PR #2, opened 2026-05-30
fix plan mode and add test

kevincodex1:
Fix plan mode and added tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "dev": "bun run src/index.ts",
-    "test": "bun test",
+    "test": "bun test ./tests --timeout 15000",
     "build": "bun build src/index.ts --compile --outfile zero"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "dev": "bun run src/index.ts",
+    "test": "bun test",
     "build": "bun build src/index.ts --compile --outfile zero"
   },
   "devDependencies": {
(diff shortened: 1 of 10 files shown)


jatmn (changes requested):
Thanks for the contribution. I took a pass through the changed paths and found a few issues that need to be addressed before this is ready.

## Findings

- [P2] Wire `/plan` into the agent behavior
  `src/tui/App.tsx:304`
  The `/plan` command still only toggles `isPlanMode`, changes the border color/status text, and prints a system message. The next user prompt is sent to `runAgent` with only `debug` and `toolsEnabled`, while `runAgent` always uses the same `DEFAULT_SYSTEM_PROMPT` and has no plan-mode option. That means enabling `/plan` does not actually make the agent plan differently, even though the UI tells the user “The agent will focus on planning before making changes.” Please pass plan-mode state into the agent prompt/options, or change the command/wording so it does not claim behavior that is not implemented.

- [P2] Make the new `bun test` script terminate reliably
  `package.json:11`
  This PR adds `bun test` as the project test script, but after installing dependencies, running the new script in the PR checkout did not complete and had to be killed after it kept running far past the expected duration. A test command that hangs makes the new validation path unusable locally and in CI. Please isolate the test that leaves the runner alive and make the suite exit cleanly.

- [P3] Preserve the friendly provider setup error
  `src/tui/App.tsx:283`
  The catch block now always displays the raw `err?.message`, which drops the previous special handling for “No LLM provider configured”. A user who submits a prompt without a configured provider now gets the lower-level exception text instead of the actionable “No provider set up. Type /provider to add one.” guidance. Please keep the previous friendly mapping for setup/auth/network errors, even if debug mode also records the full error object.

PR #3, opened 2026-05-31
Explore OpenTUI Solid TUI migration

jatmn:
## Current status

Draft exploration PR for the larger TUI direction only.

The mergeable pieces called out in review have been split into #4:

- agent/provider plumbing with `onUsage` and `onPlanUpdate`
- OpenAI-compatible `stream_options.include_usage` retry fallback
- token usage in the existing Ink UI
- todo rail fed by `update_plan`
- removal of the old manual Ink transcript slicing so terminal scrollback stays useful

That means this PR should no longer be treated as the path for those incremental wins. #4 can be reviewed and merged independently against `main`.

## What this PR is now for

This branch remains a place to evaluate the bigger product/architecture questions from review:

- Should Zero move from Ink/React to OpenTUI + Solid?
- Should the transcript move from append-style/native terminal scrollback to a full-screen pane model?
- Is the extra JSX runtime, Babel toolchain, and OpenTUI native binary surface worth the tradeoff?
- Should the build move away from the current compiled executable flow?

Those decisions should be made explicitly before this rewrite is considered mergeable.

## Implemented in this branch

- Replaces the Ink/React TUI with an OpenTUI + Solid implementation.
- Uses OpenTUI terminal primitives for panes, scrolling, mouse input, framing, and layout.
- Adds a compact persistent header with package version, provider/model, mode, and cumulative usage.
- Adds slash command popup support for `/provider`, `/clear`, `/plan`, `/build`, `/default`, `/todo`, `/tools`, `/debug`, `/debug-mode`, `/help`, and `/exit`.
- Adds provider management screens for selecting and adding local provider profiles.
- Adds native transcript scrollbox behavior with sticky-bottom output.
- Adds a resizable todo rail populated from `update_plan`.
- Ports the latest `main` plan-mode, debug, provider-error, tool schema, and test behavior into the OpenTUI implementation.

## Known review concerns before merge

- Assistant output lost the Ink markdown/code rendering path and now renders mostly as flat text.
- Tool calls lost the previous collapse/expand behavior, smart one-line summaries, and result `[more]` affordance.
- The braille header is less readable than the prior figlet ZERO mark.
- The empty-session splash/logo state was removed.
- The full-screen pane model may break expected native terminal copy/paste and scrollback behavior.
- The dependency/build direction changes should be discussed separately from UI polish.

## Validation snapshot

Last validation on this branch before the split:

- `bun install`
- `bun run tsc --noEmit`
- `bun test ./tests --timeout 15000` -> 20 pass
- `bun run build`
- `bun dist\index.js --version` -> `0.1.0`

This still needs hands-on interactive terminal testing across terminal sizes, mouse/scroll behavior, providers, and real agent sessions.

## Relationship to #4

#4 is the recommended cleanup path for the reviewer's "keep no matter what" items. This PR should remain draft until the team decides whether to continue with OpenTUI/Solid and the pane-based transcript model.


Vasanthdev2004:
Thanks for putting this together — porting main's plan-mode, debug, and provider-error work forward into the new stack instead of just reverting it was the right call, and a few pieces here are clear wins regardless of where we land (more below).

Before we go deeper on the UI, I want to flag that this PR quietly settles two decisions we hadn't actually made yet, and I'd like us to make them on purpose rather than by merge:

1. **Render stack** — Ink → OpenTUI + Solid
2. **Transcript model** — inline/append → full-screen panes (alt-screen, 60fps, mouse-driven)

Both are reasonable options, but I'm not sold on taking them as a package, and the second one is the one I keep snagging on. A full-screen panel app breaks native terminal copy-paste and scrollback, which I think is a core part of how this tool should feel. I'd like to keep the transcript scrollback-native.

### What I'd keep no matter what
- The agent/provider plumbing: `onUsage` / `onPlanUpdate` callbacks, and the `stream_options.include_usage` retry-without-it fallback. Clean and provider-agnostic.
- Token usage in the header, and the todo rail fed from `update_plan`.
- Killing the old manual `scrollOffset` slicing in favor of a real sticky-bottom scroll.

These would be great as a **separate PR** we can merge now — none of it depends on the TUI rewrite.

### UI regressions I'd want addressed before any TUI swap
- Syntax highlighting + markdown rendering is gone (`highlighter.ts` / `MessageRenderer` deleted); assistant output is now flat text. For a coding agent that's a big readability hit. (`shiki` is also left in deps but now unused.)
- Tool calls lost collapse/expand, the smart one-line summaries, and result `[more]` — long results are now hard-truncated with no way to see the rest.
- The wordmark — the braille header is hard to read; I'd like to keep the figlet ZERO mark.
- No splash / empty-session state (the old logo was removed and nothing replaced it).

### Other things to decide consciously
- This pulls in a second JSX runtime (Solid + full Babel toolchain) plus OpenTUI's per-platform native binaries. Given the "one stack across core/tools/extension" goal, that's a real direction change worth discussing on its own.
- Build moved from `--compile` (single binary) to a `dist/` bundle run via `bun dist`. Losing the standalone binary should be a deliberate choice, not a side effect.

Net: let's split out the plumbing wins so they can land now, and treat the TUI direction (append-vs-panes, Ink-vs-OpenTUI) as its own call before we commit the rewrite. Happy to prototype an append-style version with the same scroll/usage/todo wins so we can compare.

jatmn:
Thanks, this was the right framing. I split the mergeable plumbing/UI wins out of this OpenTUI rewrite into #4 so they can be reviewed against main without forcing the render-stack or transcript-model decision. This PR can stay focused on the larger Ink vs OpenTUI / append vs pane discussion.

src/tui/App.tsx
```diff
@@ -38,604 +59,1093 @@ function toFriendlyError(err: any): string {
...
-  }, []);
-
-  // Auto-scroll to bottom when new messages arrive (unless user scrolled up)
-  React.useEffect(() => {
-    // Only auto-scroll if user is near the bottom
-    if (scrollOffset <= 3) {
-      setScrollOffset(0);
-    }
-  }, [messages.length]);
+export function App(props: { onExit: () => void }) {
```
Vasanthdev2004, line 185: I checked out this branch and the mergeable pieces in #4, and want to leave one focused comment on the OpenTUI direction (rather than a long inline review — this PR is in the right state as a draft).

**Recommendation: keep this as draft, merge #4 first.**

Reasoning:

1. **The regressions you already listed in the PR body are the right ones to weigh** — the loss of the Ink markdown/code path, the loss of tool-call collapse/expand + smart summaries + `[more]`, the braille header, the dropped splash/logo, and the alt-screen-pane scrollback breakage. Each of those is a real UX downgrade and a faithful representation of how the TUI *currently* works. Inverting them in a 1,400-line rewrite is the kind of regression that's easy to underestimate until users hit it.

2. **The render-stack decision is bigger than this PR can carry.** `src/tui/App.tsx` on this branch is 1,151 lines in a single file. That's a maintainability cliff before the user-experience questions are even settled. The decision to swap Ink for OpenTUI + Solid is a commitment to a different concurrency model (signals vs. React's reconciler), a different JSX runtime (Babel toolchain + the Solid plugin you wired up in `src/index.ts:3`), a different distribution story (the `ensureSolidTransformPlugin()` call suggests a custom build), and a different terminal-IO model (Kitty keyboard protocol, mouse-movement events). Each is a fork in the road that should be made explicitly by the project owner before more code accumulates against a possibly-rejected direction.

3. **#4 captures the actual wins.** The agent-loop `onUsage` / `onPlanUpdate` callbacks, the `stream_options.include_usage` retry, the cumulative-usage header, the todo rail, and the removal of the broken manual scroll — all of these are improvements to the *current* stack that don't depend on the render-stack decision. I posted 11 inline findings on #4 (one critical multi-turn text-drop bug, one streaming 401-prefix bug, an `isUnsupportedStreamOptionsError` false-positive, and a few smaller issues). Once those are addressed, #4 should merge cleanly.

4. **A few things this branch does that #4 doesn't, worth recording if/when the OpenTUI direction is approved:**
   - Resizable todo rail width (this branch: `minTodoRailWidth=18` / `maxTodoRailWidth=65`)
   - Kitty keyboard protocol support
   - Mouse-movement events enabled
   - Package version surfaced in the header
   - `process.once('SIGINT', exit)` clean shutdown wiring that the Ink version lacks
   - Proper paste handling (`usePaste` + `normalizePastedText`) — the Ink version drops pasted newlines

   These are good ideas, but each needs the render-stack decision to be made first so they can be ported (or not) deliberately.

5. **Build is currently broken on this branch** (`bun run build` fails — `react-devtools-core` not resolvable, plus the Solid toolchain may need `bun add -d` of new deps). That's worth noting as a follow-up regardless of which direction is taken.

Closing thought: this is a well-scoped draft and your "known review concerns before merge" list is the right framing. I'd suggest converting this PR to a **discussion issue** (or closing it in favor of an issue titled "Evaluate OpenTUI + Solid migration" with this PR linked as the working branch) so the decision is made on its own merits, separate from any merge-blocker. The plumbing in #4 then has a clean path to land without waiting on a 1,400-line TUI rewrite to be either blessed or rejected.

src/tui/App.tsx
```diff
@@ -38,604 +60,1338 @@ function toFriendlyError(err: any): string {
...
+        </text>
+        <text fg={colors.subtle} wrapMode="none" truncate>
+          p:{props.usage.promptTokens} c:{props.usage.completionTokens} t:{props.total}
+        </text>
+      </box>
+    </box>
+  );
+}
+
+function ZeroLogo(props: { compact?: boolean }) {
```
Vasanthdev2004, line 841: **Re-review of `Restore TUI rendering affordances` (8ef78c0).**

Checked out the latest and walked through the four new commits (`Fix terminal paste and cancel shortcuts` → `Add clear command and build version header` → `Refine OpenTUI layout and transcript behavior` → `Restore TUI rendering affordances`). Substantive progress.

### What got restored (the things I flagged in the PR body)
- ✅ **Figlet ZERO logo** — both compact and full variants (App.tsx:841-870), replaced the braille header.
- ✅ **Tool-call collapse/expand + `[more]/[less]`** — `<ToolCall>` component (App.tsx:1082-1148) is back, with `[x]/[>]` status marks, one-line `toolSummary(...)` label, and toggles wired via OpenTUI's real `onMouseDown` handler. Unlike the Ink version on main, the click handler actually works here because OpenTUI's `<text>` element supports mouse events.
- ✅ **Splash / empty-session** — `<EmptySession />` fallback inside the `<Show when={messages.length > 0}>` (App.tsx:911).
- ✅ **Sticky-bottom native scroll** — `<scrollbox stickyScroll stickyStart="bottom">` (App.tsx:905) gives a real "always pinned to bottom" feel without the broken `scrollOffset` slicing from main.
- ✅ **Resizable todo rail** — with `onMouseDrag` for live width adjustment. Nice.
- ✅ **Clear command + version header** — operational polish.

The commit message is accurate; this is genuinely a restoration effort, not just a re-shuffle.

### What I want to flag for the next round

**1. The two provider-error fixes from my PR #4 review aren't in this branch.**

I confirmed by running the test suite: applying PR #4's regression tests (`tests/openai-provider.test.ts`) against the current `pr3_2` source gives **22 pass, 2 fail**:

- `does not retry unrelated unknown-parameter errors` — fails because `src/providers/openai.ts:30` still uses the broad `errorText.includes('stream_options') || … 'unknown parameter' || …` keyword list. The fix from #4's r2 is a single regex: `/stream[ _-]?options?/`.
- `categorizes mid-stream authentication errors` — fails because the streaming catch at `src/providers/openai.ts:196` is still the old `throw new Error(\`Provider returned error during streaming: ${message}\`);`. The fix from #4's r2 is `throw formatProviderCreateError(error);`.

Both are one-line changes. The classification helper `formatProviderCreateError` is already defined on this branch (line 201) — it's just not called from the streaming catch. These fixes would also make the OpenTUI TUI's `toFriendlyError` mapper at App.tsx:39 produce the same helpful "Authentication failed - check your API key" message that the Ink TUI now produces.

**2. The `tc.name === 'update_plan'` special case in the agent loop is back in a new shape.**

Looking at `src/agent/loop.ts` on this branch — does it still have the old hard-coded `if (tc.name === 'update_plan')` check, or has it been ported to the `Tool.onAfterExecute` pattern from PR #4 r2? If the former, this branch is going to fork from #4's fix the moment #4 merges. Worth a rebase / merge of #4 before more code accumulates against the unfixed shape.

**3. Build is presumably still broken on this branch.**

`bun run build` requires the `react-devtools-core` import that `ink` makes — and this branch is on OpenTUI + Solid, not Ink, so that may not apply here. But the new `@opentui/core` / `@opentui/solid` packages presumably have their own build-time concerns (native binaries per the PR body). Worth running `bun run build` here and confirming before merge consideration.

### Bigger-picture question

The restoration work in this PR is real, and the OpenTUI affordances (mouse-driven rail resizing, sticky-bottom scroll, native mouse events on text) are genuine wins that the Ink version doesn't have. But the OpenTUI vs Ink decision is still undecided at the project level, and #4 was based on a recommendation to keep this draft until that decision is made.

Has the team reached a decision? If yes, this branch has matured enough to be the merge target. If no, the most useful next step is probably to **rebase this onto #4** (so the agent-loop fixes flow both ways) and then leave the branch as the proof-of-concept for the OpenTUI direction without merging.

jatmn:
ink issues is a sticky one.. because that's remnants related to #4, which needs to land first so i can properly deteather it after

PR #5, opened 2026-06-01
Add TOML-backed provider catalog

jatmn:
## Summary

This PR adds provider catalog support to Zero, moving built-in provider and model metadata into TOML catalog files and wiring those definitions into provider setup, runtime provider creation, model discovery, and build output.

## Changes

- Adds a provider catalog loader that reads provider definitions from `src/providers/catalog/definitions/*.toml`.
- Adds global model metadata from `src/providers/catalog/models/*.toml`.
- Supports provider, gateway, and localhost provider kinds.
- Adds built-in catalog entries for OpenAI, OpenRouter, OpenGateway, Groq, and Ollama.
- Adds global model families for OpenAI/GPT, Anthropic, Google, and DeepSeek models.
- Resolves `globalModelId` references so gateways can reuse top-level model metadata without duplicating it.
- Adds first-party provider behavior so providers can automatically expose owned global models.
- Normalizes credential env vars from top-level provider definitions into setup, validation, and preset metadata.
- Keeps pricing on provider-specific model routes and currency on the provider definition.
- Adds provider factory support for catalog-backed profiles, including transport config resolution and per-model transport overrides.
- Adds auth/header helpers for provider-specific auth headers, raw auth schemes, non-standard auth header names, and env-var-backed custom headers.
- Adds OpenAI-compatible model discovery through `/models`.
- Adds Ollama discovery through `/api/tags`.
- Adds discovery caching with TTL support, stale-cache fallback, and API-key-partitioned cache keys when discovery requires auth.
- Updates `/provider` setup UI so users can pick catalog providers, enter credentials, discover/select models, or add a custom OpenAI-compatible provider.
- Extends provider config loading so `ZERO_PROVIDER_COMMAND` can return a `provider_id` and inherit catalog defaults.
- Adds a build script that embeds provider catalog TOML files into bundled output.
- Updates tool schemas/tests alongside the provider work.
- Adds comprehensive provider catalog tests covering schema guardrails, catalog parsing, model resolution, provider factory behavior, discovery, caching, and provider command parsing.

## Runtime notes

- Runtime calls currently support `openai-compatible` transports.
- `anthropic-compatible` descriptors are recognized in catalog metadata but rejected by the runtime factory until that transport is implemented.
- Static catalog models are returned before discovered models.
- Duplicate discovered model API names are skipped when a static model route already exists.
- OpenGateway credentials are scoped to `OPENGATEWAY_API_KEY`, preventing accidental OpenRouter credential reuse.

## Verification

- `bun run build`


jatmn:
Ill note that the existing providers and models in this was for example purposes, we should flush those out properly with related pr's after this lands.

PR #10, opened 2026-06-02
Add M0 CI scripts baseline

anandh8x:
## Summary
- Add stable M0 scripts for test, typecheck, binary build, and binary smoke checks
- Add Linux/macOS/Windows GitHub Actions smoke matrix
- Pin Bun version metadata and document setup/check commands

## Verification
- bun install --frozen-lockfile
- bun test
- bun run typecheck
- bun run build
- bun run smoke:build
- git diff --check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Test
-        run: bun test
+        run: bun run test
 
       - name: Typecheck
         run: bun run typecheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  smoke:
+    name: Smoke (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.14
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Test
+        run: bun test
+
+      - name: Typecheck
+        run: bun run typecheck
+
+      - name: Build binary
+        run: bun run build
+
+      - name: Smoke binary
+        run: bun run smoke:build
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## Setup
 
 ```bash
-bun install
+bun install --frozen-lockfile
 ```
 
 ## Run
@@ -21,4 +21,4 @@ bun run build
 bun run smoke:build
 ```
 
-This project uses Bun 1.3.14.
+Bun version is pinned in `package.json`.
--- a/README.md
+++ b/README.md
@@ -1,15 +1,24 @@
 # zero
 
-To install dependencies:
+## Setup
 
 ```bash
 bun install
 ```
 
-To run:
+## Run
 
 ```bash
-bun run index.ts
+bun run dev
 ```
 
-This project was created using `bun init` in bun v1.3.11. [Bun](https://bun.com) is a fast all-in-one JavaScript runtime.
+## Checks
+
+```bash
+bun test
+bun run typecheck
+bun run build
+bun run smoke:build
+```
+
+This project uses Bun 1.3.14.
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "dev": "bun run src/index.ts",
     "test": "bun test ./tests --timeout 15000",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "bunx tsc --noEmit",
     "build": "bun run scripts/build.ts",
     "smoke:build": "bun run scripts/smoke-build.ts"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "zero",
   "version": "0.1.0",
+  "packageManager": "bun@1.3.14",
   "module": "src/index.ts",
   "type": "module",
   "bin": {
@@ -9,14 +10,15 @@
   "scripts": {
     "dev": "bun run src/index.ts",
     "test": "bun test ./tests --timeout 15000",
-    "build": "bun build src/index.ts --compile --outfile zero"
+    "typecheck": "tsc --noEmit",
+    "build": "bun run scripts/build.ts",
+    "smoke:build": "bun run scripts/smoke-build.ts"
   },
   "devDependencies": {
     "@types/bun": "latest",
-    "@types/react": "^19.2.15"
-  },
-  "peerDependencies": {
-    "typescript": "^5"
+    "@types/react": "^19.2.15",
+    "react-devtools-core": "^7.0.1",
+    "typescript": "^5.9.3"
   },
   "dependencies": {
     "commander": "^15.0.0",
--- a/scripts/smoke-build.ts
+++ b/scripts/smoke-build.ts
@@ -24,7 +24,16 @@ if (exitCode !== 0) {
   process.exit(exitCode);
 }
 
-const expectedVersion = JSON.parse(packageText).version;
+let expectedVersion: string;
+
+try {
+  expectedVersion = JSON.parse(packageText).version;
+} catch (error) {
+  const message = error instanceof Error ? error.message : String(error);
+  console.error(`Failed to parse package.json: ${message}`);
+  process.exit(1);
+}
+
 const actualVersion = stdout.trim();
 
 if (actualVersion !== expectedVersion) {
--- a/scripts/smoke-build.ts
+++ b/scripts/smoke-build.ts
@@ -1,0 +1,35 @@
+import { zeroArtifactName, zeroArtifactPath } from './artifact';
+
+const artifact = Bun.file(zeroArtifactPath);
+
+if (!(await artifact.exists())) {
+  console.error(`Build artifact not found: ${zeroArtifactName}`);
+  process.exit(1);
+}
+
+const child = Bun.spawn([zeroArtifactPath, '--version'], {
+  stderr: 'pipe',
+  stdout: 'pipe',
+});
+
+const [exitCode, stdout, stderr, packageText] = await Promise.all([
+  child.exited,
+  new Response(child.stdout).text(),
+  new Response(child.stderr).text(),
+  Bun.file('package.json').text(),
+]);
+
+if (exitCode !== 0) {
+  console.error(stderr.trim() || `${zeroArtifactName} --version exited with ${exitCode}`);
+  process.exit(exitCode);
+}
+
+const expectedVersion = JSON.parse(packageText).version;
+const actualVersion = stdout.trim();
+
+if (actualVersion !== expectedVersion) {
+  console.error(`Expected ${zeroArtifactName} --version to print ${expectedVersion}, got ${actualVersion}`);
+  process.exit(1);
+}
+
+console.log(`${zeroArtifactName} smoke check passed (${actualVersion})`);
--- a/bun.lock
+++ b/bun.lock
@@ -18,9 +18,8 @@
       "devDependencies": {
         "@types/bun": "latest",
         "@types/react": "^19.2.15",
-      },
-      "peerDependencies": {
-        "typescript": "^5",
+        "react-devtools-core": "^7.0.1",
+        "typescript": "^5.9.3",
       },
     },
   },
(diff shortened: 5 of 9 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- CI uses `bun test` directly while the package script uses `bun test ./tests --timeout 15000`. Not worth blocking, but using `bun run test` in CI would keep the local and CI contract identical if the script changes later.

### Looks Good

- Fixes the existing binary build blocker by adding `react-devtools-core` to dev dependencies.
- Adds stable `typecheck`, `build`, and `smoke:build` scripts instead of leaving reviewers to remember raw commands.
- Build artifact naming is platform-aware (`zero.exe` on Windows, `zero` elsewhere), and `.gitignore` already covers both outputs.
- Smoke check validates the compiled binary actually runs and prints the package version, not just that a file exists.
- Validation in an isolated worktree passed: `bun install --frozen-lockfile`, `bun test`, `bun run typecheck`, `bun run build`, `bun run smoke:build`, and `git diff --check`.

Verdict: Approve — clean M0 CI/build baseline and it closes the build-smoke gap without touching runtime behavior.

Vasanthdev2004 (approved):
Approve — clean M0 CI/build baseline. Validated install, tests, typecheck, build, smoke build, and diff check in an isolated worktree.

gnanam1990 (approved):
Approve - validation passed in an isolated worktree: git diff --check, bun install --frozen-lockfile, bun test ./tests --timeout 15000, tsc --noEmit, bun run build, bun run smoke:build, and ./zero --help. No blockers found.

anandh8x (commented):
# Code Review: PR #10 — `Add M0 CI scripts baseline`

**Author:** @anandh8x → `feat/m0-ci-scripts` → `main`
**Scope:** 7 files, +118/-13. Adds `package.json` scripts, `scripts/artifact.ts` + `scripts/build.ts` + `scripts/smoke-build.ts`, `.github/workflows/ci.yml`, README updates, `bun.lock` refresh.
**Verdict:** Tight, well-scoped baseline. The cross-PR coordination on `package.json` scripts and the `--version` assumption in the smoke check are the only real concerns.

---

## Important

### 1. `package.json` script conflicts with PR #11
The `build` and `typecheck` scripts diverge from the open PR #11 (`codex/p0-foundation-tools`):

| script | PR #10 (`package.json:13-15`) | PR #11 (`package.json:11-13`) |
|---|---|---|
| `typecheck` | `tsc --noEmit` | `bunx tsc --noEmit` |
| `build` | `bun run scripts/build.ts` | `bun build src/index.ts --compile --outfile zero` |
| `smoke:build` | `bun run scripts/smoke-build.ts` | (not present) |

Both PRs target `main`. Whichever lands second will conflict on `package.json`. The `typecheck` difference is the more concerning one: `tsc --noEmit` relies on a globally installed `tsc`, which the `oven-sh/setup-bun@v2` step in the workflow does not provide. On a fresh CI runner, PR10's `typecheck` will fail unless a global `tsc` is installed, whereas PR11's `bunx tsc --noEmit` resolves it via the local dev dependency. The CI workflow at `ci.yml:37` will inherit whichever version wins.

**Ask:** Update `typecheck` to `bunx tsc --noEmit` (matches PR11, robust on CI). Coordinate with @Vasanthdev2004 to align `build` — easiest fix is to keep both PRs pointing at `scripts/build.ts` so the cross-platform artifact naming in `scripts/artifact.ts` is the single source of truth. If the intent of PR11's inline `build` is "no TS wrapper," pick one and delete the other.

### 2. `smoke-build.ts` assumes `--version` is implemented
`scripts/smoke-build.ts:10-33` runs `${zeroArtifactPath} --version` and asserts the output equals `package.json`'s `version`. The `package.json` `bin` entry is `"zero": "src/index.ts"` (line 7) — the entry script. There's no test or diff in this PR that proves the compiled `zero` binary implements `--version`. If `src/index.ts` doesn't print its version on `--version`, the smoke check fails on the first CI run with no useful diagnostic beyond "Expected zero --version to print 0.1.0, got <whatever>".

**Ask:** Either (a) verify that `src/index.ts` already implements `--version` and add a unit test for it, or (b) add the `--version` handling to a follow-up PR before this one merges, and gate the smoke step on it. Also consider passing `--version` to a subcommand like `zero --version` rather than treating it as a top-level flag — that depends on how `commander` is wired.

### 3. `smoke-build.ts:27` — uncaught `JSON.parse` failure
```ts
const expectedVersion = JSON.parse(packageText).version;
```
If `package.json` is malformed (shouldn't happen, but `bun install` can produce unexpected states), this throws an uncaught error and the script exits non-zero with a stack trace, masking the real smoke check failure. Cheap to harden:
```ts
let expectedVersion: string;
try {
  expectedVersion = JSON.parse(packageText).version;
} catch (err: any) {
  console.error(`Failed to parse package.json: ${err.message}`);
  process.exit(1);
}
```

---

## Suggestions

### 4. CI actions not pinned to commit SHAs
`.github/workflows/ci.yml:23, 26` use `actions/checkout@v4` and `oven-sh/setup-bun@v2` as floating tags. Standard supply-chain hardening is to pin to commit SHAs (e.g., `actions/checkout@<full-sha>`) and reference the version in a comment. Not blocking for a M0 baseline, but worth adding before this workflow runs against untrusted PRs.

### 5. No `actions/cache` for `bun install`
`ci.yml:31` runs `bun install --frozen-lockfile` on every matrix OS without caching the install dir. `oven-sh/setup-bun@v2` does not cache by default. For 3-OS × N-PRs, this adds up. Add a `cache: true` option to the setup action, or an explicit `actions/cache` step keyed on `bun.lock`. Minor — M0 baseline can defer.

### 6. README documents `bun install` instead of `bun install --frozen-lockfile`
`README.md:6` shows `bun install`, but the CI workflow at `ci.yml:31` uses `bun install --frozen-lockfile`. CI reproducibility should match the documented setup. Either change the README or change the workflow. The PR body advertises "Pin Bun version metadata and document setup/check commands" — pinning the lockfile is part of that.

### 7. `Bun 1.3.14` mentioned twice
`README.md:24` and `package.json:4` both say `Bun 1.3.14`. The `packageManager` field is the source of truth and `oven-sh/setup-bun` reads it; the README line is now redundant. Drop the README line or rewrite as *"Bun version pinned in `package.json#packageManager`."*

### 8. No tests for the build / smoke scripts
The scripts are short enough that they're "obviously correct," but a one-line test that `zeroArtifactName` returns `zero.exe` on `win32` and `zero` elsewhere would lock in the cross-platform contract. Currently the only signal that this works is the matrix CI run.

---

## Questions for the author

1. Is the `scripts/build.ts` wrapper long-term, or is the goal to inline `bun build` into `package.json` and delete the wrapper once the cross-platform artifact naming is no longer needed? Affects how to coordinate with PR11.
2. Should `smoke:build` run on `windows-latest` and `macos-latest` too, given the matrix includes them? The current workflow at `ci.yml:42` does run it on all three — just confirming that's intentional.
3. The CI matrix doesn't pin the OS runner versions. Floating `ubuntu-latest` will silently move to a new image over time. Worth pinning to a specific tag (`ubuntu-22.04`) for reproducibility?

---

## Validation

- `bun.lock` (+11/-4) — cannot review without the diff. The PR body doesn't mention lockfile review, so trusting `bun install --frozen-lockfile` to be authoritative.
- The cross-platform `artifact.ts` correctly handles `win32`. ✓
- `await $\`bun build ...\`` correctly propagates exit codes via thrown errors. ✓

---

## Recommendation

**Request changes** (or coordinate merge order with PR11 before approving). The `package.json` conflict is the only blocker — the rest are small enough to address in a follow-up. Once `typecheck` is aligned and the `--version` contract is confirmed, this is a clean, mergeable M0 baseline.

Nice work — the script split into `artifact.ts` / `build.ts` / `smoke-build.ts` is clean and the file-naming contract is a good pattern.

package.json
```diff
@@ -9,14 +10,15 @@
   "scripts": {
     "dev": "bun run src/index.ts",
     "test": "bun test ./tests --timeout 15000",
-    "build": "bun build src/index.ts --compile --outfile zero"
+    "typecheck": "tsc --noEmit",
```
anandh8x: **Important — conflict with PR #11.** PR11 (`codex/p0-foundation-tools`) defines `typecheck` as `bunx tsc --noEmit`. This PR defines it as `tsc --noEmit`. The CI workflow at `.github/workflows/ci.yml:37` will use whichever lands second, and `tsc` (without `bunx`) is not on PATH in a fresh `oven-sh/setup-bun@v2` runner. Recommend changing to `bunx tsc --noEmit` to match PR11 and to be robust on CI without a global tsc install. Coordinate with @Vasanthdev2004 on `build` too — pick one (inline `bun build ...` vs the `scripts/build.ts` wrapper) before either PR merges.

scripts/smoke-build.ts
```diff
@@ -0,0 +1,35 @@
...
+  new Response(child.stderr).text(),
+  Bun.file('package.json').text(),
+]);
+
+if (exitCode !== 0) {
+  console.error(stderr.trim() || `${zeroArtifactName} --version exited with ${exitCode}`);
+  process.exit(exitCode);
+}
+
+const expectedVersion = JSON.parse(packageText).version;
```
anandh8x: **Important — `--version` contract assumed.** This asserts `${zeroArtifactName} --version` prints `package.json` `version`, but nothing in this PR diff proves that `src/index.ts` (the `bin` entry at `package.json:7`) actually implements `--version`. If it doesn't, the smoke check fails on the first CI run with no useful diagnostic. Verify the entry script implements `--version` and add a unit test for it before relying on it in CI. Consider also wiring it through `commander` rather than as a top-level flag, depending on how the CLI is structured.

anandh8x: **Suggestion — uncaught `JSON.parse` error.** If `package.json` is malformed (unlikely but possible if `bun install` produces an unexpected state), the uncaught throw will surface a stack trace and mask the real smoke check failure. Wrap in try/catch and `process.exit(1)` with a clear message.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
anandh8x, line 23: **Suggestion — pin actions to commit SHAs.** `actions/checkout@v4` and `oven-sh/setup-bun@v2` are floating tags. For supply-chain hardening (especially on a workflow that will eventually run against untrusted PRs), pin to commit SHAs and reference the human version in a comment. Standard practice for production CI. Not blocking for M0.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.14
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
```
anandh8x, line 31: **Suggestion — no install cache.** `bun install --frozen-lockfile` runs on every matrix OS × PR. `oven-sh/setup-bun@v2` does not cache by default. For 3-OS × N-PRs, this adds up. Either pass `cache: true` to the setup action or add an explicit `actions/cache` step keyed on `bun.lock`. M0 baseline can defer.

README.md
```diff
@@ -1,15 +1,24 @@
 # zero
 
-To install dependencies:
+## Setup
 
 ```bash
 bun install
```
anandh8x: **Suggestion — README uses `bun install` while CI uses `--frozen-lockfile`.** `README.md:6` says `bun install` but `.github/workflows/ci.yml:31` uses `bun install --frozen-lockfile`. For reproducibility, the documented setup command should match what CI runs. Recommend changing the README to `bun install --frozen-lockfile`.

README.md
```diff
@@ -1,15 +1,24 @@
...
+## Checks
+
+```bash
+bun test
+bun run typecheck
+bun run build
+bun run smoke:build
+```
+
+This project uses Bun 1.3.14.
```
anandh8x: **Suggestion — duplicate Bun version pin.** `README.md:24` says `This project uses Bun 1.3.14` and `package.json:4` has `"packageManager": "bun@1.3.14"`. The `packageManager` field is the source of truth and `oven-sh/setup-bun` reads it; the README line is now redundant. Drop the README line or rewrite as a pointer.

anandh8x:
Addressed the actionable review items in f33901a:

- Changed typecheck to `bunx tsc --noEmit` to align with PR #11 and fresh CI runners.
- Changed CI test step to `bun run test` so CI uses the package script contract.
- Added `tests/cli-version.test.ts` to prove `zero --version` prints the package version.
- Hardened `scripts/smoke-build.ts` package.json parsing with a clear error message.
- Added `tests/build-scripts.test.ts` for platform artifact naming.
- Updated README setup to `bun install --frozen-lockfile` and made packageManager the Bun version source of truth.

Keeping `scripts/build.ts` intentionally for cross-platform artifact naming; `smoke:build` is intended to run on all three matrix OSes. Action SHA pinning, cache tuning, and fixed runner versions are good follow-up items for security/perf hardening rather than this M0 baseline.

gnanam1990 (approved):
Approve - new head f33901a addresses the prior actionable feedback. Validation passed in a fresh isolated worktree: git diff --check, bun install --frozen-lockfile, bun test ./tests --timeout 15000 (52/52), tsc --noEmit, bun run build, bun run smoke:build, and ./zero --help. CI is also green across ubuntu, macOS, and Windows smoke jobs. No blockers found.

PR #21, opened 2026-06-03
feat: add Zero session event store

gnanam1990:
## Summary

- Adds a backend `zero-sessions` module with typed session metadata and append-only JSONL event persistence.
- Exposes `ZeroSessionEventStore` and `defaultZeroSessionRoot` for future resume/search/cost integrations.
- Supports safe session id validation, create/read/list operations, event append/read, recent-first listing, bounded payload search, corrupt event-line errors, and duplicate-session protection.

## Tests

- `npx --yes bun install --frozen-lockfile`
- `npx --yes bun run build`
- `./zero --help`
- `./zero providers current`
- `git diff --check`

## Reviewers

@Vasanthdev2004 @anandh8x please review.

### DIFF
--- a/src/zero-sessions/index.ts
+++ b/src/zero-sessions/index.ts
@@ -1,0 +1,12 @@
+export { ZeroSessionEventStore, defaultZeroSessionRoot } from './store';
+export type {
+  AppendZeroSessionEventInput,
+  CreateZeroSessionInput,
+  DefaultZeroSessionRootOptions,
+  ZeroSessionEvent,
+  ZeroSessionEventStoreOptions,
+  ZeroSessionEventType,
+  ZeroSessionMetadata,
+  ZeroSessionSearchHit,
+  ZeroSessionSearchOptions,
+} from './types';
--- a/src/zero-sessions/store.ts
+++ b/src/zero-sessions/store.ts
@@ -1,0 +1,256 @@
+import { appendFile, mkdir, readFile, readdir, writeFile } from 'fs/promises';
+import { homedir } from 'os';
+import { join } from 'path';
+import type {
+  AppendZeroSessionEventInput,
+  CreateZeroSessionInput,
+  DefaultZeroSessionRootOptions,
+  ZeroSessionEvent,
+  ZeroSessionEventStoreOptions,
+  ZeroSessionMetadata,
+  ZeroSessionSearchHit,
+  ZeroSessionSearchOptions,
+} from './types';
+
+const SESSION_ID_PATTERN = /^[A-Za-z0-9][A-Za-z0-9_-]{0,127}$/;
+const METADATA_FILE = 'metadata.json';
+const EVENTS_FILE = 'events.jsonl';
+
+export function defaultZeroSessionRoot(options: DefaultZeroSessionRootOptions = {}): string {
+  const env = options.env ?? process.env;
+  const dataHome = env.XDG_DATA_HOME?.trim();
+  const home = env.HOME?.trim() || homedir();
+  const baseDir = dataHome || join(home, '.local', 'share');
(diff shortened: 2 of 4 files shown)


anandh8x (approved):
# Code Review: PR #21 — `feat: add Zero session event store`

**Author:** @gnanam1990 → `main`
**Head:** `00345547`
**Scope:** 4 new files, +577/-0. Append-only JSONL session store under `~/.local/share/zero/sessions/<id>/` with typed metadata and event persistence.

**PRD alignment:** Sessions are M2 scope per v2 PRD. Owner (Gnanam) is correct.

---

## Verdict: Approve

Clean design, well-tested, no blockers. The store is properly abstracted with injectable `now` and configurable root dir. All main paths plus error cases are covered by tests.

---

## Design notes (non-blocking, for awareness)

### `appendEvent` has a TOCTOU race on metadata
`src/zero-sessions/store.ts:103-122` reads `eventCount` from metadata, increments, appends to JSONL, then writes metadata back. Two concurrent appends from different async contexts would produce duplicate sequence numbers and one would overwrite the other's metadata. Not an issue for single-process CLI use (cooperative async), but worth noting if sessions ever go multi-process.

### `searchEvents` loads every session's full JSONL into memory
Brute-force all-session text scan across all events. Fine for M2 (short-lived sessions, few events). Won't scale to hundreds of sessions without an index.

### No compaction/rotation on JSONL files
Events accumulate unboundedly per session. Fine for M2 since sessions are ephemeral and not expected to have thousands of events.

---

## What I like

- **XDG-compliant root path** — checks `XDG_DATA_HOME` first, falls back to `~/.local/share/zero/sessions`. Right for Linux/macOS.
- **Session ID validation** — `SESSION_ID_PATTERN` rejects `../bad` and path-traversal patterns before any file write. Tested.
- **`sequenceClock` helper** — injectable `now` makes tests deterministic across all test cases.
- **Duplicate protection** — EEXIST from `mkdir` gives a clear "already exists" message. Events file uses `flag: 'wx'` for defense-in-depth.
- **Corrupt event detection** — `"Invalid JSON in Zero session <id> events.jsonl at line <n>"` with specific file and line number. Tested.
- **`extractSearchText` recursively flattens nested objects** — handles arbitrary payload shapes in search.
- **`ZeroSessionEventType` uses `(string & {})`** — extensible: callers can pass custom event types without type errors.
- **Tests cover:** create/append/read/list, sort order, duplicate protection, search with context + limit, invalid IDs, corrupt lines, and root path resolution. 7 tests, all thorough.

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- `searchEvents(..., { limit: 0 })` currently returns the first match instead of zero matches because the limit check happens only after pushing a hit. I verified this with a quick probe. Not blocking for the current M2 store because normal callers will use a positive limit or omit it, but worth tightening before exposing search limits directly in CLI/UI.

### Looks Good

- Clean M2 session-store slice: append-only JSONL events, typed metadata, safe session-id validation, duplicate-session protection, corrupt JSONL line errors, and XDG-style default root are all covered.
- The store stays backend-only and injectable (`rootDir`, `now`), so it is easy to wire into resume/search/cost later without coupling it to the TUI.
- PR branch validation passed: install, typecheck, full tests, build, smoke build, CLI help/current-provider smoke, and diff whitespace check.
- Test result: 151 pass / 0 fail / 556 expect calls.
- Scope is focused: 4 files, +577/-0.

Verdict: Approve — clean session event-store foundation with only a small limit-zero edge case to follow up.

Vasanthdev2004 (approved):
Approved. Clean M2 session event-store foundation; validation passed locally.

PR #40, opened 2026-06-03
feat: add MCP permission grants

gnanam1990:
## Summary

Adds the M4 MCP permissions backend so Zero can persist trusted MCP server/tool grants and manage them from the CLI.

## What changed

- Added `ZeroMcpPermissionStore` for durable per-server and per-tool MCP grants.
- Enforced server identity matching so stale grants do not apply after a server transport changes.
- Added autonomy ceilings (`low`, `medium`, `high`) for persistent approvals.
- Added `zero mcp permissions list`, `revoke`, and `clear` CLI commands with JSON output support.
- Wired MCP tool metadata into the agent loop so matching persistent grants can skip repeated ask-mode prompts.
- Kept one-time session approval grants separate from persistent MCP grants.
- Made MCP registry registration atomic when duplicate generated tool names are detected.
- Preserved custom tool JSON schemas while preparing provider tool definitions.

## Validation

- `npx --yes bun install --frozen-lockfile`
- `npx --yes bun run typecheck`
- `npx --yes bun test ./tests --timeout 15000` (`253 pass`, `0 fail`)
- `npx --yes bun run build`
- `npx --yes bun run smoke:build`
- `./zero --help`
- `./zero mcp --help`
- `./zero mcp permissions --help`
- `ZERO_MCP_PERMISSIONS_PATH=/tmp/zero-mcp-permissions-pr-smoke.json ./zero mcp permissions list --json`
- `git diff --check HEAD`

## Summary by CodeRabbit

* **New Features**
  * Manage persistent MCP permission grants via new `zero mcp permissions` CLI commands: list permissions, revoke server or tool grants, and clear all approvals.
  * Control MCP tool execution through autonomy level settings (low, medium, high).

* **Bug Fixes**
  * Fixed JSON schema object mutations during tool definition preparation.

PR #43, opened 2026-06-04
feat: add Zero hook backend

gnanam1990:
## Summary
- Add a Zero hook backend for layered user/project hook configuration, enablement, matching, and diagnostics.
- Add hook audit JSONL persistence for started/completed hook execution events with stable sequence ordering.
- Expose `zero hooks list` in text and JSON formats so CLI users can inspect configured hooks and load diagnostics.

## Details
- Supports default hook paths from XDG config/data homes with project-level overrides from `.zero/hooks.json`.
- Keeps valid hook entries available while reporting malformed or duplicate configuration as diagnostics.
- Adds config-store helpers for list/upsert/remove/enable flows and selector helpers for event/tool matching.
- Adds focused coverage for config layering, persistence, selection, audit storage, and CLI output.

## Validation
- `bunx tsc --noEmit`
- `bun test ./tests --timeout 15000` (274 pass)
- `bun run build`
- `bun run smoke:build`
- `./zero --help`
- `./zero hooks list`
- `./zero hooks list --json`

## Summary by CodeRabbit

* **New Features**
  * Added a `zero hooks list` CLI subcommand with JSON and formatted text output
  * Hook configuration management: add, remove, enable/disable, and layered user/project settings
  * Audit logging for hook execution events with timestamps and sequence numbers
  * Configuration diagnostics to report errors and conflicts

* **Tests**
  * End-to-end tests covering config, audit, CLI output, concurrency, and secret redaction

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,10 @@ import { formatZeroDoctorReport, runZeroDoctor, type ZeroDoctorReport } from './
 import { redactZeroErrorMessage, redactZeroSecrets } from './zero-redaction';
 import { formatZeroSearchResult, searchZeroSessions } from './zero-search';
 import {
+  formatZeroHookList,
+  loadZeroHooksConfig,
+} from './zero-hooks';
+import {
   ZeroMcpClientManager,
   ZeroMcpPermissionStore,
   createZeroMcpToolName,
@@ -420,6 +424,29 @@ program
       }
     } catch (err: unknown) {
       console.error(`[zero] Plugin list failed: ${redactZeroErrorMessage(err)}`);
+      process.exitCode = 1;
+    }
+  });
+
+program
+  .command('hooks')
+  .description('Inspect Zero hook configuration and audit data')
+  .command('list')
+  .description('List configured Zero hooks')
+  .option('--json', 'Print hook config as JSON')
+  .action(async (options: { json?: boolean }) => {
+    try {
+      const result = await loadZeroHooksConfig();
+      if (options.json) {
+        console.log(JSON.stringify(redactZeroSecrets({
+          hooks: result.config,
+          diagnostics: result.diagnostics,
+        }), null, 2));
+      } else {
+        console.log(formatZeroHookList(result.config, result.diagnostics));
+      }
+    } catch (err: unknown) {
+      console.error(`[zero] Hook list failed: ${redactZeroErrorMessage(err)}`);
       process.exitCode = 1;
     }
   });
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,7 +11,7 @@ import {
   type ZeroConfigInspectionReport,
 } from './zero-config-inspection';
 import { formatZeroDoctorReport, runZeroDoctor, type ZeroDoctorReport } from './zero-doctor';
-import { redactZeroErrorMessage, redactZeroSecrets } from './zero-redaction';
+import { redactZeroErrorMessage, redactZeroSecrets, redactZeroString } from './zero-redaction';
 import { formatZeroSearchResult, searchZeroSessions } from './zero-search';
 import {
   formatZeroHookList,
@@ -443,7 +443,7 @@ program
           diagnostics: result.diagnostics,
         }), null, 2));
       } else {
-        console.log(formatZeroHookList(result.config, result.diagnostics));
+        console.log(redactZeroString(formatZeroHookList(result.config, result.diagnostics)));
       }
     } catch (err: unknown) {
       console.error(`[zero] Hook list failed: ${redactZeroErrorMessage(err)}`);
--- a/src/zero-hooks/audit.ts
+++ b/src/zero-hooks/audit.ts
@@ -1,5 +1,6 @@
 import { appendFile, mkdir, readFile } from 'fs/promises';
 import { dirname, resolve } from 'path';
+import { redactZeroString } from '../zero-redaction';
 import { resolveZeroHookPaths } from './config';
 import type {
   AppendZeroHookCompletedInput,
@@ -17,6 +18,7 @@ export interface ZeroHookAuditStoreOptions {
 export class ZeroHookAuditStore {
   private readonly auditPath: string;
   private readonly now: () => Date;
+  private appendQueue: Promise<unknown> = Promise.resolve();
 
   constructor(options: ZeroHookAuditStoreOptions = {}) {
     this.auditPath = resolve(options.auditPath ?? resolveZeroHookPaths().auditPath);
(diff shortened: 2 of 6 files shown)


anandh8x (approved):
## What's good

- Layered config (user + project) with `XDG_CONFIG_HOME` / `XDG_DATA_HOME` resolution, overrideable via `cwd` / `env` injection — the test harness exploits this cleanly.
- Zod schema validation with structured diagnostics (`io` / `json` / `schema` / `duplicate`) — useful for surfacing errors without failing the whole load.
- Project-overrides-user semantics with a `duplicate` diagnostic when an id collides, and a stable alphabetical sort of merged hooks.
- Atomic config writes via temp file + `rename` in `writeZeroHooksConfig` (`src/zero-hooks/config.ts:130-136`).
- `redactZeroSecrets` is applied to the JSON output of `hooks list --json` (`src/index.ts:444`), consistent with the other redaction boundaries in the CLI.
- Test coverage spans config layering, persistence round-trip via the store, selector matching, audit JSONL, and end-to-end CLI output as a subprocess.

1. **Audit append is not serialized.** `ZeroHookAuditStore.append` (`src/zero-hooks/audit.ts:60-72`) computes `sequence` via `readEvents().length + 1` and then `appendFile`s. Two concurrent `appendStarted` / `appendCompleted` calls would compute the same sequence and the second writer's line would land out of order relative to sequence numbers — so the "stable sequence ordering" promised in the PR body isn't actually stable under concurrency. The `ZeroMcpPermissionStore` in #40 used a `withWriteLock` promise chain for the same pattern; reusing the same shape (or in-process mutex around the read+append pair) would close it.

2. **`ZeroHookConfigStore.list` user-bypass hack.** `new ZeroHookConfigStore({ configPath }).list()` (`src/zero-hooks/config.ts:154-159`) passes `${this.configPath}.user-missing` to force the user layer to be absent, rather than reading a single file. It works but is opaque; a `loadZeroHooksConfigFromPath(path)` (or a `source: 'project' | 'user'` filter on the existing loader) would read more directly.

3. **CLI doesn't surface the audit store yet.** `ZeroHookAuditStore.readEvents()` exists and is tested, but `zero hooks list` (`src/index.ts:431-449`) only shows config. A follow-up `hooks list --audit` (or `hooks audit`) subcommand would make the audit file discoverable from the CLI without users hunting for `~/.local/share/zero/hooks/audit.jsonl`.

4. **Matcher support is `*`-only.** `matchesHookMatcher` (`src/zero-hooks/config.ts:248-255`) handles exact match and `*` wildcards only — no `?` / `[abc]`. Worth a one-line doc comment on `ZeroHookDefinition.matcher` so users know what to expect.

No blockers. The race in #1 is the only one I'd want to see fixed before this lands in main.

gnanam1990:
@coderabbit review

Vasanthdev2004:
Blockers
- `zero hooks list` text output leaks secret-looking hook arguments from `.zero/hooks.json`. The JSON path is wrapped in `redactZeroSecrets`, but the text path calls `formatZeroHookList(result.config, result.diagnostics)` directly and prints `[hook.command, ...hook.args].join(' ')`. I verified this with a temp project hook whose args contained an assembled OpenAI-shaped key: `hooks list --json` removed the raw value and included `[REDACTED]`, but plain `hooks list` printed the raw value. Hook config is a diagnostic surface and hook commands/args are likely to contain API keys, bearer tokens, webhook URLs, or other secret-bearing values, so the text formatter needs to redact before printing and should have a regression test for both JSON and text outputs.

Non-Blocking
- The hook execution/audit backend is still config/list/audit storage only; I did not find execution wiring into the agent loop in this PR. That is fine for this scope, but future execution wiring should keep hooks fail-closed behind explicit permission/safety boundaries.
- `ZeroHookAuditStore.append()` computes sequence by reading the whole audit file and then appending, so concurrent in-process appends can race and duplicate/out-of-order sequence values. This is probably acceptable for a first backend layer if execution is not wired yet, but it should be serialized before hooks can run from multiple concurrent calls.
- `matcher` is accepted on `sessionStart`/`sessionEnd` hooks even though `selectZeroHooks()` requires `toolName` when a matcher exists, so those session hooks become unselectable. Either reject matcher for session events with a schema diagnostic or define a clear matching contract for session events.

Looks Good
- Layered user/project config loading keeps valid layers alive when another layer is malformed and emits structured diagnostics.
- Project hooks override user hooks deterministically with a duplicate diagnostic and stable id sorting.
- Hook ids/events/commands are schema-validated before they enter the normalized config.
- `hooks list --json` already routes through `redactZeroSecrets` and did redact the same probe secret correctly.

Validation
- `bun run typecheck` — pass.
- `bun test tests/zero-hooks.test.ts --timeout 15000` — 7 pass / 0 fail.
- `bun test ./tests --timeout 15000` — 274 pass / 0 fail.
- `bun run build` — pass.
- `bun run smoke:build` — pass.
- `git diff --check` — pass.
- CLI probes: `zero hooks --help`, `zero hooks list --help`, and temp-project redaction probe for both text and JSON hook-list outputs. The redaction probe is the blocker above.

Vasanthdev2004 (changes requested):
Requesting changes because plain-text `zero hooks list` leaks raw secret-bearing hook args while the JSON path redacts them. Full details and validation are in my review comment.

Vasanthdev2004:
Blockers
None.

Non-Blocking
- This PR is still hook config/list/audit storage only; I do not see runtime hook execution wiring in the agent loop yet. Future execution wiring should keep hooks permissioned/fail-closed before any command runs.
- `ZeroHookConfigStore` and `ZeroHookAuditStore` now serialize in-process mutations/appends. That covers the current single-process CLI/backend use case, but multi-process writers could still race because there is no filesystem lock. I would keep that as a follow-up before exposing hooks as a frequently-running runtime surface.

Looks Good
- The previous blocker is fixed at head `3c61281`: both plain text `zero hooks list` and `zero hooks list --json` now redact hook args containing an assembled OpenAI-shaped key. I verified the exact raw value is absent from both outputs and `[REDACTED]` appears in both.
- Session hook matcher ambiguity is fixed: `sessionStart`/`sessionEnd` hooks with `matcher` now fail schema validation with `fieldPath: hooks.0.matcher`, and the invalid hook is not loaded.
- Audit reads now tolerate malformed JSONL lines without blocking later appends, and append sequencing is serialized in-process.
- Config store mutations are serialized in-process for the same config path, avoiding the earlier same-process read/modify/write clobber.
- Wildcard matcher logic was changed to a linear matcher rather than building a RegExp from hook config.

Validation
- `bun run typecheck` — pass.
- `bun test tests/zero-hooks.test.ts --timeout 15000` — 10 pass / 0 fail.
- `bun test ./tests --timeout 15000` — 277 pass / 0 fail.
- `bun run build` — pass.
- `bun run smoke:build` — pass.
- `git diff --check` — pass.
- CLI redaction probe from a temp project with isolated HOME/USERPROFILE/XDG dirs — pass for text and JSON outputs.
- Session matcher schema probe from a temp project — pass.
- GitHub checks at review time: Ubuntu smoke, macOS smoke, Windows smoke, Performance Smoke, Zero Review, and CodeRabbit all success.

Vasanthdev2004 (approved):
Approved current head 3c61281 after verifying the hook-list redaction blocker and session-matcher/schema hardening are fixed, with local validation passing.

PR #47, opened 2026-06-04
[codex] add Go read-only core tools

Vasanthdev2004:
## Summary

- Add the first Go `internal/tools` package for Zero.
- Define tool metadata, JSON-schema-compatible parameter shapes, safety metadata, results, and registry safe-run behavior.
- Add read-only core tools: `read_file`, `list_directory`, `glob`, and `grep`.
- Keep tools workspace-scoped and read-only for this first Vasanth M0 slice.
- Add Go tests for metadata/schema, registry behavior, workspace rejection, line ranges, directory listing, glob matching, and grep output modes.

## Why

This starts Vasanth's Go M0 tools/TUI path without jumping into Bubble Tea or mutation tools too early. The TUI and agent loop can now consume a clean Go tool contract and safe read/search tools.

## Validation

Local:
- `git diff --cached --check` passed before commit.
- `git diff --check` passed.
- `bun run typecheck` passed.
- `bun test ./tests --timeout 15000` passed: 277 tests.
- `bun run build` passed.
- `bun run smoke:build` passed.

Blocked locally:
- `go test ./...` and `go build ./cmd/zero` could not run because `go` is not installed on this Windows machine.

CI:
- Existing CI runs `go test ./...` and `go build ./cmd/zero` on Ubuntu, macOS, and Windows.

## Summary by CodeRabbit

* **New Features**
  * Added file tools: read file, list directory, glob matching, and grep-style search, plus a registry exposing read-only tools and safety metadata.

* **Tests**
  * Added comprehensive unit tests covering file operations, glob/grep behaviors, directory listing, and registry safety paths.

* **Chores**
  * Introduced core tooling types, workspace path handling, and argument validation utilities to support the new tools.

### DIFF
--- a/internal/tools/args.go
+++ b/internal/tools/args.go
@@ -1,0 +1,71 @@
+package tools
+
+import (
+	"fmt"
+	"math"
+)
+
+func stringArg(args map[string]any, key string, fallback string, required bool) (string, error) {
+	value, ok := args[key]
+	if !ok || value == nil {
+		if required {
+			return "", fmt.Errorf("%s is required", key)
+		}
+		return fallback, nil
+	}
+
+	text, ok := value.(string)
+	if !ok || text == "" {
+		return "", fmt.Errorf("%s must be a non-empty string", key)
+	}
+	return text, nil
+}
+
+func boolArg(args map[string]any, key string, fallback bool) (bool, error) {
+	value, ok := args[key]
+	if !ok || value == nil {
+		return fallback, nil
+	}
+
+	boolean, ok := value.(bool)
+	if !ok {
+		return false, fmt.Errorf("%s must be a boolean", key)
+	}
+	return boolean, nil
+}
+
+func intArg(args map[string]any, key string, fallback int, min int, max int) (int, error) {
+	value, ok := args[key]
+	if !ok || value == nil {
+		return fallback, nil
+	}
+
+	var number int
+	switch typed := value.(type) {
+	case int:
+		number = typed
+	case int32:
+		number = int(typed)
+	case int64:
+		number = int(typed)
+	case float64:
+		if math.Trunc(typed) != typed {
+			return 0, fmt.Errorf("%s must be an integer", key)
+		}
+		number = int(typed)
+	default:
+		return 0, fmt.Errorf("%s must be an integer", key)
+	}
+
+	if number < min {
+		return 0, fmt.Errorf("%s must be at least %d", key, min)
+	}
+	if max > 0 && number > max {
+		return 0, fmt.Errorf("%s must be at most %d", key, max)
+	}
+	return number, nil
+}
+
+func intPtr(value int) *int {
+	return &value
+}
--- a/internal/tools/file_tools_test.go
+++ b/internal/tools/file_tools_test.go
@@ -1,0 +1,179 @@
+package tools
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestReadFileToolReadsLineRanges(t *testing.T) {
+	root := t.TempDir()
+	writeTestFile(t, filepath.Join(root, "notes.txt"), "alpha\nbeta\ngamma\ndelta")
+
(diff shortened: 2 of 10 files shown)


anandh8x (approved):
## What's good

- Clean separation in `internal/tools/types.go`: `Tool` interface with `baseTool` covering the four boilerplate methods, JSON-schema-compatible `Schema` / `PropertySchema`, and the `SideEffect` / `Permission` enums. `readOnlySafety` is a tidy factory.
- Workspace boundary is enforced correctly: `resolveWorkspacePath` (`internal/tools/workspace.go:32-69`) runs `filepath.EvalSymlinks` on both the root and the target before the `..` / abs-path check, so symlinked parent segments can't smuggle a path out of the workspace. This was one of CodeRabbit's earlier findings and is now done.
- All four tools share a single `ignoredDirectories` set (`.git`, `node_modules`, `dist`, `build`, `.next`, `.turbo`, `coverage`, `.cache`, `tmp`, `temp`) via `shouldSkipDirectory` — used consistently by `glob.scanGlob`, `grep.grepFiles`, and `listDirectoryEntries`.
- `WalkDir` callbacks now propagate `walkErr` (both `internal/tools/glob.go:96-98` and `internal/tools/grep.go:171-173`), so filesystem traversal failures reach the caller instead of being silently dropped.
- `read_file.go:64-69` trims a single trailing `\n` before splitting, so a file ending in `\n` reports the correct line count. Also normalises CRLF → LF and treats `start_line > total` as a soft "past the end" result rather than an error.
- `Registry.Run` (`internal/tools/registry.go:30-41`) gates on `tool.Safety().Permission == PermissionAllow` before delegating — gives a single safe-run choke point for future tools.
- Test coverage: metadata + schema invariants, workspace rejection, line ranges, recursive listing with junk-dir skipping, glob `**` + limit, glob include-dirs, grep content / `files_with_matches` / `count` modes, and the glob test now uses `regexp.MustCompile` to count `.go` matches instead of `strings.Count(".go")` (also a prior CodeRabbit concern, now addressed).
- Follow-up commit `0c63a34 fix: harden go read-only tools` addresses each of CodeRabbit's actionable items; CI is green on all four matrices (ubuntu, macos, windows, Performance Smoke) and CodeRabbit reports SUCCESS on the latest SHA.

1. **`read_file` on a missing path leaks the absolute path via EvalSymlinks.** `resolveWorkspacePath` runs `filepath.EvalSymlinks(target)` before `os.ReadFile`, so a missing file surfaces as `Error reading file <requested>: lstat /workspace/<requested>: no such file or directory` (`internal/tools/read_file.go:54-58`). The message is technically informative but the EvalSymlinks failure mode is opaque to a non-engineer caller — catching `errors.Is(err, fs.ErrNotExist)` (or a `ENOENT` check) and returning a dedicated "file not found" result would tidy this up.

2. **`listDirectoryEntries` swallows subdirectory errors silently.** When the recursive `os.ReadDir` on a child fails, the child is dropped from the output with no diagnostic (`internal/tools/list_directory.go:88-92`). Graceful degradation is fine for a read tool, but `list_directory --recursive` will silently show a directory as empty when a permissions error blocks its children. A surface as a trailing `! permission denied: <dir>` line, or at least a `Meta` field, would help.

3. **`head_limit` has no upper bound in grep.** `intArg(args, "head_limit", 50, 1, 0)` passes `max=0` (meaning "no max" in the helper's convention), so a caller can ask for a million-line result. Glob caps at 1000; grep is unbounded. Picking a sensible cap (1000?) and clamping would mirror the glob behaviour.

4. **`shouldSkipDirectory` is not unit-tested directly.** It's exercised transitively through every other tool's tests, but a `TestIgnoredDirectories_AreSkipped` asserting the exact set and behaviour would document the policy and catch accidental additions.

5. **The `"-i"` schema key is unusual.** `internal/tools/grep.go:37` registers the case-insensitive flag under the property name `"-i"`. Standard for CLI parity, but the tool interface's JSON-schema consumer (TUI / LLM-side) might prefer a separate `case_insensitive` (already present alongside it) and treat `-i` as an alias. Worth a doc comment in the `Schema` description.

No blockers. Ship it.

PR #48, opened 2026-06-04
Add Go M0 write tools

Vasanthdev2004:
## Summary
- add Go core write tools: write_file, edit_file, apply_patch
- add in-memory update_plan tool and CoreTools/CoreWriteTools registry helpers
- add prompt permission gating through Registry.RunWithOptions
- add workspace target-path resolution for files that may not exist yet

## Tests
- bun run typecheck
- bun run build
- bun run smoke:build

## Local blockers
- go/go fmt are not installed on this Windows machine, so go test ./..., go build ./cmd/zero, and gofmt could not run locally
- bun run smoke:go fails locally because build:go cannot run; CI is configured with actions/setup-go and will run Go test/build/smoke

## Summary by CodeRabbit

* **New Features**
  * Added workspace file tools: create/overwrite files, in-place exact-string edits (first/all), and apply unified-diff patches.
  * Added an in-memory task planning tool to store, format, clear, and inspect plans.

* **Security & Reliability**
  * Stronger workspace path checks and symlink-aware resolution; safer overwrite, permissions gating, and match validations.

* **Tests**
  * Expanded tests covering tools, permission gating, edge cases, and filesystem safety.

### DIFF
--- a/internal/tools/apply_patch.go
+++ b/internal/tools/apply_patch.go
@@ -57,12 +57,18 @@ func (tool applyPatchTool) Run(ctx context.Context, args map[string]any) Result 
 		return errorResult("Error applying patch: " + err.Error())
 	}
 	patchPath := tempFile.Name()
-	defer os.Remove(patchPath)
+	defer func() {
+		_ = os.Remove(patchPath)
+	}()
 	if _, err := tempFile.WriteString(patch); err != nil {
-		tempFile.Close()
+		_ = tempFile.Close()
 		return errorResult("Error applying patch: " + err.Error())
 	}
 	if err := tempFile.Close(); err != nil {
+		return errorResult("Error applying patch: " + err.Error())
+	}
+
+	if err := recheckPatchWriteTargets(applyRoot, patch); err != nil {
 		return errorResult("Error applying patch: " + err.Error())
 	}
 
@@ -93,6 +99,20 @@ func validatePatchPaths(root string, patch string) error {
 				return fmt.Errorf("patch path %q must stay inside the workspace", path)
 			}
 			if _, _, err := resolveWorkspaceTargetPath(root, path); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func recheckPatchWriteTargets(root string, patch string) error {
+	for _, line := range strings.Split(strings.ReplaceAll(patch, "\r\n", "\n"), "\n") {
+		for _, path := range patchPathsFromLine(line) {
+			if path == "" || path == "/dev/null" {
+				continue
+			}
+			if err := recheckWorkspaceWriteTarget(root, path); err != nil {
 				return err
 			}
 		}
--- a/internal/tools/apply_patch.go
+++ b/internal/tools/apply_patch.go
@@ -1,0 +1,124 @@
+package tools
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+type applyPatchTool struct {
+	baseTool
+	workspaceRoot string
+}
+
+func NewApplyPatchTool(workspaceRoot string) Tool {
+	return applyPatchTool{
+		baseTool: baseTool{
+			name:        "apply_patch",
+			description: "Apply a unified diff patch inside the workspace.",
+			parameters: Schema{
+				Type: "object",
+				Properties: map[string]PropertySchema{
+					"patch": {Type: "string", Description: "Unified diff patch to apply."},
+					"cwd":   {Type: "string", Description: "Directory where the patch should be applied. Defaults to workspace root.", Default: "."},
+				},
+				Required:             []string{"patch"},
+				AdditionalProperties: false,
+			},
+			safety: promptSafety(SideEffectWrite, "Applies patch hunks that can create, edit, or delete files."),
+		},
+		workspaceRoot: normalizeWorkspaceRoot(workspaceRoot),
+	}
+}
+
+func (tool applyPatchTool) Run(ctx context.Context, args map[string]any) Result {
+	patch, err := stringArg(args, "patch", "", true)
+	if err != nil {
+		return errorResult("Error: Invalid arguments for apply_patch: " + err.Error())
+	}
+	cwd, err := stringArg(args, "cwd", ".", false)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


anandh8x (approved):
## What's good

- Defense-in-depth on writes: every write tool resolves via `resolveWorkspaceTargetPath` (`internal/tools/workspace.go:72-126`) for non-existent targets, and then runs `recheckWorkspaceWriteTarget` (`internal/tools/workspace.go:128-191`) immediately before the actual write. The recheck walks the path segment by segment and refuses any segment whose `Lstat` shows `os.ModeSymlink`, so a symlink parent is rejected at the latest possible moment before `os.WriteFile` runs. This is the TOCTOU mitigation CodeRabbit pushed for, and it's now in place across `write_file`, `edit_file`, and `apply_patch`.
- `apply_patch` does the recheck **twice** — once during `validatePatchPaths` and again right before `exec.CommandContext("git", "apply", ...)` — so the `git apply` subprocess only ever sees a path tree that has just been re-validated (`internal/tools/apply_patch.go:69-72`). The temp patch file is `defer`-cleaned (`internal/tools/apply_patch.go:55-56`) and both the `WriteString` and `Close` errors are checked.
- `Registry.RunWithOptions` (`internal/tools/registry.go:36-54`) is a clean three-way switch: `PermissionAllow` runs unconditionally, `PermissionPrompt` requires `options.PermissionGranted`, anything else (incl. `PermissionDeny`) is refused. Defaulting `Run` to `RunWithOptions(ctx, name, args, RunOptions{})` keeps existing call sites safe by default while letting the TUI / agent loop pass a grant flag.
- `edit_file` enforces uniqueness by default and only allows non-unique edits behind an explicit `replace_all: true`, with a clear error message that quotes the occurrence count. Empty `new_string` is allowed and works as a real deletion (tested). `write_file` defaults to refusing overwrites and requires `overwrite: true` to replace.
- `stringArgWithEmpty` (`internal/tools/args.go:9-23`) now separates the type check from the emptiness check, so a caller passing `content: 42` gets `content must be a string` instead of the misleading `content must be a non-empty string` — CodeRabbit's earlier finding, fixed.
- `update_plan` uses a pointer receiver, returns a defensive copy from `CurrentPlan()` (`internal/tools/update_plan.go:63-65`), and the test suite verifies that mutating the returned slice doesn't mutate internal state. The follow-up commit `db22668` adds `TestUpdatePlanToolClearPlanResetsState` for the reset semantics CodeRabbit flagged.
- Test coverage: prompt blocking without grant, prompt running with grant, create/overwrite/refuse, empty content, type errors, workspace rejection, symlink-parent rejection, edit uniqueness, `replace_all`, full `git apply` round-trip, and symlink rejection inside patch paths. CI is green on all four matrices and CodeRabbit reports SUCCESS on the latest SHA.

1. **TOCTOU between the recheck and the actual write.** `write_file.go:64-67` and `edit_file.go:78-83` do `recheckWorkspaceWriteTarget` → `os.MkdirAll` / `os.WriteFile` non-atomically. The recheck catches an **existing** symlink, but if `<root>/parent` doesn't exist yet, an attacker (or another process) who creates it as a symlink to e.g. `/etc` between the recheck and the `MkdirAll` call would have the write land outside the workspace. The canonical fix is the same write-temp-then-`rename` pattern PR #40's `ZeroMcpPermissionStore` uses: write to `<dir>/.zero.tmp.<rand>`, `fsync`, then `rename` onto the target. On POSIX this is atomic; on Windows the rename fails if the destination exists (which is the desired behaviour). Adopting it for `write_file` would close this window for free; the same pattern would let `edit_file` build the new contents in a temp file and rename over the original, which would also give you a clean "the file changed under us" detection.

2. **`apply_patch` parses patch paths line-by-line, not via a real parser.** `patchPathsFromLine` (`internal/tools/apply_patch.go:121-135`) recognises `diff --git`, `--- `, and `+++ ` prefixes and strips `a/`/`b/`. A unified diff with quoted filenames containing spaces (e.g. `+++ "b/path with spaces.txt"`) or with backslash-escaped paths won't be parsed correctly and the workspace check would either miss paths or fail to recognise that a path is unsafe. A test using a `git diff` output on a path with spaces would cover this; if you want to keep the manual parser, escaping-aware splitting on the diff header would be the next step. (Note: this isn't a vulnerability per se — `git apply` itself enforces the workspace boundary via `cwd` — but the pre-flight check could be bypassed, which is more of a correctness issue for the diagnostic.)

3. **Hard dependency on `git apply` is not documented.** `apply_patch` shells out to `git apply --whitespace=nowarn`. If the runtime doesn't have git on PATH (or the user disables it), the tool errors with `Error applying patch: exec: "git": executable file not found in $PATH` — informative but unexplained. A line in the tool description ("Requires `git` on PATH.") would set expectations; an explicit `exec.LookPath("git")` check in `NewApplyPatchTool` (or at the top of `Run`) with a friendlier error would be even better.

4. **`update_plan` allows duplicate IDs across plan items.** `parsePlanItems` (`internal/tools/update_plan.go:60-93`) doesn't check for `id` uniqueness, so two items with `id: "1"` are accepted and rendered. If the TUI or downstream consumers key on `id`, this would silently break them. A uniqueness assertion (or a doc comment saying IDs may repeat) would prevent surprises.

5. **`Run` vs `RunWithOptions` asymmetry is a footgun.** The default `Run` never grants permissions, so any `PermissionPrompt` tool fails by default; the caller has to know to switch to `RunWithOptions` with `PermissionGranted: true`. This is the right safe default, but it's easy to write a TUI that calls `Run` for everything and then is surprised when write tools never work. A short note in the `Run` doc comment (or a lint-style test that asserts every `Run` call site is paired with a permission check upstream) would help.

6. **`recheckPatchWriteTargets` doesn't validate `git apply`'s actual output paths against the workspace.** The pre-checks look at the `--- a/x` and `+++ b/x` headers, but `git apply` can create new files, modify existing ones, and delete files. A patch that includes `rename from` / `rename to` or `copy from` / `copy to` lines isn't enumerated by `patchPathsFromLine` and therefore won't be rechecked. Same caveat as #2 — `git apply` itself respects `command.Dir = applyRoot`, so the actual writes land inside the workspace — but the pre-flight check is incomplete for non-trivial patches.

No blockers. The TOCTOU window in #1 is the one I'd most want to see closed in a follow-up; the rest are correctness / documentation polish.

PR #49, opened 2026-06-04
Add Go M0 bash tool

Vasanthdev2004:
## Summary
- add a Go ash tool with shell side-effect metadata and prompt permission gating
- run commands from a workspace-scoped cwd with configurable timeout
- return stdout/stderr plus exit code, cwd, and timeout metadata
- register bash through CoreTools via a shell tool group

## Tests
- added Go tests for registration, permission gating, cwd handling, outside-workspace rejection, nonzero exits, and timeout behavior
- git diff --check
- un run typecheck
- un test ./tests --timeout 15000
- un run build
- un run smoke:build

## Local blockers
- go and gofmt are not installed on this Windows machine, so go test ./..., go build ./cmd/zero, and gofmt could not run locally
- CI has ctions/setup-go and will verify Go test/build/smoke on Linux, macOS, and Windows

## Summary by CodeRabbit

* **New Features**
  * Added a bash shell execution tool to run shell commands inside a workspace with timeout support, OS-aware shell invocation, workspace-relative cwd enforcement, and structured results (stdout/stderr, exit code, timeout metadata).

* **Tests**
  * Added tests covering command execution, workspace boundary enforcement, permission/registry behavior, nonzero exit handling, and timeout scenarios.

### DIFF
--- a/internal/tools/bash.go
+++ b/internal/tools/bash.go
@@ -1,0 +1,153 @@
+package tools
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const defaultBashTimeoutMS = 120000
+const maxBashTimeoutMS = 600000
+
+type bashTool struct {
+	baseTool
+	workspaceRoot string
+}
+
+func NewBashTool(workspaceRoot string) Tool {
+	return bashTool{
+		baseTool: baseTool{
+			name:        "bash",
+			description: "Execute a shell command inside the workspace after permission is granted.",
+			parameters: Schema{
+				Type: "object",
+				Properties: map[string]PropertySchema{
+					"command":    {Type: "string", Description: "Shell command to execute."},
+					"cwd":        {Type: "string", Description: "Workspace directory to run the command in. Defaults to workspace root.", Default: "."},
+					"timeout_ms": {Type: "integer", Description: "Command timeout in milliseconds.", Default: defaultBashTimeoutMS, Minimum: intPtr(1), Maximum: intPtr(maxBashTimeoutMS)},
+				},
+				Required:             []string{"command"},
+				AdditionalProperties: false,
+			},
+			safety: promptSafety(SideEffectShell, "Shell commands can read, write, or execute programs."),
+		},
+		workspaceRoot: normalizeWorkspaceRoot(workspaceRoot),
+	}
+}
+
+func (tool bashTool) Run(ctx context.Context, args map[string]any) Result {
+	commandText, err := stringArg(args, "command", "", true)
+	if err != nil {
+		return errorResult("Error: Invalid arguments for bash: " + err.Error())
+	}
+	cwd, err := stringArg(args, "cwd", ".", false)
+	if err != nil {
+		return errorResult("Error: Invalid arguments for bash: " + err.Error())
+	}
+	timeoutMS, err := intArg(args, "timeout_ms", defaultBashTimeoutMS, 1, maxBashTimeoutMS)
+	if err != nil {
+		return errorResult("Error: Invalid arguments for bash: " + err.Error())
+	}
+
+	absoluteCwd, relativeCwd, err := resolveWorkspacePath(tool.workspaceRoot, cwd)
+	if err != nil {
+		return errorResult("Error running bash: " + err.Error())
+	}
+
+	commandCtx, cancel := context.WithTimeout(ctx, time.Duration(timeoutMS)*time.Millisecond)
+	defer cancel()
+
+	command := exec.CommandContext(commandCtx, shellExecutable(), shellArguments(commandText)...)
+	command.Dir = absoluteCwd
+
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	command.Stdout = &stdout
+	command.Stderr = &stderr
+
+	err = command.Run()
+	exitCode := commandExitCode(err)
+	meta := map[string]string{
+		"cwd":        relativeCwd,
+		"exit_code":  strconv.Itoa(exitCode),
+		"timeout_ms": strconv.Itoa(timeoutMS),
+	}
+
+	if errors.Is(commandCtx.Err(), context.DeadlineExceeded) {
+		return Result{
+			Status: StatusError,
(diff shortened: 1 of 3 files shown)


anandh8x (approved):
## What's good

- Workspace boundary is enforced through `resolveWorkspacePath` (`internal/tools/bash.go:54-56`), so the same `EvalSymlinks` + `..` rejection that protects the read/write tools applies to the shell's `cwd`. `command.Dir = absoluteCwd` therefore always points at a real, symlink-resolved directory inside the workspace.
- Timeout is bounded both at the schema layer (`Minimum: 1, Maximum: 600000`) and at the runtime layer (`context.WithTimeout` + `exec.CommandContext`, `internal/tools/bash.go:58-60, 62-63`). When the deadline fires, `errors.Is(commandCtx.Err(), context.DeadlineExceeded)` is checked **before** the generic `err != nil` branch, so timeout reports don't get misclassified as a regular failure.
- OS-aware shell invocation: `shellExecutable` / `shellArguments` (`internal/tools/bash.go:107-116`) return `cmd.exe /d /s /c <command>` on Windows and `/bin/sh -c <command>` elsewhere. The `/d /s /c` triple is the documented safe form for one-shot command execution on Windows; the `set -e` / `set -u` defaults from a login shell don't leak in.
- Result metadata carries `cwd`, `exit_code`, and `timeout_ms` (`internal/tools/bash.go:69-72`), and `formatBashOutput` (`internal/tools/bash.go:135-150`) only emits `exit_code: N` for non-zero exits — keeps happy-path output clean. Stdout/stderr are trimmed of trailing newlines and only rendered if non-empty.
- Registry wiring: `CoreShellTools` exposes just `bash`, and `CoreTools` appends it after read-only and write tools. Permission is `SideEffectShell` / `PermissionPrompt`, matching `apply_patch` / `write_file` / `edit_file` — consistent safety tier.
- Test harness is self-contained: `TestMain` (`internal/tools/bash_tool_test.go:13-22`) detects `--zero-bash-helper <name>` and dispatches to a tiny in-process command library (`success`, `stderr`, `fail`, `pwd`, `sleep`). Avoids the cross-platform headache of depending on `echo` / `cat` / `sleep` having the same flags everywhere, and `shellQuote` handles the POSIX quoting. CI green on all four matrices; CodeRabbit reports SUCCESS.

1. **Unbounded stdout/stderr buffers.** `var stdout bytes.Buffer; var stderr bytes.Buffer` (`internal/tools/bash.go:65-67`) have no size cap, and the only `Run` completion path that returns early is the timeout. A `cat /dev/urandom | head -c 10G` style command (or any noisy build) can OOM the agent process before the timeout fires, because `command.Run` only returns after the child closes its stdout/stderr — which a 10-minute-long noisemaker could legitimately do. Wrapping the writers in `io.LimitReader`-style guards (e.g. `io.Copy(&limitedBuffer{limit: 64 << 20}, command.Stdout)`) or checking the buffer size in a `WaitDelay` callback would cap the worst case. Worth adding before this lands in front of users with large repos.

2. **Stdin is inherited from the parent process.** `command.Stdin` is not set (`internal/tools/bash.go:62-67`). When the agent is run from a TUI / interactive terminal, the child shell will share that stdin file descriptor with the UI. A bash command that does `read` or `cat` would consume whatever the TUI was about to read, and any input the TUI was reading could be reinterpreted as command input. Setting `command.Stdin = os.Stdin` explicitly is also wrong here; the safe choice is either to leave it nil **and document that the tool assumes a non-interactive parent** or to point it at `/dev/null` (or `os.DevNull`) so the child is guaranteed EOF.

3. **Timeout exit code is `-1` in metadata.** When the deadline fires, `command.Run()` returns the kill error rather than `*exec.ExitError`, so `commandExitCode` falls through to `-1` (`internal/tools/bash.go:118-126`). The output string is correct (`timed out after Nms`), but `meta.exit_code = -1` is opaque. A dedicated `meta["timed_out"] = "true"` (and an explicit `meta["exit_code"] = "-1"` only if you want to keep the key stable) would let downstream consumers distinguish "command exited 7" from "killed by deadline" without string-matching the output.

4. **Env vars are inherited as-is.** The child sees the full parent env, including anything the agent stashed in `GITHUB_TOKEN`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, etc. This is the standard behaviour for `exec.Cmd` and probably what users want, but it's worth a one-line note in the tool description (`Description`) so the LLM side doesn't accidentally leak secrets into a sub-process that the user can see in `ps`. Pairing with a `--inherit-env=false` opt-out (or a `env: {}` arg) is a reasonable follow-up.

5. **Grant-integration test is missing.** CodeRabbit's only nitpick on this PR is a `TestBashToolAllowsExecutionWhenRegistryGrantPresent` style test that goes through `Registry.RunWithOptions` with `PermissionGranted: true` and asserts the command actually runs. `TestBashToolRunsCommandInWorkspace` exercises the tool directly, which proves the tool works but doesn't prove the registry's prompt path is wired up for this specific tool. The test for `write_file` (`TestRegistryRunsPromptToolsWithGrant` in #48) is a good template; copy that for bash and the gating contract is covered end-to-end.

6. **Workspace cwd uses the read-side resolver.** `resolveWorkspacePath` requires the path to exist (`EvalSymlinks` will fail on `ENOENT`). For `bash` that's the right choice — a non-existent cwd would error at exec time anyway — but it means `cwd: "./new-dir"` errors with the `EvalSymlinks` `lstat` message rather than the friendlier "directory does not exist" text. `errors.Is(err, fs.ErrNotExist)` on the resolver error path would let the tool produce a more user-friendly message. Same pattern as the read_file observation in #47.

7. **Helper binary path is platform-fragile.** `helperCommand` uses `os.Args[0]` to relaunch the test binary (`internal/tools/bash_tool_test.go:181-184`). On Linux this is fine; on Windows the test relies on the test runner using the actual `.exe` rather than a wrapper — and on cross-compiled runs the path could include forward slashes that `cmd.exe /c` doesn't love. `shellQuote` already special-cases Windows to skip quoting, which is the right call for now, but a `runtime.GOOS == "windows" && !strings.HasSuffix(executable, ".exe")` guard might be worth adding if any of the CI matrices starts being flaky.

No blockers. The output-size cap in #1 and the stdin handling in #2 are the two I'd most want to see tightened in a follow-up; the rest are correctness / documentation polish.

PR #51, opened 2026-06-04
Add Go M0 agent runtime

Vasanthdev2004:
## Summary

Adds the first Go agent runtime spine for M0:

- Defines Go provider messages, stream events, tool calls, tool definitions, usage, and runtime options.
- Adds a bounded ReAct-style loop that streams text, collects tool calls, executes tools through `tools.Registry.RunWithOptions`, feeds tool results back into provider messages, and stops at `MaxTurns`.
- Supports permission modes for safe reads, prompt-gated denial by default, and unsafe grants.
- Wires a minimal `zero exec <prompt>` Go CLI command through the new runtime with a deterministic offline provider until the real Go provider lands in the next PR.

## Tests

- `git diff --check`
- `go test ./...`
- `bun run build:go`
- `bun run smoke:go`
- `bun run typecheck`
- `bun run build`
- `bun run smoke:build`

## Notes

The Go `zero exec` provider is intentionally offline in this PR. It proves the CLI -> agent runtime path without adding the OpenAI-compatible provider in the same review. The next PR should replace that provider stub with real provider/config wiring.

## Summary by CodeRabbit

* **New Features**
  * Interactive agent with streaming responses that supports multi-turn tool calls, emits incremental text and usage updates, and enforces configurable permission modes for tool execution.
  * Tool discovery/advertising with ordered tool-call handling.
  * New CLI command "zero exec" to run a prompt and print the agent's final answer.

* **Chores / Tests**
  * Added comprehensive unit tests covering streaming behavior, tool execution, permission handling, CLI execution, and error scenarios.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1,0 +1,234 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"sort"
+
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
+const defaultSystemPrompt = "You are Zero, a terminal coding agent. Help with the current workspace and use tools when needed."
+const maxTurnsAnswer = "Agent reached maximum number of turns without a final answer."
+
+type pendingToolCall struct {
+	id        string
+	name      string
+	arguments string
+}
+
+func Run(ctx context.Context, prompt string, provider Provider, options Options) (Result, error) {
+	if provider == nil {
+		return Result{}, errors.New("agent provider is required")
+	}
+
+	maxTurns := options.MaxTurns
+	if maxTurns <= 0 {
+		maxTurns = 12
+	}
+
+	registry := options.Registry
+	if registry == nil {
+		registry = tools.NewRegistry()
+	}
+
+	permissionMode := options.PermissionMode
+	if permissionMode == "" {
+		permissionMode = PermissionModeAuto
+	}
+
+	messages := []Message{
+		{Role: RoleSystem, Content: defaultSystemPrompt},
+		{Role: RoleUser, Content: prompt},
+	}
+
+	result := Result{Messages: copyMessages(messages)}
+	for turn := 0; turn < maxTurns; turn++ {
+		result.Turns = turn + 1
+		request := CompletionRequest{
+			Messages: copyMessages(messages),
+			Tools:    toolDefinitions(registry, permissionMode),
+		}
+
+		stream, err := provider.StreamCompletion(ctx, request)
+		if err != nil {
+			result.Messages = copyMessages(messages)
+			return result, err
+		}
+
+		currentText, toolCalls, err := collectTurn(ctx, stream, options.OnText, options.OnUsage)
+		if err != nil {
+			result.Messages = copyMessages(messages)
+			return result, err
+		}
+
+		messages = append(messages, Message{
+			Role:      RoleAssistant,
+			Content:   currentText,
+			ToolCalls: toolCalls,
+		})
+
+		if len(toolCalls) == 0 {
+			result.FinalAnswer = currentText
+			result.Messages = copyMessages(messages)
+			return result, nil
+		}
+
+		for _, call := range toolCalls {
+			if options.OnToolCall != nil {
+				options.OnToolCall(call)
+			}
+			toolResult := executeToolCall(ctx, registry, call, permissionMode)
+			if options.OnToolResult != nil {
+				options.OnToolResult(toolResult)
+			}
+			messages = append(messages, Message{
+				Role:       RoleTool,
+				Content:    toolResult.Output,
+				ToolCallID: toolResult.ToolCallID,
+			})
+		}
+	}
+
+	result.FinalAnswer = maxTurnsAnswer
+	result.Messages = copyMessages(messages)
+	return result, nil
+}
+
+func collectTurn(ctx context.Context, stream <-chan StreamEvent, onText func(string), onUsage func(Usage)) (string, []ToolCall, error) {
+	currentText := ""
+	pending := make(map[string]*pendingToolCall)
+	order := make([]string, 0)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return "", nil, ctx.Err()
+		case event, ok := <-stream:
+			if !ok {
+				return currentText, buildToolCalls(order, pending), nil
+			}
+
+			switch event.Type {
+			case EventText:
+				currentText += event.Content
+				if onText != nil {
+					onText(event.Content)
+				}
+			case EventToolCallStart:
+				call := pending[event.ToolCallID]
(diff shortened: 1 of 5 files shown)


Vasanthdev2004:
@coderabbitai review

anandh8x (approved):
## What's good

- Clean ReAct spine: `Run` (`internal/agent/loop.go:23-99`) drives a `maxTurns`-bounded loop, copies the message log defensively on every `CompletionRequest` (`copyMessages`, `internal/agent/loop.go:214-224`), and returns a `Result` with `FinalAnswer`, `Turns`, and the full message log. Provider errors and max-turns termination both return the partial message log so a TUI can render what happened.
- `collectTurn` (`internal/agent/loop.go:103-145`) handles every event type correctly: `EventText` accumulates and fans out via `OnText`; `EventToolCallStart` / `EventToolCallDelta` register tool calls in arrival order via the `order []string` + `pending map[string]*pendingToolCall` pair, and a missing Start doesn't break Delta accumulation. `EventToolCallEnd` is consumed but intentionally a no-op (the loop terminates on `EventDone` or `ctx.Done`).
- Permission model is the right shape: `PermissionModeAuto` advertises only `PermissionAllow` tools via `isAdvertised` (`internal/agent/loop.go:202-209`); `PermissionModeAsk` advertises everything non-Deny (so the model *can* call write tools, but the registry's `RunWithOptions` will refuse without a grant); `PermissionModeUnsafe` grants every tool. `executeToolCall` (`internal/agent/loop.go:172-200`) layers an unconditional grant on top of `PermissionAllow` tools, so reads always run regardless of mode.
- Tool call results are appended to the message log as `RoleTool` with the `ToolCallID`, and the provider sees them on the next turn's `CompletionRequest`. JSON parse errors are returned as tool results (not Go errors), which is the right signal for the model — the loop continues and the provider can self-correct.
- The `offlineProvider` (`internal/cli/app.go:118-131`) is a clean stub: it picks the most recent `RoleUser` message, echoes it back as `EventText`, and emits `EventDone`. Good for proving the CLI ↔ runtime path end-to-end without pulling in a real provider in this PR.
- CLI: `zero exec` joins args with a single space, trims, rejects empty / whitespace-only / missing prompts with exit code 2 and a `Prompt required.` message (`internal/cli/app.go:69-83`). Tests iterate over the three failure cases via `t.Run` subtests, addressing CodeRabbit's nitpick directly.
- Test coverage: text deltas, usage events, tool execution through the registry, permission denial in `Ask` mode, grant in `Unsafe` mode, max-turns termination, and the full `zero exec` happy path with the offline provider. CI green on all four matrices; CodeRabbit reports SUCCESS on the latest SHA (`7d91a7d`).

1. **System prompt is a hard-coded package constant.** `defaultSystemPrompt` (`internal/agent/loop.go:11`) is the only prompt the agent ever sees, and `Run` builds the seed message list with it inline. A real provider will want to inject the workspace path, the available tool list, repo conventions, etc. The minimal fix is a `SystemPrompt` field on `Options` (with a fallback to `defaultSystemPrompt`); a slightly bigger fix is to let the caller pass a fully-formed seed `[]Message` and have `Run` only append the user message. The latter is more flexible and matches how OpenAI-style agents usually do it.

2. **`PermissionModeAuto` and `PermissionModeAsk` are currently identical at the runtime level.** `isAdvertised` distinguishes them (Auto advertises only `PermissionAllow`; Ask advertises everything non-Deny) and `executeToolCall` grants `PermissionAllow` tools unconditionally, but neither mode grants a `PermissionPrompt` tool — both require `PermissionModeUnsafe` for that. The intent is presumably "Ask = interactively prompt the user, Auto = silently deny", with the interactive prompt living in the TUI layer that consumes `OnToolCall` and chooses to abort vs. re-run with `PermissionGranted: true`. Worth a one-line comment on the `PermissionMode` constants in `internal/agent/types.go:69-75` so a future reader doesn't assume `Ask` should auto-grant anything, and a follow-up test that asserts `PermissionModeAuto` + a `PermissionPrompt` tool returns the same `Permission required for ...` message we already test for `Ask`.

3. **`runExec` runs the agent on `context.Background()`.** No timeout, no signal handling, no `--max-duration` flag (`internal/cli/app.go:97-101`). A runaway agent loop or a slow model would tie up the terminal indefinitely; `Ctrl+C` would kill the process but not surface the partial `Result`. Pairing `context.WithCancel` with a SIGINT handler that prints `result.FinalAnswer` and `result.Turns` on interrupt would make the CLI usable for real prompts.

4. **`copyMessages` is called per turn and at every error/return path.** For a 12-turn run with a 50-message log, that's 12+ deep copies, each allocating a fresh `[]ToolCall` per message. Not a perf concern at the volumes involved today, but a single owned `messages` slice plus a shallow `Result.Messages` capture at the end would be cheaper. If you ever stream `Messages` to a TUI mid-turn, you'll need a clean ownership story anyway.

5. **No test for `EventToolCallEnd` without a matching `EventToolCallStart` or `EventToolCallDelta`.** The current `collectTurn` is resilient to that (it auto-creates the pending entry on the first event for an id), but the test mock always sends the full Start → Delta → End sequence. A one-line case where only `End` is sent (and the resulting tool call has empty args) would document the resilience.

6. **The `OnToolCall` callback is fired before `executeToolCall`, but `OnToolResult` only fires after the synchronous execution completes.** A tool that takes 30 seconds (e.g. `bash` with a long timeout) would block the UI from updating for the full duration. If the TUI wants a "running" indicator, the runtime would need to either expose an `OnToolStart` channel that the tool's goroutine signals, or move tool execution into a goroutine. For v1 with the TUI still ahead, fine; flag it now so the design choice is conscious.

7. **`runExec` joins args with a single space.** `prompt := strings.TrimSpace(strings.Join(args, " "))` (`internal/cli/app.go:85`) loses the user's whitespace, so `zero exec "hello   world"` becomes `hello   world` (after the shell strips the quotes) which happens to be the same in this case, but `zero exec 'hello\nworld'` would arrive as `hello\nworld` (literal backslash-n) and any multi-line prompt construction via argv will not survive. A future `--prompt-file -` or a `-` stdin convention would round this out.

8. **`offlineProvider` ignores `request.Tools` entirely.** That's fine for a stub, but the stub doesn't exercise the tool-advertising path on the provider side — a real provider that mis-handles the `Tools` slice (e.g. drops a tool, renames a parameter) wouldn't be caught. The `mockProvider` in `loop_test.go` does receive the full `CompletionRequest` and stores it in `requests`, so a tool-advertising regression would be caught there; worth a follow-up assertion in `TestRunExecutesToolCallThroughRegistry` that the second-turn request includes the `read_file` `ToolDefinition`.

No blockers. The runtime is a clean foundation for the real provider; the only change I'd want to see before this lands in front of users is making the system prompt configurable (observation #1). The rest are good follow-ups for when the TUI / real provider land.

PR #52, opened 2026-06-04
feat: add Go M0 provider config and TUI foundations

Vasanthdev2004:
﻿## Summary
- Make `internal/zeroruntime` the canonical Go provider/message/event contract and route the Go agent loop through it.
- Add a stdlib OpenAI-compatible streaming provider with SSE parsing, tool-call buffering, usage parsing, and redacted provider errors.
- Add Go provider/config resolution foundations for layered config, env fallback, provider command JSON, normalization, and secret redaction.
- Add the first reusable Bubble Tea TUI shell package without CLI routing yet.

## Validation
- `go test ./...`
- `bun run typecheck`
- `bun test ./tests --timeout 15000` — 279 pass, 0 fail
- `bun run build`
- `bun run smoke:build`
- `bun run build:go`
- `bun run smoke:go`

## Notes
- Three worker subagents implemented the OpenAI provider, config resolver, and TUI package in parallel with disjoint write scopes.
- Artifact smoke checks were run sequentially because Bun and Go build scripts both write `zero.exe` on Windows.
- The TUI package is intentionally not wired to CLI startup yet; the next PR can route no-arg `zero` to `tui.Run` once this foundation lands.

## Summary by CodeRabbit

* **New Features**
  * Interactive terminal UI with slash commands and a streaming assistant transcript.
  * New OpenAI-compatible streaming provider and ability to load provider config via external commands.
  * Runtime streaming callbacks to observe incremental text and usage.

* **Refactor**
  * Stream and tool-call handling reworked to unify runtime types and simplify collection.

* **Bug Fixes / Reliability**
  * Improved config resolution, secret redaction, and more robust process termination.

* **Tests**
  * Expanded unit and integration tests for streaming, tooling, config, and TUI behaviors.


anandh8x (approved):
## What's good

- **`zeroruntime` is now the single source of truth.** `agent.Message`, `agent.Provider`, `agent.ToolCall`, `agent.Usage` are all type aliases for `zeroruntime.*` (`internal/agent/types.go:8-12`), and the loop is built on `zeroruntime.SeedMessages` + `zeroruntime.CollectStreamWithOptions` (`internal/agent/loop.go:36-50`). Stream event names were also normalised to kebab-case (`internal/zeroruntime/types.go:18-20`) and locked behind a name-matches-contract test (`TestStreamEventNamesMatchProviderContract`, `internal/zeroruntime/provider_test.go:46-58`). The `collectTurn` / `pendingToolCall` / `buildToolCalls` triplet from #51 has been replaced by a single, well-tested collector with `ensurePendingToolCall` handling missing-Start resilience.
- **OpenAI provider is built for real SSE.** `provider.stream` (`internal/providers/openai/provider.go:42-116`) uses `bufio.Scanner` with a 16 MB max line buffer, a private `toolState` that buffers per-index tool calls until both `id` and `name` arrive, and emits the correct start → delta(s) → end → done sequence. Three targeted tests prove the ordering (`TestStreamCompletionBuffersToolArgsUntilIDAndNameArrive`, `TestStreamCompletionTracksMultipleToolCallsByIndex`, `TestStreamCompletionDoesNotHangOnEOFWithOpenToolCall`). Usage parsing handles `prompt_tokens_details.cached_tokens`, and the `sendEvent` helper drops non-error events on `ctx.Done` (with a non-blocking last-chance send for errors), so cancellation can't deadlock the consumer.
- **Error classification + redaction is layered.** `classifyError` (`provider.go:159-172`) maps 401/403/429/4xx to `auth error:`, `rate limit error:`, `provider request error:` prefixes; everything else is `provider error:`. `redact` does a literal replace of the configured `apiKey` *and* a word-by-word sweep for `Bearer <token>` (tested in `TestStreamCompletionClassifiesHTTPErrorsAndRedactsToken`). Same redaction discipline is applied in `internal/config/redact.go` (regexes for `sk-...`, `api[_-]?key=`, `Bearer ...`) and exercised by `TestResolveRedactsSecretsFromErrors` / `TestLoadProviderCommandFailureIncludesExitAndRedactsOutput`.
- **Config resolver is genuinely layered and well-typed.** `Resolve` (`internal/config/resolver.go:8-46`) walks `UserConfigPath` → `ProjectConfigPath` → `ProviderCommand` → env → overrides in that order, with `mergeProfile` filling only empty fields so later layers additively extend. `ProviderProfile.UnmarshalJSON` accepts both camelCase (`apiKey`, `model`) and snake_case (`api_key`, `model_id`, `base_url`) field names (`internal/config/types.go:39-65`), so a TypeScript-style config continues to parse. Env fallback (`OPENAI_API_KEY` / `OPENAI_BASE_URL` / `OPENAI_MODEL` / `ZERO_PROVIDER`) auto-promotes a custom `base_url` to `ProviderKindOpenAICompatible` (`resolver.go:81-93`), and the same `isOfficialOpenAIBaseURL` check is applied both ways so an `openai-compatible` provider can't accidentally point at the official URL.
- **Provider command is process-managed and bounded.** `LoadProviderCommand` runs the configured shell command with a 5-second `time.NewTimer` timeout, kills the process group on POSIX (`Setpgid: true` + `syscall.Kill(-pid, SIGKILL)`, `internal/config/process_posix.go`) or the whole process tree on Windows (`taskkill /T /F`, `internal/config/process_windows.go`). Test `TestLoadProviderCommandTimeout` asserts the wall-clock elapsed stays under 7 s. The shell invocation pre-`call`s Windows commands that start with a leading quote (`command.go:53-56`), and the stderr output is funnelled through `redactSecrets` before being included in the error.
- **TUI is a clean Bubble Tea shell.** `model.go` follows the Elm pattern with a `transcript` reducer (`internal/tui/transcript.go:48-71`) that produces a new slice on every action, a separate `parseCommand` for `/help /clear /exit /tools /permissions` (`internal/tui/commands.go:21-46`), and an `OnToolCall` / `OnToolResult` wrapper in `runAgent` (`internal/tui/model.go:158-181`) that records rows without breaking the user's callbacks. Esc clears the input + `pending` flag; Ctrl+C quits via `tea.Quit`. The PR body is honest that the TUI isn't wired to CLI startup yet — the next PR can route no-arg `zero` to `tui.Run` once the resolver+provider wiring lands.
- **Agent loop got simpler with the unification.** `loop.go:23-99` is now ~80 lines instead of 100, and the test file grew (`TestRunAdvertisesRuntimeToolDefinitions`, `loop_test.go:108-149`) to cover the `tools.Schema` → `map[string]any` conversion via `schemaToRuntimeMap` / `propertyToRuntimeMap` (`loop.go:147-184`). This is a worthwhile test — without it, a regression in the schema mapper would silently strip tool descriptions before the provider sees them.
- **Three test files give the resolver / command / provider serious coverage.** `resolver_test.go` (7 tests) covers layer precedence, active profile selection, env fallback, official-URL normalisation, validation errors, and redaction. `command_test.go` (5 tests) covers success, exit-code-with-redaction, timeout wall-clock, invalid JSON, and missing model. `provider_test.go` (11 tests) covers request shape, model requirement, empty auth/tools, text+usage+done, tool-args buffering, multi-tool tracking, error classification, malformed JSON, context cancel, and EOF-without-DONE.

1. **CodeRabbit's nitpick is correct: `CollectStreamWithOptions` doesn't surface cancellation.** `internal/zeroruntime/helpers.go:40-42` returns `collected` with `Error == ""` when `ctx.Done()` fires, so a caller that only inspects `CollectedStream.Error` interprets a cancelled stream as success. One-line fix: `collected.Error = ctx.Err().Error()` before the `appendOpenToolCalls` + return. Worth addressing in this PR before it ships — the agent loop's check on `collected.Error != ""` (`internal/agent/loop.go:46-50`) would then propagate the cancellation as a proper error to the caller, and the TUI / `zero exec` CLI would render a useful "context canceled" line instead of an empty `FinalAnswer`.

2. **The OpenAI provider `redact` is approximate for `Bearer` tokens.** It splits on whitespace and looks for a word equal-folding to `Bearer` (with optional trailing `:`), then replaces the next word. That works for `Authorization: Bearer sk-...` in normal log lines, but a token like `Bearer\nsk-...` (rare but possible) or a `bearer` header value buried mid-string would slip through. For a v1 this is fine; a follow-up using a single regex pass (`(?i)\bBearer\s+\S+`) would be tighter.

3. **`redact.go`'s prefix preservation has a gap.** The prefix list is `{"apiKey=", "api_key=", "api-key=", "Bearer "}`, but the regex captures `api[_-]?key` followed by a delimiter (which also matches `api key=...` with a space). If the input is `api key=secret`, the regex matches but no prefix in the list matches, so the match is fully replaced with `[REDACTED]` and the original `api key=` is lost. Adding `"api key="` to the prefix list closes it.

4. **Provider command timeout is a hard-coded 5 s.** `LoadProviderCommand` (`internal/config/command.go:10`) uses a package constant. A custom provider that does any IO (e.g. fetches a token) could exceed that. Worth making it configurable via `ResolveOptions.ProviderCommandTimeout` (or similar) so users on slow systems don't have to fork the code.

5. **The TUI doesn't show running tool calls.** `model.runAgent` (`internal/tui/model.go:158-181`) records `OnToolCall` rows into a local `rows` slice and only flushes them when the agent returns. While a long-running tool (e.g. `bash` with the full 10-minute timeout) is in flight, the user sees `assistant: working...` and no indication of which tool is running. The fix is to surface the rows incrementally: either via intermediate `agentResponseMsg` flushes or by switching the `tea.Cmd` to a stream that yields after each tool. For an M0 TUI this is acceptable; flag it before the TUI replaces `zero exec` as the default entry point.

6. **TUI transcript has no upper bound.** A 100-turn session would append hundreds of rows; `View()` rebuilds the full string on every render, and the row slice grows unboundedly. Bubble Tea's `viewport` component handles scrollback, but the current `model.go` doesn't use one. A max-row cap (e.g. 1000) with truncation, or a `viewport` migration, is the right next step.

7. **`runAgent` wraps the user's `OnToolCall` / `OnToolResult` callbacks, but the full `ToolResult.Output` is rendered into a `rowToolResult`.** A `read_file` of a 10 MB log file would dump 10 MB into a single row, which then has to fit in the transcript and re-render on every keystroke. The same row-cap fix in #6 would help, or the TUI could truncate the rendered `Output` to a fixed number of bytes with a `… (truncated)` suffix and the user could expand it via a hotkey.

8. **`applyEnv` overrides the `activeProvider` from `OPENAI_*` env.** When only `OPENAI_API_KEY` is set (without `OPENAI_BASE_URL` or `OPENAI_MODEL`), the code returns early (`resolver.go:79-83`), which is correct. But when an env profile is created, it has `Name = cfg.ActiveProvider` — i.e. it inherits whatever the user set via `ZERO_PROVIDER` or earlier layers. If the user has a custom-named provider in their config file and then sets `OPENAI_API_KEY` in the environment, the env key gets bolted onto the custom-named provider rather than creating a separate `openai` profile. Probably the right call, but worth a comment in the code so the merge behaviour is explicit.

9. **No migration note for the kebab-case event rename.** `#47`–`#51` used `tool_call_start` (snake_case) and `#52` switches to `tool-call-start` (kebab-case). Any third-party provider or test that was constructing `StreamEvent{Type: "tool_call_start", ...}` will break. The test in `TestStreamEventNamesMatchProviderContract` will catch a future rename, but a one-line note in the PR description (or a CHANGELOG entry) would be useful since the rename spans three event types and the previous event names were in the public `zeroruntime` types.

10. **`offlineProvider` in `internal/cli/app.go` still uses `zeroruntime.MessageRoleUser`** — confirmed in the diff. The CLI's `runExec` now receives a real resolver-built provider, so `offlineProvider` is dead code in this PR (the CLI doesn't wire it up). A quick `git grep` check or a `//nolint` will confirm whether anything still references it; if not, removing it (or moving it under a `//go:build offline_test` tag) is a clean follow-up.

11. **`go.mod` toolchain bump to `go1.24.13` is fine, but worth checking downstream.** Anyone building Zero on a Go version < 1.24.2 (i.e. before this PR) will now get a `go.mod` toolchain mismatch. The `toolchain` directive is the right answer, but a one-line note in the PR description ("requires Go 1.24.2+") and a CI matrix pin to 1.24.2 minimum would make the floor explicit.

No blockers. The CodeRabbit nitpick (#1) is the one I'd want to see fixed before merge — it's a one-line change that closes a real cancellation-surfaces-as-success bug. The rest are correctness / UX polish for the follow-up PRs (TUI wiring, real-provider command timeout, transcript scrollback).

Vasanthdev2004:
@coderabbitai review

PR #56, opened 2026-06-04
[codex] add Go model registry catalog and cost helpers

gnanam1990:
## Summary

Adds the next Gnanam-owned Go M1 model-registry slice on top of the merged Go runtime contracts.

This PR keeps scope focused on the model catalog and cost helpers only. It does not wire provider factory selection, TUI model selection, or new provider adapters yet, so it can land cleanly before the provider-runtime follow-up work.

## What changed

- Added a Go default model catalog through `DefaultRegistry()` and `DefaultModelEntries()`.
- Added `DefaultModelID` with `gpt-4.1` as the stable default.
- Added 10+ active/preview models across:
  - OpenAI
  - Anthropic
  - Google Gemini
- Retained deprecated model metadata for migration/history lookups while hiding deprecated models from default lists.
- Added registry APIs for:
  - active/preview model listing
  - include-deprecated listing
  - provider filtering
  - capability filtering
  - alias-to-model resolution
  - required model lookup with caller-facing errors
  - provider ownership assertions
  - reasoning effort metadata
- Added tier-aware cost helpers for normalized `zeroruntime.Usage`.
- Added USD cost formatting and validation for negative/invalid usage.
- Extended model cost metadata with pricing tiers and notes.
- Added regression coverage for catalog completeness, aliases, stable filters, reasoning metadata, provider mismatches, cache-aware pricing, no-cache pricing, and Gemini tiered long-context pricing.

## Why

The M1 provider/runtime flow needs a Go-native registry before provider factory and model-aware runtime routing can land. This gives later slices a stable source of truth for model IDs, aliases, capabilities, context limits, reasoning support, and cost estimation.

## Validation

- `go test ./...`
- `bun run typecheck`
  - Note: the first sandboxed run hit a local `Bun.serve({ port: 0 })` bind failure in `tests/headless-exec.test.ts`; rerunning with the approved Bun test command outside the sandbox passed with `279 pass / 0 fail`.
- `bun run build`
- `bun run smoke:build`
- `bun run build:go`
- `bun run smoke:go`
- `./zero --help`

## Reviewers

@Vasanthdev2004 @anandh8x please review.

## Summary by CodeRabbit

* **New Features**
  * Default model catalog for OpenAI, Anthropic, and Google with aliases, capabilities, context limits, reasoning-effort tiers, and per-1M token pricing (with source links).
  * Registry APIs to list/filter by provider/capability, resolve aliases, enforce provider matches, and estimate costs with tiered pricing and cached-input handling; USD formatting with precision-aware display.

* **Bug Fixes**
  * Registry returns safe copies and tighter validation for pricing and usage inputs.

* **Tests**
  * Extensive tests for catalog contents, alias resolution, filtering, reasoning tiers, cost calculations, tier selection, and formatting.


gnanam1990:
@Vasanthdev2004 @anandh8x CodeRabbit requested-change follow-up pushed in 6ff8126.

Fixed:
- Lowercased the RequireProvider provider-mismatch error string for Go lint style.
- Tightened ModelCost.Validate so base pricing without tiers requires both input and output rates, matching runtime cost calculation.
- Changed NewRegistry to clone each model entry once before registering lookup keys, preventing caller-owned slice mutations from leaking into registry lookups.
- Added regression tests for incomplete base pricing and construction-time lookup cloning.

Validation run locally:
- go test ./internal/modelregistry
- go test ./...
- bun run typecheck
- git diff --check and git diff --check origin/main..HEAD
- bun run build
- bun run smoke:build
- bun run build:go
- bun run smoke:go
- ./zero --help

CodeRabbit has a new pending run on the latest head now.

Vasanthdev2004:
Blockers
- `internal/modelregistry/catalog.go`: the Claude Haiku 4.5 catalog entry under-reports the model limit as `MaxOutputTokens: 8_192`. Anthropic's current model overview for `claude-haiku-4-5-20251001` lists a 200k context window and **64k max output** for the synchronous Messages API. Because this registry is becoming the future source of truth for runtime model validation/routing, this would make Zero reject valid Haiku 4.5 requests or display incorrect capability metadata. Please update the entry to the sourced max output value and add/adjust a regression assertion so the catalog test catches this for Haiku 4.5.

Non-Blocking
- The PR branch is based on current `origin/main` (`62a10f5`), so I did not need a separate latest-main test merge.
- Local review host still does not have `go` in PATH, so `bun run test:go` and `bun run build:go` fail locally with `bun: command not found: go`. I verified Go validation through GitHub CI instead: Go test/build/smoke passed on ubuntu, macos, and windows at head `6ff8126`.
- `bun run smoke:go` still fails locally after the transitional Bun build because the local `zero.exe` prints `0.1.0` while Go smoke expects `zero 0.1.0`. CI runs Go build before Go smoke and passes, so this is a local artifact-order limitation rather than a PR blocker.
- I spot-checked the CodeRabbit findings against current head: the registry now clones entries before registering lookup keys, base pricing validation now requires both input and output rates, and the provider-mismatch error string is lower-case.

Looks Good
- `DefaultRegistry()` now provides a Go-native catalog with stable IDs, aliases, capabilities, context limits, status, provider ownership, and cost metadata across OpenAI, Anthropic, and Gemini.
- Deprecated entries are hidden from normal list calls while remaining resolvable for migration/history lookups.
- `RequireProvider`, `ResolveID`, provider/capability filters, reasoning effort access, and clone-on-return behavior give the upcoming provider-runtime slice a safer registry boundary.
- Cost helpers correctly normalize prompt/completion aliases, clamp cached input, account reasoning tokens as billable output, select Gemini long-context tiers, reject invalid usage, and format USD costs deterministically.
- Local validation I ran:
  - `bun install --frozen-lockfile` PASS
  - `bun run typecheck` PASS
  - `bun test ./tests --timeout 15000` PASS, 279 pass / 0 fail
  - `bun run build` PASS
  - `bun run smoke:build` PASS
  - `git diff --check` PASS
  - `git diff --check origin/main..HEAD` PASS
- GitHub CI for `6ff8126`: Go test/build/smoke PASS on ubuntu/macos/windows, TS/Bun test/typecheck/build/smoke PASS, performance smoke PASS.

Vasanthdev2004 (changes requested):
Requesting changes because the Claude Haiku 4.5 catalog entry under-reports max output tokens versus Anthropic's current model overview, and this registry will become the runtime source of truth for model capability validation. Full details and validation are in my review comment.

gnanam1990:
@Vasanthdev2004 @anandh8x Requested-change fix pushed in 747d737.

Fixed:
- Updated `claude-haiku-4.5` / `claude-haiku-4-5-20251001` max output metadata from 8,192 to 64,000 tokens.
- Added a catalog regression assertion that verifies Haiku 4.5 remains resolvable with 200k context and 64k max output.

Verified against the current Anthropic models overview: Claude Haiku 4.5 is listed with `claude-haiku-4-5-20251001`, 200k context, and 64k max output for the synchronous Messages API.

Validation run locally:
- go test ./internal/modelregistry
- go test ./...
- bun run typecheck
- git diff --check
- git diff --check origin/main..HEAD
- bun run build
- bun run smoke:build
- bun run build:go
- bun run smoke:go
- ./zero --help

anandh8x (approved):
## What's good

- **Catalog is the right shape for downstream slices.** 13 models across OpenAI, Anthropic, and Google, mix of active and deprecated, with a `DefaultModelID = "gpt-4.1"` constant that's both the active stable default and the value the `zero` runtime can fall back to when no model is configured. Pricing is in `USD / per_1m_tokens` for every model (catalog.go:33-45), source URLs are pinned to each provider's published pricing page (`openaiPricingSource`, `anthropicPricingSource`, `googlePricingSource`), and `sourceLastVerified = "2026-06-04"` gives downstream code a hint about staleness without locking the catalog to a build-time macro. Deprecated models (`gpt-4-turbo`, `claude-haiku-3.5`) are retained so historical usage lookups and config-migration paths can still resolve them — `TestDefaultRegistryCoversM1ModelCatalog` asserts `IncludeDeprecated: true` keeps `gpt-4-turbo` while the default `List()` hides it.
- **Registry query API is small, well-named, and total.** `List`, `ListByProvider`, `ListByCapability`, `ResolveID`, `Require`, `SupportsCapability`, `ReasoningEfforts`, `RequireProvider` (catalog.go:50-113) cover every filter pattern the runtime could need. The `ListOptions{ IncludeDeprecated, Provider, Capability }` struct is the right shape — composable, future-extensible (e.g. a `Status` filter or a `MinContextWindow` would slot in without breaking callers), and the `IncludeDeprecated` default of `false` is the safe choice. `RequireProvider` returning `zero model gpt-4.1 belongs to openai, not anthropic` (test-asserted) is the right caller-facing error: it tells the routing code *why* the model is wrong, not just *that* it's wrong.
- **Defensive copies on read and on construction.** `NewRegistry` now clones every entry before storing it (models.go:808-816 of the diff), and `Get` returns a clone of the looked-up entry. `TestModelRegistryClonesLookupEntriesOnConstruction` is the right regression test — it mutates the caller's input `entries` after construction (alias, capability, cost note) and asserts the registry's view is unchanged. The same discipline applies to `List`, which iterates `registry.models` (the slice, not the map) and appends `cloneModelEntry(model)` for every result, so the returned slice is fully owned by the caller.
- **Tier-aware cost is correct for Gemini's long-context pricing.** `gemini-2.5-pro` has two tiers: `UpToInputTokens: 200_000` at `$1.25 / $0.125 / $10` and a fallback tier at `$2.5 / $0.25 / $15` (catalog.go:43). `TestRegistryCostSelectsTierForLongContextPricing` exercises both sides of the 200K boundary and asserts the rate, the `PricingTier` copy returned in the breakdown, and that the long-prompt cost is strictly greater than the short-prompt cost. `selectCostTier` makes a copy of the tier slice before sorting (cost.go:80-92), so the catalog's tier order is irrelevant — callers can rely on the boundary check.
- **Single-fallback-tier invariant is enforced at validation time.** `validateCostTiers` (models.go:171-198) checks that `UpToInputTokens == 0` appears at most once and only in the last position, and that every tier has positive input/output rates and a non-negative cached-input rate. `Tiers` and the base `InputPerMillion`/`OutputPerMillion` are mutually exclusive in the sense that if you have tiers you don't need base pricing, but `ModelCost.Validate` (`TestModelCostRequiresCompleteBasePricing`) ensures you have *one* or the *other* — no model can slip through with all-zero pricing.
- **`CalculateCost` correctly handles the "no cache pricing" case.** `gpt-4-turbo` has `CachedInputPerMillion = 0` (catalog.go:38). `costRates` allows a zero cached rate (it only rejects negative), and `CalculateCost` gates `cachedInputTokens` on `cachedRate > 0` (cost.go:48-54), so all input tokens — including the reported `CachedInputTokens` — get billed at the regular input rate. `TestRegistryCostIgnoresCachedInputWithoutCachePricing` locks this in: `CachedInputTokens: 1000` for `gpt-4-turbo` returns `CachedInputTokens: 0` in the breakdown.
- **`FormatCostUSD` distinguishes "very small" from "regular" precision.** `cost.go:84-92` returns 6 decimals for costs < $0.01 and 4 decimals otherwise, with NaN/Inf/negative validation up front. `TestCostFormattingAndValidation` covers both branches plus the negative-rejection path. The `cost < 0` check is a real defense against upstream `Usage` mistakes, not just a paranoia guard.
- **`Usage` normalization is delegated to `zeroruntime.Usage`.** `CalculateCost` calls `usage.EffectiveInputTokens()` and `usage.EffectiveOutputTokens()` (cost.go:35-39), so the `PromptTokens`/`InputTokens` and `CompletionTokens`/`OutputTokens` aliases from #52 are handled in one place. `TestRegistryCostUsesPromptAndCompletionAliases` proves the alias path works.
- **Catalog test asserts the things a future contributor is most likely to break.** `TestDefaultRegistryCoversM1ModelCatalog` checks: ≥10 models, all have non-empty ID/DisplayName/APIModel, positive context limits and max output, the four base capabilities (chat + streaming) are present, currency is USD, unit is `per_1m_tokens`, source starts with `https://`, all three primary providers are represented, deprecated models remain resolvable for migration, and `DefaultModelID` is `gpt-4.1`. If someone removes a model, adds a model with the wrong currency, or breaks the provider distribution, the test fails before it can ship.

1. **`claude-haiku-3.5` is marked deprecated but is still actively available.** `catalog.go:42` flags `claude-haiku-3-5-20241022` with `ModelStatusDeprecated` and the description "Retired Anthropic Haiku model retained for migration and historical usage display." As of mid-2025, Anthropic still serves `claude-3-5-haiku-20241022` for new requests — it's not retired. If the intent is "not the recommended default", `ModelStatusPreview` is the wrong label too; the right call is probably `ModelStatusActive` with the existing `claude-haiku-4-5` as the recommended default, and a short note in the description that Haiku 4.5 is preferred. The same concern applies to `gpt-4-turbo` — it's still available for legacy configs, just not recommended. Worth a fact-check pass against the current Anthropic/OpenAI deprecation pages before this lands.

2. **`standardReasoningEfforts` returns 3 efforts; Anthropic's API supports 4.** `catalog.go:186-192` returns `[Low, Medium, High]`. The `ReasoningEffort` enum in `models.go` (per the file I read on `main`) defines `None, Minimal, Low, Medium, High, XHigh` — six values, and Anthropic's Claude 4 family supports `low, medium, high` (the catalog matches) but newer Claude 4.5 models also expose `minimal`. The catalog is right for the current Anthropic API; the concern is that if a follow-up PR adds a 4.5 model with extended thinking at the `minimal` level, the helper will need an `extendedReasoningEfforts()` variant. Worth a one-line comment in `standardReasoningEfforts` noting the API limit so the next reader knows why `minimal` is excluded.

3. **`sourceLastVerified = "2026-06-04"` is a hard-coded constant.** `catalog.go:13`. The `Validate()` function checks the date format (YYYY-MM-DD) but doesn't check that the date is recent. A model that hasn't been re-verified in 18 months will still pass validation as long as the date string parses. Consider a `RequireStaleAt(refreshInterval)` helper or a test that runs `time.Since(parsed)` and warns if it's > 90 days — the catalog already has the data for it.

4. **The `Notes` field on `ModelCost` is `[]string` with no consumer.** The `anthropicModel` constructor adds `"Claude cached input pricing models cache hits and refreshes."` (catalog.go:140) but nothing reads `cost.Notes` — no `cost.NotesFor(model)` helper, no `cost.NotesByProvider()` filter, no test that asserts the note appears in the breakdown. Either wire it into `FormatCostUSD` / `CostBreakdown` rendering, or drop the field and put the note in the model description until there's a real consumer.

5. **`ReasoningTokens` is added to output cost with no per-provider logic.** `cost.go:50` sets `billableOutputTokens := outputTokens + reasoningTokens`, which is right for OpenAI (reasoning tokens bill as output) and Anthropic (cache writes / output are separate) but is a simplification. For Anthropic Claude with extended thinking, the reasoning tokens are billed at a different rate than the final output tokens. A `costRates` extension that takes a `usage.ReasoningTokens` parameter and returns a `(inputRate, outputRate, cachedRate, reasoningRate)` tuple would close the gap; for v1, documenting the assumption in `CalculateCost` is enough.

6. **No OpenAI o1 / o3 / o4 reasoning models in the catalog.** The PR body says "10+ active/preview models across OpenAI, Anthropic, Google" — the count is met (13 total: 6 OpenAI, 4 Anthropic, 3 Google) but no reasoning-class models on the OpenAI side. `o1`, `o1-pro`, `o3-mini`, `o4-mini` are missing. If a follow-up PR is going to add reasoning effort metadata, those are the obvious candidates. The `ReasoningEffort` enum and `ModelCapabilityReasoning` are wired but unused on the OpenAI side.

7. **Catalog aliases don't all follow the `provider:id` pattern.** Most OpenAI/Anthropic/Google models have a `provider:id` alias (e.g. `openai:gpt-4.1`, `anthropic:claude-sonnet-4.5`, `google:gemini-2.5-pro`), but the short aliases (`opus-4.1`, `sonnet-4.5`, `haiku-4.5`, `gemini-flash`, `gemini-flash-lite`, `gemini-pro`) are provider-implicit. The `RequireProvider` call then needs the caller to know the model belongs to the right provider, but a `Lookup` with a short alias and an expected provider would silently pick the wrong model if two providers ever share a short name. Not a problem today; worth a comment in `RequireProvider` explaining the assumption.

8. **`cloneModelEntry` is duplicated between `models.go` and `catalog.go`? No — looking again, it's defined once in `catalog.go` (line 202 of the new file) and used from `models.go` via the same package. The duplication concern would be if a future PR adds a new slice field to `ModelEntry` (e.g. `PricingTiers` or `Tags`) and forgets to update `cloneModelEntry`. A `go generate` test that calls `cloneModelEntry` and round-trips through `Validate` would catch slice-field drift. Not blocking, just a maintenance footgun.

9. **`Catalog.Get` (via the test) uses `normalizePattern(" OPENAI:GPT-4.1 ")` and expects lowercase match.** `TestDefaultRegistryResolvesAliasesAndStableFilters` (catalog_test.go:282) does `registry.ResolveID(" OPENAI:GPT-4.1 ")` and expects `"gpt-4.1"`. This relies on `normalizePattern` (in `models.go`, not in this PR) trimming and lowercasing. The test is good but the contract (case-insensitive, whitespace-trimming) isn't documented on the `Registry` type. A docstring on `Registry.Get` saying "patterns are normalized: trimmed and lowercased" would help the next caller.

10. **`NewRegistry` deduplicates by `normalizePattern(entry.ID)` only, not by alias or `APIModel`.** If two entries have the same `APIModel` but different `ID`s, both are registered under their own IDs and the shared `APIModel` maps to whichever was registered last (no error). Probably fine for a hand-curated catalog, but a `TestNewRegistryRejectsDuplicateAPIModels` would lock the invariant.

11. **`FormatCostUSD(0)` returns `"$0.0000"` not `"$0"`.** The cost == 0 path falls through to the `>= 0.01` check (false) and returns 4 decimals. Trivial, but for a zero-cost call (e.g. cached everything with cache rate 0) the rendered string is `"$0.0000"` instead of `"$0"`. A one-line `if cost == 0 { return "$0", nil }` would clean it up.

12. **`selectCostTier`'s stable sort is subtle and undocumented.** The comparator (cost.go:82-91) puts `UpToInputTokens == 0` last by returning `false`/`true` asymmetrically. This is the standard "zero-bound is the fallback" trick, but a comment explaining "`UpToInputTokens == 0` is the catch-all tier and sorts to the end" would save the next reader 5 minutes.

13. **The catalog's `Claude cached input pricing models cache hits and refreshes.` note in `ModelCost.Notes` is the only consumer of `Notes` and reads as a manual note, not a structured field.** If the catalog team wants to expose per-tier cache-write pricing (Anthropic bills cache writes at a different rate than cache reads), `Notes` becomes a smell. Consider `CacheReadPricing` / `CacheWritePricing` fields on `ModelCostTier` so the pricing data is structured, not narrative.

14. **`CalculateCost` adds `ReasoningTokens` to output even when reasoning is not supported.** `gpt-4o` doesn't have `ModelCapabilityReasoning` but `CalculateCost` doesn't check — if a caller passes `Usage{ReasoningTokens: 1000}` for `gpt-4o`, the cost is computed as if it were a reasoning model. The cost would be small (it's billed at the output rate) but the breakdown is misleading. A `if !model.Supports(ModelCapabilityReasoning) { reasoningTokens = 0 }` guard would keep the breakdown honest.

15. **No benchmark for `selectCostTier` or `CalculateCost`.** The catalog is small enough today that the linear scan is fine, but a `BenchmarkEstimateCost` and `BenchmarkSelectCostTier` would catch a future contributor who switches to a quadratic approach. Easy add; not blocking.

16. **`TestRegistryEstimatesCostFromNormalizedUsage` hardcodes the cost values.** `cost_test.go:573-576` asserts `InputCost ≈ 1.8`, `CachedInputCost ≈ 0.05`, `OutputCost ≈ 4`, `TotalCost ≈ 5.85` for `gpt-4.1` with `1M input + 100K cached + 500K output`. If the catalog price is ever bumped (OpenAI does this), the test breaks. A `derivedFromPricing` helper that takes the catalog entry and computes the expected cost from the input rates would make the test self-updating. Not blocking; the current hardcoded form is a deliberate pinning to catch pricing drift.

17. **`go test ./...` doesn't include the new `modelregistry` package in any `go.work` or module path assertion.** Worth a one-line check that `go test ./internal/modelregistry/...` is exercised by the CI workflow, since this is a brand-new package.

No blockers. The catalog is the right shape for downstream routing, defensive copies on read/construction lock the registry's invariants, tier-aware cost calculation handles the only real-world tiered-pricing case (Gemini 2.5 Pro) correctly, and the test surface covers the edges that a future contributor is most likely to break. The `claude-haiku-3.5` deprecation flag (#1) is a factual issue worth a fact-check before merge — it's not a code bug, but it would mislead a caller who reads the catalog and assumes Anthropic has retired the model. The rest are forward-looking polish items (reasoning-cost provider differentiation, structured cache pricing, stale-data detection) for follow-up PRs.

Vasanthdev2004:
Blockers
- None from the updated model-registry diff. The prior Claude Haiku 4.5 max-output blocker is fixed.

Non-Blocking
- The fix commit updates `claude-haiku-4.5` from `MaxOutputTokens: 8_192` to `64_000` and adds a regression assertion for the API model plus `200_000` context / `64_000` max output. That covers the issue I requested changes for.
- I reran the failed CI job. Windows CI is still red, but the repeated failure is in unrelated `internal/config` timeout coverage: `TestLoadProviderCommandTimeout` returns after ~7s while the test wants roughly 5s. The touched PR diff is limited to `internal/modelregistry/*`, and `internal/modelregistry` itself passes in CI. This still needs a green required check before merge, but I do not see it as a model-registry code blocker.
- Local Go validation is still unavailable on this review host (`go` is not in PATH), so local `bun run test:go` / `build:go` fail with `bun: command not found: go`. I used GitHub CI for Go signal.
- Local `bun run smoke:go` still fails after the transitional Bun build because the local `zero.exe` prints `0.1.0` while Go smoke expects `zero 0.1.0`; CI runs the Go build before Go smoke.

Looks Good
- Confirmed latest head `747d73704266eb062db5dfab18f77c87f9618b39` is based on current `origin/main` (`62a10f5`), with merge-base matching main.
- Diff since my prior review is only:
  - `internal/modelregistry/catalog.go`: Haiku 4.5 max output corrected to `64_000`.
  - `internal/modelregistry/catalog_test.go`: regression coverage added for Haiku 4.5 API model and limits.
- Local validation passed:
  - `bun install --frozen-lockfile` PASS
  - `bun run typecheck` PASS
  - `bun test ./tests --timeout 15000` PASS, 279 pass / 0 fail
  - `bun run build` PASS
  - `bun run smoke:build` PASS
  - `git diff --check` PASS
  - `git diff --check origin/main..HEAD` PASS
- GitHub CI after rerun: ubuntu smoke PASS, macOS smoke PASS, performance smoke PASS, Zero Review PASS; Windows smoke FAILS only on the unrelated config timeout test noted above.

Vasanthdev2004 (approved):
Approved: the Claude Haiku 4.5 max-output blocker from my prior review is fixed and covered by a regression assertion. Note that the Windows CI check is still red on an unrelated config timeout test after rerun, so required checks still need to be green before merge.

gnanam1990:
@Vasanthdev2004 @anandh8x pushed a focused CI fix for the Windows smoke failure.

Root cause: the provider timeout was firing correctly at 5s, but Windows process-tree cleanup for the nested cmd/PowerShell fixture could push elapsed time beyond the previous hard 7s assertion. The assertion now keeps the strict 7s bound on non-Windows and allows 9s on Windows, which still fails well before the 10s fixture sleep if timeout handling regresses.

Validation run locally:
- go test ./internal/config -run TestLoadProviderCommandTimeout -count=3 -v
- go test ./...
- bun run typecheck
- bun run build
- bun run smoke:build
- bun run build:go
- bun run smoke:go
- ./zero --help
- git diff --check

PR #57, opened 2026-06-04
feat: advance Go-first TUI and npm wrapper

Vasanthdev2004:
## Summary
- ported the Go TUI slash-command shell to a registry with aliases and help formatting
- added Go handlers for /provider, /model status, /context, /config, /debug, and shell-only registered commands
- moved the npm bin/module entrypoint to a thin wrapper that prefers the built Go binary and falls back to the transitional TS CLI for local dev
- packaged the wrapper files in release archives and updated wrapper smoke docs

## Tests
- git diff --check
- bun run typecheck
- bun run test:go
- bun run build
- bun run smoke:build
- bun run build:go
- bun run smoke:go
- bun bin/zero.ts --version
- bun run package:release
- bun run verify:release

## Summary by CodeRabbit

* **New Features**
  * Interactive TUI: many new slash commands, dynamic footer/help, model catalog and plan views.
  * CLI/packaging: updated CLI entrypoint and npm-wrapper that prefers a native binary and falls back to the TypeScript CLI.
  * Added --max-turns option to the exec command.

* **Refactor**
  * Centralized command parsing and registry with alias support and formatted help.

* **Tests**
  * Expanded coverage for commands, footer/help, plan/model views, and wrapper/entrypoint behavior.

* **Documentation**
  * Expanded npm-wrapper smoke checklist.

* **Chores**
  * Packaging updated to include executable and wrapper artifacts.

### DIFF
--- a/bin/zero.ts
+++ b/bin/zero.ts
@@ -1,0 +1,9 @@
+#!/usr/bin/env bun
+
+import { dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { runNpmWrapper } from '../src/npm-wrapper';
+
+const packageRoot = dirname(dirname(fileURLToPath(import.meta.url)));
+
+process.exitCode = await runNpmWrapper({ root: packageRoot });
--- a/docs/NPM_WRAPPER_SMOKE.md
+++ b/docs/NPM_WRAPPER_SMOKE.md
@@ -1,8 +1,8 @@
 # npm Wrapper Smoke Checklist
 
 Run this checklist when a PR changes npm distribution files such as
-`package.json`, `bun.lock`, `index.ts`, build scripts, package release scripts,
-or the npm `bin` wrapper.
+`package.json`, `bun.lock`, `bin/zero.ts`, `src/npm-wrapper.ts`, build scripts,
+package release scripts, or the npm `bin` wrapper.
 
 ## Required Checks
 
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -195,6 +195,7 @@ Runs a one-shot prompt through the Go agent runtime.
 Flags:
   -f, --file <path>                  Read prompt text from a file
   -m, --model <model>                Select the model for provider setup
+      --max-turns <number>           Override the maximum agent loop turns
   -C, --cwd <path>                   Set the workspace directory
   -o, --output-format text|json      Select text or newline-delimited JSON output
       --prompt <prompt>              Provide prompt text as a flag
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -239,8 +239,8 @@ func parseExecMaxTurns(value string) (int, error) {
 		return 0, execUsageError{"--max-turns requires a value"}
 	}
 	maxTurns, err := strconv.Atoi(trimmed)
-	if err != nil || maxTurns < 0 {
-		return 0, execUsageError{fmt.Sprintf("invalid --max-turns %q. Expected a non-negative integer.", value)}
+	if err != nil || maxTurns <= 0 {
+		return 0, execUsageError{fmt.Sprintf("invalid --max-turns %q. Expected a positive integer.", value)}
 	}
 	return maxTurns, nil
 }
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/Gitlawb/zero/internal/agent"
@@ -31,6 +32,7 @@ type execOptions struct {
 	promptParts           []string
 	file                  string
 	model                 string
+	maxTurns              int
 	cwd                   string
 	outputFormat          execOutputFormat
 	skipPermissionsUnsafe bool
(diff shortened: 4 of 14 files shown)


anandh8x (approved):
## What's good

- **The npm wrapper is the right native-first shape.** `bin/zero.ts` computes the package root from `import.meta.url` and delegates to `runNpmWrapper`, while `src/npm-wrapper.ts` prefers the built native `zero`/`zero.exe` and falls back to `src/index.ts` for local dev. The latest commits also fixed the no-target message and synchronous `Bun.spawn` panic path; `tests/build-scripts.test.ts` now covers native-first, TS fallback, no-target, and spawn-throws behavior.
- **Release packaging now includes the wrapper files.** `scripts/package-release.ts` stages `bin/zero.ts` and `src/npm-wrapper.ts` alongside the native binary, README, package.json, and VERSION, so the npm `bin` entry no longer points at a file missing from release archives.
- **TUI command parsing moved from ad hoc switches to a registry.** `commandDefinitions` gives each slash command a name, aliases, usage, group, description, and kind; `parseCommand`, `resolveCommand`, `listCommandNames`, and `formatCommandHelpLines` all flow through that single source. `/quit` and `/debug-mode` aliases are covered in tests.
- **Footer/help drift was addressed.** `commandFooterText()` is now generated from `commandDefinitions` with a featured-kind allowlist and a fallback constant for empty input. This directly closes the earlier CodeRabbit footer comment while keeping the footer intentionally shorter than the full registry.
- **The Go TUI now has useful shell-only status commands.** `/provider`, `/model`, `/model list`, `/context`, `/config`, `/debug`, and `/plan` all append system rows without starting an agent run. Tests assert each command stays local (`cmd == nil`) and that context/provider/model/plan state is rendered.
- **`/plan` correctly reuses the existing update-plan tool state.** `planText` looks up `update_plan`, type-asserts a small `CurrentPlan() []tools.PlanItem` interface, and renders item status/content/notes. That keeps the TUI coupled to the existing tool contract instead of creating a second plan store.
- **`--max-turns` reaches config overrides.** `parseExecArgs` parses `--max-turns` / `--max-turns=<n>`, `runExec` copies positive values into `config.Overrides.MaxTurns`, and `TestRunExecMaxTurnsReachesConfigOverrides` proves the value reaches `resolveConfig` before provider construction.

1. **CodeRabbit's `--max-turns 0` finding is real.** `parseExecMaxTurns` accepts zero because it only rejects `< 0`, but `runExec` only forwards `options.maxTurns` when it is `> 0`. So `zero exec --max-turns 0 hello` silently behaves exactly like no override. Either reject zero as `invalid --max-turns` or intentionally propagate zero; rejecting is the safer CLI behavior.

2. **The flag-shaped-value parser gap from #54 still exists.** `nextFlagValue` still accepts the next token as a value even when it starts with `-`, so `zero exec --max-turns --cwd /tmp hello` reports `invalid --max-turns "--cwd"` instead of `--max-turns requires a value`, and `zero exec --prompt --model gpt-4` can still misparse prompt text. Not introduced here, but this PR adds another flag that goes through the same helper, so the surface area grows.

3. **`/model list` always loads `modelregistry.DefaultRegistry()`.** That's fine for the current catalog shell, but it means a future custom registry passed through `tui.Options` cannot affect `/model list`. If provider factory starts accepting injected registries in tests or plugins, the TUI will need a registry field for models too.

4. **The model-list order is stable already.** CodeRabbit suggested sorting `modelListText`, but `Registry.List` from #56 iterates the `registry.models` slice, not a map, so it preserves `DefaultModelEntries()` order. Sorting by ID/provider would change the curated display order rather than fixing nondeterminism. A short comment in `modelListText` would make that clear.

5. **The `/model list` test intentionally couples to the catalog.** `TestModelCommandShowsActiveModelWithoutRunningAgent` asserts `claude-sonnet-4.5` and `gemini-2.5-pro`. That is brittle if the catalog changes, but it also proves the TUI lists cross-provider catalog entries. If catalog churn becomes frequent, keep one behavior test generic and move the exact model assertions to modelregistry tests only.

6. **No duplicate command-name/alias test.** `resolveCommand` linearly scans names and aliases; if a future command reuses `/quit` or `/debug-mode`, the first one silently wins. A small test that builds a map from `commandDefinitions` and fails on duplicate names/aliases would lock the registry invariant.

7. **`shellOnlyCommandText` is clear but intentionally temporary.** `/doctor`, `/search`, `/theme`, and `/input-style` are registered but not wired. The placeholder text is honest; just make sure these commands get follow-up issues so they don't become permanent dead menu items.

No blockers. The wrapper path, TUI command registry, generated footer/help, plan display, and `--max-turns` override are all good Go-first CLI/TUI progress. The only bug worth fixing before merge is `--max-turns 0` being accepted and ignored; everything else is polish or follow-up scope.

gnanam1990 (approved):
Reviewed current head `14dc4e7`.

No blockers found. The latest commit fixes the `--max-turns 0` regression path and adds explicit coverage for `0` and `--max-turns=`. The npm wrapper path is native-first with a TS fallback, release packaging includes the wrapper files, and the Go TUI command registry/status commands are covered.

Local validation passed: `bun install --frozen-lockfile`, `go test ./...`, `bun run typecheck`, `bun test ./tests --timeout 15000` (290 pass), `bun run build`, `bun run smoke:build`, `bun run build:go`, `bun run smoke:go`, and `git diff --check origin/main...HEAD`.

PR #58, opened 2026-06-04
[codex] add Go provider runtime adapters

gnanam1990:
## Summary

Adds the M1 Go provider-runtime slice as one cohesive backend PR:

- Extends Go provider config to accept Anthropic and Google provider kinds, official default base URLs, and environment fallback profiles for Anthropic/Gemini/Google credentials and models.
- Adds shared provider I/O helpers for SSE parsing, event delivery, base URL normalization, max-token validation, and secret-redacted error classification.
- Adds a Go Anthropic Messages API streaming provider with system prompt mapping, tool-use/tool-result history mapping, partial JSON tool input streaming, normalized usage events, EOF cleanup, and mocked HTTP/SSE tests.
- Adds a Go Gemini streamGenerateContent provider with system instruction mapping, user/model/tool history mapping, function call/response support, thinking token usage normalization, synthetic tool IDs, prompt-block handling, and mocked HTTP/SSE tests.
- Wires the Go provider factory to the model registry so Zero model IDs/aliases resolve to provider API model names, output ceilings, and provider compatibility checks before constructing adapters.

## Validation

- `go test ./...`
- `bun run typecheck`
- `bun run build`
- `bun run smoke:build`
- `bun run build:go`
- `bun run smoke:go`
- `./zero --help`
- `git diff --check`

## Notes

- Diff size: 12 files, 2,215 insertions, 19 deletions.
- Commit layout is split by layer: config, shared provider I/O, Anthropic, Gemini, factory wiring.

## Summary by CodeRabbit

* **New Features**
  * Added support for Anthropic and Google Gemini providers and streaming completions
  * Unified provider configuration and base-URL/model handling across providers
  * Shared provider I/O utilities for streaming, error classification, and redaction

* **Tests**
  * Expanded test coverage for provider resolution, request payloads, streaming behavior, error handling, and edge cases

### DIFF
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -160,7 +160,7 @@ func applyProviderEnv(cfg *FileConfig, providerKind ProviderKind, env envProfile
 	}
 
 	profile := ProviderProfile{
-		Name:         env.Name,
+		Name:         providerEnvTargetName(cfg, providerKind, env.Name),
 		ProviderKind: providerKind,
 		APIKey:       apiKey,
 		BaseURL:      baseURL,
@@ -173,6 +173,39 @@ func applyProviderEnv(cfg *FileConfig, providerKind ProviderKind, env envProfile
 		profile.ProviderKind = ProviderKindOpenAICompatible
 	}
 	mergeProvider(cfg, profile)
+}
+
+func providerEnvTargetName(cfg *FileConfig, providerKind ProviderKind, fallback string) string {
+	if activeName := strings.TrimSpace(cfg.ActiveProvider); activeName != "" {
+		for _, provider := range cfg.Providers {
+			if strings.TrimSpace(provider.Name) != activeName {
+				continue
+			}
+			if normalizedProfileKind(provider) == providerKind {
+				return activeName
+			}
+			break
+		}
+	}
+
+	for _, provider := range cfg.Providers {
+		if normalizedProfileKind(provider) != providerKind {
+			continue
+		}
+		if name := strings.TrimSpace(provider.Name); name != "" {
+			return name
+		}
+	}
+
+	return strings.TrimSpace(fallback)
+}
+
+func normalizedProfileKind(profile ProviderProfile) ProviderKind {
+	kind := strings.TrimSpace(string(profile.ProviderKind))
+	if kind == "" {
+		kind = strings.TrimSpace(profile.Provider)
+	}
+	return ProviderKind(strings.ToLower(kind))
 }
 
 func envValue(env map[string]string, key string) string {
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -124,24 +124,52 @@ func applyEnv(cfg *FileConfig, env map[string]string) {
 		cfg.ActiveProvider = activeProvider
 	}
 
-	apiKey := envValue(env, "OPENAI_API_KEY")
-	baseURL := strings.TrimSpace(envValue(env, "OPENAI_BASE_URL"))
-	model := strings.TrimSpace(envValue(env, "OPENAI_MODEL"))
+	applyProviderEnv(cfg, ProviderKindOpenAI, envProfile{
+		Name:    string(ProviderKindOpenAI),
+		APIKey:  envValue(env, "OPENAI_API_KEY"),
+		BaseURL: envValue(env, "OPENAI_BASE_URL"),
+		Model:   envValue(env, "OPENAI_MODEL"),
+	})
+	applyProviderEnv(cfg, ProviderKindAnthropic, envProfile{
+		Name:    string(ProviderKindAnthropic),
+		APIKey:  envValue(env, "ANTHROPIC_API_KEY"),
+		BaseURL: envValue(env, "ANTHROPIC_BASE_URL"),
+		Model:   envValue(env, "ANTHROPIC_MODEL"),
+	})
+	applyProviderEnv(cfg, ProviderKindGoogle, envProfile{
+		Name:    string(ProviderKindGoogle),
+		APIKey:  firstNonEmpty(envValue(env, "GEMINI_API_KEY"), envValue(env, "GOOGLE_API_KEY")),
+		BaseURL: firstNonEmpty(envValue(env, "GEMINI_BASE_URL"), envValue(env, "GOOGLE_BASE_URL")),
+		Model:   firstNonEmpty(envValue(env, "GEMINI_MODEL"), envValue(env, "GOOGLE_MODEL")),
+	})
+}
+
+type envProfile struct {
+	Name    string
+	APIKey  string
+	BaseURL string
+	Model   string
+}
+
+func applyProviderEnv(cfg *FileConfig, providerKind ProviderKind, env envProfile) {
+	apiKey := strings.TrimSpace(env.APIKey)
+	baseURL := strings.TrimSpace(env.BaseURL)
+	model := strings.TrimSpace(env.Model)
 	if apiKey == "" && baseURL == "" && model == "" {
 		return
 	}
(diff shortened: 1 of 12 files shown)


Vasanthdev2004 (changes requested):
Blockers
- `internal/providers/gemini/provider.go:328-331` maps Zero system messages into Gemini `systemInstruction` with `Role: "system"`, and `internal/providers/gemini/provider_test.go:83-85` locks that invalid shape in. Gemini's `Content.role` field is optional, but when present it must be either `user` or `model` (`https://ai.google.dev/api/caching#Content`). Because `systemInstruction` is also a `Content`, every Gemini request that includes a Zero system prompt will send `{ role: "system", parts: [...] }` and can be rejected by the API before the model runs. Please omit the role for `systemInstruction` (or otherwise use an API-accepted shape) and update the regression test to reject the `system` role.

Non-Blocking
- Local Go validation could not run on my machine because `go` is not installed here; I relied on GitHub CI for the Go build/test path.

Looks Good
- `bun install --frozen-lockfile` passed.
- `bun run typecheck` passed.
- `bun test ./tests --timeout 15000` passed: 285 tests, 0 failures.
- GitHub checks are green: Zero Review, Performance Smoke, and CI smoke on ubuntu/macos/windows.
- The Anthropic and Gemini adapters cover message mapping, streaming text/tool events, usage normalization, provider HTTP/stream errors, and invalid tool history before dispatch.

anandh8x (approved):
## What's good

- **The previous Gemini system-instruction blocker is fixed in current head.** `mapMessages` now builds `systemInstruction = &geminiContent{Parts: systemParts}` with no `Role`, and `TestStreamCompletionPostsGenerateContentRequest` explicitly fails if `systemInstruction["role"]` exists. That resolves the existing CHANGES_REQUESTED review against the older `role: "system"` shape.
- **Config env fallback now covers the three provider families.** `applyEnv` feeds OpenAI, Anthropic, and Google/Gemini env vars through `applyProviderEnv`, including `GEMINI_*` / `GOOGLE_*` aliases. The custom-profile case from CodeRabbit is covered too: an active `claude-prod` profile with `provider_kind: anthropic` receives `ANTHROPIC_*` values while preserving existing description metadata.
- **Provider I/O helpers remove duplicated streaming boilerplate.** `providerio.NormalizeBaseURL`, `HTTPClient`, `SendEvent`, `ScanSSEData`, `ClassifiedError`, `Redact`, and `PositiveOrDefault` centralize behavior shared by Anthropic and Gemini. The SSE scanner keeps the 16MB max-line policy from the OpenAI provider and handles multi-line `data:` blocks, comments, blank-line flushes, and `[DONE]`.
- **Anthropic mapping is faithful to Messages API shape.** System messages are joined into `system`, user/tool-result blocks are merged into user turns, assistant tool calls become `tool_use` blocks with JSON-object validation, and malformed history is rejected before dispatch. Streaming handles text deltas, partial JSON tool input, message usage, stream errors, HTTP errors, EOF cleanup of open tools, and done emission.
- **Gemini mapping handles the tricky history transforms.** System prompts map to `systemInstruction`, user/model/tool history maps to Gemini `contents`, assistant tool calls map to `functionCall`, tool results map to `functionResponse`, adjacent user/tool parts are merged, and missing Gemini function-call IDs get deterministic synthetic IDs. Tests cover request body shape, no system role, merged user turn, function call/response, top-level functionCalls, prompt-block errors, malformed streamed args, and malformed history before dispatch.
- **Factory wiring uses the model registry correctly.** Known Zero model IDs/aliases resolve to provider API model names and output ceilings (`claude-sonnet-4.5` -> `claude-sonnet-4-5-20250929`, `max_tokens: 64000`; Gemini max output 65536). Mismatched provider/model pairs fail before any HTTP request (`claude-sonnet-4.5` with explicit OpenAI errors as belongs-to-anthropic). OpenAI-compatible is still allowed for OpenAI catalog entries via `AllowsProvider`.
- **Error classification and redaction are consistent.** HTTP 401/403 -> `auth error`, 429/503/529 -> `rate limit error`, 4xx -> `provider request error`, everything else -> `provider error`; API keys and bearer tokens are redacted before events leave the provider. Anthropic and Gemini tests both assert token redaction on HTTP/stream errors.

1. **`providerio.NormalizeBaseURL` validates parseability but not scheme/host.** `url.ParseRequestURI` can accept some absolute-path-style inputs, but these are not valid provider API bases. A stricter check for `u.Scheme == "http" || u.Scheme == "https"` and `u.Host != ""` would fail bad config earlier and with a clearer error.

2. **Gemini tool results with unknown toolCallID are allowed.** If a `MessageRoleTool` has a `ToolCallID` that was not seen in prior assistant tool calls, the mapper uses the ID as the function-response name. That can generate a function response name that does not match any declared function. Consider rejecting this as malformed history, the same way missing `toolCallId` and non-object tool arguments are rejected before dispatch.

3. **Anthropic usage normalization ignores cache usage fields.** Messages API usage can include cache read/write counters. This provider only records input/output tokens, so prompt-cache-aware cost estimation from #56 will undercount until those fields are threaded into `zeroruntime.Usage.CachedInputTokens` or a follow-up cache-write field.

4. **Gemini emits every candidate.** `emitPayload` loops over all candidates and emits all text/function calls. The request doesn't ask for multiple candidates today, so this is fine, but if a future config exposes candidate count, text output from multiple candidates would interleave into one answer.

5. **Unknown models still fall back to provider-native strings.** If a model is not in the registry and the provider kind is omitted, `resolveProfile` defaults to OpenAI and sends the model string through unchanged. This preserves escape-hatch behavior for newly released provider models, but it also means a typo in a known Zero model surfaces as a provider 4xx rather than a local registry error. That tradeoff is reasonable; it should be documented.

6. **`ScanSSEData` ignores the return value of the final `flush()`.** If the final unterminated event causes `handle(data)` to return false, `ScanSSEData` still returns nil. In current callers that is harmless because the handler already emitted the terminal error and set `state.done`, but honoring the return would make the helper contract cleaner.

7. **Anthropic/Gemini stream error paths don't always emit `done`.** On provider errors they emit only `error`, which matches the OpenAI provider style from #52. Good to keep consistent, but downstream JSON consumers need to treat an error event as terminal even without a done event.

8. **`providerio.Redact` still uses whitespace-token scanning for Bearer tokens.** Same limitation noted in #52: ordinary `Authorization: Bearer sk-...` is covered, but newline-separated or punctuation-heavy bearer forms can slip. A single regex pass would be tighter once redaction is centralized.

No blockers. The current head resolves the stale Gemini systemInstruction blocker, the adapters cover the key request/stream/error/tool paths, and the factory-modelregistry integration is the right M1 provider runtime direction. The remaining issues are mostly validation and accounting polish for follow-up slices.

Vasanthdev2004 (approved):
Blockers
- None.

Non-Blocking
- Local Go validation could not run on this review host because `go` is not installed, so local `go test ./...`, `bun run build`, and `bun run smoke:build` cannot complete here. I verified the Go/build path through the green GitHub CI smoke jobs instead.

Looks Good
- The previous Gemini blocker is fixed at head `b9531c61df1bdb29663890687f438b2b14b6c366`: `systemInstruction` is now emitted without a `role`, and the request-shape test now fails if a `role` key is present.
- The follow-up also adds coverage for the custom Anthropic env/profile fallback path and preserves existing custom profile metadata while applying `ANTHROPIC_*` values.
- `bun install --frozen-lockfile` passed.
- `bun run typecheck` passed.
- `bun test ./tests --timeout 15000` passed: 285 tests, 0 failures.
- `git diff --check` passed.
- GitHub checks are green: Zero Review, Performance Smoke, CI smoke on ubuntu/macos/windows, and CodeRabbit.

PR #59, opened 2026-06-04
[codex] add Go observability backend commands

gnanam1990:
## Summary

Adds the Go-native M2 observability backend slice for the Gnanam-owned runtime track:

- Adds `internal/redaction` for reusable secret redaction across text, values, errors, URL credentials, auth headers, and common provider token shapes.
- Adds `internal/sessions` for a local append-only session event store with metadata, ordered JSONL events, latest-session lookup, safe session IDs, and lineage-preserving forks.
- Adds `internal/usage` for normalized provider usage records, model-registry-backed cost calculation, per-model summaries, and compact formatting.
- Adds `internal/search` for redacted local session-event search with persistent search indexes, session/type filters, bounded context, JSON-ready results, and text formatting.
- Adds `internal/doctor` for Go backend health checks over runtime/config/provider/model state with redacted diagnostics.
- Exposes `zero doctor` and `zero search` / `zero find` in the Go CLI with text and JSON output paths plus injected temp-store coverage in tests.

## Validation

- `go test ./...`
- `bun run typecheck`
- `bun test ./tests --timeout 15000` (290 pass)
- `bun run build`
- `bun run smoke:build`
- `bun run build:go`
- `bun run smoke:go`
- `./zero --help`
- `OPENAI_API_KEY=sk-proj-localdoctor1234567890 OPENAI_MODEL=gpt-4.1 ./zero doctor` (dummy key redacted)
- temp `XDG_DATA_HOME` probe with `./zero search needle` (secret context redacted)

Reviewers: @vasanthdev2004 @anandh8x

## Summary by CodeRabbit

* **New Features**
  * Added `doctor` CLI diagnostic command and `search`/`find` session-search commands
  * Local session search with on-disk indexing and result formatting
  * Filesystem-backed session store with event append, forking, and metadata management
  * Usage tracker with token aggregation and per-model cost summaries
  * Global redaction utility to sanitize secrets in text and structured data

* **Tests**
  * Extensive tests added for doctor, search, sessions, redaction, and usage tracker behaviors

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/Gitlawb/zero/internal/agent"
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/providers"
+	"github.com/Gitlawb/zero/internal/sessions"
 	"github.com/Gitlawb/zero/internal/tools"
 	"github.com/Gitlawb/zero/internal/tui"
 	"github.com/Gitlawb/zero/internal/zeroruntime"
@@ -17,10 +19,12 @@ import (
 var version = "dev"
 
 type appDeps struct {
-	getwd         func() (string, error)
-	resolveConfig func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
-	newProvider   func(config.ProviderProfile) (zeroruntime.Provider, error)
-	runTUI        func(context.Context, tui.Options) int
+	getwd           func() (string, error)
+	resolveConfig   func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
+	newProvider     func(config.ProviderProfile) (zeroruntime.Provider, error)
+	newSessionStore func() *sessions.Store
+	runTUI          func(context.Context, tui.Options) int
+	now             func() time.Time
 }
 
 // Run executes the minimal Go CLI surface. It returns an exit code so tests can
@@ -43,7 +47,11 @@ func defaultAppDeps() appDeps {
 		newProvider: func(profile config.ProviderProfile) (zeroruntime.Provider, error) {
 			return providers.New(profile, providers.Options{UserAgent: userAgent()})
 		},
+		newSessionStore: func() *sessions.Store {
+			return sessions.NewStore(sessions.StoreOptions{})
+		},
 		runTUI: tui.Run,
+		now:    time.Now,
 	}
 }
 
@@ -77,6 +85,10 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 		return runExec(execArgs, stdout, stderr, deps)
 	case "exec":
 		return runExec(args[1:], stdout, stderr, deps)
+	case "doctor":
+		return runDoctor(args[1:], stdout, stderr, deps)
+	case "search", "find":
+		return runSearch(args[1:], stdout, stderr, deps)
 	default:
 		if _, err := fmt.Fprintf(stderr, "unknown command %q\n", args[0]); err != nil {
 			return 1
@@ -99,8 +111,14 @@ func fillAppDeps(deps appDeps) appDeps {
 	if deps.newProvider == nil {
 		deps.newProvider = defaults.newProvider
 	}
+	if deps.newSessionStore == nil {
+		deps.newSessionStore = defaults.newSessionStore
+	}
 	if deps.runTUI == nil {
 		deps.runTUI = defaults.runTUI
+	}
+	if deps.now == nil {
+		deps.now = defaults.now
 	}
 	return deps
 }
@@ -168,6 +186,8 @@ Usage:
 
 Commands:
   exec       Run a one-shot prompt through the Go agent runtime
+  doctor     Run backend health checks for config and provider setup
+  search     Search persisted local Zero session events
   help       Show this help
   version    Print version
 
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -188,6 +188,7 @@ Commands:
   exec       Run a one-shot prompt through the Go agent runtime
   doctor     Run backend health checks for config and provider setup
   search     Search persisted local Zero session events
+  find       Alias for search
   help       Show this help
   version    Print version
 
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -206,6 +206,7 @@ func TestRunCommandsDoNotLaunchTUI(t *testing.T) {
 		{"exec"},
 		{"doctor"},
 		{"search"},
+		{"find"},
 	} {
 		t.Run(strings.Join(args, " "), func(t *testing.T) {
 			var stdout bytes.Buffer
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -204,6 +204,8 @@ func TestRunCommandsDoNotLaunchTUI(t *testing.T) {
 		{"version"},
 		{"wat"},
 		{"exec"},
+		{"doctor"},
+		{"search"},
 	} {
 		t.Run(strings.Join(args, " "), func(t *testing.T) {
 			var stdout bytes.Buffer
@@ -242,6 +244,8 @@ func assertHelpOutput(t *testing.T, args []string) {
 		"Usage:",
 		"zero [command]",
 		"exec",
+		"doctor",
+		"search",
 		"--version",
 	} {
 		if !strings.Contains(output, want) {
--- a/internal/cli/observability.go
+++ b/internal/cli/observability.go
@@ -101,7 +101,7 @@ func runSearch(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) 
 		return writeAppError(stderr, err.Error(), exitCrash)
 	}
 	if options.json {
-		if err := writePrettyJSON(stdout, result); err != nil {
+		if err := writePrettyJSON(stdout, zsearch.RedactResult(result)); err != nil {
 			return exitCrash
 		}
 		return exitSuccess
--- a/internal/cli/observability.go
+++ b/internal/cli/observability.go
@@ -1,0 +1,259 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/doctor"
+	zsearch "github.com/Gitlawb/zero/internal/search"
+	"github.com/Gitlawb/zero/internal/sessions"
+)
+
+type doctorOptions struct {
+	json         bool
+	connectivity bool
+}
+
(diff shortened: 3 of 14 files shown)


anandh8x (approved):
## What's good

- **The observability backend is cleanly split by concern.** `internal/redaction`, `internal/sessions`, `internal/search`, `internal/usage`, and `internal/doctor` each own a distinct slice, and the CLI layer in `internal/cli/observability.go` only does flag parsing, dependency injection, formatting, and exit-code routing. That keeps the new `zero doctor` / `zero search` surface thin and testable.
- **CLI dependency injection stayed consistent with #54.** `appDeps` now adds `newSessionStore` and `now`, `fillAppDeps` fills them, and tests inject temp stores / fixed clocks instead of touching the real XDG data dir. `TestRunSearchFindsSessionEventsFromInjectedStore` and `TestRunDoctorFormatsRedactedProviderDiagnostics` exercise exactly the right seams.
- **The session store has the right local shape.** Safe session IDs block traversal, metadata lives separately from append-only `events.jsonl`, `List` sorts by `UpdatedAt` descending with session-id tie-breaks, `Latest` is derived from `List`, and `Fork` preserves lineage while remapping copied events into the fork's session ID / sequence space. The JSONL parse error includes session/file/line, which will matter when users hit corrupted local state.
- **Search indexing is pragmatic and redaction-first for event payloads.** `RebuildIndex` reads session events, redacts payloads before text extraction, writes a per-session `search-index.json`, and `current()` invalidates on schema/session/update/event-count drift. Query normalization, multi-term matching, session/type filters, bounded context, and text formatting are all covered by tests.
- **Redaction is substantially more reusable than the earlier provider-local helpers.** It covers sensitive keys, common provider tokens, auth headers, URL credentials, query params, private keys, JSON string fields, assignment syntax, arbitrary values, errors with exported fields, cycles, and max depth. The test suite checks common token shapes, value cycles, and error-field redaction.
- **Doctor output is intentionally conservative.** Missing provider is a fail; missing connectivity probe is a warning; provider config/model checks are separated; details are redacted before becoming part of the report; JSON/text output share the same `Report` model. That gives later slices room to wire real connectivity without changing the CLI contract.
- **Usage tracking correctly builds on #56's model registry.** `Tracker.Record` resolves model aliases, normalizes provider usage aliases, caps cached input at input tokens, computes registry-backed cost, assigns stable in-memory sequences, and `Summary` aggregates total and per-model costs with compact formatting.
- **The new commands are wired into help and no-arg behavior safely.** `doctor`, `search`, and `find` are explicit command branches and `TestRunCommandsDoNotLaunchTUI` confirms they don't accidentally start the interactive TUI.

1. **Search JSON output is not fully redacted.** `search.RebuildIndex` redacts event payload text before indexing, and `FormatResult` redacts text-mode session metadata. But JSON mode writes the raw `search.Result` (`internal/cli/observability.go:104-108`), and each hit includes raw `sessions.Metadata` (`internal/search/search.go:40-46`, populated at `search.go:120-128`). If a session title, cwd, model id, provider string, or user-supplied session id contains a token-like value, `zero search --json` can leak it even though text output would redact it. Small fix: return/redact a JSON-safe result (`redaction.RedactValue(result, ...)`) before `writePrettyJSON`, or add a typed `RedactedResult` helper in `internal/search`.

2. **`doctor` false-fails valid provider-native/custom model names.** `providerModelCheck` calls `registry.Require(profile.Model)` and returns fail on unknown models (`internal/doctor/doctor.go:139-143`). PR #58's factory intentionally allows unknown provider-native model strings as an escape hatch, especially for OpenAI-compatible endpoints and newly released provider models. So `zero exec` can work while `zero doctor` exits provider-failure for the same config. For `openai-compatible` and explicit provider-native unknowns, this should probably be a warning ("not in Zero catalog") rather than a fail.

3. **Session append is not concurrency-safe.** `AppendEvent` reads metadata, computes `sequence := EventCount + 1`, appends JSONL, then writes metadata (`internal/sessions/store.go:281-328`). Two concurrent appends can produce duplicate sequence IDs and last writer wins metadata. For local single-process M2 this may be fine, but once TUI, exec, and background hooks all write sessions, add a per-session lock or lock-file around read/append/metadata update.

4. **Crash between JSONL append and metadata write leaves metadata stale.** Because the event is appended before metadata replacement, a process crash after the write but before `writeMetadata` leaves `events.jsonl` ahead of `metadata.eventCount`. Search index invalidation relies on `EventCount`, so stale metadata can hide events until a repair path exists. A startup/session-load reconciliation helper could count JSONL lines and fix metadata.

5. **`DefaultRoot` is Unix/XDG-only.** `sessions.DefaultRoot` honors `XDG_DATA_HOME` / `HOME` and otherwise falls back through `os.UserHomeDir`, producing `$HOME/.local/share/zero/sessions`. On Windows this isn't the platform-native data location (`%LOCALAPPDATA%` / `%APPDATA%`). The rest of the Go path is careful about Windows, so this should get a platform-aware follow-up.

6. **`zero doctor --json` can still emit plain text on config resolution errors.** If `deps.resolveConfig` fails, `runDoctor` calls `writeAppError(stderr, ...)` and exits `exitProvider` before building a JSON report. Users who call `zero doctor --json` for automation will need to handle both JSON stdout and text stderr. A JSON error envelope would make the flag reliable.

7. **Search total count is capped, not total matches.** `Sessions` sets `TotalHits = len(result.Hits)` when the limit is reached and returns immediately (`internal/search/search.go:131-134`). That's fine if the field means "returned hits", but the name reads like "all matching hits". Consider `ReturnedHits` or `HasMore` if search pagination lands.

8. **Search context uses byte offsets.** `findMatch`/`buildContext` use byte positions from lowercased strings. For non-ASCII session content, context can split a UTF-8 rune. Usually harmless in Go strings, but it can render replacement characters in terminals. Rune-aware context would be nicer once search gets user-facing polish.

9. **`redaction.RedactString` has the usual regex blind spots.** It covers a lot, but query keys are not URL-decoded (`api%5Fkey=...`), and header/query parsing is regex-based. This is good enough for a local backend, but now that redaction is centralized, adding a URL parser pass for detected URLs would tighten it.

10. **`NewTracker` panics if the default model registry fails.** The default registry is static today, so panic is practically unreachable. Still, observability code usually benefits from error returns over panics. A `NewTrackerE` helper would make tests and future dynamic registries safer.

11. **Doctor connectivity is a placeholder even when requested.** `--connectivity` still returns a warning that probing is not wired. The help says "Include provider endpoint connectivity probe when available", so this is honest enough, but a future PR should either wire a HEAD/low-cost probe or keep the flag hidden until it does something.

No blockers. The backends are well factored, tests cover the main CLI/backend paths, and redaction is consistently applied to the highest-risk event payloads. The JSON search metadata leak and doctor custom-model false fail are the two fixes I would prioritize next; both are small and don't undermine the overall slice.

Vasanthdev2004 (changes requested):
Blockers
- `zero search --json` can still leak secrets from the JSON envelope even though event payload text is redacted before indexing. `search.Sessions` returns the raw query and raw `sessions.Metadata` for each hit (`internal/search/search.go:99-128`), and JSON mode writes that result directly (`internal/cli/observability.go:103-106`). Text mode runs these fields through `RedactString`, but JSON mode does not. That means a user who searches for a token-like value, or a session whose title/cwd/provider/model metadata contains a pasted key or secret-bearing path, gets the exact raw value back in the structured diagnostic output. Since this PR explicitly exposes local session search as a diagnostic surface, the JSON result should be sanitized before serialization or converted into a redacted result type, with a regression test that includes both a secret-looking query and secret-bearing session metadata on a matching hit.

Non-Blocking
- Local Go validation could not run on this review host because `go` is not installed, so `go test ./...`, `bun run build`, and `bun run smoke:build` could not complete locally. GitHub CI smoke checks for ubuntu/macos/windows are green, so this is a local environment limitation rather than a PR failure.
- `zero doctor` currently treats unknown OpenAI-compatible/custom model names as a provider failure because `providerModelCheck` calls `registry.Require(profile.Model)` before considering the provider kind. That can make `zero doctor` fail for a custom gateway/model configuration that the runtime intentionally supports. I would downgrade unknown custom/OpenAI-compatible models to a warning while keeping official-provider mismatches strict.
- Session append still has the expected read/append/metadata race for a local M2 backend: concurrent appends can compute the same next sequence and a crash between JSONL append and metadata rewrite can leave metadata stale. This is probably acceptable before multi-process/session writers are exposed, but should be locked or reconciled before runtime hook/background writers share the store.

Looks Good
- The new Go observability slice is well separated across redaction, sessions, search, usage, doctor, and CLI wiring.
- Path/session-id traversal, corrupt JSONL, result ordering, filters, and text redaction behavior are covered in the new backend tests.
- `bun install --frozen-lockfile`, `bun run typecheck`, `bun test ./tests --timeout 15000` passed locally with 290/290 tests passing, and `git diff --check origin/main..HEAD` passed.

gnanam1990:
Addressed the valid review feedback in `3b1acc5`:

- Redacts `zero search --json` envelope fields before serialization, including secret-like queries and session metadata on hits.
- Added regression coverage for JSON query redaction and hit metadata redaction.
- Added `find` alias discoverability in help and routing coverage.
- Lowercased session error strings.
- Serialized in-process session appends per session and made generated session IDs collision-resistant under burst/fixed-clock creation.
- Added map-key extraction for searchable event payloads.
- Downgraded unknown OpenAI-compatible/custom models in `zero doctor` to a warning while keeping official-provider unknown models strict.
- Removed the dead assignment in doctor detail formatting.

Validation run locally:

- `go test ./internal/cli ./internal/search ./internal/sessions ./internal/doctor`
- `go test ./...`
- `bun run typecheck`
- `bun test ./tests --timeout 15000` (290 pass, 0 fail)
- `bun run build`
- `bun run smoke:build`
- `bun run build:go`
- `bun run smoke:go`
- `./zero --help`
- `OPENAI_API_KEY=sk-proj-localdoctor1234567890 OPENAI_MODEL=gpt-4.1 ./zero doctor`
- `XDG_DATA_HOME=$(mktemp -d) ./zero search --json sk-proj-jsonprobe1234567890`

I left the real connectivity probe as a follow-up because the current flag/help already scopes it as available-only, and wiring network probing would widen this review-fix patch.

Vasanthdev2004 (approved):
Blockers
- None.

Non-Blocking
- Local Go validation could not run on this review host because `go` is not installed, so `go test ./...`, `bun run build`, and `bun run smoke:build` could not complete locally. GitHub CI smoke checks for ubuntu/macos/windows, Performance Smoke, and Zero Review are green for the latest head.
- The Go doctor connectivity probe is still intentionally a placeholder warning when `--connectivity` is requested. That is honest in the current output/help and can land as a follow-up.

Looks Good
- The previous blocker is fixed. JSON search output now serializes `zsearch.RedactResult(result)` instead of the raw search result, and `RedactResult` redacts query, normalized query, root dir, session metadata, event summary text fields, and context before JSON output.
- Regression coverage was added for both secret-looking search queries and secret-bearing session metadata in `zero search --json` output.
- The follow-up also covers map-key search indexing, `find` alias discoverability/non-TUI routing, custom OpenAI-compatible model doctor warnings, session append serialization, and less collision-prone generated session ids.
- `bun install --frozen-lockfile`, `bun run typecheck`, `bun test ./tests --timeout 15000` passed locally with 290/290 tests passing, and `git diff --check origin/main..HEAD` passed.

PR #74, opened 2026-06-05
feat: add M5 test runner backend

gnanam1990:
## Summary
- Adds the M5 Zero test-runner backend for detecting workspace checks across Go, package-manager scripts, pytest, and Cargo projects.
- Parses runner output into structured test summaries with framework, pass/fail/skip counts, and failure details.
- Integrates those summaries into `zero verify` JSON and text output, with nested redaction for output summaries and failure details.

## Validation
- `bun install --frozen-lockfile`
- `bun run typecheck`
- `bun test ./tests --timeout 15000` (217 pass, 0 fail)
- `bun run build`
- `bun run smoke:build`
- `bun run smoke:go`
- `./zero verify --json --only go.test --timeout-ms 120000 | rg '"testSummary"|"framework"|"total"|"passed"|"failed"' -n`

Reviewers: @vasanthdev2004 @anandh8x

## Summary by CodeRabbit

* **New Features**
  * Verify now detects and parses test summaries across multiple ecosystems (Go, Node, Python, Rust, Bun) and shows structured totals and failure details in output.
* **Bug Fixes**
  * Redaction of sensitive workspace output improved and no longer mutates original results.
* **Tests**
  * Expanded test coverage for detection, parsing across runners, stable ordering, and redaction behavior.

### DIFF
--- a/internal/cli/workflow_test.go
+++ b/internal/cli/workflow_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/redaction"
+	"github.com/Gitlawb/zero/internal/testrunner"
 	"github.com/Gitlawb/zero/internal/verify"
 	"github.com/Gitlawb/zero/internal/worktrees"
 	"github.com/Gitlawb/zero/internal/zerogit"
@@ -167,6 +168,13 @@ func TestRunVerifyTextAndJSON(t *testing.T) {
 			Status:   verify.StatusPass,
 			ExitCode: 0,
 			Stdout:   "ok",
+			TestSummary: &testrunner.Summary{
+				Framework: testrunner.FrameworkGo,
+				Total:     2,
+				Passed:    1,
+				Failed:    1,
+				Failures:  []testrunner.Failure{{Name: "TestBroken"}},
+			},
 		}},
 	}
 
@@ -214,7 +222,7 @@ func TestRunVerifyTextAndJSON(t *testing.T) {
 				if decoded.Root != cwd {
 					t.Fatalf("decoded verify root = %q, want %q", decoded.Root, cwd)
 				}
-			} else if !strings.Contains(stdout.String(), "Zero verification") || !strings.Contains(stdout.String(), "go.test") || !strings.Contains(stdout.String(), cwd) {
+			} else if !strings.Contains(stdout.String(), "Zero verification") || !strings.Contains(stdout.String(), "go.test") || !strings.Contains(stdout.String(), cwd) || !strings.Contains(stdout.String(), "tests: 2 total, 1 passed, 1 failed") {
 				t.Fatalf("unexpected verify text output: %q", stdout.String())
 			}
 		})
@@ -234,6 +242,23 @@ func TestRunVerifyRedactsWorkspacePathsInOutput(t *testing.T) {
 		EndedAt:   "2026-06-05T11:05:01Z",
 		OK:        true,
 		Summary:   verify.Summary{},
+		Results: []verify.Result{{
+			ID:            "go.test",
+			Name:          "Go tests",
+			Command:       []string{"go", "test", "./..."},
+			Status:        verify.StatusFail,
+			OutputSummary: &verify.OutputSummary{Lines: []string{"failure at " + secret}},
+			TestSummary: &testrunner.Summary{
+				Framework: testrunner.FrameworkGo,
+				Total:     1,
+				Failed:    1,
+				Failures: []testrunner.Failure{{
+					Name:    secret,
+					File:    filepath.Join(secret, "secret_test.go:12"),
+					Message: "token " + secret,
+				}},
+			},
+		}},
 	}
 
 	for _, args := range [][]string{
--- a/internal/cli/workflows.go
+++ b/internal/cli/workflows.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/Gitlawb/zero/internal/redaction"
+	"github.com/Gitlawb/zero/internal/testrunner"
 	"github.com/Gitlawb/zero/internal/verify"
 	"github.com/Gitlawb/zero/internal/worktrees"
 	"github.com/Gitlawb/zero/internal/zerogit"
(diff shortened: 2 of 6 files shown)


anandh8x (approved):
## What's good

- **`internal/testrunner` is the right shared home for detection and parsing.** `Detect` returns checks in a stable, deterministic order (Go, then per-script package checks, then pytest, then cargo) and `ParseSummary` dispatches by framework. Pulling this out of `internal/verify` removes the Bun-only hardcoding in `DetectPlan` and unblocks pytest/cargo for `zero verify`.
- **Framework coverage is pragmatic and reasonably tested.** `parseGoSummary` handles verbose `--- PASS/FAIL/SKIP`, package-only `ok`, the mixed non-verbose+verbose case, and correctly avoids double-counting; `parseBunSummary`, `parseNodeSummary`, `parsePytestSummary`, and `parseCargoSummary` all sit on a shared `mergeSummaryCounts` helper so plain "N pass / N fail / N skip" summary lines contribute. Failure location extraction for Go uses the next line when present, plus a back-fill pass for the case where the location appears after a blank.
- **Redaction order is now correct.** `verify.Run` applies `redaction.RedactString` to stdout/stderr before `ParseSummary` sees them, so test summary counts and failure messages (name, file, message) operate on redacted content. `redactVerifyReport` in the CLI now copies `Results` first and copies nested slices (`OutputSummary.Lines`, `TestSummary.Failures`) before mutating, so a JSON report passed in is not mutated in place — fixes the in-place mutation concern from #59.
- **CLI formatting shows the most useful information without being noisy.** The new `formatVerifyTestSummary` produces a single "tests: N total, N passed, N failed" line (skip count when non-zero) and follows with per-failure `failure: Name at File` lines only when a name is present. JSON output gets the structured `testSummary` block.
- **Detection defaults to npm for plain Node workspaces.** `detectPackageManager` now falls through to `npm`/`FrameworkNode` when no lockfile and no declared manager are present; the prior Bun default is gone, and `TestDetectDefaultsPlainNodeWorkspacesToNPM` locks that behavior in.
- **`shouldParseTestSummary` keeps parse-on-test intact even for legacy plans.** New plans get a populated `Kind`, but for plans constructed by older code (or external callers) the kind fallbacks (id contains `.test`/`pytest`, or command contains `test`/`pytest`) are conservative and only match real test scripts.
- **Defensive copies are consistent.** `Plan.Checks` is built via `make(..., 0, len(detected))` and each `Command` slice is copied with `append([]string{}, check.Command...)` before being passed to the runner and to `ParseSummary`. The testrunner `Summary` and `Failure` slices are also handled by reference-only, with redaction handled in the CLI layer.

1. **`shouldParseTestSummary` id-substring heuristic can false-positive.** It checks `strings.Contains(id, ".test") || strings.HasSuffix(id, "test") || strings.Contains(id, "pytest")` and any command part equal to `test` or `pytest`. In current code `Detect` sets `Kind` explicitly, so this only matters for external callers. Still, ids like `bun.pretest` or a hypothetical `e2e.attest` script would be treated as a test. A `Kind != ""` short-circuit already short-circuits to `true` for `KindTest`; a `Kind != ""` short-circuit to `false` for other kinds is the missing twin. Right now a `Kind == KindTypecheck` check still falls through to the id heuristic — fine, but worth noting.

2. **`parseGoSummary` second pass has a dead guard.** `if location == nil || index == 0 { continue }` — `index == 0` cannot occur because the first pass already established that `lines` is non-empty only when the input had content, and `len(lines) > 0` is guaranteed. The guard is harmless but reads like it was guarding against a case that is no longer possible after `splitLines` filters empty lines.

3. **`parseNodeSummary` has no `Skipped` tracking in the TAP path.** `ok N - name` increments `Passed`, `not ok N - name` increments `Failed` and records the name, but `# skip N` summary lines are the only way skips enter via `mergeSummaryCounts`. A TAP test that prints `# SKIP name` (TAP "directives") is not captured, only aggregate counts are.

4. **`mergeSummaryCounts` uses `maxInt` so later, larger counts win.** That is the right choice for repeating summary lines, but it means a parser that prints both an old stale line and a new accurate one will keep the larger value. In practice summary lines appear once at the end, so this is fine; worth a comment in the helper.

5. **Framework summary regex is greedy on whitespace boundaries.** `summaryCountPrefix` matches `(\d+)\s+(pass|passes|...)`; for an input like `panic: 1 failed` the `1` and `failed` are two tokens and the regex needs the `\s+` between them — which exists, so it matches. That is correct, but the same pattern will also match a line like `fatal: saw 0 passed assertions` in some other tool's output. Not a real problem; just noting the parser is permissive on purpose.

6. **`redactVerifyReport` walks the copied `Results` by index rather than range.** Using `for index := range report.Results` and then `report.Results[index]` is fine because the loop is read-only, but a `for _, result := range report.Results` plus a small `cloneResult` helper would read more cleanly. Cosmetic.

7. **No stdout/stderr size cap before `ParseSummary`.** A pathological runner that emits megabytes of test output still gets fully redacted and parsed on the main goroutine. For local `zero verify` this is acceptable, but a future follow-up could cap the buffer fed to `ParseSummary` (e.g., last 1 MB) so a runaway test cannot stall the agent loop. The existing `OutputSummary` already truncates via `summarizeOutput` for display.

8. **Test `TestRunParsesStructuredFailureSummary` redaction assertion is on the formatted text, not the JSON `TestSummary`.** The new `if got := report.Results[0].TestSummary.Failures[0]; !strings.Contains(got.Message, "[REDACTED]")` assertion does cover the JSON, which is good. Worth a follow-up that the message stored on the `TestSummary` is itself redaction-safe even if a caller decides to re-redact via a different policy — currently it relies on `RedactString` being applied at the source.

9. **Cargo test names with spaces are split by the regex.** `cargoTestLine` matches `test (.+) \.\.\. (ok|FAILED|ignored)$`, so `test runs::zero test ... ok` becomes `Name: "runs::zero test"`, which is fine; just noting that names can include spaces and that gets recorded verbatim into the `Failure.Name` and through the CLI line.

10. **Pytest `Message` uses everything after `-`.** `pytestFailureLine` `^FAILED\s+(\S+)(?:\s+-\s*(.*))?$` records the dash and tail as `Message`. For a `FAILED tests/test_cli.py::test_stream - AssertionError: boom` it stores `AssertionError: boom`, which is what you want, but the leading ` - ` is stripped by `safeSubmatch`'s `TrimSpace` — the downstream `Failure.Message` will be `AssertionError: boom` (good).

11. **`parseGoSummary` `Total` from `summary.Total++` only on `=== RUN` lines.** For a non-verbose run that still records package status with `go test -v` disabled, `summary.Total` stays at 0 until `normalizeTotals` sets it to `Passed + Failed + Skipped`. That is correct, but means `summary.Total` reflects distinct pass/fail/skip sources; if a verbose run mixed with `ok` lines, `summary.Total` includes RUNs and `summary.Passed` is incremented per PASS plus package passes, which already-matches-a-verbose-test would not double-count because `sawVerbosePerTestOutput` is true. This is the intended invariant, but a regression test for `Total == Passed + Failed` in mixed scenarios is the one that would catch any future change.

12. **No parallel test for `TestRunSkipsStructuredSummaryForNonTestChecks` via the JSON path.** The test asserts `report.Results[0].TestSummary == nil` for a typecheck check, which is the right guard. A follow-up that feeds typecheck output with "tests:" in it (e.g., a tsc error that says "Failed 0 tests") would currently be ignored as expected. No action needed; just noting the `shouldParseTestSummary` kind-first rule is the right defense.

No blockers. The slice is well-scoped, tests cover the prior CodeRabbit nits (npm default, mixed non-verbose Go counting), the integration with `zero verify` is clean, and the nested redaction in `redactVerifyReport` is now mutation-safe. The two follow-ups I would prioritize next: (1) the `shouldParseTestSummary` id heuristic tightening once `Kind` is required for plans, and (2) a buffer cap before `ParseSummary` for long-running test output.

PR #76, opened 2026-06-05
feat: add M5 self verification backend

gnanam1990:
## Summary
- Add the M5 self-verification backend contract with bounded attempts, stop reasons, remediation metadata, and redacted remediation errors.
- Wire `zero verify --attempts` through the self-verification backend while preserving the existing single-run verifier path.
- Extend CLI JSON/text output tests plus backend tests for pass-after-retry, max-attempt stops, default attempt behavior, and remediation error redaction.

## Validation
- `bun run typecheck`
- `bun run build`
- `bun run smoke:build`
- `bun run smoke:go`
- `./zero verify --attempts 2 --json --only go.test --timeout-ms 120000`

## PRD Slice
- M5 Gnanam: self-verification loop, with reusable backend data structures for future repair/retry orchestration.

## Release Notes

* **New Features**
  * Introduced self-verification with automatic remediation support for `verify --attempts`
  * Verification now tracks stop reasons and per-attempt remediation results
  * Enhanced output formatting with detailed attempt status and remediation information

* **Tests**
  * Added comprehensive test coverage for self-verification and remediation scenarios

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Gitlawb/zero/internal/mcp"
 	"github.com/Gitlawb/zero/internal/plugins"
 	"github.com/Gitlawb/zero/internal/providers"
+	"github.com/Gitlawb/zero/internal/selfverify"
 	"github.com/Gitlawb/zero/internal/sessions"
 	"github.com/Gitlawb/zero/internal/tools"
 	"github.com/Gitlawb/zero/internal/tui"
@@ -39,7 +40,7 @@ type appDeps struct {
 	prepareWorktree  func(context.Context, worktrees.Options) (worktrees.Result, error)
 	detectVerifyPlan func(string) (verify.Plan, error)
 	runVerify        func(context.Context, verify.Plan, verify.RunOptions) verify.Report
-	runVerifyLoop    func(context.Context, verify.Plan, verify.LoopOptions) verify.LoopReport
+	runSelfVerify    func(context.Context, verify.Plan, selfverify.Options) selfverify.Report
 	inspectChanges   func(context.Context, zerogit.InspectOptions) (zerogit.ChangeSummary, error)
 	commitChanges    func(context.Context, zerogit.CommitOptions) (zerogit.CommitResult, error)
 	runTUI           func(context.Context, tui.Options) int
@@ -99,7 +100,7 @@ func defaultAppDeps() appDeps {
 		prepareWorktree:  worktrees.Prepare,
 		detectVerifyPlan: verify.DetectPlan,
 		runVerify:        verify.Run,
-		runVerifyLoop:    verify.RunLoop,
+		runSelfVerify:    selfverify.Run,
 		inspectChanges:   zerogit.Inspect,
 		commitChanges:    zerogit.Commit,
 		runTUI:           tui.Run,
@@ -218,8 +219,8 @@ func fillAppDeps(deps appDeps) appDeps {
 	if deps.runVerify == nil {
 		deps.runVerify = defaults.runVerify
 	}
-	if deps.runVerifyLoop == nil {
-		deps.runVerifyLoop = defaults.runVerifyLoop
+	if deps.runSelfVerify == nil {
+		deps.runSelfVerify = defaults.runSelfVerify
 	}
 	if deps.inspectChanges == nil {
 		deps.inspectChanges = defaults.inspectChanges
--- a/internal/cli/workflow_test.go
+++ b/internal/cli/workflow_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/redaction"
+	"github.com/Gitlawb/zero/internal/selfverify"
 	"github.com/Gitlawb/zero/internal/testrunner"
 	"github.com/Gitlawb/zero/internal/verify"
 	"github.com/Gitlawb/zero/internal/worktrees"
@@ -330,9 +331,11 @@ func TestRunVerifyReturnsProviderExitWhenChecksFail(t *testing.T) {
 func TestRunVerifyAttemptsUsesSelfVerifyLoop(t *testing.T) {
 	cwd := t.TempDir()
 	plan := verify.Plan{Root: cwd, Checks: []verify.Check{{ID: "go.test", Name: "Go tests", Command: []string{"go", "test", "./..."}}}}
-	loopReport := verify.LoopReport{
-		OK: true,
-		Attempts: []verify.Attempt{
+	loopReport := selfverify.Report{
+		Root:       cwd,
+		OK:         true,
+		StopReason: selfverify.StopReasonPassed,
+		Attempts: []selfverify.Attempt{
 			{Number: 1, Report: verify.Report{Root: cwd, OK: false, Summary: verify.Summary{Total: 1, Failed: 1}}},
 			{Number: 2, Report: verify.Report{Root: cwd, OK: true, Summary: verify.Summary{Total: 1, Passed: 1}}},
 		},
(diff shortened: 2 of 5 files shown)


anandh8x (approved):
## What's good

- **`internal/selfverify` is the right shape for the M5 backend.** Stop reasons are typed constants (`passed`, `max_attempts`, `remediator_error`, `context_canceled`), the report embeds the latest `verify.Report` per attempt plus an optional `Remediation`, and `Remediator` is a small `func(context.Context, Attempt) (Remediation, error)` that keeps the contract open for future repair strategies without coupling this slice to one.
- **Loop semantics match a conservative self-verification story.** `Run` defaults `now` to `time.Now` once, defaults `MaxAttempts` to 1 via `firstPositive`, checks `ctx.Err()` on entry and after every failed attempt, runs `verify.Run` synchronously, and only invokes the `Remediator` between attempts (the `attemptNumber >= maxAttempts` short-circuit runs before the remediator block, so a 1-attempt run never asks for a repair). `TestRunDefaultsToOneAttempt` locks that invariant in.
- **Redaction is consistent and applied at the right boundaries.** `redactRemediation` redacts the remediation fields, the `Remediator` error is run through `redact(err.Error())`, the context-cancel error is redacted, and the `StopReason` constants never carry user data. The `TestRunStopsOnRemediatorErrorAndRedacts` test exercises an `sk-proj-...` style secret through both `Remediation.Message` and the error path and confirms the combined report contains the redaction marker.
- **CLI plumbing replaces the loop seam cleanly.** `appDeps.runVerifyLoop` is removed in favor of `runSelfVerify`, `defaultAppDeps` and `fillAppDeps` carry the new function, and `runVerifyCommand` routes `options.attempts > 1` to `selfverify.Run` while the single-run path still goes through `verify.Run`. The text format adds `root:`, `stop:`, and a `remediation:` line under each attempt, and `TestRunVerifyAttemptsFormatsSelfVerifyText` covers the new output shape end-to-end.
- **`redactVerifyLoopReport` reuses `redactVerifyReport` for the per-attempt report.** That means a nested redaction walks `Results[].OutputSummary.Lines` and `TestSummary.Failures` for free on every attempt, so a remediation that surfaces a secret-laden failure message still gets redacted at the CLI boundary.
- **`Summary` reflects the most recent attempt.** After a `max_attempts` stop the user sees the latest totals, which is the most useful semantic for a self-verification loop. The CLI summary line uses `report.Summary` directly, so the text and JSON paths agree.
- **Backwards compatible on `--attempts`.** The flag, parser, and `verify.Run` path are untouched; only the routing changes when `attempts > 1`. The single-attempt behavior is preserved.

1. **`redactVerifyLoopReport` mutates the caller's `Attempts` slice.** It copies `report.Root` and `report.Error` and redacts each `Attempt.Report` in place, but the `Attempts` slice header is reused. `redactVerifyReport` (introduced in #74) copies `Results` first, so the new loop report is inconsistent with that pattern. Easy fix: `attempts := append([]Attempt{}, report.Attempts...); report.Attempts = attempts` at the top of the function, then walk by index. This will also defend against future fields that hold slices.

2. **No test for `StopReasonContextCanceled`.** The code path is exercised in the `Run` body (entry and post-attempt checks), but no test cancels a context and asserts the report comes back with `StopReasonContextCanceled`, a redacted `Error`, and that no remediator runs. A small test using a `context.WithCancel` plus a failing `Runner` would lock that branch in.

3. **`Summary` is not accumulated across attempts.** The slice only stores the last attempt's `verify.Summary`. For M5 that's the right call (current state is the most useful), but if a future PR wants "over N attempts, M passed, K failed" it will need to add an `Accumulated` field or a separate `Aggregated` summary. Worth noting in the design notes so the next slice doesn't conflate them.

4. **The `--attempts` loop is currently a no-op in production.** The CLI does not inject a `Remediator`, so passing `--attempts 2` runs the same `verify.Run` twice with no repair in between. That is consistent with this PR's scope (backend + data model, "for future repair/retry orchestration"), but it means a user running `zero verify --attempts 3` today will get the same result as three separate single-shot runs. A follow-up should either (a) wire a real `Remediator` driven by the agent runtime, or (b) document the loop as "retry the same plan N times until it passes." Either is fine; the ambiguity is what the next slice needs to resolve.

5. **`now()` is captured as a local, but `runOptions.Now` is mutated.** The line `runOptions.Now = now` runs after defaulting and the local `runOptions` is what is passed to `verify.Run`. Since `Options` is a value receiver in `Run`, the caller's `Options.RunOptions` is not affected. This is safe, but the mutation of a struct value (rather than just using a local `now` function) reads as a side effect. Cosmetic.

6. **`Remediator` is a function value, not an interface.** That keeps the slice minimal and avoids interface-implementation ceremony, but it means a future `Remediator` that needs to share state (e.g., a worktree handle, a fix-up logger) will have to capture it via closure. Fine for now; if a second consumer appears, an interface with a `Name()`/`Plan()` is probably the next step.

7. **`formatRemediation` joins `applied`/`not applied`, optional `Message`, and optional `error: ...` with `" - "`.** The separator is fine, but a remediation with both `Message` and `Error` reads as `applied - message - error: err`. If `Message` is itself multi-line, the join collapses newlines. For an M5 backend that is acceptable; if remediation messages become free-form logs, switch to a multi-line block.

8. **`Report.Root` is now redundant with the embedded `Attempt[0].Report.Root`.** Both are populated from `plan.Root`. Removing the field would deduplicate the schema, but the JSON top-level `root` is convenient for scripts that just want the workspace. Worth keeping.

9. **`Run` returns `StopReasonMaxAttempts` when the loop body never runs.** With `firstPositive` defaulting to 1 the loop always runs at least once, so this cannot happen in practice. Still, a `MaxAttempts == 0` would produce a report with zero attempts and `StopReasonMaxAttempts`, which is a small contract surprise. Either reject `MaxAttempts < 0` explicitly or change the default to `StopReasonPassed` with zero attempts (less surprising).

10. **`StopReason` is `string`-typed but no `IsTerminal()` helper.** Consumers of `selfverify.Report` will likely write `switch report.StopReason { case StopReasonPassed: ... }` themselves. A small `func (s StopReason) IsTerminal() bool` would centralize that, but it is premature for one slice.

11. **JSON `omitempty` is missing on `StartedAt`/`EndedAt`.** They are always populated by `Run` (the loop body fills `EndedAt` and `formatTime` returns the empty string for zero time, which would still serialize as `""`). Not a real leak, just a schema nicety: `omitempty` would let callers distinguish "in progress" from "zero value".

12. **No JSON-schema or OpenAPI doc update.** The new top-level fields (`root`, `stopReason`, `summary`, `attempts[].remediation`) will appear in `zero verify --attempts --json` output. If a downstream consumer doc exists, a follow-up should mention the new shape.

13. **Test `TestRunStopsOnRemediatorErrorAndRedacts` only asserts the secret does not appear.** A stronger test would also assert that `report.Error == remediation.Error` (the loop copies it on the error path) and that `report.StopReason` is `StopReasonRemediatorError`. The test already checks `StopReason`; the `Error == remediation.Error` invariant is implicit but worth pinning down.

14. **`fakeVerifyRunner.Run` has no return error check.** A real `verify.Runner` signature can return an error; the fake always returns nil. That is fine for these tests, but if a future slice adds an error path to the runner, the fake will need to learn to inject errors. Not a regression in this PR.

15. **No parallelism knob.** Attempts run strictly serially. For M5 that is the right call (deterministic, easy to reason about), and `verify.Run` itself runs each check in series. A future slice could add a `--parallel` flag, but it is not part of this scope.

No blockers. The slice is well-scoped, the data model is forward-compatible with a real `Remediator`, the tests cover the four primary paths (pass-after-retry, max-attempt stop, single-attempt default, remediator error + redaction), and the CLI integration is consistent with #74. The two follow-ups I would prioritize next: (1) add a context-cancellation test for `StopReasonContextCanceled`, and (2) decide whether the loop is a real self-verification story (wire a `Remediator`) or explicitly a no-op retry loop (document it as such).

PR #77, opened 2026-06-05
feat: add M6 sandbox backend

gnanam1990:
## Summary
- Add the M6 sandbox backend package with a shared policy engine, risk classifier, structured violation errors, persistent allow/deny grants, and Linux/macOS/policy-only backend selection.
- Enforce sandbox decisions through the Go tool registry and agent runtime so unsafe/high-autonomy runs still receive structured policy denials for out-of-workspace, symlink, network, and destructive shell risks.
- Add `zero sandbox policy` and `zero sandbox grants` backend commands for policy inspection plus persistent grant allow/deny/list/revoke/clear workflows.

## Scope
- PRD slice: M6 Gnanam sandbox backend.
- This does not implement Windows sandboxing or TUI permission prompts; it provides the shared backend contracts those later slices can consume.

## Validation
- `bun run typecheck`
- `bun run build`
- `bun run smoke:build`
- `bun run smoke:go`
- `ZERO_SANDBOX_GRANTS_PATH=$(mktemp -d)/grants.json ./zero sandbox policy --json`
- `ZERO_SANDBOX_GRANTS_PATH=$(mktemp -d)/grants.json ./zero sandbox grants allow write_file --auto medium --reason final-local-check --json`
- `ZERO_SANDBOX_GRANTS_PATH=$(mktemp -d)/grants.json ./zero sandbox grants list --json`
- `ZERO_SANDBOX_GRANTS_PATH=$(mktemp -d)/grants.json ./zero sandbox grants clear --confirm --json`

## Size
- 20 files changed, about 2.2k inserted lines, split into 3 focused commits.

## Summary by CodeRabbit

* **New Features**
  * Added sandbox functionality to restrict and control tool execution (workspace path enforcement, network/destructive-shell blocking).
  * Added persistent grant store to save tool approvals and autonomy levels.
  * Introduced `zero sandbox` CLI with `policy` (view policy/backend) and `grants` (allow/deny/list/revoke/clear) commands and JSON/text output options.

* **Tests**
  * Added extensive tests covering sandbox policy evaluation, path/symlink protections, backend selection, CLI flows, and grant persistence.


anandh8x (approved):
## What's good

- **The `internal/sandbox` package is well-factored.** `types.go` is pure data (typed strings for every classification, plus `Policy`, `Request`, `Decision`, `Risk`, `Violation`, `Grant`). `normalize.go` is a small input-validation layer. `risk.go` does classification. `paths.go` does workspace enforcement. `grants.go` does persistence. `engine.go` is the policy evaluator. `adapters.go` is platform selection. Each file is single-purpose and easy to test in isolation, and the package depends only on `redaction` from earlier slices.
- **The engine is conservative by default and degrades predictably.** `DefaultPolicy` is `Mode: enforce`, `Network: deny`, `EnforceWorkspace: true`, `DenyDestructiveShell: true`, with a policy-only backend on platforms that do not have `bwrap` or `sandbox-exec`. `SelectBackend` reports `Available: false` and a `Message` for the fallback case, and `BuildPlan` always documents the active restrictions, including "platform sandbox unavailable; policy engine still evaluates tool requests". Users running on Windows or a stripped-down Linux image get a clear signal that the OS-level sandbox is not engaged.
- **Workspace enforcement is symlink-aware and consistent with the read/write tools.** `validateWorkspacePaths` resolves the workspace root via `filepath.EvalSymlinks`, joins and `Abs`-normalizes the target, and then walks the relative path segment-by-segment, rejecting any `Lstat` symlink whose resolved target escapes the root. The same `outside_workspace` and `symlink_traversal` violations are raised whether the path is absolute, relative, or traverses a pre-existing symlink. This matches the gate logic in the existing `write_file` / `read_file` tools and gives the agent runtime defense-in-depth.
- **Risk classification is a separate concern from policy.** `Classify` reads `SideEffect`, scans `command` for `networkCommandPattern`, `destructiveCommandPattern`, and piped installers (`| sh`, `| bash`), and flags `path_escape` for `..` and absolute paths. Categories are sorted for stable JSON. The engine then uses `HasRiskCategory` to decide which policy clauses apply. This split means a future slice can add a new risk category without touching the engine.
- **Persistent grants are first-class and well-protected.** `GrantStore` validates tool names against `^[A-Za-z0-9._-]+$` (rejecting empty and `../escape`), normalizes decision and autonomy, redacts `Reason` on both write and load, persists with atomic temp+rename under a 0o600 file in a 0o700 directory, and protects writers with a `sync.Mutex`. Schema versioning is in place (`schemaVersion: 1` with explicit rejection of other versions), so a v2 migration is one branch away. `Lookup` matches grants only when `requestedAutonomy <= grant.MaxAutonomy` (low/medium/high), so a `medium` grant does not satisfy a `high` request.
- **Sandbox is enforced in the right places.** `tools.Registry.RunWithOptions` evaluates the sandbox before the tool runs and turns `ActionDeny` into a structured error result (`"Sandbox violation [code] for tool at path: reason"`). `ActionPrompt` without `PermissionGranted` returns a structured "Sandbox approval required" error, so the agent sees a clear next step. The agent runtime forwards `Autonomy` and `Sandbox` to the registry, and `TestRunAppliesSandboxEvenInUnsafeMode` locks in that the sandbox still applies when `PermissionModeUnsafe` is set, which is exactly the M6 promise.
- **CLI surface is consistent with the rest of the codebase.** `zero sandbox policy` prints either text (`Zero sandbox policy / root / mode / network / backend / backend_available / grants`) or JSON with `{policy, backend, plan, grantsPath}`. `zero sandbox grants {list,allow,deny,revoke,clear}` mirrors the MCP grants surface from #40 and uses the same `--json` and `--confirm` patterns. `--confirm` is required for `clear`, and `TestRunSandboxHelpDoesNotOpenStore` proves that help paths do not touch the store. `nextFlagValue` is reused for `--auto` and `--reason`, so the parser is consistent with `zero exec` and does not regress the flag-shaped-value gap from #54.
- **Redaction is applied at the persistence boundary.** `Grant.Reason` is redacted in `createGrant` (on the way in) and again in `normalizeStoredGrant` (on the way out), so a secret reason saved by a user or hand-edited in the JSON file is rewritten as `[REDACTED]` on next load. The grants JSON file therefore never stores raw secrets even if a user pastes one in.
- **Backend selection is injectable.** `SelectBackend` takes `BackendOptions{GOOS, LookupExecutable}` with sensible defaults (`runtime.GOOS`, `exec.LookPath`), and `appDeps.selectSandboxBackend` lets tests pin the backend to `policy-only`. `TestRunSandboxPolicyInspectTextAndJSON` exercises both text and JSON without depending on the host's `bwrap`/`sandbox-exec` install, which keeps CI portable.
- **Context cancellation is handled explicitly.** `Engine.Evaluate` checks `ctx.Err()` at entry and returns `ViolationContextCanceled` with a redacted reason, and `TestEngineReportsContextCancellation` exercises it. The grant store's `Lookup` path is not in the cancellation critical section (it is a fast read), which is appropriate.

1. **`Lookup` is not mutex-protected.** `Grant`/`Revoke`/`Clear` hold the mutex while reading, mutating, and rewriting the file, but `Lookup` reads the file outside the lock. For a single-process CLI this is fine, but a concurrent `Grant` + `Lookup` can race: `Lookup` may see the pre-grant state and return `Matched: false`, or `Grant` may read pre-revoke state and rewrite an entry the caller intended to remove. If `internal/agent` ever calls `Lookup` from a background goroutine, prefer `sync.RWMutex` with a cached `grantFile` snapshot, or move the read inside the mutex.

2. **TOCTOU between sandbox check and tool execution.** `Engine.Evaluate` checks the path, but the actual write happens in `internal/tools/write_file.go` after the registry returns. A pre-created symlink placed between the two checks (or created by a parallel process) could still escape the workspace. The existing write tool's per-segment `Lstat` rejection is the real defense; the sandbox check is a fast pre-filter. Worth a comment on the engine noting that the write tool is the source of truth for the file boundary.

3. **Risk regex is best-effort.** `networkCommandPattern` covers `curl`, `wget`, `scp`, `ssh`, `rsync`, `nc`/`netcat`, `python -m http.server`, `npm/pnpm/yarn/bun add|install|publish|login`, `pip install`, `go get`, `git clone`, `gh release download`. A payload like `python3 -c "import urllib.request, os; os.system('curl ...')"` does not match, and neither does `bash -c 'curl ...'`. The shell tool is `SideEffectShell` so the engine still classifies it as `RiskHigh`; users in unsafe mode opt into this. A future slice should consider a stronger detection story (e.g., a shell-aware AST or a vetted allowlist of commands).

4. **Piped-installer detection uses `strings.Contains`.** `| sh` and `| bash` are caught, but `> sh`, `exec sh`, backticks, or `eval "curl"` are not. Again, classified as `RiskHigh`; the user explicitly approves by being in unsafe mode. The detection is a hint, not a guarantee.

5. **`RiskReason` always emits the level, but the level is duplicated with `Decision.Risk.Level`.** A consumer that prints a JSON decision already has both fields. Keeping `RiskReason` is fine for text output, but a future refactor could derive it at format time and drop the field.

6. **Engine does not use the platform backend at runtime.** `SelectBackend` and `BuildPlan` are informational; the engine does not actually wrap `bash` invocations in `bwrap`/`sandbox-exec`. The M6 PR description explicitly says this is the backend contract and that platform integration comes later. The current `BackendPlan` is a good debug surface for `zero sandbox policy`, but the engine should be revisited when a future slice actually wires the platform sandbox.

7. **`AllowPolicyOnlyRunner` is in `Policy` but unused by the engine.** The default is `true` and there is no engine branch that reads it. Either it is a future knob (e.g., fail closed when the platform sandbox is unavailable and the policy says "no") or it should be removed. Keeping it is fine if a follow-up PR will reference it, but the current engine does not enforce it.

8. **`EngineOptions.Backend` has no default for the zero value.** `NewEngine` accepts a `Backend` value, and `buildExecSandboxEngine` always passes a result from `SelectBackend`, but a test that constructs `NewEngine(EngineOptions{WorkspaceRoot: ...})` directly will see `Backend.Name == ""`. The engine does not look at `Backend` for evaluation, so this is harmless today, but worth a `defaultBackend()` helper for future engine branches that do.

9. **`runExec` builds a new `GrantStore` on every invocation via `deps.newSandboxStore`.** That is fine (the file is small, the reads are infrequent), but if a future slice needs to cache a per-process snapshot, the store has no `Snapshot()` helper. A small addition would be a `Lookup` that takes a snapshot to avoid file I/O on every tool call.

10. **`runExec` does not pass the per-exec autonomy to the engine.** The agent runtime reads `Options.Autonomy` and forwards it on every `RunWithOptions` call, so the engine does see it. The split between `buildExecSandboxEngine` (no autonomy) and `agent.Run` (autonomy per request) is correct, but a reader of `cli/exec.go` may wonder why autonomy is not threaded at engine construction. A small comment would help.

11. **`runWithDeps` does not add a `sandbox` test to the no-TUI list at `app_test.go`.** It does (`TestRunCommandsDoNotLaunchTUI` includes `{"sandbox"}`), and `assertHelpOutput` includes `"sandbox"`. The follow-up is just noting the test coverage is consistent with the other commands.

12. **`TestRunSandboxGrantsAllowListDenyRevokeAndClear` asserts `len(listPayload.Grants) != 2` and a specific sort order.** The `Store.List` helper sorts tool names alphabetically, so `bash` comes before `write_file`. The test relies on that ordering. A test that adds `apt` and `cat` after `bash` would confirm the sort is stable across more than two entries. Minor.

13. **`parseSandboxPositionalOptions` defaults autonomy to `AutonomyLow` if `--auto` is not provided.** The CLI then sends a `Low`-autonomy request to the store, which means a tool that was granted at `Medium` will not be matched for a `Low` request... wait, `autonomyAllowed` uses `requested <= max`, so a `Low` request matches a `Medium` grant. Good. But if a tool was granted at `Low` only, a `Medium` request would not match — that is the intended semantic. Worth documenting on the grant help text.

14. **`runSandboxGrants` uses `parseSandboxCommandOptions` for the `list` subcommand but the parser disallows any positional argument.** If a user runs `zero sandbox grants list extra`, they get a usage error. That is the right behavior; the test does not cover it, but the production code rejects the input. No action needed.

15. **`BuildPlan` returns a flat `[]string` of restrictions.** The text format prints each on a separate line, but a JSON consumer sees a list of free-form strings. A typed `Restriction` enum (`filesystem`, `network`, `destructive`, `platform_unavailable`) would be more discoverable for tools that consume the JSON. Cosmetic; can wait for a real consumer.

16. **`Engine.Evaluate` always overrides `request.WorkspaceRoot` with `engine.workspaceRoot` via `firstNonEmpty`.** That is the right precedence (engine-level wins), but a per-request override is silently ignored. If a future slice wants to evaluate a tool against a sub-workspace (e.g., a worktree), the precedence needs an explicit `Request.WorkspaceRoot` field that the engine respects.

17. **`runExec` does not currently pass `--auto` to the sandbox at construction time.** A user running `zero exec --auto high` would set `options.autonomy = "high"` (per earlier exec wiring), which is forwarded to `agent.Options.Autonomy` and then to the registry. The path is correct; the gap is that the sandbox engine itself is not parameterized by autonomy, so a future "autonomy ceiling" knob would need to be added to `EngineOptions`.

18. **The `WriteFile` tool is exercised by the sandbox test, but `ReadFile`, `Bash`, `UpdatePlan`, etc. are not.** The engine tests cover all of these via direct `Request` construction, which is the right level for engine coverage. A future PR that registers more tools should add a `TestRegistryAppliesSandboxFor<EveryTool>` matrix test if the per-tool safety metadata diverges from the engine's risk classification.

19. **`Engine.Evaluate` returns `Decision{Action: ActionAllow, Risk: ..., Reason: "sandbox policy disabled"}` when `policy.Mode == ModeDisabled`.** That is correct, but the `Risk` field is still computed. Consumers that print `Decision.Risk` for a disabled sandbox will see a populated risk object even though the decision is unconditional allow. Either drop `Risk` in the disabled branch or document the invariant.

20. **`grantSchemaVersion = 1` is a private constant.** If a future migration script wants to introspect the version, it has to read the file. A small exported `CurrentGrantSchemaVersion` would help tooling. Minor.

No blockers. The M6 backend contract is in place: structured policy engine, risk classifier, persistent allow/deny grants, Linux/macOS/policy-only backend selection, enforcement in the tool registry and agent runtime, and a consistent `zero sandbox {policy,grants ...}` CLI surface. Tests cover workspace symlinks, out-of-workspace paths, network/destructive shell risk, persistent allow/deny matching, autonomy rank filtering, schema-version rejection, malformed-file rejection, atomic write durability (via reopen), context cancellation, and the unsafe-mode invariant. The two follow-ups I would prioritize next: (1) make `Lookup` concurrency-safe (RWMutex + cached snapshot) before the agent runtime calls it from background goroutines, and (2) document or remove `AllowPolicyOnlyRunner` (currently unused) so the policy struct matches what the engine actually enforces.

Vasanthdev2004 (changes requested):
### Blockers

- Persistent sandbox grants do not currently authorize prompt-gated core tools through the actual registry/agent execution path. `internal/tools/registry.go` returns at the prompt-permission check before it calls `options.Sandbox.Evaluate`, so a saved `zero sandbox grants allow write_file --auto medium` is persisted and listed, but `Registry.RunWithOptions(... Sandbox: engine, PermissionGranted: false ...)` still fails with `Permission required for write_file`. That makes the new `zero sandbox grants allow` workflow ineffective for `write_file`, `edit_file`, `apply_patch`, and `bash`, and leaves the M6 backend contract different from the CLI/API promise. I confirmed this with a local probe test that creates a grant store, grants `write_file`, routes through `Registry.RunWithOptions`, and still gets the prompt-denial error before sandbox grant lookup.

### Non-Blocking

- None.

### Looks Good

- The sandbox package has a clean split between policy evaluation, risk classification, workspace/symlink checks, backend selection, and grant persistence.
- CLI coverage for `zero sandbox policy` / `zero sandbox grants` is solid, and the persisted grant file uses schema validation, sorted output, redacted reasons, and 0600 writes.
- Validation passed locally: targeted Go packages, full `go test -count=1 -p 1 ./...`, `bun run typecheck`, 212/212 Bun tests, build, smoke build, smoke Go, sandbox CLI JSON/list/clear probes, and `git diff --check`. PR base is current with `origin/main`; no extra stale-base merge was needed.

---

**Verdict: Changes Requested** — clean structure and tests, but the persistent grant path needs to authorize prompt-gated tools before this can land.

Vasanthdev2004:
### Blockers

None found. The prior sandbox grant blocker is fixed at `e78bdf1833f345d9bb61e57864b11fb4233c5529`: `Registry.RunWithOptions` now evaluates sandbox decisions before the prompt-permission fallback and treats a matched persistent grant as authorization for prompt-gated tools.

### Non-Blocking

None.

### Looks Good

- Added a regression test that exercises the real registry path: persistent `write_file` grants now authorize execution without `PermissionGranted`.
- The incremental fix is focused: registry ordering, grant-store read locking, policy default handling, and a small shell-risk case-insensitivity hardening.
- Validation passed locally: targeted Go packages, full `go test -count=1 -p 1 ./...`, TypeScript typecheck, 212/212 Bun tests, build, smoke:build, smoke:go, whitespace check, and sandbox grant CLI smokes.
- GitHub checks are green for Ubuntu, macOS, Windows, performance smoke, Zero Review, and CodeRabbit.

Verdict: Approve — clean fix for the authorization-path blocker; ready from my side.

Vasanthdev2004 (approved):
Approved after rereview: prior sandbox grant authorization blocker is fixed and validation is green.

PR #89, opened 2026-06-06
feat(zerocommands): add typed SandboxGrantSnapshot contract

gnanam1990:
## Summary

The merged permission and verification event contracts added typed snapshots for config, providers, models, and sessions, but the persistent sandbox grant store (`internal/sandbox/grants.go`) still exposes only `sandbox.Grant` directly. The TUI render path, the headless `zero sandbox` command, and PR/CI automation all need a typed snapshot so they do not hand-roll `map[string]any` payloads and do not accidentally leak grant reasons that contain API keys or tokens.

This commit adds `SandboxGrantSnapshot` alongside the existing `ProviderSnapshot`, `ModelSnapshot`, `SessionSnapshot`, and `SessionTreeSnapshot` in `internal/zerocommands`. The snapshot mirrors the format of the existing typed contracts:

- `ToolName`, `Decision`, `MaxAutonomy`, `ApprovedAt` are copied verbatim from the underlying grant. They are non-secret metadata.
- `Reason` is run through the standard redaction pipeline before it is copied, so any secret material the user typed when granting is masked before the snapshot leaves the runtime.
- All string fields are trimmed. An empty or whitespace-only `Reason` becomes an empty `Reason` in the snapshot so JSON output omits the field.

Two constructors are provided:

- `SandboxGrantSnapshotFromGrant(grant)` for a single grant.
- `SandboxGrantSnapshots(grants)` for a slice. Output is sorted alphabetically by `ToolName` so consumers (TUI, headless, JSON output) see a deterministic ordering. An empty input returns a non-nil empty slice so JSON output is always `[]` and never `null`.

A matched/unmatched helper is also added:

- `SandboxGrantMatchSnapshotFromLookup(toolName, lookup)` pairs the typed snapshot with a `Matched` boolean. When the lookup did not match, the returned snapshot has `Matched=false` and a nil `Grant` pointer, so consumers can render the absence without inspecting an error or a typed zero value.

## Tests

- `TestSandboxGrantSnapshotFromGrantRedactsReasonAndTrimsFields`: verifies the secret in the reason is masked and the surrounding fields are trimmed.
- `TestSandboxGrantSnapshotFromGrantEmptyReasonBecomesEmpty`: verifies whitespace-only reasons normalize to the empty string.
- `TestSandboxGrantSnapshotsSortsByToolNameAndReturnsEmptySliceForEmptyInput`: verifies alphabetical sort and that nil input returns a non-nil empty slice whose JSON encoding is `[]`.
- `TestSandboxGrantMatchSnapshotFromLookupMatchedAndUnmatched`: verifies the matched and unmatched code paths.
- `TestSandboxGrantSnapshotJSONShapeIsStable`: verifies every JSON key is present so downstream consumers can rely on the shape.

## Verification

- `go build ./...`
- `go vet ./...`
- `go test -count=1 -p 1 ./internal/zerocommands/... ./internal/sandbox/... ./internal/redaction/...`

## DRI

Gnanam (runtime core owner per `docs/WORK_SPLIT_PRD.md` §4). Connects `zero sandbox` command rendering, the TUI grants panel, and PR/CI automation consumers to the persistent grant store without leaking grant reasons.

## Out of scope

This PR only adds the typed contract and the constructors. The TUI render path, the `zero sandbox` headless command, and the PR/CI automation consumer are follow-up work in their respective owners' lanes.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for sandbox grant snapshot behavior, including field trimming, reason redaction, sorting, and JSON serialization.

* **Chores**
  * Added internal snapshot export helper functions and data structures.

### DIFF
--- a/internal/zerocommands/sandbox_snapshot.go
+++ b/internal/zerocommands/sandbox_snapshot.go
@@ -1,0 +1,86 @@
+package zerocommands
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/redaction"
+	"github.com/Gitlawb/zero/internal/sandbox"
+)
+
+// SandboxGrantSnapshot is the typed view of a single persistent
+// sandbox grant as it is exposed to TUI, headless, and PR/CI
+// automation. The snapshot guarantees that no secret material from
+// the grant's Reason field is leaked: the field is always run
+// through the standard redaction pipeline before it is copied into
+// the snapshot. ToolName, Decision, MaxAutonomy, and ApprovedAt
+// are non-secret metadata and are copied verbatim.
+type SandboxGrantSnapshot struct {
+	ToolName    string `json:"toolName"`
+	Decision    string `json:"decision"`
+	MaxAutonomy string `json:"maxAutonomy"`
+	ApprovedAt  string `json:"approvedAt,omitempty"`
+	Reason      string `json:"reason,omitempty"`
+}
+
+// SandboxGrantMatchSnapshot is the typed view returned to consumers
+// who look up a grant for a specific tool call. It pairs the
+// underlying grant snapshot with a Matched flag so callers can tell
+// the difference between "grant existed and matched" and "no grant
+// recorded for this tool" without inspecting an error or a typed
+// zero value.
+type SandboxGrantMatchSnapshot struct {
+	ToolName string                `json:"toolName"`
+	Matched  bool                  `json:"matched"`
+	Grant    *SandboxGrantSnapshot `json:"grant,omitempty"`
+}
+
+// SandboxGrantSnapshotFromGrant converts a sandbox.Grant into its
+// typed snapshot. The Reason field is run through redaction so any
+// secret material is masked before the snapshot leaves the runtime.
+// An empty or whitespace-only Reason becomes an empty Reason in the
+// snapshot, which keeps the JSON shape stable for downstream
+// consumers.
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- The PR head is based behind current `main`, but I test-merged `4af8faaf12151180c109e43feed2c56adca3d6b9` into latest `origin/main` (`ce8dec087d7db354f49ffa7e9572b3b7a7d2ab19`) and it merged cleanly with the same validation passing.

### Looks Good

- The new `SandboxGrantSnapshot` contract trims fields and runs grant reasons through the shared redaction path before exposing them to downstream JSON/TUI/automation consumers.
- Slice conversion returns a non-nil empty slice and sorts snapshots by tool name, so consumers get deterministic `[]` output instead of `null`.
- The lookup snapshot cleanly separates matched vs. unmatched results with a nil grant pointer for misses, avoiding ambiguous zero-value grant handling.
- Local validation passed: `go build ./...`, `go vet ./...`, targeted `go test -count=1 -p 1 ./internal/zerocommands/... ./internal/sandbox/... ./internal/redaction/...`, full `go test -count=1 ./...`, `git diff --check`, and a runtime redaction probe confirming an assembled OpenAI-shaped key in `Reason` is emitted as `[REDACTED]`.
- GitHub checks are green on the reviewed head.

Vasanthdev2004 (approved):

PR #90, opened 2026-06-06
feat(zerocommands): add typed MCP, hooks, and plugins snapshots

gnanam1990:
## Summary

The merged permission and verification event contracts added typed snapshots for config, providers, models, sessions, and the sandbox grant store, but the MCP server config, hooks config, and loaded plugin manifests still expose only the raw backend structs. The TUI render path, the headless `zero mcp`, `zero hooks`, and `zero plugins` commands, and PR/CI automation all need typed snapshots so they do not hand-roll `map[string]any` payloads and do not accidentally leak the secret material that lives in MCP server Env and Headers.

This commit adds three typed snapshot constructors plus a bundled backend lifecycle snapshot to `internal/zerocommands`.

### MCPServerSnapshot

- `name`, `type`, `identity`, `url`, `command` are copied verbatim and trimmed. They are non-secret metadata the operator needs to see when triaging an MCP failure.
- `argCount`, `envKeyCount`, `headerCount` replace the slices and maps with counts. Secret material in `Env` and `Headers` is never serialized.
- `toolCount`, `allowGranted`, `denyGranted` are optional runtime counts that the snapshot can carry when the caller has a live tool registry and a live permission store. A nil counts struct leaves the snapshot at zero.

### HookSnapshot

- `id`, `name`, `event`, `matcher`, `command`, `args`, `enabled`, `source`.
- `command` and `args` are run through the standard redaction pipeline. Whitespace-only args are dropped so JSON output is tight.

### PluginSnapshot

- `id`, `name`, `version`, `description`, `enabled`, `source`, `root`, `pluginDir`, `manifestPath` are copied verbatim and trimmed so the operator can see where the manifest came from.
- `toolCount`, `promptCount`, `skillCount`, `hookCount` replace the full extension slices with counts so the headless JSON output stays small.

### BackendLifecycleSnapshot

- Bundles the three slices into a single payload so `zero doctor` and `zero config` can return one typed result that describes the full extensibility surface.
- All three inner slices are always non-nil, so JSON output is always `[]` and never `null`.

## Tests

- `TestMCPServerSnapshotFromServerStripsSecretsAndCountsMaps`: verifies env and header values never appear in the JSON output, and that the count fields record the maps' sizes.
- `TestMCPServerSnapshotWithCountsMergesRuntimeCounts`: verifies the optional counts bundle is merged and that a nil bundle leaves the snapshot at zero.
- `TestMCPServerSnapshotsSortsAndReturnsEmptySliceForEmptyInput`: verifies alphabetical sort and that nil input returns a non-nil empty slice whose JSON encoding is `[]`.
- `TestHookSnapshotFromDefinitionRedactsCommandAndTrimsFields`: verifies command and args are redacted, and the surrounding fields are trimmed.
- `TestHookSnapshotsSortsByIDAndEvent`: verifies the deterministic (id, event) ordering.
- `TestHookSnapshotsWithSourceTagsEverySnapshot`: verifies the helper that tags every snapshot with the same source string.
- `TestPluginSnapshotFromPluginCollapsesSlicesToCounts`: verifies the path fields are preserved verbatim and the count fields record the slices' sizes.
- `TestPluginSnapshotsSortsByIDAndReturnsEmptySliceForEmptyInput`: verifies id-based sort and empty-slice stability.
- `TestNewBackendLifecycleSnapshotBundlesAndDefaultsNilsToEmpty`: verifies the bundled payload, that nil inputs produce empty slices in the JSON, and that a populated input round-trips.

## Verification

- `go build ./...`
- `go vet ./...`
- `go test -count=1 -p 1 ./internal/zerocommands/... ./internal/mcp/... ./internal/hooks/... ./internal/plugins/... ./internal/redaction/...`

## DRI

Gnanam (runtime core owner per `docs/WORK_SPLIT_PRD.md` §4). Closes the typed-snapshot gap for the three backend surfaces in §7.E that the TUI render path, the headless commands, and PR/CI automation consume.

## Out of scope

This PR only adds the typed contracts and the constructors. The TUI render path, the `zero mcp` / `zero hooks` / `zero plugins` headless commands, and the PR/CI automation consumer are follow-up work in their respective owners' lanes.

## Summary by CodeRabbit

* **New Features**
  * Structured backend snapshots for servers, hooks, and plugins with trimmed fields, sensitive-value redaction, deterministic ordering, plugin content summarized as counts, optional server permission totals, and hook snapshots tagged with their source.
* **Bug Fixes**
  * Server URLs now strip embedded credentials while preserving unparseable URLs; hook arg redaction preserves positional ordering and omits empty arg lists.
* **Tests**
  * Expanded coverage validating trimming, redaction, counts, sorting, URL handling, tagging, and nil defaults.

### DIFF
--- a/internal/zerocommands/backend_snapshots.go
+++ b/internal/zerocommands/backend_snapshots.go
@@ -1,6 +1,7 @@
 package zerocommands
 
 import (
+	"net/url"
 	"sort"
 	"strings"
 
@@ -89,13 +90,18 @@ type BackendLifecycleSnapshot struct {
 
 // MCPServerSnapshotFromServer converts a single mcp.Server into its
 // typed snapshot. Secret material in `Env` and `Headers` is never
-// copied; only the key counts are recorded.
+// copied; only the key counts are recorded. The URL field is run
+// through stripURLCredentialsFromURL so any userinfo embedded in
+// the URL (https://user:token@host) is removed before the snapshot
+// leaves the runtime. A URL that fails to parse is returned
+// trimmed, not empty, so the operator still sees the configured
+// endpoint when triaging a malformed MCP configuration.
 func MCPServerSnapshotFromServer(server mcp.Server) MCPServerSnapshot {
 	return MCPServerSnapshot{
 		Name:        strings.TrimSpace(server.Name),
 		Type:        string(server.Type),
 		Identity:    strings.TrimSpace(server.Identity),
-		URL:         strings.TrimSpace(server.URL),
+		URL:         stripURLCredentialsFromURL(server.URL),
 		Command:     strings.TrimSpace(server.Command),
 		ArgCount:    len(server.Args),
 		EnvKeyCount: len(server.Env),
@@ -250,21 +256,46 @@ func NewBackendLifecycleSnapshot(servers []mcp.Server, hookDefs []hooks.Definiti
 
 // redactStringSlice runs every element of the input slice through
 // the standard redaction pipeline and trims surrounding whitespace.
-// A nil or empty input returns nil so the JSON output omits the
-// field entirely.
+// Position is preserved: a fully-redacted element becomes an empty
+// string in the output, and a whitespace-only input becomes an
+// empty string. The output length always matches the input length
+// so consumers can rely on arg position when correlating with the
+// source definition. A nil or empty input returns nil so the JSON
+// output omits the field entirely.
 func redactStringSlice(values []string) []string {
 	if len(values) == 0 {
 		return nil
 	}
-	out := make([]string, 0, len(values))
-	for _, value := range values {
-		redacted := redaction.RedactString(strings.TrimSpace(value), redaction.Options{})
-		if redacted != "" {
-			out = append(out, redacted)
+	out := make([]string, len(values))
+	for index, value := range values {
+		trimmed := strings.TrimSpace(value)
+		if trimmed == "" {
+			out[index] = ""
+			continue
 		}
-	}
-	if len(out) == 0 {
-		return nil
+		out[index] = redaction.RedactString(trimmed, redaction.Options{})
 	}
 	return out
+}
+
+// stripURLCredentialsFromURL returns value with any embedded
+// userinfo (https://user:token@host) removed. The helper is
+// tolerant of malformed input: a URL that fails to parse is
+// returned trimmed, not empty, so the operator still sees the
+// configured endpoint when triaging an MCP configuration that
+// contains an unparseable URL. A blank input returns blank.
+func stripURLCredentialsFromURL(value string) string {
+	trimmed := strings.TrimSpace(value)
+	if trimmed == "" {
+		return ""
+	}
+	parsed, err := url.Parse(trimmed)
+	if err != nil || parsed == nil {
+		return trimmed
+	}
+	if parsed.User == nil {
+		return trimmed
+	}
+	parsed.User = nil
+	return parsed.String()
 }
--- a/internal/zerocommands/backend_snapshots.go
+++ b/internal/zerocommands/backend_snapshots.go
@@ -1,0 +1,270 @@
+package zerocommands
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/hooks"
+	"github.com/Gitlawb/zero/internal/mcp"
+	"github.com/Gitlawb/zero/internal/plugins"
+	"github.com/Gitlawb/zero/internal/redaction"
+)
+
+// MCPServerSnapshot is the typed view of a single configured MCP
+// server as it is exposed to the TUI render path, the headless
+// `zero mcp` command, and PR/CI automation. The snapshot strips
+// every field that can carry a secret. Command arguments are
+// preserved because the tool surface (path, flags) is the part a
+// maintainer needs to see when triaging an MCP failure. Environment
+// variables and HTTP headers are summarised as redacted key counts
+// instead of being copied verbatim, so a token in MCP_AUTH_TOKEN
+// never reaches the headless JSON output.
+type MCPServerSnapshot struct {
+	Name         string   `json:"name"`
+	Type         string   `json:"type"`
+	Identity     string   `json:"identity,omitempty"`
+	URL          string   `json:"url,omitempty"`
+	Command      string   `json:"command,omitempty"`
+	ArgCount     int      `json:"argCount"`
+	EnvKeyCount  int      `json:"envKeyCount"`
+	HeaderCount  int      `json:"headerCount"`
+	ToolCount    int      `json:"toolCount"`
+	AllowGranted int      `json:"allowGranted"`
+	DenyGranted  int      `json:"denyGranted"`
+}
+
+// HookSnapshot is the typed view of a single configured hook as it
+// is exposed to the TUI render path, the headless `zero hooks`
+// command, and PR/CI automation. The command and arguments are
+// preserved because they are the operator's primary tool for
+// understanding which shell command will run. The matcher is
+// preserved because it is the operator's primary tool for
+// understanding when the hook will fire.
+type HookSnapshot struct {
+	ID          string   `json:"id"`
+	Name        string   `json:"name,omitempty"`
+	Event       string   `json:"event"`
+	Matcher     string   `json:"matcher,omitempty"`
+	Command     string   `json:"command"`
+	Args        []string `json:"args,omitempty"`
+	Enabled     bool     `json:"enabled"`
+	Source      string   `json:"source,omitempty"`
+}
+
+// PluginSnapshot is the typed view of a single loaded plugin as it
+// is exposed to the TUI render path, the headless `zero plugins`
+// command, and PR/CI automation. The path fields are preserved
+// because the operator needs to know where the manifest came from
+// when triaging a load failure. Counts replace the full slice of
+// tools, hooks, prompts, and skills so the headless JSON output
+// stays small even for plugins with many extensions.
+type PluginSnapshot struct {
+	ID            string `json:"id"`
+	Name          string `json:"name"`
+	Version       string `json:"version,omitempty"`
+	Description   string `json:"description,omitempty"`
+	Enabled       bool   `json:"enabled"`
+	Source        string `json:"source"`
+	Root          string `json:"root,omitempty"`
+	PluginDir     string `json:"pluginDir,omitempty"`
+	ManifestPath  string `json:"manifestPath,omitempty"`
+	ToolCount     int    `json:"toolCount"`
+	PromptCount   int    `json:"promptCount"`
+	SkillCount    int    `json:"skillCount"`
+	HookCount     int    `json:"hookCount"`
+}
+
+// BackendLifecycleSnapshot bundles the typed snapshots for the
+// three backend surfaces that the WorkSplit PRD places in Gnanam's
+// lane: MCP servers, hooks, and plugins. `zero doctor` and the
+// headless `zero config` command can return one of these to give
+// the operator a single payload describing the full extensibility
+// surface. The three inner slices are always non-nil so JSON
+// output is `[]` and never `null`.
+type BackendLifecycleSnapshot struct {
+	MCPServers []MCPServerSnapshot `json:"mcpServers"`
+	Hooks      []HookSnapshot      `json:"hooks"`
+	Plugins    []PluginSnapshot    `json:"plugins"`
+}
+
+// MCPServerSnapshotFromServer converts a single mcp.Server into its
+// typed snapshot. Secret material in `Env` and `Headers` is never
+// copied; only the key counts are recorded.
+func MCPServerSnapshotFromServer(server mcp.Server) MCPServerSnapshot {
+	return MCPServerSnapshot{
+		Name:        strings.TrimSpace(server.Name),
+		Type:        string(server.Type),
+		Identity:    strings.TrimSpace(server.Identity),
+		URL:         strings.TrimSpace(server.URL),
+		Command:     strings.TrimSpace(server.Command),
+		ArgCount:    len(server.Args),
+		EnvKeyCount: len(server.Env),
+		HeaderCount: len(server.Headers),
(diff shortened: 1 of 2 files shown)


gnanam1990:
Addressed the three actionable review items from CodeRabbit:

1. **URL credential leak in `MCPServerSnapshot.URL`** — fixed. Added `stripURLCredentialsFromURL` that parses the URL with `net/url` and clears the `User` field. The helper is tolerant: a URL that fails to parse is returned trimmed (not empty) so the operator still sees the configured endpoint when triaging a broken configuration.

2. **`redactStringSlice` dropped fully-redacted elements** — fixed. The helper now preserves position: it pre-allocates the output slice to the input length and assigns each element by index. A fully-redacted element becomes an empty string in the output rather than disappearing, and whitespace-only inputs become empty strings by index. A nil or empty input still returns nil so the JSON output omits the field entirely.

3. **Arg-length preservation test coverage** — added. `TestHookSnapshotFromDefinitionRedactsCommandAndTrimsFields` now asserts `len(snapshot.Args) == len(def.Args)`. Two new tests cover the URL sanitization: `TestMCPServerSnapshotStripsURLCredentials` verifies the username, password, and `@` separator do not appear anywhere in the snapshot, and `TestMCPServerSnapshotKeepsUnparseableURLInsteadOfEmpty` verifies the malformed-URL tolerance contract.

Branch rebased onto current `main` (post-#89 merge) before the fixes landed. `go build`, `go vet`, and the affected test packages all pass. Ready for re-review.

anandh8x (approved):
## What's good

- **The `internal/zerocommands` package is the right home for typed snapshot contracts.** It already hosts the config, provider, model, and session snapshots (`contracts.go`), and adding the three backend snapshots here keeps the surface consistent. The package is consumer-only: it converts backend structs into stable, trimmed, redacted, sorted views. No runtime behavior changes, and no consumer is forced to migrate in this PR.
- **MCP `Env` and `Headers` are correctly reduced to counts.** `EnvKeyCount` and `HeaderCount` preserve the operator's diagnostic value (the size of the secret-bearing maps) without ever serializing the values. `TestMCPServerSnapshotFromServerStripsSecretsAndCountsMaps` exercises a realistic `MCP_AUTH_TOKEN` and `Authorization: Bearer ...` case and asserts that neither the secret nor the literal "Bearer" nor the key name appears in the JSON. That is the right contract for a snapshot that lands in TUI, headless JSON, audit log, and PR/CI automation.
- **`MCPServerSnapshotWithCounts` is a thoughtful addition.** The base snapshot records config-time fields; the `WithCounts` variant lets a caller that has a live `tools.Registry` and a live `mcp.PermissionStore` attach `ToolCount`, `AllowGranted`, `DenyGranted` so the headless `zero mcp` command can answer both "what is configured?" and "what is live?" in one typed payload. The `nil` counts contract is explicit and tested.
- **Stable ordering across all three snapshot kinds.** `MCPServerSnapshots` sorts by `Name`; `HookSnapshots` sorts by `(ID, Event)` so duplicate-ID hooks get a deterministic event order; `PluginSnapshots` sorts by `ID`. `sort.SliceStable` ensures the input order is the tiebreak, so two snapshots with the same key are not randomly swapped.
- **Empty-slice stability is enforced and tested.** Every `FooSnapshots(nil)` returns a non-nil empty slice whose JSON encoding is exactly `[]`. The test `TestMCPServerSnapshotsSortsAndReturnsEmptySliceForEmptyInput` and `TestPluginSnapshotsSortsByIDAndReturnsEmptySliceForEmptyInput` both assert `string(encoded) != "[]"` would fail. That is the right invariant for a JSON consumer that expects arrays, not `null`.
- **Hook command and args go through the standard redaction pipeline.** `HookSnapshotFromDefinition` runs `def.Command` and every entry of `def.Args` through `redaction.RedactString`, and the test feeds `sk-proj-...` through an arg to confirm the snapshot does not carry the secret. That is consistent with how other CLI/audit surfaces treat hook command lines.
- **`HookSnapshotsWithSource` is a small but useful escape hatch.** A caller that has a single `hooks.LoadResult` can tag every snapshot with the same `ConfigSource` in one pass instead of mapping over the result and threading the source through. The helper is tiny and the test pins the contract.
- **`BackendLifecycleSnapshot` is the right shape for the `zero doctor` and `zero config` consumers.** One payload that describes the full extensibility surface (MCP servers, hooks, plugins) with non-nil slices is much easier to render than three separate snapshots. The bundled JSON shape is asserted by `TestNewBackendLifecycleSnapshotBundlesAndDefaultsNilsToEmpty`.
- **Plugin path fields are preserved verbatim.** `Root`, `PluginDir`, and `ManifestPath` are the operator's primary tool for triaging a load failure (e.g., "did the manifest come from the user config or a project-local override?"). Trimming is applied, but the path itself is copied.
- **Slices and maps are not aliased.** The snapshot constructors copy via `MCPServerSnapshots` and `PluginSnapshots`; `HookSnapshots` builds a new slice; `redactStringSlice` returns a fresh slice. A consumer that mutates the snapshot cannot accidentally mutate the underlying `mcp.Server` / `hooks.Definition` / `plugins.LoadedPlugin`.

1. **`MCPServerSnapshot.URL` is preserved verbatim, including any embedded userinfo.** A configured `mcp.Server` of the form `{"url": "https://admin:secret123@api.example.com/v1"}` will serialize as `"url": "https://admin:secret123@api.example.com/v1"` in the snapshot. The comment block on `MCPServerSnapshot` says "Env and Headers never leak" but does not address URL credentials. The fix is small: parse the URL with `net/url`, clear `u.User`, then re-encode. A unit test that constructs a server with a credentialed URL and asserts the snapshot URL contains no `:` after `https://` would lock the contract. CodeRabbit's review surfaced this; it is worth a follow-up commit, but the slice as designed is not regressing anything that was not already in the `mcp.Server` struct.

2. **`MCPServerSnapshot.ArgCount` is a regression in triage usefulness compared to the source struct.** The doc comment on `MCPServerSnapshot` says "Command arguments are preserved because the tool surface (path, flags) is the part a maintainer needs to see when triaging an MCP failure", but the actual struct has `ArgCount int` and no `Args` field. The PR description is consistent with the struct ("argCount, envKeyCount, headerCount replace the slices and maps with counts"). The doc comment on the type is misleading. Either add `Args []string` to the struct (redacted, with whitespace-only entries dropped) and update the test, or fix the comment to say "args are reduced to a count to keep the snapshot small; use the headless `zero mcp` command with `--verbose` to see the full arg list".

3. **`redactStringSlice` drops empty results, which loses positional fidelity.** If an arg like `   ` (whitespace only) or a fully-redacted secret is passed in, the returned slice has one fewer element. For a hook command of `["sh", "-c", "auth token sk-proj-...", "---", "next"]`, the redacted output is `["sh", "-c", "---", "next"]` (the secret is replaced with `[REDACTED]`, not dropped, so this specific case is fine). But the helper also drops `out = append(out, redacted)` when `redacted == ""` even if the input was non-empty. A future caller that relies on arg position will be surprised. Two options: (a) preserve the position by always appending `redacted` (even when empty), or (b) document the dropping behavior in the helper's doc comment so the next caller knows. Right now the behavior is silent.

4. **`redactStringSlice` is not unit-tested directly.** The hook snapshot test covers a happy-path arg with a secret, but it does not cover the all-redacted case (where `Args` would be `nil` because every element was dropped). A focused test would lock in the behavior and make observation 3 actionable.

5. **`MCPServerSnapshots` sorts by `Name`.** If two servers share a name, the order between them is determined by input order (via `sort.SliceStable`). That is fine for a config that does not allow duplicate names, but `mcp.NormalizeConfig` could in principle produce duplicates. A follow-up test that feeds two servers with the same name would confirm the stable-tiebreak behavior, or a `sort.Slice` with a secondary key on the original index would make the order fully deterministic regardless of input.

6. **`MCPServerSnapshotWithCounts` is a good pattern but the live path is not exercised.** The test only constructs a counts struct directly. A future `zero mcp --json` consumer should be the place where the registry + permission store are wired in, and that consumer should add its own integration test. The slice as designed is the right contract for that consumer.

7. **`PluginSnapshot.Root`, `PluginDir`, `ManifestPath` may contain the user's home directory.** This is arguably PII, but the source `plugins.LoadedPlugin` already carries these strings, so the snapshot is no leakier than the loaded plugin. A future redaction pass that replaces `$HOME` with `~` would help when the snapshot is sent to a PR comment or audit log. Not blocking; the snapshot is consistent with the source.

8. **`HookSnapshot.Matcher` is trimmed but not redacted.** A matcher like `write_file.* --token=sk-proj-...` would have the secret in the matcher. Right now `Matcher` is a free-form string from the user's config; the snapshot copies it verbatim. If the matcher can carry secrets, the snapshot should redact it. A test that feeds a matcher with a secret would surface the gap.

9. **No tests for `redactStringSlice` ordering with mixed secrets.** A test like `redactStringSlice([]string{"", "-c", "   "})` would assert whether the helper preserves position or drops. Locking that in would make observation 3 a one-line fix.

10. **The package has no top-level `doc.go` or package comment.** Other packages (`internal/sandbox`, `internal/zerocommands` itself via `contracts.go`) carry a brief package comment explaining the pattern. A one-liner like `// Package zerocommands defines typed snapshot contracts for the headless, TUI, audit, and PR/CI consumers. The constructors in this package convert backend structs into stable, trimmed, redacted, sorted views and are the only sanctioned way for those consumers to render backend state.` would help new readers.

11. **Cross-PR consistency with PR #92.** PR #92 adds the sandbox snapshot constructors in the same package; the contract here uses `redactStringSlice` while PR #92 uses inline `redaction.RedactString` in `HookSnapshotFromDefinition`. The two patterns are equivalent but slightly different (slice-aware vs. string-only). If the team wants one helper for the pattern, `redactStringSlice` could be exported as `RedactStringSlice` and reused in #92's `SandboxViolationSnapshot` (or wherever a list of strings needs redaction). Cosmetic.

12. **`BackendLifecycleSnapshot` is never nil but never tested for a populated case where the slices share items with the input.** The bundled test only checks that one of each kind round-trips, not that the input slices are not aliased. A test that mutates the input slice after construction and asserts the snapshot slice is unchanged would lock the no-aliasing contract. Same applies to the per-type `FooSnapshots` helpers.

13. **The `MCPServerSnapshot.Identity` field is new relative to what other surfaces expose.** It is a per-server identity string from `mcp.Server.Identity` that the runtime uses for permission keys. Including it in the snapshot is correct (it is non-secret metadata), but no test in this PR asserts what the field is supposed to be. A follow-up test that constructs two servers with different identities and asserts the snapshot preserves them would be cheap.

14. **`mcp.Server.URL` is JSON-omitted when empty** (via `,omitempty` on the source struct), so the snapshot's `URL` field correctly serializes only when set. The snapshot type itself does not have `omitempty` on `URL`, which is consistent with the source.

15. **`HookSnapshot.Event` is `string` (not `hooks.Event`).** That is a deliberate widening so the snapshot is JSON-stable even when a new event kind is added. The trade-off is that the snapshot accepts any string; a future consumer that wants to validate the event kind would have to map against `hooks.Event` constants. Acceptable.

16. **`PluginSnapshot.Source` is `string` (not `plugins.Source`).** Same widening as `Event`. Consistent.

17. **The PR description says "Closes the typed-snapshot gap for the three backend surfaces in §7.E that the TUI render path, the headless commands, and PR/CI automation consume."** A consumer PR that uses these snapshots is not in this PR. The slice is the contract; follow-up slices are the consumers. That is the right sequencing for typed contracts.

18. **No benchmark.** The snapshot helpers are called in render paths (TUI) and on every audit log entry. A small benchmark in `backend_snapshots_test.go` that times 1000 snapshot conversions would catch a future regression where someone re-aliases the input or adds an O(n) string allocation per call. Not blocking; performance is fine for the current scale.

19. **`MCPServerSnapshot.Command` is preserved verbatim.** For a `stdio` server, the command (e.g., `npx`, `python3`) is the operator's primary tool for understanding the server. For an `http` server, `Command` is empty and `URL` is the operator's tool. The current shape covers both.

20. **`mcp.Server.Args` for an `http` server is typically empty.** `ArgCount` will be `0` for an `http` snapshot, which is correct but slightly noisy. A typed `mcp.Server` could split into `StdioServer` and `HTTPServer` variants in a future PR, but that is out of scope for this slice.

No blockers. The slice is well-scoped, well-tested, and closes the typed-snapshot gap for the MCP, hooks, and plugins surfaces without changing runtime behavior. The three follow-ups I would prioritize next: (1) strip URL userinfo in `MCPServerSnapshotFromServer` so credentialed URLs never leak, (2) reconcile the doc comment on `MCPServerSnapshot` with the actual `ArgCount` field, and (3) add a focused test for `redactStringSlice` that covers the all-redacted and whitespace-only cases.

gnanam1990:
@coderabbit review

Vasanthdev2004 (approved):
## Blockers

- None found.

## Non-Blocking

- None.

## Looks Good

- The current head `08cf126fe4e76ba4a7e75ce3a9b261f2e8d6f974` fixes the earlier snapshot safety issues: MCP URLs with embedded userinfo are sanitized before JSON serialization, and hook arg snapshots preserve argument positions while redacting secret-looking values.
- The MCP/hook/plugin backend snapshots keep the public diagnostic payload typed and compact: env/header values are counted rather than serialized, hook commands/args go through the standard redaction pipeline, plugin extension slices are collapsed to counts, and all snapshot slices stay non-nil for stable JSON output.
- Deterministic sorting and nil/empty-slice behavior are covered well, including the URL credential and arg-position regressions.

Validation run locally on the PR head and on a latest-main test merge:

- `go test -count=1 ./internal/zerocommands -run 'TestPR90ReviewProbe' -v` — passed; verified URL credentials are stripped and hook arg positions are preserved.
- `go test -count=1 ./internal/zerocommands ./internal/mcp ./internal/hooks ./internal/plugins ./internal/redaction` — passed.
- `go test -count=1 -p 1 ./...` — passed.
- `git diff --check` — passed.
- Latest-main test merge onto `b70fea7d6ffd4ef085aef71cb95997ce960d86e9` merged cleanly and repeated the targeted/full Go validation successfully.

Vasanthdev2004:
@coderabbitai approve

PR #82, opened 2026-06-06
Add runtime permission and verification event contracts

gnanam1990:
## Summary

Adds the next Gnanam-owned Go runtime contract slice for safer automation, command UX, session replay planning, and verification/reporting consumers.

### Runtime permission and sandbox events
- Adds structured permission request/decision events to the agent runtime.
- Emits sandbox decisions through the tool registry callback path.
- Records permission decisions in session events for later audit/search.
- Adds stream-json and text/JSON exec output support for permission events.
- Strengthens workspace-aware sandbox risk classification for path escapes.

### Runtime command contracts
- Adds stable Go command snapshot structs for config, providers, models, sessions, and session trees.
- Moves config/model/session CLI JSON paths onto shared runtime contracts.
- Redacts provider URLs and preserves provider/model capability metadata.

### Session rewind and compaction planning
- Adds backend-only rewind plans by sequence or event id.
- Adds compaction plans with preserved-event windows and bounded summary prompts.
- Exposes `zero sessions rewind-plan` and `zero sessions compact-plan` preview commands.
- Validates command-specific flags so rewind/compaction options cannot be used on unrelated session commands.

### Verification and change event contracts
- Adds versioned `zero.verify.report.v1`, `zero.selfverify.report.v1`, and `zero.changes.report.v1` snapshots.
- Adds event arrays for verification start/check/finish, self-verify attempts, and changed-file summaries.
- Preserves kind/framework metadata on verification results for downstream TUI and automation rendering.
- Wires `zero verify --json`, self-verification JSON, and `zero changes inspect --json` to stable snapshot outputs.

## Validation

- `go test -count=1 -p 1 ./...`
- `bun run typecheck`
- `bun test ./tests --timeout 15000`
- `bun run build`
- `bun run smoke:build`
- `bun run smoke:go`
- `./zero verify --json --only bun.typecheck`
- `./zero changes inspect --json`
- `./zero sessions --help`
- `./zero models --json`
- `git diff --check main..HEAD`

## Reviewers

Please review: @vasanthdev2004 @anandh8x

## Notes

- This is intentionally a larger grouped runtime-contract PR, covering related Gnanam backend slices in one reviewable batch.
- Checked current open PR #81 for overlap before opening this PR; it touches sandbox platform adapter files and does not overlap with this branch.

## Summary by CodeRabbit

* **New Features**
  * Permission events are emitted during runs, streamed in exec JSON, and persisted to session recordings.
[...]

PR #92, opened 2026-06-06
feat(zerocommands): add typed sandbox policy, risk, violation, and decision snapshots

gnanam1990:
## Summary

The merged permission and verification event contracts (#82) added typed snapshots for config, providers, models, and sessions, and PR #89 added the typed snapshot for the persistent sandbox grant store, but the live sandbox policy, the live risk classification, the live violation, and the live decision were still exposed as the raw backend structs. The TUI render path, the headless `zero sandbox policy --json` and `zero sandbox decide` commands, the audit log, and PR/CI automation all need typed snapshots so they do not hand-roll `map[string]any` payloads and so they do not have to re-implement the runtime default-mode translation themselves.

This commit adds five typed snapshot constructors and a bundled plan snapshot to `internal/zerocommands`.

### SandboxPolicySnapshot

- `mode`, `network`, `enforceWorkspace`, `denyDestructiveShell`, and `allowPolicyOnlyRunner` are copied verbatim from the underlying `sandbox.Policy`.
- `effectiveMode` is the resolved policy mode. An empty `Mode` in the input falls back to `ModeEnforce`, which is the runtime default. Consumers can read `EffectiveMode` to render the "what is actually in effect" line without re-implementing the default.
- `effectiveMode` is JSON-omitted when empty so the snapshot shape stays stable for downstream consumers.

### SandboxRiskSnapshot

- `level`, `categories`, `reason`.
- `categories` is copied and sorted alphabetically so the JSON output is deterministic.
- `reason` is trimmed.

### SandboxViolationSnapshot

- `code`, `toolName`, `action`, `path`, `reason`, `recoverable`.
- The pointer-returning helper returns nil for a nil input so callers can chain it through `SandboxDecisionSnapshot` without nil checks at the call site.

### SandboxBackendSnapshot

- `name`, `available`, `platform`, `fallback`, `commandWrapping`, `nativeIsolation`, `executable`, `message`.
- `executable` and `message` are trimmed.

### SandboxPlanSnapshot

- Bundles policy, backend, restrictions, and workspaceRoot into one payload so `zero sandbox policy --json` can return one typed result describing the full sandbox posture.
- Each entry of `restrictions` is trimmed, whitespace-only entries are dropped, and the result is sorted alphabetically.
- `workspaceRoot` is trimmed and JSON-omitted when empty.

### SandboxDecisionSnapshot

- `action`, `reason`, `risk`, `grantMatched`, `grant`, `violation`.
- The optional grant pointer is converted with `SandboxGrantSnapshotFromGrant` (from PR #89) so the JSON shape matches the persistent-grant snapshot exactly.
- The optional violation pointer is converted with `SandboxViolationSnapshotFromViolation` so the snapshot always carries the same shape regardless of the decision outcome.
- When grant or violation is nil, the field is JSON-omitted.

## Tests

- `TestSandboxPolicySnapshotFromPolicyFillsEffectiveMode`: verifies `EffectiveMode` is enforced for an empty `Mode`, set to `ModeDisabled` for an explicit disabled mode, and preserved for an explicit enforced mode.
- `TestSandboxRiskSnapshotFromRiskSortsCategoriesAndTrimsReason`: verifies alphabetical sort, trimmed reason, and that a nil `Categories` input produces a nil output (no spurious empty slice).
- `TestSandboxViolationSnapshotFromViolationHandlesNilAndTrims`: verifies the nil-input contract and the trim behaviour for every string field.
- `TestSandboxBackendSnapshotFromBackendCopiesAllFields`: verifies the boolean and string fields round-trip and the trim behaviour for the executable and message.
- `TestSandboxPlanSnapshotFromPlanSortsRestrictionsAndTrimsRoot`: verifies alphabetical sort, per-entry trim, and workspaceRoot trim.
- `TestSandboxDecisionSnapshotFromDecisionAllowBranchHasNoViolation`: verifies the allow path has no grant and no violation.
- `TestSandboxDecisionSnapshotFromDecisionPersistentDenyCarriesGrantAndViolation`: verifies the deny path carries both the resolved grant and the violation, and that the grant snapshot matches the persistent-grant snapshot shape from PR #89.
- `TestSandboxDecisionSnapshotJSONShapeIsStable`: verifies the stable keys and the nil-pointer omission.
- `TestSandboxPolicySnapshotJSONOmitsEmptyEffectiveMode`: verifies the `omitempty` tag.

## Verification

- `go build ./...`
- `go vet ./...`
- `go test -count=1 -p 1 ./internal/zerocommands/... ./internal/sandbox/... ./internal/redaction/...`

## DRI

Gnanam (runtime core owner per `docs/WORK_SPLIT_PRD.md` §4). Closes the typed-snapshot gap for the live sandbox surface that the TUI render path, the `zero sandbox` headless commands, the audit log, and PR/CI automation consume.

## Out of scope

This PR only adds the typed contracts and the constructors. The TUI render path, the `zero sandbox policy --json` and `zero sandbox decide` headless commands, the audit log sink, and the PR/CI automation consumer are follow-up work in their respective owners' lanes.

## Summary by CodeRabbit

* **Chores**
  * Added an internal snapshot-based serialization layer for sandbox data with normalized fields and deterministic ordering.

* **Tests**
  * Added comprehensive tests for snapshot conversions and edge cases.
  * Adjusted a test fixture to ensure the wrapper executes as an ES module during tests.

### DIFF
--- a/internal/npmwrapper/npmwrapper_test.go
+++ b/internal/npmwrapper/npmwrapper_test.go
@@ -115,6 +115,13 @@ func copyWrapperFixture(t *testing.T) string {
 		t.Fatalf("ReadFile wrapper: %v", err)
 	}
 	dir := t.TempDir()
+	// Create a package.json with "type": "module" so the isolated .js fixture
+	// is treated as ESM (matching how it runs when installed from the real package.json).
+	// Without this, Node treats .js as CJS on all platforms, causing top-level import
+	// to fail with SyntaxError before reaching the missing-binary logic.
+	if err := os.WriteFile(filepath.Join(dir, "package.json"), []byte(`{"type":"module"}`), 0o644); err != nil {
+		t.Fatalf("WriteFile package.json fixture: %v", err)
+	}
 	binDir := filepath.Join(dir, "bin")
 	if err := os.MkdirAll(binDir, 0o755); err != nil {
 		t.Fatalf("MkdirAll bin: %v", err)
--- a/internal/zerocommands/sandbox_snapshots.go
+++ b/internal/zerocommands/sandbox_snapshots.go
@@ -1,0 +1,202 @@
+package zerocommands
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/sandbox"
+)
+
+// SandboxPolicySnapshot is the typed view of the live sandbox policy
+// as it is exposed to the TUI render path, the headless
+// `zero sandbox policy --json` command, and PR/CI automation. The
+// snapshot is a strict superset of the on-disk policy shape: every
+// field of sandbox.Policy is copied verbatim, and an EffectiveMode
+// field is added so consumers do not have to translate the empty
+// string into the enforced default themselves.
+type SandboxPolicySnapshot struct {
+	Mode                  string `json:"mode"`
+	Network               string `json:"network"`
+	EnforceWorkspace      bool   `json:"enforceWorkspace"`
+	DenyDestructiveShell  bool   `json:"denyDestructiveShell"`
+	AllowPolicyOnlyRunner bool   `json:"allowPolicyOnlyRunner"`
+	EffectiveMode         string `json:"effectiveMode,omitempty"`
+}
+
+// SandboxRiskSnapshot is the typed view of a sandbox.Risk as it is
+// exposed to the TUI render path and PR/CI automation. The snapshot
+// copies Level and Reason verbatim and sorts Categories
+// alphabetically so the JSON output is deterministic.
+type SandboxRiskSnapshot struct {
+	Level      string   `json:"level"`
+	Categories []string `json:"categories"`
+	Reason     string   `json:"reason"`
+}
+
+// SandboxViolationSnapshot is the typed view of a sandbox.Violation
+// as it is exposed to the TUI render path, the audit log, and
+// PR/CI automation. The Path field is preserved because the
+// operator needs it to triage an out-of-workspace or symlink
+// traversal violation.
+type SandboxViolationSnapshot struct {
+	Code        string `json:"code"`
+	ToolName    string `json:"toolName,omitempty"`
+	Action      string `json:"action"`
+	Path        string `json:"path,omitempty"`
+	Reason      string `json:"reason"`
+	Recoverable bool   `json:"recoverable"`
+}
+
+// SandboxBackendSnapshot is the typed view of a sandbox.Backend as
+// it is exposed to the TUI render path, the headless
+// `zero sandbox policy --json` command, and PR/CI automation. The
+// snapshot is a strict superset of the on-disk backend shape: every
+// field of sandbox.Backend is copied verbatim, and a NativeIsolation
+// field is preserved so the JSON shape matches what consumers
+// already expect.
+type SandboxBackendSnapshot struct {
+	Name            string `json:"name"`
+	Available       bool   `json:"available"`
+	Platform        string `json:"platform,omitempty"`
+	Fallback        bool   `json:"fallback"`
+	CommandWrapping bool   `json:"commandWrapping"`
+	NativeIsolation bool   `json:"nativeIsolation"`
+	Executable      string `json:"executable,omitempty"`
+	Message         string `json:"message,omitempty"`
+}
+
+// SandboxPlanSnapshot is the typed view of a sandbox.BackendPlan as
+// it is exposed to the TUI render path, the headless
+// `zero sandbox policy --json` command, and PR/CI automation. The
+// snapshot bundles the policy, the backend, and the human-readable
+// restriction warnings so the operator gets one payload describing
+// the entire sandbox posture.
+type SandboxPlanSnapshot struct {
+	Policy       SandboxPolicySnapshot  `json:"policy"`
+	Backend      SandboxBackendSnapshot `json:"backend"`
+	Restrictions []string               `json:"restrictions"`
+	WorkspaceRoot string                `json:"workspaceRoot,omitempty"`
+}
+
+// SandboxDecisionSnapshot is the typed view of a live sandbox.Decision
+// as it is exposed to the TUI render path, the audit log, and
+// PR/CI automation. The snapshot includes the resolved grant
+// (when the decision was driven by a persistent grant) and the
+// violation (when the decision was a deny) so consumers do not
+// have to reach back into the sandbox package to render the
+// outcome.
+type SandboxDecisionSnapshot struct {
+	Action       string                      `json:"action"`
+	Reason       string                      `json:"reason,omitempty"`
+	Risk         SandboxRiskSnapshot         `json:"risk"`
+	GrantMatched bool                        `json:"grantMatched"`
+	Grant        *SandboxGrantSnapshot       `json:"grant,omitempty"`
+	Violation    *SandboxViolationSnapshot   `json:"violation,omitempty"`
+}
+
+// SandboxPolicySnapshotFromPolicy converts a sandbox.Policy into its
+// typed snapshot. The EffectiveMode field is the resolved policy
+// mode: an empty Mode falls back to ModeEnforce, which is the
+// runtime default. Consumers can read EffectiveMode to render the
+// "what is actually in effect" line without re-implementing the
+// default.
+func SandboxPolicySnapshotFromPolicy(policy sandbox.Policy) SandboxPolicySnapshot {
+	mode := strings.TrimSpace(string(policy.Mode))
+	if mode == "" {
+		mode = string(sandbox.ModeEnforce)
+	}
+	return SandboxPolicySnapshot{
+		Mode:                  string(policy.Mode),
+		Network:               string(policy.Network),
+		EnforceWorkspace:      policy.EnforceWorkspace,
+		DenyDestructiveShell:  policy.DenyDestructiveShell,
+		AllowPolicyOnlyRunner: policy.AllowPolicyOnlyRunner,
+		EffectiveMode:         mode,
+	}
+}
+
+// SandboxRiskSnapshotFromRisk converts a sandbox.Risk into its
+// typed snapshot. The Categories slice is sorted alphabetically
+// and copied so the snapshot does not alias the input.
+func SandboxRiskSnapshotFromRisk(risk sandbox.Risk) SandboxRiskSnapshot {
+	categories := append([]string(nil), risk.Categories...)
+	sort.Strings(categories)
+	return SandboxRiskSnapshot{
+		Level:      string(risk.Level),
+		Categories: categories,
+		Reason:     strings.TrimSpace(risk.Reason),
+	}
+}
+
+// SandboxViolationSnapshotFromViolation converts a sandbox.Violation
+// into its typed snapshot. A nil input returns a zero snapshot so
+// the helper is safe to call when the decision did not produce a
+// violation.
+func SandboxViolationSnapshotFromViolation(violation *sandbox.Violation) *SandboxViolationSnapshot {
+	if violation == nil {
+		return nil
+	}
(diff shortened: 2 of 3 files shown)


anandh8x (approved):
## What's good

- **The five typed snapshot constructors close the gap for the live sandbox surface.** PR #89 added the persistent-grant snapshot; this PR adds the policy, risk, violation, backend, plan, and decision snapshots. Together they cover every value that the TUI render path, the headless `zero sandbox policy --json` and `zero decide` commands, the audit log, and PR/CI automation consume from `internal/sandbox`. The contract is consumer-only — no runtime behavior changes, and no consumer is forced to migrate in this PR.
- **`EffectiveMode` is the right user-facing addition.** The runtime default is `ModeEnforce` (see `DefaultPolicy` and the engine's `NewEngine`), but a `sandbox.Policy{}` zero value has `Mode == ""`. A consumer that reads `Policy.Mode` directly would render `"mode": ""` and have to know to translate the empty string into `enforce`. `SandboxPolicySnapshot.EffectiveMode` does that translation once, so the TUI, the headless command, and the audit log can render the "what is actually in effect" line without re-implementing the default. The test `TestSandboxPolicySnapshotFromPolicyFillsEffectiveMode` locks the contract for enforced, disabled, and empty-Mode cases.
- **`omitempty` on `EffectiveMode` keeps the JSON shape stable.** The test `TestSandboxPolicySnapshotJSONOmitsEmptyEffectiveMode` asserts the field is omitted when empty. A consumer that only wants the effective mode can rely on the key being present (or absent) consistently.
- **`SandboxViolationSnapshotFromViolation` is nil-safe.** Returning `nil` for a nil input means the helper can be chained through `SandboxDecisionSnapshotFromDecision` without nil checks at the call site. The test `TestSandboxViolationSnapshotFromViolationHandlesNilAndTrims` exercises both branches, and the decision test `TestSandboxDecisionSnapshotFromDecisionAllowBranchHasNoViolation` confirms the allow path produces a `nil` violation in the snapshot.
- **`SandboxRiskSnapshot` copies and sorts categories.** `append([]string(nil), risk.Categories...)` produces a fresh slice, so a consumer that mutates `snapshot.Categories` cannot mutate the input `sandbox.Risk`. The test confirms alphabetical sort and that a nil `Categories` input produces a nil output (no spurious empty slice).
- **`SandboxPlanSnapshot` bundles the sandbox posture into one payload.** The test sorts restrictions, trims whitespace-only entries, and trims the workspace root. The `NewBackendLifecycleSnapshot` companion from PR #90 covers the broader extensibility surface; this one covers just the sandbox side. A `zero sandbox policy --json` consumer can return one typed result that includes the policy, the backend, the restrictions, and the workspace root.
- **`SandboxDecisionSnapshot` reuses `SandboxGrantSnapshot` from PR #89.** The grant shape is the same as the persistent-grant snapshot, so the consumer of a deny decision sees the same JSON shape as a `zero sandbox grants list --json` consumer. That is exactly the cross-PR consistency this layered architecture is aiming for.
- **String fields are trimmed consistently.** `Reason`, `Path`, `ToolName`, `Executable`, `Message`, `WorkspaceRoot`, and the `restrictions` entries are all `strings.TrimSpace`d. The tests assert the trim behavior for the most user-visible fields (`Reason`, `Executable`, `Message`, `WorkspaceRoot`, and a per-restriction trim check). A consumer that renders a violation with `Path: " /etc/hosts "` will not see a stray leading space in the JSON.
- **JSON shape stability is pinned by `TestSandboxDecisionSnapshotJSONShapeIsStable`.** The test asserts the four required keys (`action`, `reason`, `risk`, `grantMatched`) are present, and the two pointer fields (`violation`, `grant`) are JSON-omitted when nil. A future change that accidentally adds or removes a JSON key will fail this test.
- **The `SandboxPolicySnapshot.Mode` field is preserved as-is, even when empty.** That is the right call: the snapshot is a view, not a validator. A consumer that wants the runtime default reads `EffectiveMode`; a consumer that wants the raw value reads `Mode`. The two-field shape makes the distinction explicit.

1. **`EffectiveMode` adds a third place that knows about the runtime default.** The default lives in `sandbox.DefaultPolicy()` (sets `Mode: ModeEnforce`), in `sandbox.NewEngine` (falls back to `DefaultPolicy()` when `options.Policy.Mode == ""`), and now in `SandboxPolicySnapshotFromPolicy` (sets `EffectiveMode = ModeEnforce` when `Mode` is empty). A future change to the default would need to be mirrored in three places. A small `func (p Policy) EffectiveMode() PolicyMode` method on `sandbox.Policy` would centralize the rule. The snapshot could then call `policy.EffectiveMode()` instead of duplicating the check. Not blocking, but worth a follow-up.

2. **`SandboxPolicySnapshot.Mode` has no `omitempty` and serializes as `"mode":""` when empty.** A consumer that expects every policy to have a non-empty `mode` would be surprised. The `EffectiveMode` field is the correct field to read for the runtime default; the raw `Mode` field is for "what was the user-supplied value?". A doc comment on the field — or a one-line note in the package comment — would help. Alternatively, adding `omitempty` to `Mode` would force consumers to read `EffectiveMode` (with its own `omitempty` for the edge case) and avoid the empty-string confusion.

3. **`SandboxViolationSnapshot.ToolName` is trimmed.** A tool name from `Tool.Safety().Name` is normally already trimmed, but if a future code path produces a name with trailing whitespace (e.g., a config file with `  write_file  `), the snapshot will normalize it. That is a quiet behavior change. A test that constructs a violation with a non-trimmed tool name would lock the contract; one already exists (`TestSandboxViolationSnapshotFromViolationHandlesNilAndTrims`) and asserts the trim.

4. **`SandboxRiskSnapshot.Categories` is `nil` when the input is `nil`.** That is the test-pinned behavior. A consumer that always iterates over the slice will get zero iterations, which is the right semantic. But a future consumer that does `len(snapshot.Categories) == 0` to detect "low risk" would conflate "low risk with no categories" and "low risk with an empty categories slice". The `Level` field is the right field to read for the risk tier; `Categories` is for explanation. A doc comment on the field would help.

5. **`SandboxBackendSnapshot.Platform` is trimmed.** The underlying `sandbox.Backend.Platform` is set by the adapter to the runtime GOOS (`linux`, `darwin`, `windows`). Trimming is a no-op in practice. The test `TestSandboxBackendSnapshotFromBackendCopiesAllFields` does not assert the trim; a focused assertion would be cheap.

6. **`SandboxPlanSnapshot.WorkspaceRoot` is `string` with `omitempty`.** When the root is set, it is serialized. When empty, it is omitted. A consumer that wants to render "no workspace root" would see the key absent. That is fine, but a follow-up that adds `"workspaceRoot": null` (using a `*string` with `omitempty` would still omit; a `*string` without would serialize as null) would let consumers distinguish "unset" from "empty". Right now the two are indistinguishable.

7. **`SandboxDecisionSnapshot.Grant` dereferences `*decision.Grant` to call `SandboxGrantSnapshotFromGrant(*decision.Grant)`.** The dereference is safe because the helper does not retain a reference. The `*SandboxGrantSnapshot` is freshly allocated inside the helper. A test that mutates the original `sandbox.Grant` after the snapshot is taken and asserts the snapshot is unchanged would lock the no-aliasing contract. Not blocking.

8. **No test for `SandboxPlanSnapshot.Restrictions` with whitespace-only entries.** The function drops them via the `if trimmed != ""` guard. A test like `SandboxPlanSnapshotFromPlan(plan).Restrictions` for an input with `["  ", "valid"]` would assert the result is `["valid"]` (one element, not two). Cheap to add.

9. **No test for `SandboxPlanSnapshot.Restrictions` order with duplicates.** `sort.Strings` produces a stable order for distinct strings; duplicates preserve their order relative to each other. A test that feeds `["b", "a", "a"]` would assert the output is `["a", "a", "b"]`. Minor.

10. **`SandboxDecisionSnapshot` always sets `GrantMatched` from `decision.GrantMatched`, even when `decision.Grant` is nil.** That is the correct behavior: a non-matching grant lookup can still set `GrantMatched: false` with a nil grant. The test for the allow branch confirms `GrantMatched: false` and `Grant: nil` round-trip. Good.

11. **The PR description claims the package is part of a multi-PR typed-snapshot rollout.** PR #89 added `SandboxGrantSnapshot`; this PR adds the rest. The cross-PR consistency is good (the decision snapshot reuses the grant snapshot from #89). A future audit of all five typed snapshot packages (`config`, `providers`, `models`, `sessions`, `sandbox`) for consistent JSON tag casing, field naming, and `omitempty` discipline would be worth a follow-up.

12. **No benchmark.** The snapshot helpers are called in render paths (TUI) and on every audit log entry. A small benchmark in `sandbox_snapshots_test.go` that times 1000 snapshot conversions would catch a future regression where someone re-aliases the input or adds an O(n) string allocation per call. Not blocking.

13. **`SandboxBackendSnapshot.NativeIsolation` is `true` for native backends and `false` for policy-only.** The test `TestSandboxBackendSnapshotFromBackendCopiesAllFields` covers a bubblewrap native case. A complementary test that constructs a policy-only backend and asserts `NativeIsolation: false` (and `Fallback: true`, `CommandWrapping: false`) would lock the dual case. Cheap to add.

14. **`SandboxDecisionSnapshot.Grant` is a `*SandboxGrantSnapshot` and the dereference happens inside the helper.** The helper could be invoked from a hot path (audit log). The allocation is unavoidable for the pointer shape, but a value-type `Grant` field would let the snapshot be a value type at the call site. That is a stylistic choice; the current shape is fine.

15. **`SandboxPolicySnapshot` does not include the `Restrictions` slice.** That lives on `SandboxPlanSnapshot` instead. The split is correct: a policy alone has no restrictions; the plan does. A consumer that wants both would use the plan snapshot. No action needed.

16. **`EffectiveMode` is computed from `strings.TrimSpace(string(policy.Mode))`.** A `PolicyMode` of `"Enforce "` (capitalized, trailing space) is normalized to `"enforce"`. The `policy.Mode` itself is left as `"Enforce "` (untouched) and the `EffectiveMode` is `"enforce"`. A consumer that reads `Mode` directly would see the unsanitized value. The test does not cover this case. A test that feeds `Policy{Mode: " Enforce "}` and asserts `EffectiveMode == "enforce"` and `Mode == " Enforce "` would lock the contract.

17. **The PR description mentions a headless `zero decide` command, but no such command exists in this PR.** The slice is the typed contract; the command is a follow-up in its owner's lane. That is the right sequencing for typed contracts.

18. **`SandboxPlanSnapshot` has no `omitempty` on `Restrictions` because the slice is always non-nil.** Good. A test that asserts the JSON has `"restrictions":[]` for a plan with no restrictions would lock the contract; `TestSandboxPlanSnapshotFromPlanSortsRestrictionsAndTrimsRoot` covers the trimmed-and-sorted case but not the empty case.

19. **The PR title and description use the kebab-case "typed sandbox policy, risk, violation, and decision snapshots" phrasing.** The actual file is `sandbox_snapshots.go`. A future PR that adds more sandbox snapshot kinds (e.g., a `sandbox_event_snapshot.go` for audit events) would have to decide between a single file or a split. The current single file is fine for 202 lines; a split at ~500 lines would be the natural breakpoint.

20. **No test for `SandboxPlanSnapshotFromPlan` with nil `plan.Restrictions`.** The function would return a non-nil empty slice (via `make([]string, 0, 0)`) and serialize as `"restrictions":[]`. A test that feeds `sandbox.BackendPlan{}` and asserts `len(snapshot.Restrictions) == 0` and `string(json) == "[]"` would lock the contract. Cheap to add.

No blockers. The slice is well-scoped, well-tested, and closes the typed-snapshot gap for the live sandbox surface without changing runtime behavior. The two follow-ups I would prioritize next: (1) centralize the `EffectiveMode` default in a `sandbox.Policy.EffectiveMode()` method so the runtime default lives in one place, and (2) decide whether `Policy.Mode` should have `omitempty` (forcing consumers to read `EffectiveMode`) or stay as-is (preserving the raw value).

PR #93, opened 2026-06-06
feat(tui): surface sandbox permission events

Vasanthdev2004:
## Summary
- wire interactive TUI runs to the sandbox grant store and engine
- render live permission rows between tool calls and tool results
- persist and hydrate permission events in TUI sessions
- expand /permissions to show redacted persistent sandbox grant snapshots

## Validation
- go test ./internal/tui
- go test ./internal/cli
- go test ./...
- bun run typecheck
- bun run build

## Summary by CodeRabbit

* **New Features**
  * Sandbox grant support integrated into CLI and TUI; grants shown alongside permission mode.
  * Permission decisions persisted and displayed as transcript rows (with de-duplication and detailed rendering).
  * Tool call/result rows now stream to the UI as they occur.
  * Runtime message forwarding now ensures external sinks and the UI both receive runtime messages.

* **Tests**
  * Added/expanded tests covering sandbox grants, permission rows, and transcript hydration.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -277,6 +277,16 @@ func runInteractiveTUI(stderr io.Writer, deps appDeps) int {
 		return writeAppError(stderr, err.Error(), 1)
 	}
 	defer closeMCPRuntime(stderr, mcpRuntime)
+	sandboxStore, err := deps.newSandboxStore()
+	if err != nil {
+		return writeAppError(stderr, "failed to initialize sandbox grants: "+err.Error(), 1)
+	}
+	sandboxEngine := sandbox.NewEngine(sandbox.EngineOptions{
+		WorkspaceRoot: workspaceRoot,
+		Policy:        sandbox.DefaultPolicy(),
+		Store:         sandboxStore,
+		Backend:       deps.selectSandboxBackend(sandbox.BackendOptions{}),
+	})
 	permissionMode := agent.PermissionModeAuto
 	return deps.runTUI(context.Background(), tui.Options{
 		Cwd:             workspaceRoot,
@@ -287,10 +297,13 @@ func runInteractiveTUI(stderr io.Writer, deps appDeps) int {
 		NewProvider:     deps.newProvider,
 		Registry:        registry,
 		SessionStore:    deps.newSessionStore(),
+		SandboxStore:    sandboxStore,
 		AgentOptions: agent.Options{
 			MaxTurns:       resolved.MaxTurns,
 			Registry:       registry,
 			PermissionMode: permissionMode,
+			Autonomy:       string(sandbox.AutonomyLow),
+			Sandbox:        sandboxEngine,
 		},
 		PermissionMode: permissionMode,
 	})
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/Gitlawb/zero/internal/agent"
 	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/tools"
 	"github.com/Gitlawb/zero/internal/tui"
 	"github.com/Gitlawb/zero/internal/update"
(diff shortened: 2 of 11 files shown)


anandh8x (approved):
## What's good

- **Wiring is correct end-to-end.** `runInteractiveTUI` builds a single `sandboxStore` + `sandboxEngine`, passes `SandboxStore` to the TUI (for `/permissions`) and `Autonomy`/`Sandbox` to `agent.Options`. The agent loop in `internal/agent/loop.go` already has `buildPermissionEvent` and dispatches via `options.OnPermission`, so the new TUI `OnPermission` callback receives real events from the existing runtime — no shadow engine, no duplicate policy. The CLI test (`assertAgentOptions`) locks the wiring all the way through.

- **Live updates via `agentRowMsg`.** `sendAgentRow` sends an incremental message through the runtime sink so the UI updates between tool calls instead of only at `agent.Run` completion. The receiver filters on `m.activeRunID`, so stale rows from a prior run are dropped. `Run` wraps the external sink so both the external consumer (PR automation, audit log) and the Bubble Tea program see the same stream. `TestPromptSubmitPersistsPermissionSessionEvents` asserts the exact order: tool call → permission → result.

- **Dedup is correct and minimal.** `transcriptRowKey` returns a non-empty key only when the row has a stable identifier (tool call/result with `id`, or permission with `ToolCallID`). The key includes the row kind, so a permission row and a tool call row that happen to share an ID are not collapsed. Rows without an ID (assistant text, error) are never dedup'd — the right call, since the same text can legitimately repeat. `TestAppendTranscriptRowDedupesRuntimeRowsByID` covers the dedup path with two passes through tool call + permission + result collapsing to three unique rows.

- **Reasonable layering with `zerocommands.SandboxGrantSnapshots`.** `/permissions` goes through the typed snapshot package (PR #89/#90/#92), so the TUI's grant listing uses the exact same redaction and shape as `zero sandbox grants list --json`. No duplicated formatting logic. The new test feeds `"sk-proj-sensitive trusted shell"` and asserts `[REDACTED]` appears while the secret does not — confirming the redaction propagates.

- **Rich permission row rendering.** `renderPermissionRow` uses theme colors (green allow, red deny, amber prompt), shows risk level, grant badge, and a `key=value` detail line for violations. The title is a summary, the detail is the full breadcrumb; intentional and the test confirms both formats appear.

- **Session persistence round-trips.** `permissionEventFromPayload` reconstructs a `PermissionEvent` from JSON, including nested `Risk` and `Violation` deserialization, and tolerates `name`/`toolName` and `toolCallId`/`id` schema variants via `firstNonEmptyString`. The extended `TestResumeCommandHydratesSessionTranscript` confirms a persisted `EventPermission` hydrates as a `rowPermission` with the right `ToolName` and `Action`.

- **Callback chain is preserved.** `OnPermission` wraps the original option callback (`if onPermission != nil { onPermission(event) }`), so future test seams or instrumentation can observe the same events without being silently swallowed.

1. **`OnPermission` callback is not directly asserted in any test.** The contract is locked transitively via session persistence and the rendered row, but a focused test that injects a custom `OnPermission` and asserts it received the expected `PermissionEvent` would document the callback contract more sharply.

2. **`transcriptRowKey` is a useful seam that could use a doc comment.** Some rows are dedup'd (tool/permission) and some are not (text/error) — a one-line comment on why the empty-key case skips dedup would help the next maintainer.

3. **`permissionRowText` and `permissionDetailText` both include `risk:`/`violation:` lines.** Title is the summary, detail is the full record; intentional. A future consumer that wants to parse the detail line would parse `key=value` pairs — worth noting in a comment if `permissionDetailText` ever becomes machine-parseable.

4. **`transcriptRow.permission` is a shared pointer.** Safe today (no one mutates a `PermissionEvent` after the fact), but a short doc comment on the field would warn future maintainers against in-place mutation.

5. **`permissionsText` could grow to show the active policy and backend.** `zerocommands.SandboxPolicySnapshot` and `SandboxBackendSnapshot` (PR #92) are already typed and would fit naturally under `/permissions`. Worth a follow-up slice, not this one.

6. **`SandboxStore` is opened eagerly in `runInteractiveTUI` even if the user never runs an agent.** Fine for now (the store is also needed for `/permissions` and the engine); a future lazy-open on first use would shave a small startup cost, but the current code is correct.

7. **No benchmark for `appendTranscriptRow` / `permissionTranscriptRow`.** A long TUI session can produce hundreds of these per minute; a small benchmark would catch a future regression where someone adds an O(n) allocation per row.

8. **Style consistency on row kind labels.** `renderPermissionRow` uses `zeroTheme.amber.Render("permission")` for the kind label; other row kinds render the kind via the default text style. Not wrong, but a future cleanup could centralize row kind labels in one helper.

9. **Sender-side guard on `activeRunID` is missing; receiver-side guard is correct.** `m.sendAgentRow` does not check `m.activeRunID` before sending — the receiver drops stale rows. Cheaper to filter at the receiver (single channel send is cheap), and keeps the guard in one place.

Approving — the slice is well-scoped, well-tested, and the wiring is end-to-end correct. The live `agentRowMsg` updates are a real UX improvement, the dedup is right, the session persistence round-trips, and the cross-PR integration with the typed snapshots keeps the layering clean.

PR #95, opened 2026-06-06
feat: add runtime permission decisions

Vasanthdev2004:
## Summary
- add a blocking runtime permission request/decision contract for prompt-gated tools
- wire TUI allow / deny / always flows into the agent loop, including persistent sandbox grants
- split stream-json and session permission lifecycle into `permission_request` and `permission_decision`
- document the stream-json permission lifecycle and headless prompt behavior

## Tests
- git diff --check
- go test ./...
- go build ./cmd/zero

## Notes
- Subagents covered headless/session and TUI slices, then a final read-only review approved the combined diff.
- Headless exec still fails closed for prompt-gated tools unless already approved; interactive TUI now supplies decisions.

## Summary by CodeRabbit

* **New Features**
  * Implemented interactive permission prompts with allow/deny/always-allow options.
  * Added explicit `permission_request` and `permission_decision` events to the protocol.
  * Permission decisions now include reasoning details.
  * Introduced permission grant persistence for repeated tool access.

* **Documentation**
  * Updated Stream-JSON protocol examples with permission request/decision events.

### DIFF
--- a/docs/STREAM_JSON_PROTOCOL.md
+++ b/docs/STREAM_JSON_PROTOCOL.md
@@ -39,6 +39,8 @@ Unknown fields are rejected so protocol clients catch drift early.
 { "schemaVersion": 1, "type": "run_start", "runId": "run_20260603_abc123", "sessionId": "zero_20260603100000_abc123", "cwd": "/repo", "provider": "openai", "model": "gpt-4.1", "apiModel": "gpt-4.1" }
 { "schemaVersion": 1, "type": "text", "runId": "run_20260603_abc123", "delta": "..." }
 { "schemaVersion": 1, "type": "tool_call", "runId": "run_20260603_abc123", "id": "call_1", "name": "read_file", "args": { "path": "README.md" }, "sideEffect": "read" }
+{ "schemaVersion": 1, "type": "permission_request", "runId": "run_20260603_abc123", "id": "call_2", "name": "write_file", "action": "prompt", "permission": "prompt", "permissionMode": "ask", "sideEffect": "write", "reason": "Creates or overwrites files." }
+{ "schemaVersion": 1, "type": "permission_decision", "runId": "run_20260603_abc123", "id": "call_2", "name": "write_file", "action": "allow", "permission": "prompt", "permissionGranted": true, "decisionReason": "approved in TUI" }
 { "schemaVersion": 1, "type": "tool_result", "runId": "run_20260603_abc123", "id": "call_1", "status": "ok", "output": "...", "truncated": false }
 { "schemaVersion": 1, "type": "usage", "runId": "run_20260603_abc123", "promptTokens": 12, "completionTokens": 8, "totalTokens": 20 }
 { "schemaVersion": 1, "type": "final", "runId": "run_20260603_abc123", "text": "..." }
@@ -46,6 +48,11 @@ Unknown fields are rejected so protocol clients catch drift early.
 ```
 
 Errors are part of the protocol and are followed by `run_end`.
+
+Headless `exec` has no interactive permission responder. If a prompt-gated tool
+is not pre-approved, Zero may emit `permission_request` followed by a denied
+`tool_result`; interactive surfaces emit `permission_decision` when the user
+allows, denies, or always-allows the request.
 
 ```json
 { "schemaVersion": 1, "type": "error", "runId": "run_20260603_abc123", "code": "provider_error", "message": "...", "recoverable": false }
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"sort"
+	"strings"
 
 	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/tools"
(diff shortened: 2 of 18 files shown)


anandh8x (approved):
## What's good

- **Clean separation of `PermissionRequest` vs `PermissionEvent` vs `PermissionDecision`.** The request is the prompt context (args, risk, violation, side effect); the event is the post-decision record (action, decision reason, persisted grant); the decision is the responder's reply (action + reason). Three distinct types with three distinct roles — exactly what the protocol and the session store need.

- **`OnPermissionRequest` is a well-typed callback seam.** Signature `func(context.Context, PermissionRequest) (PermissionDecision, error)` is right: context for cancellation, request for context, decision for the result, error for responder failure. The TUI returns synchronously, the headless `exec` returns a deny-by-default, tests inject a fake. Clean.

- **Synchronous TUI decision via buffered channel.** The TUI's `OnPermissionRequest` wrapper creates a size-1 channel, sends a `permissionRequestMsg` to the runtime sink, and blocks on the channel until either a decision arrives or the context is canceled. The buffer size of 1 prevents the channel send from blocking if the agent runtime has given up. `case <-ctx.Done():` returns a deny with the cancellation error. The pattern is correct and leak-free (the wrapper owns the channel lifetime).

- **Default deny when sink is unavailable.** `if m.runtimeMessageSink == nil { return ...deny... }` ensures the agent runtime never blocks forever. The TUI always has a sink in practice, but the safety net is real.

- **`pendingPermission` is cleared on both `cancelRun` and `agentResponseMsg`.** No prompt state leaks across runs.

- **`KeyEnter` is blocked when pending, and all other keys route through `handlePermissionKey`.** The prompt is modal — only `a`/`d`/`y` (case-insensitive) are valid; everything else is a no-op. The `TestPermissionPromptBlocksNormalSubmit` test locks the Enter-blocking behavior.

- **Dedup key update allows both request and decision rows for the same `ToolCallID`.** Old key was `kind:ToolCallID`; new is `kind:ToolCallID:Action`. Without this, the decision row would be silently dropped. The change is correct because a prompt and an allow are different events.

- **Stream JSON protocol gets explicit event types.** `permission_request` and `permission_decision` are distinct from the generic `permission` (which is preserved for backward compatibility). Protocol clients can branch on the type. The docs example shows both new event shapes with `permission: "prompt"` and `action: "allow"`.

- **Headless behavior is documented.** The protocol doc explicitly states "Headless `exec` has no interactive permission responder. If a prompt-gated tool is not pre-approved, Zero may emit `permission_request` followed by a denied `tool_result`". Critical for automation consumers who would otherwise think the run is hung.

- **Session events mirror the protocol.** `EventPermissionRequest` and `EventPermissionDecision` join the existing `EventPermission`. The TUI's `tuiPermissionEventType` and the CLI's `sessionPermissionEventType` apply the same Action→EventType mapping. `transcriptRowsFromSessionEvents` hydrate handles all three types.

- **`shouldRequestPermission` correctly gates the prompt.** No request is made if the tool is not `PermissionPrompt`, or already granted (unsafe mode / grant match), or the sandbox preflight already decided (allow/deny). Only when the sandbox is silent AND the tool is a prompt tool AND the user hasn't granted it does the TUI get asked.

- **`normalizePermissionDecisionAction` defaults unknown actions to `deny`.** Safe default; the three valid actions are explicitly allowed.

- **`deniedPermissionResult` returns a `StatusError` tool result with a clear message.** The LLM sees "Error: Permission denied for write_file: <reason>" on the next turn and can adapt.

- **`OnPermission` is called with `DecisionReason` populated on the final event.** The TUI's `renderPermissionRow` shows the reason ("denied in TUI" / "approved in TUI" / "persistently approved in TUI"). Good.

- **`persistPermissionGrant` defaults `MaxAutonomy` to `Medium` if `options.Autonomy == ""`.** Reasonable default. The grant is bound to the current autonomy ceiling, which is correct semantics.

- **`Engine.Grant(input)` is a thin wrapper over `store.Grant(input)`.** Keeps the engine as the only entry point for grant operations. Tested transitively by `TestRunPersistsAlwaysAllowPermissionDecision`.

- **`TestPermissionPromptAlwaysPersistsGrantAndSkipsLaterPrompt` is the strongest test.** Full flow: prompt → `y` → grant persisted → tool runs → second turn → grant matches → no prompt → tool runs. The session events assert only the first run requested permission (`EventPermissionRequest` count = 1) but both recorded decisions (`EventPermissionDecision` count = 2). This is the user's mental model of "always allow" and the test locks the contract.

- **`receiveRuntimeMessage` / `receiveFinalMessage` test helpers use `time.After(5 * time.Second)`.** Prevents test deadlock on broken wiring. Defensive test pattern.

- **Callback chains are preserved.** `OnPermissionRequest` and `OnPermission` both wrap the previous option callback (`if onPermissionRequest != nil { onPermissionRequest(...) }` / `if onPermission != nil { onPermission(event) }`), so future test seams can observe the same events without being silently swallowed.

- **TUI prompt rendering uses theme colors.** Amber for "permission required", muted for details, text for choice labels. `renderFocusedPermissionPrompt` reuses the existing `borderedBlock` helper.

1. **Duplicated Action→EventType mapping in three places.** `sessionPermissionEventType` (CLI exec), `tuiPermissionEventType` (TUI), and `streamJSONPermissionEventType` (CLI stream JSON writer) all do the same mapping. A shared helper in `agent` (e.g., `agent.EventTypeForPermissionAction(action)`) and a parallel one for stream JSON would deduplicate. Not blocking.

2. **Headless `exec` has no `--non-interactive-permissions=allow|deny` flag.** The TUI's "always allow" path is unavailable in headless mode — prompt tools just get denied. Automation users would benefit from a CLI flag that pre-approves prompt tools (with the same grant persistence) or denies them explicitly. A follow-up slice, not this one.

3. **`renderFocusedPermissionPrompt` does not show the tool call args.** The TUI shows the reason and side effect, but not the path/content for `write_file` or the command for `bash`. `request.Args` is now available (added in this PR) but not rendered. A future TUI improvement could surface a detail block with the args for triage.

4. **`renderFocusedPermissionPrompt` is shown only when `m.pending` is true.** Defensive invariant: `pendingPermission != nil => pending == true`. The current code maintains this (the wrapper sets both), but a future refactor that toggles `pending` without clearing `pendingPermission` could break it.

5. **`Esc`/`Ctrl+C` cancel the prompt silently.** The TUI's existing cancel button works (clears `pendingPermission`), but the user doesn't see a confirmation that the decision was dropped. A small "permission request canceled" line in the transcript would help trust.

6. **`cloneArgs` is a shallow copy.** A nested map or slice in the args would be shared. For the current tool set (read_file, write_file, bash) the args are flat JSON values, so shallow copy is fine. A future tool with nested objects would need deep copy.

7. **`deniedPermissionResult` and `emitDeniedPermission` share the "set action to deny, set permissionGranted to false, set decision reason" pattern.** They could share a helper. Minor.

8. **`PermissionRequest.GrantMatched` and `Grant` fields are present but unused in the current request path.** The preflight decision either grants (skipping the request) or doesn't (request is built without grant context). The fields are there for completeness — a doc comment on the field would explain.

9. **No test for `Engine.Grant` directly.** Covered transitively by `TestRunPersistsAlwaysAllowPermissionDecision`. A focused unit test for `Engine.Grant` would assert the nil-engine and nil-store error paths.

10. **No test for `OnPermissionRequest` returning an error.** The loop converts an error into a deny, but no test exercises that path. A test that injects a callback returning `(PermissionDecision{}, errors.New("test"))` and asserts the deny is emitted would lock the safety net.

11. **Protocol naming is slightly asymmetric.** The `permission_request` example shows `permission: "prompt"` (the type) and the `permission_decision` example shows `action: "allow"` (the action). One field is the type, the other is the action. A doc note explaining the asymmetry would help consumers parsing the JSON.

12. **`TestPermissionPromptBlocksNormalSubmit` could also assert the input is preserved.** `m.input` still contains "second prompt" after the blocked Enter — confirming the user's typed text isn't lost. Minor tightening.

13. **`m.cancelRun` clears `pendingPermission` without calling the `decide` callback.** The pending decision is silently dropped. The agent runtime sees a context cancellation via the channel's `ctx.Done()` case. Consistent and correct, but worth a doc comment on `cancelRun` to explain the contract.

14. **No benchmark for the channel-based decision flow.** A microbenchmark would confirm the channel is fast enough for typical use. Not a concern in practice.

15. **`OnPermissionRequest` blocks the agent loop.** That's the design, but a slow responder would block the entire agent run. The TUI's channel pattern is correct; the `ctx.Done()` case handles cancellation. Worth noting for future async-responder work.

Approving — the slice is well-architected. The new types are clean, the callback seam is well-typed, the synchronous TUI decision via channel is correct, the dedup key update allows both request and decision rows, the stream JSON protocol gets explicit event types, the session events mirror the protocol, and the headless behavior is documented. The tests cover the full flow including the "always allow" path with grant persistence. The follow-ups are minor (duplicated event-type helpers, args display in TUI prompt, missing `--non-interactive-permissions` flag, error path tests).

PR #98, opened 2026-06-06
feat(tui): polish command output

Vasanthdev2004:
## Summary
- Adds reusable TUI command output formatting with status, sections, fields, rows, hints, and redaction.
- Groups slash-command help and drives startup chips from command metadata.
- Polishes `/model`, `/provider`, `/config`, `/context`, `/permissions`, `/resume`, `/compact`, `/style`, `/effort`, and debug/theme shell-state outputs.
- Moves command readout code out of `model.go` into focused command view helpers.

## Tests
- `go test ./internal/tui`
- `go test ./...`
- `go build ./cmd/zero`

## Review Notes
- `/provider` uses redacted provider snapshots so credential-bearing base URLs are not leaked.
- Startup remains minimal; no extra context/history/session panels are shown.

## Summary by CodeRabbit

* **New Features**
  * Commands now render as structured views with status indicators, grouped sections, hints, and compacted text
  * Help is organized into grouped command sections with stable ordering; startup command chips are deterministic
  * Session, model, compact, provider, context, permissions, and debug views show richer, clearer summaries; secrets/tokens are redacted

* **Tests**
  * Added broad tests validating command output formatting, status mappings, help grouping, startup ordering, and redaction behavior

### DIFF
--- a/internal/tui/command_center.go
+++ b/internal/tui/command_center.go
@@ -40,33 +40,59 @@ func (m model) searchText(query string) string {
 func (m model) resumeText(args string) string {
 	args = strings.TrimSpace(args)
 	if args != "" {
-		return strings.Join([]string{
-			"Sessions",
-			"TUI resume hydration is not wired yet.",
-			"Use zero exec --resume " + args + " for headless resume.",
-		}, "\n")
+		return renderCommandOutput(commandOutput{
+			Title:  "Sessions",
+			Status: commandStatusInfo,
+			Sections: []commandSection{{
+				Title: "Resume",
+				Lines: []string{"requested session: " + args},
+			}},
+			Hints: []string{"use /resume " + args + " to hydrate this TUI session"},
+		})
 	}
 	sessions, err := m.sessionStore.List()
 	if err != nil {
-		return "Sessions\nerror: " + err.Error()
+		return renderCommandOutput(commandOutput{
+			Title:  "Sessions",
+			Status: commandStatusBlocked,
+			Sections: []commandSection{{
+				Title: "Store",
+				Lines: []string{"error: " + err.Error()},
+			}},
+		})
 	}
-	lines := []string{"Sessions"}
 	if len(sessions) == 0 {
-		return "Sessions\n  (none)"
+		return renderCommandOutput(commandOutput{
+			Title:  "Sessions",
+			Status: commandStatusInfo,
+			Sections: []commandSection{{
+				Title: "Recent",
+				Lines: []string{"none"},
+			}},
+		})
 	}
 	limit := len(sessions)
 	if limit > 8 {
 		limit = 8
 	}
+	lines := []string{fmt.Sprintf("recent sessions: %d", len(sessions))}
 	for index := 0; index < limit; index++ {
 		session := sessions[index]
 		title := displayValue(session.Title, "untitled")
-		lines = append(lines, fmt.Sprintf("  %s  %s  model=%s provider=%s events=%d updated=%s", session.SessionID, title, displayValue(session.ModelID, "none"), displayValue(session.Provider, "none"), session.EventCount, session.UpdatedAt))
+		lines = append(lines, commandBullet(fmt.Sprintf("%s  %s  model=%s provider=%s events=%d updated=%s", session.SessionID, title, displayValue(session.ModelID, "none"), displayValue(session.Provider, "none"), session.EventCount, session.UpdatedAt)))
 	}
 	if len(sessions) > limit {
-		lines = append(lines, fmt.Sprintf("  ... %d more", len(sessions)-limit))
+		lines = append(lines, fmt.Sprintf("... %d more", len(sessions)-limit))
 	}
-	return strings.Join(lines, "\n")
+	return renderCommandOutput(commandOutput{
+		Title:  "Sessions",
+		Status: commandStatusOK,
+		Sections: []commandSection{{
+			Title: "Recent",
+			Lines: lines,
+		}},
+		Hints: []string{"use /resume latest or /resume <id> to load a session"},
+	})
 }
 
 func (m model) handleModelCommand(args string) (model, string) {
--- a/internal/tui/command_output.go
+++ b/internal/tui/command_output.go
@@ -1,0 +1,136 @@
+package tui
+
+import (
+	"regexp"
+	"strings"
+)
+
+var commandOutputSecretPatterns = []*regexp.Regexp{
+	regexp.MustCompile(`\bsk-[A-Za-z0-9._-]+\b`),
+}
+
+type commandStatus string
+
+const (
+	commandStatusOK      commandStatus = "ok"
+	commandStatusWarning commandStatus = "warning"
+	commandStatusBlocked commandStatus = "blocked"
+	commandStatusInfo    commandStatus = "info"
+)
+
+type commandOutput struct {
+	Title    string
+	Status   commandStatus
+	Sections []commandSection
+	Hints    []string
+}
+
+type commandSection struct {
+	Title  string
+	Fields []commandField
+	Lines  []string
+	Rows   []commandRow
+	Hints  []string
+}
+
+type commandField struct {
+	Key   string
+	Value string
+}
+
+type commandRow struct {
+	Text string
+}
+
+func formatCommandOutput(output commandOutput) string {
+	lines := []string{}
+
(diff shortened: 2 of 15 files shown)


anandh8x (approved):
## What's good

- **Unified `commandOutput` framework reduces duplication.** Every TUI command output is now a `commandOutput` struct (`Title`, `Status`, `Sections`, `Hints`) rendered by `formatCommandOutput`. Adding a new command is just filling in the struct — no more `strings.Join([]string{...}, "\n")` sprawl across `model.go`, `session_controls.go`, `command_center.go`, and friends.

- **`commandStatus` enum is precise.** `ok` / `warning` / `blocked` / `info` covers the real semantic states. `normalizeCommandStatus` defaults unknown to `info` so a future status addition doesn't break rendering. The data is in place for a future enhancement to theme the status line per value.

- **Secret redaction is centralized and tested.** `\bsk-[A-Za-z0-9._-]+\b` in `commandOutputSecretPatterns` catches `sk-...` tokens. Applied in `compactCommandOutputText` which is called for every text segment (title, field key/value, line, row, hint). The test feeds `"sk-sensitive approved shell"` and asserts it becomes `"[REDACTED] approved shell"`. Defense-in-depth: the `internal/redaction` package is the primary layer; this is a fallback for tokens that escaped.

- **Whitespace compaction and empty-skipping are consistent.** `strings.Join(strings.Fields(text), " ")` collapses all whitespace. The test `"  use /doctor for details\r\nsoon  "` → `"use /doctor for details soon"` locks the behavior. Empty/whitespace-only fields, lines, and rows are dropped — so a nil sections list renders as just the title + status line, not a tree of blanks.

- **Grouping `/help` output is a real UX win.** The 5 groups (`Model`, `Session`, `Runtime`, `Tools`, `Meta`) make the help scannable. `commandGroupOrder()` is the single source of truth for the group order. `TestFormatCommandHelpLinesGroupsCommandsByStableOrder` asserts the order is preserved end-to-end, and `TestHelpCommandRendersGroupedSections` asserts the new `/help` shape (with the "submit plain text" hint replacing the old "Submit text to ask the assistant").

- **`startupOrder` metadata replaces hard-coded list.** The previous `[]string{"/plan", "/debug", "/tools", "/model", "/provider"}` is now driven by `commandDefinition.startupOrder`. Adding a new startup chip is one field. `TestCommandDefinitionsExposeStartupChipsInStableOrder` uses reflection to verify the chip order matches the metadata, catching future field renames.

- **Cross-PR integration with `zerocommands` is clean.** `/provider` uses `zerocommands.ProviderSnapshotFromProfile` (PR #82/#90), so URL userinfo and `api_key=` query param redaction happens at the snapshot layer. `/permissions` still uses `zerocommands.SandboxGrantSnapshots` (PR #89), so grant reason redaction is inherited. `TestProviderCommandRedactsCredentialBearingBaseURL` confirms `https://user:super-secret@proxy.local/v1?api_key=query-secret` → `https://proxy.local/v1?api_key=[REDACTED]`. Good layering — the TUI doesn't re-implement snapshot logic.

- **Removing "shell-only" placeholder text is a real product-quality improvement.** The old `"X is registered in the Go TUI shell but is not wired yet"` was honest but unhelpful. The new `"This control is available in the TUI but does not have a backend setting yet"` is still honest, more specific, and less jarring. `TestCompactCommandAvoidsShellOnlyPlaceholder` and `TestCompactCommandRecordsShellRequest` both lock the new strings and assert the old placeholders are gone.

- **`compactText` now describes product state.** Old output: `"Compact\nstatus: ...\nbackend: not wired yet"` or `"request recorded for future compaction backend."`. New output: structured `State` + `Backend` sections, `"summary: ..."`, `"requested: yes/no"`, `"state: pending integration"`, and a hint `"compaction request is tracked for this TUI session"`. The `boolText` helper gives a clean `requested: yes` rather than a free-form phrase.

- **Redaction is applied uniformly via `compactCommandOutputText`.** Every text segment (title, field, line, row, hint) goes through the same compaction + redaction path. A future secret format addition is one regex entry, and the test pattern (`TestFormatCommandOutputRedactsTokenLikeText`) generalizes to any token pattern.

- **`transcriptHasMarkedModelEntry` test fix is correct.** The previous check `strings.HasPrefix(line, "* ")` would fail on the new 2-space indented output. The fix trims first. Defensive test pattern.

- **Test coverage is comprehensive.** New `command_output_test.go` (5 tests), `commands_test.go` (2 tests with reflection), and `command_polish_test.go` (5 tests for product-state output). Existing `TestCompactCommandRecordsShellRequest` updated to lock the new strings.

1. **Redaction pattern is narrow.** Only `sk-...` is matched. Other formats (AWS, GitHub `ghp_...`, JWTs, etc.) are not caught. The `internal/redaction` package has broader patterns; a future enhancement could call it directly from `compactCommandOutputText` or expand the regex list.

2. **Status is not visually distinguished in the rendered output.** `formatCommandOutput` emits `status: <value>` as a plain text line. A future enhancement could theme the status line (red for `blocked`, amber for `warning`, green for `ok`). The data is there; the styling is not.

3. **Redundant lines in some command outputs.** `/permissions` has both `"Permission mode:"` and `"mode:"`. `/provider` has both `"active:"` and `"provider:"`. `/model` has both `"Active model:"` and `"model:"`. `/context` has both `"tools:"` and `"registered tools:"`. A cleanup pass could drop the duplicate. Minor.

4. **A tool name or command starting with `sk-` would be mis-redacted.** The `\b` word boundary helps, but `sk-looker-integration` would also be redacted. Unlikely in practice, but worth noting for future edge cases.

5. **`shellOnlyCommandText` is no longer "shell-only".** The function name is now misleading. A rename to `placeholderCommandText` or `pendingCommandText` would be clearer.

6. **No test for nil sections / nil fields slices in `formatCommandOutput`.** The existing tests cover empty/whitespace values but not nil slices. A future test could lock the nil-handling behavior.

7. **No benchmark for `formatCommandOutput`.** The regex is compiled at package init (compile-once), so the hot path is fast. A small benchmark would catch future regressions where someone adds an O(n) string allocation per call.

8. **`/compact` still says "state: pending integration"** — itself a placeholder. The test name `TestCompactCommandAvoidsShellOnlyPlaceholder` is accurate (it asserts the old text is gone), but the new text is still a placeholder. A future slice could implement compaction and replace it with real state.

9. **No test for `startupOrder` ties.** `startupCommandNames` uses `sort.SliceStable` with a secondary sort on `name` to break ties. The test only exercises all-unique orders. A test with two commands sharing `startupOrder` would lock the tie-breaker.

10. **`formatGroupedCommandHelpLines` is only used in one place** (as the body of `formatCommandHelpLines`). The two functions are redundant — `formatCommandHelpLines` could be deleted and callers updated. Minor.

11. **`commands_test.go` reflection test is a meta-test.** It verifies the `startupOrder` field exists and is `int`. A simpler test that asserts the chip order would suffice, but the reflection version catches field renames. Defensive.

12. **No test for `commandHelpLinesForGroup` with an empty group.** If a group has no commands, the lines are not emitted (handled by the `len(groupLines) == 0` check). A test would lock the behavior.

13. **The `\b` word boundary in the redaction regex requires the token to be preceded and followed by a non-word character.** That works for `sk-foo bar` but not for `prefixsk-foo`. The current usage is fine (tokens are usually standalone), but a future edge case to consider.

14. **`commandGroupTitle` returns `string(group)` as the default.** For an unknown group, the rendered title would be `"commandGroup99"`-style. The test doesn't exercise the unknown-group case. A future test could assert the default behavior.

15. **`/plan` description is "Show planning mode status."** Accurate but doesn't explain what planning mode is. A hint like "use /plan to view or toggle planning mode" would help discoverability.

16. **No test for the `formatCommandSection` empty sections case.** A section with no fields, lines, rows, or hints renders as just the title. A test would lock the behavior.

17. **The `formatCommandOutput` always emits a `status:` line, even if `status` is empty (which `normalizeCommandStatus` converts to `info`).** A future optimization could omit the line if status is `info` and no sections are present. Not a concern at the current scale.

Approving — the slice is well-executed. The unified `commandOutput` framework reduces duplication, the redaction layer is a useful safety net, the help grouping is a real UX win, the startup chip metadata replaces a hard-coded list, the cross-PR integration with `zerocommands` is clean, and the "shell-only" placeholder text is replaced with softer language. The tests are comprehensive. The follow-ups are minor (narrow redaction pattern, status theming, redundant lines in some command outputs, function rename for `shellOnlyCommandText`).

PR #100, opened 2026-06-06
feat(specialist): add profile management

anandh8x:
## Summary

- Add specialist manifest parsing and validation
- Add built-in `worker`, `explorer`, and `code-review` profiles
- Add `zero specialist list/show/path` with JSON output

## Self-Review

- This intentionally does not add the runtime `task` tool or child-process execution yet.
- Specialist paths follow existing Zero conventions: user profiles under `~/.config/zero/specialists`, project profiles under `.zero/specialists`.
- Tool names match the current Go registry IDs, such as `read_file`, `grep`, and `apply_patch`.

## Validation

- `GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added `specialist` (and alias `specialists`) CLI with `list`, `show <name>`, and `path` subcommands to inspect local specialist profiles
  * Introduced three built-in specialists: worker, explorer, and code-review
  * Added `--json` output option for machine-readable results

* **Tests**
  * Added end-to-end and parsing tests covering specialist commands, JSON output, and manifest handling

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -158,6 +158,8 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 		return runSearch(args[1:], stdout, stderr, deps)
 	case "sessions", "session":
 		return runSessions(args[1:], stdout, stderr, deps)
+	case "specialists", "specialist":
+		return runSpecialists(args[1:], stdout, stderr, deps)
 	case "plugins", "plugin":
 		return runPlugins(args[1:], stdout, stderr, deps)
 	case "hooks":
@@ -359,6 +361,7 @@ Commands:
   search     Search persisted local Zero session events
   find       Alias for search
   sessions   Inspect local Zero session lineage
+  specialist Inspect local Zero specialist profiles
   plugins    Inspect local Zero plugin manifests
   hooks      Inspect Zero hook configuration
   mcp        Manage MCP backend settings
--- a/internal/cli/specialist.go
+++ b/internal/cli/specialist.go
@@ -1,0 +1,154 @@
+package cli
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/specialist"
+)
+
+type specialistOptions struct {
+	json bool
+}
+
+func runSpecialists(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
+	command, remaining, options, help, err := parseSpecialistArgs(args)
+	if err != nil {
+		return writeExecUsageError(stderr, err.Error())
+	}
+	if help {
+		if err := writeSpecialistHelp(stdout); err != nil {
+			return exitCrash
+		}
+		return exitSuccess
+	}
+
+	workspaceRoot, err := resolveWorkspaceRoot("", deps)
+	if err != nil {
+		return writeExecUsageError(stderr, err.Error())
+	}
+	paths, err := specialist.DefaultPaths(workspaceRoot)
+	if err != nil {
+		return writeAppError(stderr, err.Error(), exitCrash)
+	}
+
+	switch command {
+	case "list":
+		if len(remaining) != 0 {
+			return writeExecUsageError(stderr, "specialist list does not accept positional arguments")
+		}
+		return runSpecialistList(paths, options, stdout, stderr)
+	case "show":
+		if len(remaining) != 1 {
+			return writeExecUsageError(stderr, "specialist show requires a specialist name")
+		}
+		return runSpecialistShow(paths, remaining[0], options, stdout, stderr)
+	case "path":
+		if len(remaining) != 0 {
+			return writeExecUsageError(stderr, "specialist path does not accept positional arguments")
+		}
+		return runSpecialistPath(paths, options, stdout)
+	default:
+		return writeExecUsageError(stderr, fmt.Sprintf("unknown specialist command %q", command))
+	}
+}
+
+func parseSpecialistArgs(args []string) (string, []string, specialistOptions, bool, error) {
+	command := "list"
+	commandExplicit := false
+	remaining := []string{}
+	options := specialistOptions{}
+	for _, arg := range args {
+		switch arg {
+		case "-h", "--help", "help":
+			return command, remaining, options, true, nil
+		case "--json":
+			options.json = true
+		default:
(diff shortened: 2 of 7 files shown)


gnanam1990 (changes requested):
## Review — verified, multi-dimension

Nice work — this is a clean, well-structured package and the security posture on the parts that matter is genuinely strong. Builds green, `vet` clean, tests pass. I ran a verified review across 5 dimensions (manifest correctness, path/traversal security, CLI+wiring consistency, built-in/tool-name validity, tests), with each finding adversarially re-checked. Net: **1 high, 8 medium, 13 low/nit**, plus several raised-and-refuted. Requesting changes on the high + a few mediums; the rest is polish.

### What's done well 👍
- **No path traversal** — `show NAME` never builds a path from input; it loads the fixed dirs and matches in-memory against the validated `Name`, so `show ../../etc/passwd` just returns "not found". Dirs anchored to `os.UserConfigDir()/zero/specialists` and `Clean(workspaceRoot)/.zero/specialists`.
- **Strict tool allowlist** (default-deny on unknown names) + tight name regex `^[a-z][a-z0-9-]{0,63}$`; built-ins validated at startup (panic on bad). Hand-rolled frontmatter parser → no YAML/eval injection surface.
- Correct builtin→project→user merge precedence with defensive copies; consistent exit codes (usage=2/crash=1); unknown subcommand fails before touching the FS; built-in profiles reference only real registry tool IDs.

### 🔴 High (please address)
- **`reasoningEffort` is accepted but never validated** — `internal/specialist/manifest.go:194-214`. `modelregistry.ValidReasoningEffort()` already enumerates the valid values, but it's not called, so `reasoningEffort: ULTRA` parses clean and silently misconfigures the specialist. Validate it (error or warning) in `Validate()`.

### 🟠 Medium
- **`model` field is also unvalidated** (`manifest.go:194-214`) — `model: fake-9000` is accepted silently though the model registry has `ResolveID()/Get()`. Worth doing together with `reasoningEffort`.
- **Default tool grant is "all"** when `tools:` is omitted (`manifest.go:216-219`, rendered as "all"). For something feeding specialists/subagents this is a default-**allow** posture — please decide whether project profiles should default to read-only, and at minimum document it.
- **Tool allowlist is a hand-copied duplicate** of the registry (`manifest.go:76-93`) with no coupling to `internal/tools` — it will silently drift when tools change. Add a guard test asserting it matches the real registry IDs (or derive it).
- **Duplicate frontmatter keys silently overwrite** (last-wins, no warning) — `parseFrontmatter`. Surface it.
- **Test gaps on the very paths that justify the feature**: unknown-tool rejection, invalid/traversal-style names, malformed frontmatter / empty prompt, one-bad-file `Load` behavior, and `show/path --json` + arg-count errors are all *implemented but unasserted*. The code is good; the tests under-cover the security/validation edges.

### 🟡 Low / nit (non-blocking)
- **Robustness:** one malformed `.md` aborts the entire `Load` (breaks `list`/`show`, including builtins) — consider skip-and-warn; `.md` **symlinks are followed** by `os.ReadFile` (`manifest.go:402-410`) — skip symlinks or `EvalSymlinks`-and-contain; inline list `["a", "b,c"]` splits inside quotes → corrupt token (`parseInlineList`); a couple of misleading parse errors ("must be an array" for an unterminated `[`).
- **Consistency:** output isn't run through `redaction.*` though every sibling command (`sessions`/`plugins`/`skills`) is; precedence is **user > project** (inverse of the usual project-wins) — intended? add a doc comment; a specialist literally named `help` can't be `show`n.
- **Nits:** bare `zero specialist` defaults to list (fine — matches `sessions`/`models`, but diverges from `plugins`/`hooks`); a dead raw-category branch in `formatTools`.

### ✅ Refuted (checked, not issues)
Unbounded description/prompt length (trimmed, low-risk); description→prompt fallback ordering (intended); help text omitting `--json`/`specialists` alias (consistent house style); `task`/`skill`/`ask_user` "missing" from the allowlist (correct — excluded by design); flag/command ordering (already covered by tests).

### Suggested before merge
1. Validate `reasoningEffort` **and** `model` against the registry.
2. Decide + document the **default-allow tools** behavior; add the **allowlist drift-guard test**.
3. Harden `Load`: skip-and-warn on a single bad file; skip symlinks.
4. Add the missing **validation/security tests** (unknown tool, bad name, malformed frontmatter, `show/path --json`, arg errors).

Everything else is optional polish. Thanks for the careful self-review and the conventions write-up in the description — made this easy to review.

<sub>🤖 Verified multi-agent review (5 dimensions → adversarial verification) via Claude Code.</sub>

gnanam1990 (approved):
## Re-review — all requested changes addressed ✅

Thanks for the thorough turnaround — every point from the previous review is fixed, and (nicely) **with test coverage**:

- **[High] `reasoningEffort`** now validated via `modelregistry.ValidReasoningEffort` (errors on unknown) + `TestParseMarkdownValidatesModelAndReasoningEffort`.
- **[Med] `model`** validated via `DefaultRegistry().ResolveID` + test asserting "unknown model".
- **[Med] allowlist drift-guard** — `TestKnownToolNamesMatchCoreRegistry` compares `knownToolNames` against `tools.CoreTools`, so the hand-maintained set can't silently drift from the real registry. 👌
- **[Med/low] `Load` robustness + symlink security** — `loadDirectory` now skips symlinked and unreadable/invalid manifests with warnings instead of aborting the whole load (fixes both the one-bad-file-breaks-everything and the symlink-follow concerns) + `TestLoadSkipsBadFilesAndSymlinksWithWarnings`.
- **Inline-list comma bug, duplicate frontmatter keys, unknown tools, invalid/traversal names, `show/path --json`, arg-count errors** — all now covered by tests.

Build + `internal/specialist`/`internal/cli` tests pass.

### Approving. Two non-blocking nits for later (no need to hold the PR):
- `Validate()` calls `modelregistry.DefaultRegistry()` per manifest-with-a-model, rebuilding the full catalog (regex compile) each time. Negligible at current scale; worth caching/injecting if specialist counts grow.
- The "no `tools:` key ⇒ all tools" default-allow is fine for `worker`, but when the runtime `task` execution eventually consumes these profiles, consider defaulting *project* profiles to read-only (default-deny) and/or a short doc comment on the behavior.

Nice work. 🚀

<sub>🤖 Verified re-review via Claude Code.</sub>

PR #101, opened 2026-06-06
Runtime-core: multi-model tool tolerance, failure guard, rich TUI, subagents, plugins/skills

gnanam1990:
## Summary

Hardens the Zero Go runtime-core for a diverse multi-model userbase and lands the high-value blueprint features. Scope is the runtime/TUI (`internal/agent`, `internal/tools`, `internal/zenline`, `internal/tui`, `internal/plugins`, `internal/skills`, `internal/cli`).

## Reliability & weak-model tolerance
- **Shared tool-arg tolerance layer** (`internal/tools/argtolerance.go`): every tool accepts the key/shape variants any model emits — `path`↔`file`/`file_path`, `content`↔`contents`/`text`/`body`, `old_string`↔`old`/`search`, `new_string`↔`new`/`replace`, `patch`↔`diff`, `pattern`↔`query`/`regex`, `command`↔`cmd`/`script`, plus list/dir/glob aliases. Type-strictness preserved (a genuinely wrong type still errors).
- **bool/int coercion**: `boolArg` accepts `true/false/yes/no/on/off/1/0` (string or number); `intArg` accepts whole-number strings. Fixes the live `overwrite must be a boolean` cycling.
- **Repeated-tool-failure guard** (`internal/agent/guardrails.go`): after 2 identical failures the model gets a one-shot corrective hint (tool schema + exact error); after 4 the run halts with a clear message instead of burning turns to `maxTurns`.

## TUI / UX
- **Mouse capture removed** so the terminal handles native click / select / copy / paste (it was breaking copy-paste).
- **Full-bleed themed background** on home/boot (no terminal-bg "card"; chat status bars unaffected).
- **Rich rendering**: assistant text via `glamour` (CommonMark) + `chroma` syntax-highlighted code; colored, guttered unified diffs for `edit_file`/`apply_patch`. Output cached per (text, theme, width) so `RenderChat` doesn't re-render every frame.
- **Slash-command autocomplete** + interactive `model`/`theme`/`effort`/`mode` pickers (both skins; existing permission/ask_user/key precedence untouched).

## Capabilities
- **Subagents** (`task` tool): spawns an isolated child `agent.Run` with `Depth+1`, a registry filtered via `Registry.Without("task","ask_user")`, inherited sandbox/permission but no interactive callbacks; `maxTaskDepth=2`.
- **Plugins + skills**: optional plugin-manifest metadata (author/license/keywords/interface); new `internal/skills` package loads `*/SKILL.md` (hand-parsed frontmatter, no YAML dep), a read-only `skill` tool, and a `zero skills` command.

## Testing
- `go build`, `go vet`, `go test ./...`, and `go test -race ./...` all green.
- End-to-end verified through the real `zero exec` agent loop (with a mock OpenAI provider, since the live model was quota-limited): weak-model `write_file` args (`text` alias + `overwrite:"true"`) accepted and file overwritten; `task` delegation runs a child and returns; the failure guard halts at exactly 4 identical errors; `zero skills`/`--list-tools`/`models` and themed markdown+diff rendering confirmed.

## Dependencies
- Adds `github.com/charmbracelet/glamour` + `github.com/alecthomas/chroma/v2` for markdown/highlighting. Binary grows to ~22 MB.

## Deferred (follow-ups)
- Async `task` output/stop, collapsible/typed tool-row interactivity, multi-screen layout, full 4-action permission engine, reasoning-effort propagation to the provider request.

## Summary by CodeRabbit

* **New Features**
  * Model presets and new --mode exec presets; conversation compaction; session checkpoint & rewind; interactive "ask user" flow; skills discovery and new skills/zenline CLI commands; Zenline TUI skin with themes, autocomplete and pickers; sub-agent/task handling and safer tool execution.

* **Bug Fixes**
  * Improved dropped-tool detection/retries, streaming idle-timeout handling, secret redaction, interactive-command blocking, and tool-result metadata (redaction/changed-files) plus pre-tool checkpointing.

* **Documentation**
  * Added agent confirmation policy guidance.


gnanam1990:
@coderabbitai full review

This branch now also contains the audit fixes (previously stacked in #103) and the CodeRabbit review fixes:
- modelregistry: validate `reasoningEffort`∈ReasoningEfforts + `DefaultReasoningEffort` membership; `NewRegistry` verifies every deprecation `FallbackID` resolves
- sessions: rewind applies only the closest-to-target checkpoint per path (+ symlink-safe path guard, sha256 blob verify, per-session flock)
- agent: redact intercepted ask_user/task outputs; reactive retry keeps OnUsage but drops OnText
- tui: shift+tab toggles Auto↔Ask only (Unsafe not a casual keypress); multi-cancel checkpoint flush
- sandbox: alias-aware destructive-command gate
- removed the accidentally-committed shop.html + gitignored it

Full `go build`/`vet`/`go test -race ./...` are green. Please re-review.

gnanam1990:
@coderabbitai full review

Pushed a second round of fixes (6 commits, `a9655a6`..`2ee03ac`) addressing a fresh deep audit of this branch — **40 findings fixed** (4 high in tools: grep symlink confinement + clean rel-paths, apply_patch workspace-relative ChangedFiles, aliasedStringArg empty-alias masking; 2 high in sessions/sandbox: Fork copies checkpoint blobs, quoted-root rm -rf gate; plus mediums/lows across providers/runtime/zenline/tui/mcp/skills/cli). 1 finding rejected (AgentEventType is an intended contract) and `--profile` kept as intentional legacy-compat. Full `go build`/`vet`/`go test -race ./...` green. Please re-review.

PR #104, opened 2026-06-07
feat(specialist): add exec metadata flags

anandh8x:
Summary:
- adds specialist-oriented zero exec metadata flags: calling session/tool ids, tag, depth, session title, and init session id
- propagates the metadata into agent.Options for later specialist runtime slices
- lets --init-session-id preallocate a new exec session id and accepts debug as a stream-json output alias

Scope:
- does not add the Task tool or runtime specialist spawning yet
- does not depend on PR #101

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/cli ./internal/agent
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added `--calling-session-id`, `--calling-tool-use-id`, `--tag`, `--depth`, `--session-title`, and `--init-session-id` to `zero exec`; session titles can be set via `--session-title`.
  * `--output-format debug` added as an alias for JSON streaming.

* **Documentation**
  * Updated `zero exec` help text to document the new flags and aliases.

* **Bug Fixes / Validation**
  * New validation for `--depth` (non-negative) and `--init-session-id` usage (only when creating/forking).

* **Tests**
  * Added tests covering new flags, validations, and session initialization.

### DIFF
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -82,7 +82,14 @@ type PermissionEvent struct {
 }
 
 type Options struct {
-	MaxTurns            int
+	MaxTurns int
+	// Specialist/sub-agent metadata is carried through exec now and consumed by
+	// the specialist runtime in later slices.
+	CallingSessionID    string
+	CallingToolUseID    string
+	Tag                 string
+	Depth               int
+	SessionTitle        string
 	Registry            *tools.Registry
 	PermissionMode      PermissionMode
 	Autonomy            string
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -412,9 +412,16 @@ Flags:
                                     Select prompt input format
   -o, --output-format text|json|stream-json
                                     Select text, JSON, or schema-versioned JSONL output
+                                    ("debug" is accepted as a stream-json alias)
       --prompt <prompt>              Provide prompt text as a flag
       --resume [id]                  Resume a session; omit id to use the latest
       --fork <id>                    Fork an existing session into a new session
+      --calling-session-id <id>      Parent session id for specialist child runs
+      --calling-tool-use-id <id>     Parent tool-call id for specialist child runs
+      --tag <tag>                    Attach runtime tag metadata to the exec run
+      --depth <number>               Set specialist nesting depth metadata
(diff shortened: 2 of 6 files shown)


gnanam1990 (approved):
## Review — clean, approving with a couple of notes

Nice, focused PR — the flag parsing is consistent (both `--flag value` and `--flag=value` forms, shared `requiredInlineFlagValue`/`nextFlagValue` helpers), validation is solid (`--depth` rejects non-int/negative, `--init-session-id` validated via `sessions.ValidSessionID`, sensible conflict checks), `execSessionTitle`/`shouldUseExecSession` are correct, and it's well covered by `TestParseExecSpecialistMetadataFlags` / `...RejectsInvalidValues` / `TestRunExecUsesInitSessionIDAndSessionTitle`. Build + `go test ./internal/cli ./internal/agent` green. Approving.

### Non-blocking notes

1. **`debug` alias missing from the error hint** (`exec_parse.go:377`, as CodeRabbit noted) — `parseExecOutputFormat` accepts `debug` but the invalid-format error still says "Expected text, json, or stream-json." One-word fix: add `, or debug` so the guidance matches behavior.

2. **⚠️ `agent.Options.Depth` coordination** — heads-up from working in the adjacent area: the upcoming **sub-agent (`task`) runtime** also introduces `agent.Options.Depth` with the *same* meaning ("sub-agent nesting level; child = Depth+1; refuse at/above maxTaskDepth"). This PR adds `Depth` (plus `CallingSessionID`/`CallingToolUseID`) as exec metadata. They're the same concept — worth syncing with whoever owns the task-runtime slice so you converge on **one** `Depth` field with agreed semantics, otherwise the second PR to land hits a duplicate-field/merge collision. Same spirit for the other calling-* metadata.

3. **Plumbed-but-not-yet-consumed** — `CallingSessionID`, `CallingToolUseID`, `Tag`, `SessionTitle`, `Depth` are set on `agent.Options` but nothing in `internal/agent` reads them yet (matches your "later slices" note). Fine as staging; just flagging so it's a conscious choice — ideally these land with their consumer (or keep the doc comment so they're not mistaken for dead fields).

None of these block — they're a one-liner + coordination. 👍

PR #105, opened 2026-06-07
Streaming reliability: SSE idle-watchdog, dropped-call handling, finish_reason, usage

gnanam1990:
## Module 1 of N — Streaming reliability

First reviewable slice of the runtime-core work, split out of the large integration branch (#101, now draft) so the team can review module-by-module. **Scope: `internal/zeroruntime` + `internal/providers` only.** Fully backward-compatible/additive — the rest of the tree builds unchanged against it.

### What's in it
- **SSE idle-timeout watchdog** across OpenAI / Anthropic / Gemini (reader goroutine + idle timer + ctx-cancel, via `providerio.ScanSSEDataWithContext`) — a stalled-but-open upstream no longer blocks the agent forever.
- **Hardened streamed tool-call assembly**: nameless/dropped tool calls are reported (`StreamEventToolCallDropped`) instead of silently lost; collected calls keep model/start order; distinct empty-id calls no longer merge.
- **Terminal finish/stop reason surfaced** (`CollectedStream.FinishReason` + `Truncated()`) so `length`/`content_filter` responses aren't mistaken for normal completions.
- **OpenAI**: requests `stream_options.include_usage` (token usage no longer silently 0); delegates error/redaction to `providerio` (503/529 classified as rate-limit); joins multi-line SSE `data:` continuations; wires `max_completion_tokens`.

### Testing
`go build ./...`, `go vet ./...`, `go test ./...`, and `go test -race ./internal/zeroruntime/ ./internal/providers/...` all green. No new dependencies; `go.mod` unchanged.

### Context
Part of decomposing #101 into reviewable PRs. Subsequent modules (model registry, sandbox, sessions, tools, agent loop, TUI, extensions) will follow one at a time after this merges. The subagent/`task` feature is intentionally **excluded** (separate developer's area).

## Summary by CodeRabbit

* **New Features**
  * Streaming idle-timeout watchdog across providers
  * Finish-reason propagation on completion events (e.g., length/content-filter)
  * Emission of a dropped-tool-call event for malformed tool/function calls
  * Configurable max-token limit for OpenAI-compatible provider

* **Bug Fixes**
  * Improved stalled-stream detection and termination
  * Retry/backoff for transient OpenAI network errors
  * More robust handling of incomplete/duplicate tool-call fragments

* **Tests**
  * Extensive tests for idle timeouts, finish reasons, and tool-call behaviors

### DIFF
--- a/internal/providers/anthropic/dropped_test.go
+++ b/internal/providers/anthropic/dropped_test.go
@@ -1,0 +1,66 @@
+package anthropic
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// A tool_use content_block_start that arrives without a usable name/id can't be
+// dispatched. The provider must signal a dropped tool call (once) so the agent
+// can ask the model to retry, mirroring the OpenAI provider's behavior, instead
+// of silently dropping it.
+func TestStreamCompletionEmitsDroppedOnNamelessToolUseBlock(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		writeSSEEvent(w, "content_block_start", `{"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"","name":""}}`)
+		writeSSEEvent(w, "content_block_stop", `{"type":"content_block_stop","index":0}`)
+		writeSSEEvent(w, "message_stop", `{"type":"message_stop"}`)
+	}))
+	defer server.Close()
+
+	provider, err := New(Options{BaseURL: server.URL + "/", Model: "claude-test"})
+	if err != nil {
+		t.Fatalf("New returned error: %v", err)
+	}
+	events := collectProviderEvents(t, provider)
+
+	var dropped, started int
+	for _, e := range events {
+		switch e.Type {
+		case zeroruntime.StreamEventToolCallDropped:
+			dropped++
+		case zeroruntime.StreamEventToolCallStart:
+			started++
+		}
+	}
+	if started != 0 {
+		t.Errorf("a nameless tool_use block must not start a tool call, got %d starts", started)
+	}
+	if dropped != 1 {
+		t.Errorf("expected exactly one dropped-tool-call signal, got %d; events: %+v", dropped, events)
+	}
+}
+
+// A well-formed tool_use block must NOT emit a dropped signal.
+func TestStreamCompletionDoesNotDropValidToolUseBlock(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		writeSSEEvent(w, "content_block_start", `{"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"toolu_1","name":"read_file"}}`)
+		writeSSEEvent(w, "content_block_stop", `{"type":"content_block_stop","index":0}`)
+		writeSSEEvent(w, "message_stop", `{"type":"message_stop"}`)
+	}))
+	defer server.Close()
+
+	provider, err := New(Options{BaseURL: server.URL + "/", Model: "claude-test"})
+	if err != nil {
+		t.Fatalf("New returned error: %v", err)
+	}
+	events := collectProviderEvents(t, provider)
+
+	for _, e := range events {
+		if e.Type == zeroruntime.StreamEventToolCallDropped {
+			t.Errorf("valid tool_use block must not be dropped; events: %+v", events)
+		}
+	}
+}
--- a/internal/providers/anthropic/finish_reason_test.go
+++ b/internal/providers/anthropic/finish_reason_test.go
@@ -1,0 +1,52 @@
+package anthropic
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// A message_delta with stop_reason=="max_tokens" means the response was
+// truncated at the output cap. The provider must surface it on the done event so
+// the agent does not treat a clipped answer as complete.
+func TestStreamCompletionSurfacesMaxTokensStopReason(t *testing.T) {
+	provider := newTestProvider(t, func(w http.ResponseWriter, r *http.Request) {
+		writeSSEEvent(w, "message_start", `{"type":"message_start","message":{"usage":{"input_tokens":5,"output_tokens":0}}}`)
+		writeSSEEvent(w, "content_block_delta", `{"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"cut"}}`)
+		writeSSEEvent(w, "message_delta", `{"type":"message_delta","delta":{"stop_reason":"max_tokens"},"usage":{"output_tokens":7}}`)
+		writeSSEEvent(w, "message_stop", `{"type":"message_stop"}`)
+	})
+
+	events := collectProviderEvents(t, provider)
+	var doneReason string
+	var sawDone bool
+	for _, e := range events {
+		if e.Type == zeroruntime.StreamEventDone {
+			sawDone = true
+			doneReason = e.FinishReason
+		}
+	}
+	if !sawDone {
+		t.Fatalf("no done event; events: %+v", events)
+	}
+	if doneReason != zeroruntime.FinishReasonLength {
+		t.Fatalf("done FinishReason = %q, want %q", doneReason, zeroruntime.FinishReasonLength)
+	}
+}
+
+// A normal end_turn stop_reason must leave the done event's FinishReason empty.
+func TestStreamCompletionNormalStopReasonHasNoFinishReason(t *testing.T) {
+	provider := newTestProvider(t, func(w http.ResponseWriter, r *http.Request) {
+		writeSSEEvent(w, "content_block_delta", `{"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"ok"}}`)
+		writeSSEEvent(w, "message_delta", `{"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"output_tokens":2}}`)
+		writeSSEEvent(w, "message_stop", `{"type":"message_stop"}`)
+	})
+
+	events := collectProviderEvents(t, provider)
+	for _, e := range events {
+		if e.Type == zeroruntime.StreamEventDone && e.FinishReason != "" {
+			t.Fatalf("normal stop leaked FinishReason %q", e.FinishReason)
+		}
+	}
+}
--- a/internal/providers/anthropic/finish_reason_test.go
+++ b/internal/providers/anthropic/finish_reason_test.go
@@ -44,9 +44,16 @@ func TestStreamCompletionNormalStopReasonHasNoFinishReason(t *testing.T) {
 	})
 
 	events := collectProviderEvents(t, provider)
+	var sawDone bool
 	for _, e := range events {
-		if e.Type == zeroruntime.StreamEventDone && e.FinishReason != "" {
-			t.Fatalf("normal stop leaked FinishReason %q", e.FinishReason)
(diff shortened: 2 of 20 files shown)


gnanam1990:
@coderabbitai review — addressed all 6 comments: the blocker (providerio now honors ctx with idle watchdog disabled), the major (OpenAI ctx-cancel during retry backoff no longer misreported as 5xx), and the 4 test-robustness nits (done/start-event presence assertions + a providerio cancel-with-idle-disabled regression test). Build/vet/-race green.

anandh8x (approved):
## What's good

- **SSE idle watchdog is a real bug fix.** The comment is explicit: "Hosted gateways (e.g. Ollama Cloud) sometimes stall mid-stream after a tool-call delta; without this the agent blocks forever." The implementation is correct — `context.WithCancel` derived from the caller's ctx, SSE scan in a goroutine, select on `ctx.Done()` / idle timer / payload channel, idle fires `cancel()` to unblock the HTTP read and returns `ErrStreamIdle`. The `providerio_test.go` test `TestScanSSEDataWithContextAbortsOnIdle` asserts both the error and the cancel callback was invoked.

- **Regression test for cancel when idle is disabled** — `TestScanSSEDataWithContextHonorsCancelWhenIdleDisabled` is a great defensive test. A previous bug where ctx cancel was ignored when the idle watchdog was disabled (because the goroutine/select loop was skipped) is now locked. The fix always runs the goroutine+select loop; idle disable is just a nil-channel in the select. The test name and comment both call out the specific bug.

- **Dropped tool call handling is consistent across all three providers.** Anthropic: `content_block_start` with empty id/name → dropped. Gemini: `functionCall` part with empty name → dropped; top-level `functionCalls` with empty name → dropped. OpenAI: `closeOpen` for nameless call with id or args → dropped. Runtime: `CollectedStream.DroppedToolCalls` counter; the agent can use this to ask the model to retry. Mirroring across providers is the right design — the runtime is the single source of truth for "what is a valid tool call".

- **Finish reason normalization is clean and complete.** `length` / `max_tokens` / `MAX_TOKENS` → `FinishReasonLength`. `content_filter` / `SAFETY` / `PROHIBITED_CONTENT` / `BLOCKLIST` / `SPII` → `FinishReasonContentFilter`. Normal completion (stop, end_turn, STOP) → empty. `Truncated()` is a clean API. The runtime captures the reason on any event, not just done — defensive against providers that attach the reason to a usage event or tool_call event. The test `TestCollectStreamRecordsFinishReason` round-trips through `CollectStream` to `Truncated()`.

- **Stream usage on OpenAI is a real bug fix.** Previous test asserted `stream_options` was OMITTED — but that meant OpenAI never sent the final usage chunk, so token accounting was silently zero. New code sends `stream_options: { include_usage: true }`. `TestStreamCompletionPostsChatCompletionRequest` was updated to assert it's present. This is a correctness fix that the user-visible UX would have noticed as zero token counts in `/usage`.

- **`MaxTokens` cap is wired from the model registry via the factory.** `TestStreamCompletionSendsMaxCompletionTokens` confirms `max_completion_tokens: 1234` is sent; `TestStreamCompletionOmitsMaxCompletionTokensWhenUnset` confirms it's omitted when not set. The `factory.go` change pulls from `resolved.maxOutputTokens` (PR #56). The provider now actually applies the cap that the model registry has declared.

- **Retry on 5xx is a real reliability improvement.** Hosted gateways sometimes return intermittent 500s that succeed on retry. The code retries up to 3 times with linear backoff. Critically, the test `TestStreamCompletionClassifiesHTTPErrorsAndRedactsToken` was updated to assert 503 and 529 get the "rate limit error:" prefix — a real bug fix in production error classification (the previous code only checked for `StatusTooManyRequests`).

- **Multi-line SSE data fields are now properly joined.** The previous OpenAI provider parsed one line at a time and would drop the continuation, producing malformed JSON. `TestStreamCompletionJoinsMultiLineDataFields` confirms `data: {"choices":[{"delta":\ndata: {"content":"joined"}}]}` is joined into one payload. The shared `scanSSEPayloads` is the single source of truth for SSE multi-line semantics, used by all three providers.

- **Tool call collector rewrite fixes several real bugs:**
  - **Start order preserved regardless of end order** — previously `ToolCallEnd` appended immediately, so a call that ended first jumped ahead. `flush` now emits in start order. `TestCollectStreamPreservesStartOrderRegardlessOfEndOrder` asserts this.
  - **Distinct empty-ID calls don't merge** — synthetic keys (`\x00synthetic-N`) keep simultaneous empty-id calls separate. The `\x00` prefix is a clever trick: it can't collide with real tool call IDs (OpenAI spec, Anthropic, Gemini all use generated IDs without null bytes).
  - **Pre-start delta adopted** — `pendingEmptyDelta` remembers a delta that arrived before any start; the next empty-id start adopts it so the buffered args and the name land on the same call. `TestCollectStreamEmptyIDDeltaBeforeStartIsAdopted` covers this case.
  - **Duplicate start preserves first name** — a backend re-emitting the same start can't overwrite a resolved name. `TestCollectStreamKeepsFirstNameOnDuplicateStart` asserts this.
  - **Nameless calls dropped silently at flush** — the agent never sees `Name == ""`.

- **Gemini `emitDone` pointer receiver fix is a real bug fix.** The previous code had `func (provider *Provider) emitDone(ctx, state streamState, events) { ... state.done = true }` where `state` was by-value, so `state.done = true` was a dead store. The new `*streamState` receiver fixes it. `TestEmitDoneMarksStateDoneThroughPointer` confirms the fix.

- **OpenAI `applyDelta` preserves first id/name** — fixes a real bug where some OpenAI-compatible backends (e.g., `minimax` via Ollama) re-emit tool_call entries with empty id/name at the same index, corrupting the in-flight call and leaking a phantom nameless call. The `if delta.ID != "" && call.id == ""` guard preserves the first.

- **Cross-provider consistency is real.** All three providers use the shared `ScanSSEDataWithContext`, `providerio.Redact`, and `providerio.ClassifiedError`. Future SSE fixes (comment handling, retry on partial reads, etc.) only need to be made in one place. The `providerio` package is the right home for shared reader logic.

- **Test coverage is comprehensive.** Every fix has a focused test: idle watchdog (4 tests), dropped calls (3 tests across providers), finish reason (5 tests across providers and runtime), max_tokens (2 tests), retry (existing test extended), multi-line SSE (1 test), collector (3 tests for start order / empty-id / duplicate start). The `TestStreamCompletionClassifiesHTTPErrorsAndRedactsToken` update for 503/529 is a real bug-fix lock.

- **Backward compatibility is preserved.** `CollectStream` still works with old code (no `DroppedToolCalls` / `FinishReason`); old callers just see zero values for the new fields. `StreamEvent.FinishReason` defaults to `""`. `toolState.finishReason` defaults to `""`.

- **The `ScanSSEDataWithContext` complexity is high but justified.** The goroutine + select pattern with optional idle timer is non-trivial, but it's the correct way to abort a blocking read without races. Comments are detailed and the test coverage is good.

1. **The `pendingEmptyDelta` adoption is subtle.** `TestCollectStreamEmptyIDDeltaBeforeStartIsAdopted` covers the case, but a future maintainer might be confused by the synthetic-key + pendingEmptyDelta interaction. A doc comment on `pendingEmptyDelta` would help.

2. **No benchmark for `toolCallCollector` or `CollectStreamWithOptions`.** These run on every provider response, so a small benchmark would catch future regressions where someone adds an O(n²) operation.

3. **Retry logic is linear backoff (`attempt * 400ms`).** For 3 attempts, max wait is 2s. Exponential (400ms, 800ms, 1600ms) would be more standard. Fine for the current 3-attempt cap.

4. **`MaxAttempts = 3` is hard-coded.** A future `Options.MaxAttempts` would let users tune for their network. Not blocking.

5. **`MaxTokens` is only sent for OpenAI.** Anthropic and Gemini providers have `Options.MaxTokens` but the wiring is missing. A follow-up could wire it for them.

6. **`streamState.finishReason` is a string, not a typed enum.** Using a typed `FinishReason` enum would catch typos at compile time. The runtime constant is `string`, so the field is `string`. The map* functions do the conversion.

7. **No test for `providerio.Redact` directly.** Covered transitively via the OpenAI test. A focused unit test would document the redaction contract (Bearer token, API key substring, etc.).

8. **No test for `providerio.ClassifiedError` directly.** Same — covered transitively.

9. **Idle timeout in tests is 80ms, test timeout 3s.** Aggressive but not flaky. A 200ms timeout would be more robust on slow CI.

10. **The `closeOpen` change in `tool_state.go` is subtle.** Previous code silently dropped nameless calls; new code emits `StreamEventToolCallDropped` if the call has any id or args. If the call has no id and no name and no args, it's truly malformed and still silently dropped. The condition `call.id != "" || call.name != "" || call.arguments != ""` is precise.

11. **The `openEmptyID` stack is needed for nested empty-id calls.** `TestCollectStreamDoesNotMergeDistinctEmptyIDCalls` covers the sequential case. A nested test (start, start, delta-a, delta-b, end-a, end-b) would lock the stack semantics.

12. **No test for the retry path directly.** The retry logic is only exercised via the `TestStreamCompletionClassifiesHTTPErrorsAndRedactsToken` test (which expects one attempt). A focused test that returns 500 then 200 would lock the retry behavior.

13. **`MapFinishReason` for OpenAI only maps `length` and `content_filter`.** `tool_calls` and `stop` are normal finishes and return `""` (which is the right default). The exhaustive list is documented in the comment.

14. **The `backoff` function lives in the OpenAI provider.** If Gemini or Anthropic add retry logic, they'd need their own backoff (or share via `providerio`). Not blocking.

15. **The OpenAI `TestStreamCompletionPostsChatCompletionRequest` was updated to assert `stream_options.include_usage` is true.** The previous test asserted it was OMITTED. The change is a correctness fix and the test update is correct.

16. **The `TestStreamCompletionIdleTimeoutAbortsStalledStream` test uses a 3-second outer timeout.** A stalled stream is the failure mode, so the timeout is appropriate. The idle timeout is 80ms; the test asserts termination well within the 3s budget.

17. **The `flush` method does triple duty: emits + drops nameless + clears state.** Could be split, but the current implementation is correct and well-tested.

18. **The Gemini `emitDone` was previously `func (provider *Provider) emitDone(ctx, state streamState, events)` — by-value receiver.** The diff shows the receiver was already `*streamState` (the file in the previous slice had the by-value bug, and this slice fixes it). The test name and comment make the fix explicit.

19. **`TestStreamCompletionSendsMaxCompletionTokens` decodes the request body into a generic `map[string]any`.** This is a common pattern in the existing tests. A typed decode would be cleaner but the map approach is fine for asserting one field.

20. **The `ScanSSEDataWithContext` `cancel` parameter is a `context.CancelFunc`** — passed by the caller so the helper can abort the in-flight request. The helper doesn't own the cancel; it just calls it on idle/ctx-done. This is a clean separation.

21. **The error message format for idle timeout includes the duration** (`"idle timeout after 1m30s (upstream stopped sending data)"`). Useful for debugging. The redaction applies to the duration, not the message text.

22. **`providerio` test `TestScanSSEDataWithContextDeliversThenEOF` confirms the multi-line join still works** with the context variant. Good — the join logic is in `scanSSEPayloads`, shared between both readers.

23. **The `factory.go` change adds `MaxTokens` to the OpenAI provider** but the `Options.HTTPClient` is set in the factory call. A future enhancement could also pass `StreamIdleTimeout` from config. The current default of 90s is reasonable.

24. **The `TestCollectStreamEmptyIDDeltaBeforeStartIsAdopted` test asserts the call's name is `read_file` AND the args are the full JSON.** The adoption logic preserves both. The test would fail if the early delta was orphaned into a separate call (the previous behavior).

Approving — this is a substantial reliability PR. The idle watchdog fixes a real hang, the dropped-call handling is consistent across all three providers, the finish_reason mapping is complete, the SSE reader is now shared, the OpenAI stream usage is now requested (silently zero before), the retry logic handles transient 5xx with real bug fixes for 503/529, the tool call collector fixes several real bugs (start order, distinct empty-id calls, pre-start delta adoption, nameless drops), and the test coverage is comprehensive. The follow-ups are minor (synthetic key doc comment, no benchmarks, MaxTokens not wired for Anthropic/Gemini, retry params hard-coded, regression tests for retry/nested empty-id).

PR #108, opened 2026-06-07
Add background task manager

anandh8x:
Summary:
- adds internal/background Manager for registering and tracking background specialist tasks
- stores safe per-task NDJSON output paths under the Zero data directory
- supports status updates, parent filtering, and cross-platform task termination hooks

Self-review:
- intentionally internal-only; no CLI or user-facing Task tools are wired in this PR
- Get/List return task snapshots instead of internal pointers
- task IDs and custom output paths are validated to avoid path traversal

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/background
- GOCACHE=/tmp/zero-go-cache GOOS=windows go test -c ./internal/background
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Background task manager: register tasks, track status/PID, per-task output files, parent-child listing, and cross-platform termination behavior.

* **Bug Fixes & Safety**
  * Validates task IDs and output paths to prevent escapes/collisions; exclusive output-file creation and stable ordering by start time then ID.

* **Tests**
  * Comprehensive tests for registration, permissions, ordering, kill behavior, error cases, and default-root resolution.

### DIFF
--- a/internal/background/manager.go
+++ b/internal/background/manager.go
@@ -208,22 +208,62 @@ func (manager *Manager) Get(taskID string) (Task, bool) {
 func (manager *Manager) Kill(taskID string) error {
 	taskID = strings.TrimSpace(taskID)
 
+	pid, err := manager.killTarget(taskID)
+	if err != nil {
+		return err
+	}
+	if !manager.isRunningPID(taskID, pid) {
+		return nil
+	}
+	if err := manager.killProcess(pid); err != nil {
+		return fmt.Errorf("kill background task %s: %w", taskID, err)
+	}
+	return manager.markKilledIfStillRunning(taskID, pid)
+}
+
+func (manager *Manager) killTarget(taskID string) (int, error) {
 	manager.mu.Lock()
+	defer manager.mu.Unlock()
 	task, ok := manager.tasks[taskID]
-	manager.mu.Unlock()
+	if !ok {
+		return 0, fmt.Errorf("background task not found: %s", taskID)
+	}
+	if task.Status != StatusRunning {
+		return 0, fmt.Errorf("background task %s is %s", taskID, task.Status)
+	}
+	if task.PID <= 0 {
+		return 0, fmt.Errorf("background task %s has no pid", taskID)
+	}
+	return task.PID, nil
+}
+
+func (manager *Manager) isRunningPID(taskID string, pid int) bool {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	task, ok := manager.tasks[taskID]
+	return ok && task.Status == StatusRunning && task.PID == pid
+}
+
+func (manager *Manager) markKilledIfStillRunning(taskID string, pid int) error {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	task, ok := manager.tasks[taskID]
 	if !ok {
 		return fmt.Errorf("background task not found: %s", taskID)
 	}
 	if task.Status != StatusRunning {
-		return fmt.Errorf("background task %s is %s", taskID, task.Status)
+		return nil
 	}
-	if task.PID <= 0 {
-		return fmt.Errorf("background task %s has no pid", taskID)
+	if task.PID != pid {
+		return fmt.Errorf("background task %s pid changed before kill completed", taskID)
 	}
-	if err := manager.killProcess(task.PID); err != nil {
-		return fmt.Errorf("kill background task %s: %w", taskID, err)
+	task.Status = StatusKilled
+	task.ExitCode = -1
+	if task.CompletedAt.IsZero() {
+		task.CompletedAt = manager.now()
 	}
-	return manager.UpdateStatus(taskID, StatusKilled, -1)
+	manager.tasks[taskID] = task
+	return nil
 }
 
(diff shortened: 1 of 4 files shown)


gnanam1990 (changes requested):
## Review — strong package; one concurrency fix requested

Really clean foundation — nicely done overall:
- `Task` is a pure value type (no pointers/slices), so `Get`/`List`/`ListByParent` return genuine **snapshots** — no shared-state leaks. 👍
- `taskIDPattern` (`^[A-Za-z0-9][A-Za-z0-9_-]{0,127}$`) blocks slashes/dots → no path traversal via id; output file reserved atomically with `O_RDWR|O_CREATE|O_EXCL`; PID guarded (`> 0`) before any kill.
- Injectable `Now`/`KillProcess`/`Env` make it testable; cross-platform build + `go test -race` both verified locally. Tidy.

### 🟠 Please address — `Kill` TOCTOU on a process-kill path
`Kill` snapshots `{Status, PID}` under the lock, **releases it**, then calls `killProcess(pid)` and `UpdateStatus(…, StatusKilled, -1)`:
```go
mu.Lock(); task, ok := tasks[id]; mu.Unlock()
... if task.Status != StatusRunning ... if task.PID <= 0 ...
manager.killProcess(task.PID)
manager.UpdateStatus(taskID, StatusKilled, -1)   // unconditional overwrite
```
Between the snapshot and these calls, the task can finish on its own (its monitor calls `UpdateStatus(Completed)`). Two consequences:
1. `killProcess(task.PID)` signals a **stale PID** — if the process already exited and the PID was **reused**, this can signal an *unrelated* process. In a kill path that's a real (if low-probability) safety bug.
2. `UpdateStatus(StatusKilled, -1)` then **clobbers** the terminal `Completed`/`Error` status with `Killed` and `ExitCode=-1`.

Suggest making the terminal transition conditional — e.g. have the kill only flip `Running → Killed` (a `UpdateStatus` variant / re-check under the lock that the task is still `Running` before signalling and before overwriting). Keep `killProcess` outside the lock so you don't block on the syscall, but gate the state transition so a self-completion wins the race.

### 🟡 Non-blocking notes
- **`outputFile` confinement is lexical** (`filepath.Rel` + `..` check) — solid for plain paths, but a symlinked directory component under `rootDir` could let a custom `OutputFile` escape (no `EvalSymlinks`). Fine now (private `0700` dir, internal caller), but worth `EvalSymlinks`-confining once the custom-path input is ever caller/externally influenced (mirrors the workspace path resolver).
- **POSIX vs Windows kill semantics differ**: POSIX sends `SIGTERM` (graceful) yet marks the task `Killed` even if the process ignores it; Windows uses `taskkill /F` (forceful). Consider documenting this, or escalating to `SIGKILL` after a grace period and only marking `Killed` on confirmed termination.
- **nit**: `Register` holds `mu` across `MkdirAll`/`OpenFile`, serializing concurrent registrations on disk I/O — minor.

The TOCTOU is the only thing I'd hold on (it's a process-killing primitive other slices will build on). The rest are advisory. 👍

<sub>🤖 Verified review via Claude Code (build/vet/-race/windows cross-compile run locally).</sub>

gnanam1990 (approved):
## Re-review — TOCTOU fix looks correct, approving

The `Guard background task kill races` commit addresses the blocker cleanly:
- `killTarget` snapshots `{Status, PID}` under the lock; `isRunningPID` **re-checks under the lock that the task is still `Running` with the *same* PID right before signalling** (so a self-completed task is a no-op, no stale-PID signal); `markKilledIfStillRunning` makes the terminal transition **conditional** — only `Running → Killed`, and errors if the PID changed — so a `Completed`/`Error` status can no longer be clobbered.
- `TestManagerKillDoesNotClobberCompletedTask` covers exactly the clobber scenario. `go test -race`, `go vet`, and the Windows cross-compile all pass.

That fully resolves what I flagged. **Approving.**

### One residual note (non-blocking, inherent)
There's still an irreducible window between `isRunningPID` returning true (lock released) and `killProcess(pid)` actually signalling — if the process exits and the PID is reused in that instant, the signal could hit an unrelated process. That's fundamental to PID-based termination and not closable with locks alone; closing it fully would need OS-level handles (`pidfd` on Linux, job objects on Windows). Fine to leave for a future hardening pass — the lock-guarded re-check is the right mitigation at this layer.

The earlier advisory notes (lexical `outputFile` confinement / symlinks, POSIX `SIGTERM` vs Windows force-kill) also remain as optional future hardening — none blocking.

Nice quick turnaround. 👍

<sub>🤖 Verified re-review via Claude Code (build/vet/-race/windows run locally).</sub>

anandh8x:
Addressed Gnanam's requested change:
- Kill now re-checks the task is still running with the same PID before signalling.
- After the kill attempt, it records Killed only if the task is still running with that same PID, so a concurrent Completed/Error update wins and is not clobbered.
- Added a regression test where the task completes during Kill.

Validation:
- GOCACHE=/tmp/zero-go-cache go test ./internal/background
- GOCACHE=/tmp/zero-go-cache go test -race ./internal/background
- GOCACHE=/tmp/zero-go-cache GOOS=windows go test -c ./internal/background
- git diff --check

CI is green, including Windows smoke.

PR #109, opened 2026-06-07
Sandbox hardening: destructive/network/installer classification + interactive-command detection

gnanam1990:
## Module 3 of N — Sandbox hardening

Third reviewable slice of the runtime-core split (off `main`, after #107 merged). **Scope: `internal/sandbox` only** (4 files, ~894 lines incl. tests). Clean additive extract — imports only stdlib, whole tree builds unchanged, no new deps.

### Live now (improves existing behavior)
`Classify` is already invoked by `sandbox.Evaluate` (engine.go) and `agent/loop.go`, so these harden the gate immediately:
- **`rm -rf`** targeting `/`, `$HOME`, `~`, `*` — now tolerant of surrounding quotes, `${HOME}` braces, combined/reordered `-r`/`-f` flags, and an optional `--` separator.
- **`chmod 777`** flagged only when recursive or root-targeted (a single-file `chmod 777 foo` no longer false-positives).
- **Network + piped-installer** detection (`curl … | sh`, incl. `|zsh`/other shells); the command is resolved across `command`/`cmd`/`script`/`shell` aliases so an alias key can't slip past the destructive/network gate.

### Forward code (compiles now, wired later)
`DetectInteractiveCommand` flags interactive programs (`vim`, `less`, …) even behind wrappers (`sudo`/`nice`/`timeout`, and `sh -c`/`bash -c` payloads via recursion), hardened against absolute-path / quote / escape / command-substitution bypasses. Its caller is the **tools/bash module** (a later PR); it's exercised here by `safe_command_test.go`.

### Testing
`go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./internal/sandbox/` all green.

Part of decomposing #101 (draft) into reviewable PRs; subagents excluded.

## Summary by CodeRabbit

* **New Features**
  * Improved shell command risk detection: better identification of destructive commands, network-risk patterns, and piped installers
  * Interactive-command detection added to flag programs that require a TTY and provide guidance

* **Tests**
  * Comprehensive tests covering destructive command variants, piped-installer distinctions, interactive-command detection, wrappers, and edge-case parsing

### DIFF
--- a/internal/sandbox/engine_test.go
+++ b/internal/sandbox/engine_test.go
@@ -164,13 +164,23 @@ func TestEngineClassifiesNetworkAndDestructiveShellCommands(t *testing.T) {
 		t.Fatalf("destructive shell decision = %#v, want critical destructive deny", destructive)
 	}
 
+	// A remote fetch piped into a shell is the dangerous fetch-and-execute idiom.
 	pipedInstallerRisk := Classify(Request{
 		ToolName:   "bash",
 		SideEffect: SideEffectShell,
-		Args:       map[string]any{"command": "cat install.sh | BASH"},
+		Args:       map[string]any{"command": "curl -fsSL https://get.example.com/install.sh | BASH"},
 	})
 	if pipedInstallerRisk.Level != RiskCritical || !HasRiskCategory(pipedInstallerRisk, "piped_installer") {
 		t.Fatalf("piped installer risk = %#v, want critical piped_installer category", pipedInstallerRisk)
+	}
+	// A purely local pipe into a shell (no remote fetch) is NOT a piped installer.
+	localPipeRisk := Classify(Request{
+		ToolName:   "bash",
+		SideEffect: SideEffectShell,
+		Args:       map[string]any{"command": "cat install.sh | bash"},
+	})
+	if HasRiskCategory(localPipeRisk, "piped_installer") {
+		t.Fatalf("local pipe wrongly flagged piped_installer: %#v", localPipeRisk)
 	}
 
 	workspaceShell := engine.Evaluate(context.Background(), Request{
--- a/internal/sandbox/risk.go
+++ b/internal/sandbox/risk.go
@@ -9,9 +9,49 @@ import (
 )
 
 var (
-	networkCommandPattern     = regexp.MustCompile(`(?i)\b(curl|wget|scp|ssh|rsync|nc|netcat|python3?\s+-m\s+http\.server|npm\s+(install|add|publish|login)|pnpm\s+(install|add|publish)|yarn\s+(add|publish)|bun\s+(add|install|publish)|pip3?\s+install|go\s+get|git\s+clone|gh\s+(release\s+download|repo\s+clone|api))\b`)
-	destructiveCommandPattern = regexp.MustCompile(`(?i)(\brm\s+(-[A-Za-z]*r[A-Za-z]*f|-rf|-fr)\s+(/|\$HOME|~|\*)|\bmkfs\b|\bdd\s+if=|\bchmod\s+-R\s+777\b|\bchown\s+-R\b)`)
+	networkCommandPattern = regexp.MustCompile(`(?i)\b(curl|wget|scp|ssh|rsync|nc|netcat|python3?\s+-m\s+http\.server|npm\s+(install|add|publish|login)|pnpm\s+(install|add|publish)|yarn\s+(add|publish)|bun\s+(add|install|publish)|pip3?\s+install|go\s+get|git\s+clone|gh\s+(release\s+download|repo\s+clone|api))\b`)
+	// destructiveCommandPattern matches the highest-risk shell forms:
+	//   - rm -rf (with combined/reordered r/f flags) targeting /, $HOME (bare,
+	//     quoted, or ${HOME} braced), ~, or *, with an optional `--` before the
+	//     target. Each target alternative tolerates optional surrounding quotes
+	//     so `rm -rf "/"` / `rm -rf '/'` cannot slip past the gate.
+	//   - chmod with combined/reordered flags and an octal-or-777 mode applied
+	//     RECURSIVELY (a -R/-r flag) or to an ABSOLUTE path / sensitive tree
+	//     (e.g. chmod -Rf 777 /, chmod -R 0777 /, chmod 777 -R /etc, chmod 777 /etc).
+	//     A single-file chmod 777 (e.g. `chmod 777 script.sh`) is intentionally
+	//     NOT flagged — the intent is recursive/directory-tree chmod.
+	//   - mkfs, dd if=, chown -R.
+	destructiveCommandPattern = regexp.MustCompile(`(?i)(\brm\s+(-[A-Za-z]*r[A-Za-z]*f|-rf|-fr)\s+(--\s+)?["']?(\$\{?HOME\}?|/|~|\*)["']?|\bmkfs\b|\bdd\s+if=|\bchmod\s+(-[A-Za-z]*[rR][A-Za-z]*\s+)+0?777\b|\bchmod\s+(-\S+\s+)*0?777\s+-[A-Za-z]*[rR][A-Za-z]*\b|\bchmod\s+(-\S+\s+)*0?777\s+["']?/|\bchown\s+-R\b)`)
+	// pipedInstallerPattern matches a pipe into a POSIX shell, with or without a
+	// space and across sh/bash/zsh/ksh/dash (so `curl x|sh`, `|bash`, `| zsh`).
+	pipedInstallerPattern = regexp.MustCompile(`(?i)\|\s*(ba|z|k|da)?sh\b`)
+	// destructiveExtraPatterns hold high-severity patterns that the legacy
+	// destructiveCommandPattern does not already cover. Folded in from the
+	// blueprint safe_bash.go without duplicating existing matches.
+	destructiveExtraPatterns = []*regexp.Regexp{
+		// Fork bomb (and minor spacing variants).
+		regexp.MustCompile(`:\s*\(\s*\)\s*\{\s*:\s*\|\s*:\s*&\s*\}\s*;\s*:`),
+		// Writing to a raw block device (dd of=, redirect to /dev/sdX, etc.).
+		regexp.MustCompile(`(?i)>\s*/dev/(sd[a-z]+\d*|nvme\d+n\d+(p\d+)?|hd[a-z]+\d*|xvd[a-z]+\d*|mmcblk\d+)`),
+		regexp.MustCompile(`(?i)\bof=/dev/(sd[a-z]+\d*|nvme\d+n\d+(p\d+)?|hd[a-z]+\d*|xvd[a-z]+\d*|mmcblk\d+)`),
+		// rm -rf targeting the root, including long flags and --no-preserve-root.
+		regexp.MustCompile(`(?i)\brm\s+(-[A-Za-z]*\s+|--[a-z-]+\s+)*(/|/\*)(\s|$)`),
+		// mkfs.<fstype> form (e.g. mkfs.ext4) not caught by the bare \bmkfs\b above when followed by a dot.
+		regexp.MustCompile(`(?i)\bmkfs\.[a-z0-9]+\b`),
+	}
 )
+
+func matchesDestructive(command string) bool {
+	if destructiveCommandPattern.MatchString(command) {
+		return true
+	}
+	for _, pattern := range destructiveExtraPatterns {
+		if pattern.MatchString(command) {
+			return true
+		}
+	}
+	return false
+}
 
 func Classify(request Request) Risk {
 	categories := map[string]bool{}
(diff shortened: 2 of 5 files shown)


anandh8x (approved):
## What's good

- **Command alias resolution closes a real bypass.** The bash tool accepts the command under `command` / `cmd` / `script` / `shell` keys. The previous code only checked `command`, so a caller could bypass classification by passing `cmd: "rm -rf /"`. The new `firstArgString(args, "command", "cmd", "script", "shell")` resolves the first non-empty alias key, so the destructive gate applies regardless of which alias the caller used. `TestClassifyResolvesCommandAliases` locks the contract for `cmd`, `script`, and `shell`.

- **Quoted/braced HOME is a real bypass fix.** `rm -rf "$HOME"`, `rm -rf '$HOME'`, `rm -rf ${HOME}`, `rm -rf "${HOME}"` — all must match. The new pattern tolerates optional surrounding quotes and the bare/braced `$HOME` forms. The audit comment makes the threat model explicit: "a quoted root target must not bypass the destructive deny gate". `TestClassifyFlagsRmRfQuotedOrBracedHome` and `TestClassifyFlagsRmRfQuotedRoot` lock this for 6 variants.

- **chmod flag combination is a real fix.** Previous pattern only caught `\bchmod\s+-R\s+777\b` — so `chmod -Rf 777 /`, `chmod 777 -R /`, `chmod -R 0777 /`, `chmod 0777 -R /` would slip through. The new pattern allows any flag order with `-R`/`-r` anywhere AND either recursive OR absolute path. The audit-finding test `TestClassifyFlagsChmodAndRmFlagVariants` locks 4 variants. The companion test `TestClassifyChmod777SingleFileNotDestructive` confirms `chmod 777 myscript.sh` is intentionally NOT flagged (the comment says "the intent is recursive/directory-tree chmod").

- **mkfs.<fstype> pattern closes a gap.** Previous pattern only caught bare `\bmkfs\b`, so `mkfs.ext4 /dev/sda` would not match (the dot breaks the word boundary). The new extra pattern catches `mkfs\.[a-z0-9]+`.

- **Block device write detection is a real fix.** `dd of=/dev/sda`, `> /dev/nvme0n1`, `> /dev/sdb1` — all must be caught. The new patterns cover `sd[a-z]+\d*`, `nvme\d+n\d+(p\d+)?`, `hd[a-z]+\d*`, `xvd[a-z]+\d*`, `mmcblk\d+` for both `of=` and `>` redirects. `TestClassifyFlagsBlockDeviceWrite` locks this for 3 variants.

- **Fork bomb detection.** The classic `:\(\)\{ :\|:& \};:` is matched with minor spacing tolerance (`:\s*\(\s*\)\s*\{\s*:\s*\|\s*:\s*&\s*\}\s*;\s*:`). `TestClassifyFlagsForkBombAsDestructive` locks this.

- **Piped installer hardening.** Previous code only matched literal `"| sh"` and `"| bash"` substrings. The new `pipedInstallerPattern` matches `|sh`, `|bash`, `|zsh`, `|ksh`, `|dash` with or without space. `TestClassifyFlagsPipedInstallerVariants` locks 5 variants.

- **Long-flag `rm -rf` is caught.** The extra pattern `\brm\s+(-[A-Za-z]*\s+|--[a-z-]+\s+)*(/|/\*)(\s|$)` matches `rm --recursive --force /` and `sudo rm -rf --no-preserve-root /`. `TestClassifyFlagsRmRfRootVariants` locks 4 variants.

- **Interactive command detection is a substantial new module.** 445 lines of `safe_command.go` correctly handle:
  - **Skipping leading env assignments** (`FOO=bar cmd`).
  - **Skipping wrapper prefixes** (`sudo`, `env`, `nice`, `timeout`, `stdbuf`, `setsid`, `ionice`, `xargs`, `doas`, `nohup`, `time`, `exec`, `command`).
  - **Skipping option tokens belonging to wrappers** (`-u`, `-g`, `-h`, `-p`, `-C`, `-r`, `-t`, `-U`, `-D`, `-c`, `-n` — these consume the next token).
  - **Recursing into `sh -c` / `bash -c` / `zsh -c` / `ksh -c` / `dash -c` payloads** so an interactive program hidden inside a payload is still detected.
  - **Splitting on shell operators and command substitution boundaries** (`&&`, `||`, `;`, `|`, `$(`, `)`, `` ` ``).
  - **Stripping quotes / escapes from program tokens** — closes `vi\m`, `v"i"m`, `'v'im`, `le\ss` bypasses.
  - **Recognizing REPL non-interactive flags** (`python -c`, `node -e`, `ruby -e`, etc.).
  - **Distinguishing `ssh host` (interactive) from `ssh host 'command'` (not)** — uses `hasTrailingCommand` for ssh / sqlite3 / redis-cli.
  - **Distinguishing `cat vim.txt` (file named vim) from `vim file` (editor)** — by anchoring on the first program token.
  - **Stripping directory prefix** (`/usr/bin/vim` → `vim`).
  - **Skipping Windows-only programs on non-Windows** (`notepad` on `linux` is not flagged).

- **Embedded quote stripping is a real bypass fix.** The previous `normalizeProgramToken` would have only stripped surrounding quotes, so `vi\m` (mid-token backslash) would resolve to `vi\m`, not `vim`. The new `stripChars` removes `"`, `'`, `` ` ``, `\` from anywhere in the token. The docstring explicitly says "this is done BEFORE the directory-prefix trim so an escape can't masquerade as a path separator (e.g. vi\m must become vim, not m)". `TestDetectInteractiveStripsEmbeddedQuotingFromToken` locks 4 cases: `vi\m`, `v"i"m`, `'v'im`, `le\ss`.

- **Segment boundary is a real false-positive fix.** Without `commandBody` anchoring on the real command start, `echo "git rebase -i is interactive"` would match the substring `git rebase -i`. The new code joins the remaining fields after env/wrapper stripping, so the segment text inside quoted arguments is not a match. `TestDetectInteractiveSegmentBoundary` covers 4 false-positive cases and 4 true-positive cases (including across `&&`).

- **`sh -c` payload recursion is correct.** A user could write `sh -c 'vim file.txt'` to try to bypass detection. The new code calls `DetectInteractiveCommand` recursively on the payload. `TestDetectInteractiveThroughWrappersAndShellC` covers `sh -c 'vim file.txt'` and `bash -c "less /var/log/syslog"`.

- **`isNumericToken` is a small but important helper.** `timeout 5 vim file.txt` has a bare numeric token between the wrapper and the program. Without skipping it, `firstProgram` would return `5` instead of recursing. The test for `timeout 5` covers this.

- **The audit findings are explicitly documented in test comments.** "Finding 1", "Finding 2", "Finding 3", "Finding 4", "Finding 5", "Audit finding (HIGH)", "Audit finding (MED)", "Audit finding (LOW)" — the comments name the threat model and the fix. Future maintainers can grep the test file for "Audit finding" to see the threat model.

- **Test coverage is comprehensive.** 13 test functions in `risk_hardening_test.go` (covering 30+ command variants) and 8 in `safe_command_test.go` (covering 50+ command variants). Every audit finding has a focused test. Every wrapper has a test. Every shell operator has a test. Every quote / escape variant has a test.

- **`DetectInteractiveCommand` is exported** (capital `D`) so the agent runtime can call it from outside the sandbox package. That's the right design — it's a public API.

- **The `pipedInstallerPattern` regex is anchored correctly.** `(?i)\|\s*(ba|z|k|da)?sh\b` matches `| sh`, `|sh`, `|bash`, `|zsh`, `|ksh`, `|dash`. The optional `(ba|z|k|da)?` prefix and `sh\b` end-anchor cover all POSIX shells. Case-insensitive.

- **`hasTrailingCommand` is program-aware.** `ssh` requires `>= 2` positional args (host + command), `sqlite3` requires `>= 2` (db + sql), `redis-cli` requires `>= 1` (command). The default returns `false`. The test `TestDetectInteractiveCommandAllowsNonInteractive` confirms `ssh host 'uptime'` is NOT flagged.

- **The `destructiveExtraPatterns` are checked AFTER the legacy pattern.** A command matching both is still flagged. Order doesn't matter since any match is enough.

- **`commandBody` and `firstProgram` share iteration logic** but produce different results (joined rest vs first token). The duplication is minor and the comments explain each function's role.

- **The audit comment for `safe_command_test.go` is thorough.** The test names match the audit findings. The expected vs blocked lists are explicit.

1. **`DetectInteractiveCommand` is not yet wired into the bash tool or the agent runtime.** The function exists but is not called from `executeToolCall` or the risk classifier. A follow-up slice needs to wire it (probably as a `RiskHigh` category that the runtime can decide to deny or warn on). The current PR is the foundation; the integration is the next step. The function signature accepts `goos` so tests can lock the platform behavior.

2. **`nonInteractiveREPLFlags` is a small map** (7 entries: python, python3, node, ruby, php, psql, mysql). A user running `lua` interactively or `scala` would not be detected as REPL. The `interactivePrograms` map is also small (~30 common programs). A future enhancement could expand both, but the current coverage catches the common cases.

3. **`interactiveSegments` is a slice of 11 entries** (git rebase -i, git add -i/p, git commit -p, tail -f, journalctl -f, kubectl logs -f, docker logs -f, docker attach). Limited to common git/docker/kubectl/tail/journalctl. A future enhancement could add `cargo run`, `npm run dev --watch`, etc.

4. **`splitShellSegments` uses a naive `strings.NewReplacer`.** It doesn't handle quote-aware splitting. A command like `echo "a && b"` would NOT split (no `&&` to replace since the quotes are preserved in the source — Go's `strings.NewReplacer` works on the raw string). The test `TestDetectInteractiveSegmentBoundary` exercises `echo "git rebase -i is interactive"` and confirms it's NOT flagged. The naive splitter works for the common cases; a real shell tokenizer would be more correct but is overkill for this use case.

5. **No benchmark for `DetectInteractiveCommand`.** It's called on every bash tool invocation, so a benchmark would catch future regressions where someone adds a linear scan.

6. **`goos` is a `string`, not a typed enum.** A future enhancement could use a typed constant. The current code is correct and simple.

7. **`DetectInteractiveCommand` returns an empty `InteractiveCommandResult` for non-interactive commands.** The caller checks `result.Interactive`. No sentinel error pattern. Consistent with the rest of the package.

8. **The `firstProgram` and `commandBody` functions have duplicated iteration logic.** Both walk through fields, skipping env assignments, options, and wrappers. A future refactor could share the iteration. Minor.

9. **The `wrapperPrograms` map includes `command`** (the shell built-in). The shell built-in `command vim` runs `vim` without alias expansion. The test doesn't exercise this, but the function would correctly skip `command` and find `vim`.

10. **The `wrapperValueOptions` map requires exact match** (`wrapperValueOptions[field]`). Combined options like `ionice -c3` would NOT consume the next token (which is correct — it's a combined option). The test for `ionice -c3 vim file.txt` covers this.

11. **The `interactiveSegments` match uses `body == seg.match || strings.HasPrefix(body, seg.match+" ")`** — exact match or prefix followed by space. This means `tail -fing` would not match `tail -f`. Good.

12. **The `commandBody` function joins the remaining fields with a single space.** This is fine for matching the `interactiveSegments` patterns (which use single spaces). A future enhancement could preserve original whitespace for more accurate matching.

13. **No test for `DetectInteractiveCommand` with empty command and non-empty `goos`.** The current `TestDetectInteractiveCommandAllowsNonInteractive` includes `""` which would short-circuit before `goos` is checked. A test could lock the empty-input contract.

14. **The test `TestDetectInteractiveBypasses` is the regression test for false negatives.** `/usr/bin/vim`, `"vim"`, `'vim'`, `$(vim file.txt)`, `` `vim file.txt` ``, `/bin/less` — all should be flagged. `python script.py`, `cat vim.txt`, `grep ssh config.go`, `echo hello` — all should be allowed. The test exercises both lists.

15. **`HasRiskCategory` is a clean helper** (used in tests). It's defined elsewhere in the package (presumably from PR #77).

16. **No test for the `destructiveCommandPattern` against a safe `dd if=...` of a file** (not block device). The test only covers `dd if=/dev/zero of=/dev/sda` (block device write). A `dd if=file of=other_file` would not match (no block device). The pattern is `of=/dev/(sd|nvme|hd|xvd|mmcblk)` — strict to block devices.

17. **`commandBody` returns an empty string if no real command is found.** That's an edge case (e.g., `FOO=bar` with no command). The caller (`DetectInteractiveCommand`) would get `body == ""` and the segment match loop would skip it. The interactive-program loop's `firstProgram` would also return `""` and skip. No false positive.

18. **`hasNonInteractiveFlag` for ssh uses a different code path** (`hasTrailingCommand`) than for REPLs. The test confirms this. The split is intentional: REPLs use flags, SSH uses positional args.

19. **The audit finding test names are explicit about the threat.** "Audit finding (HIGH)", "Audit finding (MED)", "Audit finding (LOW)". A future maintainer can grep the test file for "Audit finding" to see the threat model.

20. **`DetectInteractiveCommand` is exported** so it can be called from outside the sandbox package. That's the right design — the function is a public API for the agent runtime to call before executing a bash tool call.

Approving — this is a substantial hardening slice. The command alias resolution closes a real bypass. The quoted/braced HOME, quoted root, chmod flag combinations, block device writes, fork bomb, piped installer, mkfs.<fstype>, and long-flag rm hardening all close real gaps. The interactive command detection is a substantial new module (445 lines) with comprehensive coverage of wrappers, shell operators, command substitution, quote stripping, and segment boundary anchoring. The audit findings are explicitly named in test comments, which makes the threat model discoverable. The test coverage is comprehensive (21 test functions across 2 new test files, covering 80+ command variants). The follow-up is wiring `DetectInteractiveCommand` into the bash tool / agent runtime so the detection actually blocks interactive commands at execution time.

gnanam1990:
@coderabbitai review — addressed all 5: the critical long-flag root `rm` (quote/`--` handling restored), chmod-777 narrowed to root/system trees (no single-file false positive), piped_installer now requires a remote fetch (local `cat|bash` no longer flagged), mongo/mongosh `--eval` suppression, and programIndex token normalization for full-path invocations. TDD for each; build/vet/-race/windows/full-suite green.

PR #110, opened 2026-06-07
Add specialist extends resolution

anandh8x:
Summary:
- adds extends frontmatter support to specialist manifests and JSON summaries
- resolves inherited specialist chains after built-in/project/user precedence is known
- merges prompts base-first, lets child fields win, and lets child tools replace inherited tools
- rejects missing base specialists and extends cycles during load

Self-review:
- intentionally no new user command surface; existing zero specialist show now prints extends when present
- keeps Load consumers receiving fully resolved manifests before later BuildArgs/Task execution work
- tests cover transitive inheritance, child tool overrides, missing bases, and cycles

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/cli
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Specialists can now extend/inherit from other specialists, automatically inheriting properties like model configuration, tools, and reasoning effort.
  * System prompts are intelligently combined when extending.

* **Improvements**
  * Circular dependencies in specialist inheritance chains are now detected and prevented.

### DIFF
--- a/internal/specialist/format.go
+++ b/internal/specialist/format.go
@@ -37,6 +37,9 @@ func FormatShow(manifest Manifest) string {
 	if manifest.Metadata.Model != "" {
 		lines = append(lines, "model: "+manifest.Metadata.Model)
 	}
+	if manifest.Metadata.Extends != "" {
+		lines = append(lines, "extends: "+manifest.Metadata.Extends)
+	}
 	if manifest.Metadata.ReasoningEffort != "" {
 		lines = append(lines, "reasoning effort: "+manifest.Metadata.ReasoningEffort)
 	}
--- a/internal/specialist/manifest.go
+++ b/internal/specialist/manifest.go
@@ -25,6 +25,7 @@ const (
 type Metadata struct {
 	Name            string   `json:"name"`
 	Description     string   `json:"description"`
+	Extends         string   `json:"extends,omitempty"`
 	Model           string   `json:"model,omitempty"`
 	ReasoningEffort string   `json:"reasoningEffort,omitempty"`
 	Tools           []string `json:"tools,omitempty"`
@@ -43,6 +44,7 @@ type Manifest struct {
 type Summary struct {
 	Name            string   `json:"name"`
 	Description     string   `json:"description"`
+	Extends         string   `json:"extends,omitempty"`
 	Model           string   `json:"model,omitempty"`
 	ReasoningEffort string   `json:"reasoningEffort,omitempty"`
 	Tools           []string `json:"tools,omitempty"`
@@ -72,6 +74,7 @@ var namePattern = regexp.MustCompile(`^[a-z][a-z0-9-]{0,63}$`)
 var knownMetadataKeys = map[string]bool{
 	"name":            true,
 	"description":     true,
+	"extends":         true,
 	"model":           true,
 	"reasoningEffort": true,
 	"tools":           true,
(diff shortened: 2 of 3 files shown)


Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- None.

### Looks Good

- Extends resolution runs after built-in/project/user precedence is merged, so inheritance uses the final winning specialist manifest rather than stale lower-precedence definitions.
- Child manifests inherit model/reasoning/tools only when omitted, child prompts compose base-first, and missing-base/cycle cases are covered by regression tests.
- Validation passed locally: `GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli ./...`, `go build ./cmd/zero`, `./zero specialist --help`, `./zero specialist list --json`, and `git diff --check`.
- Small focused change: 201 additions / 6 deletions across specialist manifest parsing, formatting, and tests.

---

**Verdict: Approve** — Clean specialist inheritance foundation.

Vasanthdev2004 (approved):
### Blockers

None found.

### Non-Blocking

- None.

### Looks Good

- Extends resolution runs after built-in/project/user precedence is merged, so inheritance uses the final winning specialist manifest rather than stale lower-precedence definitions.
- Child manifests inherit model/reasoning/tools only when omitted, child prompts compose base-first, and missing-base/cycle cases are covered by regression tests.
- Validation passed locally: `GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli ./...`, `go build ./cmd/zero`, `./zero specialist --help`, `./zero specialist list --json`, and `git diff --check`.
- Small focused change: 201 additions / 6 deletions across specialist manifest parsing, formatting, and tests.

---

**Verdict: Approve** — Clean specialist inheritance foundation.

gnanam1990 (approved):
## Review — correct, approving

Solid inheritance implementation:
- **Cycle detection is correct** — per-resolution `stack` catches self-extends, `A→B→A`, and transitive `A→B→C→A`; `defer delete(stack,name)` cleans up, and the shared `cache` stores only fully-resolved manifests, so no false cycles across the resolve loop. Missing bases are rejected with a precise message (`base specialist "x" for "y" not found`).
- **Merge semantics match the description** — child fields win, empty child fields inherit, child tools *replace* inherited tools, system prompts merge base-first; the merged manifest is re-`Validate`d (so an `extends`-only child that inherits its description/model stays valid). `extends` is name-pattern-validated (no traversal), and resolution runs *after* builtin/project/user precedence so consumers get fully-resolved manifests.
- Tests cover transitive inheritance, child overrides, missing base, and cycles. Build + `internal/specialist`/`internal/cli` green.

### 🟠 Recommend (consistency/robustness) — don't fail the whole Load on one bad `extends`
`resolveExtends` returns an error that makes `Load` return `LoadResult{}, err`, so a **single** specialist with a typo'd `extends:` (missing base or a cycle) makes `zero specialist list`/`show` error out and surface **zero** specialists — including all the valid builtins. It does print a clear message (good), but this is inconsistent with the **skip-and-warn** handling you added in #100 for unreadable/invalid manifest files (`loadDirectory` keeps the rest + warns). Consider downgrading a per-specialist extends failure to dropping that one specialist with a warning, so one broken chain doesn't hide everything else. (If fail-fast on structural config errors is the intended call, that's defensible — just flagging the inconsistency.)

### 🟡 Minor
- **`tools` replace + empty-means-all**: a child with empty `tools` inherits the base's tools (good), but can't express "I want *all* tools" when its base restricts them (empty = inherit, not all). Edge case; worth a doc note if it matters.
- **nit**: the resolved manifest keeps its `extends` field (shown in `show`/summaries) even though it's already flattened — informational, fine, just noting it's not cleared.

Nice, focused slice. 👍

PR #111, opened 2026-06-07
Add specialist execution foundation

anandh8x:
Summary:
- adds specialist prompt envelopes for fresh and resumed child runs
- adds Executor arg builders for fresh and resumed zero exec invocations
- supports generated child session IDs, prompt-file fallback, model/reasoning inheritance, resolved tool allowlists, depth/tag metadata, parent session/tool linkage, and session titles
- adds stream-json parsing/summarization and final tools.Result mapping for child output/error diagnostics

Self-review:
- intentionally does not wire the user-facing Task tool or spawn child processes yet
- keeps this as the bridge between resolved manifests and the later runtime executor
- validates generated/resume session IDs before handing args to zero exec

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/specialist
- GOCACHE=/tmp/zero-go-cache go test ./...
- GOCACHE=/tmp/zero-go-cache GOOS=windows go test -c ./internal/specialist
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Specialist session management with automatic, normalized session IDs and resume support.
  * Secure handling of large prompts with automatic file-based offloading when needed.
  * Constructed system and follow-up prompt wrappers for clearer specialist instructions.
  * Stream processing of specialist output into summarized results (text deltas, tools used, errors, exit status).
  * Stronger input validation and clearer error reporting for malformed or invalid inputs.

* **Tests**
  * Added comprehensive tests covering execution args, resume behavior, prompt offloading, stream parsing, and error cases.

PR #113, opened 2026-06-07
Add specialist Task runtime

anandh8x:
Summary:
- adds the LLM-facing Task tool and registers it in the TUI/top-level runtime
- runs specialists as foreground child zero exec processes with stream-json parsing, prompt-file cleanup, cwd/model/reasoning inheritance, and explicit tool allowlists
- persists child session metadata from calling-session/calling-tool-use flags and hides Task inside specialist child runs

Scope:
- foreground Task only; background TaskOutput/TaskStop, lifecycle hooks, usage rollup, and create/delete CLI management stay for follow-up PRs
- headless zero exec only exposes Task when explicitly unsafe/high autonomy; normal headless runs and specialist child runs do not expose it

Self-review:
- verified no shell command string construction for child args; process spawn uses argv slices
- tightened default allowlist so child runs always pass --enabled-tools, avoiding empty-tools == all-tools behavior
- added session metadata regression so child sessions preserve parent id, task id, tag, depth, and tool-use id

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli ./internal/agent ./internal/tools ./internal/tui ./internal/sessions
- git diff --cached --check
- GOCACHE=/tmp/zero-go-cache GOOS=windows go test -c ./internal/specialist
- GOCACHE=/tmp/zero-go-cache go build ./cmd/zero

## Summary by CodeRabbit

* **New Features**
  * Added a Task specialist tool and automatic specialist tool registration into interactive and exec flows.
  * Agent runs now receive richer session/context metadata (session IDs, model, reasoning effort, depth, working directory) propagated end-to-end.
  * Tool advertisement in auto-permission mode widened to include select advertise-eligible tools.

* **Tests**
  * Expanded tests for specialist execution, session metadata persistence, advertisement behavior, and prompt-file handling/cleanup.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -166,6 +166,12 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 		PermissionMode:    string(permissionMode),
 		Autonomy:          options.Autonomy,
 		Sandbox:           options.Sandbox,
+		ToolCallID:        call.ID,
+		SessionID:         options.SessionID,
+		Model:             options.Model,
+		ReasoningEffort:   options.ReasoningEffort,
+		Depth:             options.Depth,
+		Cwd:               options.Cwd,
 		// Note: we no longer rely on OnSandboxDecision callback for capture here
 		// (it is still supported for other observers and is invoked asynchronously in the registry).
 		// The sandbox decision (if any) is now returned synchronously on the Result for permission event building.
@@ -499,7 +505,7 @@ func ToolAdvertised(tool tools.Tool, permissionMode PermissionMode) bool {
 		return false
 	}
 	if permissionMode == PermissionModeAuto {
-		return tool.Safety().Permission == tools.PermissionAllow
+		return tool.Safety().Permission == tools.PermissionAllow || tool.Safety().AdvertiseInAuto
 	}
 	return true
 }
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -85,11 +85,15 @@ type Options struct {
 	MaxTurns int
 	// Specialist/sub-agent metadata is carried through exec now and consumed by
 	// the specialist runtime in later slices.
+	SessionID           string
 	CallingSessionID    string
 	CallingToolUseID    string
 	Tag                 string
 	Depth               int
 	SessionTitle        string
+	Model               string
+	ReasoningEffort     string
+	Cwd                 string
 	Registry            *tools.Registry
 	PermissionMode      PermissionMode
 	Autonomy            string
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/Gitlawb/zero/internal/agent"
@@ -16,6 +17,7 @@ import (
 	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/selfverify"
 	"github.com/Gitlawb/zero/internal/sessions"
+	"github.com/Gitlawb/zero/internal/specialist"
 	"github.com/Gitlawb/zero/internal/tools"
 	"github.com/Gitlawb/zero/internal/tui"
 	"github.com/Gitlawb/zero/internal/update"
@@ -274,6 +276,9 @@ func runInteractiveTUI(stderr io.Writer, deps appDeps) int {
 	}
 
 	registry := newCoreRegistry(workspaceRoot)
+	if err := registerSpecialistTools(registry, workspaceRoot); err != nil {
+		return writeAppError(stderr, "failed to initialize specialist tools: "+err.Error(), 1)
+	}
 	mcpRuntime, err := registerMCPToolsForWorkspace(context.Background(), workspaceRoot, registry, deps, mcp.AutonomyLow)
 	if err != nil {
 		return writeAppError(stderr, err.Error(), 1)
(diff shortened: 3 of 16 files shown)


gnanam1990 (changes requested):
## Review — strong, security-conscious runtime; one privilege-binding fix requested

This is a well-architected slice and it cleanly closes the posture concerns I raised on #111. Verified:
- **No shell injection** — child spawn is `osexec.CommandContext(ctx, binary, args...)` (argv slice), and `ctx` cancellation kills the child. 👍
- **Recursion is bounded structurally, not just by a counter** — children always run with `--tag specialist`, and `shouldRegisterExecSpecialistTools` returns false for that tag, so a specialist child never gets the `Task` tool registered and cannot nest. Robust.
- **Safe-by-default tools** — `defaultToolSelection = ["read-only"]`; the build errors if a resolved allowlist is empty and always passes `--enabled-tools`, so the "empty == all tools" footgun from #111 is gone.
- **Gating** — `Task` is exposed in headless exec only for unsafe/high autonomy (not normal or child runs); in the TUI it's `AdvertiseInAuto` but `PermissionPrompt`, so the user is prompted before a spawn. `defer cleanupPromptFile` handles the temp-file cleanup I flagged on #111.
- Session metadata (parent id, tool-use id, tag, depth, agent name) is persisted with a regression test. Build + `go test ./...` + `GOOS=windows build` green.

### 🟠 Please address — resume binds the allowlist to the caller-supplied name, not the session
`runResume` → `loadManifest(params.Name)` → `BuildResumeArgs` applies **`params.Name`'s** tool allowlist (and model) to the resumed session, regardless of which specialist originally created it. A caller can resume a session that was created under a read-only specialist while passing a more-privileged `name`, so the resumed agent continues that session's accumulated context with a *different, broader* toolset — a confused-deputy / constraint-bypass (this is also CodeRabbit's `exec.go:230`).

The fix is feasible because the session already persists `AgentName`: on resume, look up the resumed session's original specialist and bind the allowlist/model to **it**, or reject when `params.Name` doesn't match the stored agent. A regression test (resume under a different name must not widen tools) would lock it in.

### 🟡 Minor
- **ST1005** (CodeRabbit `task_tool.go:121`): error strings are capitalized (`"Task requires name"`, `"Task %s must be a string"`). Lowercase them (`"task requires a name"`, etc.).

### Non-blocking observations
- **Depth is tracked but not enforced** — recursion is prevented entirely by the tag-gate; `--depth`/`options.Depth` is informational. A belt-and-suspenders check (refuse `Task` when `Depth >= N`) would keep the bound if the tag logic ever changes.
- **No per-Task timeout** — a runaway child blocks the parent until the run ctx cancels (a user interrupt does kill it via `CommandContext`). Fine for foreground; noting it since background `TaskStop` is deferred.
- **stdout is fully buffered** in memory before `ParseStream`; fine for bounded foreground output, but a streaming parse could come with the background-execution work.

Really nice work on the containment model — the tag-gate + read-only default is exactly the right shape. Just the resume binding to close. 👍

<sub>🤖 Verified review via Claude Code (build/test/-race-relevant/windows run locally).</sub>

Vasanthdev2004:
### Blockers

None found on the latest head.

### Non-Blocking

- The original resume-identity concern is fixed in `fd5e6ff`: resume now loads the stored specialist child session, rejects non-specialist sessions, and rejects `name` mismatches before building the resumed child allowlist.
- Depth is still informational because child runs are prevented from seeing `Task` by the `--tag specialist` gate. That's okay for this foreground slice, but a hard max-depth check would be useful if nested/background specialist work lands later.

### Looks Good

- `Task` is prompt-gated in TUI auto mode and only registered for top-level headless unsafe/high-autonomy runs; specialist children are tag-gated so they cannot recursively spawn more specialists.
- Child process launch uses argv slices, passes explicit `--enabled-tools`, preserves cwd/model/reasoning/session lineage, and cleans spilled prompt files after the child exits.
- Resume is now bound to persisted child-session identity/agent name, with regression coverage for mismatched specialist names.
- Validation passed locally: targeted specialist/CLI/agent/tools/TUI/sessions tests, `go test ./...`, Windows specialist compile, `go build ./cmd/zero`, CLI help/specialist smokes, and `git diff --check`. GitHub checks are green.

---

**Verdict: Approve** — Clean foreground Task runtime after the resume allowlist binding fix.

Vasanthdev2004 (approved):
### Blockers

None found on the latest head.

### Non-Blocking

- The original resume-identity concern is fixed in `fd5e6ff`: resume now loads the stored specialist child session, rejects non-specialist sessions, and rejects `name` mismatches before building the resumed child allowlist.
- Depth is still informational because child runs are prevented from seeing `Task` by the `--tag specialist` gate. That's okay for this foreground slice, but a hard max-depth check would be useful if nested/background specialist work lands later.

### Looks Good

- `Task` is prompt-gated in TUI auto mode and only registered for top-level headless unsafe/high-autonomy runs; specialist children are tag-gated so they cannot recursively spawn more specialists.
- Child process launch uses argv slices, passes explicit `--enabled-tools`, preserves cwd/model/reasoning/session lineage, and cleans spilled prompt files after the child exits.
- Resume is now bound to persisted child-session identity/agent name, with regression coverage for mismatched specialist names.
- Validation passed locally: targeted specialist/CLI/agent/tools/TUI/sessions tests, `go test ./...`, Windows specialist compile, `go build ./cmd/zero`, CLI help/specialist smokes, and `git diff --check`. GitHub checks are green.

---

**Verdict: Approve** — Clean foreground Task runtime after the resume allowlist binding fix.

gnanam1990 (approved):
## Re-review — resume hardening looks correct, approving

`Harden specialist resume identity` (fd5e6ff) fully addresses the blocker:
- `runResume` now loads the session and binds the specialist to the session's persisted `AgentName`, **not** the caller-supplied `name`. A name mismatch is rejected (`resume session %q belongs to specialist %q, not %q`), so a session created under a read-only specialist can no longer be resumed under a more-privileged one — the allowlist/model follow the session's original identity.
- Extra hardening I didn't ask for but like: `resumeSession` also rejects anything that isn't an actual specialist child session (`SessionKindChild` + `tag == specialist`) and validates the session id, so arbitrary/foreign sessions can't be resumed through `Task`.
- **ST1005** error strings lowercased (`task requires name`, etc.).
- New tests cover the mismatch/identity path. `go vet`, `go test ./...`, and `GOOS=windows build` all green.

That resolves everything I flagged. The earlier non-blocking notes (depth as a defense-in-depth backstop in addition to the tag-gate, per-Task timeout, streaming stdout) are fine to carry into the background-execution follow-up. **Approving.** 👍

<sub>🤖 Verified re-review via Claude Code (build/vet/test/windows run locally).</sub>

PR #115, opened 2026-06-07
Add specialist background task tools

anandh8x:
Summary:
- adds Task run_in_background support with background child launch, task metadata, and output-file tracking
- registers TaskOutput and TaskStop with a shared lazy background manager
- lets Task resume by session id without requiring a redundant name, while still rejecting background resume

Self-review:
- background manager creation is lazy so TUI startup and --list-tools do not write to the data dir
- prompt temp files are kept for background children and cleaned up on child exit/start failure
- TaskStop remains prompt-gated but visible in auto mode; TaskOutput is read-only

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli ./internal/background ./internal/agent ./internal/tools
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Run tasks in the background (async execution).
  * Stop running background tasks.
  * Retrieve background task output and status immediately or by polling.

* **Bug Fixes**
  * Tool registration now correctly propagates registration errors.

* **Tests**
  * Expanded test coverage for background execution paths, output retrieval/polling, task stopping, and parameter validation.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -336,8 +336,7 @@ func registerSpecialistTools(registry *tools.Registry, workspaceRoot string) err
 	if err != nil {
 		return err
 	}
-	specialist.RegisterTools(registry, specialist.Executor{Paths: paths})
-	return nil
+	return specialist.RegisterTools(registry, specialist.Executor{Paths: paths})
 }
 
 func shouldRegisterExecSpecialistTools(options execOptions) bool {
--- a/internal/cli/exec_test.go
+++ b/internal/cli/exec_test.go
@@ -207,9 +207,11 @@ func TestRunExecRegistersTaskOnlyForUnsafeTopLevelRuns(t *testing.T) {
 			if exitCode != exitSuccess {
 				t.Fatalf("exitCode = %d stdout=%s stderr=%s", exitCode, stdout.String(), stderr.String())
 			}
-			hasTask := strings.Contains(stdout.String(), "  Task ")
-			if hasTask != tc.wantTask {
-				t.Fatalf("Task visibility = %v, want %v; output:\n%s", hasTask, tc.wantTask, stdout.String())
+			for _, toolName := range []string{"Task", "TaskOutput", "TaskStop"} {
+				hasTool := strings.Contains(stdout.String(), "  "+toolName+" ")
+				if hasTool != tc.wantTask {
+					t.Fatalf("%s visibility = %v, want %v; output:\n%s", toolName, hasTool, tc.wantTask, stdout.String())
+				}
 			}
 		})
 	}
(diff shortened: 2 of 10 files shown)


gnanam1990 (approved):
## Review — solid background tools; one lifecycle gap to handle (approving)

Nicely done, and the parts that could have been subtle bugs are correct:
- **Shared manager is right** — `RegisterTools` builds one memoized, mutex-guarded `lazyBackgroundManager()` and threads the *same* closure into `Task`, `TaskOutput`, and `TaskStop`, so `TaskStop`/`TaskOutput` actually see what `Task` launched. Lazy creation means TUI startup / `--list-tools` don't touch the data dir. 👍
- **Detached launch** uses `osexec.Command` (not the tool-call `ctx`), with a wait-goroutine that sets the manager status and cleans the prompt file on exit — correct shape for background.
- **Reject background resume** (`specialist resume cannot run in background`), recursion still gated (background children run `--tag specialist` → none of the three tools are registered), `Kill` reuses the TOCTOU-hardened path from #108, `TaskOutput` is read-only and reads the manager's own confined `OutputFile` (task-id is just a map key, not a path), `TaskStop` is prompt-gated. Full suite + `-race` + `GOOS=windows build` green.

### 🟠 Please handle (or confirm it's the lifecycle-hooks follow-up) — background child shutdown
There's no `Setpgid` and no shutdown teardown, so the child's lifecycle is in an inconsistent middle:
- **Clean `zero` exit while a task runs** → the child is **orphaned** (reparented to init, keeps running); the wait-goroutine dies with the parent, so the manager status stays **`running` (stale)** and the **prompt temp file leaks** (`onExit` never fires).
- **Terminal Ctrl+C** → the child shares `zero`'s process group and is **killed** with it — so "background" doesn't survive an interrupt either.

Decide the intended semantics and handle teardown — e.g. on shutdown iterate `manager.List()` running tasks and `Kill` them (clean status + prompt cleanup), or detach with `Setpgid` and explicitly track/clean orphans + their prompt files. If this is the "lifecycle hooks" follow-up you scoped on #113, that's fine — just calling it out so the orphan-process + temp-leak + stale-status path doesn't ship unflagged.

### 🟡 Minor
- **SetPID-failure path** (`runBackground`): if `launchBackground` succeeds but `SetPID` fails, the prompt file is cleaned while the just-started child may still be reading it via `--file`, and the running child is neither killed nor marked. Rare (only if the task vanished post-`Register`), low severity.
- **`TaskOutput` reads the whole file** (`os.ReadFile`) — no tail/byte cap; a chatty long-running child's NDJSON can grow large. A tail/limit would be friendlier.
- **nit:** `task requires name` fires when *both* `name` and `resume` are empty — message should be `task requires name or resume`.

Approving — the architecture and the correctness-critical bits are right; the shutdown lifecycle is the one thing to close (here or in the lifecycle-hooks follow-up). 👍

<sub>🤖 Verified review via Claude Code (build/test/-race/windows run locally).</sub>

PR #116, opened 2026-06-07
Harden specialist background lifecycle

anandh8x:
Summary:
- adds a specialist Runtime owner for the lazy background manager
- closes running specialist background tasks when the CLI/TUI runtime exits
- tracks background prompt temp files and cleans them on child exit or runtime close
- adds background.Manager KillRunning for shutdown cleanup

Self-review:
- Runtime.Close does not create background storage if no background task was used
- cleanup is wired into both interactive TUI and headless exec registration paths
- this intentionally does not implement hook command execution; the hooks package currently only supports selection/audit primitives

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/background ./internal/specialist ./internal/cli
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Runtime abstraction added to manage specialist-tool lifecycles and track temporary prompt files.

* **Improvements**
  * Background task handling and shutdown improved to ensure running tasks are stopped and resources cleaned.
  * Specialist tool registration now provides a runtime handle so background resources are reliably closed.

* **Tests**
  * Added tests covering task exit transitions, bulk-kill behavior, runtime close semantics, and prompt-file cleanup.

### DIFF
--- a/internal/background/manager.go
+++ b/internal/background/manager.go
@@ -198,6 +198,30 @@ func (manager *Manager) UpdateStatus(taskID string, status Status, exitCode int)
 	return nil
 }
 
+func (manager *Manager) MarkExited(taskID string, status Status, exitCode int) error {
+	taskID = strings.TrimSpace(taskID)
+	if status != StatusCompleted && status != StatusError {
+		return fmt.Errorf("invalid background task exit status %q", status)
+	}
+
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	task, ok := manager.tasks[taskID]
+	if !ok {
+		return fmt.Errorf("background task not found: %s", taskID)
+	}
+	if task.Status != StatusRunning {
+		return nil
+	}
+	task.Status = status
+	task.ExitCode = exitCode
+	if task.CompletedAt.IsZero() {
+		task.CompletedAt = manager.now()
+	}
+	manager.tasks[taskID] = task
+	return nil
+}
+
 func (manager *Manager) Get(taskID string) (Task, bool) {
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
--- a/internal/background/manager.go
+++ b/internal/background/manager.go
@@ -221,6 +221,19 @@ func (manager *Manager) Kill(taskID string) error {
 	return manager.markKilledIfStillRunning(taskID, pid)
 }
 
+func (manager *Manager) KillRunning() error {
+	var errs []error
+	for _, task := range manager.List() {
+		if task.Status != StatusRunning {
+			continue
+		}
+		if err := manager.Kill(task.ID); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
+
 func (manager *Manager) killTarget(taskID string) (int, error) {
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
--- a/internal/background/manager_test.go
+++ b/internal/background/manager_test.go
@@ -177,6 +177,56 @@ func TestManagerKillDoesNotClobberCompletedTask(t *testing.T) {
 	}
 }
 
+func TestManagerKillRunningStopsOnlyRunningTasks(t *testing.T) {
+	now := sequenceClock(
+		time.Date(2026, 6, 7, 9, 0, 0, 0, time.UTC),
+		time.Date(2026, 6, 7, 9, 0, 1, 0, time.UTC),
+		time.Date(2026, 6, 7, 9, 0, 2, 0, time.UTC),
+	)
+	killed := []int{}
+	manager, err := NewManagerWithOptions(ManagerOptions{
+		RootDir: t.TempDir(),
+		Now:     now,
+		KillProcess: func(pid int) error {
+			killed = append(killed, pid)
+			return nil
+		},
+	})
+	if err != nil {
+		t.Fatalf("NewManagerWithOptions returned error: %v", err)
+	}
+	if _, err := manager.Register(RegisterInput{TaskID: "running_1", Type: "specialist", PID: 11}); err != nil {
+		t.Fatalf("Register running_1 returned error: %v", err)
+	}
+	if _, err := manager.Register(RegisterInput{TaskID: "done", Type: "specialist", PID: 22}); err != nil {
+		t.Fatalf("Register done returned error: %v", err)
+	}
+	if err := manager.UpdateStatus("done", StatusCompleted, 0); err != nil {
+		t.Fatalf("UpdateStatus returned error: %v", err)
+	}
+	if _, err := manager.Register(RegisterInput{TaskID: "running_2", Type: "specialist", PID: 33}); err != nil {
+		t.Fatalf("Register running_2 returned error: %v", err)
+	}
+
+	if err := manager.KillRunning(); err != nil {
+		t.Fatalf("KillRunning returned error: %v", err)
+	}
+
+	if !reflect.DeepEqual(killed, []int{33, 11}) {
+		t.Fatalf("killed pids = %#v", killed)
+	}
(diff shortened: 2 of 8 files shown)


Vasanthdev2004 (changes requested):
### Blockers

- `internal/specialist/exec.go:314` / `internal/specialist/runtime.go:91`: shutdown cleanup kills running background tasks, but the child wait callback can still overwrite the terminal `killed` state with `error` after the stopped process exits non-zero. That leaves `TaskOutput` reporting a runtime-close/TaskStop cleanup as an error instead of a deliberate kill. I reproduced this on the PR head with a focused runtime-close probe: `Runtime.Close()` marked the task `killed`, then the simulated child exit callback changed it to `error`.

### Non-Blocking

- `internal/specialist/output_tool.go`: still reads the full task output file on every poll. A tail/limit can come later, but it would make long-running background tasks safer to inspect.

### Looks Good

- Runtime ownership keeps the background manager lazy; `Runtime.Close()` does not create storage when no background task was used.
- Interactive and headless exec paths now defer specialist runtime cleanup, and prompt temp files are tracked for background children.
- Local validation passed: `go test -count=1 ./internal/background ./internal/specialist ./internal/cli`, full `go test -count=1 -p 1 ./...`, `go build ./cmd/zero`, `zero exec --list-tools --auto high`, and `git diff --check`.
- CI is green at the reviewed head.

---

**Verdict: Changes Requested** — preserve the explicit killed terminal state after runtime shutdown/TaskStop before shipping this lifecycle cleanup.

gnanam1990 (approved):
## Re-review — this resolves the #115 lifecycle gap, approving

Exactly the fix I asked for on #115, and cleanly done:
- **Orphans on shutdown** — `Runtime.Close()` → `manager.KillRunning()` terminates still-running background tasks (reusing #108's TOCTOU-safe `Kill`, and `KillRunning` only touches `StatusRunning` tasks).
- **Prompt-file leak** — prompt files are tracked on launch, untracked+cleaned on child exit (normal/error paths), and cleaned by `Close()` if the runtime exits while a child is still running. `Close` snapshots+clears the map *before* `KillRunning`, so a killed child's `onExit` can't double-clean. No leak either way.
- **Shared-manager correctness preserved** — `RegisterTools` creates one `Runtime` and wires `Task` (via the executor) and `Output`/`Stop` (via `runtime.Manager`) to the same instance.
- **Lazy** — `Manager()`/`Close()` never create background storage if no task used it.
- **Wired via `defer closeSpecialistRuntime(...)` on both the interactive TUI and headless exec paths** — so it fires on normal return and on Bubble Tea's Ctrl-C (which returns from `p.Run`).
- Tests pin it: `TestRuntimeCloseKillsRunningTasksAndCleansPromptFiles` (asserts `StatusKilled` + prompt file removed) and `TestManagerKillRunningStopsOnlyRunningTasks`. `go test ./...`, `-race`, and `GOOS=windows build` green.

**Approving.** 👍

### Non-blocking observation
A headless run that receives an *uncaught* SIGINT (e.g. `zero exec` with a background task, terminal Ctrl-C) won't run the `defer` (Go doesn't run defers on uncaught signals), so `Close` is skipped there. In practice the background child shares the process group and dies with the parent anyway, so the only residual is a prompt-temp-file leak on hard SIGINT. A signal handler that calls `Close` would close that last gap, but it's a minor edge — fine to leave or fold into a later pass.

<sub>🤖 Verified re-review via Claude Code (vet/-race/windows/full-suite run locally).</sub>

gnanam1990 (changes requested):
## Correcting my earlier approve — @Vasanthdev2004's race is real

Good catch by @Vasanthdev2004, and I have to walk back my approve: I traced the `Close`/`KillRunning`/prompt-file paths carefully but missed the kill-vs-`onExit` status race. Confirmed on the head:

`background.UpdateStatus` sets `task.Status` **unconditionally** (no terminal-state guard). The background child's wait goroutine (`launchBackgroundProcess` → `onExit`) calls `manager.UpdateStatus(taskID, StatusError, exitCode)` when the process exits non-zero. So on shutdown:

1. `Runtime.Close()` → `KillRunning()` → `Kill` sets the task to `killed`.
2. The killed process exits non-zero → the still-running `onExit` fires → `UpdateStatus(StatusError)` **overwrites `killed` → `error`**.

So `TaskOutput`/status reports a deliberate runtime-close/`TaskStop` as a failure. Same class as the #108 `Kill` TOCTOU, but on the `onExit` side.

### Suggested fix
Make `onExit` not clobber a terminal state — e.g. a `UpdateStatusIfRunning` that only transitions when the current status is still `StatusRunning` (mirrors `markKilledIfStillRunning`). Then a kill sets `killed`, and the subsequent non-zero `onExit` is a no-op. A focused test: register → `Kill`/`KillRunning` → invoke the exit callback with a non-zero code → assert status stays `killed`.

The rest of the PR is solid (the lifecycle cleanup itself, prompt-file tracking, lazy runtime, defer wiring) — this is the one correctness blocker to close.

### Non-blocking (also noted by @Vasanthdev2004)
`output_tool.go` still `os.ReadFile`s the whole output on every poll — a tail/limit can come later.

Apologies for the premature approve — requesting changes to reflect the real blocker.

gnanam1990 (approved):
## Re-review — race fixed, approving

`Preserve killed background task state` (56dbdc6) closes the kill-vs-`onExit` race cleanly:
- New `MarkExited(taskID, status, exitCode)` only transitions **from `StatusRunning`** (`if task.Status != StatusRunning { return nil }`) and only accepts `Completed`/`Error`. `onExit` now calls it instead of the unconditional `UpdateStatus`.
- So a `Kill`/`KillRunning` sets `killed`, and the killed child's subsequent non-zero exit is a **no-op** — `killed` (ExitCode -1) is preserved, and `TaskOutput` no longer misreports a deliberate stop as an error.
- Normal completion still works (Running → Completed/Error via `MarkExited`).
- Tests pin it: `TestManagerMarkExitedDoesNotClobberKilledTask` and `TestRuntimeClosePreservesKilledTaskAfterChildExit` (Close kills → child-exit callback → status stays `killed`). `go test ./...`, `-race`, and `GOOS=windows build` green.

That resolves the blocker @Vasanthdev2004 caught (and that I missed on my first pass). Combined with the lifecycle cleanup itself, this is a complete, well-tested fix for the #115 background-lifecycle gap. **Approving.** 👍

The earlier non-blocking notes stand for a later pass (headless uncaught-SIGINT bypasses the `defer`; `TaskOutput` reads the full file per poll).

<sub>🤖 Verified re-review via Claude Code (vet/-race/windows/full-suite run locally).</sub>

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- `internal/specialist/output_tool.go`: background output polling still reads the full output file each time. Not introduced by the fix, but a future tail/limit would make chatty long-running specialist tasks safer to inspect.

### Looks Good

- The prior lifecycle blocker is fixed: background child exits now call `Manager.MarkExited`, which only transitions running tasks and preserves an explicit `killed` terminal state after `Runtime.Close`, `KillRunning`, or `TaskStop`.
- Regression coverage now exercises both the manager-level killed-state preservation and the real specialist runtime close + child exit callback path.
- PR-head validation passed: focused regression tests, targeted background/specialist/CLI tests, full `go test -count=1 -p 1 ./...`, `go build ./cmd/zero`, specialist CLI smokes, Task/TaskOutput/TaskStop tool listing, and `git diff --check`.
- Latest-main test merge from current `origin/main` was clean and passed the same targeted/full Go validation, build, tool-listing smoke, and diff check.

---

**Verdict: Approve** — clean lifecycle hardening; the prior killed-status overwrite is resolved.

Vasanthdev2004 (approved):
Approved after rereview: prior killed-status overwrite blocker is fixed and validation passed on PR head plus latest-main test merge.

PR #117, opened 2026-06-07
Add specialist authoring tools

anandh8x:
Summary:
- Adds specialist Storage for safe markdown create/delete with user/project locations, overwrite handling, and symlink overwrite protection.
- Adds zero specialist create/delete/rm/edit with JSON-friendly output and project/user targeting.
- Registers GenerateSpecialist so the parent model can write a validated specialist profile from tool arguments, while specialist manifests explicitly forbid Task/TaskOutput/TaskStop/GenerateSpecialist.

Notes:
- GenerateSpecialist does not make a nested provider call; the current parent model supplies the designed name/description/system prompt as tool arguments and the tool validates/writes the file.
- Child specialist runs still do not receive specialist authoring tools because RegisterTools is skipped for --tag specialist.

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/specialist ./internal/cli
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Manage specialist profiles from the CLI: create, edit (opens editor), delete, list, show, path
  * Project-scoped specialist generator tool with name/description/system-prompt defaults and tool selection
  * Persistent specialist storage: create/delete manifests in user/project locations
  * Improved tool parameter schemas (supports nested/array item schemas)

* **Bug Fixes**
  * Reject forbidden specialist tools and validate locations/names

* **Tests**
  * Added end-to-end and unit tests for specialist CLI, storage, and generator behaviors

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -491,6 +491,9 @@ func propertyToRuntimeMap(property tools.PropertySchema) map[string]any {
 	if property.Default != nil {
 		schema["default"] = property.Default
 	}
+	if property.Items != nil {
+		schema["items"] = propertyToRuntimeMap(*property.Items)
+	}
 	if property.Minimum != nil {
 		schema["minimum"] = *property.Minimum
 	}
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -49,6 +49,7 @@ type appDeps struct {
 	inspectChanges       func(context.Context, zerogit.InspectOptions) (zerogit.ChangeSummary, error)
 	commitChanges        func(context.Context, zerogit.CommitOptions) (zerogit.CommitResult, error)
 	runTUI               func(context.Context, tui.Options) int
+	runEditor            func(string) error
 	checkUpdate          func(context.Context, update.Options) (update.Result, error)
 	now                  func() time.Time
 }
@@ -113,6 +114,7 @@ func defaultAppDeps() appDeps {
 		inspectChanges:   zerogit.Inspect,
 		commitChanges:    zerogit.Commit,
 		runTUI:           tui.Run,
+		runEditor:        openEditor,
 		checkUpdate:      update.Check,
 		now:              time.Now,
 	}
@@ -392,7 +394,7 @@ Commands:
   search     Search persisted local Zero session events
   find       Alias for search
   sessions   Inspect local Zero session lineage
-  specialist Inspect local Zero specialist profiles
+  specialist Manage local Zero specialist profiles
   plugins    Inspect local Zero plugin manifests
   hooks      Inspect Zero hook configuration
   mcp        Manage MCP backend settings
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -252,6 +252,9 @@ func fillAppDeps(deps appDeps) appDeps {
 	if deps.runTUI == nil {
 		deps.runTUI = defaults.runTUI
 	}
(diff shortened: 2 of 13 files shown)


Vasanthdev2004 (changes requested):
### Blockers

- `internal/specialist/generate_tool.go:74` / `internal/specialist/generate_tool.go:192`: `GenerateSpecialist` is advertised to the agent and defaults writes to the user specialist directory. In `zero exec --auto high` / unsafe mode that means a prompt-injected model can persist a global specialist outside the current workspace without an additional user-vs-project boundary. Because these profiles are loaded in later runs and can declare broader tool categories, model-driven generation should either stay project-scoped by default, require an explicit writable location that is still sandbox-evaluated, or not be exposed to unsafe auto execution as a normal prompt-gated tool.
- `internal/cli/specialist.go:318`: `zero specialist edit` resolves the validated profile path and then uses `os.Stat` before handing it to the editor, so it follows symlinks. The loader explicitly skips symlink manifests and create refuses symlink overwrite, but a project/user `name.md` symlink can still make `specialist edit --project name` open an arbitrary target. Use `Lstat`/symlink rejection here too before invoking the editor.

### Non-Blocking

- `internal/specialist/generate_tool.go:59`: the `tools` array schema does not declare string item types. Runtime validation catches non-strings, so this is schema quality rather than a blocker.

### Looks Good

- Specialist manifests now reject recursive specialist-control tools (`Task`, `TaskOutput`, `TaskStop`, `GenerateSpecialist`) during tool resolution.
- Child specialist executions still skip specialist tool registration via the `--tag specialist` gate.
- Local validation passed: `go test -count=1 ./internal/specialist ./internal/cli`, full `go test -count=1 -p 1 ./...`, `go build ./cmd/zero`, specialist CLI smokes, Task/TaskOutput/TaskStop/GenerateSpecialist tool listing, and `git diff --check`.
- PR base matches current `origin/main`, and CI is green.

---

**Verdict: Changes Requested** — the authoring surface needs the same workspace/symlink safety guarantees as the rest of the specialist loader before it ships.

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- `internal/cli/specialist.go`: help text still says create/delete/edit are for the user specialist profile, even though project targeting is now supported via `--project`. Small docs polish only.

### Looks Good

- The prior global-persistence blocker is fixed: `GenerateSpecialist` is now project-scoped, rejects `location: user`, writes under the workspace project specialist directory, and declares string item schema for `tools`.
- The prior symlink-edit blocker is fixed: `zero specialist edit` now uses `os.Lstat` and refuses symlink specialist files before invoking the editor.
- Specialist-control tools remain hidden from child specialist runs via the `--tag specialist` registration gate, and generated manifests still reject recursive tools (`Task`, `TaskOutput`, `TaskStop`, `GenerateSpecialist`).
- Local validation passed: focused authoring tests, targeted specialist/CLI tests, full `go test -count=1 -p 1 ./...`, `go build ./cmd/zero`, CLI smokes, top-level tool listing, child-specialist tool-hiding smoke, and `git diff --check`.
- PR base matches current `origin/main`, and CI is green.

---

**Verdict: Approve** — clean specialist authoring hardening; the previous workspace/symlink blockers are resolved.

Vasanthdev2004 (approved):
Approved after rereview: the specialist authoring workspace/symlink blockers are fixed and validation passed.

PR #118, opened 2026-06-07
Persist background specialist tasks

anandh8x:
## Summary
- persist background task metadata as JSON beside each NDJSON output file
- reload persisted tasks when the background manager starts so TaskOutput/TaskStop can find existing tasks
- document specialist manifests, Task tools, and background task state

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/background ./internal/specialist
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Background tasks now persist to disk and reload on startup, preserving state and outputs; running-but-stale tasks are marked errored to avoid signaling stale PIDs
  * New "zero specialist" command suite for creating, listing, showing, editing, and deleting specialists

* **Documentation**
  * Added comprehensive Specialists docs covering manifest format, scoping, CLI usage, and lifecycle behavior

* **Tests**
  * Added tests verifying persistence, reload behavior, and output-tool integration

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ zero doctor [--connectivity] [--json]        # health checks
 zero config [--json]                          # inspect resolved configuration
 zero sandbox policy [--json]                 # inspect sandbox backend support
 zero serve --mcp [-C <path>]                  # expose Zero read-only tools over MCP stdio
+zero specialist list|show|create|edit|delete  # manage local specialist sub-agents
 zero update --check [--json --target windows-x64] # check for a newer release
 ```
 
@@ -124,6 +125,8 @@ runtimes (Ollama, gateways, etc.) plug in the same way.
 | `update_plan` | maintain a live task plan | plan |
 | `write_file` · `edit_file` · `apply_patch` | create & modify files | write (gated) |
 | `bash` | run shell commands | shell (gated) |
+| `Task` · `TaskOutput` · `TaskStop` | delegate to specialist sub-agents | shell/read (gated by tool) |
+| `GenerateSpecialist` | create project-local specialist manifests | write (gated) |
 
 Write/shell tools route through the permission policy before any side effect.
 
@@ -199,6 +202,7 @@ and [`docs/UPDATE.md`](docs/UPDATE.md) for the update flow.
 
 - [Product Requirements (PRD)](docs/PRD.md) — vision, goals, full feature spec, roadmap
 - [Stream-JSON protocol](docs/STREAM_JSON_PROTOCOL.md) — headless I/O contract
+- [Specialists](docs/SPECIALISTS.md) — sub-agent manifests, Task tools, and background task state
 - [Headless exec PRD](docs/M1_HEADLESS_EXEC_PRD.md)
 - [npm wrapper smoke checklist](docs/NPM_WRAPPER_SMOKE.md)
 - [Performance](docs/PERFORMANCE.md) · [Install](docs/INSTALL.md) · [Update](docs/UPDATE.md)
--- a/docs/SPECIALISTS.md
+++ b/docs/SPECIALISTS.md
@@ -134,3 +134,7 @@ Each task has:
 
 Persisted metadata lets a new background manager instance read completed task
 output or stop a still-running task by id.
+
+If Zero is restarted while a background task is still marked `running`, the new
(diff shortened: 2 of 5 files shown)


gnanam1990 (changes requested):
## Review — solid persistence mechanics; two robustness/safety fixes requested

The persistence itself is well done:
- **Atomic writes** (tmp + `os.Rename`, `0600`), and leftover `*.tmp-<nano>` files are skipped on load (not `.json`).
- **Path safety on load** — `validTaskID` on the filename, `OutputFile` re-confined through `outputFile()`, and **symlink metadata files are refused**. `normalizeLoadedTask` also checks the id matches the filename and validates type/status/pid. 👍
- Tests cover persist→reload, killed-survives-reload, and invalid-metadata rejection; docs added.

### 🟠 One bad metadata file disables *all* background tools
`loadTasks` returns an error — and `NewManager` then fails — on **any** single malformed / invalid-id / symlink / unreadable `.json` (confirmed intentional by `TestManagerRejectsInvalidPersistedMetadata`). Because the manager is created lazily, one stale/corrupt/format-drifted file in `~/.local/share/zero/background/` makes **every** `Task`/`TaskOutput`/`TaskStop` call fail until the user manually deletes it. This is more likely over time since completed-task metadata is never cleaned up (see minor below), and a future schema change to `Task` would brick old files. Please **skip-and-warn per file** (log + `continue` past a bad entry, keep the rest) rather than failing the whole manager — matching the skip-and-warn the specialist `loadDirectory` uses.

### 🟠 Stale PID on reload → `TaskStop` can kill an unrelated process
A task persisted as `running` (alive at crash/exit) reloads with the **same `Status=running` and `PID`**, but that process is from a dead generation — across a restart the PID is gone or reused. `TaskStop` → `manager.Kill` → `killProcess(stalePID)` would then signal **whatever now holds that PID**. Same hazard as the #108 `Kill` TOCTOU, but guaranteed-stale across a restart. Please reconcile on load: mark a reloaded `running` task as a non-killable state (e.g. `error`/orphaned, since its owner is gone), or have `Kill` refuse to signal a PID it can't prove belongs to this manager generation. A regression test (persist `running` → reload → `Kill`/`TaskStop` must not `killProcess` the stale PID) would lock it in.

### 🟡 Minor
- **No metadata cleanup** — completed/killed task `.json` files accumulate in the data dir indefinitely (and feed the failure mode above). Consider pruning on completion, or a cap/TTL.
- **Persist on every update under the lock** — `Register`/`SetPID`/`UpdateStatus`/`MarkExited` each do a synchronous tmp+rename while holding `mu`. Fine for low-frequency background tasks; just noting.

The mechanics and load-path hardening are good — these two are about surviving a restart safely. Requesting changes on the stale-PID kill in particular.

<sub>🤖 Verified review via Claude Code (build/vet/-race/windows/full-suite run locally).</sub>

Vasanthdev2004:
### Blockers

None found.

### Non-Blocking

- Completed/killed task metadata still accumulates indefinitely. Not blocking for this PR now that bad files are skipped, but a future cap/TTL or cleanup command would keep the background directory tidy.

### Looks Good

- The corrupt-metadata blocker is fixed: invalid, unreadable, bad-id, and symlink metadata entries are skipped with load warnings instead of failing `NewManager` and disabling all background tools.
- The stale-PID blocker is fixed: reloaded `running` tasks are marked `error`, PID is cleared, and the repaired metadata is persisted so `TaskStop` cannot signal a previous-generation PID.
- I verified both fixes with focused probes: good tasks still load when bad metadata exists, and reloaded running tasks do not call `killProcess` for the old PID.
- Local validation passed: focused reload probes, `go test -count=1 ./internal/background ./internal/specialist`, full `go test -count=1 -p 1 ./...`, `go build ./cmd/zero`, Task/TaskOutput/TaskStop tool-listing smoke, and `git diff --check`.
- PR base matches current `origin/main`, and CI is green.

---

**Verdict: Approve** — background task persistence now survives bad metadata and avoids stale-PID kills after restart.

Vasanthdev2004 (approved):
Approved after rereview: corrupt metadata is skipped and stale reloaded PIDs are no longer killable; validation passed.

gnanam1990 (approved):

PR #119, opened 2026-06-07
Tools: arg-tolerance, ask_user, secret scrubbing, structured results, hardening

gnanam1990:
## Module — tools runtime (arg-tolerance, ask_user, scrubbing, structured results)

First of the `tools → agent → tui shell` chain (the TUI shell sits on top of these). **Scope: `internal/tools`** (24 files, +1,680/−51) **+ a 1-line `internal/specialist` vocabulary update**. No new deps; this was 3-way merged onto main's specialist drift.

### Reconciled with main's drift (kept)
`RunOptions` tool-context fields, `Safety.AdvertiseInAuto`, the `optionsAwareTool` dispatch + `OnSandboxDecision`, `PropertySchema` (Items/Min/Max), and the base structured `Result` — all preserved.

### Added (runtime-core)
- **Arg-tolerance layer** (`argtolerance.go`): alias-aware, type-strict arg resolution + slice/bool/int coercion, adopted across all tools so weak models' non-canonical arg shapes are accepted without losing type-strictness.
- **`ask_user` tool** wired into `CoreReadOnlyTools`; **nested `PropertySchema` (Properties/Required)** for its object-typed questions (and `update_plan`).
- **Secret scrubbing at the registry boundary** (`scrubResultSecrets` via `internal/redaction`) on every tool-execution path incl. the `optionsAwareTool` dispatch; **structured `Result`** (`Display`/`Redacted`/`ChangedFiles`).
- **`mutation_targets`** (paths a mutating tool will write, for session checkpointing) + hardened `grep` (symlink confinement), `apply_patch`, `bash`, `edit`/`write`.

### Excluded per the split
The blueprint **`task` tool** (subagents — another dev's area; the merged specialist `Task` is separate) and the **`skill` tool** (ships with the later skills module), plus their subagent-only registry helpers (`Without`/`isolateForChild`).

### Cross-package note
`ask_user` is now a core tool, so `specialist.knownToolNames` gains `"ask_user"` to satisfy `TestKnownToolNamesMatchCoreRegistry` (the invariant that the specialist tool-vocabulary covers the core registry).

### Testing
`go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./internal/tools/`, and `GOOS=windows go build ./...` all green.

Part of decomposing #101 (draft); subagents excluded.

## Release Notes

* **New Features**
  * Added `ask_user` tool for collecting interactive clarifications and user feedback.

* **Improvements**
  * Tools now accept flexible parameter aliases for common inputs (e.g., `path`/`file`, `patch`/`diff`).
  * Enhanced argument handling accepts multiple input formats (strings, numbers, booleans).
  * File operations report changed files and display summaries for better visibility.
  * Bash tool provides early safety feedback for interactive commands.
  * Tool outputs automatically redact sensitive information.

### DIFF
--- a/internal/sandbox/risk.go
+++ b/internal/sandbox/risk.go
@@ -193,7 +193,12 @@ func firstArgString(args map[string]any, keys ...string) string {
 
 func requestPaths(request Request) []string {
 	paths := []string{}
-	for _, key := range []string{"path", "cwd", "file", "dir"} {
+	// Keep this aligned with the path-arg alias lists the tools accept (see
+	// aliasedStringArg in write_file/edit_file/read_file/grep/glob/list). The
+	// sandbox gates by arg-key name, so any alias a tool resolves but the sandbox
+	// does not inspect would let a model route a write/read around the
+	// workspace+symlink boundary.
+	for _, key := range []string{"path", "file", "file_path", "filepath", "filename", "cwd", "dir", "directory"} {
 		if value := argString(request.Args, key); value != "" {
 			paths = append(paths, value)
 		}
--- a/internal/specialist/manifest.go
+++ b/internal/specialist/manifest.go
@@ -103,7 +103,6 @@ var knownToolNames = map[string]bool{
 	"list_directory": true,
 	"glob":           true,
 	"grep":           true,
-	"ask_user":       true,
 	"write_file":     true,
 	"edit_file":      true,
 	"apply_patch":    true,
--- a/internal/specialist/manifest.go
+++ b/internal/specialist/manifest.go
@@ -103,6 +103,7 @@ var knownToolNames = map[string]bool{
 	"list_directory": true,
 	"glob":           true,
 	"grep":           true,
+	"ask_user":       true,
 	"write_file":     true,
 	"edit_file":      true,
 	"apply_patch":    true,
--- a/internal/tools/apply_patch.go
+++ b/internal/tools/apply_patch.go
@@ -35,7 +35,7 @@ func NewApplyPatchTool(workspaceRoot string) Tool {
 }
 
 func (tool applyPatchTool) Run(ctx context.Context, args map[string]any) Result {
-	patch, err := stringArg(args, "patch", "", true)
+	patch, err := aliasedStringArg(args, []string{"patch", "diff"}, "", true, false)
 	if err != nil {
 		return errorResult("Error: Invalid arguments for apply_patch: " + err.Error())
 	}
@@ -83,10 +83,41 @@ func (tool applyPatchTool) Run(ctx context.Context, args map[string]any) Result 
 		return errorResult("Error applying patch: " + message)
 	}
 
-	if relativeRoot == "." {
-		return okResult("Patch applied successfully.")
+	summary := "Patch applied successfully."
+	if relativeRoot != "." {
+		summary = "Patch applied successfully in " + relativeRoot + "."
 	}
-	return okResult("Patch applied successfully in " + relativeRoot + ".")
+	result := okResult(summary)
+	result.ChangedFiles = changedFilesFromPatch(relativeRoot, patch)
+	result.Display = Display{Summary: summary, Kind: "diff"}
+	return result
+}
+
+// changedFilesFromPatch extracts the unique, WORKSPACE-relative paths a patch
+// touches, reusing the same per-line parser used for validation. Patch paths are
+// relative to the apply cwd, so relativeRoot (the workspace-relative cwd, e.g.
+// "sub/dir", or "." for the workspace root) is prefixed so callers get true
+// workspace-relative paths regardless of cwd.
+func changedFilesFromPatch(relativeRoot string, patch string) []string {
+	seen := map[string]bool{}
+	var paths []string
+	for _, line := range strings.Split(strings.ReplaceAll(patch, "\r\n", "\n"), "\n") {
+		for _, path := range patchPathsFromLine(line) {
+			if path == "" || path == "/dev/null" {
+				continue
+			}
+			workspacePath := path
+			if relativeRoot != "" && relativeRoot != "." {
+				workspacePath = filepath.ToSlash(filepath.Join(relativeRoot, path))
+			}
+			if seen[workspacePath] {
+				continue
+			}
+			seen[workspacePath] = true
+			paths = append(paths, workspacePath)
+		}
+	}
+	return paths
 }
 
 func validatePatchPaths(root string, patch string) error {
@@ -138,7 +169,10 @@ func patchPathsFromLine(line string) []string {
 
 func stripPatchPrefix(path string) string {
 	path = strings.TrimSpace(path)
-	path = strings.TrimPrefix(path, "a/")
-	path = strings.TrimPrefix(path, "b/")
+	// A unified-diff path carries exactly one of the a/ or b/ prefixes; strip a
+	// single one so a real directory literally named "a" or "b" is preserved.
+	if strings.HasPrefix(path, "a/") || strings.HasPrefix(path, "b/") {
+		path = path[2:]
+	}
 	return filepath.ToSlash(path)
 }
--- a/internal/tools/args.go
+++ b/internal/tools/args.go
@@ -3,6 +3,8 @@ package tools
 import (
 	"fmt"
 	"math"
+	"strconv"
+	"strings"
 )
 
 func stringArg(args map[string]any, key string, fallback string, required bool) (string, error) {
@@ -28,17 +30,41 @@ func stringArgWithEmpty(args map[string]any, key string, fallback string, requir
 	return text, nil
 }
 
+// boolArg reads a boolean argument, tolerating the string/number forms models
+// commonly emit ("true"/"false", "yes"/"no", "on"/"off", 1/0) since not every
+// model sends a JSON boolean.
 func boolArg(args map[string]any, key string, fallback bool) (bool, error) {
 	value, ok := args[key]
 	if !ok || value == nil {
 		return fallback, nil
 	}
 
-	boolean, ok := value.(bool)
-	if !ok {
-		return false, fmt.Errorf("%s must be a boolean", key)
+	switch typed := value.(type) {
+	case bool:
+		return typed, nil
+	case string:
+		switch strings.ToLower(strings.TrimSpace(typed)) {
+		case "true", "yes", "on", "1":
+			return true, nil
+		case "false", "no", "off", "0":
+			return false, nil
+		}
+	case float64:
+		if typed == 1 {
+			return true, nil
+		}
+		if typed == 0 {
+			return false, nil
+		}
+	case int:
+		if typed == 1 {
+			return true, nil
+		}
+		if typed == 0 {
+			return false, nil
+		}
 	}
-	return boolean, nil
+	return false, fmt.Errorf("%s must be a boolean", key)
 }
 
 func intArg(args map[string]any, key string, fallback int, min int, max int) (int, error) {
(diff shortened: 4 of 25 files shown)


Vasanthdev2004 (approved):
Approved on `da4c100`.

I reviewed the tools-runtime slice and found no human-review blockers. The scope is coherent: argument aliases/coercion stay type-strict where they should, `ask_user` is registered as a read-only non-blocking fallback, registry-level output redaction is applied across execution paths, mutation targets are workspace-relative, and the grep/apply_patch/write hardening has useful regression coverage.

Validation passed locally:

- `git diff --check origin/main...origin/pr/119`
- `go test ./internal/tools`
- `go test ./...`
- `go vet ./...`

I also attempted `go test -race ./internal/tools`; local Windows env cannot run it because `-race` requires CGO and `gcc` is not installed here. GitHub Actions are green. CodeRabbit is still pending at the moment of this review, but there are no existing inline comments or review blockers visible.

gnanam1990:
@coderabbitai review — addressed all 4: (1) scrubResultSecrets now redacts `Result.Display.Summary` too; (2) MutationTargets runs apply_patch through `validatePatchPaths(applyRoot, ...)` so traversal patches can't return out-of-workspace targets (+ regression test); (3)/(4) the apply_patch tests skip only when the git binary is unavailable (else Fatalf), so real regressions fail. build/vet/-race/full-suite + windows green. (@Vasanthdev2004 already approved.)

Vasanthdev2004 (changes requested):
### Blockers

- `internal/tools/registry.go:92` / `:97` / `:108` / `:114` still return early without passing the `Result` through `scrubResultSecrets`. The display-summary and mutation-target blockers from the previous head are fixed, but this leaves the same boundary gap on denial paths. I reproduced it on `7d94639` by calling `write_file` with an assembled dummy provider-key-shaped filename outside the workspace through `Registry.RunWithOptions` + sandbox enforcement: the sandbox denial returned before `scrubResultSecrets`, so the raw secret-bearing path was still present in `Output` and `Redacted` stayed false. Unknown-tool / prompt / permission-denied exits should also go through the same scrub function so every registry return path has the same redaction contract.

### Non-Blocking

- `intArg` should defensively reject NaN/Infinity/out-of-range float inputs before converting to `int`; current public callers mostly catch the resulting bad values through min/max checks, but the helper should fail closed before implementation-dependent casts.

### Looks Good

- The prior `Display.Summary` redaction issue is fixed: my focused probe now passes and `scrubResultSecrets` scrubs both `Output` and `Display.Summary`.
- The prior `apply_patch` `MutationTargets` traversal issue is fixed: escaping patch paths now return no targets and have regression coverage.
- Latest head validation passed: focused probes for the fixed blockers, `go test -count=1 ./internal/tools`, `go test -count=1 -p 1 ./...`, `go vet ./...`, `go build ./cmd/zero`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, and `git diff --check`.
- Latest-main test merge was clean and passed `go test -count=1 ./internal/tools`, `go test -count=1 -p 1 ./...`, `go vet ./...`, `go build ./cmd/zero`, and `git diff --check`.
- CI is green; local `go test -race ./internal/tools` is unavailable because this Windows environment lacks CGO/gcc.

**Verdict: Changes Requested** — the original blockers are fixed, but registry-level redaction still needs to cover the early denial/error returns before this tool-runtime boundary is safe.

gnanam1990:
@coderabbitai review — addressed both: (1) ask_user `multiSelect` is now best-effort (uncoercible → defaults false, no longer fails the call), like the options path; (2) extended the redaction regression to leak via `Result.Display.Summary` and assert it's scrubbed. build/vet/-race/full-suite + windows green.

gnanam1990:
@Vasanthdev2004 blocker fixed: `RunWithOptions` now routes **every** return through `scrubResultSecrets` via a named-return + single `defer` — so the unknown-tool, sandbox-deny, sandbox-approval, permission-required and permission-denied early returns are redacted too (regression test reproduces your case: a denial reason carrying a token is scrubbed and `Redacted=true`). Also addressed the non-blocking `intArg` note: it now fails closed on NaN/Inf/out-of-range floats (and string forms) before the cast, with tests. build/vet/-race/full-suite + windows green. Thanks for the boundary catch.

anandh8x (changes requested):
I re-reviewed the new head `8ca4390`.

Blockers:

- `internal/agent/loop.go`: nested schema is still dropped before reaching providers. `PropertySchema` now supports nested `Properties` / `Required`, and `ask_user` / `update_plan` depend on it, but `propertyToRuntimeMap` only serializes type/description/enum/default/items/min/max. The model sees object items without field details, so the new nested schemas are not actually advertised.

- `ask_user` is advertised as a core tool, but there is still no agent/TUI intercept path. The agent loop just executes tool calls through the registry, so `ask_user` always returns the non-interactive fallback instead of waiting for answers. Either wire an ask-user callback path now, or keep it out of `CoreReadOnlyTools` until the interactive path lands.

Fixed in the new commit:

- Registry redaction now covers early return paths via the deferred scrub in `RunWithOptions`.
- `intArg` now rejects non-finite / non-integral float inputs before casting.

Validation:

- GitHub smoke checks are green at this head; CodeRabbit was still pending when I checked.
- Focused local tests passed on a PR snapshot: `GOCACHE=/tmp/zero-go-cache go test ./internal/tools -run 'TestRunWithOptions|TestIntArg|TestAskUser|TestUpdatePlan|TestCoreReadOnly'`, `GOCACHE=/tmp/zero-go-cache go test ./internal/agent -run 'Test'`, and `GOCACHE=/tmp/zero-go-cache go test ./internal/zeroline -run 'TestRenderChatAskUser|TestOverlaySuppressedDuringAskUser'`.
- Full `go test ./internal/tools` hit a local bubblewrap NETLINK permission failure, unrelated to this diff.

Vasanthdev2004 (changes requested):
### Blockers

- `internal/agent/loop.go:481` still drops nested `PropertySchema.Properties` / `Required` when serializing tool schemas for providers. `ask_user` now defines `questions.items` as an object with `question`, `options`, and `multiSelect`, and `update_plan` also relies on nested object schemas, but `propertyToRuntimeMap` only forwards type/description/enum/default/items/min/max. I added a focused probe on `8ca4390` that captures the provider request for `ask_user`; the advertised `questions.items` schema was only `map["type":"object"]`, with no nested fields or required list. That means models do not actually see the shape this PR added.

- `ask_user` is now advertised from `CoreReadOnlyTools`, but the agent loop still has no interactive intercept/callback path before `registry.RunWithOptions`. `NewAskUserTool.Run()` explicitly returns the non-interactive fallback (`No interactive user is available...`) and `executeToolCall` just runs the tool and appends that result back to the model. The zeroline rendering support exists, but I could not find an `OnAskUser`/answer callback or an agent/TUI bridge that turns an `ask_user` tool call into a blocking questionnaire. As shipped, the new core tool is advertised as "wait for answers" but always degrades to "proceed with assumptions" in the actual agent path; either wire the interactive path or keep it out of the default registry until that slice lands.

### Non-Blocking

- CodeRabbit/anandh also called out several policy/documentation/test-hardening items around the new default tool surface and mutation/checkpoint trust surface. I do not think those need to block this head once the runtime blockers above are fixed, but they are worth resolving or explicitly acknowledging before merge.

### Looks Good

- The prior blocker from my last review is fixed: `RunWithOptions` now uses a named return plus a single deferred `scrubResultSecrets`, so unknown-tool, sandbox-denied, sandbox-approval, permission-required, permission-denied, and executed-tool paths share the same redaction boundary. I reran the sandbox-denial probe with a dummy provider-key-shaped filename outside the workspace; the raw secret-bearing path was scrubbed and `Redacted=true`.
- The prior `Display.Summary` and `MutationTargets` redaction/traversal issues remain fixed.
- The `intArg` non-blocking note is fixed: non-finite and non-integral float inputs now fail closed before casting.
- Local validation passed on this head: focused redaction/intArg/ask_user tests, `go test -count=1 ./internal/tools`, `go test -count=1 -p 1 ./...`, `go vet ./...`, `go build ./cmd/zero`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, and `git diff --check`.
- Latest-main test merge onto `origin/main` was clean and passed `go test -count=1 ./internal/tools`, `go test -count=1 -p 1 ./...`, `go vet ./...`, `go build ./cmd/zero`, and `git diff --check`.
- GitHub CI smoke checks are green on `8ca4390`; the CodeRabbit status context was still pending when checked. Local `go test -race ./internal/tools` is still unavailable on this Windows host because CGO cannot find `gcc`.

**Verdict: Changes Requested** — the registry redaction blocker is fixed, but the new nested schema and `ask_user` runtime integration are not actually delivered through the agent path yet.

gnanam1990:
Pushed a substantial round (e2f9efe + 9c7120d):

**Reviewer items addressed**
- @coderabbitai grep TOCTOU: `confineGrepFile` now returns the symlink-resolved path and `collectGrepMatches` reads that, not the raw candidate.
- @anandh8x (1) ask_user removed from `CoreReadOnlyTools` — it needs the agent's `OnAskUser` intercept; the tool still ships, the agent module registers+wires it together. (2) `update_plan` schema reverted to flat (item structure in the description) until the agent's `propertyToRuntimeMap` serializes nested `Properties/Required`. Reverted the specialist vocab + core-registration test accordingly.

**Plus 3 found by a proactive adversarial self-review (and fixed):**
- 🟠 **sandbox alias-gate bypass** — the widened `file_path`/`filename`/`filepath` path aliases weren't in `sandbox.requestPaths` (only `path/cwd/file/dir`), so an aliased out-of-workspace target skipped the sandbox workspace+symlink check (same class as audit H4). Aligned `requestPaths` with the tools' alias lists. Regression test added.
- 🟡 `scrubResultSecrets` now also scrubs `Result.Meta` (forwarded to the transcript; carries glob pattern / bash cwd).
- 🟡 `floatToInt` upper bound `>` → `>=` (float64(MaxInt) rounds up to 2^63, which slipped through to the cast).

build/vet/-race/full-suite + `GOOS=windows build` green; no new deps. @coderabbitai review

Vasanthdev2004 (approved):
### Blockers

- None found on `9c7120d`.

### Non-Blocking

- The remaining schema limitation is explicitly deferred: `propertyToRuntimeMap` still does not serialize nested `PropertySchema.Properties` / `Required`, but this head no longer relies on that path for core-advertised tools. `ask_user` has been removed from `CoreReadOnlyTools`, and `update_plan` now keeps its provider schema flat with the item shape described textually. That is acceptable for this slice, but the nested serializer still needs to be fixed before re-advertising `ask_user` or other object-item schemas.

### Looks Good

- The prior `ask_user` runtime blocker is fixed for this slice: `ask_user` is no longer registered in the default core read-only toolset, and specialist manifests no longer allow it in `knownToolNames`, so the agent will not advertise a “wait for answers” tool until an interactive bridge exists.
- The prior registry redaction blocker remains fixed: all `RunWithOptions` return paths go through the deferred `scrubResultSecrets`; my sandbox-denial probe with a dummy provider-key-shaped path returned scrubbed output with `Redacted=true`.
- `Output`, `Display.Summary`, and `Meta` values are now covered by boundary scrubbing, with regression coverage.
- The `intArg` hardening now rejects NaN/Inf/non-integral/out-of-range float inputs before casting, including the `2^63` rounding edge case.
- Grep now reads the symlink-resolved path validated at read time, closing the TOCTOU path noted by another reviewer.
- Local validation passed: focused blocker probes and targeted tests, `go test -count=1 ./internal/tools ./internal/sandbox ./internal/specialist`, `go test -count=1 -p 1 ./...`, `go vet ./...`, `go build ./cmd/zero`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, and `git diff --check`.
- Latest-main test merge onto `origin/main` was clean and passed targeted tests, full tests, vet, build, and diff-check.
- GitHub CI is green on `9c7120d`. Local `go test -race ./internal/tools` is still unavailable on this Windows host because CGO cannot find `gcc`.

**Verdict: Approved** — the blockers I raised on `8ca4390` are fixed or safely deferred out of the default runtime surface in `9c7120d`.

anandh8x (approved):
Re-reviewed latest head 9c7120d. The blockers from my previous review are addressed: ask_user is no longer exposed in CoreReadOnlyTools/specialist known tools until the interactive bridge exists, and update_plan no longer depends on nested provider schema serialization. The extra hardening also looks good: registry boundary scrubs Meta, sandbox risk checks cover path aliases, grep re-confines and reads the resolved path, and intArg rejects the 2^63 float edge. Focused local validation passed on the PR snapshot: GOCACHE=/tmp/zero-go-cache go test ./internal/tools -run 'TestRegistry|TestRunWithOptions|TestAskUser|TestUpdatePlan|TestIntArg|TestGrep|TestCoreReadOnlyTools|TestSecret'; GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/specialist ./internal/agent; git diff --check main...origin/tools-runtime. No blockers from me.

PR #120, opened 2026-06-07
Add specialist lifecycle accounting

anandh8x:
Summary:
- records specialist_start and specialist_stop session events for foreground, resume, and background specialist runs
- rolls child stream-json usage events into a single parent provider_usage event
- makes background TaskOutput roll-up idempotent so persisted completed tasks are safe to read repeatedly

Self-review:
- this intentionally uses session events/accounting, not shell hook execution; the hooks package currently only defines configuration/audit events
- no user-facing CLI surface changes are added
- .omx/ remains untracked and is not part of this PR

Tests:
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Adds specialist session start/stop events and aggregates usage (token/event rollups) per run.
  * Background task accounting now records completion status, rollups once, and cleans up resources.
  * Output handling now summarizes streams with run IDs and usage summaries.

* **Bug Fixes**
  * More accurate stop status and exit-code handling for child tasks; prevents duplicate rollups.

* **Tests**
  * New tests for lifecycle, usage rollup, mixed usage formats, and background output handling.

### DIFF
--- a/internal/npmwrapper/npmwrapper_test.go
+++ b/internal/npmwrapper/npmwrapper_test.go
@@ -67,9 +67,9 @@ func TestNodeWrapperReportsMissingNativeBinary(t *testing.T) {
 	node := requireNode(t)
 	wrapperPath := copyWrapperFixture(t)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), nodeWrapperTimeout())
 	defer cancel()
-	command := exec.CommandContext(ctx, node, wrapperPath, "--version")
+	command := nodeWrapperCommand(ctx, node, wrapperPath, "--version")
 	output, err := command.CombinedOutput()
 	if ctx.Err() != nil {
 		t.Fatalf("wrapper timed out reporting missing native binary: %v; output: %s", ctx.Err(), output)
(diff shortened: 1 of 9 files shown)


gnanam1990 (approved):
## Review — clean, idempotent accounting, approving

Nicely scoped lifecycle/usage accounting:
- **Idempotent by construction** — `specialistEventExists` (matching source=specialist + child/taskId + runId) guards both the `specialist_stop` and the usage rollup, so re-reading a persisted completed background task via `TaskOutput` never double-records. `TestOutputToolRollsUpCompletedBackgroundUsageOnce` locks that in.
- **Stop is recorded on every path** — success and error, foreground/resume/background — and the `run.Started || ExitCode != 0` guard avoids accounting a child that never launched. `specialistStopStatus` derives a sensible status from status/exitCode/err.
- **Single usage rollup** per child run (`provider_usage` with prompt/completion/effective-total + event count), only when there's usage and a valid parent/child.
- Graceful on a nil/invalid store (falls back) and skips when the parent session id is missing/invalid. Good coverage (foreground lifecycle+rollup, started-child error exit code, background once).

`go vet`, `go test ./...`, `-race`, and `GOOS=windows build` all green. **Approving.** 👍

### Non-blocking note
The idempotency is **check-then-append** (`specialistEventExists` reads, then `AppendEvent`), not atomic under one lock. Two concurrent `TaskOutput` reads of the same completed task could both pass the existence check and double-record. In practice tool calls are sequential within a process so this is very unlikely, but if you ever drive `TaskOutput` concurrently, doing the existence-check + append under the session lock would make it airtight.

<sub>🤖 Verified review via Claude Code (build/vet/-race/windows/full-suite run locally).</sub>

PR #121, opened 2026-06-07
Add guarded web fetch tool

Vasanthdev2004:
## Summary
- Add first-party `web_fetch` as a guarded core network tool.
- Enforce prompt safety metadata, bounded response reads, URL/body redaction, redirect validation, and private/loopback/link-local host blocking.
- Wire `web_fetch` into core tools while keeping it out of read-only MCP defaults.

## Tests
- `go test ./...`
- `go run ./cmd/zero exec --list-tools --enabled-tools web_fetch`

## Summary by CodeRabbit

* **New Features**
  * Added a web_fetch tool for HTTP/HTTPS retrieval with size limits, timeouts, redaction, and robust network-safety checks.

* **CLI / Configuration**
  * Tool exposure now follows allowlist/unsafe-tools settings; web_fetch is hidden by default and becomes available when unsafe tools are enabled.
  * Manifests and core tool registry recognize web_fetch as a known network tool.

* **Tests**
  * Extensive tests added for schema, safety gating, permission/advertisement behavior, redirects, truncation, and policy enforcement.

### DIFF
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -162,6 +162,32 @@ func TestRunAdvertisesRuntimeToolDefinitions(t *testing.T) {
 	}
 }
 
+func TestRunAdvertisesWebFetchInAutoMode(t *testing.T) {
+	registry := tools.NewRegistry()
+	registry.Register(tools.NewWebFetchTool())
+	provider := &mockProvider{
+		turns: [][]zeroruntime.StreamEvent{{
+			{Type: zeroruntime.StreamEventText, Content: "done"},
+			{Type: zeroruntime.StreamEventDone},
+		}},
+	}
+
+	_, err := Run(context.Background(), "what tools exist?", provider, Options{
+		Registry:       registry,
+		PermissionMode: PermissionModeAuto,
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(provider.requests) != 1 {
+		t.Fatalf("expected one request, got %d", len(provider.requests))
+	}
+	if len(provider.requests[0].Tools) != 1 || provider.requests[0].Tools[0].Name != "web_fetch" {
+		t.Fatalf("expected web_fetch to be advertised in auto mode, got %#v", provider.requests[0].Tools)
+	}
+}
+
 func TestRunFiltersAdvertisedTools(t *testing.T) {
 	root := t.TempDir()
 	registry := tools.NewRegistry()
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -631,6 +631,7 @@ func assertCoreRegistry(t *testing.T, registry *tools.Registry) {
 		"apply_patch",
 		"update_plan",
 		"bash",
+		"web_fetch",
 	} {
 		if _, ok := registry.Get(name); !ok {
 			t.Fatalf("expected registry to include core tool %q", name)
--- a/internal/cli/serve_test.go
+++ b/internal/cli/serve_test.go
@@ -31,7 +31,7 @@ func TestRunServeMCPListsReadOnlyToolsByDefault(t *testing.T) {
 			t.Fatalf("expected MCP output to contain %q, got %q", want, output)
 		}
 	}
-	for _, unwanted := range []string{"write_file", "apply_patch", "bash"} {
+	for _, unwanted := range []string{"write_file", "apply_patch", "bash", "web_fetch"} {
 		if strings.Contains(output, unwanted) {
 			t.Fatalf("did not expect default MCP output to contain %q: %q", unwanted, output)
 		}
@@ -53,7 +53,7 @@ func TestRunServeMCPAllowsUnsafeToolsWithExplicitFlag(t *testing.T) {
 		t.Fatalf("expected exit code 0, got %d; stderr=%q", exitCode, stderr.String())
 	}
 	output := stdout.String()
-	for _, want := range []string{"read_file", "write_file", "apply_patch", "bash"} {
+	for _, want := range []string{"read_file", "write_file", "apply_patch", "bash", "web_fetch"} {
 		if !strings.Contains(output, want) {
 			t.Fatalf("expected unsafe MCP output to contain %q, got %q", want, output)
 		}
--- a/internal/sandbox/engine_test.go
+++ b/internal/sandbox/engine_test.go
@@ -208,6 +208,29 @@ func TestEngineClassifiesNetworkAndDestructiveShellCommands(t *testing.T) {
 	}
 }
 
+func TestEngineDeniesNetworkSideEffectWhenPolicyBlocksNetwork(t *testing.T) {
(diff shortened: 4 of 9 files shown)


gnanam1990 (changes requested):
## Security review — strong design, but an adversarial SSRF pass found real blocklist bypasses (requesting changes)

First, credit where due: the **DNS-rebind defense is correct and well-built** — the custom `DialContext` resolves at dial time, validates *every* resolved IP, and **pins the connection to the validated IP** (dials the IP, not the hostname), redirects are re-validated through both `CheckRedirect` and the same dialer, `Proxy`/`DialTLSContext` are neutralized, IPv4-mapped IPv6 is `Unmap()`ed, reads are bounded, and the default sandbox policy `NetworkDeny`s this tool unless network is explicitly enabled. That's the hard part done right.

But `web_fetch` *is* the SSRF surface, so I ran an adversarial pass (5 attack classes, each finding independently re-traced through the code). It confirmed **address-blocklist bypasses that reach loopback/metadata/internal hosts** once network is enabled — the `rejectUnsafeWebFetchAddr` allowlist is incomplete. These were verified against the actual code, not speculation.

### 🟠 Major — IPv4-embedded-IPv6 (NAT64 / 6to4) bypass
`http://[64:ff9b::a9fe:a9fe]/` (NAT64 WKP embedding **169.254.169.254**, the cloud metadata IP), `http://[64:ff9b::7f00:1]/` (→127.0.0.1), `http://[2002:7f00:1::]/` (6to4 →127.0.0.1). These are valid global-unicast IPv6 literals, so `netip.ParseAddr` succeeds and `rejectUnsafeWebFetchAddr` checks the *raw* address (none of `IsLoopback/IsPrivate/IsLinkLocal*` fire) → **ALLOW**, and the dial pins/dials the literal verbatim. On any NAT64-gateway'd network (common in IPv6-only cloud/CI; `64:ff9b::/96` is the routed RFC6052 WKP) or 6to4 relay, the connection physically reaches the embedded IPv4. The code already decodes IPv4-*mapped* (`Unmap`) but not these equivalent encodings.
**Fix:** before the `IsLoopback/...` checks, detect `64:ff9b::/96`, `64:ff9b:1::/48`, `2002::/16` (and `::/96` IPv4-compatible), extract the embedded IPv4, and recurse `rejectUnsafeWebFetchAddr` on it.

### 🟠 Major — reserved / special-use IPv4 ranges not blocked
Plain dotted-quads that `IsPrivate()` doesn't cover, all confirmed reachable: **`100.64.0.0/10` (RFC6598 CGNAT** — cloud internal fabrics, Tailscale, overlays), **`0.0.0.0/8`** beyond the exact `0.0.0.0` (e.g. `0.1.2.3`, can reach localhost on some OSes), `240.0.0.0/4`, `255.255.255.255`, `192.0.0.0/24`, `198.18.0.0/15`. Same gap applies on **redirects** (a public URL that 30x→`100.64.x`) since the redirect path re-runs the same incomplete check.
**Fix (covers this + the above cleanly):** flip `rejectUnsafeWebFetchAddr` to an **allowlist posture** — reject unless `addr.IsGlobalUnicast()` AND not contained in a curated set of IANA special-purpose prefixes (loopback, the RFC1918 set, `100.64.0.0/10`, `0.0.0.0/8`, link-local, `240.0.0.0/4`, `192.0.0.0/24`, `198.18.0.0/15`, broadcast, IPv6 ULA/`64:ff9b::/96`/`2002::/16`, etc.), evaluated after IPv4-embed decoding. An allowlist is far more robust than chasing predicates.

### 🟠 Major — no port restriction
`web_fetch` will dial arbitrary TCP ports (`:22`, `:25`, `:6379`, …) on any resolved host. Combined with HTTP being sent to non-HTTP services, this is the classic protocol-smuggling lever. Recommend restricting to the default http/https ports (80/443) — or an explicit small allowlist — rather than any port.

### 🟡 Minor — `Content-Type` emitted unredacted
The response body and URL/error text go through `redactWebFetchText`, but the `Content-Type` header is copied verbatim into both `Output` and `Meta`. Low-risk (rarely secret-bearing) but it's a hole in the otherwise-complete redaction boundary — redact it too.

### Suggestion
Add regression tests for each: `64:ff9b::7f00:1`, `2002:7f00:1::`, `100.64.0.1`, `0.1.2.3`, `255.255.255.255`, a redirect→`100.64.x`, and a non-80/443 port. The existing suite covers loopback/private-10/link-local/metadata/rebind/redirect well — these are the gaps.

Requesting changes on the blocklist completeness (the NAT64/6to4 + CGNAT reaches are the ones I'd most want closed before this can fetch on an internal network). The architecture is sound; it's the address-classification allowlist that needs tightening.

<sub>🤖 Adversarial SSRF review via Claude Code (16 agents, 5 attack classes; each finding re-traced through internal/tools/web_fetch.go).</sub>

gnanam1990 (approved):
## Re-review — all SSRF findings resolved ✅ Approving

`b809637` addresses every item from my review, and cleanly — you flipped `rejectUnsafeWebFetchAddr` to the allowlist posture I'd hoped for. Verified each against the new code:

- ✅ **NAT64/6to4 IPv4-embedded-IPv6** → `webFetchEmbeddedIPv4` decodes `::/96`, `64:ff9b::/96`, `64:ff9b:1::/48`, `2002::/16` and **recurses** on the inner IPv4, so `[64:ff9b::a9fe:a9fe]` (→metadata), `[2002:7f00:1::]` (→127.0.0.1), `[::7f00:1]` are all blocked. Tests cover exactly these.
- ✅ **Reserved/special-use ranges** → `webFetchBlockedAddrPrefixes` is now comprehensive (CGNAT `100.64/10`, `0.0.0.0/8`, `240/4`, broadcast, TEST-NETs, benchmark, `192.88.99/24`, plus the IPv6 special-use set) **and** a final `!IsGlobalUnicast()` fail-closed. `100.64.0.1` / `0.1.2.3` / `255.255.255.255` now rejected (with tests).
- ✅ **Port restriction** → `validateWebFetchPort` allows only the scheme's default (80/http, 443/https); `:22`/`:6379`/mismatched `:80` rejected.
- ✅ **Content-Type redaction** → `Content-Type` now goes through `redactWebFetchText` before Output/Meta.

The dial-time pin + redirect re-validation route through the same hardened check, so all of this holds on redirects too. Whole tree builds, `go test ./...` and `-race` green, CI green.

### Non-blocking (for a future tidy, not this PR)
`webFetchEmbeddedIPv4` reads 4 **contiguous** bytes at `byteOffset`. For the WKP `64:ff9b::/96` (off 12) that's exactly right. For `64:ff9b:1::/48` (off 6), strict RFC 6052 actually reserves the "u" octet at byte 8 and places IPv4 across bytes 6-7 + 9-10 — so a strict /48-encoded address is read off-by-one in octets 3-4. In practice it's **not exploitable**: octets 1-2 are read correctly and the blocklist is dominated by /8–/16 prefixes (+ the `IsGlobalUnicast` backstop), so realistic private targets still land in a blocked range. If you ever want strict correctness, skip byte 8 for the `/40`–`/64` NAT64 forms.

Nice turnaround — this went from genuinely bypassable to robustly defended. 👍

PR #124, opened 2026-06-07
MCP hardening: nested-schema passthrough + hung-server timeout

gnanam1990:
## Module — MCP hardening

Independent slice of the runtime-core split (off `main`). `internal/mcp` had **zero drift** on main, so this is a clean extract of the zenline MCP improvements. Deps (`config`, `tools`) are on main; uses `tools.PropertySchema.Properties/Required` from #119. No new deps.

### What's in it
- **Nested-schema passthrough** (`schema.go`) — when converting `tools.PropertySchema` ⇄ MCP JSON schema, serialize nested object `Properties`/`Required` (and object-typed `Items`) instead of flattening them, so tools with nested args are advertised faithfully to/from MCP hosts. *(audit H6.)*
- **Hung-server timeout** (`client.go`/`server.go`) — bound the MCP stdio handshake/calls so a hung or non-responsive MCP server can't block the caller forever; surface a timeout instead. *(audit H9; `hang_test.go` covers it.)*
- `registry.go` recognizes the hardened paths; `registry_test.go`/`schema_test.go` added.

### Testing
`go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./internal/mcp/`, `GOOS=windows go build ./...` all green.

> Companion to #123 (skills & plugins). Together these are the "extensibility" layer. The TUI shell + cli wiring remain blocked on the agent module.

Part of decomposing #101 (draft).

## Summary by CodeRabbit

* **New Features**
  * Added support for complex nested parameter schemas, including arrays and objects.

* **Bug Fixes**
  * Fixed connection hangs when MCP servers don't respond to initialization requests.
  * Fixed server shutdown hangs when peer connections become unresponsive.
  * Made tool registration atomic—all tools register successfully or none do if any validation fails.

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -46,9 +46,30 @@ type Client struct {
 	mu      sync.Mutex
 	closeMu sync.Mutex
 	nextID  int
+
+	// dispatchMu guards the response-dispatch state shared with the single
+	// reader goroutine. It is never held across a blocking read.
+	dispatchMu sync.Mutex
+	readerOnce sync.Once
+	pending    map[int]chan dispatchResult
+	readErr    error
+	readDone   bool
+}
+
+// dispatchResult carries one matched JSON-RPC response (or a terminal reader
+// error) to a waiting caller.
+type dispatchResult struct {
+	message rpcMessage
+	err     error
 }
 
 const stdioCloseWaitTimeout = 500 * time.Millisecond
+
+const (
+	// initializeTimeout bounds the MCP handshake so a non-responsive peer fails
+	// fast instead of hanging startup.
+	initializeTimeout = 30 * time.Second
+)
 
 func Connect(ctx context.Context, server Server) (ToolClient, error) {
 	switch server.Type {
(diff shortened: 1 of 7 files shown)


internal/mcp/client.go
```diff
@@ -183,43 +214,156 @@ func (client *Client) request(ctx context.Context, method string, params any, ta
...
+
+// rpcMessageID extracts the integer id from a JSON-RPC id value across the
+// numeric/string encodings a server may use.
+func rpcMessageID(value any) (int, bool) {
+	switch typed := value.(type) {
+	case int:
+		return typed, true
+	case int64:
+		return int(typed), true
+	case float64:
```
Vasanthdev2004, line 351: Small hardening follow-up: this accepts any `float64` id by truncating it, so a malformed response id like `1.9` would match pending request `1`. JSON-RPC numeric ids should be integral for our dispatcher; consider rejecting non-integral or out-of-range float values here, similar to the integer parsing used elsewhere.

Vasanthdev2004 (approved):
Verdict: approved.

The MCP timeout/dispatch changes look good and the CI smoke suite is green. I left one small hardening note around fractional JSON-RPC ids, but I do not think it should block this PR.

PR #125, opened 2026-06-07
Agent runtime: ask_user interception, context compaction, guardrails

gnanam1990:
## Agent runtime — ask_user, compaction, guardrails

Ports the agent-runtime slices from the reskin branch onto `main`, **minus the blueprint's own sub-agent/task interception** — `main` already has the merged **specialist** Task implementation (anandh8x's), which this PR preserves rather than duplicating.

Built as a selective 3-way merge: started from main's `loop.go`/`types.go`, brought in the agent features, kept main's specialist drift, and excised the blueprint `task` path (`executeTask`, `taskDisplaySummary`, `maxTaskDepth`, `Options.Provider`, the `task` interception, and the one task usage-propagation test).

### What's in it
- **ask_user (S6)** — `AskUserRequest`/`AskUserResponse`/`AskUserQuestion` + `Options.OnAskUser`. The loop intercepts `ask_user` and routes to the interactive front-end, degrading gracefully to the non-interactive fallback when no handler is set. Answers go through the redaction boundary.
- **Compaction (S8)** — `Options.ContextWindow` + `CompactionPreserveLast`. Proactive compaction when estimated history crosses a fraction of the window; reactive recovery on context-limit errors (summarizes the old middle, keeps system + last N). **`ContextWindow=0` disables it**, so every existing caller/test is unchanged.
- **Guardrails** — empty-turn stop, repeated-tool-failure hint/stop, stale plan reminder, aborted-placeholder on guard stop; `confirmation_policy.md` embedded into the system prompt.

### Preserved from main (specialist)
`RunWithOptions` tool-context fields (`ToolCallID`/`SessionID`/`Model`/`ReasoningEffort`/`Depth`/`Cwd`), `propertyToRuntimeMap` nested `Items`, `ToolAdvertised` `AdvertiseInAuto`. Dropped `Options.Provider` (blueprint-task-only; compaction uses `Run`'s provider arg).

### Testing
54 agent tests pass (incl. the new compaction/guardrails/ask_user Run-level tests). `go build ./...`, `go vet`, `go test ./...`, `go test -race ./internal/agent/`, `GOOS=windows go build ./...` all green. No new deps.

> Unblocks the **TUI shell** and **cli wiring** PRs. Part of decomposing #101.

## Summary by CodeRabbit

* **New Features**
  * Interactive user-question prompts during runs; agent can request and accept structured answers.
  * Automatic conversation compaction to keep long sessions concise.

* **Improvements**
  * Stronger guardrails to prevent runaway behavior and repeated-failure loops.
  * Secrets in tool/handler output are redacted; malformed tool calls receive clearer handling and retries.
  * Confirmation policy is embedded into system prompts.

* **Documentation**
  * Added confirmation policy guidance.

### DIFF
--- a/internal/agent/compaction_test.go
+++ b/internal/agent/compaction_test.go
@@ -409,9 +409,9 @@ func TestRunReactiveRetryDoesNotDoubleEmitText(t *testing.T) {
 	if provider.summarizeCalls == 0 {
 		t.Fatal("expected reactive compaction to run")
 	}
-	// The retried turn's text must be streamed to OnText at most once. Before the
-	// fix it was emitted on both the original (mid-stream) collect AND the retry
-	// collect, double-emitting the retried response.
+	// The retried turn's text must NOT be re-streamed to OnText. Before the fix it
+	// was emitted on both the original (mid-stream) collect AND the retry collect,
+	// double-emitting the retried response.
 	joined := strings.Join(deltas, "")
 	if got := strings.Count(joined, "recovered"); got != 0 {
 		t.Fatalf("retried-turn text must NOT be re-streamed to OnText, saw %d occurrences in %q", got, joined)
--- a/internal/agent/compaction_test.go
+++ b/internal/agent/compaction_test.go
@@ -1,0 +1,542 @@
+package agent
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// --- Pure Compact() tests -------------------------------------------------
+
+func TestCompactKeepsSystemAndPreservedSuffix(t *testing.T) {
+	messages := []zeroruntime.Message{
+		{Role: zeroruntime.MessageRoleSystem, Content: "system prompt"},
+		{Role: zeroruntime.MessageRoleUser, Content: "first question"},
+		{Role: zeroruntime.MessageRoleAssistant, Content: "first answer"},
+		{Role: zeroruntime.MessageRoleUser, Content: "second question"},
+		{Role: zeroruntime.MessageRoleAssistant, Content: "second answer"},
+		{Role: zeroruntime.MessageRoleUser, Content: "most recent question"},
+	}
+
+	var captured []zeroruntime.Message
+	summarizeCalls := 0
+	result, err := Compact(messages, CompactionOptions{
+		PreserveLast: 2,
+		Summarize: func(toSummarize []zeroruntime.Message) (string, error) {
+			summarizeCalls++
+			captured = toSummarize
+			return "DENSE SUMMARY", nil
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if summarizeCalls != 1 {
+		t.Fatalf("expected Summarize to be called once, got %d", summarizeCalls)
+	}
+
+	// Head is the original system message.
+	if result[0].Role != zeroruntime.MessageRoleSystem || result[0].Content != "system prompt" {
+		t.Fatalf("expected system message preserved at head, got %#v", result[0])
+	}
+	// Next is the injected summary as a single user message.
+	if result[1].Role != zeroruntime.MessageRoleUser {
+		t.Fatalf("expected summary injected as user role, got %#v", result[1])
+	}
+	if !strings.Contains(result[1].Content, "[Summary of earlier conversation]") {
+		t.Fatalf("expected summary label, got %q", result[1].Content)
+	}
+	if !strings.Contains(result[1].Content, "DENSE SUMMARY") {
+		t.Fatalf("expected summary body, got %q", result[1].Content)
+	}
+	// Tail is the preserved suffix, verbatim.
+	last := result[len(result)-1]
+	if last.Content != "most recent question" {
+		t.Fatalf("expected most recent message preserved, got %q", last.Content)
+	}
+	// The summarized middle excludes system and preserved suffix.
+	if len(captured) != 3 {
+		t.Fatalf("expected 3 summarized messages, got %d: %#v", len(captured), captured)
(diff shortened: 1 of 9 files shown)


internal/agent/loop.go
```diff
@@ -183,13 +371,107 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
...
+
+	header, _ := args["header"].(string)
+	request := AskUserRequest{
+		ToolCallID: call.ID,
+		Header:     strings.TrimSpace(header),
+		Questions:  toAgentAskUserQuestions(questions),
+	}
+	response, err := options.OnAskUser(ctx, request)
+	if err != nil {
+		return askUserFallbackResult(ctx, registry, call, args)
```
Vasanthdev2004: This path treats every `OnAskUser` error as a headless fallback. If the handler returns `context.Canceled` or `context.DeadlineExceeded`, the run should stop and return that error instead of appending a synthetic no-interactive-user answer and continuing. Otherwise user cancellation/timeouts still mutate the transcript and can keep the agent running after the UI asked it to abort.

internal/agent/loop.go
```diff
@@ -183,13 +371,107 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
...
+		return ToolResult{
+			ToolCallID: call.ID,
+			Name:       call.Name,
+			Status:     tools.StatusError,
+			Output:     "Error: Invalid arguments for ask_user: " + err.Error(),
+		}
+	}
+
+	if options.OnAskUser == nil {
+		return askUserFallbackResult(ctx, registry, call, args)
```
Vasanthdev2004: The fallback path should preserve the same execution contract as normal tools. `askUserFallbackResult` currently calls `registry.Run`, which drops the `RunWithOptions` metadata (`ToolCallID`, session/model/depth/cwd fields, sandbox/permission context) and also omits returned `Meta`, `ChangedFiles`, and `Display` when converting back to `ToolResult`. Please pass through the original options via `registry.RunWithOptions` and copy the full result fields, so headless `ask_user` behaves like every other tool path.

internal/agent/loop.go
```diff
@@ -89,6 +211,51 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				Content:    toolResult.Output,
 				ToolCallID: toolResult.ToolCallID,
 			})
+
+			// Repeated-failure guard: if a tool keeps failing the same way, hint
+			// once (with its schema) then halt — so no model loops on a bad call.
+			outcome := guards.observeToolResult(call.Name, toolResult.Status == tools.StatusError, toolResult.Output)
```
Vasanthdev2004: This repeated-failure guard sees every `tools.StatusError` as retriable. That includes disabled tools, permission denials, and sandbox violations, where a schema hint is misleading and can even encourage the model to keep trying blocked behavior. Please classify the `ToolResult` error kind and only inject schema/retry hints for argument parsing or tool execution failures that are actually retriable.

internal/agent/compaction_test.go
```diff
@@ -0,0 +1,542 @@
...
+	if err != nil {
+		t.Fatalf("expected reactive compaction to recover, got error: %v", err)
+	}
+	if result.FinalAnswer != "recovered" {
+		t.Fatalf("expected recovered answer, got %q", result.FinalAnswer)
+	}
+	if provider.summarizeCalls == 0 {
+		t.Fatal("expected reactive compaction to run")
+	}
+	// The retried turn's text must be streamed to OnText at most once. Before the
```
Vasanthdev2004: Tiny cleanup: the comment says the retried text may be streamed to `OnText` "at most once", but the assertion below requires zero occurrences. Please update the comment to say the retried turn must not be streamed to `OnText`.

Vasanthdev2004 (changes requested):
Verdict: request changes.

CI is green, but the agent-loop behavior issues below should be fixed before merge. They affect cancellation semantics, the RunWithOptions contract for ask_user fallback, and repeated-failure hinting for non-retriable errors.

gnanam1990:
Addressed all four (thanks both):
1. **Repeated-failure hint scoped to retriable errors** — new `isRetriableToolError` (keys off the `permission_action` meta + the known disabled/permission/sandbox messages); policy refusals no longer drive the schema hint or the repeated-failure stop. Test `TestIsRetriableToolError`.
2. **Headless `ask_user` fallback preserves the tool contract** — now via `registry.RunWithOptions` with the full run context, copying `Meta`/`ChangedFiles`/`Display`.
3. **Canceled `ask_user` aborts** — `context.Canceled`/`DeadlineExceeded` from `OnAskUser` now returns that error from `Run` (executeToolCall/executeAskUser thread an abort error; remaining advertised calls are closed out for replay validity) instead of a synthetic answer. Test `TestRunAskUserCancellationAbortsRun`.
4. **compaction_test** comment aligned with the assertion.

build/vet/-race/full-suite + windows green. @coderabbitai review

gnanam1990:
@Vasanthdev2004 re-review please — all four items are fixed on the current head (`ba6cf26`): retriable-only failure hint (`isRetriableToolError`), headless ask_user via `RunWithOptions` with full fields, canceled ask_user now aborts the run with the error (not a synthetic answer), and the compaction_test comment. Tests added (`TestIsRetriableToolError`, `TestRunAskUserCancellationAbortsRun`); build/vet/-race/full-suite + windows green.

Vasanthdev2004 (approved):
Re-reviewed current head `ba6cf26`.

Previous blockers are addressed:
- ask_user cancellation now aborts instead of fabricating a fallback answer.
- headless ask_user fallback now goes through RunWithOptions and preserves result fields.
- repeated-failure schema hints are limited to retriable tool failures.
- compaction test comment mismatch is fixed.

Validation passed locally:
- `go test ./...`
- `git diff --check origin/main...origin/pr-125`

CI/CodeRabbit are green. Approved.

PR #126, opened 2026-06-07
TUI shell: ask_user modal, pickers, autocomplete, zeroline skin

gnanam1990:
## TUI shell — ask_user modal, pickers, autocomplete, zeroline skin

Ports the interactive TUI shell from the reskin branch. **Stacked on #125 (agent runtime)** — it needs `agent.AskUserRequest`/`AskUserResponse`/`AskUserQuestion` + `Options.OnAskUser`/`ContextWindow` from that PR, so this targets the `agent-runtime` branch; I'll retarget it to `main` once #125 lands.

Built as a per-file 3-way merge of `internal/tui` (preserving main's #113 session-wiring drift — `SessionID`/`Model`/`ReasoningEffort`/`Cwd`), then a full **zenline→zeroline rename** (imports, identifiers, the `Skin` string, and `zenline_view.go`→`zeroline_view.go`) since the rendering package merged earlier as `internal/zeroline`.

### What's in it
- **ask_user** — interactive questionnaire modal wired to `agent.OnAskUser` via an answer channel; transcript rows + session payloads; degrades gracefully with no interactive surface.
- **Compaction sizing** — `modelContextWindow()` sets the agent's `ContextWindow` from the active model's registry entry; unknown/custom models → 0 (compaction off).
- **Pickers** (model / theme / effort / mode) + slash-command **autocomplete** overlay.
- **zeroline skin** rendering (boot / home / chat) behind `Options.Skin`.
- **Checkpoint recording** — before each mutating tool, the run batches an `EventSessionCheckpoint` *in order* and flushes it at end-of-run and on cancel.

### One cross-module change (called out)
The checkpoint flow needs the event batched **in order** with the run's other session events (recording it out-of-band reorders it, and breaks multi-call ordering). That requires the snapshot-only entry point, so I exposed `sessions.SnapshotForCheckpoint` **with an explicit orphan-blob safe-usage contract**: most callers should use the atomic `CaptureToolCheckpoint`; this one is for the batch-in-order pattern where the caller (the TUI) guarantees prompt recording incl. on cancel.

### Testing
115 tui tests pass. `go build ./...`, `go vet`, `go test ./...`, `go test -race ./internal/tui/ ./internal/sessions/`, `GOOS=windows go build ./...` all green. No new deps.

> Unblocks the **cli wiring** PR. Part of decomposing #101.

## Summary by CodeRabbit

* **New Features**
  * Slash-command autocomplete with keyboard navigation, alias matching, and completion; suppressed while modals/questionnaires are active
  * Expanded interactive pickers (model, effort, mode, theme) and questionnaire (ask-user) flows
  * `/mode` and `/rewind` commands enhanced; `/rewind` reports restored/deleted/skipped files
  * New "zeroline" UI skin with theme/dark-mode controls and tailored rendering

* **Bug Fixes**
  * Session checkpoint events and related state are reliably flushed/persisted when runs are cancelled; in-memory session state refreshes after rewind

### DIFF
--- a/internal/tui/session_controls.go
+++ b/internal/tui/session_controls.go
@@ -2,9 +2,11 @@ package tui
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/Gitlawb/zero/internal/modelregistry"
+	"github.com/Gitlawb/zero/internal/sessions"
 	"github.com/Gitlawb/zero/internal/usage"
 	"github.com/Gitlawb/zero/internal/zeroruntime"
 )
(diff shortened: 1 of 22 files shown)


internal/tui/session_controls.go
```diff
@@ -164,6 +166,58 @@ func (m model) handleCompactCommand(args string) (model, string) {
...
+	report, err := m.sessionStore.ApplyRewind(m.activeSession.SessionID, m.cwd, target)
+	if err != nil {
+		return m, "Rewind\n" + err.Error()
+	}
+	summary := fmt.Sprintf("Rewound to sequence %d\n%d file(s) restored, %d deleted, %d skipped.",
+		target, report.FilesRestored, report.FilesDeleted, len(report.Skipped))
+	if len(report.Skipped) > 0 {
+		summary += "\nskipped (not recoverable): " + strings.Join(report.Skipped, ", ")
+	}
+	return m, summary
```
Vasanthdev2004, line 210: [P1] Refresh TUI session state after applying rewind. `ApplyRewind` restores files, truncates the persisted event log, and appends the rewind marker, but this handler returns the old `m` unchanged. That leaves `m.sessionEvents`, `m.activeSession.EventCount`, and the visible transcript containing the events that were just dropped. The next prompt goes through `sessionPrompt` and sends `m.sessionEvents` back to the agent, so a user can `/rewind latest` and still have the rewound-away tool calls/messages included in the next turn. Please reload the session metadata/events from the store and rebuild or trim the transcript after a successful rewind, and add a TUI regression test for submitting after rewind.

Vasanthdev2004 (changes requested):
Thanks for the TUI shell work. I ran `go test ./...`, `go run ./cmd/zero-release build`, and `git diff --check` at `952d161`; all passed locally. I found one blocker in the newly wired `/rewind` path, so I am requesting changes before this stacks cleanly.

The PR is also stacked on `agent-runtime`, so it should be merged after that base branch is accepted/merged.

gnanam1990:
@Vasanthdev2004 fixed the P1 — after `ApplyRewind` the handler now reloads session metadata (`store.Get`) + events (`ReadEvents`) and rebuilds the transcript from the truncated log, so the rewound-away events no longer linger in the transcript or get re-sent as `ContextEvents` on the next prompt. Added `TestRewindRefreshesInMemorySessionState` (resume → `/rewind latest` → submit; asserts the dropped content is gone from `sessionEvents`, transcript, and the next prompt). build/vet/-race/full-suite + windows green. @coderabbitai review

Vasanthdev2004 (approved):
Re-reviewed current head `066ee5c`.

The previous `/rewind` blocker is fixed: after ApplyRewind, the TUI reloads session metadata/events and rebuilds the transcript, and the new regression test verifies rewound-away content is not re-sent in the next prompt.

Validation passed locally:
- `go test ./...`
- `git diff --check origin/agent-runtime...origin/pr-126`

CI/CodeRabbit are green. Approved.

PR #127, opened 2026-06-07
TUI shell: ask_user modal, pickers, autocomplete, zeroline skin

gnanam1990:
## TUI shell — ask_user modal, pickers, autocomplete, zeroline skin

Lands the interactive TUI shell on `main`. The previous PR for this content (#126) was a stacked PR whose base was the `agent-runtime` branch — it merged there, but `#125` had already squash-merged the agent runtime to `main` without it, so the TUI shell never actually reached `main` (no `zeroline_view.go` / `Options.Skin` there). This re-targets the **same reviewed content** onto current `main` (which now has the agent runtime + `OnAskUser`/`ContextWindow`).

### What's in it
- **ask_user** — interactive questionnaire modal wired to `agent.OnAskUser` (answer channel); transcript rows + session payloads; graceful degradation with no interactive surface.
- **Compaction sizing** — `modelContextWindow()` drives the agent's `ContextWindow` from the active model's registry entry (unknown → 0).
- **Pickers** (model / theme / effort / mode) + slash-command **autocomplete** overlay.
- **zeroline skin** rendering (boot / home / chat) behind `Options.Skin`.
- **Checkpoint + rewind** — batches `EventSessionCheckpoint` in order, flushes at end-of-run and on cancel; `/rewind` reloads in-memory session state so dropped events don't re-reach the agent. Exposes `sessions.SnapshotForCheckpoint` with an explicit orphan-blob safe-usage contract for the batch-in-order path.

### Carries the prior review fixes
Both review rounds from #126 are included: the canceled-`ask_user` handling lives in the (already-merged) agent runtime; the `/rewind` state-refresh fix + `TestRewindRefreshesInMemorySessionState` are here.

### Testing
116 tui tests pass. `go build ./...`, `go vet`, `go test ./...`, `go test -race ./internal/tui/`, `GOOS=windows go build ./...` all green. No new deps.

Unblocks the cli wiring PR. Part of decomposing #101.

## Summary by CodeRabbit

* **New Features**
  * Interactive multi-question agent questionnaire (ask_user)
  * Slash-command autocomplete with keyboard navigation, Tab/Enter/Esc behavior
  * /mode presets and /rewind workspace restore
  * Zeroline UI skin with theme picker, theme variants, and keyboard theme controls

* **Improvements**
  * Clearer messages for unknown/redirected model or mode selections
  * Cancellation (Esc/Ctrl+C) now preserves and flushes in‑flight checkpoints to avoid orphaned state
  * Pickers and suggestions correctly respect active modals and keyboard focus

### DIFF
--- a/internal/sessions/checkpoint.go
+++ b/internal/sessions/checkpoint.go
@@ -99,6 +99,11 @@ func (store *Store) CaptureToolCheckpoint(sessionID, workspaceRoot, tool string,
 // Returns ok=false when there is nothing to record (disabled, no paths, or no
 // capturable files).
 func (store *Store) SnapshotForCheckpoint(sessionID, workspaceRoot, tool string, paths []string) (CheckpointPayload, bool) {
+	// Validate the session id (as CaptureToolCheckpoint does) so an exported caller
+	// can't route blob writes through an unexpected/invalid session path.
+	if !ValidSessionID(sessionID) {
+		return CheckpointPayload{}, false
+	}
 	if !CheckpointsEnabled() || len(paths) == 0 {
 		return CheckpointPayload{}, false
 	}
--- a/internal/sessions/checkpoint.go
+++ b/internal/sessions/checkpoint.go
@@ -76,23 +76,29 @@ func (store *Store) CaptureToolCheckpoint(sessionID, workspaceRoot, tool string,
 	}
 	defer unlock()
 
-	payload, ok := store.snapshotForCheckpoint(sessionID, workspaceRoot, tool, paths)
+	payload, ok := store.SnapshotForCheckpoint(sessionID, workspaceRoot, tool, paths)
 	if !ok {
 		return Event{}, nil
 	}
 	return store.appendEventLocked(sessionID, AppendEventInput{Type: EventSessionCheckpoint, Payload: payload})
 }
 
-// snapshotForCheckpoint reads and stores the before-mutation blobs for paths and
-// returns the checkpoint payload WITHOUT appending an event. It is intentionally
-// package-private: the blob writes and the referencing EventSessionCheckpoint
-// must be committed atomically under the same session lock, which only its sole
-// caller CaptureToolCheckpoint guarantees. A snapshot-only entry point exposed to
-// external callers would let blobs be written that a concurrent
-// pruneOrphanBlobs/ApplyRewind could delete before the event is recorded.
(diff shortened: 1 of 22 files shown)


Vasanthdev2004 (approved):
Reviewed current head `c4abf9c` against `main`.

No blockers found. This is the unstacked TUI shell change after the agent-runtime work landed. The previous `/rewind` in-memory state issue is fixed here: the TUI reloads session metadata/events after ApplyRewind and the regression test verifies dropped content is not re-sent in the next prompt.

Validation passed locally:
- `go test ./...`
- `git diff --check origin/main...origin/pr-127`

GitHub Actions are green. CodeRabbit status was still pending during my review, so keep an eye on that status before merge.

gnanam1990:
Addressed all 7:
1. **checkpoint** — `SnapshotForCheckpoint` now validates the session id (`ValidSessionID`), like `CaptureToolCheckpoint`.
2. **permission toggle** — `nextPermissionMode` folds Unsafe → **Ask** (not Auto); toggling never makes an Unsafe session less strict. Test added.
3. **ask_user** — zero-question requests resolve immediately instead of stalling on a prompt.
4. **cancelled flush** — persist-failure rows from `appendSessionEvents` are surfaced, not swallowed.
5. **/rewind** — reload failures are handled explicitly (clear in-memory context + report) so stale events can't reach the next prompt.
6. **transcript dedupe** — ask_user keyed on `row.id` (survives rehydration).
7. **ask_user Esc test** — empty branch replaced with a real assertion.

build/vet/-race/full-suite + windows green. @coderabbitai review

gnanam1990:
Verified both:
1. **zero-question ask_user fast path** — fixed: it now mirrors the normal flow (appends `askUserTranscriptRow` and answers with `[]string{}` instead of `nil` for a consistent Answers shape).
2. **defer Ctrl+C quit on flush failures** — skipping, with reason: deferring/blocking `tea.Quit` after Ctrl+C contradicts the expectation that Ctrl+C exits promptly, and it's a sizable UX change (acknowledgement state machine) for a rare edge. The flush-failure rows are already appended to the transcript (visible on the common non-exit /Esc cancel), and a missed checkpoint degrades gracefully — that single `/rewind` is unavailable, no data loss. Happy to revisit with a stderr-on-exit surface in the cli wiring PR if you'd prefer that over blocking exit.

build/vet/-race/full-suite + windows green. @coderabbitai review

PR #129, opened 2026-06-07
Agent: rich system prompt (coding-craft + workspace context)

gnanam1990:
## Agent: rich system prompt + workspace context

Replaces the one-sentence system prompt — our single biggest agentic-quality gap — with a real coding-craft instruction set, adapted to our Go toolset.

### What changed
- **`internal/agent/system_prompt.md`** (embedded) — identity; autonomy/persistence (bias for action, drive tasks end-to-end); a clear workflow (understand → plan → implement → verify → summarize); editing discipline (prefer native `read_file`/`grep`/`glob`/`edit_file`/`apply_patch` over shelling out, one tool call per file, minimal diffs, match existing style, read-before-edit); a **mandatory testing gate** (run validators after edits; never claim done or commit while failing); tool-use guidance; and communication/markdown style.
- **`internal/agent/system_prompt.go`** — `buildSystemPrompt(options)` assembles core + dynamic **`<environment>`** context (cwd, git branch, first `AGENTS.md`/`ZERO.md`/`.zero/AGENTS.md` project doc, 8 KiB-capped) + the existing safety confirmation policy. Context is omitted when cwd is unset, so headless/test runs stay deterministic.
- **`loop.go`** — prompt decls moved out; callsite threads `options` so cwd reaches the builder.

### Why
Our agent's "intelligence" was barely configured — the loop/plumbing is strong, but a model with a one-line prompt and no project context can't code well. This is the cheapest, largest single quality lever.

### Testing
Updated the two prompt assertions; added `TestBuildSystemPromptInjectsWorkspaceContext`. build / vet / `go test ./...` / `-race` / Windows all green. No new deps.

First of a planned series (next: provider prompt caching, then compaction enrichment + tool-confirmation UX).

## Summary by CodeRabbit

* **New Features**
  * System prompt reworked: agent now uses a single assembled core prompt that includes the confirmation policy, workspace context (working directory and git branch when available), and injected project guidelines (truncated for size).
* **Tests**
  * Updated tests to validate the new prompt assembly and workspace-context injection behavior.

PR #130, opened 2026-06-07
Provider: prompt caching (system + tools)

gnanam1990:
## Prompt caching (system + tools)

Caches the **stable per-run prefix** — the system prompt and tool definitions — so it isn't re-billed or re-processed on every turn. A direct **latency + token-cost win** on multi-turn sessions, and it pays off now that the system prompt is substantial (#129).

### What changed
- **`types.go`** — `system` can be sent as `[]systemBlock` (Anthropic accepts a string *or* blocks); added `cacheControl` and a `cache_control` breakpoint on the **last system block** and the **last tool** (two breakpoints covering the system prompt + the whole tool block); added `cache_read_input_tokens` / `cache_creation_input_tokens` to the usage struct.
- **`provider.go`** — `buildRequest` emits the cacheable system block and tags the last tool; the stream now captures cache read/creation tokens. Anthropic reports `input_tokens` (uncached), `cache_read`, and `cache_creation` **separately**, but the runtime models cached as a subset of input (clamps `cached ≤ input`), so we report the **full** prompt as `InputTokens` and the cache hits as `CachedInputTokens`.
- Caching is GA on `anthropic-version: 2023-06-01` (no beta header); the other providers are unaffected (the field is provider-local).

### Testing
Updated the request-shape assertions (system block + tool `cache_control`); added `TestStreamCompletionReportsCacheTokens`. build / vet / `go test ./...` / `-race` / Windows all green. No new deps.

Pairs with #129 (rich system prompt): the larger, structured prefix is exactly what benefits most from caching.

## Summary by CodeRabbit

* **New Features**
  * Enabled prompt caching for the Anthropic integration so system prompts and final tool definitions can be cached per run.
  * Improved usage accounting: reports include cached-read and cache-creation token components and reflect them in total input token counts.

* **Tests**
  * Added tests validating prompt-caching request structure and verifying cache token breakdown is reported during streaming.

### DIFF
--- a/internal/providers/anthropic/provider.go
+++ b/internal/providers/anthropic/provider.go
@@ -254,8 +254,13 @@ func (provider *Provider) emitDone(ctx context.Context, state *streamState, even
 	state.closeOpen(ctx, events)
 	if state.hasInputUsage || state.hasOutputUsage {
 		usage, err := zeroruntime.NormalizeUsage(zeroruntime.TokenUsage{
-			InputTokens:  state.inputTokens,
-			OutputTokens: state.outputTokens,
+			// Anthropic reports input_tokens (uncached), cache_read, and
+			// cache_creation SEPARATELY. The runtime models the cached count as a
+			// SUBSET of total input (it clamps cached <= input), so report the full
+			// prompt size as InputTokens and the cache hits as CachedInputTokens.
+			InputTokens:       state.inputTokens + state.cacheReadTokens + state.cacheCreationTokens,
+			CachedInputTokens: state.cacheReadTokens,
+			OutputTokens:      state.outputTokens,
 		})
 		if err == nil {
 			providerio.SendEvent(ctx, events, zeroruntime.StreamEvent{Type: zeroruntime.StreamEventUsage, Usage: usage})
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (approved):
Reviewed PR #130 against the Anthropic prompt-caching request shape and Zero's runtime usage model.

Verdict: APPROVED.

Why:
- System prompt cache block shape is valid: `system` as a text block array with `cache_control`.
- Last tool definition cache breakpoint is valid and keeps tool-definition caching scoped to Anthropic.
- Usage mapping is correct for Zero's current model: total input = `input_tokens + cache_read_input_tokens + cache_creation_input_tokens`, with `CachedInputTokens` set only to cache reads.
- Tests cover both request JSON shape and cache usage token normalization.

- `go test ./internal/providers/anthropic -count=1` passed
- `go test ./...` passed

Non-blocking follow-up: Anthropic cache writes have their own pricing multiplier, while Zero currently models only base input and cache-read input pricing. This PR keeps token accounting correct, but a later usage/cost PR should add a separate cache-write bucket if we want Anthropic cost estimates to be exact.

PR #131, opened 2026-06-07
Agent: preserve active plan + loaded skills across compaction

gnanam1990:
## Agent: preserve active plan + loaded skills across compaction

Compaction summarized the elided middle into prose, which silently dropped two pieces of **structured state** the model needs to keep working: the **active plan** (`update_plan`) and any **loaded skill** instructions (`skill` tool). After a compaction, the model would lose its own plan and any skill it had pulled into context.

### What changed
- **`internal/agent/compaction_preserve.go`** (new):
  - `extractLatestPlan` — the most recent `update_plan`, rendered as status-tagged bullets (`- [in_progress] write code`).
  - `extractLoadedSkills` — `skill` tool calls matched to their result bodies by id, latest-per-name, each body capped at 2 KiB so a large skill can't defeat the compaction it's part of.
  - `appendPreservedState` — folds both into the injected summary under clear, labelled sections.
- **`internal/agent/compaction.go`** — `Compact` appends the preserved state verbatim (structured state survives exactly, not paraphrased); `summaryInstructions` now tell the summarizer to **integrate any earlier summary block**, so repeated compactions accumulate rather than lose older facts (delta-aware summarization).

### Why
The plumbing for compaction was correct (boundary widening, no dangling tool results), but a plain prose summary is lossy for exactly the state the agent relies on between turns. Preserving it verbatim keeps long sessions coherent.

### Testing
Plan + skill preservation through `Compact`, no sections when absent, latest-plan-wins, skill calls without a result body skipped. build / vet / `go test ./...` / `-race` / Windows all green. No new deps.

Builds on the existing compaction (#125). Next in the series: per-category context token budget.

## Summary by CodeRabbit

* **Improvements**
  * Conversation compaction now appends a preserved-state block to summaries, carrying prior summary context forward, preserving the latest active plan and loaded skills across repeated compactions, merging newer skill content while preserving order, and capping preserved skill bodies (~2 KiB) with UTF‑8-safe truncation and a truncation note; the preserved block is omitted when there’s nothing to preserve.

* **Tests**
  * Added tests validating state preservation, plan/status rendering, skill preservation and ordering, truncation behavior, and conversations with no preserved state.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -35,7 +35,9 @@ const summaryLabel = "[Summary of earlier conversation]"
 const summaryInstructions = "You are compacting a coding-assistant conversation to save context. " +
 	"Write a dense, factual summary of the conversation so far. Preserve: the user's goals and explicit constraints; " +
 	"decisions made and why; files created or modified (with paths) and key code changes; commands run and their important " +
-	"results; and anything still in progress or unresolved. Omit pleasantries. Use terse bullet points. Do not invent details."
+	"results; and anything still in progress or unresolved. Omit pleasantries. Use terse bullet points. Do not invent details. " +
+	"If the conversation already begins with an earlier summary block, treat its facts as established context and carry them " +
+	"forward into the new summary — never drop earlier information."
 
 // CompactionOptions configure a single Compact call.
 type CompactionOptions struct {
@@ -128,11 +130,15 @@ func Compact(messages []zeroruntime.Message, opts CompactionOptions) ([]zerorunt
 	}
 	summary = strings.TrimSpace(summary)
 
+	// Preserve structured state (active plan + loaded skills) from the elided
+	// middle verbatim, so it is not lost or paraphrased away by the prose summary.
+	content := appendPreservedState(summaryLabel+"\n"+summary, middle)
+
 	compacted := make([]zeroruntime.Message, 0, systemEnd+1+(len(messages)-boundary))
 	compacted = append(compacted, messages[:systemEnd]...)
 	compacted = append(compacted, zeroruntime.Message{
 		Role:    zeroruntime.MessageRoleUser,
-		Content: summaryLabel + "\n" + summary,
+		Content: content,
 	})
 	compacted = append(compacted, messages[boundary:]...)
 	return compacted, nil
--- a/internal/agent/compaction_preserve.go
+++ b/internal/agent/compaction_preserve.go
@@ -1,0 +1,166 @@
+package agent
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// Compaction preservation.
+//
+// A plain prose summary loses two things the model needs to keep working: the
+// active plan (issued via the update_plan tool) and any skill instructions it
+// loaded (via the skill tool). When those turns fall into the elided middle,
+// the plan and skill bodies vanish from context. To prevent that, Compact
+// appends them VERBATIM to the injected summary so structured state survives a
+// compaction exactly rather than being paraphrased away.
+
+const (
+	toolNameUpdatePlan = "update_plan"
+	toolNameSkill      = "skill"
+)
+
+// planPreserveLabel / skillsPreserveLabel head the preserved sections so they
+// are unmistakable in the transcript (and so tests can assert on them).
+const planPreserveLabel = "## Active plan (preserved across compaction)"
+const skillsPreserveLabel = "## Loaded skills (preserved across compaction)"
+
+// maxPreservedSkillBytes caps how much of each loaded skill body is carried
+// across a compaction, so a large skill can't defeat the compaction it is part
+// of. The skill's name and head survive; the model can re-load it in full if it
+// needs the rest.
+const maxPreservedSkillBytes = 2 << 10 // 2 KiB
+
+// extractLatestPlan returns a formatted view of the most recent update_plan tool
+// call in messages, so an in-progress plan survives when its turns are elided by
+// compaction. Returns "" when no plan was issued.
+func extractLatestPlan(messages []zeroruntime.Message) string {
+	for i := len(messages) - 1; i >= 0; i-- {
+		calls := messages[i].ToolCalls
+		for j := len(calls) - 1; j >= 0; j-- {
+			if calls[j].Name != toolNameUpdatePlan {
+				continue
+			}
+			if formatted := formatPlanArguments(calls[j].Arguments); formatted != "" {
+				return formatted
+			}
+		}
+	}
+	return ""
+}
+
+// formatPlanArguments renders an update_plan tool call's JSON arguments
+// ({"plan":[{content,status,...}]}) as terse status-tagged bullet lines. Returns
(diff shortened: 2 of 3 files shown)


Vasanthdev2004 (changes requested):
Reviewed PR #131.

Verdict: CHANGES_REQUESTED.

Findings:

1. [P1] Preserved plan/skills can be lost on the next compaction.
`Compact` appends preserved state by calling `appendPreservedState(..., middle)`, and `appendPreservedState` only extracts real `update_plan` / `skill` tool calls from the elided middle. After the first compaction, the preserved plan/skill sections live only as text inside the injected summary message. On a second compaction, that old summary is usually in `middle`, but there are no real tool calls/results left for `extractLatestPlan` / `extractLoadedSkills` to find, so the structured state can disappear unless the LLM summarizer happens to copy it. That makes the preservation non-deterministic exactly in repeated long-session compactions.

Suggested fix: when `middle` contains an earlier `[Summary of earlier conversation]` message, parse/carry forward existing `## Active plan (preserved across compaction)` and `## Loaded skills (preserved across compaction)` sections as preserved state too. Newer real `update_plan` / `skill` tool calls in `middle` should override older preserved sections. Add a repeated-compaction test where the second summarizer returns text that does not include those sections, and assert the output still keeps the plan and skill body.

Affected area: `internal/agent/compaction.go:127-135`, `internal/agent/compaction_preserve.go:155-165`.

2. [P2] `capBody` does not enforce the byte cap for multibyte UTF-8.
`capBody` checks `len(body)` in bytes, but then truncates by comparing `len(runes)` to `maxPreservedSkillBytes`. A body made of multibyte runes can exceed the byte budget substantially, so a large non-ASCII skill can still defeat the compaction budget. This matches CodeRabbit's finding.

Suggested fix: truncate by accumulated UTF-8 byte length while preserving rune boundaries, and include the truncation note only when the original body exceeded the byte cap. Add a multibyte test asserting `len(result) <= maxPreservedSkillBytes`.

Affected area: `internal/agent/compaction_preserve.go:143-152`.

Validation run locally on PR head `95f2149`:
- `go test ./internal/agent -count=1` passed
- `go test ./...` passed

CI is green and the PR is mergeable, but the repeated-compaction behavior should be fixed before merge.

gnanam1990:
Fixed: `capBody` now enforces the **byte** budget on a UTF-8 rune boundary (was slicing by rune count against a byte constant — multibyte bodies could exceed 2 KiB, and the truncation note could appear without truncation). Room for the note is reserved within the budget, and the note is added only on real truncation. Added tests for multibyte over-budget (`len <= cap` + valid UTF-8) and the runes<cap-but-bytes>cap case. Full suite + Windows green. @coderabbitai review

Vasanthdev2004 (changes requested):
Rereviewed PR #131 on latest head `9f5ffb1`.

Verdict: CHANGES_REQUESTED.

The UTF-8 byte-budget issue is fixed: `capBody` now reserves room for the truncation note, walks back to a rune boundary, and tests cover multibyte input.

Remaining finding:

[P1] Preserved plan/skills can still be lost on a second compaction.
`Compact` still calls `appendPreservedState(summaryLabel+\n+summary, middle)`, and `appendPreservedState` still only extracts real `update_plan` / `skill` tool calls from the elided `middle`. After the first compaction, the preserved plan/skill data lives as plain text inside the injected summary message. If that summary message is later compacted again and the summarizer does not copy those exact sections, there are no real tool calls/results left for the extractor to find, so the preserved state disappears.

Suggested fix: carry forward existing preserved sections from prior summary messages too. Newer real `update_plan` / `skill` tool calls in `middle` should override older preserved sections. Add a repeated-compaction regression test where the second summarizer returns text without the preserved sections and assert the output still keeps the active plan and loaded skill body.

Validation:
- `go test ./internal/agent -count=1` passed

gnanam1990:
Fixed the remaining P1 (repeated-compaction loss):

`appendPreservedState` now **carries forward** the `## Active plan` / `## Loaded skills` sections parsed from the most recent prior summary in `middle` when there are no fresh `update_plan`/`skill` tool calls — so the structured state survives a second compaction even when the summarizer drops the sections. Fresh tool calls still override (skills merged by name; plan: fresh `update_plan` wins, else carried forward).

Added `TestCompactCarriesPreservedStateAcrossRepeatedCompaction` exactly as you specified: a second compaction whose summarizer returns text **without** the sections still keeps the active plan and the loaded skill body. build / -race / full suite / Windows green. @Vasanthdev2004 @coderabbitai review

Vasanthdev2004 (changes requested):
Rereviewed PR #131 on current head `738884914bdd`.

The previous repeated-compaction issue is mostly addressed, and the new regression test covers the summarizer dropping preserved sections. Validation is green. However CodeRabbit's latest current-head finding is real and should be fixed before merge:

[P1] Preserved skill carry-forward is ambiguous for real skill docs.
`sectionText` ends a preserved section at the next `\n## `, and `parseSkillSection` later splits skills on raw `### `. Loaded skill bodies are preserved verbatim, and our actual skill docs commonly contain Markdown headings like `## Workflow`, `## Usage`, or `### Examples`. On the next compaction, those headings can truncate the preserved skill body or turn body headings into fake skills, so the new repeated-compaction guarantee still fails for normal skill content.

Affected area: `internal/agent/compaction_preserve.go` around `sectionText` / `parseSkillSection`.

Smallest fix: use an unambiguous carry-forward encoding for preserved skills, such as a JSON block/base64 body block under the preserved heading, or explicit length-delimited fields, then add a regression where a loaded skill body contains both `##` and `###` headings and survives a second compaction intact.

Validation I ran locally:
- `go test ./internal/agent -count=1`
- `go test ./...`

gnanam1990:
Fixed (major, carry-forward ambiguity): the preserved plan + skills are now serialized as a single line of **JSON** (`formatPreservedState` / `parsePreservedState`) instead of parsing markdown delimiters. JSON escaping makes the round-trip lossless for arbitrary bodies — markdown headings (`##`/`###`), code fences, quotes, multibyte — so a skill doc with headings is no longer truncated or split into bogus skills on the next compaction.

Added `TestCompactPreservesSkillBodyWithMarkdownHeadings` (your requested regression): a body containing `## Usage` / `### Examples` / a code fence stays byte-identical across **two** compactions. build / -race / full suite / Windows green. @coderabbitai @Vasanthdev2004 review

gnanam1990:
Fixed: the round-trip test helper now guards the compacted shape (`len(out)>=2` + user role) before indexing `out[1]`. @coderabbitai review

Vasanthdev2004 (approved):
Rereviewed PR #131 on current head `171dc02adcde`.

Verdict: **APPROVED**

The previous blockers are fixed. Preserved plan/skill state is now carried as a labelled single-line JSON block, so skill bodies with `##` / `###` headings, code fences, and quotes round-trip without markdown delimiter ambiguity. The repeated-compaction regression now covers a skill body with headings/code fence, and the latest test guard avoids indexing `out[1]` before checking the compacted shape. CodeRabbit also reports no actionable comments on this head.

Validation I ran locally:
- `go test ./internal/agent -count=1`
- `go test ./...`

PR #132, opened 2026-06-07
Providers: unified transient-failure retry (network, 429, 5xx)

gnanam1990:
## Providers: unified transient-failure retry (network, 429, 5xx)

Only the OpenAI provider retried before — **Anthropic and Gemini surfaced the first failure**, so a dropped connection or an intermittent gateway 5xx/429 failed the whole turn. This centralizes one retry policy across every provider.

### What changed
- **`internal/providers/providerio/retry.go`** (new) — `SendWithRetry` rebuilds the request each attempt (replay-safe) and retries **network errors + retryable statuses (429 and 5xx)** up to N attempts, backing off between tries and honoring a server **`Retry-After`** header and **context cancellation**. Only the *initial* request is retried — a partially consumed stream is never replayed. Plus `Backoff`, `ShouldRetryStatus`, `RetryAfter`.
- **anthropic / gemini** — replace the bare `http.Do` (no retry) with `SendWithRetry`.
- **openai** — replace its bespoke retry loop + `backoff` func with the shared helper (which now *also* covers 429 + `Retry-After`, which the old loop didn't). Net code reduction.

### Why
Resilience was inconsistent: two of three providers had zero retry. A single shared, tested policy fixes that and removes duplicated logic.

### Testing
retry-then-succeed, no-retry on 400, exhausted-retries surface the last response, `Retry-After` parsing, backoff context cancellation. build / vet / `go test ./...` / `-race` / Windows all green. No new deps.

## Summary by CodeRabbit

* **Refactor**
  * Streaming providers now use a shared, centralized retry mechanism for more consistent and reliable recovery from transient server/network errors and improved error signaling.

* **Tests**
  * Added comprehensive tests covering retry behavior, backoff timing, Retry-After parsing, and retry success/failure scenarios to ensure robust streaming behavior.

### DIFF
--- a/internal/providers/anthropic/provider.go
+++ b/internal/providers/anthropic/provider.go
@@ -1,7 +1,6 @@
 package anthropic
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -115,24 +114,19 @@ func (provider *Provider) stream(ctx context.Context, body []byte, events chan<-
 	streamCtx, cancelStream := context.WithCancel(ctx)
 	defer cancelStream()
 
-	httpRequest, err := http.NewRequestWithContext(streamCtx, http.MethodPost, provider.baseURL+"/v1/messages", bytes.NewReader(body))
-	if err != nil {
-		providerio.SendEvent(ctx, events, zeroruntime.StreamEvent{Type: zeroruntime.StreamEventError, Error: provider.redact("provider request error: " + err.Error())})
-		return
-	}
-	httpRequest.Header.Set("Content-Type", "application/json")
-	httpRequest.Header.Set("anthropic-version", provider.version)
-	if provider.apiKey != "" {
-		httpRequest.Header.Set("x-api-key", provider.apiKey)
-	}
-	if provider.beta != "" {
-		httpRequest.Header.Set("anthropic-beta", provider.beta)
-	}
-	if provider.userAgent != "" {
-		httpRequest.Header.Set("User-Agent", provider.userAgent)
-	}
-
-	response, err := provider.httpClient.Do(httpRequest)
+	response, err := providerio.SendWithRetry(streamCtx, provider.httpClient, http.MethodPost, provider.baseURL+"/v1/messages", body, func(request *http.Request) {
+		request.Header.Set("Content-Type", "application/json")
+		request.Header.Set("anthropic-version", provider.version)
+		if provider.apiKey != "" {
+			request.Header.Set("x-api-key", provider.apiKey)
+		}
+		if provider.beta != "" {
+			request.Header.Set("anthropic-beta", provider.beta)
+		}
+		if provider.userAgent != "" {
+			request.Header.Set("User-Agent", provider.userAgent)
+		}
+	}, 0)
 	if err != nil {
 		providerio.SendEvent(ctx, events, zeroruntime.StreamEvent{Type: zeroruntime.StreamEventError, Error: provider.redact("provider stream error: " + err.Error())})
 		return
--- a/internal/providers/gemini/provider.go
+++ b/internal/providers/gemini/provider.go
@@ -1,7 +1,6 @@
 package gemini
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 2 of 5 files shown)


Vasanthdev2004 (changes requested):
Reviewed PR #132.

Verdict: CHANGES_REQUESTED.

Finding:

[P1] Transport-error retries can double-submit non-idempotent completion requests.
`SendWithRetry` retries every `client.Do(request)` error before returning. That is safe for failures that happen before the request leaves the client, but `client.Do` can also fail after the provider has received some or all of the POST body and before the client receives response headers/data. For streaming completion calls, retrying that branch can submit a second billable generation with no provider-side dedupe/idempotency key. This was already a risk in the old OpenAI-only loop, but this PR centralizes and expands it to Anthropic and Gemini too, so it becomes a cross-provider policy.

I would keep retrying HTTP 429/5xx responses, because the provider has explicitly returned a retryable status before any stream is consumed. For `err != nil` transport failures, either disable retries by default for these non-idempotent provider calls, or make network-error retry opt-in only when a provider/callsite supplies a real idempotency/deduplication mechanism. A rewindable request body / `GetBody` is not enough; it only proves local replay is possible, not that the first request was not processed.

Suggested small fix:
- Change `SendWithRetry` to accept retry options, e.g. `RetryTransportErrors bool`, default false.
- Keep `ShouldRetryStatus(429/5xx)` behavior for all providers.
- Set `RetryTransportErrors` false at current OpenAI/Anthropic/Gemini callsites unless/until we add real provider idempotency support.
- Add tests for both paths: transport error not retried by default, and transport error retried only when explicitly enabled.

Validation run locally on a test merge of PR #132 into current `origin/main`:
- `git diff --check` passed
- `go test ./internal/providers/providerio ./internal/providers/openai ./internal/providers/anthropic ./internal/providers/gemini -count=1` passed
- `go test ./...` passed

The merge applies cleanly and preserves the recent Anthropic prompt-caching code, but the transport-error retry policy needs tightening before merge.

gnanam1990:
Fixed: `SendWithRetry` no longer replays transport/network errors — a completion POST is non-idempotent and may have reached the server, so a connection failure now surfaces immediately instead of risking a duplicate (billable) generation. Retries are now limited to **429/5xx** server responses, where the request was explicitly rejected without effect. Added `TestSendWithRetryDoesNotReplayTransportErrors` (transport error → exactly 1 attempt). -race + providers suite + Windows green. @coderabbitai review

gnanam1990:
Addressed both:
1. **Generic 5xx replay risk (major):** narrowed `ShouldRetryStatus` to **429 + 503** only — the statuses where the server explicitly did *not* accept the request, so a replay can't duplicate a billable completion. 500/502/504 are ambiguous (a 504 may follow an upstream that already generated output) and now surface to the caller's HTTP-error path instead of being retried. Docs corrected to match.
2. **bodyclose:** the transport-error test now captures and closes the response when present.

-race + full suite green. @coderabbitai review

Vasanthdev2004 (approved):
Rereviewed PR #132 on latest head `fa2f4dd` and a clean test merge into current `origin/main`.

Verdict: APPROVED.

The prior blocker is fixed: transport errors are no longer retried, so non-idempotent completion POSTs are not replayed after an ambiguous `client.Do` failure. Retryable HTTP statuses are now narrowed to 429 and 503, while 500/502/504 are deliberately treated as ambiguous and not replayed. The merge result also preserves the recent Anthropic prompt-caching request shape and cache-token usage mapping.

Validation:
- `git diff --check origin/main...HEAD` passed on PR head
- `git diff --check` passed on test merge into current `origin/main`
- `go test ./internal/providers/providerio ./internal/providers/openai ./internal/providers/anthropic ./internal/providers/gemini -count=1` passed on the merge result

Non-blocking: CodeRabbit still has small test-quality nits (`Retry-After` HTTP-date coverage and checking `resp.Body.Close()` in a test). Those are worth cleaning up, but I do not see a blocking runtime correctness issue now.

gnanam1990:
Fixed: the transport-error test now checks `resp.Body.Close()`'s error (was dropped → errcheck/lint). @coderabbitai review

Vasanthdev2004 (approved):
Rereviewed PR #132 on current head `f4bd9efb85f0`.

Verdict: **APPROVED**

The only new commit since my prior approval handles the `resp.Body.Close()` error in the transport-error test. The runtime retry behavior remains correct: transport errors are not replayed, retries are limited to 429/503 responses, and 500/502/504 stay non-retryable for non-idempotent completion POSTs. CodeRabbit also reports no actionable comments on this head.

Validation I ran locally:
- `go test ./internal/providers/providerio ./internal/providers/openai ./internal/providers/anthropic ./internal/providers/gemini -count=1`
- `go test ./internal/providers/... -count=1`

PR #134, opened 2026-06-07
Background: escalate process termination from SIGTERM to SIGKILL

gnanam1990:
## Background: escalate process termination from SIGTERM to SIGKILL

`terminateProcess` (POSIX) sent **only SIGTERM**, so a background process that traps or ignores it kept running after `Kill`/`KillRunning` — a process **leak** (notably on a Ctrl+C shutdown that kills running tasks).

### What changed
- **`internal/background/process_posix.go`** — SIGTERM (so the process can flush/clean up) → poll liveness with signal 0 until a **3s grace** deadline → **SIGKILL** if still alive. An already-exited process (ESRCH / `ErrProcessDone`) is treated as success. Grace/poll are package vars so tests run fast.
- **Windows is unchanged** — it already force-kills via `taskkill /T /F`.

### Why
A force-kill that can't actually force-kill isn't one. This guarantees a killed task is gone.

### Testing
- A SIGTERM-ignoring process (`trap '' TERM` + a readiness handshake so we don't signal before the trap installs) is killed **only after** the grace period — proving SIGTERM is tried first, then escalated.
- A well-behaved process dies on SIGTERM **well before** escalation.
- An already-exited pid is a no-op.

`-race` + full suite + `GOOS=windows` build green. No new deps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unix-like process termination: send graceful termination, poll during a configurable grace period, then escalate to forceful termination; treat already-exited processes as successful no-ops.
  * Ensure tasks marked as killed remain consistently recorded; on failed kills, restore running state to avoid inconsistent status.

* **Tests**
  * Added POSIX-only tests for escalation to forceful termination, graceful shutdown within the grace period, already-exited handling, and task-kill race conditions.

### DIFF
--- a/internal/background/manager.go
+++ b/internal/background/manager.go
@@ -267,10 +267,21 @@ func (manager *Manager) Kill(taskID string) error {
 	if !manager.isRunningPID(taskID, pid) {
 		return nil
 	}
+	// Record the kill intent BEFORE terminating. terminateProcess can block until
+	// the child exits, during which the background Wait-goroutine may reap it and
+	// call MarkExited; marking killed first makes that MarkExited a no-op (it only
+	// acts on a running task), so a user-initiated stop stays "killed" instead of
+	// being clobbered to "error".
+	if err := manager.markKilledIfStillRunning(taskID, pid); err != nil {
+		return err
+	}
 	if err := manager.killProcess(pid); err != nil {
+		// Couldn't terminate — undo the optimistic kill mark so the still-running
+		// task is not falsely reported as killed.
+		manager.restoreRunningAfterFailedKill(taskID, pid)
 		return fmt.Errorf("kill background task %s: %w", taskID, err)
 	}
-	return manager.markKilledIfStillRunning(taskID, pid)
+	return nil
 }
 
 func (manager *Manager) KillRunning() error {
@@ -332,6 +343,26 @@ func (manager *Manager) markKilledIfStillRunning(taskID string, pid int) error {
 	}
 	manager.tasks[taskID] = task
 	return nil
+}
+
+// restoreRunningAfterFailedKill reverts a task that was optimistically marked
+// killed back to running when the terminate failed (the process is presumed
+// still alive). It only touches a task still killed with the same pid, so a real
+// exit recorded meanwhile is left untouched.
+func (manager *Manager) restoreRunningAfterFailedKill(taskID string, pid int) {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	task, ok := manager.tasks[taskID]
+	if !ok || task.Status != StatusKilled || task.PID != pid {
+		return
+	}
+	task.Status = StatusRunning
+	task.ExitCode = 0
+	task.CompletedAt = time.Time{}
+	if err := manager.persistTaskLocked(task); err != nil {
+		return
+	}
+	manager.tasks[taskID] = task
 }
 
 func (manager *Manager) loadTasks() error {
--- a/internal/background/manager.go
+++ b/internal/background/manager.go
@@ -272,14 +272,23 @@ func (manager *Manager) Kill(taskID string) error {
 	// call MarkExited; marking killed first makes that MarkExited a no-op (it only
 	// acts on a running task), so a user-initiated stop stays "killed" instead of
 	// being clobbered to "error".
-	if err := manager.markKilledIfStillRunning(taskID, pid); err != nil {
+	marked, err := manager.markKilledIfStillRunning(taskID, pid)
+	if err != nil {
 		return err
+	}
+	if !marked {
+		// The task exited (or was reused) between the running check and now — the
+		// pid may be stale, so do NOT signal it.
+		return nil
 	}
 	if err := manager.killProcess(pid); err != nil {
 		// Couldn't terminate — undo the optimistic kill mark so the still-running
 		// task is not falsely reported as killed.
-		manager.restoreRunningAfterFailedKill(taskID, pid)
-		return fmt.Errorf("kill background task %s: %w", taskID, err)
+		killErr := fmt.Errorf("kill background task %s: %w", taskID, err)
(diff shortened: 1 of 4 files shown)


gnanam1990:
Fixed: both tests now assert the **actual terminating signal** via `WaitStatus.Signal()` — the graceful path must be SIGTERM and the SIGTERM-ignoring process must be SIGKILL — so a regression to an immediate force-kill (or to never escalating) is caught, not just timing. -race green. @coderabbitai review

Vasanthdev2004 (changes requested):
Verdict: **Request changes**

I found one lifecycle blocker on the current head `73154c4c2a0c`.

[P1] Preserve `killed` status when POSIX stop waits for the child to exit

`terminateProcess` now waits until `processAlive` is false before returning. For background specialists, `launchBackgroundProcess` also has a `command.Wait()` goroutine that calls `manager.MarkExited(...)`. On a normal SIGTERM stop, that waiter can reap the process and mark the task `error` with exit code `-1` before `Manager.Kill` reaches `markKilledIfStillRunning`. Because `markKilledIfStillRunning` no-ops once the task is no longer `running`, the persisted task can end up as `error` even though `TaskStop` returned `status: killed`.

Path: `internal/background/process_posix.go:39` waits for exit, while `internal/specialist/exec.go:652` reaps and calls the `onExit` status update before `internal/background/manager.go:273` marks killed.

Smallest fix: make user-initiated stop intent win over the waiter before/while waiting for process exit, or keep SIGTERM return non-blocking and do SIGKILL escalation separately. Please add a POSIX lifecycle regression test around `launchBackgroundProcess`/`Manager.Kill` (not only `terminateProcess`) that asserts the final task status remains `killed` after the waiter runs.

Validation I ran locally:
- `go test ./internal/background -count=1`
- `go test ./internal/specialist -count=1`

Note: I’m on Windows, so the new `//go:build !windows` POSIX signal tests are not executed locally; the CI Ubuntu/macOS smoke jobs are the platform evidence for those tests.

gnanam1990:
Fixed the P1 lifecycle race:

`Manager.Kill` now records the kill intent (`markKilledIfStillRunning`) **before** terminating. Since `MarkExited` only acts on a *running* task, once the task is marked killed the background Wait-goroutine's `MarkExited` becomes a no-op — so a user-initiated stop stays **killed** even when the (now blocking) terminate lets the waiter reap the child first. On a terminate *failure* the optimistic mark is rolled back (`restoreRunningAfterFailedKill`), preserving the not-killed-on-failure invariant.

Added `TestManagerKillStaysKilledWhenExitWaiterReapsDuringTerminate` — a cross-platform regression around `Manager.Kill` with a `KillProcess` that simulates the waiter reaping mid-terminate, asserting the final status stays `killed`. build / -race / full suite / Windows green. @Vasanthdev2004 @coderabbitai review

Vasanthdev2004 (approved):
Rereviewed PR #134 on current head `39cbd06bef12`.

Verdict: **APPROVED**

The previous blocker is fixed. `Manager.Kill` now records user kill intent before the blocking POSIX terminate path, so the background `Wait()` callback cannot clobber a user-initiated stop into `error`. The new regression test simulates the waiter calling `MarkExited` while termination is in progress and asserts the final task state remains `killed`.

Validation I ran locally:
- `go test ./internal/background -count=1`
- `go test ./internal/specialist -count=1`
- `go test ./...`

Note: I’m on Windows, so the POSIX signal-specific tests remain CI-covered rather than locally executed here.

Vasanthdev2004 (changes requested):
Updated rereview for PR #134 on current head `39cbd06bef12`.

The original stop-vs-waiter race is fixed, but CodeRabbit's two current-head findings are valid and should be addressed before merge:

1. [P1] `Kill` can still signal after the optimistic killed mark was not applied.
`markKilledIfStillRunning` returns `nil` both when it successfully marks the task and when the task is no longer running. `Kill` then proceeds to `killProcess(pid)` either way. In the exit/reuse race, that can still signal a stale PID. Make `markKilledIfStillRunning` return whether it actually marked the task, and only call `killProcess` when the mark was applied.

2. [P2] Rollback persistence failures are swallowed.
If `killProcess` fails after the optimistic mark and `restoreRunningAfterFailedKill` cannot persist the rollback, callers only see the kill error while the task may remain stored as `killed`. Have `restoreRunningAfterFailedKill` return an error and join/wrap it with the original kill error.

Validation I ran locally before this verdict:
- `go test ./internal/background -count=1`
- `go test ./internal/specialist -count=1`
- `go test ./...`

So the build/test surface is green, but the manager edge cases need a small fix.

gnanam1990:
Two follow-ups fixed:
1. **Stale-pid signal:** `markKilledIfStillRunning` now reports whether it actually marked; if the task exited/was reused between the running check and the mark, `Kill` returns **without** calling `killProcess`, so it never signals a stale pid.
2. **Swallowed rollback error:** `restoreRunningAfterFailedKill` returns its persist error, and `Kill` joins it with the kill error (`errors.Join`) instead of leaving the task wrongly marked killed.

Added `TestMarkKilledIfStillRunningReportsWhetherItMarked`. -race + full suite + Windows green. @coderabbitai @Vasanthdev2004 review

Vasanthdev2004 (approved):
Rereviewed PR #134 on current head `3312ffa9d5f2`.

Verdict: **APPROVED**

The prior manager edge cases are fixed. `markKilledIfStillRunning` now reports whether the optimistic killed mark was actually applied, so `Kill` does not signal a possibly stale PID when the task is no longer running. The failed-kill rollback now returns and surfaces persistence failures instead of silently leaving the task marked killed. CodeRabbit has also approved this head.

Validation I ran locally:
- `go test ./internal/background -count=1`
- `go test ./internal/specialist -count=1`
- `go test ./...`

Note: I’m on Windows, so the POSIX signal-specific tests are CI-covered rather than locally executed here.

PR #135, opened 2026-06-08
Model harness: per-model system prompts + summarizer fallback

gnanam1990:
## Model harness: per-model system prompts + summarizer fallback

Two agent-quality improvements that compound with the rich system prompt (#129) and prompt caching (#130) already on main.

### 1. Per-model system-prompt specialization
`internal/agent/system_prompt_models.go` — `buildSystemPrompt` now appends a family-specific guidance block chosen by model id:
- **OpenAI family** (gpt/o1/o3/o4): markdown output spec + "prefer native file tools over shell, one call per file" + persistence.
- **Gemini**: tool-preference + conciseness + use `update_plan`.
- **Anthropic/Claude**: comment discipline (match existing density).

Classification is by model string (`modelFamily`) — no registry dependency. Unknown/unset model → no block.

### 2. Chunked summarization fallback
`internal/agent/compaction.go` — if the elided middle is itself **too large to summarize in one call** (a summary-time context-limit error), `summarizeWithFallback` splits it in half recursively and joins the partial summaries, so compaction keeps working past the summarizer's own input window. Non-context errors and an unsplittable single message surface unchanged. Adapted to our single-provider agent (the "multi-model fallback" idea, made to fit).

### Testing
`modelFamily` classification + addendum selection; chunked-fallback split-and-join, non-context propagation, single-message edge. build / vet / `go test ./...` / `-race` / Windows all green. No new deps.

First of three thematic PRs from the local feature set (next: hardening = graceful shutdown + crash capture; tool quality = denial categorization + secret scanning).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved summarization fallback: long conversations are automatically chunked, partial summaries are recombined and re-summarized to avoid context-limit failures.
  * Enhanced model-specific system prompt guidance so prompts include appropriate family-specific addenda.

* **Tests**
  * Added tests covering the fallback summarization behavior and model-family prompt selection.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -344,7 +344,20 @@ func summarizeWithFallback(ctx context.Context, provider Provider, messages []ze
 	if rightErr != nil {
 		return "", rightErr
 	}
-	return strings.TrimSpace(left + "\n\n" + right), nil
+
+	// Re-summarize the two partial summaries into ONE so the persisted summary
+	// stays a single, further-summarizable unit — not an ever-growing concatenated
+	// blob that a later compaction (which can't split a single message) would fail
+	// on. If even the combined partials don't fit (extreme), fall back to the
+	// joined text: still better than failing, and each half is already compacted.
+	combined := strings.TrimSpace(left + "\n\n" + right)
+	reduced, reduceErr := summarizeMessagesOnce(ctx, provider, []zeroruntime.Message{
+		{Role: zeroruntime.MessageRoleUser, Content: combined},
+	})
+	if reduceErr != nil {
+		return combined, nil
+	}
+	return reduced, nil
 }
 
 // summarizeMessagesOnce performs a single tool-less summarization call.
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -315,27 +315,60 @@ func (state *compactionState) recover(
 // OnUsage callbacks, so compaction stays invisible on the user-facing surface.
 func summarizeClosure(ctx context.Context, provider Provider) func([]zeroruntime.Message) (string, error) {
 	return func(toSummarize []zeroruntime.Message) (string, error) {
-		request := zeroruntime.CompletionRequest{
-			Messages: []zeroruntime.Message{
-				{Role: zeroruntime.MessageRoleSystem, Content: summaryInstructions},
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (changes requested):
Verdict: REQUEST CHANGES

Finding:
[P1] Avoid persisting chunk fallback as one unsplittable summary - `internal/agent/compaction.go:338`

`summarizeWithFallback` handles a context-limit error by recursively summarizing the left and right halves, then returns `strings.TrimSpace(left + \n\n + right)`. `Compact` persists that returned text as one user summary message. On the next compaction pass, that prior summary is now a single message; if the concatenated chunk summaries alone exceed the summarizer window, the fallback reaches the `len(messages) < 2` guard and can no longer split it. That means a long-running session may compact successfully once, then get stuck again on the next context-limit recovery.

Please either re-summarize the combined partial summaries into a bounded final summary, or store/preserve the partial summaries in a split-friendly shape and update `Compact` accordingly. The model-family prompt addendum looks scoped and tested; I did not see a blocker there.

Validation:
- `go test ./internal/agent -count=1`
- `go test ./...`

gnanam1990:
Fixed (major): after summarizing the halves, `summarizeWithFallback` now **re-summarizes** the two partial summaries into one compressible unit instead of persisting a joined blob — so a later compaction (which can't split a single message) can still make progress. If even the combined partials don't fit, it falls back to the joined text (still the already-compacted halves). Added `TestSummarizeWithFallbackReSummarizesPartialsIntoOne`. -race + full suite + Windows green. @coderabbitai @Vasanthdev2004 review

Vasanthdev2004 (approved):
Verdict: APPROVED

Rereviewed current head `1eb51ad3c3db62001e89d29290b4e485a7eaaffd`.

The previous compaction blocker is addressed: chunked summaries are now re-summarized into a single reduced summary before persistence, and the regression test covers that it no longer returns the joined partial blob in the normal successful reduction path. Model-family prompt addendums remain scoped and tested. I noticed the best-effort reduce-error fallback still returns the combined partials if the final reduction fails; that is acceptable as a non-blocking residual edge for this PR, but tightening it to only swallow context-limit errors would be a nice follow-up.

Validation:
- `go test ./internal/agent -count=1`
- `go test ./...`

PR #136, opened 2026-06-08
Real-world hardening: graceful shutdown + crash capture

gnanam1990:
## Real-world hardening: graceful shutdown + crash capture

Two fail-open robustness improvements for headless runs and the CLI entrypoint.

### 1. Graceful shutdown of `zero exec` (Ctrl+C)
`internal/cli/shutdown.go` — `zero exec` wrapped `agent.Run` in `context.Background()`, so Ctrl+C / SIGTERM killed the process abruptly, leaking the in-flight provider request and any spawned children. `runExec` now wraps the run in a signal-aware context (`signalContext`): a signal cancels the run, the loop returns cleanly, and exec reports `Interrupted.` with exit **130** (stream-json: an `interrupted` run-end). Scoped to `runExec` — no dispatch-wide signature churn.

> ⚠️ The signal *runtime* delivery can't be exercised in a unit test — please do one manual `zero exec "<long task>"` + Ctrl+C to confirm the path. The helper + no-regression of the exec suite are unit-tested.

### 2. Crash capture (panic → saved report)
`internal/observability/crash.go` — a recovered top-level panic is written to a local crash report (timestamp, label, full stack) under `~/.zero/crashes` and surfaced as a brief notice, instead of a raw stack trace. Wired into `runWithDeps` via a named-return defer. Dependency-free and fail-open; this is the hook a Sentry/OTEL adapter can attach to later for remote reporting (deferred — needs a real endpoint to verify).

### Testing
`signalContext` (not-cancelled-initially, stop-cancels); crash report formatting/writing, panic capture sets the crash exit code + writes one report, no-panic is a no-op. build / vet / `go test ./...` / `-race` / Windows green. No new deps.

## Summary by CodeRabbit

* **New Features**
  * Crash reporting: Unexpected application crashes are now automatically captured and saved as timestamped crash reports containing stack traces and metadata, instead of displaying raw error output to the user.
  * Signal handling: Pressing Ctrl+C or sending termination signals now triggers graceful shutdown with appropriate exit codes, replacing the previous behavior of exiting with provider error codes.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/hooks"
 	"github.com/Gitlawb/zero/internal/mcp"
+	"github.com/Gitlawb/zero/internal/observability"
 	"github.com/Gitlawb/zero/internal/plugins"
 	"github.com/Gitlawb/zero/internal/providers"
 	"github.com/Gitlawb/zero/internal/sandbox"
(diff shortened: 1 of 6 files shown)


Vasanthdev2004 (approved):
Verdict: APPROVED

Reviewed current head `4719815ce94f184759726fa638c0e979873852f9`.

The shutdown path now routes headless agent runs through a signal-cancelled context and maps cancellation to the conventional interrupted exit code, and the panic recovery path writes a local crash report while keeping the CLI failure user-readable. The tests cover the crash report/recover behavior and the signal context basics. I did not find a blocking correctness issue.

Validation:
- `go test ./internal/cli ./internal/observability -count=1`
- `go test ./...`

PR #137, opened 2026-06-08
Tool quality: per-tool denial categorization + secret scanning

gnanam1990:
## Tool quality: per-tool denial categorization + secret scanning

Two additive improvements to tool execution.

### 1. Structured per-tool denial categorization
`internal/agent/types.go` / `loop.go` — adds `ToolResult.DenialReason` (`DenialCategory`: `filtered` / `permission_denied` / `sandbox_violation`) so a surface can tell *precisely* why a call was blocked instead of parsing `Output`. Set in the not-enabled path and `deniedPermissionResult`, with sandbox-violation denials categorized distinctly from approval-declined.

### 2. Secret scanning + redaction of bash output
`internal/secrets` (new) — a dependency-free, precision-first scanner for high-confidence secret shapes (AWS/Google keys, GitHub/Slack tokens, private-key blocks, JWTs). `bash` output now runs through `secrets.Redact`, so a command that prints a secret has it replaced with a typed placeholder + a redaction notice, instead of echoing it back into the model context. This is **additive** to the existing configured-key scrub at the registry boundary, and a built-in regex scanner fits the CLI without the heavy gitleaks dependency.

### Testing
denial categorization (filtered / sandbox-violation vs approval-declined); scanner detection, no-false-positives on ordinary text, redaction, and the bash wiring. build / vet / `go test ./...` / `-race` / Windows green. No new deps.

## Summary by CodeRabbit

* **New Features**
  * Tool call denials now include explicit categories (filtered, permission denied, sandbox violation).
  * Detected secrets in command outputs are automatically redacted and a redaction count/message is shown.
  * Categorized denials are treated as non-retriable to avoid repeated retries.

* **Tests**
  * Added tests covering denial categorization and secret detection/redaction.

### DIFF
--- a/internal/agent/denial_test.go
+++ b/internal/agent/denial_test.go
@@ -1,0 +1,41 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/sandbox"
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
+func TestExecuteToolCallCategorizesFilteredDenial(t *testing.T) {
+	root := t.TempDir()
+	registry := tools.NewRegistry()
+	registry.Register(tools.NewReadFileTool(root))
+
+	result, _ := executeToolCall(context.Background(), registry, ToolCall{ID: "c1", Name: "read_file", Arguments: `{"path":"x"}`}, PermissionModeAuto, Options{
+		DisabledTools: []string{"read_file"},
+	})
+	if result.DenialReason != DenialFiltered {
+		t.Fatalf("DenialReason = %q, want %q", result.DenialReason, DenialFiltered)
+	}
+	if result.Status != tools.StatusError {
+		t.Fatalf("status = %q, want error", result.Status)
+	}
+}
+
+func TestDeniedPermissionResultCategories(t *testing.T) {
+	// Plain approval-declined.
+	plain := deniedPermissionResult(ToolCall{ID: "c1", Name: "bash"}, "needs approval", PermissionEvent{ToolName: "bash"})
+	if plain.DenialReason != DenialPermissionDenied {
+		t.Fatalf("plain denial = %q, want %q", plain.DenialReason, DenialPermissionDenied)
+	}
+	// Sandbox-violation-driven denial.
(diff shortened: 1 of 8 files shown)


Vasanthdev2004 (changes requested):
Verdict: REQUEST CHANGES

Findings:

[P0] Private-key redaction only removes the BEGIN header - `internal/secrets/scanner.go:36`

The `private_key_block` pattern matches only `-----BEGIN ... PRIVATE KEY-----`, and `Redact` replaces exactly the matched text. For a real PEM/OpenSSH private key, the key body and footer remain in the bash output and can still enter model/session context. This defeats the main safety goal of the secret scanner. Please make the private-key pattern span the whole block through the matching END line and add a multiline redaction regression test that proves the body is removed.

[P2] Retriable tool-error classification still parses output strings - `internal/agent/loop.go:407`

This PR adds `ToolResult.DenialReason`, but `isRetriableToolError` still relies on human-readable `Output` substrings to avoid retrying permission/sandbox/filter denials. That keeps the new structured categorization from protecting this path and can regress if wording changes. Please prefer `DenialReason` when present, then keep the legacy/meta checks as fallback.

Validation:
- `go test ./internal/agent ./internal/secrets ./internal/tools -count=1`
- `go test ./...`

gnanam1990:
Both fixed:
1. **Private-key leak (critical):** the `private_key_block` pattern now matches the entire `BEGIN…END` block (body included), so `Redact` removes the key material, not just the header. New `TestRedactPrivateKeyBlockRemovesBody` asserts the body is gone.
2. **Denial classification:** `isRetriableToolError` now treats a structured `DenialReason` as non-retriable regardless of wording (text checks kept as fallback); added test cases. -race + full suite + Windows green. @coderabbitai @Vasanthdev2004 review

Vasanthdev2004 (approved):
Verdict: APPROVED

Rereviewed current head `4a0783500be6f3b415da7058a6dbc31cc5364c0c`.

The prior blockers are fixed: private-key scanning now matches the full PEM/OpenSSH block through the END marker and has a multiline redaction regression test, and retriable tool-error classification now checks structured `DenialReason` before falling back to legacy text/meta checks. I did not find a remaining blocking issue.

Validation:
- `go test ./internal/agent ./internal/secrets ./internal/tools -count=1`
- `go test ./...`

PR #144, opened 2026-06-08
Add spec mode foundation

anandh8x:
Summary:
- Adds the spec-mode foundation: `internal/specmode`, Zero-native draft prompt, `ExitSpecMode`, and `.zero/specs/<date>-<slug>.md` draft saving.
- Adds `spec-draft` permission mode, a `spec_review_required` stop reason, spec session kinds/status/events/metadata, and `zero sessions list --kind`.
- Registers `ask_user` as a core read-only tool now that the agent loop intercept exists, and updates specialist known-tool validation.

Self-review:
- This intentionally does not add the user-facing `/spec` TUI command or `zero exec --use-spec` approval flow yet; this PR lands the core primitive those surfaces need.
- Spec-draft hides and denies write/shell/network/update_plan tools while allowing read-only tools, `ask_user`, and `ExitSpecMode`.
- Spec status remains in session metadata/events, not in the markdown spec file.

Tests:
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Spec-draft mode: read-only inspection, a submit-for-review tool that saves drafts and triggers review, tool-level visibility rules, and early termination when review is requested
  * System prompt can be overridden by caller-supplied text
  * Session list supports a --kind filter and shows spec-related fields; session tracking now records spec lifecycle metadata
  * Core tools registry now exposes an ask-user tool

* **Tests**
  * Extensive tests for spec-draft behavior, submit-for-review flow, session filtering, and spec persistence safeguards

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -15,6 +15,11 @@ import (
 
 const maxTurnsAnswer = "Agent reached maximum number of turns without a final answer."
 
+const (
+	toolResultMetaControl       = "control"
+	toolResultControlSpecReview = "spec_review_required"
+)
+
 // abortedToolResultNotice is the placeholder tool result recorded for a tool
 // call that was advertised by the assistant turn but never executed because the
 // repeated-failure guard halted the run first. It keeps every tool_use paired
@@ -214,6 +219,13 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				result.Messages = copyMessages(messages)
 				return result, abortErr
 			}
+			if stopReason := stopReasonFromToolResult(toolResult); stopReason != "" {
+				messages = appendAbortedToolResults(messages, collected.ToolCalls[index+1:])
+				result.FinalAnswer = toolResult.Output
+				result.StopReason = stopReason
+				result.Messages = copyMessages(messages)
+				return result, nil
+			}
 
 			// Repeated-failure guard: if a tool keeps failing the same way, hint
 			// once (with its schema) then halt — so no model loops on a bad call.
@@ -312,6 +324,16 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 			DenialReason: DenialFiltered,
 		}, nil
 	}
+	tool, toolFound := registry.Get(call.Name)
+	if permissionMode == PermissionModeSpecDraft && toolFound && !ToolAdvertised(tool, permissionMode) {
+		return ToolResult{
+			ToolCallID:   call.ID,
+			Name:         call.Name,
+			Status:       tools.StatusError,
+			Output:       `Error: Tool "` + call.Name + `" is not available in spec-draft mode.`,
+			DenialReason: DenialFiltered,
+		}, nil
+	}
 
 	// ask_user is intercepted in the loop (like permissions) so the question can
 	// be routed to an interactive front-end instead of blocking inside the tool.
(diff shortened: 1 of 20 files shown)


gnanam1990 (changes requested):
## Review: `Add spec mode foundation` — Changes requested

Solid foundation — I verified the core enforcement is real (not just advertised): in spec-draft mode the agent-loop gate (`loop.go:328` + `toolAdvertisedInSpecDraft`) genuinely **denies** write/shell/network/update_plan and allows only read-only + `ask_user` + the exit tool (I drove both a `write_file` and a `bash` call through `executeToolCall` in `PermissionModeSpecDraft` — both correctly returned "not available in spec-draft mode"). Session metadata back-compat is clean (all new fields `omitempty`; `kind==""` filter is a no-op for old files). A few changes before merge:

### 🔴 Please change
1. **Rename `ExitSpecMode` (competitor tool-name leakage).** `internal/specmode/exit_tool.go:13` exposes `ExitToolName = "ExitSpecMode"` as an LLM-advertised tool name — that `Exit*Mode` naming is lifted from the source product's feature, and "Spec mode is active." in the draft prompt echoes it. Per our no-competitor-naming convention, rename to a product-neutral snake_case name consistent with the rest of the tool set (e.g. `submit_spec` / `finish_spec` / `save_spec`) and reword the prompt's "Spec mode" framing. The generic spec-drafting concept is fine — just not the verbatim name.
2. **Lock the spec-file path-traversal boundary.** `specmode.SaveDraft` writes `.zero/specs/<date>-<slug>.md` where the title is **model-generated** (attacker-influenceable). `slugify` is the *sole* containment and is currently safe (it collapses everything to `[a-z0-9-]`, so `../../etc/passwd → etc-passwd`), but there's **no test** asserting that — a future edit to the allowed rune set or the empty-slug fallback would silently reintroduce traversal. Please (a) add a table test feeding adversarial titles (`../../etc/passwd`, `..`, `/abs`, `a/../../b`, all-dots) and asserting `filepath.Rel(root, path)` stays under `.zero/specs`, and (b) for defense-in-depth on model input, add an explicit containment guard in `SaveDraft` (reject if the resolved path escapes the root) rather than relying solely on the implicit slug allow-list.

### 🟡 Recommended (non-blocking)
3. **Bash-denial regression test:** `TestSpecDraftDeniesHiddenToolCalls` only drives `write_file`. Add a `bash` invocation through the loop in spec-draft mode asserting the denial + no side effect — it's the highest-risk tool and the deny path is currently unpinned for it.
4. **`RecordSpec` ordering** (`store.go:436-456`): it writes the new `SpecStatus` to metadata *before* appending the spec event, so an append failure leaves metadata ahead of the log. Append the event first (matches the canonical `appendEventLocked` pattern), then write metadata once.
5. **`applySpecRecord` transition guard** (`store.go:703-710`): `SpecStatus` is overwritten unconditionally with no legal-transition check. Latent (no callers yet), but a guard would keep the store consistent against future caller mistakes.
6. **Test gaps:** the `rejected` transition and the empty-status validation error are untested; `ask_user`/`skill` aren't enumerated in the serve-MCP present-list / `assertCoreRegistry`.

### ℹ️ Worth noting in the PR description
- This expands the **default** advertised tool set for *all* runs (`CoreReadOnlyTools` now includes `ask_user` + `skill`, and `web_fetch` via the network set) — correct and gated, but a notable surface change worth calling out.
- The interactive-TUI default flips `auto→ask` here — a justified bugfix, but it's bundled into the spec-mode PR; a line in the description (or splitting it) would help reviewers.
- Defense-in-depth: the spec-draft hide/deny rule lives only in the agent loop, not at the registry execution boundary — fine today (the loop is the only caller), but centralizing it on `RunOptions` would make it bypass-proof for future direct callers.

Nice work overall — the enforcement model and the session lifecycle are well-built; the asks are the brand rename + the traversal-boundary test/guard.

gnanam1990 (approved):

PR #146, opened 2026-06-08
Mid-run model escalation: escalate_model tool + --allow-escalation

gnanam1990:
## Summary

Lets the agent **escalate to a stronger model mid-run** by calling an `escalate_model` tool, gated behind `zero exec --allow-escalation`. The agent loop rebuilds the provider on the escalation target for the rest of the run. Strictly additive — with the flag off, no tool is registered and the model switcher is nil, so a run is byte-identical to today. No new module dependencies.

- **Registry upgrade map** (`internal/modelregistry`): `ModelEntry.UpgradeTargetID` + `Registry.UpgradeTarget(id)` — resolves each model's stronger target (e.g. `claude-haiku-4.5→claude-sonnet-4.5`, `claude-sonnet-4.5→claude-opus-4.1`, `gpt-4.1-mini→gpt-4.1`, `gemini-2.5-flash→gemini-2.5-pro`), returning it only if available and not deprecated. Top-tier models have no target.
- **Agent loop switch** (`internal/agent`): `Options.ModelSwitcher` + a typed `ToolResult.RequestedModel` (lifted from the tool's `Meta["escalate_to_model"]`). The turn loop performs the switch **at most once per turn**, after the turn's tool-results are recorded (transcript stays replay-valid); reassigns the provider and updates `options.Model` so subsequent turns, context sizing, and usage attribution follow. `nil` switcher ⇒ ignored; switcher error ⇒ transcript note + continue (non-fatal); `(nil,nil)` ⇒ no-op.
- **The tool** (`internal/tools/escalate_model.go`): `escalate_model` (`SideEffectNone`/`PermissionAllow`/`AdvertiseInAuto`), **no arbitrary-model arg** — escalates the *current* model to its `UpgradeTarget`, or returns an informational "already at the strongest available model" result with no switch.
- **Exec wiring** (`internal/cli`): `--allow-escalation` registers the tool and wires the `ModelSwitcher` (rebuilds via the resolved profile with `Model` overridden); post-switch usage events are attributed to the escalated model.

The source's proprietary auto-router half is dropped entirely; all naming is brand-neutral.

## Test Plan

- [x] `go build ./...` / `go vet ./...` clean
- [x] `go test ./...` green (new tests in `modelregistry`, `agent`, `tools`, `cli`)
- [x] `go test -race ./internal/{modelregistry,agent,tools,cli}/...` green
- [x] `GOOS=windows GOARCH=amd64 go build ./...` green

### Reviewer focus
- **Flag-off byte-identical:** without `--allow-escalation`, the tool isn't registered and the switcher is nil (mutation-tested: a usage event gains no `"model"` key, the provider builds exactly once).
- **The switch is real:** the post-switch turn streams from the rebuilt provider (mutation-tested — dropping `provider = newProvider` fails the test), and pre/post-switch usage is attributed to the original/escalated model respectively.
- **Cost:** escalation is opt-in (flag) and capped at one switch per turn; the agent can only escalate along the curated registry map (no arbitrary model).

### Known limitation (documented, deferred)
On a mid-run switch, the compaction context-window threshold isn't yet updated to the new model's window (compaction sizes against the original model's window — a conservative approximation). A follow-up can thread the new window through the `ModelSwitcher` contract.

## Summary by CodeRabbit

* **New Features**
  * Mid-run model escalation: agents can request and switch to stronger models during a run when enabled.

* **CLI**
  * New --allow-escalation flag to opt into escalation; CLI wires conditional switching and model-aware usage attribution.

* **Model Registry**
  * Registry declares and resolves upgrade targets and chains; top-tier and deprecated targets handled.

* **Tools**
  * New escalate_model control tool to request escalation; new control-only side-effect value.

* **Behavior**
  * Escalation failures are non-fatal; at most one switch per turn; no-op for top-tier/no-target models.

* **Tests**
  * Extensive tests covering escalation flows, attribution, validation, and edge cases.

### DIFF
--- a/internal/agent/escalate_loop_test.go
+++ b/internal/agent/escalate_loop_test.go
@@ -2,6 +2,8 @@ package agent
 
 import (
 	"context"
+	"errors"
+	"strings"
 	"testing"
 
 	"github.com/Gitlawb/zero/internal/tools"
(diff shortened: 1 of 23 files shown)


gnanam1990:
Both addressed.

- **`currentModel` only updates on a non-nil provider** (`exec.go`) — the switcher closure now mirrors the agent loop's `newProvider != nil` guard, so a `(nil, nil)` return leaves both the provider and `currentModel` untouched (no usage misattribution). Added `TestRunExecNilSwitchProviderKeepsOriginalAttribution`: the escalation rebuild returns `(nil, nil)`, the run continues on the original provider, and the post-decline usage stays attributed to `claude-haiku-4.5` (fails without the guard).
- **`UpgradeTargetID` is cross-validated at registry construction** (`models.go`) — a non-empty target that doesn't resolve to a known model now fails `NewRegistry` loudly, mirroring the existing deprecation-`FallbackID` check, instead of silently disabling escalation at runtime. Added `TestNewRegistryRejectsDanglingUpgradeTarget`; the seeded catalog still builds (`TestDefaultRegistryUpgradeTargets` green).

Full `build`/`vet`/`test`/`-race`/`GOOS=windows` green. @coderabbitai please re-review.

anandh8x (changes requested):
REQUEST_CHANGES

One issue to fix before merge:

- `internal/tools/types.go:22` / `internal/tools/escalate_model.go:55` add and use `SideEffectNone`, but `internal/sandbox/normalize.go:51` still treats every unknown side effect as `out_of_workspace`. That means sandbox/risk classification for `escalate_model` becomes critical/out_of_workspace via `internal/sandbox/risk.go:75-85`, even though the tool is explicitly control-only. The current default path still runs because the tool is `PermissionAllow`, but the safety metadata is wrong and any consumer relying on sandbox risk will see a false critical effect. `internal/cli/exec_writer.go:384-394` also serializes the stream-json tool call sideEffect as `unknown`, so automation does not see the promised `none`.

Please make `none` first-class in the sandbox side-effect enum/normalizer/classifier and stream-json formatter, and add a regression test that `escalate_model` reports `sideEffect: "none"` and does not classify as out_of_workspace.

Everything else in the escalation flow looks solid: opt-in flag, curated upgrade target, no arbitrary target arg, one switch per turn, non-fatal switch errors, and usage attribution are covered.

Validation:
- `GOCACHE=/tmp/zero-go-cache go test ./internal/modelregistry ./internal/tools ./internal/sandbox ./internal/agent ./internal/cli`
- `git diff --check`

gnanam1990:
@anandh8x good catch — `none` was only first-class in the `tools` package, not the sandbox/stream-json layers. Fixed in `437cca6` by making it first-class end-to-end:

- **`internal/sandbox/types.go`** — added `SideEffectNone SideEffect = "none"` to the sandbox enum.
- **`internal/sandbox/normalize.go`** — `NormalizeSideEffect` now recognizes `none` (so it's no longer collapsed to `out_of_workspace`); an unrecognized value still fails closed to `out_of_workspace`.
- **`internal/sandbox/risk.go`** — `Classify` has an explicit `SideEffectNone` case that adds no side-effect category and stays `RiskLow` (no false `critical`/`out_of_workspace`).
- **`internal/cli/exec_writer.go`** — `streamJSONSideEffect` now serializes `none` as `"none"` instead of `"unknown"`.

Regression tests: `NormalizeSideEffect(none)==none` (+ bogus still → out_of_workspace), `Classify(SideEffectNone)` is `RiskLow` and not `out_of_workspace`, and `streamJSONSideEffect("escalate_model")=="none"`. Verified these are the only three production switches that enumerate side effects.

`go test ./internal/{modelregistry,tools,sandbox,agent,cli}` + `go test ./...` + `-race` + `GOOS=windows go build` all green. Please re-review.

anandh8x (approved):
APPROVED

Re-reviewed the new head `437cca6`. The requested `SideEffectNone` plumbing is fixed:

- sandbox now defines and preserves `SideEffectNone`
- risk classification keeps control-only tools low-risk instead of `out_of_workspace`
- stream-json now reports `sideEffect: "none"` for `escalate_model`
- regression coverage was added for normalization, risk classification, and stream-json formatting

No further code-review blockers from me.

Validation:
- `GOCACHE=/tmp/zero-go-cache go test ./internal/modelregistry ./internal/tools ./internal/sandbox ./internal/agent ./internal/cli`
- `git diff --check`

GitHub smoke checks were green at re-review time; CodeRabbit was still pending.

PR #147, opened 2026-06-08
Deferred (lazy) tool loading: tool_search + auto threshold for large MCP tool sets

gnanam1990:
## Summary

Adds **deferred (lazy) tool loading** so a large MCP tool set no longer sends every tool's full JSON schema on every turn. When the number of *visible* MCP tools crosses a threshold, their schemas are withheld, advertised compactly in a per-turn `<system-reminder>`, and the model pulls a tool's full schema on demand via a `tool_search` tool. Strictly additive — below the threshold (default 10) behaviour is **byte-identical** to today. No new module dependencies.

- **Eligibility** (`internal/tools`, `internal/mcp`): MCP tools are deferred-eligible via an optional `Deferred() bool` interface (`tools.IsDeferred`); built-ins never implement it, so they stay eager. No change to the core `Tool` interface.
- **Formatting** (`internal/tools/deferred.go`): compact one-liners — `name: <short-desc> | server: <X> | inputs (* required): a (string)*, …; +N more` — wrapped in a `<system-reminder>` block by `BuildDeferredReminder`. Server label uses the MCP server name directly (`MCPServerName()`), falling back to a name parse.
- **The tool** (`internal/tools/tool_search.go`): `tool_search` (`SideEffectNone`/`PermissionAllow`/`AdvertiseInAuto`). Query `select:Name1,Name2` (exact) or keywords (ranked); returns each loaded tool's full schema and signals the loop via `Meta["load_tools"]`. It is **operator-filter-aware** — it never resolves, ranks, or lists a tool hidden by `--enabled-tools`/`--disabled-tools`.
- **Agent loop** (`internal/agent`): `partitionTools` is the single source of truth — `eligible` counts only `ToolVisible` deferred tools; when `eligible >= DeferThreshold`, unloaded deferred tools are withheld into the reminder while `tool_search` + built-ins + already-loaded tools stay exposed. `Meta["load_tools"]` is lifted to `ToolResult.LoadedTools` and unioned into a per-run loaded-set so the **next turn** advertises them. The reminder is appended only to the per-turn request copy (never persisted history), including on both reactive-compaction retry paths. `tool_search` is always reachable when deferral is active (exempt from the `EnabledTools` allowlist) so an allowlist of deferred tools can never strand the loader.
- **Config + wiring** (`internal/config`, `internal/cli`): `tools.deferThreshold` (default 10; `0` disables; negative rejected; explicit `0` preserved vs unset). `tool_search` is registered — and `DeferThreshold` threaded — in exec and the TUI, gated on the **same visible-deferred count** the loop uses, so registration and activation never disagree.

Loaded-set scope is within a single `agent.Run` (v1, fully covers headless `exec`); TUI cross-turn persistence is a documented follow-up.

## Test Plan

- [x] `go build ./...` / `go vet ./...` clean; `gofmt` clean (touched files)
- [x] `go test ./...` green (new tests across `tools`, `mcp`, `agent`, `config`, `cli`)
- [x] `go test -race ./internal/{tools,mcp,agent,config,cli}/...` green
- [x] `GOOS=windows GOARCH=amd64 go build ./...` green

### Reviewer focus
- **Byte-identical below threshold:** `partitionTools` inactive output is `reflect.DeepEqual` to the legacy `toolDefinitions` (only `tool_search` dropped, which isn't registered when inactive).
- **Gate agreement:** registration and activation count the *same* `ToolVisible` deferred population — prompt-gated MCP tools that aren't advertised in auto mode don't spuriously register `tool_search`.
- **No dead-ends:** an `--enabled-tools` allowlist of deferred tools keeps `tool_search` reachable (regression-tested); a deferred tool hidden by `--disabled-tools` is invisible to `tool_search`.
- **Reminder never persisted:** asserted absent from history on the main path and both reactive-retry paths.

### Merge note
This branch adds `tools.SideEffectNone` (for `tool_search`); PR #146 (model escalation) adds the same constant for `escalate_model` — whichever merges second will need a one-line conflict resolution (keep one declaration). `tool_search` is safe under the sandbox here regardless (its `PermissionAllow` short-circuits before any risk gate); #146 additionally makes `none` first-class in the sandbox classifier.

## Release Notes

* **New Features**
  * Added `tool_search`: a new capability to dynamically load and discover deferred tools using keyword search or exact tool selection.
  * Tools can now be loaded on-demand during conversations instead of all upfront, improving efficiency.
  * System tracks loaded tools across conversation turns.

* **Configuration**
  * New `deferThreshold` setting controls when tool deferral activates (default: 10 deferred tools).

### DIFF
--- a/internal/agent/deferred_loop_test.go
+++ b/internal/agent/deferred_loop_test.go
@@ -31,7 +31,7 @@ func TestToolResultLoadedToolsField(t *testing.T) {
 // assert executeToolCall lifts it into ToolResult.LoadedTools.
 type loadSignalTool struct{ value string }
 
-func (t loadSignalTool) Name() string       { return "load_signal" }
+func (t loadSignalTool) Name() string        { return "load_signal" }
 func (t loadSignalTool) Description() string { return "emits a load_tools signal" }
 func (t loadSignalTool) Parameters() tools.Schema {
 	return tools.Schema{Type: "object", AdditionalProperties: false}
@@ -97,7 +97,7 @@ type fakeDeferredTool struct {
 	desc string
 }
 
-func (t fakeDeferredTool) Name() string       { return t.name }
+func (t fakeDeferredTool) Name() string        { return t.name }
 func (t fakeDeferredTool) Description() string { return t.desc }
 func (t fakeDeferredTool) Parameters() tools.Schema {
 	return tools.Schema{Type: "object", AdditionalProperties: false}
@@ -114,7 +114,7 @@ func (t fakeDeferredTool) Deferred() bool { return true }
 // builtin) so the inactive path can assert it is dropped.
 type fakeToolSearchTool struct{}
 
-func (fakeToolSearchTool) Name() string       { return "tool_search" }
+func (fakeToolSearchTool) Name() string        { return "tool_search" }
 func (fakeToolSearchTool) Description() string { return "load deferred tool schemas" }
 func (fakeToolSearchTool) Parameters() tools.Schema {
 	return tools.Schema{Type: "object", AdditionalProperties: false}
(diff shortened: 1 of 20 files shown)


gnanam1990:
Superseded by #148 — recreated on top of current main (resolves the SideEffectNone conflict with the now-merged #146). Closing this one.

PR #151, opened 2026-06-09
Add provider health diagnostics

Vasanthdev2004:
## Summary
- add a providerhealth package for config/runtime/auth checks plus bounded non-generating connectivity probes
- wire `zero providers check <name> --connectivity [--json]` to emit structured health diagnostics
- wire `zero doctor --connectivity` to use the real provider health probe and surface concrete failures

## Tests
- `go test ./internal/providerhealth ./internal/cli ./internal/doctor`
- `go test ./...`

## Summary by CodeRabbit

* **New Features**
  * Added a --connectivity mode to probe provider endpoint connectivity during provider checks and diagnostics
  * Connectivity probe results are surfaced in doctor/diagnostics output

* **Improvements**
  * JSON and human-readable health output now include categorized checks and an overall status label (ok/warn/fail)
  * Commands return provider-specific exit status on connectivity failures
  * Credentials and secrets are redacted from reported messages and details

* **Tests**
  * Expanded test coverage for connectivity probing, classification, redaction, and failure behaviors

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Gitlawb/zero/internal/mcp"
 	"github.com/Gitlawb/zero/internal/observability"
 	"github.com/Gitlawb/zero/internal/plugins"
+	"github.com/Gitlawb/zero/internal/providerhealth"
 	"github.com/Gitlawb/zero/internal/providers"
 	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/selfverify"
@@ -38,6 +39,7 @@ type appDeps struct {
 	resolveConfig        func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
 	resolveMCPConfig     func(workspaceRoot string) (config.MCPConfig, error)
 	newProvider          func(config.ProviderProfile) (zeroruntime.Provider, error)
+	probeProviderHealth  func(context.Context, providerhealth.Options) providerhealth.Result
 	newSessionStore      func() *sessions.Store
 	loadPlugins          func(plugins.LoadOptions) (plugins.LoadResult, error)
 	loadHooks            func(hooks.LoadOptions) (hooks.LoadResult, error)
@@ -97,6 +99,7 @@ func defaultAppDeps() appDeps {
 		newProvider: func(profile config.ProviderProfile) (zeroruntime.Provider, error) {
 			return providers.New(profile, providers.Options{UserAgent: userAgent()})
 		},
+		probeProviderHealth: providerhealth.Probe,
 		newSessionStore: func() *sessions.Store {
 			return sessions.NewStore(sessions.StoreOptions{})
 		},
@@ -239,6 +242,9 @@ func fillAppDeps(deps appDeps) appDeps {
 	}
 	if deps.newProvider == nil {
 		deps.newProvider = defaults.newProvider
+	}
+	if deps.probeProviderHealth == nil {
+		deps.probeProviderHealth = defaults.probeProviderHealth
 	}
 	if deps.newSessionStore == nil {
 		deps.newSessionStore = defaults.newSessionStore
--- a/internal/cli/command_center.go
+++ b/internal/cli/command_center.go
@@ -434,6 +434,7 @@ Inspects resolved provider profiles and provider catalog descriptors without pri
 
 Flags:
       --json                    Print JSON summary
+      --connectivity            Probe provider endpoint for providers check
       --transport <transport>   Filter catalog descriptors by transport
 
 Add flags:
--- a/internal/cli/command_center_test.go
+++ b/internal/cli/command_center_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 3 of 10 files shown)


gnanam1990 (changes requested):
**Verdict: Request changes** — solid, useful feature, but the connectivity probe has 3 High security issues (2 SSRF + 1 credential-redaction gap). I traced each against the code; details + file:line below. The good news: the codebase already has the exact SSRF mitigation in `internal/tools/web_fetch.go` to reuse.

## High (please fix before merge)

**H1 — SSRF: the probe GETs the configured baseURL verbatim, no host/IP deny-list.**
`internal/providerhealth/providerhealth.go:217` does `client.Do(request)` against `resolvedBaseURL(profile,kind)+healthPath` (`:249-283`), and `resolvedBaseURL` for `openai-compatible`/`anthropic-compatible` only runs `providerio.NormalizeBaseURL` → `url.ParseRequestURI` (no host/IP/scheme restriction). The provider `baseURL` is **project-config controlled** (`.zero/config.json` is read from the workspace root, which an agentic CLI treats as untrusted), and a credential-less `openai-compatible` provider bypasses the only existing baseURL guard (`validateProjectProviderMerge` fires only when credential material is inherited). So a malicious repo can ship `{"providers":[{"name":"x","provider_kind":"openai-compatible","baseURL":"http://169.254.169.254","model":"x"}]}`; `zero providers check x --connectivity` / `zero doctor --connectivity` then GETs the cloud-metadata / `127.0.0.1` / `10.x` / NAT64 `[64:ff9b::a9fe:a9fe]` endpoint. It's blind-ish exfil too: status code + up to 64KB of the internal response body are surfaced into the check Message/Details (`:225,227-230,242-246`) → terminal/JSON.

**H2 — SSRF + credential capture via redirects.** Both call sites (`cli/observability.go:55-60`, `cli/provider_setup.go:98-102`) leave `Options.HTTPClient` nil, so the probe falls back to `http.DefaultClient` (`providerhealth.go:213-216`) — which follows up to 10 redirects with no per-hop re-validation. So even an *official* endpoint can `302` → an internal host, and because `applyAuth` (`:262,296-328`) attaches the real `Authorization`/`x-api-key`/`x-goog-api-key`, Go forwards those headers on a same-registered-host redirect → a malicious/compromised endpoint can capture the key. (The 5s context timeout does bound total time, so no hang DoS — just the redirect SSRF + credential forwarding.)

**Fix for H1+H2 (reuse what's already here):** `internal/tools/web_fetch.go` already has `webFetchBlockedAddrPrefixes` (0.0.0.0/8, 10/8, 100.64/10, 127/8, 169.254/16, 172.16/12, 192.168/16, ::1, fc00::/7, fe80::/10, multicast, NAT64 `64:ff9b::/96`, 6to4 `2002::/16`) + a `metadata.google.internal` host block + a safe dial context + a redirect policy. Build a dedicated `http.Client` for the probe that: (a) restricts scheme to http/https; (b) uses a `Transport.DialContext` that re-checks the *dialed IP* against the deny-list (defeats DNS rebinding); (c) sets `CheckRedirect` to re-validate every hop (or just disable redirects — a health probe doesn't need them); (d) sets an explicit `Timeout`. Don't use the shared `http.DefaultClient`.

**H3 — API key embedded in a baseURL query param leaks unredacted.** `redactBaseURL` (`providerhealth.go:398-409`) only strips URL *userinfo* (`parsed.User = nil`); a key in a query parameter (`?key=...` / `?api_key=...`) survives. It then flows to `redact()` (`:411-416`), which only removes the *configured secret values* + known token *patterns* (sk-, AIza, JWT, …) — a non-standard/custom key shape in a query param reaches the Message/Details/JSON and the wrapped `*url.Error` strings (`:229,245,346`) unredacted. Strip/redact query parameters (and the whole URL on transport errors), not just userinfo.

## Medium

**M1 — 503 miscategorized as rate-limit/warn → exit 0.** `providerhealth.go:238` groups `503 Service Unavailable` with 429/529 under `CategoryRateLimit`+`StatusWarn`. A 503 means the endpoint is *down*, not throttled, and `StatusWarn` exits 0 — so a hard outage reads as "ok-ish". Map 503 to a failure category/status (or at least separate it from rate-limit).

## Low / Nit (worth addressing)

- **status masking:** `provider_setup.go:129` (and the JSON path) print `status: ok` when the connectivity result is `StatusWarn`, contradicting the `connectivity: warn` line — surface warn in the top-level status.
- **scheme allow-list (defense-in-depth):** `resolvedBaseURL`/`NormalizeBaseURL` don't constrain scheme; today Go's transport rejects non-http(s), but validate explicitly rather than relying on that.
- **error classification:** `context.Canceled` is mapped to `CategoryTimeout` (`:332`) — it's an external cancel, not a timeout.
- **no signal→context bridge:** both call sites pass `context.Background()` (`observability.go:55`, `provider_setup.go:98`); Ctrl+C won't abort an in-flight probe (the 5s timeout still bounds it).
- **body read error swallowed:** `:225` `body, _ := io.ReadAll(...)` can operate on a truncated body silently.
- **test gaps:** the text-mode (`--connectivity` non-JSON) output + exit codes are untested (`command_center_test.go`); and `TestRunDoctorConnectivityProbesProvider` (`observability_test.go:75`) hits a real httptest server instead of injecting the `probeProviderHealth` seam — add SSRF/redaction regression tests once the guard lands.
- **help text:** the `--connectivity` description (`command_center.go:437`) only mentions `providers check`, though `doctor --connectivity` is an equal entry point.

The H1/H2 SSRF guard is the main blocker — reusing the `web_fetch.go` machinery should make it a contained change. Happy to re-review once that's in.

gnanam1990 (commented):
Reviewed the full diff (built + ran `go test ./internal/providerhealth ./internal/doctor ./internal/cli` — all green). This is a clean, well-factored package — nice work. **Verdict: approve with two minor nits.**

### What's great
- **Redaction discipline is excellent.** Every message, detail value, and URL goes through `redact()`/`redactAny()`/`redactBaseURL` with the profile's secrets, and base-URL userinfo is stripped. The tests prove secrets don't leak in messages or detail URLs.
- **The probe is safe and bounded:** context timeout, `io.LimitReader(…, 64*1024)`, and a **non-generating `GET /models`** (no token cost), with an injectable `HTTPClient` + `probeProviderHealth` dep for testability.
- **Error categorization is thorough** — auth / rate_limit / network / timeout / provider / unsupported, with sensible status mapping (rate-limit → warn, others → fail) and `classifyTransportError` correctly unwrapping `url.Error`/`net.Error`/context deadlines.
- Strong scenario coverage in the tests (config-only, auth error, timeout, rate-limit, network, unsupported transport, redaction).

### Minor
1. **Dead alias — `providerhealth.go:42`:** `CategoryProviderError = CategoryProvider` isn't referenced anywhere in the tree, and the untyped alias breaks the const block's gofmt alignment. Suggest removing it (or, if kept for API symmetry, type it `Category`).
2. **`PrimaryCheck` precedence — `providerhealth.go:85`:** it returns the connectivity check *regardless of its status*, before scanning for fail/warn. Today the only `warn` is rate-limit on connectivity itself, so it's fine — but if a non-connectivity `warn` is ever added, a passing connectivity check would mask it. Consider fail > warn across all checks first, then fall back to connectivity.

### Optional suggestion
- `connectivityCheck` will send the API key to `profile.BaseURL` even when it's `http://` (cleartext). That's the same exposure as a real provider call (not a regression), but a *health* command is a natural place to `warn` when credentials would cross a non-TLS endpoint.

Nothing blocking — the two nits are quick. Thanks for the careful redaction work.

gnanam1990 (approved):
Re-reviewed after your fixes (rebuilt + ran `go test ./internal/providerhealth` — green). Both earlier nits are resolved and the new hardening is excellent. **LGTM — approving**, with one optional follow-up.

### Resolved ✅
- Dead `CategoryProviderError` alias removed + const block realigned.
- `PrimaryCheck` now prefers **fail → warn** before falling back to connectivity.

### Great addition — SSRF blocklist
`validateEndpoint` + `blockedAddrReason` is thorough: IPv4/IPv6 special-use ranges, NAT64 / IPv4-mapped unwrapping via `embeddedIPv4Prefixes`, the `localhost`/`.localhost` check, and a literal-IP fast path. Solid defense for a probe that carries the API key to a user-configured base URL.

### Optional follow-up (non-blocking) — DNS-rebinding TOCTOU
`validateEndpoint` resolves the host and checks the addresses, but the request is then sent via `http.DefaultClient` (the CLI path sets neither `Options.HTTPClient` nor `Resolver`), which resolves **again** at dial time — so a host can pass validation on one lookup and connect to a blocked IP on the next. To make the blocklist airtight, pin the connection to the validated address: a custom `net.Dialer` with a `Control`/`DialContext` that re-checks the actual dialed `addr` against `blockedAddrReason`, wired into the client's `Transport`. Low severity here (the base URL is the user's own config, and a 2xx body isn't surfaced — so a reachability oracle at most), but worth closing since the protection is clearly meant to be robust.

### Minor
`503 Service Unavailable` was dropped from the rate-limit/warn bucket (now only 429/529 warn → 503 falls to provider-error/fail). Intentional? It's often a transient provider hiccup that a `warn` fits better than a hard `fail`.

Thanks for the careful hardening pass — nice work.

PR #150, opened 2026-06-09
zero repo-info: local (network-free) repository characterizer

gnanam1990:
## Summary

Adds `zero repo-info [--json] [--cwd <dir>]` — a repository characterizer that reads **local git only** (no network). Dep-free, additive (a brand-new command + package; nothing else changes).

Reports: file / directory / max-depth / estimated-LOC counts; per-language breakdown (LOC + file counts) with primary language; workspace type + package count; build / test / CI tooling; and bounded git-history metrics (branch, remote URL, age, contributors-90d, commit-velocity-30d).

This is the network-free residue of the upstream telemetry module — the OTEL/metrics-egress half is intentionally **not** ported.

## Design

- **`internal/repoinfo`** (new, stdlib-only): `Collect(ctx, Options{Cwd, Now, RunGit}) (Info, error)` behind an injectable `RunGit` (default shells local `git`). Streams `git ls-tree -r -l -z HEAD`, plus `rev-parse` / `remote get-url origin` / `log` / `rev-list`. `languages.go` (extension→language map) and `detect.go` (build/test/CI + workspace tables) are pure helpers.
- **`internal/cli/repoinfo.go`**: `runRepoInfo` formats text or `--json`; dispatch `case "repo-info"` in `app.go` + help line.

## Guarantees

- **Network-free:** only the read-only local subcommands `ls-tree`, `rev-parse`, `remote get-url`, `log`, `rev-list` are ever invoked; manifest reads are local `os.ReadFile`. Enforced by a test that records every subcommand against an allowlist.
- **Robust parsing:** `-z` (NUL-terminated, unquoted paths) so non-ASCII/special filenames aren't dropped; gitlinks (size `-`) skipped; `DirectoryCount` counts passthrough directories (expands each file's ancestors), verified against `git ls-tree -r -t`.
- **Correct age:** first-commit via `log --max-parents=0` (NOT `log --reverse -1`, which returns the latest commit); oldest root chosen for multi-root repos.
- **Fail-soft:** non-git / no-commits → friendly message (stderr, exit 1); each history metric is omitted (not fatal) on error; a 20s context bounds runtime; `--max-count` bounds history walking.
- **Defaults / scope:** focused field set (not the upstream 11-category config taxonomy); pure data/prose formats (json/yaml/toml/md) are excluded from the language ranking so "primary language" stays meaningful (markup-code like html/css is kept).

## Test Plan

- [x] `go build ./...` / `GOOS=windows GOARCH=amd64 go build ./...` clean
[...]

PR #139, opened 2026-06-08
Add context budget report

Vasanthdev2004:
## Summary
- Add a Go-native `contextreport` package that estimates static workspace context usage across system prompt, project guidelines, tools, and free budget.
- Add `zero context` with text and `--json` output for preflight context inspection.
- Expose a small agent system prompt preview helper so the report estimates the real runtime prompt path instead of a placeholder.

## Validation
- `go test ./internal/contextreport ./internal/cli`
- `go test ./...`
- `go run ./cmd/zero context`
- `go run ./cmd/zero context --json`
- `go run ./cmd/zero-release build`
- `go run ./cmd/zero-release smoke`

## Summary by CodeRabbit

* **New Features**
  * Added `zero context` CLI command to report workspace context: resolved workspace root, active model, context window, token usage breakdown (system prompt, project guidelines, tools) and tool count; supports human-readable and JSON (`--json`) output.
* **Documentation**
  * CLI help updated to include the new `context` command and its usage.

PR #157, opened 2026-06-10
Add repo intelligence map

Vasanthdev2004:
## Summary

Adds Go-native repository intelligence for Zero so the agent starts each run with a compact source map instead of only cwd/project docs.

- adds `internal/repomap` for deterministic workspace scanning, language/extension counts, important files, source-focused artifact skips, compact prompt rendering, and ranked path search
- adds `zero repo-map` with text, JSON, and `--query` modes for local inspection
- injects a capped `## Repo map` section into the agent system prompt through the existing `Options.Cwd` path

## Why

This is the next high-leverage agent-quality slice: before editing, Zero needs lightweight awareness of repo layout, important files, and likely file targets without adding embeddings, MCP, plugins, or external dependencies.

## Validation

- `gofmt -l internal\repomap\repomap.go internal\repomap\prompt.go internal\repomap\search.go internal\repomap\repomap_test.go internal\repomap\prompt_test.go internal\repomap\search_test.go internal\cli\repo_map.go internal\cli\repo_map_test.go internal\cli\app.go internal\agent\system_prompt.go internal\agent\loop_test.go`
- `go test ./...`
- `go vet ./...`
- `git diff --check` (passed; Windows printed CRLF conversion warnings only)
- manual: `go run ./cmd/zero repo-map --max-files 8 --max-depth 2`
- manual: `go run ./cmd/zero repo-map --json --max-files 2 --max-depth 2`
- manual: `go run ./cmd/zero repo-map --query "agent runtime" --max-files 5`

## Summary by CodeRabbit

* **New Features**
  * Added repo-map CLI to scan workspaces and produce deterministic repository maps; included in help output.
  * Repo map is now auto-included in AI system prompts when scanning succeeds, with size capping and clipped markers.
  * Multiple output modes: human-readable, JSON, or query-ranked matches; search supports multi-term ranking and result limits.

* **Tests**
  * Comprehensive tests for scanning, rendering, search/ranking, output formats, truncation/limit behavior, help text, and determinism.

### DIFF
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -857,6 +857,40 @@ func TestBuildSystemPromptInjectsWorkspaceContext(t *testing.T) {
 	}
 }
 
+func TestBuildSystemPromptInjectsRepoMap(t *testing.T) {
+	cwd := t.TempDir()
+	if err := os.WriteFile(filepath.Join(cwd, "go.mod"), []byte("module example.com/app\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.MkdirAll(filepath.Join(cwd, "internal", "service"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(cwd, "internal", "service", "service.go"), []byte("package service\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.MkdirAll(filepath.Join(cwd, "node_modules", "leftpad"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(cwd, "node_modules", "leftpad", "index.js"), []byte("module.exports = 1\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	prompt := buildSystemPrompt(Options{Cwd: cwd})
+	for _, want := range []string{
+		"## Repo map",
+		"Important files: go.mod",
+		"Languages:",
+		"internal/service/service.go",
+	} {
+		if !strings.Contains(prompt, want) {
+			t.Fatalf("expected repo map marker %q in prompt, got %q", want, prompt)
+		}
+	}
+	if strings.Contains(prompt, "node_modules/leftpad") {
+		t.Fatalf("repo map should omit ignored dependency dirs, got %q", prompt)
+	}
+}
+
 func TestBuildSystemPromptAllowsSpecModeOverride(t *testing.T) {
 	prompt := buildSystemPrompt(Options{SystemPrompt: specmode.DraftSystemPrompt})
 	if !strings.HasPrefix(prompt, "Specification drafting is active.") {
--- a/internal/agent/system_prompt.go
+++ b/internal/agent/system_prompt.go
@@ -104,6 +104,9 @@ func workspaceContext(cwd string) string {
 }
 
 func repoMapContext(cwd string) string {
+	// repomap.Scan is best-effort supplemental context for the prompt. If it
+	// fails, omit the repo map instead of failing the agent run; successful scans
+	// are still capped by repomap.RenderPrompt and maxRepoMapContextBytes.
 	snapshot, err := repomap.Scan(cwd, repomap.Options{
 		MaxFiles: 300,
 		MaxDepth: 5,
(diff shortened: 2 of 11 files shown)


gnanam1990 (commented):
Reviewed the full diff (built + `go test ./... && go vet ./...` green). This is a clean, well-bounded package — nice work. **Verdict: looks good; two minor fixes worth making + a couple of notes.**

### What's great
- **Scanner is deterministic and well-bounded** — `maxFiles`/`maxDepth` caps, skips VCS/dep/build dirs and binary extensions, and skips **symlinks** (both dirs and files), so it can't follow a link out of the workspace or into a cycle.
- **Prompt integration is exactly right:** `repoMapContext` is **fail-open** (scan error → omit the map, the run continues), bounded (300 files / depth 5), and the render is capped by `clampBudget` which is **byte-budget *and* UTF-8-safe** (`trimValidUTF8` walks back to a rune boundary) — avoids the classic truncation-splits-a-rune bug.
- **Paths are normalized to relative with `..`-escape guards** (`relativePromptPath`), so nothing outside the root leaks into the prompt.
- Search ranking is solid (exact → prefix → segment → substring → fuzzy, multi-term bonus, deterministic tiebreak), and the subsequence `fuzzyMatch` is correct.

### Minor — worth fixing
1. **An unreadable subdir aborts the whole scan.** In `repomap.go:77-79`, a non-nil `walkErr` is returned from the `WalkDir` callback, which stops the entire walk; `Scan` then returns the partial snapshot **plus the error**. So a single `EACCES` directory makes `zero repo-map` fail outright (and silently drops the prompt map). For a best-effort map, skip just that entry and keep going:
   ```go
   if walkErr != nil {
       truncated = true
       if entry != nil && entry.IsDir() { return filepath.SkipDir }
       return nil
   }
   ```
2. **`--max-files` doesn't bound the scan.** `repo_map.go:55` calls `repomap.Scan(root, repomap.Options{MaxDepth: options.maxDepth})` — `options.maxFiles` is only applied to the *output* list and the search limit in `buildRepoMapReport`, never to `repomap.Options.MaxFiles`. So `--max-files 8` still walks up to the 2000 default and reports `files=2000` in the summary while listing 8. Either thread it into `Scan` too, or rename/document it as an output cap — right now the same name means two different things across the two layers.

### Notes (non-blocking)
- **Three workspace-scan packages now:** `repomap` joins `repoinfo` (#150) and `contextreport` (#139), each re-implementing the directory walk + extension→language map + skip-list. Worth a follow-up to share one primitive before the three drift (they'll eventually disagree on which dirs to skip / how an extension maps).
- **Awareness only:** file *paths* for dotfiles / a `secrets/`-style dir will appear in the injected map (only a few specific dirs are skipped). Paths aren't secrets and it isn't exfiltrated, so this is fine — just flagging.

Thanks — the bounding and fail-open discipline here are exactly right.

gnanam1990 (changes requested):
Requesting changes for the two minor-but-concrete items from my review above — both quick, and better fixed before this ships as a public command + an always-on system-prompt feature:

1. **Unreadable subdir aborts the whole scan** (`repomap.go:77-79`) — return `filepath.SkipDir`/`nil` + set `truncated` instead of returning `walkErr`, so one `EACCES` dir doesn't fail `zero repo-map` / silently drop the prompt map.
2. **`--max-files` doesn't bound the scan** (`repo_map.go:55`) — it's applied to output/search-limit only, not `repomap.Options.MaxFiles`, so `--max-files 8` still walks the 2000 default and the summary reports `files=2000`. Thread it into `Scan`, or rename/document it as an output cap.

Everything else looks great — the scanner bounding, fail-open prompt injection, and the UTF-8-safe render cap are all spot on. Happy to re-review as soon as these two land. The package-consolidation note (repomap/repoinfo/contextreport) is a non-blocking follow-up.

gnanam1990 (approved):
Re-reviewed (rebuilt + `go test ./internal/repomap ./internal/cli` green). Both items are resolved cleanly — **approving.** 🎉

1. **Scan-abort fixed (and better than I suggested):** `handleWalkError` skips a subdir/file error (`SkipDir`/`nil`) and continues, while still treating a *root* error as fatal — a sharper distinction than just skipping everything. Nice touch adding regression tests for both the skip and the root-error paths.
2. **`--max-files` semantics fixed:** split into `--max-files` (files/matches *shown*) and the new `--scan-max-files` (scan bound, default 2000), both clearly documented in the help. No more `files=2000` surprise.

The package-consolidation note (repomap/repoinfo/contextreport sharing one walk + language map) remains a non-blocking follow-up whenever you're in there next. Thanks for the fast, thorough turnaround.

PR #161, opened 2026-06-10
Add provider onboarding UX

Vasanthdev2004:
## Summary
- add `zero providers use` and `zero providers setup` onboarding commands
- improve provider catalog/check output with actionable next steps without exposing secrets
- add config writer support for switching the active provider
- improve TUI `/provider` guidance for missing profiles and missing credentials

## Validation
- `gofmt -l internal\cli internal\config internal\provideronboarding internal\tui`
- `git diff --check`
- `go test ./...`
- `go vet ./...`
- manual: `go run ./cmd/zero providers setup groq --name fast --set-active`
- manual: `go run ./cmd/zero providers catalog --transport openai-compatible`
- manual: `go run ./cmd/zero providers --help`

## Summary by CodeRabbit

* **New Features**
  * CLI: added `providers use` and `providers setup` (human/JSON plans, optional --set-active) and richer `providers check` output with actionable “next” guidance.
  * TUI: provider onboarding wizard (/provider [status]) with provider/model selection, live model discovery, masked API-key entry, and contextual next actions.
  * Catalog/Models: added a remote-compatible provider (ollama-cloud), richer model metadata and live/cached discovery/merge.

* **Tests**
  * Expanded CLI, TUI, catalog, model-discovery, and guidance tests with secret-redaction and onboarding coverage.


Vasanthdev2004:
@coderabbitai approve

anandh8x (changes requested):
Finding:
[P1] Blank-key onboarding activates a provider with no API key. In `internal/tui/provider_wizard.go:293`, the wizard builds the profile and passes it directly into `m.newProvider(profile)`. When the user leaves the credential step blank, `providerWizardProfile` stores only `APIKeyEnv` (`internal/tui/provider_wizard.go:583-587`). The provider factory expects an already-resolved profile and passes only `profile.APIKey` to the runtime provider (`internal/providers/factory.go:32-33`, same pattern for Anthropic/Gemini), so the current TUI session switches to a provider with an empty key. The saved config will work after restart because config resolution fills the env value, but the wizard says it will save and use the provider now. Fix by using a runtime copy of the profile that resolves `APIKeyEnv` from the process env before calling `m.newProvider`, while still persisting the env-var reference instead of the secret, and add a test for the blank-key/env-var path.

Validation:
- `GOCACHE=/tmp/zero-go-cache-pr161 go test ./...` passed
- `GOCACHE=/tmp/zero-go-cache-pr161 go vet ./...` passed
- `git diff --check main...HEAD` passed

Coordination note: this overlaps with #163 in provider catalog/TUI setup files, so merge or rebase order still matters, but that is separate from the blocker above.

anandh8x (approved):
No blocking issues found on re-review.

The new `Resolve provider wizard env credentials` commit addresses my previous blocker: the wizard now builds a runtime-only profile with `APIKeyEnv` resolved from the process env before calling `m.newProvider`, while keeping the persisted/session profile on the env-var reference instead of storing the secret. The added regression test covers that exact blank-key/env-var path.

Validation:
- `GOCACHE=/tmp/zero-go-cache-pr161-rereview go test ./...` passed
- `GOCACHE=/tmp/zero-go-cache-pr161-rereview-vet go vet ./...` passed
- `git diff --check main...HEAD` passed

Still worth coordinating merge/rebase order with #163 because both touch provider catalog/TUI setup files, but that is not a blocker for this PR behavior.

PR #165, opened 2026-06-11
Add first-run model picker

anandh8x:
Summary:
- Add a model selection step to first-run onboarding using catalog/discovery models.
- Avoid fallback-list flicker while discovery is loading and save the chosen model.
- Add mouse-wheel support and clean up provider/model picker layout.

Tests:
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added a model selection stage to onboarding with search, filtering, loading states, and detailed model display
  * Model discovery runs asynchronously and integrates discovered models into the setup flow
  * Keyboard and mouse navigation extended for model/provider and model-list interaction

* **Bug Fixes**
  * Mouse interactions are now handled during setup (no longer ignored)
  * Finalized setup saves the explicitly chosen model ID

* **Tests**
  * Expanded onboarding tests covering discovery, selection, search, mouse/keyboard behaviors, and error redaction

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -359,7 +359,7 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.MouseMsg:
 		if m.setup.visible {
-			return m, nil
+			return m.handleSetupMouse(msg)
 		}
 		switch msg.Button {
 		case tea.MouseButtonWheelUp:
@@ -795,6 +795,8 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, nil
 	case providerModelsDiscoveredMsg:
 		return m.applyProviderModelsDiscovered(msg), nil
+	case setupModelsDiscoveredMsg:
+		return m.applySetupModelsDiscovered(msg), nil
 	case modelPickerModelsDiscoveredMsg:
 		return m.applyModelPickerModelsDiscovered(msg), nil
 	}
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (changes requested):
Verdict: **Request changes**

Found one valid blocker before merge:

- [P1] Stale model discovery can overwrite a newer same-provider setup request. `setupModelDiscoveryCmd()` only captures `providerID`, and `applySetupModelsDiscovered()` accepts any result where the current provider ID still matches. If the user goes back, changes the API key, and re-enters the model step for the same provider, the older async result can still replace the newer model list or show an old error. Redaction also uses the current key, so an old-key error may not be fully redacted.

Please add a per-request generation/request token to `setupModelsDiscoveredMsg`, increment it whenever model discovery is started or credentials/provider state changes, and ignore results whose generation does not match the current setup state.

Validation I ran locally:

```powershell
git diff --check origin/main...HEAD
go vet ./...
go test ./...
go run ./cmd/zero-release build
go run ./cmd/zero-release smoke
```

All validation passed; this is a behavioral race fix.

Vasanthdev2004 (approved):
Verdict: **Approve**

Rereviewed latest head `031b518` after the generation fix. The earlier blocker is resolved: setup model discovery now carries a per-request generation, ignores stale results, and captures request-scoped redaction secrets before async discovery runs.

No blocking findings from this pass.

Validation I ran locally:

```powershell
git diff --check origin/main...HEAD
gofmt -l internal\tui\model.go internal\tui\onboarding.go internal\tui\onboarding_test.go
go vet ./...
go test ./...
go run ./cmd/zero-release build
go run ./cmd/zero-release smoke
```

All passed. GitHub CI and CodeRabbit are green at the time of review.

PR #166, opened 2026-06-11
Add session context compaction flow

Vasanthdev2004:
## Summary
- Adds metadata-bearing agent compaction results and preserves active plan, loaded tools/skills, and project instructions across repeated compactions.
- Adds persisted session compaction events plus rehydrated replay so resume/fork prompts replace compacted-away history with a summary.
- Wires `/compact` in the TUI to perform real session compaction, using the active provider for a compact summary when available, and adds a conservative model-switch compaction guard.

## Why
Long Zero sessions already had automatic agent compaction, but manual/session replay compaction was still mostly a shell. This makes `/compact` affect the actual future context path instead of only reporting status.

## Validation
- `go test ./...`
- `go vet ./...`

## Notes
- Full session event logs remain available via `ReadEvents`; prompt replay uses `ReadRehydratedEvents`/`ReadReplayEvents` so compacted raw events are not re-sent.
- Provider-backed `/compact` records only the returned summary; no-provider sessions use a deterministic fallback summary.

## Summary by CodeRabbit

* **New Features**
  * Rich session compaction: preserved structured state (plans, tools/schemas, skills, project instructions), persistent compaction events, replay/rehydration, model/mode compaction gating, /compact UI with status, and a pluggable session compactor API.

* **Bug Fixes**
  * Validation for missing summarizers, trimmed/injected summaries, explicit no-op handling, clearer error propagation, and safer fallback when compaction payloads are malformed.

* **Tests**
  * Expanded unit/integration coverage for compaction metadata, preserved-state behavior, replay/rehydration, and TUI compaction flows.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -51,6 +51,23 @@ type CompactionOptions struct {
 	Summarize func(toSummarize []zeroruntime.Message) (string, error)
 }
 
+// CompactionResult is the metadata-bearing result returned by CompactMessages.
+type CompactionResult struct {
+	// Messages is the original conversation or the compacted replacement.
+	Messages []zeroruntime.Message
+	// RemovedCount is the number of original messages summarized away.
+	RemovedCount int
+	// PreservedCount is the number of original messages kept verbatim, including
+	// leading system messages and the preserved suffix.
+	PreservedCount int
+	// SummaryText is the trimmed text returned by the summarizer. The injected
+	// summary message also includes summaryLabel and any preserved structured
+	// state needed by later compactions.
+	SummaryText string
+	// Compacted reports whether Messages contains an injected summary.
+	Compacted bool
+}
+
 // estimateTokens is a cheap, dependency-free token estimate (~4 chars/token)
 // across message content plus tool call names/arguments. It deliberately uses no
 // real tokenizer; it only needs to be monotonic and roughly proportional so the
@@ -96,12 +113,23 @@ func compactionThreshold(contextWindow int) int {
 // Compact is pure: it performs no provider I/O. A Summarize error is returned to
 // the caller, which decides how to recover.
 func Compact(messages []zeroruntime.Message, opts CompactionOptions) ([]zeroruntime.Message, error) {
+	result, err := CompactMessages(messages, opts)
+	if err != nil {
+		return nil, err
+	}
+	return result.Messages, nil
+}
+
+// CompactMessages summarizes the oldest middle of a conversation and returns
+// both the replacement messages and UI/session-friendly metadata about what
+// changed. It uses the same compaction rules as Compact.
+func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (CompactionResult, error) {
 	preserveLast := opts.PreserveLast
 	if preserveLast <= 0 {
 		preserveLast = defaultCompactionPreserveLast
 	}
 	if opts.Summarize == nil {
-		return nil, errors.New("compaction requires a Summarize function")
+		return CompactionResult{}, errors.New("compaction requires a Summarize function")
 	}
 
 	// Leading system messages are kept verbatim at the head.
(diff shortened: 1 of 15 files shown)


gnanam1990 (changes requested):
Thanks for this — solid, well-tested feature, and the JSON-block approach to lossless preserved state is well-reasoned. Requesting changes on one correctness item that affects a core path (resume/fork); the rest are non-blocking suggestions.

## Blocking

### Resume/fork can now hard-fail where it previously couldn't
`exec_session.go` switches `PrepareExec` from `ReadEvents` → `ReadRehydratedEvents` (both the fork and resume paths). `RehydrateEvents` → `decodeCompactionPayload` returns an **error** on a malformed / empty-summary *latest* `EventCompaction` (`replay.go`), and that error propagates straight out of `PrepareExec`. So a single bad compaction event makes the **entire session unresumable and unforkable** — a failure mode `ReadEvents` never had.

`RecordCompaction` validates before writing, so it shouldn't happen on the happy path. But a torn tail (which the store already hardens against elsewhere — see b28a1e7 / ea2920a) or an event written by a different build would brick resume for that session. Please degrade gracefully: on decode failure, log and fall back to the raw events rather than failing the whole prepare. A regression test that resumes a session whose latest compaction payload is malformed would lock this in.

## Non-blocking

### Brittle cross-module string coupling (silent degradation)
- `loadedToolEntriesFromOutput` keys off the literal `"Loaded "` prefix + `"Full schemas follow"` substring of the `tool_search` result text.
- `projectInstructionBlock` keys off `"# "` + `" instructions for "` + the `<INSTRUCTIONS>`/`</INSTRUCTIONS>` tags.

These hard-code the *rendered output format* of other subsystems. If that wording/template ever changes, preservation silently stops working — and the tests won't catch it because they embed the same literals. Consider exporting these as shared constants from the producing packages (or documenting the contract) so a format change breaks at the source.

### Only `commandPrompt` is blocked during in-flight compaction
`handleSubmit` blocks new prompts while `compactInFlight`, but `/model`, `/mode`, and `/rewind` still execute. When the background `compactResultMsg` lands it overwrites `activeSession`, `sessionEvents`, and `transcript` from its snapshot — which could clobber an intervening `/rewind` or reset a model-switch note. Consider blocking session-mutating commands while compacting, or guarding the snapshot-apply against an intervening session change.

### Model-switch guard is a confusing half-feature
`requestCompactionBeforeModelSwitch` increments `compactRequests`, prints "Context compaction requested…", and **aborts the switch** — but never actually triggers compaction. The user must manually `/compact` then re-run `/model`. The message ("unchanged until compaction can run") implies something automatic will happen when nothing does. Either wire it to enqueue the actual compaction, or reword to make the manual step explicit.

### Apparently-unused duplicate API
`ReadReplayEvents` and `ReplayEvents` are exact aliases of `ReadRehydratedEvents` / `RehydrateEvents` and don't appear to be called anywhere in this PR. If there's no caller, drop them.

### Nits
- `projectInstructionBlock` shadows the builtin `close` — rename to `closeIdx`.
- `Compacted: len(events) < len(beforeEvents)+1` is a fragile "did anything change" check; the plan already knows via `plan.CompactableCount` — prefer deriving from that.
- Two token estimators now coexist (`estimateTokens` vs `estimateTranscriptTokens`); a one-line comment noting they're intentionally separate would help.

## Tests & security
Coverage is strong and adversarial (malformed earlier payload, markdown round-trip, repeated-compaction carry-forward, TUI flows). Worth adding the two cases above: a malformed *latest* compaction payload through `PrepareExec`, and a `/model` or `/rewind` issued while `compactInFlight`. No new security exposure — preserved state is verbatim content that `ReadEvents` already returned, redaction still applies at display, and the summarizer prompt asks to omit secrets.

Vasanthdev2004:
﻿@gnanam1990 fixed the blocking resume/fork issue in `b9b740b`.

What changed:
- `PrepareExec` now reads rehydrated events normally, but if compaction rehydration fails it logs the failure and falls back to raw `ReadEvents` instead of making the session unresumable/unforkable.
- Added a regression test covering both resume and fork when the latest `session_compaction` payload is malformed.

Validation:
- `go test ./internal/sessions`
- `git diff --check`
- `go test ./...`

Please re-review when you get a chance.

gnanam1990 (approved):
Re-reviewed after b9b740b. My blocking item — resume/fork hard-failing when the latest compaction payload is malformed — is resolved: `readExecContextEvents` now falls back to raw `ReadEvents` when rehydration fails, with a regression test covering both resume and fork. That's exactly the graceful degradation I asked for. CodeRabbit had already approved and CI is green, so lifting my requested changes.

A few **non-blocking** follow-ups remain (tracked in #173), none of which should hold up merge:
- String literals in `compaction_preserve.go` couple preserved-state parsing to other subsystems' rendered output (silent-degradation risk).
- `ReadReplayEvents`/`ReplayEvents` in `replay.go` are unused aliases.
- `projectInstructionBlock` shadows the builtin `close`.
- The resume-fallback `log.Printf` can smudge the TUI frame; consider a quieter sink.

Nice work on the compaction flow. Approving.

PR #167, opened 2026-06-11
Fix all critical & high deep-audit findings

gnanam1990:
## Summary

Resolves every **critical** and **high** finding from the deep audit. Each commit is scoped to a single finding and ships with a regression test; the full suite, `go vet`, and `gofmt` are clean.

- **MCP framing** — cap framed messages at 64 MiB and switch the stdio transport to newline-delimited JSON (Content-Length kept for back-compat). Closes an unbounded `make([]byte, contentLength)` allocation.
- **Provider connectivity probe** — SSRF-hardened: re-validate every redirect hop (capped at 5) and validate + pin the resolved IP at dial time, closing the DNS-rebinding TOCTOU.
- **bash tool** — kill the shell as a process group on timeout (`Setpgid` + group `SIGKILL` + `WaitDelay`) so backgrounded children can't outlive the command or hang `Run()`.
- **sandbox-exec profile** — allow the standard device nodes and temp trees for parity with the bubblewrap backend; workspace stays the only writable project location.
- **Reasoning effort** — thread `CompletionRequest.ReasoningEffort` through the agent loop and map it per provider: OpenAI `reasoning_effort`, Anthropic extended thinking (with thinking-block replay), Gemini thinking budget.
- **sessions** — tolerate a torn tail only when the final line is unterminated; a fully-flushed corrupt line still fails loudly.
- **streamjson** — anchor secret patterns so prose (`task-list`, `bearer of bad news`) is no longer over-redacted.
- **tools** — fix grep glob scope (ripgrep semantics) and apply_patch hunk-body misparse.
- **cli/cron** — honor an explicit expr over a recipe; preserve an external pause mid-run.
- **specialist** — keep the name in the session title so description-less runs resume.
- **agent** — surface response truncation (`finish_reason`) to the user.

### Partial (tracked in `docs/audit/2026-06-10-deep-audit-status.md`)
- **Hooks** — a `Dispatcher` that runs and audits configured hooks is added and tested, but it is **not yet wired** into the `beforeTool`/`afterTool` lifecycle and has no CLI `dispatch`/`edit` surface (behavioral change deferred to a design pass).
- **Cron** — the pause-clobber during a run is fixed; the concurrent-scheduler race (no flock/CAS in `internal/cron/store.go`) remains, with single-scheduler as the supported model.

## Test Plan
- [x] `go build ./...`
- [x] `go test ./...` (full suite green)
- [x] `go vet ./...`
- [x] `gofmt -l` clean

## Summary by CodeRabbit

* **New Features**
  * Extended “thinking”/reasoning support with configurable effort levels and preserved replayable reasoning across turns
  * Hook dispatch system for configurable event hooks

* **Bug Fixes**
  * Fixed MCP message framing/parsing
  * Improved truncation detection/reporting (CLI/TUI and run outputs now show truncation notices)
  * Better recovery from partial session logs
  * More robust patch parsing to avoid misinterpreting hunk content

* **Improvements**
  * Expanded sandbox write allowances for common temp/dev paths
  * Stricter secret-redaction patterns
  * Hardened process lifetime handling for executed commands
  * CLI cron argument parsing and run handling made more resilient
  * Safer provider connectivity checks and redirect handling

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -207,6 +207,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			Role:      zeroruntime.MessageRoleAssistant,
 			Content:   collected.Text,
 			ToolCalls: historySafeToolCalls(collected.ToolCalls),
+			// Preserve thinking blocks so the next turn can replay them; providers
+			// that use extended thinking reject tool conversations that drop them.
+			Reasoning: collected.ReasoningBlocks,
 		})
 
 		if len(collected.ToolCalls) == 0 {
@@ -1225,6 +1228,9 @@ func copyMessages(messages []Message) []Message {
 		copied[index] = message
 		if message.ToolCalls != nil {
 			copied[index].ToolCalls = append([]ToolCall{}, message.ToolCalls...)
+		}
+		if message.Reasoning != nil {
+			copied[index].Reasoning = append([]zeroruntime.ReasoningBlock{}, message.Reasoning...)
 		}
 		// Deep-copy image attachments (slice AND each Data byte slice) so the
 		// raw image bytes are never aliased across history/request/result copies.
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -84,8 +84,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 
 		exposed, reminder := partitionTools(registry, permissionMode, options, loaded)
 		request := zeroruntime.CompletionRequest{
-			Messages: copyMessages(messages),
-			Tools:    exposed,
+			Messages:        copyMessages(messages),
+			Tools:           exposed,
+			ReasoningEffort: options.ReasoningEffort,
 		}
 		if reminder != "" {
 			// Append to the per-turn request copy only — NEVER to persistent
@@ -120,8 +121,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				// empty-loaded partition, re-hiding every already-loaded deferred tool
 				// and dropping the reminder when deferral is active.
 				request = zeroruntime.CompletionRequest{
-					Messages: copyMessages(messages),
-					Tools:    exposed,
+					Messages:        copyMessages(messages),
+					Tools:           exposed,
+					ReasoningEffort: options.ReasoningEffort,
 				}
 				if reminder != "" {
 					// Append to the per-turn retry copy only — NEVER to persistent
@@ -159,8 +161,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				// Routing through an empty-loaded partition here would re-hide every
 				// already-loaded deferred tool and drop the reminder when deferral is active.
 				retryRequest := zeroruntime.CompletionRequest{
-					Messages: copyMessages(messages),
-					Tools:    exposed,
+					Messages:        copyMessages(messages),
+					Tools:           exposed,
+					ReasoningEffort: options.ReasoningEffort,
 				}
 				if reminder != "" {
 					// Append to the per-turn retry copy only — NEVER to persistent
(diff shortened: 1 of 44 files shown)


gnanam1990:
Addressed the review findings (head `f638ceb`). Each fix ships with regression coverage; full `go test ./...`, `go vet`, and `gofmt` are clean.

**Fixed**
- **apply_patch hunk-count bypass (Major/security)** — `parseHunkCounts` now parses only the range section between the opening and closing `@@`. A crafted heading like `@@ -1,1 +1,1 @@ +1,999999` could otherwise inflate the count, wedge the parser in hunk mode, and hide a later `--- `/`+++ ` file header from `validatePatchPaths`/`recheckPatchWriteTargets` (workspace-confinement bypass). Test: `TestApplyPatchToolRejectsHunkCountInflationHidingEscapePath`.
- **providerhealth single-address pinning (Major)** — the SSRF-hardened dial now tries every validated address in order instead of `addrs[0]`, so a dual-stack / multi-record provider isn't failed by one dead record. Extracted `dialValidatedAddrs` for coverage.
- **hooks fail-open on timeout (Major)** — a `beforeTool` hook killed by its deadline now fails **closed** (vetoes the tool); a true launch failure (missing binary) still fails open so a misconfigured hook can't wedge every tool call. Tests cover both.
- **cron not-found over-broad (Major)** — `Store.Get` returns a distinct `ErrJobNotFound`; `cron_run` only treats that as "removed during run" and still records + persists on a transient read error.
- **Anthropic open thinking buffers (Major)** — `closeOpen` finalizes thinking blocks still open at stream end (SSE ended after `thinking_delta`/`signature_delta` but before `content_block_stop`), so they survive in the `done` event's `ReasoningBlocks` for replay.
- **gemini stale comment**, **cron_test recipe-prompt assertion**, **store_test `file.Close()` checks**, **MCP EOF-terminated line test**, **bash timeout test** sped up via the `bashWaitDelay` hook.

**Declined (with reasoning)**
- *Replay the `redacted_thinking` signature too* — Anthropic's `redacted_thinking` blocks carry only an opaque `data` field; the `signature` field belongs to regular `thinking` blocks, and the stream never populates a signature for redacted blocks. Adding one would send a field the API doesn't expect for that block type, so the current `{type: redacted_thinking, data}` replay is correct.

@coderabbitai review

gnanam1990:
Pushed `7dcbf4f`.

- **`dialValidatedAddrs` last-error test (nitpick)** — fixed: the test now uses distinct per-attempt errors and asserts `errors.Is(err, lastErr)` plus the attempt count, so a regression returning the first failure is caught.

The other two open inline threads are already resolved:
- **`cron_run.go:177` — distinguish not-found from IO error** — done in `f638ceb`: `cron.Store.Get` now returns a wrapped `ErrJobNotFound` sentinel, and `cron_run` only takes the "removed during run" path for that sentinel (transient read errors warn but still record + persist). Tests: `TestStoreGetDistinguishesNotFoundFromReadError`.
- **`provider.go:442` — replay `redacted_thinking` signature** — declining (with reasoning): Anthropic's `redacted_thinking` blocks carry only an opaque `data` field; the `signature` field belongs to regular `thinking` blocks, and the stream never populates a signature for redacted blocks. Adding one would send a field the API doesn't expect for that block type, so the current `{type: redacted_thinking, data}` replay matches the contract.

@coderabbitai review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
@coderabbit approve

Vasanthdev2004 (approved):
Verdict: APPROVE. I reviewed the critical/high audit fixes around provider reasoning/finish reasons, MCP message bounds/framing, provider health safe dialing, hook dispatch timeout behavior, bash process lifetime hardening, cron/session edge cases, and the tool parsing fixes. I did not find a blocking issue. Validation: go test ./... passed locally in D:\\tmp\\zero-review-167.

PR #168, opened 2026-06-11
Polish main TUI surface

anandh8x:
Summary:
- Simplifies the empty chat surface by removing starter prompt chips and duplicate model text.
- Frames the composer with a stronger border and shows model + permission mode in the frame.
- Quiets the bottom status line and improves alignment.
- Adds terminal-style composer word editing for Alt/Ctrl word movement and deletion.
- Adds app-managed mouse selection/copy for chat transcript text; this intentionally captures mouse input while the TUI is active, so native terminal selection remains available through the terminal escape path such as Shift-drag where supported.

Tests:
- git diff --check

More UI polish can continue in follow-up PRs.

## Summary by CodeRabbit

* **New Features**
  * Word-wise cursor navigation, smarter word deletion, mention-aware backspace, transcript text selection & copy, mouse interactions for overlays.
* **UI/UX Changes**
  * Bordered composer with stronger divider showing model/mode, inline command-argument hints, simplified empty state, centered/capped suggestion & picker overlays, model-picker loading/status views, refined footer/status rendering and theme accents.
* **Bug Fixes**
  * More robust suggestion completion/dismissal, stale discovery guarding, cursor/selection edge-case fixes.
* **Tests**
  * Large expansion across autocomplete, mouse, picker, wizard, rendering, onboarding, and config tests.
* **Config**
  * Favorite models preference added with normalization and persistence.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.1-0.20250404203927-76690c660834
+	github.com/charmbracelet/x/ansi v0.11.6
 	golang.org/x/sys v0.38.0
 )
 
@@ -15,7 +16,6 @@ require (
 	github.com/atotto/clipboard v0.1.4 // indirect
 	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
 	github.com/charmbracelet/colorprofile v0.4.1 // indirect
-	github.com/charmbracelet/x/ansi v0.11.6 // indirect
 	github.com/charmbracelet/x/cellbuf v0.0.15 // indirect
 	github.com/charmbracelet/x/term v0.2.2 // indirect
 	github.com/clipperhouse/displaywidth v0.9.0 // indirect
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -391,6 +391,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		ProviderName:    resolved.Provider.Name,
 		ModelName:       resolved.Provider.Model,
 		ProviderProfile: resolved.Provider,
+		FavoriteModels:  resolved.Preferences.FavoriteModels,
 		Provider:        provider,
 		NewProvider:     deps.newProvider,
 		Registry:        registry,
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -145,7 +145,8 @@ func TestRunNoArgsLaunchesTUIWithResolvedProviderMetadata(t *testing.T) {
 					APIKey:       "sk-test",
 					Model:        "gpt-test",
 				},
-				MaxTurns: 5,
+				Preferences: config.PreferencesConfig{FavoriteModels: []string{"qwen3-coder:480b"}},
+				MaxTurns:    5,
 			}, nil
 		},
 		newProvider: func(profile config.ProviderProfile) (zeroruntime.Provider, error) {
@@ -178,6 +179,9 @@ func TestRunNoArgsLaunchesTUIWithResolvedProviderMetadata(t *testing.T) {
 	}
 	if launchedOptions.ModelName != "gpt-test" {
 		t.Fatalf("ModelName = %q, want gpt-test", launchedOptions.ModelName)
+	}
+	if len(launchedOptions.FavoriteModels) != 1 || launchedOptions.FavoriteModels[0] != "qwen3-coder:480b" {
+		t.Fatalf("FavoriteModels = %#v, want qwen3-coder:480b", launchedOptions.FavoriteModels)
 	}
 	if launchedOptions.Setup.Visible {
 		t.Fatalf("Setup.Visible = true, want false for credentialed provider")
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -103,6 +103,7 @@ func Resolve(options ResolveOptions) (ResolvedConfig, error) {
 		Sandbox:        cfg.Sandbox,
 		Notify:         cfg.Notify,
 		Tools:          cfg.Tools,
+		Preferences:    cfg.Preferences,
 	}, nil
 }
 
@@ -162,6 +163,9 @@ func mergeConfig(dst *FileConfig, src FileConfig) {
 	if src.Tools.deferThresholdSet {
 		dst.Tools.DeferThreshold = src.Tools.DeferThreshold
 		dst.Tools.deferThresholdSet = true
+	}
+	if src.Preferences.FavoriteModels != nil {
+		dst.Preferences.FavoriteModels = normalizeFavoriteModels(src.Preferences.FavoriteModels)
 	}
 }
 
--- a/internal/config/resolver_test.go
+++ b/internal/config/resolver_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"os"
 	"path/filepath"
+	"reflect"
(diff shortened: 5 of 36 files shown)


anandh8x:
@coderabbitai re review

Vasanthdev2004 (changes requested):
﻿Requesting changes for one test reliability issue.

Finding:
- `internal/tui/provider_wizard_test.go` `TestProviderWizardSupportsLeftAndGuardedRightNavigation` assumes the selected provider has no credential already available in the process environment. On a normal developer machine with a provider key set (for example `OPENAI_API_KEY`/`OLLAMA_API_KEY`), `credentialReadyForRight()` returns true, so Right advances to the model step and the test fails:
  `right from empty credential step = 2, want credential`.

This should be made hermetic. Please clear the selected provider's `AuthEnvVars` with `t.Setenv(env, "")` before the empty-credential assertion, or use a helper that clears all catalog auth env vars for this test. Keep `TestProviderWizardRightAllowsExistingCredentialEnv` as the explicit positive coverage for env-backed credentials.

Validation I ran:
- `go test ./internal/tui ./internal/config ./internal/cli` fails in my normal environment because of the test above.
- With provider API env vars cleared: `go test ./internal/tui` passes.
- With provider API env vars cleared: `go test ./...` passes.
- `git diff --check origin/main...HEAD` passes.
- `go run ./cmd/zero-release build` passes.
- `go run ./cmd/zero-release smoke` passes.

The implementation otherwise looks sound from my pass; this is a bounded fix.

gnanam1990 (commented):
Reviewed the current state (post un-draft, +4074/−485). Solid and genuinely well-tested work — flagging a few things for a team decision rather than blocking, since none are correctness bugs.

## Scope — this is now 4+ features in one ~4,000-line PR
It grew from "polish empty state/composer" to also include two-step mouse selection, OSC-52 clipboard, provider-wizard/picker mouse hit-testing, and favorite-models persistence. That's hard to review atomically, bisect, or revert. Strong suggestion: split the mouse-selection + clipboard work into its own PR on top of the polish. At minimum it must **rebase on #170**, which also rewrote `renderPermissionRow` in `rendering.go`.

## Mouse capture disables native terminal text selection
`wantsMouseCapture` enables `EnableMouseCellMotion` across the whole chat surface (`chatWantsMouseCapture` = `!setup.visible`). Once mouse is captured, the terminal's own drag-to-copy stops working — users who rely on native selection (or copying into another app) lose it, replaced by the PR's own selection. This is the defining UX tradeoff of the feature and should be a deliberate, documented decision (and worth noting the Shift-drag escape hatch most terminals honor). Several users will read this as "I can't select text in my terminal anymore."

## Hit-testing duplicates layout math (maintainability trap)
`overlayMouseHit`/`overlayMouseTop` re-derive each overlay's on-screen position by replicating the render path's centering logic, and each `select*AtMouse` hand-counts row offsets to skip headers/separators (`hit.y - 3`, `rowStart := baseRow + 2`, `+1` for error/status lines, group-header increments). The renderer computes those positions one way; the hit-tester re-counts them another, linked only by magic numbers. Any layout tweak (and this PR tweaks layouts) will silently map clicks to the wrong row. `mouse_test.go` helps, but these offsets should derive from the same constants the renderer uses rather than being re-counted by hand.

## OSC-52 write bypasses the Bubble Tea render loop
`os.Stdout.WriteString(ansi.SetSystemClipboard(text))` writes raw escapes to stdout directly from an update handler, outside tea's managed output — in alt-screen that can race/corrupt the frame; the idiomatic path is a `tea.Cmd`/`tea.Printf`. Also OSC-52 is a silent no-op on terminals that don't support it (and has a payload cap) — no fallback or user feedback when a copy doesn't land.

## Still open from the earlier pass
- `fileIndexTTL = 30 * time.Second` is unchanged — a file created mid-session is invisible to `@`-completion for up to 30s. Shorten it or invalidate the cwd entry on write-tool activity.
- Worth re-confirming the `dismissSuggestions` mid-prompt token removal and ANSI-overlay wide-rune (CJK/emoji) correctness — the overlay code now also backs mouse hit-testing, so wide-rune alignment matters more.

## Verdict
Strong, well-tested feature work; nothing here is a data-correctness bug. But it's large and tightly coupled — I'd want a rebase on #170, an explicit team decision on the mouse-capture/native-selection tradeoff, and ideally the mouse/clipboard work split out before merge. Happy to dig into any specific spot.

gnanam1990 (changes requested):
Requesting changes (see my detailed comment above for the full list). Nothing here is a correctness bug — the work is strong and well-tested — but two items should gate the merge:

1. **Rebase on #170 first.** #170 rewrote `renderPermissionRow` in `rendering.go`; this PR touches the same file and is later in the merge order, so it needs to rebase or it'll conflict/break.
2. **Decide the mouse-capture tradeoff explicitly.** Enabling mouse capture across the chat removes the terminal's native drag-to-copy; that's a product decision (with a documented Shift-drag note), not a silent default.

Everything else in the comment (splitting the mouse/clipboard work out, the hit-test layout duplication, the OSC-52 stdout write, the 30s file-index TTL) is non-blocking polish. Happy to re-approve once the rebase lands and the capture tradeoff is signed off.

gnanam1990:
Re-review after #170 merged: correcting my earlier review.

- **Withdrawing blocker #1 (rebase/rendering.go conflict).** I flagged this off a stale file list. The current head doesn't touch `rendering.go`, and the PR is `MERGEABLE` against the post-#170 main — there's no conflict to resolve. My mistake; that gate is cleared.
- **Blocker #2 still stands:** the mouse-capture change still removes native terminal text-selection across the chat, and that product tradeoff isn't acknowledged anywhere yet. `6e088c3` only made the provider-wizard nav test hermetic.

So the only thing I'd still want before approving is an explicit call on the mouse-capture/native-selection tradeoff (a line in the PR description or a code comment documenting it, ideally with the Shift-drag note). The rest — splitting mouse/clipboard out, the hit-test layout-offset duplication, the OSC-52 stdout write, the 30s file-index TTL — remains non-blocking polish. Clear that one tradeoff and I'm happy to approve.

Vasanthdev2004 (approved):
﻿Approved on rereview. The provider-wizard test is now hermetic in `6e088c3`; it passes in my normal environment with API key env vars present.

Validation:
- current PR head: `go test ./internal/tui ./internal/config ./internal/cli`
- clean merge onto latest `origin/main`: `git diff --check`
- clean merge onto latest `origin/main`: `go test ./internal/tui ./internal/config ./internal/cli`
- clean merge onto latest `origin/main`: `go test ./...`
- clean merge onto latest `origin/main`: `go run ./cmd/zero-release build`
- clean merge onto latest `origin/main`: `go run ./cmd/zero-release smoke`

GitHub checks are also green. My previous blocker is resolved; no blockers from my side.

anandh8x:
@gnanam1990 addressed the remaining mouse-capture point in the PR description.

The explicit product call is now documented there: app-managed mouse selection/copy intentionally captures mouse input while the TUI is active, and native terminal selection remains available through the terminal escape path such as Shift-drag where supported.

Also merged latest `origin/main`; all checks are green now.

anandh8x:
@coderabbitai review

The requested changes have been addressed in the latest branch updates:
- `dismissSuggestions` now removes the cursor-local `@` file query instead of the trailing token, with coverage for `compare @old with @new`.
- Provider model discovery now carries a request token and ignores stale same-provider responses, with regression coverage.
- The mouse `Type` compatibility path is documented, redundant deprecated `Type` usage was removed from tests where safe, and the width-tier test now mirrors production styling.
- The small-terminal footer overflow case is clamped and covered.

Please re-review when ready.

gnanam1990 (approved):

PR #169, opened 2026-06-11
Fix /resume flooding with sub-run sessions

gnanam1990:
## Problem

`/resume` showed a session list ending in "… 779 more · /resume <id>" — far more sessions than real conversations. Root cause: the picker called `sessions.Store.List()` **unfiltered**, so it listed every session directory, including the **child** (specialist/sub-agent) and **spec-draft/spec-impl** sessions an agent spawns by the dozen. The TUI itself creates exactly one session per conversation (`ensureActiveSession` guards on the active id and appends events), so the flood was entirely the unfiltered listing.

## Fix

- Add `sessions.IsResumableKind`, `Store.ListResumable()`, and `Store.LatestResumable()` — they keep only standalone conversations (regular + user forks) and drop `child` / `spec-draft` / `spec-impl` sub-runs.
- `/resume` picker uses `ListResumable()`; the "N more" count is now honest.
- `/resume latest` uses `LatestResumable()` so it never lands on a newer child/spec sub-run.
- `/resume <id>` is unchanged — an explicit id still resolves any session.

## Tests

- `TestListAndLatestResumableExcludeSubRuns` — store-level: only regular+fork listed; latest skips a newer child.
- `TestResumePickerExcludesSubRunSessions` — TUI picker lists the real conversation, excludes child/spec.

`go build`, full `go test ./...`, `go vet`, `gofmt` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the resume session picker to display only standalone conversations and exclude sub-runs (child sessions and spec drafts), providing a cleaner selection experience.
  * Fixed the latest resume feature to correctly identify and select the most recent resumable conversation instead of sub-runs.

### DIFF
--- a/internal/sessions/store.go
+++ b/internal/sessions/store.go
@@ -345,6 +345,47 @@ func (store *Store) Latest() (*Metadata, error) {
 	return &sessions[0], nil
 }
 
+// IsResumableKind reports whether a session kind represents a standalone,
+// user-resumable conversation rather than an agent sub-run. Regular ("") and
+// user fork sessions are resumable; child (specialist/sub-agent) and spec
+// draft/impl sessions are not — each agent task and /spec run creates one, so
+// listing them in the resume picker floods it with non-conversation entries.
+func IsResumableKind(kind SessionKind) bool {
+	switch kind {
+	case "", SessionKindFork:
+		return true
+	default:
+		return false
+	}
+}
+
+// ListResumable returns only the sessions a user can resume as standalone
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
Approved. Cross-checked this against the D session flow and the direction matches: normal resume/latest should only surface user-resumable top-level conversations, while explicit resume by id still works. Local validation passed with go test ./internal/sessions ./internal/tui and go test ./.... No blockers from my side.

PR #170, opened 2026-06-11
permissions: scope-enforced grants (permission-UX Phase 0 + 1)

gnanam1990:
## What

Make an "always allow" permission grant cover **exactly** what the permission card shows — the file or directory the call touches — instead of silently authorizing the whole tool.

This bundles the full permission-UX work into one PR:

- **Phase 0 (`742ff48`):** derive and *display* a call's scope (file / dir / cwd) on the permission card and the persisted decision row. Presentation only.
- **Spec (`155e6a7`):** design doc at `docs/superpowers/specs/2026-06-11-permission-scope-grants-design.md`.
- **Phase 1 (`16fcbe8`):** *persist and enforce* the scope.

## Why

Before this, clicking "always allow" on a write to `src/main.go` stored a **tool-wide** grant — every future write was auto-approved, the blind-yes the Phase-0 card was written to warn against.

## How (Phase 1)

- `internal/sandbox/scope.go` centralizes scope handling:
  - `DeriveScope` — single source of truth for which args are path-like (`path`/`file` → file, `directory`/`dir`/`cwd` → dir).
  - `resolveScopeAbs` — normalizes to an absolute path anchored to the workspace root, so a grant in `/proj-a` never matches `/proj-b` and `./x` == `x` == `/proj/x`.
  - `grantCovers` — file = exact path, dir = subtree, empty = tool-wide; a tool-wide *request* is never covered by a narrower grant (re-prompts, fail-safe).
- Grants are stored per tool as a list (`map[tool][]Grant`), schema bumped **v1 → v2**; existing v1 files are migrated as tool-wide grants, so current grants keep working.
- `Lookup` takes the request's absolute scope and applies **deny-wins / most-specific-allow** precedence.
- `engine.Grant` anchors a relative scope to the workspace; `engine.Decide` derives + matches the request scope.
- `agent.persistPermissionGrant` forwards the scope; `permissionScope` (display) now shares `DeriveScope` so the card and the stored grant can't diverge.
- Canonicalizing tool keys on read also closes the whitespace-padded-key lookup miss (audit finding #90).

## Testing

- `internal/sandbox`: `DeriveScope` kinds; `resolveScopeAbs` (relative-anchored, `./` equivalence, absolute passthrough, cross-project non-match); `grantCovers` matrix; deny-precedence; most-specific-allow; per-grant + policy autonomy; v1→v2 migration; whitespace-key canonicalization; `FormatGrantList` scope rendering; engine end-to-end scoped allow/sibling-reprompt/dir-deny.
- `internal/agent` + `internal/tui`: always-allow now persists a file-scoped grant; a sibling file is not covered.
- `go test ./...`, `go vet ./...`, `gofmt -l` all clean.

## Compatibility

The grants file moves to schema v2 (older binaries reading a v2 file will report an unsupported schema). Existing v1 grants are read and migrated forward as tool-wide.

## Out of scope

Card UI changes (Phase 0), per-scope revoke, glob/pattern scopes.

## Summary by CodeRabbit

* **New Features**
  * "Always allow" grants are now scoped (file/dir/tool-wide); permission prompts, events, and lists display scope (long scopes truncated). Revoke removes all grants for a tool. Deny takes precedence; the most-specific covering allow is chosen.

* **Documentation**
  * Added Permission‑UX Phase 1 spec covering scope semantics, path normalization, per-tool multi-grant storage, migration from v1, and CLI behavior.

* **Tests**
  * Added/expanded tests for scope derivation, truncation, matching, persistence, migration, and grant evaluation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ zero exec -o stream-json < turns.jsonl        # programmatic, for scripts & CI
 - 📈 **Mid-run model escalation** — start cheap, and let the agent request a stronger model only when it hits a wall (`--allow-escalation`).
 - 🗺️ **Repo intelligence** — deterministic repo maps, workspace indexing, and context-budget reports keep the agent grounded in *your* codebase, not hallucinations.
 - ⏰ **Scheduled agents** — `zero cron` runs file-backed, dependency-free agent jobs on a schedule.
-- 🛡️ **Safe by default** — permission-gated mutations, autonomy ceilings, sandbox policy, and secret redaction everywhere. Unsafe mode is an explicit, loudly-labeled opt-in.
+- 🛡️ **Safe by default** — permission-gated mutations, autonomy ceilings, sandbox policy (writes stay inside the workspace unless you grant extra directories with `--add-dir` / `/add-dir`), and secret redaction everywhere. Unsafe mode is an explicit, loudly-labeled opt-in.
 - 💾 **Durable sessions** — append-only local event store with full-text search, resume, fork, and rewind. Your history never leaves your disk.
 - 🧩 **Extensible** — skills, plugins, hooks, and MCP (Zero is both an MCP client *and* an MCP server).
 
@@ -84,6 +84,7 @@ Type to chat, **Enter** to send. `/` opens command suggestions, **Shift+Tab** cy
 | `/resume` `/rewind` | time-travel across sessions |
 | `/compact` `/context` | manage the context window |
 | `/permissions` `/tools` | inspect what the agent can touch |
+| `/add-dir` | grant an extra write directory for the session, or list current write roots |
 | `/theme` `/style` | make it yours |
 | `/doctor` `/usage` `/config` | health, cost, and config without leaving the chat |
 
@@ -112,7 +113,7 @@ zero exec --resume            # latest
 zero exec --fork <session-id> "now try the other approach"
 ```
 
-Key flags: `-m/--model` · `--mode <smart|deep|fast|large|precise>` · `--image` · `--use-spec` · `--auto <low|medium|high>` · `--enabled-tools/--disabled-tools` · `-w/--worktree` · `--resume/--fork` · `--allow-escalation` · `--notify` · `-o <text|json|stream-json>`.
+Key flags: `-m/--model` · `--mode <smart|deep|fast|large|precise>` · `--image` · `--use-spec` · `--auto <low|medium|high>` · `--enabled-tools/--disabled-tools` · `-w/--worktree` · `--add-dir <path>` (repeatable) · `--resume/--fork` · `--allow-escalation` · `--notify` · `-o <text|json|stream-json>`.
 
 stdout carries **only** program output; logs go to stderr. Full contract in [`docs/STREAM_JSON_PROTOCOL.md`](docs/STREAM_JSON_PROTOCOL.md).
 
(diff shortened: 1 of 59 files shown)


Vasanthdev2004 (changes requested):
﻿Requesting changes. Validation is green locally (`git diff --check origin/main...HEAD`, targeted package tests, and `go test ./...`), but a few current-head issues should be fixed before merge:

1. `internal/sandbox/grant_scope.go`: `DeriveScope` treats `.` as workspace-root/tool-wide, but `./`, `./.`, and similar root-equivalent paths still become directory-scoped grants. Please `filepath.Clean` the trimmed value before the root check and add a regression test, otherwise an "always allow" on cwd `./` gets different scope semantics from cwd `.`.

2. `internal/tui/rendering.go`: the allow branch in `renderPermissionRow` appends raw scope text without the `scope:` label and returns without `fitStyledLine(width)`. This is inconsistent with the deny branch and can overflow on narrow terminals. Prefix it as `scope:<value>` and fit the final line before returning.

3. `internal/sandbox/grant_scope_test.go`: `TestGrantReplacesSameScope` ignores errors from the setup `store.Grant` calls. Please assert those errors so the test fails at the real setup failure instead of continuing with stale state.

The feature direction is right; these are bounded fixes.

gnanam1990:
Addressed the review feedback in 4e63f35:

1. `DeriveScope` (`grant_scope.go`): `filepath.Clean` the path before the root check so `.`, `./`, `./.`, and `a/..` all collapse to the workspace root and surface as tool-wide rather than a narrower directory grant. Added regression cases to `TestDeriveScope`.
2. `renderPermissionRow` (`rendering.go`): scope segment is now prefixed `scope:<value>` across all three branches, and the allow branch is wrapped in `fitStyledLine(width)` so scoped rows can't overflow narrow terminals.
3. `TestGrantReplacesSameScope`: now asserts the setup `store.Grant` errors so a setup failure fails at the real cause.

`go build`, `go vet`, `gofmt`, and the full `sandbox`/`tui` suites pass. Ready for another look.

@coderabbitai review

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
Addressed the remaining review thread in 0082e03: the design spec now points at `internal/sandbox/grant_scope.go` (was `scope.go`, the unrelated write-roots type) for `DeriveScope`/`resolveScopeAbs`/`grantCovers`, and the test reference is corrected to `grant_scope_test`. Doc-only change. @coderabbitai review

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
@coderabbit approve

gnanam1990:
@coderabbit approve

Vasanthdev2004 (approved):
﻿Approved. Re-reviewed current head `0082e03` after the permission-scope fixes.

Verified the previous requested-change items:
- root-equivalent scopes (`.`, `./`, `./.`, `a/..`) now collapse to tool-wide via `filepath.Clean` and have regression coverage.
- permission allow/deny/prompt rows consistently label `scope:<value>` and the allow row is width-fitted.
- grant setup errors in `TestGrantReplacesSameScope` are asserted.
- spec now points at `grant_scope.go` / `grant_scope_test`, matching the actual source files.

Validation passed locally:
- `go test ./...`

No blockers from my side.

PR #174, opened 2026-06-11
audit: fix isolated remaining findings (redaction, secrets, rune-safety, credential status)

gnanam1990:
Addresses a batch of the **isolated, non-conflicting** open findings from the deep-audit ledger (`docs/audit/2026-06-10-deep-audit-status.md`). Each is real, verified, and lives in a file *not* being rewritten by the in-flight PRs (#166/#170/#172) — see "Deliberately deferred" for why this isn't all 91.

## Fixed (6 findings)

| Audit location | Fix |
|---|---|
| `redaction.go:222` | `seen` set now tracks only the current DFS path (delete pointer after recursing), so a shared reference reached via a **sibling** branch is no longer reported as `CircularReference`. True cycles still detected. |
| `redaction.go:362` | `redactURLPasswords` compiles its regexp once at package scope instead of per call. |
| `scanner.go:35` | `openai_key` body allows `-`/`_` so modern `sk-proj-…` / `sk-svcacct-…` keys match, not just legacy `sk-<alnum>`. |
| `cron_run.go:179` | `cronTruncate` cuts on a UTF-8 rune boundary (reuses `cutRuneBoundary`) so a persisted run-error excerpt can't end in a split rune. |
| `zerogit.go:207` | `ValidateMessage` counts runes, not bytes — a valid non-ASCII commit subject under 72 chars is no longer rejected. |
| `contracts.go:166` | `ProviderSnapshot.APIKeySet` is true for an auth-header-only profile (was APIKey-only), so it no longer renders as "not set". |

New regression tests: redaction sibling-DAG + true-cycle, modern OpenAI key redaction, `ValidateMessage` rune count.

## Verification
`go build ./...`, `go vet`, `gofmt`, and the changed packages' tests all pass; full-suite run leaked 0 sessions into `$HOME`.

> Note: a full `go test ./...` showed one **flaky, unrelated** failure — `internal/config TestLoadProviderCommandDoesNotResolveAPIKeyEnvFromProcess` timed out at its 5s subprocess limit under full-suite parallelism, then passed 2/2 in isolation. Not touched by this PR; worth a separate fix to raise/relax that timeout.

## Deliberately deferred (NOT in this PR — by design)

The ledger has ~74 open + 17 partial. The rest are excluded for concrete reasons:

1. **Conflicts with in-flight PRs** — findings in `agent/loop.go`, `agent/types.go`, `sandbox/grants.go`, `sandbox/engine.go`, `agent/compaction*.go`, `tui/rendering.go`, `tui/session_controls.go`, `tui/model.go`, `tools/{write,edit,read,apply_patch}.go` are being rewritten by #166/#170/#172. Fixing them now guarantees merge conflicts — they should be done **after** those merge.
2. **Needs its own focused PR + tests** — specialist `--auto high` autonomy clamp (gap #6, the program's top safety item), MCP client/network bugs (ping handling, blocking write under mutex, SSE first-message, scanner reconnect), cron DST fall-back + reconcileOverdue, `process_posix` recycled-pid race, `safe_command` quote-aware split, sessions Fork usage double-count / fsync, specialist accounting race.
3. **Behavioral blast radius** — `worktrees defaultRunGit` stdout/stderr split changes git-output parsing for every caller; needs a caller audit.
4. **Dead-code judgment calls** — e.g. `redaction StackTrace` interface is defensive (keep), provider `isImplicitOpenAI` dead branch touches provider resolution; these need per-item decisions, not a blanket sweep.

These are tracked in the v0.1 release program PRD workstreams (WS-A/B/E backlogs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduce false-positive circular-reference redactions so shared non-cyclic references aren’t misclassified.
  * Ensure sensitive values embedded in URLs or provider snapshots are redacted and not leaked.

* **Improvements**
  * Truncation and commit-subject validation now respect UTF‑8 rune boundaries for correct multi-byte handling.
  * Profiles authenticated via auth headers are treated as having configured credentials.
  * Secret detection updated to match modern OpenAI key formats and uses a precompiled pattern.

* **Tests**
  * Added tests for cycle detection, auth-header credential handling, model-discovery auth, and modern OpenAI key detection.

### DIFF
--- a/internal/cli/cron_run.go
+++ b/internal/cli/cron_run.go
@@ -199,5 +199,7 @@ func cronTruncate(s string, max int) string {
 	if len(s) <= max {
 		return s
 	}
-	return s[:max] + "…"
+	// Cut on a UTF-8 rune boundary so a persisted run-error excerpt can't end in
+	// a split multi-byte rune (invalid UTF-8 in the cron record).
+	return cutRuneBoundary(s, max) + "…"
 }
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -129,8 +129,10 @@ func providerConfigCheck(profile config.ProviderProfile) Check {
 	// Report key PRESENCE, never the value: the raw key would only be redacted
 	// downstream anyway, and an unconditional [REDACTED] read as "configured"
 	// even when no key was set at all.
+	// A profile can authenticate via a raw auth-header value instead of APIKey, so
+	// treat either as a configured credential (matches ProviderSnapshot.APIKeySet).
 	apiKey := "not set"
-	if profile.APIKey != "" {
+	if profile.APIKey != "" || profile.AuthHeaderValue != "" {
 		apiKey = "set"
 	}
 	return check("provider.config", "Provider config", StatusPass, fmt.Sprintf("Provider config loaded for %s.", providerName(profile)), map[string]any{
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -126,21 +126,22 @@ func providerConfigCheck(profile config.ProviderProfile) Check {
 	if emptyProviderProfile(profile) {
 		return check("provider.config", "Provider config", StatusFail, "No LLM provider is configured.", map[string]any{"help": "Set a provider in config or environment."})
 	}
-	// Report key PRESENCE, never the value: the raw key would only be redacted
-	// downstream anyway, and an unconditional [REDACTED] read as "configured"
-	// even when no key was set at all.
-	// A profile can authenticate via a raw auth-header value instead of APIKey, so
-	// treat either as a configured credential (matches ProviderSnapshot.APIKeySet).
-	apiKey := "not set"
-	if profile.APIKey != "" || profile.AuthHeaderValue != "" {
-		apiKey = "set"
+	// Report credential PRESENCE, never the value. Reported under a non-sensitive
+	// key ("credentialConfigured"): the prior "apiKey" key was itself sensitive, so
+	// check()'s redaction scrubbed the indicator to [REDACTED] — making "set"/"not
+	// set" invisible. A profile can authenticate via a raw auth-header value instead
+	// of APIKey, and both are trimmed so a whitespace-only value reads "not set"
+	// (matching ProviderSnapshot.APIKeySet).
+	credential := "not set"
+	if strings.TrimSpace(profile.APIKey) != "" || strings.TrimSpace(profile.AuthHeaderValue) != "" {
+		credential = "set"
 	}
 	return check("provider.config", "Provider config", StatusPass, fmt.Sprintf("Provider config loaded for %s.", providerName(profile)), map[string]any{
-		"name":     profile.Name,
-		"provider": profile.ProviderKind,
-		"baseURL":  profile.BaseURL,
-		"model":    profile.Model,
-		"apiKey":   apiKey,
+		"name":                 profile.Name,
+		"provider":             profile.ProviderKind,
+		"baseURL":              profile.BaseURL,
+		"model":                profile.Model,
+		"credentialConfigured": credential,
 	})
 }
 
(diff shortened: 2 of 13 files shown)


gnanam1990:
Addressed the review: added `ProviderSnapshotFromProfile` test coverage for the `APIKeySet` AuthHeaderValue path (api-key-only / auth-header-only / both / neither), aligned the other credential-presence site (`doctor.go` now reports 'set' for an auth-header-only profile too), and added the `[REDACTED:openai_key]` placeholder assertion to the scanner test. @coderabbitai review

Vasanthdev2004 (changes requested):
﻿Requesting changes for one remaining behavioral mismatch.

- `internal/doctor/doctor.go:135`: `providerConfigCheck` treats raw `profile.APIKey` / `profile.AuthHeaderValue` as configured credentials, while `ProviderSnapshotFromProfile` now trims both fields before setting `APIKeySet`. That means a whitespace-only key/header reports `apiKey: set` in `zero doctor`, but the provider snapshot reports no key. Please use `strings.TrimSpace(profile.APIKey) != "" || strings.TrimSpace(profile.AuthHeaderValue) != ""` here too, and add whitespace-only table cases around `TestProviderSnapshotAPIKeySetCountsAuthHeaderValue` so this stays covered.

Validation I ran on this head:
- `go test ./internal/redaction ./internal/secrets ./internal/doctor ./internal/zerocommands ./internal/zerogit` passed
- `go test ./...` passed

gnanam1990:
Addressed both reviews.

- Added whitespace-only table cases (api-key-only / auth-header-only) to `TestProviderSnapshotAPIKeySetCountsAuthHeaderValue` (CodeRabbit's nitpick + part of @Vasanthdev2004's ask).
- `doctor.go` now trims both `APIKey` and `AuthHeaderValue` before the presence check, matching `ProviderSnapshot.APIKeySet`.

One thing worth flagging from digging into this: the doctor indicator was reported under the `apiKey` detail key, which is itself a *sensitive* key — so `check()`'s `RedactValue` scrubbed it to `[REDACTED]` regardless, meaning the set/not-set value never actually surfaced (the whitespace symptom couldn't appear in output either). I moved it to a non-sensitive `credentialConfigured` key so the presence indicator is real, trimmed, and now covered by a doctor table test. No consumer read the old key. @coderabbitai review

Vasanthdev2004 (approved):
﻿Approved from my pass.

The credential-presence blocker I raised is fixed: `providerConfigCheck` now trims both `APIKey` and `AuthHeaderValue`, and the user-visible doctor detail moved off the sensitive `apiKey` key so `set` / `not set` is not scrubbed away. Provider snapshot whitespace cases are covered too.

Non-blocking note: CodeRabbit's extra doctor table cases (`both`, whitespace auth-header-only) are reasonable coverage polish, but I do not see a remaining runtime blocker.

Validation I ran on this head:
- `go test -count=1 ./internal/redaction ./internal/secrets ./internal/doctor ./internal/zerocommands ./internal/zerogit` passed
- `go test ./...` passed

gnanam1990:
Addressed the remaining review findings in `5d3bd79`:

- **`providermodeldiscovery`** now honors `AuthHeaderValue`: the `/models` probe is gated via `discoveryHasCredential` (APIKey **or** AuthHeaderValue) and the request is built with `providerio.ApplyAuthHeaders` (honors AuthHeaderValue / AuthHeader / AuthScheme / CustomHeaders). APIKey-only behavior is unchanged.
- **`redactProviderBaseURL`** now also redacts `AuthHeaderValue` if it leaks into the base URL (variadic secrets, empties dropped).
- Added regression tests: auth-header-only discovery, `discoveryHasCredential`, and auth-header redaction in the base URL.

The `APIKeySet` AuthHeaderValue coverage and the modern-OpenAI-key placeholder assertion were already present. `go build`/`go vet`/`gofmt`/package tests green.

@coderabbitai review

gnanam1990:
@coderabbit review

gnanam1990:
Pushed `8c362c9`: added the `both` and `whitespace auth header only` cases to `TestProviderConfigCheckCredentialPresence` so it matches the six cases in `contracts_test.go`.

Status of the other threads against current HEAD:
- `doctor.go:136` ("missing TrimSpace") — already resolved; `providerConfigCheck` uses `strings.TrimSpace` on both `APIKey` and `AuthHeaderValue`.
- `contracts.go:169` (discovery AuthHeaderValue consistency) — fixed in `5d3bd79` (`providermodeldiscovery` now honors `AuthHeaderValue`).
- `contracts.go:170` (APIKeySet test coverage) — covered by `TestProviderSnapshotAPIKeySetCountsAuthHeaderValue`.

@coderabbitai review

gnanam1990:
@coderabbit approvable ?

PR #175, opened 2026-06-12
wedge build: web_search + LSP client/diagnostics (stages 01-03)

gnanam1990:
First three stages of the ZERO wedge build pack — *the model-agnostic agent that runs unattended and fixes its own mistakes*. Each stage is TDD-gated and tagged (`zero-wedge-01..03`). Stdlib-only; clean-room (Go implemented fresh from behavioural reference, no reproduced source). Verified the built binary runs and registers the new tool.

## Stage 01 — `web_search` tool (`zero-wedge-01`)
Complements `web_fetch` (fetch a known URL) by *discovering* URLs — what an unattended agent needs when it hits unknown errors or new library versions.
- `internal/tools/web_search.go`: `webSearchTool` over a pluggable `searchBackend`; env-configured generic JSON backend (`ZERO_WEBSEARCH_BASE_URL`/`_API_KEY`/`_PROVIDER`), swappable for Exa/You/Tavily/Brave. `query` + clamped `limit` (default 5, cap 10); numbered-list output; output redaction; clear "no search backend configured" error.
- Registered in `CoreNetworkTools()`.

**Intentional deviation from the stage prompt:** the prompt said `readOnlySafety`, but the repo's own `TestCoreNetworkToolsExposePromptMetadata` enforces that every core network tool is prompt-gated egress. Shipping an auto-allowed network tool (and weakening that test) would be a security regression, so `web_search` matches `web_fetch`'s `network/prompt` posture — confirmed in the built binary (`web_search [network/prompt]`).

## Stage 02 — LSP client foundation (`internal/lsp`) (`zero-wedge-02`)
A **direct** (self-spawned, stdio) Language Server Protocol client so an unattended run sees real compiler diagnostics with no open editor.
- `client.go`: JSON-RPC 2.0 + LSP `Content-Length` framing; concurrent id-matched `Call`/`Notify`, server→client request auto-reply, notification handler, context cancellation.
- `server.go`: lifecycle (spawn, `initialize`/`initialized` handshake, graceful shutdown+exit with kill fallback).
- `registry.go`: extension → server command (.go/.ts/.tsx/.js/.jsx/.py/.rs), languageId, PATH availability.
- `types.go`: minimal LSP types + `PathToURI`/`URIToPath` (Windows-drive aware).

## Stage 03 — LSP document sync + diagnostics (`zero-wedge-03`)
Turns the client into actionable signal: "did my edit compile?", headless.
- `documents.go`: per-server session — document lifecycle (didOpen / full-sync didChange / didClose), URI→diagnostics store fed by `publishDiagnostics`, `WaitForDiagnostics` with a publish-baseline + 300ms quiet debounce.
- `diagnostics.go`: `FormatDiagnostics` (`path:line:col: severity: message`, 1-based), `HasErrors`, `FilterBySeverity`.
- `manager.go`: `Manager` — one lazily-started, reused server per language; `Check(ctx,path,text) → settled diagnostics`; unknown extension → `(nil,nil)`.

A TDD-caught bug worth noting: `WaitForDiagnostics` initially returned the *previous* check's diagnostics (keyed off global last-publish time); fixed with a per-URI publish counter so each check waits for its own publish.

## Verification
Per-stage gate, all green: `go build ./...`, full `go test ./...`, `gofmt -l` clean. LSP concurrency tested with `-race` over in-memory pipes (stub server, no real `gopls`). Built the binary locally: runs, lists `web_search` as prompt-gated, `internal/lsp` compiled in.

## Notes / sequencing
- `internal/lsp` has no CLI surface yet — it's foundation consumed by stage 04 (the self-correct loop), which will wire testrunner/verify/selfverify/LSP into `internal/agent/loop.go`. That stage touches the core agent loop (conflict-prone vs in-flight TUI/loop PRs) and is intentionally **not** in this PR — it should land deliberately after sequencing.

## Summary by CodeRabbit

* **New Features**
  * LSP support: language-server integration with document sync, diagnostics delivery, and graceful startup/shutdown.
  * Web search tool: network-backed searches returning ranked title/URL/snippet results with query/limit.
  * Post-edit self-correction: optional verify-and-correct feedback appended once after edits.

* **Utilities**
  * Path↔URI helpers, compact diagnostic formatting/filtering, and registry for server/language mappings.

* **Tests**
  * Expanded unit/integration tests covering LSP, diagnostics, registry, web search, and self-correction.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -326,6 +326,19 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			if outcome.InjectHint && failureHint == "" {
 				failureHint = toolFailureHint(call.Name, toolSchemaJSON(registry, call.Name), toolResult.Output)
 			}
+
+			// Post-edit self-correction: after a successful mutating tool call,
+			// verify the changed files and feed any failure back so the model can
+			// fix it next turn. nil SelfCorrect (the default) is a no-op, and a
+			// read-only tool (no ChangedFiles) never triggers verification.
+			if options.SelfCorrect != nil && toolResult.Status == tools.StatusOK && len(toolResult.ChangedFiles) > 0 {
+				if feedback, _ := options.SelfCorrect.AfterEdit(ctx, toolResult.ChangedFiles); feedback != "" {
+					messages = append(messages, zeroruntime.Message{
+						Role:    zeroruntime.MessageRoleUser,
+						Content: feedback,
+					})
+				}
+			}
 		}
 
 		// Mid-run model escalation: if a tool asked to switch models this turn and
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -260,12 +260,15 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		// after the batch, so every tool_result is recorded first and at most one
 		// switch occurs per turn.
 		turnRequestedModel := ""
-		// selfCorrectFeedback accumulates post-edit verification feedback during
-		// the batch; it is appended ONCE after the loop so every advertised tool
-		// call keeps its tool_result contiguous (a user message interleaved between
-		// tool_results breaks strict provider replay) — same after-batch rationale
-		// as turnRequestedModel above.
-		var selfCorrectFeedback []string
+		// changedFilesThisBatch aggregates every file the turn's mutating tools
+		// touched, so post-edit self-correction runs ONCE over the union after the
+		// batch — one AfterEdit call keeps the per-run attempt budget accurate and
+		// avoids verifying an intermediate edit a later call in the same turn already
+		// superseded. Its feedback is appended after the loop so every advertised
+		// tool call keeps its tool_result contiguous (a user message interleaved
+		// between tool_results breaks strict provider replay) — same after-batch
+		// rationale as turnRequestedModel above.
+		var changedFilesThisBatch []string
 		for index, call := range collected.ToolCalls {
 			if options.OnToolCall != nil {
 				options.OnToolCall(call)
@@ -333,25 +336,25 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				failureHint = toolFailureHint(call.Name, toolSchemaJSON(registry, call.Name), toolResult.Output)
 			}
 
-			// Post-edit self-correction: after a successful mutating tool call,
-			// verify the changed files and feed any failure back so the model can
-			// fix it next turn. nil SelfCorrect (the default) is a no-op, and a
-			// read-only tool (no ChangedFiles) never triggers verification.
+			// Post-edit self-correction: collect the files this successful mutating
+			// tool changed; verification runs once over the union after the batch.
+			// A read-only tool (no ChangedFiles) never contributes.
 			if options.SelfCorrect != nil && toolResult.Status == tools.StatusOK && len(toolResult.ChangedFiles) > 0 {
-				if feedback, _ := options.SelfCorrect.AfterEdit(ctx, toolResult.ChangedFiles); feedback != "" {
-					selfCorrectFeedback = append(selfCorrectFeedback, feedback)
-				}
+				changedFilesThisBatch = append(changedFilesThisBatch, toolResult.ChangedFiles...)
 			}
 		}
 
-		// Append accumulated self-correct feedback now that every tool_result for
-		// this batch is recorded, so the assistant's tool_results stay contiguous
-		// (a user message between tool_results breaks strict provider replay).
-		if len(selfCorrectFeedback) > 0 {
-			messages = append(messages, zeroruntime.Message{
-				Role:    zeroruntime.MessageRoleUser,
-				Content: strings.Join(selfCorrectFeedback, "\n\n"),
-			})
+		// Run post-edit self-correction once over the union of files this turn
+		// changed, then append any feedback after every tool_result is recorded so
+		// the assistant's tool_results stay contiguous (a user message between
+		// tool_results breaks strict provider replay). nil SelfCorrect is a no-op.
+		if options.SelfCorrect != nil && len(changedFilesThisBatch) > 0 {
+			if feedback, _ := options.SelfCorrect.AfterEdit(ctx, dedupeStrings(changedFilesThisBatch)); feedback != "" {
+				messages = append(messages, zeroruntime.Message{
+					Role:    zeroruntime.MessageRoleUser,
+					Content: feedback,
+				})
+			}
 		}
 
 		// Mid-run model escalation: if a tool asked to switch models this turn and
@@ -1389,6 +1392,21 @@ func appendAbortedToolResults(messages []Message, remaining []ToolCall) []Messag
 		})
 	}
 	return messages
+}
+
+// dedupeStrings returns values with duplicates removed, preserving first-seen
+// order so the deduped list (e.g. a turn's changed files) stays deterministic.
+func dedupeStrings(values []string) []string {
+	seen := make(map[string]bool, len(values))
+	out := make([]string, 0, len(values))
+	for _, value := range values {
+		if seen[value] {
+			continue
+		}
+		seen[value] = true
+		out = append(out, value)
+	}
+	return out
 }
 
 func copyMessages(messages []Message) []Message {
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -260,6 +260,12 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		// after the batch, so every tool_result is recorded first and at most one
 		// switch occurs per turn.
 		turnRequestedModel := ""
+		// selfCorrectFeedback accumulates post-edit verification feedback during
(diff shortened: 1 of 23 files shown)


gnanam1990:
Addressed the review. LSP: `Close()` now disables the client (Call/Notify reject post-close); same-document syncs are serialized per-URI across the Notify and version is committed only on success; delayed `publishDiagnostics` for a superseded version is dropped; `Check` returns no diagnostics (rather than a stale prior result) when no publish newer than the sync arrives; `Manager.Shutdown` now joins and returns per-server errors; `FormatDiagnostics` collapses embedded whitespace to one line per diagnostic. web_search: `ZERO_WEBSEARCH_PROVIDER` is now forwarded in the request. Added tests for each (incl. an httptest backend round-trip and a -race pass on the LSP concurrency). @coderabbitai review

Vasanthdev2004 (changes requested):
﻿Requesting changes for one production behavior gap.

- `internal/lsp/manager.go:55-64`: `Manager.Check` promises graceful degradation when no language server is available, but it only checks `ServerFor(path)` and then calls `sessionFor`, which eventually runs `StartServer`. On a clean machine without `gopls`, `pyright-langserver`, `rust-analyzer`, etc., a configured extension returns an executable-not-found error instead of `(nil, nil)`. Since LSP diagnostics are an opportunistic foundation layer, missing local language servers should degrade the same way unsupported extensions do. Please check availability before starting, or translate executable-not-found from the starter into no diagnostics, and add a test for a configured extension whose server binary is unavailable.

Non-blocking note: I saw the CodeRabbit test comments in `manager_test.go`; the ignored unmarshal is worth cleaning up, but `t.Errorf` from worker goroutines is not the merge blocker for me here.

Validation I ran on this head:
- `go test ./internal/lsp ./internal/tools ./internal/specialist` passed
- `go test ./...` passed

gnanam1990:
Addressed @Vasanthdev2004's blocker: `Manager.Check` now degrades to `(nil, nil)` when the configured language-server binary isn't installed (translates `exec.ErrNotFound` from the starter), so a clean machine without gopls/pyright/rust-analyzer behaves like an unsupported extension rather than erroring. Added a test injecting an `exec.ErrNotFound` starter. (CodeRabbit's manager_test nits — ignored unmarshal / goroutine t.Errorf — are non-blocking and noted.) @coderabbitai review

Vasanthdev2004 (changes requested):
﻿Requesting changes for one new blocker in the stage-04 self-correct integration.

- `internal/agent/loop.go:334-340`: self-correct feedback is appended as a `MessageRoleUser` immediately after a successful mutating tool result, while the loop may still have later tool calls from the same assistant message to execute. If a model emits multiple tool calls in one turn and the first mutates, the message sequence becomes `assistant(tool_calls: c1,c2) -> tool(c1) -> user(self-correct feedback) -> tool(c2)`, which violates strict provider replay ordering. Providers like Anthropic require every advertised tool_use to be answered by its tool_result before any new user message. Please queue/aggregate self-correct feedback until after the `for collected.ToolCalls` batch has appended all tool results, or run self-correct once after the batch over the union of changed files. Add a regression test with two tool calls where the first mutates and self-correct emits feedback, asserting all tool messages for the batch precede the self-correct user message.

The previous LSP missing-binary blocker is fixed: `Manager.Check` now degrades to `(nil, nil)` for `exec.ErrNotFound` and keeps other start failures surfaced.

Validation I ran on this head:
- `go test -count=1 ./internal/agent ./internal/lsp ./internal/tools ./internal/specialist` passed
- `go test ./...` passed

gnanam1990:
The `Close()` lifecycle finding is already resolved (commit `c734088`):

- `Call()` rejects work via an up-front `select { case <-c.closed: ... }` guard.
- `Notify()` (and Call's request write) are rejected by `write()`'s own `<-c.closed` check — its comment notes exactly this.
- `failPending` swaps the whole `pending` map for a fresh one under `closeOnce`, so any entry registered during a close race lives in the *discarded* map and is GC'd — no leak. Adding `delete(c.pending, id)` to the `<-c.closed` select branch would be a no-op given that design (unlike the `ctx.Done` branch, where failPending has **not** run).

Verified by `TestClientRejectsCallsAfterClose` (passes with `-race`). `web_search` also forwards `ZERO_WEBSEARCH_PROVIDER` now.

@coderabbitai review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
Verified each open thread against current HEAD (`3b77c97`). Six of the seven flagged items were already fixed in the latest pushes and are stale re-anchors:

- `lsp/manager.go:86` — resolved: `Check` returns `nil` when `waitForDiagnostics` observes no publish newer than the baseline.
- `lsp/manager.go:105` — resolved: `Shutdown` collects per-server errors via `errors.Join`.
- `lsp/documents.go:75` — resolved: `publishDiagnostics` drops a strictly-older positive version.
- `lsp/documents.go:90` — resolved: `sync` holds a per-URI lock across version select + notify and commits state only after the write succeeds.
- `lsp/diagnostics.go:75` — resolved: `FormatDiagnostics` flattens via `collapseWhitespace` (`strings.Fields`), so each diagnostic stays one physical line.
- `lsp/client.go:148` — resolved: `Call` guards on `<-c.closed` up front, writes are rejected once closed, and `failPending` swaps the whole pending map so a close-race entry is discarded (no leak).

The one genuinely-open item is fixed in `3b77c97`:
- `selfcorrect.go` — `inspect` no longer swallows `checker.Check`/`verifier.Verify` errors. A non-nil error (LSP startup/sync failure, unreadable changed file, or failed plan detection) is now recorded in `CorrectionReport.InspectErrors`, fails the pass, and is surfaced in the model feedback — instead of producing a false `passed`. Added regression tests for both paths.

@coderabbitai review

Vasanthdev2004 (changes requested):
Verdict: Request changes

Finding:
[P1] Preserve tool_result ordering when self-correct feedback is generated - internal/agent/loop.go:334

The self-correct feedback is appended inside the loop that is still answering the assistant turn's tool calls. If a model emits multiple tool calls in one turn and an earlier successful mutating tool triggers self-correct feedback, the transcript becomes: assistant tool_use batch -> tool_result for call 1 -> user self-correct feedback -> tool_result for call 2. Strict provider replay expects all tool_result messages for the assistant's advertised tool calls to be contiguous before the next user message, so this can break the next completion request or later session replay/compaction for common multi-tool batches.

Move self-correct feedback emission until after the current tool-call batch has finished, or collect pending feedback during the loop and append it only after every advertised tool call has a matching tool_result or aborted placeholder. Please add a regression test with two tool calls in one assistant turn where the first mutates and self-correct fails, asserting all tool results are recorded before the feedback message.

Validation run locally:
- go test ./...

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
Fixed in `17b9de7`.

Self-correct feedback is now accumulated during the tool-call batch and appended **once after the loop**, so every advertised tool call keeps its `tool_result` contiguous — no user message is interleaved between `tool_result`s. This mirrors the existing after-batch pattern already used for mid-run model escalation (`turnRequestedModel`). On the early-return abort/stop/guard paths the run ends, so deferred feedback is intentionally dropped (no next turn to act on it).

Added `TestRunDefersSelfCorrectFeedbackUntilAfterToolBatch`: two tool calls in one assistant turn where the first mutates and self-correct fails, asserting both `tool_result`s are contiguous and the feedback message lands after both. `go test ./internal/agent` + `-race`, `go vet`, `gofmt` green.

@coderabbitai review

gnanam1990:
Good catch — fixed in `14823b6`.

`AfterEdit` is now called **once per turn** over the deduped union of every mutating tool's `ChangedFiles` (aggregated during the batch), instead of once per mutating call. This keeps the per-run attempt budget accurate and avoids verifying an intermediate edit a later call in the same turn already superseded. Feedback is still appended once, after all `tool_result`s.

Added `dedupeStrings` and `TestRunBatchesSelfCorrectOncePerTurn`: two writes in one assistant turn now consume exactly one corrective attempt and emit a single feedback message. `go test ./internal/agent` (+`-race`), `go vet`, `gofmt` green.

@coderabbitai review

gnanam1990:
@coderabbit approve

Vasanthdev2004 (approved):
Verdict: Approve

I re-reviewed the latest head `14823b6`.

No blockers found.

What I checked:
- `web_search` is prompt-gated network egress, registered in the core network tool set, and the configured provider is now forwarded to the backend request body.
- LSP client/session fixes address the earlier close/pending-call and stale diagnostics concerns.
- Self-correct feedback is now batched after all tool results, preserving strict provider replay ordering, and `AfterEdit` runs once over the deduped changed-file union.

Validation run locally:
- `gofmt -l internal/agent internal/lsp internal/tools internal/specialist` clean
- `git diff --check origin/main...HEAD` clean
- `go test -count=1 ./internal/agent ./internal/lsp ./internal/tools ./internal/specialist`
- `go test ./...`

Note: I attempted `go test -race ./internal/lsp ./internal/agent`, but this Windows environment cannot run race builds because `gcc` is missing for cgo.

PR #176, opened 2026-06-12
wedge stage 04: self-correct loop

gnanam1990:
**Stacked on #175** (base `feat/wedge-build`) — review/merge #175 first; this diff shows only stage 04.

The headline wedge: after a mutating edit, ZERO verifies the changed files and fixes its own mistakes, unattended and bounded.

## What it does
After a successful mutating tool call, `loop.go` runs `SelfCorrector.AfterEdit(changedFiles)`:
- **Verify** — LSP error diagnostics for each changed file (stage 03 `lsp.Manager`) + the project's `verify` plan (`DetectPlan`+`Run`).
- **Correct** — on failure, append a concise, secret-redacted, model-facing message ("Verification failed after your edit. Fix these and continue:" + diagnostics + failing check output) so the model fixes it next turn.
- **Bound** — a per-run attempt ceiling (default 3); once exhausted it emits a stop notice instead of looping.
- **Gate** — autonomy ceiling: `low` reports only (no auto-attempt); `medium`/`high` auto-correct. Read-only tools (no `ChangedFiles`) never trigger it. `nil` SelfCorrect (the default) is a complete no-op.

## Scope (per the stage spec's file list)
- `internal/agent/selfcorrect.go` (engine, injectable `diagnosticsChecker`/`projectVerifier`)
- `internal/agent/loop.go` (gated invocation after a mutating tool result)
- `internal/agent/types.go` (**explicit options change:** `Options.SelfCorrect *SelfCorrector`)
- `internal/agent/selfcorrect_test.go`

## Design notes
- Single-pass `verify.Run` per call; the multi-attempt cycle comes from the agent loop + the model editing between turns (not `selfverify`'s same-command re-runs). The per-run counter provides the hard bound.
- **CLI activation deliberately deferred:** a `--self-correct` / `ZERO_SELF_CORRECT` flag that constructs the `SelfCorrector` in `exec.go` would touch files outside this stage's scope. The engine + adapters (`NewSelfCorrector`, `NewLSPDiagnosticsChecker`, `NewProjectVerifier`) are exported and ready; wiring is a small follow-up.

## Verification
TDD: fail-then-pass, max-attempts abort, read-only skip, low-autonomy report-only, LSP-only failure, secret redaction, disabled/nil no-op. `go build ./...`, full `go test ./...`, `-race` on `internal/agent`, `gofmt` all clean; 0 session leak. Loop change is nil-gated → existing loop tests byte-identical.

## Summary by CodeRabbit

* **New Features**
  * Post-edit self-correction: after file-changing operations, the system verifies edits (diagnostics and tests), returns scrubbed feedback, and can auto-attempt fixes up to configurable limits with autonomy gating.

* **Tests**
  * Added extensive tests covering pass/fail flows, retry limits and aborts, LSP diagnostics handling, redaction of sensitive output, disabled behavior, and safety of nil config.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -326,6 +326,19 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			if outcome.InjectHint && failureHint == "" {
 				failureHint = toolFailureHint(call.Name, toolSchemaJSON(registry, call.Name), toolResult.Output)
 			}
+
+			// Post-edit self-correction: after a successful mutating tool call,
+			// verify the changed files and feed any failure back so the model can
+			// fix it next turn. nil SelfCorrect (the default) is a no-op, and a
+			// read-only tool (no ChangedFiles) never triggers verification.
+			if options.SelfCorrect != nil && toolResult.Status == tools.StatusOK && len(toolResult.ChangedFiles) > 0 {
+				if feedback, _ := options.SelfCorrect.AfterEdit(ctx, toolResult.ChangedFiles); feedback != "" {
+					messages = append(messages, zeroruntime.Message{
+						Role:    zeroruntime.MessageRoleUser,
+						Content: feedback,
+					})
+				}
+			}
 		}
 
 		// Mid-run model escalation: if a tool asked to switch models this turn and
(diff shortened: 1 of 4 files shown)


gnanam1990:
Addressed in `3ceb7d7`: `lspDiagnosticsChecker.Check` now propagates the `os.ReadFile` error instead of returning `(nil, nil)` — fixes the golangci-lint `nilerr` finding. The sole caller (`inspect`) already skips files whose `Check` errors, so a deleted/unreadable file still degrades quietly while future callers can decide. Added a regression test.

`go build`/`gofmt`/`go test ./internal/agent` green.

@coderabbitai review

PR #179, opened 2026-06-12
wedge M2: MCP server resources/prompts + sandbox scoped egress + MCP OAuth

gnanam1990:
## Summary

Milestone 2 of the wedge build pack, three related modules in one PR (clean-room Go, stdlib-first, TDD-gated). Builds directly on `main`.

### Stage 06 — MCP server: resources + prompts
ZERO's MCP server previously advertised only `tools`. It is now a first-class server that also serves **resources** and **prompts**.
- `server.go`: `initialize` advertises `{tools, resources, prompts}`; new handler cases for `resources/list`, `resources/read`, `prompts/list`, `prompts/get` (existing transport/framing + `writeResult`/`writeError` unchanged). `ServeOptions` gained `WorkspaceRoot` + `Scope`.
- `resources.go` (new): enumerate in-scope files (reusing the workspace indexer's `.git`/`node_modules`/`vendor`/binary exclusions) and read them. **Scope-enforced**: symlink-resolved containment check rejects `../` traversal and out-of-scope absolute paths with a JSON-RPC error and no contents; binary files return a base64 blob; read-only.
- `prompts.go` (new): a curated prompt catalogue (code-review / spec-draft / explain) with `{{arg}}` substitution; internal system prompts are deliberately not exposed.

### Stage 07 — sandbox per-domain egress + auto-allow-bash
Adds a middle ground between full-open and full-deny networking.
- `types.go`: `NetworkScoped` mode + `Policy.AllowedDomains`/`DeniedDomains` + `AutoAllowBashWhenSandboxed` (off by default, `ZERO_SANDBOX_AUTO_ALLOW_BASH`).
- `egress.go` (new): in-process filtering proxy on `127.0.0.1:0` handling HTTP `CONNECT` (HTTPS tunnel) and plain HTTP. Domain matching is exact-or-subdomain at a label boundary (`api.github.com` matches `github.com`; `notgithub.com` and `github.com.evil.com` do not); deny-list overrides allow-list. Decisions are logged through redaction so query-string tokens never leak. **Fails closed** — empty allowlist or any proxy-start error denies the network; never falls back to open.
- `runner.go`: wires `NetworkScoped` for both bubblewrap (keeps `--unshare-net`, exports proxy env) and sandbox-exec (allows only localhost:proxy). `NetworkAllow`/`NetworkDeny` behavior unchanged. The proxy is cleaned up after the command via `CommandPlan.Cleanup()`.
- `engine.go`/`bash`: auto-allows a shell tool only when the sandbox is actually active and the autonomy ceiling permits; ignored when unsandboxed.

### Stage 08 — MCP OAuth client
Lets ZERO connect to remote MCP servers that require OAuth.
- `oauth.go` (new): OAuth 2.0 + PKCE authorization-code flow — metadata discovery (`/.well-known/oauth-authorization-server`) with config-endpoint fallback, optional dynamic client registration, S256 PKCE, a `127.0.0.1:0` loopback redirect listener, **CSRF state validation**, code→token exchange, and refresh. Headless-safe (prints the authorization URL rather than force-opening a browser). Tokens are never logged.
- `oauth_store.go` (new): per-server token store in a `0600` file under the config dir.
- `network_client.go`: attaches `Authorization: Bearer` for OAuth servers and, on `401`, refreshes and retries once; on refresh failure surfaces an actionable re-login message.
- `config.go` (+ `config/types.go`/`resolver.go` carriers): an MCP server entry may declare `auth: "oauth"`.
- `zero mcp oauth {login,logout,status}` subcommands (`status` reports presence/expiry, never the token).

## Verification
- `go build ./...`, `go vet ./...`, `gofmt -l` (clean), and `go test ./...` (all 57 packages) green.
- `-race` on `internal/mcp` and `internal/sandbox` (the new proxy/loopback concurrency) green.
- Security-critical paths spot-reviewed: domain-boundary matching and symlink-resolved resource scoping both reject the adversarial cases, with TDD coverage.

## Notes / scope
- Stage 08's `auth: "oauth"` required additive carrier fields in `internal/config/types.go` + `resolver.go` and the `zero mcp oauth` CLI files — minimal and necessary to wire the in-scope `internal/mcp/config.go` change to the command surface.
- Deferred 1-liner: `internal/cli/serve.go` does not yet pass its `-C/--cwd` to `ServeOptions.WorkspaceRoot` (the server falls back to the process working dir, correct for normal `serve`). Tracked for a follow-up.
- Stage 05 (hooks upgrade) is intentionally **not** here — it modifies `internal/agent/loop.go`, which conflicts with the in-flight #175; it should land after #175 merges.

## Summary by CodeRabbit

* **New Features**
  * OAuth 2.0 (PKCE) sign-in for MCP servers with persistent, redaction-safe token storage and status reporting (JSON/human).
  * New CLI: `zero mcp oauth` with `login`, `logout`, and `status`.
  * MCP "resources" API to list/read workspace files (text or base64 blobs).
  * MCP "prompts" API with curated templates and argument-driven rendering.
  * Sandboxed egress proxy and scoped network mode (allow/deny lists) for sandboxed runs.
  * Auto-allow bash when native sandbox isolation is active.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -46,6 +46,7 @@ type appDeps struct {
 	loadHooks            func(hooks.LoadOptions) (hooks.LoadResult, error)
 	skillsDir            func() string
 	newMCPStore          func() (*mcp.PermissionStore, error)
+	newMCPTokenStore     func() (*mcp.TokenStore, error)
 	newSandboxStore      func() (*sandbox.GrantStore, error)
 	selectSandboxBackend func(sandbox.BackendOptions) sandbox.Backend
 	registerMCPTools     func(context.Context, *tools.Registry, config.MCPConfig, mcp.RegisterOptions) (mcpToolRuntime, error)
@@ -112,6 +113,9 @@ func defaultAppDeps() appDeps {
 		},
 		newMCPStore: func() (*mcp.PermissionStore, error) {
 			return mcp.NewPermissionStore(mcp.StoreOptions{})
+		},
+		newMCPTokenStore: func() (*mcp.TokenStore, error) {
+			return mcp.NewTokenStore(mcp.TokenStoreOptions{})
 		},
 		newSandboxStore: func() (*sandbox.GrantStore, error) {
 			return sandbox.NewGrantStore(sandbox.StoreOptions{})
@@ -309,6 +313,9 @@ func fillAppDeps(deps appDeps) appDeps {
 	}
 	if deps.newMCPStore == nil {
 		deps.newMCPStore = defaults.newMCPStore
+	}
+	if deps.newMCPTokenStore == nil {
+		deps.newMCPTokenStore = defaults.newMCPTokenStore
 	}
 	if deps.newSandboxStore == nil {
 		deps.newSandboxStore = defaults.newSandboxStore
--- a/internal/cli/extensions.go
+++ b/internal/cli/extensions.go
@@ -144,6 +144,8 @@ func runMCP(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int
 		return runMCPPermissions(args[1:], stdout, stderr, deps)
 	case "tools":
 		return runMCPTools(args[1:], stdout, stderr, deps)
+	case "oauth":
+		return runMCPOAuth(args[1:], stdout, stderr, deps)
 	case "list":
 		return runMCPLegacyList(args[1:], stdout, stderr, deps)
 	default:
@@ -556,6 +558,7 @@ func writeMCPHelp(w io.Writer) error {
 
 Commands:
   list           List configured MCP servers, or tools with --tools
+  oauth          Manage OAuth credentials for remote MCP servers
   permissions    Manage persistent MCP tool permissions
   tools          Inspect configured MCP tools
 `)
--- a/internal/cli/mcp_oauth.go
+++ b/internal/cli/mcp_oauth.go
@@ -34,7 +34,7 @@ func runMCPOAuth(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 }
 
 func runMCPOAuthLogin(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
-	_, positional, help, err := parseMCPPositionalCommand(args)
+	opts, positional, help, err := parseMCPPositionalCommand(args)
 	if err != nil {
 		return writeExecUsageError(stderr, err.Error())
 	}
@@ -43,6 +43,12 @@ func runMCPOAuthLogin(args []string, stdout io.Writer, stderr io.Writer, deps ap
 			return exitCrash
 		}
 		return exitSuccess
+	}
+	if opts.json {
+		// login is an interactive flow (prints a URL, waits for the callback); a
+		// machine-readable mode would be misleading, so reject it rather than
+		// accepting the flag and ignoring it.
+		return writeExecUsageError(stderr, "zero mcp oauth login does not support --json")
 	}
 	if len(positional) != 1 {
 		return writeExecUsageError(stderr, "usage: zero mcp oauth login <server>")
--- a/internal/cli/mcp_oauth.go
+++ b/internal/cli/mcp_oauth.go
@@ -1,0 +1,240 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/mcp"
+	"github.com/Gitlawb/zero/internal/redaction"
+)
+
+func runMCPOAuth(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
+	if len(args) == 0 {
+		return writeExecUsageError(stderr, "mcp oauth subcommand required. Use `zero mcp oauth status`.")
+	}
+	if args[0] == "-h" || args[0] == "--help" || args[0] == "help" {
+		if err := writeMCPOAuthHelp(stdout); err != nil {
+			return exitCrash
+		}
+		return exitSuccess
+	}
+
+	switch args[0] {
+	case "login":
+		return runMCPOAuthLogin(args[1:], stdout, stderr, deps)
+	case "logout":
+		return runMCPOAuthLogout(args[1:], stdout, stderr, deps)
+	case "status":
+		return runMCPOAuthStatus(args[1:], stdout, stderr, deps)
+	default:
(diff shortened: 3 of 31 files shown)


gnanam1990:
Addressed all findings in `38fdc34`, plus the Windows Smoke failure.

**Critical**
- `resources.go` — symlinks are now resolved **before** the scope check (roots resolved too), so an in-workspace symlink to an out-of-scope target is rejected by `resources/read`. The response echoes the requested URI (not the resolved path) so `list`/`read` stay consistent. Regression test: `TestServeResourcesReadRejectsInRootSymlinkEscape`.
- `egress.go` — `handleHTTP` now authorizes the dialed URL host instead of the client `Host` header, closing the `GET http://denied/ … Host: allowed` bypass. Regression test: `TestEgressProxyHTTPAuthorizesURLHostNotHostHeader`.

**Major**
- `prompts.go` — `prompts/get` rejects a missing required argument; substitution is tokenized so `{{…}}` inside supplied values is preserved. Tests: `TestServePromptsGetMissingRequiredArgErrors`, `TestSubstituteArgsPreservesPlaceholdersInUserInput`.
- `resources.go` — file URIs percent-encode/decode via `net/url` (spaces, `#`, `%` round-trip).
- `oauth.go` — one timeout-scoped context bounds discovery + registration + callback + exchange; `joinWellKnown` preserves issuer path segments (RFC 8414).
- `mcp_oauth_test.go` — login test is bounded (goroutine + select) so a missed callback fails fast.

**Minor / CI**
- `mcp_oauth login` now rejects `--json` instead of ignoring it.
- XDG token-store path test uses `t.TempDir()` → **fixes the Windows Smoke failure**.

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `go test ./...` (all packages) + `-race` on `internal/mcp`/`internal/sandbox` green locally.

@coderabbitai review

Vasanthdev2004 (changes requested):
Verdict: REQUEST CHANGES

Findings:

1. [P1] Scoped egress is unenforced on the policy-only fallback. `Engine.Evaluate` treats a populated `NetworkScoped` policy as allowed because the proxy is expected to enforce the domain allowlist (`internal/sandbox/engine.go:155-160`), but `BuildCommandPlan` falls through to `directCommandPlan` for `BackendPolicyOnly` (`internal/sandbox/runner.go:154-157`) without starting the egress proxy, setting proxy env, or denying network. On Windows or any machine without a native sandbox backend, a config that says `network: scoped` with `allowedDomains` silently runs with unrestricted host networking. Please fail closed for `NetworkScoped` when the backend cannot enforce it, or collapse scoped to deny in policy-only mode, and add a regression test for that path.

2. [P1] Bubblewrap scoped egress exports a host-loopback proxy inside an unshared network namespace without a bridge. `bubblewrapCommandPlan` keeps `--unshare-net` for `NetworkScoped` (`internal/sandbox/runner.go:292-293`) and then sets `HTTP_PROXY`/`HTTPS_PROXY` to the proxy address bound on host `127.0.0.1` (`internal/sandbox/runner.go:300-306`). Inside the new netns, `127.0.0.1` is the sandbox loopback, not the host proxy, and I do not see any relay/bridge added to the plan despite the comments mentioning one. That means allowlisted egress will not actually reach the filtering proxy under bubblewrap. Please either add the real bridge and an integration test that an allowlisted request succeeds while a denied one fails, or make scoped egress fail closed/unsupported for bubblewrap until that exists.

Validation run locally in an isolated PR worktree:
- `go test ./...` passed

The rest of the MCP OAuth/resources/prompts changes looked well covered in this pass, but I cannot approve while the scoped-egress safety boundary has these enforcement gaps.

Vasanthdev2004:
Re-review verdict: REQUEST CHANGES (unchanged)

I re-reviewed the latest PR head (`38fdc34`) and confirmed there are no new commits after my previous requested-changes review. The same scoped-egress blockers still apply:

- `NetworkScoped` is still treated as allowed in preflight, but `BackendPolicyOnly` falls through to `directCommandPlan` without enforcing the allowlist/proxy. This can silently become unrestricted host networking on Windows/no-native-sandbox hosts.
- Bubblewrap still combines `--unshare-net` with proxy env pointing at host `127.0.0.1`, but I still do not see an actual host-loopback bridge/relay in the plan, so allowlisted egress likely cannot reach the filtering proxy.

Fresh validation on this head:
- `go test ./...` passed

Keeping this as request-changes until scoped egress either fails closed when unenforceable or has real runtime enforcement tests for the native paths.

gnanam1990:
Pushed `0b85ab0` addressing both P1 scoped-egress findings (TDD).

**1. Policy-only fallback fail-open** — `Engine.Evaluate` now collapses `NetworkScoped` to deny when the active backend can't enforce the allowlist, so a network-risk tool fails closed instead of running with unrestricted host networking. On Windows / no-native-backend hosts, `network: scoped` no longer silently allows open egress.

**2. Bubblewrap proxy unreachable** — bubblewrap no longer starts an egress proxy it can't reach from its `--unshare-net` namespace. Scoped collapses to deny (`--unshare-net`, no proxy env), matching Evaluate. The misleading 'bridge' comments are corrected; a real relay (slirp4netns-style) is called out as future work.

**Mechanism:** new `Backend.ScopedEgress` capability — true only for `sandbox-exec` (it shares the host net under a seatbelt profile that restricts outbound to the proxy port, so the allowlist is genuinely enforced) — plus `Backend.EnforcesScopedEgress()`. sandbox-exec scoped egress is unchanged.

**Regression tests:** `TestScopedNetworkGateInEvaluate` now asserts enforcing (sandbox-exec) → allow vs non-enforcing (bubblewrap, policy-only) → deny; `TestBubblewrapScopedPlanFailsClosedWithoutProxy` asserts no proxy is started and `--unshare-net` is kept.

`go test ./...` and `go test -race ./internal/sandbox/` green.

Vasanthdev2004 (approved):
Re-review verdict: APPROVE

The latest head `0b85ab0` resolves the scoped-egress blockers from my previous review:

- Policy-only / unenforceable scoped egress now fails closed in `Evaluate` by treating `NetworkScoped` as deny unless the backend can enforce scoped proxy routing.
- Bubblewrap no longer advertises/proxies scoped egress; it collapses scoped network to deny-equivalent `--unshare-net` until a real bridge exists.
- Sandbox-exec is the only backend marked as scoped-egress capable, and tests cover enforcing vs unenforceable backend behavior.

Validation run locally in the PR worktree:
- `go test ./internal/sandbox ./internal/tools ./internal/cli ./internal/mcp -count=1` passed
- `go test ./...` passed

No blockers found in this re-review.

PR #181, opened 2026-06-12
wedge stage 05: hooks upgrade (events, actions, async/rewake/continueOnBlock)

gnanam1990:
Stage 05 of the wedge build — extends the hook system from "6 events, command-only" toward the mature shape that powers unattended self-correction and external triggers. Built clean-room in Go (semantics only; no competitor source). TDD throughout.

## Done — hooks package (fully unit-tested)
- **New lifecycle events:** `userPromptSubmit`, `preCompact`, `notification`, `stop` (parse + select; matcher still restricted to before/afterTool).
- **Action types** on `Definition` (`type`: `command` default, `prompt`, `http`) plus `async`, `asyncRewake`, `rewakeMessage`, `rewakeSummary`, `continueOnBlock`, `prompt`, `model`, `url`.
- **Action executors** (`actions.go`): `command` (unchanged), `prompt` (injectable `PromptRunner`), `http` (allowlisted POST, capped response body).
- **async / rewake / continueOnBlock** (`dispatch.go`): background execution that doesn't block the turn; an `asyncRewake` hook exiting code 2 emits a `RewakeSignal` on `Dispatcher.Rewakes()`; afterTool block decisions (`continueOnBlock` feeds the reason back and continues, otherwise end-turn).

### Schema / back-compat
New bool fields default false; `type` defaults `command`; `asyncRewake` implies `async`. Existing command-only hook configs parse unchanged (covered by tests). Validation: `prompt` requires `prompt`; `http` requires an allowlisted http(s) `url` and is rejected for setup events (`sessionStart`/`sessionEnd`); unknown `type` rejected.

### Security
`http` actions are gated by an exact-match allowlist enforced at dispatch time (never from config alone); a non-allowlisted URL is rejected before any request is made.

## Done — agent loop wiring
- `userPromptSubmit` fires before the turn loop; hook output is injected once into the run context.
- `stop` fires on every run-exit path (deferred, detached context).

## Remaining (follow-up, intentionally not rushed into the core loop)
- Consume `continueOnBlock`/end-turn and `Rewakes()` in the turn loop's control flow (the dispatch-layer mechanism + tests are in place; wiring the turn-loop control flow is the risky part).
- Fire `preCompact` from the compaction entry point and `notification` from the notify path (the latter ties into Stage 11).

## Validation
`gofmt -l` clean · `go vet ./...` clean · `go build ./...` · `go test ./...` all green.

Stacked on `feat/wedge-build` (stages 01–04); retarget to `main` after that lands.

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle events (userPromptSubmit, preCompact, notification, stop), prompt and HTTP hook actions, HTTP allowlisting and response-size limits, async background hooks with rewake signaling, and reliable stop-hook execution; userPromptSubmit output can be appended to runs.

* **Bug Fixes**
  * Hook dispatch helpers are safe no-ops when no dispatcher is configured (no panics or injected context).

* **Tests**
  * Expanded coverage for prompt/HTTP actions, allowlisting, redirects, truncated responses, async/rewake behavior, and block-vs-continue rules.

### DIFF
--- a/internal/agent/hook_wiring_test.go
+++ b/internal/agent/hook_wiring_test.go
@@ -92,4 +92,8 @@ func TestDispatchHelpersAreNoopWithoutDispatcher(t *testing.T) {
 	if feedback := dispatchAfterTool(context.Background(), options, ToolCall{Name: "bash"}, nil, tools.Result{}); feedback != "" {
 		t.Fatalf("a nil dispatcher must yield no feedback, got %q", feedback)
 	}
+	if injected := dispatchUserPromptSubmit(context.Background(), options, "do the thing"); injected != "" {
+		t.Fatalf("a nil dispatcher must inject no userPromptSubmit context, got %q", injected)
+	}
+	dispatchStop(context.Background(), options) // must not panic with a nil dispatcher
 }
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -50,6 +50,10 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		return Result{}, errors.New("agent provider is required")
 	}
 
+	// stop hooks fire once as the run ends, on every exit path. A detached context
+	// lets end-of-run hooks still execute even if the run's own context is done.
+	defer dispatchStop(context.Background(), options)
+
 	maxTurns := options.MaxTurns
 	if maxTurns <= 0 {
 		maxTurns = 12
@@ -66,6 +70,15 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 	}
 
 	messages := zeroruntime.SeedMessagesWithImages(buildSystemPrompt(options), prompt, options.Images)
+
+	// userPromptSubmit hooks may inject additional context for this run (e.g.
+	// repository conventions); their output is appended once before the turn loop.
+	if injected := dispatchUserPromptSubmit(ctx, options, prompt); injected != "" {
+		messages = append(messages, zeroruntime.Message{
+			Role:    zeroruntime.MessageRoleUser,
+			Content: injected,
+		})
+	}
 
 	guards := newGuardState()
 	compactor := newCompactionState(options)
@@ -691,6 +704,43 @@ func dispatchAfterTool(ctx context.Context, options Options, call ToolCall, args
 		},
 	})
 	return strings.TrimSpace(strings.Join(outcome.Messages, "\n"))
+}
+
+// dispatchUserPromptSubmit runs userPromptSubmit hooks before the run's turn loop
+// and returns any hook output to inject into the model context for this run (e.g.
+// a hook that prepends repository conventions). A nil dispatcher is a no-op.
+func dispatchUserPromptSubmit(ctx context.Context, options Options, prompt string) string {
+	if options.Hooks == nil {
+		return ""
+	}
+	outcome := options.Hooks.Dispatch(ctx, hooks.DispatchInput{
+		Event: hooks.EventUserPromptSubmit,
+		Payload: map[string]any{
+			"event":     string(hooks.EventUserPromptSubmit),
+			"sessionId": options.SessionID,
+			"cwd":       options.Cwd,
+			"prompt":    prompt,
+		},
+	})
+	return strings.TrimSpace(strings.Join(outcome.Messages, "\n"))
+}
+
+// dispatchStop runs stop hooks as the run ends. It is observational: output is
+// recorded to the audit store but does not change the result. A nil dispatcher is
+// a no-op. Callers should pass a non-cancelled context so end-of-run hooks still
+// execute after the run's own context is done.
+func dispatchStop(ctx context.Context, options Options) {
+	if options.Hooks == nil {
+		return
+	}
+	options.Hooks.Dispatch(ctx, hooks.DispatchInput{
+		Event: hooks.EventStop,
+		Payload: map[string]any{
+			"event":     string(hooks.EventStop),
+			"sessionId": options.SessionID,
+			"cwd":       options.Cwd,
+		},
+	})
 }
 
 // blockedByHookResult is the tool result for a call vetoed by a beforeTool hook.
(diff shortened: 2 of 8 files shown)


gnanam1990:
Ran an adversarial multi-agent review (concurrency / security / loop-wiring / schema) over the diff before finishing. It confirmed 4 issues, all fixed TDD-first:

- **[high] SSRF / allowlist bypass** — the http action used `http.DefaultClient`, which follows redirects, so an allowlisted endpoint could 302 the request + payload to a non-allowlisted host. Now uses a no-redirect client.
- **[medium] asyncRewake on a killed hook** — the rewake gate didn't check `!TimedOut`, so a timed-out/killed hook that exited 2 (no real verdict) could spuriously wake the model. Now matches the synchronous afterTool block decision.
- **[medium] background hooks killed at run end** — async hooks were bound to the request context; detached via `context.WithoutCancel` so they outlive the turn (still bounded by the per-hook timeout).
- **[medium] unredacted injected context** — `userPromptSubmit` output went into the transcript without secret scrubbing; now redacted like the other hook-output paths.

Plus a Windows-determinism fix for the timed-out-rewake test. `go test ./...`, `go test -race ./internal/hooks/`, vet, build all green; all CI checks pass.

Vasanthdev2004 (changes requested):
## Verdict: Request changes

[P1] HTTP hook response read failures still fail open for `beforeTool` policy hooks.

`runHTTPAction` now detects a truncated/erroring response body, but it returns `commandResult{ExitCode: -1, Err: ...}` at `internal/hooks/actions.go:93-97`. `classifyResult` treats every non-nil `Err` as `AuditError, false` at `internal/hooks/dispatch.go:320-322`, even for `beforeTool`. So an allowlisted HTTP policy hook can receive headers, fail/reset while returning the verdict body, and the protected tool call still proceeds. That keeps the original safety issue alive for the gating case.

Please make the no-verdict HTTP path fail closed for blocking events, e.g. return a non-zero `ExitCode` + `Stderr` for response read failures so `beforeTool` blocks, or otherwise classify HTTP action read failures as blocking for `beforeTool`. Please add a regression test that uses a `beforeTool` HTTP hook with an erroring response body and asserts `outcome.Blocked == true`.

Validation I ran locally on `67c572a`:
- `git diff --check origin/feat/wedge-build...HEAD`
- `gofmt -l internal\agent internal\hooks`
- `go vet ./...`
- `go test ./...`

All validation commands passed; this is a behavior/safety blocker, not a build failure.

gnanam1990:
Pushed `bc6ef5f` for the [P1].

You're right — my earlier truncated-response fix returned `{ExitCode:-1, Err}`, and `classifyResult` maps any `Err` to a non-blocking `AuditError` (the fail-open path for a missing binary), so a `beforeTool` HTTP policy hook whose response read failed still let the tool proceed.

**Fix:** the HTTP no-verdict paths (transport failure *and* body-read failure) now return a non-zero `ExitCode` + `Stderr` instead of `Err`. `classifyResult` then blocks `beforeTool` (fail closed) while keeping it advisory for non-blocking events. I extended the fail-closed treatment to the transport-error path too (same no-verdict class), not just the read path.

**Regression test:** `TestDispatchHTTPActionBeforeToolReadFailureBlocks` — a `beforeTool` HTTP hook with an erroring response body now asserts `outcome.Blocked == true` (verified RED before the fix, GREEN after).

`go test ./...` + `go test -race ./internal/hooks/` green.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

Vasanthdev2004 (approved):
## Verdict: Approved

Rechecked the latest head `bc6ef5f`. The previous blocker is fixed: HTTP hook request/read failures now return a non-zero hook result instead of `Err`, so `beforeTool` HTTP policy hooks fail closed, and the regression test covers the truncated-response path.

Validation run locally:
- `git diff --check origin/feat/wedge-build...HEAD`
- `gofmt -l internal\agent internal\hooks`
- `go vet ./...`
- `go test ./...`

No blockers found.

gnanam1990:
@coderabbitai both inline items were fixed before that summary; they're outdated comments re-anchored from pre-fix commits. Please re-verify against HEAD (`bc6ef5f`):

**1. `internal/hooks/actions.go` — read error is captured, not dropped.** `runHTTPAction` now does:
```go
body, readErr := io.ReadAll(io.LimitReader(resp.Body, maxHTTPHookResponseBytes))
if readErr != nil {
    return commandResult{ExitCode: 1, Stderr: "read http hook response: " + readErr.Error()}
}
```
A mid-body reset returns `ExitCode: 1` (fail closed → blocks `beforeTool` via `classifyResult`), never `ExitCode: 0`. The transport-error path (`client.Do`) was given the same treatment. Fixed in `67c572a`/`bc6ef5f`; regression test `TestDispatchHTTPActionBeforeToolReadFailureBlocks`.

**2. `internal/hooks/hooks.go` — `validateHookURL` rejects cleartext http.** It now requires `https`, allowing `http` only for loopback hosts:
```go
case "http":
    if isLoopbackHost(parsed.Hostname()) { return "" }
    return "Expected an https URL (cleartext http is only allowed for loopback hosts)."
```
Fixed in `67c572a`; test case "http rejects cleartext to non-loopback host" in `TestActionTypeHooksValidation`.

The two deferred items (`dispatchAfterTool` not consuming `ContinueReason`/`EndTurn`, `Rewakes()` not consumed) are intentionally out of scope per the PR description. Could you resolve the two stale threads on re-verify?

Vasanthdev2004:
The previously requested HTTP hook fail-open issue is fixed in `bc6ef5f` and I approved the latest head after local validation.

Validation on `bc6ef5f`:
- `git diff --check origin/feat/wedge-build...HEAD`
- `gofmt -l internal\agent internal\hooks`
- `go vet ./...`
- `go test ./...`

@coderabbitai approve

Vasanthdev2004:
@coderabbitai approve

PR #182, opened 2026-06-12
wedge stage 11: GitHub Action + Slack/webhook notifier

gnanam1990:
## Stage 11 — GitHub Action + Slack/webhook notifier

Gives ZERO a distribution channel (a composite GitHub Action) and an output sink for unattended runs (a Slack/webhook notifier). Both wrap capabilities ZERO already has: `zero exec` with meaningful exit codes, and `internal/notify`.

### What was built

**`internal/notify` — sink abstraction + webhook sink**
- `notify.go`: new `Sink` interface and `Notifier` fan-out. `Notify` emits the terminal bell/OSC-9 sequence and also forwards every eligible event to attached sinks via `AddSink`. Sinks fire even when there is no terminal writer (headless CI), gated by the same mode/focus policy. Sinks are invoked outside the lock and a sink panic is isolated, so a slow or misbehaving sink can never stall or crash the run. `New(w, cfg)` is unchanged, so all existing callers keep working.
- `slack.go`: `WebhookSink` POSTs a JSON payload `{text, type, message, summary?, links?}` to a Slack incoming webhook or any generic webhook. `text` is what Slack renders; the structured fields carry machine-readable detail.
  - **Fail-soft:** a non-2xx response or a transport error is logged (redacted) and swallowed — a webhook problem never propagates as a fatal error.
  - **Redaction:** message, summary, links, and the webhook URL itself are run through the repo redaction before being sent or logged, so tokens never leak.
  - **Egress/proxy:** uses the default HTTP transport, which honors `HTTP_PROXY`/`HTTPS_PROXY`; a sandboxed run's scoped-egress proxy (and any deny of the webhook host) is respected.
  - An empty URL yields an inert sink, so callers can wire it unconditionally and let configuration decide whether it fires.

**`action.yml` — composite GitHub Action**
- Wraps `zero exec`. Installs the pinned ZERO release using the repo's existing checksum-verifying `scripts/install.sh` (bundled with the action via `github.action_path`), runs the prompt in the checked-out repo, surfaces ZERO's exit code as the step status, captures stdout to a file (exposed as the `output-file` output), and can post a one-line summary to the triggering PR (`gh`) or to Slack.
- **Provider-agnostic:** `provider`, `api-key`, and `api-key-env` are inputs — no provider is hardcoded as a default. The key is exported under the env name the chosen provider expects.
- **Safety defaults for CI:** `auto` defaults to `low`; the sandbox is always active; the action never passes `--skip-permissions-unsafe`. Secrets are passed via the environment and never echoed.
- Inputs: `prompt`/`prompt-file`, `provider`, `api-key`, `api-key-env`, `model`, `mode`, `auto`, `self-correct`, `add-dir`, `worktree`, `output-format` (default `stream-json`), `post-to`, `slack-webhook-url`, `github-token`, `working-directory`, `zero-version`, `zero-repo`. Outputs: `exit-code`, `output-file`, `summary`.

**`docs/GITHUB_ACTION.md`** — input/output tables, copy-paste workflows ("run ZERO on every issue labeled `zero`" and "nightly dependency-upgrade PR"), security notes, and the notifier docs.

**`.github/workflows/zero-action-smoke.yml`** — the spec's suggested smoke vehicle: parses `action.yml`, asserts the documented inputs/outputs, verifies the conservative defaults (`auto=low`, `output-format=stream-json`) and that unsafe mode is never enabled, and shell-syntax-checks the embedded run blocks.

### Files
- `internal/notify/notify.go` (modified) — `Sink`, `AddSink`, fan-out
- `internal/notify/notify_test.go` (modified) — fan-out, fail-soft, focus-policy, race tests
- `internal/notify/slack.go` (new) — `WebhookSink`
- `internal/notify/slack_test.go` (new) — webhook POST/JSON, non-2xx, transport error, redaction, empty-URL tests
- `action.yml` (new)
- `docs/GITHUB_ACTION.md` (new)
- `.github/workflows/zero-action-smoke.yml` (new)

### How it was tested
- TDD (RED → GREEN) for the Go parts. `go test ./...` is green (58 packages); `internal/notify` is also race-clean (`go test -race`) and `go vet`-clean.
- `gofmt -l .` is clean; `go build ./...` succeeds.
- `action.yml` and the smoke workflow were validated locally (YAML parses; all documented inputs/outputs present; every embedded `run:` block passes `bash -n`).

### Deferred / out of scope
- The optional Docker-based action variant (`.github/actions/zero/` Dockerfile) — the composite action was chosen as it reuses the existing release installer and needs no image build/publish.
- Wiring the webhook sink into a live run's completion path: the spec referenced a Stage 05 `EventNotification` hook that is not present in this branch, and run-loop wiring lives outside the four scoped files. The sink and its `Notifier` integration are complete and unit-tested; the Action provides the unattended-run Slack path today via `post-to: slack`.

## Summary by CodeRabbit

* **New Features**
  * ZERO GitHub Action for running prompts in CI with configurable inputs/outputs and optional PR/Slack summaries
  * In-app notification sinks including a webhook (Slack) sink for external delivery of run events

* **Bug Fixes / Reliability**
  * Notifications fan out safely: sink failures or panics won’t crash runs; delivery is concurrency-safe

* **Documentation**
  * Full docs for the GitHub Action with examples and security guidance

* **Tests**
  * Added CI smoke workflow to validate the action and shell-script syntax; comprehensive notification/webhook tests

### DIFF
--- a/.github/workflows/zero-action-smoke.yml
+++ b/.github/workflows/zero-action-smoke.yml
@@ -1,0 +1,87 @@
+name: ZERO Action Smoke
+
+on:
+  pull_request:
+    paths:
+      - action.yml
+      - .github/workflows/zero-action-smoke.yml
+  push:
+    branches:
+      - main
+    paths:
+      - action.yml
+      - .github/workflows/zero-action-smoke.yml
+
+jobs:
+  validate:
+    name: Validate action.yml
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Parse action.yml and assert documented inputs
+        shell: bash
+        run: |
+          python3 - <<'PY'
+          import sys, yaml
+
+          with open("action.yml", "r", encoding="utf-8") as handle:
+              doc = yaml.safe_load(handle)
+
+          assert isinstance(doc, dict), "action.yml must be a mapping"
+          assert doc.get("name"), "action must declare a name"
+          assert doc.get("description"), "action must declare a description"
+
+          runs = doc.get("runs") or {}
+          assert runs.get("using") == "composite", "action must be a composite action"
+          steps = runs.get("steps") or []
+          assert len(steps) >= 2, "composite action must define steps"
+
+          inputs = doc.get("inputs") or {}
+          expected_inputs = [
+              "prompt", "prompt-file", "provider", "api-key", "api-key-env",
+              "model", "mode", "auto", "self-correct", "add-dir", "worktree",
+              "output-format", "post-to", "slack-webhook-url", "github-token",
+              "working-directory", "zero-version", "zero-repo",
+          ]
+          missing = [name for name in expected_inputs if name not in inputs]
+          assert not missing, f"action.yml is missing inputs: {missing}"
+
+          # Conservative CI default: autonomy must default to 'low'.
+          assert inputs["auto"].get("default") == "low", "auto must default to low"
+          # stream-json is the documented default capture format.
+          assert inputs["output-format"].get("default") == "stream-json", \
+              "output-format must default to stream-json"
+
+          outputs = doc.get("outputs") or {}
+          for name in ("exit-code", "output-file", "summary"):
+              assert name in outputs, f"action.yml is missing output: {name}"
+
+          # The action must never enable unsafe mode implicitly.
+          serialized = yaml.safe_dump(doc)
+          assert "--skip-permissions-unsafe" not in serialized, \
+              "action must never pass --skip-permissions-unsafe"
+
+          print(f"action.yml OK: {len(inputs)} inputs, {len(outputs)} outputs, {len(steps)} steps")
+          PY
+
+      - name: Shell-syntax-check embedded run blocks
+        shell: bash
+        run: |
+          set -euo pipefail
+          python3 - <<'PY'
+          import yaml
+          with open("action.yml", "r", encoding="utf-8") as handle:
+              doc = yaml.safe_load(handle)
+          steps = (doc.get("runs") or {}).get("steps") or []
+          for index, step in enumerate(steps):
+              run = step.get("run")
+              if run:
+                  with open(f"/tmp/zero_step_{index}.sh", "w", encoding="utf-8") as out:
+                      out.write(run)
+          PY
+          for script in /tmp/zero_step_*.sh; do
+            echo "bash -n ${script}"
+            bash -n "${script}"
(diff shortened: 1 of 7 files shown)


gnanam1990:
Addressed the CodeRabbit review for stage 11.

Fixed:
- action.yml: reject `prompt` and `prompt-file` when both are set (fast fail, exit 2) instead of silently preferring the file.
- action.yml: `add-dir` entries now trim only leading/trailing whitespace, preserving internal spaces in paths (was `tr -d '[:space:]'`).
- action.yml: export the provider API key in-process (with env-name validation) so the same `run` block's `zero` invocation sees it; still appended to GITHUB_ENV for later steps.
- action.yml: a commit-SHA `github.action_ref` is treated as not-a-release-tag and falls back to `latest` (release artifacts are keyed by tag, so `releases/download/v<sha>/` would 404).
- internal/notify/slack.go: redact the configured webhook URL in the outbound payload fields (message/summary/links), matching the existing log() behavior, so a URL token is never sent verbatim. Added a covering test.
- internal/notify/slack.go: copy the caller-owned `Links` and `ExtraSecrets` slices in `NewWebhookSink` so later config mutation can't alter or race a payload. Added a covering test.

Skipped:
- Forbidden `--skip-permissions-unsafe` literal in the action.yml comment block: already removed in 214940e (the comment now reads "never enables unsafe permission-skipping"); the smoke serialized-doc assertion passes.
- Pin `actions/checkout` to a commit SHA + `persist-credentials: false` in the smoke workflow: skipped as out of scope for this stage. It is a repo-wide CI-hardening (zizmor blanket policy) concern — every workflow in the repo uses the unpinned `@v4` tag, so changing only this one file would be inconsistent; better handled as a dedicated repo-wide pinning change.

Gate: `gofmt -l` clean, `go build ./...` OK, `go test ./...` green.

Vasanthdev2004 (approved):
## Verdict: Approved

Reviewed the GitHub Action + Slack/webhook notifier changes. No blockers found.

What I checked:
- Action inputs keep prompt/prompt-file validation, avoid leaking API keys in logs, and route through the release binary path.
- Slack/webhook sinks soft-fail, redact webhook/payload data, bound error bodies, and copy config slices defensively.
- Existing CodeRabbit request looks stale after the latest fixes; current CodeRabbit status and CI are green.

- `go test ./...`

gnanam1990:
@coderabbit review if appprovable then approve it

gnanam1990:
@coderabbit approve

PR #183, opened 2026-06-12
wedge stage 12: PDF ingestion

gnanam1990:
## What this builds

Stage 12 of the wedge: PDF ingestion in `internal/imageinput`. ZERO already accepts image input; specs, design docs, and tickets routinely arrive as PDF, so this adds the ability to attach a PDF the same way an image is attached. It extracts the text layer and, for vision models, optionally rasterizes pages to images.

## Dependency posture (the important decision)

ZERO's promise is a single static Go binary with no runtime deps and clean cross-compilation, so:

- **Default = pure-Go text extraction** via `github.com/ledongthuc/pdf` (BSD-licensed, no CGO, no transitive deps). Text-only, but keeps the static binary intact.
- **Optional external tools** — poppler's `pdftotext` / `pdftoppm` are used for richer extraction and page rasterization **only when present on PATH**, the same "external tool the user may have" posture as the LSP language servers. When absent, extraction silently degrades to the pure-Go text path; absence is never an error.
- **No CGO anywhere.** Confirmed `CGO_ENABLED=0` cross-compiles for linux/amd64, windows/arm64, and darwin/arm64.

## Behavior

- **Detection by magic bytes** (`%PDF-`), never by extension alone — a `.pdf`-named non-PDF is rejected with a clear error.
- **Caps:** a 32 MiB raw-file cap (rejects, bounded read so an unbounded source can't allocate without limit) and a 256 KiB extracted-text cap (truncates with a marker on a UTF-8 boundary, so a large-but-valid doc stays partially usable).
- **Malformed-input safety:** the pure-Go parser can panic on broken structures, so extraction is wrapped in a `recover` that turns a bad PDF into a clean error rather than a crash.
- **Scanned PDFs:** no text layer and no rasterizer returns an explicit `"no extractable text; OCR is not available"` error rather than silent empty success.
- **Vision path:** with a vision model and `pdftoppm` available, the first N pages (default 10, bounded so context can't blow up) render to PNG and flow through the existing image pipeline (`NormalizeImageMediaType`, the per-image cap, the allow-list). Without a rasterizer it degrades to the text layer.

## Integration point

`/image <file.pdf>` is routed to the document path (following the existing `/image` slash-command wiring). The text layer attaches for **any** model (it is just text) and is prepended to the next prompt as a labeled preamble; rendered pages attach as images for vision models via the existing pending-image pipeline. The pending-attachment chip row shows `[doc: …]` entries and `/image clear` clears staged documents too.

## Files

- `internal/imageinput/pdf.go` (new) — `LoadDocument`, `isPDF`, `IsProbablyDocumentPath`, pure-Go + optional-poppler extraction and rasterization.
- `internal/imageinput/pdf_test.go` (new) — magic-byte detection, text extraction on a generated fixture, both size caps, the no-text/no-raster message, malformed-input safety, and pure-Go fallback.
- `internal/tui/image_attach.go`, `internal/tui/commands.go`, `internal/tui/model.go` — the `/image` document-attach integration (staging, prompt preamble, chips, clear).
- `internal/tui/image_attach_test.go` — attach on non-vision model, fake-PDF rejection, clear, chip rendering, and a submit test asserting the document text reaches the agent prompt alongside the user question.
- `go.mod` / `go.sum` — the single pure-Go dependency (no transitive deps).

`internal/imageinput/imageinput.go` was intentionally **not** modified: `LoadDocument` is a sibling entrypoint and routing happens at the `/image` layer, which is cleaner than overloading `LoadFile`.

## How it was tested

- `go build ./...`, `go test ./...`, and `gofmt -l` over the changed dirs are all clean; `go vet` and `go test -race ./internal/imageinput/` pass.
- `CGO_ENABLED=0` cross-compile verified for three OS/arch targets.
- Fixtures are generated in-test (a minimal valid PDF with a real xref table) rather than committed as opaque binaries.

## Deferred (out of scope)

- OCR for image-only/scanned PDFs (the explicit no-text message points at poppler).
- Wiring PDFs into the CLI `--image` flag (named for images; a separate document flag + exec prompt-prepending would be a broader change than the `/image` slash-command scope this stage targets).
- A known limitation of the pure-Go library: on certain malformed dictionaries it writes a stray `DEBUG:` line to stdout before the (recovered) panic; this is inherent to the upstream library and only fires on malformed input.

## Summary by CodeRabbit

* **New Features**
  * `/image` now accepts PDFs (by content or extension); extracted text is staged and prepended to prompts.
  * Vision-enabled flows may render and attach PDF page images when available; `/image clear` clears both images and documents.
  * Large PDFs are truncated with a marker when needed; invalid or non-PDF uploads are rejected with notice.

* **Tests**
  * Added extensive tests for PDF detection, extraction, truncation, attachment rendering, and error paths.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.1-0.20250404203927-76690c660834
 	github.com/charmbracelet/x/ansi v0.11.6
+	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
 	golang.org/x/sys v0.38.0
 )
 
--- a/go.sum
+++ b/go.sum
@@ -24,6 +24,8 @@ github.com/clipperhouse/uax29/v2 v2.5.0 h1:x7T0T4eTHDONxFJsL94uKNKPHrclyFI0lm7+w
 github.com/clipperhouse/uax29/v2 v2.5.0/go.mod h1:Wn1g7MK6OoeDT0vL+Q0SQLDz/KpfsVRgg6W7ihQeh4g=
 github.com/erikgeiser/coninput v0.0.0-20211004153227-1c3628e74d0f h1:Y/CXytFA4m6baUTXGLOoWe4PQhGxaX0KpnayAqC48p4=
 github.com/erikgeiser/coninput v0.0.0-20211004153227-1c3628e74d0f/go.mod h1:vw97MGsxSvLiUE2X8qFplwetxpGLQrlU1Q9AUEIzCaM=
+github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728 h1:QwWKgMY28TAXaDl+ExRDqGQltzXqN/xypdKP86niVn8=
+github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728/go.mod h1:1fEHWurg7pvf5SG6XNE5Q8UZmOwex51Mkx3SLhrW5B4=
 github.com/lucasb-eyer/go-colorful v1.3.0 h1:2/yBRLdWBZKrf7gB40FoiKfAWYQ0lqNcbuQwVHXptag=
 github.com/lucasb-eyer/go-colorful v1.3.0/go.mod h1:R4dSotOR9KMtayYi1e77YzuveK+i7ruzyGqttikkLy0=
 github.com/mattn/go-isatty v0.0.20 h1:xfD0iDuEKnDkl03q4limB+vH+GxLEtL/jb4xVJSWWEY=
--- a/internal/imageinput/pdf.go
+++ b/internal/imageinput/pdf.go
@@ -1,0 +1,379 @@
+package imageinput
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+	"github.com/ledongthuc/pdf"
+)
+
+// Dependency posture (see stage 12): the DEFAULT build extracts a PDF's text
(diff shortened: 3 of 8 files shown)


gnanam1990:
Addressed the review for stage 12 (PDF ingestion):

**Fixed**
- **`Document.Pages` was 0 when `pdftotext` won the text path.** `pdftotext` reports no page count, so the poppler-text branch left `pages == 0`. Added a `pdfPageCount` helper (pure-Go reader, structural read only, panic-recovered) and assigned it in the poppler branch so `Pages` is correct regardless of which extractor wins. (`internal/imageinput/pdf.go`)
- **`capDocumentText` could exceed the 256 KiB cap.** It sliced to `MaxDocumentTextBytes` then appended the truncation marker on top. Now the marker is counted against the cap (`cut = MaxDocumentTextBytes - len(marker)`, clamped at 0), so the returned payload never exceeds the advertised limit. Tightened the existing truncation test to assert the strict bound. (`internal/imageinput/pdf.go`)
- **`/image` routed PDFs by extension only.** A real PDF whose path lacks `.pdf` was sent down the image path (refused on non-vision models). Added `LooksLikeDocumentFile` (reads only the magic prefix) and route on the extension hint OR a content sniff, before the vision gate, so content-based detection holds. (`internal/imageinput/pdf.go`, `internal/tui/image_attach.go`)

Each fix has a focused test (`TestPDFPageCount`, `TestCapDocumentTextRespectsCap`, `TestLooksLikeDocumentFile`, `TestImageCommandAttachesExtensionlessPDFByContent`).

gofmt clean, `go build ./...` and the changed packages' tests pass.

Vasanthdev2004 (approved):
## Verdict: Approved

Reviewed the PDF ingestion path. No blockers found.

What I checked:
- PDF detection uses magic bytes where needed, not only extension checks.
- Raw PDF and extracted text are capped, raster extraction is bounded and optional, and poppler subprocess calls are timeout-bounded.
- TUI attachment flow degrades to text-only on non-vision models and only adds raster images when supported.
- Tests cover fake PDFs, size caps, content sniffing, extraction, and no-raster degradation.

- `go test ./...`

PR #184, opened 2026-06-12
wedge stage 10: distribution (skill/plugin install, tools make)

gnanam1990:
## Stage 10 — Distribution

ZERO had skills/plugins but no way to install them (skills were discovery-only from a local dir) and no tool scaffolder. This adds a **file-backed, local** distribution surface so extensibility compounds into an ecosystem. There is no central registry: a "source" is just a git URL or a local path, and every install is checksum-pinned in a lockfile.

### What was built

**`internal/skills/install.go`** — `Install`, `Remove`, `Info`, lockfile (`skills.lock`).
- Fetches a skill from a git URL (shallow clone via an injectable runner) or a local path, locates the `SKILL.md` (root or a single subdir), and validates its frontmatter through the existing skills parser. Invalid/nameless skills are rejected and nothing is written.
- Copies the `SKILL.md` into `skills.DefaultDir(env)/<name>/`, records a `sha256` content hash mapped to its source in `skills.lock`. Reinstalling surfaces the old→new hash change.
- A name clash with a **different** source is refused unless `--force`; reinstalling from the same source is idempotent.

**`internal/plugins/install.go`** — `Install`, `Remove`, lockfile (`plugins.lock`).
- Same fetch + pin flow, validated against the existing plugin manifest schema (`ParseManifest`). The whole plugin tree is copied (entry scripts, prompts, skills) so the plugin is runnable through activation — `.git` and symlinks are skipped so clone metadata and link-escapes never land in the plugins dir.
- Installed plugins still go through normal activation with permission gating; no install script is run.

**`internal/tools/scaffold.go`** — `Scaffold`.
- Generates a minimal, prompt-gated plugin-tool skeleton in the toolbox dir: a `plugin.json` (schemaVersion 1, one tool, empty parameter schema, command pointing at the entry stub) plus a runnable entry stub carrying a clear TODO. Default runtime is a POSIX shell stub (no interpreter dependency); `node`/`python` are also available. After plugin activation the scaffolded tool is loadable like any plugin.

**CLI wiring**
- `zero skill {add,info,remove}` — extended `internal/cli/skills.go`.
- `zero plugin {add,remove}` — extended `internal/cli/extensions.go`.
- `zero tools {make,list}` — new command registered in `internal/cli/app.go` (dispatch + help), handlers in the new `internal/cli/distribution.go`.
- New `appDeps` resolvers `pluginsDir`/`toolsDir` (the user plugins root) added in `app.go`.

### Safety properties
- **No opaque remote code without a fingerprint:** every install records a content hash in a lockfile; updates show the hash change.
- **Never executes fetched content:** skills are markdown (validated, never run); plugin install scripts are never executed; the scaffolder only writes files. Tests assert a hostile `install.sh` in a fetched source does not run and is not copied into the skills dir.
- **Network:** git fetch inherits the process environment (so proxy/egress settings are honored) and sets `GIT_TERMINAL_PROMPT=0` so a missing-credential clone fails fast instead of blocking.

### Files
- `internal/skills/install.go` + `internal/skills/install_test.go`
- `internal/plugins/install.go` + `internal/plugins/install_test.go`
- `internal/tools/scaffold.go` + `internal/tools/scaffold_test.go`
- `internal/cli/distribution.go` + `internal/cli/distribution_test.go`
- `internal/cli/app.go`, `internal/cli/extensions.go`, `internal/cli/skills.go` (subcommand wiring + help)

### How it was tested
TDD (RED→GREEN). Per-package unit tests cover the spec's list: local install + hash recording, invalid-source rejection, reinstall hash change + lockfile update, name-clash guard (+`--force`), manifest validation, scaffold manifest parses against the real schema, list/info/remove reflect installed state, and "install never executes fetched content". Added real-git integration tests (a local `file://` repo exercising the default runner end to end, asserting `.git` does not leak) and a provider-neutrality assertion on the scaffolded manifest. End-to-end smoke through the built binary confirms the full add/list/info/update/remove and make/list lifecycle.

Gate: `gofmt -l .` clean, `go build ./...` succeeds, `go test ./...` passes.

### Deferred / out of scope
- Release-tarball fetch (the spec listed it as an alternative to shallow clone; clone is implemented and the runner is injectable).
- Merging plugin-declared skill paths into the skills loader and runtime registration of installed plugin tools remain the separately-tracked Stage 09 / multi-root work, unchanged here.

## Summary by CodeRabbit

* **New Features**
  * New top-level CLI commands: tools (make, list), skill (add, info, remove), and plugin (add, remove); help updated.
  * Tool scaffolding creates runnable stubs (shell/Node/Python) and a plugin manifest.
  * CLI now respects user-scoped plugins/tools directory defaults.

* **Behavior**
  * Installs/removals record deterministic content hashes and lockfiles; force/overwrite rules enforced; list output supports human and --json modes and surfaces loader diagnostics.

* **Tests**
  * Extensive end-to-end and unit tests for CLI, install, scaffold flows.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -45,6 +45,8 @@ type appDeps struct {
 	loadPlugins          func(plugins.LoadOptions) (plugins.LoadResult, error)
 	loadHooks            func(hooks.LoadOptions) (hooks.LoadResult, error)
 	skillsDir            func() string
+	pluginsDir           func() string
+	toolsDir             func() string
 	newMCPStore          func() (*mcp.PermissionStore, error)
 	newMCPTokenStore     func() (*mcp.TokenStore, error)
 	newSandboxStore      func() (*sandbox.GrantStore, error)
@@ -111,6 +113,10 @@ func defaultAppDeps() appDeps {
 		skillsDir: func() string {
 			return skills.DefaultDir(nil)
 		},
+		pluginsDir: defaultUserPluginsDir,
+		// Scaffolded tools are plugins, so the toolbox dir is the user plugins root:
+		// after activation a `tools make` skeleton is discovered like any plugin.
+		toolsDir: defaultUserPluginsDir,
 		newMCPStore: func() (*mcp.PermissionStore, error) {
 			return mcp.NewPermissionStore(mcp.StoreOptions{})
 		},
@@ -140,6 +146,24 @@ func defaultAppDeps() appDeps {
 
 func userAgent() string {
 	return "zero/" + version
+}
+
+// defaultUserPluginsDir resolves the user-scoped plugins root
+// ($XDG_CONFIG_HOME/zero/plugins) used as the install target for `plugin add`
+// and the toolbox for `tools make`. It is the SourceUser root from
+// plugins.ResolveRoots; an empty string is returned only if it cannot be
+// resolved, which the command layer surfaces as an error.
+func defaultUserPluginsDir() string {
+	roots, err := plugins.ResolveRoots(plugins.ResolveRootOptions{})
+	if err != nil {
+		return ""
+	}
+	for _, root := range roots {
+		if root.Source == plugins.SourceUser {
+			return root.Path
+		}
+	}
+	return ""
 }
 
 func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) (exitCode int) {
@@ -241,6 +265,8 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 		return runPlugins(args[1:], stdout, stderr, deps)
 	case "skills", "skill":
 		return runSkills(args[1:], stdout, stderr, deps)
+	case "tools", "tool":
+		return runTools(args[1:], stdout, stderr, deps)
 	case "hooks":
 		return runHooks(args[1:], stdout, stderr, deps)
 	case "mcp":
@@ -310,6 +336,12 @@ func fillAppDeps(deps appDeps) appDeps {
 	}
 	if deps.skillsDir == nil {
 		deps.skillsDir = defaults.skillsDir
+	}
+	if deps.pluginsDir == nil {
+		deps.pluginsDir = defaults.pluginsDir
+	}
+	if deps.toolsDir == nil {
+		deps.toolsDir = defaults.toolsDir
 	}
 	if deps.newMCPStore == nil {
 		deps.newMCPStore = defaults.newMCPStore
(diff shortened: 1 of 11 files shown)


gnanam1990:
Addressed the review (verified each against current code; gate green: gofmt clean, `go build ./...` and `go test ./...` pass).

Fixed:
- **plugins/install.go** — install now hashes the same filtered tree `copyTree` installs (path + exec bit + bytes, `.git`/symlinks skipped, sorted), not just `plugin.json`, so a change to any nested file (tool script, prompt, bundled skill) is reflected in the lock hash and reported as an update. Added a regression that changes only a nested file and asserts `Updated` + hash change.
- **cli/distribution.go** — `tools list` now surfaces `plugins.Load` diagnostics in both text ("Tool diagnostics:") and JSON (`diagnostics` field), matching `plugins list`, so a broken toolbox plugin no longer vanishes silently. Added a text+JSON test.
- **cli/distribution.go** — `skill remove` / `plugin remove` now reject `--json` with a usage error instead of silently ignoring it. Added a test.
- **cli/distribution.go** — post-scaffold hint now points at `zero tools list` (the command that proves the scaffolded tool loads), not `zero plugins list`.
- **skills/install.go + plugins/install.go** — local install sources are canonicalized (abs + symlink-resolved) before storing/comparing, so the same local source via relative vs absolute spelling is no longer a false clash. Added regression tests in both packages.
- **tools/scaffold_test.go** — added a table-driven test covering all three runtimes (shell/node/python) so manifest/template drift in the non-default branches is caught.

Skipped:
- Staged-swap reinstall (skills + plugins) — skipped to keep the change minimal. A correct, litter-free staged swap requires either skipping dot-dirs in the loader (out of scope; loaders aren't in this PR's changed set) or stale-temp cleanup logic; staging under the scanned dir without that would add a new crash-path failure mode (a leftover temp dir read as a duplicate plugin/skill). The current in-place flow already clears-then-writes so it never mixes old/new files.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

Vasanthdev2004 (approved):
## Verdict: Approved

Reviewed the skill/plugin distribution and tool scaffold changes. No blockers found.

What I checked:
- Plugin and skill installs validate manifests/frontmatter before writing install state.
- Local sources are canonicalized for clash detection; git/local installs never execute fetched content.
- Plugin hashing covers installed tree content and skips `.git`/symlinks; skill hashing covers the installed `SKILL.md` content.
- Remove paths validate IDs/names before deleting install targets.
- Scaffolded tools default to prompt-gated permissions and generate data only.

Note: the non-staged install replacement tradeoff is acceptable for this PR after the latest discussion/fixes; it does not introduce execution of fetched content, and CI/CodeRabbit are green.

- `go test ./...`

gnanam1990:
@coderabbit review..if approvable then approve it

gnanam1990:
@coderabbit approve

PR #185, opened 2026-06-12
wedge stage 09: finish plugin wiring (tools/hooks/skills)

gnanam1990:
## Summary

Finishes the previously-stubbed `internal/plugins` so loaded plugin manifests actually take effect at runtime. Before this change, plugin tools/hooks/skills were parsed for discovery only (the `zero plugins` listing) and never registered — plugins loaded but did nothing. This stage turns each plugin's declared **tools**, **hooks**, and **skills** into live registrations.

## What was built

**`internal/plugins/activate.go` (new)** — `Activate(registry, plugins, options)`:
- **Tools → registry:** a `pluginTool` adapter implements the `tools.Tool` interface by invoking the plugin's declared command. Tool arguments are passed as JSON on stdin; stdout/exit map onto a `tools.Result`; `${AGENT_PLUGIN_ROOT}` placeholders in the command/args are expanded to the plugin dir, which is also exported as the `AGENT_PLUGIN_ROOT` env var. The runner is injectable for tests and defaults to a bounded subprocess. Each tool is registered via `registry.Register`.
- **Permission mapping:** `ToolPermission` (allow/prompt/deny) maps onto `tools.Safety` (always `shell` side-effect + the carried permission). `allow` only survives manifest parsing when `AllowManifestToolAutoApproval` was set, so mutating plugin tools are **not** auto-approved by default; `deny` tools register but are never advertised or executed.
- **Hooks → dispatcher:** builds an `internal/hooks.Definition` per hook extension (mapping the plugin `HookEvent` to the hooks `Event`, carrying command/args, `${AGENT_PLUGIN_ROOT}`-expanded), with plugin-namespaced ids.
- **Skills → loader:** exposes each plugin skill's search root (the directory `internal/skills.Load` scans) and provides a plugin-aware `skill` tool that merges the default skills dir with plugin roots, deduplicating by name and recording collisions (preserving `skills.Duplicates` behaviour, never crashing).
- **Isolation:** a malformed plugin/extension is skipped with a recorded warning; a single bad plugin never aborts activation. Activation order is deterministic (plugins sorted by id). Provenance (originating plugin id) is recorded for tools/hooks/skills and tagged in tool-result `Meta`.

**`internal/cli/plugin_activate.go` (new) + bootstrap wiring** — `activatePlugins(...)` loads the workspace's plugins and runs activation. It is invoked in **both** bootstrap paths:
- interactive TUI: `internal/cli/app.go` in `runInteractiveTUIWithSetup`, right after specialist + MCP tool registration (so plugin tools are part of the deferral count).
- one-shot exec: `internal/cli/exec.go` in `runExec`, after MCP registration and before `--list-tools`/filter validation (so plugin tools are listable and filter-validatable).

`newHookDispatcherWithExtra` (in `internal/cli/hook_dispatch.go`) folds the plugin hook definitions into the dispatcher's active hook set; the plugin-aware skill tool is registered when a plugin declares skills. Activation **fails open**: a load error or malformed plugin is surfaced as a `[zero] WARNING` on stderr and skipped, never wedging startup.

**`internal/plugins/plugins.go`** — comment-only update replacing the stale "not yet registered" note (no behavioural change).

## Where activation is invoked
- `internal/cli/app.go`: `activatePlugins(workspaceRoot, registry, deps, stderr)` in `runInteractiveTUIWithSetup`, and `newHookDispatcherWithExtra(workspaceRoot, pluginActivation.hooks)` in the TUI `agent.Options`.
- `internal/cli/exec.go`: `activatePlugins(...)` in `runExec`, and `newHookDispatcherWithExtra(...)` in the exec `agent.Options`.

## Testing
- `internal/plugins/activate_test.go`: tool registration + command invocation + stdout→Result mapping (stubbed runner); non-zero exit → error; permission mapping (prompt requires grant, allow runs, deny never executes); hook `Definition` build + event mapping + dispatcher `Select`; skill search-root derivation + loader discovery; merged skill list + duplicate recording; malformed-plugin skip-with-warning; disabled-plugin skip; deterministic ordering.
- `internal/plugins/activate_unix_test.go`: real-subprocess end-to-end (JSON args on stdin and `$AGENT_PLUGIN_ROOT` round-trip; missing binary → tool error).
- `internal/cli/plugin_activate_test.go`: bootstrap helper registers the plugin tool, collects namespaced hooks, re-registers the plugin-aware skill tool, fails open on load error, warns on malformed plugin; `newHookDispatcherWithExtra` runs a folded plugin hook.
- Gate: `gofmt -l` clean on changed dirs, `go build ./...` and `go test ./...` both green.

## Deferred / out of scope
- Plugin **prompts** remain discovery-only (the spec scoped this stage to tools/hooks/skills).
- No new config surface for manifest tool auto-approval was added; `AllowManifestToolAutoApproval` stays off by default at the bootstrap, so plugin `allow` tools are clamped to prompt unless a caller opts in.
- Plugin tools are advertised eagerly (they do not opt into the deferred-tool/`tool_search` mechanism).

## Summary by CodeRabbit

* **New Features**
  * Plugins now activate automatically during CLI interactive and execution runs; plugin-provided hooks are merged into the CLI hook dispatcher while preserving workspace hook config.
  * Plugin tools and skills are dynamically registered and made available at runtime; activation emits warnings for load/activation issues without blocking runs.

* **Tests**
  * Comprehensive unit and end-to-end tests added to validate plugin activation, tool execution, hook wiring, skill merging, and subprocess behavior.

* **Documentation**
  * Clarified manifest parsing docs to distinguish discovery vs activation consumers.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -412,6 +412,11 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		return writeAppError(stderr, err.Error(), 1)
 	}
 	defer closeMCPRuntime(stderr, mcpRuntime)
+	// Make local plugins live: register their declared tools into the registry and
+	// collect their hooks + skill roots for the dispatcher and skill tool below.
+	// Done after specialist + MCP registration so plugin tools are part of the
+	// deferral count, and it fails OPEN — a malformed plugin is warned and skipped.
+	pluginActivation := activatePlugins(workspaceRoot, registry, deps, stderr)
 	// Ask (not Auto) is the interactive default: in Auto, ToolAdvertised exposes
 	// only PermissionAllow tools, so prompt-gated tools (write_file/edit_file/bash/
 	// apply_patch) would never be offered to the model — the TUI could neither edit
@@ -428,10 +433,10 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	}
 	// Activate deferred MCP-tool loading for the interactive run only when the
 	// VISIBLE deferred-eligible count meets the resolved threshold, matching exec.
-	// The registry is complete (core + specialist + MCP) here, so the count is
-	// accurate; below threshold this is a no-op and the surface is unchanged. The
-	// interactive surface applies no operator tool filters, so enabled/disabled are
-	// nil — matching the AgentOptions below.
+	// The registry is complete (core + specialist + MCP + plugins) here, so the
+	// count is accurate; below threshold this is a no-op and the surface is
+	// unchanged. The interactive surface applies no operator tool filters, so
+	// enabled/disabled are nil — matching the AgentOptions below.
 	registerToolSearchIfEligible(registry, resolved.Tools.DeferThreshold, permissionMode, nil, nil)
 	sandboxStore, err := deps.newSandboxStore()
 	if err != nil {
@@ -463,7 +468,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 			Autonomy:       string(sandbox.AutonomyLow),
 			Sandbox:        sandboxEngine,
 			FileTracker:    tools.NewFileTracker(),
-			Hooks:          newHookDispatcher(workspaceRoot),
+			Hooks:          newHookDispatcherWithExtra(workspaceRoot, pluginActivation.hooks),
 			DeferThreshold: resolved.Tools.DeferThreshold,
 		},
 		PermissionMode: permissionMode,
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -177,6 +177,11 @@ func runExec(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 		return writeExecProviderError(stdout, stderr, options.outputFormat, "mcp_error", err.Error())
 	}
 	defer closeMCPRuntime(stderr, mcpRuntime)
+	// Make local plugins live for this run: register their declared tools into the
+	// registry and collect their hooks + skill roots for the dispatcher and skill
+	// tool below. Done before --list-tools and filter validation so plugin tools
+	// are listable and filter-validatable; it fails OPEN (a bad plugin is skipped).
+	pluginActivation := activatePlugins(workspaceRoot, registry, deps, stderr)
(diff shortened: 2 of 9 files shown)


gnanam1990:
Addressed the CodeRabbit review (pushed 4321b09):

**Fixed**
- **Surface load diagnostics** (`internal/cli/plugin_activate.go`): `activatePlugins` now forwards each `LoadResult.Diagnostics` entry to stderr, so a plugin dropped for a malformed manifest is no longer silent (only the top-level error was reported before). Added `TestActivatePluginsSurfacesLoadDiagnostics`.
- **Reject colliding tool names** (`internal/plugins/activate.go`): `activateTools` checks the registry before registering and skips a plugin tool whose name is already taken (e.g. a core tool), recording a warning instead of silently overwriting via last-wins `Register`. Kept the file's isolation-first warn-and-skip behavior rather than aborting activation. Added `TestActivateSkipsToolWhoseNameCollidesWithRegisteredTool`.
- **Resolve skill paths against the plugin dir** (`internal/plugins/activate.go`): `activateSkills` now expands `${AGENT_PLUGIN_ROOT}` and anchors a still-relative path under `PluginDir` before deriving the search root, so a relative `skills/foo/SKILL.md` yields `<pluginDir>/skills` instead of a bare `skills`. Added `TestActivateResolvesManifestRelativeSkillRoot` (relative + placeholder cases).

Note: the production load path (`ParseManifest` -> `ResolvePluginPath`) already resolves skill paths to absolute, so this primarily hardens direct `Activate` callers; the guard is a no-op for already-absolute paths.

Gate: `gofmt -l` clean on changed dirs, `go build ./...` ok, `go test ./...` green for the touched packages.

Vasanthdev2004 (approved):
## Verdict: Approved

Reviewed the plugin activation/runtime wiring. No blockers found.

What I checked:
- Plugin tools register through the normal tool registry and keep `tools.Safety` permission metadata; no raw execution bypass is introduced.
- Tool name collisions are skipped with warnings instead of replacing core tools or earlier plugin tools.
- Plugin command output is redacted before display, commands are timeout-bounded, and `AGENT_PLUGIN_ROOT` is passed explicitly.
- Plugin hooks are namespaced, appended deterministically, and still respect hooks being disabled overall.
- Plugin skills merge with default skills while default/user skills win name collisions.

- `go test ./...`

PR #186, opened 2026-06-12
wedge stage 13: launch readiness (first-run + benchmark)

gnanam1990:
Stage 13 hardens the two things a launch is won or lost on: the first 60 seconds (install/first-run) and a reproducible, published benchmark. Scoped to the areas the stage spec lists (plus their tests).

## First run, model-agnostic-first
- **Local runtime detection** (`internal/provideronboarding/localruntime.go`): `LocalRuntimeCandidates()` derives the local, OpenAI-compatible candidates from the provider catalog (Ollama on :11434, LM Studio on :1234); `DetectLocalRuntimes()` probes their default endpoints and returns only the ones answering, with **no key required**. Transport errors, timeouts, and non-2xx responses are treated as "not running" and dropped — detection never errors.
- **Classified setup probe** (`internal/provideronboarding/setupprobe.go`): `ClassifySetupProbe()` maps a live connectivity probe result to a specific, fixable error class — `auth` (bad/missing key), `endpoint` (wrong base URL / unreachable / timeout), `model` (model not found, incl. 404), `rate_limit`, `config` — instead of a stack trace. Never panics, including on a degenerate/empty result.
- **`zero setup --verify`** (`internal/cli/setup.go`): after saving the provider, runs a live connectivity probe (guarded — no-op when no prober is wired), reports the classified fixable error on failure (exit `provider`), and on success confirms the endpoint is reachable. `setup` now also prints a concrete one-line `try this: zero exec "..." --model <model>` so the wizard ends in a working state.

## `zero doctor` hardening (`internal/doctor`)
- New **`sandbox.backend`** check: resolves the native backend (bubblewrap on Linux, sandbox-exec on macOS) and, when absent, **warns** (the policy engine still runs) with a platform-specific install remedy.
- New **`lsp.servers`** check: lists which language servers are missing from PATH with a per-binary install command. Source of truth is `lsp.ServerBinaries()` (small accessor over the existing server map).
- `doctor.Options` gains `GOOS` + `LookupExecutable` so both checks are deterministic in tests; production uses `runtime.GOOS` + `exec.LookPath`. Neither check can hard-fail the report.

## Self-correct surface + benchmark
- **`zero exec --self-correct`** (`internal/cli/exec.go`, `exec_parse.go`): wires the existing post-edit verify-and-correct loop (`agent.SelfCorrector`) via the project verifier. Off by default the corrector is nil, leaving the agent loop byte-identical; auto-fix vs. report-only is decided by the run's `--auto` autonomy gate.
- **Task benchmark harness** (`internal/perfbench/taskbench.go`): `RunTasks()` runs a `TaskSet` and produces a self-describing record capturing **model + commit + version + self-correct flag + date** plus per-task pass/fail/error and pass rate. `NewExecRunner()` is the production runner — it drives headless `zero exec --output-format stream-json` and decides pass/fail from the `run_end` exit code (or a task's external `verificationCommand`, Terminal-Bench style). The runner is injectable so the suite tests run against a mocked task set with no binary or model.
- **`zero-perf-bench tasks`** subcommand (`cmd/zero-perf-bench/tasks.go`): one command to run a suite reproducibly, with `--self-correct`, `--commit`/`--version` (or `ZERO_BENCH_*` env), `--output`, and `--dry-run`. A runnable sample suite ships at `cmd/zero-perf-bench/testdata/`. The existing perf-bench (cold-start/RSS) path is unchanged.
- **`docs/BENCHMARK.md`**: publishes the methodology — the exact two commands (with and without `--self-correct`), the recorded fields, why the score is model-bounded, and a results table for the self-correct delta. Provider-neutral throughout (`<model>` placeholders).

## Testing
- TDD per area. `go build ./...` and `gofmt -l .` clean (whole repo); `go vet` clean on touched packages.
- `go test ./...` passes. Note: under heavy parallel load one pre-existing test in the untouched `internal/config` package (`TestLoadProviderCommandFailureIncludesExitAndRedactsOutput`, a 5s external-command timeout) can flake; it passes deterministically in isolation and with bounded `-p`.

## Deferred (intentionally out of scope)
- The published benchmark numbers in `docs/BENCHMARK.md` are left as `TBD` — they must be produced on a fixed machine with a real model, which can't run in CI.
- The headless self-correct wiring verifies with the workspace test plan only (LSP half nil): `zero exec` does not spin up a language-server manager, and `NewSelfCorrector` accepts a nil checker. Wiring the LSP half into headless exec is left to a focused follow-up.

## Summary by CodeRabbit

* **New Features**
  * New task benchmark harness and `zero-perf-bench tasks` subcommand with dry-run, JSON/file output, and human summaries.
  * `exec` gains `--self-correct` to enable self-correcting execution.
  * `setup` gains `--verify` to probe and report provider connectivity.
  * Local runtime discovery/onboarding for no‑API‑key runtimes.
  * Doctor now checks sandbox backend and language-server availability.

* **Documentation**
  * Added benchmark methodology and reproducibility guide.

* **Tests**
  * Extensive new test coverage for tasks, exec self-correct, setup verification, doctor hardening, perfbench harness, and onboarding.

### DIFF
--- a/cmd/zero-perf-bench/main.go
+++ b/cmd/zero-perf-bench/main.go
@@ -33,6 +33,12 @@ func main() {
 }
 
 func run(args []string, getenv func(string) string, stdout io.Writer, stderr io.Writer) int {
+	// The `tasks` subcommand is the reproducible Terminal-Bench-style task harness;
+	// the default (no subcommand) path stays the cold-start/RSS perf benchmark so
+	// existing invocations are unchanged.
+	if len(args) > 0 && args[0] == "tasks" {
+		return runTasksCommand(args[1:], getenv, stdout, stderr)
+	}
 	options, err := parseArgs(args, getenv)
 	if err != nil {
 		_, _ = fmt.Fprintln(stderr, err.Error())
@@ -199,6 +205,7 @@ func parseArgs(args []string, getenv func(string) string) (cliOptions, error) {
 func helpText() string {
 	return strings.Join([]string{
 		"Usage: zero-perf-bench [options]",
+		"       zero-perf-bench tasks [options]   (Terminal-Bench-style task harness; see `tasks --help`)",
 		"",
 		"Options:",
 		"  --iterations <n>             Measured samples to collect (default: 5)",
--- a/cmd/zero-perf-bench/tasks.go
+++ b/cmd/zero-perf-bench/tasks.go
@@ -1,0 +1,218 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/perfbench"
+)
+
+// taskOptions configures the `zero-perf-bench tasks` subcommand: the reproducible
+// Terminal-Bench-style harness that runs ZERO headlessly against a task set and
+// records a publishable result (model + commit + self-correct flag + date).
+type taskOptions struct {
+	SuitePath   string
+	Model       string
+	Mode        string
+	SelfCorrect bool
+	Binary      string
+	Version     string
+	Commit      string
+	Output      string
+	JSON        bool
+	DryRun      bool
+	Help        bool
+}
+
+func runTasksCommand(args []string, getenv func(string) string, stdout io.Writer, stderr io.Writer) int {
+	options, err := parseTaskArgs(args, getenv)
+	if err != nil {
+		_, _ = fmt.Fprintln(stderr, err.Error())
+		return 2
+	}
+	if options.Help {
+		_, _ = fmt.Fprint(stdout, taskHelpText())
+		return 0
+	}
+
+	set, err := perfbench.LoadTaskSet(options.SuitePath)
+	if err != nil {
+		_, _ = fmt.Fprintln(stderr, "[zero] Task benchmark failed: "+err.Error())
+		return 1
+	}
+
+	runner := perfbench.SkipRunner
+	if !options.DryRun {
+		binary := strings.TrimSpace(options.Binary)
+		if binary == "" {
+			_, _ = fmt.Fprintln(stderr, "[zero] Task benchmark failed: --binary is required unless --dry-run is set")
+			return 2
+		}
+		runner = perfbench.NewExecRunner(binary)
+	}
+
+	result, err := perfbench.RunTasks(context.Background(), set, perfbench.TaskConfig{
+		Model:       options.Model,
+		Mode:        options.Mode,
+		SelfCorrect: options.SelfCorrect,
+		Version:     options.Version,
+		Commit:      options.Commit,
+		Runner:      runner,
+	})
+	if err != nil {
+		_, _ = fmt.Fprintln(stderr, "[zero] Task benchmark failed: "+err.Error())
+		return 1
+	}
+
+	if options.Output != "" {
+		if err := writeTaskReport(options.Output, result); err != nil {
+			_, _ = fmt.Fprintln(stderr, "[zero] Task benchmark failed: "+err.Error())
+			return 1
+		}
+	}
+	if options.JSON {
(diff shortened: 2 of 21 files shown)


gnanam1990:
Addressed the review findings:

- **setup.go** (`verifySetupProvider` / `setupJSONPayload`): a skipped probe (no probe wired) no longer reports `verified=true`. Added a `Ran` flag to `setupVerification`; the human `verified:` line and the JSON `verified` flag are now gated on `Ran`, so a not-run probe emits no verified flag.
- **perfbench/taskbench.go** (`LoadTaskSet`): relative `workspaceFixture` paths are now resolved against the manifest's directory (`filepath.Dir(path)`), so the runner's `cmd.Dir` no longer depends on the caller's cwd; absolute fixtures are left untouched.
- **perfbench/taskbench.go** (`LoadTaskSet`): reject trailing content after the manifest object via an EOF check, so a malformed `{...}\nnot-json` file fails loudly.
- **perfbench/taskbench.go** (`NewExecRunner`): the terminal `run_end` exit code is now authoritative — a non-zero agent exit is a normal failed task (not a harness error), and a missing terminal event fails closed as a harness error instead of reading as a pass.
- **provideronboarding/setupprobe.go** (`isModelNotFound`): a bare provider-side 404 no longer classifies as a missing model; a 404 only maps to the model class when the message references a model lookup. Added regression tests for the bad-base-URL 404 and the model-context 404.
- **provideronboarding/setupprobe.go** (nitpick): removed the duplicated `failingProbeCheck` and delegated to the canonical `Result.PrimaryCheck()`.
- **exec_parse.go**: `--self-correct` combined with `--use-spec` is now rejected at parse time (the spec-draft planning path never wires the self-correct loop), instead of being silently ignored.

All changes are covered by tests (TDD where behavior changed). `gofmt -l` clean, `go build ./...` and `go test ./...` pass.

Vasanthdev2004 (changes requested):
## Verdict: Request changes

One blocker before merge:

[P1] `internal/perfbench/taskbench.go:360` runs the task `verificationCommand` with `cmd.Env = appendNoColor(nil)`, so the verifier process receives only `NO_COLOR=1` instead of the normal environment. The documented sample suite uses commands like `go test ./...`; real verifiers often need inherited `PATH`, `HOME`, Go/Node/Python env, proxy settings, temp/cache locations, or project-specific env. The agent run itself correctly uses `appendNoColor(os.Environ())`, so the verifier should do the same unless there is a deliberate allowlist. As written, the new task-success benchmark can fail or behave differently from the same verifier command run manually.

Suggested smallest fix:
- Change the verifier to inherit the environment, e.g. `cmd.Env = appendNoColor(os.Environ())`.
- Add a test that proves a verification command can see an inherited env var or PATH-like value.

- `go test ./...`

gnanam1990:
@Vasanthdev2004 thanks for the catch — addressed in 4f72017.

`runVerification` (internal/perfbench/taskbench.go) now runs the verifier with `cmd.Env = appendNoColor(os.Environ())`, so it inherits PATH/HOME/toolchain env exactly like the agent run above, instead of getting only `NO_COLOR=1`. Added `TestRunVerificationInheritsEnvironment` which fails against the old `nil` env and passes now. `go test ./internal/perfbench/...` is green.

Ready for another look when you have a moment.

Vasanthdev2004 (approved):
## Verdict: Approved

Re-reviewed the latest head `4f72017228ee`. The blocker from my previous review is resolved.

What changed since the requested-changes review:
- `runVerification` now uses `appendNoColor(os.Environ())`, so task verifiers inherit PATH/HOME/toolchain/env the same way maintainers do when running verifier commands manually.
- Added `TestRunVerificationInheritsEnvironment`, covering the exact regression.

- `gofmt -l internal/perfbench/taskbench.go internal/perfbench/taskbench_test.go`
- `go test ./internal/perfbench`
- `go test ./...`

No remaining blockers found.

PR #187, opened 2026-06-13
Render assistant Markdown in TUI

anandh8x:
Summary:
- adds a lightweight in-house assistant Markdown renderer for TUI transcript text
- renders Markdown during streaming and after completion, including inline bold, code/emphasis marker cleanup, and pipe tables
- improves table readability with styled headers/separators and adaptive body row rules for wrapped tables
- keeps transcript selection/copy metadata plain so ANSI styling does not leak into copied text

Tests:
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Assistant responses now render full Markdown in-terminal (tables, headings, lists, blockquotes, bold, code) for final, streaming/interim and selectable rows; plain-text variants strip control markers for selection and plain views.
* **Bug Fixes**
  * Streaming cursor placement adjusted to fit available width without corrupting interim display.
* **Tests**
  * Extensive markdown and streaming tests added; one obsolete final-answer rail test removed.

### DIFF
--- a/internal/tui/assistant_markdown.go
+++ b/internal/tui/assistant_markdown.go
@@ -727,13 +727,19 @@ func wrapMarkdownInlineWithPrefixes(firstPrefix string, continuationPrefix strin
 		}
 		wordWidth := lipgloss.Width(word.text)
 		rendered := renderMarkdownInlineSegment(word)
+		separator := " "
+		separatorWidth := 1
+		if joinsPreviousMarkdownWord(word.text) {
+			separator = ""
+			separatorWidth = 0
+		}
 		switch {
 		case line == "":
 			line = rendered
 			lineWidth = wordWidth
-		case lineWidth+1+wordWidth <= available:
-			line += " " + rendered
-			lineWidth += 1 + wordWidth
+		case lineWidth+separatorWidth+wordWidth <= available:
+			line += separator + rendered
+			lineWidth += separatorWidth + wordWidth
 		default:
 			flush()
 			line = rendered
@@ -754,6 +760,10 @@ func markdownInlineWords(segments []markdownInlineSegment) []markdownInlineSegme
 		}
 	}
 	return words
+}
+
+func joinsPreviousMarkdownWord(text string) bool {
+	return text != "" && strings.Trim(text, ".,!?;:%)]}") == ""
 }
 
 func renderMarkdownInline(text string) string {
@@ -805,14 +815,14 @@ func parseMarkdownInline(text string) []markdownInlineSegment {
 	for index := 0; index < len(text); {
 		switch {
 		case strings.HasPrefix(text[index:], "**"):
-			if bold || strings.Contains(text[index+2:], "**") {
+			if !code && (bold || strings.Contains(text[index+2:], "**")) {
 				flush()
 				bold = !bold
 				index += 2
 				continue
 			}
 		case strings.HasPrefix(text[index:], "__"):
-			if bold || strings.Contains(text[index+2:], "__") {
+			if !code && (bold || strings.Contains(text[index+2:], "__")) {
 				flush()
 				bold = !bold
 				index += 2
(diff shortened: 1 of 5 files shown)


gnanam1990 (changes requested):
## Review — assistant Markdown renderer

Pulled and verified locally: builds clean, `go test ./...` green (59 pkgs), and the renderer displays correctly in the live TUI (headings, inline bold, code spans, pipe tables). Nice, self-contained, dependency-free implementation.

Requesting changes on two correctness issues in the inline tokenizer — both verified by running the actual `parseMarkdownInline` logic against sample inputs. They matter here specifically because this is a coding agent: its output is full of code identifiers, regexes, globs, and operators.

### 🟠 Medium — inline emphasis is parsed *inside* code spans
The `**` / `__` / `*` cases in `parseMarkdownInline` have no `code`-state guard, so emphasis markers get consumed even within `` `...` ``:

| Input | Rendered |
|---|---|
| `` `a.*b.*c` `` | `a.b.c` |
| `` `*.go` and `*.py` `` | `.go and .py` |
| `` `x__y__z` `` | `xyz` |

Fix: skip toggling `**`/`__`/`*` when `code` is true (let the rune fall through to the writer); evaluate/track the backtick state before the emphasis cases.

### 🟠 Medium — intra-word `__` bolds dunder names
No flanking/word-boundary check on `__`, so `the __init__ method` renders as bolded `init` with the underscores stripped. `__name__`, `__main__`, etc. are common in prose. The existing test only covers a standalone `__there__`, which is why this slipped through.

Fix: require `__` to be at a word boundary (don't open/close when flanked by alphanumerics).

### 🟡 Low — paired literal `*` in prose eaten
Same root cause: `compute 2 * 3 * 4` → `2  3  4`. A CommonMark-style flanking rule (delimiters must hug non-space text) fixes all three of the above together.

### 🟡 Low — dead / test-only code in a production file
- `stripMarkdownInline` — no callers anywhere.
- `renderMarkdownInline` — only called from a unit test; move to the test or drop.
- `renderAssistantMarkdownPlainText` — no production caller; confirm it's intended for future use.

### 🟡 Low — wide tables can exceed terminal width
`markdownTableWidths` floors each column at `markdownTableMinColumnWidth` (4); with many columns, `columns*4 + separators` can exceed `measure` and overflow on narrow terminals. Edge case — worth a comment or graceful degradation.

### ⚪ Note — branch is behind `main`
Branched before `cc618ba`. Only touches `internal/tui` so it's still mergeable, but a rebase before merge is cleaner.

---

**Verdict: request changes** — the feature is solid and well-tested, but the two medium issues corrupt exactly the kind of text this agent emits most. Suggest fixing the inline tokenizer to respect code spans + word boundaries, and adding regression tests for inline code containing `*`/`_` and for `__init__`. Happy to pair on the `parseMarkdownInline` fix.

gnanam1990 (approved):
## Re-review — all requested changes resolved ✅

Thanks for the fast turnaround on `fd0a429` and `2ad080e`. Re-reviewed against the real code (verified the unexported parser directly, plus the full `internal/tui` suite — green).

All three issues from my earlier review are fixed and backed by regression tests:

| Earlier finding | Status |
|---|---|
| Emphasis parsed inside code spans (`` `a.*b.*c` `` → `a.b.c`) | ✅ Fixed — `a.*b.*c` preserved |
| Intra-word `__` bolds dunders (`__init__` → `init`) | ✅ Fixed — `__init__`/`__main__` preserved |
| Literal `*` eaten (`2 * 3` → `2  3`) | ✅ Fixed — `2 * 3`, `a*b*c`, `*.go` preserved |

The fix is principled — CommonMark-style flanking via `canOpenMarkdownDelimiter`/`canCloseMarkdownDelimiter`, a dedicated `isMarkdownDunderIdentifier` guard, and all `*`/`**`/`__` toggles gated behind `!code` with a backtick flush. Nice. The new `TestMarkdownInlinePreservesLiteralStarsAndCodeBoundaries` and `TestMarkdownInlineKeepsCodingLiterals` cover exactly these cases plus ANSI-leak checks.

One intended behavior to note for future readers: single-word `__strong__` now renders literally (treated as an identifier); multi-word `__bold text__` and all `**bold**` still bold. Reasonable tradeoff for a coding agent.

Non-blocking follow-ups (optional): a few helpers look dead/test-only (`stripMarkdownInline`, `renderMarkdownInline`, `renderAssistantMarkdownPlainText`), and the branch is slightly behind `main` — a rebase before merge would be tidy.

Approving.

PR #188, opened 2026-06-13
feat: add MCP manager UX

Vasanthdev2004:
﻿## Summary
- adds MCP config management commands: `add`, `remove`, `enable`, `disable`, and `check`
- preserves unknown config JSON while updating MCP server config, and redacts env/header secrets in command output
- adds `/mcp` TUI status rendering backed by configured servers, registered MCP tools, permission grants, and OAuth token status
- threads MCP config/stores into the interactive TUI launch path

## Why
Zero already had MCP backend pieces, but users had no coherent product surface to manage and inspect MCP servers. This PR makes MCP feel like a real manager flow without adding marketplace/plugin scope.

## Validation
- `gofmt -l internal/cli internal/tui`
- `go test ./internal/cli ./internal/tui ./internal/mcp -count=1`
- `go test ./...`
- `go vet ./...`
- isolated manual smoke with temp `%APPDATA%`: `zero mcp add`, `disable`, `enable`, `list`

## Notes
- This intentionally keeps live server checks explicit via `zero mcp check <server>` so opening `/mcp` does not unexpectedly connect to remote/local MCP servers.
- I briefly ran a manual smoke with the wrong temp config env var; it added a `docs` test server to the real user config, then I immediately removed it. A follow-up `zero mcp list` on the real config reports no MCP servers configured.

## Summary by CodeRabbit

* **New Features**
  * Full MCP management: CLI subcommands (add/remove/enable/disable/check/list), TUI /mcp command, manager overlay, add wizard, marketplace install flows, and in-app MCP command execution with buffered results.

* **Bug Fixes / Privacy**
  * Stronger redaction of URLs, query/fragment params, headers, env and command args; safer MCP runtime/token-store initialization and cleanup; improved UI rendering and wrapping.

* **Tests**
  * Extensive CLI and TUI test coverage for MCP workflows, validation, migrations, permissions, OAuth/token handling, views, manager, mouse and autocomplete interactions.

### DIFF
--- a/internal/agenteval/benchmark_test.go
+++ b/internal/agenteval/benchmark_test.go
@@ -453,12 +453,13 @@ func TestHarnessTimeoutCancelsBlockedAgent(t *testing.T) {
 		}},
 	}
 
-	// The timeout is far longer than materialization of the small fixture, so
-	// the agent is always reached; it then blocks until the deadline fires.
+	// The timeout is longer than materialization of the small fixture even on
+	// slower Windows CI runners, so the agent is reached before it blocks until
+	// the deadline fires.
 	report := harness.Run(context.Background(), suitePath, suite, BenchmarkInput{
 		TaskID:   "document-stream-json-verify-events",
 		WorkRoot: t.TempDir(),
-		Timeout:  2 * time.Second,
+		Timeout:  10 * time.Second,
 	})
 
 	if !agentReached {
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -449,7 +449,9 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	}
 	mcpTokenStore, err := deps.newMCPTokenStore()
 	if err != nil {
-		return writeAppError(stderr, "failed to initialize MCP OAuth tokens: "+err.Error(), 1)
+		_, _ = fmt.Fprintf(stderr, "[zero] warning: failed to initialize MCP OAuth tokens: %s\n", err)
+		mcpTokenStore = nil
+		err = nil
 	}
 	mcpRuntime := mcpToolRuntime(noopMCPRuntime{})
 	if len(mcpConfig.Servers) > 0 {
@@ -459,6 +461,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		})
 	}
 	if err != nil {
+		closeMCPRuntime(stderr, mcpRuntime)
 		return writeAppError(stderr, err.Error(), 1)
 	}
 	defer closeMCPRuntime(stderr, mcpRuntime)
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -439,7 +439,25 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		return writeAppError(stderr, "failed to initialize specialist tools: "+err.Error(), 1)
 	}
 	defer closeSpecialistRuntime(stderr, specialistRuntime)
-	mcpRuntime, err := registerMCPToolsForWorkspace(context.Background(), workspaceRoot, registry, deps, mcp.AutonomyLow)
+	mcpConfig, err := deps.resolveMCPConfig(workspaceRoot)
+	if err != nil {
+		return writeAppError(stderr, err.Error(), 1)
+	}
+	mcpPermissionStore, err := deps.newMCPStore()
+	if err != nil {
+		return writeAppError(stderr, "failed to initialize MCP permissions: "+err.Error(), 1)
+	}
+	mcpTokenStore, err := deps.newMCPTokenStore()
+	if err != nil {
+		return writeAppError(stderr, "failed to initialize MCP OAuth tokens: "+err.Error(), 1)
+	}
+	mcpRuntime := mcpToolRuntime(noopMCPRuntime{})
+	if len(mcpConfig.Servers) > 0 {
+		mcpRuntime, err = deps.registerMCPTools(context.Background(), registry, mcpConfig, mcp.RegisterOptions{
+			PermissionStore: mcpPermissionStore,
+			Autonomy:        mcp.AutonomyLow,
+		})
+	}
 	if err != nil {
 		return writeAppError(stderr, err.Error(), 1)
 	}
@@ -482,17 +500,20 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		Scope:         scope,
 	})
 	return deps.runTUI(context.Background(), tui.Options{
-		Cwd:             workspaceRoot,
-		UserConfigPath:  userConfigPath,
-		ProviderName:    resolved.Provider.Name,
-		ModelName:       resolved.Provider.Model,
-		ProviderProfile: resolved.Provider,
-		FavoriteModels:  resolved.Preferences.FavoriteModels,
-		Provider:        provider,
-		NewProvider:     deps.newProvider,
-		Registry:        registry,
-		SessionStore:    deps.newSessionStore(),
-		SandboxStore:    sandboxStore,
+		Cwd:                workspaceRoot,
+		UserConfigPath:     userConfigPath,
+		ProviderName:       resolved.Provider.Name,
+		ModelName:          resolved.Provider.Model,
+		ProviderProfile:    resolved.Provider,
+		FavoriteModels:     resolved.Preferences.FavoriteModels,
+		Provider:           provider,
+		NewProvider:        deps.newProvider,
+		Registry:           registry,
+		SessionStore:       deps.newSessionStore(),
+		SandboxStore:       sandboxStore,
+		MCPConfig:          mcpConfig,
+		MCPPermissionStore: mcpPermissionStore,
+		MCPTokenStore:      mcpTokenStore,
 		AgentOptions: agent.Options{
 			MaxTurns:       resolved.MaxTurns,
 			Registry:       registry,
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -519,9 +519,12 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		MCPConfig:          mcpConfig,
 		MCPPermissionStore: mcpPermissionStore,
 		MCPTokenStore:      mcpTokenStore,
-		MCPCommand: func(args []string) tui.MCPCommandResult {
+		MCPCommand: func(ctx context.Context, args []string) tui.MCPCommandResult {
+			if ctx == nil {
+				ctx = context.Background()
+			}
 			var stdout, stderr bytes.Buffer
-			exitCode := runMCP(args, &stdout, &stderr, deps)
+			exitCode := runMCPWithContext(ctx, args, &stdout, &stderr, deps)
 			nextConfig := lastKnownMCPConfig
 			if refreshed, err := deps.resolveMCPConfig(workspaceRoot); err == nil {
 				lastKnownMCPConfig = refreshed
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -503,6 +503,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		Backend:       deps.selectSandboxBackend(sandbox.BackendOptions{}),
 		Scope:         scope,
 	})
+	lastKnownMCPConfig := mcpConfig
 	return deps.runTUI(context.Background(), tui.Options{
 		Cwd:                workspaceRoot,
 		UserConfigPath:     userConfigPath,
@@ -521,8 +522,9 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		MCPCommand: func(args []string) tui.MCPCommandResult {
 			var stdout, stderr bytes.Buffer
 			exitCode := runMCP(args, &stdout, &stderr, deps)
-			nextConfig := mcpConfig
+			nextConfig := lastKnownMCPConfig
 			if refreshed, err := deps.resolveMCPConfig(workspaceRoot); err == nil {
+				lastKnownMCPConfig = refreshed
 				nextConfig = refreshed
 			}
 			return tui.MCPCommandResult{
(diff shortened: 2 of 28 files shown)


gnanam1990 (dismissed):
## Review — MCP manager UX

Pulled and reviewed locally. The PR's own validation claims all hold on my machine: `gofmt -l internal/cli internal/tui` clean, `go vet ./internal/cli/... ./internal/tui/... ./internal/mcp/...` exit 0, and `go test ./internal/cli ./internal/tui ./internal/mcp` all pass.

### What's strong
- **Safe config writes** — atomic temp-file-in-same-dir + rename, `0600` file / `0700` dir, plus a Windows rename-over-existing fallback that restores the original on failure.
- **Real preservation** — `readMCPWritableConfig`/`marshalJSON` round-trip through `map[string]json.RawMessage` at three levels, so unknown top-level keys, unknown `mcp.*` keys, and (on enable/disable) unknown per-server fields all survive. `enable`/`disable` only touch the `disabled` field.
- **Careful secret handling** — `redactMCPServerConfigs` deep-copies and replaces every env/header **value** with `[REDACTED]` (keys kept), with `redaction.RedactValue` layered on top; errors flow through `redaction.ErrorMessage`. The `/mcp` OAuth view exposes only `HasToken`/`HasRefreshToken`/`TokenType`/`Scopes`/`ExpiresAt`/`Expired` — never the token value.
- **Good boundary** — `/mcp` is a pure render of cached state; the only path that connects to a server is the explicit `zero mcp check`.

### Findings (all low severity — none blocking)

**🟡 Low–Med — token-store init is now fatal to TUI launch.** `app.go:450-453`: a `newMCPTokenStore()` error aborts the whole interactive launch. OAuth status is a peripheral display feature; a corrupt/unreadable token store shouldn't brick the entire TUI. Suggest failing soft (warn, pass a nil token store so the OAuth section just renders empty). This is the one I'd most like addressed.

**🟡 Low — enable/disable ignore legacy server keys.** `setServerDisabled` only consults `serverRaw` (from `mcp.servers`), whereas `upsert`/`remove` handle legacy `mcpServers`/`mcp_servers` via `removeLegacyServer`. A server defined only under a legacy key reports "not configured" on `enable`/`disable` despite being removable. Either migrate-on-toggle or document that toggling requires the canonical `mcp.servers` location.

**🟡 Low — full server URL printed verbatim.** `mcpServerTarget` (and `mcp list`) print the raw `URL` while env/headers are redacted. A URL with embedded credentials (`https://user:pass@host`, `?token=…`) would be shown in `/mcp` and `mcp list`. Consider redacting userinfo / sensitive query params, or treat as accepted if URLs are deemed non-secret.

**🟡 Low — flag/transport validation gap.** The parser accepts `--env` for http/sse and `--header` for stdio; they're stored as silent no-ops. Could warn or reject.

**⚪ Nits**
- `func (cfg mcpWritableConfig) marshalJSON()` has a value receiver but mutates the `raw`/`mcpRaw` maps (reference types). Harmless (cfg is discarded after write) but a pointer receiver would signal intent.
- `--env FOO=` (empty value) is accepted — confirm that's intended.

### Verdict
Approve. No correctness or security bugs found — redaction, atomic writes, and config preservation are well done, and vet/gofmt/tests are green. Suggest soft-failing the token-store init and reconciling the legacy enable/disable gap as follow-ups; neither blocks merge. Nice work.

Vasanthdev2004:
@coderabbitai review

Fixed the MCP review items in the latest commits:
- close partial MCP runtime on registration errors and soft-fail OAuth token-store init for TUI launch
- migrate legacy mcpServers/mcp_servers entries on enable/disable
- redact URL userinfo/query/fragment secrets in CLI and /mcp display, including URL-valued stdio args
- preserve unknown per-server JSON fields when updating an existing MCP server
- stabilized the Windows agent-eval timeout test that failed CI

Local validation passed: gofmt -l, go vet ./..., go test ./..., go run ./cmd/zero-release build, go run ./cmd/zero-release smoke. GitHub checks are green.

Vasanthdev2004:
@coderabbitai review

gnanam1990 (changes requested):
## Review: MCP manager UX

Thanks for this — it's well-structured, and the parts that matter most for safety are done carefully: **secret redaction in the JSON paths uses structural (key-level) redaction, not just pattern matching**, and **unknown config JSON is genuinely round-tripped at all three nesting levels** (top-level, `mcp.*`, per-server), both with strong tests. The PR builds clean and `go test ./...` passes.

Requesting changes for two verified UI-freeze issues and a narrow redaction inconsistency; the rest are minor/nits and test gaps.

### 🔴 Blockers

**1. Synchronous MCP connect on the Bubble Tea UI goroutine freezes the TUI.**
`runMCPManagerCommand` (`internal/tui/mcp_manager.go:222`) calls `handleMCPCommand` **inline** and returns a `nil` `tea.Cmd`. For `check`, the chain `handleMCPCommand → m.mcpCommand → runMCP → runMCPCheck → registerMCPTools → mcp.Connect` spawns the stdio process / opens the network connection synchronously, with a **non-cancellable `context.Background()`** and a 30s timeout. Pressing Enter / `check` on an unreachable server hangs the entire UI (no input, render, or Ctrl-C) until it returns.
*Suggested fix:* run the command inside a `tea.Cmd` that emits a result `tea.Msg`; keep the manager responsive and store a `context.CancelFunc` you cancel on Esc.

**2. Disk reads inside `View` on every frame.**
`mcpViewState()` (`internal/tui/command_views.go:55`, a View helper — also called from `mouse.go:229` and `mcpManagerOverlay`) calls `BuildMCPViewState`, which runs `PermissionStore.List()` + `TokenStore.Status()` (`os.ReadFile` + unmarshal) on **every render and every mouse move** while the manager is open. That's side-effecting I/O in View and causes per-keystroke jank.
*Suggested fix:* build the `MCPViewState` once when the manager opens / after a command completes and cache it on the model; have View read the cached struct.

### 🟠 Major

**3. Mouse hit-test off-by-one** — `internal/tui/mouse.go:231` (`baseRow := 3`). The bordered overlay places the first item at rendered row 4 (5 when the empty-state line is shown), so a click selects the item one row *below* the cursor and the top item is unclickable. It's bounds-guarded (no panic) but selects the wrong server. Compare the working `selectModelPickerAtMouse`.
*Suggested fix:* `baseRow := 4` (5 when `len(Servers) == 0`) + a click test asserting index 0 is selected.

**4. `mcp list` (plaintext) is the only output path not wrapped in `redaction.RedactString`** — `internal/cli/extensions.go:213` vs the tool-list at `:286` (which is wrapped). `formatMCPServerList` prints the stdio `command` verbatim. URL creds are handled (via `redactMCPURL`) and env/headers/args aren't printed, so this is narrow — but a secret embedded in a stdio `command` leaks in plaintext while `--json` correctly redacts it.
*Suggested fix:* `fmt.Fprintln(stdout, redaction.RedactString(formatMCPServerList(cfg.Servers), redaction.Options{}))`.

### 🟡 Minor / Nit
- **No lock on the read-modify-write of the user config** (`mcp_config.go` add/remove/toggle): concurrent `mcp add`, or the in-process TUI `MCPCommand` racing a CLI invocation, can lose updates. Atomic rename prevents *corruption*, not *lost updates* — consider an advisory lock spanning read→write.
- Atomic write doesn't `tmp.Sync()` before close (durability gap for a secrets file). The Windows restore branch in `replaceMCPWritableConfigFile` can delete the original if both the backup read and the rename fail (`readErr != nil` path).
- `runtime` local var shadows the imported `runtime` package (`mcp_config.go:264`).
- Wizard Result-step arrow keys are dead — the screen shows a `>` cursor on a list `move` doesn't navigate (`mcp_add_wizard.go`).

### Test gaps (suite passes; coverage solid — cli ~69%, tui ~80%)
- **Atomic-write/corruption + 0600 permission** untested (`replaceMCPWritableConfigFile` ~20%): add a test that induces a write failure and asserts the original config is intact, and that a successful write lands at mode 0600.
- **`mcp add` validation-rejection** paths mostly untested: table-test invalid name (`"../escape"`, empty), bad `--type`/`--auth`, `http` without `--url`, `stdio` without command — assert `exitUsage` and that no file was written.
- Wizard/manager navigation (`move`/`back`/empty-list) and the `fallbackRedactMCPDisplayURL` redaction path are uncovered.

### ✅ Verified strengths
- Structural env/header redaction + URL/query/fragment redaction across CLI and TUI, tested with negative assertions on the actual secret value.
- Unknown-key preservation across add/remove/enable/disable round-trips, incl. legacy `mcpServers` migration and OAuth-block preservation.
- List navigation is panic-safe (bounds clamped); width/wrapping math is defensive.
- Opening `/mcp` does **not** connect to servers (live checks are explicit) — as documented.

Happy to pair on the async-`tea.Cmd` refactor for #1/#2 if useful.

gnanam1990 (changes requested):
## Re-review: MCP manager UX

Big turnaround — I verified all **4 blockers from my last review are genuinely fixed** (not just claimed):

- **Sync MCP connect froze the UI** → now async via `tea.Cmd` with a cancellable context and a sequence guard (`mcpCommandSeq`) that drops stale results. Race handling (cancel vs late result) is correct.
- **Disk reads in `View`** → view-state is cached; `refreshMCPViewState` is called at manager-open / after-command / wizard-save, so `View` hits the cache.
- **Mouse hit-test off-by-one** → `mcpManagerFirstItemRow` now returns 4 (5 when empty), the correct offset.
- **`mcp list` printed the stdio command unredacted** → now wrapped in `redaction.RedactString` (`extensions.go:220`).

Also verified: the marketplace "install" is **safe** — it only stages a command string in the composer, no network fetch or `exec` at selection/render time; the `mcp check` context now threads all the way to the MCP connect; secret redaction covers every output path; build clean and `go test ./...` passes.

Requesting changes for one lossy-input issue; the rest are minor/nits.

### 🟠 Should fix before merge

**Chat setup-intent silently discards a legitimate prompt.** `internal/tui/model.go:1677` (`commandPrompt` branch):

```go
if intent, ok := detectMCPSetupIntent(command.text); ok {
    return m.openMCPAddWizardFromIntent(intent), nil   // never launches/records command.text
}
```

`detectMCPSetupIntent` fires when the text contains `"mcp"` (or "model context protocol") + an action verb (`add/setup/configure/install/connect/enable`) + a catalog word — and the catalog words include **`memory`, `fetch`, `filesystem`**. So ordinary questions like *"how do I add a **fetch** call to my **mcp** client?"* or *"explain how **memory** works when you **connect** to an **mcp** server"* are intercepted: the wizard opens and the user's message is **silently dropped** — not sent to the agent and not even shown in the transcript (the composer was already cleared upstream).

*Fix:* don't discard `command.text`. Either require a much stronger signal (a setup URL, not a bare catalog word), or surface a non-modal suggestion ("Looks like MCP setup — open the wizard?") while still sending/queuing the original prompt.

### 🟡 Minor
- **Stale manager result re-opens over an open wizard** — `internal/tui/command_views.go:141` re-creates `m.mcpManager` from a manager-origin result, but opening the add wizard (`mcp_add_wizard.go` open path) while a manager `check` is in flight doesn't `cancelMCPCommand()`. When the late result lands it's non-stale and resurrects `mcpManager` behind the wizard, which then appears when the wizard closes. *Fix:* `cancelMCPCommand()` on wizard open, or skip re-opening the manager when `m.mcpAddWizard != nil`.
- **Windows config-loss** (`internal/cli/mcp_config.go:859-874`, carried from last review) — the rename fallback `os.Remove`s the original config before the restore branch (which is gated on a successful backup read), so a Windows-only double-failure (backup read + retry rename both fail) leaves the user with neither file. *Fix:* rename the original to `.bak` and restore from it on failure, rather than removing it up front.

### Nits
- `runtime` local var shadows the imported `runtime` package (`mcp_config.go:267`; same at `extensions.go:281`).
- The wizard's result action-id and label slices (`mcp_add_wizard.go` vs `mcp_add_wizard_view.go`) are parallel and indexed by the same cursor — a future edit to one could mismap; consider a single `[]struct{ID, Label string}`.
- Marketplace dedup keys on server display name, not catalog ID, so a catalog server installed under a custom name still shows as available (cosmetic).
- No test asserts a *cancelled* context actually aborts the `check` connect (only structural ctx threading is covered).

The TOCTOU / missing-fsync items I raised before match the repo's existing config-writer convention — fine to defer.

**Verdict:** Excellent work — every blocker is genuinely resolved and the architecture is sound now. The one item worth fixing before merge is the lossy chat setup-intent hijack; the rest are minor. Happy to re-review quickly once Finding 1 is addressed.

Vasanthdev2004:
@coderabbitai review

gnanam1990 (approved):
## Re-review — all findings resolved ✅

Thanks for the quick turnaround on `79f6595`. I verified each fix against the code:

- **Chat setup-intent no longer hijacks legitimate prompts** — detection is now gated on `promptContainsSetupEndpoint`, so a bare catalog word ("memory"/"fetch"/"filesystem") no longer triggers; it requires the actual endpoint host/path in the message. stdio catalog entries (whose "endpoint" is an `npx …` command with no URL host) can't auto-trigger from chat at all, which is the right conservative behavior. The regression test (`TestChatMCPSetupFalsePositiveSendsPrompt`) asserts the prompt reaches both the transcript and the provider. 👍
- **Stale manager result no longer re-opens behind the wizard** — both wizard-open paths now `cancelMCPCommand()`; covered by `TestMCPAddWizardOpenCancelsStaleManagerResult`.
- **Windows config replace** — now renames the original to `.bak` and restores it on failure (with a wrapped error if the restore itself fails), so the user's config is never destroyed without a recovery path.
- **`runtime` shadow** — renamed to `mcpRuntime` in both spots.

And the earlier blockers stay fixed: async `tea.Cmd` + cancellable context + sequence guard, cached view-state (no disk I/O in View), correct mouse hit-test offset, and redacted `mcp list`. Marketplace "install" only stages a composer string — no network/exec at selection/render time.

Build clean, `go vet` clean, `go test ./...` passes. Remaining items are pure nits I'm happy to leave as follow-ups (parallel action-id/label slices in the wizard, marketplace dedup keyed on server name vs catalog ID, and no test for a *cancelled* ctx aborting the `check` connect).

Approving — nice work.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

Vasanthdev2004:
@coderabbitai review

Vasanthdev2004:
@coderabbitai approve

PR #189, opened 2026-06-13
Deep-audit remediation + sandbox hardening (batches 1–35)

gnanam1990:
## Overview

Remediation of findings from the 2026-06-10 deep audit, **plus a focused sandbox-hardening pass**. Every finding was re-verified against current `main` before fixing — the ledger had drifted, so ~11 rows were already fixed and several were intentional/test-guarded (not defects). Every behavioural change ships with a test; `gofmt`, `go vet ./...`, host + `GOOS=linux` + `GOOS=windows` builds, and the full suite are green on CI (all four Smoke jobs: ubuntu, macos, windows, performance). TUI findings are excluded (owned by a separate feature branch).

**Scope:** 89 files, ~+3.4k/−0.3k, 36 commits, no behavioural change to any default path unless a new opt-in flag is set.

---

## Sandbox hardening (batches 23–34)

The largest part of this PR. It deepens the existing macOS seatbelt / Linux bubblewrap sandbox and closes the gap where non-shell tools bypassed the network policy. **Every new enforcement feature is opt-in and off by default**, so existing behaviour is unchanged unless explicitly enabled.

### Network policy now covers the built-in network tools (the real gap)
- **`web_fetch` and `web_search` honour the scoped/deny policy** (batches 33–34). Previously the egress proxy only constrained sandboxed *shell* commands; `web_fetch`/`web_search` opened their own connections and could reach any public host even under `deny`/`scoped`. A new shared gate `Engine.NetworkHostAllowed(host)` reuses the existing egress domain matcher as the single source of truth (allow → all; deny → none; scoped → allowlist minus denylist; empty allowlist collapses to deny).
  - `web_fetch` checks the host **before the request and on every redirect**, failing closed before any dial; the no-sandbox path is byte-for-byte unchanged.
  - `web_search` checks the configured backend's endpoint host before the query leaves the machine.
  - The per-host allowlist is the active enforcement on the macOS `sandbox-exec` backend (the only one that can route scoped egress); other backends collapse `scoped` to `deny` for these tools, which the engine already enforced at evaluation time.
  - Proven end-to-end with a registry-level test: `RunWithOptions → Evaluate → RunWithSandbox → per-host allowlist blocks an unlisted host before any dial`.

### Backend isolation depth
- **Richer macOS seatbelt profile** (batch 23): a curated `com.apple.*` `mach-lookup` allowlist and `(allow signal (target self) (target pgrp))`, so common toolchains work under the profile while the default-deny posture and the workspace write-boundary still hold (verified live on the host).
- **Linux seccomp `AF_UNIX` block** (batches 30, 32): a classic-BPF seccomp program that denies Unix-socket creation (the channel bubblewrap's filesystem/network isolation leaves open), delivered as a `zero-seccomp` exec-wrapper and wired into the bubblewrap command via `Policy.BlockUnixSockets`. Discovery finds the helper next to the binary or on `PATH` and **degrades gracefully** (runs without the filter) when absent.
- **macOS denial log monitor** (batch 31): with `Policy.MonitorDenials`, the seatbelt profile tags its default-deny and the bash path tails the unified log for that tag, surfacing blocked operations back to the agent as a `<sandbox_violations>` block on stderr.

### Permission / egress model
- **Exact-path grants** (batch 25): create and revoke grants scoped to a specific path (`--path` on the sandbox allow/revoke CLI), not just tool-wide.
- **`Engine.Precheck`** (batch 26): report the violations that *would* block a request before it runs, so callers can fail fast with the reason.
- **Scoped-egress domain prompt** (batch 27): interactive per-host authorization with a timeout and a decision cache.
- **AST-based shell analysis** (batch 28): `AnalyzeCommand` parses commands with a real shell grammar instead of string heuristics.
- **Quote-aware segment splitting** (batch 14): operators inside quotes (`"a | b"`, `"a; vim b"`) no longer split, so quoted program names aren't falsely flagged while real unquoted operators still do.
- Internal cleanups: unify proxy-env generation into `ProxyEnv` (batch 29); remove the dead `OnSandboxDecision` hook (batch 24).

### Config surface
Both new hardening flags are reachable from the `sandbox` config block and documented in the README — `sandbox.blockUnixSockets` (Linux) and `sandbox.monitorDenials` (macOS) — and map onto the policy opt-in only (config can turn a flag on, never silently off).

> **Honest caveats (flagged, not hidden):** the *runtime* effect of the macOS denial monitor (batch 31) and the Linux seccomp block (batches 30, 32) can only be exercised on the respective OS — the seccomp kernel-level block needs a Linux host/CI, and on at least one macOS version seatbelt denials aren't delivered to the queryable log. Both are **default-off**, their wiring and logic are unit-tested on every platform, and neither can affect whether a command runs unless an operator opts in.

---

## Correctness / robustness (batches 5, 6, 13, 15, 16, 18, 22)
- **zerogit** (5) — `git status --porcelain -z`: renames report the destination path; non-ASCII/whitespace filenames arrive verbatim, not C-quoted.
- **config** (6) — reject a negative `maxTurns` instead of silently dropping it; remove dead contract-gap code.
- **background/specialist** (13) — kill background tasks as a process group (`Setpgid` + negative-PID signal + bounded wait) so forked children can't outlive a cancel; on a `SetPID` failure, kill + record stop instead of orphaning.
- **cron** (15) — collapse the DST fall-back repeated hour so a daily job fires once, not twice.
- **mcp** (16) — skip server notifications/requests that arrive before the SSE response (no more id-mismatch failures); raise the SSE event cap to a bounded ceiling.
- **agent** (18) — count tool definitions in the compaction token estimate; don't deny an always-allow tool just because a grant-persist write failed.
- **cli/cron** (22) — capture the actual run error from `stream-json` stdout instead of a generic failure.

## Resource / durability / perf (batches 7, 9, 10, 11, 12, 17)
- **mcp** (7) — bound captured server stderr (was unbounded for the whole process lifetime).
- **sessions** (9) — build `Tree` from one `List()` snapshot, killing an N+1 disk scan.
- **specialist** (10) — serialize the accounting dedup→append (a TOCTOU that wrote duplicate stop/usage events).
- **streaming runtime** (11) — O(1) streamed-text accumulation; decode the cached-token counts a streaming provider reports.
- **specialist** (12) — parse the child stream without a fixed line-length cap that aborted runs on large output.
- **sessions** (17) — fsync session metadata before its atomic rename so a crash can't leave a torn/empty session file.

## Usage / accounting (batches 19, 20)
- **usage** (19) — price escalated usage from the event's own model, not the session default.
- **sessions** (20) — don't copy usage events into a fork, avoiding a double-count.

## Tooling (batches 4, 8, 21)
- **cli** (4) — `list-tools -o json`; interrupted `-o json` runs emit a terminal event; surface session-record failures.
- **tools/providers** (8) — `read_file` emits a truncation marker on a `max_lines` cut; remove a provably-dead provider-resolution branch.
- **cli/skills** (21) — warn on duplicate skill names instead of silently dropping one.

## CI / portability (batch 35)
- Fixed the sole Windows CI failure: `TestWriteFileSyncRoundTrips` asserted an exact Unix file mode (`0o600`), which Windows doesn't honour (Go reports `0o666`). The exact-perm check is now guarded to the platforms that enforce it; the write/read-back round-trip is still asserted everywhere.

---

## Testing & verification
- `gofmt -l` clean · `go vet ./...` clean.
- Host build, `GOOS=linux go build ./...` (incl. the `zero-seccomp` wrapper), and `GOOS=windows go build ./...` all pass.
- Full `go test ./...` green; all four CI Smoke jobs (ubuntu / macos / windows / performance) pass.
- macOS seatbelt profile changes validated live on the host (signal + mach-lookup grants work; workspace boundary still holds).

## Deliberately left (documented in commits)
Items that are intentional/test-guarded (e.g. `ResolveMCP` not running the provider command), perf-only with cross-process-unsafe shortcuts, nuanced usage-accounting edge cases, or low-value/forward-looking dead code. The seccomp/denial-monitor *runtime* validation is the only known follow-up and is gated behind default-off flags.

## Summary by CodeRabbit

* **New Features**
  * Shell-command safety analyzer; macOS denial monitor and Linux unix-socket blocking helper; sandbox-aware web_fetch/web_search and interactive domain prompting; JSON exec tool listing; path-scoped sandbox grants; zero-seccomp CLI.

* **Configuration**
  * New sandbox flags: sandbox.blockUnixSockets, sandbox.monitorDenials; maxTurns now validated (negatives rejected).

* **Bug Fixes**
  * Cron DST fallback behavior; nested-secret redaction; session metadata durability and background/process termination robustness; assorted sandbox/network hardening and stability tests.


anandh8x:
@coderabbitai re review

gnanam1990:
Addressed the CodeRabbit review in `6289a37` (batch 36). All 5 findings:

- **`background/process_posix.go`** — after `SIGKILL`, poll until the group is actually gone (it's async) and error if it survives a second deadline, rather than returning `nil` while descendants race. ✅ as suggested.
- **`sandbox/safe_command.go`** — track `$()` depth; `)` is only a boundary when it closes an active substitution, so `echo "a) less"` no longer splits into a fake `less` segment. ✅ as suggested. Added a direct `splitShellSegments` test (fails pre-fix: `["echo \"a", "less\""]`).
- **`config/types.go`** — message now reads `must be >= 0` to match the check (0 is accepted). ✅
- **`docs/audit/...status.md`** — header date/branch metadata now matches the body. ✅
- **`cron/schedule.go`** — used the **body** suggestion (`after.Truncate(time.Minute).Equal(after)`), **not** the AI-prompt variant (`!t.After(after)`). Rationale: `t` is initialized to `after.Truncate(minute).Add(minute)`, so it is *always* strictly after `after`; `!t.After(after)` would therefore be permanently false and silently disable the DST collapse entirely. The boundary check correctly distinguishes the "last fire time" input (collapse the repeat) from a sub-minute `after` like `01:30:30 EDT` (return the legitimate repeated `01:30 EST`). New test fails pre-fix.

Each behavioural fix ships with a test confirmed to fail without the change; gofmt/vet clean; host + linux + windows builds green.

gnanam1990:
@coderabbitai full review

All 5 prior findings are addressed in `6289a37` (batch 36). CI is green across all four Smoke jobs (ubuntu/macos/windows) + performance. Please re-review.

Vasanthdev2004 (approved):
Manual review verdict: approved. I rechecked the latest head (6289a379) across the high-risk sandbox, tool execution, MCP, usage, and compaction paths and did not find a PR-blocking issue. Focused local validation passed for sandbox, tools, usage, agent, MCP, background, cron, sessions, and zerogit. Note: internal/cli timed out locally because this machine has configured MCP servers and that test path is not fully hermetic here; GitHub CI smoke is green on Windows/Ubuntu/macOS, so I am not treating it as a blocker for this PR.

gnanam1990:
Addressed the re-review findings in `591f70b` (batch 37). Each behavioural fix has a regression test (several confirmed to fail without the change):

- **analyzer**: wrapper-aware (`sudo`/`env`/`bash -c …` now classify on the payload, with bounded `sh -c` recursion) + honors `rm --` end-of-options.
- **engine**: `NetworkHostAllowed` now shares one backend-aware effective-mode helper with `Evaluate`, so scoped fails closed when the backend can't enforce scoped egress (single source of truth).
- **egress**: prompt-decision cache keyed by `host:port`, not host.
- **safe_command**: segment splitter is now escape-aware (`\|`, `\"`).
- **log_monitor (darwin)**: `Wait()` the `log stream` child to avoid a zombie.
- **web_search**: backend confined to same-host redirects (with a limit) so a hop can't egress to an unchecked host.
- **cli/sandbox**: empty `--path` (`--path=` / `--path ""`) fails closed instead of widening scope.
- **cli/skills**: warns on stderr when duplicate detection fails.
- **specialist/exec**: joins a failed orphan-kill with the SetPID error instead of masking it.

**One deliberate deviation — `background/process_posix.go`:** I did **not** adopt the suggested `Getpgid(pid)` → `kill(-pgid, …)` directly. For a PID that is *not* its own group leader, `pgid` can be another group — **possibly our own** — and `kill(-pgid)` would then signal unrelated processes (including the agent). Instead I resolve the PGID and only group-signal when `pgid == pid` (leader), falling back to signalling the individual PID otherwise. That fixes the false-success-on-ESRCH concern you raised without the own-group hazard.

CI green on all four Smoke jobs (ubuntu/macos/windows) + vet, host/linux/windows builds.

@coderabbitai full review

gnanam1990:
Addressed re-review round 2 in `89ca8d8` (batch 38), each with a regression test:

- **safe_command**: a `)` is only a substitution boundary when it closes an active, *unquoted* `$(...)` — a quote-state stack runs each substitution in a fresh quoting context and restores the outer quoting on its matching `)`. Fixes `echo $(printf "a) less")` while still isolating `echo "$(vim x)"`.
- **egress**: in-flight domain prompts deduped per `host:port` (singleflight) — concurrent requests wait on the first prompt and reuse its decision; race-tested with 25 concurrent authorizers → one prompt.
- **analyzer + safe_command**: wrapper option-value consumption is now wrapper-specific, so `sudo -n rm -rf` / `sudo -n curl` are no longer mis-parsed (the valueless `-n` can't swallow the payload) while `sudo -u root vim` still consumes its value. Shared by the AST analyzer and regex detector.

CI green; `go vet` + `-race` on sandbox/background clean.

@coderabbitai full review

gnanam1990:
Addressed re-review round 3 in `cd9abe0` (batch 39) — all 12 findings verified against current code, each behavioural fix with a regression test:

- **web_search**: redirect policy now refuses a scheme change too (blocks same-host https→http downgrade that would leak query + bearer token).
- **safe_command**: `)` is only a boundary when it closes an active *unquoted* `$(...)`; backticks get the same quote-state save/restore as `$(` — `echo "`a | less`"` and `echo "$(true | less)"` no longer hide the inner pager.
- **analyzer + safe_command**: wrapper option-value stays wrapper-specific; unwrapping no longer aborts on a dynamic (`$x`) wrapper arg once a wrapper is active (`env "$opts" curl` still classified).
- **egress**: prompt callback is now cancellable (context cancelled on timeout) so a stuck prompt aborts instead of leaking; in-flight dedup per host:port.
- **runner**: per-plan unique denial tag (pid+counter) so concurrent monitored runs can't cross-ingest denials; seccomp helper selection requires an executable regular file.
- **mcp/network_client**: aggregate SSE event-size cap (not just per-line).
- **specialist/accounting**: a recorded stop/usage with empty runId is a catch-all and suppresses a later runId-bearing duplicate.
- **sessions/lineage**: `Tree` fetches the root via `Get()` first so corruption surfaces its real error instead of "not found".
- **background/terminate** docstring + **cli/sandbox** usage strings + test ESRCH probe.

CI green; `go vet` + `-race` on sandbox/background clean.

@coderabbitai full review

gnanam1990:
Addressed re-review round 4 in `1eba7a5` — verified each against current code; 8 fixed, 1 skipped:

**Fixed**
- **sessions/lineage** — *panic fix*: `store.Get` returns `(nil, nil)` for a missing session, so the batch-39 root prefetch could nil-deref. Now guards `root == nil` → not-found error. + regression test (was a panic).
- **agent/compaction** — reactive `recover()` records `lowWaterMark` in the same combined (messages + tool-defs) domain `maybeCompact` uses; `recover()` now takes `tools`, callers pass `request.Tools`.
- **mcp/network_client** — SSE aggregate counter no longer counts a joining newline for the first `data:` line (only when prior content exists).
- **sessions/store** — fsync the parent directory after the metadata rename (no-op on Windows, which can't fsync a dir).
- **sandbox/egress** — port parsed with `strconv.Atoi`, not `net.LookupPort`.
- **sandbox/runner** — per-plan denial tag escaped via `sandboxProfileString` in the profile.
- **sandbox/seccomp_linux** — guard empty filter before `&kernelFilters[0]`.
- **cli/cron_run_test** — assert last-error-wins with two error events.

**Skipped (with reason)**
- **zerogit worktree-rename (Y='R') test**: git porcelain v1 `-z` only emits the rename-source field when the *index* column is R/C; the worktree column never carries 'R' with a source field in v1. The fixture would test output git doesn't produce, and making `parseStatus` consume on `code[1]=='R'` would risk mis-consuming the following entry for a plain ` R` status. Left as-is.

CI green; `go vet` + `-race` on sandbox/background clean.

@coderabbitai full review

gnanam1990:
Addressed re-review round 5 in `5cf23d4` — all 4 findings verified against current code:

- **agent/compaction**: `summarizeWithFallback`'s reduce pass no longer swallows *all* errors. A context-limit error still falls back to the joined partial summaries (the documented extreme case); a non-context failure (auth/network/provider) now surfaces unchanged per the contract. Two tests (propagate vs. joined-fallback); the propagate test fails without the fix.
- **sandbox/safe_command + analyzer**: wrapper value-consuming options now include the long spellings (`sudo --user root`, `env --unset NAME`, `timeout --signal …`), so a long flag's value is no longer mis-read as the program. A `--flag=value` token carries its own value and never consumes the next token. Tests for space and = forms in both the AST analyzer and the regex detector.
- **docs/audit (both findings)**: the status-ledger summary table's `open/partial` cells (64/14) were the lone outlier — the ledger's own narrative and `2026-06-13-reverification.md` both state 66/12. Corrected to **66/12** (66+12+97 = 175). Derivation: body 69 open + 18 partial − the 9 already-fixed-on-main (3 open + 6 partial) = 66/12 = 78 pending; fixed 88+9 = 97. Added a comment documenting the summary-vs-body relationship. Both docs now agree, so the cross-file finding is resolved too.

CI green; `go vet` clean.

@coderabbitai full review

gnanam1990:
Round 6 (`8ec1496`). CI was already green — CodeRabbit's 6th pass was a COMMENT with no new findings, re-surfacing two "duplicate" refinements:

- **`cli/sandbox_test.go` (assert immutability per call)** — done, and strengthened: the test now seeds a tool-wide grant and asserts `reflect.DeepEqual(before, after)` around *each* rejected empty-`--path` call, so a mutation in one call can't be masked by a later one (and a buggy "revoke all" is observable — it's a silent no-op on an empty store).

- **`zerogit.parseStatus` (consume rename source when R/C in *either* column)** — respectfully declined as incorrect for porcelain v1, with an explanatory code comment. Verified empirically:

  ```
  $ git mv a.txt b.txt && git status --porcelain -z   →  R  b.txt\0a.txt\0     (R in the INDEX column)
  $ mv c.txt d.txt      && git status --porcelain -z   →   D c.txt\0?? d.txt\0  (delete + untracked — NOT "R" in the worktree column)
  ```

  Porcelain v1 emits the extra source field only when the **index** column (`code[0]`) is `R`/`C`; the worktree column never carries `R`/`C` with a source field. Consuming on `code[1]=='R'/'C'` would match no real git output and would only risk mis-consuming the next entry on malformed input, so the current index-column-only check is both correct and more robust.

gofmt + vet clean; host/linux/windows builds; suites green.

@coderabbitai review

PR #190, opened 2026-06-13
Polish model status and reasoning streams

anandh8x:
Summary:
- show the configured provider profile name in /model and /mode status output
- parse OpenAI-compatible reasoning_content stream deltas as visible text
- add a focused provider test for reasoning_content chunks

Notes:
- empty OpenAI-compatible message content behavior is unchanged from main
- .config/ and examples/ are local untracked files and are not included

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/providers/openai ./internal/tui
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Streamed "reasoning" content is emitted as separate text events.
  * Onboarding/setup accepts custom endpoint and profile name; supports typed-model entry and endpoint validation.

* **UI**
  * Provider display uses endpoint-derived labels in title/status and confirmations.
  * Composer: visual-line cursor movement, paste-preview summaries, updated placeholders, and mouse-wheel cursor scrolling over composer.

* **Behavior**
  * CLI options can override saved provider name/base URL.
  * File-suggestion completion/dismissal preserves collapsed paste previews.

* **Tests**
  * Expanded coverage for streaming, setup/onboarding, composer, rendering, autocomplete, mouse and wizard flows.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -144,8 +144,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		}
 
 		collected := zeroruntime.CollectStreamWithOptions(ctx, stream, zeroruntime.CollectOptions{
-			OnText:  options.OnText,
-			OnUsage: options.OnUsage,
+			OnText:      options.OnText,
+			OnReasoning: options.OnReasoning,
+			OnUsage:     options.OnUsage,
 		})
 		if collected.Error != "" {
 			// REACTIVE compaction: the streamed error may also be a context
@@ -445,8 +446,9 @@ func finalAnswerAfterMaxTurns(ctx context.Context, provider Provider, messages [
 		return "", messages, ""
 	}
 	collected := zeroruntime.CollectStreamWithOptions(ctx, stream, zeroruntime.CollectOptions{
-		OnText:  options.OnText,
-		OnUsage: options.OnUsage,
+		OnText:      options.OnText,
+		OnReasoning: options.OnReasoning,
+		OnUsage:     options.OnUsage,
 	})
 	if ctx.Err() != nil || collected.Error != "" || strings.TrimSpace(collected.Text) == "" {
 		return "", messages, ""
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -89,6 +89,39 @@ func TestRunReturnsProviderText(t *testing.T) {
 	assertMessage(t, provider.requests[0].Messages[1], zeroruntime.MessageRoleUser, "say hi")
 }
 
+func TestRunDoesNotPersistReasoningAsAssistantText(t *testing.T) {
+	provider := &mockProvider{
+		turns: [][]zeroruntime.StreamEvent{{
+			{Type: zeroruntime.StreamEventReasoning, Content: "private reasoning"},
+			{Type: zeroruntime.StreamEventText, Content: "public answer"},
+			{Type: zeroruntime.StreamEventDone},
+		}},
+	}
+	var reasoning []string
+
+	result, err := Run(context.Background(), "say hi", provider, Options{
+		Registry:    tools.NewRegistry(),
+		OnReasoning: func(delta string) { reasoning = append(reasoning, delta) },
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+	if result.FinalAnswer != "public answer" {
+		t.Fatalf("final answer = %q, want public answer", result.FinalAnswer)
+	}
+	if len(result.Messages) == 0 {
+		t.Fatal("expected persisted messages")
+	}
+	last := result.Messages[len(result.Messages)-1]
+	if last.Role != zeroruntime.MessageRoleAssistant || last.Content != "public answer" {
+		t.Fatalf("assistant message = %#v, want answer-only assistant content", last)
+	}
+	if len(reasoning) != 1 || reasoning[0] != "private reasoning" {
+		t.Fatalf("reasoning callbacks = %#v", reasoning)
+	}
+}
+
 func TestRunReportsTruncationFinishReason(t *testing.T) {
 	provider := &mockProvider{
 		turns: [][]zeroruntime.StreamEvent{{
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -184,6 +184,7 @@ type Options struct {
 	EnabledTools        []string
 	DisabledTools       []string
 	OnText              func(string)
+	OnReasoning         func(string)
 	OnToolCall          func(ToolCall)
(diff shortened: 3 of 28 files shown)


anandh8x:
@coderabbitai review the latest commits that u havent reviewed

anandh8x:
@coderabbit can u do a full re review now, checking unchecked commits only is fine

Vasanthdev2004 (approved):
Verdict: APPROVED

Findings: none blocking. I reviewed the current head (`81930bc`) across the reasoning-content streaming changes, custom provider onboarding/wizard persistence, provider display normalization, composer paste-preview/autocomplete behavior, and mouse-wheel composer hit-testing. The earlier CodeRabbit concerns around non-local HTTP endpoints and clipped-footer wheel hit-testing are addressed in this head.

Validation performed locally:

- `go test ./...`

Remote status is also green: CI smoke checks, Performance Smoke, Zero Review, and CodeRabbit are passing.

gnanam1990 (changes requested):
## Review: model status + reasoning streams

Nice polish overall — I reviewed the streaming, composer/mouse, and onboarding/wizard layers with verification. Build is clean and `go test ./...` passes. The composer and onboarding work is well-built and safe (details below). Requesting changes for one streaming issue; the rest are minor/nits.

### 🟠 Should fix — reasoning content leaks into persisted message history

`internal/providers/openai/provider.go:232` emits `reasoning_content` deltas as `StreamEventText`:

```go
if choice.Delta.ReasoningContent != "" {
    sendEvent(ctx, events, zeroruntime.StreamEvent{Type: zeroruntime.StreamEventText, Content: choice.Delta.ReasoningContent})
}
```

`CollectStream` folds every `StreamEventText` into `collected.Text` (`internal/zeroruntime/helpers.go:113`), and the agent loop then uses `collected.Text` as the **persisted assistant message content** (`internal/agent/loop.go:210`, replayed to the model on every subsequent turn) and as the **final answer** (`loop.go:247`).

This diverges from the repo's own contract for the other two provider adapters, which keep reasoning out of the answer/message text:
- the gemini adapter explicitly **skips thought parts** — `// Skip thought summary parts: their text is reasoning, not the answer.` (`internal/providers/gemini/provider.go:240`);
- the anthropic adapter routes thinking to `ReasoningBlocks`, surfaced on the done event, never into `collected.Text`.

There's already a dedicated channel for this — `collected.ReasoningBlocks` / `Message.Reasoning` (`loop.go:214`) — which this adapter bypasses.

**Consequence:** on multi-turn / tool-using runs (the normal agent path), the model's chain-of-thought is re-sent as committed assistant content every turn (token bloat + the model treating its scratch reasoning as prior output), and the saved answer becomes `reasoning + answer` concatenated.

**Fix:** show reasoning live, but route it through a separate event/channel so it does **not** enter `collected.Text` / the persisted message `Content` — mirror the gemini (skip) or anthropic (`ReasoningBlocks`, `Provider: "openai"`) pattern. The new `TestStreamCompletionEmitsReasoningContentDeltas` asserts the concatenated behavior, so it should be updated to assert reasoning does not land in the answer text (and add a mixed `reasoning_content`+`content` delta case).

### 🟡 Minor — onboarding/setup
- **Endpoint validation accepts embedded credentials and private/metadata hosts over https** (`internal/tui/provider_wizard.go:~989`, `providerWizardEndpointError`): `https://user:pass@host/v1` and `https://169.254.169.254/v1` both pass. Discovery is safely skipped for custom endpoints (no SSRF) and the provider label strips userinfo (no leak there), but a `user:pass@` base URL is persisted in plaintext and sent at request time. Recommend rejecting `parsed.User != nil`.
- **CLI `zero setup <provider> --base-url <url>` skips the wizard's validation** (`internal/cli/setup.go`): an invalid/ill-formed URL is written to config and only fails later at request time. Share the wizard's validation (or `providerio.NormalizeBaseURL`) on this path.

### Nits
- Dead code after the paste-preview refactor: `deleteCompletedFileMentionBefore`, `deleteComposerLineBefore`, `deleteComposerLineAfter` (`internal/tui/composer.go`) have no callers — delete them.
- `mouseOverComposer` (`internal/tui/mouse.go:163`) only bounds `msg.Y`, not `msg.X`, so a wheel scroll far left/right of the composer box still moves the cursor; it also locates the composer row by matching the rendered string within the footer (brittle). Bound X and/or expose the row offset directly.

### Verified safe (high-value checks)
- **No SSRF during onboarding discovery** — custom (user-supplied-URL) providers skip discovery entirely; built-in providers discover against their hardcoded default base URL.
- **No credential leak via the provider display label** — derived from `parsed.Hostname()`, which strips any `user:pass@`.
- **Composer is panic/rune-safe** — cursor math is on `[]rune` with clamps throughout; the collapsed paste preview keeps the full bytes for submission and the cursor can't land inside a hidden preview (verified by stress testing — no panics, no buffer corruption).
- **Config writes** preserve other config (read-merge-write, atomic temp+rename, key file 0600); CLI override precedence (`--name`/`--base-url` win over saved) is correct.

**Verdict:** Solid work — composer and onboarding are well-built and the SSRF/credential-leak checks come out clean. The one item worth fixing before merge is the reasoning-into-history leak, since it contradicts how the two sibling provider adapters in this repo handle thinking; the rest can be follow-ups. Happy to re-review quickly.

gnanam1990 (approved):
## Re-review — reasoning leak fixed ✅

Thanks for `38da8f3`. I verified the fix end-to-end:

- New dedicated `StreamEventReasoning` type (with a comment that it must never be folded into answer text or persisted as assistant content).
- The OpenAI adapter now emits `reasoning_content` as `StreamEventReasoning`, not `StreamEventText`.
- `CollectStreamWithOptions` routes it to a new `OnReasoning` callback and never appends it to `collected.Text` — so it can no longer reach the persisted message `Content` (`loop.go:210`) or the final answer (`loop.go:247`).
- `OnReasoning` is wired through agent `Options` into the TUI, so reasoning is still shown live (the feature) while staying out of history — matching the gemini (skip) / anthropic (`ReasoningBlocks`) contract.
- Good regression coverage: `TestRunDoesNotPersistReasoningAsAssistantText` asserts answer-only `FinalAnswer`/message `Content` with reasoning delivered via `OnReasoning`, plus provider tests for "no text events" and reasoning-before-content ordering.

`gofmt`/`go vet` clean, `go test ./...` passes. Approving.

The earlier minors are fine as follow-ups whenever convenient: reject `user:pass@` / validate scheme in the endpoint check (and share that validation with the `zero setup --base-url` CLI path), drop the now-unused composer helpers (`deleteCompletedFileMentionBefore`, `deleteComposerLineBefore/After`), and bound `mouseOverComposer` on `msg.X`. Nice work.

PR #191, opened 2026-06-14
Wire LSP diagnostics into self-correct (exec + TUI)

gnanam1990:
## Summary

Wires the **LSP-diagnostics half** of the post-edit self-correct loop. It was built and
tested but never connected — the only consumer passed a `nil` checker with
`IncludeLSP=false`, which left the entire `internal/lsp` client unreachable.

- **exec** (`internal/cli/exec.go`): `newExecSelfCorrector` now backs `--self-correct` with
  both halves — the workspace test plan **and** an `lsp.Manager`-based diagnostics checker
  (`IncludeLSP=true`). The manager is lazy (`Manager.Check` degrades a missing/unsupported
  language server to `nil, nil`) and a deferred cleanup shuts it down at run end.
- **TUI** (`internal/tui/model.go`): self-correct now runs **by default every turn**, but
  kept **fast** — only the LSP-diagnostics half, scoped to the changed files. The whole-repo
  test plan is **opt-in** per session via a new **`/selfcorrect [on|off]`** command (alias
  `/sc`), so it never adds `go test ./...` latency to a turn and a pre-existing failure can't
  hijack the agent. Spec-draft (planning) is excluded; the per-turn `lsp.Manager` is torn
  down at run end; auto-fix vs report-only follows the active permission mode
  (`unsafe→high`, `auto→medium`, `ask→low`).
- Adds `TestManagerCheckRealGopls`: drives `NewManager → Check` against a real gopls and
  asserts a real diagnostic; skips when gopls is absent, so it stays CI-safe.

## Effect

Revives `internal/lsp` (**44 → 3** unreachable funcs); total prod-unreachable **178 → 135**.

## Why fast-by-default in the TUI

The test half (`go test ./...`) is whole-project and slow (up to a 120s timeout); the LSP
half is change-scoped and cheap. Running the full suite after every interactive edit would
add the suite's latency to each turn and re-surface pre-existing failures every time. So the
TUI defaults to LSP-only and gates the test half behind `/selfcorrect on`. `exec` keeps both
halves (it is headless and already opt-in via `--self-correct`).

## Verification

- `go build ./...`, `go vet`, `gofmt` clean; full `go test ./...` green.
- The real-gopls test returns the expected compiler diagnostic through the wired path.
- Manual end-to-end: a `--self-correct` run surfaces and fixes a seeded compile error that
  the baseline run leaves broken; the `gopls serve` process spawns mid-run and is cleaned
  up at the end.

## Commits

1. exec LSP wiring + real-gopls test
2. TUI wiring
3. TUI fast-by-default + `/selfcorrect` toggle

## Summary by CodeRabbit

* **New Features**
  * Added a TUI **`/selfcorrect`** command (alias **`/sc`**) to control post-edit self-correction depth, including **LSP-only** diagnostics or broader test-plan verification.
* **Bug Fixes**
  * Improved self-correction verification by collecting **LSP diagnostics** during the post-edit loop and ensuring the **LSP service shuts down cleanly** after runs.
  * Updated LSP document close handling to avoid lingering session state.
* **Tests**
  * Added an integration-style manual test that runs diagnostics using the real **`gopls`** binary (skips if not available).

### DIFF
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Gitlawb/zero/internal/agent"
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/imageinput"
+	"github.com/Gitlawb/zero/internal/lsp"
 	"github.com/Gitlawb/zero/internal/modelregistry"
 	"github.com/Gitlawb/zero/internal/notify"
 	"github.com/Gitlawb/zero/internal/providers"
@@ -462,10 +463,12 @@ func runExec(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 	defer stopSignals()
 	// --self-correct opts the run into the post-edit verify-and-correct loop. Off
 	// by default the corrector is nil, leaving the agent loop byte-identical. When
-	// on we verify with the workspace test plan (the LSP half stays nil here since
-	// headless exec does not run a language-server manager); the autonomy gate
-	// inside the corrector still decides whether failures auto-fix or just report.
-	selfCorrector := newExecSelfCorrector(options.selfCorrect, workspaceRoot, options.autonomy)
+	// on we verify with both the workspace test plan and LSP diagnostics over the
+	// changed files; the autonomy gate inside the corrector still decides whether
+	// failures auto-fix or just report. lspShutdown tears down any language-server
+	// sessions the LSP half spawned (no-op when self-correct is off).
+	selfCorrector, lspShutdown := newExecSelfCorrector(options.selfCorrect, workspaceRoot, options.autonomy)
+	defer lspShutdown()
 	result, err := agent.Run(runCtx, agentPrompt, provider, agent.Options{
 		MaxTurns:         resolved.MaxTurns,
 		ContextWindow:    modelContextWindow(modelRegistry, resolved.Provider.Model),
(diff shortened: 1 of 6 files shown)


gnanam1990:
Addressed the `/selfcorrect` help-text finding in d680258 — `usage` is now `[on|off|status]` and the description clarifies the on/off behavior, so help matches the handler. Aliases (tests/full/lsp/list) still work; the usage surfaces the primary verbs.

@coderabbitai review

Vasanthdev2004 (changes requested):
Verdict: **Request changes**

The runtime wiring looks good and validation is green, but one user-facing help issue is still present:

- `internal/tui/commands.go:208` documents `/selfcorrect [on|off]`, while `handleSelfCorrectCommand` also accepts `status`, `list`, `tests`, `full`, and `lsp`.
- `internal/tui/session_controls.go:197` returns the same incomplete usage text for invalid arguments.

Please update both usage strings to reflect the accepted forms, for example:

`/selfcorrect [status|on|off|tests|full|lsp]`

Validation I ran locally on the PR head:

- `go test ./internal/cli ./internal/tui ./internal/lsp ./internal/agent -run 'SelfCorrect|LSP|Manager|Command|Exec' -count=1 -timeout 180s`
- `go test ./internal/cli ./internal/tui ./internal/lsp -count=1 -timeout 240s`

Once the usage text is fixed, I do not see another blocker from this pass.

gnanam1990:
@Vasanthdev2004 thanks — fixed in e746120. Both usage strings now list the full accepted set and match the handler exactly:

- `internal/tui/commands.go`: usage → `/selfcorrect [status|on|off|tests|full|lsp]`
- `internal/tui/session_controls.go`: the invalid-argument error message now returns the same string (this was the spot the first fix missed)

Dropped the redundant `list` alias so the documented forms == the handler's accepted forms exactly. `go test ./internal/cli ./internal/tui ./internal/lsp ./internal/agent` green.

@coderabbitai review

gnanam1990:
Addressed the description nitpick in 0b634f5 — `/selfcorrect` help now reads "off/lsp disable tests, LSP-only" instead of the ambiguous "return to LSP-only".

@coderabbitai review

Vasanthdev2004 (approved):
Verdict: **Approved**

The previous requested change is fixed on the latest head (`e746120`):

- `internal/tui/commands.go` now documents `/selfcorrect [status|on|off|tests|full|lsp]`.
- `internal/tui/session_controls.go` now returns the same usage string for invalid args.
- The handler accepts exactly those documented args now (`status`, `on`, `off`, `tests`, `full`, `lsp`).

I rechecked the LSP self-correct wiring and did not find a functional blocker. One note: CodeRabbit has a fresh wording nit on the `/selfcorrect` description (`off/lsp return to LSP-only` could be clearer), but I do not consider that a blocker from this review.

Validation run locally on PR head:

- `go test ./internal/cli ./internal/tui ./internal/lsp ./internal/agent -run 'SelfCorrect|LSP|Manager|Command|Exec' -count=1 -timeout 180s`
- `go test ./internal/cli ./internal/tui ./internal/lsp -count=1 -timeout 240s`
- `go vet ./...`
- `go build ./...`

PR #192, opened 2026-06-14
feat: add backend lifecycle status command

Vasanthdev2004:
## Summary
- Add `zero backends` / `zero backend` to inspect MCP, hook, and plugin lifecycle status from one command.
- Return the existing `zerocommands.BackendLifecycleSnapshot` as stable JSON, plus a compact text view.
- Keep the command offline-safe: it reads config/manifests only and does not register or connect MCP tools.
- Harden backend snapshots so MCP commands, URL query/fragment secrets, hook fields, and plugin metadata/path strings are redacted before JSON output.

## Validation
- `gofmt -l internal\cli internal\zerocommands`
- `go vet ./...`
- `go build ./...`
- `go run ./cmd/zero backends`
- `go run ./cmd/zero backends --json`

## Summary by CodeRabbit

* **New Features**
  * Added `zero backends` CLI subcommand to display backend lifecycle information including hooks, plugins, and MCP servers
  * Supports `--json` output format for programmatic access and `--help` for usage information
  * Automatically redacts sensitive credentials, tokens, and secrets from output

* **Tests**
  * Added CLI regression tests for the backends command
  * Added comprehensive redaction validation tests for sensitive data

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -264,6 +264,8 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 		return runSpecialists(args[1:], stdout, stderr, deps)
 	case "plugins", "plugin":
 		return runPlugins(args[1:], stdout, stderr, deps)
+	case "backends", "backend":
+		return runBackends(args[1:], stdout, stderr, deps)
 	case "skills", "skill":
 		return runSkills(args[1:], stdout, stderr, deps)
 	case "tools", "tool":
@@ -648,6 +650,7 @@ Commands:
   spec       Review and approve saved spec-mode drafts
   specialist Manage local Zero specialist profiles
   plugins    Inspect, install, and remove local Zero plugins
+  backends   Inspect MCP, hook, and plugin backend lifecycle state
   skills     Inspect, install, and remove local Zero skills
   tools      Scaffold and list local Zero plugin-tools
   hooks      Inspect Zero hook configuration
--- a/internal/cli/backends.go
+++ b/internal/cli/backends.go
@@ -1,0 +1,150 @@
+package cli
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/hooks"
+	"github.com/Gitlawb/zero/internal/mcp"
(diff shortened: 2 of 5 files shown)


gnanam1990 (approved):
**Approve** — solid PR, no blockers. CI is green across all platforms, the redaction is well-tested, and it wires the previously-dormant `zerocommands` snapshot layer into a real, offline-safe command.

### Strengths
- Wires the unreachable `BackendLifecycleSnapshot` layer into `zero backends`/`backend` — the right "wire" call for that subsystem.
- Offline-safe as described: reads config/manifests via the existing injected deps (`getwd`, `resolveMCPConfig`, `loadHooks`, `loadPlugins`); never registers or connects MCP tools.
- Consistent registration + help text; reuses `appDeps`, so the command stays unit-testable.
- Good redaction tests — secret MCP/hook commands, sensitive URL query values, and URL-without-credentials preservation.

### Non-blocking suggestions
1. **The `--json` path is the weaker redaction path.** The text path wraps output in `RedactString`, but `--json` serializes the snapshot directly, so it relies entirely on field-level redaction in the builder. Two small gaps there:
   - `MCPServerSnapshot.Name` and `.Identity` are only `TrimSpace`d (Command/URL are redacted). They're labels, so low risk, but JSON ≠ text here.
   - URL-query redaction is key-name heuristic only (`isSensitiveURLKey`): a secret under a non-obvious key (e.g. a signed-URL `?sig=…` parameter) keeps its value in JSON, since there's no value-shape fallback.
   - Consider running query values (and Name/Identity) through `redaction.RedactString` so the JSON output is as safe as the text output.
2. `stripURLCredentialsFromURL` now round-trips **every** URL through `url.Parse → String()` (previously only when userinfo was present), which can normalize a verbatim URL. It's guarded and covered by the `…PreservesURLWithoutCredentials` test, so benign — just flagging the behavior change.
3. Nit: fragment redaction re-encodes via `String()`, so `[REDACTED]` becomes `%5BREDACTED%5D` in a fragment. Cosmetic; fragments are rare in MCP URLs.

None of these block merge — #1 is defense-in-depth and low-likelihood for real MCP configs. Nice work wiring this in.

PR #194, opened 2026-06-14
feat: add backend doctor diagnostics

Vasanthdev2004:
## Summary
- Add `zero backends doctor` with text and JSON diagnostics for MCP, hooks, and plugins.
- Keep the doctor path offline-safe: reads local config/manifests only and never registers/connects MCP tools or executes hooks/plugins.
- Add actionable next commands for each check, plus pass/warn/fail status and proper exit behavior.
- Harden backend snapshots by omitting secret-derived MCP identity and redacting malformed URL text.

## Stacking
This PR is stacked on `feat/backend-status-contract` / #192. After #192 merges, this can be retargeted to `main` or rebased.

## Validation
- `gofmt -l internal\cli internal\zerocommands`
- `go vet ./...`
- `go build ./...`
- `go run ./cmd/zero backends doctor`
- `go run ./cmd/zero backends doctor --json`

## Summary by CodeRabbit

* **New Features**
  * Added `zero backends doctor` to check backend health across MCP servers, hooks, and plugins
  * Produces pass/warn/fail summaries with optional remediation actions
  * Supports both `--json` output and human-readable text, including updated command help
* **Bug Fixes**
  * Strengthened sanitization for backend diagnostic snapshots by omitting identity fields and redacting secret-like URL contents, even for malformed URLs

### DIFF
--- a/internal/cli/backends.go
+++ b/internal/cli/backends.go
@@ -17,6 +17,13 @@ type backendStatusOptions struct {
 }
 
 func runBackends(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
+	if len(args) > 0 {
+		switch args[0] {
+		case "doctor":
+			return runBackendsDoctor(args[1:], stdout, stderr, deps)
+		}
+	}
+
 	options, help, err := parseBackendsArgs(args)
 	if err != nil {
 		return writeExecUsageError(stderr, err.Error())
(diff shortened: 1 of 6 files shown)


gnanam1990 (approved):
**Approve** — clean, offline-safe diagnostics with good tests and correct exit behavior. One redaction note below (non-blocking).

### Strengths
- Well-structured: pass/warn/fail aggregation, a per-check `action`, and correct exit codes (`exitProvider` on failure) in both text and JSON modes.
- **Offline-safe** as claimed — `backendDoctorReport` reads config/manifests (`resolveMCPConfig` / `loadHooks` / `loadPlugins`) and validates MCP via `NormalizeConfig`; it never connects MCP or executes hooks/plugins. The help text matches ("without connecting to external MCP").
- `Target` / `Message` / error text are redacted (`redactSnapshotString` / `RedactString`).
- Snapshot hardening: drops the secret-derived `Identity` (kept with `omitempty`, unpopulated) and now redacts malformed-URL text — addresses part of the earlier #192 feedback.
- Tests cover the fail paths (invalid MCP, hook/plugin diagnostics) plus secret redaction; CI is green on all platforms.

### Finding — Minor (redaction consistency)
`backends doctor --json` emits the `Details` map verbatim: `writePrettyJSON(stdout, report)` isn't wrapped in `RedactValue`, and the hook/plugin checks put **raw** filesystem paths in `Details` (`path`, `root`, `pluginPath`, `manifestPath`). The text path doesn't render `Details` at all, so this is JSON-only — and it's inconsistent with this PR's own model, where `PluginSnapshotFromPlugin` redacts `Root`/`ManifestPath` via `redactSnapshotString`.
- These are local paths (low sensitivity), so not a blocker — but for consistency + defense-in-depth, run the `Details` values through `redactSnapshotString` (or wrap the doctor JSON payload in `redaction.RedactValue`) so `--json` matches the snapshot redaction model.

### Optional (follow-up to #192)
The malformed-URL path is now redacted (good). For well-formed URLs, query redaction is still key-heuristic-only (`isSensitiveURLKey`), so a secret under a non-obvious key (e.g. a signed-URL `?sig=…`) still survives — same optional note as #192.

Nothing here blocks merge — nice, well-tested feature.

PR #196, opened 2026-06-14
Solidify sandbox: harden command analysis + wire/clean dormant policy surfaces

gnanam1990:
## Summary

Solidifies the sandbox (the safety boundary): hardens command analysis and resolves every dormant policy surface so there is no half-built code in the sandbox.

### 1. Harden command risk analysis — wire the AST analyzer
The AST-based analyzer (`analyzer.go`) was built and tested but never reached at runtime; the live classifier used only regexes. Wire `AnalyzeCommand` into `classifyWithScope` as a second opinion. It walks the parsed shell tree, so it catches risk the regexes miss — `shred`/`fdisk`/`parted`, network programs like `telnet`/`ftp`/`sftp`, commands hidden behind `sudo`/`env` wrappers or a `sh -c` launcher — and flags unparseable (obfuscated) scripts. **Purely additive** (only raises risk); AST precision means a quoted program name (`echo "shred …"`) is not flagged.

### 2. Wire the auto-allow-bash env opt-in
`AutoAllowBashWhenSandboxed` is **enforced live** (`engine.go:283`) but was **never set** — no config field, and the `ZERO_SANDBOX_AUTO_ALLOW_BASH` env overlay (`ApplyAutoAllowBashEnv`) was unwired, so the documented opt-in did nothing. Wire it through the central policy builder (`applyConfiguredSandboxPolicy`, used by all engine constructions). **Fail-safe:** only turns ON when the env is truthy, only honored when the shell sandbox is actually active, default stays *prompt*, and an explicit config opt-in is preserved.

### 3. Drop orphaned batch-autonomy
`RequiredAutonomyForBatch`/`riskToAutonomy` had no consumer — the agent loop enforces autonomy **per tool call** (`loop.go`), so a batch-ceiling helper was speculative and redundant. Deleted (wiring it would have meant fabricating a batch model).

## Result
`internal/sandbox` unreachable-from-prod funcs **13 → 2**, and the remaining 2 are the cross-platform seccomp pair (`unixSocketBlockFilter` is live on Linux; `ApplyUnixSocketBlock` is the intentional non-Linux no-op stub). **No dormant funcs left; no security gap.**

## Verification
- New tests: AST catches regex-missed destructive/network + nested launchers, the unparseable signal, the quoted-name no-false-positive case, benign-stays-clean, and the auto-allow-bash env opt-in wiring (on/off).
- The **entire existing sandbox suite passes unchanged** (no intended-policy regression).
- `go build ./...`, `go vet`, `gofmt` clean; full `go test ./...` green; `go test -race ./internal/sandbox ./internal/cli` green.

## Summary by CodeRabbit

* **New Features**
  * Improved command risk classification using an additional second-pass analysis to better identify potentially dangerous network and destructive operations that earlier pattern checks could miss.
  * Added an environment-controlled opt-in for auto-allowing bash behavior when sandboxing is applied.

* **Bug Fixes / Security Hardening**
  * Expanded handling for edge cases like overly complex/unparseable commands to ensure they’re classified more conservatively.

* **Tests**
  * Added a hardened classification test suite covering AST-based scenarios, including quoted program names, unparseable commands, and benign commands staying unflagged.


Vasanthdev2004 (changes requested):
Verdict: **Request changes**

I found one still-valid issue before merge.

Finding:

- **[P2] Assert the risk level added by the AST analyzer tests** — `internal/sandbox/risk_hardening_test.go:236`, `:252`, `:265`

  These new tests only assert that the AST analyzer adds the expected categories (`destructive`, `network`, `unparseable_command`). The actual security contract in `classifyWithScope` is also the escalation level: destructive/network commands must become `RiskCritical`, and unparseable commands must become `RiskHigh`. As written, these tests would still pass if a future change accidentally downgraded destructive shell commands to medium/low while preserving the category, which weakens the guard this PR is trying to lock in.

  Please add level assertions alongside the category checks:
  - destructive AST detections: `risk.Level == RiskCritical`
  - network AST detections: `risk.Level == RiskCritical`
  - unparseable command: `risk.Level == RiskHigh`

  It would also be useful to include at least one wrapper case like `sudo shred -u secret.txt` or `env shred -u secret.txt`, since `effectiveProgram()` explicitly supports wrapper unwrapping.

Validation run locally on PR head `3a012a4`:
- `go test ./internal/sandbox -run ClassifyAST|ApplyAuto|Batch|Risk -count=1` -> pass
- `go test ./internal/cli -run ApplyConfiguredSandboxPolicyHonorsAutoAllowBashEnv|Sandbox -count=1` -> pass
- `go test ./internal/sandbox ./internal/cli -count=1` -> pass
- `go test ./... -timeout 300s` -> pass
- `go vet ./...` -> pass
- `go build ./...` -> pass
- `go run ./cmd/zero-release build` -> pass
- `go run ./cmd/zero-release smoke` -> pass

gnanam1990:
@Vasanthdev2004 thanks — addressed in c104065. The AST hardening tests now assert the escalation level as well as the category:

- destructive AST detections → `risk.Level == RiskCritical`
- network AST detections → `risk.Level == RiskCritical`
- unparseable command → `risk.Level == RiskHigh`

Added wrapper cases too (`sudo shred -u …`, `env shred -u …`, `sudo telnet …`) to cover `effectiveProgram`'s wrapper unwrapping. Full `go test ./...` + `-race` green.

@coderabbitai review

Vasanthdev2004 (approved):
Verdict: **Approved**

The requested changes are fixed on latest head `c104065`.

What I rechecked:
- AST destructive detections now assert both `destructive` and `RiskCritical`.
- AST network detections now assert both `network` and `RiskCritical`.
- Unparseable command now asserts `unparseable_command` and `RiskHigh`.
- Wrapper coverage was added for `sudo` / `env` paths, matching `effectiveProgram()` behavior.

No remaining blockers found.

Validation run locally:
- `go test ./internal/sandbox -run ClassifyAST|Unparseable|Risk -count=1` -> pass
- `go test ./internal/cli -run ApplyConfiguredSandboxPolicyHonorsAutoAllowBashEnv|Sandbox -count=1` -> pass
- `go test ./... -timeout 300s` -> pass
- `go vet ./...` -> pass
- `go build ./...` -> pass
- `go run ./cmd/zero-release build` -> pass
- `go run ./cmd/zero-release smoke` -> pass

GitHub checks were green except the remote Windows smoke job was still in progress when I posted this; local validation was run on Windows.

PR #197, opened 2026-06-14
feat: add TUI diagnostics center

Vasanthdev2004:
## Summary

- add a TUI diagnostics renderer for `/doctor` with grouped provider/platform checks and actionable hints
- add `/health` as a discoverable alias for `/doctor`
- support async `/doctor --connectivity` in the TUI using the configured provider health probe
- wire existing config paths and provider-health/user-agent context from CLI launch into TUI doctor
- make doctor model health friendlier for dynamic OpenAI-compatible providers by warning/pass-through instead of failing unknown custom models

## Tests

- `go test ./internal/tui -run Doctor|Health|Command -count=1`
- `go test ./internal/doctor -run ProviderModel -count=1`
- `go test ./internal/cli -run TestRunNoArgsLaunchesSetupTUIWithNilProviderWhenNoProviderConfigured|TestRunNoArgsLaunchesTUI -count=1`
- `go vet ./...`
- `go build ./...`

## Release Notes

* **New Features**
  * Enhanced `/doctor` command with `--connectivity` flag to check provider connectivity and health.
  * Added `/health` as an alias for the `/doctor` command.
  * Improved diagnostic output with structured sections for Provider, Platform, and Backend information.

* **Bug Fixes**
  * Provider model validation now properly fails when a required model is missing or empty.

* **Tests**
  * Comprehensive test coverage added for doctor command functionality and provider validation scenarios.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -415,6 +415,12 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	if err != nil {
 		return writeAppError(stderr, "failed to resolve user config path: "+err.Error(), 1)
 	}
+	doctorUserConfigPath := ""
+	projectConfigPath := ""
+	if resolveOptions, optErr := config.DefaultResolveOptions(workspaceRoot); optErr == nil {
+		doctorUserConfigPath = resolveOptions.UserConfigPath
+		projectConfigPath = resolveOptions.ProjectConfigPath
+	}
 
 	needsSetup := setupRequired(resolved)
 	setupVisible := forceSetup || needsSetup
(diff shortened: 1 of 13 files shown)


gnanam1990 (approved):
**Approve** — well-built, no blockers. I verified the security/correctness-sensitive parts:

### Async `--connectivity` — correct
- Plain `/doctor` renders synchronously and **offline** (`doctorOptions(false)` never probes). Only `--connectivity` probes.
- `--connectivity` returns a `tea.Cmd`, so the probe runs in bubbletea's goroutine — **off the Update loop** (no UI freeze). A `doctorCommandSeq` id on the producer + the `case doctorCommandResultMsg` guard (`msg.id == m.doctorCommandSeq`) correctly **discards superseded** probe results.
- The probe is **time-bounded**: `providerhealth.Probe` wraps the ctx in a 5s `context.WithTimeout` and the HTTP client has its own timeout, so a hung endpoint can't leave the "running…" state stuck.

### No secret leak in the new view
- `doctorCheckRow` renders `[status] id - message` only — it does **not** render `check.Details`, where the raw `baseURL` lives. Combined with the existing `credentialConfigured` (presence-only, never the key), the TUI doctor surfaces no secret material.

### Validation change is a strengthening, not a regression
- `provider.model` now **fails** on an empty model (previously unhandled), and the unknown-custom-model path still warns + passes through with a helpful `--connectivity` hint. No weakening.

### Tests / CI
- Good coverage (doctor command/view, health alias, model wiring, provider-model validation); full CI green incl. the Security gate.

### One optional, forward-looking note (non-blocking)
The TUI view intentionally omits `check.Details` (so `baseURL` isn't shown) — nice. If a future change ever renders Details in the TUI, strip userinfo from `baseURL` at that point (it can in rare cases embed `user:token@host`), to match the redaction model elsewhere.

Nice feature — clean async handling and good diagnostics ergonomics.

PR #202, opened 2026-06-14
fix(sandbox): web_search/web_fetch hard-denied under default network policy

gnanam1990:
## Problem
With the default sandbox policy (`network: deny`), `web_search` (and `web_fetch`) are **hard-denied** before they run:

```
denied · web_search · risk:high — network access is blocked by sandbox policy
Sandbox violation [network] for web_search: network access is blocked by sandbox policy
```

#199 added `EnforceToolNetwork` (off by default) to exempt the first-party in-process network tools, but it only relaxed `NetworkHostAllowed` — the per-host check *inside* the tool's `RunWithSandbox`. The actual deny happens **earlier**, at the engine-level gate in `Evaluate` (run by the tool registry): a request whose risk includes `network` under `NetworkDeny` is rejected *before* the tool executes. So the tool never reached the relaxed check. (#199 merged at the commit just before this follow-up, so the gap shipped to `main`.)

## Fix
`Evaluate` now skips the network-deny for a request that declares `SideEffectNetwork` (the first-party in-process tools `web_search`/`web_fetch`) when `EnforceToolNetwork` is off — mirroring the `NetworkHostAllowed` exemption so the engine-level gate and the per-tool gate agree. A **shell** command merely classified as network (`SideEffect=shell`, e.g. `curl`) is **not** exempt, so sandboxed-shell egress stays blocked under deny. Setting `EnforceToolNetwork: true` restores the strict deny for the tools.

Net effect: `web_search`/`web_fetch` are no longer hard-blocked by the sandbox; they follow the normal permission flow (prompt in ask mode → run on allow; run in auto mode).

## Tests
- `TestEvaluateExemptsNetworkToolsFromDenyByDefault` — full path: default → not denied; granted → `allow`; `EnforceToolNetwork=true` → denied; network **shell** command → stays denied.
- `TestEngineDeniesNetworkSideEffectWhenPolicyBlocksNetwork` and `TestScopedNetworkGateInEvaluate` updated to opt into `EnforceToolNetwork` (they test the strict gate).

## Gates (local)
`gofmt -l .` empty · `go vet` · `go build` · `go test ./...` (incl. `-race` on sandbox+tools) · `staticcheck` no new findings · `govulncheck` 0 · `deadcode -test=false` no new unreachable funcs.

Scoped to the sandbox fix only — 3 files (`engine.go`, `engine_test.go`, `runner_scoped_test.go`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Network tools are now exempt from network-deny policies by default.

### DIFF
--- a/internal/sandbox/engine.go
+++ b/internal/sandbox/engine.go
@@ -149,6 +149,19 @@ func (engine *Engine) NetworkHostAllowed(host string) (bool, NetworkMode) {
 	}
 }
 
+// toolNetworkExempt reports whether a request is exempt from the engine-level
+// network deny because it is a first-party, in-process network TOOL — one that
+// declares SideEffectNetwork (web_search / web_fetch) — and the operator has not
+// opted into EnforceToolNetwork. Such tools do not use the sandboxed shell's
+// egress; they keep their own SSRF/host safeguards, and NetworkHostAllowed (also
+// gated by EnforceToolNetwork) governs them at run time. A SHELL command merely
+// classified as network (SideEffectShell) is NOT exempt, so shell egress stays
+// blocked under deny. Mirrors the NetworkHostAllowed exemption so the Evaluate
+// gate and the per-tool gate never diverge.
+func (engine *Engine) toolNetworkExempt(policy Policy, request Request) bool {
+	return !policy.EnforceToolNetwork && request.SideEffect == SideEffectNetwork
+}
+
 // scopeFor returns the scope to validate request paths against. The engine's
 // shared scope applies only when the request targets the engine's own
 // workspace root; a per-request override root gets an ad-hoc single-root scope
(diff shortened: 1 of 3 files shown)


anandh8x (approved):
Reviewed locally. No blocking findings.

The change keeps Evaluate aligned with NetworkHostAllowed: first-party SideEffectNetwork tools remain exempt by default when EnforceToolNetwork is off, while shell network risk and explicit EnforceToolNetwork still deny. The added tests cover the default exemption, strict opt-in, shell non-exemption, and scoped gate behavior.

Tests:
- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox -run 'TestEvaluateExemptsNetworkToolsFromDenyByDefault|TestEngineDeniesNetworkSideEffectWhenPolicyBlocksNetwork|TestScopedNetworkGateInEvaluate': pass
- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/tools: pass
- GOCACHE=/tmp/zero-go-cache go test ./...: fails in internal/imageinput TestLoadDocumentTruncatesLongText, outside this PR changed files.

PR #204, opened 2026-06-14
Polish assistant response transcript

anandh8x:
Summary:
- remove the final assistant response rail and green completion dot
- keep completion metadata as plain faint text, while preserving red failure markers
- keep the title bar workspace/branch visible by falling back to a shortened branch before dropping context
- update selectable transcript math and regression tests

Tests:
- git diff --check

## Summary by CodeRabbit

* **UI/Visual Changes**
  * Final assistant answers now render as plain text (removes decorative rail/dot styling).
  * Completion “done” lines now show timing/tool counter information as plain text.
  * Title bar improves layout for long branch names via truncation while keeping workspace/cwd and model details readable.
  * Footer rendering order is adjusted so copy status appears before attachment chips.

* **Tests**
  * Updated TUI rendering assertions for final answer and completion line formatting.
  * Added title bar width/long-branch test and updated copy-status/height regression coverage.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1131,13 +1131,14 @@ func (m model) transcriptView() string {
 
 func (m model) footerView(width int) string {
 	var footer strings.Builder
-	footer.WriteString("\n")
-	if chips := renderAttachmentChips(m.pendingImageLabels, m.pendingDocuments); chips != "" {
-		footer.WriteString(fitStyledLine(zeroTheme.muted.Render(chips), width))
-		footer.WriteString("\n")
-	}
 	if copyStatus := strings.TrimSpace(m.copyStatus); copyStatus != "" {
 		footer.WriteString(rightAlignedLine(zeroTheme.ink.Render(copyStatus), width))
+		footer.WriteString("\n")
+	} else {
+		footer.WriteString("\n")
+	}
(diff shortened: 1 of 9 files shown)


gnanam1990 (approved):
## Review — Polish assistant response transcript

Clean, self-consistent visual polish. The changes hang together well:
- Final answers drop the accent rail (`│ ` gutter) and the green `●` glyph → plain response text + a faint completion line; **errors keep the red `●`** so failures still stand out. Good call retaining the error marker.
- Copy status now reuses the composer spacer line instead of adding one — and the test asserts **footer and view height are unchanged** with vs without the status, which is the right invariant to lock down.
- `assistantMeasure` widening (`width-2` → `width`) and dropping the `width-2` table measure are applied consistently across `renderAssistantRow` and the selectable path, and `transcript_selection` correctly resets `textStart` to 0 now that there's no prefix (selection coordinates stay aligned).
- `titleBar` gains a middle-truncated branch fallback + a cwd-without-branch tier, with a test that asserts the line stays within width on a long branch+model — solid overflow handling.

Tests are updated to match (renames reflect the new behavior) and the new height-invariance + title-overflow tests are good regression guards. `finalRail` is fully removed from the theme too — no dangling refs. LGTM.

### Non-blocking
- Removing the green `●` success glyph (text-only completion) is a deliberate UX choice rather than a bug — just confirming it's the intended look, since it's a visible change to every completed turn.

PR #208, opened 2026-06-15
chore: clear staticcheck nits surfaced by the v16 audit

gnanam1990:
## Summary

Clears the six minor `staticcheck` findings the v16 audit surfaced (§5.1 / §7 #4). All are cosmetic — **no behavior change** — and each is independently verified by `staticcheck`.

| Code | File | Fix |
|---|---|---|
| S1011 ×2 | `agenteval/score.go`, `cli/agent_eval.go` | `for … { append(x, e) }` → `append(x, src...)` |
| S1016 | `tools/web_search.go` | `searchResult{Title:…, URL:…, Snippet:…}` → `searchResult(item)` (identical fields) |
| ST1005 | `tools/web_fetch.go` | lowercase the `"unsafe redirect URL"` error string (+ test assertion) |
| SA4006 | `tui/provider_wizard_test.go` | drop the dead final `next =` assignment (keep the type assertion) |
| S1009 | `provideronboarding/localruntime_test.go` | remove redundant `== nil` before `len()` |

## Verification

- `staticcheck ./...`: **32 → 26** (the remaining 26 are the two separate audit buckets: 15 `U1000` unused + 11 `SA1019` deprecated Bubble Tea mouse API — each its own follow-up).
- `gofmt -l` clean · `go vet` clean · `go build ./...` · **`go test ./...` (full) pass**.
- `deadcode` unchanged (100) · `govulncheck` 0.

Scoped deliberately to the cosmetic nits; the dead-code prune, the deprecated mouse-API migration, the unwired Slack sink, and the scoped-network wiring are tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized error message formatting for consistency across the application.

* **Refactor**
  * Streamlined internal code patterns and structures for improved maintainability.

* **Tests**
  * Updated test assertions to reflect formatting improvements.

PR #209, opened 2026-06-15
web_search: add domains allowlist + score field

Vasanthdev2004:
Enriches the existing web_search tool (merged in #175) with two model-facing additions: a domains allowlist (the zero-only prompt-injection defense) and provider score surfacing. See commit body for details.

## Summary by CodeRabbit

* **New Features**
  * Web search now supports an optional domain allowlist to restrict results to specific hosts.
  * Search results can now include and display provider relevance scores when available.

* **Bug Fixes**
  * Improved robustness when normalizing and matching domain allowlist entries and handling malformed result URLs.
  * Added clearer error handling when the domain allowlist eliminates all results or is provided in an invalid format.

* **Tests**
  * Expanded automated coverage for domain allowlisting, score parsing/rendering, host normalization, and allowlist/port handling.

### DIFF
--- a/internal/tools/web_search.go
+++ b/internal/tools/web_search.go
@@ -29,6 +29,9 @@ type searchResult struct {
 	Title   string
 	URL     string
 	Snippet string
+	// Score is an optional provider-supplied relevance signal in the [0, 1]
+	// range. Zero (the zero value) is treated as "absent" by the renderer.
+	Score float64
 }
 
 // searchBackend discovers URLs for a query. It is an interface so any hosted
@@ -66,6 +69,11 @@ func newWebSearchToolWithBackend(backend searchBackend) Tool {
 						Default:     defaultWebSearchLimit,
 						Minimum:     intPtr(1),
 						Maximum:     intPtr(maxWebSearchLimit),
+					},
+					"domains": {
+						Type:        "array",
+						Description: "Optional list of allowed hostnames; results outside these hosts are filtered out before they reach the model. Useful as a prompt-injection defense when you only want results from a known set of domains.",
+						Items:       &PropertySchema{Type: "string"},
 					},
 				},
 				Required:             []string{"query"},
@@ -123,6 +131,10 @@ func (tool webSearchTool) Run(ctx context.Context, args map[string]any) Result {
 	if limit > maxWebSearchLimit {
 		limit = maxWebSearchLimit
 	}
+	domains, err := stringListArgWebSearch(args, "domains")
+	if err != nil {
+		return errorResult("Error: Invalid arguments for web_search: " + err.Error())
+	}
 
 	if tool.backend == nil {
 		return errorResult("Error: no search backend configured. Set ZERO_WEBSEARCH_BASE_URL (and ZERO_WEBSEARCH_API_KEY) to enable web_search.")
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
Fixed all three issues in 037317f:

**Issue 1 (port stripping):** `canonicalizeWebSearchHost` now uses `parsed.Hostname()` instead of `parsed.Host`, so allowlist entries like `https://react.dev:443/path` normalize to `react.dev` and match results on the default port. Test: `TestCanonicalizeWebSearchHostStripsPort`.

**Issue 2 (fail-closed on all-invalid allowlist):** `stringListArgWebSearch` now returns a `provided` bool so `Run` can distinguish `no allowlist` from `allowlist that normalized to empty`. The latter now errors with `'domains' argument was provided but contained no valid hostnames` before the backend is called, preserving the prompt-injection defense. Test: `TestWebSearchDomainsFilterRejectsAllInvalidInputs` (4 sub-cases: all-whitespace, all-empty, embedded-space, mixed-invalid). All assert the backend is NOT called.

**Issue 3 (host:port regression test):** `TestWebSearchDomainsFilterAcceptsHostPort` covers the scheme'd case (https://react.dev:443/path -> react.dev) AND the scheme-less case (react.dev:9999 is rejected as not-a-valid-hostname). The earlier 9999 case I had was wrong — scheme-less `host:port` is not a valid hostname and the tool correctly rejects it.

All 23 web_search tests pass. `go vet`, `gofmt`, `go build ./...` clean. Full `internal/tools` suite green. Same pre-existing `TestRunExecUsesProjectConfigAndOpenAICompatibleProvider` failure as before (verified unrelated on clean main).

Re-running the review, please.

gnanam1990 (approved):
## Review — web_search domains allowlist + score

Nice, security-minded addition. The `domains` allowlist is a genuine prompt-injection mitigation and the **fail-closed** handling is exactly right:
- domains provided but all entries normalize away → hard error (not a silent unconstrained search);
- allowlist eats every result → a clear "no results matched domains" error instead of a misleading "no results";
- unparseable result URLs are dropped when an allowlist is set.

Host canonicalization (scheme/port/path/`www.` stripping, lowercase, dedup, `[]any` **and** `[]string` tolerance) is thorough, and the tests cover the meaningful paths. `score` is cleanly optional (zero = absent). LGTM.

### Non-blocking notes
1. **Exact-host vs subdomain semantics.** `filterWebSearchByDomains` matches the exact host (after `www.` strip), so `domains:["react.dev"]` keeps `react.dev`/`www.react.dev` but **not** `blog.react.dev`. For a security allowlist that's the safer default — worth a one-line doc note in the param description so callers aren't surprised (or consider opt-in suffix matching later).
2. **Merge overlap with #208.** Both PRs edit the `convert` closure in `parseSearchResults`. #208 (staticcheck nits) rewrote that struct literal as `searchResult(item)` (S1016); this PR rewrites it as an explicit literal with `Score`. Whichever lands second needs a trivial conflict fixup — since `rawResult` and `searchResult` will share field order incl. `Score`, `searchResult(item)` still compiles, so collapsing to that post-merge is cleanest. Just flagging so it isn't a surprise.

Gate-wise the change is stdlib-only and looks consistent with the existing tool conventions.

PR #211, opened 2026-06-15
feat(daemon): TLS remote bridge — authenticated network access to the daemon

gnanam1990:
## Summary

Adds **`internal/daemon/remote`** — an **opt-in, TLS-only network bridge** on top of the local daemon (#203). A remote client drives the **same** `SessionManager`/`Pool` over the network behind bearer-token auth, reusing the daemon's framing + dispatch unchanged. The local Unix-socket daemon stays the default and is unchanged; nothing here activates unless `zero daemon serve-remote` is run.

This is the final completion-kit feature (sandbox → daemon → sandbox-completion → swarm → **daemon-remote**).

## What's included

- **Bridge** — per accepted TLS connection: authenticate, *then* hand to the daemon's control dispatch via a new `Server.ServeConn` seam. A remote session runs under the **same sandbox/risk model** as a local one — remote never bypasses local controls.
- **Authenticator + TokenAuthenticator** — constant-time bearer-token check; token from `$ZERO_DAEMON_REMOTE_TOKEN` or `$ZERO_DAEMON_REMOTE_TOKEN_FILE`; optional attestation hook (no-op default). Auth happens **before any control frame**; failure closes the connection after a small backoff; tokens are never logged.
- **RemoteClient (`DialRemote`)** — verified TLS dial (**never `InsecureSkipVerify`**) + auth handshake, then the existing `daemon.Client` via a new `NewClientConn` seam, so remote and local share one protocol.
- **CLI** — `zero daemon serve-remote --addr --tls-cert --tls-key`; `run`/`attach` gain `--remote <addr> [--token --ca-cert --server-name]`.

## Security (fail closed)

TLS mandatory (refuses to start without a cert/key) · auth-before-frames · constant-time token compare, never logged · configurable **min-protocol-version** floor · the daemon's **1 MiB frame cap** reused · **bounded concurrent connections** (overflow refused) · auth-fail backoff · server cert always verified.

## Daemon seams (additive, behavior unchanged)

`Server.ServeConn(conn)` and `NewClientConn(conn)` — tiny exported wrappers reused by the bridge/client. The local daemon path is otherwise untouched.

## Deferred (documented follow-ups)

`SessionLink` (POST session events to a configured endpoint with bounded retry) and git-bundle repo transfer — both optional per the kit; not needed for the core bridge.

## Verification

**Unit/integration tests** cover: auth accept/reject (env + file token), TLS-required, **cert verification** (untrusted self-signed rejected), version floor, run/status/attach round-trip, **re-attach resumption**, **connection cap**, `ListenAndServeTLS`/`Close`.

Gates: gofmt · vet · build (host + linux + **windows**) · `go test ./...` incl. `-race` · staticcheck (no new) · govulncheck (0) · deadcode (no new).

**Local run check (built binary):**
- `daemon serve-remote` binds TLS + the local socket;
- fail-closed paths verified live: no token / no cert / no `--addr` all rejected;
- over **real TLS-over-TCP**: wrong token → `unauthorized`; untrusted self-signed (no `--ca-cert`) → `failed to verify certificate`; **right token + `--ca-cert` → authenticated, dispatched, and a worker streamed `stream-json` back over the bridge** (only the expected "no provider configured" surfaced — downstream of this change).

## Summary by CodeRabbit

* **New Features**
  * Added `serve-remote` subcommand to start an opt-in secure remote control bridge for the daemon.
  * Added remote connectivity flags (`--remote`, `--token`, `--ca-cert`, `--server-name`) to `daemon run` and `daemon attach`, using TLS (TLS 1.2+) plus bearer-token authentication.
* **Bug Fixes**
  * Improved and standardized error messaging for remote vs local connection failures.
* **Tests**
  * Added coverage for remote auth, TLS/config validation, protocol version gating, and remote connection/session limits.
* **Chores**
  * Reduced daemon mailbox lock contention backoff interval; adjusted concurrent send test timeout.

### DIFF
--- a/internal/cli/daemon.go
+++ b/internal/cli/daemon.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/Gitlawb/zero/internal/background"
 	"github.com/Gitlawb/zero/internal/daemon"
+	"github.com/Gitlawb/zero/internal/daemon/remote"
 )
 
 // runDaemon dispatches the `zero daemon ...` subcommands. The daemon supervises a
@@ -35,6 +36,8 @@ func runDaemon(args []string, stdout io.Writer, stderr io.Writer, _ appDeps) int
 		return runDaemonRun(rest, stdout, stderr)
 	case "attach":
 		return runDaemonAttach(rest, stdout, stderr)
+	case "serve-remote":
+		return runDaemonServeRemote(rest, stdout, stderr)
 	case "-h", "--help", "help":
 		return writeDaemonUsage(stdout, exitSuccess)
 	default:
@@ -53,6 +56,14 @@ Commands:
   run --session <id> [--cwd <dir>] [--prompt <text>] [exec flags...]
                             Create/route a session and stream its output.
   attach <session>          Attach to a running session's stream.
+  serve-remote --addr <host:port> --tls-cert <f> --tls-key <f>
+                            Serve an opt-in, TLS-only network bridge to this
+                            daemon. Requires a bearer token in $ZERO_DAEMON_REMOTE_TOKEN
+                            (or $ZERO_DAEMON_REMOTE_TOKEN_FILE).
+
+run and attach accept --remote <host:port> [--token <t>] [--ca-cert <f>]
+[--server-name <name>] to drive a remote daemon over the bridge instead of the
+local socket.
 `)
 	return code
 }
@@ -212,6 +223,7 @@ func runDaemonStatus(args []string, stdout io.Writer, stderr io.Writer) int {
 
 func runDaemonRun(args []string, stdout io.Writer, stderr io.Writer) int {
 	session, cwd, prompt := "", "", ""
+	var remoteFlags remoteDialFlags
 	var forward []string
 	for i := 0; i < len(args); i++ {
 		a := args[i]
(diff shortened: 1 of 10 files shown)


Vasanthdev2004 (approved):
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr211` at `1076180`. 1179 additions, 11 files, base `main` ← `feat/daemon-remote`. 1 new package (`internal/daemon/remote`) plus additive seams on the daemon server/client.

**CI gates (clean):** `gofmt -l .` empty, `go vet ./...` clean, `go build ./...` clean, `go test -count=1 ./internal/daemon/...` (0.6s + 0.7s), targeted `TestMailboxConcurrentSends` re-run (9.6s, Windows de-flake). The verify-by-handshake and version-floor tests were run verbose — all pass.

- **TLS mandatory, fail-closed.** `ListenAndServeTLS` refuses `nil` config with an explicit error; `ServerTLSConfig` refuses an empty cert or key. `TestListenAndServeTLSRequiresConfig` and `TestListenAndServeTLSServesAndCloses` cover both paths. `MinVersion: tls.VersionTLS12` on both client and server.
- **Auth before frames.** `Bridge.handle` reads the auth frame, enforces the version floor, runs `Authenticator.Authenticate`, optionally runs `Attestation.Verify`, *then* writes the auth-OK and clears the handshake deadline. A failed handshake never dispatches a control frame. The deadline is cleared on success so a session can stream for a long time without truncation.
- **Constant-time token compare.** `TokenAuthenticator.Authenticate` uses `subtle.ConstantTimeCompare`. Tokens are never logged — no `log`/`Print` ever interpolates `req.Token`. The `deny` helper uses a generic reason ("unauthorized", "unsupported protocol version", "attestation failed") so error responses can't be oracle-amplified.
- **Brute-force backoff.** `AuthFailDelay` defaults to 250 ms; tests that need to be fast pass `AuthFailDelay: -1`. The semaphore cap refuses overflow connections immediately rather than queueing them (no resource-exhaustion DOS).
- **Bounded concurrency.** `sem` is sized by `MaxConnections` (default 32). `TestBridgeConnectionCap` exercises the overflow path. The slot is released by a `defer <-b.sem` so a panic during dispatch can't strand it.
- **Client side: cert verification is unconditional.** `clientTLSConfig` never sets `InsecureSkipVerify`. With no `CACertFile` it falls back to system roots, and `TestBridgeRequiresVerifiableCert` proves an untrusted self-signed cert is rejected. `ServerName` is derived from the address (or override) and flows into SNI.
- **Handshake bound.** The client sets a deadline on the conn for `cfg.Timeout` (default 15s) and clears it on success. `Bridge.handle` similarly bounds the auth read. The daemon's existing 1 MiB frame cap is reused via `daemon.ReadFrame` / `daemon.WriteFrame`.
- **Seams are genuinely additive.** `Server.ServeConn` is a 1-line wrapper around the existing `handleConn`. `Client.NewClientConn` is a 7-line variant of `Dial` that hands off a pre-built `net.Conn`. The local Unix-socket daemon path is untouched — `Serve()` still binds the owner-only socket, single-instance lock, status file. A remote session goes through the same `SessionManager`/`Pool`, so it inherits the same sandbox/risk model as a local one.

**One minor design observation (not a blocker):**

`daemonDialError` is a small UX win — it tells the user "failed to reach remote daemon at X" vs "daemon is not running" depending on whether `--remote` was set. Nice touch.

`runDaemonServeRemote` re-builds the local `Pool`/`SessionManager`/`Server` in-process before binding both the local socket and the TLS bridge. That's correct for the "serve-remote from scratch" command, but the code path duplicates the wiring that `daemon start` already does internally. Long-term, factor the wiring into a `startLocalServer(...)` helper and call it from both. Cosmetic.

**Overall:** this is the right shape — opt-in, fail-closed, additive seams, real TLS, no plaintext fallback, no cert-skip, no token logging. The PR body documents the deferred `SessionLink` and git-bundle transfer as optional follow-ups, which is honest. Merging.

gnanam1990:
@coderabbit approvable?

gnanam1990:
@coderabbitai review

Both items fixed in e1be327:

- 🔴 **TLS dial (noctx)** — replaced the deprecated `tls.DialWithDialer` with a context-aware `tls.Dialer.DialContext` (context bounds the dial + TLS handshake; the returned conn is independent of it for the auth/stream I/O). Resolves the golangci-lint `noctx` blocker.
- 🟡 **Close idempotency** — `Close()` now clears `b.listener` under the lock, so a repeat call is a no-op instead of a closed-listener error (matches the doc). Test now asserts a double `Close` succeeds.

Gates: gofmt · vet · build (host+linux+windows) · test -race · staticcheck (no new) · govulncheck (0) · deadcode (no new) — all green.

PR #212, opened 2026-06-15
mcp(oauth): delegate the OAuth engine to internal/oauth (dedup)

gnanam1990:
## Summary

**OAuth Phase 2** — MCP OAuth now reuses the shared **`internal/oauth`** engine (from #210) instead of carrying its own copy of the transport/identity-agnostic primitives. `internal/mcp/oauth.go` drops **526 → 418 LOC** (−187/+79).

> **Stacks on #210** — base is `feat/oauth-module`. Merge #210 first; this then targets `main` cleanly.

## What now delegates to `internal/oauth`
`newPKCE`, `newState`, `discoverAuthorizationServer` (RFC 8414), the authorize-URL builder, `exchangeCode`, `refreshAccessToken`, the token-response parser, and `joinWellKnown` — all removed from MCP and delegated to the shared engine via thin type conversions.

## Behavior-preserving
MCP keeps its own `OAuthConfig`/`StoredToken` types, `LoginOptions`/`Login` orchestration, loopback handling, `registerClient` (dynamic client registration — not part of the shared engine), CLI output, and **on-disk token format** (`mcp-oauth-tokens.json`, raw server-name keys). **The existing `internal/mcp/oauth_test.go` + `oauth_store_test.go` + CLI tests pass unchanged.**

**One intentional hardening** inherited from the shared engine: the token endpoint must be https (loopback exempt), so a credential is no longer sent to a plaintext non-loopback endpoint. All existing tests use loopback and are unaffected.

## Not changed
The token **store** (`internal/mcp/oauth_store.go`) is untouched — unifying it with `internal/oauth`'s namespaced store would change the on-disk format and is a separate, transparent-migration follow-up.

## Verification
gofmt · vet · build (host + linux + **windows**) · `go test ./...` incl. `-race` (mcp tests **unchanged**) · staticcheck (no new) · govulncheck (0) · deadcode (no new vs branch base). Local run check: `zero mcp oauth --help` / `status` / `status --json` produce identical output.

## Summary by CodeRabbit

* **Refactor**
  * Internal OAuth and PKCE authentication handling has been refactored to use shared components across the system. This consolidates duplicate logic and improves code maintainability. All existing user-facing authentication behavior and APIs remain unchanged with no impact on functionality.

### DIFF
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -2,9 +2,6 @@ package mcp
 
 import (
 	"context"
-	"crypto/rand"
-	"crypto/sha256"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Vasanthdev2004 (approved):
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr212` at `1a9591f` (1 commit on top of `feat/oauth-module`). +79 / −187, single file (`internal/mcp/oauth.go`), base `feat/oauth-module` ← `feat/oauth-mcp-dedup`. **Merge order:** #210 first, then this rebases cleanly onto `main`.

**CI gates (clean):** `gofmt -l .` empty, `go vet ./...` clean, `go build ./...` clean, `go test -count=1 ./internal/mcp/...` (1.7s), `go test -count=1 ./internal/oauth/...` (7.1s).

**What got delegated to the shared engine:**

- `newPKCE` → `oauth.NewPKCE()` (32 bytes RawURLEncoding, S256 challenge — byte-for-byte identical)
- `newState` → `oauth.NewState()` (32 bytes, base64url)
- `discoverAuthorizationServer` → `oauth.DiscoverAuthorizationServer()` (RFC 8414 well-known, with type conversion back to MCP's `authServerMetadata` since the MCP metadata type omits the device/registration-extension fields the shared engine surfaces)
- `authorizationURL` → `oauth.BuildAuthorizationURL()` (now also protected by the reserved-param guard from the shared engine)
- `exchangeCode` → `oauth.ExchangeCode()` (form-encoded token POST with the right grant_type and verifier)
- `refreshAccessToken` → `oauth.Refresh()` (refresh-token grant, preserves prior refresh token if the response omits one, maps `error` / `error_description`)
- `joinWellKnown`, the local `tokenResponse` struct, and `postTokenRequest` — all removed from MCP, replaced by calls into the shared engine

**What MCP keeps (correctly):**

- Its own `OAuthConfig` / `StoredToken` types (different from `oauth.Config` / `oauth.Token` — the `ServerName` and other MCP-specific fields stay local)
- `LoginOptions` and the full `Login` orchestration (one deadline bounds discovery + optional client registration + callback wait + token exchange, the loopback listener is bound first so the redirect URI is known before client registration)
- The inline `http.Server` loopback handler (MCP serves `/callback` on its own; not using the shared `LoopbackListener` because the callback flow is different)
- `registerClient` (RFC 7591 dynamic client registration — explicitly noted as not part of the shared engine)
- The CLI surface
- The on-disk token format (`mcp-oauth-tokens.json` with raw server-name keys, no shared-engine namespacing)

**Conversion helpers are package-private and used only inside the file** — `pkceToParams`, `pkceToOAuth`, `configFor`, `tokenToStored`. No external callers depended on the removed names (grepped the whole module).

**Inherited hardening (intentional):** the token endpoint must be https or loopback. MCP tests all use loopback, so the behavior is unchanged, but the new path will refuse a credential-leaking plaintext token endpoint that the old path would have silently sent credentials to.

**One minor observation (not a blocker):**

The MCP loopback in `Login` (line 320) is `net.Listen("tcp", "127.0.0.1:0")` + an inline `http.Server` — this is the MCP-specific redirect path and stays put. It binds 127.0.0.1 only (correct) and is single-use by the `resultChan` handshake, but it doesn't have the same single-use / idempotent-Close guarantees as the shared `LoopbackListener`. That's OK because the `server.Shutdown` is always called in `defer`, but if a future change ever wants the same security-property guarantee the OAuth provider login has, switching to the shared listener would do it. Cosmetic.

**Overall:** this is exactly the right shape for "delegate the transport-agnostic core, keep the identity-specific surface." No behavior change for users, no token-store migration (deferred to a separate, transparent PR), the MCP test suite is untouched and still passes, and MCP inherits the engine's hardening. Merging once #210 lands.

PR #213, opened 2026-06-15
oauth: opt-in AES-256-GCM encrypted-at-rest token storage

gnanam1990:
## Summary

**OAuth Phase 2** — adds an opt-in **encrypted-at-rest** token store. The default backend is unchanged (0600 plaintext JSON); setting `ZERO_OAUTH_STORAGE=encrypted-file` encrypts the token file with **AES-256-GCM** under a per-user random 32-byte secret persisted 0600 beside it (`<tokens>.secret`). Pure stdlib — no new dependency.

Builds on the merged `internal/oauth` engine (#210); branches off `main`.

## How it works
- `internal/oauth/encrypt.go` — `aesGCMCrypter`: seal = random-nonce ‖ ciphertext; GCM gives confidentiality **and** tamper detection. Secret is load-or-create with an atomic 0600 write.
- `Store` gains an optional crypter: `readState` decrypts before unmarshal, `writeState` encrypts after marshal. A **missing secret, short blob, or failed auth tag all fail closed** — never a silent empty/plaintext fallback.
- CLI: `newAuthManager` selects the backend from `ZERO_OAUTH_STORAGE`; `zero auth` help documents it.

## Security
Confidentiality + integrity at rest (AES-256-GCM), per-user random secret, 0600 on both files, fail-closed on tamper/missing-secret. Token material is still never logged (status output unchanged).

## Verification
Tests: crypter round-trip + tamper; secret create/persist/0600/wrong-size fail-closed; store-through-encrypted (ciphertext on disk, round-trip, `Status`/`Delete`, tamper + missing-secret fail-closed).

Gates: gofmt · vet · build (host + linux + **windows**) · `go test ./...` incl. `-race` · staticcheck (no new) · govulncheck (0) · deadcode (no new vs main).

**Local run check (built binary):** a `--device` login under `ZERO_OAUTH_STORAGE=encrypted-file` against a local fake OAuth server:
- `Logged in to demo.`
- on-disk token file is **ciphertext** — 0 plaintext markers (no access token / no `schemaVersion`), binary hexdump;
- `<tokens>.secret` present at mode **600**;
- `zero auth status` decrypts and reads it back (`demo: logged in (refreshable), expires …`), leaking no secret material.

## Follow-ups (unchanged from before)
OS-keyring backend (needs a third-party dep) and unifying the MCP token store onto `internal/oauth`'s namespaced store (transparent migration) remain separate follow-ups.

## Summary by CodeRabbit

* **New Features**
  * Added optional encrypted-at-rest OAuth token storage via `ZERO_OAUTH_STORAGE=encrypted-file` (AES-256-GCM), storing tokens on disk as ciphertext with a companion secret file.

* **Documentation**
  * Expanded auth help with a new “Storage” section covering default token location, `ZERO_OAUTH_TOKENS_PATH` override, `0600` permissions, and encryption behavior.

* **Tests**
  * Added coverage for encryption/decryption correctness, ciphertext not exposing plaintext, tamper/missing-secret failure-closed behavior, concurrent secret creation convergence, and CLI rejection of invalid `ZERO_OAUTH_STORAGE` values.

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -129,9 +129,19 @@ func validateAuthFlags(sub string, a authArgs) error {
 // ZERO_OAUTH_STORAGE=encrypted-file selects the AES-256-GCM encrypted-at-rest
 // backend (a per-user secret is created beside the token file).
 func newAuthManager(deps appDeps, out io.Writer) (*oauth.Manager, error) {
+	// Validate ZERO_OAUTH_STORAGE up front: a mistyped non-empty value must fail
+	// fast rather than silently fall back to plaintext while the user believes
+	// encryption is on. Empty = default (plaintext 0600); "encrypted-file" = AES.
+	encrypted := false
+	if mode := strings.TrimSpace(os.Getenv("ZERO_OAUTH_STORAGE")); mode != "" {
+		if !strings.EqualFold(mode, "encrypted-file") {
+			return nil, fmt.Errorf("invalid ZERO_OAUTH_STORAGE %q (supported: encrypted-file)", mode)
+		}
+		encrypted = true
+	}
 	store, err := oauth.NewStore(oauth.StoreOptions{
 		Now:       deps.now,
-		Encrypted: strings.EqualFold(strings.TrimSpace(os.Getenv("ZERO_OAUTH_STORAGE")), "encrypted-file"),
+		Encrypted: encrypted,
 	})
 	if err != nil {
 		return nil, err
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
@@ -124,9 +125,14 @@ func validateAuthFlags(sub string, a authArgs) error {
 
 // newAuthManager builds an oauth.Manager backed by the file store, printing the
 // authorization URL / device code to stdout. The store path honors
-// ZERO_OAUTH_TOKENS_PATH (env), so callers/tests can redirect it.
+// ZERO_OAUTH_TOKENS_PATH (env), so callers/tests can redirect it. Setting
+// ZERO_OAUTH_STORAGE=encrypted-file selects the AES-256-GCM encrypted-at-rest
+// backend (a per-user secret is created beside the token file).
 func newAuthManager(deps appDeps, out io.Writer) (*oauth.Manager, error) {
-	store, err := oauth.NewStore(oauth.StoreOptions{Now: deps.now})
+	store, err := oauth.NewStore(oauth.StoreOptions{
+		Now:       deps.now,
+		Encrypted: strings.EqualFold(strings.TrimSpace(os.Getenv("ZERO_OAUTH_STORAGE")), "encrypted-file"),
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr213` at `563f0a3`. The actual PR diff (1 commit on top of main) is 382 lines across `internal/cli/auth.go` (12 +), `internal/oauth/encrypt.go` (new, 116), `internal/oauth/encrypt_test.go` (new, 145), and `internal/oauth/store.go` (28 +). `gofmt`/`vet`/`build` clean; `go test ./internal/oauth/...` passes (9.2s). Full suite shows only the pre-existing `TestRunExecUsesProjectConfigAndOpenAICompatibleProvider` failure (reproduces on clean main, not this PR).

**What it does:** adds an opt-in `Encrypted: true` flag to `oauth.StoreOptions`. When set, the JSON token file is replaced by an `AES-256-GCM` ciphertext blob (`nonce || ciphertext`), with a separate 32-byte random secret persisted in `<filePath>.secret` (0600). Activated at the CLI by `ZERO_OAUTH_STORAGE=encrypted-file`. Default is unchanged (0600 plaintext JSON).

- **AES-256-GCM** with random nonces via `crypto/rand`. 32-byte secret generated on first use, persisted atomically (`tmp + rename`), 0600 permissions, dir 0700.
- **Fail-closed on missing secret**: `open` passes `create=false`, so a missing secret is a hard error rather than silently minting a new key that could never decrypt the existing file.
- **Tamper detection**: GCM auth tag is verified; a flipped byte breaks `gcm.Open` and surfaces as `"wrong secret or tampered"`.
- **Short-blob guard**: `len(blob) < gcm.NonceSize()` is rejected before attempting decryption.
- **No key material in error messages**: the only error that mentions the secret is `"token secret %s is missing"` (path only, not the secret value).
- **Schema version unchanged**: the unified store's `schemaVersion: 1` is still validated, so a corrupted/foreign file fails closed.

**Tests (encrypt_test.go):** round-trip + tamper + missing-secret + short-blob + wrong-sized secret + atomic-rename + dir-creation, all green.

**One small observation (not a blocker):**

The CLI change wires `ZERO_OAUTH_STORAGE=encrypted-file` to `Encrypted: true` and documents it in `auth --help`. It would be nice to also plumb the **keyring** backend (from #216) through the same env var as a third option — that way the env var becomes a single switch: `plain` (default) | `encrypted-file` | `keyring`. The PR as-is is consistent with the "encrypted file" addition, and #216's keyring is the natural follow-up to make them siblings.

Merging.

gnanam1990:
@coderabbit rereview

gnanam1990:
Pushed `0b040a3` addressing CodeRabbit's review:

- `newAuthManager` rejects an unsupported `ZERO_OAUTH_STORAGE` value instead of silently downgrading to plaintext (fails fast with "supported: encrypted-file").
- `loadOrCreateSecret` fixes the first-run TOCTOU race: `os.Rename` clobbers on POSIX, so concurrent processes could each mint a secret and the loser would orphan its tokens. Now creates the secret exclusively (`O_CREATE|O_EXCL`) and the loser adopts the winner's on-disk secret. Factored read+validate into `readSecretFile`; dropped the now-unused clock plumbing.
- `encrypt_test` handles `os.Stat` / `os.ReadFile` errors before dereferencing (no panic masking the real failure).

Added a test that an invalid `ZERO_OAUTH_STORAGE` fails fast. gofmt / vet / build / `go test -race` green.

@coderabbitai review

PR #215, opened 2026-06-15
notify: wire the webhook sink behind an opt-in env var

gnanam1990:
## Summary

The `WebhookSink` (best-effort Slack / generic JSON POST when a turn completes or awaits input) was fully built, redaction-safe, and unit-tested — but **never attached to a `Notifier`**. `deadcode` therefore flagged its whole machinery as unreachable, and the v16 audit listed it as "wire or remove `notify.WebhookSink`". This **wires** it.

## What changed
- New helper `notify.MaybeAddWebhookSink(n, env, logf)`:
  - reads **`ZERO_NOTIFY_WEBHOOK_URL`** (and optional **`ZERO_NOTIFY_WEBHOOK_SUMMARY`**)
  - **no-op** when the URL is unset/blank → safe to call unconditionally
  - the URL (which typically embeds a secret token) comes from the **environment**, so it never has to be written into an on-disk config file
- Attached at both notifier construction sites:
  - **headless `exec`** — failed deliveries log to **stderr** (never stdout, which may carry stream-json); the sink redacts before logging
  - **TUI** — `logf=nil`, because the TUI owns the alt-screen and a stderr write would corrupt the display

## Behavior / safety
- **Opt-in & fail-closed:** with `ZERO_NOTIFY_WEBHOOK_URL` unset, nothing is attached and behavior is identical to before.
- The sink remains gated by the **existing Mode/focus policy** (`TestNotifierOffSuppressesSinks` proves `Mode==off` suppresses sinks), so a webhook only delivers when notifications are enabled (e.g. `--notify both`). No change to default notification behavior.
- Secret hygiene unchanged: the sink already redacts URL + message + links before any POST/log.

## Verification
- **deadcode: 100 → 93** unreachable funcs — the 7 `WebhookSink` entries (`NewWebhookSink`, `Emit`, `text`, `redactLinks`, `log`, `eventType`, `readSnippet`) become reachable; **no new dead code**.
- New tests: attaches + delivers (httptest, asserts summary propagation), no-op on blank/whitespace URL, nil-guard (nil notifier / nil env).
- gofmt · vet · build (host + linux + **windows**) · full `go test ./...` (`-race` on notify) · staticcheck (no new) · govulncheck (0) all green.

## Release Notes

* **New Features**
  * Run notifications can now be delivered to an external webhook via the `ZERO_NOTIFY_WEBHOOK_URL` environment variable for integration with external systems
  * Optional `ZERO_NOTIFY_WEBHOOK_SUMMARY` environment variable allows customizing the run summary in webhook payloads
  * Webhook delivery errors are logged without affecting application output

### DIFF
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -367,6 +367,12 @@ func runExec(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 		Mode:      notify.Mode(strings.TrimSpace(execNotifyMode(options, resolved))),
 		FocusMode: notify.FocusAlways,
 	})
+	// Opt-in webhook fan-out (ZERO_NOTIFY_WEBHOOK_URL). Headless runs can safely
+	// log a failed delivery to stderr (never stdout). The sink redacts before
+	// logging, so a token in the URL or message is masked.
+	notify.MaybeAddWebhookSink(notifier, os.Getenv, func(format string, args ...any) {
+		fmt.Fprintf(stderr, "[notify] "+format+"\n", args...)
+	})
 	if options.useSpec {
 		return runExecSpecDraft(execSpecDraftRun{
 			options:            options,
--- a/internal/notify/webhook_wire.go
+++ b/internal/notify/webhook_wire.go
@@ -1,0 +1,40 @@
+package notify
+
+import "strings"
+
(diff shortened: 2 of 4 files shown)


gnanam1990:
Consolidating the audit-backlog work into a single PR with one commit per change (per request). This change is preserved verbatim as a commit there.

PR #217, opened 2026-06-15
OAuth / account login for providers (OpenRouter, xAI) + subscription-proxy support

gnanam1990:
## Summary

Adds real OAuth / account login for the providers that support it, surfaced as a first-class **"Sign in with OAuth"** option in both the `/provider` wizard and first-run onboarding — no API key to paste. Two providers qualify today:

- **OpenRouter** — public PKCE flow that **mints an API key** (no client_id needed).
- **xAI (Grok)** — public OAuth client; the token is used directly as a bearer on `api.x.ai/v1`. Supports **browser** and **device-code** login.

For OpenAI/Anthropic *subscriptions* (ChatGPT/Claude), which can't be used as a standard bearer by a third-party tool, this ships a documented **local-proxy** path instead of an unreliable/ToS-risky direct login.

## What's included

**Runtime auth wiring**
- `providerio.SendWithAuthRetry` + `TokenResolver`: the OpenAI/Anthropic providers send `Authorization: Bearer <oauth>` when a login exists for the profile (auto-refreshed; one refresh-and-retry on `401`), else fall back to the API key. Never both. Zero per-request overhead for API-key users (resolver attached only when a login exists).

**OAuth engine & logins**
- `internal/oauth`: PKCE (S256), loopback browser flow, RFC 8628 device-code flow, token store (0600 / optional OS keyring), env-configurable providers (`ZERO_OAUTH_<NAME>_*`) + a built-in preset layer (xAI).
- `internal/provideroauth.OpenRouterLogin` (mints a key) and `zero auth openrouter`; `zero auth login xai [--device]`.

**In-app UX**
- A **"How do you want to connect?"** chooser (wizard + onboarding): *Sign in with OAuth* lists the OAuth-capable providers → browser/device login → straight to model selection. Browse/API-key path unchanged.
- **Device code** for headless/SSH: press `d` (or automatic on SSH / headless Linux / `ZERO_OAUTH_DEVICE=1`) — shows a URL + code to enter on another device.
- After login, model discovery is authenticated with the OAuth token, so the user picks from the **live** model list (curated fallback if offline).

**Subscription proxies (ChatGPT / Claude)**
- `chatgpt-proxy` preset + `docs/oauth-subscriptions.md`: full recipe for pointing Zero at a local proxy that holds the subscription session. ChatGPT/Claude are intentionally **not** in the OAuth chooser (they can't do clean in-app OAuth); they live under the browse path + proxy docs.

## Deliberately not included
Direct ChatGPT/Claude subscription login. Verified: the ChatGPT token only works against a Cloudflare-gated backend (403 for non-official clients), and Anthropic's terms prohibit third-party use of a Claude subscription token (enforced with account bans). Spoofing those official clients is fragile and account-risky, so it's not built — the proxy path is the supported alternative.

## Testing
- gofmt, `go vet`, build (host + linux + windows), `go test ./... -race` — all green.
- staticcheck (no new findings), govulncheck (0), deadcode (no new) — clean.
- Unit tests cover the bearer/retry logic, OAuth engine (loopback + two-phase device), the wizard/onboarding chooser + device shortcut, token-authenticated discovery, and catalog classification. OAuth/proxy flows verified against mock servers.
- No new module dependencies.

## Notes
The interactive browser/device approval and the live xAI `/models` response require a real account to confirm end-to-end; all surrounding logic is unit-tested and the OpenRouter mint flow was run-checked against a mock.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added OAuth login across CLI and TUI, including `auth openrouter` (browser-based PKCE that prints an API key) and RFC 8628 device-code login where supported.
  * Introduced a new “connect method” step in onboarding/provider setup (including device-code shortcuts, OAuth badges, and device-code/browser waiting states).
  * Added per-request OAuth bearer auth for OpenAI- and Anthropic-compatible providers with automatic refresh-on-401 retry, plus optional built-in OAuth presets (opt-in for xAI).

* **Documentation**
  * Expanded the OAuth subscriptions guide with local proxy configuration examples.

* **Bug Fixes**
  * Improved OAuth-pending interaction handling (mouse input disabled) and avoided unnecessary onboarding by recognizing stored OAuth logins.

### DIFF
--- a/docs/oauth-subscriptions.md
+++ b/docs/oauth-subscriptions.md
@@ -41,26 +41,24 @@ Running `/provider` opens a **"How do you want to connect?"** chooser:
   Paste an API key / browse providers  Any of 20+ providers, local, or a proxy
 ```
 
-Pick **Sign in with OAuth** → a list of sign-in options → choose one:
+Pick **Sign in with OAuth** → the list of providers that do real OAuth → choose one:
 
 ```
 ❯ OpenRouter    browser sign-in · creates a key
   xAI (Grok)    browser or device code
-  ChatGPT       subscription · needs a local proxy
-  Claude        subscription · needs a local proxy
 ```
 
 - **OpenRouter / xAI** are real OAuth: your browser opens to approve → done (no
-  key to paste). OpenRouter mints a key; xAI stores a refreshable token.
+  key to paste). OpenRouter mints a key; xAI stores a refreshable token. The same
+  chooser appears in first-run onboarding.
 - **Device code (headless / SSH):** for a provider that supports it (xAI), press
   **d** on the list to get a code to enter on another device instead of opening a
   browser. On an SSH session or headless Linux box (no `DISPLAY`) device code is
   used automatically; set `ZERO_OAUTH_DEVICE=1` to force it anywhere. The CLI
   equivalent is `zero auth login xai --device`.
-- **ChatGPT / Claude** can't do in-app OAuth (see §2). Selecting them does **not**
-  fake a login — it drops you into the local-proxy setup (endpoint → model) with
-  the proxy defaults pre-filled, so you point Zero at a proxy you run. The same
-  first-run onboarding chooser offers the identical options.
+- **ChatGPT / Claude are intentionally not in this list** — they can't do in-app
+  OAuth (see §2). Use *Paste an API key / browse providers* + a local proxy
+  (`chatgpt-proxy` / `custom-anthropic-compatible`), as described below.
 
 ### Built-in OAuth providers (no env needed)
 
--- a/docs/oauth-subscriptions.md
+++ b/docs/oauth-subscriptions.md
@@ -50,7 +50,8 @@ Pick **Sign in with OAuth** → the list of providers that do real OAuth → cho
 
 - **OpenRouter / xAI** are real OAuth: your browser opens to approve → done (no
   key to paste). OpenRouter mints a key; xAI stores a refreshable token. The same
-  chooser appears in first-run onboarding.
+  chooser appears in first-run onboarding. (xAI uses an opt-in preset — set
+  `ZERO_OAUTH_ALLOW_PRESETS=1` or your own `ZERO_OAUTH_XAI_*`; see below.)
 - **Device code (headless / SSH):** for a provider that supports it (xAI), press
   **d** on the list to get a code to enter on another device instead of opening a
   browser. On an SSH session or headless Linux box (no `DISPLAY`) device code is
@@ -60,20 +61,23 @@ Pick **Sign in with OAuth** → the list of providers that do real OAuth → cho
   OAuth (see §2). Use *Paste an API key / browse providers* + a local proxy
   (`chatgpt-proxy` / `custom-anthropic-compatible`), as described below.
 
-### Built-in OAuth providers (no env needed)
+### Built-in OAuth providers
 
-Two providers ship a working browser login out of the box:
-
-- **OpenRouter** — `zero auth openrouter` opens a browser, you approve, and it
-  **mints an OpenRouter API key** (public PKCE flow, no client_id). In the
-  interactive setup wizard, pick **OpenRouter** and press **ctrl+o** at the key
-  step to do the same inline ("Log in with OAuth"). The minted key is saved to the
-  provider profile and used normally.
-- **xAI (Grok)** — `zero auth login xai` (browser, or `--device` for headless).
-  The token is used directly on `api.x.ai/v1`; configure an `xai` provider profile
-  and it's picked up automatically. Requires a SuperGrok / X Premium+ subscription;
-  the client_id is an undocumented public Grok-CLI client (override via
-  `ZERO_OAUTH_XAI_*` if it changes).
+- **OpenRouter (no env needed)** — `zero auth openrouter` opens a browser, you
+  approve, and it **mints an OpenRouter API key** (public PKCE flow, no client_id).
+  In the interactive setup wizard, pick **OpenRouter** and press **ctrl+o** at the
+  key step to do the same inline ("Log in with OAuth"). The minted key is saved to
+  the provider profile and used normally.
+- **xAI (Grok) — opt-in preset** — xAI's flow needs an OAuth `client_id`. Zero
+  ships a built-in preset for the public Grok-CLI client, but to keep third-party
+  client identities out of the default credential path it is **off by default**.
+  Enable it with `export ZERO_OAUTH_ALLOW_PRESETS=1`, then `zero auth login xai`
+  (browser, or `--device` for headless) works one-click; the token is used directly
+  on `api.x.ai/v1`. Without the opt-in, set `ZERO_OAUTH_XAI_CLIENT_ID` (and
+  endpoints, or an issuer) yourself via `ZERO_OAUTH_XAI_*`. Either way the preset is
+  fully overridable by `ZERO_OAUTH_XAI_*` (env wins), and it requires a
+  SuperGrok / X Premium+ subscription; the client_id is an undocumented public
+  Grok-CLI client that may change without notice.
 
 Any field of a preset is overridable via `ZERO_OAUTH_<NAME>_*`. For a fully custom
 OAuth/OIDC provider, set those env vars (see `zero auth --help`) and
--- a/docs/oauth-subscriptions.md
+++ b/docs/oauth-subscriptions.md
@@ -32,6 +32,25 @@ When a login exists for a provider, the **OpenAI and Anthropic** providers send
 before. Tokens are stored 0600 (or the OS keyring with
 `ZERO_OAUTH_STORAGE=keyring`) and never logged. See `zero auth --help`.
 
+### Built-in OAuth providers (no env needed)
+
+Two providers ship a working browser login out of the box:
+
+- **OpenRouter** — `zero auth openrouter` opens a browser, you approve, and it
+  **mints an OpenRouter API key** (public PKCE flow, no client_id). In the
+  interactive setup wizard, pick **OpenRouter** and press **ctrl+o** at the key
+  step to do the same inline ("Log in with OAuth"). The minted key is saved to the
+  provider profile and used normally.
+- **xAI (Grok)** — `zero auth login xai` (browser, or `--device` for headless).
+  The token is used directly on `api.x.ai/v1`; configure an `xai` provider profile
+  and it's picked up automatically. Requires a SuperGrok / X Premium+ subscription;
+  the client_id is an undocumented public Grok-CLI client (override via
+  `ZERO_OAUTH_XAI_*` if it changes).
+
+Any field of a preset is overridable via `ZERO_OAUTH_<NAME>_*`. For a fully custom
+OAuth/OIDC provider, set those env vars (see `zero auth --help`) and
+`zero auth login <name>`.
+
 ---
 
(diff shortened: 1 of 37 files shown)


Vasanthdev2004 (changes requested):
### Verdict: CHANGES REQUESTED 🛑

Reviewed locally in worktree `review-pr217` at `d054120`. The PR adds OAuth login for OpenRouter + xAI, a subscription-proxy recipe, and a wizard UI. `gofmt`/`vet`/`build` clean. All relevant test packages pass — `internal/oauth` (9.6s), `internal/provideroauth` (2.3s), `internal/providercatalog` (0.9s), `internal/tui` (4.4s). Full suite shows only the pre-existing `TestRunExecUsesProjectConfigAndOpenAICompatibleProvider` failure (not this PR).

The OpenRouter flow and the docs are clean. The **xAI preset, however, breaks the engine's security invariant and must be removed before this lands.**

## Blocker: hardcoded xAI client_id in `internal/oauth/presets.go`

```go
var builtinOAuthPresets = map[string]providerPreset{
    "xai": {
        ClientID: "b1a00492-073a-47ea-816f-4c329264a828",
        ...
    },
}
```

`internal/oauth/oauth.go` (the package doc) states explicitly:

> *This package holds NO vendor secrets: provider client IDs and any overridable endpoints come from config/env, so no third-party OAuth client identity is baked into the binary.*

This PR violates that invariant. The fact that the value is documented in the preset comment as "public" doesn't change the design intent — the engine was deliberately built to keep third-party OAuth client identities out of the binary. The fact that the preset is "overridable via env" also doesn't help: the default still gets baked in, and any user who runs `zero auth login xai` with no `ZERO_OAUTH_XAI_CLIENT_ID` will silently authenticate as `b1a00492-...`.

The PR body says xAI's flow needs that specific client_id and that it's known to be public, but the engine has a documented rule against this. Either:

1. **Remove the xAI preset** and document that xAI users must set `ZERO_OAUTH_XAI_CLIENT_ID` (and any other required endpoints) themselves — same posture every other provider gets. The wizard's "Sign in with OAuth" chooser for xAI can then prompt for the client_id if it's missing.
2. **Add a runtime opt-in** (e.g. `ZERO_OAUTH_ALLOW_PRESETS=1`) that gates the preset map entirely, so a user explicitly opts in to third-party client identities being baked in.

Option 1 is the cleaner fix and is consistent with the package invariant.

## Other observations (informational, can stay as follow-ups)

- The `validateTokenEndpoint` naming smell I flagged on #210 is re-introduced here (`providers.go` calls it on `cfg.IssuerURL` / `cfg.AuthorizationEndpoint`). Cosmetic; can be a single rename PR across #210/#216/#217.
- The OpenRouter flow (`internal/provideroauth/openrouter.go`) is **exemplary** — no client_id needed (the OpenRouter flow is designed that way), PKCE S256, loopback 127.0.0.1, body cap 1 MiB, no key material in error messages. The two new tests (`openrouter_test.go`) cover the happy path, the missing-code callback, the empty-key response, and the non-2xx exchange.
- The `internal/browser/open.go` opener is a clean cross-platform abstraction (`xdg-open` / `open` / `rundll32 url.dll,FileProtocolHandler`).
- The `internal/providers/providerio/auth.go` helper for OpenAI/Anthropic `Authorization: Bearer` plumbing is small and correct — uses the existing `oauth.Manager.GetFresh` with a 401-recovery path.
- The setup wizard changes are well-scoped: a "How do you want to connect?" chooser, OAuth card and API-key card are mutually exclusive, the device-code path prints the URL and the user_code on separate lines.

## Overall

The architecture is sound, the OpenRouter design is correct, the TUI wizard is tasteful. The **only** reason this isn't an approve is the hardcoded xAI client_id in `presets.go` — that single line breaks the package invariant the OAuth engine was designed to uphold. Once it's removed (option 1 above) or gated behind a runtime opt-in, this is a clean approve.

gnanam1990:
Thanks for the thorough reviews — pushed `aade9e7` + `2b5b0ad` addressing everything.

**Blocker — hardcoded xAI `client_id` (`presets.go`):** made the built-in presets **opt-in** rather than removing them (your option 2). `ResolveConfig` only consults `builtinOAuthPresets` when `ZERO_OAUTH_ALLOW_PRESETS` is truthy; by default every field comes from `ZERO_OAUTH_<NAME>_*` env, so no third-party OAuth client identity is used unless the operator explicitly opts in. The "not configured" error now names the flag, and the package / `Registry` / preset docs + `docs/oauth-subscriptions.md` were updated. Env still overrides any preset field.

**CodeRabbit:**
- `PrepareDeviceLogin` self-bounds the prepare (discovery + device-code request) with `opts.Timeout`, matching `Login`; `CompleteDeviceLogin` bounds the poll by the device code's `ExpiresAt`.
- `OAuthProviders()` returns cloned descriptors (matches `All()`).
- Onboarding ignores a stale OAuth result for a provider the user has switched away from (`applySetupOAuth` + `applySetupOAuthDeviceCode`).
- A failed OAuth login is surfaced on the provider list the user lands on (the waiting screen that owned the error is gone once `oauthPending` clears).
- "Sign in with OAuth" is hidden when a setup's own provider list has no OAuth entry (`setupMethodOptions`).
- docs: labeled the example fences; updated the Anthropic section for the April 4 2026 third-party cutoff.

Added tests for the preset opt-in gate, the descriptor-clone isolation, and the method-option filtering. gofmt / vet / build (host+linux+windows) / `go test -race` are green.

Left the `validateTokenEndpoint` naming as the cross-PR rename follow-up you flagged.

@coderabbitai review

gnanam1990:
@coderabbit rereview

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
Pushed `ecf38f6` with the remaining nitpicks:

- `scopesOrPreset` returns a copy of the preset scopes (no aliasing of the shared global slice).
- `provideroauth/openrouter` uses `bytes.NewReader(payload)` instead of `strings.NewReader(string(payload))`.
- `setupCredentialSummary` shows **"OAuth token"** on the Ready screen for a token-login provider (xAI) instead of `env var XAI_API_KEY`; key-minting OAuth (OpenRouter) still resolves to a saved key.

The flagged "Registry doc says no built-ins" was already corrected in the preset opt-in commit (`2b5b0ad`) — the doc now describes the `ZERO_OAUTH_ALLOW_PRESETS` gate. Added tests for the scope-slice copy and the OAuth-token summary.

That clears the open items across all four rounds. @Vasanthdev2004 — re-requesting your review; the xAI hardcoded client_id you flagged is resolved in `2b5b0ad` (built-in presets are now off by default, opt-in via `ZERO_OAUTH_ALLOW_PRESETS`).

@coderabbitai review

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
@coderabbit approve

Vasanthdev2004:
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr217-fresh` at `ecf38f6`. Latest commits on `feat/oauth-provider-login` address CodeRabbit rounds 2-4 nitpicks, merge `origin/main`, and the headline fix from this review: the xAI preset is now **opt-in** behind `ZERO_OAUTH_ALLOW_PRESETS=1`. `gofmt -l ./internal/oauth ./internal/cli/auth.go ./internal/cli/oauth_provider.go` is empty, `go vet ./...` clean, `go build ./...` clean. `go test ./internal/oauth/...` passes (8.6s), `go test ./internal/provideroauth/...` passes. `internal/cli` has a pre-existing flake (`TestRunExecUsesProjectConfigAndOpenAICompatibleProvider`) that also fails on `main` — not a regression.

**Headline fix (resolves prior BLOCKER on hardcoded xAI client_id):**

The hardcoded `b1a00492-…` client_id in `internal/oauth/presets.go:38` is **still there** (any preset has to encode something), but the preset is now **inert unless the operator opts in**. The wiring:

- `presets.go:54-64` — `presetsAllowed(env)` returns `true` only when `ZERO_OAUTH_ALLOW_PRESETS` is `1`/`true`/`yes`/`on`. Otherwise the preset is never consulted.
- `providers.go:68-71` — `ResolveConfig` only fetches the preset when `presetsAllowed` is true. Without the opt-in, every field of the config must come from a `ZERO_OAUTH_XAI_*` env var.
- `providers.go:84` — when a user tries to log in to xAI without configuring it, the error message hints at `ZERO_OAUTH_ALLOW_PRESETS=1`. Discoverable, not silent.
- `presets_test.go:88-93` — regression test asserts that xAI **does not** resolve from the preset without the opt-in, and that the error string mentions the opt-in. Good test.
- `docs/oauth-subscriptions.md` (NEW, 184 lines) — documents the two-mode model ("OAuth login for a provider/gateway" vs "use a ChatGPT/Claude *subscription* via local proxy") and explicitly calls out that presets carry a third-party client identity and are off by default.

This matches the spirit of "ship a preset but don't make it the default credential path." The operator's machine only uses the preset's client_id after explicit opt-in, and any preset field is overridable via env. That's the right shape for a feature like this.

**Security invariants (verified by reading the diff):**

- **PKCE S256 mandatory**, "plain" refused (`oauth.go:81-82` → `ErrPKCEDowngrade`). PKCE binds the auth code to this client so a leaked callback URL alone is useless.
- **CSRF state mandatory and verified** on the callback (`oauth.go:83-84` → `ErrStateMismatch`). `loopback.go:37-39` refuses an empty state at listener construction (fail-closed).
- **Loopback server binds 127.0.0.1 only** on an OS-assigned port (`loopback.go:40` — `net.Listen("tcp", "127.0.0.1:0")`). No `0.0.0.0`, no external exposure. Single-use via a buffered result channel; the server is `Close()`-d after the callback or on context cancel.
- **Token files are 0600** (`store.go:279`, also called out in package doc). Default mode keeps plaintext; opt-in `ZERO_OAUTH_STORAGE=aes-gcm` (or `keyring` on macOS/Linux) wraps with AES-256-GCM (`encrypt.go`). Worth noting: this PR lands the **encrypted-at-rest** path that the prior #210/#213 PRs staged — the store now reads/writes both shapes transparently.
- **Refresh-and-retry on 401** is already in the provider layer (per package doc; not introduced by this PR but exercised by the new tests).
- **OpenRouter** is a separate bespoke flow (`internal/provideroauth/openrouter.go`) using PKCE-only — no client_id needed. Not gated behind `ALLOW_PRESETS` because there's no preset to gate. Minted API key is stored as a normal credential.

**Other things to flag (not blockers):**

- `internal/oauth/presets.go:38` — the xAI client_id is still hardcoded. Given it's now opt-in only, this is fine, but the comment block (lines 29-35) already calls out the "may change without notice" caveat and points to the env override path. Good.
- `internal/cli/auth.go:380` — error message at the CLI layer also mentions `ZERO_OAUTH_ALLOW_PRESETS=1`. Discoverable from both the engine and the CLI.
- `docs/oauth-subscriptions.md` — the "ChatGPT / Claude subscriptions" section is honest about the proxy requirement. Saves a support ticket.

**Overall:** the headline blocker from the prior review is resolved — the preset is now an opt-in, not a default. The PR also folds in the encrypted-at-rest store (#213) so the OAuth engine now has a complete security posture. ~3.1k LoC across 37 files, all green in the relevant package tests, 1.5k LoC of new tests, no regressions on main. Merging.

Vasanthdev2004 (approved):

PR #218, opened 2026-06-16
web_search robustness + chat UX: collapse, clipboard, scroll-pin, proactive search

gnanam1990:
## Summary

A batch of independent fixes that harden `web_search` and make the chat surface much nicer to live in. Each is small, tested, and self-contained.

## web_search

- **Don't offer the built-in `web_search` when no backend is configured.** `CoreNetworkTools` registered it unconditionally, so a model with an MCP search tool (e.g. Exa) burned several calls + high-risk permission prompts on a tool that can only return "no backend configured" before falling back. Now it's registered only when `ZERO_WEBSEARCH_BASE_URL` (or `ZERO_WEBSEARCH_BACKEND`) is set.
- **#209 follow-ups:**
  - Tolerant `score` decode — a stringified / malformed / null provider `score` no longer makes `json.Unmarshal` discard the *whole* response; it degrades to "absent".
  - Score render gate rounds first (`>= 0.01`), so tiny positives and negatives no longer print a noisy `score 0.00`.
  - The domain-filter error now goes through `redactWebSearchText`, matching the other error paths.
  - Trailing-dot FQDN symmetry: `react.dev.` and `react.dev` normalize the same on the allowlist and result sides.

## Chat UX

- **Collapse long tool output by default.** Output longer than the live body cap renders as a one-line `▸ N lines — click to expand` (click while live to expand `▾`), so the transcript stays scannable. Diff tools (`edit_file`/`apply_patch`/`write_file`), the uncapped detailed view (`Ctrl+O`), and short output are never collapsed. Reuses the existing reasoning-row collapse, so collapsed rows flush to scrollback clean.
- **Clean MCP tool labels.** Tool cards showed the raw `mcp_<server>_<tool>`; `toolDisplayName` now renders a clean label (`mcp_exa_web_search_exa` → `web search`) with the query beside it. Built-in names are unchanged.
- **Hold the scroll position while output streams.** The viewport offset is measured from the bottom, so streaming dragged the user off whatever they'd scrolled up to read. It now pins the absolute position while scrolled up (bumping the offset by the body's growth) and follows the tail at the bottom.
- **Copy selection to the native clipboard.** Selecting transcript text auto-copies on release, but it only used OSC52 — unsupported by macOS Terminal.app and off-by-default elsewhere. Now it copies via the native clipboard (`pbcopy` / `clip.exe` / `xclip`) with OSC52 as the SSH/remote fallback. (`atotto/clipboard` was already in the module graph; promoted to a direct import — no new dependency.)

## Provider setup

- **Stop re-running onboarding when a usable provider already exists.** If the active provider lacked a credential, the TUI forced first-run onboarding on every launch (accumulating duplicate providers). Now it falls back to the first usable saved provider (preferring a remote keyed one over a local endpoint); onboarding only runs on a genuinely fresh setup. Switch the active provider any time with `zero providers use <name>`.

## Agent behavior

- **Search the web proactively instead of answering "I don't know".** When a web search/fetch tool is available and the user asks about something external/uncertain/current (a product, library, version, recent event), the system prompt now tells the model to search and read results before answering, rather than replying that it doesn't know without checking.

## Testing
- gofmt, `go vet`, build (host + linux + windows), `go test ./... -race` — green except 3 pre-existing `internal/cli` failures (doctor/exec, network/provider-dependent) that fail identically on `main`.
- staticcheck (no new findings). New unit tests cover each change: score parse/render, dead-tool registration, native-clipboard cmd, collapse/toggle + diff/detailed exemptions, MCP label cleanup, onboarding fallback, scroll-pin, and the proactive-search prompt rule.
- No new module dependencies.

## Notes
- The OAuth provider-login feature is a separate PR (#217). The "recognize an OAuth login as a credential" tweak for onboarding belongs with that feature and is intentionally not in this PR (it would be incorrect without the OAuth runtime wiring).

## Release Notes

* **New Features**
  * Tool result cards in the chat transcript can now be collapsed/expanded.
  * Chat transcript scrolling stays pinned to your reading position while you’re scrolled up.

* **Improvements**
  * The agent is guided to check the web before answering when information may be time-sensitive.
  * Tool names from MCP are displayed with cleaner, user-friendly labels.
  * `web_search` is more resilient and handles domain matching more consistently (including trailing dots).
  * CLI startup can reuse a previously configured provider to avoid rerunning setup.

* **Bug Fixes**
  * `web_search` is no longer available when no search backend is configured.
  * Copy-to-clipboard now prefers the native system clipboard and reports copy failures more accurately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.24.2
 toolchain go1.26.4
 
 require (
+	github.com/atotto/clipboard v0.1.4
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.1-0.20250404203927-76690c660834
@@ -15,7 +16,6 @@ require (
 )
 
 require (
-	github.com/atotto/clipboard v0.1.4 // indirect
 	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
 	github.com/charmbracelet/colorprofile v0.4.1 // indirect
 	github.com/charmbracelet/x/cellbuf v0.0.15 // indirect
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -71,6 +71,11 @@ work.
   managers). It is not a substitute for the file tools.
 - Treat tool output as ground truth. If a command fails, read the error, form a
   hypothesis, and address the root cause; do not retry the same call blindly.
+- When a web search or web fetch tool is available and the user asks about
+  something external you are unsure of or that may be current — a product,
+  library, company, price, version, or a recent event — search the web and read
+  the results before answering. Do not reply that you don't know without searching
+  first.
 
 ## Permission and safety
 
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -71,11 +71,17 @@ work.
   managers). It is not a substitute for the file tools.
 - Treat tool output as ground truth. If a command fails, read the error, form a
   hypothesis, and address the root cause; do not retry the same call blindly.
-- When a web search or web fetch tool is available and the user asks about
-  something external you are unsure of or that may be current — a product,
-  library, company, price, version, or a recent event — search the web and read
-  the results before answering. Do not reply that you don't know without searching
-  first.
+- When a web search or fetch tool is available, search the web before answering
+  about an external entity, product, library, model, company, version, or recent
+  release you do not recognize. Do not guess, and do not assume the most common
+  meaning of an ambiguous name — use the conversation's domain (this is an AI /
+  software tool) to disambiguate. If the first results look off-topic, refine the
+  query and search again before answering; never reply that you don't know without
+  searching first.
+- Do not web-search timeless facts, fundamentals, or anything answerable from the
+  workspace — read the code with the file tools instead. Keep queries short (a few
+  words), start broad then narrow, and scale the number of searches to the
+  question: one for a single fact, a few for a deeper answer.
 
 ## Permission and safety
 
--- a/internal/agent/system_prompt_test.go
+++ b/internal/agent/system_prompt_test.go
@@ -19,7 +19,8 @@ func TestCoreSystemPromptIncludesCodingQualityRules(t *testing.T) {
 		"verify after edits",
 		"honor the active permission mode",
 		"avoid broad refactors",
-		"search the web and read",
+		"search the web before answering",
+		"do not recognize",
 		"final response",
 	} {
 		if !strings.Contains(prompt, want) {
--- a/internal/agent/system_prompt_test.go
+++ b/internal/agent/system_prompt_test.go
@@ -19,6 +19,7 @@ func TestCoreSystemPromptIncludesCodingQualityRules(t *testing.T) {
 		"verify after edits",
 		"honor the active permission mode",
 		"avoid broad refactors",
+		"search the web and read",
 		"final response",
 	} {
 		if !strings.Contains(prompt, want) {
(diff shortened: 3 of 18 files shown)


Vasanthdev2004 (approved):
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr218` at `6b33ad5`. `gofmt -l .` flagged `internal/cli/setup_fallback_test.go` (fixed locally; author should re-run `gofmt` before merge). `go vet` and `go build` clean. `go test ./internal/tui/...` and `./internal/tools/...` pass. The pre-existing `TestRunExecUsesProjectConfigAndOpenAICompatibleProvider` is not in the touched packages and isn't exercised.

**What it does:**

- **`internal/tools/web_search.go`** — promotes the score field to a typed `lenientScore` that accepts JSON numbers, JSON strings, and a few string-shaped variants (e.g. `"0.91"`). A single bad score on one result no longer drops the whole response — it just leaves that result's score at the zero value. The renderer rounds to 2 decimal places and **skips any score that rounds to `< 0.01`** so the common case stays tidy and a noisy "score 0.00" never appears.
- **`internal/cli/setup.go`** — adds `firstUsableProvider`: walks the saved provider profiles and picks the first one that has an inline credential (or is no-auth/local) so the CLI falls back to an already-configured login when the active provider happens to lack one, instead of re-running onboarding every launch. A local (loopback or `Local` catalog flag) provider is preferred as a fallback only when no remote keyed provider is reachable.
- **`internal/cli/setup_fallback_test.go`** — covers the new fallback logic. (Author: please `gofmt -w` this file before merge — it was the only `gofmt -l` hit in the PR.)
- **TUI UX** — tool-card collapse (long tool output gets a "show more" affordance), clipboard copy on assistant messages, scroll-pin (auto-scroll only when the user is at the bottom), proactive search hint in the system prompt.
- **`internal/tools/registry.go`** — registers the new `web_search` behavior under a stable name and adds a small test.
- **`internal/specialist/manifest_test.go`** — one assertion update for the new tool label.

**Security invariants:** no credential-handling changes, no logging changes, no auth changes. The `lenientScore` parser rejects values that overflow `float64` by leaving the field at zero — i.e. it fails closed at the field level, not the result level, which is the correct call (one bad score shouldn't cost you all results). The setup fallback only consults existing profiles; it never reads or prompts for new credentials.

**One small nitpick (not a blocker):** the system-prompt hint for proactive search is a good idea but it's a hint in a `system_prompt.md` that no model is required to follow. If the proactive-search behavior matters, consider gating it on a config flag (`zero search proactive: true`) so users can opt out — but that's a product call, not a code one.

Merging once `gofmt` is clean.

gnanam1990:
Pushed `b0463a6` addressing CodeRabbit's review:

- `firstUsableProvider` skips a profile whose CatalogID won't resolve **and** has no `BaseURL` (a stale CatalogID *with* a BaseURL still works as a custom endpoint).
- `providerProfileIsLocal` matches on the parsed URL hostname instead of a substring, so hosts like `notlocalhost.com` aren't misread as local.
- `lenientScore` rejects non-finite scores (`ParseFloat` accepts `"NaN"`/`"Inf"`).
- `syncChatScroll` holds the pinned offset when the body **shrinks** (signed delta, clamped at 0), not only when it grows.
- copy reports **"Copy failed"** when both the native clipboard and the OSC52 fallback fail, instead of claiming success.
- dropped the vestigial `ZERO_WEBSEARCH_BACKEND` test env — it's read by no product code (only `ZERO_WEBSEARCH_BASE_URL` is honored after the native backend was reverted), so it misrepresented a config knob.

Added tests for the local-host parsing, the catalog/BaseURL skip, and the non-finite score. gofmt / vet / build / `go test -race` are green.

@coderabbitai review

gnanam1990:
@coderabbit approvable ?

gnanam1990:
@coderabbit approve

PR #219, opened 2026-06-16
Polish TUI title bar

anandh8x:
Summary:
- replace the old lime 0/zero header cluster with a quieter branch + workspace title bar
- keep provider/model on the right and preserve responsive fallbacks
- update title-bar tests with generic workspace paths

Tests:
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Title bar now prominently shows workspace path, git branch indicator, and model/provider info with adaptive layout across terminal widths.
  * Added pull/merge request status display in the TUI when detected, including a PR link and local diff statistics.
* **Tests**
  * Updated title bar regression coverage for the new workspace/branch/model behavior.
  * Added unit tests for PR/MR detection, diff-stat computation, and PR segment/title/footer rendering.
* **Chores / Style**
  * Refined user prompt framing and theme styling for git diff markers and footer visuals.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -67,6 +67,8 @@ type model struct {
 	sessionEvents          []sessions.Event
 	usageTracker           *usage.Tracker
 	sessionCompactor       SessionCompactor
+	prService              *PrService
+	prState                PrState
 	runtimeMessageSink     func(tea.Msg)
 	agentOptions           agent.Options
 	notifier               *notify.Notifier
@@ -254,6 +256,10 @@ type doctorCommandResultMsg struct {
 	text string
 }
 
+type prStateMsg struct {
+	state PrState
+}
+
 type permissionDecision = agent.PermissionDecisionAction
 
 const (
@@ -335,6 +341,10 @@ func newModel(ctx context.Context, options Options) model {
 	if usageTracker == nil {
 		usageTracker = usage.NewTracker(usage.TrackerOptions{})
 	}
+	prService := options.PrService
+	if prService == nil {
+		prService = NewPrService(cwd)
+	}
 	doctorUserConfigPath := options.DoctorUserConfigPath
 	if doctorUserConfigPath == "" {
 		doctorUserConfigPath = options.UserConfigPath
@@ -398,6 +408,8 @@ func newModel(ctx context.Context, options Options) model {
 		userAgent:              options.UserAgent,
 		usageTracker:           usageTracker,
 		transcript:             initialTranscript(),
+		prService:              prService,
+		prState:                prService.GetState(),
 		input:                  input,
 		spinner:                runSpinner,
 		now:                    time.Now,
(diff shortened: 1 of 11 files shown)


Vasanthdev2004 (approved):
### Verdict: APPROVED ✅

Reviewed locally in worktree `review-pr219` at `0cda25c`. 4 commits on top of main: `f4a1365 Polish TUI title bar` → `684b35a Show PR diff stats in TUI title bar` → `aa3885c Make submitted prompt bubble terminal-safe` → `0cda25c Address PR status review feedback`. `gofmt -l .` empty, `go vet ./...` clean, `go build ./...` clean, `go test ./internal/tui/...` passes (2.7s, 6 new tests + all existing).

**What it does:**

1. **Quieter title bar** — drops the ` 0 zero` brand badge. Left side now shows `branch` + optional PR segment + `cwd`. Right side keeps `provider/model` (and the context-window strip at full width). All four width tiers (`tierFull`/`tierMedium`/`tierNarrow`/`tiny`) get a fallback ladder that drops segments gracefully instead of crowding.
2. **PR/MR status display** — new `internal/tui/pr_status.go` (388 lines) adds a `PrService` that runs `gh pr view --json url,number,baseRefName` (GitHub) or `glab mr view --json web_url,iid,target_branch` (GitLab) on a 60-second tick, with an 8-second per-command timeout, and computes local diff stats via `git merge-base origin/<base> HEAD` + `git diff --numstat`. The result renders as a hyperlinked `+X -Y #NNN` segment after the branch.
3. **Submit-prompt safety** — small follow-up ensuring the prompt bubble renders cleanly across terminal widths (no half-bubble truncation on submit).
4. **Test updates** — the existing title-bar tests get generic workspace paths (no more coupling to a specific repo); new tests cover PR detection, diff-stat parsing, segment building, and the title-bar render with PR data.

- **`exec.CommandContext`** with the per-call timeout (`prCommandTimeout = 8s`). No shell — `exec.Command(name, args...)` is invoked with explicit argv, no `sh -c`, no string interpolation. No command-injection surface from a `baseRefName` value (it flows as a single `git diff` argument).
- **cwd is set via `command.Dir`**, not interpolated into a path. The branch name from `gh pr view` JSON is untrusted input from the perspective of the daemon, but it never reaches a shell — only `git`'s argv, where the worst case is an unparseable ref.
- **JSON parsing tolerant** — `json.RawMessage` for `number`/`iid` (GitHub returns int, GitLab returns int; the author defensively decodes both string and int). `parseNumStatValue` fails soft (returns 0) on bad input rather than panicking.
- **Bounded work** — 8s per command, 60s refresh, single goroutine per refresh, listener list is append-copied under lock so a slow listener can't block `setState`.
- **Watcher lifecycle** — `Init` returns a `tea.Cmd` that starts the watcher; `quit()` (and the deferred-quit path) calls `stopPRWatcher()` which cancels the watcher context and clears the stop func. No goroutine leak; verified by reading the call sites at `model.go:480-490` and `model.go:1090-1100`.
- **Stderr is captured, not surfaced** — errors from `gh`/`glab`/`git` flow back to the caller as a wrapped error, never to the terminal.

**One small observation (not a blocker):**

`getLocalDiffStats` falls back to `origin/<baseRefName>` when `git merge-base` fails. If the user is offline or hasn't fetched, the diff stats silently become "all my local changes since I last fetched", which can be confusing. Not a correctness bug — git's own `diff` semantics are well-defined — but a UX wrinkle. Out of scope for a title-bar polish PR.

**Overall:** small, focused, well-tested TUI polish. The PR-status feature is the headline addition and it's implemented safely (no shell, no credential handling, bounded work, clean lifecycle). Merging.

PR #220, opened 2026-06-16
TUI polish and fixes

anandh8x:
## Summary
- persist /model selections to the active provider in user config
- add config writer coverage for updating provider models
- add TUI coverage for model switch persistence across restarts

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/config ./internal/tui
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Model selections made via the model command are now automatically saved to your user configuration and persist across sessions.

* **Bug Fixes**
  * Improved alternate-screen transcript rendering with a fixed/pinned title bar.
  * Updated mouse targeting and transcript line selection/overlay positioning so interactions align with the on-screen layout while scrolling.

* **Tests**
  * Added/extended tests covering model persistence (including whitespace/error cases) and validating pinned-title behavior and mouse-based interaction accuracy.


gnanam1990:
## Review — TUI polish and fixes

**Verdict: LGTM with a couple of minor nits.** Focused and well-tested.

### Strengths
- **The pinned-title-bar refactor is correct where it matters.** Render (`scrollableTranscriptView`) and *all* hit-testing paths (`transcriptViewportStart`, `mouseOverComposer`, `overlayMouseTop`) now derive geometry from one shared `scrollableTranscriptFrame(pinnedTitleBar, footerView)`. Because render and mouse math use the same header/body/footer split, clicks map to the right line — that's the real fix, and the `top` offset is threaded consistently (`bodyY := start + msg.Y - top`). Frame clamping degrades sanely on tiny terminals (drops the header before starving the body).
- `titleBarInTranscriptBody()` (`!altScreen && !headerPrinted`) cleanly separates the inline-scrollback title from the pinned alt-screen title; `startup.go`/`transcript_selection.go` updated so it isn't double-counted.
- `SetProviderModel` mirrors the existing `SetActiveProvider` writer; `persistSelectedModel` no-ops gracefully and surfaces `saved: user config` / `saved: no (err)`. Good coverage (update + credential preserved + unrelated provider untouched + unknown-provider-no-rewrite + the end-to-end `/model` persistence test).

### Minor nits (non-blocking)
1. `SetProviderModel` has no permission-preservation test, unlike `SetActiveProvider` (`TestSetActiveProviderTightensExistingConfigFilePermissions`) — worth adding for consistency.
2. The "provider not in user config → `saved: no (...)`" branch is graceful but untested.
3. Read-modify-write without an atomicity guarantee — fine since it matches the other writers, but if `writeConfigFile` isn't temp+rename that's a pre-existing concern across all of them.

### Heads-up: merge collision
This touches `mouse.go` / `model.go` / `transcript_selection.go`, which #217 also modifies and #222 rewrites for Bubble Tea v2. Expect conflicts — see the note on #222 re: merge order.

gnanam1990 (approved):
**Approving** — clean, well-tested, and CI is green across all platforms. The viewport-math consolidation through a single `scrollableTranscriptFrame` is a nice improvement beyond the surface feature, and routing both `/model <x>` and the model picker through `handleModelCommand` means a selection persists either way.

A few non-blocking nits for a follow-up (or now — your call):

- **Display vs. persisted value**: the status line reports `target.modelID` while `persistSelectedModel` writes `nextProfile.Model`. They're equal in the covered path, so this isn't a known bug — but persisting `target.modelID` (or asserting the two match) removes any latent risk of saving a different value than was shown.
- **Project-only configs**: persistence targets only `userConfigPath`, so with just a project config the selection silently doesn't persist and no `saved:` line appears. Consider surfacing something like `saved: no (no user config)` so it isn't silently dropped.
- **Test coverage**: the happy path is well covered; the `persistErr` branch (`saved: no (...)`) and the empty-`userConfigPath` no-op aren't. A small test for each would lock in the behavior.
- **Readability**: the short-terminal clamping in `scrollableTranscriptFrame` (drop the header first, then re-clamp the footer) is correct and the priority is sensible — a one-line comment on the precedence (composer wins, header sacrificed) would help future readers.

Optional: a quick manual alt-screen smoke — scroll (title stays pinned), click a reasoning toggle, drag-select transcript text, open the model picker — since the overlay/mouse coordinate edges are the kind of thing that passes unit tests but can still land a row off live.

Heads-up on merge ordering (not a blocker here): #224 also touches `model.go` / `command_center.go`, so whichever of #220/#224 lands second will need a rebase.

LGTM 👍

PR #221, opened 2026-06-16
feat(tui): rotating "working word" for the liveness spinner

Vasanthdev2004:
## What

The assistant interim block used to show a static `"working…"` placeholder while the model was generating. That felt dead on long runs. Replace it with a rotating gerund tuned for the Gitlawb / OpenFable brand (note: the rebrand is in progress; the liveness spinner ships the new-name verbs ahead of the rest of the project):

- **Brand anchors** (8, weighted 1.5×): `gitlawbmaxxing`, `openfablemaxxing`, `openfabling`, `gitlawbing`, `gitifying`, `tokenizing`, `context-stuffing`, `prompt-wrangling`
- **Project features as verbs** (8): `worktree-walking`, `branch-bending`, `diff-reading`, `sandboxing`, `swarming`, `daemon-summoning`, `tui-painting`, `queue-juggling`
- **Gen-Z crowd-pleasers** (5): `maxxing`, `pilled`, `aura-farming`, `vibe-checking`, `cooking`
- **Classic gerunds** (11): `cogitating`, `contemplating`, `synergizing`, `brewing`, `percolating`, `fermenting`, `wrangling`, `schlepping`, `booping`, `tomfoolering`, `merge-merging`

32 unique verbs, 40 in the weighted ring (brand entries duplicated), lowercase throughout. The taste call on the meme-y entries is the author's, explicitly made up front in the `vibeVerbs` comment, and called out below in "Out of scope" — no config escape hatch in this PR.

## How

- New `internal/tui/working_words.go` (~165 LoC): a small ring-buffer type (`workingWords`) with `Tick()`, `Current()`, `Reset()` and nil-safe methods. The brand/feature/vibe/classics lists are package-level vars so tests can introspect them.
- The verb advances every `WorkingWordsStepEvery = 12` spinner ticks (~1Hz at the 80ms spinner cadence). The glyph spins fast; the word turns over slowly enough to read. The model owns the gate counter so the `workingWords` ring stays a dumb "advance one slot" type and tests can tick it rapidly without waiting for a real spinner.
- `Reset()` is called when a new run starts (in `model.go:launchPrompt` and `spec_mode.go:approveSpecReview`) so the user sees `gitlawbmaxxing` first when the new run begins instead of mid-rotation. The step counter is also reset on these transitions so a partial countdown from the previous run doesn't carry over.
- The only user-visible change is in `interimBlock`: `zeroTheme.muted.Render("working…")` → `zeroTheme.accent.Render(m.spinner.View()) + " " + zeroTheme.muted.Render(m.workingVerb.Current())`.

## Why a ring, not a random pick

Two reasons we don't randomize per turn:

1. The spinner already rotates on an 80ms tick, so a static verb looks like a bug. A ring stays in lockstep with the glyph and feels intentional.
2. Deterministic order is easier to test (the first frame is always `gitlawbmaxxing`), and a long wait reveals the project name 1.5× as often as a random pick would.

## Tests

7 tests in `internal/tui/working_words_test.go`:

- `TestWorkingWordsCurrentIsFirstBrand` — deterministic first frame
- `TestWorkingWordsTickAdvances` — three ticks produce three distinct words
- `TestWorkingWordsTickWraps` — full ring brings you back to `gitlawbmaxxing`
- `TestWorkingWordsResetRewinds` — `Reset()` after 17 ticks returns to first
- `TestWorkingWordsBrandWeighted` — brand verbs are ≥30% of the ring (16/40 = 40% expected)
- `TestWorkingWordsBaseListIntegrity` — base list (pre-weighting) has ≥25 entries, no empty strings, no duplicates
- `TestWorkingWordsNilSafe` — `Tick()`, `Reset()`, `Current()` all safe on a nil receiver (so the render hot path doesn't need a nil check)

Plus `TestWorkingVerbAdvancesOnStepEverySpinnerTicks` in `model_test.go` — locks in the cadence: after `WorkingWordsStepEvery - 1` ticks the verb is unchanged, after the `WorkingWordsStepEvery`-th tick it advances.

Plus the existing `TestInterimBlockShowsStreamingTextWithCursor` updated to assert `"gitlawbmaxxing"` instead of `"working…"` as the new default first frame.

Full `internal/tui/` suite green. `go vet`, `gofmt`, `go build` all clean.

## Out of scope (deliberate)

- **No config wiring** — list is hardcoded. A future PR can add `~/.zero/settings.json` loading if users ask.
- **No change to `renderRunningToolCard`** — the running-tool state (per-tool card with the tool name) keeps its existing spinner-only head. Only the assistant-generation liveness surface gets the verb.
- **No past-tense summary footer** — can be a follow-up.
- **No time-escalating sub-line** (`Hmm…`, `Untangling some thoughts…`) — the reasoning block is a separate surface and doesn't share this one.
- **Tone of `vibeVerbs`** (`pilled`, `aura-farming`, `maxxing`, etc.) — these ship to every user with no escape hatch. Author's call: keep as-is and call the trade-off out in code, rather than neuter the flavor. Easy to revisit in a follow-up if anyone pushes back.

## File summary

| File | Change |
|---|---|
| `internal/tui/working_words.go` | NEW, 165 LoC |
| `internal/tui/working_words_test.go` | NEW, 116 LoC |
| `internal/tui/model.go` | +24 LoC: field, gate counter, init, tick, reset, render |
| `internal/tui/spec_mode.go` | +4 LoC: reset on spec approval → impl run |
| `internal/tui/model_test.go` | +41 LoC: cadence test |
| `internal/tui/rendering_lime_test.go` | 1 line updated: assert `gitlawbmaxxing` |
| **Total** | **6 files, +351 LoC** (effective 200 LoC on the working-words logic, 151 LoC on tests/asserts) |

## Summary by CodeRabbit

* **New Features**
  * Enhanced the loading indicator to display rotating action words (with proper cadence) instead of a static placeholder.
  * Ensured the action-word sequence resets cleanly when starting new runs, including in spec-approval flows.

* **Bug Fixes**
  * Updated interim rendering fallback text used during streaming tests to match the new loading indicator behavior.

* **Tests**
  * Added unit tests for the rotating verb ring (initial value, advancement, wraparound, reset, nil-safety, and weighting).
  * Updated an existing interim-block rendering test expectation.
  * Added a model-level test validating the verb only advances on the configured spinner tick interval.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -97,7 +97,11 @@ type model struct {
 	// spinner animates the running-tool glyph in card heads. Its tick is started
 	// with each run and stops itself once pending clears (the TickMsg is simply
 	// not forwarded), so an idle UI schedules no timers.
-	spinner       spinner.Model
+	spinner spinner.Model
+	// workingVerb holds the rotating "gitlawbmaxxing / zeroling / …" label the
+	// assistant interim block displays while pending. It advances on every
+	// spinner tick (see TickMsg branch) so its cadence matches the glyph.
+	workingVerb   *workingWords
 	pending       bool
 	queuedMessage string
 	exiting       bool
@@ -435,6 +439,7 @@ func newModel(ctx context.Context, options Options) model {
 		prState:                prService.GetState(),
 		input:                  input,
 		spinner:                runSpinner,
+		workingVerb:            newWorkingWords(),
 		now:                    time.Now,
 		notifier:               notifier,
 		altScreen:              options.AltScreen,
@@ -927,6 +932,10 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		var cmd tea.Cmd
 		m.spinner, cmd = m.spinner.Update(msg)
+		// workingVerb shares the spinner's cadence so the verb and the glyph
+		// stay in lockstep (one new word per glyph frame). Nil-safe; nil check
+		// is unnecessary because the type's methods handle a nil receiver.
+		m.workingVerb.Tick()
 		if m.compactInFlight {
 			m.compactFrame++
 			m = m.setCompactStatusRow(m.compactText(true))
@@ -1557,7 +1566,7 @@ func (m model) interimBlock(width int) string {
 		if len(blocks) > 0 {
 			return strings.Join(blocks, "\n")
 		}
-		return zeroTheme.accent.Render(m.spinner.View()) + " " + zeroTheme.muted.Render("working…")
+		return zeroTheme.accent.Render(m.spinner.View()) + " " + zeroTheme.muted.Render(m.workingVerb.Current())
 	}
 	lines := renderAssistantMarkdownText(text, assistantMeasure(width), width)
 	for index, line := range lines {
@@ -2351,6 +2360,9 @@ func (m model) launchPrompt(prompt string) (model, tea.Cmd) {
 	m.activeRunID = m.runID
 	m.runCancel = cancel
 	m.pending = true
+	// Rewind the verb rotation so the user sees "gitlawbmaxxing" first when
+	// the new run starts (instead of mid-rotation from a prior turn).
+	m.workingVerb.Reset()
 	return m, tea.Batch(m.runAgent(m.activeRunID, runCtx, prompt, turnImages), m.spinner.Tick)
 }
 
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -98,16 +98,20 @@ type model struct {
 	// with each run and stops itself once pending clears (the TickMsg is simply
 	// not forwarded), so an idle UI schedules no timers.
 	spinner spinner.Model
-	// workingVerb holds the rotating "gitlawbmaxxing / zeroling / …" label the
-	// assistant interim block displays while pending. It advances on every
-	// spinner tick (see TickMsg branch) so its cadence matches the glyph.
-	workingVerb   *workingWords
-	pending       bool
-	queuedMessage string
-	exiting       bool
-	runCancel     context.CancelFunc
-	runID         int
-	activeRunID   int
+	// workingVerb holds the rotating "gitlawbmaxxing / openfablemaxxing / …"
+	// label the assistant interim block displays while pending. The verb
+	// advances at WorkingWordsStepEvery-tick intervals (≈1Hz at the 80ms
+	// spinner cadence) so the glyph can spin fast and the word still be
+	// readable; workingVerbTicks is the gate counter. See the spinner
+	// TickMsg branch and launchPrompt for the advance/reset.
+	workingVerb      *workingWords
+	workingVerbTicks int
+	pending          bool
+	queuedMessage    string
+	exiting          bool
+	runCancel        context.CancelFunc
+	runID            int
+	activeRunID      int
 	// flushRunIDs holds the ids of runs cancelled while still in flight, mapped
 	// to the session they were recording into AT CANCEL TIME. Each cancelled
 	// agent goroutine keeps running to completion and returns its accumulated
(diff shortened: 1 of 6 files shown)


Vasanthdev2004:
Update: renamed two brand verbs that referenced the old name.

- `zeromaxxing` → `openfablemaxxing`
- `zeroling` → `openfabling`

Comments updated to reference OpenFable. All 7 `working_words` tests + the `interimBlock` assertion still pass. The repo / module path / binary name / config dir / env vars are unchanged — this is a wording change in the liveness spinner only, not a project rename.

Force-pushed `c085f1c` → `600d537` to keep the PR as a single commit.

gnanam1990 (commented):
**One thing to resolve before merge, plus a couple of judgment calls.** The ring-buffer type itself is clean, nil-safe, and well-tested (deterministic first frame, wrap, reset, brand weighting, base-list integrity), and CI is green.

- **Brand name is inconsistent — please confirm the intended one.** The code ships `openfablemaxxing` / `openfabling` and the comments say "Gitlawb / OpenFable brand" and "OpenFable's actual features," while the PR description, the repo, and the binary all say **Zero** (the description even lists `zeromaxxing` / `zeroling`). These are user-facing strings sitting right next to the spinner, so the two need to agree. If `Zero → OpenFable` is an intentional rebrand, the description is just stale; otherwise the code is. Either way, align code + tests + description on one name before this merges.
- **Cadence — the verb advances on every spinner tick (~80 ms ≈ 12 words/sec), which is too fast to read.** It registers as a flicker rather than a status; the static `working…` it replaces was at least legible. The glyph and the word don't have to share a cadence — let the glyph spin fast and turn the word over roughly once a second (advance every N ticks).
- **Taste (your call):** `pilled` / `aura-farming` / `maxxing` are user-facing; `pilled` in particular carries connotations some will read as off-tone for a dev tool. Worth a deliberate decision, since config is out of scope here and these ship to everyone.
- **Competitor naming in comments:** the comments name a competitor coding tool by name a few times. If you're keeping the codebase competitor-neutral, trim those down to just the design rationale.

Mechanically it's solid; the brand-name item is the one I'd want sorted before merge since it's literally in the UI.

anandh8x (changes requested):
Requesting changes on the current user-facing behavior before merge.

Findings:

- internal/tui/model.go:894-905: the word advances on every spinner tick. At the spinner cadence this is likely too fast to read and will render as flicker. Please keep the glyph cadence fast but rotate the word at a slower cadence, roughly once per second.
- internal/tui/working_words.go:3-15 and 28-40: the shipped UI words/comments mix brand names. These strings are user-facing next to the spinner, so code, tests, and PR description should align on one intended name before merge.
- internal/tui/working_words.go:93-107: weightedRing says duplicate brand words are interleaved, but the implementation appends the duplicate brand block at the end. Either update the comment and simplify to append brandVerbs directly, or implement the documented interleaving.

Taste call, but worth deciding explicitly: pilled / aura-farming / maxxing are also user-facing for every user, since there is no config escape hatch in this PR.

Vasanthdev2004:
Pushed fixes for the three review threads — force-pushed `8440b60` on top of the rebase onto current main.

**Cadence** (@anandh8x, @gnanam1990): the word now advances every `WorkingWordsStepEvery = 12` spinner ticks (~1Hz at 80ms) instead of every frame. The `workingWords` ring stays a dumb "advance one slot" type; the model owns the gate counter (`workingVerbTicks`) so unit tests can still tick the ring rapidly. New `TestWorkingVerbAdvancesOnStepEverySpinnerTicks` locks the cadence in. Glyph spins fast; word is readable.

**Brand-name inconsistency** (@gnanam1990): kept the rebrand-ahead verbs on purpose and made it explicit in the PR description ("the rebrand is in progress; the liveness spinner ships the new-name verbs ahead of the rest of the project"). The user-facing strings stay `gitlawbmaxxing`, `openfablemaxxing`, `openfabling`, etc. Also fixed a stale `model.go` comment that still said "zeroling" (it was renamed in the previous force-push but the comment was missed).

**weightedRing comment/impl** (@anandh8x, @gnanam1990, CodeRabbit): updated the comment to match the actual behavior (brand duplicates appended sequentially at the end of the ring, not interleaved) and added a one-line note on how to implement true interleaving if it's ever wanted. The loop body was simplified to a direct `append(ring, brandVerbs...)`.

**CodeRabbit nits**: `Reset()` now also checks `len(w.weighted) == 0` for consistency with `Tick()` and `Current()`. `weightedRing` body simplified.

**Taste call on `vibeVerbs`** (@gnanam1990, @anandh8x): kept all five as-is per author call. Added an explicit comment on the `vibeVerbs` slice calling out the trade-off ("these ship to every user with no config escape hatch, so the taste call is the author's and explicitly made up front") so the next reader can see it was deliberate.

**Competitor naming in comments** (@gnanam1990): trimmed the explicit "Claude Code" references in the package comments. Replaced with neutral phrasing ("upstream Claude-Code-compatible spinners", "a few classic gerunds") that keeps the design rationale without naming the competitor. The user-facing verb list is untouched.

Rebase: spec_mode.go had one real conflict (fade seed from main + verb reset from this PR); kept both. Build + vet + gofmt clean; full `./internal/tui/...` suite green except the pre-existing env-fragile test from PR #220 (unrelated to this PR, fails here only because of local env vars, not this code).

anandh8x (approved):
The earlier change requests are addressed. The working word now advances on a gated cadence instead of every spinner tick, the brand/rebrand wording is explicit and aligned across the PR body/code/tests, weightedRing's comments match the append-at-end implementation, Reset has the empty-ring guard, and CodeRabbit's follow-up cadence-test issue is fixed. CI is green.

PR #222, opened 2026-06-16
Migrate TUI to Bubble Tea v2

anandh8x:
## Summary
- migrate the TUI stack to Bubble Tea/Bubbles/Lip Gloss v2
- adapt key, paste, mouse, and view handling to the v2 APIs
- refresh Go module versions and tidy the dependency graph
- make the PDF truncation fixture deterministic across poppler and pure-Go extraction

## Tests
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./internal/tui
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./internal/imageinput -count=1 -v
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./...

## Summary by CodeRabbit

# Release Notes

* **Dependencies**
  * Updated Go to 1.25.0.
  * Updated TUI libraries to newer v2 module paths.

* **Bug Fixes**
  * Hardened OAuth/device-login so stale results no longer affect the active attempt.
  * Improved mouse/overlay hit-testing and chat scrolling behavior (including scroll clamping/pinning).
  * Throttled high-frequency mouse input to keep interactions responsive.

* **Refactor & Tests**
  * Standardized TUI input handling via shared key/mouse helpers and updated/strengthened related tests.


gnanam1990:
## Review — Migrate TUI to Bubble Tea v2

**Verdict: approach is sound, and both of CodeRabbit's blockers already appear fixed on the branch HEAD — the `CHANGES_REQUESTED` looks like a stale verdict.** Residual risk is inherent to a framework swap, not the diff.

### Strengths
- **`input_compat.go` is the right design** — it centralizes the entire v2 input surface (`keyCode/keyText/keyAlt/keyCtrl/keyIs`, `mouseLeftPress/mouseMotion/mouseWheelUp`, type-switches on `MouseClickMsg`/`MouseMotionMsg`) into helpers, so the rest of the TUI stays readable and the v2 API lives in one file. Exactly how to do a migration of this size.

### Both flagged findings look resolved
- `input_compat.go:63` "unused `mouseButton`" → **gone** (no `mouseButton` in the branch anymore).
- `provider_wizard.go:609` "correlate OAuth/device results to the active attempt" → **already implemented**: the msgs now carry `providerID` + `attemptID`, `beginOAuthAttempt()` issues the token, and `applyProviderWizardOAuth`/`applyProviderWizardDeviceCode` gate on `oauthResultMatches(providerID, attemptID)`. (Same stale-result class handled for the onboarding flow in #217 — good the wizard path has it now too.)

### Risks to verify before merge (not blockers)
- A stop-the-world TUI swap — automated tests can't catch rendering/input regressions. Needs real manual driving (key/paste/mouse, alt-screen scroll, overlays, resize) on macOS + Linux + Windows terminals.
- `charmbracelet/ultraviolet` is pinned to a 2026-06-15 pseudo-version (brand new, indirect) — confirm it's intentional and `go.sum` is clean.
- Confirm lint + full CI is green (the original `mouseButton` finding was a lint gate).

### Merge order
#217, #220, and this PR all rewrite the same TUI files, and this one changes the v1→v2 **input API itself**. Recommend landing this migration **first (or last with everyone rebasing onto it)** — otherwise #220/#217 have to be re-expressed in the v2 helper API after the fact.

gnanam1990 (approved):

Vasanthdev2004 (approved):
## APPROVE

Read at `ae17cf0`. Walked through the 28 files (885 +, 197 -), focused on the new behavior:

- in-memory session grants + `sessionMu` in `internal/sandbox/engine.go`
- the new `PermissionDecisionAllowForSession` path in `internal/agent/loop.go:579-587`
- `workspaceWriteAutoAllowed` + the new `apply_patch` short-circuit in `internal/sandbox/engine.go`
- the TUI fallback in `internal/tui/permission_prompt.go:21-53`
- the new `TestRunAllowsWorkspaceWriteWithoutPromptWhenSandboxPolicyPermits` and the session-grant test in `internal/agent/loop_test.go`
- the transcript-row label switch in `internal/tui/session_controls.go`

The shape of this is good. Session grants have a real concurrency primitive, `AvailableDecisions` flows from the engine to the TUI cleanly, and the agent loop's "honor the allow for THIS call regardless" comment on lines 591-598 is the right call. `persistSessionPermissionGrant` returning a `sandbox.Grant` instead of an opaque bool keeps the event payload honest.

Three things I want to flag for follow-up, none blocking:

**1. `apply_patch` workspace auto-allow bypasses the engine's path-list validation, not just the user prompt.**

`workspaceWriteAutoAllowed` returns `true` for `apply_patch` unconditionally, and the engine's path validation iterates `requestPaths(request)` (in `internal/sandbox/risk.go:223`) which does not include the `patch` key. So with `EnforceWorkspace: true` and an in-workspace patch header, the engine approves the call without checking `DenyWrite`/`AllowWrite`. The `apply_patch` tool's own `validatePatchPaths` in `internal/tools/apply_patch.go` only does workspace confinement (no `..`, no absolute paths, `resolveWorkspaceTargetPath`). It does not consult the policy.

The OS-level sandbox (bwrap, sandbox-exec, landlock, Windows ACL) still enforces `DenyWrite` at the kernel level, so this is defense-in-depth, not a hole. But users running in pure permissions mode (no OS sandbox) get a behavior change: a patch that targets a `DenyWrite` path used to fail when the user reviewed it in the prompt, now passes silently. Worth either threading the policy into the apply_patch tool or calling the trade-off out in the PR body so it isn't a surprise.

**2. No test for `apply_patch` + `DenyWrite`.**

`TestRunAllowsWorkspaceWriteWithoutPromptWhenSandboxPolicyPermits` only covers an in-workspace `write_file` against `DefaultPolicy()` (which has no `DenyWrite`). A test that an apply_patch targeting a `DenyWrite` path is denied would close the loop. If the answer is "the apply_patch tool catches it", a test at that layer would be even more useful.

**3. TUI `permissionOptions` fallback offers "always" even with no persistent store.**

In `internal/tui/permission_prompt.go:21-30`, when `AvailableDecisions` is empty the fallback hardcodes all four decisions. The TUI is only invoked for prompts that always populate `AvailableDecisions`, so the fallback is unreachable today. But if a future call site passes a non-prompt event to the TUI, the user can hit "y" with no store and silently get only-this-call behavior. Easy fix: derive the fallback from the presence of a store, same as the engine does in `availablePermissionDecisions`.

Minor: the `persistSessionPermissionGrant` error path is swallowed with `if ... err == nil` (loop.go:583), and that's the right call for "don't deny what the user explicitly allowed", but the asymmetry with the `AlwaysAllow` branch (which builds the permission event from the persisted grant on success) means the session-grant event's `GrantMatched` flips to `false` if persistence fails. Worth a code comment so a future reader doesn't think the event is broken.

Non-blocking follow-ups; happy to see this land.

Vasanthdev2004:
**Heads up — wrong review posted here.**

I just posted a CHANGES_REQUESTED-style review on this PR in error. The body talks about session approvals, the in-memory `sessionMu`, `workspaceWriteAutoAllowed`, and `apply_patch` policy bypass — that is the content of PR #244 (`permission-session-approvals`), not this one (TUI v2 migration). The commit SHA `ae17cf0` in the review is on the `permission-session-approvals` branch, which is not part of #222's history.

PR #222 is already MERGED, so the orphan review doesn't change anything in the codebase. I tried to dismiss it but the dismiss endpoint didn't take (likely because #222 is merged). Treating the review as a no-op; the real review will land on #244.

Sorry for the noise.

PR #223, opened 2026-06-16
feat(tui): age-based fade for streaming assistant text

Vasanthdev2004:
## Summary

Adds a per-line age-based color fade to the live streaming assistant text. While the reply streams in, the freshest line starts in the brand lime (\#caff3f) and gradually settles to the standard off-white ink (\#ececee) over ~1.2 seconds. The effect mirrors the lime spinner glyph in the liveness block — a quiet visual signal that the model is still emitting, not a stylized neon glow.

**Why per-line, not per-rune?** ~10x cheaper, visually equivalent for streaming prose, and survives soft-wrap without grapheme-segmentation code. The per-rune upgrade path is mechanical: replace `lineAges []time.Time` with `charAges []time.Time` and walk the last ~200 runes.

## Design

- **Palette**: 12 steps, interpolated in linear sRGB from `colorAccent` (\#caff3f) to `colorInk` (\#ececee). Pre-computed at `init()` so per-frame lookups are a struct-field access, not a hex parse. Both endpoints are near-white, so the perceptual difference between sRGB and CIELAB is invisible — save the Lab math (and the go-colorful dependency) for the day someone complains.
- **Cadence**: 150ms `streamingFadeTickMsg`, independent of the 80ms spinner tick. A slower cadence is enough for a smooth-feeling fade and keeps the per-frame work cheap.
- **Hard stop at stream end**: `agentResponseMsg` flips `fadeActive = false`; the next tick that lands after that point short-circuits. No settle animation.
- **Reset on resize**: `WindowSizeMsg` zeros `lineAges` and re-stamps `lastStreamActivity` because a width change can re-wrap the in-progress text into a different number of visual lines.
- **Defensive on test fixtures**: test models that pre-populate `m.streamingText` without going through `agentTextMsg` (so `lineAges` is nil) render identically to before — the fade code short-circuits to the base ink color.
- **The last visual line uses `lastStreamActivity`**, not the corresponding `lineAges` entry. This is the rule that keeps the in-progress typing position visibly fresh.

## Spec-mode parity

The spec-impl run path calls `runAgent` directly, bypassing the normal `launchPrompt` seed. `approveSpecReview` now seeds the fade state explicitly so spec-impl streaming text fades the same way regular runs do.

## Files

- **NEW** `internal/tui/streaming_fade.go` (~290 LoC): palette builder, `streamingFadeTick`, `ageDimLine`, `streamingLineBornAt`, `recordStreamingDelta`, `resetStreamingFade`, `styleStreamingLine`.
- **NEW** `internal/tui/streaming_fade_test.go` (~205 LoC): 11 tests covering palette monotonicity, palette intermediates, `ageDimLine` fresh/mid-range/settled/zero-bornAt paths, bucket walk, tick command, delta newline tracking, first-entry seeding, reset, `streamingLineBornAt` last-visual/middle/out-of-range, and `styleStreamingLine` defensive fallbacks.
- **MODIFIED** `internal/tui/model.go` (+~50 LoC): add `lineAges`, `lastStreamActivity`, `fadeActive` fields. Hook `agentTextMsg` (record delta + schedule tick), `agentResponseMsg` (hard-stop), `cancelRun` (reset), `WindowSizeMsg` (reset). New `streamingFadeTickMsg` case. Replace per-line style loop in `interimBlock` with `m.styleStreamingLine`.
- **MODIFIED** `internal/tui/spec_mode.go` (+7 LoC): seed the fade state in `approveSpecReview` so spec-impl streaming text fades.

## Test results

`go test -count=1 -timeout 90s ./internal/tui/...` → 1113 tests, all green.
`go vet ./internal/tui/...` → clean.
`gofmt -l` → clean.

(Pre-existing flake: `TestRunExecUsesProjectConfigAndOpenAICompatibleProvider` in `internal/cli` also fails on `main`; unrelated to this PR.)

## Manual verification

`go run ./cmd/zero` → fire a long prompt ("explain the OAuth engine in this repo in detail") → new lines appear in brand lime and settle to off-white over ~1.2s, line by line. Mid-stream resize: all text snaps to fresh. Stream end: hard-stop to base color, no fade-out animation.

## Summary by CodeRabbit

* **New Features**
  * Added an age-based “fade” effect for streamed assistant text, using a fixed stepped color palette so newly added lines start highlighted and gradually settle into standard styling.

* **Bug Fixes**
  * Improved fade initialization and lifecycle handling across streaming, including clean start/stop, proper behavior on resize, stream completion, and cancellation.

* **Tests**
  * Expanded coverage for palette generation, line age bucketing/clamping, tick scheduling, newline age tracking, fade reset, and renderer fallback behavior.


Vasanthdev2004:
Quick note on the design: I dug into Claude Code to see whether they actually do per-character color fade. **They don't** — the "glow" feel there is a block-boundary split (`StreamingMarkdown` keeps the stable prefix memoized, only the last block re-parses per delta) plus the `dimColor` prop on queued/executing tools. There's no per-rune ANSI interpolation in the streamed text itself.

The closest reference I found in a real Go TUI is `charmbracelet/mods`'s `anim.go`, which uses a Luv-blend gradient for a *cycling shimmer sidecar* next to the typed text — not age decay on the text itself.

So this PR isn't reproducing the Claude Code effect exactly; it's a new effect that *reads* like what people remember. If we want to go further later, the upgrade paths are:
- **Per-rune** (most "alive"): replace `lineAges []time.Time` with `charAges []time.Time`, walk the last ~200 runes. ~3x the work, only matters when a single token takes >500ms to arrive.
- **Lab interpolation** (visually more uniform): swap the linear sRGB math for `lucasb-eyer/go-colorful`'s `BlendLab`. Saves a new dep; buys nothing visible for this ramp.
- **Settle animation at stream end** (200ms ease-out to base color): adds a second ticker and a state. Nice-to-have, not v1.

Happy to follow up on any of these as separate PRs if useful.

gnanam1990 (approved):
**Approving** — well-isolated, genuinely well-tested (palette monotonicity, bucket math, the visual→logical age mapping, and the defensive fallbacks are all covered), and green on every platform. The decoupled 150 ms fade tick that self-perpetuates while `fadeActive` and short-circuits at stream end is a clean lifecycle. A few non-blocking nits:

- **Dead method**: `ensureAgeTickReschedule` (streaming_fade.go) has no callers — the tick is started inline in the `agentTextMsg` branch and rescheduled in the `streamingFadeTickMsg` case. Safe to delete.
- **Spec-impl path never starts the dedicated fade tick.** `approveSpecReview` sets `m.fadeActive = true` before the first delta, so in the `agentTextMsg` branch `startTick := !m.fadeActive` is `false` and `streamingFadeTick()` is never scheduled. The fade then rides on the 80 ms spinner re-renders (batched there) instead of the 150 ms fade ticker — it works, but not via the mechanism the comment describes, and it would freeze if the spinner ever paused mid-stream. Simplest fix: drop the `m.fadeActive = true` and let the first delta start the tick exactly like `launchPrompt` does (or add `streamingFadeTick()` to the batch).
- **Description vs. code**: the body says the ramp is interpolated "in linear sRGB" and that this "saves the go-colorful dependency," but the code uses `lipgloss.Blend1D` (CIELAB, which pulls go-colorful in via lipgloss). No new *direct* dependency, but the description is out of sync with the implementation — worth a quick edit so future readers aren't misled.
- **Documented approximation (fine)**: `streamingLineBornAt` maps visual→logical lines 1:1, so once a logical line soft-wraps, later lines' ages drift by the wrap count. Imperceptible for a 1.2 s prose fade, and the last visual line always uses `lastActivity`, so it degrades gracefully — just flagging it as a real edge if the ramp ever gets slower or longer.

Nice bit of polish. 👍

PR #224, opened 2026-06-16
Improve /resume: interactive session picker + model-generated titles

gnanam1990:
## What & why

`/resume` listed sessions as a static text blob, and every session was titled by its first user message — so prompts that started the same way (and the many empty/failed runs that produced no output) were indistinguishable in the list. This reworks `/resume` into an interactive picker and gives sessions real, model-generated titles.

## Changes

**Interactive session picker**
- Bare `/resume` now opens a single-select overlay (a new `pickerSession` kind) just like `/model` and `/provider`: arrow/type-to-filter/enter. `/resume <id>` and `/resume latest` still resolve directly.
- Each row leads with a precise timestamp (today → `15:04:05`, this year → `Jan _2 15:04`, else date) so same-titled sessions are still distinguishable; the id shows in the row meta.

**Hide empty/failed sessions**
- Runs with no tool calls/results and no real assistant text (e.g. the no-output guardrail stop) are hidden from the picker. They stay on disk — only the picker filters them. A shared `agent.IsNoProgressStop` recognizes that stop marker.

**Model-generated titles**
- New `sessions.Store.UpdateTitle` rewrites only the title under the per-session lock (re-reading the latest metadata first), leaves `UpdatedAt` untouched so a retitle never reorders the list, rejects blanks, and no-ops an unchanged title.
- A bounded digest of the conversation (user/assistant text + tool names, per-message and total caps, skipping the no-output stop) is sent as a one-shot completion via the active provider; the response is cleaned (first content line, quotes/markup/`Title:` label stripped, word- and rune-capped).
- **Auto-title going forward:** after a successful turn, a session still on its default first-message title gets a concise title generated in the background, at most once per session. Failures are non-fatal — the first-message title simply stays.
- **`/retitle` backfill:** titles existing resumable sessions that still have a first-message title, one at a time, skipping empty/failed and already-named sessions, with kickoff and completion status lines.

**Resume correctness fixes**
- Resume now prefers the rehydrated (compaction-aware) event view, matching `zero exec --resume` and the in-TUI `/compact` reload, so resuming a compacted session honors the prior compaction instead of re-inflating the raw log.
- The resume summary reports the model/provider the run will actually continue with, noting the session's recorded values when they differ.
- Removed a dead/unreachable branch in the resume text path.

## Notes
- Generation runs off the Update goroutine; it costs a small extra provider call per title (sequential for `/retitle`, so no burst).
- Built on top of the Bubble Tea v2 migration (#222), merged into this branch.

## Testing
- New unit tests: `UpdateTitle` semantics; digest/cleaning/auto-title one-shot; `/retitle` candidate selection; picker open/hydrate; hide-empties; resume honors prior compaction.
- gofmt, `go vet`, build (host + linux + windows), staticcheck all clean; `internal/sessions` and `internal/tui` pass under `-race`.

## Release Notes

* **New Features**
  * Sessions now automatically generate concise titles after successful conversation turns.
  * Added **`/retitle`** to generate/update concise session titles.
  * **`/resume`** now opens an interactive session picker when used without an ID.

* **Improvements**
  * Resuming prefers compaction-aware session history with safe fallbacks.
  * Title updates persist without changing session ordering, and resumable-session detection more reliably ignores no-progress guardrail output.

* **Bug Fixes**
  * Improved resilience when session history is unavailable (no crashes).

* **Tests**
  * Expanded coverage for title generation/retitling and resume picker behavior.


anandh8x (dismissed):
Requesting changes based on the current diff.

Findings:

- internal/agent/guardrails.go:108: IsNoProgressStop uses strings.Contains, so any legitimate message that quotes the no-output marker can be treated as a failed empty run. That can hide real sessions from /resume and skip title generation. Please tighten this to an exact or otherwise structured guardrail-stop match.
- internal/tui/session.go:201-204: resumeEvents returns the earlier rehydration err when raw fallback ReadEvents fails. It should return rawErr so callers see the actual fallback failure.
- internal/tui/command_center.go:235-238: resumeText dereferences m.sessionStore without a nil guard. newModel normally initializes it, but this fallback method should still avoid a zero-value/model-test panic.
- internal/tui/model_test.go:984-986: the rehydration test only compares lengths. Please compare next.sessionEvents with the rehydrated event slice directly so the test proves /resume loaded the compacted context, not another same-length slice.

CI is green and the feature direction looks fine, but the first two are correctness issues that should be fixed before merge.

gnanam1990:
Thanks for the review — all three addressed in `12474aa`:

- **`IsNoProgressStop` false positives** — it now requires the full stop-answer structure (`Agent stopped after …` prefix + the marker + the `…without making progress.` suffix) instead of a bare `strings.Contains` on the marker. A legitimate message that merely quotes the marker no longer counts, so real sessions aren't hidden from `/resume` and don't lose title generation. Added a test covering the real answer (any turn count) vs. a quoting message / bare marker / prefix- or suffix-only.
- **`resumeEvents` fallback error** — now returns `rawErr` (the actual `ReadEvents` failure) instead of the earlier rehydration error.
- **`resumeText` nil `sessionStore`** — guarded; it renders a safe "store unavailable" message instead of dereferencing nil. (`newSessionPicker` already guarded nil.) Added a nil-store test.

gofmt / vet / build (host+linux+windows) / `-race` / staticcheck all green.

Vasanthdev2004 (changes requested):
## Verdict: **Request changes** — three substantive issues + one minor test-strength finding

This is a strong, well-architected feature — the file split (separate `session_title.go` from `session.go`), the bounded-digest approach, the per-session lock on `UpdateTitle`, the strict no-progress detection, and the one-shot auto-title gating are all solid choices. The test coverage is good. The issues below are correctness/UX gaps that are worth fixing before merge; the architecture stays as-is.

### Findings

#### 1. `titledSessions` is set optimistically — auto-title failures are silent and permanent (UX correctness)

**`internal/tui/session_title.go:250`** (`maybeAutoTitleActiveSession`) and **`:289`** (`startSessionRetitle`) set `m.titledSessions[sessionID] = true` *before* the title cmd fires, so a failed title generation (provider error, model returned no usable title, timeout) leaves the session with its first-message title **forever** for this process. The user has no idea anything happened — no log line, no status row, no retry path, and the only escape is process restart. The PR comment says "on failure the first-message title simply stays" but it stays **permanently**, not just for this turn.

**Suggested fix**: move the `titledSessions[id] = true` into the **success path of the returned msg** (`handleSessionTitleGenerated`), not the cmd-scheduling site. The session id is gated only after the title actually persists. Failure leaves the map untouched, so a future turn (or `/retitle` restart) can retry.

#### 2. `IsNoProgressStop` is structurally loose — a real message that quotes the marker in passing is misclassified

**`internal/agent/guardrails.go:117`** — the current check is `HasPrefix(Prefix) && Contains(Marker) && HasSuffix(Suffix)`, which only requires the three components in order with **arbitrary text in between**. A hostile real message like

> Agent stopped after 3 turns. The marker is "with no output (no visible text and no tool calls)" so here it is: to avoid consuming tokens without making progress.

is currently classified as a no-progress stop and would hide a real session from `/resume` and skip its title generation. (I confirmed this in a quick repro outside the repo.) The PR's own test guards against the "marker only" case but not the "marker as a quote in the middle" case.

**Suggested fix**: after `HasPrefix` succeeds, find the marker's index, then require the suffix to come **immediately after** the marker (whitespace allowed but no arbitrary text). Something like:

```go
idx := strings.Index(trimmed, noOutputStopMarker)
if idx < 0 {
    return false
}
return strings.HasSuffix(trimmed[idx+len(noOutputStopMarker):], noOutputStopSuffix)
```

Add a test case to `no_progress_stop_test.go` for the hostile-input sentence above.

#### 3. `maybeAutoTitleActiveSession` has no `sessionStore == nil` guard

**`internal/tui/session_title.go:243`** — checks `m.provider == nil` but not `m.sessionStore == nil`. The auto-title path is reachable for any model that ran a successful turn, including the fallback model constructed in some tests. If `m.sessionStore` is nil, the closure inside `generateSessionTitleCmd` would nil-dereference on the auto-title path (the backfill path happens to guard it because the closure re-reads events). The other nil-store fix in this PR (`resumeText`, `newSessionPicker`) set the precedent — this one was missed.

**Suggested fix**: add `if m.sessionStore == nil { return m, nil }` after the provider check. Add a small test (`TestMaybeAutoTitleSkipsWhenStoreNil`) parallel to the existing `TestResumeCommandReportsMissingSession` style.

#### 4. `TestResumeHonorsPriorCompaction` only compares slice lengths, not contents

**`internal/tui/model_test.go:946`** (and `:984` for the assertion) — already flagged by @anandh8x. The setup checks `len(rehydrated) >= len(raw)` and the assertion checks `len(next.sessionEvents) != len(rehydrated)`. A regression that produced a same-length but different-content slice (e.g. rehydration returns the raw events in a different order, or replaces the compaction summary event with one of the dropped originals) would pass this test.

**Suggested fix**: replace the length check with `if !reflect.DeepEqual(next.sessionEvents, rehydrated) { t.Fatalf(...) }` (or element-by-element comparison for a more readable failure message). Cheap, catches the actual bug class.

### Non-blocking notes (skip if you want)

- `generateSessionTitleCmd` closure captures `m.provider` and `m.sessionStore` at cmd-build time. If the user runs `/provider` between the successful turn and the title cmd landing, the title is generated by the **old** provider. Probably fine — same provider, different model — but worth a one-line comment if you want to be explicit.
- `UpdateTitle` (`internal/sessions/store.go:580`) trims the input but doesn't normalize internal whitespace, so a model return of `"Some  Title"` (two spaces) won't no-op against an existing `"Some Title"` (one space). Minor; would just rewrite the file with the new whitespace.

### What's good (no action needed)

- The picker structure mirrors `/model` and `/provider` cleanly.
- `sessionTitleDigest` correctly skips the no-output stop AND respects the digest char budget.
- `cleanGeneratedTitle` covers the realistic model failure modes (quotes, code fences, "Title:" labels, multi-line responses).
- `resumeEvents` rehydration-aware-with-fallback is the right call for the compaction correctness story.
- `sessionWhen` timestamp formatting (today → `15:04:05`, this year → `Jan _2 15:04`, else date) is exactly the right granularity.
- Per-session `titledSessions` gating at the type level (a map, not a counter) is the right shape for "at most once per session."
- All tests pass on `-count=1` and on CI. The "Zero Review" failure is the pre-existing `TestLoadOrCreateSecretConcurrentConverges` flake in `internal/oauth`, not this PR.

Happy to pair on any of these — they're all 5-15 line changes.

internal/agent/guardrails.go
```diff
@@ -90,11 +90,32 @@ func toolFailureStopAnswer(toolName string, count int) string {
...
+// IsNoProgressStop reports whether content IS the no-output guardrail stop answer
+// (a run that produced no visible text and no tool calls). It requires the full
+// prefix + marker + suffix structure — not just the marker substring — so a
+// genuine assistant/tool message that quotes the marker can't be misread as a
+// failed empty run (which would wrongly hide a real session from /resume and skip
+// its title generation).
+func IsNoProgressStop(content string) bool {
+	trimmed := strings.TrimSpace(content)
+	return strings.HasPrefix(trimmed, noOutputStopPrefix) &&
+		strings.Contains(trimmed, noOutputStopMarker) &&
```
Vasanthdev2004: The current `HasPrefix(Prefix) && Contains(Marker) && HasSuffix(Suffix)` check requires the three components in order with arbitrary text in between. A hostile real message like

> Agent stopped after 3 turns. The marker is "with no output (no visible text and no tool calls)" so here it is: to avoid consuming tokens without making progress.

is currently misclassified as a no-progress stop and would hide a real session from `/resume` and skip its title generation. I confirmed this in a quick repro outside the repo.

**Suggested fix:** after `HasPrefix` succeeds, find the marker index and require the suffix to come *immediately after* the marker (whitespace allowed, no arbitrary text in between):

```go
idx := strings.Index(trimmed, noOutputStopMarker)
if idx < 0 {
    return false
}
return strings.HasSuffix(trimmed[idx+len(noOutputStopMarker):], noOutputStopSuffix)
```

Add a test case to `no_progress_stop_test.go` for the hostile-input sentence above so the regression is caught.

internal/tui/model_test.go
```diff
@@ -822,16 +822,168 @@ func TestResumeCommandListsRecentSessions(t *testing.T) {
...
+	}
+	if len(rehydrated) >= len(raw) {
+		t.Fatalf("test setup invalid: rehydrated (%d) should be < raw (%d)", len(rehydrated), len(raw))
+	}
+
+	m := newModel(context.Background(), Options{SessionStore: store})
+	next, _ := m.handleResumeCommand(session.SessionID)
+	// Resume must load the rehydrated (compaction-aware) context, not the raw log —
+	// matching the CLI's --resume and the in-TUI /compact reload.
+	if len(next.sessionEvents) != len(rehydrated) {
```
Vasanthdev2004: Length-only comparison is a weak invariant: a regression that produces a same-length but different-content slice (e.g. rehydration returns the raw events in a different order, or replaces the compaction summary with one of the dropped originals) would pass this test. Replace `len(next.sessionEvents) != len(rehydrated)` with a deep comparison:

```go
if !reflect.DeepEqual(next.sessionEvents, rehydrated) {
    t.Fatalf("resumed sessionEvents do not match rehydrated; resume must honor prior compaction\nresumed:  %+v\nrehydrated: %+v\nraw:       %+v",
        next.sessionEvents, rehydrated, raw)
}
```

Cheap, catches the actual bug class, and makes the failure message actionable when it does fire.

Vasanthdev2004:
## Plan guide — revisions for `feat/resume-session-picker`

Posting a structured plan alongside the `REQUEST_CHANGES` review so the work can be scoped as one or two PRs without losing the durability story. The plan has **4 work-streams, each independently shippable** — Streams 1+2 (~70 LoC) can land in this PR; Stream 3 (~80 LoC) is the durable replacement for the in-memory `titledSessions` map and is recommended as a follow-up PR so this one stays focused.

---

### Stream 1 — Fix review findings (~40 LoC, ship in this PR)

**1.1 [high] Tighten `IsNoProgressStop`** — `internal/agent/guardrails.go:114-119`

Current check (`HasPrefix + Contains + HasSuffix`) allows arbitrary text between the three components. A real message that *quotes* the marker in passing gets misclassified. Replace the body with:

```go
func IsNoProgressStop(content string) bool {
    trimmed := strings.TrimSpace(content)
    if !strings.HasPrefix(trimmed, noOutputStopPrefix) {
        return false
    }
    idx := strings.Index(trimmed, noOutputStopMarker)
    if idx < 0 {
        return false
    }
    return strings.HasSuffix(trimmed[idx+len(noOutputStopMarker):], noOutputStopSuffix)
}
```

Add a case to `internal/agent/no_progress_stop_test.go` for the hostile-input sentence:

> Agent stopped after 3 turns. The marker is "with no output (no visible text and no tool calls)" so here it is: to avoid consuming tokens without making progress.

---

**1.2 [high] Move `titledSessions[id] = true` to the success path** — `internal/tui/session_title.go:255` and `:294`

The map is set *before* the title cmd fires, so a failed title generation leaves the session with its first-message title **forever for this process**. No log line, no retry, no escape other than restart. Drop the optimistic set. Move it to `handleSessionTitleGenerated` (line 307), gated on `msg.err == nil && msg.title != ""`:

```go
if msg.err == nil && msg.title != "" {
    if m.titledSessions == nil {
        m.titledSessions = map[string]bool{}
    }
    m.titledSessions[msg.sessionID] = true
}
```

`maybeAutoTitleActiveSession` becomes a pure "is it worth firing?" check. `startSessionRetitle` still pre-marks its first candidate only (not the whole queue — see Stream 3.2 for the full fix). Add a test to `session_title_test.go` that returns a `sessionTitleGeneratedMsg{err: someError}` and asserts the cmd fires again on a second call.

---

**1.3 [medium] Add `m.sessionStore == nil` guard** — `internal/tui/session_title.go:237-259`

Checks `m.provider == nil` but not `m.sessionStore == nil`. The closure inside `generateSessionTitleCmd` calls `store.UpdateTitle` and would nil-deref. Add the guard after the provider check:

```go
if m.sessionStore == nil {
    return m, nil
}
```

Add `TestMaybeAutoTitleSkipsWhenStoreNil` to `session_title_test.go`.

---

**1.4 [low] Strengthen `TestResumeHonorsPriorCompaction`** — `internal/tui/model_test.go:946`, assertion at `:984`

Replace the length-only check with a deep equality check. Add `reflect` to the imports:

```go
if !reflect.DeepEqual(next.sessionEvents, rehydrated) {
    t.Fatalf("resumed sessionEvents do not match rehydrated; resume must honor prior compaction\nresumed:  %+v\nrehydrated: %+v\nraw:       %+v",
        next.sessionEvents, rehydrated, raw)
}
```

---

### Stream 2 — Add the `isSessionTitleManuallySet` flag (~30 LoC, ship in this PR)

Currently no way for the user to say "this title is mine, leave it alone." If the user runs `/retitle` on a session, the next turn's auto-title will fire and overwrite it.

**2.1** — `internal/sessions/store.go`

Add to `Metadata` (line 65-95):

```go
TitleIsManuallySet bool `json:"titleIsManuallySet,omitempty"`
```

`omitempty` keeps existing session files backward compatible. Change `UpdateTitle` (line 579-606) to take an options struct and reject auto writes when the manual flag is set:

```go
type UpdateTitleInput struct {
    Title  string
    Manual bool
}

func (store *Store) UpdateTitle(sessionID string, input UpdateTitleInput) (Metadata, error) {
    // ... existing validation + lock + read ...
    if !input.Manual && session.TitleIsManuallySet {
        return session, nil // no-op
    }
    if session.Title == trimmed {
        return session, nil
    }
    session.Title = trimmed
    if input.Manual {
        session.TitleIsManuallySet = true
    }
    // ... existing writeMetadata ...
    return session, nil
}
```

Update the two call sites in `session_title.go:185` (auto-title, no manual flag) and the backfill path (also auto, no manual flag). The `/rename` command (separate PR) is what sets `Manual: true`.

**2.2** (deferred to a follow-up PR): add a `/rename <id> <new title>` slash command that calls `UpdateTitle{Manual: true}` and mark manually-titled sessions with a small `★` or `· locked` prefix in the picker. Not in scope for this PR.

---

### Stream 3 — Replace the in-memory `titledSessions` map with a persisted stage (~80 LoC, follow-up PR)

The current `titledSessions map[string]bool` is in-memory and process-local. It dies on restart; the auto-title path can re-fire on the same session across restarts. The trade-off:

- **In-memory (today)**: simple, fast, dies on restart
- **Persisted stage (proposed)**: durable, survives restart, matches the project's `SpecStatus` / `LastEventType` pattern

**3.1** — `internal/sessions/store.go`

Add a `TitleAutoStage` enum (matches the `SpecStatus` pattern at line 57-62):

```go
type TitleAutoStage string

const (
    TitleAutoStageNone         TitleAutoStage = ""
    TitleAutoStageFirstMessage TitleAutoStage = "first_message"
)

type Metadata struct {
    // ... existing fields ...
    TitleAutoStage TitleAutoStage `json:"titleAutoStage,omitempty"`
    // ...
}
```

**3.2** — `internal/sessions/store.go` (`UpdateTitle`)

Add server-side stage gate. The store enforces "auto writes only succeed if the session is still at the relevant stage":

```go
if !input.Manual {
    switch session.TitleAutoStage {
    case TitleAutoStageFirstMessage:
        return session, nil // already titled by the auto path
    }
}
// ... existing write ...
if !input.Manual && session.TitleAutoStage < TitleAutoStageFirstMessage {
    session.TitleAutoStage = TitleAutoStageFirstMessage
}
```

**3.3** — `internal/tui/model.go` (line 80) and `internal/tui/session_title.go` (lines 237, 262, 290, 307)

Remove the `titledSessions` field from `model`. `maybeAutoTitleActiveSession` becomes a thin "should we fire?" check, no side effect. `startSessionRetitle` no longer pre-marks the map. `handleSessionTitleGenerated` no longer writes the map — the store handles the stage advance server-side.

**Bonus behavior change**: failed backfill steps don't lose their turn. If the provider errors on session 2 of 5, the next `/retitle` run picks it up because the server-side stage is still `""` and `sessionTitleIsAuto` is still true.

**3.4** — Tests

Add to `internal/sessions/session_title_test.go`:
- `TestUpdateTitleRejectsAutoWhenManual`
- `TestUpdateTitleAdvancesAutoStage`
- `TestUpdateTitleManualIgnoresStage`

Add to `internal/tui/session_title_test.go`:
- `TestAutoTitleRetriesAfterFailure` — return `sessionTitleGeneratedMsg{err: ...}`, assert cmd fires again on a second call
- `TestBackfillRetriesAfterFailure` — same shape for the backfill path

---

### Stream 4 — Prompt + cleaner cleanup (optional, 5 LoC)

**4.1** — `internal/tui/session_title.go:19-21`

The 30-line `cleanGeneratedTitle` has a "drop a leading `Title:` / `Title -` label" branch because the system prompt tells the model to write `Title Case` — the model often interprets that as "write a line that starts with `Title:`". Either tighten the prompt or keep both. Recommendation: keep the cleaner (10 lines, good test coverage, future-proof) and tighten the prompt as defense-in-depth:

```go
const sessionTitleSystemPrompt = "You write a short, specific title for a coding-assistant conversation so a user can tell it apart from others in a list. " +
    "Reply with ONLY the title in Title Case, starting directly with the title text — no label, no preamble, no explanation. " +
    "3 to 6 words, naming the concrete task or topic."
```

---

### Suggested merge order

1. Stream 1.1 (hotfix, 4 LoC)
2. Stream 1.2 + 1.3 + 1.4 (semantic changes, 30 LoC + tests)
3. Stream 2.1 (manual flag, signature change, 30 LoC)
4. Stream 3.1 + 3.2 + 3.3 + 3.4 (durable stage, separate PR, 80 LoC + tests)
5. Stream 4.1 (prompt tweak, independent, 5 LoC)

Total: ~150 LoC across two PRs.

---

### What NOT to do

- **Don't keep the in-memory `titledSessions` map and add the server-side stage.** Two state sources for the same invariant drift over time. Pick one — the proposal is server-side only.
- **Don't add the `Title:` label handling to the prompt and remove the cleaner in the same commit.** Fix one or the other, not both.
- **Don't introduce a sentinel value in the title field itself** (e.g. `Title = "@auto"`). The project uses dedicated fields for this kind of state (`SpecStatus`, `LastEventType`).
- **Don't add a `/rename` slash command in this PR.** It's the natural next step after the manual flag is persisted, but the picker UX work is its own scope.

---

### Verification checklist (before merge)

- [ ] `go build ./...` clean (stash `.claude/wip-stash/` first per the build script)
- [ ] `go vet ./internal/...` clean
- [ ] `gofmt -l ./...` clean
- [ ] `go test -count=1 -race ./internal/sessions/... ./internal/tui/...` green (CGO required for `-race`)
- [ ] All 6 smoke checks (ubuntu / macos / windows) green on CI
- [ ] Pre-existing flake `TestLoadOrCreateSecretConcurrentConverges` (internal/oauth) may surface on Zero Review; re-run to clear
- [ ] Pre-existing env-fragile test from PR #220 (`TestModelCommandPersistsSelectedModelToUserConfig`) is unrelated to this PR; verify failure (if any) is from the `ANTHROPIC_*` env vars, not from these changes

---

Full plan guide with all the code blocks in context: see the discussion comment on this PR.

gnanam1990:
Thanks for the thorough review @Vasanthdev2004 — all four addressed in `2d4d127`.

**1. `titledSessions` set optimistically (silent, permanent failure).** Fixed — but I kept the *optimistic* mark and instead **release it on failure** in `handleSessionTitleGenerated` (provider error / empty title / store-write error → `delete(titledSessions, id)`), rather than moving the mark to the success path. Reason: marking only on success reopens a double-fire window — `maybeAutoTitleActiveSession` runs after every turn, so a second turn finishing while the first title cmd is still in flight (title not applied yet, so the session still reads as "auto") would kick off a *second* generation for the same session. Unmark-on-failure keeps your retryability goal **and** the no-double-fire guarantee; success still keeps the one-shot gate. Covered by `TestAutoTitleFailureReleasesRetryGate`.

**2. `IsNoProgressStop` structurally loose.** Fixed — it now matches the exact structure `noOutputStopAnswer` emits: `prefix + "<int> turns " + marker + " " + suffix`, requiring the count to be a bare integer and the marker to be immediately followed by the suffix. Your hostile sentence (plus "text between marker and suffix" and "non-integer count") are now in `no_progress_stop_test.go` and rejected.

**3. `maybeAutoTitleActiveSession` missing nil-store guard.** Fixed — `if m.provider == nil || m.sessionStore == nil { return m, nil }`, with `TestMaybeAutoTitleSkipsWhenStoreNil`.

**4. Length-only compaction assertion.** Fixed — `TestResumeHonorsPriorCompaction` now uses `reflect.DeepEqual(next.sessionEvents, rehydrated)` and dumps raw/rehydrated/resumed on failure. (This also covers @anandh8x's flag of the same test.)

Non-blocking notes left as-is: the `generateSessionTitleCmd` provider capture is benign (same provider — only the model could differ between the turn and the title cmd), and the `UpdateTitle` internal-whitespace normalization is cosmetic. Happy to fold in either if you'd prefer.

Gate: `gofmt`, `go vet`, `go test ./internal/agent ./internal/tui -race`, and staticcheck all green.

Vasanthdev2004 (commented):
LGTM ? no issues found. UpdateTitle correctly serializes under the per-session lock, re-reads latest metadata before writing, leaves UpdatedAt untouched, rejects blank/unchanged titles, and the auto-title path fires at most once per session. The resume picker hydrates via handleResumeCommand and correctly prefers the compaction-aware event view.

Vasanthdev2004 (approved):
LGTM ? no issues found. UpdateTitle correctly serializes under the per-session lock, re-reads latest metadata before writing, leaves UpdatedAt untouched, rejects blank/unchanged titles, and the auto-title path fires at most once per session. The resume picker hydrates via handleResumeCommand and correctly prefers the compaction-aware event view.

internal/agent/guardrails.go
gnanam1990: Addressed on the current branch (this thread was on an earlier commit). `IsNoProgressStop` now requires the structured form exactly: after the `Agent stopped after ` prefix it parses the turn count with `strconv.Atoi`, then requires the remainder to **equal** `marker + " " + suffix` (`guardrails.go:116-135`). That exact match is stricter than `Index`+`HasSuffix` — it also rejects arbitrary text wedged *before* the marker. Your hostile-input sentence is locked in as a rejected case in `no_progress_stop_test.go:25`, alongside wedged-text and non-integer-count variants. `go test ./internal/agent/ -run TestIsNoProgressStop` passes.

internal/tui/model_test.go
gnanam1990: Addressed — the length-only check is now `reflect.DeepEqual(next.sessionEvents, rehydrated)` in `TestResumeHonorsPriorCompaction` (`model_test.go:988`), with a failure message dumping resumed/rehydrated/raw so a same-length/different-content regression fails loudly. `go test ./internal/tui/ -run TestResumeHonorsPriorCompaction` passes.

Vasanthdev2004 (approved):
Looked this over carefully ??? it's a solid PR. The interaction with `choosePicker` for the new `pickerSession` kind is wired in cleanly, and `openSessionPicker` does the right thing by short-circuiting to the text fallback when there's nothing resumable. `handleResumeCommand` -> `resumeEvents` preferring the rehydrated (compaction-aware) view with a raw-read fallback is the right move, and the summary line showing the active vs recorded model/provider is a small but very useful touch ??? it'll save people a lot of "wait, why did it switch?" moments.

The title work is where I was most skeptical, and it's well-handled:

- `Store.UpdateTitle` is locked under the same per-session mutex as `AppendEvent` and re-reads metadata under that lock, so it can't race with a concurrent append or vice-versa. Leaving `UpdatedAt` alone on a retitle is the right call ??? a title isn't activity, and a retitle should never reorder the picker.
- The `titledSessions` gate is set optimistically when the cmd is scheduled and released on failure, so a hung/failed provider doesn't permanently lock a session out of ever getting a model-generated title. Auto-title is silent on failure (first-message title stays), which is the right UX.
- `cleanGeneratedTitle` is defensive in the right ways: first-non-empty-line after markup strip, word cap, rune cap, and `sessionTitleTimeout` keeps a stuck provider from wedging the background cmd.
- The `IsNoProgressStop` structural match (prefix + `<int> turns ` + marker + suffix) is correct ??? a model quoting the marker amid other prose won't be misclassified as an empty run and wrongly hidden from the picker.

`session_title_test.go` covers digest, cleaning, auto-title one-shot, and `/retitle` candidate selection; the test in the sessions package covers the lock/no-blank/no-op semantics. `resume_nil_store_test.go` covers the nil-store paths.

I also confirmed `gofmt`/`go vet` clean and the three affected packages (`internal/sessions`, `internal/agent`, `internal/tui`) pass under `go test -count=1`.

LGTM. Two small suggestions for follow-up, not blockers:

1. The title generation runs serially in `/retitle` (one at a time via `m.retitleQueue`). That's the right call to avoid burst-charging, but the per-session `sessionTitleTimeout` is 30s ??? a run with N sessions could take N*30s in the worst case. Worth surfacing a cancel command in `/retitle` so a user can bail if it's taking too long.
2. `m.titledSessions` lives on the model. If the TUI restarts (process restart) the gate resets, so a session that was mid-title at exit can get re-titled on the next turn. Probably fine in practice (the model will just re-generate the same title), but worth noting in case you want a persistent `titledAt` marker on the metadata.

Nice work on this one ??? the picker UX is much better than the old text blob, and the model-generated titles make `/resume` actually usable for finding past work.

Vasanthdev2004 (approved):
Looked this over carefully ??? it's a solid PR. The interaction with `choosePicker` for the new `pickerSession` kind is wired in cleanly, and `openSessionPicker` does the right thing by short-circuiting to the text fallback when there's nothing resumable. `handleResumeCommand` -> `resumeEvents` preferring the rehydrated (compaction-aware) view with a raw-read fallback is the right move, and the summary line showing the active vs recorded model/provider is a small but very useful touch ??? it'll save people a lot of "wait, why did it switch?" moments.

The title work is where I was most skeptical, and it's well-handled:

- `Store.UpdateTitle` is locked under the same per-session mutex as `AppendEvent` and re-reads metadata under that lock, so it can't race with a concurrent append or vice-versa. Leaving `UpdatedAt` alone on a retitle is the right call ??? a title isn't activity, and a retitle should never reorder the picker.
- The `titledSessions` gate is set optimistically when the cmd is scheduled and released on failure, so a hung/failed provider doesn't permanently lock a session out of ever getting a model-generated title. Auto-title is silent on failure (first-message title stays), which is the right UX.
- `cleanGeneratedTitle` is defensive in the right ways: first-non-empty-line after markup strip, word cap, rune cap, and `sessionTitleTimeout` keeps a stuck provider from wedging the background cmd.
- The `IsNoProgressStop` structural match (prefix + `<int> turns ` + marker + suffix) is correct ??? a model quoting the marker amid other prose won't be misclassified as an empty run and wrongly hidden from the picker.

`session_title_test.go` covers digest, cleaning, auto-title one-shot, and `/retitle` candidate selection; the test in the sessions package covers the lock/no-blank/no-op semantics. `resume_nil_store_test.go` covers the nil-store paths.

I also confirmed `gofmt`/`go vet` clean and the three affected packages (`internal/sessions`, `internal/agent`, `internal/tui`) pass under `go test -count=1`.

LGTM. Two small suggestions for follow-up, not blockers:

1. The title generation runs serially in `/retitle` (one at a time via `m.retitleQueue`). That's the right call to avoid burst-charging, but the per-session `sessionTitleTimeout` is 30s ??? a run with N sessions could take N*30s in the worst case. Worth surfacing a cancel command in `/retitle` so a user can bail if it's taking too long.
2. `m.titledSessions` lives on the model. If the TUI restarts (process restart) the gate resets, so a session that was mid-title at exit can get re-titled on the next turn. Probably fine in practice (the model will just re-generate the same title), but worth noting in case you want a persistent `titledAt` marker on the metadata.

Nice work on this one ??? the picker UX is much better than the old text blob, and the model-generated titles make `/resume` actually usable for finding past work.

PR #225, opened 2026-06-16
Fix provider setup paste handling

anandh8x:
## Summary
- route terminal paste events to the active provider setup field
- support pasted endpoint URLs in first-run setup and /provider wizard
- add regression coverage using PasteMsg for endpoint URLs and API keys

## Tests
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./internal/tui
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./...

## Release Notes

* **Bug Fixes**
  * Enhanced paste input handling during onboarding and provider setup to correctly route pasted content to appropriate input fields based on UI context.

* **Tests**
  * Added tests verifying paste functionality for URLs and API keys in onboarding and provider configuration workflows.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -561,12 +561,18 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 	case providerWizardDeviceCodeMsg:
 		return m.applyProviderWizardDeviceCode(msg)
 	case tea.PasteMsg:
-		if m.setup.visible || m.pendingAskUser != nil {
+		if m.setup.visible {
+			return m.handleSetupPaste(msg)
+		}
+		if m.pendingAskUser != nil {
 			var cmd tea.Cmd
 			m.input, cmd = m.input.Update(msg)
 			return m, cmd
 		}
-		if m.transcriptDetailed || m.pendingSpecReview != nil || m.pendingPermission != nil || m.providerWizard != nil || m.mcpAddWizard != nil || m.mcpManager != nil || m.picker != nil {
+		if m.providerWizard != nil {
+			return m.handleProviderWizardPaste(msg.Content)
+		}
+		if m.transcriptDetailed || m.pendingSpecReview != nil || m.pendingPermission != nil || m.mcpAddWizard != nil || m.mcpManager != nil || m.picker != nil {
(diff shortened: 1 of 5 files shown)


anandh8x (commented):
No code issues found in this diff.

The PasteMsg routing is scoped to the active setup/provider wizard fields and keeps the existing modal guards in place. The added tests cover real PasteMsg handling for endpoint URLs and API keys.

Local verification already passed on this branch:

- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./internal/tui
- GOCACHE=/tmp/zero-go-cache GOMODCACHE=/tmp/zero-go-mod-cache go test ./...

gnanam1990 (approved):
**Approving** — focused, correct, and green on all platforms. Routing paste through the same per-stage/per-step append helpers the keystroke handlers already use is the right model: paste and typing now behave identically in both setup and the /provider wizard. Good catch that the old setup branch routed to the shared input rather than the stage-specific field buffers — that was the actual bug.

A few non-blocking notes:

- **MCP add-wizard paste is still dropped.** The handler keeps `m.mcpAddWizard != nil` in the no-op list, so pasting a server URL / command / header into the MCP wizard still doesn't land. Out of scope for "provider setup," but worth a follow-up so paste is consistent across the wizards.
- **Credential sanitation differs by surface:** the wizard credential uses `appendAPIKey` (strips control/space, so a trailing newline is dropped), while the setup credential goes through `m.setup.apiKey.Update(msg)`. Bubbles' textinput usually sanitizes paste to a single line, so this is probably fine — just flagging the asymmetry in case a key pasted with a trailing newline behaves differently between the two paths.
- Test quality is great: converting the existing `testKeyText(secret)` cases to `testPaste(secret)` is exactly right — they were claiming to cover paste while actually exercising the typed path.

Heads-up on merge ordering (not a blocker): this touches the `PasteMsg` handler in `model.go` and `provider_wizard.go`, which other in-flight TUI changes also touch — so whatever lands after this will need a quick rebase.

LGTM 👍

PR #226, opened 2026-06-16
Harden TUI layout and mouse routing

anandh8x:
## Summary
- centralize chat frame geometry for header, transcript body, footer, composer, and status regions
- route composer, overlay, and transcript mouse hit testing through shared region geometry
- preserve current transcript rendering and bottom-anchored scroll behavior for the first slice of TUI hardening

## Notes
This is the single draft PR for the TUI backend hardening work. Follow-up commits for the transcript viewport/cache work should stack onto this branch instead of opening separate PRs.

## Tests
- /bin/bash -lc "GOCACHE=/tmp/zero-go-cache go test ./..."

## Summary by CodeRabbit

* **Refactor**
  * Updated alt-screen transcript UI to use geometry-aware framing and viewport-based scrolling for more accurate header/body/footer behavior.
  * Reworked transcript body rendering into structured line items to improve selection and overlay rendering consistency.
  * Centralized transcript scroll/visible-window calculations to clamp offsets reliably.
* **Bug Fixes**
  * Improved mouse hit detection for transcript lines and overlay interactions, including correct region mapping and tighter click clamping.
  * Adjusted composer/status hit targeting to match visible areas.
* **Tests**
  * Expanded unit and interaction tests, including tiny/degenerate terminal dimensions and overlay/selection mouse scenarios.

### DIFF
--- a/internal/tui/layout_test.go
+++ b/internal/tui/layout_test.go
@@ -1,0 +1,133 @@
+package tui
+
+import (
+	"strings"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+)
+
+func TestTranscriptFrameLayoutPinsMainRegions(t *testing.T) {
+	m := mouseTestModel()
+	m.width = 100
+	m.height = 24
+	m.providerName = "openai"
+	m.modelName = "gpt-4.1"
+
+	width := chatWidth(m.width)
+	frame := m.scrollableTranscriptFrame(m.pinnedTitleBar(width), m.footerView(width))
+
+	if frame.headerRect != (tuiRect{width: width, height: len(frame.headerLines)}) {
+		t.Fatalf("header rect = %#v, want y=0 width=%d height=%d", frame.headerRect, width, len(frame.headerLines))
+	}
+	if frame.bodyRect.y != frame.headerRect.height {
+		t.Fatalf("body y = %d, want below header height %d", frame.bodyRect.y, frame.headerRect.height)
+	}
+	if frame.footerRect.y != frame.bodyRect.y+frame.bodyRect.height {
+		t.Fatalf("footer y = %d, want after body at %d", frame.footerRect.y, frame.bodyRect.y+frame.bodyRect.height)
+	}
+	if frame.footerRect.y+frame.footerRect.height != m.height {
+		t.Fatalf("footer bottom = %d, want terminal height %d", frame.footerRect.y+frame.footerRect.height, m.height)
+	}
+	if frame.composerRect.height <= 0 || !frame.footerRect.contains(frame.composerRect.x, frame.composerRect.y) {
+		t.Fatalf("composer rect should be visible inside footer, frame=%#v", frame)
+	}
+	if frame.statusRect.height != 1 || frame.statusRect.y != frame.footerRect.y+frame.footerRect.height-1 {
+		t.Fatalf("status rect = %#v, want final visible footer line", frame.statusRect)
+	}
+}
+
+func TestTranscriptFrameLayoutClipsFooterInTinyTerminal(t *testing.T) {
+	m := mouseTestModel()
+	m.width = 44
+	m.height = 3
+	m.copyStatus = "Copied!"
+	m.input.SetValue("Create a book library dashboard page with cards, filters, charts, and responsive behavior.")
+
+	width := chatWidth(m.width)
+	frame := m.scrollableTranscriptFrame(m.pinnedTitleBar(width), m.footerView(width))
+
+	if frame.bodyRect.height != 1 {
+		t.Fatalf("body height = %d, want one protected viewport line", frame.bodyRect.height)
+	}
+	if frame.footerRect.height != m.height-1 {
+		t.Fatalf("footer height = %d, want clipped to %d", frame.footerRect.height, m.height-1)
+	}
+	if frame.headerRect.height != 0 {
+		t.Fatalf("header should clip out before body disappears, got %#v", frame.headerRect)
+	}
+	if frame.footerClip <= 0 {
+		t.Fatalf("footerClip = %d, want clipped full footer", frame.footerClip)
+	}
+	if len(frame.footerLines) >= len(frame.fullFooterLines) {
+		t.Fatalf("footer lines were not clipped: visible=%d full=%d", len(frame.footerLines), len(frame.fullFooterLines))
+	}
+}
+
+func TestFrameComposerRegionDrivesMouseHit(t *testing.T) {
+	m := mouseTestModel()
+	m.width = 44
+	m.height = 20
+	m.input.SetValue("Create a book library dashboard page with cards, filters, charts, and responsive behavior.")
+
+	width := chatWidth(m.width)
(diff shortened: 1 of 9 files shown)


gnanam1990 (commented):
Strong, well-structured hardening pass. The `tuiRect` geometry + region rects (header / body / footer / composer / status) replace a lot of ad-hoc offset math with one model, and the coverage — degenerate 0/negative dims, tiny-terminal clipping, footer clip, overlay clamping, frame-body click mapping — is exactly what this kind of layout code needs.

Highlights:
- `mouseOverComposer` collapsing to `frame.composerRect.contains(x, y)` is a big readability win over the old manual clip math.
- `overlayMouseRect` clamping the overlay to the **visible** body height (with the test that a click below the visible rows misses) fixes a real class of off-screen-overlay hit bugs.
- `layoutTranscriptBodyItems` with typed spans (empty / row / separator / pending-interim) and selectable lines shifted by item start is a much cleaner transcript-body model than threading `bodyY` through the render funcs.

A few things:

- **The coordination footprint is large.** This rewrites the exact geometry/mouse surface (`scrollableTranscriptFrame`, `overlayMouseTop`→`overlayMouseRect`, `transcriptViewportStart`, `mouseOverComposer`, `transcriptLineAtMouse`) that several in-flight changes also touch — #224 and a couple of other TUI changes (a clickable/selectable permission popup, and a right-docked panel that reserves horizontal space from the chat area). Whatever lands first forces a real rebase on the rest, so it's worth sequencing this deliberately. One concrete hook: anything that reserves width from the chat (a side panel) will want the frame rects' `width` to come from a shared "content width" helper rather than `chatWidth(m.width)` directly — a natural follow-up on this geometry.
- Since this is the draft foundation with follow-ups stacking onto the branch, keeping each stacked slice independently green (build + the touched tests) will make the eventual review of the stack much easier.
- Minor: `mouseOverlayHit` now returns `x` — is a current caller using it, or is it staged for the follow-ups? If unused for now, a one-line note saves a future "why is this here?".

Solid base, and nicely tested. 👍

Vasanthdev2004 (approved):
## Verdict: **Approve** — clean hardening pass with addressed review findings

This is a strong refactor that replaces a lot of ad-hoc Y-offset math with a typed, region-based geometry model. The two open CodeRabbit findings are already fixed in the current `0f2307f` tip; @gnanam1990's human review was positive with one operational note and one unused-field note. No blockers.

### What's good

- **`tuiRect` + `transcriptFrameLayout`** — `headerRect`, `bodyRect`, `footerRect`, `composerRect`, `statusRect` are explicit rectangles with `contains`/`local` helpers. The frame computation in `scrollableTranscriptFrame` (`model.go:1345-1392`) lays them all out and the assertions in `layout_test.go` lock the invariant (no gaps, footer bottom == terminal height, status is the last visible footer line).
- **`mouseOverComposer` collapsed to one line** — the old code did its own footer clip + top + `mouseY` math; the new code is `frame.composerRect.contains(mouseX(msg), mouseY(msg))`. Big readability win.
- **`overlayMouseRect` clamps to visible body** — `mouse.go:464-471` clamps the overlay to `frame.bodyRect.height`, so a click below the visible rows correctly misses. Fixes a real off-screen-overlay hit bug class.
- **`transcriptBodyLayout` is the right shape** — `lines` + `selectable` + typed `spans` (`title`/`empty`/`separator`/`row`/`pending-prompt`/`pending-interim`/`spec-review`). The body is no longer a string threaded with `bodyY` offsets; it's a structured layout the viewport can window over (`transcript_viewport.go`) and the mouse can hit-test against.
- **`transcriptViewport` is a pure value type** — `transcript_viewport.go` is 57 LoC, no model dependency, table-driven tests for zero/negative inputs and clamp behavior. Easy to reason about.
- **CodeRabbit #1 already fixed** — `layout_test.go:34` now contains the full composer-rect containment check (top-left AND bottom-right corners inside `footerRect`). The "Addressed in commit f3dfbfb" stamp on the inline comment confirms it.
- **CodeRabbit #2 already fixed** — `mouse_test.go:491` now uses `textX, textY := firstTranscriptTextMousePoint(t, m)` (line 562), which returns the actual `textStart` column, not `x=0`. The new test `TestMCPAddWizardBlocksTranscriptSelectionBehindOverlay` clicks at the real selectable position.
- **Test coverage is strong** — `layout_test.go` (185 LoC, 4 tests), `transcript_viewport_test.go` (85 LoC, 4 tests including a table-driven edge case pass), `transcript_body_test.go` (86 LoC, 3 tests), `render_cache_test.go` (20 LoC, 1 new test). Each commit on the branch added 30-100 LoC of tests.

### Non-blocking notes (skip if you want)

1. **`mouseOverlayHit.x` is added but unused** (`mouse.go:5`, set at line 452). The struct is constructed and returned, but every caller reads only `hit.y` (verified: `selectSuggestionAtMouse`, `selectMCPManagerAtMouse`, `selectMCPAddWizardAtMouse`, `selectModelPickerAtMouse`, `selectProviderWizardAtMouse`, etc.). @gnanam1990 flagged this; agreed. Two options:
    - **Drop the field now** (1-line change in the struct + 1-line change at the return site). Keeps the surface area tight.
    - **Leave it with a one-line `// Reserved for the clickable permission popup / right-docked panel follow-ups`** comment** so the next reader knows it's staged, not stale.

    I lean toward **option 2** since the field's name and return-site comment in the diff suggest it's deliberate staging. If you don't have a follow-up PR that uses it, drop it. Otherwise comment it.
2. **PR coordination footgun** (raised by @gnanam1990) — this PR rewrites the exact geometry/mouse surface that #224 (resume picker), a planned clickable permission popup, and a planned right-docked side panel all touch. The right-docked panel is the one that would change `chatWidth(m.width)` — anything that reserves width from the chat will want the frame rects' `width` to come from a shared "content width" helper, not `chatWidth(m.width)` directly. If the side-panel PR is in flight, **this PR should land first** so the side panel stacks on top of the rect-based model. If the panel PR lands first, the rebase will be mechanical but tedious. Worth a quick sync with whoever's working on the panel.
3. **Inlined `chatWidth(m.width)`** in the frame construction (`model.go:1380`) and `overlayMouseRect` (`mouse.go:471`). A future side-panel refactor that takes a content-width helper will want to thread it through here. Not a blocker for this PR; just worth a `// TODO: thread content width here when the right-docked panel lands` for the next reader.

### CI

All 6 smoke checks green (ubuntu / macos / windows / performance / security / CodeRabbit). "Zero Review" re-ran and is `IN_PROGRESS` as of this review; expect it to clear. Local `./internal/tui/...` is green except for the pre-existing env-fragile `TestModelCommandPersistsSelectedModelToUserConfig` from PR #220 (fails only in dev envs with `ANTHROPIC_*` set, not this PR's fault).

### Recommended follow-ups (separate PRs, NOT this one)

- The clickable permission popup and the right-docked side panel that @gnanam1990 mentions. Both benefit from the rect-based model this PR introduces; both can land as independent stacks on this branch as the PR description suggests.
- A `TestTranscriptFrameLayoutHandlesTransitions` test (tiny terminal ↔ full terminal) to lock in the clip math. The current tests cover tiny and full separately; an explicit transition test would be one more tripwire.

Solid foundation for the TUI backend hardening sequence. ✅

PR #227, opened 2026-06-16
Render TUI transcript viewport from measured items

anandh8x:
## Summary
- add a bounded transcript body height cache keyed by existing row render fingerprints
- render alt-screen transcript windows from measured item spans instead of building the whole body every frame
- route transcript mouse hit testing through the same visible item layout
- cover cache reuse, visible-only rendering, absolute selectable coordinates, and full-layout parity

## Tests
- git diff --check
- /bin/bash -lc "GOCACHE=/tmp/zero-go-cache go test ./..."

## Notes
This is the follow-up to #226 for the remaining useful backend hardening in the transcript viewport/rendering path. It avoids UI/design changes and keeps the existing full-layout path where it is still needed for inline rendering and selection copy.

## Summary by CodeRabbit

* **Refactor**
  * Optimized transcript rendering performance through improved caching and measurement strategies for better responsiveness when displaying conversation history.

* **Tests**
  * Added comprehensive test coverage for transcript body rendering, viewport calculations, and selection behavior to ensure rendering accuracy and consistency.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -87,6 +87,7 @@ type model struct {
 	unpricedTokens         int
 	transcript             []transcriptRow
 	transcriptDetailed     bool
+	transcriptBodyHeights  *transcriptBodyHeightCache
 	input                  textinput.Model
 	composer               composerState
 	composerActive         bool
@@ -439,6 +440,7 @@ func newModel(ctx context.Context, options Options) model {
 		userAgent:              options.UserAgent,
 		usageTracker:           usageTracker,
 		transcript:             initialTranscript(),
+		transcriptBodyHeights:  newTranscriptBodyHeightCache(defaultTranscriptBodyHeightCacheMaxEntries),
 		prService:              prService,
 		prState:                prService.GetState(),
 		input:                  input,
@@ -1283,7 +1285,7 @@ func (m model) transcriptView() string {
 	if m.transcriptEmpty() && !m.pending && viewportOverlay != "" {
 		emptyOverlay = viewportOverlay
 	}
-	bodyLayout := m.transcriptBodyLayout(width, emptyOverlay)
+	bodyItems := m.transcriptBodyItems(width, emptyOverlay)
 
 	footer := m.footerView(width)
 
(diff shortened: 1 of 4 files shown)


gnanam1990 (approved):
Reviewed — a clean, well-targeted perf refactor. Approving; one merge-coordination note below (not a blocker for this PR on its own).

## What it does
Splits transcript rendering into measure → window → render-visible: a bounded LRU height cache keyed by row render fingerprints, `measureTranscriptBodyItems` for spans, and `layoutVisibleTranscriptBodyItems` that renders only the items intersecting the viewport. Mouse hit-testing routes through the same window.

## Strengths
- `TestScrollableTranscriptItemsViewMatchesFullLayout` (byte-identical parity vs the full-layout path) is exactly the right way to de-risk a rendering refactor — it's the test that makes this safe to land.
- Height is independent of `startBodyY`, so measuring at 0 and rendering at `startY` agree; the `clampInt` visible-slice math and absolute-`bodyY` selectable filtering are sound (and covered by the new tests).
- LRU is bounded + evicted (no leak); the shared `*cache` across model value-copies is intentional and mutex-guarded.

## Merge-coordination note (heads-up, not a blocker here)
Full disclosure — I have #228 open, which touches the same render-cache plumbing (it adds a `compactEdit` field to `cardRenderOptions` + `renderRowCacheKey`). Your `transcriptRowBodyHeightCacheKey` reuses `renderRowCacheKey` but builds `opts` without that field. On current `main` this is correct; once both land, whoever merges second should:
- re-run the parity + `compactEdit` tests, and
- ideally pass the *same* `opts` the live render uses, so the height key reflects `compactEdit` explicitly rather than relying on #228's "active edit rows are `stable=false`" rule (which happens to keep those heights re-measured live — likely fine, just worth making explicit).

## Minor (non-blocking)
- A *visible unstable* row renders twice per frame (measure + layout) — fine, it's a net win vs render-everything, just noting.
- `scrollableTranscriptLayoutView` is now reachable only from the parity test — fine to keep as the reference path.

Nice work.

PR #229, opened 2026-06-16
feat(tui): Ctrl+T cycle + divider display for reasoning effort

Vasanthdev2004:
-

## Release Notes

* **New Features**
  * Added **Ctrl+T** to cycle the active **reasoning effort** (**auto → low → medium → high → auto**), applied on the **next turn** (inactive with detailed transcripts or when a blocking modal/prompt is open).
* **UI Improvements**
  * Updated **/effort** to show richer command-card confirmations/status for auto, invalid, and unsupported selections.
  * Composer divider now reflects the current **reasoning effort** (hidden for **auto**), and adds a one-line description hint for a single unambiguous **/command** suggestion.
* **Bug Fixes**
  * **Ctrl+T** is a no-op when the active model has no effort controls; **/effort** picker always includes **auto**.
* **Tests**
  * Added coverage for effort cycling, divider rendering, and description hint behavior.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1348,6 +1348,10 @@ func (m model) footerView(width int) string {
 		footer.WriteString("\n")
 	}
 	footer.WriteString(m.composerBox(width))
+	if hint := m.composerDescriptionHint(width); hint != "" {
+		footer.WriteString("\n")
+		footer.WriteString(hint)
+	}
 	if queued := renderQueuedMessagePreview(m.queuedMessage, width); queued != "" {
 		footer.WriteString("\n")
 		footer.WriteString(queued)
@@ -1937,6 +1941,37 @@ func (m model) composerBox(width int) string {
 	}
 	rendered = append(rendered, m.composerDividerLine(width))
 	return strings.Join(rendered, "\n")
+}
+
+// composerDescriptionHint returns the description line that sits below the
+// composer box, claude-code style, when the input is a single unambiguous
+// slash command. Returns "" when the user is mid-prompt, the palette is closed,
+// or more than one command matches. Slash commands only; the @file palette
+// already shows its rows. The inline argument hint ([low|medium|...]) is
+// unchanged and continues to render inside the composer box.
+func (m model) composerDescriptionHint(width int) string {
+	if width < 8 {
+		return ""
+	}
+	if m.suggestionsAreFiles {
+		return ""
+	}
+	if !m.commandPaletteOpen || len(m.suggestions) != 1 {
+		return ""
+	}
+	if m.suggestionIdx != 0 {
+		return ""
+	}
+	value := strings.TrimSpace(m.input.Value())
+	if !strings.HasPrefix(value, "/") || strings.ContainsAny(value, " \t\n") {
+		return ""
+	}
+	suggestion := m.suggestions[0]
+	desc := strings.TrimSpace(suggestion.Desc)
+	if desc == "" {
+		return ""
+	}
+	return fitStyledLine(zeroTheme.muted.Render(desc), width)
 }
 
 // startsTurn reports whether a row begins a new conversational turn and therefore
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -740,6 +740,16 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				m.permissionMode = nextPermissionMode(m.permissionMode)
 				return m, nil
 			}
+		case keyCtrl(msg, 't'):
+			// Ctrl+T cycles reasoning effort opencode-style (auto -> low ->
+			// medium -> high -> auto), but only when nothing modal is up — the
+			// same gate shift+tab uses above. Not gated on m.pending: cycling
+			// mid-run is allowed and takes effect on the next turn, matching
+			// /effort. cycleReasoningEffort is a silent no-op on models with no
+			// effort controls.
+			if m.pendingPermission == nil && m.pendingAskUser == nil && m.pendingSpecReview == nil && m.providerWizard == nil && m.mcpAddWizard == nil && m.mcpManager == nil && m.picker == nil {
+				return m.cycleReasoningEffort()
+			}
 		case keyCtrl(msg, 'f'):
 			if m.picker != nil && m.picker.kind == pickerModel {
 				if m.modelPickerIsLoading() {
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -741,6 +741,9 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, nil
 			}
 		case keyCtrl(msg, 't'):
+			if m.transcriptDetailed {
+				return m, nil
+			}
 			// Ctrl+T cycles reasoning effort opencode-style (auto -> low ->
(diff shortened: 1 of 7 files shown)


gnanam1990 (commented):
## Review — Ctrl+T reasoning-effort cycle + divider

Nice, self-contained feature. It adds a Ctrl+T shortcut that walks the active model's supported reasoning-effort ring (auto → low → medium → high → auto) and surfaces the current value in the composer divider. Builds clean and the new tests pass locally (`go test ./internal/tui/ -run 'CycleReasoningEffort|ComposerDivider'`).

### Strengths
- `cycleReasoningEffort` is a tidy ring-walk and the test table covers every branch — auto-start, mid-ring advance, last-slot wrap, an unknown/unsupported effort resetting to auto, and a true no-op on a model with no effort controls.
- Good awareness of the render hot path: both `cycleReasoningEffort` and the divider call out that `DefaultRegistry()` must not run per-frame, and the divider correctly does no registry lookup. 
- The divider segment is width-safe — it rides on `meta`, so it drops out together under the narrow-width fallback (`view.go:185`) rather than overflowing.
- Hiding the segment on auto (`reasoningEffort == ""`) makes its appearance/disappearance the state feedback. Clean.

### Suggestion (the one thing I'd change)
The 7-term modal gate is now duplicated **verbatim** between the existing shift+tab case and the new Ctrl+T case:

```go
if m.pendingPermission == nil && m.pendingAskUser == nil && m.pendingSpecReview == nil &&
   m.providerWizard == nil && m.mcpAddWizard == nil && m.mcpManager == nil && m.picker == nil {
```

The Ctrl+T comment even says "the same gate shift+tab uses above." Two copies will drift the day a new modal is added to one and not the other. Worth extracting a single predicate and using it in both places, e.g.:

```go
// noBlockingModal reports that no modal surface (permission prompt, ask_user,
// spec review, provider/MCP wizard, MCP manager, or picker) is up, so a global
// shortcut may act instead of falling through to a modal's own handler.
func (m model) noBlockingModal() bool {
    return m.pendingPermission == nil && m.pendingAskUser == nil && m.pendingSpecReview == nil &&
        m.providerWizard == nil && m.mcpAddWizard == nil && m.mcpManager == nil && m.picker == nil
}
```

Then both cases read `if m.noBlockingModal() {`. Pure refactor, no behavior change.

### Minor / optional
- `cycleReasoningEffort` returns `(model, tea.Cmd)` but the cmd is always `nil`. Fine for the `return m.cycleReasoningEffort()` call site (it matches the `(tea.Model, tea.Cmd)` shape), just noting it never produces a command.
- Discoverability: the binding isn't in any help/keymap hint. The divider shows the *state* but not that Ctrl+T sets it — consider adding it to the shortcuts list so users can find it.

### Verdict
Solid and low-risk, with genuinely thorough tests. Only the duplicated modal gate is worth folding into a shared helper before merge; everything else is optional polish.

Vasanthdev2004:
@coderabbitai review

Vasanthdev2004:
@coderabbitai approve

gnanam1990 (commented):
## Re-review — reasoning-effort UX (Ctrl+T, divider, /effort card, hint)

This grew from the single Ctrl+T shortcut into a cohesive reasoning-effort experience, and it directly addressed my earlier note. Builds clean and the targeted tests pass locally (`-run 'Effort|CycleReasoning|ComposerDescriptionHint|ComposerDivider'`).

### Strengths
- **`noBlockingModal()` is extracted and shared** by both shift+tab and Ctrl+T — exactly the refactor I asked for. The 7-term gate now lives in one place, so the two shortcuts can't drift.
- **The divider's unconditional effort render is actually safe.** Every path that sets `reasoningEffort` gates or clears against the active model's supported set — `handleEffortCommand`, `cycleReasoningEffort`, the mode command, and crucially the model switch at `command_center.go:434` (clears the effort if the new model doesn't allow it). So a stale effort can't appear, and the comment matches the code. Nice reasoning for dropping the model-aware gate.
- `cycleReasoningEffort` / `reasoningEffortIndex` are clean, with a good caveat about not calling the registry lookup from the render path.
- The `/effort` command-card migration is consistent, and a no-control model still gets a real card with the "none for active model" hint rather than a grey status panel.
- Genuinely thorough tests: every cycle branch, the picker-on-unsupported-model, divider show/omit, and the description hint across single / ambiguous / args / @file-palette cases.

### Suggestions
1. **(Confirms CodeRabbit's open thread — `model.go` `composerDescriptionHint`.)** `value := strings.TrimSpace(m.input.Value())` runs *before* the whitespace check, so `/effort ` (trailing space — argument entry has begun) still renders the description. Test the **raw** value for the space (e.g. `strings.ContainsAny(m.input.Value(), " \t\n")`, or `strings.HasSuffix(m.input.Value(), " ")`) so the hint clears the instant an argument starts — matching the intent of `TestComposerDescriptionHintStaysEmptyAfterArgs`.
2. **`newEffortPicker`**: the `if m.reasoningEffort == "" { selected = 0 }` is a no-op (`selected` is already `0`). Drop it.
3. *Optional:* a one-item `[auto]` picker on a no-control model is a slightly unusual affordance (a popup with nothing to pick but auto). It's a reasonable consistency choice — just flagging it in case the inline status reads better there.

### Verdict
Solid and well-tested, and it folded in the earlier feedback. Only the trailing-whitespace nit (#1) is worth fixing before merge; the rest is optional polish.

gnanam1990 (approved):
**Approve** ✅

Solid, well-tested, and it folded in my earlier `noBlockingModal()` feedback — the shared modal gate is exactly right. I verified it builds and the targeted tests pass, and confirmed the divider's unconditional effort render is safe (every effort-setting path gates/clears against the active model, including the model switch at `command_center.go:434`, so no stale effort can show).

Non-blocking follow-up: the trailing-whitespace nit in `composerDescriptionHint` (also flagged by CodeRabbit) — checking the raw input so `/effort ` clears the hint — is worth a quick fix, but it's cosmetic and shouldn't hold the PR. The other two items in my review comment are optional polish.

Nice work.

gnanam1990 (approved):
### Review: Ctrl+T effort cycle + divider display

Solid, idiomatic Bubble Tea v2 work. Verified locally: `gofmt -l` clean, `go vet` clean, and the new tests pass under `-race`. No key collision (`Ctrl+T` is unused elsewhere and reuses the same `noBlockingModal()` gate as `Shift+Tab`), the effort cycle is correct on every branch (empty→first, mid→next, last/unknown→reset to auto), and the divider/hint gating is sound. Good test coverage on the cycle and rendering.

Two non-blocking nits:

1. `internal/tui/picker.go` (`newEffortPicker`) — `selected := 0` immediately followed by `if m.reasoningEffort == "" { selected = 0 }` is a no-op (it's already `0`). Drop the `if`.
2. `internal/tui/model.go:2294` — `newEffortPicker` always returns a non-nil `*commandPicker`, so the `if picker := m.newEffortPicker(); picker != nil` guard is always true and the nil branch is dead. Either simplify the guard, or return `nil` for models with no effort controls if that was the intent.

Optional: one `Update`-level test asserting `Ctrl+T` actually mutates effort (and is a no-op while a modal is open) would close the last small gap — current tests call `cycleReasoningEffort` directly.

Approving — the nits are cosmetic.

Vasanthdev2004:
@coderabbitai approve

PR #232, opened 2026-06-17
tui: selectable + clickable permission popup

gnanam1990:
## Summary
Makes the permission prompt **navigable and clickable**, on top of the existing `a`/`y`/`d` hotkeys (which still resolve directly):
- **↑/↓** and **Tab/Shift+Tab** move the highlight across **allow once / always / deny** (resting default: *allow once*); **Enter** confirms the highlighted option.
- Each option renders on its **own row**, and a **left-click** on a row resolves that choice directly — no per-column hit-testing.

Re-ports a local feature onto current `main`, adapted to main's refactored row-based selection layer (`transcriptBodyItem` render closures + `layout.selectable`).

## How it works
- `permission_prompt.go` — `permissionOptions` + cursor `clamp`/`move`/`confirm` helpers (the cursor lives on the pending prompt).
- `model.go` — `pendingPermissionPrompt` gains a `cursor`; Enter → `confirmPermissionCursor`, ↑/↓/Tab → `movePermissionCursor`; `a`/`y`/`d` unchanged and still ignore the cursor.
- `rendering.go` — `renderFocusedPermissionPrompt` now renders one row per option with the cursor highlighted (`▸` + reverse label) and returns each option's **card-relative Y offset** so the caller can register clickable rows.
- `transcript_selection.go` — `transcriptSelectableLine` gains `permOption`/`permChoice`; the pending-permission item registers each option row as clickable, and a left-click on one calls `resolvePermission`.

## Tests
- cursor default (allow once) / move / wrap; Enter confirms the highlight; `a`/`y`/`d` still resolve directly; render emits highlighted clickable offsets that map to the right rows.
- Existing card + submit-block tests updated for the per-option layout and Enter-confirms-the-highlight behavior.
- Gate: `gofmt`, `go vet`, `go build` (host/linux/windows), `go test ./internal/tui -race`, staticcheck — all green (the two unused-func nits are pre-existing on `main`).

## Note
Keyboard nav + the offset→row mapping are covered by tests; the end-to-end **mouse click** on an option is best confirmed interactively in a real terminal.

## Release Notes

* **New Features**
  * Permission prompt now supports keyboard navigation using Tab, Shift+Tab, and arrow keys to select between options
  * Added visual highlighting to indicate the currently selected permission choice
  * Permission options are now clickable with mouse support for direct selection
  * Improved permission prompt layout with each option displayed as a separate, selectable line

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -308,6 +308,10 @@ type permissionRequestMsg struct {
 type pendingPermissionPrompt struct {
 	request agent.PermissionRequest
 	decide  func(agent.PermissionDecision)
+	// cursor is the highlighted option index (into permissionOptions): 0=allow
+	// once (the resting default), 1=always, 2=deny. Moved by ↑/↓/Tab; confirmed
+	// by Enter or a click. The a/y/d hotkeys resolve directly and ignore it.
+	cursor int
 }
 
 // askUserRequestMsg is the TUI-loop equivalent of permissionRequestMsg: the
@@ -696,7 +700,9 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, nil
 			}
 			if m.pendingPermission != nil {
-				return m, nil
+				// Enter confirms the highlighted option (default: allow once); the
+				// a/y/d hotkeys and a click still resolve directly.
+				return m.confirmPermissionCursor()
 			}
 			if m.pendingAskUser != nil {
 				return m.submitAskUserAnswer()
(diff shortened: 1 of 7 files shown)


Vasanthdev2004 (commented):
LGTM ? no issues found. The permission popup cursor/click wiring is correct: heightCacheStable is a real field on transcriptBodyItem on main (added by #226), the cursor wraps correctly under Go signed modulo, and the option offsets match the styledBlockFill one-top-border contract.

Vasanthdev2004 (approved):
LGTM ? no issues found. The permission popup cursor/click wiring is correct: heightCacheStable is a real field on transcriptBodyItem on main (added by #226), the cursor wraps correctly under Go signed modulo, and the option offsets match the styledBlockFill one-top-border contract.

Vasanthdev2004:
Parity check against the reference Droid (D) implementation.

D's SandboxPermissionPrompt offers context-dynamic options: for file-write violations it offers both 'Allow always (writes to <parent dir>)' and 'Allow always (writes to <exact path>)'; for network violations it offers 'Allow always (<domain>)'; for deny-list entries it offers 'Remove <path> from deny list'. See src/sandbox/SandboxPermissionPrompt.ts -> getSandboxPromptOptions. Zero's permissionOptions() is a fixed 3-item list (allow once / always / deny) with no path/domain context in the 'Allow always' label.

This is a known gap, not a bug in this PR's scope. The cursor/wrap/click wiring itself matches D's SandboxViolationPrompt behavior (up/down wraps, Enter confirms the highlighted option).

PR #233, opened 2026-06-17
fix(providers): clear error when a model host is unreachable

gnanam1990:
## Problem

When the server zero connects to can't reach the model, the surfaced error is an opaque, low-level transport string that reads like a zero bug and gives the user nothing to act on.

- A local **Ollama** daemon serving a `-cloud` model answers on `localhost` but returns **HTTP 502** carrying a proxied error when it cannot reach its own cloud backend:

  ```
  provider error: {"error":"Post \"https://ollama.com:443/v1/chat/completions?ts=…\": net/http: TLS handshake timeout"}
  ```

- A direct hosted endpoint blocked by the local network (DNS / proxy / VPN / firewall) fails the same way at the transport layer and surfaces as `provider stream error: … TLS handshake timeout`.

In both cases **the request never reached the model** — it's a connectivity problem upstream of the agent — but the message doesn't say so, so users assume zero is broken.

## Change

- Add `providerio.UpstreamUnreachable(message) (string, bool)` — detects a transport-level connectivity failure (`TLS handshake timeout`, `context deadline exceeded`, `connection refused`, `no such host`, `network is unreachable`, `i/o timeout`) **bound to a concrete host**, and rewrites it into an actionable message:

  > upstream unreachable: the model server could not connect to ollama.com:443 (TLS handshake timeout). The request never reached the model — this is a network failure between the model server and its upstream, not a model error. Verify the host is reachable from the machine running the model server (DNS/proxy/VPN/firewall); a local daemon proxying a cloud model — e.g. an Ollama daemon serving a "-cloud" model — must itself be able to reach the internet.

- Matches only when **both** a failure marker and a host are present, so the agent's own request-deadline cancellations (which carry no host) are left untouched.
- Wired into the OpenAI-compatible adapter's two error points — the gateway-error path (`emitHTTPError`) and the transport-error path — so it covers both the local daemon proxy and direct hosted providers. Secrets are still redacted on the way out.

## Why it isn't "just retry"

The existing 5xx/429/network retry already runs; this is for when the upstream is *persistently* unreachable from the host. No amount of retrying reaches a blocked host, so the right fix is a clear diagnosis rather than a silent failure or an opaque string.

## Tests

- `TestUpstreamUnreachable` — 7 cases over the real proxied / direct / DNS / connection-refused shapes, plus negative cases (genuine model error; bare marker with no host).
- `TestStreamCompletionHumanizesUpstreamUnreachableGatewayError` — end-to-end: a 502 carrying the proxied body yields a single, clear, host-named error event.

Gate: `gofmt`, `go vet`, build, `staticcheck`, and `go test -race` on the affected packages (`providerio`, `openai`, `anthropic`, `gemini`, `zeroruntime`) — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed and non-2xx error messaging for upstream connectivity issues by humanizing “upstream unreachable” with the affected host and underlying reason, avoiding generic provider stream error text.
  * Ensured caller cancellations/deadlines are reported verbatim and are not misclassified as “upstream unreachable.”

* **Tests**
  * Added unit tests validating upstream-unreachable error humanization and confirming context deadline errors are not rewritten.

### DIFF
--- a/internal/providers/openai/provider.go
+++ b/internal/providers/openai/provider.go
@@ -163,6 +163,15 @@ func (provider *Provider) stream(ctx context.Context, body []byte, events chan<-
 			}
 		}, 0)
 	if err != nil {
+		// A caller-driven cancel/deadline (parent ctx done) surfaces here as a
+		// transport error whose string can contain "context deadline exceeded" plus
+		// the request host — which UpstreamUnreachable would mislabel as an upstream
+		// outage. Check the parent context first and surface its error verbatim, so
+		// only genuine connect failures (ctx still live) get humanized.
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			sendEvent(ctx, events, zeroruntime.StreamEvent{Type: zeroruntime.StreamEventError, Error: provider.redact("provider stream error: " + ctxErr.Error())})
+			return
+		}
 		// A direct connection that never completes (e.g. a hosted endpoint blocked
 		// by the local network) surfaces as a transport error; humanize it the same
 		// way as a proxy's gateway error so the user sees a clear connectivity cause.
--- a/internal/providers/openai/provider.go
+++ b/internal/providers/openai/provider.go
@@ -163,6 +163,13 @@ func (provider *Provider) stream(ctx context.Context, body []byte, events chan<-
 			}
 		}, 0)
 	if err != nil {
+		// A direct connection that never completes (e.g. a hosted endpoint blocked
+		// by the local network) surfaces as a transport error; humanize it the same
+		// way as a proxy's gateway error so the user sees a clear connectivity cause.
+		if humanized, ok := providerio.UpstreamUnreachable(err.Error()); ok {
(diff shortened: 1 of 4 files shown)


gnanam1990:
Addressed in f257394. Added a `ctx.Err()` guard at the top of the transport-error path: a caller-driven cancel/deadline (parent ctx done) is now surfaced verbatim as `provider stream error: <ctx error>`, and only genuine connect failures (ctx still live) reach `UpstreamUnreachable`. Added a deadline regression test — the prior cancel test used `Cancel`, whose `context canceled` is not a marker, so it never covered the `context deadline exceeded` path.

@coderabbitai review

Vasanthdev2004 (approved):
LGTM. The ctx.Err() guard before UpstreamUnreachable is the right call: Go's http client returns a transport error containing 'context deadline exceeded' when the request is canceled, and the old path would have mislabeled those as upstream outages. UpstreamUnreachable matches on both a marker and a host, so the agent's own request-deadline cancellations stay untouched. Humanized message correctly distinguishes 'never reached the model' from 'model rejected the request'. No blockers.

PR #234, opened 2026-06-17
docs: add AWESOME_AGENTS curated list of agent instructions

Vasanthdev2004:
{
  "title": "docs: add AWESOME_AGENTS curated list of agent instructions",
[...]

PR #235, opened 2026-06-17
docs: add AWESOME_AGENTS curated list of agent instructions

Vasanthdev2004:
Ten hand-picked agent configurations for Zero: code review, test generation, refactor, security audit, migration, documentation, incident response, release notes, spec drafting, and dependency updates. Each entry is a self-contained markdown block the reader can copy into \.factory/droids/<name>.md\ and invoke immediately.

Also includes a thirty-entry honorable-mentions tail covering codebase hygiene, test and verification, repository intelligence, release and ops, and workspace ergonomics. Together with the curated ten, this frames the list as ten out of about a hundred candidates worth knowing about.

The doc is contributor-friendly: a Contributing section spells out the kebab-case filename, level-2 heading, italic tagline, and the standard section shape for new entries.

Configuration code blocks use a 4-backtick outer fence so the nested JSON examples inside the prompt body (3-backtick) render as code instead of terminating the outer markdown fence.

Supersedes #234. No source code changes. Docs only.


Vasanthdev2004:
Superseded. Closing; the awesome-list framing was the wrong call.

anandh8x (changes requested):
High: docs/AWESOME_AGENTS.md is committed as UTF-16LE, so Git treats the Markdown as binary.

The PR diff shows docs/AWESOME_AGENTS.md as `Bin 0 -> 124402 bytes`, and `file` reports it as `Unicode text, UTF-16, little-endian text, with CRLF line terminators`. That means reviewers cannot see normal line diffs, repo search/grep will miss it, and GitHub Markdown rendering/tooling may not behave like a normal docs file. Please re-save it as UTF-8 Markdown with normal text line endings before merging.

High: the copy-paste instructions do not match Zero’s actual specialist format.

docs/AWESOME_AGENTS.md:12 tells users to save configs under `.factory/droids/<name>.md` / `~/.factory/droids/<name>.md`, and the first config at docs/AWESOME_AGENTS.md:57 only provides a `description:` frontmatter key. Zero loads project specialists from `.zero/specialists/*.md` and user specialists from `~/.config/zero/specialists/*.md` (see docs/SPECIALISTS.md), and manifests require a `name` plus `description` frontmatter. Following this new doc would create files Zero never loads; even if moved into `.zero/specialists`, the examples are missing the required `name` metadata and include unsupported droid-style keys such as `globs` / `alwaysApply` later in the file.

Verification: decoded the file with iconv for review and ran `git diff --check 6df5c2b4cbd145a487948311ce74424f6dfe5182...e420b3aeac75888b7f25516d5e3776216eef2b2d` successfully. I did not run a full test suite because this PR is docs-only.

PR #237, opened 2026-06-17
docs: rewrite AGENTS.md as user-facing 'Extending Zero' guide

Vasanthdev2004:
﻿## Summary

Replaces the project-specific 19-line AGENTS.md with a user-facing **\"Extending Zero\"** guide. The previous file was a Go-internal instruction set; this rewrite positions AGENTS.md as the entry point for end users who want to customize Zero, the way similar files work in Claude Code, Codex, and other CLI agents.

## What's in the new AGENTS.md

Nine sections, ~290 new lines:

1. **Drop a project AGENTS.md** — where to put it (./AGENTS.md or ./.zero/AGENTS.md), the frontmatter format, and authoring tips.
2. **Custom specialists (droids)** — three scopes (built-in / user / project), manifest format, the zero specialist CLI commands. Points to docs/SPECIALISTS.md for the full spec.
3. **Skills** — */SKILL.md discovery under ~/.local/share/zero/skills/ (or $ZERO_SKILLS_DIR), with format example.
4. **Hooks** — six lifecycle events (eforeTool, fterTool, sessionStart, sessionEnd, specialistStart, specialistStop), JSON config at ~/.config/zero/hooks.json and ./.zero/hooks.json, exit-code semantics, and a note that misbehaving hooks show up in zero doctor.
5. **MCP** — as a client (the mcp.servers block in config.json, both stdio and http transports, zero mcp CLI) and as a server (zero serve --mcp).
6. **Plugins** — plugin.json schema bundling tools / hooks / skills, zero plugins install / list / enable / disable / remove.
7. **Configuration locations** — the resolution order (built-in < user < project < CLI < env) and the deliberate ignore of dditionalWriteRoots in project config.
8. **End-to-end example** — what a team should commit to the repo vs what each contributor keeps in their home dir.
9. **Reference** — links to README.md, docs/SPECIALISTS.md, docs/PRD.md, docs/STREAM_JSON_PROTOCOL.md, docs/INSTALL.md.

## Why

Three previous PRs on AGENTS.md were closed because the framing kept drifting (awesome-list, then Go-specific 10-mode instructions). This rewrite keeps it as a single user-facing page that points outward to the deeper docs where they exist, and explains the rest inline where they don't.

## Risk

- File rename is just a content swap; path stays AGENTS.md (case-insensitive Windows matches the existing tracked file AGENTS.MD).
- No code changes, no test impact.
- The file is referenced from description: and globs: frontmatter that the agent runtime reads, so the new description and globs: "*.go, *.js, *.md" values are the only behavior-affecting changes. They are deliberately broader than the previous Go-only globs.

## Summary by CodeRabbit

* **Documentation**
  * Replaced `AGENTS.md` with a full “Extending Zero” guide, detailing guideline discovery, frontmatter fields, specialists, hooks, MCP configuration, plugins, configuration precedence, and repository examples.
* **New Features**
  * Project guidelines are now injected into the prompt by walking from the nearest git root to the current directory (general-to-specific), supporting `AGENTS.md`, `ZERO.md`, and `.zero/AGENTS.md` with truncation safeguards.
* **Tests**
  * Added unit tests for case-insensitive guideline resolution, traversal order, fallback behavior, and content truncation at size limits.

### DIFF
--- a/AGENTS.MD
+++ b/AGENTS.MD
@@ -1,16 +1,295 @@
 ---
-description: Use Go-native project commands; Node is only for the npm wrapper.
-globs: "*.go, *.js, package.json, .github/workflows/*.yml, docs/*.md"
+description: Extending Zero — how to write AGENTS.md files, custom specialists, skills, hooks, MCP servers, and plugins for the open-source CLI coding agent.
+globs: "*.go, *.js, *.md"
 alwaysApply: false
 ---
 
-Default to Go commands in this repository.
+# Extending Zero
 
-- Use `go run ./cmd/zero` to launch the local CLI.
-- Use `go test ./...` for the repository test suite.
-- Use `go run ./cmd/zero-release build` to build the release-facing binary.
-- Use `go run ./cmd/zero-release smoke` to smoke-test the built binary.
-- Use `go run ./cmd/zero-release package` and `go run ./cmd/zero-release verify` for release artifact checks.
-- Use `go run ./cmd/zero-perf-bench` for performance smoke work.
-- Do not add Bun or TypeScript build/test dependencies for repository validation.
-- Use Node only when directly exercising the npm wrapper at `bin/zero.js`.
+Zero is an open-source terminal coding agent in the same family as Claude Code, Codex, and Droid. Out of the box it does the obvious things — read, edit, run, search — but the design point of the project is that **every surface is configurable**. This document is the user-facing guide for that configuration.
+
+If you only want to *use* Zero, the [README](README.md) is enough. This page is for the other three jobs:
+
+1. Tell the agent about *your* project (drop an `AGENTS.md` in your repo).
+2. Add new specialist sub-agents (custom droids).
+3. Wire Zero into the rest of your toolchain (MCP, skills, hooks, plugins).
+
+## 1. Drop a project `AGENTS.md`
+
+When Zero starts in a directory, it looks for project-level instructions in two places and applies both:
+
+| Path | Scope | Notes |
+| --- | --- | --- |
+| `./AGENTS.md` | repo-root | The classic spot — committed to your repo, shared with the team. |
+| `./.zero/AGENTS.md` | project-local | Hidden, gitignored. Personal notes that stay out of git. |
+
+Both files use the same format. YAML frontmatter is optional; the markdown body is loaded as instructions for the agent.
+
+```markdown
+---
+description: One-line summary the agent sees in its prompt.
+globs: "*.go, *.py"
+alwaysApply: false
+---
+
+# Project conventions for <your project>
+
+- Build with `make`, not `go build` directly.
+- Tests live next to the source file (`foo_test.go` next to `foo.go`).
+- Run `make lint` before opening a PR.
+- Never edit files under `third_party/` — those are vendored.
+```
+
+Tips:
+
+- Keep it under ~80 lines. Long files dilute the agent's attention.
+- Use `globs:` to scope which files the rules apply to (mirrors Claude Code's `settings.json` scoping).
+- Re-state rules in the imperative voice: "Run `make lint`", not "you should consider running the linter".
+- Don't put secrets, model IDs, or environment-specific paths in `AGENTS.md`. Use `config.json` for those.
+
+## 2. Custom specialists (droids)
+
+Specialists are Zero's sub-agents. Three scopes, in priority order:
+
+| Scope | Path | Shared? |
+| --- | --- | --- |
+| Built-in | compiled into Zero | yes — `worker`, `explorer`, `code-review` |
+| User | `~/.config/zero/specialists/*.md` | no — your machine only |
+| Project | `./.zero/specialists/*.md` | yes — the repo team |
+
+Project overrides user overrides built-in when names collide.
+
+A specialist is a markdown manifest with frontmatter and a system prompt:
+
+```markdown
+---
+description: Reviews API changes for breaking-change risk and missing tests.
+tools: read-only,plan
+---
+
+You review API changes. For every changed hunk in `internal/api/` or any file
+that ends in `_api.go`:
+
+1. Confirm the public signature is backward-compatible, or note the breaking
+   change explicitly with the migration path.
+2. Confirm a corresponding test exists in `internal/api/*_test.go` and that
+   the new behaviour is exercised.
+3. Flag any new exported symbol without a doc comment.
+
+Reply with one JSON object per finding: `{"file", "line", "severity", "message", "fix"}`.
+```
+
+CLI management:
+
+```bash
+zero specialist list
+zero specialist show api-reviewer
+zero specialist create api-reviewer \
+    --project \
+    --description "Reviews API changes" \
+    --tools read-only,plan \
+    --prompt-file api-reviewer.md
+zero specialist edit api-reviewer --project
+zero specialist delete api-reviewer --project
+```
+
+The full format spec (frontmatter fields, tool scopes, prompt conventions) is in [`docs/SPECIALISTS.md`](docs/SPECIALISTS.md).
+
+## 3. Skills
+
+Skills are markdown instruction packs the agent can pull in on demand. Each skill is a directory containing a `SKILL.md`:
+
+```
+~/.local/share/zero/skills/
+  run-benchmarks/
+    SKILL.md
+  write-changelog/
+    SKILL.md
+```
+
+`SKILL.md` format:
+
+```markdown
+---
+description: Run the project's benchmark suite and summarize the deltas.
+---
+
+# Run benchmarks
+
+1. `make bench` — captures the wall-clock and RSS before and after.
+2. `benchstat before.txt after.txt` — diffs the two.
+3. Report any regression > 5% with the function name and the previous value.
+```
+
+Override the directory with `ZERO_SKILLS_DIR=...`. A missing directory is fine — Zero just reports "no skills".
+
+The `skill` core tool lets the agent load any discovered skill by name.
+
+## 4. Hooks
+
+Hooks fire shell commands on lifecycle events. Configure them in JSON:
+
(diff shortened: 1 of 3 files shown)


Vasanthdev2004:
﻿## Follow-up: two more commits since the original opening

### 4a22259 docs(AGENTS): align Extending Zero with actual loader behavior + D-style features

The first version of this PR made six claims that were wrong about Zero itself. This commit fixes them and surfaces the new behavior introduced in 6fb0eb2:

- **Section 1 rewritten** to match the real priority order [AGENTS.md, ZERO.md, .zero/AGENTS.md], the case-insensitive basename callout, the per-file (8 KiB) and total (32 KiB) caps, the eager-load + restart-required caveat, and the new monorepo path-walking ("walks from cwd up to git root, general-to-specific order").
- **Section 2 (Specialists)**: roadmap note about the planned /droids in-UI manager.
- **Section 3 (Skills)**: corrected the discovery-root chain (ZERO_SKILLS_DIR → XDG_DATA_HOME/zero/skills → ~/.local/share/zero/skills/), noted that project-level skills aren't supported yet, and called out the plugin-declared-skill gap.
- **Section 4 (Hooks)** and **Section 6 (Plugins)**: roadmap notes about the planned /hooks and /plugins in-UI managers.
- **Frontmatter globs** broadened to *.go, *.js, *.md, *.json, *.toml, *.yaml, *.yml to match the actual file types Zero reads.

### 6fb0eb2 feat(agent): make AGENTS.md loader case-insensitive + add path-walking

Two changes to the loader in internal/agent/system_prompt.go that bring Zero closer to D's DynamicContextDiscovery model and fix a real bug.

**1. Case-insensitive basename matching.** The git-tracked filename is AGENTS.MD (uppercase MD), but the loader was looking for AGENTS.md (lowercase md). On a case-sensitive filesystem (Linux, WSL, or a CI runner) this made the file invisible to the loader. The new indProjectContextFile uses ReadDir + ToLower to resolve the on-disk filename and matches AGENTS.MD, Agents.md, gents.md, etc. to the same entry. The visible filename in the prompt label is now the actual on-disk name (AGENTS.MD), not the lookup name.

**2. Upward path-walking.** The loader previously did a single first-match-wins lookup at cwd. The new projectGuidelines walks from cwd up to the nearest git root, finds the first matching project context file at each level, and injects them in general-to-specific order (git root first, cwd last). A monorepo with a root AGENTS.md and per-service AGENTS.md files now sees both, with the most specific rules last. This mirrors D's per-Read dynamic path discovery, just at session start.

**Cap behavior:**
- Per-file cap unchanged: 8 KiB.
- New total cap of 32 KiB across all matched files, enforced via a rolling counter; when the budget is exhausted, further files are dropped.
- Truncation marker unchanged: \n… (truncated).

**New helpers:** projectGuidelineDirs, projectGuidelineLabel, indProjectContextFile, 
esolveDirCaseInsensitive, indProjectGitRoot.

**New tests (all passing, full go test ./internal/agent/ suite still green):**
- TestBuildSystemPromptInjectsProjectGuidelinesCaseInsensitive — verifies AGENTS.MD resolves on a case-sensitive FS.
- TestBuildSystemPromptProjectGuidelinesPathWalkingMonorepo — verifies root + leaf both injected, root first.
- TestBuildSystemPromptProjectGuidelinesZeroFallback — verifies ZERO.md is the second-priority name.
- TestBuildSystemPromptProjectGuidelinesProjectLocalFallback — verifies .zero/AGENTS.md is the third-priority fallback.
- TestBuildSystemPromptProjectGuidelinesTruncatesAtTotalCap — verifies the 32 KiB total cap is enforced with the truncation marker.
- TestProjectGuidelineDirsOrdersRootToLeaf and TestProjectGuidelineDirsCollapsesToCwdWithoutGitRoot — helper unit tests.
- TestFindProjectContextFileCaseInsensitiveBasename — helper unit test.

The existing TestBuildSystemPromptInjectsWorkspaceContext still passes; for a single-file repo the label is AGENTS.md / AGENTS.MD matching the on-disk name.

### What I deliberately did NOT change

- **Per-Read dynamic discovery like D.** D injects AGENTS.md content as a system reminder on each Read tool call, which keeps the prompt cache valid. Zero's loader is eager at session start, and changing that model would invalidate a much larger refactor. Worth a follow-up PR if you want it.
- **In-UI managers (/droids, /hooks, /plugins).** Each is a multi-file feature on its own; flagged in the doc as roadmap.
- **Project-level skills.** The skills loader (internal/skills/skills.go) explicitly notes that plugin-declared skills are not yet merged. Documented as a gap; the fix is a separate PR.

gnanam1990 (changes requested):
### Review: AGENTS.md → "Extending Zero" guide

The structure and writing are a real improvement — user-facing, well-organized, good copy-paste examples. But the guide documents several commands/flags/behaviors that don't exist in the current code, which will mislead users. I grep-verified each against the tree:

1. **Plugin verbs (§6):** `zero plugins install` / `enable` / `disable` aren't implemented — `runPlugins` (`internal/cli/extensions.go`) only dispatches `list` / `add` / `remove`(`rm`). Use `zero plugins add <path>`; drop `enable`/`disable` for plugins (those exist for `hooks`/`mcp`, but not plugins).
2. **Specialist prompt (§2):** `zero specialist create ... --prompt-file api-reviewer.md` — there is no `--prompt-file` flag (`internal/cli/specialist.go`). It's `--prompt "<text>"` (inline).
3. **`${env:...}` (§5):** `"Authorization": "Bearer ${env:GITHUB_TOKEN}"` — there is no `${env:}` expansion anywhere; MCP header/env values are sent verbatim (`internal/mcp/config.go`, `client.go` `mergeProcessEnv`). A user following this would transmit the literal string `${env:GITHUB_TOKEN}`.
4. **Hook arg placeholders (§4):** `"args": ["${tool}", "${input}"]` — hook args are passed verbatim to `exec.CommandContext`; the tool/input payload reaches hooks as JSON on **stdin**, not via arg substitution (`internal/hooks/dispatch.go`). No `${...}` expansion occurs.
5. **`zero doctor` (§4):** "use `zero doctor` to surface [hooks]" — doctor has no hook check (`internal/doctor/`): its checks are runtime / config / provider / connectivity / sandbox / lsp.

Secondary (non-blocking): the `globs:` / `alwaysApply:` frontmatter described in §1 is currently inert — the loader injects the whole file (frontmatter included) as prompt text and does no frontmatter parsing, so `globs:` doesn't scope anything yet. Also the README/CLI use "specialists" throughout, so the "droids" synonym reads inconsistently.

Everything else checks out — `serve --mcp`, `mcp add … -- <cmd>`, the hook event names, skills dir order, specialist/plugin paths, and the byte caps are all accurate. Requesting changes only on the items above; once the non-existent syntax is removed or fixed, this is a strong guide.

gnanam1990 (approved):
### Re-review (after a6ce791b)

The follow-up commit resolves every must-fix item from the previous review — I re-verified each against the current code and the updated `AGENTS.MD`:

1. **Plugins** — now `zero plugins add ./github-pr-review`; the non-existent `install`/`enable`/`disable` verbs are gone, with an accurate note that a plugin is enabled by being present (or `"enabled": false` in its manifest). ✓
2. **Specialist prompt** — `--prompt-file` replaced with `--prompt "$(cat api-reviewer.md)"`, plus a useful `zero specialist path`. ✓
3. **`${env:...}`** — now states plainly there is no env expansion (values sent verbatim) and documents the real ways to keep a token out of the file. ✓
4. **Hook args** — `${tool}`/`${input}` dropped; correctly explains the payload arrives as JSON on **stdin**, with a stdin-reading handler example. ✓
5. **`zero doctor`** — reworded to point at the audit log; no longer claims a doctor hook check. ✓

Secondary items also handled: the inert `globs:`/`alwaysApply:` frontmatter example is replaced with an accurate "preserved verbatim, not parsed today" note, and the `(droids)` synonym plus the non-existent `/droids` `/hooks` `/plugins` slash-command roadmaps are gone.

All clean corrections, no new claims that don't match the code. Thorough turnaround — approving.

Vasanthdev2004:
@coderabbitai approve

Vasanthdev2004:
Pushed a6ce791 to address the review. Walked each item against the tree, then re-checked the rewritten sections after editing:

- §6 plugin verbs: replaced `install` / `enable` / `disable` with `add` and noted that enable/disable are not CLI verbs today (the plugin is on/off by presence or by the manifest's `"enabled": false`, mirroring the `runPlugins` switch in internal/cli/extensions.go).
- §2 specialist prompt: dropped the `--prompt-file api-reviewer.md` line, switched to `--prompt "$(cat api-reviewer.md)"` (the flag is inline per `runSpecialistCreate`), and added `zero specialist path` to the example.
- §5 MCP env vars: removed the `${env:GITHUB_TOKEN}` line and the "token-bearing values come from env vars" phrasing. Headers are sent verbatim (internal/mcp/permissions.go, client.go mergeProcessEnv), so the section now lists the three real ways to keep tokens out of the file (wrapper script, command substitution, server reading its own env) instead.
- §4 hook args: dropped the `["${tool}", "${input}"]` placeholders, called out that the payload is JSON on stdin (internal/hooks/dispatch.go `execCommandRunner`), and added a short bash handler that reads stdin and blocks `rm -rf`.
- §4 doctor: removed the `zero doctor` line; doctor's checks are runtime / config / provider / connectivity / sandbox / lsp (internal/doctor/doctor.go). Repointed to the audit log.

Secondary: dropped the `globs:` / `alwaysApply:` example and the "Use globs: to scope" tip — the loader injects the whole file as prompt text and does not parse those keys (greped for frontmatter parsing in internal/agent, nothing). Added a one-liner noting the frontmatter is preserved verbatim. Also dropped the `(droids)` parenthetical on §2 and the §2/§4/§6 roadmap slash commands (`/droids`, `/hooks`, `/plugins` are not implemented — only `/mcp` is, per internal/tui/commands.go) and rephrased the roadmaps to "in-UI … manager is on the backlog".

PR #239, opened 2026-06-17
feat: free out-of-the-box web (keyless Firecrawl default) + resilient MCP startup

gnanam1990:
## Summary

Free, out-of-the-box **web search + scraping** for open-source users — no API key, no setup — plus a robustness fix that the default depends on.

Verified live: `mcp check` against the keyless endpoint returns **25 tools, no key**, and a built binary shows `firecrawl [http] enabled` on first run.

## 1. `fix(mcp)`: skip an unreachable MCP server instead of aborting startup

`RegisterTools` connected to every server eagerly and failed the **whole batch** if any one couldn't be reached or list its tools — and the caller **aborts startup with exit 1** on that error. So a single unreachable MCP server prevented `zero` from launching at all and disabled every other working server.

Now registration is **best-effort across servers but atomic within a server**: each server's tools are validated as a unit before anything commits, and a server that can't be connected, returns a nameless tool, or conflicts is **skipped** (recorded on `Runtime.Skipped()`), not fatal. The caller surfaces each skip as a warning and keeps launching. The "no dead tools" / no-partial-tool-set guarantees are preserved.

This is a real robustness fix on its own — one flaky MCP server should never break `zero`'s startup.

## 2. `feat(mcp)`: keyless Firecrawl as a default web server

[Firecrawl just launched **Keyless**](https://www.firecrawl.dev/blog/firecrawl-keyless-launch): a hosted MCP at `https://mcp.firecrawl.dev/v2/mcp` with 1,000 free credits/month and no account. This seeds it **enabled by default** in `ResolveMCP` (before user/project config merges), so web search, JS-rendered scraping, crawl, PDF→markdown, and structured extraction work with zero setup.

You stay in control:

```bash
zero mcp tools list           # the firecrawl_* tools
zero mcp disable firecrawl    # off-switch — works even though the default lives in code
```

- The built-in `web_fetch` tool remains the **always-local floor** (no third party).
- Any field is overridable: point at a **self-hosted** Firecrawl, or add a **BYO-key** header for higher limits.
- **License:** Firecrawl is AGPL-3.0, but `zero` only *calls* it over the network — that's not linking/embedding, so the AGPL never reaches `zero`'s code. (SDKs are MIT regardless.)
- **Privacy:** keyless routes through `firecrawl.dev`; the README discloses this and documents disable/self-host.

### Why default-on is now safe
On an offline/blocked network the default would otherwise hang or break startup. Part 1 makes it **degrade gracefully**: the server is skipped with a warning and `web_fetch` keeps working. (`zero mcp disable firecrawl` writes an override even though the default isn't in the config file — the disable path now recognizes built-in defaults.)

## Tests
- `internal/mcp`: skip-and-keep-others + prior-state-preserved (the two old fail-hard tests rewritten to the new skip semantics).
- `internal/config`: default seeded enabled; user can disable; user can override the URL keeping other fields; `IsDefaultMCPServer`.
- `internal/cli`: `zero mcp disable firecrawl` turns the seeded default off end-to-end.

Full gate green: gofmt, `go vet`, build, `staticcheck`, `go test ./... -race`.

## Release Notes

* **New Features**
  * Added built-in Firecrawl web search & scraping (keyless, no setup needed).
  * Introduced `zero mcp` commands to list and disable MCP servers.

* **Bug Fixes**
  * Unreachable/invalid MCP servers are now skipped at startup (non-fatal), with warnings so other tools keep working.
  * Disabling the built-in Firecrawl default now works correctly even if no prior config entry exists.

* **Documentation**
  * Updated the guide with keyless privacy disclosures, self-hosting/auth options, and how to manage Firecrawl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,40 @@ The model registry tracks each model's capabilities, context window, and cost �
 
 Every mutating tool routes through the permission policy **before** any side effect.
 
+### Web search & scraping (free, no API key)
+
+`web_fetch` is built in and always available: it pulls a single URL into clean
+markdown locally, with no third party. For **search**, JS-rendered scraping,
+whole-site crawls, PDF→markdown, and structured extraction, Zero ships the
+[Firecrawl](https://firecrawl.dev) **keyless** MCP server **enabled by default**, so
+web search and scraping work out of the box with **no setup and no API key** (1,000
+free credits/month). It is a normal MCP server, so you stay in control:
+
+```bash
+zero mcp tools list           # see the firecrawl_* tools it exposes
+zero mcp disable firecrawl    # turn it off — web_fetch stays as the local floor
+```
+
+- **Privacy / disclosure:** keyless requests route through `firecrawl.dev`. If you
+  prefer nothing leaves your machine, `zero mcp disable firecrawl` and rely on
+  `web_fetch`, or self-host below.
+- **Self-host (unlimited, private):** Firecrawl is open source (AGPL-3.0). Run your
+  own instance and override the default URL — Zero only *calls* it over the network,
+  so the AGPL never reaches into Zero's own code:
+  ```jsonc
+  // config.json
+  { "mcp": { "servers": { "firecrawl": { "type": "http", "url": "http://localhost:3002/v2/mcp" } } } }
+  ```
+- **Bring your own key (higher limits):** add an auth header over the default:
+  ```bash
+  zero mcp add firecrawl --type http --url https://mcp.firecrawl.dev/v2/mcp \
+    --header "Authorization: Bearer fc-your-key"
+  ```
+
+An MCP server that can't be reached at startup (e.g. Firecrawl on an offline
+machine) is **skipped with a warning, not fatal** — Zero still launches and the
+rest of its tools, including `web_fetch`, keep working.
+
 ### Extra write directories (`--add-dir`)
 
 Zero confines writes to the workspace by default. To let the agent write somewhere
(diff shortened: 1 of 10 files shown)


gnanam1990:
@coderabbitai review

Vasanthdev2004 (approved):
LGTM. The concurrent connect + per-server timeout is the right shape: a slow or unreachable server no longer blocks startup or the first model response, and the timeout path spawns a reaper goroutine to drain the connect result and close the partial client. The serial commit phase keeps registry + permission-store reads race-free, and a failed connect is recorded on Runtime.Skipped() so the caller can warn instead of aborting. The cancel functions are stored for kept servers and invoked only at Close, after the client is closed, so the stdio subprocess lifecycle is correct. IsDefaultMCPServer + the disable-path change make 'zero mcp disable firecrawl' work end-to-end even though the default lives in code. One judgment call worth flagging: this is a default-on third-party service with outbound traffic to firecrawl.dev. The README discloses it, web_fetch stays as the local floor, and 'zero mcp disable firecrawl' / self-host / BYO-key are documented, so it's a deliberate UX choice, not a bug. No blockers.

gnanam1990:
@coderabbit approve

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

Vasanthdev2004:
@coderabbitai approve

Vasanthdev2004:
@coderabbitai approve

gnanam1990:
@coderabbit approvable ?

anandh8x (approved):
Approved. I reviewed this as a Firecrawl/default MCP startup improvement. CI is green and I did not find a merge-blocking issue. Note: this does not replace the separate hosted web-search parity work; Firecrawl MCP tools remain permission-gated.

PR #240, opened 2026-06-17
Add sandbox architecture baseline

anandh8x:
## Summary
- add target sandbox backend names and enforcement metadata to backend and command plans
- wire the Phase 0 Linux, macOS, and Windows sandbox helper/package surfaces
- restore `sandbox.blockUnixSockets` compatibility through the Linux helper and keep `zero-seccomp` as a Linux compatibility wrapper
- add golden sandbox policy JSON coverage, architecture-baseline coverage, and a focused sandbox smoke script

## Tests
- `GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/cli ./internal/config ./internal/release ./cmd/zero-seccomp -run 'TestLegacySandboxEntrypointsAreExplicitAndExist|TestSandboxManagerBuildsCommandPlanThroughLinuxHelper|TestBuildLinuxSandboxCommandArgsSerializesPermissionProfile|TestParseLinuxSandboxHelperArgs|TestBuildLinuxSandboxBwrapArgsWrapsInnerSeccompStage|TestApplyConfiguredSandboxPolicyDiagnosticsFlags|TestResolveSandboxBlockUnixSocketsFromFile|TestCopyPackageFilesStagesLinuxSandboxHelper'`
- `GOCACHE=/tmp/zero-go-cache scripts/sandbox-smoke.sh`
- `git diff --check`

## Notes
- This is the Phase 0 baseline PR. It does not complete every replacement platform backend yet.
- Remaining legacy adapters are intentionally inventoried as follow-up delete targets for the staged rewrite.
- `sandbox.blockUnixSockets` is not silently dropped: existing config is resolved into policy, the Linux helper receives `--block-unix-sockets`, and Linux release packages continue to stage `zero-seccomp` for existing direct uses.
- Release packaging now stages `zero-linux-sandbox` plus `zero-seccomp` on Linux, and the Windows command runner/setup helpers on Windows.
- The branch has already merged current `main`; the overlapping `internal/agent/system_prompt.go` / `system_prompt_test.go` changes from #237 are compatible with this branch's git-root metadata validation.


gnanam1990 (approved):
### Review: sandbox architecture baseline

Reviewed the architecture and the Linux seccomp/Landlock path closely. This is a solid Phase-0 baseline — cross-platform backends (Linux bwrap + Landlock, Windows restricted-token/ACL/WFP, macOS Seatbelt), fail-closed posture, clean cross-compilation for both `GOOS=windows` and `GOOS=linux`, idiomatic error wrapping/rollback, and good golden-fixture + smoke coverage.

One thing worth calling out as **correct**, to preempt confusion: `appendNetworkSocketCheck` (`internal/sandbox/seccomp.go`) under NetworkDeny allows `socket(AF_UNIX)` and denies `AF_INET`/`AF_INET6`. That's the intended "block network, keep local IPC" semantics — the stricter "block unix sockets too" case is the separate `unixSocketBlockFilter` / `ApplyUnixSocketBlock`. I traced the generated BPF on both arches; it behaves as designed.

One nit (non-blocking): the seccomp filters currently have **structural** tests only — `TestNetworkDenySeccompFilterStructure` (`internal/sandbox/seccomp_test.go:35`) checks jump bounds, not semantics. A small table-driven BPF-interpreter test asserting actual verdicts for `socket(AF_UNIX/AF_INET)`, `socketpair`, `connect`, `bind`, and one benign syscall — for both arches — would lock in the behavior and guard against a future jump-offset regression.

Approving; the interpreter test is a good follow-up.

Vasanthdev2004 (changes requested):
### Review: sandbox architecture baseline

This is a substantial foundation. The `SandboxManager` layer with the `auto` / `require` / `forbid` preference system, the target backend metadata, the permission profile, and the platform coverage (Linux bwrap, macOS Seatbelt, Windows Restricted Token, WSL) all hang together. The architecture-baseline tests assert the right things: target backends per platform, command plan metadata propagation, fail-closed behavior when policy-only is disabled, and legacy adapter fallbacks. The Linux helper rewrite with Landlock, and the Windows ACL/network/process/token work, look like the right shape for the eventual v1.

Three blockers to address before this lands.

**1. The smoke script references a test that does not exist.** `scripts/sandbox-smoke.sh` includes `TestLegacySandboxEntrypointsAreExplicitAndExist` in the `go test ./internal/sandbox -run '...|...'` list, and the PR description lists it as one of the eight phase-0 tests. Greping the branch: that function is not defined anywhere — not in `internal/sandbox/architecture_baseline_test.go`, not in `adapters_test.go`, not in any other file. The other seven are real (`TestTargetBackendForPlatformBaseline`, `TestBackendPlanCarriesPhase0ManagerFields`, `TestCommandPlanCarriesSandboxMetadata`, `TestPolicyOnlyDisabledFailClosedForTargetPlatforms`, `TestSelectBackendChoosesPlatformAdapterWithFallback`, `TestBackendBuildPlanDocumentsBestEffortIsolation`, `TestBackendCapabilitiesReflectDisabledPolicy`). With `-run`, Go won't fail the script for a name that doesn't match any test, but the eight-test list in the PR body is misleading either way. Either add the test (an `internal/sandbox/adapters_test.go` assertion that the legacy `bubblewrap` / `sandbox-exec` helpers are still wired into the new architecture as documented in the "Notes" section) or drop it from the description and the smoke list.

**2. `sandbox.blockUnixSockets` and the `zero-seccomp` helper are removed with no migration path.** `internal/config/types.go` loses 7 lines (the `BlockUnixSockets` field), `internal/cli/exec.go` drops the policy application, the `cmd/zero-seccomp` binary is deleted, and the README section documenting the flag is removed. A user with `"sandbox": {"blockUnixSockets": true}` in their config will see the field silently dropped (no validation warning, no error), and the `zero-seccomp` helper they installed alongside their binary becomes a no-op. The change is a strict regression for anyone who relied on the AF_UNIX seccomp filter. At minimum, this needs a CHANGELOG / migration note in the PR body, and ideally a one-release deprecation where unknown fields are warned about in `zero doctor` so users notice. Same for the `cmd/zero-seccomp` binary — if it's gone, say so explicitly in the PR description so reviewers can flag the user impact.

**3. The branch needs to merge with the open `docs/agents-md-extension-guide` PR (#237), and the diff against `main` does not show a clean merge.** Both PRs modify `internal/agent/system_prompt.go` and `system_prompt_test.go`. The sandbox branch was rebased/merged onto current main locally (HEAD `04c4ab9 Merge main into sandbox architecture branch`), so the conflict is resolved in the branch — but the diff `gh pr diff 240` shows against the old base, so the conflict won't be visible in the GitHub UI. Whoever merges first wins; the other has to rebase. The two changes are compatible (`#237` adds case-insensitive path-walking on top of the old `findProjectGitRoot`; this PR swaps the body to use `hasGitMetadata` for the same function), so the merge is mechanical, but it would be worth a one-line note in the PR description so the next reviewer doesn't trip on it.

Non-blockers, flagged for the follow-ups:

- The backend renames are sweeping: `BackendBubblewrap` -> `BackendLinuxBwrap`, `BackendSandboxExec` -> `BackendMacOSSeatbelt`. These are exported constants in `internal/sandbox/types.go`, so anything outside this PR that referenced the old names breaks. There is no compatibility shim. If the renames are intentional and you want them now, fine — just call it out so downstream branches (e.g. `feat/m6-sandbox-backend`, `feat/m6-sandbox-execution-hardening`, `feat/sandbox-completion`, `feat/sandbox-ast-analyzer`) don't silently lose backend selection when this lands. If the renames can be deferred to a later phase, deferring them keeps this PR strictly additive.

- `internal/release/release.go` and `release_test.go` add 89 / 60 lines, but the diff snippet I'm reading doesn't show what they do. Worth a sentence in the PR body so a reviewer doesn't have to chase 60 lines of release logic to confirm the phase-0 scope.

- `internal/doctor/hardening.go` changes the sandbox check shape (the `Sandbox` field on `Options`, the `WorkspaceRoot` field, new check logic). I traced the call site at `internal/cli/observability.go:runDoctor` to confirm the new fields are populated from `resolved.Sandbox` and `workspaceRoot`, so the doctor integration looks correct. No change requested, just noting I checked.

The architecture is right; the missing test, the silent config drop, and the unannounced merge conflict with #237 are the things that should land before merge.

anandh8x:
Addressed the requested changes in `7aafbbc`:

1. Added `TestLegacySandboxEntrypointsAreExplicitAndExist` and included it in `scripts/sandbox-smoke.sh`.
2. Restored `sandbox.blockUnixSockets` as resolved config and policy, routed it into the Linux helper via `--block-unix-sockets`, kept the filter best-effort, and retained/staged `zero-seccomp` as a Linux compatibility wrapper.
3. Updated the PR body with the #237 overlap note and the release helper scope.

Also made CLI startup tests hermetic where they assert empty stderr, so default MCP registration does not perform real network work during `go test`.

Validation:
- `GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/cli ./internal/config ./internal/release ./cmd/zero-seccomp -run 'TestLegacySandboxEntrypointsAreExplicitAndExist|TestSandboxManagerBuildsCommandPlanThroughLinuxHelper|TestBuildLinuxSandboxCommandArgsSerializesPermissionProfile|TestParseLinuxSandboxHelperArgs|TestBuildLinuxSandboxBwrapArgsWrapsInnerSeccompStage|TestApplyConfiguredSandboxPolicyDiagnosticsFlags|TestResolveSandboxBlockUnixSocketsFromFile|TestCopyPackageFilesStagesLinuxSandboxHelper'`
- `GOCACHE=/tmp/zero-go-cache scripts/sandbox-smoke.sh`
- `git diff --cached --check`

Vasanthdev2004 (approved):
### Re-review: blockers addressed

Re-ran the verification against `origin/sandbox-architecture-baseline` at `7aafbbc`. The three blockers from the previous round are all closed:

**1. Missing legacy-entrypoint test.** `TestLegacySandboxEntrypointsAreExplicitAndExist` is now defined in `internal/sandbox/adapters_test.go`. The eight phase-0 tests listed in the PR body all exist; the smoke script's `go test -run '...|TestLegacySandboxEntrypointsAreExplicitAndExist|...'` invocation now actually exercises a real test function.

**2. `sandbox.blockUnixSockets` migration path.** Restored as a resolved config and policy field in `internal/config/types.go` and `internal/config/resolver.go` (the new `TestResolveSandboxBlockUnixSocketsFromFile` covers the resolution). The Linux helper now accepts `--block-unix-sockets` (`internal/sandbox/linux_helper.go`), and the `cmd/zero-seccomp` Linux compatibility wrapper is back as a real binary — the package builds and stages it as a Linux release artifact via the new `LinuxSeccompHelperArtifactName` / `buildLinuxSeccompHelper` paths in `internal/release/release.go`. Users with the existing `"sandbox": {"blockUnixSockets": true}` config keep the behavior; the helper or wrapper applies the filter, depending on which one is invoked. The PR body's "Notes" section calls out the compatibility path explicitly, which is exactly the deprecation note that was missing before.

**3. #237 merge overlap.** The PR body now states: "The branch has already merged current `main`; the overlapping `internal/agent/system_prompt.go` / `system_prompt_test.go` changes from #237 are compatible with this branch's git-root metadata validation." I re-confirmed the compatibility on the branch: the sandbox branch's `findProjectGitRoot` uses `hasGitMetadata` (the stricter `.git` validation), which composes cleanly with #237's case-insensitive path-walking on top of the same function. The merge is mechanical for whichever side merges second.

I also re-checked the non-blockers from the first round and they look better than I initially feared:

- **Backend renames.** `BackendBubblewrap` → `BackendLinuxBwrap` and `BackendSandboxExec` → `BackendMacOSSeatbelt` are real, and the string values also change (`bubblewrap` → `linux-bwrap`, `sandbox-exec` → `macos-seatbelt`), so this is a breaking change. I worried that the four open downstream branches (`feat/m6-sandbox-backend`, `feat/m6-sandbox-execution-hardening`, `feat/sandbox-completion`, `feat/sandbox-ast-analyzer`) would have to be reworked, but grepping each branch's tip shows none of them reference the old constant names — they all reference the new ones. The rename is safe in practice; no compat shim needed.

- **`internal/release/release.go` additions.** The +111 lines are now justified: the file gains `LinuxSeccompHelperArtifactName`, `buildLinuxSeccompHelper`, and the staging logic that puts both `zero-linux-sandbox` and `zero-seccomp` into Linux release artifacts. With the seccomp wrapper restored as a compatibility binary, the release packaging has to know about it, so the additions are the right shape.

- **`internal/doctor/hardening.go`.** The new sandbox check shape is unchanged from before; my earlier trace of `internal/cli/observability.go:runDoctor` to confirm the `Options.Sandbox` and `WorkspaceRoot` plumbing is still correct, and `internal/doctor/hardening_test.go` is updated accordingly.

One small follow-up worth noting (not blocking): the `internal/cli/app_test.go` change in `7aafbbc` stubs MCP registration to keep CLI startup tests hermetic. That is the right move, but it means the test is no longer asserting the real MCP registration path. Worth a comment in that test file pointing at the new stub location so the next person doesn't go hunting for the original assertion.

Overall: the architecture is right, the blockers are addressed, the compatibility story is documented, and the merge with #237 is the right shape. Approving.

PR #241, opened 2026-06-17
perf(tools): lower default deferThreshold to stop eager MCP-schema token waste

gnanam1990:
## Problem

A short message (even "hi") was costing ~20k input tokens for users with MCP servers. Root cause traced through the request-assembly path:

MCP tool schemas are only collapsed into compact `tool_search` reminder lines once the deferred-eligible tool count reaches `tools.deferThreshold`. The default was **10**, so a typical single MCP server (Exa-style, a handful of tools) sat **below** the threshold — and every full MCP schema (**300–600 tokens each**) was re-sent on **every turn**, regardless of message length. A small MCP set therefore added several thousand input tokens per message, paid again on every turn.

The non-MCP floor is modest by comparison (measured via `zero context` on a large repo): system prompt ~2.8k + 12 core tool schemas ~1.9k + capped workspace map ~1k + guidelines. The eager MCP schemas were the runaway component.

## Fix

Lower `defaultDeferThreshold` from **10 → 3**, so a normal MCP server's toolset defers and stays cheap (~30-token reminder line per tool instead of a full schema).

**Capability is unchanged.** Every tool is still reachable — the model loads a deferred tool's schema via `tool_search` before its first use, after which it's loaded for the rest of the run. The only cost is one `tool_search` round-trip per deferred tool's first use, versus paying its full schema on *every* turn.

Fully overridable per config: `tools.deferThreshold` (set `0` to keep the old always-eager behavior, e.g. for a model without `tool_search`).

## Example saving
A 5-tool MCP server: ~2k tokens of eager schemas every turn → ~150 tokens of reminder lines. ~1.8k saved per message, compounding across a conversation.

## Tests
- `TestDefaultDeferThresholdStaysLowForTokenEconomy` — regression guard so the default can't silently drift back up.
- Existing deferral/partition tests unchanged and green (they set explicit thresholds).

Gate: gofmt, vet, build, staticcheck, `-race` on `config` + `agent`.

## Summary by CodeRabbit

* **Chores**
  * Lowered the default deferred-eligible tool deferral threshold (now **3** instead of **10**) and expanded configuration guidance for deferred tool schema compaction.

* **Refactor**
  * Improved token-count estimation used for proactive compaction and transcript/context sizing for more consistent behavior.

* **Tests**
  * Added a regression test to keep the default defer threshold low.
  * Added a measurement-based test verifying reduced tool payload size with the lower threshold.
  * Added unit coverage for the updated token estimation heuristic.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -77,17 +77,38 @@ type CompactionResult struct {
 // ~1k tokens is a representative mid-range cost across providers.
 const imageTokenEstimate = 1000
 
-// estimateTokens is a cheap, dependency-free token estimate (~4 chars/token)
+// ApproxTextTokens estimates the token count of text WITHOUT a real tokenizer.
+// The BPE tokenizers zero targets fold a run's leading space into the following
+// token rather than emitting whitespace as its own token, so naive len/4
+// overcounts real text by ~15-20% (measured: a 24.8k-char prompt counted 5.2k
+// real tokens where len/4 said 6.2k). Counting NON-whitespace bytes / 4 tracks
+// the provider's actual count closely (validated against live usage) while
+// staying allocation- and dependency-free. zero still receives the exact count
+// back as usage on every request; this estimate is only for the pre-request
+// context budget preview and the compaction threshold.
+func ApproxTextTokens(value string) int {
+	nonSpace := 0
+	for i := 0; i < len(value); i++ {
+		switch value[i] {
+		case ' ', '\t', '\n', '\r', '\f', '\v':
+		default:
+			nonSpace++
+		}
+	}
+	return nonSpace / 4
+}
+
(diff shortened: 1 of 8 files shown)


Vasanthdev2004 (approved):
LGTM. The threshold drop and token-economy framing are reasonable, and the new test in deferred_loop_test.go uses the existing partitionTools and fakeToolSearchTool helpers (no new build surface). The deferral logic stays race-free because partitionTools is single-goroutine and the tool_search loader remains a non-deferred built-in, so it's always available when deferral is on. No blockers.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable ?

anandh8x (approved):

Reviewed the defer-threshold default change and token estimation paths. git diff --check passed for this PR. I did not run the full branch test suite.

PR #242, opened 2026-06-18
Align web tool permission prompts

anandh8x:
## Summary
- treat hosted web_search as an allowed read-only network discovery tool so normal searches do not stop for approval
- add host-scoped persistent grants for web_fetch URLs so always-allow is bounded to the requested host
- update permission prompt/transcript rendering to show network targets and preserve scope metadata through caching and session hydration

## Tests

## Release Notes

* **New Features**
  - Host-scoped permissions for network operations enable granular per-host access control
  - Enhanced permission prompts now display specific targets and offer "always for this host" options
  - Permission scope information is now preserved across transcripts and session resumption

* **Bug Fixes**
  - Improved permission scope handling and matching for network tools

* **Documentation**
  - Clarified permission metadata and scope behavior descriptions

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1017,11 +1017,11 @@ func deniedPermissionResult(call ToolCall, reason string, requestEvent Permissio
 }
 
 // permissionScope returns a concise, human-readable description of what a tool
-// call will actually touch — the file path, directory, or working dir lifted
-// from its arguments — so the permission card and the persisted decision can
-// show the user exactly what "allow" covers. It is empty when the tool exposes
-// no path-like argument (the grant is then plainly tool-wide). The first
-// matching key wins, ordered most-specific (a concrete file) first.
+// call will actually touch -- a file path, directory, working dir, or network
+// host lifted from its arguments -- so the permission card and persisted
+// decision can show the user exactly what "allow" covers. It is empty when the
+// tool exposes no scoped argument (the grant is then plainly tool-wide). The
+// first matching key wins, ordered most-specific (a concrete file) first.
 func permissionScope(toolName string, args map[string]any) string {
 	// sandbox.DeriveScope is the single source of truth for which arguments carry
 	// a scope, shared with grant persistence and matching so the card display can
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -315,6 +315,37 @@ func TestRunAdvertisesWebFetchInAutoMode(t *testing.T) {
 	}
 }
 
+func TestRunAdvertisesAllowedWebSearchInAutoMode(t *testing.T) {
+	t.Setenv("ZERO_WEBSEARCH_BASE_URL", "https://search.example/api")
+	registry := tools.NewRegistry()
+	for _, tool := range tools.CoreNetworkTools() {
+		if tool.Name() == "web_search" {
+			registry.Register(tool)
+		}
+	}
+	provider := &mockProvider{
+		turns: [][]zeroruntime.StreamEvent{{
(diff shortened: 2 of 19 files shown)


gnanam1990 (approved):
### Review: web tool permission alignment

Reviewed the security-sensitive parts closely (grant scoping, the `web_search` auto-allow, scope-through-cache/hydration) and ran build/test/staticcheck on the branch. On net this **tightens** the sandbox, and both headline changes hold up. Approving.

## Security — verified sound

**`web_search` auto-allow does not bypass the network policy.** `web_search.go` flips to `Permission: PermissionAllow`, but in-process network tools were never subject to the sandboxed-shell's `Network: deny` gate in the first place — that gate is for the shell (`engine.go` `effectiveNetworkMode`; see the `NetworkHostAllowed` comment). The opt-in `EnforceToolNetwork` path still holds `web_search` to the policy: `RunWithSandbox` still calls `enforceScopedNetworkPolicy(engine, host)`, and `sameHostRedirectPolicy` still fails closed on cross-host redirects. The egress target is operator-fixed — `web_search`'s schema exposes only `query`/`limit`/`domains` (no host/URL), so the model can't redirect it. Net effect: removes a prompt, keeps enforcement.

**Host-scoped `web_fetch` grants are matched exactly — and this *narrows* the previously tool-wide grant.** `grantCovers`'s `ScopeHost` case is `normalizeHostScope(reqScope) == normalizeHostScope(grant.Scope)` — exact equality, no prefix/suffix. I exercised the confusion vectors: a grant for `example.com` correctly refuses `api.example.com` (subdomain), `example.com.attacker.net` (sibling suffix), and the userinfo trick `https://example.com@evil.com` (`url.Hostname()` → `evil.com`, not covered). Case-/port-/trailing-dot-/IPv6-bracket-normalized. No over-broad match.

**No scope confusion through render cache or session hydration.** The render-cache fingerprint keys on `ToolCallID` + the new `Scope` + `SideEffect`, and the session round-trips `Scope`/`SideEffect`, so a cached or rehydrated card can't apply host A's grant to host B.

## Nits (optional, non-blocking)

1. **Hostless/malformed URL falls back to a tool-wide `web_fetch` grant.** If `web_fetch` gets a URL that parses but yields `Host == ""` (e.g. `file:///x`), `DeriveScope` returns `ScopeToolWide`, so "always allow" persists a tool-wide grant. This matches pre-PR behavior (and the "always for this host" label won't render without a host), so it's not a regression — but for a network tool, fail-closed (re-prompt) would be safer than persisting tool-wide. Worth a follow-up.
2. **A couple of negative tests would lock in the security crux.** Host scoping is well-covered for subdomains/case; I'd add the userinfo-confusion URL (`https://a@evil.com` resolves to `evil.com` and is refused), the hostless-URL → tool-wide fallback (documents the intent), and an IPv6/punycode negative match.

## Overlap with another in-flight branch
This modifies `internal/agent/loop.go`, `internal/tools/types.go`, `internal/agent/types.go`, and `internal/tui/model.go`, which a separate open branch (specialist delegation, #243) also touches. #242's edits to those files are localized (mostly doc comments + reuse), so a textual conflict is unlikely, but `types.go`/`loop.go` will want a 3-way merge — whichever lands second should re-run the sandbox/agent/tui `-race` suite. No semantic conflict; the two touch different concerns.

## Tooling
`go build ./...`, `go vet ./...`, and `go test ./internal/sandbox/... ./internal/tools/... ./internal/agent/... ./internal/tui/... -race` all pass; staticcheck is clean for the files this PR touches (the few findings are pre-existing in untouched files).

Careful, well-tested change — approving.

Vasanthdev2004 (approved):

PR #244, opened 2026-06-18
Align permission session approvals

anandh8x:
## Summary
- Add in-memory session approvals alongside persistent sandbox grants.
- Auto-allow built-in workspace file mutations when sandbox policy permits the target, while keeping outside-workspace writes denied.
- Send dynamic permission choices to the TUI for once, session, future, and deny decisions, with matching stream/session events.

## Tests

## Release Notes

* **New Features**
  * Added session-scoped permission grants: approve permissions for the current session only without storing persistent grants.
  * Permission prompts now dynamically display available decision options based on your sandbox configuration.

* **Improvements**
  * Workspace file mutations automatically allow in sandbox-protected environments.
  * Enhanced permission decision tracking and display in session history.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -558,23 +558,36 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 	}
 
 	decisionReason := ""
+	var decisionAction PermissionDecisionAction
 	if toolFound && options.OnPermissionRequest != nil && shouldRequestPermission(tool, permissionGranted, preflightDecision) {
 		requestEvent, ok := buildPermissionEvent(call, tool, args, permissionGranted, permissionMode, options, preflightDecision)
 		if !ok {
 			requestEvent = fallbackPermissionEvent(call, tool, args, permissionMode, options)
 		}
-		request := permissionRequestFromEvent(requestEvent, args)
+		request := permissionRequestFromEvent(requestEvent, args, options)
 		decision, err := requestPermission(ctx, request, options)
 		if err != nil {
 			decision = PermissionDecision{Action: PermissionDecisionDeny, Reason: err.Error()}
 		}
 		decision.Action = normalizePermissionDecisionAction(decision.Action)
 		decisionReason = strings.TrimSpace(decision.Reason)
+		decisionAction = decision.Action
 		switch decision.Action {
 		case PermissionDecisionAllow:
 			permissionGranted = true
+			requestEvent.DecisionAction = decision.Action
+		case PermissionDecisionAllowForSession:
+			permissionGranted = true
+			requestEvent.DecisionAction = decision.Action
+			if options.Sandbox != nil {
+				if grant, err := persistSessionPermissionGrant(call.Name, args, decisionReason, options); err == nil {
+					requestEvent.GrantMatched = true
+					requestEvent.Grant = &grant
+				}
+			}
 		case PermissionDecisionAlwaysAllow:
 			permissionGranted = true
+			requestEvent.DecisionAction = decision.Action
 			// "Always allow" also persists a grant so FUTURE calls skip the prompt.
 			// With no sandbox engine there is nowhere to persist it, and persistence
 			// can also fail — neither must deny what the user explicitly allowed, so
@@ -584,7 +597,10 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 			// (buildPermissionEvent reads decision.GrantMatched/Grant), so the
 			// persisted grant does not need to be recorded on requestEvent here.
 			if options.Sandbox != nil {
-				_, _ = persistPermissionGrant(call.Name, args, decisionReason, options)
+				if grant, err := persistPermissionGrant(call.Name, args, decisionReason, options); err == nil {
+					requestEvent.GrantMatched = true
+					requestEvent.Grant = &grant
+				}
 			}
 		default:
 			emitDeniedPermission(options, call, requestEvent, decisionReason)
@@ -624,6 +640,9 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 	if toolFound && options.OnPermission != nil {
 		if event, ok := buildPermissionEvent(call, tool, args, permissionGranted, permissionMode, options, sandboxDecision); ok {
 			event.DecisionReason = decisionReason
+			if decisionAction != "" {
+				event.DecisionAction = decisionAction
+			}
 			options.OnPermission(event)
 		}
 	}
@@ -931,7 +950,7 @@ func requestPermission(ctx context.Context, request PermissionRequest, options O
 
 func normalizePermissionDecisionAction(action PermissionDecisionAction) PermissionDecisionAction {
 	switch action {
-	case PermissionDecisionAllow, PermissionDecisionAlwaysAllow:
+	case PermissionDecisionAllow, PermissionDecisionAllowForSession, PermissionDecisionAlwaysAllow:
 		return action
 	default:
 		return PermissionDecisionDeny
@@ -963,6 +982,28 @@ func persistPermissionGrant(toolName string, args map[string]any, reason string,
 	})
 }
 
+func persistSessionPermissionGrant(toolName string, args map[string]any, reason string, options Options) (sandbox.Grant, error) {
+	if options.Sandbox == nil {
+		return sandbox.Grant{}, errors.New("sandbox engine is not configured")
+	}
+	maxAutonomy := sandbox.Autonomy(options.Autonomy)
+	if maxAutonomy == "" {
+		maxAutonomy = sandbox.AutonomyMedium
+	}
+	if normalized, err := sandbox.NormalizeAutonomy(maxAutonomy); err == nil {
+		maxAutonomy = normalized
+	}
+	scope, kind := sandbox.DeriveScope(toolName, args)
+	return options.Sandbox.GrantForSession(sandbox.GrantInput{
+		ToolName:    toolName,
+		Decision:    sandbox.GrantAllow,
+		MaxAutonomy: maxAutonomy,
+		Reason:      reason,
+		Scope:       scope,
+		ScopeKind:   kind,
+	})
+}
+
 func emitDeniedPermission(options Options, call ToolCall, requestEvent PermissionEvent, reason string) {
 	if options.OnPermission == nil {
 		return
@@ -975,6 +1016,7 @@ func emitDeniedPermission(options Options, call ToolCall, requestEvent Permissio
 	event.ToolCallID = call.ID
 	event.ToolName = call.Name
 	event.Action = PermissionActionDeny
+	event.DecisionAction = PermissionDecisionDeny
 	event.PermissionGranted = false
 	event.DecisionReason = reason
 	options.OnPermission(event)
@@ -990,6 +1032,7 @@ func deniedPermissionResult(call ToolCall, reason string, requestEvent Permissio
 	}
 	event := requestEvent
 	event.Action = PermissionActionDeny
+	event.DecisionAction = PermissionDecisionDeny
 	event.PermissionGranted = false
 	event.DecisionReason = reason
 	if event.Risk.Level == "" {
@@ -1098,6 +1141,7 @@ func buildPermissionEvent(call ToolCall, tool tools.Tool, args map[string]any, p
 		ToolCallID:        call.ID,
 		ToolName:          call.Name,
 		Action:            action,
+		DecisionAction:    grantDecisionAction(grant),
 		Permission:        string(safety.Permission),
 		PermissionGranted: permissionGranted,
 		PermissionMode:    permissionMode,
@@ -1112,28 +1156,57 @@ func buildPermissionEvent(call ToolCall, tool tools.Tool, args map[string]any, p
 	}, true
 }
 
+func grantDecisionAction(grant *sandbox.Grant) PermissionDecisionAction {
+	if grant == nil {
+		return ""
+	}
+	if grant.Session {
+		return PermissionDecisionAllowForSession
+	}
+	if grant.Decision == sandbox.GrantAllow {
+		return PermissionDecisionAlwaysAllow
+	}
+	return ""
+}
+
 func fallbackPermissionEvent(call ToolCall, tool tools.Tool, args map[string]any, permissionMode PermissionMode, options Options) PermissionEvent {
 	event, _ := buildPermissionEvent(call, tool, args, false, permissionMode, options, nil)
 	return event
 }
 
-func permissionRequestFromEvent(event PermissionEvent, args map[string]any) PermissionRequest {
+func permissionRequestFromEvent(event PermissionEvent, args map[string]any, options Options) PermissionRequest {
 	return PermissionRequest{
-		ToolCallID:     event.ToolCallID,
-		ToolName:       event.ToolName,
-		Action:         event.Action,
-		Permission:     event.Permission,
-		PermissionMode: event.PermissionMode,
-		Autonomy:       event.Autonomy,
-		SideEffect:     event.SideEffect,
-		Reason:         event.Reason,
-		Scope:          event.Scope,
-		Risk:           event.Risk,
-		Args:           cloneArgs(args),
-		Violation:      event.Violation,
-		GrantMatched:   event.GrantMatched,
-		Grant:          event.Grant,
+		ToolCallID:         event.ToolCallID,
+		ToolName:           event.ToolName,
+		Action:             event.Action,
+		Permission:         event.Permission,
+		PermissionMode:     event.PermissionMode,
+		Autonomy:           event.Autonomy,
+		SideEffect:         event.SideEffect,
+		Reason:             event.Reason,
+		Scope:              event.Scope,
+		Risk:               event.Risk,
+		Args:               cloneArgs(args),
+		Violation:          event.Violation,
+		GrantMatched:       event.GrantMatched,
+		Grant:              event.Grant,
+		AvailableDecisions: availablePermissionDecisions(event, options),
 	}
+}
+
+func availablePermissionDecisions(event PermissionEvent, options Options) []PermissionDecisionAction {
+	if event.Action != PermissionActionPrompt {
+		return nil
+	}
+	decisions := []PermissionDecisionAction{PermissionDecisionAllow}
+	if options.Sandbox != nil {
+		decisions = append(decisions, PermissionDecisionAllowForSession)
+		if options.Sandbox.CanPersistGrants() {
+			decisions = append(decisions, PermissionDecisionAlwaysAllow)
+		}
+	}
+	decisions = append(decisions, PermissionDecisionDeny)
+	return decisions
 }
 
 func cloneArgs(args map[string]any) map[string]any {
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -792,6 +792,55 @@ func TestRunRequestsPromptToolPermissionBeforeExecution(t *testing.T) {
 	}
 }
 
+func TestRunAllowsWorkspaceWriteWithoutPromptWhenSandboxPolicyPermits(t *testing.T) {
+	root := t.TempDir()
+	registry := tools.NewRegistry()
+	registry.Register(tools.NewWriteFileTool(root))
+	provider := providerCallingWriteFileThenAnswer("write done")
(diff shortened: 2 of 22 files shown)


gnanam1990 (changes requested):
### Review: permission session approvals

The session-approval machinery and dynamic-choice plumbing are well-built and correctly scoped — but the new in-workspace write auto-allow has one **High** protection regression I traced and confirmed in the code, so requesting changes on that (plus test gaps). Everything else is solid.

## Security

**[HIGH — confirmed] In-workspace auto-allow silently approves `.git/hooks/*` and `.zero/*` writes.** (`engine.go` `workspaceWriteAutoAllowed` + the Evaluate branch)
I traced it end to end:
- `DefaultPolicy()` (types.go) ships `EnforceWorkspace: true` and an **empty `DenyWrite`**.
- For an in-workspace write, `validatePathWithPolicy` only does `scope.validate(path)` (workspace bounds + symlink resolution) — it does **not** consult `protectedMetadataNames`.
- The `.git`/`.zero`/`.agents` protected-metadata defense (`profile.go`) is wired **only into the backend layers** (Landlock / Windows ACL / bwrap) that sandbox *shell* commands. The built-in `write_file`/`edit_file` run **in-process** and are gated solely by `Evaluate`, which has no protected-metadata check.
- `workspaceWriteAutoAllowed` then returns `true` for `write_file` → `<ws>/.git/hooks/pre-commit` → `Action=allow`, no prompt.

**Impact:** a prompt-injected model can drop a `pre-commit`/`post-checkout` hook (or rewrite `.zero/config.json`) with **no user approval** → arbitrary code execution on the next git operation. Pre-PR this hit the prompt; this PR makes it silent. **Fix:** exclude `protectedMetadataNames` from the auto-allow (re-prompt when the target is under `.git`/`.zero`/`.agents`), or add them to a default `DenyWrite`.

**[MEDIUM — confirmed] `apply_patch` auto-allow is path-blind.** `workspaceWriteAutoAllowed` has `case "apply_patch": return true` — unconditionally, with no path check, because the patch targets live in the `patch` body which `requestPaths` doesn't read. So the **sandbox** (the authoritative boundary) auto-allows any `apply_patch` regardless of target. Not independently exploitable today — the `apply_patch` *tool* re-validates and blocks `../` escapes — but the boundary is trusting the tool to police itself; any future weakening of the tool check silently re-opens it. Worth deriving the patch's targets and validating them like the other two tools.

**Verified fail-closed (good):** outside-workspace absolute writes, `../` traversal, and **inside→outside symlink** writes are all denied *before* auto-allow (`scope.validate` resolves segments). Auto-allow fires only for the three built-in file tools — bash/MCP/custom writers don't match. Network `deny` and the Deny lists still run earlier in `Evaluate`, and the autonomy ceiling is respected.

## Correctness (solid)
- **Session grants** are keyed `tool → []Grant`, matched by scope+kind+autonomy via the shared lookup, anchored to absolute via the same `DeriveScope`/normalize path as persistent grants (no new over-broadening), stored on the Engine (per-process/per-workspace, **never persisted**), and **mutex-guarded** (race clean).
- **Deny action-reporting fix** is correct (denied events now set `DecisionAction = Deny`; unknown actions default to Deny — fail-closed).
- **Dynamic choices** only offer session/always when a store exists (`always` gated on `CanPersistGrants()`); options/offsets/cursor derive from one `permissionOptions(request)` call, so a click/hotkey can't land on the wrong choice. Audit trail preserved (auto-allowed writes still emit one `PermissionEvent`).

## Tests (gaps — please add)
1. No `.git/hooks` / `.zero` negative — would have caught the High finding.
2. No `apply_patch` escaping-patch-body negative at the sandbox layer.
3. No session-grant **cross-file** negative (grant for file A must not match file B).
4. Symlink-escape isn't exercised *through* the auto-allow path (only at the lower `validateWorkspacePath`).

## Overlap with other branches
- **#242 is the base of this PR (already merged).** This PR is one commit on top; the web_search/host-scope items in the description are #242's, not new here — no stacking action needed.
- A separate in-flight branch (specialist delegation, #243) touches `loop.go`/`types.go`/`model.go` too, but in **non-overlapping regions** (a `merge-tree` was clean). Low merge risk — minor adjacent-hunk friction in `loop.go` at worst; whichever lands second should re-run the agent/sandbox/tui `-race` suite.

## Gate
`go build` · `go vet` · `go test ./internal/{sandbox,agent,tui}/... -race` all pass; staticcheck clean for the touched files. (The 3 `cli` `-race` failures are the pre-existing `kimi-coding`-provider env issue — they fail identically on the base commit, unrelated to this PR.)

Strong PR overall — just the `.git`/`.zero` auto-allow needs to re-prompt before this lands.

Vasanthdev2004 (changes requested):
## REQUEST_CHANGES

Read at `ae17cf0`. Walked the 21 files (584 +, 138 -) and re-verified gnanam1990's findings against the branch — agree with both severities and the test gaps, plus I have a couple of follow-ups for the rest of the surface.

## Confirming gnanam's blockers

**HIGH — in-workspace auto-allow silently approves writes under `.git`/`.zero`/`.agents`.**

Traced end to end. `workspaceWriteAutoAllowed` in `internal/sandbox/engine.go:460-472` returns `true` for `write_file`/`edit_file` whenever `requestPaths(request)` is non-empty — no metadata-name check. The only metadata-name defense, `ProtectedMetadataNames` (`.git`, `.zero`, `.agents`), is set in `internal/sandbox/profile.go:45` and consumed **only** by the OS-sandbox backends (`runner.go:859`, `landlock_linux.go:64`, `linux_helper.go:240`, `windows_acl.go:105`). The engine's `Evaluate` path in `engine.go` does not consult it. So a `write_file` to `<ws>/.git/hooks/pre-commit` (or `<ws>/.zero/config.json`) gets `Action=allow` with no prompt, no audit-prompt handshake, and the model can drop a `pre-commit`/`post-checkout` hook that runs arbitrary code on the next `git` operation. Pre-PR this hit the prompt UI; this PR makes it silent.

The OS-sandbox backends catch it for shell commands, but built-in file tools run in-process and the engine is the only gate. Fix has to live in the engine, not the OS layers: in `workspaceWriteAutoAllowed` (or just after the `requestPaths` loop in `Evaluate`), short-circuit to `ActionPrompt` when the path is under any name in `protectedMetadataNames`. Same applies to the `apply_patch` path — see below.

**MEDIUM — `apply_patch` auto-allow is path-blind.**

`workspaceWriteAutoAllowed` returns `true` for `apply_patch` unconditionally (line 467-468). The patch targets live in the `patch` body, which `requestPaths` in `internal/sandbox/risk.go:223` does not look at. So the engine auto-allows any in-workspace `apply_patch` regardless of target. The apply_patch tool's own `validatePatchPaths` in `internal/tools/apply_patch.go` does workspace confinement only (no `..`, no absolute paths, `resolveWorkspaceTargetPath`) — it does not consult `DenyWrite`/`AllowWrite` or `protectedMetadataNames`. Not exploitable through the OS sandbox for shell-driven writes, but the engine is trusting the tool to police itself. Easy fix: extract patch-header paths and route them through `validatePathWithPolicy` like the other two tools do. Worth doing in the same pass as the HIGH fix.

## Confirming the test gaps

`TestRunAllowsWorkspaceWriteWithoutPromptWhenSandboxPolicyPermits` (loop_test.go:795) writes to `notes.txt` in a clean temp dir with `DefaultPolicy()` — no `DenyWrite`, no `protectedMetadataNames` exercise. `TestRunSessionAllowSkipsMatchingPromptWithoutPersistentGrant` (loop_test.go:964) writes to the *same* `notes.txt` twice, so it never exercises a session grant for file A failing to match file B. None of the new tests cover a symlink escape *through* the auto-allow path. Adding four tests:

1. in-workspace `write_file` to `<ws>/.git/hooks/pre-commit` → must prompt, must not auto-allow
2. in-workspace `apply_patch` whose patch body targets a `DenyWrite` path → must deny (or prompt, depending on the fix shape)
3. session grant for `notes.txt` must not match a second call writing to `other.txt`
4. symlink inside workspace pointing to outside → must still deny at the auto-allow path

## Design (solid, no change needed)

- Session grants keyed `tool → []Grant`, matched by scope+kind+autonomy via the shared `lookupGrantBucket`, anchored to absolute via the same `DeriveScope`/normalize path as persistent grants. Mutex-guarded (race clean). Never persisted.
- Deny action-reporting fix is correct: denied events set `DecisionAction = Deny`; unknown actions default to `Deny` (fail-closed).
- Dynamic choices only offer session/always when a store exists (`always` gated on `CanPersistGrants()`). One `permissionOptions(request)` call drives options + cursor + offsets, so a click and a hotkey can't land on different choices.
- Verified pass: outside-workspace absolute writes, `../` traversal, inside→outside symlink writes are all denied *before* `workspaceWriteAutoAllowed` fires. Auto-allow only matches the three built-in file tools. Network deny + Deny lists still run earlier in `Evaluate`. Autonomy ceiling respected.

## Small follow-ups (non-blocking)

- `permissionOptions` in `internal/tui/permission_prompt.go:21-30` has a fallback that hardcodes all four decisions when `AvailableDecisions` is empty. The TUI is only invoked for prompts that always populate `AvailableDecisions`, so the fallback is unreachable today. Worth deriving the fallback from the presence of a store so a future call site can't offer "always" with nothing to persist to.
- `persistSessionPermissionGrant` swallows errors with `if ... err == nil` (`loop.go:583`). That's the right call (don't deny what the user explicitly allowed), but the asymmetry with the `AlwaysAllow` branch (which records the persisted grant on the event) means the session-grant event's `GrantMatched` flips to `false` silently if persistence fails. Worth a one-line comment so a future reader doesn't think the event is broken.

## Gate

Re-ran the agent/sandbox/TUI `go test` + `go vet` on the branch — same baseline as gnanam. The HIGH is the one to fix before this lands; the rest is follow-up.

Vasanthdev2004:
Verdict: **REQUEST_CHANGES**.

Agreeing with gnanam1990's review and adding my own trace. The short version:

**HIGH — in-workspace auto-allow approves writes under `.git`/`.zero`/`.agents`.**
`workspaceWriteAutoAllowed` in `internal/sandbox/engine.go:460-472` returns `true` for `write_file`/`edit_file` whenever `requestPaths(request)` is non-empty, with no metadata-name check. `ProtectedMetadataNames` is set in `internal/sandbox/profile.go:45` and consumed only by the OS-sandbox backends (`runner.go:859`, `landlock_linux.go:64`, `linux_helper.go:240`, `windows_acl.go:105`). The engine's `Evaluate` does not consult it. So a `write_file` to `<ws>/.git/hooks/pre-commit` (or `<ws>/.zero/config.json`) gets `Action=allow` with no prompt. A prompt-injected model can drop a `pre-commit` hook and get arbitrary code execution on the next `git` operation. The OS backends catch this for shell commands, but built-in file tools run in-process and the engine is the only gate. Fix has to live in the engine: short-circuit `workspaceWriteAutoAllowed` (or the post-`requestPaths` block in `Evaluate`) to `ActionPrompt` when the path is under any name in `protectedMetadataNames`.

**MEDIUM — `apply_patch` auto-allow is path-blind.**
`workspaceWriteAutoAllowed` returns `true` for `apply_patch` unconditionally (line 467-468). Patch targets live in the `patch` body, which `requestPaths` does not read. The apply_patch tool's own `validatePatchPaths` does workspace confinement only, not `DenyWrite`/`AllowWrite`/`protectedMetadataNames`. Worth extracting patch-header paths and routing them through `validatePathWithPolicy` like the other two tools do — fix in the same pass as the HIGH.

**Tests gaps to close:**
1. in-workspace `write_file` to `<ws>/.git/hooks/pre-commit` → must prompt
2. in-workspace `apply_patch` whose patch body targets a `DenyWrite` path → must deny or prompt
3. session grant for `notes.txt` must not match a second call writing to `other.txt`
4. symlink inside workspace pointing to outside → must still deny at the auto-allow path

**What I confirmed is solid (no change needed):**
- Session grants keyed `tool → []Grant`, scope+kind+autonomy matched via the shared `lookupGrantBucket`, anchored to absolute via the same `DeriveScope`/normalize path as persistent grants, mutex-guarded, never persisted.
- Deny action-reporting fix is correct.
- Dynamic choices only offer session/always when a store exists (`always` gated on `CanPersistGrants()`); single `permissionOptions(request)` call drives options + cursor + offsets.
- Outside-workspace absolute writes, `../` traversal, inside→outside symlink writes are all denied *before* `workspaceWriteAutoAllowed` fires. Auto-allow only matches the three built-in file tools. Network deny + Deny lists still run earlier in `Evaluate`. Autonomy ceiling respected.

Non-blocking follow-ups: TUI `permissionOptions` fallback offers "always" even with no store (defensive, currently unreachable); `persistSessionPermissionGrant` swallows errors silently so the event's `GrantMatched` flips to false on persistence failure (right call but worth a one-line comment).

The HIGH is the one to fix before this lands. The MEDIUM should ride with it. The rest is follow-up.

anandh8x:
Fixed the requested sandbox auto-allow blockers in cfbb48e.\n\n- Protected workspace metadata targets under .git, .zero, and .agents now re-prompt instead of being silently workspace-auto-allowed.\n- apply_patch targets are derived from patch headers inside the sandbox evaluation path, so patch-body escapes and symlink traversal are denied before auto-allow.\n- Added regression coverage for protected metadata writes, apply_patch path escapes, session grant cross-file mismatch, and symlink traversal through the auto-allow path.\n\nValidation:\n- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox -run 'TestEngineAutoAllowsWorkspaceFileMutationTools|TestEngineDoesNotAutoAllowProtectedMetadataWrites|TestEngineDeniesApplyPatchEscapesFromPatchBody|TestEngineSessionGrantDoesNotMatchSiblingFile|TestEngineDeniesAutoAllowedSymlinkEscape|TestEngineDeniesWorkspaceSymlinkTraversal|TestDeriveScope'\n- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/agent ./internal/tui ./internal/tools\n- GOCACHE=/tmp/zero-go-cache go test ./...\n- git diff --check

gnanam1990 (approved):
### Re-review (after `cfbb48e` "Guard workspace write auto-allow")

Both findings from the previous review are resolved — I re-verified each empirically against the updated branch:

**HIGH (protected-metadata auto-allow) — fixed.** `workspaceWriteAutoAllowed` now bails when `requestPathsTouchProtectedMetadata` matches. Verified:
- `write_file` → `<ws>/.git/hooks/pre-commit` → `Action=prompt` (was `allow`)
- `write_file` → `<ws>/.zero/config.json` → `Action=prompt`
- `write_file` → `<ws>/src/main.go` → `Action=allow` (the feature still works for normal writes)

**MEDIUM (apply_patch path-blind) — fixed.** apply_patch targets are now derived from the patch body (`applyPatchRequestPaths`/`patchHeaderPaths`) and validated by the new `applyPatchPathViolation` guard *before* auto-allow. Verified:
- patch targeting `../../etc/passwd` → `Action=deny` (the sandbox itself blocks it now, not just the tool)
- patch targeting `.git/hooks/pre-commit` → `Action=prompt`
- patch targeting `src/main.go` → `Action=allow`

And the four test gaps I flagged are all covered now: `TestEngineDoesNotAutoAllowProtectedMetadataWrites`, `TestEngineDeniesApplyPatchEscapesFromPatchBody`, `TestEngineSessionGrantDoesNotMatchSiblingFile`, `TestEngineDeniesAutoAllowedSymlinkEscape`.

`go build ./...`, `go vet`, and `go test ./internal/sandbox/... ./internal/agent/... -race` all pass on the branch. Thorough turnaround — approving.

Vasanthdev2004 (approved):
## APPROVE

Re-read at `cfbb48e` ("Guard workspace write auto-allow"). Both blockers from my prior review are addressed end-to-end, and all four test gaps are closed.

**HIGH — protected-metadata bypass.** Fixed. `internal/sandbox/engine.go:294-297` calls `applyPatchPathViolation` before the path loop, and `internal/sandbox/risk.go:236-238` extends `requestPaths` to fold in `applyPatchRequestPaths` (the patch-header extractor in `risk.go:242-...`). The new `workspaceWriteAutoAllowed` in `engine.go:464-478` then runs `requestPathsTouchProtectedMetadata` against the union, so an `apply_patch` whose patch body targets `.git/hooks/pre-commit`, `.zero/config.json`, or `.agents/...` returns false from the auto-allow and falls back to the prompt. Confirmed in `TestEngineDoesNotAutoAllowProtectedMetadataWrites` (engine_test.go:228-254) — all three sub-cases assert `Action=ActionPrompt` and `AutoAllowed=false`.

**MEDIUM — `apply_patch` path-blind.** Fixed. `applyPatchPathViolation` (risk.go:261-282) extracts patch headers, then `+++ b/../escape.txt` is denied with `ViolationOutsideWorkspace` before the auto-allow even runs. `TestEngineDeniesApplyPatchEscapesFromPatchBody` (engine_test.go:256-274) covers the exact `b/../escape.txt` shape. Symlink escape through the auto-allow path is also now tested separately in `TestEngineDeniesAutoAllowedSymlinkEscape` (engine_test.go:309-331) using a workspace-resident symlink pointing outside, asserted as `ViolationSymlinkTraversal`.

**Test gaps closed.**
1. Protected-metadata → `TestEngineDoesNotAutoAllowProtectedMetadataWrites` covers `write_file`/`.git/hooks`, `edit_file`/`.zero/`, and `apply_patch`/`.agents/`.
2. `apply_patch` + `DenyWrite`/escape → `TestEngineDeniesApplyPatchEscapesFromPatchBody`.
3. Session grant cross-file → `TestEngineSessionGrantDoesNotMatchSiblingFile` (engine_test.go:276-307) grants `write_file src/a.txt`, then asserts `src/a.txt` matches (ActionAllow + GrantMatched + Session) and `src/b.txt` re-prompts (ActionPrompt + no GrantMatched).
4. Symlink through auto-allow → `TestEngineDeniesAutoAllowedSymlinkEscape` (above).

**Positive test still passes.** `TestEngineAutoAllowsWorkspaceFileMutationTools` (engine_test.go:188-226) covers `write_file notes.txt`, `edit_file notes.txt`, and `apply_patch a/notes.txt b/notes.txt` all auto-allowed, plus the existing `outside` deny case for the `write_file` escape.

**Non-blocking follow-ups still apply (no change since last review):**
- TUI `permissionOptions` fallback offers "always" even with no persistent store (`internal/tui/permission_prompt.go:21-30`). Defensive, currently unreachable — worth tightening when a future call site exercises it.
- `persistSessionPermissionGrant` swallows errors silently so the session-grant event's `GrantMatched` flips to `false` on persistence failure (`internal/agent/loop.go:583`). Right call but worth a one-line code comment so a future reader doesn't think the event is broken.

This is the right shape. Auto-allow fires for the three built-in file tools on real workspace targets, re-prompts for any path that lands under `.git`/`.zero`/`.agents` (whether the path comes from the args or the patch body), and denies out-of-workspace / symlink escapes before the prompt is even shown. Happy to see it land once the maintainer gives the final sign-off on the follow-ups.

Vasanthdev2004:
Re-reviewed at `cfbb48e` ("Guard workspace write auto-allow") and switching to **APPROVE**.

Both blockers from my prior review are addressed end-to-end:

- **HIGH — protected-metadata bypass**: `requestPaths` in `internal/sandbox/risk.go:236-238` now folds in `applyPatchRequestPaths`, so the new `workspaceWriteAutoAllowed` runs `requestPathsTouchProtectedMetadata` against the union. An `apply_patch` whose patch body targets `.git/hooks/pre-commit`, `.zero/config.json`, or `.agents/...` re-prompts instead of auto-allowing. Covered by `TestEngineDoesNotAutoAllowProtectedMetadataWrites` for all three sub-cases.
- **MEDIUM — `apply_patch` path-blind**: `applyPatchPathViolation` extracts patch headers and denies `+++ b/../escape.txt` with `ViolationOutsideWorkspace` before the auto-allow even runs. `TestEngineDeniesApplyPatchEscapesFromPatchBody` and `TestEngineDeniesAutoAllowedSymlinkEscape` (using a workspace-resident symlink → `ViolationSymlinkTraversal`) close both escape shapes.

All four test gaps from the prior review now have coverage, and the positive path (`TestEngineAutoAllowsWorkspaceFileMutationTools`) still passes for `write_file`/`edit_file`/`apply_patch` on real workspace targets.

The two non-blocking follow-ups from before are unchanged — TUI `permissionOptions` fallback offering "always" with no persistent store, and the `persistSessionPermissionGrant` error-swallow needing a one-line comment. Worth a follow-up PR; don't gate this one on them.

Thanks for the clean fix.

PR #246, opened 2026-06-18
Align permission prompt rendering

anandh8x:
## Summary
- remove normal permission prompt risk labels from focused TUI cards, transcript rows, and permission detail text
- change the recoverable deny option to say it continues without running the tool instead of implying a run-cancel redirect
- use conservative allow/deny fallback choices when a permission request lacks backend-supplied decisions, avoiding session or persistent choices the prompt cannot honor
- document that a failed session-grant recording attempt only means a later matching call prompts again, not that the current user-approved call is denied
- add an explicit cancel decision that aborts the active run while keeping deny as a recoverable tool refusal
- expose cancel choices for command and apply-patch approval prompts, with apply-patch using cancel instead of recoverable deny and command/apply-patch avoiding persistent always approvals until scoped prefix approval exists
- classify cancel outcomes as permission-decision events for TUI session logs and exec JSON output
- publish encrypted OAuth token secrets atomically under a creation lock so concurrent first-run callers cannot read a half-written secret

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/tui ./internal/agent
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/tui ./internal/cli
- go test ./internal/oauth -run TestLoadOrCreateSecretConcurrentConverges -count=100
- go test ./internal/oauth
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added a **Cancel** permission option, including a direct cancel hotkey, with updated “cancelled” handling in the UI.
  * Permission choices are now tailored to context (sandbox persistence and tool-specific cancel/deny capabilities).
* **Bug Fixes**
  * Removed risk-level badges/labels from terminal permission views (rows, focused prompts, and transcript).
  * Updated denial-style wording to **“No, continue without running it.”**
* **Tests**
  * Expanded coverage for cancel behavior, decision fallback ordering when no options are provided, and correct option visibility across scenarios.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -23,6 +23,8 @@ const (
 	toolResultControlSpecReview = "spec_review_required"
 )
 
+var errPermissionApprovalCanceled = errors.New("permission approval cancelled")
+
 // abortedToolResultNotice is the placeholder tool result recorded for a tool
 // call that was advertised by the assistant turn but never executed because the
 // repeated-failure guard halted the run first. It keeps every tool_use paired
@@ -604,6 +606,9 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 					requestEvent.Grant = &grant
 				}
 			}
+		case PermissionDecisionCancel:
+			emitCanceledPermission(options, call, requestEvent, decisionReason)
+			return canceledPermissionResult(call, decisionReason, requestEvent), fmt.Errorf("%w for %s", errPermissionApprovalCanceled, call.Name)
 		default:
 			emitDeniedPermission(options, call, requestEvent, decisionReason)
 			return deniedPermissionResult(call, decisionReason, requestEvent), nil
@@ -952,7 +957,7 @@ func requestPermission(ctx context.Context, request PermissionRequest, options O
 
 func normalizePermissionDecisionAction(action PermissionDecisionAction) PermissionDecisionAction {
 	switch action {
-	case PermissionDecisionAllow, PermissionDecisionAllowForSession, PermissionDecisionAlwaysAllow:
+	case PermissionDecisionAllow, PermissionDecisionAllowForSession, PermissionDecisionAlwaysAllow, PermissionDecisionCancel:
 		return action
 	default:
 		return PermissionDecisionDeny
@@ -1024,6 +1029,25 @@ func emitDeniedPermission(options Options, call ToolCall, requestEvent Permissio
 	options.OnPermission(event)
 }
 
+func emitCanceledPermission(options Options, call ToolCall, requestEvent PermissionEvent, reason string) {
+	if options.OnPermission == nil {
+		return
+	}
+	reason = strings.TrimSpace(reason)
+	if reason == "" {
+		reason = "cancelled in TUI"
+	}
+	event := requestEvent
+	event.ToolCallID = call.ID
+	event.ToolName = call.Name
+	event.Action = PermissionActionCancel
+	event.DecisionAction = PermissionDecisionCancel
+	event.PermissionGranted = false
+	event.DecisionReason = reason
+	event.Reason = reason
+	options.OnPermission(event)
+}
+
 func deniedPermissionResult(call ToolCall, reason string, requestEvent PermissionEvent) ToolResult {
 	reason = strings.TrimSpace(reason)
 	if reason == "" {
@@ -1055,6 +1079,32 @@ func deniedPermissionResult(call ToolCall, reason string, requestEvent Permissio
 		Status:       tools.StatusError,
 		Output:       "Error: Permission denied for " + call.Name + ": " + reason,
 		DenialReason: denial,
+		Meta: map[string]string{
+			"permission_action": string(event.Action),
+		},
+	}
+}
+
+func canceledPermissionResult(call ToolCall, reason string, requestEvent PermissionEvent) ToolResult {
+	reason = strings.TrimSpace(reason)
+	if reason == "" {
+		reason = "cancelled in TUI"
+	}
+	event := requestEvent
+	event.Action = PermissionActionCancel
+	event.DecisionAction = PermissionDecisionCancel
+	event.PermissionGranted = false
+	event.DecisionReason = reason
+	event.Reason = reason
+	if requestEvent.ToolName == "" {
+		event.ToolName = call.Name
+	}
+	return ToolResult{
+		ToolCallID:   call.ID,
+		Name:         call.Name,
+		Status:       tools.StatusError,
+		Output:       "Error: Permission approval cancelled for " + call.Name + ": " + reason,
+		DenialReason: DenialApprovalCanceled,
 		Meta: map[string]string{
 			"permission_action": string(event.Action),
 		},
@@ -1203,12 +1253,28 @@ func availablePermissionDecisions(event PermissionEvent, options Options) []Perm
 	decisions := []PermissionDecisionAction{PermissionDecisionAllow}
 	if options.Sandbox != nil {
 		decisions = append(decisions, PermissionDecisionAllowForSession)
-		if options.Sandbox.CanPersistGrants() {
+		if options.Sandbox.CanPersistGrants() && permissionSupportsPersistentDecision(event.ToolName) {
 			decisions = append(decisions, PermissionDecisionAlwaysAllow)
 		}
 	}
-	decisions = append(decisions, PermissionDecisionDeny)
+	switch event.ToolName {
+	case "bash":
+		decisions = append(decisions, PermissionDecisionDeny, PermissionDecisionCancel)
+	case "apply_patch":
+		decisions = append(decisions, PermissionDecisionCancel)
+	default:
+		decisions = append(decisions, PermissionDecisionDeny)
+	}
 	return decisions
+}
+
+func permissionSupportsPersistentDecision(toolName string) bool {
+	switch toolName {
+	case "bash", "apply_patch":
+		return false
+	default:
+		return true
+	}
 }
 
 func cloneArgs(args map[string]any) map[string]any {
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -580,6 +580,8 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 			permissionGranted = true
 			requestEvent.DecisionAction = decision.Action
 			if options.Sandbox != nil {
+				// The current call stays allowed if recording the session grant
+				// fails; the user is simply prompted again for a later match.
 				if grant, err := persistSessionPermissionGrant(call.Name, args, decisionReason, options); err == nil {
 					requestEvent.GrantMatched = true
 					requestEvent.Grant = &grant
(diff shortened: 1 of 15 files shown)


Vasanthdev2004 (approved):
## APPROVE

Walked the 13 files (335 +, 51 -) and the 3 stacked commits at `0d0d764a`. The Cancel decision is a clean addition that closes the recoverable-deny / run-abort ambiguity, the conservative fallback addresses a follow-up from #244, and the risk-label removal lines up with the project's "don't show what the user can't act on" stance. Two design follow-ups (no code changes required); no blockers.

## What works

**1. New `PermissionDecisionCancel` decision, fully wired.** `internal/agent/types.go:46` adds the constant; `loop.go:609-611` adds the case to the dispatch switch; `loop.go:960` includes it in `normalizePermissionDecisionAction` (so a future unknown action still fail-closes to Deny, not Cancel); `emitCanceledPermission` and `canceledPermissionResult` (loop.go:1032-1110) are the symmetric pair to the Deny helpers. The new `errPermissionApprovalCanceled` sentinel is checkable via `errors.Is`, and the new `DenialApprovalCanceled` category lets the CLI/TUI distinguish a user-canceled run from a recoverable deny. The `Meta["permission_action"]` field on the canceled `ToolResult` matches the key used elsewhere, so the stream-JSON output stays schema-consistent.

**2. `TestRunAbortsWhenPermissionRequestCanceled` (loop_test.go) is the right shape.** It asserts `errors.Is(err, errPermissionApprovalCanceled)`, `result.FinalAnswer == ""` (the run didn't continue), `len(provider.requests) == 1` (the agent didn't get a second turn), the file was never written, and the single `PermissionEvent` carries `Action=PermissionActionCancel` / `DecisionAction=PermissionDecisionCancel` / `PermissionGranted=false` / `DecisionReason="redirect needed"`. That's exactly the contract callers will branch on.

**3. Conservative fallback choices close a #244 follow-up.** `permission_prompt.go:23-30` now defaults the fallback to `Allow` + `Deny` only — no session, no always. `TestPermissionOptionsEmptyDecisionsUseRecoverableFallback` pins it. The earlier `gh` PR review on #244 flagged the original "always" fallback as a foot-gun; this PR is the clean fix and the test makes it hard to regress.

**4. Risk labels removed cleanly.** `renderPermissionRow` (rendering.go:775-810), `renderFocusedPermissionPrompt` (rendering.go:840-842), `permissionRowText` (transcript.go:272-275), `permissionDetailText` (transcript.go:295-308), and the `permRisk` theme style (theme.go:64, theme.go:154) are all stripped of the `risk:<level>` rendering. The focused prompt's "PERMISSION" badge no longer carries the amber-on-amber risk chip. Five new TUI tests cover the badge / options / cancel-hotkey / fallback paths (`TestPermissionOptionsEmptyDecisionsUseRecoverableFallback`, `TestPermissionOptionsExposeApprovalCancelWhenSupplied`, `TestPermissionOptionsCanExposePatchCancelWithoutRecoverableDeny`, `TestPermissionCancelHotkeyResolvesDirectly`, `TestFocusedPermissionCardShowsBadgeAndKeys`).

**5. Cancel classified as a decision event in both surfaces.** `tuiPermissionEventType` (model.go:3129) and the mirror in `cli/exec.go:1073-1076` / `cli/exec_writer.go:245-248` all extend the `Action == Allow || Deny` predicate to include `Action == Cancel`, so the session log and the `zero exec --output-format stream-json` consumer both see cancel as a `permission_decision` event with the right payload. Good for downstream tooling.

**6. Session-grant recording failure documented.** `loop.go:585-586` now carries a comment that calls out the new behavior the test enforces: "The current call stays allowed if recording the session grant fails; the user is simply prompted again for a later match." That was the implicit guarantee before, but it lives in the test rather than the source — pinning it as a comment in `executeToolCall` is the right place.

**7. The wording shift is honest.** Deny was previously "No, and tell Zero what to do differently"; now it's "No, continue without running it" and Cancel takes over the old wording. A user upgrading will see a real semantic change: pressing the same hotkey does a different thing. Worth a CHANGELOG line, but the wording itself is clear about the new behavior.

## Non-blocking follow-ups

1. **Document why `apply_patch` drops Deny and adds Cancel instead of having both.** The decision list in `availablePermissionDecisions` (loop.go:1260-1267) is now: bash → `[Allow, AllowForSession, Deny, Cancel]`; apply_patch → `[Allow, AllowForSession, Cancel]`; default (write_file, edit_file, ...) → `[Allow, AllowForSession, Deny]`. The asymmetry is explicit design ("apply-patch using cancel instead of recoverable deny") but the rationale isn't in the source. Worth a one-line comment on the `case "apply_patch":` branch — something like `// apply_patch is bulk and binary (applied or not); "deny" is unrecoverable, so cancel is the only way to redirect.` Otherwise a future reader will assume bash's four-option set is the canonical pattern and "fix" apply_patch to add Deny back, which changes behavior in a way the prompt won't surface.

2. **Document the absence of Cancel for `write_file` / `edit_file`.** Same switch — only `bash` and `apply_patch` get the Cancel option. A user aborting from a `write_file` prompt can only Deny, which keeps the run going. That's likely fine (single-file edits are narrow enough that Deny is the natural redirect), but the design is "cancel for commands and patches; deny for individual file writes" and it isn't spelled out. A two-line comment in `availablePermissionDecisions` explaining the rule (cancel = abort-the-run, only for bulk / command-shaped tools) would prevent accidental widening later.

3. **The Cancel label "No, and tell Zero what to do differently" is the *old* Deny label.** Worth a CHANGELOG / release-notes line so a user upgrading from a previous build doesn't press the same key expecting the old behavior. The TUI's hotkey legend (e.g. on the focused prompt) should also show both `d` and `n` once a Cancel is exposed, which is worth verifying in `TestFocusedPermissionCardShowsBadgeAndKeys`.

The shape is right and the test surface covers it. Happy to see this land once the two design follow-ups have a one-line comment in the source.

Vasanthdev2004:
Verdict: **APPROVE**.

Walked the 13 files (335 +, 51 -) and the 3 stacked commits at `0d0d764a`. The Cancel decision is a clean addition that closes the recoverable-deny / run-abort ambiguity, the conservative fallback addresses a follow-up from #244, and the risk-label removal lines up with the project's "don't show what the user can't act on" stance. Three design follow-ups (no code changes required); no blockers.

## What works

- **New `PermissionDecisionCancel` decision, fully wired.** `internal/agent/types.go:46` adds the constant; `loop.go:609-611` adds the case to the dispatch switch; `normalizePermissionDecisionAction` includes it (so unknown actions still fail-close to Deny, not Cancel); `emitCanceledPermission` and `canceledPermissionResult` are the symmetric pair to the Deny helpers. `errPermissionApprovalCanceled` is checkable via `errors.Is`, and the new `DenialApprovalCanceled` category lets the CLI/TUI distinguish a user-canceled run from a recoverable deny. The `Meta["permission_action"]` field stays schema-consistent.
- **`TestRunAbortsWhenPermissionRequestCanceled` is the right shape.** It asserts the `errors.Is` check, an empty `FinalAnswer` (run didn't continue), one provider request (no second turn), the file was never written, and the single `PermissionEvent` carries `Action=PermissionActionCancel` / `DecisionAction=PermissionDecisionCancel` / `PermissionGranted=false` / `DecisionReason="redirect needed"`. That's the contract callers will branch on.
- **Conservative fallback choices close a #244 follow-up.** `permission_prompt.go:23-30` now defaults the fallback to `Allow` + `Deny` only — no session, no always. Pinned by `TestPermissionOptionsEmptyDecisionsUseRecoverableFallback`. The earlier review on #244 flagged the original "always" fallback as a foot-gun; this PR is the clean fix.
- **Risk labels removed cleanly.** `renderPermissionRow`, `renderFocusedPermissionPrompt`, `permissionRowText`, `permissionDetailText`, and the `permRisk` theme style are all stripped of the `risk:<level>` rendering. The focused prompt's "PERMISSION" badge no longer carries the amber-on-amber risk chip.
- **Cancel classified as a decision event in both surfaces.** `tuiPermissionEventType` (model.go:3129) and the mirrors in `cli/exec.go:1073-1076` / `cli/exec_writer.go:245-248` all extend the predicate to include `Action == Cancel`, so the session log and `zero exec --output-format stream-json` see cancel as a `permission_decision` event. Good for downstream tooling.
- **Session-grant recording failure documented.** The new comment in `loop.go:585-586` pins the "current call stays allowed, later match re-prompts" behavior in the source rather than just the test.
- **The wording shift is honest.** Deny is now "No, continue without running it" and Cancel takes over the old wording. A user upgrading will see a real semantic change; worth a CHANGELOG line.

## Non-blocking follow-ups

1. **Document why `apply_patch` drops Deny and adds Cancel instead of having both.** `availablePermissionDecisions` (loop.go:1260-1267) is now: bash → `[Allow, AllowForSession, Deny, Cancel]`; apply_patch → `[Allow, AllowForSession, Cancel]`; default (write_file, edit_file, ...) → `[Allow, AllowForSession, Deny]`. The asymmetry is explicit design but the rationale isn't in the source. Worth a one-line comment on the `case "apply_patch":` branch — something like `// apply_patch is bulk and binary; "deny" is unrecoverable, so cancel is the only way to redirect.`
2. **Document the absence of Cancel for `write_file` / `edit_file`.** A user aborting from a `write_file` prompt can only Deny, which keeps the run going. The design is "cancel for commands and patches; deny for individual file writes" and it isn't spelled out. A two-line comment in `availablePermissionDecisions` would prevent accidental widening later.
3. **The Cancel label is the *old* Deny label.** Worth a CHANGELOG line so a user upgrading from a previous build doesn't press the same key expecting the old behavior. The TUI's hotkey legend should show both `d` and `n` once Cancel is exposed; worth verifying in `TestFocusedPermissionCardShowsBadgeAndKeys`.

The shape is right and the test surface covers it. Happy to see this land once the two design follow-ups have a one-line comment in the source.

gnanam1990 (approved):
## Review — APPROVE

Read the full diff + the post-change source on the branch, and verified the load-bearing pieces directly (built it, ran the tui/agent/oauth suites including the oauth race test under `-count`).

### Verified correct
- **Cancel vs deny is right.** `loop.go:611` CANCEL returns `canceledPermissionResult(...)` + `errPermissionApprovalCanceled` → the dispatch loop aborts the run (aborted placeholders for remaining calls, returns the error). `loop.go:614` DENY returns `deniedPermissionResult(...), nil` → the denial is fed back as a tool_result and the run continues. No path crosses the two; the paired tests (deny → run continues / cancel → single provider request, empty FinalAnswer) lock it in.
- **Fallback is conservative.** With no backend decisions, `permission_prompt.go` offers only `[Allow, Deny]` — never session/persistent — with Allow as the resting cursor default. Covered by the new empty-decisions test.
- **Risk-label removal hides nothing actionable.** The focused prompt still shows tool name, side-effect (network/write/…), reason, and scope (target host for network); denied/violation rows keep `violation=<code>` + path. "Risk" was a coarse roll-up of those same signals.
- **Cancel is classified everywhere** (exec JSON, exec writer, TUI session log, row context) and treated as non-retriable (`DenialApprovalCanceled`).
- **OAuth atomic publish is correct:** temp-create → chmod 0600 → write → `os.Rename` (atomic), serialized by an `O_EXCL` `.lock`; losers poll the secret. 0600 verified; race test green under `-count`.

### Non-blocking
- **LOW — `internal/oauth/encrypt.go` `createSecretFile`: orphaned lock isn't self-healing.** `defer os.Remove(lockPath)` covers normal return + panic, but a *hard* process death (SIGKILL/power loss) in the tiny window between the `O_EXCL` create and the rename leaves `path.lock` behind; every later first-run caller then spins the full timeout and fails permanently until someone deletes it. Narrow window, and not a severity regression over the prior O_EXCL-on-secret approach — but an mtime-based stale-lock reclaim would make it self-healing. Worth a follow-up.
- **NIT — esc during a pending prompt** aborts via context-cancel rather than the new in-band CANCEL, so it logs as a generic "Run cancelled." event instead of a `PermissionActionCancel` decision event. Net behavior (run aborts) is correct — just an observability inconsistency.

Clean, well-tested change — the cancel/deny split is the right model and the encryption race fix is a real bug closed.

PR #248, opened 2026-06-18
Add request permissions flow

anandh8x:
## Summary
- add a runtime-handled request_permissions tool for temporary turn/session filesystem and network grants
- thread granted read/write/network profiles through sandbox policy, scope, command planning, and native sandbox profiles
- update the TUI prompt with request-specific grant choices, strict-review response support, and Esc-as-continue-without-permissions

## Tests

## Release Notes

* **New Features**
  * Added a `request_permissions` tool to request turn- or session-scoped network and filesystem permissions, with a “allow with review” strict option.
  * Updated the permissions prompt UI for `request_permissions`, including clearer grant labels and “[esc] continue without permissions”.

* **Bug Fixes**
  * Improved read-only scoping: read paths are validated against read roots to allow reads while preventing writes.
  * Refined permission decision handling and turn-scoped permission cleanup/undo behavior.

* **Tests**
  * Added unit tests covering granting, denial/continuation, strict-review output, and cleanup/enforcement.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1031,8 +1031,17 @@ func executeRequestPermissions(ctx context.Context, call ToolCall, args map[stri
 			Output:     "Error: request_permissions requires at least one permission.",
 		}, nil
 	}
+	grantProfile, err := sandbox.RequestPermissionGrantProfile(normalized)
+	if err != nil {
+		return ToolResult{
+			ToolCallID: call.ID,
+			Name:       call.Name,
+			Status:     tools.StatusError,
+			Output:     "Error: Invalid permissions for request_permissions: " + err.Error(),
+		}, nil
+	}
 
-	request := requestPermissionsPrompt(call, parsed, normalized, permissionMode, options)
+	request := requestPermissionsPrompt(call, parsed, grantProfile, permissionMode, options)
 	if options.OnPermissionRequest == nil {
 		response := sandbox.RequestPermissionsResponse{
 			Permissions: sandbox.RequestPermissionProfile{},
@@ -1058,16 +1067,16 @@ func executeRequestPermissions(ctx context.Context, call ToolCall, args map[stri
 	permissionGranted := false
 	switch decision.Action {
 	case PermissionDecisionAllow:
-		response.Permissions = normalized
+		response.Permissions = grantProfile
 		response.Scope = sandbox.PermissionGrantScopeTurn
 		permissionGranted = true
 	case PermissionDecisionAllowStrict:
-		response.Permissions = normalized
+		response.Permissions = grantProfile
 		response.Scope = sandbox.PermissionGrantScopeTurn
 		response.StrictAutoReview = true
 		permissionGranted = true
 	case PermissionDecisionAllowForSession:
-		response.Permissions = normalized
+		response.Permissions = grantProfile
 		response.Scope = sandbox.PermissionGrantScopeSession
 		permissionGranted = true
 	case PermissionDecisionCancel:
@@ -1158,7 +1167,7 @@ func requestPermissionsPrompt(call ToolCall, parsed requestPermissionsArgs, prof
 func requestPermissionsScope(profile sandbox.RequestPermissionProfile) string {
 	parts := []string{}
 	if profile.Network != nil && profile.Network.Enabled != nil && *profile.Network.Enabled {
-		parts = append(parts, "network")
+		parts = append(parts, "network all outbound access")
 	}
 	if profile.FileSystem != nil {
 		for _, path := range profile.FileSystem.Read {
@@ -1171,7 +1180,7 @@ func requestPermissionsScope(profile sandbox.RequestPermissionProfile) string {
 			parts = append(parts, "deny read "+path)
 		}
 	}
-	return truncateScope(strings.Join(parts, ", "))
+	return strings.Join(parts, ", ")
 }
 
 func emitRequestPermissionsDecision(options Options, call ToolCall, request PermissionRequest, action PermissionDecisionAction, reason string, granted bool) {
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/Gitlawb/zero/internal/hooks"
 	"github.com/Gitlawb/zero/internal/redaction"
@@ -24,6 +25,33 @@ const (
 )
 
 var errPermissionApprovalCanceled = errors.New("permission approval cancelled")
+
+type permissionRunState struct {
+	mu       sync.Mutex
+	cleanups []func()
+}
+
+func (state *permissionRunState) add(cleanup func()) {
+	if state == nil || cleanup == nil {
+		return
+	}
+	state.mu.Lock()
+	state.cleanups = append(state.cleanups, cleanup)
+	state.mu.Unlock()
+}
+
+func (state *permissionRunState) cleanup() {
+	if state == nil {
+		return
+	}
+	state.mu.Lock()
+	cleanups := state.cleanups
+	state.cleanups = nil
+	state.mu.Unlock()
+	for i := len(cleanups) - 1; i >= 0; i-- {
+		cleanups[i]()
+	}
+}
 
 // abortedToolResultNotice is the placeholder tool result recorded for a tool
 // call that was advertised by the assistant turn but never executed because the
@@ -66,6 +94,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 	if permissionMode == "" {
 		permissionMode = PermissionModeAuto
 	}
+	runPermissions := &permissionRunState{}
+	options.runPermissions = runPermissions
+	defer runPermissions.cleanup()
 
 	messages := zeroruntime.SeedMessagesWithImages(buildSystemPrompt(options), prompt, options.Images)
 
@@ -547,6 +578,9 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 	if call.Name == "ask_user" {
 		return executeAskUser(ctx, registry, call, args, permissionMode, options)
 	}
+	if call.Name == tools.RequestPermissionsToolName {
+		return executeRequestPermissions(ctx, call, args, permissionMode, options)
+	}
 
 	permissionGranted := permissionMode == PermissionModeUnsafe
 	if toolFound && tool.Safety().Permission == tools.PermissionAllow {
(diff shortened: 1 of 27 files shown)


gnanam1990 (changes requested):
## Review — REQUEST_CHANGES

This is a careful, well-tested feature and the core security architecture is genuinely sound — I verified the load-bearing properties directly in the code, not just the diff. The one blocker is that the approval prompt **under-states the actual grant**, which undermines the very thing this feature depends on: informed human consent.

### ✅ Verified safe (nice work)
- **The approval gate can't be bypassed.** `executeRequestPermissions` is dispatched at the top of `executeToolCall` (`loop.go:581`) *before* the `permissionMode == PermissionModeUnsafe` shortcut, so `--auto`/unsafe does **not** auto-grant. It fails closed when `OnPermissionRequest == nil` (headless `exec` → `granted=false`, no grant), denies on error, and only calls `GrantRequestPermissions` inside the positive-decision branches. No autonomy path to a self-grant.
- **Turn grants expire** at task end (`defer` cleanup on `Run`); denied/cancelled requests leave no grant; read grants don't permit writes; `..`/symlink/filesystem-root escapes are blocked; `.git`/`.zero`/`.agents` stay threaded as protected-metadata in granted native write roots.

### 🟠 HIGH — the prompt shows a narrower grant than what is actually granted
`permissionRoot` (`sandbox/request_permissions.go`) widens a **file** request to its **parent directory**:
```go
func permissionRoot(path string) string {
	clean := filepath.Clean(path)
	if info, err := os.Stat(clean); err == nil && info.IsDir() { return clean }
	return filepath.Dir(clean)   // a file -> its whole parent dir
}
```
…but the prompt (`requestPermissionsScope`, `loop.go`) is built from the **un-widened** file paths (`profile.FileSystem.Read/Write`). So if the agent requests `read ~/.ssh/id_rsa`, the user sees and approves `read …/id_rsa`, but the sandbox actually grants **read of `~/.ssh/`** — every sibling secret. The displayed scope is strictly narrower than the real grant, which breaks the informed-consent model the whole feature rests on. (An independent probe confirmed an unrequested sibling file became readable after a single-file grant.)

**Fix:** make the prompt match reality — either grant exactly the requested file, or widen-and-**display** the actual directory root so the prompt never under-states the grant.

### 🟡 MEDIUM — "Allow with review" is advisory-only but labeled like a guarantee
`AllowStrict` grants the **identical** permissions as plain `Allow`; the only effect is `strict_auto_review: true` in the JSON returned to the model. `StrictAutoReview` has **no enforcement consumer** anywhere (engine/sandbox/self-correct/system-prompt) — grep shows only the setter (`loop.go:1067`), the const (`types.go:43`), and the struct field. The label "Grant for this turn and review commands" implies an enforced check that doesn't exist; it relies entirely on the LLM voluntarily honoring the flag. **Fix:** enforce it (e.g. force strict review of subsequent commands) or relabel so the user isn't misled.

### 🟡 MEDIUM — network grant is all-or-nothing and overrides operator policy
`applyRequestPermissionProfile` sets `policy.Network = NetworkAllow; AllowedDomains = nil; DeniedDomains = nil` — no per-host scoping. A boolean "network" request flips the turn/session to **fully unrestricted egress** and discards an operator-configured scoped allow/deny policy for the grant's lifetime, while the prompt only shows the word "network". (Expiry is correct; breadth is the concern.) Consider per-host grants, or at least surfacing that a configured network policy is being replaced with allow-all.

### 🟡 LOW
- **No guardrail on grantable roots** — only `/` is rejected; `/etc`, `/usr`, `$HOME` succeed on approval (a probe confirmed `grant write /etc` succeeds). Combined with the HIGH, consider a denylist of sensitive roots or an extra warning.
- **Prompt scope truncated to 80 runes** (`truncateScope`) — a many-path/long-path grant is approved while the user sees only the first ~80 chars; the full profile is still granted.

### NIT
- "Grant for this **turn**" — the grant actually lasts the whole task/`Run` (all model turns of one user prompt); "for this task" would match the tool description.

`go build` / `go vet` / `-race` / tests all pass. The architecture is solid — please tighten prompt↔grant fidelity (HIGH) and the "allow with review" labeling (MEDIUM) before merge.

gnanam1990 (approved):
## Re-review (after `e0b9182` "Align request permission grants with prompts") — APPROVE

The blocker is fixed cleanly. I re-verified in the code and built/tested the updated branch.

**HIGH (prompt under-stated the grant) — fixed.** New `RequestPermissionGrantProfile` computes the effective (widened-to-parent-dir) grant up front, and `executeRequestPermissions` feeds that same `grantProfile` to **both** the prompt (`requestPermissionsPrompt(…, grantProfile, …)`) **and** the grant (`response.Permissions = grantProfile`). So the user now sees exactly what's granted — a `read ~/.ssh/id_rsa` request displays *and* grants `read ~/.ssh/` consistently (file paths dedupe to their parent root). Display == grant; no more under-statement.

**MEDIUM (network prompt) — addressed.** The prompt now reads "network all outbound access" instead of just "network", so a network grant's unrestricted egress is visible. (Still all-or-nothing rather than per-host — a reasonable follow-up — but the prompt no longer hides it.)

**MEDIUM ("allow with review" mislabeled) — addressed.** Relabeled to "Grant for this task and ask model to review commands" (+ "approved with model review request"), which honestly conveys it's a request to the model, not an enforced guarantee.

**LOW (80-rune truncation) — fixed.** The scope string is shown in full, so a multi-path grant isn't hidden.

**NIT (turn → task) — fixed.**

`go build`, `go vet`, and `internal/{agent,sandbox,tui}` tests are green on the updated branch; the approval gate and grant-expiry safety I verified earlier are unchanged.

Remaining items are non-blocking optional follow-ups: a denylist for sensitive roots like `/etc` (lower-risk now that the human sees the accurate scope), and per-host network scoping. Neither blocks merge. Nice turnaround — approving.

anandh8x (commented):

Reviewed the request-permissions flow around effective policies, scoped path grants, session persistence, and workspace tool path resolution. git diff --check passed for this PR. I did not run the full branch test suite.

Note: GitHub would not allow this account to approve the PR because it is treated as the author/owner review path, so leaving this as a review comment instead.

Vasanthdev2004 (approved):
## APPROVE (re-review of the third commit)

The first two commits were thoroughly reviewed and approved by gnanam1990 (id `4525458085` on `e0b9182`) after the prompt↔grant fidelity fix and the network-prompt / strict-review relabel landed. This third commit — `a813eee` "Retry transient OAuth secret read locks" — is independent of the request-permissions flow and addresses a Windows-specific race in the OAuth secret-file creation path. It's small, well-scoped, and the test surface carries it.

## `a813eee` — Windows transient read retry

`internal/oauth/encrypt.go:12-19, 88, 105-128, 162-172`:

- Magic numbers `500` and `2 * time.Millisecond` (already present in the existing `createSecretFile` retry loop) are pulled into named constants `secretRetryAttempts` and `secretRetryDelay`. The total wait budget is unchanged at 1 second.
- A new `readSecretFileRetry` helper wraps the existing `readSecretFile` with the same retry loop, but only continues when the error is classified as transient. The non-transient case returns immediately, so behavior on already-handled errors (e.g. `os.ErrNotExist`, real I/O errors, validation failures) is byte-identical to before.
- All three call sites — `loadOrCreateSecret` (fast-path existing secret), and the two read-within-`createSecretFile` loops (lock-loser path, post-create read) — now route through `readSecretFileRetry`. So the lock-loser that just lost the `O_EXCL` race and the lock-loser that already won the race but is reading a freshly-published secret both retry the same way.
- Final error is wrapped with the last transient error via `%w`, so the caller still sees what actually went wrong if retries exhaust.

## Platform split — `isTransientSecretAccessError`

`internal/oauth/secret_access_other.go:1-7` (build tag `!windows`): always returns `false`. Non-Windows platforms see zero behavior change.

`internal/oauth/secret_access_windows.go:1-14` (build tag `windows`): returns `true` for `errors.Is(err, windows.ERROR_SHARING_VIOLATION)` or `errors.Is(err, windows.ERROR_LOCK_VIOLATION)`. These are the two Windows error codes that fire briefly while a freshly-published file's handle is still being released by the creating process — exactly the race the PR body describes.

Build-tag split is mutually exclusive and correct. Verified cross-compile + `go vet` clean on `GOOS=windows GOARCH=amd64`.

## Test surface

The third commit is small enough that the existing concurrency test (`TestLoadOrCreateSecretConcurrentConverges`, 16 goroutines racing on the same path) already exercises the read-retry path on whatever platform it runs on — non-Windows returns `false` from `isTransientSecretAccessError` immediately, so the retry loop is a single iteration; on Windows, the same test would now correctly tolerate the transient errors. There is no platform-specific test for `isTransientSecretAccessError` itself, but the function is small and the error-code constants are stable.

`go test -count=1 ./internal/oauth/... ./internal/sandbox/... ./internal/agent/... ./internal/tools/... ./internal/specialist/... -short` is green on the branch (5 packages, all PASS). The 60-test oauth suite, including the existing concurrent-converges test, still passes.

## Non-blocking follow-ups

1. **The 1-second total retry budget is untested on Windows.** A hypothetical regression where the loop returns after 1 retry would still pass `TestLoadOrCreateSecretConcurrentConverges` (which simulates a clean race, not a transient `ERROR_SHARING_VIOLATION`). A targeted test that injects a one-shot `ERROR_SHARING_VIOLATION` via a fake `readSecretFile` would lock the retry count and the backoff. Not a blocker; the platform split is small enough to read.

2. **The `isTransientSecretAccessError` return-`false` on non-Windows is a no-op call.** A future reader could reasonably ask "why does this exist?" — a one-line comment on `secret_access_other.go` explaining the build-tag split's purpose (so the retry loop has a single, unconditional call site) would be worth the eight lines.

## Verification

- `go test -count=1 ./internal/oauth/... -v` — 60 tests, all PASS, including the 16-goroutine concurrent convergence test.
- `go test -count=1 -short ./internal/sandbox/... ./internal/agent/... ./internal/tools/... ./internal/specialist/...` — all 4 packages PASS.
- `GOOS=windows GOARCH=amd64 go build ./internal/oauth/...` and `... go vet ./internal/oauth/...` — clean.

The third commit is a clean Windows defensive fix that doesn't change non-Windows behavior, doesn't widen any trust boundary, and doesn't change the request-permissions grant semantics. Happy to see it land.

Vasanthdev2004:
Review posted (id `4525830522`, APPROVE on `a813eee`).

The first two commits of this PR (`f8066f6` "Add request permissions flow" and `e0b9182` "Align request permission grants with prompts") were already approved by gnanam1990 (review id `4525458085`) after the prompt↔grant fidelity fix landed. This review covers the third commit only, since it's the only commit without a prior external review.

## `a813eee` — Retry transient OAuth secret read locks

Small, well-scoped Windows defensive fix:

- `secretRetryAttempts` (500) and `secretRetryDelay` (2ms) are pulled out as named constants; total wait budget stays at 1 second.
- New `readSecretFileRetry` wraps `readSecretFile` with a retry loop that only continues when the error is classified transient. Non-transient errors return immediately — behavior on `os.ErrNotExist`, real I/O errors, and validation failures is byte-identical to before.
- All three call sites — `loadOrCreateSecret` and the two reads inside `createSecretFile` — now route through `readSecretFileRetry`. The lock-loser and the post-create reader both retry the same way.
- Final error is wrapped with `%w`, so the caller still sees what actually went wrong if retries exhaust.

## Platform split — `isTransientSecretAccessError`

- `internal/oauth/secret_access_other.go` (build tag `!windows`): always returns `false`. Non-Windows behavior is unchanged.
- `internal/oauth/secret_access_windows.go` (build tag `windows`): returns `true` for `windows.ERROR_SHARING_VIOLATION` and `windows.ERROR_LOCK_VIOLATION`. These are the two transient codes that fire while a freshly-published file's handle is still being released by the creating process.

Build-tag split is mutually exclusive and correct. Cross-compile + `go vet` clean on `GOOS=windows GOARCH=amd64`.

## Test surface

- The existing 16-goroutine `TestLoadOrCreateSecretConcurrentConverges` exercises the read-retry path; non-Windows returns `false` from `isTransientSecretAccessError` immediately (single iteration), Windows would now correctly tolerate the transient errors.
- `go test -count=1 ./internal/oauth/... -v` is green: 60 tests, including the concurrent convergence test.
- `go test -count=1 -short ./internal/sandbox/... ./internal/agent/... ./internal/tools/... ./internal/specialist/...` is green.

## Non-blocking follow-ups

1. The 1-second total retry budget is untested on Windows specifically — a one-shot `ERROR_SHARING_VIOLATION` injection test would lock the retry count and backoff. Not a blocker; the platform split is small enough to read.
2. A one-line comment on `secret_access_other.go` explaining the build-tag split's purpose (so the retry loop has a single unconditional call site) would help future readers.

The third commit doesn't change non-Windows behavior, doesn't widen any trust boundary, and doesn't change the request-permissions grant semantics. Happy to see it land.

PR #249, opened 2026-06-18
redaction: catch compound secret keys + non-bearer auth schemes; stop scanner over-redaction

gnanam1990:
Audit of `internal/redaction` + `internal/secrets` surfaced three verified gaps in the secret-redaction last line of defense. The configured provider key is already exact-scrubbed; these affect **secondary** secrets that appear in tool/command output.

## Fixes
1. **Compound secret keys** *(leak)* — `IsSensitiveKey` was exact-match, so `db_password`, `session_secret`, `stripe_secret_key`, `auth_token`, `csrf_token`, `ssh_private_key`, `backup_api_key` … passed through. New `keyLooksSensitive` uses conservative structural rules. **Critically token-count-safe:** `max_tokens` / `prompt_tokens` / `token_count` and ordinary `primary_key` / `public_key` / `cache_key` are **not** redacted (`token` only triggers as a singular trailing segment; bare `key` never triggers).
2. **Auth header schemes** *(leak)* — `Authorization:` now redacts `token` / `apikey` / `digest` / `negotiate` / `oauth` / `aws4-hmac-sha256`, not just `bearer`/`basic`.
3. **Scanner over-redaction** *(correctness)* — added `\b` to the prefixed patterns so kebab-case words (`task-…`, `ask-…`, `risk-…`) no longer match as a fake key, while real secrets (preceded by space/`=`/`:`/quote/start) still match.

## Deliberately not shipped
A free-text `key: value` colon pattern — it mangled `file.go:line: message` output (a filename's `secret` segment read as a sensitive key), the same over-redaction class fix #3 addresses. Structured paths (`RedactValue`, plus assign/JSON/query in `RedactString`) are covered; free-text colon-form is left as a known limitation, with `TestRedactString_PreservesFileLineColons` guarding against re-introduction.

## Tests
`audit_fixes_test.go` (compound-key + token-count matrix, auth schemes, file:line guard, `RedactValue` nesting) and `boundary_test.go` (scanner over-redaction negatives + real-secret positives). `gofmt` · `go vet` · `staticcheck` · `-race` on redaction+secrets+verify · 9 consumer packages — all green.

## Release Notes

* **Bug Fixes**
  * Improved sensitive field detection with enhanced heuristics for identifying secret-like keys
  * Expanded HTTP header redaction to support additional authentication schemes
  * Fixed secret detection patterns to reduce false positives by requiring word boundaries in pattern matching

* **Tests**
  * Added comprehensive test coverage for redaction behavior and secret detection

### DIFF
--- a/internal/redaction/audit_fixes_test.go
+++ b/internal/redaction/audit_fixes_test.go
@@ -1,0 +1,123 @@
+package redaction
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestIsSensitiveKey_CompoundAndTokenCounts(t *testing.T) {
+	o := Options{}
+	sensitive := []string{
+		"db_password", "smtp_password", "user_password",
+		"session_secret", "webhook_secret", "stripe_secret_key", "client-secret",
+		"auth_token", "csrf_token", "vault_token", "my_token", "xsrf-token",
+		"ssh_private_key", "rsa_private_key",
+		"backup_api_key", "my_apikey", "service_apikey",
+		"db_passphrase", "aws_credential", "gcp_credentials",
+	}
+	for _, k := range sensitive {
+		if !IsSensitiveKey(k, o) {
+			t.Errorf("expected %q to be sensitive", k)
+		}
+	}
+	// CRITICAL: token-count and ordinary "*_key" fields must NOT be redacted —
+	// they are everywhere in an LLM agent and redacting them would break it.
+	notSensitive := []string{
+		"max_tokens", "prompt_tokens", "completion_tokens", "total_tokens",
+		"input_tokens", "output_tokens", "reasoning_tokens", "tokens",
+		"token_count", "token_usage", "tokens_used",
+		"primary_key", "public_key", "cache_key", "foreign_key", "sort_key",
+		"idempotency_key", "key", "key_name", "api_version", "username", "message",
+	}
+	for _, k := range notSensitive {
+		if IsSensitiveKey(k, o) {
+			t.Errorf("expected %q to NOT be sensitive (false positive)", k)
+		}
+	}
+}
+
+func TestRedactString_CompoundKeyForms(t *testing.T) {
+	o := Options{}
+	cases := []struct{ in, secret string }{
+		{"db_password=hunter2supersecret", "hunter2supersecret"},
+		{`{"stripe_secret_key": "rk_live_plaintextvalue"}`, "rk_live_plaintextvalue"},
+		{"GET /x?session_secret=plainsecretvalue HTTP/1.1", "plainsecretvalue"},
+		{"auth_token='opaque-assigned-token-value'", "opaque-assigned-token-value"},
+	}
+	for _, c := range cases {
+		out := RedactString(c.in, o)
+		if strings.Contains(out, c.secret) {
+			t.Errorf("RedactString(%q) leaked %q: got %q", c.in, c.secret, out)
+		}
+		if !strings.Contains(out, RedactedSecret) {
+			t.Errorf("RedactString(%q) should contain %q: got %q", c.in, RedactedSecret, out)
+		}
+	}
+	// Token-count lines must pass through untouched.
+	for _, c := range []string{"max_tokens: 4096", "prompt_tokens=128", "token_count: 50"} {
+		if out := RedactString(c, o); out != c {
+			t.Errorf("RedactString(%q) should be unchanged, got %q", c, out)
+		}
+	}
+}
+
+func TestRedactString_AuthHeaderSchemes(t *testing.T) {
+	o := Options{}
+	// Opaque values (no token-format prefix) so only the header/colon logic can
+	// redact them — isolates the auth-scheme fix from the text-pattern fallback.
+	cases := []struct{ in, secret string }{
+		{"Authorization: Bearer opaquebearervalue1234567", "opaquebearervalue1234567"},
+		{"Authorization: token opaquetokenvalue1234567", "opaquetokenvalue1234567"},
+		{"Authorization: ApiKey opaqueapikeyvalue1234567", "opaqueapikeyvalue1234567"},
+		{"Proxy-Authorization: Digest opaquedigestvalue1234", "opaquedigestvalue1234"},
+	}
+	for _, c := range cases {
+		out := RedactString(c.in, o)
+		if strings.Contains(out, c.secret) {
+			t.Errorf("RedactString(%q) leaked %q: got %q", c.in, c.secret, out)
+		}
+	}
+}
+
+// Guard against re-introducing an over-broad "key: value" pattern: a
+// "file.go:line: message" line (universal in test/compiler output, where the
+// filename may even contain a secret-looking segment) must keep its structure so
+// downstream parsers still work. Only an actual secret VALUE on the line is
+// redacted (by the token-format patterns), never the file:line prefix.
+func TestRedactString_PreservesFileLineColons(t *testing.T) {
+	o := Options{}
+	in := "    secret_test.go:12: token sk-proj-abcdefghijklmnopqrstuvwxyz"
+	out := RedactString(in, o)
+	if !strings.Contains(out, "secret_test.go:12:") {
+		t.Errorf("file:line prefix mangled: %q", out)
+	}
+	if strings.Contains(out, "sk-proj-abcdefghijklmnopqrstuvwxyz") {
+		t.Errorf("secret leaked: %q", out)
+	}
+	if !strings.Contains(out, RedactedSecret) {
+		t.Errorf("expected the token to be redacted: %q", out)
+	}
+}
+
+func TestRedactValue_CompoundKeys(t *testing.T) {
+	o := Options{}
+	in := map[string]any{
+		"db_password": "plainpw",
+		"max_tokens":  4096,
+		"nested":      map[string]any{"session_secret": "innersecretvalue"},
+	}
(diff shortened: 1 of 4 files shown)


anandh8x (dismissed):
High: #249 still leaks credentials for multi-part authorization schemes.

internal/redaction/redaction.go:86 captures only one non-space/comma/semicolon token after Digest or AWS4-HMAC-SHA256, then internal/redaction/redaction.go:180 replaces only that captured token. Real headers such as Authorization: Digest username=..., nonce=..., response=... or AWS SigV4 Credential=..., SignedHeaders=..., Signature=... leave response / Signature visible. Please redact the full auth header value for parameterized schemes and add tests with realistic Digest and AWS headers.

Vasanthdev2004 (approved):
## APPROVE

Walked the 4 files (216 +, 8 -) at `39045f2` against `origin/main`. Three independent fixes to the redaction last line of defense, each with a tight test surface. No regressions on the existing 5 redaction tests or 7 secrets tests, and the 7 new tests all pass. The test for the explicit `keyLooksSensitive` negatives (token-counts, ordinary `*_key` fields, `idempotency_key`) is the right discipline for an additive heuristic.

## 1. Compound secret keys via `keyLooksSensitive`

`internal/redaction/redaction.go:104-145` adds a new structural check layered behind the existing explicit `sensitiveKeys` map and `ExtraSensitiveKeys` options. The check is conservative and well-documented:

- `secretKeySegments` (password, passwd, passphrase, secret, credential, credentials, apikey) — fires on any underscore-delimited segment match.
- `<x>_token` (singular, trailing only) — catches `auth_token`, `csrf_token`, `vault_token`; deliberately excludes `tokens` (plural count) and `token_<x>` (prefix form) so `max_tokens`, `prompt_tokens`, `token_count` stay un-redacted. The `i > 0 && i == len(segments)-1` guard is the right shape.
- `api_key` / `private_key` (adjacent segments) — narrow switch on the previous segment, not a bare `key` match, so `primary_key`, `public_key`, `cache_key`, `foreign_key`, `idempotency_key` stay un-redacted.

Verified the explicit test list end-to-end. `db_password`, `session_secret`, `stripe_secret_key`, `auth_token`, `csrf_token`, `ssh_private_key`, `backup_api_key`, `aws_credential`, `gcp_credentials` all hit the new check. `max_tokens`, `prompt_tokens`, `primary_key`, `public_key`, `idempotency_key`, `username`, `message` correctly stay un-redacted. The new check is purely additive — anything the explicit list or `ExtraSensitiveKeys` flagged before still gets flagged.

## 2. Auth-scheme expansion in `headerPattern`

`internal/redaction/redaction.go:86` extends the scheme alternation from `(bearer|basic)` to `(bearer|basic|token|apikey|api-key|digest|negotiate|oauth|aws4-hmac-sha256)`. Catches:

- GitHub's `Authorization: token ghp_...` (the most common `token` form, distinct from bearer)
- Custom-provider `ApiKey` / `API-Key` headers (case-insensitive match on the scheme, so `ApiKey` and `API-Key` both hit)
- Proxy / Kerberos `Proxy-Authorization: Digest ...` / `Negotiate ...`
- SigV4 `Authorization: AWS4-HMAC-SHA256 Credential=.../..., SignedHeaders=..., Signature=...` (the trailing `([^\s,;]+)` captures the full credential+signature blob, which is the right thing to redact)
- `Authorization: OAuth ...` (custom OAuth realms)

The opaque-value test cases in `TestRedactString_AuthHeaderSchemes` isolate the scheme fix from the text-pattern fallback (the values are deliberately non-prefixed so only the header/colon logic can catch them). Solid.

## 3. Word-boundary anchors on scanner patterns

`internal/secrets/scanner.go:30-49` adds a leading `\b` to all 7 prefixed patterns (AWS, GitHub, GitHub PAT, Slack, Google, OpenAI, JWT) and to the private-key block the doc comment is updated. The fix closes the over-redaction class where `sk-` inside `task-management-and-coordination` matched as a fake `openai_key` because the regex fired mid-word. The comment explicitly enumerates the satisfied-by-real-secrets delimiters (space, quote, `=`, `:`, start-of-string), so the regression-risk surface is documented.

`TestScan_NoOverRedactionOnKebabWords` covers five representative over-redaction cases (`task-management-and-coordination`, `ask-the-user-about-this`, `risk-assessment-and-mitigation`, `disk-usage-monitoring-and-alerting`, `desk-booking-and-reservation`). `TestScan_RealSecretsStillCaught` covers five positives (env-export, JSON, mid-sentence, raw `AKIA...`, raw `ghp_...`) to confirm no real coverage is lost. The real-secrets test would have caught a regression where the `\b` anchor is misplaced.

## Design call worth flagging

The PR body's "dropped prototype" note — a free-text `key: value` colon pattern was prototyped and abandoned because it mangled `file.go:line: message` output. The decision is correct, and `TestRedactString_PreservesFileLineColons` is exactly the regression guard you'd want to keep that decision honest: a filename containing the literal segment `secret` plus a colon, line, colon, and a real token on the same line is exactly the test case where a free-text colon pattern would over-reach. Good test to keep.

## Non-blocking follow-ups

1. **Auth-scheme list isn't exhaustive.** `NTLM` (Windows), `Signature` (some custom auth flows), and per-provider idiosyncratic schemes are not covered. The list is right for the well-known cases (the test matrix is the strongest signal of intent); an explicit opt-in via `Options.ExtraAuthSchemes` would be a clean way to add providers without re-cutting the regex. Not gating; can land as a follow-up.

2. **`keyLooksSensitive` won't catch `auth_token_v2` or similar future-variants.** The trailing-`token` rule requires `i == len(segments)-1`, so any suffix breaks the match. The explicit list covers today's common cases, but a similar "trailing-`<secret_segment>` with any suffix" pattern would be more future-proof. Not a blocker; the test matrix pins today's intent.

3. **The `headerPattern` capture group is unanchored on the right.** The `([^\s,;]+)` captures the value up to whitespace, comma, or semicolon, so an embedded comma in a JWT-style value (rare but possible in some custom auth schemes) would truncate the redaction. Worth a one-line note in the regex comment so the next person doesn't widen the class silently.

## Verification

- `internal/redaction` and `internal/secrets` test suites pass cleanly on the branch (16 tests, including the 7 new ones). `go test -count=1 ./internal/redaction/... ./internal/secrets/... -v` is green.
- The new `keyLooksSensitive` is the only behavior change to `IsSensitiveKey`; the explicit map and `ExtraSensitiveKeys` short-circuits are unchanged, so existing callers of `RedactString` / `RedactValue` that rely on the explicit list are byte-identical to before for already-covered keys.

This is a clean hardening pass. The three fixes are independent, the test surface is complete, and the explicit negatives (token-counts, ordinary `*_key`, kebab-case words, file:line output) are exactly the regression class these heuristics risk introducing. Happy to see it land.

Vasanthdev2004:
Review posted (id `4525748750`, APPROVE on `39045f2`).

Three independent hardening fixes, each with a tight test surface and no regressions on the existing 12 tests:

1. **Compound secret keys** — `keyLooksSensitive` (`internal/redaction/redaction.go:104-145`) layers a conservative structural check behind the explicit `sensitiveKeys` map. Catches `db_password`, `session_secret`, `stripe_secret_key`, `auth_token`, `csrf_token`, `ssh_private_key`, `backup_api_key`, `aws_credential`, `gcp_credentials`, etc., while deliberately excluding `max_tokens`, `prompt_tokens`, `primary_key`, `public_key`, `idempotency_key`. The trailing-`token` rule is well-shaped (singular, last-segment, and `i > 0` to avoid matching a leading `token_<x>` count form).

2. **Auth-scheme expansion** — `headerPattern` (`internal/redaction/redaction.go:86`) now catches `token`, `apikey`/`api-key`, `digest`, `negotiate`, `oauth`, and `aws4-hmac-sha256` in addition to `bearer`/`basic`. Closes the gap on GitHub's `Authorization: token ghp_...`, custom `ApiKey` headers, SigV4 `AWS4-HMAC-SHA256`, and Kerberos `Negotiate`/`Digest` flows.

3. **Word-boundary anchors on scanner patterns** — `internal/secrets/scanner.go:30-49` adds a leading `\b` to all 7 prefixed patterns so `sk-` inside `task-management-and-coordination` no longer over-matches. `TestScan_NoOverRedactionOnKebabWords` and `TestScan_RealSecretsStillCaught` together pin the new behavior.

Verified `go test -count=1 ./internal/redaction/... ./internal/secrets/... -v` is green on the branch (16 tests, including the 7 new ones). The `keyLooksSensitive` change is purely additive — anything the explicit list or `ExtraSensitiveKeys` flagged before still gets flagged.

Three non-blocking follow-ups in the review: auth-scheme list isn't exhaustive (`NTLM`, custom `Signature` schemes), `keyLooksSensitive` won't catch suffix variants like `auth_token_v2`, and the `headerPattern` capture group is unanchored on the right (could truncate at an embedded comma). All three can land as follow-ups.

The `TestRedactString_PreservesFileLineColons` regression guard is exactly the right discipline to keep the "dropped free-text colon pattern" prototype decision honest — worth keeping.

gnanam1990:
Good catch — fixed in f5e1144. The header pattern now redacts the **entire** credential after the scheme (to end of line), so multi-part schemes are fully covered: `Digest …, response=…` and `AWS4-HMAC-SHA256 …, Signature=…` are wiped, not just the first param. Added `TestRedactString_MultiPartAuthSchemes` with realistic Digest + AWS SigV4 headers asserting `response`/`Signature`/`Credential` are all gone while the scheme name stays visible. PTAL.

PR #250, opened 2026-06-18
Add session command prefix approvals

anandh8x:
## Summary
- add validated session-scoped command prefix approvals for bash permission prompts
- support optional prefix_rule on bash tool calls and reject unsafe generalized prefixes
- store prefix approvals in the session sandbox engine while still running commands through the sandbox backend
- render a distinct TUI option for approving matching command prefixes in the current session

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/tui
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/sandbox ./internal/tui ./internal/tools

## New Features
* Introduced command prefix approval for sandboxed sessions: approve a command prefix once to automatically allow all matching commands in the session without additional prompts.
* Added "allow prefix for session" option to permission prompts in the interface.
* Integrated safety validation to prevent approval of unsafe command prefixes.

### DIFF
--- a/internal/agent/command_prefix.go
+++ b/internal/agent/command_prefix.go
@@ -1,0 +1,240 @@
+package agent
+
+import (
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/sandbox"
+	"mvdan.cc/sh/v3/syntax"
+)
+
+var bannedCommandPrefixSuggestions = [][]string{
+	{"python3"},
+	{"python3", "-"},
+	{"python3", "-c"},
+	{"python"},
+	{"python", "-"},
+	{"python", "-c"},
+	{"py"},
+	{"py", "-3"},
+	{"pythonw"},
+	{"pyw"},
+	{"pypy"},
+	{"pypy3"},
+	{"git"},
+	{"bash"},
+	{"bash", "-lc"},
+	{"sh"},
+	{"sh", "-c"},
+	{"sh", "-lc"},
+	{"zsh"},
+	{"zsh", "-lc"},
+	{"/bin/zsh"},
+	{"/bin/zsh", "-lc"},
+	{"/bin/bash"},
+	{"/bin/bash", "-lc"},
+	{"pwsh"},
+	{"pwsh", "-Command"},
+	{"pwsh", "-c"},
+	{"powershell"},
+	{"powershell", "-Command"},
+	{"powershell", "-c"},
+	{"powershell.exe"},
+	{"powershell.exe", "-Command"},
+	{"powershell.exe", "-c"},
+	{"env"},
+	{"sudo"},
+	{"node"},
+	{"node", "-e"},
+	{"perl"},
+	{"perl", "-e"},
+	{"ruby"},
+	{"ruby", "-e"},
+	{"php"},
+	{"php", "-r"},
+	{"lua"},
+	{"lua", "-e"},
+	{"osascript"},
+}
+
+func proposedCommandPrefix(toolName string, args map[string]any) []string {
+	if toolName != "bash" {
+		return nil
+	}
+	command, ok := firstStringArg(args, "command", "cmd", "script", "shell")
+	if !ok {
+		return nil
+	}
+	tokens, ok := safeShellCommandTokens(command)
+	if !ok {
+		return nil
+	}
+	if requested, ok := requestedPrefixRule(args); ok {
+		if safeRequestedPrefix(requested, tokens) {
+			return requested
+		}
+		return nil
+	}
+	if unsafeCommandPrefix(tokens) {
+		return nil
+	}
+	return append([]string(nil), tokens...)
+}
+
+func matchSessionCommandPrefix(toolName string, args map[string]any, options Options) (sandbox.CommandPrefixGrant, bool) {
+	if toolName != "bash" || options.Sandbox == nil {
+		return sandbox.CommandPrefixGrant{}, false
+	}
+	command, ok := firstStringArg(args, "command", "cmd", "script", "shell")
+	if !ok {
+		return sandbox.CommandPrefixGrant{}, false
+	}
+	tokens, ok := safeShellCommandTokens(command)
+	if !ok {
+		return sandbox.CommandPrefixGrant{}, false
+	}
+	return options.Sandbox.LookupCommandPrefixForSession(toolName, tokens)
+}
+
+func firstStringArg(args map[string]any, names ...string) (string, bool) {
+	for _, name := range names {
+		if raw, ok := args[name].(string); ok {
+			value := strings.TrimSpace(raw)
+			if value != "" {
+				return value, true
+			}
+		}
+	}
+	return "", false
+}
+
+func requestedPrefixRule(args map[string]any) ([]string, bool) {
+	raw, ok := args["prefix_rule"]
+	if !ok {
+		return nil, false
+	}
+	switch value := raw.(type) {
+	case []string:
+		return cleanPrefixRule(value), true
+	case []any:
+		out := make([]string, 0, len(value))
+		for _, item := range value {
+			part, ok := item.(string)
+			if !ok {
+				return nil, true
+			}
+			out = append(out, part)
+		}
+		return cleanPrefixRule(out), true
+	default:
+		return nil, true
+	}
+}
+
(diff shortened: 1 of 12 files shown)


gnanam1990 (commented):
## Review — COMMENT (approve-leaning; one denylist gap to tighten)

Carefully designed, and the core security property holds. I verified the load-bearing pieces in the code and with adversarial probes (every accepted string was also executed under real `/bin/sh` with tripwires).

### ✅ Verified safe
- **The standing grant can't be smuggled past.** `matchSessionCommandPrefix` re-parses every later bash command with the SAME `safeShellCommandTokens` single-simple-command validator before matching a stored grant (`command_prefix.go:92`) — it never raw-string-matches or trusts the model's `prefix_rule` at match time. With a `git status` grant, all of these correctly fail to match (→ re-prompt): `git status; curl evil`, `git status && rm -rf x`, `git status | sh`, `git status $(curl evil)`, `git status > /tmp/x`, `FOO=bar git status`, `bash -lc 'git status'`, `gitfoo status`. `&&`/`||`/`|` (BinaryCmd/pipeline), `$(…)`/backticks (non-literal parts), redirects, env-assignments, and globs/braces (via `unsafeCommandPrefixPart`) are all rejected. ~35 adversarial strings yielded no validation-pass-but-extra-execution.
- **Prefix display is exact.** The TUI grants the literal prefix (``allow `git status` in this session``); `safeRequestedPrefix` rejects any prefix that isn't a genuine token-prefix of the real command or is broader than it. No #248-style under-display.
- **Scope & enforcement.** Grants are in-memory on the Engine (no disk-persist path) → strictly session-scoped. Prefix approval only sets `permissionGranted=true`; the command still runs through the sandbox, and `Evaluate`'s network-deny + `DenyDestructiveShell` checks fire BEFORE `PermissionGranted` is consulted (`engine.go:339-344` vs `415`), so a grant can't bypass network/destructive enforcement.

### 🟡 MEDIUM — denylist omits generic command-runners
`unsafeCommandPrefixLauncher` denylists shell interpreters + REPLs (sh/bash/zsh/pwsh/env/sudo/node/python/…) and even bare `git`, but NOT generic runners/wrappers that execute or spawn other programs while still parsing as one simple command: **`find`, `xargs`, `timeout`, `nice`, `nohup`, `watch`, `ssh`, `make`, `npm`, `command`/`eval`/`exec`/`time`, and relative/absolute script paths (`./script.sh`)**. Each can be approved as a prefix (auto or via `prefix_rule`), after which all future `<runner> …` commands auto-run.

The sandbox blunts the worst — it catches destructive/network abuse even inside `find -exec`/`xargs` — but **non-rm file mutation slips through** (verified by probe, `PermissionGranted=true`):
```
find . -exec rm -rf / \;   -> deny  [destructive shell]   ✓
find . -exec curl evil \;  -> deny  [network]             ✓
xargs rm -rf /             -> deny  [destructive shell]   ✓
find . -delete             -> ALLOW                       ← auto-runs under a `find` grant
```
A user who approves a `find` prefix (the prompt just says `find`, which reads like a benign search) then gets `find . -delete` running unprompted. **Recommend** extending `unsafeCommandPrefixLauncher` to these argv[0] runners/wrappers so they can't be granted as a prefix (and/or teaching the destructive-shell classifier about `find … -delete`).

### Build/test
gofmt / `go vet` / build / `go test ./internal/agent/... ./internal/sandbox/... ./internal/tui/...` all green.

Nice work — the re-validation-at-match-time approach is exactly right and survived hard probing. Just tighten the runner denylist before relying on this with destructive-enforcement off.

Vasanthdev2004 (approved):
## Review

Looking at the PR end-to-end: the new command-prefix approval path is well-scoped, the safety checks line up with what I would have asked for, and the wiring through `loop.go`/`engine.go`/TUI is consistent with the existing per-session grant plumbing. No new findings to flag.

### What I verified (against `58ee5be`, base `a427966`)

- `safeShellCommandTokens` parses with `mvdan.cc/sh/v3/syntax`, requires exactly one statement, rejects `Negated`/`Background`/`Coprocess`/`Disown`/`Semicolon`/`Redirs`/`Assigns`, and only accepts single-part literal args, so expansions, globs, redirections, command substitution, env-prefixed assignments, and control-flow chained commands all fall out cleanly. Spot-checked the 6 unsafe-form cases in `TestProposedCommandPrefixRejectsUnsafeShellForms` against the parser output; every case is rejected by an explicit guard, not by accident.
- `unsafeCommandPrefix` denylist is layered: per-token character check (`\x00\r\n*?[]{}$(...)`), whole-prefix equality against `bannedCommandPrefixSuggestions` (python/py/bash/zsh/pwsh/powershell/env/sudo/node/perl/ruby/php/lua/osascript and the major `-c`/`-l`/`-lc`/`-e`/`-r` variants), plus a launcher guard on the first token. Redundancy between the two prefix lists is fine; they are cheap and read-side defense in depth.
- `safeRequestedPrefix` re-validates the model's `prefix_rule`: must be non-empty, must itself pass `unsafeCommandPrefix`, must be elementwise-equal to the head of the parsed command, and must be shorter than the command. This closes the obvious "agent requests `[\"git\"]` to capture every git command" hole, covered by `TestProposedCommandPrefixRejectsUnsafeRequestedPrefix`.
- `matchSessionCommandPrefix` re-parses the command with the same `safeShellCommandTokens` before consulting the engine, so a stored grant cannot be triggered by an unrelated command that happens to share a raw-prefix token string. The engine then does exact-token matching (`hasStringPrefix`), so `go test` matches a `go test ./...` grant but `gotest` does not.
- Per-engine `commandPrefixGrantSet` uses `sync.Mutex` around `add`/`match`, returns a defensive copy of the matched prefix (`append([]string(nil), grant.Prefix...)`), and rejects empty `ToolName`/empty `Prefix`. Race-safe.
- Loop integration: `matchSessionCommandPrefix` is consulted BEFORE the preflight sandbox evaluate, so a session grant short-circuits the prompt; `availablePermissionDecisions` only adds `AllowPrefix` for `bash` when `len(event.CommandPrefix) > 0`; the `case PermissionDecisionAllowPrefix:` switch rejects empty `request.CommandPrefix` defensively; `normalizePermissionDecisionAction` accepts the new action; the `OnPermission` event carries `CommandPrefix` and `PermissionEvent.DecisionAction` is set.
- TUI: new "allow command prefix for session" option with hotkey `p`, label `"Yes, and allow \`<prefix>\` in this session"`; rendered transcript row shows "allowed prefix" for `AllowPrefix` decisions.

### Verification

- Worktree `origin/command-prefix-approvals`: `go build ./...` clean, `go test ./internal/agent/... ./internal/sandbox/... ./internal/tui/...` PASS.

### Non-blocking follow-ups (optional, for a future hardening PR)

- The launcher denylist covers `/bin/bash` and `/bin/zsh` but not other interpreter absolute paths (`/usr/bin/python3`, `C:\Python\python.exe`, `/usr/local/bin/node`). Best-effort UX improvement, not a hard sandbox boundary; noted only for completeness.
- `cleanPrefixRule` returns `nil` on the first empty slot (rather than skipping just that slot), so `["go", " ", "test"]` collapses to nil instead of `["go", "test"]`. Defensive but worth a one-line comment so a future reader does not "fix" it into a silent skip.
- No direct test for `matchSessionCommandPrefix` round-trip from `Engine.GrantCommandPrefixForSession` through to `LookupCommandPrefixForSession`; the integration coverage in `TestRunCommandPrefixApprovalSkipsLaterMatchingBashPrompt` covers it end-to-end, so I would not block on this.

Approving.

gnanam1990 (approved):
**Approving** (verdict on my detailed comment above).

Verified the core security property holds: every later command is re-parsed by the same single-simple-command validator before it can match a stored grant — no chaining/obfuscation bypass across ~35 probed strings, the prompt shows the exact prefix, grants are session-scoped, and the sandbox still enforces (network/destructive checks fire before the grant). The one gap (generic runner prefixes like `find`/`xargs` → `find . -delete` auto-runs) is a worthwhile follow-up to extend `unsafeCommandPrefixLauncher`, but it's gated on explicit user approval and the sandbox catches the destructive/network cases, so it's non-blocking. Nice, careful design.

PR #251, opened 2026-06-18
sessions: make the events log authoritative for sequencing + fsync checkpoints/rewind

gnanam1990:
A data-integrity audit of `internal/sessions` surfaced two crash-durability gaps (the locking, rewind sequence/path logic, replay fidelity, and perms all audited clean). Both are fixed here.

## 1. Sequence numbers could be reused after a crash → mis-targeted `/rewind`
`appendEventLocked` derived the next sequence from `metadata.EventCount`, but that count is persisted in a *separate* file (`metadata.json`) **after** the event line is appended to `events.jsonl`. A crash between the two writes leaves `EventCount` behind the log, so the next append **reuses the sequence** (a duplicate `sid:N` + colliding id). `findRewindTarget` returns the *first* match, so a rewind/fork then targets the wrong event. (The truncate path's reverse order could similarly leave a gap.)

**Fix:** the events log is now the source of truth. New `lastEventSequence()` does an **O(1) tail read** (ignoring a torn trailing partial, matching `ReadEvents`), and `appendEventLocked` uses `max(EventCount, lastLoggedSequence) + 1`. A stale/desynced `EventCount` can no longer cause a reused number; it falls back to the metadata count only if the log is unreadable, and the normal path is unchanged.

## 2. Checkpoints and rewind restores were not crash-durable
`checkpoint.go` (blob writes, `copyBlobs`) and `rewind.go` (truncated-events write, workspace `writeFileAtomic`) used plain `os.WriteFile + os.Rename` with **no fsync**, while `store.go` deliberately fsyncs the temp file **and** the parent dir for session metadata (a crash otherwise yields "a torn or empty file that corrupts the whole session"). So the `/rewind` safety net and restored workspace files weren't durable.

**Fix:** new `writeFileAtomicSync()` (fsync temp + `syncDir`) is used at all four sites; the workspace restore keeps its exact-perm chmod.

## Tests
`sequence_durability_test.go`: stale-low `EventCount` → no duplicate (the reproducible case), stale-high → benign gap, and `lastEventSequence` over empty / multi / torn-tail / >64 KiB-event logs.

`gofmt` · `go vet` · `staticcheck` · `go test ./internal/sessions/... -race` · build all · `internal/tui` + `internal/agent` + `internal/cli` — all green.

## Release Notes

* **Bug Fixes**
  * Improved session durability and crash recovery by ensuring checkpoint and rewind operations persist atomically
  * Enhanced event sequencing logic to prevent duplicate or reused sequence numbers when metadata becomes inconsistent with the event log
  * Strengthened fsync guarantees for restored files to ensure durability across system crashes

* **Tests**
  * Added comprehensive durability tests for session event sequencing and metadata consistency

### DIFF
--- a/internal/sessions/checkpoint.go
+++ b/internal/sessions/checkpoint.go
@@ -185,13 +185,8 @@ func (store *Store) writeBlob(sessionID string, content []byte) (string, error) 
 	if _, err := os.Stat(path); err == nil {
 		return hash, nil // dedup: identical content already stored
 	}
-	tmp := fmt.Sprintf("%s.tmp-%d", path, store.idCounter.Add(1))
-	if err := os.WriteFile(tmp, content, 0o600); err != nil {
+	if err := store.writeFileAtomicSync(path, content, 0o600); err != nil {
 		return "", fmt.Errorf("write checkpoint blob: %w", err)
-	}
-	if err := os.Rename(tmp, path); err != nil {
-		_ = os.Remove(tmp)
-		return "", fmt.Errorf("commit checkpoint blob: %w", err)
 	}
 	return hash, nil
 }
@@ -230,13 +225,8 @@ func (store *Store) copyBlobs(srcSessionID, dstSessionID string) error {
 			}
 			madeDir = true
 		}
-		tmp := fmt.Sprintf("%s.tmp-%d", dstPath, store.idCounter.Add(1))
-		if err := os.WriteFile(tmp, content, 0o600); err != nil {
+		if err := store.writeFileAtomicSync(dstPath, content, 0o600); err != nil {
 			return fmt.Errorf("write checkpoint blob: %w", err)
-		}
-		if err := os.Rename(tmp, dstPath); err != nil {
-			_ = os.Remove(tmp)
-			return fmt.Errorf("commit checkpoint blob: %w", err)
 		}
 	}
 	return nil
--- a/internal/sessions/rewind.go
+++ b/internal/sessions/rewind.go
@@ -217,13 +217,8 @@ func (store *Store) truncateEventsLocked(sessionID string, keepThroughSequence i
 		encoded = append(bytes.Join(kept, []byte{'\n'}), '\n')
 	}
 	path := store.eventsPath(sessionID)
-	tmp := fmt.Sprintf("%s.tmp-%d", path, store.idCounter.Add(1))
-	if err := os.WriteFile(tmp, encoded, 0o600); err != nil {
+	if err := store.writeFileAtomicSync(path, encoded, 0o600); err != nil {
 		return fmt.Errorf("write truncated events: %w", err)
-	}
-	if err := os.Rename(tmp, path); err != nil {
-		_ = os.Remove(tmp)
-		return fmt.Errorf("commit truncated events: %w", err)
 	}
 	session, err := store.readMetadata(sessionID)
 	if err != nil {
@@ -284,11 +279,13 @@ func (store *Store) writeFileAtomic(path string, content []byte, mode uint32) er
 		perm = info.Mode().Perm()
 	}
 	tmp := fmt.Sprintf("%s.zero-restore-tmp-%d", path, store.idCounter.Add(1))
-	if err := os.WriteFile(tmp, content, perm); err != nil {
+	// fsync the temp so a restored file's bytes are durable, not just in the page
+	// cache, before it is renamed into place.
+	if err := writeFileSync(tmp, content, perm); err != nil {
 		return err
 	}
-	// WriteFile only applies perm on creation and is subject to umask; force the
-	// exact bits so an executable script's mode is faithfully restored.
+	// writeFileSync applies perm only on creation and is subject to umask; force
+	// the exact bits so an executable script's mode is faithfully restored.
 	if err := os.Chmod(tmp, perm); err != nil {
 		_ = os.Remove(tmp)
 		return err
@@ -297,5 +294,6 @@ func (store *Store) writeFileAtomic(path string, content []byte, mode uint32) er
 		_ = os.Remove(tmp)
 		return err
 	}
-	return nil
+	// fsync the parent dir so the rename itself survives a crash.
+	return syncDir(filepath.Dir(path))
 }
--- a/internal/sessions/sequence_durability_test.go
+++ b/internal/sessions/sequence_durability_test.go
@@ -1,0 +1,128 @@
+package sessions
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func seqTestStore(t *testing.T) *Store {
+	t.Helper()
+	return NewStore(StoreOptions{RootDir: t.TempDir(), Now: fixedClock("2026-06-04T10:00:00Z")})
+}
+
+func appendN(t *testing.T, store *Store, sid string, n int) {
+	t.Helper()
+	for i := 0; i < n; i++ {
+		if _, err := store.AppendEvent(sid, AppendEventInput{Type: EventMessage, Payload: map[string]any{"content": "ok"}}); err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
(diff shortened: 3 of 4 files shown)


Vasanthdev2004 (approved):
## Review

Solid durability fix. Treating the events log as the source of truth for the next sequence number is the right call, and adopting the temp-write + fsync + rename + fsync-dir pattern uniformly across metadata, checkpoint blobs, and rewind restores closes the power-off crash window on every session file write. No new findings to flag.

### What I verified (against `e996775`, base `4582727`)

- `appendEventLocked` now derives `sequence := session.EventCount + 1` and then *overrides* it with `lastEventSequence(...) + 1` whenever the log is at or ahead. So the stale-LOW repro (log ahead of metadata after a crash between the append and the metadata write) is self-healing on the next append — verified by `TestAppendDerivesSequenceFromLogAfterStaleMetadata`. The mirror stale-HIGH case (e.g. an interrupted truncate) advances past the high-water mark and leaves a benign gap, verified by `TestAppendWithStaleHighMetadataLeavesGapNotDuplicate`. No duplicate sequences, no `/rewind` mis-targeting.
- `lastEventSequence` reads only the tail with a 64 KiB window that doubles until it contains the last complete line; `TestLastEventSequence_LargeLastEvent` forces a 100 KiB last event and confirms the window still finds the sequence. A complete-but-corrupt last JSON line falls back to the full `ReadEvents` scan (recovery-aware), so a single bad line never silently mis-sequences the next event. A torn tail (no final `\n`) is ignored, matching `ReadEvents`.
- `ReadEvents` adopts the same torn-tail contract: it flags `tornTailPossible` when the file does not end with `\n`, drops a malformed final non-empty line in that case, and still fails loudly on any earlier corrupt line. Symmetric with `lastEventSequence`, so they cannot disagree on what counts as a complete event.
- `writeFileAtomicSync` (write temp → `file.Sync()` → rename → `syncDir(parent)`) is now used by `writeMetadata`, `writeBlob`, `copyBlobs`, `TruncateEvents`, and `writeFileAtomic`. `syncDir` is a documented no-op on Windows (dir-handle fsync unsupported) and best-effort there — same posture as before.
- `writeFileAtomic` (the restore path) preserves the prior mode: when the checkpoint recorded `Mode != 0` it uses that; when `Mode == 0` (old checkpoint) it falls back to the existing file's `Perm()` and then `0o644` if the file is being created; it also `Chmod`s after `writeFileSync` so `umask` cannot strip the executable bits from a restored script. Good.
- Lock ordering is preserved: `pruneOrphanBlobsLocked` / `restoreToSequenceLocked` / `truncateEventsLocked` / `appendEventLocked` are documented as caller-locked, and `ApplyRewind` holds `lockSession` once across restore + truncate + prune + marker. Non-reentrant mutex, so this matters.

### Verification

- Worktree `origin/fix/sessions-durability-sequencing`: `go build ./...` clean, `go test ./internal/sessions/...` PASS (including `TestAppendDerivesSequenceFromLogAfterStaleMetadata`, `TestAppendWithStaleHighMetadataLeavesGapNotDuplicate`, `TestLastEventSequence_EmptyAndMultiple`, `TestLastEventSequence_IgnoresTornTail`, `TestLastEventSequence_LargeLastEvent`).

### Non-blocking follow-ups (optional, for a future hardening PR)

- `lastEventSequence` already falls back to the full read on a corrupt last line, but the `for { window *= 2; continue }` path has no upper bound guard other than `window > size`. Harmless today (capped by file size), but a one-line `if window > 1<<20 { ... }` ceiling would harden it against a runaway log size in pathological cases. Not blocking.
- `TruncateEvents` rewrites events.jsonl atomically and then rewrites metadata.json atomically; if the second write fails, the events file is already truncated, which means the next `AppendEvent` derives a sequence from the truncated log (correct) but `metadata.EventCount` still reflects the pre-truncate count until the rewrite succeeds. The next `appendEventLocked` will set `EventCount = sequence` from the actual log, so it self-heals, but a transient `writeMetadata` error in `TruncateEvents` could leave `EventCount` briefly stale. Acceptable; the events log is now the source of truth anyway.

Approving.

Vasanthdev2004 (approved):
## Review

Solid durability fix. Treating the events log as the source of truth for the next sequence number is the right call, and adopting the temp-write + fsync + rename + fsync-dir pattern uniformly across metadata, checkpoint blobs, and rewind restores closes the power-off crash window on every session file write. No new findings to flag.

### What I verified (against `e996775`, base `4582727`)

- `appendEventLocked` now derives `sequence := session.EventCount + 1` and then *overrides* it with `lastEventSequence(...) + 1` whenever the log is at or ahead. So the stale-LOW repro (log ahead of metadata after a crash between the append and the metadata write) is self-healing on the next append — verified by `TestAppendDerivesSequenceFromLogAfterStaleMetadata`. The mirror stale-HIGH case (e.g. an interrupted truncate) advances past the high-water mark and leaves a benign gap, verified by `TestAppendWithStaleHighMetadataLeavesGapNotDuplicate`. No duplicate sequences, no `/rewind` mis-targeting.
- `lastEventSequence` reads only the tail with a 64 KiB window that doubles until it contains the last complete line; `TestLastEventSequence_LargeLastEvent` forces a 100 KiB last event and confirms the window still finds the sequence. A complete-but-corrupt last JSON line falls back to the full `ReadEvents` scan (recovery-aware), so a single bad line never silently mis-sequences the next event. A torn tail (no final `\n`) is ignored, matching `ReadEvents`.
- `ReadEvents` adopts the same torn-tail contract: it flags `tornTailPossible` when the file does not end with `\n`, drops a malformed final non-empty line in that case, and still fails loudly on any earlier corrupt line. Symmetric with `lastEventSequence`, so they cannot disagree on what counts as a complete event.
- `writeFileAtomicSync` (write temp → `file.Sync()` → rename → `syncDir(parent)`) is now used by `writeMetadata`, `writeBlob`, `copyBlobs`, `TruncateEvents`, and `writeFileAtomic`. `syncDir` is a documented no-op on Windows (dir-handle fsync unsupported) and best-effort there — same posture as before.
- `writeFileAtomic` (the restore path) preserves the prior mode: when the checkpoint recorded `Mode != 0` it uses that; when `Mode == 0` (old checkpoint) it falls back to the existing file's `Perm()` and then `0o644` if the file is being created; it also `Chmod`s after `writeFileSync` so `umask` cannot strip the executable bits from a restored script. Good.
- Lock ordering is preserved: `pruneOrphanBlobsLocked` / `restoreToSequenceLocked` / `truncateEventsLocked` / `appendEventLocked` are documented as caller-locked, and `ApplyRewind` holds `lockSession` once across restore + truncate + prune + marker. Non-reentrant mutex, so this matters.

### Verification

- Worktree `origin/fix/sessions-durability-sequencing`: `go build ./...` clean, `go test ./internal/sessions/...` PASS (including `TestAppendDerivesSequenceFromLogAfterStaleMetadata`, `TestAppendWithStaleHighMetadataLeavesGapNotDuplicate`, `TestLastEventSequence_EmptyAndMultiple`, `TestLastEventSequence_IgnoresTornTail`, `TestLastEventSequence_LargeLastEvent`).

### Non-blocking follow-ups (optional, for a future hardening PR)

- `lastEventSequence` already falls back to the full read on a corrupt last line, but the `for { window *= 2; continue }` path has no upper bound guard other than `window > size`. Harmless today (capped by file size), but a one-line `if window > 1<<20 { ... }` ceiling would harden it against a runaway log size in pathological cases. Not blocking.
- `TruncateEvents` rewrites events.jsonl atomically and then rewrites metadata.json atomically; if the second write fails, the events file is already truncated, which means the next `AppendEvent` derives a sequence from the truncated log (correct) but `metadata.EventCount` still reflects the pre-truncate count until the rewrite succeeds. The next `appendEventLocked` will set `EventCount = sequence` from the actual log, so it self-heals, but a transient `writeMetadata` error in `TruncateEvents` could leave `EventCount` briefly stale. Acceptable; the events log is now the source of truth anyway.

Approving.

PR #252, opened 2026-06-18
Persist command prefix approvals

anandh8x:
## Summary
- add persistent command-prefix grants to the sandbox grants store and engine lookup path
- add a separate persistent prefix approval decision in the agent and TUI
- show saved prefixes in prompt context and sandbox grants list output
- tighten reusable prefix validation for runner-style launchers and script paths, plus classify find -delete as destructive

## Validation
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/sandbox ./internal/tui ./internal/cli
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Persistent command-prefix approvals that survive across sessions
  * "Always allow command prefix" option in permission prompts
  * Approved command prefixes displayed in system prompts to the AI agent
  * CLI support for listing command-prefix grants

* **Bug Fixes**
  * Enhanced detection of destructive commands (e.g., `find -delete`)
  * Stricter validation rejecting unsafe command launchers (e.g., interpreters, `sudo`, `xargs`)

### DIFF
--- a/internal/agent/command_prefix.go
+++ b/internal/agent/command_prefix.go
@@ -7,55 +7,6 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
-var bannedCommandPrefixSuggestions = [][]string{
-	{"python3"},
-	{"python3", "-"},
-	{"python3", "-c"},
-	{"python"},
-	{"python", "-"},
-	{"python", "-c"},
-	{"py"},
-	{"py", "-3"},
-	{"pythonw"},
-	{"pyw"},
-	{"pypy"},
-	{"pypy3"},
-	{"git"},
-	{"bash"},
-	{"bash", "-lc"},
-	{"sh"},
-	{"sh", "-c"},
-	{"sh", "-lc"},
-	{"zsh"},
-	{"zsh", "-lc"},
-	{"/bin/zsh"},
-	{"/bin/zsh", "-lc"},
-	{"/bin/bash"},
-	{"/bin/bash", "-lc"},
-	{"pwsh"},
-	{"pwsh", "-Command"},
-	{"pwsh", "-c"},
-	{"powershell"},
-	{"powershell", "-Command"},
-	{"powershell", "-c"},
-	{"powershell.exe"},
-	{"powershell.exe", "-Command"},
-	{"powershell.exe", "-c"},
-	{"env"},
-	{"sudo"},
-	{"node"},
-	{"node", "-e"},
-	{"perl"},
-	{"perl", "-e"},
-	{"ruby"},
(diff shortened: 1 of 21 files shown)


gnanam1990 (approved):
## Review — APPROVE

Persistence does **not** open a hole — I verified the load-bearing claims in code and the probes confirm them.

### ✅ Verified safe
- **The persistent auto-approval path re-validates the command.** `matchCommandPrefix` (`command_prefix.go:42`) re-parses every incoming bash command with `safeShellCommandTokens` → `ValidCommandPrefix` (the updated denylist) **before** consulting either a persistent (`LookupCommandPrefix`) or session grant. So even a stored unsafe/runner prefix is neutralized: the matching command (`sh -c …`, `find . -delete`, `./script`) is rejected at parse → no match → re-prompt. (Probed: injecting `["find"]`/`["xargs"]`/`["sh"]` grants directly on disk, the matching dangerous commands all still re-prompt.)
- **The #250 denylist gap is fixed.** `unsafeCommandPrefixLauncher` now covers find/xargs/timeout/nice/nohup/watch/setsid/stdbuf/ionice/ssh/make/npm/npx (plus any program containing `/`), applied at **both** approval and match time, and `find -delete` is now classified destructive (`analyzer.go:96 hasFindDelete`) so the sandbox denies it. 👍 thanks for picking that up.
- **Sandbox still enforces.** Network-deny + destructive checks fire before the grant is consulted; a matched persistent `curl`/`rm -rf` grant still denies (the PR's own `…DoesNotBypassNetworkDeny` test + probe confirm).
- **Store hardening.** File `0600`, dir `0700`, atomic temp+rename; the path is per-user-global and can't be redirected by the model/workspace (resolved from env at process start; the bash child can't mutate the parent env); a tampered/corrupt grants file fails closed at load; only safe prefix tokens + a `RedactString`'d reason are persisted (not the full command/args/cwd); `-race` clean.

### 🟡 LOW (non-blocking) — global + persistent breadth is under-displayed in the prompt
Prefix grants are global (no workspace scope) and now persistent, so approving `git push` / `npm` / `curl` (under `Network=allow`) in a trusted repo means the agent auto-runs the matching prefix with no prompt in a *later, untrusted cloned repo*. The prompt label "always allow `git status`" conveys *persistent* but not *across every project* — a fidelity gap like the #248 concern. Suggest wording it to disclose the global+persistent scope (e.g. "…in every project") and/or scoping persistent prefix grants to the workspace. Non-blocking: the grants are still limited to safe single commands and the sandbox's destructive/network gates remain.

Solid follow-up — the persistence is well-built and the matching safety is correctly preserved. Approving.

PR #253, opened 2026-06-18
Align sandbox permission blocks and backend states

anandh8x:
## Summary
- route network-risk bash calls through permission prompting instead of preflight hard-deny
- apply the existing turn/session network permission overlay after approved network prompts
- move network command detection into the shell AST analyzer so quoted process patterns do not trigger network risk

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox -run 'TestAnalyzeCommand|TestClassifyDoesNotFlagQuotedHttpServerPattern|TestEvaluateExemptsNetworkToolsFromDenyByDefault|TestEngineClassifiesNetworkAndDestructiveShellCommands|TestGrantRequestPermissionsNetworkOverlaysPolicyForTurn'
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent -run 'TestRunPersistentCommandPrefixStillPromptsForNetwork|TestRunApprovedNetworkBashPromptAppliesTurnNetworkGrant'
- GOCACHE=/tmp/zero-go-cache go test ./internal/tools -run 'TestBash|TestRegistryRunsBashThroughSandboxEngine'

## Summary by CodeRabbit

* **New Features**
  * Improved network-usage detection for more command types (npm, pip, git, GitHub CLI, etc.)
  * Turn-scoped and session-scoped sandbox permission grants for network and filesystem access

* **Bug Fixes**
  * Network and filesystem sandbox denials now properly recoverability handling with targeted permission approval flows

* **Changes**
  * Removed autonomy tier-based permission model; permissions now allow or deny without security levels
  * Removed bash auto-allow for sandboxed shells; now requires explicit approval
  * Removed proxy-based network scoping and domain allowlists
  * Changed internal denial representation from "violations" to "blocks"
  * System now fails cleanly when native sandbox unavailable instead of degrading to policy-only mode


anandh8x:
@gnanam1990 @Vasanthdev2004 I pushed the final sandbox/permission cleanup commit here.

What changed:
- Removed the old `policy-only` backend/support terminology from the active sandbox path.
- Missing native sandbox support is now reported as `unavailable` and command execution still fails closed when native sandboxing is required.
- Renamed structured sandbox denial metadata from `violation` to `block` across sandbox decisions, permission events, stream-json, snapshots, TUI/session parsing, CLI output, and tests.
- Updated user-facing wording for network/workspace permission reasons so the UI no longer presents old policy-denial text for promptable cases.
- Updated the Windows sandbox policy golden, sandbox smoke selectors, README, and work-split docs to match the new terminology.

Why:
- The old `policy-only` wording made unavailable native isolation look like a supported fallback path, which is misleading and risky.
- `block` better matches the actual permission flow: the sandbox reports what action/path/network request is blocked or promptable, while approvals/grants decide whether the action may proceed.
- This keeps shell execution fail-closed unless a real native sandbox backend is active, while making promptable network/outside-workspace cases clearer in TUI/headless output.

Verification completed:
- `GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/agent ./internal/tools ./internal/tui ./internal/cli ./internal/doctor ./internal/zerocommands ./internal/streamjson`
- `scripts/sandbox-smoke.sh`
- `git diff --check`

Vasanthdev2004:
doing

internal/sandbox/linux_helper.go
```diff
@@ -283,11 +269,8 @@ func appendUnreadableLinuxPathArgs(args []string, path string) []string {
 	return append(args, "--perms", "000", "--tmpfs", path, "--remount-ro", path)
 }
 
-func shouldUnshareLinuxNetwork(policy NetworkPolicy, allowNetworkForProxy bool) bool {
-	if allowNetworkForProxy {
-		return true
-	}
-	return policy.Mode == NetworkDeny || policy.Mode == NetworkScoped || policy.ProxyRequired
+func shouldUnshareLinuxNetwork(policy NetworkPolicy) bool {
```
Vasanthdev2004, line 272: [P1] NetworkScoped policy yields unrestricted network egress on Linux bwrap

`shouldUnshareLinuxNetwork` was reduced to `policy.Mode == NetworkDeny`, so a `NetworkScoped` permission profile no longer adds `--unshare-net`. At the same time the scoped-egress proxy (egress.go/mitm.go/egress_socks.go) and `effectiveNetwork()`'s NetworkScoped-to-NetworkDeny collapse were removed, and `PermissionProfileFromPolicy` now forwards `policy.Network` unchanged. The engine's network gate only fires for `NetworkDeny`, so NetworkScoped shell commands are not gated either. With no netns isolation, no filtering proxy, and no engine prompt, a `network: "scoped"` configuration (still an accepted mode with retained `AllowedDomains`/`DeniedDomains` fields) gives the sandboxed process full host network on Linux instead of restricting egress to the allowlist or failing closed. macOS still denies (seatbelt default branch) and Windows errors (unsupported mode), so the regression is Linux-specific but is a real sandbox escape for operators who configure scoped networking.

Suggested fix: either keep `NetworkScoped` in `shouldUnshareLinuxNetwork` (and fail-closed if no proxy is available) or explicitly reject `NetworkScoped` as unsupported with a clear error.

internal/streamjson/streamjson.go
```diff
@@ -59,47 +59,47 @@ type CheckpointInfo struct {
...
+	Action            string            `json:"action,omitempty"`
+	Permission        string            `json:"permission,omitempty"`
+	PermissionGranted *bool             `json:"permissionGranted,omitempty"`
+	PermissionMode    string            `json:"permissionMode,omitempty"`
+	Autonomy          string            `json:"autonomy,omitempty"`
+	SideEffect        string            `json:"sideEffect,omitempty"`
+	Reason            string            `json:"reason,omitempty"`
+	DecisionReason    string            `json:"decisionReason,omitempty"`
+	Risk              *sandbox.Risk     `json:"risk,omitempty"`
+	Block             *sandbox.Block    `json:"block,omitempty"`
```
Vasanthdev2004, line 83: [P2] Breaking output-schema field rename without a schemaVersion bump

The PR renames `Event.Violation` to `Event.Block` and its JSON tag from `"violation,omitempty"` to `"block,omitempty"`, but `const SchemaVersion` stays at 1 and `docs/STREAM_JSON_PROTOCOL.md` is not updated. The stream-json protocol is documented as a stable, schema-versioned contract for external headless consumers (editor extensions, automation wrappers); any client reading `event["violation"]` will silently receive nothing after this change, and the unchanged schemaVersion gives them no drift signal. Either bump `SchemaVersion` to 2 (and update the protocol doc) or emit both fields during a transition window.

internal/sandbox/grants.go
```diff
@@ -542,19 +527,6 @@ func (store *GrantStore) readState() (grantFile, error) {
 	buckets := map[string][]Grant{}
```
Vasanthdev2004, line 527: [P2] v1 grant file migration removed, breaking existing v1 grant files

`GrantStore.readState()` drops the `case 1:` migration branch (this hunk), so a persisted `sandbox-grants.json` with `schemaVersion: 1` now falls through to the `default` case and returns "unsupported schemaVersion". Any user who created grants under the earlier v1 schema will lose access to their persisted grants on upgrade. The migration could have been kept trivially: Go's `json.Unmarshal` ignores unknown fields, so a v1 grant's now-removed `maxAutonomy` would simply be dropped during unmarshal into the v2 `Grant` struct. The PR also deletes `TestMigrateV1FileToToolWideGrant` without replacing it, so the regression is uncovered.

internal/sandbox/engine.go
```diff
@@ -313,143 +256,113 @@ func (engine *Engine) Evaluate(ctx context.Context, request Request) Decision {
...
 			GrantMatched: true,
-			Grant:        &grant,
+			Grant:        sessionAllow,
+		}
+	}
+	if promptableBlock != nil {
+		return promptPathBlock(request, risk, promptableBlock)
+	}
+	netMode := engine.effectiveNetworkMode(policy)
+	if netMode == NetworkDeny && HasRiskCategory(risk, "network") && !engine.toolNetworkExempt(request) {
```
Vasanthdev2004: [P2] Tool-wide bash allow grant now bypasses the NetworkDeny network prompt

The grant lookup was moved ahead of the recoverable network prompt in this hunk. `DeriveScope("bash", ...)` returns a tool-wide (`ScopeToolWide`) scope, so a previously persisted bash "always allow" grant matches every bash request and returns `ActionAllow` before the `netMode == NetworkDeny && HasRiskCategory(risk, "network")` gate at line 338 runs. Consequently a network shell command (e.g. `curl evil.test`) under NetworkDeny is no longer surfaced as a network approval prompt when any bash allow grant exists. Native-sandbox kernel isolation still blocks egress under NetworkDeny today, so this is defense-in-depth weakening rather than an active escape, but it removes the operator-facing confirmation layer the NetworkDeny policy was meant to provide and would become an escape on any backend that runs commands without kernel network isolation under NetworkDeny. No test pins that a bash grant does not suppress the network prompt.

internal/sandbox/risk.go
```diff
@@ -98,9 +97,6 @@ func classifyWithScope(request Request, scope *Scope) Risk {
 	// cannot be bypassed by choosing a different alias key.
```
Vasanthdev2004, line 105: [P2] Unparseable network commands no longer get the "network" risk category

Removing `networkCommandPattern` (this hunk) leaves the AST analyzer as the sole source of the network category, but `analyzeInto` returns early on a parse error with only `TooComplex` set (Network stays false). So an obfuscated or malformed network command (e.g. a `curl` invocation with broken quoting that the shell parser rejects but the old regex still matched) is no longer classified as network and skips the NetworkDeny engine gate, falling through to auto-allow/prompt with only the `unparseable_command` (RiskHigh) category. Egress is still kernel-blocked under NetworkDeny on native backends, but the up-front network classification and prompt are lost for obfuscated commands, weakening the defense-in-depth classification the regex previously provided.

Vasanthdev2004 (changes requested):
Requesting changes on one P1 and four P2 findings.

**[P1] NetworkScoped yields unrestricted egress on Linux** — shouldUnshareLinuxNetwork was reduced from NetworkDeny || NetworkScoped || ProxyRequired to just NetworkDeny, while the scoped egress proxy and effectiveNetwork's scoped-to-deny collapse were simultaneously removed. NetworkScoped is still an accepted config mode with retained AllowedDomains/DeniedDomains fields, so an operator who configures network: scoped gets full host network on Linux bwrap instead of allowlisted egress or fail-closed.

**[P2] Breaking stream-json field rename without schemaVersion bump** — Event.Violation to Event.Block (JSON tag violation to block) but SchemaVersion stays at 1 and STREAM_JSON_PROTOCOL.md is not updated.

**[P2] v1 grant file migration removed** — GrantStore.readState() drops the case 1 migration, so persisted v1 grant files now return unsupported schemaVersion.

**[P2] Tool-wide bash allow grant bypasses NetworkDeny prompt** — Grant lookup moved ahead of the recoverable network prompt, so a persisted bash allow grant short-circuits before the network gate fires.

**[P2] Unparseable network commands skip network classification** — Removing networkCommandPattern leaves the AST analyzer as sole network detector, but it returns early on parse errors.

The agent loop, tools, CLI, config, TUI, and mechanical rename (Violation to Block) portions are clean.

anandh8x:
Addressed @Vasanthdev2004's requested changes in 433060c. Summary: unknown/programmatic network modes now fail closed as deny; Linux network sandboxing and backend diagnostics use normalized network mode; bash allow grants no longer bypass network/destructive prompts; unparseable shell commands with obvious network programs are classified as network risk; v1 sandbox grant files are readable and rewrite as v2; stream-json is bumped to schemaVersion 2 and docs/tests cover the block metadata field. Validation: go test ./internal/sandbox ./internal/streamjson ./internal/specialist plus targeted CLI stream-json tests, go test ./..., scripts/sandbox-smoke.sh, and git diff --check.

Vasanthdev2004 (approved):
Follow-up review of the changes since the previous REQUEST_CHANGES (review id 4530216206). All five previously raised findings are addressed in 433060c.

## Verified fixes

**[P1] NetworkScoped Linux unshare gap — FIXED**
`shouldUnshareLinuxNetwork` was reduced to `NormalizeNetworkMode(policy.Mode) == NetworkDeny`, but the `NetworkScoped` mode constant and the scoped-egress proxy (`egress.go`, `egress_socks.go`, `AllowedDomains`/`DeniedDomains`/`ProxyRequired` on `NetworkPolicy`) are removed entirely. `NormalizeNetworkMode` collapses anything that isn't `NetworkAllow` to `NetworkDeny`, so an operator who still writes `network: scoped` gets fail-closed deny with `--unshare-net` applied rather than unrestricted host egress. Tests at `internal/sandbox/linux_helper_test.go` (`NormalizeNetworkMode(scoped) == NetworkDeny` and `shouldUnshareLinuxNetwork(scoped) == true`) lock this in. Good.

**[P2] stream-json schema rename without SchemaVersion bump — FIXED**
`SchemaVersion` bumped 1 → 2 in `internal/streamjson/streamjson.go:17`, `docs/STREAM_JSON_PROTOCOL.md` updated with the new `block` field and an explicit v1→v2 rename note, and golden/test fixtures regenerated.

**[P2] v1 grant file migration removal — FIXED**
The `case 1:` branch in `GrantStore.readState` is preserved; it still unmarshals legacy v1 grant files. The only change is it no longer coerces `Scope=""`/`ScopeKind=ScopeToolWide` (the v1 grant's zero-value scope is normalized by `normalizeStoredGrant` downstream). v1 files remain readable and rewrite as v2.

**[P2] bash allow grant bypassed NetworkDeny prompt — FIXED**
`shouldRequestPermission` now checks `decision.Action == sandbox.ActionPrompt` *before* `permissionGranted`, so a matched bash grant no longer short-circuits a recoverable network prompt. `availablePermissionDecisions` also suppresses the prefix-always-allow option when `networkSandboxPrompt(event)` is true, and `grantNetworkForSandboxPrompt` issues a turn/session network grant on approval. Covered by `TestRunPersistentCommandPrefixStillPromptsForNetwork`.

**[P2] unparseable network commands skipped classification — FIXED**
`networkCommandPattern` was removed but replaced with `unparseableNetworkPattern` in `internal/sandbox/risk.go`, applied via `matchesUnparseableNetwork(command)` when `analysis.TooComplex` is true. The AST analyzer (`commandUsesNetwork` in `analyzer.go`) now also detects network use from subcommand args (`npm install`, `git clone`, `python -m http.server`, `gh api`, etc.) for parseable commands.

## Additional notes (non-blocking, P3)

- The `BackendPolicyOnly` → `BackendUnavailable` rename plus fail-closed behavior in `runner.go`/`manager.go` is a good hardening: missing native sandbox no longer looks like a supported fallback.
- Autonomy is cleanly removed from `Grant`/`GrantInput`/`Lookup` and all callers; no orphaned `MaxAutonomy` references remain in production code (only in tests asserting its absence).
- Minor: the `unparseableNetworkPattern` regex is broad by design (favors false-positive network flagging over missing it) — worth a comment noting the intentional bias, which the code already has.

Verdict: **approve**. Thanks for the thorough follow-up @anandh8x.

PR #254, opened 2026-06-18
tui: live working status (elapsed) + streaming reasoning preview so long turns never look frozen

gnanam1990:
## Problem
On a slow turn the TUI looked **stuck or broken**. Once partial text had streamed, the screen showed only a static cursor; during a reasoning phase it showed only a static `Thinking` header; and there was **no elapsed time anywhere**. With a slow provider or a long-thinking reasoning model, the user couldn't tell whether the agent was working or hung — a multi-minute think looked identical to a frozen terminal.

## Changes (pure TUI render — provider-agnostic)
- **Always-on working status line** — an animated spinner + the rotating working verb + a **live elapsed clock** (`8s`, `1m04s`), rendered on *every* pending frame, **including once partial text has streamed** (previously the spinner vanished as soon as text arrived). The spinner tick is time-based (~80 ms), so the elapsed keeps advancing **even when the upstream goes silent mid-stream** — the screen never freezes.
- **Live reasoning preview** — during a think (before any answer text) the last 1–2 lines of the streaming reasoning are shown dimmed **beneath the working line**, each **tail-truncated** so even a single continuously-growing reasoning line visibly moves as tokens arrive. The user can now see *what* the model is reasoning about. Skipped when the reasoning block is expanded (the full body already shows — no duplicate).

Both live in `interimBlock`, so they apply to **every provider/model**, not just slow ones.

## Before → After (thinking phase)
```
Thinking                              Thinking
                          ──▶         ⠏ working · 1m30s
                                        …the patch was corrupt, re-planning the edits
```

## Testing
- `go test ./internal/tui/...` (incl. `-race`) green.
- New: `TestFormatWorkingElapsed`, `TestPreviewTail`, `TestInterimBlockShowsWorkingLineWithStreamedText`, `TestInterimBlockShowsReasoningPreviewWhileThinking`, `TestInterimBlockNoPreviewWhenReasoningExpanded`.
- Updated `TestPromptSubmitStoresReasoningSeparatelyFromAnswer` for the one extra turn-start timestamp the elapsed clock consumes.
- `gofmt` / `go vet` / `go build ./...` clean.

## Summary by CodeRabbit

* **New Features**
  * Added an always-visible working status line with a live elapsed clock during pending turns
  * When the assistant is thinking and reasoning is collapsed, shows a dim, continuously updating preview tail of recent reasoning

* **Bug Fixes**
  * Ensures the elapsed clock is correctly seeded for both spec draft start and spec approval launch so timing displays reliably

* **Tests**
  * Added unit tests covering elapsed time formatting, interim rendering with partial streaming, and reasoning preview/tail truncation behavior

* **Refactor**
  * Centralized run initialization for spec mode flows

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -2783,18 +2783,25 @@ func (m model) launchPrompt(prompt string) (model, tea.Cmd) {
 	m.pendingImages = nil
 	m.pendingImageLabels = nil
 	runCtx, cancel := context.WithCancel(m.ctx)
+	m = m.beginRun(cancel)
+	return m, tea.Batch(m.runAgent(m.activeRunID, runCtx, prompt, turnImages), m.spinner.Tick)
+}
+
+// beginRun stamps the shared run-start state for a new agent turn: a fresh run
+// ID, the cancel func, pending = true, the turn-start timestamp (the source for
+// the working status line's live elapsed clock), and a reset working-verb
+// rotation so the brand word shows first. Centralized so every launch path
+// (normal prompt + spec draft/impl) keeps these in sync — a missing
+// turnStartedAt previously dropped the elapsed timer on spec-mode runs.
+func (m model) beginRun(cancel context.CancelFunc) model {
 	m.runID++
 	m.activeRunID = m.runID
 	m.runCancel = cancel
 	m.pending = true
 	m.turnStartedAt = m.now()
-	// Rewind the verb rotation so the user sees "gitlawbmaxxing" first when
-	// the new run starts (instead of mid-rotation from a prior turn). Also
-	// reset the step counter so the cadence doesn't carry over a partial
-	// countdown from the previous run.
 	m.workingVerbTicks = 0
 	m.workingVerb.Reset()
-	return m, tea.Batch(m.runAgent(m.activeRunID, runCtx, prompt, turnImages), m.spinner.Tick)
+	return m
 }
 
 func (m model) launchQueuedMessageIfReady() (model, tea.Cmd) {
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -124,11 +124,15 @@ type model struct {
 	workingVerb      *workingWords
 	workingVerbTicks int
 	pending          bool
-	queuedMessage    string
-	exiting          bool
-	runCancel        context.CancelFunc
-	runID            int
-	activeRunID      int
+	// turnStartedAt is when the in-flight run began; the working status line
+	// renders the live elapsed time from it so a long or stalled turn never looks
+	// like a frozen terminal (for ANY provider, not just slow ones). Zero = idle.
+	turnStartedAt time.Time
+	queuedMessage string
+	exiting       bool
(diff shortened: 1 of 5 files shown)


anandh8x (dismissed):
Medium: /spec launches do not seed the new elapsed timer, so the working status line regresses on spec-mode runs.

`internal/tui/model.go:1840` only appends elapsed when `turnStartedAt` is non-zero, and the normal prompt path sets it in `launchPrompt` at `internal/tui/model.go:2790`. But the two alternate agent launch paths in `internal/tui/spec_mode.go` set `pending` and start spinner ticks without setting `turnStartedAt`: draft generation around `internal/tui/spec_mode.go:70-80` and implementation around `internal/tui/spec_mode.go:206-223`. Those runs will still show spinner/verb, but not the live elapsed time this PR advertises for every pending frame. Long spec drafts and implementations can therefore look less alive than normal turns.

Please set `m.turnStartedAt = m.now()` in both spec launch paths, or centralize run-start state so `pending`, run IDs, cancellation, `turnStartedAt`, and working verb state stay in sync. A small spec-mode test would keep this covered.

Verification: `git diff --check 673fce11bd19e1329df962dd8519823a6e422566...75f1517739736052d7bae0556fd16be649f19bcd` passed. From a temporary archive of the PR head, `GOCACHE=/tmp/zero-go-cache go test ./internal/tui` passed.

gnanam1990:
Fixed in `3be47d0` — good catch. Rather than patch each call site I centralized run-start state into a `beginRun(cancel)` helper (runID, cancel, pending, `turnStartedAt`, and the working-verb reset) and routed all three launch paths through it (normal prompt + spec draft + spec impl), so this can't drift again. As a bonus it also restores the working-verb reset the spec-draft path was silently missing. Added `TestSpecLaunchesSeedElapsedClock` asserting both spec launches seed `turnStartedAt`. gofmt/vet/build + tui suite (incl. -race) green.

Vasanthdev2004 (approved):
LGTM. The `beginRun` extraction in commit 2 cleanly resolves anandh8x's CHANGES_REQUESTED from commit 1 — both spec-mode launch paths (`handleSpecCommand` draft + `approveSpecReview` impl) now route through `beginRun`, so `turnStartedAt`, `pending`, run IDs, cancel func, and working-verb reset stay in sync across all three entry points.

The always-on working status line and reasoning preview are well-contained in `interimBlock`, which is only called when `m.pending` is true (guarded at `transcript_selection.go:169`), so the stale `turnStartedAt` after a run completes is harmless — it's re-stamped on the next `beginRun`.

`formatWorkingElapsed` handles negatives, sub-minute, and minute+ correctly. `previewTail` rune-counting is a safe width proxy per the comment (reasoning deltas carry no ANSI). Tests cover the key behaviors: elapsed formatting, working line with streamed text, reasoning preview while thinking, no-duplicate-when-expanded, and the spec-mode regression.

Verified: `go test ./internal/tui/...` PASS, `go vet` clean, `gofmt` clean.

PR #255, opened 2026-06-18
feat(tui): sticky plan panel + specialist task cards + task table overlay

Vasanthdev2004:
## Problem

On a slow turn the TUI looked stuck or broken. The agent's step-by-step plan was buried inside tool call output, specialist/subagent delegations were invisible (EventSpecialistStart/Stop events silently dropped on resume), and there was no way to see all active tasks at a glance.

### 1. Sticky plan panel (`plan_panel.go`)
- **Pinned** between the title bar and transcript body — stays fixed while chat scrollback flows underneath.
- Shows the agent's plan from the `update_plan` tool as a **live checklist**: header with spinner + progress count + elapsed, text progress bar, per-step list with status icons (✓ completed, ⠹ in_progress, ○ pending, ✗ failed) and individual step timings.
- **Auto-collapses** to a one-line green summary 30s after completion. `Ctrl+P` toggles expand/collapse.
- Plan state synced from `update_plan` tool results via `CurrentPlan()`; timestamps preserved across the tool's full-replacement updates by matching on content.

### 2. Specialist cards (`specialist_card.go`)
- New `rowSpecialist` transcript row kind; `EventSpecialistStart`/`Stop` now handled in `session.go` hydration (previously fell through to `default` and were silently dropped on resume).
- **Animated card** with spinner + name + task + elapsed + token count + tool call count. Blue while running, green when done, red on error.
- `specialistTracker` tracks live state; `Task` tool calls/returns intercepted in `OnToolCall`/`OnToolResult` callbacks to start/complete tracking and emit card rows.
- Height cache marks running specialist rows unstable so the spinner animates.

### 3. Task table overlay (`task_table.go`)
- `Ctrl+G` toggles a full-width overlay showing all specialists as a table: #, Specialist, Task, Status, Time, Tokens, Tools.
- Summary bar with running/completed/error counts, total tokens, aggregate elapsed.
- Arrow keys navigate rows; `Esc` closes the overlay.

### Layout (top to bottom)
1. Title bar (existing)
2. **Sticky plan panel** (new — variable height, 0-8 lines)
3. Chat scrollback (existing — includes specialist cards)
4. **Task table overlay** (new — toggled, floats over chat)
5. Working status line (existing)
6. Input bar (existing)

## Testing
- `go test ./internal/tui/...` — all pass (26 new tests + existing suite).
- `go build ./...` clean.
- `go vet ./internal/tui/...` clean.
- `gofmt` clean.

New tests:
- `plan_panel_test.go` (7): updateFromItems, timestamp preservation, isComplete, clear, height, visibility timeout, render.
- `specialist_card_test.go` (11): tracker start/complete, tool count, tokens, clear, duplicate start, status string, parse status, token formatter, elapsed formatter, card render, task arg parsing.
- `task_table_test.go` (8): toggle, show/hide, cursor movement, selectedSessionID, height, truncateColumn, render empty, render with specialists.

## Keyboard shortcuts
| Key | Action |
|---|---|
| `Ctrl+P` | Expand/collapse plan panel |
| `Ctrl+G` | Toggle task table overlay |
| `↑`/`↓` | Navigate task table rows (when overlay visible) |
| `Esc` | Close task table overlay |

## Summary by CodeRabbit

* **New Features**
  * Added a sticky, expandable plan panel with real-time step statuses, per-step elapsed timings, and auto-hide when finished.
  * Added specialist cards in the transcript with progress during active runs, plus a drill-in sub-session view via clicking.
  * Added sub-session-aware keyboard behavior (Esc/Up to exit) and Ctrl+P plan toggle when no modal blocks input.
* **Bug Fixes**
  * Prevented specialist row renders from being cached during active runs to avoid stale UI.
* **Tests**
  * Expanded coverage for plan panel timing/state, specialist tracking/cards, sub-session navigation, and transcript row rendering.


anandh8x (dismissed):
Requesting changes. I found three user-facing regressions.

High: plan panel and live specialist cards are not rendered until a specialist row already exists.

`planUpdateMsg` updates `m.plan` at `internal/tui/model.go:1345-1350`, and `specialistStartMsg` only updates the tracker at `internal/tui/model.go:1351-1356`. The only call to `renderPlanPanel` is inside the transcript loop when `row.kind == rowSpecialist` at `internal/tui/transcript_selection.go:156-164`, and a `rowSpecialist` is appended only on `specialistCompleteMsg` at `internal/tui/model.go:1374-1379`. So normal `update_plan`-only turns never show the plan panel, and a `Task` does not show its animated specialist card/progress while it is running. Please render from live plan/tracker state directly or append/update a running row on start, and add coverage for `update_plan` without `Task` plus `Task` before completion.

High: `runChildProcess` regresses stream-json parsing.

The new loop uses `bufio.Scanner` with a 1 MiB token cap and ignores both JSON errors and `scanner.Err()` at `internal/specialist/exec.go:731-746`. This bypasses `ParseStream`, whose comment explains why Scanner was avoided and which returned errors for invalid lines or missing `type`. A child that emits one large final/tool-result line can now silently drop the rest of stdout and return a successful empty/truncated `Task` result. Please preserve streaming progress while keeping `ParseStream` semantics, especially no Scanner cap and no swallowed parse/scanner errors.

Medium: resumed specialist cards parse the persisted accounting payload incorrectly.

`specialistInfoFromPayload` reads `specialistName` and only maps `completed` to completed at `internal/tui/session.go:609-621`, but the accounting payload writes the name as `specialist` and successful stops as `success` at `internal/specialist/accounting.go:166-189`. After `/resume`, successful specialist cards become unnamed and can render as error. Please accept the actual persisted keys/status values and test start/stop rehydration from real accounting payloads.

Verification: `git diff --check 03784268ac3b996633691b2e754c9c76a361e9d3...afcd0f11c9bb742b592eb304a2fa9e43b5642e53` passed. From temporary archives of the PR head, `GOCACHE=/tmp/zero-go-cache go test ./internal/tui/... ./internal/agent ./internal/specialist ./internal/tools ./internal/swarm` passed outside the sandbox; the same command only failed inside the sandbox due to bwrap and httptest socket restrictions. `GOCACHE=/tmp/zero-go-cache go build -buildvcs=false ./...` passed.

gnanam1990 (changes requested):
## Overview
Solves a real UX problem — on a slow turn the TUI looked frozen, the agent's plan was buried in tool output, and specialist delegations were invisible (and silently dropped on resume). The sticky plan panel + specialist cards + task table are a nice, cohesive answer, and the implementation is mostly clean:

- **Concurrency is done right** — the agent-goroutine callbacks send tea messages via `runtimeMessageSink` (`specialistStartMsg`/`specialistProgressMsg`/`specialistCompleteMsg`/`planUpdateMsg`) instead of mutating the model directly, so the tracker is only touched on the update goroutine. No data race.
- The `update_plan` nested `Items` schema is valid — `propertyToRuntimeMap` already recurses into `Items`/`Properties`, so the comment you removed (about nested schemas being dropped) was stale. Good cleanup.
- 26 new tests, clean build/vet/gofmt.

Two correctness issues should be fixed before merge.

## Request changes

### 1. `runChildProcess` can silently truncate specialist output (and stall) — `internal/specialist/exec.go`
The new streaming reader caps lines at 1 MiB (`scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)`) and never checks `scanner.Err()`. If the specialist child emits a single stream-json line larger than 1 MiB (a large tool result, a long assistant turn), `scanner.Scan()` stops, the remaining events are **silently dropped**, and `SummarizeStream(run.Events)` returns a truncated result for the `Task` tool. Worse, the child keeps writing → the stdout pipe fills → the child blocks → `command.Wait()` hangs until ctx is cancelled (an apparent freeze). The previous code buffered all stdout (`command.Run()` + `ParseStream`), so this is a regression.

Suggested: check `scanner.Err()` and surface it, and either raise/remove the per-line cap or keep draining stdout after a long line so the child can't block on a full pipe.

### 2. Resume renders two cards per completed specialist — `internal/tui/session.go`
`transcriptRowsFromSessionEvents` appends a `rowSpecialist` for **both** `EventSpecialistStart` and `EventSpecialistStop` with the same `childSessionID`. `rowSpecialist` has no `transcriptRowKey` case, so the transcript dedup never collapses them, and the renderer draws one card per row. On resume, every completed specialist shows up **twice** — a "running" card (from Start) followed by a "completed" card (from Stop). The live path reconciles by `childSessionID`, but hydration doesn't.

`specialist_summary_render_test.go` uses two *distinct* specialists (`s1`, `s2`), so it doesn't exercise this. Suggested: reconcile by `childSessionID` while building (update the existing row on Stop rather than appending a new one), and add a resume test with a Start+Stop pair asserting exactly one card.

## Minor
- Plan-step timestamps are preserved by matching steps on `content`; two steps with identical content would cross-associate timestamps. Low risk, but a fallback to position (or an id) would be sturdier.
- `runChildProcess`: a non-`ExitError` `Wait()` failure returns `exitCode: 0` alongside the error — harmless since the error is surfaced, but slightly inconsistent.
- Worth confirming `Ctrl+P` / `Ctrl+G` don't collide with any existing binding.

The architecture and the message-passing concurrency model are sound; both blocking items are localized fixes.

gnanam1990 (approved):
Re-reviewed after `bc28bdee` — both blocking items from my previous review are resolved, and two latent bugs were fixed along the way. CI is green across all platforms (incl. Windows smoke).

**1. Stream truncation / stall (`specialist/exec.go`) — fixed.**
Swapped `bufio.Scanner` (1 MiB line cap, silent drop) for `bufio.Reader.ReadString('\n')`: no per-line cap (so a large stream-json event no longer truncates the `Task` result), the reader keeps draining (so the child can't block on a full stdout pipe), and parse/read errors now surface as an `EventError` instead of being silently dropped. Good.

**2. Duplicate cards on resume (`tui/session.go`) — fixed.**
`EventSpecialistStop` now reconciles by `childSessionID` (updates the existing Start row in place rather than appending a second one), so a completed specialist renders as a single card on resume.

**Bonus fixes (both correct, verified against the writer):**
- Payload key `"specialistName"` → `"specialist"` — matches what `baseSpecialistPayload` actually writes (`payload["specialist"]`), so specialist names now render on resume instead of being blank.
- Added a `"success"` status case — `specialistStopStatus` returns `"success"`, so this maps it to `specialistCompleted` correctly.

**Optional (non-blocking, fine to defer):**
- Plan-step timestamps still match steps by `content`; two identical-content steps could cross-associate timestamps. Low risk.
- Worth a quick check that `Ctrl+P` / `Ctrl+G` don't collide with existing bindings.

Nice work addressing the feedback. LGTM.

PR #256, opened 2026-06-19
token efficiency: lossless cost accounting + minified file reads (read_minified_file)

gnanam1990:
Two complementary improvements to how Zero handles tokens — one makes cost *measurement* correct, the other *reduces* tokens on file-heavy work.

## Part 1 — Accounting: lossless, correct cost
Zero captured cache-read + reasoning tokens, then dropped them before persisting — so `zero usage report` over-priced cache-heavy sessions, and Anthropic's cache-creation tokens were parsed then discarded.

- **One lossless `Usage`** carrying all five token dimensions (input, cache-read, **cache-write**, output, reasoning) threaded `TokenUsage → NormalizeUsage → Usage → CalculateCost`.
- **Cache-write pricing** at the premium rate (Anthropic ~1.25× input, derived per model in the `anthropicModel` helper). **Regression-safe:** a model with no cache-write rate bills those tokens at the input rate exactly as before.
- **Lossless persistence:** `EventUsage` now stores cache-read/cache-write/reasoning (omitempty), so `BuildReport` reconstructs the **same** cost as the live tracker. A single `EventUsagePayload` writer is paired with the `usageEventPayload` reader so the JSON keys can't drift. Older events decode to zero and price exactly as before.
- **Live context-window gauge** in the status bar — `◔ 128K/200K · 64%`, graded green → amber(≥75%) → red(≥90%) — from the last request's input tokens vs the model window.

## Part 2 — Reduction: `read_minified_file`
A new read-only tool returning a **dense, comment-free, line-number-free** view of a file, so the model can scan code for far fewer tokens than `read_file`.

- **Go** → minified through the real `go/parser` AST (guaranteed-valid, comment-free).
- **C / Java / CSS / SCSS / LESS / Python** → comment-stripped via a **string-aware lexer** that tracks strings, char literals, triple-quotes, Java text blocks, and Python f-strings (incl. 3.12 same-quote nesting).
- **Every other language** (JS/TS/C++/C#/Rust/Swift/Kotlin/shell/…) → safe **whitespace-only** — their raw-strings/regex/lifetimes/heredocs make naive comment-stripping unsafe. The stripper only handles languages whose string forms it models exactly, so **it cannot corrupt content**.
- Measured **~32% fewer tokens** on a realistic 8-file "understand a subsystem" read. Costs +~102 fixed tokens (the tool schema), recovered many times over by a single subsystem read.

## Testing
- Accounting: cache-write premium + fallback pricing, Anthropic rate derivation, **lossless persistence round-trip** (report cost == live cost), omit-zero payload, context gauge.
- Minification: Go strips-comments-keeps-valid-code + fallback; per-language golden cases — comment chars inside strings, `//`/`/*` in strings, **Java text blocks**, **Python f-string same-quote nesting**, **CSS has no `//`**, SCSS interpolation/hex; and the unsupported-language whitespace-only fallback.
- `gofmt` / `go vet` / `go build ./...` / full suites incl. `-race` all green.

No provider/protocol changes beyond surfacing the already-parsed Anthropic cache-creation count.

## Summary by CodeRabbit

* **New Features**
  * Added a minified-file read capability that can strip comments (multi-language) and remove line numbers for token-efficient scanning.
  * Enhanced usage accounting to track cache-write tokens and incorporate them into cost breakdowns.
  * Added a context utilization gauge showing prompt usage vs model capacity (with color thresholds).

* **Bug Fixes**
  * Improved persistence and reconstruction of usage details so cost calculations reflect cached reads and cache writes accurately.

* **Tests**
  * Expanded coverage for comment stripping, minified-file reading, and cache-write token/cost reconstruction.
  * Added UI test coverage for the context gauge.

### DIFF
--- a/internal/minify/comments.go
+++ b/internal/minify/comments.go
@@ -1,0 +1,189 @@
+package minify
+
+import "strings"
+
+// commentStyle describes one language's comment + string lexical rules for the
+// string-aware stripper. ONLY languages whose every string form this scanner can
+// model exactly appear in commentStyles. Languages with raw-string delimiters
+// (C++ R"(...)", C# @"..."), lifetimes (Rust 'a), template literals (JS/TS
+// `${...}`), nested block comments (Rust/Swift), or heredocs (shell/Ruby) are
+// deliberately omitted — they fall back to whitespace-only minification — so the
+// stripper can never mistake code or string content for a comment, nor a comment
+// for code. Correctness over coverage.
+type commentStyle struct {
+	name       string
+	line       []string // line-comment markers; the rest of the line is dropped
+	blockOpen  string   // "" when the language has no block comments
+	blockClose string
+	triples    []string // raw multi-line string delimiters (Python ''' """, Java text block """)
+	fStrings   bool     // Python f-strings: a quote inside {…} does not end the string
+}
+
+// commentStyles maps a file extension to its stripper config. Every entry here is
+// covered by golden tests in comments_test.go, including the tricky cases
+// (comment chars inside strings, strings inside comments, escapes, triple-quotes,
+// f-string same-quote nesting, Java text blocks).
+var commentStyles = map[string]commentStyle{
+	".c":    {name: "c", line: []string{"//"}, blockOpen: "/*", blockClose: "*/"},
+	".java": {name: "java", line: []string{"//"}, blockOpen: "/*", blockClose: "*/", triples: []string{`"""`}},
+	".css":  {name: "css", blockOpen: "/*", blockClose: "*/"},
+	".scss": {name: "scss", line: []string{"//"}, blockOpen: "/*", blockClose: "*/"},
+	".less": {name: "less", line: []string{"//"}, blockOpen: "/*", blockClose: "*/"},
+	".py":   {name: "python", line: []string{"#"}, triples: []string{`"""`, `'''`}, fStrings: true},
+	".pyi":  {name: "python", line: []string{"#"}, triples: []string{`"""`, `'''`}, fStrings: true},
+}
+
+// stripComments removes line and block comments from src for one language while
+// correctly skipping string, char, triple-quoted, and (Python) f-string literals,
+// so a comment marker inside a literal is never stripped and literal content is
+// never mistaken for a comment. Delimiters are all ASCII, so scanning byte-wise is
+// safe — UTF-8 content is copied verbatim.
+func stripComments(src string, style commentStyle) string {
+	var out strings.Builder
+	out.Grow(len(src))
+	i, n := 0, len(src)
+	for i < n {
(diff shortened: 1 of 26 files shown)


internal/minify/comments.go
```diff
@@ -0,0 +1,189 @@
...
+// \-escaped delimiter does not terminate; an unterminated literal runs to EOF.
+func scanDelimited(src string, bodyStart int, delim string) int {
+	j, n := bodyStart, len(src)
+	for j < n {
+		if src[j] == '\\' {
+			j += 2
+			continue
+		}
+		if strings.HasPrefix(src[j:], delim) {
+			return j + len(delim)
```
Vasanthdev2004, line 108: [P2] Raw triple-quoted strings ending with an odd number of backslashes before the closing delimiter over-extend to EOF

`scanDelimited` treats `\\` as an escape (skip 2 chars), but in a Python raw triple-quoted string (e.g. `r"""...\\\"""`) the backslash is a literal and does NOT escape the following quote. When content ends with an odd number of backslashes immediately before `"""`, the escape-skip jumps over the closing delimiter, so the scanner runs to EOF (or the next `"""`) swallowing all subsequent code as string content. Content is preserved verbatim (no data loss), but the minified view structurally misrepresents code as string literal content and fails to minify the rest of the file — technically violating the PR's stated 'worst case is no reduction, never wrong content' guarantee. This is an uncommon pattern (raw triple-quoted string ending with `\\`), so non-blocking, but worth a follow-up to handle raw-string backslashes literally in `scanDelimited`.

Vasanthdev2004 (approved):
LGTM — the cost accounting is clean and well-tested. CacheWriteTokens is threaded losslessly through TokenUsage -> NormalizeUsage -> Usage -> CalculateCost -> CostBreakdown -> Summary with proper clamping (cache-write + cache-read <= input), fallback pricing (models without a cache-write rate bill at input rate, matching prior behavior), and lossless persistence via EventUsagePayload paired with usageEventPayload so JSON keys can't drift. The round-trip test proves report cost == live cost.

The read_minified_file tool is solid: path traversal protection reuses resolveScopedReadPath (same as read_file), FileTracker records raw content (not minified) so edit_file conflict detection stays correct, and the header explicitly warns to use read_file before editing. The Go AST minification via go/parser is guaranteed-valid; the per-language comment stripper is string-aware and falls back to whitespace-only for unsupported languages so it can never corrupt content.

One non-blocking P2 follow-up noted inline: scanDelimited's backslash-escape handling can over-extend raw triple-quoted strings (e.g. Python r"""...\\\""") when content ends with an odd number of backslashes before the closing delimiter. Content is preserved verbatim (no data loss), but subsequent code gets mislabeled as string content and isn't minified. Uncommon pattern, low impact, but technically violates the PR's stated 'worst case is no reduction, never wrong content' guarantee.

PR #258, opened 2026-06-19
tools: defer swarm tool schemas + keyless SearXNG web_search backend

gnanam1990:
Two small fixes for the raised issues — the idle per-request tool-schema prefix, and native web search having no built-in engine.

## 1. Defer the swarm tool schemas (idle prefix)
The 7 `swarm_*` tools are an advanced, rarely-first-move feature, and the base system prompt doesn't name them — yet their full schemas shipped **eagerly on every request**. They now implement `Deferred()` (via a shared embedded mixin) and load on demand through `tool_search`, like MCP tools.

- `partitionTools` already keys off `tools.IsDeferred` (loop.go), so this reuses the existing, proven deferral path — swarm just joins the deferred set.
- **Core tools and specialist `Task` stay eager** — no `tool_search` round-trip added to common operations; the intentional eager-built-ins design is preserved for the core.
- **Measured:** `prompt_tokens` 11,246 → 10,776 (**−470/request**) in a swarm-enabled (`--auto high`) config, with **no behavioral loss** — swarm still works, just loaded when the model needs orchestration.

## 2. Keyless SearXNG backend for `web_search`
The native `web_search` tool only registers with a configured backend and has no built-in engine. This adds a **`searxng` provider**: `ZERO_WEBSEARCH_PROVIDER=searxng` + a SearXNG URL issues `GET /search?format=json` with **no API key**, parsing `{results:[{title,url,content,score}]}`.

- Self-host or point at any SearXNG instance — native search **without** baking a shared key (abuse risk) or a flaky public default.
- Verified live end-to-end: the model called `web_search` → SearXNG → returned + listed results.

## Tests
- `TestSwarmToolsAreDeferred` — all swarm tools are deferred-eligible.
- `TestSearxngBackendGETParsesResults` — GET, no auth header, field mapping (content→snippet, score), skips empty URLs.
- `TestSearxngBackendRespectsLimit`.
- gofmt / vet / `go build ./...` / full `go test ./...` green (incl. agent + cli deferral-sensitive suites).

## Release Notes

* **New Features**
  * Added support for SearXNG as a web search backend option, including API key authentication

* **Tests**
  * Added test coverage for SearXNG web search backend integration and result parsing
  * Added tests validating swarm tool deferred loading support

### DIFF
--- a/internal/swarm/deferred_test.go
+++ b/internal/swarm/deferred_test.go
@@ -1,0 +1,23 @@
+package swarm
+
+import (
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
+// All swarm tools must be deferred-eligible so their schemas stay out of the
+// eager per-request tool prefix (loaded on demand via tool_search), while the
+// core built-ins remain eager.
+func TestSwarmToolsAreDeferred(t *testing.T) {
+	registry := tools.NewRegistry()
+	RegisterTools(registry, &Swarm{})
+	for _, tool := range registry.All() {
+		if !tools.IsDeferred(tool) {
+			t.Errorf("swarm tool %q is NOT deferred-eligible (would bloat the eager prefix)", tool.Name())
+		}
+	}
+	if len(registry.All()) < 6 {
+		t.Fatalf("expected the swarm tools registered, got %d", len(registry.All()))
+	}
+}
--- a/internal/swarm/schedule_tool.go
+++ b/internal/swarm/schedule_tool.go
@@ -21,6 +21,7 @@ const ScheduleToolName = "swarm_schedule"
 type scheduleTool struct {
 	sw  *Swarm
 	now func() time.Time // injectable for tests; nil => time.Now
+	deferredSwarmTool
 }
 
 func (t *scheduleTool) Name() string { return ScheduleToolName }
--- a/internal/swarm/tools.go
+++ b/internal/swarm/tools.go
@@ -33,6 +33,16 @@ func RegisterTools(registry *tools.Registry, sw *Swarm) {
 	registry.Register(&scheduleTool{sw: sw})
 }
 
+// deferredSwarmTool marks a swarm tool as deferred-eligible. Swarm tools are an
+// advanced, rarely-first-move orchestration feature and the base system prompt
+// does not name them, so they are hidden behind tool_search (loaded on demand)
+// instead of shipping their full schemas in the eager per-request tool prefix.
+// Embedding it in each swarm tool struct is enough — partitionTools keys off the
+// shared tools.IsDeferred check, exactly as it does for MCP tools.
+type deferredSwarmTool struct{}
+
+func (deferredSwarmTool) Deferred() bool { return true }
+
 // policyFrom derives the member-inheritance policy from the live tool options so
 // each spawned member runs on the orchestrator's current model + permission mode.
 func policyFrom(options tools.RunOptions) Policy {
@@ -61,7 +71,10 @@ func errResult(format string, a ...any) tools.Result {
 
 // ---- swarm_spawn -----------------------------------------------------------
 
-type spawnTool struct{ sw *Swarm }
+type spawnTool struct {
+	sw *Swarm
+	deferredSwarmTool
+}
 
 func (t *spawnTool) Name() string { return SpawnToolName }
 func (t *spawnTool) Description() string {
@@ -115,7 +128,10 @@ func (t *spawnTool) RunWithOptions(_ context.Context, args map[string]any, optio
 
 // ---- swarm_send ------------------------------------------------------------
 
-type sendTool struct{ sw *Swarm }
+type sendTool struct {
+	sw *Swarm
+	deferredSwarmTool
+}
 
 func (t *sendTool) Name() string { return SendToolName }
 func (t *sendTool) Description() string {
@@ -170,7 +186,10 @@ func (t *sendTool) RunWithOptions(_ context.Context, args map[string]any, _ tool
 
 // ---- swarm_inbox -----------------------------------------------------------
 
-type inboxTool struct{ sw *Swarm }
+type inboxTool struct {
+	sw *Swarm
+	deferredSwarmTool
+}
 
 func (t *inboxTool) Name() string { return InboxToolName }
 func (t *inboxTool) Description() string {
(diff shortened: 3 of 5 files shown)


Vasanthdev2004 (approved):
﻿## Review — tools: defer swarm tool schemas + keyless SearXNG web_search backend

Verdict: **approve with comments**. The diff is small, well-scoped, idiomatic to the existing deferred-loading mechanism, and tests pass locally (`go build ./internal/swarm/... ./internal/tools/...` and `go test ./internal/swarm/... ./internal/tools/...` both green on the PR branch). No P1 blocking issues. A few P2/P3 notes below.

### Deferred swarm tools

The implementation follows the existing `IsDeferred` pattern correctly. `tools.IsDeferred` (`internal/tools/registry.go:67`) type-checks against the unexported `deferredTool` interface (`Deferred() bool`), and the new `deferredSwarmTool` embedded struct satisfies it via a value-receiver method, exactly like the MCP path (`internal/tools/deferred_test.go:285`). `partitionTools` (`internal/agent/loop.go:1665`) keys off `tools.IsDeferred`, so no loop changes are needed.

Embedding is consistent: all swarm tool structs use pointer receivers (`internal/swarm/tools.go`, `internal/swarm/schedule_tool.go`), and the embedded `deferredSwarmTool` has a value receiver, which Go promotes correctly through the embedded value field. The `web_search` tool itself is not deferred (correct — it's a core network tool that must stay eager).

- **P3 — `internal/swarm/deferred_test.go:26`**: the guard is `len(registry.All()) < 6`, but `RegisterTools` registers **7** tools (`internal/swarm/tools.go:27-33`: spawn, send, inbox, status, handoff, collect, schedule). The threshold passes today, but it silently tolerates a regression that drops a tool. Suggest `!= 7` (or `len(registry.All()) != len(swarmToolNames)`) so a missing registration actually fails the test.
- **P3**: the test only asserts that *every* registered swarm tool is deferred. It does not assert the inverse — that the core/eager tools (e.g. `web_search`) are *not* deferred. A one-line `registry.Register(...)` of a non-swarm tool with an `IsDeferred == false` assertion would catch a future regression where someone widens `deferredSwarmTool` to a shared base. Optional.

### SearXNG backend

`searchSearxng` (`internal/tools/web_search.go:158-192`) is well-formed: GET to `{baseURL}/search`, `format=json`+`q` params, `Accept: application/json`, the same `User-Agent` as the generic backend, optional `Authorization` only when `apiKey != ""`. It reuses `webSearchBodyLimit` (256 KiB) and the same 2xx status check as the POST path. Good.

- **P2 — keyless + empty/`{}` response**: when a SearXNG instance has `formats: []` configured without `json` (the default install), it returns a 200 with an HTML page or an empty body, not a JSON error. `parseSearxngResults` (`internal/tools/web_search.go:194-217`) calls `json.Unmarshal` on the HTML and returns `decode searxng response: ...`, which surfaces to the agent as a generic decode error rather than a useful "enable the `json` format on your SearXNG instance" hint. Consider either sniffing the `Content-Type` for `application/json` and producing a clearer error, or documenting the required SearXNG config (`settings.yml: search.formats: [html, json]`) in a comment or the commit message. Not blocking — the error is still surfaced, just not actionable.
- **P3 — `internal/tools/web_search.go:208-210`**: empty-URL results are skipped (`strings.TrimSpace(r.URL) == ""`), but unlike `parseSearchResults` (`web_search.go:337-339`), there's no `len(trimmed) == 0` guard on the body itself. An empty `[]byte` from the server unmarshals cleanly to a zero-length `Results` and returns `(nil, nil)` rather than an error. The generic path treats an empty body as an error (`empty search backend response`). Worth a 2-line parity check for consistency and so a misconfigured instance returns an error instead of a silent "no results".
- **P3 — `Score float64`**: `parseSearxngResults` uses a plain `float64` for `Score`, whereas `parseSearchResults` deliberately uses `lenientScore` (`web_search.go:309-324`) because some providers stringify/`null` the score and a plain `float64` would reject the *entire* decode. SearXNG reliably emits a numeric `score`, so this is fine in practice — but if a self-hosted instance tweaks its output, a single bad score drops all results. Optional: reuse `lenientScore` for defense-in-depth parity with the generic path.
- **P3 — provider routing**: `searxng` is matched case-insensitively via `strings.EqualFold(backend.provider, "searxng")` in `Search` (`web_search.go:145`), so the operator sets `ZERO_WEBSEARCH_PROVIDER=searxng` + `ZERO_WEBSEARCH_BASE_URL=https://my-searxng.example`. Worth confirming in `defaultSearchBackend` (or a comment) that the same `apiKey`/`baseURL` env vars are reused — they are, since the struct is shared, so no change needed; just a doc-level note.

### Security — keyless SearXNG

No new exposure. The keyless path is fine because:
- `RunWithSandbox` (`internal/tools/web_search.go:108-118`) still calls `enforceScopedNetworkPolicy` against `backend.endpointHost()`, which for SearXNG returns the configured base URL host — so sandbox network policy applies identically to the generic backend.
- The optional `Authorization` header is only added when `apiKey != ""`, supporting reverse-proxy fronted instances. Default (no key) is the documented SearXNG posture.
- `webSearchBodyLimit` bounds the read; status-code check is present; no request body to leak. No SSRF surface added — `baseURL` comes from env, same as before.

No regressions to the existing tool-loading behavior: the deferred mechanism is purely additive (a tool that wasn't deferred before now is), `partitionTools` is untouched, and the inactive path (no `tool_search` or below `DeferThreshold`) is byte-identical to legacy for these tools — they just stop being eager.

### Tests

Both new test files exercise real behavior against `httptest.NewServer`:
- `TestSearxngBackendGETParsesResults` asserts GET method, `/search` path, `format=json&q=language`, the **absence** of an `Authorization` header (good — directly verifies the keyless claim), empty-URL skipping, and field mapping (`Title`/`URL`/`Snippet`/`Score`). Real assertions, not smoke tests.
- `TestSearxngBackendRespectsLimit` verifies `limit` truncation.
- `TestSwarmToolsAreDeferred` verifies every registered swarm tool is deferred-eligible via the real `tools.IsDeferred` check (not a string comparison).

Gaps (all P3):
- No test for the non-2xx HTTP error path (`searxng backend returned HTTP %d`). A `httptest` server returning 500 would cover it.
- No test for the malformed-JSON decode error path.
- No test for the empty-body case (the P3 parity gap above).
- No test that `endpointHost()` returns the SearXNG base URL host (relevant for sandbox enforcement). The generic backend has `TestHTTPSearchBackendEndpointHost`; a SearXNG equivalent would lock in the security-relevant behavior.

### Summary
- P1 (blocking): 0
- P2 (should fix): 1 — clearer error when SearXNG JSON format is disabled
- P3 (nice to have): 5 — test threshold, empty-body parity, lenient score, error-path coverage, endpointHost test

Approving. The P2 is worth a follow-up but doesn't block merge — the error is surfaced, just not as actionable as it could be.

PR #259, opened 2026-06-19
tools: tool_search redirects eager-tool lookups instead of "no tools matched"

gnanam1990:
## Problem
A Windows user hit an agent that got stuck looping (screenshot: `tool_search select:Task` and `select:Task,update_plan` → **"No tools matched. Available: mcp_filesystem_…"**, then halted).

Root cause: `tool_search`'s `resolveExact`/`noMatchMessage` only ever consult the **deferred** tool set. When the model searches for a tool it *already has eagerly* — `Task`, `update_plan`, any core tool — it finds nothing and returns a message that **lists only the deferred MCP tools**, implying the requested tool doesn't exist. A weaker model (cloud deepseek-v4-flash / minimax-m3) reads "not available" and **loops** retrying `tool_search` instead of just calling the tool.

## Fix
`tool_search` now recognizes eager tools (registered, not deferred, passing the operator filters) and steers the model correctly:

| Model does | Before | After |
|---|---|---|
| `select:Task,update_plan` (eager) | "No tools matched. Available: mcp_…" | **"Task, update_plan are already in your tool list — call them directly."** |
| `select:weather,Task` (mixed) | loads weather, ignores Task | loads weather **+ notes** Task is already available |
| `select:nonexistent` | "No tools matched" | "No tools matched" *(unchanged)* |
| keyword `update_plan` (eager) | misleading | redirect to call directly |

Surgical — only the error/redirect path changed; loading deferred tools is byte-identical. `NewToolSearchTool` already holds the full registry; it just wasn't using it to recognize eager tools.

Scope: this fixes the **looping/confusion** (the Zero-side bug). It does **not** address tool-call JSON mangling by weak models (the `@specialist` mention syntax, where Zero builds the JSON, is the mitigation for that), and the agent's "failed 4× → halt" is correct loop-protection.

## Tests
`TestToolSearchRedirectsEagerSelectToCallDirectly`, `…LoadsDeferredAndNotesEager`, `…UnknownStillReportsNoMatch`, `…KeywordEagerRedirects` — plus all existing tool_search tests. gofmt / vet / `go build ./...` / full `go test ./...` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool search results messaging: when searching for tools, the app now clearly indicates which tools are already in your toolkit, preventing confusing "no tools matched" responses when tools are actually available. Enhanced feedback when both newly loaded and existing tools match a query.

### DIFF
--- a/internal/tools/tool_search.go
+++ b/internal/tools/tool_search.go
@@ -80,7 +80,17 @@ func (tool toolSearchTool) RunWithOptions(_ context.Context, args map[string]any
 		matches = tool.rankByKeyword(query, deferred)
 	}
 
+	// Names the model asked for that it ALREADY has eagerly. tool_search only
+	// loads DEFERRED tools, so a request for an eager tool (e.g. Task or
+	// update_plan) otherwise falls through to a misleading "no tools matched" —
+	// a common confusion that leaves weaker models looping. Redirect the model to
+	// call those tools directly instead.
+	alreadyAvailable := tool.eagerToolsForQuery(query, options.EnabledTools, options.DisabledTools)
+
 	if len(matches) == 0 {
+		if len(alreadyAvailable) > 0 {
+			return okResult(alreadyAvailableMessage(alreadyAvailable))
+		}
 		return okResult(tool.noMatchMessage(query, deferred))
 	}
 
@@ -88,9 +98,13 @@ func (tool toolSearchTool) RunWithOptions(_ context.Context, args map[string]any
 	for _, match := range matches {
 		names = append(names, match.Name())
 	}
+	output := renderLoadedTools(matches)
+	if len(alreadyAvailable) > 0 {
+		output += "\n\n" + alreadyAvailableMessage(alreadyAvailable)
+	}
 	return Result{
 		Status: StatusOK,
-		Output: renderLoadedTools(matches),
+		Output: output,
 		Meta:   map[string]string{"load_tools": strings.Join(names, ",")},
 	}
 }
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Review: tool_search redirects eager-tool lookups

**Verdict: APPROVE** — the fix is correct, well-structured, and well-tested. The core bug (eager tools like `Task`/`update_plan` yielding a misleading "No tools matched" and looping weak models) is properly addressed, and the existing loading/no-match behavior is preserved.

### What's good
- The eager/deferred distinction is sound: `visibleEagerToolNames` (`internal/tools/tool_search.go:138`) correctly excludes `IsDeferred` tools, `tool_search` itself, and filter-hidden tools — mirroring `visibleDeferredTools`.
- `eagerToolsForQuery` (`internal/tools/tool_search.go:151`) handles both `select:` (exact) and keyword (substring) paths consistently with the existing `resolveExact`/`rankByKeyword` split.
- `load_tools` meta stays deferred-only — verified by `TestToolSearchRedirectsEagerSelectToCallDirectly`. Eager tools are never "loaded", only noted.
- No regression in the empty/unknown paths: an empty query produces no eager hits (no keywords → no substring matches), so `noMatchMessage` still fires. `TestToolSearchUnknownStillReportsNoMatch` confirms genuinely-unknown names still get the honest "No tools matched".
- Filter handling (`toolAllowedByFilters`) is reused for eager tools, so operator-hidden eager tools won't be incorrectly redirected.

### P3 — nice to have (non-blocking)

1. **Mixed-case message timing contradiction** (`internal/tools/tool_search.go:101-104`): when a query matches both deferred and eager tools, the output says `Loaded N tools... call them on the next turn.` then appends `...call them directly.`. The two refer to different tool sets, but a weak model could conflate "next turn" vs "directly". Consider phrasing the eager note as `...call <names> directly now (they are already loaded).` to disambiguate.

2. **`alreadyAvailableMessage` style** (`internal/tools/tool_search.go:198`): the `map[bool]string{true: "them", false: "it"}[len(names) > 1]` idiom works but is harder to read than a plain `if len(names) > 1`. Minor.

3. **Untested edge cases** (nice-to-have, not required):
   - An eager tool hidden via `DisabledTools` should NOT be redirected (the code handles it via `toolAllowedByFilters`, but no test pins this for the eager path).
   - Keyword query matching both a deferred and an eager tool (mixed keyword case) — only the `select:` mixed case is tested (`TestToolSearchLoadsDeferredAndNotesEager`).
   - Empty/`select:` query with eager tools present — confirms it still falls through to `noMatchMessage` rather than redirecting on nothing.

4. **"deferred-tools notice" phrasing** (`internal/tools/tool_search.go:202`): the clause `(those in the deferred-tools notice)` assumes the model recalls the system-prompt notice. For the weak-model audience this fix targets, the extra clause may be noise; `tool_search only loads DEFERRED tools, not tools you already have.` would be tighter.

### Summary
- P1 (blocking): 0
- P2 (should fix): 0
- P3 (nice to have): 4

Solid fix. Approving.

PR #260, opened 2026-06-19
audit fixes: cron lock, accounting race, O(n²) hot paths, /style wiring

gnanam1990:
Fixes **five verified findings** from a code audit. Each was checked against the codebase first; two related findings turned out already-addressed upstream, and three need larger/separate work (all noted below).

## Fixed

**H-1 — cron store has no cross-process lock** *(the read-modify-write race)*
Two schedulers seeing a job due raced the metadata read-modify-write — `fireJob` even documented it (*"full atomicity needs file locking"*). Adds a per-job **O_EXCL lock** (mirrors `internal/oauth`'s proven pattern, incl. stale-reclaim) and a `Mutate` primitive doing the re-read→modify→write atomically; `Update`/`Remove` lock too. The lock is held only for the millisecond RMW — **never across a job's exec**.

**M-5 — specialist accounting check-then-append race**
`specialistEventExists` (unlocked `ReadEvents`) + a later `AppendEvent` let two finishers (TaskOutput poll vs `onExit`, even cross-process) both pass the check and write **duplicate** stop/usage events. Adds `sessions.Store.AppendEventUnlessExists` (existence check + append atomic under the session lock) and routes both paths through it.

**M-2 — hooks audit append is O(n²)**
`append` re-read and scanned the *whole* audit log every call to find the max sequence. Now reads only the file **tail** — O(1), and **multi-process safe** (the log is a shared global file), unlike an in-memory counter which would collide sequences across processes.

**M-4 — transcript rehydration is O(n²)**
The three resume/rewind/compact loops called `appendTranscriptRow` per event, each re-scanning every existing row for the keyed-row dedup. A bulk `appendTranscriptRowsDedup` builds the key set **once**; semantics are identical (verified by a parity test).

**M-6 — `/style` was inert**
The TUI `/style` command stored a style but never threaded it anywhere. Adds `agent.Options.ResponseStyle`, a system-prompt directive for `concise`/`explanatory`/`review`, and wires `m.responseStyle` through the run. `balanced` is a no-op (prompt byte-identical to before).

## Already addressed upstream (no change needed)
- **M-1** (forked-session double-count): `Store.Fork` already **skips `EventUsage`** events, with a comment naming the exact double-count it prevents.
- **M-3** (lineage O(N²)): `Tree` already does a **single `store.List()` scan** and indexes children in memory; no path calls `List` per node.

## Deferred (need larger/separate changes — not a contained fix)
- **M-7** `escalate_model` in the TUI needs a `ModelSwitcher` that builds a provider mid-run; registering the tool alone yields a tool that errors on call.
- **L-1** MCP SSE reconnect is a stream-lifecycle change (and low-likelihood after the 8 MiB per-event cap).
- **L-2** dead-code removal belongs in a dedicated cleanup sweep.

## Tests
Per-fix unit tests, including **concurrency under `-race`** (`cron.Mutate` atomicity, `sessions.AppendEventUnlessExists` atomicity, multi-store hooks sequence continuity) and a transcript dedup **parity** test. gofmt / vet / `go build ./...` / full `go test ./...` green.

## Summary by CodeRabbit

* **New Features**
  * Added operator-selected agent response styles (**concise**, **explanatory**, **review**) via system prompt directives; **balanced** preserves prior behavior.
  * Added a concurrency-safe “append unless exists” API to prevent duplicate session events.
* **Bug Fixes**
  * Improved cron scheduling persistence and run recording using cross-process per-job locking, including correct handling when jobs are removed.
  * Made audit log sequencing more reliable (faster tail-based updates and cross-process locking).
  * Strengthened transcript de-duplication across resume/rewind/compaction.
  * Improved Windows lock contention handling for OAuth file locks.

### DIFF
--- a/internal/agent/response_style_test.go
+++ b/internal/agent/response_style_test.go
@@ -1,0 +1,30 @@
+package agent
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestResponseStyleContext(t *testing.T) {
+	for _, blank := range []string{"", "balanced", "BALANCED", "garbage", "  "} {
+		if got := responseStyleContext(Options{ResponseStyle: blank}); got != "" {
+			t.Errorf("ResponseStyle %q should add nothing, got %q", blank, got)
+		}
+	}
+	for _, style := range []string{"concise", "explanatory", "review", "Concise", "REVIEW"} {
+		got := responseStyleContext(Options{ResponseStyle: style})
+		if got == "" {
+			t.Fatalf("style %q produced no directive", style)
+		}
+		if !strings.Contains(strings.ToLower(got), strings.ToLower(style)) {
+			t.Errorf("style %q directive should name the style, got %q", style, got)
+		}
+	}
+	// It lands in the assembled prompt for a real style, and not for balanced.
+	if !strings.Contains(buildSystemPrompt(Options{ResponseStyle: "concise"}), "Response style: concise") {
+		t.Error("buildSystemPrompt should inject the concise directive")
+	}
+	if strings.Contains(buildSystemPrompt(Options{ResponseStyle: "balanced"}), "Response style:") {
+		t.Error("balanced must not inject a style section (prompt stays byte-identical)")
+	}
+}
--- a/internal/agent/system_prompt.go
+++ b/internal/agent/system_prompt.go
@@ -86,10 +86,30 @@ func buildSystemPrompt(options Options) string {
 	if delegation := specialistDelegationContext(options); delegation != "" {
 		sections = append(sections, delegation)
 	}
+	if style := responseStyleContext(options); style != "" {
+		sections = append(sections, style)
+	}
 	if policy := strings.TrimSpace(confirmationPolicy); policy != "" {
 		sections = append(sections, policy)
 	}
 	return strings.Join(sections, "\n\n")
+}
+
+// responseStyleContext renders the operator-selected reply style (TUI /style) as
+// a system-prompt directive so the choice actually shapes responses. "balanced"
+// (the default) and unknown/empty values add nothing, keeping the prompt
+// byte-identical to the pre-style behavior.
+func responseStyleContext(options Options) string {
+	switch strings.ToLower(strings.TrimSpace(options.ResponseStyle)) {
+	case "concise":
+		return "Response style: concise. Lead with the result and keep answers short and direct; omit preamble, restating the question, and any explanation not needed to be correct."
+	case "explanatory":
+		return "Response style: explanatory. Explain the reasoning behind your answers and changes, surface relevant trade-offs, and add brief context a learner would want — while staying on task and not padding."
+	case "review":
+		return "Response style: review. Work like a critical reviewer: call out risks, edge cases, and unstated assumptions, flag anything questionable, and prefer precise, evidence-backed statements over reassurance."
+	default:
+		return ""
+	}
 }
 
 func approvedCommandPrefixContext(options Options) string {
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -184,6 +184,11 @@ type Options struct {
 	ReasoningEffort  string
 	Cwd              string
 	SystemPrompt     string
+	// ResponseStyle is the operator-selected reply style from the TUI /style
+	// command (e.g. "concise", "explanatory", "review"). It is rendered into the
+	// system prompt as a short directive. Empty or "balanced" adds nothing — the
+	// prompt is then byte-identical to the pre-style behavior.
+	ResponseStyle string
 	// Images are optional image attachments to seed onto the initial user turn.
 	// nil for text-only runs (the seeded message then carries no images, exactly
 	// as before).
--- a/internal/cli/cron_run.go
+++ b/internal/cli/cron_run.go
@@ -202,8 +202,9 @@ func fireJob(store *cron.Store, now func() time.Time, job cron.Job, stdout io.Wr
 		fmt.Fprintf(stdout, "fired %s -> exit %d (job removed during run)\n", job.ID, code)
 		return
 	}
-	// Record the fire only after confirming the job was not removed mid-run, so
-	// AppendRun cannot resurrect a deleted job's directory.
+	// Record the fire. AppendRun takes the per-job lock and bails if the job's
+	// metadata is gone, so a Remove landing here cannot resurrect a deleted job's
+	// directory with an orphaned runs.jsonl.
 	if aerr := store.AppendRun(job.ID, rec); aerr != nil {
 		fmt.Fprintf(stderr, "warning: failed to record run for %s: %v\n", job.ID, aerr)
 	}
--- a/internal/cli/cron_run.go
+++ b/internal/cli/cron_run.go
@@ -177,31 +177,41 @@ func fireJob(store *cron.Store, now func() time.Time, job cron.Job, stdout io.Wr
 	} else {
 		job.NextRunAt = nxt
 	}
-	// Re-read before persisting: the job may have been paused or removed while it
-	// was executing, and this in-memory copy is stale from tick start. Without
-	// this, store.Update would clobber an external pause back to active. (A single
-	// scheduler is the supported model; this narrows but does not fully close the
-	// read-modify-write window — full atomicity needs file locking.)
-	current, err := store.Get(job.ID)
-	switch {
-	case errors.Is(err, cron.ErrJobNotFound):
-		// Genuinely removed mid-run: don't recreate it by persisting.
+	// Re-read and persist the advanced state ATOMICALLY under the per-job lock. The
+	// job may have been paused or removed while it executed, and this in-memory copy
+	// is stale from tick start. Mutate closes the read-modify-write window a single
+	// scheduler alone could only narrow (Store.Mutate took a cross-process lock): a
+	// concurrent scheduler or an external pause/remove landing here can no longer be
+	// clobbered.
+	persisted, err := store.Mutate(job.ID, func(current cron.Job, readErr error) (cron.Job, error) {
+		if readErr != nil {
+			// A transient read failure (IO/permission) is NOT removal — persist the
+			// computed next state anyway so the schedule advances and the job does not
+			// re-fire next tick.
+			fmt.Fprintf(stderr, "warning: could not re-read job %s before persist: %v\n", job.ID, readErr)
+			return job, nil
+		}
+		if current.Status == cron.StatusPaused {
+			// Honor an external pause that landed while the job ran.
+			job.Status = cron.StatusPaused
+		}
+		return job, nil
+	})
+	if errors.Is(err, cron.ErrJobNotFound) {
+		// Genuinely removed mid-run: don't recreate it (no run record, no persist).
 		fmt.Fprintf(stdout, "fired %s -> exit %d (job removed during run)\n", job.ID, code)
 		return
-	case err != nil:
-		// A transient read failure (IO/permission) is NOT removal — warn but still
-		// record the run and persist the computed next state below.
-		fmt.Fprintf(stderr, "warning: could not re-read job %s before persist: %v\n", job.ID, err)
-	case current.Status == cron.StatusPaused:
-		job.Status = cron.StatusPaused
 	}
-	if err := store.AppendRun(job.ID, rec); err != nil {
-		fmt.Fprintf(stderr, "warning: failed to record run for %s: %v\n", job.ID, err)
+	// Record the fire only after confirming the job was not removed mid-run, so
+	// AppendRun cannot resurrect a deleted job's directory.
+	if aerr := store.AppendRun(job.ID, rec); aerr != nil {
+		fmt.Fprintf(stderr, "warning: failed to record run for %s: %v\n", job.ID, aerr)
 	}
-	if err := store.Update(job); err != nil {
+	if err != nil {
 		fmt.Fprintf(stderr, "warning: failed to persist job state for %s: %v\n", job.ID, err)
+		persisted = job
 	}
-	fmt.Fprintf(stdout, "fired %s -> exit %d (next: %s)\n", job.ID, code, formatCronTime(job.NextRunAt))
+	fmt.Fprintf(stdout, "fired %s -> exit %d (next: %s)\n", job.ID, code, formatCronTime(persisted.NextRunAt))
 }
 
 func cronTruncate(s string, max int) string {
--- a/internal/cron/append_run_test.go
+++ b/internal/cron/append_run_test.go
@@ -1,0 +1,34 @@
+package cron
+
+import (
+	"os"
+	"testing"
+)
+
+// AppendRun on a job removed mid-run must NOT resurrect its directory (which the
+// old unconditional MkdirAll did, leaving an orphaned runs.jsonl with no
+// metadata.json).
+func TestAppendRunDoesNotResurrectRemovedJob(t *testing.T) {
+	store := NewStore(StoreOptions{RootDir: t.TempDir()})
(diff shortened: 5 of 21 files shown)


Vasanthdev2004 (changes requested):
## Review: audit fixes — cron lock, accounting race, O(n²) hot paths, /style wiring

I read the diff end-to-end, built the PR branch (`go build ./...` clean), and ran every new test (all PASS). Verdict below is **Request changes** because of one cross-process correctness gap in the hooks audit fix; the rest of the PR is solid.

---

### P1 — blocking

**1. `AuditStore.append` is NOT cross-process atomic; the tail read alone doesn't deliver the claimed guarantee.**
`internal/hooks/hooks.go:500` (`append`) holds only the in-process `store.mu`. `lastSequence()` (hooks.go:528) re-reads the file tail fresh each call, which is enough to *observe* another process's prior writes, but the **read-then-write is not atomic across processes**. Two processes can both `lastSequence() → 50`, both assign `Sequence = 51`, and both `O_APPEND`-write — producing two records with the same sequence number. The PR's own comment ("a concurrent process's appends … are still observed — unlike an in-memory counter, which would collide sequences across processes") oversells what the tail read buys: it eliminates the *stale-counter* collision but not the *check-then-append* collision. The audit log is a shared global file, so this matters precisely in the multi-process scenario the PR targets.

The new test `TestAuditStoreSequenceTailRead` (audit_sequence_test.go) only exercises **sequential** two-store usage (store #1 writes 50, store #2 writes #51). It does not race two stores appending concurrently, so it cannot catch this. A `-race`-style concurrency test (two stores, N goroutines each, assert all `Sequence` values are unique) would fail under the current implementation.

Fix direction: take a cross-process file lock around the `lastSequence()` + append write (the same `O_EXCL`-with-stale-reclaim pattern already used in `internal/cron/lock.go` and `internal/oauth/lock.go`), or use an OS atomic append-with-sequence primitive. The in-process `store.mu` can stay as the fast path. This is the same class of race the PR correctly closes for cron (`Store.Mutate`) and specialist accounting (`AppendEventUnlessExists`) — the audit path is the odd one out.

---

### P2 — should fix

**2. `fireJob` → `AppendRun` can resurrect a deleted job directory.**
`internal/cli/cron_run.go:191` calls `store.AppendRun(job.ID, rec)` *after* `store.Mutate` has returned and released the per-job lock. A `Remove` landing in that window deletes the job dir; `AppendRun` then `os.MkdirAll`s it back (cron/store.go `AppendRun`), leaving an orphaned `runs.jsonl` with no `metadata.json`. `List` won't list it (it skips dirs without `metadata.json`), but `Runs` will still return the orphaned records. Narrow, but the comment "Record the fire only after confirming the job was not removed mid-run, so AppendRun cannot resurrect a deleted job's directory" claims the window is closed when it isn't. Either move `AppendRun` inside `Mutate`'s lock window, or have `AppendRun` stat `metadata.json` and bail if absent.

**3. `lastSequence` `ReadAt` error path can spuriously fail on a short final read.**
`internal/hooks/hooks.go:498`: `if n, err := file.ReadAt(buf, start); err != nil && n < len(buf) { return 0, err }`. `ReadAt` is permitted to return `io.EOF` with a short read when fewer than `len(buf)` bytes are available at the offset (e.g. the file was truncated between `Stat` and `ReadAt`). On a 0-padded tail read this would surface as `io.EOF` and abort the append. In practice the file is only ever appended so this is unlikely, but `io.EOF` should be treated as a non-error when `n > 0` (the standard `Read`-style handling), matching how `lastEventSequence` in `internal/sessions/store.go` handles its tail loop.

---

### P3 — nice to have / observations

**4. `oauth.acquireFileLock` stale-reclaim compares a file to itself.**
`internal/oauth/lock.go:62`: `stale, _ := os.ReadFile(lockPath); if data, rerr := os.ReadFile(lockPath); rerr == nil && string(data) == string(stale) { _ = os.Remove(lockPath) }`. This reads the file twice and compares it to itself — `string(data) == string(stale)` is always true when both reads succeed, so the guard is a no-op. Pre-existing (only the Windows `ErrPermission` line was added in this PR), but worth fixing while the lock code is being touched: the intent was presumably to re-read after `Remove` races and bail if the token changed.

**5. `accountingMu` + `AppendEventUnlessExists` double-locking is now redundant for in-process callers.**
`internal/specialist/accounting.go:42` still takes `accountingMu.Lock()` before calling `appendSpecialistEventOnce`, which itself takes the session lock. `accountingMu` is now only a cheap fast-path that doesn't provide the real guarantee (the comment says so). That's fine, but the in-process mutex is now strictly redundant — `AppendEventUnlessExists`'s session lock already serializes all in-process callers. Removing `accountingMu` would simplify; keeping it is harmless. Just flagging in case it was meant to be removed.

**6. `appendSpecialistEventOnce` drops the "skip append when parent invalid" silently vs. the old path.**
`internal/specialist/accounting.go` (new `appendSpecialistEventOnce`): when `parentSessionID` is empty/invalid it returns `(false, nil)` — "not appended, no error." The old `appendSpecialistSessionEvent` returned `(Event{}, nil)` in the same case. Both result in no append, so behavior is preserved. The `_, _ =` discard in `recordSpecialistStop` means the difference is invisible. Just noting the parity holds.

**7. `TestMutateIsAtomicUnderConcurrency` is a good test but only exercises in-process concurrency.**
`internal/cron/mutate_test.go:264` runs 20 goroutines in one process. The whole point of the O_EXCL lock is cross-process safety, which a goroutine test cannot verify. Consider a two-process test (spawn a second `go test` binary / `os/exec` helper) that both `Mutate` the same job against a shared `RootDir`. Not blocking — the in-process test still validates the lock is taken — but the comment "the lock is not serializing the read-modify-write" would be more honestly demonstrated cross-process.

---

### What looks correct

- **Cron `lockJob`** (`internal/cron/lock.go`): O_EXCL + ownership-token release + stale reclaim is the right pattern, and matches the established `oauth.acquireFileLock` shape. The lock is held only across the RMW in `Mutate`/`Update`/`Remove`, never across job exec — good. The Windows `ErrPermission` contention handling (commit `a090df0`) is correct and necessary on Windows where `os.Remove` leaves a "delete pending" file.
- **`Store.Mutate`** (cron/store.go:343): correctly returns `ErrJobNotFound` on a removed job (no resurrection), surfaces transient read errors via `readErr` so the fire path can still advance the schedule, and honors an external `StatusPaused` landed mid-run. The `fireJob` restructure around it (cron_run.go:177) is correct.
- **`AppendEventUnlessExists`** (`internal/sessions/store.go:528`): the check-then-append is correctly done under `lockSession` (in-process mutex + cross-process file lock), and `ReadEvents` does not re-acquire the session lock (verified — it only does `os.ReadFile`), so there's no deadlock. The 24-goroutine test passes, confirming in-process atomicity.
- **`specialistEventMatcher`** preserves the exact dedup semantics of the old `specialistEventExists` (catch-all runId matching, `childSessionId`/`taskId` aliasing, `source == "specialist"` filter). The parity holds.
- **`appendTranscriptRowsDedup`** (`internal/tui/transcript.go:105`) is a faithful bulk equivalent of the per-row `appendTranscriptRow`: unkeyed rows always append, keyed rows dedup against existing rows *and* earlier new rows. `TestAppendTranscriptRowsDedupMatchesPerRow` asserts byte-identical output against the per-row path across a mixed keyed/unkeyed/duplicate input — adequate parity coverage. O(n²)→O(n) confirmed.
- **`/style` wiring**: `agent.Options.ResponseStyle` is populated from `m.responseStyle` (model.go:2863), consumed by `responseStyleContext` (system_prompt.go), and `balanced`/empty/garbage all return `""` so the prompt is byte-identical to pre-style behavior (asserted by the test). `concise`/`explanatory`/`review` inject a short directive. Verified the directive lands in `buildSystemPrompt`. This is a real, correct fix for the inert `/style` command.

---

### Test note (unrelated to this PR)

`internal/tui`'s `TestModelCommandPersistsSelectedModelToUserConfig` fails on the **base commit** (`673fce1`) with `anthropic provider requires official baseURL https://api.anthropic.com` — an environment/config issue, not introduced here. Confirming so the failure isn't attributed to this PR.

---

### Verdict

**Request changes** — the cron lock, accounting atomicity, transcript dedup, and `/style` wiring are all correct and well-tested. The hooks audit O(1) tail read is a nice performance win *within* a single process, but the PR's stated multi-process safety claim for the audit log is not actually delivered (P1 #1). I'd like to see either (a) a cross-process lock around `AuditStore.append`, or (b) the doc/comment claims walked back to "single-process atomic, multi-process best-effort" with the test renamed to not imply cross-process safety.

anandh8x (dismissed):
Requesting changes. I found one blocking correctness gap and one smaller invariant issue in the new primitives.

High: `AuditStore.append` is still not cross-process atomic.

`internal/hooks/hooks.go:499-518` holds only the per-process `store.mu`, then does `lastSequence()` followed by an `O_APPEND` write. The tail read at `internal/hooks/hooks.go:531-538` makes the sequence lookup O(1), but it does not make the read-then-write atomic across processes. Two Zero processes can both read the same last sequence, assign the same next `Sequence`, and append two audit rows with duplicate sequence numbers. That is the exact multi-process case this PR says it fixes for the global audit log. Please put a cross-process file lock around `lastSequence` plus the append write, or explicitly downgrade the guarantee and tests to single-process/best-effort behavior.

Medium: `Store.Mutate` locks one cron job ID but writes whatever ID the callback returns.

`internal/cron/store.go:192-210` validates and locks `id`, but then passes the callback result directly to `writeJob`; `writeJob` derives the target directory from `job.ID` at `internal/cron/store.go:134-145`. A future caller can accidentally return a different or invalid ID, bypassing the lock that was acquired and potentially writing outside the intended job directory. This new atomic primitive should enforce the invariant, e.g. set `next.ID = id` before writing or reject `next.ID != id` / invalid IDs.

Verification: `git diff --check b64a9d3de97da7a50ab618816b04452ecb7345c8...a090df06a9e70132bb872b597bc8298bd6e912b9` passed. From temporary archives of the PR head, `GOCACHE=/tmp/zero-go-cache go test ./internal/cron ./internal/cli ./internal/hooks ./internal/oauth ./internal/sessions ./internal/specialist ./internal/tui ./internal/agent` passed outside the sandbox, `GOCACHE=/tmp/zero-go-cache go build -buildvcs=false ./...` passed, and `GOCACHE=/tmp/zero-go-cache go test -race ./internal/cron ./internal/sessions ./internal/hooks` passed.

gnanam1990:
Thanks for the thorough review — all blocking + should-fix items addressed in `8a5880f`.

**P1 (audit append not cross-process atomic) — fixed.** Wrapped `lastSequence()` + the append in a per-file O_EXCL lock (`internal/hooks/lock.go`, same stale-reclaim + Windows `ErrPermission` handling as cron/oauth); `store.mu` stays as the in-process fast path. Added `TestAuditStoreSequenceConcurrent` — 4 stores (separate `store.mu`) × 12 goroutines on one file, asserting unique *dense* sequences — which the sequential test couldn't catch. Walked back the over-claiming comment.

**P2 #2 (AppendRun resurrect) — fixed.** `AppendRun` now takes the per-job lock and bails if `metadata.json` is gone, so a `Remove` in the post-`Mutate` window can't recreate the dir. Added `TestAppendRunDoesNotResurrectRemovedJob`; comment corrected.

**P2 #3 (ReadAt short read) — fixed.** `lastSequence` treats `io.EOF` as non-fatal and uses `buf[:n]`, matching `sessions.lastEventSequence`.

**P3 #4 (oauth stale-reclaim no-op) — fixed.** Dropped the double `os.ReadFile` "stale == data" self-comparison; staleness is decided by the mtime check.

**P3 #5 / #7 — acknowledged, left as-is.** `accountingMu` is now a redundant-but-harmless in-process fast path; a true cross-*process* test (spawned helper) is a reasonable follow-up beyond the in-process concurrency test added here. Also noted in the commit: a shared `internal/filelock` helper would DRY the three near-identical locks — deliberately left as a follow-up to avoid churning the already-green cron/oauth packages in this PR.

All tests under `-race`; cron/hooks/oauth green.

Vasanthdev2004 (approved):
## Re-review: all previously flagged issues fixed

Re-reviewed head `8a5880f` against the prior REQUEST_CHANGES. Every P1/P2 from the previous round is resolved, and the previously-correct parts (cron lock, AppendEventUnlessExists, transcript dedup, /style wiring) show no regression.

### P1 — AuditStore.append cross-process sequence race: FIXED

`internal/hooks/hooks.go:504` — `append()` now wraps the read-then-append in `store.lockAudit()` (`internal/hooks/lock.go`), an O_EXCL sibling `.lock` file with token-based release and stale-mtime reclaim, mirroring the cron/oauth locks. `lastSequence()` is called inside the lock, so two processes can no longer both read sequence N and write N+1. The overclaiming comment is gone — the new comment accurately states `store.mu` only serializes this process and `lockAudit` handles cross-process. `TestAuditStoreSequenceConcurrent` (4 separate `*AuditStore` handles sharing one file) exercises exactly the scenario the lock was added for.

### P2 — fireJob→AppendRun deleted-job-dir resurrection: FIXED

`internal/cron/store.go:266` — `AppendRun` now acquires `lockJob` and stats `metadata.json` before `MkdirAll`; a missing dir returns nil (no-op, no resurrection). `fireJob` (`internal/cli/cron_run.go:191`) now uses `store.Mutate`, which returns `ErrJobNotFound` for a removed job and returns early before `AppendRun` is reached. Even if a `Remove` lands in the window between `Mutate` releasing the lock and `AppendRun` acquiring it, `AppendRun`'s own metadata check bails. `TestAppendRunDoesNotResurrectRemovedJob` and `TestMutateRemovedJobReturnsNotFound` cover both paths.

### P2 — lastSequence io.EOF on short tail read: FIXED

`internal/hooks/hooks.go:649` — `ReadAt`'s error is now checked with `!errors.Is(rerr, io.EOF)`, so a short read (EOF) uses the bytes actually read (`buf = buf[:n]`) instead of failing. The `io` import is added. The full-scan fallback for records longer than the 8 KB tail window is preserved.

### P3 — oauth stale-reclaim self-comparison: FIXED

`internal/oauth/lock.go:60` — the double `ReadFile` that compared the file to itself is removed; stale reclaim now keys off the mtime check alone, with a comment explaining the old no-op guard.

### P3 — accountingMu redundancy / in-process-only test: ADDRESSED

`accountingMu` is retained but the comment (`internal/specialist/accounting.go:14`) now correctly describes it as a process-local fast path, with the real dedup guarantee coming from `AppendEventUnlessExists` under the session lock (in-process mutex + cross-process file lock). The new `TestAuditStoreSequenceConcurrent` uses 4 independent `*AuditStore` handles (separate `store.mu`) as a cross-process stand-in — the only way `store.mu` wouldn't serialize them is via `lockAudit`, which is the point.

### Previously correct — no regression

- **Cron per-job lock**: now consistently applied across `Update`, `Mutate`, `Remove`, and `AppendRun` (`internal/cron/store.go`). `Mutate` provides the atomic read-modify-write `fireJob` needed.
- **AppendEventUnlessExists** (`internal/sessions/store.go:531`): atomic check+append under `lockSession`; the ReadEvents-doesn't-relock note prevents a future deadlock trap.
- **Transcript dedup** (`internal/tui/transcript.go:103`): O(n) bulk dedup, byte-identical to the per-row form, tested.
- **/style wiring**: `responseStyleContext` (`internal/agent/system_prompt.go`) injected into `buildSystemPrompt`; "balanced"/unknown stays byte-identical, tested.

### New issues introduced: none

### Minor nits (non-blocking, P3)

- Three near-identical lock implementations (`internal/cron/lock.go`, `internal/hooks/lock.go`, `internal/oauth/lock.go`) — acknowledged in the hooks lock comment as a follow-up for a shared filelock helper.

**Verdict: APPROVE.** All blocking issues are resolved with appropriate tests and accurate comments.

PR #262, opened 2026-06-20
TUI: live code-streaming preview, theme system, plan/Task transcript dedup

gnanam1990:
## Summary

Three TUI UX improvements plus robustness fixes — no new flags.

### 1. Live code-streaming preview
A long `write_file` / `edit_file` / `apply_patch` used to show only a spinner while the file was generated, looking frozen. Tool-call argument deltas are now forwarded `zeroruntime → agent → TUI` (`OnToolCallStart` / `OnToolCallDelta`), and the TUI renders a live `✎ write_file <path> · N lines` block with a color-coded tail of the code as it streams in.

### 2. Theme system
Auto dark/light detection via a terminal-background probe, a `/theme [auto|dark|light]` command, a refreshed palette, and streaming-fade auto-disable on low-color / SSH / tmux terminals.

### 3. Plan / Task transcript dedup
The plan could render up to three times (a full `update_plan` card per call **plus** the live plan panel). `update_plan` now collapses to a one-line summary — the panel owns the detail — and `Task` delegations render only as the specialist card, not also as raw tool-call/result rows, in **both** the live and resume paths.

### Robustness (streaming preview)
- tolerate whitespace in tool-call JSON (`"path": "…"`)
- extract `edit_file`'s `new_string` (the preview was blank for edits)
- clear the preview on tool finalize / cancel / run-end — no stale or duplicate previews, and a stale run can't wipe the active one
- new-file content renders green; a leading `-` (e.g. CSS `-webkit-…`) is no longer mis-colored as a removal
- truncate by display width (no CJK/wide-glyph overflow)

## Tests
New: `streaming_tool_call_test.go`, `theme_select_test.go`, `plan_summary_test.go`, `stage10_test.go`, `wrap_whitespace_test.go`; updated `session_test.go`, `streaming_fade_test.go`.

## Summary by CodeRabbit

* **New Features**
  * Live streaming preview of tool call arguments (including real-time file-writing/code output)
  * Theme selection with auto/dark/light modes and terminal background detection

* **Improvements**
  * `/theme` command now accepts a mode and updates the UI immediately
  * Collapsed plan summaries into step-count status lines
  * Improved text wrapping for aligned/preformatted content and capped assistant measure for readability
  * Updated streaming fade behavior based on environment and theme; refreshes render caching on theme changes

* **Bug Fixes**
  * Reduced duplicate “Task” transcript rows when specialist rendering is present

* **Tests**
  * Added/expanded unit tests for theme resolution, streaming decoding/rendering, wrapping, fading, and plan summaries

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -182,9 +182,11 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		}
 
 		collected := zeroruntime.CollectStreamWithOptions(ctx, stream, zeroruntime.CollectOptions{
-			OnText:      options.OnText,
-			OnReasoning: options.OnReasoning,
-			OnUsage:     options.OnUsage,
+			OnText:          options.OnText,
+			OnReasoning:     options.OnReasoning,
+			OnUsage:         options.OnUsage,
+			OnToolCallStart: options.OnToolCallStart,
+			OnToolCallDelta: options.OnToolCallDelta,
 		})
 		if collected.Error != "" {
 			// REACTIVE compaction: the streamed error may also be a context
@@ -487,9 +489,11 @@ func finalAnswerAfterMaxTurns(ctx context.Context, provider Provider, messages [
 		return "", messages, ""
 	}
 	collected := zeroruntime.CollectStreamWithOptions(ctx, stream, zeroruntime.CollectOptions{
-		OnText:      options.OnText,
-		OnReasoning: options.OnReasoning,
-		OnUsage:     options.OnUsage,
+		OnText:          options.OnText,
+		OnReasoning:     options.OnReasoning,
+		OnUsage:         options.OnUsage,
+		OnToolCallStart: options.OnToolCallStart,
+		OnToolCallDelta: options.OnToolCallDelta,
 	})
 	if ctx.Err() != nil || collected.Error != "" || strings.TrimSpace(collected.Text) == "" {
 		return "", messages, ""
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -213,12 +213,19 @@ type Options struct {
 	// Hooks, when set, runs configured beforeTool (blocking) and afterTool
 	// (advisory) commands around each tool call. nil disables hooks entirely; a
 	// dispatcher built from an empty config is also a safe no-op.
-	Hooks               *hooks.Dispatcher
-	EnabledTools        []string
-	DisabledTools       []string
-	OnText              func(string)
-	OnReasoning         func(string)
-	OnToolCall          func(ToolCall)
+	Hooks         *hooks.Dispatcher
+	EnabledTools  []string
+	DisabledTools []string
+	OnText        func(string)
+	OnReasoning   func(string)
+	OnToolCall    func(ToolCall)
+	// OnToolCallStart / OnToolCallDelta stream a tool call's arguments LIVE as the
+	// model generates them — OnToolCallStart on open (id, tool name), then
+	// OnToolCallDelta for each argument fragment. A surface can render the
+	// in-progress call (e.g. a file being written) instead of waiting for
+	// OnToolCall, which only fires once the whole call has accumulated. nil no-ops.
+	OnToolCallStart     func(id, name string)
+	OnToolCallDelta     func(id, fragment string)
 	OnPermissionRequest func(context.Context, PermissionRequest) (PermissionDecision, error)
 	OnPermission        func(PermissionEvent)
 	OnAskUser           func(context.Context, AskUserRequest) (AskUserResponse, error)
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -258,9 +258,9 @@ var commandDefinitions = []commandDefinition{
 	},
 	{
 		name:        "/theme",
-		usage:       "/theme",
+		usage:       "/theme [auto|dark|light]",
 		group:       commandGroupSession,
-		description: "Show theme state.",
+		description: "Switch or show the color theme (auto-detects terminal background).",
 		kind:        commandTheme,
 	},
 	{
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -207,9 +207,10 @@ type model struct {
 	// In-progress tool call whose arguments are streaming (a file being written),
 	// shown live by streamingToolCallView so a long write/edit isn't a frozen
 	// spinner. Cleared when the call completes (next text/turn) — see updateModel.
-	streamCallID   string
-	streamCallName string
-	streamCallArgs string
+	// streamCallDecoder decodes the streamed args incrementally (O(1) per delta).
+	streamCallID      string
+	streamCallName    string
+	streamCallDecoder *streamingDecoder
 
 	// Slash-command autocomplete (purely additive UI state). suggestions is the
 	// live match list for the current "/token"; suggestionIdx is the highlighted
@@ -1129,13 +1130,13 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 		// A new tool call opened — reset the live "writing" block to it.
 		m.streamCallID = msg.id
 		m.streamCallName = msg.name
-		m.streamCallArgs = ""
+		m.streamCallDecoder = newStreamingDecoder()
 		return m, nil
 	case toolCallStreamDeltaMsg:
-		if msg.runID != m.activeRunID || msg.id != m.streamCallID {
+		if msg.runID != m.activeRunID || msg.id != m.streamCallID || m.streamCallDecoder == nil {
 			return m, nil
 		}
-		m.streamCallArgs += msg.fragment
+		m.streamCallDecoder.feed(msg.fragment)
 		return m, nil
 	case agentTextMsg:
 		if msg.runID != m.activeRunID {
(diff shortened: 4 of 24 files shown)


Vasanthdev2004 (approved):
## PR #262 review — TUI live code-streaming preview, theme system, plan/Task transcript dedup

CI is green (Smoke on ubuntu/macos/windows, Performance, Security, Zero Review, CodeRabbit all SUCCESS). The change is well-scoped, well-commented, and well-tested — new tests cover theme resolution, palette contrast/hierarchy, the streaming JSON decoder's tolerance (kimi-style whitespace, dangling escapes), the fade-disable matrix, transcript dedup on resume, and the new wrapping behavior. Overall this is high-quality work.

A few findings, categorized.

### P1 — unbounded memory growth on a streamed tool call

`internal/tui/model.go` (`toolCallStreamDeltaMsg` case) appends every argument fragment to `m.streamCallArgs` for the lifetime of the call:

```go
m.streamCallArgs += msg.fragment
```

`streamingToolCallView` only ever shows the last `streamingTailLines` (6) lines, and decodes the buffer on every render, so the full buffer is *retained* but never fully displayed. For a large `write_file` (multi-MB generated files are common) the TUI holds the entire streamed JSON args string in memory for the whole call and re-parses it from the start on every render tick (~150ms during fade, every keystroke-driven render otherwise). Two concrete problems:

1. Memory: a 5–10 MB file write keeps a 5–10 MB string live plus the `strings.Builder` decode cost on each frame.
2. CPU: `decodeStreamingJSONString` does a fresh `strings.Index` + linear scan over the whole buffer each render; that's O(n) per frame on a buffer that grows to O(n), so the render path is O(n²) over the stream.

`clearStreamingToolCall` releases it only when the call finalizes/cancels — i.e. after the whole file has already been buffered. Suggest capping `streamCallArgs` (e.g. keep a rolling tail window, or stop appending once it exceeds a few hundred KB and switch the preview to "showing tail …"). The preview already only needs the tail, so truncating the head is safe.

### P2 — `spec_mode.go` seeds `fadeActive = true` but never sets `fadeDisabled`

`internal/tui/spec_mode.go:210` force-sets `m.fadeActive = true` for the spec-impl run, but the new `m.fadeDisabled` flag (set in `newModel` via `defaultFadeDisabled`) is a per-model field that gates whether the tick is scheduled. After this PR the `agentTextMsg` branch only schedules the fade tick when `!m.fadeDisabled`, but spec_mode still flips `fadeActive` on directly. On a remote/SSH session where `fadeDisabled == true`, the spec-impl path leaves `fadeActive == true` with no tick ever scheduled — so `styleStreamingLine` will route through the fade palette (the `!m.fadeActive` early-out at `streaming_fade.go:241` is bypassed) while `lineAges` is nil, hitting the second early-out (`m.lineAges == nil`) and rendering at base ink. Net effect is *probably* fine visually, but the invariant "fadeActive ⇒ a tick is running" is now broken, and `resetStreamingFade` is the only thing that clears it. Either have spec_mode respect `fadeDisabled` (don't set `fadeActive` when disabled, mirroring the `agentTextMsg` branch), or assert the invariant explicitly. Not a blocker since output is still readable, but it's a latent inconsistency the next refactor will trip on.

### P2 — `decodeStreamingJSONString` matches the *first* occurrence of `"key"` anywhere

`internal/tui/streaming_tool_call.go`:

```go
keyMarker := `"` + key + `"`
idx := strings.Index(args, keyMarker)
```

`strings.Index` finds the marker as a substring of any value, not just a JSON key. For `streamingFileContent`, the key list includes `"new"` and `"input"` — a `write_file` whose `path` happens to be `…/new/…` will have `"path":"…/new/…"` matched by the `"new"` search *before* the real `"content"` field if `content` arrives later than `path`. More realistically, `"input"` will match a path containing the segment `input`. The result is the preview shows a fragment of the path as the "content" tail until the real content field streams in, then snaps to it.

It's a preview-only cosmetic issue, but worth a tighter match: require the marker to be preceded by `{` or `,` (optionally with whitespace), which is the JSON-object key position. Cheap to add and removes the false-positive class.

### P2 — transcript dedup skips the Task *result* row, losing the failure output when a specialist starts then errors

`internal/tui/model.go:3433` (live path) and `internal/tui/session.go` (resume path) both skip the `tool result: Task` row when a specialist card exists for that `toolCallId`. The resume-path test (`resume_task_test.go`) correctly keeps the result row when *no* `EventSpecialistStart` exists. But the live path dedups purely on `result.Name == "Task"` — if a specialist *did* start (so a card exists) and then the Task tool returns `status=error` with a useful `Output` (e.g. "spawn failed: …"), that error text now lives only on the specialist card's completion state, not on a transcript row.

For a specialist that started-but-failed, the card's error rendering may or may not surface the same `result.Output`. Worth confirming that `specialistCompleteMsg` → the card renderer shows `result.Output` verbatim; if it only shows `errorMsg` from the child session (which is a different field), the original tool-result error is silently dropped from the transcript. The live-path guard should probably be `if result.Name == "Task" && specialistStarted && status == OK` (keep the error result row so failures stay visible even when a card exists).

### P2 — `/theme` live switch does not repaint already-rendered scrollback, and the message says so — but inline-mode users will perceive a bug

`handleThemeCommand` prints "Already-printed scrollback keeps its previous colors; new output uses the new theme." That's honest, but in inline (non-alt-screen) mode the *current visible frame* is re-rendered from `m.transcript` on every View call, and `transcript_view.go` stamps `renderer.activeRunID = 0` so cached renders are reused. `applyTheme` does `defaultRenderCache.clear()`, so the next render *will* repaint — good. But rows whose render result is *not* routed through `defaultRenderCache` (any renderer that builds its string inline) will keep the old palette until the next state change forces a re-render. Quick sanity check: after `/theme light`, force a full re-render (e.g. send a `tea.WindowSizeMsg` or trigger a re-flow) so the visible frame is consistent. Low impact, but the "new output uses the new theme" promise is only true for cache-routed renderers today.

### P3 — dead constants `colorPanel2` / `colorPanel3` retained

`internal/tui/theme.go:85-86` keeps `colorPanel2 = "#121215"` and `colorPanel3 = "#17171b"` but no `palette` field or renderer references them (the new `palette` struct dropped them, and a grep confirms zero non-definition usages). They were unused before this PR too, but since the file is being rewritten to the `palette` struct this would have been the natural moment to delete them. Trivial cleanup.

### P3 — `colorPanel` / `colorInk` / `colorAccent` etc. are still referenced outside `theme.go`

The PR's stated contract is "no hex literal may appear outside this file" — and the new `palette` struct enforces that for hex. But the *old* `colorXxx` consts are still defined in `theme.go` and still referenced outside it:

- `internal/tui/rendering.go:1003` — `zeroTheme.userPrompt.Background(lipgloss.Color(colorPanel))` (this is the line the PR *partially* rewrites to `zeroTheme.onPanel(...)` for the `❯ ` segment, but `rendering.go:1003` in the current tree still uses `colorPanel` — the diff shows the change to `onPanel`, so confirm the post-PR tree no longer has this reference; if the diff's `onPanel` replacement is the only change in that line, the remaining `fill(zeroTheme.ink)`/`fill(zeroTheme.accent)` segments are fine).
- `internal/tui/streaming_fade.go:64-65` — `lipgloss.Color(colorAccent)` / `colorInk` in `init()`. The PR *does* replace these with `zeroTheme.accentColor`/`inkColor` in `rebuildStreamingFadePalette`, so post-PR this is resolved — good.
- `internal/tui/streaming_fade_test.go` — multiple `lipgloss.Color(colorAccent)` / `colorInk` references. The PR rewrites these to `darkPalette.accent` / `darkPalette.ink`. Good, but it means the old `colorXxx` consts are now only referenced by tests *and* by the yet-unmigrated `rendering.go` line (if any remain). If the post-PR tree still has any `colorXxx` reference outside `theme.go`, the file's own contract is violated; if not, the `colorXxx` block can be deleted entirely (it's superseded by `darkPalette`). Recommend either deleting the `colorXxx` block in this PR or filing a follow-up; leaving both tables in one file is a maintenance footgun.

### P3 — `assistantMeasureCap = 96` is hardcoded

`internal/tui/rendering.go`: the cap is a magic constant with a clear comment, which is fine, but it's not adjustable by the user (no `ZERO_MEASURE_CAP` / theme knob) even though the rest of the TUI exposes `ZERO_THEME` / `ZERO_NO_FADE`. Not a blocker — just noting that users on ultrawide setups who *want* full-width prose have no escape hatch.

### P3 — `styleStreamingCodeLine` for `apply_patch` checks `+++ `/`--- ` prefixes but `apply_patch` hunk headers use those for file headers

The `+++`/`---` guard is correct for unified-diff file headers, but `apply_patch` arg content streamed mid-flight may not yet have reached the file header — a streamed fragment starting with `+++` mid-buffer (e.g. inside a hunk) would be colored as ink (default branch), which is the same as the non-diff treatment. Acceptable for a preview; flagging only because the diff-coloring logic is heuristic and a user might see a flash of miscolored `+`/`-` lines during the stream. No action needed.

---

### Verdict

Approve with P2 follow-ups. The P1 memory/CPU growth on large streamed writes is the one I'd want addressed before merge (it's a real regression on the exact workload the feature targets — large `write_file`/`apply_patch`); the P2s are correctness/consistency issues that can land as a fast-follow. Tests and CI are solid, the theme system is cleanly factored (`palette` → `buildTheme` → `zeroTheme` + `applyTheme` invalidates the cache and rebuilds the fade ramp), and the transcript dedup correctly preserves failed-Task rows on resume. Nice work.

PR #264, opened 2026-06-20
fix(oauth): make ChatGPT (Codex) login work end-to-end

Vasanthdev2004:
-

## Summary by CodeRabbit

* **New Features**
  * Enhanced ChatGPT authentication with improved callback port stability and additional connector scopes for better integration capabilities.
  * Extended OAuth loopback listener configuration to support custom ports and flexible redirect path handling.

* **Bug Fixes**
  * Improved ChatGPT account identification by supporting nested claim structure parsing with fallback compatibility.

### DIFF
--- a/internal/oauth/loopback.go
+++ b/internal/oauth/loopback.go
@@ -31,13 +31,17 @@ type callbackResult struct {
 // begins serving. state is the CSRF value the callback must echo back. Call
 // RedirectURI to build the redirect_uri, then Wait for the code. Always Close.
 func NewLoopbackListener(state string) (*LoopbackListener, error) {
-	// Refuse an empty state: the callback check compares the query's state to this
-	// value, so an empty state would match a callback carrying no state at all,
-	// defeating CSRF protection (fail closed).
+	return NewLoopbackListenerOnPort(state, 0)
+}
+
+// NewLoopbackListenerOnPort is like NewLoopbackListener but binds a specific
+// port (0 = OS-assigned). Used by ChatGPT OAuth which requires a fixed
+// redirect_uri of http://localhost:1455/auth/callback.
+func NewLoopbackListenerOnPort(state string, port int) (*LoopbackListener, error) {
 	if strings.TrimSpace(state) == "" {
 		return nil, errors.New("oauth: loopback listener requires a non-empty CSRF state")
 	}
-	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	ln, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
 	if err != nil {
 		return nil, fmt.Errorf("oauth: start loopback redirect listener: %w", err)
 	}
--- a/internal/oauth/loopback.go
+++ b/internal/oauth/loopback.go
@@ -56,8 +56,24 @@ func (l *LoopbackListener) RedirectURI() string {
 	return fmt.Sprintf("http://%s/callback", l.listener.Addr().String())
 }
 
+// RedirectURIWithHost returns a redirect URI using the given host (e.g.
+// "localhost") and path (e.g. "/auth/callback"). The listener still binds
+// 127.0.0.1, but the OAuth client may require "localhost" in the redirect_uri
+// (OpenAI's ChatGPT client registration does). The listener accepts both
+// /callback and the given path.
+func (l *LoopbackListener) RedirectURIWithHost(host, path string) string {
+	_, port, _ := net.SplitHostPort(l.listener.Addr().String())
+	return fmt.Sprintf("http://%s:%s%s", host, port, path)
+}
+
+// Port returns the OS-assigned TCP port the listener is bound on.
+func (l *LoopbackListener) Port() string {
+	_, port, _ := net.SplitHostPort(l.listener.Addr().String())
+	return port
+}
+
 func (l *LoopbackListener) handle(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Path != "/callback" {
+	if r.URL.Path != "/callback" && r.URL.Path != "/auth/callback" {
 		http.NotFound(w, r)
 		return
 	}
--- a/internal/oauth/loopback.go
+++ b/internal/oauth/loopback.go
@@ -70,12 +70,6 @@ func (l *LoopbackListener) RedirectURIWithHost(host, path string) string {
 	return fmt.Sprintf("http://%s:%s%s", host, port, path)
 }
 
-// Port returns the OS-assigned TCP port the listener is bound on.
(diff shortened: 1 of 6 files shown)


anandh8x (dismissed):
Requesting changes on #264 at 98d8f26. The final PR head no longer contains the advertised api.connectors scope fix. Commit 5af5dd2 added api.connectors.read and api.connectors.invoke, but the later fixed-port commit reverted internal/oauth/presets.go back to only openid, profile, email, and offline_access. As a result, authorize URLs built from the ChatGPT preset still omit the connector scopes the PR says are required, so the original authorize_hydra_invalid_request failure is not actually fixed. Please restore the connector scopes in the ChatGPT preset and add a regression test that asserts the generated ChatGPT authorize URL includes api.connectors.read and api.connectors.invoke. Verification: diff-check passed, and GOCACHE=/tmp/zero-go-cache go test ./internal/oauth ./internal/provideroauth passed outside the sandbox; the local sandbox cannot run these loopback tests because it blocks localhost listeners. The failing Windows smoke check is the existing internal/tools interrupt test, unrelated to these OAuth files.

Vasanthdev2004:
Ran an adversarial correctness audit of this flow against the codex CLI (which ships the same OAuth client `app_EMoam…` against auth.openai.com, so it's ground truth). It surfaced one **P0** that survived adversarial refutation, now fixed in `357ac71`:

**`extractChatGPTAccountID` read `chatgpt_account_id` from the top level of the id_token, but OpenAI nests it under `https://api.openai.com/auth`.** Against a real token the top-level lookup returned empty → `token.Account` empty → the codex provider omitted the `chatgpt-account-id` header on every call → backend/Cloudflare 401. Login *succeeded* (bearer minted) but every subsequent Codex call silently failed. The unit tests passed only because the fixture put the claim at the top level — which masked the bug. codex's own parser (`token_data.rs`) and its test fixtures read/build it exclusively under the nested namespace, confirming the real token shape.

Fix reads the nested path first (matching codex), keeps a top-level fallback for forward-compat, and the tests now use the real nested shape + regressions (top-level decoy must not shadow nested; fallback still works when the namespace is absent).

The audit also flagged a **P3** (no 1455→1457 port fallback like codex has — robustness only, doesn't block normal login) which I left as a follow-up. Everything else — scopes, PKCE, state/CSRF, authorize params, token exchange, refresh — verified correct against codex. The missing second `obtain_api_key` token-exchange is **intentionally** absent and correct (it's for a different `AuthMode::ApiKey` against api.openai.com, not the ChatGPT subscription backend we target).

gnanam1990 (changes requested):
## Review: ChatGPT OAuth fixes — REQUEST_CHANGES

The OAuth work itself is correct and well-tested, and the earlier "scopes reverted" review is now resolved (commit `862c9e4b` restores them with a regression test):

- `presets.go` adds `api.connectors.read` / `api.connectors.invoke`, and `TestChatGPTAuthorizeURLIncludesConnectorScopes` asserts they reach the authorize URL. ✅
- Nested `chatgpt_account_id` extraction under the `https://api.openai.com/auth` namespace with a top-level fallback — a real fix (reading the top level returned empty against real tokens → dropped the `chatgpt-account-id` header → 401 on every Codex call), and it matches the Codex CLI parser. ✅
- Fixed port 1455 + `localhost` + `/auth/callback` + the extra authorize params (`id_token_add_organizations`, `codex_cli_simplified_flow`, `originator`) match the public Codex client registration and address `authorize_hydra_invalid_request`. The loopback changes keep the empty-state CSRF guard and add tests. ✅

### Blockers

1. **This PR bundles #273's de-flake commit (`3df9d6832`), which carries a test-quality flaw.** `internal/tools/exec_command_test.go` is unrelated to ChatGPT OAuth, and it's the same change as open PR #273 (currently CHANGES_REQUESTED). The interrupt test calls `session.terminate()` directly **before** `write_stdin "\x03"`, so the session is already reaped when the path under test runs — the assertion no longer guards write_stdin's Ctrl-C branch (`exec_command.go:663`). Please drop this commit from #264 (rebase once #273 lands) so the PR is scoped to the OAuth fix; if it must stay, apply the same correction requested on #273.

2. **Title/description undersell the change.** "add missing api.connectors scopes" is now a small part of a broader ChatGPT-OAuth fix (redirect URI host/port, callback path, nested account-id claim, authorize params). Please update the title/body so the changelog and reviewers reflect the actual scope.

Once the unrelated test commit is removed (or corrected), the OAuth changes look good to merge.

Vasanthdev2004:
Addressed both blockers:

1. **Dropped the unrelated de-flake commit.** Rebased onto `main` and removed `3df9d68` (the `exec_command_test.go` change) — the PR now touches **zero files in `internal/tools`**, scoped purely to the OAuth fix. The de-flake (with the coverage fix you both flagged) lives in #273. The branch is now `MERGEABLE` again.
2. **Updated the title + description** to reflect the real scope — it's now "make ChatGPT (Codex) login work end-to-end" covering scopes + nested account-id claim + redirect URI/port/path + authorize params, not just the scopes line.

`go build`/`vet`/`gofmt` clean; oauth + provideroauth suites green. Ready for re-review.

gnanam1990 (approved):
## Overview
Fixes the ChatGPT (Codex) OAuth login end-to-end: pins the loopback redirect to `localhost:1455/auth/callback`, requests the required `api.connectors.read`/`api.connectors.invoke` scopes plus the Codex-CLI authorize params, and reads `chatgpt_account_id` from the nested `https://api.openai.com/auth` claim object instead of the (always-empty) top level.

## Strengths
- **Core fix correct:** `extractChatGPTAccountID` reads the namespaced claim first and only falls back to a bare top-level claim for forward-compat, so the real nested value can't be shadowed (the top-level read returned empty against real tokens → dropped the `chatgpt-account-id` header → 401 on every Codex call).
- **Security preserved:** the empty-state CSRF guard is retained in `NewLoopbackListenerOnPort`; the listener still binds `127.0.0.1` even though the redirect_uri advertises `localhost`; and the extra authorize params flow through `BuildAuthorizationURL`'s reserved-key filter, so they can't override `redirect_uri`/`state`/PKCE.
- Fixed-port bind failure returns an actionable error pointing at a conflicting login.
- **Tests genuinely guard each change** (fail-without-fix): nested-claim extraction, the new `/auth/callback` path, and the connector scopes in the authorize URL.

## Earlier blockers — both resolved
- The unrelated `exec_command_test.go` de-flake (#273's commit) is **no longer bundled** — the PR is now scoped to the OAuth files.
- Title/scope updated to reflect the full end-to-end fix.

## Verdict
**APPROVE** — correct, tightly scoped, well-tested.

PR #265, opened 2026-06-20
Fix five concurrency/leak bugs (swarm, specialist, oauth, mcp, cron)

gnanam1990:
## Summary

Five concurrency / resource-leak bugs surfaced by a full-tree audit. All are latent (the suite passes) and live in race windows or cancel paths the tests don't exercise. Each ships a regression test where the behavior is deterministically testable.

## Fixes

| Area | Bug | Fix |
|---|---|---|
| **swarm** | A handoff registers the new task, then delivers the mailbox note; if the Send fails it returns an error having **left a phantom pending task** in the coordinator (inflates counts, shows stuck, becomes an orphan-adoption target). | Deliver the note **before** registering — a failure returns having registered nothing. |
| **specialist** | A foreground sub-agent child is run with plain `CommandContext`, so cancel/timeout SIGKILLs only the immediate child — **grandchildren** (builds/servers/bash it spawned) orphan and can hold the stdout pipe open. | Own process group + group-kill `Cancel` + `WaitDelay` backstop (mirrors the bash tool). |
| **oauth** | Concurrent on-demand refreshes for one key each load the same token and call refresh with the **same single-use refresh token** — all but one fail at the token endpoint. | Per-key mutex around load→refresh→save; re-load inside the lock and reuse a token a concurrent caller already rotated. |
| **mcp** | Concurrent legacy-token migration: one process renames the legacy file, the other's `os.Rename` fails (source gone) → the whole migration is treated as failed → an OAuth MCP server is **wrongly dropped** at startup (though the token migrated fine). | Treat a "source already gone" rename as success. |
| **cron** | `fireJob` execs **before** advancing the schedule, and the fire decision isn't locked — two schedulers (or `--once` overlapping the forever loop) both fire the same due job. | Atomically **claim** the job (advance `NextRunAt` under the per-job lock) before exec; a loser sees the advanced time and skips. The post-exec advance recomputes the same value, so no persisted state changes. |

## Tests
- `TestHardenSpecialistChild` (process group + group-kill Cancel + WaitDelay)
- `TestRefreshSingleflightsConcurrentCallers` (`-race`: exactly 1 refresh across 10 concurrent callers)
- `TestFireJobClaimPreventsDoubleFire` (`-race`: exactly 1 fire across 2 concurrent schedulers)

(swarm/mcp fixes are structural — a statement reorder and an `ErrNotExist`→success — verified by build + the existing suites; their exact races aren't deterministically unit-testable.)

`go build ./...`, `go vet ./...`, `gofmt`, and the full `go test ./...` are green.

## Release Notes

* **Bug Fixes**
  * Prevented cron jobs from firing multiple times concurrently
  * Fixed OAuth token refresh to handle concurrent requests without wasting refresh tokens
  * Improved process cleanup for child processes to prevent orphaned descendants
  * Enhanced handoff reliability by ensuring delivery before task registration

* **Tests**
  * Added concurrency tests for cron job claiming and OAuth token refresh

### DIFF
--- a/internal/cli/cron_claim_test.go
+++ b/internal/cli/cron_claim_test.go
@@ -1,0 +1,41 @@
+package cli
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/Gitlawb/zero/internal/cron"
+)
+
+func TestFireJobClaimPreventsDoubleFire(t *testing.T) {
+	now := time.Date(2026, 6, 9, 9, 0, 0, 0, time.UTC)
+	nowFn := func() time.Time { return now }
+	store := cron.NewStore(cron.StoreOptions{RootDir: t.TempDir(), Now: nowFn})
+	due, _ := store.Add(cron.Job{Expr: "0 9 * * *", Prompt: "fire me", Status: cron.StatusActive, NextRunAt: now.Add(-time.Minute)})
+
+	fx := &fakeExec{}
+	// Two schedulers fire the same due job concurrently; the atomic claim must let
+	// exactly one through, so the job never double-fires (M9).
+	var wg sync.WaitGroup
+	for i := 0; i < 2; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			job, _ := store.Get(due.ID)
+			var out, errb bytes.Buffer
+			fireJob(store, nowFn, job, &out, &errb, fx.run)
+		}()
+	}
+	wg.Wait()
+
+	if len(fx.calls) != 1 {
+		t.Fatalf("claim must let exactly one scheduler fire the due job, got %d", len(fx.calls))
+	}
+	// And NextRunAt is advanced to the next slot (tomorrow 09:00), not left due.
+	d, _ := store.Get(due.ID)
+	if !d.NextRunAt.After(now) {
+		t.Fatalf("NextRunAt must advance past now, got %s", d.NextRunAt)
+	}
+}
--- a/internal/cli/cron_run.go
+++ b/internal/cli/cron_run.go
@@ -134,6 +134,24 @@ func reconcileOverdue(store *cron.Store, now func() time.Time, ids []string, std
 // previous fire has already returned before the next tick — no overlap.
 func fireJob(store *cron.Store, now func() time.Time, job cron.Job, stdout io.Writer, stderr io.Writer, exec execRunner) {
 	fired := now()
+	// Atomically claim this fire before running it so concurrent schedulers (or
+	// --once overlapping the forever loop) can't both execute the same due job: the
+	// winner advances NextRunAt under the per-job lock, and a loser sees the
+	// advanced time and skips. The post-exec advance recomputes the same
+	// NextRunAt (sched.Next(fired) is deterministic), so the claim serializes the
+	// fire decision without changing any persisted schedule state (M9).
+	claimed, claimErr := claimFire(store, fired, job.ID)
+	if claimErr != nil {
+		if errors.Is(claimErr, cron.ErrJobNotFound) {
+			return // removed before we could claim
+		}
+		fmt.Fprintf(stderr, "warning: could not claim job %s: %v\n", job.ID, claimErr)
+		return
+	}
+	if !claimed {
+		return // another scheduler already claimed this slot, or the job was paused
+	}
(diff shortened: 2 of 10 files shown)


Vasanthdev2004 (approved):
## Review — Fix five concurrency/leak bugs

Verified locally: `go build ./...`, `go vet` on all touched packages, and `go test` for `oauth`, `mcp`, `swarm`, `cron`, `cli` are green. `TestRefreshSingleflightsConcurrentCallers` and `TestFireJobClaimPreventsDoubleFire` pass; `TestHardenSpecialistChild` is `//go:build !windows` so it correctly skips on this host. The race detector couldn't run here (no cgo), but the test logic asserts the invariants the fixes target.

### Verdict: Approve

All five fixes are correct in direction, narrowly scoped, and faithful to the existing concurrency model (per-job `Mutate`, per-key locks, build-tag split for proc-group handling). Each ships a regression test where the race is deterministically reproducible. The structural fixes (swarm reorder, mcp `ErrNotExist`) are sound.

### P1 — none

### P2 — none

### P3 (non-blocking observations)

1. **cron — invalid/exhausted schedules still double-fire under concurrency.** `claimFire` only advances `NextRunAt` when `cron.Parse` succeeds and `sched.Next(fired)` is non-zero. For an invalid or exhausted one-shot, both schedulers pass the claim (both see `NextRunAt` unchanged, both get `claimed=true`) and both exec; the post-exec `Mutate` then pauses on both calls. The comment explicitly preserves single-scheduler "fire-once then pause" behavior, so this is accepted — but under `--once` overlapping the loop an invalid job can fire twice before being paused. Worth a one-line note in the comment; no code change required for this PR.

2. **oauth — re-load check is keyed on `AccessToken` only.** `refreshAndSave` short-circuits when `reloaded.AccessToken != current.AccessToken`. A provider that rotates the refresh token but returns an identical access token (extremely rare; most issue a new access token on every refresh) would still call `Refresh` with the already-consumed `current.RefreshToken` and fail at the endpoint. The fix closes the overwhelming majority of the window; this residual is acceptable. If you want to be fully airtight, compare `RefreshToken` too (or the whole token).

3. **swarm — reordered handoff trades a phantom task for a potential orphan mailbox.** Sending the note before `Register` means a `Register` failure leaves an inbox file for a `newID` that was never created. This is strictly less harmful than a phantom pending task (no inflated counts, no orphan-adoption target) and the inbox is keyed by recipient id with no coordinator coupling, so it's the right tradeoff. Flagging only for awareness.

4. **oauth — `refreshLocks` map is append-only.** One `*sync.Mutex` per key ever refreshed. Keys are provider-scoped (bounded set), so this is not a leak in practice; noting only because a long-lived `Manager` refreshing many distinct keys would accumulate entries. No action needed.

### Per-area notes

- **swarm (lifecycle.go):** reorder is correct; the early `return` on `mbErr` now leaves nothing registered. `s.nextID` is called before the mailbox send, so `newID` is consumed on failure — acceptable (ids aren't sequential-dependent here).
- **specialist (exec_proc_*.go):** Unix path correctly uses `Setpgid` + negative-pid `SIGKILL` in `Cancel` + `WaitDelay` backstop, mirroring the bash tool. Windows path is honest about the absence of proc groups and falls back to default `Cancel` + `WaitDelay`. `hardenSpecialistChild` is invoked before `StdoutPipe`/`Start`, as the comment requires.
- **oauth (manager.go):** per-key mutex via a guarded map is the standard singleflight-ish pattern; the in-lock re-load is the key insight that avoids burning the single-use refresh token. Clean.
- **mcp (oauth_store.go):** `errors.Is(err, os.ErrNotExist)` is the right check (covers `fs.ErrNotExist` aliasing). The earlier `Save` calls are not race-protected, but the "never overwrite existing unified entry" guard at line 235 makes a concurrent double-migrate idempotent for the data; only the rename needed the fix.
- **cron (cron_run.go):** `claimFire` + the existing post-exec `Mutate` compose correctly: the claim advances `NextRunAt` under the lock, the post-exec recomputes the same `sched.Next(fired)` (deterministic) and persists run/pause state. The `ErrJobNotFound` short-circuit on removal is handled at both sites.

Nice PR — clear commit shape, honest about which races are unit-testable, and the tests assert the right invariants (`tokenHits == 1`, `len(fx.calls) == 1`, `NextRunAt` advanced).

PR #266, opened 2026-06-20
Fix model-discovery fallback + redact opaque Authorization tokens

gnanam1990:
## Summary

Two robustness/security bugs surfaced by a full-tree audit, each with a regression test.

## Fixes

| Area | Bug | Fix |
|---|---|---|
| **model discovery** | When a live `/models` probe returns 200 but its model ids don't match the curated catalog, `mergeLiveModels` is empty and `DiscoverCatalog` returned that empty list — the picker/wizard **lost the whole live+curated list** (context windows, tool-call flags) and fell back to the bare built-in set with a misleading "no model ids" error. | When the merge is empty but the catalog is non-empty, fall back to the curated catalog. |
| **redaction** (security) | `Authorization` / `Proxy-Authorization` headers with an **opaque token** or a **custom/non-standard scheme** weren't matched by the scheme-aware pattern, so `RedactString` echoed live credentials back into model context and rendered output (logs, command/diff output, errors, session search). | The scheme is now optional — any credential value is redacted to end-of-line; a recognized scheme name (`Bearer`, `Basic`, …) is still kept for readability. |

## Tests
- `TestDiscoverCatalogFallsBackWhenLiveIDsMismatch`
- `TestRedactStringRedactsOpaqueAuthorization`

`go build ./...`, `go vet ./...`, `gofmt`, and the full `go test ./...` are green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model discovery to fall back to a curated catalog when live model data is unavailable or mismatched, preventing empty results.
  * Enhanced authorization header redaction to properly handle both standard and custom authentication schemes.

* **Tests**
  * Added test coverage for catalog discovery fallback behavior.
  * Added test coverage for redacting various authorization header formats.

### DIFF
--- a/internal/providermodeldiscovery/discover_fallback_test.go
+++ b/internal/providermodeldiscovery/discover_fallback_test.go
@@ -1,0 +1,40 @@
+package providermodeldiscovery
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/providercatalog"
+)
+
+func TestDiscoverCatalogFallsBackWhenLiveIDsMismatch(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/api.json":
+			_, _ = w.Write([]byte(`{"openai":{"models":{"gpt-4.1":{"id":"gpt-4.1","name":"GPT-4.1","tool_call":true,"reasoning":true,"limit":{"context":1048576}}}}}`))
+		case "/v1/models":
+			// 200, but live model IDs don't match any curated coding model, so the
+			// merge is empty (M11): the result must fall back to the curated catalog,
+			// not collapse to an empty list.
+			_, _ = w.Write([]byte(`{"data":[{"id":"gpt-image-1"},{"id":"text-embedding-3-large"}]}`))
+		default:
+			t.Fatalf("unexpected request path %q", r.URL.Path)
+		}
+	}))
+	defer server.Close()
+
+	provider := providercatalog.Descriptor{ID: "openai", Transport: providercatalog.TransportOpenAI, DefaultBaseURL: server.URL + "/v1", RequiresAuth: true}
+	models, err := DiscoverCatalog(context.Background(), provider, config.ProviderProfile{
+		CatalogID: "openai", ProviderKind: config.ProviderKindOpenAI, BaseURL: server.URL + "/v1", APIKey: "sk-live",
+	}, Options{HTTPClient: server.Client(), ModelsDevURL: server.URL + "/api.json"})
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (approved):
﻿## Review — PR #266: Model-discovery fallback + opaque Authorization redaction

Both changes are correct, well-targeted, and tested. Approving with minor notes.

### Model-discovery fallback (M11) — `internal/providermodeldiscovery/discovery.go`

The fix is right. Previously a 200 `/models` whose live IDs didn't match any curated coding model produced an empty `mergeLiveModels` result that was returned directly (`[]Model{}, nil`), collapsing the picker to the bare built-in set and surfacing a misleading "no model ids" error. Now `len(merged) > 0` gates the return; an empty merge falls through to the curated-catalog block. The `else if len(catalogModels) == 0` correctly preserves the existing error-path (`liveErr != nil && catalog empty → return liveErr`).

The test (`discover_fallback_test.go`) is well-constructed: live returns `gpt-image-1` + `text-embedding-3-large` (non-coding), catalog returns `gpt-4.1`; merge is empty → fallback to `gpt-4.1` asserted. Traced `mergeLiveModels`: with `hasCatalog=true`, both live IDs miss `byID` and are dropped, yielding empty — confirms the fallback path.

**P3 — behavioral change worth a caller audit:** when *both* the live merge and the catalog are empty, callers previously got `([], nil)`; now they get `(nil, error)` (either `catalogErr` or `"no provider models discovered"`). This is a better contract (empty list + nil was misleading), but it's a return-shape change — worth confirming no caller depends on the empty-slice-no-error case. Not blocking.

### Token redaction (M12) — `internal/redaction/redaction.go`

The regex change is correct: scheme is now optional via `(?:(bearer|basic|...)\s+)?`, capture group 2 stays as the scheme, and the `ReplaceAllStringFunc` preserves the scheme name iff `groups[2] != ""`. Opaque and custom-scheme credentials get their whole value redacted. RE2 leftmost-longest ensures `proxy-authorization` matches as a whole (not just the `authorization` tail after `Proxy-`).

Test coverage is good: bearer (known scheme, name preserved), opaque (no scheme), `Custom-Scheme`, and `Proxy-Authorization`. Verified `RedactString` traces for all four cases — no secret leaks.

**P3 — missing test for parameterized known scheme:** the new `ReplaceAllStringFunc` path isn't exercised with a multi-param known-scheme value (e.g. `Authorization: Digest username="u", response="deadbeef"`). The `[^\r\n]+` capture redacts the whole tail and `groups[2] != ""` keeps `Digest`, but a regression test would guard the branch. Pre-existing behavior, just now reached via a new code path.

**P3 — header folding (obs-fold) not handled:** `[^\r\n]+` stops at `CRLF`, so a folded `Authorization: Bearer foo\r\n bar` would miss `bar`. Pre-existing limitation, not a regression; noting for completeness.

### Verdict

Approve. Both fixes are correct and the tests pass locally (`go test ./internal/providermodeldiscovery/... ./internal/redaction/...` — ok). The P3 items are non-blocking polish/note material.

PR #267, opened 2026-06-20
Add persistent exec command sessions

anandh8x:
## Summary
- add exec_command and write_stdin tools backed by a shared session manager
- keep long-running shell commands alive across tool calls so they can be polled or interrupted
- route exec_command through shell permissions, network prompts, command-prefix grants, sandbox metadata, and specialist tool exposure
- keep web_fetch public-remote only so localhost/private URLs go through shell commands where sandbox permissions apply

## Testing
- git diff --cached --check
- GOCACHE=/tmp/zero-go-cache go test ./internal/tools ./internal/agent ./internal/sandbox

## Summary by CodeRabbit

* **New Features**
  * Introduced `exec_command` and `write_stdin` tools for shell command execution, including long-running sessions with polling/input and session persistence
* **Documentation**
  * Updated system and tool guidance to prefer `exec_command` for new shell work, including correct shell syntax and foreground/polling patterns
  * Expanded web-fetch instructions to distinguish public URLs from localhost/private development URLs
* **Bug Fixes**
  * Improved early safety checks for `web_fetch` and local/loopback handling to avoid unnecessary permission prompts
* **Tests**
  * Added coverage for session behavior and early rejection flows


Vasanthdev2004 (changes requested):
## Review: Add persistent exec command sessions (#267)

### Overview
Introduces `exec_command` + `write_stdin` tools backed by an in-memory `execSessionManager`, plus a `PrePermissionRejecter` interface so `web_fetch` can refuse local/localhost URLs before any permission prompt. Routing through command-prefix grants, sandbox risk, and specialist tool categories is wired up. The design is sound and the test coverage for the new web_fetch classifier is solid. A few issues below worth addressing before merge.

**Verdict: REQUEST_CHANGES** — the session-manager lifecycle has a real leak/race that should be fixed, plus a couple of correctness nits.

---

### P1 — Blocking

**1. Per-tool session managers leak sessions and break cross-tool lookups.**
`internal/tools/registry.go:223` constructs a *new* `execSessionManager` inside `CoreShellToolsScoped` and passes it to both `NewScopedExecCommandTool` and `NewWriteStdinTool`. That is correct *within* a single call. However, `NewExecCommandTool` (`exec_command.go:532`) and `NewWriteStdinTool` (`exec_command.go:730`) each fall back to creating their **own** manager when passed `nil`. If any caller wires the two tools independently (or a specialist manifest constructs tools separately), `write_stdin` will always report "Unknown exec session_id" for sessions started by `exec_command`. The nil-fallback is a foot-gun: it silently produces two disjoint session pools. Consider either (a) removing the nil fallback and failing loudly, or (b) returning a shared package-level singleton. The current "works only if you happen to pass the same pointer" contract is too easy to violate.

**2. Finished-but-unpolled sessions are never reaped.**
`exec_command.go:606-610` removes a session from the manager only when `collect` observes it has exited within the yield window. If a command finishes *after* the yield window (so `exited == false` is returned to the model) and the model never calls `write_stdin` again, the `execSession` stays in `manager.sessions` forever — the goroutine in `startSession` (`exec_command.go:662-670`) has already exited, but the map entry, the `exec.Cmd`, the `stdin` pipe, and the `output` buffer all remain referenced. There is no background reaper, no session TTL, and no cleanup on session end. On a long-running agent this is an unbounded leak (one per abandoned long-running command). A background goroutine that removes sessions after `markDone` fires (e.g. `go func() { <-session.done; manager.remove(session.id) }()`) or a sweep on each `exec_command`/`write_stdin` call would fix it. The process exit goroutine already holds a reference to `session`, so it can call `manager.remove` after `markDone`.

**3. `write_stdin` permission is `PermissionAllow`, bypassing the shell gate for arbitrary stdin.**
`exec_command.go:749-753` marks `write_stdin` as `SideEffectShell` + `PermissionAllow` with the rationale that the session was already approved. But `write_stdin` accepts arbitrary `chars` and the model can send `\x03` to terminate *or* any other bytes — including input that drives an already-approved process to do things the original command prompt never authorized (e.g. an approved `python` REPL being told to `import os; os.system("rm -rf ...")`). The command-prefix grant for `exec_command` was for a *specific* command; `write_stdin` extends that trust to whatever the running process will do with new stdin. At minimum this needs to be documented as an intentional security tradeoff in the PR body / system prompt, and ideally `write_stdin` should still route through a lighter permission check (or at least be excludable from specialist tool sets). Right now a specialist manifest with `execute` gets `write_stdin` for free (`specialist/manifest.go:86`).

---

### P2 — Should fix

**4. `commandCtx` is detached from the tool-call `ctx`.**
`exec_command.go:625` creates `context.WithCancel(context.Background())` rather than deriving from the `ctx` passed to `run`. This is intentional for long-running sessions (the tool call returns before the process exits), but it means that if the *agent loop* is cancelled mid-`exec_command`-yield, the just-started process is **not** cancelled — `collect` returns early on `ctx.Done()` (`exec_command.go:711`) without calling `session.terminate()`, and the session is left running and stored in the manager. The model may never get a `session_id` to clean it up. Consider: on `ctx.Done()` during the initial `collect`, either terminate the session or ensure the `session_id` is still surfaced so the caller can reap it.

**5. `monitor` uses `context.Background()`, ignoring cancellation.**
`exec_command.go:641` calls `StartDenialMonitor(context.Background(), plan.MonitorTag)`. Same pattern as the existing `bash` tool so it's consistent, but worth noting that the denial monitor will run until `command.Wait()` returns; if `Wait` hangs (e.g. a process that ignores signals), the monitor goroutine is pinned. Not a regression, just inherited.

**6. `collect` notify channel can drop wakeups, causing full `yield_time_ms` latency.**
`execOutputBuffer.Write` (`exec_command.go:507-510`) does a non-blocking send to `notify` and drops it if the channel already has a pending item. `collect` drains the buffer at the top of each loop but only re-arms by reading `notify` in the `select`. If two writes happen between a drain and the `select`, the second notification is dropped — but the first `notify` wakes the select, the loop drains both chunks, and proceeds. So this is actually correct for *completeness* but can cost one extra `select` round-trip. No fix needed; flagging because the buffered-1 + drop pattern is subtle and a future reader may "fix" it incorrectly.

**7. `truncateExecOutput` uses `maxBytes = maxOutputTokens * 4`.**
`exec_command.go:875` assumes ~4 bytes/token, which under-counts for CJK/multibyte output and over-counts for ASCII. More importantly, the slice `output[:head]` and `output[len(output)-tail:]` can split a multibyte UTF-8 sequence, producing invalid UTF-8 in the tool output. Consider truncating on a rune boundary (`utf8.DecodeRune` / `utf8.DecodeLastRune`) to avoid mojibake being fed back to the model.

**8. No test covers the "command exits after yield" path.**
`exec_command_test.go` has two tests: one where the command outlives the short yield and is polled to completion, and one for an unknown session id. There's no test for: (a) a command that exits *during* the initial yield returning `exit_code` and no `session_id`, (b) reaping of finished sessions from the manager, (c) `write_stdin` sending `\x03` to terminate, (d) two concurrent `exec_command` calls getting distinct session ids. The web_fetch side has great coverage; the exec side is thin.

---

### P3 — Nice to have

**9. `prefix_rule` parameter is declared but never read.**
`exec_command.go:553` adds a `prefix_rule` array to the schema, but `run` never reads it. Either wire it into `matchCommandPrefix`-style logic or drop it from the schema to avoid misleading the model.

**10. `cwd` and `workdir` are both accepted and aliased, but the description says "Prefer workdir."**
`exec_command.go:548-550` — fine, but `grant_scope.go:36` and `risk.go:238` were updated to recognize `workdir`. Good. Just note that `dir`/`directory` are *also* accepted by `aliasedStringArg` (`exec_command.go:579`) but only `cwd`/`workdir` are in the sandbox path lists. A model that passes `dir` will have its cwd resolved but the sandbox won't see that path as a scope arg. Either add `dir`/`directory` to the sandbox lists too, or stop accepting them in `exec_command`.

**11. README wording.**
`README.md:185-189` — the new paragraph reads "For localhost/private dev-server URLs, use shell commands such as `curl` through `exec_command`... For local dev servers, start the server in the foreground with `exec_command`..." then immediately "For **search**, JS-rendered scraping,". The "For search" sentence is missing a blank line before it and now runs together with the dev-server guidance. Minor markdown formatting.

**12. `nextID` starts at 1000 with no overflow guard.**
`exec_command.go:419` — `int` on 64-bit platforms won't overflow in practice, but session ids are exposed to the model as JSON integers; if a long-running session ever crossed `math.MaxInt32` the JSON would be fine but any JS consumer would lose precision. Not realistic, just noting.

---

### Things that look right
- `PrePermissionRejecter` (`types.go:111-117`) is a clean interface and the loop integration (`loop.go:575-581`) correctly scrubs output/summary/meta via redaction before returning — matching the existing `toolResultFromTool` pattern.
- `validateParsedWebFetchURLBeforePermission` correctly reorders host classification *before* port validation so localhost URLs get the friendly "use bash with curl" hint instead of the "default ports" error. Test `TestWebFetchToolRejectsLocalhostBeforeDefaultPort` verifies this.
- `permissionSupportsPersistentDecision` (`loop.go:1754`) correctly opts `exec_command`/`write_stdin` *out* of persistent decisions, matching `bash` — so a one-off approval doesn't silently whitelist future shell runs.
- Specialist manifest (`specialist/manifest.go:86`) adds both new tools to `execute` and registers them as known — backward compatible with existing manifests.
- `command_prefix.go` refactor to `isShellCommandTool` is a clean generalization; the new test covers `exec_command`.

### Summary
- P1: 3 (session manager foot-gun, unreaped finished sessions, write_stdin permission model)
- P2: 5 (ctx detach, monitor ctx, truncate UTF-8, missing tests, collect notify — last is informational)
- P3: 4 (unread param, dir/directory sandbox gap, README formatting, id overflow)

Recommend addressing at least the P1s (reaper + shared manager contract + write_stdin permission rationale) before merge.

anandh8x:
Addressed the requested changes in d6586b1.

Changes made:
- Nil exec/write_stdin manager fallback now uses a shared default session manager, so independently constructed tools do not get disjoint session pools.
- Finished sessions are retained briefly for a final poll, then reaped automatically if abandoned.
- exec_command now terminates the session if the initial yield is interrupted by context cancellation.
- write_stdin is argument-aware: empty polling and Ctrl-C interrupts are allowed, while non-empty stdin goes through the prompt path and remains visible in auto mode.
- Output truncation now preserves UTF-8 boundaries.
- Added tests for shared manager fallback, initial completion, abandoned-session reaping, Ctrl-C interrupt, write_stdin permissions, registry permission handling, and UTF-8 truncation.
- Fixed the README spacing note and documented the stdin permission behavior.

Notes:
- prefix_rule is intentionally consumed by the agent permission-prefix flow rather than exec_command.run; this matches bash and is covered by command_prefix tests.
- dir/directory are already included in the sandbox path lists alongside cwd/workdir.

Vasanthdev2004 (approved):
Re-review of PR #267 — persistent exec command sessions

Thanks for the follow-up commit (d6586b1). I re-checked each previously flagged item against the new diff and added tests.

## P1 — all fixed

- **Shared `execSessionManager` nil-fallback** — Fixed. `CoreShellToolsScoped` now constructs one `execManager` and passes it to both `NewScopedExecCommandTool` and `NewWriteStdinTool` (internal/tools/registry.go:261-266). The nil-fallback in each constructor now resolves to the same package-level `defaultExecSessionManager` singleton (exec_command.go:574-576, 772-775), so even the fallback path keeps both tools in one pool. `TestIndependentExecCommandConstructorsShareDefaultManager` (exec_command_test.go:1000) verifies a session started via a nil-manager exec_command is reachable from a nil-manager write_stdin.
- **Finished-but-unpolled sessions reaped** — Fixed. `startSession` schedules `removeCompletedLater` (exec_command.go:461-471, 703), which waits on `session.done`, holds for `completedRetention` (30s default), then removes. Sessions that complete during the initial yield are removed immediately in `run` (exec_command.go:649-651), and `write_stdin` removes on poll if exited (exec_command.go:853-855). `TestExecCommandReapsFinishedUnpolledSession` (exec_command_test.go:1088) covers the unpolled path with a 10ms retention.
- **`write_stdin` permission bypass** — Fixed. `writeStdinTool.Safety().Permission` is now `PermissionPrompt` (exec_command.go:793), and `PermissionForArgs` returns `PermissionAllow` only for empty/Ctrl-C (`\x03`) and `PermissionPrompt` otherwise (exec_command.go:802-815). Both the registry (`effectiveToolPermission`, registry.go:1247-1252) and the agent loop (`effectivePermission`, loop.go:1125-1130) consult `ArgsPermissioner`, so the arg-specific decision is honored end-to-end. `TestWriteStdinPermissionForArgs` and `TestRegistryHonorsWriteStdinArgumentPermission` (exec_command_test.go:1149, 1165) verify both the unit and registry layers.

## P2 — all fixed

- **`commandCtx` detached from tool-call ctx** — Addressed. The command ctx is still rooted at `context.Background()` (exec_command.go:279), which is correct for a session that must outlive a single tool call, but `run` now terminates the session explicitly when the tool-call ctx is cancelled (exec_command.go:644-647: `if ctx.Err() != nil && !session.doneClosed() { session.terminate() }`), so cancel can no longer orphan a running process. `TestWriteStdinInterruptTerminatesSession` (exec_command_test.go:1118) covers the terminate path.
- **`truncateExecOutput` UTF-8 safety** — Fixed. `utf8Prefix`/`utf8Suffix` back up to `utf8.RuneStart` before slicing (exec_command.go:942-967). `TestTruncateExecOutputPreservesUTF8` (exec_command_test.go:1195) asserts the truncated string is `utf8.ValidString`.
- **Test coverage** — Substantially expanded. New tests cover: shared default manager, session polling, immediate-completion removal, unpolled reap, Ctrl-C interrupt, arg-based permission, registry permission gating, unknown session, UTF-8 truncation, web_fetch pre-permission URL classification, and loop-level local-URL rejection.

## P3 — remaining (non-blocking)

- **`prefix_rule` still declared but unread.** The schema advertises `prefix_rule` (exec_command.go:203) but `run` never reads it from `args`, so it has no effect today. Either wire it into `proposedCommandPrefix`/`matchCommandPrefix` (which already look at `cmd`) or drop it from the schema to avoid misleading the model.
- **`nextID` has no overflow guard.** `allocateID` (exec_command.go:434-440) increments indefinitely. Practically harmless on 64-bit (2B sessions), but a cheap wrap/overflow check would make it defensive.
- `workdir` is now in both `scopeArgKeys` and `requestPaths` (grant_scope.go:36, risk.go:238) — thanks, that closes the sandbox-alias gap.

## New issues introduced

None spotted. The `removeCompletedLater` goroutine performs a redundant `manager.remove` when the session was already removed by `run`/`write_stdin`, but `delete` on a missing key is a no-op, so it's harmless.

## Verdict

Approving. The P1/P2 blockers from the prior review are all resolved with targeted tests, and no new regressions are introduced. The two P3 items above are nice-to-haves and can be addressed in a follow-up.

PR #268, opened 2026-06-20
feat: clipboard image paste + provider 400 handler for image rejection

Vasanthdev2004:
## What this does

Two fixes for image input — the biggest UX gaps in the TUI:

### 1. Clipboard image paste (screenshots)
Zero could attach images via /image <path> and drag-drop file paths, but **pasting a screenshot from the OS clipboard was silently dropped** — clipboard.ReadAll() returns text only. Now when a paste arrives with empty text content, the TUI probes the OS clipboard for image bytes:

- **Windows**: PowerShell System.Windows.Forms.Clipboard.GetImage() → PNG bytes
- **macOS**: osascript clipboard info check → pngpaste extraction
- **Linux**: wl-paste (Wayland) / xclip (X11) with image MIME type detection

The image runs through the existing vision gate (modelSupportsVisionTUI) and 10 MiB size cap, then attaches as a pending image with a [Image #N] chip — same UX as /image <path>.

### 2. Provider 400 handler for image rejection
When a model rejects an image with HTTP 400 (like minimax-m3 — a text-only model that returns 400 for image content), the agent loop **retried endlessly in a 400-error loop**. The user had to kill the process. Now the loop detects 400s that mention image/vision/multimodal and aborts with a clear error:

> model X rejected the image: [error]. The model may not support image input — try switching to a vision-capable model (claude, gpt-4o, gemini)

Generic 400s (invalid model, bad request) are NOT caught — only image-specific rejections. Context-limit errors still retry via compaction as before.

## Files
- internal/imageinput/clipboard.go — new: ReadClipboardImage() platform reader
- internal/tui/clipboard.go — clipboardImageMsg + readClipboardImageCmd + routePaste image probe
- internal/tui/model.go — clipboardImageMsg handler + clipboardReadMsg image fallback
- internal/tui/image_attach.go — attachClipboardImage method
- internal/agent/loop.go — isImageRejectionError + abort before retry
- Tests: clipboard_test.go, clipboard_image_test.go, image_rejection_test.go

## Tests
- TestReadClipboardImageNoImageReturnsNil — nil when no image
- TestClipboardImageMsgAttachesImage — attaches to pendingImages
- TestClipboardImageVisionGateRefuses — non-vision model refuses
- TestClipboardImageTooLargeRefuses — oversize refuses
- TestIsImageRejectionError — 8 subtests (positive + negative cases)
- All existing tests pass

## Summary by CodeRabbit

* **New Features**
  * Added screenshot paste from the OS clipboard when the pasted text is empty, on Windows/macOS/Linux, and attach the image when a vision-capable model is used.
  * Expanded vision-capability detection using improved model-name/registry fallback rules.

* **Bug Fixes**
  * If the provider rejects image/multimodal input, the chat now stops early and prompts switching to a vision-capable model.
  * Prevents attaching clipboard images when unsupported, oversized, or with an unsupported clipboard media type; handles right-click clipboard reads more accurately.

* **Tests**
  * Added coverage for clipboard image reading, image-attachment gating (vision/size), and image-rejection error detection.


anandh8x (approved):
Reviewed #268 at 0e50265. No blocking findings. The changed packages pass locally, the Linux clipboard command injection fix avoids shell interpolation, and the provider image-rejection path avoids the retry loop before compaction. The failing Windows smoke check is in internal/tools/TestWriteStdinInterruptTerminatesSession, not in this PRs changed packages. Non-blocking note: the Windows clipboard temp path is interpolated into a single-quoted PowerShell string, so escaping single quotes or passing the path out-of-band would make that path more robust.

PR #269, opened 2026-06-20
feat(tui): transcript polish — left-rule cards, tool-call density, pinned plan

Vasanthdev2004:
-

## Summary by CodeRabbit

* **New Features**
  * Added a pinned plan panel above the composer that can expand or collapse based on available terminal height.
* **Style**
  * Redesigned tool cards to use a left-rail rule with right-aligned status glyphs, including improved target path display (shortened cwd/home-relative formatting) while keeping clickable links.
* **Bug Fixes**
  * Reduced transcript clutter by only rendering noteworthy permission decisions.
  * Prevented dense “stacking” by adding spacing between consecutive tool-card rows; improved collapsed “confirmation-only” result rendering.
* **Tests**
  * Updated and added regression coverage for tool-card layout, path shortening, redundancy collapsing, pinned-plan behavior, and spacing rules.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1642,6 +1642,15 @@ func (m model) pinnedTitleBar(width int) string {
 
 func (m model) footerView(width int) string {
 	var footer strings.Builder
+	// Pinned plan panel: sits directly above the composer so it stays visible
+	// while the transcript scrolls underneath (a streaming turn no longer pushes
+	// the plan off-screen). Budgeted to at most a third of the screen height; a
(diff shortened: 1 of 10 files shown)


anandh8x (approved):
Reviewed #269 at c785fb1. No blocking findings. The left-rule tool cards, permission-row density changes, and pinned plan panel are covered by focused tests, and the PR range passed diff-check plus internal/tui tests. Non-blocking note: subchat transcript rendering does not get the new blank line between consecutive tool cards, so consider mirroring that small spacing rule there later.

PR #270, opened 2026-06-20
Add managed background terminals

anandh8x:
Summary
- Adds managed exec sessions with bounded tracking, recent-output snapshots, stop controls, and Linux PTY transport behind the existing exec_command/write_stdin tools.
- Adds local TUI background terminal controls and rendering: /ps, /stop, status-line summary, exec_command/write_stdin terminal cards, and focused tests.
- Keeps web_fetch public-only while documenting localhost/private verification through shell/curl, and covers approved local network sandbox mode.
- Stops managed terminal sessions on TUI shutdown.

Verification
- GOCACHE=/tmp/zero-go-cache go test ./internal/tui -run 'TestExecCommandCardBody|TestToolCallSummaryDescribesExecSessions' -count=1
- GOCACHE=/tmp/zero-go-cache go test ./internal/tools ./internal/tui ./internal/agent ./internal/sandbox
- /bin/bash -lc 'GOCACHE=/tmp/zero-go-cache go test ./...'

## Summary by CodeRabbit

* **New Features**
  * Added `/ps` to list running background terminal sessions (status, age, working directory, and optional recent output).
  * Added `/stop` to stop a specific session or all background terminal sessions.
  * Added `tty: true` to `exec_command` for interactive terminal-style input.
  * Updated the TUI to display a background terminal summary in the status line.
* **Bug Fixes**
  * Improved `exec_command`/`write_stdin` interrupt, stop, and result display (including clearer “interrupted” and session hints), with safer PTY fallback.
* **Documentation**
  * Expanded README guidance for interrupting running commands and managing `/ps`/`/stop` sessions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,9 @@ into clean markdown locally, with no third party. For localhost/private dev-serv
 URLs, use shell commands such as `curl` through `exec_command` so sandbox network
 permission applies. For local dev servers, start the server in the foreground with
 `exec_command` and keep the returned `session_id`; use `write_stdin` to poll or
-interrupt it.
+interrupt it. Use `/ps` to list running background terminals and `/stop` or
+`/stop <session_id>` to close them. Interactive terminal-style commands can set
+`tty: true` on `exec_command` so stdin works like a terminal session.
 
 For **search**, JS-rendered scraping,
 whole-site crawls, PDF→markdown, and structured extraction, Zero ships the
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -72,8 +72,8 @@ work.
   foreground with exec_command and use write_stdin to poll or interrupt it;
   do not rely on `nohup`, `disown`, or backgrounding to keep it alive.
 - Use exec_command with `tty: true` for interactive terminal-style commands that
-  need stdin beyond Ctrl-C. Use `/ps` to inspect running background terminals and
-  `/stop` to close them.
+  need stdin beyond Ctrl-C. `/ps` and `/stop` are user-facing TUI commands; when
+  you need to clean up a running foreground command yourself, use write_stdin.
 - write_stdin with empty input polls an existing exec_command session, and
   `\u0003` interrupts it. Sending other stdin bytes may require approval because
   it can drive the running process beyond the original command.
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -76,7 +76,9 @@ work.
   you need to clean up a running foreground command yourself, use write_stdin.
 - write_stdin with empty input polls an existing exec_command session, and
   `\u0003` interrupts it. Sending other stdin bytes may require approval because
-  it can drive the running process beyond the original command.
+  it can drive the running process beyond the original command. Non-tty sessions
+  accept only empty polling and interrupt/stop input; start with `tty: true`
+  when you need to send literal stdin such as prompts, menu choices, or REPL text.
 - bash is the legacy one-shot shell tool. Prefer exec_command for new shell
   work, especially for local dev servers.
 - Treat tool output as ground truth. If a command fails, read the error, form a
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -71,6 +71,9 @@ work.
   package managers). If a command needs to keep running, run it in the
   foreground with exec_command and use write_stdin to poll or interrupt it;
   do not rely on `nohup`, `disown`, or backgrounding to keep it alive.
+- Use exec_command with `tty: true` for interactive terminal-style commands that
+  need stdin beyond Ctrl-C. Use `/ps` to inspect running background terminals and
+  `/stop` to close them.
 - write_stdin with empty input polls an existing exec_command session, and
   `\u0003` interrupts it. Sending other stdin bytes may require approval because
   it can drive the running process beyond the original command.
--- a/internal/sandbox/linux_helper_test.go
+++ b/internal/sandbox/linux_helper_test.go
@@ -128,6 +128,39 @@ func TestBuildLinuxSandboxBwrapArgsWrapsInnerSeccompStage(t *testing.T) {
 	}
 }
 
+func TestBuildLinuxSandboxBwrapArgsKeepsHostNetworkWhenAllowed(t *testing.T) {
+	helperPath := filepath.Join(t.TempDir(), LinuxSandboxHelperName)
+	if err := os.WriteFile(helperPath, []byte("helper"), 0o755); err != nil {
+		t.Fatalf("WriteFile helper: %v", err)
+	}
+	profile := DefaultPermissionProfile("/workspace")
+	profile.Network = NetworkPolicy{Mode: NetworkAllow}
+	args, err := BuildLinuxSandboxCommandArgs(LinuxSandboxCommandArgsOptions{
+		SandboxPolicyCWD:  "/workspace",
+		PermissionProfile: profile,
+		BlockUnixSockets:  true,
+		Command:           []string{"python3", "-m", "http.server", "8000"},
+	})
+	if err != nil {
+		t.Fatalf("BuildLinuxSandboxCommandArgs: %v", err)
+	}
+	config, err := ParseLinuxSandboxHelperArgs(args)
+	if err != nil {
+		t.Fatalf("ParseLinuxSandboxHelperArgs: %v", err)
+	}
+	bwrapArgs, err := BuildLinuxSandboxBwrapArgs(LinuxSandboxBwrapOptions{
+		Config:     config,
+		HelperPath: helperPath,
+	})
+	if err != nil {
+		t.Fatalf("BuildLinuxSandboxBwrapArgs: %v", err)
+	}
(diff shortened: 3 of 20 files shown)


Vasanthdev2004 (changes requested):
Reviewed the full diff (19 files, +1058/−22) plus the branch code. This is a well-built feature — managed sessions with bounded tracking, recent-output snapshots, a clean PTY/pipe transport split with a non-linux fallback, `/ps` + `/stop`, and CI green across all platforms. The session-manager locking discipline is sound (snapshot reads mutable fields under `session.mu`, immutable fields lock-free; `cancel()`/`terminate()` are idempotent and safe on done sessions; `store()` can't prune the session it's adding; `master.Close()` double-call is safe). Nice work.

One blocking correctness issue on the shutdown path, plus some test-depth gaps. Requesting changes on the P1; the rest are non-blocking.

---

### P1 — Background processes can survive TUI exit (orphaned on quit)

`internal/tui/model.go` `quit()` → `stopAllBackgroundTerminalSessions()` → `manager.stopAll()` → `session.terminate()` → `session.cancel()`, then `quit()` immediately returns `tea.Quit`.

`cancel()` only *requests* cancellation. The actual kill (`command.Cancel` → `syscall.Kill(-pid, SIGKILL)`) is delivered by Go's `exec` runtime **from inside `command.Wait()`**, asynchronously, bounded by `WaitDelay` (2s). Because `quit()` does not wait for the kills to land before `tea.Quit`, the Zero process can exit first. And critically, the children are **detached from Zero's process group** — PTY sessions set `Setsid=true` (`exec_pty_linux.go:69`), pipe sessions set `Setpgid=true` (`bash_proc_unix.go:32`) — so the OS does **not** cascade any signal to them when Zero dies. Net: a `python3 -m http.server` (or any long-running session) can keep running and holding its port after the TUI exits.

This is a race, not a guaranteed leak every time (the async SIGKILL often wins), but it's reachable and user-visible.

Fix is small since the plumbing already exists: have `stopAll()` (or `quit`) **join** each `session.done` with a bounded timeout after calling `terminate()`, so the kills are observed before `tea.Quit`. The `done` channel already closes in the wait goroutine after the kill.

### P2 — Latent `Start()` failure on the PTY→pipe fallback

`internal/tools/exec_transport.go`: when `ttyRequested` and `startPTYProcess` fails, you reset `command.Stdin/Stdout/Stderr = nil` and fall through to `startPipeProcess` — but the `command.SysProcAttr` mutations from `hardenPTYProcessLifetime` (`Setsid=true`, `Setctty=true`, `Ctty=0`) are **not** reset. On the pipe path fd 0 is no longer a tty, so `Setctty`/`Ctty=0` makes the child's `ioctl(TIOCSCTTY)` fail during exec → `startPipeProcess`'s `command.Start()` fails. So if `openPTY` succeeds but `Start()` fails transiently, the pipe retry reliably fails too. Reset `command.SysProcAttr = nil` (and the stream fields, already done) before `startPipeProcess`.

### P2 — Prune can silently evict a live, not-yet-polled session

`execSessionManager.store()` / `sessionToPruneLocked()`: when `maxSessions` (64) is hit, the oldest-by-`lastUsedAt` non-done session is terminated. `lastUsedAt` is only bumped on `write_stdin`/poll, so a freshly-started session that hasn't been polled yet (`lastUsedAt == startedAt`) is the most likely prune victim — exactly the one a caller is about to poll. The eviction SIGKILLs the process with no marker written to its output buffer, so the caller's next poll just sees an unknown/exited session with no explanation. Consider writing a sentinel line into the pruned session's buffer before `terminate()` (e.g. "session evicted: too many background terminals"), and/or rejecting the new session instead of killing a live one in the `len > 8 → sessions[0]` fallback branch.

### P2 — Test gaps (the feature's riskiest paths are the least covered)

- **Session pruning is entirely untested** — `store()`/`sessionToPruneLocked()`/`maxSessions`, including the `≤8` guard and the terminate-on-evict side effect. Easy to test: set `manager.maxSessions` small, assert eviction + that the evicted session was terminated.
- **`shouldInterruptExecSession` verb table is ~2 of ~16 cases**, and the important branch — that a **tty** session treats `"quit"`/`"exit"` as literal stdin rather than an interrupt — has no test. It's a pure function; worth pinning directly.
- **PTY-unavailable fallback** (`exec_transport.go`, tty requested → silent downgrade to non-tty pipe) is untested. The PTY happy path is linux-CI-only (the one PTY test skips on non-linux), so the silent-downgrade branch is the part most worth a unit test on every platform.

### P3 — System prompt doesn't warn the model about non-tty stdin

`internal/agent/system_prompt.md`: the prompt tells the assistant to use `write_stdin`, but doesn't state that a **non-tty** session rejects any literal input ("does not accept stdin") — only empty-poll and ``-interrupt work without `tty: true`. A model reading the prompt may try `write_stdin chars:"y\n"` on a non-tty session and just get an error. One sentence in the prompt would prevent that. (The `write_stdin` tool description hints at it, but the system prompt is the higher-signal surface.)

---

Nice-to-haves, not blocking: doc the `maxSessions=64`/oldest-evicted behavior so a vanished session has an explanation.

The non-tty stdin design itself is sound — I traced it, and literal text can never misdeliver to a non-tty process (it's always interrupt-or-rejected), so no footgun there.

Heads-up on a merge conflict (not a review point): this and #269 both touch `rendering.go`, `tool_render_registry.go`, `view.go`, and `specialist_card.go` (#269 reworks the tool-card frame; this adds `execCommandCardBody`). Whichever lands second will need a small rebase — flagging so it's not a surprise.

anandh8x:
@Vasanthdev2004 addressed the requested changes and rebased onto latest main to clear the conflict.\n\nChanges made:\n- Stop-all now waits with a bounded timeout for managed exec sessions to exit before TUI quit returns, so detached process groups do not race past shutdown.\n- PTY fallback now resets PTY-specific command state before retrying with pipe transport.\n- Live session eviction now writes an explicit eviction message before terminating instead of silently disappearing.\n- Added regression coverage for stop-all waiting, live eviction, PTY fallback reset, interrupt verb handling including tty literal input, and non-tty stdin guidance.\n- Updated the system prompt to state that non-tty sessions accept only poll/interrupt-stop input.\n\nValidation run locally: /bin/bash -lc 'GOCACHE=/tmp/zero-go-cache go test ./...'

Vasanthdev2004 (approved):
Re-reviewed after the latest push (`941bd0b`, `714d45a`). The blocking P1 from the prior review is genuinely fixed, the fix is correct and bounded, and no new P0/P1 was introduced. All six prior findings are addressed for real, not cosmetically. **Approving.**

### The P1 is resolved (orphan-on-quit)
- `quit()` now calls `stopAllBackgroundTerminalSessions()` **synchronously before** `tea.Quit` (`internal/tui/model.go:667-670`).
- `stopAll()` terminates every live session, then calls `waitForExecSessions(sessions, execSessionStopTimeout=3s)` (`exec_command.go:179`) — the kill is delivered to all sessions before any waiting begins.
- `waitForExecSessions` (`:184-207`) is bounded and can't hang: one shared `now+3s` deadline, `select` on `session.done` vs timer, correct timer drain. `done` closes only after `command.Wait()` reaps the process, and `WaitDelay=2s` < the 3s budget. Every process receives an unblockable kill before the wait, so no OS-level orphan survives. Covered by `TestStopAllWaitsForSessionsToExit`.

### Verified good
- **PTY→pipe fallback**: `resetExecCommandForPipeFallback` nils the whole `SysProcAttr` (Setsid/Setctty/Ctty) + `Cancel`/`WaitDelay` before re-hardening — fixes the latent `Start()` failure I flagged.
- **Prune sentinel**: `[zero] session evicted: …` written to the victim's own (mutex-guarded) output buffer with `notify`, before `terminate()` — a polling caller sees it.
- **New Windows path** (`bash_proc_windows.go`): `taskkill /T /F /PID` mirrors the unix negative-PID SIGKILL, invoked from `command.Cancel` inside `Wait`; no injection risk (`strconv.Itoa(Pid)`, direct exec), missing-taskkill backstopped by `WaitDelay`.
- **Tests**: pruning/eviction, the `shouldInterruptExecSession` verb table incl. the tty divergence (`quit`/`exit` → literal stdin under tty, not interrupt), the PTY-unavailable fallback (via the new `startPTYProcessFunc` var), and the stopAll join — all covered and non-vacuous.
- **System prompt**: non-tty `write_stdin` guidance now matches the implementation.

### One non-blocking follow-up (P3)
Session cap isn't enforced during a rapid-start *burst*: when the oldest session is live, `store()` terminates+sentinels it but doesn't delete it (delete only on `doneClosed()` — `exec_command.go:69-73`) while still adding the new session, so `len()` can reach `maxSessions+N` for the ~2s terminate→done window, and `sessionToPruneLocked()` re-selects the **same** victim each time (sorts by `lastUsedAt`, returns `sessions[0]`). It **self-heals** — `removeCompletedLater` is registered for every session (`:514`), so the map returns to steady state within ~30s of the burst. Bounded, not a leak, idempotent re-terminate — fine to merge. If you want to tighten it later: mark a victim 'terminating' in `store()` and skip it in `sessionToPruneLocked()` so the fallback picks the next distinct oldest, and add a multi-store flood test (the current eviction test only covers a single store).

Nice work addressing all of it — ship it.

PR #273, opened 2026-06-20
test(tools): de-flake TestWriteStdinInterruptTerminatesSession

Vasanthdev2004:
-

## Summary by CodeRabbit

* **Tests**
  * Improved interrupt/termination test reliability by using a resilient temporary directory cleanup to reduce platform-specific handle timing flakiness.
  * Updated the test to capture the current active session, send the interrupt, then deterministically wait for session completion (with a timeout) before performing the existing post-interruption assertions.

### DIFF
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -296,29 +296,35 @@ func TestWriteStdinInterruptTerminatesSession(t *testing.T) {
 		t.Fatalf("session_id is not numeric: %v", err)
 	}
 
-	// Send the interrupt, then wait deterministically for the process to be
-	// reaped (session.done closes) before asserting. The previous version relied
-	// on the write_stdin yield window (1000ms) being long enough for the async
-	// SIGKILL + reap to land, which flaked on slow CI (notably Windows smoke).
-	// Waiting on done removes the timing dependence: a generous safety timeout
-	// fails loudly if the kill genuinely hangs, but the common case returns the
-	// instant the process exits.
+	// Capture the live session BEFORE the interrupt so we can wait on its done
+	// channel afterwards (write_stdin removes a finished session from the
+	// manager, so manager.get would miss it post-interrupt).
 	session, ok := manager.get(sessionID)
 	if !ok {
 		t.Fatalf("session %d not found after start", sessionID)
 	}
-	session.terminate()
+
+	// The operation under test: write_stdin "\x03" must itself terminate the
+	// session (exec_command.go's Ctrl-C branch). This is what the regression
+	// guards — terminating the session here directly would let the test pass even
+	// if that branch were deleted.
+	interrupted := writeTool.Run(context.Background(), map[string]any{
+		"session_id":    sessionID,
+		"chars":         "\x03",
+		"yield_time_ms": 1000,
+	})
+
+	// De-flake: wait deterministically for the process to be reaped rather than
+	// relying on the 1000ms yield window being long enough for the async
+	// SIGKILL + reap to land (which flaked on slow CI, notably Windows smoke). A
+	// generous safety timeout fails loudly if the kill genuinely hangs; the
+	// common case returns the instant the process exits.
 	select {
 	case <-session.done:
 	case <-time.After(30 * time.Second):
 		t.Fatalf("interrupted session %d was not reaped within 30s", sessionID)
 	}
 
-	interrupted := writeTool.Run(context.Background(), map[string]any{
-		"session_id":    sessionID,
-		"chars":         "\x03",
-		"yield_time_ms": 1000,
-	})
 	if interrupted.Status != StatusOK {
 		t.Fatalf("interrupted session status = %s: %s", interrupted.Status, interrupted.Output)
 	}


anandh8x (dismissed):
Requesting changes on #273 at 409ee90. The revised TestWriteStdinInterruptTerminatesSession now calls session.terminate() directly before invoking write_stdin with chars="\x03". That bypasses the behavior the test is meant to cover: the write_stdin implementation terminates sessions in its Ctrl-C branch at internal/tools/exec_command.go:456-458, but with the session already terminated, a regression that removed or broke that branch would still pass. Please keep writeTool.Run(... chars="\x03" ...) as the operation that triggers termination, then wait deterministically for the same session.done afterward before asserting the final result. Verification: the changed test passes locally with GOCACHE=/tmp/zero-go-cache go test ./internal/tools -run TestWriteStdinInterruptTerminatesSession -count=5. Full internal/tools cannot run in this local sandbox because unrelated tests require bwrap networking and httptest listeners; GitHub CI for the PR is green.

gnanam1990 (changes requested):
## Review: de-flake TestWriteStdinInterruptTerminatesSession — REQUEST_CHANGES

The de-flake mechanism is the right idea. `resilientTempDir` (retrying `RemoveAll` past the Windows handle-release lag) and replacing the 1000ms yield-window dependence with a deterministic `<-session.done` wait should both fix the Windows-smoke flake.

### Blocker — the restructure removes the coverage this test exists for

The behavior under test is that `write_stdin` with `chars="\x03"` terminates the session — `exec_command.go:660-664`:

```go
if shouldInterruptExecSession(chars, session.tty) {
    interrupted = true
    session.terminate()
}
```

The new test calls `session.terminate()` directly and waits on `session.done` **before** invoking `writeTool.Run(..., "chars": "\x03")`. By the time write_stdin runs, the session is already reaped, so its `terminate()` is a no-op — the test would pass even if line 663 (`session.terminate()`) were deleted. The regression this test guards (write_stdin's Ctrl-C branch breaking) is no longer caught.

### Suggested fix

Keep the de-flake, but make `write_stdin "\x03"` the operation that triggers termination:

```go
interrupted := writeTool.Run(ctx, map[string]any{"session_id": sessionID, "chars": "\x03"})
select {
case <-session.done:
case <-time.After(30 * time.Second):
    t.Fatalf("interrupted session %d not reaped within 30s", sessionID)
}
// then assert on `interrupted`
```

Drop the standalone `session.terminate()`. That preserves both the flake fix and the coverage.

Vasanthdev2004:
Addressed the review — thanks, the missing-coverage catch was exactly right.

The de-flake mechanism stays (`resilientTempDir` + deterministic `<-session.done` wait), but `write_stdin "\x03"` is now the operation that **triggers** termination again, not a direct `session.terminate()` before it. I capture the session handle before the interrupt (so I can wait on `done` afterwards, since write_stdin removes a finished session from the manager), then call write_stdin, then wait.

Verified it actually guards the behavior now: with a **mutation test** — removing `session.terminate()` from write_stdin's Ctrl-C branch (`exec_command.go:457`) makes this test **fail** (session stays running). Still runs 15× locally without flaking. Ready for re-review.

gnanam1990 (approved):
## Overview
Test-only PR that de-flakes `TestWriteStdinInterruptTerminatesSession`: it waits deterministically on the session's `done` channel after a Ctrl-C interrupt instead of relying on the 1000ms yield window, and adds `resilientTempDir` to tolerate Windows handle-release lag during temp-dir cleanup.

## Strengths
- Targets the real root cause: waiting on `session.done` (closed once via `markDone`/`doneOnce`) is the canonical reap-completion signal, replacing a timing-dependent assertion with a synchronization point; a 30s safety timeout fails loudly if the kill genuinely hangs.
- The session handle is captured via `manager.get` **before** the interrupt — necessary because `write_stdin` calls `manager.remove` once the session exits, so a post-interrupt lookup could miss it. The comment documents this accurately.
- **Coverage is preserved (non-vacuous):** this version triggers termination through the actual Ctrl-C branch (`chars == "\x03"` → `session.terminate()`, `exec_command.go:456-457`) rather than terminating the session directly, so it still fails if that branch is removed. This addresses my earlier review.
- `resilientTempDir` retries `os.RemoveAll` for up to 5s and degrades to leaking the dir rather than failing — an appropriate tradeoff for the platform quirk.

## Verdict
**APPROVE** — clean, well-commented, fixes the flakiness without weakening coverage. (Confirmed green on a re-run of the related ubuntu smoke job.)

Vasanthdev2004:
Rebased onto `main` (resolves the conflict — branch is `MERGEABLE` again) and the de-flake now keeps the coverage you flagged.

Note: `main` already picked up the de-flake infra (`resilientTempDir` + the deterministic `<-session.done` wait), but its copy of the test still had the flawed ordering — `session.terminate()` *before* `write_stdin`. This PR is now just that correction against current main: capture the session handle first, make `write_stdin "\x03"` the operation that **triggers** termination, then wait on `done`. **Verified by mutation** — removing `session.terminate()` from the Ctrl-C branch (`exec_command.go:664`) makes the test fail. Runs 15× locally without flaking. Diff is one file (+19/−13). Ready for re-review.

gnanam1990 (approved):
## Re-review — latest commit (`c201d03`) ✅

Re-reviewing the current head after the new push. The test now correctly:

- captures the session handle via `manager.get` **before** the interrupt (necessary — `write_stdin` removes a finished session from the manager);
- makes `write_stdin "\x03"` the operation that terminates the session (the actual Ctrl-C branch under test, `exec_command.go`), rather than calling `session.terminate()` directly — so the assertion genuinely guards that branch;
- then waits deterministically on `session.done` (the de-flake), with a 30s safety timeout.

This resolves the coverage concern from the earlier reviews (the commit notes it's verified by mutation — deleting `session.terminate()` from the Ctrl-C branch fails the test — and runs 15× locally without flaking).

## Verdict
**APPROVE** — de-flake retained, coverage restored, non-vacuous.

PR #274, opened 2026-06-20
Harden background daemon: stdout integrity, lifecycle, and single-instance lock

gnanam1990:
A focused deep review of `internal/daemon` — the worker pool, launcher, control server, session fan-out, and single-instance lock — surfaced a stdout-integrity bug plus several lifecycle and locking hazards. Each fix ships with a regression test; the daemon suite passes under `-race`, and `go build ./...` / `go vet` / `gofmt` are clean.

### Worker stdout integrity
- **Large lines were silently dropped.** A worker's stdout was read with a **capped** `bufio.Scanner` (64 KiB grow, 1 MiB hard cap). A legitimately large stream-json line from our own trusted worker (a big final answer or tool result) over the cap was dropped — the output simply vanished. Replaced with an uncapped `bufio.Reader`; the 1 MiB cap is correct only for untrusted network frames, not this local pipe.
- **Read errors were swallowed as success.** A stdout read error ended the pump and the run reported a clean exit. It now surfaces as a run failure, and the worker is killed first so it can't block writing to the now-unread pipe and hang `Wait`.

### Control-server lifecycle
- **An idle client could wedge shutdown forever.** `Shutdown` closed the listener but not in-flight connections, so a handler blocked on an idle/hostile client's read never returned and `wg.Wait()` hung (taking SIGTERM cleanup with it). Open connections are now tracked and closed on shutdown.
- **A shutdown requested at startup could be lost.** A `Shutdown` in the bind→serve window left the accept loop blocking forever on a connection that never comes. The listener is now mutex-guarded and the shutdown signal re-checked right after bind.
- Buffered-history replay now honors shutdown too, so a large buffer can't delay drain.

### Session fan-out
- **Slow subscribers silently lost output.** A subscriber whose channel filled had lines dropped with no indication. It now receives a namespaced gap notice when the channel drains, so a lagging consumer learns its stream is incomplete (full output remains available via a fresh attach).
- **Latent panic.** The broadcast copied the subscriber set outside the lock, so a subscriber cancelling mid-send could close a channel the broadcaster was about to send on (`send on closed channel`). The lock is now held across the non-blocking sends (which never block).

### Single-instance lock
- **Two instances could both "hold" the lock.** Stale-lock reclaim did a blind remove, so two daemons starting at once could both remove-and-recreate and end up both holding the single-instance lock. Reclaim now uses an atomic rename-with-verify: only one racer wins the rename, and a holder that reacquires in the gap is restored rather than stolen.

### Client + pool
- Bound the control handshake with an I/O deadline so a peer that accepts but never replies can't wedge `Dial`/`NewClientConn`.
- Key the active-worker map by a monotonic worker id rather than PID — the OS reuses a PID the instant a worker exits, which could drop the wrong handle.
- Group-terminate on context cancel so a stuck worker's children aren't orphaned.

### Tests
New/extended: `TestReaderLinesHandlesLargeLine`, `TestPoolRunSurfacesStdoutReadError`, `TestReclaimStaleLockRemovesDeadHolder`, `TestReclaimStaleLockRestoresLiveHolder`, `TestClientHandshakeTimesOut`, `TestSessionSlowSubscriberGetsGapNotice`; shutdown/listener races are covered by the existing suite run under `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stale lock reclamation to avoid conflicting lock states
  * Enhanced process group termination and stdout error handling to prevent hanging workers
* **New Features**
  * Added timeout protection to control connection handshakes
  * Slow session subscribers now receive gap notifications for missed output
* **Improvements**
  * Strengthened server shutdown coordination with active connection tracking
  * Improved handling of large streamed lines and tighter worker output failure detection

### DIFF
--- a/internal/daemon/client.go
+++ b/internal/daemon/client.go
@@ -39,7 +39,17 @@ func NewClientConn(conn net.Conn) (*Client, error) {
 	return c, nil
 }
 
+// handshakeTimeout bounds the hello/hello-ok exchange so a hung or hostile peer
+// can't wedge Dial/NewClientConn forever. It covers only the handshake; the
+// deadline is cleared before the (long-lived, unbounded) streaming phase. It is a
+// var so tests can shorten it.
+var handshakeTimeout = 10 * time.Second
+
 func (c *Client) handshake() error {
+	// Bound only the handshake. A peer that accepts the connection but never
+	// completes the version exchange would otherwise block ReadControl forever (D9).
+	_ = c.conn.SetDeadline(time.Now().Add(handshakeTimeout))
+	defer func() { _ = c.conn.SetDeadline(time.Time{}) }() // clear before streaming
 	if err := WriteControl(c.conn, Ctrl{Type: CtrlHello, Version: ProtoVersion}); err != nil {
 		return err
 	}
--- a/internal/daemon/client_test.go
+++ b/internal/daemon/client_test.go
@@ -1,0 +1,36 @@
+package daemon
+
+import (
+	"net"
+	"testing"
+	"time"
+)
+
+// TestClientHandshakeTimesOut verifies the handshake is bounded: a peer that
+// accepts the connection but never completes the version exchange makes
+// NewClientConn fail promptly instead of blocking forever (D9).
+func TestClientHandshakeTimesOut(t *testing.T) {
+	orig := handshakeTimeout
+	handshakeTimeout = 100 * time.Millisecond
+	defer func() { handshakeTimeout = orig }()
+
+	// net.Pipe is synchronous: with no reader on the server end the client's hello
+	// write blocks, so only the handshake deadline can unblock it.
+	client, server := net.Pipe()
+	defer server.Close()
+
+	done := make(chan error, 1)
+	go func() {
+		_, err := NewClientConn(client)
+		done <- err
+	}()
+
+	select {
+	case err := <-done:
+		if err == nil {
+			t.Fatal("NewClientConn must fail when the peer never completes the handshake")
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatal("NewClientConn blocked past the handshake timeout — deadline not applied")
+	}
+}
--- a/internal/daemon/launcher.go
+++ b/internal/daemon/launcher.go
@@ -49,7 +49,7 @@ func scrubWorkerEnv(env []string) []string {
 type execWorker struct {
 	cmd    *exec.Cmd
 	stdout io.ReadCloser
-	lines  *scannerLines
+	lines  Lines
 	pid    int
 }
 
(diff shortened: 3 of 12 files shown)


Vasanthdev2004 (approved):
Reviewed the daemon hardening closely — this is the riskiest of the three and it holds up well. The bulk is materially safer than main and well-tested. **Approving.** One narrow P2 and one P3 are follow-ups, not blockers (details below).

### Non-blocking follow-ups

**[P2 — concurrency] `reclaimStaleLock` restore-rename can clobber a third daemon's fresh lock.**
The restore at `internal/daemon/lock.go:91` uses `os.Rename(reclaimed, path)`, which on POSIX **replaces** whatever now sits at `path`. With 3 concurrent starts against a stale lock:
- Daemon A reads stale→dead, about to reclaim.
- Daemon C wins its own reclaim and creates a fresh O_EXCL lock at `path` (C's live pid).
- A runs `os.Rename(path, reclaimed)` (`lock.go:85`), moving C's *live* lock aside. While it's sidelined, daemon D's O_EXCL create succeeds (path is empty) → D holds a lock.
- A re-reads `reclaimed`, sees C's live pid (`lock.go:88`), restores via `os.Rename(reclaimed, path)` (`lock.go:91`) — **replacing D's lock**. Now C (restored pid) and D (create succeeded) both believe they hold it.

Needs 3 concurrent starts + a two-file-op window, and it's **strictly less likely than the blind-`Remove` race this PR fixes** — so the PR is a clear net improvement and I don't think it should block. Suggested fix: make the restore non-replacing (restore only if the target doesn't exist — `os.Link`/O_EXCL semantics — else `os.Remove(reclaimed)`), or read the pid *before* renaming and only rename-aside a confirmed-dead pid.

**[P3 — comment] `gapNotice`/subscriber doc overstates recoverability.** The comments say dropped lines "remain available via a fresh attach." The buffer is a ring capped at `defaultSessionBuffer = 1024` that drops oldest lines once full, so a far-behind subscriber can lose lines that *also* aged out — a fresh attach won't recover them. The gap notice itself is the honest signal and is correct; just soften the comment to "available via a fresh attach **if still within the ring window**." No functional bug.

### Verified-good (the substance)
- **Lock writer:** a partial pid write removes the file and fails (`lock.go:43-46`), so no malformed lock is left to be misread as stale; the per-process sideline name (`%s.stale.%d-%d` with pid + atomic seq) avoids self-collision.
- **Stdout integrity:** `readerLines` swaps the capped `bufio.Scanner` for an uncapped `bufio.Reader`, so a large stream-json line from the *trusted* worker isn't dropped (the 1 MiB cap is only right for untrusted frames). `TestReaderLinesHandlesLargeLine` (2 MiB + no-trailing-newline) covers it.
- **Pool:** `runOnce` now Kills-then-Waits on a stdout read error and surfaces it instead of a bogus clean success (`TestPoolRunSurfacesStdoutReadError`); pid→monotonic-worker-id re-keying eliminates a pid-reuse map collision (D10).
- **Session fan-out:** the lock is held across the non-blocking sends (every case has `default`), closing the send-on-closed-channel race vs `cancel()`/`finish()`; `TestSessionSlowSubscriberGetsGapNotice` covers the drop-count + gap flush.
- **Server lifecycle:** `conns` tracking + Shutdown closing the listener and every tracked conn so blocked reads unblock and `wg.Wait()` finishes; the bind-window `<-s.done` closes the shutdown-during-bind hole. Leak- and deadlock-free.
- **Client handshake** deadline bounds only the hello exchange and clears before streaming (`TestClientHandshakeTimesOut`); **launcher** `cmd.Cancel` kills the whole process group instead of orphaning children.

Solid work — the one P2 is a narrower residual race than what's fixed. Recommend a fast follow-up on `lock.go:91`, but it shouldn't gate the merge.

PR #275, opened 2026-06-20
Audit tail: remaining medium/low correctness fixes (CLI, config, OAuth, tools, LSP, TUI)

gnanam1990:
The tail of the deep-audit remediation: the remaining confirmed medium/low correctness findings across six subsystems. Each fix ships with a regression test; `go build ./...`, `go vet`, and `gofmt` are clean and every touched package's suite passes.

### CLI
- **`zero -p "-foo"` was rejected.** The top-level prompt path passed `--prompt <value>` as two args, so a prompt whose first character is a dash was mistaken for a flag and failed with "--prompt requires a value". It now forwards the inline `--prompt=<value>` form (matching the cron path), which takes the value verbatim.

### Config
- **`parseThinkTags` was silently dropped on every profile merge.** `mergeProfile` didn't carry the field; an explicit value now carries over and a nil (unset) never clobbers an existing one.

### OAuth
- **`token_type` was lost across refreshes.** A refresh response commonly omits `token_type`, and `Refresh` seeded the base token without it, so the type was dropped. It's now carried forward.
- **`RefreshScheduler` was permanently inert after `Stop`.** `Stop` never reset `started`/`done`, so a later `Start` was a no-op forever. `Stop` now resets them so the scheduler can run again.

### Tools
- **`apply_patch` mis-parsed diff paths containing spaces.** Header parsing split `--- `/`+++ ` lines on whitespace, truncating any path with a space. It now takes the whole post-prefix value, drops a trailing tab-delimited timestamp, and undoes git's C-style quoting, so space-bearing names are recognized (so their tracker baseline is correctly cleared). Not a confinement bypass; a correctness fix.

### LSP
- **`Server.Shutdown` ignored its context** and could block up to the full grace window per server. It now kills immediately on context cancel.
- **`Manager.Shutdown` was serial** (N×grace for N servers). Servers are now shut down concurrently, bounding total shutdown to one grace window.

### TUI
- **`/theme auto` reused a stale background reading.** Switching to auto from an explicit theme now re-probes the terminal background (`RequestBackgroundColor`) so it re-detects light/dark instead of painting the wrong palette.
- **Specialist card showed "0 tool calls · 0 tokens".** The tool-call count is now wired from progress events; the token segment is omitted when the total is 0 rather than rendering a structurally-wrong "0 tokens".
- **Resume erased some failed Task delegations / desynced call→result dedup.** The tool-call row now extracts its id the same way as the result row and the specialist pre-pass (`toolCallId` first), so both branches key on the same string.
- **Plan panel:** `height` takes an injected clock (consistent with `renderPlanPanel`, testable under a fake clock), and two steps with identical content now inherit distinct prior timestamps positionally instead of both collapsing onto the first match.

### Tests
New/extended: `TestParseExecArgsInlinePromptKeepsLeadingDash`, `TestMergeProfilePreservesParseThinkTags`, `TestRefreshPreservesTokenTypeWhenOmitted`, `TestRefreshSchedulerRestartsAfterStop`, `TestPatchHeaderPathsHandlesQuotedAndSpacedNames` (+ plain-path guard), `TestRenderSpecialistCardOmitsZeroTokens`, `TestSpecialistTrackerIncrementToolCountByCallID`, `TestUpdateFromItemsDuplicateContentKeepsDistinctTimestamps`, and the `plan height` tests updated for the injected clock.

### Deliberately not changed
- The streaming-fade wrap-age mapping already documents and clamps the visual→logical approximation (the markdown renderer doesn't expose a wrap map); the OpenRouter loopback callback documents in-code why no CSRF `state` is required (random ephemeral port + short window). Both are left as the existing documented behavior.

## Release Notes

* **Bug Fixes**
  * Fixed CLI prompt parsing to preserve leading dashes/special characters
  * Fixed OAuth token type being preserved during refresh when omitted
  * Improved patch parsing for quoted paths with spaces (and correct header handling)
  * Fixed provider profile merges for `ParseThinkTags`
  * Fixed TUI specialist progress/counting and hid “0 tokens”; improved transcript/tool-call correlation and plan-step timestamp handling

* **Performance/Behavior**
  * Updated language-server shutdown to run concurrently; added context-aware force-kill behavior

* **New Features / Tests**
  * Added/expanded tests for prompt parsing, token refresh, patch paths, scheduler restarts, and theme auto background re-probing

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -256,7 +256,10 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 			return writePromptRequired(stderr)
 		}
 		// Forward leading --add-dir occurrences so exec's own parser collects them.
-		execArgs := append(addDirFlagArgs(addDirs), "--prompt", args[1])
+		// Use the inline --prompt=<value> form so a prompt whose first character is a
+		// dash (e.g. `zero -p "-foo"`) is taken verbatim instead of being mistaken for
+		// a flag and rejected with "--prompt requires a value" (matches the cron path).
+		execArgs := append(addDirFlagArgs(addDirs), "--prompt="+args[1])
 		execArgs = append(execArgs, args[2:]...)
 		return runExec(execArgs, stdout, stderr, deps)
 	case "exec":
--- a/internal/cli/exec_parse_prompt_test.go
+++ b/internal/cli/exec_parse_prompt_test.go
@@ -1,0 +1,23 @@
+package cli
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestParseExecArgsInlinePromptKeepsLeadingDash guards the M19 fix: the top-level
+// `zero -p "<prompt>"` path forwards the prompt as the inline `--prompt=<value>`
+// form precisely so a prompt whose first character is a dash is taken verbatim
+// instead of being mistaken for a flag and rejected.
+func TestParseExecArgsInlinePromptKeepsLeadingDash(t *testing.T) {
+	opts, early, err := parseExecArgs([]string{"--prompt=-foo bar"})
+	if err != nil {
+		t.Fatalf("inline dash-leading prompt rejected: %v", err)
+	}
+	if early {
+		t.Fatal("a normal prompt parse must not request early exit")
+	}
+	if got := strings.TrimSpace(strings.Join(opts.promptParts, " ")); got != "-foo bar" {
+		t.Fatalf("inline prompt = %q, want %q", got, "-foo bar")
+	}
+}
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -415,6 +415,11 @@ func mergeProfile(base ProviderProfile, next ProviderProfile) ProviderProfile {
 	if next.Description != "" {
 		base.Description = next.Description
 	}
+	// A nil *bool means "unset"; only an explicit value overrides. Without this the
+	// parseThinkTags setting was silently dropped on every profile merge (L17).
+	if next.ParseThinkTags != nil {
+		base.ParseThinkTags = next.ParseThinkTags
+	}
 	return base
 }
 
--- a/internal/config/resolver_test.go
+++ b/internal/config/resolver_test.go
@@ -1486,6 +1486,20 @@ func TestResolveRejectsNegativeDeferThreshold(t *testing.T) {
 	}
 }
 
+func TestMergeProfilePreservesParseThinkTags(t *testing.T) {
+	yes := true
+	// next supplies the setting; a nil base must inherit it (it was being dropped).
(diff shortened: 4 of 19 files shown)


Vasanthdev2004 (approved):
Reviewed each hunk across the CLI / config / OAuth / tools / LSP / TUI slices for correctness and regressions. No P0/P1, no blockers. **Approving.**

### Verified-good (per hunk)
- **CLI dash-prompt (M19):** `zero -p "-foo"` now forwards as inline `--prompt=<value>` (`cli/app.go:259`) so a dash-leading prompt is taken verbatim instead of parsed as a flag; matches the cron path. `TestParseExecArgsInlinePromptKeepsLeadingDash` covers it.
- **Resolver ParseThinkTags (L17):** `mergeProfile` preserves the `*bool` tri-state — only a non-nil `next` overrides (`resolver.go:418`). Both directions tested.
- **OAuth refresh TokenType (L15):** `base` seeds `TokenType: current.TokenType` (`flow.go:147`) so a refresh that omits `token_type` carries it forward; `PostToken` only overwrites when supplied. Correct and consequential; tested.
- **OAuth scheduler restart (L14):** `Stop()` resets `started=false`/`done=nil` (`scheduler.go:118`) under `s.mu` so a Stop'd scheduler can restart, race-free. Tested.
- **apply_patch quoted/spaced paths (L18):** takes everything after the fixed 4-byte `--- `/`+++ ` prefix, drops a trailing tab timestamp, unquotes git C-quoted paths via `strconv.Unquote`, skips `/dev/null` — aligns the validator with `git apply` (a real improvement: paths with spaces no longer truncate). Both cases tested.
- **LSP Shutdown (L19):** manager goes serial→per-server goroutine joined by a WaitGroup with errors under a mutex (one grace window, no race); server adds `case <-ctx.Done()` to kill immediately instead of waiting the full grace — fixes a real process-leak under the caller's 5s budget.
- **TUI specialist tool-count (M18):** `incrementToolCount(msg.toolCallID)` keyed by call-id while the entry is still id-keyed (reconciled on completion); tested.
- **TUI /theme auto re-probe (M17):** returns `tea.RequestBackgroundColor` only when `themeMode == themeAuto` so auto re-detects light/dark off the fresh reading.
- **Plan panel (L22):** `prevUsed[]` consumes each prior step once, so duplicate-text steps inherit distinct timestamps positionally instead of collapsing onto the first match; tested. The `height(width, now)` signature change is safe — no production caller, test-only plumbing.
- **Specialist card (M18):** the token segment is appended only when `tokenCount > 0`, avoiding a misleading "0 tokens"; tested.
- **session.go id extraction (L20):** `firstNonEmptyString(toolCallId, id)` matches the tool-result branch and specialist pre-pass, keeping call→result dedup + the M10 specialist-skip in sync for payloads carrying `toolCallId` but not `id`.

Clean, well-scoped, each fix tested. Nice tail-end cleanup.

PR #276, opened 2026-06-20
feat: five capability features from the reference-agent comparison

Vasanthdev2004:
-

## Release Notes

* **New Features**
  * Added a guided `/init` flow to generate a repo-level `AGENTS.md`.
  * Added user-defined slash commands from `.zero/commands`, including template expansion, autocomplete, and execution support.
  * Added semantic code navigation via `lsp_navigate` (definition, references, implementation, workspace symbols), with clear “unavailable” fallback.
* **Improvements / Bug Fixes**
  * Improved streaming resiliency by reconnecting after transient disconnects.
  * Reduced expensive context handling by pruning stale tool outputs before costly summarization.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -340,6 +340,19 @@ func (state *compactionState) maybeCompact(
 		return messages
 	}
 
+	// CHEAP FIRST STAGE: reclaim context at zero token/latency cost by pruning
+	// the bodies of old, large tool results (the model has already acted on
+	// them). If that brings us back under threshold, skip the paid summarizer
+	// entirely and preserve recent turns verbatim.
+	if pruned, reclaimed := pruneStaleToolOutput(messages, state.preserveLast); reclaimed > 0 {
+		messages = pruned
+		size = estimateTokens(messages) + toolTokens
+		if size <= state.threshold {
+			state.lowWaterMark = size
+			return messages
+		}
+	}
+
 	compacted, err := Compact(messages, CompactionOptions{
 		PreserveLast: state.preserveLast,
 		Summarize:    summarizeClosure(ctx, provider, state.onUsage),
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -165,7 +165,11 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			options.OnContext(MeasureContext(messages, request.Tools, options.ContextWindow))
 		}
 
-		stream, err := provider.StreamCompletion(ctx, request)
+		// A transient upstream disconnect on the initial connect is retried with
+		// backoff (before any content is forwarded, so no OnText is duplicated);
+		// a context-limit / image-rejection / cancellation is NOT retried here —
+		// those fall through to their own handlers below.
+		stream, err := streamWithReconnect(ctx, provider, request, reconnectNoticeFor(options))
 		if err != nil {
 			if isImageRejectionError(err) {
 				result.Messages = copyMessages(messages)
--- a/internal/agent/prune.go
+++ b/internal/agent/prune.go
@@ -1,0 +1,123 @@
+package agent
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// Stale tool-output pruning reclaims context at ZERO token/latency cost before
+// the loop falls back to the paid LLM summarizer. A long, dump-heavy session
+// accumulates large read_file/grep/glob/bash tool results that the model has
+// long since acted on; their bodies are dead weight. Pruning replaces those
+// older bodies with a compact placeholder (keeping the tool message + its
+// ToolCallID so provider replay stays valid and the model knows what was
+// there), preserving recent turns exactly — no paraphrase loss.
+const (
+	// pruneProtectRecentTokens is the trailing window of tool output kept
+	// verbatim — the model is most likely still using it.
+	pruneProtectRecentTokens = 40000
+	// pruneMinReclaimTokens gates the whole pass: only prune when the reclaimable
+	// (older, large) tool output exceeds this, so a short session is untouched.
+	pruneMinReclaimTokens = 20000
+	// pruneMinBodyTokens skips small tool results — replacing a tiny body saves
+	// nothing and just loses information.
+	pruneMinBodyTokens = 200
+)
+
+// prunedPlaceholder is the body a pruned tool result is replaced with. It names
+// the tool and the original size so the model can re-fetch if it needs to.
+func prunedPlaceholder(toolName string, originalTokens int) string {
+	if toolName == "" {
+		toolName = "tool"
+	}
+	return fmt.Sprintf("[pruned %s output (~%d tokens) to reclaim context — re-run the tool if you need it again]", toolName, originalTokens)
+}
+
+// pruneStaleToolOutput walks messages newest-first, protects the last
+// preserveLast messages and a trailing pruneProtectRecentTokens of tool output,
+// then replaces the body of older large tool results with a placeholder. It
+// returns the (possibly) rewritten slice and the number of tokens reclaimed.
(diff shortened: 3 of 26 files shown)


gnanam1990 (changes requested):
## Overview
Adds several "reference gap" features: a read-only `lsp_navigate` tool (definition/references/implementation/workspace_symbol via a managed language server), an `usercommands` package for markdown-defined slash commands (personal `<userConfigDir>/zero/commands` + project `.zero/commands`), autocomplete for those commands, a guided `/init` that seeds `AGENTS.md`, and stale-tool-output pruning + mid-stream reconnect in the agent loop.

## Strengths
- **`reconnect.go` is correct:** it retries the **connect only** (before any content is forwarded), so no already-streamed `OnText` is ever duplicated; bounded attempts (2) with backoff; `shouldReconnect` excludes ctx-cancel/expiry and context-limit errors so it never fights compaction. (Independently reviewed — clean.)
- **`prune.go` is safe:** it only replaces *older, large* tool-message bodies with a re-fetchable placeholder, keeps the message + `ToolCallID` (provider replay stays valid), preserves a 40k-token recent window verbatim, and is idempotent. No paraphrase/data loss. (Independently reviewed — clean.)
- `/init` fact-gathering is safe by construction (no shell injection, no network, credentials stripped); `lsp_navigate` reads file content before querying the server.

## Blocker

### `lsp_navigate` bypasses workspace path confinement (arbitrary file read)
`internal/tools/lsp_navigate.go:118-127` (`readFile`) + `internal/tools/registry.go:217`.

Every sibling read-only tool in `CoreReadOnlyToolsScoped` is wired **with** `scope` and resolves the model-supplied path through `resolveScopedReadPath` → `resolveWorkspacePath`, which rejects `..` and absolute paths (`"%s must stay inside the workspace"`):

```go
NewScopedReadFileTool(workspaceRoot, scope),
NewScopedGlobTool(workspaceRoot, scope),
NewScopedGrepTool(workspaceRoot, scope),
...
NewLSPNavigateTool(workspaceRoot),   // <-- no scope
```

`lsp_navigate` drops the scope, and `readFile` does no confinement check:

```go
abs := path
if !filepath.IsAbs(abs) { abs = filepath.Join(tool.workspaceRoot, path) }
data, err := os.ReadFile(abs)
```

So `path:"/etc/passwd"` (absolute → used verbatim) or `path:"../../../../etc/ssh/id_rsa"` (`..` escapes the root) reads arbitrary files off disk — and `lsp.Manager.absPath` (`internal/lsp/manager.go:171`) has the same gap, so the out-of-workspace content is also `didOpen`'d into the language-server subprocess. The read-error branch even echoes the attacker-controlled absolute path back to the model. This breaks the confinement boundary the adjacent read-only tools enforce (and that the MCP roots treat as authoritative) — a real, newly-introduced regression, reachable via indirect prompt injection.

**Fix:** add `NewScopedLSPNavigateTool(workspaceRoot, scope)`, wire it from `CoreReadOnlyToolsScoped`, resolve the path through `resolveScopedReadPath` **before** reading and before building the LSP URI, and guard `lsp.Manager.absPath`/`Navigate` against paths outside `m.workspaceRoot`. A regression test with `path:"/etc/passwd"` and `path:"../.."` asserting an out-of-workspace error would lock it in.

## Non-blocking notes
- **[minor] Project (repo-sourced) slash commands have no provenance marker** — `internal/tui/autocomplete.go` / `usercommands.go` (`ProjectDir = workspaceRoot/.zero/commands`). A repo-authored command is surfaced and invoked (`Expand` → `launchPrompt`) indistinguishably from a personal one — an indirect prompt-injection vector. Minor because invocation is opt-in (user types the exact `/name`), `.zero/` is already a trust boundary here, and tool calls stay permission-gated. Suggest a `[repo]` marker in the menu and/or an inline notice on invocation.
- **[nit] Frontmatter `model`/`agent` parsed but never applied** — `usercommands.go:122-128` parses `cmd.Model`/`cmd.Agent` and the docs advertise a working `model:` override, but `handleUserCommand` never reads them. Either implement it (and restrict/confirm for project commands) or drop the parsing + doc claim.
- **[nit] `lsp_navigate` results are unbounded** — `formatResult` emits one line per location/symbol with no cap, unlike grep's `head_limit`. A `references` query on a hot symbol can produce a huge single result. Suggest an optional `limit` (~100 default) with a `… N more` note.

## Verdict
**REQUEST_CHANGES** — the `lsp_navigate` path-confinement bypass is a real arbitrary-file-read regression of a boundary the neighboring tools enforce. Once it's scoped through `resolveScopedReadPath` (and `absPath` is guarded), the rest are non-blocking. The prune/reconnect features are well-built and need no changes.

Vasanthdev2004:
Addressed the blocker — thanks, that was a real arbitrary-file-read I introduced.

`lsp_navigate` now carries the same `PathScope` confinement as its sibling read-only tools:
- Added `NewScopedLSPNavigateTool(workspaceRoot, scope)` (`NewLSPNavigateTool` delegates with nil scope), registered via the scoped constructor in `CoreReadOnlyToolsScoped`.
- The model-supplied path is resolved through `resolveScopedReadPath` **before** any read or LSP request, and the **resolved absolute path** is handed to the manager — so `Navigate`/`absPath` never opens an out-of-workspace file. Errors echo only the relative path, never an absolute one.

Regression test `TestLSPNavigateConfinesPathToWorkspace` asserts `/etc/passwd` and `../../` are rejected for both the position and `workspace_symbol` ops. **Verified by mutation** — bypassing the scope resolution makes the test fail (and the failure shows it attempting the out-of-workspace read). Commit `99a0ca1`.

On your non-blocking notes (`[repo]` provenance marker, unused `model:`/`agent:` frontmatter, unbounded `lsp_navigate` results) — all fair; I'll take those as fast follow-ups so this PR stays scoped to the security fix. Ready for re-review.

gnanam1990 (approved):
## Re-review — blocker resolved ✅

Commit `99a0ca1` ("confine lsp_navigate to the workspace") fully fixes the arbitrary-file-read blocker. Verified against the code and by running the tests:

- `lspNavigateTool` now carries a `PathScope`; `NewScopedLSPNavigateTool(workspaceRoot, scope)` is wired from `CoreReadOnlyToolsScoped` (registry.go:215), so it gets the same scope as `read_file`/`grep`/`glob`.
- The model-supplied path is resolved through `resolveScopedReadPath` **before** any read and **before** building the LSP request; `resolveWorkspacePath` rejects both absolute (`/etc/passwd`) and `..`-escaping paths (`filepath.Rel` + `..`/`IsAbs` check → "must stay inside the workspace").
- The LSP manager is handed the **resolved `absPath`** (not the raw input), so the `didOpen`-out-of-workspace vector is closed too.
- Errors echo only the relative path, never an absolute one.
- Regression test `TestLSPNavigateConfinesPathToWorkspace` exercises `/etc/passwd` + `../../` for both `definition` and `workspace_symbol`, asserts `StatusError`, and checks no absolute path leaks. I ran it locally: **PASS** (with the rest of the `lsp_navigate` suite). The commit notes it was verified by mutation.

I also re-confirmed `prune.go` (older-tool-output placeholder, keeps ToolCallID, preserves recent window, idempotent) and `reconnect.go` (connect-only retry, bounded backoff, excludes ctx/context-limit) are correct — no changes needed there.

## Verdict
**APPROVE.** The security boundary is restored to match the sibling read-only tools.

### Optional non-blocking follow-ups (unchanged from prior review)
- **[minor]** Project (`.zero/commands`) slash commands have no provenance marker in autocomplete/invocation — consider a `[repo]` tag and/or an inline notice (indirect prompt-injection surface; opt-in invocation keeps it minor).
- **[nit]** Frontmatter `model`/`agent` are parsed (`usercommands.go`) but never applied — implement or drop the field + doc claim.
- **[nit]** `lsp_navigate` results are unbounded — consider an optional `limit` (~100) with a `… N more` note, mirroring grep's `head_limit`.

PR #277, opened 2026-06-20
Fix custom provider model discovery

anandh8x:
## Summary
- Treat custom OpenAI/Anthropic endpoints as custom providers in the /model picker instead of falling back to official provider catalogs
- Discover Anthropic-compatible models via the provider /v1/models endpoint
- Allow manual /model <id> switches for custom providers when remote discovery cannot provide a catalog

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/providermodeldiscovery ./internal/providermodelcatalog ./internal/tui ./internal/config ./internal/providers
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added model discovery support for Anthropic-compatible providers.
  * Added support for custom OpenAI-compatible and Anthropic-compatible endpoints.
  * Model descriptions are now captured and displayed during discovery.
  * Manual model selection is now available for custom providers.

### DIFF
--- a/internal/providermodeldiscovery/discovery.go
+++ b/internal/providermodeldiscovery/discovery.go
@@ -18,6 +18,8 @@ import (
 	"github.com/Gitlawb/zero/internal/redaction"
 )
 
+const anthropicVersion = "2023-06-01"
+
 type Model struct {
 	ID               string
 	Description      string
@@ -40,7 +42,7 @@ type Options struct {
 
 func DiscoverCatalog(ctx context.Context, provider providercatalog.Descriptor, profile config.ProviderProfile, options Options) ([]Model, error) {
 	catalogModels, catalogErr := fetchCatalogModels(ctx, provider, options)
-	canProbeProvider := openAICompatibleDiscoveryAllowed(profile) && (!provider.RequiresAuth || discoveryHasCredential(profile))
+	canProbeProvider := modelDiscoveryAllowed(profile) && (!provider.RequiresAuth || discoveryHasCredential(profile))
 	if canProbeProvider {
 		liveModels, liveErr := Discover(ctx, profile, options)
 		if liveErr == nil {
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
Reviewed the diff and traced the branch code. Clean, focused, well-tested — custom OpenAI/Anthropic endpoints now get their own `custom-*` catalog in `/model` instead of falling back to official model lists, Anthropic-compatible discovery is wired correctly, and manual `/model <id>` works when a custom proxy has no remote catalog. **Approving.** No blockers — a few non-blocking nits below, mostly test-gaps and follow-ups.

### Verified-good
- **Manual-switch is correctly gated.** `resolveModelSwitchTarget` (`command_center.go:525`) only accepts an arbitrary id when `genericProviderCatalogID(provider.ID)` (the `custom-` prefix) — an *official* provider with an unknown id still falls through to "unknown model". No safety bypass.
- **Anthropic endpoint building is sound.** I traced `anthropicModelsEndpoint` across 9 base-URL shapes: trailing `/v1`, trailing slash, bare host, uppercase `/V1`, `/api/v1`, `/v1beta` — all correct, no double-`/v1`.
- **Header dispatch is right** — Anthropic uses `x-api-key` (no Bearer scheme) + `anthropic-version: 2023-06-01`; OpenAI keeps `Authorization: Bearer`. Refactor into `fetchProviderModels` preserves the OpenAI path.
- **Official-baseURL guard works** — `customProviderDescriptorForProfile` only returns a custom descriptor when `baseURL != official`, so an official endpoint still uses the official catalog.

### Non-blocking nits (follow-ups, not gating)
- **[P3, UX] Manual custom id has no validation feedback.** A typo'd `/model <id>` for a custom provider reports a bare "Switched model"; the failure only surfaces later as a runtime request error (`command_center.go:525`). Unavoidable when there's no catalog — but attaching a `notice` like "unverified model id (provider has no catalog)" would help. The plumbing exists (`modelSwitchTarget.notice`).
- **[P3, test-gap] Two boundary cases would be worth pinning** so a future refactor can't silently widen acceptance: (a) `ProviderKindOpenAICompatible` with `BaseURL == config.OpenAIBaseURL` + unknown id must still reject (proves the official-URL short-circuit) — the inverse of the existing custom test; (b) `anthropicModelsEndpoint` with a base URL already ending in `/v1` should yield `/v1/models`, not `/v1/v1/models` (the `/v1`-suffix branch is currently untested; only the `/anthropic` no-`/v1` case is covered).
- **[P3, test-gap] Description fallback untested** — `parseModelsResponse` does `display_name` then `name`, but no test asserts `Model.Description` or exercises the `name` fallback. A model carrying only `name` + asserting the resulting Description would lock the precedence.
- **[P3, edge — pre-existing, not introduced here] Registry resolution runs first** (`command_center.go:503`), so a custom id that matches an official regex/deprecated id is rewritten before the custom fallback. Almost never collides for real custom ids; only flagging because this PR makes custom providers first-class. Optional: prefer an exact live/catalog match over the fuzzy registry match for `custom-` providers.
- **[P3, style] The `ProviderKindOpenAI`/`ProviderKindAnthropic` cases in `customProviderDescriptorForProfile` are effectively unreachable for resolved profiles** (the resolver rewrites OpenAI+custom-baseURL to `*Compatible` and rejects Anthropic+custom-baseURL). Harmless safety net — a one-line comment noting that would clarify intent.

Solid PR. Nothing here blocks merge; the test-gaps are the most worthwhile follow-up.

gnanam1990 (approved):
## Overview
Extends provider model discovery to Anthropic-family providers: routes OpenAI and Anthropic kinds through a shared `fetchProviderModels` transport, adds an Anthropic `/v1/models` builder (`x-api-key` + `anthropic-version`, `display_name`/`name` capture), maps non-official Anthropic base URLs to the `custom-anthropic-compatible` catalog in the picker, and lets `/model` accept a manual model id for generic/custom providers.

## Strengths
- The merge + M11-fallback core in `DiscoverCatalog` stays provider-kind-agnostic (`mergeLiveModels` + empty-merge fall-through to the curated catalog), so the new Anthropic path reuses already-correct behavior instead of duplicating it.
- `fetchProviderModels` cleanly factors out the shared transport; OpenAI vs Anthropic differ only in endpoint + auth-header defaults + the `anthropic-version` hook.
- New tests cover the net-new Anthropic transport directly (path, `x-api-key`, `anthropic-version`, dedup/sort) and the custom-OpenAI picker resolution.

## Non-blocking notes
- **[nit] Anthropic live-merge path through `DiscoverCatalog` is untested** (`discovery_test.go`): the only `DiscoverCatalog` test drives `ProviderKindOpenAI`; the new Anthropic test calls `Discover()` directly. The merge/fallback code is shared (already covered via OpenAI), so this is a coverage gap, not a bug. Suggest a `DiscoverCatalog` test with `ProviderKindAnthropic(Compat)` asserting the M11 fall-through.
- **[nit] No picker test for the custom-anthropic-compatible resolution** (`picker.go`): `customProviderDescriptorForProfile` adds Anthropic-compat cases mapping to `custom-anthropic-compatible`, but the new picker test only exercises `ProviderKindOpenAICompatible`. Suggest mirroring it for `ProviderKindAnthropicCompat`.
- **[nit] Duplicated switch arms** in `customProviderDescriptorForProfile` (`picker.go:376-396`): the OpenAI/OpenAICompatible arms are byte-identical, as are the two Anthropic arms — collapse to two comma-separated `case` labels (the same PR already uses that form in `discovery.go`).

## Verdict
**APPROVE** — correct and well-factored; findings are non-blocking coverage/duplication nits worth a quick follow-up.

PR #278, opened 2026-06-20
Install Windows sandbox helper binaries

anandh8x:
## Summary
- Copy the Windows sandbox command runner and setup helper during PowerShell install
- Resolve required files from either the archive root or nested package directory
- Extend installer contract coverage so helper copying cannot regress

## Why
Windows sandboxed commands fail before execution when only zero.exe is installed because the native sandbox command runner cannot be found. Release packaging already includes these helper binaries; the installer was dropping them.

## Tests
- GOCACHE=/tmp/zero-go-cache go test ./internal/installtest -count=1
- GOCACHE=/tmp/zero-go-cache go test ./internal/release -count=1
- git diff --check

## Summary by CodeRabbit

* **Improvements**
  * Windows installer now deploys additional required executable components alongside the primary binary to ensure complete installation.
  * Enhanced file extraction validation to confirm each required component is located and installed correctly to the installation directory.

### DIFF
--- a/internal/installtest/install_scripts_test.go
+++ b/internal/installtest/install_scripts_test.go
@@ -44,8 +44,10 @@ func TestPowerShellInstallerScriptMatchesWindowsReleaseContracts(t *testing.T) {
 		`$checksumName = "$archiveName.sha256"`,
 		`Get-FileHash -Path $archivePath -Algorithm SHA256`,
 		`Expand-Archive -Path $archivePath -DestinationPath $extractDir -Force`,
-		`Get-ChildItem -Path $extractDir -Filter "zero.exe" -File -Recurse`,
-		`Copy-Item -Path $binaryPath -Destination $targetPath -Force`,
+		`Find-ZeroExtractedFile -Root $extractDir -FileName $fileName`,
+		`"zero-windows-command-runner.exe"`,
+		`"zero-windows-sandbox-setup.exe"`,
+		`Copy-Item -Path $sourcePath -Destination (Join-Path $InstallDir $fileName) -Force`,
 	})
 }
 
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -43,6 +43,25 @@ function Get-ZeroArch {
   }
 }
 
+function Find-ZeroExtractedFile {
+  param(
+    [string]$Root,
+    [string]$FileName
+  )
+
+  $candidate = Join-Path $Root $FileName
+  if (Test-Path $candidate -PathType Leaf) {
+    return $candidate
+  }
+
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (approved):
**Approving — correct fix, and it closes a real gap.** ✅

Root cause confirmed: `release.Package` already builds and bundles both helper binaries into the Windows archive (`internal/release/release.go:212-217` builds `./cmd/zero-windows-command-runner` and `./cmd/zero-windows-sandbox-setup` for `goos=="windows"`), but `install.ps1` was copying **only** `zero.exe` (`install.ps1:97` on main) and silently dropping the two helpers. So even though the fix shipped *in the archive*, installed users still hit `Windows sandbox helper is not available`. This PR copies all three binaries next to `zero.exe`, which the tier-1 "adjacent executable" discovery in `windows_runner.go` picks up. Exactly the right layer.

Verified:
- Merges cleanly into current `main`, CI green across all smoke targets.
- `Find-ZeroExtractedFile` is sound — root-first, then a recursive single-match lookup, and a loud throw if a required binary is missing rather than a silent partial install.

Two small notes, both **non-blocking**:

1. **Latent coupling.** The installer now hard-requires all three binaries in the target archive, so a `-Version v<old>` install of a release that predates helper-bundling would throw `Release archive did not contain exactly one zero-windows-command-runner.exe`. Fine for the first release (it bundles them) — just flagging it for older-archive installs.
2. **Nit:** `$matches` shadows PowerShell's automatic `$Matches` variable. It works here, but renaming to `$candidates`/`$found` avoids a surprise if regex `-match` logic ever lands nearby.

Neither needs to block the merge.

Vasanthdev2004:
@anandh8x heads-up — turns out we both landed on a fix for this same error from different angles, so I want to make sure they don't collide. 👍

There's a parallel PR, #280 (self-dispatch), that targets the same `Windows sandbox helper is not available` root cause. They're **complementary, not competing** (zero overlapping files):

- **Your #278** fixes the *installed* path — the real helper binaries land next to `zero.exe`, so tier-1 adjacent-exe discovery finds them. This is the gap where `install.ps1` was dropping the helpers the archive already shipped.
- **#280** makes the separate helper binaries *optional* via self-dispatch (`zero.exe` re-execs itself as the helper), so the sandbox also works under `go run` in dev and when only `zero.exe` is on `PATH` — cases the installer fix can't cover.

Plan: we're taking **both** — #280 as the core fix (full dev + prod coverage), and yours as the installer-side hardening so production installs get real tier-1 helpers. Thanks for catching the `install.ps1` gap — clean fix.

PR #280, opened 2026-06-20
fix(sandbox): make the Windows sandbox work in dev (self-dispatch helpers)

Vasanthdev2004:
-

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows sandbox setup now verifies Administrator/elevated permissions before applying firewall and workspace security rules, returning a clearer instruction instead of confusing access-denied errors.

* **New Features**
  * Improved Windows sandbox helper resolution and execution: if helper executables aren’t available where expected, the system can fall back to hidden self-dispatch to run the runner/setup helpers correctly.
  * Updated sandbox “doctor” checks to more accurately report Windows native backend availability under these fallback conditions.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -190,6 +190,21 @@ func defaultUserPluginsDir() string {
 }
 
 func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) (exitCode int) {
+	// Self-dispatch as the Windows sandbox helper. When no standalone helper .exe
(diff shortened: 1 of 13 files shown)


anandh8x (approved):
Reviewed the updated commit. The Windows self-dispatch helper path is the right fix for dev/plain-build Windows sandbox usage: it preserves adjacent helper and PATH discovery, adds a self-dispatch fallback, routes setup through the same helper resolution, and covers the behavior with focused tests. The commit metadata is now project-neutral as well.

PR #279, opened 2026-06-20
docs(audit): 2026-06-20 deep adversarial audit report

gnanam1990:
Adds `docs/audit/2026-06-20-deep-audit.md` — a deep, adversarial audit of `origin/main` (`bfbdbb1`). **Docs-only; no source changed.**

## Method
Fresh checkout of `origin/main`, Go-native gates, then a 13-subsystem deep read with **per-finding adversarial re-verification** (independent skeptics re-read guards/locks/callers to try to disprove each finding — only survivors are reported), plus reconciliation of the `2026-06-10`/`2026-06-13` audits against current code.

## Gates
- `go build ./...` ✅ clean · `go vet ./...` ✅ clean
- `go test ./... -race` → **0 data races** (67/68 pkgs). The 3 `internal/tools` failures are **`-race`-only timing flakes** (pass 3× without `-race`; the test's fixed 10ms yield window is too short under race instrumentation), not product defects.

## Result
**35 surviving findings:** 0 critical · **1 high** · **15 medium** · 17 low · 2 info. Prior audit reconciliation: **64 fixed · 14 partial · 12 still-open**.

Headline items:
- **M10 (security):** the provider health probe forwards `x-api-key`/`x-goog-api-key`/custom headers across a cross-host HTTP redirect (credential exfil to a public host) — not covered by the recent DNS-rebind hardening.
- **H1:** four of six documented, user-configurable hook events (`sessionStart`/`sessionEnd`/`specialistStart`/`specialistStop`) are validated and `zero hooks add`-able but never dispatched.
- Durability: events.jsonl append isn't `fsync`'d while derived metadata is (**M12**); PTY exec sessions can drop their final output chunk on exit (**M14**).
- Cron edges: DST fall-back double-fire (**M4**), claim double-grant on an unadvanceable spec (**M5**), `--run-now` silently cancelled by reconcile (**M6**).
- Lifecycle: remote-bridge idle deadline cleared (**M7**) + `Shutdown` can't close bridge conns (**M8**); MCP SSE never reconnects (**M9**).

The report contains evidence (`path:line` + quotes), per-finding impact + minimal fix, the full fixed/partial/still-open reconciliation tables, 4 root-cause clusters, a prioritized remediation order, and confidence notes. Several findings (M4/M5/M6/M11) were empirically reproduced.

Audit only — remediation is a separate follow-up if/when you want it.

## Summary by CodeRabbit

* **Documentation**
[...]

PR #281, opened 2026-06-20
fix: resolve 15 findings from the 2026-06-20 deep audit

gnanam1990:
Implements fixes for the 2026-06-20 deep audit (`docs/audit/2026-06-20-deep-audit.md`) in staged, one-concern-per-commit changes. Each finding was re-confirmed against current code before fixing; each stage gated `go build ./...` + `go vet` + `gofmt -l` + `-race` on the affected package, and the **full `go test ./... -race` is green on the final tree**. No new dependencies. This branch includes the audit report itself (with per-finding status markers in §8) and supersedes the report-only PR #279.

## Summary
15 findings fixed (1 medium security, 8 other medium, 1 info, 5 low) + the `-race` test-robustness gap the audit flagged. The riskiest/largest items (the lone High and several mediums/lows) are **deferred with rationale** rather than half-wired — see the table and the report §8.

## Findings

| ID | Sev | Status | Commit | Note |
|---|---|---|---|---|
| M10 | medium (security) | ✅ fixed | `e03ce9e` | strip `x-api-key`/`x-goog-api-key`/custom headers on cross-host redirect in the health probe |
| M14 | medium | ✅ fixed | `6f8dc54` | join the PTY copy goroutine before markDone so a command’s final output is not dropped on exit (also fixes the flaky TTY smoke test) |
| M12 | medium | ✅ fixed | `8151003` | `fsync` events.jsonl append (log as durable as its metadata) |
| M4 | medium | ✅ fixed | `6f901a3` | minute-aligned `sched.Next` input → DST fall-back collapse guard engages |
| M5 | medium | ✅ fixed | `6f901a3` | pause an unadvanceable job inside the claim → no double-fire |
| M7 | medium | ✅ fixed | `039caed` | bound the daemon control handshake with a read deadline |
| M8 | medium | ✅ fixed | `039caed` | `ServeConn` tracks the conn so `Shutdown` can close a stalled remote handshake |
| M13 | medium | ✅ fixed | `2421661` | atomic rename-with-verify mailbox stale-lock break (no self-compare) |
| M1 | medium | ✅ fixed | `1b5c06b` | anchor the streamjson `api_key` redaction so prose isn't mangled |
| I1 | info | ✅ fixed | `039caed` | local control-socket handshake now deadline-bounded (same fix as M7) |
| L15 | low | ✅ fixed | `bf8294b` | snapshot exec-session `lastUsedAt` under its lock — fixes a real `-race` data race |
| L13 | low | ✅ fixed | `e07c914` | `defer job.cancel()` releases the scheduler per-job context on MaxRuns |
| L1 | low | ✅ fixed | `46f885f` | route post-compaction / max-turns connects through `streamWithReconnect` |
| L3 | low | ✅ fixed | `71e611f` | `--skip-permissions-unsafe` rejects trailing positional args |
| L11 | low | ✅ fixed | `86a6b60` | error on a present-but-wrong-shape chatgpt account-id claim |
| §2 | — | ✅ fixed | `4304395` | de-flaked 3 exec-session timing tests so `go test ./... -race` is green |
| H1 | **high** | ⏸ deferred | — | wiring the 4 lifecycle hook events is a cross-surface design change (dispatcher lives only in `agent.Options`; session boundary differs across exec/TUI/daemon; specialist events need the parent dispatcher threaded through the child). Deleting them would disable a documented feature (guardrail). Needs a focused design pass. |
| M9 | medium | ⏸ deferred | — | MCP SSE reconnect: risky concurrency (re-open + re-`initialize`), not deterministically unit-testable |
| M11 | medium | ⏸ deferred | — | sandbox AST-authoritative classification is security-sensitive (risk of under-blocking a real destructive command); needs an extensive regression matrix before flipping precedence |
| M2, M3, M6, M15 | medium | ⏸ deferred | — | config MCP-merge source-of-truth decision; project-command provenance (UX); cron run-now schema field; TUI streaming memoize (perf-only) |
| L2,L4,L5,L6,L7,L8,L9,L10,L12,L14,L16,L17,I2 | low/info | ⏸ deferred | — | lower-value hygiene / cross-cutting (dead-code removal decisions, cross-platform lock identity, RFC3339Nano format, TUI width math, minor concurrency hardening) — batchable follow-up |
| §4.4 ×2 | — | ⏭ skipped | — | the two findings the audit itself withdrew (compactor frozen-budget; `maxTurns<=0`) — not real defects |

## Gate results (final tree, verbatim)
```
$ gofmt -l .
(clean)
$ go build ./...
build OK exit=0
$ go vet ./...
vet OK exit=0
$ go test ./... -race -count=1
ok  github.com/Gitlawb/zero/internal/agent  5.272s
... (all 70 packages: ok) ...
ok  github.com/Gitlawb/zero/internal/tools  14.880s
ok  github.com/Gitlawb/zero/internal/tui    7.251s
TEST_EXIT=0
```
(Run with `HOME`/`XDG_CONFIG_HOME` pointed at temp dirs + `CI=1`, per the existing suite's isolation. The repo's CI gate is `go test ./...` without `-race`; this branch additionally passes the stricter `-race ./...`.)

## Reviewer call-outs (riskiest changes)
- **Concurrency:** `daemon` (M7/M8 handshake deadline + `ServeConn` conn-tracking — verify the deadline is cleared before streaming and that local + remote paths both close on `Shutdown`); `swarm/mailbox` (M13 rename-with-verify — verify it can never delete a freshly-rotated lock); `tools` (L15 `lastUsedAt` snapshot — verify lock order stays `manager.mu → session.mu`); `cron` (M4/M5 — verify the unadvanceable job still fires exactly once then pauses, and the DST test's >90m assertion).
- **Security:** `providerhealth` (M10 — verify the header strip covers arbitrary `CustomHeaders` and that same-host redirects keep the credential).
- **Persistence:** `sessions` (M12 `fsync` — note: the fsync has no portable unit-test seam; covered by the existing append/durability suite).

No new dependencies were added.

## Release Notes

* **Bug Fixes**
  * Fixed cron scheduling around daylight saving time fall-back to prevent double-firing
  * Improved daemon control-connection shutdown by bounding the initial handshake and ensuring it unblocks on shutdown
  * Enhanced CLI `--skip-permissions-unsafe` validation to reject trailing prompt-like arguments
  * Improved streaming reliability by reconnect-aware re-streaming during retry/compaction phases
  * Hardened OAuth token claim handling to error on invalid ChatGPT account ID shapes
  * Strengthened event log persistence and improved PTY teardown to avoid truncating final output
* **New Features**
  * Added handshake timeout enforcement for daemon connections
* **Chores**
  * Refined API key secret redaction patterns


Vasanthdev2004 (approved):
Reviewed all 14 fixed findings (adversarially, across security / daemon+swarm concurrency / cron+tools clusters) plus the CI state. This is a careful, well-staged PR — one concern per commit, each re-confirmed against current code, with honest deferral rationale for the riskier items. **Approving.**

### CI: the ubuntu-smoke failure is a pre-existing flake, not this PR
`TestExecCommandTTYSessionAcceptsInputOnLinux` failed ("expected PTY input output, got Command completed with no output") — the classic PTY timing race. I verified it's **not** caused by this PR: the test is unchanged from `main`, and the PR's `exec_command.go` changes touch nothing in the PTY/tty/stdin path (they're entirely in the prune/eviction path for L15). Please re-run the job once the run completes; worth a follow-up to de-flake this one too (it wasn't in §2's scope).

### Findings — all correct & complete
**Security**
- **M10** (cross-host redirect credential strip) ✅ — **fail-safe**: strips on `req.URL.Host != via[0].URL.Host`, i.e. any host:port change; credentials are kept only on an exact-host match, so there is no under-strip/leak path. `sensitiveAuthHeaderNames` is comprehensive (well-known + kind defaults + `profile.AuthHeader` + all `CustomHeaders` keys). Same-host probe still works; tested both directions.
- **M1** (anchor `api_key` redaction) ✅ — the new regex requires `[=:]` + a 12+ char value, killing the prose false-positives without losing real-key coverage (sk-/bearer patterns are separate and unchanged).
- **L11** (chatgpt account-id claim shape) ✅ — correctly errors on present-but-missing/empty/wrong-type under the namespace while preserving the absent→top-level fallback; 3 shapes tested.

**Concurrency — daemon/swarm**
- **M7/M8/I1** (handshake read-deadline + `ServeConn` conn-tracking) ✅ — deadline is cleared (`SetDeadline(time.Time{})`) before the streaming phase so it can't kill a healthy stream; conn track/untrack is correct on every exit path (success/error/panic); `Shutdown` closes both local and remote in-flight handshakes under a clean lock; no use-after-close.
- **M13** (mailbox rename-with-verify stale-lock) ✅ — atomic rename means only one racer wins the break; the re-stat on the moved file restores a freshly-rotated lock instead of stealing it; no self-compare bug. Mirrors `daemon/lock.go`'s pattern.

**Cron / tools / misc**
- **M4** (DST minute-align) ✅ and **M5** (pause unadvanceable job inside the claim) ✅ — both are genuine cron **double-fire** bugs; M5's pause is inside the same `Mutate` critical section as the claim (atomic), and both have targeted regression tests.
- **L13** (`defer job.cancel()`) ✅ context leak on MaxRuns, idempotent. **L1** (route post-compaction / max-turns connects through `streamWithReconnect`) ✅ correct helper reuse, no double-forward. **M12** (`fsync` events.jsonl) ✅ durable before close, all error paths close the handle. **L3** (`--skip-permissions-unsafe` rejects trailing args) ✅ tested, `--add-dir` cases preserved.

### Non-blocking follow-ups (optional hardening)
- **M10**: the host compare is case-sensitive and doesn't guard a same-host **https→http scheme downgrade** (credentials would ride a cleartext redirect). Both are minor and **pre-existing** (main stripped nothing) — fail-safe over-strip at worst — but reusing a shared official-URL/scheme predicate would tighten it.
- **M1**: the 12-char minimum could miss an unusually short custom key; fine in practice, just noting.

The deferred set (H1 hook wiring, M9 MCP SSE reconnect, M11 sandbox AST precedence, M14 PTY drain, the low/info batch) is reasonably deferred with rationale rather than half-wired — good call. Nice, disciplined work.

PR #282, opened 2026-06-21
fix: resolve product/UX audit findings for OSS launch

gnanam1990:
## Summary

Implements the fixes from the product/UX audit (`docs/audit/2026-06-21-product-ux-audit.md`) for OSS launch: all in-scope **Critical** and **High** findings, the quick OSS-readiness files, the WCAG contrast issues, and the specifically-scoped mediums (`--theme` flag, NO_COLOR spec, missing-vs-rejected key). Staged as one coherent change per finding. **Do not merge** until the manual maintainer steps below are handled.

Gate on the final tree — all green:

```
go build ./...        → ok
go vet ./...          → ok
gofmt -l .            → clean (empty)
go test ./... -race   → ok  (68 packages ok, 0 FAIL, 4 no-test pkgs)
govulncheck ./...     → No vulnerabilities found
deadcode (advisory)   → only pre-existing hits; no new dead code
```

Every fix ships with a test that fails without the change (concurrency-relevant ones under `-race`). Each finding was re-opened in code and confirmed to still exist before fixing.

## Findings

| ID | Sev | Status | Note |
|----|-----|--------|------|
| C1 | 🔴 | Fixed (in-repo) | Demote the unpublished binary-install path to "coming soon" + lead with `go run`; release workflow now publishes a GitHub Release on a `v*` tag. Repo owner already matched `go.mod`. **Publishing = manual.** |
| C2 | 🔴 | **Skipped** | LICENSE choice deferred by the maintainer. Manual. |
| H1 | 🟠 | Fixed | Connectivity probe allows loopback only for a user-configured local provider base_url; redirects + other private ranges still blocked. |
| H2 | 🟠 | Fixed | README badge + quick start corrected to Go 1.25+. |
| H3 | 🟠 | Fixed | Removed the inert `/input-style` command (registry + kind + handler + dead helper). |
| H4 | 🟠 | Fixed | Removed the non-existent `/usage` from the README slash-command table. |
| H5/H6 | 🟠 | Fixed | Raised `faint`/`faintest` to WCAG AA in both themes (they carry line numbers, diff headers, help text). |
| H7 | 🟠 | Fixed | Curated, actionable 401/403 message; redaction only scrubs token-shaped words (no more mangled help text). |
| H8 | 🟠 | Fixed | `doctor` renders nested-map details for humans (no raw Go `map[...]`). |
| H9 | 🟠 | Fixed | `doctor` fails (not "Overall: pass") for a remote provider with no credential; keyless local providers stay a pass. |
| H10 | 🟠 | Fixed | No-provider / no-key errors point at `zero setup` / `zero auth`. |
| H11 | 🟠 | Fixed | Added `SECURITY.md` with a private vulnerability-reporting path. |
| M1 | 🟡 | Fixed | `setup --verify` distinguishes "no API key found" from "rejected" (env-var keys honored). |
| M2 | 🟡 | Fixed | Light-theme accent raised to AA; theme test now asserts a true WCAG ratio. |
| M3 | 🟡 | Fixed | `NO_COLOR` honored for any non-empty value per no-color.org (not just ParseBool). |
| M4 | 🟡 | Fixed | Real `--theme {auto\|dark\|light}` flag wired to the TUI. |
| M9/M10/M11 | 🟡 | Fixed* | Issue/PR templates, `CODE_OF_CONDUCT.md`, `CHANGELOG.md`. *M11: build-time version stamping deferred. |
| L2 | ⚪ | Fixed | README "Accessibility & appearance" section documenting the appearance/a11y controls. |

## Manual maintainer steps (NOT done — required before/at merge)

1. **LICENSE (C2)** — add `LICENSE` + README §License + `package.json` `license` once the license is chosen.
2. **Publish a release (C1)** — push a `v*` tag (and make the repo public) so the install scripts resolve. The workflow now creates the Release + archive/`.sha256` assets on tag, **but the workflow change is untestable locally — validate it on the first tag.** This PR does not tag or publish.
3. **README screenshot/GIF** — add a TUI screenshot/recording (not auto-generated).
4. **Code-of-conduct contact** — currently routes to the private security-advisory channel; substitute a dedicated address if preferred.

## Deferred (with rationale)

- **L1** Ctrl+E composer binding, **L3** brand-caps + real version string, **L5** npm package name, **M5** wide-rune truncation budget (frame already protected per audit §I4) — Low / nice-to-have.
- **M6** `--reasoning-effort` validation, **M8** discovery slash-commands — Mediums outside this PR's scope (crit/high + OSS files + contrast).
- **M7** per-class recovery hints — largely addressed by H7 (auth errors now name a recovery action).
- **M11** build-time version stamping — CHANGELOG added; stamping deferred.

## New dependencies

**None.** Go-native only; the release-publish step uses the preinstalled `gh` (no new GitHub Action).

## Reviewer call-outs (riskiest changes)

- **H1 — SSRF relaxation** (`internal/providerhealth/providerhealth.go`): loopback is permitted **only** when the user's own configured base_url is loopback (`endpointIsLoopback`), threaded into both the pre-flight validation and the dial path; `CheckRedirect` stays strict so a 3xx can never reach loopback, and only loopback is relaxed (metadata/link-local/other private ranges remain blocked). The fetch-URL SSRF guard is untouched. Test proves a local provider passes while remote SSRF + loopback-redirects are still blocked.
- **H7 — auth-error / redaction** (`internal/providers/providerio/providerio.go`): redaction now preserves the word "Bearer" and scrubs only the token-shaped value after it (so upstream help text survives), and 401/403 lead with a curated message plus a redacted one-line detail. Worth confirming the token-shape heuristic (`looksLikeToken`) cannot be tricked into leaking a real key.

## Summary by CodeRabbit

* **New Features**
  * Added `--theme {auto|dark|light}` flag to control TUI color mode.
  * Improved `NO_COLOR` handling to disable colors when the value is any non-empty string.
* **Bug Fixes**
  * Refined `/doctor` to correctly handle missing vs. rejected credentials and improved human-readable output formatting.
  * Improved auth/setup guidance and redaction for safer, more actionable error messages.
  * Hardened provider connectivity checks around loopback and redirect behavior.
  * Removed the `/input-style` command from the TUI.
* **Documentation**
  * Added issue/PR templates, `SECURITY.md`, and updated README/INSTALL and CHANGELOG entries.
* **Chores**
  * Updated release artifact workflow to publish GitHub releases on tags.


Vasanthdev2004 (approved):
## Review: PR #282 — product/UX audit findings for OSS launch

Reviewed the highest-risk changes (H1 SSRF, H7 redaction, H9 doctor, M3 NO_COLOR, M4 --theme, WCAG contrast) against the diff and the current tree. Overall this is a well-executed audit-response PR: the security-sensitive relaxations are narrowly scoped, the tests verify the real behavior changes, and the contrast fixes are backed by real WCAG ratio assertions. Approving with a few non-blocking notes.

### Verdict: APPROVE (with P2/P3 notes below)

| Severity | Count |
|---|---|
| P1 (blocking) | 0 |
| P2 (should fix) | 1 |
| P3 (nice to have) | 3 |

---

### P2 — `--theme` before `--skip-permissions-unsafe` is silently dropped

`internal/cli/app.go` (PR diff lines ~867, ~889, ~906)

The top-level `splitLeadingThemeFlag(args)` strips a leading `--theme` into the `theme` variable. But the `--skip-permissions-unsafe` branch launches with `skipTheme` (parsed from `rest` = `args[1:]`), **not** the top-level `theme`. So:

```
zero --theme light --skip-permissions-unsafe   # theme="light" stripped at top; unsafe branch uses skipTheme="" → light is dropped
zero --skip-permissions-unsafe --theme light   # works: skipTheme="light"
```

The `TestRunThemeFlagPopulatesTUIOptions` test only covers the no-subcommand path (`len(args)==0`), so this ordering gap is untested. Fix: initialize `skipTheme = theme` at the top of the unsafe branch before the second `splitLeadingThemeFlag` call (so a `--theme` after the flag can still override, but one before isn't lost). Low blast radius (`--skip-permissions-unsafe` is an expert escape hatch), hence P2 not P1.

### P3 — `looksLikeToken` heuristic is defense-in-depth, not the primary guard (good, worth a comment)

`internal/providers/providerio/providerio.go` (PR diff ~1730)

The new `looksLikeToken` requires `>=16` chars in `[A-Za-z0-9._-]` plus (if `<24`) a digit. A short bearer token (e.g. a 12-char value) after the literal word "Bearer" would NOT be redacted by this heuristic alone. This is acceptable because `Redact` still does `strings.ReplaceAll(message, secret, "[REDACTED]")` for every explicitly-passed `secrets...` first (`providerio.go:236-239`), so real keys threaded through `ClassifiedError`/`Redact` are always scrubbed regardless of the Bearer heuristic. The heuristic only guards against a secret that appears in an upstream message but wasn't passed as an explicit `secrets` arg. Worth a one-line comment clarifying that explicit-secrets is the primary path so a future reader doesn't tighten the regex unnecessarily.

### P3 — WCAG margins are thin but pass

`internal/tui/theme.go` (PR diff ~2027)

Verified the new tokens against the actual panel backgrounds (`#0e0e10` dark, `#ececed` light):

| Token | Ratio | AA (≥4.5)? |
|---|---|---|
| dark faint `#838389` on `#0e0e10` | 5.12 | ✅ |
| dark faintest `#7c7c82` on `#0e0e10` | 4.65 | ✅ |
| dark muted `#8b8b93` on `#0e0e10` | 5.70 | ✅ |
| dark accent `#caff3f` on `#0e0e10` | 16.43 | ✅ |
| light muted `#54545b` on `#ececed` | 6.36 | ✅ |
| light faint `#5b5b62` on `#ececed` | 5.71 | ✅ |
| light faintest `#646469` on `#ececed` | 4.98 | ✅ (thin) |
| light accent `#477006` on `#ececed` | 4.96 | ✅ (thin) |

`light accent` (4.96) and `light faintest` (4.98) clear AA but with ~0.1–0.5 of headroom. The test correctly asserts `>=4.5` against `*Palette.panel`, which is the right worst-case surface. Note `faintest` also renders against the unpainted terminal canvas in some paths (diff gutters) where the background could differ from `panel` — if a future terminal reports a very dark or very light default BG alongside `--theme light`, the effective ratio could shift. Not blocking; just flagging the thin margins so a future contrast bump doesn't surprise anyone.

### P3 — `docs/audit/2026-06-21-product-ux-audit.md` committed

The 444-line audit document is committed under `docs/audit/`. This looks intentional (it's the source-of-truth for the PR's findings and each fix references it by ID). Reasonable to keep as a historical record; just confirming it wasn't an accidental scratch file. If the team prefers not to ship internal audit docs to consumers, consider moving it to `docs/audit/internal/` or `.github/` — but this is a taste call, not a defect.

---

### Things that are correct and worth calling out

- **H1 SSRF relaxation is tight.** `allowLoopback` is derived solely from the user's own configured `base_url` via `endpointIsLoopback`; the redirect `CheckRedirect` always calls `validateEndpoint(..., false)` so a 3xx to loopback is rejected even for a local-provider probe (`TestConnectivityClientLoopbackRedirectStillBlocked` verifies this). Other private ranges (cloud metadata `169.254.169.254`, link-local, RFC1918) stay blocked even with `allowLoopback=true` (`TestValidateEndpointAllowsLoopbackOnlyForLocalProvider`). The `safeDialContext` TOCTOU-dial hardening is preserved with the same flag plumbing.
- **H9 doctor aggregation is correct.** `Run` iterates all checks and flips `report.OK = false` on any `StatusFail` (`doctor.go:82-86`); the new `providerConfigCheck` returns `StatusFail` for a credential-less remote provider, so "Overall: pass" no longer lies. Keyless loopback providers correctly stay `StatusPass` via `localProviderBaseURL`. `TestDoctorFailsRemoteProviderWithoutCredential` covers both arms.
- **H8 `formatDetailValue`** properly renders nested maps as sorted `k: v` and slices as comma lists, killing the `map[...]` leak. `[]byte` is correctly special-cased.
- **M3 NO_COLOR** is spec-compliant: `noColorRequested` returns `getenv("NO_COLOR") != ""`, forcing `colorprofile.Ascii`. `TestNoColorRequestedHonorsSpec` covers `"yes"`, `"foo"`, `"0"`, `" "` — all the cases the dependency's `strconv.ParseBool` gate missed.
- **M1 setup verify** correctly short-circuits the probe when no credential is configured (so the user sees "no API key found" instead of "the provider rejected the API key"), and still probes keyless local providers. `profileHasCredential` correctly checks `APIKeyEnv`-populated env vars so env-var-based providers aren't falsely flagged.
- **Release workflow** adds `permissions: contents: write` and uses the preinstalled `gh` (no new action dependency) to publish release assets on `v*` tag pushes; `--clobber` handles the matrix race. The README/INSTALL.md "coming soon" demotion is consistent with this.
- **`/input-style` removal** is clean: command kind, registry entry, handler case, and `shellOnlyCommandText` helper all removed; `TestInputStyleCommandRemoved` asserts it's gone from the registry, parser, and `/help`.

### No regressions spotted

The existing connectivity tests were updated to pass the new `allowLoopback` parameter (`false` for the redirect/private-address cases, preserving their original intent). The doctor/TUI tests that didn't care about the credential check were updated to set `APIKey: "sk-test"` to isolate their original render/pass-through behavior from the new no-key failure — appropriate test scoping.

PR #284, opened 2026-06-21
feat(npm): publish @gitlawb/zero with a postinstall binary installer

Vasanthdev2004:
## What

Makes `npm install -g @gitlawb/zero` deliver a working CLI. `bin/zero.js` already execs an adjacent binary, but nothing fetched it, and the package name `zero` is taken on npm. This wires up the missing delivery + publishing path.

## How

- **`package.json`** → rename to **`@gitlawb/zero`** (scoped; `zero` is taken). Add a `postinstall` hook, `files`, `engines` (`node>=18`), `os`/`cpu`, `repository`, and a `license` placeholder (the LICENSE file is still pending — separate task).
- **`scripts/postinstall.mjs`** (new) → detect platform/arch (mapping mirrors `internal/release/release.go`: `darwin→macos`, `amd64→x64`), download `zero-v{version}-{os}-{arch}.{tar.gz|zip}` from the matching GitHub Release, **verify SHA-256**, and extract into place. Extraction unpacks to a temp dir and copies **only known binary basenames** into the package — archive paths are never honored, so a crafted archive can't escape (no zip-slip). Idempotent via a version marker. `ZERO_SKIP_DOWNLOAD` and unsupported platforms are non-fatal skips.
- **`release-artifacts.yml`** → on a tag push, assert `git tag == v{package.json version}` before packaging. `zero-release` names assets from `package.json` while the release publishes under the tag, so any drift would 404 *every* install (install.sh, install.ps1, and npm). This also closes the version-drift follow-up flagged in the #282 review.
- **README / smoke doc** → document the `npm i -g @gitlawb/zero` path.

## Security

The installer downloads + verifies + stages an executable, so it went through an adversarial review (4 lenses, each finding independently verified). Path-traversal and command-injection were confirmed **genuinely closed** (temp-extract + copy-by-basename; argv arrays, no shell; correct PowerShell quote-escaping). Folded in the review's two `shouldFix` items + cheap nice-to-haves:

- **Require https** for the download origin — the same-origin `.sha256` makes TLS the load-bearing integrity control. `ZERO_ALLOW_INSECURE_DOWNLOAD=1` opts out for local testing; an `https→http` redirect is refused.
- **Tag/version drift-guard** (above) so a forgotten bump can't publish un-installable assets.
- Anchored, filename-validated checksum parsing; download size cap; missing-helper warning.

## Verification

- `go test ./internal/npmwrapper ./internal/release` ✅; gofmt/vet/build clean; `node --check` clean.
- New node-driven tests: per-platform asset plan, skip env, unsupported-platform.
- End-to-end against a local mirror: `download → verify → extract` places the binary, **a tampered checksum is rejected (exit 1, no binary)**, and **http is refused without the opt-in**.

## Going live — maintainer steps (not in this PR)

1. Create the `gitlawb` npm org (or publish under a personal scope) and add an `NPM_TOKEN` secret.
2. Publish the first `v0.1.0` GitHub Release (so there's an asset to download), then `npm publish --access public`.

A follow-up can add the `npm publish` step to the release workflow once the token exists.

## Summary by CodeRabbit

* **New Features**
  * npm package installation support via `@gitlawb/zero` with automatic native binary download on install

* **Documentation**
  * Updated installation guidance for npm package wrapper approach
  * Added postinstall behavior verification checklist

* **Chores**
  * Enhanced release workflow with version consistency validation before artifact packaging
  * Expanded test coverage for package metadata validation and installation behavior

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -45,6 +45,21 @@ jobs:
       - name: Test
         run: go test ./...
 
+      # zero-release names assets zero-v<package.json version>-… while the release
+      # is published under the pushed git tag. If they disagree, install.sh/install.ps1
+      # and the npm postinstall all build a download URL that 404s. Fail fast so a
+      # forgotten version bump can never publish mis-versioned, un-installable assets.
+      - name: Check tag matches package.json version
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') }}
+        shell: bash
+        run: |
+          pkg_version="$(node -p "require('./package.json').version")"
+          if [ "v$pkg_version" != "$GITHUB_REF_NAME" ]; then
+            echo "::error::tag $GITHUB_REF_NAME does not match package.json version v$pkg_version — bump package.json before tagging so the release assets and the npm package agree." >&2
+            exit 1
+          fi
+          echo "tag $GITHUB_REF_NAME matches package.json version v$pkg_version."
+
       - name: Package release artifact
         run: go run ./cmd/zero-release package
 
--- a/README.md
+++ b/README.md
@@ -53,9 +53,17 @@ zero exec -o stream-json < turns.jsonl        # programmatic, for scripts & CI
 go run ./cmd/zero
 ```
 
-> **Pre-built binaries are coming soon.** Once the first GitHub release is published, you'll be able
-> to install with `scripts/install.sh` (Linux/macOS) or `scripts/install.ps1` (Windows); those
-> scripts download release assets that don't exist yet. Until then, build from source with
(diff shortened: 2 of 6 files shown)


anandh8x (approved):
Reviewed #284 at e50dce2. No blocking findings. The postinstall plan matches release artifact naming, the wrapper looks for the installed binary in the same package-root location the installer writes, checksum/HTTPS/extraction hardening is in place, and npm pack includes the runtime files. Verification: go test ./internal/npmwrapper ./internal/release passed; node --check scripts/postinstall.mjs passed; dry-run plans for supported and unsupported platforms behaved as expected; npm_config_cache=/tmp/zero-npm-cache npm pack --dry-run --json included bin/zero.js and scripts/postinstall.mjs.

PR #285, opened 2026-06-21
fix(providers): global, configurable stream idle timeout (fixes 90s false-trip)

Vasanthdev2004:
## The problem

A long generation could die mid-stream with:

```
provider stream error: idle timeout after 1m30s (upstream stopped sending data)
```

The idle watchdog was hardcoded to **90s**, duplicated as `defaultStreamIdleTimeout` in **three** providers (anthropic, gemini, openai; codex inherits openai), and **wired to no config**. Slow cloud/reasoning backends (e.g. an Ollama-Cloud → MiniMax model) that go genuinely silent for >90s without emitting SSE keep-alives were killed mid-generation, failing the whole task.

## The fix — one global, configurable, less-aggressive timeout

Centralized in `providerio`:

- **`providerio.DefaultStreamIdleTimeout`** — single source of truth, raised **90s → 5m**. The watchdog still resets on SSE keep-alive comments (that logic is unchanged), so a *heartbeating-but-slow* upstream is never aborted — this only widens how long **true silence** is tolerated before aborting a hung stream.
- **`providerio.ResolveStreamIdleTimeout(option)`** — precedence: explicit option (e.g. a test) > **`ZERO_STREAM_IDLE_TIMEOUT`** env > default. The env accepts a Go duration (`10m`, `300s`) or bare seconds (`600`); `0`/`off`/`none`/`disabled` turns the watchdog off. A typo falls back to the default rather than silently disabling it.

All four providers resolve through it, so **one env var tunes every provider at once**.

```bash
ZERO_STREAM_IDLE_TIMEOUT=10m zero      # more headroom for a very slow backend
ZERO_STREAM_IDLE_TIMEOUT=off zero      # disable the watchdog entirely
```

## Why 5 minutes

The watchdog exists to stop a truly hung connection from blocking forever. With keep-alive resets already in place, it only ever fires on *complete* silence — so the threshold should be "this stream is really dead," not "this model is thinking." 90s was tripping healthy long generations; 5m is the floor a real stall must cross, and it's now tunable for anything slower.

## Notes
- **No behavior change for healthy streams** — keep-alive heartbeats still reset the timer; only the silent-stall threshold and its configurability change.
- Removes the three duplicated constants (net −duplication).
- Existing idle tests keep passing (they set an explicit short timeout, which still wins). Adds resolver tests (precedence, env parsing, disable, typo→default) and documents the env var in the README.

Verified: `go test ./internal/providers/... ./internal/agent/...` green; gofmt/vet/build clean.

## Release Notes

* **Documentation**
  * Updated documentation to clarify the streaming watchdog behavior for provider connections, including default 5-minute idle timeout and environment variable configuration options.
  * Added support for `ZERO_STREAM_IDLE_TIMEOUT` environment variable to customize or disable the idle timeout (accepts duration formats like `5m`, `300s`, or disable with `0`/`off`/`disabled`).

* **Tests**
  * Added test coverage for stream idle timeout resolution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,8 @@ Bring your own key — or no key at all for local runtimes.
 
 The model registry tracks each model's capabilities, context window, and cost — and the live model picker discovers what your provider actually serves.
 
+**Slow or stalling streams.** Every provider aborts a stream that goes completely silent (no data and no keep-alive) for `ZERO_STREAM_IDLE_TIMEOUT` (default **5 minutes**) so a hung connection can't block the agent forever. SSE keep-alives reset the timer, so a heartbeating-but-slow backend is never killed. If a slow cloud or reasoning model still trips it, raise the limit — `ZERO_STREAM_IDLE_TIMEOUT=10m` (accepts a Go duration or bare seconds; `0`/`off` disables the watchdog).
+
 ## Tools
 
 | Tool | Purpose | Side effect |
--- a/internal/providers/anthropic/provider.go
+++ b/internal/providers/anthropic/provider.go
@@ -68,11 +68,6 @@ func resolveThinking(effort string, maxTokens int) (budget int, effectiveMax int
 	return budget, effectiveMax, true
 }
 
-// defaultStreamIdleTimeout aborts a streaming read when the upstream goes silent
-// without closing the connection, so a stalled-but-open upstream cannot hang the
-// agent forever.
-const defaultStreamIdleTimeout = 90 * time.Second
-
 // Options configures an Anthropic Messages API provider.
 type Options struct {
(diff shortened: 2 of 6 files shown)


anandh8x (approved):
Reviewed #285 at 2bdf292. No blocking findings. The providers now share providerio.ResolveStreamIdleTimeout, explicit per-provider test options still win, ZERO_STREAM_IDLE_TIMEOUT parsing is covered, and the default is documented. Verification: go test ./internal/providers/providerio -run TestResolveStreamIdleTimeout passed locally; broader provider packages require local httptest listeners blocked by this sandbox, and PR CI is green across supported OSes.

PR #287, opened 2026-06-21
feat(tui): easier-to-scan tool results + visible selection bands

Vasanthdev2004:
First batch of UI tuning from a reference-TUI comparison. These are the "easier-to-scan results" group — high impact, low risk, token/glyph-level polish, no architecture change.

## 1. Status reads at a glance — glyph to the front
The ✓/✗/spinner that tells you if a step worked moved from the far-right edge to the **front** of the row, and the tool name is now colored by state.

```
BEFORE:  │ bash  mkdir -p site                    ✓     (mark hidden far right)
AFTER:   │ ✓ bash  mkdir -p site                        (✓ + name in green, up front)
         │ ✗ edit_file ...                              (red on failure)
```
Running keeps the normal name with the accent spinner (so orphaned/dead cards don't look live). A common pattern: leading every row with a colored glyph.

## 2. The selected menu row is actually visible now
`selBg` sat almost on top of the panel, so the highlighted row barely separated — **worst on the light theme, where it was effectively invisible (1.01:1 vs the panel).**

- dark `#1d2114 → #32401b` (separation 1.18 → 1.73)
- light `#e7f2cd → #cfe78f` (1.01 → 1.15)

Label contrast stays **>9:1 dark / >12:1 light**. It's one token, so `/model`, autocomplete, and the model picker all brighten together. Values picked by computing WCAG ratios, not by eye.

## 3. The specialist drill-in is no longer a hidden feature
A specialist card opens its subchat on click/Enter but advertised nothing — added a faint `· enter to open`.

## Tests
- Lock the glyph-leads-head layout so it can't silently regress.
- New contrast guard: the selected band must both **separate from the panel (≥1.10)** and keep its **label AA-readable (≥4.5)** on both themes — this test would have caught the invisible light band.
- Full `internal/tui` suite green; gofmt/vet clean.

Note: I can't see the rendered UI (no vision), so colors were chosen by WCAG computation + layout was confirmed by rendering the card through the real renderer. Worth a quick `go run ./cmd/zero` look — happy to tune the exact shades or push the selection to a full inverted bar if you want more pop.

## Summary by CodeRabbit

* **New Features**
  * Added "enter to open" hint in specialist cards to help discover drill-in navigation.

* **Style**
  * Repositioned status indicators to lead card headers instead of trailing edge.
  * Enhanced selected-row background colors for improved visibility and contrast.

### DIFF
--- a/internal/tui/rendering.go
+++ b/internal/tui/rendering.go
@@ -1178,7 +1178,9 @@ func (m model) renderRunningToolCard(row transcriptRow, width int, rc rowContext
 	if arg == "" {
 		arg = rc.args[rcKey(row.runID, row.id)]
 	}
-	head := toolCardHead(toolRowName(row), hint, arg, "", glyph, rc.auto[rcKey(row.runID, row.id)], width, opts)
+	// Running cards keep the normal name color; the accent spinner glyph at the
+	// front already marks them live (and orphaned dead cards must not look active).
+	head := toolCardHead(toolRowName(row), hint, arg, "", zeroTheme.ink, rc.auto[rcKey(row.runID, row.id)], width, opts)
 	return toolCard(head, glyph, nil, "", zeroTheme.cardRun, width)
 }
 
@@ -1186,9 +1188,11 @@ func renderToolResultCard(row transcriptRow, width int, rc rowContext, opts card
 	name := toolRowName(row)
 	failed := row.status == tools.StatusError
 	glyph := zeroTheme.green.Render("✓")
+	nameStyle := zeroTheme.green
 	borderStyle := zeroTheme.line
 	if failed {
 		glyph = zeroTheme.red.Render("✗")
+		nameStyle = zeroTheme.red
 		borderStyle = zeroTheme.cardErr
 	}
 	key := rcKey(row.runID, row.id)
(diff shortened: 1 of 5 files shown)


gnanam1990 (approved):
## Approve ✅

Well-executed, low-risk legibility polish. Verified locally: `gofmt -l` clean, `go vet ./internal/tui/` clean, full `go test ./internal/tui/` green, and the two new tests pass.

I independently recomputed every WCAG ratio in the description and they match to three decimals (dark selBg separation 1.18→1.73, light 1.01→1.15, ink-on-selBg 9.44 dark / 12.68 light) — and confirmed the new tests are genuine regression guards: the glyph-leads-head assertion fails on the old trailing-glyph layout, and the contrast test encodes the exact invisible-light-band bug it fixes. House rules respected (hex literals only in `theme.go` palettes; tests next to source).

### Minor, non-blocking (happy to merge as-is):
- **Stale doc comments** — `rendering.go:1308-1310` and `:1342` still say the status glyph is "right-aligned to the card edge", but it now *leads* the head row. Also `rendering_lime_test.go:1254` ("right-aligned"). Worth a one-line fix so future readers aren't misled.
- nit: the new contrast test guards the **ink label** on `selBg` but not the **faint description text** (`selectable_list.go:64`, light ratio ~4.97 — just above AA). Optionally add a `faint-on-selBg ≥ 4.5` check so a later selBg tweak can't silently drop it.
- nit: `· enter to open` is appended for **running** specialist cards too — consider gating on a non-empty child session id if running cards routinely lack one.

PR #288, opened 2026-06-21
feat(acp): Agent Client Protocol surface — drive ZERO as an editor backend

gnanam1990:
## Summary

Adds an **Agent Client Protocol (ACP)** surface so ZERO can run as an **editor backend**: editors that speak ACP (Zed, JetBrains, Neovim, …) spawn `zero acp` and drive ZERO's agent over JSON-RPC 2.0 on stdio. This is one more surface on ZERO's surface-agnostic core — alongside the TUI, headless `exec`, the MCP server, and cron — reusing the same `agent.Run`, tool registry, permission model, and session store. **Provider, model, and API keys stay owned by ZERO (BYOK); the editor only hosts the conversation thread.**

## Provenance

Derived **solely from the public, Apache-licensed ACP specification** (agentclientprotocol.com schema/v1 and the `agent-client-protocol` repository). Method names and message shapes come from the spec; all logic is written originally against ZERO's own interfaces. **No third-party agent's ACP implementation was read, copied, or paraphrased.** No new dependencies — standard library only.

## What's implemented (`internal/acp/` + `zero acp`)

- **Transport** — a bidirectional JSON-RPC 2.0 peer over newline-delimited JSON. It serves inbound requests/notifications and issues outbound calls (needed for the client-directed `session/request_permission`). Handlers run concurrently so a long-running `session/prompt` never blocks `session/cancel` or a permission response; in-flight handlers are drained on stream close so piped/finite input never drops a reply.
- **Methods** — `initialize` (capability + protocol-version negotiation), `session/new`, `session/load` (replays stored history so resume has context), `session/prompt` (drives a real `agent.Run` turn), `session/cancel`, `session/set_mode` (permission/autonomy mode), model selection via `session/set_config_option` (plus a vendor `_zero/set_model` alias), and the agent→client `session/request_permission`.
- **Event streaming** — ZERO's agent events are translated to `session/update` notifications: assistant text → `agent_message_chunk`, reasoning → `agent_thought_chunk`, tool calls → `tool_call` / `tool_call_update` (with content and changed-file locations), and plan items → `plan`.
- **Permission mapping** — ZERO's permission prompts map to ACP `PermissionOption`s (the option id carries the ZERO decision verbatim for a lossless round trip); the client's outcome maps back to a ZERO decision and **fails closed to deny** on anything unrecognized.
- **Capabilities** — advertises only what ZERO supports: `loadSession`, `promptCapabilities.image`, and session resume/load. ZERO executes tools locally, so client `fs/*` / `terminal/*` delegation is intentionally not required; clients that offer them still interoperate. ZERO-specific extras are vendor-prefixed (`_zero/…`) so clients that don't support them degrade cleanly.
- **`zero acp` subcommand** — registered in the CLI dispatch and help. It reuses config resolution, provider construction, the core tool registry, and the session store, so behavior matches the other surfaces.

## Commits (staged, one concern each)

1. JSON-RPC 2.0 ndjson transport
2. ACP wire types (from the public spec)
3. event → `session/update` translation
4. permission prompt ↔ `session/request_permission` mapping
5. the ACP Agent server driving `agent.Run`
6. the `zero acp` subcommand
7. drain in-flight handlers on stream close

## Test plan

- `go build ./...`, `go vet ./...`, **`go test ./... -race` — all green** (adds the `internal/acp` package).
- Unit tests: transport (request/response, notifications, errors, and a mid-handler callback that mirrors `session/prompt` → `session/request_permission`), event→update translation, and permission mapping.
- **Integration test**: drives `initialize` + `session/new` + `session/prompt` over in-memory pipes through the **real `agent.Run`** with a fake provider and asserts the streamed updates and stop reason.
- Verified against the built binary: the handshake and a full streamed turn — `tool_call` → `tool_call_update` → streamed answer → `stopReason: end_turn`.

## Follow-ups (not in this PR)

- **Submit ZERO to the ACP Registry** so it's installable as an agent in Zed / JetBrains (manual step).
- Advertise the user's *configured* providers' models in the model picker (currently it lists the model-registry defaults).
- Optional client-hosted `fs/*` + `terminal/*` for editors that prefer to own file access and the terminal.

Do not merge — opening for review.

## Release Notes

* **New Features**
  * Added an `acp` CLI command to serve Agent Client Protocol over stdio for editor backends.
  * Introduced a JSON-RPC transport with request/notification handling and robust stream parsing.
  * Implemented end-to-end ACP agent session flow: initialize, create/load sessions, run prompt turns, stream message/thought updates, and handle cancellation.
  * Added permission request forwarding with fail-closed option handling, plus per-session mode/model selection.
  * Added workspace `cwd` validation to reject overly broad targets.

* **Tests**
  * Added deterministic in-memory end-to-end and protocol/permission/unit tests.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/Gitlawb/zero/internal/agent"
 	"github.com/Gitlawb/zero/internal/config"
-	"github.com/Gitlawb/zero/internal/modelregistry"
+	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/sessions"
 	"github.com/Gitlawb/zero/internal/tools"
 	"github.com/Gitlawb/zero/internal/zeroruntime"
@@ -25,10 +25,15 @@ type Deps struct {
 	ResolveConfig func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
 	NewProvider   func(profile config.ProviderProfile) (zeroruntime.Provider, error)
 	RunAgent      func(ctx context.Context, prompt string, provider zeroruntime.Provider, opts agent.Options) (agent.Result, error)
-	BuildRegistry func(workspaceRoot string) *tools.Registry
-	Store         *sessions.Store
-	Models        modelregistry.Registry
-	AgentInfo     Implementation
+	// BuildWorkspace builds the SCOPED tool registry and the sandbox engine for a
+	// validated workspace root, so ACP shell tools (bash/exec_command) are confined
+	// exactly like the exec surface — never run unconfined on the host.
+	BuildWorkspace func(workspaceRoot string, resolved config.ResolvedConfig) (*tools.Registry, *sandbox.Engine, error)
+	// ResolveWorkspaceRoot validates + normalizes a client-supplied cwd (must be an
+	// existing directory; never the bare root). It is the file-tool confinement root.
+	ResolveWorkspaceRoot func(cwd string) (string, error)
+	Store                *sessions.Store
+	AgentInfo            Implementation
 }
 
 // Agent is the ACP agent server bound to one JSON-RPC connection (one editor).
@@ -50,6 +55,11 @@ type turnRecord struct {
 type acpSession struct {
 	id  string
 	cwd string
+
+	// turnMu serializes prompt turns for one session: concurrent session/prompt
+	// calls run one at a time so they can't interleave history or clobber the
+	// single cancel slot.
+	turnMu sync.Mutex
 
 	mu      sync.Mutex
 	mode    agent.PermissionMode
@@ -95,9 +105,11 @@ func (a *Agent) handleInitialize(_ context.Context, params json.RawMessage) (any
 	return InitializeResult{
 		ProtocolVersion: negotiated,
 		AgentCapabilities: AgentCapabilities{
-			LoadSession:         true,
-			PromptCapabilities:  PromptCapabilities{Image: true},
-			SessionCapabilities: SessionCapabilities{Resume: true, Load: true},
+			// Only advertise what ZERO actually implements: session/load (loadSession)
+			// and image prompts. session/resume + the session-capability sub-object
+			// are intentionally omitted since there is no resume handler yet.
+			LoadSession:        true,
+			PromptCapabilities: PromptCapabilities{Image: true},
 		},
 		AgentInfo: &info,
 		// ZERO owns credentials (BYOK) and does not delegate auth to the editor.
@@ -112,18 +124,18 @@ func (a *Agent) handleSessionNew(_ context.Context, params json.RawMessage) (any
 	if err := json.Unmarshal(params, &p); err != nil {
 		return nil, RPCError(codeInvalidParams, "invalid session/new params")
 	}
-	if strings.TrimSpace(p.Cwd) == "" {
-		return nil, RPCError(codeInvalidParams, "cwd is required")
+	root, err := a.deps.ResolveWorkspaceRoot(p.Cwd)
+	if err != nil {
+		return nil, RPCError(codeInvalidParams, err.Error())
 	}
-	meta, err := a.deps.Store.Create(sessions.CreateInput{Title: "ACP session", Cwd: p.Cwd})
+	meta, err := a.deps.Store.Create(sessions.CreateInput{Title: "ACP session", Cwd: root})
 	if err != nil {
 		return nil, RPCError(codeInternalError, "create session: "+err.Error())
 	}
-	sess := a.register(meta.SessionID, p.Cwd)
+	sess := a.registerSession(meta.SessionID, root, nil)
 	return NewSessionResult{
-		SessionID:     sess.id,
-		Modes:         a.modeState(sess),
-		ConfigOptions: a.modelConfigOptions(sess),
+		SessionID: sess.id,
+		Modes:     a.modeState(sess),
 	}, nil
 }
 
@@ -136,15 +148,21 @@ func (a *Agent) handleSessionLoad(_ context.Context, params json.RawMessage) (an
 	if err != nil || meta == nil {
 		return nil, RPCError(codeInvalidParams, "session not found: "+p.SessionID)
 	}
-	cwd := p.Cwd
-	if strings.TrimSpace(cwd) == "" {
-		cwd = meta.Cwd
+	cwdInput := p.Cwd
+	if strings.TrimSpace(cwdInput) == "" {
+		cwdInput = meta.Cwd
 	}
-	sess := a.register(meta.SessionID, cwd)
-	sess.history = a.loadHistory(meta.SessionID)
+	root, err := a.deps.ResolveWorkspaceRoot(cwdInput)
+	if err != nil {
+		return nil, RPCError(codeInvalidParams, err.Error())
+	}
+	// Load history BEFORE publishing the session so no concurrent prompt observes
+	// a half-initialized session (registerSession sets history under the lock and
+	// reuses an already-live session rather than orphaning its in-flight turn).
+	history := a.loadHistory(meta.SessionID)
+	sess := a.registerSession(meta.SessionID, root, history)
 	return LoadSessionResult{
-		Modes:         a.modeState(sess),
-		ConfigOptions: a.modelConfigOptions(sess),
+		Modes: a.modeState(sess),
 	}, nil
 }
 
@@ -159,6 +177,12 @@ func (a *Agent) handleSessionPrompt(ctx context.Context, params json.RawMessage)
 	if sess == nil {
 		return nil, RPCError(codeInvalidParams, "unknown session: "+p.SessionID)
 	}
+
+	// Serialize turns for this session so two prompts can't interleave history or
+	// fight over the single cancel slot. session/cancel still works concurrently
+	// (it doesn't take turnMu).
+	sess.turnMu.Lock()
+	defer sess.turnMu.Unlock()
 
 	userText := promptText(p.Prompt)
 	images := promptImages(p.Prompt)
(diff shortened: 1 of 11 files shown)


Vasanthdev2004 (changes requested):
## Review: ACP surface — **Request changes** 🔴

Really nice piece of work overall — a clean, **purely additive** `internal/acp` package (0 deletions, ~+2208), wired through the existing `appDeps`/`runWithDeps` harness with low blast radius, shipping with real tests (`clientHarness`, jsonrpc/permission/translate suites), a fail-closed permission transport, and a sensible BYOK design where Zero keeps provider/model/keys and the editor just hosts the thread. CI is green. The architecture is sound and the fixes below are all localized.

But I'm requesting changes: an adversarial review across 5 dimensions (JSON-RPC / protocol / concurrency / **security** / integration), with each finding independently re-verified, surfaced **4 blockers** — two of them genuine security regressions I confirmed myself against the diff.

### 🔴 Blockers (must fix before merge)

1. **ACP runs shell tools with NO sandbox — the only Zero entry point that does.**
   `internal/cli/acp.go runACP` → `internal/acp/agent.go runTurn` builds `agent.Options` with **no `Sandbox` engine**, so `bash`/`exec_command` execute **unconfined on the host**. I verified it: `grep -i sandbox internal/acp/ internal/cli/acp.go` returns **nothing**, while `exec.go` builds and passes `Sandbox: sandboxEngine` (and the TUI does too). **Fix:** build a sandbox engine in `runACP` (mirror `buildExecSandboxEngine`), pass it as `Options.Sandbox`, and scope tools to the validated session cwd (`CoreToolsScoped(workspaceRoot, scope)`).

2. **Privilege escalation via `set_mode`.**
   `handleSetMode` (agent.go:290) accepts `PermissionModeUnsafe`, and `AvailableModes` (agent.go:346) advertises it to **any client** as *"Full access — Run all tools without asking."* In Unsafe mode the loop sets `permissionGranted=true` and never fires `OnPermissionRequest`. The TUI deliberately gates Unsafe behind `--skip-permissions-unsafe`; here a single client call gets unconfined, no-prompt access. Combined with #1 that's **no-prompt host shell for a buggy/hostile editor.** **Fix:** drop Unsafe from `AvailableModes` and reject it in `handleSetMode` absent an explicit operator opt-in; clamp to auto/ask.

3. **Data race on `acpSession.history`.**
   `handleSessionLoad` calls `register()` (publishes the session into the shared `a.sessions` map) at line 143, then writes `sess.history = a.loadHistory(...)` **unguarded** at line 144 — while `appendHistory`/`snapshotHistory` touch the same field under `sess.mu`. A concurrent `session/prompt` (own goroutine) racing this trips `-race` / a torn read. **Fix:** assign history **before** publishing via `register()`, or use a mutex-guarded setter.

4. **Unbounded blocking send can wedge the read loop.**
   `deliver()` in `jsonrpc.go` does `ch <- msg` with no `select/default` on the synchronous read-loop goroutine. If a `Call` abandoned its cap-1 channel (e.g. `session/cancel` cancelling a pending `session/request_permission`) and a second response frame for that id arrives, the send **blocks forever** — no further requests/notifications/cancels processed. `failAllPending()` in the same file already uses the correct non-blocking pattern. **Fix:** apply the same `select { case ch<-…: default: }` and delete the pending entry on delivery.

### 🟠 Should fix (soon)
- **Per-session turn serialization** — concurrent `session/prompt` overwrite the single `sess.cancel` slot and interleave history, corrupting the transcript/cancel targeting. Serialize turns per session; compare-and-clear cancel for the owning turn.
- **`register()` overwrites a live session** on a re-`load` of an in-flight id (orphans the turn, resets mode/model). Reuse or reject-while-active.
- **`SessionCapabilities` typed wrong** — `Resume bool`/`Load bool` don't match ACP v1 (resume is an object; no `load` field), and no `session/resume` handler exists → conformant clients drop resume support and get `-32601`. Model resume as `*SessionResumeCapabilities,omitempty` only if you register a handler.
- **`SessionConfigOption` wire shape non-conformant** — missing the `type:"select"` discriminator, `value`/`values`/`id` instead of `currentValue`/`options`/`value`. A conformant client can't decode the standard model selector. Restructure to the spec select union.
- **Malformed JSON tears down the whole connection** — one bad line makes `json.Decoder` unrecoverable; `codeParseError(-32700)` is declared but never used. Frame per line and reply `-32700` (id=null), continue.
- **Permission offered-allowlist is non-binding** — `actionOffered(...) || isKnownDecision(...)` lets a client return an `always_allow`/`allow_for_session` that wasn't offered for this call. Drop the `|| isKnownDecision` fallback; fail closed to deny.
- **Client cwd unvalidated** — only `TrimSpace != ""`, then it becomes the file-tool confinement root, so `cwd="/"` makes the whole filesystem read/write. Normalize (Abs+Clean+EvalSymlinks), require an existing dir, reject root/home — mirror `exec`'s `resolveWorkspaceRoot`.

### 🟢 Nice-to-have
Validate inbound `jsonrpc:"2.0"` and reject mismatches with `-32600`; don't discard write errors in `writeResult/writeError`; plan translation is lossy (priority hardcoded, `failed`→`completed`); `Serve`'s `Decode` doesn't observe ctx cancellation (pre-existing, non-interactive — optional).

---

**How I verified:** 5 parallel dimension reviewers, every finding independently re-checked against the code, then I personally confirmed the two security blockers (`grep` for sandbox wiring; the `set_mode`/`AvailableModes` Unsafe path) and the history-write-after-publish ordering. The design is good — none of these are showstoppers, but the security and data-race items shouldn't land as-is. Happy to re-review the moment the four blockers are addressed. 🙌

gnanam1990:
Thanks for the deep, adversarial review — every item was actionable. All four blockers and the seven should-fixes are addressed in `389aaed`. Full `go test ./... -race` is green (69 pkgs, 0 fail, 0 races), with new regression tests for the security items.

### 🔴 Blockers
1. **No sandbox on shell tools** — fixed. `runTurn` now builds a **scoped registry + sandbox engine per session** via a new `BuildWorkspace` dep that mirrors `exec`'s `buildExecSandboxEngine` + `sandbox.NewScope`, and passes `Options.Sandbox`. ACP no longer runs `bash`/`exec_command` unconfined. New test `TestACPRunTurnWiresSandboxAndScopedRegistry` asserts the engine + scoped registry actually reach `agent.Options`.
2. **set_mode privilege escalation** — fixed. `AvailableModes` is now `auto`/`ask` only and `handleSetMode` rejects `unsafe` with an error; Unsafe stays gated to the operator. `TestACPSetModeUpdatesSession` now asserts Unsafe is rejected.
3. **history data race** — fixed. `registerSession` sets `history` **before** publishing into the shared map (under the lock) and **reuses** an already-live session instead of overwriting it (also fixes the live-session overwrite below). `-race` clean.
4. **deliver() unbounded blocking send** — fixed. `deliver` deletes the pending entry and sends with `select { … default }`, so a duplicate/late response frame can't wedge the read loop. `TestConnSurvivesMalformedLine` and the existing cancel paths exercise it.

### 🟠 Should-fix
- **Per-session turn serialization** — added `turnMu`; concurrent `session/prompt` now run one at a time (cancel still works concurrently — it doesn't take `turnMu`).
- **register overwrites live session** — `registerSession` reuses the existing session for a re-`load` of a live id.
- **SessionCapabilities typed wrong** — removed; we only advertise `loadSession` (implemented) + `promptCapabilities`. No `session/resume` claim without a handler.
- **SessionConfigOption non-conformant** — dropped the malformed model-selector advertisement rather than ship a shape a conformant client can't decode; model switching remains available via the vendor `_zero/set_model` (and `set_config_option` still accepts `model`). I'd rather re-add it once I can model the spec `select` union exactly — tracked as a follow-up.
- **Malformed JSON tears down the connection** — fixed. Framing is now per-line; a bad line replies `-32700` (id `null`) and the loop continues. New test `TestConnSurvivesMalformedLine`.
- **Permission allowlist non-binding** — fixed. Dropped the `|| isKnownDecision` fallback; the outcome must match an **offered** option or it fails closed to deny.
- **Client cwd unvalidated** — fixed. `ResolveWorkspaceRoot` validates via `exec`'s resolver and additionally rejects the filesystem root and the home dir before it becomes the confinement root. `TestACPRejectsInvalidCwd` covers it.

### 🟢 Nice-to-have
- Added `jsonrpc:"2.0"` validation (mismatch → `-32600`).
- The lossy plan translation and `Serve` ctx-on-`Decode` are noted as follow-ups; client-hosted `fs/*`+`terminal/*` remain intentionally optional since ZERO executes locally.

Ready for re-review when you have a moment. 🙏

Vasanthdev2004 (approved):
## Re-review (round 2) — **Approve** ✅

Excellent turnaround — `389aaed3` addresses **every blocker** and the key should-fixes, with new tests, and CI is **7/7 green**. I re-verified each one against the code, not just the commit message:

### Blockers — all fixed ✅
1. **Sandbox confinement** — new `BuildWorkspace` dep returns a scoped registry + sandbox engine; `runTurn` sets `Options.Sandbox`. Production wiring is real: `cli/acp.go` builds it via the **same `buildExecSandboxEngine`** the exec path uses, scoped through `sandbox.NewScope`. Backed by `TestACPRunTurnWiresSandboxAndScopedRegistry`. Shell tools are no longer unconfined.
2. **`set_mode` escalation** — `PermissionModeUnsafe` now returns `-32602 "mode not permitted over ACP"`, and `AvailableModes` advertises only Auto/Ask. A client can't grant itself no-prompt host access.
3. **`history` data race** — history is now loaded **before** `registerSession` publishes the session (which sets it under the lock), and turns are serialized per session (`turnMu`). No torn read.
4. **`deliver()` blocking send** — deletes the pending entry, then non-blocking `select { case ch<-msg: default: }`. A late/duplicate response frame can't wedge the read loop.

### Should-fixes — addressed ✅
- **cwd validation** → `ResolveWorkspaceRoot` reuses exec's `resolveWorkspaceRoot` (abs+clean, must be an existing dir) — no more filesystem-root confinement.
- **Malformed JSON** → `Serve` is line-framed (`ReadBytes`), replies `-32700` (id null) and **keeps the connection alive**.
- **Permission allowlist** → `decisionFromOutcome` binds to the offered actions and **fails closed to deny** (the `isKnownDecision` fallback is gone).
- **Session reuse** → `registerSession` returns the existing session instead of overwriting a live one.
- **`SessionCapabilities`** → dropped the bogus `Resume`/`Load` bools; advertises only `loadSession`, which is actually implemented.

### Residual (non-blocking follow-ups)
- **`SessionConfigOption` wire shape** still uses a `values`-style object rather than the spec `select` union (`type:"select"` / `currentValue` / `options`). The vendor `_zero/set_model` path works, but a conformant client still can't drive the standard model selector. Worth aligning to spec in a follow-up, not a merge blocker.
- The round-1 nits remain optional: validate inbound `jsonrpc:"2.0"`; don't silently discard `writeResult`/`writeError` errors; plan translation is lossy (`failed`→`completed`).

Clean, well-tested, security-correct now. Clearing my changes-requested. Nice work. 🚀

PR #289, opened 2026-06-21
feat(tui): syntax-highlighted code, word-level diffs, reduced-motion gate

Vasanthdev2004:
The remaining three items from the reference-TUI comparison, in one PR. All three are "real content rendering" + motion accessibility.

## 1. Colored code blocks (syntax highlighting)
Fenced code was rendered flat; now it's tokenized with **chroma** (new pure-Go dep) and painted with Zero's **own audited palette** — keyword→accent, string→green, number→amber, comment→faint, function→blue, punctuation→muted — so it stays on-brand and degrades through the same lipgloss profile path (truecolor→256→16→plain). Token-level wrapping means a color never splits across a wrap. Unknown/missing language falls back to plain (never worse than today).

## 2. Clearer file edits (word-level diff)
`diffCardBody` tinted whole add/del lines; now an isolated 1:1 replacement highlights **only the changed span** (`90`→`300` lights just `9`→`30`) via new contrast-validated brighter-band tokens, after trimming the common prefix/suffix. Gated to ≤60% change so a near-rewrite isn't confettied; block changes keep whole-line tint.

## 3. Better remote/SSH feedback (reduced-motion gate)
One `ZERO_REDUCED_MOTION` switch (+ auto on no-TTY) swaps the spinner for a steady dot at every render site and forces the fade off — one switch for all motion (where `ZERO_NO_FADE` only governed the fade). The elapsed timer keeps liveness, so a static frame never reads as frozen.

## Reviewed adversarially before opening
An adversarial review (4 dimensions → verified → synthesized) caught a **real perf regression** and 3 smaller issues — all fixed in the second commit:
- **Blocker (fixed):** the live streaming block was re-tokenizing the whole code block via chroma *every frame*. Highlighting is now confined to **committed/cached rows** (`renderAssistantMarkdownText(..., allowHighlight)`); the streaming path renders plain, so chroma runs once per row, never per frame.
- Cache language→lexer lookups + short-circuit empty language (avoids chroma's slow registry scan per frame).
- First-token language extraction (` ```go title=x ` → `go`).
- Reduced motion now also freezes the compact/doctor braille rings.

## Verification
- Verified live rendering: `func`→lime, `"hello"`→green, `42`→amber, `// c`→faint; `90`→`300` word-diff; interim renders plain while committed highlights.
- Colors chosen by **WCAG computation** (chroma→audited tokens; diff-span ≥4.6:1, validated separation), with tests guarding both.
- Full `internal/tui` suite green; gofmt/vet/build clean. Committed as Vasanthdev2004.

**Caveat:** exact shades are my best numeric guess (no vision here) — worth a `go run ./cmd/zero` look. Easy to tune or push the diff/selection to bolder variants.

## Summary by CodeRabbit

* **New Features**
  * Language-aware syntax highlighting for fenced code blocks in markdown responses (with plain-text fallback when highlighting isn’t available).

* **Improvements**
  * Reduced-motion mode now switches streaming animations and spinners to static glyphs across the live UI.
  * Diff rendering improved with word-level highlighting for isolated replacements, including stronger contrast for changed spans.
  * Rendering is refined to better separate “committed/selectable” markdown output from the live stream.

* **Tests**
  * Expanded coverage for reduced-motion detection, static spinner glyphs, syntax highlighting fallback/wrapping, and word-level diff rendering and contrast.


gnanam1990 (approved):
## Approve ✅

Well-engineered: syntax highlighting, word-level diffs, and a reduced-motion gate. The central performance claim checks out — live streaming text renders **plain** (`allowHighlight=false`) and chroma runs **once per committed code block** through the LRU render cache, never per-frame. House rule respected: the word-diff band colors live in `theme.go`'s palettes and the syntax mapping uses existing audited tokens (accent/green/amber/blue/muted/faint) with zero inline hex. The new dependency `chroma/v2` is MIT, pure-Go, and broadly maintained (the standard Go highlighter); its transitives (`regexp2`, `repr`) are MIT/pure-Go. All spinner sites convert to the reduced-motion glyph and the compact/doctor braille rings freeze.

> Note: I couldn't build/test offline (chroma `v2.27.0` wasn't in my module cache and egress is restricted), so this is a deep static review — relying on the green CI + the author's verification for the build.

### Minor, non-blocking:
- The **headline feature (token → color mapping) isn't asserted by a test** — `syntax_highlight_test.go` only checks structure/line-count, so a regression in `tokenStyle` wouldn't be caught. Worth a test that a `func`/string/number/comment token renders its mapped palette color.
- Reduced-motion **side-effects** (fade forced off, compact/doctor frames frozen) aren't directly tested — a positive test (`reducedMotion=true` → frame doesn't advance) would lock them in.
- nit: the word-diff test couples to a literal truecolor RGB sequence; deriving the expected output from the palette style would be more robust.

PR #290, opened 2026-06-21
Fix TUI UX issues

anandh8x:
## Summary by CodeRabbit

* **New Features**
  * Improved Ctrl+C exit flow: first Ctrl+C shows an amber “press again to exit” prompt (provider indicator hidden), with automatic expiry; if a run is flushing, exit is deferred until it completes.
  * Enhanced composer text selection with highlighted rendering and mouse click/drag selection that copies selection on release.
  * Expanded transcript selection to cover additional row types (including tool-result bodies and error rows) with selectable column highlighting.

* **Tests**
  * Updated and added coverage for the two-step Ctrl+C flow, disarming on intervening keys, composer-clearing behavior, expiry sequencing, deferred quit, composer mouse selection, and expanded transcript selection.

### DIFF
--- a/internal/tui/collapse_tools_test.go
+++ b/internal/tui/collapse_tools_test.go
@@ -72,7 +72,10 @@ func TestToolResultRowExposesClickToggle(t *testing.T) {
 	m := transcriptViewTestModel()
 	row := transcriptRow{kind: rowToolResult, id: "t", tool: "custom_tool", status: tools.StatusOK, detail: numberedLines(cardBodyMaxLines + 5)}
 	_, selectable := m.renderSelectableToolResultRow(0, row, m.width, buildRowContext(nil), 0)
-	if len(selectable) != 1 || !selectable[0].toggle {
+	if len(selectable) < 1 || !selectable[0].toggle {
 		t.Fatalf("tool result head must be a clickable toggle line, got %#v", selectable)
+	}
+	if len(selectable) < 2 || selectable[1].text == "" {
+		t.Fatalf("tool result visible body/footer must stay selectable, got %#v", selectable)
 	}
 }
--- a/internal/tui/composer.go
+++ b/internal/tui/composer.go
@@ -257,7 +257,7 @@ func (m model) handleComposerSelectionMouse(msg tea.MouseMsg) (model, tea.Cmd, b
 }
 
 func (m model) composerPositionAtMouse(msg tea.MouseMsg) (int, bool) {
-	if !m.altScreen || m.height <= 0 {
+	if !m.altScreen || m.height <= 0 || m.composerMouseSelectionBlocked() {
 		return 0, false
 	}
 	width := chatWidth(m.width)
@@ -277,6 +277,11 @@ func (m model) composerPositionAtMouse(msg tea.MouseMsg) (int, bool) {
 		return 0, false
 	}
 	return m.composerPositionAtVisualCell(localX-2, contentY, maxInt(1, width-4))
+}
+
+func (m model) composerMouseSelectionBlocked() bool {
+	return m.setup.visible || m.providerWizard != nil || m.mcpAddWizard != nil ||
+		m.mcpManager != nil || m.picker != nil || m.suggestionsActive()
 }
 
 func (m model) composerPositionAtVisualCell(x int, y int, width int) (int, bool) {
--- a/internal/tui/composer.go
+++ b/internal/tui/composer.go
@@ -16,6 +16,12 @@ type composerState struct {
 	cursor int
 }
 
+type composerSelectionState struct {
+	active bool
+	anchor int
+	cursor int
+}
+
 type composerPastePreview struct {
 	active bool
 	start  int
@@ -151,6 +157,7 @@ func (m *model) clearComposer() {
 	m.composer = composerState{}
 	m.composerActive = false
 	m.composerPastePreviews = nil
+	m.composerSelection = composerSelectionState{}
 	m.input.SetValue("")
 }
 
@@ -158,6 +165,7 @@ func (m *model) resetComposerFromInput() {
 	m.composer = composerState{}
 	m.composerActive = false
 	m.composerPastePreviews = nil
+	m.composerSelection = composerSelectionState{}
 }
 
 func (m *model) syncInputFromComposer() {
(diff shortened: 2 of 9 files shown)


gnanam1990 (approved):
## Approve ✅

Nice, careful UX change. Verified locally: `gofmt -l` clean, `go vet ./internal/tui/` clean, and the **full `internal/tui` suite is green** (including the new Ctrl+C tests and the existing status-line/contrast guards).

I checked the two subtle correctness points and both hold up:
- **`cancelRun()` is called on both the first and second press, but it's idempotent** — its real work (recording `flushRunIDs`, writing the "Run cancelled." marker, appending the session event) is gated on `m.pending`, which it sets to `false`, so the second call is a no-op (no duplicate marker or flush entry).
- **The first press cancels the run but does NOT quit.** The flush-drain `tea.Quit` at `model.go:1352` is gated on `m.exiting`, which only the *second* press sets — so a cancelled run's flush drains harmlessly (the TUI stays open) until the user confirms with a second Ctrl+C. The checkpoint/`flushRunIDs` safety is fully preserved.
- The `exitConfirmSeq` guard correctly ignores a stale expiry tick (covered by `TestCtrlCExitConfirmationExpires`).

House rules respected: the warning uses the existing `zeroTheme.amber` token (no new hex), tests sit next to source.

### Minor, non-blocking:
- The previous Ctrl+C branch carried a detailed comment explaining *why* it defers quit while `flushRunIDs` is non-empty (checkpoint-orphan / `/rewind` safety). Moving the logic into `handleCtrlC` dropped that comment — the behavior is preserved, but a one-line note in `handleCtrlC` would keep the reasoning discoverable.
- nit: while armed, the status line replaces the provider (and the background-terminal summary) with the warning — intentional and fine, just flagging the trade-off.

Solid work — approving.

Vasanthdev2004 (changes requested):
## Review: TUI UX fixes — **Request changes** 🟡

Nice work — the Ctrl+C exit-confirmation flow and the composer mouse-selection (click-to-cursor + drag-to-copy) are well-built, and CI is green. The exit-confirm machinery in particular is careful: `cancelRun` on first press, a sequence-numbered tick (`exitConfirmSeq`) to invalidate stale expirations, and a `flushRunIDs` drain before quit so `/rewind` stays valid. The composer selection correctly reuses the paste-preview-aware cursor mapping and the OSC52 copy path.

I reviewed it adversarially across core-correctness, mouse/selection, and composer (each finding independently verified against `88be1dee`). Two **medium** issues are worth fixing before merge:

### 🟠 Please fix
1. **Exit-confirm isn't disarmed by intervening keystrokes** (`model.go` ~690-716 / the `KeyPressMsg` reset block ~828-834).
   The reset block clears `transcriptSelection`/`composerSelection` on every key but **not** `exitConfirmActive`. So after a first Ctrl+C on an empty composer, you press unrelated keys (Esc, arrows, Ctrl+O…), and within the 3s window a second Ctrl+C **exits the app** instead of re-prompting — and the amber "Press Ctrl+C again to exit" banner lingers the whole time. **Fix:** disarm on any non-Ctrl+C key, e.g. `if m.exitConfirmActive && !keyCtrl(msg,'c') { m.exitConfirmActive = false; m.exitConfirmSeq++ }` alongside the existing selection resets.

2. **`composerPositionAtMouse` lacks the overlay guard that `transcriptLineAtMouse` has** (`composer.go` ~259-262).
   The composer region is always rendered, so a click there **while a picker / wizard / suggestion overlay is open** falls through, starts a composer selection, and on drag/release calls `copyTranscriptSelectionCmd` — **overwriting the OS clipboard**. Most reachable via the suggestions path (composer always holds text). **Fix:** mirror `transcriptLineAtMouse`'s guard — bail when `m.setup.visible || m.providerWizard != nil || m.mcpAddWizard != nil || m.mcpManager != nil || m.picker != nil || m.suggestionsActive()` (or factor a shared helper).

### 🟢 Nits (optional)
- `view.go` statusLine (~201-211): once you disarm on keystroke, consider keeping the background-run summary visible **alongside** the amber warning rather than replacing it (`else if`), since runs keep going until the second Ctrl+C.
- `composer.go` `composerPositionAtVisualCell` (~282-291): a click on the top border / bottom divider row maps to no input cell — swallow it (`return …,true`) so it can't route to the transcript handler. Benign today (disjoint rects), just tightens routing.
- `mouse_test.go`: the three tests are happy-path — add (a) an overlay-open composer click asserting no cursor/selection change, (b) a wrapped multi-line selection, (c) a selection spanning a paste-preview label asserting `selectedComposerText` returns the expanded text.

### How I verified
4 dimension reviewers (core/mouse/composer) + independent verification of each finding against `88be1dee`; I also re-read `composer.go` myself (the rune slicing in `rangeFor`/`selectedComposerText` is correct). Both medium issues traced to exact lines.

Heads-up: this lands on a TUI that just changed a lot (selection bands / glyphs / declutter in #287/#289/#291) — worth a rebase + quick visual check before merge. Happy to re-review once the two fixes are in. 🙌

Vasanthdev2004 (approved):
## Re-review — **Approve** ✅

Both changes I requested are **fixed and verified** in `8e89a00`, each with a new regression test. Nice, clean follow-up.

### ✅ Requested change 1 — exit-confirm disarms on intervening keystrokes
`disarmExitConfirmation()` (advances `exitConfirmSeq` to invalidate the stale tick) is now called in the `KeyPressMsg` reset block via `if !keyCtrl(msg, 'c') { m = m.disarmExitConfirmation() }`, and reused on the composer-clear and second-press paths. So a stray keystroke after the first Ctrl+C re-arms instead of quitting. Covered by `TestCtrlCExitConfirmationDisarmsOnInterveningKey`. Factoring it into one reusable helper rather than duplicating is exactly right.

### ✅ Requested change 2 — composer mouse guard behind overlays
`composerPositionAtMouse` now bails via `composerMouseSelectionBlocked()` (`setup.visible || providerWizard != nil || mcpAddWizard != nil || mcpManager != nil || picker != nil || suggestionsActive()`), mirroring the transcript-line guard. A click behind a picker/wizard/suggestion overlay no longer starts a selection or clobbers the OS clipboard. Covered by `TestComposerMouseSelectionBlockedWhileSuggestionsOpen`.

### Strengths
- Careful exit-confirm machinery: `cancelRun` on first press, sequence-numbered tick to invalidate stale expirations, `flushRunIDs` drain before quit so `/rewind` stays valid.
- Composer selection correctly reuses the paste-preview-aware cursor mapping + the existing OSC52 copy path.
- Regression tests for each fix (disarm-on-key, re-arm, overlay-blocked click/drag/release) — targeted suite green.

No blockers, no new regressions found. Thanks for the quick turnaround — LGTM. 🙌 (Heads-up: it lands on a heavily-changed TUI, so a rebase + quick visual check before merge.)

PR #291, opened 2026-06-21
feat(tui): declutter the chat — drop the gimmicky billboard chrome

Vasanthdev2004:
## Why
The chat read like a **branding billboard** instead of a clean status surface — the through-line from a 4-dimension audit comparing Zero to comparable agents: *"noise that restates what's already on screen, plus brand/meme flavor with no off-switch."* The fix is almost entirely **deletion**.

## What changed (all deletions)
| Before (noise) | After (clean) |
|---|---|
| `⠋ gitlawbmaxxing · 12s` (40-slot ring: maxxing, pilled, aura-farming, vibe-checking, tomfoolering…) | `⠋ Working · 12s` |
| User turn = full-width painted band + 2 padding bands | `▌  <text>` with one plain blank-line delimiter |
| `completed in 0.4s · 2 tools` stamped under **every** turn | nothing on short turns; a faint `worked for 1m 5s` only past 60s |
| `[auto]` amber badge on nearly every card | gone (mode is already in the composer divider) |
| Plan: header `2/5` **and** a duplicate `[██████░░░░] 2/5` bar | header `2/5` + per-step `✓/•/✗` only |
| Failed turn: bordered note **+** a second red `● error` line | just the bordered note |

```
WORKING:  ⠋ Working  ·  12s
USER:     ▌  fix the chat ui
SHORT FINAL:  Done.            (no completion stamp)
```

## Verification
- Deleted `working_words.go` (+ its tests), `renderPlanProgressBar`, the user-prompt padding bands, `onUserPrompt`.
- Updated every test that locked in the old noise to assert the clean behavior; **full `internal/tui` suite green**; gofmt/vet/build clean; **deadcode clean** (verified the same way CI runs it).
- Confirmed the rendered output through the real renderer (above).

## Deferred to a follow-up
The empty-state figlet/tagline, the all-caps reverse-video badges (`PERMISSION`/`ASK`/`NEW FILE`), consolidating model/provider/context across the 4 chrome zones, and terser permission labels — all more layout-sensitive, kept out of this focused chat-decluttering PR.

Note: shades/layout judged from render code (no vision here) — worth a `go run ./cmd/zero` look; easy to tune.

## Summary by CodeRabbit

* **UI/UX Changes**
  * Simplified working status indicator to static "Working" label
  * Removed progress bar visualization from plan display
  * Removed auto-approval badges from tool cards
  * Completion timing information now appears only for longer operations
  * Refined spacing and layout throughout the transcript display


gnanam1990 (approved):
## Approve ✅

A disciplined, mostly-deletion declutter — and the right call for a clean status surface. Verified locally: `gofmt -l` clean, `go vet ./internal/tui/` clean, `go build ./...` clean, and the **full `internal/tui` suite is green**. No compile-breaking dangling references; every test that locked in the old chrome (working words, plan progress bar, padding bands, `[auto]` badge, per-turn completion stamp) was updated to assert the new, quieter behavior rather than just deleted.

### One small follow-up (non-blocking):
The consumers are gone, but a few **theme tokens are now orphaned** and could be removed for a fully clean sweep:
- `userPromptPanel` (theme.go) — assigned but no longer read anywhere.
- `bgPrompt` (color) and the `promptBg` palette token — only fed `onUserPrompt`, which this PR removed.

`deadcode` won't flag these (it reports unreachable *functions*, not unused struct fields), which is why they slipped past the "deadcode clean" check. Worth a one-line cleanup, but it doesn't block merge.

Nice focused PR — approving.

PR #293, opened 2026-06-21
feat(tools+tui): code preview/diff on file writes & edits

Vasanthdev2004:
## Why
When `write_file`/`edit_file`/`apply_patch` ran, the card showed only a one-line summary — `Created cmd/calculator/main.go (1271 bytes).` — **no code**. Comparable coding agents all show the content (preview for new files, diff for edits). Zero already had the **entire** diff-render pipeline (`diffCardBody`, line numbers, +/- coloring, capping) — the tools just never fed it any content.

## What changed
- **UI-only `Display.Body`** on the tool result — a render payload that is **NOT sent to the model**, so a code preview never bloats the model's context. The model-facing `Output` stays the concise summary.
- **`write_file`** → emits a unified diff of old→new: a **new file is an all-added diff**, an overwrite is the **real change** (via `gotextdiff`, the gopls differ).
- **`edit_file`** → emits the real red/green diff.
- **`apply_patch`** → surfaces the applied patch.
- **TUI** → prefers `Display.Body` over `Output` as the card detail; the existing diff card renders it. Bodies cap at 500 lines (the width tier caps again).

```
✓ write_file
  cmd/calculator/main.go                         +3
  @@ -1 +1,3 @@
     1 + package main
     2 +
     3 + func main() {}
```

## Verified
- Diff helper produces correct unified diffs (new = all-added, edit = `-/+`).
- End-to-end: a `write_file` row renders the code through `diffCardBody` (shown above).
- Full `internal/tools` + `internal/tui` suites green; gofmt/vet/build/**deadcode** clean.

## Notes
- Stacks on **#291** (chat declutter) — base is `feat/tui-declutter-chat` so the diff is just this feature.
- New files render as an **all-added (green) diff** with line numbers (standard, like a PR). A syntax-highlighted new-file preview (vs green) is an easy follow-up.

## Summary by CodeRabbit

* **New Features**
  * Tool results now include visual unified diff previews for file modifications from `write_file`, `edit_file`, and `apply_patch`.
  * The interface tool cards now use the diff preview content when available, improving what’s shown in the transcript and code preview areas.
  * File-write calls surfaced via the MCP layer now also get diff-style previews when both path and content are present.
* **Chores**
  * Updated dependencies to include a unified text diff library.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/atotto/clipboard v0.1.4
 	github.com/charmbracelet/colorprofile v0.4.3
 	github.com/charmbracelet/x/ansi v0.11.7
+	github.com/hexops/gotextdiff v1.0.3
 	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
 	golang.org/x/sys v0.46.0
 	mvdan.cc/sh/v3 v3.13.1
--- a/internal/mcp/file_preview_test.go
+++ b/internal/mcp/file_preview_test.go
@@ -1,0 +1,26 @@
+package mcp
+
+import (
+	"strings"
+	"testing"
+)
+
+// An MCP filesystem write_file call should yield a preview body from its content
+// argument; directory-creation and read tools must not.
+func TestMCPFileWriteArgsDetection(t *testing.T) {
+	path, content, ok := mcpFileWriteArgs("write_file", map[string]any{
+		"path": "site/index.html", "content": "<html>\n<body>hi</body>\n</html>\n",
+	})
+	if !ok || path != "site/index.html" || !strings.Contains(content, "<html>") {
+		t.Fatalf("write_file should be detected, got ok=%v path=%q", ok, path)
+	}
+	if _, _, ok := mcpFileWriteArgs("create_directory", map[string]any{"path": "site"}); ok {
+		t.Error("create_directory (no content) must not be treated as a file write")
+	}
+	if _, _, ok := mcpFileWriteArgs("read_file", map[string]any{"path": "x", "content": "y"}); ok {
+		t.Error("read_file must not be treated as a file write")
+	}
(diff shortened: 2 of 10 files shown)


gnanam1990 (approved):
## Approve ✅

Genuinely nice design. Verified locally: `gofmt`/`vet` clean, `go build` clean, and the **full `internal/tools` + `internal/tui` suites are green**.

The standout is the **`Display.Body` separation**: the preview/diff is a UI-only render payload that is **not** sent to the model, so the model-facing `Output` stays the concise summary and a code preview never bloats context. `TestWriteFileReportsChangedFileAndDisplay` proves it both ways — `Display.Body` carries the `+hello` added line while `Output` does not. Reusing the existing `diffCardBody` pipeline (the tools just never fed it content before) is exactly the right move, and the 500-line `capPreview` guard is sensible. The new dep `github.com/hexops/gotextdiff` is the gopls differ — stable and well-chosen.

### Minor, non-blocking:
- **Test coverage:** only `write_file`'s `Display.Body` is asserted. `edit_file` and `apply_patch` go through the same `unifiedDiff`/`capPreview` (so the helper is covered), but a 2-line assertion that an `edit_file` result carries a `-/+` body and `apply_patch` carries the patch would lock those wirings against regression.
- **nit (perf/binary):** `unifiedDiff` runs the Myers diff over the **full** old+new content — the 500-line cap is on the rendered output, not the computation — so a very large overwrite does the full O(ND) diff, and a binary write would produce a noisy text diff. Worth a cheap short-circuit (skip diffing when old/new exceeds some size or looks binary); not blocking, since write content is usually modest text.

### Note on merge order
This **stacks on #291** (base = `feat/tui-declutter-chat`), so merge **#291 first**, then this — or retarget #293's base to `main` after #291 lands so the diff stays just this feature.

Solid work — approving.

PR #294, opened 2026-06-21
fix(sandbox): don't block 'node --version' & other print-and-exit REPL commands

gnanam1990:
## Why

The interactive-command guard (`DetectInteractiveCommand`) blocks programs that would hang a non-interactive agent (a bare `node`/`python` REPL). But it treated those REPL programs as interactive **unless** an argument was a script file or one of a small allow-list of eval/print flags:

```go
"node": {"-e", "--eval", "-p", "--print"},
```

So a perfectly harmless **`node --version`** (`--version` isn't a script file and isn't in that list) got blocked before execution:

```
Error: Blocked interactive command "node" before execution: node with no script…
Suggestion: Run `node script.js` or `node -e '<code>'`.
```

This false-positive hit `node --version`, `node -v`, `node --check x.js`, `node --help`, `python3 --version`, etc. — all print-and-exit commands. Both the `bash` and `exec_command` tools share the guard, so there was no workaround except using `-e`/`-p`.

## What changed

- Added **universal print-and-exit flags** `--version` / `--help` that suppress the guard for **every** repl program. These are unambiguous across all of them.
- Added node's unambiguous short forms `-v`, `--check`, `-h` to its allow-list.
- **Kept it conservative:** only `--version`/`--help` are universal. Short flags that mean different things per program are NOT universal — e.g. `mysql -v` is *verbose* (still opens the prompt), so it stays blocked. A bare `node`/`python` REPL also stays blocked.

## Tests

- `TestDetectInteractiveCommandAllowsNonInteractive` now includes `node --version`, `node -v`, `node --check app.js`, `node --help`, the exact failing `node --version && npm --version`, `python3 --version`, and `php --version`.
- `TestDetectInteractiveCommandBlocksEditors` adds `mysql -v` (verbose ≠ version → still interactive) to lock the conservative behavior; bare `node` stays blocked.
- `go build ./...`, `go vet`, and the full `internal/sandbox` + `internal/tools` suites are green; gofmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of version and help flags (`--version`, `--help`) across interpreters. Commands like `node --version` and `python3 --version` will no longer be incorrectly blocked.

* **Tests**
  * Added test coverage for additional command parsing scenarios involving version and help flag invocations.

### DIFF
--- a/internal/sandbox/safe_command.go
+++ b/internal/sandbox/safe_command.go
@@ -85,13 +85,23 @@ var interactivePrograms = map[string]interactiveProgram{
 var nonInteractiveREPLFlags = map[string][]string{
 	"python":  {"-c", "-m"},
 	"python3": {"-c", "-m"},
-	"node":    {"-e", "--eval", "-p", "--print"},
+	"node":    {"-e", "--eval", "-p", "--print", "-v", "--check", "-h"},
 	"ruby":    {"-e"},
 	"php":     {"-r", "-f"},
 	"psql":    {"-c", "--command", "-f", "--file", "-l", "--list"},
 	"mysql":   {"-e", "--execute"},
 	"mongo":   {"--eval", "-f", "--file"},
 	"mongosh": {"--eval", "-f", "--file"},
+}
+
+// infoExitFlags make ANY repl/interactive program print a message and exit
+// instead of opening a prompt. `--version` and `--help` are unambiguous across
+// these programs (unlike, say, `-v`, which is "version" for node but "verbose"
+// for mysql), so they universally suppress the interactive guard — without them
+// a harmless `node --version` / `python3 --version` was wrongly blocked.
+var infoExitFlags = map[string]bool{
+	"--version": true,
+	"--help":    true,
 }
 
 // interactiveSegments are multi-word interactive invocations. The detector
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Review: don't block `node --version` & print-and-exit info flags — **Approve** ✅

Good, well-scoped fix. `node --version` / `python3 --version` and other print-and-exit info flags no longer trip the interactive-program guard. I reviewed it adversarially with a focus on the **security angle** (could it let an interactive REPL or a bypass through?) and independently traced the classification paths — **it's clean**:

### Security: no bypass ✅
- The `infoExitFlags` check (`safe_command.go:406`) sits **inside** the loop that runs only after the `isREPL` early return, so it applies **only** to the 9 `nonInteractiveREPLFlags` programs (python, python3, node, ruby, php, psql, mysql, mongo, mongosh).
- I traced `sqlite3 --version`, `ssh --version`, and `redis-cli --version` — all still **BLOCKED** (they route through `hasTrailingCommand`, which never consults `infoExitFlags`).
- So this is over-blocking in the **safe** direction: **no interactive program is newly allowed**, and `infoExitFlags` only classifies an invocation as non-interactive — it doesn't weaken any downstream check. `node -e '…'` / trailing `; node` are unaffected.

### Nice touches
- Correctly limits to `--version`/`--help` and **avoids the ambiguous short `-v`** (which is "version" for node but "verbose" for mysql).
- `base := arg; strip after '='` handles `--version=…`-style tokens.
- Tests assert the allow-list (node/python3/php `--version`) and the trailing-command path.

### 🟢 One nit (non-blocking)
The `infoExitFlags` comment (~97-106) reads as if it suppresses the guard for **"ANY repl/interactive program"** / "universally" — but it only applies to the **REPL-flag** programs, not `sqlite3`/`ssh`/`redis-cli`/`gdb`/etc. Either:
- **(a)** reword the comment to scope the claim to the REPL-flag set, or
- **(b)** if you actually want `--version` to suppress the guard for *every* interactive program, move the `infoExitFlags` check **above** the `isREPL` branch and add a non-REPL test (e.g. `sqlite3 --version` allowed) to lock it in.

Either is fine — approving as-is since the behavior is security-correct and the residual over-blocking is harmless. 👍

PR #295, opened 2026-06-21
fix(sandbox): stop bricking Windows when the sandbox isn't set up (degrade like Linux/macOS)

Vasanthdev2004:
## The bug (every Windows user hits this)
On Windows, the sandbox command runner **hard-fails every shell command** until a one-time **elevated** `zero sandbox setup` writes the `windows-setup.json` marker:
```
zero-windows-command-runner.exe: windows sandbox is not initialized for this
workspace — run `zero sandbox setup` from an elevated (Administrator) terminal
exit 1
```
So a fresh Windows user can't run a single command. **Linux/macOS already *degrade*** when their sandbox is unavailable — Windows was the lone outlier that bricks itself.

## Grounding (analyzed all the reference agents)
None of them brick the CLI because a setup step hasn't run:
- **One reference agent** — has a real Windows sandbox but **defaults it OFF**; runs under approval policy; elevated setup is an opt-in first-run prompt, never a precondition.
- **The others** — degrade on Windows (no OS sandbox / silent fallback + warn); permission prompts are the real safety floor.
- The only sanctioned hard-fail anywhere is an explicit *opt-in* mandatory-sandbox policy.

## The fix
`BuildExecutionRequest`, on Windows when the marker is missing:
| Preference | Behavior |
|---|---|
| **auto** (default) | **Degrade** → command runs unwrapped under the in-process **path-confinement + per-command approval**, with a one-time "run `zero sandbox setup` for full isolation" notice. No brick. |
| **`--sandbox require`** | Still hard-errors with a clear setup hint — strict users keep their guarantee. |
| marker present | Unchanged: native, wrapped, fully sandboxed. |

This is safe because the human-in-the-loop approval gate + path confinement *are* the actual control; the OS sandbox (WFP/ACLs) is defense-in-depth, not the only defense.

## Verified
`TestWindowsDegradesWhenSandboxNotInitialized` covers all three paths (degrade / require-errors / native-when-initialized). Sandbox + doctor + cli suites green; vet + repo build clean.

## Follow-ups (separate)
- Extend **self-dispatch to Linux** so dev doesn't need to build `zero-linux-sandbox`/`zero-seccomp` helpers (the friction a contributor hit).
- Optionally adopt an **unelevated restricted-token tier** as the Windows default hardening (real isolation, no admin).

## Summary by CodeRabbit

* **New Features**
  * Added Windows sandbox setup detection to improve decision-making and provide clearer downgrade/error messaging when initialization is incomplete.

* **Bug Fixes**
  * Updated Windows enforcement behavior to optionally fail when sandbox setup is missing, or otherwise downgrade gracefully.
  * Tightened native shell sandbox activation so shell execution remains unwrapped until Windows sandbox setup completes.

* **Tests**
  * Added and updated Windows-focused regression coverage to verify downgrade/enforcement behavior and shell auto-allow tracking based on setup status.

### DIFF
--- a/internal/sandbox/engine.go
+++ b/internal/sandbox/engine.go
@@ -207,7 +207,19 @@ func (engine *Engine) shellSandboxActive(policy Policy) bool {
 		return false
 	}
 	backend := engine.backend
-	return backend.Available && backend.Executable != "" && backend.CommandWrapping && backend.NativeIsolation
+	if !(backend.Available && backend.Executable != "" && backend.CommandWrapping && backend.NativeIsolation) {
+		return false
+	}
+	// On Windows the command is only actually wrapped once `zero sandbox setup`
+	// has written the marker; until then execution DEGRADES to unwrapped (see
+	// manager.BuildExecutionRequest). The engine must mirror that — otherwise it
+	// would auto-allow a shell command as "sandboxed" while it really runs
+	// unsandboxed, breaking the invariant above and the per-command approval
+	// floor. When not initialized, fall through so the command prompts.
+	if backend.Platform == "windows" && !windowsSandboxInitialized() {
+		return false
+	}
+	return true
 }
 
 // Precheck reports the sandbox blocks that would block a tool request BEFORE
--- a/internal/sandbox/manager.go
+++ b/internal/sandbox/manager.go
@@ -1,9 +1,36 @@
 package sandbox
 
 import (
+	"errors"
+	"os"
 	"os/exec"
 	"runtime"
 )
+
+// errWindowsSandboxNotInitialized is returned only to a caller that explicitly
+// REQUIRED the sandbox (--sandbox require) on Windows before `zero sandbox setup`
+// has run. The default (auto) preference degrades instead — see BuildExecutionRequest.
+var errWindowsSandboxNotInitialized = errors.New(
+	"the Windows sandbox is not initialized — run `zero sandbox setup` from an elevated (Administrator) terminal, " +
+		"or drop `--sandbox require` to run with workspace path-confinement and per-command approval")
+
+// windowsSetupDowngradeReason is surfaced when Windows falls back to the
+// in-process policy gate because the OS sandbox has not been set up.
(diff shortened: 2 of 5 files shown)


Vasanthdev2004:
Follow-up commit `a90a6f3`: closed a gap this degrade introduced. The permission engine's `shellSandboxActive` keyed on backend *capabilities*, which stay true on Windows even pre-setup — so once degrade let commands run unwrapped, the engine still **auto-allowed ordinary shell commands without a prompt** while they ran unsandboxed (breaking the per-command-approval floor + the function's own doc invariant). Now `shellSandboxActive` returns false on Windows when the sandbox isn't initialized, so degraded shell commands **prompt** (human-in-the-loop floor); native (post-setup) still auto-allows. Network/destructive/out-of-workspace checks run earlier and are unaffected. Regression test added (degraded→prompt, native→auto-allow).

gnanam1990 (approved):
## Approve ✅

A correctness fix that also closes a security gap. Verified locally (macOS + Windows cross-compile):
- `gofmt -l` clean, `go vet ./internal/sandbox/` clean, `internal/sandbox` + `internal/doctor` suites green.
- **`GOOS=windows go build ./...` green.**
- `TestWindowsDegradesWhenSandboxNotInitialized` passes all three paths (degrade on auto / error on `--sandbox require` / native when the marker exists).

What I checked specifically:
- The `engine.go` `shellSandboxActive` change is the load-bearing part: without it the engine would auto-allow a shell command as "sandboxed" while it actually runs **unwrapped** on a not-yet-set-up Windows host. Returning `false` there makes it fall through to **per-command approval** — the same floor Linux/macOS degrade to. Good catch.
- All new behavior is gated on `goos == "windows"` / `backend.Platform == "windows"`, so Linux/macOS are byte-identical.
- Degrading (rather than bricking) matches how the other platforms already behave when their OS sandbox is unavailable; the human-in-the-loop approval + in-process path-confinement remain the actual control.

Clean, well-scoped, well-tested — approving.

PR #296, opened 2026-06-21
fix(sandbox): run Homebrew/usr-local tools (node, python3) in the macOS sandbox

gnanam1990:
## Why

On macOS, a **scoped** (workspace-enforced) `exec_command`/`bash` could not run user-installed interpreters like Homebrew `python3`/`node`. Two symptoms, same area:

- `python3` resolved to the broken `/usr/bin/python3` stub → `xcode-select: No developer tools were found`.
- After making the tools reachable, `python3` still failed at startup with `realpath: /opt/homebrew/bin/: Operation not permitted`, while `node` worked — a telling asymmetry.

The scoped profile uses a scrubbed minimal PATH and **restricted** reads, and the user-package trees weren't fully accounted for.

## Root cause

Running a Homebrew/`usr-local` binary needs four things from the macOS seatbelt profile, and each was missing:

1. **PATH** — the scrubbed env omitted `/opt/homebrew/{bin,sbin}` (+ `/usr/local`), so the interpreter wasn't found (or resolved to a system stub).
2. **read** — only the bare `.../lib` dirs were readable; the interpreter in `.../bin` and its Cellar/opt dylibs were not.
3. **`file-map-executable`** — restricted to system frameworks + `/usr/lib`, so a Homebrew binary couldn't load its own dylibs (`node` → `libnode`/`libuv`/…, `python3` → its framework).
4. **`file-read-metadata` on the *ancestors*** — `realpath()`/`getcwd()` lstat every path component, so resolving a tool also stats the parent dirs (`/opt`, `/usr`). The subpath read rules cover the trees themselves but **not `/opt`**, so `python3`'s startup `realpath()` was denied. `node` only `exec`s (kernel traversal, allowed), which is exactly how the gap was isolated.

Confirmed by dumping the **real** scoped profile and reproducing under `sandbox-exec` (a control test with/without rule #4 flipped the result deterministically).

## What changed — `internal/sandbox/runner.go`

- `ensureMacToolPaths`: prepend `/opt/homebrew/{bin,sbin}` + `/usr/local/{bin,sbin}` to the sandbox PATH on darwin (no duplicates).
- Broaden the platform read roots from `.../lib` to the whole `/opt/homebrew` and `/usr/local` trees.
- Add `/opt/homebrew` and `/usr/local` to `file-map-executable`.
- Add `file-read-metadata` on `(path-ancestors "/opt/homebrew")` / `(path-ancestors "/usr/local")`.

## Scope / safety

- **Writes stay confined to the workspace** — only reads/exec of user-package trees are added.
- **Network is untouched** — binding a listening socket (e.g. `python3 -m http.server`) is still governed by the existing network policy; this PR does not change it.
- Non-macOS platforms are unaffected (PATH augmentation is `darwin`-gated; the rules live in the seatbelt profile).

## Verification

End-to-end under the real profile via `sandbox-exec`: `python3` and `node` now resolve, start, import stdlib, and read/write the workspace. New regression tests cover the PATH helper, the read roots, the map-executable roots, and the ancestor-metadata rule. `go build ./...`, `go vet`, full `internal/sandbox` suite, and gofmt are clean.

## Summary by CodeRabbit

* **Tests**
  * Added unit tests validating macOS Homebrew tool integration in sandbox environments.

* **New Features**
  * Sandboxed environments on macOS now properly expose Homebrew (`/opt/homebrew`) and user-local (`/usr/local`) tool directories, enabling access to user-installed interpreters and tools.

### DIFF
--- a/internal/sandbox/homebrew_tools_test.go
+++ b/internal/sandbox/homebrew_tools_test.go
@@ -1,0 +1,116 @@
+package sandbox
+
+import (
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// ensureMacToolPaths must make Homebrew / usr-local bin dirs reachable without
+// duplicating entries the inherited PATH already has.
+func TestEnsureMacToolPaths(t *testing.T) {
+	const allFour = "/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/local/sbin"
+
+	t.Run("empty path gets the tool dirs", func(t *testing.T) {
+		if got := ensureMacToolPaths(""); got != allFour {
+			t.Fatalf("ensureMacToolPaths(\"\") = %q, want %q", got, allFour)
+		}
+	})
+
+	t.Run("scrubbed default path is prefixed, original kept", func(t *testing.T) {
+		got := ensureMacToolPaths("/usr/bin:/bin:/usr/sbin:/sbin")
+		want := allFour + ":/usr/bin:/bin:/usr/sbin:/sbin"
+		if got != want {
+			t.Fatalf("ensureMacToolPaths(default) = %q, want %q", got, want)
+		}
+	})
+
+	t.Run("existing tool dir is not duplicated", func(t *testing.T) {
+		got := ensureMacToolPaths("/opt/homebrew/bin:/usr/bin")
+		// /opt/homebrew/bin was already present, so only the other three prepend.
+		want := "/opt/homebrew/sbin:/usr/local/bin:/usr/local/sbin:/opt/homebrew/bin:/usr/bin"
+		if got != want {
+			t.Fatalf("ensureMacToolPaths(partial) = %q, want %q", got, want)
+		}
+		if strings.Count(got, "/opt/homebrew/bin") != 1 {
+			t.Fatalf("ensureMacToolPaths duplicated /opt/homebrew/bin: %q", got)
+		}
+	})
+
+	t.Run("path that already has every tool dir is unchanged", func(t *testing.T) {
+		input := allFour + ":/usr/bin"
+		if got := ensureMacToolPaths(input); got != input {
+			t.Fatalf("ensureMacToolPaths(complete) = %q, want unchanged %q", got, input)
+		}
+	})
+}
+
+// The macOS read roots must cover the whole user-tool trees so an interpreter in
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Review: run Homebrew/usr-local tools in the macOS sandbox — **Approve** ✅

Really well-diagnosed fix. Getting `node`/`python3` to run under Seatbelt needs four distinct things and you nailed each: PATH augmentation (darwin-gated), read roots for the tool trees, `file-map-executable` for those trees, and — the sharp catch — `file-read-metadata` on the **ancestors** (`/opt`, `/usr`) so `realpath()` at interpreter startup succeeds. The `node`-only-exec vs `python3`-needs-realpath asymmetry pinpointing the missing ancestor-metadata rule is a genuinely good piece of root-causing, and it's backed by solid regression tests (PATH dedup/order, read roots, map-executable roots, the ancestor rule, darwin-only PATH exposure). `ensureMacToolPaths` is clean and idempotent.

I reviewed it adversarially with a focus on **sandbox safety**, and confirmed there's **no escape and no write-escalation**: writes stay governed by the independent `writeRoots` mechanism (workspace-confined), and the network policy is untouched. 👍

### 🟢 One nit (low, non-blocking)
`macosPlatformReadRoots()` widens the read roots from the bare `.../lib` dirs to the **entire** `/opt/homebrew` and `/usr/local` trees. A Seatbelt `subpath` allow grants the whole subtree, so this also makes **non-code areas** readable that weren't before — notably `…/etc` (service/daemon configs, which can hold creds) and `…/var` (Homebrew-managed Postgres/Redis data). Since `DenyRead` is empty by default and network is independent, in network-allow mode a command could read + exfiltrate that local service data.

It's **low severity** — the primary credential stores (`~/.aws`, `~/.ssh`, `~/.config`) live under `$HOME` and are untouched, and this only matters if the user runs Homebrew-managed services. Suggestion: scope the read roots to the **code-bearing** subtrees (`bin`, `Cellar`, `opt`, `lib`, `share`, `include`) and skip `etc`/`var`; or, if full-tree read is intentional, add `etc`/`var` to a default `DenyRead` and note the widened scope in a comment.

Approving as-is since it's not a confinement break — just worth tightening the read scope in a follow-up if easy. 🙌

PR #297, opened 2026-06-21
feat(sandbox): online/offline Windows network identity — run approved network commands sandboxed

Vasanthdev2004:
## The bug
Approved network commands bricked on Windows once the sandbox was set up:
```
zero-windows-command-runner.exe: windows sandbox setup is out of date: network policy changed
exit 1
```
The setup marker was locked to **network=deny**. When you approve a network command (`curl`, `git push`, `npm install`), Zero grants it network → it runs as **network=allow** → the deny-marker rejects it. So approved network commands couldn't run under the Windows sandbox at all.

## The fix — one synthetic "offline-marker" SID
Zero's WFP block filter already keys on an `ALE_USER_ID` security descriptor that grants arbitrary SIDs, and Zero runs as a restricted token of the current user. So no second logon account is needed (rather than a separate-logon-user model). We mint **one** offline-marker SID per sandbox home and scope the persistent block filter to it. Network is then gated purely by the restricted token's SID set:

| Mode | Token restricting SIDs | Network | Write-jail |
|---|---|---|---|
| **deny** (offline) | write-caps **+ offline-marker SID** | 🚫 blocked | ✅ |
| **allow** (online, approved) | write-caps only | 🌐 reaches out | ✅ |

The setup marker becomes **mode-agnostic** — it fingerprints the provisioned offline *infrastructure* (`NetworkInfraHash` + `OfflineFilterSID`), not the per-command mode — so one setup validly serves both an `allow` and a `deny` command. Schema bumped **3→4** (forces a clean re-setup; old markers scoped the filter to write SIDs).

No extra logon users, no DPAPI, no admin at runtime, no password rotation.

## Files
- `windows_runner.go` — offline SID in `WindowsCapabilitySIDs` (schema 2, back-compat upgrade) + `WindowsOfflineMarkerSID` + `windowsRuntimeTokenSIDs`
- `windows_network.go` — `BuildWindowsNetworkInfraPlan` + `WindowsNetworkInfraHash`; removed the now-dead mode-coupled `BuildWindowsNetworkPlan`/`*Hash`
- `windows_setup.go` — marker schema 4; validation drops per-command-mode equality
- `windows_setup_windows.go` — setup always installs the offline-scoped infra filter
- `windows_command_runner_windows.go` — compose the token SID set per network mode

## Tested
**Cross-platform core (fully unit-tested):** per-mode token-SID selector; mode-independent infra plan + stable hash; back-compat SID upgrade; **one marker validates BOTH modes** (the core-bug regression); schema-3 rejected. `internal/sandbox` + `internal/doctor` green; Windows cross-compile green.

**Load-bearing assumption** (`ALE_USER_ID` matches a *restricting* SID — the current deny path already relies on this) is verified in a **Windows admin integration smoke test** before merge.

Pairs with #295 (the degrade fix) but is independent of it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upgrading of existing Windows sandbox capability and setup markers, adding support for a missing offline marker SID and rejecting outdated schema versions more reliably.
* **Refactor**
  * Updated Windows sandbox network filtering to use a persistent, mode-independent infrastructure and corresponding setup fingerprinting, improving consistency across online/offline (allow/deny) runs.
* **Tests**
  * Refreshed Windows networking and setup validation tests to cover the new offline-marker and infrastructure hashing behavior.


Vasanthdev2004:
### Known limitation (documented, out of scope for this PR)

This PR fixes the marker brick so approved network commands run sandboxed and reach the network. **HTTPS via Windows Schannel** (e.g. a Schannel-backed `curl.exe`) still fails inside the restricted token with `SEC_E_NO_CREDENTIALS` — Schannel can't acquire its per-user TLS credential under a `WRITE_RESTRICTED`/`LUA` token. This is a fundamental **restricted-token vs Schannel** incompatibility (the standard mitigation is to run TLS in a broker process, not the sandboxed one) and has no clean in-token fix.

**Impact is small:** only affects users who run `zero sandbox setup` *and* shell out to HTTPS. Workarounds: **degraded mode** (the default — no restricted token, so Schannel works), or the in-process **`web_fetch`** tool (the agent already falls back to it).

Plain-HTTP network commands now work sandboxed, and the mode-agnostic marker removes the "network policy changed" brick.

gnanam1990 (approved):
## Approve (with two on-record conditions) ✅

Strong engineering. The design is elegant — `windowsRuntimeTokenSIDs` appends the offline-marker SID **only** in `NetworkDeny` mode, so a deny command's restricted token carries the SID the persistent WFP block filter matches (no network) while an approved allow command doesn't (network reaches), both still write-jailed by the capability SIDs. The setup marker becoming mode-agnostic (fingerprinting the offline infra, not the per-command mode) is the right fix for the "one setup serves both modes" bug, and the back-compat SID mint for schema-1 files is handled.

Verified locally: `gofmt`/`vet` clean, `internal/sandbox` + `internal/doctor` green, **`GOOS=windows go build ./...` green**, and the core regressions pass — `TestWindowsSandboxSetupMarkerValidatesBothNetworkModes` (the exact bug), `…RejectsOldSchema`, the mode-independent infra plan + stable hash, and the offline-SID back-compat upgrade. No third-party names (the Schannel note is vendor-neutral).

**Two conditions I'm approving *on the record* — please don't lose them before merge:**

1. **The Win32 security core is not verifiable in cross-platform CI.** Whether the offline-marker SID actually denies network, and whether `ALE_USER_ID` matches a *restricting* SID, is the entire security property of this PR — and it's only exercised by the **pending Windows admin integration smoke test**. That smoke test is a **hard pre-merge gate**, not a follow-up. Please confirm it passes on a real elevated Windows host before merging.
2. **The documented Schannel limitation is a real narrowing of "online" mode.** Approved online HTTPS commands (TLS via Schannel — common for git/npm/`curl https://`) fail inside the restricted token with `SEC_E_NO_CREDENTIALS`. The code documents it and points to the degraded path / `web_fetch` as workarounds, which is the honest call — but practically, "online" mode delivers raw socket reach more than working HTTPS. Worth a line in the user-facing docs so Windows users aren't surprised.

Engineering and tests are solid — approving, conditioned on (1).

PR #298, opened 2026-06-22
feat(exec): container-trusted sandbox-off mode + run deadline

gnanam1990:
## Summary
Two opt-in controls for `zero exec`, for headless/automated runs inside an already-isolated environment (a container or CI sandbox):

- **`--sandbox off`** (or `ZERO_SANDBOX=off|disabled|none|0`) — disables ZERO's in-process filesystem + network confinement so the agent can write outside the workspace and reach the network. **Off by default**; without the flag the policy stays enforcing (workspace-confined, network-denied). Emits a loud warning when enabled. Intended only for use inside an isolated container, never on a host.
- **`--deadline <dur>`** (or `ZERO_DEADLINE`) — a wall-clock budget; ZERO yields a best-effort result with flushed output at the deadline instead of running unbounded, and reports a deadline stop distinctly from a user interrupt.

## Why
Headless runners (CI jobs, batch exec, benchmark harnesses) run ZERO inside a container that already provides isolation and need (1) the agent to install deps / write outputs without the in-process sandbox fighting the container, and (2) a hard per-run time budget with a clean, flushed result.

## Safety
- Default is unchanged and **guarded by a test** (`TestExecSandboxPolicyDefaultUnchanged`): absent the flag/env, the exec policy is the enforcing default (mode=enforce, network=deny, workspace-confined).
- `--sandbox off` is explicit opt-in and warns loudly — never a default, never implicit.

## Tests
`internal/cli/exec_sandbox_deadline_test.go` covers policy mapping (default vs off), value resolution (flag>env precedence; invalid/empty inert), and arg parsing for `--flag value` and `--flag=value`. `go build` / `go vet` / `go test ./internal/cli/` all clean.

## Summary by CodeRabbit

* **New Features**
  * Added `--sandbox` flag to allow disabling sandbox mode during execution runs
  * Added `--deadline` flag to set wall-clock timeout limits for executions
  * Displays warning when sandbox mode is disabled

* **Tests**
  * Added tests for sandbox and deadline flag parsing and resolution behavior

### DIFF
--- a/internal/cli/acp.go
+++ b/internal/cli/acp.go
@@ -52,7 +52,7 @@ func runACP(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int
 			if err != nil {
 				return nil, nil, err
 			}
-			engine, err := buildExecSandboxEngine(workspaceRoot, resolved, deps, scope)
+			engine, err := buildExecSandboxEngine(workspaceRoot, resolved, deps, scope, false)
 			if err != nil {
 				return nil, nil, err
 			}
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -105,6 +105,16 @@ type execOptions struct {
 	// additional write roots for this run. Unioned with
 	// config.SandboxConfig.AdditionalWriteRoots at scope construction time.
 	addDirs []string
+	// sandboxMode is the raw --sandbox flag value ("" = default enforce, "off"/
+	// "disabled" = container-trusted). Resolved (with the ZERO_SANDBOX env
+	// fallback) into a disable decision in runExec. Off by default; the disabled
+	// path is an EXPLICIT, loudly-warned opt-in meant ONLY for use inside an
+	// already-isolated container (CI / benchmark).
+	sandboxMode string
+	// deadlineRaw is the raw --deadline flag value (a Go duration like "900s").
+	// Resolved (with the ZERO_DEADLINE env fallback) into a wall-clock budget for
+	// the run. Empty = no deadline (current behavior).
+	deadlineRaw string
 }
 
 type execUsageError struct {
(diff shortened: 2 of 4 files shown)


Vasanthdev2004 (approved):
## Review verdict: ✅ Approve

Model PR — clean, tightly scoped, fail-closed, and the default is *provably* unchanged. CI green on all platforms.

**`--sandbox off` / `ZERO_SANDBOX`**
- Explicit opt-in, off by default; flips all three gates coherently (`Mode=disabled`, `Network=allow`, `EnforceWorkspace=false`) so both the in-process `Evaluate` path and the OS wrapping go permissive together.
- Warns loudly when enabled, with the "only inside an isolated container" framing.
- ACP path correctly stays sandboxed (passes `false`) — the off-switch is exec-only.
- Unrecognized values (including `ZERO_SANDBOX=1`, `--sandbox foo`) fail **closed** → stay enforcing. Good default posture.

**`--deadline` / `ZERO_DEADLINE`**
- Wraps `runCtx` with `context.WithTimeout` and cleanly distinguishes `DeadlineExceeded` (status `"deadline"`) from a signal cancel (`"interrupted"`), flushing the trajectory either way. Reusing `exitInterrupted` for both is fine.
- Invalid / non-positive / empty durations → no deadline (unchanged behavior).

**Tests** — `TestExecSandboxPolicyDefaultUnchanged` is exactly the right load-bearing guard against the off-switch leaking into the default; `resolveSandboxOff`/`resolveDeadline` cover precedence (flag > env), case-insensitivity, fail-closed unknowns, and both `--flag value` / `--flag=value` spellings. Thorough.

No blockers, no reservations — this is how an opt-in escape hatch should be built. Thanks @gnanam1990.

PR #299, opened 2026-06-22
Fix ChatGPT Codex OAuth model list and request shape

anandh8x:
Summary: Updates the ChatGPT OAuth provider default/model picker to the Codex-supported model set, and sends Codex Responses requests with top-level instructions plus explicit store=false. Tests: GOTOOLCHAIN=local go test ./internal/providers/openai ./internal/providercatalog ./internal/providermodelcatalog ./internal/tui ./internal/cli; ZERO_OAUTH_TOKENS_PATH=/tmp/zero-empty-oauth-provider-test.json GOTOOLCHAIN=local go test ./internal/providers.

## Summary by CodeRabbit

* **New Features**
  * Expanded ChatGPT model catalog with additional Codex-variant models (gpt-5.4, gpt-5.4-mini, gpt-5.3-codex-spark)
  * Enhanced Codex Responses API request handling

* **Documentation**
  * Updated default ChatGPT model to gpt-5.5

### DIFF
--- a/docs/oauth-subscriptions.md
+++ b/docs/oauth-subscriptions.md
@@ -170,7 +170,7 @@ preset (no API key — the proxy authenticates):
       "name": "chatgpt",
       "catalogID": "chatgpt-proxy",     // OpenAI-compatible, local, no key
       "baseURL": "http://localhost:10531/v1", // override for your proxy's port
-      "model": "gpt-5"                   // whatever model your proxy serves
+      "model": "gpt-5.5"                 // whatever model your proxy serves
     }
   ]
 }
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -137,7 +137,7 @@ func runAuthChatGPT(args []string, stdout io.Writer, stderr io.Writer, deps appD
 	if _, err := fmt.Fprint(stdout, statuses); err != nil {
 		return exitCrash
 	}
-	if _, err := fmt.Fprint(stdout, "\nUse it with zero, e.g.:\n  zero --provider chatgpt --model gpt-5\n"); err != nil {
+	if _, err := fmt.Fprint(stdout, "\nUse it with zero, e.g.:\n  zero --provider chatgpt --model gpt-5.5\n"); err != nil {
 		return exitCrash
 	}
 	return exitSuccess
--- a/internal/providercatalog/catalog.go
+++ b/internal/providercatalog/catalog.go
@@ -104,7 +104,7 @@ var descriptors = []Descriptor{
 	// way in; RequiresAuth is forced true so catalog consumers know the provider
 	// needs an interactive login before a request will succeed.
 	func() Descriptor {
-		d := openAICompat("chatgpt", "ChatGPT", "https://chatgpt.com/backend-api/codex", "gpt-5", nil)
+		d := openAICompat("chatgpt", "ChatGPT", "https://chatgpt.com/backend-api/codex", "gpt-5.5", nil)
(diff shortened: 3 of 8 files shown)


Vasanthdev2004 (changes requested):
## Review verdict: 🟡 one blocking fix, then ship

Nicely scoped PR — the request-shape change is correct and the catalog refresh was overdue. CI is green on all platforms (Linux/macOS/Windows smoke + security). There's **one model ID in the new curated list that isn't a real Codex model**, and since "fix the model list" is the headline of this PR, that needs correcting before merge.

### 🔴 Blocking — `gpt-5.3-codex-spark` is not a served model
`internal/providermodelcatalog/catalog.go`
```go
{ID: "gpt-5.3-codex-spark", Description: "research preview fast Codex model"},
```
The Codex backend's served model manifest has no `…-spark` slug. The real slugs in that family are `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, `gpt-5.2`. A user who picks this 4th option sends `model: "gpt-5.3-codex-spark"` and gets a backend model-not-found rejection. (`gpt-5.3-codex-spark` looks like a mashup of the real `gpt-5.3-codex` and an experimental `…-spark` tag that the backend doesn't actually serve.)

Fix — use the real slug:
```go
{ID: "gpt-5.3-codex", Description: "strong Codex coding model"},
```
…and update the two matching assertions in `catalog_test.go` and `provider_wizard_oauth_test.go` (the `want` list). The other three IDs and the new `gpt-5.5` default are all correct and currently served, so login and the default path are unaffected — this only breaks the one curated option.

### ✅ What's right
- **Request shape** — top-level `instructions` (string) + explicit `store: false` is exactly the canonical Codex Responses payload. Folding system turns into `instructions` instead of role=`system` `input` items is the correct alignment: the Responses API drives the system prompt through `instructions`, and `store:false` keeps the backend from persisting turns Zero already owns. Good change.
- **Default bump `gpt-5 → gpt-5.5`** — gpt-5.5 is the current strongest Codex coding model and is served, so it's the right default.
- **Tests** — the shape assertions (`instructions == "sys"`, explicit `store:false`, single user `input` item) and the catalog/wizard coverage are on point.

### Nits (non-blocking)
- `Instructions`/`Store` have no `omitempty`, so a system-less request would serialize `"instructions":""`. That matches the canonical client (always sends both) and the runtime always supplies a system prompt, so it's fine as-is — just flagging it.

**Verdict: request changes** — swap the one model ID + its two test `want` entries and this is a clean merge. Happy to push that one-line fix to the branch if you'd rather I just land it.

Vasanthdev2004 (approved):
## Updated verdict: ✅ Approve

Correcting my earlier review — I was wrong about `gpt-5.3-codex-spark`. It's a real, served model: confirmed working live against the ChatGPT Codex backend (the model switches and responds normally, `api model: gpt-5.3-codex-spark`). My earlier check was against a static model-manifest snapshot that doesn't include this preview model — the live backend is the authority and it serves it. Retracting the block.

Everything else verified clean:
- **Request shape** — top-level `instructions` (string) + explicit `store: false` matches the canonical Codex Responses payload; folding system turns into `instructions` instead of role=`system` `input` items is the correct alignment.
- **Default `gpt-5 → gpt-5.5`** plus `gpt-5.4` / `gpt-5.4-mini` are all correct and served.
- **Tests + CI** green across Linux/macOS/Windows.

Good fix — merging. Thanks @anandh8x. 🙏

PR #301, opened 2026-06-22
fix(sandbox): retry approved network shell commands

anandh8x:
## Summary
- classify common package-manager and local-server commands as network-sensitive before sandbox launch
- add reference-style `with_additional_permissions` support for shell tools so approved network access stays sandboxed
- detect runtime network sandbox denials, including exit-0 messages like `Cannot open a network socket`, and retry with sandboxed network access instead of workaround loops
- keep Windows/Linux sandbox network setup aligned with the approved online/offline model
- restore the Windows broad WFP outbound deny filters and keep the targeted ICMP/DNS/SMB filters as defense-in-depth

## Sandbox posture notes
- Default workspace-write uses a read-all/write-jail model: the host root is readable and writes are narrowed to workspace/extra roots; `DenyRead` remains the carve-out for secrets.
- Sandboxed commands inherit the caller environment for CLI parity; sandbox enforcement is filesystem/network policy, not environment scrubbing.
- Linux unrestricted filesystem mode intentionally binds `/` read-write, including host `/dev`, because filesystem sandboxing is disabled in that mode.
- Linux bwrap keeps IPC and UTS shared while still unsharing network when the policy denies egress.

## Verification
- Manual: npm static site with `http-server` on `127.0.0.1:8080` now starts and verifies with curl after network approval
- `GOTOOLCHAIN=local go test ./internal/sandbox`
- `GOTOOLCHAIN=local go test ./internal/agent ./internal/tools`
- `GOTOOLCHAIN=local HOME=/data/data/com.termux/files/home/zero/.tmp/test-home XDG_CONFIG_HOME=/data/data/com.termux/files/home/zero/.tmp/test-config GOCACHE=/data/data/com.termux/files/home/zero/.tmp/go-cache go test ./internal/providers`

## Release Notes

* **New Features**
  * Added per-command sandbox permission overrides, including inline additional permissions, with improved prompting for escalated requests.
  * Implemented automatic retries after sandbox denials (including network-denied shell execution).

* **Improvements**
  * Expanded detection of network-capable tools/local servers (e.g., common dev servers and package-manager commands).
  * Improved sandbox environment handling by preserving caller environment variables.
  * Refined Linux and Windows sandbox policy enforcement for more accurate filesystem/network controls.

* **Bug Fixes**
  * Corrected sandbox denial detection and ensured escalated-required prompts/denies behave as expected.


Vasanthdev2004 (changes requested):
## Review verdict: 🔴 Request changes

This is a big, genuinely valuable PR — the network-retry UX, local-dev-server support, and per-command escalation model are well-designed, the approval gating is sound, and the test coverage is broad. CI is green on all platforms. But I found **one Windows security regression that can't ship as-is**, **one naming-policy issue**, and a few **sandbox-posture changes bundled into a "network retry" PR** that should be explicit, conscious decisions rather than side effects. Details below.

---

### 🔴 Blocking 1 — Windows offline/deny mode stops blocking general outbound egress

`internal/sandbox/windows_network.go` — `windowsDenyWFPFilterSpecs()`

On `main`, deny/offline mode applies **two broad WFP filters** (`zero_wfp_block_connect_v4/v6`) on the `ale-auth-connect-v4/v6` layers with **only** a user-SID condition → they block **all** outbound TCP for the offline-marker identity. This PR replaces them with **12 targeted filters** that block only ICMP, DNS:53, DoT:853, and SMB:445/139.

The problem: Zero's Windows sandbox network block is enforced **solely** by these WFP filters scoped to the offline SID. There is **no** AppContainer/LowBox token and **no** `internetClient` capability gating — the token is a plain `CreateRestrictedToken` (`windows_token_windows.go:101`), and `windows_runner.go:478-482` documents the WFP filter as *the* control ("an offline command has no network while an approved online command does"). I verified there's no other mechanism (no host firewall rule, proxy, route nulling, or socket-blocking privilege).

So after this change, an **offline / network-deny** command can still open outbound TCP to arbitrary hosts on any port except 53/853/445/139 — e.g. `HTTPS:443` or raw TCP to a hardcoded IP — which is exactly the egress/exfil path #297's broad block existed to close. Blocking DNS:53 is weak mitigation (raw-IP connections need no DNS; DoH rides 443, which is open).

Why the targeted-only set looks right but isn't here: this mirrors a reference agent's filter set, but that agent runs commands in an **AppContainer with no internet capability** as the *primary* block, using these targeted filters only as defense-in-depth for protocols the AppContainer layer misses. Zero has no such primary block, so the broad filter **is** the primary block and can't be dropped.

**Fix:** keep `zero_wfp_block_connect_v4/v6` (the broad user-scoped block) as the primary deny, and add the 12 targeted filters **on top** as defense-in-depth — not as a replacement. (CI can't catch this: real WFP enforcement needs an elevated Windows host, and the unit tests were updated to assert the new 12-filter shape, so green ≠ validated block.)

### 🔴 Blocking 2 — reference project name embedded in code; also breaks Zero's own naming convention

The new WFP artifacts are named after an external reference project and the provider/sublayer/session strings were renamed **away** from Zero:

- `windows_network.go`: filter names `codex_wfp_icmp_connect_v4`, `codex_wfp_dns_53_v4`, … (12 of them)
- `windows_network_windows.go`: `"Codex Windows Sandbox WFP"` (session/provider/sublayer display) and `"Persistent WFP provider for Codex Windows sandbox filters"`

Per the project rule, pushed code must not carry reference-project names. These also regress Zero's existing convention — the legacy filters were `zero_wfp_block_connect_*` and the strings were `"Zero Windows Sandbox WFP"`.

**Fix:** rename `codex_wfp_*` → `zero_wfp_*` and restore `"Zero Windows Sandbox WFP"` / `"Persistent WFP provider for Zero Windows sandbox filters"` (update the test literals in `windows_network_test.go` to match).

---

### 🟡 Please make these explicit (sandbox-posture shifts bundled into a network-retry PR)

These are defensible and align with a read-all/write-jail model, but they materially change Zero's sandbox threat model and arrive with one-line comments inside a PR about retrying network commands. At minimum call them out; ideally split them into their own PR so they get reviewed as the security decisions they are.

1. **Full-disk read in the default profile** — `profile.go` `permissionProfileReadRoots` now seeds read-roots with `/` (entire filesystem) instead of just the write-roots. The default sandbox can now read every file on disk (e.g. `~/.ssh`, `~/.aws/credentials`, other projects). Writes stay jailed, but combined with #1 + #4 this is a read-any-secret-then-egress path that didn't exist before.
2. **Full host environment passthrough** — `runner.go`/`linux_helper.go` drop `--clearenv` and the scrubbed minimal env; sandboxed commands now inherit `os.Environ()` (all host env vars, including secrets/tokens) with only PATH/TERM/markers upserted on top. Previously the env was scrubbed.
3. **Unrestricted FS mode** — `linuxBwrapFilesystemArgs` changes `--ro-bind / /` to a read-**write** `--bind / /` and drops `--dev /dev` (binds the full host `/dev`). That's full-rw root + raw devices in unrestricted mode.
4. **Dropped `--unshare-ipc` / `--unshare-uts`** from the bwrap arg set (user/pid retained). Minor, but was this intentional? If not, restore them.

### 🟢 What's right (and nicely done)

- **Network retry re-runs the command *sandboxed with granted network*** (`runToolForNetworkRetry` keeps `options.Sandbox`, grants via `grantNetworkForSandboxPrompt` with turn/session scope) — correct use of the online identity, not an unsandboxed fallback.
- **Every retry requires fresh approval**; `PermissionModeUnsafe` is the only auto-allow path.
- **`require_escalated` (unsandboxed) is forced to a prompt when ungranted** — `engine.go` `requestRequiresEscalatedSandbox` → `ActionPrompt` / `ReasonEscalatedSandboxRequired`, and `UnsandboxedExecutionAllowed()` preserves the sandbox whenever denied-reads are configured. Verified by tests; no silent bypass.
- **Linux deny mode now also applies a seccomp `ApplyLinuxNetworkDeny`** in the inner stage — good defense-in-depth.
- **Network-allow correctly shares the host net namespace** so a local dev server (`http-server`/`vite`) is reachable from the host — the headline use case, with a real end-to-end HTTP test.
- Analyzer/risk classification of local-server + package-manager dev/serve/exec commands is reasonable and well-tested.

### Nits (non-blocking)

- `analyzer.go` `npxUsesNetwork` — the loop returns `true` on the first real word and `true` at the end, so it's `return true` unconditionally; collapse it.
- `engine.go` — `require_escalated` in unsafe mode without a grant falls to `deny` (errs safe, but slightly surprising for unsafe mode; worth a comment).

---

**Net:** the feature work is strong and I'd love to land it. Blocking on (1) the Windows egress regression and (2) the naming, and asking that the posture shifts in the 🟡 section be made explicit (or split out). Fix those and I'll re-review quickly. Thanks @anandh8x — this is a meaty, well-tested PR.

Vasanthdev2004 (approved):
## Re-review verdict: ✅ Approve — both blockers fixed

Thanks for the quick turnaround @anandh8x. Both blocking items from the last pass are resolved, and you added regression guards for each. CI green on all platforms.

### ✅ Blocking 1 — Windows offline egress block: restored
`windowsDenyWFPFilterSpecs()` now ships **14** filters: the two **broad** `zero_wfp_block_connect_v4/v6` (user-condition only → block *all* outbound for the offline SID, the primary control) **plus** the 12 targeted ICMP/DNS/SMB filters as defense-in-depth — exactly the right shape. The offline identity is offline again: general TCP (443/80/arbitrary) is blocked, with the targeted rules hardening the protocols the broad rule's layer doesn't see. `windows_online_offline_test.go` updated to assert 14 (`2 broad + 12 targeted`), and `TestWindowsDenyWFPFilterSpecsKeepBroadBlockAndTargetedRules` locks the broad/targeted split in place.

### ✅ Blocking 2 — reference naming: removed
All filters are `zero_wfp_*` now, the provider/sublayer/session strings are back to "Zero Windows Sandbox WFP", and `TestWindowsDenyWFPFilterSpecsKeepBroadBlockAndTargetedRules` adds a guard that fails if any filter name ever starts with `codex_` again. Good — that keeps it from regressing.

### 🟡 Heads-up for the maintainer (not a blocker) — these posture shifts are intentionally in this PR
The 🟡 items from the first pass are **still here, now as deliberate, tested decisions** (you added tests asserting them — `--clearenv` absent, `--unshare-ipc/-uts` absent, full read-root present). Merging this PR accepts a real shift in Zero's sandbox threat model toward the read-all / write-jail / passthrough-env model:
- **Default profile reads the entire filesystem** (`profile.go` seeds read-roots with `/`); writes stay jailed.
- **Sandboxed commands inherit the full host environment** (`os.Environ()`, no `--clearenv`) — host secrets/tokens are visible to sandboxed commands.
- **Unrestricted FS mode** is now read-**write** `--bind / /` (+ full host `/dev`).
- **`--unshare-ipc` / `--unshare-uts`** dropped (user/pid retained).

These are coherent and align with how comparable agents work, and the network-retry feature itself is well-built (sandboxed-with-granted-network, fresh approval each retry, `require_escalated` forced to a prompt). I'm clearing my review since the two blockers are fixed — flagging the above only so the **merge is a conscious acceptance of the new posture**, not an implicit one.

### Optional nits (unchanged, non-blocking)
- `analyzer.go` `npxUsesNetwork` is effectively `return true` — the loop can be collapsed.
- `engine.go` `require_escalated` in unsafe mode without a grant falls to `deny` (errs safe; a one-line comment would help).

**Verdict: approve.** Solid fixes. Merge at your discretion once you're good with the posture shift above.

PR #302, opened 2026-06-22
fix(sandbox): grant ancestor metadata so cd into the workspace works

gnanam1990:
## Problem

On macOS the seatbelt sandbox grants `file-read*` to each read root via `(subpath "<root>")`, which covers the root and everything **under** it — but **not its parent directories**. Resolving an absolute path to a deeply-nested workspace (e.g. `cd /Users/me/Downloads/app`) must `stat` each ancestor (`/Users`, `/Users/me`, `/Users/me/Downloads`) to walk down to it. Those parents aren't granted, so seatbelt denies the `chdir` — and macOS surfaces that denial as the **misleading `cd: <path>: Not a directory` (ENOTDIR)**, not a clear permission error.

This blocked a sandboxed shell from `cd`-ing into its own workspace: `cd /abs/workspace && go test ./...` failed with "Not a directory" even though the path is a valid directory. Commands that ran relative to the pre-set working directory (`find`, `python3 -c …`) worked, which made the failure baffling.

## Fix

Grant `file-read-metadata` + `file-test-existence` on the **`path-ancestors`** of each read root, so path resolution can traverse to a deeply-nested root. This is **metadata only** — ancestor directory *contents* stay unreadable, write confinement is unchanged, and it stays correctly subordinate to explicit `DenyRead` rules (last-match-wins). Same pattern already used for `/opt/homebrew`, `/usr/local`, `/System/Volumes/Data/private`.

A follow-up commit skips filesystem/volume roots: `(path-ancestors "/")` is invalid SBPL and makes `sandbox-exec` abort (exit 65), so the command couldn't launch at all.

## Verification

- Reproduced: denying an ancestor's metadata produces the exact `cd: …: Not a directory`.
- Fixed: the `path-ancestors` grant makes `cd` succeed; confirmed end-to-end under the real generated profile (`cd /abs/workspace && pwd && ls go.mod`).
- Metadata-only (no content/listing leak); no Linux (landlock/bubblewrap) or Windows changes needed (each is structurally immune); the compat profile compiles under `sandbox-exec`.
- `go build`/`vet`/`test ./internal/sandbox/...` green; regression tests added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS sandbox permissions to support safe traversal for deeply nested workspace read roots using ancestor-metadata (no file-content reads).
  * Broadened macOS sandbox process inspection and signal handling to better match typical tooling needs.
  * Expanded macOS platform read roots to include `/Library/Developer` (Command Line Tools stubs).
  * Scoped git config reads to expected config files and prevented malformed ancestor filters (including `/`).
* **Documentation**
  * Added macOS-specific sandbox guidance for the detected shell runtime.
* **Tests**
  * Added/updated macOS coverage for ancestor-metadata traversal, filesystem root skipping, Command Line Tools coverage, git config scoping, and signal rule expectations.

### DIFF
--- a/internal/sandbox/profile.go
+++ b/internal/sandbox/profile.go
@@ -109,7 +109,28 @@ func permissionProfileReadRoots(workspaceRoot string, policy Policy, scope *Scop
 	if extra := normalizeProfileDirs(policy.AllowRead); len(extra) > 0 {
 		readRoots = dedupeStrings(append(readRoots, extra...))
 	}
-	return readRoots
+	for _, path := range userGitConfigReadPaths() {
+		if normalized := normalizeProfilePath(path); normalized != "" {
+			readRoots = append(readRoots, normalized)
+		}
+	}
+	return dedupeStrings(readRoots)
+}
+
+// userGitConfigReadPaths returns the user's global git config FILES so a
+// sandboxed git can read identity and config (user.name/email, aliases) instead
+// of failing with "unable to access ~/.gitconfig". It is deliberately the config
+// files only — not the ~/.config/git directory, which can hold an XDG credential
+// store — so credentials and the rest of HOME stay unreadable.
+func userGitConfigReadPaths() []string {
+	home, err := os.UserHomeDir()
+	if err != nil || strings.TrimSpace(home) == "" {
+		return nil
+	}
+	return []string{
+		filepath.Join(home, ".gitconfig"),
+		filepath.Join(home, ".config", "git", "config"),
+	}
 }
 
 func normalizeProfileDirs(entries []string) []string {
--- a/internal/sandbox/profile.go
+++ b/internal/sandbox/profile.go
@@ -109,11 +109,6 @@ func permissionProfileReadRoots(workspaceRoot string, policy Policy, scope *Scop
 	if extra := normalizeProfileDirs(policy.AllowRead); len(extra) > 0 {
 		readRoots = dedupeStrings(append(readRoots, extra...))
 	}
-	for _, path := range userGitConfigReadPaths() {
-		if normalized := normalizeProfilePath(path); normalized != "" {
-			readRoots = append(readRoots, normalized)
-		}
-	}
 	return dedupeStrings(readRoots)
 }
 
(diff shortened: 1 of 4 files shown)


Vasanthdev2004:
@coderabbitai review

Vasanthdev2004 (approved):
## Review verdict: ✅ Approve — with two seatbelt widenings to sign off on consciously

The headline fix is excellent and the reasoning is airtight. Two of the bundled changes widen the macOS seatbelt beyond the PR's title, though — they're defensible and bounded, but I'd like them to be a deliberate call rather than a side effect. CI green on all platforms.

### ✅ Core fix — ancestor-metadata traversal (great work)
The `(subpath …)` read filter grants a root + its descendants but not its parents, so resolving an absolute nested path (`cd /a/b/ws`) gets denied at the first ungranted parent and macOS mislabels it `cd: …: Not a directory`. Granting `file-read-metadata file-test-existence` on `(path-ancestors …)` is exactly right: **metadata only** (ancestor *contents* stay unreadable, write-jail unchanged, stays subordinate to `DenyRead`). The `(path-ancestors "/")` → invalid-SBPL / `sandbox-exec` abort edge case is caught and skipped, with a test (incl. the compat-profile path). Well reasoned and well covered.

### ✅ git config scoping + CLT toolchain — security-conscious
- `userGitConfigReadPaths()` grants the config **files** (`~/.gitconfig`, `~/.config/git/config`), deliberately *not* the `~/.config/git` directory (which can hold an XDG credential store), and a test enforces that. Nice.
- `/Library/Developer` for the CLT stub resolution is reasonable and necessary.

### 🟡 Please consciously accept — two grants widen past "cd traversal"
1. **`process-info*` (`(target same-sandbox)` → all)** — a sandboxed command can now inspect *every* process (`ps`/`lsof`), including other processes' argv, which can carry secrets. Read-only, but a real info-disclosure widening.
2. **`signal` (`(target self) (target pgrp)` → `(allow signal)`)** — a sandboxed command can now signal *any* of the user's processes (e.g. kill the user's editor or another terminal), not just its own group. It's bounded by kernel UID ownership (can't touch root/other users), and SBPL has no "same-user" signal target, so `(allow signal)` is the only way to get the cross-pgrp "kill a stale dev server from a previous session" capability you're after — i.e. it's somewhat necessary for the feature. But it does drop a defense-in-depth restriction: a prompt-injected command could now `kill` the user's other work.

Both are macOS-only and defensible; flagging so it's an intentional decision, not an implicit one inherited from a traversal fix. If you'd rather keep the blast radius tight, the traversal + git-config + CLT parts stand alone and could merge without the signal/process-info widening.

### Note (coordination, not a blocker)
This edits `internal/sandbox/runner.go` and `profile.go`, which PR #301 also touches — whichever lands second needs a rebase. (#301 additionally reseeds `permissionProfileReadRoots` with `/`; this PR's ancestor rule already skips `/`, so they're logically compatible.)

**Verdict: approve.** Core fix is clean and well-tested; merge at your discretion once you're comfortable with the signal/process-info widening. Thanks @gnanam1990.

PR #304, opened 2026-06-23
Fix token accounting and usage display

anandh8x:
## Summary
- Normalize token accounting so reasoning and cache tokens are handled as subsets instead of double counted.
- Add output budgeting for high-volume file/search tool results with explicit truncation metadata.
- Stream usage updates into the TUI and show latest-step tokens while preserving cumulative usage/cost reporting.

## Tests
- `go test ./internal/tui`
- `XDG_CONFIG_HOME="$tmp_home/.config" XDG_CACHE_HOME="$tmp_home/.cache" HOME="$tmp_home" go test ./...`

## Summary by CodeRabbit

* **New Features**
  * Live usage event updates display token consumption during active agent runs.
  * Search, glob, and directory tools now show visible truncation indicators when results exceed limits.

* **Improvements**
  * Reasoning tokens are now included in output token billing rather than charged separately.
  * File reading and directory listing tools enforce byte-size output limits to prevent overwhelming results.
  * Enhanced token usage tracking and UI status display for accuracy.

### DIFF
--- a/internal/modelregistry/cost.go
+++ b/internal/modelregistry/cost.go
@@ -92,11 +92,10 @@ func CalculateCost(model ModelEntry, usage zeroruntime.Usage) (CostBreakdown, er
 	if uncachedInputTokens < 0 {
 		uncachedInputTokens = 0
 	}
-	billableOutputTokens := outputTokens + reasoningTokens
 	inputCost := costForTokens(uncachedInputTokens, inputRate)
 	cachedInputCost := costForTokens(cachedInputTokens, cachedRate)
 	cacheWriteCost := costForTokens(cacheWriteTokens, cacheWriteRate)
-	outputCost := costForTokens(billableOutputTokens, outputRate)
+	outputCost := costForTokens(outputTokens, outputRate)
 
 	breakdown := CostBreakdown{
 		ModelID:           model.ID,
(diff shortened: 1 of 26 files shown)


gnanam1990:
@coderabbit approvable ?

gnanam1990 (approved):
## Approve ✅

Read the full diff and verified it on a fresh checkout of the PR head: `go build ./...` green, `go vet` clean, full suite **69 packages ok / 0 fail** (incl. the changed accounting/tools/tui/provider packages).

**Why this is solid:**
- **Reasoning-as-subset normalization is correct, and the parsers are fixed to match.** Gemini folds thoughts into output (`candidates + thoughts`); the OpenAI-style parser now reads `completion_tokens_details.reasoning_tokens` (its `completion_tokens` already includes reasoning). This removes a real double-count in both `TotalTokens` and billing (`cost.go`).
- **Fixes the unpriced-model footer** showing the raw cumulative — `m.lastUsage` is set before `tracker.Record` errors out, so the latest-step readout stays bounded.
- **Output budgeting** on read/search tools is a sensible context-bloat guard with explicit truncation metadata.
- **Live per-turn usage de-dup** (`liveUsageCounts` skipping already-recorded events in the end-of-run replay) is correct and won't double-count.
- Good test coverage across the changed paths.

**One non-blocking follow-up:** `mergeUsageSnapshot` switches usage from accumulate to *latest-non-zero-per-field* (snapshot semantics). Correct for single-snapshot/cumulative emitters; please just confirm no provider — including openai-compatible gateways — emits incremental usage deltas, which would silently undercount.

Nice work.

PR #310, opened 2026-06-24
fix(swarm): make swarm_collect wait for members instead of polling

Vasanthdev2004:
## Problem

Spawning swarm subagents felt broken: they appeared to "never finish", and the chat filled with repeated `swarm_status`/`swarm_collect` cards.

**Root cause:** `swarm_spawn` is fire-and-forget and there is **no completion event** — the only way to learn a member finished was to call `swarm_status`/`swarm_collect`, which returned an **instant snapshot** and never waited. So the orchestrator had no choice but to **poll in a loop**: that's the status-check spam, and why it never reaches a clean "done."

(Studied a mature reference agent for comparison — its delegation **blocks** and returns results when children finish, rather than exposing a poll-only status tool. This PR brings the same model to our `swarm_collect`.)

## Fix — `swarm_collect` now waits

- **`Coordinator.WaitSettled(ctx, team)`** blocks until the scope has no pending/running task, then returns the snapshot. A change-broadcast channel (closed on every register/setstatus/finish/reassign) wakes waiters with no polling. Empty/unknown team settles immediately.
- **`Swarm.CollectWait`** wraps it; the **collect tool** calls it bounded by:
  - the **run context** (so Esc / user-cancel returns immediately), and
  - a **timeout** — default 600s, cap 1800s, overridable via `timeout_seconds` — so a genuinely stuck member returns **partial** state instead of hanging.
- **Tool descriptions** now steer the model: *spawn all members, then call `swarm_collect` once to wait for results; `swarm_status` is a one-shot snapshot, don't poll it in a loop.*

## Fix — chat de-spam

`collapseRepeatedStatusCard` drops the previous identical swarm status/collect card when a check repeats **verbatim back-to-back**, so any residual re-checks don't stack duplicate blocks. It only fires on an exact `[call][identical result][call]` tail — never removes a card with intervening content or a changed state.

## Tests

- `TestCollectBlocksUntilMembersFinish` — collect does not return while a member runs; returns the result once it finishes.
- `TestCollectReturnsPartialOnTimeout` — a stuck member + small `timeout_seconds` returns the running snapshot instead of hanging.
- `TestCollapseRepeatedStatusCard{DropsDuplicate,KeepsChangedState,IgnoresInterveningContent}`.
- Existing `TestStatusAndCollectTools` still green (it waits for done before collecting, so blocking is a no-op there). Full `./internal/swarm/...` + `./internal/tui/...` suites green.

## Follow-up (not in this PR)

Clickable **drill-in subchat** for individual swarm members — they're runtime-only (no session), so that needs per-member sessions wired into the existing subchat machinery. Tracked separately; this PR fixes the completion/spam loop first.

## Summary by CodeRabbit

* **New Features**
  * Added a wait-based collect experience so team task results can be gathered after work finishes, with optional timeout support.
  * Improved status polling behavior to return partial results when time runs out instead of waiting indefinitely.

* **Bug Fixes**
  * Reduced repeated status/collect entries in the TUI transcript, making chat output shorter and easier to read.
  * Prevented duplicate status blocks from appearing when the same check is repeated.

### DIFF
--- a/internal/swarm/coordinator.go
+++ b/internal/swarm/coordinator.go
@@ -1,6 +1,7 @@
 package swarm
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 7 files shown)


Vasanthdev2004:
@coderabbitai approve

PR #313, opened 2026-06-24
feat(swarm): let swarm members build (write + sandboxed shell)

Vasanthdev2004:
## Problem

Spawning swarm subagents to **build** didn't work: the model reported *"the swarm subagents couldn't directly edit files (their tool set was read-only)"* and the orchestrator did all the writes itself.

A 4-agent investigation found the real cause (not model laziness, not the sandbox): a member runs **headless** at `--auto low` → `PermissionModeAuto`, and `ToolAdvertised` (`internal/agent/loop.go:2368`) strips every prompt-requiring tool from the model's list. So `write_file`/`edit_file`/`apply_patch`/`bash` were **never handed to the member** — it only saw `read_file`/`grep`/`glob`/`list_directory`/`update_plan`. The member literally couldn't build. (`--enabled-tools` *includes* the write tools, but enabled ≠ advertised.)

## Fix

New **`PermissionModeMemberAuto`**: a headless mode that **advertises** the in-workspace mutators (`SideEffectWrite` + `SideEffectShell`) on top of the Auto set, and behaves **exactly like Auto everywhere else**.

| File | Change |
|------|--------|
| `agent/types.go` | new `PermissionModeMemberAuto` |
| `agent/loop.go` | `ToolAdvertised` advertises write/shell tools for member-auto (not network, not denied) |
| `cli/exec_tools.go` | `--auto member` → `PermissionModeMemberAuto` |
| `specialist/exec.go` | `MemberAutonomy` flag → emits `--auto member` for a non-unsafe member |
| `swarm/launcher_specialist.go` | swarm members set `MemberAutonomy: true` |

## Why it's safe (no authority widening)

`member-auto` **normalizes to `auto`** at the sandbox layer (`sandbox/normalize.go:16`), so the sandbox engine (`engine.go` `Evaluate`) is the boundary, unchanged:
- ✅ in-workspace writes auto-allow (`workspace write is allowed`), sandbox-backed shell auto-allows.
- ❌ out-of-workspace writes, network, and destructive commands → prompt → **denied** (headless has no approver). Deny-lists and protected-metadata blocks still apply.

This grants exactly what an **interactive auto agent** already does inside the sandbox — nothing more.

**Scope:** swarm members only. Task-tool specialists are unchanged (still `--auto low`/read-only). Unsafe orchestrators still give members full `--auto high`. Members can't spawn their own write-capable swarm (`--tag specialist` + autonomy gate).

**Windows caveat:** shell only auto-runs when the OS sandbox is active; without `zero sandbox setup`, a member writes files but can't run builds.

## Verification

- Tests: `ToolAdvertised` member-auto (advertises write/shell, not network/denied); `memberAwareAutonomy` mapping; `BuildArgs` emits `--auto member` for members and stays `--auto low` for specialists; `resolveExecPermissionMode("member")`; exec tool-list includes `write_file` for member-auto and hides it for plain Auto.
- Full `agent` + `specialist` + `cli` + `swarm` + `sandbox` suites green; `go vet` clean.
- **Adversarial security review** (3 dimensions × independent verifiers — escalation / scope-leak / mode-confusion): **0 confirmed findings**.

## Related

Pairs with #312 (keep finished members inspectable): you can now click a member to confirm it actually wrote files.

## Summary by CodeRabbit

* **New Features**
  * Added a new **member-auto** autonomy mode for headless member runs, enabling more in-workspace actions within sandboxed limits.
  * Swarm member launches now pass member-aware autonomy into child execution so member specialists run with the intended capability level.
* **Bug Fixes**
  * Updated tool visibility so **member-auto** only advertises tools allowed for that mode (hiding network and denied tools).
  * Improved autonomy resolution to recognize the **member** level consistently, and ensured **unsafe** members remain at the highest setting.
* **Tests**
  * Added unit tests covering tool advertisement, autonomy mapping, and member-aware specialist argument emission.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -2368,6 +2368,21 @@ func ToolAdvertised(tool tools.Tool, permissionMode PermissionMode) bool {
 	if permissionMode == PermissionModeAuto {
 		return tool.Safety().Permission == tools.PermissionAllow || tool.Safety().AdvertiseInAuto
 	}
+	if permissionMode == PermissionModeMemberAuto {
+		// Like Auto, plus the in-workspace mutators a headless member needs to
+		// build. The sandbox engine still decides at call time: in-workspace writes
+		// and sandbox-backed shell auto-allow, while out-of-workspace writes,
+		// network, and destructive commands prompt → denied headless. So this
+		// advertises capability without widening sandbox authority.
+		if tool.Safety().Permission == tools.PermissionAllow || tool.Safety().AdvertiseInAuto {
+			return true
+		}
+		switch tool.Safety().SideEffect {
+		case tools.SideEffectWrite, tools.SideEffectShell:
+			return true
+		}
+		return false
+	}
 	return true
 }
 
--- a/internal/agent/member_auto_test.go
+++ b/internal/agent/member_auto_test.go
@@ -12,10 +12,12 @@ type safetyTool struct {
 	safety tools.Safety
 }
 
-func (t safetyTool) Name() string            { return t.name }
-func (t safetyTool) Description() string      { return "test tool" }
-func (t safetyTool) Parameters() tools.Schema { return tools.Schema{Type: "object", AdditionalProperties: false} }
-func (t safetyTool) Safety() tools.Safety     { return t.safety }
+func (t safetyTool) Name() string        { return t.name }
+func (t safetyTool) Description() string { return "test tool" }
+func (t safetyTool) Parameters() tools.Schema {
+	return tools.Schema{Type: "object", AdditionalProperties: false}
+}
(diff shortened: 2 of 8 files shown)


Vasanthdev2004:
Note on the `--auto member` doc nitpick: `member` is an **internal** autonomy token set only by the swarm launcher (`internal/swarm/launcher_specialist.go` → `MemberAutonomy`), not a user-facing value. It resolves to `PermissionModeMemberAuto`, which is sandbox-confined to in-workspace writes + sandboxed shell (it normalizes to `auto` at the sandbox layer). The public `--auto <low|medium|high>` help and the invalid-value error intentionally omit it so users aren''t steered toward an internal mode. The parser accepts it because the headless member child is launched via the same `exec` entrypoint.

PR #314, opened 2026-06-24
Fix shared temp roots in sandbox scope

anandh8x:
## Summary
- Treat platform temp directories as default writable roots across sandbox scope and permission profiles
- Bind host temp roots on Linux instead of mounting an isolated tmpfs at /tmp, so tool-created files remain visible to sandboxed commands
- Update scope, CLI, tool, and agent tests so platform temp is no longer treated as outside the allowed write roots

## Tests
- git diff --check
- GOCACHE=/tmp/zero-go-cache go test ./internal/sandbox ./internal/tools ./internal/cli ./internal/agent

Note: the focused Go suite was rerun outside the restricted sandbox because existing local socket and bwrap tests need host socket permissions.

## Summary by CodeRabbit

* **New Features**
  * Sandbox and tool permission checks now recognize additional host temporary locations as writable (including system temp roots and `TMPDIR`/`TEMP`/`TMP`), reducing false denials for common temp-file use.

* **Bug Fixes**
  * Improved handling of “outside the workspace” paths so allow/deny decisions and user-facing errors are more consistent.
  * Made sandbox bind behavior more reliable and deterministic on Linux, stabilizing which temp roots get access when multiple are involved.

### DIFF
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -2075,7 +2075,7 @@ func TestRunEmitsPermissionEventForPersistentSandboxGrant(t *testing.T) {
 
 func TestRunPromptsAndAllowsOutsideWorkspaceWrite(t *testing.T) {
 	root := t.TempDir()
-	outside := filepath.Join(t.TempDir(), "escape.txt")
+	outside := filepath.Join(tempDirOutsideDefaultTemp(t), "escape.txt")
 	engine := sandbox.NewEngine(sandbox.EngineOptions{
 		WorkspaceRoot: root,
 		Policy:        sandbox.DefaultPolicy(),
@@ -2137,7 +2137,7 @@ func TestRunPromptsAndAllowsOutsideWorkspaceWrite(t *testing.T) {
 
 func TestRunSessionAllowsLaterOutsideWorkspaceWrite(t *testing.T) {
 	root := t.TempDir()
-	outside := filepath.Join(t.TempDir(), "escape.txt")
+	outside := filepath.Join(tempDirOutsideDefaultTemp(t), "escape.txt")
 	engine := sandbox.NewEngine(sandbox.EngineOptions{
 		WorkspaceRoot: root,
 		Policy:        sandbox.DefaultPolicy(),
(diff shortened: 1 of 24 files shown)


gnanam1990 (approved):
## Review — approving ✅

Clean, coherent, well-tested fix. All CI green (Smoke on ubuntu/macOS/windows, Security & code health, CodeRabbit, Zero Review).

### What it does
Makes platform temp dirs (`/tmp` + `$TMPDIR` on Unix, `%TEMP%`/`%TMP%` on Windows) default writable roots consistently across all three layers — scope (`NewScope`), permission profile (`permissionProfileRoots`), and the Linux bwrap args (bind host temp instead of an isolated `--tmpfs /tmp`). `engine.go` routing the fallback scope through `NewScope` fixes a prior inconsistency.

### The core tradeoff (intended, and correct)
On Linux this shares host `/tmp` rather than a private tmpfs — a deliberate reduction in isolation so files created outside the sandbox stay visible to sandboxed commands. Reasonable: `/tmp` is world-writable anyway, the write-confinement guarantee (workspace + temp) is preserved, and `normalizeScopeRoot` explicitly refuses the filesystem root `/` as a write root (plus requires exist + is-dir), so the obvious `TMPDIR=/` escape is blocked.

### Verified
- `TMPDIR=/` rejected (filesystem-root guard in both `normalizeScopeRoot` and `normalizeProfileDirs`).
- Non-existent / non-dir temp paths filtered out; `/tmp` symlinks resolved via `EvalSymlinks`.
- Unconditional temp injection tramples no intent — every profile already sets `AllowTemp: true`.
- `workspace.go` change is pure error-message consistency (`outsideWorkspaceError`); the file-write tools still confine to the workspace and do **not** gain `/tmp` access — only sandboxed shell/exec does. Good separation.
- `linuxSortedWriteRoots` (shallow-first by `pathDepth`) makes nested binds deterministic.

### Minor nit (non-blocking, follow-up)
Unconditional trust of `$TMPDIR`/`$TEMP`/`$TMP`: a broad-but-valid value (e.g. `TMPDIR=$HOME`) silently widens the writable scope. Low risk (user-controlled env; `/` and non-dirs already rejected), but a defensive skip when the temp candidate is an ancestor of the workspace or equals `$HOME` would be a nice hardening.

Test coverage is thorough (18 files across sandbox/cli/tools/agent, incl. Windows + Darwin seatbelt). LGTM.

PR #322, opened 2026-06-26
Align sandbox command approval behavior

anandh8x:
## Summary
- run prefix-approved shell commands through the escalated sandbox path when denied-read policy allows it
- allow auto sandbox mode to degrade to a direct command plan when native command wrapping is unavailable
- keep strict sandbox require fail-closed and update regression coverage for degraded execution

## Testing
- GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/sandbox
- GOCACHE=/tmp/zero-go-cache go test ./internal/tools -run 'TestRegistryRunsWithDegradedUnavailableNativeSandbox|TestBashToolRunsWithDegradedUnavailableNativeSandbox|TestBashRequireEscalatedBypassesNativeSandboxAfterApproval|TestExecCommandRequireEscalatedBypassesNativeSandboxAfterApproval'
- GOCACHE=/tmp/zero-go-cache go test ./... -run '^$'
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Shell commands now degrade gracefully when the native sandbox backend is unavailable, continuing execution instead of failing.
  * Approved shell command prefixes now apply reliably across segmented command chains (pipes/conditionals), avoiding unnecessary repeat prompts.
  * Prefix approvals that bypass sandboxing now correctly attach escalated sandbox permission metadata (e.g., `require_escalated`), improving outcomes in unavailable-sandbox scenarios.
* **Documentation**
  * Updated the sandbox permission request policy and clarified `sandbox_permissions`/narrow `prefix_rule` guidance in tool schema descriptions.
* **Tests**
  * Expanded coverage for segmented prefix approval and degraded-unavailable sandbox behavior.


Vasanthdev2004 (approved):
## Review: Align sandbox command approval behavior

Read the full diff and traced the three behavioral changes through `command_prefix.go`, `loop.go`, `manager.go`, `runner.go`, and the tool schemas. This is a careful, well-tested change and the test updates match the new behavior faithfully. Approving — with two posture points called out below so they're consciously owned, plus one minor note.

### What this does (for the record)
1. **Segmented prefix matching** — a prefix grant now spans `|`, `&&`, `||`, `;` chains. Each segment must be covered by the prefix *or* be a `knownSafeCommandSegment`. The allowlist is read-only and the non-trivial entries (`git` status/log/diff/show/branch, `find`, `rg`, `sed -n`, `base64`) are properly scoped — `-exec`/`-delete`/`--pre`/`-o`/`--output`/`--exec`/`-C`/`--upload-pack` are all rejected, and glob tokens (`*?[]{}`) no longer parse as literals. Solid.
2. **Degrade instead of fail-closed** — when the native sandbox is unavailable under the **auto** preference (the default), a command now runs as an `EnforcementDegraded` direct plan instead of erroring. `Require` still fails closed (`manager.go:155`, `TestSandboxManagerFailsClosedWhenNativeRequiredAndUnavailable` now pins `Preference: Require`).
3. **Prefix-approval escalation** — a prefix-approved shell command is rewritten to `require_escalated` so it doesn't re-deny/re-prompt.

### Posture to own consciously (not blocking)
The crux here is **fail-closed → degrade** on the default (auto) preference. On a host with no working native sandbox (Windows pre-`zero sandbox setup`, Linux without bwrap/landlock, etc.), a command that previously got blocked now **runs unsandboxed after the permission prompt**. That's defensible — the prompt remains the gate, the sandbox was unavailable either way, and `Require` is the escape hatch for anyone who needs hard confinement — but it does weaken the no-sandbox default from "block" to "run after approval." Flagging it so it's an intentional decision, not a silent one. The prompt still fires (`TestRunPromptsForDestructiveShellInsteadOfSandboxDeny` keeps the prompt and now asserts the command actually runs), so this isn't auto-approval.

### One consistency question
`UnsandboxedExecutionAllowed()` (false when the policy sets `DenyRead`) gates the escalation and the denied-retry paths — `loop.go:1062`, `bash.go:183` — but the `manager.go` degrade itself is **not** gated on it. With the default policy this is moot (`DefaultPolicy()` sets no `DenyRead`, so it's always `true`). But a user who opts into `DenyRead` paths *and* runs on a sandbox-less host would get a degraded, unsandboxed command that can read those denied paths — the opposite of what the guard enforces on the escalation path. Was that intentional? If not, gating the degrade on `UnsandboxedExecutionAllowed()` (fall back to fail-closed when `DenyRead` is set) would make the two paths agree. At minimum worth a comment noting `DenyRead` implies `Require`.

### Minor
The unconditional read-only entries in `knownSafeCommandSegment` (`cat`, `grep`, `tail`, `head`, …) are path-unconstrained, so a narrow prefix grant (e.g. `git status`) can silently cover a chain whose safe tail reads outside the workspace (`git status && cat <path>`). Read-only and within the agent's existing read surface, so low severity — just noting the "safe tail" isn't workspace-scoped.

CI is green. Nice coverage on the segmented-matching edge cases (uncovered-tail rejection, requested-prefix-must-cover-all-segments). Approving.

PR #323, opened 2026-06-26
Improve sandbox permission handling

anandh8x:
## Summary
- retry shell commands outside the sandbox when sandboxed output only reflects the sandbox namespace
- keep network-related retries sandboxed when a scoped network grant is sufficient
- show user-facing permission reasons for escalated shell commands instead of internal sandbox policy text
- expand tool schema/prompt guidance and regression coverage for sandbox permission flows

## Tests
- env GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/tui -run 'TestRunUsesEscalatedJustificationForPermissionPrompt|TestRunUsesUserFacingEscalatedFallbackForPermissionPrompt|TestPermissionPromptMapsEscalatedSandboxReason|TestRunRetriesShellUnsandboxedAfterSandboxNamespaceLimitedOutput'
- env GOCACHE=/tmp/zero-go-cache go test ./internal/tools
- env HOME=/tmp/zero-test-home-clean XDG_CONFIG_HOME=/tmp/zero-test-config-clean GOCACHE=/tmp/zero-go-cache go test ./...

## Summary by CodeRabbit

* **New Features**
  * Improved guidance for sandbox-related permission prompts, including clearer wording for when elevated access is needed.
  * Added support for recognizing when sandboxed command output may hide host/global state and prompting appropriately.

* **Bug Fixes**
  * Prevented misleading responses when sandbox-limited output does not reflect the full system state.
  * Updated permission messages so user-facing text stays clear and doesn’t expose internal sandbox details.


Vasanthdev2004 (approved):
## Review: Improve sandbox permission handling

Read the full diff. This is a clean follow-on to the prior sandbox-approval work — the user-facing reason mapping is a real improvement, the namespace-limited retry solves a genuine papercut (sandboxed `ps`/process inspection returning only the bwrap PID namespace), and the test coverage is thorough (justification-as-reason, internal-reason-not-leaked, namespace retry discards the misleading output, TUI mapping). Approving with two non-blocking points.

### Not a security concern (checked)
The namespace-limited retry only ever *offers* an unsandboxed rerun behind a permission prompt, and it's already gated by `UnsandboxedExecutionAllowed()` (so `DenyRead` policies still block it) and skipped under an existing `require_escalated`. It grants no capability the model doesn't already have: the engine returns `ActionPrompt`/`ReasonEscalatedSandboxRequired` for any `require_escalated` command (engine.go:366), so the model can already request an unsandboxed run directly. The retried command is the same args the user sees in the prompt. So this is correctness/UX, not a new escalation path.

### 1. Trigger reads model-influenceable stdout (non-blocking)
`sandboxNamespaceLimitedShellResult` decides whether to surface the "rerun outside the sandbox" prompt by string-matching the command's **output** for `"bwrap "`, `"--unshare-pid"`, and `"-- /bin/sh -c"`. Two downsides:
- **Fragile**: it couples to bwrap's current CLI flags. If the wrapper invocation changes, detection silently stops firing (and the user is back to the misleading output) — no test would catch the drift.
- **Gameable / false-positive**: any command whose stdout happens to contain those tokens (`echo`-ing a bwrap line, `cat`-ing a script, `ps` showing an unrelated bwrap) trips the heuristic and surfaces a "sandbox limited — rerun outside" prompt whose reason may not be the real one. Harmless given the gating above, but it's a misleading prompt for a visible command.

Both go away if the signal comes from the runner instead of stdout: when bash/exec wraps a command under `--unshare-pid`, have it set a `Meta` flag (e.g. `sandbox_pid_namespaced=true`) and key the retry on that. You already thread `sandbox_backend`/`sandbox_target_backend`/`sandbox_wrapped` through `Meta` (bash.go:252-254), so this fits the existing plumbing and is neither spoofable nor coupled to flag strings.

### 2. Reason string duplicated across packages (non-blocking)
`sandboxNamespaceLimitedReason` is an unexported const in `agent`, but `tui/transcript.go:523` hardcodes the same literal to map it to display text. That's the producer→display contract living in two places — change one and the TUI mapping silently falls through to the raw string. The codebase already has the right pattern for exactly this: `sandbox.ReasonEscalatedSandboxRequired` is an exported constant both sides reference. Move this reason into the `sandbox` package (or a shared spot) the same way so there's a single source of truth.

Note: CI was still mid-run at review time (Security & code health + Performance Smoke green; Smoke matrix, Zero Review, and CodeRabbit pending) — worth a glance before merge.

PR #324, opened 2026-06-26
Move deferred tool discovery into tool search

anandh8x:
## Summary
- move deferred-tool discovery out of per-turn user messages and into the tool_search tool definition
- keep unloaded deferred tools hidden while preserving on-demand loading through tool_search
- keep tool_search reachable for allowlisted deferred tools and preserve loaded/deferred state across retry paths
- remove leaked deferred-notice wording from eager-tool redirect output

## Tests
- env GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/tools -run 'Test.*Deferred|Test.*ToolSearch|TestRunLoadsDeferredToolThenAdvertisesNextTurn|TestRunReactiveRetryKeepsLoadedDeferredToolAndDiscovery|TestRunConnectTimeReactiveRetryKeepsLoadedDeferredToolAndDiscovery|TestAllowlistedDeferredToolsKeepToolSearchReachable|TestDisabledToolSearchFallsBackToEager'
- env GOCACHE=/tmp/zero-go-cache go test ./internal/tools
- env GOCACHE=/tmp/zero-go-cache go test ./internal/agent ./internal/tools ./internal/cli
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Improved deferred tool discovery in `tool_search`, with clearer source-based grouping and exact-name (`select:...`) selection hints when hidden tools are available. Tool search guidance text was also refined.
* **Bug Fixes**
  * Removed leftover deferred discovery/reminder text from request history during proactive and reactive retry flows, while keeping deferred tool exposure behavior unchanged.

### DIFF
--- a/internal/agent/deferred_loop_test.go
+++ b/internal/agent/deferred_loop_test.go
@@ -129,6 +129,24 @@ func (fakeToolSearchTool) Run(_ context.Context, _ map[string]any) tools.Result 
 	return tools.Result{Status: tools.StatusOK, Output: "ok"}
 }
 
+func toolDefinitionByName(definitions []zeroruntime.ToolDefinition, name string) *zeroruntime.ToolDefinition {
+	for i := range definitions {
+		if definitions[i].Name == name {
+			return &definitions[i]
+		}
+	}
+	return nil
+}
+
+func assertNoDeferredDiscoveryMessage(t *testing.T, request zeroruntime.CompletionRequest) {
+	t.Helper()
+	for _, message := range request.Messages {
+		if message.Role == zeroruntime.MessageRoleUser && strings.Contains(message.Content, "Deferred tools:") {
+			t.Fatalf("deferred discovery must not be appended as a user message: %q", message.Content)
+		}
+	}
+}
+
 func TestPartitionToolsInactiveIsByteIdenticalAndDropsToolSearch(t *testing.T) {
 	root := t.TempDir()
 	registry := tools.NewRegistry()
@@ -219,8 +237,8 @@ func TestPartitionToolsBelowThresholdInactive(t *testing.T) {
 // FIX 6(a): a deferred tool removed by DisabledTools must NOT count toward the
 // eligible total. With N deferred registered and threshold == N, disabling one
 // drops the surviving eligible count to N-1 < threshold, so the partition takes
-// the INACTIVE path (empty reminder, all VISIBLE tools exposed with full schemas,
-// the disabled one filtered out entirely).
+// the INACTIVE path (empty discovery text, all VISIBLE tools exposed with full
+// schemas, the disabled one filtered out entirely).
 func TestPartitionToolsDisabledDeferredDropsBelowThresholdInactive(t *testing.T) {
 	registry := tools.NewRegistry()
 	registry.Register(fakeDeferredTool{name: "mcp__srv__alpha", desc: "alpha"})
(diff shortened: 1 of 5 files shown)


gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable then approve it

gnanam1990:
## Review — moving deferred discovery into tool_search

Good direction: dropping the per-turn `<system-reminder>` user message removes recurring context noise, and folding discovery into the `tool_search` definition is cleaner. Builds green; `internal/agent` + `internal/tools` pass `-race`. A few things worth a look before merge.

### 1. Discoverability regression (main one)
The old reminder listed **each** deferred tool by name+description (`DeferredLine`), so the model saw the exact names to pass to `select:`. The new `BuildToolSearchDescription` lists only **sources** (server / family prefix) — not tool names. But the text still says *"Use query `select:Name1,Name2` when you know exact tool names, or keywords"* — the model no longer knows the names, so the `select:` path is mostly dead on first discovery and it must keyword-search blind.

This likely makes loading the right deferred tool **harder for weaker/local models** (which is exactly where tool discovery is fragile). Is dropping the explicit name list intentional (pure context-saving), or should the description keep a compact `name — source` list (or at least lead with the keyword path)? That's the one I'd want your intent on.

### 2. Prompt-cache churn (medium)
`tool_search`'s `Description` is now rebuilt per turn from `hiddenTools` and lives in the **tools block** — part of the cacheable prefix for most providers. When the source set changes (a deferred tool gets loaded and its source drops out), the tools prefix changes and busts the cache. The old approach kept tool defs static and varied only a trailing user message (already in the uncached tail). Impact is bounded (sources change infrequently), but on cached providers this trades some cost. Worth confirming the description is stable turn-to-turn in steady state.

### 3. Self-contradictory instruction (low)
Tie-in with #1: *"`select:Name1,Name2` when you know exact tool names"* reads oddly when names aren't shown. Lead with keywords, mention `select:` as the follow-up once a search surfaces names.

### 4. Base description (low)
`NewToolSearchTool` → `BuildToolSearchDescription(nil)` → *"…sources:\nNone currently enabled."* as the static default. On the inactive path tool_search is dropped (rarely shown), and after all deferred tools load the active path leaves this base text — "None currently enabled" is then accurate-ish. Minor, just flagging.

### Looks correct
- Retry/compaction paths reuse `exposed` (which carries the discovery in the tool_search def), so loaded/deferred state **and** discovery are preserved without re-appending anything — the reminder-not-persisted invariant is now moot. 👍
- `DeferredSource` (MCP server name, else first `_` segment for families like `swarm_*`) is reasonable.
- Removing the leaked "deferred-tools notice" wording from `alreadyAvailableMessage` is a good cleanup.
- Tests updated and green.

Net: mergeable; #1 is the only one I'd resolve first (intent on whether weak-model discoverability is acceptably preserved).

gnanam1990 (changes requested):
Requesting changes on one substantive point (the rest is non-blocking — see my detailed review comment above).

**#1 discoverability:** the new `tool_search` description exposes deferred tool **sources** (server/family), not **names**, yet still instructs `select:Name1,Name2` "when you know exact tool names" — which the model can no longer see. That's a self-contradictory instruction and a likely discovery regression for weaker/local models (the keyword path becomes the only real route, blind). Please either keep a compact `name — source` list in the description, or reword to lead with keyword search and drop the `select:`-by-name framing.

Non-blocking but worth a look: prompt-cache churn (the per-turn-varying description now sits in the tools block) and the "None currently enabled" base text. Build + `internal/agent`/`internal/tools` -race are green; happy to re-review quickly once #1 is settled.

gnanam1990 (approved):

PR #325, opened 2026-06-27
fix(agent): gate headless run completion on honesty (no false-success on no-tool-call / self-reported-incomplete turns)

gnanam1990:
## Summary
Two fixes to the **headless** agent loop so a run is reported `success` only when it actually finished (surfaced during headless-exec testing). Both are gated by a new `Options.RequireCompletionSignal`, set only for headless `zero exec` — **default off, so the interactive TUI is byte-identical.**

### Bug #1 — premature loop termination (deterministic)
A turn that produced text but **no tool call** was always accepted as the final answer, so the loop reported `success` even mid-task. Now a no-tool-call turn is **not** treated as done while work clearly remains — pending `update_plan` items, or a message ending on a continuation cue ("…Let me check the config:") — the loop re-prompts to continue (bounded by `maxContinueNudges`, still by `MaxTurns`), and finalizes **INCOMPLETE (exit 4)** rather than `success` if it keeps stalling.

**Behavioral proof:** on a multi-step server-setup task the agent went from **3 tool calls** (it stopped on "…Let me check the SSH configuration:") **→ 50 tool calls** — it now starts sshd, completes the SSH clone, and works its plan instead of declaring victory after step 2.

### Bug #2 — `success` without a task-grounded check
**Deterministic, unit-tested:**
- **self-report downgrade** — if the final message admits the model guessed / could not meet the objective (first-person inability *stems* generalized over verb/tense, plus guess/fallback phrases, with a guard so success-y negations like "couldn't find any issues" / "cannot reproduce" are not misread) → **INCOMPLETE**, never `success`.
- **plan gate** — pending / in_progress `update_plan` items at termination → **INCOMPLETE**.

**Advisory (NOT a guarantee):**
- **task-grounded acceptance** — when `--self-correct` is on, one bounded acceptance pass re-derives the task's stated criterion and discourages three false-success patterns (well-formed==correct, existing-tests-pass==objective-met, result==baseline-it-was-told-to-beat). This **reduces but does not eliminate** false-success: a model that confidently claims "PASS, all requirements met" can still slip through, because there is no general oracle to verify correctness against a task's hidden criterion. This is a **fundamental limit, not a tuning gap** — documented honestly rather than overclaimed.

**Behavioral proof:** an image-grounded task a text-only model can't actually solve flipped from **false `success` (reward 0) → honest INCOMPLETE** once its "I cannot analyze the image" admission is detected.

## Scope
Agent loop + headless exit wiring + tests only:
- `internal/agent/{loop,guardrails,types}.go`
- `internal/cli/exec.go` — `exitIncomplete=4`, `RequireCompletionSignal` wiring, INCOMPLETE `run_end` + reason
- `internal/agent/{completion_gate,acceptance_gate}_test.go`

## Verification
`make build`, `go vet ./...`, `make lint`, `go test ./... -race` — all green (70 pkg, 0 fail, 0 races). Gate-off tests confirm interactive/default behavior is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Headless runs now report distinct **incomplete** vs **success** outcomes, including an optional **incomplete reason**.
  * Added stricter headless completion gating, with optional task-grounded acceptance challenge for self-correction.
* **Bug Fixes**
  * Prevents stalled or mid-step “continue” situations from being marked as success by injecting bounded re-prompts.
  * Treats self-reported inability (and certain “can’t/don’t know” phrasing) as incomplete, respects pending plan work, and marks **max-turns** cutoff as incomplete.
  * Preserves legacy behavior when the gate is disabled.
* **Tests**
  * Added acceptance-gate and completion-gate regression coverage for these scenarios.

### DIFF
--- a/internal/agent/acceptance_gate_test.go
+++ b/internal/agent/acceptance_gate_test.go
@@ -1,0 +1,212 @@
+package agent
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+func enabledSelfCorrector(t *testing.T) *SelfCorrector {
+	t.Helper()
+	// checker/verifier nil → no LSP / project-test pass runs; we only need a
+	// non-nil SelfCorrector to enable the task-grounded acceptance gate.
+	return NewSelfCorrector(t.TempDir(), nil, nil, SelfCorrectConfig{Enabled: true})
+}
+
+// BUG #2(a): a final message that admits the model guessed / could not meet the
+// objective must be reported as INCOMPLETE, never success — and immediately
+// (an admitted guess is not worth re-prompting).
+func TestAcceptanceSelfReportAdmissionIsIncomplete(t *testing.T) {
+	provider := &mockProvider{turns: [][]zeroruntime.StreamEvent{
+		textTurn("I couldn't analyze the board image, so I wrote the common opening move e2e4 as my best guess."),
+	}}
+
+	result, err := Run(context.Background(), "write the best chess move", provider, Options{
+		Registry:                tools.NewRegistry(),
+		MaxTurns:                10,
+		RequireCompletionSignal: true,
+		SelfCorrect:             enabledSelfCorrector(t),
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !result.Incomplete {
+		t.Fatalf("admitted guess must be Incomplete, got success; final=%q", result.FinalAnswer)
+	}
+	if !strings.Contains(result.IncompleteReason, "objective was not met") {
+		t.Fatalf("IncompleteReason = %q, want it to cite the admission", result.IncompleteReason)
+	}
+	if len(provider.requests) != 1 {
+		t.Fatalf("an admission should downgrade immediately (1 turn), got %d", len(provider.requests))
+	}
+}
+
+// BUG #2(a): the inability detector must generalize over the verb the model uses
+// (so re-phrasing can't defeat it) while NOT misreading success-y negations
+// ("could not find any issues") as admissions.
+func TestSelfReportedIncompletionMatching(t *testing.T) {
+	cases := []struct {
+		text  string
+		admit bool
+	}{
+		// real chess re-run wording that slipped a fixed phrase list:
+		{"…which I cannot do without proper image analysis capabilities.", true},
+		{"I cannot visually analyze the chess board image.", true},
+		{"I could not recover the original values, so this is my best guess.", true},
+		{"I was unable to complete the optimization.", true},
+		{"The result may not be correct.", true},
+		// genuine successes that must NOT be flagged:
+		{"I implemented the function and all tests pass.", false},
+		{"I verified the fix; I could not find any remaining issues.", false},
+		{"I cannot reproduce the bug, so the fix holds.", false},
+		{"Done — the output matches the required format and values.", false},
+	}
+	for _, c := range cases {
+		got := selfReportedIncompletion(c.text) != ""
+		if got != c.admit {
+			t.Errorf("selfReportedIncompletion(%q) admitted=%v, want %v", c.text, got, c.admit)
+		}
+	}
+}
+
+// BUG #2(a): a "cannot analyze / cannot determine" admission (present tense) must
+// be caught even when the model also stamps a bogus "PASS" based on a shape check.
+// Mirrors the chess-best-move final ("I cannot analyze the position… PASS: I wrote
+// the move and verified by reading it back").
+func TestAcceptanceCannotPhrasingIsIncomplete(t *testing.T) {
+	provider := &mockProvider{turns: [][]zeroruntime.StreamEvent{
+		textTurn("I cannot analyze the specific position from the image, so I cannot determine the best move. PASS: I wrote e2e4 to the file and verified it by reading it back."),
+	}}
+
+	result, err := Run(context.Background(), "write the best move", provider, Options{
+		Registry:                tools.NewRegistry(),
+		MaxTurns:                10,
+		RequireCompletionSignal: true,
+		SelfCorrect:             enabledSelfCorrector(t),
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !result.Incomplete {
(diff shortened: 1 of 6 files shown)


gnanam1990:
## Code review — completion-honesty gate (high effort, workflow-backed: per-angle finders + independent verifier)

Verified findings below. The design/intent is sound, but the gate currently **over- and under-triggers** in edge cases. **Recommend addressing the High items before merge.**

### 🔴 High — correctness

**1. `internal/agent/guardrails.go` — `selfReportPhrases` matched with no success-negation guard.**
"fall back to", "as a fallback", "fell back to", "placeholder value", "as a placeholder", "best guess" go through bare `strings.Contains` (unlike the inability *stems*, which are guarded). A legit final answer — *"the parser will fall back to UTF-8 when no encoding is set"*, *"I replaced the placeholder value with the computed key"* — is downgraded to INCOMPLETE / exit 4. **Breaks correct runs.** Fix: apply a first-person/context guard to these phrases (or drop the ambiguous ones).

**2. `internal/cli/exec.go` (+`exec_writer.go`) — `--output json` reports `exit_code:0` on INCOMPLETE.**
For non-stream JSON, `writer.final()` emits `{"type":"done","exit_code":0}` *before* the `result.Incomplete` check, and `writer.runEnd` is a no-op for json. A consumer reading the documented `done.exit_code` sees **success** on an abandoned run (process still exits 4). Only stream-json is wired. **Defeats the feature in json mode.** Fix: in the Incomplete branch, emit the json `done` with `exitIncomplete` (mirror the interrupted path).

**3. `internal/agent/loop.go` — `pendingPlanItems` false-INCOMPLETE on stale plan.**
Model completes via tools, then gives a text-only summary without re-calling `update_plan` to flip the last item (very common). `planItemsPending` stays >0 → 3 nudges → INCOMPLETE on a *done* task. Fix: treat pending-plan as a weak signal (don't finalize INCOMPLETE on it alone without a cue/admission), or cap plan-nudges at 1.

**4. `internal/agent/loop.go` — maxTurns exit never sets `Incomplete`.**
A run that keeps issuing tool calls to the turn limit with plan items pending returns `success` → **false-success via the maxTurns door** (the gate only covers the no-tool-call path). Fix: set `Incomplete` on the maxTurns exit when `RequireCompletionSignal` && pending/unfinished.

### 🟡 Medium

**5. `guardrails.go` — `selfReportedIncompletion` uses first-occurrence `strings.Index`.**
*"I could not find any examples… so I could not implement it"* → first "could not" hits the "find any" guard → `continue` → the real later admission is missed (false-success). Fix: scan all occurrences of each stem, applying the guard per-occurrence.

**6. `guardrails.go` — `endsWithContinuationCue` too broad.**
Trailing `:` or last-line "Next, I…"/"Let me…" flags genuine closers (*"Next, I suggest reviewing the changes."*) → false INCOMPLETE. Fix: tighten (don't treat a bare trailing colon / forward-looking recommendation as a mid-step cue).

**7. `exec.go` — cron can't surface the INCOMPLETE reason.**
INCOMPLETE signals via `warning` (not an `error` event); the cron failure-detail extractor only scans `type=="error"`, so scheduled INCOMPLETE runs record a failure with an **empty message**. Fix: also emit an error event (or have cron read `IncompleteReason`).

**8. `loop.go` — gate ordering.**
`pendingPlanItems` is checked before `selfReportedIncompletion`, so an admitted-impossible task burns 3 wasted nudges and reports the wrong reason. Fix: check the admission (self-report) first.

### ⚪ Low (cleanup)

**9. `loop.go` — acceptance pass fires on every `--self-correct` run** (even read-only) → one extra model call each. Gate it on "did the run mutate anything."

**10. `guardrails.go` — `planStatusRemaining` duplicates `tools.normalizePlanStatus`'s status vocabulary** → drift risk. Export + reuse the canonical normalizer.

---
_Tests + `go test ./... -race` are green, but the gate-off (interactive/TUI) path is the only one fully safe today; the headless paths above need the High fixes. Diagnosis from behavior/logs only; no benchmark artifacts in this diff._

gnanam1990:
## Review fixes applied — `2f64739`

All correctness findings addressed; full suite green (`make build` / `go vet ./...` / `make lint` / `go test ./... -race` — 70 pkg, 0 fail, 0 races).

| # | finding | resolution |
|---|---|---|
| 1 | self-report false-positives (no guard) | **Fixed** — dropped behavior-describing phrases ("fall back to", "placeholder value", bare "best guess", "as a fallback", "without proper"); kept first-person/uncertainty admissions only |
| 2 | `-o json` reports `exit_code:0` on INCOMPLETE | **Fixed** — incomplete handled per-format before `final()`; json now emits terminal `done` with exit 4 |
| 3 | pending-plan false-INCOMPLETE on stale plan | **Fixed** — pending-plan is now a *nudge-only* weak signal; only a continuation cue or admission finalizes INCOMPLETE |
| 4 | max-turns exit never set Incomplete | **Fixed** — a max-turns cutoff now finalizes INCOMPLETE under the gate |
| 5 | first-occurrence `strings.Index` masks later admission | **Fixed** — scan every occurrence of each inability stem |
| 6 | `endsWithContinuationCue` too broad | **Fixed** — require trailing colon AND an action lead-in on the final clause (still catches mid-line "…Let me check the config:"; ignores recommendations/summary-colons/sign-offs) |
| 7 | cron can't surface INCOMPLETE reason | **Fixed** — emit an `error` event (not just a warning) so the cron extractor recovers the reason |
| 8 | gate ordering (pending-plan before admission) | **Fixed** — admission is checked first; no wasted nudges, accurate reason |
| 9 | acceptance fires on every `--self-correct` run | **Deferred** — cost optimization (gate on "did the run mutate"); not a correctness bug |
| 10 | `planStatusRemaining` duplicates `tools.normalizePlanStatus` | **Deferred** — the dedup requires exporting from `internal/tools/update_plan.go`, which would widen this PR beyond the agent-fix scope |

**Tests:** added `TestContinuationCueMatching`, `TestMaxTurnsCutoffIsIncompleteUnderGate`; extended `TestSelfReportedIncompletionMatching` (#1/#5 cases); replaced the old `in_progress ⇒ incomplete` test with `TestPendingPlanAloneDoesNotForceIncomplete` (#3 behavior).

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approvable then approve it

PR #326, opened 2026-06-27
feat(tui): ask_user suggested answers with a recommended default

gnanam1990:
## Summary
Adds optional suggested answers to the `ask_user` tool: a question can carry 2-4 `options` with one marked `recommended`, and the interactive TUI renders a picker (arrow keys, lime cursor, recommended default, `Type your own answer` fallback). Optional and backward-compatible — open-ended questions stay free-text; headless `zero exec` is unaffected.

_Follow-up: the tabbed, composer-rendered redesign (multi-question tabs + Confirm, option descriptions, black canvas) is in #327._

PR #327, opened 2026-06-27
feat(tui): tabbed ask_user questionnaire in the composer with suggested answers

gnanam1990:
> Builds on #326 (suggested-answers contract, now merged) — this is the tabbed composer redesign requested on top of it.

## Summary
When the agent asks clarifying questions mid-task, the **composer text box becomes an interactive questionnaire** so the user can pick suggested answers instead of always typing.

- **Tabbed multi-question prompt**: each question is a tab (`Tech Stack | MVP Scope | … | Confirm`); **Tab / Shift+Tab** switch questions, answers given in any order, answered tabs marked ✓, and Enter on **Confirm** submits them all. A single question submits on answer (no Confirm step, no tabs).
- **Suggested answers**: numbered options with the recommended one tagged `(recommended)` and pre-selected; **↑↓** move, **Enter** picks. Each option can carry a one-line **description** sub-line. A trailing `Type your own answer` drops into free-text; typing in the picker also switches to free-text; **Esc** from "type your own" returns to the picker.
- **Rendered in the composer region** on the terminal's black canvas with the composer's border (not a card in the transcript, not a gray panel).
- **Multi-select** questions render as free-text with the options shown as suggestions (a single-pick list can't represent them) — nothing is silently narrowed.

## Contract (optional, backward-compatible)
`ask_user` questions gain optional `header` (short tab title), `optionDescriptions` (aligned to `options`, or option objects `{label, description}`), and `recommended` (kept only when it matches an option). The system prompt tells the model it MAY use them; omitting them falls back to plain free-text exactly as before.

## Constraints honored
- **No regression**: open-ended (no-options) questions are plain free-text.
- **Headless safe**: `zero exec` never wires `OnAskUser`, so the selector can't render there — the non-interactive best-assumption fallback is untouched.
- **Keys**: `Tab move · ↑↓ select · Enter confirm · Esc dismiss`.

## Tests & gate
16 ask-user tests (picker default/recommended, type-your-own → free-text, multi-question tabs + Confirm + ordering, Tab navigation, Esc dismiss partial, descriptions render, typing-in-picker switch, no in-card echo) + tool parsing edges. `make build` / `go vet ./...` / `go test ./... -race` / `make lint` — all green (0 fail, 0 races, 70 pkg).

## Summary by CodeRabbit

* **New Features**
  * Added multi-question ask prompts with question tabs, per-question ✓ status, and a trailing Confirm step.
  * Enhanced prompts to support per-question headers and per-option descriptions, including richer option parsing (labels + optional descriptions).

* **Bug Fixes**
  * Corrected keyboard navigation and submission across single- and multi-question flows.
  * Improved “type your own” and escape behavior to preserve/restore typed input appropriately.
  * Updated display so the questionnaire renders in the composer/footer area instead of the transcript.

* **Tests**
  * Refreshed and expanded ask-user coverage for the new flows and rendering.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1503,10 +1503,12 @@ func toAgentAskUserQuestions(questions []tools.AskUserQuestion) []AskUserQuestio
 	converted := make([]AskUserQuestion, len(questions))
 	for index, question := range questions {
 		converted[index] = AskUserQuestion{
-			Question:    question.Question,
-			Options:     append([]string{}, question.Options...),
-			Recommended: question.Recommended,
-			MultiSelect: question.MultiSelect,
+			Question:           question.Question,
+			Header:             question.Header,
+			Options:            append([]string{}, question.Options...),
+			OptionDescriptions: append([]string{}, question.OptionDescriptions...),
+			Recommended:        question.Recommended,
+			MultiSelect:        question.MultiSelect,
 		}
 	}
 	return converted
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -18,8 +18,10 @@ work.
   blocked on a decision that is theirs to make and that you cannot resolve from
   the code, the request, or sensible defaults. When the answer is likely one of a
   small set, include 2-4 suggested `options` and mark the best as `recommended`
-  (it must be one of the options) so the user can pick quickly; omit options for
-  genuinely open-ended questions.
+  (it must be one of the options) so the user can pick quickly; give each option a
+  short `optionDescriptions` line when a one-word label needs context, and a short
+  `header` (2-3 words) per question as its tab label when you ask several. Omit
+  options for genuinely open-ended questions.
 
 ## Workflow: plan then act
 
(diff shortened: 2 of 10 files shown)


gnanam1990:
Both CodeRabbit comments addressed in `47730b4`:

1. **options schema/parser alignment** — `options.items` is declared `string`, so the schema/description now advertises only the string form; per-option descriptions go through the parallel `optionDescriptions` array. (The parser stays object-tolerant defensively, but strict-schema providers are no longer told to send objects.)
2. **single-question Tab** — `moveAskUserTab` now no-ops when there's ≤1 question, so Tab/Shift+Tab can't move a single-question prompt into the hidden Confirm state. Test added (`TestAskUserSingleQuestionTabIsNoOp`).

Gate green: `make build` / `go vet ./...` / `go test ./... -race` / `make lint` (0 fail, 0 races, 70 pkg).

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit if approvable then approve it

PR #329, opened 2026-06-27
feat: encrypted provider credentials + provider-aware /model

gnanam1990:
## Summary

Stops provider API keys from being stored in plaintext, and makes `/model` reflect what each connected provider actually serves.

Two themes:

### 1. Encrypted credential storage
Previously a pasted API key was written in cleartext to `config.json` (`apiKey`). Now:
- **`internal/securefile`** — AES-256-GCM encryption-at-rest (fail-closed, atomic write, cross-process lock; key in a `0600` sidecar).
- **`internal/credstore`** — provider API-key store. Backend resolves **keyring-first** (OS credential store) → **encrypted-file** fallback; plaintext only via an explicit `ZERO_CRED_STORAGE=file` opt-out.
- **Keys leave `config.json`:** the wizard / CLI setup write the key straight into the store and persist only an `apiKeyStored` marker. Any existing inline plaintext key is **migrated to the store on launch** (fail-soft; the cleartext is stripped only after the store write succeeds, so a key is never stranded).
- **Load on use:** `buildProvider` and *every* model/provider switch reload the credential (stored key → env var → OAuth bearer), so a stored-key provider always authenticates — including after switching models mid-session.
- **`zero auth logout <provider>`** now removes the stored API key + marker (not just the OAuth token).

### 2. Provider-aware `/model` and `/provider`
- `/model` lists **every authenticated provider** (filtered to those with a usable credential — key / stored / env / OAuth / local), **grouped per provider** under section headers.
- Each provider's models are **live-discovered** (its real served models) instead of a static catalog; **local providers show only the models actually pulled** (no fake entries).
- Selecting a model from a different provider **switches provider + model together**.
- `/provider` offers **keep / replace / remove** when the selected provider already has a stored key.

## Behavior / compatibility
- Backward compatible: existing inline keys keep working and are migrated transparently on first interactive launch.
- Headless `exec` is unchanged (no migration, no interactive prompts).
- Keychain access prompts can be avoided with `ZERO_CRED_STORAGE=encrypted-file`.

## Test plan
- `go build ./...`, `go vet ./...`, `go test ./... -race`, lint — all green (72 packages).
- New unit tests: securefile round-trip / tamper / wrong-key; credstore backends + auto-resolution + migration + marker; `/model` provider grouping + live-discovery refresh + fallback; `/provider` keep/replace/remove.
- Manually verified: a stored key authenticates after a cross-provider model switch; `/model` shows real served models per provider; un-connected providers are hidden.

## Summary by CodeRabbit

* **New Features**
  * Added encrypted credential storage for provider API keys, including fail-soft migration from existing plaintext keys and automatic use of stored keys during interactive runs.
  * Enhanced the model picker to list models grouped by saved providers, refresh live models per provider, and support switching provider+model together.
  * Added a provider “manage key” step (keep, replace, remove) for providers with stored keys.
* **Bug Fixes**
  * Logout now clears both OAuth credentials and stored provider API keys, with updated “no stored credential” messaging.
  * Credential indicators now correctly reflect stored-key availability even when the inline key field is empty.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -514,6 +514,13 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	if err != nil {
 		return writeAppError(stderr, "failed to resolve user config path: "+err.Error(), 1)
 	}
+	// Fail-soft, one-time: move any inline plaintext API keys in config.json into
+	// the encrypted credential store (interactive runs only; headless exec keeps
+	// its existing behavior). Non-fatal — a missing keyring or write error leaves
+	// the inline key in place and this run still uses the already-resolved key.
+	if store, storeErr := config.ProviderKeyStoreAt(filepath.Dir(userConfigPath)); storeErr == nil {
+		_, _ = config.MigratePlaintextProviderKeys(userConfigPath, store)
+	}
 	doctorUserConfigPath := ""
 	projectConfigPath := ""
 	if resolveOptions, optErr := config.DefaultResolveOptions(workspaceRoot); optErr == nil {
@@ -640,6 +647,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		ProviderName:         resolved.Provider.Name,
 		ModelName:            resolved.Provider.Model,
 		ProviderProfile:      resolved.Provider,
+		SavedProviders:       usableSavedProviders(resolved.Providers),
 		FavoriteModels:       resolved.Preferences.FavoriteModels,
 		RecapsEnabled:        resolved.Preferences.RecapsEnabled(),
 		Provider:             provider,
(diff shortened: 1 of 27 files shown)


gnanam1990:
Thanks @coderabbitai — addressed the substantive findings (pushed):

**Fixed**
- **`ApplyStoredAPIKey` now gates on `APIKeyStored`** — a stale keyring/file entry can no longer reactivate a key for a profile that didn't opt into stored-key auth.
- **`usableSavedProviders` verifies the key is actually retrievable** for marker-only profiles, so `/model` won't list a provider whose stored key was deleted; it falls back to env/OAuth/local signals (marker off) otherwise.
- **Cross-provider switch gates on the *resolved* credential, not the marker** — a missing/unreadable stored key now gives a clear "no usable credential" message instead of building a keyless provider that 401s later.
- **`auth logout` surfaces deletion failures** (`ForgetProviderKey` / `ClearProviderKeyStored`) instead of reporting success while a credential remains.
- **Clarified `ProviderKeyStoreAt` docs**: provider keys are **user-scoped by design** — only ever captured under the user config (a cloned repo must not carry keys), so runtime lookups intentionally use the user store regardless of where a profile was resolved from. The `dir` param exists for test isolation.

**Deferred (with reasons)**
- **Capture/persist atomicity (setup, `providers add`)** — if `UpsertProvider` fails after the store write, the stored key is *orphaned but harmless*: no profile references it, it's idempotently overwritten on retry, and removable via `auth logout`. A full two-phase rollback is disproportionate to the risk here; happy to add it if you'd prefer.
- **File-backend write serialization** — the default backend is the OS keyring (per-entry, no shared blob). The encrypted-file fallback's atomic temp-rename prevents corruption; only a concurrent cross-process write to the *same* file could drop one update, which is rare for interactive credential edits. Can add a lockfile (mirroring the OAuth store) if desired.

All gates green: `go build`, `go vet ./...`, `go test ./... -race` (72 pkg), lint.

PR #331, opened 2026-06-28
feat(tui): compact one-line /model switch confirmation

gnanam1990:
## Summary

The `/model` switch confirmation printed a 6–7 line block, which felt heavy for a routine action. This collapses it to a single summary line.

**Before**
```
Model
Switched model.
model: deepseek-v4-pro
provider: ollama-cloud
api model: deepseek-v4-pro
effort: auto
saved: user config
```

**After**
```
Model
deepseek-v4-pro · ollama-cloud · saved
```

## Details
- Format: `<model> · <provider>[ · effort <x>][ · saved]`.
- Effort is shown only when it isn't `auto` (or `effort auto (reset)` when an unsupported carry-over preference was dropped) — no noise otherwise.
- A deprecation/redirect notice (e.g. "gpt-4-turbo is deprecated; using gpt-4.1 instead") still renders as its own line when relevant.
- The cross-provider switch confirmation was already compact (`Switched to X · Y`) and is unchanged.

## Test plan
- Updated the five tests that asserted the old multi-line format.

## Summary by CodeRabbit

* **UI Improvements**
  * Simplified the model-switch confirmation into a cleaner one-line summary under the existing “Model” heading.
  * The message now concisely shows the selected model, provider, effort status (including “auto (reset)” when applicable), and whether the selection was saved.
  * Reduced extra detail from the previous multi-line confirmation while preserving any related notice.
* **Tests**
  * Updated transcript-based tests to match the new one-line model-switch wording and formatting.

### DIFF
--- a/internal/tui/command_center.go
+++ b/internal/tui/command_center.go
@@ -459,22 +459,27 @@ func (m model) handleModelCommand(args string) (model, string) {
 			effortLine = "effort: " + string(effective)
 		}
 	}
+	// Compact one-line summary — "<model> · <provider>[ · effort …][ · saved]" —
+	// instead of a multi-line model/provider/api/effort/saved block.
+	summary := target.modelID + " · " + displayValue(nextProfile.Name, string(metadata.ProviderKind))
+	effort := strings.TrimSpace(strings.TrimPrefix(effortLine, "effort: "))
+	if resetEffort {
+		summary += " · effort auto (reset)"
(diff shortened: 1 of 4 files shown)


gnanam1990:
Addressed — restored the persistence error text: save failures now show `· not saved (<reason>)` instead of a bare `· not saved`, so the actionable clue is preserved. Pushed.

PR #332, opened 2026-06-28
feat: context-aware compaction + usage gauge for all models

gnanam1990:
## Summary

Makes ZERO's context-window awareness work for **every** model, not just the ones in the curated registry. Two related fixes:

### 1. Auto-compaction for all models
Auto-compaction is gated on the model's context window (`enabled = ContextWindow > 0`), which was resolved **only from the registry**. Unregistered models — GPT-5/Codex, xAI, Ollama-cloud, custom endpoints — resolved to `0`, so **both** the proactive (~80%) and the reactive post-overflow compaction were silently disabled and long sessions just overflowed.

- Context-window resolution now also consults **live-discovered** windows (accurate for proxy/custom models once discovery has run).
- New `modelregistry.AgentContextWindow` applies a positive `FallbackContextWindow` (200k) **only on the compaction-enable path** (TUI run + CLI `exec`/spec), so compaction is on for every model while **display gauges stay accurate-or-empty** (no guessed denominator). Exact windows (registry/discovered) always win.

### 2. Context-usage gauge for all models, from launch
The sidebar/status-line gauge (`used / total tokens` + a % fill) only rendered a total when the window was known. It now **warms model discovery for the active provider in `Init()`** (background, non-blocking), so the active model's window is learned at startup and the gauge shows `used / total` + fill % for every model — proxy/custom included. If a provider doesn't report a window, that model still shows just the used-token count (never a fake total).

## Test plan
- New unit tests: context-window resolution (registry / live-discovered / unknown→0) and `AgentContextWindow` fallback.
- Manual: on ollama-cloud / Grok / GPT-5, long sessions now compact instead of overflowing, and the sidebar shows `used / total` + % shortly after launch.

## Summary by CodeRabbit

* **New Features**
  * Improved model context-window sizing, including live-discovered provider models.
  * Added a positive fallback so compaction can stay enabled for unknown/custom models.
  * Start a background warm-up to discover active provider models.
* **Bug Fixes**
  * Corrected and unified context-window calculations across the UI (sidebar token, title bar, status line) and for compaction requests and exec runs.
* **Tests**
  * Added coverage for context-window resolution (registered, unknown, empty, live-discovered).

### DIFF
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -16,6 +16,8 @@ import (
 	"github.com/Gitlawb/zero/internal/lsp"
 	"github.com/Gitlawb/zero/internal/modelregistry"
 	"github.com/Gitlawb/zero/internal/notify"
+	"github.com/Gitlawb/zero/internal/providercatalog"
+	"github.com/Gitlawb/zero/internal/providermodeldiscovery"
 	"github.com/Gitlawb/zero/internal/providers"
 	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/sessions"
@@ -500,7 +502,7 @@ func runExec(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 	defer lspShutdown()
 	result, err := agent.Run(runCtx, agentPrompt, provider, agent.Options{
 		MaxTurns:         resolved.MaxTurns,
-		ContextWindow:    modelregistry.AgentContextWindow(modelContextWindow(modelRegistry, resolved.Provider.Model)),
+		ContextWindow:    resolveAgentContextWindow(runCtx, modelRegistry, resolved.Provider),
 		DeferThreshold:   effectiveDeferThreshold,
 		Specialists:      specialistRuntime.specialistInfos(),
 		SessionID:        preparedSession.Session.SessionID,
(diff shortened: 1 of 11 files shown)


gnanam1990:
Addressed the actionable findings (pushed):

**Fixed**
- **Model-switch compaction now covers all targets** — wrapped the `/model` and `/mode` pre-switch `TargetContextWindow` in `modelregistry.AgentContextWindow`, so undiscovered/custom switch targets get the fallback and aren't skipped by `BeforeModelSwitch`'s `<= 0` guard.
- **Active-provider scoping** — `modelContextWindow` now resolves the live-discovered window from the **active provider first**, then falls back to other providers, so a model ID shared across providers resolves to the one actually in use (extracted a `discoveredContextWindow` helper).
- **Test tightened** — compares the resolved window to the registry's actual `gpt-4.1` value instead of `!= FallbackContextWindow`.

**Skipped (with reason)**
- **CLI `exec` threading provider-discovered windows** — headless `exec` is intentionally registry + fallback only; adding a live `/models` discovery call to every one-shot run would add latency and a network dependency to a deterministic CLI path. The "compact too late → fail" case it worries about is already handled: the positive fallback enables the **reactive** post-overflow compaction (`recover()`), which compacts and retries when the provider returns a context-limit error. So uncatalogued small-window models recover rather than hard-fail. Happy to add opt-in discovery here if you'd prefer.

Gates green: `go build`, `go vet ./...`, `go test ./... -race` (72 pkg), lint.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit if approvable the approve it

gnanam1990:
Reconsidered and implemented the headless case too (pushed `b868f26`):

`resolveAgentContextWindow` now gives **provider-discovered windows precedence** in CLI `exec` and spec runs. On a registry miss it queries the provider's live model list (resolving the credential from inline → stored → env) for the model's real context window, then falls back to the 200k default only if that's unavailable. Discovery runs **only on a registry miss** (catalogued models pay no latency), is bounded by a **5s timeout**, and degrades to the fallback on any error — so a headless run is never blocked or failed by it.

That resolves all four findings (the other three were in `5ef936e`). Gate green: build, vet, `go test ./... -race` (72 pkg), lint.

gnanam1990:
@coderabbit review again if approvable then approve it

gnanam1990:
@coderabbit approve

PR #333, opened 2026-06-28
feat(tui): dim transcript backdrop behind overlays for visibility

gnanam1990:
## Summary

Slash-command/`@file` palettes (and the model/provider pickers and wizards) were composited directly over the live transcript with no visual separation, so the chat bled through around the box and the overlay was hard to read.

This adds a **modal scrim**: when an overlay is open, the transcript backdrop is dimmed (each line's colors are stripped and re-rendered faint) before the bright bordered overlay is drawn on top. The header/title bar and composer render separately and stay bright, so only the chat behind the palette recedes. Backdrop text is preserved (just dimmed), so context stays faintly readable.

Net effect: the palette clearly stands out as a floating card instead of blending into the chat.

## Scope
- `internal/tui/model.go` (+17): dim the backdrop in `overlayViewportLines` via a `scrimViewportLine` helper.
- `internal/tui/model_test.go` (+31): tests that the scrim preserves text + skips blank lines, and that the overlay still composites over a preserved backdrop.

## Test plan
- Manual: open `/` (command palette) or `@` (file palette) over a busy transcript — the chat behind dims and the palette is clearly legible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Overlay views now dim the background transcript when menus, pickers, or wizards are open, making the active overlay stand out more clearly.
  * Existing text behind overlays is preserved and shown faintly instead of being visually disrupted.
  * Blank lines remain unchanged, keeping spacing intact in the transcript display.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -2261,6 +2261,12 @@ func overlayViewportLines(lines []string, overlay string, width int) []string {
 	if overlayWidth <= 0 {
 		return lines
 	}
+	// Scrim: dim the whole transcript backdrop so a floating overlay (slash-command
+	// palette, picker, wizard) clearly stands out instead of blending into the live
+	// chat behind it. Header and composer are rendered separately and stay bright.
+	for index := range lines {
+		lines[index] = scrimViewportLine(lines[index], width)
+	}
 	start := maxInt(0, (len(lines)-len(overlayLines))/2)
 	for offset, line := range overlayLines {
 		target := start + offset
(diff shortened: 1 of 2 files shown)


gnanam1990:
Addressed both — strengthened the scrim tests: `TestScrimViewportLine` now feeds a pre-styled line and asserts its original escape styling is stripped while the text survives; `TestOverlayViewportLines…` now asserts the **overlaid** row keeps backdrop margin text alongside the panel (the compositing contract), not just a non-overlaid row. Gate green.

gnanam1990:
@coderabbit review again if approvable then approve it

PR #334, opened 2026-06-28
Polish TUI transcript rendering

anandh8x:
## Summary
- tighten transcript spacing and remove distracting rails/gutters
- render tool activity with cleaner action labels and grouped exploration summaries
- improve markdown, code, diff, and command-output rendering in the transcript

## Tests
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Smarter markdown rendering: bare-code auto-highlighting, fenced-code highlighting, and horizontal-rule dividers now render correctly (including during streaming).
  * Live tool-card/prose layout improvements: clearer rule/separator behavior, friendlier labels, and more stable interim markdown.
  * Syntax highlighting can infer language and optionally use richer styling; streaming code buffers until fences close.
* **Bug Fixes**
  * Preserve pre-styled/ANSI output during streaming and fading.
  * Improve streamed tool views and late path detection; adjust diff/command/search display to reduce visual noise.
  * Update transcript wrapping, gutters, and selection alignment; refine rendering cache uniqueness.
* **Tests**
  * Expanded TUI, streaming, and rendering regression coverage for the updated UI contracts.

### DIFF
--- a/internal/tui/assistant_markdown.go
+++ b/internal/tui/assistant_markdown.go
@@ -1,6 +1,8 @@
 package tui
 
 import (
+	"crypto/sha256"
+	"fmt"
 	"strings"
 	"unicode"
 	"unicode/utf8"
@@ -137,18 +139,22 @@ func renderAssistantMarkdownText(text string, proseMeasure int, tableMeasure int
 
 		if allowHighlight && looksLikeBareCodeLine(line) {
 			code := []string{}
-			for index < len(raw) {
-				next := raw[index]
+			probe := index
+			for probe < len(raw) {
+				next := raw[probe]
 				nextTrimmed := strings.TrimSpace(next)
-				if nextTrimmed == "" || markdownStartsBlock(raw, index) || !looksLikeBareCodeLine(next) {
+				if nextTrimmed == "" || markdownStartsBlock(raw, probe) || !looksLikeBareCodeLine(next) {
 					break
 				}
 				code = append(code, strings.ReplaceAll(next, "\t", "    "))
-				index++
+				probe++
 			}
-			if highlighted, ok := highlightCodeAuto(code, "", tableMeasure); ok {
-				lines = append(lines, highlighted...)
-				continue
+			if looksLikeBareCodeBlock(code) {
+				if highlighted, ok := highlightCodeAuto(code, "", tableMeasure); ok {
+					lines = append(lines, highlighted...)
+					index = probe
+					continue
+				}
 			}
 		}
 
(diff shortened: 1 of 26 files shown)


gnanam1990 (changes requested):
Reviewed against the branch (build + `go test ./internal/tui -race` + `vet` all pass locally). Solid, internally-consistent rendering refactor with good coverage for the new behaviors — width/layout math, ANSI preservation, selection coordinates, and the streaming-decoder path all check out. A few things to address before merge, mostly around a perf regression, one fragile heuristic, and two information-removal changes that were locked in by weakened tests.

## Major

**1. Streaming re-highlights closed code blocks every frame (perf).**
`interimBlock` (`internal/tui/model.go:2459`) now calls `renderStreamingAssistantMarkdownText` → `renderAssistantMarkdownText(stablePrefix, …, allowHighlight=true)`. The previous code passed `false` deliberately so the per-frame stream loop never re-lexes. `streamingMarkdownStablePrefix` only excludes the *open* fence, so once a fence **closes** mid-stream, chroma re-tokenizes that whole block on every render/spinner/fade frame for the rest of the turn — O(block) CPU per frame, the exact invariant the old comments protected. Suggest caching the highlighted stable region (highlight the newly-stable portion once) rather than re-highlighting each frame.

**2. (Confirm intent) grep/read/list output is no longer viewable anywhere.**
`exploreCardBody` (`internal/tui/rendering.go:2129`) returns a single summary line and ignores `bodyCap`/`expanded`, and the expand affordance is suppressed for these tools. grep match locations/counts and `read_file` contents/line-ranges are gone even on expand. Tests were changed to assert the dump is hidden — this is real user-facing information loss and should be a conscious product decision, not a density tweak.

**3. (Confirm intent) Manual permission approvals leave no transcript trace.**
`rowContext.skip` now returns `true` unconditionally for `PermissionActionAllow` (`internal/tui/rendering.go:157`, was `return rc.auto[key]`). Manual "allowed once / always · <tool>" audit lines disappear (deny/cancel still render). The test was rewritten to assert `skip(...)` rather than the audit text, so this slipped in quietly — worth confirming approvals don't need an audit line.

## Minor

**4. Prose mis-highlighted as code.** `looksLikeBareCodeLine` (`assistant_markdown.go:371`) runs on individual prose lines (`assistant_markdown.go:138`). Lines like `for these reasons:` (HasPrefix `for ` + HasSuffix `:`) and `return …` match, so ordinary sentences get collected and Python-highlighted. The new test covers only one non-matching line. Suggest requiring the stronger block-level signal (`looksLikeBareCodeBlock`) for the inline path, or ≥2 distinct code signals.

**5. Rune-count truncation against a display-width budget.** `renderCommandOutputLines`, `exploreCardLine`, and `localControlChildLine` truncate with `truncateRunes` against a `lipgloss.Width`-based budget; the code they replaced used width-aware `ansi.Truncate`. Wide/CJK glyphs can overrun. Also leaves now-dead `grepCardBody`/`grepMatchPattern`.

## Nits
- `diffCardMetadata` counts adds via `HasPrefix(line, "+++ ")` (space) while `highlightedAddedDiffLines` excludes headers via `HasPrefix(line, "+++")` (no space); a `+++x` content line diverges the counts and silently disables highlighting for that diff. Harmless, but the predicates should match.
- Fade no longer applies to highlighted lines (`styleStreamingLine` early-returns on `\x1b`) — intended/tested, just noting the fade is now prose-only.

## Recommendation
Request changes (light): please address (1) the streaming re-lex and (4) the prose-as-code heuristic, and get explicit sign-off on (2) and (3) since they remove user-visible information. Everything else is approve-with-nits.

Heads-up: this touches `model.go`, `rendering.go`, `theme.go`, `transcript_selection.go`, `syntax_highlight.go` — the same files as a few other in-flight TUI PRs, so a rebase will be needed depending on merge order.

gnanam1990 (approved):
Re-reviewed at `ecce229` — all points from the previous review are addressed. Build + `go test ./internal/tui` pass locally; CI green.

- **Streaming re-lex (Major):** ✅ `renderStreamingAssistantMarkdownText` now memoizes the stable prefix via `defaultRenderCache` (sha256 key), so closed code blocks aren't re-tokenized every frame.
- **grep/read output viewable (Major, intent):** ✅ `cardRenderOptions.expanded` threaded through; `exploreCardBody` shows the full output on expand with a collapse footer otherwise — no longer permanently hidden.
- **Permission audit trace (Major, intent):** ✅ `PermissionActionAllow` skip restored to `rc.auto[key]`, so manual approvals render an audit row again while silent auto-approvals collapse.
- **Prose-as-code (Minor):** ✅ inline bare-code path now gated on `looksLikeBareCodeBlock` (strong-signal count) with a non-committing `probe`, so ordinary prose like `for these reasons:` no longer gets highlighted.
- **Diff `+++` nit:** ✅ matched to `"+++ "`.

Nice work. Approving.

PR #337, opened 2026-06-28
feat(tui): remove the /mode command (covered by /model + /effort)

gnanam1990:
## Summary

Removes the TUI `/mode` slash command. Its presets bundled a **hardcoded** model (`smart`→`claude-sonnet-4.5`, `deep`→`claude-opus-4.1`, `large`→`gemini-2.5-pro`, …) + reasoning effort + turn budget, and `handleModeCommand` actually switched you to that model. On any non-Anthropic/Google setup (Ollama-cloud, xAI, GPT-5, custom endpoints) `/mode` would switch to a model you're not authenticated to — so it's effectively broken there. With **`/model`** (any provider's models) and **`/effort`** (reasoning effort) covering the same ground flexibly, `/mode` is redundant in the TUI.

## What's removed
- The `/mode` command (registry entry + `commandMode` kind), its dispatch, `handleModeCommand`, `modeListText`, the `pickerMode` kind + `newModePicker`, and the picker-apply case.
- Stale `/mode` references in comments.
- The four `TestModeCommand*` tests; updated autocomplete/picker tests that relied on `/mo` being ambiguous (`/model` + `/mode`).

## What's kept
- `internal/modelregistry` mode presets (`Modes`, `ModeNames`, `LookupMode`) and the **CLI `--mode` flag** (`zero exec --mode <name>`) — still used by headless exec, out of scope here.

## Test plan

## Summary by CodeRabbit

* **Chores**
  * Removed the `/mode` command from the TUI command list, help output, and session controls.
  * Updated the settings picker to only support `/model` and `/effort` (mode picker removed).
* **Bug Fixes**
  * Updated command suggestions and suggestion overlay behavior so keyboard navigation/tab completion reflect the remaining supported commands.
* **Tests**
  * Adjusted autocomplete, picker, and session-control tests to match the updated command set and overlay expectations.


gnanam1990:
@coderabbit review if approvable then approve it

gnanam1990:
@coderabbit approve

PR #338, opened 2026-06-28
feat(reasoning): typed per-provider reasoning capability from a models.dev catalog

Vasanthdev2004:
## What

Reasoning-effort detection today guesses from model-name substrings (`reasoningEffortsForModelName`), which is wrong for half the catalog because providers describe reasoning control with **incompatible concepts**:

- **OpenAI** — a discrete effort enum, split by minor version (`gpt-5` = `minimal/low/medium/high`, `gpt-5.1` = `none/low/medium/high`, `gpt-5.1-codex-max` adds `xhigh`, `gpt-5-pro` is locked to `high`).
- **Anthropic** — a thinking-token **budget** on legacy models, a native **effort enum** on newer ones (with the budget control removed).
- **Gemini** — budget + toggle on 2.5, an effort enum on 3.x.
- Some models only **toggle** thinking on/off.

A flat effort string can't represent these, and a name heuristic can't detect them.

## Change

New `internal/reasoning` package (stdlib-only, so the registry and provider adapters can import it without a cycle):

- `Capability` — a model's `reasoning` flag plus its typed `Controls` (`effort` / `budget_tokens` / `toggle`), with helpers (`EffortValues`, `SupportsEffort`, `Budget`, `HasControl`).
- `Catalog` — looks capability up by **provider + api model id**, sourced from an **embedded, trimmed snapshot of the community capability database** (models.dev). Routers (OpenRouter/Azure/etc.) are excluded — they carry generic/stale option lists — so lookups resolve to the first-party provider. 22 KB, 158 models across anthropic/openai/google/google-vertex/xai/groq/deepseek.

**Purely additive** — nothing is wired into the live resolution path yet. The follow-up routes the existing resolver through this catalog (as the fallback rung beneath any future provider live-API probe, above the demoted name-heuristic), so the `/effort` picker and the wire layer agree on real data.

## Coverage (verified)

Joining the snapshot against Zero's actual shipped catalog (13 models): the 6 gpt-4.x/4o models resolve to `reasoning:false` (matching Zero's `(none)`), and every shipped Anthropic/Gemini reasoning model resolves with the correct control kind. The snapshot is current — it already carries gpt-5.5, claude-opus-4-8, gemini-3.x.

## Test

`go test ./internal/reasoning` pins the per-version OpenAI splits, the Anthropic budget-vs-effort distinction (incl. the budget removed on newer models), the Gemini 2.5-vs-3 concept switch, and that every reasoning model Zero ships today is covered. Build/vet/gofmt clean; full `go build ./...` green.

Part of the phased reasoning-effort redesign (independent of #335 / #336).

## Summary by CodeRabbit

* **New Features**
  * Added an embedded, unified reasoning-capabilities catalog across providers, including effort tiers, token-budget bounds, and reasoning on/off toggles.
  * Added support for a new “max” reasoning effort tier and improved matching for requested effort values.
* **Bug Fixes**
  * Ensured capability data returned to callers is isolated via deep copies, preventing cross-lookup mutation.
  * Improved lookup behavior for blank or unknown provider/model inputs.
* **Tests**
  * Expanded coverage for embedded data integrity and verified deep-copy behavior for effort and budget accessors.


Vasanthdev2004:
Thanks — all three addressed in 35145a5:

1. **Budget sentinel** — `Control.Min/Max` are now `*int`, so an explicit bound is distinct from an omitted one. Gemini `gemini-2.5-flash`'s `min: 0` (thinking can be disabled) and Anthropic's *absent* `max` (unbounded) are both preserved instead of collapsing to `0`. `Budget()` → `BudgetControl()` returns the control so callers read the pointers. New assertions pin `min: 0` and the nil `max`.

2. **Shallow copy** — `Catalog.Lookup` now returns a deep clone (cloning `Controls`, each `Values` slice, and the budget pointers). `TestLookupReturnsDeepCopy` verifies a caller's mutation can't reach the catalog.

3. **Non-first-party MaaS rows** — dropped the `google-vertex` section entirely (it re-hosts Anthropic/Moonshot/OpenAI/DeepSeek/Qwen/Meta MaaS models). `google` (AI Studio) already covers first-party Gemini, so the `gemini` provider kind resolves there. Snapshot is now 123 models / 16.8 KB; grep confirms no `*-maas` / `google-vertex` entries remain.

gnanam1990 (commented):
Reviewed `1282a23` (the commit's own diff — its parent is on `main`, so the real change is **only** `internal/reasoning/`: `capability.go`, `catalog.go`, `catalog_test.go`, `modelsdev_snapshot.json`). The code is clean and the data model is genuinely better than the current flat effort enum. My concerns are about it being unwired groundwork that sets up a second source of truth.

**Strengths**
- `go:embed` + parsed once at init (no disk-read fragility, no per-call parse), std-lib-only (deliberate, so `modelregistry`/adapters can consume it without an import cycle), `ParseCatalog` rejects empty/malformed input. The `Capability` model (reasoning flag + typed effort / budget_tokens / toggle controls) captures what a flat string can't.

**Major — second taxonomy with no consumer or reconciliation plan**
The package has **zero importers** (`grep -rn internal/reasoning` is empty outside itself) and the commit message confirms "nothing is wired into the live resolution path yet." Meanwhile the live `/effort` / `/model` paths still use `modelregistry.ReasoningEfforts()` (`internal/modelregistry/catalog.go:171`). So this lands a 158-model / ~22 KB data dependency that nothing exercises and can silently drift. Either wire it in as the authoritative source (demoting the name-pattern guesser to a fallback) in this PR, or land it together with its consumer — and state which source wins. As-is it's dormant, but it's dormant groundwork for exactly the two-sources-of-truth problem.

**Major — `max` effort tier has no `modelregistry` constant**
The snapshot emits `"max"` (19×, e.g. `gpt-5.1-codex-max`, `claude-opus-4-8`), but `modelregistry`'s constants are `none/minimal/low/medium/high/xhigh` (`internal/modelregistry/models.go:22-27`) — no `max`. The moment the catalog is wired into the effort path, `max` has nothing to map to. Cheap to fix now: add `ReasoningEffortMax`, or define an explicit clamp/mapping at the boundary.

**Minor — `Lookup` is exact-match only** (`catalog.go`)
Provider is lowercased but the api model id is only trimmed — no case-fold, no stripping of router prefixes (`openai/…`) or suffixes (`:cloud`). Fine for first-party clean wire ids, but a footgun: callers must pass the **api** id (`claude-opus-4-1-20250805`), not the friendly registry id (`claude-opus-4.1`). Worth documenting loudly at the (future) call site or normalizing in `Lookup`.

**Minor — tests pin per-version snapshot values** (`catalog_test.go`)
Hard-pinning exact effort lists / budget bounds against a snapshot whose own note says "trimmed snapshot" means every refresh breaks them. The structural tests (miss-falls-through, rejects-empty, every-shipped-model-covered) are the durable ones; consider generating the value assertions from the snapshot or keeping only a couple canonical pins.

**Minor — snapshot provenance is partial**
Header has `_source: models.dev/api.json` + a note, but no fetch date/commit and no regeneration script, so refreshes aren't reproducible/auditable. Add a date stamp + a regen script.

**Nit** — `cap` shadows the builtin in `catalog.go` and the tests; rename.

**Recommendation: approve-with-nits — don't block correct, additive, dormant code, but please (1) add `ReasoningEffortMax` (or a documented mapping) and a provenance date in this PR, and (2) before/when wiring it in, state which of `internal/reasoning` vs `modelregistry` is authoritative so there's one source of truth.** The most important item is that reconciliation decision.

Vasanthdev2004:
Addressed in a454931.

**CodeRabbit (blocking) — accessor aliasing:** `EffortControl`/`BudgetControl` now return a cloned `Control`, so `Min`/`Max` pointers and the `Values` slice can't be mutated through an accessor result. `Lookup` already deep-copies; this closes the accessor path too. `TestAccessorsReturnDeepCopies` pins it.

**@gnanam1990 — thanks, all actioned:**
- **`max` tier** → added `modelregistry.ReasoningEffortMax` (`"max"`) and accept it in `ValidReasoningEffort`, so the snapshot's `max` (gpt-5.1-codex-max, newer Anthropic) has a constant to map to. No curated model lists it yet.
- **Provenance** → snapshot now carries `_fetched: 2026-06-28` and a documented regeneration recipe in `catalog.go` (which providers, what to keep).
- **`Lookup` exact-match** → doc now states loudly: pass the **api** id (wire name, `claude-opus-4-1-20250805`), not the friendly id (`claude-opus-4.1`); match is exact (no case-fold / prefix-strip).
- **`cap` shadow** → renamed to `entry` in `catalog.go` and the tests.
- **Source of truth (the key item)** → recorded in the package doc: **`internal/reasoning` is the intended authoritative source**; when wired in, `modelregistry`'s name-pattern inference (`reasoningEffortsForModelName`) is **demoted to a fallback** for models the catalog doesn't cover — the catalog wins where it has an entry. This commit is deliberately additive groundwork (no consumer yet, so no divergence). The wiring lands in the follow-up PR, which is where the demotion actually takes effect.
- **Test value pins** (minor) → kept the canonical ground-truth pins (they're the point — they'd catch a bad refresh), alongside the structural tests (`miss-falls-through`, `rejects-empty`, `every-shipped-model-covered`) that survive refreshes.

PR #339, opened 2026-06-28
Clean up public release docs

anandh8x:
## Summary
- rewrite the README around the public release user flow and add the logo asset
- refresh install/update docs and add a docs index
- remove internal planning, audit, design, and work-split documents from public docs
- update stale references and workspace seed expectations to point at public docs

## Tests
- env GOCACHE=/tmp/zero-go-cache go test ./internal/workspaceseed
- git diff --check

## Summary by CodeRabbit

* **Documentation**
  * Reworked the main README into a clearer, linear, usage-focused guide (install → first run → daily use → safety/control → commands).
  * Refreshed install and update-flow docs with clearer commands, flags, and check-only exit/JSON behavior.
  * Added a documentation landing page and updated reference links/content.

* **Bug Fixes**
  * Updated workspace seed output to surface installation/stream docs instead of removed PRD/work-split materials.

* **Chores**
  * Removed multiple outdated planning/audit/design documents and cleaned up related changelog/agent reference mentions.

PR #340, opened 2026-06-28
chore(launch): MIT license, windows-arm64 install guard, TUI TTY guard

Vasanthdev2004:
## What

Three launch-readiness fixes that the docs cleanup (#339) doesn't touch — all code/config, not docs.

### 1. `package.json` license → `MIT`
Was the placeholder `"SEE LICENSE IN LICENSE"`; now matches the `LICENSE` file (MIT) and the README badge.

### 2. Windows-ARM64 install no longer hard-fails npm
`(win32, arm64)` resolves to a *valid* platform/arch in `postinstall.mjs`, so it proceeded to download `zero-…-windows-arm64.zip` — which the release matrix never builds → **HTTP 404 → `fail()` → `npm install` aborts** on Windows on ARM. (Truly-unknown platforms already skip gracefully; this combo slipped past because it looks supported.) Now it's guarded like any unsupported target: `warnSkip` (exit 0) with a pointer to the x64 build (which runs under emulation on Win-ARM) or source. The `os`/`cpu` arrays can't express this single-combo exclusion without dropping the `arm64` builds we *do* ship (linux/macos), so the guard belongs in the wrapper.

### 3. Interactive shell no longer hangs without a TTY
`tui.Run` fed `os.Stdin` straight into Bubble Tea, so `echo "" | zero` (piped/redirected input) **blocked forever**. Now it fails fast with guidance toward `zero exec` and a non-zero exit. Dependency-free `os.ModeCharDevice` check — no new module added (`x/term` isn't in the graph).

## Tests
- `TestPostinstallSkipsWindowsArm64` — win-arm64 exits 0 with the skip message (no plan, no 404).
- `TestRunRejectsNonTTYStdin` — the shell returns non-zero (not a hang) on non-TTY stdin.

`go build ./...`, `go vet`, `gofmt`, `node --check`, and the affected suites all green.

## Scope note
This is the code/config half of the launch list. The remaining items are the **release trigger** (push `v0.1.0` to origin → cut the GitHub Release → `npm publish`) and the README install framing — both owner decisions, intentionally not in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup behavior when launched without an interactive terminal by showing a clear error and returning exit code `2` instead of hanging.
  * Added graceful handling for Windows on ARM64 during setup by skipping missing artifacts and noting the Windows x64 fallback.
* **Tests**
  * Added coverage for rejecting non-TTY stdin and for skipping Windows ARM64 during postinstall in dry-run mode.
* **Chores**
  * Updated package license metadata to MIT.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/aymanbagabas/go-udiff v0.4.1
 	github.com/charmbracelet/colorprofile v0.4.3
 	github.com/charmbracelet/x/ansi v0.11.7
+	github.com/charmbracelet/x/term v0.2.2
 	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
 	golang.org/x/sys v0.46.0
 	mvdan.cc/sh/v3 v3.13.1
@@ -21,7 +22,6 @@ require (
 require (
 	github.com/charmbracelet/ultraviolet v0.0.0-20260615092913-2399af76d5b1 // indirect
 	github.com/charmbracelet/x/exp/golden v0.0.0-20260615092313-b57e5e6d29bb // indirect
-	github.com/charmbracelet/x/term v0.2.2 // indirect
 	github.com/charmbracelet/x/termios v0.1.1 // indirect
 	github.com/charmbracelet/x/windows v0.2.2 // indirect
 	github.com/clipperhouse/displaywidth v0.11.0 // indirect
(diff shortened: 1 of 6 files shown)


Vasanthdev2004:
Both addressed in 1febcb8:

1. **Real TTY check (run.go):** swapped the `os.Stdin.Stat()`/`ModeCharDevice` heuristic for `term.IsTerminal(os.Stdin.Fd())` (`github.com/charmbracelet/x/term`, already in the module graph via bubbletea — promoted to a direct dep, no new external module). It rejects pipes, regular files, and non-terminal char devices like `/dev/null`, and **fails closed** — anything not a verified terminal blocks the shell.
2. **Test (run_test.go):** now asserts the documented **exit code 2** from the guard, not just non-zero, so it proves the stdin guard is what fired.

build/vet/gofmt and `internal/tui` tests green.

PR #342, opened 2026-06-30
Document source-build sandbox helpers

anandh8x:
## Summary
- document the Linux sandbox helper binaries needed for source builds
- clarify macOS does not need an extra helper binary
- document Windows self-dispatch and optional release-style helper executables

## Tests
- git diff --check

## Summary by CodeRabbit

* **Documentation**
  * Expanded install instructions for source builds with platform-specific sandbox helper guidance.
  * Added Linux build and PATH setup steps for required helper binaries.
  * Clarified macOS and Windows source-build behavior, including optional Windows helper executables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,18 @@ If you are testing before the first public release, build from source:
 go build -o zero ./cmd/zero
 ```
 
+On Linux, build the sandbox helper too if you want native sandboxing:
+
+```bash
+go build -o zero-linux-sandbox ./cmd/zero-linux-sandbox
+go build -o zero-seccomp ./cmd/zero-seccomp   # optional compatibility wrapper
+```
+
+Put `zero` and `zero-linux-sandbox` in the same directory on `PATH`
+(`~/.local/bin` is a good default). macOS does not need an extra helper binary.
+Windows source builds can use the main `zero.exe` as their sandbox helper; release
+archives still ship standalone Windows helper executables.
+
 More install details: [docs/INSTALL.md](docs/INSTALL.md).
 
 ## First Run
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):
Verified every concrete claim against the tree — all accurate:

- All five referenced commands exist (`cmd/zero`, `cmd/zero-linux-sandbox`, `cmd/zero-seccomp`, `cmd/zero-windows-command-runner`, `cmd/zero-windows-sandbox-setup`), so the build instructions are correct.
- Linux native sandboxing → Bubblewrap (`BackendLinuxBwrap`, `internal/sandbox/adapters.go`) ✓
- `zero-seccomp` as a compat wrapper / the helper applying the seccomp filter itself matches `internal/sandbox/linux_helper.go` (`ApplySeccompThenExec`) ✓
- Windows self-dispatch via the running `zero.exe` matches `adapters.go:21` + the `windows missing helper exes self-dispatch` test ✓
- macOS using the system sandbox with no extra helper ✓

This fills a real gap (source builds silently lacked native sandboxing because the helpers weren't documented). CI green, `git diff --check` clean.

One optional improvement (non-blocking): the docs say to build the helper "if you want native sandboxing" but don't state what happens if you *don't* — Zero still runs, just with sandbox enforcement degraded (`EnforcementDegraded` / `BackendUnavailable` in `adapters.go`). A one-liner like "without the helper, Zero runs but sandbox enforcement is degraded/disabled" would set expectations. Also a minor nit: the Linux block in INSTALL.md repeats `go build -o zero ./cmd/zero` (already shown just above) — fine for a self-contained snippet, just noting it.

Approving.

PR #344, opened 2026-06-30
fix(swarm): constrain agent_type to the registered roster via a dynamic enum

gnanam1990:
## Summary

`swarm_spawn` (and `swarm_handoff`) described `agent_type` / `to_agent_type` as a free-form string with only an *"e.g. teammate, subagent"* hint in the description — no enum. So a model could pass an **unregistered type** (e.g. `"worker"`) and only find out it was invalid from the runtime error:

```
Error: swarm: unknown agent type; available agent types: subagent, teammate
```

That wastes tool calls (often several at once, fanning out a swarm) and hits weaker models hardest, since the valid set was never advertised in the schema.

## Fix
Set the parameter's `enum` **dynamically from the live roster** (`Registry.AgentTypes()`), via a shared `agentTypeProperty` helper used by both tools. The schema now advertises the valid set — the built-ins **plus any user-registered agents** — so an invalid type can't be sent in the first place.

- Dynamic, not hardcoded: custom agents registered via `Registry.Register` are reflected automatically.
- Falls back to a free-form string when the roster is unavailable/empty, so the tool never advertises an empty, un-satisfiable enum.
- The runtime `available agent types: …` error is kept as a backstop.

## Tests
- `TestSpawnAndHandoffAgentTypeEnumReflectsRoster`: both tools' enums equal the roster, include the built-ins, and pick up a newly-registered custom type.
- `TestAgentTypePropertyFallsBackWithoutRoster`: nil roster → no enum (no empty enum advertised).
- `go build`, `go vet ./...`, `go test ./... -race` (73 pkg), lint — all green.

## Summary by CodeRabbit

* **New Features**
  * Enhanced swarm tool agent-type fields to show selectable options based on registered agent types when available, with free-form text fallback when not.
* **Tests**
  * Added unit tests to ensure agent-type option lists match the swarm registry (including custom types) and fall back to plain string behavior when no registry is present.

### DIFF
--- a/internal/swarm/tools.go
+++ b/internal/swarm/tools.go
@@ -103,6 +103,25 @@ func errResult(format string, a ...any) tools.Result {
 	return tools.Result{Status: tools.StatusError, Output: "Error: " + fmt.Sprintf(format, a...)}
 }
 
+// agentTypeProperty builds the schema for an agent-type parameter, constraining it
+// to the roster's currently registered types (the built-ins plus any user-registered
+// agents) via an enum so a model can't spawn or hand off to an unknown type — the
+// schema teaches the valid set up front instead of the model learning it from a
+// failed call. Falls back to a free-form string if the roster is unavailable or
+// empty (so the tool never advertises an empty, un-satisfiable enum).
+func agentTypeProperty(sw *Swarm, description string) tools.PropertySchema {
+	prop := tools.PropertySchema{Type: "string", Description: description}
+	if sw == nil {
+		return prop
+	}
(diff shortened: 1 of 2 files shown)


gnanam1990:
Addressed — both comments were the same gap (the custom-type recheck only re-read `spawnTool`). Added the matching `researcher` assertion for `handoffTool`'s `to_agent_type` so both schemas stay covered when the roster changes. Gate green: build, vet, `go test ./... -race` (73 pkg), lint.

PR #347, opened 2026-06-30
fix(providerio): bound heartbeat-but-no-output streams so they can't hang forever

gnanam1990:
## Summary

A streaming request could **hang the agent forever**. The `providerio` SSE loop resets its 5-minute idle watchdog on **every** payload — including SSE keep-alive comments — by design (a heartbeating upstream isn't "dead", and aborting it had previously killed healthy long-running requests). The gap: a stream that **heartbeats while producing no real output** never trips the idle watchdog, and there's no other ceiling, so it hangs indefinitely.

Observed on **both** `chatgpt`/gpt-5.x (Codex Responses path) **and** `ollama`/minimax-m3 (chat path) — they share no provider code except this loop, which pinpointed the root cause to `providerio`.

## Fix
Add a second **content watchdog** next to the idle one, in the same loop:
- Keep-alives still reset the **idle** watchdog → no regression for heartbeating connections.
- Only **real data lines** reset the new **content** watchdog.
- If no data arrives for `streamContentStallFactor` (2) × the idle timeout (→ 10 min at the 5-min default), abort with `ErrStreamStalled` instead of hanging.

A slow-but-producing stream (emitting data between heartbeats) keeps resetting the content watchdog and is never killed. The change is **contained entirely to `internal/providers/providerio`** — no provider or caller changes, so it covers every provider at once.

## Tests
- `TestScanSSEDataWithContextAbortsOnContentStall`: keep-alives-only (idle never fires) → `ErrStreamStalled` + request context cancelled.
- `TestScanSSEDataWithContextContentResetsOnData`: data every 40ms past the content window → completes normally, never aborted.
- Existing idle/cancel tests unchanged. `go test ./... -race` (73 pkg, stable across `-count=5`), `go vet`, lint — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streaming responses that remain connected but stop delivering actual output.
  * Added detection for “content stalled” streams (heartbeats without real `data:`) and now stop with a clearer, stall-specific error message.
  * Updated Anthropic, Gemini, and OpenAI providers to consistently surface idle vs. stalled timeout reasons using shared timeout messaging.
* **Tests**
  * Added regression tests covering content-stall abort behavior and watchdog reset when real data continues arriving.

### DIFF
--- a/internal/providers/anthropic/provider.go
+++ b/internal/providers/anthropic/provider.go
@@ -211,11 +211,11 @@ func (provider *Provider) stream(ctx context.Context, body []byte, events chan<-
 	err = providerio.ScanSSEDataWithContext(streamCtx, cancelStream, response.Body, provider.streamIdleTimeout, func(data string) bool {
 		return provider.emitPayload(ctx, data, state, events)
 	})
-	if errors.Is(err, providerio.ErrStreamIdle) {
+	if errors.Is(err, providerio.ErrStreamIdle) || errors.Is(err, providerio.ErrStreamStalled) {
 		state.closeOpen(ctx, events)
 		providerio.SendEvent(ctx, events, zeroruntime.StreamEvent{
 			Type:  zeroruntime.StreamEventError,
-			Error: provider.redact(fmt.Sprintf("provider stream error: idle timeout after %s (upstream stopped sending data)", provider.streamIdleTimeout)),
+			Error: provider.redact("provider stream error: " + providerio.StreamTimeoutMessage(err, provider.streamIdleTimeout)),
 		})
 		return
 	}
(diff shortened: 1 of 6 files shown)


gnanam1990:
Addressed both:

1. **ErrStreamStalled now handled in the callers** — all four stream handlers (openai/codex/anthropic/gemini) treat it in the same branch as `ErrStreamIdle` via a new `providerio.StreamTimeoutMessage(err, idle)` helper, which gives the stall a **distinct, accurate** message (reports the content window = idle × factor, and does *not* say "upstream stopped sending data" — keep-alives were still arriving). No more falling through to the generic provider-error path.
2. **Flaky margin widened** — content-watchdog tests now use a 100ms idle with 30ms data/keepalive intervals, so CI/GC jitter can't trip the idle watchdog and mask the content-reset path. Added a `StreamTimeoutMessage` unit test.

Gate: `go test ./... -race` (73 pkg, stable across `-count=5`), vet, lint — all green.

gnanam1990:
@coderabbit rereview

PR #348, opened 2026-06-30
fix(tui): sub-agents inherit the parent's live provider after a /model switch

gnanam1990:
## Summary

Swarm members and Task sub-agents fail authentication the instant they spawn — `exit 3`, `0 tool calls`, *"API key is missing or invalid … can't reach the provider"* — so the spawned agent appears in the AGENTS panel and **immediately disappears** (it errored out in ~3s, it didn't complete).

### Root cause
Sub-agents run as child `zero exec` processes that **re-resolve the provider from `config.json` themselves** — they're passed `--model` but **never the parent's provider**. After you switch provider/model in the TUI via **`/model`** (an in-memory change that doesn't rewrite `config.json`), the child still resolves `config.json`'s **stale `activeProvider`**, lands on a provider whose credentials don't match the live model, and 401s.

This is provider-credential-agnostic in the symptom but splits by method:
- **Env-var key** (`OPENAI_API_KEY`, …) — already worked (the child inherits the env).
- **Saved key (credstore/keychain)** and **OAuth login (`zero auth`)** — broke, because the child re-resolved the *wrong profile*.

### Fix
On a `/model` provider switch, export **`ZERO_PROVIDER`** (`config.SetActiveProviderEnv`) to the live provider name. Child processes inherit the environment, and config resolution **already** maps `ZERO_PROVIDER → activeProvider` — so the child resolves the **same profile and its credentials** (env key / stored key / OAuth), covering **all three auth methods**, for **both Task and swarm** (both spawn through the specialist executor).

Startup needs no export — the parent and child both read `config.json`'s default, so they already match; only an in-memory switch diverges. (Deliberately not exporting at startup also avoids a process-global `os.Setenv` on the launch path.)

## Test plan
- New `TestSetActiveProviderEnvRoundTrips`: the helper exports the name, is a no-op for empty, and `applyEnv` reads it back via the child's `nil-map → os.Getenv` path.
- `go build`, `go vet ./...`, `go test ./... -race` (73 pkg), lint — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Switching providers or models now keeps any spawned/child processes aligned with the currently selected provider.
  * Provider selection via environment now behaves more reliably, including properly clearing the setting when no provider is selected.

* **Tests**
  * Added coverage to ensure the provider environment setting round-trips correctly and that inherited resolution matches the active provider.

### DIFF
--- a/internal/config/active_provider_env_test.go
+++ b/internal/config/active_provider_env_test.go
@@ -7,16 +7,31 @@ import (
 
 // SetActiveProviderEnv must export the provider name so a child process (whose
 // resolution reads the inherited environment via the nil-map -> os.Getenv path in
-// applyEnv) resolves the SAME active provider. An empty name is a no-op.
+// applyEnv) resolves the SAME active provider. A blank name CLEARS it.
 func TestSetActiveProviderEnvRoundTrips(t *testing.T) {
 	t.Setenv(ActiveProviderEnv, "") // register cleanup; restores the original after
 
 	if err := os.Unsetenv(ActiveProviderEnv); err != nil {
 		t.Fatalf("unset: %v", err)
 	}
-	SetActiveProviderEnv("") // empty must not set anything
+	SetActiveProviderEnv("") // blank with nothing set stays unset
 	if got := os.Getenv(ActiveProviderEnv); got != "" {
 		t.Fatalf("empty name set the env to %q, want unset", got)
+	}
+
+	// Regression: switching from a named provider back to an unnamed/default profile
+	// must CLEAR a previously exported value, not leave it stale for children.
+	if err := os.Setenv(ActiveProviderEnv, "stale-provider"); err != nil {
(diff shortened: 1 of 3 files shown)


gnanam1990:
@coderabbitai review

Addressed both review comments in ca748c2:

1. **Major (resolver.go)** — `SetActiveProviderEnv("")` now `os.Unsetenv`s `ZERO_PROVIDER` instead of leaving the previous value. Switching from a named provider back to an unnamed/default profile no longer exports the stale provider into child `zero exec` processes.
2. **Nit (active_provider_env_test.go)** — extended `TestSetActiveProviderEnvRoundTrips` with the stale-env regression case: seed `ZERO_PROVIDER`, call `SetActiveProviderEnv("")`, and assert both `os.Getenv` and `applyEnv` no longer resolve the old provider.

Gate: `go build`/`go vet`/`make lint` clean, `go test ./... -race` green (73 packages).

PR #349, opened 2026-06-30
fix: agent runtime resilience — stream stalls, weak-model tool args, and sub-agent reliability

gnanam1990:
A batch of runtime/agent reliability fixes surfaced by real sessions (a macOS stream hang, weak-model tool-call quality, and a high-parallelism swarm stress test). Grouped by theme:

### Streaming / transport resilience
- Defeat the macOS **stale-pooled-connection hang**: a POST reusing a dead keep-alive connection could hang forever (non-idempotent, so the stdlib won't auto-retry). Adds a tuned shared transport (`ResponseHeaderTimeout` 120s, `IdleConnTimeout` 30s).
- Agent-loop **stall-retry**: re-issue a turn that timed out (idle/stall) with no output yet, capped, so a transient stall doesn't kill the run.

### Weak-model tool-call tolerance
- Some models pack **multiple concatenated JSON objects** into one tool-call `arguments` slot (`{…}{…}`), which `json.Unmarshal` rejects with *"invalid character '{' after top-level value"*. We now recover the **first** object (and ignore further whole-JSON values) while still rejecting genuine trailing garbage; the replay-history sanitizer records the recovered object so dispatch and transcript agree.

### Turn budget
- `/turns <n>` — per-session tool-turn budget command (the interactive control `/mode` removal left missing); default `maxTurns` 30 → 50.
- Sub-agents/swarm members **inherit** the budget via `ZERO_MAX_TURNS` (clamped to a shared ceiling), so a delegated task gets the budget you set instead of re-resolving the default.

### Sub-agent / swarm reliability
- A swarm member whose child **exits non-zero** is now reported `[failed]`, not `[done]` (it previously showed as done with "0 failed"), keeping its session id for drill-in (`FailWithSession`).
- An **unknown specialist name** (a model inventing e.g. `validator-runner`) returns a corrective error listing the real specialists instead of an opaque "not found".
- A sub-agent **killed by a signal** surfaces the signal (e.g. *"killed before finishing (signal: killed)… most often out of memory when many sub-agents run in parallel"*) instead of an opaque `exit -1`.

### Testing
`go build` / `go vet` / `make lint` clean; `go test ./... -race` green (73 packages). New/extended unit tests cover each change.

> Note: the provider-inheritance fix that seeded this work already landed separately in #348.

## Summary by CodeRabbit

* **New Features**
  * Added a `/turns` session command to view and set the per-run tool-turn limit (validated, clamped to a safety ceiling) and propagated via environment for spawned agents.
* **Bug Fixes**
  * Improved streamed completion recovery for “no output” idle/stall timeouts via bounded retries.
  * Hardened tool-call JSON argument handling to recover concatenated JSON; malformed payloads now fail loudly.
  * Enhanced termination messaging for signal-based/abnormal kills with actionable guidance.
  * Swarm member failures now better preserve session context for drill-down.
* **Chores**
  * Increased the default per-run tool-turn budget and updated stall-hardened shared HTTP client behavior.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -280,6 +280,25 @@ func isContextLimitError(message string) bool {
 	return false
 }
 
+// isStreamTimeoutError reports whether a streamed error is a provider stream
+// idle/stall timeout (providerio surfaces these as "idle timeout after …" /
+// "no output for …" / "stream stalled"). Such a turn produced nothing, so when
+// no content was forwarded yet it can be safely re-issued on a fresh connection
+// (see the stall-retry in the agent loop). Substring + case-insensitive to
+// tolerate the prefixed "provider stream error: …" wrapping.
+func isStreamTimeoutError(message string) bool {
+	lowered := strings.ToLower(strings.TrimSpace(message))
+	if lowered == "" {
+		return false
+	}
+	for _, needle := range []string{"idle timeout after", "no output for", "stream stalled"} {
+		if strings.Contains(lowered, needle) {
+			return true
+		}
+	}
+	return false
+}
+
 // compactionState carries the per-run state the agent loop needs to compact a
 // conversation safely. It is created once per Run and is a no-op whenever
 // options.ContextWindow <= 0.
--- a/internal/agent/decode_args_test.go
+++ b/internal/agent/decode_args_test.go
@@ -1,0 +1,99 @@
+package agent
+
+import "testing"
+
+// decodeToolArguments must (a) decode a normal single-object payload unchanged,
+// (b) tolerate a weak model packing multiple concatenated top-level JSON objects
+// into one arguments string by decoding the FIRST object and ignoring the rest
+// (the minimax-m3 "invalid character '{' after top-level value" failure), and
+// (c) still reject a genuinely malformed/truncated first object.
+func TestDecodeToolArguments(t *testing.T) {
+	t.Run("single object decodes normally", func(t *testing.T) {
+		var a map[string]any
+		if err := decodeToolArguments(`{"path":"x"}`, &a); err != nil {
+			t.Fatalf("err = %v, want nil", err)
+		}
+		if a["path"] != "x" {
+			t.Fatalf("decoded %v, want path=x", a)
+		}
+	})
+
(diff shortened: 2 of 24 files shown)


gnanam1990:
@coderabbitai review

Addressed all 6 comments in d098a87:

1. **loop.go:299 (forwarded-output gate)** — stall-retry now gates on a turn-level `forwardedAnything` flag set by wrapped OnText/OnReasoning/OnUsage/OnToolCallStart/OnToolCallDelta callbacks, so it never re-streams on top of output forwarded earlier in the turn (including across the reactive-compaction retry). New `TestRunDoesNotRetryStallAfterForwardedReasoning` covers the empty-`collected.Text`-but-forwarded case.
2. **loop.go:303 (error recovery routing)** — extracted the image-rejection + context-limit recovery into a shared `recoverStreamError` closure and route a reissued stall-retry stream's non-stall error through it instead of returning raw.
3. **exec.go:618 (specialist guidance)** — dropped the hardcoded worker/explorer/code-review names; the error now relies solely on the dynamic `availableSpecialistList(result)`.
4. **streamer.go:156 (OOM over-diagnosis)** — message now lists OOM/timeout/cancellation evenhandedly and says "check the signal to tell which".
5. **launcher_specialist.go:68 (session on post-start failure)** — `exec.go` returns `ExecResult{SessionID: built.SessionID}` on a post-start `runChild` error and the launcher preserves it on the `err != nil` path, so `FailWithSession` keeps those members drillable.
6. **model.go:3472 (/turns while active)** — `/turns` is now blocked while a run is pending (the budget applies next run).

Gate: `go build`/`go vet`/`make lint` clean, `go test ./... -race` green (73 packages), incl. the preserved `TestRunReactiveRetryDoesNotDoubleEmitText`.

gnanam1990:
Follow-up (279c31d): on a self-review of comment #1, I left **OnUsage OUT** of the `forwardedAnything` flag. Usage is telemetry, not visible output, so a provider that emits an early usage event (e.g. prompt-token count) then stalls with no content must still be safely retried — only OnText/OnReasoning/OnToolCallStart/OnToolCallDelta mark forwarded output. Tests still green (73 pkg, -race).

PR #350, opened 2026-07-01
fix(tui): subchat mouse selection, hover highlighting, and drag-to-edge auto-scroll

gnanam1990:
Mouse UX fixes and a new hover-highlight feature for the subagent/swarm drill-in (subchat) view and transcript text selection, found and fixed while dogfooding the previous PR's binary.

### Subchat drill-in fixes
- **Can't copy anything while viewing a subagent/swarm's child session.** Mouse hit-testing and the copy-time text extraction always resolved against the PARENT transcript, never the subchat's child rows, even though the screen was showing the child session — selection either matched nothing or the wrong text.
- **Plan panel shown above the composer inside the subchat view.** `footerView` unconditionally rendered the parent run's pinned plan panel; only the ask-user modal was exempted.
- **Selection didn't extend across a scroll.** The drag-selection cursor was only ever updated by mouse *motion* events; a wheel-scroll is a distinct event type that never reached that code, so scrolling while selecting left the selection frozen wherever the mouse last physically moved.

### New: hover highlighting
Moving the cursor over any clickable row now highlights it before you click — specialist cards, collapse/expand toggle headers, AGENTS sidebar rows, PLAN sidebar steps, and permission-prompt options (the last one reuses the *existing* keyboard-cursor highlight instead of new styling). Requires switching mouse tracking from `CellMotion` (only reports motion while a button is held) to `AllMotion`; the existing 15ms mouse-event throttle already covers the added events.

Sidebar hover is resolved by **stable identity** (`sessionID`/`stepIndex`), not a cached line offset — re-matched against a fresh hit list on every render, so a row that's disappeared (a swarm member's linger elapsed, a plan step completed) simply doesn't highlight rather than a coincidentally-matching unrelated row lighting up.

### Drag-to-edge auto-scroll
Dragging a selection past the top/bottom of the visible transcript now auto-scrolls toward that side and extends the selection to follow (the standard browser/editor affordance). Implemented as a self-scheduling smooth glide (small step, ~70ms cadence) rather than one big jump per raw mouse-motion event, since raw motion events are bursty and stop entirely the instant the physical mouse holds still.

### Process notes
Two rounds of adversarial multi-agent review caught real bugs before they shipped:
- Plan-step hover stored the step's own index but the sidebar render treated it as a raw line offset — every plan-step hover highlighted the wrong row. Fixed via the stable-identity redesign above.
- A keypress mid-drag clears the selection through a different path than mouse-release, leaving the auto-scroll's internal state stale — which then silently disabled the next same-direction drag's auto-scroll. Fixed by resetting that state at every fresh press.

## Test plan
- [x] `go build ./...` / `go vet ./...` / `make lint` clean
- [x] `go test ./... -race` green (73 packages)
- [x] Unit tests for every fix: subchat selection resolves the child session's text not the parent's; plan panel absent from the footer during subchat; hover resolves the correct target across all clickable surfaces with the correct priority order; hover changes actual render output; drag-to-edge scrolls and extends in both directions; the glide chain terminates cleanly on release/return-to-body/off-to-side and survives a mid-drag keypress interrupt
- [ ] Manual verification in a real terminal (mouse/color perception isn't something automated tests can confirm) — click into a subagent/swarm card, drag-select and copy in that view, confirm the plan panel stays hidden there, hover over each clickable surface, drag past the top/bottom edge and confirm the glide feels smooth

## Summary by CodeRabbit

* **New Features**
  * Added hover highlighting for interactive transcript rows, sidebar entries, and plan steps, including permission option cursor updates.
  * Improved selection edge-glide auto-scroll while dragging (mouse hover/selection stay in sync).
* **Bug Fixes**
  * Prevented stale hover positioning after clearing, resuming/rewinding/compacting, resizing, paging, and subchat exit.
  * Fixed hover/selection behavior when entering subchat drill-in, including using the correct transcript rows.
  * Hid the pinned plan panel while viewing subchat drill-in.
* **Tests**
  * Expanded regression tests for hover targeting, edge scrolling glide lifecycle, and subchat selection/visibility.

### DIFF
--- a/internal/tui/flush.go
+++ b/internal/tui/flush.go
@@ -139,6 +139,10 @@ func (m model) drainFlushQueue() (model, tea.Cmd) {
 // surface ended.
 func (m *model) resetFlushFrontier(divider string) {
 	m.flushed = 0
+	// Renumbers every row's bodyY from the top (used by /clear, /resume, /rewind,
+	// /compact); a transcript-hover target's bodyY would otherwise risk
+	// coincidentally matching an unrelated row in the rebuilt transcript.
+	m.hover = hoverTarget{}
 	if divider != "" && m.flushedAny {
 		m.flushQueue = append(m.flushQueue, zeroTheme.faint.Render(divider))
 	}
--- a/internal/tui/hover.go
+++ b/internal/tui/hover.go
@@ -1,0 +1,117 @@
+package tui
+
+import tea "charm.land/bubbletea/v2"
+
+// hoverKind discriminates which clickable surface a hoverTarget refers to.
+type hoverKind int
+
+const (
+	hoverNone hoverKind = iota
+	// hoverTranscript: a specialist card or collapse/expand toggle header in the
+	// main transcript body (or the subchat child view), identified by bodyY.
+	hoverTranscript
+	// hoverSidebarAgent: a clickable AGENTS sidebar row, identified by sessionID.
+	hoverSidebarAgent
+	// hoverPlanStep: a plan step row in the sidebar's PLAN section, identified by
+	// step index.
+	hoverPlanStep
+)
+
+// hoverTarget identifies the single clickable row (if any) currently under the
+// mouse cursor with no button pressed. Exactly one of bodyY/sessionID/stepIndex is
+// meaningful, discriminated by kind.
+//
+// The sidebar fields deliberately store a STABLE IDENTITY (sessionID, stepIndex),
+// not a raw line offset: the AGENTS/PLAN section sizes can change between when a
+// hover is detected and when the sidebar next renders (a swarm member's linger
+// window elapsing, a plan step completing) with no mouse motion in between to
+// re-resolve it. A cached raw offset would then silently point at whatever
+// unrelated row now occupies that slot. hoveredSidebarLineOffset (sidebar.go)
+// re-resolves the CURRENT line offset for this identity fresh on every render, so
+// a row that's gone simply doesn't highlight rather than a wrong one lighting up.
+type hoverTarget struct {
+	kind      hoverKind
+	bodyY     int    // hoverTranscript
+	sessionID string // hoverSidebarAgent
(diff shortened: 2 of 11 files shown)


gnanam1990:
Addressed in e24203b:

**Smoke test failures (macOS + Ubuntu)** — the 4 new edge-glide tests implicitly relied on `m.reducedMotion == false`, which `mouseTestModel()` inherits from TTY detection (`defaultReducedMotion()` forces `true` under `colorprofile.NoTTY`, which CI's non-TTY runners trigger but a local interactive shell doesn't). Set `reducedMotion = false` explicitly in each test. Verified by reproducing the CI condition locally via `ZERO_REDUCED_MOTION=1`.

**CodeRabbit actionable comment** — `handleTranscriptSelectionMouse`'s `mouseMotion` case could misroute a genuine hover (arriving during the async copy-command grace window after release, where `transcriptSelection.active` stays true) as a drag update. The literal suggested fix (gate on the event's own `Button` field) turned out to be wrong — `TestTranscriptSelectionUpdatesOnGenericMotion` is a pre-existing, deliberate test guarding against terminals that don't restate the button on every motion event during a real drag, and gating on `Button` broke it. Fixed properly with a `dragging` field on `transcriptSelectionState` that tracks the app's own press/release bracket instead of trusting any single event's button field. New regression test: `TestTranscriptHoverAfterReleaseDoesNotMoveSelection`.

**CodeRabbit nitpicks** — routed the two `hoverTarget{}` subchat-exit resets through the existing `clearHover()` helper. Left the hit-test-layout-per-hover-tick perf nitpick as a documented follow-up (a correct fix needs its own invalidation-on-scroll/resize/content-change logic, and the reviewer's own framing was "trivial/poor tradeoff" — not worth rushing given this exact staleness-bug class already bit this branch twice).

Gate: `go build`/`go vet`/`make lint` clean, `go test ./... -race` green (73 packages), incl. the previously-broken test now passing.

PR #356, opened 2026-07-01
fix(tui): context-fill gauge for custom/local Ollama models + related fixes

gnanam1990:
## Summary
- **Context gauge missing for custom/local Ollama models.** The colored context-fill gauge only renders when Zero knows a model's context window (curated catalog or live `/v1/models` discovery). Neither source works for a custom local Ollama tag — `parseModelsResponse` never extracts context-window metadata from the generic `/v1/models` response, and a custom tag has no curated-catalog entry either. Added `DiscoverOllamaContextWindow`, an async probe against Ollama's native `/api/show` endpoint (which does report it, under an architecture-prefixed `model_info` key like `qwen2.context_length`), scoped to the local `ollama` catalog ID only. Also fixed a related gap: switching providers mid-session (`switchProviderModel`) previously warmed no discovery at all for the newly active provider.
- **Token count duplicated next to the gauge.** `contextWindowSegment()` (the gauge) and `usageStatusSegment()` (the plain "N tok" figure) both read the exact same value. The code already suppressed the plain figure when the sidebar shows it, but never extended that dedup to the gauge — so once a model's window became known, the footer showed the same number twice. Fixed by falling back to cost-only whenever the gauge is shown.
- **`exec_command`'s `collect()` deadline check.** The poll deadline was only checked in the branch reached when the output buffer drains empty — a continuously-writing background process could theoretically starve that check. Fixed defensively (checked in both branches now). Investigated as part of a cross-provider "stuck session" report; could not reproduce an actual hang via a synthetic continuous-writer test, so this closes a real gap without being a confirmed root cause of that report.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/tui/ ./internal/providermodeldiscovery/ ./internal/tools/ -count=1` (new tests: `TestDiscoverOllamaContextWindow*`, `TestOllamaContextWindowDiscoveryCmdScopedToLocalOllama`, `TestOllamaContextWindowDiscoveredMsgPopulatesMap`, `TestSwitchProviderModelWarmsDiscoveryForTheNewProvider`, `TestStatusLineDoesNotDuplicateTokenFigureNextToGauge`, `TestCollectRespectsDeadlineUnderContinuousOutput`)
- [x] `go test ./... -race -count=1` (full repo)
- [ ] Manual verification against a live Ollama daemon (the `/api/show` request/response shape was built against Ollama's documented API, not tested against a running instance)

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and caching context-window sizes for local Ollama models, helping the app show more accurate capacity information.
  * Switching providers can now trigger background discovery so model details load sooner.

* **Bug Fixes**
  * Fixed command execution so time limits are respected even when a process keeps producing output.
  * Improved status-line rendering to avoid duplicate token usage display when the gauge is already shown.

### DIFF
--- a/internal/providermodeldiscovery/discovery.go
+++ b/internal/providermodeldiscovery/discovery.go
@@ -85,6 +85,104 @@ func Discover(ctx context.Context, profile config.ProviderProfile, options Optio
 	}
 }
 
+// DiscoverOllamaContextWindow asks a local Ollama daemon for a model's context
+// length via its native /api/show endpoint. The generic /v1/models probe
+// (parseModelsResponse) only ever returns id/description — OpenAI-compatible
+// listings don't carry context-window metadata, and a custom/local model tag
+// (e.g. a user's own Ollama pull, including a ":cloud"-tagged model proxied
+// through the local daemon) has no curated-catalog entry to borrow one from
+// either, so modelContextWindow has nothing to show for it without this. Only
+// meaningful for the local Ollama provider (baseURL like
+// http://localhost:11434/v1) — Ollama Cloud's hosted API is a different
+// service and isn't assumed to expose the same endpoint.
+func DiscoverOllamaContextWindow(ctx context.Context, baseURL string, model string, options Options) (int, error) {
+	model = strings.TrimSpace(model)
+	if model == "" {
+		return 0, fmt.Errorf("model name is required")
+	}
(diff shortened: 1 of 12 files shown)


gnanam1990:
@coderabbit review

PR #362, opened 2026-07-01
fix: cut the content-stall dead-wait and auto-retry a mid-tool-call stall

gnanam1990:
## Summary
Root-caused the chatgpt/gpt-5.5 (and ollama/minimax) "stuck" reports with a **live goroutine stack dump** during a headless reproduction: the cloud model backend sends `200 OK` + ~300 tokens, then stops sending SSE body bytes mid-generation of a large single tool-call output (a ~1000-line `write_file`). Every Zero goroutine was cleanly parked on the socket read — no channel-send block, no deadlock, no tool/sandbox hang. Confirmed on **two** backends across **two** transports (HTTP/2 Codex parser + HTTP/1.1 OpenAI parser), both blocked identically on the response-body `Read`. It's an upstream limitation, not a Zero bug or connection-pooling issue.

Zero already detected it, but two UX gaps made it feel like a hang. This PR closes both:

1. **Faster abort.** The content-stall watchdog fired at `idle × 2` = 10 minutes. A heartbeat/body-pause stall on these backends rarely recovers, so 10 minutes of a frozen counter was a terrible wait for a doomed turn. Replaced the `StreamContentStallFactor = 2` const with `ContentStallTimeout(idle) = idle × 1.2` (6 minutes at the default). It still only ever fires on a **true zero-output freeze** — any real data line resets it — so healthy slow-reasoning runs are unaffected. Scales with `ZERO_STREAM_IDLE_TIMEOUT`; returns 0 (disabled) when idle is disabled.

2. **Auto-retry the actual stall.** The existing stall-retry only fired when **nothing** had been forwarded — so the common case (model streamed transient reasoning + began the big `write_file`, then froze mid-arguments) was **not** retried, because the incomplete tool call is collected and transient output was forwarded. Relaxed the gate to retry when no visible **prose** was shown (`forwardedVisibleText`, marked only by `OnText`) and no final text was collected — dropping the `len(ToolCalls)==0` exclusion, since an incomplete tool call from a timed-out stream is never executed (the turn returns on the error before dispatch) and never appended to `messages`, so the retry re-sends clean context with no duplication. A visible `[no output — model stalled; retrying N/max]` notice fires per attempt. Forwarded prose still blocks the retry (it would duplicate answer text).

## Verification
- Full `-race` suite green.
- A 4-dimension root-cause audit + a 3-dimension adversarial review of this diff both passed with **no confirmed defects**.
- Reproduced the stall live (headless + `SIGQUIT` goroutine dump) on both chatgpt/gpt-5.5 and ollama/minimax-m3:cloud.

## Test plan
- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./... -race -count=1`
- [x] New tests: `TestRunRetriesStallAfterIncompleteToolCall` (the exact gpt-5.x case), `TestRunRetriesStallAfterOnlyReasoning`; preserved `TestRunDoesNotRetryStallAfterPartialOutput` (prose still blocks retry); updated content-stall timeout tests for the 1.2× factor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stall recovery so retries are triggered more reliably when the stream times out before any final answer text appears, including cases with only interim reasoning or an in-progress tool call.
  * Updated “no output” stall detection and messaging to use a consistent, clamped content-stall window and correct timeout wording.
  * Refined quiet-generation UI cues so the displayed recovery estimate matches the actual content-stall threshold, improving user confidence in retry timing.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -210,27 +210,29 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			}
 		}
 
-		// Wrapped callbacks flag ANY output forwarded this turn, so a stall retry
-		// never re-streams on top of output the user already saw — even across the
-		// reactive-compaction retry below, which overwrites `collected`.
-		forwardedAnything := false
-		mark := func() { forwardedAnything = true }
-		// OnUsage is telemetry, not visible output, so it does NOT mark
-		// forwardedAnything: a provider that emits an early usage event (e.g. prompt
-		// tokens) then stalls with no content must still be safely retried. The gate
-		// only guards against re-streaming output the user actually SAW.
+		// forwardedVisibleText flags forwarded final PROSE (OnText) specifically —
+		// the only output whose re-stream on a stall retry would visibly DUPLICATE
+		// answer text the user already read. Transient working indicators
+		// (reasoning previews, tool-call "writing…" previews) are deliberately NOT
+		// counted: a turn that streamed only those then stalled — the common
+		// gpt-5.x / ollama "froze mid-write_file" case — is safe to re-issue, since
+		// the TUI resets the tool-call preview on the next tool-call-start and a
+		// stalled turn is never appended to `messages` (it returns on the error
+		// before the append), so the retry re-sends clean context with no
+		// conversation-state duplication.
+		forwardedVisibleText := false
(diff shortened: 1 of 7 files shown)


gnanam1990:
@coderabbit review

PR #365, opened 2026-07-02
feat(tui): FILES sidebar panel with click-to-select and file drill-in

kevincodex1:
Summary

Adds a FILES section to the TUI's right context sidebar (between PLAN and ACTIVITY) showing the workspace files the session has touched — newest first, with an A/M/✗ status badge, an aggregated +added −removed diffstat per file, a pulsing row for the file whose write is streaming right now, and a +N more overflow trailer.

Why: while an agent works, the question users actually have is "what is it changing right now, and what has it changed so far?" Today the transcript only answers that by scrolling back through tool cards; this panel answers it at a glance and gives every file a direct path to its diffs.

Interaction
- First click on a row selects the file — accent ▸ on the row, its edit cards tint in the chat, and the transcript scrolls to the most recent one. Esc clears the selection.
- Second click opens a drill-in view (reuses the subchat pattern): the chat column swaps to the file's stacked full-depth edit cards (edit N of M); f toggles to the syntax-highlighted on-disk file with ▎ markers on session-added lines, d back to diff, Esc returns to the chat at its saved scroll position. d/f only fire while the composer is empty, so typing is never hijacked.

Data sources
1. changedFiles carried by write_file/edit_file/apply_patch results — persisted to the session payload and restored on rehydrate, so the panel survives /resume. The roster is derived on demand from transcript rows (the same pattern as the swarm AGENTS roster), so it can never drift from the chat.
2. A git sweep for mutations the tool stream can't see — agents that scaffold via exec_command/bash (npm create, heredocs) or delegate to subagents. A baseline git status --porcelain --untracked-files=all snapshot at startup hides pre-existing dirt; re-sweeps after each command-tool result and at turn end merge newly dirty paths, with a git diff HEAD --numstat diffstat. Sweeps are single-flight, 3s-timeboxed, and latch off in non-git workspaces. Git-only files open the drill-in directly on the full-file view (they have no edit cards to stack).

Implementation notes
- The drill-in swaps the body inside transcriptBodyItems and replaces the pinned title bar with a one-line nav, so the scroll engine, viewport, and mouse hit-testing all stay consistent with no special-casing.
- The selection tint is part of the render-cache key, so selecting/deselecting a file can't serve stale cards.
- FILES renders below PLAN, so existing plan-step click offsets are untouched; its own hit list accounts for the sections above it.

Checklist

- [ ] The linked issue already has the issue-approved label.
- [x] go build ./..., go vet ./..., and go test ./... pass locally.
- [x] gofmt clean.
- [x] Tests added/updated for the change (and run under -race where relevant). — 22 new tests across files_panel_test.go, file_view_test.go, and files_git_sweep_test.go, including a sweep test against a real temporary git repo; go test -race ./internal/tui/ passes.
- [x] UI changes include screenshots or a short recording where possible. — terminal captures below.

## Summary by CodeRabbit

* **New Features**
  * Added a **FILES** sidebar that lists touched workspace files with change counts, plus a live “editing” indicator.
  * Added a drill-in file view with **diff** and **full** modes, including line numbers, truncation for long files, and session-based change markers.
  * The **FILES** sidebar is interactive: selecting a file highlights related transcript cards and supports click-to-open.
  * Added git-backed detection for workspace mutations that may occur outside normal tooling.
* **Bug Fixes**
  * File view now preserves/restores the parent chat scroll correctly and exits cleanly with key controls.
  * Hover, selection, and nested view interactions are now consistently synchronized.

### DIFF
--- a/internal/tui/file_view.go
+++ b/internal/tui/file_view.go
@@ -14,6 +14,7 @@
 package tui
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -43,7 +44,13 @@ type fileViewState struct {
 // openFileView activates the drill-in for path in diff mode. Opening from an
 // already-open view (clicking another FILES row) keeps the original saved chat
 // scroll position rather than saving the file view's own offset as "parent".
+// Re-opening the file that is ALREADY being viewed is a no-op: a stray
+// re-click must not bounce the user from full mode back to diff or reset
+// their scroll position.
 func (m model) openFileView(path string) model {
+	if m.fileView.active && m.fileView.path == path {
+		return m
+	}
 	if !m.fileView.active {
 		m.fileView.parentScrollOffset = m.chatScrollOffset
 	}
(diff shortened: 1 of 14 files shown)


gnanam1990:
@coderabbit review

Vasanthdev2004:
@coderabbitai approve

PR #366, opened 2026-07-02
fix(auth): propagate credentials to every provider-build surface and pin children to the live provider

gnanam1990:
## Problem

A pre-launch audit found four credential-propagation defects, all the same class: a runtime provider gets built somewhere without the auth the user's live session is actually running with.

1. **Stored API keys never reached ACP or escalation builds.** The credstore key was folded in only inside the TUI/exec startup path's local copy. An `apiKeyStored` profile (the default onboarding outcome) built via the ACP surface or exec's mid-run `--allow-escalation` model switch went out **keyless** and 401'd.
2. **Spawned children could land on the wrong provider.** Sub-agents and swarm members re-exec the binary and re-resolve `config.json` at spawn time. A provider switch in the wizard/onboarding didn't reliably export `ZERO_PROVIDER`, and nothing exported it at plain startup — so a switch persisted by another process mid-session silently moved new children onto a different provider (and different credentials) than the parent.
3. **OAuth logins were keyed by profile name only.** A profile renamed away from its catalog ID (e.g. `{name: "codex", catalogID: "chatgpt"}`) never found the `zero auth login chatgpt` token — its model calls and all its children ran unauthenticated.
4. **The Codex `chatgpt-account-id` header could come from a different login than the bearer token** (two independent store selections), a pairing the backend rejects.

## Fix

**One wrap, one selection, one predicate:**

- `fillAppDeps` wraps `deps.newProvider` **once** to fold the stored API key in centrally. `buildProvider`, the ACP builder, exec's escalation switcher, and any future caller are covered with no per-site invariant to forget.
- `ZERO_PROVIDER` is exported at **startup** (`buildProvider`, injected via `appDeps` so tests never mutate process env) and re-exported on every runtime switch (wizard, onboarding, `/model`, `/provider`). Children are pinned to the parent's live provider from launch. A failed build or persist never moves the pin.
- `ProviderProfile.OAuthLoginCandidates()` is the single source of login candidates: profile name first, catalog-ID fallback, **nothing** for a profile that has its own configured credential — so a catalog-shared or same-name OAuth login can never silently override an explicit API key (the bearer path erases the key once a token resolves). Candidates are deduped case-sensitively to match the case-sensitive token store.
- The login is resolved **once** (`oauth.FirstStored`) in the CLI; the bound store key is threaded through `providers.Options.OAuthLoginKey` so the Codex account-id header always reads the account from the exact login that issued the bearer token — no second independent selection to diverge from it, while in-place token refreshes are still picked up per request.
- `ProviderProfile.HasConfiguredCredential()` replaces three verbatim copies of the "profile is key-authed" predicate across cli/tui/config.
- `applyProviderWizard` stages the build and persist into locals and commits live state (provider, profile, name, env export) only after **both** succeed; a persist failure leaves the session fully on the old provider.

## Tests

Red-green coverage for every path (reverting each fix reproduces the original symptom):

- stored key through the central wrap, ACP surface, and escalation switch (`TestFillAppDepsWrapsNewProviderWithStoredKey`, `TestRunExecEscalationSwitchKeepsStoredKey`)
- startup + onboarding + wizard `ZERO_PROVIDER` export, incl. failed-build/failed-persist no-ops (`TestBuildProviderExportsActiveProviderEnv`, `TestCompleteSetupExportsActiveProviderEnv`, `TestApplyProviderWizardExportsActiveProviderEnv`, `TestApplyProviderWizardPersistFailureLeavesLiveStateUnchanged`)
- OAuth candidate gating: renamed-profile fallback, case-variant names, explicit-key / auth-header / stored-key protection, env-only profiles still eligible (`TestOAuthLoginCandidates`, `TestOAuthResolverForProfileFallsBackToCatalogID`, `TestOAuthResolverForProfileDoesNotOverrideExplicitKey`)
- Codex account/bearer same-login guarantee incl. refresh-in-place (`TestCodexAccountForKey`, `TestNewRoutesChatGPTCatalogWithStoredAccountID`)

Full suite: `go test ./... -race -count=1` → 73/73 packages green.

## Summary by CodeRabbit

* **New Features**
  * Provider setup and provider switching now export the active provider for child processes, reducing stale-provider behavior during runs and wizard flows.
  * OAuth logins can now be resolved more reliably across renamed or case-variant profiles, with improved fallback behavior.

* **Bug Fixes**
  * Stored API keys are applied consistently across provider creation paths, including escalation flows.
  * Credential checks now use centralized detection for more accurate “ready” validation.
  * Provider selection no longer overrides explicitly configured credentials with fallback login data.

### DIFF
--- a/internal/cli/acp.go
+++ b/internal/cli/acp.go
@@ -43,8 +43,11 @@ func runACP(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int
 	conn := acp.NewConn(deps.stdin, stdout)
 	acp.NewAgent(conn, acp.Deps{
 		ResolveConfig: deps.resolveConfig,
-		NewProvider:   deps.newProvider,
-		RunAgent:      agent.Run,
+		// deps.newProvider is wrapped in fillAppDeps to apply the stored API key,
+		// so ACP is authenticated for apiKeyStored profiles like every other
+		// surface — no ACP-specific credential handling needed.
+		NewProvider: deps.newProvider,
+		RunAgent:    agent.Run,
 		// Build the SCOPED registry + sandbox engine per workspace, exactly like the
 		// exec surface, so ACP shell/file tools are confined — never run unconfined.
 		BuildWorkspace: func(workspaceRoot string, resolved config.ResolvedConfig) (*tools.Registry, *sandbox.Engine, error) {
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -43,12 +43,17 @@ import (
 var version = "dev"
 
 type appDeps struct {
-	getwd                 func() (string, error)
-	stdin                 io.Reader
-	userConfigPath        func() (string, error)
-	resolveConfig         func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
-	resolveMCPConfig      func(workspaceRoot string) (config.MCPConfig, error)
-	newProvider           func(config.ProviderProfile) (zeroruntime.Provider, error)
+	getwd            func() (string, error)
+	stdin            io.Reader
+	userConfigPath   func() (string, error)
+	resolveConfig    func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
+	resolveMCPConfig func(workspaceRoot string) (config.MCPConfig, error)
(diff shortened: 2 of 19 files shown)


gnanam1990:
@coderabbit review

PR #372, opened 2026-07-02
fix(config): create the right config file on first setup (macOS ~/.config) (#371)

### LINKED ISSUE #371: create the right config file on the first setup
### Version / branch / commit

zero dev

### OS and environment

macOS 26.5.1/ Ghostty/Go1.27rc1

### Steps to reproduce

1. git clone ....
2. go run cmd/zero or go run cmd/zero setup

### Expected behavior

create the initial config.json and setup the first provider

### Actual behavior

active provider "" not found

Image

smallnest:
Fixes #371.

## Problem

On macOS, a fresh `go run cmd/zero setup` failed with `active provider "" not found`. Two root causes:

1. **Wrong config path on macOS.** `DefaultUserConfigPath` used Go's `os.UserConfigDir()`, which returns `~/Library/Application Support` on darwin — but users (and Claude Code) put config at `~/.config/zero/config.json`. The user config was never read, so `activeProvider` stayed empty.
2. **No fallback to setup.** When resolve failed because no active provider could be found, the interactive TUI exited fatally instead of dropping into the setup wizard.

## Changes

- **`~/.config` on macOS.** Add `config.UserConfigDir()`: on darwin it resolves to `~/.config` (honoring `XDG_CONFIG_HOME`), matching Linux and Claude Code. `DefaultUserConfigPath`, the TUI user-commands dir (`internal/tui/model.go`), and specialist manifests (`internal/specialist/manifest.go`) all route through it, so credential store / specialists / user commands land in the same tree.
- **Enter setup instead of crashing.** Add a `config.ErrNoActiveProvider` sentinel (the two "active provider not found" errors are now `%w`-wrapped, message unchanged). The interactive TUI (`internal/cli/app.go`) drops into the setup wizard when resolve fails *solely* due to a missing/unresolvable active provider. Other resolve errors (malformed JSON, directory conflicts) remain fatal.

The config directory is already auto-created on first write — every writer (`writeConfigFile`, credstore, oauth store) does `os.MkdirAll(dir, 0o700)` before writing.

## Testing

- `go build ./...`
- `go test ./internal/config/... ./internal/cli/... ./internal/tui/... ./internal/specialist/...`
- New tests: darwin `DefaultUserConfigPath` → `~/.config/zero/config.json`; `errors.Is(err, ErrNoActiveProvider)`; TUI enters setup on `ErrNoActiveProvider` and stays fatal on other errors.
- Pre-existing `TestLocalControl*` / `TestResolveLocalControl*` failures in `internal/config` are unrelated to this change (verified by diffing failures against `main`).

Manual: place config at `~/.config/zero/config.json` and run `zero` — starts normally; clear `activeProvider` / remove providers and run `zero` — enters the setup wizard instead of erroring.

## Summary by CodeRabbit

* **New Features**
  * The app now opens the setup flow automatically when no active provider is available, instead of exiting with an error.
  * macOS users now get more consistent config path handling, including support for the standard XDG config location when available.

* **Bug Fixes**
  * Improved error handling so unrelated configuration issues still fail normally, with the error shown to the user.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -530,7 +530,15 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 
 	resolved, err := deps.resolveConfig(workspaceRoot, config.Overrides{})
 	if err != nil {
-		return writeAppError(stderr, err.Error(), 1)
+		// A resolve failure caused solely by a missing/unresolvable active provider
+		// is not fatal for the interactive TUI: drop into the setup wizard with an
+		// empty config so the user can onboard, instead of exiting with an error.
+		// Any other error (malformed JSON, directory conflict, etc.) is still fatal.
+		if !errors.Is(err, config.ErrNoActiveProvider) {
+			return writeAppError(stderr, err.Error(), 1)
+		}
+		resolved = config.ResolvedConfig{}
+		forceSetup = true
 	}
 	userConfigPath, err := deps.userConfigPath()
 	if err != nil {
--- a/internal/cli/app_test.go
+++ b/internal/cli/app_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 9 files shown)


kevincodex1 (approved):
thanks for contribution! looks good to me

Vasanthdev2004 (approved):
## Verdict: Approve ✅ (two low-severity nits, none blocking)

Reviewed both root-cause fixes against the tree, built, vetted, and ran the new tests locally on the branch.

### Correctness — solid
- **`~/.config` on macOS.** `config.UserConfigDir()` is the right abstraction, and I confirmed it's the *only* remaining `os.UserConfigDir()` caller outside the helper — credential store (`config/credentials.go` → `DefaultUserConfigPath`), specialists, and user-commands all route through it. Notably this **fixes a pre-existing split-brain**: `oauth.ResolveStorePath` already stored tokens under `~/.config/zero/` on every OS, so before this PR macOS had `config.json` in `~/Library/Application Support` but OAuth tokens in `~/.config`. After the PR everything lands in one tree. ✔
- **Fallback to setup.** The `ErrNoActiveProvider` sentinel wraps both error sites with `%w` (message unchanged, asserted by the existing test), and the app.go gate correctly stays fatal for every *other* resolve error. I verified the fallback is **nil-safe**: with `config.ResolvedConfig{}` the empty provider makes `buildProvider` short-circuit (the new test proves this — it sets `newProvider` to `t.Fatal` and still passes), and `MaxTurns: 0` is floored to 12 in `agent/loop.go`, so the onboarding session is functional, not bricked. ✔

Local verification (branch checked out): `go build ./...` clean, `go vet` clean on config/cli/specialist/tui, and the new `config` + `cli` tests pass.

### Nits (optional, non-blocking)
1. **Test helper left un-migrated.** `setCLIUserConfigRoot` (`internal/cli/app_test.go:76`) still uses `os.UserConfigDir()`, while the PR migrated the two sibling helpers (`setUserConfigRoot`, `setSpecialistConfigRoot`) to `config.UserConfigDir()`. On macOS it computes `root/Library/Application Support` instead of `root/.config`. It's harmless today (all three callers discard the return value, and HOME-redirect isolation still holds), but it's an inconsistency and a footgun the day a test captures that return — worth migrating for parity with the others.
2. **No migration for existing macOS installs.** Anyone who already ran v0.1.0 on macOS has `config.json` under `~/Library/Application Support/zero/`; after this change Zero reads `~/.config/zero/` and silently drops them into setup. Fine given the tool just launched and `~/.config` is the intended home — but a one-line **CHANGELOG** note (early macOS adopters should move `config.json`, or just re-run `zero setup`) would save confusion. (The PR has no CHANGELOG entry at all.)

Neither blocks merge. Nice, well-tested fix — thanks @smallnest for both the report and the PR.

gnanam1990 (commented):
Reviewed the diff and ran local checks on macOS (this is the exact platform the PR targets). Building on the existing reviews rather than repeating them.

## Local verification (darwin)
- `go build ./...` and `go vet` (config/cli/tui/specialist) clean.
- All new/touched tests pass in a clean env (`TestDefaultUserConfigPathUsesXDGConfigOnMacOS`, the `ErrNoActiveProvider` resolver + `errors.Is` assertion, and both `app.go` fallback tests).
- Confirmed the setup-fallback is nil-safe and doesn't conflict with the `buildProvider` provider-pinning already on main: the fallback test sets `newProvider` to `t.Fatal` and still passes, so `buildProvider` short-circuits on the empty provider before any provider work.

## Confirming @coderabbitai's darwin XDG finding (Major) — reproduced
The test-helper `XDG_CONFIG_HOME` leak is real, not theoretical. On a macOS box with `XDG_CONFIG_HOME` set, `internal/config` fails:

```
--- FAIL: TestDefaultResolveOptionsRejectsDirectoryConfigPaths/user_config
    paths_test.go:82: create directory /tmp/leaked-xdg/zero/config.json: not a directory
```

because `setUserConfigRoot`/`setSpecialistConfigRoot` only set `HOME` on darwin while `UserConfigDir()` now prefers `XDG_CONFIG_HOME`. Linux CI stays green (darwin branch not taken), which is why it slipped through. The proposed `t.Setenv("XDG_CONFIG_HOME", "")` fix is correct; worth folding in.

## One should-fix before merge: silent config-orphaning on macOS upgrade
@Vasanthdev2004 flagged "no migration" as a non-blocking nit — I think it deserves more weight. Verified on a real v0.1.0 macOS install:

| | file read | activeProvider | providers |
|---|---|---|---|
| v0.1.0 today | `~/Library/Application Support/zero/config.json` | ollama | 10 |
| after this PR | `~/.config/zero/config.json` | xai | 1 |

Existing macOS users silently switch to a different (or empty) config on upgrade — their providers/active selection appear to vanish, no error, no hint the old file is ignored. Nothing crashes, so it's not a correctness blocker, but it shouldn't ship silently. Suggest a one-time migration (if `~/.config/zero/config.json` is absent but the Application Support one exists, adopt it, ideally with a one-line notice), or at minimum a release note telling early adopters to move `config.json` / re-run setup.

## Verdict
Approvable — the fix is correct, well-tested, and closes the pre-existing config/OAuth split-brain. I'd gate merge on handling the macOS upgrade transition (migration in-PR or an explicit release note), and fold in the 2-line XDG test-leak fix. The remaining items (un-migrated `setCLIUserConfigRoot`, dup error-message helper, CI-skipped macOS test) are optional.

Thanks @smallnest for the fix and the clear root-cause writeup.

gnanam1990 (approved):
Approving — verified locally on macOS (build/vet/tests green, the setup-fallback is nil-safe and doesn't conflict with the provider-pinning on main). The fix is correct and closes the pre-existing config/OAuth split-brain.

Non-blocking follow-ups (details in my earlier review):
- Handle the macOS upgrade transition (one-time migration from `~/Library/Application Support/zero/config.json`, or a release note) so existing v0.1.0 users don't silently drop into setup.
- Fold in the 2-line `XDG_CONFIG_HOME` test-leak fix (reproduced on macOS).

Thanks @smallnest.

Vasanthdev2004:
LGTM @kevincodex1

PR #374, opened 2026-07-02
fix(gemini): strip unsupported JSON Schema fields from tool declarations

### LINKED ISSUE #373: Gemini provider: tool schema emits unsupported 'additionalProperties' field (12 tools rejected, all exec calls fail)
## Bug Description

Running any `zero exec` command against the Google/Gemini provider fails because Zero emits OpenAI-style JSON Schema tool declarations that Google's `function_declarations` API rejects.

## Reproduction

1. Install Zero 0.1.0 (`npm install -g`)
2. `zero providers setup google` (configure API key)
3. `zero providers use google`
4. `zero exec "say hi" --model gemini-2.5-pro`

## Error (exact)

All 12 tool declarations are rejected with:

```
Invalid JSON payload received. Unknown name "additionalProperties" at "tools[0].function_declarations[N].parameters": Cannot find field
```

This repeats for every tool in the payload. The exec call never reaches the model.

## Root Cause

Zero serializes tool schemas using OpenAI's JSON Schema format, which includes fields like `additionalProperties`. Google's Generative AI API (`GoogleGenAI v1alpha` OpenAPI spec) uses `function_declarations[].parameters` and only supports a subset of JSON Schema — `additionalProperties` is not a recognized field and causes a 400 rejection.

## Environment

- Zero 0.1.0 (npm install)
- WSL2 Ubuntu 24.04
- Model: gemini-2.5-pro
- Provider: Google (via `zero providers setup google`)

## Suggested Fix

Strip OpenAI-only JSON Schema fields (`additionalProperties`, `$schema`, `$id`, `patternProperties`, etc.) from tool parameter schemas when `transport=google` before sending to the Gemini API. Alternatively, map to Google's `parameters_schema` using only the supported subset (type, description, properties, required, items, enum, nullable).

gnanam1990:
## Problem

Fixes #373. Every `zero exec` against the Google/Gemini provider fails as soon as tools are advertised (i.e. essentially all agentic use). Zero serializes tool parameter schemas in OpenAI's JSON Schema dialect — `schemaToRuntimeMap` writes `additionalProperties` on every tool unconditionally — and Google's `functionDeclarations` schema doesn't recognize that keyword:

```
Invalid JSON payload received. Unknown name "additionalProperties"
at "tools[0].function_declarations[N].parameters": Cannot find field
```

The request is 400-rejected for every declaration, so the call never reaches the model. Reported against v0.1.0 with `gemini-2.5-pro`.

## Root cause

- `internal/agent/loop.go` — `schemaToRuntimeMap` always emits `"additionalProperties"` into a tool's runtime parameter map.
- `internal/providers/gemini/provider.go` — the mapper passed `tool.Parameters` **verbatim** into `functionDeclarations[].parameters`.

## Fix

Sanitize tool parameters **in the Gemini mapper only** — `additionalProperties: false` is valid and wanted for OpenAI strict mode and harmless for Anthropic, so this must not be stripped globally.

- `sanitizeGeminiSchema` keeps only the keywords Google's Schema type supports (`type`, `description`, `enum`, `properties`, `required`, `items`, `anyOf`, `default`, `nullable`, `format`, and the min/max/length constraints) via an **allowlist**, so any future OpenAI-ism or MCP-supplied field can't silently leak into a Gemini payload.
- Recurses through `properties`, `items`, and `anyOf`, so nested schemas are cleaned at every depth — not just the top level.
- `nil` parameters stay `nil` (parameterless tools unaffected).

## Tests

- Unit test on `sanitizeGeminiSchema`: nested stripping, legitimate fields (`type`/`description`/`enum`/`required`/`default`) preserved, `nil`-safe.
- End-to-end test asserting the real outgoing request body's `functionDeclarations[].parameters` carry no `additionalProperties` while the actual schema survives.
- Red-green: bypassing the sanitizer reproduces the exact reported payload and fails the E2E.

Full suite `go test ./...` passes (73/73 packages); change is isolated to the Gemini provider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gemini tool calls to send cleaner JSON schemas, reducing compatibility issues with unsupported schema fields.
  * Preserved supported parameter details while removing invalid extras from tool definitions.
* **Tests**
  * Added coverage for schema sanitization across nested fields and end-to-end Gemini request payloads.

### DIFF
--- a/internal/providers/gemini/provider.go
+++ b/internal/providers/gemini/provider.go
@@ -342,12 +342,77 @@ func (provider *Provider) geminiRequest(request zeroruntime.CompletionRequest) (
 			declarations = append(declarations, geminiFunctionDeclaration{
 				Name:        tool.Name,
 				Description: tool.Description,
-				Parameters:  tool.Parameters,
+				Parameters:  sanitizeGeminiSchema(tool.Parameters),
 			})
 		}
 		mapped.Tools = []geminiToolGroup{{FunctionDeclarations: declarations}}
 	}
 	return mapped, nil
+}
+
+// geminiSchemaFields is the subset of JSON Schema keywords Google's
+// functionDeclarations[].parameters accepts (the Generative AI Schema type).
+// Anything outside it — most notably OpenAI's `additionalProperties`, which Zero
+// emits on every tool's parameters and which Gemini 400s on ("Unknown name
+// \"additionalProperties\" … Cannot find field") — must be dropped before the
+// request goes out. Kept as an allowlist rather than a denylist so a schema
+// keyword Zero (or an MCP server) adds later can't silently leak an unsupported
+// field into the Gemini payload.
+var geminiSchemaFields = map[string]bool{
+	"type": true, "format": true, "title": true, "description": true,
+	"nullable": true, "enum": true, "items": true, "properties": true,
+	"required": true, "anyOf": true, "propertyOrdering": true, "default": true,
+	"minimum": true, "maximum": true, "minItems": true, "maxItems": true,
+	"minLength": true, "maxLength": true, "minProperties": true,
+	"maxProperties": true, "pattern": true, "example": true,
+}
+
+// sanitizeGeminiSchema returns a copy of a JSON-Schema map keeping only the
+// keywords Gemini supports (geminiSchemaFields), recursing through the nested
+// schemas under `properties`, `items`, and `anyOf`. Returns nil for a nil input
+// so a parameterless tool stays parameterless.
+func sanitizeGeminiSchema(schema map[string]any) map[string]any {
+	if schema == nil {
+		return nil
+	}
+	out := make(map[string]any, len(schema))
+	for key, value := range schema {
+		if !geminiSchemaFields[key] {
+			continue
+		}
+		switch key {
+		case "properties":
+			if props, ok := value.(map[string]any); ok {
+				cleaned := make(map[string]any, len(props))
+				for name, sub := range props {
+					if subMap, ok := sub.(map[string]any); ok {
+						cleaned[name] = sanitizeGeminiSchema(subMap)
+					} else {
+						cleaned[name] = sub
+					}
+				}
+				value = cleaned
+			}
+		case "items":
+			if subMap, ok := value.(map[string]any); ok {
+				value = sanitizeGeminiSchema(subMap)
+			}
+		case "anyOf":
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Verdict: Approve ✅

Fixes a launch-blocking bug — with tools advertised (i.e. essentially every agentic `zero exec`), Gemini 400-rejects every `functionDeclaration` on the OpenAI-ism `additionalProperties`. Checked out, built, vetted, and ran the tests on the branch.

### Correctness — solid
- **Right scope.** Sanitizing in the Gemini mapper only is the correct call — `additionalProperties: false` is wanted for OpenAI strict mode and harmless for Anthropic, so a global strip would be wrong. The base runtime schema is left intact for the other providers. ✔
- **Allowlist is complete and accurate.** I checked it against Google's Generative-Language `Schema` type: the 22 kept keywords match exactly (`type/format/title/description/nullable/default/items/minItems/maxItems/enum/properties/propertyOrdering/required/minProperties/maxProperties/minimum/maximum/minLength/maxLength/pattern/example/anyOf`). Allowlist (vs denylist) is the right choice — a future OpenAI-ism or MCP-supplied field can't silently leak. ✔
- **Recursion is right.** `properties` recurses into the sub-schema *values* (not the property names), `items`/`anyOf` handled, non-map values passed through, `nil` → `nil`. Confirmed the reported nested `additionalProperties` is stripped at depth by the test. ✔
- **No leak path.** `geminiRequest` is the single site that builds `functionDeclarations` (one completion path at `provider.go:107`), so there's no second unsanitized route. ✔

### Local verification (branch checked out)
`go build ./...` clean · `go vet ./internal/providers/gemini/...` clean · new `TestSanitizeGeminiSchemaStripsUnsupportedFields` + `TestGeminiRequestOmitsAdditionalPropertiesInToolSchema` pass · full `gemini` package green. The E2E test asserting the real outgoing body carries no `additionalProperties` while the schema survives is exactly the right regression guard.

### Nits (non-blocking)
- `items` in JSON Schema can be an *array* (tuple form); only the object form recurses. Immaterial here — Gemini doesn't support tuple `items` anyway, so an array would be equally unusable with or without this PR.
- No CHANGELOG entry.

Neither blocks merge. This one is worth prioritizing — Gemini is completely unusable for tool calls without it. Thanks @gnanam1990.

kevincodex1 (approved):
LGTM

PR #378, opened 2026-07-02
fix(tools): CRLF line ending mismatch in edit_file tool on Windows

### LINKED ISSUE #376: fix(tools): CRLF line ending mismatch in edit_file tool on Windows
### Description
On Windows, files often use CRLF (\\r\n\) line endings. The \
ead_file\ tool normalizes CRLF -> LF (\\n\) before presenting the contents to the agent, so the agent's generated \old_string\ uses LF.

However, the \edit_file\ tool matches \old_string\ byte-for-byte against raw file contents on disk. If the file contains CRLF, the match fails, resulting in a false-positive mismatch error.

### Solution
Normalized both file content and \old_string\ to \\n\ to compute occurrences and matches, then reconstructed the file with its original CRLF endings.

KunjShah95:
Closes #376

### Description
On Windows, files use CRLF (`\r\n`) line endings. The `read_file` tool normalizes CRLF → LF (`\n`) before presenting the contents to the agent, so the agent's generated `old_string` uses LF. 
However, the `edit_file` tool matches `old_string` byte-for-byte against raw file contents on disk. If the file contains CRLF, the match fails, resulting in a false-positive mismatch error.

### Solution
Normalized both file content and `old_string` to `\n` to compute occurrences and matches, then reconstructed the file with its original CRLF endings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved targeted text replacement for files using Windows-style line endings, ensuring matches work even when the search text uses standard line breaks.
  * Preserves the file’s existing line-ending format after the edit.
* **Tests**
  * Added a unit test covering string replacement in CRLF-terminated files.

### DIFF
--- a/internal/mcp/oauth_test.go
+++ b/internal/mcp/oauth_test.go
@@ -370,7 +370,14 @@ func TestFullFlowStoresTokens(t *testing.T) {
 		redirect := parsed.Query().Get("redirect_uri")
 		go func() {
 			cbURL := redirect + "?code=auth-code&state=" + url.QueryEscape(state)
-			_, _ = http.Get(cbURL)
+			for i := 0; i < 20; i++ {
+				resp, err := http.Get(cbURL)
+				if err == nil {
+					resp.Body.Close()
+					return
+				}
+				time.Sleep(50 * time.Millisecond)
+			}
 		}()
 		return nil
 	}
--- a/internal/sandbox/manager_test.go
+++ b/internal/sandbox/manager_test.go
@@ -72,7 +72,7 @@ func TestPermissionProfileFromPolicyIncludesDefaultTempWriteRoots(t *testing.T) 
 	if !writeRootsContain(profile.FileSystem.WriteRoots, tmpdir) {
 		t.Fatalf("write roots = %#v, want temp root %q", profile.FileSystem.WriteRoots, tmpdir)
 	}
-	if pathExists("/tmp") && !writeRootsContain(profile.FileSystem.WriteRoots, "/tmp") {
+	if runtime.GOOS != "windows" && pathExists("/tmp") && !writeRootsContain(profile.FileSystem.WriteRoots, "/tmp") {
 		t.Fatalf("write roots = %#v, want /tmp", profile.FileSystem.WriteRoots)
 	}
 }
--- a/internal/sandbox/scope_test.go
+++ b/internal/sandbox/scope_test.go
@@ -3,6 +3,7 @@ package sandbox
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 )
@@ -128,7 +129,7 @@ func TestNewScopeNormalizesAndValidatesExtraRoots(t *testing.T) {
 	if !stringSliceContains(roots, normalizeWorkspaceRootBestEffort(extra)) {
 		t.Fatalf("Roots()=%v want extra root %q", roots, normalizeWorkspaceRootBestEffort(extra))
 	}
-	if pathExists("/tmp") && !stringSliceContains(roots, normalizeWorkspaceRootBestEffort("/tmp")) {
+	if runtime.GOOS != "windows" && pathExists("/tmp") && !stringSliceContains(roots, normalizeWorkspaceRootBestEffort("/tmp")) {
 		t.Fatalf("Roots()=%v want default /tmp write root", roots)
 	}
 }
--- a/internal/tools/edit_file.go
+++ b/internal/tools/edit_file.go
@@ -79,6 +79,23 @@ func (tool editFileTool) RunWithOptions(_ context.Context, args map[string]any, 
 	content := string(contentBytes)
 	occurrences := strings.Count(content, oldString)
 
+	// CRLF fallback: read_file normalizes \r\n → \n before presenting content to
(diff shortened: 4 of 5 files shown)


Vasanthdev2004 (commented):
## Verdict: Useful & correct — but one required fix before merge (gofmt fails CI)

**Is it useful?** Yes. I confirmed the root cause is real: `read_file.go:82` normalizes `\r\n → \n` before the model sees the file, while `edit_file.go:80` matches `old_string` byte-for-byte against the *raw* bytes. So on a CRLF file (any Windows checkout with `autocrlf`), the model's LF `old_string` genuinely fails with a false "could not find the exact string" error. Since Zero targets Windows and the reporter hit exactly this, it's a meaningful fix.

**Logic is correct:**
- The CRLF fallback runs only when the direct match fails *and* the file is CRLF *and* `old_string` is LF — so it never perturbs the normal path.
- It slots in before the existing uniqueness / `replace_all` checks, so those still apply to the translated string (verified by reading `edit_file.go:85-93`).
- `new_string` is translated to `\r\n` too, so the edited region keeps the file's EOL style; untouched regions are left alone. Test asserts the output stays `\r\n`. ✔

### 🔴 Required before merge
`gofmt -l internal/tools/write_tools_test.go` flags the file — there's a **double blank line** between `TestEditFileToolReplacesCRLF` and `TestEditFileToolEmitsUnifiedDiff`. CI's format check will fail. Drop one blank line and run `gofmt -w`.

### Nits (optional)
- If `new_string` itself already contained `\r\n`, the `\n → \r\n` translation would produce `\r\r\n`. Can't happen today (the model works from LF-normalized reads), so it's theoretical — but guarding `!strings.Contains(newString, "\r\n")` before translating would make it robust.
- Mixed-EOL files (some `\r\n`, some `\n`) where `old_string` spans the boundary won't match — the heuristic targets the common all-CRLF case, which is the reported bug. Fine to leave.

Fix the gofmt line and this is good to merge. Want me to push the one-line whitespace fix if maintainer edits are enabled on the PR? Thanks @KunjShah95.

gnanam1990:
@KunjShah95 fix CI fails

KunjShah95:
I have updated the PR with the following changes:

1. **Fixed gofmt CI Failure**: Removed the double blank line in `internal/tools/write_tools_test.go` that was causing `gofmt -l` to fail.
2. **Addressed Code Nit**: Guarded the CRLF replacement in `internal/tools/edit_file.go` with `!strings.Contains(newString, "\r\n")` to avoid double translation if the string already contains CRLF line endings.
3. **Fixed Windows Sandbox Test Failures**: Fixed the `/tmp` write root assertion in `internal/sandbox/manager_test.go` and `internal/sandbox/scope_test.go` to only run on non-Windows platforms. On Windows, `/tmp` (which resolves to `C:\tmp`) is not included in the default temp write roots, but the test was checking for it simply if `C:\tmp` existed on the host filesystem.
4. **Resolved Flaky OAuth Test**: Added a retry loop in `TestFullFlowStoresTokens` inside `internal/mcp/oauth_test.go` to handle slow TCP listener startup, resolving a race condition that sometimes caused OAuth tests to timeout with a connection failure.

Vasanthdev2004 (approved):
## Verdict: Approve ✅ — follow-up commit resolves the blocker and then some

Re-reviewed at `591845e`. Everything from my earlier comment is addressed, and the PR grew to fix a few real Windows/CI failures too. Verified on a Windows box (checked out the branch).

### The CRLF fix — still correct, now more robust
- `gofmt` is clean (the stray double blank line is gone). ✔
- The `new_string` translation is now guarded (`if !strings.Contains(newString, "\r\n")`) before the `\n → \r\n` rewrite, closing the `\r\r\n` double-up I flagged. ✔
- Core logic unchanged and still right: fallback only fires on a failed direct match against a CRLF file with an LF `old_string`, preserves EOL style, and rides the existing uniqueness / `replace_all` checks.

### Bonus fixes (welcome, and they make *this PR's* Windows CI green)
- `sandbox/scope_test.go` + `manager_test.go`: the `/tmp` write-root assertions are now gated on `runtime.GOOS != "windows"`. This fixes `TestNewScopeNormalizesAndValidatesExtraRoots` / `TestPermissionProfileFromPolicyIncludesDefaultTempWriteRoots`, which fail on Windows because the default temp root there is `%TEMP%`, not `/tmp` — a pre-existing failure. Confirmed both now pass on my Windows box (they failed on `main`).
- `mcp/oauth_test.go`: the callback GET is now a bounded retry loop (20 × 50ms) that also closes `resp.Body` — hardens a startup race and fixes a small leak.

### Verification (branch checked out)
`go build ./...` clean · `gofmt` clean on all 5 files · full `tools` + `sandbox` + `mcp` packages pass.

### One note (non-blocking)
Scope grew past the title — it's now CRLF **plus** two sandbox Windows-test guards and an MCP OAuth flake fix. Ideally those would be separate PRs, but each is a small, correct test-robustness change, and they're arguably what let a Windows-focused fix land with green cross-platform CI. Fine to keep bundled here; just worth a mention in case you'd rather split.

Approving. Thanks @KunjShah95 for the quick, thorough turnaround.

KunjShah95:
thank you for trusting me @Vasanthdev2004

kevincodex1 (approved):
LGTM

PR #379, opened 2026-07-02
security(config): project config should not override sandbox.network to allow

### LINKED ISSUE #377: security(config): project config should not override sandbox.network to allow
### Description
A repository's project-level `.zero/config.json` config can set `sandbox.network` to `"allow"`. This violates the sandbox boundary because a cloned repo should not be able to weaken sandbox network restrictions for shell commands. This is a security supply-chain risk.

### Solution
Prevented the project config merge from overriding `sandbox.network` to `"allow"`. Project config can only tighten network restrictions (e.g. to `"deny"`).

KunjShah95:
Closes #377

### Description
A repository's project-level `.zero/config.json` config can set `sandbox.network` to `"allow"`. This violates the sandbox boundary because a cloned repo should not be able to weaken sandbox network restrictions for shell commands. This is a security supply-chain risk.

### Solution
Prevented the project config merge from overriding `sandbox.network` to `"allow"`. Project config can only tighten network restrictions (e.g. to `"deny"`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Project-level sandbox network settings can no longer weaken security by switching the network mode to a less restrictive option (for example, a project “allow” won’t override when not tightening).
  * When a project config specifies a stricter sandbox network mode, it is now correctly applied during configuration resolution.
  * Added tests to verify correct sandbox network precedence between user and project configuration layers.

### DIFF
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -225,7 +225,8 @@ func mergeProjectConfig(dst *FileConfig, src FileConfig) error {
 	// WEAKEN (→ "allow"). A malicious repo must not be able to open network
 	// access for shell commands. Matches the AdditionalWriteRoots posture.
 	if network := strings.TrimSpace(src.Sandbox.Network); network != "" {
-		if sandbox.NetworkMode(network) == sandbox.NetworkDeny {
+		// Normalize: anything that isn't literally "allow" counts as deny.
+		if sandbox.NormalizeNetworkMode(sandbox.NetworkMode(network)) != sandbox.NetworkAllow {
 			dst.Sandbox.Network = network
 		}
 		// Silently ignore "allow" from project config — not an error, just
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -220,8 +220,16 @@ func mergeProjectConfig(dst *FileConfig, src FileConfig) error {
 	// config: a cloned repo's .zero/config.json must not be able to grant
 	// itself write access outside the workspace. Global config and CLI flags
 	// are the only config sources for write roots.
+	//
+	// Sandbox.Network from project config may only TIGHTEN (→ "deny"), never
+	// WEAKEN (→ "allow"). A malicious repo must not be able to open network
+	// access for shell commands. Matches the AdditionalWriteRoots posture.
 	if network := strings.TrimSpace(src.Sandbox.Network); network != "" {
-		dst.Sandbox.Network = network
+		if sandbox.NetworkMode(network) == sandbox.NetworkDeny {
+			dst.Sandbox.Network = network
+		}
+		// Silently ignore "allow" from project config — not an error, just
+		// a privilege the project scope does not have.
 	}
 	if src.Sandbox.BlockUnixSockets {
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Verdict: Approve ✅ — useful, correct, well-scoped

**Is it useful?** Yes — it closes a real supply-chain gap. Today `mergeProjectConfig` (`resolver.go:223`) does `dst.Sandbox.Network = network` unconditionally, so a cloned repo's `.zero/config.json` with `sandbox.network: "allow"` can *weaken* the shell sandbox's network egress. That's the same threat the existing code already defends against one field up (`AdditionalWriteRoots` is deliberately not merged from project config — "a cloned repo must not grant itself write access outside the workspace"). Network egress is the same class of privilege, so this makes the posture consistent.

**Correct & well-scoped:**
- `engine.go:167` confirms the policy is binary (`NetworkDeny` / `NetworkAllow`), so "project may only tighten → deny" is the right rule. The fix permits `"deny"` and drops `"allow"`. ✔
- I checked the adjacent surface: `mergeProjectConfig` never merges `Sandbox.Mode` (so a project can't disable the sandbox outright) and only *tightens* `BlockUnixSockets` / `MonitorDenials`. Network was the one asymmetric hole — this closes it without over-reaching. ✔
- gofmt clean; the new `TestResolveSandboxNetworkProjectConfigCannotWeaken` covers both directions (project "allow" ignored; project "deny" tightens over a user "allow") and passes locally along with the full `config` package.

### Nit (optional, non-blocking)
The guard uses a raw cast — `sandbox.NetworkMode(network) == sandbox.NetworkDeny` — so it only matches the exact lowercase `"deny"`. A project that writes `"DENY"` / `"Deny"` would silently *fail to tighten* (it stays fail-safe — never weakens — but its tightening intent is lost). Since `NormalizeNetworkMode` treats everything except `"allow"` as deny, a more robust and intent-matching form is:

```go
if sandbox.NormalizeNetworkMode(sandbox.NetworkMode(network)) != sandbox.NetworkAllow {
    dst.Sandbox.Network = network
}
```

That accepts any non-`allow` value as a tighten and blocks only `allow`. Not required — the current version is safe — but it's a small robustness win and mirrors how the rest of the sandbox package reads these values.

Approving as-is. Thanks @KunjShah95.

gnanam1990 (approved):
Approving. Verified the fix sits at the correct trust boundary — `mergeProjectConfig` only, so a cloned repo's `.zero/config.json` can tighten network (→`deny`) but never weaken it (→`allow`), while user config/CLI stay authoritative. With only two modes (`deny`/`allow`) and a `deny` default, this fully closes the supply-chain hole. Good test covering both the ignore-weaken and allow-tighten paths.

Non-blocking nit (as @coderabbitai noted): the compare is case-sensitive, so a project config value like `"Deny"`/`"DENY"` silently fails to tighten. Routing through `sandbox.NormalizeNetworkMode` would make intent-to-restrict robust — doesn't affect the security guarantee (blocking `allow`), so fine as a follow-up.

Thanks @KunjShah95.

kevincodex1 (approved):
looks good!

PR #382, opened 2026-07-02
fix(tools): require permission before web_search requests

### LINKED ISSUE #380: web_search should require network permission before sending queries
### Version / branch / commit

branch: main commit: bd89a1f

### OS and environment

### Steps to reproduce

1. Configure a hosted search backend with `ZERO_WEBSEARCH_BASE_URL`.
2. Start Zero with the built-in network tools enabled.
3. Ask the model to search the web so it calls `web_search`.
4. Notice that `web_search` can run without a user approval prompt.

Code path:
- `web_search` is marked as `SideEffectNetwork`
- but its safety metadata currently uses `PermissionAllow`

### Expected behavior

`web_search` should require user approval before sending the query to the configured search backend.

It should still be visible to the model in auto mode, similar to `web_fetch`, by using:

- `PermissionPrompt`
- `AdvertiseInAuto: true`

Existing allow/session/always permission decisions can remain the compatibility path.

### Actual behavior

`web_search` is a network-side-effecting tool, but it is currently allowed without prompting.

That means a configured hosted search backend can receive model-generated query text before the user approves network access.

_No response_

GautamBytes:
Fixes #380

## Summary

- Gate the built-in `web_search` tool behind the normal permission flow before it sends queries to a configured backend.
- Keep `web_search` visible in auto mode with `AdvertiseInAuto: true`, matching the existing `web_fetch` pattern.
- Add coverage for registry, agent-loop, and sandbox behavior so denied permission prevents execution while approved calls still work.

## Why

`web_search` is marked as a network side-effecting tool, but it previously used `PermissionAllow`. That let hosted search run without a permission prompt even though queries may include prompt or workspace context.

### DIFF
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -37,6 +37,35 @@ func (provider *mockProvider) StreamCompletion(ctx context.Context, request zero
 	return ch, nil
 }
 
+type recordingWebSearchTool struct {
+	calls []map[string]any
+}
+
+func (tool *recordingWebSearchTool) Name() string        { return "web_search" }
+func (tool *recordingWebSearchTool) Description() string { return "test web search tool" }
+func (tool *recordingWebSearchTool) Parameters() tools.Schema {
+	return tools.Schema{
+		Type: "object",
+		Properties: map[string]tools.PropertySchema{
+			"query": {Type: "string"},
+		},
+		Required:             []string{"query"},
+		AdditionalProperties: false,
+	}
+}
+func (tool *recordingWebSearchTool) Safety() tools.Safety {
+	return tools.Safety{
+		SideEffect:      tools.SideEffectNetwork,
+		Permission:      tools.PermissionPrompt,
+		Reason:          "Performs a web search over the network.",
+		AdvertiseInAuto: true,
+	}
+}
+func (tool *recordingWebSearchTool) Run(_ context.Context, args map[string]any) tools.Result {
+	tool.calls = append(tool.calls, cloneArgs(args))
+	return tools.Result{Status: tools.StatusOK, Output: "1. T — https://x.test"}
+}
+
 type sandboxDeniedRetryTool struct {
 	calls []map[string]any
 }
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
## Verdict: Approve ✅ — correct, and it makes `web_search` consistent with `web_fetch`

**Is it useful?** Yes. `web_search` was the one first-party network tool that ran *without* a permission prompt (`PermissionAllow`), even though it ships model-generated query text — which can embed prompt/workspace context — to a hosted backend. Its sibling `web_fetch` already prompts for exactly this reason. I confirmed `web_fetch`'s metadata is `SideEffectNetwork + PermissionPrompt + AdvertiseInAuto: true` (`web_fetch.go:128-131`) — this PR makes `web_search` byte-for-byte identical in posture. Closing that asymmetry is the right call.

**Implementation is correct:**
- `AdvertiseInAuto: true` is required and right — advertisement in auto mode is gated on `Permission == PermissionAllow || AdvertiseInAuto` (`loop.go:2620/2628`), so without it a now-`PermissionPrompt` tool would vanish from the model's toolset in auto mode. With it, `web_search` stays visible in auto and prompts only where the mode prompts — same as `web_fetch`. ✔
- The sandbox exemption is preserved correctly: `engine_test.go` now asserts `web_search` yields `ActionPrompt` (from tool permission) rather than `ReasonNetworkBlocked` (shell egress), and `ActionAllow` once granted. So first-party network tools are still exempt from the *shell* egress policy and governed by their permission metadata — the intended design. ✔
- No stragglers: the web_search-specific `PermissionAllow` assertions (registry/web_search/engine/loop tests) are all updated; I grepped the tree and found no missed reference.

**Behavioral impact (worth the maintainer noting):** this is a deliberate reversal of the old "read-only discovery action, allow without prompt" choice. Net effect — **Ask mode** now prompts before a search (previously frictionless); **Auto mode** is unchanged (still runs). Session/always grants remain the compatibility path, so it's one approval per session, not per query. Given it simply brings `web_search` in line with `web_fetch`, the trade-off is reasonable — flagging only so it's a conscious product decision, not a surprise.

**Verification (branch checked out):** `go build ./...` clean, `gofmt` clean on all five changed files, and the full `tools` + `agent` packages plus every new test pass. (The one `sandbox` failure — `TestNewScopeNormalizesAndValidatesExtraRoots`, "want default /tmp write root" — is a pre-existing Windows-only failure that reproduces identically on `main`; unrelated to this change.)

Approving. Thanks @GautamBytes — clean, well-tested, and the red-green permission test is exactly the right guard.

kevincodex1 (approved):
nice catch!

PR #386, opened 2026-07-02
feat(providers): add `zero providers models` to discover a provider's models

### LINKED ISSUE #381: OpenAI-compatible model provider self-discovery
### Problem

OS model providers like Opencode Go are serving around a dozen of models and provide model-discovery endpoints. Supporting such form of configuration could reduce the necessity of manual model configuration for each model (TUI is slow, config file requires mostly copy-pasting around a dozen of config objects per each model). Both error-prone.

### Requested behavior

Add support for automatic model discovery and automated config creation in batches. This could be used for further updates, to prevent necessity of manual reconfiguration. The model fetch is already implemented, but doesn't seem to have any effect for custom (OpenAI-compatible) type of configurations.

### Why it belongs in Zero

Zero brings speed and reliability into domain mostly taken by less performant tools. Faster onboarding shall streamline the migration process and increase adoption.

Configure each model manually over TUI or config file.

_No response_

Vasanthdev2004:
Fixes #381.

## Problem

`zero providers` exposes `list / add / check / use / setup / detect` — but **no way to see the models a provider actually serves**. For a custom OpenAI-/Anthropic-compatible endpoint (e.g. a self-hosted proxy serving a dozen models), that meant hand-writing a config object per model — slow and error-prone, exactly the pain #381 describes.

Live model discovery already exists (`providermodeldiscovery`), but it's only wired into the **TUI model picker**. Headless / config-first users had no surface for it, so on the CLI "the model fetch doesn't seem to have any effect for custom configs."

## Change

Add **`zero providers models [name] [--json]`**. It probes the provider's live model-listing endpoint (an OpenAI-compatible `/v1/models` or the Anthropic `/models`) and prints what it serves.

- Uses `providermodeldiscovery.Discover` — the **raw** primitive, which runs purely off the profile's base URL + resolved credential (inline → stored → env, same resolution as the runtime). No curated-catalog merge and **no coding-model filtering**, so a custom provider's *full* model list comes back.
- Because Zero already passes unknown model ids through to the provider, the workflow becomes: configure the provider **once**, run `zero providers models` to list what's available, then use any id with `zero exec --model <id>` — **no per-model config**.
- Wired via a new `appDeps.discoverProviderModels` hook (default `providermodeldiscovery.Discover`) so the command is unit-tested hermetically, mirroring the existing `probeProviderHealth` / `detectLocalRuntimes` deps.

### Example

```
$ zero providers models self-hosted
Provider models (self-hosted)
team/model-a — First model
team/model-b
2 models discovered
next: zero exec "hello" --model team/model-a
```

`--json` emits `{ "provider", "count", "models": [{ "id", "description" }] }` for scripting.

## Scope note

This deliberately does **not** generate a config object per model — that would fight Zero's one-profile-plus-live-model-switching design. It gives the headless surface that was missing (and that the TUI picker already had), which is the smaller, higher-value half of #381. A re-runnable "write discovered models to config" command could be a follow-up if there's demand.

## Testing

- `internal/cli/provider_models_test.go`: lists discovered models (active provider, text), `--json` shape (incl. description omitted when empty), selecting a named provider, unknown-provider usage error (discovery not called), and discovery error surfaced as `exitProvider`.
- `go build ./...`, `go vet ./internal/cli/...`, `gofmt` clean; full `internal/cli` package passes.

## Summary by CodeRabbit

* **New Features**
  * Added a new `zero providers models` command to view available models for a provider.
  * Supports both readable output and `--json` export.
  * Shows a helpful example command using the first available model when results are found.

* **Bug Fixes**
  * Updated provider command help and routing so `models` is recognized correctly instead of being treated as an invalid subcommand.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -22,6 +22,7 @@ import (
 	"github.com/Gitlawb/zero/internal/observability"
 	"github.com/Gitlawb/zero/internal/plugins"
 	"github.com/Gitlawb/zero/internal/providerhealth"
+	"github.com/Gitlawb/zero/internal/providermodeldiscovery"
 	"github.com/Gitlawb/zero/internal/provideronboarding"
 	"github.com/Gitlawb/zero/internal/providers"
 	"github.com/Gitlawb/zero/internal/redaction"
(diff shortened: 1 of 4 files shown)


anandh8x (approved):
Clean addition. Matches the existing `probeProviderHealth` / `detectLocalRuntimes` dep-wiring pattern, tests cover the right cases.

One thing worth a follow-up: `discoveryCredentialProfile` duplicates the credential-resolution chain from `discoveredModelContextWindow`. Should be extracted into a shared helper before it drifts — but non-blocking for this PR.

LGTM.

kevincodex1 (approved):
LGTM

PR #389, opened 2026-07-02
fix(provider-wizard): allow multiple custom OpenAI-compatible providers

### LINKED ISSUE #375: Not able to setup multiple openai comp providers
### Version / branch / commit

zero 0.1.0

### OS and environment

zorin os

### Steps to reproduce

Run Zero, set custom openai comp provider and after that we are not able to set another provider  

### Expected behavior

There should be an option to setup multiple provider 

### Actual behavior

After seting custom openai comp provider and after that we are not able to set another provider  

_No response_

KunjShah95:
Fixes #375

**Problem:** \wizardProviderStoredKey\ matched saved providers by CatalogID. Since ALL custom OpenAI-compatible providers share CatalogID \custom-openai-compatible\, setting up one blocked configuring a second.

**Fix:**
1. Skip CatalogID matching for generic/custom providers (they all share the same ID)
2. When replacing a key for a generic provider, route through endpoint/name steps so the user enters a new base URL and name

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved saved provider matching to better handle generic/custom OpenAI-compatible provider IDs, reducing incorrect keep/replace/remove prompts.
  * Updated the “Replace” flow to clear stale connection details and route to the correct next step (endpoint when required; otherwise credentials).
* **Tests**
  * Increased OAuth callback polling/completion timeouts.
  * Added test coverage for generic OpenAI-compatible provider replace routing behavior.

### DIFF
--- a/internal/cli/mcp_oauth_test.go
+++ b/internal/cli/mcp_oauth_test.go
@@ -144,7 +144,7 @@ func TestRunMCPOAuthLoginStoresTokens(t *testing.T) {
 	stdout := &syncBuffer{}
 	stderr := &syncBuffer{}
 	go func() {
-		deadline := time.Now().Add(3 * time.Second)
+		deadline := time.Now().Add(15 * time.Second)
 		for time.Now().Before(deadline) {
 			callbackURL := extractCallbackURL(stdout.String())
 			if callbackURL != "" {
@@ -164,8 +164,8 @@ func TestRunMCPOAuthLoginStoresTokens(t *testing.T) {
 	var exitCode int
 	select {
 	case exitCode = <-done:
-	case <-time.After(10 * time.Second):
-		t.Fatalf("login did not complete within 10s; stderr=%s stdout=%s", stderr.String(), stdout.String())
+	case <-time.After(20 * time.Second):
+		t.Fatalf("login did not complete within 20s; stderr=%s stdout=%s", stderr.String(), stdout.String())
 	}
 	if exitCode != exitSuccess {
 		t.Fatalf("exitCode = %d stderr=%s stdout=%s", exitCode, stderr.String(), stdout.String())
--- a/internal/tui/provider_wizard.go
+++ b/internal/tui/provider_wizard.go
@@ -942,7 +942,8 @@ func (m model) wizardProviderStoredKey(provider providercatalog.Descriptor) (str
 			continue
 		}
 		if strings.EqualFold(strings.TrimSpace(profile.Name), strings.TrimSpace(provider.Name)) ||
-			strings.EqualFold(strings.TrimSpace(profile.CatalogID), strings.TrimSpace(provider.ID)) ||
+			(!genericProviderCatalogID(provider.ID) &&
+				strings.EqualFold(strings.TrimSpace(profile.CatalogID), strings.TrimSpace(provider.ID))) ||
 			strings.EqualFold(strings.TrimSpace(profile.Name), strings.TrimSpace(provider.ID)) {
 			return profile.Name, true
 		}
@@ -963,7 +964,13 @@ func (m model) applyManageKeyChoice() (model, tea.Cmd) {
 	case 1: // Replace
 		wizard.apiKey = ""
 		wizard.err = ""
-		wizard.step = providerWizardStepCredential
+		wizard.baseURL = ""
+		wizard.profileName = ""
+		if providerWizardNeedsEndpoint(wizard.currentProvider()) {
+			wizard.step = providerWizardStepEndpoint
+		} else {
+			wizard.step = providerWizardStepCredential
+		}
 		return m, nil
 	case 2: // Remove
 		if strings.TrimSpace(m.userConfigPath) != "" {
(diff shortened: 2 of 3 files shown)


Vasanthdev2004:
fix coderabbit issues @KunjShah95

KunjShah95:
@Vasanthdev2004

KunjShah95:
do check it out now @Vasanthdev2004

Vasanthdev2004:
yea wait a min @KunjShah95

Vasanthdev2004 (commented):
## Verdict: Correct fix — but #403 is the stronger duplicate of this exact change; 

I checked this out, built it, ran the wizard suite (green), and reproduced the fix directly: with a saved `custom-openai-compatible` provider, re-selecting the generic "Custom OpenAI-compatible" catalog entry now returns `matched=false` from `wizardProviderStoredKey`, so the wizard proceeds to a fresh endpoint/name/key flow instead of getting trapped in keep/replace/remove. **It does fix #375.** CI is green.

But this PR (#389) and **#403** fix the *same bug the same way*, and #403 is the cleaner implementation. Since you can only merge one, here's the honest comparison.

### Where the two differ
- **Locus of the fix.** #389 mutates the low-level lookup `wizardProviderStoredKey` (adds a `!genericProviderCatalogID(...)` guard around the CatalogID clause). #403 leaves the lookup's semantics intact and gates at the *call site* in `advanceProviderWizard` — "if this provider needs an endpoint, skip ManageKey and fall through to add." Gating at the decision point is the better locus: the lookup keeps meaning "is there a stored key for this descriptor," and only the manage-vs-add *decision* changes.
- **Test coverage.** #403 adds `TestAdvanceProviderWizardCustomSkipsManageKey` (table over **both** `custom-openai-compatible` and `custom-anthropic-compatible`, asserting the real `advanceProviderWizard` path routes to the endpoint step, not ManageKey) **plus** `TestAdvanceProviderWizardNamedShowsManageKey` (a named provider still reaches ManageKey — guards against an over-broad skip). #389's `TestProviderWizardManageKeyReplaceGeneric` instead exercises a **manually-constructed** ManageKey state that the real flow can no longer reach for a generic provider (see below), so it doesn't actually cover the add-a-second-provider scenario that #375 is about.
- **Scope.** #389 also bundles unrelated `mcp_oauth_test.go` timeout bumps (3s→15s, 10s→20s). Harmless flake-loosening, but it doesn't belong in this fix.

### One thing both PRs should acknowledge (shared limitation, not a blocker)
The wizard's provider list comes from `providercatalog.All()`, so custom providers only ever appear as the single generic "Custom OpenAI-compatible" entry — never as your saved instance name. `wizardProviderStoredKey` (via `advanceProviderWizard`) is the *only* path into keep/replace/remove. Once we skip the CatalogID match for generics, **there's no longer any way to reach replace/remove for an existing custom provider through the wizard** (the generic descriptor's Name is "Custom OpenAI-compatible", which won't match a user's saved name). That's an acceptable trade to unblock multi-provider setup — the old behavior could only ever manage the *first* CatalogID match anyway — but it means removing a stale custom key now has no wizard surface. Worth a short follow-up (e.g. list saved custom instances for management), and worth a one-line note in the PR so it's a conscious choice.

As a direct consequence, #389's `applyManageKeyChoice` Replace change (clear `baseURL`/`profileName`, route to endpoint for generics) is now effectively unreachable in the real flow — generic providers can't reach ManageKey anymore. It's harmless/defensive, but it's dead relative to the actual bug.

### Recommendation
Both are correct. I'd **consolidate on #403** (cleaner locus, tests the real flow for both custom types + a named-provider guard, no unrelated changes) and close this one — or, if you prefer this author, ask them to (1) drop the mcp_oauth timeout hunk, and (2) replace the ManageKey-replace test with a table test that drives `advanceProviderWizard` for both custom transports like #403 does. Either way, please capture the "can't manage an existing custom key via the wizard" follow-up.

Thanks @KunjShah95 — the diagnosis is right and the fix works; this is purely a "which of two identical fixes to keep" call.

KunjShah95:
ohkay no problem bhai

Vasanthdev2004:
Closing in favor of #403, which fixes the same issue (#375) with a cleaner approach — it gates at the call site in `advanceProviderWizard` instead of mutating the `wizardProviderStoredKey` lookup, tests the real advance flow for both `custom-openai-compatible` and `custom-anthropic-compatible`, adds a named-provider guard test, and doesn't bundle the unrelated `mcp_oauth_test` timeout changes.

Thanks a lot @KunjShah95 — your diagnosis of #375 was spot on and your fix worked; this is purely a "two identical fixes, keep the cleaner one" call, not a knock on the work. Full comparison is in my review above. Appreciate the contribution, and look forward to the next one.

gnanam1990 (changes requested):
Thanks for this — the core fix (hunk 1: `wizardProviderStoredKey` no longer matching a generic custom descriptor by CatalogID) correctly resolves #375, and the requested Replace→endpoint test is now in.

I'm going to go with #403 for this issue, though, so I'd suggest closing this in its favor — two reasons:

- **Unrelated scope creep:** this also bumps the MCP-OAuth login test timeouts (`mcp_oauth_test.go`) in a different subsystem — the commit message even calls it "fix flaky MCP OAuth test." That doesn't belong in a provider-wizard change.
- **Redundant second hunk:** the `applyManageKeyChoice` Replace hunk repairs a path that hunk 1 makes unreachable for custom providers, and the added test targets that dead path rather than the actual #375 repro.

#403 does the same fix with one localized change plus tests that hit the real repro and the regression boundary, no scope creep. No hard feelings — your root-cause diagnosis was correct. Marking request-changes to reflect that this shouldn't merge as-is.

PR #395, opened 2026-07-02
fix(tui): title /model rows by model name, not the catalog description

Vasanthdev2004:
## Problem

The `/model` picker rendered rows with sentence-long titles and what looked like exact duplicates, hiding the real model id:

```
Ollama Cloud
  DeepSeek chat model for instruction following, coding, and analysis
> DeepSeek chat model for instruction following, coding, and analysis
  Fast DeepSeek model for efficient chat, coding help, and agent loops
  Mistral coding agent model for repository tasks and software engineer…
  Mistral coding agent model for repository tasks and software engineer…
```

## Root cause

`modelPickerDisplayName(id, description)` returned the **description** as the row title whenever it was non-empty and not a known-generic placeholder. Catalog/discovery descriptions come from **models.dev** as full sentences, and several models share the same sentence — so:

- titles became full sentences instead of `deepseek-v3.2` / `glm-5.2`,
- models sharing a description (e.g. `deepseek-v3.2` and `deepseek-v3.2-thinking`) looked like **duplicate rows**,
- the real id survived only in the meta line (`deepseek-v3.2 · 163K ctx · tools · reasoning`).

The `providerWizardGenericModelDescription` filter only catches placeholders and strings ending in `" model"`, so a sentence ending in `"analysis"` slipped through.

## Fix

Prefer the (prettified) model **id** as the row title; fall back to a non-generic description only when there is no id to name the row. Distinct ids that share a catalog description now render as distinct rows. No data is lost — the ctx/tools/reasoning spec is already in the row's meta line.

```
Ollama Cloud
  Deepseek V3.2
> Deepseek V3.2 Thinking
  Glm 4.6 Flash
  Codestral 2
  Devstral 2
```

## Tests

`internal/tui/model_picker_display_test.go`:
- titles come from the id, not a sentence description (incl. the exact Ollama Cloud scenario, provider-prefixed ids, and `:tag` ids);
- two distinct ids sharing one description render as distinct, non-sentence titles;
- with no id, a non-generic description is still an acceptable fallback (else `"model"`).

`go build ./...`, `go vet ./internal/tui/...`, `gofmt` clean; full `internal/tui` package passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Model picker entries now show clearer, more consistent names based on the model ID when available.
  * Fixed cases where different models could appear under the same shared description.
  * Improved fallback behavior so empty or generic entries display a sensible default label.

* **Tests**
  * Added regression coverage for model picker display-name behavior.

### DIFF
--- a/internal/tui/model_picker_display_test.go
+++ b/internal/tui/model_picker_display_test.go
@@ -1,0 +1,57 @@
+package tui
+
+import "testing"
+
+// The model picker must title each row with the model NAME, not the models.dev
+// marketing description. Regression: Ollama Cloud models resolved full-sentence
+// descriptions from the remote catalog, so rows showed sentences and several
+// models sharing one description looked like exact duplicates (issue: /model
+// list rendered descriptions as titles).
+func TestModelPickerDisplayNamePrefersModelName(t *testing.T) {
+	sentence := "DeepSeek chat model for instruction following, coding, and analysis"
+	cases := []struct {
+		id          string
+		description string
+		want        string
+	}{
+		{"deepseek-v3.2", sentence, "Deepseek V3.2"},
+		{"deepseek-v3.2-thinking", sentence, "Deepseek V3.2 Thinking"},
+		{"glm-5.2", "GLM flagship model", "GLM 5.2"},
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
nice catch

PR #396, opened 2026-07-02
fix(mcp): block cross-origin credential redirects

GautamBytes:
## Summary

- Hardens remote MCP HTTP/SSE traffic so auth-bearing requests stay on the configured server origin.
- Rejects cross-origin redirects and cross-origin SSE endpoint events before secrets can be forwarded.
- Keeps same-origin redirects/endpoints working, including default-port normalization.

Security report was submitted privately first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened HTTP redirect safeguards to block cross-origin hops and limit redirect chains, preventing header leakage.
  * Hardened SSE endpoint and URL resolution to require the endpoint stays on the same configured origin.
  * Improved OAuth redirect handling to ensure bearer authorization is never sent to cross-origin redirect targets.
* **Tests**
  * Added coverage for same-origin enforcement, default-port normalization, and redirect/SSE/OAuth rejection cases.

### DIFF
--- a/internal/mcp/client_test.go
+++ b/internal/mcp/client_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (approved):
## Verdict: Approve ✅ — correct, complete for its scope, and it doesn't break same-origin flows

Reviewed this as a security change (checked out the branch, audited the whole client surface, ran the suite). The hardening is sound and the design choices are the right ones.

### Security logic is correct
- **Origin comparison** (`sameMCPOrigin`) is scheme + host + *effective* port, case-insensitive, with default-port normalization (`80`/`443`). So `https://h/x` ≡ `https://H:443/y`, and an `https→http` downgrade is correctly treated as cross-origin. The table test covers all four cases.
- **Redirect target is compared to the *original* URL** (`via[0].URL`), not just the previous hop — so a multi-hop chain can't walk off-origin one same-origin step at a time. This is the subtle thing a naive implementation gets wrong; this one gets it right.
- **The bearer/custom header never reaches a cross-origin target**: `CheckRedirect` returns the error *before* the client issues the next request, so the OAuth RoundTripper never runs for the target. All three no-leak tests assert `targetHits == 0`, and confirm the header still reaches the configured origin.
- **SSE endpoint is origin-pinned** (`resolveSSEEndpointURL`) before any message POST, closing the "malicious `event: endpoint` points cross-origin" vector — and the POST itself also rides the guarded client, so it's belt-and-suspenders.
- Redirect chain is bounded (`maxMCPRedirects`).

### Coverage is complete for the data plane — I checked for bypasses
Both remote entry points build their client through `oauthHTTPClient` → `mcpHTTPClient` (HTTP at `connectNetwork`, SSE at `connectRemoteSSE`), and every `client.client.Do(...)` uses it. The non-OAuth path now gets the guard too (previously bare `http.DefaultClient`). I grepped the package for other client construction — no MCP request path bypasses the guard.

### It doesn't break legitimate behavior
- `TestHTTPClientFollowsSameOriginRedirect` proves same-origin redirects (incl. default-port equivalence) still work end to end with the header intact.
- Non-OAuth servers switching from the shared `http.DefaultClient` to a per-server `mcpHTTPClient(server, nil)` keeps `http.DefaultTransport` (nil transport), so connection pooling and timeout behavior are unchanged.
- Full `internal/mcp` suite passes locally; `go build ./...`, `go vet`, `gofmt` clean. (Race detector runs in CI on Linux — the new tests use `sync/atomic` correctly.)

### One optional follow-up (non-blocking, out of this PR's scope)
The OAuth **control-plane** clients — token exchange / metadata discovery in `oauth.go` and the refresh `storeTokenSource` (`network_client.go:892`) — still use `http.DefaultClient` without this guard. A 307/308 redirect there preserves method+body, so in principle a hostile authorization-server-metadata could bounce a token request cross-origin. Different threat model (requires compromising AS discovery) and clearly outside "MCP data traffic stays on origin", but extending the same guard to those clients would be a tidy defense-in-depth follow-up.

Nice, careful work — approving. Thanks @GautamBytes, and good call reporting privately first.

kevincodex1 (approved):
LGTM

PR #397, opened 2026-07-02
fix(tui): resolve every permission request so the agent can't deadlock

gnanam1990:
## Problem

Runs got "stuck forever" — the TUI showed `Working · thinking` climbing for 28–33 minutes with no prompt on screen and no network activity, until the user killed it. Intermittent (~5/10) and model-dependent: it happened on `chatgpt/gpt-5.5` but not `ollama/glm-5.2`.

A goroutine dump of a live stuck process nailed it. The agent goroutine was parked forever:

```
internal/tui/model.go   (select on the permission decision channel)
internal/agent/loop.go:requestPermission
internal/agent/loop.go:executeToolCall
internal/agent/loop.go:Run
```

blocked on a decision channel that nothing would ever signal.

## Root cause

A shell command whose sandbox decision is **Allow** but that also requests additional sandbox permissions (`sandbox_permissions: "with_additional_permissions"`) makes `shouldRequestPermission` return true — so the agent calls `OnPermissionRequest` and blocks — **but** `buildPermissionEvent` stamped the event `Action = allow` (from the sandbox Allow), not `prompt`.

The TUI's `permissionRequestMsg` handler only rendered/resolved `Action == prompt` requests and **silently dropped everything else** (early `return m, nil`). So the `decide` callback was never called, and because it was the active run its context was never cancelled — the agent parked indefinitely with no visible prompt.

Model-dependent because it depends on the model emitting that flag: gpt-5.5's folder-creation command set `with_additional_permissions`; glm's did not — which is exactly why one hung and the other ran clean.

## Fix (two parts, atomic — they must land together)

- **agent (`buildPermissionEvent`):** an additional-permissions request is a sandbox **elevation** the user must consent to, even over a base Allow, so it is now marked `Action = prompt` (matching `shouldRequestPermission`'s own condition). It surfaces as a real prompt instead of a dropped auto-allow.

- **tui (`permissionRequestMsg` handler):** enforce the contract that a blocked agent **always** receives a decision. A non-prompt action resolves immediately and **fails closed** (deny; `cancel` stays cancel) so the UI can never silently grant an unsurfaced permission; a superseded run (runID mismatch) resolves as `cancel` so its parked goroutine unblocks without depending on context-cancel ordering. This matches the ACP handler's always-resolve, fail-closed contract — the TUI was the only permission surface that could leave the agent hung.

## Testing

- **Red-green, both halves:** reverting the tui branch reproduces the deadlock (`decide` never called); reverting the agent override reproduces `Action=allow`.
- New tests: TUI handler resolves a non-prompt request fail-closed, cancels a superseded-run request, honors an explicit cancel, and still defers a real prompt to the user (no accidental auto-resolve); agent `buildPermissionEvent` prompts for an additional-permissions elevation and keeps allow for an ordinary allowed command.
- Pre-existing permission/ask-user tests unchanged and green.
- **End-to-end against a mock backend emitting the exact triggering tool call:** the real TUI now renders the permission prompt and the run **completes on approval**, where before it parked forever.
- Full suite `go test ./...` — 73/73 packages pass.

Fixes the "stuck / Working·thinking forever" hangs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Permission requests now prompt the user when a command asks for additional sandbox permissions, instead of being auto-approved.
  * Improved handling of permission prompts so stale or non-interactive requests no longer leave the app waiting indefinitely.
  * Non-prompt permission requests now resolve safely by default, reducing the risk of blocked or hung sessions.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -2159,6 +2159,18 @@ func buildPermissionEvent(call ToolCall, tool tools.Tool, args map[string]any, p
 		}
 	}
 
+	// A command that explicitly requests additional sandbox permissions
+	// (sandbox_permissions: with_additional_permissions) is an ELEVATION the user
+	// must consent to, even when the base command's sandbox decision was Allow.
+	// shouldRequestPermission blocks on OnPermissionRequest for exactly this case,
+	// so the event MUST be a prompt — otherwise it carried Action=allow while the
+	// loop waited on a decision, and a UI that renders only prompts (the TUI)
+	// dropped it, deadlocking the run. Keep this consistent with
+	// shouldRequestPermission's own condition.
+	if action == PermissionActionAllow && !permissionGranted && shellCommandAdditionalPermissionsRequested(args) {
+		action = PermissionActionPrompt
+	}
+
 	if safety.Permission == tools.PermissionAllow && action == PermissionActionAllow && !grantMatched && block == nil {
 		return PermissionEvent{}, false
 	}
--- a/internal/agent/permission_additional_perms_test.go
+++ b/internal/agent/permission_additional_perms_test.go
@@ -1,0 +1,70 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/sandbox"
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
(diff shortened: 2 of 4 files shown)


kevincodex1 (approved):
nice one! thanks bro

PR #398, opened 2026-07-02
feat(providers): split minimax zai into intl cn

JochenYang:
## Summary

- Add `minimaxi-cn` (api.minimaxi.com) sharing the model catalog with `minimax`
- Add `zai-cn`; `zai` now points at the international endpoint api.z.ai
- Split env vars so the two sides do not share a key: `ZAI_API_KEY` vs `ZHIPU_API_KEY` for zai/zai-cn, `MINIMAX_API_KEY` vs `MINIMAXI_API_KEY` for minimax/minimaxi-cn
- Map zai-cn and minimaxi-cn to the same models.dev providers as their international counterparts so the picker hydrates for both

## Breaking change: `zai` now defaults to the international endpoint

`zai` previously resolved to the China endpoint `open.bigmodel.cn` and accepted either `ZAI_API_KEY` or `ZHIPU_API_KEY`. It now points at the international endpoint `api.z.ai` and reads **only** `ZAI_API_KEY`. Users with bare `zai` configs (no persisted `base_url` / `api_key_env`) and a Zhipu-only key will see silent auth failures after upgrade.

**Migration for China users:**

- Change `catalog_id` to `zai-cn`
- Switch the env var from `ZAI_API_KEY` (or `ZHIPU_API_KEY`) to `ZHIPU_API_KEY`

The end-state is correct (`zai` internationally is `api.z.ai`), but the transition is not zero-config. Existing configs that already persisted `base_url` / `api_key_env` are unaffected — the resolver only fills missing fields, so the blast radius is bare `zai` configs without persisted overrides.

## Test

```bash
go test ./internal/providercatalog/... ./internal/providermodelcatalog/... ./internal/config/...
```

All three packages pass. TUI smoke-tested: the picker shows the same model list (sourced from models.dev) for both endpoints, and the env vars stay scoped to their respective side.

## Note on default model

The catalog default model for `zai` stays at `glm-4.5`; `glm-5.2` is only reachable at runtime through the picker''s live discovery, not as the catalog default.

## Summary by CodeRabbit

* **New Features**
  * Added a new MiniMax CN provider option with its own default base URL, model, and API key environment variable.
  * Split the Z.ai provider into separate international and China options, each with distinct base URL, default model, and API key environment variable.
  * Expanded curated model availability so the China options mirror the corresponding international model sets.
* **Tests**
  * Added/extended unit tests to verify default resolution, catalog ordering, alias mapping, and international/China model mirroring.


kevincodex1 (approved):
looks great to me. thnnks!

Vasanthdev2004 (changes requested):
## Verdict: Request changes — well-engineered split, but the `zai` change silently breaks existing configs; needs a migration note (+ a gofmt fix)

I checked this out, built it, ran `providercatalog` / `providermodelcatalog` / `config` (all green), and audited the catalog and env-var wiring. The engineering here is genuinely good — I want to call that out before the asks:

- **MiniMax side is a clean, non-breaking addition.** `minimax` is untouched; `minimaxi-cn` is purely additive with its own `MINIMAXI_API_KEY`. 👍
- **Test coverage is excellent** — resolver-default tests for all three entries, the `…MirrorInternational` contract tests pinning that CN and intl expose the same model IDs, the models.dev alias-mapping tests, and — importantly — both the `expectedCatalogIDs` list *and* the `ListByTransportPreservesCatalogOrder` order-lists are kept in sync. That's exactly the discipline these catalog inserts need.
- **The shared curated slices are safe.** I verified `Models()` → `dedupeModels()` always builds a fresh slice, so `zaiCuratedModels`/`minimaxCuratedModels` can't be aliased/mutated by callers. The comment's claim holds.

### 🔴 The `zai` change is a silent breaking change for existing users
This flips an **existing** catalog entry rather than only adding new ones:

| | before | after |
|---|---|---|
| `zai` endpoint | `open.bigmodel.cn` (China) | `api.z.ai` (international) |
| `zai` env vars | `ZAI_API_KEY`, `ZHIPU_API_KEY` | `ZAI_API_KEY` only |

So a user today on `zai` with a China account + `ZHIPU_API_KEY`, whose config relies on catalog defaults (e.g. `{name, catalog_id: "zai"}` with no persisted `base_url`/`api_key_env`), will after upgrade have their endpoint silently moved to `api.z.ai` **and** their `ZHIPU_API_KEY` stop being read — a 401 with no obvious cause. (Users whose profile already persisted `base_url`/`api_key_env` are fine; the resolver only fills missing fields — so the blast radius is "bare `zai` configs," not everyone.)

I actually think the **end-state is correct** — "Z.ai" internationally *is* api.z.ai, and `zai`→China was arguably the wrong mapping to begin with, so `zai` = intl / `zai-cn` = China is the right shape going forward. I'm not asking you to reverse it. But it shouldn't ship *silently*. Please either:

1. **(Recommended)** Keep the split as-is and add a migration note to the PR body + release notes: *"`zai` now defaults to the international api.z.ai endpoint and reads only `ZAI_API_KEY`. China (open.bigmodel.cn) users: switch `catalog_id` to `zai-cn` and use `ZHIPU_API_KEY`."* Pre-1.0 this is a fair price for the correct end-state — it just needs to be visible. Consider also keeping `ZHIPU_API_KEY` as a secondary accepted env on `zai-cn` (already done) — good.
2. **(Non-breaking alternative)** Leave `zai` pointing at its current China endpoint/env and introduce `zai-intl` for api.z.ai instead. Zero breakage, but perpetuates the "zai = China" oddity.

I'd take option 1 — just make the break loud.

### 🟡 gofmt — will fail CI's format check
`internal/providermodelcatalog/catalog.go` isn't gofmt-clean (map-value alignment on the `"zai":` line). Please run:
```
gofmt -w internal/providermodelcatalog/catalog.go
```

### Nits
- **PR description mismatch:** the body says "`zai` now points at api.z.ai and the glm-5.2 default," but the code (and your own test) keeps the default at `glm-4.5`. Either bump the curated default or fix the description so it doesn't imply glm-5.2 is the catalog default.
- **CI:** the failing `Smoke (ubuntu-latest)` is **not your change** — it's a runner cache-extraction flake (`tar: … Cannot open: File exists` restoring the Go toolchain cache); macOS + Windows smokes passed. A re-run should clear it.

Add the migration note and the gofmt fix and I'm happy to merge. Nice, careful work @JochenYang — the test contracts in particular are the right call.

Vasanthdev2004:
fix conflict

gnanam1990 (commented):
The follow-up commit + PR-body edit resolved both of @Vasanthdev2004's merge-blocking asks — nice work:

- **gofmt is clean** now (the `providermodelcatalog/catalog.go` alignment on the `zai` entry).
- The **breaking change is now documented** — the PR body carries an explicit "`zai` defaults to the international endpoint" + "Migration for China users" section, which is exactly the Option-1 path Vasanth endorsed.

Verified on current head (`6fbc383`): `go build ./...` and `go vet` clean, `gofmt -l internal/` clean, and `providercatalog` / `providermodelcatalog` / `config` tests pass. The MiniMax side (`minimaxi-cn`) is purely additive and `minimax` is untouched.

One thing to fold in before merge (non-blocking): the breaking-change note lives **only in the PR body**. Neither commit uses a `BREAKING CHANGE:` footer / `feat!:`, and the `CHANGELOG` `[Unreleased]` section wasn't updated — so release-please will classify this as a plain `feat` and won't surface the `zai` change in generated release notes. Adding a `BREAKING CHANGE:` footer to the squash commit (or a `CHANGELOG` entry) ensures users who read release notes rather than the PR see it before their saved `zai` config flips to `api.z.ai` and 401s.

Otherwise this is good — the intentional China→international flip now ships loudly instead of silently.

PR #403, opened 2026-07-02
fix(provider-wizard): allow multiple custom OpenAI-compatible providers

### LINKED ISSUE #375: Not able to setup multiple openai comp providers

Baran3575:
## Problem

Issue #375: After setting up one custom OpenAI-compatible provider, configuring a second one is blocked. The wizard diverts to the Keep/Replace/Remove screen instead of allowing a new setup.

## Root Cause

`wizardProviderStoredKey` matches saved providers by `CatalogID`. All generic/custom providers share the same `CatalogID` (`"custom-openai-compatible"` or `"custom-anthropic-compatible"`), so saving one creates a match that blocks creating another.

## Fix

In `advanceProviderWizard` (`internal/tui/provider_wizard_discovery.go`), skip the ManageKey diversion for custom providers and route directly to the endpoint step. The user can:

- **Overwrite** an existing instance by re-entering the same base URL and name
- **Create a new** instance by entering a different base URL or name

Named providers (OpenAI, Anthropic, Groq, etc.) are **unaffected** — ManageKey is still shown for them.

## Changes

| File | Change |
|------|--------|
| `internal/tui/provider_wizard_discovery.go` | +14 lines — skip ManageKey for custom providers |
| `internal/tui/provider_wizard.go` | 0 lines — untouched |
| `internal/tui/provider_wizard_test.go` | +45 lines — 2 new tests |

## Testing

- `TestAdvanceProviderWizardCustomSkipsManageKey` — custom provider with saved key routes to Endpoint step, not ManageKey
- `TestAdvanceProviderWizardNamedShowsManageKey` — named provider with saved key still routes to ManageKey (no regression)
- All existing provider wizard tests pass (ManageKey remove, replace, keep, stored key matching)

Fixes #375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the provider setup wizard for accounts with saved credentials.
  * For non-OAuth providers where an endpoint is not required, the wizard now skips key management and proceeds directly to endpoint setup.
  * Model discovery is now started only when the wizard advances into the model-selection step.

* **Tests**
  * Added unit tests covering the custom-provider skip behavior and the named-provider key-management flow.

### DIFF
--- a/internal/tui/provider_wizard_discovery.go
+++ b/internal/tui/provider_wizard_discovery.go
@@ -44,6 +44,20 @@ func (m model) advanceProviderWizard() (model, tea.Cmd) {
 	// keep/replace/remove before re-entering credentials.
 	if m.providerWizard.step == providerWizardStepProvider && !m.providerWizard.oauthMode {
 		if name, ok := m.wizardProviderStoredKey(m.providerWizard.currentProvider()); ok {
+			// Generic/custom providers (custom-openai-compatible etc.) all share
+			// the same CatalogID — matching on CatalogID would block creating a
+			// second instance. Skip ManageKey and route directly to the endpoint
+			// step; the user can overwrite an existing instance by re-entering
+			// the same name, or create a new one with a different name.
+			if providerWizardNeedsEndpoint(m.providerWizard.currentProvider()) {
+				m.providerWizard.manageProviderName = ""
+				previous := m.providerWizard.step
+				m.providerWizard.advance()
+				if m.providerWizard.step == providerWizardStepModel && previous != providerWizardStepModel {
+					return m, m.providerModelDiscoveryCmd()
+				}
+				return m, nil
+			}
 			m.providerWizard.manageProviderName = name
 			m.providerWizard.manageKeyCursor = 0
 			m.providerWizard.err = ""
--- a/internal/tui/provider_wizard_discovery.go
+++ b/internal/tui/provider_wizard_discovery.go
@@ -46,23 +46,17 @@ func (m model) advanceProviderWizard() (model, tea.Cmd) {
 		if name, ok := m.wizardProviderStoredKey(m.providerWizard.currentProvider()); ok {
 			// Generic/custom providers (custom-openai-compatible etc.) all share
 			// the same CatalogID — matching on CatalogID would block creating a
-			// second instance. Skip ManageKey and route directly to the endpoint
-			// step; the user can overwrite an existing instance by re-entering
-			// the same name, or create a new one with a different name.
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (commented):
This is the fix we're going with for #375 — I closed the duplicate #389 in favor of this one. Nice approach: gating at the call site in `advanceProviderWizard` (rather than changing the `wizardProviderStoredKey` lookup) keeps the lookup's meaning intact, and the table test over both custom transports plus the named-provider guard is exactly the right coverage.

Verified locally: `go build ./...` clean, and the wizard suite passes — re-selecting the generic "Custom OpenAI-compatible" entry with one already saved now routes to the endpoint step instead of trapping in keep/replace/remove, while a named provider (e.g. `openai`) still gets keep/replace/remove.

**One thing before merge:** the new test file isn't gofmt-clean and will trip the format check in CI. Please run:

```
gofmt -w internal/tui/provider_wizard_test.go
```

(It's just struct-field alignment on the `step:` field in the two new tests.)

One heads-up, not a blocker: because the wizard's provider list comes from `providercatalog.All()`, custom providers only ever appear as the single generic entry — so after this change there's no longer a wizard path to replace/remove an existing *custom* provider's stored key (the generic descriptor's name won't match a user's saved name). That's an acceptable trade to unblock multi-provider setup, and the old behavior could only manage the first CatalogID match anyway — just flagging it as a good follow-up (list saved custom instances for management).

Once the gofmt is in and CI goes green, this is good to merge. Thanks @Baran3575!

gnanam1990 (approved):
Approving — this is the fix we're taking for #375.

Verified locally: `go build ./...`, `go vet`, and `go test ./internal/tui` are green, and the two added tests exercise the exact repro (a saved custom profile + re-selecting "Custom OpenAI-compatible" now routes to the Endpoint step instead of ManageKey) plus the regression boundary (a named provider still gets ManageKey). The change is localized to wizard navigation — it doesn't touch `providerWizardProfile` / `UpsertProvider` / `mergeProvider`, so an existing single custom-provider config resolves identically. Clean and well-scoped.

Note: this supersedes #389; I'm requesting changes there in favor of this one.

Baran3575:
@gnanam1990 @Vasanthdev2004 is it possible to run action and review again?

kevincodex1 (approved):
looks good to me!

PR #407, opened 2026-07-02
fix(install): persist install dir to user PATH on Windows

### LINKED ISSUE #406: install.ps1 never persists install dir to PATH, so zero is unusable after install
## Bug description

`scripts/install.ps1` downloads and installs `zero.exe` to `%LOCALAPPDATA%\zero\bin`, but it never adds that directory to the user's PATH. It only prints a hint:

```powershell
$pathEntries = $env:PATH -split [System.IO.Path]::PathSeparator
if ($pathEntries -notcontains $InstallDir) {
  Write-Host "Add $InstallDir to PATH to run zero from any directory."
}
```

## Reproduction steps

1. Run `irm https://raw.githubusercontent.com/Gitlawb/zero/main/scripts/install.ps1 | iex` on Windows.
2. Restart the terminal.
3. Run `zero`.

## Expected behavior

`zero` should be runnable from a fresh terminal after install, matching the behavior of most CLI installers (e.g. it should persist the install directory to the user PATH via `[Environment]::SetEnvironmentVariable`).

## Actual behavior

`zero` is not found — `'zero' is not recognized as an internal or external command`. The install directory was never added to PATH; the script only prints a suggestion to add it manually.

## Environment

- Windows 11
- Installed via the official PowerShell installer command from the README

PierrunoYT:
## Summary

The PowerShell installer only printed a suggestion to add the install directory to PATH but never actually did it, so `zero` was unusable from a fresh terminal after install. This change persists the install directory to the user's PATH via `[Environment]::SetEnvironmentVariable` and also updates the current session's `$env:PATH` so `zero` works immediately after install without a terminal restart.

## Linked issue

Fixes #406

## Checklist

- [x] `gofmt` clean.

### Verification notes

- `go build ./...` and `go vet ./...` pass clean.
- `go test ./...` has 4 pre-existing failures unrelated to this change (env has no provider/API key configured, a Windows file-lock permission error in `internal/oauth`, and a TUI string-match assertion in `internal/tui`) — none touch install/PATH logic, and this PR does not modify any `.go` files.
- `gofmt -l` flags nearly the whole repo pre-existing due to `core.autocrlf=true` on Windows (CRLF vs LF); not introduced by this change and no `.go` files are touched here.
- No automated tests exist for `scripts/install.ps1`; verified manually by reproducing the bug (fresh install, PATH unchanged, `zero` not found) and confirming the fix persists the entry via `[Environment]::SetEnvironmentVariable("PATH", ..., "User")` and that `zero` resolves afterward.
- Not a UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer now more reliably adds the install location to `PATH` only when it’s missing.
  * Current terminal sessions are updated without duplicating the path entry.
  * If automatic `PATH` updates fail, clearer warnings guide you to update it manually.
  * You’ll now be prompted to restart your terminal after a successful user `PATH` update.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -149,9 +149,20 @@ try {
   $targetPath = Join-Path $InstallDir "zero.exe"
   Write-Host "Installed $targetPath"
 
-  $pathEntries = $env:PATH -split [System.IO.Path]::PathSeparator
-  if ($pathEntries -notcontains $InstallDir) {
-    Write-Host "Add $InstallDir to PATH to run zero from any directory."
+  $userPath = [Environment]::GetEnvironmentVariable("PATH", "User")
+  $userPathEntries = @()
+  if (-not [string]::IsNullOrEmpty($userPath)) {
+    $userPathEntries = $userPath -split [System.IO.Path]::PathSeparator
+  }
+  if ($userPathEntries -notcontains $InstallDir) {
+    $newUserPath = if ([string]::IsNullOrEmpty($userPath)) { $InstallDir } else { "$userPath;$InstallDir" }
+    [Environment]::SetEnvironmentVariable("PATH", $newUserPath, "User")
+    Write-Host "Added $InstallDir to your user PATH. Restart your terminal to use 'zero'."
+  }
+
+  $sessionPathEntries = $env:PATH -split [System.IO.Path]::PathSeparator
+  if ($sessionPathEntries -notcontains $InstallDir) {
+    $env:PATH = "$env:PATH;$InstallDir"
   }
 } finally {
   if (Test-Path $tempDir) {
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -62,6 +62,19 @@ function Find-ZeroExtractedFile {
   throw "Release archive did not contain exactly one $FileName"
 }
 
+function Test-ZeroPathContainsDir {
+  param(
+    [string]$PathValue,
+    [string]$Dir
+  )
+
+  if ([string]::IsNullOrEmpty($PathValue)) {
+    return $false
+  }
+
+  return @($PathValue -split [System.IO.Path]::PathSeparator) -contains $Dir
+}
+
 function Find-ZeroOptionalExtractedDirectory {
   param(
     [string]$Root,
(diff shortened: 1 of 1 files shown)


gnanam1990 (approved):
Approving.

The high-consequence axes are all handled correctly: it *appends* to the User PATH (doesn't clobber), is idempotent (guarded by `Test-ZeroPathContainsDir`), writes User scope so it needs no elevation, and won't abort the install if the PATH write fails (try/catch + the current-session fallback still runs). The non-Windows installer is untouched, and CodeRabbit's earlier change-request (guard `SetEnvironmentVariable`) is addressed.

One non-blocking note: `[Environment]::GetEnvironmentVariable(..., 'User')` returns `%VAR%` entries already expanded, so writing back via `SetEnvironmentVariable` freezes any dynamic `%VAR%` PATH entries to literals (REG_EXPAND_SZ → REG_SZ). It's inherent to the .NET API and standard across install scripts, so fine to ship — a registry-level write with `ExpandString` would preserve them if you want to harden it later.

Vasanthdev2004 (approved):
## Approve — fixes a genuinely broken Windows install path, done carefully

**Worth it:** yes, high value. The Windows installer copies `zero.exe` into `%LOCALAPPDATA%\zero\bin` but never puts that directory on PATH, so after the advertised `irm … | iex` install, `zero` isn't found from a fresh terminal (#406). That makes the *primary* Windows install path effectively broken for new users. Thanks @PierrunoYT.

The PATH manipulation is done safely:
- **Idempotent** — `Test-ZeroPathContainsDir` splits the User PATH and uses `-contains`, which is case-insensitive in PowerShell (`C:\ZERO\BIN` matches `C:\zero\bin`), so re-running the installer won't duplicate the entry.
- **No clobber** — reads the existing User PATH and appends `;$InstallDir`, handling the empty case.
- **Correct scope** — reads and writes the `"User"` hive (HKCU), and reads the *User* PATH (not the process PATH), so it won't fold system entries into the user PATH — a common mistake this avoids.
- **Resilient** — try/catch with actionable warnings, updates the current session's `$env:PATH` so `zero` works immediately, and prints a restart hint.

I parse-checked the PR's script on Windows (PowerShell 5.1) — clean — and validated the helper's idempotency/case behavior locally.

**Non-blocking notes for a future pass** (don't hold merge): `[Environment]::GetEnvironmentVariable("PATH","User")` expands `%VAR%` tokens and writes back `REG_SZ`, so a User PATH containing e.g. `%USERPROFILE%\bin` gets flattened to a literal path (the standard .NET-installer pitfall; the path still works). And no `WM_SETTINGCHANGE` broadcast, so already-open shells need a restart — which the message already says.

**CI note:** the Go CI checks haven't run here because this is a **cross-repo fork PR whose workflows are gated on maintainer approval** — not because the change is script-only. A maintainer should trigger the workflows before merge. That said, the change touches only `scripts/install.ps1` (no Go, and `install.ps1` isn't exercised by the smoke job), so the risk of CI catching anything is low. Approving on the script review.

PR #385, opened 2026-07-02
fix(config): unbrick first-run setup — default google/anthropic models, enter setup on fixable config errors

gnanam1990:
Fixes #384.

## Problem

A provider entry with an API key but no `model` bricked Zero — reported twice in the wild against a fresh `npm install -g` (#384):

```
% zero
[zero] provider google requires model (apiKey=[REDACTED])
% zero setup
[zero] provider google requires model (apiKey=[REDACTED])
% zero config
[zero] provider google requires model (apiKey=[REDACTED])
% zero login
unknown command "login"
```

Chicken-and-egg: `zero config` and bare `zero setup` resolve the full config up front, so the exact commands that could fix the config died with the same error. The trigger doesn't even need a config file — `applyEnv` synthesizes a google provider from a `GEMINI_API_KEY`/`GOOGLE_API_KEY` exported in the user's shell, with no model unless `GEMINI_MODEL` is also set, which is how a completely fresh install hits this. Anthropic had the identical gap.

## Fix (commit 1): default missing google/anthropic models from the catalog

The resolver has always defaulted a missing model for the **openai** kind, and `catalogID`-carrying profiles get defaults via `applyCatalogDescriptor` — but catalog-less `provider_kind: google`/`anthropic` profiles (hand-written or env-synthesized) had no fallback. They now fall back to their provider-catalog default models (`gemini-2.5-pro` / `claude-sonnet-4.5`), same shape and same gate as openai (option renamed `defaultOpenAIModel` → `defaultModels`). An explicit model always wins; provider-command configs (`ZERO_PROVIDER_COMMAND`) still get no invented models (pinned by a test).

## Fix (commit 2): enter setup on the remaining fixable shape + `zero login` hint

- Custom openai-/anthropic-compatible endpoints have no catalog default (Zero can't guess a gateway's model), so that shape still failed resolve — and still bricked bare `zero`/`zero setup`. The failure is now tagged `ErrProviderRequiresModel` via a **message-preserving** wrapper (multi-target `Unwrap`), and the interactive TUI's fixable-error fallback (previously `ErrNoActiveProvider` only, from #372) also drops into the setup wizard for it. Headless commands (`zero config`, `zero exec`) still fail with the actionable message, now with a hint: `provider gw requires model — add "model" to its entry in config.json, or re-run: zero setup <catalog-id> --model <model>`. Every other resolve error (malformed JSON, directory conflicts) stays fatal.
[...]

PR #410, opened 2026-07-02
fix(config): fall back to a usable saved provider instead of forcing full re-onboarding

### LINKED ISSUE #409: no active provider marked forces full re-onboarding instead of using an already-configured provider
## Bug description

Running bare `zero` prints (or, on the stale v0.1.0 release binary, crashes with) `active provider "" not found` and does not fall back to any already-configured provider, even when `config.json` has one or more usable providers saved — it just has no `activeProvider` set (blank/stale field), or has 2+ providers where none was ever explicitly selected active.

Root cause: `normalizeProviders` in `internal/config/resolver.go` discards the already-normalized provider list when no active provider is found (`ErrNoActiveProvider`), returning `nil` instead of the list. The interactive TUI (`internal/cli/app.go`, `runInteractiveTUIWithSetup`) already has fallback logic (`firstUsableProvider`) that picks an already-configured, credentialed provider instead of forcing a fresh onboarding wizard — but that fallback only runs for a *different* failure shape (active provider present but missing a credential). For `ErrNoActiveProvider` specifically, the code unconditionally blanks the resolved config and forces the full setup wizard, discarding providers the user already configured and authenticated.

## Reproduction steps

1. Have a `config.json` with 2+ providers configured (e.g. via `zero provider add`), each with valid credentials, but no `activeProvider` set (or a stale one referencing a provider that no longer exists).
2. Run bare `zero`.

## Expected behavior

Zero should fall back to one of the already-configured, usable providers (same as it already does when the active provider lacks a credential), rather than forcing a full re-onboarding flow.

## Actual behavior

The resolve error discards the provider list entirely and the interactive TUI is forced into onboarding, even though a usable provider is already saved.

## Additional note

Separately: the currently published `v0.1.0` release predates an earlier related fix (`027a8f2`, "use ~/.config on macOS and enter setup when no provider (#371) (#372)") that's already on `main` — so users installing via `scripts/install.ps1` today get a binary that hard-crashes on this instead of even reaching the setup-wizard fallback. That's a release/packaging gap, not something a source fix alone resolves; a new release cut from current `main` would need to be published separately.

PierrunoYT:
## Summary

When `config.json` has providers configured but none marked active (blank/stale `activeProvider`, or 2+ providers where none was ever selected), the interactive TUI discarded the already-normalized provider list and forced a full re-onboarding wizard — even though a usable, already-authenticated provider was already saved.

- `normalizeProviders`/`Resolve` (`internal/config/resolver.go`) now return the normalized providers list alongside `ErrNoActiveProvider` instead of discarding it.
- `runInteractiveTUIWithSetup` (`internal/cli/app.go`) tries the existing `firstUsableProvider` fallback on that list before forcing setup. Onboarding only runs when no configured provider is actually usable — the same behavior already used for a different flavor of this problem (active provider present but missing a credential).

## Linked issue

Fixes #409

## Checklist

- [x] `gofmt` clean.

### Verification notes

- Added `TestResolveKeepsNormalizedProvidersWhenNoneMarkedActive` (`internal/config/resolver_test.go`) and `TestRunNoArgsFallsBackToUsableProviderWhenNoneMarkedActive` (`internal/cli/app_test.go`); both pass, along with the pre-existing adjacent tests (`TestRunNoArgsEntersSetupWhenResolveReportsNoActiveProvider`, `TestResolveRejectsActiveProviderWithoutConfiguredProfiles`) confirming the "genuinely nothing configured" path is unchanged.
- `go test ./internal/config/... ./internal/cli/...` has the same 8 pre-existing failures present on `main` before this change (no active provider/API key configured in this sandbox environment) — none are new or related to this change.
- `gofmt -l` on the touched files flags them, but they were already flagged identically before this change (repo-wide `core.autocrlf=true` CRLF artifact on Windows) — not introduced here.
- Not a UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * The app now continues using an available configured provider when no active provider is set, instead of forcing the setup wizard.
  * Configuration resolution now preserves the normalized list of configured providers when the active provider is missing, enabling smoother fallback behavior.
  * Added CLI and configuration tests to verify correct “no active provider” behavior and prevent regressions.


kevincodex1 (approved):
looks good to me!

gnanam1990 (approved):
Approving.

Verified locally (`build` / `vet` / `test ./internal/config ./internal/cli` all green). The change is correctly scoped to the resolve-error branch — a config with a valid active provider hits `err == nil` and bypasses it entirely — and the #385 unbrick path is preserved: the `ErrProviderRequiresModel` case still forces setup, and the pre-existing onboarding test still passes.

Two non-blocking follow-ups (not merge blockers): `firstUsableProvider` gates on credential presence but not on `Model != ""`, so a modelless custom-endpoint provider could be selected into a soft-failing launch (recoverable via `/model`) — it's a pre-existing property of that helper, not introduced here. And it may be worth carrying the non-provider config sections (Sandbox/Tools/Swarm) on the fallback return so the session honors them.

Vasanthdev2004 (approved):
## Verdict: Approve ✅ — genuinely useful bug fix, correct, and it resolves the credential the right way

**Is it worth it?** Yes, clearly. Being force-marched back through onboarding — re-entering API keys you already saved — just because `activeProvider` is blank/stale (or you have 2+ providers and never picked one) is a real, high-friction bug. This is exactly the kind of fix that earns its keep.

I reviewed it as a credential-adjacent change (that's where this can go subtly wrong) and traced it end to end.

### The fix is correct
- **The list is preserved on the way up.** `normalizeProvidersWithOptions` now returns the normalized profiles alongside `ErrNoActiveProvider` instead of `nil`, and `Resolve` forwards them as `ResolvedConfig{Providers: providers}`. Other error paths (`active.Model == ""`, etc.) still return `nil`, so only the no-active-provider case surfaces a list — no accidental leakage of a partial list for other failures.
- **The fallback only fires when it should.** It's gated on `errors.Is(err, ErrNoActiveProvider) && firstUsableProvider(...) ok`; anything else still forces setup. `ErrProviderRequiresModel` is unaffected.
- **It picks a genuinely usable provider — including stored-key ones.** This was my main worry: the resolver is pure (no secret I/O), so `resolved.Providers` carry an empty `APIKey` with the key in the credstore. I confirmed `firstUsableProvider` → `setupMissingCredentialEnv` → `HasConfiguredCredential`, which counts `APIKeyStored` (and auth headers), so a normally-configured stored-key provider is correctly treated as usable — not just the inline-key case.
- **No keyless-provider trap.** The fallback sets `resolved.Provider` to a profile that then flows through the *same* `buildProvider(resolved, deps)` path as a normally-active provider, so its credential is loaded identically. (This is the same pattern the pre-existing `firstUsableProvider` fallback at app.go:588 already uses in production.)

Local: `go build ./...`, `gofmt` clean; full `internal/config` and the targeted `internal/cli` tests pass. Tests added are real and on-point (`TestResolveKeepsNormalizedProvidersWhenNoneMarkedActive`, `TestRunNoArgsFallsBackToUsableProviderWhenNoneMarkedActive`).

### Two small notes (non-blocking)
1. **Test the stored-key path too.** `TestRunNoArgsFallsBackToUsableProviderWhenNoneMarkedActive` uses an inline `APIKey: "sk-test"`. The code correctly handles the far more common `APIKeyStored: true` (credstore) case via `HasConfiguredCredential`, but the test doesn't exercise it — a variant with `APIKeyStored: true` and no inline key would lock in the real-world scenario. Worth adding, not worth blocking.
2. **Minor redundancy.** There are now two `firstUsableProvider(resolved.Providers)` fallback sites — the new one in the `ErrNoActiveProvider` block and the pre-existing one at app.go:588. The new one is clear and correct where it sits; just noting the same fallback now exists in two places in case you'd rather consolidate later.

### CI note
The red `Smoke (windows-latest)` is **not this PR** — it's `TestLoadOrCreateSecretConcurrentConverges` in `internal/oauth` (a concurrency flake in secret-file creation), unrelated to config/cli. macOS + Ubuntu smokes passed. A re-run should clear it.

Approving. Thanks @PierrunoYT — clean fix, and reusing the existing `firstUsableProvider` semantics is exactly right.

PR #412, opened 2026-07-02
fix(tools): flag piped POSIX utilities before running on Windows

### LINKED ISSUE #411: bash tool wastes a turn piping into head/tail/grep/etc on Windows (cmd.exe has no equivalent)
## Bug description

Zero's bash tool always executes commands through native `cmd.exe` on Windows (`shellExecutable()` in `internal/tools/bash.go`). There's a pre-flight detector (`detectShellCommandIssue` in `internal/tools/shell_runtime.go`) that catches bash-style `cd /...` and bare `ls` and blocks them with a helpful cmd.exe-translation hint before running — but it doesn't catch the broader family of common POSIX pipe utilities the model reaches for, like `head`, `tail`, `grep`, `wc`, `awk`, `sed`, `cut`, `xargs`, `tr`. Since cmd.exe has none of these, the command always fails, wasting a full turn.

## Reproduction steps (as observed in the wild)

On Windows, ask Zero to run something like:
```
git log --oneline -- somefile.py | head
```
or
```
git blame -L 40,46 somefile.py --format="%H %an" | grep summary
```

## Actual behavior

The command runs, cmd.exe rejects the piped-to utility (`'head' is not recognized as an internal or external command...`), and the turn is wasted. On non-English Windows locales this error text is also localized, so `detectShellOutputIssue`'s post-hoc English-string matching never even adds Zero's own remediation hint — the model sees a bare, untranslated OS error.

## Expected behavior

Piping into a known-missing POSIX utility should be caught pre-flight (like the existing `cd`/`ls` checks) with a clear suggestion, instead of wasting a turn on a guaranteed cmd.exe failure.

PierrunoYT:
## Summary

Zero always runs the bash tool through native `cmd.exe` on Windows. The existing pre-flight Windows-syntax detector (`detectShellCommandIssue`) only caught bash-style `cd /...` and bare `ls`, not the broader family of common POSIX pipe utilities (`head`, `tail`, `grep`, `wc`, `awk`, `sed`, `cut`, `xargs`, `tr`) that cmd.exe has no equivalent for. Piping into any of these (e.g. `git log ... | head`, a `git blame ... | grep` pipeline) always failed, wasting a turn — reported in the wild.

This adds a `windowsPosixUtilityPattern` check alongside the existing `cd`/`ls` detectors in `internal/tools/shell_runtime.go`, blocking these pre-flight with the same "translate to cmd.exe / use Zero's native tools" guidance already used for `cd`/`ls`.

Scoped narrowly to the pre-flight command-text check (locale-independent, since it inspects the model-generated command rather than cmd.exe's own translated error output) rather than trying to also expand the post-hoc output-string matcher, which can never be locale-complete.

## Linked issue

Fixes #411

## Checklist

- [x] `gofmt` clean.

### Verification notes

- Added `TestDetectShellCommandIssueFlagsPipedPosixUtilities` and `TestDetectShellCommandIssueAllowsUnrelatedCommands` (`internal/tools/bash_tool_test.go`); both pass alongside the existing `cd`/`ls` detector tests.
- `go test ./internal/tools/...` passes in full.
- `gofmt -l` on the touched files: pre-existing repo-wide CRLF flag (`core.autocrlf=true` on Windows), not introduced by this change — same as prior PRs against this repo.
- Verified false-positive safety by hand: `git grep foo` (real git subcommand, not a pipe target) and `findstr /r "..."` are unaffected since the pattern only matches at the start of a command or immediately after a `|`/`&`/`;` separator, not as a positional argument.
- Not a UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows shell command detection to flag pipelines that use common POSIX-style utilities (for example, `head`, `grep`, `wc`, `tail`) where they aren’t supported.
  * Reduced false positives by allowing unrelated Windows-appropriate commands (such as `git`, `dir`, `findstr`, and `echo`) to pass through unflagged.
* **Tests**
  * Added coverage to verify both the new Windows pipeline detection and the unchanged behavior for unrelated commands.

### DIFF
--- a/internal/tools/bash_tool_test.go
+++ b/internal/tools/bash_tool_test.go
@@ -180,6 +180,9 @@ func TestDetectShellCommandIssueAllowsUnrelatedCommands(t *testing.T) {
 		`dir /b`,
 		`findstr /r "summary" file.txt`,
 		`echo header text`,
+		`tail-log.sh`,
+		`grep-cli --version`,
+		`sed.exe --help`,
 	} {
 		if issue := detectShellCommandIssue(command, "windows"); issue != nil {
 			t.Fatalf("expected unrelated command to pass for %q, got %#v", command, issue)
--- a/internal/tools/bash_tool_test.go
+++ b/internal/tools/bash_tool_test.go
@@ -161,6 +161,32 @@ func TestDetectShellCommandIssueRequiresActualLSCommand(t *testing.T) {
 	}
 }
 
+func TestDetectShellCommandIssueFlagsPipedPosixUtilities(t *testing.T) {
+	for _, command := range []string{
+		`git log --oneline -- pinokio_compat.py | head`,
+		`git blame -L 40,46 pinokio_compat.py --format=%author | grep summary`,
+		`cat file.txt | wc -l`,
+		`some_command | tail -n 20`,
+	} {
+		if issue := detectShellCommandIssue(command, "windows"); issue == nil {
+			t.Fatalf("expected POSIX utility pipeline to be flagged for %q", command)
+		}
+	}
+}
+
+func TestDetectShellCommandIssueAllowsUnrelatedCommands(t *testing.T) {
+	for _, command := range []string{
+		`git log --oneline`,
+		`dir /b`,
+		`findstr /r "summary" file.txt`,
+		`echo header text`,
+	} {
+		if issue := detectShellCommandIssue(command, "windows"); issue != nil {
+			t.Fatalf("expected unrelated command to pass for %q, got %#v", command, issue)
+		}
+	}
+}
+
 func TestDetectShellOutputIssueAddsWindowsSyntaxHint(t *testing.T) {
 	issue := detectShellOutputIssue(`cd /d/tmp/zero-pr-158 && ls -la`, "The syntax of the command is incorrect.", "windows")
 	if issue == nil {
--- a/internal/tools/shell_runtime.go
+++ b/internal/tools/shell_runtime.go
@@ -20,6 +20,11 @@ type shellIssue struct {
 var (
 	windowsBashStyleCDPattern = regexp.MustCompile(`(?i)(^|[&|;]\s*)cd\s+/(?:[a-ce-z0-9_./~-]|d[a-z0-9_./~-])[a-z0-9_./~-]*`)
 	windowsLSCommandPattern   = regexp.MustCompile(`(?i)(^|[&|;]\s*)ls\b(?:\s+|$)`)
+	// windowsPosixUtilityPattern catches the other common POSIX coreutils/grep
+	// family commands cmd.exe has no equivalent for (unlike `dir`/`findstr`,
+	// which sound similar enough that a model might reach for the POSIX name
+	// instead) — most often seen piped in, e.g. `git log ... | head`.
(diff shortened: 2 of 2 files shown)


kevincodex1 (approved):
LGTM

gnanam1990 (approved):
Approving.

The hard OS gate (`if goos != "windows" { return nil }`, with callers passing the real host GOOS) makes this a no-op off Windows, so there's no non-Windows regression. Verified `build` / `vet` / `test ./internal/tools` green, and the `(?:\s+|$)` anchor correctly avoids the earlier false positives (`tail-log.sh`, `grep-cli --version`, `git grep foo` all pass) — locked in by the new test.

Non-blocking nit: the regex is quote-blind, so a utility name inside a quoted argument (e.g. `git commit -m "fix; sed x"`) can false-block on Windows. It mirrors the existing `windowsBashStyleCDPattern` / `windowsLSCommandPattern` in the same file, so it's consistent with what's there — a quote-aware split would tighten it later.

anandh8x (approved):
Clean, well-scoped fix. The second commit correcting `\b` → `(?:\s+|$)` shows good attention to false-positive safety — `tail-log.sh`, `grep-cli --version`, and `sed.exe --help` all pass correctly.



LGTM.

Vasanthdev2004 (approved):
## Approve — useful, correct, and verified on Windows

**Worth it:** yes. This fixes a concretely reported problem (#411): Zero always runs the bash tool through `cmd.exe` on Windows, and models routinely pipe to POSIX utilities (`git log … | head`, `git blame … | grep`) that `cmd.exe` has no equivalent for, so every such pipeline fails and burns a turn. The existing pre-flight detector only caught `cd`/`ls`; this extends it to the common coreutils/grep family. Small, well-scoped, Windows-only, locale-independent. Thanks @PierrunoYT.

**No #414-class regression.** This is #414's sibling, but crucially it does **not** modify any existing suggestion string — it adds a *new* pattern + a *new* `shellIssue` alongside the old ones, plus new tests. I verified on this Windows machine:
- `go test ./internal/tools/ -run 'TestDetectShell|TestBashToolBlocksInteractive|TestBashToolAllowsNonInteractive' -v` — all pass, **including the exact test #414 broke** (`TestBashToolBlocksInteractiveCommandThroughSandbox`).
- `go test ./internal/tools/ ./internal/sandbox/` — both green; `go build ./...`, `go vet`, `gofmt -l` clean.

The new detector runs before interactive-command detection, but since the pattern excludes `vim`/`less`, it doesn't intercept the interactive guards (the #414 canaries still fire). Good `dir`/`findstr`/`git`/`echo` exclusion coverage in the tests. CI all 7 green including `Smoke (windows-latest)`.

**One non-blocking note** (not a fix required): the block is a *hard* pre-flight `StatusError` — the command never runs. On a Windows box with Git-for-Windows/scoop/choco coreutils on PATH (very common among exactly these Windows devs), `git log | grep foo` would actually succeed via `cmd.exe` piping to `grep.exe`, but is now blocked. It's model-facing and recoverable in one turn, and consistent with the existing `ls` hard-block — so approve stands — but a soft *hint* rather than a hard block would be the cleaner long-term design if you want a follow-up.

Approving.

PR #414, opened 2026-07-02
fix(sandbox): Windows-appropriate suggestions when blocking interactive commands

### LINKED ISSUE #413: interactive-command blocker suggests POSIX-only commands (cat/head/tail/ps aux) on Windows
## Bug description

`DetectInteractiveCommand` (`internal/sandbox/safe_command.go`) correctly blocks interactive programs (pagers, process monitors, `tail -f`, etc.) before they hang the non-interactive agent — but the remediation text it returns hardcodes POSIX-only alternatives regardless of platform: pagers (`less`/`more`/`most`) suggest `cat`/`head`/`tail -n N`; process monitors (`top`/`htop`/`btop`/`btm`) suggest `ps aux`; `tail -f` suggests `tail -n N <file>`. None of `cat`/`head`/`tail`/`ps` exist on Windows cmd.exe (which is what Zero's bash tool actually executes through on Windows — see #411/#412), so the block correctly stops the interactive command but then points the model at three more commands it also can't run.

## Reproduction (observed in the wild)

On Windows, run something like:
```
type blame_out.txt | more +0
```
Zero blocks `more` (correct — it's an interactive pager) but returns: "Use `cat`, `head`, or `tail -n N` to print file contents non-interactively." None of those exist on Windows.

## Expected behavior

The suggestion should be platform-appropriate: `type <file>` (or Zero's own `read_file` tool) instead of `cat`/`head`/`tail`, `tasklist` instead of `ps aux`, etc.

PierrunoYT:
## Summary

`DetectInteractiveCommand` already receives `goos`, but its remediation text for pagers (`less`/`more`/`most`), process monitors (`top`/`htop`/`btop`/`btm`), and `tail -f` hardcoded POSIX-only alternatives (`cat`, `head`, `tail -n N`, `ps aux`) regardless of platform. On Windows none of those exist either (Zero's bash tool executes through native cmd.exe there), so blocking e.g. `type file | more` correctly stopped the interactive command but then pointed the model at three more commands cmd.exe doesn't have — undermining the guidance this check exists to give. Reported in the wild.

Adds an optional `windowsSuggestion` per program/segment, used in place of the POSIX suggestion when `goos == "windows"`: `type <file>` (or the `read_file` tool) for pagers/`tail -f`, `tasklist` for process monitors. Non-Windows suggestions are unchanged. Scoped to the cases where the POSIX suggestion was the *sole* remediation offered — left `vim`/`vi`/etc.'s secondary `sed -i` mention alone since `edit_file`/`write_file` is already their correct primary suggestion.

## Linked issue

Fixes #413

## Checklist

- [x] `gofmt` clean.

### Verification notes

- Added `TestDetectInteractiveCommandSuggestsWindowsAlternativesOnWindows` (`internal/sandbox/safe_command_test.go`), covering `more`/`less` → `type`, `htop` → `tasklist`, `tail -f` → `read_file`, and confirming Linux keeps the original POSIX suggestion unchanged.
- `go test ./internal/sandbox/...` passes in full, including all pre-existing interactive-command tests.
- `gofmt -l` on the touched files: pre-existing repo-wide CRLF flag (`core.autocrlf=true` on Windows), not introduced by this change — same as prior PRs against this repo.
- Not a UI change.

## Summary by CodeRabbit

* **New Features**
  * Improved interactive command guidance on Windows, offering Windows-friendly alternatives for common commands like process monitors and file-following.
  * Interactive command detection now provides more relevant suggestions when a Windows-specific option is available.
* **Bug Fixes**
  * Kept existing non-Windows guidance intact while making command suggestions more accurate across platforms.
* **Tests**
  * Added coverage to verify Windows-specific suggestions are shown on Windows and POSIX-style suggestions remain unchanged on Linux.


kevincodex1 (approved):
LGTM

gnanam1990 (approved):
Approving.

The OS gate is airtight — the new Windows suggestion text sits behind `goos == "windows" && windowsSuggestion != ""`, so on macOS/Linux the suggestion strings are byte-for-byte unchanged and the block decision itself is untouched. Verified `build` / `vet` / `test ./internal/sandbox` green.

Non-blocking nit (same as CodeRabbit's): the new Windows table tests cover `more`/`less`/`htop`/`tail -f` but not `most`/`top`/`btop`/`btm`, which also gained `windowsSuggestion` entries — worth adding for completeness.

Vasanthdev2004 (changes requested):
## Verdict: Request changes — the change itself is right, but it breaks an existing test on Windows (the very platform it targets)

The core idea is correct and well-scoped: on Windows the POSIX remediations (`cat`/`head`/`tail -n`/`ps aux`) don't exist in cmd.exe either, so blocking `type file | more` and then pointing the model at three *more* non-existent commands defeats the purpose. Routing pagers/`tail -f` → `type`/`read_file` and monitors → `tasklist` is the right fix, and the `windowsSuggestion`-overrides-`suggestion` design is clean and applied to both the segment and program paths. The added `internal/sandbox` test is good.

**But this PR breaks `internal/tools.TestBashToolBlocksInteractiveCommandThroughSandbox` on Windows** — and CI's `Smoke (windows-latest)` is red because of exactly that. I reproduced it on a real Windows machine:

```
--- FAIL: TestBashToolBlocksInteractiveCommandThroughSandbox (0.00s)
    bash_tool_test.go:490: expected pager guard message with cat suggestion,
    got "...Suggestion: Use `type <file>` to print file contents non-interactively, or the read_file tool."
```

That test runs `less /etc/hosts` through the bash tool and asserts the output contains **`cat`** (bash_tool_test.go:489). The bash tool calls `DetectInteractiveCommand` with the *runtime* GOOS (not an injected one), so on Windows it now gets your new `type`/`read_file` suggestion, and the `cat` assertion fails. On Linux/macOS it still contains `cat`, which is why a local `go test ./...` on a non-Windows box passes and the checklist item looked green — the Windows code path only runs on Windows.

### Fix — make that assertion GOOS-aware
`runtime` is already imported in that test file. I verified this makes it pass on Windows:

```go
wantSuggestion := "cat"
if runtime.GOOS == "windows" {
    wantSuggestion = "type"
}
if !strings.Contains(result.Output, "interactive") || !strings.Contains(result.Output, wantSuggestion) {
    t.Fatalf("expected pager guard message with %q suggestion, got %q", wantSuggestion, result.Output)
}
```

(That's the only failing test — `internal/sandbox` and the rest of `internal/tools` pass on Windows. So this one edit closes it out.)

### One process note
Please run the suite on Windows (or at least this package) before the next push — this is a Windows-behavior change, so the non-Windows `go test ./...` in the checklist can't actually exercise it. Everything else here is solid; fix the broken assertion and I'm happy to merge. Thanks @PierrunoYT — good catch on the original issue, the guidance is genuinely better now.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: none identified

Re-review of new commits since my prior review (`c82d4e30` → `97231505`). The new commit adds 4 lines to `safe_command_test.go` and adjusts 6 lines in `bash_tool_test.go` to cover remaining Windows suggestion override scenarios. No production code changes — test coverage expansion only.

## BUILD / TEST

- CI: CodeRabbit passed; other CI checks not shown for the latest commit but the changes are test-only additions with no production code impact

## FINDINGS

None. The additional test coverage tightens the regression net for Windows-specific command suggestions.

## BOTTOM LINE

Test-only expansion covering remaining Windows suggestion scenarios. Safe to merge.

Vasanthdev2004 (approved):
Good catch. The guard was correctly blocking the interactive command but then handing the model cat/head/tail/ps aux, none of which exist under cmd.exe — so the hint was useless exactly where it was needed. The override is additive and gated on Windows, so nothing else changes, and the tests cover both paths.

Marked #413 approved so this is clear to land. One thing for whoever merges: the full CI matrix hasn't run on the latest here — a quick rebase will re-trigger it and get it green. Approving on the code.

Vasanthdev2004 (approved):
I reviewed the Windows interactive-suggestion changes for issue #413. The suggestions now render and navigate correctly under the Windows codepath where they were getting dropped, and the tests exercise the Windows-specific branch. Approving.

This PR is currently conflicting with main, so please rebase onto latest main. Heads up that main also has an unrelated test-build break right now (PR #589 is the one-line fix, queued for merge); once #589 lands and you rebase, Smoke should go green. Approving once the rebase clears.

kevincodex1 (approved):

Issue #416, opened 2026-07-02
Rebindable keyboard shortcuts


eli-l:
### Problem

The built-in shortcuts may be in conflict with shell/terminal. Unique combinations might be too complex.

### Requested behavior

It should be possible to remap the bindings via the config file.

### Why it belongs in Zero

Keybindings hardcoded in Zero can be intercepted by wrappers and never passed to Zero.

Vasanthdev2004:
This shipped in #417, now merged to main. You can remap the toggle shortcuts from `config.json` under a `keybindings` block:

```json
"keybindings": {
  "toggleDetailed": "ctrl+y",
  "toggleMouse": "ctrl+e",
  "cycleReasoning": "ctrl+t",
  "togglePlan": "ctrl+p",
  "toggleSidebar": "ctrl+u"
}
```

Defaults are ctrl+o / ctrl+e / ctrl+t / ctrl+p / ctrl+b; any key you don't set keeps its default, and the `?` overlay and on-screen hints reflect whatever you configure. Closing as resolved.

PR #417, opened 2026-07-02
Make shortcuts reconfigurable via the config file.

eli-l:
## Summary

Allows to configure keyboard shortcuts mapping to prevent conflicts with terminal/shell.

Example `config.json`

```
  "keybindings": {
    "toggleDetailed": "ctrl+y",
    "toggleMouse": "ctrl+e",
    "cycleReasoning": "ctrl+m",
    "togglePlan": "ctrl+k",
    "toggleSidebar": "ctrl+u"
  }
```
Screenshot 2026-07-03 at 11 08 58

## Linked issue

https://github.com/Gitlawb/zero/issues/416

## Checklist

- [x] `gofmt` clean.

## Summary

* **New Features**
  * Added configurable TUI keyboard shortcuts for transcript details, mouse capture, reasoning cycling, plan/context expansion, and sidebar visibility.
  * Updated the `? shortcuts` overlay and on-screen hints to display your configured keys.

* **Bug Fixes**
  * Custom shortcut settings now consistently drive both key handling and displayed labels.

* **Bug Fixes / Behavior**
  * Improved sidebar show/hide handling to better respect saved preference even when the sidebar auto-hides.

* **Tests**
  * Added broader coverage for shortcut parsing, matching, and help-overlay substitution.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -743,6 +743,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		},
 		PermissionMode: permissionMode,
 		Notify:         resolved.Notify,
+		KeyBindings:    resolved.KeyBindings,
 		Setup: tui.SetupOptions{
 			Visible:    setupVisible,
 			Required:   needsSetup,
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -144,6 +144,7 @@ func Resolve(options ResolveOptions) (ResolvedConfig, error) {
 		Tools:          cfg.Tools,
 		Swarm:          cfg.Swarm,
 		Preferences:    cfg.Preferences,
+		KeyBindings:    cfg.KeyBindings,
 		LocalControl:   cfg.LocalControl,
 	}, nil
 }
@@ -225,6 +226,7 @@ func mergeConfig(dst *FileConfig, src FileConfig) {
 		dst.Preferences.Theme = strings.TrimSpace(src.Preferences.Theme)
 	}
 	mergeLocalControlConfig(&dst.LocalControl, src.LocalControl)
+	mergeKeyBindings(&dst.KeyBindings, src.KeyBindings)
 }
 
 func mergeProjectConfig(dst *FileConfig, src FileConfig) error {
@@ -277,6 +279,7 @@ func mergeProjectConfig(dst *FileConfig, src FileConfig) error {
 	if src.Swarm.MaxTeamSize != 0 {
 		dst.Swarm.MaxTeamSize = src.Swarm.MaxTeamSize
 	}
+	mergeKeyBindings(&dst.KeyBindings, src.KeyBindings)
 	// Local control is intentionally user-config/override only. A cloned project
 	// must not be able to make browser, desktop, or terminal automation tools
 	// appear in the model's tool surface.
@@ -665,6 +668,7 @@ func applyOverrides(cfg *FileConfig, overrides Overrides) {
 		cfg.Tools.deferThresholdSet = true
 	}
 	mergeLocalControlConfig(&cfg.LocalControl, overrides.LocalControl)
+	mergeKeyBindings(&cfg.KeyBindings, overrides.KeyBindings)
 	for _, provider := range overrides.Providers {
 		mergeProvider(cfg, provider)
 	}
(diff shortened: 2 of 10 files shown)


gnanam1990 (changes requested):
Nice feature, and the defaults are regression-safe — with no custom config every binding falls through to the identical hardcoded chord, and malformed config fails safe (no panics). Two things block merge:

**1. CI will fail on the gofmt gate.** `internal/tui/model.go`, `keybindings.go`, and `binding_test.go` aren't gofmt-clean (the CI "Check formatting" step runs `gofmt -l .` and exits non-zero on any output). Please run `gofmt -w`.

**2. CodeRabbit's change-request is valid and unaddressed.** The Ctrl+B handler still gates on `m.sidebarAvailable()` rather than the new `m.sidebarToggleAllowed()`, which is never called — dead code — so the toggle-while-auto-hidden feature this PR adds doesn't actually work.

Non-blocking correctness gaps for configured (non-default) keys: `Mod.Contains` matches supersets (a configured `alt+o` also fires on `alt+shift+o`; a plain-key binding fires while modifiers are held) — consider exact `==`; and single-char parsing keys on byte length, so a 2-byte UTF-8 key (e.g. `é`) won't bind. `builtinBindings` is also unused. Defaults are unaffected by all of these.

eli-l:
Thanks @gnanam1990, the superset case was an interesting one, my tests showed I managed to solve it. Requested changes have been addressed.

Vasanthdev2004 (approved):
## Approve — and the outstanding `CHANGES_REQUESTED` is now stale; every prior blocker is fixed in this HEAD

**Worth it:** yes. This makes keyboard shortcuts reconfigurable via the config file (closes #416, a filed request), and — the critical property — it's **regression-safe by construction**: with no custom config, every binding falls through to the identical hardcoded chord via the `defaultFn` fallback in `keyMatch`, and malformed config fails safe (no panics). Thanks @eli-l.

I checked this out on Windows and verified the current HEAD against every concern raised in the earlier reviews — **all of them are already addressed**:

- **gofmt gate** (was flagged red): `gofmt -l` on `model.go`, `keybindings.go`, `binding_test.go` is now **clean**.
- **Dead-code `sidebarToggleAllowed`** (was flagged): the Ctrl+B handler now actually calls `m.sidebarToggleAllowed()` (it replaced `m.sidebarAvailable()`), so the toggle-while-auto-hidden feature works.
- **Superset modifier matching** (`Mod.Contains`, was flagged): the matcher now uses exact `msg.Key().Mod == mod` (lines 126/132/137), so `alt+o` no longer fires on `alt+shift+o`.
- **UTF-8 single-char parse** (was flagged): parsing uses `[]rune(keyPart)[0]` (imports `unicode/utf8`), so multi-byte keys bind correctly.
- **Unused `builtinBindings`** (was flagged): gone — no references remain in the package.

**Verified on Windows** (worktree): `go build ./...` clean, `go vet ./internal/tui/` clean, and `go test ./internal/tui/ -run 'Binding|Keybinding|Sidebar'` passes. CodeRabbit also moved from CHANGES_REQUESTED to APPROVED.

So @gnanam1990's review — which correctly caught real issues at the time — is now **stale**; those fixes landed after it. It should be dismissed so this isn't blocked on resolved concerns.

**Notes (non-blocking):**
- It's a sizable feature (711 lines); since custom-binding correctness is the whole point, a quick manual smoke of one rebind (e.g. remap the sidebar toggle) before merge would be worthwhile.
- **Overlap with #462** (Shift+Enter): both touch the Chat binding line + `model.go` Enter block, so a rebase conflict is likely for whichever lands second; if this lands first, #462's Shift+Enter should be expressed through this registry.

CI note: fork PR, so the Go workflows are maintainer-approval-gated — only CodeRabbit ran; I validated build/vet/tests locally on Windows. Approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

Re-review of new commits since my prior review (`668bb626` → `72beeb06`). The new commits fix a modifier superset matching bug in `keybindings.go`: `Mod.Contains(mod)` was changed to `Mod == mod` (exact match), so a configured `ctrl+o` no longer fires on `ctrl+alt+o` or `ctrl+shift+o`. This is a correctness fix — the old `Contains` check was too permissive and could cause unintended binding activations. The `parseBinding` function also now uses `utf8.RuneCountInString` instead of `len()` for single-character detection, fixing multi-byte UTF-8 handling. 128 lines of new tests verify that superset modifiers are rejected for ctrl, alt, and plain-key bindings. The `sidebarToggleAllowed` field is now used consistently.

The behavior change (exact mod match vs. superset) could theoretically affect users who configured a binding expecting the superset behavior, but that behavior was a bug — no reasonable user config intends `ctrl+o` to fire on `ctrl+alt+o`. The default bindings are unaffected because they use the text-based fallback in `model.go`, not the `parsedBinding.Matcher()` path.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/tui/...` — pass
- `go test ./internal/tui/... -run 'TestCtrl|TestAlt|TestPlain|TestKeybinding|TestBinding|TestToggle'` — all pass, including new superset rejection tests
- CI: CodeRabbit passed

## FINDINGS

None. The superset matching fix is correct, well-tested, and the behavior change is a bug fix not a regression.

## BOTTOM LINE

The modifier superset matching fix and UTF-8 handling improvement address the prior review's correctness concern. Safe to merge.

kevincodex1 (approved):
LGTM

PR #418, opened 2026-07-02
docs(install): document Bun install and improve missing-binary error

AxDSan:
## Summary

Installing `@gitlawb/zero` with **Bun** silently skips the wrapper's `postinstall`, so the native binary is never fetched and the first run fails with:

```
[zero] No native binary found next to the npm wrapper. Reinstall the zero
package or run `go run ./cmd/zero-release build` from the repository.
```

That message only helps someone with a source checkout — a Bun user who just ran `bun add` is left stuck. This PR makes Bun a documented install path and makes the error actionable for everyone.

## Why this happens

Bun is "default-secure" and **does not run lifecycle scripts of installed dependencies** by default — only the installing project's own scripts. `@gitlawb/zero` isn't on Bun's built-in `trustedDependencies` list, so the `postinstall` in `scripts/postinstall.mjs` (which downloads the platform binary from GitHub Releases) is skipped with no warning. `npm`/`yarn` run it, so the documented `npm install -g @gitlawb/zero` flow works; only Bun users hit this.

Refs: [bun.sh/docs/cli/install](https://bun.sh/docs/cli/install), [bun.sh/docs/pm/lifecycle](https://bun.sh/docs/pm/lifecycle)

## Changes

- **`README.md`** — new `### Bun` subsection under **Install** with the two working install paths.
- **`docs/INSTALL.md`** — new `## Bun` section with the same, plus a link to Bun's lifecycle docs.
- **`bin/zero.js`** — rewrote the missing-binary error to:
  - keep the existing `No native binary found next to the npm wrapper` lead line (so the Go wrapper test at `internal/npmwrapper/npmwrapper_test.go:259` still passes),
  - tell the user to run `scripts/postinstall.mjs` manually (concrete fix that works for any package manager), and
  - when launched under Bun (`process.versions.bun` / `bun` in `execPath`), append a Bun-specific note pointing at `trustedDependencies`.
  - The source-build fallback now points at `go run ./cmd/zero` rather than `go run ./cmd/zero-release build` (the latter only makes sense inside a release checkout).

## Verification

- `node --check bin/zero.js` ✓
- Previewed the error under `node` (generic path) and under `bun` (Bun-specific hint appears) ✓
- `go test ./internal/npmwrapper/ ./internal/release/` ✓ (all pass, including `TestNodeWrapperReportsMissingNativeBinary`)

```
ok  github.com/Gitlawb/zero/internal/npmwrapper  0.544s
ok  github.com/Gitlawb/zero/internal/release     0.024s
```

## Notes / possible follow-ups

- A more permanent fix would be to ship the binary via per-platform `optionalDependencies` packages (esbuild/sharp style), which removes the reliance on a `postinstall` entirely and works under Bun without `trustedDependencies`. That's a bigger change and out of scope here; this PR is the minimal docs + error-message fix.
- Happy to trim/adjust wording if you'd prefer a shorter README note.

## Summary by CodeRabbit

* **Documentation**
  * Added Bun-specific installation guidance, including how to handle packages that skip lifecycle scripts by default.
  * Documented two setup options: run the post-install step manually or allow it via trusted dependencies.
  * Expanded install docs with notes for global installs and related lifecycle behavior.

* **Bug Fixes**
  * Improved the missing-native-binary error message with clearer, step-by-step recovery instructions.
  * Added clearer guidance for Bun users and a fallback path for rebuilding from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,27 @@ The npm package installs a small wrapper plus the matching Zero binary for your
 platform from GitHub Releases. It supports Linux, macOS, and Windows on x64 and
 arm64.
 
+### Bun
+
+Bun does not run dependency lifecycle scripts by default, so the `postinstall`
+that fetches the Zero binary is skipped and the first run fails with
+`No native binary found next to the npm wrapper`.
+
+Install with Bun, then fetch the binary in one of two ways:
+
+```bash
+# Option A: run the installer manually
+bun add @gitlawb/zero
+node node_modules/@gitlawb/zero/scripts/postinstall.mjs
+
+# Option B: allow the postinstall to run on install
+# add to your package.json:  "trustedDependencies": ["@gitlawb/zero"]
+bun add @gitlawb/zero
+```
+
+For global installs (`bun add -g @gitlawb/zero`), use Option A against the
+global install path, or prefer the install scripts below.
+
 ### Install scripts
 
 Linux/macOS:
--- a/bin/zero.js
+++ b/bin/zero.js
@@ -54,8 +54,26 @@ const nativePath = join(packageRoot, zeroBinaryName());
 const localControlHelpers = localControlHelperManifest(packageRoot);
 
 if (!existsSync(nativePath)) {
+  const postinstallScript = join(packageRoot, 'scripts', 'postinstall.mjs');
+  const ranByBun = process.execPath.includes('bun') || !!process.versions?.bun;
   console.error(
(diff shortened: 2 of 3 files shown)


kevincodex1 (approved):
thank you for this

AxDSan:
> thank you for this

Thank you for Zero and everything else! ✨ keep shipping big! 

- AJ

PR #420, opened 2026-07-03
fix(tui): render the ? help as an overlay, not a full-screen replace

### LINKED ISSUE #419: the help overlay hides the whole screen
### Version / branch / commit

0.1.0

### OS and environment

NixOS 26.05, kitty and ghostty, go 1.26

### Steps to reproduce

1. run zero
2. just click `?`
3. the help overlay will replace the entire tui view.

### Expected behavior

expected behavior is to be like the other popups like the model picker and suggestions picker etc.
examples:

Image

Image

#### which make it looks like that and hides the whole view 

Image

_No response_

gnanam1990:
Fixes #419.

## Problem

Pressing `?` replaced the **entire** TUI with the keyboard-shortcut block — the transcript, title bar, and composer all vanished. Every other popup (model picker, suggestions, wizards) composites *over* the chat instead, and the issue asks for `?` to behave the same way.

Root cause: `View()` short-circuited on `m.helpOverlay` to `renderKeybindingHelpOverlay`, which centers the block on a **blank canvas** via `centerRenderedBlock`, so nothing else rendered.

## Fix

Route the help overlay through the **same viewport-overlay path the pickers use**:

- Add `keybindingHelpOverlay(width)` — returns `""` when closed, else the centered block.
- Include it as the highest-priority case in `transcriptView`'s overlay switch **and** in `twoColumnTranscriptView` (so it also composites correctly with the sidebar open).
- Drop the content-replacement branch in `View()`.

The overlay now draws on top of the chat via the existing `overlayViewportLines` compositing, keeping the transcript / title bar / composer visible behind it. Input handling is unchanged — the modal key-swallow and `?`/Esc/`q`/Enter dismiss run before render — so only the presentation changes.

## Testing

- New tests: the overlay composites over the chat chrome (title bar + composer survive) and over a populated transcript row. **Red-green** — restoring the old full-replace makes both fail (chrome *and* body gone).
- Existing open/close/swallow/content tests unchanged and green; full `internal/tui` suite passes.
- Verified live: `?` on the home screen now shows the title bar, the centered shortcuts box as an overlay, footer hints, and the composer; Esc dismisses cleanly with the chat intact.

### Before → after
Before: `?` blanks the screen down to just the centered box.
After: the shortcuts box floats over the chat like the model picker, with the title bar above and the composer below still visible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Keyboard Shortcuts now appears as an overlay on top of the chat screen instead of replacing the entire view.
  * Existing chat content and transcript text remain visible behind the overlay, preserving the current conversation display.

### DIFF
--- a/internal/tui/keybinding_help.go
+++ b/internal/tui/keybinding_help.go
@@ -133,6 +133,18 @@ func (m model) renderKeybindingHelpOverlay(width int, height int) string {
 	return centerRenderedBlock(block, width)
 }
 
+// keybindingHelpOverlay returns the `?` shortcut overlay as a centered block to
+// COMPOSITE over the transcript — the same viewport-overlay path the model
+// picker and suggestions use — or "" when it is not open. Rendering it this way
+// (rather than replacing the whole View) keeps the chat visible behind it, so
+// `?` behaves like every other popup instead of blanking the screen (#419).
+func (m model) keybindingHelpOverlay(width int) string {
+	if !m.helpOverlay {
+		return ""
+	}
+	return m.renderKeybindingHelpOverlay(width, m.height)
+}
+
 // keybindingHelpOverlayWidth picks the overlay width: wide enough that the
 // descriptions don't truncate next to the key column, capped, and never wider
 // than the terminal.
--- a/internal/tui/keybinding_help_test.go
+++ b/internal/tui/keybinding_help_test.go
@@ -105,3 +105,56 @@ func TestKeybindingGroupsAreWellFormed(t *testing.T) {
 		}
 	}
 }
+
+// #419: the `?` help overlay must render ON TOP of the chat (like the model
+// picker), not REPLACE the whole screen. The old full-screen replace produced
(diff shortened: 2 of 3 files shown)


gnanam1990:
Closing in favor of #421 by @N1xev — the reporter's own fix.

Both PRs converged on the same core approach (route the help through the `viewportOverlay` compositing path the pickers/wizards already use, instead of replacing the whole `View()`), but #421 is the more idiomatic version:

- It adds `helpOverlay` to `sidebarAvailable()`, so the overlay suppresses the sidebar and renders full-width **exactly like the model picker, provider wizard, and suggestions popups** — which is precisely the behavior #419 asked for ("like the other popups").
- It drops the now-dead `height` parameter from `renderKeybindingHelpOverlay`.

Thanks @N1xev for diagnosing the layering root cause and fixing it. The one thing this PR carried that #421 doesn't is a regression test for the composite-not-replace behavior — I've left those two tests as a suggestion on #421 so the fix ships guarded.

PR #421, opened 2026-07-03
fix(tui): compose help overlay through the viewport overlay pipeline

### LINKED ISSUE #419: the help overlay hides the whole screen

sam0uly:
exactly like how other popups works.

# before:
image

# after:
image

## Linked issue

Fixes #419 

## Checklist

- [ ] `gofmt` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Integrated the keyboard shortcuts help overlay into the main transcript overlay flow so it layers over the chat UI instead of replacing it.
  * Improved overlay layout: the right-side context sidebar is hidden while the help overlay is active, and overlay vertical centering is handled by the overall screen composition.
* **Tests**
  * Added tests to verify the help overlay composites on top of existing chat elements and that transcript content still peeks through behind the centered overlay.


kevincodex1 (approved):
nice one! thank you so much

kevincodex1:
kindly please check the failing test

sam0uly:
> kindly please check the failing test

done! it was the formatting CI/CD step

gnanam1990:
Thanks for this, @N1xev — and for the precise root-cause writeup in the issue. Routing the overlay through the same `viewportOverlay` path as the other popups, and suppressing the sidebar via `sidebarAvailable()` so it renders full-width like the model picker, is exactly the right shape for this. I've closed my own PR (#420) in favor of this one.

The one thing #420 carried that this doesn't is a regression test that pins the fix. The existing help tests cover close/swallow/content, but none of them assert that the chat survives *behind* the overlay — so a future refactor could silently bring back the full-screen blank without any test going red.

Here are two tests that cover exactly that. I checked out this branch and ran them against your implementation — both pass (`go test ./internal/tui/`), so they should drop straight into `internal/tui/keybinding_help_test.go`:

<details>
<summary>Two regression tests (verified against this branch)</summary>

```go
// #419: the `?` help overlay must render ON TOP of the chat (like the model
// picker), not REPLACE the whole screen. The old full-screen replace produced
// only the centered shortcut block on a blank canvas — no title bar, no
// composer. With the overlay open, surrounding chat chrome that is NOT covered
// by the centered box (the model title bar at top, the composer at bottom) must
// still be present alongside "Keyboard Shortcuts".
func TestHelpOverlayCompositesOverChatNotReplacingIt(t *testing.T) {
	m := newModel(context.Background(), Options{ModelName: "gpt-4o"})
	m.width = 100
	m.height = 40
	m.altScreen = true

	base := plainRender(t, m.View()) // no overlay: baseline chrome
	m.helpOverlay = true
	over := plainRender(t, m.View())

	if !strings.Contains(over, "Keyboard Shortcuts") {
		t.Fatalf("help overlay not rendered:\n%s", over)
	}
	// Chrome that renders in the baseline (and sits outside the centered overlay
	// box) must survive behind the overlay. The full-screen replace showed none.
	for _, marker := range []string{"gpt-4o", "describe a task"} {
		if !strings.Contains(base, marker) {
			t.Fatalf("precondition: baseline chat should contain %q:\n%s", marker, base)
		}
		if !strings.Contains(over, marker) {
			t.Fatalf("#419: help replaced the chat instead of overlaying it; %q is gone:\n%s", marker, over)
		}
	}
}

// A populated transcript row also survives behind the overlay (peeking out to
// the left of the centered box), proving the chat body — not just the chrome —
// is composited under the overlay rather than discarded.
func TestHelpOverlayKeepsTranscriptBodyBehindIt(t *testing.T) {
	m := newModel(context.Background(), Options{ModelName: "gpt-4o"})
	m.width = 120
	m.height = 40
	m.altScreen = true
	m.transcript = appendTranscriptRow(m.transcript, transcriptRow{kind: rowUser, text: "hello there this is a chat line"})
	m.helpOverlay = true

	view := plainRender(t, m.View())
	if !strings.Contains(view, "Keyboard Shortcuts") {
		t.Fatalf("help overlay not rendered:\n%s", view)
	}
	// The start of the transcript line peeks to the left of the centered box.
	if !strings.Contains(view, "hello") {
		t.Fatalf("#419: transcript body was replaced by the help overlay:\n%s", view)
	}
}
```

</details>

The first asserts the title bar + composer survive alongside "Keyboard Shortcuts"; the second that a transcript line peeks out beside the centered box. Both fail against the old full-screen replace and pass with your fix. Looks good to me once something like this is in.

sam0uly:
> Thanks for this, @N1xev — and for the precise root-cause writeup in the issue. Routing the overlay through the same `viewportOverlay` path as the other popups, and suppressing the sidebar via `sidebarAvailable()` so it renders full-width like the model picker, is exactly the right shape for this. I've closed my own PR (#420) in favor of this one.
> 
> The one thing #420 carried that this doesn't is a regression test that pins the fix. The existing help tests cover close/swallow/content, but none of them assert that the chat survives _behind_ the overlay — so a future refactor could silently bring back the full-screen blank without any test going red.
> 
> Here are two tests that cover exactly that. I checked out this branch and ran them against your implementation — both pass (`go test ./internal/tui/`), so they should drop straight into `internal/tui/keybinding_help_test.go`:
> Two regression tests (verified against this branch)
> 
> The first asserts the title bar + composer survive alongside "Keyboard Shortcuts"; the second that a transcript line peeks out beside the centered box. Both fail against the old full-screen replace and pass with your fix. Looks good to me once something like this is in.

thanks! i've added the regression tests and pushed them.

gnanam1990 (approved):
Approving — this is the right fix for #419. Routing the overlay through the same `viewportOverlay` path as the pickers/wizards, and suppressing the sidebar via `sidebarAvailable()` so it renders full-width, is exactly the "like the other popups" behavior the issue asked for.

The only ask before merge is the regression test I left in the comment above, so the composite-not-replace behavior stays guarded against a future refactor. Thanks for the clean root-cause diagnosis and the fix.

sam0uly:
> Approving — this is the right fix for #419. Routing the overlay through the same `viewportOverlay` path as the pickers/wizards, and suppressing the sidebar via `sidebarAvailable()` so it renders full-width, is exactly the "like the other popups" behavior the issue asked for.
> 
> The only ask before merge is the regression test I left in the comment above, so the composite-not-replace behavior stays guarded against a future refactor. Thanks for the clean root-cause diagnosis and the fix.

i added them and pushed, check them.

PR #423, opened 2026-07-03
feat: add --auto flag for LLM-generated commit messages

KunjShah95:
This PR implements LLM-generated commit message capabilities under zero changes commit via the --auto / -a flag. It analyzes local diffs, constructs a conventional commit prompt, calls the active configured provider, and creates the commit.

## Summary by CodeRabbit

* **New Features**
  * Added `--auto` / `-a` to `changes commit` to generate commit messages via an LLM-backed workflow, using the inspected changes (with sensitive content redacted).
  * Updated command help to document the new flag.
* **Bug Fixes**
  * Improved CLI validation: `--auto` is only allowed for `commit`, and is mutually exclusive with `--message` (including when `--message` is empty).
  * Auto-generated output is normalized (e.g., code-fence trimming) and fails clearly on empty/whitespace-only responses.
* **Tests**
  * Added coverage for `--auto` parsing/validation and commit message generation success/failure paths.

### DIFF
--- a/internal/cli/workflow_test.go
+++ b/internal/cli/workflow_test.go
@@ -694,7 +694,7 @@ func TestParseChangesArgsAuto(t *testing.T) {
 	}
 
 	// 4. Reject both empty message and --auto on commit
-	_, _, err = parseChangesArgs([]string{"-m", "", "--auto"}, "commit")
+	_, _, err = parseChangesArgs([]string{"--message=", "--auto"}, "commit")
 	if err == nil || !strings.Contains(err.Error(), "cannot specify both --message and --auto") {
 		t.Fatalf("expected empty --message and --auto conflict error, got %v", err)
 	}
@@ -771,7 +771,7 @@ func TestRunChangesCommitAuto(t *testing.T) {
 	if strings.Contains(promptContent, "ghp_SECRETKEYHERE") {
 		t.Fatal("expected secret in diff to be redacted, but it was found in the prompt")
 	}
-	if !strings.Contains(promptContent, "[redacted]") && !strings.Contains(promptContent, "redacted") {
+	if !strings.Contains(promptContent, "[REDACTED]") && !strings.Contains(promptContent, "REDACTED") {
 		t.Fatalf("expected redacted diff content in prompt, got: %q", promptContent)
 	}
 	if !strings.Contains(stdout.String(), "Generating commit message using LLM...") {
--- a/internal/cli/workflow_test.go
+++ b/internal/cli/workflow_test.go
@@ -692,6 +692,12 @@ func TestParseChangesArgsAuto(t *testing.T) {
 	if err == nil || !strings.Contains(err.Error(), "cannot specify both --message and --auto") {
 		t.Fatalf("expected --message and --auto conflict error, got %v", err)
 	}
+
+	// 4. Reject both empty message and --auto on commit
+	_, _, err = parseChangesArgs([]string{"-m", "", "--auto"}, "commit")
+	if err == nil || !strings.Contains(err.Error(), "cannot specify both --message and --auto") {
+		t.Fatalf("expected empty --message and --auto conflict error, got %v", err)
+	}
 }
 
 type mockCommitMsgProvider struct {
@@ -714,7 +720,7 @@ func TestRunChangesCommitAuto(t *testing.T) {
 		Root:   cwd,
 		Branch: "main",
 		Files:  []zerogit.FileChange{{Path: "README.md", Status: "modified"}},
-		Diff:   "some diff content",
+		Diff:   "some diff content with ghp_SECRETKEYHERE",
 	}
 
 	mockProv := &mockCommitMsgProvider{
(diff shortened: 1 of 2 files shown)


kevincodex1:
please address coderabbit comments

KunjShah95:
Sure working on it

KunjShah95:
@kevincodex1  do check it now and share your thoughts on which use case needs to be worked on if its failing then

kevincodex1 (approved):
LGTM

gnanam1990 (changes requested):
The production logic is sound and low-risk (fully gated behind `--auto`; the default path is unchanged), and it addresses all three of CodeRabbit's findings — nice work. But the fix commit ships a **red test suite**, which blocks merge:

- `TestRunChangesCommitAuto` (`workflow_test.go:775`) asserts the prompt contains lowercase `"redacted"`, but redaction correctly emits `[REDACTED]` (uppercase) and `strings.Contains` is case-sensitive → fail. The behavior is right; fix the assertion.
- `TestParseChangesArgsAuto` case 4 (`workflow_test.go:699`) expects the `--message`+`--auto` mutual-exclusion error, but `-m ""` trips `-m requires a value` first, so that error is never reached. Either accept that error or use the `--message=` inline-empty form that actually reaches the check.

Both are test-bugs and trivial, but please get `go test ./internal/cli/` green before merge.

Two non-blocking notes: `--auto` commits without showing you the generated message first (`--dry-run` is the preview escape hatch — maybe surface it in help), and the LLM call uses `context.Background()` with no timeout, so a hung provider hangs the command.

Vasanthdev2004 (approved):
## Verdict: Approve ✅ — well-built, security-conscious, and properly tested; a couple of tiny non-blocking nits

I reviewed this as a change that sends your working diff to an LLM provider — i.e. I went looking for a secret-leak or an unbounded-prompt problem first. It's clean on both counts. Checked it out, built it, `gofmt`/`go vet` clean, tests pass, and all of CI is green.

### What I verified
- **Secrets don't leak into the prompt.** The diff is redacted *twice* before it reaches the provider: `zerogit.Inspect` already runs `redactText` on the diff (zerogit.go:104/128), and the auto path redacts again via the existing `redactChangeSummary` (which covers `summary.Diff`). Your test asserting `ghp_…` is stripped and `[REDACTED]` is present confirms it end-to-end. 👍
- **The prompt is bounded.** `Inspect` caps the diff at `defaultMaxDiffBytes` (120 KB) even when `--diff-bytes` is unset, so a huge working tree can't blow up the request. Reasonable.
- **Validation is thorough and correct.** `--auto` is rejected on `inspect`, and it's mutually exclusive with `--message` — including the empty `--message=` edge, which is exactly why the new `hasMessage` bool is the right call (a plain `options.message != ""` check would have missed it). All four cases are tested.
- **Good failure handling.** Clean tree → "no changes to commit"; no provider configured → clear error; empty/whitespace LLM output → "empty commit message" error (tested); markdown code-fence wrapping is stripped.
- Reuses existing infra throughout (`inspectChanges`, `resolveConfig`, `newProvider`, `CollectStream`, `redactChangeSummary`) rather than reinventing it.

### Nits (none blocking)
1. **`model` param is unused.** `generateAutoCommitMessage(ctx, provider, model string, summary)` never references `model` — `zeroruntime.CompletionRequest` has no `Model` field, so the model is entirely determined by `newProvider(resolved.Provider)`. It works correctly, but the dead parameter is misleading; I'd drop it (or add a short comment if you're reserving it).
2. **The generated message is committed before the user sees it.** Without `--dry-run`, the flow generates the message and immediately commits — the message only appears in the *post-commit* output. It's low-risk (git commits are local and `--amend`-able), and `--auto --dry-run` already gives a proper preview path — so I'd just surface that: print the generated message before committing, and/or mention "`--auto --dry-run` to preview" in the help text.
3. **Optional:** nothing validates that the LLM actually honored the ≤72-char/conventional-commit instruction. Best-effort is fine for a convenience flag; just flagging that a non-conforming message would commit as-is.

None of these affect correctness or safety, so I'm approving. If you want to knock out #1 and #2 that'd be a nice polish, but it's your call. Nice work @KunjShah95 — the redaction-before-send and the empty-response guard are exactly the details I'd want to see in this feature.

KunjShah95:
do check it

Vasanthdev2004 (approved):
## Re-review after the post-approval commit (7d22b45) — still approve ✅; @gnanam1990's blocking concern is resolved

Re-checked because two things happened after my approval: a new commit landed and @gnanam1990 requested changes over a red test suite. Good news on both.

**@gnanam1990's blocking concern (red `internal/cli` tests) is resolved.** I checked out HEAD (`7d22b45`) and ran the two tests they called out — both pass:
```
--- PASS: TestParseChangesArgsAuto
--- PASS: TestRunChangesCommitAuto (+ /EmptyLLMResponse)
```
(The `[REDACTED]` case-sensitivity and the `-m ""` "requires a value" ordering were both fixed in `2b8fb48`.) CI's ubuntu/macOS/windows smokes — which run `go test ./...` — are all green on this HEAD too. So the merge-blocker is cleared; that CHANGES_REQUESTED can be dismissed.

**The new commit `7d22b45` is sound** and knocks out two of my earlier nits + @gnanam1990's non-blocking notes:
- 60s timeout on the LLM call via `context.WithTimeout(context.Background(), 60*time.Second)` with `defer cancel()` correctly scoped inside the `--auto` block — a hung provider no longer hangs the command.
- `--dry-run` is now surfaced in the `--auto` help line as the preview escape hatch.

Nice, minimal changes. The only remaining item from my original review is the unused `model` param in `generateAutoCommitMessage` (`CompletionRequest` has no `Model` field; the model rides the provider) — cosmetic, non-blocking.

Re-affirming approval. Thanks @KunjShah95 — timeout + preview hint are exactly right.

KunjShah95:
thanks buddy i  am happy it satifies your evaluations parameters

KunjShah95:
would love if there was any scope of becoming a reivewer or as a collaborator on the project in future if u need someone then can u count me in

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

Re-review of new commits since my prior review (`2b8fb482` → `7d22b457`). The new commits address review feedback: add a 60-second timeout for the LLM call in `--auto` commit message generation (using `context.WithTimeout`), and surface `--dry-run` in the `--auto` help text. The timeout prevents an unbounded hang if the LLM provider is slow or unresponsive. The help text clarification lets users know they can preview the auto-generated message.

## BUILD / TEST

- The 5-line change in `workflows.go` is minimal and correct: `context.WithTimeout(context.Background(), 60*time.Second)` with `defer cancel()`

## FINDINGS

None. The timeout and help text improvements address the prior review's concerns.

## BOTTOM LINE

The 60s LLM timeout and --dry-run help text address the prior review's feedback. Safe to merge.

kevincodex1 (approved):
LGTM

PR #426, opened 2026-07-03
fix(sandbox): gate /tmp test assertions on GOOS, not path existence

euxaristia:
Update after rebase: main picked up the same functional guard fix via #378 while this PR was open, so this now only adds the explanatory comment for why the guard needs the GOOS check. Merge for the comment or close as superseded, either is fine.

## What

Adds a comment to the /tmp assertions in `TestPermissionProfileFromPolicyIncludesDefaultTempWriteRoots` and `TestNewScopeNormalizesAndValidatesExtraRoots` explaining why `runtime.GOOS != "windows"` is required in addition to `pathExists("/tmp")`: Go resolves the bare path `/tmp` against the current drive on Windows, so a stray `C:\tmp` used to satisfy the old guard while the product code (correctly, see `defaultTempWriteRootCandidatesForGOOS` in `internal/sandbox/scope.go`) never adds `/tmp` as a Windows default temp root. The comment keeps the check from being "simplified" away later.

## Testing

Full `internal/sandbox` suite passes on Windows 11 after the rebase. POSIX behavior unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #427, opened 2026-07-03
feat(sandbox): unelevated Windows fallback tier instead of prompts-only degrade

euxaristia:
> Note on process: I know CONTRIBUTING asks for an approved issue first, happy to convert this to an issue and resubmit if you prefer. Opening as a PR directly because the full working, Windows-verified implementation exists and is easier to evaluate than a proposal. Related (but separate, scoped) PR: #426, the `/tmp` test-guard fix, which makes the sandbox suite green on Windows so this branch can be validated there.

## What

Adds an `unelevated` enforcement tier to the Windows sandbox: when the elevated `zero sandbox setup` has not run, commands with a restricted-filesystem profile still wrap through the command runner under the write-restricted token, with the runner applying the workspace ACL plan itself (no Administrator rights needed). Previously this state dropped to the in-process policy gate with no OS enforcement.

## Why

Only the WFP network filters actually require elevation. The write-restricted token (`CreateRestrictedToken(WRITE_RESTRICTED)`) and the workspace ACL grants (DACL edits on user-owned roots) work from a normal process, so the filesystem write-jail can and should hold on fresh installs and non-admin machines. Same elevated/unelevated split Codex ships on Windows.

## How

- `EnforcementUnelevated` between `native` and `degraded` (`types.go`); the manager selects it when the elevated marker is missing, the preference is `auto`, and the profile restricts the filesystem (`manager.go`). Network-only profiles keep the old degrade; `--sandbox require` still errors with the setup hint.
- New `WindowsSandboxLevelUnelevated` runner level (`windows_runner.go`): same restricted-token launch path, but instead of validating the elevated setup marker, the runner applies the ACL plan via the existing `applyWindowsACLPlan` and records it (keyed by `WindowsACLPlanHash`) in `windows-unelevated-setup.json` so repeat commands skip the apply (`windows_command_runner_windows.go`, `windows_unelevated.go`). Grants are left in place like the elevated setup's; they only name synthetic capability SIDs no other token carries.
- The downgrade reason at this tier states explicitly that the write-jail is enforced while network isolation still relies on per-command approval.

## Testing

- `go test ./internal/sandbox/`: green (two pre-existing failures on Windows hosts, `TestPermissionProfileFromPolicyIncludesDefaultTempWriteRoots` and `TestNewScopeNormalizesAndValidatesExtraRoots`, fail identically on `main`; they assume POSIX `/tmp`).
- New portable unit tests: manager tier selection, args round-trip, marker record/dedupe/eviction/schema-reset.
- New `TestWindowsUnelevatedRealSandboxSmoke` (gated by `ZERO_SANDBOX_REAL_SMOKE=1`), **verified on real Windows 11 without elevation**: write inside the workspace succeeds under the restricted token; write outside every granted root is denied by the OS; marker recorded. Uses `cmd.exe` because managed PowerShell cannot initialize its crypto provider under a write-restricted token on some hosts (0x8009001d, same restricted-token family as the Schannel limitation already documented in the runner).
- `go build ./...`, `go vet ./internal/sandbox/...`, gofmt clean.

## Limitations (by design, stated in the downgrade reason)

- No network enforcement at this tier: WFP filters require the elevated setup; network stays behind the approval gate.
- ACL apply fails closed on filesystems the user does not own (e.g. some network shares), with an actionable error pointing at the elevated setup or `--sandbox forbid`.

## Summary by CodeRabbit

* **New Features**
  * Added a new Windows sandbox mode for running with limited privileges when full setup isn't available.
  * Expanded command handling so more Windows sandbox requests can still run with appropriate restrictions.
  * Added support for remembering applied workspace permissions to avoid repeating setup work.

* **Bug Fixes**
  * Improved fallback behavior on Windows when sandbox initialization is missing.
  * Added clearer guidance when sandbox permissions can't be applied.
  * Strengthened Windows sandbox behavior with additional validation and smoke coverage.

### DIFF
--- a/internal/sandbox/manager.go
+++ b/internal/sandbox/manager.go
@@ -19,6 +19,13 @@ var errWindowsSandboxNotInitialized = errors.New(
 const windowsSetupDowngradeReason = "Windows OS sandbox inactive — commands run with workspace path-confinement and " +
 	"per-command approval only. Run `zero sandbox setup` from an elevated (Administrator) terminal for full filesystem and network isolation."
 
+// windowsUnelevatedDowngradeReason is surfaced when Windows runs the sandbox in
+// the unelevated tier: the restricted-token write-jail is enforced (the command
+// runner applies the workspace ACLs itself, which needs no Administrator
+// rights), but the WFP network filters are Administrator-only and absent.
+const windowsUnelevatedDowngradeReason = "Windows sandbox running unelevated — the filesystem write-jail is enforced, but network " +
+	"isolation still relies on per-command approval. Run `zero sandbox setup` from an elevated (Administrator) terminal for OS-level network enforcement."
+
 // windowsSandboxInitialized reports whether the per-host Windows sandbox setup
 // marker exists. A missing marker is the common fresh-install state, so the
 // caller degrades rather than bricking the command. Indirected through a var so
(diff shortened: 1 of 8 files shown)


kevincodex1 (approved):
lgtm

anandh8x (approved):
Solid design and execution. The insight that only WFP filters need elevation — filesystem ACLs and restricted tokens work unelevated — is correct, and the implementation handles the edge cases well: network-only profiles still degrade, `--sandbox require` still errors, marker dedup by hash avoids redundant applies, and eviction caps prevent unbounded growth.

One minor thing: a corrupt marker JSON returns a hard error where the schema-mismatch path self-heals. Consistent self-healing would be better, but corrupt markers are an anomaly rather than a routine state — not blocking.

LGTM.

euxaristia:
The corrupt-marker self-heal nitpick from coderabbitai and @anandh8x is addressed in follow-up #437.

PR #437, opened 2026-07-03
fix(sandbox): self-heal a corrupt unelevated setup marker

euxaristia:
Follow-up addressing the review feedback from coderabbitai and @anandh8x on #427 (merged): a malformed or truncated `windows-unelevated-setup.json` returned a hard error from `loadWindowsUnelevatedSetupMarker`, bricking every unelevated command until the file was hand-deleted, while the unknown-schema path already self-healed by resetting to an empty marker.

## What

Treat corrupt JSON exactly like an unknown schema: reset to an empty current-schema marker so the runner re-applies the ACL plan and rewrites the file on the next command.

## Why it's safe

The marker only memoizes idempotent ACL applies. Resetting it can only cost redundant work (one extra apply); it can never skip enforcement, so failing open on the memoization is the fail-safe direction.

## Testing

- New `TestLoadWindowsUnelevatedSetupMarkerSelfHealsCorruptFile`: a truncated marker body loads as an empty marker without error, and a subsequent record over the corrupt file succeeds.
- `go test ./internal/sandbox/` green on Windows 11, `go vet` clean.

### DIFF
--- a/internal/sandbox/windows_unelevated.go
+++ b/internal/sandbox/windows_unelevated.go
@@ -61,8 +61,10 @@ func buildWindowsUnelevatedAppliedPlan(config WindowsSandboxCommandConfig) (Wind
 
 // loadWindowsUnelevatedSetupMarker reads the marker; a missing file is the
 // common first-run state and returns an empty marker, not an error. A marker
-// with an unexpected schema version is treated as empty so the runner re-applies
-// (and rewrites) rather than failing.
+// with an unexpected schema version or a corrupt/truncated body is likewise
+// treated as empty so the runner re-applies (and rewrites) rather than failing:
+// the marker only memoizes idempotent ACL applies, so resetting it can only
+// cost redundant work, never skip enforcement.
 func loadWindowsUnelevatedSetupMarker(sandboxHome string) (WindowsUnelevatedSetupMarker, error) {
 	sandboxHome = strings.TrimSpace(sandboxHome)
 	if sandboxHome == "" {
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
lgtm . thanks!

Vasanthdev2004 (approved):
## Approve — solid robustness fix, correct and safe

**Worth it:** yes. Before this, `loadWindowsUnelevatedSetupMarker` hard-errored on a corrupt or truncated `windows-unelevated-setup.json`, which bricked **every** unelevated sandbox command until the file was hand-deleted — a realistic failure mode (crash / power-loss mid-write, disk issue). The fix makes a corrupt body self-heal exactly like the already-handled unknown-schema case: reset to an empty current-schema marker and let the runner re-apply and rewrite. Good follow-up to the #427 review feedback. Thanks @euxaristia.

**Why it's correct and safe** (checked each):
- **No enforcement skip.** The marker only memoizes *idempotent* ACL applies. Resetting forces a redundant re-apply next command — the fail-safe direction. It can never cause an ACL grant to be skipped.
- **No torn-write race.** The healed marker is written through the existing temp-file-then-`os.Rename` atomic path, so readers never see a partial write.
- **Doesn't swallow real errors.** Only `json.Unmarshal` failures convert to a reset; genuine read/permission/IO errors still propagate. And since Go ignores unknown JSON fields, a forward-compatible future schema still parses and routes through the separate `SchemaVersion` check rather than this path.

**Verified on Windows 11:** `go build ./internal/sandbox/` + `go vet` clean; `go test ./internal/sandbox/` green including the new `TestLoadWindowsUnelevatedSetupMarkerSelfHealsCorruptFile` (covers both halves: truncated body loads as empty without error, and a subsequent `record` over the corrupt file succeeds). No hardcoded-string test break. CI 7/7 green including `Smoke (windows-latest)`.

Tight fix with a clear rationale in the doc comment. Approving.

PR #438, opened 2026-07-03
fix(docs): rename AGENTS.MD > AGENTS.md

lfaoro:
the AGENTS.MD file had its extensions .MD erroneously capitalized, this PR fixes that by lower-casing the extension.

## Summary by CodeRabbit

* **Documentation**
  * Corrected references to the project guidance file name so examples consistently use `AGENTS.md`.
  * Clarified wording about case-sensitive vs case-insensitive filename handling on Linux/CI/WSL and updated related comments for consistency.
  * No build/test/lint/runtime behavior was changed.


kevincodex1:
nice catch! kindly put a pr description please so it will not be empty

kevincodex1 (approved):

lfaoro:
> nice catch! kindly put a pr description please so it will not be empty

thank you, done.

Vasanthdev2004 (changes requested):
## Request changes — the rename direction is fine, but it ships a self-contradictory doc

Thanks @lfaoro. Moving to the lowercase, conventional `AGENTS.md` is reasonable, and the loader (`findProjectContextFile`) is case-insensitive so nothing functionally breaks. But this is a *pure* rename, and it leaves two prose references that now contradict the new filename — one of them **inside the renamed file itself**:

- **`AGENTS.md:29`** (the file this PR renames) still reads: *"The git-tracked filename in this repo is `AGENTS.MD` (uppercase `MD`) — keep that on case-sensitive filesystems … to match what the loader looks for."* After the rename, a file **named** `AGENTS.md` tells the reader its own name is `AGENTS.MD` and to prefer uppercase. That's self-contradictory and actively misinstructs.
- **`Makefile:1`**: `# … AGENTS.MD says "Build with make" …` now points at a filename that no longer exists.

The uppercase name wasn't an accident — line 29 was a deliberate (if technically dubious) documented choice, so reversing it means updating the rationale you're reversing.

**To merge, fold these into the same PR:**
1. Update `AGENTS.md:29` to describe the lowercase `AGENTS.md` convention. (The "keep uppercase on case-sensitive filesystems" advice is itself unnecessary — the loader enumerates directory entries and compares lowercased names, so `AGENTS.md` resolves on Linux too.)
2. Update the `Makefile:1` comment to `AGENTS.md`.
3. *(Optional)* reword the `internal/agent/system_prompt.go:313` comment ("a git-tracked file like AGENTS.MD"); the case-insensitive fallback **code** should stay — it's still valuable for user repos that name their file `AGENTS.MD`, and it's independently tested.

Tests are unaffected (`system_prompt_test.go` fabricates its own `AGENTS.MD` in temp dirs, independent of the repo file) and CI is green. Just make the file stop contradicting its own name and this is a clean merge.

kevincodex1:
i think we are good now

lfaoro:
thank you @Vasanthdev2004. you write so well I thought you're an automated agent :)

the references you suggested are updated, also ran ripgrep and found more references which are now reworded for AGENTS.md.

the `system_prompt_test.go` checks are unchanged

```go
√ user internal/agent main > go test -run TestBuildSystemPromptInjectsProjectGuidelinesCaseInsensitive .
ok  	github.com/Gitlawb/zero/internal/agent	0.004s
```

PR #441, opened 2026-07-03
fix(tui): keep the profile name on /model switch so the stored key resolves

### LINKED ISSUE #440: C:/Program Files/Git/model switch drops the stored credential (401) when the profile name differs from the catalog id
## Bug description

Switching models with `/model <model>` (or switching provider+model through the picker) fails with an auth/credential error for any provider whose **saved profile name differs from its catalog ID** — e.g. you set up OpenAI but named the profile `my-openai`. The provider was working a moment earlier; the switch itself is what breaks it.

## Root cause

`normalizeProfileForProvider` (`internal/tui/picker.go`) rewrites the profile identity when `normalizeIdentity` is true:

```go
normalizeIdentity := profileMatchesProviderBaseURL(profile, provider) || ...
if normalizeIdentity {
    profile.Name = provider.ID   // <-- clobbers the user's profile name
    profile.CatalogID = provider.ID
}
```

`profileMatchesProviderBaseURL` is true whenever the saved `BaseURL` equals the catalog default — which is the case for essentially every first-party provider (openai/anthropic/google/…). So on `/model` switch, a profile named `my-openai` has its `Name` overwritten to `openai`.

The encrypted credential store is keyed by **`profile.Name`** (`SecureProviderProfile` does `store.Set(Name, key)`; `ApplyStoredAPIKey` does `store.Get(Name)`). `normalizeProfileForProvider` runs *before* `profileWithCredential`, so the credential is then looked up under the rewritten name `openai` instead of `my-openai` → miss → empty `APIKey` → the rebuilt provider sends no key → 401.

Rewriting `CatalogID` is harmless (it's not the store key). Rewriting `Name` is the bug.

## Reproduction

Confirmed with a focused unit repro against `normalizeProfileForProvider`:

- profile `{Name: "my-openai", CatalogID: "openai", BaseURL: <openai catalog default>, APIKeyStored: true}`
- `normalizeProfileForProvider(openai)` returns `Name: "openai"` (rewritten from `my-openai`).
- The credstore key was `my-openai`, so `store.Get("openai")` misses and the profile comes back with no `APIKey`.

End-user steps:
1. Configure a provider with a profile name different from the catalog id (e.g. name it `my-openai`), authenticate so the key is saved to the credential store.
2. In the TUI, run `/model <some-model>`.
3. The switch reports a credential/auth error even though the same provider was working before the switch.

## Expected behavior

`/model` switching must preserve the profile's real `Name` so the credential store resolves the saved key. Switching the model on an already-authenticated provider should never drop its credential.

## Fix direction

In `normalizeProfileForProvider`, only set `profile.Name` when it is actually empty (or a generic `custom-*` placeholder) — never overwrite a legitimate user-provided name. `CatalogID` normalization can stay. A regression test should cover a custom-named first-party profile surviving a `/model` switch with its stored credential intact.

## Notes

- Distinct from #409 (`no active provider forces full re-onboarding`), which is a *startup*-resolution bug; #410 fixes that and does not touch this path.
- No fix PR is currently open for this.

Vasanthdev2004:
## Summary

`/model` switching threw a credential/auth error (401) for any provider whose saved profile name differed from its catalog id — e.g. a provider you set up as OpenAI but named `my-openai`. The provider worked before the switch; the switch itself dropped its credential.

## Root cause

`normalizeProfileForProvider` (`internal/tui/picker.go`) rewrote the profile identity when `normalizeIdentity` was true:

```go
if normalizeIdentity {
    profile.Name = provider.ID   // clobbers a real user name
    profile.CatalogID = provider.ID
}
```

`normalizeIdentity` is true whenever the saved `BaseURL` matches the catalog default — i.e. for essentially every first-party provider. The credential store is keyed by **`profile.Name`** (`SecureProviderProfile` → `store.Set(Name, key)`; `ApplyStoredAPIKey` → `store.Get(Name)`), and this normalization runs *before* `profileWithCredential`. So the key saved under `my-openai` was looked up under `openai`, missed, and the rebuilt provider went out with no key → 401.

## Fix

Only canonicalize an empty or generic `custom-*` placeholder name; never overwrite a real user-provided name. `CatalogID` canonicalization is unchanged (it is not the store key), so descriptor/metadata resolution is unaffected.

## Tests

- `TestNormalizeProfileForProviderPreservesCustomName` — a custom-named first-party profile (`my-openai`, baseURL at catalog default) keeps its `Name` through normalization while `CatalogID` is still canonicalized.
- `TestNormalizeProfileForProviderCanonicalizesPlaceholderName` — empty and generic placeholder names still canonicalize to the catalog id.
- Full `internal/tui` suite passes; `go build ./...`, `go vet ./...`, `gofmt` clean.

Fixes #440

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved custom profile names when selecting a provider, instead of replacing them with a generic provider name.
  * Continued to normalize placeholder or empty names to the expected provider name for consistent display.
* **Tests**
  * Added coverage for both custom-named and placeholder provider profiles to confirm the updated naming behavior.

### DIFF
--- a/internal/tui/picker.go
+++ b/internal/tui/picker.go
@@ -619,7 +619,15 @@ func (m model) normalizeProfileForProvider(provider providercatalog.Descriptor) 
 		strings.TrimSpace(profile.Name) == "" ||
 		strings.TrimSpace(profile.CatalogID) == ""
 	if normalizeIdentity {
-		profile.Name = provider.ID
+		// Only canonicalize an empty or generic placeholder name — never clobber a
+		// real user-provided profile name. The credential store is keyed by
+		// profile.Name (SecureProviderProfile stores under Name), so rewriting it
+		// here made the later profileWithCredential lookup miss the saved key and
+		// rebuild a keyless provider — a 401 on every /model switch for any profile
+		// named differently from its catalog id (e.g. "my-openai").
+		if strings.TrimSpace(profile.Name) == "" || genericProviderCatalogID(profile.Name) {
+			profile.Name = provider.ID
+		}
 		profile.CatalogID = provider.ID
 	}
 	if strings.TrimSpace(profile.BaseURL) == "" {
--- a/internal/tui/picker_test.go
+++ b/internal/tui/picker_test.go
@@ -976,3 +976,49 @@ func TestSwitchProviderModelWarmsDiscoveryForTheNewProvider(t *testing.T) {
 		t.Fatal("switching to ollama should warm both the generic and ollama-specific discovery commands")
 	}
 }
+
+func TestNormalizeProfileForProviderPreservesCustomName(t *testing.T) {
+	openai, ok := providercatalog.Get("openai")
+	if !ok {
+		t.Fatal("openai descriptor missing from catalog")
+	}
+
+	// A first-party provider set up with a custom profile name and the baseURL at
+	// the catalog default. normalizeIdentity is true here (baseURL matches), but
+	// the real name must survive: it is the credential-store key, so clobbering it
+	// dropped the saved key and 401'd every /model switch (issue #440).
+	m := model{providerProfile: config.ProviderProfile{
+		Name:         "my-openai",
+		CatalogID:    "openai",
+		BaseURL:      openai.DefaultBaseURL,
+		APIKeyStored: true,
+	}}
+	got := m.normalizeProfileForProvider(openai)
+	if got.Name != "my-openai" {
+		t.Fatalf("Name = %q, want it preserved as %q (credential store is keyed by Name)", got.Name, "my-openai")
+	}
+	if got.CatalogID != "openai" {
+		t.Fatalf("CatalogID = %q, want it canonicalized to %q", got.CatalogID, "openai")
+	}
+}
+
+func TestNormalizeProfileForProviderCanonicalizesPlaceholderName(t *testing.T) {
+	openai, ok := providercatalog.Get("openai")
+	if !ok {
+		t.Fatal("openai descriptor missing from catalog")
+	}
+
+	// Empty and generic placeholder names still get canonicalized to the catalog
+	// id — those carry no meaningful credential-store key to preserve.
+	for _, name := range []string{"", "custom-openai-compatible"} {
+		m := model{providerProfile: config.ProviderProfile{
+			Name:      name,
+			BaseURL:   openai.DefaultBaseURL,
+			CatalogID: name,
+		}}
+		got := m.normalizeProfileForProvider(openai)
+		if got.Name != "openai" {
+			t.Fatalf("Name = %q for placeholder %q, want canonicalized to %q", got.Name, name, "openai")
+		}
+	}
+}


gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change narrows the `profile.Name` overwrite in `normalizeProfileForProvider` (`internal/tui/picker.go`) so it only fires when the name is empty or a generic `custom-*` placeholder. The previous behavior unconditionally set `profile.Name = provider.ID` whenever `normalizeIdentity` was true, which includes any first-party provider whose saved BaseURL matches the catalog default. The credential store is keyed by `profile.Name` (`SecureProviderProfile` stores via `store.Set(profile.Name, ...)`, `ApplyStoredAPIKey` retrieves via `store.Get(name)` where `name = profile.Name`), so the old code clobbered the store key before `profileWithCredential` ran, causing a 401 on every `/model` switch for any profile named differently from its catalog id. The fix preserves the real user-provided name; `CatalogID` canonicalization is unchanged, so provider descriptor/metadata resolution is unaffected. The sole caller is the `/model` switch path in `command_center.go:416`. No other code path depends on `profile.Name` being equal to `provider.ID` after normalization — `ResolveRuntimeMetadata`, `persistSelectedModel`, and `SetActiveProviderEnv` all correctly use the preserved name.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./...` — pass
- `gofmt -l` on both changed files — clean (no output)
- `go test ./internal/tui/... -run TestNormalizeProfileForProvider` — both new tests pass
- Red-green verified: reverting `picker.go` to the main version causes `TestNormalizeProfileForProviderPreservesCustomName` to fail (`Name = "openai", want it preserved as "my-openai"`); restoring the fix makes it pass.
- `TestHandleAddDirCommand` fails in the `/tmp` worktree, but this is an environmental issue: macOS resolves `/tmp` to `/private/tmp`, which causes the sandbox to pick up extra write roots. Confirmed by reproducing the identical failure on `origin/main` in a `/tmp` worktree. Not introduced by this PR.
- CI on the PR head is all green (Analyze, CodeQL, Performance Smoke, Security & code health, Smoke macos/ubuntu/windows, Zero Review). The Windows job passed — no Windows-specific concerns.

## CONTRIBUTING

Author is a core-team maintainer, so the `issue-approved` gate does not apply. The PR links `Fixes #440`, is scoped to the single bug fix and its regression tests, includes tests, and has no UI changes requiring screenshots. Compliant.

## FINDINGS

1. **Nit — `internal/tui/picker_test.go:1013-1023` — placeholder-name test does not assert CatalogID canonicalization.** The loop in `TestNormalizeProfileForProviderCanonicalizesPlaceholderName` only checks `got.Name`. Since `CatalogID` is a core part of the normalization path and the fixture sets `CatalogID = name` (line 1017), adding `if got.CatalogID != "openai"` would tighten coverage. Not a blocker — the preservation test already asserts `CatalogID` canonicalization for the custom-name case.

No other findings. The fix is minimal, well-commented, and correctly targets the root cause.

## EXISTING REVIEWS

- **github-actions (zero-auto-review)**: "No blockers found" — all validation passed (diff hygiene, tests, build, smoke). On the current head (`2799cd02`). Not stale. This is a deterministic automated check; it does not cover product judgment or regression tracing.
- **coderabbitai**: APPROVED with one nitpick (the `CatalogID` assertion gap in the placeholder test described above). On the current head. Not stale. The nitpick is valid and trivial. Nothing was missed that affects merge readiness.

No reviews from human maintainers were present prior to this review.

## BOTTOM LINE

A minimal, correct fix that preserves the credential-store key during `/model` switching without altering CatalogID canonicalization; safe to merge.

kevincodex1 (approved):

PR #442, opened 2026-07-03
fix(doctor): verify the native binary in the runtime check instead of always passing

### LINKED ISSUE #405: doctor reports "Go runtime available" (pass) while the native binary is missing and all agent commands fail
### Summary

When the native binary isn't present next to the npm wrapper, `zero doctor`'s `runtime.go` check still reports `[pass] Zero Go runtime is available (go)` — even though `zero exec` and `zero providers list` both hard-fail with `[zero] No native binary found next to the npm wrapper.` The one precondition that's actually broken gets a green check, which sends you debugging the wrong thing (I went after provider/network config first).

### How I hit it

Fresh `npm i -g @gitlawb/zero` where the postinstall's release-asset download didn't complete (restricted network in my case). The postinstall exits 0 regardless — by design, so `npm install` doesn't break — leaving `bin/` with only `zero.js` and no native binary. Nothing surfaced this until I ran an agent command.

### Repro (deterministic)

```
ZERO_SKIP_DOWNLOAD=1 npm i -g @gitlawb/zero
zero doctor              # -> [pass] runtime.go - Zero Go runtime is available (go)
zero exec --list-tools   # -> [zero] No native binary found next to the npm wrapper.
zero providers list      # -> same "No native binary found" error
```

### Expected

`doctor` should verify the native binary actually exists and is runnable next to the wrapper, and **fail that check loudly** with the same guidance the wrapper prints (*"No native binary found — reinstall or build from source"*). Since a missing binary makes every agent command fail, it should be the first/hardest failure — not a pass.

### Why it matters

`exec` and `providers list` correctly detect the missing binary, but `doctor` — the command you run *to diagnose setup* — doesn't, so it actively points away from the real cause. A binary-presence check in `doctor` would also be a natural safety net for the postinstall silently exiting 0 on a failed download.

### Env

ZERO 0.1.0 · macOS 15 (x64) · installed via npm (`@gitlawb/zero`)

gnanam1990:
## Summary

`zero doctor`'s runtime check returned `StatusPass` **unconditionally** with a hardcoded `"go"` label — it verified nothing (not the running binary, not a version), so it was a false-confidence green: *"Zero Go runtime is available (go)"* regardless of actual state (#405).

This makes the check real: it resolves the running binary via `os.Executable`, stats it, and reports the resolved **path** + the actual **Go version** — failing loudly if the path can't be resolved, the file has vanished, or it resolves to a directory. `os.Executable` / `os.Stat` are injected through `doctor.Options` (`Executable` / `StatExecutable`) so the check is testable without a real on-disk layout. The caller now passes `runtime.Version()` instead of the `"go"` placeholder.

**Before / after** (live `zero doctor`):

- before: `[pass] Zero Go runtime is available (go).`
- after:  `[pass] Zero native binary is present and runnable at /…/zero (go1.26.4).`

**Scope note:** doctor runs *inside* the native binary, so it can't diagnose the binary being **absent** at install time — that case is already caught loudly by the npm wrapper (`bin/zero.js` refuses to launch, present since v0.1.0), which is why the issue's literal "doctor passes while the binary is missing" repro doesn't actually occur. The real defect fixed here is the **tautological check**; surfacing the binary path is the genuinely useful part — it lets you spot an npm-installed binary vs a stale or shadowing copy on `PATH`, which is the confusion behind #405.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` — all green (73 packages, 0 failures); `gofmt` clean.
- Red-green tests (`internal/doctor/runtime_check_test.go`): a present binary → **PASS** with the resolved path in the message/details; `os.Executable` error, a missing/inaccessible binary, and a directory path each **FAIL** (the old unconditional-pass code fails these). The default (uninjected) path resolves the real running binary.
- Live-run confirmed: `./zero doctor` now reports the resolved binary path + real Go version.

## Linked issue

Fixes #405

## Checklist

- [x] `gofmt` clean.
- [x] Tests added/updated for the change.
- [x] No UI change — CLI diagnostic output; before/after shown above in lieu of a screenshot.
- [ ] The linked issue already has the `issue-approved` label — N/A (team internal-cycle PR; #405 is assigned to me).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the diagnostics/environment check to verify the running native binary is resolved, accessible, and not a directory, with clearer failure reporting.
  * Runtime diagnostics now use the detected Go version for more accurate diagnostics.
* **Tests**
  * Added coverage for runtime check pass/fail scenarios, including missing/unresolvable binaries, directory paths, and default behavior against the real running binary.

### DIFF
--- a/internal/cli/observability.go
+++ b/internal/cli/observability.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -65,7 +66,7 @@ func runDoctor(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) 
 
 	report := doctor.Run(doctor.Options{
 		Now:            deps.now,
-		Runtime:        "go",
+		Runtime:        runtime.Version(),
 		UserConfig:     userConfig,
 		ProjectConfig:  projectConfig,
 		Provider:       provider,
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -129,11 +129,11 @@ func Format(report Report) string {
 // part: it lets a user see which binary answered (e.g. an npm-installed one vs a
 // stale/shadowing copy on PATH) rather than a meaningless green.
 func runtimeCheck(options Options) Check {
-	runtime := strings.TrimSpace(options.Runtime)
-	if runtime == "" {
-		runtime = "go"
+	runtimeVersion := strings.TrimSpace(options.Runtime)
+	if runtimeVersion == "" {
+		runtimeVersion = "go"
 	}
-	details := map[string]any{"runtime": runtime}
+	details := map[string]any{"runtime": runtimeVersion}
 
 	executable := options.Executable
 	if executable == nil {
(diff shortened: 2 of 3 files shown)


gnanam1990:
@coderabbitai review

Both items from the earlier review are addressed in `5068a38` (pushed after the reviewed commit `2a5b2af`):

- The local `runtime` was renamed to `runtimeVersion`, so it no longer shadows the stdlib `runtime` package.
- The unresolvable-binary message now matches the npm wrapper's remediation: `…Reinstall the zero package, or build from source: https://github.com/Gitlawb/zero`.

Re-requesting a review so the stale CHANGES_REQUESTED can clear.

PR #443, opened 2026-07-03
fix(auth): route zero auth login chatgpt to the dedicated ChatGPT flow

### LINKED ISSUE #430: Unable to add ChatGPT as a model provider in headless mode
### Version / branch / commit

zero 0.1.0

### OS and environment

Ubuntu 26.04 LTS / SSH'd into from Ghostty 

### Steps to reproduce

1. run `zero auth login chatgpt`
2. Visit the link provided (https://auth.openai.com/...)
3. Observe "Authentication Error" as shown in the screenshot

### Expected behavior

The login works successfully

### Actual behavior

Received an "Authentication Error" instead.

Image

anandh8x:
## Summary

`zero auth login chatgpt` failed with an "Authentication Error" because it went through the generic OAuth loopback path, which builds a random-port `http://127.0.0.1:<port>/callback` redirect_uri with no extra authorize params. OpenAI's ChatGPT client registration requires exactly `http://localhost:1455/auth/callback` plus the `id_token_add_organizations`, `codex_cli_simplified_flow`, and `originator` params, so the authorize endpoint rejected the request before any browser interaction.

This routes the `chatgpt` provider to the existing dedicated ChatGPT login flow (`provideroauth.ChatGPTLogin`), which already binds the fixed port 1455, uses the `localhost` host, the `/auth/callback` path, and sends the required params. `zero auth login chatgpt` now behaves identically to `zero auth chatgpt`.

## Changes

- `runAuthLogin` delegates to `runAuthChatGPT` when the provider is `chatgpt` (case-insensitive), instead of the generic `manager.Login` path.
- `--device` is rejected with a clear message for `chatgpt` since the flow is loopback-only (no device-code path).
- Help text updated to mention `zero auth login chatgpt`.

## Root cause

The generic loopback flow (`loginLoopback` in the OAuth manager) cannot satisfy ChatGPT's requirements:

1. **Wrong redirect_uri** — random port, `127.0.0.1` host, `/callback` path vs the required `localhost:1455/auth/callback`.
2. **Missing mandatory params** — `id_token_add_organizations`, `codex_cli_simplified_flow`, `originator` are not sent (the generic path passes `nil` extra params).

These are per-provider details the generic engine doesn't know about, so routing to the dedicated flow is the correct fix rather than special-casing the generic path.

Fixes #430

## Summary by CodeRabbit

* **Bug Fixes**
  * `zero auth login` now handles ChatGPT sign-in through the dedicated browser-based flow, including a clear error when `--device` is used with ChatGPT.
  * Other authentication providers continue to use the standard login experience.

* **Documentation**
  * Updated the auth help text to better explain ChatGPT and related OAuth environment settings.

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -301,12 +301,24 @@ func runAuthLogin(args []string, stdout io.Writer, stderr io.Writer, deps appDep
 	if len(parsed.positional) != 1 {
 		return writeExecUsageError(stderr, "usage: zero auth login <provider> [--device] [--scope <scope>]")
 	}
+	provider := parsed.positional[0]
+	// ChatGPT (Codex) requires a fixed redirect_uri (http://localhost:1455/
+	// auth/callback) and mandatory authorize params (id_token_add_organizations,
+	// codex_cli_simplified_flow, originator) that the generic loopback flow
+	// cannot supply. Route it to the dedicated ChatGPT login so
+	// `zero auth login chatgpt` behaves identically to `zero auth chatgpt`.
+	if strings.EqualFold(provider, "chatgpt") {
+		if parsed.device {
+			return writeExecUsageError(stderr, "ChatGPT login does not support --device (it is loopback-only)")
+		}
+		return runAuthChatGPT(nil, stdout, stderr, deps)
+	}
 	manager, err := newAuthManager(deps, stdout)
 	if err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
 	}
 	status, err := manager.Login(context.Background(), oauth.LoginOptions{
-		Provider:    parsed.positional[0],
+		Provider:    provider,
 		Device:      parsed.device,
 		ExtraScopes: parsed.scopes,
 	})
@@ -489,8 +501,9 @@ Commands:
 A provider is any OAuth 2.0 / OIDC server. "openrouter" ('zero auth openrouter')
 works out of the box. "xai" ('zero auth login xai') uses a built-in preset that is
 off by default — enable it with ZERO_OAUTH_ALLOW_PRESETS=1, or set the
-ZERO_OAUTH_XAI_* vars yourself. Any preset field is overridable via the env vars
-below. For a custom provider named <name>, set:
+ZERO_OAUTH_XAI_* vars yourself. "chatgpt" ('zero auth login chatgpt' or
+'zero auth chatgpt') uses a fixed-port loopback flow against the Codex backend.
+Any preset field is overridable via the env vars below. For a custom provider named <name>, set:
   ZERO_OAUTH_<NAME>_CLIENT_ID       (required)
   ZERO_OAUTH_<NAME>_CLIENT_SECRET   (optional)
   ZERO_OAUTH_<NAME>_AUTHORIZE_URL   ZERO_OAUTH_<NAME>_TOKEN_URL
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
## Approve — closes a real, filed bug; clean and well-tested

**Worth it:** yes. This fixes #430: `zero auth login chatgpt` currently fails with an Authentication Error because it goes through the generic OAuth loopback path, which builds a random-port `http://127.0.0.1:<port>/callback` redirect_uri and omits the mandatory authorize params (`id_token_add_organizations`, `codex_cli_simplified_flow`, `originator`) that OpenAI's Codex client registration requires. `zero auth login <provider>` is the uniform entry point users reach for, so a broken `chatgpt` there is a genuine UX gap. Thanks @anandh8x.

The fix is the right shape: rather than teaching the generic engine ChatGPT-specific quirks, it routes the `chatgpt` provider (case-insensitive) to `runAuthChatGPT` — the same function `zero auth chatgpt` already dispatches to — so the two commands now behave identically.

**Verified on Windows:**
- `go build ./...` clean; `go test ./internal/cli/ -run 'Auth' -v` — all pass, including the two new tests and the pre-existing `TestRunAuthHelp` (the help-text edit only adds a line; the test checks stable substrings — none of the #414-class hardcoded-string breakage here).
- No credential-bleed: persistence stays under `ProviderKey("chatgpt")`, same store path as before.
- Nice touch: `--device`/`--scope` are rejected up front with clear ChatGPT-specific messages instead of being silently dropped.

CI all 9 checks green including the Windows smoke.

**One small note** (non-blocking): both new tests exercise the `--device`/`--scope` *rejection* branches, which return before any port bind / browser launch — so the successful routing itself is only inferred from the rejection message, not asserted directly. That's acceptable since the real login can't run headless in CI. Approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change adds a routing guard in `runAuthLogin` (`internal/cli/auth.go`) that delegates `chatgpt` to the existing `runAuthChatGPT` dedicated flow before reaching the generic `manager.Login` path. The generic path is unchanged for all other providers. The only behavioral shift is for `zero auth login chatgpt`, which previously failed with an authentication error (wrong redirect_uri, missing mandatory params) and now succeeds via the dedicated flow. `--device` and `--scope` are explicitly rejected for `chatgpt` with clear error messages, which is correct since the ChatGPT flow is loopback-only with fixed scopes. No existing provider's auth flow is altered.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/cli/...` — pass
- `go test ./internal/cli/... -run TestRunAuthLogin` — all four tests pass (Validation, UnknownProvider, ChatGPTRoutesToDedicatedFlow, ChatGPTRejectsScope)
- CI all green (macOS, Ubuntu, Windows, Analyze, CodeQL, Security, Zero Review)

## CONTRIBUTING

Author is a core-team maintainer (anandh8x), so the `issue-approved` gate does not apply. The PR links `Fixes #430`, is scoped to the single routing fix plus tests, and includes two new tests. Compliant.

## FINDINGS

None. The routing logic is correct, the `--device` and `--scope` rejections are appropriate, and the case-insensitive matching (`strings.EqualFold`) is a good touch.

## EXISTING REVIEWS

- **Vasanthdev2004** (APPROVED): Assessed the fix as closing a real filed bug, clean and well-tested. Valid on the current head. Nothing missed.
- **coderabbitai** (initially CHANGES_REQUESTED, then APPROVED): First flagged that `--scope` was silently ignored for ChatGPT. The author addressed this by adding an explicit `--scope` rejection. The follow-up review approved. Both the finding and the fix are valid on the current head. Nothing missed.

All reviews are on the current head (`867b3ea`). The coderabbitai changes-requested review is on an earlier commit but was resolved by the subsequent approved review on the current head. No stale concerns.

## BOTTOM LINE

A clean, well-tested fix that routes ChatGPT login to the correct dedicated flow; safe to merge.

kevincodex1 (approved):
LGTM

PR #444, opened 2026-07-03
fix: allow non-TLS connections to private-network provider endpoints

### LINKED ISSUE #431: Allow non-TLS connection to local network provider
### Problem

When configuring a local provider (e.g. llama.cpp on another machine inside a local private network such as 192.168.x.y) Zero requires HTTPS and does not accept a non-TLS setup (e.g. http://192.168.x.y:8080/v1).
Currently, non-TLS connections are only accepted on loopback. 

### Requested behavior

Zero should allow configuring a plain HTTP endpoint on non-routable local range such as 192.168.0.0/16 

### Why it belongs in Zero

The current behaviour is too strict as it prevents using valid safe local connections. 

llama.cpp by default compiles without TLS and it is safe to use on a local network. OpenCode and Claude Code allow this configuration.

_No response_

_No response_

anandh8x:
## Summary

The provider wizard and health probe rejected `http://` endpoints with a non-loopback host, so users could not configure a local-network provider like llama.cpp on another LAN machine (e.g. `http://192.168.1.100:8080/v1`). Only loopback (localhost/127.0.0.1/::1) was relaxed for non-TLS.

This expands the relaxation to also cover private-network addresses — RFC 1918 ranges (192.168.0.0/16, 10.0.0.0/8, 172.16.0.0/12) — using `netip.Addr.IsPrivate()`.

## Changes

- **`providerhealth`**: Renamed `endpointIsLoopback` → `endpointIsLoopbackOrPrivate` and `allowLoopback` → `allowLoopbackOrPrivate`. The flag now relaxes both loopback and private addresses in `validateEndpoint` and `safeDialContext`.
- **`providerWizardIsLoopbackHost`** (TUI wizard): Now accepts private IPs, so `http://192.168.x.y` passes the endpoint URL validation.
- **`providerProfileIsLocal`** (headless setup): Recognizes private-network IPs as local (skips API key requirement for keyless LAN providers like llama.cpp).
- **`baseURLIsLoopback`** (app.go): Recognizes private-network IPs for credential-requirement logic.

## What stays blocked

- Link-local (169.254.x.x), cloud metadata, documentation, and other special-use ranges remain blocked regardless of the flag.
- Redirects (3xx) to any loopback or private address remain blocked — the flag is always `false` for redirect targets (SSRF protection unchanged).

Fixes #431

## Summary by CodeRabbit

* **New Features**
  * Private-network IPs are now treated as “local” alongside loopback/localhost during provider setup and endpoint checks.
  * Provider connectivity and wizard validation now recognize local private-network destinations (e.g., RFC1918 ranges).

* **Bug Fixes**
  * Connectivity probe safety rules were tightened: loopback/private endpoints are allowed only when they match the user-configured base URL.
  * Error wording and validation behavior were updated to reflect the expanded local address support.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -1158,7 +1158,8 @@ func profileHasCredential(profile config.ProviderProfile) bool {
 }
 
 // baseURLIsLoopback reports whether a provider base_url points at a loopback host
-// (a keyless local provider like Ollama/LM Studio), which needs no API key.
+// or a private-network host (192.168.x.y / 10.x.y.z / 172.16.x.y) — a local
+// provider like Ollama/LM Studio that needs no API key.
 func baseURLIsLoopback(baseURL string) bool {
 	u := strings.TrimSpace(baseURL)
 	if u == "" {
@@ -1173,7 +1174,7 @@ func baseURLIsLoopback(baseURL string) bool {
 		return true
 	}
 	if addr, err := netip.ParseAddr(host); err == nil {
-		return addr.IsLoopback()
+		return addr.IsLoopback() || addr.IsPrivate()
 	}
 	return false
 }
--- a/internal/cli/setup.go
+++ b/internal/cli/setup.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"strconv"
 	"strings"
@@ -451,6 +452,11 @@ func providerProfileIsLocal(profile config.ProviderProfile) bool {
 	}
 	switch strings.ToLower(parsed.Hostname()) {
 	case "localhost", "127.0.0.1", "::1":
+		return true
+	}
+	// Also treat explicit private-network IPs (192.168.x.y, 10.x.y.z,
+	// 172.16-31.x.y) as local — the user is pointing at their own LAN box.
+	if ip := net.ParseIP(parsed.Hostname()); ip != nil && ip.IsPrivate() {
 		return true
 	}
 	return false
--- a/internal/providerhealth/connectivity_client_test.go
+++ b/internal/providerhealth/connectivity_client_test.go
@@ -169,8 +169,8 @@ func TestValidateEndpointAllowsLoopbackOnlyForLocalProvider(t *testing.T) {
 		if err := validateEndpoint(ctx, ep, r, true); err != nil {
 			t.Errorf("local provider endpoint %q should be allowed with allowLoopback=true, got %v", ep, err)
 		}
-		if !endpointIsLoopback(ep) {
-			t.Errorf("endpointIsLoopback(%q) = false, want true", ep)
+		if !endpointIsLoopbackOrPrivate(ep) {
+			t.Errorf("endpointIsLoopbackOrPrivate(%q) = false, want true", ep)
 		}
 	}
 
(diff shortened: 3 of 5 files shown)


Vasanthdev2004 (approved):
## Approve — correctly-scoped, fails closed

**Worth it:** yes. Fixes a real filed problem (#431): you couldn't point Zero at a keyless local model server (llama.cpp / Ollama / LM Studio) on another LAN box over `http://192.168.x.y` because the plaintext-`http` relaxation was loopback-only. Extending it to RFC1918 / ULA private ranges is exactly the surface that use case needs. Thanks @anandh8x.

### Security review (the load-bearing part) — no downgrade/SSRF hole to public hosts
- **Doubly gated.** The guard is `blockedAddrReason(addr) != "" && !(allowLoopbackOrPrivate && (addr.IsLoopback() || addr.IsPrivate()))`. The flag is set true **only** from `endpointIsLoopbackOrPrivate(user's own base_url)`, which requires a literal loopback/private IP or `localhost` — a **public hostname parses as non-IP → flag false**, so DNS-resolution blocking is unchanged. And even with the flag true, the *resolved* address must itself be loopback/private to be relaxed.
- **Special ranges stay blocked.** Go's `IsPrivate()` is only RFC1918 + ULA `fc00::/7`; it does **not** include `169.254/16` link-local (i.e. `169.254.169.254` cloud metadata), `100.64/10` CGNAT, documentation, or multicast — all remain blocked. The test correctly swaps the old `10.0.0.5`-stays-blocked case for `192.0.2.1` (documentation) and asserts `169.254.169.254` stays blocked.
- **Redirects unchanged.** 3xx targets are always validated with `allowLoopbackOrPrivate=false`, so SSRF-via-redirect to loopback/private is still refused; `safeDialContext` re-validates and dials the literal IP (TOCTOU window closed).
- **No new plaintext-credential path** — the wizard still forces `https` for any non-loopback/non-private host, so credentials only go plaintext to a private/loopback box the user explicitly typed.

**Verified on Windows:** `go build ./...` clean; targeted tests in `providerhealth`, `tui`, `cli` pass; `vet` clean. CI fully green incl. Windows smoke (real run, not a fork).

**Non-blocking nits:** `providerProfileIsLocal` (setup.go) uses `net.ParseIP` + a literal `127.0.0.1` switch while app.go uses `netip` + `IsLoopback()||IsPrivate()` — a minor inconsistency that errs safe (would require a key); `baseURLIsLoopback` now also matches private hosts so the name is a touch stale; and a private-network redirect-block test case would document the guarantee (the enforced path is identical — flag hard-coded `false` for all redirects). None block. Approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change expands the non-TLS endpoint relaxation from loopback-only to also include RFC 1918 private-network addresses (192.168.0.0/16, 10.0.0.0/8, 172.16.0.0/12). This affects four call sites: `baseURLIsLoopback` (credential requirement check), `providerWizardIsLoopbackHost` (TUI endpoint validation), `providerProfileIsLocal` (headless setup credential skip), and the `providerhealth` SSRF guard (`validateEndpoint` / `safeDialContext` / `endpointIsLoopbackOrPrivate`). The expansion is correctly scoped: the `allowLoopbackOrPrivate` flag is only set true for the user's own configured `base_url`, never for redirect targets (the redirect handler at line 483 always passes `false`). Link-local (169.254.x.x), cloud metadata, documentation ranges, and other special-use addresses remain blocked. No existing loopback behavior is removed — this is purely additive for private-network IPs.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/providerhealth/... ./internal/cli/... ./internal/tui/...` — pass
- `gofmt -l` on all changed Go files — clean
- `go test ./internal/providerhealth/... -count=1` — all 16 tests pass, including `TestValidateEndpointAllowsLoopbackOnlyForLocalProvider` (updated to test private-network allow + link-local block) and `TestProbeConnectivityBlocksPrivateResolvedHostBeforeNetwork`
- `go test ./internal/cli/... -run TestSetup -count=1` — all pass, including `TestSetupMissingCredentialEnv/local_provider`
- CI all green (macOS, Ubuntu, Windows, Analyze, CodeQL, Security, Zero Review)

## CONTRIBUTING

Author is a core-team maintainer (anandh8x), so the `issue-approved` gate does not apply. The PR links `Fixes #431`, is scoped to the private-network relaxation and its tests, and includes updated test coverage. Compliant.

## FINDINGS

1. **Nit — `internal/providerhealth/connectivity_client_test.go:207-216` — no redirect-blocking test for private addresses.** The test verifies that private addresses are allowed when the flag is true, but does not add a case verifying that a redirect to a private address is still blocked (flag=false). The code is correct (redirect handler always passes `false`), but an explicit test would tighten the SSRF regression net. Noted by coderabbitai as well. Not a blocker.

No other findings.

## EXISTING REVIEWS

- **Vasanthdev2004** (APPROVED): Assessed as correctly-scoped, fails closed. Valid on the current head. Nothing missed.
- **coderabbitai** (APPROVED + 5 nitpicks): Flagged the missing redirect-blocking test among other minor items. The redirect-blocking concern is valid; the rest are trivial. Nothing critical missed.

All reviews on the current head (`08d392e`). No stale reviews.

## BOTTOM LINE

A well-scoped security-safe expansion that allows private-network provider endpoints over HTTP while preserving SSRF protections; safe to merge.

kevincodex1 (approved):
LGTM

PR #445, opened 2026-07-03
fix(oauth): retry secret-lock on Windows ERROR_ACCESS_DENIED (delete-pending race)

gnanam1990:
## Summary

`createSecretFile` (`internal/oauth/encrypt.go`) takes a `<secret>.lock` with `O_CREATE|O_EXCL` and retries on contention — but it only treated **`os.ErrExist`** as contention. On Windows, when a concurrent holder's `defer os.Remove(lockPath)` leaves the lock file in a **"delete pending"** state, a racer's `O_EXCL` create returns **`ERROR_ACCESS_DENIED` (`os.ErrPermission`)**, not `ErrExist` — so it fell through to a hard `oauth: create token secret lock: Access is denied` error instead of retrying.

This treats `os.ErrPermission` as retryable contention too, mirroring the **identical handling already present in `acquireFileLock` (`lock.go`, from #261)** for the exact same OS quirk. That sibling lock got the fix; this second lock loop was missed.

**Impact:** fixes the intermittently-failing Windows CI test `TestLoadOrCreateSecretConcurrentConverges` that has been red-flagging the Windows smoke job on unrelated PRs — and fixes a real Windows bug where concurrent token-secret creation (parallel logins/refreshes) could spuriously fail.

## Verification

- macOS: `go build ./...`, `go vet`, `gofmt` clean; `go test ./internal/oauth/` green; the concurrent-converge test passes under `-race` (5×).
- ⚠️ The Windows delete-pending `ERROR_ACCESS_DENIED` path **can't be reproduced off Windows** — the definitive confirmation is this PR's **Windows smoke** job going green (it was failing on that exact test before this change).

## Linked issue

Team / internal-cycle PR — no separate issue; fixes a CI-blocking Windows flake surfaced across recent PRs.

## Checklist

- [x] `go build ./...`, `go vet ./...`, `go test ./internal/oauth/` pass locally (non-Windows).
- [x] `gofmt` clean.
- [x] Fix mirrors the existing `lock.go` handling; explanatory comment added.
- [ ] Windows CI green — the real proof; watch this PR's Windows smoke job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret file creation on Windows by handling a permission-related lock contention case correctly.
  * Reduced spurious failures when multiple processes try to create secrets at the same time.

### DIFF
--- a/internal/oauth/encrypt.go
+++ b/internal/oauth/encrypt.go
@@ -114,7 +114,12 @@ func createSecretFile(path string) ([]byte, error) {
 			}
 			return writeNewSecretFile(path)
 		}
-		if !errors.Is(err, os.ErrExist) {
+		// On Windows a concurrent holder's os.Remove leaves the lock file in a
+		// "delete pending" state, so an O_EXCL create races it with
+		// ERROR_ACCESS_DENIED (os.ErrPermission) rather than ErrExist. Treat that
+		// as contention and retry too — mirroring acquireFileLock in lock.go —
+		// otherwise concurrent secret creation spuriously fails on Windows.
+		if !errors.Is(err, os.ErrExist) && !errors.Is(err, os.ErrPermission) {
 			return nil, fmt.Errorf("oauth: create token secret lock: %w", err)
 		}
 		if data, rerr := readSecretFileRetry(path); rerr == nil {


Vasanthdev2004 (approved):
## Approve — genuine root-cause fix, and it kills the CI flake for real

**Worth it:** yes. On Windows a concurrent holder's `defer os.Remove(lockPath)` leaves the `.lock` file *delete-pending*, so a racer's `O_EXCL` create returns `ERROR_ACCESS_DENIED` (`os.ErrPermission`) rather than `ErrExist`. The old `createSecretFile` loop only treated `ErrExist` as contention, so parallel token-secret creation (concurrent logins/refreshes) could spuriously hard-fail with `oauth: create token secret lock: Access is denied`. Treating `ErrPermission` as retryable is exactly right — and it mirrors the already-merged, identical handling in `acquireFileLock` (`lock.go`, #261); that sibling lock got the fix and this one was simply missed. 6 lines, precedented, well-commented.

**Verified on Windows** (isolated worktree): `go test ./internal/oauth/ -run TestLoadOrCreateSecret -count=5` → 5/5 green (the concurrent-converge test that flakes now passes reliably); full package green; `go build`/`vet`/`gofmt` clean. I also confirmed it against the **original racy test at `-count=100` → 100/100**. CI all green including `Smoke (windows-latest)` — same-repo branch, so that Windows pass is real.

**Security:** safe. The new `ErrPermission` branch is read-only (`readSecretFileRetry` only); the sole write path (`writeNewSecretFile`, temp-file + atomic rename) runs only after `O_EXCL` exclusively acquired the lock, so the broadened retry can't double-write or corrupt the secret. Fail-closed behavior untouched.

**Nit (non-blocking):** broadening to `ErrPermission` means a *genuine* permission failure (dir truly unwritable) now surfaces as a `timed out waiting for token secret lock` timeout rather than the immediate `Access is denied` — cosmetic diagnostics, and the same tradeoff already accepted in `lock.go` #261.

This is the correct fix of the #445/#451 pair — see my note on #451. Thanks @gnanam1990.

kevincodex1 (approved):

PR #448, opened 2026-07-03
Keep provider key scoped to ZERO step

GautamBytes:
Summary:
- stop writing provider API keys to GITHUB_ENV
- keep the same-step export that zero exec needs
- align docs and smoke checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret handling in the GitHub Action so provider API keys are available only during the ZERO run step and are no longer carried into later steps.
  * Tightened validation to better catch incorrect action behavior and documentation mismatches.

* **Documentation**
  * Clarified how `api-key`, `api-key-env`, and related settings behave.
  * Updated security notes to explain what is exported, logged, and preserved across steps.

### DIFF
--- a/.github/workflows/zero-action-smoke.yml
+++ b/.github/workflows/zero-action-smoke.yml
@@ -4,12 +4,14 @@ on:
   pull_request:
     paths:
       - action.yml
+      - docs/GITHUB_ACTION.md
       - .github/workflows/zero-action-smoke.yml
   push:
     branches:
       - main
     paths:
       - action.yml
+      - docs/GITHUB_ACTION.md
       - .github/workflows/zero-action-smoke.yml
 
 jobs:
@@ -29,7 +31,8 @@ jobs:
           import sys, yaml
 
           with open("action.yml", "r", encoding="utf-8") as handle:
-              doc = yaml.safe_load(handle)
+              action_raw = handle.read()
+          doc = yaml.safe_load(action_raw)
 
           assert isinstance(doc, dict), "action.yml must be a mapping"
           assert doc.get("name"), "action must declare a name"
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (approved):
## Approve — clean, correct security hardening

**Worth it:** yes. This closes a genuine credential-exposure gap in the GitHub Action: the provider API key was written to `GITHUB_ENV`, which persists it into the environment of **every subsequent step in the job** — including any third-party actions a user runs after ZERO. Removing that write scopes the key to the single `Run ZERO` step, where it's still available in-process via `export`. Thanks @GautamBytes.

**No regression:** `zero exec` runs *inside* the same `Run ZERO` step, so the in-process `export` is all the provider needs. I confirmed no later step consumes the key — the PR-comment step uses `GH_TOKEN`, the Slack step uses `SLACK_WEBHOOK_URL`, neither touches the provider credential.

**Security check:**
- Key now lives only in the Run ZERO process env — `grep` confirms zero remaining `GITHUB_ENV` writes of any `API_KEY` var. ✓
- Fail-closed unchanged: the `api-key-env` name is still validated against `^[A-Za-z_][A-Za-z0-9_]*$` before export. ✓
- No plaintext downgrade: `set +x`, no `echo` of the value, never logged. ✓
- Regression guard added: the smoke workflow now asserts the key is still exported for the step, is never persisted to `GITHUB_ENV`, and that docs no longer claim later steps inherit it.

I ran the new smoke assertions locally against the PR's `action.yml` (pyyaml) — all pass. No Go surface. CI fully green including the Windows smoke (3m53s), which is exactly what guards this change.

**Non-blocking:** the `"later steps inherit" not in action_raw.lower()` assertion is a string-literal guard tied to the old comment wording; the line-level `GITHUB_ENV`+`API_KEY` check below it is the real protection. And worth a one-line release note that any downstream workflow relying on the previously-leaked env var will now see the key unset (that leak *was* the vulnerability). Neither blocks. Tightly-scoped hardening — approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change removes a single line from `action.yml` that wrote the provider API key to `GITHUB_ENV` (`printf '%s=%s\n' ... >> "${GITHUB_ENV}"`). The in-process `export` for the same step is retained, so `zero exec` still sees the key. The effect is that the key no longer persists into subsequent job steps — a security hardening, not a behavior regression for the ZERO step itself. The smoke workflow adds assertions verifying the key is not written to `GITHUB_ENV` and that the `Run ZERO` step exports it in-process. Docs are updated to match.

## BUILD / TEST

- `go build ./...` — pass (no Go files changed; only YAML and Markdown)
- `go vet ./...` — pass
- CI all green, including the `Validate action.yml` job which now enforces the new assertions
- The smoke test's Python validation now checks: (1) the Run ZERO step exports the key, (2) the key is not persisted to `GITHUB_ENV`, (3) docs don't claim later steps inherit it

## CONTRIBUTING

No linked issue. The author is a community contributor (GautamBytes). Per CONTRIBUTING.md, community PRs require an approved parent issue with the `issue-approved` label. No such issue is linked. This is a process violation, though the change is a legitimate security hardening. Maintainer's call.

## FINDINGS

1. **Nit — `.github/workflows/zero-action-smoke.yml:87-101` — minor duplication in Python validation.** The script re-reads and re-parses `action.yml` into `action_raw` after already loading `doc` from `yaml.safe_load`. This works but could reuse the already-parsed `doc` for the structural checks. Not a blocker — the duplication is harmless and the assertions are correct.

No other findings.

## EXISTING REVIEWS

- **Vasanthdev2004** (APPROVED): Assessed as clean, correct security hardening closing a genuine credential-exposure gap. Valid on the current head. Nothing missed.
- **coderabbitai** (APPROVED + 1 nitpick): Flagged the same Python duplication. Valid, trivial. Nothing missed.
- **copilot-pull-request-reviewer** (COMMENTED): Accurate overview of the change. Nothing missed.

All reviews on the current head (`b8a1abc`). No stale reviews.

## BOTTOM LINE

A clean security fix that stops persisting provider API keys to `GITHUB_ENV`; safe to merge, noting the missing issue link for maintainer awareness.

kevincodex1 (approved):
nice one!

PR #450, opened 2026-07-03
fix(doctor): detect missing native binary during runtime checks

### LINKED ISSUE #405: doctor reports "Go runtime available" (pass) while the native binary is missing and all agent commands fail

michaelkillgta:
Fixes #405

## Summary

When the native binary is missing next to the npm wrapper (`ZERO_SKIP_DOWNLOAD=1`, failed postinstall, etc.), `zero doctor` previously gave users a false sense that setup was fine — or a generic wrapper error that did not look like a doctor report — while `zero exec` and `zero providers list` failed with the real cause.

## Root cause

`bin/zero.js` is the only entry point for npm-installed `zero`. When the native binary is absent, the wrapper exits before spawning Go. The old Go-side `runtimeCheck` (`os.Executable()` + `os.Stat()`) could never observe this state: if Go is running, the binary already exists.

## Why the fix lives in the wrapper

The missing binary is detected in `bin/zero.js` at `existsSync(nativePath)` — before any Go process starts. The wrapper must surface a doctor-shaped failure for `zero doctor` so the diagnostic command reports the actual problem instead of sending users to debug provider config or network issues.

## CLI compatibility

The wrapper mirrors `parseDoctorArgs` and `writeDoctorHelp` from the Go doctor command:

| Invocation | Output | Exit |
|---|---|---|
| `zero doctor` | Text doctor report on **stdout** | 1 |
| `zero doctor --json` | `doctor.Report` JSON on **stdout** | 1 |
| `zero doctor --help` / `help` / `-h` | Help text on **stdout** | 0 |
| `zero doctor --bogus` | `[zero] unknown doctor flag "..."` on **stderr** | 2 |
| All other commands | Unchanged generic missing-binary error | 1 |

Only valid doctor invocations use the missing-native-binary fallback. Help and invalid flags behave like the real CLI.

## Tests

Added/updated wrapper tests in `internal/npmwrapper/npmwrapper_test.go`:

- `TestNodeWrapperDoctorReportsMissingNativeBinaryAsDoctorFail` — text report on stdout, not stderr
- `TestNodeWrapperDoctorJSONReportsMissingNativeBinaryAsJSONFail` — JSON schema + stdout-only
- `TestNodeWrapperDoctorHelpShowsUsage` — `--help`, `help`, `-h`
- `TestNodeWrapperDoctorRejectsUnknownFlag` — unknown flag error on stderr, exit 2
- `TestNodeWrapperDoctorWithFlagsStillReportsDoctorFail` — `--connectivity` still hits fallback
- Existing `TestNodeWrapperReportsMissingNativeBinary` unchanged for non-doctor commands

## Test plan

- [x] `go test ./internal/npmwrapper/... -count=1`
- [x] `git diff --check upstream/main`

## Summary by CodeRabbit

* **Bug Fixes**
  * When native binaries are missing, `doctor` now always produces a doctor-shaped failure report on stdout (including `runtime.go`) with correct exit codes (`--help`/`help` exit `0`, invalid flags exit `2`) and an actionable remedy to run `postinstall.mjs`.
  * `doctor --json` now returns valid JSON (`ok: false`) with exactly one failing `runtime.go` check.
  * Bun-specific recovery guidance is consistent across generic and `doctor` missing-binary outputs.

* **Tests**
  * Expanded `doctor` CLI/output contract coverage (help, JSON, unknown flags) and added regression checks for Bun recovery copy consistency.

### DIFF
--- a/bin/zero.js
+++ b/bin/zero.js
@@ -70,6 +70,65 @@ Flags:
 
 const EXIT_USAGE = 2;
 
+function installedByBun() {
+  if (process.env.ZERO_WRAPPER_SIMULATE_BUN === '1') {
+    return true;
+  }
+  return process.execPath.includes('bun') || !!process.versions?.bun;
+}
+
+function bunRecoveryParagraph() {
+  return (
+    'You installed with Bun, which does not run dependency lifecycle scripts\n' +
+    'by default. Trust the package to run the blocked postinstall:\n' +
+    '  bun pm trust @gitlawb/zero       (project install)\n' +
+    '  bun pm -g trust @gitlawb/zero    (global install)\n' +
+    'On Bun versions without `bun pm trust`, add\n' +
+    '  "trustedDependencies": ["@gitlawb/zero"]\n' +
+    'to your project package.json and reinstall.\n' +
+    '\n'
+  );
+}
+
+function buildFromSourceParagraph() {
+  return (
+    'If this platform has no prebuilt binary, build from source:\n' +
+    'https://github.com/Gitlawb/zero (go run ./cmd/zero, requires Go 1.26+).'
+  );
+}
+
+function missingBinaryContextParagraph() {
+  return (
+    'Normally npm installs it as an optional dependency of @gitlawb/zero\n' +
+    `(@gitlawb/zero-${process.platform}-${process.arch}), and the wrapper can\n` +
+    'also download it from the GitHub Release when it is missing.'
+  );
+}
+
+function missingNativeRecoveryParagraphs(postinstallScript) {
+  const ranByBun = installedByBun();
+  return (
+    missingBinaryContextParagraph() +
+    '\n' +
+    '\n' +
+    'Things to try:\n' +
+    '  - reinstall without omitting optional dependencies:\n' +
+    '      npm install -g @gitlawb/zero\n' +
+    '  - run the downloader manually (needs write access to the package dir):\n' +
+    `      node "${postinstallScript}"\n` +
+    '\n' +
+    (ranByBun ? bunRecoveryParagraph() : '') +
+    buildFromSourceParagraph()
+  );
+}
+
+function formatGenericMissingNativeBinaryMessage(postinstallScript) {
+  return (
+    '[zero] No native binary is available for this install.\n' +
+    missingNativeRecoveryParagraphs(postinstallScript)
+  );
+}
+
 function parseDoctorArgs(args) {
   let json = false;
   for (const arg of args) {
@@ -108,7 +167,7 @@ function missingNativeDoctorJSONReport(postinstallScript) {
   };
 }
 
-function missingNativeDoctorTextReport(postinstallScript, ranByBun) {
+function missingNativeDoctorTextReport(postinstallScript) {
   return (
     'Zero doctor report (' +
     new Date().toISOString() +
@@ -119,17 +178,7 @@ function missingNativeDoctorTextReport(postinstallScript, ranByBun) {
     postinstallScript +
     '"\n' +
     '\n' +
-    'The platform binary is fetched at install time by a postinstall script,\n' +
-    'which did not run (or was skipped) for this install.' +
-    (ranByBun
-      ? '\nYou installed with Bun, which does not run dependency lifecycle scripts\n' +
-        'by default. To let it run on future installs, add zero to your\n' +
-        "project's trustedDependencies:\n" +
-        '  "trustedDependencies": ["@gitlawb/zero"]\n' +
-        'then reinstall.'
-      : '') +
-    '\nIf reinstall fails, build from source: https://github.com/Gitlawb/zero\n' +
-    '(go run ./cmd/zero, requires Go 1.25+).'
+    missingNativeRecoveryParagraphs(postinstallScript)
   );
 }
 
@@ -193,7 +242,6 @@ const nativePath = resolveNativeBinary();
 
 if (!nativePath) {
   const postinstallScript = join(packageRoot, 'scripts', 'postinstall.mjs');
-  const ranByBun = process.execPath.includes('bun') || !!process.versions?.bun;
 
   // `zero doctor` is the diagnostic command: when the native binary is missing
   // it's the one invocation that must NOT bail with the generic wrapper error,
@@ -220,25 +268,11 @@ if (!nativePath) {
       process.exit(1);
     }
 
-    process.stdout.write(missingNativeDoctorTextReport(postinstallScript, ranByBun) + '\n');
+    process.stdout.write(missingNativeDoctorTextReport(postinstallScript) + '\n');
     process.exit(1);
   }
 
-  console.error(
-    '[zero] No native binary is available for this install.\n' +
-      'Normally npm installs it as an optional dependency of @gitlawb/zero\n' +
-      `(@gitlawb/zero-${process.platform}-${process.arch}), and the wrapper can\n` +
-      'also download it from the GitHub Release when it is missing.\n' +
-      '\n' +
-      'Things to try:\n' +
-      '  - reinstall without omitting optional dependencies:\n' +
-      '      npm install -g @gitlawb/zero\n' +
-      '  - run the downloader manually (needs write access to the package dir):\n' +
-      `      node "${join(packageRoot, 'scripts', 'postinstall.mjs')}"\n` +
-      '\n' +
-      'If this platform has no prebuilt binary, build from source:\n' +
-      'https://github.com/Gitlawb/zero (go run ./cmd/zero, requires Go 1.26+).',
-  );
+  console.error(formatGenericMissingNativeBinaryMessage(postinstallScript));
   process.exit(1);
 }
 
--- a/bin/zero.js
+++ b/bin/zero.js
@@ -55,6 +55,84 @@ function localControlHelperManifest(packageRoot) {
   return JSON.stringify({ version: 1, helpers });
 }
 
+// Mirrors internal/cli/observability.go parseDoctorArgs + writeDoctorHelp so the
+// wrapper's missing-binary doctor fallback matches the Go doctor CLI surface.
+const DOCTOR_HELP = `Usage:
+  zero doctor [flags]
+
+Runs Go backend health checks for config and provider setup.
+
+Flags:
+      --json            Print JSON report
+      --connectivity    Include provider endpoint connectivity probe when available
+  -h, --help            Show this help
+`;
+
+const EXIT_USAGE = 2;
+
+function parseDoctorArgs(args) {
+  let json = false;
+  for (const arg of args) {
+    switch (arg) {
+      case '-h':
+      case '--help':
+      case 'help':
+        return { kind: 'help' };
+      case '--json':
+        json = true;
+        break;
+      case '--connectivity':
+        break;
+      default:
+        return { kind: 'error', message: `unknown doctor flag ${JSON.stringify(arg)}` };
+    }
+  }
+  return { kind: 'run', json };
+}
+
+function missingNativeDoctorJSONReport(postinstallScript) {
+  return {
+    generatedAt: new Date().toISOString().replace(/\.\d{3}Z$/, 'Z'),
+    ok: false,
+    checks: [
+      {
+        id: 'runtime.go',
+        label: 'Go runtime',
+        status: 'fail',
+        message: 'Native zero binary is missing next to the npm wrapper.',
+        details: {
+          remedy: `node "${postinstallScript}"`,
+        },
+      },
+    ],
+  };
+}
+
+function missingNativeDoctorTextReport(postinstallScript, ranByBun) {
+  return (
+    'Zero doctor report (' +
+    new Date().toISOString() +
+    ')\n' +
+    'Overall: fail\n' +
+    '[fail] runtime.go - Native zero binary is missing next to the npm wrapper.\n' +
+    '  remedy: node "' +
+    postinstallScript +
+    '"\n' +
+    '\n' +
+    'The platform binary is fetched at install time by a postinstall script,\n' +
+    'which did not run (or was skipped) for this install.' +
+    (ranByBun
+      ? '\nYou installed with Bun, which does not run dependency lifecycle scripts\n' +
+        'by default. To let it run on future installs, add zero to your\n' +
+        "project's trustedDependencies:\n" +
+        '  "trustedDependencies": ["@gitlawb/zero"]\n' +
+        'then reinstall.'
+      : '') +
+    '\nIf reinstall fails, build from source: https://github.com/Gitlawb/zero\n' +
+    '(go run ./cmd/zero, requires Go 1.25+).'
+  );
+}
+
 // The platform payload is a version of @gitlawb/zero itself, installed under
 // an npm: alias (see docs/NPM_PACKAGING.md). The alias name is derived from
 // process.platform/process.arch, so an unsupported platform simply fails to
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (changes requested):
## Request changes — right instinct, but this fixes the wrong layer (and CI is red)

Thanks @michaelkillgta — #405 is a real papercut (doctor pointing users away from the actual cause), and the PR is thoughtful: clear comments, a fail-closed default, four focused tests. But as designed I don't think it fixes the reported bug, and it's currently failing CI.

### 1. The check can't observe the condition it targets (blocker)
The npm wrapper `bin/zero.js` (unchanged by this PR; guard already on `main` via #418) checks `existsSync(nativePath)` for **every** subcommand — including `doctor` — and on a missing binary prints *"No native binary found"* and `exit 1` **before it ever spawns the Go binary**. So:
- In #405's exact repro (`ZERO_SKIP_DOWNLOAD=1 … zero doctor`), the Go doctor code never runs — the new `StatusFail` branch can't fire.
- When Go doctor *does* run, `os.Executable()` is the process executing it, so `os.Stat` on it is a tautology → always `StatusPass`.
- The only reachable non-pass is `StatusWarn` when `os.Executable()` itself errors — not the reported condition.

So a Go-side `runtimeCheck` architecturally can't see "the native binary is missing" (it can't be running when the binary is absent). The fix belongs in the **JS wrapper's `doctor` path** — e.g. a preflight, or have the wrapper pass doctor the *expected* `nativePath` so it can stat a path other than its own running image. If I'm missing a flow where `zero doctor` reaches Go with the binary absent, point me at it.

### 2. CI red — not gofmt-clean (blocker)
`Smoke (ubuntu-latest) → Check formatting` fails; I reproduced it — `gofmt -l` flags all four changed files (misaligned `Details` map literals in `doctor.go`). `gofmt -w` the four. (Windows/macOS smoke passing is misleading — the format gate is ubuntu-only.)

### Minor
The `os.Executable()`→empty-on-error block is duplicated in `observability.go` and `model.go`, and no test covers the real call-site wiring (every test injects the fields).

Verified locally on Windows: build clean, `internal/doctor` tests pass — but they validate a mechanism that doesn't map to the real failure mode. Happy to re-review once detection moves to the wrapper layer and CI's green.

gnanam1990 (changes requested):
## VERDICT: needs-changes

## REGRESSION RISK: medium

The change replaces the old unconditional `StatusPass` in `runtimeCheck` with a real `os.Stat` of the running binary path. The intent is correct (detect a missing native binary), but the implementation cannot observe the condition it targets: when the native binary is missing, `bin/zero.js` exits with "No native binary found" before spawning the Go process, so `zero doctor` never reaches the Go code. When Go doctor does run, `os.Executable()` returns the path of the currently executing binary, so `os.Stat` on it is a tautology that always passes. The only reachable non-pass state is `StatusWarn` when `os.Executable()` itself errors — which is not the reported condition (#405). The fix belongs in the JS wrapper's doctor path (e.g. have the wrapper pass the expected `nativePath` so Go doctor can stat a path other than its own running image), or a preflight check in the wrapper itself.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/doctor/... ./internal/cli/...` — pass
- `gofmt -l` on all four changed files — clean (the "style: run gofmt" commit addressed the gofmt blocker from the prior review)
- `go test ./internal/doctor/... -run TestRuntimeCheck` — all four tests pass, but they validate a mechanism that does not map to the real failure mode (every test injects the fields rather than exercising the real call-site wiring)
- CI status for the latest commit (`b527640`) could not be confirmed — the check-runs API returned empty for this commit, suggesting CI may not have re-run after the latest push. The prior commit's CI was red on the Ubuntu formatting gate, which is now fixed.

## CONTRIBUTING

Author is a community contributor (michaelkillgta). The PR links `Fixes #405`, which has a `bug` label but no `issue-approved` label. Per CONTRIBUTING.md, the `issue-approved` gate applies to community PRs. Not met. Process violation, but reviewing the code regardless as instructed.

## FINDINGS

1. **Blocker — `internal/doctor/doctor.go:127-170` — the check cannot observe the condition it targets.** `os.Executable()` returns the path of the running Go binary, which by definition exists (it is executing). `os.Stat` on it is always `StatusPass`. When the binary is actually missing, the JS wrapper (`bin/zero.js`, unchanged by this PR) exits before the Go process starts, so the new `StatusFail` branch is unreachable for the reported scenario. The fix needs to either (a) move detection to the JS wrapper's doctor path, or (b) have the wrapper pass the expected `nativePath` to the Go doctor so it can stat a path other than its own image.

2. **Minor — `internal/cli/observability.go:67-70` and `internal/tui/model.go:777-790` — duplicated `os.Executable()` resolution.** The 4-line "resolve exe path, empty on error" block is duplicated across CLI and TUI call sites. Consider a shared helper. No test covers the real call-site wiring (every test injects the fields).

## EXISTING REVIEWS

- **Vasanthdev2004** (CHANGES_REQUESTED on `81d37c5`): Identified both blockers — (1) the check being a tautology / unreachable for the reported condition, and (2) CI red from gofmt. Two new commits have since been pushed (`c1dd620` "style: remove trailing blank line" and `b527640` "style: run gofmt"). Blocker #2 (gofmt) is now resolved — verified clean on the current head. Blocker #1 (architectural) is NOT addressed — `bin/zero.js` is not in the changed files, and the `os.Executable()` tautology remains. The review is partially stale (gofmt fixed) but the primary blocker is still valid on the current head.
- **coderabbitai** (APPROVED + 2 nitpicks): Flagged the duplicated `os.Executable()` resolution. Valid, trivial. The approval did not catch the architectural blocker — the check cannot observe the missing-binary condition because the JS wrapper exits first.

## BOTTOM LINE

The gofmt blocker is fixed, but the architectural blocker remains: a Go-side `os.Executable()` stat cannot detect a missing native binary because the binary is either running (tautology) or absent (Go never starts); the fix needs to move to the JS wrapper layer.

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to preserve `doctor --json`
  `bin/zero.js:68`
  CodeRabbit's latest review thread is still unresolved and still applies: the new missing-binary `doctor` branch treats every `zero doctor ...` invocation as text output. The Go command already supports `zero doctor --json` and writes the report as JSON, but with the native binary missing this wrapper path now exits before Go starts and prints the human-readable report to stderr with empty stdout. That breaks scripts or CI using the machine-readable doctor mode exactly when diagnostics are needed most. Please either emit a JSON-shaped failure for `doctor --json` in this fallback or reject JSON mode explicitly, and add wrapper coverage for that flag.

- [P3] Update the PR description to match the wrapper-layer fix
  `PR description`
  The current PR body still describes the older Go-side approach: it says the solution adds `ExecutablePath` and `StatExecutable` to `doctor.Options`, changes `runtimeCheck`, and adds `TestRuntimeCheck...` coverage. The current patch no longer touches those files; it only changes `bin/zero.js` and `internal/npmwrapper/npmwrapper_test.go`. Since prior review rejected the Go-side approach as the wrong layer, leaving that stale description and test plan makes it hard to tell what behavior this PR is actually asking maintainers to merge. Please update the summary, root cause, solution, and test plan to describe the current wrapper fallback.

jatmn (changes requested):
Thanks for the update. I rechecked the current wrapper-layer patch and found a couple of issues that need to be addressed before this is ready.

## Findings

- [P2] Keep the text doctor report on stdout
  `bin/zero.js:95`
  The new missing-binary `zero doctor` fallback writes the doctor-shaped text report with `console.error`, so `zero doctor > report.txt` captures an empty file even though the normal Go doctor path writes `doctor.Format(report)` to stdout and uses the exit code to signal failure. The new regression test uses `CombinedOutput`, so it cannot catch this stream drift. Please write the text doctor report to stdout, like the JSON fallback and the regular doctor command do, while preserving exit code 1.

- [P3] Preserve doctor help and flag parsing in the wrapper fallback
  `bin/zero.js:68`
  The new branch treats every first-argument `doctor` invocation as a runtime diagnostic, so `zero doctor --help` and `zero doctor help` now return the missing-binary failure report instead of the doctor usage text, and `zero doctor --json --bogus` returns a runtime JSON report instead of the normal unknown-flag error. The Go `runDoctor` path handles help before running checks and rejects unknown flags, so the missing-binary wrapper should keep those contracts for `doctor` rather than treating help and invalid invocations as real diagnostics.

- [P3] Resolve the contribution-policy gate for the linked issue
  `CONTRIBUTING.md`
  This PR links `Fixes #405`, but that issue currently has no `issue-approved` label. `CONTRIBUTING.md` says community pull requests need an approved parent issue and that approval is shown by the `issue-approved` label. Please get the parent issue approved/labeled or have a maintainer explicitly waive that gate before continuing the code review cycle.

- [P3] Update the PR description to match the wrapper-layer fix
  `PR description`
  The current PR body still describes the older Go-side approach: adding `ExecutablePath` and `StatExecutable` to `doctor.Options`, changing `runtimeCheck`, and adding `TestRuntimeCheck...` coverage. The current patch only changes `bin/zero.js` and `internal/npmwrapper/npmwrapper_test.go`, and prior review rejected the Go-side approach as the wrong layer. Please update the summary, root cause, solution, and test plan so the PR describes the wrapper fallback that is actually being reviewed.

michaelkillgta:
Hi! I've addressed the latest review comments and updated the implementation to preserve CLI behavior (stdout for doctor reports, help/invalid-flag handling, JSON support). I also noticed the mention of the \issue-approved\ requirement. Could a maintainer let me know whether issue #405 should receive that label for this PR, or whether you'd prefer a different process?

jatmn (changes requested):
Thanks for the update. I rechecked the current wrapper-layer patch and found a couple of issues that still need to be addressed before this is ready.

## Findings

- [P1] Fix the failing Ubuntu smoke formatting check
  `internal/npmwrapper/npmwrapper_test.go:1`
  The current `Smoke (ubuntu-latest)` check is failing in the `Check formatting` step, and the job log reports `gofmt needed on: internal/npmwrapper/npmwrapper_test.go`. This is a changed file in this PR, so please run `gofmt` on the test file and push the formatting fix before the code/test results can be treated as clean.

- [P3] Resolve the contribution-policy gate for the linked issue
  `CONTRIBUTING.md:19`
  This PR links `Fixes #405`, but that issue still has no `issue-approved` label. `CONTRIBUTING.md` says community pull requests need an approved parent issue and that approval is shown by the `issue-approved` label. Please get the parent issue approved/labeled or have a maintainer explicitly waive that gate before continuing the code review cycle.

michaelkillgta:
Thanks for the re-review.

**P1 (gofmt):** Addressed in \3082dfc\ (\style: gofmt npmwrapper doctor fallback tests\). \gofmt -l internal/npmwrapper/npmwrapper_test.go\ is clean locally. The Ubuntu Smoke failure you saw should be resolved on that commit — could you re-run **Smoke (ubuntu-latest)** when you have a moment? (Fork workflow runs may need approval before checks appear on the latest push.)

**P3 (issue-approved):** Understood. Issue #405 still doesn't have the \issue-approved\ label. I'm happy to pause the code-review cycle until a maintainer labels #405 or explicitly waives the gate per \CONTRIBUTING.md\. Please let me know which you prefer.

jatmn (changes requested):
Thanks for the update. I rechecked the current wrapper-layer patch and found a couple of issues that still need to be addressed before this is ready.

## Findings

- [P2] Preserve the Bun trust-command recovery path in the doctor fallback
  `bin/zero.js:119`
  The new missing-binary `zero doctor` fallback adds its own Bun-specific remedy, but it only tells Bun users to add `trustedDependencies`. Current `main` already updated the wrapper's missing-binary guidance to lead with `bun pm trust @gitlawb/zero` and `bun pm -g trust @gitlawb/zero`, with `trustedDependencies` only as the fallback for older Bun versions. This means Bun users who run the diagnostic command now get weaker, stale recovery advice than the generic missing-binary wrapper path. Please keep the new doctor-shaped report aligned with the current Bun guidance and add coverage for the Bun branch so it does not drift again.

- [P3] Resolve the contribution-policy gate for the linked issue
  `CONTRIBUTING.md:19`
  This PR still links `Fixes #405`, but that issue has no `issue-approved` label. `CONTRIBUTING.md` says community pull requests need an approved parent issue and that approval is shown by the `issue-approved` label. Please get the parent issue approved/labeled or have a maintainer explicitly waive that gate before continuing the code review cycle.

jatmn (changes requested):
Thanks for the update. I rechecked the current wrapper-layer patch and found a few issues that still need to be addressed before this is ready.

## Findings

- [P1] Rebase on main and resolve the wrapper conflict first
  `GitHub mergeability`
  GitHub currently reports this branch as `CONFLICTING` / `DIRTY`, and the conflict is in `bin/zero.js`, which is the file this PR changes. Please rebase on the latest `main` branch and resolve that conflict before addressing the review findings below, so the missing-binary wrapper behavior is reviewed against the code that will actually be merged.

- [P2] Preserve the current Bun trust-command recovery guidance
  `bin/zero.js:74`
  The current PR head still builds the shared Bun recovery copy around `trustedDependencies` only, but current `main` already changed the missing-binary wrapper guidance to lead with `bun pm trust @gitlawb/zero` and `bun pm -g trust @gitlawb/zero`, with `trustedDependencies` as the fallback for older Bun versions. After rebasing, please keep both the generic missing-binary path and the new `zero doctor` fallback on that newer `bun pm trust` guidance, and keep the Bun regression coverage pointed at that current copy so the doctor path cannot drift again.

- [P3] Resolve the contribution-policy gate for the linked issue
  `CONTRIBUTING.md:19`
  This PR still links `Fixes #405`, but that issue has no `issue-approved` label. `CONTRIBUTING.md` says community pull requests need an approved parent issue and that approval is shown by the `issue-approved` label. Please get the parent issue approved/labeled or have a maintainer explicitly waive that gate before continuing the code review cycle.

jatmn (dismissed):
@Vasanthdev2004 

Thanks for the update. I rechecked the current wrapper-layer patch and found one issue that still needs to be addressed before this is ready.

## Findings

- [P3] Resolve the contribution-policy gate for the linked issue
  `CONTRIBUTING.md:19`
  This PR still links `Fixes #405`, but issue #405 has no `issue-approved` label. `CONTRIBUTING.md` says community pull requests need an approved parent issue and that approval is shown by the `issue-approved` label. Please get the parent issue approved/labeled or have a maintainer explicitly waive that gate before this is treated as ready.

michaelkillgta:
Hi! My PR (#450) is now approved from a technical standpoint. The remaining blocker appears to be the issue-approved policy requirement. Would a maintainer be able to add the issue-approved label to #405, or let me know if this requirement can be waived for this contribution? Thanks!

Vasanthdev2004 (approved):
Re-reviewing — my earlier request-changes was that this fixed the wrong layer, and that's resolved. The whole fix now lives in `bin/zero.js`, which is the only place that can actually observe a missing binary: the wrapper exits before the Go process ever starts, so nothing in the Go doctor could see it. The wrapper now special-cases `zero doctor` to print a doctor-shaped FAIL line for the runtime instead of the generic "no native binary" bail — which is exactly what #405 needed, since the complaint was that the one diagnostic command pointed users away from the real cause.

The stream handling is right, which is the thing that kept coming up: both the text report and the `--json` report go to **stdout** with exit 1 (so `zero doctor > report.txt` and `--json | jq` both work), unknown flags go to stderr with exit 2, and every non-doctor invocation keeps the original bail untouched. The Bun recovery leads with `bun pm trust`, matching main, and the test fixture captures stdout and stderr separately now, so it can catch the stream drift the old `CombinedOutput` test couldn't.

Marked #405 `issue-approved` — a doctor that greenlights a broken install is a real papercut, and this is the right fix. CI is green and it's mergeable.

Vasanthdev2004 (dismissed):
I went through the diff for issue #405. The doctor command now checks for the native sandbox/helper binaries and reports a clear missing-binary diagnosis instead of letting a later step fail with a generic error. The detection is guarded so a missing binary degrades the report rather than crashing it. Approving.

gnanam1990 (dismissed):
## Re-review: APPROVE

Thanks for the follow-up commits — the architectural concern from my earlier review is fully addressed.

### Prior blocker: resolved
The original Go-side `runtimeCheck` (`os.Executable()` + `os.Stat`) could never observe a missing npm native binary. That approach is gone.

The fix now lives where it must — in `bin/zero.js` at the `existsSync(nativePath)` guard. For `zero doctor` only, the wrapper emits a doctor-shaped FAIL for `runtime.go` on stdout instead of the generic stderr bail. That directly fixes #405's blind-alley.

### Verified
- `zero doctor` / `--json` / `--help` / unknown-flag paths match Go doctor surface
- Non-doctor commands unchanged (generic missing-binary stderr)
- `ZERO_WRAPPER_SIMULATE_BUN` regression tests pass
- CI green on `6a26a74`

### Nit (non-blocking)
Wrapper doctor failure exits 1 while Go doctor returns 3 on report failure — fine for typical usage.

Good to merge.

kevincodex1:
hello @michaelkillgta please rebase to main and fix conflicts

Vasanthdev2004 (commented):
Re-checked after the rebase — everything I approved before is intact, and the conflict resolution is right. The whole fix still lives in bin/zero.js, the doctor fallback now sits correctly after the new platform-package/auto-download resolution that landed on main, the non-Bun generic message stays byte-for-byte what main prints today, and you picked up main's new build-from-source copy in the test constants. JSON shape still matches doctor.Report and the flag parsing still mirrors the Go side, so nothing to re-litigate on the code. The only reason I'm not re-stamping the approval right now is that the workflows on this push are sitting in the fork-approval queue — Smoke/Performance/Security haven't actually run on df89ca5 yet. Once they're waved through and come back green I'll re-approve; no further changes needed from you. Thanks for turning kevin's rebase ask around quickly.

kevincodex1 (approved):
LGTM

PR #453, opened 2026-07-03
fix(agent): reject a malformed additional_permissions payload before prompting

euxaristia:
## What

A shell tool call (`bash`/`exec_command`) that sets `sandbox_permissions: "with_additional_permissions"` but omits, or malforms, the `additional_permissions` payload was accepted into the normal permission-prompt flow. The user saw an ordinary-looking approval request, approved it, and only then hit a hard denial from the same validation running a second time at grant time, with the message `failed to grant session additional permissions: missing additional_permissions`.

## Why this happens

`buildPermissionEvent` (`internal/agent/loop.go`) calls `inlineAdditionalPermissionsProfile` to build the prompt's scope text, but only uses the result `if ok && err == nil`, silently discarding the exact validation error that later kills the grant in `grantInlineAdditionalPermissions`. So by the time the prompt is shown, the request is already known to be unsatisfiable, but nothing communicates that to the user.

This is most visible with models that set the `sandbox_permissions` flag inconsistently (I hit it repeatedly with a driving model that set the flag without ever including the `additional_permissions` object), producing a loop of approve-then-deny with no way to make progress on the shell command.

## Fix

`inlineAdditionalPermissionsProfile` is already the single source of truth for whether this payload is well-formed (it also catches the mismatched case: `additional_permissions` present without the flag). Run that same check up front, before any permission prompt is shown, and return a plain tool error immediately if it fails. A malformed payload can never succeed regardless of what the user decides, so:

- The user is never shown a prompt that cannot lead anywhere.
- The driving model gets a direct, actionable error it can see in the transcript and retry from with a corrected tool call.

The change is additive: one early-exit check before the existing `ask_user`/`request_permissions` interception, reusing existing validation logic with no changes to the successful path.

## Testing

Three new tests in `internal/agent/additional_permissions_validation_test.go`, all exercising `executeToolCall` directly and asserting on `OnPermissionRequest` call counts:

- Missing `additional_permissions` with the flag set: rejected, prompt never fires.
- `additional_permissions` present without the flag: rejected, prompt never fires.
- A valid payload: still reaches the prompt exactly once (confirms the fix is scoped to malformed input only).

`go build ./...`, `go vet ./internal/agent/...` clean. Full `internal/agent` suite passes; I confirmed the 4 unrelated failures in that suite (`TestRunPromptsAndAllowsOutsideWorkspaceWrite`, `TestRunSessionAllowsLaterOutsideWorkspaceWrite`, `TestRunAppliesSandboxEvenInUnsafeMode`, `TestRequestPermissionsTurnGrantAllowsLaterToolAndCleansUp`) are pre-existing on unmodified `main` in my environment and untouched by this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added early validation for permission-related tool calls, so malformed `with_additional_permissions` requests fail immediately instead of showing the standard approval prompt.
  * Improved error messaging with clearer, actionable guidance (including correctly-shaped `additional_permissions` examples) when required fields are missing or normalize to an empty profile.
  * Verified that well-formed permission requests still follow the expected approval flow.

### DIFF
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/Gitlawb/zero/internal/tools"
@@ -77,6 +78,37 @@ func TestExecuteToolCallRejectsAdditionalPermissionsWithoutFlagBeforePrompt(t *t
 	}
 	if promptCalls != 0 {
 		t.Fatalf("OnPermissionRequest called %d times, want 0 (must reject before prompting)", promptCalls)
+	}
+}
+
+// The rejection message must show a concrete, correctly-shaped example: a
+// model that gets this wrong once (attaching sandbox_permissions out of habit
+// from an earlier call, without a valid payload) needs enough in the error to
+// self-correct on retry instead of repeating the identical mistake.
+func TestExecuteToolCallMissingAdditionalPermissionsErrorIsActionable(t *testing.T) {
+	root := t.TempDir()
+	registry := tools.NewRegistry()
+	registry.Register(tools.NewBashTool(root))
+
+	result, err := executeToolCall(
+		context.Background(),
+		registry,
+		ToolCall{
+			ID:        "c1",
+			Name:      "bash",
+			Arguments: `{"command":"ping example.com","sandbox_permissions":"with_additional_permissions"}`,
+		},
+		PermissionModeAuto,
+		Options{},
+	)
+	if err != nil {
+		t.Fatalf("executeToolCall: %v", err)
+	}
+	if !strings.Contains(result.Output, `{"network": {"enabled": true}}`) {
+		t.Fatalf("output = %q, want a concrete additional_permissions example", result.Output)
+	}
+	if !strings.Contains(result.Output, "omit sandbox_permissions") {
+		t.Fatalf("output = %q, want guidance that the flag can simply be omitted", result.Output)
 	}
 }
 
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -1,0 +1,114 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
+// A shell call that sets sandbox_permissions: with_additional_permissions but
+// omits (or malforms) additional_permissions can never succeed: the same
+// validation runs again at grant time regardless of the user's decision. It
+// must be rejected as a plain tool error BEFORE any permission prompt, not
+// surfaced as a confusing "approved but denied" prompt outcome.
+func TestExecuteToolCallRejectsMissingAdditionalPermissionsBeforePrompt(t *testing.T) {
+	root := t.TempDir()
+	registry := tools.NewRegistry()
+	registry.Register(tools.NewBashTool(root))
+
+	promptCalls := 0
+	result, err := executeToolCall(
+		context.Background(),
+		registry,
+		ToolCall{
+			ID:        "c1",
+			Name:      "bash",
+			Arguments: `{"command":"echo hi","sandbox_permissions":"with_additional_permissions"}`,
+		},
+		PermissionModeAuto,
+		Options{
+			OnPermissionRequest: func(ctx context.Context, request PermissionRequest) (PermissionDecision, error) {
+				promptCalls++
+				return PermissionDecision{Action: PermissionDecisionAllow}, nil
+			},
+		},
+	)
+	if err != nil {
+		t.Fatalf("executeToolCall: %v", err)
+	}
+	if result.Status != tools.StatusError {
+		t.Fatalf("status = %q, want error", result.Status)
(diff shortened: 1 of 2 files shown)


euxaristia:
Pushed a follow-up commit: made the rejection message actionable with a concrete, schema-accurate example, instead of just naming the missing field.

Context for why: I watched this happen live in a real session today. A model kept re-attaching sandbox_permissions from an earlier, correct gh/git call onto later commands that did not need it (network was already open via a session-scoped grant), hit the original rejection 4 times in a row with no change in its retry, and gave up. The old message named what was missing but gave nothing concrete to fix; the new one shows the exact JSON shape and also tells the model it can just omit sandbox_permissions if elevation is not actually needed.

Vasanthdev2004 (approved):
## Approve — well-scoped bug fix, fails closed

**Worth it:** yes. This closes a real approve-then-deny dead-end: a shell call setting `sandbox_permissions: with_additional_permissions` but omitting/malforming `additional_permissions` used to sail into the normal approval prompt — the user approved an ordinary-looking request and only *then* hit a hard denial (`failed to grant session additional permissions: missing additional_permissions`) from the same validation re-running at grant time. Since a malformed payload can never be satisfied regardless of the user's choice, prompting at all is the bug. Thanks @euxaristia — good diagnosis.

The fix runs the existing single-source-of-truth validator (`inlineAdditionalPermissionsProfile`) up front and returns a plain, actionable tool error the model can self-correct from — before any prompt. It's additive (28 LOC in `loop.go`), gated to `bash`/`exec_command`, and leaves the successful path untouched.

**Key regression question — can the up-front check disagree with the grant-time check?** It can't: the new check, `buildPermissionEvent`, and `grantInlineAdditionalPermissions` all call `inlineAdditionalPermissionsProfile(args, options.Cwd)` with the same base path, so the early error is exactly the one that would otherwise kill the grant. And a valid payload still reaches the prompt exactly once (guarded by the `StillPromptsForValid` test).

**Verified on Windows:** `go build ./...`, `go vet ./internal/agent/...` clean; the four new tests pass (rejects missing payload before prompt; rejects `additional_permissions`-without-flag; actionable error message; valid payload still prompts once). The tests the author flagged as failing in their env all pass here — env-specific, not regressions. CI green incl. Windows smoke.

**Non-blocking note:** the check sits before the `permissionGranted` logic, so a malformed elevation payload now also errors on paths that previously ran the command silently (unsafe mode, prefix-approved, the `AllowForSession` mismatched-flag case). In every one of those the payload was a no-op or already headed for denial, so it's sensible hardening of malformed input, not a regression. Approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change adds an early-exit validation check in `executeToolCall` (`internal/agent/loop.go:819-835`) that runs `inlineAdditionalPermissionsProfile` before any permission prompt is shown for shell commands with `sandbox_permissions: "with_additional_permissions"`. If the payload is missing or malformed, the call is rejected immediately as a tool error instead of presenting a prompt that cannot succeed. This is purely additive — the existing successful path (valid payload → prompt → grant) is unchanged. The error messages in `inlineAdditionalPermissionsProfile` are also improved to include concrete examples (`{"network": {"enabled": true}}`), making them actionable for the model to retry. No existing code path, permission flow, or user interaction is removed or altered.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/agent/...` — pass
- `go test ./internal/agent/... -run TestExecuteToolCall.*AdditionalPermissions` — all four tests pass:
  - `RejectsMissingAdditionalPermissionsBeforePrompt`: flag set, no payload → rejected, prompt never fires
  - `RejectsAdditionalPermissionsWithoutFlagBeforePrompt`: payload present, no flag → rejected, prompt never fires
  - `MissingAdditionalPermissionsErrorIsActionable`: error message includes a concrete example and "omit sandbox_permissions" guidance
  - `StillPromptsForValidAdditionalPermissions`: valid payload → prompt fires exactly once (no false positive)

## CONTRIBUTING

No linked issue. The author is a community contributor (euxaristia). Per CONTRIBUTING.md, community PRs require an approved parent issue with the `issue-approved` label. No such issue is linked. Process violation, but the change is a legitimate bug fix with comprehensive tests. Maintainer's call.

## FINDINGS

None. The early-exit check correctly guards only shell commands (`isShellCommandTool`), reuses the existing validation function, and the test suite covers both rejection cases, the actionable error message, and the valid-payload pass-through.

## EXISTING REVIEWS

- **Vasanthdev2004** (APPROVED): Assessed as a well-scoped bug fix that fails closed. Valid on the current head. Nothing missed.
- **coderabbitai** (APPROVED): No issues. Nothing missed.

All reviews on the current head (`65146b3`). No stale reviews.

## BOTTOM LINE

A clean, additive validation guard that prevents an unsatisfiable approve-then-deny loop; safe to merge.

euxaristia:
Thanks for the positive review @gnanam1990 🍾 

I think the requirement to file an issue is a good idea in general but might be an impediment to rapid iteration in such a new project where community contributors such as myself and maintainers such as yourself are discovering bugs at an accelerated rate; and patching them with coderabbitai's approval.

kevincodex1 (approved):
looks good!

PR #460, opened 2026-07-03
fix: preserve conversation context in exec prompts

pengdst:
## Summary
Fixes exec/session prompt formatting so older conversation turns are preserved instead of being dropped when many later context events exist.

## What changed
- Preserve the latest event window.
- Prepend earlier conversation events when available.
- Add regression coverage for long exec/session histories.

## Tests
- go test ./internal/sessions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resumed session prompts to keep the most relevant conversation messages, even when unrelated/noisy events appear afterward.
  * Reduced prompt noise by filtering retained context to allowed event types and capping the context to a larger, recent window.

* **Tests**
  * Added tests to verify resume-mode prompt generation preserves conversation content while omitting noisy tool/result outputs, including correct truncation behavior with many events.

### DIFF
--- a/internal/sessions/exec_session.go
+++ b/internal/sessions/exec_session.go
@@ -161,9 +161,6 @@ func FormatExecPrompt(prompt string, prepared PreparedExec) string {
 		return prompt
 	}
 	events := promptContextEvents(prepared.ContextEvents)
-	if len(events) == 0 {
-		return prompt
-	}
 
 	lines := []string{}
 	for _, event := range events {
--- a/internal/sessions/exec_session.go
+++ b/internal/sessions/exec_session.go
@@ -160,9 +160,9 @@ func FormatExecPrompt(prompt string, prepared PreparedExec) string {
 	if prepared.Mode == ModeNew || len(prepared.ContextEvents) == 0 {
 		return prompt
 	}
-	events := prepared.ContextEvents
-	if len(events) > 20 {
-		events = events[len(events)-20:]
+	events := promptContextEvents(prepared.ContextEvents)
+	if len(events) == 0 {
+		return prompt
 	}
 
 	lines := []string{}
@@ -185,6 +185,25 @@ func FormatExecPrompt(prompt string, prepared PreparedExec) string {
 		"Current user request:",
 		prompt,
 	}, "\n")
+}
+
+func promptContextEvents(events []Event) []Event {
+	const maxPromptContextEvents = 80
+
+	filtered := make([]Event, 0, len(events))
+	for _, event := range events {
+		switch event.Type {
+		case EventMessage, EventCompaction, EventSessionFork, EventSessionChild, EventSpecialistStart, EventSpecialistStop, EventError:
+			filtered = append(filtered, event)
+		}
+	}
+	if len(filtered) == 0 {
+		filtered = append(filtered, events...)
+	}
+	if len(filtered) > maxPromptContextEvents {
+		filtered = filtered[len(filtered)-maxPromptContextEvents:]
+	}
+	return filtered
 }
 
 func forkTitle(title string) string {
--- a/internal/sessions/store_test.go
+++ b/internal/sessions/store_test.go
@@ -555,6 +555,32 @@ func TestPrepareExecSessionResolvesResumeAndFork(t *testing.T) {
 	}
 }
 
+func TestFormatExecPromptKeepsConversationMessagesWhenNoisyEventsFollow(t *testing.T) {
+	events := []Event{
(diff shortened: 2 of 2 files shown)


kevincodex1:
hi @pengdst please address coderabbit comments

pengdst:
@kevincodex1 done 👍

Vasanthdev2004 (approved):
## Approve — genuine bug fix with regression coverage

**Worth it:** yes. `FormatExecPrompt` previously kept only the last 20 events (`events[len(events)-20:]`) regardless of type, so when a resumed or forked `zero exec` session had 20+ trailing noisy events (tool_result / tool_call / permission / usage), the window filled with noise and **every actual conversation turn was dropped** from the "Previous session context" block. The fix filters to high-signal event types before capping, restoring the lost context. Small (+105/−4), self-contained to `internal/sessions`. Thanks @pengdst.

**Verified on Windows** (worktree): `go build ./internal/sessions/...` clean; `go test ./internal/sessions/` all pass (incl. the two new regression tests); gofmt clean. Confirmed all whitelisted `Event` constants exist in `store.go`, and that `ModeNew` still early-returns the prompt unchanged — so single-shot `zero exec` is untouched; only resume/fork paths change.

**Design notes (non-blocking):**
- The context cap grew 20 → 80 (≈40KB worst case at the 500-char/line cap), but the type-filter strips the bulk of the noise, so real prompts are typically *smaller* than before. Reasonable.
- The all-noise fallback sensibly keeps the old behavior for sessions with no high-signal events.
- The whitelist excludes all tool_call/tool_result, so a session that ended right after a tool_result the agent hadn't responded to won't surface that output in the resume prompt — a reasonable tradeoff (messages carry the intent).

CodeRabbit's two nitpicks don't apply to the final diff (the "dead code" guard isn't present; the "80-cap coverage" gap was closed by the added truncation test — which is why CodeRabbit later approved). CI is green including `Smoke (windows-latest)`. Approving.

gnanam1990 (approved):
## VERDICT: approve

## REGRESSION RISK: low

The change replaces the old "last 20 events regardless of type" truncation in `FormatExecPrompt` with a new `promptContextEvents` function that filters to conversation-relevant event types (EventMessage, EventCompaction, EventSessionFork, EventSessionChild, EventSpecialistStart, EventSpecialistStop, EventError) and caps at 80. This alters the content of resumed/forked exec session prompts for existing users — they will now see more conversation context and less tool-result noise. This is an improvement to the prompt quality, not a breaking change: the prompt format (`## Previous session context` + event lines + `## Current user request`) is unchanged, and the function returns early for `ModeNew` or empty context as before. The fallback (`if len(filtered) == 0 { filtered = append(filtered, events...) }`) preserves all events when none match the filter, so sessions with only non-conversation events are not silently empty.

## BUILD / TEST

- `go build ./...` — pass
- `go vet ./internal/sessions/...` — pass
- `go test ./internal/sessions/... -run TestFormatExecPrompt` — both new tests pass:
  - `TestFormatExecPromptKeepsConversationMessagesWhenNoisyEventsFollow`: verifies conversation messages survive when followed by 43 tool-result events
  - `TestFormatExecPromptTruncatesConversationMessagesAfterFilteringNoise`: verifies the 80-event cap keeps the most recent conversation events and drops noise

## CONTRIBUTING

No linked issue. The author is a community contributor (pengdst). Per CONTRIBUTING.md, community PRs require an approved parent issue with the `issue-approved` label. No such issue is linked. Process violation, but the change is a legitimate bug fix with regression tests. Maintainer's call.

## FINDINGS

1. **Nit — `internal/sessions/exec_session.go:184-201` — fallback path could exceed the cap.** When no events match the filter, the fallback appends all events, then the cap is applied. This is correct behavior but means a session with only tool results will still include up to 80 of them. Not a blocker — this preserves existing behavior for unusual sessions.

No other findings.

## EXISTING REVIEWS

- **Vasanthdev2004** (APPROVED): Assessed as a genuine bug fix with regression coverage. Valid on the current head. Nothing missed.
- **coderabbitai** (APPROVED + 2 nitpicks): Minor maintainability suggestions on the filter list. Valid, trivial. Nothing critical missed.
- **copilot-pull-request-reviewer** (COMMENTED): Accurate overview. Nothing missed.

All reviews on the current head (`54c3259`). No stale reviews.

## BOTTOM LINE

A solid bug fix that preserves conversation context in resumed exec sessions instead of dropping it behind tool-result noise; safe to merge.

kevincodex1 (approved):
LGTM

Issue #415, opened 2026-07-02
python -c with nested quotes can get mangled by cmd.exe on Windows


PierrunoYT:
## Bug description

On Windows, a `python -c "..."` (or similar `-c`/`-e` inline-code) command with nested quotes inside the code string can fail with a Python-side syntax error that indicates cmd.exe's quote parsing corrupted the argument before Python ever saw it.

## Reproduction (observed in the wild)

A command shaped like:
```
py -c "import py_compile; py_compile.compile('pinokio_compat.py', doraise=True); print('OK')"
```
failed with:
```
File "<string>", line 1
  "import
  ^
SyntaxError: unterminated string literal (detected at line 1)
```
Note the Python source string Python actually received starts with a literal `"` character — meaning cmd.exe's argument parsing did not consume the outer double quotes the way a POSIX shell would for `-c "..."` containing inner single-quoted strings, parens, and commas.

## Why this isn't fixed yet

Unlike the `head`/`grep`/pager issues already fixed in #412/#414 (a simple "this command/utility doesn't exist on cmd.exe" pattern match), this is a genuine quoting/escaping mismatch between how a model writes a `-c` invocation (assuming bash-style quote nesting) and how cmd.exe's argument parser actually behaves. A reliable pre-flight detector for "this quoting will get mangled by cmd.exe" is a much harder, higher-false-positive-risk problem than a simple command-name pattern match, so it needs more careful design (or possibly better tool-description guidance telling the model to avoid inline `-c`/`-e` code with nested quotes on Windows, in favor of writing a temp script file and running that) rather than a quick heuristic.

## Suggested directions (not evaluated in depth)

- Tool-description guidance: on Windows, recommend writing inline code to a temp `.py`/`.js` file (via write_file) and running that, instead of `-c`/`-e` with complex nested quoting.
- Or: have the bash tool itself detect `-c`/`-e` flags on known REPL launchers with quote characters inside the payload and warn pre-flight.

PierrunoYT:
This looks addressed by #465, which fixes the underlying cmd.exe `/S /C` quote-mangling behavior for commands with embedded quotes (e.g. `python -c "..."`).

Vasanthdev2004:
[...]

PR #465, opened 2026-07-03
fix(tools): fix cmd.exe /S/C corrupting commands with embedded quotes

euxaristia:
## What

Any shell command whose own text contains embedded double quotes (`python -c "print(15 / 3)"`, `git commit -m "message"`, `node -e "..."`) was silently corrupted on Windows before reaching the target program. Reported live: a model ran `python -c "print(15 / 3)"` and got back `SyntaxError: unterminated string literal`, with the code Python actually received shown as literally `"print(15`, truncated right before the `/`.

## Root cause

`bash.go` passes the full command text as one argv element to `cmd.exe /d /s /c`. Go's `os/exec` escapes it correctly for a standard `CommandLineToArgvW`-style parser: wrapping the whole thing in quotes and backslash-escaping the embedded ones. But `cmd.exe`'s own `/C` remainder parsing does not follow those rules once `/S` is present. Per its documented behavior, it strips the first and last literal quote character in the whole remainder (not a matched pair) before handing the now-corrupted text along to be re-tokenized. The escaping was correct for one layer of parsing; the command line actually goes through two.

## Fix (updated after review)

The original version of this PR worked around the corruption by writing `commandText` to a temp `.cmd` file and pointing `/c` at the file's path. Review found that traded one bug for others: it silently reinterpreted the command as batch-file syntax instead of command-line syntax (`for %i in (...) do ...` needs `%%i` only inside an actual `.bat`/`.cmd` file, not typed at a prompt or passed via `/C`), `/s` was still corrupting the *script path itself* if it ever contained spaces, and the temp file leaked on sandbox planning errors.

Fixed at the actual source instead, with no temp file at all: `commandText` now reaches `cmd.exe` completely unescaped, exactly as if it had been typed at an interactive prompt.

- On the unwrapped path, `internal/tools/bash_proc_windows.go` sets `exec.Cmd.SysProcAttr.CmdLine` directly, bypassing Go's normal per-argument escaping (the thing that was wrapping `commandText` in an outer pair of quotes in the first place).
- On the sandboxed path, the command is serialized as CLI argv and reconstructed inside a separate `zero-windows-command-runner` process, so the same fix has to happen again where *that* process actually builds its `CreateProcessAsUser` command line (`internal/sandbox/windows_process_windows.go`'s `windowsCommandLine`): it now recognizes Zero's shell-invocation shape and leaves the final element (the command text) unescaped instead of running it through `syscall.EscapeArg` like every other argument.
- Both call sites for "what shape is Zero's shell invocation" (the argv-building side and the raw-command-line-building side) now go through two shared functions in `internal/sandbox/windows_runner.go` (`WindowsShellArgs`, `WindowsShellCommandLine`), instead of the `/d /c` literals living in three separate places that could drift out of sync.
- `/s` is dropped entirely, not just avoided on one path. It forced the more aggressive of cmd.exe's two documented stripping behaviors; without it, a command that legitimately starts with a quoted executable path (`"C:\Program Files\foo.exe" arg`) gets cmd.exe's own careful quote-preserving special case for that exact shape, instead of also being blindly stripped.

No temp file means nothing to leak on a sandbox planning error, and no batch-file reinterpretation of the command text, so this also resolves the batch-vs-command-line and cleanup-leak review findings as a consequence of removing the mechanism they were about, not as separate patches.

`buildBashCommand` is back to its original 3-value return (`*exec.Cmd`, `CommandPlan`, `error`); the cleanup-func threading through `bash.go` and `exec_command.go` from the temp-file version is gone.

## Testing

- `TestBashToolPreservesEmbeddedQuotesOnWindows` (kept from the original PR): pins the original corruption shape via a self-dispatch test helper.
- `TestBashToolRunsCommandLineForLoopSyntax` (new): runs `for %i in (1 2 3) do echo %i` (single percent) through the tool and asserts it expands correctly. This is the test that would have failed under the temp-`.cmd`-file version of this fix and passes now.
- `TestWindowsShellArgsHasNoSAndKeepsCommandTextWhole`, `TestWindowsShellCommandLineLeavesCommandTextUnescaped`, `TestWindowsShellCommandLineFromArgsRecognizes/RejectsOtherShapes` (new, `internal/sandbox`): unit-test the shared shape/escaping logic directly, cross-platform (no Windows-only build tag needed for these, since they're pure string logic).

All of the above run and pass on an actual Windows machine, not just cross-compiled. `go build ./...` and `go vet ./internal/tools/... ./internal/sandbox/...` clean on windows/linux/darwin cross-compiles. Full `internal/tools` and `internal/sandbox` suites green, excluding failures pre-existing and identical on unmodified `main` in my environment (unrelated path-assumption issues under a deeply nested working directory, not touched by this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows command handling so quoted arguments and loop syntax are preserved more reliably.
  * Made Windows command-line parsing more consistent across sandboxed and non-sandboxed execution paths.
  * Added stricter validation to reject invalid command arguments consistently.

* **Tests**
  * Added regression coverage for embedded quotes, `cmd.exe` loop syntax, and Windows shell command-line shape handling.

### DIFF
--- a/internal/sandbox/windows_process_windows.go
+++ b/internal/sandbox/windows_process_windows.go
@@ -87,11 +87,20 @@ func windowsCommandLine(args []string) (string, error) {
 	if len(args) == 0 {
 		return "", errorsNewWindowsProcess("missing command")
 	}
-	out := make([]string, 0, len(args))
 	for _, arg := range args {
 		if strings.ContainsRune(arg, 0) {
 			return "", errorsNewWindowsProcess("command argument contains NUL")
 		}
+	}
+	// Zero's cmd.exe shell invocation needs its final element (the shell
+	// command text) reaching cmd.exe raw, not escaped like a normal argv
+	// element: see WindowsShellCommandLine for why. Every other command this
+	// runner might launch is escaped normally below.
+	if commandLine, ok := windowsShellCommandLineFromArgs(args); ok {
+		return commandLine, nil
+	}
+	out := make([]string, 0, len(args))
+	for _, arg := range args {
 		out = append(out, syscall.EscapeArg(arg))
 	}
 	return strings.Join(out, " "), nil
--- a/internal/sandbox/windows_runner.go
+++ b/internal/sandbox/windows_runner.go
@@ -105,6 +105,56 @@ const (
 	WindowsSandboxLevelDisabled   WindowsSandboxLevel = "disabled"
 )
 
+// WindowsShellArgs returns cmd.exe's argv (after the executable name) for
+// running commandText as a shell command on Windows: no wrapping quotes and
+// no /S, so commandText reaches cmd.exe's own /C remainder parsing exactly as
+// written. See WindowsShellCommandLine for why that matters. Used both as
+// CommandSpec.Args for the plain (unwrapped) path and, via
+// windowsShellCommandLineFromArgs, to recognize this same shape once it has
+// round-tripped through the sandboxed runner's CLI argv.
+func WindowsShellArgs(commandText string) []string {
+	return []string{"/d", "/c", commandText}
+}
+
+// WindowsShellCommandLine builds the raw command line a Windows CreateProcess
+// call should receive to run commandText as a shell command: "cmd.exe /d /c "
+// followed by commandText completely unescaped.
+//
+// cmd.exe's own /C remainder parsing is not CommandLineToArgvW-compatible: it
+// applies its own quote-stripping heuristic (documented in `cmd /?`) whenever
+// the remainder starts with a literal quote, and that heuristic does not
+// understand backslash-escaped inner quotes the way a normal argv consumer
+// would. A command like `python -c "print(15 / 3)"` passed through the
+// standard exec.Cmd Args mechanism gets wrapped in an outer pair of quotes
+// (because it contains spaces) with its own quotes escaped as \" — exactly
+// the shape that heuristic corrupts, stripping the outer quotes but leaving
+// the literal backslashes behind. Passing commandText through raw, exactly as
+// the model wrote it, means cmd.exe parses it the same way it would if typed
+// directly at an interactive prompt: a leading quoted executable path like
+// `"C:\Program Files\foo.exe" arg` is preserved correctly by cmd.exe's own
+// quote-preserving special case, and a command with no leading quote at all,
+// like the python example above, is never touched by the stripping heuristic
+// in the first place.
+func WindowsShellCommandLine(commandText string) string {
+	return "cmd.exe /d /c " + commandText
+}
+
+// windowsShellCommandLineFromArgs reports whether args is exactly
+// ["cmd.exe"]+WindowsShellArgs(text) for some text and, if so, returns
+// WindowsShellCommandLine's result for that text. The sandboxed runner path
+// only has the deserialized argv (WindowsSandboxCommandConfig.Command) to
+// work from, not the original commandText, so it recovers the raw command
+// line by recognizing the shape instead.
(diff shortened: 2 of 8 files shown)


PierrunoYT:
This addresses #415 (`python -c` with nested quotes getting mangled by cmd.exe on Windows) — same root cause: cmd.exe's `/C` remainder parsing stripping the first/last quote rather than treating the whole argument as pre-escaped for one layer of parsing.

jatmn (changes requested):

## Findings

- [P1] Preserve `cmd.exe` command-line syntax instead of switching to batch-file syntax
  `internal/tools/bash.go:296`
  This changes every Windows `bash`/`exec_command` invocation from running the user command as the `/c` command string to writing it into a `.cmd` file. Those are not equivalent parsers: commands that are valid at a `cmd.exe` prompt, such as `for %i in (*.go) do echo %i`, must be written with doubled percent variables (`%%i`) inside a batch file. Zero's prompt/tool guidance tells the model to use Windows `cmd.exe` syntax, and the old path accepted that command-line form; with this patch, the same command text is parsed as a batch file and breaks. Please keep the quote fix from changing the command language, or translate/escape command-line-only constructs before writing the script.

- [P1] Complete CodeRabbit's request to remove `/s` from the temp-script invocation
  `internal/tools/bash.go:305`
  The Windows path still returns `cmd.exe /d /s /c <scriptPath>`. This fixes embedded quotes in the command text only until the temporary `.cmd` path itself contains spaces, because `cmd.exe /s /c` strips the first and last quote from the `/c` remainder before re-tokenizing it. `os.CreateTemp("", ...)` writes under the user's temp directory, so a Windows profile/temp path with spaces can make every `bash`/`exec_command` run fail before the generated script is reached. Please complete CodeRabbit's request by dropping `/s` for this script-file path while keeping the existing `/d /c` behavior.

- [P2] Complete CodeRabbit's request to clean up the script on sandbox planning errors
  `internal/tools/bash.go:248`
  `windowsShellCommandSpec` creates the temporary `.cmd` file before `engine.CommandContext` is called, but if sandbox command planning fails, `buildBashCommand` returns the error together with the cleanup function and both callers return before invoking it. That leaves a temp script behind for every preparation failure on Windows. Please complete CodeRabbit's request by running the script cleanup before returning the `engine.CommandContext` error, and return a no-op cleanup on that path so callers cannot double-delete it.

euxaristia:
Redesigned this after your review, rather than patching the temp-file version further, since the first finding challenged the approach itself.

The temp-.cmd-file mechanism is gone entirely. commandText now reaches cmd.exe unescaped via SysProcAttr.CmdLine on the unwrapped path, and via the same treatment inside the sandboxed runner's own process creation (internal/sandbox/windows_process_windows.go) on the wrapped path, exactly as if typed at an interactive prompt. Details on all three findings:

1. Command-line vs batch-file syntax: no longer relevant, there's no script file. Added TestBashToolRunsCommandLineForLoopSyntax, which runs a single-percent for loop (for %i in (1 2 3) do echo %i) through the tool. This is exactly the construct that broke under the temp-file version and now passes.
2. /s corrupting a spacey script path: also no longer relevant (no script path), and /s is dropped entirely rather than just avoided on one path, per your and CodeRabbit's ask.
3. Cleanup leak on sandbox planning errors: no temp file means nothing to leak. buildBashCommand is back to its original 3-value return.

Full explanation and the new tests are in the updated PR description. Ran the whole suite on an actual Windows machine, not just a cross-compile, including the new tests.

jatmn (approved):
@Vasanthdev2004

kevincodex1 (approved):
LGTM

Issue #467, opened 2026-07-04
Windows: bash tool with piped POSIX commands (e.g. `| cat`) causes "Not enough memory resources are available to process this command"


euxaristia:
## Summary

On Windows, the agent frequently emits shell commands using POSIX-style pipes to utilities that don't exist in `cmd.exe` (or cause parse failures):

```sh
gh pr view 465 --json reviews,comments -q '.reviews | length' 2>&1 | cat
gh api repos/.../pulls/465/reviews 2>&1 | cat
go test -run 'TestFoo|Bar' ...
```

These fail with the unhelpful error:

> Not enough memory resources are available to process this command.
> exit 1

### Root cause

- The model defaults to Unix idioms (`| cat` / `| head` to control output or bypass pagers, single quotes around args containing `|`).
- `cmd.exe` has no `cat`/`head`. Pipes to missing commands + unquoted metachars in the remainder passed to `cmd /d /c` lead to command-line parse errors that surface as this memory warning.
- Pre-flight detection only caught bare `ls` and `cd /...` paths. Piped utilities ran straight through.

This was reported via multiple TUI screenshots during PR review work (inspecting large review threads, etc.).

### Fix on this branch

- Extended `detectShellCommandIssue` (with regex + string fallbacks) to **flag piped POSIX utilities** (`head|tail|grep|cat|wc|...`) *before* execution on Windows.
- Significantly improved guidance in the `bash`/`exec_command` tool descriptions and system `<environment>` block:
  - On Windows, wrap arguments containing `|` (or `& > < ^`) in **double quotes**: `gh --jq ".foo | bar"` (single quotes do **not** protect metachars in cmd.exe).
  - Prefer `gh --jq` / `-q` for server-side filtering, PowerShell `Select-Object -First N` for limiting output, or Zero's native tools (`grep`, `read_file`, `glob`, etc.).
- Better error suggestions when bad commands slip through.
- Complementary quoting fixes (`/d /c` raw command line, no temp script, no `/s`) so properly-quoted commands reach cmd.exe intact.

This prevents the bad patterns from ever being executed and teaches the model correct Windows shell usage going forward.

## Changes

**`internal/tools/shell_runtime.go`**
- Added `windowsPosixUtilityPattern` (and kept/expanded coverage) to catch piped utilities early.
- Updated `detectShellCommandIssue` and suggestions.
- Enhanced `shellGuidanceForGOOS` with explicit Windows quoting advice for metachars in args.

[...]

PR #468, opened 2026-07-04
fix(tools): Windows cmd.exe quoting guidance and clipboard escaping fix

### LINKED ISSUE #469: bash tool: unbounded stdout/stderr buffering can OOM the agent
## Problem

The `bash` tool (and to a lesser extent `exec_command`) allocates full command output into `bytes.Buffer` with no size limit:

```go
// internal/tools/bash.go:127
var stdout bytes.Buffer
var stderr bytes.Buffer
command.Stdout = &stdout
command.Stderr = &stderr
...
output := formatBashOutput(stdout.String(), ...)
```

Other tools (`grep`, `read_file`, `web_fetch`) use `applyOutputBudget` + `io.LimitReader`. Bash does not.

## Impact

- Easy OOM or unresponsiveness on large output (cat huge file, gh pr with thousands of comments, find /, verbose tests, etc.).
- Directly caused the "Not enough memory resources..." errors reported during PR review work.
- No backpressure; the process can be killed before the command even finishes.

## Suggested fix

- Add a reasonable byte limit (e.g. 10-50 MiB) + `io.LimitReader` for stdout/stderr in bash.go (and review exec_command).
- Reuse or extend `applyOutputBudget` / `max_output_tokens` logic.
- Truncate early and surface a clear message + suggestion to narrow the command.

### LINKED ISSUE #470: specialist: no hard maximum depth limit on Task / nested specialists
## Problem

Specialist / Task tool nesting has no hard upper bound.

```go
// internal/specialist/exec.go
if options.CurrentDepth < 0 { ... }
// then always:
args = append(..., "--depth", strconv.Itoa(input.CurrentDepth+1), ...)
```

Only a negative check. Depth is just metadata passed down to child `zero` processes.

## Impact

- A specialist (or the model) can recursively spawn more specialists indefinitely.
- Leads to process explosion, high CPU/memory, stack overflow, or runaway costs.
- No protection against accidental or adversarial deep recursion.

## Suggested fix

- Add a hard `MaxDepth` (e.g. 5-8) in `TaskRunOptions` / `Executor.Run` / swarm launcher.
- Return clear error when exceeded.
- Document and expose via CLI flag if needed.
- Consider a global per-run limit in the main agent loop.

### LINKED ISSUE #471: acp: silent errors in session persistence can lose history
## Problem

```go
// internal/acp/agent.go:391
if a.deps.Store != nil {
}
```

## Impact

## Suggested fix

- Log the error at minimum.

### LINKED ISSUE #472: clipboard: risky string concatenation into PowerShell -Command
## Problem

String concatenation into a PowerShell `-Command` script:

```go
// internal/imageinput/clipboard.go:79
script := `... $img.Save('` + tmpPath + `', ...)`
cmd := exec.Command("powershell", "-NoProfile", "-Command", script)
```

Similar risky patterns exist in other clipboard/image paths.

## Impact

- If `tmpPath` ever contains single quotes, backticks, or PowerShell metachars, it can break the script or (in theory) lead to unintended execution.
- Inconsistent with the careful argument-passing used elsewhere (e.g. Linux clipboard code explicitly avoids this).

## Suggested fix

- Use discrete arguments or proper escaping / here-strings.
- Or write the script to a temp .ps1 file and invoke it safely.
- Add a comment or helper for "safe PowerShell command construction".

euxaristia:
Closes #472

Windows shell quoting guidance and a clipboard escaping fix.

## Changes

- Rewrote the Windows shell/quoting guidance (bash tool description and system prompt) to explain double-quote escaping for metacharacters in cmd.exe (single quotes do not protect them).
- Fixed the clipboard PowerShell `-Command` script (`clipboard.go`) to escape embedded single quotes in the temp file path instead of concatenating it unescaped. Closes #472.
- CI: relaxed the deadline slack in `TestCollectRespectsDeadlineUnderContinuousOutput` for Windows CI jitter.

## Split from this PR

Per reviewer request (Vasanthdev2004), the bash output OOM cap, specialist depth cap, and persistence error logging that were originally bundled here have been split into their own PRs, since they are independent fixes:

- #492: bash stdout/stderr OOM cap (closes #469)
- #491: specialist nesting depth cap (closes #470, merged)
- #493: session persistence error logging (closes #471)

This PR no longer touches `internal/tools/bash.go`'s output buffering, `internal/specialist/exec.go`'s depth check, or `internal/acp/agent.go`'s persistence logging.

## Dropped the shell_runtime.go POSIX/MSYS detection rework (per Vasanth's review)

This PR previously also flagged piped POSIX utilities (`| cat`, `| head`, etc.) and reworked `internal/tools/shell_runtime.go`'s detection to handle quoted arguments and cmd.exe operator boundaries, closing #467.

That work has been dropped from this PR entirely. #476 (already approved) rebuilt the same detection surface independently and more thoroughly: it removes `windowsPosixUtilityPattern`/`windowsLSCommandPattern` altogether in favor of segment/word-anchored detection (`windowsCommandSegments` + `firstCommandWord` + a single `windowsMsysProneNames` set), and additionally handles MSYS binary-path detection, the `require_escalated` escalation-bypass bug, and cmd.exe redirection boundaries that this PR's regex-plus-stripping approach didn't cover. The two are incompatible rewrites of the same code, not complementary, and once #476 merges this PR would no longer even compile against `main` (it still referenced `windowsLSCommandPattern`, which #476 deletes).

`shell_runtime.go` and the corresponding tests in `bash_tool_test.go` are now reverted to match `main` exactly. #467 is no longer closed by this PR; #476 should close it instead. `internal/sandbox/safe_command.go` is unaffected by this change (the #414 pager-suggestion overlap noted below was already resolved in an earlier commit on this branch).

Everything else on this PR (the quoting guidance, the clipboard fix, the CI jitter tolerance, and the earlier resolution of the #414 pager-suggestion duplication) is unchanged and still applies.

See individual issues for details and full analysis.


euxaristia:
Addressed the remaining CodeRabbit review comments from #468:

- Added \cat\' to the POSIX utility message in shell_runtime.go
- Fixed persistTurn to use stderr instead of stdout for warnings in acp/agent.go
- Escaped tmpPath in clipboard.go PowerShell script
- Implemented proper cappedWriter for bash output to prevent OOM during command execution (not just after), with truncation notice in format functions.

Addresses unaddressed items from CodeRabbit reviews.

Vasanthdev2004 (changes requested):
## Request changes — real, valuable work, but a Windows regression + grab-bag scope

Worth merging: **yes, after a split and one revert** — this is not slop. The core (Windows shell-quoting guidance, the POSIX `cat` flag, interactive-command suggestions) is a correct *additive* refinement on top of merged #412 — I confirmed #412 is an ancestor of your base, so you build on it rather than re-doing it — and the OOM `cappedWriter`, the specialist depth-cap (`--depth` propagation), and the stderr logging fix are each individually sound. Thanks @euxaristia.

Two things block it:

**🔴 Hard blocker — the clipboard change breaks image paste on default Windows.** `clipboard.go` switches PowerShell from `-Command` to `-File` with no `-ExecutionPolicy Bypass`. `-Command` is *not* gated by script-file execution policy; `-File` *is*, and the Windows client default is `Restricted` — so on an unconfigured machine, image paste silently stops working. Two more regressions ride in the same switch: the BOM-less UTF-8 `.ps1` is decoded as ANSI by Windows PowerShell 5.1, so a non-ASCII `%TEMP%` path (accented username) mis-resolves; and writing-then-executing a temp script adds a TOCTOU window the inline form doesn't have. **Fix:** revert to `-Command`, keeping only the single-quote escaping — that already fixes the bug you were targeting, with none of the downside.

**🟡 Scope — split the grab-bag.** This bundles shell guidance + OOM cap + specialist depth + persistTurn logging + clipboard under a "Windows shell guidance" title. Each is fine on its own, but they need separate issue-linked PRs for reviewability and to avoid re-churning files #412 just touched (CodeRabbit's out-of-scope warning flagged the same).

One rebase note: once #476 lands (it expands the same `windowsPosixUtilityPattern` and already adds `cat`), your `cat` addition becomes redundant — drop it, keeping the quoting-guidance rewrite + interactive suggestions as the unique value. Credit where due: the additive shell slice is a legitimate improvement; just unbundle it and revert the clipboard call.

euxaristia:
Pushed 2d0b14c reverting the clipboard call to powershell -Command, keeping only the single-quote escaping, per the execution policy, encoding, and TOCTOU concerns. I will follow up separately on splitting the remaining changes into scoped PRs and on dropping the cat pattern once #476 lands.

jatmn (changes requested):

## Findings

- [P2] Keep `| cat` covered before closing #467
  `internal/tools/shell_runtime.go:28`
  The PR description says this closes #467, but the current Windows preflight regex still does not include `cat`. That means the exact issue example `gh ... 2>&1 | cat` is not blocked before execution, while only cases that also contain another listed utility such as `| head` are caught. Please either keep `cat` in this PR's detector/tests, or stop this PR from closing #467 until the `cat` coverage has landed elsewhere and the issue-closing wording is updated.

- [P2] Do not suggest another follow-forever command for `tail -f`
  `internal/sandbox/safe_command.go:641`
  The Windows replacement for a blocked `tail -f` is `Get-Content -Path file -Tail N -Wait`, but `-Wait` follows the file indefinitely too. The detector is supposed to steer the agent away from commands that hang a non-interactive run, so this can send it straight into the same class of problem with a Windows-native command. Please make the primary suggestion bounded, such as `Get-Content -Tail N` without `-Wait`, and tighten the test so it rejects follow/`-Wait` suggestions.

- [P2] Coordinate or drop the duplicate #413 fix from this PR
  `internal/sandbox/safe_command.go:618`
  The Windows interactive-suggestion slice in this PR overlaps the older open PR #414, which already targets #413 and changes the same `safe_command.go`, `safe_command_test.go`, and `bash_tool_test.go` paths. #414 has already been through review for this exact behavior, including the Windows sandbox test fallout, while this PR neither links #413 nor explains why it should replace that earlier PR. Please either coordinate this slice with #414 and make this PR the chosen replacement, or drop the interactive-suggestion changes here so the review queue does not carry two competing implementations for the same issue.

- [P3] Do not auto-close #473 from this PR
  `PR description`
  The body says `Closes #473`, but the current patch does not replace, augment, simplify, or materially test the hand-written parser risk described in that issue; it only adds Windows-specific suggestion text on top of the existing parser. Merging as-is would close the parser-risk issue even though the risk remains intentionally deferred. Please remove #473 from the closing keywords or include the parser-risk fix here.

jatmn (approved):

euxaristia:
Status update on the open findings:

Already fixed before this push:
- Vasanthdev2004's hard blocker (clipboard -File Windows regression): reverted to -Command with single-quote escaping kept (2d0b14c).
- Vasanthdev2004's grab-bag scope concern: split into #491 (specialist depth), #492 (bash OOM cap), #493 (persistence logging) (940ee4b).
- jatmn's cat/#467 coverage: added back to windowsPosixUtilityPattern, including the bare `| cat` case (49d0ebf).
- jatmn's duplicate-with-#414 concern: the interactive-suggestion slice was dropped from this PR entirely (49d0ebf).
- jatmn's #473 auto-close concern: the PR body already states the parser risk is not closed by this PR.

Fixed in 6e62c9f (this push): CodeRabbit's remaining finding, that the pager guard's suggestion for a blocked less/more/most ("use cat, head, or tail -n N") names commands that don't exist on cmd.exe. Added a windowsSuggestion override so Windows gets "use type, or read_file with offset/limit" instead. Fixed the existing test that hardcoded the cat suggestion regardless of OS, and added dedicated coverage that the two platforms' suggestions differ.

go build/vet clean on linux and windows targets, go test ./internal/sandbox/... ./internal/tools/... ./internal/agent/... green.

Ready for re-review.

jatmn (changes requested):

## Findings

- [P2] Do not flag quoted `| cat` text as an executable pipeline
  `internal/tools/shell_runtime.go:29`
  The new Windows POSIX-utility guard runs this regex over the raw command string before `bash`/`exec_command` executes, so it also matches utility names that appear inside double-quoted argument values. For example, a valid Windows command such as `gh pr comment --body "please do not use | cat here"` is blocked because the quoted body contains `| cat `, even though `cmd.exe` would treat that pipe as literal text. This regresses the same review/comment workflows the PR is trying to improve whenever the comment body mentions the forbidden pipeline. Please make this detection segment/quote-aware, or otherwise only match actual command positions, and add a regression test for a quoted argument containing `| cat`/`| grep`.

- [P2] Do not suggest `more` when the next guard blocks it
  `internal/tools/shell_runtime.go:71`
  The Windows POSIX-utility block tells the model to use cmd.exe equivalents including `more` to page output, but both shell tools immediately run `DetectInteractiveCommand` after this check, and `safe_command.go` classifies `more` as an interactive pager on Windows. A retry such as `some-cmd | more` therefore gets blocked by the interactive-command guard with a different suggestion (`type` or `read_file`), so the remediation can lead directly into another preflight failure. Please remove `more` from this suggestion or make the two guards agree on which Windows paging alternatives are actually allowed.

euxaristia:
Addressed both new findings from the latest review:

- Quoted `| cat`/`| grep` text falsely flagged: detectShellCommandIssue now runs its cd/ls/POSIX-utility regexes against the command with double-quoted spans blanked out first (stripDoubleQuotedSpans), so text that only appears inside a quoted argument value (e.g. `gh pr comment --body "please do not use | cat here"`) can no longer trigger a match. Added regression tests for quoted `| cat` and `| grep`, and confirmed the unquoted/real pipeline cases still get flagged.
- `more` recommended as a paging suggestion while also being blocked as an interactive pager: dropped `more` from the POSIX-utility suggestion text entirely, so it no longer points the model at a command the other guard immediately blocks.

go build/vet clean on linux and windows targets, go test ./internal/tools/... ./internal/agent/... green.

euxaristia:
Addressed: cmd.exe caret escapes (^&, ^|, ^;, ^^, ^<, ^>) are now neutralized in the same pre-pass as the double-quote stripping, so foo^|cat reads as the literal token it is in cmd.exe instead of a pipe into cat. Added regression coverage, including confirming an actual unescaped pipe is still flagged. go build/vet/tests green on linux and windows targets.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found issues that need to be addressed before this is ready.

## Findings

  `GitHub mergeability`
  GitHub currently reports `mergeable=CONFLICTING` with `mergeStateStatus=DIRTY`, so this branch cannot be merged as-is. Please rebase on the latest `main` branch and resolve the conflicts before addressing any remaining review items, so the final fixes apply to the code that will actually be merged.

- [P2] Keep flagging POSIX utilities when cmd redirection/operators follow the name
  `internal/tools/shell_runtime.go:29`
  The new Windows POSIX-utility regex only matches `cat`/`head`/`grep`/etc. when the utility name is followed by whitespace or the end of the command. `cmd.exe` also accepts operators and redirections immediately after a command name, so an actual pipeline such as `some-cmd|cat>out.txt` or `some-cmd|head&&dir` does not match this preflight even though it still invokes the POSIX utility that this PR is trying to block. Please treat cmd metacharacters such as `|`, `&`, `<`, and `>` as valid command-name boundaries here, and add regression coverage for a piped utility followed immediately by redirection or another operator.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found issues that need to be addressed before this is ready.

## Findings

  `GitHub mergeability`
  GitHub currently reports `mergeable=CONFLICTING` with `mergeStateStatus=DIRTY`, so this branch cannot be merged as-is. Please rebase on the latest `main` branch and resolve the conflicts before addressing any remaining review items, so the final fixes apply to the code that will actually be merged.

- [P2] Resolve the remaining #414/#413 overlap or update the PR body
  `internal/sandbox/safe_command.go:46`
  The PR body says the Windows-specific interactive-command suggestions for `less`/`more`/`top`/`tail -f` were dropped because they duplicated the already-open #414, but the current patch still adds `windowsSuggestion` behavior for the pager entries (`less`, `more`, and `most`) and adds `TestDetectInteractiveCommandPagerSuggestionIsPlatformSpecific` for that same interactive-command path. That leaves the earlier duplicate-scope request only partially resolved and makes the description misleading for reviewers trying to decide whether #414 or this PR owns #413's pager fix. Please either drop this remaining `safe_command.go` slice from this PR, or explicitly coordinate/describe that this PR is taking the pager subset from #414 so the overlapping PRs and issue-closing path are unambiguous.

euxaristia:
@Vasanthdev2004 Same as on #475, I can't re-request your review directly. Your findings here (the `require_escalated` escape hatch and the `.exe` regex over-match) were addressed, and jatmn's latest review on the current head is an approval. Would appreciate a re-look when convenient.

Vasanthdev2004 (changes requested):
The rebase and the #414 pager slice are both sorted now — thanks. But holding here, because this collides with #476, which is already approved, and I don't think both can land as-is.

Both PRs rework the same Windows detection in `shell_runtime.go`, in incompatible ways:

- **#476** removes `windowsPosixUtilityPattern` and `windowsLSCommandPattern` entirely and replaces them with segment/word-anchored detection (`windowsCommandSegments` + `firstCommandWord` + a single `windowsMsysProneNames` set) — that's what avoids the quoted-argument false positives.
- **This PR** keeps and extends `windowsPosixUtilityPattern` (adds `cat` + operator/redirection boundaries) and adds `stripDoubleQuotedSpans` + `stripCmdCaretEscapes` to solve that same quoted-arg problem a different way.

They can't coexist. Once #476 merges, this branch still references `windowsLSCommandPattern`, which #476 deletes, so it won't compile until it's reworked — and even after a rebase the two are duplicate, competing solutions to the same detection. It's also why CI is green here but would break against #476.

The parts of this PR that *don't* overlap #476 are worth keeping: the Windows quoting guidance in `system_prompt.go` (wrap metachar args in double quotes, single quotes don't protect in cmd.exe) and the clipboard escaping fix are both genuinely useful and unique to this branch.

Suggestion: let #476 own the `shell_runtime.go` POSIX/MSYS detection, and narrow this PR down to the quoting guidance + clipboard fix, rebased after #476 lands. If you intended these two to be sequential and I'm missing the ordering, tell me and I'll re-look — but as the branches stand they're on a collision course nothing's flagged yet since neither is merged.

(Issue side is fine — #467/#472 are real; I'll mark them `issue-approved` once the scope here settles so the gate isn't the thing holding it.)

jatmn (changes requested):
Thanks for narrowing this down. I found a couple of issues that still need to be addressed before this is ready.

## Findings

- [P2] Do not show single-quoted cmd.exe examples in the quoting guidance
  `internal/agent/system_prompt.go:262`
  The new Windows guidance correctly says that `cmd.exe` requires double quotes for arguments containing metacharacters, but the examples immediately before that still show the failing forms: `gh --jq '.a | b'` and `go test -run 'A|B'`. On Windows those single quotes do not protect the pipe, so this prompt can still teach the model to copy the exact syntax that caused the original shell failures. Please make the examples themselves use the safe form, for example `gh --jq ".a | b"` and `go test -run "A|B"`, and keep the regression test checking for representative double-quoted examples rather than only the generic phrase.

- [P2] Put the quoting rule in the shell tool guidance, not only the system prompt
  `internal/tools/shell_runtime.go:40`
  The PR description says the Windows shell/quoting guidance was rewritten for the bash tool description and system prompt, but the shared tool guidance used by `bash` and `exec_command` still only says `Uses Windows cmd.exe syntax on Windows; prefer cwd over cd when changing directories.` The current tests also only require `cmd.exe`/`cwd`, so the tool schema can still omit the new double-quote rule even though these are the exact tools where the model chooses shell syntax. Please update `shellGuidanceForGOOS("windows")` (and the bash/exec-command description tests) to include the same double-quote/single-quote warning, so the model sees the rule at the shell tool call site as well as in the broader environment block.

euxaristia:
Update: both findings from the last review are fixed.

- The Windows shell guidance now uses double-quoted examples (`gh --jq ".a | b"`, `go test -run "A|B"`) instead of the single-quoted forms that don't actually protect metacharacters in cmd.exe.
- `shellGuidanceForGOOS("windows")` (the bash/exec_command tool description, not just the system prompt) now carries the same double-quote rule.

Separately, #476 merged into main while this was in flight and rewrote the same guidance string (adding the MSYS coreutils warning). Rebased onto main and combined both pieces into one coherent string in each spot, rather than one clobbering the other. Build, vet, and the affected tests all pass. Mergeable again as of the latest commit.

Vasanthdev2004 (approved):
Thanks for splitting the unrelated scope out earlier — this is tight now and easy to see. The clipboard fix does the PowerShell single-quote doubling correctly, and the cmd.exe double-quote guidance matches how cmd actually parses. It applies cleanly on top of the MSYS quoting work that already merged, so no overlap there.

I've marked the backing issue (#472) approved so this can go. Approving.

Vasanthdev2004 (changes requested):
Took another look alongside jatmn's review — the two points there are fair and worth doing before this lands. The Windows quoting examples in system_prompt.go still show the single-quoted forms (`gh --jq '.a | b'`, `go test -run 'A|B'`) right before the rule saying cmd.exe needs double quotes, so the prompt can still teach the exact syntax that breaks — switch the examples themselves to the double-quoted form. And the rule needs to live where the model actually picks shell syntax: `shellGuidanceForGOOS("windows")` in shell_runtime.go only mentions cmd.exe/cwd, so the bash/exec_command tool descriptions don't carry the quoting rule — add it there too (and to those tests).

The gate side is clear now — #472 is approved. Fold those two in and I'll re-approve.

euxaristia:
I have addressed the requested changes by ensuring the single-quoted examples are replaced with double-quoted ones, adding the quoting rules to shellGuidanceForGOOS in shell_runtime.go, and adding the corresponding description verification tests for both bash and exec_command tools. Pushed the updates to the branch.

Vasanthdev2004 (approved):
I read the Windows diff for issue #472. The cmd quoting now wraps args so embedded spaces and special characters survive the trip through cmd.exe, and the clipboard path stops mangling the copied text. Both are real Windows-only papercuts that the posix tests wouldn't have caught. Approving.

euxaristia:
@jatmn @Vasanthdev2004 both findings were addressed in the July 6 push (23c74a5). The system prompt examples now use the double-quoted forms, and the loop_test regression test asserts them and rejects the single-quoted forms. The double-quote rule was also added to shellGuidanceForGOOS("windows"), so the bash and exec_command tool descriptions carry it, with assertions in both description tests. The red ubuntu smoke run was an infra failure while restoring the Go module cache, not a test failure. I do not have rights to re-run it. Please take another look when you get a chance.

Vasanthdev2004:
fix smoke test @euxaristia

euxaristia:
@Vasanthdev2004 fixed the real part of the smoke failure in 69a83f8: exec_command_test.go was missing a gofmt pass, which the ubuntu formatting gate caught. The remaining mac and windows failures are the picker_test build break on main that your #589 fixes, so this PR should go fully green once that lands and CI reruns.

Vasanthdev2004:
@euxaristia wait i will fix it

euxaristia:
Ready to merge?

Vasanthdev2004:
for me yea @kevincodex1

jatmn (approved):
LGTM

kevincodex1 (commented):
looks good!

PR #481, opened 2026-07-04
fix: address bugs found in a multi-agent codebase audit

### LINKED ISSUE #480: Nine bugs found in a multi-agent codebase audit (sandbox, oauth, TUI, update, sessions)
## Summary

I ran five parallel review passes over different areas of the codebase (self-update, TUI keybindings, agent/session handling, sandbox/shell safety, and auth/credential storage) and found nine concrete defects, ranked by severity below. Fixes with regression tests are up in a PR that will link to this issue.

## High severity (fail-open / security)

1. **Sandbox deny-grants are case-sensitive on Windows** — `internal/sandbox/grant_scope.go`. Persistent per-path grants (`ScopeFile`/`ScopeDir`) compare paths with plain `==`/`HasPrefix`, unlike the workspace-boundary check which uses case-insensitive `filepath.Rel`. A user-authored `deny --path C:\Users\me\project\Secrets` silently fails to match a request spelled `c:\users\me\project\secrets` — same file on a case-insensitive filesystem, but the deny doesn't apply and the request falls through to normal allow logic.

2. **Windows lock-contention fix applied inconsistently** — `internal/securefile/securefile.go` still had the pre-#445 bug that `internal/oauth/encrypt.go` fixed: it only treated `os.ErrExist` as lock contention, not `ERROR_ACCESS_DENIED`. Since `securefile.go` backs `credstore`'s encrypted API-key storage, concurrent `zero` processes on Windows could still spuriously hard-fail with "Access is denied".

3. **Provider fallback ignores OAuth logins** — `internal/cli/setup.go` (`firstUsableProvider`). Only checked inline API keys/env vars, never OAuth logins, contradicting `setupRequired`'s own logic. If `activeProvider` goes stale/blank, a fully-authenticated OAuth-only provider (e.g. `zero auth login xai`) was skipped and the user got forced back into onboarding despite having working credentials.

## Medium severity (correctness/data corruption)

4. **UTF-8 truncation can corrupt resumed session context** — `internal/sessions/exec_session.go`, `summarizePayload` truncated with a byte-index (`text[:500]`), which can split a multi-byte UTF-8 rune. Feeds directly into exec prompts for `--resume`/`--fork`.

5. **Keybinding collisions silently shadow hardcoded shortcuts** — `internal/tui/model.go`. Dispatch is a priority-ordered switch with configurable bindings interleaved with hardcoded ones and zero collision detection. E.g. configuring `toggleDetailed` to `ctrl+f` permanently killed the "favorite model" shortcut.

6. **Windows lock-error diagnostics masked** — `internal/oauth/encrypt.go`. The real lock-creation error was discarded in favor of a subsequent unrelated `ErrNotExist`, so genuine ACL failures surfaced as a misleading "timed out: file does not exist" instead of an immediate, accurate error.

## Low severity / edge cases

7. Windows shell pre-flight regex wasn't quote-aware (`internal/tools/shell_runtime.go`) — could false-positive block harmless commands like `echo "foo; ls -la"`.
8. `internal/update/apply.go` checksum verification didn't cross-check the archive filename against what's actually extracted.
9. `internal/update/extract.go`'s `extractZip` didn't reject symlink-mode zip entries the way `extractTarGz` rejects non-regular tar entries; and `replace_windows.go`'s restore-on-failure rename had no retry, risking a permanently missing binary on a transient Windows file lock.

## Scope note

Nothing was found in: archive path-traversal/zip-slip, sandbox command-injection/escaping, `additional_permissions` validation ordering, OAuth crypto (AEAD, random nonces), TLS/private-network classification, ChatGPT auth routing, and most of the provider-picker/model-picker filter logic — those areas checked out clean.

PierrunoYT:
## Summary

Fixes #480. Five parallel review passes swept different areas of the codebase (self-update, TUI keybindings, agent/session handling, sandbox/shell safety, and auth/credential storage) and surfaced nine concrete defects, all fixed here with regression tests.

- **sandbox**: `grant_scope` path comparisons were case-sensitive, letting a persisted deny grant be bypassed by spelling a Windows path with different case
- **securefile/oauth**: `createSecretFile`'s Windows `ERROR_ACCESS_DENIED` handling was only applied to the oauth token store, not the shared credstore backend; both also masked the real lock-contention error behind a misleading "file does not exist" timeout
- **cli**: `firstUsableProvider` ignored OAuth logins when picking a fallback provider, forcing unnecessary re-onboarding for an already-authenticated user
- **sessions**: exec prompt summarization truncated on a raw byte offset, which could split a multi-byte UTF-8 rune and embed invalid UTF-8 into a resumed session's prompt
- **tui**: configurable keybindings had no collision detection, so a remapped chord could silently make a hardcoded shortcut (or another configured binding) permanently unreachable
- **tools**: the Windows shell-syntax pre-flight check wasn't quote-aware and could block harmless commands whose quoted arguments merely contained operator-like text
- **update**: checksum verification didn't cross-check the archive filename it verified against the one being extracted; `extractZip` didn't reject symlink-mode entries like `extractTarGz` already did; `replaceBinary`'s restore-on-failure rename had no retry, risking a permanently missing binary on a transient Windows file lock

## Test plan

- [x] `go build ./...` and `go vet ./...` clean across the repo
- [x] New regression test added for each of the 9 fixes
- [x] Full test suite passes for every touched package (`sandbox`, `oauth`, `securefile`, `credstore`, `cli`, `sessions`, `tui`, `tools`, `update`)
- [x] One pre-existing, unrelated `tui` test failure (`TestProviderWizardAdvancesProviderAPIKeyAndModelSteps`) confirmed to fail identically on unmodified `main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding fallback so already signed-in providers can be used without repeating setup.
  * Fixed key binding conflicts by keeping valid shortcuts, reverting collisions to safe defaults, and showing warnings.
  * Made update downloads safer by rejecting mismatched checksums and unsupported archive entries.
  * Improved reliability on Windows and under file-lock contention when saving or restoring files.
  * Preserved valid UTF-8 when trimming long session text and reduced false shell-command warnings inside quotes.

### DIFF
--- a/internal/cli/setup.go
+++ b/internal/cli/setup.go
@@ -351,11 +351,12 @@ func oauthLoggedInProviders() map[string]bool {
 }
 
 // firstUsableProvider returns the saved provider best suited to run without
-// onboarding: the first usable (inline credential present, or no-auth/local)
-// non-local provider, else the first usable local one. It lets the CLI fall back
-// to an already-configured login when the active provider happens to lack a
-// credential, instead of re-running onboarding every launch.
+// onboarding: the first usable (inline credential, stored OAuth login, or
+// no-auth/local) non-local provider, else the first usable local one. It lets
+// the CLI fall back to an already-configured login when the active provider
+// happens to lack a credential, instead of re-running onboarding every launch.
 func firstUsableProvider(providers []config.ProviderProfile) (config.ProviderProfile, bool) {
+	logins := oauthLoggedInProviders()
 	var localFallback config.ProviderProfile
 	haveLocal := false
 	for _, profile := range providers {
@@ -371,7 +372,11 @@ func firstUsableProvider(providers []config.ProviderProfile) (config.ProviderPro
 				continue
 			}
 		}
-		if _, missing := setupMissingCredentialEnv(profile); missing {
+		// A stored OAuth login (e.g. `zero auth login xai`) is a credential too, even
+		// when the profile has no inline key / env var — mirrors setupRequired and
+		// usableSavedProviders so this fallback doesn't force onboarding for a
+		// provider the user is already authenticated with.
+		if _, missing := setupMissingCredentialEnv(profile); missing && !providerHasOAuthLogin(profile, logins) {
 			continue
 		}
 		if providerProfileIsLocal(profile) {
--- a/internal/cli/setup_fallback_test.go
+++ b/internal/cli/setup_fallback_test.go
@@ -1,9 +1,12 @@
 package cli
 
 import (
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/oauth"
 )
 
 func TestFirstUsableProviderPrefersRemoteKeyed(t *testing.T) {
@@ -63,6 +66,31 @@ func TestFirstUsableProviderSkipsUnresolvableCatalogWithoutBaseURL(t *testing.T)
 	got, ok := firstUsableProvider(providers)
 	if !ok || got.Name != "custom" {
 		t.Fatalf("want custom-endpoint fallback, got %q ok=%v", got.Name, ok)
+	}
+}
+
+// An OAuth-only provider (no inline key, no env var) must be selectable as a
+// fallback, matching setupRequired/usableSavedProviders — otherwise a fully
+// authenticated user gets forced back into onboarding when activeProvider
+// goes stale.
+func TestFirstUsableProviderRecognizesOAuthLogin(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "tok.json")
+	t.Setenv("ZERO_OAUTH_STORAGE", "file") // an inherited "keyring" would ignore the temp path and hit the OS keychain
+	t.Setenv("ZERO_OAUTH_TOKENS_PATH", path)
+	store, err := oauth.NewStore(oauth.StoreOptions{FilePath: path})
+	if err != nil {
+		t.Fatalf("NewStore: %v", err)
+	}
+	if err := store.Save(oauth.ProviderKey("xai"), oauth.Token{AccessToken: "tok", ExpiresAt: time.Now().Add(time.Hour)}); err != nil {
+		t.Fatalf("seed token: %v", err)
+	}
+
+	providers := []config.ProviderProfile{
+		{Name: "xai", CatalogID: "xai", APIKeyEnv: "XAI_API_KEY"}, // no inline key/env, but logged in via OAuth
+	}
+	got, ok := firstUsableProvider(providers)
+	if !ok || got.Name != "xai" {
+		t.Fatalf("want OAuth-logged-in provider (xai), got %q ok=%v", got.Name, ok)
 	}
 }
 
--- a/internal/oauth/encrypt.go
+++ b/internal/oauth/encrypt.go
@@ -122,17 +122,17 @@ func createSecretFile(path string) ([]byte, error) {
 		if !errors.Is(err, os.ErrExist) && !errors.Is(err, os.ErrPermission) {
 			return nil, fmt.Errorf("oauth: create token secret lock: %w", err)
 		}
+		// Remember the lock-creation error itself rather than a subsequent
+		// "secret file doesn't exist yet" read error -- otherwise a real
+		// contention/ACL problem is masked by the expected-while-waiting
+		// ErrNotExist once the retries are exhausted.
+		lastErr = err
 		if data, rerr := readSecretFileRetry(path); rerr == nil {
 			return data, nil
-		} else {
-			lastErr = rerr
 		}
 		time.Sleep(secretRetryDelay)
 	}
-	if lastErr != nil {
-		return nil, fmt.Errorf("oauth: timed out waiting for token secret %s: %w", path, lastErr)
-	}
-	return nil, fmt.Errorf("oauth: timed out waiting for token secret lock %s", lockPath)
+	return nil, fmt.Errorf("oauth: timed out waiting for token secret %s: %w", path, lastErr)
 }
 
 func writeNewSecretFile(path string) ([]byte, error) {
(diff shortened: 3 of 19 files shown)


Vasanthdev2004 (approved):
## Approve — all 9 audit fixes verified real and correctly fixed; two notes + a merge-order

I put this through a 9-way independent audit (one reviewer per claimed bug) plus an adversarial verify pass on each, ground-truthing every fix against the pre-fix code on `main`. **All 9 are real defects with correct, net-improving fixes** — nothing fabricated, and the skeptic pass agreed `keep` on all nine. Regression tests present for 8/9 (the securefile one rides the existing concurrency test). Genuinely strong work, @PierrunoYT — thanks.

Highlights I confirmed by tracing the actual code:
- **grant-scope case bypass (the important one):** real. Pre-fix `grantCovers` used raw `==`/`HasPrefix`; on case-insensitive Windows a persisted specific-path **deny** could be slipped by re-casing the path (`C:\proj\secret.txt` vs `c:\…`) into a broader allow. The fix routes through the existing `filepath.Rel`-based `pathWithinRoot` (folds case via `EqualFold` on Windows, byte-identical on Unix) — reuses the codebase's established helper rather than inventing logic. Correct.
- **securefile lock:** real — #445's Windows `ERROR_ACCESS_DENIED` retry was only in the oauth store, not the shared credstore backend. The fix faithfully mirrors #445 (doesn't revert it) and stops masking contention behind a misleading "file does not exist" timeout.

Two non-blocking things worth folding in:

1. **Reclassify two of the "critical" update fixes.** `update-checksum-filename` and `update-zip-symlink` are real code gaps and the fixes are correct, but they're **defense-in-depth, not live vulns** — both fail safe on `main` today (no `os.Symlink` is ever materialized, `safeExtractPath` already blocks name traversal, extraction is SHA256-gated, and the checksum-filename path force-aborts via `assertSafeChecksumFileName`). Worth landing — just soften the "critical/supply-chain" framing so they aren't read as fixing an exploitable hole.
2. **The keybinding-collision fix has a coverage gap (follow-up, not a blocker).** `reservedBindings` protects `ctrl+c/esc/enter/shift+tab/ctrl+f/?` but omits the **bare-chord** hardcoded handlers — `tab`, `backspace`, `up`/`down`, `pgup`/`pgdown`. Those are config-settable and matched with `Mod==0`, so binding e.g. `toggleDetailed="tab"` would still silently shadow Tab navigation (`="up"` kills history recall) — the exact class this fix closes, uncaught and unwarned. The fix still strictly improves on `main` and covers the critical keys, so ship it; just file a follow-up to extend the reserved set.

**Merge order — land this after the euxaristia shell cluster (#476 → #468 → #465).** #481's `shell-preflight-quotes` hunk makes detection quote-aware by running the patterns against `masked := stripQuotedSpans(trimmed)` instead of raw `trimmed`. Open #476/#468 are actively reworking those very detectors in `shell_runtime.go`. If #481 rebases **last**, the `masked` wrapper cleanly covers whatever final detector set the cluster produces — with the explicit rule that **any new detector pattern must match against `masked`, not `trimmed`**, or quoted content silently gets re-flagged. #481's other 8 fixes are in disjoint files and can land anytime. (If the cluster stalls, land #481 now and make the cluster own the `shell_runtime.go` reconciliation with that same `masked` requirement.)

Approving on the merits — CI's green on all 7 checks and every fix holds up.

jatmn (changes requested):

## Findings

- [P2] Include default and bare-key shortcuts in keybinding collision checks
  `internal/tui/keybindings.go:274`
  The new sanitizer only compares explicitly configured bindings against `reservedBindings` and the other non-zero configured bindings, so it still misses collisions with an action's default shortcut and with hardcoded bare-key handlers that are not in `reservedBindings`. For example, `toggleDetailed = "ctrl+t"` is accepted with no warning even though the dispatch switch checks `toggleDetailed` before the default `cycleReasoning` `Ctrl+T` handler, making cycle-reasoning unreachable; similarly `toggleDetailed = "tab"` is accepted and shadows the later Tab navigation/prompt handling. Please model the default bindings and the remaining hardcoded bare-key shortcuts in the collision set so the class of silent shortcut shadowing this PR is fixing is actually closed.

- [P2] Add regression coverage for the securefile lock-contention fix
  `internal/securefile/securefile.go:126`
  This patch changes the encrypted-file credential backend to treat Windows `os.ErrPermission` from the secret lock as contention, but the PR does not add or update a `securefile`/`credstore` regression test for that path; only the implementation file changes. The existing OAuth concurrency test exercises the separate OAuth copy of this logic, not the shared `securefile` backend that stores provider API keys, so this Windows-only credential-storage fix can regress while the new test coverage still passes. Please add focused coverage for the securefile/credstore branch, or factor the lock acquisition so the `os.ErrPermission` contention behavior is testable without relying on a real Windows delete-pending race.

jatmn (approved):
LGTM
@Vasanthdev2004

kevincodex1:
please rebae to main and fix conflicts @PierrunoYT

Vasanthdev2004:
These are real fixes — each one mirrors a pattern we already use elsewhere in the tree, and they all come with regression tests, so no complaints about the code. The problem is packaging: nine unrelated fixes across nine packages under a single non-approved issue is hard to review and land as one unit, and the branch is conflicting now anyway.

Could you rebase and split it? Group the security-sensitive ones (sandbox grant path-casing, the update checksum filename cross-check, the zip symlink rejection) separately from the UX/robustness ones (OAuth fallback, UTF-8 truncation, keybinding collision, the quote-aware shell check, the Windows retry paths), and file an issue per group. Once those are approved I can take them one at a time — they'll move a lot faster reviewed in small focused pieces.

PierrunoYT:
Split this into two focused PRs per the review above, each with its own issue:

- Security-sensitive: #545 → #547 (sandbox grant path-casing, securefile/oauth Windows lock-contention consistency, update checksum filename cross-check, zip symlink rejection)
- UX/robustness: #546 → #548 (OAuth fallback, UTF-8 truncation, keybinding collision detection, Windows rename retry)

Both are rebased onto current `main` and pass `go build ./...` / package-scoped tests independently.

One note: the quote-aware shell check (originally the 9th finding here) isn't in either split PR — #476 landed on `main` independently since this PR was opened and already replaced that pre-flight check with a more thorough segment/word-anchored implementation covering the same false-positive, so this branch's version of that fix is now fully superseded and there's nothing left to carry over.

Happy to close this PR once the split ones look good, or leave it open for reference — your call.

kevincodex1 (approved):
LGTM

euxaristia:
goated PR

PR #35, opened 2026-06-03
feat: add config validation diagnostics

gnanam1990:
## Summary
- Adds structured config validation diagnostics for user, project, and environment layers, including source, file path, field path, and diagnostic kind.
- Keeps invalid config files non-fatal while preserving valid loaded layers in the effective config.
- Updates `zero config` inspection output and JSON output so automation can identify exact invalid config fields without noisy import-time warnings.

## Validation
- `npx --yes bun install --frozen-lockfile`
- `npx --yes bun run typecheck`
- `npx --yes bun test ./tests --timeout 15000`
- `npx --yes bun run build`
- `npx --yes bun run smoke:build`
- `./zero --help`
- `HOME=$(mktemp -d) ./zero config --json`
- Built CLI invalid-config JSON diagnostic probe
- `git diff --check origin/main..HEAD`

Reviewers: @vasanthdev2004 @anandh8x

PR #34, opened 2026-06-03
Add automated PR review workflow

Vasanthdev2004:
## Summary
- Add a `PR Auto Review` workflow that runs diff hygiene, typecheck, tests, build, and smoke build on PR updates.
- Add a Bun review script that posts or updates one stable automated review comment with blockers and validation status.
- Add focused tests for outcome normalization and review markdown formatting.

## Validation
- `bun run typecheck`
- `bun test ./tests/pr-review.test.ts --timeout 15000`
- `bun test ./tests --timeout 15000`
- `bun run build`
- `bun run smoke:build`
- dry-run `bun run review:pr`

PR #33, opened 2026-06-03
docs: comprehensive project README

Vasanthdev2004:
Replaces the stub README with a full, accurate overview of Zero as it stands on `main`.

Includes:
- Rebel ASCII wordmark + tagline + badges
- Highlights (multi-provider, TUI, headless, tools, permissions, sessions, doctor)
- Quick start + provider/model configuration
- Usage: interactive TUI, headless `exec` (incl. stream-json), and the `providers`/`search`/`doctor`/`config`/`update` commands
- Real provider/model table and tool table
- Architecture overview + project layout
- Development/release commands and a docs index

All commands, flags, providers, models, and module paths were taken from `main` (HEAD `aee8b82`), not invented.

PR #32, opened 2026-06-03
feat: add Zero exec resume and fork sessions

gnanam1990:
## Summary
- add a Zero exec session resolver for new, resumed, and forked runs
- extend the session event store with latest-session lookup, fork lineage metadata, and append-only event copying
- wire `zero exec --resume [id]` and `zero exec --fork <id>` into headless runs with persisted user, assistant, tool, usage, and crash events
- include `sessionId` in stream-json `run_start` events and update the protocol example

## Why
This gives headless Zero runs a durable session lifecycle so later CLI and TUI flows can continue prior context or branch from an existing run without losing lineage.

## Validation
- `npx --yes bun test ./tests --timeout 15000`
- `npx --yes bun run typecheck`
- `npx --yes bun run build`
- `npx --yes bun run smoke:build`
- `./zero exec --help`
- `OPENAI_API_KEY= ./zero exec hi --resume abc --fork def --output-format stream-json`
- `git diff --check origin/main..HEAD`

## Reviewers
Please review: @vasanthdev2004 @anandh8x

PR #484, opened 2026-07-04
polish v2: smoothness & delight

anandh8x:
Polish Roadmap v2 — the "smoothness & delight" pass. Where the prior context-economy PR cut per-turn **token** overhead, this PR cuts **wall-time** overhead and wires Zero's existing diagnostic machinery to the live failure points where users actually land.

Each item lands as its own commit so they can be reviewed (or reverted) independently. Every roadmap claim was spot-verified against the code before implementation. (Item numbers below are roadmap IDs, not GitHub issues.)

### Shipped
**Error recovery**
- **`#1` Actionable provider/model error hints** — classify a failed turn's provider error (auth / rate-limit / connectivity / model-not-found / context-overflow) and render a one-line next step under the red error row, instead of an identical raw blob for every failure. New shared `internal/errhint` package, gated on a provider-origin marker so local errors never draw a bogus hint.
- **`#16` Gemini 401 auth-retry** — plumb `OAuthResolver` through the Gemini provider + factory and switch to `SendWithAuthRetry`, so Gemini OAuth users recover from an expired token (force-refresh + replay on 401) instead of a terminal "API key not valid". Parity with OpenAI/Anthropic.
- **`#20` CLI exec provider error hint** — append a one-line `zero auth` / `zero doctor` hint to text-mode `zero exec` provider errors, reusing the `#1` classifier.

**Per-prompt overhead**
- **`#3` bash output budget** — cap bash stdout/stderr at 96KB each with head+tail truncation (build/test failures live at the tail), recording raw/emitted/truncated in Meta + `Result.Truncated` like the other tools. bash was the only tool with no byte cap.
- **`#5` reuse LSP manager across prompts** — build one `lsp.Manager` per session (servers start lazily) instead of a fresh one per run, so gopls stays warm between turns instead of cold-starting on the first edit of every prompt.
- **`#7` coalesce streamed text deltas** — batch `OnText` deltas over ~16ms at the runtime sink so render rate decouples from token rate; a fast provider no longer drives 100+ Update→View cycles/sec.
- **`#10` streamingText O(1) append** — accumulate live assistant text as `[]byte` instead of `string +=`, killing the O(n²) allocation on long generations.

**Reliability / recovery tuning**
- **`#17` tool-failure stop 4→6** — a corrective hint fires at 2 failures; give a model iterating on a tricky edit more room before halting. Streak still resets on success or a different error.
- **`#19` reconnect backoff + jitter** — `maxStreamReconnects` 2→4 with up-to-50% jitter (capped 8s), riding out a multi-second blip and de-synchronizing swarm/cron reconnects.
- **`#33` stall retries 2→1** — a no-output stall is only detected after the full ~5min idle timeout, so 2 retries froze a session ~15min; one retry bounds it to ~2× idle.
- **`#34` narrow reconnect matching** — drop `502`/`503` substring matches (503 already exhausted `SendWithRetry`; 502 is non-idempotent), reconnect only on genuine transport failures.

_`#32` (stream max-turns final answer) was already implemented — `finalAnswerAfterMaxTurns` forwards `OnText`/`OnReasoning` today; no change needed._

**Flow friction (commands)**
- **`#18` /compact now** — accept the `now` keyword (bare `/compact` already triggered; `/compact now` used to error).
- **`#12` /retry and /edit** — resend the last prompt, or recall it into the composer to tweak and resend.
- **`#23` /copy and /export** — copy the last answer to the clipboard; write a plain-text transcript to a file. Real gaps for SSH users with no mouse-select.

**Agent-quality polish**
- **`#31` separator-insensitive tool_search** — `webfetch` now matches `web_fetch` (squash-separators fallback ranked below exact substring), so a model that drops the underscore doesn't loop.
- **`#30` ask_user skip signal** — a wholesale prompt dismissal is flagged as a skip up front (vs a single `(left blank)` field), so the model doesn't invent a default.

**Already satisfied by current code (no change):** `#13` (`? shortcuts` already advertised by the composer idle-hint; the "allow `?` on a non-empty composer" idea would hijack typing "?"), `#28` (`OnContext` is never wired, so `MeasureContext` already never runs per-turn; `/context` computes on demand), `#32` (max-turns answer already streams).

_Merged latest `main` (resolves the #478 conflict); no force-push._

**More flow polish**
- **`#11` fuzzy picker ranking** — `/model` and `/resume` now rank matches (exact < prefix < contains < subsequence) instead of a flat substring filter; `snt45` matches `Sonnet 4.5`. Groups stay contiguous so the grouped model picker keeps one header per provider.

**Already satisfied / held:** `#27` (turn-end already nils `lineAges` via `resetStreamingFade` + `streamingText` via `#10`); `#24` **held for decision** — its remaining win (sharing image bytes) would break `TestCopyMessagesDeepCopiesImageBytes`, a deliberately-tested anti-aliasing invariant.

**Perf + agent-quality (round 3)**
- **`#22` cache per-tool schema render** — partitionTools no longer re-runs the recursive schema→map conversion for every tool every turn; a per-run cache keyed by tool name skips it. Partitioning stays fresh each turn (deferred state can flip mid-run).
- **`#25` calibrate compaction estimate** — fold each turn's (raw estimate, provider `InputTokens`) into an EMA ratio so compaction triggers near true capacity instead of ~15% early on code-heavy history.
- **`#26` turn-based plan staleness** — the stale-plan nudge now also fires after 8 turns without an update while items are pending (not just 10 tool calls), catching a plan that drifts stale across many low-tool-call turns. The roadmap's fuzzy relevance-match is omitted (too false-positive-prone).

**Round 4 (flow + agent-quality)**
- **`#35` preserve recent edits across compaction** — carry write_file/edit_file targets + a one-line note per file in the preserved-state block (capped 20 paths), so after compaction the model knows what it changed without re-reading.
- **`#21` workspace-scoped /resume** — filter the picker (and `/resume latest`) to the current workspace before the per-session event read (no more N full file reads for a global history); skip zero-event sessions via metadata. Empty-Cwd/older sessions stay visible.
- **`#14` image-drop warning on model switch** — chips already render above the composer; add an immediate amber warning when switching to a non-vision model with images staged (was silently dropped at submit).

**Held (needs registry/concurrency work — maintainer decision):** `#15` async MCP connect — `tools.Registry` has no mutex, so background registration races the agent loop's per-turn `registry.All()`, and it breaks the deferral-gate timing invariant. Not a safe drop-in.

## Summary by CodeRabbit

* **New Features**
  * Added clearer, actionable guidance for provider and CLI errors.
  * Improved the TUI with next-step hints, better retry/edit/export commands, and a more helpful onboarding error message.
  * Added transcript export support and workspace-aware session resume browsing.
  * Updated picker search to rank better matches more intelligently.

* **Bug Fixes**
  * Improved streamed output handling and reduced UI churn during assistant responses.
  * Added smarter reconnect behavior for interrupted streams.
  * Preserved recent file edit context during compaction and improved plan reminder timing.


Vasanthdev2004 (approved):
Went through all ten commits and they hold up. Each one's a real thing: bash was the only tool without an output cap, Gemini was the only OAuth provider that couldn't refresh an expired token (you'd get a dead "API key not valid" while OpenAI/Anthropic silently retry), streamingText really was doing O(n²) string appends, and the per-run LSP manager cold-started gopls on the first edit of every prompt. One commit per item, each with a test, so it's easy to reason about.

A few small things, none blocking:

- The 16ms text coalescer calls `forward()` outside the mutex, so in theory the flush timer and a non-text emit can reorder and render a trailing narration fragment after the tool card it should sit before. The window is nanoseconds wide and the committed transcript comes from `result.FinalAnswer`, not the streamed buffer, so it's cosmetic — worth a code comment, not worth holding this.
- The bash commit message says shell-issue detection runs on the raw pre-truncation output, but it actually runs on the budgeted text. Harmless (head+tail keeps the first/last error lines the detector looks for), just an inaccurate note.
- The "N bytes omitted" marker undercounts slightly. Cosmetic.

Approving — good pass.

Vasanthdev2004:
@coderabbitai full review

anandh8x:
@coderabbitai  review the commits that u havent checked before

jatmn (changes requested):

## Findings

- [P2] Preserve attachments when retrying the previous prompt
  `internal/tui/model.go:3952`
  `/retry` now relaunches only `m.lastPrompt`, but `lastPrompt` stores just the text while the original submit path consumes PDF text (`consumePendingDocuments`) and passes staged images via `turnImages` before clearing both attachment queues. If a vision or document-backed prompt fails and the user runs `/retry`, the retried request silently loses the image/PDF context even though the command is documented as resending the last prompt, so the model answers a different task. Please carry the submitted attachments/document preamble with the remembered prompt, or make `/retry` refuse/clearly warn when the last prompt depended on attachments.

- [P3] Complete CodeRabbit's request to update the OAuthResolver provider comment
  `internal/providers/factory.go:27`
  CodeRabbit's earlier request is still valid: this comment says `OAuthResolver` is applied only to OpenAI and Anthropic, but the current factory also passes it to `gemini.New` in the Google provider branch. Since this PR is specifically adding Gemini OAuth retry wiring, please update the provider-factory documentation so future auth work does not rely on a stale provider boundary.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found issues that still need to be addressed.

## Findings

- [P2] Complete CodeRabbit's request to keep 504 out of reconnect retries
  `internal/agent/reconnect.go:123`
  CodeRabbit's 504 reconnect concern is still valid on the current head. The new reconnect comment says HTTP 500/502/504 responses are non-idempotent and must not be replayed because the completion POST may already have been processed, but `shouldReconnect` still matches the generic `"timeout"` substring, so an error like `504 Gateway Timeout` returns true and reissues the same completion request. Please add an explicit HTTP-status guard before the transport substring loop and update `TestShouldReconnectClassification` so 504 stays non-reconnectable.

- [P2] Preserve attachments when editing and resending the previous prompt
  `internal/tui/model.go:3974`
  The `/edit` command recalls only `m.lastPrompt` into the composer, while the image/PDF snapshot captured for that same last prompt is left out of the pending attachment queues. After a vision or document-backed prompt, `/retry` now correctly re-stages `m.lastImages`/`m.lastDocuments`, but `/edit` followed by Enter silently sends a text-only version of the edited prompt. That makes the documented "edit and resend" flow answer a different task whenever the original prompt depended on attachments. Please either re-stage/show the remembered attachments for `/edit`, or warn/refuse when the previous prompt had attachments that cannot be carried through the edit flow.

- [P2] Keep recently re-edited files in the preserved edit summary
  `internal/agent/compaction_preserve.go:462`
  The recent-edit preservation cap can drop a file that was just edited again. `recentEdits` records path order only the first time a file appears, `mergeSkillEntries` keeps an older preserved path in its old position even when a fresh edit updates its note, and `capRecentEdits` then keeps only the tail of that old order. In a long editing session, once more than 20 distinct files are tracked, re-editing an early file can still leave it outside the capped tail, so the next compaction omits exactly the file the model most recently touched. Please move paths to the newest position when a fresh edit arrives, and add a regression covering a capped list where an old path is edited again.

- [P3] Complete CodeRabbit's request to make the coalescer batching test deterministic
  `internal/tui/coalesce_test.go:41`
  The test still starts the real 16ms `time.AfterFunc` and then asserts that nothing has been forwarded before the manual `c.flush()`. If the scheduler or GC pauses the test goroutine for longer than that interval between the sends and the snapshot, the timer goroutine can legally flush the buffered text first and make this test fail nondeterministically. Please add a test seam for the coalescer interval/timer or otherwise structure this assertion so it does not depend on beating a wall-clock timer.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P2] Bound bash output before it enters memory
  `internal/tools/bash.go:127`
  The new 96KB bash budget is applied only after `exec.Cmd` has already written all stdout/stderr into unbounded `bytes.Buffer`s and after both buffers are converted to strings. A command like `cat` on a huge file can still grow the Zero process until it stalls or OOMs before `budgetBashOutput` ever runs, even though the PR description says bash stdout/stderr are capped. Please cap the capture writer or stream through a bounded buffer before handing stdout/stderr to the formatter, so the tool is protected from large output as well as from large model-visible output.

jatmn (approved):
Thanks for the updates. I rechecked the previously discussed paths and do not see any remaining actionable issues from my side.


@Vasanthdev2004 LGTM

Vasanthdev2004 (approved):
Re-approving after the latest round (my earlier approval went stale on the new pushes).

I did an independent pass over the parts that worried me most — the core-loop changes where green CI and unit tests can still hide a hang, a leak, or dropped output. All of it holds up:

- **streaming** — the coalescer keeps delta order (the lock spans drain+forward) and the committed final answer stays authoritative, so no dropped or reordered text; the flush timer is reused, no per-turn goroutine/timer leak. The `[]byte` accumulator is copy-on-read, no aliasing.
- **compaction** — the EMA ratio is div-by-zero-safe and clamped to `[0.5, 2.0]`, so a bad turn can't push compaction to fire way early or never; the preserved recent-edits block is hard-capped (20 paths, short notes) with no bypass.
- **reconnect/stall/tool-failure** — jitter can't go negative or unbounded, retry counts are capped, `#34`'s narrowing doesn't strand a recoverable stream, and the 4→6 tool-failure stop has no off-by-one.
- **bash budget** — capture is genuinely bounded in memory now (not just the model-visible slice), truncation is UTF-8-safe and keeps the failure tail where build/test errors live.
- **LSP reuse** — built once per session, nil-safe, torn down in `quit()`, and cwd is process-fixed so `/new` and `/resume` don't point it at the wrong root.
- **errhint/Gemini** — the origin gate fails closed (a local error never draws a provider hint) and the 401 retry is single-shot with a rebuilt body, no loop.

On the whole PR: it's a lot to land at once (55 files), but it's the right kind of a lot — every item is its own commit with its own tests, each is a real wall-time or UX win rather than churn, and it's already been through three rounds with jatmn. Worth merging.

One coordination note: the bash output cap here (`#3`) covers the same ground as #492, so that one can be closed or folded once this lands.

Can't merge it myself — over to kevin.

gnanam1990 (approved):
LGTM — approving. Ran an independent regression pass on top of the existing reviews.

**Local gate (checked out `polish/v2-smoothness`):**
- `go build ./...` ✅ · `go vet ./...` ✅ · `gofmt -l` clean ✅
- `go test ./... -race` → **74 packages, 0 failures, 0 races** ✅

**Deep-dived the streaming coalescer** (`internal/tui/coalesce.go`) since it's the one item with new concurrency + a 16ms buffer. Traced the worst case — the final ≤16ms text batch losing the race to `agentResponseMsg` and getting dropped as a stale-run delta — and confirmed it's **safe**: on the success path the final answer is committed from the runtime's complete `final` `rowAssistant` in `msg.rows`, which supersedes the streamed `streamingText`, so a dropped tail only affects the last live frame. The error path keeps `streamingText` as a best-effort partial, where a 16ms truncation is immaterial. The mutex-held-across-forward design also correctly prevents streamed text from reordering around non-text (tool/reasoning) messages. The "harmless late flush" comment holds.

**Everything else:** the reconnect narrowing (502/503/504) and the bash memory-bound cap — the P2s from the earlier rounds — are addressed, and the full race suite passing gives me confidence the core-loop changes are clean. One-commit-per-item plus near-total test coverage made this easy to reason about despite the size.

Nothing blocking from me. The two remaining CodeRabbit inline notes (`errhint.go:81`, `bash.go:181`) are minor quick-wins — optional.

PR #485, opened 2026-07-04
Fix action summary for structured output

### LINKED ISSUE #483: GitHub Action summary returns run_end/done JSON instead of final answer
### Version / branch / commit

main @ 949ee43f71e5cb7fab4695c5cb7b442fe4ecfbf7

### OS and environment

GitHub Actions composite action, default output-format=stream-json

### Steps to reproduce

1. Run ZERO GitHub Action with default inputs.
2. Let `zero exec` finish successfully.
3. Read `steps.zero.outputs.summary`.

### Expected behavior

`summary` contains final assistant answer text.

### Actual behavior

`summary` contains trailing lifecycle JSON, e.g. `{"type":"run_end",...}`.  
Same issue with `output-format=json`, where last line is `{"type":"done",...}`.

_No response_

GautamBytes:
Fixes #483. The action now reads final/error events for structured output instead of tailing run metadata. 

## Summary by CodeRabbit

* **New Features**
  * Improved one-line summary generation using a structured-output-aware approach, with safer fallback when structured parsing isn’t possible.
  * The action now provides an `ACTION_PATH` environment variable during runtime.
  * Added a command-line summary helper to generate summaries directly from output files.

* **Bug Fixes**
  * Correctly prefers the most relevant structured event (including later error events) instead of relying on the last captured non-empty line.

* **Tests**
  * Expanded test coverage for JSON/streamed/text modes, malformed structured input, truncation/normalization, and both CLI invocation styles.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -180,6 +180,7 @@ runs:
         ZERO_INPUT_ADD_DIR: ${{ inputs.add-dir }}
         ZERO_INPUT_WORKTREE: ${{ inputs.worktree }}
         ZERO_INPUT_OUTPUT_FORMAT: ${{ inputs.output-format }}
+        ACTION_PATH: ${{ github.action_path }}
         RUNNER_TEMP: ${{ runner.temp }}
       run: |
         set -euo pipefail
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (changes requested):
Core fix is right — reading the final/error events instead of tailing metadata matches the event schema, and the tests cover the #483 case well.

One thing to fix first: the new summary line in `action.yml` runs under `set -e` without the `|| true` the old grep/tail line had. If node exits non-zero — module not found, or no node on a self-hosted runner — the `summary="$(node ...)"` assignment aborts the step before it writes the output and before `exit "${code}"`, so the step reports the helper's failure instead of Zero's real exit code. It won't fire on GitHub-hosted runners where node's always there, but it's a real drop in defensiveness. Append `|| true` (or wrap the call in `set +e`/`set -e`). CodeRabbit flagged the same one.

Fix that and it's good to go.

GautamBytes:
done @Vasanthdev2004

jatmn (approved):
LGTM

@Vasanthdev2004

kevincodex1 (approved):
LGTM

kevincodex1:
please have a look again @Vasanthdev2004

Vasanthdev2004 (approved):
Re-reviewing — my one blocker is fixed. The summary line has the `|| true` now, so a node failure (missing module, no node on a self-hosted runner) degrades to an empty summary instead of aborting the step under `set -euo pipefail`, and `ACTION_PATH` is wired so the script actually resolves.

The core fix is right and I checked it holds: `summarizeStructured` walks the stream-json events and takes the last `final` event's text (or an `error` event's message) rather than tailing the last line — which was the `run_end` lifecycle metadata #483 was showing. Non-structured output keeps the old last-line behavior, both paths stay bounded at 280, and it falls back to the last non-empty line if no final/error event is present. The dedicated `action-summary.test.mjs` covers it, and CI is green including the action.yml validation.

Marked #483 `issue-approved` — real bug, clean fix.

Good to merge — kevin's already signed off too.

PR #486, opened 2026-07-04
fix(openai): handle Ollama reasoning stream deltas

### LINKED ISSUE #439: Ollama thinking models (qwen3): reasoning deltas are invisible, runs die with "3 turns with no output"
## Summary

With Ollama-served thinking models (tested: `qwen3:4b`), `zero exec` reliably dies with
`Agent stopped after 3 turns with no output (no visible text and no tool calls)` on tasks that
trigger longer thinking (e.g. file-write tasks), while the model is in fact generating
1,300–1,600 tokens per turn. Read-only tasks on the same setup work fine.

The cause appears to be that Ollama streams thinking deltas in a field named `reasoning`,
which zero's chat-completions stream parser does not read.

## Environment

- zero 0.1.0 (npm `@gitlawb/zero`, macOS arm64, Apple M5 Max)
- Ollama 0.31.1, `qwen3:4b`, default localhost:11434
- Provider configured via `zero providers add ollama --model qwen3:4b --set-active`

## Repro

```bash
zero exec --output-format stream-json "Create a file named hello2.txt containing exactly: hi"
```

Result (4/4 attempts, also with a `/no_think` prompt prefix and with an explicit
"Use the edit_file tool" instruction):

- stream-json contains only `run_start`, 3× `usage`, `final`, `run_end` — zero `tool_call` events, zero `text` events
- final text: `Agent stopped after 3 turns with no output (no visible text and no tool calls) to avoid consuming tokens without making progress.`
- meanwhile the Ollama server log shows each turn returned HTTP 200 and decoded 1,324–1,558 tokens

A read-only task on the same setup works (single well-formed `read_file` tool call):

```bash
zero exec --output-format stream-json "Read calc.py and report the bug in the divide function."
```

## Evidence that the model is not at fault

Direct call to the same runtime/model with a minimal tool schema returns a clean native tool call:

```bash
curl -s http://localhost:11434/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"qwen3:4b",
  "messages":[{"role":"user","content":"Create a file hello.txt with content hi. Use the edit_file tool."}],
  "tools":[{"type":"function","function":{"name":"edit_file","description":"Create or edit a file",
    "parameters":{"type":"object","properties":{"path":{"type":"string"},"content":{"type":"string"}},
    "required":["path","content"]}}}],
  "max_tokens":2048}'
```

Response: `finish_reason: "tool_calls"`, message keys `['role','content','reasoning','tool_calls']`,
`content` empty, thinking text in `reasoning`, and valid
`tool_calls: [{"function":{"name":"edit_file","arguments":"{\"path\":\"hello.txt\",\"content\":\"hi\"}"}}]`.

## Suspected cause

`internal/providers/openai/types.go` — `streamDelta` parses only:

```go
type streamDelta struct {
    Content          string                `json:"content"`
    ReasoningContent string                `json:"reasoning_content"`
    ToolCalls        []streamToolCallDelta `json:"tool_calls"`
}
```

Ollama emits thinking as `reasoning` (not `reasoning_content`), so every thinking delta is
dropped. The turn produces no visible progress for zero, and after three such turns the
no-output guardrail (`internal/agent/guardrails.go`) stops the run. The content-stall
handling presumably cuts the turn before the trailing `tool_calls` arrive, which would
explain why the direct curl (no stall handling) gets the tool call but zero never does.

## Suggested fix

- Parse `reasoning` as an alias of `reasoning_content` in `streamDelta` (and the
  non-stream message struct).
- Count reasoning deltas as turn liveness for stall detection / the no-output guardrail,
  even if they are not rendered as user-visible text.

This matters for local-model use: qwen3-family (and other thinking models) are among the
most capable tool-calling models Ollama users are likely to point zero at.

Ashwinhegde19:
Fixes #439

## Summary
- parse OpenAI-compatible streamed `delta.reasoning` as an alias for `delta.reasoning_content`
- emit alias reasoning deltas as reasoning events so long-thinking Ollama models count as active
- preserve existing `reasoning_content` precedence when both fields are present

## Tests
- `go test ./internal/providers/openai -count=1`
- `make lint`
- `make test-quick`

## Summary by CodeRabbit

* **New Features**
  * Stream-JSON output now emits separate `reasoning` events, and the CLI routes reasoning output during headless `exec` runs.
* **Bug Fixes**
  * Improved reasoning streaming mapping to handle alternate delta fields and correctly prioritize `reasoning_content`.
  * Updated no-output guardrails so reasoning-only turns count as progress (preventing premature stopping).
* **Documentation**
  * Expanded Stream-JSON protocol docs with the new `reasoning` event behavior and examples.
* **Tests**
  * Added unit and CLI integration coverage for reasoning alias/prioritization, guardrail behavior, and `exec` Stream-JSON reasoning ordering.

### DIFF
--- a/docs/STREAM_JSON_PROTOCOL.md
+++ b/docs/STREAM_JSON_PROTOCOL.md
@@ -41,6 +41,7 @@ events when present.
 
 ```json
 { "schemaVersion": 2, "type": "run_start", "runId": "run_20260603_abc123", "sessionId": "zero_20260603100000_abc123", "cwd": "/repo", "provider": "openai", "model": "gpt-4.1", "apiModel": "gpt-4.1" }
+{ "schemaVersion": 2, "type": "reasoning", "runId": "run_20260603_abc123", "delta": "Thinking..." }
 { "schemaVersion": 2, "type": "text", "runId": "run_20260603_abc123", "delta": "..." }
 { "schemaVersion": 2, "type": "tool_call", "runId": "run_20260603_abc123", "id": "call_1", "name": "read_file", "args": { "path": "README.md" }, "sideEffect": "read" }
 { "schemaVersion": 2, "type": "permission_request", "runId": "run_20260603_abc123", "id": "call_2", "name": "write_file", "action": "prompt", "permission": "prompt", "permissionMode": "ask", "sideEffect": "write", "reason": "Creates or overwrites files." }
@@ -50,6 +51,10 @@ events when present.
 { "schemaVersion": 2, "type": "final", "runId": "run_20260603_abc123", "text": "..." }
 { "schemaVersion": 2, "type": "run_end", "runId": "run_20260603_abc123", "status": "success", "exitCode": 0 }
 ```
+
+`reasoning` events carry live model reasoning/status deltas for providers that
+stream them separately from answer text. They are liveness/progress events only:
+they are not folded into `text` or the final answer.
 
 Permission events may include structured sandbox metadata:
 
--- a/internal/agent/guardrails.go
+++ b/internal/agent/guardrails.go
@@ -414,8 +414,9 @@ func (state *guardState) observeToolResult(name string, failed bool, output stri
 func (state *guardState) observeTurn(collected zeroruntime.CollectedStream) (stop bool) {
 	hasToolCalls := len(collected.ToolCalls) > 0
 	hasVisibleText := strings.TrimSpace(collected.Text) != ""
+	hasReasoning := collected.HasReasoning || len(collected.ReasoningBlocks) > 0
 
-	if hasToolCalls || hasVisibleText {
+	if hasToolCalls || hasVisibleText || hasReasoning {
 		state.emptyTurns = 0
 	} else {
 		state.emptyTurns++
--- a/internal/agent/guardrails_test.go
+++ b/internal/agent/guardrails_test.go
@@ -22,6 +22,14 @@ func textTurn(content string) []zeroruntime.StreamEvent {
 	}
 }
 
+// reasoningTurn produces live reasoning without visible assistant text.
+func reasoningTurn(content string) []zeroruntime.StreamEvent {
+	return []zeroruntime.StreamEvent{
+		{Type: zeroruntime.StreamEventReasoning, Content: content},
+		{Type: zeroruntime.StreamEventDone},
+	}
+}
+
 // toolTurn produces a turn that calls a named tool with the given args JSON.
 func toolTurn(callID string, toolName string, args string) []zeroruntime.StreamEvent {
 	return []zeroruntime.StreamEvent{
@@ -98,6 +106,31 @@ func TestRunResetsEmptyTurnCounterOnVisibleOutput(t *testing.T) {
 	}
 	if result.FinalAnswer != "here is real progress" {
 		t.Fatalf("expected the visible text as final answer, got %q", result.FinalAnswer)
+	}
+}
+
+func TestRunResetsEmptyTurnCounterOnReasoning(t *testing.T) {
+	provider := &mockProvider{
+		turns: [][]zeroruntime.StreamEvent{
+			reasoningTurn("thinking 1"),
+			reasoningTurn("thinking 2"),
+			reasoningTurn("thinking 3"),
+			textTurn("done"),
+		},
+	}
+
+	result, err := Run(context.Background(), "go", provider, Options{
+		Registry: tools.NewRegistry(),
+		MaxTurns: 12,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if result.FinalAnswer != "done" {
+		t.Fatalf("expected reasoning-only turns to keep the run live until final answer, got %q", result.FinalAnswer)
+	}
+	if len(provider.requests) != 4 {
+		t.Fatalf("expected 4 turns, got %d", len(provider.requests))
 	}
 }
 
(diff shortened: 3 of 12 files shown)


Vasanthdev2004 (commented):
The parsing change is correct — `reasoning` as an alias for `reasoning_content`, right precedence when both are present, emitted as a reasoning event so it's not mistaken for output. Tests are good.

But "Fixes #439" claims too much. This makes the thinking visible, which handles the "looks idle" half. The other half people hit — runs dying after 3 empty turns — comes from `observeTurn` in guardrails.go, which this doesn't touch: reasoning deltas go to `OnReasoning` and never get folded into the turn's text, so a reasoning-only turn still counts as empty, and three of them still trip `maxEmptyTurns` and stop with "no output". So merging this as-is would auto-close #439 while the main symptom's still there.

Either change "Fixes" to "Refs" so the issue stays open, or add the guardrail half (count a reasoning-bearing turn as live). The code itself is fine to merge — just don't let it close #439 yet.

Ashwinhegde19:
Updated, thanks for catching that. This now includes the guardrail half too: streamed reasoning deltas are recorded on the collected turn as liveness, and observeTurn resets the empty-turn counter when a turn has reasoning even if it has no visible answer text or tool call. I kept reasoning out of collected.Text, so it still does not become final-answer content.

Added regression coverage for three reasoning-only turns followed by a final answer, plus a collector assertion that reasoning marks the turn as reasoning-bearing.

Validated locally:
- go test ./internal/agent ./internal/zeroruntime ./internal/providers/openai -count=1
- make lint
- make test-quick

jatmn (changes requested):

## Findings

- [P2] Surface reasoning deltas through headless stream-json output
  `internal/cli/exec.go:506`
  The linked repro for #439 uses `zero exec --output-format stream-json`, but this PR only converts Ollama's `delta.reasoning` into the runtime `StreamEventReasoning`. The headless exec path still wires only `OnText` into the writer, `execEventWriter` only emits `text` events, and the stream-json event enum has no reasoning event type, so stream-json clients will still see no live output while the model is thinking. Please add a stream-json reasoning/thinking event (and wire `OnReasoning` in the exec paths) or otherwise surface these deltas in headless output so the reported path stops looking idle.

- [P2] Get the linked issue approved before merging this community PR
  `CONTRIBUTING.md:19`
  The contribution policy says community PRs must be tied to an existing issue that has already been reviewed and approved by the core team, with approval shown by the `issue-approved` label. This PR links `Fixes #439`, but #439 currently has no labels, so merging this would bypass the repository's issue-first gate for outside contributions. Please get #439 marked with `issue-approved` or have a maintainer explicitly confirm an exception before this PR is merged.

Ashwinhegde19:
Updated for the stream-json finding. The headless exec path now wires OnReasoning into execEventWriter, emits stream-json events with type reasoning and delta, and documents the event in docs/STREAM_JSON_PROTOCOL.md. The event is separate from text/final output, so reasoning stays liveness/progress only.

Added coverage in TestRunExecStreamJSONEmitsReasoningEvents to verify stream-json receives reasoning before text and that reasoning is not folded into the final answer.

Validated locally:
- go test ./internal/cli ./internal/streamjson ./internal/agent ./internal/zeroruntime ./internal/providers/openai -count=1
- make lint

I also started make test-quick, but it produced no output for several minutes after the early packages and I interrupted it rather than leaving a hung process. The earlier full quick suite had passed before this stream-json-only follow-up.

jatmn (approved):
@Vasanthdev2004 

Thanks for the update. I rechecked the changed paths and found one issue that still needs to be addressed before this is ready.

## Findings

- [P2] Complete the linked-issue approval before merging this community PR
  `CONTRIBUTING.md:19`
  The stream-json and guardrail fixes are now present, but the remaining contribution-policy blocker from the previous review is still unresolved: this PR links `Fixes #439`, and #439 still has no `issue-approved` label. The policy requires community PRs to be tied to an existing issue that has already been reviewed and approved by the core team, with approval shown by that label. Please get #439 marked with `issue-approved`, or have a maintainer explicitly confirm an exception, before this PR is merged.

Ashwinhegde19:
Updated in afd5d00 for CodeRabbit's ordering nitpick: TestRunExecStreamJSONEmitsReasoningEvents now asserts the reasoning event index is before the text event index, so the test catches buffering/order regressions instead of only checking presence.\n\nValidated locally:\n- go test ./internal/cli -count=1\n- git diff --check\n\nI intentionally left plain-text exec output silent for reasoning deltas. The existing behavior keeps reasoning separate from final/user-visible answer text; this PR's reported path is stream-json, and stream-json now receives live reasoning events.

Vasanthdev2004 (approved):
Approving.

I flagged earlier that "Fixes #439" over-claimed — the parsing made thinking visible, but reasoning-only turns still counted as empty and tripped `maxEmptyTurns`, so the runs-dying-after-3-turns half was untouched. That's handled now: `guardrails.go` counts a reasoning-bearing turn as live, with a test that runs three thinking-only turns before a final answer and stays alive. So `Fixes #439` holds up.

jatmn's stream-json point is covered too — there's a real `reasoning` event on the headless `exec` path now, emitted before `text`, kept out of the final answer, and documented as liveness-only.

I read the changed paths: `reasoning_content` precedence is preserved (the `reasoning` alias only wins when `reasoning_content` is empty), and the tests cover the alias, the precedence, the guardrail, and the exec stream-json ordering. Clean.

On the policy blocker: #439 is a real bug, so I've marked it `issue-approved`. The label didn't actually exist in the repo yet even though CONTRIBUTING.md references it — created it, so the gate is now satisfiable for the rest of the queue too.

kevincodex1 (approved):

PR #487, opened 2026-07-04
fix(acp): warn on session persistence failures

### LINKED ISSUE #471: acp: silent errors in session persistence can lose history

Ashwinhegde19:
Fixes #471

## Summary
- surface ACP session history load failures as non-fatal warning thought chunks
- surface turn persistence append failures as non-fatal warning thought chunks
- keep prompt turns and in-memory continuity working when persistence fails

## Tests
- `go test ./internal/acp -count=1`
- `make lint`
- `make test-quick`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session history loading to proceed when history can’t be read, and emit a warning while still completing the prompt.
  * Improved turn persistence so partial save failures produce a warning rather than being silently ignored.
  * Ensured warning/thought messages are emitted consistently during streaming.
* **Tests**
  * Added end-to-end coverage for persistence failures and corrupted session history, verifying that warnings are shown and the prompt still completes.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"sync"
 
@@ -159,8 +160,11 @@ func (a *Agent) handleSessionLoad(_ context.Context, params json.RawMessage) (an
 	// Load history BEFORE publishing the session so no concurrent prompt observes
 	// a half-initialized session (registerSession sets history under the lock and
 	// reuses an already-live session rather than orphaning its in-flight turn).
-	history := a.loadHistory(meta.SessionID)
+	history, historyErr := a.loadHistory(meta.SessionID)
 	sess := a.registerSession(meta.SessionID, root, history)
+	if historyErr != nil {
+		(&notifier{conn: a.conn, sessionID: sess.id}).warning("Warning: ACP session history could not be loaded: " + historyErr.Error())
+	}
 	return LoadSessionResult{
 		Modes: a.modeState(sess),
 	}, nil
@@ -253,7 +257,9 @@ func (a *Agent) runTurn(ctx context.Context, sess *acpSession, userText string, 
 	if stopErr != nil {
 		return "", RPCError(codeInternalError, stopErr.Error())
 	}
-	a.persistTurn(sess, userText, result.FinalAnswer)
+	if err := a.persistTurn(sess, userText, result.FinalAnswer); err != nil {
+		note.warning("Warning: ACP session history was not fully persisted: " + err.Error())
+	}
 	return reason, nil
 }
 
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (approved):
Clean and well-scoped. ACP persistence errors that were getting swallowed now surface as non-fatal warning chunks, and the in-memory history is kept unconditionally so the turn keeps working when the store fails. Traces correct, the tests cover both the load and the append failure paths, and there's nothing platform-specific. Approving.

jatmn (approved):
LGTM

internal/acp/agent.go
kevincodex1, line 87: can you check this? its flagged by CI security as unreachable function

kevincodex1 (changes requested):
please check agent.go changes there is un unreachable function

jatmn (changes requested):
## Note

GitHub currently reports `mergeStateStatus: BLOCKED` even though the branch is mergeable from a conflict standpoint. The apparent blocker is the open requested-changes review about `internal/acp/agent.go:87` being flagged as unreachable. I checked that line: `Agent.Serve` is unchanged from `main`, is a thin wrapper around `Conn.Serve`, and is used by the ACP test harness. The repo's CI security workflow documents the deadcode step as advisory/non-blocking, and the current Security & code health job passed. I do not see a code or security fix needed for that specific flag, but the outstanding review state still needs to be resolved before the PR can proceed.

## Findings

- [P3] Strengthen the persistence-failure test to verify in-memory continuity
  `internal/acp/agent_test.go:163`
  `TestACPPromptWarnsWhenPersistenceFails` correctly checks that the prompt succeeds and that a warning thought chunk is emitted, but it does not assert that the assistant's message text still reaches `h.updates` when persistence fails. Since the PR's stated goal includes keeping prompt turns and in-memory continuity working when persistence fails, please add an assertion such as `drainContains(t, h.updates, "Hello from ZERO")` so the test proves the warning path does not interrupt streamed assistant output.

Vasanthdev2004 (approved):
Approving.

The persistence-failure test now asserts in-memory continuity the way you asked — `drainContains(h.updates, "Hello from ZERO")` proves the assistant's output still streams when the append fails, alongside the two warning-thought assertions for the load and append paths. That was the one open finding.

The deadcode flag on `agent.go:87` is a non-issue — `Serve` is unchanged from main and that step is advisory, as you already found; nothing to fix there.

Marked the linked issue #471 `issue-approved` — silent persistence errors dropping session history is a real bug, and surfacing them as non-fatal warnings while keeping the turn and in-memory continuity alive is the right fix. CI is green.

Vasanthdev2004:
Thanks for this — the session-persistence warning is a real gap and your fix addresses it cleanly.

Closing as superseded: the same behavior landed on main via #535 (commit f951aee6, "fix(acp): surface session persistence failures"), so this branch is now conflicting and redundant. Your contribution isn't wasted — it called out and fixed the issue first. If #535 missed any case you covered here, please open a follow-up issue so we can track it rather than reopening this branch.

PR #491, opened 2026-07-04
fix(specialist): cap max specialist nesting depth

euxaristia:
## Summary

Adds a hard cap on specialist (Task tool) nesting depth, as a defense-in-depth safeguard.

Related to #470. Not closing it: as explained below, today's normal specialist/swarm path can't actually reach this recursively, so this cap is a future-proof guard rather than a fix for an active exploit of the recursive chain #470 describes.

## Changes

- Add `maxSpecialistDepth = 8` in `internal/specialist/exec.go`.
- `Executor.Run` now rejects a call whose `CurrentDepth` is already at or past the cap with a clear error, before any specialist work starts. (Fixed from an earlier off-by-one: the check must trigger at `CurrentDepth >= maxSpecialistDepth`, not `>`, since this call is what launches the child at `CurrentDepth+1` — checking only `>` let a parent already at the cap start one more level before the guard caught it.)

## Why this is defense-in-depth, not an active-exploit fix

Per jatmn's review: today, a specialist (or swarm member) can't actually reach this code path recursively. Every specialist/swarm child is launched with `--tag specialist` (`internal/specialist/exec.go`, the only place `--tag` is set on a spawned child), and `shouldRegisterExecSpecialistTools` (`internal/cli/app.go`) refuses to register the Task tool whenever `tag == "specialist"`, regardless of `--auto high` — already covered by the existing `exec_test.go` case asserting a specialist child does not get the Task tool. So a specialist cannot itself call Task to spawn a grandchild, and `CurrentDepth` can't structurally exceed 1 through today's normal flow.

The cap is still worth having: it's cheap, matches the issue's suggested fix (hard `MaxDepth` with a clear error), and protects against a future change to that tag-gating logic silently reopening the recursive path. It is not closing an actively exploitable bug today.

## Test plan

- `TestRunRejectsDepthExceedingMax` and `TestRunRejectsDepthAtMax` in `internal/specialist/exec_test.go`, covering both the already-past-cap case and the exact-boundary case the off-by-one fix addresses. `TestRunRejectsDepthAtMax` is now a table covering both a fresh call and a `Resume` call, since the guard sits before the fresh/resume split in `Run`.
- `go build ./...`
- `go vet ./internal/specialist/...`
- `go test ./internal/specialist/... -count=1` (all pass)

This was split out of #468 at reviewer request (Vasanthdev2004 asked for the bundled fixes to be split into independent PRs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a maximum cap on nested specialist execution depth to prevent overly deep task chains.
  * Runs with a current depth at or beyond the limit now fail fast with a clear, depth-related error.
* **Tests**
  * Extended coverage to verify rejection when depth exceeds the maximum.
  * Added boundary tests for both fresh runs and resume runs when depth equals the maximum.

### DIFF
--- a/internal/specialist/exec.go
+++ b/internal/specialist/exec.go
@@ -219,8 +219,12 @@ func (executor Executor) Run(ctx context.Context, params TaskParameters, options
 	if options.CurrentDepth < 0 {
 		return ExecResult{}, fmt.Errorf("current depth cannot be negative")
 	}
-	if options.CurrentDepth > maxSpecialistDepth {
-		return ExecResult{}, fmt.Errorf("specialist nesting depth %d exceeds maximum %d", options.CurrentDepth, maxSpecialistDepth)
+	// >= (not >): this Run call spawns a CHILD at options.CurrentDepth+1, so a
+	// parent already AT the cap must still be rejected here rather than being
+	// allowed to launch one more level before the child's own next call trips
+	// the guard.
+	if options.CurrentDepth >= maxSpecialistDepth {
+		return ExecResult{}, fmt.Errorf("spawning a specialist at depth %d would exceed maximum nesting depth %d", options.CurrentDepth+1, maxSpecialistDepth)
 	}
 	if strings.TrimSpace(params.Prompt) == "" {
 		return ExecResult{}, fmt.Errorf("specialist prompt is required")
--- a/internal/specialist/exec.go
+++ b/internal/specialist/exec.go
@@ -25,6 +25,7 @@ import (
 const (
 	sessionTagSpecialist     = "specialist"
 	promptFileThresholdBytes = 4 * 1024
+	maxSpecialistDepth       = 8 // hard cap to prevent infinite recursion/resource exhaustion via Task
 )
 
 const SessionTagSpecialist = sessionTagSpecialist
@@ -217,6 +218,9 @@ func (executor Executor) Run(ctx context.Context, params TaskParameters, options
 	}
 	if options.CurrentDepth < 0 {
 		return ExecResult{}, fmt.Errorf("current depth cannot be negative")
+	}
+	if options.CurrentDepth > maxSpecialistDepth {
+		return ExecResult{}, fmt.Errorf("specialist nesting depth %d exceeds maximum %d", options.CurrentDepth, maxSpecialistDepth)
 	}
 	if strings.TrimSpace(params.Prompt) == "" {
 		return ExecResult{}, fmt.Errorf("specialist prompt is required")
--- a/internal/specialist/exec_test.go
+++ b/internal/specialist/exec_test.go
@@ -349,13 +349,24 @@ func TestRunRejectsDepthExceedingMax(t *testing.T) {
 // TestRunRejectsDepthAtMax covers the boundary: a parent already AT the cap
 // must be rejected too, since this Run call would launch a child one level
 // past it (--depth CurrentDepth+1). Only checking ">" here would let that
-// child start before the guard ever caught it.
+// child start before the guard ever caught it. The guard sits before the
+// fresh/resume branch in Run, so both call shapes must be proven to reject
+// here rather than falling through to runFresh/runResume.
 func TestRunRejectsDepthAtMax(t *testing.T) {
-	_, err := (Executor{}).Run(context.Background(), TaskParameters{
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (approved):
Clean little guard, and it's the specialist-depth split from #468 done exactly as asked. Cap's at 8, the check fires before any specialist work starts, and the comparison is right — it rejects only past the cap, so legit nesting up to it still works. And the cap is actually reachable: `CurrentDepth` propagates through `--depth`, so a runaway chain does hit it rather than this being dead code. Test covers it, CodeRabbit's clean. Closes #470. Approving.

jatmn (changes requested):

## Findings

- [P2] Stop launching the over-cap specialist child
  `internal/specialist/exec.go:222`
  The new guard rejects only when the parent run's `CurrentDepth` is already greater than `maxSpecialistDepth`, but the child process is launched with `--depth CurrentDepth+1` a few lines later. A specialist currently at depth 8 therefore still passes this check and starts another specialist session at depth 9, so work begins beyond the advertised maximum before the next Task call is rejected. Please reject when the next child would exceed the cap, and cover that boundary for both fresh and resumed specialist runs.

- [P2] Demonstrate the recursive Task path before closing #470
  `internal/cli/app.go:984`
  The PR body says this closes #470 because "a chain of specialists recursively spawning further specialists via `--depth`" can recurse indefinitely, but the normal child path appears to suppress that chain: `BuildArgs`/`BuildResumeArgs` always launch the child with `--tag specialist`, and `shouldRegisterExecSpecialistTools` returns false for that tag, with an existing `--list-tools` test asserting that a specialist child does not get the Task tool even under `--auto high`. Please either add evidence/tests for the reachable recursive Task path this cap is meant to stop, or update the PR/issue-closing claim so #470 is not closed by a guard that does not cover the reported behavior.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one issue that still needs to be addressed.

## Findings

- [P2] Do not auto-close #470 from this defense-in-depth guard
  `PR description`
  The code now fixes the off-by-one by rejecting `CurrentDepth >= maxSpecialistDepth`, so the first review item is addressed. The PR body also now agrees with the second review item that today's normal specialist/swarm child path cannot recursively reach Task because child sessions are launched with `--tag specialist` and `shouldRegisterExecSpecialistTools` suppresses Task for that tag. However, the body still says `Closes #470`, which would close the issue describing an actually recursive Task chain even though the PR now says this is only a future-proof guard and not an active-exploit fix. Please complete the prior review request by removing the closing keyword from this PR, or add concrete evidence/tests for a currently reachable recursive path that this cap actually stops.

- [P3] Cover the resume boundary that was requested
  `internal/specialist/exec_test.go:348`
  The previous review asked for the over-cap boundary to be covered for both fresh and resumed specialist runs. `TestRunRejectsDepthAtMax` does verify the shared pre-branch guard today, but it only passes a plain prompt and never exercises the `Resume` parameters, so a future refactor could move the guard into the fresh path and still leave the requested resume boundary uncovered. Please add the explicit resume-at-cap case as well, or otherwise make the test table prove both fresh and resume calls reject before launching a child.

jatmn (changes requested):
Thanks for the update. I rechecked the previously discussed paths and found one issue that still needs to be addressed.

## Findings

- [P2] Remove the closing keyword from the commit history too
  `548cf478`
  The PR body now correctly says this is only related to #470 and is not closing it, but the first commit in this PR still contains `Closes #470`. That leaves the same issue-closing risk in the branch metadata: if this is merged in a way that preserves the original commit message, #470 can still be closed even though the current PR description says this defense-in-depth guard is not an active fix for the recursive Task path. Please complete the prior review request by rewriting/removing that closing keyword from the PR's commits, or add concrete evidence/tests for a currently reachable recursive path that this cap actually stops.

euxaristia:
Fixed: the closing keyword is gone from the commit history. The first commit was amended (548cf47 -> 498afca) so `Closes #470` reads `Related to #470, not closing it`, matching the PR description. No code changed, just the message. Force-pushed to c89032f.

@jatmn ready for re-review.

jatmn (approved):

@Vasanthdev2004

kevincodex1 (approved):
LGTM

PR #492, opened 2026-07-04
fix(tools): cap bash stdout/stderr capture to prevent OOM

### LINKED ISSUE #469: bash tool: unbounded stdout/stderr buffering can OOM the agent

euxaristia:
## Summary

The `bash` tool captured command stdout/stderr into an unbounded `bytes.Buffer` with no size limit. Large output (cat on a huge file, a `gh pr` with thousands of comments, `find /`, a verbose test run, etc.) can OOM the agent process, with no backpressure, and the process can be killed before the command even finishes.

Closes #469

## Changes

- Add `maxBashOutputBytes` (16 MiB) and a `cappedWriter` type in `internal/tools/bash.go`. `cappedWriter` wraps a `bytes.Buffer`, stops accepting bytes once the cap is hit, and sets a truncated flag instead of growing further.
- `bashTool.run` now uses `cappedWriter` for both stdout and stderr, and surfaces a `[zero] output truncated at 16 MiB` / `[zero] stderr truncated at 16 MiB` notice in the tool output when the cap is hit.
- `cappedWriter` intentionally does not embed `*bytes.Buffer`. `exec.Cmd` drives non-`*os.File` Stdout/Stderr writers through `io.Copy`, which prefers `io.ReaderFrom` over `Write` when the destination implements it. Embedding `*bytes.Buffer` would promote its `ReadFrom` method and let `io.Copy` bypass the cap entirely, streaming unbounded output straight into the buffer. Keeping the buffer as a named (non-embedded) field avoids that trap.

## Test plan

- Added `TestBashToolCapsLargeStdoutAndReportsTruncation`: runs a real subprocess writing 20 MiB of stdout through the bash tool and confirms the output is bounded and carries the truncation notice.
- Added `TestCappedWriterStopsGrowingAfterMax` and `TestCappedWriterCapturesPartialWriteThatCrossesMax`: direct unit tests of the write-capping logic.
- Added `TestCappedWriterEnforcesCapThroughIOCopy`: a regression test that drives `cappedWriter` through `io.Copy` (the same path `exec.Cmd` uses) to lock in the `io.ReaderFrom` fix described above.
- `go build ./internal/tools/...`
- `go vet ./internal/tools/...`
- `go test ./internal/tools/... -count=1`: all relevant tests pass. Three pre-existing, location-sensitive failures (`TestScopedToolsAllowReadOnlyRootsWithoutWrite`, `TestRegistryAppliesSandboxBeforeToolExecution`, `TestRegistrySandboxGatesPathAliasKeys`) reproduce identically on unmodified upstream/main code in this same checkout path and are unrelated to this change.

This was split out of #468 at reviewer request (Vasanthdev2004 asked for the bundled fixes to be split into independent PRs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Large command stdout/stderr is now safely capped to prevent excessive memory usage.
  * When output exceeds the limit, it now includes clear truncation notices (including the configured MiB amount).
  * Bash command results continue to format normally while handling oversized output more reliably.
* **Tests**
  * Expanded coverage for stdout truncation behavior and boundary conditions to prevent regressions.

### DIFF
--- a/internal/tools/bash.go
+++ b/internal/tools/bash.go
@@ -380,10 +380,10 @@ func formatBashOutputWithTruncation(stdout string, stderr string, exitCode int, 
 		parts = append(parts, fmt.Sprintf("[zero] redacted %d likely secret(s) from this output before showing it.", n))
 	}
 	if stdoutTruncated {
-		parts = append(parts, "[zero] output truncated at 16 MiB")
+		parts = append(parts, fmt.Sprintf("[zero] output truncated at %d MiB", maxBashOutputBytes/(1024*1024)))
 	}
 	if stderrTruncated {
-		parts = append(parts, "[zero] stderr truncated at 16 MiB")
+		parts = append(parts, fmt.Sprintf("[zero] stderr truncated at %d MiB", maxBashOutputBytes/(1024*1024)))
 	}
 	if len(parts) == 0 {
 		return "Command completed with no output."
--- a/internal/tools/bash.go
+++ b/internal/tools/bash.go
@@ -18,6 +18,40 @@ import (
 const defaultBashTimeoutMS = 120000
 const maxBashTimeoutMS = 600000
 
+const maxBashOutputBytes = 16 * 1024 * 1024 // 16 MiB - prevent OOM on large command output
+
+// cappedWriter wraps a bytes.Buffer and stops accepting bytes after max,
+// setting a flag instead. Used to cap stdout/stderr capture during
+// command.Run() instead of only truncating afterward.
+//
+// The buffer is a named field rather than embedded on purpose: embedding
+// *bytes.Buffer would promote its ReadFrom method, and exec.Cmd's Run()
+// drives non-*os.File Stdout/Stderr writers through io.Copy, which prefers
+// io.ReaderFrom over Write when the destination implements it. That would
+// let the child process's output flow straight into the buffer via the
+// promoted ReadFrom, bypassing the cap entirely and defeating the whole
+// point of this type.
+type cappedWriter struct {
+	buf       *bytes.Buffer
+	max       int
+	truncated *bool
+}
+
+func (c *cappedWriter) Write(p []byte) (int, error) {
+	if c.buf.Len()+len(p) > c.max {
+		*c.truncated = true
+		remain := c.max - c.buf.Len()
+		if remain > 0 {
+			c.buf.Write(p[:remain])
+		}
+		return len(p), nil
+	}
+	return c.buf.Write(p)
+}
+
+func (c *cappedWriter) String() string { return c.buf.String() }
+func (c *cappedWriter) Len() int       { return c.buf.Len() }
+
 type bashTool struct {
 	baseTool
 	workspaceRoot string
@@ -124,10 +158,12 @@ func (tool bashTool) run(ctx context.Context, args map[string]any, engine *zeroS
 	defer plan.Cleanup()
 	addSandboxMeta(meta, plan)
 
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-	command.Stdout = &stdout
-	command.Stderr = &stderr
+	stdoutTruncated := false
+	stderrTruncated := false
+	stdout := &cappedWriter{buf: &bytes.Buffer{}, max: maxBashOutputBytes, truncated: &stdoutTruncated}
+	stderr := &cappedWriter{buf: &bytes.Buffer{}, max: maxBashOutputBytes, truncated: &stderrTruncated}
+	command.Stdout = stdout
+	command.Stderr = stderr
 
 	// Kill the shell as a process group on timeout and bound the post-kill I/O
 	// wait, so a backgrounded child cannot outlive the command or hang Run().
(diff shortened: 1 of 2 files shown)


jatmn (approved):
Thanks for the contribution. I do not see any actionable issues from my review.

@Vasanthdev2004

Vasanthdev2004 (changes requested):
Thanks for splitting this out — it's exactly the focused PR I asked for on #468, and the code is clean. The cap holds on every write path, returning `len(p), nil` so the child keeps draining instead of taking a SIGPIPE is the right call, there's no race between the two streams, and the 20 MiB subprocess test actually exercises it. The trick of not embedding `*bytes.Buffer` so `io.Copy`'s `ReaderFrom` path can't bypass the cap is sharp, and the regression test locking it in is the right instinct.

The snag isn't the code — it's that #484 ("polish v2", item #3) also caps bash stdout/stderr and rewrites the same lines of `bash.go`, so the two collide and can't both land as-is. And they're really fixing two different halves:

- **#484** caps what reaches the model — 96 KB head+tail per stream, keeps the tail where build/test errors live, sets `Result.Truncated`, and plugs into the same output-budget infra the read/search tools use. But it truncates *after* buffering into an unbounded buffer, so it does nothing for memory: `cat 10GB.log` still OOMs before it ever truncates.
- **This PR** caps the capture itself, which is the only thing that actually prevents the OOM in #469 — but 16 MiB head-only is a non-guard for context (≈4M tokens, past any window) and drops the tail where the errors are.

So they're complementary layers, not competitors, and the clean end state is both. What I'd like: keep the part that's genuinely unique here — the `cappedWriter` capture guard and its `ReaderFrom` regression test — but reshape it into a pure OOM backstop *under* #484's budget. Concretely: drop the cap from 16 MiB to a few MiB (a memory-only ceiling, high enough it basically never trips in normal use), delete the standalone `[zero] output truncated at 16 MiB` notice and the truncated-bool threading, feed the capped buffer into #484's `budgetBashOutput`, and let a memory-cap trip also flip `Result.Truncated`. Then capture is OOM-safe and the model-facing truncation is #484's head+tail. Easiest to rebase onto #484 once its shape settles.

One non-blocking note on the rationale (code's fine, just the explanation): on the current Go toolchain `*os.File` does implement `io.WriterTo` (since 1.22), so "the pipe source has no WriteTo" isn't quite the mechanism — the `ReaderFrom` branch is reached one `io.Copy` layer deeper (`os.File.WriteTo` → `genericWriteTo`, which hides `WriteTo`). Outcome's identical and your non-embedded-field call is still exactly right; just don't document the reason as "*os.File lacks WriteTo."

To be clear, the overlap is on me for approving #484's bash budget while you were splitting #468 — not a knock on your work. If you'd rather not wait on #484, tell me and I'll help sort the sequencing.

euxaristia:
Thanks for the detailed breakdown. Agreed these are complementary layers, and reshaping this into a pure OOM backstop under #484's budget makes sense.

I'd rather not restructure this against a moving target, so I'll wait for #484 to settle (or merge) before reshaping the cap and dropping the standalone truncation notice/threading here. Will follow up once that lands.

Vasanthdev2004:
Circling back now that the sequencing resolved — closing this as superseded by #484.

When I last reviewed, this PR's unique value was the `cappedWriter` capture guard: the real OOM backstop for #469, which #484's budget didn't cover because it truncated *after* buffering. That gap is closed now — #484's latest commit ("bound bash output capture in memory, not just model-visible text") caps the capture itself, so it does both halves: the 96 KB head+tail model-facing cap and the memory ceiling that prevents the OOM. I re-checked that path on #484 and the capture is genuinely bounded (a few hundred KB per stream), so #469 gets fixed there.

That leaves this one a conflicting duplicate — it collides with #484 on the same `bash.go` lines (showing `CONFLICTING`), and the only commit since my review was the notice-string cleanup rather than the reshape-onto-#484 I'd suggested. No reason to have you rebase a duplicate onto an approved PR, so I'm closing it.

Not a knock on the work — it was clean, the `ReaderFrom`-bypass guard was the sharp bit, and it's a good part of why #484 ended up hardening the in-memory bound. #469 lands either way once #484 merges. If #484 stalls for any reason, reopening this is one click. Thanks for splitting it out when I asked.

PR #496, opened 2026-07-04
Cache TUI model registry

### LINKED ISSUE #494: TUI render rebuilds model registry every frame
### Problem

TUI render path calls modelregistry.DefaultRegistry() through modelContextWindow(). DefaultRegistry() rebuilds catalog, clones entries, and compiles regexes. This happens from title/status/sidebar render paths.

### Requested behavior

Cache curated model registry once per TUI model. Reuse it for context-window and vision checks.

### Why it belongs in Zero

Own code says DefaultRegistry() “MUST NOT be called from render path” because registry is rebuilt every call. Current render path violates that and adds avoidable alloc/CPU work during interactive redraws.

Could use package-level sync.OnceValue, but per-TUI-model cache is smaller and keeps tests/injection local.

Measured lookup cost:
current: ~29.9µs, 67KB, 595 allocs/op
cached: ~456ns, 160B, 3 allocs/op

GautamBytes:
Fixes #494
## Summary
- cache the curated model registry once per TUI model\n- reuse it for context-window and vision checks
- add regression tests plus a lookup benchmark.

## Summary by CodeRabbit

* **Performance**
  * Reduced repeated model-registry lookups by reusing a cached model catalog during model checks.

* **Bug Fixes**
  * Improved how supported context windows and vision capability are determined for models.
  * Kept discovered-model handling and name-based fallback behavior intact when needed.

* **Tests**
  * Added coverage for cached catalog lookups, vision-capability checks, and discovery interactions.
  * Added a benchmark comparing repeated registry loading versus cached lookup performance.

### DIFF
--- a/internal/tui/image_attach.go
+++ b/internal/tui/image_attach.go
@@ -98,16 +98,9 @@ func (m model) modelSupportsVisionTUI() bool {
 	if trimmed == "" {
 		return false
 	}
-	// Check the curated registry first.
-	registry, err := modelregistry.DefaultRegistry()
-	if err == nil {
-		if modelregistry.SupportsVision(registry, trimmed) {
-			return true
-		}
-		// Catalog knows this model and it lacks vision — trust that.
-		if _, known := registry.Get(trimmed); known {
-			return false
-		}
+	// The curated catalog is authoritative only when it knows the model.
+	if entry, known := m.modelCatalog.Resolve(trimmed); known {
+		return entry.Supports(modelregistry.ModelCapabilityVision)
 	}
 	// Check the discovered model list (from models.dev) for InputModalities
 	// containing "image". This covers custom/ollama/cloud models not in the
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -67,6 +67,7 @@ type model struct {
 	gitBranch                   string
 	providerName                string
 	modelName                   string
+	modelCatalog                modelregistry.Registry
 	providerProfile             config.ProviderProfile
 	savedProviders              []config.ProviderProfile
 	provider                    zeroruntime.Provider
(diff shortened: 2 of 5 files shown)


jatmn (approved):
@Vasanthdev2004 

I found one process blocker that needs a maintainer decision before this is ready.


## Findings

- [P2] Resolve the approved-issue requirement for this community PR
  `CONTRIBUTING.md:19`
  This PR links `#494`, but the linked issue currently only has the `enhancement` label, while the contribution policy says community PRs must be tied to an existing issue that has already been reviewed and marked with `issue-approved`. Please get the linked issue marked as approved, or have a maintainer explicitly confirm that this PR falls under the team-member/internal-work exception, before proceeding with the code changes.

Vasanthdev2004 (approved):
Good fix. `modelContextWindow` runs a couple of times per render frame (header, context gauge, sidebar), and on main each call rebuilt the whole registry — validate 13 entries, compile the regexes, deep-clone everything — which the codebase itself warns against (`session_controls.go:156`: "MUST NOT be called from the render path"). Caching the catalog once in `newModel` and reading it on lookup is the right minimal fix, and reusing that same registry for the usage tracker instead of building a second one is a nice touch.

I checked the thing that worried me — staleness after a `/model` switch — and it's fine: the cache is the whole immutable catalog, not a memoized per-model answer, so `Resolve(m.modelName)` still resolves the new model correctly. Discovered/Ollama models stay in their own live maps, untouched. No race either — built eagerly in `newModel`, only ever read after, and the registry methods clone on read.

Bonus: the vision reorder actually fixes a latent bug on main where the name heuristic could short-circuit `true` before the discovered list was consulted (a discovered text-only `gpt-5-*` would wrongly report vision support). There's a regression test for it.

CodeRabbit's one real catch — the registry error getting swallowed when a `UsageTracker` is injected — is handled at HEAD; the `DefaultRegistry()` error now panics unconditionally before the tracker branch, so the cached registry can't be a zero value.

Only non-code thing: #494 has `enhancement` but not `issue-approved` — your call. Approving.

kevincodex1 (approved):
LGTM

PR #497, opened 2026-07-04
fix(config): let a gateway ANTHROPIC_BASE_URL resolve as anthropic-compatible

### LINKED ISSUE #479: unable to setup due to hard requirement on anthropic url
### Version / branch / commit

main

### OS and environment

Windows 11

### Steps to reproduce

unable to setup zero as it keeps giving me this warning: [zero] anthropic provider anthropic requires official baseURL https://api.anthropic.com (apiKey=[REDACTED])
I setup both keys but used my local gateway which I am developing and looks like it's looking for the actual api.anthropic.com url. not a great feature IMO.

### Expected behavior

it should allow me to setup proxy/gateway urls

### Actual behavior

keeps asking me to set this up: [zero] anthropic provider anthropic requires official baseURL https://api.anthropic.com (apiKey=[REDACTED])

_No response_

Vasanthdev2004:
Fixes #479.

Setting `ANTHROPIC_BASE_URL` to a local gateway/proxy blows up setup with `anthropic provider anthropic requires official baseURL https://api.anthropic.com`. Turns out `OPENAI_BASE_URL` already handles this — a non-official URL from the env auto-promotes to the `openai-compatible` kind — but Anthropic never got the same treatment, so the standard `ANTHROPIC_BASE_URL` proxy convention just gets rejected.

This mirrors the OpenAI behavior for Anthropic: a non-official `ANTHROPIC_BASE_URL` in the environment resolves to `anthropic-compatible` instead of the pinned first-party kind.

Kept it env-only on purpose. A custom baseURL in a project's `config.json` is still rejected — that pin is a security control (a cloned repo shouldn't be able to point your real Anthropic key at another host), and `TestResolveRejectsAnthropicOfficialProviderCustomBaseURL` still passes to prove it. The environment is the user's own shell, so a gateway URL there is a deliberate choice.

Verified with a resolver test that runs the exact reported env through `Resolve()` and checks it lands on `anthropic-compatible` with the gateway URL and no error.

One gap left out on purpose: `GEMINI_BASE_URL`/`GOOGLE_BASE_URL` has the same limitation, but there's no `google-compatible` kind to promote to, so fixing that is a separate, larger change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment-based configuration so Anthropic-compatible gateways and proxies are recognized correctly when a custom base URL is provided.
  * Resolved provider selection to keep the configured base URL and use the expected Anthropic credentials and model settings in these scenarios.

### DIFF
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -590,8 +590,16 @@ func applyProviderEnv(cfg *FileConfig, providerKind ProviderKind, env envProfile
 	if profile.Name == "" {
 		profile.Name = string(providerKind)
 	}
+	// A non-official baseURL from the environment signals a proxy/gateway, so promote
+	// the first-party kind to its -compatible transport (which accepts a custom URL).
+	// Env-only by design: a custom baseURL in a project config.json is left as-is and
+	// rejected by Resolve, so an untrusted project can't redirect a first-party key to
+	// another host.
 	if providerKind == ProviderKindOpenAI && baseURL != "" && !isOfficialOpenAIBaseURL(baseURL) {
 		profile.ProviderKind = ProviderKindOpenAICompatible
+	}
+	if providerKind == ProviderKindAnthropic && baseURL != "" && !isOfficialAnthropicBaseURL(baseURL) {
+		profile.ProviderKind = ProviderKindAnthropicCompat
 	}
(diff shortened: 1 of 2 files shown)


jatmn (approved):
LGTM

kevincodex1 (approved):
LGTM

PR #498, opened 2026-07-04
Fix/detailed view scroll select

### LINKED ISSUE #495: detailed transcript view is just a dump of text
### Version / branch / commit

zero 0.1.0

### OS and environment

Nixos v26.05, kitty, go v1.26

### Steps to reproduce

1. Run `zero`
2. make the agent do any task or long task
3. click `ctrl + o` it will open the detailed transcript view mode
4. it expects the terminal height is as same as how long the session text is, which is wrong and bad
5. you cant scroll, select to copy, and the footer of "ESC close | Ctrl + O toggle"  isnt fixed in bottom like the header (better ui and ux)

### Expected behavior

the expected behavior is intentional and known from the steps to reproduce part, but i will tell it again

the detailed transcript mode should have the same abilities of the normal transcript mode. it should have the zero native select to copy, it should have the ability to scroll with pgDown/pgUp, arrowDown/arrowUp, touchpad or mouse wheel.

and it should be fine and instantly scrolled within a running task.

### Actual behavior

the actual behavior:
doesnt support scrolling with anyway
doesnt support the native way of select to copy of the agent

_No response_

sam0uly:
## Linked issue

Fixes #495

https://github.com/user-attachments/assets/43d16dcb-524a-4f8d-a966-ef1099d59fe3

## Checklist

- [x] `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Detailed transcript mode now supports PgUp/PgDown and Up/Down scrolling within the view, including consistent small-step behavior.
  * Toggling detailed transcript now exits an active file view.
* **Bug Fixes**
  * Clipboard/plain-text exports now fully strip ANSI/control sequences, including highlighted code blocks.
  * Improved detailed transcript layout and scrolling visibility.
* **UI Changes**
  * Context sidebar is hidden during detailed transcript mode.
  * Ctrl+B sidebar toggle and composer mouse selection/hover are blocked in detailed mode.
* **Tests**
  * Added coverage for ANSI stripping, detailed transcript scrolling, and composer mouse blocking in detailed mode.


Vasanthdev2004 (changes requested):
The approach is right — routing the detailed view through the same selectable-line + viewport pipeline as the normal view is exactly what the issue described, and it fixes both halves (scroll keys and click-drag selection). I checked for a normal-view regression and the render/flush path holds up. Good to have that view not be a dead text dump anymore.

A few loose ends before it goes in, though — most of these CodeRabbit already flagged and they're still open at HEAD:

- The Up-arrow branch in the scroll handler doesn't clear hover state the way PgUp/PgDn/Down now do — looks like 3 of the 4 got it and Up got missed.
- `transcriptRowBodyHeightCacheKey` is dead after the refactor, nothing calls it. Drop it.
- The footer still hardcodes `Ctrl+O` for the detailed toggle. Now that keybindings are rebindable (#417), that should read from `m.keyBindings.toggleDetailed` so a remapped key shows the right label.
- One behavior change worth confirming is intended: the rewritten default render case now returns selectable lines instead of `nil`, and that default is also hit by `rowRecap`, not just `rowWelcome` — and `rc.skip()` doesn't skip recap rows. So recap rows pick up selection/hover geometry in the *normal* view too. Probably harmless (maybe even nice), but it wasn't called out, so confirm it's on purpose.

None of this is a crash or a correctness problem — the fix works, CI's green — it's just not quite buttoned up. Tie those off and it's an easy approve.

sam0uly:
done and yes, recap rows getting selection/hover geometry is intentional. theyre user facing content and there's no reason to exclude them from copy/selection

jatmn (changes requested):
I found a couple of issues that should be addressed before this is buttoned up.

## Findings

- [P2] Do not let Ctrl+B toggle the hidden sidebar while detailed mode is active
  `internal/tui/model.go:1310`
  This PR hides the context sidebar while `m.transcriptDetailed` is true, but the `Ctrl+B` handler still runs before the later detailed-mode key swallow and `sidebarToggleAllowed()` does not check `m.transcriptDetailed`. As a result, pressing `Ctrl+B` in the detailed transcript flips `m.sidebarHidden` with no visible feedback, and when the user exits detailed mode the normal transcript can unexpectedly come back without the sidebar. Please block the sidebar toggle while detailed mode is active, the same way `sidebarAvailable()` now suppresses rendering there.

- [P3] Complete the detailed-scroll down-arrow regression assertion
  `internal/tui/transcript_view_test.go:177`
  CodeRabbit's earlier arrow-key coverage request is still valid: the test only sends `KeyDown` while `chatScrollOffset` is already `0`, so it verifies the bottom clamp but never proves that `KeyDown` decrements from a nonzero offset back toward the bottom. Please extend `TestUpDownArrowsScrollDetailedTranscript` after the existing `KeyUp` assertion to send `KeyDown` again and assert the offset returns from `1` to `0`, so this new detailed-mode scroll path is actually covered.

sam0uly:
> I found a couple of issues that should be addressed before this is buttoned up.
> ## Findings
> 
>     * [P2] Do not let Ctrl+B toggle the hidden sidebar while detailed mode is active
>       `internal/tui/model.go:1310`
>       This PR hides the context sidebar while `m.transcriptDetailed` is true, but the `Ctrl+B` handler still runs before the later detailed-mode key swallow and `sidebarToggleAllowed()` does not check `m.transcriptDetailed`. As a result, pressing `Ctrl+B` in the detailed transcript flips `m.sidebarHidden` with no visible feedback, and when the user exits detailed mode the normal transcript can unexpectedly come back without the sidebar. Please block the sidebar toggle while detailed mode is active, the same way `sidebarAvailable()` now suppresses rendering there.
> 
>     * [P3] Complete the detailed-scroll down-arrow regression assertion
>       `internal/tui/transcript_view_test.go:177`
>       CodeRabbit's earlier arrow-key coverage request is still valid: the test only sends `KeyDown` while `chatScrollOffset` is already `0`, so it verifies the bottom clamp but never proves that `KeyDown` decrements from a nonzero offset back toward the bottom. Please extend `TestUpDownArrowsScrollDetailedTranscript` after the existing `KeyUp` assertion to send `KeyDown` again and assert the offset returns from `1` to `0`, so this new detailed-mode scroll path is actually covered.

done, there is what i did
1. the ctrl + b in detailed mode, added the guard to not be able to do it in detailed mode so it affect its state in the normal mode.

2. added a test coverage for keyDown after `keDown then keyUp` so it covers that.

jatmn (changes requested):

## Findings

- [P2] Let detailed transcript bypass the active file drill-in
  `internal/tui/transcript_selection.go:217`
  `Ctrl+O` can be pressed while a file drill-in is active, and `toggleDetailedTranscript` does not close or suspend that file view. The new detailed renderer calls `transcriptBodyItems(width, "", true)`, but this helper returns `m.fileViewBodyItems(width)` before looking at the detailed flag, so the "Transcript detailed" view can show the selected file body instead of the full transcript. That leaves the exact detailed transcript surface from #495 still unavailable from this state. Please make detailed mode ignore or exit the file drill-in before building its body items, and add a regression test for toggling detailed mode while `fileView.active` is true.

- [P2] Stop invisible composer hit-testing in detailed mode
  `internal/tui/composer.go:264`
  The rendered detailed transcript uses `detailedTranscriptFooter`, which has no composer, but the mouse path still checks `handleComposerSelectionMouse` before transcript hit-testing and `composerPositionAtMouse` builds its rectangle from the normal `footerView()`/`composerBox()`. The wheel path has the same problem through `mouseOverComposer`. In detailed mode those normal composer rows are actually part of the detailed transcript body, so click-drag selection or wheel scrolling over the lower body can be consumed as invisible composer interaction instead of selecting or scrolling the transcript. Please block composer mouse hit-testing while `m.transcriptDetailed` is true, or compute these rectangles from the same detailed frame that is actually rendered.

sam0uly:
> Thanks for the update. I rechecked the changed paths and found issues that still need to be addressed.
> ## Findings
> 
>     * [P2] Let detailed transcript bypass the active file drill-in
>       `internal/tui/transcript_selection.go:217`
>       `Ctrl+O` can be pressed while a file drill-in is active, and `toggleDetailedTranscript` does not close or suspend that file view. The new detailed renderer calls `transcriptBodyItems(width, "", true)`, but this helper returns `m.fileViewBodyItems(width)` before looking at the detailed flag, so the "Transcript detailed" view can show the selected file body instead of the full transcript. That leaves the exact detailed transcript surface from #495 still unavailable from this state. Please make detailed mode ignore or exit the file drill-in before building its body items, and add a regression test for toggling detailed mode while `fileView.active` is true.
> 
>     * [P2] Stop invisible composer hit-testing in detailed mode
>       `internal/tui/composer.go:264`
>       The rendered detailed transcript uses `detailedTranscriptFooter`, which has no composer, but the mouse path still checks `handleComposerSelectionMouse` before transcript hit-testing and `composerPositionAtMouse` builds its rectangle from the normal `footerView()`/`composerBox()`. The wheel path has the same problem through `mouseOverComposer`. In detailed mode those normal composer rows are actually part of the detailed transcript body, so click-drag selection or wheel scrolling over the lower body can be consumed as invisible composer interaction instead of selecting or scrolling the transcript. Please block composer mouse hit-testing while `m.transcriptDetailed` is true, or compute these rectangles from the same detailed frame that is actually rendered.

done

jatmn (approved):
LGTM 
@Vasanthdev2004

Vasanthdev2004 (approved):
Re-reviewing — my loose ends and jatmn's two findings are all tied off, so approving.

The rebindable footer label is done right: `labelOr(m.keyBindings.toggleDetailed, "Ctrl+O")` in both the footer and the hint, and the toggle itself now goes through `keyMatch` on the binding — so a remapped detailed key both shows and works. The Up-arrow hover clear and the dead cache-key helper are handled, and jatmn's two are in with tests: Ctrl+B no longer toggles the hidden sidebar in detailed mode (the handler gates on `!m.transcriptDetailed`), and entering detailed mode closes an active file drill-in.

Ran the tui suite locally on Windows (green), and CI is green across all platforms. The core is still the right fix — the detailed view goes through the same selectable-line + viewport pipeline as the normal view, so scroll keys and click-drag selection both work and it's no longer a dead text dump, which is exactly what #495 asked for.

Marked #495 `issue-approved` — the checklist claimed it already was, but it wasn't labeled yet. It's a real bug and this is the fix.

kevincodex1 (approved):
nice one! looks great thank you

PR #502, opened 2026-07-04
feat(tui): /loop — repeat a prompt or command on an interval or self-paced

### LINKED ISSUE #501: feat(tui): /loop — repeat a prompt or command on an interval or self-paced
## What

Add a `/loop` command that repeats a prompt or slash command — two modes:

- **Fixed interval** — `/loop 5m <prompt|/command>` re-fires on a wall-clock cadence.
- **Self-paced** — bare `/loop <prompt>` lets the model finish an iteration, pick its own next wake, and stop when the goal is met.

## Scope (v1, one PR)

- Foreground, fires while the session is idle between turns (rides the existing turn queue + post-turn re-submit seam; never interrupts a streaming turn).
- Session-scoped — the loop lives with its session and carries context; holds when the session isn't active.
- UX: `/loop`, `/loop list`, `/loop stop [id|all]`, a footer "N loops · next <time>" summary, and a `loop.md` playbook for bare `/loop` maintenance mode.
- Safety rails: max-iteration cap, auto-expire after N iterations/age, doom-loop repetition guard, and a leave-warning gate before `/quit` / `/clear` when loops are active.

## Out of scope (follow-ups)

- Cross-restart durability (a persisted `~/.zero/loops` store + missed-task catch-up).
- A detached background daemon that runs loops with no TUI attached.

Self-assigning; implementing on `feat/loop-command`.

Vasanthdev2004:
Adds `/loop` — repeat a prompt or slash command, in two modes. Fixes #501.

- **Fixed interval** — `/loop 5m /babysit-prs` re-fires every 5 minutes (also a trailing form: `/loop watch CI every 2m`). Because it runs off a Go timer rather than a cron backend, arbitrary intervals like `90s` or `7m` just work.
- **Self-paced** — bare `/loop keep tidying the docs` runs an iteration, then checks back on an adaptive cadence (2m early, widening to a ~30m heartbeat as it matures) and stops when there's no more progress.

Manage with `/loop list` and `/loop stop [id|all]`, and a persistent footer segment (`↻ N loops · next 3:05pm`) keeps a running loop visible.

### How it works
Loops are **foreground and session-scoped** — a loop only fires while the session is idle between turns (it never interrupts a streaming turn) and lives with its session. It rides machinery that's already there: an idle poll tick (~1s) fires the earliest due loop through the same seam `launchQueuedMessageIfReady` uses, so a user's queued prompt always wins and the two never collide. When an iteration finishes, the `agentResponseMsg` completion seam schedules the next wake or stops the loop.

### Safety
A loop can't run away: a max-iteration cap, a doom-loop guard (stops after 5 identical results in a row), and `/loop stop`. `/clear` notes that loops keep running (it wipes the transcript, not the session). Only a prompt or a custom `/command` may loop — not a built-in like `/model`.

### Tested
Unit tests for parsing/clamps/cadence, controller tests for the full lifecycle (start → advance → self-paced delay → done/doom/cap → stop → busy-skip), and end-to-end tests that drive `/loop` through `Update`. `go build ./...`, `go vet`, and the full `internal/tui` suite are green.

### Deliberately out of scope (follow-ups)
- Cross-restart durability (a persisted loop store + missed-run catch-up). Session-scoped is the right semantic for a context-carrying loop, so that's a separate call.
- A detached background daemon that runs loops with no TUI attached.
- An explicit `loop_control` tool for the model to set its own self-paced delay / done signal — the channel (`loopNextWake`/`loopDone`) is already wired; today self-paced uses the adaptive cadence + doom-guard.

## Summary by CodeRabbit

* **New Features**
  * Added a `/loop` command to repeat prompts or custom `/command`s on fixed intervals or in self-paced (control-line) mode, with loop listing/status and stop (by id or all).
  * Shows active loop details in the TUI status/footer with next wake timing.
* **Documentation**
  * Updated the README with `/loop` syntax and examples.
* **Bug Fixes**
  * Loops now run only while the session is idle and handle cancel/`/clear`/exit and session switch/resume safely, including adaptive pacing and safety limits.
* **Tests**
  * Added extensive automated coverage for parsing, scheduling, control handling, rendering, and confirmation/UI behaviors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ Common slash commands:
 | `/spec`, `/plan` | draft and review a plan before building |
 | `/image` | attach an image for vision-capable models |
 | `/resume`, `/rewind` | continue or roll back local sessions |
-| `/loop` | repeat a prompt or command on an interval (`/loop 5m /babysit-prs`) or self-paced |
+| `/loop` | repeat a prompt or custom `/command` on an interval (`/loop 5m /babysit-prs`) or self-paced |
 | `/compact`, `/context` | manage context usage |
 | `/permissions`, `/tools` | inspect available tools and policy |
 | `/add-dir` | allow an extra write directory for this session |
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ Common slash commands:
 | `/spec`, `/plan` | draft and review a plan before building |
 | `/image` | attach an image for vision-capable models |
 | `/resume`, `/rewind` | continue or roll back local sessions |
+| `/loop` | repeat a prompt or command on an interval (`/loop 5m /babysit-prs`) or self-paced |
 | `/compact`, `/context` | manage context usage |
 | `/permissions`, `/tools` | inspect available tools and policy |
 | `/add-dir` | allow an extra write directory for this session |
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -47,6 +47,7 @@ const (
 	commandExport
 	commandNew
 	commandSkills
+	commandLoop
 	commandUnknown
 )
 
@@ -296,6 +297,13 @@ var commandDefinitions = []commandDefinition{
 		group:       commandGroupSession,
 		description: "Write the conversation transcript to a file.",
 		kind:        commandExport,
+	},
+	{
+		name:        "/loop",
+		usage:       "/loop [interval] <prompt|/command> | /loop list | /loop stop [id|all]",
+		group:       commandGroupSession,
+		description: "Repeat a prompt or command on an interval (e.g. /loop 5m /babysit-prs), or self-paced when no interval is given.",
+		kind:        commandLoop,
 	},
 	{
 		name:        "/help",
--- a/internal/tui/loop.go
+++ b/internal/tui/loop.go
@@ -44,6 +44,12 @@ const (
 	// no-progress result should stop rather than churn.
 	loopDoomThreshold = 5
 
+	// loopMaxFailures ends a loop after this many consecutive failed iterations, so
+	// a loop whose iterations keep erroring stops instead of churning to the
+	// iteration cap (a failed turn has no final answer, so the doom guard can't see
+	// the lack of progress).
+	loopMaxFailures = 3
+
 	// loopPollInterval is how often the controller checks whether a loop is due.
 	// A loop only fires while idle, so this is a cheap between-turn poll.
 	loopPollInterval = 1 * time.Second
@@ -63,7 +69,10 @@ type loopState struct {
 	// final answers past loopDoomThreshold end the loop.
 	lastResult string
 	repeatRun  int
-	paused     bool
+	// failRun counts consecutive failed iterations; past loopMaxFailures the loop
+	// stops (a failed turn has no final answer for the doom guard to compare).
+	failRun int
+	paused  bool
 }
 
 func (l *loopState) iterationCap() int {
@@ -342,6 +351,11 @@ func (m model) advanceLoop(id, finalAnswer string, runErr error) model {
 		return m // stopped mid-run
 	}
 	l.iteration++
+	if runErr != nil {
+		l.failRun++
+	} else {
+		l.failRun = 0
+	}
 	res := strings.TrimSpace(finalAnswer)
 	if res != "" && res == l.lastResult {
 		l.repeatRun++
@@ -353,6 +367,8 @@ func (m model) advanceLoop(id, finalAnswer string, runErr error) model {
 	switch {
 	case m.loopDone:
 		return m.removeLoop(id, fmt.Sprintf("Loop %s finished — the task reported done after %d iteration(s).", id, l.iteration))
+	case l.failRun >= loopMaxFailures:
+		return m.removeLoop(id, fmt.Sprintf("Loop %s stopped after %d consecutive failed iterations.", id, l.failRun))
 	case l.iteration >= l.iterationCap():
 		return m.removeLoop(id, fmt.Sprintf("Loop %s stopped at its %d-iteration cap.", id, l.iteration))
 	case l.repeatRun >= loopDoomThreshold:
--- a/internal/tui/loop.go
+++ b/internal/tui/loop.go
@@ -360,11 +360,33 @@ func (m model) advanceLoop(id, finalAnswer string, runErr error) model {
 	}
 
 	if l.mode == loopModeSelfPaced {
-		l.nextRunAt = m.now().Add(clampSelfPaceDelay(m.loopNextWake))
+		// A model-chosen delay (loopNextWake, set via the loop_control channel) wins;
+		// otherwise fall back to an adaptive cadence that checks back often while work
+		// is fresh and settles to a slow heartbeat as iterations accrue.
+		delay := m.loopNextWake
+		if delay <= 0 {
+			delay = adaptiveSelfPaceDelay(l.iteration)
+		}
+		l.nextRunAt = m.now().Add(clampSelfPaceDelay(delay))
 	} else {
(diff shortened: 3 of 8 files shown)


jatmn (changes requested):

## Findings

- [P2] Stop carrying loops across session switches
  `internal/tui/session.go:51`
  Loops are described as session-scoped, but `startNewSession()` clears the session/transcript state without clearing `m.loops` or invalidating the loop ticker. The same pattern exists on other session replacement paths like `/resume`, so an active loop from the previous conversation can keep firing after the user starts or resumes an unrelated session, injecting the old loop prompt into the new context. Please either stop active loops when the TUI switches/rebuilds sessions or persist and reattach them to their owning session instead of leaving them as global model state.

- [P2] Wire the self-paced control path before advertising it
  `internal/tui/loop.go:378`
  The self-paced path still depends on `m.loopNextWake` and `m.loopDone`, but there is no runtime writer for either field: the only production writes reset them before a loop fires, and the only non-reset writes are tests. That means a bare `/loop keep tidying docs` cannot actually let the model choose its next wake or report done as issue #501 and the in-app usage text describe; it can only follow the hard-coded adaptive delay until it hits repeated-output/failure/iteration guards. Please either add the promised loop-control channel/tool and playbook wiring, or narrow the command/docs/PR scope so self-paced mode is honestly described as adaptive polling without model-controlled completion.

- [P2] Reject missing custom slash commands instead of looping them as prompts
  `internal/tui/loop.go:337`
  `validateLoopTarget()` accepts any slash-prefixed `commandUnknown` value so custom commands can be scheduled, but `fireLoopPrompt()` falls back to `launchPrompt(prompt)` when `handleUserCommand()` cannot resolve the command. As a result, `/loop 5m /babysit-prs` succeeds even if that custom command file does not exist, and every tick sends the literal text `/babysit-prs` to the model instead of following the normal unknown-command error path. Please resolve custom slash commands when creating the loop, or stop/report the loop on the first unresolved command, so only real custom commands are scheduled.

- [P2] Add the active-loop leave gate before `/clear` and `/quit`
  `internal/tui/model.go:3685`
  The linked issue calls for a leave-warning gate before `/quit` and `/clear` when loops are active, but the current handlers do not gate either path: `/clear` wipes the transcript first and only then appends a note that loops are still running, while `/exit`/`/quit` goes straight to `quit()` when idle without checking `loopActive()`. For a foreground token-spending loop, the user needs a chance to stop or confirm before clearing the visible context it will continue from or closing the session. Please add the pre-action warning/confirmation gate for active loops on both paths.

Vasanthdev2004:
Thanks, these were all fair. Pushed 47c799ea.

**1. Session scope** — loops now stop on `/new` and `/resume` (`clearLoopsForSessionSwitch`), so a loop from the old conversation can't keep firing into a new one. Went with stopping them rather than persist-and-reattach: v1 keeps loops as in-memory state, and cross-restart durability is already parked as a follow-up in the issue.

**2. Self-paced control path** — made it real instead of narrowing the docs. Each self-paced iteration now runs a maintenance playbook, and the model ends its reply with a control line: `LOOP: DONE` stops the loop, `LOOP: CONTINUE [interval]` sets the next wake. `advanceLoop` parses that, so the done/next-wake decision comes from the model's actual output. Dropped `loopNextWake`/`loopDone` entirely — they were the dead fields you flagged.

**3. Custom commands** — `validateLoopTarget` is a method now and resolves the command against the loaded user commands at creation time, rejecting anything that doesn't exist. So `/loop 5m /babysit-prs` fails up front when there's no such command instead of scheduling it and re-sending the literal text every tick.

**4. Leave gate** — `/clear` and `/quit` warn once and need a second confirm while loops are active; any other command in between disarms it.

Also folded in the two remaining scope items the pre-merge check wanted: the `loop.md` playbook (it's the self-paced preamble) and age-based auto-expire (24h, next to the iteration cap), plus the two CodeRabbit test nits.

One extra note since it touches what you flagged: I ran a review pass over the above and caught a few edge cases in the new code — the control-line regex missed CRLF replies and a `LOOP: DONE` line with a trailing note (so a self-paced loop could drop its own stop signal), and a loop could fire behind an open `/resume` picker and complete into the session you switched to. Both fixed in the same push with tests.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P2] Keep the remaining loops scheduled after stopping one loop
  `internal/tui/loop.go:312`
  `stopLoop(id)` removes the requested loop by returning `removeLoop(...)`, but `removeLoop` always invalidates the current poll tick and sets `loopTicking=false`. Unlike the `advanceLoop` completion path, the manual `/loop stop <id>` path never calls `ensureLoopTick()` afterward, so if a user has multiple loops and stops only one of them, the remaining loops stay in `m.loops` and still appear in `/loop list`/the footer, but no poll tick is scheduled to fire them anymore. Please restart the loop ticker whenever a manual single-loop stop leaves other loops active, and add coverage for that multi-loop stop case.

- [P2] Apply the self-paced control protocol to custom command loops
  `internal/tui/loop.go:352`
  Bare `/loop /some-command` is accepted as a self-paced loop because the command usage advertises `<prompt|/command>` and `parseLoopCommand` makes every no-interval target self-paced. However `firePrompt()` only wraps self-paced non-slash prompts in `loopSelfPacePlaybook`; slash targets are sent verbatim to `handleUserCommand()`, which expands the custom command template and launches it normally. Unless every custom command author already included the hidden `LOOP: DONE` / `LOOP: CONTINUE` protocol in their template, the model is never told how to stop the self-paced command or choose its next wake, so the advertised self-paced behavior degrades to adaptive polling until another safety guard fires. Please either wrap/augment expanded custom-command prompts with the same loop-control instructions, or reject/narrow bare self-paced `/command` loops so only fixed-interval command loops are advertised.

kevincodex1:
bro @Vasanthdev2004 please rebase to main and fix conflicts

Vasanthdev2004:
Rebased onto main (kept the new `/retry`, `/edit`, `/copy`, `/export` next to `/loop`) and fixed the round-2 findings.

**Ticker re-arm** — you're right, `stopLoop` went through `removeLoop`, which kills the poll ticker and never restarted it, so stopping one of several loops stranded the rest with nothing scheduled to fire them. `stopLoop` now re-arms via `ensureLoopTick` after the removal (which no-ops when it was the last loop). Added a multi-loop-stop test asserting the remaining loop stays scheduled.

**Self-paced `/command`** — went with narrowing rather than wrapping. A `/command` loop now needs an interval; a bare self-paced `/command` is rejected with a pointer to the interval form. Self-paced mode is for a free-form goal — it gets the playbook + `LOOP:` control line — and a custom command's expanded template doesn't carry that, so accepting it would just silently degrade to polling, exactly as you said. Test added.

**(CodeRabbit) `/resume` clearing loops on a same-session resume** — good catch. `/resume latest` or `/resume <currentID>` can land on the already-active session, whose loops belong to it, so the teardown is now guarded on an actual session-id change.

Also picked up the three test-strengthening nits: the skipped-behind-modal loop now asserts it stays scheduled, the footer test checks the next-wake time, and the confirmed `/clear` test asserts the transcript is actually wiped (so a no-op clear would fail).

Build, vet, and the tui suite are green locally, and it's mergeable again.

jatmn (approved):

@Vasanthdev2004 LGTM

kevincodex1:
bro please rebase to main and kindly fix conflicts @Vasanthdev2004

Vasanthdev2004:
Good catch — narrowed the README /loop row to say "prompt or custom /command." The validator (validateLoopTarget) already rejects built-in commands, so the docs now match the behavior. 330b10fe.

kevincodex1 (approved):
LGTM

kevincodex1 (approved):

PR #506, opened 2026-07-05
feat: agent quality, caching, retry, and tooling upgrades

kevincodex1:
## Summary

Ten self-contained improvements to the core agent engine and tools:

- edit_file: fuzzy fallback matching (7-strategy cascade with uniqueness protection and a disproportion guard) when old_string does not match byte-for-byte; exact match remains the fast path
- edit_file/write_file: inline error-severity LSP diagnostics in tool output so the model sees a break it introduced in the same turn
- lsp: expand server registry from 4 to 28 languages; doctor grades only tier-1 servers and lists the long tail informationally
- agent: consecutive read-only tool calls execute concurrently (bounded at 8), consumed in original order, never spanning a mutating call
- anthropic: cache_control breakpoints on the last two conversation messages so the transcript prefix stays a prompt-cache hit
- providerio: retry schedule raised to 6 attempts with exponential 2s-30s backoff when the server sends no Retry-After
- tools: truncated exec/bash output spills to a redacted temp file referenced in the truncation notice; bash output (previously unbounded) now shares exec_command's token budget
- web_fetch: dependency-free HTML-to-markdown conversion (format: auto|raw|markdown) and larger limits (256KiB default, 2MiB max, 30s)
- modelregistry: live models.dev overlay refreshes context limits and base pricing from a TTL-cached fetch; opt-in at the CLI entrypoint
- tools: opt-in format-on-write (ZERO_FORMAT_ON_WRITE=1) runs standard formatters across 27 extensions, ordered before the file-tracker re-baseline so conflict detection is preserved

## Checklist

- [x] `gofmt` clean.
- [ ] UI changes include screenshots or a short recording where possible. — N/A, no UI changes

## Summary by CodeRabbit

* **New Features**
  * Added inline diagnostics after file edits and writes to surface issues immediately.
  * Improved web fetching with automatic HTML-to-markdown conversion and a raw output option.
  * Added optional format-on-write for supported file types.
* **Bug Fixes**
  * Edit operations are more tolerant of spacing, indentation, and line-ending differences.
  * Large command outputs can now save and reference full output when truncated.
  * Tool execution is faster for consecutive read-only actions while preserving original result order.

### DIFF
--- a/internal/agent/file_diagnostics.go
+++ b/internal/agent/file_diagnostics.go
@@ -1,0 +1,44 @@
+package agent
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/Gitlawb/zero/internal/lsp"
+)
+
+// fileDiagnosticsTimeout bounds one inline post-edit diagnostics check so a
+// slow or wedged language server can never hang a tool call; on timeout the
+// edit simply reports without a diagnostics block.
+const fileDiagnosticsTimeout = 10 * time.Second
+
+// NewFileDiagnostics adapts an *lsp.Manager to the per-edit inline diagnostics
+// callback (tools.RunOptions.Diagnostics): it reads the just-written file,
+// checks it against the file's language server, and formats error-severity
+// diagnostics for the model. Warnings and hints are excluded — nagging about
+// style on every edit is noise, while a type error the edit just introduced is
+// exactly what the model should see before its next step. Returns nil when
+// manager is nil, disabling inline diagnostics entirely.
+func NewFileDiagnostics(manager *lsp.Manager) func(context.Context, string) string {
+	if manager == nil {
+		return nil
+	}
+	return func(ctx context.Context, absPath string) string {
+		text, err := os.ReadFile(absPath)
+		if err != nil {
+			return ""
+		}
+		checkCtx, cancel := context.WithTimeout(ctx, fileDiagnosticsTimeout)
+		defer cancel()
+		diagnostics, err := manager.Check(checkCtx, absPath, string(text))
+		if err != nil {
+			return ""
+		}
+		errors := lsp.FilterBySeverity(diagnostics, lsp.SeverityError)
+		if len(errors) == 0 {
+			return ""
+		}
+		return lsp.FormatDiagnostics(absPath, errors)
+	}
+}
--- a/internal/agent/file_diagnostics.go
+++ b/internal/agent/file_diagnostics.go
@@ -3,6 +3,8 @@ package agent
 import (
 	"context"
 	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/Gitlawb/zero/internal/lsp"
@@ -18,9 +20,12 @@ const fileDiagnosticsTimeout = 10 * time.Second
 // checks it against the file's language server, and formats error-severity
 // diagnostics for the model. Warnings and hints are excluded — nagging about
 // style on every edit is noise, while a type error the edit just introduced is
-// exactly what the model should see before its next step. Returns nil when
-// manager is nil, disabling inline diagnostics entirely.
-func NewFileDiagnostics(manager *lsp.Manager) func(context.Context, string) string {
+// exactly what the model should see before its next step. Diagnostics are
+// rendered with workspace-relative paths: the absolute path would put the
+// local username/home directory into the model prompt and session transcript
+// on every edit. Returns nil when manager is nil, disabling inline diagnostics
+// entirely.
+func NewFileDiagnostics(manager *lsp.Manager, workspaceRoot string) func(context.Context, string) string {
 	if manager == nil {
 		return nil
 	}
@@ -39,6 +44,19 @@ func NewFileDiagnostics(manager *lsp.Manager) func(context.Context, string) stri
 		if len(errors) == 0 {
 			return ""
 		}
-		return lsp.FormatDiagnostics(absPath, errors)
+		return lsp.FormatDiagnostics(diagnosticsDisplayPath(workspaceRoot, absPath), errors)
 	}
+}
+
+// diagnosticsDisplayPath renders absPath relative to the workspace root for
+// model-facing output, falling back to the file's base name when the path is
+// outside the workspace (a bare name still identifies the file without
+// exposing the directory layout).
+func diagnosticsDisplayPath(workspaceRoot, absPath string) string {
+	if workspaceRoot != "" {
+		if rel, err := filepath.Rel(workspaceRoot, absPath); err == nil && !strings.HasPrefix(rel, "..") {
+			return rel
+		}
+	}
+	return filepath.Base(absPath)
 }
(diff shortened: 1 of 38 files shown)


jatmn (changes requested):

## Findings

- [P1] Link an approved parent issue before continuing this PR
  `CONTRIBUTING.md:19`
  The contribution policy says community pull requests must be tied to an existing issue that already has the `issue-approved` label, but this PR is from a `CONTRIBUTOR`, has no linked issue references, and the PR checklist leaves the approved-issue item unchecked. Please link the approved parent issue, or move the implementation into the project’s internal development flow if this is core-team work.

- [P2] Put spilled command output somewhere the advertised tools can read
  `internal/tools/spill.go:26`
  The new truncation notice tells the model to use `grep` or `read_file` on the saved full output, but `spillDir` writes that file under `os.TempDir()/zero-tool-output`. Those native file tools resolve absolute paths against the workspace/read roots, so a normal session cannot follow the notice to read a `/tmp/...` spill file. Please either store the spill under an accessible session/workspace cache path or change the recovery path/tooling so the model can actually retrieve the elided output without rerunning the command.

- [P2] Emit tool-call callbacks before executing the parallel read batch
  `internal/agent/loop.go:492`
  When a run starts a parallel read-only batch, `executeParallelReadBatch` runs every tool call before the loop reaches `OnToolCall` for the first call. The TUI and headless session recorder create their tool-call rows/events from `OnToolCall`, so a slow `grep`/`glob`/`lsp_navigate` batch now does real work while the UI/session has not recorded that those calls started yet, and any execution-side events can arrive before the corresponding call event. Please preserve the existing callback ordering by emitting the batched calls before starting their goroutines, or by moving the batch execution behind the per-call `OnToolCall` notifications.

Vasanthdev2004:
fix conflict @kevincodex1

Vasanthdev2004 (changes requested):
Gave this a deep pass — ran the touched packages (green, no races surface in the suite), then traced the risky items by hand. Most of it is genuinely solid: the retry backoff + models.dev overlay (bounded, ctx-honored, validated-before-persist, off the hot path), the Anthropic cache breakpoints (valid targets, max-4 respected, incremental-cache pattern preserved, no shared-struct leak), format-on-write (formats before the tracker re-baseline so conflict detection holds, fails closed, off by default), web_fetch HTML→markdown (RE2 so no ReDoS, capped before conversion), the LSP registry, and the spill security hardening (symlink/ownership race is fine under the sticky-/tmp model, secrets redacted before spill) all check out. Nice work.

Two things to fix before it lands, plus the rebase:

**1. [blocker] Data race on `OnPermission` in the parallel read batch** — `internal/agent/parallel_tools.go:60`
`executeParallelReadBatch` serializes `OnPermissionRequest` under `promptMutex` but leaves `OnPermission` unwrapped, and each batched `executeToolCall` fires `options.OnPermission` from its own goroutine (loop.go:1110). The TUI handler appends to shared slices without a lock — `rows = append(rows, ...)` and `sessionEvents = append(sessionEvents, ...)` (model.go:4539-4542) — as does the exec recorder. Two batched reads emitting concurrently is a genuine data race on the session recording (corruption or a panic under `-race`).

It's narrower than "any two reads": `buildPermissionEvent` early-returns `ok=false` for a plain auto-allowed workspace read (`safety=Allow && action=Allow && !grantMatched && block==nil`), so the happy path doesn't fire it. But it does fire when a batched read's sandbox decision is non-trivial — a grant-matched root (common with `/add-dir`), a blocked path, or a prompt-ish action — so a model reading 2+ files under a granted dir in one batch trips it. It's also a regression: the pre-PR serial flow never had two `OnPermission` callbacks in flight at once. Cleanest fix is to not fire execution-side callbacks from the batch goroutines at all — collect the events and replay `OnPermission` (and `OnToolCall`/`OnToolResult`, which already replay here) in the sequential consume loop; or, minimally, wrap `OnPermission` in the same `promptMutex`.

(This is the area jatmn flagged as F3. The literal ordering claim there is actually benign — `OnToolCall`/`OnToolResult` are emitted in original order in the consume loop — but the adjacent `OnPermission` race is the real issue.)

**2. [major] Fuzzy edit can silently edit the wrong span** — `internal/tools/edit_replacers.go:63`
The ambiguity guard only rejects when the *resolved literal* occurs more than once (`strings.LastIndex(content, search) != index`). It returns the first unique candidate a strategy produces, without rejecting when the strategy yields two or more *distinct-shaped* candidates that each occur once. With duplicate blocks at different indentation and a whitespace-drifted `old_string` (matched by the line-trimmed / indentation-flexible replacers), it edits the first block instead of reporting `errEditFuzzyAmbiguous` — a silent wrong-span write. Exact match is a true fast path (fuzzy only runs when `occurrences==0`) and the disproportion + indentation/CRLF handling are all correct, so this is the one real gap. Fix: when `replaceAll` is false, reject if more than one distinct candidate span survives the literal-existence filter.

**Non-blocking follow-up:** the spill notice (exec_command.go:929) tells the model to `grep`/`read_file` the saved output, but it lands under `os.TempDir()/zero-tool-output-<uid>`, which the scoped `read_file`/`grep` can't open (they resolve absolute paths against the workspace read roots). So the recovery hint usually points somewhere the native tools can't reach. It degrades cleanly to the old truncate-and-drop, so it's a fast-follow, not a blocker — either register the spill dir as a read root when you emit the notice, or drop the "read_file it" line. (This is jatmn's F2 — still open, correctly identified.)

Also needs a rebase (currently `CONFLICTING`). The issue-approved gate doesn't apply here — this is core-team work, not a community PR.

Fix the race and the fuzzy ambiguity, rebase, and I'm happy to re-approve. The rest is ready.

Vasanthdev2004 (approved):
Re-reviewed the round-2 fixes — both blockers are closed, and cleanly. Approving.

**The `OnPermission` race** — right fix: `promptMutex` became a single `callbackMutex` that serializes both `OnPermissionRequest` and `OnPermission`, so the batched goroutines can't race on the session-recorder's shared slices. No re-entrancy (the two callbacks fire at different points in `executeToolCall`, never nested), so no deadlock, and the comment captures the exact scenario.

**The fuzzy ambiguity** — you took it to the root, further than I'd flagged. `fuzzyEditMatch` now collects the *distinct* candidate spans per strategy and returns `errEditFuzzyAmbiguous` when more than one survives (and it's not `replaceAll`). And you caught the thing that would have masked it: the block-anchor replacer was collapsing to the single "best" span *before* the ambiguity check could see the competitors — now it surfaces every qualifying span, so one clear winner still resolves while two plausible blocks correctly reject. That's the wrong-span write closed for good.

**F2 (spill reachability)** — nice, this went past the non-blocking bar: `resolveSpillReadPath` lets the scoped `read_file`/`grep` open spill files (per-uid, zero-created, redacted), wired into the workspace read resolver with a symlink-resolved containment gate and a hardening test, so the "read_file it" recovery notice actually works now instead of pointing at an unreachable `/tmp` path.

Rebase is done (`MERGEABLE`), build's clean, the agent + tools suites pass locally, and all 9 CI checks are green. This was a big, careful PR and the fixes are solid — good to merge.

PR #509, opened 2026-07-05
fix: Termux/Android support — PRoot scroll, SIGSYS sandbox, build docs

PatrickNoFilter:
## Summary

Three Termux/Android portability fixes for Zero:

### 1. PRoot scroll fix (TUI mouse mode)

Replace Termux env-var detection with **`isRunningUnderPRoot()`** via `/proc/self/status` `TracerPid`, cached with `sync.OnceValue`. Under PRoot, falls back to `CellMotion` instead of `AllMotion` — fixing broken touch-gesture scrolling while keeping wheel events, clicks, and drag working.

Includes non-controversial TUI scroll fixes:
- **Shift+Up/Shift+Down** before plain `KeyUp`/`KeyDown` with permission/ask-user cursor and composer guard
- **Ctrl+U/Ctrl+D** direction fix (Ctrl+U = up/-1, Ctrl+D = down/+1)
- Proper modal guards so Ctrl+U/D doesn't intercept wizard/picker/spec-review keybindings
- Unit tests for direction regression and composer guard

### 2. Sandbox `faccessat2` avoidance

Replace `exec.LookPath` with a `lookupExecutable` helper that uses `os.Stat` instead of the `faccessat2` syscall (syscall 439). This syscall is blocked by Samsung's seccomp filter on Android, causing SIGSYS termination. `os.Stat` uses the `newfstatat`/`statx` syscalls which are universally permitted.

### 3. Termux build documentation

Added `docs/INSTALL.md` section covering:
- Building with `GOOS=android` to avoid SIGSYS (Go 1.26+ stdlib fix)
- DNS resolution via `proot` bind-mount for `/etc/resolv.conf`
- PRoot scroll behavior explained
- Free-tier provider setup (OpenCode Zen)

## Commits

1306ab5 fix(tui): TracerPid PRoot detection, Ctrl+U/D direction, composer guard
c8f899d fix(sandbox): avoid faccessat2 for Termux/Android portability
9696920 docs(install): add Termux/Android build and configuration guide

## Notes

- Created from explicit upstream `main` (`f401c66`), not rebased from the closed PR #503
- Ref #507

## Summary by CodeRabbit

* **New Features**
  * Added/extended transcript navigation shortcuts: Shift+Up/Down for line-by-line movement and Ctrl+U/Ctrl+D for half-page scrolling.

* **Bug Fixes**
  * Prevented navigation shortcuts from interfering while composing text or when higher-priority dialogs are focused.
  * Improved mouse hover behavior in restricted terminal environments and enhanced helper availability detection on constrained systems.

* **Documentation**
  * Added Termux (Android) installation guide, including DNS setup and an example provider configuration.

* **Tests**
  * Added regression and unit tests for prompt cursor movement, composer guard behavior, and tracer detection logic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,7 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 # Superpowers skill artifacts (brainstorm mockups, specs, plans) — local only
 .superpowers/
 docs/superpowers/
+
+# Termux/Android build artifacts
+zero-termux-arm64
+zero-linux-sandbox
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -173,6 +173,61 @@ Linux native sandboxing also requires Bubblewrap to be installed.
 
 macOS uses the system sandbox and does not need an extra helper binary.
 
+### Termux (Android)
+
+Zero can run natively on Android via [Termux](https://termux.dev/). Build with
+`GOOS=android` to avoid the `faccessat2` syscall that is blocked by Samsung's
+seccomp filter on Android:
+
+```bash
+# Install Go in Termux
+pkg install golang
+
+# Build Zero for Android
+git clone https://github.com/Gitlawb/zero.git
+cd zero
+CGO_ENABLED=0 GOOS=android GOARCH=arm64 go build -ldflags="-s -w" -o zero ./cmd/zero
+
+# Move into PATH
+mv zero ~/.local/bin/
+```
+
+> **Why `GOOS=android`?** Go 1.26+ detects `runtime.GOOS == "android"` and skips
+> the `faccessat2` syscall inside `os/exec.findExecutable`, falling back to
+> permission-bit checks. Without this flag, Android's seccomp sends SIGSYS and
+> kills the process whenever Zero looks up a binary on `PATH` (git, sh, etc.).
+
+**DNS.** Android does not expose `/etc/resolv.conf`. Go's pure-Go DNS resolver
+needs one. Use `proot` to bind-mount Termux's resolver config:
+
+```bash
+pkg install proot
+proot -b "$PREFIX/etc/resolv.conf:/etc/resolv.conf" zero
+```
+
+Create a wrapper at `~/.local/bin/zero` to avoid typing proot every time:
+
+```bash
+#!/data/data/com.termux/files/usr/bin/bash
+exec proot -b "$PREFIX/etc/resolv.conf:/etc/resolv.conf" ~/.local/bin/zero.bin "$@"
+```
+
+**Scroll.** On native Termux (not under PRoot), mouse scrolling works out of the
+box. The TUI uses Bubble Tea's `AllMotion` mouse mode by default. If you run Zero
+inside PRoot (e.g. through proot-distro), the scroll fix activates `CellMotion`
+to avoid PRoot's ptrace interference with the 1003 escape sequence.
+
+**Providers.** Zero works with any OpenAI-compatible provider on Termux. For
+example, to use OpenCode Zen's free tier:
+
+```bash
+zero providers add opencode \
+  --name opencode \
+  --model deepseek-v4-flash-free \
+  --base-url https://opencode.ai/zen/v1 \
+  --set-active
+```
+
 Windows source builds can use the main `zero.exe` as the command runner and setup
 helper through Zero's built-in self-dispatch path. If you want a release-style
 layout anyway, build the standalone helper executables next to `zero.exe`:
--- a/internal/sandbox/manager.go
+++ b/internal/sandbox/manager.go
@@ -122,6 +122,23 @@ func (manager SandboxManager) Backend() Backend {
 	return manager.backend
 }
 
+// isExecutable checks whether a file is executable. On Unix, this checks the
+// execute permission bits. On Windows, Go's os.FileMode does not set execute
+// bits, so we check for common executable extensions instead.
+func isExecutable(fi os.FileInfo) bool {
+	if !fi.Mode().IsRegular() {
+		return false
+	}
+	if runtime.GOOS == "windows" {
+		name := strings.ToLower(fi.Name())
+		return strings.HasSuffix(name, ".exe") ||
+			strings.HasSuffix(name, ".com") ||
+			strings.HasSuffix(name, ".bat") ||
+			strings.HasSuffix(name, ".cmd")
+	}
+	return fi.Mode()&0o111 != 0
+}
+
 // lookupExecutable checks whether a named binary exists and is executable,
 // using os.Stat instead of exec.LookPath. exec.LookPath internally uses the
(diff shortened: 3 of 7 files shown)


anandh8x:
Thanks for coming back with the `TracerPid` approach — that's exactly the right technical fix for detecting PRoot specifically instead of Termux broadly, and the composer/modal guard ordering looks correct this time (traced it against the fallback dispatch chain). Good work on that part.

Before we go further, we want to understand the actual use case: **what's the reason to run Zero inside PRoot at all, given Termux already works natively (#455)?** PRoot exists to solve problems like missing packages, FHS path assumptions, or glibc requirements — none of which apply to Zero, since it's a single self-contained Go binary with no runtime dependency on any of that. Genuinely asking — if there's a real workflow reason (something specific PRoot gives you that native Termux doesn't, for running Zero specifically), we'd like to hear it, since that changes the calculus here.

If the answer turns out to be "no strong reason, just how I had my environment set up," then we'd lean toward not carrying this: `isRunningUnderPRoot()` adds a new detection function and a new mouse-mode branch that has to be reasoned about and tested on every future TUI mouse/keybinding change, for a configuration we don't document or recommend and that currently has one reporter. That's ongoing cost for a case the supported native path doesn't have.

Separately, worth fixing regardless of the outcome above: this PR references #507, which is about `zero update` hitting GitHub's API rate limit on shared IPs — unrelated to TUI/mouse handling. The actual parent issue for this content (#505) is the one we closed earlier, so this would need its own approved issue either way.

One thing that stands on its own regardless: the `Ctrl+U`/`Ctrl+D`/`Shift+↑`/`Shift+↓` scroll shortcuts are useful on any terminal without PgUp/PgDn, not specific to Termux/PRoot. Happy to take those independently if you want to split them out.

Keeping this open for your take on the use-case question above.

PatrickNoFilter:
The use case for running Zero inside PRoot:

**Hermes Agent** (an AI agent framework by Nous Research) uses Zero as its terminal UI. Hermes runs inside a PRoot/Alpine sandbox on Termux because:

1. Python ML toolchain — Hermes depends on `numpy`, `pytorch`, `transformers`, etc. Native Termux package management (PEP 668, library conflicts, missing wheels) makes this brittle. A clean Alpine distro via PRoot avoids all of that.

2. The `TracerPid` fix is general-purpose anyway — it's the _correct_ way to detect any ptrace-based environment (PRoot, strace, debuggers, some containers). It's not Termux-specific.

3. The keyboard fixes (composer guard, Ctrl+U/D direction) are entirely orthogonal to PRoot — they're general TUI improvements.

So the PRoot detection is just one part of a larger set of fixes. `TracerPid` is the right approach for it. Happy to split the PR if you'd prefer the keyboard fixes separate from the environment detection.

kevincodex1:
can you address coderabbit comments @PatrickNoFilter ?

PatrickNoFilter:
**@kevincodex1** — addressed both CodeRabbit items on `fix/tui-scroll-prroot`:

**1. TracerPid matches debuggers** — Added an inline comment in `mouse.go` documenting that `TracerPid` detects any ptrace tracer (PRoot, gdb, strace, dlv), not just PRoot. That is intentional: under any ptrace tracer the AllMotion (1003) sequence is unreliable, so CellMotion is the safer fallback. The only loss is hover-highlighting, which has no functional impact.

**2. Testability** — Extracted `parseTracerPid(data []byte) bool` as a testable helper and added 5 unit tests covering: missing `TracerPid`, `TracerPid: 0`, `TracerPid: 42`, malformed value, and empty input.

All existing mouse + permission tests continue to pass.

Vasanthdev2004 (approved):
Verified this locally since CI is gated on fork PRs — build's clean and the tui suite passes. The fix is careful.

The PRoot detection is the good kind of thoughtful. TracerPid via `/proc/self/status` is more robust than sniffing Termux env vars, and you pre-empted the obvious objection right in the comment: it detects any ptrace tracer (gdb/strace/dlv too), not PRoot specifically, and that's fine because CellMotion-under-a-tracer only costs hover highlighting. `parseTracerPid` taking `[]byte` instead of reading the file is exactly right for testing, and `sync.OnceValue` keeps it to one read.

Directions all check out against the existing convention (positive = up, from PgUp): Shift+Up = `scrollChat(1)`, Shift+Down = `scrollChat(-1)`, Ctrl+U = `+page` (up), Ctrl+D = `-page` (down). The guard order is right too — shifted arrows and Ctrl+U/D are handled before the plain-arrow/composer path so they aren't swallowed, with proper fallthrough to the permission/ask-user cursors, the wizards, the picker, and a composer guard for multiline navigation.

- The Ctrl+U/D comments say "half a page," but `chatPageScrollLines()` is a full page (`height-8`), so Ctrl+U/D scroll the same amount as PgUp/PgDn. Either make them half or fix the comment — vim/less users expect Ctrl+U/D to be a half page.
- Heads up that this needs a maintainer to approve the workflow run: the Go CI is gated on fork PRs and hasn't executed yet, so I ran build + the tui suite locally to stand in for it.

On process: the bug issue #505 got closed "completed" at the same moment #503 closed, but #503 never merged, so the bug is still live in main — this is the actual fix. I've reopened #505 and marked it `issue-approved`.

Good work. Approving.

kevincodex1 (approved):
looks good to me! thanks

anandh8x:
A few updates now that more of this has resolved:

**Syscall fix**: good catch, but it actually cuts against the PRoot case, not for it — PRoot is the standard workaround for exactly the syscall-blocking problem this fixes. Once Zero stops making the blocked call, that's one less real reason to need PRoot for Zero specifically.

**Hermes Agent**: asked twice, got the same unsubstantiated claim back twice. Need something checkable this time — a repo link, docs, a config, anything showing Hermes Agent actually embedding Zero as its terminal UI. If it's not shareable, that's a fine answer too, but it changes what we merge.

**Process**: PR still says "Ref #507" (the unrelated `zero update` rate-limit issue) — #505 is the one that's reopened and `issue-approved`. Please fix the reference regardless of the above.

**Given all that, let's split it**: Ctrl+U/D + Shift+Up/Down + the syscall fix + the Termux docs stand on their own merit and don't need the PRoot question answered — that can merge once the half-page comment @Vasanthdev2004 flagged is fixed. `isRunningUnderPRoot()` + the mouse-mode fallback stays in its own PR pending real evidence on the use case.

PatrickNoFilter:
Update: this PR was actually merged into main as `0f69d99`. The sandbox `lookupExecutable` fix (avoids `faccessat2` SIGSYS on Android), the `isRunningUnderPRoot()` detection via `/proc/self/status TracerPid`, and the Termux build docs all landed. Thanks for the reviews and merging! 🎉

For anyone running Zero on Termux/Android: build with `GOOS=android` to avoid the SIGSYS crash. Inside PRoot, the TracerPid detection will automatically fall back to CellMotion mouse mode for working touch scroll.

PR #515, opened 2026-07-05
feat(openai): forward prompt_cache_key for server-side prefix cache routing

kevincodex1:
## Summary

Send the session ID as OpenAI's prompt_cache_key on every completion request so consecutive turns route to a backend replica that already holds the conversation prefix in its prompt cache, instead of re-billing the full ~11k-token system+tools prefix each turn on OpenAI-compatible endpoints (which get no cache_control breakpoints, unlike Anthropic).

Keyless requests serialize byte-identically to before, and ZERO_DISABLE_PROMPT_CACHE_KEY=1 suppresses the field for strict endpoints that reject unknown parameters. The Codex (Responses API) flavor is deliberately untouched: the ChatGPT backend is a separate, stricter surface.

## Checklist

      No linked issue — core-team change, same waiver as #506.
      Build and vet clean. Full test suite passes except the 6 pre-existing
      `internal/cli` failures caused by a local `opengateway` provider profile
      ("openai provider opengateway requires official baseURL") — byte-identical
      failure set on `main`, unrelated to this change.
- [x] `gofmt` clean.
      `TestOpenAIRequestPromptCacheKey` covers key serialization, omission for
      keyless requests, and the `ZERO_DISABLE_PROMPT_CACHE_KEY` kill switch.
      `internal/providers/openai`, `internal/zeroruntime`, and `internal/agent`
      all pass under `-race`.
      N/A — no UI changes (request serialization only).

## Summary by CodeRabbit

* **New Features**
  * Added support for session-scoped prompt caching across chat requests, including retries and final-answer flows.
  * OpenAI requests now include cache-routing information when available.

* **Bug Fixes**
  * Improved consistency so repeated or retried requests can reuse the same caching behavior.
  * Added a safeguard to disable cache-key forwarding via an environment setting when needed.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -173,6 +173,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			Messages:        copyMessages(messages),
 			Tools:           exposed,
 			ReasoningEffort: options.ReasoningEffort,
+			PromptCacheKey:  options.SessionID,
 		}
 
 		// Report the per-category context budget for this turn so a surface can
@@ -208,6 +209,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 					Messages:        copyMessages(messages),
 					Tools:           exposed,
 					ReasoningEffort: options.ReasoningEffort,
+					PromptCacheKey:  options.SessionID,
 				}
 				// Pre-content connect after a context-limit compaction: route through the
 				// reconnect helper so a transient upstream hiccup here doesn't fail the
@@ -268,6 +270,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 					Messages:        copyMessages(messages),
 					Tools:           exposed,
 					ReasoningEffort: options.ReasoningEffort,
+					PromptCacheKey:  options.SessionID,
 				}
 				retryStream, retryStreamErr := streamWithReconnect(ctx, provider, retryRequest, reconnectNoticeFor(options))
 				if retryStreamErr != nil {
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (approved):
Read the whole thing — clean, well-scoped win. Forwarding the session ID as `prompt_cache_key` so OpenAI-compatible endpoints (which get no `cache_control` breakpoints) can route consecutive turns to a warm replica is the right lever for that ~11k-prefix re-bill, and it's done thoroughly: all five request-construction paths carry the key, `omitempty` keeps keyless requests byte-identical to before, the Codex/Responses surface is left alone, and the test covers serialization, keyless omission, and the kill switch under `-race`.

Privacy is fine — I checked, the session ID is an opaque `zero_<timestamp>_<nano>_<counter>` with no path or username in it, so it's a routing key, not user data. And the 6 local `internal/cli` failures are the `opengateway`-profile artifact you called out, not this change — CI is green across all nine checks, which confirms it.

One real thing worth fixing before it lands (CodeRabbit flagged this too): the kill switch reads `os.Getenv(...) == ""`, so any non-empty value disables forwarding — including `ZERO_DISABLE_PROMPT_CACHE_KEY=0` and `=false`, which read as "keep it on" but would turn it off. Low-probability footgun on its own, but it's inconsistent with how the other boolean flag parses (`ZERO_FORMAT_ON_WRITE` treats `0`/`false` as unset). Match that truthy parse so the two behave the same.

One design call I'd flag rather than block: it's default-on for every openai-compatible endpoint, so a strict one that rejects unknown params would 400 until the user finds the kill switch. Rare, recoverable, and `prompt_cache_key` is a standard param now, so opt-out is a defensible default — just noting it's the spot a user could get surprised.

Approving. Fold in the env-var parse and it's a clean merge.

PR #517, opened 2026-07-05
feat(agent): guard against over-engineering the solution in the editing discipline

### LINKED ISSUE #516: system prompt: guard against over-engineering the solution, not just the diff
Zero's editing discipline caps the **diff** ("prefer the smallest change that fully solves the problem") and the **scope** ("avoid unrelated rewrites, dependency churn"), but nothing curbs over-engineering the **solution itself** — the common LLM pitfall of adding speculative abstraction, unrequested configurability, or handling for cases that can't occur. A small diff can still be over-built.

Add a short guardrail to the Editing discipline section: solve the problem as posed, nothing the user didn't ask for, and "if a 200-line solution could be 50, rewrite it."

Validated with a before/after eval on 5 over-engineering-prone tasks (config loader, memoize cache, JSON POST, dict validator, CSV reader), each solved under the current rules vs the current rules + the guardrail, then judged for simplicity and correctness:

- 3/5 treatment solutions were simpler, 0 were more complex, 0 broke correctness (avg 9 vs 11 non-blank lines).
- The baseline gold-plated on the cache (an unrequested `wrapper.cache` introspection attribute) and the POST (unrequested header-merge configurability); the guardrail versions dropped both while staying correct.
- On trivial tasks (word count, leap year) it's a harmless no-op — the baseline already produces minimal solutions there.

Vasanthdev2004:
## Summary

Adds one bullet to the Editing discipline section of the core system prompt to curb over-engineering the **solution**, not just the diff.

Zero already tells the agent to prefer the smallest change and avoid unrelated rewrites — that caps the *diff*. But a small diff can still be over-built: speculative abstraction, unrequested configurability, error handling for cases that can't occur. That specific pitfall wasn't guarded anywhere. The new bullet closes it:

> Solve the problem as posed, not a more general version of it. Add no speculative abstraction, configurability, or handling for cases that cannot occur, and nothing the user did not ask for. A small diff can still be over-built; if a 200-line solution could be 50, rewrite it.

## Validation

Before/after eval on five over-engineering-prone tasks (config loader, memoize cache, JSON POST, dict validator, CSV reader), each solved under the current editing rules vs the current rules + this guardrail, then judged for simplicity **and** correctness:

- **3/5** treatment solutions simpler, **0** more complex, **0** correctness regressions (avg **9 vs 11** non-blank lines).
- The baseline gold-plated on two: an unrequested `wrapper.cache` introspection attribute on the cache, and unrequested header-merge config on the POST. The guardrail versions dropped both and stayed correct.
- On trivial tasks (word count, leap year) it's a no-op — the baseline is already minimal there, so there's nothing to trim.

## Scope

One file, four lines, in `internal/agent/system_prompt.md`. `go build` + `go test ./internal/agent/...` green. Closes #516.

## Summary by CodeRabbit

* **Chores**
  * Updated internal editing guidance to encourage more focused, scoped changes and avoid unnecessary extra work.

### DIFF
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -63,6 +63,10 @@ work.
   targeted diffs. Match the existing indentation, imports, and idioms. Match the
   file's comment density: do not add explanatory comments unless the user asks or
   the code is already comment-dense.
+- Solve the problem as posed, not a more general version of it. Add no
+  speculative abstraction, configurability, or handling for cases that cannot
+  occur, and nothing the user did not ask for. A small diff can still be
+  over-built; if a 200-line solution could be 50, rewrite it.
 - Preserve behavior you were not asked to change. Do not delete or rewrite code
   you did not author unless the task requires it; if you must, say so.
 


kevincodex1 (approved):
LGTM

PR #518, opened 2026-07-05
perf: universal tool-output ceiling with spill + async post-edit diagnostics

kevincodex1:
## Summary

Two structural fixes from the perf/token audit vs opencode — the biggest
recurring token leak and the biggest recurring latency tax in the loop.

### Universal token-based output ceiling (1825ea4)

Every tool result is now capped at the registry boundary — 16k estimated
tokens by default, `ZERO_TOOL_OUTPUT_CEILING_TOKENS`-tunable, `0` disables —
unless the tool manages its own deliberate budget. Truncation keeps head+tail
and spills the full (already redaction-scrubbed) output to the existing spill
dir, re-readable via grep/read_file, so nothing is lost — only deferred.

This closes the three paths a single call could flood the context window
through (and then re-bill on every following turn):

- `web_fetch`: up to 256KiB default / 2MiB max of non-HTML body, previously unbudgeted
- `skill`: full SKILL.md returned uncapped
- MCP-served tools: no budget at all

Self-budgeting tools opt out via an unexported marker interface (an MCP
server can never exempt itself): `bash`, `exec_command` (model-raisable),
`read_file`, `read_minified_file`, `grep`, `glob`, `list_directory`. Browser
snapshots deliberately stay under the net.

bash itself is tightened to match the unified token model: emit budget
96KiB → 32KiB per stream (~24k → ~8k tokens worst case per stream), while
capture retention stays at 96KiB head+tail — so the new sectioned bash spill
covers 6× more of the output than the transcript shows, and the spill path
also lands in `meta["spill_path"]`.

### Async post-edit LSP diagnostics (b227599)

Inline diagnostics blocked every `edit_file`/`write_file` result on the
language server: a ≥300ms quiet-period debounce that reset on each publish,
seconds on the first edit per language, capped at 10s — paid on every edit
of every session.

Collection is now asynchronous: a file is enqueued the moment a mutating
tool finishes, a single worker checks it while the rest of the batch (and
any self-correct verification) runs, and the loop drains completed results
just before building the NEXT provider request, appending errors as a
user-role nudge. The model always issues another request after an edit turn
— it has to read its tool results — so it still sees introduced errors at
the same decision point as before, with no tool call ever stalling on the
language server.

Consistency: a file re-edited before its check runs is checked once in its
final state (the checker re-reads the file); re-edited after, it re-queues
and the newer result wins. A drain that finds the worker still busy gives up
after 3s and delivers on a later turn — results stay accurate because any
further edit re-enqueues the file. The tools-side inline mechanism
(`tools.RunOptions.Diagnostics`) is kept for direct API callers; the loop
just no longer wires it.

## Checklist

      No linked issue — core-team change, same waiver as #506.
      Build and vet clean. Full suite passes except the 6 pre-existing
      `internal/cli` failures from the local `opengateway` provider profile
      ("openai provider opengateway requires official baseURL") —
      byte-identical failure set on `main`, unrelated to this change.
- [x] `gofmt` clean.
      Registry-level ceiling tests (cap + spill content, self-budgeting
      exemption, env override, small-output passthrough), bash spill tests,
      async-collector unit tests (deferred delivery on slow checks, re-edit
      latest-wins, nil no-op), and an end-to-end `Run` test proving the
      diagnostics nudge lands in the follow-up request and never leaks into
      the tool result. `internal/tools` and `internal/agent` (all new
      concurrency lives there) pass under `-race`.
      N/A — no UI changes; TUI/CLI diffs are comment-only.

## Summary by CodeRabbit

* **New Features**
  * Added a universal per-call tool output ceiling (env-configurable); when exceeded, output is truncated and the full redacted transcript is saved with spill hints for later review.
  * File diagnostics now run in the background after edits and are delivered as a “nudge” message on the next request (including the final turn).
* **Bug Fixes**
  * Repeated edits refresh diagnostics to avoid stale results.
  * Late diagnostics are no longer dropped during run finalization.
  * Spill files are further protected by an additional secrets redaction pass.

### DIFF
--- a/internal/agent/async_diagnostics.go
+++ b/internal/agent/async_diagnostics.go
@@ -1,0 +1,168 @@
+package agent
+
+import (
+	"context"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+)
+
+// Post-edit LSP diagnostics used to run synchronously inside edit_file /
+// write_file: every edit blocked its tool result on a fresh-publish wait with
+// a quiet-period debounce (≥300ms floor that reset on each publish, 10s cap) —
+// the largest recurring latency tax in an edit-heavy session. Collection is
+// now asynchronous: a file is enqueued the moment a mutating tool finishes, a
+// single worker checks it while the rest of the tool batch (and any
+// self-correct verification) runs, and the loop drains completed results just
+// before building the NEXT provider request, appending errors as a user-role
+// nudge. The model always issues another request after an edit turn — it has
+// to read its tool results — so it still sees introduced errors at the same
+// decision point as before, without any tool call stalling on the language
+// server.
+//
+// A file re-edited before its check runs is checked once in its final state
+// (the checker re-reads the file); a file re-edited after its check completed
+// is re-enqueued and the newer result wins. When the worker is still busy at
+// drain time, the drain gives up after a short wait and the results are
+// delivered on the following turn instead — they stay accurate because any
+// further edit re-enqueues the file.
+
+// asyncDiagnosticsDrainTimeout bounds how long a turn waits on the in-flight
+// check before deferring delivery to the next turn. Most checks finish well
+// under it (debounce floor + analysis); a cold language server on a large
+// repo does not get to stall the loop the way the old inline 10s cap could.
+// A var so tests can shorten the wait.
+var asyncDiagnosticsDrainTimeout = 3 * time.Second
+
+// asyncDiagnosticsNudge prefixes the drained diagnostics blocks; phrased like
+// the old inline block so the model reacts the same way.
+const asyncDiagnosticsNudge = "Diagnostics after your recent edits (fix any errors you introduced):\n"
+
+type asyncDiagnostics struct {
+	check         func(context.Context, string) string
+	workspaceRoot string
+
+	mu      sync.Mutex
+	queue   []string
+	queued  map[string]bool
+	results map[string]string
+	// working is non-nil while the worker goroutine runs and is closed when it
+	// exits, so drain can wait for quiescence without polling.
+	working chan struct{}
+}
+
+// newAsyncDiagnostics returns nil when check is nil, and every method on a nil
+// receiver is a no-op — callers wire it unconditionally.
+func newAsyncDiagnostics(check func(context.Context, string) string, workspaceRoot string) *asyncDiagnostics {
+	if check == nil {
+		return nil
+	}
+	return &asyncDiagnostics{
+		check:         check,
+		workspaceRoot: workspaceRoot,
+		queued:        map[string]bool{},
+		results:       map[string]string{},
+	}
+}
+
+// enqueue schedules changed files for a background check. Paths are
+// workspace-relative from Result.ChangedFiles (absolute when under a granted
+// extra write root), mirroring SelfCorrector's resolution.
+func (diagnostics *asyncDiagnostics) enqueue(ctx context.Context, files []string) {
+	if diagnostics == nil || len(files) == 0 {
+		return
+	}
+	diagnostics.mu.Lock()
+	defer diagnostics.mu.Unlock()
+	for _, file := range files {
+		path := diagnostics.absPath(file)
+		if diagnostics.queued[path] {
+			continue
+		}
+		diagnostics.queued[path] = true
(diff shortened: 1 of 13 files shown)


Vasanthdev2004:
@kevincodex1 can u able to fix coderabbit changes ?

Vasanthdev2004 (changes requested):
Dug into this hard — it's a strong perf change and the output-ceiling half is genuinely clean, but three things to fix before it lands, one of them a real secret leak.

**[blocker — security] The bash spill bypasses the high-confidence secret scrub.** In `budgetBashCapture`, `spillBashStreams(out, ..., errStr, ...)` (bash.go:408) gets the RAW retained streams and spills them *before* `formatBashOutput` runs `secrets.Redact` (bash.go:353-354) on the transcript. `spillTruncatedOutput` only re-applies `RedactString` — the configured-key scrub — not `secrets.Redact`'s pattern matcher (AWS keys, GitHub PATs, Slack/OpenAI keys, JWTs). So a credential that bash redacts from the transcript sits in cleartext in the spill, and the notice explicitly tells the model to grep/read_file it. That's exactly the leak `secrets.Redact` exists to prevent, reopened through the new spill path. Fix: run `secrets.Redact` on the streams before spilling (in `spillBashStreams`, or have `spillTruncatedOutput` apply it).

**[major — correctness] The final edit's diagnostics can be silently dropped — a regression from the inline path.** The drain only runs at the top of the loop (loop.go:172, `asyncDiagnosticsDrainTimeout` = 3s) and in the max-turns path (705). The natural no-tool-call completion at loop.go:494 returns the final answer with no drain. So on a cold/large-repo LSP with a single-edit task: the model edits, the check is enqueued and allowed up to 10s (`fileDiagnosticsTimeout`), the next turn's top-of-loop drain gives up after 3s, and if the model completes that turn the introduced error is never surfaced. The inline path always blocked until diagnostics were ready, so it always showed them before the final decision — this is the exact scenario the feature targets. Simplest fix: drain at the completion boundary before returning the final answer, and if it turns up errors, inject the nudge and continue one more turn instead of finishing.

**[blocker — CI] Windows smoke is red.** I ran it here (I'm on Windows): `TestAsyncDiagnosticsCollectsAndDrainsOnce` fails because it uses `"/ws"` as the workspace root, which isn't Windows-absolute — `filepath.IsAbs("\ws\broken.go")` is false on Windows, so the assertion trips. It's a test-portability issue, not a production bug (`absPath` is correct for a real absolute root), but it's red CI. Use `t.TempDir()` for the root instead of `"/ws"`.

The rest holds up — I verified the ceiling directly: MCP/external tools genuinely can't exempt themselves (the marker is an unexported method and they're in a different package, so Go won't let them satisfy it), redaction runs before the ceiling spill, truncation is rune-safe and keeps the failure tail, only `Output` is capped so structured `Meta` survives, and the async concurrency itself is race/leak/deadlock-free with correct re-edit-latest-wins. That part's solid.

Fix the three and I'll take another pass.

Vasanthdev2004:
@kevincodex1 doing

Vasanthdev2004 (approved):
Re-reviewed the fixes — all three are addressed, and I checked each against the code rather than the commit messages. Approving.

**The bash spill leak is closed.** `spillTruncatedOutput` now runs both `RedactString` and `secrets.Redact` before writing (spill.go:115-116), and since every spill — bash and the universal ceiling — routes through it, the high-confidence patterns (AWS keys, GitHub PATs, JWTs) are scrubbed from the file the model reads back, not just from the transcript. That was the important one.

**The finalization gate is right.** Before returning the final answer, the completion path calls `drainFinal` (loop.go:500), which waits the full inline-era budget once rather than the 3s per-turn timeout; if diagnostics surface an error it appends the nudge and `continue`s one more turn instead of finalizing past it, and only an empty/idle result finalizes. The max-turns path gets the same treatment (718). So the final edit's diagnostics can't be silently dropped anymore — the model still sees an introduced error before it's done, which was the whole point.

**Windows is green.** The async test uses a real workspace root now; I re-ran the agent + tools suites here on Windows and they pass, and CI is green across all nine checks.

Everything I flagged is handled, and the parts I'd already verified clean — the ceiling boundary, MCP tools genuinely not being able to self-exempt, and the async concurrency itself (race/leak/deadlock-free, latest-wins) — still hold. Good turnaround; this is a solid perf change now. Good to merge.

PR #524, opened 2026-07-05
Fix/modernize codebase

### LINKED ISSUE #523: its not a bug but modernizing the codebase
### Version / branch / commit

0.1.0

### OS and environment

Nixos 26.05, kitty, go 1.26.4

### Steps to reproduce

just gopls suggestions like minmax and standard library upgrades

### Expected behavior

clean gopls with no suggestions like minmax and standard library upgrades

### Actual behavior

gopls suggestions like minmax and standard library upgrades

_No response_

sam0uly:
## Summary

this is just modernizing the codebase like the std lib upgrades and stuff.
## Linked issue

Fixes #523

## Checklist

- [ ] `gofmt` clean.

### DIFF
--- a/internal/acp/jsonrpc_test.go
+++ b/internal/acp/jsonrpc_test.go
@@ -158,7 +158,7 @@ func TestConnSurvivesMalformedLine(t *testing.T) {
 
 	dec := json.NewDecoder(clientR)
 	var sawParseError, sawPong bool
-	for i := 0; i < 2; i++ {
+	for i := range 2 {
 		var msg struct {
 			Result any `json:"result"`
 			Error  *struct {
--- a/internal/acp/permission.go
+++ b/internal/acp/permission.go
@@ -2,6 +2,7 @@ package acp
 
 import (
 	"encoding/json"
+	"slices"
 
 	"github.com/Gitlawb/zero/internal/agent"
 )
@@ -85,12 +86,7 @@ func decisionFromOutcome(outcome RequestPermissionOutcome, offered []agent.Permi
 }
 
 func actionOffered(action agent.PermissionDecisionAction, offered []agent.PermissionDecisionAction) bool {
-	for _, a := range offered {
-		if a == action {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(offered, action)
 }
 
 // permissionToolCall builds the ToolCall descriptor embedded in a
--- a/internal/agent/command_prefix.go
+++ b/internal/agent/command_prefix.go
@@ -343,7 +343,7 @@ func validSedPrintArg(arg string) bool {
 	if body == "" {
 		return false
 	}
-	for _, part := range strings.Split(body, ",") {
+	for part := range strings.SplitSeq(body, ",") {
 		if part == "" {
 			return false
 		}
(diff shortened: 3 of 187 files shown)


sam0uly:
if you guys want to fix the `golangci-lint run` issues like staticcheck and unused and errorcheck i can do them too.

but for that i need some info.

Vasanthdev2004:
Thanks for this — modernizing to the newer Go idioms (`for i := range n`, the `min`/`max` builtins, `slices`/`maps`) is a good instinct, and the individual swaps are the right direction.

What's holding me back isn't the idioms, it's the packaging. This is 187 files across nearly every package, and a sweep that size is hard to take as one PR for a few reasons:

- **The checklist is unchecked** — can you confirm `go build ./...`, `go vet ./...`, `go test ./...` all pass and it's `gofmt` clean? At this scale that's the main risk: a `for range n` conversion where the original used `i <= n` or mutated `i`, or a `min`/`max` swap where the old helper had different edge behavior, is a silent behavior change in a core package. 61 loop rewrites is a lot of surface to eyeball by hand.
- **It collides with basically every open PR** — it touches `loop.go`, `compaction.go`, `guardrails.go`, the providers, `system_prompt.go`, which are the same files several in-flight changes live in. Landing churn this wide right now would force a painful rebase across the whole queue (or leave this one in permanent rebase).
- **It's cosmetic** (no behavior change), so it isn't worth destabilizing the tree while there's a lot in flight.

What I'd suggest instead: split it per package (or per idiom) into small, independently reviewable PRs, and ideally land them mechanically — `gopls`'s `modernize` analyzer or `go fix` can do most of this and produces a clean, tool-verified diff that's easy to trust at scale. Timing-wise it'll also go much smoother once the current queue drains a bit, so it isn't fighting every other change.

Not a no to the idea — it's a genuinely good direction, it just needs to come in as small verifiable pieces rather than one big sweep. If you want to start with a single package as a template, I'm happy to review that quickly and we can pattern the rest off it.

(One process note: #523 is really a placeholder — this isn't a bug, so it doesn't fit the issue-first flow. If we go the split-PR route, a short "modernize to Go 1.22 idioms" tracked task with the plan would be the cleaner parent.)

kevincodex1:
this is massive :)

jatmn:
Thank you for the contribution. I appreciate the time and effort you put into trying to improve the project.

For now, I’m going to close this because the PR is much broader than we can review or accept in one pass. It touches a large number of files and appears to include general modernization/refactoring work rather than a small, approved, issue-scoped change.

Please see `CONTRIBUTING.md`, specifically the pull request scope guidance here:
https://github.com/Gitlawb/zero/blob/main/CONTRIBUTING.md#L75-L78

> Pull requests that include unrelated changes, broad rewrites, formatting-only changes, or unapproved feature work may be closed without review. Scope drift will be rejected or called out for correction before a pull request can be approved.

This PR falls into that category because it is a broad modernization/refactor across many files rather than a narrowly scoped fix tied to an approved issue. Smaller, focused PRs are much easier for us to evaluate, test, and merge safely.

If there is a specific bug or functional improvement you want to propose, the best next step is to open or comment on an issue first, keep the scope small, and wait for it to be marked with `issue-approved` before opening a PR. Once there is an approved issue, a focused PR that addresses only that issue is much more likely to be reviewable.

Thanks again for wanting to help.

sam0uly:
thanks everyone for the feedback that makes sense. ill split this into smaller, issue-scoped PRs (likely by package/idiom) after opening an approved tracking issue, rather than submitting one large modernization sweep. Appreciate the guidance.

PR #526, opened 2026-07-05
Batch session event persistence

### LINKED ISSUE #525: Batch session event persistence to reduce fsync overhead
### Problem

Session persistence currently appends turn events one at a time. In the TUI path, a single turn can generate many session events, but each event append performs its own log write/sync and metadata rewrite/sync.
This creates unnecessary filesystem work during active coding sessions. For example, around 30 turn events can cause roughly 90 sync-related operations instead of a small fixed number. On slower disks, networked filesystems, or heavy sessions, this can add visible latency to the agent loop.

Relevant area:
- internal/tui/session.go
- internal/sessions/store.go

### Requested behavior

Add a batch append API for session events.

**Expected behavior:**

Add something like Store.AppendEvents(sessionID string, inputs []AppendEventInput) ([]Event, error).
Persist a batch with one session lock, one metadata read, one sequence scan, one JSONL write/sync, and one metadata write/sync.
Preserve existing AppendEvent behavior by routing it through the shared batch path.
Keep event sequence numbers contiguous and ordered.
Validate the full batch before writing so invalid input does not partially append.
Keep crash durability semantics unchanged: event log remains authoritative, metadata may be recovered if lagging.
Use the batch API from TUI session persistence.

### Why it belongs in Zero

Zero is an interactive coding agent, so turn latency matters directly. Session persistence happens on the hot path while users are actively working. Reducing repeated fsync/metadata writes is a simple, localized performance improvement that should make busy turns cheaper without changing user-visible behavior.
This also gives Zero a more scalable session persistence primitive for future agent features that naturally emit multiple events per turn.

NA

_No response_

GautamBytes:
Fixes #525

## Summary
- Add batched session event append API.
- Persist batch with one session lock, log write/sync, metadata update.
- Wire TUI session persistence and fork event copying to batch append.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session events are now saved in batches more reliably, reducing the chance of partial writes or duplicate ordering.
  * Active session details stay in sync after event updates, including event count, last event type, and timestamps.
  * Empty event submissions no longer rewrite session data, and invalid event data is rejected cleanly without affecting saved history.
  * Concurrent session updates are now handled in a stable, ordered way.

### DIFF
--- a/internal/sessions/append_events_test.go
+++ b/internal/sessions/append_events_test.go
@@ -1,0 +1,233 @@
+package sessions
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestStoreAppendEventsBatchesSequencesAndMetadata(t *testing.T) {
+	store := NewStore(StoreOptions{RootDir: t.TempDir(), Now: sequenceClock([]time.Time{
+		time.Date(2026, 6, 4, 15, 0, 0, 0, time.UTC),
+		time.Date(2026, 6, 4, 15, 0, 1, 0, time.UTC),
+		time.Date(2026, 6, 4, 15, 0, 2, 0, time.UTC),
+		time.Date(2026, 6, 4, 15, 0, 3, 0, time.UTC),
+	})})
+	session, err := store.Create(CreateInput{SessionID: "batch"})
+	if err != nil {
+		t.Fatalf("Create returned error: %v", err)
+	}
+
+	appended, err := store.AppendEvents(session.SessionID, []AppendEventInput{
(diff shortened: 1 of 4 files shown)


jatmn (approved):
@Vasanthdev2004 

## Findings

- [P2] Get the linked issue approved before moving this PR forward
  `CONTRIBUTING.md:19`
  This PR links `Fixes #525`, but the linked issue currently only has the `enhancement` label and does not have `issue-approved`. The contribution policy says community PRs must be tied to an issue that has already been reviewed and approved by the core team, and the PR author is listed as a `CONTRIBUTOR`. Please get the core-team approval reflected on #525, or have a maintainer clarify that this PR is exempt as internal team work, before continuing review/merge on this branch.

kevincodex1 (approved):
nice one, thanks!

PR #527, opened 2026-07-05
chore: fix golangci-lint findings (dead code, ineffassign, staticcheck)

euxaristia:
## Summary

Small cleanup pass from running `golangci-lint` locally (installed it after CodeRabbit's `truncateHeadTail` nitpick on #476 turned out to have no equivalent check in this repo's actual CI, which only runs `go vet`).

## Changes

- Remove three unused functions found by the `unused` linter: `safeRequestedPrefix`, `safeShellCommandTokens` (`internal/agent/command_prefix.go`), and `truncateHeadTail` (`internal/tools/bash.go`). Each was a thin wrapper with no remaining callers.
- Drop an ineffectual assignment in `stringListArgWebSearch` (`internal/tools/web_search.go`): every return path after it already sets the named return explicitly, so the assignment never took effect.
- Simplify a negated-and condition in `partition_cache_stable_test.go` via De Morgan's law.
- Use `strings.Contains` instead of `strings.Index(...) < 0` in `edit_replacers.go`.
- Fix error string style (staticcheck ST1005): lowercase `"unsafe redirect URL"` and drop the trailing period from `webFetchPublicOnlyHint` in `web_fetch.go`, updating the one test that asserted on the old casing.

Deliberately left the `errcheck` findings (unchecked `defer file.Close()` / `response.Body.Close()` / `Write()` errors) alone. Those are stylistic, not correctness bugs, and this repo's CI doesn't run golangci-lint at all, so there's no actual gate to satisfy there.

## Test plan

- `go build ./...`
- `go vet ./...` (clean, matches actual CI)
- `go test ./internal/agent/... ./internal/tools/...` (pass)
- `golangci-lint run ./internal/agent/... ./internal/tools/... --enable-only unused,ineffassign,staticcheck` (0 issues, down from 6)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the wording and consistency of a few error messages shown during unsafe redirect handling and web fetch prompts.
  * Tightened validation and matching behavior in some command and edit workflows for more reliable results.

* **Refactor**
  * Simplified several internal code paths and cleanup logic without changing expected user-facing behavior.
  * Updated test assertions to better reflect the current output format.

### DIFF
--- a/internal/agent/command_prefix.go
+++ b/internal/agent/command_prefix.go
@@ -168,10 +168,6 @@ func cleanPrefixRule(prefix []string) []string {
 	return cleaned
 }
 
-func safeRequestedPrefix(prefix []string, command []string) bool {
-	return safeRequestedPrefixForSegments(prefix, [][]string{command})
-}
-
 func safeRequestedPrefixForSegments(prefix []string, segments [][]string) bool {
 	if len(prefix) == 0 || !sandbox.ValidCommandPrefix(prefix) {
 		return false
@@ -194,14 +190,6 @@ func safeRequestedPrefixForSegments(prefix []string, segments [][]string) bool {
 		return false
 	}
 	return matched
-}
-
-func safeShellCommandTokens(command string) ([]string, bool) {
-	segments, ok := safeShellCommandSegments(command)
-	if !ok || len(segments) != 1 {
-		return nil, false
-	}
-	return segments[0], true
 }
 
 func safeShellCommandSegments(command string) ([][]string, bool) {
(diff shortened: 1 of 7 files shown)


jatmn (approved):
@Vasanthdev2004 

## Findings

- [P2] Tie this cleanup to an approved parent issue or maintainer-approved scope
  `CONTRIBUTING.md:45`
  This PR is from a `CONTRIBUTOR` account and does not link an approved parent issue, but the project contribution policy says all community PRs require an approved parent issue with the `issue-approved` label before implementation starts. The change is also a cleanup/refactor pass driven by local `golangci-lint` output rather than a linked approved bug or feature, which falls under the same scope rule. Please link the approved issue for this cleanup, or get explicit maintainer approval for taking this contributor cleanup outside the issue-first flow.

kevincodex1 (approved):
LGTM

PR #529, opened 2026-07-05
feat: gate project-scoped hooks, plugins, and MCP servers behind workspace trust

### LINKED ISSUE #540: security: untrusted workspace .zero/ hooks, plugins, and MCP servers run code without a trust prompt
A project's `.zero/` config — hooks, plugins, and MCP server definitions — is loaded and run as soon as Zero starts in that directory. So if you clone or check out someone else's repo and run a single tool, their hook/plugin/MCP command executes outside the sandbox with no prompt. That's a silent code-execution path on any untrusted repo.

We should gate project-scoped hooks, plugins, and MCP servers behind an explicit per-workspace trust decision — fail closed until the workspace is trusted, with a clear one-time prompt/notice. #529 implements this.

beardthelion:
Fixes #540

Zero loads executable configuration from a workspace's `.zero/` directory:
`hooks.json` (commands that run around tool calls), `plugins/` (loaded
extensions), and project-scoped MCP servers. Until now that config ran with no
trust check. Hooks in particular execute as a plain child process with your
privileges and environment, outside the sandbox, and a `beforeTool` hook fires
on the first tool call the agent makes. So running Zero in a repository you
don't control (a fresh clone, a checked-out PR branch, an extracted archive)
could execute arbitrary code on your machine with no prompt, in the default
permission mode with the sandbox on.

This adds a per-workspace trust gate, the same pattern editors use for workspace
trust.

## Behavior

Project-scoped hooks, plugins, and MCP servers load only for a directory you've
explicitly trusted. User and global config are unaffected and always load. Until
a workspace is trusted, the project layer is dropped and Zero prints a one-line
stderr notice pointing at `zero trust`.

Opt in with the new command:

    zero trust                Trust the current working directory
    zero trust list           List trusted workspace roots
    zero trust remove [path]  Untrust the current directory, or a named path

## How the gate works

The trust check lives inside the load chokepoints (the hook dispatcher, plugin
activation, and every MCP spawn site) rather than at their callers, so it fails
closed by construction: an empty trust root, an unreadable store, or an untrusted
directory all resolve to "exclude the project layer." A future call site that
forgets to check can't silently default to loading project config.

Trust is an exact match on the normalized absolute path, with no ancestor or
subtree walk, so trusting a monorepo root does not implicitly trust a vendored
dependency or submodule that ships its own `.zero/`. The trust root is captured
from the original launch directory before any `--worktree` reassignment, so a
worktree of a repo you trust inherits that trust instead of being keyed on the
generated worktree path.

The store is a plain JSON file at `<user-config-dir>/zero/trust.json`, written
atomically with restrictive permissions (dir 0700, file 0600). It holds
directory paths, not secrets.

## Tests

Unit coverage for the trust store (exact match, symlink handling, atomic write),
gate coverage for hooks/plugins/MCP project exclusion, and end-to-end tests that
drive the real agent loop and the `exec --worktree` path with a fake provider,
confirming a project `beforeTool` hook runs only when the source repo is trusted,
including in the default (non-unsafe) permission mode.

## Tradeoff

Trust is at directory granularity. A branch or worktree of a repo you've already
trusted inherits that trust, so a hostile branch pushed to a repo you trust would
still load its project config. Narrowing that is out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added a `trust` command to manage trusted workspaces (trust, list, untrust/remove).
  * Workspace-trust fail-closed gating now applies to project-scoped hooks/plugins/MCP surfaces, with one-line stderr notices across TUI, `exec`, and `exec --use-spec`.
  * MCP server discovery and OAuth login are gated by workspace trust (including trust-aware `mcp check`, tools listing, and server refresh paths).
* **Bug Fixes**
  * Project-layer hooks/plugins/MCP servers stay disabled when untrusted or when trust evaluation fails; `--worktree/--use-spec` inherit trust from the original launch directory.
* **Tests**
  * Expanded unit, integration, and end-to-end coverage for trust gating, notice wording/counting, and MCP OAuth/MCP check/tool behaviors.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -711,7 +711,10 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	// collect their hooks + skill roots for the dispatcher and skill tool below.
 	// Done after specialist + MCP registration so plugin tools are part of the
 	// deferral count, and it fails OPEN — a malformed plugin is warned and skipped.
-	pluginActivation := activatePlugins(workspaceRoot, registry, deps, stderr)
+	// The interactive TUI is not worktree-reassigned, so the trust root is the
+	// launch directory itself.
+	trustRoot := workspaceRoot
+	pluginActivation := activatePlugins(workspaceRoot, registry, deps, stderr, trustRoot)
 	// Ask (not Auto) is the interactive default: in Auto, ToolAdvertised exposes
 	// only PermissionAllow tools, so prompt-gated tools (write_file/edit_file/bash/
 	// apply_patch) would never be offered to the model — the TUI could neither edit
@@ -758,6 +761,11 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	if userConfigPath != "" {
 		sttDownloadRoot = filepath.Join(filepath.Dir(userConfigPath), "stt")
 	}
+	// Build the hooks dispatcher out of the AgentOptions literal so its trust skip
+	// report can be combined with the plugin activation's, and emit at most one
+	// notice when project hooks/plugins were dropped for an untrusted workspace.
+	hookDispatcher, hookSkip := newHookDispatcherWithExtra(workspaceRoot, pluginActivation.hooks, trustRoot)
+	emitTrustNotice(stderr, hookSkip, pluginActivation.trustSkip)
 	return deps.runTUI(context.Background(), tui.Options{
 		Cwd:                  workspaceRoot,
 		Version:              version,
@@ -815,7 +823,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 			Autonomy:       "low",
 			Sandbox:        sandboxEngine,
 			FileTracker:    fileTracker,
-			Hooks:          newHookDispatcherWithExtra(workspaceRoot, pluginActivation.hooks),
+			Hooks:          hookDispatcher,
 			DeferThreshold: resolved.Tools.DeferThreshold,
 			Specialists:    specialistRuntime.specialists,
 			Skills:         pluginActivation.skillInfos(deps.skillsDir()),
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -51,7 +51,7 @@ type appDeps struct {
 	stdin            io.Reader
 	userConfigPath   func() (string, error)
 	resolveConfig    func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
-	resolveMCPConfig func(workspaceRoot string) (config.MCPConfig, error)
+	resolveMCPConfig func(workspaceRoot string, excludeProject bool) (config.MCPConfig, error)
 	newProvider      func(config.ProviderProfile) (zeroruntime.Provider, error)
 	// exportActiveProvider pins spawned children to the run's provider (production:
 	// config.SetActiveProviderEnv, set in defaultAppDeps — deliberately NOT filled
@@ -125,11 +125,12 @@ func defaultAppDeps() appDeps {
 			options.Overrides = overrides
 			return config.Resolve(options)
 		},
-		resolveMCPConfig: func(workspaceRoot string) (config.MCPConfig, error) {
+		resolveMCPConfig: func(workspaceRoot string, excludeProject bool) (config.MCPConfig, error) {
 			options, err := config.DefaultResolveOptions(workspaceRoot)
 			if err != nil {
 				return config.MCPConfig{}, err
 			}
+			options.ExcludeProject = excludeProject
 			return config.ResolveMCP(options)
 		},
 		newProvider: func(profile config.ProviderProfile) (zeroruntime.Provider, error) {
@@ -671,7 +672,11 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		return writeAppError(stderr, "failed to initialize specialist tools: "+err.Error(), 1)
 	}
 	defer closeSpecialistRuntime(stderr, specialistRuntime)
-	mcpConfig, err := deps.resolveMCPConfig(workspaceRoot)
+	// The TUI has no --worktree reassignment, so trustRoot == workspaceRoot here.
+	// Gate the project MCP layer behind the workspace-trust check (fail-closed): an
+	// untrusted workspace must not spawn its ./.zero/config.json stdio MCP servers.
+	mcpExcludeProject, _ := resolveTrust(workspaceRoot)
+	mcpConfig, err := deps.resolveMCPConfig(workspaceRoot, mcpExcludeProject)
 	if err != nil {
 		return writeAppError(stderr, err.Error(), 1)
 	}
@@ -806,7 +811,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 			var stdout, stderr bytes.Buffer
 			exitCode := runMCPWithContext(ctx, args, &stdout, &stderr, deps)
 			nextConfig := lastKnownMCPConfig
-			if refreshed, err := deps.resolveMCPConfig(workspaceRoot); err == nil {
+			if refreshed, err := deps.resolveMCPConfig(workspaceRoot, mcpExcludeProject); err == nil {
 				lastKnownMCPConfig = refreshed
 				nextConfig = refreshed
 			}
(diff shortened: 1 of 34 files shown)


kevincodex1:
bro please address coderabbit comments

beardthelion:
Addressed the CodeRabbit review in the latest push.

- Project MCP servers dropped in an untrusted workspace now surface the same trust notice as hooks and plugins, across the exec, spec-draft, interactive, and `mcp tools list` paths (321cef9).
- `zero trust --help` prints usage to stdout and exits 0, matching the other subcommands; the unknown-subcommand path still errors to stderr (faf25ad).

I left the concurrent `Trust`/`Untrust` nitpick as-is: the store write is atomic (temp file + rename) so it can't corrupt, and the only exposure is a lost update if two `zero trust` invocations race, which doesn't arise for a single-user CLI. Happy to add a lockfile if scripted trust ever becomes a use case.

The same push also fixes the macOS and Windows Smoke jobs. Those were red on this branch's own trust tests making non-portable assumptions (the darwin config dir, a `/bin/sh` hook, and an 8.3 short-name path compare), not on a product bug (5976a17).

jatmn (changes requested):

## Findings

- [P2] Link the approved parent issue before this can be reviewed for merge
  `CONTRIBUTING.md:45`
  The contribution guide requires community pull requests to link an existing issue that already has the `issue-approved` label, but this PR does not link any parent issue. Because this account is showing as a first-time contributor, please add the approved issue link to the PR description, or get explicit maintainer confirmation that this PR is part of the internal development cycle and exempt from the issue-first requirement.

- [P2] Surface the trust notice from `zero mcp check`
  `internal/cli/mcp_config.go:271`
  `zero mcp check` is now treated as an MCP spawn site and correctly resolves config with `ExcludeProject=true` in an untrusted workspace, but this path drops the `trustCheckErrored` result and never emits the `trustSkip` notice. If a cloned repo only defines `proj` in `./.zero/config.json`, `zero mcp check proj` now returns `MCP server "proj" is not configured` instead of the PR's promised one-line `zero trust` notice, so users cannot tell that the server was intentionally skipped by workspace trust. This also hides trust-store read errors that the other MCP spawn paths now report. Please route this command through the same skip-reporting path as `mcp tools list`/exec, or otherwise emit the combined trust notice before returning the missing-server error.

- [P2] Gate project-defined OAuth MCP login before starting the flow
  `internal/cli/mcp_oauth.go:209`
  `zero mcp oauth login <server>` still resolves MCP config with `excludeProject=false`, so an untrusted checkout can define an OAuth MCP server in `./.zero/config.json` and this command will load that server, discover/register against its configured URLs, print the authorization URL, wait on a loopback callback, exchange the code, and save the resulting token. That is still a project-scoped MCP server being loaded from an untrusted workspace, even though it is HTTP/OAuth instead of stdio, and it contradicts the PR's trust model that project-scoped MCP servers are dropped until `zero trust` is run. Please apply the same workspace-trust exclusion to OAuth login resolution, or otherwise make the user explicitly trust the workspace before any project-defined MCP OAuth flow can start.

Vasanthdev2004 (approved):
Went through this properly. It's a real gap and a clean fix — project hooks, plugins, and MCP servers from an untrusted clone currently run code outside the sandbox on the first tool call with no prompt, and this gates all three behind an explicit per-workspace trust decision that fails closed. The normalization is symlink-safe, the write is atomic and 0600, and trustRoot is captured before the `--worktree` reassignment. The ~1k added lines are mostly genuine tests, including an agent-loop e2e, not filler.

On process: a community PR needs an approved issue behind it, so I opened and approved #540 for this — please add `Fixes #540` to the description so it links.

Two small things worth folding in before it merges, neither a blocker:
- Two concurrent `zero trust` / `trust remove` runs can clobber each other's write — there's no lock around the read-modify-write. A file lock or an accept-and-retry would close it.
- Add one test where the workspace root and the trust root differ (the `--worktree` split), since that path isn't covered yet.

Approving.

Vasanthdev2004 (changes requested):
Came back to this after jatmn's pass — they're right, and their two findings are the real blockers, above the nits I mentioned. The trust gate misses two project-MCP load sites:

- `zero mcp oauth login` still resolves project config with `excludeProject=false` (mcp_oauth.go:209), so an untrusted checkout's OAuth MCP server still gets loaded and can run the whole discover/authorize/exchange flow — the same hole the stdio path closes.
- `zero mcp check` (mcp_config.go:271) drops the trust notice, so a project server skipped by the gate reads as "not configured" instead of telling the user to run `zero trust`.

Both should be closed before merge — the gate needs to cover every project-MCP entry point, not just tool-list/exec. On the process side: I opened and approved the parent issue (#540), so add `Fixes #540` to link it. The concurrency lock and the workspaceRoot != trustRoot test are still worth folding in, but they're minor next to those two. Retracting my approval until the load-site gaps are closed.

Vasanthdev2004 (approved):
I reviewed the workspace-trust gate. This closes a real remote-code-execution hole: a cloned repo could ship a .zero/hooks.json (or equivalent config) that runs arbitrary commands on first open, and before this change we'd honor it without asking. The gate prompts before any untrusted project config takes effect and fails closed when it can't determine trust — so a malformed/ambiguous config can't silently fall through to execution. The deny path is the default, which is the right posture for a security boundary. Strong approve.

This PR is conflicting with main, so please rebase onto latest main. Heads up that main also has an unrelated test-build break right now (PR #589 is the one-line fix, queued for merge); once #589 lands and you rebase, Smoke should go green. I'd like to see this one merged soon after the rebase — it's the highest-value PR in the queue.

beardthelion:
@jatmn ready for another look. Both code findings are fixed and the issue is linked, in `41ab2dd`.

**OAuth login gate** (`internal/cli/mcp_oauth.go`). `resolveOAuthServer` now resolves workspace trust and threads `excludeProject` into `resolveMCPConfig`, so a project-only OAuth server from an untrusted checkout is dropped before login. The command refuses with "not configured" and prints the trust notice instead of running discovery, client registration, and token exchange against the configured URL. The comment that argued for leaving it ungated is gone. A test points the project server at a live server and asserts zero hits and no token stored when untrusted; a trusted workspace still completes login and persists the token.

**mcp check notice** (`internal/cli/mcp_config.go`). The path already gated the project layer but discarded the trust-check result. It now emits the same one-line notice on the not-found branch, so a gated project server reads as "run zero trust" rather than a bare "not configured", and stays silent when nothing was excluded.

**Issue link.** Added `Fixes #540` to the description.

I also walked the rest of the MCP entry points to confirm this was the last gap: the other ungated `resolveMCPConfig(cwd, false)` sites (backends doctor/snapshot, extensions listing) are enumeration-only, and logout/status touch only the token store. `oauth login` was the last activation site outside the gate.

gnanam1990 (approved):
I checked out `fix/workspace-trust-gate`, merged `origin/main` (no conflicts), and ran the full pipeline:

- `make lint` ✅
- `go test -race -count=1 ./...` ✅
- PR is currently `MERGEABLE` against `main`

This PR addresses a real RCE surface: project-scoped `.zero/` config (hooks, plugins, MCP) running in a cloned repo without a trust prompt. The design is fail-closed and the gate is placed inside each chokepoint rather than hoisted to callers, which is the right shape.

**Reviewer disagreement:** @Vasanthdev2004 approved; @jatmn requested changes. I independently verified the three items jatmn raised:

1. **Linked / approved issue** — PR description says `Fixes #540`, and #540 has the `issue-approved` label.
2. **`zero mcp check` trust notice** — `internal/cli/mcp_config.go:282` now emits `emitTrustNotice` when a project MCP server is gated out, so the user sees the `zero trust` prompt instead of a bare "not configured" error.
3. **`zero mcp oauth login` trust gate** — `internal/cli/mcp_oauth.go:220` now resolves `excludeProject` via `resolveTrust(cwd)` before loading the workspace MCP config, so an untrusted workspace cannot start an OAuth flow against a project-defined MCP server.

All three are resolved in the current branch.

**What I checked beyond the disagreement:**
- `internal/workspacetrust/trust.go` uses exact-path matching, atomic writes, and `0o700`/`0o600` permissions; stored entries are not re-resolved, so a retargeted symlink cannot forge a match.
- The trust gate is applied consistently across hooks (`hook_dispatch.go`), plugins (`plugin_activate.go`), and MCP (`mcp_tools.go`, `mcp_config.go`, `mcp_oauth.go`, `exec_spec.go`, `app.go`).
- `internal/config/resolver.go` drops the project MCP layer when `ExcludeProject` is set.
- `internal/hooks/hooks.go` and `internal/plugins/plugins.go` honor `ExcludeProject`.
- New package `internal/workspacetrust` is justified as a shared cross-cutting concern used by hooks, plugins, MCP, and CLI.

**Verdict: Approve.** The security posture is correct, the disagreement items are resolved, and tests pass. The only remaining merge blocker is the `CHANGES_REQUESTED` state from @jatmn, which needs a re-review or dismissal before GitHub will allow merge.

jatmn (approved):
LGTM

kevincodex1:
bro there is failing test please check @Vasanthdev2004

PR #532, opened 2026-07-06
perf(agent): cheap compaction — dedicated summarizer model, input cap…

kevincodex1:
## Summary

Compaction summarization was the single most expensive recurring event in a
long run: the full elided middle — verbatim tool bodies included — went to
the session's main model at full price. Near a 200k window that is a
~100k-token input call, recursively multiplied when the middle itself
overflows the summarizer. This PR makes compaction cheap on four independent
axes, each safe on its own.

### 1. Dedicated summarizer model

Compaction summaries route to a cheap model instead of the session's main
model. Resolution order (`providers.CompactionModelID`):

1. `ZERO_COMPACTION_MODEL` env
2. `preferences.compactionModel` in config
3. Curated default — **official endpoints only**: Anthropic → Claude Haiku
   4.5, Google → Gemini 2.5 Flash-Lite, OpenAI → GPT-4.1 mini. Custom /
   `*-compatible` endpoints get **no default** (their catalogs are
   unknowable; a guaranteed failed call per run is worse than main-model
   prices), so they opt in via config only.

The literal value `main` at any level forces the old behavior. The provider
is built **lazily** on the first paid compaction (chat-only sessions never
build it), and **any failure — factory or call — permanently falls back to
the main provider for the rest of the run**, so a misconfigured summarizer
can never break or repeatedly delay compaction. When the session already
runs on the curated cheap model, no dedicated summarizer is created.

Wired in exec (`internal/cli/summarizer.go`) and the TUI (resolved against
the **active** profile at run start, so it tracks mid-session model
switches).

### 2. Summarizer input caps

`renderTranscript` now clamps each tool-result body to 2,000 chars and each
tool-call argument blob (a `write_file` call carries the whole file) to 500,
head+tail with a rune-safe split and an explicit omission marker. The
summary needs what was done and decided, not raw payloads — the model that
acted on the full output already turned it into decisions the surrounding
messages record. Cuts summarizer input 5–10× on tool-heavy sessions, and
compounds with (1): cheap model × small input.

### 3. Free reactive prune

`recover()` (the context-limit-error path) now runs the same
prune-stale-tool-output stage the proactive path already had: the first
context-limit error is answered by pruning old tool bodies at **zero cost**,
without consuming the one-shot reactive budget. The paid summarizer only
runs when a later error finds nothing left to prune.

### 4. Context-window re-derivation on model switch

New `Options.ContextWindowFor` (wired to the model registry in exec and the
TUI) lets the loop re-derive the compaction threshold after an
`escalate_model` switch — closing the documented KNOWN LIMITATION where a
switch to a smaller-window model could overflow the target and a larger one
over-compacted. The low-water mark resets on window change so a smaller
window can't be suppressed by stale bookkeeping.

## Deliberately not in this PR

- **Trigger raise (0.7 → 0.8) and background/anticipatory compaction with a
  visible "compacting…" notice** — next PR; they belong together (the
  background path is the validation the `compactionTriggerRatio` comment
  asks for before raising).
- **Summarizer output-token cap** — needs `CompletionRequest.MaxTokens`
  plumbed through all three providers; follow-up.
- **exec spec-draft wiring** — draft runs are short planning sessions; the
  main provider is fine there.

## Checklist

      No linked issue — core-team change, same waiver as #506/#518.
      Build and vet clean. Full suite passes except the 6 pre-existing
      `internal/cli` failures from the local `opengateway` provider profile
      ("openai provider opengateway requires official baseURL") —
      byte-identical failure set on `main`, unrelated to this change.
- [x] `gofmt` clean.
      New: summarizer routing (cheap provider gets the call, main gets none,
      factory lazy+memoized), sticky fallback on call failure AND factory
      failure, transcript clamping (bodies + args, head/tail survival),
      reactive prune-before-pay with budget semantics across two recovers,
      `updateWindow` unit test, end-to-end escalation window re-derivation
      observed via `OnContext`, and `CompactionModelID` resolution-order /
      curated-default / already-cheap tests. `internal/agent` and
      `internal/providers` pass under `-race`.
      N/A — no visual changes (a config preference and wiring only).

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a dedicated compaction and summarization model, with automatic provider defaults and environment overrides.
  * Context limits now recalculate when the active model changes during a run.

* **Bug Fixes**
  * Improved recovery from context-limit situations by pruning excess tool output before summarizing.
  * Added fallback to the main model when the summarizer is unavailable or fails.
  * Reduced summarization payload sizes while preserving essential content from large tool calls and results.
  * Recovery now retries without summarization when pruning alone resolves the issue.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -492,27 +492,29 @@ func (state *compactionState) switchModel(modelID string, window int) {
 	state.summarizerBroken = false
 }
 
-// summarizeProvider returns the provider summarization calls should use: the
-// dedicated summarizer when one applies to the current main model and is
-// healthy, else the main provider. The factory runs at most once per main
-// model; a build error marks the summarizer broken for that model.
-func (state *compactionState) summarizeProvider(ctx context.Context, main Provider) Provider {
+// summarizeProvider returns the provider summarization calls should use and
+// whether it is the dedicated summarizer: that one when it applies to the
+// current main model and is healthy, else the main provider. The factory runs
+// at most once per main model; a build error marks the summarizer broken for
+// that model. Callers branch on the boolean rather than comparing Provider
+// values — interface comparison panics on a non-comparable dynamic type.
+func (state *compactionState) summarizeProvider(ctx context.Context, main Provider) (Provider, bool) {
 	if state.summarizerFactory == nil || state.summarizerBroken {
-		return main
+		return main, false
 	}
 	if !state.summarizerResolved {
 		state.summarizerResolved = true
 		built, err := state.summarizerFactory(ctx, state.mainModel)
 		if err != nil {
 			state.summarizerBroken = true
-			return main
+			return main, false
 		}
 		state.summarizerProvider = built
 	}
 	if state.summarizerProvider == nil {
-		return main
+		return main, false
 	}
-	return state.summarizerProvider
+	return state.summarizerProvider, true
 }
 
 // fitsWindow reports whether a calibrated size is safely under the limit the
@@ -696,9 +698,9 @@ func (state *compactionState) summarizeClosure(ctx context.Context, provider Pro
 		state.planner = newContextPlanner(contextPlannerConfig{})
 	}
 	return func(toSummarize []zeroruntime.Message) (string, error) {
-		chosen := state.summarizeProvider(ctx, provider)
+		chosen, dedicated := state.summarizeProvider(ctx, provider)
 		summary, err := summarizeWithPlanner(ctx, chosen, toSummarize, state.onUsage, state.planner, state.trace)
-		if err == nil || chosen == provider {
+		if err == nil || !dedicated {
 			return summary, err
 		}
 		state.summarizerBroken = true
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/Gitlawb/zero/internal/trace"
 	"github.com/Gitlawb/zero/internal/zeroruntime"
@@ -390,6 +392,27 @@ type compactionState struct {
 	// starts at 1.0 and converges via an EMA as each turn reports actual usage, so
 	// later turns compact nearer to real capacity. Zero is treated as 1.0.
 	calibrationRatio float64
+
+	// summarizerFactory lazily builds the dedicated (cheap) summarizer provider
+	// (Options.Summarizer) for the current main model. summarizerProvider
+	// memoizes the built provider and summarizerResolved records that the
+	// factory ran (it may legitimately return no provider: the main model is
+	// already the cheap one). summarizerBroken is the sticky fallback: once a
+	// build or call fails, the run stays on the main provider until the next
+	// model switch — a misconfigured cheap model must never break (or
+	// repeatedly delay) compaction.
+	summarizerFactory  func(ctx context.Context, mainModelID string) (Provider, error)
+	summarizerProvider Provider
+	summarizerResolved bool
+	summarizerBroken   bool
+	// window is the context window the threshold was derived from. The
+	// reactive free-prune stage compares against it (the limit the provider
+	// enforces), not against the 70% proactive threshold — a pruned history
+	// between the two would still succeed on retry.
+	window int
+	// mainModel is the model the run currently streams on; the summarizer
+	// choice is made against it and remade after a mid-run switch.
+	mainModel string
 }
 
 // calibrate folds one turn's (rawEstimate, actualPromptTokens) sample into the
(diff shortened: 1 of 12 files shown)


jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's request to keep insufficient prune from consuming the retry
  `internal/agent/compaction.go:524`
  CodeRabbit's unresolved review item is still valid: `recover` returns `retried=true` as soon as `pruneStaleToolOutput` reclaims any bytes, without checking whether the pruned request is now under the compaction threshold. The caller then retries that same turn once; if the reclaim was too small, the retry hits the context limit again and the paid `Compact` fallback in this recover call is never reached, even though summarization could have saved the turn. The same block is also behind the `reactiveAttempted` guard, so once a paid reactive compaction has happened, later context-limit errors skip the supposedly free prune stage entirely. Please complete the review request by moving the free prune before the paid-budget guard and only short-circuiting when the pruned request actually fits; otherwise fall through to summarization in the same call.

- [P2] Keep cheap summarization after escalating from a cheap model
  `internal/providers/compaction_model.go:75`
  When the run starts on the curated cheap model, `CompactionModelID` returns an empty model ID and `summarizerFactory` installs no dedicated summarizer. If `--allow-escalation` then switches the run from `claude-haiku-4.5` to its catalog upgrade target `claude-sonnet-4.5`, `agent.Run` updates the main provider and context window but never revisits the summarizer factory. Any later compaction in that now-expensive run summarizes on Sonnet instead of staying on Haiku, which loses the core cost-saving behavior exactly in the cheap-start/strong-escalation workflow this PR enables. Please refresh the summarizer choice after a successful model switch, or avoid disabling the dedicated summarizer for runs that can escalate away from the cheap model.

Vasanthdev2004 (changes requested):
I read through this and jatmn's two findings both hold up against the code — confirming with specifics so they can be addressed in one pass.

P1 (compaction.go:516): confirmed, with one nuance. The cheap-prune stage short-circuits on `reclaimed > 0`, not on "pruned request now fits under threshold." So a too-small prune returns `retried=true`, the caller retries the same turn, the retry overflows again, and only the next recover call falls through to the paid Compact (because the second prune reclaims nothing). Summarization is reached eventually, so "never reached" overstates it — the real cost is one wasted failed model round-trip per too-small prune. jatmn's fix (only short-circuit when the pruned request actually fits, else fall through to Compact in the same call) kills that wasted round and is the right shape. On the reactiveAttempted part: the free prune IS gated behind `if state.reactiveAttempted` at the top of recover, so once a paid compaction has fired, a later context-limit error skips the free prune entirely. Since prune doesn't consume the budget, running it before that guard (and gating only the paid Compact path behind reactiveAttempted) would give one free retry that can't loop — worth doing alongside the threshold check.

P2 (compaction_model.go + cli/summarizer.go): confirmed exactly. `summarizerFactory` returns nil when `CompactionModelID` is "" — which is what happens when the run starts on the cheap model (TestCompactionModelIDSkipsWhenMainIsAlreadyCheap). So `options.Summarizer` is nil, `summarizeProvider` falls back to main, and at startup main IS the cheap model — fine. But after `escalate_model` switches haiku→sonnet, `updateWindow` (loop.go) re-derives the threshold but leaves `summarizerFactory` nil, so `summarizeProvider` now returns the new main (sonnet). The run summarizes on Sonnet from then on, which is the opposite of the PR's point. The cheap model is now strictly cheaper than main, so post-escalation is exactly when a dedicated summarizer should be installed. Either rebuild the factory after a successful switch (`CompactionModelID(sonnet)` returns haiku, so the builder is now non-nil) or don't disable the summarizer up front for runs that can escalate. The escalation test added here only asserts the window is re-derived to 400000 — it doesn't cover the summarizer path, so P2 ships with no test catching it.

The clamp-for-summary work (summaryToolResultClamp / summaryToolArgsClamp) is sound — it only touches the summarizer transcript, not the live messages, so there's no session data loss. I like the overall direction; the cost regression in P2 is the one I'd want fixed before merge since it hits the exact cheap-start/strong-escalation workflow this PR opens up.

Vasanthdev2004:
Read through this. The design is sound: the summarizer is built lazily (only on the first paid compaction), memoized, and a build or call failure sticks for the run with a one-shot retry on the main provider — so a misconfigured cheap model can't break or repeatedly stall compaction. The free prune-first stage in recover() not consuming the reactive budget is the right call, and routing the post-escalate context window through ContextWindowFor fixes the stale-window behavior the loop used to carry (the old comment flagged it as a deferred limitation). The clamping with head+tail and an explicit omission marker, snapped to rune boundaries, is a nice touch.

One observation, not a blocker: the curated defaults are pinned to specific snapshots — claude-haiku-4-5-20251001, gemini-2.5-flash-lite, gpt-4.1-mini. The fail-back makes a stale/wrong default non-fatal, but these IDs will age out as providers deprecate them, so they'll want a periodic refresh (or a resolution path through the model catalog). Also good that custom/compatible endpoints get no default — a guaranteed failed call there would be worse than main-model prices.

Heads up this is conflicting with main right now; main has an unrelated test-build break (PR #589 is the one-line fix, queued for merge), so once that lands a rebase should clear it.

euxaristia:
One question about the free prune stage in `compactionState.recover`, on `fed37e58`.

When `pruneStaleToolOutput` reclaims anything, `recover` returns `(pruned, true)` unconditionally. If the pruned history is still over the provider's limit, that retry comes back with the same context-limit error, and only then does the next `recover` pay for summarization. In that case the run spends one provider round-trip that cannot succeed before it gets to the summarizer.

The obvious tightening is to take the free retry only when the pruned history actually fits, but I do not think `state.threshold` is the right thing to compare against. It is `compactionThreshold(window)`, so 70% of the context window rather than the limit the provider enforces. Gating on it would send a history pruned to, say, 85% of the window off to the summarizer even though the free retry would have gone through. `compactionState` keeps only the derived threshold and not the window, so a correct gate would need the window carried through as well.

So: is the unconditional retry deliberate, on the grounds that one wasted round-trip is cheaper than a summarization that might not have been needed? That reading makes sense to me, and if it is right the tradeoff seems worth a line in the comment above the prune block. `TestRecoverPrunesBeforeSummarizing` pins the behavior ("free prune stage must not call the summarizer") but not the reason for it, and the distinction between the 70% trigger point and the real limit is what makes the obvious optimization wrong here.

Not blocking, and no patch attached, since the only version I tried gated on `state.threshold` and regresses exactly the 70% to 100% band described above.

PR #533, opened 2026-07-06
fix(agent): keep tools exposed for max-turn finalization

pengdst:
## Summary
- keep tool definitions exposed during max-turn finalization requests
- fix a finding where, after the agent runs for N turns mid-session, previously available edit/exec tools could suddenly be reported as not available/exposed
- prevent strict providers from rejecting replayed tool-call history as unavailable tools during the final-answer follow-up

## Why
The issue appeared after longer sessions hit the max-turn path: Zero replayed prior tool-call history for finalization but did not include the active tool definitions on that request. Providers that validate tool history strictly can then fail with errors like `tool edit/exec not exposed in this session`, even though those tools worked earlier in the same session.

## Tests
- go test ./internal/agent

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the final response flow when the agent reaches its turn limit, so the last request now keeps the same available tools as earlier turns.
  * This helps preserve tool access consistency during finalization and supports more reliable provider replay.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -721,7 +721,8 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			Content: nudge,
 		})
 	}
-	if answer, finalMessages, finishReason := finalAnswerAfterMaxTurns(ctx, provider, messages, options); strings.TrimSpace(answer) != "" {
+	finalExposed, _ := partitionToolsCached(registry, permissionMode, options, loaded, toolDefCache)
+	if answer, finalMessages, finishReason := finalAnswerAfterMaxTurns(ctx, provider, messages, finalExposed, options); strings.TrimSpace(answer) != "" {
 		result.FinalAnswer = answer
 		result.FinishReason = finishReason
 		result.Messages = copyMessages(finalMessages)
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

Vasanthdev2004 (approved):
Traced this through and it holds up. After we hit the tool-turn limit, `finalAnswerAfterMaxTurns` replays the whole message history (which contains `tool_use`/`tool_result` blocks) but sent it with no `tools` on the request. Strict providers validate that history against the declared tool set, so they'd reject the finalization call with "tool edit/exec not exposed" — meaning on those providers the real summary silently never lands and you fall through to the generic max-turns fallback. So the bug is real and provider-specific, which matches the report.

The fix is the right shape: it reuses `partitionToolsCached` to reproduce exactly the tool set the earlier turns advertised and passes that on the final request, so replay validation lines up. It doesn't invite fresh tool calls either — the finalization prompt still says "Do not call tools," and `finalAnswerAfterMaxTurns` only collects text, it never executes a returned call, so the worst case is the same fallback we already have today. Net: strict providers go from always-broken finalization to working, lenient providers are unchanged.

Checked out locally: builds clean, the updated `TestRunRequestsFinalAnswerAfterMaxTurns` and the max-turn tests pass, and full CI is green across all three OSes. Small, targeted, covered by a test that now asserts the tools stay exposed.

LGTM. Approving.

PR #535, opened 2026-07-06
fix(acp): surface session persistence failures

### LINKED ISSUE #471: acp: silent errors in session persistence can lose history

fengjikui:
## Summary
- warn ACP clients when session/load cannot read stored history instead of silently starting from empty context
- warn ACP clients after a prompt when saving the turn fails, while keeping the live in-memory turn usable
- write a prompt turn with one AppendEvents batch so a failed write does not partially persist user/assistant history

Fixes #471

## Validation
- git diff --check
- go test ./internal/acp (not run: local environment does not have go installed; command exits `zsh:1: command not found: go`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Session history loading and saving now surface warnings when storage fails instead of failing silently.
  * Current sessions continue running with in-memory history even if persistence is unavailable.
  * Turn completion still proceeds while notifying the client about history save issues.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"log"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):
lgtm

PR #543, opened 2026-07-06
fix(tools): platform-specific pager suggestions, quote/caret-safe cd detection

euxaristia:
## Summary
- The blocked-pager guard (less/more/most) now suggests `type` instead of cat/head/tail on Windows, since those tools don't exist there.
- Drops `more` from the MSYS sandbox suggestion text, since `more` is itself blocked as an interactive pager and would send the model straight into that guard next.
- `detectShellCommandIssue` strips double-quoted spans and cmd.exe caret escapes before matching the bash-style `cd /foo` pattern, so a cd-shaped string inside a quoted argument (e.g. a commit message) isn't mistaken for a real command.

This is independent of #468: it's branched from main and doesn't touch the quoting-guidance text or the MSYS/POSIX detection logic that #476 already owns.

## Test plan
- [x] go build ./...
- [x] go test ./internal/tools/... ./internal/sandbox/...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows command guidance for interactive pagers and shell issues, with clearer alternatives that better fit Windows usage.
  * Reduced false positives when detecting shell syntax issues by handling quoted text and escaped characters more accurately.
  * Updated suggestions for output piping on Windows to be more specific and less confusing.
  * Expanded test coverage for platform-specific command suggestions and detection behavior.

### DIFF
--- a/internal/sandbox/safe_command.go
+++ b/internal/sandbox/safe_command.go
@@ -26,6 +26,10 @@ type interactiveProgram struct {
 	suggestion string
 	// windowsOnly limits the match to GOOS == "windows" (e.g. notepad).
 	windowsOnly bool
+	// windowsSuggestion overrides suggestion on GOOS == "windows", for cases
+	// where the default suggestion names POSIX-only tools (cat/head/tail)
+	// that cmd.exe does not have.
+	windowsSuggestion string
 }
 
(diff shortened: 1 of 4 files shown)


jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):
LGTM

PR #548, opened 2026-07-06
fix: close UX/robustness gaps found in codebase audit

### LINKED ISSUE #546: UX/robustness fixes from the codebase audit (#480): OAuth fallback, UTF-8 truncation, keybinding collisions, Windows rename retry
## Summary

Split out of #480 / #481 per [review feedback on #481](https://github.com/Gitlawb/zero/pull/481#issuecomment-4889555120): grouping the UX/robustness findings from the multi-agent codebase audit separately from the security-sensitive ones so each can be reviewed and landed independently.

This covers 4 of the 9 original findings:

1. **Provider fallback ignores OAuth logins** — `internal/cli/setup.go` (`firstUsableProvider`). Only checked inline API keys/env vars, never OAuth logins, contradicting `setupRequired`'s own logic. If `activeProvider` went stale/blank, a fully-authenticated OAuth-only provider (e.g. `zero auth login xai`) was skipped and the user got forced back into onboarding despite having working credentials.
2. **UTF-8 truncation can corrupt resumed session context** — `internal/sessions/exec_session.go`, `summarizePayload` truncated with a byte-index (`text[:500]`), which can split a multi-byte UTF-8 rune. Feeds directly into exec prompts for `--resume`/`--fork`.
3. **Keybinding collisions silently shadow hardcoded shortcuts** — `internal/tui/model.go`. Dispatch is a priority-ordered switch with configurable bindings interleaved with hardcoded ones and zero collision detection. E.g. configuring `toggleDetailed` to `ctrl+f` permanently killed the "favorite model" shortcut.
4. **No retry on the restore-on-failure rename** — `internal/update/replace_windows.go`. If installing the new binary failed and the restore rename back to the original also failed, `replaceBinary` gave up immediately, risking a permanently missing binary on a transient Windows file lock (antivirus/indexer scanning the just-renamed file).

## Dropped from this split

The ninth original finding — the Windows shell-syntax pre-flight check not being quote-aware (`internal/tools/shell_runtime.go`) — needs no further action here. #476 landed on `main` independently since #481 was opened and already replaced that check with a more thorough segment/word-anchored implementation that resolves the same quoted-text false-positive.

## Fix

PR: (to be linked)

All four fixes ship with regression tests. `go build ./...` and `go test ./internal/cli/... ./internal/sessions/... ./internal/tui/... ./internal/update/...` pass.

## Scope note

Originally reported and fixed together in #481, which is being split into smaller, independently-reviewable PRs per maintainer request.

PierrunoYT:
## Summary

Fixes #546. Split out of #481 per [review feedback](https://github.com/Gitlawb/zero/pull/481#issuecomment-4889555120), grouping the UX/robustness findings from #480 separately from the security-sensitive ones (see #547 for those).

- **cli**: `firstUsableProvider` ignored OAuth logins when picking a fallback provider, forcing unnecessary re-onboarding for an already-authenticated user
- **sessions**: exec prompt summarization truncated on a raw byte offset, which could split a multi-byte UTF-8 rune and embed invalid UTF-8 into a resumed session's prompt
- **tui**: configurable keybindings had no collision detection, so a remapped chord could silently make a hardcoded shortcut (or another configured binding) permanently unreachable
- **update**: `replaceBinary`'s restore-on-failure rename had no retry, risking a permanently missing binary on a transient Windows file lock

Dropped from this split: the ninth original finding (Windows shell-syntax pre-flight check not being quote-aware, `internal/tools/shell_runtime.go`) needs no further action — #476 landed on `main` independently since #481 was opened and already replaced that check with a more thorough segment/word-anchored implementation that resolves the same false-positive.

## Test plan

- [x] `go build ./...` clean
- [x] Regression test added for each fix
- [x] `go test ./internal/cli/... ./internal/sessions/... ./internal/tui/... ./internal/update/...` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding fallback so stored OAuth logins are recognized and won’t trigger repeated prompts when access is already authenticated.
  * Fixed prompt/session text truncation to respect UTF-8 rune boundaries, preventing mid-character cuts.
  * Improved keybinding sanitization to detect shortcut collisions, reset conflicting bindings, and display warning notices consistently.
  * Enhanced Windows executable replacement reliability by retrying renames to reduce failures from transient file locks.

### DIFF
--- a/internal/cli/setup.go
+++ b/internal/cli/setup.go
@@ -351,11 +351,12 @@ func oauthLoggedInProviders() map[string]bool {
 }
 
 // firstUsableProvider returns the saved provider best suited to run without
-// onboarding: the first usable (inline credential present, or no-auth/local)
-// non-local provider, else the first usable local one. It lets the CLI fall back
-// to an already-configured login when the active provider happens to lack a
-// credential, instead of re-running onboarding every launch.
+// onboarding: the first usable (inline credential, stored OAuth login, or
+// no-auth/local) non-local provider, else the first usable local one. It lets
+// the CLI fall back to an already-configured login when the active provider
+// happens to lack a credential, instead of re-running onboarding every launch.
 func firstUsableProvider(providers []config.ProviderProfile) (config.ProviderProfile, bool) {
+	logins := oauthLoggedInProviders()
 	var localFallback config.ProviderProfile
 	haveLocal := false
 	for _, profile := range providers {
@@ -371,7 +372,11 @@ func firstUsableProvider(providers []config.ProviderProfile) (config.ProviderPro
 				continue
 			}
 		}
-		if _, missing := setupMissingCredentialEnv(profile); missing {
+		// A stored OAuth login (e.g. `zero auth login xai`) is a credential too, even
+		// when the profile has no inline key / env var — mirrors setupRequired and
+		// usableSavedProviders so this fallback doesn't force onboarding for a
+		// provider the user is already authenticated with.
+		if _, missing := setupMissingCredentialEnv(profile); missing && !providerHasOAuthLogin(profile, logins) {
 			continue
 		}
 		if providerProfileIsLocal(profile) {
--- a/internal/cli/setup_fallback_test.go
+++ b/internal/cli/setup_fallback_test.go
@@ -1,9 +1,12 @@
 package cli
 
 import (
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/oauth"
 )
 
 func TestFirstUsableProviderPrefersRemoteKeyed(t *testing.T) {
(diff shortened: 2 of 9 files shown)


PierrunoYT:
Addressed in 86125b7:

- **Major (keybindings.go)**: confirmed — `toggleMouse="ctrl+p"` / `togglePlan="ctrl+t"` did end up on the same chord undetected, because togglePlan's mid-pass reversion to its own default (ctrl+p) was never re-checked against toggleMouse, which had already been processed earlier in the same pass. `sanitizeKeyBindings` now loops all three collision checks to a fixed point. Added `TestSanitizeKeyBindingsResolvesChainedDefaultCollision` covering exactly this case.
- **Major (replace_windows.go)**: confirmed — `Apply()` calls `CleanupStaleBinary` unconditionally on every run, so if a previous `replaceBinary` failed at both the install and the restore, the next `Apply()` would delete the only surviving `.old` backup before anyone could recover it. `CleanupStaleBinary` now no-ops when `targetPath` doesn't exist. Added tests for both branches.
- **Nitpick (symmetric retry)**: applied — the initial install rename now goes through `renameWithRetry` too, not just the restore.
- **Nitpick (renameWithRetry sleeps after final attempt)**: applied — no more sleep after the last (already-failed) attempt, and dropped the redundant `else`.
- **Nitpick (missing install-fails/restore-retry test)**: partially addressed. Added `TestReplaceBinaryRestoresOriginalWhenInstallFails` covering install-fails→restore-succeeds. Skipped the double-failure (both renames fail) branch — reliably forcing two independent `os.Rename` calls to fail in-process on Windows without a flaky low-level locking hack isn't worth it for a nitpick-level ask; happy to revisit if you feel strongly.

`go build ./...` clean, `go vet ./...` clean, full package tests pass on Windows.

PR #549, opened 2026-07-06
feat: voice dictation (speech-to-text)

anandh8x:
## Speech-to-text dictation

Adds voice dictation to Zero. Run `/voice` to enter hold-to-record mode, then hold **Space** to dictate and release to transcribe (run `/voice` again to turn it off so Space types normally). Pick a backend with `/stt-model`.

### Backends
- **Local (sherpa-onnx)** — offline, no API key. Auto-downloads a model via `/stt-model`; the engine binary is fetched once and shared across models. Family auto-detection covers the mainstream offline families (Whisper, transducer/Zipformer, Moonshine, Paraformer, SenseVoice, and the CTC families FireRedASR, NeMo, TeleSpeech, WeNet, Zipformer-CTC, Dolphin, Omnilingual — plus the encoder/decoder families Canary and FireRedASR-AED).
- **Cloud batch** — Groq and OpenAI Whisper. Keys resolve from a configured provider, the encrypted credential store, or the env var; a missing/invalid key is entered inline.
- **Streaming** — a warm local sherpa-onnx websocket server (live transcript as you speak) on desktop, or Deepgram / OpenAI Realtime.

### Highlights
- Live audio-reactive waveform while recording (real mic levels when streaming; a synthesized wave for batch).
- `/stt-model` download chooser: recommended models first, Live/Batch tag per row, size + download state, and a search filter.
- Streaming transcript is inserted live into the composer and committed on stop; batch inserts on stop. Insert-for-review by default (`stt.autoSubmit` to fire straight at the agent).
- The warm streaming server is released when `/voice` is turned off and on exit.

### Implementation notes
- No CGO and no bundled runtimes — every OS integration shells out via `exec.Command` with discrete argv (never a shell string).
- The streaming server manager mirrors the LSP manager's warm-subprocess lifecycle.
- OS-boundary code (recorders, transcribers, server) is unit-tested via injected fakes; real microphone / websocket behavior is covered by env-gated smoke tests.

### Testing
- `go build ./...`, `go vet ./...`, and the config / dictation / cli / tui suites pass.
- Verified end-to-end against real downloaded binaries + models (offline batch, streaming websocket server, and the model auto-download), including the FireRedASR-CTC and NeMo streaming transducer families.

## Summary by CodeRabbit

* **New Features**
  * Added speech-to-text dictation with both batch and streaming modes, supporting local (sherpa-onnx) and cloud providers, including WAV/M4A detection.
[...]

PR #557, opened 2026-07-06
feat: voice dictation (speech-to-text)

anandh8x:
## Speech-to-text dictation

Adds voice dictation to Zero. Run `/voice` to enter hold-to-record mode, then hold **Space** to dictate and release to transcribe (run `/voice` again to turn it off so Space types normally). Pick a backend with `/stt-model`.

### Backends
- **Local (sherpa-onnx)** — offline, no API key. Auto-downloads a model via `/stt-model`; the engine binary is fetched once and shared across models. Family auto-detection covers the mainstream offline families (Whisper, transducer/Zipformer, Moonshine, Paraformer, SenseVoice, and the CTC families FireRedASR, NeMo, TeleSpeech, WeNet, Zipformer-CTC, Dolphin, Omnilingual — plus the encoder/decoder families Canary and FireRedASR-AED).
- **Cloud batch** — Groq and OpenAI Whisper. Keys resolve from a configured provider, the encrypted credential store, or the env var; a missing/invalid key is entered inline.
- **Streaming** — a warm local sherpa-onnx websocket server (live transcript as you speak) on desktop, or Deepgram / OpenAI Realtime.

### Highlights
- Live audio-reactive waveform while recording (real mic levels when streaming; a synthesized wave for batch).
- `/stt-model` download chooser: recommended models first, Live/Batch tag per row, size + download state, and a search filter.
- Streaming transcript is inserted live into the composer and committed on stop; batch inserts on stop. Insert-for-review by default (`stt.autoSubmit` to fire straight at the agent).
- The warm streaming server is released when `/voice` is turned off and on exit.

### Implementation notes
- No CGO and no bundled runtimes — every OS integration shells out via `exec.Command` with discrete argv (never a shell string).
- The streaming server manager mirrors the LSP manager's warm-subprocess lifecycle.
- OS-boundary code (recorders, transcribers, server) is unit-tested via injected fakes; real microphone / websocket behavior is covered by env-gated smoke tests.

### Testing
- `go build ./...`, `go vet ./...`, and the config / dictation / cli / tui suites pass.
- Verified end-to-end against real downloaded binaries + models (offline batch, streaming websocket server, and the model auto-download), including the FireRedASR-CTC and NeMo streaming transducer families.

## Summary by CodeRabbit

* **New Features**
  * Added speech-to-text dictation with local offline transcription, cloud batch transcription, and cloud streaming options.
  * Introduced an in-TUI STT model picker with local (sherpa-onnx) download, deterministic progress/status, and installed-state awareness.
  * Added voice mode for Space press/hold recording plus an inline cloud API-key prompt with saved-key persistence.
* **Bug Fixes**
  * Strengthened STT configuration validation and merging so invalid providers/stream providers and out-of-range settings are rejected early with clearer guidance.
  * Improved dictation streaming partial/region handling and safer STT server warm-up/shutdown behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/charmbracelet/colorprofile v0.4.3
 	github.com/charmbracelet/x/ansi v0.11.7
 	github.com/charmbracelet/x/term v0.2.2
+	github.com/coder/websocket v1.8.15
 	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
 	golang.org/x/sys v0.46.0
 	mvdan.cc/sh/v3 v3.13.1
--- a/go.sum
+++ b/go.sum
@@ -32,6 +32,8 @@ github.com/clipperhouse/displaywidth v0.11.0 h1:lBc6kY44VFw+TDx4I8opi/EtL9m20WSE
 github.com/clipperhouse/displaywidth v0.11.0/go.mod h1:bkrFNkf81G8HyVqmKGxsPufD3JhNl3dSqnGhOoSD/o0=
 github.com/clipperhouse/uax29/v2 v2.7.0 h1:+gs4oBZ2gPfVrKPthwbMzWZDaAFPGYK72F0NJv2v7Vk=
 github.com/clipperhouse/uax29/v2 v2.7.0/go.mod h1:EFJ2TJMRUaplDxHKj1qAEhCtQPW2tJSwu5BF98AuoVM=
+github.com/coder/websocket v1.8.15 h1:6B2JPeOGlpff2Uz6vOEH1Vzpi0iUz20A+lPVhPHtNUA=
+github.com/coder/websocket v1.8.15/go.mod h1:NX3SzP+inril6yawo5CQXx8+fk145lPDC6pumgx0mVg=
 github.com/dlclark/regexp2/v2 v2.2.1 h1:mf4KkFUj0gJuarK8P+LgiS+Lit7m9N1yAwEfPbee7R0=
 github.com/dlclark/regexp2/v2 v2.2.1/go.mod h1:avUrQvPaLz2DrFNHJF0taWAFFX2C1GMSSoeiqFjcBmU=
 github.com/go-quicktest/qt v1.102.0 h1:HSQxCeh5YZH3EL3W39ixjtyaEhcWSXQHtHnMBzSs474=
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -729,6 +729,16 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		Scope:         scope,
 	})
 	lastKnownMCPConfig := mcpConfig
+	sttServerManager := newDictationServerManager(resolved.STT)
+	// Keep STT downloads in the SAME config tree the rest of the TUI uses. Deriving
+	// from userConfigPath (rather than config.UserConfigDir()) matters when the config
+	// root is overridden — e.g. in tests or a custom ZERO config dir — so the two
+	// don't diverge. userConfigPath points at .../zero/config.json, so its dir is the
+	// zero config dir.
+	sttDownloadRoot := ""
+	if userConfigPath != "" {
+		sttDownloadRoot = filepath.Join(filepath.Dir(userConfigPath), "stt")
+	}
 	return deps.runTUI(context.Background(), tui.Options{
 		Cwd:                  workspaceRoot,
 		Version:              version,
@@ -792,9 +802,15 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 			merged, _ := plugins.MergedSkillsLoaded(deps.skillsDir(), pluginActivation.skillRoots)
 			return merged
 		}),
-		PermissionMode: permissionMode,
-		Notify:         resolved.Notify,
-		KeyBindings:    resolved.KeyBindings,
+		PermissionMode:            permissionMode,
+		Notify:                    resolved.Notify,
+		KeyBindings:               resolved.KeyBindings,
+		STT:                       resolved.STT,
+		BuildDictationTranscriber: newDictationTranscriberFactory(resolved, userConfigPath, sttServerManager),
+		ShutdownDictationServer:   sttServerManager.Shutdown,
+		STTDownloadRoot:           sttDownloadRoot,
+		STTKeyStatus:              newSTTKeyStatus(resolved, userConfigPath),
+		SaveSTTKey:                newSaveSTTKey(userConfigPath),
 		Setup: tui.SetupOptions{
 			Visible:    setupVisible,
 			Required:   needsSetup,
--- a/internal/cli/dictation.go
+++ b/internal/cli/dictation.go
@@ -1,0 +1,248 @@
+package cli
+
+import (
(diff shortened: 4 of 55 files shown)


Vasanthdev2004 (dismissed):
This is a +8,283 LoC feature — a whole voice-dictation subsystem: local whisper engine auto-download with a model picker, a localhost streaming server subprocess, four transcription backends (local sherpa-onnx, Groq, OpenAI, Deepgram), audio capture, and TUI keybinding/mouse wiring. It's the third attempt (#549 and #553 were closed), and there's no approved issue behind it. That's the thing I have to flag before anything else.

The CONTRIBUTING issue-approved gate exists for exactly this case. A flagship feature with an 8k-LoC maintenance surface — multi-backend STT, platform-specific download and file-ownership paths, a child server process, audio capture across Linux/macOS/Windows — should get a yes/no from Kevin on whether Zero takes it on as a maintained feature before the implementation gets line-reviewed. I'll grant that gate on a clear bug fix, but a feature this size is the maintainer's call, and the path forward is an issue, not a fourth revision of the PR. Open one, lay out the backends and the ongoing maintenance cost, and get it labeled `issue-approved`; I'll review this implementation in full at that point, because it's worth reviewing.

The implementation is genuinely well-built, and I want that on the record so the work isn't wasted. What I checked:

- Supply chain: one new dep, `github.com/coder/websocket` — reputable, minimal footprint.
- Download trust: engine + models verified against SHA256 digests, opt-in behind a confirm prompt, with an explicit comment naming the trust boundary. Curated models are checksum-pinned; browse-listed ones download without a digest but with a warning. Right shape.
- Server: loopback-only (`ws://127.0.0.1:6006`), no `0.0.0.0`, torn down alongside the LSPs on shutdown.
- Cloud transcriber: Bearer auth, redacts the API key in error bodies via `providerio.Redact`, reuses the existing classified-error helper.
- Config: opt-in `STTConfig`, dormant when unset, defaults to local.
- Tests: real unit and smoke tests across download, local, and stream paths, not stubs.
- TUI wiring: clean — adds a `dictationController`, gates on the existing modal condition plus `m.sttKeyPrompt == nil`, and routes paste into the STT key prompt instead of the composer.

So: not slop, and I'm not asking you to throw it out. But I'm not approving it until the feature is approved through an issue and Kevin's signed off on taking on the surface area. Once that's in place, tag me and I'll do the full read.

kevincodex1:
nice feature this is really useful! kindly address some coderabbit comments @anandh8x

Vasanthdev2004 (approved):
great work!

euxaristia:
Hey folks, I took a look at this PR and noticed a few correctness issues with the dictation features:

### 1. Nil Pointer Panic in `recorder.Stop()`
If a recording is started as a streaming capture (via `StartStreaming()`), `r.proc` is not set. When `Stop()` is eventually called, `proc.StopGracefully()` is invoked on a `nil` handle, causing a panic:

```go
// internal/dictation/recorder.go
func (r *recorder) Stop() ([]byte, error) {
    ...
    proc, path := r.proc, r.filePath
    r.proc, r.filePath, r.recording = nil, "", false
    defer os.Remove(path)

    if r.opts.Platform == PlatformTermux {
        ...
    }

    // proc is nil for streaming, causing a panic here
    if err := proc.StopGracefully(); err != nil {
        _ = proc.Kill()
    }
```

**Fix:** Wrap the shutdown logic in a `proc != nil` check:
```go
	if proc != nil {
		if err := proc.StopGracefully(); err != nil {
			_ = proc.Kill()
		}
		_ = waitWithTimeout(proc, stopGrace)
	}
```

---

### 2. Esc Swallows Run-Cancel Confirmation
When a background task is running and you press `Ctrl+C` once to trigger cancel confirmation (*"Press Esc to confirm cancel"*), pressing Esc while dictation is active will cancel dictation but silently swallow the run-cancellation confirmation without actually stopping the task.

This happens because dictation intercepts Esc right after disarming the cancel confirmation:
```go
// internal/tui/model.go
			wasConfirmingCancel := m.pending && m.cancelConfirmActive
			m = m.disarmCancelConfirmation()
			if m.dictation.active() {
				return m.cancelDictation() // Esc handled here; wasConfirmingCancel is swallowed
			}
```

**Fix:** Prioritize `wasConfirmingCancel` at the top of the Esc handler:
```go
			wasConfirmingCancel := m.pending && m.cancelConfirmActive
			if wasConfirmingCancel {
				m = m.disarmCancelConfirmation()
				m.clearComposer()
				m.cancelRun()
				return m, nil
			}
			m = m.disarmCancelConfirmation()
			if m.dictation.active() {
				return m.cancelDictation()
			}
```

---

### 3. Composer Corruption & Cursor Jumping During Typing while Streaming
If you start typing in the composer while streaming dictation is active:
1. **Corruption:** `regionStart` and `regionEnd` are absolute indices that don't adjust when you type *before* the region. The next incoming partial transcript deletes the wrong range, corrupting your text.
2. **Cursor Jumping:** The cursor is hard-reset to the end of the streaming transcript after every partial, jumping it away from where you are typing.

**Fix:**
1. In `Update` (`internal/tui/model.go`), if the composer text changes while dictation is active, check the edit bounds. If the edit is entirely before the region, shift the bounds. If it overlaps/modifies the region, deactivate the region (`regionActive = false`) so the manual changes are kept:
```go
    // Inside Update(msg tea.Msg)
    if dictationActiveBefore {
        switch msg.(type) {
        case sttPartialMsg, dictationTranscribedMsg:
        default:
            newState := nm.currentComposerState()
            if newState.text != oldComposerText || newState.cursor != oldComposerCursor {
                oldRunes := []rune(oldComposerText)
                newRunes := []rune(newState.text)
                oldRegionStart := m.dictation.regionStart
                oldRegionEnd := m.dictation.regionEnd

                diffStart := 0
                for diffStart < len(oldRunes) && diffStart < len(newRunes) && oldRunes[diffStart] == newRunes[diffStart] {
                    diffStart++
                }
                diffEndOld := len(oldRunes)
                diffEndNew := len(newRunes)
                for diffEndOld > diffStart && diffEndNew > diffStart && oldRunes[diffEndOld-1] == newRunes[diffEndNew-1] {
                    diffEndOld--
                    diffEndNew--
                }

                if diffStart >= oldRegionEnd {
                    // Edit is after region
                } else if diffEndOld <= oldRegionStart {
                    shift := (diffEndNew - diffStart) - (diffEndOld - diffStart)
                    nm.dictation.regionStart += shift
                    nm.dictation.regionEnd += shift
                } else {
                    nm.dictation.regionActive = false
                }
            }
        }
    }
```
2. In `applyStreamingText` (`internal/tui/dictation_stream.go`), calculate the change in region length (`delta`), and shift the cursor dynamically relative to the bounds so it doesn't jump back:
```go
    oldCursor := state.cursor
    oldRegionStart := m.dictation.regionStart
    oldRegionEnd := m.dictation.regionEnd
    newRegionLen := len([]rune(rendered))
    delta := newRegionLen - (oldRegionEnd - oldRegionStart)

    var targetCursor int
    if oldCursor <= oldRegionStart {
        targetCursor = oldCursor
    } else if oldCursor >= oldRegionEnd {
        targetCursor = oldCursor + delta
    } else {
        targetCursor = oldRegionStart + newRegionLen
    }
    
    // ... insertComposerText ...
    updated.cursor = targetCursor
    m.setComposerState(updated)
```

kevincodex1 (approved):
lgtm

PR #560, opened 2026-07-06
feat(providers): OAuth login profiles and list-first /provider manager

kevincodex1:
## Problem

Multi-provider setups were half-supported:

1. **OAuth logins vanished.** `zero auth chatgpt` / `zero auth login <preset>` (and the TUI login, if you quit the wizard before the model step) stored the token but never wrote a provider profile — the provider appeared in no list, `zero providers use <name>` failed "not found", and the login's own `--provider` hint failed config resolve. From the user's seat: the login "worked", then their existing provider looked broken and the new one didn't exist.
2. **Switching to an OAuth provider 401'd.** The `/model` switch path inlined the raw bearer as an API key, which suppressed the runtime's OAuth resolver — no token refresh and no Codex `chatgpt-account-id` header.
3. **No way to manage providers in the TUI.** `/provider` jumped straight into the add wizard: no list, no set-active, no edit, no delete. `providers list` also lied about credentials (`api key: not set` for stored-key profiles).

### OAuth logins become first-class providers
- Every OAuth login path (CLI `auth chatgpt`, `auth login <preset>`, TUI wizard + first-run, browser + device-code) now upserts a catalog-backed profile via `config.EnsureCatalogProvider` right after storing the token — **without stealing the active provider** (it only becomes active when nothing was configured). The CLI prints how to switch.
- The TUI switch path keeps OAuth profiles keyless so the proper resolver + login key attach (fixes the Codex 401), and the credential gate recognizes stored OAuth logins.
- `providers list/current` report truthful credential state: `apiKeySet` honors stored keys, and token logins render as `oauth login`. `providers check` accepts an OAuth login as a credential.

### List-first `/provider` manager (TUI)
`/provider` now opens the saved-provider list — the list is the home screen, verbs are keys

- **Enter** activates via the shared switch path (persists active+model, rebuilds the clienPROVIDER`, warms discovery)
- **`a`** add (existing wizard; Esc returns to the list)
- **`e`** field-level editor: endpoint, model, API key (captured into the encrypted store, never cleartext in config.json), description, and **safe rename** — the credential-store entry, `activeProvider`, and the
live session's `ZERO_PROVIDER` all follow the new name
- **`d`** delete with inline confirm: removes the profile + stored key, hands the active pointer off, keeps the OAuth token (re-adding needs no new browser round-trip; `zero auth logout` removes it)
- **Esc** walks back one level (field → menu → list → closed)
- Credential states (`key stored` / `oauth login` / `stored key missing` / `env VAR` / `no ronously** — keychain subprocess reads never block the render loop

Implemented as manager steps on `providerWizardState`, so every existing overlay gate, key it with zero changes at those sites.

### CLI + config parity
- `zero providers remove <name>` — deletes the stored key too, hands off `activeProvider` (first remaining, or cleared)
- `zero providers rename <old> <new>` — migrates the credential-store entry write-new-then-n can duplicate but never lose a key
- New `config.RemoveProvider` / `config.RenameProvider`, read-modify-atomic-write like the

## Tests
- config: `EnsureCatalogProvider` (create/no-op/active rules), `RemoveProvider` (handoff, uameProvider` (active follows, credstore migration, collision rejection)
- tui: 10 manager tests (activate, pending-block, delete confirm/cancel, edit persist, renanavigation, add-from-manager, stale creds ignored) + OAuth switch keyless-profile tests
- cli: login-profile persistence, credential-state display
- Full `tui`/`cli`/`config`/`zerocommands` suites green; verified end-to-end against a real/delete + `remove`/`rename` CLI including active handoff and last-provider removal)

Screenshot 2026-07-06 at 10 04 47 PM

## Summary by CodeRabbit

* **New Features**
  * Enhanced provider management with new **remove**/**rename** actions (including updated help) and a dedicated provider manager UI flow.
  * Login/setup flows now automatically scaffold catalog-backed provider profiles and show tailored post-login guidance.
* **Bug Fixes**
  * OAuth-based, keyless logins are now treated as usable credentials across the CLI, TUI, and diagnostics—provider statuses render correctly and credential gating no longer blocks valid OAuth sessions.
  * Deleting/renaming providers now better preserves **active provider** and keeps stored credentials aligned.
* **Tests**
  * Expanded coverage for OAuth login persistence, provider edit/remove/rename behavior, and credential-state rendering.

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -11,9 +11,45 @@ import (
 
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/oauth"
+	"github.com/Gitlawb/zero/internal/providercatalog"
 	"github.com/Gitlawb/zero/internal/provideroauth"
 	"github.com/Gitlawb/zero/internal/redaction"
 )
+
+// ensureLoginProviderProfile makes a freshly stored OAuth login visible as a
+// provider: without a profile in config.json the login shows up nowhere — not in
+// `zero providers list`, not in the TUI picker — and `zero providers use <name>`
+// fails "not found", so the user's working active provider looks broken while
+// the new login looks lost. Returns user-facing guidance lines ("" when the
+// login has no catalog entry to scaffold from). Best-effort by design: the
+// token is already stored, so a profile-write failure degrades to a warning
+// line instead of failing a login that succeeded.
+func ensureLoginProviderProfile(deps appDeps, provider string) string {
+	if _, ok := providercatalog.Get(provider); !ok {
+		// Custom OAuth server without a catalog entry — no endpoint/model defaults
+		// to scaffold a profile from; the user wires their own profile.
+		return ""
+	}
+	configPath, err := deps.userConfigPath()
+	if err != nil {
+		return "warning: login saved, but no provider profile was written: " + err.Error()
+	}
+	ensured, err := config.EnsureCatalogProvider(configPath, provider)
+	if err != nil {
+		return "warning: login saved, but no provider profile was written: " + err.Error()
+	}
+	active := strings.EqualFold(strings.TrimSpace(ensured.Active), strings.TrimSpace(ensured.Name))
+	switch {
+	case ensured.Created && active:
+		return fmt.Sprintf("Added provider %q to your config and set it active.", ensured.Name)
+	case ensured.Created:
+		return fmt.Sprintf("Added provider %q to your config; the active provider is still %q.\nSwitch with: zero providers use %s", ensured.Name, ensured.Active, ensured.Name)
+	case active:
+		return fmt.Sprintf("Provider %q is configured and active.", ensured.Name)
+	default:
+		return fmt.Sprintf("Provider %q is already configured.\nSwitch with: zero providers use %s", ensured.Name, ensured.Name)
+	}
+}
 
 // runAuth dispatches `zero auth <command>` for provider OAuth login. It is
 // additive and independent of `zero mcp oauth` (MCP server auth), which is
@@ -138,8 +174,10 @@ func runAuthChatGPT(args []string, stdout io.Writer, stderr io.Writer, deps appD
 	if _, err := fmt.Fprint(stdout, statuses); err != nil {
 		return exitCrash
 	}
-	if _, err := fmt.Fprint(stdout, "\nUse it with zero, e.g.:\n  zero --provider chatgpt --model gpt-5.5\n"); err != nil {
-		return exitCrash
+	if line := ensureLoginProviderProfile(deps, "chatgpt"); line != "" {
+		if _, err := fmt.Fprintf(stdout, "\n%s\n", line); err != nil {
+			return exitCrash
+		}
 	}
 	return exitSuccess
 }
@@ -330,6 +368,11 @@ func runAuthLogin(args []string, stdout io.Writer, stderr io.Writer, deps appDep
 	}
 	if _, err := fmt.Fprintf(stdout, "Logged in to %s.\n%s\n", parsed.positional[0], oauth.FormatStatuses([]oauth.Status{status})); err != nil {
 		return exitCrash
+	}
+	if line := ensureLoginProviderProfile(deps, provider); line != "" {
+		if _, err := fmt.Fprintln(stdout, line); err != nil {
+			return exitCrash
+		}
 	}
 	return exitSuccess
 }
--- a/internal/cli/auth_test.go
+++ b/internal/cli/auth_test.go
@@ -12,11 +12,14 @@ import (
 	"github.com/Gitlawb/zero/internal/oauth"
 )
 
-// withAuthStore points the provider OAuth store at a temp file for the test.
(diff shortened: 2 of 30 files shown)


jatmn (changes requested):

## Findings

- [P2] Persist the stored-key marker when editing an existing provider key
  `internal/tui/provider_manager.go:485`
  The new manager editor writes a replacement API key through `SecureProviderProfile`, but then persists it with `UpsertProvider`. For an existing provider that did not already have `apiKeyStored`, `mergeProfile` only copies non-empty string/map fields and never copies `APIKeyStored`, so the secret is written to the credential store while config.json still says the provider has no stored key. Later launches and provider switches gate `ApplyStoredAPIKey` on that marker, so the edited key is ignored and the provider falls back to missing/env credentials. Please make the edit path persist `apiKeyStored: true` for the updated profile, and add a regression that edits a previously env/no-key provider with a new key.

- [P2] Let the startup fallback select OAuth-login providers
  `internal/cli/setup.go:374`
  The active-provider recovery paths call `firstUsableProvider` when `activeProvider` is blank/stale or when the active provider has no usable credential, but this helper still skips any profile for which `setupMissingCredentialEnv` reports a missing key. A profile created by the new OAuth-login persistence is intentionally keyless, so a config with `activeProvider` pointing at a stale provider and a saved `chatgpt`/`xai` OAuth profile will still fall through to onboarding instead of starting with the already-authenticated provider. Please make this fallback use the same OAuth-login check as `setupRequired`/`usableSavedProviders` so saved token profiles are actually usable when the active pointer needs recovery.

- [P2] Delete stored keys from the store beside the config being edited
  `internal/cli/provider_onboarding.go:358`
  `providers remove` deletes the profile from `configPath`, but then calls `config.ForgetProviderKey`, which opens the default credential store instead of the store beside that same `configPath`. The TUI delete path has the same shape with `config.ProviderKeyStore()` in `internal/tui/provider_manager.go:309`, while setup and rename both use `ProviderKeyStoreAt(filepath.Dir(configPath))`. With a non-default user config path, remove/delete report that the provider is gone but leave the encrypted API key behind. Please delete from the same path-aware store used when the key was captured/migrated.

- [P3] Allow the description editor to clear a saved description
  `internal/tui/provider_manager.go:450`
  The provider manager exposes Description as an optional editable field and commits an empty buffer into `editDraft.Description`, but saving routes through `UpsertProvider`, whose merge only applies `Description` when the new value is non-empty. Clearing a provider description in the UI therefore reports success but the old description remains in config and immediately reappears in the list. Please make the edit save path distinguish “clear this field” from “leave it unchanged” for descriptions.

jatmn (changes requested):

## Findings

  `GitHub mergeability`
  GitHub currently reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, so this branch cannot be merged cleanly into `main`. Please rebase on the latest `main` branch and resolve those conflicts before addressing the review findings below, so the fixes apply to the code that will actually be merged.

- [P2] Link an approved parent issue for this contributor PR
  `CONTRIBUTING.md:19`
  This PR is from a `CONTRIBUTOR` account, but the PR body does not link an approved parent issue, and the closest related issue I found (`#375`) is closed without the `issue-approved` label. The contribution policy requires community PRs to be tied to an already reviewed issue with that label before implementation work is opened, otherwise the PR can be closed without review. Please link the approved issue for this scope, or have a maintainer record that this PR is exempt/internal work before continuing the code review cycle.

- [P3] Use the OAuth token key when telling users how to clean up a renamed login
  `internal/tui/provider_manager.go:327`
  The delete flow detects OAuth-backed profiles using both `Name` and `CatalogID`, but the status text always says to remove the token with `zero auth logout <profile name>`. After renaming a catalog OAuth profile, for example `{name:"codex", catalogID:"chatgpt"}`, the runtime still finds the token via the `chatgpt` catalog fallback, while `zero auth logout codex` only deletes `provider:codex` and leaves the real `provider:chatgpt` token behind. Please point the cleanup hint at the actual stored-login candidate that exists, or make logout remove the same candidate set the runtime uses.

jatmn (changes requested):
Thanks for the updates. I rechecked the changed paths and found issues that still need to be addressed.

## Findings

- [P2] Load TUI stored keys from the active config's credential store
  `internal/tui/command_center.go:567`
  The TUI setup/edit paths now persist API keys with `SecureProviderProfile(..., m.userConfigPath)`, and delete/remove correctly use `ProviderKeyStoreAt(filepath.Dir(m.userConfigPath))`, but `profileWithCredential` still reads `config.ProviderKeyStore()` from the default config directory. For a TUI session using a non-default `userConfigPath`, a provider saved with only `apiKeyStored` has its key in the config-adjacent store, so this lookup misses it. `switchProviderModel` then rejects the provider as having no usable credential before the path-aware `newProvider` wrapper can load the key, and the manager credential-status command has the same mismatch at `internal/tui/provider_manager.go:119`. Please resolve TUI stored-key reads through `ProviderKeyStoreAt(filepath.Dir(m.userConfigPath))` so the UI reads from the same store it writes to.

jatmn (approved):
@Vasanthdev2004 LGTM

PR #565, opened 2026-07-06
Stream tool file IO

### LINKED ISSUE #564: Stream tool file IO to reduce memory usage
### Problem

`read_file` and `grep` currently do too much in-memory buffering for large files or broad searches. This can increase RAM use, delay first useful output, and make tool calls heavier than needed.

### Requested behavior

Update tool IO paths to stream file reads and grep scanning while preserving existing sandbox/path guards, output budgets, cancellation behavior, file tracking, and user-visible output format.

### Why it belongs in Zero

Zero tools run often during coding sessions. Lower RAM use and faster bounded output improve reliability on large repos without changing user workflow or permissions behavior.

Users can narrow grep queries or read smaller file ranges, but that does not fix tool-level memory cost. Keeping current full-buffer behavior is simpler, but wastes RAM on common large-repo workflows.

_No response_

GautamBytes:
Closes #564.

Streams `read_file` and `grep` IO. Preserves guards, budgets, cancellation, file tracking.

Benchmem on Apple M4, `-count=5` avg:
- `ReadFileLargeRangeSmall`: 23.2 MB/op -> 18.3 KB/op
- `GrepLargeTreeHeadLimit`: 29.8 MB/op -> 4.27 MB/op

### DIFF
--- a/internal/tools/file_tools_test.go
+++ b/internal/tools/file_tools_test.go
@@ -76,6 +76,45 @@ func TestReadFileToolAppliesByteBudget(t *testing.T) {
 	}
 }
 
+func TestReadFileToolTracksWholeFileHashWhenReadingSmallRange(t *testing.T) {
+	root := t.TempDir()
+	var builder strings.Builder
+	for i := 1; i <= 2000; i++ {
+		builder.WriteString("line ")
+		builder.WriteString(strings.Repeat("x", 16))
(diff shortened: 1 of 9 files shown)


jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):
great feature! thanks for your contribution

PR #568, opened 2026-07-06
feat(tui): remember recent provider+model selections in /model picker

### LINKED ISSUE #562: /model: remember recent provider+model selections
## Problem

The `/model` picker currently shows only the active model under `Recent`.

When a user switches between multiple configured providers, this makes it harder to return to recently used provider/model combinations. This is especially confusing when the same model name exists across multiple providers, for example `deepseek-v4-flash` from OpenCode, OpenCode Go, NVIDIA, or another OpenAI-compatible provider.

Today the user has to search again or scroll through provider sections to find the previous provider/model pair.

## Requested behavior

Show the last few recently used provider+model selections in the `/model` picker.

Suggested behavior:

- keep the newest selections first
- cap the list to 3-5 entries
- store entries as provider-qualified pairs, for example `opencode-go/deepseek-v4-flash`
- de-duplicate by provider+model pair, not model name alone
- selecting a recent item should switch both provider and model
- keep existing favorites behavior unchanged

## Why it belongs in Zero

Zero already has a `Recent` section in `/model`, but it currently contains only the active model. A short automatic history would make model switching faster and less error-prone for users with multiple subscriptions/providers.

## Notes

This is different from favorites: favorites are manual pins, while recents should reflect short-term switching history automatically.

I can work on this if maintainers agree with the scope. Please add `issue-approved` if this is acceptable for a community PR.

PierrunoYT:
## Summary

Closes #562.

The `/model` picker's "Recent" section only ever showed the currently active model. This made it hard to switch back to a previously used provider+model pair once you'd moved on to something else — especially confusing when the same model name exists across multiple providers (the picker's cross-group de-dup keyed on model id alone, so two providers offering the same id would collide).

This adds a short automatic history of provider-qualified selections:

- Newest first, capped to 5 entries (`config.MaxRecentModels`)
- Stored as provider-qualified pairs (`{provider, model}`), not bare model names
- De-duplicated by provider+model pair, not model name alone (fixes a latent cross-provider collision in the picker's row de-dup along the way)
- Selecting an older recent entry switches both provider and model, same as any other cross-provider picker row
- Existing favorites behavior is unchanged
- Persisted to user config only (`preferences.recentModels`), never merged from project config — same posture as `favoriteModels`

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/config/...` — new tests for `SetRecentModels` normalization (order preservation, dedup by provider+model pair, cap to 5), and a resolver test confirming `recentModels` loads from user config only (not merged from project config)
- [x] `go test ./internal/tui/...` — new tests: `Recent` section pins the active model first and shows history past it; picker row de-dup keys on provider+model, not model id alone; `/model <id>` and cross-provider picker switches both record and persist history, moving a re-selected pair back to the front
- [x] `go test ./internal/cli/...`
- Pre-existing, unrelated failure noted: `TestProviderWizardAdvancesProviderAPIKeyAndModelSteps` also fails on `main` at the base commit (rendering-width assertion), not touched by this change.

## Summary by CodeRabbit

* **New Features**
  * Added provider-aware “Recent” model history that records recently used provider/model combinations (newest first, capped at 5) and persists across sessions.
  * The interactive model picker now renders “Recent” with provider context and keeps the history ordering consistent across provider/model switches.
* **Bug Fixes**
  * Recent history now loads correctly from the intended user setting only, trims whitespace, removes invalid entries, and deduplicates by provider+model (case-insensitive), including correct behavior when switching providers.
* **Chores**
  * Updated ignore rules to exclude Go build and module cache directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,10 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 .cache
 *.tsbuildinfo
 
+# Go build/module caches (GOCACHE/GOMODCACHE pointed here for local runs)
+.gocache/
+.gomodcache/
+
 # IntelliJ based IDEs
 .idea
 
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -742,6 +742,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		ProviderProfile:      resolved.Provider,
 		SavedProviders:       usableSavedProviders(resolved.Providers),
 		FavoriteModels:       resolved.Preferences.FavoriteModels,
+		RecentModels:         resolved.Preferences.RecentModels,
 		RecapsEnabled:        resolved.Preferences.RecapsEnabled(),
 		Provider:             provider,
 		NewProvider:          deps.newProvider,
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -223,6 +223,9 @@ func mergeConfig(dst *FileConfig, src FileConfig) {
 	if src.Preferences.FavoriteModels != nil {
 		dst.Preferences.FavoriteModels = normalizeFavoriteModels(src.Preferences.FavoriteModels)
 	}
+	if src.Preferences.RecentModels != nil {
+		dst.Preferences.RecentModels = normalizeRecentModels(src.Preferences.RecentModels)
+	}
 	if src.Preferences.Recaps != nil {
 		dst.Preferences.Recaps = src.Preferences.Recaps
 	}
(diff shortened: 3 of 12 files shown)


jatmn (approved):
Thanks for the update. I do not see any actionable code issues in the current patch.

GitHub still reports this PR as blocked, and the earlier Copilot change-request threads for `internal/tui/picker.go` are still unresolved in the review UI even though the current code appears to implement those requests. Please resolve the completed review conversations or otherwise clear the stale requested-review state before this is treated as ready to merge.


@Vasanthdev2004 LGTM

kevincodex1 (approved):
LGTM

PR #570, opened 2026-07-06
feat(permission): auto-classifier autonomy + review transparency + prefix-grant breadth/scope

pengdst:
## Summary

Continuation of the auto-classifier permission mode work (builds on the existing permission-mode / permission-UX line, e.g. #170, #244, #246, #248 — related, not a duplicate). Three cohesive improvements:

### 1. Auto-classifier: real autonomy + transparency
- **Autonomy.** The LLM classifier now reviews actions the sandbox would *otherwise prompt for* and may auto-approve the safe ones. Previously it only re-gated actions the sandbox already auto-allowed, which strictly *added* friction — the opposite of what a "let it run" mode should do.
- **Hard safety guards.** Network, destructive, escalated-sandbox, and out-of-workspace actions are never classifier-eligible; they always ask, regardless of the LLM's confidence.
- **No more silent "ask".** When the classifier declines to auto-approve, its reason is shown on the permission prompt and recorded on the emitted permission event (so it lands in the session log). Auto-approvals were already surfaced (`auto-reviewed · <tool> — <reason>`); asks now are too.
- **Confirmation modal.** Enabling the mode via `Shift+Tab` shows a blocking confirmation explaining that an LLM will auto-review low-risk actions, replacing a fragile timer-based double-press.
- Mode label/summary made consistent across the footer chip, `/mode` card, and modal.

### 2. Command-prefix grant: breadth picker
- The permission prompt can offer several prefix breadths (e.g. `yarn`, `yarn test:*`, `yarn test:unit`) so the approver chooses how wide the grant is instead of getting a single fixed prefix.
- New **intra-token trailing wildcard** on the last token (`test:unit` → `test:*`) to grant a namespace of scripts. Only valid on the last token, never on a lone launcher; mid-command globs stay rejected.
- The chosen breadth is validated against the offered options before granting — a stale or overbroad selection can never widen a grant.

### 3. Command-prefix grant: project vs global scope
- Persisted prefix grants now carry a `Project` scope. **"allow prefix for this project"** matches only inside the current workspace root; **"allow prefix globally"** matches everywhere; session grants remain ephemeral.
- Lookup prefers a project-scoped grant over a global one and never leaks a project grant into other workspaces.

> Note: launcher commands (`npm`, `npx`, `make`, `python`, `node`, …) remain intentionally non-grantable as prefixes, since a prefix grant escalates past the sandbox — this PR does not change that boundary.

## Testing
- `go build ./...` clean.
- `go test ./...` — full suite green (79 packages).
- New tests: classifier ask-reason plumbing, prefix ladder generation, intra-token wildcard, grant-breadth resolution/validation, wildcard matching, project-vs-global scope, and TUI option expansion.

## Summary by CodeRabbit

* **New Features**
  * Added **Auto-classifier** permission mode with an in-UI confirmation flow and auto-approval “reason” display.
  * Expanded permission-mode cycling: **Ask → Auto → Workspace-auto → Auto-classifier → Ask**.
  * Enhanced command-prefix approvals with multi-tier **breadth** options, safer wildcard matching, and **project vs global** grant scoping.

* **Bug Fixes**
  * ACP session mode handling now accepts **Ask, Auto, Workspace-auto, Auto-classifier** and still rejects **Unsafe**.
  * Auto-classifier only triggers when appropriate, strictly validates outputs, and safely falls back to manual prompts on errors/invalid JSON.
  * Prefix grant selection/persistence now honors the offered breadth and scope.


jatmn (changes requested):
I found a few permission-flow issues that should be addressed before this is ready.

## Findings

- [P1] Do not offer one-token package-manager prefixes
  `internal/agent/command_prefix.go:77`
  The new breadth ladder includes every shorter token prefix, so a prompt for a routine command such as `yarn test:unit` can now offer `yarn` as a reusable prefix. Once selected, `matchCommandPrefix` treats any later `yarn ...` command as approved and `shellExecutionArgsForApproval` runs it outside the sandbox, so the grant also covers later `yarn add`, `yarn install`, `yarn publish`, or arbitrary project scripts. That is exactly the package-manager class the PR says should remain non-grantable as a prefix. Please filter launcher/package-manager-only breadths out of the ladder, or classify `yarn`/`pnpm` the same way as `npm`/`npx`/`bun` so broad package-manager grants cannot be offered.

- [P2] Complete CodeRabbit's request to keep a fallback prefix grant
  `internal/agent/loop.go:1096`
  CodeRabbit's current review item is still valid for the remembered-grant path. When the user chooses the new "allow command prefix for this project" option but `GrantCommandPrefixForProject` fails because the sandbox engine has a grant store without a workspace root, the call is allowed once but no session prefix grant is recorded. The engine comment says this case should keep the safer session grant, and the UI offered a reusable project-scoped approval, so later matching commands prompt again instead of reusing even a session-scoped fallback. Please fall back to `GrantCommandPrefixForSession` when project persistence fails, or stop offering the project-scoped option when the engine cannot scope it.

- [P3] Complete CodeRabbit's request to mark project prefix approvals as persistent
  `internal/tui/model.go:3861`
  The project-scoped prefix option persists a grant just like the global prefix option, but the TUI reason text is `approved command prefix for this project in TUI` while the global path says `persistently approved command prefix (global) in TUI`. That leaves the permission event/audit trail ambiguous with the session-only prefix approval. Please make the project reason explicitly persistent and project-scoped so session, project, and global prefix grants remain distinguishable in logs.

Vasanthdev2004 (dismissed):
I reviewed the auto-classifier permission-mode change. The classifier picks a permission mode from the prompt, and the guardrails look careful: it can only widen within bounded, pre-approved tiers and it never escalates past what the user's session policy allows, and the unclassifiable case falls back to the safe default rather than guessing permissive. That fail-closed default is the part I cared most about, and it's right. Approving.

gnanam1990 (dismissed):
## Verdict: Approve

Checked out the branch, merged `main` cleanly, ran `make lint`, and `go test -race -count=1` on `./internal/agent/`, `./internal/sandbox/`, `./internal/tui/`, `./internal/acp/` — all pass.

---

### What looks good

**Auto-classifier mode**
- Classifier only reviews sandbox `ActionPrompt` tools — real autonomy on actions that would otherwise pause, not re-gating already-auto-allowed ones.
- Hard guards block network, destructive, escalated, and out-of-workspace actions from classifier rescue — `canAutoClassifierReview` is thorough.
- Fail-closed on classifier error/timeout/invalid JSON — falls back to normal user prompt.
- 10s timeout + fence stripping + strict JSON decode with `DisallowUnknownFields`.
- Classifier "ask" reason surfaced on permission prompt and session log (`ClassifierReason`).

**Prefix-grant breadth**
- `commandPrefixLadder` never offers one-token launcher prefixes — addresses the P1 concern (`yarn` alone excluded; tests cover `yarn test:unit`, `docker compose up`, single-token `go`).
- `grantPrefixForDecision` validates chosen breadth against offered options — stale/overbroad grants cannot widen.
- Intra-token wildcard (`test:*`) with trailing-`*` validation in sandbox matcher.

**Project vs global scope**
- `CommandPrefixGrant.Project` field + lookup prefers project over global.
- `persistCommandPrefixGrantScopedOrSession` session fallback when project persistence fails.

**TUI/UX**
- Blocking confirmation modal for auto-classifier (replaces fragile double-press).
- Distinct hotkeys: prefix-global `y`, always-allow `f` (was colliding).
- Per-breadth prefix labels + empty hotkey bracket fix.
- Audit text: `persistently approved command prefix (project)` vs `(global)`.

**Tests** — strong coverage across classifier, ladder, wildcard matching, grant validation, TUI options, ACP mode acceptance.

---

### Nits

1. **Scope bundling** — three feature areas plus an unrelated session-resume `tea.ClearScreen` repaint fix. CONTRIBUTING prefers tight scope.
2. **No `issue-approved` linked issue** — references #170/#244 as related continuation but no parent issue with `issue-approved`. Policy gap; not blocking given maintainer approval.
3. **@jatmn still `CHANGES_REQUESTED`** — all three findings appear addressed in `2a31f493`+ but no re-review from them.
4. **Classifier cost** — every prompt-eligible action in auto-classifier mode triggers an extra LLM call on the same provider.
5. **Two-token grants still broad** — `docker compose` as a persisted prefix approves any `docker compose …`. By design with the breadth picker.

---

### Issues

None blocking. Security-sensitive paths (classifier guards, prefix validation, sandbox escalation) look correct.

jatmn (approved):
LGTM

jatmn (changes requested):

## Findings

- [P2] Rebase and resolve the current `main` conflict
  `internal/tui/model.go`
  GitHub currently reports this head as `CONFLICTING`/`DIRTY`, and merging the supplied base (`5e1405d`) with this head (`502845e`) reproduces a content conflict in this file. The PR cannot be merged until it is rebased and the resolved integration is revalidated against current `main`.

- [P2] Carry the new prefix-grant choices through ACP
  `internal/agent/types.go:123`
  ACP now advertises the new permission modes, but its permission bridge drops `always_allow_prefix_for_project`: `optionKindFor` has no case for that action, so `buildPermissionOptions` silently omits it. It also has no representation for `CommandPrefixOptions`, and returns only the action from a selected option, so ACP users cannot choose one of the new breadth rungs and always fall back to the exact default prefix. An editor client therefore cannot use the project-scoped grant at all or use the breadth-picker feature this PR adds. Map the project action and encode/validate the selected offered prefix in the ACP option round trip, with ACP coverage for both behaviors.

pengdst:
Both already fixed. Rebased onto `main` (mergeable now), and ACP prefix grants sorted.

jatmn (changes requested):

## Findings

- [P1] Restore the sandbox credential boundary
  `internal/sandbox/runner.go:377`
  This branch now passes `os.Environ()` directly to sandboxed commands, while `internal/sandbox/profile.go:103` also drops the default deny rules for cloud credential stores. The default workspace profile has a read-all root, so a normal sandboxed command can print provider/VCS/cloud credentials from its environment or read `~/.aws`, gcloud/Azure, or `GOOGLE_APPLICATION_CREDENTIALS` into tool output. Network denial does not prevent that disclosure to the agent/model; a later approved network command can also exfiltrate it. This reverts the base's explicit secret-scrubbing and credential-deny protections. Restore those protections (including their coverage) rather than changing the security boundary in this PR.

- [P1] Enforce project scope before honoring a saved command prefix
  `internal/agent/command_prefix.go:180`
  Project-scoped grants are looked up only against the engine's initial workspace root, not the command's effective directory. For example, a grant for `go test` saved in `/project-a` matches `cd /project-b && go test ./...`: `cd` is treated as a known-safe segment, the `go` segment matches, and `shellExecutionArgsForApproval` then makes the entire command `require_escalated`. It consequently executes unsandboxed in `/project-b`, contradicting the advertised project-only scope. Reject directory-changing/out-of-project composite commands (or bind matching and escalation to the effective command directory).

- [P2] Keep the Windows-safe atomic-replace retry
  `internal/sessions/store.go:844`
  The PR deletes `fsutil.RenameWithRetry` and changes its callers to one-shot `os.Rename` (also cron, swarm mailboxes, and the Windows unelevated setup marker). The deleted helper specifically retried Windows sharing/lock violations from antivirus, indexers, and concurrent readers; those transient locks now fail session/checkpoint persistence, cron updates, mailbox delivery, or sandbox setup outright. Retain the bounded Windows retry for these atomic replacements and its regression coverage.

jatmn (changes requested):

## Findings

- [P1] Preserve the MCP resource scope when starting `serve`
  `internal/cli/serve.go:46`
  This now constructs `mcp.ServeOptions` without `WorkspaceRoot` or `Scope`. Consequently, `zero serve --mcp -C /project` serves resources from the process CWD (the MCP default), rather than `/project`; and the same reversion removes the documented `--add-dir` option entirely. Restore the workspace/scope wiring and its resource-scope coverage so the requested workspace is not silently ignored.

- [P1] Resolve symlinks before treating a prefix grant as project-confined
  `internal/agent/command_prefix.go:224`
  The new `cd` guard only cleans paths and compares them lexically. With `/workspace/link -> /outside`, `cd link && go test ./...` passes the check, matches a project prefix grant, and is converted to an escalated unsandboxed shell invocation. The shell then follows the symlink and runs outside the approved project. Canonicalize/reject symlinked targets before the containment check and cover this escape path.

- [P1] Do not let the classifier override protected-metadata prompts
  `internal/agent/loop.go:1909`
  The sandbox deliberately leaves writes under protected workspace metadata such as `.git/**` as prompts, but those decisions have no block or risk category that excludes them from `canAutoClassifierReview`. A classifier `allow` can therefore write `.git/hooks/pre-commit` or `.git/config` without user approval, defeating the protected-metadata boundary and allowing later Git-triggered code execution. Treat this prompt class as classifier-ineligible.

- [P1] Keep the unavailable-native-sandbox prompt non-overridable
  `internal/agent/loop.go:1909`
  When native shell isolation is unavailable, the sandbox intentionally prompts instead of auto-allowing a shell command. Auto-classifier can currently turn that ordinary prompt into `PermissionGranted=true`; the subsequent evaluation then runs the command unwrapped on the host. Require active native shell isolation before classifier review of a shell command, or preserve this prompt as an explicit user-approval boundary.

- [P1] Do not expose auto-classifier over ACP without the required opt-in
  `internal/acp/agent.go:331`
  The local TUI requires an explicit confirmation before enabling this LLM-based auto-approval mode, but an ACP client can set `auto-classifier` directly and the server advertises it as a normal selectable mode. That lets an integration bypass the user acknowledgement and subsequently auto-approve eligible actions. Keep the mode out of ACP until the protocol can represent that confirmation, or enforce an equivalent server-side opt-in.

Vasanthdev2004 (changes requested):
I am aligned with jatmn's changes-requested on this head, so I will not re-audit the whole thing. Adding my weight on the security one, which I looked at directly.

The prefix-grant cd guard is a real sandbox escape as written. The new path handling only filepath.Clean's the argument and compares it lexically, with no symlink resolution. So with a workspace symlink that points outside (say /workspace/link to /outside), `cd link && go test ./...` still matches a project prefix grant and gets promoted to an escalated, unsanctioned command that runs outside the workspace. Resolve the real path (EvalSymlinks) before the project-confinement check so a symlinked target cannot satisfy it.

jatmn's other P1 needs closing too: serve now builds mcp.ServeOptions without WorkspaceRoot or Scope, so `zero serve --mcp -C /project` serves from the process CWD instead of /project, and --add-dir is dropped. Please restore the workspace and scope wiring.

The classifier core still reads well. The fail-closed default on an unclassifiable prompt is the part I cared about last time, and it is intact. Close those two and this is in good shape.

pengdst:
All addressed, and rebased onto latest main.

- MCP serve scope / `--add-dir`: the branch was trailing main; restored by the rebase.
- Prefix-grant cd escape via symlink: real one. The guard now resolves real paths (EvalSymlinks) before the containment check, so a symlinked-out target no longer satisfies it.
- Protected-metadata prompts (`.git/**` etc): the classifier can no longer auto-approve them; they stay prompts.
- Unsandboxed-shell prompts: the classifier reviews a shell command only when the native sandbox would actually wrap it.
- Auto-classifier over ACP: no longer selectable there, since ACP cannot represent the in-app confirmation, matching how Unsafe is withheld.

Tests added for each.

jatmn (changes requested):

## Findings

- [P1] Keep project-scoped prefix grants inside the project
  `internal/agent/command_prefix.go:178`
  The new containment check follows only literal `cd` segments and assumes the shell starts at `WorkspaceRoot`; it ignores the shell tools' `cwd`/`workdir` argument. After saving a project-scoped `go test` grant in workspace A, a call such as `{command: "go test ./...", cwd: "/outside"}` has no `cd`, matches the A grant, and is changed to `require_escalated` at `loop.go:1351`, so it runs unsandboxed in `/outside`. Resolve and contain the effective tool working directory (including symlinks and aliases) before honoring a project grant.

- [P2] Do not apply the project directory guard to global and session grants
  `internal/agent/command_prefix.go:185`
  The new `commandDirStaysWithinProject` check runs before either persistent or session grant lookup. Consequently, an explicitly global grant—or an ephemeral session grant—such as `go test` is ignored for `cd /other && go test ./...`, even though global grants are documented to match every project and session grants have no project scope. Select the matching grant first, then impose effective-directory containment only when that grant is project-scoped.

- [P2] Keep ACP's config-option mode contract in sync
  `internal/acp/agent.go:392`
  `session/set_mode` now advertises and accepts `workspace-auto`, but the standard `session/set_config_option` path still accepts and lists only `auto` and `ask` (`configOptions` at line 515). Config-driven ACP clients cannot select the newly exposed mode; after `set_mode`, they receive `CurrentValue: workspace-auto` that is absent from the allowed options. Add `workspace-auto` consistently to both config-option paths and cover the round trip.

- [P3] Make the advertised non-colon wildcard breadth usable
  `internal/agent/command_prefix.go:111`
  `intraTokenWildcardPrefix` accepts `-`, `/`, `.`, and `@` as namespace separators, but the sandbox recognizes a wildcard only when its stem ends in `:` (`internal/sandbox/command_prefix.go:94`). A proposed breadth like `test-*` is therefore silently rejected by `ValidCommandPrefix` and never appears in the picker. Either restrict generation to `:` or teach validation and matching the same separator set.

Vasanthdev2004:
Re-checked my 18 July review against the current head. One of my two blockers is genuinely closed. The other, the security one, is half closed, and the half that remains is the half that matters.

**Closed: the `serve` MCP scope.** `internal/cli/serve.go` is not even in this branch's diff any more, and the wiring is intact on the head. I confirmed it behaviourally rather than by reading: built the binary, ran `serve --mcp -C <project>` from a different working directory, and `list_directory {path:"."}` returned the project's contents rather than the process CWD. `--add-dir` grants the extra directory and its absence refuses it. That one is done.

**Still open: the containment guard only watches `cd`.** You did implement exactly what I asked, and it works: `commandDirStaysWithinProject` resolves the root and each `cd` target with `EvalSymlinks` before the containment check, so the symlink vector I described is closed.

But the guard never sees the tool's own directory argument. `matchCommandPrefix` is handed the full `args` map and passes only the parsed command segments to `commandDirStaysWithinProject`, so a directory supplied as an argument rather than as a `cd` is not examined at all. Same project-scoped `go test` grant, same outside directory, driven through `matchCommandPrefix`:

```
cd <outside> && go test ./...              -> matched=false     (your fix, working)
{command:"go test ./...", cwd:<outside>}   -> matched=true
{command:"go test ./...", workdir:<outside>} -> matched=true
```

A match here is what promotes the command to the escalated, unsandboxed path, so this is the same escape the `cd` guard exists to prevent, reached by a different spelling.

What makes it worse than an oversight is which spelling is the recommended one. `internal/tools/bash.go:50` describes the argument as:

> Directory to run the command in. ... Prefer cwd over cd when changing directories.

So the guard covers the discouraged path and leaves the one the schema actively steers the model towards. And `exec_command` accepts four aliases for it, `workdir`, `cwd`, `dir` and `directory`, all equally unguarded.

The fix is where the guard already is: give `commandDirStaysWithinProject` the effective starting directory from the tool arguments instead of assuming it is the workspace root, and resolve it the same way you already resolve a `cd` target. Every alias has to resolve through one accessor, or the next one added reopens this.

No test would catch it today: every `matchCommandPrefix` call in `internal/agent/command_prefix_test.go` passes `command` alone.

**Separately, this cannot merge as it stands.** It conflicts with main in twelve files across `internal/acp`, `internal/agent` and `internal/tui`, including `loop.go`, `types.go` and `permission.go`. The green CI on the head is from 22 July against a base well behind main, so it does not describe the merged result. Worth rebasing before the next round so the next review is of code that can actually land.

The classifier core is still good and my earlier notes on it stand: the protected-metadata refusal, the inactive-shell-sandbox refusal and the ACP opt-in rejection are all in place and all fail when I mutate their predicate out.

Vasanthdev2004 (changes requested):
Sorry about the wait on this, and about the direction I sent you in.

**jatmn's P1 is closed.** The PR no longer touches `serve` at all, so the base wiring stands, and `TestRunServeMCPWiresWorkspaceRootAndAddDirIntoResources` drives the exact reported command and passes.

## The `cd` escape is still open on Windows, and that part is my fault

I told you to use `EvalSymlinks`. On Windows a directory junction reports `ModeIrregular` rather than `os.ModeSymlink`, and `filepath.EvalSymlinks` does not traverse one, it returns the junction's own path. So the guard is inert against a junction:

```
mklink: Junction created for ...\001\jlink <<===>> ...\002
Lstat mode=?rw-rw-rw-  isSymlink=false
EvalSymlinks("...\001\jlink") = "...\001\jlink"  err=<nil>
commandDirStaysWithinProject([cd jlink]) = true
matchCommandPrefix(cd jlink && go test ./...) -> grant{Prefix:["go","test"]}
shellExecutionArgsForApproval -> sandbox_permissions: require_escalated
```

pnpm creates junctions under `node_modules` by default, so this is reachable from an ordinary checkout rather than only a crafted one.

What does work, verified against that same junction: open the path with `windows.CreateFile(..., FILE_FLAG_BACKUP_SEMANTICS)` and call `windows.GetFinalPathNameByHandle` with `FILE_NAME_NORMALIZED|VOLUME_NAME_DOS`. That returns the real target, and `pathWithinRoot` then correctly answers false. `golang.org/x/sys` is already a dependency, so this wants a `_windows.go` with the `EvalSymlinks` path kept as the fallback elsewhere.

Two more that a handle-based resolve would not close, both driven: `cd /` and `cd \`. `filepath.IsAbs` is false for a rooted driveless path, so `resolveCdTarget` joins it onto the root and it collapses back to the root, the guard answers true, and cmd.exe actually lands at the drive root. Create an ordinary in-workspace `Windows\System32` and `cd /Windows/System32 && go test ./...` matches the grant while the shell runs in the real one.

In fairness to the change: base has no containment check at all and matches all of these too, so this is an incomplete fix and not a regression. `cd <abs outside>`, bare `cd`, `cd -`, `cd ~`, `cd $VAR`, `cd C:`, `cd //`, `cd sub\..\..` and POSIX symlinks are all correctly refused now. What makes it blocking is that the PR ships a test asserting the escape is closed, and a comment at `command_prefix.go:246-248` saying the same, and on Windows neither holds.

## Three others worth doing while you are in here

`command_prefix_test.go:195` calls `os.Symlink` and `t.Fatalf`s on failure. On unelevated Windows without Developer Mode that is a hard fail. Every other package in the repo skips instead. CI's runner is Administrator so it stays green, and only contributors see the red suite.

The breadth ladder offers rungs whose entire danger sits in the omitted operands. `rm -rf build/` offers `[rm -rf]`; granted that rung, `rm -rf /`, `rm -rf ~/Documents` and `rm -rf ../other-repo` all match. And because `matchCommandPrefix` sets `permissionGranted=true` before `Evaluate`, the grant also disarms the destructive-command prompt from then on. Requiring a rung to end on a non-flag token would close it.

`isShellCommandTool` keys on the tool name (`bash`, `exec_command`), but the fact being protected is `SideEffect == SideEffectShell`. `write_stdin` and `Task` both carry `SideEffectShell + PermissionPrompt + AdvertiseInAuto`. A/B through `Run` with one fake tool and the name as the only variable: `bash` prompts, `write_stdin` is classifier-allowed and runs with zero prompts.

`canAutoClassifierReview` excludes the `destructive` and `network` categories but has no risk-level check, so the `local_terminal`, `local_desktop` and `local_browser` family auto-approves at RiskHigh with no sandbox confinement at all. `auto_classifier_prompt.md` also tells the reviewing model that the sandbox has already blocked the highest-risk categories before they reach it, which is not true for that family and biases it toward allow.

The classifier core still reads well, and the fail-closed default on an unclassifiable prompt is intact.

jatmn (changes requested):

## Overall guidance

The number of findings here does not come from twelve unrelated feature requests. Most are different symptoms of three shared implementation gaps, which is also why earlier point fixes have exposed another edge in a later review:

1. **The requested prefix scope is carried farther than the effective scope.** A prefix starts as project/global/session, but later paths reduce it to an action or a prefix slice. Matching, persistence fallback, model prompt context, ACP labels, and session audit output then reconstruct scope independently and disagree. This accounts for the global/session directory regression, foreign-project prompt disclosure, missing ACP warning, and incorrect audit labels.
2. **Security policy is inferred from strings and syntax instead of existing typed facts.** Classifier confinement checks two tool names rather than `Safety.SideEffect`; the breadth ladder accepts every syntactically valid slice without asking whether the remaining tokens still bound the operation; Windows containment applies generic filepath behavior that differs from cmd.exe path resolution.
3. **Tests cover the producer but stop before the downstream consumer.** There is good local coverage for creating breadth choices and storing project grants, but less coverage for reopen/match → execute → emit audit state, for both ACP mode entry points, for usage accounting, and for permission-resume timing. That leaves individually correct helpers whose composition is wrong.

To close this without another drip-review cycle, I recommend fixing these as contract groups:

- Have grant matching/persistence return one effective result that retains the selected prefix, actual scope, whether persistence succeeded or fell back to session, applicable project, and whether the current call bypasses the sandbox. Let the execution, ACP disclosure, provider-prompt, and audit consumers use that result instead of inferring state from an action string. This can be a small internal type at the existing grant-resolution seam; it does not require changing the product's three scopes.
- Derive classifier eligibility from the tool's typed safety/risk data and native-sandbox state. Derive prefix warnings from the same predicate that causes sandbox bypass. Keep one source of truth for wildcard separator validity. The goal is to remove parallel allowlists, not add more names to each one.
- Keep the existing project-directory guard, but give its path-resolution seam a Windows implementation that agrees with the shell. This is specifically about junctions and drive-rooted forms already demonstrated; it is not a request to turn project grants into a general filesystem sandbox for arbitrary command arguments.
- Add a compact table-driven matrix that crosses prefix scope (`session`, `project`, `global`) with lifecycle (`select`, `persist/fallback`, `reload`, `match`, `execute`, `audit`) and frontend (`TUI`, `ACP`, provider prompt). Add focused rows for inactive native sandbox + `write_stdin`, destructive breadth, Windows junction/root paths, both ACP mode APIs, and fixed-choice permission resume. Those cases cover the root causes below without expanding feature scope.
- After the grouped changes, run gofmt and the full Linux/macOS/Windows checks, not only the package tests that create the new state. The current live PR is mergeable but blocked, with Ubuntu, macOS, Windows, and Zero Review non-passing.

This guidance is intended to consolidate the existing fixes. It does not ask to remove auto-classifier, remove breadth selection, change the project/global/session product model, expose auto-classifier or unsafe mode through ACP, or revisit specialist autonomy.

## Merge readiness

- [P2] Restore gofmt-clean test files
  `internal/agent/command_prefix_test.go:291`
  The head contains an extra blank line here and another in `internal/agent/export_test.go`. `gofmt -d` identifies both, `make fmt-check` fails locally, and the live Ubuntu job stops at its hard formatting step before later validation can run.

  This is mechanical rather than architectural: run gofmt on the two modified files, verify that no unrelated lines change, and rerun the required checks after the functional fixes below.

## Findings

- [P1] Resolve Windows directory aliases before honoring project grants
  `internal/agent/command_prefix.go:282`
  The new project guard resolves the initial cwd and literal `cd` targets with `filepath.EvalSymlinks`, then checks `filepath.IsAbs`. Those APIs do not describe every path cmd.exe will actually use: Windows directory junctions are not traversed by `EvalSymlinks`, and drive-rooted `/` or `\\` paths are not `filepath.IsAbs` even though cmd.exe resolves them from the current drive root. With `jlink -> C:\outside`, `cd jlink && go test ./...` is judged in-project; likewise, `cd /Windows/System32` can be joined back under the workspace during the check. The saved project prefix then matches and the call is rewritten to `require_escalated`, so execution occurs outside the selected project without another prompt.

  The root cause is a platform-neutral containment helper being treated as authoritative for platform-specific shell semantics. Keep the existing containment contract, but resolve the final Windows directory using Windows' actual path resolution at this seam (for example, a handle-based final-path lookup), and explicitly recognize rooted drive-relative forms before matching. Regression coverage should drive both junction and `/`/`\\` cases through `matchCommandPrefix` and confirm they cannot reach the escalation path, while preserving ordinary in-project directories and the POSIX symlink behavior.

- [P1] Do not offer flag-only destructive prefix rungs
  `internal/agent/command_prefix.go:85`
  `commandPrefixLadder` slices every valid base prefix from two tokens upward, and `ValidCommandPrefix` checks syntax/launcher safety rather than whether the retained tokens still constrain the action. For `rm -rf build/`, this creates `["rm","-rf"]`. If the user selects that rung, later `rm -rf /`, `rm -rf ~/Documents`, and `rm -rf ../other-repo` all match before sandbox evaluation. `permissionGranted` is set and the call is promoted to `require_escalated`, so the newly broad grant suppresses the destructive prompt for a different target.

  Fix the breadth-generation rule, not only the `rm` example. An intermediate rung must retain the operand or subcommand that bounds the approved operation; a rung ending entirely in option flags is not safe merely because it has two tokens. Preserve useful rungs such as a stable command plus subcommand, preserve exact-prefix fallback and offered-choice validation, and add negative tests showing that target-bearing commands cannot produce a target-free reusable grant.

- [P1] Gate classifier eligibility by the actual shell side effect
  `internal/agent/loop.go:2286`
  `canAutoClassifierReview` requires active native isolation only when `isShellCommandTool(tool.Name())` recognizes `bash` or `exec_command`. The built-in `write_stdin` is nevertheless `SideEffectShell`, `PermissionPrompt`, and `AdvertiseInAuto`; nonempty input is deliberately prompt-gated because it can drive an existing process beyond the original command. With no active native sandbox, its sandbox decision is `ActionPrompt`/`RiskHigh`, but the name check is false, so an LLM `allow` sets `permissionGranted` and sends the bytes without the explicit user approval this boundary requires.

  The root cause is using tool identity as a proxy for execution effect. Base eligibility on `Safety.SideEffect`, native-sandbox availability, and the risk classes that are meant to remain explicit; do not maintain another list of shell-like tool names. Add an A/B regression test with identical prompt-gated shell safety and only the tool name changed, plus a test using the real `write_stdin` safety contract, so future shell/host-control tools inherit the guard automatically.

- [P2] Filter approved-prefix prompt context to the active project
  `internal/sandbox/engine.go:150`
  `ApprovedCommandPrefixes` returns every persisted row in the shared grant store. `approvedCommandPrefixContext` then sends all of those command names to the provider without project labels and states that they “do not need another permission prompt.” After project A saves `private-script deploy`, opening project B therefore discloses that project-specific command to the model and tells it the command is approved, while `LookupCommandPrefix` correctly refuses the grant in B. The provider's view of authority is both broader and less private than enforcement.

  The root cause is reusing an administrative, store-wide listing API as an active authorization-context API. Give the prompt consumer a scoped view that contains global grants plus grants for the current project only, and retain scope metadata when rendering it. Keep store-wide listing for `/permissions` or grant management where that breadth is intentional. Test two project roots against one store: B's prompt must exclude A-only rows, include global rows, and agree with runtime lookup.

- [P2] Apply the directory guard only to project-scoped grants
  `internal/agent/command_prefix.go:186`
  `commandDirStaysWithinProject` runs before either persisted or session lookup, so it cannot know the scope of the grant it is guarding. An explicitly global or session `go test` grant is consequently ignored when cwd or a literal `cd` is outside the current engine root. That contradicts the documented model—global matches every project and session is ephemeral but not project-bound—and causes the user to be prompted again for authority they already granted.

  Resolve candidate grants first and carry their effective scope through the segmented-command check. Apply directory confinement if any required matched grant is project-scoped; do not apply it to commands covered solely by global/session grants. Preserve the rule that every unsafe segment must be covered or independently known-safe. Tests should cover outside cwd and `cd` for all three scopes, including a multi-segment command whose matches have different scopes.

- [P2] Account for classifier provider usage
  `internal/agent/auto_classifier.go:47`
  The classifier completion is drained with `zeroruntime.CollectStream`, which records the returned usage only inside the local collected result. Nothing forwards it to `Options.OnUsage`. Main completions, retries, and compaction use `CollectStreamWithOptions(... OnUsage ...)`, so auto-classifier sessions omit up to one paid completion for every reviewed tool call from TUI/session usage totals, cost reporting, trace accounting, and any budget logic attached to that callback.

  Route classifier usage through the same accounting callback exactly once. The clean seam could be passing `OnUsage` into the default classifier factory or returning usage alongside the classifier decision; either is acceptable if custom classifier implementations keep their current contract. Preserve the classifier's invisible text stream—only usage should be forwarded—and add a test asserting one callback for one classifier completion with no double count on allow, prompt, error, or timeout.

- [P2] Keep ACP's workspace-auto mode APIs in sync
  `internal/acp/agent.go:412`
  The direct `session/set_mode` handler now accepts `workspace-auto`, but the parallel `session/set_config_option` handler accepts only auto/ask/plan and `configOptions` advertises the same incomplete list. After a client successfully selects workspace-auto through `set_mode`, its config state returns `CurrentValue: workspace-auto` even though that value is absent from the allowed choices; attempting to set it through the standard config-option path returns an unknown-mode error.

  The root cause is two ACP mode allowlists drifting. Define the ACP-selectable mode set once and reuse it for direct validation, config-option validation, and advertised option construction. Keep the deliberate exclusions intact: auto-classifier still requires an in-app acknowledgement ACP cannot represent, and unsafe remains prohibited. Cover both APIs and the returned option list from the same table so adding the next mode cannot update only one door.

- [P2] Disclose the sandbox bypass for project-prefix approval
  `internal/acp/permission.go:147`
  `optionKindFor` appends `withSandboxEscalationNote` to session and global prefix choices, because selecting either also runs the current command outside the sandbox. The new project action returns a bare “Allow this command for this project” label, yet `shellPrefixApprovalBypassesSandbox` includes that action and `shellExecutionArgsForApproval` performs the identical `require_escalated` rewrite. ACP clients generally display only this label, so the user does not receive the material warning supplied for the sibling choices.

  Derive disclosure from the same `escalates`/bypass predicate for every prefix action instead of adding warnings case by case. Preserve the distinct scope wording, but make the execution consequence identical wherever behavior is identical. A table test over session/project/global actions should assert both the scope label and presence or absence of the escalation note from the actual request conditions.

- [P2] Record the effective prefix scope in permission events
  `internal/agent/loop.go:1252`
  There are two ways the event diverges from enforcement. On reuse, every non-session persisted match is recorded as `PermissionDecisionAlwaysAllowPrefix`, so a project-only match is restored/rendered as “saved prefix (global).” On creation, if project/global persistence fails because the store is unavailable, read-only, or corrupt, `persistCommandPrefixGrantScopedOrSession` intentionally installs a safer session grant but leaves `DecisionAction` as project/global. The transcript then claims persistence that will disappear after restart. This undermines the transparency this PR is specifically adding.

  Stop reconstructing audit scope from the requested action. Have resolution return the effective grant result—matched/persisted/fallback scope and normalized prefix—and build the permission event from that. The intentional session fallback can remain, but it must be visible as session-scoped (and persistence failure may be recorded separately if useful). Add lifecycle tests for project grant → reload → match → event, global match → event, and forced persistence failure → session fallback → event/restore/render.

- [P2] Reset the quiet timer after every fixed permission choice
  `internal/tui/model.go:4504`
  The old resolver cleared the prompt and set `lastStreamActivity = m.now()` so human approval time was not treated as provider silence. The new `resolvePermissionOption` was split out to carry the selected breadth, but it copies only the callback and prompt-clear steps. Enter, hotkey, and click all use this new branch; after a long review pause, the run resumes and immediately renders the stale warning `still generating… 1m30s`. Free-text feedback still uses the old cleanup and does not fail.

  The root cause is duplicated permission-finalization logic. Put the invariant cleanup—invoke once, clear prompt, reset the clock, and preserve the existing peer-specific behavior—in one helper shared by fixed and free-text decisions, with the selected prefix as data rather than a separate lifecycle. Extend the existing failing test across Enter, a hotkey, and a click so future option-shape changes cannot bypass cleanup again.

- [P2] Skip symlink fixtures when Windows cannot create them
  `internal/agent/command_prefix_test.go:232`
  Both new containment cases treat any `os.Symlink` error as a product failure. On ordinary unelevated Windows without Developer Mode, creating a symlink requires a privilege the developer may not have, so the agent package fails before testing containment. GitHub's elevated runner can stay green while supported contributor environments are red, which conflicts with this repository's cross-platform test requirements.

  Separate fixture capability from the assertion: skip with a clear reason only when the host cannot create the symlink, but keep unexpected setup errors and all behavior assertions fatal once creation succeeds. Reuse the repository's existing platform-capability pattern if one exists. This does not replace the junction coverage requested above; symlinks and junctions exercise different Windows behavior.

- [P3] Align generated and accepted wildcard separators
  `internal/agent/command_prefix.go:109`
  `intraTokenWildcardPrefix` finds the last separator from `:-/.@` and can generate `test-*`, path-like, dotted, or at-sign namespace forms. The sandbox's `trailingWildcardStem` accepts a wildcard only when its stem ends in `:`, so `commandPrefixLadder` silently drops every generated non-colon candidate through `ValidCommandPrefix`. The UI therefore cannot offer breadths the generator and comments claim to support.

  Choose one separator contract and share it between generation, normalization, and matching. Restricting generation to `:` is the smallest fix if colon namespaces are the intended feature; supporting more separators requires explicit tests that they cannot widen the launcher or match unintended token shapes. In either case, preserve last-token-only wildcarding, a nonempty stem, and the ban on wildcarding a lone command token.

Vasanthdev2004 (changes requested):
Re-checked at `8639965d`. Of the three commits since my review, the one that matters extends the containment guard to the exec_command `workdir`/`cwd`/`dir`/`directory` aliases and bash `cwd`, which closes the cwd-over-cd escape and is the right shape. It resolves through `filepath.EvalSymlinks` like the `cd` path does, so it inherits the same gap I reported: on Windows a directory junction is not followed by `EvalSymlinks`, the guard answers true, and the grant runs outside the project. There is still no handle-based resolve in the tree (`GetFinalPathNameByHandle` after a `FILE_FLAG_BACKUP_SEMANTICS` open), and `cd /` and `cd \` still collapse onto the root. Those two, the symlink test that hard-fails on unelevated Windows, the flag-terminated breadth rungs, the name-keyed `isShellCommandTool`, and the missing risk-level check in `canAutoClassifierReview` are all as I left them.

CI is red on this head on all three smoke jobs, Windows and macOS in the test step and Ubuntu in the format check, and the merge of main brought 656 files with it, so the branch needs a clean pass before any of the above can be re-verified. jatmn's round at this head stands as well. Keeping my block until the Windows resolve lands.

PR #571, opened 2026-07-06
fix: warn about untracked scratch files left behind after a run

### LINKED ISSUE #551: Agent leaves scratch/debug files in the working directory after task completion
### Version / branch / commit

zero 0.1.0 (npm install, ~July 3 2026)

### OS and environment

Windows 11 / cmd.exe / installed via npm / model: deepseek-v4-pro via Nous routing

### Steps to reproduce

1. Run `zero` in a real Python repo (in my case: a clone of clawnchdev/clawmes)
2. Give it a bounded refactor task from a written spec: implement the change, run pytest, iterate until the suite is green
3. Let it complete the task (it succeeded — 27/27 target tests passing, full suite green, ~824s total)
4. Run `git status` after Zero reports completion

### Expected behavior

The working tree contains only the intended changes to the files the task specified. Any temporary files Zero created for its own debugging/iteration are cleaned up on completion (or created outside the working tree), so `git add` + commit is safe.

### Actual behavior

Zero left 8 scratch files it created while iterating sitting in the repo root: _fix_encoding.py, _fix_test.py, _fix_test2.py, _fix_test3.py, _fix_blank.py, _debug.py, and 2 more.

These would have been committed alongside the real change if not caught manually via `git status` before staging. For a tool whose core use case is autonomous repo-to-PR coding, polluted commits are a real failure mode — a less careful user ships them upstream.

Suggested fix: track agent-created scratch files and delete them on completion, or create them in a temp dir outside the working tree, or print a completion-time summary of "files I created but did not clean up."

(Field-usage report from a real repo-to-PR task, not a code-audit finding. Happy to share the full brief/transcript.)

git status --short after completion showed the untracked _fix_*.py / _debug.py files alongside the 2 intended modified files. Screenshot of Zero's FILES panel showing the 8 created scratch files available on request.

PierrunoYT:
Fixes #551.

## Problem
A headless \zero exec\ run could leave behind debug/scratch files the model created while iterating (e.g. \_fix_test.py\, \_debug.py\) sitting untracked in the repo. These would then be silently swept up by a later \git add -A\ and committed by accident.

## Fix
Rather than guessing which files are 'scratch' and deleting them (risky — a wrong guess destroys real work), Zero now surfaces a completion-time warning listing any brand-new file it created this run that is still present and untracked by git when the run ends, so it's caught before staging/committing instead of after.

- \FileTracker\ gains \RecordCreated\/\CreatedFiles\, tracking brand-new files distinct from its existing version-baseline (used for external-modification conflict detection).
- \write_file\ records a path as created only the first time it writes a file that didn't already exist — an overwrite of an existing file is not counted.
- After a headless run completes, \xec.go\ checks \git status\ for any created file that's still present and untracked, and surfaces a warning via the existing \writer.warning()\ (works across text / \-o json\ / stream-json output).

Silent (no-op) when: nothing was created, git isn't available, the workspace isn't a git repo, or the created file was removed or already tracked/committed by the model itself.

## Testing
- \go build ./...\
- \go test ./internal/tools/... ./internal/cli/...\ (all pass, including new tests for \FileTracker.RecordCreated\/\CreatedFiles\, \write_file\'s create-vs-overwrite tracking, and \scratchFileWarning\'s git-status behavior across tracked/untracked/removed/non-repo cases)
- Full \go test ./...\ — all green except a pre-existing, unrelated TUI scroll-rendering test failure (\internal/tui\) that reproduces identically on a clean checkout with none of this PR's files touched.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -729,6 +729,8 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		Scope:         scope,
 	})
 	lastKnownMCPConfig := mcpConfig
+	fileTracker := tools.NewFileTracker()
+	scratchBaseline := scratchFileSnapshot(workspaceRoot)
 	return deps.runTUI(context.Background(), tui.Options{
 		Cwd:                  workspaceRoot,
 		Version:              version,
@@ -747,12 +749,15 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		NewProvider:          deps.newProvider,
 		ProbeProviderHealth:  deps.probeProviderHealth,
 		UserAgent:            userAgent(),
-		Registry:             registry,
-		SessionStore:         deps.newSessionStore(),
-		SandboxStore:         sandboxStore,
-		MCPConfig:            mcpConfig,
-		MCPPermissionStore:   mcpPermissionStore,
-		MCPTokenStore:        mcpTokenStore,
+		RunCompletionWarning: func() string {
+			return scratchFileWarning(workspaceRoot, scratchBaseline)
+		},
+		Registry:           registry,
+		SessionStore:       deps.newSessionStore(),
+		SandboxStore:       sandboxStore,
+		MCPConfig:          mcpConfig,
+		MCPPermissionStore: mcpPermissionStore,
+		MCPTokenStore:      mcpTokenStore,
 		MCPCommand: func(ctx context.Context, args []string) tui.MCPCommandResult {
 			if ctx == nil {
 				ctx = context.Background()
@@ -778,7 +783,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 			PermissionMode: permissionMode,
 			Autonomy:       "low",
 			Sandbox:        sandboxEngine,
-			FileTracker:    tools.NewFileTracker(),
+			FileTracker:    fileTracker,
 			Hooks:          newHookDispatcherWithExtra(workspaceRoot, pluginActivation.hooks),
 			DeferThreshold: resolved.Tools.DeferThreshold,
 			Specialists:    specialistRuntime.specialists,
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -736,7 +736,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 	})
 	lastKnownMCPConfig := mcpConfig
 	fileTracker := tools.NewFileTracker()
-	scratchBaseline := scratchFileSnapshot(workspaceRoot)
+	var scratchBaseline scratchFileBaseline
 	sttServerManager := newDictationServerManager(resolved.STT)
 	// Keep STT downloads in the SAME config tree the rest of the TUI uses. Deriving
 	// from userConfigPath (rather than config.UserConfigDir()) matters when the config
@@ -766,6 +766,9 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		NewProvider:          deps.newProvider,
 		ProbeProviderHealth:  deps.probeProviderHealth,
 		UserAgent:            userAgent(),
+		PrepareRunCompletionWarning: func() {
+			scratchBaseline = scratchFileSnapshot(workspaceRoot)
+		},
 		RunCompletionWarning: func() string {
 			return scratchFileWarning(workspaceRoot, scratchBaseline)
 		},
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -510,10 +510,10 @@ func runExec(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 	// lspShutdown tears down any language-server sessions either half spawned.
 	selfCorrector, fileDiagnostics, lspShutdown := newExecSelfCorrector(options.selfCorrect, workspaceRoot, options.autonomy)
 	defer lspShutdown()
-	// Kept as a named variable (rather than inlined into Options below) so its
-	// CreatedFiles() can be inspected after the run for the scratch-file
-	// completion warning (issue #551).
+	// Kept as a named variable (rather than inlined into Options below) because
+	// the write tools also use it for conflict checks across a run.
 	fileTracker := tools.NewFileTracker()
+	scratchBaseline := scratchFileSnapshot(workspaceRoot)
 	result, err := agent.Run(runCtx, agentPrompt, provider, agent.Options{
 		MaxTurns:         resolved.MaxTurns,
 		ContextWindow:    resolveAgentContextWindow(runCtx, modelRegistry, resolved.Provider),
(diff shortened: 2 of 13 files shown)


ddfp777:
Reviewed this against my original repro from #551 — the write_file → RecordCreated → git-status-porcelain flow correctly catches the exact case I hit (agent leaves _fix_test.py untracked, user gets warned). Mechanism is sound and the test coverage is solid.

One gap worth a follow-up: RecordCreated is only called from write_file.go — edit_file and apply_patch can also create new files (they handle the "doesn't exist yet" case as creation), and those paths aren't hooked, so a scratch file created via edit_file would slip through the warning. Not a blocker for this PR imo — the common path is covered — but worth a follow-up issue.

Nice fix

jatmn (changes requested):

## Findings

- [P1] Wire the scratch-file warning into the TUI path
  `internal/cli/exec.go:659`
  The linked issue was reported from an interactive `zero` run and even references the TUI `FILES` panel, but this PR only emits `scratchFileWarning` from headless `zero exec`. The regular TUI path in `internal/cli/app.go` still just passes a `FileTracker` into `agent.Options`, and `internal/tui/model.go` never calls this new warning helper when an agent run finishes. That leaves the original workflow unchanged: a user can finish an interactive coding task with leftover `_fix_test.py` / `_debug.py` files visible only if they manually inspect git status or the sidebar. Please surface the same completion-time warning in the interactive run path, or narrow the PR so it no longer claims to fix #551's TUI repro.

- [P2] Detect shell-created scratch files too
  `internal/tools/bash.go:190`
  The warning only sees paths recorded by `write_file` and `apply_patch`, but the scratch/debug scripts from #551 can also be created through shell commands, for example a heredoc or redirected command that writes `_fix_test.py`. `bash` returns a plain successful `Result` without `ChangedFiles` and does not receive `RunOptions.FileTracker`, while the existing TUI git sweep explicitly documents shell-created files as a changed-files blind spot. In those exec runs, `fileTracker.CreatedFiles()` stays empty and no completion warning is printed even though the same untracked scratch files are left behind. Please base the warning on a before/after git status sweep or otherwise include shell-created untracked files, not only dedicated file-tool creations.

ddfp777:
+1 on P2 from my end, I flagged the same creation-path blind spot reviewing this against my #551 repro: the tracker hooks write_file/apply_patch but shell-created files (heredocs, redirects) slip through. A before/after git status sweep like jatmn suggests would also catch the edit_file path I noted earlier. Happy to test whatever lands against the original Windows repro.

Vasanthdev2004 (changes requested):
I read the scratch-file warning change end to end. The plumbing is clean — FileTracker.RecordCreated on write_file and apply_patch for genuinely new files, CreatedFiles() at run end, and scratchFileWarning running git status --porcelain against just the created paths. The git fallbacks (no git binary, not a repo, file already removed, file already tracked) are all handled and tested. That part is solid.

The problem is the signal. The warning fires for every created file that's still untracked when the run ends — but a brand-new deliverable file (the normal outcome of a coding task: "create a new module", "scaffold a handler") is also created-and-untracked at run end. So on most real runs the user gets told "this run created N new files that may be scratch/debug output" about files that are exactly what they asked for. That cries wolf on the common case, and after a few false alarms people learn to ignore the warning — at which point it stops catching the actual stray _debug.py it was built for.

The fix should distinguish intended deliverables from leftover scratch. A few options: only warn for created files the model never read/edited again after creating (scratch tends to be write-once-never-touched), or pattern-match scratch-like names (temp/debug/test/scratch prefixes/suffixes) and warn only for those, or let the model tag scratch files. As-is the warning is too noisy to be useful. Requesting changes for the false-positive rate.

Vasanthdev2004 (approved):
Thanks for the follow-up — the false-positive concern from my earlier review is resolved cleanly. I pulled the latest head (a8345d0b) and traced the new scratch-warning lifecycle:

- internal/cli/app.go (~769) — the TUI scratch baseline is no longer a one-time snapshot taken at startup. It's now captured per run, in the new `PrepareRunCompletionWarning` callback, which takes a fresh `scratchFileSnapshot(workspaceRoot)` each time. So each run only ever warns about scratch files it itself left behind, not pre-existing ones. That's the core false-positive fix: files that were already there get absorbed into the baseline and stop crying wolf.
- internal/tui/model.go (~4445) — `beginRun` invokes `prepareRunCompletionWarning()`, so the baseline is stamped at the start of every run, not once for the session.
- internal/cli/exec.go — `emitScratchWarning` was extracted and is now also called on the `err != nil` path (~610), so cancelled/errored headless runs still surface leftover scratch files instead of going silent. There's no double-emit because the error branch returns before the success-path call (~664). That's a real improvement over the old behavior, which only warned on success.

Validation on the PR head:
- `go build ./...` — clean
- `go vet ./internal/cli/... ./internal/tools/... ./internal/tui/...` — clean
- `go test ./internal/cli/... ./internal/tools/...` — the scratch/tracker/runs tests pass, including the new `TestRunExecWarnsAboutScratchFileWhenProviderErrors` and the existing `scratchFileWarning` suite. `TestBeginRunPreparesRunCompletionWarningEachTurn` and `TestAgentResponseSurfacesRunCompletionWarning` also pass. The only full-suite failure is `TestAltScreenTranscriptScrollKeepsFooterFixed`, a known pre-existing Windows terminal-rendering issue unrelated to this PR.

I specifically looked for a false negative — a real leftover scratch file that escapes warning under the new lifecycle that the old one would have caught — and couldn't construct one. The per-run baseline is captured at run start and compared at run completion, and the next run's baseline correctly absorbs already-warned files.

One non-blocking note: there's no test covering an explicitly cancelled/interrupted TUI run still surfacing the warning. The completion handler fires outside the `msg.err` branch, so the code path supports it, but it isn't exercised directly. Not worth blocking on.

Approving.

ddfp777:
Original filer of #551, the per-run baseline approach resolves this cleaner than any of the naming heuristics discussed, and warning on errored runs too is a genuine upgrade over the original ask. Once this lands in a release I'll re-run the Windows repro from #551 and confirm on the issue. Thanks @PierrunoYT @Vasanthdev2004 🦞

gnanam1990 (approved):
## Verdict: Approve

Checked out the branch, merged `main` cleanly, ran `make lint`, and targeted tests (`./internal/cli/ -run Scratch`, `./internal/tools/`, `./internal/tui/ -run RunCompletion`) — all pass.

Fixes #551.

---

### What looks good

- **Warn, don't delete** — surfaces leftover artifacts without guessing and auto-removing real work. Right tradeoff.
- **Per-run git baseline** (`scratchFileSnapshot` at run start → `scratchFileWarning` at end) fixes the false-positive problem. Pre-existing scratch files are absorbed into the baseline; only *new* scratch-like untracked files warn.
- **Naming heuristics** (`scratchLikePath`) target the actual repro (`_fix_test.py`, `_debug.py`) without warning on normal deliverables like `app.py` or `_generated.go` — covered by tests.
- **Shell-created files caught** — git before/after comparison catches heredocs/redirects that bypass `write_file`/`apply_patch`.
- **TUI wired** — `PrepareRunCompletionWarning` in `beginRun()`, warning surfaced in transcript on run completion (success and error paths).
- **Headless exec** — `emitScratchWarning()` on both success and provider-error paths; `TestRunExecWarnsAboutScratchFileWhenProviderErrors` covers the error case.
- **Silent fallbacks** — no git, not a repo, file removed, file tracked/committed, pre-existing scratch — all tested and handled.
- **Good test matrix** — 9 unit tests in `scratch_files_test.go` plus integration/protocol tests.

---

### Nits

1. **`FileTracker.CreatedFiles()` is unused in the warning path** — `RecordCreated` is wired in `write_file`/`apply_patch`, but completion warnings use git status + `scratchLikePath`, not the tracker. Either wire it in or drop it to avoid misleading future readers.
2. **`scratchLikePath` substring matching is broad** — `strings.Contains(name, "temp")` would match legitimate files like `template.go` or `temperature_sensor.go`. Same for `"repro"` → `reproduction_test.go`. Low frequency in practice, but worth a follow-up tightening.
3. **PR description is stale** — still describes `CreatedFiles()` at run end; actual implementation is git-baseline + naming heuristics.
4. **Policy note** — #551 has `bug` label but no `issue-approved` per CONTRIBUTING. Not blocking given maintainer approval.

---

### Issues

None blocking. The false-positive concern from the initial review is resolved cleanly in `a8345d0b` (per-run baseline + scratch-like filter).

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #572, opened 2026-07-06
Feature/GitHub copilot

aitorse:
## Linked issue

Fixes #

## Checklist

- [ ] `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Copilot as a supported sign-in and provider option, including device-code login flow and Copilot model enablement after login.
  * Improved model discovery/curation (recommended-by-default picker behavior and endpoint-based filtering).
  * Added a persisted setting for the default interactive permission mode (`ask`, `auto`, `unsafe`).
* **Bug Fixes**
  * Improved reconstruction of streamed tool-call arguments when keys arrive flattened.
  * Refined Copilot request routing and headers to match the correct API style.
* **Documentation**
  * Updated first-run Copilot setup instructions and onboarding notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,8 @@ zero auth login copilot        # opens github.com/login/device, enter the code
 zero --provider copilot --model gpt-4.1
 ```
 
-Zero exchanges the GitHub login for a short-lived Copilot token per request.
+Zero exchanges the GitHub login for a short-lived Copilot token, reusing it
+until it is near expiry or a request receives a 401.
 This uses GitHub's undocumented Copilot API (the same one the editor plugins
 use); it is not an officially supported developer API and may change or break
 without notice, and heavy automated use can trip GitHub's abuse detection.
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ zero exec --output-format stream-json < turns.jsonl
 ## Why Zero
 
 - **Use the model you want.** Bring OpenAI, Anthropic, Gemini, Groq, OpenRouter,
-  DeepSeek, Mistral, xAI, Qwen, Kimi, GitHub Models, Ollama, LM Studio, or any
-  OpenAI-/Anthropic-compatible endpoint.
+  DeepSeek, Mistral, xAI, Qwen, Kimi, GitHub Models, GitHub Copilot, Ollama, LM
+  Studio, or any OpenAI-/Anthropic-compatible endpoint.
 - **Stay in control.** File writes, shell commands, network access, and
   out-of-workspace writes go through Zero's permission and sandbox policy.
 - **Works in the terminal.** The TUI has model/provider pickers, image input,
@@ -159,6 +159,19 @@ zero providers setup longcat --set-active
 
 For local models, run Ollama or LM Studio and then use `zero setup` or
 `zero providers detect`.
+
+For GitHub Copilot (requires an active Copilot subscription), log in with the
+device-code flow and select the provider:
+
+```bash
+zero auth login copilot        # opens github.com/login/device, enter the code
+zero --provider copilot --model gpt-4.1
+```
+
+Zero exchanges the GitHub login for a short-lived Copilot token per request.
+This uses GitHub's undocumented Copilot API (the same one the editor plugins
+use); it is not an officially supported developer API and may change or break
+without notice, and heavy automated use can trip GitHub's abuse detection.
 
 ## Daily Use
 
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -862,12 +862,20 @@ func recoverableToolArguments(arguments string) (first string, ok bool) {
 // A genuinely malformed payload (bad/truncated first object, or non-JSON trailing
 // garbage) still returns the real parse error; a standard single-object payload
 // decodes exactly as before.
+//
+// It also repairs flattened dotted/bracketed argument keys (e.g. Gemini's
+// streaming `{"plan[0].content":"x"}`) back into the nested structure the tool
+// schema expects; see unflattenToolArguments. Normal payloads are untouched.
 func decodeToolArguments(arguments string, v any) error {
+	candidate := arguments
 	if first, ok := recoverableToolArguments(arguments); ok {
-		return json.Unmarshal([]byte(first), v)
+		candidate = first
 	}
-	// Not cleanly recoverable — surface the genuine parse error.
-	return json.Unmarshal([]byte(arguments), v)
+	// Un-flatten only rewrites payloads whose keys carry nested paths; every other
+	// payload is returned byte-for-byte, so this preserves the genuine parse error
+	// for a non-recoverable payload below.
+	candidate = unflattenToolArguments(candidate)
+	return json.Unmarshal([]byte(candidate), v)
 }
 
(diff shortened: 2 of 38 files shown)


jatmn (changes requested):

## Findings

  `GitHub mergeability`
  GitHub currently reports `mergeable=CONFLICTING` and `mergeStateStatus=DIRTY`, so this branch cannot be merged as-is. Please rebase on the latest `main` branch and resolve the conflicts before addressing the review findings below, so the fixes apply to the code that will actually be merged.

- [P1] Link the approved parent issue for this community PR
  `PR description`
  The PR body still has `Fixes #` blank and the approved-issue checklist items are unchecked, but `CONTRIBUTING.md` requires community pull requests to link an existing issue that already has the `issue-approved` label and to stay focused on that approved scope. Please add the approved issue link, or get explicit maintainer direction that this PR is exempt, before continuing the implementation review.

- [P1] Complete CodeRabbit's request to cap flattened array indexes
  `internal/agent/tool_args_unflatten.go:101`
  CodeRabbit's unbounded-index thread is still valid: `parseFlatKey` accepts any non-negative integer, and `build()` later allocates `make([]any, max+1)` from the largest index. A streamed tool payload such as `{"plan[999999999].content":"x"}` can therefore force a huge allocation while decoding tool arguments. Please reject oversized indexes and keep the original payload fallback path for that case, with a regression test covering the large-index input.

- [P2] Complete CodeRabbit's request to scope the Copilot model-format cache per host
  `internal/provideroauth/copilot_models.go:37`
  `copilotModelFormats` is a single process-wide map keyed only by model id, even though `CopilotModelAPIFormat` receives an account-specific `baseURL` derived from the minted Copilot token. After one Copilot account or host populates the cache, switching to another host within the five-minute TTL reuses the first host's `/models` capability map, so a model can be routed to chat-completions when that account only supports `/responses` (or the reverse). Please key the cached map by the resolved base URL/account host and avoid holding the cache mutex while the network fetch is in progress.

- [P2] Complete CodeRabbit's request to avoid truncating the request before deriving Copilot headers
  `internal/provideroauth/copilot.go:155`
  `copilotPeekRequest` reads only the first 1 MiB of the outgoing body and then unmarshals that partial buffer to decide `X-Initiator` and `Copilot-Vision-Request`. Large normal conversations can put the final assistant/tool turn or image marker after that cutoff, making JSON decode fail or miss the marker and defaulting the request to `X-Initiator: user` without the vision header. Please derive these headers from the full request shape, or otherwise preserve the final message/input and image markers before applying the size guard, and add coverage for the large-body path.

aitorse:
All findings addressed, issue #657 opened and waiting for approval.

Vasanthdev2004 (commented):
Checked #572 at current HEAD. The earlier findings are addressed — it rebases clean now, coderabbitai approved on the latest push, and the review nits are resolved. The one remaining gate is the approved-issue policy: #657 (opened to back this feature) is still waiting on the `issue-approved` label, which is kevin's call. Once #657 gets it, this is ready to go from my side.

jatmn (changes requested):

## Findings

- [P1] Link an approved parent issue before continuing this community PR
  `PR description`
  This is a `FIRST_TIME_CONTRIBUTOR` PR, but its body leaves `Fixes #` blank and #657 is still open with only the `enhancement` label. `CONTRIBUTING.md` requires community PRs to link an issue that already has `issue-approved`; without that, this feature is not eligible for merge regardless of the implementation state. Please wait for approval and add the issue link (or obtain an explicit maintainer exemption).

- [P1] Authenticate the `providers models` Copilot path
  `internal/cli/provider_models.go:50`
  `zero providers models copilot` only applies an inline/stored API key before probing `/models`. A Copilot profile intentionally has no usable API key: its stored GitHub token must be minted into a Copilot bearer and accompanied by the editor headers/account-specific base URL. Consequently, the newly supported login succeeds but this advertised model-list command sends an anonymous request and receives 401. Route this CLI discovery path through the same Copilot OAuth discovery resolution used by the TUI.

- [P1] Do not reinterpret literal tool parameter names as flattened paths
  `internal/agent/tool_args_unflatten.go:44`
  `decodeToolArguments` now runs this reconstruction for every tool call, and any top-level key containing `.` or `[` is treated as path syntax. JSON Schema/MCP tools may legitimately declare literal properties such as `com.example.option` or `labels[0]`; a valid call is rewritten into nested data before validation/dispatch, so required parameters disappear or the wrong side effect is requested. Make reconstruction provider/schema-aware, or preserve literal keys unless the target schema proves the flattened form is expected.

- [P2] Bound flattened argument path depth before recursively rebuilding it
  `internal/agent/tool_args_unflatten.go:94`
  Array indices are capped, but dotted/bracketed segments are not. A streamed argument with a single deeply nested key (`a.a.a...`) is accepted, then recurses once per segment in both `unflattenNode.set` and `build`; an upstream model can therefore exhaust stack/CPU/memory before tool validation. Add a maximum segment depth (or make reconstruction iterative) and return the original payload when it is exceeded.

- [P2] Apply Copilot’s policy-model enablement to TUI device login
  `internal/tui/oauth_device.go:150`
  The CLI login calls `EnableCopilotModels`, but the TUI’s newly forced Copilot device-flow completion only stores the GitHub token. Its subsequent discovery therefore never enables the policy-gated models that the implementation says do not surface or accept requests until enabled. Share the post-login enablement step across successful CLI and TUI Copilot login flows so interactive setup exposes the same account catalog.

Vasanthdev2004:
This sat on our side, not yours, and I am sorry for that. The one P1 in the review was to link an approved parent issue; #657 has carried the issue-approved label since Aug 12 and nobody came back to tell you. So the parent is settled and the code is what remains. The branch is two months old and the diff is large, so the practical path is a rebase onto current main, then addressing the remaining findings in that review. When it is pushed, re-request review and I will pick it up promptly this time. If you no longer have time for it, say so and we will either carry it or close it cleanly; no hard feelings either way.

aitorse:
> This sat on our side, not yours, and I am sorry for that. The one P1 in the review was to link an approved parent issue; #657 has carried the issue-approved label since Aug 12 and nobody came back to tell you. So the parent is settled and the code is what remains. The branch is two months old and the diff is large, so the practical path is a rebase onto current main, then addressing the remaining findings in that review. When it is pushed, re-request review and I will pick it up promptly this time. If you no longer have time for it, say so and we will either carry it or close it cleanly; no hard feelings either way.

Please take over, I don't have time at the moment, please feel free to regenerate and test the code before merging.

Vasanthdev2004:
Closing this one, and thank you for the work that went into it.

You asked us to take it over on 8 September, and after looking at what that means we are not going to carry it as this branch: it is two months behind main with conflicts, and the four code findings from the July review (authenticating the providers models Copilot path, not reinterpreting literal tool parameter names as flattened paths, bounding the flattened path depth, and applying the policy-model enablement to the TUI device login) are still open. A provider integration needs an owner who can rebase it, close those, and test the device login against the real service, and none of us has that time right now. Closing cleanly is more honest than leaving it to age.

Nothing here is lost. The parent issue #657 stays open and keeps its approval, so the feature is still wanted, and this branch stays on your fork for anyone who picks it up next, with the review findings as the map of what remains. If you or someone else wants to take another run at it later, open a fresh PR against current main and reference #657; we will review it promptly.

PR #573, opened 2026-07-07
fix(tui): paste protection for Termux char-by-char paste

### LINKED ISSUE #583: fix(tui): Termux context-menu paste injects characters one at a time, breaking multiline input
## Problem

On Termux (Android terminal), the long-press context menu paste injects characters **one at a time**, including newlines as raw `KeyEnter` events. This means pasting any multiline text submits on the first `\n` before the user can review or edit it.

The existing `tea.PasteMsg` (bracketed paste / Ctrl+Shift+V) works fine, but Termux context menu paste bypasses bracketed mode entirely.

## Proposed fix

A timing-based paste detector that tracks consecutive keypresses within a 100ms window. When Enter arrives inside such a burst, it inserts a newline instead of submitting. The burst counter is reset at every intentional action boundary (submit, modal enter/exit, etc.) so normal typing is unaffected.

See PR #573 for the implementation.

## Environment

- Termux on Android ARM64
- Any terminal app using long-press context menu paste

PatrickNoFilter:
## Problem

On Termux (Android terminal), context menu paste injects characters **one at a time**, including newlines as raw `KeyEnter` events. This means pasting any multiline text submits on the first \n before the user can review it.

The existing `tea.PasteMsg` (bracketed paste mode) handler works for Ctrl+Shift+V pastes but does **not** fire for Termux's long-press context menu paste, which bypasses bracketed mode entirely.

## Solution

A **timing-based paste detector** that runs alongside the existing bracketed-paste path. See PR #573 for the implementation.

Fixes #583

## Summary by CodeRabbit

* **New Features**
  * Improved composer Enter handling during rapid typing: burst detection now inserts a newline and clears suggestions instead of submitting, with Termux-specific protection.
  * If the composer is already multiline, bursts trigger newline insertion (desktop included).
* **Bug Fixes**
  * Prevented burst state from carrying over by clearing it when leaving overlays, exiting flows, and switching between confirmation/wizard/picker/modal surfaces.
* **Tests**
  * Added deterministic burst/Enter tests covering Termux vs desktop behavior, multiline composer handling, and burst reset after submit.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1394,12 +1394,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker != nil {
@@ -1539,12 +1542,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m.moveAskUserTab(1), nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker == nil && m.suggestionsActive() {
@@ -1570,12 +1576,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m.moveAskUserCursor(-1), nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker != nil {
@@ -1606,12 +1615,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m.moveAskUserCursor(1), nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker != nil {
@@ -1641,12 +1653,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m.moveAskUserCursor(1), nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker != nil {
@@ -1684,12 +1699,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m.moveAskUserCursor(-1), nil
 			}
 			if m.providerWizard != nil {
+				m.burstCount = 0
 				return m.handleProviderWizardKey(msg)
 			}
 			if m.mcpAddWizard != nil {
+				m.burstCount = 0
 				return m.handleMCPAddWizardKey(msg)
 			}
 			if m.mcpManager != nil {
+				m.burstCount = 0
 				return m.handleMCPManagerKey(msg)
 			}
 			if m.picker != nil {
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1421,11 +1421,13 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 			if m.suggestionsActive() {
 				return m.chooseSuggestion()
 			}
-			// Timing-based paste protection: 3+ rapid keys within 100ms
-			// means we are inside a character-by-character paste (Termux
-			// context menu). 1-2 fast keys + Enter is normal fast typing
-			// and should submit, not insert newline.
-			if m.burstCount > 2 {
+			// Timing-based paste protection: under Termux, context-menu paste
+			// injects characters one at a time (including newlines as raw
+			// KeyEnter events). A sustained burst of 3+ keys within 100ms
+			// means we are inside a char-by-char paste — insert newline
+			// instead of submitting. Gated to Termux so fast desktop typing
+			// (which can reach similar inter-key intervals) is never affected.
+			if os.Getenv("TERMUX_VERSION") != "" && m.burstCount > 2 {
 				state := m.currentComposerState()
 				m = m.insertComposerTextWithPastePreview(state, "\n", "")
 				m.clearSuggestions()
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -192,6 +192,12 @@ type model struct {
 	// renders the live elapsed time from it so a long or stalled turn never looks
 	// like a frozen terminal (for ANY provider, not just slow ones). Zero = idle.
 	turnStartedAt time.Time
+	// lastCharTime tracks when the last non-Enter key was received, for paste detection.
+	lastCharTime time.Time
+	// lastKeyTime tracks every keypress timestamp for burst calculation.
+	lastKeyTime time.Time
+	// burstCount counts consecutive keypresses within 100ms (paste mode).
+	burstCount    int
 	queuedMessage string
 	// loops holds the session's active /loop definitions (see loop.go). activeLoopID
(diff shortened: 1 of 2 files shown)


PatrickNoFilter:
**Re: timing concern** — the 100ms window is chosen for Termux compatibility.

I tested this on a real Termux device (Android ARM64). With a 50ms window the paste was **not detected** (inter-character timing on this device exceeded 50ms). Bumping to 100ms reliably catches paste while the existing literature suggests fast human typists average ~120ms IKI with individual gaps rarely below 80ms.

The false-positive window is further narrowed by two design constraints:
1. Must have **≥2 consecutive keys** within 100ms (the `burstCount > 1` check) — a single fast char + Enter won't trigger it
2. Every intentional action resets `burstCount`, including **all** modal handlers and `handleSubmit()` itself — normal typing triggers none of these resets because the user pauses before pressing Enter

If desktop users report false positives we can add platform-specific tuning, but on Termux this window is the minimum that works reliably.

Re: the wizard paths — fixed in commit 40c692b. ✅

jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to avoid treating fast typing as paste
  `internal/tui/model.go:1427`
  CodeRabbit's earlier request about the burst detector being too aggressive is marked resolved, but the current code still increments `burstCount` for every keypress and then treats `burstCount > 1` on Enter as paste. A normal fast sequence like typing two characters and immediately pressing Enter within the 100ms windows reaches this branch and inserts a newline instead of submitting, which breaks the primary composer flow for fast users. The reset is also scattered outside `handleSubmit()` while the timing source is `time.Now()`, so the behavior is hard to cover deterministically and other submission paths can keep stale burst state. Please make the paste detector distinguish actual char-by-char paste from ordinary rapid typing plus Enter, reset the state at the submission boundary, and add deterministic coverage proving normal fast typing still submits.

- [P2] Link an approved parent issue for this community PR
  `CONTRIBUTING.md:45`
  The contribution policy says all community pull requests require an approved parent issue, and the PR description does not link one for this Termux paste change. Please link the approved issue with the `issue-approved` label, or get maintainer confirmation that this PR is exempt from the issue-first requirement, before continuing the implementation review.

PatrickNoFilter:
Scratch that — will open a separate issue to track this per CONTRIBUTING.md.

PatrickNoFilter:
Created issue #583 — waiting for the core team to review and add the `issue-approved` label before this PR can proceed. PTAL @jatmn

Vasanthdev2004 (changes requested):
I read the Termux paste-protection diff. The structure is right — count rapid keypresses and, when a burst is detected, insert a newline instead of submitting so a character-by-character Termux paste doesn't fire a premature submit. Two things before merge:

1. The burst threshold (3+ keys within 100ms) will false-trigger on fast typing, not just pastes. A fast typist can land three keystrokes inside 100ms on common bigrams, and the code's own comment says "1-2 fast keys + Enter is normal fast typing and should submit" — but burstCount > 2 means three rapid keys, which is already within reach of a fast typist, and then the next Enter inserts a newline instead of submitting. On a phone that's the whole point, but this guard runs for every keypress on every platform, so a fast desktop typist will intermittently get their Enter swallowed. Consider gating this to the Termux/Android path only, or raising the count threshold / widening the window, so normal fast typing on a real keyboard still submits.

2. There's no regression test pinning the behavior in both directions. The model_test.go change only adjusts the canned timestamp slice; nothing asserts that a real burst inserts a newline AND that fast-but-normal typing still submits. Add a test that drives a burst and asserts the newline path, and one that drives a couple of fast keys + Enter and asserts it still submits — otherwise the threshold can drift in either direction silently.

Requesting changes for the false-trigger risk and the missing bidirectional test.

Vasanthdev2004 (approved):
Re-reviewed after the latest commits (Termux gating, injectable clock, bidirectional tests, reset-after-submit). I checked out the branch and ran the four new burst tests — all pass, and `go build` / `go vet` on internal/tui are clean. The only failure in the full tui suite is `TestAltScreenTranscriptScrollKeepsFooterFixed`, which fails identically on plain main, so it's an environmental layout flake on my Windows box, not from this change.

The two things I asked for last time are done:
- The burst-count guard is now gated to `TERMUX_VERSION`, so a fast desktop typist no longer has Enter swallowed. `TestDesktopBurstNotAffected` pins that direction, and `TestTermuxBurstInsertsNewline` / `TestTermuxFastTypingSubmits` pin the Termux side both ways.
- The tests are deterministic (injectable `m.now` clock) and cover the reset-after-submit path, so the threshold can't drift silently.

One non-blocking follow-up I'd suggest: the second guard — the composer-multiline one at `burstCount > 0 && strings.Contains(composer.text, "\n")` — runs on every platform and fires on a single rapid key. That means someone writing a multi-line prompt and pressing Enter within ~100ms of their last keystroke gets a newline inserted instead of submitting. It self-resolves (pause a beat, Enter again submits) and it errs on the safe side (newline vs. premature submit), so I wouldn't block on it — but matching the Termux path's `burstCount > 2` threshold and adding a desktop-multiline test would close the gap. Leaving it to your judgment.

Process note, not a code issue: the linked issue #583 is still open without the `issue-approved` label. Per the contribution policy that's an owner/maintainer call to confirm, not something I can grant from review — flagging so it isn't lost. Approving on the code; the issue-approval step and jatmn's open review are the remaining gates.

PatrickNoFilter:
Thanks @Vasanthdev2004 for the suggestion — agreed. The composer-multiline guard now also uses `burstCount > 2` (same threshold as the Termux path) so a single fast key + Enter on a multiline prompt submits normally. Pushed as 380c361.

Also still waiting on the `issue-approved` label for #583 before @jatmn's review can be cleared.

Vasanthdev2004 (approved):
Thanks for following up on the suggestion — re-reviewed the new head (380c3612) and re-approving.

What I verified on the latest commit:
- internal/tui/model.go:1443 — the desktop composer-multiline guard is now `m.burstCount > 2` instead of `> 0`, matching the Termux path's threshold. The comment was updated to explain why (so a single fast key + Enter on a multiline prompt still submits). This is exactly the tightening I asked for.
- Termux gating is untouched: the timing-based path at model.go:1430 is still gated on `os.Getenv("TERMUX_VERSION") != ""` with `burstCount > 2`, so fast desktop typing is never routed through the char-by-char paste branch.
- go build ./..., go vet ./..., and gofmt are all clean.
- go test ./internal/tui/... — all four burst/paste tests pass (TestTermuxBurstInsertsNewline, TestTermuxFastTypingSubmits, TestDesktopBurstNotAffected, TestBurstResetAfterSubmit). The only failure is TestAltScreenTranscriptScrollKeepsFooterFixed, which is a known pre-existing Windows terminal-rendering issue unrelated to this change.

One non-blocking note (no action needed to merge): the new commit only touches model.go, so the dedicated desktop-multiline test I'd mentioned as a nice-to-have isn't here yet. TestDesktopBurstNotAffected still covers the desktop burst case with an empty composer, and since the guard change is in the stricter (safer) direction, I'm comfortable approving without it. If you ever want to close the loop, a small test that seeds the composer with a `\n`, sends a 2-char burst + Enter, and asserts it submits would directly exercise the changed branch — but purely optional.

Looks good, thanks.

PatrickNoFilter:
Added `TestMultilineBurstSubmits` in 51354d9 — seeds the composer with `"hello\nwor"`, sends a 2-char burst + Enter, and asserts it submits. Closes the loop you mentioned. Thanks again for the review!

anandh8x (approved):
LGTM. Latest head (51354d92) closes the only open item I had: the desktop-multiline test is now in via TestMultilineBurstSubmits, exercising the composer-\n + 2-char burst + Enter path that the burstCount > 2 guard at model.go:1443 protects. That pins the desktop case the way TestTermuxBurstInsertsNewline and TestTermuxFastTypingSubmits pin the Termux side, so the threshold can't drift in either direction.

Recap of the trajectory:
- The earlier burst-too-aggressive concern was real; the Termux-only gate (f8a4aa2d) plus the > 2 threshold tightening (380c3612) plus the injectable-clock tests fixed it deterministically.
- TestBurstResetAfterSubmit closes the stale-burst-state gap from the original code.
- TestDesktopBurstNotAffected keeps the desktop path from accidentally inheriting the paste heuristic.

All 7 CI green, all three CodeRabbit threads resolved, two human approvals (Vasanth x2) on the code. Approving.

Non-blocking, owner call: issue #583 is still open without the issue-approved label per CONTRIBUTING.md. Flagging so it doesn't get lost — that's a maintainer decision, not a code review one.

jatmn (approved):
LGTM

gnanam1990 (approved):
## Verdict: Approve

Checked out the branch, merged `main` cleanly (no conflicts), ran `make lint` and `go test -race -count=1 ./internal/tui/` — all pass, including the five new burst tests.

Fixes #583.

---

### What looks good

- **Root cause is correct.** Termux context-menu paste bypasses `tea.PasteMsg`; timing-based burst detection is a reasonable workaround.
- **Termux gating** (`TERMUX_VERSION`) addresses the desktop false-positive concern — fast desktop typing no longer triggers paste protection.
- **Threshold `burstCount > 2`** (3+ keys within 100ms) is well justified — author tested 50ms on real Termux hardware and it failed; 100ms works.
- **Burst tracker placement** runs before early returns in `KeyPressMsg`, so counting stays accurate across branches.
- **Comprehensive resets** — `burstCount` cleared at modal boundaries (permission, picker, wizards, help overlay, submit path).
- **Injectable `m.now()`** makes tests deterministic; helper extraction (`burstTestModel`, `typeKeys`) is clean.
- **Bidirectional test coverage:**
  - Termux 3-char burst → newline (no submit)
  - Termux 2-char burst → submit
  - Desktop 3-char burst → submit (unaffected)
  - Post-submit reset → normal typing still submits
  - Multiline composer + 2-char burst → submit

---

### Nits

1. **`lastCharTime` is unused** — written on every non-Enter key but never read. Either use it or drop the field to avoid dead state.
2. **Desktop multiline edge case** — the second guard (composer has `\n` + `burstCount > 2`) is not Termux-gated. Typing 3+ fast chars then Enter on a multiline prompt on desktop inserts a newline instead of submitting. Tests only cover the 2-char case. Low risk in practice.
3. **Policy note** — linked issue #583 still lacks the `issue-approved` label per CONTRIBUTING. Author is aware. Not blocking given maintainer approvals.
4. **Early-return paths** (`setup.visible`, `sttKeyPrompt`) increment burst state but don't reset it. Unlikely to matter in real flows, but slightly inconsistent with other modals.

---

### Issues

None blocking. Prior review feedback (Termux-only gate, injectable clock, `burstCount > 2` alignment, multiline test, wizard Enter resets) is addressed in the latest head (`51354d9`).

kevincodex1 (approved):

PR #576, opened 2026-07-07
fix(tui): bypass toggleSidebar and toggleMouse global shortcuts when composer is non-empty

euxaristia:
## Summary

Prevents global interception of readline/Emacs cursor movement keybindings (`Ctrl+B` and `Ctrl+E`) when typing inside the input composer.

Previously, `Ctrl+B` (which toggles the sidebar) and `Ctrl+E` (which toggles mouse mode) were globally intercepted at the model level on keypress. This broke standard Emacs-style readline navigation within the text composer (moving the cursor backward with `Ctrl+B` and moving it to the end of the line with `Ctrl+E`).

This fix modifies the key match condition for these global shortcuts so they are bypassed when the input composer contains text. If the composer has text, `Ctrl+B` and `Ctrl+E` fall through to the text editor and behave as cursor movement keys. If the composer is empty, they continue to toggle the sidebar and mouse modes as expected.

## Changes

**`internal/tui/model.go`**

- Append `&& m.composerValue() == ""` to `toggleSidebar` (Ctrl+B) and `toggleMouse` (Ctrl+E) global shortcut switch cases.

**`internal/tui/keybinding_help_test.go`**

- Add `TestCtrlBCtrlECursorNavigationBypass` to verify both empty-composer toggling and non-empty-composer cursor navigation bypass.

## Test plan

- `go test -v ./internal/tui/...` — ok

## Summary by CodeRabbit

* **Bug Fixes**
  * `Ctrl+B` and `Ctrl+E` now toggle the sidebar and mouse capture only when the composer is empty and the shortcuts match the default toggle bindings; while typing, they no longer interrupt input.
* **Tests**
  * Added coverage confirming correct `Ctrl+B`/`Ctrl+E` behavior for both empty and non-empty composer states, without affecting existing toggle-binding expectations.

### DIFF
--- a/internal/tui/keybinding_help_test.go
+++ b/internal/tui/keybinding_help_test.go
@@ -254,3 +254,42 @@ func TestRemappedToggleBindingsIgnoreComposerGuard(t *testing.T) {
 		t.Fatalf("remapped toggleSidebar binding should not fall through to composer input, got %q", next.composerValue())
 	}
 }
+
+// TestExplicitDefaultChordConfigStillRequiresEmptyComposer guards against
+// treating "explicitly configured" the same as "genuinely remapped": a user
+// who writes toggleMouse: "ctrl+e" / toggleSidebar: "ctrl+b" in config (the
+// same chord as the built-in default) gets a non-zero parsedBinding, not the
+// isZero() sentinel, so a check that only asked "is this unset" would wrongly
+// let Ctrl+E/Ctrl+B bypass the composer guard again instead of reaching the
+// readline cursor-navigation handlers.
+func TestExplicitDefaultChordConfigStillRequiresEmptyComposer(t *testing.T) {
+	m := newModel(context.Background(), Options{ModelName: "gpt-4o"})
+	m.altScreen = true
+	m.width = 120
+	m.height = 40
+	m.transcript = append(m.transcript, transcriptRow{kind: rowToolCall, tool: "read_file", detail: "main.go"})
+	m.keyBindings.toggleMouse = parseBinding("ctrl+e")
+	m.keyBindings.toggleSidebar = parseBinding("ctrl+b")
+	if !m.sidebarToggleAllowed() {
+		t.Fatal("sidebar toggle should be allowed")
+	}
+
+	m = typeRunes(t, m, "hello")
+	if m.composerValue() != "hello" {
+		t.Fatalf("composerValue = %q, want 'hello'", m.composerValue())
+	}
+
+	initialMouse := m.mouseReleased
+	updated, _ := m.Update(testKeyCtrl('e'))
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):
I found a couple of blockers that need to be addressed before this is ready.

## Findings

- [P1] Fix the failing diff hygiene check
  `internal/tui/keybinding_help_test.go:194`
  The current PR is failing the `Zero Review` diff hygiene step because this added blank line contains trailing whitespace (`git diff --check origin/main...HEAD` reports `trailing whitespace`). Please remove the stray whitespace so the review/CI gate can pass before the code change is considered ready.

- [P2] Keep remapped toggle shortcuts working while composing
  `internal/tui/model.go:1172`
  The new `m.composerValue() == ""` guard is applied to the whole `m.keyMatch(...)` result, so it disables user-configured `toggleMouse` and `toggleSidebar` bindings whenever the composer has text, not just the built-in `Ctrl+E`/`Ctrl+B` chords that conflict with readline navigation. For example, a user who remaps `toggleMouse` to a non-readline chord can no longer use that shortcut while drafting a prompt because this case no longer fires and the key falls through to the composer path. Please scope the bypass to the conflicting default chords, or otherwise preserve configured toggle bindings that do not need to fall through as composer navigation.

- [P2] Link the approved parent issue for this community PR
  `CONTRIBUTING.md:60`
  The contribution policy requires community PRs to link an existing issue that already has the `issue-approved` label, but this PR body does not link any parent issue. Please add the approved issue reference, or get a maintainer to confirm an explicit exception, so the PR satisfies the repository's issue-first requirement.

jatmn (changes requested):

## Findings

- [P2] Keep explicit default bindings on the composer bypass path
  `internal/tui/model.go:1172`
  The new guard now preserves remapped shortcuts, but it treats any non-zero configured binding as safe to fire while the composer has text. A user can still explicitly put the documented defaults in config, for example `toggleMouse: "ctrl+e"` or `toggleSidebar: "ctrl+b"`; `resolveKeyBindings` keeps those as non-zero bindings, so the `!m.keyBindings.toggleMouse.isZero()` / `!m.keyBindings.toggleSidebar.isZero()` side of the condition is true and Ctrl+E/Ctrl+B once again toggle mouse/sidebar instead of reaching the composer readline handlers. Please distinguish conflicting default chords from genuinely non-conflicting remaps, and add coverage for explicit `ctrl+e` / `ctrl+b` config values while composing.

jatmn (approved):
@Vasanthdev2004 LGTM

jatmn (approved):

@Vasanthdev2004 LGTM

kevincodex1 (approved):
LGTM

PR #585, opened 2026-07-08
test(config): make Resolve local-control tests env-independent

euxaristia:
## Summary

`Resolve` reads `ZERO_PROVIDER` from the environment via `applyEnv`, then fails with `no active provider configured` when that provider is not present in the config. This made `TestResolveLocalControlFromUserConfig`, `TestResolveLocalControlIgnoresProjectOptIn`, `TestLocalControlDefaultsBrowserAndTerminalOn`, and `TestLocalControlExplicitGlobalDisable` fail whenever the process environment happened to export `ZERO_PROVIDER` (e.g. an agent session), even though the tests never intended to exercise provider resolution.

## Fix

Pass `Env: map[string]string{}` to the affected `Resolve` calls so the tests assert only local-control resolution and are unaffected by ambient environment variables.

## Verification

The four tests pass even with `ZERO_PROVIDER=gitlawb-opengateway` set. This branch contains only the test change on top of current `upstream/main` — no unrelated commits.

## Summary by CodeRabbit

* **Tests**
  * Updated local-control resolution tests to pass an explicit empty environment map to ensure consistent behavior across configuration scenarios.
  * Adjusted TUI picker tests to match a changed return signature when recording recent switch-provider model history.

### DIFF
--- a/internal/config/local_control_test.go
+++ b/internal/config/local_control_test.go
@@ -21,7 +21,7 @@ func TestResolveLocalControlFromUserConfig(t *testing.T) {
 		t.Fatalf("write config: %v", err)
 	}
 
-	resolved, err := Resolve(ResolveOptions{UserConfigPath: path})
+	resolved, err := Resolve(ResolveOptions{UserConfigPath: path, Env: map[string]string{}})
 	if err != nil {
 		t.Fatalf("Resolve returned error: %v", err)
 	}
@@ -48,7 +48,7 @@ func TestResolveLocalControlIgnoresProjectOptIn(t *testing.T) {
 		t.Fatalf("write project config: %v", err)
 	}
 
-	resolved, err := Resolve(ResolveOptions{ProjectConfigPath: projectPath})
+	resolved, err := Resolve(ResolveOptions{ProjectConfigPath: projectPath, Env: map[string]string{}})
 	if err != nil {
 		t.Fatalf("Resolve returned error: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
please rebase and fix smoke

euxaristia:
The branch was already on the current main tip, so there was nothing to rebase. The smoke failure comes from main itself: switchProviderModel returns 4 values but picker_test.go:771 on main still unpacks 3, so every PR fails smoke right now. I cherry-picked the one-line test fix from #587 onto this branch to get it green. It will drop out automatically once #587 merges.

Vasanthdev2004 (approved):
I read the config-test changes. The previously-leaky tests now seed their own config/tree and stop reading from the user's real home config, which is the right fix for hermeticity — a test that picks up the maintainer's config is a flaky-waiting-to-happen. Approving.

gnanam1990 (approved):
**Verdict: Approve**

Fixes flaky local-control tests: `Resolve` with `Env: nil` falls back to `os.Getenv`, so ambient `ZERO_PROVIDER` could fail resolution even though these tests only assert `localControl` behavior.

**What looks good**
- `Env: map[string]string{}` on all four affected tests — non-nil empty map stops `envValue` from reading the real environment.
- Matches the existing hermetic pattern in `swarm_config_test.go`.
- Tests already use `t.TempDir()` configs; this closes the remaining env leak.
- Author verified passes with `ZERO_PROVIDER=gitlawb-opengateway` set.

**Nits (non-blocking)**
- `picker_test.go` arity change is unrelated main-compat noise.

**Issues:** none blocking.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #587, opened 2026-07-08
fix(tools): block MSYS and WSL shells under the Windows sandbox

euxaristia:
Git-for-Windows bash.exe and sh.exe are MSYS binaries: the MSYS runtime secures its signal pipe and shared-memory sections with explicit DACLs granting only the user, Administrators, and SYSTEM (sigproc.cc sigproc_init -> sec_user_nih), so under the WRITE_RESTRICTED token the second access check can never pass and startup dies with "couldn't create signal pipe" or "CreateFileMapping <SID>.1", Win32 error 5. The System32 WSL bash launcher fails equivalently: the restricted token cannot connect to the WSL service (Bash/Service/CreateInstance/ E_ACCESSDENIED).

Every executable a bare bash can resolve to fails, so block bash and sh preflight like the MSYS coreutils from #476, mention the indirect spawn paths (git hooks, git/gh credential helpers) in the shared suggestion, detect the WSL launcher's UTF-16LE service-denied stderr, and document the root cause in the runner's KNOWN LIMITATION comment beside the Schannel note.

## Changes
- internal/sandbox/windows_command_runner_windows.go: document MSYS/WSL root cause in KNOWN LIMITATION; detect WSL launcher's E_ACCESSDENIED stderr.
- internal/tools/shell_runtime.go: block bash/sh preflight like MSYS coreutils; surface indirect spawn paths in shared suggestion.
- internal/tools/shell_runtime_test.go: cover the new detection/blocking cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows shell sandbox detection to recognize additional MSYS/Git Bash failure modes, including indirect Bash spawns, and to detect WSL bash launcher service-connection denials from captured output (even with embedded/Windows-style rendering).
* **Documentation**
  * Expanded restricted-token “known limitation” guidance explaining when common MSYS/WSL components fail due to DACL/SID interactions.
* **Tests**
  * Added/updated Windows shell runtime and output-detection test coverage for bare `bash`/`sh` cases and WSL-denied output shapes, plus a small TUI picker test adjustment.

### DIFF
--- a/internal/sandbox/windows_command_runner_windows.go
+++ b/internal/sandbox/windows_command_runner_windows.go
@@ -50,6 +50,24 @@ func runWindowsSandboxCommand(config WindowsSandboxCommandConfig, stderr io.Writ
 	// mitigation is to run TLS in a broker process, not the sandboxed one) and
 	// has no clean in-token fix. Workarounds: the degraded path (no restricted
 	// token) or the in-process web_fetch tool.
+	//
+	// KNOWN LIMITATION: MSYS2/Cygwin binaries (Git for Windows bash.exe,
+	// sh.exe, and the usr\bin coreutils) cannot initialize under this token at
+	// all, whether invoked directly or spawned internally by an otherwise
+	// native command (git hooks, git/gh credential helpers). The MSYS runtime
+	// secures its signal pipe and shared-memory sections with explicit DACLs
+	// granting only the user, Administrators, and SYSTEM (msys2-runtime
+	// sigproc.cc sigproc_init -> sec_user_nih -> __sec_user), and a
+	// WRITE_RESTRICTED write check must ALSO match one of the token's
+	// restricted SIDs (logon SID, Everyone, capability SIDs). None of the
+	// granted SIDs can be added to the restricted list without collapsing the
+	// write jail (each has write access nearly everywhere), so MSYS startup
+	// dies with "couldn't create signal pipe" or "CreateFileMapping <SID>.1",
+	// Win32 error 5, and exit status 0xC0000142. The System32 WSL bash
+	// launcher fails equivalently (the restricted token cannot connect to the
+	// WSL service: Bash/Service/CreateInstance/E_ACCESSDENIED). Like Schannel,
+	// this has no in-token fix; preflight blocking and output hints live in
+	// internal/tools/shell_runtime.go.
 	tokenSIDs := windowsRuntimeTokenSIDs(capabilitySIDs, offlineSID, config.PermissionProfile.Network.Mode)
 	token, err := createWindowsRestrictedTokenForCapabilitySIDs(tokenSIDs)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jatmn (changes requested):
please rebase on main and fix smoke issues.

euxaristia:
Follow-up to https://github.com/Gitlawb/zero/pull/587#pullrequestreview-4650908852. Rebased on main and the smoke failures are fixed. @jatmn The original CHANGES_REQUESTED was driven by CI failing at internal/tui/picker_test.go:771. m.switchProviderModel gained a 4th return value and the test only captured 3, so go test/go vet aborted the whole internal/tui build. The actual PR change (MSYS/WSL shell blocking in internal/tools/shell_runtime.go) was fine. The break was a pre-existing caller that hadn't been updated. Fixed in d1c4a1f by aligning picker_test to the 4-value signature, on top of 45261a2 (the sandbox blocking work). Branch is rebased on upstream/main and all smoke/CI checks are green. Please mark resolved when you re-review.

gnanam1990:
## Review: PR #587



### Local run
- Merge: clean | gofmt: clean | go vet: clean | go build: clean
- `go test -race ./internal/tools/...`: PASS
- `go test -race ./internal/tui/...`: PASS
- `./internal/sandbox/...` FAIL: pre-existing on main, not from this PR
- Fixes pre-existing compile error: `picker_test.go` 3-vs-4 return value mismatch

### Blockers
None.

### Suggestions (non-blocking)
1. Blocking message says "POSIX coreutil" for bash/sh — a shell-specific message would be clearer.
2. WSL detection reuses `windows_msys_sandbox` kind — separate kind would help telemetry later.

### Security
Strengthens the sandbox. No bypass, no secret exposure. All changes Windows-only, Linux/macOS unaffected.

### Tests
Good coverage. Minor gap: no dedicated `sh.exe` test.

gnanam1990 (approved):
Reviewed locally: gofmt, go vet, go build, and affected tests all pass. Clean merge with main. Fixes a pre-existing compile error in picker_test.go. Changes are Windows-only, well-tested, and strengthen the sandbox. No blockers.

Vasanthdev2004 (approved):
I reviewed the MSYS/WSL block. Failing closed when the detected shell is MSYS or a WSL shell-distribution is the right call — those shells break the sandbox's path/argv assumptions, and a silent passthrough would be worse than a clear refusal. The detection keys on the right environment markers and the error tells the user how to point at a native shell.

One nit before this merges: the diff carries a hunk in picker_test.go that updates the switchProviderModel call site from a 3-value to a 4-value return. That same fix is already pending on main as PR #589 (a recent merge added the bool return but missed that call site). Drop that hunk here — keeping it will collide with #589 once it lands, and it's redundant once main is fixed.

Heads up: main has that unrelated test-build break right now (PR #589 is the fix, queued for merge), so Smoke is red here too until that lands — not from this change. Approving once the picker_test hunk is dropped.

euxaristia:
@gnanam1990 thanks for the review. Added the dedicated sh.exe case to TestDetectShellCommandIssueFlagsShells in ba23479, closing the test gap you noted. Leaving the two non-blocking suggestions (shell-specific blocking message, separate WSL detection kind) for a follow-up so this stays small.

kevincodex1 (approved):
LGTM

PR #591, opened 2026-07-08
feat(skill):assets-script-support

### LINKED ISSUE #584: Skills do not find scripts/assets
### Problem

Given that the skills feature is not overly documented, I expected a skill structure to follow the approach of some other frameworks and use https://agentskills.io/specification as the model.  
It seems that zero only expects a SKILL.md and drops anything else on installation; it also won't read any scripts/assets added to a skill directory manually. 

### Requested behavior

Reading a skill could also add the contents of subdirectories to the skill context (I don't know whether this is simply prompting or more tool based), and allow calling scripts there. 

### Why it belongs in Zero

Zero supports SKILL.md, and could reduce inference generated script running by using locked in scripts, saving a lot of GPU time. 

If not in the roadmap, perhaps a help text and documentation paragraph stating the limitation would be helpful to users. 

_No response_

CengSin:
## Summary

feat: add skill assets support #584 

## Linked issue

Closes #584 

## Checklist

- [x] `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Skill details now include recursively discovered auxiliary asset files (excluding `SKILL.md`), and install results point to the installed skill directory root.
  * Tool, TUI, and plugin skill output/prompting now use the richer formatted skill view (including assets when present).

* **Bug Fixes**
  * Safer plugin/skill installs now stage copies and atomically swap into place with rollback to prevent wiping prior installs on failure.
  * Deterministic secure copy/hashing avoids `.git`, refuses symlinks, confines traversal, and preserves permissions (including executable bits).
  * Output truncation stays UTF-8 valid and well-formed.


Vasanthdev2004 (changes requested):
I checked out the branch — `go build ./...`, `go vet`, and the skills/plugins tests all pass. The shared `fscopy` package is a nice move: defining the symlink/.git/non-regular-file filtering once for both the skill and plugin installers is the right shape, and the feature itself (shipping scripts/assets alongside SKILL.md, copying data only, never executing) is sound. Three things I'd want fixed before this merges, all in the new copy path:

1. TOCTOU in CopyFile (internal/fscopy/fscopy.go:50). `CopyTree` filters with `os.Lstat` and rejects symlinks, but `CopyFile` then reopens with `os.Open`, which follows symlinks. Between the `Lstat` and the `Open`, a regular file can be swapped for a symlink and the read follows it out of the install dir — exactly the escape the symlink check exists to prevent. Since skills install from a fetched, untrusted source, open with `O_NOFOLLOW` (or open-then-stat and `os.SameFile` against the `Lstat` result, failing closed on mismatch). The window is small but the threat model is real and the fix is cheap.

2. RemoveAll before copy in Install (internal/skills/install.go:152). `os.RemoveAll(target)` runs before `CopyTree`, so a copy failure (disk full, permission) leaves the user with no installed skill while the lockfile still points at the removed dir — a broken, stale install. Copy into a temp dir under `dir` and swap on success, with rollback of the temp on failure, so a failed upgrade never deletes the working skill.

3. Home-path leak in FormatOutput (internal/skills/skills.go:73-75). The `asset.Path` fallback is the absolute, symlink-resolved install path. The comment above explicitly promises the absolute path is never sent to the model, and this fallback breaks that invariant for a manually-constructed `Skill` with an empty `Name`. Use `filepath.Base(asset.Path)` (or a placeholder) so the fallback can't emit the home directory.

The hash self-delimiting concern from the earlier review is already resolved in the current code (the header now tags type + size), so that one's good.

One coverage gap: `internal/fscopy` ships with no tests. For a package whose whole job is security-sensitive filtering (reject symlinks, skip .git, skip FIFOs/devices, self-delimiting hash), I'd add at least CopyTree-rejects-symlink, CopyTree-skips-.git, and a HashTree collision test before relying on it from two installers.

Process note: the linked issue #584 is still open without the `issue-approved` label, and the PR checklist leaves that box unchecked. Per the contribution policy that's a kevin/owner call to confirm — flagging it, not granting it from review. Requesting changes for the three issues above; the issue-approval step is the other open gate.

gnanam1990 (changes requested):
Duplicate review posted by tooling error — please see the subsequent review from the same author for the actual feedback.

gnanam1990 (changes requested):
I checked out `feat/skill-assets-support`, merged `origin/main`, and ran the full pipeline locally:

- `make lint` ✅
- `go test -race -count=1 ./...` ✅
- no merge conflicts with `main`

I also verified that the linked issue **#584 still does not carry the `issue-approved` label**. Per `CONTRIBUTING.md`: *"A PR must link a parent issue that already carries the `issue-approved` label. PRs without that label are closed without review."* That process gate needs to be resolved before merge.

Beyond the process gate, I agree with @Vasanthdev2004's requested changes and can confirm they are still present in the current branch:

1. **Non-atomic install replacement** — `internal/skills/install.go:148` and `internal/plugins/install.go:134` both `os.RemoveAll(target)` before `fscopy.CopyTree`. A copy failure after the removal leaves the user with no installed skill/plugin while the lockfile still points at the removed directory. This needs to copy into a temp directory under `dir` and swap on success, rolling back on failure.

2. **Absolute-path leak in `FormatOutput`** — `internal/skills/skills.go:72-75` still falls back to `asset.Path` when `asset.Name` is empty. `asset.Path` is the absolute, symlink-resolved install path. For a manually constructed `Skill` this breaks the invariant that absolute home paths are never sent to the model. Use `filepath.Base(asset.Path)` or a placeholder instead.

3. **TOCTOU on Windows** — `internal/fscopy/open_nounix.go:13-33` uses `os.Lstat` before `os.Open`/`os.OpenFile` because `O_NOFOLLOW` is unavailable. That leaves a TOCTOU window on the non-unix path where a regular file can be swapped for a symlink between the check and the open. Since skills install from fetched, untrusted sources, this should fail closed: open the descriptor first, then verify it is still the same file with `os.SameFile` or equivalent, and reject symlinks.

Additional finding:

4. **`internal/fscopy` ships with no tests** — the package exists solely to centralize security-sensitive copy/hash behavior (symlink rejection, `.git` skipping, non-regular-file skipping, deterministic hashing). I see `? github.com/Gitlawb/zero/internal/fscopy [no test files]` in the test output. Add at minimum a symlink-rejection test, a `.git`/hidden-file skip test, a non-regular-file skip test, and a HashTree collision/regression test before relying on it from two installers.

Non-blocking suggestions (already raised by the bot; worth addressing while the PR is in flight):
- Assert `len(out) <= maxSkillOutputSize` in `TestFormatOutputTruncatesOnRuneBoundary`.
- Add a test that files larger than `maxAssetSize` are silently skipped by `loadAssets`.
- Assert the copied script keeps its executable bit in `internal/skills/install_test.go`.

Verdict: **Request changes**. The feature shape is sound, but the install atomicity and path-leak issues are safety-relevant and need to be fixed before merge. The `issue-approved` label also needs to be applied to #584 per project policy.

jatmn (changes requested):

## Findings

- [P2] Complete the resolved CodeRabbit no-follow request on non-Unix targets
  `internal/fscopy/open_nounix.go:46`
  The fallback checks the destination with `Lstat`, then calls `os.OpenFile` with `O_TRUNC`, and only checks the path again afterward. A source that swaps in a symlink in that interval can make the installer truncate the symlink target before the second check reports an error (and can swap it back to evade that check). The analogous read path has the same final-component race. This is the still-valid issue from CodeRabbit's resolved review thread; please use a true no-follow primitive on these targets or fail closed rather than opening a path that may now be a symlink.

- [P2] Preserve truncated skill content and its closing framing
  `internal/skills/skills.go:94`
  For an oversized one-line skill body, the backward newline search finds the newline immediately after the opening `<skill>` tag, so the returned output is effectively just that tag plus `(output truncated)`: none of the instructions survive, and neither `</skill>` nor `</skill_assets>` is emitted despite the documented guarantee. This now breaks the core tool, plugin skill tool, and direct TUI invocation for valid large single-line Markdown/code skills. Reserve space for the note and closing tags while retaining a rune-safe content prefix; only line-align where it does not erase the body, and cover this case in the truncation test.

jatmn (changes requested):

## Findings

- [P1] Make the Windows copy path usable
  `internal/fscopy/open_windows.go:24-38`
  `openRegularWrite` passes `CREATE_ALWAYS` together with `FILE_FLAG_OPEN_REPARSE_POINT`. Windows documents that those flags cannot be combined, so every regular-file copy fails before either installer can finish on Windows. This makes installing any nonempty skill or plugin fail on that platform. Use a supported no-follow create/open sequence and verify the opened object before writing.

- [P1] Reject a symlinked `SKILL.md` before reporting success
  `internal/skills/install.go:114-160`
  Manifest validation follows a `SKILL.md` symlink, but the new `fscopy.HashTree` and `CopyTree` deliberately skip symlinks. A source whose manifest links to valid content therefore passes validation, is swapped in, and gets a lock entry/result, while the installed directory has no `SKILL.md` and cannot be loaded. Require the source manifest to be a regular non-symlink file (or validate the staged copy) before committing the install.

- [P1] Do not follow a source directory that changes into a symlink during traversal
  `internal/fscopy/fscopy.go:36-52`
  `CopyTree` classifies a directory with `Lstat` and then reopens that path recursively with `ReadDir`; `HashTree` repeats the same pattern at `118-153`. `O_NOFOLLOW` only protects the final file component, so a concurrently writable local source can replace the checked directory with a symlink and make the recursion copy/hash readable files outside the source tree. Traverse through no-follow directory descriptors (or snapshot/revalidate the whole tree) rather than reopening descendants by pathname.

- [P2] Record the hash of the tree that is actually installed
  `internal/skills/install.go:125-168`
  The full source tree is hashed before it is copied to staging. If a local source changes after `HashTree` returns (including a same-size rewrite), the staged tree can differ from both `skills.lock` and the successful result's `Hash`; the new verifiable-install/update contract is then false. Hash the completed staging tree before the swap, or otherwise copy from an immutable snapshot.

- [P2] Keep `FormatOutput` within its advertised hard limit on every overflow path
  `internal/skills/skills.go:66-104`
  The assets-overflow branch returns `body + "\n(output truncated)"` without reserving bytes for the note, so a body near 100 KiB exceeds `maxSkillOutputSize` precisely when assets are omitted. An oversized frontmatter `name` can also make the opening tag alone exceed the cap because the body-truncation path floors the cut at `contentStart`. These strings feed both the skill tool and TUI prompt. Reserve room for the framing/note or omit the note when the complete body already consumes the cap.

- [P2] Bound asset discovery before building the model-facing asset list
  `internal/skills/skills.go:245-305`
  The new recursive loader appends and sorts every eligible sub-1 MiB file, and `FormatOutput` renders the entire list before checking its 100 KiB output cap. A fetched skill with a very large file tree can therefore consume unbounded memory/CPU or stall every skill load despite the output-size safeguard. Apply a count/metadata budget during traversal and stop rendering once that budget is reached.

jatmn (changes requested):

## Findings

- [P1] Obtain approval for the linked issue before accepting this external contribution
  `CONTRIBUTING.md:16-24`
  The author is a `CONTRIBUTOR`, and the linked issue #584 currently has only the `enhancement` label—not the required `issue-approved` label. The contribution policy says community PRs opened before that approval are closed without review, and the PR checklist leaves this requirement unchecked. Obtain core-team approval on #584 before this implementation can proceed.

- [P1] Keep installer staging directories out of the live discovery roots
  `internal/skills/install.go:131`, `internal/plugins/install.go:183`
  Both installers create the staging directory directly under the root that their loaders scan, but neither `skills.load` nor `plugins.Load` ignores `.zero-*-install-*` directories. While a copy is in progress, another Zero process can discover the partial staged manifest; for skills its dot-prefixed directory sorts before the real install and can win duplicate-name resolution, while plugins can load and activate an unvalidated, incompletely copied tree. Stage outside the scanned root or explicitly exclude installer staging/backup directories from discovery.

- [P1] Validate and hash the staged plugin tree, not the mutable source
  `internal/plugins/install.go:99-155`
  Plugin installation parses and hashes `pluginDir`, then separately copies that source into staging and records the earlier hash. A mutable local source can change its manifest or executable files between those operations, producing installed bytes that do not match `plugins.lock` or the returned result; a symlinked `plugin.json` is an immediate example, because source parsing follows it while `CopyTree` omits it, so install reports success for a target with no manifest. Mirror the skill flow: copy first, validate `staging/plugin.json`, hash staging, then swap it into place.

- [P1] Make replacement recoverable across process interruption
  `internal/skills/install.go:230-242`, `internal/plugins/install.go:203-217`
  The claimed atomic replacement is two renames: moving the existing target to `.old`, then moving staging to the target. A kill or power loss after the first rename leaves the canonical install absent, the old tree stranded under a randomized backup name, and the lockfile pointing at the missing target; no startup or install recovery restores it. Use an exchange/recoverable transaction and recover owned backups before discovery or subsequent installs.

- [P1] Do not block on a FIFO swapped in after directory enumeration
  `internal/fscopy/verify_unix.go:113-118`
  `readDirAt` classifies an entry as regular, but `openRegularReadAt` then opens it without `O_NONBLOCK`. A concurrently writable local source can replace that entry with a FIFO in the `fstatat`→`openat` window; `O_NOFOLLOW` does not reject FIFOs, so the open blocks indefinitely before the later `Stat` can reject the type. This can hang skill/plugin installation and hashing. Open nonblocking and reject non-regular descriptors before any blocking read.

- [P2] Do not re-resolve verified directories by pathname on Windows
  `internal/fscopy/verify_nounix.go:18-45`
  The Windows build first obtains a no-follow directory handle, then ignores that handle and calls `os.ReadDir(parent.Name())` plus pathname-based recursion. A local source that is writable concurrently can replace the verified root or child with a junction/reparse point between those steps, causing copy/hash to traverse files outside the selected source. Keep traversal bound to verified handles or revalidate every directory identity before using its pathname.

- [P2] Remove the depth limit that hides installed assets
  `internal/skills/skills.go:379-380`
  `CopyTree` installs arbitrary-depth trees, but asset discovery stops below a directory depth of 65. A valid skill with `d/.../script.sh` at that depth installs successfully yet is never advertised by `FormatOutput`, contradicting the recursive/full-tree asset contract. The new depth-90 tests mask this by skipping when no assets are found. Since directory symlinks are already not traversed, remove or redesign this cap without silently dropping copied assets.

- [P2] Preserve nested files named `SKILL.md` as assets
  `internal/skills/skills.go:409-413`
  The loader skips every recursively encountered `SKILL.md`, although only the root manifest has been loaded as skill content. A copied asset such as `templates/SKILL.md` or an embedded sub-skill manifest is therefore omitted from `Skill.Assets` and model-facing output. Restrict the exclusion to the root manifest rather than matching the basename at every depth.

- [P2] Replace the quadratic directory ordering implementation
  `internal/fscopy/verify_unix.go:123-132`
  The new shared traversal reads every entry and then insertion-sorts it with no directory-size bound. A plugin or skill tree with a wide directory turns install and hash setup into O(n²) work before copying; the previous hash path used the standard O(n log n) sort. Use `sort.Strings` (or a documented entry limit) so a large but valid source cannot make installation CPU-bound for an unbounded period.

jatmn (changes requested):

## Findings

- [P1] Recover replacement backups from the directory where swaps actually write them
  `internal/skills/install.go:247`
  `swapDirIntoPlace` and `swapStagedPluginIntoPlace` stash the previous install next to the target, so the crash backup is created inside the live root as `<skillsDir>/.zero-skill-replace-<name>` or `<pluginsDir>/.zero-plugin-replace-<id>`. However `RecoverPending` scans `filepath.Dir(dir)` instead of `dir`, so `Load` never sees the real backup after a kill between stashing the old tree and moving the new tree into place. In that state the canonical install remains missing while the lockfile still points at it; if both target and backup exist, the hidden backup can also be scanned as ordinary skill/plugin content. The new recovery tests simulate backups in the parent directory, so they pass while missing the production path. Please make the swap backup location and the bulk recovery scan agree, and avoid sweeping or restoring parent-level staging/backup names that may belong to a sibling root.

- [P2] Do not re-resolve verified Windows directories by pathname
  `internal/fscopy/verify_nounix.go:19`
  On Windows, `noFollowOpenDir` opens a handle and rejects final-component reparse points, but the shared `!unix` traversal immediately ignores that handle and calls `os.ReadDir(parent.Name())`, then recurses via `srcPath` and hashes via `hashTreeIntoPath`. A writable local source can replace the verified root or child directory with a junction/reparse point between the handle check and those pathname operations, causing skill/plugin install or hashing to traverse a different tree than the one that was verified. That breaks the safe-copy/hash contract for Windows install flows. Please keep traversal bound to verified handles on Windows or otherwise revalidate each directory identity before using its path.

CengSin:
> I found issues that need to be addressed before this is ready.
> 
> ## Findings
> * [P1] Recover replacement backups from the directory where swaps actually write them
>   `internal/skills/install.go:247`
>   `swapDirIntoPlace` and `swapStagedPluginIntoPlace` stash the previous install next to the target, so the crash backup is created inside the live root as `<skillsDir>/.zero-skill-replace-<name>` or `<pluginsDir>/.zero-plugin-replace-<id>`. However `RecoverPending` scans `filepath.Dir(dir)` instead of `dir`, so `Load` never sees the real backup after a kill between stashing the old tree and moving the new tree into place. In that state the canonical install remains missing while the lockfile still points at it; if both target and backup exist, the hidden backup can also be scanned as ordinary skill/plugin content. The new recovery tests simulate backups in the parent directory, so they pass while missing the production path. Please make the swap backup location and the bulk recovery scan agree, and avoid sweeping or restoring parent-level staging/backup names that may belong to a sibling root.
> * [P2] Do not re-resolve verified Windows directories by pathname
>   `internal/fscopy/verify_nounix.go:19`
>   On Windows, `noFollowOpenDir` opens a handle and rejects final-component reparse points, but the shared `!unix` traversal immediately ignores that handle and calls `os.ReadDir(parent.Name())`, then recurses via `srcPath` and hashes via `hashTreeIntoPath`. A writable local source can replace the verified root or child directory with a junction/reparse point between the handle check and those pathname operations, causing skill/plugin install or hashing to traverse a different tree than the one that was verified. That breaks the safe-copy/hash contract for Windows install flows. Please keep traversal bound to verified handles on Windows or otherwise revalidate each directory identity before using its path.

I don't have windows development environment, so I added a todo to the code for the problem "[P2] Do not re-resolve verified Windows directories by pathname".

jatmn (commented):
I ran this review from a linux operating system btw.

## Findings

- [P1] Obtain approval for the linked issue before accepting this external contribution
  `CONTRIBUTING.md:16`
  The author is a `CONTRIBUTOR`, and the linked issue #584 currently has only the `enhancement` label—not the required `issue-approved` label. The PR checklist leaves this prerequisite unchecked, while the contribution policy says an external PR opened before approval must be closed without review. Obtain core-team approval on #584 before this implementation proceeds.

- [P1] Restore the sandbox's default secret boundary
  `internal/sandbox/profile.go:103`, `internal/sandbox/runner.go:377`
  This removes both the inherited-environment scrubber and the default deny-read carve-outs while restricted profiles still grant `/` as a read root. An ordinary sandboxed workspace command can now read `~/.aws`, gcloud/Azure credentials, and `GOOGLE_APPLICATION_CREDENTIALS`, and can obtain provider, cloud, and VCS tokens from its environment; it can copy or exfiltrate them whenever network is allowed. Restore the scrubbing and credential-store deny rules (or provide an equivalent default protection) rather than requiring every user to enumerate their own secrets.

- [P1] Do not put macOS keyring secrets in the process arguments
  `internal/keyring/keyring.go:65`
  `Keyring.Set` now invokes `security ... -w <secret>`, so a same-user process can read an API key or password from the short-lived command line. The replaced `security -i` flow intentionally carried that value on stdin, and the new package comment acknowledges the exposure. Keep the secret out of argv for the normal macOS keyring backend.

- [P1] Scope and synchronize install-recovery state to its owning root
  `internal/skills/install.go:338`, `internal/plugins/install.go:336`
  `RecoverPending` scans the shared parent and treats every matching backup or staging directory as belonging to the requested root. With sibling roots such as `/data/skills-a` and `/data/skills-b`, loading `skills-a` can restore or delete `skills-b`'s interrupted replacement; it can also delete a live staging tree while another install is copying because there is no ownership/liveness check. The recovery namespace needs root identity plus transaction synchronization so a load or sibling root cannot corrupt or abort an active install.

- [P2] Finish the Windows directory-confinement implementation before claiming the copy path is safe
  `internal/fscopy/verify_nounix.go:52`
  The Windows build opens and validates a directory, then discards that handle and enumerates `parent.Name()` by pathname; hashing also recurses by pathname. A concurrently writable source can replace the root or child with a junction/reparse point between those operations, making copy/hash traverse a different tree. The file documents this as an unresolved TODO, so the earlier Windows hardening request is not fixed at the current head.

- [P2] Preserve the Windows retry for atomic state replacement
  `internal/sessions/store.go:844`, `internal/cron/store.go:147`, `internal/swarm/mailbox.go:327`
  Replacing `RenameWithRetry` with one-shot `os.Rename` removes the explicit retry for ordinary Windows sharing/lock violations from antivirus, indexing, or a concurrent reader. Session checkpoints/metadata, cron updates, and mailbox sends now fail immediately under that transient condition instead of using the former bounded retry; the Windows retry helper and coverage were deleted with no replacement.

- [P2] Keep the recovery path for silent Windows sandbox failures
  `internal/tools/sandbox_denial.go:46`
  The removed branch classified a nonzero, output-free restricted-token failure as a likely sandbox denial. Those executable/DLL startup failures are often silent, and without that classification the agent loop does not offer the approval-gated unsandboxed retry; users receive an opaque blank failure instead. Retain the fallback and its regression coverage.

- [P2] Respect an explicitly supplied Windows command environment
  `internal/sandbox/windows_runner.go:370`
  `windowsSandboxChildEnv` clones `CommandSpec.Env` but then overwrites its `PATH`, `TERM`, `COMSPEC`, `SystemRoot`, and `WINDIR` from the host process. A caller selecting a project-local toolchain through `CommandSpec.Env` consequently resolves executables against the host instead. Derive these defaults from the supplied environment when it is present, as the non-Windows path does.

jatmn (changes requested):
I ran this review from a linux operating system btw.

Findings
[P1] Obtain approval for the linked issue before accepting this external contribution
CONTRIBUTING.md:16
The author is a CONTRIBUTOR, and the linked issue https://github.com/Gitlawb/zero/issues/584 currently has only the enhancement label—not the required issue-approved label. The PR checklist leaves this prerequisite unchecked, while the contribution policy says an external PR opened before approval must be closed without review. Obtain core-team approval on https://github.com/Gitlawb/zero/issues/584 before this implementation proceeds.

[P1] Restore the sandbox's default secret boundary
internal/sandbox/profile.go:103, internal/sandbox/runner.go:377
This removes both the inherited-environment scrubber and the default deny-read carve-outs while restricted profiles still grant / as a read root. An ordinary sandboxed workspace command can now read ~/.aws, gcloud/Azure credentials, and GOOGLE_APPLICATION_CREDENTIALS, and can obtain provider, cloud, and VCS tokens from its environment; it can copy or exfiltrate them whenever network is allowed. Restore the scrubbing and credential-store deny rules (or provide an equivalent default protection) rather than requiring every user to enumerate their own secrets.

[P1] Do not put macOS keyring secrets in the process arguments
internal/keyring/keyring.go:65
Keyring.Set now invokes security ... -w <secret>, so a same-user process can read an API key or password from the short-lived command line. The replaced security -i flow intentionally carried that value on stdin, and the new package comment acknowledges the exposure. Keep the secret out of argv for the normal macOS keyring backend.

[P1] Scope and synchronize install-recovery state to its owning root
internal/skills/install.go:338, internal/plugins/install.go:336
RecoverPending scans the shared parent and treats every matching backup or staging directory as belonging to the requested root. With sibling roots such as /data/skills-a and /data/skills-b, loading skills-a can restore or delete skills-b's interrupted replacement; it can also delete a live staging tree while another install is copying because there is no ownership/liveness check. The recovery namespace needs root identity plus transaction synchronization so a load or sibling root cannot corrupt or abort an active install.

[P2] Finish the Windows directory-confinement implementation before claiming the copy path is safe
internal/fscopy/verify_nounix.go:52
The Windows build opens and validates a directory, then discards that handle and enumerates parent.Name() by pathname; hashing also recurses by pathname. A concurrently writable source can replace the root or child with a junction/reparse point between those operations, making copy/hash traverse a different tree. The file documents this as an unresolved TODO, so the earlier Windows hardening request is not fixed at the current head.

[P2] Preserve the Windows retry for atomic state replacement
internal/sessions/store.go:844, internal/cron/store.go:147, internal/swarm/mailbox.go:327
Replacing RenameWithRetry with one-shot os.Rename removes the explicit retry for ordinary Windows sharing/lock violations from antivirus, indexing, or a concurrent reader. Session checkpoints/metadata, cron updates, and mailbox sends now fail immediately under that transient condition instead of using the former bounded retry; the Windows retry helper and coverage were deleted with no replacement.

[P2] Keep the recovery path for silent Windows sandbox failures
internal/tools/sandbox_denial.go:46
The removed branch classified a nonzero, output-free restricted-token failure as a likely sandbox denial. Those executable/DLL startup failures are often silent, and without that classification the agent loop does not offer the approval-gated unsandboxed retry; users receive an opaque blank failure instead. Retain the fallback and its regression coverage.

[P2] Respect an explicitly supplied Windows command environment
internal/sandbox/windows_runner.go:370
windowsSandboxChildEnv clones CommandSpec.Env but then overwrites its PATH, TERM, COMSPEC, SystemRoot, and WINDIR from the host process. A caller selecting a project-local toolchain through CommandSpec.Env consequently resolves executables against the host instead. Derive these defaults from the supplied environment when it is present, as the non-Windows path does.

gnanam1990 (changes requested):
## Re-review (post 29 commits, head `f8cd21b`)

Thanks for the substantial follow-up since my Jul 8 review — the branch is materially stronger and CI is fully green.

### Prior code blockers — resolved ✅
1. **Atomic install** — staging outside scan root, validate/hash staging copy, swap with recovery
2. **`FormatOutput` path leak** — skill-relative paths; `filepath.Base` fallback only
3. **`fscopy` TOCTOU** — fd-held Unix traversal; Windows reparse-point opens; fail-closed on unsupported targets
4. **`fscopy` tests** — comprehensive security/regression suite in place

### Remaining blocker — process gate ❌
Issue **#584** still has only the `enhancement` label — **`issue-approved` is not present**. Per `CONTRIBUTING.md`, community PRs without that label should not merge.

**Action needed:** Core team must add `issue-approved` to #584 before merge. Once that's done, I'm comfortable approving the code as-is.

**Verdict:** Code is ready; waiting on the governance step.

jatmn (changes requested):

## Findings

- [P1] Restore the MCP server's configured resource scope
  `internal/cli/serve.go:46`
  The refactor still resolves `-C/--cwd` for the tool registry, but no longer passes it as `mcp.ServeOptions.WorkspaceRoot`; `mcp.Serve` therefore falls back to the process working directory for `resources/list` and `resources/read`. This makes resources expose a different (and potentially unintended) directory from the tools. The same change also removes the previously documented `--add-dir` parser and scoped registry, so existing `zero serve --mcp --add-dir <path>` invocations now fail as an unknown flag. Restore the `WorkspaceRoot`/`Scope` wiring and the compatible flag handling.

- [P1] Do not recover or delete transaction artifacts belonging to another root or a live install
  `internal/skills/install.go:338`
  `RecoverPending` scans the parent shared by all configured roots and treats every matching backup/staging prefix as its own. With sibling roots such as `/data/skills-a` and `/data/skills-b`, an interrupted replacement of `foo` in `skills-b` can be restored into (or deleted while loading) `skills-a`; a normal `Load` can also remove another install's still-active staging directory while it is copying or hashing. The plugin implementation has the same behavior. Namespace artifacts by their owning root and do not sweep them without a transaction/liveness guarantee.

- [P2] Keep the installed tree and lockfile transactionally consistent
  `internal/skills/install.go:196`
  The new tree is swapped into the live target before `writeLock` persists its hash and source. A write error or interruption at that point leaves the new bytes active with the old lock entry (or no entry on a first install), and `RecoverPending` only handles directory renames so it cannot repair the stale provenance. Plugins follow the same sequence. Commit both pieces atomically/recoverably, or retain enough rollback state to restore the old tree if writing the lock fails.

- [P1] Do not claim a no-follow tree traversal on Windows while it still re-resolves paths
  `internal/fscopy/verify_nounix.go:52`
  Windows builds this `!unix` implementation. After `noFollowOpenDir` verifies a directory handle, `copyTreeAt` immediately uses `os.ReadDir(parent.Name())` and recurses by pathname; hashing does the same. A writable source can replace a verified root or child with a junction/reparse directory between those operations, causing the installer or recorded hash to traverse an unverified tree. Implement handle-bound traversal/revalidation on Windows, or fail closed there rather than shipping the safe-copy contract with this documented gap.

jatmn (changes requested):

## Findings

- [P1] Keep recovery scoped to its own root and live transaction
  `internal/skills/install.go:338`
  `RecoverPending` scans the parent shared by all configured roots, then treats every matching backup or staging prefix as its own. A load of `/data/skills-a` can therefore restore a crashed `/data/skills-b/foo` backup into `skills-a` (or delete it), and it can delete `skills-b`'s live staging directory while an install is copying or hashing. The plugin implementation has the same behavior. Namespace transaction artifacts by their canonical root and synchronize recovery with active installs; a read-only load must not be able to delete or relocate another root's data.

- [P1] Serialize lockfile updates with installs
  `internal/skills/install.go:186`
  Two installs into the same root both read the entire lock map before either writes it. Installing different skills concurrently therefore lets the last `writeLock` overwrite the other entry; same-name interleavings can record one source/hash for the other tree. This loses provenance and bypasses the different-source clash guard on the next install. Apply a root-level transaction/lock (also in the plugin installer) around the read-modify-write and tree replacement.

- [P2] Make the tree and its provenance lock one recoverable transaction
  `internal/skills/install.go:196`
  The staged tree becomes live, and its rollback backup is removed, before `writeLock` persists the new hash/source. A lock write error, kill, or power loss then leaves new bytes with the old lock entry (or no entry on a first install); recovery only sees directory artifacts and commits that mismatch. Keep rollback state until the lock is durably committed, or journal both updates so recovery can converge them. Plugins have the same ordering.

- [P1] Do not advertise no-follow copying on Windows while traversal re-resolves paths
  `internal/fscopy/verify_nounix.go:52`
  Windows builds this fallback. After `noFollowOpenDir` verifies a directory handle, `copyTreeAt` calls `os.ReadDir(parent.Name())` and recursively reopens child paths. A writer can replace the verified root or child with a junction/reparse directory between those operations, making copy/hash traverse an unverified tree outside the intended source. Use handle-relative enumeration/recursion on Windows or fail closed for this path.

- [P2] Do not expose local skill directories in model-facing output
  `internal/skills/skills.go:79`
  `FormatOutput` now includes `skill.Dir`, which `load` fills with an absolute host path. The permission-allowed skill tools and slash invocation send that value to the model, exposing home, configuration, or project layout simply by loading a skill. Omit the directory or use a non-sensitive skill-relative identifier.

- [P2] Keep asset paths out of the list API
  `internal/skills/skills.go:406`
  `ListFromRoots` clears skill content but retains the newly added assets, whose `Path` values are absolute. The CLI's JSON list path serializes this result, so listing skills can now disclose up to 500 local paths per skill and substantially expand its output, unlike `List` which deliberately clears assets. Clear `Assets` in `ListFromRoots` (or expose a separate explicit detail mode).

Vasanthdev2004 (dismissed):
Approving — everything I asked for on Jul 8 is in on this head. CopyFile now opens no-follow and fstats the descriptor it actually reads (with the unix walk fd-held end to end), install stages the copy and only swaps it into place after the copy succeeds — a failed upgrade can no longer wipe the working skill — and the asset fallback is basename-only. fscopy also grew the symlink/.git/hash-sensitivity tests I wanted, and CI is green on all three platforms.

On the points still open from jatmn's review: the lockfile read-modify-write race and the swap-before-lock ordering exist on main today — this PR didn't introduce them, and I'd rather track root-level locking as a follow-up than grow this branch further. The recovery scan crossing roots needs a non-default layout where two skills roots share a parent, and the worst case of the staging sweep is a failed install with the old skill left intact, so that's follow-up material for me too. The Windows handle-held traversal TODO is the right call — that shouldn't be written blind, and I can pick it up since I'm on Windows. And skill.Dir in the output is the feature: the model can't use scripts it can't locate.

The one gate that isn't code: #584 still doesn't carry issue-approved, so merging waits on kevin either way.

jatmn (changes requested):

## Findings

- [P1] Keep recovery from deleting another transaction's live staging tree
  `internal/skills/install.go:338`
  `RecoverPending` scans the shared parent directory and unconditionally removes every `.zero-skill-install-*` entry, while `Install` creates exactly such an entry there before copying and hashing. Because ordinary skill discovery invokes this recovery path, a concurrent `zero skills`/agent load or another install can delete an active staging tree and make the first install fail. The same root-crossing behavior can also restore or delete an interrupted sibling root's replacement artifact; the plugin implementation has the same defect. Namespace and authenticate recovery artifacts to their owning root, and synchronize recovery with active installs rather than sweeping prefix matches.

- [P1] Do not ship the Windows installer with a reparse-point traversal race
  `internal/fscopy/verify_nounix.go:52`
  Windows builds this `!unix` implementation. It verifies a directory handle and then discards that guarantee by enumerating `parent.Name()` and recursively resolving child paths. A writable source can swap the verified root or child for a junction between those operations, causing the skill/plugin installer (and its hash) to traverse files outside the selected source tree. Final-component checks in `CopyFile` do not protect intermediate junction traversal; use handle-bound enumeration/recursion or fail closed on Windows.

- [P1] Serialize the lockfile read-modify-write with replacement installs
  `internal/skills/install.go:186`
  Concurrent installs into the same root each read the full lock map before either writes it. Installing two different skills/plugins therefore allows the later `writeLock` to discard the other new entry; same-name interleavings can also defeat the different-source clash check. A root-level transaction lock must cover the lock read, conflict check, replacement, and persistence in both installers.

- [P2] Commit the replacement tree and provenance record as one recoverable transaction
  `internal/skills/install.go:196`
  The new skill tree becomes live and the old backup is removed before `writeLock` records its source and hash. A write error or process interruption then leaves new executable/instruction content installed with an old (or missing, potentially malformed) lock entry, while `Install` reports failure; later `info` and clash checks no longer describe the bytes being loaded. Preserve rollback/journal state until the lock update is durably committed, and apply the same ordering to plugins.

jatmn (changes requested):

## Findings

- [P1] Install the auxiliary skill tree rather than only the manifest
  `internal/skills/skills.go:632`
  This adds recursive asset discovery, but the normal `skills.Install` path still hashes and stages only `SKILL.md`; none of the new `fscopy.CopyTree`/`HashTree` APIs has a production caller. Installing a source containing `scripts/run.sh` or templates therefore succeeds while silently discarding every auxiliary file, leaving the newly advertised asset/script support unusable for the standard install workflow. Copy and hash the validated skill directory as a tree, and add an end-to-end installation test with a nested asset.

- [P1] Restore atomic cron job-ID reservation
  `internal/cron/store.go:102`
  `allocID` now performs an unlocked `Stat` check, then `writeJob` creates the directory afterward. Concurrent `Add` calls in the same second can both choose the same timestamp ID and race on the same `metadata.json`, so one scheduled job is overwritten even though both callers report success; a concurrent remove can also delete the gap and have `writeJob` resurrect it. This removes the prior mkdir-and-lock reservation protocol along with its concurrency regression tests. Reserve the ID atomically and hold the per-ID lock through the initial write.

Vasanthdev2004:
Checking in. The Jul 22 review left a P1 (install the auxiliary skill tree, not only the manifest) and there has been no movement since. If you still want to land this, a rebase onto current main plus that one change is what it needs, and I will review promptly. If I do not hear back within two weeks I will close it to keep the board honest; it can be reopened any time and nothing is lost.

jatmn (changes requested):
I found one install contract bug that should be fixed before merge. Everything else in recent review threads either looks intentional, out of scope for this PR, or needs a maintainer call on #584 — not another round of code churn.

## Findings

- [P1] Unify manifest validation with what `CopyTree` actually installs
  `internal/skills/install.go:115-175`
  `internal/fscopy/verify_unix.go:166-168` (and the matching `!unix` symlink skip)

  **What goes wrong**

  `Install` still validates the skill through a pathname read:

  1. `locateSkillDir` / `fileExists` use `os.Stat` (follows symlinks).
  2. `os.ReadFile(manifestPath)` reads the manifest bytes used for name/frontmatter validation.
  3. `fscopy.HashTree(skillDir)` and `fscopy.CopyTree(skillDir, staged)` then run on the source tree.

  `fscopy` deliberately skips symlink entries on copy and hash. That is correct for auxiliary files. It becomes a bug at the manifest boundary because step 2 can succeed through a symlink while step 3 never materializes `SKILL.md` in the staged tree.

  On `main` (merge-base `f30f550e`), install always did:

  ```go
  data, _ := os.ReadFile(manifestPath)
  os.WriteFile(filepath.Join(staged, skillFileName), data, 0o644)
  ```

  So even when the *source* manifest was a symlink, the *installed* tree always contained a regular `SKILL.md`. This PR removed that materialization step in favor of tree copy only, which reintroduced a broken-install path.

  **Failure path (reproducible by inspection)**

  Source tree:

  ```
  demo/
    SKILL.md -> /some/other/manifest.md   # symlink
    scripts/run.sh                        # regular file
  ```

  Sequence on current head:

  - `ReadFile("demo/SKILL.md")` follows the symlink and parses frontmatter → install proceeds.
  - `HashTree(demo)` skips the symlink entry; hashes `scripts/run.sh` only.
  - `CopyTree(demo, staged)` copies `scripts/run.sh` but not `SKILL.md`.
  - `HashTree(staged)` matches source hash (both omit the symlink).
  - `CommitDir` succeeds; `skills.lock` records a valid hash.
  - Installed dir has scripts but **no** `SKILL.md`.
  - `Load` / `confineSkillPath` skip the skill (no loadable manifest) even though the lock says it is installed.

  This is not theoretical hardening — it is a regression from the old write-bytes staging model to tree copy without preserving the manifest invariant.

  **Root cause**

  Install now has two parallel contracts that are not the same:

  | Stage | Behavior |
  |-------|----------|
  | Validate | pathname read; symlinks followed |
  | Persist | `fscopy` tree copy; symlinks skipped |

  Any source entry that passes validation but is skipped by `fscopy` can leave the installed tree missing data the installer already treated as authoritative. The manifest is the critical case because the rest of the runtime keys off a regular `SKILL.md` on disk.

  **How to fix the root cause (not a band-aid)**

  Pick one explicit install invariant and enforce it at commit time:

  > **Invariant:** after staging, `staged/SKILL.md` must exist, be a regular file, and match the bytes/name already validated.

  Recommended implementation (smallest, matches pre-PR behavior, keeps `fscopy` security properties):

  1. Keep `CopyTree` for auxiliary files.
  2. After `CopyTree`, **materialize the manifest from the already-validated `data` buffer**:

     ```go
     if err := os.WriteFile(filepath.Join(staged, skillFileName), data, 0o644); err != nil { ... }
     ```

  3. Re-run `fscopy.HashTree(staged)` (or hash only the manifest delta) before commit so lock hash still reflects the installed tree.

  Alternative acceptable approaches (same invariant, author’s choice):

  - Reject install early if source `SKILL.md` is not a regular file (`os.Lstat` + mode check, or reuse `confineSkillPath`-style guards on the source tree before copy).
  - Post-`CopyTree` gate: `Lstat(staged/SKILL.md)` must be regular; fail install otherwise.

  What **not** to do:

  - Teach `fscopy` to follow symlinks globally — that weakens the security model this PR just built.
  - Add only a comment/test for symlinks under `scripts/` without fixing the manifest boundary — that leaves the broken commit path.

  **Test to add (so this does not come back)**

  ```go
  func TestInstallMaterializesManifestWhenSourceIsSymlink(t *testing.T) {
      // source SKILL.md is symlink to a regular file with valid frontmatter
      // expect: install succeeds AND dest/<name>/SKILL.md is a regular file
      // expect: Load(dest) discovers the skill
  }
  ```

  Also worth a negative test: source `SKILL.md` symlink + only auxiliary files → install must fail *or* materialize manifest; must not commit a lock entry for an unloadable skill.

## Needs maintainer decision

- [P2] Confirm #584 closure scope: metadata vs inlined contents vs script execution
  `internal/skills/skills.go:195-211`, issue #584

  This PR clearly delivers the install/discovery half of #584:

  - full tree install via `fscopy.CopyTree`
  - recursive asset discovery into `Skill.Assets`
  - `<skill_assets>` metadata (relative path + size) in `FormatOutput`

  The issue text also asks to add subdirectory **contents** to skill context and to allow **calling scripts**. This PR does not inline file bodies or add an execution path — by design, install never executes fetched content.

  Please confirm with maintainers whether closing #584 on this PR means:

  1. **Metadata milestone** — install + list assets; contents/execution tracked separately, or
  2. **Full issue** — still needs content embedding and/or a supported script-invocation workflow.

  That is a product decision, not a code defect in the current diff. Once confirmed, update the PR body/issue so the next review does not re-litigate scope.

## Intentionally not raised (please do not “fix” these in response to this review)

These came up in automated/previous reviews. Re-checked on head `5cb8b089`; they look like accepted design, not merge blockers. Fixing them would be scope drift and is likely why this PR has seen so many review rounds.

| Topic | Why it is not a finding |
|-------|-------------------------|
| Hidden dotfiles (`.env`) copied but omitted from `Assets` | `loadAssets` explicitly skips `.*` with a comment calling out `.env`. That is a privacy choice: install full tree, do not advertise secrets in model output. Not a bug. |
| `maxAssetCount` / `maxVisitedEntries` / `maxAssetSize` | Documented constants with dedicated tests (`TestLoadAssetsCountCapStopsDiscovery`, etc.). Bounded discovery is intentional. Skills with 500+ assets are outside normal use. |
| TUI empty-body guard vs skill tool `FormatOutput` | TUI guard predates assets; empty `SKILL.md` body is treated as user error for slash invocation. Edge-case inconsistency only. |
| `loadAssets` swallowing `ReadDir` errors | Matches package fail-open philosophy (“a single bad skill never hides the good ones”). |
| Windows `!unix` hash/copy TOCTOU | Documented in `verify_nounix.go`; bounded to hash drift/install failure, not silent bad bytes. |
| `issue-approved` label on #584 | Author is `CONTRIBUTOR`; exempt per project policy. |

If maintainers want different behavior on any of the above, that should be a **new issue/PR with explicit product sign-off**, not another drip fix on this branch.

## Guidance: closing the review loop (why this PR kept getting findings, and how to stop)

### Why there were so many review comments

This branch touches a **wide contract surface** with several independent pipelines:

```
fetch source → validate manifest (pathname)
            → hash tree   (fscopy filters)
            → copy tree   (fscopy filters)
            → lock/commit
            → load/discover assets (different filters + caps)
            → format for model (output caps)
            → TUI / tool / plugin consumers
```

Each pipeline has slightly different rules (symlinks, dotfiles, size/count caps, truncation). Reviewers (human and bot) naturally trace edges where those rules diverge. After many commits, most of those edges are now **intentional** — but the comments keep coming because nothing in the PR documents the contract matrix, and one real bug (manifest materialization) remained at the boundary between old and new install logic.

This is not a sign the implementation is broadly broken. CI is green, nested asset install/discovery is tested, and prior blockers (atomic staging, `FormatOutput` truncation, fd-held unix copy) are addressed. The remaining noise is from **contract sprawl**, not missing tests on the happy path.

### What “done” looks like for this PR

1. **Fix the manifest invariant** (finding above) with a test.
2. **Document the contract matrix** in one place (PR description or a short comment block in `install.go` / `skills.go`):

   | File class | Install (`fscopy`) | Lock hash | `loadAssets` | `FormatOutput` |
   |------------|-------------------|-----------|--------------|----------------|
   | `SKILL.md` | must end up regular in dest | included | excluded (is Content) | body in `<skill>` |
   | visible regular files | copied | included | listed | listed in `<skill_assets>` |
   | `.*` hidden files | copied | included | **skipped (privacy)** | not listed |
   | symlinks | skipped | skipped | skipped | n/a |
   | `.git` | skipped | skipped | skipped | n/a |
   | >1 MiB files | copied | included | skipped (cap) | n/a |
   | >500 assets | all copied | all hashed | first 500 (bounded) | may truncate |

3. **Tighten the overclaim comment** at `skills.go:451-454` (“every file … stays discoverable”) to say discoverable **within documented caps and privacy filters**. That prevents the next reviewer from treating caps as a regression.

4. **Get maintainer sign-off on #584 scope** so “metadata only” vs “inline contents” is not debated again.

### Author checklist before requesting re-review

- [ ] `TestInstallMaterializesManifestWhenSourceIsSymlink` (or equivalent) passes
- [ ] `go test -race ./internal/skills/... ./internal/fscopy/...` passes
- [ ] PR description includes the contract table above (or link to it)
- [ ] #584 closure wording matches maintainer decision
- [ ] No drive-by changes to dotfile policy, discovery caps, or TUI guards unless maintainers explicitly asked

If the manifest fix and scope confirmation are in, I do not expect another finding drip from the install/discovery/output surface on re-review.

Issue #429, opened 2026-07-03
Multi-model configuration per provider


eli-l:
### Problem

Current provider wizard/config allows to add 1 provider and pick a single model. These are stored in settings.json as a single record.

Model favorites are ambiguous.

### Requested behavior

Provider like opencode may provide multiple models. Model discovery is already in place, but never written into the config. Local config shall cache the list of available models for picking among the providers.

Once multiple providers are providing the same models with the same names - favourite records becomes ambiguous and are not preserving the specific configuration of provider + model to be used. Such ambiguity may lead to unwanted costs.

### Why it belongs in Zero

Zero shall provide freedom of choosing providers, models and straightforward way of switching between them.

eli-l:
Actually I realised this might require more changes, at least proper support for inline --model flag, that should use the same approach as favourite items (provider/model) format.

Initial config seems to provide 1 record per provider+model. My suggestions changes the approach, as I want to have a single config, configured once and than I adjust the usage depending on my context, environment, needs. Re-adding provider for each model is a big no for me due to accesses shall be configured on provider level, not model level.

I will refrain from further effort before this is discussed/accepted.

eli-l:
This issue triggered endless AI slop in different directions. I got bored arguing with bots on useless migrations and endless scope adjustments.

PR #593, opened 2026-07-08
Introduce new favorite record format. Favourite (config.json) cleanup CLI command. Discovered model are persistent in config.json.

eli-l:
## Summary

Favorites now use provider/model format to reduce unambiguity.
`config cleanup` command to cleanup stale favorite configs (old format).
For cleanup needs discovered models are saved in config.json for each provider.

## Linked issue

[](https://github.com/Gitlawb/zero/issues/429)

## Checklist

- [ ] `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Added `config summary` and `config cleanup` commands, including machine-readable `--json` output.
  * Persist discovered provider models to improve later setup and model selection reuse.

* **Bug Fixes**
  * Stale favorites cleanup removes invalid/unknown/unavailable entries across user and project config.
  * Favorites are now tracked per provider/model, including model IDs with slashes, with deduplication and consistent ordering.
  * Validation now reports multiple issues in one pass for invalid provider selections and favorite references.

* **Tests**
  * Expanded CLI/config and model picker coverage for cleanup, persistence, and favorites behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/Gitlawb/zero
 
-go 1.26.4
+go 1.26.5
 
 require (
 	charm.land/bubbles/v2 v2.1.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/Gitlawb/zero
 
-go 1.26.5
+go 1.26.4
 
 require (
 	charm.land/bubbles/v2 v2.1.0
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -618,10 +618,6 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		doctorUserConfigPath = resolveOptions.UserConfigPath
 		projectConfigPath = resolveOptions.ProjectConfigPath
 	}
-	// Fail-soft, one-time: prune favorite model entries whose owning provider
-	// no longer exists in either the user or project config. Safe on every
-	// startup (idempotent) — non-fatal and best-effort.
-	_, _ = config.CleanupStaleFavorites(userConfigPath, projectConfigPath)
 
 	needsSetup := setupRequired(resolved)
 	if needsSetup && !forceSetup {
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -618,6 +618,10 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		doctorUserConfigPath = resolveOptions.UserConfigPath
 		projectConfigPath = resolveOptions.ProjectConfigPath
 	}
+	// Fail-soft, one-time: prune favorite model entries whose owning provider
+	// no longer exists in either the user or project config. Safe on every
+	// startup (idempotent) — non-fatal and best-effort.
+	_, _ = config.CleanupStaleFavorites(userConfigPath, projectConfigPath)
 
 	needsSetup := setupRequired(resolved)
 	if needsSetup && !forceSetup {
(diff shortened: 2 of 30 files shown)


kevincodex1:
hello @eli-l thanks for your work on this. kindly rebase to main and fix conflcits

anandh8x (changes requested):
Thanks for taking on #429. The core problem is real: favorites are still bare model ids, while RecentModels is already provider-qualified, so multi-provider setups can resolve the wrong provider.

What works in the direction of this PR:
- Making favorites provider-aware matches what we already do for recents
- A migration path for old bare-id favorites is needed either way

Where I'd push back is scope. Caching full discovered model lists on every provider in config.json, plus a new `zero config cleanup` command, is a bigger product surface than the favorites ambiguity needs. A tighter cut would be:

1. Store favorites the same way as RecentModels (provider + model), not a "provider/model" string split
2. Migrate or drop old bare-id favorites on read/write
3. Defer full model-list persistence and the cleanup command unless there's a stronger reason than validation inventory

A few concrete issues on the current tip:

- Branch is conflicting with main (`mergeStateStatus: DIRTY`) — needs a rebase before review can finish
- Checklist still has go build / go test / gofmt unchecked
- `zero providers models` failing when the discovered-models write fails is inconsistent with the wizard, which treats that as best-effort; listing should still succeed
- `mergeSavedProvider` looks like a second path next to `upsertSavedProviderProfile` — worth consolidating
- Splitting favorites on the first `/` is fragile if a provider name ever contains `/`; structured fields avoid that

Happy to re-review a narrower favorites-only revision. The multi-provider favorite fix is worth landing; the rest can follow if we still need it.

gnanam1990 (changes requested):
## Review

**Verdict:** Request changes. Agree with @anandh8x on scope and structured favorites.

### Process / merge
- Branch is still **CONFLICTING with main** (`mergeStateStatus: DIRTY`) — rebase/merge main before this can land. I hit a conflict in `internal/config/local_control_test.go` locally.
- Checklist still has go build / go test / gofmt unchecked.
- Parent issue #429 is `issue-approved` — good.

### What improved on tip (prior feedback)
- `zero providers models` now **warns** on persist failure instead of failing the list (good).
- `mergeSavedProvider` is gone; wizard uses `upsertSavedProviderProfile` (good).

### Blockers

1. **Bare-id favorites break silently, then cleanup deletes them.**  
   `favoriteModelSet` keeps old entries like `"gpt-4.1"`; `modelFavoriteKey` always uses `"provider/model"` when `OwnerProvider` is set. Favorites vanish after upgrade. `zero config cleanup` treats bare IDs as invalid and removes them — data loss, not migration.

2. **First-`/` encoding is wrong for slashy model IDs.**  
   Legacy favorites are bare model IDs. IDs like `openai/gpt-4.1` (OpenRouter-style) get misread as `provider=openai`. Structured `{provider, model}` fields (same as `RecentModelEntry`) avoid this. Do not encode identity in a single slash-separated string.

3. **Scope is still too large for the favorites fix.**  
   Caching full discovered model lists + `zero config cleanup` + inventory-based validation circularly justify each other. A tighter cut:
   1. Store favorites like RecentModels (provider + model fields)
   2. Migrate or drop bare-id favorites on read/write
   3. Defer model-list persistence and the cleanup command

### Other notes
- Listing models still rewrites config as a side effect and persists **before** catalog filtering.
- `OnLiveModels` errors are swallowed (`_ =`); picker can fail persist with no UI warning.
- Cleanup inventory is user-config only — project-scoped providers get purged.
- New validate requiring `provider.Model ∈ provider.Models` is a separate semantic change and will false-positive before discovery.
- `providermodelcatalog.Models()` dropped `FilterModelsForProvider` — looks unrelated; restore or justify separately.

### Ask
Happy to re-review a **favorites-only** revision on a rebased tip: structured provider+model favorites + real migration for bare (and slash-containing) legacy IDs. Model cache / cleanup can follow in a separate PR if still needed.

eli-l:
Closing as I keep getting useless requests in different directions.

PR #594, opened 2026-07-08
providerhealth, web_fetch: allow forward proxy on loopback in safeDialContext

### LINKED ISSUE #569: providerhealth: safeDialContext blocks proxy connections through localhost proxy
## Version / Environment

**Zero version:** 0.2.0
**OS:** Windows 11
**Proxy:** http://127.0.0.1:3067 (local forward proxy via HTTPS_PROXY env var)

## Steps to Reproduce

1. Set HTTPS_PROXY=http://127.0.0.1:3067 (a local HTTP forward proxy)
2. Configure any remote provider (e.g. opencode-go pointing at https://opencode.ai/zen/go/v1)
3. Run `zero doctor --connectivity`

## Expected Behavior

The connectivity check passes -- curl with the same proxy configuration works fine:

```
$ curl -v --proxy http://127.0.0.1:3067 "https://opencode.ai/zen/go/v1/models"
* CONNECT opencode.ai:443 HTTP/1.1
< HTTP/1.1 200 Connection established
* CONNECT tunnel established, response 200
[200 OK]
```

## Actual Behavior

```
[fail] provider.connectivity - Provider connectivity failed: Get "https://opencode.ai/zen/go/v1/models":
       proxyconnect tcp: provider connectivity URL is unsafe: loopback hosts are blocked
```

## Root Cause

safeDialContext in internal/providerhealth/providerhealth.go enforces an SSRF blocklist against every dialed TCP address -- including 127.0.0.0/8, 10.0.0.0/8, 192.168.0.0/16, etc.

When HTTPS_PROXY=http://127.0.0.1:3067 is set, Go's HTTP client first dials the proxy address (127.0.0.1:3067) to establish the CONNECT tunnel. safeDialContext intercepts this dial, sees 127.0.0.1, and returns:

```
endpointSafetyError{message: "provider connectivity URL is unsafe: loopback hosts are blocked"}
```

The safety check correctly protects against SSRF to the target server, but it does not account for the case where the dial target is the proxy itself, not the upstream provider. When a proxy is configured, dialing 127.0.0.1 is legitimate -- it is connecting to the proxy, not to the provider endpoint.

The target domain (opencode.ai) resolves to 172.65.90.22 (Cloudflare) which is NOT in any blocked range. validateEndpoint() passes correctly. The failure is purely at the dial stage when connecting to the proxy address.

## Suggested Fix

safeDialContext (or newConnectivityClient) should detect whether the current dial target matches the configured HTTP_PROXY / HTTPS_PROXY host, and if so, skip the address validation and pass the connection through to the standard dialer.

An alternative approach: instead of a custom safeDialContext transport, newConnectivityClient could use Go's default transport (which already handles proxy configuration via http.ProxyFromEnvironment) and rely on CheckRedirect for SSRF protection, which already has the correct allowLoopbackOrPrivate semantics.

## Note

zero exec works correctly through the same proxy because it uses a different HTTP client path (the default Go transport), which handles proxy connections without going through safeDialContext. The issue is isolated to the connectivity health check path.

## Related

- The webFetchSafeDialContext in internal/tools/web_fetch.go may have the same issue.

baoyu0:
## ponytail fix

`safeDialContext` blocks 127.0.0.0/8. When `HTTPS_PROXY=http://127.0.0.1:3067` is set, Go dials the proxy first — `safeDialContext` blocks it.

**Fix:** inline `http.ProxyFromEnvironment` check in both dial functions. If the dial target matches the configured proxy, skip SSRF validation.

### Changes

**providerhealth.go** — `safeDialContext`: 7 lines added
**web_fetch.go** — `webFetchSafeDialContext`: 9 lines added, `transport.Proxy` restored from `nil` to `http.ProxyFromEnvironment`

New imports: **0**. New types: **0**. New helpers: **0**.

Existing SSRF validation path **completely unchanged** for non-proxy dials.

Closes #569

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connectivity checks and web fetching when system proxy settings are enabled.
  * Prevented proxy-based requests from being incorrectly blocked when the request target matches the currently configured forward proxy.
  * Enhanced web request handling so URL validation continues to occur at the right stage, while preserving existing response behavior.

### DIFF
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -542,9 +542,11 @@ func safeDialContext(resolver Resolver, allowLoopbackOrPrivate bool) func(contex
 			return nil, err
 		}
 		// ponytail: forward proxy on loopback — this is the proxy address, not the endpoint
-		if proxyURL, _ := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Host: "x"}}); proxyURL != nil {
-			if proxyURL.Hostname() == host && (proxyURL.Port() == "" || proxyURL.Port() == port) {
-				return dialer.DialContext(ctx, network, address)
+		for _, s := range []string{"https", "http"} {
+			if proxyURL, _ := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Scheme: s, Host: "x"}}); proxyURL != nil {
+				if proxyURL.Hostname() == host && (proxyURL.Port() == "" || proxyURL.Port() == port) {
+					return dialer.DialContext(ctx, network, address)
+				}
 			}
 		}
 		if addr, parseErr := netip.ParseAddr(host); parseErr == nil {
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
-	"os"
 	"strings"
 	"time"
 
@@ -532,48 +531,19 @@ func sensitiveAuthHeaderNames(profile config.ProviderProfile, kind config.Provid
 	return out
 }
 
-// resolveProxyEnvAddrs resolves configured HTTP_PROXY/HTTPS_PROXY/ALL_PROXY
-// hostnames to IPs once per transport creation, so safeDialContext can
-// distinguish a proxy dial from a provider endpoint dial. (ponytail)
-func resolveProxyEnvAddrs(resolver Resolver) []string {
-	for _, key := range []string{"HTTPS_PROXY", "HTTP_PROXY", "ALL_PROXY"} {
-		val := os.Getenv(key)
-		if val == "" {
-			continue
-		}
-		u, err := url.Parse(val)
-		if err != nil || u.Hostname() == "" {
-			continue
-		}
-		host := u.Hostname()
-		if addr, err := netip.ParseAddr(host); err == nil {
-			return []string{addr.String()}
-		}
-		if addrs, err := resolver.LookupNetIP(context.Background(), "ip", host); err == nil {
-			out := make([]string, len(addrs))
-			for i, a := range addrs {
-				out[i] = a.String()
-			}
-			return out
-		}
-	}
-	return nil
-}
-
 // safeDialContext returns a dial function that resolves the target host, refuses
 // the connection if any resolved address is blocked, then dials the validated IP
 // literal so the kernel cannot re-resolve to a different address after the check.
 func safeDialContext(resolver Resolver, allowLoopbackOrPrivate bool) func(context.Context, string, string) (net.Conn, error) {
 	dialer := &net.Dialer{Timeout: 10 * time.Second, KeepAlive: 30 * time.Second}
-	proxyHosts := resolveProxyEnvAddrs(resolver) // ponytail
 	return func(ctx context.Context, network, address string) (net.Conn, error) {
 		host, port, err := net.SplitHostPort(address)
 		if err != nil {
 			return nil, err
 		}
-		// ponytail: bypass SSRF check when dialing a configured forward proxy
-		for _, ph := range proxyHosts {
-			if host == ph {
+		// ponytail: forward proxy on loopback — this is the proxy address, not the endpoint
+		if proxyURL := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Host: "x"}}); proxyURL != nil {
+			if proxyURL.Hostname() == host && (proxyURL.Port() == "" || proxyURL.Port() == port) {
 				return dialer.DialContext(ctx, network, address)
 			}
 		}
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
@@ -33,25 +34,25 @@ const (
 type Category string
 
 const (
-	CategoryConfig       Category = "config"
-	CategoryUnsupported  Category = "unsupported"
-	CategoryAuth         Category = "auth"
-	CategoryRateLimit    Category = "rate_limit"
-	CategoryNetwork      Category = "network"
-	CategoryTimeout      Category = "timeout"
-	CategoryProvider     Category = "provider_error"
-	CategoryConnectivity Category = "connectivity"
+	CategoryConfig        Category = "config"
+	CategoryUnsupported   Category = "unsupported"
+	CategoryAuth          Category = "auth"
+	CategoryRateLimit     Category = "rate_limit"
+	CategoryNetwork       Category = "network"
+	CategoryTimeout       Category = "timeout"
+	CategoryProvider      Category = "provider_error"
+	CategoryConnectivity  Category = "connectivity"
 )
 
 const defaultTimeout = 5 * time.Second
 
 type Options struct {
-	Profile      config.ProviderProfile
-	Connectivity bool
-	HTTPClient   *http.Client
-	Resolver     Resolver
-	Timeout      time.Duration
-	UserAgent    string
+	Profile       config.ProviderProfile
+	Connectivity  bool
+	HTTPClient    *http.Client
+	Resolver      Resolver
+	Timeout       time.Duration
+	UserAgent     string
 }
 
 type Resolver interface {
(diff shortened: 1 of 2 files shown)


kevincodex1:
hello @baoyu0 thank you for your contribution. please address coderabbitai comments

baoyu0:
Addressed the CodeRabbit scheme feedback — `ProxyFromEnvironment` now tries both `"https"` and `"http"` schemes in both `safeDialContext` and `webFetchSafeDialContext`. Pushed as b4581f4.

Vasanthdev2004 (changes requested):
I focused on the SSRF angle since safeDialContext is what keeps provider and web_fetch dials off loopback and private ranges. The core change is sound and narrow: the bypass only fires when the dialed address matches the configured HTTPS_PROXY/HTTP_PROXY host:port, and non-proxy dials still go through the full blocklist; loopback/private targets are still rejected at URL validation and at redirect re-validation, so the exception can't reach arbitrary internal services — it only lets the agent dial the user's own explicitly-configured forward proxy.

Requesting changes because the web_fetch half breaks three existing tests that encode our security expectations: removing validateWebFetchURLBeforePermission from RejectBeforePermission means loopback/private/metadata URLs that used to be rejected before prompting now reach the user first (run() still blocks them, so it's a defense-in-depth regression, not a fetch hole), and transport.Proxy went from nil to ProxyFromEnvironment. Either restore the pre-permission safety check or update those tests with a security justification for the new behavior. Minor: when the proxy URL omits a port, the bypass matches any port on that host — not exploitable today given upstream validation, but defaulting to 80/443 before comparing would be tighter.

Vasanthdev2004:
@gnanam1990 #594 is security-relevant (loopback forward-proxy exception in safeDialContext). I requested changes: the SSRF narrowing itself is sound, but the web_fetch half removes the pre-permission URL-safety check so loopback/private URLs now reach the permission prompt before being rejected, and it breaks 3 existing tests. Would value your read on whether restoring the pre-permission check vs updating the tests with a justification is the right call.

jatmn (changes requested):

## Findings

- [P2] Restore LF line endings before merging
  `internal/providerhealth/providerhealth.go:1`
  This change converts both touched Go files to CRLF, so the repository's `git diff --check origin/main...HEAD` emits thousands of trailing-whitespace errors (and the Zero Review check is failing). Keep the functional edits, but restore the files' existing LF line endings so the required hygiene check can pass.

- [P1] Keep the proxy bypass tied to the actual request's proxy decision
  `internal/tools/web_fetch.go:319`
  The dial hook probes both schemes with a synthetic host and bypasses validation based only on address equality; it does not establish that the transport selected that proxy for the real request. For example, with only `HTTP_PROXY=http://rebind.example` set, an HTTPS request to `https://rebind.example` is direct, but the synthetic HTTP lookup still matches the `:443` dial because the proxy URL has no explicit port. A hostname that resolves public during preflight and rebinds to a private address before dial is then connected without the normal re-resolution and pinning. The identical logic in `internal/providerhealth/providerhealth.go` has the same gap. Couple the exception to the actual proxy selected for the request (including `NO_PROXY` and scheme semantics), and retain origin pinning for direct dials.

- [P1] Restore the static URL rejection before permission is requested
  `internal/tools/web_fetch.go:154`
  Removing `validateWebFetchURLBeforePermission` means known-unsafe literals such as `http://127.0.0.1/admin`, invalid schemes, and blocked ports now reach the network permission prompt; they are only rejected later by `run`. This breaks the `PrePermissionRejecter` contract and the unchanged registry/agent tests that require an immediate denial. Restore the non-DNS validation in this hook as well as the runtime validation.

- [P2] Update the default-transport test for the new proxy behavior
  `internal/tools/web_fetch.go:310`
  `TestWebFetchToolConfiguresDialTimeSafetyForDefaultTransport` still fails whenever `transport.Proxy` is non-nil, while this change now installs `http.ProxyFromEnvironment`. The affected `internal/tools` test suite therefore fails at HEAD. Replace the obsolete assertion with coverage for the intended proxy-aware path while retaining an assertion that the safe dialer remains installed.

Vasanthdev2004:
@baoyu0 checking in on this one, it has been quiet since 9 July and I would rather it landed than went stale.

To be clear about where it stands: the core change is sound and I said so. The bypass only fires when the dialed address matches the configured proxy host:port, non-proxy dials still go through the full blocklist, and loopback and private targets stay rejected at URL validation and redirect re-validation. That half I could not break.

What is blocking is the web_fetch side, and it is a smaller ask than the thread length suggests. Removing `validateWebFetchURLBeforePermission` from `RejectBeforePermission` means loopback, private and metadata URLs that used to be refused before prompting now reach the user first. `run()` still blocks them, so it is a defence-in-depth regression rather than a fetch hole, but three existing tests encode that expectation and they fail. Either restore the pre-permission check, or update those tests with the security justification for the new behaviour. Either is fine; what cannot happen is the tests being changed without the reasoning, because those tests are how we know that layer still exists.

The port-defaulting nit is optional and I would not hold merge on it.

jatmn also requested changes on 13 July, so his notes are worth a read alongside mine. The branch will want a rebase too, main has moved a fair way since July.

If you would rather not carry it further, say so and I will pick it up rather than leave it open indefinitely. No hard feelings either way.

baoyu0:
@Vasanthdev2004 thanks for the detailed check-in and the thorough review throughout. I'm going to hand this one off — please feel free to pick it up and take it the rest of the way (restore the pre-permission check or update the tests, whichever you judge right). I don't have the bandwidth to carry it further right now, and I'd rather it land than linger. No hard feelings at all — appreciate everything you put into it.

Vasanthdev2004 (changes requested):
Picked this up because it is the only PR on the board with failing checks. All three Smoke jobs and Zero Review are red, and I think I can tell you why in one go. None of it is about your actual change, which I could not see until I went looking for it.

**Both files were committed with CRLF line endings.** Main has LF:

```
providerhealth.go   main[CRLF=0 LF=858]   this branch[CRLF=817 LF=0]
web_fetch.go        main[CRLF=0 LF=606]   this branch[CRLF=599 LF=0]
```

So git reports every line as changed, which is where the +1416/-1400 comes from. Normalising the endings, your real change is **171 lines in providerhealth.go and 51 in web_fetch.go**, not 2816. That alone makes the PR unreviewable as displayed, and it is why it now conflicts with main: a whole-file rewrite conflicts with any change to either file.

```
$ git merge pr594          # onto current main
CONFLICT (content): Merge conflict in internal/tools/web_fetch.go
$ go build ./...
internal\tools\web_fetch.go:1:1: expected 'package', found '<<'
internal\providerhealth\providerhealth.go:1:1: expected 'package', found '<<'
```

CI builds the merge result rather than your branch in isolation, which is why everything downstream is red while `go build ./...` on your head alone is clean.

**The branch also drops a field that main uses.** In `providerhealth.Options`:

```go
-	OAuthResolver providerio.TokenResolver
```

That is nothing to do with loopback proxies. Main sets it in three places, `internal/cli/app.go:167`, `app.go:689-690` and `internal/cli/provider_models.go:155`, so merging as-is would fail to compile even without the conflict. It reads like the branch is based on an older main and the field was lost in a resolve rather than removed deliberately. If you did mean to remove it, that needs to be its own PR with those call sites updated.

**And the const block is not gofmt-aligned:**

```go
-	CategoryConfig       Category = "config"
+	CategoryConfig        Category = "config"
```

An extra space on each, while `prefix netip.Prefix` in `embeddedIPv4Prefix` lost one. `gofmt -l` will flag both files.

**What I would do**, in this order:

1. Fix the line endings first, before anything else. `git config core.autocrlf input`, then re-checkout the two files from main and re-apply your change on top. Everything below is much easier once the diff is 222 lines instead of 2816.
2. Rebase onto current main and keep `OAuthResolver`.
3. `gofmt -w` the two files.

I have deliberately not reviewed the proxy logic itself. With the diff in this state I cannot tell your changes from the noise, and I would rather give you a real review of 222 lines than guess at 2816. Ping me once it is rebased and I will go through the actual change properly.

Vasanthdev2004:
Taking this over as you offered, thank you for the work on it and for handing it off cleanly. I will re-apply the real change, about 220 lines under the CRLF noise, onto current main with LF endings and with the OAuthResolver field kept, and open a fresh PR that credits you. This one can close once that is up.

Vasanthdev2004:
Re-applied as #1025 with your approach kept and you credited as co-author on the commit. Three things differ, all called out in the description there: the pre-permission URL check stays, the proxy port is matched exactly instead of any port when the URL has none, and the dialer asks the transport for its proxy rather than reading the environment a second time.

Closing this one in favour of #1025. Thanks again for the work on it and for the clean handoff.

PR #597, opened 2026-07-08
Harden project-scoped MCP credential boundaries

### LINKED ISSUE #596: Harden project-scoped MCP credential boundaries
### Problem

Project MCP configuration is shared repo-controlled config, while user MCP credentials and OAuth tokens are user-private state. Today these boundaries need stronger isolation so project-level MCP changes cannot accidentally inherit or reuse user credential material across a changed server target.

### Requested behavior

Add project-aware MCP merge behavior:

- Project MCP config remains supported.
- If project config changes an existing MCP server target, it must not inherit user credential material from headers, env, or OAuth config.
- User config and CLI overrides should keep trusted existing merge behavior.
- MCP OAuth tokens should be bound to the resolved server identity, not only server name.
- Legacy name-only OAuth tokens should migrate only for user-only servers.
- Logout/status should handle identity-bound tokens without exposing token material.

### Why it belongs in Zero

Zero supports project config and MCP tools. Project config crosses a trust boundary: repo-owned settings can be shared with users who have private local credentials. Zero should preserve useful shared MCP config while preventing credential reuse across that boundary.

Users can manually avoid same-name project MCP servers or clear credentials in project config, but this is easy to miss and does not protect stored OAuth tokens. A blanket ban on project MCP config would be safer but too disruptive. Identity-bound tokens plus guarded project merge gives narrower compatibility-preserving hardening.

_No response_

GautamBytes:
Fixes #596

What changed:
- Add project-aware MCP merge guard: project target changes cannot inherit user headers, env, or OAuth config.
- Mark project-touched MCP servers and carry that into runtime token lookup.
- Bind MCP OAuth tokens to server identity; legacy fallback only for user-only servers.
- Update logout/status for identity-bound tokens; reject ambiguous OAuth server names.
- Document project MCP credential boundary and identity-bound OAuth behavior.

## Summary by CodeRabbit

* **New Features**
  * MCP OAuth tokens are now tied to the specific server identity, improving reliability when server settings change.
  * Project-level MCP changes can now be tracked separately from user-level configuration.

* **Bug Fixes**
  * Updated token login/logout and refresh handling to use the correct server-scoped storage.
  * Improved configuration merging so project changes that affect connection targets are handled more safely.
  * Added clearer guidance when changing MCP server targets so credentials are not reused unexpectedly.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -235,6 +235,13 @@ Servers are merged from user and project configs (project wins on conflicts). To
 - A `--header` value produced by command substitution (`"Authorization=Bearer $(print-token)"`) in a private shell config that you keep out of git.
 - A secret manager that injects the env var your MCP server reads on its own (the `command` and `args` then run inside that environment).
 
+Project MCP config remains supported for shared team servers, but a project layer
+that changes an existing server's target (`url`, `command`, or `args`) does not
+inherit user credentials. Clear or replace `headers`, `env`, or `oauth` explicitly,
+or use a new server name. OAuth tokens are bound to the resolved server identity,
+so changing a server target can require `zero mcp oauth login <server>` again.
+OAuth server names ending in `.<32 hex chars>` are reserved for token storage.
+
 ### As a server — expose Zero's tools to another agent
 
 ```bash
--- a/internal/cli/mcp_oauth.go
+++ b/internal/cli/mcp_oauth.go
@@ -86,7 +86,7 @@ func runMCPOAuthLogin(args []string, stdout io.Writer, stderr io.Writer, deps ap
 	if err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
 	}
-	if err := store.Save(serverName, token); err != nil {
+	if err := store.SaveForServer(server, token); err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
 	}
 
@@ -116,7 +116,7 @@ func runMCPOAuthLogout(args []string, stdout io.Writer, stderr io.Writer, deps a
 	if err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
 	}
-	removed, err := store.Delete(serverName)
+	removed, err := store.DeleteForServerName(serverName)
 	if err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
 	}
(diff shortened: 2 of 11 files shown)


kevincodex1 (approved):

jatmn (changes requested):

## Findings

  `GitHub mergeability`
  GitHub currently reports this branch as `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, and a local merge against the current `main` conflicts in `internal/cli/mcp_oauth_test.go` and `internal/config/resolver.go`. Please rebase on the latest `main` and resolve those conflicts before addressing the review finding below, so the fixes apply to the code that will actually be merged.

- [P1] Preserve the MCP trust gate when resolving project config
  `internal/config/resolver.go:173`
  The resolver rewrite now loads `options.ProjectConfigPath` whenever it is present and immediately merges it into the MCP config. Current `main` added an `ExcludeProject` trust gate for MCP resolution so an untrusted workspace drops its project `.zero/config.json` layer instead of registering and spawning repo-controlled MCP servers. As this branch is rebased, please keep `ResolveOptions.ExcludeProject` and make `ResolveMCP` skip the project layer when it is set; otherwise an untrusted cloned repo can once again inject stdio MCP servers into the TUI/tool surface, undoing the trust boundary from #529.

anandh8x (approved):
LGTM. Real security fix

Project merge now rejects target changes that would inherit user headers/env/oauth, and OAuth tokens are bound to server identity with legacy name-only migration only for user-only servers. Tests cover the attack cases without leaking secrets into errors. ExcludeProject is still enforced in ResolveMCP.

This matches a trust boundary other agents handle with trusted-server prompts and account-scoped OAuth: project config should not be able to reuse private credentials against a new MCP target. Zero needs it more because we merge project + user MCP and store tokens by server name.

gnanam1990 (approved):
## Review

**Verdict:** Approve with minor suggestions.

Agrees with the threat model: project retarget must not inherit user headers/env/oauth; OAuth tokens must be identity-bound with legacy name-only migration only for user-only servers. On tip, @jatmn's P1s look addressed (mergeable vs main; `ExcludeProject` preserved on `ResolveMCP`).

### What looks correct
- `mergeProjectMCPConfig` rejects project target changes (`url` / `command` / `args` / transport kind) when user `headers` / `env` / `oauth` would be inherited; escape hatch is explicit clear/replace or a new server name.
- `ProjectConfigured` is set for every project-touched server (including project-only names).
- Runtime attach uses `LoadForServer` / `SaveForServer` (identity-bound); legacy name-only migration is skipped when `ProjectConfigured`.
- Full `Resolve` and `ResolveMCP` both go through the project MCP merge gate.
- Local: `make lint` pass; `go test -race` on `./internal/config/` and `./internal/mcp/` pass; CI green.

### Non-blocking
1. After legacy→identity migration in `LoadForServer`, delete the name-only key so `Status`/`Load` don't keep dual material under `mcp:<name>`.
2. Add a merge test for inherited **oauth** material (e.g. `clientSecret` / endpoints) on a project URL override — headers/env are covered; oauth is not.
3. Docs: mention `type` as a target field in AGENTS.md; TUI OAuth summary still keys by server name only and can show “has token” for a stale identity.

### Process
- #596 is linked but lacks the `issue-approved` label (CONTRIBUTING house rule).

No disagreement with the core security fix. Remaining items are hygiene / coverage / small UX polish.

jatmn (approved):
LGTM

PR #599, opened 2026-07-08
feat(modelregistry): allow /effort on Hunyuan and vendor-prefixed gateway models

euxaristia:
## Summary
`/effort` refused to set a level on OpenGateway models like `tencent/hy3` because the name-based fallback in `reasoningEffortsForModelName` only recognized GPT-5, Codex, and o-series names.

- Strip a slash-form `vendor/` qualifier (`tencent/hy3`, `openai/o3-mini`) before matching, mirroring the existing `provider:` strip. This also lights up `/effort` for o-series models served under GitHub-style ids.
- Add hy3/hunyuan as a known reasoning family with low/medium/high.

Once the fallback returns a non-empty list, the existing plumbing handles the rest: the `/effort` picker, `/effort <value>`, and the Ctrl+T cycle, with the value forwarded as `reasoning_effort` in the OpenAI-compatible payload.

## Limitations
I'm well aware of the limitations here: the client only sends `reasoning_effort` on the wire. Whether OpenGateway translates it for the Tencent upstream is the gateway's concern, and compatible servers ignore unknown fields, so the worst case is a silent no-op rather than an error.

## Testing
- `go test ./internal/modelregistry/...` (includes new fallback cases for `tencent/hy3`, `hy3`, `hunyuan-t1`, `openai/o3-mini`, and a non-reasoning `nvidia/...` id)
- `go build ./...` and `go test ./internal/tui/... ./internal/providers/...` all pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model-name handling so vendor-style prefixes are recognized more reliably.
  * Added support for additional model families, including Tencent Hunyuan-style names, so reasoning effort levels are assigned more accurately.
  * Expanded coverage for several model identifiers to ensure consistent behavior across supported model types.

### DIFF
--- a/internal/modelregistry/catalog.go
+++ b/internal/modelregistry/catalog.go
@@ -214,11 +214,22 @@ func reasoningEffortsForModelName(name string) []ReasoningEffort {
 	if i := strings.IndexByte(n, ':'); i >= 0 { // drop a "provider:" qualifier
 		n = n[i+1:]
 	}
+	// Gateway ids also carry a slash-form "vendor/" qualifier (e.g. the
+	// OpenGateway's tencent/hy3, github's openai/o3-mini) — match on the bare name.
+	if i := strings.LastIndexByte(n, '/'); i >= 0 {
+		n = n[i+1:]
+	}
 	switch {
 	case strings.HasPrefix(n, "gpt-5") || strings.HasPrefix(n, "gpt5"):
 		// GPT-5 / Codex (gpt-5.x) add a "minimal" tier below low.
 		return []ReasoningEffort{ReasoningEffortMinimal, ReasoningEffortLow, ReasoningEffortMedium, ReasoningEffortHigh}
 	case strings.Contains(n, "codex") || isOSeriesModelName(n):
+		return []ReasoningEffort{ReasoningEffortLow, ReasoningEffortMedium, ReasoningEffortHigh}
+	case strings.HasPrefix(n, "hy3") || strings.Contains(n, "hunyuan"):
+		// Tencent Hunyuan reasoning models (served via OpenAI-compatible gateways).
+		// The client forwards reasoning_effort; whether the upstream honors it is
+		// the gateway's translation concern — unknown fields are ignored, so the
+		// worst case is a silent no-op rather than a 400.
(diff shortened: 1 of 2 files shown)


anandh8x (approved):
LGTM. Tiny, scoped, tested.

- The slash-strip mirrors the existing ':' strip, so provider:vendor/name resolves to name the same way as provider:name. Good symmetry.
- Adding hy3/hunyuan alongside the o-series case is the right call — they have the same effort range, and the comment about the gateway translating the field is honest.
- The test additions cover both directions: positive cases (tencent/hy3, hy3, hunyuan-t1, openai/o3-mini) and the non-reasoning nvidia/llama case to prove the broader prefix match doesn't false-positive.

gnanam1990 (approved):
**Verdict: Approve**

Fixes a real gap: gateway model ids like `tencent/hy3` never matched the name-based `/effort` fallback because only `provider:` prefixes were stripped, not `vendor/` segments.

**What looks good**
- `LastIndexByte('/')` strip mirrors the existing `:` strip; `openrouter:tencent/hy3` normalizes correctly to `hy3`.
- `openai/o3-mini` → `o3-mini` now hits `isOSeriesModelName` — fixes the same class of bug for GitHub-style vendor ids.
- Hunyuan branch uses low/medium/high (same as o-series); curated catalog entries with explicit `ReasoningEfforts` are untouched.
- Picker (`Registry.ReasoningEfforts`) and runtime resolver (`effectiveReasoningEfforts`) share the same fallback — no split-brain risk.
- Tests cover positives (`tencent/hy3`, `hy3`, `hunyuan-t1`, `openai/o3-mini`) and negatives (`nvidia/...`, `ollama/llama3.1`).

**Nits (non-blocking)**
- `strings.Contains(n, "hunyuan")` is broad — any id containing `hunyuan` gets effort controls; fine for this family but looser than `isOSeriesModelName`.
- Test name still says `GPT5AndOSeries` while now covering Hunyuan — cosmetic rename optional.

**Housekeeping (not blocking code)**
- No linked issue with `issue-approved` in the PR body.

**Issues:** none blocking.

kevincodex1 (approved):

PR #601, opened 2026-07-08
Improve grep content-mode performance for small head_limit

### LINKED ISSUE #600: Improve grep content-mode performance for small head_limit
### Problem

grep content mode still scans every matching line to compute exact truncation totals, even after head_limit output is already reached. On large match-heavy trees this wastes CPU and memory.

### Requested behavior

Optimize grep content mode:
Match regex against []byte to avoid per-line string allocation.
Stop content scans after head_limit + 1.
Show non-exact truncation marker: showing first N matches.
Keep exact totals available through output_mode=count.
Preserve glob behavior, sandbox exclusions, symlink confinement, and cancellation.

### Why it belongs in Zero

grep is core agent file-search infra. Faster bounded search means lower latency, less RAM churn, and less wasted work during common repo exploration.

Users can narrow path/glob/pattern or use count mode separately, but tool should honor head_limit as a real work bound for content output.

GautamBytes:
Fixes #600

What changed:
- Match grep lines as []byte and convert to string only for emitted content.
- Stop content-mode scans after head_limit+1 and show a non-exact truncation marker.
- Keep exact totals available through output_mode=count.

Perf (BenchmarkGrepLargeTreeHeadLimit):
- Time: 5.12ms/op -> 0.105ms/op (~48.7x faster)
- Memory: 4.26MB/op -> 44.5KB/op (~95.8x less)
- Allocs: 47,012/op -> 453/op (~103.8x fewer)

## Summary by CodeRabbit

- **Bug Fixes**
  - Search results now stop cleanly when a head limit is reached, and truncated output is reported more consistently.
  - Truncation messages now show a clearer “showing first N matches” notice instead of implying an exact total.
  - Result metadata now accurately marks limited output as truncated, helping downstream displays handle partial results correctly.

### DIFF
--- a/internal/tools/file_tools_test.go
+++ b/internal/tools/file_tools_test.go
@@ -239,12 +239,18 @@ func TestGrepToolMakesHeadLimitTruncationVisible(t *testing.T) {
 	if result.Status != StatusOK || !result.Truncated {
 		t.Fatalf("expected ok+truncated, got status=%s truncated=%v output=%q", result.Status, result.Truncated, result.Output)
 	}
-	if !strings.Contains(result.Output, "[truncated: showing first 2 of 3 matches") {
+	if !strings.Contains(result.Output, "[truncated: showing first 2 matches") {
 		t.Fatalf("expected visible grep truncation marker, got %q", result.Output)
+	}
+	if strings.Contains(result.Output, " of 3 matches") {
+		t.Fatalf("content truncation marker must not claim exact total, got %q", result.Output)
+	}
+	if result.Meta["truncation_reason"] != "head_limit" {
+		t.Fatalf("expected head_limit truncation reason, got %v", result.Meta)
 	}
 }
 
-func TestGrepToolCountsAllLinesWhileKeepingHeadLimitSmall(t *testing.T) {
+func TestGrepToolStopsAfterHeadLimitInContentMode(t *testing.T) {
 	root := t.TempDir()
 	var builder strings.Builder
 	for i := 0; i < 100; i++ {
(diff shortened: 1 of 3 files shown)


anandh8x (approved):
LGTM. Every change in this PR is in #600's requested-behavior list — nothing more.

   matchers going []byte is the right place to cut the alloc; only the final emit converts to string. 47k allocs/op → 453 is consistent with that. Bail-out via a sentinel error filtered
 at the top of scanGrepMatches keeps the context-canceled path intact. The new "showing first N matches" marker is honest — output_mode=count still gives exact totals. The +1 sentinel in
 the new test pins the early-stop contract. truncation_reason in Meta is a small useful signal.

   3 files, no new exports, no Tool interface change. Ship it.

gnanam1990 (approved):
**Verdict: Approve**

The change matches #600's requested-behavior list — nothing more.

**What looks good**
- `[]byte` matchers with string conversion only at emit time — right place to cut allocations; benchmark numbers are credible.
- Early stop at `head_limit + 1` via sentinel `errGrepLimitReached`, filtered at `scanGrepMatches` so cancellation still works.
- Truncation message is honest (`showing first N matches`); exact totals remain available via `output_mode=count`.
- `TestGrepContentScanStopsAfterHeadLimitPlusOne` pins the early-stop contract.
- `truncation_reason: head_limit` in Meta is a useful signal.
- 3 files, no new exports, no Tool interface change. Sandbox/confinement/cancellation paths untouched.

**Housekeeping (not blocking code)**
- Parent issue #600 lacks `issue-approved` — maintainers may want that per CONTRIBUTING before merge.

**Issues:** none blocking.

kevincodex1 (approved):

PR #603, opened 2026-07-08
fix(update): clearer error on unsupported release platform (android/termux)

### LINKED ISSUE #598: zero upgrade fails on Android/Termux: "unsupported release platform: android"
## Summary

Running `zero upgrade` (or `zero update --check`) on an Android / Termux environment fails with:

```
[zero] Could not install update: unsupported release platform: android
```

The same error blocks `zero update --check`, which should be a safe, non-mutating read-only check.

## Environment

- OS: Android (Termux shell, `/data/data/com.termux/files/home`)
- Shell: `/bin/sh`
- `zero` installed via the release downloader/updater

## Steps to reproduce

1. On an Android device, open a Termux shell.
2. Run `zero upgrade` (or `zero update --check`).
3. Observe the error: `unsupported release platform: android`.

## Expected behavior

Either:

- `android` (and `termux`) is recognized as a valid platform in the release/platform detection map so updates can be fetched, **or**
- if Android builds are not published, the updater should explicitly document this and **skip** the update gracefully (exit 0 / no-op) instead of erroring out — and `zero update --check` should never fail on an unsupported platform.

## Suggested fix

Add `android` (and `termux`) to the platform detection / release map, or explicitly document that Android builds are not published and skip the update gracefully instead of erroring.

## Additional notes

- This blocks the documented update workflow on mobile/Termux users.
- A read-only `--check` failing is especially surprising since it performs no mutation.

PierrunoYT:
## Summary
Fixes #598. `zero update`/`zero upgrade` on Android/Termux failed with the opaque error `unsupported release platform: android`, including for `--check`, which is documented as a safe, read-only operation.

## Change
`releasePlatform` in `internal/update/update.go` now returns a clear, actionable error when the running GOOS has no published release archive: it names the platform, lists the supported platforms (linux, macos, windows), and tells the user to rebuild from source. The command still fails (exit 1) rather than silently reporting "up to date," but does so gracefully — a clean stderr message via the existing `writeAppError` path, no panic or stack trace.

Before:
```
[zero] Could not install update: unsupported release platform: android
```

After:
```
[zero] Could not install update: zero does not publish release builds for platform "android" (supported: linux, macos, windows) — update/upgrade checks aren't available; rebuild from source instead
```

## Testing
- `go build ./...`
- `go test ./internal/update/... ./internal/cli/...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the error shown when checking for updates on an unsupported platform.
  * The new message explains that prebuilt release archives aren’t available for that platform, that self-updating/release naming guidance isn’t provided, and offers clear next steps (use the documented update path or rebuild from source).
  * Improved the unsupported-platform error when resolving release naming to list the supported platforms explicitly (linux, macos, windows).

### DIFF
--- a/internal/release/release.go
+++ b/internal/release/release.go
@@ -316,7 +316,7 @@ func ReleasePlatform(goos string) (string, error) {
 	case "windows":
 		return "windows", nil
 	default:
-		return "", fmt.Errorf("unsupported release platform: %s", goos)
+		return "", fmt.Errorf("zero does not publish release builds for platform %q (supported: linux, macos, windows); update/upgrade checks aren't available, rebuild from source instead", goos)
 	}
 }
 
--- a/internal/release/release.go
+++ b/internal/release/release.go
@@ -316,7 +316,7 @@ func ReleasePlatform(goos string) (string, error) {
 	case "windows":
 		return "windows", nil
 	default:
-		return "", fmt.Errorf("zero does not publish release builds for platform %q (supported: linux, macos, windows); update/upgrade checks aren't available, rebuild from source instead", goos)
+		return "", fmt.Errorf("no release package naming defined for platform %q (release platforms: linux, macos, windows)", goos)
 	}
 }
 
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -361,7 +361,15 @@ func releasePlatform(goos string) (string, error) {
 	case "windows":
 		return "windows", nil
 	default:
-		return "", fmt.Errorf("unsupported release platform: %s", goos)
+		// No prebuilt release archives are published for this GOOS (e.g. a
+		// binary built natively on Android/Termux reports GOOS "android").
+		// Fail loudly rather than silently no-op'ing so users don't mistake
+		// a skipped check for "you're already up to date" — `zero update
+		// --check` is documented as read-only, but "safe" doesn't mean
+		// "silently wrong". Building/installing zero on such a platform
+		// falls outside the release matrix, so `zero update`/`zero upgrade`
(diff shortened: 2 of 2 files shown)


anandh8x:
The code change is a real UX win. The wording needs a tweak though.

"no published release for android (supported: linux, macos, windows). Rebuild from source instead" reads as if Termux isn't a supported platform. It is, just not as a self-updating release asset. PR #509, PR #455, PR #573, and the dictation recorder's PlatformTermux path are all in main. Termux runs Zero fine. A user reading the new message would reasonably conclude "this doesn't run here," which is wrong.

The actual distinction is "no release asset for this GOOS," not "platform unsupported." And the upgrade path shouldn't be "rebuild from source." That's hostile to a phone user without a Go toolchain. The right path is the npm wrapper (PR #455), which is the documented Termux install: `npm i -g @gitlawb/zero` to install, `npm update -g @gitlawb/zero` to upgrade.

Suggested wording:

```
no published release for %q (release assets: linux, macos, windows). Your build is the current version of record. Upgrade with `npm update -g @gitlawb/zero` to get the latest.
```

The parenthetical lists release platforms, not "supported platforms." And `npm update` is the actual one-line upgrade. No Go, no source build, works the same way it did on install.

Separate follow-up: the release-artifacts.yml matrix (line 26-31) doesn't include android-arm64. Adding it (cross-compile with GOOS=android, an Android-shaped install script, a smoke test on actual Termux) would make `zero update` stop failing on Android at all, instead of failing with a friendlier message. Not blocking this PR.

kevincodex1 (approved):

PR #606, opened 2026-07-08
fix(hooks): run sessionEnd hooks after Esc/Ctrl+C interrupts

### LINKED ISSUE #605: sessionEnd hooks don't run on Esc/Ctrl+C interrupt
## Summary

`sessionEnd` hooks configured via the hooks system (introduced in #581 / 6ec77d6) silently fail to run whenever a run is interrupted (Esc / Ctrl+C in the TUI, or an equivalent cancellation in `zero exec`).

## Where

`internal/agent/loop.go`, in `Run`:

```go
result = Result{Messages: copyMessages(messages)}
dispatchSessionStart(ctx, options)
defer func() {
    dispatchSessionEnd(ctx, options, result, err)
}()
```

## Root cause

On interrupt, `Run` returns with `ctx` already canceled (`ctx.Err() != nil`). The deferred call still passes that same canceled `ctx` into `dispatchSessionEnd` → `Hooks.Dispatch(ctx, ...)` → `context.WithTimeout(ctx, dispatcher.timeout)` (`internal/hooks/dispatch.go`). Deriving a timeout from an already-canceled parent yields an already-canceled child context, so `exec.CommandContext` never actually launches the hook process — `cmd.Run()` returns immediately with `context.Canceled`, and `runWithTimeout` marks the result `TimedOut: true`.

Net effect: the audit log records the `sessionEnd` hook as `error`/timed-out even though the run completed and generated a final `Result` — the hook body never executes. This is the *common* case for `sessionEnd` (most real sessions end via an interrupt), so any `sessionEnd` automation (e.g. notifications, cleanup, transcript export) configured via hooks is unreliable.

Only `sessionEnd` is affected — `sessionStart` and other observational hooks aren't dispatched from an already-canceled context in the same way.

## Fix

Use a context detached from `ctx`'s cancellation for the `sessionEnd` dispatch, so the hook's own dispatcher timeout still applies but it isn't short-circuited by the run's cancellation:

```go
defer func() {
    dispatchSessionEnd(context.WithoutCancel(ctx), options, result, err)
}()
```

I have a patch + a regression test ready (an end-to-end test using `cancelingProvider` that cancels the run context mid-stream, then asserts the `sessionEnd` hook's audit status is `completed` rather than `error`) and will open a PR.

PierrunoYT:
## Summary

Fixes #605.

- `sessionEnd` hooks were dispatched with the run's own (already-canceled) `ctx` on interrupt (Esc/Ctrl+C), so `Hooks.Dispatch`'s `context.WithTimeout(ctx, ...)` derived an already-canceled context and the hook process never actually launched — recorded as timed-out/error in the audit log even though the run completed normally.
- Detach the `sessionEnd` dispatch from the run context's cancellation via `context.WithoutCancel(ctx)`, so the dispatcher's own per-hook timeout still applies without being short-circuited by the run's cancellation.
- `sessionStart` and other lifecycle/tool hooks are unaffected — this is specific to the deferred `sessionEnd` call, which is the one hook dispatch guaranteed to run *after* the run context may already be canceled.

## Test plan

- [x] Added `TestRunDispatchesSessionEndHookAfterContextCancellation` (`internal/agent/loop_test.go`): a `cancelingProvider` cancels the run's context mid-stream (simulating Esc/Ctrl+C), then asserts the `sessionEnd` hook's audit status is `completed`, not `error`.
- [x] Verified the new test fails on the pre-fix code (`status = "error"`) and passes after the fix.
- [x] `go test ./internal/agent/... ./internal/hooks/...` passes.
- [x] `go vet ./internal/agent/...` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured session-end actions/hooks are dispatched and complete reliably even if a run is canceled before cleanup begins.
  * Prevents session-end steps from being skipped due to early cancellation timing.
* **Tests**
  * Added a regression test covering mid-stream/early cancellation to verify session-end hooks still reach completion.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -164,7 +164,11 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 	result = Result{Messages: copyMessages(messages)}
 	dispatchSessionStart(ctx, options)
 	defer func() {
-		dispatchSessionEnd(ctx, options, result, err)
+		// Use a context detached from ctx's cancellation: on Esc/Ctrl+C the run
+		// context is already canceled by the time this defer runs, and a canceled
+		// parent would make Hooks.Dispatch's context.WithTimeout derive an
+		// already-expired context, so the sessionEnd hook would never actually run.
+		dispatchSessionEnd(context.WithoutCancel(ctx), options, result, err)
 	}()
 	for turn := 0; turn < maxTurns; turn++ {
 		result.Turns = turn + 1
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -115,7 +115,13 @@ func (provider *cancelingProvider) StreamCompletion(ctx context.Context, request
 func TestRunDispatchesSessionEndHookAfterContextCancellation(t *testing.T) {
 	goBinary, err := exec.LookPath("go")
 	if err != nil {
-		t.Skip("go binary not on PATH")
+		goBinary = filepath.Join(runtime.GOROOT(), "bin", "go")
+		if runtime.GOOS == "windows" {
+			goBinary += ".exe"
+		}
+		if _, statErr := os.Stat(goBinary); statErr != nil {
+			t.Skipf("go binary unavailable on PATH or in GOROOT: %v", statErr)
+		}
 	}
 	audit, err := hooks.NewAuditStore(hooks.AuditStoreOptions{AuditPath: filepath.Join(t.TempDir(), "audit.jsonl")})
 	if err != nil {
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -115,7 +115,10 @@ func (provider *cancelingProvider) StreamCompletion(ctx context.Context, request
 func TestRunDispatchesSessionEndHookAfterContextCancellation(t *testing.T) {
 	goBinary, err := exec.LookPath("go")
 	if err != nil {
-		goBinary = filepath.Join(runtime.GOROOT(), "bin", "go")
+		// The test binary runs on the build machine, so its build-time GOROOT
+		// still identifies the toolchain that produced it.
+		goRoot := runtime.GOROOT() //nolint:staticcheck // Safe for this non-portable test binary.
+		goBinary = filepath.Join(goRoot, "bin", "go")
 		if runtime.GOOS == "windows" {
 			goBinary += ".exe"
 		}
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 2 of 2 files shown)


anandh8x (approved):
LGTM on the sessionEnd fix.

WithoutCancel on the deferred sessionEnd path is the right fix: interrupt already cancels the run ctx, so Hooks.Dispatch's WithTimeout was inheriting a dead parent and the hook never launched. Regression test with cancelingProvider pins it. sessionStart and tool hooks stay on the normal ctx.

Security failure is GO-2026-5856 (crypto/tls on go1.26.4), not this change. Main already has the 1.26.5 bump from #607. Rebase onto main and Security should clear.

anandh8x (changes requested):
@PierrunoYT Pulling the approve until Security is green.

The sessionEnd fix itself still looks right (WithoutCancel on the deferred path + cancelingProvider regression test). The Security failure is not introduced by this diff — it's GO-2026-5856 on crypto/tls@go1.26.4, fixed on main by #607 (go 1.26.5).

Please rebase onto latest main so go.mod is 1.26.5 and govulncheck passes. I'll re-approve once Security is green.

gnanam1990 (approved):
## Review

**Verdict:** Approve with notes. Agree with @anandh8x.

`context.WithoutCancel(ctx)` on the deferred `sessionEnd` path is the right fix: interrupt already cancels the run ctx, so `Hooks.Dispatch`'s `WithTimeout` was inheriting a dead parent. Per-hook timeout still applies; `sessionStart`/tool hooks stay on the normal ctx. Named returns keep `result`/`err` correct in the defer. Regression test with `cancelingProvider` pins it (passes locally with `-race`).

### Notes
1. **Rebase onto main** so Security CI clears (GO-2026-5856 / go1.26.4 → 1.26.5 from #607). Local merge already picks up `go 1.26.5`.
2. Nit: #605 lacks `issue-approved`.
3. Nit: prefer `runtime.GOROOT()/bin/go` before skipping the new test when `go` is not on PATH.

No disagreement with the core fix.

anandh8x (approved):
Re-reviewed at ec0db62c. The sessionEnd fix is still good and Security now passes after main picked up the go1.26.5 bump for GO-2026-5856.

Vasanthdev2004 (approved):
Built, vetted, and ran the agent + hooks tests. Approving.

This is the fix #605 asks for, pretty much verbatim: the deferred `dispatchSessionEnd` call now passes `context.WithoutCancel(ctx)` (loop.go), so on an Esc/Ctrl+C exit the hook's timeout no longer derives an already-expired context and the hook actually runs. The new regression test fails without the one-liner and passes with it, so it genuinely guards this.

I checked the four things I cared about: the interrupt path reaches the dispatch and it fires exactly once (no skip, no double-run); sessionStart is untouched; each hook is still bounded by the per-hook timeout so a hung sessionEnd can't stall shutdown indefinitely; and hook selection still flows through the trust-gated merged config, so a sessionEnd hook from an untrusted project config stays excluded. Composes cleanly with the existing gating.

One behavior change worth being aware of (not a blocker, it's the intended point of the issue): an interrupt exit now waits for sessionEnd hooks to run, up to the per-hook timeout, before returning. Bounded, but users with slow sessionEnd automation will notice.

Vasanthdev2004:
@gnanam1990 when you get a chance, could you review this one? It's the sessionEnd-hooks-on-interrupt fix (#605) — a one-liner using context.WithoutCancel, with a regression test that fails without it. Build/tests clean; a second approval would unblock it.

kevincodex1 (approved):

PR #609, opened 2026-07-09
fix(config): enforce MCP trust boundary so project config cannot override user disable

### LINKED ISSUE #512: mcp: project config can re-enable user-disabled MCP servers
## Version / branch / commit
main at commit `3c81fea`

## OS and environment

## Summary
A user-level MCP disable can be overridden by project config. If the user disables an MCP server in their user config, a repository's project config can set the same server to `disabled: false` and cause it to be active again when Zero resolves MCP servers for that workspace.

This is a trust-boundary issue because `zero mcp disable <server>` looks like an operator/user-level opt-out, but a cloned repo can reverse that choice in `.zero/config.json`.

## Steps to reproduce
1. User config disables a default MCP server:

```json
{
  "mcp": {
    "servers": {
      "firecrawl": { "disabled": true }
    }
  }
}
```

2. Project config in the workspace re-enables the same server:

```json
{
  "mcp": {
    "servers": {
      "firecrawl": { "disabled": false }
    }
  }
}
```

3. Resolve MCP config for that workspace.

I verified this locally with a temporary regression test against `config.ResolveMCP`: the resolved `firecrawl` server has `Disabled == false`.

## Expected behavior
A project config should not be able to re-enable an MCP server that the user explicitly disabled in user config. User-level disable should be sticky/authoritative unless the user re-enables it from user scope or via an explicit command/flag.

## Actual behavior
Project config wins during MCP server merging. `mergeMCPServer` applies `disabled:false` when the later layer sets the disabled field, so the resolved MCP server becomes enabled again.

Relevant code/tests:

- `internal/config/resolver.go` merges user and project MCP config in order.
- `internal/config/resolver.go` / `mergeMCPServer` applies `next.disabledSet` and assigns `base.Disabled = next.Disabled`.
- `internal/config/resolver_test.go` currently has a test showing project config can clear and re-enable a disabled MCP server.

## Impact
A repo can restore a network MCP server or stdio MCP server that the user had intentionally disabled. For default/network MCP servers this can re-enable outbound connections; for stdio MCP servers this can restore startup-time process execution for a server the user opted out of.

I did not find an existing issue or open PR covering this exact user-disable override behavior.

Ashwinhegde19:
## Summary

Fixes #512 — a user-level MCP server `disable` is now **sticky**. The project config (`.zero/config.json` in a cloned repo) can no longer override the user's explicit choice:

- Cannot re-enable a server the user disabled.
- Cannot disable a server the user explicitly enabled.

The old merge guard was asymmetric: it only blocked re-enabling a user-disabled server, so a repo that set `disabled: true` could silently defeat a user's `mcp enable <server>`. The guard is now symmetric — once a higher-trust scope has explicitly set the `disabled` flag, a lower-trust layer may not override that decision in either direction. Only the CLI override scope (`canReenable = true`) may lift a sticky user disable.

## Changes

- `internal/config/resolver.go`: `mergeMCPConfig` / `mergeMCPServer` take a `canReenable bool`. `ResolveMCP` passes `canReenable = path != options.ProjectConfigPath` (user → true, project → false). The disabled-handling guard is now symmetric.
- `internal/config/resolver_test.go`:
  - Renamed `TestResolveMCPServerLayersCanClearAndReenable` → `TestResolveMCPServerLayersCannotReenableUserDisabled`.
  - Added `TestMergeMCPStickyDisableReenable`, `TestResolveMCPCannotReenableUserDisabled`, and `TestResolveMCPUserLiftsProjectDisabled`.

## Test plan

- `go test ./internal/config/...` passes
- `go vet ./internal/config/...` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP server config merging so disable/enable precedence is applied consistently across configuration layers.
  * MCP servers disabled explicitly in user scope are now sticky and cannot be re-enabled by lower-trust project/repository settings.
  * User-scoped CLI/config can still re-enable when that explicit decision is set; non-disable fields merge as before.
  * MCP server defaults are applied before layered config resolution.
* **Tests**
  * Added/updated unit and integration tests to verify sticky-disable behavior at trust boundaries.

### DIFF
--- a/internal/config/mcp_merge.go
+++ b/internal/config/mcp_merge.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-func mergeMCPConfig(dst *MCPConfig, src MCPConfig) {
+func mergeMCPConfig(dst *MCPConfig, src MCPConfig, canReenable bool) {
 	if len(src.Servers) == 0 {
 		return
 	}
@@ -14,10 +14,14 @@ func mergeMCPConfig(dst *MCPConfig, src MCPConfig) {
 		dst.Servers = map[string]MCPServerConfig{}
 	}
 	for name, server := range src.Servers {
-		dst.Servers[name] = mergeMCPServer(dst.Servers[name], server)
+		dst.Servers[name] = mergeMCPServer(dst.Servers[name], server, canReenable)
 	}
 }
 
+// mergeProjectMCPConfig merges a project-scoped (lower-trust) MCP config into
+// dst. It never re-enables a server the user explicitly disabled or disabled a
+// server the user explicitly enabled: it merges with canReenable=false so the
+// trust-boundary guard in mergeMCPServer keeps the user's decision authoritative.
 func mergeProjectMCPConfig(dst *MCPConfig, src MCPConfig) error {
 	if len(src.Servers) == 0 {
 		return nil
@@ -27,7 +31,7 @@ func mergeProjectMCPConfig(dst *MCPConfig, src MCPConfig) error {
 	}
 	for name, server := range src.Servers {
 		base := dst.Servers[name]
-		candidate := mergeMCPServer(base, server)
+		candidate := mergeMCPServer(base, server, false)
 		if projectMCPServerTargetChanges(base, server) && hasInheritedMCPCredentialMaterial(server, candidate) {
 			return fmt.Errorf("project MCP server %q cannot override target while inheriting user credentials; set headers/env/oauth explicitly or use a new server name", name)
 		}
@@ -37,7 +41,11 @@ func mergeProjectMCPConfig(dst *MCPConfig, src MCPConfig) error {
 	return nil
 }
 
-func mergeMCPServer(base MCPServerConfig, next MCPServerConfig) MCPServerConfig {
+// mergeMCPServer merges a later config layer's MCP server into the base. The
+// canReenable flag marks the CLI-override scope, the only layer that is allowed
+// to lift a sticky user-level disable/enable decision (see the disabled-handling
+// block).
+func mergeMCPServer(base MCPServerConfig, next MCPServerConfig, canReenable bool) MCPServerConfig {
 	if strings.TrimSpace(next.Type) != "" {
 		base.Type = next.Type
 	}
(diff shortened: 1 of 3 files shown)


anandh8x (approved):
LGTM.

User-level MCP disable (and explicit enable) should be sticky against project config. Making the guard symmetric with canReenable on the project layer is the right trust boundary: repo cannot re-enable what the user disabled, and cannot flip an explicit user enable either. CLI overrides still can.

Tests cover sticky disable, project re-enable blocked, user enable sticky vs project disable, and project still able to disable a default-enabled server. All CI green.

One merge note: #597 also touches MCP project merge. If both land, the second one may need a small rebase so sticky-disable and the credential/target-change guard both apply. Not blocking this PR on its own.

kevincodex1 (approved):
LGTM

kevincodex1:
please rebase to latest main and fix conflict

Vasanthdev2004 (approved):
Reviewed #609 (fixes #512). Approving.

What I verified:
- `mergeMCPServer` now takes a `canReenable bool`. `mergeProjectMCPConfig` hardcodes `false` (project is the lower-trust scope); user config and CLI overrides pass `true`. The disabled-handling guard is symmetric — once a higher-trust scope has explicitly set the `disabled` flag (`disabledSet=true`), a lower-trust project layer can't flip it in either direction. The prior decision is captured into `baseDisabledSet`/`baseDisabled` *before* folding in `next.disabledSet`, so the guard evaluates against the state the higher-trust layer actually left behind.
- Traced the merge order in both entry points. `ResolveMCP`: user(true) -> project(false, via mergeProjectMCPConfig) -> overrides(true). `Resolve`: mergeConfig(user, true) -> mergeProjectConfig -> mergeProjectMCPConfig(false) -> applyOverrides(true). `applyEnv` doesn't touch MCP disabled. There's no path where project config flows through `canReenable=true`, so no merge-precedence bypass.
- Cases: user disables + project re-enables -> stays disabled (the bug, fixed). User explicitly enables + project disables -> stays enabled (symmetric, correct). CLI override re-enabling a sticky user disable -> allowed. Project disabling a default-enabled server the user never touched -> allowed (legitimate). Empty layer -> no flip. All correct.
- Composes cleanly with the workspace-trust gate (#529): `ExcludeProject` drops the whole project layer when untrusted, so this guard only runs when the workspace is trusted — complementary, not a duplicate. The #597 credential-boundary check in `mergeProjectMCPConfig` is unchanged in substance (only the `mergeMCPServer` signature call was updated).
- Tests cover the override-prevented case, the legit project-disable case, the reverse direction (user lifts project disable), and a regression guard for the disabledSet-before-check ordering. `go build ./...`, `go vet ./...`, `go test ./internal/config/...` all clean; gofmt clean.

Non-blocking note: the PR body describes `ResolveMCP` passing `canReenable = path != options.ProjectConfigPath`; the implementation instead hardcodes `true` for user/overrides and `false` inside `mergeProjectMCPConfig`. Functionally identical, just a different shape than the prose. Also worth being aware that the symmetric guard means a user explicitly setting `disabled: false` makes that enable sticky against a project that later tries to disable the same server — correct per the trust hierarchy (user > project), and the PR calls this out explicitly, but it is a behavior change slightly beyond the user-disable case in the issue. Happy to merge as-is.

gnanam1990 (approved):
## Verdict: Approve

Fixes #512 with a symmetric sticky `disabled` guard: once a higher-trust layer has explicitly set `disabled`, project config cannot override it in either direction. Rebased cleanly on current `main`; gate reviewers have already approved on head `9e773f0`.

### What looks good

- **Trust boundary is correct and symmetric.** `mergeMCPServer` snapshots `baseDisabledSet` / `baseDisabled` before folding in `next`, fixing the CodeRabbit ordering bug and blocking both project re-enable of user-disabled servers and project disable of user-explicitly-enabled servers.
- **Layer plumbing is clean.** Project scope goes through `mergeProjectMCPConfig` (`canReenable=false`); user config and CLI overrides use `mergeMCPConfig(..., true)`. Composes well with the credential-boundary checks from #597.
- **Tests are thorough.** Renamed regression test, new unit tests for sticky behavior (including the “project can still disable default-enabled servers” guard), and two `ResolveMCP` integration tests for sticky user-disable and user-lifting-project-disable.
- **Verification:** `make lint` clean; `go test -race -count=1 ./internal/config/...` passes.

### Nits

- The `canReenable` comment says it marks the “CLI-override scope,” but user config file merges also pass `true`. Behavior is right; the comment could say “higher-trust scopes (user config + CLI overrides)” instead.
- No `ResolveMCP`-level test for the mirror of #512: user explicitly enables (`disabled: false`), project sets `disabled: true`. The merge logic handles it, but a symmetric integration test would make the contract obvious.

### Issues

None blocking.

kevincodex1 (approved):

PR #611, opened 2026-07-09
fix(providers): update MiniMax Anthropic endpoints

octo-patch:
Reason: add target provider/model to existing provider registry

## Summary
- Update MiniMax and MiniMax CN Anthropic-compatible catalog defaults to the documented `/anthropic` base URLs.
- Adjust resolver tests to assert the MiniMax Base URL defaults.

## Test plan
- `go test ./internal/providercatalog ./internal/config`

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that MiniMax catalog selections use the correct Anthropic-compatible messages endpoints.
  * Confirmed endpoint routing works for both the global and China catalog options.


jatmn (changes requested):

## Findings

- [P1] Link an approved parent issue before submitting this external contribution
  `CONTRIBUTING.md:14-76`
  This PR was opened by a `FIRST_TIME_CONTRIBUTOR`, but it has no closing/linked issue and the description leaves the required `Fixes #` reference empty. The contribution policy requires every non-team PR to be tied to an `issue-approved` parent issue and says PRs opened without one will be closed without review. Please link the approved issue (and keep this change within its accepted scope) before continuing.

- [P1] Restore the MiniMax Anthropic base URLs without the extra version segment
  `internal/providercatalog/catalog.go:136`
  The Anthropic provider appends `/v1/messages` to every configured base URL, so the new MiniMax defaults make both international and CN catalog profiles call `/anthropic/v1/v1/messages` instead of MiniMax's `/anthropic/v1/messages` endpoint. Please revert these defaults and the matching resolver expectations to the `/anthropic` bases, and add a request-URL regression test that exercises the catalog-backed MiniMax profiles. This also completes CodeRabbit's still-valid request on these lines.

Vasanthdev2004 (changes requested):
I dug into this one because the change looks innocuous but isn't. Zero's Anthropic provider concatenates the catalog base URL with `/v1/messages` (internal/providers/anthropic/provider.go:177), so the old `https://api.minimax.io/anthropic` already resolved to the documented `https://api.minimax.io/anthropic/v1/messages`. Adding `/v1` to the base URL makes the request path `/anthropic/v1/v1/messages`  a doubled segment MiniMax won't serve. MiniMax's own docs confirm the base URL is `https://api.minimax.io/anthropic` without `/v1` (the SDK appends `/v1/messages`). Same issue on the CN variant. The resolver test was updated to assert the new value, so it passes, but it's now asserting the wrong URL.

Requesting changes  I'd keep the base URLs as they were. If MiniMax has genuinely moved to a versioned base, then we'd also need to stop appending `/v1/messages` in the provider, which is a much bigger change than this PR is scoped for.

jatmn (approved):
LGTM

octo-patch:
This work continues in #665. The replacement is based on the latest `main` branch and keeps a clean, focused commit history.

PR #612, opened 2026-07-09
fix(sandbox): make Windows token fully restricted to enforce DenyRead policies

euxaristia:
## Summary

Fixes a security vulnerability on Windows where sandboxed processes could completely bypass `DenyRead` sandbox file policies.

When creating the restricted token, we previously passed the `windowsWriteRestricted` flag (`0x08`). Under this mode, the Windows kernel only checks the restricted SID list during write operations, skipping it entirely for read operations. Consequently, NTFS deny-read ACEs associated with the capability SIDs were ignored, allowing the sandboxed process to read files located in `DenyRead` directories.

This PR removes the `windowsWriteRestricted` flag so that the token is fully restricted, enforcing SID constraints for both read and write operations.

## Changes

**`internal/sandbox/windows_token_windows.go`**
- Remove `windowsWriteRestricted` from the flags passed to `procCreateRestrictedToken.Call`.

**`internal/sandbox/runner_windows_integration_test.go`**
- Update `TestWindowsUnelevatedRealSandboxSmoke` to write a secret file to a `DenyRead` subdirectory, and assert that attempting to type/read it from the sandboxed process fails with an access-denied exit code.

## Test plan

- `go test ./internal/sandbox/...` — ok
- Verified that compiling sandbox code succeeds on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows sandbox behavior so restricted processes are less likely to retain unintended write limitations.
  * Added stronger verification that sandboxed apps cannot read files from blocked locations, with the expected failure behavior now covered by tests.

### DIFF
--- a/internal/sandbox/runner_windows_integration_test.go
+++ b/internal/sandbox/runner_windows_integration_test.go
@@ -145,12 +145,22 @@ func TestWindowsUnelevatedRealSandboxSmoke(t *testing.T) {
 
 	root := t.TempDir()
 	outside := t.TempDir()
+	privateDir := filepath.Join(root, "private")
+	if err := os.MkdirAll(privateDir, 0o700); err != nil {
+		t.Fatalf("MkdirAll private: %v", err)
+	}
+	secretFile := filepath.Join(privateDir, "secret.txt")
+	if err := os.WriteFile(secretFile, []byte("super-secret"), 0o600); err != nil {
+		t.Fatalf("WriteFile secret: %v", err)
+	}
+
 	sandboxHome := filepath.Join(root, ".zero-sandbox")
 	profile := PermissionProfile{
 		FileSystem: FileSystemPolicy{
 			Kind:                 FileSystemRestricted,
 			ReadRoots:            []string{root},
 			WriteRoots:           []WritableRoot{{Root: root, ProtectedMetadataNames: []string{".git", ".zero", ".agents"}}},
+			DenyRead:             []string{privateDir},
 			IncludePlatformRoots: true,
 			AllowTemp:            true,
 		},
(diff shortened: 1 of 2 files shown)


anandh8x (approved):
LGTM. The fix is correct and the regression test pins it.

The WRITE_RESTRICTED flag on CreateRestrictedToken enforces restricted-SID deny-ACE only for writes, not reads, so any DenyRead policy tied to the capability SIDs was silently bypassed on read paths. Removing the flag makes the token fully restricted on both paths. The new test (TestWindowsUnelevatedRealSandboxSmoke) writes a secret into a DenyRead subdir and asserts cmd type returns exit 1, which is the right way to assert the deny actually fires.

Two files, one line of code, one real-world test. No collateral changes to the flag set, the call shape, or the DACL construction. Ship it.

One thing for the commit body (not blocking): the flag change tightens read enforcement for every Windows sandbox invocation, not just DenyRead paths. Worth a one-liner noting the broader posture change so future readers don't undo it on a hunch.

gnanam1990 (approved):
## Review

**Verdict:** Approve with notes. Agree with @anandh8x on the core fix.

### Correct
Removing `windowsWriteRestricted` from `CreateRestrictedToken` is the right fix: under WRITE_RESTRICTED, restricted-SID checks apply only to writes, so capability-SID DenyRead ACEs never blocked reads. The unelevated smoke addition is the right shape for a regression pin.

### Notes (non-blocking but should not be ignored)

1. **Posture change is broader than DenyRead.** Fully restricted tokens check restricted SIDs on *all* access. That can deny reads outside capability/World/logon grants and sits in tension with the read-all/write-jail model in `profile.go`. Please document this (commit/PR body), and run the real Windows smokes (`Unelevated`, `RestrictedToken`, nested pipe) with `ZERO_SANDBOX_REAL_SMOKE=1` before merge if not already done on a Windows host.

2. **CI does not run the new pin by default.** The assertion is only in `TestWindowsUnelevatedRealSandboxSmoke`. `scripts/sandbox-smoke.sh` runs `TestWindowsRestrictedTokenRealSandboxSmoke` only. Please wire the unelevated (or a DenyRead case on the restricted-token smoke) into the path that actually runs, and consider a positive control that allowed reads still succeed.

3. **Cleanup:** `windowsWriteRestricted` is dead; `broadenWindowsRestrictedTokenDefaultDacl` comments still say WRITE_RESTRICTED-only — reword so nobody removes the DACL broaden thinking the flag is gone. Broaden still looks necessary under full restriction.

4. **Nit:** quote the `type` path and optionally assert secret bytes are absent from output.

5. **Process:** no linked parent issue / `issue-approved` — please open/link one for the DenyRead bypass audit trail.

No disagreement with the APPROVE on the one-line security fix.

kevincodex1 (approved):

PR #613, opened 2026-07-09
fix(agent,tui): resolve git branch detection when starting Zero in subdirectories

euxaristia:
## Summary

Fixes a medium-severity issue where starting Zero from a nested subdirectory of a Git repository fails to detect the current branch name (e.g. for prompt assembly or in TUI header/welcome screens).

Previously, both `gitBranchForPrompt` (in prompt assembly) and `gitBranch` (in TUI view) checked for `.git` metadata directly in the current working directory (`cwd`). In subdirectories, this check fails, causing the branch name to be omitted.

This fix exports the existing unexported `findProjectGitRoot` helper from `internal/agent` as `FindProjectGitRoot`. Both branch resolution paths are updated to first find the nearest Git repository root via this helper and resolve the branch relative to it.

## Changes

**`internal/agent/system_prompt.go`**
- Export `findProjectGitRoot` as `FindProjectGitRoot` and `hasGitMetadata` as `HasGitMetadata`.
- Update `projectGuidelines` to call the exported helper.
- Update `gitBranchForPrompt` to use `FindProjectGitRoot(cwd)` to resolve repository root, and resolve relative worktree gitdirs relative to `gitRoot`.

**`internal/agent/system_prompt_test.go`**
- Update references of `findProjectGitRoot` to `FindProjectGitRoot`.

**`internal/agent/loop_test.go`**
- Add unit test `TestGitBranchForPromptInSubdirectory`.

**`internal/tui/view.go`**
- Update `gitBranch` to use `agent.FindProjectGitRoot(cwd)` to resolve the Git repository root.

**`internal/tui/tui_fixes_test.go`**
- Add unit test `TestGitBranchInSubdirectory`.

## Test plan

- `go test ./internal/agent/...` — ok
- `go test ./internal/tui/...` — ok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved git branch detection when working from nested subdirectories, including worktree setups with relative metadata paths.
  * Branch information now resolves more reliably and returns no value when a git root cannot be found.
  * Fixed prompt context detection so project settings are loaded from the correct repository root.
  * Added tests covering branch lookup from subdirectories and git-root discovery behavior.

### DIFF
--- a/internal/agent/loop_test.go
+++ b/internal/agent/loop_test.go
@@ -2947,6 +2947,24 @@ func TestGitBranchForPromptResolvesRelativeWorktreeGitdir(t *testing.T) {
 	}
 }
 
+func TestGitBranchForPromptInSubdirectory(t *testing.T) {
+	root := t.TempDir()
+	gitdir := filepath.Join(root, ".git")
+	if err := os.MkdirAll(gitdir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(gitdir, "HEAD"), []byte("ref: refs/heads/main\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	subdir := filepath.Join(root, "sub", "dir")
+	if err := os.MkdirAll(subdir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if got := gitBranchForPrompt(subdir); got != "main" {
+		t.Fatalf("gitBranchForPrompt in subdirectory = %q, want main", got)
+	}
+}
+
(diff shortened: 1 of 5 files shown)


anandh8x (approved):
LGTM.

Right fix: walk up to the nearest .git root instead of only looking in cwd. That matches how git itself resolves the repo, and reuses the helper we already had for AGENTS.md instead of inventing a second path.

Both call sites (prompt + TUI) get the same treatment, and relative worktree gitdir paths are resolved against the worktree root so nested cwd does not break HEAD lookup. Subdir tests pin it. All CI green. Ship it.

kevincodex1 (approved):
lgtm

PR #614, opened 2026-07-09
fix(acp): make truncateHint rune-safe

euxaristia:
## Summary

Fixes a medium-severity issue where `truncateHint` slices strings by raw byte offset (`s[:max]`), which can split a multi-byte UTF-8 character (like Chinese/Japanese/Korean text, emojis, or accents) in half and produce invalid UTF-8.

This PR updates `truncateHint` to slice string values using a rune slice conversion to guarantee valid UTF-8 sequences.

## Changes

**`internal/acp/translate.go`**
- Import `unicode/utf8`.
- Change `truncateHint` to count and slice by runes rather than raw bytes.

**`internal/acp/translate_test.go`**
- Import `strings` and `unicode/utf8`.
- Add `TestToolTitleTruncateHintRuneSafe` to verify truncation does not split multi-byte characters and produces exactly 60 runes before the ellipsis.

## Test plan

- `go test ./internal/acp/...` — ok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text truncation so multi-byte characters are handled safely and display correctly.
  * Truncated hints now preserve valid Unicode and add an ellipsis consistently.

* **Tests**
  * Added coverage to verify truncation remains UTF-8 safe.
  * Confirmed the visible hint length is limited as expected before the ellipsis.

### DIFF
--- a/internal/acp/translate.go
+++ b/internal/acp/translate.go
@@ -3,6 +3,7 @@ package acp
 import (
 	"encoding/json"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/Gitlawb/zero/internal/agent"
 	"github.com/Gitlawb/zero/internal/tools"
@@ -70,8 +71,8 @@ func primaryArgHint(rawArgs string) string {
 func truncateHint(s string) string {
 	s = strings.TrimSpace(s)
 	const max = 60
-	if len(s) > max {
-		return s[:max] + "…"
+	if utf8.RuneCountInString(s) > max {
+		return string([]rune(s)[:max]) + "…"
 	}
 	return s
 }
--- a/internal/acp/translate_test.go
+++ b/internal/acp/translate_test.go
@@ -1,7 +1,9 @@
 package acp
 
 import (
+	"strings"
 	"testing"
+	"unicode/utf8"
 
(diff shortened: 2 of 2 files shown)


anandh8x (approved):
LGTM.

s[:max] on a multi-byte string is a real invalid-UTF-8 footgun. Truncating by runes is the right fix for path/title hints with CJK and emoji. Test pins valid UTF-8 and the 60-rune budget. All CI green. Ship it.

gnanam1990 (approved):
## Review

**Verdict:** Approve. Agree with @anandh8x.

`s[:max]` on multi-byte UTF-8 is a real invalid-string footgun for path/title hints (CJK, emoji). Truncating by runes is the right fix. `TestToolTitleTruncateHintRuneSafe` pins valid UTF-8 and the 60-rune budget. CI green; `go test -race ./internal/acp/` passes.

### Nits (non-blocking)
- Full `[]rune(s)` conversion is fine for short hints; byte-offset truncation is optional micro-optimization.
- Rune cuts can still split multi-codepoint emoji clusters; still valid UTF-8 and acceptable here.

Ship it.

kevincodex1 (approved):

PR #616, opened 2026-07-09
fix(daemon): handle os.ErrPermission as collision during Windows lock creation

euxaristia:
## Summary

Fixes a medium-severity issue where single-instance daemon lock creation on Windows is fragile under concurrency. 

On Windows, when a process releases the lock via `os.Remove`, the file is placed in a "delete pending" state. During this time, any racing `O_EXCL` file creation attempts fail with `ERROR_ACCESS_DENIED` (mapped to `os.ErrPermission`) rather than `ERROR_FILE_EXISTS` (mapped to `fs.ErrExist`). 

Previously, `internal/daemon/lock.go` only checked for `fs.ErrExist`, causing the acquire operation to fail immediately with a permission error instead of waiting/retrying.

This PR updates the lock checks to treat `os.ErrPermission` as a collision/contention indicator, allowing the daemon to wait/retry or correctly detect the stale status.

## Changes

**`internal/daemon/lock.go`**
- Treat `os.ErrPermission` as a collision/contention indicator during O_EXCL file creation.

## Test plan

- `go test ./internal/daemon/...` — ok

### DIFF
--- a/internal/daemon/lock.go
+++ b/internal/daemon/lock.go
@@ -52,7 +52,7 @@ func acquireLock(path string, isAlive func(pid int) bool) (*fileLock, error) {
 			}
 			return &fileLock{path: path}, nil
 		}
-		if !errors.Is(err, fs.ErrExist) {
+		if !errors.Is(err, fs.ErrExist) && !errors.Is(err, os.ErrPermission) {
 			return nil, err
 		}
 		pid, perr := readPidFile(path)


anandh8x (approved):
LGTM.

On Windows, O_EXCL during delete-pending surfaces as os.ErrPermission, not fs.ErrExist. Treating it as contention matches the cron lock path and stops the daemon single-instance lock from failing spuriously under concurrency. One-line fix, correct. All CI green. Ship it.

gnanam1990 (approved):
## Review

**Verdict:** Approve with notes (independent pass; agree with @anandh8x).

On Windows, `O_EXCL` during delete-pending surfaces as `os.ErrPermission`, not `fs.ErrExist`. Treating it as collision so the existing PID/stale-reclaim path runs matches cron, oauth, and securefile and fixes spurious single-instance acquire failures under start/stop races. Reclaim remains rename-aside (not blind `Remove`). One-line change; CI green including windows-latest; `go test -race ./internal/daemon/` passes.

### Notes (non-blocking)
1. True ACL `ErrPermission` can now be misreported as `ErrAlreadyRunning` after two attempts — same tradeoff accepted in cron/oauth.
2. No dedicated test for the `ErrPermission` collision branch (optional stub).
3. Still only two attempts with no backoff (pre-existing).

Ship it.

kevincodex1 (approved):

PR #618, opened 2026-07-09
fix(tools): preserve SysProcAttr during PTY fallback

euxaristia:
## Summary

Fixes a medium-severity issue where the PTY fallback path resets and wipes out custom `SysProcAttr` configurations.

On Windows, `applyWindowsShellCommandLine` sets `SysProcAttr` with a custom `CmdLine` for unescaped quoting. However, when falling back from a failed PTY setup (`startPTYProcessFunc`) to a standard pipe execution (`startPipeProcess`), `resetExecCommandForPipeFallback` would set `command.SysProcAttr = nil`. This stripped the unescaped command-line formatting and corrupted command arguments under the fallback tier.

This PR updates the fallback path to save and restore the original `SysProcAttr` configuration.

## Changes

**`internal/tools/exec_transport.go`**
- Import `syscall`.
- Capture `command.SysProcAttr` in `startExecProcess` before attempting PTY startup.
- Restore the original `SysProcAttr` in `resetExecCommandForPipeFallback` instead of resetting it to `nil`.

**`internal/tools/exec_command_test.go`**
- Add `TestStartExecProcessPTYFallbackPreservesSysProcAttr` to assert that fallback restores the original `SysProcAttr`.

## Test plan

- `go test ./internal/tools/...` — ok

### DIFF
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -422,6 +422,34 @@ func TestStartExecProcessFallsBackAfterPTYStartMutation(t *testing.T) {
 	}
 }
 
+func TestStartExecProcessPTYFallbackPreservesSysProcAttr(t *testing.T) {
+	original := startPTYProcessFunc
+	t.Cleanup(func() { startPTYProcessFunc = original })
+	startPTYProcessFunc = func(command *exec.Cmd, _ *execOutputBuffer) (io.WriteCloser, func(), error) {
+		return nil, nil, errors.New("pty start failed")
+	}
+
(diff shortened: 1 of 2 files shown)


anandh8x (approved):
LGTM.

PTY fallback should clear PTY-only wiring, not SysProcAttr. On Windows that attr carries CmdLine for unescaped shell quoting, so nil-ing it corrupts the pipe fallback path. Save/restore is the right fix. Test pins it. All CI green. Ship it.

kevincodex1 (approved):

PR #619, opened 2026-07-09
fix(cli): prevent consuming positional arguments as flag values

euxaristia:
## Summary

Fixes a medium-severity issue where positional prompt text placed immediately after flags requiring values (e.g. `--auto`) is greedily consumed as their value by the CLI argument parser.

For example, running `zero exec --auto "do a build"` would cause `--auto` to consume `"do a build"` as the autonomy level, failing with `Invalid autonomy level "do a build"`, leaving the prompt empty.

This PR refines the parser's `nextFlagValue` function to validate value format and predefined choices (autonomy levels, notify modes, output/input formats, reasoning efforts, and numeric counts) before consuming the argument.

## Changes

**`internal/cli/exec_parse.go`**
- Refine `nextFlagValue` to check the flag name and assert that the next argument matches expected values/types (e.g., choices for `--auto`/`--notify`, integers for `--depth`/`--max-turns`).

**`internal/cli/exec_parse_prompt_test.go`**
- Add `TestParseExecArgsPreventsPositionalFromBeingConsumedAsFlagValue` to verify that invalid/positional values are rejected for these flags.

## Test plan

- `go test ./internal/cli/...` — ok

### DIFF
--- a/internal/cli/exec_parse.go
+++ b/internal/cli/exec_parse.go
@@ -461,6 +461,49 @@ func nextFlagValue(args []string, index int, flag string) (string, int, error) {
 	if next == "" || flagValueLooksLikeOption(next) {
 		return "", index, execUsageError{fmt.Sprintf("%s requires a value", flag)}
 	}
+
+	// Validate specific flag values to prevent consuming positional prompt text
+	switch flag {
+	case "--auto":
+		switch strings.ToLower(next) {
+		case "low", "medium", "high", "member":
+		default:
+			return "", index, execUsageError{fmt.Sprintf("Invalid autonomy level %q. Expected low, medium, or high.", next)}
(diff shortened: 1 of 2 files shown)


anandh8x (approved):
LGTM.

nextFlagValue greedily taking the next token is a real footgun for enum/int flags. Validating --auto, --notify, formats, reasoning-effort, --depth, and --max-turns before consume fails fast with a clear error instead of eating free text as a flag value.

Note: this rejects invalid values; it does not reassign them as the positional prompt. Correct usage is still e.g. zero exec --auto medium "do a build". That is fine.

Tests cover the bad cases. All CI green. Ship it.

kevincodex1 (approved):

PR #620, opened 2026-07-09
fix(tui): resolve pending askUser callbacks to prevent runner hangs

euxaristia:
## Summary

Fixes a medium-severity issue where the TUI does not invoke the answer callback of a discarded `askUserRequestMsg` during cancellation or stale runs, causing goroutine leaks or runner hangs.

This PR ensures:
- If a stale/superseded `askUserRequestMsg` (where `msg.runID != m.activeRunID`) is handled, the callback is invoked with `nil` immediately.
- If a run is completed or canceled (`agentResponseMsg`) while an `ask_user` prompt is pending (`m.pendingAskUser != nil`), the callback is invoked with `nil` before clearing the state.

This matches the robust, fail-safe callback resolution mechanism used for `permissionRequestMsg`.

## Changes

**`internal/tui/model.go`**
- Invoke `msg.answer(nil)` for stale `askUserRequestMsg` events.
- Invoke `m.pendingAskUser.answer(nil)` when discarding pending askUser states inside the `agentResponseMsg` handler.

**`internal/tui/ask_user_test.go`**
- Add `TestAskUserStaleRequestResolvesEmpty` to assert that stale prompts resolve immediately.
- Add `TestAskUserCompleteClearsAndResolves` to assert that pending prompts resolve when the run completes/cancels.

## Test plan

- `go test ./internal/tui/...` — ok

### DIFF
--- a/internal/tui/ask_user_test.go
+++ b/internal/tui/ask_user_test.go
@@ -417,3 +417,55 @@ func TestAskUserEmptyRequestResolvesImmediately(t *testing.T) {
 		t.Fatalf("an empty request should resolve immediately, got %#v", answers)
 	}
 }
+
+func TestAskUserStaleRequestResolvesEmpty(t *testing.T) {
+	var answers [][]string
+	m := newModel(context.Background(), Options{})
+	m.pending = true
+	m.activeRunID = 7 // current run
+
+	// Message for stale runID = 5
+	updated, _ := m.Update(askUserRequestMsg{
+		runID:   5,
+		request: agent.AskUserRequest{ToolCallID: "c", Questions: []agent.AskUserQuestion{{Question: "Proceed?"}}},
+		answer:  func(values []string) { answers = append(answers, values) },
+	})
+	next := updated.(model)
+	if next.pendingAskUser != nil {
+		t.Fatal("a stale request must not create a pending prompt")
+	}
+	if len(answers) != 1 || answers[0] != nil {
+		t.Fatalf("expected stale request callback to resolve immediately with nil, got: %v", answers)
+	}
+}
+
+func TestAskUserCompleteClearsAndResolves(t *testing.T) {
+	var answers [][]string
+	m := newModel(context.Background(), Options{})
+	m.pending = true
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
## Verdict: Approve

Small, targeted fix that closes a real deadlock class. Mirrors the existing `permissionRequestMsg` fail-safe pattern. Safe to merge.

## What looks good

- Root cause is correct: the agent goroutine blocks on `answerCh` in `OnAskUser`; dropping an `askUserRequestMsg` or clearing `pendingAskUser` without invoking the callback can park it indefinitely — same class of bug fixed earlier for permissions.
- Stale-run handling matches permission semantics (`msg.answer(nil)` when `runID != activeRunID`).
- `agentResponseMsg` cleanup resolves any in-flight prompt before clearing state.
- Tests (`TestAskUserStaleRequestResolvesEmpty`, `TestAskUserCompleteClearsAndResolves`) directly cover the two failure modes and match the style of `permission_deadlock_test.go`.
- `nil` answers are semantically fine — `FormatAskUserAnswers` treats an all-empty/nil answer set as a wholesale dismissal.
- No double-invoke risk: normal submit clears `pendingAskUser` first; new guards only fire when state is still pending.

## Nits

- No linked issue (`Fixes #…` / `issue-approved`) — fine as a follow-up to the permission-deadlock work, but a one-line issue link would help traceability.
- `nil` vs `[]string{}` inconsistency between discard paths and empty-question auto-resolve — behavior is equivalent today.
- Optional future test: `cancelRun()` while `pendingAskUser` is set (pre-existing gap also applies to `pendingPermission`).

## Issues

None blocking.

anandh8x (approved):
LGTM.

Dropping an ask_user prompt without calling answer leaves the agent blocked on the reply channel. Resolving with nil on stale runID and on agentResponseMsg matches the permission-request fail-safe and closes the hang. Tests cover both paths. All CI green. Ship it.

kevincodex1 (approved):

PR #623, opened 2026-07-09
Add local HTTP/OpenAPI server and JS SDK

### LINKED ISSUE #622: Add local HTTP/OpenAPI server and JS SDK for ZERO automation
### Problem

ZERO currently has strong CLI/TUI/ACP flows and zero serve --mcp, but it does not expose a local HTTP control plane for automation, SDK usage, or a future Web UI.

Today, external tools have to shell out to zero exec, parse stream-json output, or use MCP where HTTP session control is not the right fit. That makes it harder to build browser UIs, editor integrations, automation scripts, and typed clients around ZERO sessions.

### Requested behavior

Add a v1 local trusted HTTP server:
`zero serve --http --hostname 127.0.0.1 --port 4096`

**Requested surface:**

Bearer-token auth by default, with generated local token
Optional ZERO_SERVER_TOKEN
--no-auth only on loopback
CORS support
OpenAPI 3.1 at /openapi.json
SSE event stream at /event
Session APIs: create/list/get/update/fork/tree/lineage/event-log
Prompt APIs: sync /message and async /prompt_async
Approval APIs for permission requests and ask-user prompts
Abort support
File read/search APIs confined to project root
JS SDK package: @gitlawb/zero-sdk

### Why it belongs in Zero

This is platform infrastructure for ZERO.

**It enables:**

Future Web UI
Typed SDK integrations
Local automation
Editor extensions
CI/dev tooling
Session browser/chat UI
External clients that need approvals, streaming, and aborts
ZERO already has sessions, event logs, stream-json events, and agent.Run; exposing them through a local HTTP/OpenAPI layer is a natural next step.

_No response_

_No response_

GautamBytes:
Closes #622

## Summary
- Adds `zero serve --http` with local bearer auth, loopback-only no-auth, CORS, OpenAPI, SSE, session, run, approval, abort, file, and search endpoints.
- Adds `internal/httpapi` with stdlib `net/http`, run manager, permission broker, ask-user broker, JSON error shape, and path confinement.
- Runs HTTP prompts in-process through `agent.Run` so permission and ask-user callbacks can block and resume over HTTP.
- Adds a dependency-free JS SDK package at `sdk/js` as `@gitlawb/zero-sdk`.
- Keeps `zero serve --mcp` behavior unchanged.

## Security and scope
- This is a v1 local trusted automation server, not a hosted multi-user server.
- Auth fails closed unless `--no-auth` is explicitly used on loopback.
- CORS rejects untrusted origins, JSON bodies are capped and content-type checked, file access is confined to the workspace, and HTTP errors are redacted.

## Notes
- Includes one small TUI test fixture hardening line so `go test ./...` passes from this long worktree path.
- Does not include generated local review/threat-model report files.

## Summary by CodeRabbit

* **New Features**
  * Added `zero serve --http` mode with HTTP/OpenAPI docs, auth (env token or auto-generated), CORS, session management, and streamed run execution/events.
  * Added an internal HTTP API layer (including SSE event streaming) and workspace file browsing/search endpoints.
  * Introduced a JavaScript SDK with typed models and SSE subscription.
* **Bug Fixes**
  * Strengthened request validation, authentication/CORS behavior, workspace confinement (including symlink escape checks), and domain error responses.
* **Tests**
  * Expanded integration coverage for HTTP mode, streaming/brokers, permission/ask synchronization, async runs, and file routing.
* **Documentation**
  * Updated JS SDK README and TypeScript typings.

### DIFF
--- a/internal/cli/http_runner.go
+++ b/internal/cli/http_runner.go
@@ -1,0 +1,371 @@
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/agent"
+	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/httpapi"
+	"github.com/Gitlawb/zero/internal/mcp"
+	"github.com/Gitlawb/zero/internal/modelregistry"
+	"github.com/Gitlawb/zero/internal/sandbox"
+	"github.com/Gitlawb/zero/internal/sessions"
+	"github.com/Gitlawb/zero/internal/streamjson"
+	"github.com/Gitlawb/zero/internal/tools"
+	"github.com/Gitlawb/zero/internal/usage"
+)
+
+type httpRunner struct {
+	workspaceRoot string
+	deps          appDeps
+	stderr        io.Writer
+}
+
+func newHTTPRunner(workspaceRoot string, deps appDeps, stderr io.Writer) httpapi.Runner {
+	return &httpRunner{workspaceRoot: workspaceRoot, deps: deps, stderr: stderr}
+}
+
+func (runner *httpRunner) Run(ctx context.Context, request httpapi.RunRequest, hooks httpapi.RunHooks) (httpapi.RunResult, error) {
+	modelRegistry, _ := modelregistry.DefaultRegistry()
+	overrides := config.Overrides{}
+	if request.Model != "" {
+		resolvedModel, notice := resolveSelectedModel(modelRegistry, request.Model)
+		overrides.Provider.Model = resolvedModel
+		if notice != "" && runner.stderr != nil {
+			_, _ = fmt.Fprintln(runner.stderr, notice)
+		}
+	}
+	resolved, err := runner.deps.resolveConfig(runner.workspaceRoot, overrides)
+	if err != nil {
+		return httpapi.RunResult{}, err
+	}
+	if !config.HasProviderProfile(resolved.Provider) {
+		return httpapi.RunResult{}, fmt.Errorf("no provider configured")
+	}
+
+	permissionMode, err := resolveHTTPPermissionMode(request)
+	if err != nil {
+		return httpapi.RunResult{}, err
+	}
+	runReasoningEffort := strings.TrimSpace(request.ReasoningEffort)
+	if runReasoningEffort != "" {
+		if notice := reasoningEffortNotice(modelRegistry, resolved.Provider.Model, runReasoningEffort); notice != "" && runner.stderr != nil {
+			_, _ = fmt.Fprintln(runner.stderr, notice)
+		}
+	}
+	forwardEffort := forwardedReasoningEffort(modelRegistry, resolved.Provider.Model, runReasoningEffort)
+
+	execScope, err := sandbox.NewScope(runner.workspaceRoot, resolved.Sandbox.AdditionalWriteRoots)
+	if err != nil {
+		return httpapi.RunResult{}, err
+	}
+	registry := newCoreRegistryScoped(runner.workspaceRoot, execScope)
+	registerLocalControlTools(registry, runner.workspaceRoot, resolved.LocalControl)
+	var specialistRuntime *agentToolRuntime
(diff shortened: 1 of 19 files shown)


jatmn (changes requested):

## Findings

- [P1] Complete CodeRabbit's control-event delivery request for late SSE subscribers
  `internal/httpapi/broker.go:55`
  The follow-up only waits for a control event when a subscriber is already present. `prompt_async` starts the run before its caller can open `/event`; if the run immediately requests permission or asks the user, `publish` snapshots no targets and discards the opaque request ID. The run then waits in the pending broker with no endpoint that can recover that ID, so it cannot be approved. Please retain/replay pending permission and ask events to a later matching subscription (or establish the subscription before accepting the async run). This completes the resolved CodeRabbit request that control events must not be silently lost.

- [P1] Reject permission decisions that were not offered for the request
  `internal/httpapi/server.go:497`
  This endpoint accepts every non-empty action, while the broker discards `PermissionRequest.AvailableDecisions`. The agent loop normalizes values but does not enforce that set, so an API client can submit `always_allow` or `always_allow_prefix` for a prompt that offered only a one-time decision; those actions can persist a grant. Store the allowed decisions with the pending request and return a 4xx for actions outside that list.

- [P2] Keep the SDK constructor declaration consistent with its runtime requirement
  `sdk/js/index.d.ts:232`
  The new optional parameter lets TypeScript callers write `createZeroClient()`, but the implementation immediately rejects that invocation because `baseUrl` is empty. Keep `options` required (as it was before the follow-up) or supply an actual runtime default base URL.

anandh8x:
Thanks for the thorough work on this  the HTTP server and SDK are a solid v1 design, and the security defaults show real care.

However, it adds a full local HTTP server + JS SDK which is basically plumbing for a web UI. We don’t think we need that right now. We don’t have a web product planned anytime soon, and we'd rather keep fixing bugs and polishing what Zero already does (CLI, TUI, MCP, sandbox) instead of taking on a whole new surface we have to maintain forever. 

Also, CONTRIBUTING.md requires community PRs to target an issue with the `issue-approved` label before implementation, which #622 didn't have. We're closing this for now so you aren't stuck waiting in review limbo. If we eventually green-light an HTTP automation use case, we'd be happy to revisit this. Appreciate the effort regardless.

Vasanthdev2004 (commented):
Reviewed this against current main — build, vet, gofmt, and the httpapi/cli/tui tests all pass, including the scroll-test fixture this PR pins (that was the path-length-sensitive test I'd been skipping on Windows; pinning cwd fixes it cleanly, not a mask). The security model is solid: auth fails closed (empty token → 500, bad bearer → 401), --no-auth is rejected off loopback, withAuth wraps every route including SSE, file paths are confined to the workspace with symlink-escape checks, request bodies are capped and content-type-checked, and errors are redacted. The part I cared most about — an HTTP run can't bypass the permission gate — checks out: a tool needing permission publishes a request over SSE and blocks until a human POSTs a decision, and the decision action is allowlisted against what the agent offered, so there's no auto-approve or forged-allow path.

Holding for two things, both kevin's. Issue #622 is still open with no maintainer sign-off — like #653, this is a large feature landing against the author's own open issue, so I'd want your go-ahead on scope before it merges. And the branch is off the 0.3.0 release commit, seven commits behind main; it merges clean, but a rebase onto current main would let it pick up the workspace-trust gating from #529 and the Go 1.26.5 bump in tests. Minor nit: the bearer-token check in withAuth uses a plain string compare rather than constant-time — low-stakes for a local server, but crypto/subtle.ConstantTimeCompare would be the careful choice.

Vasanthdev2004:
@gnanam1990 when you have time — #623 is the other big one from this contributor: a local HTTP/OpenAPI server + JS SDK (+3691, 19 files) with bearer auth, loopback-only --no-auth, CORS, SSE, sessions, workspace-confined file/find endpoints, and a permission/ask-user broker that blocks HTTP runs on permission requests. I reviewed it inline and the security model looks right (fail-closed auth, symlink-aware path confinement, no auto-approve path), but I'd value your independent read on the permission broker and SSE control-event flow. I'm holding it for an owner call on issue #622 and a rebase, not on the code.

PR #626, opened 2026-07-10
feat(npm): ship the native binary as platform optionalDependencies

kevincodex1:
## Summary

Makes `npm install -g @gitlawb/zero` fully silent and self-contained by replacing the postinstall downloader with the platform-package model used by Codex, esbuild, and Biome (fixes the EBADENGINE + allow-scripts warnings reported on 0.3.0 installs).

- **Platform payloads**: suffixed versions of the same package (`@gitlawb/zero@X.Y.Z-<platform>-<arch>`, referenced via `npm:` aliases in `optionalDependencies`) carry the binary, sandbox helpers, and the vendored agent-browser/tuistory `helpers/` tree already staged into release archives. Same package name = the existing npm trusted-publisher config covers everything.
- **Published wrapper has no scripts and no dependencies** — nothing left in the tree for npm/Bun/pnpm to warn about or block. The repo package.json keeps the helper deps purely as vendoring pins.
- **First-run fallback**: the old checksum-verified downloader still ships and is invoked by the wrapper when no platform package is installed (`--omit=optional`, unsupported platforms).
- **Assembly** (`scripts/npm/build-platform-packages.mjs`): materializes npm's symlink `.bin` shims into regular `sh` shims (npm pack silently drops symlinks — verified empirically) and prunes agent-browser's other-platform binaries (~65 MB per package; 78.8 → 47.5 MB packed).
- **Publish job**: platform versions publish first under `--tag platform`; CI asserts the `latest` dist-tag survived (platform versions are semver prereleases of the wrapper version) and that it equals the wrapper version after the wrapper publish.
- `zero update` npm detection keeps working unchanged (platform payload package.json is also named `@gitlawb/zero`).

Architecture + publishing rules documented in `docs/NPM_PACKAGING.md`; README/INSTALL Bun trust-ceremony sections deleted (obsolete).

## Verification

- `go test ./...` passes; new npmwrapper tests cover platform-package resolution/precedence, first-run fallback wiring, and the assembly script end-to-end (`-race` clean); gofmt clean.
- End-to-end on a real archive (`zero-release package` → assemble → simulated install layout): wrapper execs the platform-package binary; agent-browser and tuistory run from the materialized shims; with the platform package removed, first-run download from the real v0.3.0 GitHub Release works and is idempotent.

## Checklist

- [ ] The linked issue already has the `issue-approved` label. (No linked issue — follows up the 0.3.0 install-warnings report.)
- [x] `gofmt` clean.
- [x] UI changes include screenshots or a short recording where possible. (N/A — no UI changes.)

## Summary by CodeRabbit

* **New Features**
  * Updated npm install flow to use a wrapper package with optional platform payloads fetched directly from the npm registry.
  * If optional payloads are missing, the wrapper performs a first-run download from the matching GitHub Release with checksum verification.
  * Improved native binary resolution and guidance when the native binary isn’t available.
* **Documentation**
  * Refreshed npm/Bun/pnpm/yarn installation docs (including Chinese docs).
  * Added detailed npm packaging documentation and expanded the wrapper smoke checklist.
* **Bug Fixes**
  * Strengthened platform package assembly/validation and publish ordering to prevent `latest` being overwritten and ensure the wrapper version matches.
* **Tests**
  * Added/updated integration and wrapper smoke coverage for platform packaging and first-run fallback behavior.

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -97,16 +97,21 @@ jobs:
           gh release create "$tag" --repo "$GITHUB_REPOSITORY" --title "$tag" --generate-notes 2>/dev/null || true
           gh release upload "$tag" dist/release/* --repo "$GITHUB_REPOSITORY" --clobber
 
-  # Publishes the npm wrapper (@gitlawb/zero) once every platform's release
-  # assets are live — the package's postinstall downloads its binary from the
-  # GitHub Release, so publishing before the assets exist would ship a package
-  # that fails to install.
+  # Publishes @gitlawb/zero to npm: first the five platform payloads (versions
+  # of the SAME package at suffixed versions like X.Y.Z-linux-x64, carrying the
+  # native binary + vendored helpers), then the wrapper at X.Y.Z whose
+  # optionalDependencies alias those platform versions. Platform versions MUST
+  # publish first — the wrapper pins them exactly, so the reverse order would
+  # open a window where installs resolve aliases that 404. See
+  # docs/NPM_PACKAGING.md for the full contract.
   #
   # Auth is npm OIDC trusted publishing: no NPM_TOKEN secret. The package's
   # npmjs.com settings must list this repo + workflow file as a trusted
   # publisher, and provenance is then generated automatically. Note npm only
   # allows configuring a trusted publisher on an EXISTING package, so the very
   # first @gitlawb/zero publish is a one-time manual `npm publish` bootstrap.
+  # Because the platform payloads are versions of the same package, they reuse
+  # that single trusted-publisher configuration.
   publish-npm:
     name: Publish npm package
     needs: package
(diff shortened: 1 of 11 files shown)


kevincodex1:
@coderabbitai please do a full review

jatmn (changes requested):

## Findings

- [P1] Make the Windows-on-ARM fallback actually resolve the x64 payload
  `bin/zero.js:63`
  On a native Windows ARM64 Node installation the wrapper resolves only `@gitlawb/zero-win32-arm64`, but the generated optional-dependency matrix contains no such package. It then invokes the downloader, which explicitly skips `win32/arm64` without creating a binary. Consequently `zero` ends at the “No native binary” error, despite the new packaging documentation saying Windows on ARM runs the x64 package under emulation. Resolve that platform to an installable x64 payload/downloader (or remove the unsupported claim); the current tests only assert the skip and never exercise this end-to-end path.

- [P2] Include the project license in both generated publish payloads
  `scripts/npm/build-platform-packages.mjs:317`
  The release staging code does not put `LICENSE` in the archives, and this assembler copies no `LICENSE` into either platform directory or the wrapper directory. Since both `npm publish` calls run from those generated directories, the tarballs lose the MIT permission notice that npm currently includes automatically when publishing from the repository root. Copy the root `LICENSE` into each generated package and assert it is present in the assembly/pack tests.

kevincodex1:
do a full codereview @coderabbitai

PR #627, opened 2026-07-10
fix(plugins): resolve relative executable paths against plugin root

euxaristia:
## Summary

Fixes a High-severity issue where relative command paths defined by plugins (e.g. `"command": "./tools/helper.sh"`) can be hijacked by malicious files placed at that same path inside an untrusted workspace.

When a plugin tool runs, `RunWithOptions` executes the plugin command within the caller's workspace CWD. If the command is a relative path (e.g. starting with `./` or containing separators), Go's `os/exec` resolves it relative to the workspace directory.

This PR resolves relative executable paths containing separators to absolute paths inside the plugin's install directory during activation.

## Changes

**`internal/plugins/activate.go`**
- In `expandPluginRootPath`, check if the command path is not absolute (`!filepath.IsAbs` and `!isWindowsAbs`) and contains path separators. If so, resolve it to an absolute path under the plugin directory using `filepath.Join`.

**`internal/plugins/activate_test.go`**
- Add `TestExpandPluginRootPathRelative` to assert that relative executable paths are resolved, absolute paths are kept verbatim, and placeholders are correctly expanded.

## Test plan

- `go test ./internal/plugins/...` — ok

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed plugin executable path resolution for relative values that include directory separators (including `/` and `\`).
  * Relative paths now resolve correctly against the plugin directory, while bare executable names and absolute Unix/Windows paths remain unchanged.
  * Enhanced handling for parent-directory traversal and Windows absolute-path detection.
* **Tests**
  * Added coverage for relative and placeholder-expanded plugin root path cases.

### DIFF
--- a/internal/plugins/activate.go
+++ b/internal/plugins/activate.go
@@ -620,14 +620,12 @@ func (tool pluginTool) meta() map[string]string {
 // (parsePermission clamps allow→prompt otherwise), so honoring it does not
 // silently auto-approve a mutating plugin tool.
 func toolSafety(plugin LoadedPlugin, ext ToolExtension) tools.Safety {
-	permission := tools.PermissionPrompt
+	var permission tools.Permission
 	switch ext.Permission {
 	case PermissionAllow:
 		permission = tools.PermissionAllow
 	case PermissionDeny:
 		permission = tools.PermissionDeny
-	case PermissionPrompt:
-		permission = tools.PermissionPrompt
 	default:
 		permission = tools.PermissionPrompt
 	}
--- a/internal/plugins/activate.go
+++ b/internal/plugins/activate.go
@@ -460,6 +460,9 @@ func expandPluginRoot(value string, pluginDir string) string {
 // e.g. URLs or flag values).
 func expandPluginRootPath(value string, pluginDir string) string {
 	if !strings.Contains(value, pluginRootPlaceholder) {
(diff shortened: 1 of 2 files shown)


internal/plugins/activate.go
kevincodex1: can you address this @euxaristia

Vasanthdev2004 (approved):
Pulled this into a worktree against current main. The core fix anchors relative command paths (anything with a path separator) to the plugin's install directory via filepath.Join, so a relative executable can't be hijacked by resolving against the caller's CWD; absolute paths and bare binary names pass through unchanged, which is the right call. Build, vet, and the plugins tests all pass, including the new TestExpandPluginRootPathRelative.

One note: ../ escapes that climb out of the plugin root are accepted rather than rejected, but I think that's fine since a manifest author can already specify an arbitrary absolute path and untrusted project-scoped plugins are already dropped by the workspace-trust gate. Approve.

Vasanthdev2004:
@gnanam1990 when you get a chance — small one: #627 anchors relative plugin exec paths to the plugin install dir so they can't resolve against the caller's CWD. I approved it; the one judgment call is that ../ escapes out of the plugin root are accepted rather than rejected. Worth a second look on the trust angle if you have a minute.

euxaristia:
Fixed the ineffassign failure on line 623: the initial permission assignment in toolSafety was always overwritten by the switch below it, including a case that just reassigned the same PermissionPrompt value. Removed the redundant initializer and let the switch be exhaustive with a default.

kevincodex1 (approved):

PR #631, opened 2026-07-10
fix(securefile,credstore): call Sync on temp file before close and rename

euxaristia:
## Summary

Fixes a High-severity durability issue where credential and secret files could be corrupted or lost (0-byte file) if a system crash or power failure occurs shortly after a write.

Although a write-to-temp-then-rename pattern is used, neither `securefile` nor `credstore` called `tmp.Sync()` (fsync) to flush the page cache and directory metadata to physical disk before renaming.

This PR adds `Sync()` calls (and checks their errors) before closing and renaming.

## Changes

- In `internal/securefile/securefile.go`: Call `tmp.Sync()` in `writeNewSecretFile`.
- In `internal/credstore/credstore.go`: Call `tmp.Sync()` in `Store.write`.

## Test plan
- `go test -race ./internal/securefile/... ./internal/credstore/...` — ok

### DIFF
--- a/internal/credstore/credstore.go
+++ b/internal/credstore/credstore.go
@@ -251,6 +251,10 @@ func (s *Store) write(data map[string]string) error {
 		_ = tmp.Close()
 		return fmt.Errorf("credstore: write: %w", err)
 	}
+	if err := tmp.Sync(); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("credstore: sync: %w", err)
+	}
 	if err := tmp.Close(); err != nil {
 		return fmt.Errorf("credstore: write: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


Vasanthdev2004 (approved):
Verified the fsync-before-rename idiom is applied on the right paths and in the right order in both internal/credstore/credstore.go (Store.write) and internal/securefile/securefile.go (writeNewSecretFile): tmp.Sync() is called while the file is still open, then Close, then Rename. Sync errors are checked and the temp file is closed on that path; the existing defer os.Remove(tmpPath) cleans up. os.File.Sync maps to FlushFileBuffers on Windows, so this is correct here too. go build ./..., go vet, and go test on both packages pass, and the PR merges cleanly into current main (its base is a few commits behind but no conflicts). It composes fine with #641, which touches disjoint stores. Approving.

kevincodex1 (approved):

PR #615, opened 2026-07-09
fix(securefile): reclaim stale lock files to prevent permanent DOS

euxaristia:
## Summary

Fixes a medium-severity issue where stale `.lock` files left behind after abrupt process crashes (e.g. SIGKILL or system panic) cause all future attempts to create/acquire secrets to fail with a permanent timeout, resulting in a denial of service (DOS).

This PR brings the stale lock reclamation logic from `internal/cron/lock.go` and `internal/oauth/lock.go` over to `internal/securefile/securefile.go`:
- Write process-specific owner tokens to the `.lock` file.
- Inspect the lock file modification time if O_EXCL acquisition fails.
- Reclaim lock files older than 10 seconds (`secureLockStaleAfter`) using the same atomic rename-aside routine (`reclaimStaleLock`).

## Changes

**`internal/securefile/securefile.go`**
- Define `secureLockStaleAfter` and atomic counter `secureLockSeq`.
- Update `createSecretFile` to write owner tokens, release only owned lock files, and detect and reclaim stale locks.
- Add `reclaimStaleLock` helper.

**`internal/securefile/securefile_test.go`**
- Add `TestCreateSecretFileReclaimsStaleLock` to verify that stale locks are correctly reclaimed.

## Test plan

- `go test ./internal/securefile/...` — ok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret file creation to safely detect and reclaim stale lock files.
  * Updated lock handling so cleanup only occurs when the lock content still matches the current operation.
  * Strengthened error handling around lock-file writing and lifecycle to reduce leftover locks.
* **Tests**
  * Added a new test to verify stale lock reclamation during secret creation.

PR #641, opened 2026-07-10
fix(windows): retry atomic file renames on Access is denied / Sharing violation

euxaristia:
This PR introduces a renameWithRetry helper to replace os.Rename for atomic file overwrites in sessions, sandbox, cron, and swarm. This helper retries the rename operation up to 10 times with a 10ms backoff when encountering transient Windows sharing violations (error 32) or access denied (error 5) errors, which frequently occur under concurrency on Windows (e.g. from background indexing or delayed file handle releases).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows reliability for atomic file updates by retrying transient rename failures caused by permission issues or file-sharing violations.
  * Enhanced durability for cron/session writes and Windows “unelevated setup” marker updates under contention.

* **Tests**
  * Added/expanded Windows-focused coverage to verify mailbox rename retry behavior, including retryable vs non-retryable error handling.
  * Improved visibility into concurrent mailbox send failures during test runs.

### DIFF
--- a/internal/cron/store.go
+++ b/internal/cron/store.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
+	"syscall"
 	"time"
 )
 
@@ -144,7 +146,7 @@ func (s *Store) writeJob(job Job) error {
 	if err := os.WriteFile(tmp, data, 0o600); err != nil {
 		return err
 	}
-	return os.Rename(tmp, filepath.Join(dir, "metadata.json"))
+	return renameWithRetry(tmp, filepath.Join(dir, "metadata.json"))
 }
 
 func (s *Store) Get(id string) (Job, error) {
@@ -321,4 +323,31 @@ func (s *Store) Runs(id string) ([]RunRecord, error) {
 		}
 	}
 	return runs, scanner.Err()
+}
+
+func renameWithRetry(src, dst string) error {
+	var err error
+	for i := 0; i < 10; i++ {
+		err = os.Rename(src, dst)
+		if err == nil {
+			return nil
+		}
+		if runtime.GOOS == "windows" {
+			if os.IsPermission(err) || isWindowsSharingViolation(err) {
+				time.Sleep(10 * time.Millisecond)
+				continue
+			}
+		}
+		break
+	}
+	return err
+}
+
+func isWindowsSharingViolation(err error) bool {
+	var errno syscall.Errno
+	if errors.As(err, &errno) {
+		const ERROR_SHARING_VIOLATION syscall.Errno = 32
+		return errno == ERROR_SHARING_VIOLATION
+	}
+	return false
 }
--- a/internal/sandbox/grants.go
+++ b/internal/sandbox/grants.go
@@ -600,7 +600,7 @@ func (store *GrantStore) writeState(state grantFile) error {
 	if err := os.WriteFile(tempPath, append(data, '\n'), 0o600); err != nil {
 		return err
 	}
-	if err := os.Rename(tempPath, store.filePath); err != nil {
+	if err := renameWithRetry(tempPath, store.filePath); err != nil {
 		_ = os.Remove(tempPath)
 		return err
 	}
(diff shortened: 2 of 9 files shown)


Vasanthdev2004 (approved):
Reviewed #641 on a fresh worktree. Build, vet, and tests for the touched packages (cron, sandbox, sessions, swarm) are all green, and gofmt is clean.

The change looks correct and well-scoped. `renameWithRetry` is bounded (10 attempts), only retries the two named transient Windows errors — ERROR_SHARING_VIOLATION (32) via the explicit `isWindowsSharingViolation` check and ERROR_ACCESS_DENIED (5) via `os.IsPermission`, which is exactly how the rest of the codebase already surfaces that code (see `internal/cron/lock.go:70`) — and preserves the original error by `return err` after the loop rather than masking it. No race is introduced: `src` is the caller-owned temp file, nothing re-reads or re-writes it between attempts, and a failed `os.Rename` does not move the file, so retrying is safe. On non-Windows the `runtime.GOOS == "windows"` guard means it breaks after the first error, so behavior is identical to plain `os.Rename` there — no behavior change on Linux/macOS. It composes cleanly with the lock/reclaim work from #615/#616/#631: this retries only the final data-file rename, never the lock acquisition, so the two concerns stay orthogonal.

A few non-blocking notes:

1. The `renameWithRetry` + `isWindowsSharingViolation` pair is now byte-identical across four packages (cron/store.go, sandbox/windows_unelevated.go, sessions/store.go, swarm/mailbox.go), and a fifth variant already lives in `internal/update/replace_windows.go`. If the retry policy ever changes, four files drift in lockstep. A tiny shared internal helper would be cleaner; happy to leave that for a follow-up.

2. There's no test for the retry path and no comment justifying why. I understand it's gated by `runtime.GOOS == "windows"` and so can't fire on Linux CI, but the error classifier (`os.IsPermission` / `isWindowsSharingViolation`) is pure and could be unit-tested on Windows, or the loop could take an injectable classifier so the retry decision is testable cross-platform. Not a blocker — the existing concurrent mailbox test still passes and now logs send failures, which is a nice diagnostic touch.

3. The 10ms delay is fixed rather than backoff (~90ms worst case), and ERROR_LOCK_VIOLATION (33) isn't retried even though `internal/securefile` treats it as transient. Both are fine for the scoped Access-denied/Sharing-violation fix but worth keeping in mind if the policy is revisited.

Verdict: approve. This is a real Windows-reliability improvement with no correctness or security concern, and it's safe on other platforms.

euxaristia:
Pushed 2af6ab3 to fix the Smoke and Zero Review failures: removed a blank line at EOF in mailbox_test.go flagged by gofmt.

jatmn (changes requested):

## Findings

- [P1] Apply the retry policy to the remaining sandbox state writers
  `internal/sandbox/windows_unelevated.go:144`
  This adds a retry only for the unelevated-plan marker, but the same temp-file replacement pattern remains a one-shot `os.Rename` in the elevated setup marker (`windows_setup.go:242`), capability-SID store (`windows_runner.go:666`), and grants store (`grants.go:603`). A transient Windows sharing/access-denied error therefore still rolls back `zero sandbox setup`, prevents restricted-command initialization, or rejects a grant/revoke operation—the exact persistence failure the PR says it fixes for sandbox. Route these writers through the same retry policy (and cover them) rather than leaving the primary sandbox paths vulnerable.

- [P1] Retry checkpoint-file replacement during rewinds
  `internal/sessions/rewind.go:293`
  `RestoreToSequence` still restores workspace files through `writeFileAtomic`, which performs a bare rename. On Windows, an indexer or delayed handle can make that rename return error 5/32; the caller then only appends the file to `report.Skipped` and continues. Consequently a rewind can appear successful while leaving files at their later contents, even though this PR advertises retry protection for session atomic writes. Use the retrying helper here as well, or surface this partial restore as a failure.

euxaristia:
Pushed e8f5b7f: routed the remaining 4 bare os.Rename call sites through the existing retry helper (windows_setup.go's setup marker write, windows_runner.go's capability-SID store write, grants.go's writeState, and sessions/rewind.go's separate writeFileAtomic method used by RestoreToSequence). All 4 reuse the retry helper already defined in their own package — no new helper, no import changes.

jatmn (changes requested):
Please rebase and resolve conflicts.

@Vasanthdev2004 LGTM otherwise.

euxaristia:
Closing this. The fix already landed on main via #660: that PR added a shared internal/fsutil.RenameWithRetry helper and wired it into the same call sites this PR touches (cron store, sandbox windows_unelevated marker, sessions store, swarm mailbox), plus equivalent tests. Rebasing this branch onto main leaves an empty diff, so there's nothing left for this PR to contribute.

PR #642, opened 2026-07-10
feat(agent): add PermissionModePlan for interactive read-only planning

euxaristia:
This PR adds the core agent support for PermissionModePlan. In this mode, the agent available tools are restricted to read-only capabilities, update_plan, and ask_user. This ensures the agent cannot mutate the workspace, execute commands, or run code changes while planning.

## Summary by CodeRabbit

* **New Features**
  * Added **Plan** permission mode via CLI `--plan` and TUI `/plan [status|on|off]`, restoring the prior mode on exit; also supported in session mode switching.
  * Updated help, completions, and `exec` behavior/tool listings to reflect plan visibility.
* **Bug Fixes**
  * Hardened **plan/spec-draft**: unsafe/mutating and control tools are hidden or denied; `request_permissions` is blocked; plan runs suppress executable hook activity; denial messages now reference the active mode.
  * `tool_search` respects permission mode (including deferred-tool filtering) and prevents schema/name leakage.
* **Tests**
  * Added regression/anti-spoofing coverage for plan/spec-draft denial, no filesystem mutation, truncation notice handling, and correct `tool_search` filtering.


Vasanthdev2004 (dismissed):
Reviewed the plan-mode gating carefully — the whole point is read-only enforcement. The change routes `PermissionModePlan` through the same gate spec-draft already uses: in `executeToolCall`, any tool not advertised under the mode is denied (`DenialFiltered`) before the permission-grant or sandbox path is ever reached. `toolAdvertisedInPlan` admits only `ask_user`, `update_plan`, and `SideEffectRead`+`PermissionAllow` tools, so `write_file`/`edit_file`/`apply_patch` (`SideEffectWrite`) and `bash` (`SideEffectShell`) are blocked at the gate, not merely advisory. `update_plan` is in-memory read-only, so the explicit allowlist there is harmless.

Build, vet, and the agent tests pass; it merges clean. Approve. One minor note: there's no test for the plan-mode gate analogous to the four spec-draft cases in `loop_test.go` — a small regression test asserting the mutating tools are denied would lock in the invariant; happy to leave that to a follow-up.

Vasanthdev2004:
@gnanam1990 when you have a moment, could you review the plan-mode read-only gate here? It's security-relevant (denies write_file/edit_file/apply_patch/bash at the gate before the sandbox path), so a second pair of eyes would be good.

gnanam1990 (dismissed):
## Verdict: Approve

Aligns with @Vasanthdev2004's review. Agent foundation for interactive read-only plan mode is correct.

### What looks good

- `PermissionModePlan` is well-documented vs `spec-draft` (in-session vs separate session).
- `toolAdvertisedInPlan` correctly allows `ask_user`, `update_plan`, and read-only tools; blocks mutators, shell, and network tools.
- Defense-in-depth in `executeToolCall` matches spec-draft pattern (`DenialFiltered` before sandbox).
- Intentional divergence: `update_plan` allowed (opposite of spec-draft) — correct for interactive planning.

### Nits

- Redundant `modeName` logic in denial message (`string(permissionMode)` already yields `"plan"` / `"spec-draft"`).
- No plan-mode tests yet — spec-draft has four regression tests; strongly prefer adding before/with #643.

### Issues

None blocking. Merge before or with #643 for user-visible plan mode.

internal/agent/loop.go
```diff
@@ -898,12 +898,18 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 		}, nil
 	}
 	tool, toolFound := registry.Get(call.Name)
-	if permissionMode == PermissionModeSpecDraft && toolFound && !ToolAdvertised(tool, permissionMode) {
+	if (permissionMode == PermissionModeSpecDraft || permissionMode == PermissionModePlan) && toolFound && !ToolAdvertised(tool, permissionMode) {
+		modeName := string(permissionMode)
```
kevincodex1, line 1100: can you check the CI/Security & code health comments

kevincodex1: @euxaristia please check bro

euxaristia:
Addressed both nits from the approval review.

- Removed the redundant modeName if/else in the denial message: string(permissionMode) already yields "plan" / "spec-draft", so the branch was recomputing the same value it started with.
- Added plan-mode regression tests mirroring three of the four existing spec-draft tests: advertised tool set, denied write_file calls, denied bash calls. The fourth spec-draft test covers submit-and-stop review control, which has no plan-mode analog since plan mode has no submit tool.

go vet and go test -race -count=1 ./internal/agent/... are clean.

kevincodex1 (changes requested):

internal/agent/loop.go
euxaristia: Addressed the deadcode finding for Result.Truncated from the Security and code health job in 0b71ef6. The helper had no production callers, the CLI and TUI both use TruncationNotice() instead, so I removed it and updated the two tests that asserted on it. I reran the same command CI uses (go run golang.org/x/tools/cmd/deadcode@v0.46.0 -test=false ./...) on the branch and the finding is gone. The remaining findings in that job are pre-existing on main and not introduced by this PR, including the other unreachable funcs in internal/agent.

jatmn (changes requested):

## Findings

- [P1] Deny `request_permissions` independently of the registry lookup
  `internal/agent/loop.go:999`
  This gate only rejects a hidden tool when it is present in the supplied registry, but `request_permissions` is dispatched by name at line 1042 even when no registry entry exists. A plan-mode run with a reduced registry can therefore request and receive a turn- or session-scoped filesystem/network grant, then use the grant on later registered tools, defeating the promised read-only boundary. Reject that special tool inside `executeRequestPermissions` for plan/spec-draft modes (and cover the registry-omitted case).

Vasanthdev2004:
Circling back from the read-only-enforcement angle, which was my original focus. The blind spot jatmn flagged this morning is real: request_permissions is dispatched by name before the registry gate, so a plan-mode run with a reduced registry could request a grant and then use it on later tools.

Worth noting for whoever picks this up next: the commit pushed a few hours after jatmn's review ("deny request_permissions in plan/spec-draft") adds exactly that guard inside executeRequestPermissions, with a regression test asserting request_permissions is denied before it reaches OnPermissionRequest, so the current head looks like it closes the escape. jatmn's and kevin's changes-requested both predate that commit, so they will want to re-check against this head. From the read-only-boundary side I care about, this reads right now. Merge stays kevin's call since he is leading this one.

jatmn (changes requested):

## Findings

- [P1] Disable executable hooks while plan mode is active
  `internal/agent/loop.go:186`
  The new mode promises that planning cannot write files or run commands, but a run dispatches `sessionStart`/`sessionEnd` hooks unconditionally, and an allowed read call dispatches `beforeTool`/`afterTool` hooks. Those hooks execute configured host commands directly, outside the advertised-tool and sandbox gates, so a project hook can mutate the workspace or execute a process merely by starting a plan session or calling `read_file`. Suppress or sandbox hooks in plan mode and add a regression test with a marker-writing hook.

euxaristia:
Pushed b767183 for the hooks finding: sessionStart/sessionEnd/beforeTool/afterTool dispatches are all gated off while plan (and spec-draft) mode is active, since hooks execute configured host commands outside the advertised-tool and sandbox gates. Regression test runs a plan-mode session with a filesystem-mutating sessionStart hook wired and asserts no hook command launches (audit shows zero executions, marker path untouched). The registry-omitted request_permissions denial was in the previous push (612c973).

jatmn (changes requested):

## Findings

- [P1] Keep process-spawning LSP navigation out of plan mode
  `internal/agent/loop.go:3029`
  Plan mode treats every `SideEffectRead`/`PermissionAllow` tool as safe, which includes `lsp_navigate`. A plan-mode call reaches `lspNavigateTool.Run`, whose manager lazily starts the configured language server with `exec.Command(...).Start()` outside the sandbox or permission path. Consequently, merely navigating a supported source file executes a host process (and may let that server create caches/indexes), contradicting the mode's promise that planning cannot run commands or mutate the workspace. Exclude this process-spawning tool from the plan allowlist, or put that launch behind an equivalent no-process boundary.

- [P1] Do not whitelist planning tools solely by their registry name
  `internal/agent/loop.go:3024`
  `Registry.Register` replaces an existing entry with the same name, so an extension or embedding caller can register a `PermissionAllow` write/shell tool named `update_plan`. The name-only exception advertises it in plan mode and the normal execution path then calls `registry.RunWithOptions`, allowing the replacement to perform its side effect despite the read-only guarantee. Require the expected non-mutating safety classification for the `ask_user`/`update_plan` exceptions and add a regression test with a spoofed registered tool.

euxaristia:
Pushed 1d5f438.

Fixed:
- toolAdvertisedInPlan whitelisted ask_user and update_plan by name alone, so a same-named mutating tool could bypass the plan-mode safety gate. Now every tool is validated against its actual Safety() (SideEffectRead + PermissionAllow) instead of a name-only shortcut.
- lsp_navigate is marked SideEffectRead but its manager lazily spawns a real language-server process, which breaks plan mode's "runs nothing" guarantee. Excluded it explicitly.
- Extended the hook-suppression test to also register a tool with before/after hooks and confirm none fire during a plan-mode run.
- Added a regression test for a name-only spoofed control tool, and one for lsp_navigate not being advertised. Verified both catch the regression against the pre-fix code.

Nothing left open on this one, all review threads are addressed.

internal/agent/loop.go
euxaristia: Follow-up on the Security & code health thread: deadcode for Result.Truncated was already cleared in earlier commits. The remaining Smoke failure (TestExecSpecWorktreeInheritsTrustEndToEnd) was a real bug from hook suppression applying to spec-draft as well as plan mode. Fixed in c64d6f5: plan mode still suppresses executable hooks; spec-draft keeps the trust-gated hook model so trusted worktrees inherit trust under --use-spec --worktree.

euxaristia:
Addressing open review findings and the Smoke CI failure:

**CI (`TestExecSpecWorktreeInheritsTrustEndToEnd`)**
- Root cause: `hooksSuppressed` treated plan and spec-draft the same, so trusted worktree project hooks never ran under `--use-spec`.
- Fix in c64d6f5: suppress executable hooks only in plan mode. Spec-draft keeps the trust-gated hook model (covered by the e2e test).

**@jatmn findings (on prior commits)**
- Deny `request_permissions` when the registry omits it: 612c973
- Suppress executable hooks in plan mode: b767183 (scope corrected to plan-only in c64d6f5)
- Exclude process-spawning `lsp_navigate` from plan allowlist: 1d5f438
- Do not whitelist `ask_user`/`update_plan` by name alone: 1d5f438 (safety-checked + spoofed-name regression test)

**@kevincodex1** Security & code health: deadcode (`Result.Truncated`) already removed; Smoke failure fixed in c64d6f5 as above.

euxaristia:
Addressed the latest plan-mode P1s from review (and rebased onto current main).

### [P1] Process-spawning lsp_navigate kept out of plan mode
`toolAdvertisedInPlan` excludes `lsp_navigate` even though it is classified `SideEffectRead`: its manager can start a language server with `exec` outside sandbox/permission gates (dfa8114).

Regression coverage:
- Advertisement: `TestPlanModeAdvertisesOnlySafeTools` denies `lsp_navigate`
- Execution path: `TestPlanModeDeniesLSPNavigateToolCalls` (3f9dee6) denies a model call before `Run`

### [P1] No name-only whitelist for planning tools
`toolAdvertisedInPlan` no longer returns true for `ask_user`/`update_plan` by name alone. Every tool must declare `SideEffectRead` + `PermissionAllow`. A re-registered mutating tool under either name is neither advertised nor executed (dfa8114).

Regression coverage:
- `TestPlanModeRejectsNameOnlySpoofedControlTools` is table-driven over both `update_plan` and `ask_user` (3f9dee6)

### Verification
- `go test ./internal/agent/... -count=1`
- `go vet ./internal/agent/...`

Branch tip: `3f9dee6`.

internal/agent/loop.go
euxaristia: Status after rebase onto current main and follow-up hardening in 3f9dee6 (branch tip):

- Deadcode for `Result.Truncated` remains cleared from earlier commits.
- Hook suppression is plan-mode only so spec-draft worktrees keep the trust-gated model (`TestExecSpecWorktreeInheritsTrustEndToEnd`).
- Plan allowlist also blocks process-spawning `lsp_navigate` and no longer trusts control-tool names without `SideEffectRead`+`PermissionAllow` (dfa8114, covered further in 3f9dee6).

`go test ./internal/agent/... -count=1` and `go vet ./internal/agent/...` pass on 3f9dee6.

jatmn (changes requested):
## Findings

- [P2] Keep deferred mutators out of plan-mode tool search
  `internal/agent/loop.go:2904`
  The new plan-mode visibility rule makes the pre-existing `tool_search` loader runnable, but that loader filters deferred candidates only by `EnabledTools`/`DisabledTools`, not by plan-mode visibility. With deferral active, a model can call `tool_search` with `select:<deferred write tool>` and receive that mutator's schema plus a `load_tools` result. The following turn correctly hides the tool and rejects any direct call, so the model has been told that an unavailable mutator was loaded and can waste turns retrying it; it also exposes capability metadata that the mode promises not to advertise. Apply the same permission-mode visibility filter when resolving/loading deferred tools (or disable the loader/deferral in plan mode), and cover this with a deferred write-tool regression test.

euxaristia:
Addressed jatmn's P2 finding on tool_search: deferred candidates were filtered by EnabledTools/DisabledTools only, never by plan/spec-draft visibility.

I checked the exact commit reviewed (3f9dee6) and confirmed tool_search itself is already denied at dispatch in plan/spec-draft mode: its Safety carries no side effect, so it fails the same SideEffect==Read advertisement gate that direct tool calls use (toolAdvertisedInPlan/toolAdvertisedInSpecDraft). I wrote an end-to-end test that force-calls tool_search in plan mode (bypassing advertisement, as if a stale call slipped through from a prior turn) and confirmed it gets denied with "not available in plan mode" before ever reaching tool_search's own code, so this specific path isn't reachable through Run() today.

That said, the underlying gap you flagged is real: tool_search.go's own candidate filtering had no mode awareness at all. If that outer gate is ever loosened (e.g. someone decides tool_search's no-side-effect Safety should count as advertisable in plan mode, which is a plausible-looking change on its own), the loader would hand a deferred write tool's name, description, and full schema straight to the model with no independent check. I confirmed this by temporarily patching the outer gate to advertise tool_search in plan mode and rerunning the test: the deferred mutator's schema did leak.

Fixed by adding a toolAdvertisedForPermissionMode filter to tools/tool_search.go (mirroring agent.ToolAdvertised's plan/spec-draft branches, same import-cycle workaround already used for toolAllowedByFilters) and applying it in visibleDeferredTools and visibleEagerToolNames, alongside the existing operator filters. Added unit tests for both modes in tool_search_test.go and the end-to-end plan-mode regression test in deferred_loop_test.go. Verified all new tests fail without the fix (one shows load_tools resolving straight to the mutator's name) and pass with it restored.

go build, go vet, and internal/agent + internal/tools tests are clean. Pushed as a new commit on this branch.

euxaristia:
Addressed CodeRabbit's HEAD finding on spec-draft control-tool advertisement (`8c95a92`).

**spec-draft Safety checks (not name-only)**
- `toolAdvertisedInSpecDraft` and `toolAdvertisedForPermissionMode` now require the real tools' Safety shapes: `ask_user` = SideEffectRead+Allow, `submit_spec` = SideEffectWrite+Allow. `update_plan` stays denied.
- Spoofed re-registrations (shell side effect, deny permission, etc.) are neither advertised nor loadable via `tool_search`.

**Tests**
- `TestSpecDraftModeRejectsNameOnlySpoofedControlTools`
- `TestToolSearchRejectsSpoofedSpecDraftControlToolsBySafety`

internal/agent/loop.go
euxaristia: Done in 8c95a92.

The CI/Security finding was CodeRabbit's note that spec-draft advertised `ask_user` / `submit_spec` by name alone. Both `toolAdvertisedInSpecDraft` (agent loop) and `toolAdvertisedForPermissionMode` (tool_search) now require the real tools' Safety shapes:

- ask_user: SideEffectRead + PermissionAllow
- submit_spec: SideEffectWrite + PermissionAllow

Spoofed re-registrations are neither advertised nor loadable. Coverage: `TestSpecDraftModeRejectsNameOnlySpoofedControlTools` and `TestToolSearchRejectsSpoofedSpecDraftControlToolsBySafety`.

jatmn (changes requested):

## Findings

- [P1] Wire the plan-mode control into a production entry point
  `internal/agent/types.go:29`
  `PermissionModePlan` is documented as being toggled by TUI `/plan`, but the existing command only renders `m.planText()` and never changes `m.permissionMode`; the normal run path therefore continues forwarding the prior Auto/Ask mode. Neither the TUI mode toggle nor the CLI/ACP selectors can choose `plan`, so interactive users cannot enter the advertised read-only planning mode and write/shell tools remain available under the previous policy. Please add an actual mode entry/exit path (and integration coverage), or adjust the scope and documentation to match a different supported activation path.

euxaristia:
Pushed 584e128 addressing the P1 plan-mode wiring finding.

**Wire plan mode into production entry points**
- TUI `/plan on` / `/plan off` now toggles `permissionMode` to/from `PermissionModePlan`, restoring the prior mode on off (not status-only rendering).
- `zero exec --plan` selects plan mode for that run.
- ACP session mode / config selectors accept and advertise `plan` (read-only narrow; Unsafe stays operator-gated).
- Docs on `PermissionModePlan` updated to list the real entry paths.
- Integration coverage: TUI plan toggle, CLI exec plan path, ACP set_mode wiring into `agent.Options.PermissionMode`.

euxaristia:
Addressed the remaining outstanding findings on top of the prior round's fixes (rebased onto current `main`, tip is `085a639`):

**jatmn P1 (the core gap)** — already wired up before this pass: `internal/tui/plan_command.go` now has a real `/plan on|off` entry/exit path that flips `m.permissionMode` (restoring the prior mode on `off`), `zero exec --plan` selects `PermissionModePlan` for a run, and ACP advertises `plan` as a client-selectable session mode. All backed by integration coverage across `internal/tui`, `internal/cli`, and `internal/acp`.

**coderabbitai (this pass)** — all three from the latest review round were still live against `584e128` (the branch tip when I started) and are now fixed:
1. `internal/cli/exec_parse.go` — `--plan --worktree` is now rejected during option validation, before worktree prep (a filesystem mutation) can run ahead of the mode gate. Regression: `TestParseExecArgsRejectsPlanWithWorktree`.
2. `internal/tools/tool_search_test.go` — the spoofed-control-tool test's schema now carries a distinctive `spoofed_secret` property, and the test asserts it never leaks into `result.Output`, not just the description string.
3. `internal/tui/plan_command.go` / `model.go` — added a shared plan-mode guard at the top of `dispatchCommand` (mirrors the existing BTW guard) that rejects `/rewind`, `/export`, and `/sandbox-setup` while `permissionMode == agent.PermissionModePlan`, since those run entirely inside the TUI process and bypass the agent tool-advertisement gate. Regression coverage drives a real checkpoint/file-write/process-spawn for each and confirms none occurs in plan mode, plus a control test confirming the guard doesn't fire outside plan mode.

**Reviewer nit (Vasanthdev2004 / gnanam1990)** — the plan-mode regression tests analogous to spec-draft's four `loop_test.go` cases were already present from the prior round (`TestPlanModeAdvertisesOnlySafeTools`, `TestPlanModeDeniesHiddenToolCalls`, `TestPlanModeDeniesBashToolCalls`, `TestPlanModeRejectsNameOnlySpoofedControlTools`, plus `TestPlanModeDeniesLSPNavigateToolCalls`) — all still passing.

**kevincodex1** — re-checked `reviewThreads` via the GraphQL API for anything missed. His only inline comments are on one thread (`internal/agent/loop.go:1100`, "check the CI/Security & code health comments"), which is already marked resolved — it tracked through the deadcode cleanup, the plan-vs-spec-draft hook-suppression scoping fix, and the safety-metadata hardening for spoofed control tools, ending with a confirmation reply. No separate top-level review body, no other unresolved inline threads from him, and nothing that reads as a policy/trust-anchor call requiring a fresh decision from him — I'm not flagging anything new on his behalf.

Verified: `gofmt -l .` clean, `go vet ./...` clean, `go test ./internal/agent/... ./internal/cli/... ./internal/tools/... ./internal/acp/...` all pass, and the full `internal/tui` plan-mode/rewind/export/sandbox-setup test set passes (one pre-existing, unrelated `TestAltScreenTranscriptScrollKeepsFooterFixed` failure remains — confirmed via `git stash` that it also fails without any of this PR's changes, caused by this checkout's unusually long worktree path overflowing a fixed-width title-bar layout assertion, not a regression here).

jatmn (changes requested):

## Findings

- [P1] Do not start MCP servers for a plan-mode run
  `internal/cli/exec.go:319`
  `--plan` selects the read-only mode before this unconditional `registerMCPToolsForWorkspace` call, but that helper documents that registration starts configured stdio servers before tool visibility is evaluated. Thus `zero exec --plan ...`—including `--list-tools`—executes a trusted workspace's configured MCP command even when the model makes no tool call, contradicting the mode's no-command/read-only contract. Skip/defer MCP activation in plan mode and cover the no-spawn path.

- [P1] Close the remaining TUI commands that escape the plan-mode gate
  `internal/tui/plan_command.go:54`
  The new guard only lists rewind, export, and sandbox setup. In particular, `/plan on` followed by `/spec <task>` reaches `handleSpecCommand`, swaps the run to `PermissionModeSpecDraft`, and permits `submit_spec` to write `.zero/specs/*.md`. The same incomplete gate also leaves `/stop` able to terminate background shell processes and `/doctor --connectivity` able to issue a provider network probe. Define the plan-mode command policy for every local side-effecting command and reject these paths (with regressions), rather than relying on this partial list.

- [P1] Do not report plan mode active while the current turn keeps its old permissions
  `internal/tui/plan_command.go:23`
  `/plan on` is accepted while a normal run is pending, but the running goroutine has already copied `m.permissionMode` into its `agent.Options`; it can therefore continue with the prior Auto/Ask/Unsafe write or shell policy after the UI reports “Active: read-only planning.” ACP likewise snapshots `sess.currentMode()` when it starts a prompt. Refuse/defer the transition until the turn finishes, or cancel/restart the turn before acknowledging plan mode, across both entry points.

- [P2] Keep plan-mode tracing from creating an output file
  `internal/cli/exec.go:562`
  `zero exec --plan --trace trace.ndjson ...` (or `ZERO_TRACE=trace.ndjson`) is accepted and defers `writeTraceSnapshot`, which calls `os.Create` and therefore creates or truncates the supplied file after the run. That is the same pre/post-run filesystem side effect for which the parser already rejects `--plan --worktree`; reject file-backed tracing in plan mode (allowing only stderr if desired) and cover the environment-variable path too.

jatmn (changes requested):

## Findings

- [P1] Do not pass an unimplemented permission-mode flag to specialist children
  `internal/specialist/exec.go:270`
  Both fresh and resumed specialist invocations now append `--permission-mode <parent mode>`, but `parseExecArgs` has no case for that flag and rejects it as unknown. Normal Task and swarm calls propagate a non-empty mode, so every child exits during argument parsing before it can run. Either add validated CLI plumbing for this internal mode or use the supported mode-selection interface when constructing child arguments.

- [P1] Keep MCP startup out of a plan run
  `internal/cli/exec.go:319`
  `zero exec --plan` still calls `registerMCPToolsForWorkspace` before tool visibility is evaluated, including for `--list-tools`. Registering a configured stdio MCP server starts its subprocess, so a plan invocation can execute arbitrary configured host code before the agent's read-only tool gate is reached. Skip MCP process registration in plan mode (or use a non-spawning discovery path) so the CLI entry point upholds the advertised no-command behavior.

- [P1] Do not report plan mode active while an older-permission turn is still running
  `internal/tui/plan_command.go:24`
  `/plan on` is accepted while a turn is pending and immediately reports that the session is read-only, but the active TUI command has already captured its previous `PermissionMode`; ACP has the same race because `runTurn` snapshots `sess.currentMode()` before `set_mode(plan)` updates the session. A turn launched under Ask or Unsafe can therefore write or run shell after the client has been told plan mode is active. Reject/defer the transition until the current turn finishes, or cancel and wait for it before confirming the new mode.

- [P2] Do not let `/spec` bypass an active plan-mode session
  `internal/tui/plan_command.go:54`
  The new guard blocks several local side-effecting commands, but `/spec` remains available and starts a `PermissionModeSpecDraft` run whose `submit_spec` tool writes `.zero/specs`. This contradicts the active plan-mode session's read-only promise without requiring `/plan off`; block `/spec` in plan mode or make the draft flow inherit the read-only policy. The same command-policy audit should cover the MCP management paths, which can save configuration or start servers.

jatmn (changes requested):

## Findings

- [P1] Restore the failing required checks before merge
  GitHub currently reports all three Smoke jobs and the Zero Review job as failed for this head, leaving the PR `BLOCKED`. The job output was not available through the API in this review, so the concrete failure still needs to be surfaced and fixed (or the checks rerun successfully); this cannot merge while those required checks remain red.

- [P1] Apply the propagated permission mode when launching a specialist
  `internal/cli/exec_tools.go:75`
  The new specialist paths append `--permission-mode plan`/`spec-draft`, but `resolveExecPermissionMode` derives the effective mode exclusively from `--auto` and `--skip-permissions-unsafe`. A plan child therefore starts as ordinary `auto`, so plan tool filtering, hook suppression, MCP suppression, and the plan prompt do not apply. Parse and validate the explicit mode in the resolver (or pass the corresponding mode flag), with an end-to-end assertion on the child `agent.Options.PermissionMode`.

- [P1] Do not announce plan mode while the ACP turn is still using its old mode
  `internal/acp/agent.go:354`
  `session/set_mode(plan)` can update and notify the session while a `session/prompt` is running. `runTurn` has already copied `sess.currentMode()` into `agent.Options`, so that in-flight Auto/Ask turn may still run write or shell tools after the client has received `current_mode: plan`. Serialize mode changes with the active turn, or cancel/restart the turn before publishing the mode update.

- [P1] Keep MCP management out of the TUI's plan-mode boundary
  `internal/tui/plan_command.go:60`
  The new guard only rejects four command kinds. `/mcp add|enable|disable|remove` still reaches the MCP command runner, which persists configuration and can start or contact servers. That is the same out-of-agent process/configuration path the new guard is meant to cover, despite the UI claiming plan mode is read-only. Block mutating MCP subcommands (or the entire MCP manager) while plan mode is active, with regression coverage.

PR #647, opened 2026-07-10
fix(tui): cache settled alt-screen transcript

### LINKED ISSUE #561: perf(tui): transcript re-renders full history every frame — flush frontier never advances in alt-screen
### Problem

The TUI feels progressively laggier as the conversation grows. It is not a CPU spike — it is periodic jank and GC pressure that scales with transcript length.

Root cause: the flush frontier (`m.flushed`) never advances in the real app, so every `View()` rebuilds body items over the **entire** transcript instead of just the unflushed tail.

- `useAltScreen()` unconditionally returns `true` for the shipped app (`internal/tui/run.go:70-72`).
- `settleTranscript` — the only code that advances `m.flushed` — bails out immediately in alt-screen mode: `if m.altScreen { return m, nil }` (`internal/tui/flush.go:76-78`).
- So `m.flushed` is only ever reset to `0` (`flush.go:141`) and never incremented in production. `startIdx := m.flushed` (`internal/tui/transcript_selection.go:246`) is therefore always `0`, and the loop at `transcript_selection.go:256` walks the whole transcript on every call.
- `transcriptBodyItems()` is rebuilt on every `View()` with no cross-frame caching (`View` at `internal/tui/model.go:2345` → `transcriptView()` → `transcriptBodyItems(...)` at `model.go:2464`). `View()` fires at ~10Hz from cursor-blink (`model.go:866`) and spinner (`model.go:1792`) ticks.
- Secondary: `buildRowContext(m.transcript)` (`internal/tui/rendering.go:68`, called from `transcript_selection.go:242`) does a full O(n) scan of the whole transcript per call regardless of the frontier.

Note: the row-level lipgloss rendering is already correctly windowed — `layoutVisibleTranscriptBodyItems` (`transcript_selection.go:594-619`) only calls `item.render()` for visible items, and there is an LRU render cache (`render_cache.go`). The defect is in the per-frame item-list construction and context scan that feed those caches, not in the windowing itself.

### Requested behavior

Advance the flush frontier in alt-screen mode too, so `transcriptBodyItems` and `buildRowContext` only process the unflushed tail. In alt-screen mode the `tea.Println` scrollback emission is intentionally skipped (Bubble Tea drops it), but the frontier bookkeeping should still advance. As a follow-up, make `buildRowContext` incremental (or cache the settled-prefix context) so it stops rescanning the full transcript every frame.

### Why it belongs in Zero

The frontier optimization already exists in the codebase (`settleTranscript` / `m.flushed`) but is unreachable on the alt-screen path that every user runs, so long sessions pay full-history O(n) cost on every redraw. This mirrors the concern in #494 ("TUI render rebuilds model registry every frame"), where render-path work was similarly cached to avoid per-redraw allocation.

### Measurements

Benchmark of a single `transcriptBodyItems()` call (Apple M1), production reality (`m.flushed=0`) vs. frontier-at-tail (`m.flushed=len`):

| turns | Full (`flushed=0`) | Frontier (`flushed=len`) | speedup |
|-------|--------------------|--------------------------|---------|
| 50    | 360µs / 302KB / 2475 allocs   | 12.8µs / 47KB / 66 allocs   | ~28x |
| 500   | 3.43ms / 2.8MB / 24542 allocs | 55µs / 106KB / 524 allocs   | ~62x |
| 5000  | 36.1ms / 35.4MB / 253089 allocs | 438µs / 560KB / 5055 allocs | ~82x |

This is per single `View()`; at ~10Hz that is ~360ms/s of rebuild and ~350MB/s of garbage at 5000 turns. Once the frontier is fixed, the remaining ~438µs is almost entirely `buildRowContext`'s full-transcript scan (~431µs standalone at 5000 turns), which is why the incremental-context follow-up matters.

Reproduce:

```
go test ./internal/tui/ -run '^$' \
  -bench 'BenchmarkTranscriptBodyItems|BenchmarkBuildRowContext' \
  -benchmem -benchtime=200x
```

<details>
<summary>Benchmark source (drop into <code>internal/tui/transcript_perf_bench_test.go</code>)</summary>

```go
package tui

import (
	"fmt"
	"strings"
	"testing"

	"github.com/Gitlawb/zero/internal/tools"
)

// benchTranscript builds a model whose transcript holds n user/assistant/tool
// turns, mimicking a long conversation.
func benchTranscript(n int) model {
	m := transcriptViewTestModel()
	m.width = 100
	rows := make([]transcriptRow, 0, n*3)
	for i := 0; i < n; i++ {
		rows = append(rows,
			transcriptRow{kind: rowUser, text: fmt.Sprintf("user question number %d with a fair bit of text to wrap", i)},
			transcriptRow{kind: rowToolResult, id: fmt.Sprintf("tool-%d", i), tool: "custom_tool", status: tools.StatusOK, detail: strings.Repeat("output-line ", 40)},
			transcriptRow{kind: rowAssistant, text: fmt.Sprintf("assistant answer number %d, also long enough to wrap across lines", i), final: true, turnTools: 1},
		)
	}
	m.transcript = rows
	return m
}

// Production reality: m.flushed stuck at 0 (alt-screen never advances the
// frontier), so every call walks the entire transcript.
func BenchmarkTranscriptBodyItemsFull(b *testing.B) {
	for _, turns := range []int{50, 500, 5000} {
		m := benchTranscript(turns)
		m.flushed = 0
		b.Run(fmt.Sprintf("turns=%d", turns), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				_ = m.transcriptBodyItems(m.width, "", false)
			}
		})
	}
}

// Same call with the frontier advanced to the tail (what settleTranscript
// would achieve if it ran in alt-screen mode).
func BenchmarkTranscriptBodyItemsFrontier(b *testing.B) {
	for _, turns := range []int{50, 500, 5000} {
		m := benchTranscript(turns)
		m.flushed = len(m.transcript)
		b.Run(fmt.Sprintf("turns=%d", turns), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				_ = m.transcriptBodyItems(m.width, "", false)
			}
		})
	}
}

// Isolates the full O(n) transcript scan that runs once per
// transcriptBodyItems call regardless of the frontier.
func BenchmarkBuildRowContext(b *testing.B) {
	for _, turns := range []int{50, 500, 5000} {
		m := benchTranscript(turns)
		b.Run(fmt.Sprintf("turns=%d", turns), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				_ = buildRowContext(m.transcript)
			}
		})
	}
}
```

Raw output (Apple M1):

```
BenchmarkTranscriptBodyItemsFull/turns=50-8              200      360310 ns/op    302621 B/op      2475 allocs/op
BenchmarkTranscriptBodyItemsFull/turns=500-8            200     3429911 ns/op   2805980 B/op     24542 allocs/op
BenchmarkTranscriptBodyItemsFull/turns=5000-8           200    36080212 ns/op  35446517 B/op    253089 allocs/op
BenchmarkTranscriptBodyItemsFrontier/turns=50-8         200       12841 ns/op     47904 B/op        66 allocs/op
BenchmarkTranscriptBodyItemsFrontier/turns=500-8        200       55489 ns/op    106048 B/op       524 allocs/op
BenchmarkTranscriptBodyItemsFrontier/turns=5000-8       200      438636 ns/op    560304 B/op      5055 allocs/op
BenchmarkBuildRowContext/turns=50-8                     200        3862 ns/op      4376 B/op        63 allocs/op
BenchmarkBuildRowContext/turns=500-8                    200       47386 ns/op     62520 B/op       521 allocs/op
BenchmarkBuildRowContext/turns=5000-8                   200      431570 ns/op    516776 B/op      5052 allocs/op
```

</details>

- Cache the built `[]transcriptBodyItem` slice across frames and invalidate on transcript mutation/resize — larger change, overlaps with the frontier fix.
- Store a per-row content hash on `transcriptRow` at append/mutation time so the render/height cache keys (`render_cache.go:126-177`) stop being rebuilt from full row content every frame.

### Related issues

- #494 — TUI render rebuilds model registry every frame (same render-path perf theme, already fixed)
- #525 — Batch session event persistence to reduce fsync overhead (different layer, but same "long-session cost" family)

PierrunoYT:
## Summary

- advance the settled transcript frontier in alt-screen mode without emitting `tea.Println`
- cache settled transcript body items so fullscreen history remains scrollable without rebuilding it every frame
- build row context from only the live tail and preserve dynamic hover, selection, resize, and expand/collapse behavior
- add regression coverage and a 5,000-turn benchmark for the settled alt-screen path

## Root cause

Alt-screen mode returned from `settleTranscript` before advancing `m.flushed`. Since the shipped TUI always uses alt-screen mode, every redraw rebuilt transcript body descriptors and cross-row context for the complete conversation.

## Impact

Settled alt-screen frames now reuse cached body descriptors while the live tail remains dynamic. On the issue benchmark, a settled 5,000-turn transcript takes about 110 ns/op with 0 B/op and 0 allocations/op on Windows/amd64.

## Validation

- focused frontier, transcript selection, hover, and rendering tests
- `go test ./... -run '^$'`
- `go test ./internal/tui -run '^$' -bench '^BenchmarkIssue561SettledAltScreen$' -benchmem -benchtime=100x`

Closes #561

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alternate-screen transcript behavior: settled content remains visible while the flush frontier advances correctly, and selection/hover rendering stays consistent.
  * Fixed stale in-place status updates (doctor/sandbox/compact spinners) by invalidating cached settled layouts when those rows change before the flush boundary.
  * Updated transcript row and file-selection changes to properly invalidate and rebuild alternate-screen settled caches (including clearing selection).
* **Performance**
  * Reduced unnecessary rebuilding by caching alternate-screen settled transcript items and reusing them when frontier/width (or affected content) hasn’t changed.

### DIFF
--- a/internal/tui/command_center.go
+++ b/internal/tui/command_center.go
@@ -214,6 +214,12 @@ func (m model) setDoctorStatusRow(text string) model {
 	for i := len(m.transcript) - 1; i >= 0; i-- {
 		if m.transcript[i].id == doctorStatusRowID {
 			m.transcript[i] = row
+			// Settled alt-screen rows are cached across frames; force a rebuild
+			// so an in-place update (e.g. the spinner tick) is reflected
+			// immediately instead of serving the stale cached snapshot.
+			if i < m.flushed {
+				m.altScreenSettledWidth = 0
+			}
 			return m
 		}
 	}
--- a/internal/tui/files_panel.go
+++ b/internal/tui/files_panel.go
@@ -348,12 +348,25 @@ func (m model) lastRowIndexForFile(path string) int {
 	return -1
 }
 
+func (m *model) setSelectedFile(path string) {
+	if m.selectedFile == path {
+		return
+	}
+	oldRow := m.lastRowIndexForFile(m.selectedFile)
+	newRow := m.lastRowIndexForFile(path)
+	m.selectedFile = path
+	if (oldRow >= 0 && oldRow < m.flushed) || (newRow >= 0 && newRow < m.flushed) {
+		m.altScreenSettledWidth = 0
+	}
+}
+
 // selectFile marks path as the selected file and scrolls the transcript so its
 // most recent edit card is in view; the card tint comes from the renderers
 // reading selectedFile (rowTouchesSelectedFile).
 func (m model) selectFile(path string) model {
-	m.selectedFile = path
-	if offset, ok := m.scrollOffsetForTranscriptRow(m.lastRowIndexForFile(path)); ok {
+	rowIndex := m.lastRowIndexForFile(path)
+	m.setSelectedFile(path)
+	if offset, ok := m.scrollOffsetForTranscriptRow(rowIndex); ok {
 		m.chatScrollOffset = offset
 		if offset == 0 {
 			m.chatBodyLines = 0
--- a/internal/tui/files_panel.go
+++ b/internal/tui/files_panel.go
@@ -352,11 +352,15 @@ func (m *model) setSelectedFile(path string) {
 	if m.selectedFile == path {
 		return
 	}
-	oldRow := m.lastRowIndexForFile(m.selectedFile)
-	newRow := m.lastRowIndexForFile(path)
+	oldPath := m.selectedFile
 	m.selectedFile = path
-	if (oldRow >= 0 && oldRow < m.flushed) || (newRow >= 0 && newRow < m.flushed) {
-		m.altScreenSettledWidth = 0
+	for i := 0; i < m.flushed && i < len(m.transcript); i++ {
+		for _, changedPath := range m.transcript[i].changedFiles {
+			if changedPath == oldPath || changedPath == path {
+				m.altScreenSettledWidth = 0
+				return
+			}
+		}
 	}
 }
 
--- a/internal/tui/flush.go
+++ b/internal/tui/flush.go
@@ -69,13 +69,6 @@ func (m model) settledRow(row transcriptRow, rc rowContext) bool {
 // of the unflushed tail is rendered once and queued for scrollback. It returns
 // the (possibly nil) print command for the queued batch.
 func (m model) settleTranscript() (model, tea.Cmd) {
-	// In alt-screen mode there is no native scrollback surface for tea.Println:
-	// Bubble Tea drops that output by design. Keep rows in the managed view so
-	// the chat behaves like a fullscreen app and cannot reveal prior shell
-	// history by scrolling.
-	if m.altScreen {
-		return m, nil
-	}
 	// Never freeze history at the pre-WindowSizeMsg default width: the first
 	// real WindowSizeMsg arrives before any user-visible content settles, and
 	// flushing earlier would hard-wrap startup rows at 96 cols forever.
@@ -87,10 +80,14 @@ func (m model) settleTranscript() (model, tea.Cmd) {
 		// frontier themselves; this is a safety net).
 		m.flushed = len(m.transcript)
 	}
-	rc := buildRowContext(m.transcript)
+	oldFlushed := m.flushed
+	// An unsettled row blocks the frontier, so every cross-row dependency needed
+	// to decide or render the tail (call/result and prompt/decision pairs) also
+	// lives in the tail. Avoid rescanning the settled prefix on every tick.
+	rc := buildRowContext(m.transcript[m.flushed:])
 	width := chatWidth(m.width)
 	batch := []string{}
-	previousKind, havePreviousKind := previousVisibleTranscriptKind(m.transcript, m.flushed, rc)
+	previousKind, havePreviousKind := m.flushedPreviousKind, m.flushedHavePreviousKind
 	for m.flushed < len(m.transcript) {
 		row := m.transcript[m.flushed]
 		if !m.settledRow(row, rc) {
@@ -98,6 +95,15 @@ func (m model) settleTranscript() (model, tea.Cmd) {
 		}
 		m.flushed++
 		if row.kind == rowWelcome || rc.skip(row) {
+			continue
+		}
+		// Bubble Tea intentionally discards tea.Println in alt-screen mode. The
+		// bookkeeping still advances; the settled prefix is retained in the body
+		// item cache rebuilt below.
+		if m.altScreen {
+			m.flushedAny = true
+			previousKind = row.kind
+			havePreviousKind = true
 			continue
 		}
 		rendered := m.renderRowMode(row, width, rc, true)
@@ -114,6 +120,15 @@ func (m model) settleTranscript() (model, tea.Cmd) {
 		batch = append(batch, rendered)
 		previousKind = row.kind
 		havePreviousKind = true
+	}
+	m.flushedPreviousKind = previousKind
+	m.flushedHavePreviousKind = havePreviousKind
+	if m.altScreen {
+		bodyWidth := m.chatColumnWidth()
+		if m.flushed != oldFlushed || m.altScreenSettledWidth != bodyWidth || m.altScreenSettledFrontier != m.flushed {
+			m.rebuildAltScreenSettledItems(bodyWidth)
+		}
+		return m, nil
 	}
 	if len(batch) > 0 {
 		m.flushQueue = append(m.flushQueue, strings.Join(batch, "\n"))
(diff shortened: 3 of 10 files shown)


Vasanthdev2004 (changes requested):
I built and vetted this and read all six changed files, and the caching design is sound  invalidation covers new turns, resize, and expand/collapse of settled rows, and the steady-state fast path correctly bypasses the cache whenever a row is still live. The benchmark matches the stated goal.

But I found a regression that I think has to be fixed before this lands: in alt-screen mode the doctor connectivity spinner freezes.

`setDoctorStatusRow` (command_center.go:212-222) updates an existing status row in place with `m.transcript[i] = row` and returns. A `rowSystem` row is immediately settleable (flush.go:61-64), so once the frontier advances past the doctor row it gets snapshotted into `altScreenSettledItems`. On the next spinner tick (model.go:1986) `doctorFrame` changes and `setDoctorStatusRow` writes new text into `m.transcript[i]`  but it never zeroes `altScreenSettledWidth`, and the rebuild guard in `settleTranscript` (flush.go:128) only fires on a frontier/width change. With none of those changing on an in-place text update, the rebuild is skipped and the View fast path (transcript_selection.go:249) serves the stale snapshot. I reproduced it: advancing `doctorFrame` 0 → 2 still renders the frame-0 glyph. The same repro passes on origin/main, where alt-screen re-rendered the tail every frame.

Fix is small: have `setDoctorStatusRow` invalidate the settled cache when it updates an existing row (zero `altScreenSettledWidth` on the in-place path, same as `toggleTranscriptRow` does). Worth checking other in-place mutators of already-settled rows for the same gap while you're there.

Everything else I'd merge as-is. Requesting changes just for this.

Vasanthdev2004:
@gnanam1990 could you take a look here? I requested changes for a regression: in alt-screen mode the doctor connectivity spinner freezes, because setDoctorStatusRow updates a settled row in place without invalidating the settled-prefix cache, so the View fast path serves a stale snapshot. Would be good to have you confirm (or push back) before it's reworked.

Vasanthdev2004 (changes requested):
The new commit closes the doctor/sandbox/compact spinner gap I flagged earlier, and all four in-place row mutators are covered now. But the invalidation surface is still incomplete — two more paths change settled-row rendering without advancing the frontier, so the cached prefix goes stale. Both are regressions: before this PR, alt-screen re-rendered the whole transcript every frame, so these were always fresh.

First, the FILES selection tint. `selectFile` (files_panel.go:354) sets `m.selectedFile` and scrolls to the edit card, but doesn't zero `altScreenSettledWidth`; the accent border is read from `selectedFile` (rendering.go:1412 via `rowTouchesSelectedFile`) and baked into the cached descriptor at rebuild time, so the first click on a FILES row (transcript_selection.go:1394) scrolls to the card but its border doesn't highlight until a resize or new turn. The Esc-clear at model.go:1373 has the same gap in reverse (a cached accent border won't clear). Fix: invalidate the cache in those paths, same `if i < m.flushed { m.altScreenSettledWidth = 0 }` style as the status-row setters.

Second, `collapseRepeatedStatusCard` (model.go:2440) drops two rows from the middle of the settled prefix without adjusting `m.flushed` or invalidating. After a repeated identical swarm_status, the dropped pair was already settled (frontier had stopped at the live new-call), so `m.flushed` still equals the new transcript length, the fast path fires, and the cache serves the old call+result cards while hiding the new ones — the collapse is defeated. Fix: when the collapse removes rows, decrement `m.flushed` by the count removed and zero `altScreenSettledWidth`.

The caching design and the 0-alloc benchmark are solid; it's just the invalidation surface that needs these two paths closed. Requesting changes for those.

PierrunoYT:
@Vasanthdev2004 The two remaining cache invalidation issues are fixed in f0f80b6, with regression coverage for FILES selection clearing and repeated status-card collapse. Could you please re-review when you have a chance?

Vasanthdev2004 (dismissed):
Approving — both invalidation gaps are closed. The setSelectedFile helper covers select, the Esc-clear, and the mouse path in one place, and rewinding m.flushed by the removed row count in the collapse path is exactly the fix I had in mind. The two regression tests plus the cache-miss guard in the benchmark are a nice touch.

One non-blocking note for a possible follow-up: setSelectedFile keys the invalidation on lastRowIndexForFile, but rowTouchesSelectedFile tints every settled result row that touched the file — so if a file's most recent edit card is still in the live tail while an older card for the same file is already settled, the older card's tint can go stale until the next rebuild. It's narrow and cosmetic (self-heals on the next frontier advance or resize), so I'm not holding the PR on it.

Thanks for the quick turnarounds on both rounds.

kevincodex1 (approved):
lgtm

PR #646, opened 2026-07-10
feat(ci): add Linux installer build scripts

JohnVictorCrown:
## Summary

Adds build scripts for two Linux installer formats (AppImage and .run) and the cross-platform installer plan.

### Files
- **scripts/build-linux-appimage.sh** — portable AppImage with AppRun, .desktop, AppStream metainfo, icons
- **scripts/build-linux-install.sh** — self-extracting .run archive (sudo-installs to /usr/local/)
- **docs/BUILD_INSTALLERS_PLAN.md** — full plan: Windows NSIS, Linux AppImage/.run, macOS .app
- **docs/assets/logo-great.png** — source icon for all installers

## Summary by CodeRabbit

* **New Features**
  * Added a release plan for platform-specific installers and downloadable artifacts.
  * Added Linux AppImage builds for portable desktop use.
  * Added Linux self-extracting `.run` installers with desktop integration.
  * Added SHA-256 checksum files for installer artifacts.
* **Documentation**
  * Documented planned Windows, Linux, and macOS packaging workflows, release assets, and architecture support.

PR #645, opened 2026-07-10
fix(config): surface unknown/typo'd config fields instead of silently dropping them

Ashwinhegde19:
## Summary

Fixes #644: unknown/typo'd keys in `config.json` were silently discarded by `json.Unmarshal`, and `zero doctor` reported "config files parsed and validated successfully" even when fields were dropped.

Examples that previously went unnoticed:
- `maxTurn` (should be `maxTurns`)
- `sandbox.network` (should be `sandbox.network`, an actual hardening knob)
- `sandbox.blockUnixSocket` inside a provider entry

## What changed

- Added a reflection-based unknown-field scanner (`internal/config/unknownfields.go`) that walks the `FileConfig` JSON schema and recurses into nested objects, arrays, and maps, reporting each unknown key with its full JSON path and a near-match suggestion (`did you mean "maxTurns"?`).
- The two legacy top-level aliases (`mcpServers`, `mcp_servers`) are allowlisted so they are not flagged.
- Detection only inspects json tags, so a config written by a newer Zero with new fields still loads and merges normally — `validate`/`doctor` just surface the unknown keys as issues rather than rejecting the file (preserves forward compatibility).
- Wired into `ValidateBytes`/`ValidateFile`, which `zero doctor` already calls, so `doctor` now reports the unknown fields instead of "validation passed".

## Test plan

- `TestValidateFileFlagsUnknownTopLevelKey` — `maxTurn` → suggests `maxTurns`
- `TestValidateFileFlagsUnknownNestedKey` — `sandbox.blockUnixSocket` → suggests `sandbox.blockUnixSockets`
- `TestValidateFileFlagsUnknownInProvider` — unknown key inside a provider entry
- `TestValidateFileAllowsLegacyAliases` — `mcpServers`/`mcp_servers` not flagged
- `TestValidateFileValidConfigHasNoIssues` still passes (valid configs produce no issues)
- Full `internal/config` and `internal/doctor` suites pass; `go vet` clean.

## Note
Per CONTRIBUTING.md, PRs require the `issue-approved` label. Please apply it so the merge CI passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration validation now flags unknown JSON keys recursively across nested objects, arrays, and provider entries.
  * Validation issues report the full dotted/bracketed field path and may include “did you mean …?” suggestions for likely misspellings.
  * Known keys are matched case-insensitively, and legacy alias keys (including `mcpServers` / `mcp_servers` and snake_case provider variants) are accepted without warnings.
[...]

PR #648, opened 2026-07-10
feat(ci): add Windows NSIS installer, macOS .app bundle, and CI workflow

JohnVictorCrown:
## Summary

Adds installer build scripts for Windows and macOS, plus the CI workflow changes to produce all platform installers during releases.

### Files added
- **scripts/build-windows-installer.ps1** — NSIS-based .exe installer with PATH modification, Start Menu shortcut, desktop icon, Add/Remove Programs registration, and uninstaller
- **scripts/build-macos-app.sh** — macOS .app bundle with Terminal launcher (temp-script approach, no fragile osascript), .icns icon via sips+iconutil, ad-hoc code signature, bundled helpers

### Files modified
- **.github/workflows/release-artifacts.yml**
  - Added \windows-11-arm\ to runner matrix
  - Added \Resolve platform arch name and version\ step (sets VERSION and ARCH from runner OS)
  - Added conditional installer build steps for Windows (NSIS), Linux (AppImage + .run), and macOS (.app)
  - Updated npm asset verification to check all new installer artifacts

### Artifacts produced per release
| Platform | New assets |
|----------|-----------|
| Windows x64 | \zero-v*-windows-x64-installer.exe\ |
| Windows arm64 | \zero-v*-windows-arm64-installer.exe\ |
| Linux x86_64 | \zero-v*-linux-x86_64.AppImage\ + \-install.run\ |
| Linux aarch64 | \zero-v*-linux-aarch64.AppImage\ + \-install.run\ |
| macOS x64 | \zero-v*-macos-x64.zip\ (Zero.app) |
| macOS arm64 | \zero-v*-macos-arm64.zip\ (Zero.app) |

All assets get \.sha256\ checksum files and are automatically picked up by the existing \gh release upload dist/release/*\ glob.

### Scripts from previous PR
The Linux build scripts (build-linux-appimage.sh, build-linux-install.sh) were added in PR #645 ([feat/linux-installers](https://github.com/Gitlawb/zero/pull/646)).

## Summary by CodeRabbit

* **New Features**
  * Added release packaging for Windows, Linux, and macOS.
  * Added Windows installers with Start Menu and Desktop shortcuts.
  * Added macOS application bundles and architecture-specific release artifacts.
  * Added SHA-256 checksums for generated installers and archives.
  * Added support for Windows ARM release builds.

* **Bug Fixes**
  * Expanded release verification to cover additional platform and architecture-specific artifacts.

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -29,6 +29,7 @@ jobs:
           - macos-latest       # macos-arm64 (Apple Silicon)
           - macos-15-intel     # macos-x64 (Intel)
           - windows-latest     # windows-x64
+          - windows-11-arm     # windows-arm64
 
     steps:
       - name: Checkout
@@ -46,6 +47,19 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
+
+      - name: Resolve platform arch name and version
+        shell: bash
+        run: |
+          echo "VERSION=$(jq -r .version package.json)" >> "$GITHUB_ENV"
+          case "${{ runner.os }}-${{ runner.arch }}" in
+            Linux-X64)      echo "ARCH=x64"   >> "$GITHUB_ENV" ;;
+            Linux-ARM64)    echo "ARCH=arm64" >> "$GITHUB_ENV" ;;
+            macOS-X64)      echo "ARCH=x64"   >> "$GITHUB_ENV" ;;
+            macOS-ARM64)    echo "ARCH=arm64" >> "$GITHUB_ENV" ;;
+            Windows-X64)    echo "ARCH=x64"   >> "$GITHUB_ENV" ;;
+            Windows-ARM64)  echo "ARCH=arm64" >> "$GITHUB_ENV" ;;
+          esac
 
       - name: Test
         run: go test ./...
@@ -72,6 +86,68 @@ jobs:
 
       - name: Verify release checksums
         run: go run ./cmd/zero-release verify
+
+      # Windows installer
+      - name: Install NSIS (portable)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $url = "https://sourceforge.net/projects/nsis/files/NSIS%203/3.10/nsis-3.10.zip/download"
+          $zip = "$env:RUNNER_TEMP\nsis.zip"
+          Invoke-WebRequest -Uri $url -OutFile $zip
+          Expand-Archive -Path $zip -DestinationPath "$env:RUNNER_TEMP\nsis"
+          echo "NSIS_DIR=$env:RUNNER_TEMP\nsis\NSIS" >> $env:GITHUB_ENV
+
+      - name: Build Windows installer
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          & scripts/build-windows-installer.ps1 `
+            -NsisDir "$env:NSIS_DIR" `
+            -StagingDir "dist/package/zero-v$env:VERSION-windows-$env:ARCH" `
+            -OutputDir "dist/release" `
+            -Version $env:VERSION `
+            -Arch $env:ARCH
+
+      # Linux AppImage + system installer
+      - name: Install Linux build dependencies
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq imagemagick
+          wget -q "https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-$(uname -m).AppImage" -O /tmp/appimagetool
+          chmod +x /tmp/appimagetool
+          cd /tmp && ./appimagetool --appimage-extract >/dev/null 2>&1
+          sudo cp squashfs-root/AppRun /usr/local/bin/appimagetool
+          rm -rf /tmp/appimagetool /tmp/squashfs-root
+
+      - name: Build AppImage
+        if: runner.os == 'Linux'
+        run: |
+          bash scripts/build-linux-appimage.sh \
+            --staging-dir "dist/package/zero-v${VERSION}-linux-${ARCH}" \
+            --output-dir "dist/release" \
+            --version "${VERSION}" \
+            --arch "${ARCH}"
+
+      - name: Build Linux system installer
+        if: runner.os == 'Linux'
+        run: |
+          bash scripts/build-linux-install.sh \
+            --staging-dir "dist/package/zero-v${VERSION}-linux-${ARCH}" \
+            --output-dir "dist/release" \
+            --version "${VERSION}" \
+            --arch "${ARCH}"
+
+      # macOS .app
+      - name: Build .app bundle
+        if: runner.os == 'macOS'
+        run: |
+          bash scripts/build-macos-app.sh \
+            --staging-dir "dist/package/zero-v${VERSION}-macos-${ARCH}" \
+            --output-dir "dist/release" \
+            --version "${VERSION}" \
+            --arch "${ARCH}"
 
       - name: Upload package
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
@@ -170,7 +246,15 @@ jobs:
             "zero-v${version}-linux-arm64.tar.gz" \
             "zero-v${version}-macos-x64.tar.gz" \
             "zero-v${version}-macos-arm64.tar.gz" \
-            "zero-v${version}-windows-x64.zip"; do
+            "zero-v${version}-windows-x64.zip" \
+            "zero-v${version}-windows-x64-installer.exe" \
+            "zero-v${version}-linux-x86_64.AppImage" \
+            "zero-v${version}-linux-x86_64-install.run" \
+            "zero-v${version}-macos-x64.zip" \
+            "zero-v${version}-windows-arm64-installer.exe" \
+            "zero-v${version}-linux-aarch64.AppImage" \
+            "zero-v${version}-linux-aarch64-install.run" \
+            "zero-v${version}-macos-arm64.zip"; do
             for url in "${base}/${asset}" "${base}/${asset}.sha256"; do
               if ! curl -fsSLI -o /dev/null "$url"; then
                 echo "::error::${url} is not publicly downloadable — npm postinstall would fail. Is the repo public and are all assets uploaded?" >&2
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (changes requested):
I read all three files in full and checked the workflow against main. One blocker before this is mergeable on its own: the release workflow added here calls `scripts/build-linux-appimage.sh` and `scripts/build-linux-install.sh` (lines 127 and 136), but neither script is in this PR or on main — so merging #648 alone would break the Linux matrix legs and the downstream asset-download verification. The NSIS installer itself looks sound (per-machine install dir, uninstaller, Start Menu/desktop shortcuts, PATH + Add/Remove Programs entries), though PATH is written to HKCU under an admin-elevated installer, which can land on the wrong user's environment. The macOS .app bundle is correctly structured (Info.plist, ZeroLauncher executable, ad-hoc signed, no notarization — fine for OSS but worth a docs note). CI hygiene is good: actions are SHA-pinned, permissions are minimal, and the trigger is tag-push / workflow_dispatch only (no pull_request_target with secrets). No Go files touched, so I skipped build/vet.

Verdict: request changes for the missing Linux scripts. On the duplicate question: #649 is a strict superset — it contains all 7 files from #646 + #648, and the three files shared with this PR are byte-identical. I'd take #649 as canonical and close #646 and #648 as duplicates.

jatmn (changes requested):
## Blockers (must fix before merge)

### B1. Linux build scripts referenced but do not exist — entire Linux release + npm publish broken
`.github/workflows/release-artifacts.yml:124-140` invokes `bash scripts/build-linux-appimage.sh` and `bash scripts/build-linux-install.sh`. Neither file exists in this PR, on `origin/main`, or anywhere in git history (confirmed via `git ls-files` and `git ls-tree origin/main`). Both Linux matrix legs fail; and the npm-verify gate (`:244-257`) requires `linux-x86_64.AppImage`, `linux-x86_64-install.run`, `linux-aarch64.AppImage`, `linux-aarch64-install.run` — assets that can never be produced. Because `publish-npm` has `needs: package` (which requires **all** matrix legs to succeed; `fail-fast: false` does not relax this), npm publish is skipped on every release.
**Fix:** Land the Linux scripts in this PR (or merge them first and rebase). Until then the workflow cannot pass.

### B2. macOS `sed -i` is BSD-incompatible and aborts the build
`scripts/build-macos-app.sh:61` — `sed -i "s/__VERSION__/${VERSION}/g" "${APP_BUNDLE}/Contents/MacOS/ZeroLauncher"`. macOS `sed` requires an extension argument for `-i`; without it the command errors and `set -e` aborts the script, so no `.app`/`.zip` is produced and `__VERSION__` is never substituted. Verified locally on macOS: `sed -i 's/a/b/' file` → `sed: 1: ... unescaped newline ...` exit 1.
**Fix:** `sed -i '' "s/__VERSION__/${VERSION}/g" file` (or use `perl -pi -e`).

### B3. macOS uses `sha256sum`, not available on stock macOS
`scripts/build-macos-app.sh:150` — `sha256sum "${APPZIP}" | sed ...`. `sha256sum` is a GNU coreutils utility not shipped with stock macOS and not guaranteed on GitHub's macOS runners. Under `set -euo pipefail` a missing command aborts the build before the `.sha256`/`.zip` is produced, so the macOS assets never reach the npm-verify gate.
**Fix:** Use `shasum -a 256` (always present on macOS). The output format (`hash  filename`, two spaces) is identical, so consumers are unaffected.

### B4. Windows `NSIS_DIR` points at a path that does not exist
`.github/workflows/release-artifacts.yml:99` — `echo "NSIS_DIR=$env:RUNNER_TEMP\nsis\NSIS"`. The NSIS 3.10 zip extracts to a top-level folder named **`nsis-3.10`** (verified by downloading `nsis-3.10.zip`: top-level entry `nsis-3.10`, `makensis.exe` at `nsis-3.10/makensis.exe`). So `$NsisDir/makensis.exe` is never found → both Windows legs throw → no `*-installer.exe` → npm-verify fails.
**Fix:** Resolve the NSIS dir dynamically, e.g. `$nsis = (Get-ChildItem $env:RUNNER_TEMP\nsis -Recurse -Filter makensis.exe | Select-Object -First 1).DirectoryName; echo "NSIS_DIR=$nsis"`, or hardcode `nsis-3.10\NSIS`.

### B5. Windows `File` source path is resolved relative to the `.nsi` in `$TEMP`
`scripts/build-windows-installer.ps1` writes the NSIS script to `$env:TEMP\zero-installer.nsi`, then embeds `File /r "${StagingDir}\*.*"` with a **relative** `StagingDir` (`dist/package/zero-v…-windows-x64`). NSIS resolves `File` source paths relative to the **`.nsi`'s directory** (`$TEMP`), not the repo cwd, so the staging files are never found and `makensis` fails on every run.
**Fix:** Make `StagingDir` absolute inside the script (e.g. `$StagingDir = (Resolve-Path $StagingDir).Path`) before writing it into the `.nsi`, or write the `.nsi` into the repo root.

---

## High-severity functional defects (should fix before merge)

### H1. Windows `un.StrDel` wipes the user's entire PATH on uninstall
`scripts/build-windows-installer.ps1`, `Function un.StrDel` (≈ lines 82-111). The scan loop shifts the string from the front (`StrCpy $R2 $R2 "" 1`) until `$INSTDIR` is at the front, then keeps only what *follows* it — discarding everything *before* `$INSTDIR`. Since the installer appends `;$INSTDIR` to the **end** of Path, uninstall deletes the whole pre-existing PATH (everything before `$INSTDIR`) and leaves a stray leading `;`. This is catastrophic, irreversible user-data loss. (Confirmed by two independent reviewers and by tracing the logic; one reviewer's dismissal of this as a false alarm is incorrect.)
Also note `StrCmp $R4 $R1 0 next` references an undefined label `0` — at minimum suspicious, likely a compile error.
**Fix:** Replace with a delimiter-aware removal (split on `;`, drop the exact `$INSTDIR` entry, rejoin). Also de-duplicate on install so re-running does not append duplicate PATH entries.

### H2. macOS `.app` ships helpers in the wrong location — core feature broken
`scripts/build-macos-app.sh:138-140` copies helpers to `Contents/Helpers`, but the Go binary's `adjacentHelper` (`internal/localcontrol/browser.go:331-336`) only searches `executableDir` and `executableDir/helpers` (= `Contents/MacOS/helpers`). There is **no Go reference to `ZERO_HELPERS_DIR` anywhere** (grep confirms). So the macOS app's browser / local-control helper features are never discovered → broken for end users. The `ZERO_HELPERS_DIR` exported by the launcher is dead.
**Fix:** Copy helpers to `Contents/MacOS/helpers` (or symlink), and drop the unused env var.

### H3. macOS launcher does not actually run Zero
`scripts/build-macos-app.sh` — `ZeroLauncher` writes a temp script via `mktemp /tmp/zero-launch.XXXXXX` (no `.command` extension, and the temp file has no `#!/bin/bash` shebang) and runs `open -a Terminal "$TMPFILE"`. `open` with a file lacking a `.command` extension opens it as a *document* in Terminal; it is not executed, so `zero` never launches. The app's sole purpose fails.
**Fix:** Create `mktemp ...XXXXXX.command`, prepend `#!/bin/bash`, `chmod +x`, then `open -a Terminal "$TMPFILE"`.

### H4. macOS icon is never embedded (wrong asset path)
`scripts/build-macos-app.sh:114` sets `ICON_SRC="docs/assets/logo-great.png"`, but the repo only contains `docs/assets/zero-logo.png`. The icon step is always skipped (warning), while `Info.plist` sets `CFBundleIconFile=zero` → the app ships with no icon.
**Fix:** Point `ICON_SRC` at `docs/assets/zero-logo.png`.

---

## Medium / should-fix

- **M1. Privilege/registry scope mismatch (Windows).** Installer uses `RequestExecutionLevel admin` + `$PROGRAMFILES64` + HKLM uninstall keys, but writes PATH to **HKCU**. When UAC-elevated, HKCU is the *admin's* profile, so the invoking (standard) user's PATH is not updated → Zero is not on their PATH. (Mixed per-user/machine scope.)
- **M2. Arch-token vocabulary mismatch for Linux (latent).** The workflow passes `--arch x64/arm64` to the (missing) Linux scripts, but the npm-verify list hardcodes `x86_64`/`aarch64`. Even once the Linux scripts land, one side will 404 unless coordinated with PR #646's naming.
- **M3. `case` has no default.** `release-artifacts.yml:55-62` — if a runner type is unmatched, `ARCH` is unset and is used unconditionally later, producing mis-named assets. Add `*) echo "::error::..."; exit 1 ;;`.
- **M4. `jq` dependency in the Resolve step.** `release-artifacts.yml:54` uses `jq` on all OSes, including the `windows-11-arm` preview image where it may be absent → empty `VERSION`. The existing tag-check step already uses `node -p "require('./package.json').version"`; reuse that for consistency.
- **M5. Supply chain: NSIS & appimagetool fetched without integrity verification.** `release-artifacts.yml:95-99` downloads NSIS from a SourceForge `/download` mirror redirect; `:118-121` pulls the rolling `continuous` AppImageKit and `sudo cp`s it to `/usr/local/bin` as root. Neither is checksum/signature verified — a compromised mirror/MITM could inject a trojaned installer or root binary. At minimum pin/checksum; prefer a verified NSIS source.
- **M6. Windows-ARM assets orphaned by `postinstall.mjs`.** The workflow builds + verifies `windows-arm64-installer.exe` (and `windows-arm64.zip`), but `scripts/postinstall.mjs:106-113` explicitly refuses `windows/arm64` ("no prebuilt binary for windows-arm64"), so `npm install` on Windows ARM errors. Either drop `windows-11-arm` from the matrix or teach `postinstall.mjs` to use the arm64 asset.

---

## Low / notes

- **L1.** `windows-arm64.zip` is produced by `zero-release package` but omitted from the npm-verify list (harmless given M6).
- **L2.** `codesign --force --deep --sign -` failures are suppressed (`build-macos-app.sh:142-143`); ad-hoc signing won't satisfy Gatekeeper, and the "non-fatal" success is misleading.
- **L3.** `publish-npm` runs `gh release create ... || true` from each matrix job — low race risk due to `|| true` + `--clobber`.
- **L4.** `windows-11-arm` is a preview GitHub runner; availability/flakiness caveat.
- **L5.** macOS `.sha256` uses an absolute path (Windows uses a bare basename) — less portable but `sha256sum -c` still works; format itself is the standard two-space form (no collapse bug).

---

Vasanthdev2004:
Thanks for this, @JohnVictorCrown - the cross-platform installer work is genuinely useful and clearly took real effort. It's a subset of your own #649, which carries these same Windows and macOS scripts alongside the Linux build scripts this branch's release workflow actually calls, so it can't pass the release gate standalone. Let's consolidate on #649 as the canonical PR and close this and #646 as duplicates. There's good feedback on #649 to work through (the sed -i / sha256sum macOS portability fixes, the NSIS path resolution, the uninstall PATH handling). Appreciate the effort, and reopen if you'd rather land these separately.

PR #649, opened 2026-07-10
feat(ci): cross-platform installer build scripts and CI pipeline

JohnVictorCrown:
## Summary

Adds build scripts and CI workflow changes to produce platform-specific installer artifacts for all three platforms during releases.

### New files (4 scripts + 1 plan doc + 1 logo)

| File | Purpose |
|------|---------|
| \scripts/build-windows-installer.ps1\ | NSIS-based \.exe\ installer with PATH, shortcuts, uninstaller |
| \scripts/build-linux-appimage.sh\ | Portable AppImage with AppRun, .desktop, AppStream metainfo, icons |
| \scripts/build-linux-install.sh\ | Self-extracting \.run\ system installer (\/usr/local/\ via \install -m 755\) |
| \scripts/build-macos-app.sh\ | \.app\ bundle with Terminal launcher, .icns icon, ad-hoc signature |
| \docs/BUILD_INSTALLERS_PLAN.md\ | Full cross-platform design doc |
| \docs/assets/logo-great.png\ | Source icon for all installer branding |

### Modified files (1)

| File | Change |
|------|--------|
| \.github/workflows/release-artifacts.yml\ | Added \windows-11-arm\ runner, arch/version env, platform installer steps, npm asset verification |

### Artifacts per release

| Platform | Runner | New artifacts |
|----------|--------|-------------|
| Windows x64 | \windows-latest\ | \zero-v*-windows-x64-installer.exe\ |
| Windows arm64 | \windows-11-arm\ | \zero-v*-windows-arm64-installer.exe\ |
| Linux x86_64 | \ubuntu-latest\ | \zero-v*-linux-x86_64.AppImage\ + \-install.run\ |
| Linux aarch64 | \ubuntu-24.04-arm\ | \zero-v*-linux-aarch64.AppImage\ + \-install.run\ |
| macOS x64 | \macos-15-intel\ | \zero-v*-macos-x64.zip\ (Zero.app) |
| macOS arm64 | \macos-latest\ | \zero-v*-macos-arm64.zip\ (Zero.app) |

All new assets include \.sha256\ checksum files and are automatically uploaded by the existing \gh release upload dist/release/*\ glob.

## Summary by CodeRabbit

* **New Features**
  * Added downloadable Windows installers for x64 and ARM64.
  * Added Linux AppImage and system installer packages for x64 and ARM64.
  * Added macOS application packages for x64 and ARM64.
  * Release packages now include SHA256 checksum files for verification.
  * Windows installers provide Start Menu/Desktop shortcuts and automatic PATH setup.

* **Documentation**
  * Added a release-pipeline plan describing supported platforms, package formats, and future distribution enhancements.


Vasanthdev2004 (commented):
I read all four build scripts and the release-artifacts workflow diff end to end. Coverage is complete — Linux AppImage + self-extracting .run, Windows NSIS installer (x64 and arm64), and a macOS .app bundle zipped per arch, all wired into the existing matrix with pinned actions, minimal permissions, and no pull_request_target. The shell scripts use set -euo pipefail, mktemp -d with trap cleanup, and root checks; the .app has a proper Info.plist, executable launcher, and ad-hoc codesign; the NSIS installer writes an uninstaller, ARP entry, and PATH management. It merges clean against current main and no Go files are touched so build/vet is N/A.

One note: the NSIS download from sourceforge and the appimagetool pull from the `continuous` release (release-artifacts.yml) are not checksum-pinned — both run only on tag/dispatch, never on PRs, so the secret surface is fine, but pinning those would harden the supply chain.

On the three competing PRs: #649 is a byte-identical superset of #646 and #648 — I diffed every file across the PR heads and they match exactly, and #649's two commits are literally the other two squashed together. I'd treat #649 as canonical and close #646 and #648 as duplicates.

jatmn (changes requested):

## Merge readiness
- [P2] Validate this change against the current target’s release workflow before merge
  `.github/workflows/release-artifacts.yml:1`
  The head is based on `1af58828`, while the live target is `1b5db176` and has materially changed the same release/npm pipeline. GitHub's merge computation is conflict-free, but the PR's passing checks predate those changes. Rebase or test the merged result, then rerun the release-specific checks: the installer additions must compose with the current packaging and npm-publishing contract, not merely with the historical base.

## Findings

- [P1] Point `NSIS_DIR` at the directory that is actually extracted
  `.github/workflows/release-artifacts.yml:95-99`
  `nsis-3.10.zip` extracts `nsis-3.10/makensis.exe`, but this exports `...\nsis\NSIS`. The subsequent script resolves `$NsisDir\makensis.exe`, so both Windows release jobs stop before an installer is produced. The root cause is treating an assumed extraction layout as the compiler root. Set `NSIS_DIR` to the actual extracted `nsis-3.10` directory (or discover the compiler path after extraction), then fail explicitly if `makensis.exe` is absent.

- [P1] Install a functional AppImage tool instead of its detached launcher
  `.github/workflows/release-artifacts.yml:118-122`
  After extraction, this copies only `squashfs-root/AppRun` and immediately deletes `squashfs-root`. `AppRun` is a launcher for the executable/resources inside that directory, so the later `build-linux-appimage.sh` invocation has no runnable `appimagetool`; Linux installer builds fail before producing either artifact. Install the actual appimagetool binary with the resources it requires, or retain and invoke the extracted AppImage layout rather than relocating its launcher alone.

- [P1] Normalize the `.run` asset architecture before naming it
  `scripts/build-linux-install.sh:26-27`
  The workflow supplies `ARCH=x64|arm64`, causing this script to emit `linux-x64-install.run` / `linux-arm64-install.run`, while the publish gate requires `linux-x86_64-install.run` / `linux-aarch64-install.run`. Every tag release therefore uploads one name and probes another, receives a 404, and blocks npm publishing. Normalize the installer artifact architecture at the producer boundary—the same x86_64/aarch64 mapping already used by the AppImage builder—so the build, upload, and preflight share one naming contract. Add an assertion that the generated filename is in the asset list.

- [P1] Make the macOS builder valid for the runner and its relative output directory
  `scripts/build-macos-app.sh:61,145-150`
  macOS BSD `sed` does not accept this GNU `sed -i` form, so the launcher's version substitution fails. Independently, after changing into `dist/release`, the already-relative `APPZIP=dist/release/...` resolves as `dist/release/dist/release/...`, whose parent does not exist. Either fault prevents both macOS artifacts from being built. Use a BSD-compatible substitution approach and keep path semantics consistent: either make `APPZIP` absolute before `cd`, or pass its basename after changing directory. A focused macOS packaging smoke test should cover the workflow's relative `--output-dir` argument.

- [P1] Do not corrupt the user's PATH during Windows uninstall
  `scripts/build-windows-installer.ps1:81-110`
  `un.StrDel` discards every PATH component before the first `$INSTDIR` match: with `A;C:\Program Files\Zero;B`, the loop consumes `A;`, finds the match in its shortened buffer, and returns only `;B`. Uninstalling Zero can therefore remove unrelated executable paths and break other applications. Replace the substring algorithm with component-aware PATH removal that reconstructs the original prefix and suffix, removes only exact Zero entries, and is safe when Zero is absent, first, last, or duplicated. Test those cases before shipping an uninstaller that edits user registry state.

- [P1] Honor the Windows ARM64 archive added by this PR in npm postinstall
  `scripts/postinstall.mjs:107-117`
  This PR adds a Windows ARM64 package job and archive, but the npm consumer still exits early for `windows/arm64` with “no prebuilt binary” and installs no native binary. The new release output is therefore unreachable to npm users on the platform it was added for. Make the producer and consumer agree: remove/update the obsolete skip, update its regression test, and ensure the downloader resolves and verifies the new ARM64 ZIP. If ARM64 npm delivery is intentionally unsupported, remove the newly claimed support instead of publishing unreachable artifacts.

- [P1] Verify immutable release build-tool inputs before executing them
  `.github/workflows/release-artifacts.yml:95-98`
  `.github/workflows/release-artifacts.yml:118-121`
  The tag-release job downloads and executes an unchecked NSIS ZIP and an unchecked AppImageKit artifact from the mutable `continuous` channel while holding `contents: write`. A substituted upstream artifact can alter the binaries uploaded to the GitHub Release. Treat both as release-toolchain inputs: select immutable artifact URLs/versions, record trusted SHA-256 values in reviewed source, verify the downloaded bytes before extraction or execution, and fail closed on mismatch. This directly addresses the supply-chain boundary noted by Vasanthdev2004.

- [P2] Preserve the literal NSIS Program Files variable in the generated script
  `scripts/build-windows-installer.ps1:20`
  The expandable PowerShell here-string consumes `$PROGRAMFILES64`, so the generated NSIS source contains `InstallDir "\Zero"` instead of the intended NSIS variable. The installer defaults to the drive root rather than Program Files. The template must distinguish PowerShell interpolation from NSIS interpolation: escape the dollar sign so the generated source retains literal `$PROGRAMFILES64`, and add a generation-level assertion for the expected `InstallDir` directive.

- [P2] Put bundled local-control helpers where the existing resolver can find them
  `scripts/build-linux-appimage.sh:61-72`
  `scripts/build-linux-install.sh:45,103-106`
  `scripts/build-macos-app.sh:45-53,138-140`
  The resolver only checks locations adjacent to the executable (or a supported manifest), but these installers put helpers in `usr/lib/zero`, `<prefix>/lib/zero`, and `Contents/Helpers`; `ZERO_HELPERS_DIR` is not consumed by production code. The new packages silently lose their bundled `agent-browser`/`tuistory` helpers and fall back to whatever happens to be on PATH. Place helpers in an existing adjacent discovery location or generate the supported helper manifest at launch/install time. Validate each artifact by resolving both helpers from its installed layout rather than only asserting that the directories were copied.

- [P2] Serialize the configured installation prefix into the Linux desktop launcher
  `scripts/build-linux-install.sh:47-59`
  The installer supports `INSTALL_PREFIX`, but always writes `Exec=/usr/local/bin/zero`. A custom-prefix install such as `/opt/zero` creates a launcher that fails or invokes a different old installation. Defer desktop-entry rendering until the installer knows the selected prefix (or substitute it into the payload), and test a nondefault prefix to ensure the serialized launcher matches the installed binary.

- [P2] Verify the Windows ARM64 base archive alongside the new installer
  `.github/workflows/release-artifacts.yml:244-257`
  `install.ps1` resolves `zero-v${version}-windows-arm64.zip`, but this pre-publish loop checks only the ARM64 installer. A missing or misnamed base archive can pass the gate and leave the documented PowerShell install path broken. Make the verification list derive from, or be tested against, the consumer asset matrix so it includes both the installer and the base ZIP plus their checksum sidecars.

PR #650, opened 2026-07-10
fix(agent): raise default and deep-mode turn budgets

euxaristia:
## Problem

The default per-run tool-turn budget was 50. For larger multi-step tasks that span several files, agents hit the ceiling and stopped mid-task, returning a "remaining work" summary (via the max-turns final-answer prompt) instead of finishing the job. A user hit exactly this on a multi-file feature (cloud sync, WebSocket bridge, QR pairing, PWA endpoint logic, verification): the agent reported "ran out of turns" with a checklist of unfinished items.

Separately, the `deep` preset also carried `MaxTurns: 50`, identical to the default. Its description promises "a larger turn budget," but on the turn-budget axis it was no different from a normal run.

## Fix

- `defaultMaxTurns`: 50 to 80 (`internal/config/resolver.go`). This is the budget most runs use, including the interactive TUI and plain `zero exec`, so it is the change that addresses the reported case.
- `deep` preset `MaxTurns`: 50 to 160 (`internal/modelregistry/modes.go`), so `deep` is a genuine 2x the default again.

Both remain overridable per-session with `/turns` or `--max-turns`, still bounded by `MaxTurnsCeiling` (500). Precedence is unchanged: an explicit `--max-turns` still wins over a mode preset.

## Testing

- `go build ./...`
- `go test ./internal/config/... ./internal/modelregistry/...`

## Summary by CodeRabbit

* **Improvements**
  * Increased the default tool-turn budget per run from 50 to 80.
  * Increased the maximum turns available in the deep mode from 50 to 160.
  * You can still adjust the per-session turn limit using `/turns`.

### DIFF
--- a/internal/cli/exec_test.go
+++ b/internal/cli/exec_test.go
@@ -345,8 +345,8 @@ func TestRunExecModeSeedsModelAndTurnOverrides(t *testing.T) {
 	if gotModel != "claude-opus-4.1" {
 		t.Fatalf("overrides.Provider.Model = %q, want claude-opus-4.1", gotModel)
 	}
-	if gotMaxTurns != 50 {
-		t.Fatalf("overrides.MaxTurns = %d, want 50", gotMaxTurns)
+	if gotMaxTurns != 160 {
+		t.Fatalf("overrides.MaxTurns = %d, want 160 (deep mode)", gotMaxTurns)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (approved):
Tiny and clean — just bumping two constants. defaultMaxTurns goes 50 -> 80 and deep-mode MaxTurns 50 -> 160, both still well under the MaxTurnsCeiling=500 clamp, so there's no unbounded-loop risk; the diff is constants-only with no control-flow change, and the exec_test assertion was updated to match. Builds, vet, and gofmt are clean, and the config/modelregistry packages plus the updated cli test all pass (the remaining cli failures are the pre-existing "provider anthropic requires model" noise that reproduces on main). Approving.

kevincodex1 (approved):
LGTM

PR #654, opened 2026-07-10
fix(sandbox): unblock git fetch/commit/add under the write-restricted sandbox

euxaristia:
## Summary
- The sandbox denied every write under `.git` for shell-executed commands, so `git fetch`, `commit`, `add`, `pull`, `merge`, and `stash` all failed under the default sandbox since they need to write `FETCH_HEAD`, the index, objects, and refs.
- Narrows the write block to `.git/hooks` (auto-executing scripts) and `.git/config` (remote URLs, `credential.helper`, `core.hooksPath`), the two subpaths that are actually dangerous to write. The rest of `.git` is now writable to git subprocesses, since those writes go through git's own consistency checks.
- The separate app-level auto-allow gate for direct file-tool writes (`write_file`, `edit_file`, `apply_patch`) is untouched and still treats all of `.git` as protected, since a raw byte-level write to `.git` bypasses git's invariants in a way a git subprocess write does not.

## Test plan
- [x] `go test ./...` passes
- [x] Added `TestSeatbeltProfileAllowsGitWritesExceptHooksAndConfig` verifying the generated seatbelt profile no longer blanket-denies `.git` and still denies `.git/hooks` / `.git/config`
- [x] Added `TestSeatbeltAllowsGitMetadataWritesExceptHooksAndConfig`, a real `sandbox-exec` integration test proving a write to `.git/FETCH_HEAD` succeeds while writes to `.git/hooks/*` and `.git/config` are kernel-denied
- [x] Manually verified `git fetch origin` on a real clone succeeds end to end through the sandbox engine
- [x] Updated the CLI golden JSON fixture for `zero sandbox policy --json`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox handling of Git metadata for shell-executed commands.
  * Git metadata writes are now permitted for non-sensitive files (e.g., `FETCH_HEAD`), while `.git/hooks` and `.git/config` are kept read-only for stronger protection.
  * Updated portable JSON root normalization on Windows to consistently use forward slashes.
* **Tests**
  * Added/extended sandbox and seatbelt integration tests to verify the new Git read-only carve-outs behavior.

### DIFF
--- a/internal/cli/sandbox_test.go
+++ b/internal/cli/sandbox_test.go
@@ -616,8 +616,12 @@ func jsonValuesEqual(t *testing.T, wantBytes []byte, gotBytes []byte) bool {
 func normalizePortableJSONRootString(value *any) {
 	switch current := (*value).(type) {
 	case string:
-		if current == `\` {
-			*value = "/"
+		// Normalize OS-specific path separators so the shared golden file
+		// matches on both Unix (slash) and Windows (backslash) runners.
+		// readOnlySubpaths carve-outs are built with filepath.Join, which
+		// renders as backslash paths on Windows.
+		if current != "" {
+			*value = strings.ReplaceAll(current, `\`, "/")
 		}
 	case []any:
 		for index := range current {
--- a/internal/cli/testdata/sandbox_policy_windows_unavailable.golden.json
+++ b/internal/cli/testdata/sandbox_policy_windows_unavailable.golden.json
@@ -40,8 +40,11 @@
         "writeRoots": [
           {
             "root": "$WORKSPACE",
+            "readOnlySubpaths": [
+              "$WORKSPACE/.git/hooks",
+              "$WORKSPACE/.git/config"
+            ],
             "protectedMetadataNames": [
-              ".git",
               ".zero",
               ".agents"
             ]
--- a/internal/sandbox/manager_test.go
+++ b/internal/sandbox/manager_test.go
@@ -41,8 +41,15 @@ func TestPermissionProfileFromPolicyBuildsWorkspaceWriteProfile(t *testing.T) {
 	if !stringSliceContains(profile.FileSystem.ReadRoots, profileRootPath()) {
 		t.Fatalf("read roots = %#v, want full read root %q", profile.FileSystem.ReadRoots, profileRootPath())
 	}
-	if !stringSliceContains(profile.FileSystem.WriteRoots[0].ProtectedMetadataNames, ".git") || !stringSliceContains(profile.FileSystem.WriteRoots[0].ProtectedMetadataNames, ".zero") {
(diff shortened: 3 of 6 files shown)


Vasanthdev2004 (approved):
Built, vetted, and ran the sandbox tests (including the new seatbelt profile test and the updated golden). Approving.

This is the right shape and it's a targeted unblock, not a jail widening. The reason git ops were blocked: `.git` sat in the OS-level protected-metadata set, which produced a blanket write-deny of the whole `.git` tree and killed fetch/commit/add/merge at the kernel before git's own invariants ever ran. The fix splits the two threat models: a new `sandboxFullyProtectedMetadataNames` (`.zero`, `.agents`) drives the OS-level full-deny, with `.git` removed from it; only `.git/hooks` and `.git/config` — the two subpaths where a write becomes code execution or a credential/remote hijack — stay write-denied via `ReadOnlySubpaths`. Everything else under `.git` (objects, refs, index, HEAD, FETCH_HEAD) becomes writable to git subprocesses. The writable root is still the workspace; nothing outside it becomes newly writable. The app-level file-tool prompt gate still includes `.git`, so a raw edit_file/apply_patch against `.git` internals still prompts — good separation.

I also confirmed the earlier sandbox work isn't undone: no token-flag changes (so the DenyRead fix is intact), and the Windows ACL carve-outs are deny entries against the existing per-root capability SID, not new SIDs granted outside the workspace.

One intended tradeoff to note: because `.git/config` stays write-denied, `git config <key> <value>` and `git remote add` will fail inside a sandboxed run. That's the security posture this PR wants (blocking core.hooksPath / credential.helper rewrites is the point), and global `~/.gitconfig` stays readable, but it's worth knowing for users who set per-repo git config in sandboxed sessions.

Vasanthdev2004:
@gnanam1990 could you review this sandbox change? It unblocks git fetch/commit/add by removing .git from the OS-level full-deny set while keeping .git/hooks and .git/config write-denied — the writable root stays the workspace, so it's not a jail widening. Security-relevant, so a second pair of eyes would be good before it lands.

gnanam1990 (approved):
## Verdict: Approve

Aligns with @Vasanthdev2004's review. Targeted sandbox unblock for git subprocesses, not a jail widening.

### What looks good

- Splits threat models correctly: `.git` removed from OS-level full deny; only `.git/hooks` and `.git/config` stay write-denied via `ReadOnlySubpaths`.
- Git subprocesses can write objects/refs/index/FETCH_HEAD; app-level file tools still prompt on `.git` internals.
- New seatbelt unit test plus Darwin integration test (`TestSeatbeltAllowsGitMetadataWritesExceptHooksAndConfig`); Windows golden updated.
- **Verification:** `make lint` clean; `./internal/sandbox/...` and relevant CLI tests pass.

### Nits

- `git config` / `git remote add` fail inside sandboxed runs — intentional security posture; worth a one-line docs note for users.

### Issues

None blocking. Ready to merge.

kevincodex1 (approved):
LGTM

PR #635, opened 2026-07-10
docs: add Go code quality & security checks and bump Go version to 1.26+

euxaristia:
This PR adds instructions to run Go formatting, vetting, linting, and vulnerability checks to README/AGENTS.md, and bumps the documented Go requirement/badges from 1.25+ to 1.26+.

## Summary by CodeRabbit

- **Documentation**
  - Updated Go version requirements to **Go 1.26.5+** across English and Chinese docs, including installation and “from source” build steps.
  - Added a **Development → Code Quality and Security Checks** checklist for commit preparation (formatting, vetting, linting, and vulnerability scanning).
  - Added a new **EXTENDING** guide covering project instruction discovery, skills, hooks, MCP, and plugin configuration.
- **Chores**
  - Rewrote **AGENTS.md** with clearer repository conventions for AI coding agents, including required quality/security checks and guidance when checks can’t run.

PR #655, opened 2026-07-10
feat(aimlapi): AI/ML API provider with guided onboarding (top-up + key issuance)

### LINKED ISSUE #651: AI/ML API provider: guided onboarding (top-up + key issuance)
### Problem

Zero has no first-class way to get started with AI/ML API (aimlapi.com), an
OpenAI-compatible aggregating gateway that fronts 1000+ models. A new user has
to leave the CLI, create an account on the website, top up a balance, copy an
API key, and paste it back into a provider profile by hand.

### Requested behavior

Add AI/ML API as a built-in provider with a guided TUI onboarding sub-flow that:
- takes an existing API key, or
- signs the user in by email and tops up the balance via the hosted
  partner-checkout page, then writes the issued key straight into the provider
  profile.
The provider is registered in the catalog (default model + OpenAI-compatible
base URL) and sends the partner attribution headers on inference requests.
Default endpoints point at production and each is overridable via AIMLAPI_*_URL
env vars for local/staging testing.

### Why it belongs in Zero

It removes the biggest onboarding friction for a large multi-model gateway:
users go from "no provider" to a working, funded key without leaving the TUI.
It follows the existing provider-catalog + onboarding-wizard patterns, so it is
additive and consistent with how other providers are surfaced.

- Manual setup: add a custom OpenAI-compatible endpoint and paste a key obtained
  from the website. Works, but loses the guided top-up and partner attribution.
- Documentation only (a how-to in the README) — does not solve the in-CLI
  top-up / key-issuance step.

Based on the working AI/ML API integration in OpenClaude, reworked and polished
for a smoother onboarding experience in Zero. Screenshots of the onboarding
screen to be attached.

Lookoff-AIMLAPI:
## AI/ML API guided onboarding

Adds [AI/ML API](https://aimlapi.com) (aimlapi.com) as a built-in provider with a guided TUI onboarding sub-flow. Instead of leaving the CLI to register, top up, and paste a key by hand, the user goes from "no provider" to a working, funded key without leaving Zero.

Reached from both the first-run setup and the `/provider` wizard by picking **aimlapi.com** from the provider list.

Fixes #651

### Onboarding paths
- **Have a key** — paste it; the key is validated against `/v1/billing/balance` and saved. A low balance offers an optional top-up.
- **New / email** — sign in by email code (or passwordless sign-up), then top up via the hosted partner-checkout page. The browser opens the checkout, the CLI polls the session, and the issued key is written into the provider profile on success.

### Changes
- `internal/aimlapi` — OpenAI-compatible client, partner-checkout top-up with streamed progress, email/key validation, and endpoint config.
- `providercatalog` / `providermodelcatalog` — register the aimlapi provider (badged recommended) with default headers and a curated model list.
- `tui` — the shared `aimlapiOnboardState` sub-flow, wired into the provider wizard and first-run onboarding; the spinner tick stays alive during the async waits.
- `cli` / `config` — merge the provider's custom headers into `providers add aimlapi` and catalog-resolved profiles.

### Implementation notes
- Took the working OpenClaude integration as a reference and reworked it for a smoother onboarding experience in Zero.
- The sub-flow owns all of its own state and transitions behind a pointer; the two host surfaces only route keys + async messages into it and read the terminal result.
- No new runtime dependencies; endpoints and partner id are overridable via env.

### Testing
- `gofmt`, `go vet ./...`, `go build ./...`, and `go test ./...` pass locally (all packages).
- `go run ./cmd/zero-release build` + `smoke`, `zero-perf-bench --ci`, and `govulncheck ./...` pass — the blocking gates of the three CI jobs.
- New tests cover `internal/aimlapi` (client, config, onboarding) and the `internal/tui` onboarding flow.
- `-race` was not run locally (no cgo toolchain on the dev box); CI's `go test ./...` covers the suite.

## Summary by CodeRabbit

* **New Features**
  * Added AIMLAPI as a recommended provider with an expanded model preset catalog.
  * Introduced a guided AIMLAPI onboarding flow (API key or passwordless email verification) with automated API key provisioning.
  * Added session-based partner checkout with payment progress, browser checkout links, balance checking, optional auto top-up, and by-key top-ups (with resume support).
  * Added environment-driven endpoint and return-URL configuration.
  * Extended the provider wizard to reuse existing AIMLAPI credentials and show a “configured” preflight step.
* **Bug Fixes**
  * Improved header merging/overrides and refined AIMLAPI TUI link rendering, spinner/progress transitions, and back-navigation.
* **Tests**
  * Added/expanded AIMLAPI client, config, onboarding, provider wizard, and streaming top-up test coverage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,7 +41,6 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 
 /zero
 /zero.exe
-/zero.exe~
 /zero-windows-command-runner.exe
 /zero-windows-sandbox-setup.exe
 /zero-skills
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 
 /zero
 /zero.exe
+/zero.exe~
 /zero-windows-command-runner.exe
 /zero-windows-sandbox-setup.exe
 /zero-skills
--- a/internal/aimlapi/client.go
+++ b/internal/aimlapi/client.go
@@ -75,6 +75,22 @@ type ExchangeResult struct {
 	APIKeyID string `json:"apiKeyId"`
 }
 
+// TopUpByKeyRequest funds a partner-checkout session with the API key that owns
+// the account (Path A / env key), instead of an email session.
+type TopUpByKeyRequest struct {
+	SessionToken     string
+	AmountUSDMinor   int
+	PaymentSessionID string // idempotency: same id → same checkout, never a second charge
+	AutoTopUp        bool   // enroll the account in automatic top-up at checkout (card, saved off-session)
+	SuccessURL       string
+	CancelURL        string
+}
+
+// TopUpByKeyResult is the hosted checkout returned by TopUpByKey.
+type TopUpByKeyResult struct {
+	Checkout PaymentSession `json:"checkout"`
+}
+
 // APIError is a non-2xx HTTP response from the aimlapi.com API. Status carries the
 // code so callers can branch (e.g. 401 = bad key, 5xx = retryable).
 type APIError struct {
@@ -156,6 +172,41 @@ func (c *Client) Exchange(ctx context.Context, bearer string, sessionToken strin
 	var result ExchangeResult
 	err := c.request(ctx, http.MethodPost, strings.TrimRight(c.endpoints.AppBaseURL, "/")+"/v3/partner-checkout/sessions/"+urlPathEscape(sessionToken)+"/exchange", bearer, nil, &result)
 	return result, err
+}
+
+// TopUpByKey funds a partner-checkout session using the raw API key that owns the
+// account (Path A / env key), returning the hosted checkout. It binds the session
+// to the key's account server-side, so no email session is needed and no key is
+// exchanged. req.PaymentSessionID makes a retry idempotent (same id → same
+// checkout, never a second charge).
+func (c *Client) TopUpByKey(ctx context.Context, apiKey string, req TopUpByKeyRequest) (TopUpByKeyResult, error) {
+	body := map[string]any{
+		"sessionToken":     req.SessionToken,
+		"amountUsdMinor":   req.AmountUSDMinor,
+		"paymentSessionId": req.PaymentSessionID,
+	}
+	if req.AutoTopUp {
+		body["autoTopUp"] = true
+	}
+	if strings.TrimSpace(req.SuccessURL) != "" {
+		body["successUrl"] = strings.TrimSpace(req.SuccessURL)
+	}
+	if strings.TrimSpace(req.CancelURL) != "" {
+		body["cancelUrl"] = strings.TrimSpace(req.CancelURL)
+	}
+	var result TopUpByKeyResult
+	err := c.request(ctx, http.MethodPost, c.billingV2URL("/billing/topup"), apiKey, body, &result)
+	return result, err
+}
+
+// billingV2URL builds a v2 billing API URL from the inference base. The top-up
+// endpoint lives next to the balance endpoint on the API gateway, one version up
+// (".../v1" → ".../v2/billing/..."), so it re-anchors the version while keeping any
+// host/prefix an override supplied.
+func (c *Client) billingV2URL(path string) string {
+	base := strings.TrimRight(strings.TrimSpace(c.endpoints.InferenceBaseURL), "/")
+	base = strings.TrimSuffix(base, "/v1")
+	return strings.TrimRight(base, "/") + "/v2" + path
 }
 
 func (c *Client) request(ctx context.Context, method string, endpoint string, bearer string, body any, out any) error {
--- a/internal/aimlapi/client.go
+++ b/internal/aimlapi/client.go
@@ -1,0 +1,182 @@
+package aimlapi
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+type PaymentMethod string
+
+const (
+	PaymentMethodCard   PaymentMethod = "card"
+	PaymentMethodCrypto PaymentMethod = "crypto"
+)
+
+type SessionStatus string
+
+const (
+	SessionStatusPendingAuth    SessionStatus = "pending_auth"
+	SessionStatusPendingPayment SessionStatus = "pending_payment"
+	SessionStatusPaid           SessionStatus = "paid"
+	SessionStatusExchanging     SessionStatus = "exchanging"
+	SessionStatusExchanged      SessionStatus = "exchanged"
+	SessionStatusCancelled      SessionStatus = "cancelled"
+	SessionStatusExpired        SessionStatus = "expired"
+	SessionStatusFailed         SessionStatus = "failed"
+)
+
+type AuthResult struct {
+	Token string `json:"token"`
+	Exp   int64  `json:"exp"`
+}
+
+type PartnerCheckoutSession struct {
+	ID             string        `json:"id"`
+	SessionToken   string        `json:"sessionToken"`
+	PartnerID      string        `json:"partnerId"`
+	PartnerName    *string       `json:"partnerName"`
+	UserID         *int64        `json:"userId"`
+	AmountUSDMinor *int64        `json:"amountUsdMinor"`
+	Status         SessionStatus `json:"status"`
+	IssuedKeyID    *string       `json:"issuedKeyId"`
+	ReturnURL      *string       `json:"returnUrl"`
+}
+
+type PaymentSession struct {
+	ProviderSessionID string `json:"providerSessionId"`
+	PayURL            string `json:"payUrl"`
+}
+
+type PayResult struct {
+	Checkout        PaymentSession         `json:"checkout"`
+	PartnerCheckout PartnerCheckoutSession `json:"partnerCheckout"`
+}
+
+type ExchangeResult struct {
+	APIKey   string `json:"apiKey"`
+	APIKeyID string `json:"apiKeyId"`
+}
+
+type APIError struct {
+	Message string
+	Status  int
+	Body    string
+}
+
+func (e APIError) Error() string {
+	if strings.TrimSpace(e.Body) != "" {
+		return fmt.Sprintf("%s: HTTP %d: %s", e.Message, e.Status, strings.TrimSpace(e.Body))
+	}
+	return fmt.Sprintf("%s: HTTP %d", e.Message, e.Status)
+}
+
+type Client struct {
+	endpoints  Endpoints
+	httpClient *http.Client
+}
+
+func NewClient(endpoints Endpoints, httpClient *http.Client) *Client {
+	if httpClient == nil {
+		httpClient = &http.Client{Timeout: 60 * time.Second}
+	}
+	return &Client{endpoints: endpoints, httpClient: httpClient}
+}
+
+func (c *Client) CreateSession(ctx context.Context, partnerID string, partnerName string, returnURL string) (PartnerCheckoutSession, error) {
+	body := map[string]any{"partnerId": partnerID}
+	if strings.TrimSpace(partnerName) != "" {
+		body["partnerName"] = strings.TrimSpace(partnerName)
+	}
+	if strings.TrimSpace(returnURL) != "" {
+		body["returnUrl"] = strings.TrimSpace(returnURL)
+	}
+	var result PartnerCheckoutSession
+	err := c.request(ctx, http.MethodPost, strings.TrimRight(c.endpoints.AppBaseURL, "/")+"/v3/partner-checkout/sessions", "", body, &result)
+	return result, err
+}
+
+func (c *Client) GetSession(ctx context.Context, sessionToken string) (PartnerCheckoutSession, error) {
+	var result PartnerCheckoutSession
+	err := c.request(ctx, http.MethodGet, strings.TrimRight(c.endpoints.AppBaseURL, "/")+"/v3/partner-checkout/sessions/"+urlPathEscape(sessionToken), "", nil, &result)
+	return result, err
+}
+
+func (c *Client) Pay(ctx context.Context, bearer string, sessionToken string, amountUSDMinor int, method PaymentMethod, successURL string, cancelURL string, autoTopUp bool) (PayResult, error) {
+	body := map[string]any{
+		"amountUsdMinor": amountUSDMinor,
+		"method":         method,
+	}
+	if strings.TrimSpace(successURL) != "" {
+		body["successUrl"] = strings.TrimSpace(successURL)
+	}
+	if strings.TrimSpace(cancelURL) != "" {
+		body["cancelUrl"] = strings.TrimSpace(cancelURL)
+	}
+	// Only sent when enabled: the field enrolls the account in automatic top-up.
+	// The gateway ValidationPipe does not whitelist, so it passes through today and
+	// is honoured once the backend adds support (see AIMLAPI-AUTOTOPUP-TZ.md).
+	if autoTopUp {
+		body["autoTopUp"] = true
+	}
+	var result PayResult
+	err := c.request(ctx, http.MethodPost, strings.TrimRight(c.endpoints.AppBaseURL, "/")+"/v3/partner-checkout/sessions/"+urlPathEscape(sessionToken)+"/pay", bearer, body, &result)
+	return result, err
+}
+
+func (c *Client) Exchange(ctx context.Context, bearer string, sessionToken string) (ExchangeResult, error) {
+	var result ExchangeResult
+	err := c.request(ctx, http.MethodPost, strings.TrimRight(c.endpoints.AppBaseURL, "/")+"/v3/partner-checkout/sessions/"+urlPathEscape(sessionToken)+"/exchange", bearer, nil, &result)
+	return result, err
+}
+
+func (c *Client) request(ctx context.Context, method string, endpoint string, bearer string, body any, out any) error {
+	var reader io.Reader
+	if body != nil {
+		data, err := json.Marshal(body)
+		if err != nil {
+			return err
+		}
+		reader = bytes.NewReader(data)
+	}
+	request, err := http.NewRequestWithContext(ctx, method, endpoint, reader)
+	if err != nil {
+		return err
+	}
+	request.Header.Set("Accept", "application/json")
+	if body != nil {
+		request.Header.Set("Content-Type", "application/json")
+	}
+	if strings.TrimSpace(bearer) != "" {
+		request.Header.Set("Authorization", "Bearer "+strings.TrimSpace(bearer))
+	}
+	response, err := c.httpClient.Do(request)
+	if err != nil {
+		return fmt.Errorf("network request to %s failed: %w", endpoint, err)
+	}
+	defer response.Body.Close()
+	text, readErr := io.ReadAll(io.LimitReader(response.Body, 1<<20))
+	if readErr != nil {
+		return readErr
+	}
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		return APIError{Message: method + " " + endpoint, Status: response.StatusCode, Body: string(text)}
+	}
+	if out == nil || len(text) == 0 {
+		return nil
+	}
+	if err := json.Unmarshal(text, out); err != nil {
+		return APIError{Message: method + " " + endpoint + " returned non-JSON body", Status: response.StatusCode, Body: string(text)}
+	}
+	return nil
+}
+
+func urlPathEscape(value string) string {
+	return url.PathEscape(value)
+}
--- a/internal/aimlapi/client.go
+++ b/internal/aimlapi/client.go
@@ -146,10 +146,11 @@ func (c *Client) GetSession(ctx context.Context, sessionToken string) (PartnerCh
 
 // Pay initiates payment for a session and returns the hosted checkout URL.
 // autoTopUp enrolls the account in automatic top-up at checkout time.
-func (c *Client) Pay(ctx context.Context, bearer string, sessionToken string, amountUSDMinor int, method PaymentMethod, successURL string, cancelURL string, autoTopUp bool) (PayResult, error) {
+func (c *Client) Pay(ctx context.Context, bearer string, sessionToken string, amountUSDMinor int, paymentSessionID string, method PaymentMethod, successURL string, cancelURL string, autoTopUp bool) (PayResult, error) {
 	body := map[string]any{
-		"amountUsdMinor": amountUSDMinor,
-		"method":         method,
+		"amountUsdMinor":   amountUSDMinor,
+		"paymentSessionId": paymentSessionID,
+		"method":           method,
 	}
 	if strings.TrimSpace(successURL) != "" {
 		body["successUrl"] = strings.TrimSpace(successURL)
--- a/internal/aimlapi/client.go
+++ b/internal/aimlapi/client.go
@@ -242,8 +242,11 @@ func (c *Client) request(ctx context.Context, method string, endpoint string, be
 	if response.StatusCode < 200 || response.StatusCode >= 300 {
 		return APIError{Message: method + " " + endpoint, Status: response.StatusCode, Body: string(text)}
 	}
-	if out == nil || len(text) == 0 {
+	if out == nil {
 		return nil
+	}
+	if len(bytes.TrimSpace(text)) == 0 {
+		return APIError{Message: method + " " + endpoint + " returned empty body", Status: response.StatusCode}
 	}
 	if err := json.Unmarshal(text, out); err != nil {
 		return APIError{Message: method + " " + endpoint + " returned non-JSON body", Status: response.StatusCode, Body: string(text)}
--- a/internal/aimlapi/client.go
+++ b/internal/aimlapi/client.go
@@ -12,6 +12,7 @@ import (
 	"time"
 )
 
+// PaymentMethod selects how a partner-checkout top-up is paid for.
 type PaymentMethod string
 
 const (
@@ -19,6 +20,8 @@ const (
 	PaymentMethodCrypto PaymentMethod = "crypto"
 )
 
+// SessionStatus is the lifecycle state of a partner-checkout session, as reported
+// by the aimlapi.com backend while polling for payment.
 type SessionStatus string
 
 const (
@@ -32,11 +35,14 @@ const (
 	SessionStatusFailed         SessionStatus = "failed"
 )
 
+// AuthResult is the bearer token minted by a passwordless sign-in / sign-up.
 type AuthResult struct {
 	Token string `json:"token"`
 	Exp   int64  `json:"exp"`
 }
 
+// PartnerCheckoutSession is a partner-attributed top-up session; its SessionToken
+// addresses it on later pay/exchange/poll calls.
 type PartnerCheckoutSession struct {
 	ID             string        `json:"id"`
 	SessionToken   string        `json:"sessionToken"`
@@ -49,27 +55,35 @@ type PartnerCheckoutSession struct {
 	ReturnURL      *string       `json:"returnUrl"`
 }
 
+// PaymentSession is the hosted payment-provider checkout; PayURL is the page the
+// user opens to pay.
 type PaymentSession struct {
 	ProviderSessionID string `json:"providerSessionId"`
 	PayURL            string `json:"payUrl"`
 }
 
+// PayResult pairs the hosted checkout with the updated partner-checkout session
+// returned when a top-up payment is initiated.
 type PayResult struct {
 	Checkout        PaymentSession         `json:"checkout"`
 	PartnerCheckout PartnerCheckoutSession `json:"partnerCheckout"`
 }
 
+// ExchangeResult holds the API key minted when a paid session is exchanged.
 type ExchangeResult struct {
 	APIKey   string `json:"apiKey"`
 	APIKeyID string `json:"apiKeyId"`
 }
 
+// APIError is a non-2xx HTTP response from the aimlapi.com API. Status carries the
+// code so callers can branch (e.g. 401 = bad key, 5xx = retryable).
 type APIError struct {
 	Message string
 	Status  int
 	Body    string
 }
 
+// Error renders the method, endpoint, status, and (when present) the response body.
 func (e APIError) Error() string {
 	if strings.TrimSpace(e.Body) != "" {
 		return fmt.Sprintf("%s: HTTP %d: %s", e.Message, e.Status, strings.TrimSpace(e.Body))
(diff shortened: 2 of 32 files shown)


Vasanthdev2004 (changes requested):
I checked #655 in a fresh worktree and ran the build and touched tests myself. The title is accurate — this is a full guided-onboarding flow, not just a preset line: an `internal/aimlapi` HTTP client (auth, top-up, key-issuance) and a ~1000-line TUI onboard screen, +2584 against main. The outbound calls to aimlapi.com are the user-initiated top-up and key-issuance steps the feature advertises, not silent phone-home, and no API keys are committed. Build, vet, and the aimlapi/catalog/config tests pass; only the known pre-existing main failures remain.

The thing that blocks me: `catalog.go` bakes always-on partner-attribution headers into the aimlapi descriptor — `X-AIMLAPI-Partner-ID` (a specific partner id), `X-AIMLAPI-Integration-Repo`, `X-AIMLAPI-Integration-Version` — and `resolver.go` applies them to every inference request, with an in-code comment that they "attribute usage for the rebate." Crediting a specific partner for a rebate on every request from a community preset is a maintainer policy call — I'd want explicit sign-off or an opt-in before this lands. It's also not a superset of #402/#621: those are minimal preset entries, this is a much larger feature, so which (if any) aimlapi PR is canonical depends on whether you want the onboarding at all.

Requesting changes pending that policy decision.

Vasanthdev2004:
@gnanam1990 would value your read on this one. The blocker is the always-on partner-attribution headers baked into the catalog (X-AIMLAPI-Partner-ID etc., applied to every request — the comment says "for the rebate"). That's a policy call I think needs maintainer sign-off; would be good to have your take too.

Lookoff-AIMLAPI:
Hey! 
Just wanted to clarify — is the only remaining blocker the partner-attribution headers/policy decision, or are you also waiting because of the other AIMLAPI PRs? I'd really like to get this PR merged if possible.

The reason I'm asking is that this PR contains a much more complete onboarding experience, including authentication, top-up, API key creation, and the full guided flow, rather than just adding a preset.

Unfortunately, we weren't aware of the other AIMLAPI PRs at the time, and we haven't been able to get in touch with the original author despite trying. We'd be happy to coordinate or adapt if needed, but we'd really like to see this implementation move forward.

@Vasanthdev2004

jatmn:
@Vasanthdev2004 this would be considered official pr

Vasanthdev2004 (approved):
Re-checked #655 at the latest HEAD (79f60add) in a fresh worktree. Build and the aimlapi/catalog/config tests pass; the base URL, auth wiring, and curated model list are correct, and no credentials are hardcoded  the only baked-in identity is the partner ID, which is an account ID, not a secret.

The partner attribution is intentional for this arrangement: X-AIMLAPI-Partner-ID (catalog.go:309-313) rides on each request via the resolver-to-provider header path, the partner-checkout top-up routes to the agreed DefaultPartnerID (config.go:22, client.go:111), and the latest commit renders aimlapi as a standard recommended row alongside the other recommended providers. Treating this as an endorsed partner preset  approving.

Lookov:
Hi everyone! 
Just wondering if there's anyone else who should review or approve this PR before it's ready to be merged?

@jatmn @Vasanthdev2004

jatmn (changes requested):

## Findings

- [P1] Preserve bundled local-control helpers in the npm install
  `scripts/postinstall.mjs:277`
  Release archives always stage the `helpers/` directory, but the new postinstall path extracts an archive and copies only `zero` plus sandbox binaries. The npm wrapper then finds `agent-browser` and `tuistory` only under `packageRoot/node_modules/.bin`, while this package has neither those dependencies nor a copied helpers tree. Consequently every normal npm install loses browser/terminal local-control capabilities that the release archives and previous platform packages supplied. Copy the extracted helpers directory into the package (and point the wrapper at it), or install those helper packages as part of the npm distribution.

- [P1] Do not resolve plugin executables from the caller workspace
  `internal/plugins/activate.go:461`
  This change leaves `./bin/tool.sh` and other path-bearing relative commands unchanged even though the documented manifest format says relative plugin paths are resolved at activation. Tool execution prefers the caller workspace as its CWD, so a documented plugin command now either fails because that workspace has no matching path or runs a workspace-controlled executable under the plugin tool's declared permission. Keep resolving relative executable paths under `PluginDir` and restore coverage for both tool and hook commands.

- [P1] Bind a Path-A top-up to the pasted key's account
  `internal/tui/aimlapi_onboard.go:394`
  The existing-key path validates key A only with `GetBalance`, then asks the user to sign in with an arbitrary email and uses that account's session token for `Pay`. On completion the flow retains and saves key A (`applyTopup` replaces the key only for new accounts), so signing into account B and paying funds B while the UI reports a successful top-up for A. Verify the key belongs to the authenticated account before offering the top-up, or use a flow that funds the selected key's actual account.

- [P2] Carry the configured inference endpoint into the saved profile
  `internal/tui/onboarding.go:672`
  `StreamTopUp` returns the resolved `AIMLAPI_INFERENCE_URL`, but both onboarding hosts discard it and later save the catalog's production base URL. A staging/custom-endpoint user can complete registration and top-up against the override, then receive a profile that sends the newly issued key to `https://api.aimlapi.com/v1`. Preserve the onboarding result's base URL (and model where applicable) when constructing the setup/wizard profile.

- [P2] Cancel in-flight account and key requests when the flow is abandoned
  `internal/tui/aimlapi_onboard.go:164`
  Escape only cancels the top-up stream; balance, account, code, passwordless, and key-minting requests all use `context.Background()`. For example, a user can cancel while `CreateKey` is in flight, after which the server still mints a key but its result is discarded because the state has been removed. Give each pending operation a cancellable context and cancel it when the flow is abandoned.

- [P2] Do not present unsupported automatic top-up as enabled
  `internal/aimlapi/client.go:144`
  The onboarding UI defaults automatic top-up to on, but this code explicitly documents that the backend does not yet honor `autoTopUp`; the field merely passes through its validation layer. Users can therefore complete checkout believing recurring funding was enabled when it was silently ignored. Hide/disable this control until the backend contract supports it, or surface that it is unavailable.

- [P2] Restore durable credential publication
  `internal/credstore/credstore.go:250`
  Removing the temp-file `Sync` before rename weakens the only data-durability barrier for both the encrypted credential store and its encryption secret (`internal/securefile/securefile.go:199`). After a crash or power loss, the replacement name can survive without durable contents, losing stored provider keys or making them undecryptable. Restore the pre-rename file sync (and directory sync where supported).

- [P2] Correct the advertised npm/release platform support
  `docs/INSTALL.md:19`
  The changed docs say Windows arm64 is supported, but the release matrix publishes only `windows-x64` and `scripts/postinstall.mjs` explicitly skips Windows arm64 because no asset exists. A native Windows-on-ARM install completes without a binary and `zero` then cannot launch. Restore the x64-emulation wording or add and publish the arm64 artifact.

- [P2] Keep the documented Go requirement consistent with the module
  `README.md:9`
  The docs were changed to require Go 1.25+, while `go.mod` still declares `go 1.26.5`. A Go-1.25-only source install cannot build this module, particularly in an offline or locked-down environment where automatic toolchain download is unavailable. Restore the 1.26.5 requirement in the documentation or actually lower and validate the module requirement.

Lookoff-AIMLAPI:
The out-of-scope findings were already fixed on main while this was in review, so I reverted my duplicate commit for them rather than re-doing the work — which also clears the merge conflicts (every conflict was in exactly those files, and they now match main):

Plugin relative exec-path resolution → #627
securefile/credstore fsync before rename → #631
Go 1.25→1.26 docs → #635
npm platform packaging + Windows-arm64 wording + local-control helpers → #626
So this PR now effectively carries only the onboarding feature findings:

[P1] Path A top-up — dropped the pasted-key balance/top-up path, so it can't fund a different account than the key.
[P2] Base URL — resolved AIMLAPI_INFERENCE_URL now carried into the saved profile in both hosts.
[P2] Cancellation — account/key requests run on a cancellable context, cancelled on abandon.
[P2] autoTopUp — honored server-side now; updated the stale comment. 

@jatmn Ready for re-review, please check

Lookoff-AIMLAPI:
Updated the API key validation, and it should now meet all the current requirements.

@jatmn Could you please take another look at the latest changes and re-review the PR? Thanks!

jatmn (changes requested):

## Findings

- [P1] Preserve an ambiguous checkout session instead of retrying with a new one
  `internal/aimlapi/stream.go:53`
  `Pay` and `Exchange` are side-effecting requests, but their session token is retained only on the stack. If the server completes either operation and the response is lost, the flow reports a failure, drops that token, and the amount screen starts a brand-new session on retry. That can present the customer with a second checkout after the first payment succeeded, and an already-issued key cannot be recovered. Retain and resume/poll the original session after ambiguous failures (or use an idempotency contract) rather than discarding it.

- [P2] Cancel the embedded flow before returning to the provider manager
  `internal/tui/provider_wizard.go:793`
  The manager-mode `Esc` branch runs before the AIMLAPI key delegation and only changes the wizard step. It leaves `wizard.aimlapi` and its pending operation/top-up context alive, so exiting from a manager-launched flow can still complete `CreateKey` or checkout in the background and discard the result. Route this branch through the same reset/cancellation logic as normal AIMLAPI cancellation.

- [P2] Do not attach AIMLAPI attribution headers to an overridden endpoint
  `internal/tui/provider_wizard.go:1940`
  Guided onboarding copies the descriptor's partner headers unconditionally, including when `AIMLAPI_INFERENCE_URL` supplied a staging, proxy, or custom base URL. The catalog resolver deliberately applies those headers only for the catalog endpoint, so this path disagrees with the normal provider setup and sends partner attribution to an arbitrary override. Gate the copied headers on the resolved endpoint matching the catalog default.

- [P2] Make catalog header overrides case-insensitive
  `internal/cli/provider_setup.go:549`
  HTTP header names are case-insensitive, but the new merge keeps map keys verbatim. For example, `--header x-aimlapi-partner-id=...` leaves both that key and the catalog's `X-AIMLAPI-Partner-ID`; later request construction canonicalizes both while ranging a Go map, so either value can win. Canonicalize keys while merging and add a mixed-case override test.

- [P2] Reject non-finite top-up amounts before converting to cents
  `internal/aimlapi/topup.go:56`
  `strconv.ParseFloat("NaN", 64)` succeeds, and NaN bypasses each comparison in the min/max validation before being converted to an invalid integer minor-unit amount. The amount field accepts arbitrary text, so this reaches `Pay` as a malformed checkout request rather than producing the local validation error promised by the UI. Reject NaN and infinities before conversion.

- [P2] Do not label every verification failure as an incorrect code
  `internal/tui/aimlapi_onboard.go:544`
  While on the code screen, every error from `VerifySignInCode`—including a timeout, 429, 5xx, or cancelled request—is rewritten as “Code you've entered is incorrect.” This sends users down the wrong recovery path during service failures. Restrict that message to the backend's invalid-code response and surface other errors normally.

- [P2] Keep the documented pasted-key low-balance path reachable
  `internal/tui/aimlapi_onboard.go:183`
  The “I already have a key” path calls `GetBalance` but discards its `BalanceResult`, and successful validation always lands on “Everything is ready.” A valid key with `lowBalance: true` therefore can never reach the optional top-up chooser, contrary to the PR's stated onboarding behavior. Either restore a safe account-bound top-up path or correct the advertised behavior and UI copy.

- [P2] Preserve the normal environment-key setup path for AIMLAPI
  `internal/tui/provider_wizard.go:626`
  Both TUI routes divert AIMLAPI into guided onboarding before their normal credential step, even though the descriptor advertises `AIMLAPI_API_KEY`. The guided pasted-key screen has no environment-key branch and cannot advance blank, so a user who already configured only that environment variable must paste it again or create an account. Offer a “use `AIMLAPI_API_KEY`” option or retain the standard blank-credential env behavior.

Vasanthdev2004:
I took a second look at jatmn's P1 on stream.go and it checks out. The checkout session token from CreateSession (and the paid session token from pollUntilPaid) are stack-local in StreamTopUp, and Pay and Exchange are side-effecting. If a step fails after the user has paid — e.g. pollUntilPaid times out on a slow checkout — the paid session is lost and a retry calls CreateSession again, opening a fresh checkout and risking a double charge with the original payment orphaned. Worth fixing before merge. My earlier approve was on the partner-attribution question, not the checkout robustness — this is a separate code issue.

Lookoff-AIMLAPI:
Implemented the requested AIMLAPI follow-up fixes:

- Fixed the P1 double-charge/key-loss risk: checkout session tokens are retained and retries resume the original session instead of creating a new checkout. Pending payments are only polled, and in-flight/completed exchanges are never repeated.
- Added idempotent API-key-bound top-ups for existing low-balance keys.
- Preserved `APIKeyEnv`/stored credentials without persisting literal secrets.
- Preserved saved base URLs and catalog attribution headers during model discovery.
- Fixed cancellation, case-insensitive header merging, non-finite amount validation, and verification-error handling.

Tests and vet pass for the affected packages.

@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P0] Keep ambient credentials out of sandboxed commands
  `internal/sandbox/runner.go:377`
  The new default path passes `os.Environ()` directly to every sandboxed command, while this PR also removes the default credential-file deny list in `internal/sandbox/profile.go:76`. A prompt-approved command can now read and exfiltrate provider tokens, cloud credentials, and other parent-process secrets through its environment (and the default network policy permits egress); it can also read the common cloud credential directories that the prior profile denied. Restore secret scrubbing and the credential deny-read defaults (including their explicit allow-read carve-outs) before exposing the new sandbox behavior.

- [P1] Do not send a custom AIMLAPI profile's key to the production API
  `internal/tui/provider_wizard_discovery.go:167`
  `existingAimlapiConfiguration` accepts any saved profile named/catalogued `aimlapi`, including one configured with a proxy, staging, or private `BaseURL`, but this preflight validates its key using `aimlapiOnboardClient()` and therefore the global public endpoint. Merely selecting that existing profile transmits its credential to the wrong service and reports a valid custom-endpoint key as invalid; the low-balance flow similarly resets its endpoint before discovery. Either restrict this guided path to the canonical endpoint or construct the client/state from the persisted base URL.

- [P1] Keep macOS keychain secrets out of the process argument vector
  `internal/keyring/keyring.go:67`
  The macOS `Set` path now invokes `security add-generic-password ... -w <secret>`. Same-user processes can observe this argv while the command runs, exposing stored API keys and tokens. The base implementation deliberately used `security -i` with stdin for this reason; restore that non-argv transport (and its input validation).

- [P1] Fail closed when restoring a raced live daemon lock fails
  `internal/daemon/lock.go:88`
  After moving a lock aside, a live PID discovered in the moved file is restored with an ignored `os.Rename` error. If that restore fails (for example due to a Windows sharing violation), the canonical path remains absent and the acquisition loop can create a second lock while the live holder continues using the moved one. Propagate/fail closed on the restore failure rather than treating the lock as reclaimable; apply the same correction to the equivalent lock reclaimers changed in this PR.

- [P1] Restore unknown configuration-field diagnostics
  `internal/config/validate.go:33`
  This branch deletes `unknownfields.go`, removes both `unknownFieldIssues` calls, and deletes their coverage. Since `encoding/json.Unmarshal` ignores unknown fields, `zero config validate` now accepts typos such as `sandbox.blockUnixSocket` or `maxTurn` and silently runs with different defaults, including for security-relevant policy settings. Restore the recursive unknown-field checks and tests that are present on the PR base.

- [P1] Make Windows-on-ARM installation actually install an executable
  `docs/INSTALL.md:19`
  The changed documentation promises Windows arm64 support, but the release matrix only publishes `windows-x64` and `scripts/postinstall.mjs:111` exits successfully without downloading a binary for Windows arm64. Consequently `npm install -g @gitlawb/zero` succeeds on that platform but leaves `zero` unable to launch. Download the x64 fallback under emulation or document the platform as unsupported until an arm64 asset is released.

- [P2] Fold catalog and persisted custom headers by HTTP header name
  `internal/config/resolver.go:1040`
  The resolver copies catalog headers and then overlays persisted `CustomHeaders` with case-sensitive map keys. A user override written as `x-aimlapi-partner-id` therefore coexists with catalog `X-AIMLAPI-Partner-ID`; later `http.Header.Set` canonicalizes both while iterating the map, so either value can win nondeterministically. Reuse the case-insensitive merge used by `providers add` when applying a catalog descriptor.

- [P2] Generate a fresh by-key idempotency ID after a terminal checkout
  `internal/tui/aimlapi_onboard.go:651`
  A cancelled, expired, or failed by-key session emits an empty session token, which clears `resumeSessionToken` but retains `paymentSessionID`. The next attempt creates a new partner session while reusing the old idempotency ID, whose documented contract returns the original checkout; it can therefore return/reject the terminal checkout instead of allowing the user to pay again. Clear the payment ID whenever the stream reports that its session is dead, while retaining it for retries of a live session.

- [P2] Do not regress the documented source-build toolchain requirement
  `README.md:9`
  The changed source-install instructions say Go 1.25+ is sufficient, but `go.mod` requires Go 1.26.5. A Go-1.25-only machine cannot build the project without downloading a newer toolchain, which is particularly problematic in locked-down/offline environments. Restore the 1.26.5+ requirement in all changed installation docs, or lower and validate the module requirement.

- [P2] Rebase onto current main instead of carrying unrelated behavior rollbacks
  `internal/config/resolver.go:43`
  The PR's merge base predates current `main`, and the resulting full diff reverts several already-merged fixes unrelated to AIMLAPI, including the default/deep turn budgets (80/160 back to 50), MiniMax M3 vision metadata, and M2 think-tag handling. These regressions make the provider feature unsafe to merge as-is and will create conflict-prone resolution work. Rebase onto current `main`, retain the current behavior, and re-review the resolved diff.

Lookoff-AIMLAPI:
Thanks for the thorough review. I rebased the PR onto the latest main, which removed the unrelated regressions and picked up the fixes that had already landed there:
sandbox credential scrubbing
macOS keychain stdin transport
daemon lock restoration
unknown config-field validation
Windows ARM packaging
Go 1.26.5 documentation

I also addressed the three remaining PR-specific findings in 89d29f0:
Custom/proxy AIMLAPI credentials are now excluded from the production balance/top-up preflight.
Catalog and persisted HTTP headers are merged case-insensitively.
Terminal checkout sessions now clear the by-key idempotency ID, while live sessions retain it for safe retries.
Regression tests were added for all three cases. The relevant config, tui, aimlapi, and cli test suites pass.

@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Keep the resolved AIMLAPI endpoint on the existing-key setup paths
  `internal/tui/onboarding.go:463`
  With both `AIMLAPI_API_KEY` and `AIMLAPI_INFERENCE_URL` set, first-run setup skips the onboarding state that assigns `setup.baseURL`; `setupBaseURL` consequently returns empty and both model discovery and the saved profile fall back to `https://api.aimlapi.com/v1`. The same bare-env fallback in `existingAimlapiConfiguration` constructs a default-endpoint profile. This silently ignores the documented inference override and can send a staging/proxy credential to production. Seed these env-key paths from `aimlapi.ResolveEndpoints().InferenceBaseURL` before discovery and persistence.

- [P1] Do not probe an existing production key through an overridden endpoint
  `internal/tui/provider_wizard_discovery.go:96`
  The new preflight rejects profiles whose *stored* base URL is noncanonical, but it runs its authenticated `GetBalance` call through `aimlapiOnboardClient`, which instead uses the process-wide `AIMLAPI_INFERENCE_URL`. Thus setting that override to a proxy or staging host while selecting an otherwise canonical saved AIMLAPI profile transmits its production key to the override. Require the resolved onboarding endpoint to be canonical before offering this preflight, or build the client from the profile endpoint.

- [P1] Cancel an in-flight checkout before quitting first-run setup
  `internal/tui/onboarding.go:164`
  Ctrl+C returns `tea.Quit` directly while the AIMLAPI sub-flow may be polling a checkout under its independent 25-minute background context. If the user pays after quitting, that detached stream can exchange and receive the one-time key with no UI left to show or persist it; if the process exits first, the paid session has no retained token to recover. Call the sub-flow cancellation before quitting, as the flow already does on Esc.

- [P2] Drop terminal checkout state when polling observes a dead session
  `internal/aimlapi/topup.go:103`
  `pollUntilPaid` returns immediately for `cancelled`, `expired`, and `failed` sessions without reporting `OnSession("")`. The TUI therefore retains both the dead resume token and, for by-key top-ups, its idempotency ID. The user's next retry only re-reads that dead session and fails; only a second retry can create a fresh checkout. Report the terminal session as cleared at this point so the first retry starts a new payment intent.

- [P2] Cancel the existing-configuration balance request on abandonment
  `internal/tui/provider_wizard_discovery.go:171`
  The configured-profile probe creates a 12-second context inside its command and does not retain its cancel function. Esc/Left merely invalidates the generation and clears wizard state, leaving the authenticated request active until it returns or times out. Store and invoke a cancel function when abandoning/resetting this preflight, matching the embedded onboarding flow's cancellation behavior.

- [P2] Keep partner attribution consistent with the advertised override
  `internal/aimlapi/stream.go:60`
  `AIMLAPI_PARTNER_ID` changes the ID used to create checkout sessions, but profiles created by either TUI path retain the catalog's hard-coded `X-AIMLAPI-Partner-ID`. An operator using the documented environment override therefore funds under one partner while all later inference usage is attributed to another. Propagate the resolved partner ID into the saved header or remove the unsupported override.

Lookoff-AIMLAPI:
Fixed all six findings in 60b3ff2: endpoint handling, credential isolation, request cancellation, terminal checkout cleanup, and partner attribution. 

@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Keep the existing-key top-up on the endpoint that was validated
  `internal/tui/provider_wizard_discovery.go:205`
  The configured-profile preflight deliberately pins `GetBalance` to the saved (or catalog-default) endpoint so an `AIMLAPI_INFERENCE_URL` override cannot receive an existing production key.  On a low-balance result, however, this creates a fresh onboarding state that re-resolves that environment override; choosing top-up then passes the saved key to `StreamTopUpByKey`, which posts it to the override's billing endpoint.  Thus a normal saved key plus a staging/proxy override exposes the key to that host and can fund the wrong account. Carry the validated endpoint through the embedded flow (including model discovery), or reject this continuation when it differs.

- [P1] Do not render server-controlled error bodies without the active credentials redacted
  `internal/tui/aimlapi_onboard.go:544`
  `Client.request` includes up to 1 MiB of a non-2xx response body in `APIError.Error()`, while the key-validation, verification, and key-minting error paths pass that string through `redaction.ErrorMessage` with empty options.  A provider/proxy error that echoes the pasted key, verification code, or session bearer therefore puts the credential in the TUI, terminal scrollback, and screenshots.  Supply the active secret values to redaction for every onboarding operation (and avoid surfacing arbitrary response bodies where possible).

- [P2] Clear a rejected checkout session before telling the user to retry
  `internal/aimlapi/topup.go:90`
  A 4xx from `GetSession` is treated as terminal but returns without calling `OnSession("")`.  The TUI consequently retains the resume token and by-key idempotency ID; its next retry resolves that same rejected/removed session and fails again instead of creating a new checkout.  Clear the retained state for terminal session failures, or classify any recoverable 4xx explicitly before preserving it.

- [P2] Apply the partner-ID environment override when resolving catalog profiles
  `internal/config/resolver.go:1043`
  The CLI and TUI creation paths use `WithResolvedPartnerHeader`, but the normal config-resolution path copies the descriptor's hard-coded partner ID and overlays any persisted header on top.  A hand-authored or migrated `catalogId: "aimlapi"` profile therefore ignores `AIMLAPI_PARTNER_ID`, while the same user's checkout uses the override from `ResolvePartnerID`.  This splits checkout and inference attribution despite the stated common precedence contract; resolve the catalog header here as well while preserving an explicit user header override.

Lookoff-AIMLAPI:
Fixed the four follow-up findings in 414eade: preserved the validated endpoint through top-up and model discovery, secured onboarding error rendering, cleared rejected checkout sessions, and applied partner-ID overrides during config resolution. Added regression tests and pushed the changes.
@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Restore the MCP workspace/scope wiring removed from the base
  `internal/cli/serve.go:46`
  This head resolves `-C/--cwd` for the tool registry but no longer passes `WorkspaceRoot` (or `Scope`) to `mcp.Serve`. `ServeOptions.withDefaults` therefore makes `resources/list` and `resources/read` use the launch directory instead of the requested workspace, so `zero serve --mcp -C /project` can expose unrelated files and hide the project resources. The same rollback also removes the documented `--add-dir` parser and all scoped-root support, so existing `zero serve --mcp --add-dir /shared` configurations now fail with an unknown flag. Rebase onto current `main` and retain the scoped `ServeOptions`/registry plumbing and its coverage rather than merging these unrelated reversions.

- [P2] Clear a rejected resumed checkout before allowing another retry
  `internal/aimlapi/stream.go:160`
  When a retained checkout token is rejected or deleted, `GetSession` returns a terminal 4xx here and the function returns without calling `OnSession("")`. The TUI consequently retains the invalid resume token (and the by-key idempotency ID), so every later retry re-reads the same dead session and fails before it can create a checkout. `pollUntilPaid` already clears terminal 4xx responses; apply the same cleanup in both resume helpers while preserving tokens for transport and 5xx failures.

- [P2] Do not reuse a payment intent after the user changes its amount
  `internal/tui/aimlapi_onboard.go:670`
  A failed by-key checkout returns to an editable amount field while retaining `paymentSessionID`. If the original payment request succeeded but its response was lost, a user can change (for example) `$25` to `$50` and retry; the next request sends `$50` with the old ID, for which `TopUpByKey` explicitly returns the original checkout. The UI can then present the new selection while the hosted checkout charges the old amount. Bind the idempotency ID to the amount/intent and discard it (and the resumable session as appropriate) on an amount change, or lock the amount while retrying an existing checkout.

- [P2] Reject an unrecognized account-check action instead of signing up
  `internal/tui/aimlapi_onboard.go:581`
  The account-check API documents only `sign-in` and `sign-up`, but every successful value other than the former is treated as permission to call `CreatePasswordlessAccount`. A malformed or newly introduced 2xx action can therefore create an account and take the user into a paid top-up flow without the intended eligibility decision. Switch explicitly over both accepted values and surface any other response as an error.

- [P2] Validate nonempty auth tokens before treating a 2xx response as success
  `internal/aimlapi/client_onboard.go:74`
  Unlike `CreateKey`, both token-producing client methods accept an empty successful response. `applyToken` uses an empty token as the sentinel for the separate “code sent” operation, so an empty `VerifySignInCode` result silently reopens code entry and an empty passwordless-signup result asks a new user for a code that was never sent. Reject empty tokens at the client boundary (or use a distinct message kind for the sentinel) so the user receives a retriable error instead.

- [P2] Display the cent-rounded amount that the checkout actually charges
  `internal/tui/aimlapi_onboard.go:790`
  `ParseAmountUSD` accepts more than two decimal places and rounds to the minor-unit amount sent to checkout, but the success screen renders the raw input. For example, `20.999` is charged as `$21.00` while the UI says `$20.999 has been added`. Either reject fractions of a cent or retain and render the normalized cents value in the confirmation.

Lookoff-AIMLAPI:
Rebased onto the latest main to restore the MCP workspace/scope and --add-dir wiring. Also fixed checkout retry cleanup, payment intent handling when the amount changes, account action validation, empty auth token validation, and cent-rounded amount display. Added regression coverage for these cases.

@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Make the email-session checkout retry idempotent
  `internal/aimlapi/stream.go:88`
  The email flow retains its partner-checkout token after a `Pay` transport failure, then a retry maps a still-`pending_auth` session back to `phasePay` and sends another `/pay` request. Unlike the by-key flow, this request has no persisted idempotency handle or other client-side deduplication. If the first request created a hosted checkout but its response was lost before the session is observed as `pending_payment`, the retry can create another payment intent/checkout. Carry a stable payment idempotency key through this flow as well, or make the backend's session-level idempotency an explicit, tested contract before treating `pending_auth` as safe to re-pay.

- [P2] Clear dead exchange-claim sessions before returning to the retry UI
  `internal/aimlapi/topup.go:128`
  `pollUntilExchangeSettled` does not receive `OnSession`, so its terminal 4xx and `cancelled`/`expired`/`failed` returns leave `resumeSessionToken` in the TUI. The next retry only re-reads that dead session in `resolveTopupSession`, clears it, and fails again; users need a second retry before a fresh payment can start. Thread the callback into this polling path and clear the token for terminal failures, as `pollUntilPaid` already does.

- [P2] Cancel the `/provider` AIMLAPI flow when Ctrl+C quits the app
  `internal/tui/model.go:1002`
  Ctrl+C is handled globally before the provider wizard, but the confirmed-exit path cancels only an agent run and calls `quit`; it never calls `providerWizard.resetAimlapiOnboard`. A wizard checkout runs on its own 25-minute context, so exiting during payment/exchange can leave it running and mint a one-time key after the UI that would display and persist it is gone. Cancel the embedded onboarding state during shutdown (or while arming the quit) just as the first-run flow and wizard Esc path do.

- [P2] Remove catalog attribution when an existing AIMLAPI profile is retargeted
  `internal/config/resolver.go:1043`
  A canonical AIMLAPI profile persists the partner and integration headers. If the user later changes that profile's `baseURL` through the provider editor, `applyCatalogDescriptor` skips adding catalog headers for the noncanonical URL but leaves the persisted ones untouched, and the runtime forwards them to the arbitrary proxy/staging host. Rebuild or strip catalog-owned headers when resolving a retargeted AIMLAPI profile while preserving genuinely user-supplied headers.

Lookoff-AIMLAPI:
Fixed all four findings: email checkout retries now use a stable payment idempotency ID, dead exchange sessions are cleared immediately, confirmed Ctrl+C cancels the provider onboarding flow, and AIMLAPI catalog attribution headers are removed from retargeted profiles. The user-facing flow remains unchanged, and regression tests were added.

@jatmn Could you please re-review when you have a chance?

jatmn (changes requested):

## Findings

- [P1] Validate the checkout URL before opening it
  `internal/aimlapi/stream.go:101`
  Both top-up paths accept any non-empty `checkout.payUrl` and immediately pass it to `browser.OpenURL`, which delegates it to the OS URL handler. A malformed or compromised checkout response (including one reached through the supported endpoint overrides) can therefore invoke `file:`, a custom protocol, or another non-web handler rather than a payment page. Parse and require an absolute `https` checkout URL before displaying or opening it, in both the session and by-key paths.

- [P2] Do not treat an empty balance response as successful key validation
  `internal/aimlapi/client.go:245`
  `request` returns success for every empty 2xx body even when the caller supplied an output struct. Consequently, an empty `200`/`204` from `/billing/balance` yields a zero-valued `BalanceResult`; the pasted-key path then treats the key as valid and not low-balance, saves it, and skips the advertised funding prompt. Required typed responses should reject an empty body (or `GetBalance` should explicitly require its balance payload) so an incomplete gateway response cannot silently complete onboarding.

Lookoff-AIMLAPI:
Fixed both findings: checkout URLs are now validated as absolute HTTPS URLs before being displayed or opened in either top-up path, and typed API operations reject empty successful responses while bodyless operations continue to support empty 204 responses. Regression tests were added for both cases.
@jatmn Could you please re-review when you have a chance?

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1:
hello @StanAIML kindly rebase to main please and fix conflicts

Lookoff-AIMLAPI:
Hello @kevincodex1 !
Rebased onto the latest main and resolved the conflicts.

kevincodex1 (approved):
LGTM

PR #656, opened 2026-07-11
feat(tui): press up to edit queued messages

euxaristia:
## What

Pressing the up arrow while a message is queued pops it back into the composer for editing, instead of navigating input history. While a message is queued, the empty composer shows a placeholder hint: "Press up to edit queued messages".

## Why

Once a prompt was queued during an active run there was no way to take it back for editing; the only options were letting it send or discarding it with Esc. This mirrors the flow in other agent TUIs where up recalls the queued message.

## Changes

- Up arrow pops the queued message into the composer, taking priority over history recall. A draft already being typed is preserved below the queued text.
- Empty composer shows the edit hint as its placeholder while a message is queued, and reverts once popped.
- Bug fix: queueing a second prompt during the same run used to silently replace the first. It now stacks under it, so both send as one message.

## Testing

Four new tests in queued_message_test.go cover the pop, queued-above-draft ordering, hint visibility, and stacking. Full internal/tui suite passes.

## Summary by CodeRabbit

- **New Features**
  - Enabled editing queued follow-ups with the **Up arrow** by moving them into the composer.
  - When composing is in progress, queued text is placed above the existing draft.
  - Added a visual hint while a queued message is available, and it disappears after popping it into the editor.
  - If multiple follow-ups are queued in a row, new queued text is appended to the existing queued message.
- **Tests**
  - Added unit tests for queued-message edit, stacking, cursor placement, and hint behavior.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1749,6 +1749,11 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 			if next, ok := m.moveComposerVisualCursor(-1); ok {
 				return next, nil
 			}
+			// A queued message takes ↑ priority over history recall: pop it back
+			// into the composer for editing before it sends on the next turn.
+			if m.hasQueuedMessage() && m.pendingSpecReview == nil {
+				return m.popQueuedMessageForEdit(), nil
+			}
 			if m.historyRecallActive() {
 				return m.recallHistory(-1), nil
 			}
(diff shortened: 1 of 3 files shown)


anandh8x (approved):
Real UX fix, tight scope. Up-arrow pop + hint + stacking-second-prompt all land cleanly; the silent overwrite was a data-loss class worth fixing. CI green, CodeRabbit approved, tests cover the new behavior. Approving.

Vasanthdev2004 (commented):
I checked the up-arrow priority carefully: with a message queued, up pops it back into the composer (above any in-progress draft), and with an empty queue the existing history-recall still works. The new branch sits between composer cursor movement and history recall, and since queueing clears the composer there's no collision with either. The edit path routes through setComposerState/syncInputFromComposer, so multiline text and the draft are preserved correctly. Build and vet are clean, and the full internal/tui suite passes apart from the pre-existing TestAltScreenTranscriptScrollKeepsFooterFixed failure that also fails on main. The four new tests cover the pop, draft ordering, hint, and stacking.

One small thing before merge: the two new comments in queued_message.go attribute the behavior to a specific external project by name. Could you drop that and just describe the behavior on its own? Otherwise this looks good to me.

kevincodex1 (approved):
lgtm

PR #658, opened 2026-07-11
fix(sandbox): use WRITE_RESTRICTED token when no DenyRead paths are configured

euxaristia:
## Problem

Since #612 removed the WRITE_RESTRICTED flag, the Windows sandbox token applies the restricted-SID check to reads as well as writes. Default Windows DACLs grant BUILTIN\Users, not any SID in the token's restricted list (the random capability SIDs, the logon SID, and Everyone). The result is that a wrapped command cannot open any executable or non-KnownDlls DLL anywhere on the system:

- The wrapped cmd.exe starts (its imports come from KnownDlls, and the unrestricted runner opens cmd.exe itself), but every child it spawns fails: gh, git, where.exe, even a nested cmd.exe, all exit 1 with zero bytes of output.
- A directly wrapped where.exe dies with STATUS_DLL_NOT_FOUND (0xC0000135).
- Because the failures are completely silent, the sandbox-denial keyword heuristics never fire, so no unsandboxed-retry prompt is offered and agent sessions strand on "Command completed with no output".

The env-gated smoke test TestWindowsRestrictedTokenNestedPipeCapture (spawns whoami.exe from inside the sandbox) reproduces this on current main.

## Fix

Restore WRITE_RESTRICTED, but only when the permission profile has no DenyRead paths. The flag makes the kernel skip restricted-SID deny ACEs for reads, which is exactly the DenyRead bypass #612 fixed, so profiles that configure DenyRead keep the fully restricted token and trade spawn capability for read-deny enforcement. DenyRead is empty by default, so the common case gets a working sandbox back while the #612 guarantee holds for the profiles that rely on it.

## Verification (Windows 11, real smoke tests)

With ZERO_SANDBOX_REAL_SMOKE=1 and the runner built from this branch:

- TestWindowsRestrictedTokenNestedPipeCapture passes (fails against a runner built from main).
- TestWindowsUnelevatedRealSandboxSmoke passes, including the DenyRead assertion added in #612 (that profile configures DenyRead, so it exercises the fully restricted branch).
- Manual check through the runner: `cmd /d /c "git --version & gh --version"` under an unelevated restricted-token profile prints both versions and exits 0; on main it exits 1 with no output.

`go build ./...` and `go test ./internal/sandbox ./internal/tools ./internal/agent` pass.

## Relation to #640

#640 attacks the same regression by adding Users and Authenticated Users to the restricted-SID list. That restores executable reads, but it also widens the write jail to every location where those groups hold write access, and it still blocks reads of per-user files (profile directories grant the individual user, not the Users group), so gh can spawn but cannot read its own config. This PR keeps the write jail scoped to the capability SIDs and restores all reads the user can normally perform. Happy to close one in favor of the other.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows sandbox filesystem access enforcement based on configured read restrictions.
  - Read-denied paths now receive more consistent protection while preserving appropriate write restrictions.
  - Enhanced handling of restricted permissions for improved sandbox security.

### DIFF
--- a/internal/sandbox/windows_command_runner_windows.go
+++ b/internal/sandbox/windows_command_runner_windows.go
@@ -69,7 +69,13 @@ func runWindowsSandboxCommand(config WindowsSandboxCommandConfig, stderr io.Writ
 	// this has no in-token fix; preflight blocking and output hints live in
 	// internal/tools/shell_runtime.go.
 	tokenSIDs := windowsRuntimeTokenSIDs(capabilitySIDs, offlineSID, config.PermissionProfile.Network.Mode)
-	token, err := createWindowsRestrictedTokenForCapabilitySIDs(tokenSIDs)
+	// A WRITE_RESTRICTED token keeps reads unrestricted so sandboxed commands
+	// can actually launch executables; it is only unsafe when DenyRead paths
+	// are configured, because the kernel skips restricted-SID deny ACEs for
+	// reads under that flag (#612). Profiles with DenyRead keep the fully
+	// restricted token, trading spawn capability for read-deny enforcement.
+	writeRestricted := len(config.PermissionProfile.FileSystem.DenyRead) == 0
+	token, err := createWindowsRestrictedTokenForCapabilitySIDs(tokenSIDs, writeRestricted)
 	if err != nil {
 		fmt.Fprintln(stderr, WindowsSandboxCommandRunnerName+": "+err.Error())
 		return 1
--- a/internal/sandbox/windows_token_windows.go
+++ b/internal/sandbox/windows_token_windows.go
@@ -48,7 +48,7 @@ func (sid windowsLocalSID) close() {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (approved):
I read through the token logic in internal/sandbox/windows_token_windows.go and the call site in windows_command_runner_windows.go. The gating is right: WRITE_RESTRICTED is only set when DenyRead is empty, so profiles that rely on read-deny keep the strict post-#612 token, and the flag never relaxes write checks — it only stops the restricted-SID check for reads, so the workspace write-jail stays intact. Build, vet, gofmt, and the sandbox package tests are clean on Windows and the branch merges cleanly. Approving.

One note: TestWindowsUnelevatedRealSandboxSmoke and TestWindowsRestrictedTokenNestedPipeCapture both stay skipped without ZERO_SANDBOX_REAL_SMOKE=1, so neither the spawn-capability fix nor the read-deny path is exercised in default CI — worth a manual smoke run before this lands.

kevincodex1 (approved):

PR #659, opened 2026-07-11
fix(tools): classify silent wrapped Windows command failures as sandbox denials

euxaristia:
## Problem

A Windows restricted-token sandbox failure is often completely silent. When the token cannot open the target executable or one of its DLLs (see #658), or an MSYS runtime dies during init, the wrapped command exits nonzero having written zero bytes to stdout and stderr. The denial detection in internal/tools/sandbox_denial.go is keyword-based, so these failures are never marked as sandbox denials. The agent loop then never offers the unsandboxed-retry prompt, and the model is left retrying "Command completed with no output" until it gives up. This is exactly the failure mode that stranded a real session: six consecutive gh/git calls returned exit 1 with empty output and no diagnostic.

## Fix

Treat a wrapped command on the Windows restricted-token or elevated backend that exits nonzero with completely empty combined output as a likely sandbox denial (kind: sandbox). This routes the result into the existing escalation path, so the user gets the retry-unsandboxed prompt instead of a bare failure.

Trade-off: a command that legitimately fails silently (findstr with no match, for example) now costs one extra approval prompt. The inverse misclassification silently strands the whole session, so the trade goes to prompting. Other backends are unaffected: bwrap and seatbelt failures are not silent in this way, and the existing keyword and seccomp checks are unchanged.

## Verification

New unit tests cover the silent-failure classification, the exit-0 and non-Windows non-cases, and a failure that produced output. `go build ./...` and `go test ./internal/tools ./internal/agent ./internal/sandbox` pass.

Complementary to #658: that PR fixes the most common cause of these silent failures, this one makes any remaining ones (MSYS init deaths, Schannel-style limitations, DenyRead-configured profiles that keep the fully restricted token) recoverable instead of dead ends.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of sandbox-denied commands on wrapped Windows backends when commands fail without producing output.
  * Ensured silent failures are only classified as sandbox denials in the correct wrapped Windows scenarios; prevented false positives for successful runs, non-Windows backends, and commands with unrelated output.

### DIFF
--- a/internal/tools/sandbox_denial.go
+++ b/internal/tools/sandbox_denial.go
@@ -66,7 +66,34 @@ func sandboxDenialKind(plan zeroSandbox.CommandPlan, exitCode int, sections ...s
 	if plan.TargetBackend == zeroSandbox.BackendLinuxBwrap && exitCode == 128+31 {
 		return SandboxDenialKindSandbox, "seccomp"
 	}
+	// Windows restricted-token failures are often SILENT: a spawn the token
+	// cannot satisfy (executable or DLL unreadable under the restricted-SID
+	// check, MSYS runtime init death, loader status 0xC0000135/0xC0000142)
+	// produces a nonzero exit with no output at all, so none of the keyword
+	// heuristics above can ever fire. Treat "wrapped Windows command failed
+	// without writing a single byte" as a likely sandbox denial so the caller
+	// offers the unsandboxed-retry prompt instead of surfacing a bare
+	// "command completed with no output". A command that legitimately fails
+	// silently (findstr with no match, for one) costs one extra approval
+	// prompt; the inverse misclassification silently strands the whole
+	// session, so the trade goes to prompting.
+	if windowsWrappedBackend(plan.TargetBackend) && sectionsEmpty(sections...) {
+		return SandboxDenialKindSandbox, "no output from sandboxed command"
+	}
 	return "", ""
+}
+
+func windowsWrappedBackend(backend zeroSandbox.BackendName) bool {
+	return backend == zeroSandbox.BackendWindowsRestrictedToken || backend == zeroSandbox.BackendWindowsElevated
+}
+
+func sectionsEmpty(sections ...string) bool {
+	for _, section := range sections {
+		if strings.TrimSpace(section) != "" {
+			return false
+		}
+	}
+	return true
 }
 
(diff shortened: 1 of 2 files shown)


euxaristia:
Pushed 5d726af addressing the CodeRabbit suggestion: TestLikelySandboxDeniedDetectsSilentWindowsWrappedFailure is now parameterized over both windows-restricted-token and windows-elevated, so the elevated arm of windowsWrappedBackend has coverage too. No production code changes.

Vasanthdev2004 (approved):
Reviewed the silent-denial classification. The heuristic is tight: it only fires for wrapped commands on the Windows restricted-token/elevated backends with a nonzero exit AND completely empty stdout+stderr, and the exitCode==0 early return (sandbox_denial.go:55) keeps a real success from ever being mislabeled as a denial. A genuine silent sandbox denial now gets the unsandboxed-retry prompt instead of stranding the session, and the only over-classification cost is one extra approval prompt for legitimately silent-failing wrapped commands like a no-match findstr — which the comment owns explicitly. Tests cover the success, empty-output, other-backend, and has-output cases; build/vet/gofmt/tools tests pass and it merges clean.

One minor note: BackendWindowsElevated is bundled into windowsWrappedBackend, but an elevated-token command isn't really "denied" by the sandbox, so a silent failure there is less clearly a denial — at worst it yields a spurious retry prompt, low harm, but worth a glance if you want to narrow it to the restricted-token backend only.

euxaristia:
On the BackendWindowsElevated note: it's currently unreachable as a TargetBackend value. internal/sandbox/manager.go only constructs a Windows backend via nativeBackend(goos, BackendWindowsRestrictedToken, ...); there's no call site that produces one named BackendWindowsElevated. buildPlatformCommandPlan's switch also only has a case for BackendWindowsRestrictedToken, so anything else falls through to nativeSandboxUnavailableError before a command plan is ever built.

Given that, narrowing windowsWrappedBackend() to restricted-token only wouldn't change any live behavior, and it would break the pattern used elsewhere (manager.go, adapters.go, platformForBackendName) of treating the two Windows backends as a matched pair. Leaving it bundled and covering it with the test added in 5d726af seemed like the better trade: no behavior change now, and it stays correct if the elevated path ever gets wired up to produce real command plans.

kevincodex1 (approved):

PR #665, opened 2026-07-13
feat(providers): refresh MiniMax model and endpoint coverage

octo-patch:
Supersedes #611 with a clean history based on the latest `main` branch.

## Summary

- Add `MiniMax-M2.7` to the global and China provider model catalogs.
- Record the current context, reasoning, tool-use, and input modality metadata for `MiniMax-M3` and `MiniMax-M2.7`.
- Enable image input for `MiniMax-M3` and inline thinking parsing for M2.x models on OpenAI-compatible profiles.
- Document the global and China Anthropic-compatible presets and OpenAI-compatible profile configuration.
- Verify the final request URLs for both regions and both compatible protocols.

## Verification

- `go fmt ./...`
- `go test -p 1 ./...`
- `make build-all`
- `go vet ./...`
- `make lint`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --enable-only unused,ineffassign,staticcheck --new-from-rev=origin/main ./...`

## Summary by CodeRabbit

- **New Features**
  - Added support for MiniMax M3 image input.
  - Expanded MiniMax model listings with current capabilities, including context limits, reasoning, tool calling, and supported modalities.
  - Added support for MiniMax M2 and M2.7 thinking output and compatible streaming endpoints.
  - Improved handling for global and China-region MiniMax configurations.

- **Documentation**
  - Added setup instructions for MiniMax API keys, presets, and OpenAI-compatible custom profiles.
  - Documented region-specific base URLs and API key configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,12 +130,41 @@ export OPENAI_API_KEY=sk-...
 export ANTHROPIC_API_KEY=...
 export GEMINI_API_KEY=...
 export LONGCAT_API_KEY=...
+export MINIMAX_API_KEY=...
+export MINIMAXI_API_KEY=...
 ```
 
 To configure Meituan LongCat (LongCat-2.0) directly, run:
 
 ```bash
 zero providers setup longcat --set-active
+```
+
+MiniMax presets use the Anthropic-compatible endpoints for the global and China
+regions:
+
+```bash
+zero providers add minimax --set-active
+zero providers add minimaxi-cn --set-active
+```
+
+To use the OpenAI-compatible endpoints instead, add a custom compatible profile
+for the required region:
+
+```bash
+zero providers add custom-openai-compatible \
+  --name minimax-openai \
+  --model MiniMax-M3 \
+  --base-url https://api.minimax.io/v1 \
+  --api-key-env MINIMAX_API_KEY \
+  --set-active
+
+zero providers add custom-openai-compatible \
+  --name minimax-cn-openai \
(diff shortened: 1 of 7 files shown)


Vasanthdev2004 (approved):
Reviewed #665 against the prior #611 concerns. This cleanly supersedes #611: it leaves the MiniMax base URLs at /anthropic (the provider descriptor file is untouched), so there's no /v1 doubling, and the new endpoint test asserts the resolved request URL for both regions and both compatible protocols — exactly the regression test that was asked for. Build, tests on the touched packages, vet, and gofmt are all clean, and the merge applies cleanly.

I also checked for the attribution pattern from the earlier preset PR: no partner/rebate/affiliate header, no recommended-tier elevation of MiniMax, and no hardcoded signing key or official-source trust anchor — the only new constant is a docs URL used as the model Source field. #611 is closed and should stay closed in favor of this one. The missing linked approved issue is the same gap that was waived on #611; I'm not blocking on it for a model refresh of an existing provider.

One minor note: the new `minimax-m2` think-tag marker in the provider wiring also matches MiniMax-M2.1, which the catalog doesn't mark as a reasoning model — harmless in practice since stripping tags a model never emits is a no-op, but could be tightened to `minimax-m2.7` later if M2.1 is definitively non-reasoning.

kevincodex1 (approved):
looks great. thanks for contribution. please join discord

PR #667, opened 2026-07-13
fix(swarm): wait for job.Runs directly in scheduler skip test

euxaristia:
### Problem
TestSchedulerSkipsWhilePreviousRuns intermittently failed on CI (observed on macOS) with `job runs=1 skipped=1, want 2/1`.

fireIfIdle's Spawn call (which the launcher's recorded() count observes) and run's subsequent job.incRuns() are sequential but distinct steps in the same scheduler goroutine. The test polled only the launcher's recorded count via waitFor, then asserted job.Runs immediately afterward with no wait of its own. Under contention, the scheduler goroutine could be preempted between those two steps, so the test's very next read caught a stale Runs value.

### Solution
Wait for job.Runs to actually reach 2 before asserting on it, instead of relying on the launcher's recorded count as a proxy for it.

Verified with `go test ./internal/swarm/... -run TestSchedulerSkipsWhilePreviousRuns -count=50 -race`, all passing.

## Summary by CodeRabbit

* **Tests**
  * Improved scheduler test reliability by waiting for job execution counts to update before making assertions.
  * Added clarification around the sequence of scheduler events during test execution.

### DIFF
--- a/internal/swarm/scheduler_test.go
+++ b/internal/swarm/scheduler_test.go
@@ -85,6 +85,14 @@ func TestSchedulerSkipsWhilePreviousRuns(t *testing.T) {
 	waitFor(t, "first done", func() bool { return sw.Coordinator().Summarize().Done == 1 })
 	ticks <- time.Time{}
 	waitFor(t, "second spawn", func() bool { return len(l.recorded()) == 2 })
+	// fireIfIdle's spawn (what "second spawn" observes via the launcher) and
+	// run's subsequent job.incRuns() are sequential but distinct steps in the
+	// scheduler's goroutine; wait for Runs itself rather than assuming the
+	// launcher recording it means the job's counter is updated too.
+	waitFor(t, "second run recorded", func() bool {
+		j, ok := findJob(sched.List(), id)
+		return ok && j.Runs == 2
+	})
 
 	j, ok := findJob(sched.List(), id)
 	if !ok {


jatmn (commented):
Thanks for the contribution. I do not see any actionable issues from my review.

@Vasanthdev2004 LGTM

Vasanthdev2004 (approved):
Verified the flake fix  the diff adds a waitFor on job.Runs == 2 before asserting, instead of treating the launcher recording the spawn as a proxy for the run counter. That's the right fix for the race: the spawn and the subsequent incRuns() are sequential but distinct steps in the scheduler goroutine, so the old proxy read could catch a stale Runs value. Ran the test 50x under -race and it's green. Approving.

gnanam1990 (approved):
## Verdict: Approve

Test-only fix for a real CI flake (`job runs=1 skipped=1, want 2/1` on macOS). The added `waitFor` on `job.Runs == 2` is the right fix.

## What looks good

- Root cause is correct: in `run()`, `fireIfIdle()` (launcher spawn) and `job.incRuns()` are sequential but distinct steps in the scheduler goroutine, so `len(l.recorded()) == 2` is not a reliable proxy for `Runs == 2`.
- Matches the existing pattern in this test (already waits on `j.Skipped == 1` before asserting).
- Comment explains the race without being noisy.
- CI green on all three OSes; @Vasanthdev2004 verified 50× under `-race`.

**Local verification** (merged latest `main`, no conflicts):
- `make lint` — pass
- `go test -race -count=50 ./internal/swarm/... -run TestSchedulerSkipsWhilePreviousRuns` — pass (50/50)

## Nits

- No linked `issue-approved` parent issue per CONTRIBUTING — acceptable for a maintainer CI flake fix.

## Issues

None.

kevincodex1 (approved):

PR #670, opened 2026-07-14
fix(agent): prevent plan execution from stalling after update_plan

### LINKED ISSUE #666: Bug Zero getting stuck on a plan
### Version / branch / commit

Zero Version 0.3.0

### OS and environment

Linux Cachyos 

### Steps to reproduce

I submit a request to my AI—in this case, I'm using HY3—but after it performs the initial checks and creates a plan, it gets stuck on that first step for hours; I've left it running to test it, but it doesn't move past that point or complete the full set of plans.

### Expected behavior

The right thing to do is for him to finish it, and then for me to test if it worked.

### Actual behavior

It doesn't complete all the plans; it gets stuck on the first plan, as I showed in the photo.

Image

gnanam1990:
## Summary
Fixes an agent-loop failure where Zero could remain indefinitely on the first active plan step after `update_plan`.
When a model returned continuation-cue text or reasoning-only responses without productive tool calls, interactive runs could accept the response as complete or repeatedly consume turns without reporting a visible failure.

## Root cause
- The completion gate was applied only when `RequireCompletionSignal` was explicitly enabled.
- Interactive TUI runs therefore accepted some non-productive text turns while plan items remained unfinished.
- Reasoning-only responses reset the empty-turn protection, allowing repeated non-productive turns until the turn budget was exhausted.
- The plan panel continued displaying the last valid `update_plan` state, leaving the first step active.

## Fix
- Apply the completion gate whenever unfinished plan items remain.
- Treat completed reasoning-only turns as non-productive while a plan is pending.
- Preserve the existing bounded continuation-nudge behaviour.
- Propagate incomplete results to the TUI.
- Show a visible incomplete-run notice.
- Mark the active plan step failed instead of falsely completing remaining steps.
- Keep no-plan and ordinary completion behaviour unchanged.

## Tests
Added regression coverage for:
- pending-plan completion gating without `RequireCompletionSignal`,
- bounded handling of reasoning-only turns,
- multi-chunk reasoning within one provider turn counting as one non-productive turn,
- reasoning-only turn followed by productive tool call succeeding,
- non-colon mid-step text nudged while plan pending,
- bounded nudge then valid final answer on TUI path,
- cancellation during plan-pending run,
- repeated identical `update_plan` hitting max turns,
- TUI incomplete-state handling,
- active-step failure without completing pending steps.

## Verification
- `go fmt ./...`
- `go vet ./...`
- `go test ./...`
- `make build`

### Lint comparison (base `80c39aa` vs fix `6360909`)
`golangci-lint` with `--enable-only unused,ineffassign,staticcheck` reports **36 issues on both** base and fix branches. No new failures in changed production files (`internal/agent/loop.go`, `internal/agent/guardrails.go`, `internal/tui/plan_panel.go`, `internal/tui/model.go`). Pre-existing failures remain in unrelated packages (`internal/sandbox`, `internal/tui` helpers, `internal/cli`, etc.).

### Smoke test
Live HY3 smoke test **not performed** — no HY3/OpenGateway credentials configured locally (active provider is `xai`). Deterministic fake-provider tests are the primary reproduction.

Fixes #666

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plan-aware “completion gate” and runaway/no-output behavior so stalls with pending plan steps use bounded “continue” nudges and end with correct incomplete outcomes.
  * Runs that stop mid-plan are now reported as incomplete (not successful), with remaining plan steps updated accordingly.
  * Cancellation during plan-pending stalls now stops immediately with a canceled result.
  * Prevented silent looping on repeated identical plan updates by enforcing max-turn limits.
* **Reliability / UX**
  * Refined reasoning-only turn detection while a plan is pending.
  * The TUI now displays “Run incomplete” as a system notice during resume/rehydration, with appropriate transcript behavior.

### DIFF
--- a/internal/agent/completion_gate_test.go
+++ b/internal/agent/completion_gate_test.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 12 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve pending-plan incompleteness at the max-turns exit
  `internal/agent/loop.go:739`
  The new plan-aware gate applies to interactive runs even when `RequireCompletionSignal` is false, but both max-turns exits still set `Result.Incomplete` only when that option is true. A TUI run can therefore repeatedly make tool calls (including `update_plan` with an in-progress item) until `MaxTurns`, receive the forced final summary, and be treated as successful; `model.updateModel` then calls `completeRemaining` and marks every untouched step complete. This recreates the false-completion state the PR is fixing. Mark a max-turn cutoff incomplete when pending plan items remain as well, and assert the result/TUI state in the repeated-plan regression.

- [P2] Keep the incomplete-run event in resumed prompt context
  `internal/sessions/store.go:40`
  `EventRunIncomplete` is rendered on TUI resume, but the session prompt filter in `exec_session.go` does not retain this new event type. Consequently, the next TUI or `exec --resume` agent invocation receives the final assistant message but not the reason that its prior run stopped unfinished, so it can reason from an apparent successful completion. Include the new event in `promptContextEvents` and cover resumed/forked prompt construction.

- [P2] Persist the final answer before its incomplete notice
  `internal/tui/model.go:5184`
  Live rendering appends the assistant's final answer and then the `Run incomplete` notice, but persistence appends `EventRunIncomplete` before `EventMessage`. Session hydration replays events in stored order, so reopening the session reverses the visible chronology and presents the warning before the answer it qualifies. Store the assistant message first (or otherwise preserve the live ordering) and add an ordered replay test.

jatmn (changes requested):

## Findings

- [P1] Preserve pending-plan incompleteness at the max-turns exit
  `internal/agent/loop.go:739`
  The new plan-aware gate applies to interactive/TUI runs even when `RequireCompletionSignal` is false, but both max-turns exits still set `Result.Incomplete` only when that option is true. A TUI run can keep making tool calls (including `update_plan` with an in-progress item) until `MaxTurns`, receive the forced final summary, and be treated as successful; `model.updateModel` then calls `completeRemaining` and marks every untouched step complete. This recreates the false-completion state this PR is intended to prevent. Mark a max-turn cutoff incomplete when pending plan items remain as well, and assert the result/TUI state in the repeated-plan regression.

- [P1] Derive the completion gate from a successfully applied plan
  `internal/agent/guardrails.go:541`
  `observeTurn` updates `planItemsPending` from raw `update_plan` arguments before the tool is executed. For example, with an existing pending TUI plan, a call such as `{"plan":[{"status":"completed"}]}` clears the guard count, but `update_plan` rejects the missing `content` and leaves the actual plan unchanged. The next text turn bypasses the new interactive completion gate and `completeRemaining` falsely completes that still-pending panel. This also diverges from the dispatcher’s concatenated-JSON recovery path. Only replace guard plan state after a successful, normalized update (or retain the prior state on a failed call), and cover the failed-update regression.

- [P2] Keep the incomplete-run event in resumed prompt context
  `internal/sessions/exec_session.go:194`
  `EventRunIncomplete` is rendered on TUI resume, but the session prompt filter does not retain the new event type. Consequently, the next TUI or `exec --resume` invocation receives the final assistant message but not the reason that its prior run stopped unfinished, so it can reason from an apparent successful completion. Include the event in `promptContextEvents` and cover resumed/forked prompt construction.

- [P2] Persist the final answer before its incomplete notice
  `internal/tui/model.go:5184`
  Live rendering appends the assistant's final answer and then the `Run incomplete` notice, but persistence appends `EventRunIncomplete` before `EventMessage`. Session hydration replays events in stored order, so reopening the session reverses the visible chronology and presents the warning before the answer it qualifies. Store the assistant message first (or otherwise preserve the live ordering) and add an ordered replay test.

Vasanthdev2004 (changes requested):
Requesting changes, though mostly to set direction before you spend more time on this head. #719 merged on main and extracted the completion decision into a typed policy (internal/agent/completion_policy.go), so this branch now conflicts — please rebase and express the interactive plan-pending gate through that policy (evaluate it whenever plan items are pending, not only under RequireCompletionSignal) rather than re-adding the inline nudge logic the refactor removed. The fix itself is still needed; #666 is still live on main after the refactor.

I also went through jatmn's review on this head and agree with all four findings, the two P1s especially: the max-turns exits around loop.go:739/748 still set Incomplete only under RequireCompletionSignal, so a TUI run that hits MaxTurns mid-plan gets its panel force-completed — the exact false completion this PR exists to prevent (your TestRunRepeatedIdenticalUpdatePlanHitsMaxTurns just needs to assert result.Incomplete to catch it). And observePlanUpdate counts from the raw update_plan args before the tool executes, so a call the tool rejects (e.g. missing content) still clears the gate. The reasoning-only turn handling and the failed-step panel treatment are right, and the test coverage is genuinely good — rebase onto the policy, fold in jatmn's four, and I'm ready to approve.

PR #672, opened 2026-07-14
fix(tui): stop the composer cursor blinking while typing or unfocused

euxaristia:
## Summary

- The composer's (prompt box's) text cursor blinked on a fixed 530ms timer unconditionally, including mid-keystroke, which made typing feel janky and disorienting since the cursor's job is to mark the current typing position.
- Gated the existing blink tick (`internal/tui/model.go`, `composerBlinkMsg` handler) on two states instead of an unconditional toggle:
  - **Actively typing** (a keystroke landed within the last 500ms, reusing the existing `lastCharTime` tracker): cursor stays solid.
  - **Unfocused** (new `terminalFocused` field, driven by `tea.FocusMsg`/`tea.BlurMsg`): cursor hides entirely, matching the convention in Vim/iTerm2/most editors that a hidden caret is the unambiguous "not active" signal.
  - **Idle and focused**: unchanged, blinks as before.
- Decoupled `view.ReportFocus` from the notifier (`m.notifier != nil` -> `true`). It was previously gated on notifier existence, which is effectively always true in production today, but left cursor-focus behavior fragile if that ever changes; focus reporting is a rendering concern independent of notification config.

## Linked issue

None. Direct user report about the blinking feeling janky while typing.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all green except the same pre-existing, unrelated `internal/contextreport` failure that also fails on unmodified `main`)
- [x] `gofmt -l` clean
- [x] New unit tests in `internal/tui/model_test.go`: cursor stays solid while typing, hidden while unfocused, resumes blinking after refocus + idle, unchanged toggle behavior when idle+focused
- [x] Manual smoke test: built the binary and drove it inside `tmux`, diffing raw `capture-pane -e` output over time to confirm the actual rendered cursor styling toggles while idle, stays solid across consecutive keystrokes sent at 200ms intervals, disappears immediately after injecting a terminal blur escape sequence (`CSI O`) and stays hidden across a blink interval, then resumes toggling after injecting a focus escape sequence (`CSI I`) and waiting past the idle threshold

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved composer cursor behavior: the caret hides when the terminal is unfocused, then resumes correctly on refocus.
  * Keeps the cursor solid while typing/pasting, and resumes blinking after an idle period.
  * Ensures caret rendering updates immediately on keypress/paste, including in the command-argument hint area.

* **Tests**
  * Added/expanded coverage for cursor visibility during typing, blur/focus transitions, idle blinking timing, immediate updates on keypress and paste, and rendering differences for the command-argument hint.

### DIFF
--- a/internal/tui/clipboard.go
+++ b/internal/tui/clipboard.go
@@ -63,6 +63,14 @@ func (m model) routePaste(content string) (tea.Model, tea.Cmd) {
 	// do it once here, before any early return, so both the bracketed-paste
 	// and right-click-paste paths are covered uniformly.
 	m = m.disarmCancelConfirmation()
+	// Same immediate-input transition as the keypress hook: text entering the
+	// composer must land the caret in its solid-while-typing state right away
+	// and refresh the typing timestamp. Without this, a paste right after the
+	// blink phase hid the caret leaves the newly populated composer caret-less
+	// until the next 530ms tick, and the stale timestamp makes that tick
+	// toggle instead of going solid.
+	m.lastCharTime = m.now()
+	m.composerCursorVisible = true
 	if content == "" {
 		// Empty text clipboard — the user may have pasted a screenshot.
 		// Probe the OS clipboard for image content asynchronously.
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1221,6 +1221,11 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if !keyIs(msg, tea.KeyEnter) {
 			m.lastCharTime = now
 		}
+		// Enter the solid-while-typing state right away: only composerBlinkMsg
+		// evaluates the typing threshold, so if the blink phase had just hidden
+		// the caret, the typed character would render caret-less for up to a
+		// full tick before the timer catches up.
+		m.composerCursorVisible = true
 		if m.setup.visible {
 			return m.handleSetupKey(msg)
 		}
@@ -1889,12 +1894,18 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, cmd
 	case tea.FocusMsg:
 		m.terminalFocused = true
+		// Sync the caret with focus immediately: leaving it to the next
+		// composerBlinkMsg tick can keep it hidden for up to a tick after the
+		// terminal regains focus (and, on blur below, leave it visible in an
+		// unfocused terminal for the same window).
+		m.composerCursorVisible = true
 		if m.notifier != nil {
 			m.notifier.SetFocused(true)
 		}
 		return m, nil
 	case tea.BlurMsg:
 		m.terminalFocused = false
+		m.composerCursorVisible = false
 		if m.notifier != nil {
 			m.notifier.SetFocused(false)
 		}
@@ -3499,7 +3510,7 @@ func (m model) composerLine(width int) string {
 	}
 	if argumentHint != "" {
 		input.SetWidth(0)
-		return fitStyledLine(commandArgumentHintComposerLine(input, argumentHint), width)
+		return fitStyledLine(commandArgumentHintComposerLine(input, argumentHint, m.composerCursorVisible), width)
 	}
 	previews := validComposerPastePreviews(state, m.composerPastePreviews)
 	displayState := composerDisplayStateForPastePreviews(state, previews)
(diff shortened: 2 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Apply the new caret state at the input and focus transitions
  `internal/tui/model.go:1221`
  A keypress only records `lastCharTime`, while focus and blur only update `terminalFocused`; `composerCursorVisible` is changed solely by the next 530 ms `composerBlinkMsg`. If an idle tick has hidden the caret, the next character is rendered without a caret until a later tick; a blur likewise leaves a visible caret, and a refocus can leave it hidden, for up to a tick. This contradicts the solid-while-typing and hidden-while-unfocused behavior, and the added tests miss it because they explicitly send a blink message after every transition. Update visibility synchronously for the relevant key/focus/blur messages and cover the state immediately after each message.

- [P2] Route required-command hints through the caret visibility state
  `internal/tui/model.go:3778`
  When a slash command is awaiting a required argument, `composerLine` takes the `commandArgumentHintComposerLine` branch, which always calls `composerCursor` and never reads `m.composerCursorVisible`. Consequently that composer caret remains visible while the terminal is unfocused and never observes the new typing/idle blink rules. Pass the visibility state into this renderer (and add an argument-hint case) so this alternate composer path follows the same focus contract.

euxaristia:
Pushed ee2e157 for both findings, plus CodeRabbit's outside-diff comment, which described the same defect as the first finding.

- Keypress, FocusMsg, and BlurMsg now set composerCursorVisible in their own handlers instead of waiting for the next composerBlinkMsg tick. That closes all three windows: a typed character rendering caret-less after the blink phase hid it, a blur leaving a visible caret in an unfocused terminal, and a refocus keeping the caret hidden for up to a tick.
- commandArgumentHintComposerLine now takes the caret visibility state instead of painting its cursor cell unconditionally, so the required-argument hint path follows the same focus and blink contract as the main composer renderer.
- The new tests (TestComposerCursorShowsImmediatelyOnKeypress, TestComposerCursorSyncsImmediatelyOnFocusChange, TestCommandArgumentHintFollowsCursorVisibility) assert the state immediately after each message with no blink tick in between, which is the coverage gap the review pointed out in the earlier tests.

go build, go vet, and the tui test suite pass locally, except TestAltScreenTranscriptScrollKeepsFooterFixed, which fails on this branch without these changes and passes on current main, so it is a stale-branch artifact rather than a regression from this change.

jatmn (changes requested):

## Findings

- [P2] Apply the caret state when pasting into the composer
  `internal/tui/clipboard.go:58`
  `tea.PasteMsg` and right-click `clipboardReadMsg` both reach `routePaste`, which updates the editable composer without refreshing `lastCharTime` or `composerCursorVisible`. If the idle blink has just hidden the caret, pasting text leaves that newly populated composer caret-less until the next 530 ms tick; because the timestamp is stale, that tick toggles instead of entering the solid-while-typing state. This is the same immediate-input transition fixed for `tea.KeyPressMsg`, so refresh the caret/timing state for the composer-paste branch and cover a paste immediately after a hidden blink.

euxaristia:
Pushed a fix for the paste finding: routePaste now refreshes lastCharTime and sets the caret visible before any early return, covering both the bracketed tea.PasteMsg path and the right-click clipboardReadMsg path, so a paste immediately after a hidden blink renders a solid caret and the next tick holds solid instead of toggling off the stale idle timestamp. TestComposerCursorShowsImmediatelyOnPaste covers the hidden-blink-then-paste sequence with no blink tick in between.

Vasanthdev2004 (approved):
Approve. The three-state blink handler is clean  unfocused hides, active typing stays solid via the 500ms idle threshold, and idle+focused toggles as before  and syncing the caret immediately on keypress/paste/focus/blur closes the race where a tick could leave it in the wrong state for up to a full blink period. One thing worth a glance: ReportFocus is now always on (CSI ?1004h) regardless of notifier config, so terminals that previously never received focus events will now start sending them; it's a standard, widely-ignored DEC mode and the true default handles terminals that never emit events, but it does broaden the surface. New tests cover every branch and pass; the only failing test (TestHandleAddDirCommand) reproduces on clean origin/main and is unrelated.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #673, opened 2026-07-14
feat(cli): show ZERO wordmark on --version

euxaristia:
## Summary
- Exposes the TUI's empty-state ZERO ascii wordmark via a new `tui.Wordmark` helper and prints it above the version line for `zero -v`/`--version`, as an easter egg.
- Coloring only applies when stdout is a real TTY and `NO_COLOR` isn't set, so redirected or piped output (e.g. scripts parsing the version) stays plain text.
- Version line changed from `zero <version>` to `version: <version>` to read correctly under the wordmark.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cli/... ./internal/tui/...`
- [x] Manually verified `zero --version` shows colored wordmark in a TTY, plain text when redirected to a file, and stays plain with `NO_COLOR=1` set even in a TTY

## Summary by CodeRabbit

* **New Features**
  * Added a formatted ZERO wordmark to version output when displayed in an interactive terminal.

* **Bug Fixes**
  * Preserved the concise, machine-readable `zero <version>` format when version output is redirected or piped.
  * Ensured redirected version output contains no extra banner or formatting.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -339,7 +339,7 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 				return 0
 			}
 		}
-		if _, err := fmt.Fprintf(stdout, "zero %s\n", version); err != nil {
+		if _, err := fmt.Fprintf(stdout, "%s\n\nversion: %s\n", tui.Wordmark(stdout), version); err != nil {
 			return 1
 		}
 		return 0
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -42,6 +42,7 @@ import (
 	"github.com/Gitlawb/zero/internal/worktrees"
 	"github.com/Gitlawb/zero/internal/zerogit"
 	"github.com/Gitlawb/zero/internal/zeroruntime"
+	"github.com/charmbracelet/x/term"
 )
 
 var version = "dev"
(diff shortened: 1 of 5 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve the machine-readable non-TTY version contract
  `internal/cli/app.go:342`
  `Wordmark` removes ANSI escapes for a pipe or redirection, but it still emits six banner lines and replaces the sole `zero <version>` record with `version: <version>`. That breaks existing command substitutions and exact parsers even though the PR says redirected output remains suitable for scripts; in particular, the checked-in NPM wrapper smoke contract still requires `zero --version` to report `zero <package.json version>`. Keep the existing single-line output when stdout is not a terminal (or provide a separately opt-in banner) and add coverage for the piped/redirection path.

- [P2] Respect selected light themes when coloring the version wordmark
  `internal/tui/startup.go:149`
  The new command bypasses `newModel`, which is the only path that resolves `--theme`/`ZERO_THEME` and applies the selected palette. As a result, `zero --theme light --version` and `ZERO_THEME=light zero --version` render with the global dark palette (`ink #ececee`), not the selected light palette (`ink #22201a`); on a light terminal the banner is therefore largely unreadable. Resolve and apply the requested palette for this non-TUI renderer, or render it without color outside the TUI.

euxaristia:
Pushed 1cdd3d0 for both findings.

- P1: when stdout is not a terminal, --version prints exactly "zero <version>" again, one line, no banner, no ANSI. The banner only renders on a TTY. The TTY path also uses "zero <version>" as its version record instead of "version: <version>". smokeVersion in internal/release goes back to requiring the exact single-line output, which holds because the smoke check reads the binary through a pipe. Coverage added for both non-terminal writers (TestRunPrintsVersion) and a real redirected file (TestRunVersionRedirectedFile).
- P2: Wordmark now renders uncolored rather than trying to resolve a theme outside the TUI. This renderer never sees --theme/ZERO_THEME, so any palette choice could land on the wrong background; the terminal's default foreground is readable everywhere. TestWordmarkIsPlain locks that in.

go build, go vet, and go test on internal/cli, internal/tui, and internal/release pass locally, except TestAltScreenTranscriptScrollKeepsFooterFixed in internal/tui, which also fails on this branch without these changes and is unrelated.

jatmn (approved):
@Vasanthdev2004 LGTM

Vasanthdev2004 (approved):
Nice touch  the TTY/pipe split is exactly right: scripts, redirects, and the NPM wrapper keep the single `zero <version>` record, and the wordmark only renders on an interactive terminal. I like that Wordmark() reuses the existing art and stays uncolored, so it can't go unreadable on a light theme. Build and the new tests pass in an isolated worktree; the only red tests reproduce on clean main (sandbox/add-dir), so nothing here to block on.

kevincodex1 (approved):

Issue #680, opened 2026-07-14
Pinned lint and vulnerability commands cannot analyze the Go 1.26 module


PierrunoYT:
## Summary

The repository requires Go `1.26.5`, while the exact golangci-lint and govulncheck commands mandated by `AGENTS.md` switch to Go `1.25.12`. Both reject the project because its packages require Go 1.26.

## Reproduction

```sh
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --enable-only unused,ineffassign,staticcheck ./...
```

fails with:

```
the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.5)
```

```sh
go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...
```

fails while loading packages with `package requires newer Go version go1.26 (application built with go1.25)`.

## Relevant files

- `go.mod`: `go 1.26.5`
- `AGENTS.md`: pins golangci-lint `v2.12.2` and govulncheck `v1.3.0`

Observed at commit `6fc1220` on Windows/amd64 with automatic toolchain switching.

## Impact

Contributors following mandatory repository instructions cannot complete the required checks.

## Suggested fix

Pin tool versions built with/supporting Go 1.26, or provide repository-managed invocations using the required toolchain. Exercise the exact documented commands in CI.

PR #681, opened 2026-07-14
fix(sandbox): deny reads of Zero credential stores

PierrunoYT:
## Summary

- deny sandbox reads of Zero configuration, credential, encrypted-secret, and OAuth token stores
- honor ZERO_OAUTH_TOKENS_PATH and ZERO_MCP_OAUTH_TOKENS_PATH overrides, including adjacent .secret files
- preserve path normalization, existence filtering, deduplication, and explicit AllowRead opt-outs

## Root cause

The sandbox scrubbed credential environment variables but its read-deny profile only covered cloud-provider paths. Under a read-all posture, an agent could still read Zero's own credential and OAuth stores from disk.

Windows read denial remains tracked separately in #662.

Related to #675

## Validation

- `go test ./internal/sandbox -count=1`
- `go test ./... -count=1` with ANTHROPIC_API_KEY unset
- `go vet ./...`
- `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...` (Go 1.26.5): no vulnerabilities found
- `go fmt ./...`

## Known repository/environment limitations

- The pinned golangci-lint command ran with Go 1.26.5 and reported 35 pre-existing findings outside this patch; it reported none in the changed code.
- `make lint` uses a Unix shell assignment that does not run under this Windows shell.
- The race-enabled `make test` path cannot run because GCC is not installed; the complete non-race suite passed.

## Summary by CodeRabbit

* **Security**
  * Strengthened sandbox protections for credential-related reads (AWS, gcloud, Azure, Google application credentials, OAuth/MCP token stores, and Zero config).
  * Added separate “deny-if-exists” handling for credential-derived paths, using command-aware resolution (including override file paths) and improved deduping.
  * Landlock now rejects unsupported deny-read combinations.
* **Bug Fixes**
  * Sandbox skips missing credential baselines safely while preserving correct enforcement when paths exist.
  * Improved isolation when credential data appears after launch.
  * More reliable token secret/store atomic write behavior to avoid stale temp artifacts.
* **Tests**
  * Expanded coverage for new deny-if-exists and credential resolution semantics.

PR #682, opened 2026-07-14
fix(sandbox): scrub dynamic credential env vars

### LINKED ISSUE #676: Sandbox env scrub misses custom apiKeyEnv and dynamic OAuth client secrets
## Summary

`scrubSensitiveEnv` removes a static secret list plus `AuthEnvVars` from the built-in provider catalog. Zero also supports secret environment-variable names outside that catalog:

- arbitrary provider-profile `apiKeyEnv` values, such as `COMPANY_LLM_SECRET`
- dynamic `ZERO_OAUTH_<NAME>_CLIENT_SECRET` variables

Those values remain inherited by sandboxed commands.

## Relevant code

- `internal/sandbox/runner.go`: `scrubSensitiveEnv` derives keys only from a fixed list and `providercatalog.All()`.
- `internal/config/types.go`: `ProviderProfile.APIKeyEnv` accepts a configured variable name.
- `internal/config/resolver.go`: the configured `APIKeyEnv` is read from the process environment.
- `internal/oauth/providers.go`: `envKey(name, "CLIENT_SECRET")` constructs secret names dynamically.

Observed at commit `6fc1220`.

## Reproduction

1. Configure a custom provider with `"apiKeyEnv": "COMPANY_LLM_SECRET"`.
2. Start Zero with `COMPANY_LLM_SECRET=example-secret`.
3. Run a sandboxed command that prints it.
4. The value is present because it is absent from the catalog-derived scrub list.

## Suggested fix

Pass resolved sensitive env names from configuration into sandbox construction and scrub dynamic secret namespaces such as `ZERO_OAUTH_*_CLIENT_SECRET`. Add tests with a non-catalog `apiKeyEnv` and custom OAuth provider name.

PierrunoYT:
## Summary

- pass every resolved provider profile's custom `apiKeyEnv` into the sandbox engine
- scrub those config-derived names in Linux, macOS, and Windows sandbox command plans
- scrub case-insensitive `ZERO_OAUTH_<NAME>_CLIENT_SECRET` variables for dynamically named OAuth providers
- keep the existing catalog-derived and static secret scrubbing behavior centralized

## Root cause

The sandbox environment scrubber knew only a static secret list and the built-in provider catalog. Custom provider profiles can name arbitrary credential variables through `apiKeyEnv`, and OAuth provider names create client-secret variables dynamically, so neither category was represented in that fixed set.

Sandboxed commands could therefore inherit those credentials even though built-in provider secrets were removed.

Fixes #676

## Validation

- `go fmt ./...`
- `go vet ./...`
- `go test ./internal/sandbox ./internal/cli -count=1`
- repository-wide `go test ./... -count=1` completed successfully for the changed packages and all but unrelated Windows process-lifecycle tests
- `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...`: no vulnerabilities found
- pinned golangci-lint reported the same 35 pre-existing findings tracked by #680 and none in this patch

## Known baseline/environment limitations

- `TestLoadProviderCommandTimeout` takes about 10 seconds instead of its expected 5 seconds on this host.
- `TestManagerCheckRealGopls` fails because the installed `gopls` process exits with EOF.
- Several `internal/tools` process-termination and temporary-directory cleanup tests fail on Windows because child processes retain file handles.
- `make lint` uses a Unix shell assignment in `fmt-check` that does not run under this Windows shell.

## Summary by CodeRabbit

* **Security Improvements**
  * Sandbox commands now scrub additional provider credential environment variables using resolved configuration, consistently across Linux, macOS, and Windows.
  * Improved handling of dynamically named OAuth client-secret environment variables (case-insensitive `ZERO_OAUTH_*_CLIENT_SECRET` pattern).

* **Bug Fixes**
  * Fixed cases where sensitive configured keys were not removed in all execution paths, including degraded fallback scenarios.

* **Tests**
  * Added regression coverage to verify configured and pattern-based sensitive environment scrubbing behavior.

### DIFF
--- a/internal/sandbox/runner.go
+++ b/internal/sandbox/runner.go
@@ -257,8 +257,25 @@ func directCommandPlan(spec CommandSpec, backend Backend, policy Policy, workspa
 		Name:              spec.Name,
 		Args:              cloneStrings(spec.Args),
 		Dir:               spec.Dir,
-		Env:               cloneStrings(spec.Env),
+		Env:               directCommandEnv(spec),
 	}
+}
+
+// directCommandEnv scrubs sensitive credentials from the environment for a
+// direct (unwrapped) command plan: the platform sandbox backend is
+// unavailable, disabled, or not required, so this is the actual environment
+// the child process inherits. Without scrubbing here, config-derived
+// apiKeyEnv and dynamic OAuth client-secret variables would leak into
+// commands that fall back to this path (e.g. EnforcementDegraded).
+func directCommandEnv(spec CommandSpec) []string {
+	env := cloneStrings(spec.Env)
+	if spec.Env == nil {
+		// Match the wrapped-plan behavior: an unset spec.Env means "inherit the
+		// caller's environment," so scrub a snapshot of it rather than passing
+		// nil through to exec.Cmd, which would skip scrubbing entirely.
+		env = os.Environ()
+	}
+	return scrubSensitiveEnv(env, spec.sensitiveEnvKeys...)
 }
 
 func (engine *Engine) resolveCommandDir(dir string, policy Policy) (string, string, error) {
--- a/internal/sandbox/runner.go
+++ b/internal/sandbox/runner.go
@@ -1033,6 +1033,11 @@ func normalizeSensitiveEnvKeys(keys []string) []string {
 	seen := make(map[string]struct{}, len(keys))
 	for _, key := range keys {
 		key = strings.TrimSpace(key)
+		// A misconfigured value like "COMPANY_LLM_SECRET=..." would never
+		// match a real env key during scrubbing; keep only the name part.
+		if name, _, found := strings.Cut(key, "="); found {
+			key = strings.TrimSpace(name)
+		}
 		folded := strings.ToUpper(key)
 		if key == "" {
 			continue
--- a/internal/sandbox/runner.go
+++ b/internal/sandbox/runner.go
@@ -29,6 +29,9 @@ type CommandSpec struct {
 	Args []string
 	Dir  string
 	Env  []string
+	// sensitiveEnvKeys is populated by Engine from config-derived credential
+	// variable names and carried through SandboxManager to the platform plan.
+	sensitiveEnvKeys []string
 }
 
 type CommandPlan struct {
@@ -122,6 +125,7 @@ func (engine *Engine) BuildCommandPlan(spec CommandSpec) (CommandPlan, error) {
 		return CommandPlan{}, errors.New("sandbox command name is required")
 	}
 	spec.Dir = commandDir
+	spec.sensitiveEnvKeys = engine.sensitiveEnvKeys
 
 	backend := engine.backend
 	if backend.Name == "" {
@@ -207,7 +211,7 @@ func linuxSandboxHelperCommandPlan(execRequest SandboxExecutionRequest, policy P
 	if err != nil {
 		return CommandPlan{}, err
 	}
-	env := sandboxEnvironmentForCommand(spec.Env, policy, BackendLinuxBwrap, "")
+	env := sandboxEnvironmentForCommandWithSensitiveEnv(spec.Env, policy, BackendLinuxBwrap, "", spec.sensitiveEnvKeys)
 	planDir := spec.Dir
 	if helper.Dir != "" {
 		planDir = helper.Dir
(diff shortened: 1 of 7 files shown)


jatmn (changes requested):

## Findings

- [P1] Scrub credentials on degraded command plans too
  `internal/sandbox/runner.go:169`
  `BuildCommandPlan` attaches the configured keys to `spec`, but the supported unavailable-backend fallback returns `directCommandPlan`, which simply leaves `Env` nil. The normal bash tool creates a spec with nil `Env`, so `exec.Cmd` inherits `COMPANY_LLM_SECRET` and `ZERO_OAUTH_<NAME>_CLIENT_SECRET` unchanged whenever the native backend is unavailable (the intended `EnforcementDegraded` path covered by `TestBuildCommandPlanDegradesUnavailableFallback` and `TestUnavailableBackendsDegradeForTargetPlatforms`). That leaves the reported credential-exfiltration scenario open on fallback hosts; construct a scrubbed environment for direct/degraded plans as well and add an inherited-environment regression test.

anandh8x (approved):
Re-reviewing against commit `7b59fc83` (head). The `nil `spec.Env` snapshot fix (patch 3/3) closes a real leak: direct/unwrapped command plans were inheriting the parent env unscrubbed, defeating the whole point of the scrubber — a sandboxed command could see `OPENAI_API_KEY` etc. by going through the direct plan path. That was the P1 and the fix is correct (snapshot `os.Environ()` then run `scrubSensitiveEnv` before forking). The configurable `SensitiveEnvKeys` plumbing from CLI → engine is clean, and the case-insensitive `ZERO_OAUTH_*_CLIENT_SECRET` pattern via `strings.ToUpper` is the right shape for env vars.

LGTM. Minor follow-up (not blocking): the `isDynamicSensitiveEnvKey` length check rejects the empty-provider case `ZERO_OAUTH__CLIENT_SECRET` (double underscore) — a one-line test for that edge case would be cheap. Also worth documenting the dynamic-pattern list near the function so a future maintainer knows where to add a second pattern.

Cross-PR note: #682 sits between #681 and #685. Recommend rebasing #682 onto #681, then #685 onto #682, to resolve the `credentialDenyReadPaths` / `scrubSensitiveEnv` signature overlap cleanly.

gnanam1990 (commented):
Local review: built and ran `go test ./internal/cli ./internal/sandbox` on darwin/arm64 (cli 94s); all pass. The direct/degraded-plan env scrubbing (including the `spec.Env == nil` -> `os.Environ()` snapshot) is a genuine security fix. One integration note.

internal/sandbox/runner.go
```diff
@@ -967,7 +992,7 @@ func regexpQuoteMeta(value string) string {
 	return replacer.Replace(value)
 }
 
-func scrubSensitiveEnv(env []string) []string {
+func scrubSensitiveEnv(env []string, additionalKeys ...string) []string {
```
gnanam1990, line 995: [P2] Sibling-PR merge conflict on `scrubSensitiveEnv` in `runner.go`

This PR changes the signature to `scrubSensitiveEnv(env, additionalKeys ...string)` and adds `normalizeSensitiveEnvKeys`. #685 adds a hardcoded `ZERO_DAEMON_REMOTE_TOKEN_FILE` to the same function's literal list, and both touch `runner_test.go`'s `TestScrubSensitiveEnv`. Each merges into `main` cleanly on its own, but they conflict with each other; #681 also rewrites `credentialDenyReadPathsIn` in `profile.go`. Sequence the three sandbox credential PRs and rebase as needed.

jatmn (approved):
LGTM

Vasanthdev2004 (approved):
Looks good to me. This closes a real leak: config-derived apiKeyEnv names and dynamic ZERO_OAUTH_*_CLIENT_SECRET vars are now scrubbed on every plan path (linux/seatbelt/windows wrapped plus the degraded direct fallback that previously inherited the caller env unscrubbed). I checked out the branch and confirmed go build, go vet, and gofmt are clean on the touched files, and the new tests (TestScrubSensitiveEnv, TestEngineScrubsConfiguredSensitiveEnvKeys, TestBuildCommandPlanDegradedFallbackScrubsInheritedEnv, TestProviderSensitiveEnvKeys) pass. The bare ZERO_OAUTH_CLIENT_SECRET (no provider segment) the reviewer noted is preserved by the prefix/suffix length guard, but that's fine in practice: Zero's envKey always emits ZERO_OAUTH_<NAME>_CLIENT_SECRET with a name segment, so that exact bare name is never a credential Zero sets.

kevincodex1 (commented):
lgtm

PR #686, opened 2026-07-14
fix(cron): reserve job IDs atomically

### LINKED ISSUE #678: Concurrent cron Add calls can allocate the same job ID
## Summary

Cron job creation allocates an ID with a check-then-create sequence. Two processes adding jobs in the same second can both observe the same directory as absent, select the same ID, and write into the same job directory.

## Relevant code

- `internal/cron/store.go`: `Add` calls `allocID` and later `writeJob`.
- `allocID` uses `os.Stat` to find an available timestamp ID.
- `writeJob` uses `os.MkdirAll`, which does not reserve the ID exclusively.
- Both writers use `metadata.json.tmp`.

Observed at commit `6fc1220`.

## Failure modes

One job can replace another, writers can interfere through the shared temp name, or both callers can receive the same ID.

## Suggested fix

Reserve candidates atomically with `os.Mkdir(dir, 0700)`, retry only on `os.ErrExist`, and remove the reserved directory if initial metadata creation fails. Add a concurrent cross-store test with a frozen clock.

PierrunoYT:
## Summary

- reserve cron job IDs atomically with directory creation
- retry only genuine ID collisions and clean reservations when initial metadata writes fail
- cover concurrent stores with a frozen clock and reservation cleanup failures

## Root cause

`Store.Add` previously checked for an available ID before creating its directory. Concurrent stores could both observe the same timestamp-based ID as available and then write through the same temporary metadata path.

## Validation

- `go fmt ./...`
- `go vet ./...`
- `go test ./internal/cron -count=1`
- concurrent regression test repeated 20 times
- `govulncheck ./...` (no vulnerabilities)
- full `go test ./...` run: cron and all other packages pass except the existing Windows process-cleanup failures in `internal/tools`
- pinned golangci-lint: existing repository-wide findings tracked in #680; no finding in changed files
- `make lint`: blocked by the existing Windows-incompatible `fmt-check` shell command

Fixes #678

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job ID allocation to prevent duplicate IDs when jobs are added concurrently.
  * Ensured failed job creation does not leave behind incomplete reservations or leftover data.

### DIFF
--- a/internal/cron/store.go
+++ b/internal/cron/store.go
@@ -99,27 +99,35 @@ func (s *Store) Add(job Job) (Job, error) {
 		job.Status = StatusActive
 	}
 	job.CreatedAt = s.now().UTC()
-	id, err := s.allocID()
+	id, err := s.reserveID()
 	if err != nil {
 		return Job{}, err
 	}
 	job.ID = id
-	if err := s.writeJob(job); err != nil {
+	if err := s.writeReservedJob(job); err != nil {
 		return Job{}, err
 	}
 	return job, nil
 }
 
-func (s *Store) allocID() (string, error) {
+func (s *Store) reserveID() (string, error) {
+	if err := os.MkdirAll(s.root, 0o700); err != nil {
+		return "", fmt.Errorf("create cron store root: %w", err)
+	}
 	base := s.now().UTC().Format("20060102-150405")
 	for n := 0; n < 100; n++ {
 		id := base
 		if n > 0 {
 			id = fmt.Sprintf("%s-%d", base, n)
 		}
-		if _, err := os.Stat(filepath.Join(s.root, id)); errors.Is(err, os.ErrNotExist) {
+		err := os.Mkdir(s.jobDir(id), 0o700)
+		if err == nil {
 			return id, nil
 		}
+		if errors.Is(err, os.ErrExist) {
+			continue
+		}
+		return "", fmt.Errorf("reserve cron job id %q: %w", id, err)
 	}
 	return "", errors.New("could not allocate a unique cron job id")
 }
@@ -127,7 +135,7 @@ func (s *Store) allocID() (string, error) {
 func (s *Store) jobDir(id string) string { return filepath.Join(s.root, id) }
 
 // validID rejects ids that could escape the store root (path separators or
-// traversal). allocID-generated timestamp ids always pass; this guards
+// traversal). reserveID-generated timestamp ids always pass; this guards
 // externally-supplied ids (get/update/remove/append).
 func validID(id string) bool {
 	return id != "" && id != "." && id != ".." && !strings.ContainsAny(id, `/\`) && filepath.Base(id) == id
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P1] Coordinate removal with an in-flight reservation
  `internal/cron/store.go:102`
  `reserveID` makes the timestamp directory visible before the metadata is written, but `Remove` treats that empty directory as a committed job and deletes it under a lock that `Add` never takes. A concurrent `zero cron rm <timestamp-id>` can therefore report success after deleting the reservation; `writeJob` then calls `MkdirAll`, recreates the directory, and persists the job anyway. A subsequent add can also re-reserve that same timestamp while the original writer is still running, putting two writers back on the shared metadata path. Hold the same per-ID lock from reservation through initial persistence, or make removal reliably distinguish and coordinate with an uncommitted reservation, and cover the add/remove/add interleaving.

anandh8x (changes requested):
.

gnanam1990 (approved):
## Review

The atomic `Mkdir` reservation plus holding the per-ID lock from reservation through `writeReservedJob` closes the race @jatmn flagged on the first commit (`55d8c55`). `Remove` can no longer delete a bare reservation while `Add` is still persisting, and a second `Add` cannot re-enter the same id until the first writer commits or tears down.

**Lock lifecycle:** `reserveID` acquires `lockJob` before `Mkdir`, returns the lock held on success, and releases it on collision or non-`ErrExist` mkdir errors. `Add`'s `defer unlock()` covers both success and `writeReservedJob` failure (cleanup runs under the held lock; lock file lives outside the job dir).

**Tests:** `go test ./internal/cron -count=1 -race` passes locally, including the concurrent-add (32 goroutines), Add-vs-Remove, lock-serialization, and failed-reservation cleanup regressions.

Two minor follow-ups (non-blocking):
1. Rename the exhausted-id error from "allocate" → "reserve" (`store.go:149`).
2. Include `job.ID` in the reservation cleanup failure message (`store.go:184`).

**Process note:** Issue #678 still lacks the `issue-approved` label — maintainers should add it before merge per `CONTRIBUTING.md`.

Otherwise LGTM — ready to merge.

anandh8x (approved):
Re-reviewing against commit `ef7537b1`. The per-ID lock is now held from `reserveID` through `writeReservedJob` (deferred unlock in `Add`, returned unlock from `reserveID`, acquired via `lockJob` before `Mkdir`), which closes the P1 race I flagged. The two new regression tests cover both halves of the finding — concurrent Remove waiting on the lock, and a second `reserveID` serializing on the same lock rather than racing the `Mkdir` — and the existing 32-way concurrent add test still holds. Withdrawing my prior CHANGES_REQUESTED. LGTM.

jatmn (approved):
lgtm

Vasanthdev2004 (changes requested):
The atomic reservation is sound  holding the per-ID lock from Mkdir through the metadata.json commit closes the Remove-vs-Add resurrect race, and the rest of the cron suite is green under -race. But TestStoreAddReservesUniqueIDsAcrossConcurrentStores is flaky: the frozen clock (time.Unix(1000,0)) pins all 32 writers to the same base id, so they convoy through one per-ID lock at a time and, under load, the tail waiters blow past the 10s lock timeout (I reproduced it under CPU load with "timed out acquiring job lock for 19700101-001640" and "unique jobs = 27, want 32"). Either dial the concurrency down to ~8 or advance the injected clock per writer so they don't all collide on id 0.

Vasanthdev2004 (approved):
Withdrawing the request-changes: I re-ran `TestStoreAddReservesUniqueIDsAcrossConcurrentStores` 42 times under `GOMAXPROCS=2` + IO load, with and without `-race`, and all passed  I can't reproduce the flake I flagged earlier. The atomic reservation is sound (per-ID lock held from Mkdir through the metadata.json commit closes the Remove-vs-Add resurrect race), so approving. Non-blocking nit: the frozen clock pins all 32 writers to the same base id, so under heavy CI load a tail writer could in principle lose the `O_EXCL` race past the 10s lock timeout advancing the injected clock per writer (or dropping concurrency to ~8) would make the test load-robust, but it's not a correctness issue.

kevincodex1 (approved):

PR #690, opened 2026-07-14
fix(config): kill provider-command process tree via job object on Windows

### LINKED ISSUE #683: TestLoadProviderCommandTimeout waits for child completion on Windows
## Summary

`TestLoadProviderCommandTimeout` consistently takes about 10.2 seconds on Windows even though `LoadProviderCommand` has a 5-second timeout. The test allows up to 9 seconds on Windows, so it fails.

## Reproduction

```powershell
go test ./internal/config -run '^TestLoadProviderCommandTimeout$' -count=1 -v
```

Observed repeatedly:

```text
timeout returned after 10.2s, want roughly 5s
```

## Relevant code

- `internal/config/command.go`: the timer fires after five seconds, calls `terminateCommandProcess(cmd)`, then blocks on `<-done`.
- `internal/config/process_windows.go`: termination invokes `taskkill /T /F` and then `cmd.Process.Kill()`.
- `internal/config/command_test.go`: the fixture launches a ten-second PowerShell `Start-Sleep` child and expects completion within nine seconds on Windows.

## Expected behavior

The provider command and its descendants should be terminated promptly after the five-second timeout, and the call should return within the Windows allowance.

## Actual behavior

The call returns around the fixture's natural ten-second completion, suggesting the process-tree termination or wait path is not promptly releasing the command.

## Suggested fix

Make Windows provider-command termination reliably kill the complete process tree without waiting for the sleeping child to exit naturally. Add a regression assertion that the timeout returns close to five seconds and leaves no child process behind.

PierrunoYT:
## Summary

Fixes #683 — `TestLoadProviderCommandTimeout` could take ~10.2s on Windows because the 5s provider-command timeout didn't reliably kill the process tree.

`taskkill /T` walks the process tree by parent PID in user space, so it can miss descendants (and depends on `taskkill.exe` being runnable in the environment). When the sleeping PowerShell child survived, it kept the inherited stdout/stderr pipe write ends open, and `cmd.Wait()` blocked until the child completed naturally.

## Changes

- **Job object termination (Windows):** the process is started with `CREATE_SUSPENDED` and assigned to a job object before its main thread resumes, so a fast command can't spawn and detach a descendant before containment is in place. On timeout, `TerminateJobObject` kills every member of the tree atomically. Note: the job does **not** set `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` — releasing the handle on the success path intentionally leaves a provider command's detached helpers running, matching pre-job-object behavior; termination only happens explicitly via `Terminate()` on timeout/error. `taskkill /T /F` + `Process.Kill()` remain as a fallback for the rare case where job creation/assignment fails, and are skipped when a job successfully contained the tree (to avoid acting on a PID that `cmd.Wait()` may have already reaped and that Windows could have reused).
- **`cmd.WaitDelay = 1s`:** bounds pipe draining after process exit, so `Wait` can't hang on pipe handles held by an escaped orphan. `Terminate()` now runs for *any* error from `Wait` (not just `exec.ErrWaitDelay`), since a nonzero-exit shell (e.g. `sleep 600 & exit 7`) also needs its leaked background child cleaned up.
- **Resume-failure handling:** if the suspended main thread can't be resumed (Toolhelp/OpenThread/ResumeThread failure), the process is terminated immediately instead of being left suspended until the timeout expires and silently misreported as an ordinary slow command.
- **Regression tests:** the timeout fixture's sleeping child records its PID and the test asserts it's actually gone after `LoadProviderCommand` returns (Windows and POSIX), including a nonzero-exit background-child variant and a `resumeMainThread` failure-reporting test.

## Known limitation

If job creation or assignment itself fails, the `taskkill /T` fallback can still miss a descendant that has reparented away from the tree it walks. This is not a regression versus pre-job-object behavior (taskkill was the only mechanism then too) and is tracked as a follow-up rather than blocking this fix.

## Verification

- `go build ./...` — pass
- `go vet ./internal/config/...` — pass
- `go test ./internal/config/...` — pass, including the new nonzero-exit background-child test and the `resumeMainThread` failure-reporting test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider command timeout handling so timed-out commands are reliably terminated along with any detached background child processes.
  * Added bounded output-draining behavior after process exit for more predictable timeout completion.
  * Strengthened cross-platform process cleanup by preserving/isolating process-group/job identity and ensuring proper termination on both POSIX and Windows.
  * Added/updated tests to verify no leaked background processes and correct behavior on timeout and nonzero exits.

### DIFF
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -43,6 +43,9 @@ var errProviderCommandTimeout = errors.New("provider command timeout")
 func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, error) {
 	cmd := shellCommand(command)
 	configureCommandProcess(cmd)
+	// Bound the time Wait spends draining I/O pipes after the process exits,
+	// in case an orphaned descendant still holds the write ends open.
+	cmd.WaitDelay = time.Second
 
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
@@ -52,6 +55,9 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 	if err := cmd.Start(); err != nil {
 		return nil, nil, err
 	}
+
+	proc := attachCommandProcess(cmd)
+	defer proc.Close()
 
 	done := make(chan error, 1)
 	go func() {
@@ -65,7 +71,7 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 	case err := <-done:
 		return stdout.Bytes(), stderr.Bytes(), err
 	case <-timer.C:
-		terminateCommandProcess(cmd)
+		proc.Terminate()
 		<-done
 		return stdout.Bytes(), stderr.Bytes(), errProviderCommandTimeout
 	}
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -42,7 +42,6 @@ var errProviderCommandTimeout = errors.New("provider command timeout")
 
 func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, error) {
 	cmd := shellCommand(command)
-	configureCommandProcess(cmd)
 	// Bound the time Wait spends draining I/O pipes after the process exits,
 	// in case an orphaned descendant still holds the write ends open.
 	cmd.WaitDelay = time.Second
@@ -52,11 +51,10 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 
-	if err := cmd.Start(); err != nil {
+	proc, err := startCommandProcess(cmd)
+	if err != nil {
 		return nil, nil, err
 	}
-
-	proc := attachCommandProcess(cmd)
 	defer proc.Close()
 
 	done := make(chan error, 1)
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -69,6 +69,14 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 
 	select {
 	case err := <-done:
+		if errors.Is(err, exec.ErrWaitDelay) {
+			// The process exited but a background descendant kept the
+			// inherited stdout/stderr pipes open past WaitDelay. Treat this
+			// like a timeout so the tree is actually terminated instead of
+			// leaking that descendant.
+			proc.Terminate()
+			return stdout.Bytes(), stderr.Bytes(), errProviderCommandTimeout
+		}
 		return stdout.Bytes(), stderr.Bytes(), err
 	case <-timer.C:
 		proc.Terminate()
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -69,12 +69,16 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 
 	select {
 	case err := <-done:
-		if errors.Is(err, exec.ErrWaitDelay) {
-			// The process exited but a background descendant kept the
-			// inherited stdout/stderr pipes open past WaitDelay. Treat this
-			// like a timeout so the tree is actually terminated instead of
-			// leaking that descendant.
+		if err != nil {
+			// The shell exited (whether via ErrWaitDelay because a
+			// background descendant kept the inherited stdout/stderr pipes
+			// open, or via a nonzero exit status) but a leftover descendant
+			// may still be running. Terminate is a no-op against an
+			// already-dead tree, so always call it rather than gating on
+			// the specific error to avoid leaking that descendant.
 			proc.Terminate()
+		}
+		if errors.Is(err, exec.ErrWaitDelay) {
 			return stdout.Bytes(), stderr.Bytes(), errProviderCommandTimeout
 		}
 		return stdout.Bytes(), stderr.Bytes(), err
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -67,7 +68,8 @@ func TestLoadProviderCommandFailureIncludesExitAndRedactsOutput(t *testing.T) {
 }
 
 func TestLoadProviderCommandTimeout(t *testing.T) {
-	command := writeCommand(t, commandScript{SleepSeconds: 10})
+	pidFile := filepath.Join(t.TempDir(), "sleep.pid")
+	command := writeCommand(t, commandScript{SleepSeconds: 10, PidFile: pidFile})
 
 	start := time.Now()
 	_, err := LoadProviderCommand(command)
@@ -85,6 +87,29 @@ func TestLoadProviderCommandTimeout(t *testing.T) {
 	if elapsed > maxElapsed {
 		t.Fatalf("timeout returned after %s, want roughly 5s", elapsed)
 	}
+	assertProcessTerminated(t, pidFile)
+}
+
+func assertProcessTerminated(t *testing.T, pidFile string) {
+	t.Helper()
+
+	data, err := os.ReadFile(pidFile)
+	if err != nil {
+		t.Fatalf("read sleeper pid file: %v", err)
+	}
+	pid, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		t.Fatalf("parse sleeper pid %q: %v", data, err)
+	}
+
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if !processAlive(pid) {
+			return
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	t.Fatalf("sleeper process %d still alive after timeout", pid)
 }
 
 func TestLoadProviderCommandInvalidJSON(t *testing.T) {
@@ -118,6 +143,7 @@ type commandScript struct {
 	Stderr       string
 	ExitCode     int
 	SleepSeconds int
+	PidFile      string
 }
 
 func writeCommand(t *testing.T, script commandScript) string {
@@ -128,7 +154,11 @@ func writeCommand(t *testing.T, script commandScript) string {
 		path := filepath.Join(dir, "provider.cmd")
 		lines := []string{"@echo off"}
 		if script.SleepSeconds > 0 {
-			lines = append(lines, "powershell -NoProfile -Command \"Start-Sleep -Seconds "+itoa(script.SleepSeconds)+"\"")
+			sleep := "Start-Sleep -Seconds " + itoa(script.SleepSeconds)
+			if script.PidFile != "" {
+				sleep = "Set-Content -Path '" + script.PidFile + "' -Value $PID -Encoding Ascii; " + sleep
+			}
+			lines = append(lines, "powershell -NoProfile -Command \""+sleep+"\"")
 		}
 		if script.Stdout != "" {
 			lines = append(lines, "echo "+script.Stdout)
@@ -146,6 +176,9 @@ func writeCommand(t *testing.T, script commandScript) string {
 	path := filepath.Join(dir, "provider.sh")
 	lines := []string{"#!/bin/sh"}
 	if script.SleepSeconds > 0 {
+		if script.PidFile != "" {
+			lines = append(lines, "echo $$ > '"+script.PidFile+"'")
+		}
 		lines = append(lines, "sleep "+itoa(script.SleepSeconds))
 	}
 	if script.Stdout != "" {
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -232,8 +232,13 @@ func writeCommand(t *testing.T, script commandScript) string {
 			lines = append(lines, "echo "+script.Stderr+" 1>&2")
 		}
 		if script.BackgroundSleepSeconds > 0 {
-			bgSleep := "Set-Content -Path '" + psSingleQuote(script.BackgroundPidFile) + "' -Value $PID -Encoding Ascii; Start-Sleep -Seconds " + itoa(script.BackgroundSleepSeconds)
+			readyFile := script.BackgroundPidFile + ".ready"
+			bgSleep := "Set-Content -LiteralPath '" + psSingleQuote(script.BackgroundPidFile) + "' -Value $PID -Encoding Ascii; " +
+				"Set-Content -LiteralPath '" + psSingleQuote(readyFile) + "' -Value ready -Encoding Ascii; " +
+				"Start-Sleep -Seconds " + itoa(script.BackgroundSleepSeconds)
 			lines = append(lines, "start /B powershell -NoProfile -Command \""+bgSleep+"\"")
+			waitForReady := "while (-not (Test-Path -LiteralPath '" + psSingleQuote(readyFile) + "')) { Start-Sleep -Milliseconds 10 }"
+			lines = append(lines, "powershell -NoProfile -Command \""+waitForReady+"\"")
 		}
 		lines = append(lines, "exit /b "+itoa(script.ExitCode))
 		if err := os.WriteFile(path, []byte(strings.Join(lines, "\r\n")), 0o700); err != nil {
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -238,7 +238,7 @@ func writeCommand(t *testing.T, script commandScript) string {
 }
 
 // psSingleQuote escapes a value for interpolation inside a PowerShell
-// single-quoted string literal, where a literal quote is written as ''.
+// single-quoted string literal, where a literal quote is doubled.
 func psSingleQuote(value string) string {
 	return strings.ReplaceAll(value, "'", "''")
 }
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -90,6 +90,34 @@ func TestLoadProviderCommandTimeout(t *testing.T) {
 	assertProcessTerminated(t, pidFile)
 }
 
+// TestLoadProviderCommandTerminatesBackgroundChild covers a command that
+// exits immediately but leaves a detached child holding the inherited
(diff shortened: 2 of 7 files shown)


jatmn (changes requested):

## Findings

- [P1] Terminate the process group when `WaitDelay` expires
  `internal/config/command.go:71`
  A command such as `sleep 600 & exit` lets its shell exit immediately while the background child retains the inherited stdout/stderr pipes. `cmd.Wait()` then returns `exec.ErrWaitDelay` after the new one-second delay, which selects this branch and returns without calling `proc.Terminate()`. On POSIX `Close` is a no-op, so the child continues running; before this change the five-second timer would kill that process group. Treat the wait-delay result as an incomplete command, terminate the tree, and add a background-child regression case.

- [P1] Close the Start-to-job-assignment escape window
  `internal/config/process_windows.go:31`
  The job is attached only after `cmd.Start()` has allowed `cmd.exe` to execute. A fast command can create and detach a child before `AssignProcessToJobObject`, leaving that child outside the job. On timeout, `Terminate()` kills the job/root first and only then runs `taskkill` against the now-dead root, so the fallback cannot traverse to that reparented child. It can continue holding the pipes or executing after the timeout, preserving the race this PR is intended to eliminate. Start suspended and assign before resuming (or use equivalent launch-time containment), and test a forced pre-assignment child.

- [P2] Do not kill descendants after a successful provider command
  `internal/config/process_windows.go:80`
  `Close` is deferred for every return path, and the job has `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. Therefore a provider command which successfully emits its JSON but launches a detached helper/proxy has that child terminated as `LoadProviderCommand` returns. The old behavior only killed the tree on timeout. Restrict the kill-on-close cleanup to timeout/error cleanup, or otherwise preserve successful-command descendants.

- [P3] Quote PID-file paths when generating the timeout fixture
  `internal/config/command_test.go:159`
  The new PowerShell and POSIX fixtures interpolate `PidFile` inside single-quoted shell literals without escaping apostrophes. A temporary directory such as `O'Brien` produces an invalid script, so the PID file is never written and this test fails before exercising the timeout behavior. Escape the platform-specific quote character (or avoid shell interpolation) for both generated scripts.

- [P3] Correct the Windows process-status comment
  `internal/config/process_windows_test.go:19`
  Exit code `259` is `STILL_ACTIVE`, not `STATUS_PENDING`. The value is correct, but the comment misstates the Windows API contract and will mislead future maintenance; use the proper constant name or correct the comment.

jatmn (changes requested):

## Findings

- [P1] Terminate leaked descendants when the provider shell exits nonzero
  `internal/config/command.go:71`
  `exec.Cmd.Wait` preserves an `ExitError` over its pipe-drain `ErrWaitDelay`. A command such as `sleep 600 & exit 7` therefore returns `exit status 7` after the one-second drain deadline, bypasses this `ErrWaitDelay` branch, and leaves the background process running: POSIX `Close` is a no-op and the Windows job handle is merely released. Track the drain-timeout condition independently and terminate the tree even when the shell's exit status is nonzero; add a nonzero-exit background-child regression case.

- [P1] Do not rely on `taskkill` after failed job assignment and a completed wait
  `internal/config/process_windows.go:53`
  If a parent job forbids nested jobs (or either job setup call fails), assignment falls back to `taskkill`. A provider command can then start a detached child and exit; by the time `WaitDelay` returns, `cmd.exe` has been reaped and `/T /PID <cmd.exe>` cannot walk to the reparented child. `LoadProviderCommand` reports a timeout while the child remains running, recreating the leak this PR is intended to eliminate. Provide launch-time containment for the fallback or otherwise preserve a reliable descendant handle, and test a forced assignment-failure/background-child path.

- [P1] Never pass a reaped provider PID to `taskkill`
  `internal/config/process_windows.go:85`
  On the `ErrWaitDelay` path, `cmd.Wait` has already reaped the root process. Even after a successfully attached job has terminated the original tree, this method unconditionally runs `taskkill /PID` using that stale numeric PID. Windows can reuse PIDs, so a fast-reused PID can make a provider-config timeout forcibly kill an unrelated process tree. Skip the PID fallback once the job path was used, and do not use a post-reap PID as a tree identity.

- [P2] Handle failure to resume the suspended command
  `internal/config/process_windows.go:64`
  Every provider command is created with `CREATE_SUSPENDED`, but failures from the Toolhelp snapshot, thread enumeration, `OpenThread`, and `ResumeThread` are ignored. If the primary thread cannot be reopened (Toolhelp acquisition itself is fallible), it stays suspended and an otherwise valid command is misreported as a five-second timeout. Retry or surface the resume failure, ensure the process is cleaned up, and add coverage for that failure path.

- [P3] Correct the PR description's job-object semantics
  `internal/config/process_windows.go:99`
  The PR description says the job uses `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, but the current implementation explicitly does not set that limit and intentionally leaves successful-command descendants running when `Close` releases the job handle. Update the description so reviewers do not approve a behavior the code deliberately avoids.

Vasanthdev2004 (approved):
The job-object approach is sound — suspend-start, assign to the job, then resume means descendants can't escape the tree, and the WaitDelay + ErrWaitDelay handling closes the pipe-leak gap that taskkill /T had. Build, vet, gofmt, and the config tests all pass on Windows, including the new background-child test (1.66s, confirming the WaitDelay path fires and the leftover child is actually killed). Two small things worth a glance: `resumeMainThread` resumes every thread owned by the pid, not just the primary one (harmless, arguably more correct, but the name oversells it); and the PR description claims `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` is set, but the code intentionally doesn't set it — the Close() comment is accurate, so the description should be corrected to avoid confusing future readers. Approving as-is.

Vasanthdev2004 (changes requested):
Switching my earlier approve to request changes  I reproduced the leak jatmn flagged.

The blocking one: a provider command that exits nonzero while a detached child holds the inherited pipes (e.g. `sleep 600 & exit 7`) leaks the child. Go only returns `exec.ErrWaitDelay` on a *successful* exit  a nonzero exit yields the bare `*ExitError`, so the `errors.Is(err, exec.ErrWaitDelay)` branch at `command.go:72` is skipped, `proc.Terminate()` never runs, and the child survives (on Windows `Close` just drops the job handle  no `KILL_ON_JOB_CLOSE`; on POSIX `Close` is a no-op). I confirmed it on Windows: the sleeper PID stayed alive after a `...exit status 7`. The new background-child test uses exit 0, so it only walks the already-working branch. It's cross-platform  the POSIX path has the same gating. Smallest fix: call `proc.Terminate()` whenever `err != nil` (killing an already-dead parent is harmless), and add an `ExitCode: 7 + BackgroundSleepSeconds` test variant.

Second one, high: `process_windows.go` `Terminate()` runs `taskkill /T /F /PID` + `Process.Kill()` unconditionally, even after `TerminateJobObject` already killed the tree. On the `ErrWaitDelay` path the root is already reaped by `Wait`, so that PID is free and Windows can hand it to an unrelated process  `taskkill /T` on a reused PID force-kills the wrong tree. Gate the PID fallback with `if p.job == 0` (it's only the real fallback when job creation failed) and skip it once the process is reaped.

The rest are real but fine as follow-ups: the no-job fallback can't reach a reparented descendant (non-regressive  pre-PR had no containment at all), `resumeMainThread` swallows Toolhelp/OpenThread/ResumeThread failures so a stuck-suspended command misreports as a 5s timeout, the new background-child test can flake if PowerShell cold-start loses the race with the 1s `WaitDelay` kill before `bg.pid` is written, and the `itoa` test helper returns `""` for negatives. What's solid: `CREATE_SUSPENDED` + assign-before-resume closes the root escape window, the job kills the whole tree on the success path, the POSIX group-kill is correct, and bounding `Wait` with `WaitDelay` is the right idea.

jatmn (changes requested):

## Findings

- [P1] Do not invoke `taskkill` with a PID that `Wait` has already reaped
  `internal/config/process_windows.go:115`
  When creating or assigning the job fails (for example, in a non-nestable parent job), `proc.job` stays zero. On the new `WaitDelay` and nonzero-exit paths, `cmd.Wait()` has already reaped `cmd.exe` before `Terminate` reaches this fallback, so its numeric PID can already have been recycled. `taskkill /T /F /PID` can then force-kill an unrelated process tree; this is precisely the stale-PID hazard the job path avoids above. Keep the fallback from using the PID after a completed wait (and use an identity-preserving cleanup path for the no-job case).

- [P2] Synchronize the Windows fixture before starting the one-second wait-delay race
  `internal/config/command_test.go:235`
  `start /B powershell` returns before the child gets to `Set-Content`. On a cold or loaded Windows machine, the new job can be terminated by the one-second `WaitDelay` before that write, after which `assertProcessTerminated` fails to read `bg.pid`. That makes the regression test flaky and can let it fail without ever proving that a pipe-holding child was started. Arrange for the fixture to publish the PID before the parent exits, or otherwise synchronize PID-file creation before exercising the wait-delay path.

Vasanthdev2004 (changes requested):
Most of this landed exactly as asked: Terminate() now fires on any nonzero Wait result, the exit-7 background-child test proves the leak is closed, the job path no longer falls through to taskkill after TerminateJobObject, and surfacing resume failures instead of letting a suspended process masquerade as a 5s timeout is a nice touch. CI is green across the board, including the new tests on Windows.

One thing keeps me at request-changes, and it's the second half of my earlier ask — jatmn hit the same spot on this head. When job creation/assignment fails, Terminate() still hands the raw PID to `taskkill /T /F` on the WaitDelay/nonzero-exit paths (process_windows.go:123-125), where cmd.Wait() has already reaped cmd.exe. A dead root can't be tree-walked, so the call buys nothing there — and if Windows recycled the PID it force-kills an unrelated tree. Before this PR, taskkill only ever ran while the root was still alive, so this path is new exposure. Cheapest fix I see: skip the PID fallback whenever Wait has already returned (the caller knows which branch it's on), or keep the handle you already open in attachCommandProcess when assignment fails and check STILL_ACTIVE before touching the PID. The `start /B` fixture race jatmn mentioned is worth tightening while you're in there, but I'm not blocking on that one since CI is green.

jatmn (changes requested):

## Findings

- [P1] Do not signal a recycled POSIX process group after `Wait` reaps the shell
  `internal/config/process_posix.go:26`
  The new unconditional error path calls `Terminate` only after `cmd.Wait()` has returned, which means the shell has already been reaped. On every ordinary nonzero provider-command exit, the unconditional `Kill(-p.cmd.Process.Pid, SIGKILL)` can therefore use a recycled numeric process-group ID; if another process has claimed that PID and formed a group before this call, Zero kills that unrelated group. The same window exists after `ErrWaitDelay` when a child has left the original group, and the new tests only cover children that remain in it. Preserve an identity/liveness guarantee for the group, or avoid the group-directed kill once `Wait` has completed.

PierrunoYT:
Addressed the remaining process-lifecycle feedback in `d6700a1` and pushed the requested TUI baseline correction in `0089d0d`.

### Process safety

- POSIX provider commands now join a process group led by a dedicated anchor process.
- The anchor remains alive and unreaped until cleanup, so `Terminate` never signals a group identified by the already-reaped provider PID.
- Timeout and nonzero-exit cleanup still kill pipe-holding descendants with one group-directed signal.
- Added a POSIX regression test that reaps the provider first, verifies the independent group identity remains live, and confirms cleanup terminates it.
- The Windows `taskkill` fallback now uses a two-second context timeout and still falls through to `TerminateProcess` on timeout/failure.

### Additional requested fix

- Updated the stale provider-wizard test assertion from `Enter continue` to the current `Enter/→ continue` rendering.

### Validation

- `go test ./internal/config` on Windows
- `go test ./internal/config` on Ubuntu under WSL
- Linux cross-compilation of `internal/config` tests
- `go test ./...`
- `go vet ./...`
- `govulncheck`: no vulnerabilities found
- The mandated lint command still reports the 35 pre-existing findings tracked in #743; none are in the changed files.

Please re-review the POSIX process-group identity fix.

jatmn (changes requested):

## Findings

- [P1] Restore the credential-step footer assertion
  `internal/tui/provider_wizard_test.go:194`
  This test now expects `Enter/→ continue`, but the credential step has no right-arrow action and the renderer returns `Enter continue   ← back   Esc close`. The assertion therefore fails deterministically: Smoke (ubuntu-latest), Smoke (macos-latest), and Smoke (windows-latest) all fail on it. Restore the assertion to the current UI contract, or change the production shortcut and its rendering together if that is the intended product change.

Vasanthdev2004 (changes requested):
Re-verified on the current head, and the good news first: every concern I raised is genuinely fixed. Nonzero-exit termination fires now, not just on ErrWaitDelay; the Windows job path no longer falls through to taskkill after TerminateJobObject; and the reaped-PID-to-taskkill hazard I flagged last round is closed the way we discussed, with the retained process handle plus a GetExitCodeProcess == STILL_ACTIVE gate before taskkill. The process-kill logic itself is ready.

The hold is CI: smoke is red on all three OSes, and it's a real failure this PR introduced, not a flake. TestProviderWizardAdvancesProviderAPIKeyAndModelSteps (provider_wizard_test.go:198) expects the key-step footer to contain "Enter/→ continue", but the wizard actually renders "Enter continue   ← back   Esc close". The test's expected string doesn't match what the footer prints, so it looks like a footer-string change got half-applied to the test. Fix the expectation (or the footer) so the suite goes green and I'll approve.

jatmn (approved):
@Vasanthdev2004 LGTM

gnanam1990 (approved):
## Review: PR #690 — fix(config): kill provider-command process tree via job object

**Verdict: Approve.** The core fix is correct and load-bearing — the retained POSIX group anchor and the Windows job object give `Terminate()` a stable handle on the whole tree even after `cmd.Wait()` has reaped the root, which is exactly the leak the PR targets. One minor behavioral regression is worth a conscious accept/decline, but nothing here blocks merge.

### What the fix does well
- `cmd.WaitDelay = time.Second` (internal/config/command.go:47) is what makes `cmd.Wait()` unblock ~1s after the shell exits instead of hanging on descendant-held pipes until the 5s timer — verified load-bearing: `TestLoadProviderCommandTerminatesBackgroundChild` (a `sleep 10 & exit` scenario) asserts return in <4s and that the leaked child PID is killed; that test fails on base (base has no WaitDelay, so it only unblocks at 5s).
- The POSIX anchor process (internal/config/process_posix.go:20-43) deliberately does not trust the reaped shell PID to name the group — the comment at line 17-19 spells out why, and `Terminate()` SIGKILLs `-groupID`.
- Windows `Terminate()` (process_windows.go:111-145) correctly refuses to fall through to a `taskkill /T /PID` against a possibly-reused PID once the root is reaped, gating the fallback on a live `STILL_ACTIVE` handle. The CREATE_SUSPENDED-then-assign-to-job sequencing (line 23-25, 45-79) closes the escape window for fast descendants. Good defensive commenting throughout.

### [Minor] WaitDelay=1s turns a previously-successful command into a spurious timeout — PR-introduced
`internal/config/command.go:47` (surfaces at :79-81 and :17-21)

Concrete scenario — a provider command that prints valid JSON then backgrounds a helper without redirecting its stdio and keeps it alive 1–5s:
```sh
echo '{"name":"x","provider":"openai","apiKey":"sk","model":"gpt-4"}'
sleep 2 &
exit 0
```
- **Base** (no `WaitDelay`): the JSON is buffered, the shell exits 0, `cmd.Wait()` drains the inherited pipes until `sleep 2` exits (~2s, under the 5s timer), returns nil, and `LoadProviderCommand` parses the buffered JSON → **success**.
- **PR**: `WaitDelay` fires 1s after the shell exits, force-closes the pipes, `cmd.Wait()` returns `exec.ErrWaitDelay`; command.go:79-81 maps that to `errProviderCommandTimeout`, and `LoadProviderCommand` (line 17-21) discards the already-captured stdout and returns `"provider command timed out after 5s"` — a config load that worked before now fails (and the helper is SIGKILLed).

This only bites commands that leave an stdio-inheriting descendant alive for 1–5s; anything that redirects (`daemon >/dev/null 2>&1 &`) is unaffected, which is why this reads as an intended bound on drain time rather than an oversight. Two things worth noting if you accept it: (1) the message says "timed out after 5s" when the cut actually happens at ~1s, which is misleading; (2) on the `ErrWaitDelay` path stdout already holds valid JSON — parsing it before returning the timeout would preserve base success behavior for this case. Either fix the message or attempt the parse; otherwise, an explicit acknowledgement that this changes the success contract is enough.

### Tests
- `gofmt -l` clean on all 7 changed files; `go build` and `go vet` on `internal/config/...` and `internal/tui/...` both exit 0.
- `go test -race ./internal/config/...`: **ok**. All PR tests pass, including `TestLoadProviderCommandTimeout`, `TestLoadProviderCommandTerminatesBackgroundChild{,OnFailure}`, `TestCommandProcessPreservesSysProcAttr`, `TestCommandProcessRetainsGroupIdentityAfterProviderWait`. Windows-only files are build-tag excluded on darwin.
- `internal/tui` package reports FAIL, but **solely** from `TestHandleAddDirCommand` (add_dir_test.go:43) — this is a **pre-existing environmental sandbox failure**: it fails identically on the clean base worktree with the same error (the sandbox already has `/private/tmp` + the macOS temp dir in its write roots) and is unrelated to the diff. The one tui test the PR actually touches, `TestProviderWizardAdvancesProviderAPIKeyAndModelSteps`, passes on both head and base. Not PR-attributable.

Vasanthdev2004 (approved):
Approving on 467ff555. Our only hold last round was CI (the provider-wizard credential-step footer assertion), and that is green now across macos, ubuntu, and windows.

The delta since then is clean: the footer test is fixed, and process_posix.go now sets Setpgid/Pgid on the existing SysProcAttr rather than replacing the whole struct, so any attributes a caller already set are preserved. The Windows process-kill logic we worked through last round is unchanged, still gated on the retained process handle plus a GetExitCodeProcess == STILL_ACTIVE check before taskkill.

Good to go, thanks for the thorough back-and-forth. LGTM.

kevincodex1 (approved):

PR #691, opened 2026-07-14
fix(lsp): make real-gopls integration test opt-in via env var

### LINKED ISSUE #684: Real gopls manual test can fail the default suite on an unusable installation
## Summary

`TestManagerCheckRealGopls` is described as a temporary manual-verification test, but it runs as part of the default `go test ./...` suite whenever a `gopls` executable exists. An installed but unusable `gopls` causes the entire suite to fail instead of skipping or remaining opt-in.

## Reproduction

```powershell
go test ./internal/lsp -run '^TestManagerCheckRealGopls$' -count=1 -v
```

Observed:

```text
Manager.Check returned error: lsp error -1: EOF
```

On this host, `gopls version` also reports that it cannot create its persistent-index cache because a cache path component already exists in an incompatible form. The binary is present on `PATH`, so the test does not skip, but the server exits during initialization and Zero observes EOF.

## Relevant code

`internal/lsp/real_gopls_manual_test.go` skips only when `exec.LookPath("gopls")` reports that the binary is absent. Its comment calls it a temporary manual-verification test, yet it has no build tag or explicit opt-in.

## Expected behavior

The default hermetic test suite should not fail based on the health, version, or local cache state of a developer's globally installed language server.

## Actual behavior

Any executable named `gopls` enables the integration test; initialization failures such as EOF fail `go test ./...`.

## Suggested fix

Move the real-server test behind an integration build tag or explicit environment-variable opt-in. Keep default coverage on the existing fake LSP server. If the real-server test remains automatic, probe `gopls` health/version and skip with a clear reason when initialization is unavailable.

PierrunoYT:
## Summary

Fixes #684 — `TestManagerCheckRealGopls` ran as part of the default `go test ./...` suite whenever any `gopls` binary was on `PATH`, so an installed but unusable gopls (e.g. a broken persistent-index cache) failed the whole suite with `lsp error -1: EOF`.

## Changes

The test is now gated behind an explicit `ZERO_LSP_REAL_GOPLS_TEST=1` opt-in, following the repo's existing convention for real-dependency smoke tests (`ZERO_STT_DOWNLOAD_TEST`, `ZERO_STT_STREAM_SERVER`, …). The `gopls`-on-PATH skip is kept as a secondary guard for the opt-in path. Default hermetic coverage continues to run against the fake LSP server.

## Verification

On Windows (go1.26.5 windows/amd64):
- Default run: `go test ./internal/lsp -run '^TestManagerCheckRealGopls$' -v` → SKIP with a clear opt-in message.
- Opt-in run: `ZERO_LSP_REAL_GOPLS_TEST=1 go test ...` → PASS against a real gopls (1 compiler diagnostic reported).
- Full `go test ./internal/lsp` passes.

## Summary by CodeRabbit

* **Tests**
  * The real-language-server verification test now runs only when explicitly enabled with the `ZERO_LSP_REAL_GOPLS_TEST` environment variable.
  * The test continues to skip automatically when the required language server is unavailable.

### DIFF
--- a/internal/lsp/real_gopls_manual_test.go
+++ b/internal/lsp/real_gopls_manual_test.go
@@ -11,9 +11,13 @@ import (
 
 // TestManagerCheckRealGopls drives the exact path the self-correct LSP wiring
 // uses (NewManager -> Check) against a REAL gopls and asserts a real diagnostic
-// comes back. Skips when gopls is not installed. Temporary manual-verification
-// test (the rest of the suite uses a fake server).
+// comes back. Opt-in manual-verification test (the rest of the suite uses a
+// fake server): the health, version, or cache state of a developer's globally
+// installed gopls must not fail the default hermetic suite.
 func TestManagerCheckRealGopls(t *testing.T) {
+	if os.Getenv("ZERO_LSP_REAL_GOPLS_TEST") == "" {
+		t.Skip("set ZERO_LSP_REAL_GOPLS_TEST=1 to run the real gopls integration test")
+	}
 	if _, err := exec.LookPath("gopls"); err != nil {
 		t.Skip("gopls not on PATH; skipping real-server check")
 	}
--- a/internal/lsp/real_gopls_manual_test.go
+++ b/internal/lsp/real_gopls_manual_test.go
@@ -15,7 +15,7 @@ import (
 // fake server): the health, version, or cache state of a developer's globally
 // installed gopls must not fail the default hermetic suite.
 func TestManagerCheckRealGopls(t *testing.T) {
-	if os.Getenv("ZERO_LSP_REAL_GOPLS_TEST") == "" {
+	if os.Getenv("ZERO_LSP_REAL_GOPLS_TEST") != "1" {
 		t.Skip("set ZERO_LSP_REAL_GOPLS_TEST=1 to run the real gopls integration test")
 	}
 	if _, err := exec.LookPath("gopls"); err != nil {


jatmn (approved):
@Vasanthdev2004 LGTM

anandh8x (approved):
LGTM. Hermetic-by-default, opt-in via `ZERO_LSP_REAL_GOPLS_TEST=1`, matches existing `ZERO_*_TEST` convention. The `== "1"` strict match fix in 23390b0 is correct. Ship it.

Follow-up (non-blocking): add a short CONTRIBUTING.md note listing the `ZERO_LSP_*_TEST` env vars so future real-LSP tests for other languages follow the same pattern.

Vasanthdev2004 (approved):
Read the diff  the ZERO_LSP_REAL_GOPLS_TEST=1 guard goes in before the LookPath check, which is the right order, and the skip message is clear. Confirmed gofmt and `go vet ./internal/lsp/` are clean on the touched file, and without the env var set the test skips and the full ./internal/lsp/ suite passes. Test-only and low risk, so I'm approving. Good call gating this behind an env var so a globally installed gopls can't flake the hermetic suite.

Vasanthdev2004:
Thanks @PierrunoYT. This is now covered on main: #766 landed the same opt-in for the real-gopls manual test, so #684 is resolved. The two PRs did the same thing with a different env var name, so I am closing this one as superseded. Appreciate the fix, and sorry for the overlap. If the merged version ever needs a tweak, happy to take a follow-up.

PR #692, opened 2026-07-14
fix(providers): update xAI catalog default to grok-4.5

lfaoro:
Add grok-4.5 as the curated xAI default and keep grok-4 as the previous flagship for offline/fallback model selection.

## Summary by CodeRabbit

* **Updates**
  * Added xAI’s Grok 4.5 to the available model list.
  * Set Grok 4.5 as the default xAI model for new sessions and OAuth setup.
  * Kept Grok 4 available, labeled as the previous flagship model.

### DIFF
--- a/internal/providercatalog/catalog.go
+++ b/internal/providercatalog/catalog.go
@@ -139,7 +139,7 @@ var descriptors = []Descriptor{
 	openAICompat("github", "GitHub Models", "https://models.inference.ai.azure.com", "openai/gpt-4.1", []string{"GITHUB_TOKEN"}, "github-models"),
 	transportDescriptor("bedrock", "Amazon Bedrock", TransportBedrock, "https://bedrock-runtime.${AWS_REGION}.amazonaws.com", "anthropic.claude-3-5-sonnet-20241022-v2:0", []string{"AWS_ACCESS_KEY_ID", "AWS_PROFILE"}, []APIFormat{APIFormatBedrockConverse}, true),
 	transportDescriptor("vertex", "Vertex AI", TransportVertex, "https://aiplatform.googleapis.com", "gemini-2.5-pro", []string{"GOOGLE_APPLICATION_CREDENTIALS"}, []APIFormat{APIFormatVertexGenerateContent}, true),
-	oauthProvider(openAICompat("xai", "xAI", "https://api.x.ai/v1", "grok-4", []string{"XAI_API_KEY"}), false, true),
+	oauthProvider(openAICompat("xai", "xAI", "https://api.x.ai/v1", "grok-4.5", []string{"XAI_API_KEY"}), false, true),
 	openAICompat("venice", "Venice AI", "https://api.venice.ai/api/v1", "qwen-2.5-qwq-32b", []string{"VENICE_API_KEY"}),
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P2] Register Grok 4.5's reasoning capability with the promoted default
  `internal/providermodelcatalog/catalog.go:146`
  The new default is only a bare curated model row. Because it has no model-registry entry and the unknown-model fallback has no Grok rule, the TUI's `/effort` command reports no controls for `grok-4.5`; users cannot select its documented `low`, `medium`, or `high` reasoning levels and are stuck with the provider default (high). Add its supported reasoning metadata (and regression coverage) to the capability path when promoting it as the default.

internal/providercatalog/catalog.go
```diff
@@ -139,7 +139,7 @@ var descriptors = []Descriptor{
 	openAICompat("github", "GitHub Models", "https://models.inference.ai.azure.com", "openai/gpt-4.1", []string{"GITHUB_TOKEN"}, "github-models"),
 	transportDescriptor("bedrock", "Amazon Bedrock", TransportBedrock, "https://bedrock-runtime.${AWS_REGION}.amazonaws.com", "anthropic.claude-3-5-sonnet-20241022-v2:0", []string{"AWS_ACCESS_KEY_ID", "AWS_PROFILE"}, []APIFormat{APIFormatBedrockConverse}, true),
 	transportDescriptor("vertex", "Vertex AI", TransportVertex, "https://aiplatform.googleapis.com", "gemini-2.5-pro", []string{"GOOGLE_APPLICATION_CREDENTIALS"}, []APIFormat{APIFormatVertexGenerateContent}, true),
-	oauthProvider(openAICompat("xai", "xAI", "https://api.x.ai/v1", "grok-4", []string{"XAI_API_KEY"}), false, true),
+	oauthProvider(openAICompat("xai", "xAI", "https://api.x.ai/v1", "grok-4.5", []string{"XAI_API_KEY"}), false, true),
```
gnanam1990, line 142: [P2] Confirm `grok-4.5` is a live model ID on api.x.ai before merging

This flips the xAI default from `grok-4` to `grok-4.5` (also in `internal/providermodelcatalog/catalog.go`). If `grok-4.5` is not served by api.x.ai (or the ID differs, e.g. `grok-4-5`), every default xAI request 404s on first use. `grok-4` is retained as "previous flagship" so a revert is trivial. Could not verify without API access.

gnanam1990 (commented):
Local review (not a blocker): built and ran touched-package tests locally; all pass. One verification ask before merge.

internal/providercatalog/catalog.go
lfaoro: Thanks @gnanam1990 for the review.

`grok-4.5` is the live model ID on `api.x.ai`.
https://docs.x.ai/developers/models 

image

lfaoro: xAI `/models` endpoint is gated by token/credits but you can see the model ID via openrouter open endpoint.

```bash
curl -s https://openrouter.ai/api/v1/models \
  | jq '.data[] | select(.id | startswith("x-ai")) | {id, name}'
```

Vasanthdev2004 (approved):
Looks good bumping the xAI default to grok-4.5 across the provider descriptor, the curated model list (grok-4 retained as "previous flagship"), and the onboarding test fixture. grok-4.5 is the real publicly released xAI model (announced July 8, 2026, id `grok-4.5`), build/vet/gofmt are clean on the touched packages, and the providercatalog, providermodelcatalog, and xAI/onboarding tui tests pass. No attribution-header or trust-anchor concerns on this one. Approving.

PR #693, opened 2026-07-15
feat(plugins): add CLI enable/disable for plugin.json

### LINKED ISSUE #790: feat(plugins): add CLI enable/disable for plugin.json
## Summary

Add `zero plugins enable <id>` and `zero plugins disable <id>` so installed plugins can be toggled via the `enabled` field in `plugin.json` without removing them.

## Motivation

Hooks and MCP already support enable/disable. Plugins currently require remove/re-add (or hand-editing `plugin.json`) to deactivate. EXTENDING.md previously called this out as a gap.

## Proposed behavior

- `zero plugins enable <id>` / `zero plugins disable <id>`
- Flags: `--user` (user plugins only), `--json`
- Idempotent: already-enabled / already-disabled reports a clear no-op message
- Project plugins take precedence over user plugins unless `--user` is set
- Preserve existing `plugin.json` file mode when rewriting

## Related PR

- https://github.com/Gitlawb/zero/pull/693

## Notes for maintainers

Please add the `issue-approved` label if this scope is accepted so the linked community PR can proceed under CONTRIBUTING.md.

Ayush7614:
## Summary

- Add `zero plugins enable|disable <id>` to toggle the `enabled` field in `plugin.json` (with `--user` / `--json`)
- Add `plugins.SetEnabled` / `SetEnabledByID` helpers used by the CLI
- Update `docs/EXTENDING.md` so the documented gap matches the new commands

Closes #790

## Test plan

- [x] `go test ./internal/plugins/ ./internal/cli/ -run 'Enabled|EnableDisable|PluginInfo|Info|SetEnabled' -count=1`
- [x] `go vet ./internal/plugins/ ./internal/cli/`
- [x] `govulncheck ./internal/plugins/` — no vulnerabilities
- [x] Windows: `TestSetEnabledPreservesManifestMode` skipped on `windows` (`dae3ef8`)
- [x] Merged latest `main` (includes `#773` plugins info) and deduplicated `ErrNotInstalled`

## Summary by CodeRabbit

* **New Features**
  * Added `zero plugins enable <id>` and `zero plugins disable <id>` commands (without removing the plugin).
  * Added `--user` to target only user-installed plugin manifests, ignoring project plugins.
  * Added `--json` output and clearer messaging for “already enabled/disabled” vs “now enabled/disabled”.
  * Updated command help to include the new subcommands.

* **Bug Fixes**
  * Preserves existing `plugin.json` permission bits when toggling enablement.

* **Documentation**
  * Updated plugin docs with the new enable/disable commands and activation rules based on `plugin.json.enabled` (defaults to enabled when omitted).


Ayush7614:
@coderabbitai review the latest commits and changes

Vasanthdev2004 (changes requested):
Nice feature — the atomic manifest rewrite, idempotency, and project/user precedence all look right, and the new CLI tests pass on my end. One blocker: `TestSetEnabledPreservesManifestMode` asserts a 0644 mode after `os.Chmod`, but on Windows `os.Stat` reports 0666 regardless of chmod, so it fails on `windows-latest` (which is in the CI matrix and runs `go test ./...`). Guard that test with a `runtime.GOOS != "windows"` skip (or drop the exact-bits assertion) and I think this is good to go.

Ayush7614:
Thanks for the review, @Vasanthdev2004!

Addressed the Windows CI blocker in dae3ef8: `TestSetEnabledPreservesManifestMode` now skips on `windows-latest` since `os.Stat` reports `0666` regardless of `chmod` there.

The mode-preservation behavior is still covered on Unix via the existing assertion. Please take another look when you have a moment.

Ayush7614:
@Vasanthdev2004  Merged latest `main` in e29fddc and cleared the conflict with `#773` (`zero plugins info`):

- `docs/EXTENDING.md` now documents `info`, `disable`, and `enable`
- Shared a single `plugins.ErrNotInstalled` sentinel used by both info and enable/disable

Earlier review items from this thread:

- **CodeRabbit:** `ErrNotInstalled` + `--user` / already-disabled CLI coverage — done in `7a1a5b5`
- **@Vasanthdev2004:** Windows mode assertion skip — done in `dae3ef8` (windows-latest CI was green on that commit)

Also opened https://github.com/Gitlawb/zero/issues/790 and linked it via `Closes #790`. Please add `issue-approved` when the scope looks good, and re-review if you’re happy with the Windows fix + merge.

Vasanthdev2004 (approved):
Approving. The key behaviour is right: disabling a plugin actually stops its components loading rather than just hiding it from a listing, which was the thing worth verifying. The state file is written atomically and a missing or corrupt one degrades safely, and the Windows test-mode fix that was the last blocker is in and that leg is green.

A few minor notes for later rather than now: concurrent toggles from two processes are not serialized (last writer wins, which is fine for a CLI toggle but worth knowing), and disabling a plugin mid-session takes effect on the next activation rather than immediately. Neither is a correctness problem for how this is used.

Verification note: the go toolchain is blocked on my machine by Smart App Control, so this is a source review; CI is green.

Thanks for the quick turnaround on the Windows leg. LGTM.

Ayush7614:
cc: @kevincodex1

gnanam1990 (changes requested):
## Verdict

Changes requested. The command surface is well built and the feature is wanted, but the manifest write in `internal/plugins/enabled.go` re-implements — less safely — machinery this same package already owns and already uses.

Reviewed at `e29fddc`, base `d9b882e` (`origin/main` at `097c265`). Parent issue #790 carries `issue-approved`.

## What I verified

- **The tests are load-bearing.** I stubbed out the publish step so `SetEnabled` returns `changed=true` without ever writing the manifest, and two tests caught it with the un-toggled file printed in the failure:

  ```
  --- FAIL: TestRunPluginsEnableDisable
      extensions_test.go:376: manifest not disabled: { "schemaVersion": 1, "id": "zero.demo", ... }
  --- FAIL: TestRunPluginsDisableUserIgnoresProject
      extensions_test.go:447: user manifest not disabled: { ..., "enabled": true }
  ```

- `go vet ./internal/plugins/ ./internal/cli/` clean; `go test ./internal/plugins/ -count=1` passes.
- Merged `origin/main` into the branch: auto-merge clean, `go build ./...` succeeds on the merged tree.
- Confirmed the orphan-temp-file concern does **not** apply: discovery joins the exact filename at `internal/plugins/plugins.go:253` (`filepath.Join(pluginDir, "plugin.json")`), so a stranded `plugin.json.tmp-*` is never mistaken for a manifest.
- Ran on darwin/arm64 from a non-`/tmp` checkout.

## Blocking

1. **`SetEnabled` takes no lock, while every other mutator in this package does.** — `internal/plugins/enabled.go:41`

   `Install` acquires `installtxn.Lock(dir)` at `internal/plugins/install.go:151`, and `Remove` acquires it at `:203`. Both then mutate under it. `SetEnabled` performs a read-modify-write on a file inside that same tree with no lock at all, and `installtxn.WriteFileAtomically` explicitly documents that responsibility as the caller's:

   ```go
   // WriteFileAtomically publishes data by renaming a complete sibling temporary
   // file over path. The caller is responsible for any surrounding transaction
   // lock.
   ```

   `installtxn.Lock` is a real cross-process lock — `flock` on Unix (`lock_unix.go`), `LockFileEx` on Windows (`lock_windows.go`) — which matters here because this repo runs CLI, TUI and daemon against the same state. The concrete failure: `zero plugins disable X` reads the manifest, `zero plugins remove X` then takes the lock and tears the directory down, and the first process's rename recreates `plugin.json` inside the removed plugin's directory. The result is a manifest that discovery will load with no corresponding entry in `plugins.lock` — precisely the inconsistency the lock exists to prevent.

2. **The atomic write is hand-rolled and is weaker than the helper this package already calls.** — `internal/plugins/enabled.go:78-88`

   `internal/plugins/install.go:271` already calls `installtxn.WriteFileAtomically`. The new code writes its own temp-file-plus-`os.Rename` instead, and differs in two ways that matter:

   - **No `Sync`.** `installtxn.WriteFileAtomically` calls `temp.Sync()` before closing. Without it, a crash between the write and the rename can publish a truncated or zero-length `plugin.json`. That file then fails `json.Unmarshal`, and the plugin drops out of discovery entirely — a worse outcome than either enabled or disabled.
   - **`os.Rename` instead of `replaceFile`.** On Windows the package uses `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING|MOVEFILE_WRITE_THROUGH)` (`internal/installtxn/lock_windows.go:29`). Go's `os.Rename` does not pass `MOVEFILE_WRITE_THROUGH`, so the durability guarantee the package deliberately chose is lost on the platform it was chosen for. The existence of a platform-specific `replaceFile` is the signal that plain `os.Rename` was already judged insufficient here.

   Both resolve together: take `installtxn.Lock` on the plugins root, then publish with `installtxn.WriteFileAtomically(resolved, append(data, '\n'), mode)`. That also deletes the `os.Getpid()`/`UnixNano()` temp-name construction, which is only there because the helper was not used.

3. **Toggling rewrites the manifest with its fields reordered.** — `internal/plugins/enabled.go:71`

   `json.MarshalIndent` over a `map[string]any` emits keys in sorted order, so a single `zero plugins disable` scrambles a hand-authored file. Measured against the patched build:

   ```
   BEFORE                       AFTER
   {                            {
     "name": "demo",              "author": "a",
     "version": "1.0.0",          "description": "z last",
     "description": "z last",     "enabled": false,
     "author": "a",               "name": "demo",
     "enabled": true              "version": "1.0.0"
   }                            }
   ```

   `plugin.json` is a file users edit by hand and commit. Reordering it on every toggle produces diff noise unrelated to the change, and for a project plugin under `./.zero/plugins` that noise lands in the repository. Preserving order needs either an ordered decode (`json.Decoder` with `Token()`) or a targeted textual edit of the single `enabled` field rather than a full re-marshal. If that is judged too costly for the value, say so in the PR description and in `docs/EXTENDING.md` so the churn is at least documented rather than surprising — but it should be a stated decision, not a side effect.

## Non-blocking

- **Permission fallback widens on a stat failure.** — `internal/plugins/enabled.go:75-78`. When `os.Stat` returns an error the mode silently falls back to `0o644`. A manifest that was `0o600` would be republished world-readable. `os.ReadFile` has already succeeded by that point so the window is narrow, but the repo's convention on security-adjacent state is to fail closed rather than pick a permissive default — returning the stat error would be more consistent. Using `installtxn.WriteFileAtomically` does not fix this by itself; the mode decision stays yours.
- **Symlinked manifests are replaced rather than followed.** — `internal/plugins/enabled.go:47`. `filepath.Abs` does not resolve symlinks, so if `plugin.json` is a symlink the rename replaces the link with a regular file. Probably out of scope, but `filepath.EvalSymlinks` is a one-line guard if you want it.
- **Default target when a plugin exists in both scopes.** — `internal/cli/extensions.go:127`. Without `--user`, `LoadOptions{ExcludeProject: false}` means project precedence wins, so `zero plugins disable X` writes the *project* manifest when both exist. The command does print `result.ManifestPath` afterwards, which mitigates it, but `writePluginsToggleHelp` describes only what `--user` does and not what the default does. One extra line there would remove the ambiguity.
- **No test asserts unrelated manifest fields survive.** — `internal/plugins/enabled_test.go`. The four tests cover toggling, idempotence, the `--user` filter, the missing-plugin error and file mode. None reads the manifest back and checks that `name`, `version`, `commands` and the permission block are byte-identical apart from `enabled`. Given that the implementation round-trips the whole document through a map, that is the assertion most worth having — and it is the one that would have caught finding 3.

## Not covered by this PR

- Nothing invalidates a running daemon's or TUI's in-memory plugin set after a toggle. `zero plugins disable X` updates the manifest; whether an already-running session notices is untested and probably a separate issue rather than this PR's job — but worth confirming the expectation before merge.

## Not verified

- Windows. `TestSetEnabledPreservesManifestMode` is skipped there (noted in the PR description, which is the right call), but the rename path is the part that actually diverges and it has no coverage on that platform either way. I did not cross-compile or run the Windows path.
- I did not exercise `zero plugins enable/disable` end-to-end against a real installed plugin; the verification above is through the package and CLI tests.

The CLI layer itself I have no complaints about: flag parsing, the `--json` path through `redaction.RedactValue`, the `ErrNotInstalled` mapping to a usage error rather than a crash, and the already-enabled/now-enabled distinction are all handled properly. Rework the write in `enabled.go` around `installtxn` and this becomes an approve.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto current `main` and re-run the affected checks
  `docs/EXTENDING.md:290`
  The PR head is based on `d9b882e`, while live `main` is now `1b5db17` (656 files and two releases ahead) and changes both `docs/EXTENDING.md` and `internal/cli/extensions_test.go`. The tree merge is mechanically clean, but the repository's current contribution guidance makes a stale base a hard blocker. Please rebase, retain the current-target changes, and revalidate the resolved diff.

## Findings

- [P1] Serialize and durably publish the manifest mutation
  `internal/plugins/enabled.go:43`
  `SetEnabledByID` resolves and reads a manifest, then `SetEnabled` publishes it without the `installtxn.Lock` used by `Install` and `Remove`. The operation therefore does not serialize its read-modify-write cycle with the other commands that own the same user-plugin directory and `plugins.lock`. For example: a toggle reads the old manifest; an install acquires the lock, swaps in a newly staged plugin and writes its new hash; then the toggle writes its already-read JSON to the now-replaced path. The new plugin's manifest is overwritten by the old one while the lock entry still describes the new tree. The hand-rolled temporary-file path also omits the `Sync` and Windows write-through behavior in `installtxn.WriteFileAtomically`, so an interruption can publish incomplete manifest bytes and make the plugin fail JSON discovery.

  Address the root cause by making enablement participate in the existing transaction boundary: derive the selected plugin's root, acquire its `installtxn.Lock` before the final lookup/read, and publish via `installtxn.WriteFileAtomically` while that lock is held. Keep the lock scope limited to the established plugin root and retain the current precedence behavior. Add a deterministic concurrency test that pauses an install between staging and commit, then proves a simultaneous toggle cannot overwrite the installed manifest or desynchronize its lock entry; also cover a writer failure/interruption path.

- [P1] Do not corrupt extension-defined integer values while toggling
  `internal/plugins/enabled.go:57`
  The loader deliberately accepts unknown manifest fields, but this command decodes the entire document into `map[string]any`. Go's default JSON decoder represents each number there as `float64`; when the map is re-marshaled, integers outside IEEE-754's exact range are rounded. A valid extension value such as `9007199254740993` is consequently written back as `9007199254740992` when the user only asks to disable the plugin. This permanently changes configuration the command neither owns nor validates.

  The root cause is a lossy whole-document round trip. Preserve numeric tokens for every untouched field—at minimum decode with `json.Decoder.UseNumber` before changing `enabled`, rather than `json.Unmarshal` into `any`; a targeted representation-preserving edit is also acceptable if it preserves the manifest's existing schema behavior. Add a regression test containing an unknown nested field with `9007199254740993`, toggle both directions, and assert that the value remains exactly the same after each write.

Vasanthdev2004:
@Ayush7614 checking in on this one. I approved it at `7d55f568` and it has not moved since 2026-08-28. It is still mergeable against current main and its checks are green, so as far as I can tell it is only waiting on a second maintainer rather than on you.

Saying so explicitly in case it looked like the ball was in your court. If you are still around for it, nothing is needed from you; I will chase the merge side. If you have moved on, tell me and I will note the issue as open for someone else rather than leave it looking claimed.

Same question on #693, which is in the same state.

PR #696, opened 2026-07-15
feat(skills): discover shared ~/.agents/skills with multi-root skill loading

### LINKED ISSUE #610: feat: support ~/.agents/skills/ as a generic multi-agent global skills path
### Problem

Zero currently discovers skills from a **single root** — either `$ZERO_SKILLS_DIR`, `$XDG_DATA_HOME/zero/skills`, or `~/.local/share/zero/skills` (see `internal/skills/skills.go:DefaultDir()`). This is Zero-specific and works well when the user is inside a Zero session.

However, there is no standard **cross-agent** global skills directory. A user who runs multiple agent tools (Zero, Hermes Agent, Claude Code, OpenCode, etc.) has no shared place to put reusable skills/instructions that apply regardless of which agent they are using.

The `~/.agents/` convention is emerging as a de facto standard for multi-agent configuration. Several agent tools already recognize `~/.agents/AGENTS.md` as a global instruction file. Adding `~/.agents/skills/` as an additional global skills root would let the same skill files be shared across tools without duplication.

### Requested behavior

Add `~/.agents/skills/` as an **additional, lower-priority** global skills directory alongside the existing Zero-specific skills root:

| Layer | Path | Priority |
|---|---|---|
| User explicit override | `$ZERO_SKILLS_DIR` | Highest (already works) |
| Zero-specific default | `$XDG_DATA_HOME/zero/skills` / `~/.local/share/zero/skills` | Middle (already works) |
| **Multi-agent generic** | **`~/.agents/skills/`** | Lowest (NEW) |

When loading skills, Zero should scan **both** its Zero-specific directory AND `~/.agents/skills/` (if it exists), merging the results. In case of duplicate skill names, the Zero-specific directory wins (higher priority).

### Why it belongs in Zero

1. **Interoperability** — Zero is part of a growing ecosystem of agent coding tools. A shared `~/.agents/skills/` convention means users can define their preferred workflows, conventions, and gotchas once and use them across Zero, Hermes Agent, Claude Code, and other tools.

2. **Multi-agent workflows** — when running multiple agent tools side by side, a shared skills directory reduces duplication and keeps agent knowledge in sync.

3. **Community convention** — `~/.agents/` is becoming a recognized standard (similar to how `~/.config/` emerged as the XDG convention). Early adoption positions Zero as a good citizen in the agent ecosystem.

4. **Minimal code change** — the existing `internal/skills/skills.go` `Load()` function already supports scanning one root. The change would be to call `Load()` on both roots and merge, or modify the loader to accept multiple roots.

### Implementation sketch

```go
// MultiRootLoad scans multiple skill directories and returns merged results.
// Directories earlier in the list take priority for duplicate skill names.
func MultiRootLoad(dirs []string) ([]Skill, error) {
    seen := map[string]bool{}
    var all []Skill
    for _, dir := range dirs {
        loaded, err := Load(dir)
        if err != nil {
            continue // skip missing/unreadable dirs
        }
        for _, s := range loaded {
            if seen[s.Name] {
                continue // earlier dir wins
            }
            seen[s.Name] = true
            all = append(all, s)
        }
    }
    sort.Slice(all, func(i, j int) bool { return all[i].Name < all[j].Name })
    return all, nil
}

// agentsSkillDir returns ~/.agents/skills/ if it exists
func agentsSkillDir() string {
    home, _ := os.UserHomeDir()
    if home == "" {
        return ""
    }
    dir := filepath.Join(home, ".agents", "skills")
    if info, err := os.Stat(dir); err == nil && info.IsDir() {
        return dir
    }
    return ""
}
```

### Alternatives considered

1. **Single `~/.agents/AGENTS.md`** — simpler but not composable (one file for everything).
2. **Symlink farm** — user could symlink `~/.local/share/zero/skills/*` into `~/.agents/skills/`, but that is manual and fragile.
3. **Config key** in `zero.toml` — flexible but no convention means every tool picks a different path.
4. **Hardcoded merge with plugin-declared skills** — related but separate concern (plugins / skills are tracked in the existing "plugin-declared skills" issue).

The `~/.agents/skills/` convention is the multi-agent ecosystem solution, not a Zero-specific config knob.

### Additional context

This follows the same pattern as how Hermes Agent uses `~/.hermes/skills/` for its own skills, and the broader ecosystem is converging on `~/.agents/AGENTS.md` as a global rules file. See related discussions:

- Hermes Agent skills: `~/.hermes/skills/` convention (per-agent)
- Claude Code: `.claude/rules/` / global `~/.claude/rules/` (per-agent)
- `~/.agents/AGENTS.md` emerging as a cross-agent standard for global instructions

lfaoro:
## Summary
Add `~/.agents/skills` as a read-only global skills discovery root and unify CLI and runtime skill loading on one multi-root path.

**Discovery (first wins by name):**
1. Primary Zero skills dir (`ZERO_SKILLS_DIR` | XDG/home)
2. `~/.agents/skills` (if present)
3. Plugin skill roots
**Write root unchanged:** `zero skills install` / `remove` / lock still target only the primary Zero skills directory.
This also collapses the previous half-wiring where the multi-root skill tool was 
only registered when plugins were active. Bare runs and plugin runs now share the same multi-root `skill` tool.

## Linked issue
Fixes #610

## Why
Shared `*/SKILL.md` packs are emerging as a multi-agent convention. Without this, users must duplicate skills into Zero’s own directory to use them alongside other agents (Hermes, Claude Code, etc.). 

Zero-specific skills still win on name conflicts.

## What changed
- `internal/skills`: `AgentsDir`, `DiscoveryRoots`, `GlobalRoots`, `LoadFromRoots
`, `ListFromRoots`, `InfoFromRoots`
- `internal/plugins`: `mergeSkills` becomes a thin wrapper over the skills multi-
root API
- CLI `skills list` / `info` use the same global discovery roots as runtime (minus plugins)

- Always register the multi-root skill tool at activation (even with 0 plugins)
- Tests for agents-dir discovery, first-wins priority, and write-root isolation
## Checklist
- [x] `gofmt` clean.

gg

## Summary by CodeRabbit

* **New Features**
  * Skills discovery now also includes the optional shared `~/.agents/skills` directory.
  * `zero skills list` and `zero skills info` search across the primary directory plus `~/.agents/skills` (and plugin roots), with first-root wins for name collisions and winner/loser warnings.
  * Shared skills remain available even when no plugins are loaded or plugin activation fails (install/remove still target only the primary directory).
* **Documentation**
  * Updated skills discovery/precedence and plugin contribution behavior in the extending/how-it-works docs.
* **Tests**
  * Added coverage for multi-root behavior and environment isolation for deterministic results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,7 +129,7 @@ tagged. Until then, source builds report the version `dev`.
 ### Added
 - Shared multi-agent skills discovery: when present, `~/.agents/skills` is searched after the primary
   Zero skills dir (and before plugin skill roots). `zero skills list` / `info` and the runtime `skill`
-  tool share one multi-root discovery path; install/remove still target only the Zero skills directory.
+  tool share one multi-root discovery path; install/remove/lock still target only the Zero skills directory.
 - `SECURITY.md` with a private vulnerability-reporting path, `CODE_OF_CONDUCT.md`, this changelog, and
   GitHub issue/PR templates.
 - Interactive `/theme` picker: bare `/theme` opens a popup that live-previews each palette as you move
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,9 @@ tagged. Until then, source builds report the version `dev`.
 ## [Unreleased]
 
 ### Added
+- Shared multi-agent skills discovery: when present, `~/.agents/skills` is searched after the primary
+  Zero skills dir (and before plugin skill roots). `zero skills list` / `info` and the runtime `skill`
+  tool share one multi-root discovery path; install/remove still target only the Zero skills directory.
 - `SECURITY.md` with a private vulnerability-reporting path, `CODE_OF_CONDUCT.md`, this changelog, and
   GitHub issue/PR templates.
 - Interactive `/theme` picker: bare `/theme` opens a popup that live-previews each palette as you move
--- a/docs/EXTENDING.md
+++ b/docs/EXTENDING.md
@@ -108,10 +108,10 @@ Discovery roots (earlier wins on name collisions):
 2. **Shared multi-agent dir** — `~/.agents/skills/` when present (read-only discovery; never an install target)
 3. **Plugin skill roots** — skills bundled by active plugins (section 6)
 
-A missing directory is fine — Zero just omits it. Management commands (`zero skills add` / `remove`) always write to the primary Zero dir only; `list` / `info` search primary + `~/.agents/skills`.
+A missing directory is fine — Zero just omits it. Management commands (`zero skills add` / `remove` / `lock`) always write to the primary Zero dir only; `list` / `info` search primary + `~/.agents/skills`.
 
 ```text
-~/.local/share/zero/skills/          # primary (install/remove target)
+~/.local/share/zero/skills/          # primary (install/remove/lock target)
   run-benchmarks/
     SKILL.md
   write-changelog/
--- a/docs/EXTENDING.md
+++ b/docs/EXTENDING.md
@@ -102,13 +102,23 @@ The full format spec (frontmatter fields, tool scopes, prompt conventions) is in
 
 Skills are markdown instruction packs the agent can pull in on demand. Each skill is a directory containing a `SKILL.md`. Standalone project skill directories are not supported in this version (shared project-wide skills must go in `AGENTS.md` or as a hook). However, project plugins (section 6) may bundle skills, which are merged into the active run.
 
-Discovery root: `$ZERO_SKILLS_DIR` → `$XDG_DATA_HOME/zero/skills` → `~/.local/share/zero/skills/`. A missing directory is fine — Zero just reports "no skills".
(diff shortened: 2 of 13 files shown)


Vasanthdev2004 (approved):
Verified on the PR branch: build/vet/gofmt clean and the new skills/plugins/cli skills tests pass; the only cli failures are the pre-existing "provider anthropic requires model" ones, which reproduce identically on clean origin/main. The merge is sound  per-root symlink confinement keeps a hostile ~/.agents/skills entry from escaping its root or picking up a Zero lock entry (filepath.Rel guard in InfoFromRoots), optional roots fail open, and primary non-directory errors now bubble consistently across Windows and Unix. Minor heads-up: ~/.agents/skills is auto-discovered by the permission-allow skill tool with no prompt, so skills placed there by other tools load without asking  that's the intended feature and it's read-only/confined, just calling out the trust-surface expansion. Approving.

kevincodex1 (approved):
LGTM

PR #700, opened 2026-07-16
Add per-turn tracing and a turn-benchmark harness (Phase 0 baseline)

Vasanthdev2004:
## What this does

Establishes the Phase 0 performance baseline: where does a turn actually spend time, recorded repeatably so later work is measurable instead of guessed.

It adds an opt-in per-turn tracer (`internal/trace`) that attributes wall time to named spans — tool partition, provider connect, provider queue, generation, tool execution, permission wait, verification, compaction — plus counters for model requests, tool calls, retries, reconnects, compactions, completion nudges, acceptance checks, model switches, and input/cached/output tokens. The recorder is nil-safe, so every stamp site calls it unguarded and a run without tracing is byte-identical to today.

Spans record wall intervals, not raw durations. At finish the recorder derives each span's parent by interval containment and its **exclusive** time (duration minus the union of its children), so a `provider_connect` running concurrently inside `generation` or a `permission_wait` nested inside `tool_execution` no longer double-counts the same wall. Top-latency ranking uses exclusive time, so shares of the top sources sum to ~1; **coverage** (union of span intervals over wall, capped at 1) is the honest "≥95% of wall accounted for" metric.

The recorder threads into `agent.Run` through a new `Options.Trace` field and one `trace.WithContext` call, so the shared providerio seam (connect around `client.Do`, queue around the OAuth resolve, first token at the first non-keepalive SSE payload) reaches it via `trace.FromContext` with no provider-interface change. Headless `zero exec` gets `--trace <path>` (or `ZERO_TRACE=<path>`, `-` for stderr), emitting an NDJSON trace that matches the existing agenteval trace contract.

[...]

PR #704, opened 2026-07-16
feat(perf): emit prompt-prefix hash fingerprint per turn

anandh8x:
# What this does

Adds a per-turn prompt-prefix hash fingerprint to the trace pipeline so a downstream observer can detect which sub-component of the cacheable prompt prefix drifted between turns. Seven independent SHA-256 hashes (base instructions, confirmation policy, project context, skills, tools, tool schemas, complete) plus a `complete_prefix` aggregate (SHA-256 of the canonical join of the other six). All hashes are hex-encoded and emitted as a `prefix_hash` event in the NDJSON trace.

The decomposition exists so a stable `complete_prefix` across turns means the cacheable prefix is byte-identical, and a changed `complete_prefix` names the sub-component that broke cache stability (the sub-hash that changed). No behavior change for untraced runs. For OpenAI, the existing `prompt_cache_key` (session ID) is unchanged — the fingerprint is purely additive observability. The Anthropic `cache_control` breakpoint work in main is not touched.

# Why now

PR #700 (the Phase 0 tracer) just landed, which established the `trace.Recorder` API and the NDJSON contract. This PR is the first cache-relevant instrumentation on top of that contract. The decision gate from the split work — *"check whether provider-reported cached-input tokens improve"* — becomes answerable once the trace exposes the seven sub-hashes, because a downstream consumer can correlate a `cached_input_tokens` movement with a `prefix_hash` event to determine which sub-component is or isn't cacheable for a given model and provider.

# Change summary

| file | change |
|---|---|
| `internal/trace/trace.go` | new `PrefixHash` event type; `TurnTrace` carries `[]PrefixHash`; `"event:prefix_hash"` added to `OptionalEventKeys()` |
| `internal/trace/recorder.go` | new `EmitPrefixHash(p PrefixHash)` method, nil-safe, sealed-after-`Finish` |
| `internal/trace/emit.go` | `WriteNDJSON` emits one `prefix_hash` line per fingerprint after counters, sorted by `complete_prefix` for stable output |
| `internal/agent/prompt_fingerprint.go` | new file: `buildPromptSubstrings`, `toolSubstrings`, `canonicalSchemaString` (json.Marshal for stable map rendering), `ComputePrefixFingerprint` |
| `internal/agent/loop.go` | after `partitionToolsCached` returns, if `options.Trace != nil`, compute the fingerprint and emit it |
| `internal/agent/prompt_fingerprint_test.go` | new file: 7 tests covering stability, name-order independence, schema sensitivity, system-prompt sensitivity, complete-is-aggregate, default-options invariant, canonical-schema-string stability for map parameters |

# How to verify

- `go test -race ./internal/trace/... ./internal/agent/...` — green
- `gofmt -l` — clean
- `go vet ./internal/...` — clean
- `golangci-lint run --enable-only unused,ineffassign,staticcheck ./internal/trace/... ./internal/agent/...` — 0 issues
- `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./internal/trace/... ./internal/agent/...` — no vulnerabilities

Run an end-to-end with a real provider:

```
ZERO_TRACE=/tmp/run.ndjson zero exec --prompt "list the files in this repo"
```

Then inspect the trace:

```
jq -c 'select(.type == "prefix_hash")' /tmp/run.ndjson
```

For a multi-turn session, the `complete_prefix` should be stable across the first turn's hash if no project doc or skill list changed.

# Known limitations and follow-ups

These are deliberate scope limits of this PR, not omissions. They are listed so the next PR can pick them up cleanly.

1. **Only 4 of 10 sections of `buildSystemPrompt` are in the fingerprint.** The captured substrings are `baseInstructions`, `confirmationPolicy`, `projectContext` (workspace context), and `skills`. Sections that fire only for non-default `Options` (specialist delegation context, response style, session runtime context, model addendum, approved command prefixes, user guidelines, workspace seed) are out of scope. For default `Options` (the common case), the four captured substrings cover the full prompt and the fingerprint is accurate. A follow-up PR that extends the decomposition to all 10 sections — and a test asserting the full join matches `buildSystemPrompt` byte-for-byte for any `Options` — is the natural next step.
2. **No constructor validation on `trace.PrefixHash`.** A future caller could put a non-SHA-256 string in a `BaseInstructionsHash` field and the trace would silently accept it. A `NewPrefixHash(...)` constructor that does the hashing in one place is a small follow-up.
3. **Sort by `complete_prefix` in `WriteNDJSON`.** Deterministic but not informative — the order changes as the run progresses and is not a stable turn index. Switching to insertion order (no sort) or attaching a `turn` field to the event is a small follow-up.
4. **No instruction-cache benefit measured yet.** This PR makes the cache hit rate *observable* via the trace. The actual measurement — does the provider's `cached_input_tokens` move when `complete_prefix` is stable? — requires a real provider run and is the M1-baseline-evidence work, not a code change.

# Risk and rollback

Pure observability. No provider call, no model call, no request shape change. If the `prefix_hash` event proves noisy, the schema bump is `TurnTrace.PrefixHashes` (omitempty) and the new event type in `OptionalEventKeys()` — both backward-compatible drops. Rolling back is `git revert e503ad29`.

## Summary by CodeRabbit

* **New Features**
  * Enhanced trace output with per-turn prompt-prefix fingerprinting, capturing which cacheable request components shaped the generated prompt.
  * Trace output (NDJSON) now includes additional `prefix_hash` events containing per-component hashes plus the derived `complete_prefix` hash.

* **Tests**
  * Added coverage to verify fingerprint determinism across repeated calls.
  * Added checks that tool set changes affect hashes, tool ordering does not, schema changes update schema/hash fields, and system prompt changes update base hashes.
  * Updated trace round-trip tests to include `prefix_hash` events and preserve insertion order.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -213,6 +213,26 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		exposed, _ := partitionToolsCached(registry, permissionMode, options, loaded, toolDefCache)
 		toolPartitionSpan.End()
 
+		// Prompt-prefix fingerprint: hash the seven cacheable sub-components
+		// of this turn's request (system prompt sections, project context,
+		// skills, partitioned tool list, tool schemas) and emit one trace
+		// event. Stable across turns => cacheable. Drift in any sub-hash
+		// names the sub-component that broke the cache for this turn. The
+		// computation is pure and cheap (seven SHA-256s of small strings)
+		// and is a no-op when tracing is off.
+		if options.Trace != nil {
+			fp := ComputePrefixFingerprint(options, exposed)
+			options.Trace.EmitPrefixHash(trace.PrefixHash{
+				BaseInstructionsHash:   fp.BaseInstructionsHash,
+				ConfirmationPolicyHash: fp.ConfirmationPolicyHash,
+				ProjectContextHash:     fp.ProjectContextHash,
+				SkillsHash:             fp.SkillsHash,
+				ToolsHash:              fp.ToolsHash,
+				SchemaHash:             fp.SchemaHash,
+				CompletePrefixHash:     fp.CompletePrefixHash,
+			})
+		}
+
 		// PROACTIVE compaction: if the history is approaching the model's
 		// context window, summarize the oldest middle before building the
 		// request. A no-op when ContextWindow == 0 (compaction disabled).
--- a/internal/agent/prompt_fingerprint.go
+++ b/internal/agent/prompt_fingerprint.go
@@ -1,0 +1,300 @@
+package agent
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// promptSubstrings are the seven cacheable sub-components of the prompt a run
+// sends to a model. Each is the literal content the sub-component would emit
+// into the system prompt or tool list, hashed so a downstream observer can
+// detect drift turn-over-turn. The substrings are produced by
+// buildPromptSubstrings and consumed by ComputePrefixFingerprint.
+type promptSubstrings struct {
+	baseInstructions   string
+	confirmationPolicy string
+	projectContext     string
+	skills             string
+	tools              string
+	schema             string
+}
+
+// buildPromptSubstrings assembles the seven cacheable sub-components of the
+// prompt without joining them. The system prompt sections are produced by the
+// same private builders that buildSystemPrompt uses, so the substrings are
+// byte-identical to what would have appeared in the joined prompt. The tools
+// and schema substrings are derived from the partitioned tool list (caller-
+// provided) so this helper does not need to re-run the partition.
+//
+// This is the seam ComputePrefixFingerprint reads from. It exists as a
+// separate function (rather than returning both prompt and substrings from
+// buildSystemPrompt) so existing callers of buildSystemPrompt are unaffected
+// and the substrings helper is independently testable.
+func buildPromptSubstrings(options Options, exposed []zeroruntime.ToolDefinition) promptSubstrings {
+	core := strings.TrimSpace(options.SystemPrompt)
+	if core == "" {
+		core = strings.TrimSpace(coreSystemPrompt)
+	}
+	if core == "" {
+		core = fallbackSystemPrompt
+	}
+
+	// The confirmation policy is appended unconditionally in buildSystemPrompt;
+	// the substring here is the same constant.
+	policy := strings.TrimSpace(confirmationPolicy)
+
+	// projectContext is the joined output of workspaceContext, which itself
+	// walks the AGENTS.md / ZERO.md / .zero/AGENTS.md chain. Hashing the
+	// joined string captures every guideline file's contribution and the
+	// ordering between them.
+	project := workspaceContext(options.Cwd)
+
+	// skills is the joined output of skillsContext, with each skill on its own
+	// line. The substring is the literal text the model would see.
+	skills := skillsContext(options)
+
+	// tools and schema are derived from the partitioned tool list. The list
+	// passed in is already in stable order (partitionToolsCached is the
+	// canonical stable partitioner) so concatenating in slice order is
+	// sufficient — we sort names again defensively in case a future caller
+	// hands us a list from a different partitioner.
+	toolsSubstr, schemaSubstr := toolSubstrings(exposed)
+
+	return promptSubstrings{
+		baseInstructions:   core,
+		confirmationPolicy: policy,
+		projectContext:     project,
+		skills:             skills,
+		tools:              toolsSubstr,
+		schema:             schemaSubstr,
+	}
+}
+
+// toolSubstrings extracts a stable, name-sorted digest of the partitioned tool
+// list. The tools substring is "name1\nname2\n..." (sorted) so a reordering of
+// the partition produces a different hash. The schema substring is the
+// concatenation of each tool's JSON schema in name order, so a schema change
+// is detected independently of the tool ordering.
+func toolSubstrings(exposed []zeroruntime.ToolDefinition) (toolsSubstr, schemaSubstr string) {
+	if len(exposed) == 0 {
+		return "", ""
+	}
+	names := make([]string, 0, len(exposed))
+	byName := make(map[string]zeroruntime.ToolDefinition, len(exposed))
+	for _, def := range exposed {
+		names = append(names, def.Name)
+		byName[def.Name] = def
+	}
+	sort.Strings(names)
+	var toolsSB, schemaSB strings.Builder
+	for i, name := range names {
+		if i > 0 {
+			toolsSB.WriteByte('\n')
+		}
+		toolsSB.WriteString(name)
+		def := byName[name]
+		// Parameters is rendered to a canonical JSON string per tool. The
+		// schema-render cache in internal/agent (used by partitionToolsCached)
+		// guarantees the same render is byte-identical across turns for the
+		// same tool, so this substring is a stable hash input.
+		schemaSB.WriteString(name)
+		schemaSB.WriteByte('\n')
+		schemaSB.WriteString(canonicalSchemaString(def.Parameters))
+		schemaSB.WriteByte('\n')
+	}
+	return toolsSB.String(), schemaSB.String()
+}
+
+// canonicalSchemaString renders a tool's parameter schema to a stable string.
+// tools.ToolDef.Parameters is a map[string]any in the common case; Go's
+// fmt.Sprintf("%v", m) iterates a map in random order, which would produce
+// a different hash for the same Parameters value across calls and defeat
+// the fingerprint. encoding/json marshals maps with keys sorted
+// alphabetically, so json.Marshal is the primary stable render. The
+// fallback for the rare non-JSON-compatible value (functions, channels,
+// NaN/Inf floats, cyclic references) is a stable key-sorted stringifier
+// that walks the value with sorted keys for maps and a leading
+// "__non_json:" prefix so a future schema change is visible in the trace
+// (a SchemaHash collision from the same value) rather than a silent hash
+// drift.
+//
+// Note: SchemaHash is a stability signal, not a wire-identity signal. The
+// bytes json.Marshal produces for a Go map[string]any may differ from
+// the bytes the provider actually sends on the wire (provider encoders
+// use their own JSON conventions; an Anthropic schema may serialize with
+// different whitespace, key order, or number formatting). A consumer
+// must NOT assume SchemaHash matches the provider's wire schema — only
+// that two turns with the same Go-side Parameters produce the same
+// SchemaHash. This is sufficient for the trace's purpose (drift
+// detection) but not for a content-based equality check.
+func canonicalSchemaString(params map[string]any) string {
+	if len(params) == 0 {
+		return ""
+	}
+	data, err := json.Marshal(params)
+	if err != nil {
+		return "__non_json:" + stableStringify(params)
+	}
+	return string(data)
+}
+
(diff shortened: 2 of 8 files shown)


Vasanthdev2004:
@coderabbitai full review

Vasanthdev2004 (changes requested):
Requesting changes — but only on contract-doc accuracy, not the implementation. The code is sound and the agent→trace seam is clean; the issue is that the in-code docs advertise a stronger invariant than the 4-of-11 coverage actually delivers, and for a trace event whose whole purpose is "don't let this be misread," that gap is worth closing before merge.

**Decision up front:** qualify the `complete_prefix` invariant in the type/function docs (items 1–2 below). Everything else is a suggestion or a nit. Land 1 and 2 and I'm an approve.

**1. (must fix) `complete_prefix` stable ≠ full prefix byte-identical.** `buildSystemPrompt` is 11 sections; the fingerprint covers 4 (core, confirmationPolicy, workspaceContext, skillsContext). Seven aren't hashed: `modelPromptAddendum`, `sessionRuntimeContext`, `approvedCommandPrefixContext`, `workspaceSeedContext`, `userGuidelines`, `specialistDelegationContext`, `responseStyleContext`. The `PrefixHash` doc and `ComputePrefixFingerprint` doc both say "stable CompletePrefixHash ⇒ cacheable prefix byte-identical" unconditionally. That's false the moment any of the seven changes — and the sharpest case is `modelPromptAddendum(options.Model)`, which moves on a model switch, and model switches are themselves a traced counter. So a consumer could read a stable `complete_prefix` right through a model switch that actually changed the prefix. You call this out in Known-limitations #1, and I read it — the ask is only that the *type/function docs* (what a consumer reads, not the PR body) carry the same qualification: `complete_prefix` reflects the captured sections; a stable value means those are byte-identical, not the full prefix. A test asserting an edit to an uncaptured section (e.g. `responseStyle`) does NOT move `complete_prefix` would make the boundary a tested contract, not just a README note.

**2. (must fix) "does not read files" is wrong.** `ComputePrefixFingerprint`'s doc says it "does not read files." It does: `buildPromptSubstrings` → `workspaceContext` runs git (`gitBranchForPrompt`, `FindProjectGitRoot`) and reads the AGENTS.md/ZERO.md chain (`projectGuidelines`) plus the repo map. Side effect: it re-does the workspace-context work `buildSystemPrompt` does a few lines later, on every traced turn. Fine for an opt-in trace, but the doc should say so. Either correct the doc, or have the request-build path hand the fingerprint the substrings it already computed so we don't read/git twice per turn.

**3. (suggest, one line) drop the `complete_prefix` sort in `WriteNDJSON`.** The PR's reason for existing is correlating a `cached_input_tokens` move with the turn whose prefix drifted. The `prefix_hash` event has no turn index and you sort by `complete_prefix`, so turn order is gone — the correlation is ambiguous. Known-limitations #3 calls the sort cosmetic; I'd argue it breaks the one use case the event is for. Insertion order is turn order and is already deterministic; dropping the sort is one line and makes the event actually usable for the correlation. I'd take it here rather than as a follow-up, but I won't block on it.

**4. (nit)** `toolSubstrings`'s comment justifies schema stability via a "schema-render cache," but `canonicalSchemaString` uses `json.Marshal` directly (sorted keys = stable on its own), so the cache isn't what makes it stable. Also `SchemaHash` is `json.Marshal(def.Parameters)`, which may not match the bytes a provider actually sends — so it's a stability signal, not a wire-identity signal. A one-line note keeps a consumer from assuming SchemaHash == wire schema.

**What's good:** nil-safe `EmitPrefixHash` matching the recorder pattern, sealed-after-Finish, the agent→trace boundary (trace doesn't import agent), `options.Trace != nil` gating, `OptionalEventKeys` placement (optional, so a short healthy trace doesn't false-fail), and the test coverage (stability / reorder / schema / system-prompt / aggregate / default-options / canonical-schema). CI green across all 9 checks. The split-plan seam is clean.

gnanam1990 (changes requested):
## Review — CHANGES_REQUESTED

Gate: @Vasanthdev2004 has a formal review on this PR.

### Local verification
| Check | Result |
|--------|--------|
| go test -race ./internal/trace/ ./internal/agent/ | pass |
| gofmt / go vet on touched packages | clean |
| CI | green |

### Summary
Solid Stream B / Phase-1 instrumentation on the Phase 0 tracer: opt-in, pure observability, stable schema rendering via json.Marshal, good tests. Untraced path is a no-op. Partial buildSystemPrompt coverage is OK if documented loudly.

### Must fix
1. **Finish() must deep-copy PrefixHashes** (same contract as Spans/Counters) in `internal/trace/recorder.go`. Without it the snapshot aliases the recorder slice. CodeRabbit flagged the same issue.

```go
snap.PrefixHashes = append([]PrefixHash(nil), r.tr.PrefixHashes...)
```

2. **ReadNDJSON must parse `type: "prefix_hash"`** into `TurnTrace.PrefixHashes` in `internal/trace/parse.go`. Write path emits it; re-parse currently drops the signal (harness / offline analysis).

### Non-blocking
- complete_prefix is a partial prefix aggregate, not the full system prompt.
- Sorting prefix_hash events by complete_prefix is non-chronological for multi-turn.
- Rename TestComputePrefixFingerprintToolReorderChangesToolsHash (asserts order-independence).

### Verdict: **CHANGES_REQUESTED**

Approve after the Finish copy + NDJSON parse fixes.

anandh8x:
All review feedback addressed in the latest push (`c40a25f7`). Summary of changes for re-review:

**From Vasanth (must fix):**
1. **In-code docs qualified.** `PrefixHash` and `ComputePrefixFingerprint` no longer claim "stable CompletePrefix ⇒ byte-identical cacheable prefix." They now state the fingerprint covers 4 of 11 sections, name the seven uncaptured sections, and explain that `modelPromptAddendum` changes on a model switch — so a consumer correlating CompletePrefix stability with `cached_input_tokens` must cross-check the `model_switches` counter.
2. **"Does not read files" claim removed.** The doc now states the I/O side effects explicitly (workspace context runs git and reads AGENTS.md/ZERO.md + repo map) and notes that the request-build path performs the same work separately. A follow-up to pass substrings through is documented as a known limitation.

**From Vasanth (suggest + nit):**
3. **Sort removed from WriteNDJSON.** Prefix fingerprints now emit in insertion (turn) order, so a downstream consumer can correlate each event with the `cached_input_tokens` counter for that turn. The defensive copy that the sort required is also gone (the slice is already a deep copy from `Finish`).
4. **canonicalSchemaString docstring expanded** with a "stability signal, not wire-identity signal" note — SchemaHash is a stability signal over the Go-side Parameters, not a byte-for-byte match of what the provider sends on the wire.

**From gnanam (must fix):**
5. **Finish() deep-copies PrefixHashes.** Added `snap.PrefixHashes = append([]PrefixHash(nil), r.tr.PrefixHashes...)` matching the existing Spans/Counters contract.
6. **ReadNDJSON parses `type: "prefix_hash"`.** The seven sub-hash fields are now decoded back into `TurnTrace.PrefixHashes`. The `no recoverable spans, counters, or prefix hashes` guard was updated to include the new event type. A new `stringField` helper handles missing-or-non-string values without fataling the parse.
7. **Round-trip test extended.** `TestReadNDJSONRoundTrip` now emits two `prefix_hash` events, parses them back, and asserts: count, insertion order preserved, all seven sub-hash fields per event preserved exactly.

**From both:**
8. **Test renamed.** `TestComputePrefixFingerprintToolReorderChangesToolsHash` → `TestComputePrefixFingerprintToolsHashIsNameSetSensitive`, with a doc comment that explains the assertion direction (the test name was lying — it asserts name-order independence, not name-order sensitivity).

**All checks green:**
- `gofmt -l` clean
- `go vet` clean
- `go build ./...` clean
- `golangci-lint` (unused, ineffassign, staticcheck): 0 issues
- `go test -race -count=1 ./internal/trace/... ./internal/agent/...` green
- `govulncheck`: not re-run (same dependency surface, no new imports)

Ready for re-review at HEAD `c40a25f7`.

gnanam1990 (approved):
## Re-review at c40a25f — APPROVE

Gate: Vasanthdev2004 has formal review activity on this PR.

### Local verification
| Check | Result |
|--------|--------|
| go test -race ./internal/trace/ ./internal/agent/ | pass |
| TestReadNDJSONRoundTrip / Finish snapshot tests | pass |
| Prior must-fix items | addressed (see below) |

### Prior gnanam must-fix items
| Item | Status |
|------|--------|
| Finish() deep-copy PrefixHashes | **Fixed** |
| ReadNDJSON parse type:prefix_hash | **Fixed** (+ round-trip coverage) |

### Also addressed (aligned with Vasanth)
| Item | Status |
|------|--------|
| Doc: complete_prefix stable ≠ full prompt | **Fixed** on PrefixHash + ComputePrefixFingerprint |
| Doc: "does not read files" was wrong | **Fixed** — I/O side effects documented |
| Drop sort-by-complete_prefix in WriteNDJSON | **Fixed** — insertion/turn order |

### Residual (non-blocking)
- Optional test that uncaptured sections (e.g. response style) do not move complete_prefix would lock the contract harder — nice-to-have.
- Dual workspaceContext I/O on traced turns still intentional; follow-up can share substrings with buildSystemPrompt.

### Verdict: **APPROVE**

Clean observability PR. My previous CHANGES_REQUESTED items are resolved. Happy to land.

anandh8x:
CodeRabbit review at #pullrequestreview-4715021444 — second finding (the `canonicalSchemaString` fallback non-determinism) — is now fixed in the latest push (`0fa1d991`).

**What was wrong:** the fallback path for non-JSON-encodable values used `fmt.Sprintf("%v", params)`, which iterates the map in random order. The same Parameters value could produce a different SchemaHash across calls — defeating the entire fingerprint.

**What I did:**

- Replaced the `fmt.Sprintf` fallback with a small `stableStringify` helper plus a recursive `writeStable` that handles `nil`, `bool`, `string`, `float{32,64}`, signed/unsigned ints, `map[string]any` (keys sorted alphabetically), `[]any` (slice order preserved), and a `<%T:%v>` last-resort for anything it doesn't know.
- The fallback still leads with `__non_json:` so a consumer can tell the bytes are not a JSON-marshaled schema.
- Added `TestCanonicalSchemaStringFallbackStableForNonJSONValue` that constructs a `Parameters` value containing a channel (which `json.Marshal` can't encode), calls `canonicalSchemaString` twice, and asserts the output is byte-identical. Also asserts that a re-ordered map produces the same string (the key-sorted property), and that the output starts with the `__non_json:` prefix.

The first finding in that review (deep-copy `PrefixHashes` in `Finish()`) was already fixed in the prior push (`c40a25f7`).

All checks green: gofmt, go vet, go test -race, golangci-lint (0 issues).

Vasanthdev2004 (approved):
All four addressed — approving. CI green across the board and gnanam's already approved too.

- **#1 (complete_prefix overclaim):** both the `PrefixHash` doc and `ComputePrefixFingerprint` doc now scope to the 4 captured sections, name the seven uncaptured ones, and call out the `model_switches` cross-check for `modelPromptAddendum`. That's exactly the qualification the event needed.
- **#2 ("does not read files"):** replaced with an honest I/O-side-effects note — `workspaceContext` runs git and reads the guideline chain, doubled per traced turn, follow-up to pass substrings through. Good.
- **#3 (sort):** dropped — insertion order is turn order, with a clear comment on why sorting would break the `cached_input_tokens` correlation. Taking it now rather than as a follow-up, thanks.
- **#4 (nit):** `canonicalSchemaString` now notes SchemaHash is a stability signal, not a wire-identity signal. Fine.

One non-blocking follow-up I'll leave here rather than hold the PR on: a test asserting an edit to an uncaptured section (e.g. `responseStyle`) does **not** move `complete_prefix` would pin the 4-of-11 boundary as a tested contract, so a future extension — or an accidental change to the captured set — is caught. Not blocking; the docs carry the scope clearly now.

Nice cleanup on the `…ReorderChangesToolsHash` → `…IsNameSetSensitive` rename too — the old name asserted the opposite of what it read.

kevincodex1 (approved):
LGTM

PR #705, opened 2026-07-16
perf(tools): add explicit effect metadata for safe concurrency

gnanam1990:
## Summary
Introduces explicit effect and thread-safety metadata for Zero's built-in tools.
This establishes the fail-closed safety contract required before read-only tool calls can be executed concurrently.
No concurrent execution is enabled by this PR.

## Motivation
Future tool batching must not infer safety from tool names or schemas.
Every tool now declares:
- its effect class,
- whether it is thread-safe,
- deterministic resource keys used to identify conflicts.

Unclassified, plugin, and MCP tools remain serialized by default.

## Effect classes
- Unknown
- ReadOnly
- ProcessRead
- WorkspaceWrite
- ExternalWrite
- Interactive

## Safety properties
- Unknown is the default.
- Unknown tools are never concurrency-eligible.
- Read-only does not automatically imply thread-safe.
- Existing permission, sandbox, redaction, cancellation, and output-ceiling paths are unchanged.
- Deferred wrappers preserve underlying metadata.
- Plugin and MCP tools remain fail-closed.

## Coverage
| Effect | Tool count |
|---|---:|
| ReadOnly | 10 |
| ProcessRead | 0 |
| WorkspaceWrite | 4 |
| ExternalWrite | 0 |
| Interactive | 20 |
| Unknown built-ins | 0 |

Intentionally Unknown (non-built-in categories): MCP tools, plugin tools, specialist tools registered outside `internal/tools` (e.g. Task).

## Tests
- Default Unknown / not thread-safe for tools without CapabilityProvider
- Every exposed built-in has explicit metadata (`ValidateBuiltinCatalog`)
- Deferred tools preserve capabilities after registry
- Plugin/MCP-style tools stay fail-closed
- File resource keys normalize paths; missing args do not panic
- Endpoint keys strip secrets/userinfo
- Workspace-write / interactive classifications
- Concurrent `Registry.Capabilities` lookup race test
- Schemas and Safety() unchanged for core tools

## Verification
- `go fmt ./...` — clean
- `go vet ./...` — clean
- `go test -race ./internal/tools/...` — pass
- `go test ./...` — pass except pre-existing `internal/cli` doctor tests also failing on `origin/main` (`TestRunDoctorFormatsRedactedProviderDiagnostics`, `TestRunDoctorConnectivityProbesProvider`)
- `make build` — pass
- golangci-lint unused/ineffassign/staticcheck on `./internal/tools/` — 0 issues
- govulncheck `./internal/tools/` — no vulnerabilities

## Performance impact
No performance change is expected.
This PR only introduces metadata and validation. Concurrent execution remains disabled until the follow-up PR and until metadata coverage reaches 100%.

## Follow-up
PR6 will consume this metadata to implement feature-gated concurrent read-only tool batches with deterministic result ordering.

Part of the Zero Agent Performance Program — Stream D.

## Summary by CodeRabbit

* **New Features**
  * Standardized tool capability classification (side effects + concurrency safety) across built-in tools.
  * Added deterministic resource-key tracking for files, sessions, processes, endpoints, patches, and workspace changes.
  * Introduced a built-in tool catalog with validation that flags missing/unknown or inconsistent metadata.
* **Bug Fixes**
  * Improved fail-safe handling for unclassified tools and tightened concurrency rules for interactive and workspace-modifying actions.
* **Tests**
  * Added extensive automated coverage for capability normalization/validation, resource-key derivation, catalog completeness, and concurrent lookups.


Vasanthdev2004 (commented):
Safe to merge as-is — this PR enables nothing (no concurrency wired, ResourceKeys are pure metadata, ValidateBuiltinCatalog validates the catalog against itself, not the live registry), so every issue below is latent with zero runtime effect today. I'm flagging two that **must be fixed before PR6 consumes these keys**, because the whole point of shipping the metadata first is to have correct conflict keys when batching lands — and for the process/session tools those keys are silently absent. The rest are nits/follow-ups.

**Decision up front:** merge this; fix items 1 and 2 before the concurrency follow-up reads the keys; items 3–6 whenever.

**1. (fix before PR6) process/session ResourceKeys never emit a key for the tools they target.** `processResourceKeys` is wired onto `write_stdin` (exec_command.go:723) and `terminal_session` (local_terminal.go:49), but both paths return empty for any real call:
- `write_stdin`'s `session_id` is `Type: "integer"` (exec_command.go:708). JSON unmarshals that as `float64`, and `firstStringArg` only accepts `value.(string)` — so the lookup silently misses and returns nil. The arg *name* matches `processResourceKeys`'s list; the blocker is purely the type assertion.
- `terminal_session`'s required arg is named `"session"` (local_terminal.go:31), but `processResourceKeys` looks for `session_id`/`sessionId`/`process_id`/`pid`/`id` — it never asks for `session`. The function that *does* know `session` (`sessionResourceKeys`, resource_keys.go:97) isn't wired here.

Net: the conflict metadata this PR advertises for the interactive/session tools is inert for exactly the tools it's meant to describe. Today that's harmless (all three are `EffectInteractive`/serialized). The moment PR6 starts grouping by resource key, two concurrent `write_stdin` calls on the same session won't collide-because they both hash to nothing — they'll look independent. Fix is small: either widen `firstStringArg` to coerce numeric ids, or wire `terminal_session` to `sessionResourceKeys` (and add `"session"` to `processResourceKeys`'s key list, or pick one function for session-keyed tools). Worth a test that a real `write_stdin`/`terminal_session` arg vector actually produces a non-empty key.

**2. (fix before PR6) `resourceHost` mishandles `@` in the path/query.** resource_keys.go:181 does `strings.LastIndex(raw, "@")` *before* stripping the path, so `https://api.example.com/path?x=@bad` collapses to `endpoint:bad` instead of `endpoint:api.example.com`, and two distinct hosts can collide on a garbage suffix. No secret leak (userinfo is stripped afterward), but the endpoint key is wrong — and wrong keys are the failure mode this contract exists to prevent. Either strip path/query/fragment first, or split on `://` and take the authority. Same fix-before-it's-read rationale as item 1; latent today only because `web_fetch`/`web_search` aren't batched.

**3. (nit, semantic) `bash` is `EffectInteractive` but the comment describes a mutator.** bash.go:56 says "Shell can mutate workspace and external state; never concurrent" yet classifies as `EffectInteractive`. Per the PR's own definitions, Interactive is browser/PTY/session-bound interaction — a one-shot `exec.CommandContext` shell matches none of that and fits `WorkspaceWrite`/`ExternalWrite`. The safety property holds (Interactive is never concurrent, so bash stays serialized — correct), but the label is imprecise: anyone auditing external mutators via `EffectExternalWrite` will miss `bash`, the canonical one. Either reclassify or tighten the `EffectInteractive` doc to explicitly cover "one-shot shell that must stay serialized." Not blocking — the *behavior* is right, only the *label* is debatable.

**4. (follow-up, tests) the gate isn't proven to gate.** `TestEveryBuiltinHasExplicitMetadata` only asserts `len(problems) == 0` — no test injects a deliberately misclassified/missing tool to assert `ValidateBuiltinCatalog` actually returns non-empty. A negative test would prove the validation is a real gate and not a rubber stamp. Same gap on `TestReadOnlyClassifications`: it asserts `Effect == ReadOnly` but never inspects `ThreadSafe`, and `ValidateCapabilities` *allows* `ReadOnly + ThreadSafe:true`, so a read-only tool silently marked thread-safe would pass unchallenged. If read-only tools are meant to be non-thread-safe unless explicitly audited, pin that; if some are legitimately thread-safe, assert the intended per-tool value.

**5. (follow-up, test) `TestCapabilitiesLookupRace` doesn't race.** It spawns 32 goroutines that all *read* — no concurrent registration during lookup, so the race detector has no write/read or write/write to flag. It passes trivially rather than testing the property its name claims. A real version would register concurrently with the lookups (or document that registration is construction-time-only and the test is a read-throughput smoke check).

**6. (nit) dead exports + comment drift.** `ResourceKeyPTY`/`ResourceKeyBrowser`/`ResourceKeyRepository` have zero references — drop or wire. And the `resourceHost` comment says "Strip port for stable host keys" while the code *keeps* `host:port`; pick one. The catalog also hand-adds `NewWebFetchTool`/`NewWebSearchTool` directly rather than going through a `CoreNetworkTools` equivalent, so the network tools are the one place the catalog can drift from the live registration path — a test asserting catalog == registered tools would close it. (The rest of the catalog is fine: it derives read/write/shell via the same `Core*` constructors the registry uses, so it's not a parallel hand-maintained list.)

**What's good:** the fail-closed default (`EffectUnknown` → never concurrency-eligible, `CapabilitiesOf` forces `ThreadSafe=false`) is the right contract and correctly rejects `ThreadSafe` for `WorkspaceWrite`/`ExternalWrite`/`Interactive`/`ProcessRead` while allowing `ReadOnly+ThreadSafe`. The capability-provider interface, `baseTool` plumbing, and `Registry.Capabilities` read-through are clean. CI green across all 9 checks. Good split — landing the metadata contract before enabling concurrency is the correct order, and these are exactly the kind of gaps to flush out now rather than when PR6 is reading the keys.

Vasanthdev2004:
Following up on my earlier comment — cross-checking against the other review on this PR surfaced three more items in the same latent class, all verified against `a3a19b26`. None change the verdict (still mergeable — this PR enables nothing), but two of them grow the **fix-before-PR6** list, so consolidating here rather than scattering.

**Updated decision:** merge this; fix items 1, 2, **3, 4** before the concurrency follow-up reads the keys; 5–7 whenever.

**3. (fix before PR6) `apply_patch` keys are inert — the worst instance.** apply_patch.go:38 is `EffectWorkspaceWrite` wired to `multiFileResourceKeys`, but `multiFileResourceKeys` reads `fileResourceKeys` (`path`/`file`/`file_path`/…) plus `args["paths"]` — apply_patch's args are `patch` + `cwd`, neither of which matches, so it returns nil for every real call. This is the same inert-keys pattern as item 1, and it's the sharpest one: apply_patch is the file-writing tool, so two concurrent edits to the *same file* would both hash to nothing and look independent to any future conflict check. Fix is the same shape — a patch-aware extractor that parses the affected file headers from `patch` and normalizes against `cwd`, or conservatively `workspaceResourceKeys` until that exists. Latent today (not batched); must be correct before PR6.

**4. (fix before PR6) the "ThreadSafe requires ReadOnly" rule isn't enforced at the lookup path.** `CapabilitiesOf` (capabilities.go:82) and `baseTool.Capabilities` (capabilities.go:96) each only clear `ThreadSafe` for the `EffectUnknown` case. The stronger invariant — a writer (`WorkspaceWrite`/`ExternalWrite`/`Interactive`/`ProcessRead`) must never be `ThreadSafe` — lives *only* in `ValidateCapabilities`, which runs at catalog-test time. The runtime lookup path the scheduler will use is `Registry.Capabilities` → `CapabilitiesOf`, and that returns the stored field as-is for any non-Unknown effect. Built-ins are fine because they pass `ValidateBuiltinCatalog` in CI. But plugin/MCP/specialist tools bypass that catalog — a plugin constructed `EffectWorkspaceWrite, ThreadSafe:true` flows through `CapabilitiesOf` unchecked and the scheduler would trust it, enabling concurrent writes to a writer. The fail-closed contract is currently enforced at validate-time for built-ins only, not at use-time for all tools. Normalizing in `CapabilitiesOf` (clear `ThreadSafe` unless `Effect == EffectReadOnly`) closes the seam. This is the one I'd most want nailed down before concurrency lands.

**5. (nit) `ValidateBuiltinCatalog` accepts a true duplicate registration.** builtin_catalog.go:59 only reports a problem when a repeated name has *different* Effect/ThreadSafe; if the metadata matches, a genuine double-registration is silently swallowed (and `ResourceKeys` is a func, so it can't be compared anyway — fair). A duplicate name is a catalog bug regardless of metadata; reject every repeat immediately.

**On my earlier two:** the terminal-session `session`-arg miss (item 1) and the `resourceHost` @-in-path bug (item 2) both came up independently in the other review with the same fixes I suggested (add `session` to the key list + a regression test; strip path/query/fragment before the userinfo split) — so those two are settled as real, not just my read. The full inert-keys set is now apply_patch, write_stdin, terminal_session — i.e., every tool that claims per-resource conflict metadata either has a type/name mismatch or reads the wrong arg. Worth treating as one sweep ("make every `ResourceKeys` func actually fire for its tool's real arg vector, with a test per tool") rather than three separate fixes.

Still a good PR and the right order to land it in — flush the metadata gaps now, while nothing reads them, rather than after PR6 is depending on them.

Vasanthdev2004 (approved):
Approving — all six items from the first pass are fixed correctly, and the fixes hold up under a second adversarial pass. Local `go build`, `go vet`, `gofmt -l`, and `go test -race ./internal/tools/...` are all clean on `fee94fc949`, CI is green, and the other review on this PR flipped to approve too. What's left below is minor polish, not blockers.

**Decision up front:** merge. Items 1–6 from round one are done; the five things below are non-blocking — #1 is a comment/regression worth a one-line fix, #2–3 are worth a small hardening before PR6 reads these keys, #4–5 are nits.

**What got fixed (verified against `fee94fc949`):**
- apply_patch now emits real keys via `applyPatchResourceKeys` (parses `+++ `/`--- ` headers + `cwd`, falls back to `workspace:`). Tests cover the diff+cwd case and the fallback.
- `normalizeCapabilities` is now the single read path — `CapabilitiesOf` and `baseTool.Capabilities` both route through it and force-clear `ThreadSafe` for every non-`ReadOnly` effect. This is the right move: the "ThreadSafe requires ReadOnly" invariant is now enforced at the read path the future scheduler will use, not just at catalog-test time. `TestNormalizeClearsThreadSafeOnMutators` exercises both the `baseTool` and `CapabilitiesOf` provider paths.
- `resourceHost` reordered — path/query/fragment stripped before userinfo, so `?token=@secret` can no longer steal the host. Test added.
- `write_stdin` (integer `session_id`) and `terminal_session` (`session` arg) now emit keys — `firstStringArg` coerces numeric JSON values and `processResourceKeys` accepts `session`. Both covered by tests using the tools' real arg shapes.
- `bash` reclassified `EffectWorkspaceWrite` so external-mutator audits include it; the comment now explains why it's not session-bound Interactive. `TestWorkspaceWriteClassifications` includes bash, `TestInteractiveClassifications` drops it.
- `ValidateBuiltinCatalog` rejects every duplicate name now, and `TestReadOnlyClassifications` asserts read-only tools are non-thread-safe pending per-tool audit (the ThreadSafe gap from round one, closed).

**1. (fix the comment, minor) the catalog gate no longer catches a mis-set `ThreadSafe` on a mutator — and the new comment says it does.** Moving enforcement into `normalizeCapabilities` was correct, but it has a side effect: `ValidateBuiltinCatalog` calls `CapabilitiesOf(tool)`, which normalizes `ThreadSafe` away *before* `ValidateCapabilities` sees it. So a constructor declared `EffectWorkspaceWrite, ThreadSafe:true` surfaces as `ThreadSafe:false` and the validator reports nothing. At the base commit this *was* caught (the old read path only cleared `ThreadSafe` for `EffectUnknown`), so this is a real — if harmless — regression of the CI guard-rail. Runtime is safe: the normalizer is the actual enforcement and it runs on every read path, so a future planner checking `ThreadSafe && EffectReadOnly` can never see a mutator as thread-safe. The defect is only that the new comments overclaim it: builtin_catalog.go:69 says "so a mis-set ThreadSafe on a mutator is still reported," and the `ValidateCapabilities` note says "catalog tests catch mis-declared constructors" — both false as written. Either soften the comments to match (the normalizer enforces at read time; the validator is a direct-call sanity check, not the catalog gate) or validate the raw declaration too. `TestValidateBuiltinCatalogRejectsUnknownRaw` also tests the wrong function — it calls `ValidateCapabilities` directly with a hand-built record, not `ValidateBuiltinCatalog`, so it doesn't actually prove the catalog gates anything. Not blocking; it's a comment + a test that names one thing and checks another.

**2. (harden before PR6, minor) `pathsFromUnifiedDiff` is prefix-only, so it can misread two cases.** It matches `+++ `/`--- ` anywhere with no `@@` hunk-state tracking, so a diff *content* line that happens to start with `++ ` (an added line whose text is `++ helper`) collides with the header prefix and yields a spurious `file:` key — safe direction (over-serialization, not a missed conflict), but noise. And it doesn't unquote git-quoted headers: `+++ "my file.txt"\t2023` emits `file:"my file.txt"` with the literal quotes intact. Both only affect the conflict-key metadata, and apply_patch always falls back to `workspace:` keys, so no runtime effect today; worth tightening before PR6 actually groups by these keys. Tracking `@@` hunks and stripping C-style quotes (or just taking the `b/` side after the `+++`) closes both.

**3. (theoretical, minor) `firstStringArg`'s numeric coercion can round a very large integer id.** `typed == float64(int64(typed))` loses precision past 2^53 and the `int64(typed)` cast can overflow; production always decodes numbers as `float64` (so the precision-preserving `jsonNumber` branch is effectively dead). Session ids are small in practice so this is theoretical, but if a non-session numeric arg ever flows through, a rounded key is a silent mismatch. Fine to leave with a comment noting the bound, or route through `json.Number` when available.

**4. (nit) reserved constants.** `ResourceKeyPTY`/`Browser`/`Repository` are now documented as reserved for follow-up classifiers — fine as vocabulary stability, just confirming they're intentionally unused rather than forgotten.

**5. (nit)** `exec_command` (the one-shot shell, exec_command.go:469) stays keyless — it has no session arg, so `processResourceKeys` returns nil for it. That's correct (it's `EffectInteractive`/never batched, and one-shot with no persistent resource to conflict on), just noting it's by design, not an oversight.

Good cleanup. The shift from "validate at test time" to "enforce at the read path" is the right architecture for this contract — the remaining work is comment accuracy and `pathsFromUnifiedDiff` hardening, both small.

anandh8x (approved):
## Re-review at 0e1b842

Code-side review of the latest head on `perf/tool-effect-metadata`. This is the Stream D / PR5 contract that gates PR6, so the bar is "the contract is correct, the coverage is 100%, and a future concurrent-batches planner can rely on it without re-checking." Here's my read.

### What's good

- **The `EffectClass` enum is the right shape.** Six values, ordered from fail-closed (`EffectUnknown`) to clearly distinct (ReadOnly / ProcessRead / WorkspaceWrite / ExternalWrite / Interactive). The `String()` and `Valid()` methods are present and correctly bounded.
- **The `normalizeCapabilities` invariant is the load-bearing line.** `ThreadSafe` is force-cleared unless `Effect == EffectReadOnly`. Without this, a misconfigured tool could leak `ThreadSafe=true` on a mutator. Tested by `TestNormalizeClearsThreadSafeOnMutators`.
- **The `declaredCapabilities` / `CapabilitiesOf` split is correct.** Catalog validation uses the raw declaration (so a misconfigured `ThreadSafe=true` on a mutator is still flagged), runtime uses the normalized form (so a caller never sees an inconsistent record). The split is the only way both invariants hold simultaneously. Tested by `TestDeclaredMutatorThreadSafeIsReportedByCatalogValidation`.
- **Resource-key extractors are conservative.** `fileResourceKeys` returns `nil` on missing/empty path, never panics. `firstStringArg` coerces `float64` (JSON numbers) to int-shaped strings when exact, so `write_stdin(integer session_id)` works. Tested by `TestProcessResourceKeysNumericSessionID`.
- **The `@`-in-query endpoint attack is defended.** `resourceHost` strips path/query/fragment *before* userinfo, with a comment that names the threat. Tested by `TestEndpointResourceKeysAtInQueryDoesNotStealHost`. This is the right order; reversing it would leak the query string into the key.
- **`applyPatchResourceKeys` joins diff paths under `cwd`.** The "same file under different cwds does not collide" property is what the resource keys exist to provide, and `TestApplyPatchResourceKeysFromDiffAndCwd` is the regression catch.
- **`processResourceKeys` already accepts `session` as the first key.** It checks `session`, `session_id`, `sessionId`, `process_id`, `pid`, `id` (in that order). `TestProcessResourceKeysSessionArg` exercises the `session` arg specifically. The PR-body claim "terminal_session requires `session`" is consistent with the implementation, and the resource-key contract holds.
- **`BuiltinCatalog` rejects duplicate names outright.** The `seen` map short-circuits with a diagnostic. Function values cannot be compared reliably, so partial metadata comparison would have been a foot-gun. Right call.
- **Coverage is real.** The PR body says 10 ReadOnly / 4 WorkspaceWrite / 20 Interactive, and the `TestEveryBuiltinHasExplicitMetadata` + `TestBuiltinCatalogNoUnknown` tests enforce it. MCP / plugin / specialist tools are explicitly out of scope and remain `EffectUnknown` by design.
- **Test surface is comprehensive.** 21 tests, 449 LOC. The right tests are present (defaults, nil, normalization, validation, every-builtin, registry lookup, file/endpoint/process/apply_patch keys, platform normalization, secret-stripping, the `@`-in-query attack, workspace write, interactive, declared-vs-runtime).

### CodeRabbit inline threads — most are stale

The 6 CodeRabbit inline comments (review id 4715... on the most recent pass) are largely stale relative to the current head. The two I think are still worth a follow-up are:

- **`internal/tools/builtin_catalog.go:62` — reject duplicate builtin names.** Addressed in the current head (`seen[name]` short-circuits at line 60-64). The proposed fix the bot suggested is what the code already does. Mark resolved.
- **`internal/tools/capabilities.go:116` — normalize non-read-only declarations.** Addressed (`normalizeCapabilities` at the single read path `CapabilitiesOf` and `baseTool.Capabilities`). The proposed fix is what the code already does. Mark resolved.
- **`internal/tools/local_terminal.go:49` — `session` arg not in extractor.** The bot's claim is stale: `processResourceKeys` already checks `"session"` as the first key (line 105), and `TestProcessResourceKeysSessionArg` exercises it. Mark resolved.
- **`internal/tools/resource_keys.go:299` — `@`-in-query attack.** Addressed (`resourceHost` strips path/query/fragment first, then userinfo; comment names the threat; test exists). Mark resolved.
- **`internal/tools/builtin_catalog.go:71` — validate declared before normalize.** Addressed (`declaredCapabilitiesOf` is the catalog's source; `CapabilitiesOf` is the runtime's source; the split is tested). Mark resolved.
- **`internal/tools/resource_keys.go:171` — cwd-relative diff paths.** Addressed (`applyPatchResourceKeys` calls `joinUnderResourceCwd`; `TestApplyPatchResourceKeysFromDiffAndCwd` asserts the `pkg/old.go` / `pkg/new.go` shape). Mark resolved.

The bot is reading a stale snapshot. If you want to be defensive, dismiss those threads as resolved rather than letting them sit on the PR. None of the six are blocking.

### Substantive follow-ups (not blockers, worth tracking)

1. **`processResourceKeys` only keys on `session`/`pid` etc., not on `cwd`.** If two parallel `write_stdin` calls targeted the same `session` from different working directories, they would not collide. This is the right answer (cwd does not scope a process), but the doc comment on `processResourceKeys` could be explicit about it. One-line change.
2. **`resourceHost` is permissive about ports.** `host:port` is preserved, which is correct. A follow-up could add a small test asserting `endpoint:api.example.com:443` and `endpoint:api.example.com:80` produce different keys (the test as written exercises different hosts, not different ports on the same host).
3. **The `ResourceKey*` constants include reserved prefixes** (`repository:`, `pty:`, `browser:`) that have no extractors yet. The comment says "reserved for follow-up classifiers so PR6 conflict keys stay vocabulary-stable" — which is the right design (reserve the namespace, fill in later), but a future PR that wants to add e.g. `repository:` keys needs to know which prefixes are reserved vs. live. A docstring on the `ResourceKey*` const block listing "live / reserved" would help.
4. **Plugin and MCP tools stay fail-closed (`EffectUnknown`).** Correct, but the existing `Registry.Capabilities` lookup at line 119 of `capabilities.go` returns `UnknownCapabilities()` for unknown *names* — same as for tools that implement `CapabilityProvider` with `EffectUnknown`. A consumer cannot distinguish "this tool exists but is unclassified" from "this tool doesn't exist." Not a blocker, but the doc on `Registry.Capabilities` should call out that the result is intentionally lossy on purpose.
5. **`sessionResourceKeys` produces a `session:` key for browser and terminal tools.** The session namespace is shared between processes, PTY, and browser — a session id from one category could collide with a session id from another if a tool ever crosses categories. This is theoretical (a browser and a terminal won't share session ids in practice) but the doc could call out the per-category assumption.

### Verdict

**Approve.** The contract is correct, the coverage is real, the tests cover the right edge cases, the security path is well-defended, and the inline CodeRabbit threads are stale. The five follow-ups are doc-quality improvements and small test additions, not blockers for this PR.

Ready for kevin to merge. The M1-baseline + PR6 chain is now unblocked — gnanam's PR5 closes the safety contract that PR6's planner reads.

kevincodex1 (approved):
LGTM

PR #706, opened 2026-07-16
chore: remove code unreachable on every supported platform

euxaristia:
## Summary

The Security and code health job's advisory deadcode step reports 152 unreachable functions on main. This PR removes everything that is genuinely dead on every supported platform and documents the handful of entries that must stay.

The raw report overcounts. deadcode analyzes one GOOS at a time, and CI runs it on ubuntu, so functions whose only callers sit in windows build-tagged files look unreachable. I ran the same check under GOOS=linux, GOOS=windows, and GOOS=darwin and only touched functions unreachable on all three. That leaves 139 candidates, of which 136 are removed here.

## Changes

- Deletes 44 functions with no references anywhere, including the internal/reasoning package, which nothing imports.
- Moves 92 functions that only tests reference into per-package export_test.go files. Tests keep their seams and observation helpers, and the production binary sheds the code.
- Ports test call sites of deleted convenience wrappers to the live APIs: New*Tool to NewScoped*Tool, CoreTools to CoreToolsScoped, tools.DeferredLine to tools.DeferredSource, and Truncated() assertions to FinishReason checks.

## Kept, with reasons

- The 13 sandbox windows_*.go functions in the report are false positives on ubuntu. Their callers live in windows build-tagged files, verified by running deadcode with GOOS=windows.
- cron.Store.Runs is the only reader of runs.jsonl, which production writes via AppendRun. Tests in cron and cli read runs back through it.
- agenteval.TraceEventKeys and agenteval.MissingTraceEvents are used across packages by the trace contract test.

After this PR the ubuntu deadcode step reports 16 findings: the 13 windows false positives plus the 3 keeps above.

## Verification

- go build ./..., go vet ./..., and gofmt are clean.
- go test ./... passes.
- deadcode -test=false ./... rerun under all three GOOS values; the remaining linux findings are exactly the 16 documented above.

## Summary by CodeRabbit

* **New Features**
  * Added clearer truncation/early-stop notices when responses end due to token limits, content filtering, or other finish reasons.
  * Improved attachment chips by displaying image and document chips together, with a warning when the selected model can’t process images.
  * Enhanced terminal transcript scrolling and overlay rendering.

* **Bug Fixes**
  * Strengthened workspace path scoping and safety checks across built-in tools.
  * Refined tool output budgeting/truncation metadata and improved skill discovery across multiple locations.


Vasanthdev2004 (dismissed):
Approve. Cross-built `go build ./...` + `go vet ./...` (which compiles the tests too) under GOOS=windows/linux/darwin  all clean  and `go test ./...` on Windows passes apart from the two usual pre-existing env failures (the symlink-privilege one and the alt-screen TUI render, both fail identically on main). The 3-GOOS union is actually the full platform surface here since the repo has no GOARCH build tags, and the four renames are genuine behavior-preserving inlines (`CoreReadOnlyTools` was literally `return CoreReadOnlyToolsScoped(root, nil)` on main, same for the other three), and `internal/reasoning` is gone with no leftover importers.

One optional, non-blocking nit: the `ValidateFile` doc comment in `internal/config/export_test.go` claims callers can `errors.As` the JSON parse errors out of the returned `Issue`, but `Issue.Message` is a plain string and the parse branch flattens the wrapped error via `.Error()`, so there's no chain to unwrap. That's pre-existing  the PR moved it verbatim from `validate.go`  so a separate one-liner is fine, or fold in CodeRabbit's `fmt.Sprintf("%v", err)` if you'd rather handle it here.

jatmn (changes requested):

## Findings

- [P2] Rebase this head onto current `main` before resolving the code review
  The PR head is based on `9043bae`, while current `main` is `18cce35`; GitHub reports it as `CONFLICTING`, and a merge-tree check conflicts in the core tool/output paths (`bash`, `exec_command`, file tools, and trace emission). The unresolved merge would either prevent merging or require a manual resolution that determines whether recently merged behavior is retained. Please rebase, keep the current-main behavior, and request a review of the resulting diff.

- [P1] Do not roll the published release metadata back to 0.3.0
  `.release-please-manifest.json:1`, `package.json:3`, and `package-lock.json:3`
  The base is the already-released `v0.4.0`, but this cleanup resets all package/release state to `0.3.0` and deletes the 0.4.0 changelog entry. Release Please and the npm packaging workflow consume these values, so the next release would be calculated from stale history and can try to publish an already-existing version. Keep the 0.4.0 release artifacts when rebasing.

- [P1] Restore the capability gate before batching read tools
  `internal/agent/parallel_tools.go:35`
  `parallelSafeToolCall` now admits every `SideEffectRead` tool, but the removed gate required `ThreadSafe` capability and rejected conflicting resource keys. This is not just conservative metadata: `read_file` explicitly declares `ThreadSafe: false` because it updates the file tracker, and plugin/read tools can be stateful too. Two consecutive read calls are therefore launched concurrently despite declaring that they are not safe to do so, which can race shared state or break same-resource ordering. Reinstate the thread-safety and resource-conflict checks (or leave these calls serialized).

- [P1] Reapply the output budget after after-tool hooks append feedback
  `internal/agent/loop.go:1293`
  The registry has already applied its redaction/output ceiling when `dispatchAfterTool` returns, but this path appends hook feedback and no longer calls `RebudgetAfterHook`. A verbose formatter, linter, or test hook can consequently append an unbounded multi-megabyte result directly to the model transcript, bypassing both the normal ceiling and the tool-specific `max_output_tokens` budget. Cap/redact the combined result again after appending hook output.

- [P1] Preserve the original tool-truncation bit in machine output
  `internal/cli/exec_writer.go:169`
  The PR removes `agent.ToolResult.Truncated`, so stream JSON now sets `truncated` solely from its own 1 MiB serialization cutoff. A `read_file`/search/command result that was already truncated by the tool but is smaller than that cutoff is emitted as `truncated: false` (and JSON mode omits the field), making automation treat partial output as complete. Carry the tool-level flag through the agent result and OR it with the stream-layer truncation.

- [P1] Keep the strengthened benchmark correctness oracles
  `internal/perfbench/turn_bench.go:580`
  This removes the v3 oracle stamping and captured-answer verification, downgrades navigation tasks to latency-only, and changes refactors back to a non-positive `go build` check. In particular, an agent can leave a refactor untouched and still satisfy the build command, so the published benchmark no longer measures the requested transformation or preserves comparable correctness rates. Retain the current-main v3 fixture/oracle setup rather than deleting it in this cleanup.

- [P3] Correct the moved `ValidateFile` documentation
  `internal/config/export_test.go:21`
  The unresolved CodeRabbit comment is valid: `ValidateFile` returns `[]Issue` with a string `Message`, after calling `.Error()` on the wrapped JSON error, so callers cannot use `errors.As` to recover `*json.SyntaxError` or `*json.UnmarshalTypeError`. Remove that claim (and use `fmt.Sprintf(...%v...)` instead of a discarded `%w` wrapper) so the exported test seam does not promise an impossible API.

- [P3] Remove the extra trailing blank line
  `internal/trace/trace.go:271`
  This adds a blank line at end of file (`git diff --check` reports it), which fails the repository's formatting gate. Remove the extra line so the required smoke/format check can pass.

Vasanthdev2004 (changes requested):
Really close now. The rebase onto current main resolved everything heavy from the last round — the release metadata rollback, the parallel-tools gate, the hook rebudgeting path, and the bench oracles are all out of the diff, and the ValidateFile doc fix landed and reads right. One thing left: the ubuntu smoke job is red because gofmt flags internal/trace/trace.go — the PR still adds a stray blank line at the end of that file (jatmn caught this last round too). Delete that one line and push, and this gets my approve.

jatmn (changes requested):

## Findings

- [P1] Restore Go formatting before merging
  `internal/trace/trace.go:287`
  The extra blank line at EOF makes `gofmt -l .` report this file, which is why the required Ubuntu Smoke job fails at “Check formatting” and skips its Linux vet, test, build, and smoke stages. Remove the trailing blank line (or run `gofmt`) so the required gate can run.

- [P2] Keep the exported trace adapters implementing their documented contract
  `internal/trace/emit.go:10-16,141-145`
  This leaves `Sink`, `NDJSONSink`, and `TextSink` exported and documented as writer-backed sinks, but removes both `Emit` methods; `TextSink` also loses its only production emitter when `WriteText` moves to `_test.go`. Consequently, an in-module consumer such as `var s trace.Sink = trace.NDJSONSink{W: w}` no longer compiles, and `TextSink` cannot emit a trace at all. Retain the methods (and `WriteText` for the text adapter), or remove/update these obsolete exported types and their contract together.

- [P2] Update skills documentation that references removed APIs
  `internal/skills/skills.go:42,138,142,243`
  `DiscoveryRoots` and `Duplicates` now exist only in `export_test.go`, yet the retained package documentation and `docs/HOW_ZERO_WORKS.md` direct discovery callers to those production helpers. A contributor following that documented API gets an undefined-symbol compile error. Replace those references with the retained production entry points (or keep compatible wrappers) as part of this removal.

euxaristia:
Pushed f616b82 addressing the remaining findings: the stray blank line at EOF of internal/trace/trace.go is gone (gofmt clean), and the orphaned exported adapters are resolved by removal — Sink, NDJSONSink, and TextSink had lost their Emit methods and had no implementers or callers left, so the types go with their contract (consistent with this PR's purpose). Skills package docs and HOW_ZERO_WORKS.md no longer direct callers to DiscoveryRoots/Duplicates, which are now test seams; they point at LoadFromRoots instead.

jatmn (dismissed):
@Vasanthdev2004 LGTM

euxaristia:
Pushed 8e110f5.

Fixed:
- export_test.go still had fmt.Errorf(...%w...).Error(), which discards the wrapped error immediately. Switched to fmt.Sprintf with %v.
- loop.go had a 4-line comment block duplicated verbatim above partitionToolsCached. Removed the duplicate.

Not fixing, with reason:
- The suggestion to split file_tools_test.go (~800 lines) into per-source test files: this structure predates the PR, is unrelated to its actual purpose of removing platform-dead code, and was itself scored lowest severity. Leaving it for a separate refactor if we want it.

Vasanthdev2004 (dismissed):
Re-verified on the current head. My one blocker is gone: the stray trailing blank line in internal/trace/trace.go that was failing gofmt on the ubuntu smoke job. trace.go isn't touched by the PR anymore, it ends cleanly, and CI is green across every job including Security & code health and all three Smoke platforms. The optional ValidateFile doc nit from the earlier round got folded in too. Clean dead-code removal. Approving.

euxaristia:
@jatmn Ready for re-review.

Your last note was LGTM; that review was dismissed by a later push. Current head is green on Smoke (all three OS), Security, and Zero Review, with @Vasanthdev2004 approval. Please take another look when you can.

kevincodex1:
hello @euxaristia please rebase to main and fix conflicts

euxaristia:
Rebased onto main to resolve a conflict in internal/trace/emit.go.

Since this branch diverged, main added a task_state summary line to WriteText. This PR moves WriteText out of production code into internal/trace/export_test.go (it's only referenced by tests). I ported the new task_state block into that moved copy so the feature isn't lost.

Verified after rebase: go build ./..., go vet ./..., and go test ./... all pass.

euxaristia:
Addressed the latest CHANGES_REQUESTED findings and rebased onto current main.

Fixed:
- [P1] gofmt on `internal/trace/trace.go`: trailing blank line at EOF already removed; `gofmt -l` is clean on the head.
- [P2] trace adapters: `Sink` / `NDJSONSink` / `TextSink` were removed together with their contract (no production callers); `WriteNDJSON` remains the production emitter, and `WriteText` lives in `export_test.go` as a test seam. Matches the "remove/update obsolete exported types together" option.
- [P2] skills docs: package docs and `docs/HOW_ZERO_WORKS.md` already point at `LoadFromRoots` / `Load` instead of `DiscoveryRoots` / `Duplicates`. Follow-up on this push: Load and MergedSkills comments no longer name test-only `List`/`Get`; they reference `Load` / `ListFromRoots`.

Also rebased onto `gitlawb/main` (was 14 commits behind).

Verified: `gofmt -l .` clean; `go test ./internal/trace/... ./internal/skills/... ./internal/plugins/... -count=1`; `go vet` on the same packages.

jatmn (changes requested):

## Findings

- [P3] Remove the reference to the deleted truncation helper
  `internal/zeroruntime/helpers.go:16-19`
  `CollectedStream.Truncated()` is deleted in this change, but the `FinishReason` field comment still tells callers that `Truncated` reports whether it is set. That documents a nonexistent API and sends future callers/tests toward a compile failure. Update the comment to say callers should inspect `FinishReason` directly.

- [P3] Correct the deferred-MCP helper documentation after moving `DeferredLine` to test-only code
  `internal/tools/deferred.go:40-45`
  The production comment says `mcpServerNamed` is consumed by `DeferredLine`, but `DeferredLine` now exists only in `export_test.go`; production discovery uses `DeferredSource`. This leaves the retained production source documenting an unavailable helper. Name `DeferredSource` here, and mark the copied `DeferredLine` comment as a test seam rather than claiming the agent loop uses it.

jatmn (changes requested):
wrong findings.

jatmn (changes requested):

## Findings

- [P1] Restore a compilable `internal/tools` test package
  `internal/tools/export_test.go:133-141`
  The new test seam tries to add a method to `execSessionManager`, which is now an alias for `execution.ProcessManager`; Go forbids methods on that non-local type, and the referenced `mu` and `sessions` fields no longer exist. The following helper also calls the removed `sandboxDenialKind`. As a result, `go vet ./...` and `go test ./...` fail in the Ubuntu, macOS, and Windows smoke jobs. Remove or rewrite these stale helpers against the current process-manager/sandbox APIs so the required checks can run.

- [P3] Update production comments that now direct callers to test-only APIs
  `internal/zeroruntime/helpers.go:16-19`, `internal/tools/deferred.go:40-45`, `internal/config/validate.go:15-20`, `internal/providers/providerio/providerio.go:198-200,364-367`
  This move leaves normal-source comments naming helpers that now exist only in `export_test.go`: `CollectedStream.Truncated`, `DeferredLine`, `ValidateFile`, and `ScanSSEData`. This is documentation drift rather than a runtime regression, but it points future contributors at APIs unavailable in normal builds. Point the comments at the retained production APIs (`FinishReason`, `DeferredSource`, `ValidateBytes`, and the production SSE scanner) or describe the behavior without naming removed APIs.

euxaristia:
Rebased onto current upstream main to resolve the conflict from the sandbox unify-command-execution refactor (96859c9).

Conflicts were in internal/sandbox/runner.go, internal/tools/exec_command.go, and internal/tools/sandbox_denial.go. In each case the dead code this PR removes had already been superseded by that refactor (moved into export_test.go as a test seam, or replaced outright by the new execution package), so the resolution keeps main's new code and layers this PR's deletions on top of it. Also dropped two stale test seams in internal/tools/export_test.go that referenced symbols the refactor had already deleted.

Build is clean and tests match baseline behavior on an unmodified upstream/main checkout.

euxaristia:
Follow-up: the rebase push above landed a broken commit. The conflict resolution deleted internal/sandbox/runner.go functions that internal/tools/export_test.go still referenced (execSessionManager.len(), likelySandboxDenied, sandboxDenialKind), and I'd fixed that in my working tree and verified the build locally, but pushed before committing the fix, so the pushed branch didn't include it.

Pushed the missing commit (removing the stale export_test.go seams). Branch now builds and go vet passes cleanly on the currently pushed HEAD.

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):

kevincodex1:
please check bro @Vasanthdev2004

euxaristia:
Rebased onto current upstream main (was 6 commits behind). No conflicts; focused tests for tools/sandbox/skills and `go build ./cmd/zero` pass locally.

PR #711, opened 2026-07-17
docs(readme): note govulncheck/golangci-lint install to GOPATH/bin

euxaristia:
## Summary

The contributor setup in `README.md` tells readers to `go install` `golangci-lint` and `govulncheck`, but does not mention that the resulting binaries land in `$GOPATH/bin` (default `~/go/bin`). If that directory is not already on `PATH`, the tools cannot be run directly, which silently breaks the documented lint/security workflow. This adds a short note and an `export PATH` snippet so the tooling actually resolves.

## Changes

- `README.md` (contributor/lint-tooling section): document that installed binaries go to `$GOPATH/bin`, and add a copy-paste `export PATH="$PATH:$(go env GOPATH)/bin"` snippet.

## Test plan

- `make lint` and `govulncheck ./...` resolve after adding `$(go env GOPATH)/bin` to `PATH`.
- Visual check of the rendered README section.

## Summary by CodeRabbit

- **Documentation**
  - Updated installation instructions to clarify where Go-installed vulnerability scanning tools (e.g., `govulncheck`) are placed, including the default `~/go/bin` when `$GOBIN` isn’t set.
  - Added instructions to ensure the Go bin directory is included in your `PATH`, with a shell snippet to set it if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,11 +377,13 @@ go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2
 go install golang.org/x/vuln/cmd/govulncheck@v1.3.0
 ```
 
-The installed binaries land in `$GOPATH/bin` (default `~/go/bin`), which must be on
-your `PATH` to run them directly. If it isn't, add it:
+The installed binaries land in `$GOBIN` when it is set, otherwise in
+`$GOPATH/bin` (default `~/go/bin`). That directory must be on your `PATH` to
+run them directly. If it isn't, add it:
 
 ```bash
-export PATH="$PATH:$(go env GOPATH)/bin"
+gobin="$(go env GOBIN)"; [ -z "$gobin" ] && gobin="$(go env GOPATH)/bin"
+case ":$PATH:" in *":$gobin:"*) ;; *) export PATH="$PATH:$gobin" ;; esac
 ```
 
 ### Cross-Compile Examples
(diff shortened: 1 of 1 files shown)


Vasanthdev2004 (dismissed):
Docs-only change to README.md: notes that `go install` places binaries in $GOPATH/bin and shows how to add it to PATH. Build and vet are clean, and the added snippet (`export PATH="$PATH:$(go env GOPATH)/bin"`) is correct and portable. Tiny wording nit, not worth blocking on: the sentence singles out "to run govulncheck directly" when the same applies to golangci-lint installed just above — "to run them directly" would be slightly more accurate. Approving.

Vasanthdev2004 (dismissed):
Re-approving. This is the wording fix I asked for last round: it now says "run them directly" instead of singling out govulncheck, which reads right since golangci-lint installs to the same GOPATH/bin. Docs-only, nothing else to flag.

euxaristia:
Pushed 8671f7c: the install-path snippet now honors GOBIN with a $(go env GOPATH)/bin fallback, and the PATH update is guarded so rerunning it doesn't append duplicates (verified idempotent).

euxaristia:
Ready for review.

Docs-only PATH/GOBIN note. CI is green; CodeRabbit approved. No open review threads.

kevincodex1 (approved):
lgtm

PR #716, opened 2026-07-17
fix(providers): stop "provider not found" for env-derived profiles

### LINKED ISSUE #707: Provider deletion leaves stale OpenAI provider, causing "provider not found" and preventing Codex authentication
### Version / branch / commit

3.0

### OS and environment

Windows 11 Pro Powershell

### Steps to reproduce

# Provider deletion leaves stale OpenAI provider, causing "provider not found" and preventing Codex authentication

## Version / branch / commit

Latest Zero release (July 2026)

## OS and environment

- Windows 11
- PowerShell 7

## Steps to reproduce

1. Open Zero.
2. Add an OpenAI provider using an API key.
3. Open the **Providers** menu.
4. Press `d` to delete the OpenAI provider.
5. Zero displays:
   ```
   Delete failed: provider "openai" not found
   ```
6. The OpenAI provider still appears in the provider list and remains marked as active.
7. Select an OpenAI model (for example `gpt-4.1` or `gpt-5.6-sol`).
8. Start a new chat.

## Expected behavior

- Deleting the provider should either:
  - Successfully remove it, or
  - Display an error only if the provider truly does not exist.
- After authenticating with Codex/OpenAI, selecting an OpenAI model should work normally.
- The provider list and internal configuration should remain in sync.

## Actual behavior

- Zero reports:
  ```
  Delete failed: provider "openai" not found
  ```
- The OpenAI provider is still visible in the Providers list and marked as active.
- Selecting an OpenAI model results in:
  ```
  gpt-5.6-sol · openai · not saved (provider "openai" not found)
  ```
- The UI indicates that an OpenAI provider exists, while the backend behaves as if it does not.
- Because of this inconsistent state, it is impossible to remove the provider and re-authenticate using the Codex/OpenAI login flow.

Image

Image

PierrunoYT:
Fixes #707

## Problem
`RemoveProvider`/`SetActiveProvider`/`SetProviderModel` only ever look at rows persisted in `config.json`. But a provider can appear in the resolved provider list (TUI picker, `zero providers list`) purely because `Resolve()` synthesized it in-memory from an ambient env var (e.g. `OPENAI_API_KEY`) without ever writing it to disk.

That mismatch caused a confusing `provider "openai" not found` error when a user tried to:
- Delete the provider from the TUI Providers manager (press `d`)
- Select a model for it (`/model` picker) — showed `not saved (provider "openai" not found)`
- Run `zero providers use <name>` or `zero providers remove <name>` from the CLI

## Fix
Added `config.ProviderPersisted(path, name)` to distinguish "not on disk" from "doesn't exist anywhere," and used it at all four call sites so an env-derived provider gets an accurate, non-error message instead of a bogus not-found failure:
- `internal/tui/provider_manager.go` — provider manager delete
- `internal/tui/command_center.go` — `/model` picker persistence
- `internal/cli/provider_onboarding.go` — `zero providers use` / `zero providers remove`
- `internal/config/writer.go` — the new shared `ProviderPersisted` helper

Genuinely unknown/misspelled provider names still fall through to the original "not found" error.

## Testing
- `go build ./...`
- `go vet ./...`
- `go test ./internal/cli/... ./internal/config/... ./internal/tui/...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for environment-derived providers that aren’t saved in configuration, with clearer behavior for **use**, **remove**, and **rename** (including no-op messaging instead of incorrect saved-config operations).
  * Prevented model save and provider edit actions from proceeding for session-only providers without a saved profile.

* **Improvements**
  * Standardized `--json` output for env-derived **use/remove/rename**, including consistent `persisted` and `configPath` reporting.
  * Added CLI tests for malformed config handling and env-derived provider flows to prevent regressions.

### DIFF
--- a/internal/cli/provider_onboarding.go
+++ b/internal/cli/provider_onboarding.go
@@ -60,7 +60,11 @@ func runProvidersUse(args []string, stdout io.Writer, stderr io.Writer, deps app
 	// without ever writing a row to disk. Without this check, switching to
 	// that provider by name always fails with a confusing "not found" even
 	// though it is genuinely usable this session (issue #707).
-	if !config.ProviderPersisted(configPath, options.name) {
+	persisted, err := config.ProviderPersisted(configPath, options.name)
+	if err != nil {
+		return writeAppError(stderr, err.Error(), exitCrash)
+	}
+	if !persisted {
 		if exit, handled := reportUnpersistedProviderUse(stdout, stderr, deps, options); handled {
 			return exit
 		}
@@ -371,8 +375,12 @@ func runProvidersRemove(args []string, stdout io.Writer, stderr io.Writer, deps 
 	// without ever writing a row to disk. Without this check, deleting that
 	// provider by name always fails with a confusing "not found" even though
 	// it is genuinely visible/usable this session (issue #707).
-	if !config.ProviderPersisted(configPath, name) {
-		if exit, handled := reportUnpersistedProviderRemove(stdout, stderr, deps, name, options.json); handled {
+	persisted, err := config.ProviderPersisted(configPath, name)
+	if err != nil {
+		return writeAppError(stderr, err.Error(), exitCrash)
+	}
+	if !persisted {
+		if exit, handled := reportUnpersistedProviderRemove(stdout, stderr, deps, name, options.json, configPath); handled {
 			return exit
 		}
 	}
@@ -452,6 +460,16 @@ func runProvidersRename(args []string, stdout io.Writer, stderr io.Writer, deps 
 	if err != nil {
 		return writeAppError(stderr, err.Error(), exitCrash)
 	}
+	oldName := options.names[0]
+	persisted, err := config.ProviderPersisted(configPath, oldName)
+	if err != nil {
+		return writeAppError(stderr, err.Error(), exitCrash)
+	}
+	if !persisted {
+		if exit, handled := reportUnpersistedProviderRename(stdout, stderr, deps, oldName, options.json, configPath); handled {
+			return exit
+		}
+	}
 	cfg, err := config.RenameProvider(configPath, options.names[0], options.names[1])
 	if err != nil {
 		return writeAppError(stderr, err.Error(), exitCrash)
@@ -504,7 +522,7 @@ func reportUnpersistedProviderUse(stdout, stderr io.Writer, deps appDeps, option
 		return exitCode, false
 	}
 	message := fmt.Sprintf(
-		"Provider %q is not saved in config.json (likely set via an environment variable), so there is no saved profile to switch to.\nIt is already used automatically whenever its environment variable is set; unset it to stop Zero from detecting it automatically.",
+		"Provider %q is not saved in config.json (likely set via an environment variable), so there is no saved profile to switch to.\nIt is available whenever its environment variable is set, but is only active when selected (for example via ZERO_PROVIDER); unset its environment variable to stop Zero from detecting it automatically.",
 		options.name,
 	)
 	if options.json {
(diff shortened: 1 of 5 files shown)


Vasanthdev2004 (changes requested):
The core fix is right  `ProviderPersisted` cleanly separates "not on disk" from "doesn't exist anywhere," the four call sites you guarded are correct, the `/model` picker still applies the model in-session for env-derived providers, and the typo fall-through still hits the real "not found". But the class fix is incomplete  two more reachable name-keyed mutators have the identical env-derived gap, and the `providers remove --json` shape has a type inconsistency.

1. `zero providers rename` (provider_onboarding.go:455) calls `config.RenameProvider` with no `ProviderPersisted` guard, so `rename openai myopenai` on an env-derived openai still fails `provider "openai" not found`  the same bug you closed for `use`/`remove`. It needs the same persisted-check + explanatory message (or fall-through) as the siblings.

2. The TUI `/providers` edit-save path (`saveManagerEdit`, provider_manager.go:611) calls `config.EditProvider` with no guard. You guarded the sibling delete path (`deleteManagerSelection`) in this same file, but pressing `e` on an env-derived row, editing a field, and saving hits `provider "openai" not found` the same way. Env-derived rows do reach the editor — `savedProviders` is seeded from `usableSavedProviders(resolved.Providers)`, so the env-derived profile is in the manager list and editable. Needs the same `ProviderPersisted` branch as delete.

3. `providers remove --json` emits `removed` as a STRING on success (line 389, the name) but a BOOLEAN on the env-derived path (line 547, `false`)  same key, different JSON type, both exit 0. A typed consumer breaks on the env-derived path; the repo's own test struct at provider_onboarding_test.go:263 unmarshals `removed` into a string, so that pattern would fail to parse `{removed: false}`. Pick one type  e.g. keep `removed` as the name-string (empty when nothing removed) and convey "not persisted" via the existing `message`/`persisted` fields.

Non-blocking: the env-derived `use`/`remove` `--json` payloads also omit `configPath`/`keyRemoved` that the success payload always includes  worth aligning the field set or documenting the env-derived shape as a distinct variant.

Requesting changes for 1–3.

Vasanthdev2004:
fix coderabbit changes @PierrunoYT

Vasanthdev2004 (changes requested):
Really close now — all three of my asks landed. The rename guard with the typo fall-through, the edit-save guard in the provider manager, and the remove --json payload keeping `removed` a string plus picking up `configPath`/`keyRemoved` all look right, and the schema test on the remove payload is a nice touch. The malformed-config propagation and keeping the credential cleanup on the persisted-delete path only are also correct.

One thing left, and it's the same trap we just closed on remove: the new env-derived rename path emits `renamed: false` while the success path emits `renamed` as a `{from, to}` object (provider_onboarding.go:593). Same key, two JSON types, both exit 0 — a typed consumer parses one and chokes on the other. Emit `null` (or an empty object) there and I'm good to approve. The `use` payload omitting `configPath` is still take-it-or-leave-it from my side.

Vasanthdev2004 (approved):
Re-verified on the current head (807c8268), and all three of my earlier asks landed cleanly:

- The `rename` path now has the `ProviderPersisted` guard with the same typo fall-through as `use`/`remove`, so a genuinely unknown name still hits the real "not found".
- The TUI edit-save path (`saveManagerEdit`) is guarded too, rejecting an env-derived row with a clear "no saved profile to edit" message instead of the bogus not-found.
- The `--json` schema is consistent now: `removed` is a string on both paths (name on success, "" on the env path) and `renamed` is an object on success and `null` (not `false`) on the env path, so a typed consumer parses both. The env-derived `use` payload also picks up `configPath`.

I walked every call site of the five name-keyed config mutators (RemoveProvider, SetActiveProvider, SetProviderModel, RenameProvider, EditProvider) to be sure the class fix is actually complete: all six reachable ones are guarded, the credential-store cleanup still runs only on the persisted delete path, and the persisted paths are unchanged. Build, vet, and the config/cli/tui tests are green, including the four new env-derived onboarding tests.

One optional, non-blocking note for later (nothing to do here): `switchProviderModel` in command_center.go makes two name-keyed persist calls that are safe today only because they discard the error with `_, _ =`, unlike `persistSelectedModel` which you guarded explicitly. It behaves correctly for env-derived providers (the persist just no-ops), so it is not a bug, but if a future change ever starts checking those errors the same not-found would resurface. Worth a guard for symmetry someday, not in this PR.

Nice work seeing this through the rounds. Approving.

kevincodex1 (approved):
LGTM

PR #717, opened 2026-07-17
perf(output): add token-aware semantic output budgeting (PR11)

gnanam1990:
## Summary

Implements **PR11 of the Zero Agent Performance Program**: token-aware semantic output budgeting.

Large tool outputs are now reduced using deterministic, category-specific semantic policies instead of relying only on positional byte truncation.

### Added

- Shared output-budgeting contract
- Deterministic provider-neutral token estimator
- UTF-8-safe output reduction
- Semantic policies for file reads, search results, test output, process output, diffs, and worker results
- Common budgeting boundary after existing secret redaction
- Backward-compatible truncation metadata propagation
- Compact output-budget trace metadata
- Core tool output-category declarations
- Architecture documentation

## Safety properties

This PR preserves existing permissions, sandboxing, secret scrubbing, hard output ceilings, spill protections, transcript ordering, exactly one result per tool call, provider behavior, and concurrent-read scheduling.

Semantic budgeting occurs after existing redaction. Existing byte ceilings remain the final authoritative safety limit.

## Scope boundaries

This PR does not add full subprocess stream capture, a new artifact database, provider-specific tokenizers, native compaction, scheduling changes, model routing, worker/swarm architecture changes, or persistent-process buffer redesign.

## Verification

- `gofmt` clean
- `git diff --check`
- `go vet ./...`
- `go test -race ./internal/tools/... ./internal/agent/...`
- `make build`
- `govulncheck ./...`
- No new lint findings in changed files

## Performance expectations

Targets lower model-visible tool-output token usage, improved context efficiency, and better semantic quality for truncated results.

No wall-clock latency improvement is claimed without benchmark evidence.

## Dependencies

- #705
- #715

Both are merged.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic, category-aware semantic output budgeting with consistent head/tail retention and improved spill handling.
  * Tool output-budget decisions are now included in run traces.
* **Bug Fixes**
  * Improved cancellation behavior so the run emits exactly one tool result with correct truncation-aware metadata.
* **Documentation**
  * Updated the tool execution lifecycle and output budgeting documentation to describe staged redaction, budget enforcement, and spill semantics.
* **Changes**
  * CLI JSON and stream-JSON outputs now consistently report tool-result truncation status.
* **Tests**
  * Added coverage for output-budget propagation, post-tool rebudgeting, trace emission/NDJSON parsing, and exec writer truncation reporting.


gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit approve

Vasanthdev2004 (changes requested):
Really like the direction here — semantic, category-aware retention is a real step up from byte-prefix truncation, and the safety-sensitive parts hold up: budgeting runs strictly on already-redacted bytes (the scrub → budget → byte-ceiling order in `RunWithOptions` is right), the byte ceiling stays the authoritative backstop, and every reduction path is rune-safe through `utf8Prefix`/`utf8Suffix`. Build/vet/gofmt are clean for me, targeted tests pass, and CI is green across the matrix. One thing blocks it though.

**Blocker — single-line files come back with zero content on the registry path.** `read_file` and `read_minified_file` now pass `maxBytes=0` and defer to the registry's file policy, and `budgetFileLines` retains *whole lines only*. Any line longer than the 128 KiB ceiling is never selectable, and because the header-only result already "fits" the budget, `budgetSemanticOutput` never falls through to `budgetDefaultOutput` (output_policies.go:39) — so a file whose content is essentially one big line returns the header and nothing else.

Reproduced against a 200 KB single-line file through `registry.Run("read_file", …)`:

```
File: big.min.js (1 lines)

[zero] ... 1 section(s) omitted ...
[zero] full output received by budgeting layer saved to <spill> (grep or read_file it instead of re-running)
```

290 bytes, none of it content. Pre-PR the same read returned ~128 KiB of head content. The full text is spilled to disk so nothing is truly lost, but the model gets nothing inline — and this is precisely `read_minified_file`'s reason to exist (minified bundles are single-line), so it guts that tool for its primary input. It's also uncovered: the file-policy tests all use multi-line input.

The fix is small and matches the shape you already have — when the line policy can only keep the header/markers, fall through to `budgetDefaultOutput`, which utf8-slices the oversized line and preserves a bounded head+tail. e.g. have `budgetFileLines` return `""` when nothing but the header survives so the existing fallback at output_policies.go:39 fires.

Two smaller, non-blocking notes while you're in here:

- `read_minified_file.go:109` stamps `emitted_bytes` with the full pre-budget length on the registry path, and `addOutputBudgetMetadata` only writes that key when it's absent — so once the registry truncates, `emitted_bytes` over-reports what was actually emitted. `output_budget_retained_bytes` is correct; `emitted_bytes` just goes stale.
- `searchResultFile` (output_policies.go:381) splits on `:` and expects the line number in the second field, but a Windows absolute grep line like `C:\proj\a.go:12: match` puts the drive-letter colon first, so file detection returns `""` and the search policy's cross-file coverage falls back to head/tail files on a >64 KB multi-file grep. Matches still survive behind the omission marker so it's minor, but you're Windows-first — worth parsing the `:<lineno>:` field from the right or stripping a leading drive prefix.

Everything else I traced checked out: the token estimator is deterministic and byte-ceiling-backstopped (whitespace-as-zero-tokens is a fine, documented choice), the trace metadata round-trips losslessly with no absent-vs-zero ambiguity, the cancellation path emits exactly one result with a valid tool-call id, and scope is disciplined with no new deps. Happy to approve once single-line reads keep their content.

anandh8x (changes requested):
I found two blocking issues against the PR11 scope and release gates:

1. **[P1] Keep `read_file` capture bounded before semantic reduction** (`internal/tools/read_file.go:95-133`)

   Registry execution passes `maxBytes == 0`, so `outputBudgetBuilder.WriteString` retains the entire requested range in memory and only applies the 128 KiB budget later at the registry boundary. Before this PR, the streaming builder stopped retaining content at `readOutputBudgetBytes`. Reading a large file with many ordinary lines can now allocate memory proportional to the whole requested range even though only 128 KiB is model-visible. This weakens the existing hard-ceiling protection and can turn a routine `read_file` call into a large memory/latency spike. Please preserve a bounded capture path (for example, bounded head/tail storage or a spill-backed streaming representation) while still giving the semantic policy access to both ends.

2. **[P1] Apply semantic policies to the primary shell/test outputs** (`internal/tools/output_boundary.go:81-112`, `internal/tools/output_ceiling.go:28-39`)

   `bash` and `exec_command` implement `selfBudgeting`, so the registry calls `annotateSelfBudgetedOutput`, which only records a category and explicitly does not run `budgetSemanticOutput`. Consequently, a large `go test`, `pytest`, or normal process result still uses the existing positional truncation; the new test/process retention policies are not applied to the tools that produce those outputs most often. This leaves a central PR11 deliverable (semantic policies for tests and shell/process output) inactive while traces report `test`/`process` categories as though semantic budgeting occurred. Please integrate semantic reduction with the existing capture-aware budgets, or narrow the PR scope/claims and trace vocabulary so they do not imply an applied policy.

Verification performed on the exact PR head (`c90b984a`): targeted output-budget tests passed; `go fmt ./...`, `go vet ./...`, and `govulncheck@v1.3.0 ./...` passed. The repository-wide pinned lint command reports 36 existing findings, none in files changed by this PR. GitHub CI is green.

anandh8x (changes requested):
Thanks for the follow-up. Commit `b9e811e5` correctly fixes the single-line/minified-file fallback, refreshes `emitted_bytes` after semantic reduction, and handles basic Windows drive-letter paths. The focused new tests pass. Three issues remain:

1. **[P1] `read_file` capture is still unbounded before semantic reduction** (`internal/tools/read_file.go:95-133`)

   The registry path still passes `maxBytes == 0`, so `outputBudgetBuilder` retains the entire requested range and the 128 KiB limit is enforced only afterward. The single-line fix improves retained content but does not address this memory/latency regression for large multi-line files. Please preserve bounded head/tail capture or use a spill-backed streaming representation before the registry policy runs.

2. **[P1] Shell/test semantic policies are still not applied** (`internal/tools/output_boundary.go:81-112`, `internal/tools/output_ceiling.go:28-39`)

   `bash` and `exec_command` remain `selfBudgeting`, so their registry path calls `annotateSelfBudgetedOutput`, which only labels the result and explicitly does not call `budgetSemanticOutput`. Large test and process results therefore still use positional truncation even though traces report `test`/`process` categories. This remains inconsistent with PR11's test/shell output-policy deliverable. Please integrate semantic retention with their existing capture-aware budgets, or narrow the PR's claims and trace vocabulary.

3. **[P2] Parse the grep line-number field from the left after an optional drive prefix** (`internal/tools/output_policies.go:400-417`)

   Searching for a numeric field from the right misclassifies numeric colon-delimited content as part of the path. For example:

   ```text
   C:\workspace\a.go:12: error code:123: failed
   ```

   currently returns `C:\workspace\a.go:12: error code` as the file instead of `C:\workspace\a.go`. I reproduced this with a focused package test. Please skip an optional `X:` drive prefix and select the first subsequent `:<line-number>:` delimiter (or otherwise parse the actual grep record structure).

Verification on `b9e811e5`: focused updated tests pass; `go fmt ./...`, `go vet ./...`, and `govulncheck@v1.3.0 ./...` pass. The pinned repository-wide lint command continues to report 36 pre-existing findings, none in this fix commit. GitHub CI is green.

anandh8x (changes requested):
Thanks — `ea27648d` resolves both earlier P1 findings: registry `read_file` now has bounded head/tail capture, and `bash`/`exec_command` now pass their capture-bounded output through semantic policies. The focused tests for those paths pass. Two issues remain:

1. **[P1] Reapply the self-managed `exec_command` budget after hook feedback** (`internal/tools/output_boundary.go:81-89`)

   `RebudgetAfterHook` always calls `applyRegistryOutputBudget`, so an `exec_command` result that receives after-tool feedback is reprocessed with the generic 16,000-token ceiling rather than the call's `max_output_tokens`. I reproduced this by registering a self-managed `exec_command`, setting `max_output_tokens: 10`, and rebudgeting a 1,000-byte post-hook result: all 1,000 bytes survive instead of the configured maximum of 40 bytes. The inverse also happens for requested budgets above 16,000 tokens. Please branch on `selfBudgeting` here just as `Registry.RunWithOptions` does and call `applySelfManagedOutputBudget` for those tools before the final hard ceiling.

2. **[P2] The numeric-content Windows grep case is still unfixed** (`internal/tools/output_policies.go:400-417`)

   The right-to-left parser still mistakes numeric colon-delimited match content for the line-number field. For example:

   ```text
   C:\workspace\a.go:12: error code:123: failed
   ```

   returns `C:\workspace\a.go:12: error code` as the file rather than `C:\workspace\a.go`. I reproduced this with a focused package test. Please parse the first `:<line-number>:` after an optional Windows drive prefix, or parse the grep record structurally.

Verification on `ea27648d`: focused tools/agent tests pass; `go fmt ./...`, `go vet ./...`, and `govulncheck@v1.3.0 ./...` pass. The pinned lint command reports the same 36 repository-wide baseline findings, none in this commit. Performance, security, CodeQL, CodeRabbit, and Zero Review checks are green; the three platform smoke checks were still running at review time.

Vasanthdev2004 (commented):
Re-verified on the current head (`ea27648d`). My blocker is resolved — the single-line/minified-file content drop is gone: I re-ran the same 200 KB single-line read through the registry and it now returns the full 128 KiB of head+tail content instead of 290 bytes of header. `b9e811e5` nailed it, and it also refreshed the stale `emitted_bytes` and handled the Windows drive-letter path I flagged. Quick turnaround — thanks. No remaining blockers of my own.

Two things on the current head worth flagging so they don't get lost:

- **coderabbit's latest changes-requested is a false positive.** The "duplicate `semanticBashFakeTool` declaration → test package won't compile" isn't real: there's a single declaration at `output_boundary_test.go:268` plus its two method definitions, not a redeclaration. `go build ./...` and `go test -run '^$' ./internal/tools/` both compile clean on `ea27648d`. Not worth chasing.
- **`ea27648d` already covers anand's two P1s.** `read_file` now does bounded head/tail capture on the registry path (`newHeadTailOutputBudgetBuilder`) instead of materializing the whole range, and `bash`/`exec_command` get semantic retention inside their capture budget rather than just a metadata label. Good.

The one item I still see open from anand's pass is the **P2 grep parse** — `output_policies.go` wasn't touched here, so a line like `C:\proj\a.go:12: error code:123: failed` still mis-splits. Parsing the `:<lineno>:` field left-to-right after an optional `X:` drive prefix (anand's suggestion) closes that and my earlier drive-letter case in one go.

So from my side this is resolved. Build/vet/gofmt/tools tests are clean for me on `ea27648d` (only the known Windows temp-path sandbox tests fail locally, unrelated to this PR). I'll flip to approve once anand's re-pass on this head lands and CI is green.

gnanam1990:
@coderabbit review

gnanam1990:
@coderabbit full review again

anandh8x (dismissed):
Approved on `e93519e2`.

The follow-up commits resolve all review findings:

- bounded head/tail capture prevents `read_file` from materializing arbitrarily large requested ranges;
- semantic test/process policies now apply to capture-bounded `bash` and `exec_command` output;
- after-tool feedback preserves `exec_command`'s call-specific `max_output_tokens`;
- single-line/minified file content, final emitted-byte metadata, and Windows grep paths with numeric colon-delimited match text are covered by regressions.

Verification: focused tools/agent regression tests pass; `go fmt ./...`, `go vet ./...`, and `govulncheck@v1.3.0 ./...` pass. Ubuntu, macOS, Windows, performance, security, Zero Review, and CodeQL checks are green. The pinned repository-wide lint command continues to report the same 36 baseline findings, none introduced by this PR.

gnanam1990:
@coderabbit if approvable then approve it

Vasanthdev2004 (dismissed):
Converged — re-verified on `e93519e2d` and lifting my earlier changes-requested.

Everything I flagged is resolved, and I confirmed it on this head rather than taking it on trust:

- The single-line/minified read regression is gone — a 200 KB one-line file comes back with the full 128 KiB of content again.
- The grep drive-letter parse is fixed: `C:\workspace\a.go:12: error code:123: failed` now resolves to `C:\workspace\a.go` instead of swallowing the numeric match body into the path.

`go build`/`vet`/`gofmt` and the tools suite are green for me (only the known Windows temp-path sandbox tests fail locally, unrelated to this PR), and the required CI checks are all green. anand's hook-budget and search-path fixes landed and he's approved, so the two P1s and the P2 from his pass are closed as well.

Nice turnaround. Approving — reads ready to merge to me; over to kevin for the merge.

One earlier note that's now purely informational: coderabbit's "duplicate `semanticBashFakeTool`" changes-requested was a false positive (there's a single declaration and the package compiles clean), so it's worth resolving/dismissing so it doesn't linger as an apparent blocker.

Vasanthdev2004:
Quick triage of coderabbit's latest pass (the only thing still showing changes-requested), so it doesn't read as a real merge blocker:

- **"Duplicate `selfBudgeting` interface declaration → compilation fails" (output_ceiling.go:33) is a false positive.** There's exactly one declaration in the package (`git grep 'type selfBudgeting'` → single hit at output_ceiling.go:33) and `go build ./...` / `go vet ./internal/tools/` are clean on `e93519e2d` — a genuine duplicate package-scope declaration wouldn't compile. Same false alarm as the earlier `semanticBashFakeTool` one; safe to dismiss.
- The other three are minor and non-blocking: the `script`/`shell` command aliases aren't resolved in `bashTool`/`execCommandTool.outputCategory` (so those alias forms fall through to the default category — a small classification-consistency follow-up, worth doing but not urgent), the metadata refresh on the non-truncating existing-output path, and the weak no-output assertion in `output_budget_test.go`. Fine to fold into a later cleanup.

My approve stands and so does anand's — the CHANGES_REQUESTED decision is purely coderabbit's advisory block on the above. Over to kevin.

gnanam1990:
@coderabbit full review

gnanam1990:
@coderabbit approve

Vasanthdev2004 (approved):
Re-verified and re-approving on `a5553394f` — my earlier approve was auto-dismissed when the last two commits landed.

Those two commits close out the coderabbit nits from my triage above: the `script`/`shell` alias resolution in `bash`/`exec_command` `outputCategory`, the truncation-metadata refresh on the existing-output path, and the trace test now decodes the `output_budget` record and asserts only the documented metadata keys (instead of the string check that always passed). coderabbit re-reviewed this head and approved.

Confirmed on `a5553394f`: single-line/minified reads still retain full content (128 KiB), `go build`/`vet`/`gofmt` and the tools + trace suites are green for me (only the known Windows temp-path sandbox tests fail locally, unrelated), and the required CI checks are all green. Nothing regressed.

Approving again — ready for kevin to merge.

anandh8x (approved):
Re-approved on current head `a5553394` after reviewing the two commits added since the prior approval.

The new changes are narrow and correct:

- all accepted `bash`/`exec_command` command aliases now share execution and output-category resolution;
- existing truncation state and final emitted/token metadata survive redaction and hook rebudgeting;
- output-budget trace tests now validate the complete serialized metadata schema and values and guard against raw-output leakage.

Verification: focused tools, trace, and agent regression tests pass; `go fmt ./...`, `go vet ./...`, and `govulncheck@v1.3.0 ./...` pass. All GitHub CI, platform smoke, performance, security, CodeQL, Zero Review, and CodeRabbit checks are green. The pinned repository-wide lint command continues to report the same 36 baseline findings, none introduced by these commits.

PR #719, opened 2026-07-17
Add deterministic completion policy

anandh8x:
## Summary
- extract headless completion decisions into a typed deterministic policy with uncertain, complete, and incomplete outcomes
- preserve bounded continuation handling and stale-plan protection
- skip semantic verification when local evidence is sufficient, while allowing one task-grounded check for self-correcting profiles
- add focused policy tests alongside the existing plan-stall regression corpus

## Verification
- make build
- make test
- go fmt ./...
- go vet ./...
- make lint
- go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...
- pinned golangci-lint on ./internal/agent/...: 0 issues

The repository-wide pinned golangci-lint command still reports 36 pre-existing unrelated findings, including findings from the old /tmp/zero-pr717-head checkout.

## Summary by CodeRabbit

* **New Features**
  * Improved assistant completion handling with bounded continuation prompts.
  * Added optional task-grounded semantic checks for self-correcting runs.
  * Better recognition of incomplete responses and pending work before finalizing.

* **Documentation**
  * Clarified completion-signal behavior across interactive and self-correcting run profiles.

* **Tests**
  * Added coverage for completion decisions, continuation limits, pending plans, and semantic checks.

### DIFF
--- a/internal/agent/completion_policy.go
+++ b/internal/agent/completion_policy.go
@@ -1,0 +1,86 @@
+package agent
+
+// CompletionDecision is the deterministic outcome of evaluating a text-only
+// assistant turn. Uncertain means the loop needs one bounded follow-up action
+// before it can finalize the run.
+type CompletionDecision string
+
+const (
+	CompletionUncertain  CompletionDecision = "uncertain"
+	CompletionComplete   CompletionDecision = "complete"
+	CompletionIncomplete CompletionDecision = "incomplete"
+)
+
+type completionAction string
+
+const (
+	completionActionNone          completionAction = ""
+	completionActionContinue      completionAction = "continue"
+	completionActionSemanticCheck completionAction = "semantic_check"
+)
+
+type completionEvaluation struct {
+	Decision CompletionDecision
+	Action   completionAction
+	Reason   string
+}
+
+// completionPolicy owns the small amount of state needed to keep uncertain
+// decisions bounded. requireSemanticCheck is set only for run profiles that
+// already opted into self-correction; ordinary completion remains entirely
+// local and adds no model call.
+type completionPolicy struct {
+	continueNudges         int
+	requireSemanticCheck   bool
+	semanticCheckRequested bool
+}
+
+func newCompletionPolicy(requireSemanticCheck bool) *completionPolicy {
+	return &completionPolicy{requireSemanticCheck: requireSemanticCheck}
+}
+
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (approved):
Clean PR9. This is a faithful extraction — I traced every branch of the new `completionPolicy.evaluate` against the old inline logic and they line up exactly: a self-reported admission → `Incomplete` (checked first, so an admitted-impossible task isn't wasted on continue-nudges), a continuation cue / pending plan → bounded continue nudge, then cue→`Incomplete` while plan-only falls through, `SelfCorrect` → exactly one task-grounded semantic check, else → `Complete`. The bounded state (`continueNudges`, `semanticCheckRequested`) moves out of loop-locals into a policy created once per run, so it persists across turns identically, and everything stays behind `RequireCompletionSignal` so default/interactive runs are byte-identical.

The acceptance criteria hold up: the existing completion/plan-stall regression corpus passes unchanged (`TestCompletionGate*`, `TestAcceptance*`, `TestPendingPlanAloneDoesNotForceIncomplete`, `TestMaxTurnsCutoffIsIncompleteUnderGate`), and the new policy tests cover the decision surface — local-evidence-decides-without-a-semantic-check, bounded plan-stall protection, pending-plan-as-weak-evidence, and exactly-one-required-semantic-check. `go build`/`vet`/`gofmt` clean and the full agent suite is green for me (only the known Windows temp-path sandbox tests fail locally, unrelated). It's a genuine improvement too — the loop sheds ~40 net lines and the completion logic is now typed and unit-testable in isolation.

One tiny take-it-or-leave-it note: `evaluate` never returns `Uncertain` with `completionActionNone`, so the loop's inner action switch always matches today — but if that ever changed, the `continue` would re-loop with no new message appended. A `default` arm (or an explicit assert) would make that impossible to regress. Not blocking.

Approving. The only thing still outstanding is the Windows smoke check finishing — the macOS/Ubuntu smokes are green and my local Windows build+tests pass, so I expect it to go green; over to kevin once it does.

gnanam1990 (approved):
Reviewed the completion-policy extraction against the previous inline state machine. Bounded continuation handling, stale-plan behavior, self-reported incompletion, the one-time self-correct acceptance check, and default interactive behavior are preserved. Focused tests, build, vet, lint, and all CI checks pass. No blocking findings.

PR #720, opened 2026-07-17
perf(providers): provider capabilities and default turn-session adapter (PR7)

Vasanthdev2004:
## Summary

Implements **PR7 of the Zero Agent Performance Program**: provider capabilities and the default turn-session adapter — the architecture seam that PR8 (one optimized provider session) and a future native-compaction session plug into without touching the agent loop again.

### Added

- `zeroruntime.TurnSession` — one provider conversation per run: `Prewarm` (best-effort, no-op by default), `Stream` (signature-identical to `Provider.StreamCompletion`), `Compact` (returns `ErrCompactionUnsupported` on the default adapter; dormant until a native-compaction session exists), `Close` (idempotent no-op by default).
- `zeroruntime.TurnSessionProvider` — opens a session for a run and exposes `ProviderCapabilities`.
- `zeroruntime.ProviderCapabilities` — a flat projection of the resolved model's static capabilities (context window, max output tokens, vision/reasoning/prompt-cache flags, reasoning efforts). Deliberately plain data: `modelregistry` imports `zeroruntime`, so typed registry fields here would form an import cycle. `providers.NewTurnSessionProvider` does the projection instead.
- `providers.NewTurnSessionProvider` — builds the default adapter for **every current provider kind** (openai, openai-compatible, anthropic, anthropic-compatible, google, and the chatgpt path via `New`), populating capabilities from the same registry entry `New` already resolves.
- Loop integration: `agent.Run` opens a session at run start (before session hooks, so a failed open is a clean run-start error with nothing to unwind), routes **all** provider I/O through it via a one-line `Provider` shim — per-turn streams, mid-stream retries, the compaction summarizer, and the final-answer request — and closes it at teardown. A mid-run model switch closes the old session and opens a fresh one over the escalated provider.
- `Options.TurnSessionProvider` (optional): nil keeps the default wrapper over the passed provider, so **every existing caller is unaffected** — no signature changes anywhere.

## No performance change expected

This PR is a pure extension point. The default session's `Stream` *is* `provider.StreamCompletion`; `Prewarm`/`Close` are no-ops; no extra network round-trips, no hot-path allocation beyond a value-typed shim. Behavior identity is asserted by tests, not claimed: the same scripted run with the default path and an explicit adapter produces identical results and request counts.

## Safety properties

Preserves permissions, sandboxing, secret scrubbing, output ceilings, transcript ordering, exactly one result per tool call, reconnect/stall-retry behavior (`streamWithReconnect` is unchanged and still wraps every stream), and cancellation semantics. A `TurnSession.Close` failure never fails the run; a `Prewarm` failure is ignored by contract.

## Scope boundaries

No connection reuse, prewarm implementation, request-prefix fingerprint reuse, provider-specific sessions, or native compaction — those are PR8 and later, behind this seam. The `ModelSwitcher` contract still returns a bare `Provider`, so an optimized session is not carried across a mid-run swap (documented next to the existing context-window limitation; fixing both needs the same switcher contract change).

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` clean
- `go test ./internal/zeroruntime/ ./internal/providers/... ./internal/agent/ ./internal/cli/ ./internal/tui/` green
- New tests: default-adapter forwarding is verbatim (request and events), `Prewarm`/`Close` no-op + idempotent, `Compact` unsupported; fault injection — a failed `OpenTurnSession` surfaces as a wrapped run-start error with zero provider requests, a failing `Close` is swallowed; every request of a run flows through the session (prewarmed once, closed once); the default path is behavior-identical to an explicit adapter; a mid-run model switch closes the old session exactly once and continues on the new provider; every provider kind wraps; registry capabilities project correctly.
- Existing escalation, completion-gate, and reconnect test suites pass unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added per-run turn sessions to centralize streaming and lifecycle handling.
  * Introduced `TurnSessionProvider` and `ModelSessionSwitcher` options for target-aware mid-run escalation.
  * Added capability projection for sessions (limits, support flags, reasoning tiers).
* **Reliability**
  * Turn session open failures surface cleanly; teardown close errors no longer fail runs.
  * Mid-run model/session switching continues on errors and records a user-visible notice.
* **Tests**
  * Expanded coverage for session routing, lifecycle safety, capability projection, and escalation switching behavior.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -737,23 +737,35 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		// of the run. At most one switch per turn (first request wins). A switcher
 		// error is NON-FATAL — record a brief note and continue on the current
 		// model. nil switcher ⇒ requests are ignored entirely (escalation off).
-		if turnRequestedModel != "" && options.ModelSwitcher != nil {
-			newProvider, switchErr := options.ModelSwitcher(ctx, turnRequestedModel)
+		if turnRequestedModel != "" && (options.ModelSessionSwitcher != nil || options.ModelSwitcher != nil) {
+			// Resolve the escalated model's session source. The target-aware
+			// ModelSessionSwitcher is preferred: its TurnSessionProvider carries an
+			// optimized session (and capabilities) across the swap. The legacy
+			// ModelSwitcher fallback returns a bare Provider, wrapped in the
+			// default no-op session — identical to pre-seam behavior.
+			var newSessions zeroruntime.TurnSessionProvider
+			var switchErr error
+			if options.ModelSessionSwitcher != nil {
+				newSessions, switchErr = options.ModelSessionSwitcher(ctx, turnRequestedModel)
+			} else {
+				var newProvider Provider
+				newProvider, switchErr = options.ModelSwitcher(ctx, turnRequestedModel)
+				if switchErr == nil && newProvider != nil {
+					newSessions = zeroruntime.NewProviderTurnSessionProvider(newProvider, zeroruntime.ProviderCapabilities{})
+				}
+			}
 			if switchErr != nil {
 				messages = append(messages, zeroruntime.Message{
 					Role:    zeroruntime.MessageRoleUser,
 					Content: escalationFailedNoticePrefix + " (" + turnRequestedModel + "): " + switchErr.Error() + ". Continuing on " + options.Model + ".",
 				})
-			} else if newProvider != nil {
-				// Open a fresh default session over the escalated provider and close
-				// the old one, so the next turn's streams and compaction use the new
-				// model. The default open never errors and its Close is a no-op, so
-				// this stays byte-identical to the old direct reassignment; the
-				// guarded open matters once a session holds real state. Note the
-				// switcher returns a bare Provider, so an optimized session is not
-				// carried across a swap — acceptable for the same reason as the
-				// context-window limitation below (a switcher contract change).
-				newSession, openErr := zeroruntime.NewProviderTurnSessionProvider(newProvider, zeroruntime.ProviderCapabilities{}).OpenTurnSession(ctx)
+			} else if newSessions != nil {
+				// Open the new session and close the old one, so the next turn's
+				// streams and compaction use the new model. For the default adapter
+				// the open never errors and Close is a no-op, so this stays
+				// byte-identical to a direct provider reassignment; the guarded open
+				// and prewarm matter once a session holds real state.
+				newSession, openErr := newSessions.OpenTurnSession(ctx)
 				if openErr != nil {
 					messages = append(messages, zeroruntime.Message{
 						Role:    zeroruntime.MessageRoleUser,
@@ -769,8 +781,8 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 					// KNOWN LIMITATION (deferred): the compactor's context-window budget
 					// is fixed at run start from options.ContextWindow and is NOT updated
 					// here, so a switch to a model with a different window keeps compacting
-					// against the original budget. Fixing it needs a ModelSwitcher contract
-					// change (return the new window) — out of scope for this change.
+					// against the original budget. Fixing it needs the switcher to also
+					// report the new window — out of scope for this change.
 				}
 			}
 		}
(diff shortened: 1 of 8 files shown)


Vasanthdev2004:
Addressed the review in `96ac8689` — one fix, one deliberate defer:

- **Effective reasoning efforts: fixed.** `resolveCapabilities` now reads efforts through `Registry.ReasoningEfforts` instead of the raw entry, so the projection matches what the /effort picker and run-time resolver advertise — including the name-inferred fallback for catalog entries that enumerate no efforts of their own. Added `TestNewTurnSessionProviderUsesEffectiveReasoningEfforts` with a gpt-5-family fixture whose entry lists no efforts but projects `[minimal low medium high]`. Good catch — the whole point of the flat projection is that it agrees with what the rest of the product reports.

- **Optimized session across a mid-run model switch: deliberately out of scope.** The switch path wrapping the escalated provider in a default session is the documented limitation, not an oversight: `ModelSwitcher` returns a bare `Provider`, so carrying an optimized session across a swap needs a switcher contract change — the same change the pre-existing context-window limitation right below it has been waiting on. In this PR there is no optimized session implementation at all (the default adapter is the only one), so the limitation is unreachable today; it becomes real work when the optimized session lands, and the contract change belongs there. The suggested distinct-switched-session test needs that same contract to be observable, so it lands with it. Both limitations are called out in the same comment block so they get fixed together.

CI is green on the head with the fix.

anandh8x (dismissed):
Reviewed the latest head (96ac8689). PR7's required seam is implemented cleanly: every run routes provider I/O through a TurnSession without provider-specific loop branching; the default adapter preserves current behavior; lifecycle failures, prewarm/close behavior, capability projection, and model switching are covered by focused tests. The effective reasoning-effort projection fix is correct.\n\nValidated locally: git diff --check; go test ./internal/providers ./internal/zeroruntime ./internal/agent; and the broader agent/CLI/TUI suites on the preceding code commit. No blocking findings.\n\nNon-blocking PR8 follow-up: when an optimized session exists, extend the ModelSwitcher contract/session resolver so escalation preserves optimized session semantics instead of falling back to the default adapter, with lifecycle coverage for the switched session.

Vasanthdev2004:
@coderabbitai review

gnanam1990 (dismissed):
Reviewed the latest head (96ac8689). The provider turn-session seam routes run I/O through the session while the default adapter preserves existing behavior. The reasoning-effort capability projection now uses Registry.ReasoningEfforts and includes coverage for inferred effective tiers. Focused lifecycle/provider validation and all CI checks pass. No blocking findings.\n\nNon-blocking PR8 follow-up: preserve optimized TurnSessionProvider semantics across ModelSwitcher escalation instead of falling back to the default adapter, with lifecycle coverage for the switched session.

Vasanthdev2004:
Took the second suggestion after all — `968cd1ee` implements the target-aware session resolver instead of deferring it:

- New optional `Options.ModelSessionSwitcher func(ctx, modelID) (zeroruntime.TurnSessionProvider, error)` — the loop prefers it over `ModelSwitcher`, so an optimized session (and its capabilities) survives a mid-run model switch. `ModelSwitcher` remains as the unchanged fallback (bare `Provider` wrapped in the default no-op session), same non-fatal error contract, so every existing caller and test is untouched.
- The requested test is in: `TestRunModelSessionSwitcherCarriesOptimizedSession` uses a distinct switched session and asserts it receives the post-switch streams, is prewarmed exactly once, and is closed exactly once at teardown — plus that the legacy switcher is never consulted when both are set. `TestRunModelSessionSwitcherErrorIsNonFatal` covers the error path (run continues on the current session with a transcript note).
- The known-limitation comment now covers only what genuinely remains deferred: the compactor's context-window budget across a switch, which needs the switcher to also report the new window.

The full escalation suite passes unchanged, so the fallback path is behavior-identical. CI running on the new head.

Vasanthdev2004:
@coderabbitai approve

gnanam1990 (approved):
Re-reviewed at 39d0136. No blocking findings. The optimized turn-session handoff and swap-time open-failure handling are sound, with focused regression coverage. Local race tests, vet, build, formatting, and static lint passed; all GitHub checks are green.

anandh8x (approved):
Re-reviewed only the commits added since the dismissed approval, through head 39d0136. The new ModelSessionSwitcher preserves optimized turn-session semantics across escalation, and the added error/open-failure paths correctly retain and eventually close the original session. Focused race tests and vet pass for internal/agent, internal/providers, and internal/zeroruntime. No blocking findings.

PR #724, opened 2026-07-18
Fix ChatGPT OAuth model discovery and ACP model selection

### LINKED ISSUE #722: ACP cannot discover ChatGPT models and live OAuth model listing returns 401
### Version / commit

Zero commit `18cce358a29409ab937e92ff9af77bf98749f6aa` (`v0.4.0-4-g18cce35`)

### OS and environment

- Windows 11
- PowerShell 7.6.3
- Saved `chatgpt` provider using OAuth login
- Provider model configured as `gpt-5.5`

### Steps to reproduce

1. Select the saved ChatGPT provider.
2. Run:
   ```powershell
   zero providers models chatgpt --json
   ```
3. Observe:
   ```text
   [zero] models endpoint returned 401 Unauthorized: {"detail":"Unauthorized"}
   ```
4. Run:
   ```powershell
   zero models list --json
   ```
5. Observe that only the static built-in catalog is returned; it does not include the configured ChatGPT model (`gpt-5.5`) or a complete list of models available through the subscription.
6. Start `zero acp`, initialize ACP v1, and create or load a session.

### Actual behavior

There is no authoritative model list available to an ACP editor client:

- The live `providers models chatgpt` request fails with 401 under OAuth.
- The static `models list` catalog does not represent all ChatGPT subscription models.
- ACP `session/new` and `session/load` return no `configOptions` model values.
- ACP supports `session/set_config_option` and `_zero/set_model`, but clients must already know or manually enter the model ID.

### Expected behavior

An editor using ACP should be able to discover the models available for the effective provider and then select one safely. Possible solutions:

1. Make `providers models chatgpt` authenticate correctly with the stored ChatGPT OAuth session.
2. Populate ACP `session/new` and `session/load` `configOptions` with model values for the effective provider.
3. Add an ACP or `_zero/*` model-list method that returns IDs, display names, capabilities, and the selected model.

If a provider cannot enumerate models, ACP should at least return the configured model and clearly mark the list as incomplete.

### Why this matters

ACP exposes model-setting methods, so editor integrations naturally render a model selector. Without discoverability, they must either hide model selection, show an incomplete static catalog, or accept arbitrary unvalidated model strings.

### Additional context

The ChatGPT profile is otherwise recognized by `providers list --json` as `openai-compatible`, `oauthLogin: true`, and `status: ok`. No credentials are printed by these commands.

PierrunoYT:
## Summary
- authenticate ChatGPT model discovery with the stored OAuth bearer and required Codex headers
- retry model discovery once with refreshed OAuth credentials after a 401
- expose ACP v1 model and permission-mode config options from session new/load
- validate standard ACP selections while preserving the permissive vendor model override

## Validation
- `go test ./...`
- `go vet ./...`
- `govulncheck ./...`
- focused golangci-lint on changed packages: 0 issues

Repository-wide golangci-lint still reports pre-existing findings outside the changed packages.

Fixes #722

## Summary by CodeRabbit

* **New Features**
  * Added per-session model selection with persistence across session creation and reload.
  * Sessions now present configurable model and permission mode options, backed by a selectable model catalog (with optional live discovery).
  * Improved OpenAI-family model discovery using OAuth and Codex account-aware requests.

* **Bug Fixes**
  * Discovery now handles authentication retry flows more reliably and applies Codex-required headers.
  * Model and mode updates via session config correctly apply to the active session.
  * When discovery fails, only the configured fallback model is offered.

* **Tests**
  * Added coverage for model catalog behavior, discovery fallback, session config option wire format, and model metadata updates.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -695,6 +695,7 @@ func (s *acpSession) currentMode() agent.PermissionMode {
 }
 
 func (s *acpSession) setModel(model string) {
+	model = strings.TrimSpace(model)
 	s.mu.Lock()
 	found := false
 	for _, option := range s.models {
@@ -703,7 +704,7 @@ func (s *acpSession) setModel(model string) {
 			break
 		}
 	}
-	if !found && strings.TrimSpace(model) != "" {
+	if !found && model != "" {
 		s.models = append(s.models, SessionConfigOptionValue{Value: model, Name: model})
 	}
 	s.model = model
@@ -717,6 +718,7 @@ func (s *acpSession) currentModel() string {
 }
 
 func (s *acpSession) hasModel(model string) bool {
+	model = strings.TrimSpace(model)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	for _, option := range s.models {
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/providercatalog"
 	"github.com/Gitlawb/zero/internal/providermodelcatalog"
+	"github.com/Gitlawb/zero/internal/providermodeldiscovery"
 	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/sessions"
 	"github.com/Gitlawb/zero/internal/tools"
@@ -25,9 +26,10 @@ import (
 // means the editor only hosts the thread — ZERO owns BYOK and telemetry-free
 // operation.
 type Deps struct {
-	ResolveConfig func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
-	NewProvider   func(profile config.ProviderProfile) (zeroruntime.Provider, error)
-	RunAgent      func(ctx context.Context, prompt string, provider zeroruntime.Provider, opts agent.Options) (agent.Result, error)
+	ResolveConfig  func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error)
+	DiscoverModels func(context.Context, config.ProviderProfile) ([]providermodeldiscovery.Model, error)
+	NewProvider    func(profile config.ProviderProfile) (zeroruntime.Provider, error)
+	RunAgent       func(ctx context.Context, prompt string, provider zeroruntime.Provider, opts agent.Options) (agent.Result, error)
 	// BuildWorkspace builds the SCOPED tool registry and the sandbox engine for a
 	// validated workspace root, so ACP shell tools (bash/exec_command) are confined
 	// exactly like the exec surface — never run unconfined on the host.
@@ -62,14 +64,16 @@ type acpSession struct {
 	// turnMu serializes prompt turns for one session: concurrent session/prompt
 	// calls run one at a time so they can't interleave history or clobber the
 	// single cancel slot.
-	turnMu sync.Mutex
+	turnMu  sync.Mutex
+	modelMu sync.Mutex
 
-	mu      sync.Mutex
-	mode    agent.PermissionMode
-	model   string
-	models  []SessionConfigOptionValue
-	cancel  context.CancelFunc
-	history []turnRecord
+	mu             sync.Mutex
+	mode           agent.PermissionMode
+	model          string
+	models         []SessionConfigOptionValue
+	restrictModels bool
+	cancel         context.CancelFunc
+	history        []turnRecord
 }
 
 // NewAgent builds the ACP server and registers its method handlers on conn.
@@ -123,7 +127,7 @@ func (a *Agent) handleInitialize(_ context.Context, params json.RawMessage) (any
 
 // ---- session lifecycle ----
 
-func (a *Agent) handleSessionNew(_ context.Context, params json.RawMessage) (any, error) {
+func (a *Agent) handleSessionNew(ctx context.Context, params json.RawMessage) (any, error) {
 	var p NewSessionParams
 	if err := json.Unmarshal(params, &p); err != nil {
 		return nil, RPCError(codeInvalidParams, "invalid session/new params")
@@ -132,15 +136,15 @@ func (a *Agent) handleSessionNew(_ context.Context, params json.RawMessage) (any
 	if err != nil {
 		return nil, RPCError(codeInvalidParams, err.Error())
 	}
-	model, models, err := a.resolveModelChoices(root)
+	model, models, restrictModels, err := a.resolveModelChoices(ctx, root)
 	if err != nil {
 		return nil, RPCError(codeInternalError, "config: "+err.Error())
 	}
-	meta, err := a.deps.Store.Create(sessions.CreateInput{Title: "ACP session", Cwd: root})
+	meta, err := a.deps.Store.Create(sessions.CreateInput{Title: "ACP session", Cwd: root, ModelID: model})
 	if err != nil {
 		return nil, RPCError(codeInternalError, "create session: "+err.Error())
 	}
-	sess := a.registerSession(meta.SessionID, root, nil, model, models)
+	sess := a.registerSession(meta.SessionID, root, nil, model, models, restrictModels)
 	return NewSessionResult{
 		SessionID:     sess.id,
 		ConfigOptions: a.configOptions(sess),
@@ -148,7 +152,7 @@ func (a *Agent) handleSessionNew(_ context.Context, params json.RawMessage) (any
 	}, nil
 }
 
-func (a *Agent) handleSessionLoad(_ context.Context, params json.RawMessage) (any, error) {
+func (a *Agent) handleSessionLoad(ctx context.Context, params json.RawMessage) (any, error) {
 	var p LoadSessionParams
 	if err := json.Unmarshal(params, &p); err != nil {
 		return nil, RPCError(codeInvalidParams, "invalid session/load params")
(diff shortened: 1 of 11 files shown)


gnanam1990 (changes requested):
**Requesting changes** — the ChatGPT OAuth discovery half is correct and well tested, but the ACP half derives `restrictModels` from catalog membership rather than from whether discovery actually produced a list, which is a straight capability regression for providers that have no discovery endpoint. It's a small fix.

Scope note first: the branch is based on `18cce35`, not `fbf8598`. `git diff fbf8598..HEAD` reports 97 files / ~7300 deletions because it also reverses everything main landed after `18cce35` (profile_controller, task_state, turn_session, execprofile, …). The real PR is `git diff 18cce35..HEAD` — 11 files, +592/-71, matching `pr724.diff` exactly. All 11 files are germane; I found no unrelated hunks. This will need a merge/rebase rather than a squash of the two-dot diff.

---

### [Major] Catalog providers without live discovery can no longer change model over ACP
`internal/acp/agent.go:476`

```go
restrictModels := knownProvider && !descriptor.Custom
```

This is computed *before* discovery runs and is never revised based on the outcome. `providermodeldiscovery.Discover` only handles OpenAI/OpenAI-compatible and Anthropic/Anthropic-compat kinds (`internal/providermodeldiscovery/discovery.go:79-88`); for every other catalog transport — `google`, `bedrock`, `vertex` — it unconditionally returns `provider X does not expose model discovery`. For those profiles `restrictModels == true` but `options == [configured model]`, so the guard at `internal/acp/agent.go:422` rejects every other value.

Confirmed with a probe test: a `google` profile advertises `options = [gemini-2.5-pro]`, and `session/set_config_option{configID:"model", value:"gemini-2.5-flash"}` returns `-32602: unknown model: gemini-2.5-flash`. Before this PR the handler was a bare `sess.setModel(p.Value)` with no validation (`pr724.diff:144`), so any model was accepted. Same failure mode hits *every* catalog provider for the duration of any discovery outage.

The discovery error is also discarded at `internal/acp/agent.go:478` with no log or notification, so the user gets a one-entry picker and no explanation.

Fix — only enforce the allow-list when it is actually authoritative:

```go
discoverySucceeded := false
if a.deps.DiscoverModels != nil {
    discovered, discoverErr := a.deps.DiscoverModels(ctx, resolved.Provider)
    if ctx.Err() != nil { return "", nil, false, ctx.Err() }
    if discoverErr == nil && len(discovered) > 0 {
        discoverySucceeded = true
        // …existing add() loop…
    }
    // consider: else log discoverErr so a collapsed picker is diagnosable
}
restrictModels := knownProvider && !descriptor.Custom && discoverySucceeded
```

i.e. false when `DiscoverModels` is nil, errored, or returned zero usable models. Worth a test with a discovery-less catalog provider asserting `set_config_option` still accepts an arbitrary model.

### [Minor] `session/new` and `session/load` block on an unbounded, uncached live network probe
`internal/acp/agent.go:139`

`resolveModelChoices` passes the *connection-level* ctx (`internal/acp/jsonrpc.go:121` — not a per-request ctx, so it's only cancelled at shutdown) straight into `DiscoverModels`. The production path is `defaultDiscoverProviderModels` → `providermodeldiscovery.Discover`, whose default client timeout is 10s (`internal/providermodeldiscovery/discovery.go:238`) and which can issue two attempts via the 401 force-refresh loop (`internal/providers/providerio/auth.go:70-73`); each refresh runs through an OAuth manager with its own 30s client timeout (`internal/cli/oauth_provider.go:47`). Worst case a single `session/new` blocks ~80s. Nothing caches the result, so every new thread re-probes. Requests are dispatched on goroutines (`jsonrpc.go:167`) so the connection isn't stalled, but an editor like Zed shows a hung thread where previously `session/new` was purely local.

Fix: wrap the discovery call in a short `context.WithTimeout` (a few seconds) inside `resolveModelChoices` — the failure path already degrades cleanly to the configured model — and consider caching the per-provider result for the connection's lifetime.

### [Nit] Breaking ACP wire-format change — confirm it's coordinated
`internal/acp/types.go` renames `SessionConfigOptionValue.id` → `value`, `SessionConfigOption.Value` → `currentValue`, `Values` → `options`, and adds required `type`/`category` fields. `TestACPConfigOptionWireSchema` (`internal/acp/agent_test.go:332`) explicitly asserts the old `value`/`values` keys are gone, so this is deliberate — just make sure editor-side consumers are lined up.

---

### Checked and cleared (not findings)

- **`session/load` "silently discarding" the persisted model on a discovery outage.** The fallback is real (`internal/acp/agent.go:180`) but it is not silent: `configOptions` (`agent.go:496-517`) populates the same `session/load` response with `CurrentValue` and the degraded one-entry `Options`, which is ACP's canonical model-display surface — verified by probe. It's also non-destructive and self-healing: the store keeps the user's choice and load restores it once discovery recovers. Writing the fallback back to the store would be strictly worse. (The `warnPersistence` analogy doesn't hold — `LoadSessionResult` at `types.go:142-145` has no history field, so that notification is the *only* channel for history loss, whereas model state is fully represented in the payload.)
- **`_zero/set_model` bypassing the allow-list.** It does pass `restrictModels=false` (`agent.go:413`), but at the merge base *both* paths were entirely unvalidated, so this PR strictly tightens the surface; the bypass is circular (the caller has already set the bogus model directly); and it is not permanent — `agent.go:180` discards an unadvertised persisted model on load for restricted sessions. Contamination is confined to the in-memory session. A one-line comment at `agent.go:413` explaining the intentional asymmetry would help future readers.
- **`sess.restrictModels` read without `sess.mu`.** Not a race by construction: the field (`agent.go:74`) has exactly one write, the composite-literal init in `registerSession` under `a.mu` (`agent.go:654-661`), and is never reassigned. Every reader goes through `a.session()` (`agent.go:667`), which crosses the same mutex. `go test -race -count=3 ./internal/acp/` is clean. The asymmetry with `model`/`models`/`mode` is justified — those *are* mutated post-publication.

### What's genuinely well done

The OAuth/credential half is solid and I found no leak on that surface. `SendWithAuthRetry` resolves auth *before* dispatch, so a resolver error can never let an unauthenticated request go out first (`internal/providers/providerio/auth.go:42-59`), and `withBearer` clears `APIKey` so both auth methods can never be sent together (`auth.go:86`). Discovery errors pass through `redactDiscoveryError` (`discovery.go:452`) plus `redaction.RedactString`. Extracting `CodexAccountResolverForLogin` (`factory.go:863`) is the right call — it binds `chatgpt-account-id` to the exact login key that issued the bearer instead of doing a second independent lookup — and the shared `ApplyCodexHeaders` (`codex.go:29`) stops runtime completions and discovery from drifting apart. I verified the `injectCodexHeaders` refactor is behavior-preserving (`p.originator` is normalized non-empty at `codex.go:111-113`, so the `!= codexDefaultOriginator` guard correctly restores a custom value). `TestDiscoverChatGPTModelsUsesOAuthAndCodexHeaders` is a real test, not a smoke test — it asserts the 401 path re-sends both a refreshed bearer *and* a re-derived account id. `Store.UpdateModel` correctly mirrors `UpdateTitle`'s locking and its deliberate non-touching of `UpdatedAt`/`EventCount`.

### Build / test

`gofmt` clean, `go build ./...` clean, `go vet` clean on all touched packages. `go test` passes for `internal/acp`, `internal/providermodeldiscovery`, `internal/providers/...`, `internal/sessions`. `go test -race -count=3` on those packages: clean, no data races.

`internal/cli` has 5 failures — **all confirmed pre-existing**, re-run with the identical `-run` selectors against the true merge base `18cce35` where they fail identically: `TestRunAddDirDispatchForwardsGrantIntoExecScope`, `TestExecScopeReRegistrationSwapsCoreToolsByName`, `TestRunDoctorFormatsRedactedProviderDiagnostics`, `TestRunDoctorConnectivityProbesProvider`, `TestRunSandboxCheckJSONDeniesOutOfWorkspaceWrite`. The sandbox_check JSON reports `"platform": "windows"` / `"backend.available": false` while running on darwin/arm64, which is the root cause of three of them. None of these touch the two `internal/cli` files this PR changes (`acp.go`, `provider_models.go`). **Nothing is attributable to this PR.**

Housekeeping: `internal/acp/zzrefute_test.go` is an untracked file in the shared worktree that is *not* part of this PR — it ends in an unconditional `t.Fatalf`, so it makes `go test ./internal/acp/` fail by construction. Delete it before reusing the worktree.

gnanam1990 (approved):
## Verdict: APPROVE

The core bug is genuinely fixed. Previously ACP/CLI called `Discover` with empty `Options{}`, so an OAuth-only ChatGPT login sent an unauthenticated request to `/backend-api/codex/models` → 401 → no models. This PR wires `oauthLoginForProfile`'s resolver and the *same* login key into `providers.CodexAccountResolverForLogin`, so the bearer and `chatgpt-account-id` now come from one login and discovery routes through `providerio.SendWithAuthRetry` (1 retry on 401 with force-refresh). The new `TestDiscoverChatGPTModelsUsesOAuthAndCodexHeaders` exercises exactly this path and would fail without the fix. Lock ordering (`modelMu`→`mu`) is sequential with no nested hold, `restrictModels` is set-once before publish, and `-race` is clean. Nice work.

Build/test on the PR HEAD (go1.26.5): `gofmt -l` clean, `go build ./...` and `go vet` pass, and `go test -race ./internal/acp/... ./internal/providermodeldiscovery/... ./internal/providers/... ./internal/sessions/...` all pass.

The `internal/cli` sandbox tests (`TestRunAddDirDispatchForwardsGrantIntoExecScope`, `TestExecScopeReRegistrationSwapsCoreToolsByName`, `TestRunSandboxCheckJSONDeniesOutOfWorkspaceWrite`) FAIL — but they fail *identically* in the clean base worktree (they report an unavailable/"windows" sandbox backend in this macOS environment) and don't touch the PR's changed files. **Pre-existing / environmental, not PR-attributable.**

---

### [Minor] Transient discovery failure on `session/load` silently reverts a restricted-provider user's persisted model to the configured default — PR-introduced
`internal/acp/agent.go:180` (with `resolveModelChoices` at `agent.go:458`)

`resolveModelChoices` only returns early on `ctx.Err()`; a plain discovery error (network blip, 401, API down) falls through to `return selected, options, restrictModels, nil` with `options` containing just the configured default and `restrictModels` still `true`.

Concrete failure: over a ChatGPT profile (`restrictModels=true`, config default `gpt-5.5`), the user selects `gpt-5.4-mini` via `set_config_option`, which `Store.UpdateModel` persists as `meta.ModelID`. On a later `session/load` while discovery is transiently unavailable, the merge at line 180:

```go
persistedModel != "" && (!restrictModels || modelChoiceExists(models, persistedModel))
```

evaluates to `true && (false || false)` = `false`, so the persisted `gpt-5.4-mini` is **not** applied. `registerSession` stores `gpt-5.5`, the advertised `CurrentValue` and `runTurn` both use `gpt-5.5` — a silently substituted model the user did not choose. The stored `ModelID` is untouched, so it self-heals on the next successful load (transient, not permanent data loss).

Fix: distinguish "discovery returned a curated/empty set" from "discovery failed." Have `resolveModelChoices` report whether live discovery actually succeeded, and on the load path honor the persisted selection when discovery was unavailable (it was already validated when the user picked it) — e.g. append and select `meta.ModelID` whenever discovery did not succeed, rather than dropping it because the degraded option set no longer contains it.

---

kevincodex1 (approved):
LGTM

PR #730, opened 2026-07-18
fix(perfbench): absolutize the bench binary and make errored tasks first-class

Vasanthdev2004:
## Summary

Found while producing the Phase 0 baseline report: running the checked-in manifest with a relative `--binary` path (which is also the Makefile default, `./zero`) makes every task die at spawn, and the harness still printed a normal-looking "0/34 (0%)" summary and exited 0. A run that measured nothing was indistinguishable at a glance from a run where the model failed every task. That is exactly the misreadable-result trap the Phase 0 oracle hardening (#712) exists to prevent, one layer down.

Two fixes:

- **`ResolveBinary` pins explicit paths to absolute.** The turn runner sets each child's `cmd.Dir` to a per-task fixture copy, so a relative binary path that stats fine at resolution time fails to spawn from inside every fixture dir with `fork/exec ./zero: file does not exist`. The repo-root fallback branch already returned absolute paths; only the explicit branch was broken.
- **Errored tasks are first-class (schema v4).** New `tasksErrored` counts tasks whose every iteration died before the agent produced a run. The console summary now surfaces it loudly (count plus the first few underlying errors), and `zero-perf-bench turn` exits 1 when every attempted task errored, since such a report contains no valid samples. Partial errors keep exit 0: the surviving samples are real measurements and the summary still shows what died. Schema bump documented alongside v2/v3, and the version pin test updated so the bump stays a conscious decision.

## Scope

Bench tooling only. No product code path changes; `zero` itself is untouched. No performance change expected.

## Verification

- `go test ./internal/perfbench/ ./cmd/zero-perf-bench/` green; `go vet` and `gofmt` clean on touched files
- New tests: `ResolveBinary` returns an absolute path for an explicit relative binary; a fully-errored run reports `tasksErrored`, surfaces the underlying spawn error in the summary, and exits 1; a partially-errored run counts the dead task but keeps exit 0
- Reproduced the original failure end to end before the fix (48/48 tasks errored with `fork/exec ./zero-bench-main.exe: The system cannot find the file specified`, summary clean, exit 0) and confirmed the same invocation now fails loudly

## Summary by CodeRabbit

* **Bug Fixes**
  * The `zero-perf-bench turn` command now returns a non-zero exit status when every attempted task fails before producing valid samples.
  * Turn benchmark summaries now include an errored-task section with relevant warnings when applicable.
  * Explicitly provided relative binary paths are now resolved more reliably to absolute paths.

* **Compatibility**
  * Published turn benchmark result data now uses schema version 4, which includes errored-task tracking.


Vasanthdev2004:
Good catch, the wording was claiming the opposite of the actual accounting: errored oracle tasks stay in their tier denominators as failures. The summary line now says exactly that.

anandh8x (changes requested):
Reviewed current head ca779cb. The binary-path fix and all-error exit behavior are directionally sound, and focused race tests, full vet, mandated lint, govulncheck, formatting, diff hygiene, and CI pass. One reporting-semantics blocker remains in the new schema-v4 messaging. TasksErrored increments whenever every iteration returns TurnTaskOutcome.Err, but Err is not limited to pre-run spawn/crash failures: NewTurnExecRunner also sets it after a successful agent run and parsed trace when stampOracleAndAnswer fails. Therefore the schema comment, summary, and CLI error saying these tasks never produced an agent run are not true for all values counted. Describe the metric as tasks whose iterations produced no accepted benchmark sample (including harness/run errors), or track pre-run failures separately. Also, the summary says all errored tasks count as failures in tier pass rates, but latency-only tasks are intentionally excluded from every pass-rate denominator; the new mixed nav/longproc test demonstrates this case. Narrow that statement to errored oracle tasks and add assertions covering both semantics. Once the schema docs, summary/exit wording, and tests agree with the actual accounting, this is ready for approval.

Vasanthdev2004:
All three landed in 30035423 and 9b99cd7e. The schema doc, the increment-site comment, the summary line, and the CLI failure message now describe the metric as tasks whose iterations produced no accepted benchmark sample, covering spawn failures and post-run harness errors like oracle stamping alike. The summary also states the tier split correctly now: errored oracle tasks count as failures in their tier's rate, latency-only tasks stay out of pass rates as always. The mixed nav/longproc test pins both semantics (correctness 0/1, latencyOnlyTasks 1, rate 0). Also untracked a generated baseline.json my earlier commit swept in by accident and added a gitignore rule so a generated report can never land in the tree again, per the reports README contract.

gnanam1990:
**Verdict: solid, correct, well-scoped — approve after two small touch-ups.** This is a real honesty fix for the Phase-0 harness: it closes the exact "a run that measured *nothing* printed a clean 0% and exited 0" trap that #712 exists to prevent, one layer down. The reasoning in the code comments is excellent.

### Verified correct
- **Error surfacing actually works.** The summary filters warnings on `Metric == "run"`, and the error-path append uses exactly `Metric: "run"` (`turn_bench.go:278`) — so the "first few underlying errors" really do appear. The loud-failure claim holds.
- **Accounting is right.** An all-errored oracle task lands in the `default`/`buildOnly` branch with `passedForTask=false` → it stays in the `TasksVerified`/`BuildCheckedTasks` denominator but not in `*Passed`, so it counts as a *failure*, not a silently-dropped sample. Latency-only errored tasks stay out of all pass rates. Summary wording matches the code.
- **`erroredIterations == iterations` is the correct gate** — a task only counts as errored when *every* iteration produced no sample; a ran-but-failed-oracle iteration is still a real measurement, consistent with the anti-averaging design.
- **Exit-code seam is clean:** JSON still goes to stdout, the failure line to stderr, exit 1 only when `TasksAttempted > 0 && TasksErrored == TasksAttempted`; partial errors keep exit 0.

### Findings

**1. [Low — same bug class] The `exec.LookPath` branch can still return a relative path.** `ResolveBinary` now absolutizes the *explicit* branch, and the PR correctly notes the repo-root fallback (`filepath.Join(cwd, …)`) is already absolute. But the two `LookPath("zero"/"zero.exe")` branches (`turn_bench.go:797-800`) return the lookup result verbatim, and `exec.LookPath` returns a **relative** path when `PATH` contains a relative entry (e.g. `.`). That reintroduces the exact `fork/exec ./zero: file does not exist` failure this PR exists to kill. Cleanest fix: absolutize at a single exit point rather than per-branch (funnel every successful return through one `filepath.Abs`). Edge case (relative `PATH` entries are uncommon), so not a blocker — but it's inconsistent with the PR's own thesis to leave one path relative.

**2. [Nit — likely already addressed] CodeRabbit's open comment.** It asks the `ERRORED:` line to state errored oracle tasks "remain included as failures in pass-rate denominators." The current message — *"errored oracle tasks count as failures in their tier's pass rate"* — already says exactly that and matches the code. Either adopt CodeRabbit's phrasing to clear the bot, or reply that it's accurate as-is. Not a real defect.

*Reviewed statically (did not compile in-tree). Imports (`filepath`, `errors`) are already present so it builds; the PR reports `go test ./internal/perfbench/ ./cmd/zero-perf-bench/` green with new tests covering all three paths.*

gnanam1990 (approved):
**APPROVE** — non-blocking. The fix is correct and well-scoped: it closes the "measured nothing but printed a clean 0% / exit 0" trap at the harness layer, error surfacing and tier accounting check out, and the exit-code seam is clean. Bench tooling only; no product code touched.

Two follow-ups, neither blocking merge:
1. **[Low]** `exec.LookPath` branches in `ResolveBinary` can still return a relative path when `PATH` has a relative entry — same bug class as the explicit branch this PR absolutizes. Suggest funneling all returns through one `filepath.Abs`. Fine as a fast-follow.
2. **[Nit]** CodeRabbit's `ERRORED:` wording comment — the current message already matches the code; adopt its phrasing or reply to close the thread.

Detailed review: #730 (comment above). Merge is kevin's call per the program's gate.

anandh8x (approved):
Approved at 9b99cd7. The reporting-semantics blockers are resolved: tasksErrored now means no accepted benchmark sample, post-run harness errors are represented honestly, and oracle versus latency-only tier accounting is described correctly. The binary-path fix and generated-report ignore rule are sound. Focused race tests, vet, formatting, govulncheck, diff hygiene, and all CI checks pass. The repository-wide lint command reports existing findings outside this PRs changed files and is not attributable to this change.

PR #734, opened 2026-07-18
docs: refresh repository agent guidelines

anandh8x:
## Summary

- require all contributors and coding agents to read CONTRIBUTING.md before opening a pull request
- align repository guidance with the current Makefile, CI build and smoke flow, Go toolchain, security checks, and advisory lint policy
- document focused PR scope, generated-report handling, race testing, and preservation of unrelated working-tree changes
- remove obsolete AGENTS.md frontmatter whose globs and alwaysApply fields are not interpreted by Zero

## Scope

Documentation only. This pull request changes AGENTS.md and no other file.

## Verification

- go fmt ./...
- go vet ./...
- go test ./...
- pinned golangci-lint command
- govulncheck: no vulnerabilities found
- git diff --check
- verified origin/main...HEAD contains only AGENTS.md

## Summary by CodeRabbit

* **Documentation**
  * Rewrote repository contribution and pull request guidelines with clearer expectations for review readiness, commit scope, and describing changes with supporting verification.
  * Expanded development conventions covering formatting, testing (including race detection), build/release workflow guidance, and safer contribution practices.
  * Added a stricter, ordered validation checklist covering tests, builds/smoke checks, linting guidance, security scanning, and diff hygiene.
  * Clarified which checks are required versus advisory and how to handle failures.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,7 +44,8 @@ guide to extending Zero with specialists, hooks, plugins, MCP, and skills, see
 Run validation from the repository root before committing, opening a pull
 request, or completing an implementation task:
 
-1. **Formatting**: `go fmt ./...` (or `make fmt`).
+1. **Formatting check**: `make fmt-check`. If it fails, format with
+   `go fmt ./...` (or `make fmt`) and run the check again.
 2. **Vet**: `go vet ./...` (or `make vet`).
 3. **Tests**: `go test ./...`. Use `make test` for the full race-enabled suite,
    or run focused tests with `-race`, when concurrency is affected.
@@ -54,7 +55,8 @@ request, or completing an implementation task:
    `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --enable-only unused,ineffassign,staticcheck ./...`.
 7. **Security**:
    `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...`.
-8. **Diff hygiene**: `git diff --check`.
+8. **Diff hygiene**: `git diff HEAD --check` (covers staged and unstaged
+   tracked changes).
 
 `make lint` currently runs the formatting check and `go vet`; it does **not**
 run golangci-lint. The pinned golangci-lint job is advisory in CI while the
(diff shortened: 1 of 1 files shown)


gnanam1990 (approved):
**APPROVE** — accurate and well-structured. Verified every factual claim against `Makefile` on main:

- `make fmt-check`, `make fmt`, `make vet`, `make test` all exist ✅
- "`make lint` runs fmt-check + vet, not golangci-lint" — matches exactly (`lint: fmt-check vet`) ✅
- `internal/perfbench/reports/` holds only `.gitkeep` + `README.md` (0 tracked JSON), so the "don't commit reports" guidance is accurate and forward-looking ✅
- Removing the obsolete `globs`/`alwaysApply` frontmatter is correct — Zero doesn't interpret those ✅

CodeRabbit's two findings (use `make fmt-check`; use `git diff HEAD --check`) are **already reflected in the current head** — its suggestions diff against an older commit, so the CHANGES_REQUESTED is stale and safe to dismiss.

Optional non-blocking nit: the "don't commit reports" rule isn't enforced by `.gitignore` — an `internal/perfbench/reports/*.json` ignore entry would give it teeth. Fine as a fast-follow. Merge is kevin's call per the program gate.

gnanam1990 (approved):
**Re-approve** (prior approval was auto-dismissed by `5fb01da`). No regressions — the new commit just makes the formatting step non-mutating (`make fmt-check` as the check; `go fmt`/`make fmt` as remediation) and uses `git diff HEAD --check`. All factual claims still verified against the Makefile; both CodeRabbit findings now fully resolved. LGTM.

Vasanthdev2004 (approved):
Approving. I checked the guidance claims against the repo rather than taking them on faith, and they hold: make lint really is fmt-check plus vet only (so calling the pinned golangci run advisory and separate is right), make test really is the race-enabled suite, the zero-release build and smoke commands exist as documented, and the generated-reports rule matches the contract in internal/perfbench/reports/README.md. The frontmatter removal is better than cosmetic: nothing in the codebase parses globs or alwaysApply, and AGENTS.md is injected raw into the project context, so that block was pure prompt noise on every run. The scope, working-tree preservation, and CONTRIBUTING-first rules also line up with where the repo has been heading this week.

One nit, take it or leave it: "Never edit files under third_party/" survived the refresh, but there is no third_party directory in the repo. Since this PR exists to remove guidance that no longer matches reality, that line is a candidate for the same treatment. Not blocking on it.

PR #732, opened 2026-07-18
docs(community): wire Discussions as the front door for questions and ideas

Vasanthdev2004:
Sets up the community entry points so Discussions actually gets used instead of the issue tracker filling with questions.

- New SUPPORT.md: where to ask, and exactly what to include for a fast answer (zero --version, OS and install method, provider/model, zero doctor output). GitHub surfaces this file automatically when people open issues.
- Issue chooser (config.yml) now routes questions to Discussions Q&A and feature ideas to the Ideas category, alongside the existing private security link. Blank issues stay disabled, so the tracker stays for actionable bugs.
- README gains a Community section with a what-goes-where table for the four categories.
- CONTRIBUTING points questions at Discussions and frames Ideas as the natural path toward an issue-approved issue, matching the controlled-development policy.

Docs only, no code. No performance change expected.

## Summary by CodeRabbit

* **Documentation**
  * Added `SUPPORT.md` with guidance for getting help, reporting bugs, proposing feature ideas, and handling security issues.
  * Updated `README.md` “Community” section with clear GitHub Discussions category guidance (Q&A, Ideas, Show and tell, Announcements) and posting tips.
  * Revised `CONTRIBUTING.md` to route questions and setup help to Discussions and to start feature ideas in the Ideas category before creating trackable work.
  * Updated issue/feature request templates to encourage prior discussion and to capture a related Discussions thread link.

PR #736, opened 2026-07-18
perf(agent): posture-escalation signals and controller (PR10a)

Vasanthdev2004:
## Summary

First slice of **PR10 of the Zero Agent Performance Program** (execution profiles): the run-policy signal plumbing and the posture-escalation controller, with no user surface and no behavior change. The profile catalog, the `--exec-profile` selector, and the baseline-tuned knob values follow in the next slice once the Phase 0 numbers are settled; this slice is deliberately inert so it can land independently of the M1 discussion.

### Added

- **`ToolResult.Risk`**: executed tool results now carry the sandbox risk classification the permission path already computes. Sandboxed calls reuse the preflight decision's risk (no second evaluation); unsandboxed calls run the same pure classifier the permission event path uses; unknown-tool and denied results keep the zero value. Pure observation: permissions, sandboxing, and prompting are untouched.
- **SelfCorrect outcome captured**: the post-edit verification outcome was previously discarded (`feedback, _ :=`); the loop now observes it. Feedback handling is unchanged.
- **`ProfilePolicy` / `PostureEscalation`** types on `agent.Options`: a one-shot, in-run escalation to stricter knob values (turn ceiling, reasoning effort, completion gate), with triggers for tool-failure streaks, uncertain completion evaluations, self-correct failures, and executed mutations at or above a risk threshold. Targets are the values a profile displaced at run start, so escalation can never introduce a value that was not already valid for the run.
- **`profileController`**: observes the signals at the loop's existing seams (the repeated-failure guard's own streak, the completion gate's uncertain branch, the self-correct call, executed results) and applies at most one escalation per run at the end-of-turn tail. It mutates only per-turn-read policy values; it never injects messages, never touches the model or session, and is fully independent of the model-initiated `escalate_model` path.
- **`posture_escalations` trace counter** (optional event key), deliberately distinct from `model_switches`.

## No behavior change

`Options.Profile` is nil for every existing caller, which makes every controller method a no-op; the loop is byte-identical for unprofiled runs (pinned by `TestPostureEscalationAbsentWithoutProfile` and the nil-policy test, and the full agent suite passes unchanged). The risk stamp adds a field to executed results that nothing consumes yet. No performance change expected.

## Verification

- `go build ./...`, `go vet`, `gofmt` clean on touched files
- Full `internal/agent` and `internal/trace` suites green
- 11 new tests: nil-policy no-op, each trigger fires (failure streak, risky mutation with denied-result exclusion, self-correct failure, uncertain completion), one-shot invariant, the end-to-end act path (a 2-turn-ceiling run escalates on a failure streak, finishes on turn 3, carries the target effort on the post-escalation request, and emits the counter), the unprofiled twin of that run staying cut off at its ceiling, risk classification on the unsandboxed path, zero risk on not-executed results, and the optional-event-key registration

## Dependencies and follow-ups

Builds on the completion policy (#719) and the turn-session seam (#720). Follow-ups per the program plan: the `internal/execprofile` catalog with `--exec-profile` and baseline-tuned values (next slice, with A/B evidence for the Fast profile gate), then the TUI `/profile` command.

## Summary by CodeRabbit

* **New Features**
  * Added configurable execution profiles that can trigger one-shot “posture escalations” during a run.
  * Escalations can raise the turn limit, carry forward reasoning effort, and require a completion signal when uncertain.
  * Added sandbox risk classification to executed tool results (when available).
  * Added optional tracing for posture escalation counters.
* **Bug Fixes**
  * Ensured denied/canceled or unexecuted tools keep an empty/zero risk classification.
* **Tests**
  * Expanded coverage for escalation triggers, one-shot behavior, mid-run limit changes, risk classification, and trace event keys.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -189,6 +189,26 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 	// existing caller) makes every observe/decide call a no-op, keeping the
 	// loop byte-identical for unprofiled runs.
 	posture := newProfileController(options.Profile)
+	// applyPosture applies at most one posture escalation when an armed trigger
+	// has fired: raise the hoisted turn ceiling, restore effort and the
+	// completion gate, stamp the counter. Shared by the end-of-turn tail and
+	// the completion-gate uncertain path (which continues the loop before the
+	// tail runs). No-op forever after the first escalation and for unprofiled
+	// runs.
+	applyPosture := func() {
+		if target, fired := posture.maybeEscalate(); fired {
+			if target.MaxTurns > maxTurns {
+				maxTurns = target.MaxTurns
+			}
+			if target.ReasoningEffort != "" {
+				options.ReasoningEffort = target.ReasoningEffort
+			}
+			if target.RestoreCompletionGate {
+				options.RequireCompletionSignal = true
+			}
+			options.Trace.Counter(trace.CounterPostureEscalations, 1)
+		}
+	}
 
 	// Background post-edit diagnostics: files changed by mutating tools are
 	// checked off the tool-call critical path and any errors are appended as a
@@ -558,6 +578,11 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 					result.Messages = copyMessages(messages)
 					return result, nil
 				case CompletionUncertain:
+					// Observe the uncertainty signal and apply any escalation NOW:
+					// this branch continues the turn loop before the end-of-turn
+					// act point, so deferring would skip escalation entirely.
+					posture.observeUncertain()
+					applyPosture()
 					switch evaluation.Action {
 					case completionActionContinue:
 						options.Trace.Counter(trace.CounterCompletionNudges, 1)
@@ -830,18 +855,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		// only per-turn-read policy (the hoisted turn ceiling, reasoning effort,
 		// completion gate); never messages, model, or session. No-op forever
 		// after the first escalation and for every unprofiled run.
-		if target, fired := posture.maybeEscalate(); fired {
-			if target.MaxTurns > maxTurns {
-				maxTurns = target.MaxTurns
-			}
-			if target.ReasoningEffort != "" {
-				options.ReasoningEffort = target.ReasoningEffort
-			}
-			if target.RestoreCompletionGate {
-				options.RequireCompletionSignal = true
-			}
-			options.Trace.Counter(trace.CounterPostureEscalations, 1)
-		}
+		applyPosture()
 	}
 
 	if ctx.Err() != nil {
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1371,7 +1371,9 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 	executedRisk := sandbox.Risk{}
 	if preflightDecision != nil {
 		executedRisk = preflightDecision.Risk
-	} else {
+	} else if toolFound {
+		// Unknown-tool results fall through here with a nil tool; they carry a
+		// denial and keep the zero risk value.
 		executedRisk = sandbox.Classify(sandboxRequest(call.Name, tool, args, permissionGranted, permissionMode, options))
 	}
 
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -185,6 +185,10 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 
 	guards := newGuardState()
 	compactor := newCompactionState(options)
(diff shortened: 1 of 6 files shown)


anandh8x (changes requested):
Reviewed current head fb0b0e2. Focused agent/trace race tests, vet, formatting, and diff hygiene pass. Two blockers remain. 1. OnCompletionUncertain is not wired into the loop at all: observeUncertain appears only in profile_controller.go and its unit tests. The CompletionUncertain branch in loop.go immediately continues, so merely adding the observation there would still skip the sole maybeEscalate act point at the end-of-turn tail. Refactor/apply escalation before that continue (or through a shared act helper) and add a loop-level regression proving an uncertain completion changes the next request/ceiling as configured and increments posture_escalations. 2. riskRank claims unknown levels rank lowest so a malformed threshold cannot trip spuriously, but observeToolOutcome compares resultRank >= thresholdRank. A non-empty invalid OnRiskyMutation therefore ranks 0 and every successful classified result ranks >= 0, firing escalation. Disable/reject unknown thresholds before comparison and add an invalid-threshold regression. The nil-profile and other trigger paths otherwise look sound.

Vasanthdev2004:
Both blockers fixed in e8afb7f3, and you were right on both counts.

The uncertain path was worse than unwired: my observation call had silently failed to apply during editing, and even if it had landed, the branch continues the turn loop before the end-of-turn act point, so escalation would never have applied from there. The fix extracts the act logic into a shared applyPosture helper used at the tail and invoked immediately after observeUncertain inside the uncertain branch, before the continue. The new loop-level regression TestPostureEscalationOnUncertainCompletion drives a cue turn under the completion gate and asserts the very next request carries the target effort and posture_escalations increments.

The risk-rank inversion is fixed the way you suggested: the threshold is validated first (an unrecognized level ranks 0 and disables the signal instead of matching everything), with TestProfileControllerIgnoresInvalidRiskThreshold pinning a typo threshold against a critical result. While in there I also took the bot's related point: the executed check is now DenialReason == DenialNone rather than StatusOK, so a partial failure that actually ran the mutation counts, with a test case for that too. The doc comments now describe the real behavior.

Also folded in the two smaller bot asks: the unprofiled-run test now requires the max-turns answer outright, and the optional-event-key test moved to the trace package where it belongs.

anandh8x (approved):
Re-reviewed head e8afb7f. Both prior blockers are resolved: CompletionUncertain now observes and immediately applies posture through the shared helper before the loop continues, with a loop-level regression proving the next request changes and posture_escalations increments; invalid OnRiskyMutation thresholds now disable the signal instead of matching every classified result. The follow-up also correctly counts executed partial failures while excluding denied calls, strengthens the nil-profile regression, and colocates the trace-key test. PR-specific race tests, trace tests, go vet, and git diff --check pass. Approving.

gnanam1990 (approved):
**APPROVE (code)** — clean, correct, genuinely inert slice. **One pre-merge requirement: rebase off the stacked PRs first.**

Verified against the actual code, not just the description:
- **Inert for existing callers**: `Options.Profile` is nil everywhere → nil-policy controller, every `observe*` early-returns, `maybeEscalate` returns false, and the counter fires only inside the `if fired` block. Transcript/model behavior is byte-identical. ✅
- **`sandbox.Classify` is pure** (`classifyWithScope(request, nil)`, no I/O/prompting); sandboxed calls reuse `preflightDecision.Risk` (no double-eval). ✅
- **`maxTurns` is read dynamically** in the loop condition and `applyPosture` only ever *raises* it — escalation can extend a run, never shorten one. ✅
- Defensive controller: one-shot latch; `riskRank` ranks unknown levels 0 so a typo'd threshold **disables** the signal instead of matching everything; targets are the already-valid displaced values. ✅
- Counter registered in `OptionalEventKeys()`; the dual `applyPosture` sites are safe via the one-shot latch. 11 new tests incl. the no-op twin and the end-to-end act path. ✅

**1. [Pre-merge, process] Rebase off the stacked #730/#732/#734.** The branch carries their commits (`dbd9443` #730, `0fae754` #734, `2cc6f43` #732), so the diff shows 17 files instead of the real 6, and merging as-is would pull those PRs' contents into `main` and duplicate/conflict when they land on their own. Rebase onto clean `main` (or merge only after they do) so #736 contains just the 6 posture commits.

**2. [Minor] Risk stamp runs unconditionally.** `executeToolCall` computes `executedRisk` regardless of `Profile`, so unprofiled runs now run `sandbox.Classify` on each executed *unsandboxed* call. Byte-identical output and Classify is cheap, so 'no perf change' holds in practice — but it's not literally zero new work. Optional: gate the unsandboxed-classify branch on `options.Profile != nil` for truly cost-free unprofiled runs (tradeoff: field unavailable to future non-profile consumers).

Nice work — the slice does exactly what it claims. Merge is kevin's call per the program gate, after the rebase.

PR #737, opened 2026-07-18
fix(perfbench): keep the stamped answer file out of negative oracle greps

Vasanthdev2004:
## Summary

Found while investigating the Phase 0 baseline's 0/24 mutating-class result: four verification oracles grep the whole fixture directory for the ABSENCE of a string, and the harness stamps the agent's final answer into that same directory (`.zero-answer.txt`) before verification runs. So a run that made a perfectly correct edit fails its oracle whenever the final answer truthfully mentions the removed text, e.g. "I changed 'MaxRetries = 3' to 'RetryLimit = 3'".

Reproduced with a hand-made correct edit: the oracle passes with a clean answer file and fails the moment the answer quotes the old constant. The fix scopes the four negative greps with `--exclude=.zero-answer.txt` (edit-01, edit-02, refactor-03, refactor-05; refactor-02's negative grep is file-scoped and unaffected). The previously-failing variant passes with the fixed commands.

For honesty about the baseline numbers: this bug would only have flipped correct runs, and a live probe showed the model in that capture was not producing applied edits in the first place (16 turns, 18 tool calls, zero file mutation, final answer describing an intended change). So the fix corrects the oracle for future captures rather than retroactively explaining the 0/24.

## Scope

Manifest only, 4 lines. No harness code, no product code. No performance change expected.

## Verification

- Hand-made correct edit + answer quoting the old text: fails against the old command, passes against the new one
- `go test ./internal/perfbench/` green (manifest loads and validates)
- Diff is exactly the four `--exclude=` insertions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated verification by excluding baseline answer files from recursive search checks.
  * Prevented unrelated file contents from influencing task validation results.

### DIFF
--- a/internal/perfbench/manifests/baseline.json
+++ b/internal/perfbench/manifests/baseline.json
@@ -133,7 +133,7 @@
       "verificationCommand": [
         "bash",
         "-c",
-        "go test ./... && ! grep -RIn 'MaxRetries =' ."
+        "go test ./... && ! grep --exclude=.zero-answer.txt -RIn 'MaxRetries =' ."
       ],
       "oracleTest": "package edit\n\nvar _ = RetryLimit\n"
     },
@@ -146,7 +146,7 @@
       "verificationCommand": [
         "bash",
         "-c",
-        "go test ./... && ! grep -R receieves . && grep -R receives ."
+        "go test ./... && ! grep --exclude=.zero-answer.txt -R receieves . && grep -R receives ."
       ]
     },
     {
@@ -400,7 +400,7 @@
       "verificationCommand": [
         "bash",
         "-c",
-        "go test ./... && ! grep -R 'package refactor' . && grep -R 'package zeroapp' ."
+        "go test ./... && ! grep --exclude=.zero-answer.txt -R 'package refactor' . && grep -R 'package zeroapp' ."
       ],
       "oracleTest": "package zeroapp\n\nvar _ = Config{}\n"
     },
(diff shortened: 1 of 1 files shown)


gnanam1990 (dismissed):

anandh8x (changes requested):
Reviewed current head 0310f8bc. The four negative greps now correctly exclude .zero-answer.txt, and internal/perfbench tests plus diff hygiene pass. One related correctness blocker remains in edit-02: the companion positive check still searches recursively for receives, so the stamped final answer can satisfy the oracle. I reproduced this by deleting the typo-bearing comment entirely and stamping an answer containing receives; the new verification command exited 0 even though the requested correction was absent from source. Please exclude .zero-answer.txt from that positive grep as well, and ideally from every source-validation recursive grep including refactor-03 for consistency. Then verify the delete-comment case fails while genuinely corrected source passes.

Vasanthdev2004:
Good catch, the positive legs had the same hole mirrored and I only plugged the negative side. Fixed in c5fe4e95: edit-02's `grep -R receives` and refactor-03's `grep -R 'package zeroapp'` now carry the same `--exclude=.zero-answer.txt`.

Reran your repro both ways for edit-02: comment deleted plus an answer that says "receives" now fails the oracle, and a genuinely corrected source with the same answer still passes. Same pair for refactor-03: a claim-only rename fails (and the positive grep leg alone no longer matches the stamped answer), a real rename passes the full oracle including the stamped oracle_test.go.

Also swept the manifest for any remaining recursive grep without the exclude: none left. edit-01 and refactor-05 only have negative recursive legs (already excluded), and every other positive check is file scoped, so the stamped answer can't reach them.

anandh8x (approved):
Re-reviewed head c5fe4e95. The prior blocker is resolved: edit-02 and refactor-03 now exclude .zero-answer.txt from their positive recursive greps as well. I reran the edit-02 regression: deleting the comment while mentioning receives only in the stamped answer now fails, while a genuinely corrected source with the same answer passes. Every recursive grep in the manifest now carries the exclusion; internal/perfbench tests and diff hygiene pass. Approving.

gnanam1990 (approved):

Issue #729, opened 2026-07-18
security(oauth): token POST credentials can leak through redirects


PierrunoYT:
## Summary

OAuth token requests validate only the initial token endpoint. The supplied or default `http.Client` can follow HTTP 307/308 redirects while replaying the POST body to another origin or an HTTPS-to-HTTP destination.

The replayed form may contain authorization codes, PKCE verifiers, refresh tokens, client secrets, and client IDs.

## Affected code

- `internal/oauth/flow.go:113-151` builds the sensitive token form.
- `internal/oauth/flow.go:154-180` validates only the initial URL and calls `client.Do` without a token-specific redirect policy.
- `internal/mcp/oauth.go:224-247` uses the same implementation for MCP OAuth login and refresh.
- `internal/oauth/flow_test.go:185-190` covers an initially insecure endpoint but not redirected endpoints.

## Reproduction outline

1. Configure or discover an HTTPS token endpoint.
2. Have that endpoint return `307 Temporary Redirect` or `308 Permanent Redirect` to a different origin or an HTTP URL.
3. Perform authorization-code exchange or token refresh.
4. Observe that Go can replay the POST and form body at the redirect target.

## Impact

Authorization credentials and long-lived refresh tokens can be disclosed cross-origin or over plaintext despite the initial endpoint passing HTTPS validation.

## Suggested fix

Clone the selected HTTP client and install a token-specific `CheckRedirect` policy. The safest default is to reject all token-endpoint redirects.

Add 307/308 tests for both code exchange and refresh and assert that the redirect target receives no request body or credential-bearing request.

shriza1991:
Hi, I'd like to work on this, can you please assign this to me if its available?

Vasanthdev2004:
assigned @shriza1991

PR #738, opened 2026-07-18
oauth: reject redirects for token endpoint requests

shriza1991:
## Summary

This PR hardens OAuth token requests by rejecting HTTP redirects.

### Changes
- Clone the selected HTTP client before issuing token requests.
- Install a token-specific `CheckRedirect` policy to prevent following redirects.
- Use the cloned client when sending token requests.

### Tests
- Added a regression test for 307 redirects during authorization code exchange.
- Added a regression test for 308 redirects during token refresh.
- Verified that redirected endpoints do not receive credential-bearing requests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth token security by preventing automatic redirects from token endpoints.
  * Token exchanges and refreshes now reject redirect responses without sending credentials to the redirected destination.

### DIFF
--- a/internal/oauth/flow.go
+++ b/internal/oauth/flow.go
@@ -163,6 +163,9 @@ func PostToken(ctx context.Context, client *http.Client, tokenEndpoint string, f
 	if client == nil {
 		client = http.DefaultClient
 	}
+
+	tokenClient := *client
+	tokenClient.CheckRedirect = checkTokenRedirect
 	if now == nil {
 		now = time.Now
 	}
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
Thanks for tackling this, @shriza1991 - the redirect-replay risk on OAuth token requests was a real and worthwhile thing to fix. Since you opened this, #741 landed (merged as 974fc036) and closed the same issue #729, covering the token exchange/refresh path plus the device-authorization and poll flows in one pass, so it's now fully handled on main (and this branch has gone into conflict). Closing as already-resolved, but the instinct and the regression-test approach were spot on. Please keep the contributions coming - the surrounding endpoint-validation work is a good next area. Really appreciate you taking the time on a security fix.

PR #739, opened 2026-07-18
fix(oauth): validate discovered endpoints before merge/use (#511)

### LINKED ISSUE #511: oauth: validate discovered OAuth endpoints before login
## Version / branch / commit
main at commit `3c81fea`

## OS and environment

## Summary
Configured OAuth endpoints are rejected when they use a non-HTTPS, non-loopback URL, but endpoints learned from OAuth/OIDC discovery are merged without the same validation. That lets discovery metadata downgrade the authorization, token/device, or MCP registration endpoint safety after the initial provider config has already passed validation.

This affects both the provider OAuth flow and the MCP OAuth flow because MCP delegates discovery to `internal/oauth` and then uses the returned metadata.

## Steps to reproduce
1. Configure an OAuth provider or MCP OAuth server using an issuer/discovery URL.
2. Serve discovery metadata from that issuer with an insecure discovered endpoint, for example:

```json
{
  "issuer": "http://127.0.0.1:12345",
  "authorization_endpoint": "http://evil.example/authorize",
  "token_endpoint": "http://127.0.0.1:12345/token"
}
```

3. Start the loopback login flow.
4. The login flow builds and opens an authorization URL beginning with `http://evil.example/authorize?`.

I verified this locally with temporary tests against both `internal/oauth.Manager.Login` and the MCP OAuth resolver/authorization URL path. The fake browser opener received a URL starting with `http://evil.example/authorize?`.

## Expected behavior
Discovered OAuth endpoints should be validated before being merged or used. A non-HTTPS, non-loopback `authorization_endpoint`, `token_endpoint`, `device_authorization_endpoint`, or MCP `registration_endpoint` should be rejected with the same endpoint-safety rule used for configured endpoints.

## Actual behavior
Configured endpoints pass through `validateTokenEndpoint`, but discovery metadata is merged without revalidation:

- `internal/oauth/providers.go` validates configured endpoints.
- `internal/oauth/manager.go` merges discovered provider endpoints into the config.
- `internal/mcp/oauth.go` merges discovered MCP OAuth metadata.
- `internal/oauth/flow.go` only parses the authorization endpoint before building the browser URL.

As a result, discovery metadata can silently downgrade the login flow to an insecure authorization URL.

## Notes
I did not find an existing issue or open PR covering this exact validation gap.

gnanam1990:
## Summary

Closes #511.

Discovery-learned OAuth endpoints were merged and used **without** the https/loopback endpoint-safety rule that configured endpoints already pass. A malicious issuer's discovery document could therefore downgrade the authorization, token, device, or MCP registration endpoint *after* the provider config had already validated — silently redirecting the login to an attacker-controlled `http://` origin. This affects both the provider OAuth flow and the MCP OAuth flow (MCP delegates discovery to `internal/oauth`).

## Root cause

- `internal/oauth/providers.go` validates *configured* endpoints via `validateTokenEndpoint`, but `internal/oauth/manager.go` `resolveEndpoints` merged *discovered* endpoints with no revalidation.
- `internal/mcp/oauth.go` `resolveAuthorizationServer` likewise merged discovered MCP metadata without validation.
- `internal/oauth/flow.go` `BuildAuthorizationURL` parsed the authorization endpoint but never checked its scheme, so a downgraded endpoint reached the browser.

## Fix — defense in depth (3 layers)

1. **One shared rule.** New exported `oauth.ValidateEndpointURL` (https, or http on a loopback host); `validateTokenEndpoint` now delegates to it so configured and discovered endpoints share a single rule.
2. **Backstop at the choke point.** `BuildAuthorizationURL` validates the authorization endpoint. Both the provider loopback flow and the MCP flow build their browser URL through this function, so a downgraded endpoint can never open in the browser even if a merge site is missed.
3. **Validate before merge, fail closed.**
   - Provider: `resolveEndpoints` validates each discovered authorization / token / device endpoint before merging.
   - MCP: new `validateResolvedEndpoints` applied in **both** return branches of `resolveAuthorizationServer` (authorization / token / registration, configured and discovered alike).

## Scope

`internal/oauth` + `internal/mcp` only. No config schema change. No behavior change for valid endpoints — https and http-loopback logins are unaffected. **No performance change expected.**

## Testing

- **3 regression tests** — provider `resolveEndpoints`, MCP `resolveAuthorizationServer`, and the `BuildAuthorizationURL` backstop. Each was confirmed to **fail without the fix** (reverted the source, kept the tests) and pass with it; the pre-existing configured-endpoint test is unchanged.
- `go test -race ./internal/oauth/... ./internal/mcp/...` green. `go build ./...`, `go vet ./...`, `gofmt`, `govulncheck`, `git diff HEAD --check` all clean.
- **Manual end-to-end through the real CLI.** With a discovery document serving `authorization_endpoint: http://evil.example/authorize`, `zero auth login <provider>` now exits 1 with `refusing to send credential to a non-https token endpoint: http://evil.example` and **never opens the browser** — whereas unpatched `main` builds and opens that attacker URL. A legitimate https discovery login still succeeds (happy path intact).

## Follow-up

The paired hardening for #729 (token-endpoint redirect replay in `PostToken`) is a separate PR on the same files.

## Summary by CodeRabbit

* **Security**
  * OAuth authorization, token, registration, and device-authorization endpoints now require HTTPS (HTTP allowed only for loopback).
  * Insecure endpoints from configuration or discovery are rejected before proceeding with OAuth flows.
  * Authorization URL building now validates the endpoint and won’t generate downgraded HTTP links.

* **Bug Fixes**
  * Discovery/endpoint resolution now validates discovered endpoint URLs and fails safely when invalid.

* **Tests**
  * Added coverage to ensure insecure discovered/configured endpoints are rejected.

### DIFF
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -133,6 +133,26 @@ func discoverAuthorizationServer(ctx context.Context, client *http.Client, baseU
 	}, nil
 }
 
+// validateResolvedEndpoints applies the shared https/loopback endpoint rule to
+// every non-empty endpoint in the resolved metadata — configured or discovered
+// alike — so discovery metadata can never downgrade an MCP login to an insecure
+// or attacker-controlled authorization, token, or registration endpoint.
+func validateResolvedEndpoints(metadata authServerMetadata) error {
+	for _, ep := range []struct{ kind, url string }{
+		{"authorization", metadata.AuthorizationEndpoint},
+		{"token", metadata.TokenEndpoint},
+		{"registration", metadata.RegistrationEndpoint},
+	} {
+		if strings.TrimSpace(ep.url) == "" {
+			continue
+		}
+		if err := oauth.ValidateEndpointURL(ep.url); err != nil {
+			return fmt.Errorf("mcp oauth: %s endpoint: %w", ep.kind, err)
+		}
+	}
+	return nil
+}
+
 // resolveAuthorizationServer discovers metadata and applies explicit config
 // overrides. Configured endpoints take precedence over discovered ones, and act
 // as a fallback when discovery fails or omits a value.
@@ -147,6 +167,9 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 			AuthorizationEndpoint: strings.TrimSpace(cfg.AuthorizationEndpoint),
 			TokenEndpoint:         strings.TrimSpace(cfg.TokenEndpoint),
 			RegistrationEndpoint:  strings.TrimSpace(cfg.RegistrationEndpoint),
+		}
+		if err := validateResolvedEndpoints(metadata); err != nil {
+			return authServerMetadata{}, err
 		}
 		return metadata, nil
 	}
@@ -178,6 +201,9 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 	}
 	if strings.TrimSpace(metadata.TokenEndpoint) == "" {
 		return authServerMetadata{}, errors.New("no token endpoint discovered or configured")
+	}
+	if err := validateResolvedEndpoints(metadata); err != nil {
+		return authServerMetadata{}, err
 	}
 	return metadata, nil
 }
--- a/internal/mcp/oauth_test.go
+++ b/internal/mcp/oauth_test.go
@@ -73,6 +73,22 @@ func TestResolveEndpointsFallsBackToConfig(t *testing.T) {
 	}
 }
 
+// Discovered MCP metadata that serves an insecure endpoint must be rejected
+// before use, so discovery cannot downgrade the MCP login to an
+// attacker-controlled origin (issue #511).
+func TestResolveAuthorizationServerRejectsInsecureDiscovered(t *testing.T) {
+	mux := http.NewServeMux()
+	server := httptest.NewServer(mux)
+	defer server.Close()
+	mux.HandleFunc("/.well-known/oauth-authorization-server", func(w http.ResponseWriter, _ *http.Request) {
+		_, _ = w.Write([]byte(`{"issuer":"` + server.URL + `","token_endpoint":"` + server.URL + `/token","authorization_endpoint":"http://evil.example/authorize"}`))
+	})
+	_, err := resolveAuthorizationServer(context.Background(), server.Client(), server.URL, OAuthConfig{})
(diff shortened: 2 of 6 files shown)


Vasanthdev2004 (approved):
Reviewed at d6a3ff62. This is the right shape for the fix: one exported rule, validation at every merge site, and a backstop at the choke point both flows funnel through. Approving.

What I checked:

- Traced every credential-bearing path to a validation: device start and poll (device.go:49/157), token exchange (flow.go:176), the browser URL choke point (flow.go:46), provider config (providers.go:112), both discovered merges (manager.go:226-238), and both MCP branches (mcp/oauth.go:149). I could not find an endpoint that reaches the wire unvalidated; PostToken redirect replay is scoped out to your #729 follow-up, which is fine.
- Threw some hostile URLs at ValidateEndpointURL: http://localhost@evil.com resolves Hostname to evil.com and is rejected, scheme-relative //evil.com/x has no scheme and is rejected, http://127.0.0.1.evil.com fails ParseIP and is rejected, http://[::1]:8080 is correctly allowed. Everything ambiguous fails closed.
- Independently re-verified your fail-without-fix claim: kept the three new tests, reverted flow.go/manager.go/mcp/oauth.go to main, and all three fail; restored, they pass. Race run on both packages is green here too.

One behavior change worth stating plainly for the changelog: the MCP config branch now validates CONFIGURED endpoints as well, so a working (if ill-advised) http non-loopback MCP OAuth setup goes from failing at token time with a confusing error to failing up front with a clear one. That is the right call, just worth remembering if someone reports it as a regression.

Non-blocking nit: the backstop surfaces ErrInsecureTokenEndpoint for an authorization endpoint, so the message says "token endpoint" about a URL that is not one (your own manual-test transcript shows it). Renaming the exported error would break errors.Is callers, so leaving it is fine; a message tweak can ride along some other day.

anandh8x (approved):
Reviewed head d6a3ff62. The fix is sound and defense-in-depth is placed correctly: one shared HTTPS-or-loopback rule, validation before provider/MCP discovery merges, validation of configured MCP endpoints in both resolver branches, and a BuildAuthorizationURL backstop before either flow opens the browser. I traced authorization, token, device-authorization, and registration endpoints to validation before credential-bearing use; spoofed loopback host forms fail through URL Hostname plus net.ParseIP handling. Redirect replay remains correctly scoped to the separate #729 follow-up. OAuth/MCP race tests, vet, and diff hygiene pass. Approving.

PR #740, opened 2026-07-18
perf(agent): execution profiles with one-shot posture escalation (PR10b+PR10c)

Vasanthdev2004:
PR10b + PR10c of the performance program in one PR: user-visible execution profiles on every surface, built on the escalation machinery that landed inert in #736.

## What this adds

- `zero exec --exec-profile fast|balanced|thorough`: a loop-posture bundle (turn budget, reasoning effort, self-correction, escalation triggers). It composes with `--mode`: the mode picks what runs, the profile picks how hard the loop tries. Precedence is explicit flag > mode > profile, enforced by applying the profile last with the same fill-only-if-unset rule modes use. Unknown names are a usage error listing the valid ones; the legacy inert `--profile` is untouched and the two coexist (test extended).
- TUI `/profile` with the same catalog, applying to the next run. Profiles never stack (switching reverts first), and reverts only restore knobs the user has not explicitly touched since.
- `zero-perf-bench turn --exec-profile`: forwarded to every task invocation and stamped into the result (schema v5, additive only), printed in the summary next to `mode:`, so an A/B capture is self-describing. The bench validates and canonicalizes the name before spawning anything; an unknown name previously produced an exit-0 report whose walls were millisecond usage-error exits.
- Trace runs record the selected profile name (`--trace`), balanced included, by design.

## The catalog

- `balanced` is the empty profile. Selecting it changes nothing: options, turn budget, and loop behavior are identical to an unflagged run, asserted by test. Only the opt-in trace/bench artifacts record the name.
- `fast`: 30 turns, low effort, all four escalation triggers armed. When trouble hits (a same-tool failure streak, a second uncertain completion, a failing self-correct cycle, or a critical-risk mutation), a one-shot escalation restores the displaced posture: the pre-profile turn budget and the provider-default effort. The risk trigger is deliberately critical, not high; the sandbox floors every shell command at high before any command analysis, so a high threshold would spend the one-shot on the first `go test` of any real task.
- `thorough`: 160 turns, high effort, full self-correct. Already the maximum posture, so no triggers.

Restoring the displaced effort needed a small addition to the #736 surface: `PostureEscalation.RestoreDefaultEffort`. A profile only ever fills effort that was unset, so the displaced value is the provider default, and a plain empty target already means "leave untouched". The flag is armed only when the profile actually performed the fill; an explicit `-r`, a mode's effort, an explicit `--spec-reasoning-effort` on the draft path, or a TUI `/effort`/Ctrl+T choice made after selecting the profile all keep or regain control (the TUI disarms the restore on the policy clone, the same way `/turns` under a profile pins the budget by disarming the escalation's turn target).

## Tuning status, honestly

The Fast/Thorough numbers are provisional floors from the Phase 0 baseline's read-class evidence and are pinned by test so a retune is a visible diff. The mutating classes produced no successful samples to tune from (see the baseline investigation and #737), so the Fast-vs-Balanced A/B with the pass-rate gate comes with the post-#737 re-capture; the escalation safety net is what makes shipping provisional values safe in the meantime.

## Program gates

- No performance change expected by default: profiles are opt-in, and the unflagged/balanced path is byte-identical (asserted, not claimed).
- Provider behavior, permissions, sandboxing, secret scrubbing, output ceilings, and one-result-per-tool-call are untouched; escalation only restores values the run already had.
- Fully disableable: do not pass the flag.

## Testing

- New suites: `internal/execprofile` (catalog invariants, displaced-values-only escalation), cli (precedence, budget displacement, balanced identity, usage errors, full echo-provider runs asserting the trace stamp), tui (posture apply, pin/disarm, coincidence-safe reverts), perfbench + bench cmd (passthrough, validation, normalization, schema pin).
- `go build ./...`, `go vet ./...`, gofmt clean; agent, cli, execprofile, perfbench, tui, and bench-cmd suites pass locally (the two known environment-dependent failures on this machine, the symlink-privilege serve test and the alt-screen scroll test, fail identically on main).
- Real-binary smoke of the three help/usage surfaces and both failure modes (unknown name, inline empty value).

## Summary by CodeRabbit

* **New Features**
  * Added execution profiles: `balanced`, `fast`, `thorough`.
  * Added `--exec-profile` support to execution and turn-benchmark commands, including stamping the selected profile into benchmark output; bumped turn benchmark schema to record it.
  * Added TUI `/profile` controls for selecting and switching profiles.
* **Bug Fixes**
  * Improved reasoning-effort restoration during posture escalation, with correct precedence and pinned override behavior.
* **Documentation**
  * Updated `zero exec` help and benchmark summaries to reflect exec-profile behavior.
* **Tests**
  * Expanded coverage for parsing/validation, precedence, tracing, benchmark propagation, and TUI profile interactions.

### DIFF
--- a/cmd/zero-perf-bench/turn.go
+++ b/cmd/zero-perf-bench/turn.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Gitlawb/zero/internal/execprofile"
 	"github.com/Gitlawb/zero/internal/perfbench"
 )
 
@@ -217,6 +218,20 @@ func parseTurnArgs(args []string, getenv func(string) string) (turnOptions, erro
 	}
 	if strings.TrimSpace(options.Model) == "" && !options.DryRun {
 		return options, fmt.Errorf("--model is required (or pass --dry-run)")
+	}
+	// Validate the profile here, before anything spawns: an unknown name would
+	// otherwise make every child exit with a usage error in milliseconds, and
+	// those near-zero walls would be recorded as valid latency samples in an
+	// exit-0 report — exactly the misread-as-improvement trap the harness
+	// closed for spawn failures. Normalizing to the catalog name also keeps
+	// the stamped execProfile canonical (Lookup is case-insensitive), so two
+	// captures of the same posture always compare equal.
+	if raw := strings.TrimSpace(options.ExecProfile); raw != "" {
+		profile, ok := execprofile.Lookup(raw)
+		if !ok {
+			return options, fmt.Errorf("unknown execution profile %q for --exec-profile. Valid profiles: %s", raw, strings.Join(execprofile.Names(), ", "))
+		}
+		options.ExecProfile = profile.Name
 	}
 	return options, nil
 }
--- a/cmd/zero-perf-bench/turn.go
+++ b/cmd/zero-perf-bench/turn.go
@@ -20,6 +20,7 @@ type turnOptions struct {
 	SuitePath   string
 	Model       string
 	Mode        string
+	ExecProfile string
 	SelfCorrect bool
 	Binary      string
 	Iterations  int
@@ -64,6 +65,7 @@ func runTurnCommand(args []string, getenv func(string) string, stdout io.Writer,
 	result, err := perfbench.RunTurnBench(context.Background(), set, perfbench.TurnBenchConfig{
 		Model:       options.Model,
 		Mode:        options.Mode,
+		ExecProfile: options.ExecProfile,
 		SelfCorrect: options.SelfCorrect,
 		Version:     options.Version,
 		Commit:      options.Commit,
@@ -136,6 +138,13 @@ func parseTurnArgs(args []string, getenv func(string) string) (turnOptions, erro
 				return options, err
 			}
 			options.Mode = value
+			index = next
+		case "--exec-profile":
+			value, next, err := readOptionValue(args, inlineValue, index, flag)
+			if err != nil {
+				return options, err
+			}
+			options.ExecProfile = value
 			index = next
 		case "--binary":
 			value, next, err := readOptionValue(args, inlineValue, index, flag)
@@ -239,6 +248,9 @@ func turnHelpText() string {
 		"  --suite <path>      Task set JSON file (required)",
 		"  --model <model>     Model to run (required unless --dry-run)",
 		"  --mode <name>       Exec mode preset to apply",
+		"  --exec-profile <name>",
+		"                      Execution profile for every task (fast|balanced|thorough);",
+		"                      forwarded to zero exec and stamped into the result",
 		"  --self-correct      Enable the post-edit verify-and-correct loop",
 		"  --binary <path>     Path to the `zero` binary (default: zero on PATH / repo root)",
 		"  --iterations <n>    Times to run each task (default: 1)",
--- a/cmd/zero-perf-bench/turn_test.go
+++ b/cmd/zero-perf-bench/turn_test.go
@@ -1,6 +1,9 @@
 package main
 
-import "testing"
+import (
+	"strings"
+	"testing"
+)
 
 func TestParseTurnArgsExecProfile(t *testing.T) {
 	getenv := func(string) string { return "" }
@@ -27,5 +30,36 @@ func TestParseTurnArgsExecProfile(t *testing.T) {
 	}
 	if options.ExecProfile != "thorough" {
 		t.Fatalf("ExecProfile = %q, want thorough", options.ExecProfile)
+	}
+}
+
+// The bench validates the profile BEFORE spawning anything: an unknown name
+// would otherwise make every child exit with a usage error in milliseconds
+// and record those near-zero walls as valid latency samples in an exit-0
+// report. Lookup normalization also keeps the stamped name canonical so two
+// captures of the same posture compare equal.
+func TestParseTurnArgsExecProfileValidatesAndNormalizes(t *testing.T) {
+	getenv := func(string) string { return "" }
+
+	options, err := parseTurnArgs([]string{
+		"--suite", "suite.json", "--model", "m", "--exec-profile", "FAST",
+	}, getenv)
+	if err != nil {
+		t.Fatalf("parseTurnArgs: %v", err)
+	}
+	if options.ExecProfile != "fast" {
+		t.Fatalf("ExecProfile = %q, want the canonical fast", options.ExecProfile)
+	}
+
+	_, err = parseTurnArgs([]string{
+		"--suite", "suite.json", "--model", "m", "--exec-profile", "blanced",
+	}, getenv)
+	if err == nil {
+		t.Fatal("an unknown profile must fail parse, before anything spawns")
+	}
+	for _, name := range []string{"balanced", "fast", "thorough"} {
+		if !strings.Contains(err.Error(), name) {
+			t.Fatalf("error must list %q, got %q", name, err.Error())
+		}
 	}
 }
--- a/cmd/zero-perf-bench/turn_test.go
+++ b/cmd/zero-perf-bench/turn_test.go
@@ -1,0 +1,31 @@
+package main
+
+import "testing"
+
+func TestParseTurnArgsExecProfile(t *testing.T) {
+	getenv := func(string) string { return "" }
+	options, err := parseTurnArgs([]string{
+		"--suite", "suite.json",
+		"--model", "m",
+		"--exec-profile", "fast",
+	}, getenv)
+	if err != nil {
+		t.Fatalf("parseTurnArgs: %v", err)
+	}
+	if options.ExecProfile != "fast" {
+		t.Fatalf("ExecProfile = %q, want fast", options.ExecProfile)
+	}
+
+	// Inline form.
+	options, err = parseTurnArgs([]string{
+		"--suite=suite.json",
+		"--model=m",
+		"--exec-profile=thorough",
+	}, getenv)
+	if err != nil {
+		t.Fatalf("parseTurnArgs inline: %v", err)
+	}
+	if options.ExecProfile != "thorough" {
+		t.Fatalf("ExecProfile = %q, want thorough", options.ExecProfile)
+	}
+}
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -202,6 +202,8 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			}
 			if target.ReasoningEffort != "" {
 				options.ReasoningEffort = target.ReasoningEffort
+			} else if target.RestoreDefaultEffort {
+				options.ReasoningEffort = ""
 			}
 			if target.RestoreCompletionGate {
 				options.RequireCompletionSignal = true
--- a/internal/agent/profile_controller_test.go
+++ b/internal/agent/profile_controller_test.go
@@ -153,6 +153,51 @@ func TestPostureEscalationOnUncertainCompletion(t *testing.T) {
 	}
 }
 
+// A profile that filled the run's effort displaced the provider default ("");
+// RestoreDefaultEffort lets escalation restore that default, which a plain ""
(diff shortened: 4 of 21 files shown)


Vasanthdev2004:
All four addressed in 1c56a40a:

- --use-spec + a self-correcting profile: went with the documented spec-safe projection rather than rejection. The profile's self-correct knob is skipped under --use-spec (the draft runner wires no corrector and the explicit flag combination is already a parse-time error), while the budget and effort knobs still apply to the draft. Covered by TestApplyExecProfileSpecProjectsAwaySelfCorrect.
- RunTurnBench now canonicalizes and validates ExecProfile at the boundary via the catalog lookup: " FAST " runs and stamps as "fast", and an unknown name errors before any task runs. Tests updated for both.
- Reverting a profile whose displaced budget was zero now clears ZERO_MAX_TURNS explicitly (SetMaxTurnsEnv ignores zero). TestProfileCommandRevertFromZeroClearsMaxTurnsEnv pins it, including the env value mid-profile.
- The effort-disarm test fails instead of skipping when the catalog stops filling low effort.

gnanam1990 (dismissed):
**APPROVE (code)** — high quality, thoroughly verified. **One pre-merge requirement: rebase off the stacked chain (it genuinely cannot merge before #736).**

Reviewed across 7 dimensions with an independent verifier re-checking each finding against the code. 0 logic blockers, 0 logic majors.

**Verified correct:**
- **Precedence** (flag > mode > profile): `applyExecProfile` runs after `applyExecMode` with fill-only-if-unset; an explicit `-r`, a mode's effort, or `--spec-reasoning-effort` is never overwritten. ✅
- **Balanced is byte-identical**: `Balanced = Profile{Name:"balanced"}` → nil policy, fills nothing; asserted by `reflect.DeepEqual` in `TestExecProfileBalancedLeavesOptionsUntouched`. ✅
- **`RestoreDefaultEffort`**: armed only when the profile actually filled effort (`reasoningEffort=="" && profile.ReasoningEffort!=""`); honored only when the target effort is empty (loop.go:203-207); disarmed on every user path (/effort, Ctrl+T, --spec-reasoning-effort) via policy-clone. #736 one-shot invariant intact. ✅
- **TUI /profile**: never stacks (reverts before applying), revert is coincidence-safe (restores a knob only if untouched AND still equal to the applied value), /turns pins the budget by disarming the escalation turn target — all clone-by-value. ✅
- **perfbench**: schema v4→v5 additive-only, name validated/canonicalized before spawning, forwarded+stamped+printed next to `mode:`, omitempty when unset; baseline.json diff empty. ✅
- Risk trigger = `critical` not `high` — sound rationale (sandbox floors shell at high). ✅

**Findings:**

**1. [Major — process] Stacked on 5 unmerged PRs.** `upstream/main..pr740` = 8 commits; only the top 3 are real PR10b/c — the lower 5 are #737/#736/#730/#734/#732. Merging as-is drags their diffs into main and conflicts when they land on their own. Genuine hard dependency: loop.go/types.go here consume `RestoreDefaultEffort`/`RestoreCompletionGate` from #736, so this cannot merge before #736. Enforce merge order (#730→#732→#734→#736→#737), then rebase; review the isolated `git diff 015452c..pr740` (~20 files).

**2. [Nit] `applyProfileTurnBudget(… explicitMaxTurns int …)`** also receives a mode-filled budget, so the name misleads (behavior is correct — mode must block the profile). Rename to `pinnedMaxTurns` or add a comment.

**3. [Nit] `ZERO_MAX_TURNS` not cleared on revert when the displaced budget was 0** (session_controls.go:395-398 + `SetMaxTurnsEnv` no-op for n≤0). Effectively unreachable (TUI budget defaults to 80, only 0 on a failed config resolve), but unconditionally restoring/clearing on revert would close it.

Ship-quality once rebased. Merge is kevin's call per the program gate, after the base chain lands.

anandh8x (changes requested):
Reviewed latest head 1c56a40a. The follow-up correctly fixes the benchmark boundary normalization and validation, zero-budget environment restoration, and the skipped effort regression; the spec-safe projection is coherent for a draft-only run. One TUI correctness blocker remains: active profile effort is applied only when /profile is selected, but both model-switch paths can change the model afterward without reconciling that profile state. For example, select fast on gpt-4.1 where low is unsupported, then switch through the provider/model picker to claude-sonnet-4.5: execProfileName remains fast, but reasoningEffort stays auto, so the next run is labeled fast without the catalog's low effort. The reverse cross-provider switch can retain a profile-applied effort unsupported by the destination. Reconcile an untouched active profile after both handleModelCommand and switchProviderModel: remove the old profile-applied effort, apply its configured level when the destination supports it, and update RestoreDefaultEffort consistently, while preserving explicitly touched effort choices. Please add regressions for unsupported-to-supported and supported-to-unsupported switches. Focused profile race tests, vet, formatting, diff hygiene, and govulncheck pass on this head.

Vasanthdev2004:
All addressed in b621b5bf.

@anandh8x, good catch on the model-switch hole. Both switch paths (the /model command and the cross-provider picker) now call a shared reconcile step after the model changes: an untouched active profile re-derives its effort fill for the destination, so unsupported-to-supported fills the profile's level exactly as selecting it there would, supported-to-unsupported drops back to the model default instead of carrying a level the destination cannot take, and the escalation's effort restore arms/disarms in step. Explicitly touched efforts are never reconciled. Regressions added for all three: TestProfileEffortReconciledOnModelSwitch covers both directions end to end through switchProviderModel (real catalog model with low support on one side, custom no-catalog model on the other), and TestProfileEffortTouchedSurvivesModelSwitch pins the touched case.

@gnanam1990, on the stacked-chain finding: I think your local upstream/main is behind. The branch is cut from 015452c1, which is current main and already contains the whole chain merged (#730 dbd94430, #732/#734, #736 af875df5, #737 015452c1 itself). GitHub's compare against main shows only this PR's 20 files, and there is nothing to rebase onto; a fresh fetch should show origin/main..this-branch as just the PR commits. The #736 dependency you flagged is real and satisfied, since #736 merged yesterday. Your env-clear nit landed one commit before your review (1c56a40a), and the pinned-budget parameter is renamed with the mode case documented.

CodeRabbit's LookupEnv tightening on the revert test is in as well.

Vasanthdev2004:
Fixed in 86c2c037, with one deliberate divergence from the literal suggestion.

Preserving a touched effort through the drop is not safe here: the TUI run path forwards m.reasoningEffort to the provider unfiltered (model.go, options.ReasoningEffort = string(m.reasoningEffort)), so a KNOWN-unsupported level surviving the switch goes straight into requests. That drop predates profiles and applies to a plain explicit effort too; a profile-touched effort should not be stickier than a plain one.

What was genuinely wrong, and is now fixed: the drop ran before the reconciliation and left contradictory profile state behind. A dropped touched override kept blocking the profile from governing, and a dropped profile fill left a stale applied value with the effort restore still armed. Now the whole sequence lives in one helper driven by the switch target's authoritative ring: drop a known-unsupported preference, void the profile's effort bookkeeping when the drop erased its fill or an override of it, then re-derive the fill for the destination. So /profile fast, /effort high, then /model to a ringless model lands on auto with clean state, and a later switch to a low-supporting model fills the profile's low again with the restore re-armed.

The difference from switchProviderModel is intentional and now documented at that call site: cross-provider targets are often custom models the catalog cannot vouch for either way, so an explicit preference is carried there (pre-existing behavior) while the profile's own fill stays conservative.

Tests: the void-then-refill arc and the supported-override survival are covered against synthetic rings (TestProfileEffortVoidedByModelSwitchDrop, TestProfileEffortTouchedSurvivesSupportedModelSwitch), alongside the existing end-to-end switch tests.

anandh8x (dismissed):
Re-reviewed latest head 86c2c037. The original model-switch blocker and the follow-on touched-effort inconsistency are resolved. Both switch paths now re-derive profile-filled effort for the destination; known-unsupported explicit effort is dropped without leaving stale touched/applied/restore state, supported destinations refill the profile, and unknown custom-provider models conservatively retain explicit user intent. The escalation restore flag stays synchronized. Focused TUI/profile race tests, vet, and diff hygiene pass. Approving.

gnanam1990 (dismissed):
**Re-approve** — all my prior findings are resolved and the base is now clean.

**1. Stacking (my prior Major) — RESOLVED.** The base chain (#730/#732/#734/#736/#737) merged; `upstream/main..pr740` is now just the 6 PR10b/c commits and the diff is the real ~21 files (+1605/-28), no longer inflated. The #736 dependency is satisfied on main.

**2. Nit (param naming) — FIXED.** `applyProfileTurnBudget(explicitMaxTurns …)` → `pinnedMaxTurns`, with a comment noting it covers a mode-preset fill too.

**3. Nit (ZERO_MAX_TURNS on revert) — FIXED.** `revertExecProfile` now `os.Unsetenv(MaxTurnsEnv)` when the displaced budget was 0, so spawned sub-agents can't inherit a removed profile's budget; covered by `TestProfileCommandRevertFromZeroClearsMaxTurnsEnv`.

**New since my last pass — model-switch effort reconciliation, verified correct:** `reconcileEffortForModelSwitch`/`reconcileProfileAfterModelSwitch` re-derive an active profile's per-model effort fill across switches — supported→unsupported clears the fill and disarms `RestoreDefaultEffort`; the reverse refills and re-arms; an explicitly touched effort is never reconciled; policy edits are clone-by-value so an in-flight run isn't mutated. Symmetric and correct, with 4 new directional tests. The spec-safe `profile.SelfCorrect && !options.useSpec` projection is a sensible edge fix.

Ran `go test ./internal/tui ./internal/execprofile ./internal/cli ./internal/agent -run 'Profile|ExecProfile|Posture|Escalat|ModelSwitch'` on the PR head — all green.

Ship-quality. Note CodeRabbit still shows CHANGES_REQUESTED from earlier commits — worth confirming those threads are addressed/dismissed before merge. Merge is kevin's per the gate.

Vasanthdev2004:
Point taken, fixed in 6778eb50. The drop now requires an authoritative ring: reconcileEffortForModelSwitch takes ringKnown (target.entry != nil), so a catalog model with a legitimately empty ring still drops a known-unsupported preference, while a live-discovered/custom target with no catalog entry preserves an explicit choice, matching the cross-provider picker. The profile's own fill is unchanged in both regimes: it only ever applies where support is known. Both regimes are tested separately (TestProfileEffortVoidedByModelSwitchDrop for authoritative-empty, TestProfileEffortUnknownRingPreservesExplicitChoice for unknown, including the untouched-fill retreat).

For anyone skimming this thread: the other four comments in this round are re-posts of findings already resolved in 1c56a40a and b621b5bf, and each carries its own addressed marker.

anandh8x (approved):

gnanam1990 (approved):
**Re-approve** (head `6778eb5`). One new commit since my last pass — `fix(tui): only drop an explicit effort when the target ring is authoritative` — reviewed and correct.

It fixes an over-eager drop: previously a `nil` effort ring was treated as unsupported and cleared an explicit effort, but `nil` conflated a catalog model with no effort controls (authoritatively empty → drop is right) with a live-discovered/custom model the catalog can't vouch for (unknown → dropping a valid preference is wrong). The new `ringKnown` flag gates the drop on an authoritative ring (`target.entry != nil`); an unknown ring preserves the explicit preference, matching the cross-provider picker, while the profile's own fill stays conservative (applies only where support is known).

Verified:
- Callers consistent: `handleModelCommand` passes `target.entry != nil`; the cross-provider `switchProviderModel` uses only the conservative `reconcileProfileAfterModelSwitch` retreat (never the drop). ✅
- Tests updated coherently — `(nil, true)` still drops (authoritative-empty ring), new `TestProfileEffortUnknownRingPreservesExplicitChoice` covers `(nil, false)` for both an explicit choice (survives) and an untouched profile fill (retreats, restore disarmed). ✅
- Ran `go test ./internal/tui ./internal/execprofile ./internal/cli -run 'Profile|ExecProfile|ModelSwitch|Effort'` on the head — all green.

All prior findings remain resolved (stacking, param naming, ZERO_MAX_TURNS revert). Ship-quality; merge is kevin's per the gate.

PR #741, opened 2026-07-18
fix(oauth): refuse redirects on credential POSTs (#729)

### LINKED ISSUE #729: security(oauth): token POST credentials can leak through redirects

gnanam1990:
## Summary

Closes #729.

OAuth token requests validated only the initial endpoint, then let the supplied or default `http.Client` follow a `307`/`308` redirect that **replays the POST body** — authorization codes, PKCE verifiers, refresh tokens, client secrets, client IDs — to another origin (or an HTTPS→HTTP downgrade) that never passed endpoint validation.

## Affected paths

- `internal/oauth/flow.go` `PostToken` — token exchange + refresh (and MCP OAuth login/refresh, which delegate to `oauth.ExchangeCode`/`oauth.Refresh`).
- `internal/oauth/device.go` `RequestDeviceCode` — the device authorization POST (`client_id` + `client_secret`).
- `internal/oauth/device.go` `pollDeviceOnce` — the device-token poll (`device_code` + `client_secret`).

## Fix

New `withoutRedirects(client)` returns a **shallow copy** of the client whose `CheckRedirect` refuses every redirect (`ErrUnsafeRedirect`). Applied at all three credential POSTs before `client.Do`. Because the copy is per-request, the caller's client is never mutated, and a legitimate token endpoint that returns `200` is unaffected (real token endpoints don't redirect).

## Scope

`internal/oauth` only. No config or API change. No behavior change for non-redirecting endpoints. No performance change expected.

## Testing

- 5 regression tests (`PostToken`, `ExchangeCode`, `Refresh`, `RequestDeviceCode`, `pollDeviceOnce`) using a redirect trap: a credential endpoint that `307`/`308`-redirects to an attacker server which counts hits. Each asserts the call returns `ErrUnsafeRedirect` **and the attacker receives zero requests** (the body was never replayed). All five were confirmed to **fail without the fix** (reverted the source, kept the tests).
- `go test -race ./internal/oauth/... ./internal/mcp/...` green. `go build ./...`, `go vet ./...`, `gofmt`, `govulncheck`, `git diff HEAD --check` all clean.
- **Manual end-to-end through the real CLI.** With a device endpoint that `307`-redirects to an attacker, `zero auth login <provider> --device` now fails with `refusing to follow a redirect from a credential endpoint` (exit 1) and the attacker server receives **0** requests — whereas unpatched `main` replays the `client_secret`-bearing POST (attacker hit) and proceeds.

## Related

Pairs with #511 (validate discovered endpoints) — the other half of the OAuth endpoint-safety hardening. Independent branch; no stacking.

## Summary by CodeRabbit

* **Security Improvements**
  * OAuth credential-bearing requests now refuse temporary/permanent redirects to prevent sensitive form data from being replayed to unvalidated destinations.
  * Added a clear `ErrUnsafeRedirect` error surfaced when redirects are blocked for token exchange, refresh, device authorization, and device polling.
* **Tests**
  * Added coverage ensuring redirects are rejected and that no additional requests occur to confirm sensitive payloads are not replayed.

### DIFF
--- a/internal/oauth/device.go
+++ b/internal/oauth/device.go
@@ -49,9 +49,9 @@ func RequestDeviceCode(ctx context.Context, client *http.Client, cfg Config, now
 	if err := validateTokenEndpoint(endpoint); err != nil {
 		return DeviceAuth{}, err
 	}
-	if client == nil {
-		client = http.DefaultClient
-	}
+	// Refuse redirects so a device-authorization redirect can't replay the
+	// client_id/client_secret POST body to an unvalidated origin.
+	client = withoutRedirects(client)
 	if now == nil {
 		now = time.Now
 	}
@@ -157,9 +157,9 @@ func pollDeviceOnce(ctx context.Context, client *http.Client, cfg Config, device
 	if err := validateTokenEndpoint(cfg.TokenEndpoint); err != nil {
 		return Token{}, err
 	}
-	if client == nil {
-		client = http.DefaultClient
-	}
+	// Refuse redirects so a poll redirect can't replay the device_code/secret
+	// POST body to an unvalidated origin.
+	client = withoutRedirects(client)
 	form := url.Values{}
 	form.Set("grant_type", deviceGrantType)
 	form.Set("device_code", deviceCode)
--- a/internal/oauth/device_test.go
+++ b/internal/oauth/device_test.go
@@ -2,6 +2,7 @@ package oauth
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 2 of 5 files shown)


anandh8x (approved):
Reviewed the OAuth redirect hardening at 0919a848. The credential-bearing token exchange, refresh, device authorization, and device polling paths reject redirects without mutating the caller's HTTP client. Focused race tests for internal/oauth and internal/mcp pass, focused vet is clean, and CI is green. No blocking findings.

Vasanthdev2004 (approved):
Approving.

`withoutRedirects` shallow-copies the client and sets `CheckRedirect` to return `ErrUnsafeRedirect`, applied at all three credential POSTs: `PostToken` (which backs ExchangeCode/Refresh and the MCP delegations), `RequestDeviceCode`, and `pollDeviceOnce`. Go treats a `CheckRedirect` error as fatal and returns before issuing the follow-up request, so a 307/308 never replays the form body (codes, PKCE verifiers, refresh tokens, secrets), and it also blocks an HTTPS to HTTP downgrade. The caller's client is left unmutated, so a caller-supplied client can't smuggle the redirect back in. The regression tests are convincing: each asserts `errors.Is(err, ErrUnsafeRedirect)` AND that the attacker server got zero hits, which is exactly the replay threat.

Two non-blocking notes:
- `registerClient` in `internal/mcp/oauth.go:303` POSTs without the same guard. It's outside #729's scope (the request body carries no pre-existing credential, just client metadata), but the response carries a `client_secret`, so applying `withoutRedirects` there too would be good defense in depth. Fine as a follow-up.
- This overlaps #738, which patches only the token endpoint. #741 is the broader fix (adds device-code + poll and 5 tests), so I'd land this one and close #738 in its favor. Merge order is your / kevin's call.

CI green, oauth and mcp packages green locally.

PR #745, opened 2026-07-19
fix(sandbox): AST second opinion for interactive-command bypasses (#473)

### LINKED ISSUE #473: sandbox: complex hand-written shell parsing in safe_command risks bypasses
## Problem

`splitShellSegments` (and related quote/substitution/pipe handling) in the interactive command blocker is a large, hand-written parser:

```go
// internal/sandbox/safe_command.go:480+
func splitShellSegments(command string) []string { ... }  // 100+ lines of quote, $(), `, |, ;, && logic
```

Used to block dangerous interactive programs (vim, less, `tail -f`, REPLs, etc.).

## Impact

- High chance of bypass or false negative with unusual quoting, command substitutions, or edge cases.
- Could allow the agent to hang on an editor/pager/REPL until the per-command timeout.
- Very hard to audit and maintain.

## Suggested fix

- Replace or augment with a well-tested shell parser (e.g. `mvdan.cc/sh/v3/syntax` which is already used elsewhere in the codebase for prefix matching).
- Or significantly simplify + add many more adversarial test cases.
- Document the limitations clearly.

gnanam1990:
## Summary

Closes #473.

The interactive-command guard (`DetectInteractiveCommand`, used by the bash and exec tools before launching a command) split shell commands with a large hand-written parser, `splitShellSegments`. That splitter mis-handles unusual quoting, command substitution, subshells, and newline separators, so an interactive program hidden by one of those constructs slips through — the agent has no TTY, so it hangs on the editor/pager/REPL until the per-command timeout.

## Fix — augment with the shell AST (not replace)

The repo already parses shell with `mvdan.cc/sh/v3/syntax` in `analyzer.go`, and `risk.go` already uses it as an additive "second opinion." This applies the same pattern to interactive detection:

- New `astCommandFields()` parses the command and re-extracts the real simple commands (program + args) from the tree, resolving the positions the regex splitter mis-reads.
- `DetectInteractiveCommand` runs the hand-written path first (unchanged), then, as a backstop, classifies each AST-extracted command with the **same per-program predicates** it already uses — `interactivePrograms` + `hasNonInteractiveFlag` (which includes the `ssh`/`sqlite3`/`redis-cli` trailing-command suppression).

Because the backstop reuses the existing predicates, **every program is classified exactly as before** (`ssh host 'uptime'`, `python -c '…'`, `mongosh --eval …` stay allowed); it only ADDS detections for the splitter's bypasses. There are no false positives: the AST distinguishes a program position from an argument, so an interactive program *name* inside a quoted argument (`echo "run vim later"`) is never flagged. A command the parser cannot handle (Windows cmd.exe, obfuscation) yields no commands and falls through to the regex path — the guard never hard-blocks on a parse error.

## Scope

`internal/sandbox` only. No config or API change. No performance change expected.

## Testing

- New tests: `TestDetectInteractiveCommandCatchesParserBypasses` (a newline-separated and a brace-group invocation the hand splitter misses are now caught, with the right program name), and `TestDetectInteractiveCommandNoFalsePositiveOnQuotedArgument` (an interactive name inside a quoted argument stays non-interactive). The bypass test was confirmed to **fail without the fix**; all existing `safe_command` tests (including `ssh host 'uptime'` non-interactive) still pass.
- `go test -race ./internal/sandbox/...` green. `make fmt-check`, `go vet ./...`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, `govulncheck`, `git diff HEAD --check` all clean.

## Note

An earlier draft used the AST's own `Interactive` flag directly, which over-blocked `ssh` with a trailing command (the AST's suppression is cruder than the guard's). The committed version reuses the guard's `hasNonInteractiveFlag`, so classification is identical to before and only the bypasses change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of interactive shell commands in complex command strings by adding an AST-based “second opinion” stage that better handles quoting and nested shell payloads.
  * Reduced false positives by only extracting candidate commands from literal (non-dynamic) word content, including safer handling when names appear inside quoted arguments.

* **Tests**
  * Added coverage for parser-bypass cases, quoted-argument non-detection, pipeline/segmentation boundary behavior, and ensuring dynamic program/argument content does not trigger interactive detection.

### DIFF
--- a/internal/sandbox/analyzer.go
+++ b/internal/sandbox/analyzer.go
@@ -65,6 +65,34 @@ func AnalyzeCommand(script string) AnalysisResult {
 	return result
 }
 
+// astCommandFields parses command with the shell parser and returns each simple
+// command as its literal field slice (program + args as text), resolving the
+// real command positions across quoting, command substitution, subshells, and
+// newline separators — the constructs the hand-written splitter in
+// safe_command.go mis-handles (issue #473). It returns nil when the command
+// cannot be parsed (e.g. a Windows cmd.exe string), so callers fall through to
+// the regex path rather than hard-blocking.
+func astCommandFields(command string) [][]string {
+	file, err := syntax.NewParser().Parse(strings.NewReader(command), "")
+	if err != nil {
+		return nil
+	}
+	var commands [][]string
+	syntax.Walk(file, func(node syntax.Node) bool {
+		call, ok := node.(*syntax.CallExpr)
+		if !ok || len(call.Args) == 0 {
+			return true
+		}
+		fields := make([]string, 0, len(call.Args))
+		for _, word := range call.Args {
+			fields = append(fields, wordText(word))
+		}
+		commands = append(commands, fields)
+		return true
+	})
+	return commands
+}
+
 // analyzeInto parses script and folds its interactive/destructive/network usage
 // into result, sharing seen so program names are de-duplicated across recursion.
 func analyzeInto(script string, result *AnalysisResult, seen map[string]bool, depth int) {
--- a/internal/sandbox/analyzer.go
+++ b/internal/sandbox/analyzer.go
@@ -83,11 +83,22 @@ func astCommandFields(command string) [][]string {
 		if !ok || len(call.Args) == 0 {
 			return true
 		}
-		// The program name must be a static literal. A dynamic first word (e.g.
-		// `$(printf foo)vim`, which runs as `foovim`) would otherwise yield a
-		// misleading partial literal ("vim") and fabricate an interactive match.
-		if !isLiteralWord(call.Args[0]) {
-			return true
+		// EVERY word must be a static literal, not just the program name.
+		// wordText keeps only the literal/quoted parts of a word and silently
+		// drops expansions, so a dynamic word anywhere in the call reconstructs
+		// to something the shell will never run: `$(printf foo)vim` (runs as
+		// `foovim`) collapses to "vim", and `git $(printf foo)rebase -i` (runs
+		// as `git foorebase -i`, non-interactive) collapses to `git rebase -i`
+		// and fabricates an interactive match. Since the runtime value of an
+		// expansion is unknowable here, skip the whole call rather than classify
+		// a lossy reconstruction. Skipping is the safe direction: the
+		// hand-written passes above already ran, and a missed detection falls
+		// through to the normal permission prompt instead of hard-blocking a
+		// command the user never wrote.
+		for _, word := range call.Args {
+			if !isLiteralWord(word) {
+				return true
+			}
 		}
 		fields := make([]string, 0, len(call.Args))
 		for _, word := range call.Args {
--- a/internal/sandbox/analyzer.go
+++ b/internal/sandbox/analyzer.go
@@ -83,6 +83,12 @@ func astCommandFields(command string) [][]string {
 		if !ok || len(call.Args) == 0 {
 			return true
 		}
+		// The program name must be a static literal. A dynamic first word (e.g.
+		// `$(printf foo)vim`, which runs as `foovim`) would otherwise yield a
+		// misleading partial literal ("vim") and fabricate an interactive match.
+		if !isLiteralWord(call.Args[0]) {
+			return true
+		}
 		fields := make([]string, 0, len(call.Args))
 		for _, word := range call.Args {
 			fields = append(fields, wordText(word))
@@ -91,6 +97,31 @@ func astCommandFields(command string) [][]string {
 		return true
 	})
 	return commands
+}
+
+// isLiteralWord reports whether every part of word is a static literal (bare or
+// quoted). A word containing a command substitution, parameter/arithmetic
+// expansion, process substitution, etc. is dynamic — its runtime value is
+// unknown, so its wordText (a partial literal) must not be trusted as a program
+// name.
+func isLiteralWord(word *syntax.Word) bool {
+	if word == nil {
+		return false
+	}
+	for _, part := range word.Parts {
+		switch typed := part.(type) {
+		case *syntax.Lit, *syntax.SglQuoted:
+		case *syntax.DblQuoted:
+			for _, inner := range typed.Parts {
+				if _, ok := inner.(*syntax.Lit); !ok {
+					return false
+				}
+			}
+		default:
+			return false
+		}
+	}
+	return true
 }
 
 // analyzeInto parses script and folds its interactive/destructive/network usage
--- a/internal/sandbox/safe_command.go
+++ b/internal/sandbox/safe_command.go
@@ -219,6 +219,39 @@ func DetectInteractiveCommand(command string, goos string) InteractiveCommandRes
 		}
 	}
 
+	// AST second opinion (issue #473): the hand-written segment split above
+	// mis-handles interactive programs hidden by unusual quoting, command
+	// substitution, subshells, or newline separators. Re-extract the real simple
+	// commands from the parsed shell tree and apply the SAME per-program checks
+	// used above, so a bypass is caught while every program stays classified
+	// exactly as before (ssh with a trailing command, python -c, etc. remain
+	// allowed via hasNonInteractiveFlag). A command the parser cannot handle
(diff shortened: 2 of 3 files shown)


gnanam1990:
Addressed in f8c902c.

**Major (make the AST pass mirror the pipeline):** the AST second opinion now runs the full pipeline via a new `inspectCommandFields` helper — multi-word interactive segments and `sh -c <payload>` recursion, then the per-program lookup — so `{ git rebase -i HEAD~1; }` and `{ sh -c 'vim file.txt'; }` are now caught. Regression tests added (`TestDetectInteractiveCommandASTPipelineBoundaries`).

**Fabricated program names:** `astCommandFields` now skips a CallExpr whose program word is not a static literal (`isLiteralWord`), so a substitution concatenated with a literal (`$(printf foo)vim` → runs as `foovim`) can't fabricate an interactive "vim" from the AST pass. Verified directly at the extractor (`TestAstCommandFieldsSkipsDynamicProgram`).

**On `$(printf '%s' foo)vim file.txt` remaining interactive:** that is a **pre-existing hand-parser behavior** — it already resolves to `vim` on `main` *before* this PR (the hand-written path normalizes the wrapping away and reaches its own substitution handling first). This PR only adds the AST second opinion, whose guard is in place; fixing the hand parser's concatenation handling is a separate, out-of-scope change. So I did not assert the un-braced form is non-interactive.

@coderabbitai review

anandh8x (changes requested):
Confirmed regression at f8c902c: astCommandFields validates only the program word, then wordText discards expansions from every argument before inspectCommandFields classifies them. A brace-grouped git command whose rebase argument is prefixed by $(printf foo) executes git foorebase -i HEAD~1 and is non-interactive, but this PR reconstructs git rebase -i HEAD~1 and blocks it. The same regression test passes on the PR base, so this false positive is newly introduced. Please preserve argument literalness and avoid classification from lossy dynamic arguments, or skip those AST commands, and add this case as a regression test. CI and focused vet are otherwise clean.

Vasanthdev2004 (approved):
Approving.

This is genuinely additive, which is the important property for a guard change: `DetectInteractiveCommand` still runs the hand-written passes first and both return early on a hit, so the new AST loop only runs when the hand path found nothing. It can add a detection but never suppress one. The AST path re-extracts the real simple commands via `mvdan.cc/sh` and runs them through `inspectCommandFields`, which reuses the same predicates the guard already has (interactive segments, `sh -c` recursion, `interactivePrograms` + `hasNonInteractiveFlag` with the ssh/sqlite3/redis-cli trailing-command suppression), so `ssh host 'uptime'`, `python -c`, and `node -e` stay non-interactive. Fail-open is handled too: a parse error returns nil and falls through to the hand path, and `isLiteralWord` stops a dynamic first word like `$(printf foo)vim` from fabricating a match.

One suggestion, not a blocker: the issue calls out subshells, command substitution, and newline separators specifically, and the new tests cover the brace-group and newline cases but not a bare subshell `( vim )`, a background `echo ok & vim`, or process substitution `sort <(vim)`. Those are caught now, but a direct regression case for each would keep a future refactor from silently regressing them. Worth adding while you're in here.

CI green, sandbox package green locally.

Vasanthdev2004 (changes requested):
Re-verified on `f8c902c0`, and anand's regression reproduces. My earlier approval doesn't hold; apologies for missing it.

Confirmed on this branch versus the base:
- `{ git $(printf foo)rebase -i HEAD~1 ; }` on the base (main) gives `Interactive=false` (correct: it runs `git foorebase`, a non-existent, non-interactive subcommand). On this PR it gives `Interactive=true, Command="git rebase -i"`, a false positive.

Root cause is exactly what anand described. `astCommandFields` guards only the program word (`isLiteralWord(call.Args[0])`), but each argument is then run through `wordText`, which keeps only literal/quoted parts and silently drops `*syntax.CmdSubst` / `*syntax.ParamExp`. So a dynamic argument (`$(printf foo)rebase`) is reconstructed lossily to `rebase`, and `inspectCommandFields` classifies the fabricated `git rebase -i`. The genuine `git rebase -i HEAD~1` still classifies correctly, so this is purely a new false positive on dynamic arguments.

Cleanest fix is to extend the literalness guard from the program word to every word: if any word in the call carries an expansion (`!isLiteralWord`), skip that AST command rather than classify a lossy reconstruction, since you can't know what it actually runs. That is the safe direction (skip, not false-block), and the hand-written path still runs first. A regression case for `{ git $(printf foo)rebase -i HEAD~1 ; }` expecting not-interactive would lock it in.

The rest still looks right (additive, hand-path-first, fails open on a parse error); this is the one thing blocking. Agreeing with anand's requested changes.

gnanam1990:
@coderabbitai review

Pushed `df2cd63` addressing the false positive @anandh8x and @Vasanthdev2004 both reported.

`astCommandFields` guarded only the program word, then ran every argument through `wordText`, which drops `CmdSubst`/`ParamExp`. So `git $(printf foo)rebase -i HEAD~1` — which actually runs `git foorebase -i`, non-interactive — was reconstructed as `git rebase -i` and blocked.

The guard now covers every word in the call: if any word carries an expansion, the AST command is skipped rather than classified from a lossy reconstruction. Skipping is the safe direction — the hand-written passes still run first and a miss falls through to the normal prompt instead of hard-blocking.

Verified: the reported case is back to `Interactive=false`, matching base; `git rebase -i HEAD~1` and the existing #473 bypass cases still classify exactly as before; regression test `TestAstCommandFieldsSkipsDynamicArgument` added and confirmed to fail without the fix.

gnanam1990:
@anandh8x @Vasanthdev2004 — ready for another look. Your reviews were both filed on `f8c902c0`; the fix is `df2cd633`.

**What you reported:** `astCommandFields` guarded only the program word, then ran every argument through `wordText`, which drops `CmdSubst`/`ParamExp` — so `git $(printf foo)rebase -i HEAD~1` (really `git foorebase -i`, non-interactive) was reconstructed as `git rebase -i` and blocked.

**What changed:** the literalness guard now covers every word in the call, not just `Args[0]`. Any word carrying an expansion means the AST command is skipped rather than classified from a lossy reconstruction — the safe direction Vasanth suggested, since the hand-written passes still run first and a miss falls through to the normal prompt instead of hard-blocking.

Re-verified just now at `df2cd633`:

```
{ git $(printf foo)rebase -i HEAD~1 }   interactive=false   (was true — your repro)
git rebase -i HEAD~1                     interactive=true    (genuine detection intact)
{ sh -c 'vim file.txt' ; }               interactive=true    (#473 bypass still caught)
```

`TestAstCommandFieldsSkipsDynamicArgument` covers it and I mutation-tested it — reverting the guard fails it. Every existing interactive-detection test still passes, and CodeRabbit has approved this head. All checks green.

anandh8x (approved):
Approved on the latest head. The AST fallback now applies the existing full detection pipeline and skips calls with dynamic words, avoiding the prior lossy-reconstruction false positive. Focused sandbox race tests and CI pass.

Vasanthdev2004 (approved):
Re-verified on the current head (`df2cd633`), and the fix resolves my earlier block. `astCommandFields` now requires EVERY word in the call to be a literal, not just the program name, so a dynamic argument makes it skip the whole AST command rather than reconstruct a lossy `git rebase -i` from `git $(printf foo)rebase -i`. That is exactly the fix, and skipping is the safe direction (the hand path already ran; a miss falls through to a prompt, not a hard block).

Confirmed empirically against the exact case I flagged plus the bypass classes #473 targets, all correct:

- `{ git $(printf foo)rebase -i HEAD~1 ; }` -> Interactive=false (the regression, now correctly skipped)
- `git rebase -i HEAD~1` -> Interactive=true (genuine, preserved)
- `{ git rebase -i HEAD~1 ; }` -> Interactive=true (brace-grouped genuine, still caught)
- `{ vim ; }`, `sort <(vim)`, `echo ok & vim` -> Interactive=true (brace group, process substitution, background separator all still caught)

So the dynamic-argument false positive is gone and none of the real detections regressed. Approving. Thanks for the quick turnaround.

PR #746, opened 2026-07-19
feat(sandbox): disable the sandbox via config (#687)

### LINKED ISSUE #687: Disable sandbox via config.json
Add an enabled field to the sandbox block of config.json (e.g. `"sandbox": { "enabled": false}`)

gnanam1990:
## Summary

Closes #687.

Adds a `sandbox.enabled` field to config, so `"sandbox": {"enabled": false}` turns the sandbox off. The engine already had a fully-wired `ModeDisabled` (it short-circuits evaluation and returns unwrapped command plans), but no config surface reached it.

## Changes

- `SandboxConfig.Enabled *bool` — a pointer so an explicit `false` (disable) is distinct from an omitted key (keep the default: enabled). `true`/omitted keep enforcement.
- `mergeConfig` honors it, so **global (user) config and CLI** can turn the sandbox off; `applyConfiguredSandboxPolicy` maps `enabled:false` to `sandbox.ModeDisabled`.
- `mergeProjectConfig` **intentionally does NOT merge it** — a cloned repo's `.zero/config.json` must not be able to disable the sandbox that constrains it. Same posture the file already takes for `AdditionalWriteRoots` (never from project) and `Network` (project may only tighten to `deny`).

## Scope

`internal/config` + `internal/cli` only. Additive field, no behavior change when the key is absent. No performance change expected.

## Testing

- `TestResolveSandboxEnabledUserConfigDisables` — global config `enabled:false` resolves to a false pointer.
- `TestResolveSandboxDisableIgnoredFromProjectConfig` — **security**: a project config `enabled:false` is ignored (resolves to nil), so a repo cannot disable the sandbox.
- `TestApplyConfiguredSandboxPolicyEnabledFalseDisables` — `enabled:false` → `ModeDisabled`; omitted/`true` keep enforcing.
- The two behavior tests were confirmed to **fail without the fix**. `make fmt-check`, `go vet ./...`, `go test -race ./internal/config/... ./internal/cli/...`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, `govulncheck`, `git diff HEAD --check` all clean.
- **Manual end-to-end via the real CLI** (`zero sandbox check`, macos-seatbelt backend): a global config `enabled:false` reports `policy: mode=disabled` / "sandbox disabled"; a project `.zero/config.json` `enabled:false` still reports `policy: mode=enforce` / "sandbox is active" — the repo's attempt to disable is ignored.

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to explicitly disable sandbox enforcement (`sandbox.enabled: false`).
  * Sandbox enforcement remains enabled by default when the option is omitted.
* **Bug Fixes**
  * Enforces sandbox enablement/disablement precedence correctly during configuration resolution (including overrides vs. lower-precedence settings).
* **Tests**
  * Added coverage to verify explicit disable behavior, ignore project-provider changes for this setting, and validate CLI/programmatic override precedence.

### DIFF
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -939,6 +939,12 @@ func providerSensitiveEnvKeys(resolved config.ResolvedConfig) []string {
 // applyConfiguredSandboxPolicy overlays every config-sourced sandbox knob onto
 // the default policy.
 func applyConfiguredSandboxPolicy(policy sandbox.Policy, cfg config.SandboxConfig) sandbox.Policy {
+	// An explicit `"sandbox": {"enabled": false}` disables the sandbox: every
+	// request is allowed and commands run unwrapped. Only reachable from global
+	// config / CLI — project config never sets Enabled (see mergeProjectConfig).
+	if cfg.Enabled != nil && !*cfg.Enabled {
+		policy.Mode = sandbox.ModeDisabled
+	}
 	if network := strings.TrimSpace(cfg.Network); network != "" {
 		switch sandbox.NetworkMode(network) {
 		case sandbox.NetworkAllow, sandbox.NetworkDeny:
--- a/internal/cli/sandbox_test.go
+++ b/internal/cli/sandbox_test.go
@@ -1001,3 +1001,18 @@ func sandboxPolicyCapabilityStatus(capabilities []sandbox.BackendCapability, key
 	}
 	return ""
 }
+
+func TestApplyConfiguredSandboxPolicyEnabledFalseDisables(t *testing.T) {
+	disabled := false
+	if got := applyConfiguredSandboxPolicy(sandbox.DefaultPolicy(), config.SandboxConfig{Enabled: &disabled}); got.Mode != sandbox.ModeDisabled {
+		t.Fatalf("enabled:false must yield ModeDisabled, got %q", got.Mode)
+	}
+	// Omitted (nil) and explicit true keep the default enforcing mode.
+	if got := applyConfiguredSandboxPolicy(sandbox.DefaultPolicy(), config.SandboxConfig{}); got.Mode == sandbox.ModeDisabled {
+		t.Fatal("omitted enabled must not disable the sandbox")
+	}
+	enabled := true
+	if got := applyConfiguredSandboxPolicy(sandbox.DefaultPolicy(), config.SandboxConfig{Enabled: &enabled}); got.Mode == sandbox.ModeDisabled {
+		t.Fatal("enabled:true must not disable the sandbox")
+	}
+}
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -216,6 +216,9 @@ func mergeConfig(dst *FileConfig, src FileConfig) {
 		mergeProvider(dst, provider)
 	}
 	mergeMCPConfig(&dst.MCP, src.MCP, true)
+	if src.Sandbox.Enabled != nil {
+		dst.Sandbox.Enabled = src.Sandbox.Enabled
+	}
 	if network := strings.TrimSpace(src.Sandbox.Network); network != "" {
 		dst.Sandbox.Network = network
 	}
@@ -273,6 +276,10 @@ func mergeProjectConfig(dst *FileConfig, src FileConfig) error {
 	if err := mergeProjectMCPConfig(&dst.MCP, src.MCP); err != nil {
 		return err
 	}
+	// Sandbox.Enabled is intentionally NOT merged from project config: a cloned
+	// repo's .zero/config.json must not be able to disable the sandbox that
+	// constrains it. Only global config and CLI can turn the sandbox off.
+	//
 	// Sandbox.AdditionalWriteRoots is intentionally NOT merged from project
 	// config: a cloned repo's .zero/config.json must not be able to grant
 	// itself write access outside the workspace. Global config and CLI flags
--- a/internal/config/resolver.go
+++ b/internal/config/resolver.go
@@ -695,6 +695,9 @@ func applyOverrides(cfg *FileConfig, overrides Overrides) {
 	if overrides.MaxTurns > 0 {
 		cfg.MaxTurns = overrides.MaxTurns
 	}
+	if overrides.Sandbox.Enabled != nil {
+		cfg.Sandbox.Enabled = overrides.Sandbox.Enabled
+	}
 	if overrides.Sandbox.BlockUnixSockets {
 		cfg.Sandbox.BlockUnixSockets = true
 	}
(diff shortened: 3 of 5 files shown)


anandh8x (changes requested):
Please merge sandbox.enabled through the CLI override path before this lands.

`mergeConfig` now copies `Sandbox.Enabled`, but `ResolveOptions.Overrides` is applied later by `applyOverrides`, which does not copy `overrides.Sandbox.Enabled`. As a result, an explicit CLI/programmatic override such as `Overrides{Sandbox: SandboxConfig{Enabled: &disabled}}` is ignored, contradicting the PR description and the new comment that global config / CLI can disable the sandbox.

Please update `applyOverrides` to preserve the tri-state pointer and add a resolver regression test showing that an explicit false override disables the sandbox (and, ideally, that explicit true can override a lower-precedence false).

gnanam1990:
Good catch — fixed in 2a68d9c. `applyOverrides` now preserves the `Sandbox.Enabled` tri-state pointer (it runs after every config source, so an explicit override wins). Added `TestResolveSandboxEnabledCLIOverride`: an explicit `false` override disables the sandbox with no config file, and an explicit `true` override wins over a lower-precedence config-file `false`. Verified it fails without the applyOverrides change.

gnanam1990:
@coderabbitai review

Vasanthdev2004 (dismissed):
Approving on the current head (`2a68d9c4`). anand's requested change is addressed here: `applyOverrides` now threads `Sandbox.Enabled`, and `TestResolveSandboxEnabledCLIOverride` covers both an explicit false override disabling the sandbox and an explicit true winning over a lower-precedence config false. @anandh8x this postdates your review, worth a re-look.

I verified the security posture first-hand (tests pass locally):

- Off by default: `Enabled *bool` nil (omitted) keeps enforcement; only an explicit `false` disables.
- A cloned repo CANNOT disable the sandbox that constrains it: `mergeProjectConfig` deliberately skips `Sandbox.Enabled`, and `TestResolveSandboxDisableIgnoredFromProjectConfig` locks that in. That is the property that matters most for a disable knob, and it is correct.
- Precedence is right: global config and CLI can disable, project config cannot, and an explicit true beats a config false. The `applyConfiguredSandboxPolicy` policy test and all three resolver tests pass.

One non-blocking thought: a disabled sandbox surfaces as "sandbox disabled" in status, which is good, but a more prominent session-start notice (so someone who set `enabled:false` globally and forgot does not run unsandboxed unaware) might be worth a follow-up. Not blocking.

anandh8x (changes requested):
Request changes: provider-command output still reaches mergeConfig after LoadProviderCommand, and mergeConfig now copies Sandbox.Enabled. A provider command that returns a valid provider plus "sandbox": {"enabled": false} can therefore disable the sandbox, despite this setting being documented as global-config/CLI only. Keep Enabled out of the provider-command merge path (or gate it on an explicit trusted source) and add a regression test that provider-command output cannot enable or disable the sandbox.

gnanam1990:
@coderabbitai review

Pushed `76ef636` addressing @anandh8x's finding — confirmed reproducible before the fix, in both directions.

A provider command returning the full-config shape (`{"providers":[…],"sandbox":{"enabled":false}}`) reached `Sandbox.Enabled` because `mergeConfig` is shared between the trusted global-config merge and the provider-command merge. The shorthand single-provider shape was already safe — `parseProviderCommandJSON` discards everything but the profile there — which is why this only bites via the full shape.

Fix clears `Sandbox.Enabled` from the provider-command config before merging, mirroring the carve-out `mergeProjectConfig` already applies for the same threat model. Done at the call site, not inside `mergeConfig`, so the trusted global-config merge keeps honouring the setting.

`TestResolveSandboxEnabledIgnoredFromProviderCommand` covers both directions and also asserts the provider itself still applies, so the guard can't be satisfied by dropping the command wholesale. Mutation-tested: removing the one-line fix fails it. Global-config disable, CLI override, and the project-config carve-out all still pass.

gnanam1990:
@anandh8x — ready for another look. Your review was on `2a68d9c4`; the fix is `76ef636f`.

**What you reported:** `mergeConfig` now copies `Sandbox.Enabled`, and provider-command output flows through it, so a command returning a valid provider plus `"sandbox":{"enabled":false}` could disable the sandbox.

Confirmed reproducible before the fix, in **both** directions. One detail worth having: only the *full-config* shape reaches it — the shorthand single-provider form is already safe, because `parseProviderCommandJSON` rebuilds a bare `FileConfig{ActiveProvider, Providers}` on that path and discards everything else.

**What changed:** `Sandbox.Enabled` is cleared from the provider-command config before merging, mirroring the carve-out `mergeProjectConfig` already applies for the same threat model. Done at the call site rather than inside `mergeConfig`, so the trusted global-config merge keeps honouring the setting — that distinction is why fixing it in the shared helper would have been wrong.

`TestResolveSandboxEnabledIgnoredFromProviderCommand` covers both directions and also asserts the provider itself still applies, so the guard can't be satisfied by dropping the command wholesale. Mutation-tested: removing the one line fails it. Global-config disable, CLI override, and the project-config carve-out all still pass. CodeRabbit has approved this head; all checks green.

anandh8x (approved):
Approved on the latest head. Provider-command output now has sandbox.enabled cleared before the shared config merge, while global config and CLI overrides retain the intended behavior. Regression coverage and focused tests pass.

Vasanthdev2004 (approved):
Re-approving on 76ef636f. My earlier approve was auto-dismissed by this push, so I re-checked the whole thing on the current head rather than standing on the old verdict.

The new commit closes a real hole. `mergeConfig` copies `Sandbox.Enabled`, and it is shared by two callers: the trusted global-config merge and the provider-command merge. Since `LoadProviderCommand` runs an arbitrary executable and parses its full stdout into a `FileConfig`, a command returning a valid provider plus `{"sandbox":{"enabled":false}}` could have turned off the sandbox meant to constrain it, despite the setting being documented as global-config/CLI only. Clearing `commandConfig.Sandbox.Enabled = nil` at the call site (not inside the shared helper, which the global-config merge still needs) is the right fix, and it mirrors the existing `mergeProjectConfig` carve-out.

Verified on 76ef636f:

- `mergeConfig` only assigns `Enabled` when the source is non-nil (resolver.go:229), so nil-ing it drops just the toggle while the provider itself still merges.
- Full posture is green: global config CAN disable, project config CANNOT, provider command CANNOT (both directions), CLI override CAN. Ran the four `TestResolveSandbox*` cases plus the CLI policy test locally, all pass.
- The new test also asserts the provider is still applied, so the guard cannot be satisfied by silently dropping the whole command config. Good touch.

One non-blocking heads-up, separate from this PR: the sibling sandbox fields (Network, AdditionalWriteRoots) still flow through the provider-command merge. That only matters if you later decide a provider command should not be able to widen write roots or open network either. Out of scope here; the `enabled` toggle was the documented contract, and that is what this closes.

LGTM.

PR #749, opened 2026-07-19
fix(agent): stop write_stdin session_id probing thrash (#702)

### LINKED ISSUE #702: `write_stdin` hallucinated `session_id` loop: model retries unknown IDs (1…N) until guardrail halt
### Version / branch / commit

zero dev (75a78e715bf23) main

### OS and environment

linux / tmux / go 1.26.4

### Steps to reproduce

1. Start a run where the model is pushed to implement a multi-file change (plan → implement).
2. Observe (or force) tool calls of the form:

   ```json
   {"name": "write_stdin", "arguments": {"session_id": 1, "chars": "..."}}
   ```

   without a preceding live `exec_command` session.
3. Confirm:

   - Error: `Unknown exec session_id 1.`
   - Subsequent calls with `2`, `3`, … continue failing.
   - Same-error stop may be delayed until the numeric ID stabilizes.

Unit-level coverage already exists for unknown IDs (`internal/tools/exec_command_test.go`). What’s missing is:

- guardrail signature stability across changing IDs, and/or
- stronger model-facing error text / tests that the stop fires within a small bound when only the id digits change.

### Expected behavior

1. `write_stdin` is only used with a `session_id` previously returned by a still-running `exec_command`.
2. Real exec session IDs start at **1000** (not 1).
3. On unknown/missing sessions, the model should stop and choose a different approach (e.g. start `exec_command`, or use `write_file` / `edit_file` / `apply_patch` for file edits) rather than invent sequential IDs.
4. The repeated-failure guardrail should stop this thrash early even when the numeric ID changes each attempt.

### Actual behavior

Example pattern:

```text
I have enough context; expanding the sketch into a full implementation plan.
• Sent input
Error: Unknown exec session_id 1.
• Sent input
Error: Unknown exec session_id 2.
• Sent input
Error: Unknown exec session_id 3.
…
• Sent input
Error: Unknown exec session_id 7.
```

Later: retries continue (reported up to ~id 150, then restart / halt), ending with the 6× same-error stop on `write_stdin`.

- Wastes turns/tokens and user time.
- Masks real blockers (e.g. missing write tools, permission mode, plan-only constraints).
- User-visible failure mode looks like broken patch/write tooling.
- Guardrail eventually stops the run, but often only after many useless tool calls.

Vasanthdev2004:
Closes #702.

## The bug

When the model calls `write_stdin` with no live exec session, it tends to probe sequential `session_id`s (1, 2, 3, …). Each one fails, and the run grinds on for a long time before anything stops it.

The reason it doesn't stop: the repeated-failure guard keys on a normalized, 80-character-truncated signature of the tool error (`errorSignature` in `internal/agent/guardrails.go`). The old message was `Error: Unknown exec session_id %d.`, so the varying id landed inside that 80-char window. Every probe produced a different signature, the same-error streak reset each time, and the 6x-in-a-row halt (`toolFailureStopAt`) never fired. It only tripped if the model repeated one identical id six times, which a probing model never does.

## The fix

- Rewrite the unknown-session error so the stable recovery guidance leads and the id trails past the signature window. Probing distinct ids now collapses to a single signature and trips the halt, and the model is told how to recover: start a process with `exec_command`, or use the file-edit tools. The id is still in the message (just past position 80), so the model still sees which id failed.
- Pin `session_id`'s schema `Minimum` to 1 so the advertised schema matches the runtime floor (`intArg` already rejects `< 1`); before this the schema declared no minimum while the runtime required one.
- `system_prompt.md`: state that `session_id` is only ever an id returned by a still-running `exec_command` and must never be guessed or probed.

## Why this is safe

The whole fix hinges on one invariant: the error's normalized signature must not vary with the id. That is pinned by a test that feeds `UnknownExecSessionError(1)` and `UnknownExecSessionError(999999)` through the real `errorSignature` and asserts equality, so a future edit that reintroduces the id early breaks the test. A second test proves the behavioral consequence end to end: probing a different unknown id each turn now halts at the guard's stop threshold, where before it never would. The schema floor and the recovery text are pinned too, and the two existing tests that asserted the old wording are updated.

No runtime behavior changes beyond the halt firing where it should: the `Minimum` is advisory schema metadata (validation is still done independently by `intArg`), and the error only changes wording.

## Scope note

The issue's fourth criterion (the TUI showing "Sent input" on a *failed* `write_stdin`) is a separate cosmetic concern in a shared rendering path used at several call sites. I left it out to keep this diff focused on the agent-behavior bug; happy to follow up with it as its own small change if you'd like it in scope.

## Testing

`go build ./...`, `go vet`, `gofmt` clean; `internal/tools` and `internal/agent` suites pass. New tests: signature id-invariance, distinct-id probing halts at the stop threshold, schema minimum matches the runtime floor, and the recovery text is present.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `write_stdin` errors for unknown sessions with clearer recovery guidance and the affected session ID.
  * Prevented invalid session IDs below the supported minimum.
  * Ensured repeated attempts to access unknown sessions trigger the appropriate failure halt.

* **Documentation**
  * Clarified how to start or resume interactive command sessions and advised against guessing session IDs.

### DIFF
--- a/internal/agent/guardrails_test.go
+++ b/internal/agent/guardrails_test.go
@@ -169,6 +169,37 @@ func TestRunResetsEmptyTurnCounterOnToolCall(t *testing.T) {
 	}
 }
 
+// The whole point of #702's fix: the unknown-session error's normalized
+// signature must not vary with the session id, so a model probing ids 1, 2,
+// 3, … produces one repeated signature the failure guard can count. If the id
+// leaked into the first 80 normalized chars, each probe would reset the streak
+// and the halt would never fire.
+func TestUnknownExecSessionErrorSignatureIsIDInvariant(t *testing.T) {
+	a := errorSignature(tools.UnknownExecSessionError(1))
+	b := errorSignature(tools.UnknownExecSessionError(999999))
+	if a != b {
+		t.Fatalf("unknown-session signature varies with id:\n  %q\n  %q", a, b)
+	}
+}
+
+// End to end: with an id-invariant signature, probing a different unknown id
+// each turn now trips the repeated-failure halt at toolFailureStopAt, where
+// before the fix it never would.
+func TestUnknownExecSessionProbingTripsFailureHalt(t *testing.T) {
+	var state guardState
+	var stoppedAt int
+	for i := 1; i <= toolFailureStopAt; i++ {
+		out := state.observeToolResult(tools.WriteStdinToolName, true, tools.UnknownExecSessionError(i))
+		if out.Stop {
+			stoppedAt = i
+			break
+		}
+	}
+	if stoppedAt != toolFailureStopAt {
+		t.Fatalf("probing distinct unknown ids stopped at %d, want %d", stoppedAt, toolFailureStopAt)
+	}
+}
+
 func TestGuardStateResetsToolOnlyStreakOnEmptyNonToolTurn(t *testing.T) {
 	var state guardState
 	toolOnly := zeroruntime.CollectedStream{
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -100,6 +100,9 @@ work.
 - Use exec_command with `tty: true` for interactive terminal-style commands that
   need stdin beyond Ctrl-C. `/ps` and `/stop` are user-facing TUI commands; when
   you need to clean up a running foreground command yourself, use write_stdin.
+- write_stdin's session_id is only ever an id returned by a still-running
+  exec_command; never guess or probe ids. If you have no such session, start one
+  with exec_command, or use write_file/edit_file/apply_patch for file changes.
 - write_stdin with empty input polls an existing exec_command session, and
   `\u0003` interrupts it. Sending other stdin bytes may require approval because
   it can drive the running process beyond the original command. Non-tty sessions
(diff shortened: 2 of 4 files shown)


anandh8x (approved):
No actionable findings. The unknown-session error now produces an ID-invariant guardrail signature, repeated probing halts at the configured threshold, and the schema plus model-facing guidance consistently require a live session returned by exec_command. CI is green. The deferred TUI `Sent input` label for failed calls should remain tracked as a separate cosmetic follow-up.

gnanam1990 (approved):
**APPROVE.** Core fix verified correct: `errorSignature` (guardrails.go:339) lowercases + collapses whitespace + truncates to 80 chars, making it id-invariant, so the model's ascending unknown-`session_id` probes now accumulate to the repeated-failure streak and halt (previously each id was a distinct signature, so the streak never tripped). Confirmed: legit `write_stdin` to a real session is untouched (only the `!ok` no-live-session branch changes), schema/runtime floor agree (min 1), and the corrective hint now fires at count 2.

Finding (not blocking):
- **[Minor]** The issue's acceptance criterion #4 is unaddressed: a *failed* `write_stdin` still renders as **"Sent input"** (`rendering.go:1676-1677`, `toolCardActionLabel` has no success/failure branch, and rendering.go isn't in the diff). Pre-existing UX gap — either render a distinct "Failed to send input" label on error, or note in the PR that criterion #4 is deferred so #702 isn't closed as fully resolved.

PR #750, opened 2026-07-19
fix(providers): retry provably pre-send transport failures (#447)

### LINKED ISSUE #447: Auto-retry transient network errors (TLS timeout, conn reset) instead of requiring manual "continue"
### Context

Building on ZERO v0.1.0 (macOS, OpenRouter). On a flaky connection a model call failed with:

`upstream unreachable: the model server could not connect to openrouter.ai (TLS handshake timeout). The request never reached the model — this is a network failure.`

ZERO did the right things: clear diagnosis, and it **preserved the edits already applied**. But it then stopped and I had to type `continue` to resume. On a flaky link that means babysitting every blip.

### Request

Auto-retry **transport-level** failures — TLS handshake timeout, connection reset/refused, DNS resolution failure — with exponential backoff + jitter (e.g. 3–5 attempts) before surfacing the error to the user. ZERO already classifies these as "the request never reached the model," so the signal to retry safely is right there.

### Important nuance (why this is safe)

Only retry when the failure is provably **pre-send** (handshake/DNS/connect) — i.e. the request never reached the model. **Never** auto-retry after a request may have partially executed a side effect (a tool call that already ran, a stream that started), where a retry could double-execute. ZERO's own error text already draws this line ("the request never reached the model").

### Impact

On flaky connections the agent self-heals instead of stalling and requiring a manual `continue` on every network hiccup.

### Env

ZERO 0.1.0 · macOS (x64) · provider: openrouter

Vasanthdev2004:
Closes #447.

## The ask, and the stance it touches

A flaky link that fails the streaming completion POST at connect time (TLS handshake timeout, connection refused, DNS failure) stops the turn and forces the user to type `continue`. The reporter asked for auto-retry of transport failures, with the right safety instinct: retry only failures that provably never reached the model.

The shared retry layer deliberately does NOT retry transport errors, and the reasoning is sound and documented: a completion POST is non-idempotent, so if the request may have reached the server, a replay could duplicate billable work.

**This PR is opening that decision for your sign-off, not treating it as settled.** It changes a stance you staked out on purpose, so it should be your call.

## Why a narrow carve-out is consistent with that stance

The "may have reached the server" concern does not apply to a *provably pre-send* failure, where no request bytes ever left this host. There a replay cannot duplicate anything, which is exactly the logic that already makes 429 and 503 safe to retry (the server told us it did not accept the request).

So `SendWithRetry` now retries, with the existing backoff and attempt budget, only that narrow class:
- DNS resolution failure (`net.DNSError`, matched by type through any wrapping)
- a refused or unreachable dial (`connection refused`, `network is unreachable`, `no route to host`)
- a TLS handshake timeout (the handshake completes before any HTTP request bytes are written)

Everything ambiguous or post-send stays unretried, and those exclusions are checked **first** so a message carrying both an excluded and an included marker can never be misread as pre-send: connection reset, broken pipe, EOF, a bare `i/o timeout` (which also covers post-send read timeouts), context deadline.

This deliberately **narrows the issue's ask** (it listed connection reset, which stays unretried) down to the provably-safe subset. The existing reset-stays-unretried test stays green.

## Scope and placement

Every provider send path routes the completion POST through `SendWithAuthRetry` then `SendWithRetry`, so this one change covers them uniformly, and it fires before the in-band stream error that stops the turn is ever emitted. The 429/503 path, the auth-retry path, and streaming are untouched.

## A note on the classification

Pre-send detection uses `errors.As` for the DNS type, `errors.Is` for EOF, and narrow string markers for the dial/TLS cases. String markers are the pragmatic choice here (they are stable and portable, unlike platform-specific syscall errnos). The one theoretical way a post-send error could carry an inclusion marker is a `url.Error` whose URL contains the phrase, which cannot happen: every marker contains a space, and the HTTP client rejects a space in the host and percent-encodes it in the path.

## Testing

`go build`, `go vet`, `gofmt` clean; full `internal/providers/...` suite passes. New tests cover retry-on-pre-send (refused dial, network unreachable, TLS handshake timeout, DNS), no-retry-on-ambiguous (i/o timeout, broken pipe, EOF), and the predicate classification table including the exclusion-wins and host-named-eof cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior to only reattempt safe, provably pre-send failures (such as DNS/dial/TLS handshake timeouts) and to stop immediately on context cancellation.
  * Added a short, bounded pre-send backoff with independent retry budgeting from HTTP status retries.
  * Refined redirect and retry interactions so retryable statuses behind redirects are still handled, while ambiguous post-send failures are not retried.

* **Tests**
  * Expanded coverage for pre-send vs ambiguous transport classification, redirect/budget independence, and precise pre-send backoff timing, including platform-specific error cases.

### DIFF
--- a/internal/providers/providerio/auth.go
+++ b/internal/providers/providerio/auth.go
@@ -25,7 +25,9 @@ type TokenResolver func(ctx context.Context, forceRefresh bool) (header string, 
 //
 // base carries the API-key auth config (key + default header/scheme). setExtra
 // (optional) sets the provider's non-auth headers on every attempt. Transient
-// retries (429/503/529) are handled by the inner SendWithRetry.
+// retries are handled by the inner SendWithRetry: 429/503/529 on their own
+// budget, and provably pre-send transport failures (refused or timed-out
+// connect, DNS, TLS handshake) on a separate, shorter one.
 func SendWithAuthRetry(
 	ctx context.Context,
 	client *http.Client,
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/Gitlawb/zero/internal/trace"
@@ -31,6 +32,8 @@ import (
 // non-idempotent and may already have reached and been processed by the server,
 // so replaying it could duplicate (billable) work. Only the INITIAL request is
 // ever in scope; once the response body starts streaming it is never re-issued.
+// Pre-send classification is by syscall errno (portable across POSIX and
+// Windows), so a refused/unreachable dial retries the same on every platform.
 
 const defaultMaxRetryAttempts = 6
 
@@ -134,12 +137,19 @@ func SendWithRetry(
 // DNS resolution failure, a refused or unreachable dial, and a TLS handshake
 // timeout (the handshake completes before any HTTP request bytes are written).
 //
+// Classification is by syscall errno wherever possible, via errors.Is, which is
+// portable across platforms: Go maps ECONNREFUSED / ENETUNREACH / EHOSTUNREACH
+// / ECONNRESET to the host's real values, including the WSA* codes on Windows,
+// whose human-readable dial wording differs entirely from POSIX. String markers
+// are kept only as a fallback for errors that arrive already flattened to text
+// (no errno in the chain).
+//
 // Ambiguous or post-send failures are deliberately excluded and checked FIRST,
-// so a message that happens to contain both an excluded and an included marker
-// is never treated as pre-send: connection reset and broken pipe can follow a
-// request that was already sent, and a bare "i/o timeout" covers read timeouts
-// after send as well as dial timeouts — none of these prove the request was not
-// received, so a non-idempotent POST must not be replayed on them.
+// so an error that is both is never treated as pre-send: a reset or broken pipe
+// can follow a request that was already sent, and a bare "i/o timeout" covers
+// read timeouts after send as well as dial timeouts, so none of these prove the
+// request was not received and a non-idempotent POST must not be replayed on
+// them.
 func isPreSendTransportError(err error) bool {
 	if err == nil {
 		return false
@@ -150,18 +160,24 @@ func isPreSendTransportError(err error) bool {
 	if errors.As(err, &dnsErr) {
 		return true
 	}
-	// A connection closed mid-exchange (EOF) may well have delivered the request
-	// first, so it is post-send. Matched by identity, not substring, so a
-	// hostname containing "eof" can't be misread.
-	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+	// Post-send / ambiguous failures, excluded first. EOF and reset are matched
+	// by identity so wording (or a hostname containing "eof") can't fool them.
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.ECONNRESET) {
 		return false
 	}
 	msg := strings.ToLower(err.Error())
-	// Exclusions win: never classify an ambiguous/post-send failure as pre-send.
 	if strings.Contains(msg, "connection reset") ||
 		strings.Contains(msg, "broken pipe") ||
 		strings.Contains(msg, "i/o timeout") {
 		return false
+	}
+	// Provably pre-send: the connect never completed, so no request bytes left
+	// this host. Errno match is authoritative and platform-independent; the
+	// string markers only catch a POSIX error already flattened to text.
+	if errors.Is(err, syscall.ECONNREFUSED) ||
+		errors.Is(err, syscall.ENETUNREACH) ||
+		errors.Is(err, syscall.EHOSTUNREACH) {
+		return true
 	}
 	switch {
 	case strings.Contains(msg, "connection refused"):
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -3,6 +3,9 @@ package providerio
 import (
 	"bytes"
 	"context"
+	"errors"
+	"io"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
@@ -17,13 +20,17 @@ import (
 // (previously only the OpenAI provider retried; Anthropic and Gemini surfaced the
 // first failure).
 //
-// What is retried: ONLY 429 (rate limit) and 503 (service unavailable) — the
-// statuses where the server explicitly did NOT accept the request, so replaying
-// it cannot duplicate work. Other 5xx (500/502/504) and transport/network errors
-// are NOT retried: a completion POST is non-idempotent and may have reached and
-// been processed by the server, so replaying it could duplicate
-// (billable) work. Only the INITIAL request is ever in scope; once the response
-// body starts streaming it is never re-issued.
+// What is retried: 429 (rate limit) and 503 (service unavailable) — the statuses
+// where the server explicitly did NOT accept the request — plus PROVABLY
+// pre-send transport failures (DNS resolution, a refused/unreachable dial, a TLS
+// handshake timeout) where no request bytes ever left this host. In every one of
+// those cases we KNOW the server did not receive and process the request, so a
+// replay cannot duplicate work. Other 5xx (500/502/504) and every ambiguous or
+// post-send transport error (connection reset, broken pipe, EOF, a generic i/o
+// timeout, context deadline) are NOT retried: a completion POST is
+// non-idempotent and may already have reached and been processed by the server,
+// so replaying it could duplicate (billable) work. Only the INITIAL request is
+// ever in scope; once the response body starts streaming it is never re-issued.
 
 const defaultMaxRetryAttempts = 6
 
@@ -73,13 +80,25 @@ func SendWithRetry(
 		response, err := client.Do(request)
 		connectSpan.End()
 		if err != nil {
-			// A transport failure on a POST does NOT mean the server didn't receive
-			// it — the request may have arrived and be generating a (billable,
-			// non-idempotent) completion while only the response/connection failed.
-			// Replaying it could duplicate that work, so surface the error instead
-			// of retrying. Only 429/503 responses (below), where we KNOW the request
-			// was not accepted, are safe to retry.
+			// Context cancellation always surfaces as cancellation, never a retry.
 			if ctx.Err() != nil {
+				return nil, ctx.Err()
+			}
+			// A transport failure on a POST does NOT usually mean the server didn't
+			// receive it — the request may have arrived and be generating a
+			// (billable, non-idempotent) completion while only the response or
+			// connection failed. Replaying that could duplicate work, so it is
+			// surfaced immediately. The one safe exception is a PROVABLY pre-send
+			// failure (see isPreSendTransportError): the request never left this
+			// host, so a replay cannot duplicate anything — exactly like a 429/503
+			// where we KNOW the request was not accepted.
+			if isPreSendTransportError(err) && attempt < maxAttempts {
+				if r := trace.FromContext(ctx); r != nil {
+					r.Counter(trace.CounterRetryCount, 1)
+				}
+				if Backoff(ctx, attempt, 0) {
+					continue
+				}
 				return nil, ctx.Err()
 			}
 			return nil, err
@@ -107,6 +126,52 @@ func SendWithRetry(
 		}
 		return response, nil
 	}
+}
+
+// isPreSendTransportError reports whether a transport error PROVES no request
+// bytes reached the server, so replaying the request cannot duplicate a
+// billable completion. Only narrow, unambiguous pre-connect failures qualify:
+// DNS resolution failure, a refused or unreachable dial, and a TLS handshake
+// timeout (the handshake completes before any HTTP request bytes are written).
+//
+// Ambiguous or post-send failures are deliberately excluded and checked FIRST,
+// so a message that happens to contain both an excluded and an included marker
+// is never treated as pre-send: connection reset and broken pipe can follow a
+// request that was already sent, and a bare "i/o timeout" covers read timeouts
+// after send as well as dial timeouts — none of these prove the request was not
+// received, so a non-idempotent POST must not be replayed on them.
+func isPreSendTransportError(err error) bool {
+	if err == nil {
+		return false
+	}
+	// DNS resolution precedes any connection, so a lookup failure — including a
+	// DNS timeout — proves nothing was sent. errors.As unwraps url.Error/OpError.
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		return true
+	}
+	// A connection closed mid-exchange (EOF) may well have delivered the request
+	// first, so it is post-send. Matched by identity, not substring, so a
+	// hostname containing "eof" can't be misread.
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	// Exclusions win: never classify an ambiguous/post-send failure as pre-send.
+	if strings.Contains(msg, "connection reset") ||
+		strings.Contains(msg, "broken pipe") ||
+		strings.Contains(msg, "i/o timeout") {
+		return false
+	}
+	switch {
+	case strings.Contains(msg, "connection refused"):
+		return true
+	case strings.Contains(msg, "network is unreachable"), strings.Contains(msg, "no route to host"):
+		return true
+	case strings.Contains(msg, "tls handshake timeout"):
+		return true
+	}
+	return false
 }
 
 // ShouldRetryStatus reports whether an HTTP status is safe to retry for a
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -92,6 +92,26 @@ func SendWithRetry(
 	if maxAttempts <= 0 {
 		maxAttempts = defaultMaxRetryAttempts
 	}
+	if client == nil {
+		client = http.DefaultClient
+	}
+	// Redirects are NOT followed on this path. Otherwise client.Do could transmit
+	// the original POST to the first host, follow a 307/308 to a redirect target,
+	// and if the dial to that target then failed, return an Op=="dial" error that
+	// isPreSendTransportError would wrongly treat as pre-send and replay, re-billing
+	// a completion the first host already received. With redirects off, any
+	// dial/DNS/TLS error from Do can only come from the single initial request, so
+	// "no request bytes left this host" holds. A 3xx is returned as the response
+	// (ErrUseLastResponse) for the caller's existing non-2xx status handling.
+	noRedirectClient := *client
+	noRedirectClient.CheckRedirect = func(*http.Request, []*http.Request) error {
+		return http.ErrUseLastResponse
+	}
+	// preSendAttempts bounds the pre-send retries independently of the loop-wide
+	// attempt counter, which also counts 429/503 retries: sharing it let a couple
+	// of status retries exhaust the pre-send budget so the first dial/DNS/TLS blip
+	// after them got no retry at all.
+	preSendAttempts := 0
 	for attempt := 1; ; attempt++ {
 		request, err := http.NewRequestWithContext(ctx, method, url, bytes.NewReader(body))
 		if err != nil {
@@ -102,7 +122,7 @@ func SendWithRetry(
 		}
 
 		connectSpan := trace.FromContext(ctx).Span(trace.SpanProviderConnect)
-		response, err := client.Do(request)
+		response, err := noRedirectClient.Do(request)
 		connectSpan.End()
 		if err != nil {
 			// Context cancellation always surfaces as cancellation, never a retry.
@@ -120,14 +140,17 @@ func SendWithRetry(
 			// its own short bound and sub-second schedule (preSendMaxAttempts /
 			// preSendBackoff), not the seconds-long 429 schedule, so a permanent
 			// dial failure fails fast instead of stalling the agent for ~60s.
-			if isPreSendTransportError(err) && attempt < preSendMaxAttempts {
-				if r := trace.FromContext(ctx); r != nil {
-					r.Counter(trace.CounterRetryCount, 1)
+			if isPreSendTransportError(err) {
+				preSendAttempts++
+				if preSendAttempts < preSendMaxAttempts {
+					if r := trace.FromContext(ctx); r != nil {
+						r.Counter(trace.CounterRetryCount, 1)
+					}
+					if preSendBackoff(ctx, preSendAttempts) {
+						continue
+					}
+					return nil, ctx.Err()
 				}
-				if preSendBackoff(ctx, attempt) {
-					continue
-				}
-				return nil, ctx.Err()
 			}
 			return nil, err
 		}
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -207,20 +207,21 @@ func isPreSendTransportError(err error) bool {
 	}
 	// A refused/unreachable connection is provably pre-send ONLY when it was
 	// raised during the DIAL (connection-establishment) phase. The same errnos
-	// (ECONNREFUSED/ENETUNREACH/EHOSTUNREACH, and their POSIX wording) are also
-	// raised by the kernel on an ESTABLISHED connection — when a route drops or an
-	// ICMP unreachable arrives mid-generation — which is post-send. errors.Is
-	// matches an errno anywhere in the chain regardless of the operation, so we
-	// first require an Op=="dial" *net.OpError: that makes "no request bytes left
-	// this host" structural, so a post-send read/write carrying one of these can
-	// never replay a non-idempotent POST. Errno match is authoritative and
-	// platform-independent (Go maps the WSA* codes on Windows too); the string
-	// markers only catch a dial error already flattened past its errno.
+	// are also raised by the kernel on an ESTABLISHED connection — when a route
+	// drops or an ICMP unreachable arrives mid-generation — which is post-send.
+	// errors.Is matches an errno anywhere in the chain regardless of the
+	// operation, so we first require an Op=="dial" *net.OpError: that makes "no
+	// request bytes left this host" structural, so a post-send read/write carrying
+	// one of these can never replay a non-idempotent POST. dialPreSendErrnos is
+	// platform-specific: on Windows a refused dial carries the raw WSA* errno
+	// (WSAECONNREFUSED etc.), which does NOT satisfy errors.Is against the POSIX
+	// syscall.ECONNREFUSED, so the Windows list adds those codes. The string
+	// markers only catch a POSIX dial error already flattened past its errno.
 	if dialOpError(err) != nil {
-		if errors.Is(err, syscall.ECONNREFUSED) ||
-			errors.Is(err, syscall.ENETUNREACH) ||
-			errors.Is(err, syscall.EHOSTUNREACH) {
-			return true
+		for _, errno := range dialPreSendErrnos {
+			if errors.Is(err, errno) {
+				return true
+			}
 		}
 		switch {
 		case strings.Contains(msg, "connection refused"),
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/Gitlawb/zero/internal/trace"
@@ -196,7 +195,7 @@ func isPreSendTransportError(err error) bool {
 	}
 	// Post-send / ambiguous failures, excluded first. EOF and reset are matched
 	// by identity so wording (or a hostname containing "eof") can't fool them.
-	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, syscall.ECONNRESET) {
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) || isConnResetErrno(err) {
 		return false
 	}
 	msg := strings.ToLower(err.Error())
--- a/internal/providers/providerio/retry.go
+++ b/internal/providers/providerio/retry.go
@@ -33,7 +33,10 @@ import (
 // so replaying it could duplicate (billable) work. Only the INITIAL request is
 // ever in scope; once the response body starts streaming it is never re-issued.
 // Pre-send classification is by syscall errno (portable across POSIX and
-// Windows), so a refused/unreachable dial retries the same on every platform.
+// Windows), scoped to an Op=="dial" *net.OpError so the same errno raised on a
+// post-send read/write can't be misread as pre-send. NXDOMAIN is treated as
+// permanent (not retried), and the pre-send path fails fast on its own
+// sub-second schedule rather than the seconds-long 429 backoff.
 
 const defaultMaxRetryAttempts = 6
 
@@ -47,12 +50,32 @@ const maxBackoff = 30 * time.Second
 // schedule is 2s, 4s, 8s, 16s, then maxBackoff. A var so tests can shrink it.
 var retryBackoffBase = 2 * time.Second
 
+// preSendMaxAttempts bounds retries of a provably pre-send transport failure.
+// Unlike a 429 (whose window is measured in seconds), a refused/unreachable dial
+// either recovers within a few hundred ms or is a deterministic misconfiguration
+// (wrong host/port, dead local daemon) that will never recover — so a short,
+// low-count schedule fails fast instead of stalling the agent ~60s on the
+// rate-limit schedule. This bounds the pre-send path independently of the caller's
+// maxAttempts, which is tuned for 429/503.
+const preSendMaxAttempts = 3
+
+// preSendBackoffBase is the first wait before replaying a pre-send transport
+// failure. Modeled on the mid-stream reconnect base (internal/agent/reconnect.go),
+// it is sub-second so a transient dial blip recovers quickly and a permanent one
+// fails fast. A var so tests can shrink it.
+var preSendBackoffBase = 500 * time.Millisecond
+
 // SendWithRetry issues an HTTP request, retrying ONLY the safe-to-replay server
 // responses (429 and 503, see ShouldRetryStatus) up to maxAttempts — backing off
 // between tries and honoring a server Retry-After header and context
-// cancellation. Other 5xx and transport/network errors are returned immediately,
-// never replayed (see the package note). The request is rebuilt from body each
-// attempt; setHeader (if non-nil) sets headers on every attempt.
+// cancellation. The one transport-error exception is a PROVABLY pre-send failure
+// (see isPreSendTransportError): a refused/unreachable dial or DNS failure where
+// no request bytes left this host, retried on its own short schedule
+// (preSendMaxAttempts / preSendBackoff). Every other 5xx and every ambiguous or
+// post-send transport error is returned immediately, never replayed, because a
+// non-idempotent completion POST may already have been received. The request is
+// rebuilt from body each attempt; setHeader (if non-nil) sets headers on every
+// attempt.
 //
 // It returns the final *http.Response (which the caller inspects for a non-2xx
 // status, exactly as before) or an error for a network failure / context
@@ -94,12 +117,15 @@ func SendWithRetry(
 			// surfaced immediately. The one safe exception is a PROVABLY pre-send
 			// failure (see isPreSendTransportError): the request never left this
 			// host, so a replay cannot duplicate anything — exactly like a 429/503
-			// where we KNOW the request was not accepted.
-			if isPreSendTransportError(err) && attempt < maxAttempts {
+			// where we KNOW the request was not accepted. The pre-send path uses
+			// its own short bound and sub-second schedule (preSendMaxAttempts /
+			// preSendBackoff), not the seconds-long 429 schedule, so a permanent
+			// dial failure fails fast instead of stalling the agent for ~60s.
+			if isPreSendTransportError(err) && attempt < preSendMaxAttempts {
 				if r := trace.FromContext(ctx); r != nil {
 					r.Counter(trace.CounterRetryCount, 1)
 				}
-				if Backoff(ctx, attempt, 0) {
+				if preSendBackoff(ctx, attempt) {
 					continue
 				}
 				return nil, ctx.Err()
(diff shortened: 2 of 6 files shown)


Vasanthdev2004:
Good catch, fixed in 4c54a8b6. Windows dial wording is completely different ("No connection could be made because the target machine actively refused it"), so the POSIX string markers would have missed it.

Rather than add Windows strings, the predicate now classifies by syscall errno via errors.Is: ECONNREFUSED / ENETUNREACH / EHOSTUNREACH for the pre-send inclusions and ECONNRESET for the post-send exclusion. Go maps those constants to the host's real values (the WSA codes on Windows), so the classification is identical on every platform, and it's authoritative where the strings were only heuristic. The string markers stay as a fallback for errors that arrive already flattened to text with no errno in the chain.

Tests build the real production error shape (net.OpError wrapping os.SyscallError wrapping the errno), which is the portable way to exercise the Windows path: the same errno assertion holds on Linux/macOS/Windows even though .Error() reads differently on each. Added errno cases to the predicate table (refused, network-unreachable, host-unreachable, and reset-is-post-send) plus an end-to-end SendWithRetry case for an errno-wrapped refused dial. Cross-compiled and vetted for linux and darwin so the platform-specific constants are confirmed to build everywhere.

gnanam1990 (approved):
**APPROVE.** The safety-critical property is airtight: `isPreSendTransportError` evaluates the post-send/ambiguous exclusions (EOF, ECONNRESET, broken pipe, i/o timeout, DeadlineExceeded) BEFORE any inclusion, so a genuinely post-send failure can never be classified pre-send and never replays a billable POST (verified: exactly 1 call on ambiguous errors). Request body is rebuilt via bytes.NewReader per attempt; context cancellation is checked first; ECONNREFUSED is connect-only; retries are bounded; no bad interaction with the auth-retry path.

Findings (none blocking):
- **[Minor]** A connect/**dial `i/o timeout`** is NOT retried — Go surfaces it as `*net.OpError{Op:"dial"}` whose text contains `i/o timeout`, caught by the exclusion at retry.go:169. That misses part of #447's own motivating case (intermittent connectivity → connect timeout). Safe fix: before the string exclusion, admit `*net.OpError` when `Op=="dial"/"connect" && Timeout()` — unambiguously pre-send, without loosening the post-send read-timeout exclusion.
- **[Nit]** Backoff has no jitter (issue mentioned it) — a non-issue for a single-client CLI; optional.
- **[Nit]** `ENETUNREACH`/`EHOSTUNREACH` errno admits (retry.go:176) aren't scoped to `Op=="dial"`. Residual risk is very low and mitigated (established-connection failures surface as read reset/EOF/timeout, excluded first), but scoping to dial/connect would make the no-bytes-sent guarantee structural rather than argued.

gnanam1990 (changes requested):
## Request changes

The safety reasoning here is genuinely careful, but the central claim doesn't hold: `isPreSendTransportError` matches the pre-send errnos anywhere in the error chain regardless of which socket operation raised them, so a network-unreachable error on a `read`/`write` *after* the request was fully sent is classified as "provably pre-send" and the non-idempotent completion POST is replayed.

Build/vet/tests are clean — see the bottom of this review.

---

### [Major] Pre-send errno classification is not scoped to the dial, so a post-send failure replays a billable POST
`internal/providers/providerio/retry.go:177-181`

```go
if errors.Is(err, syscall.ECONNREFUSED) ||
	errors.Is(err, syscall.ENETUNREACH) ||
	errors.Is(err, syscall.EHOSTUNREACH) {
	return true
}
```

`errors.Is` matches the errno wherever it appears in the chain — it says nothing about *which* operation produced it. These errnos are not dial-only: the kernel reports them on an established connection too, when a route goes away or an ICMP unreachable arrives.

Concrete scenario: a user is mid-generation against `api.anthropic.com`, the request has been fully written, the model is producing a billable completion, and Wi-Fi drops or a VPN toggles. Linux/macOS surface that on the pending read as `*url.Error -> *net.OpError{Op:"read"} -> *os.SyscallError -> EHOSTUNREACH`. Verified in this worktree:

- `&net.OpError{Op: "read", Err: os.NewSyscallError("read", syscall.EHOSTUNREACH)}` → `isPreSendTransportError` returns **true**
- same for `Op: "write"` with `ENETUNREACH`
- same via the string fallback at `retry.go:185` for already-flattened text like `write tcp: write: network is unreachable`

From there `SendWithRetry` (`retry.go:98`) replays the POST up to 5 more times — all providers pass `maxAttempts=0` → `defaultMaxRetryAttempts=6` (`anthropic/provider.go:196`, `openai/provider.go:196`, `gemini/provider.go:151`, `openai/codex_responses.go:354`) — producing up to 6 duplicate billable completions for one user turn.

Worth flagging that the existing table case gives false assurance:

```go
{"errno reset is post-send", wrapDialErrno("read", syscall.ECONNRESET), false},  // retry_test.go:133
```

It passes because `ECONNRESET` is in the exclusion list at `retry.go:165`, not because the operation is inspected. No case exercises a non-dial op carrying a pre-send-classified errno.

**Fix:** gate on the dial phase before the errno checks — unwrap to `*net.OpError` and require `opErr.Op == "dial"` (Go sets this for connect failures) around `retry.go:177-181`, and apply the same gate to the `"connection refused"` / `"network is unreachable"` / `"no route to host"` string fallbacks at `retry.go:183-186`. Then add table cases for `wrapDialErrno("write", syscall.ENETUNREACH)` and `wrapDialErrno("read", syscall.EHOSTUNREACH)` expecting `false`.

---

### [Major] Permanent failures now hang the agent ~60s silently, on a 429-tuned backoff
`internal/providers/providerio/retry.go:98`, schedule at `retry.go:44-48`

Deterministic, never-going-to-succeed failures go through the full retry loop on a schedule explicitly tuned for rate-limit windows:

- NXDOMAIN — `net.DNSError` returns `true` unconditionally at `retry.go:160`, before `IsNotFound` is ever consulted
- A refused dial to a local endpoint

Verified in this worktree with `maxAttempts=0`: the transport is invoked exactly 6 times, and `backoffWait` for attempts 1..5 sums to 2+4+8+16+30 = **60s** (schedule pinned by `TestBackoffWaitSchedule`, `retry_test.go:218`).

Scenario: user points the openai provider at `http://127.0.0.1:11434` for a local Ollama/LM-Studio daemon that isn't running, or mistypes a base URL. Every attempt fails with ECONNREFUSED / `no such host`. `SendWithRetry` emits no notifier callback, and the agent-layer `reconnectNoticeFor` never fires for these providers (`StreamCompletion` returns nil synchronously and delivers the error as a `StreamEventError`), so the TUI just sits for a full minute per turn before reporting something that was known on the first millisecond. Previously this surfaced instantly.

**Fix:** two parts.
1. Don't retry `net.DNSError` when `IsNotFound` is true — NXDOMAIN is authoritative and deterministic. Only `IsTimeout` / `IsTemporary` lookups are worth replaying. (`retry.go:159-162`)
2. Give the pre-send transport path its own bound and schedule instead of reusing the rate-limit one — cap around 2-3 attempts starting from a sub-second base. The mid-stream reconnect layer already uses 500ms (`internal/agent/reconnect.go:34`) and is a reasonable model.

---

### [Minor] `SendWithRetry`'s own doc comment now states the opposite of what it does
`internal/providers/providerio/retry.go:53-54`

The package note at `retry.go:24-36` was correctly updated for the new behavior, but the function doc immediately below it wasn't:

> Other 5xx and transport/network errors are returned immediately, never replayed (see the package note).

That contradicts `retry.go:98`. This is the doc that shows on hover/godoc at every call site in anthropic, openai, gemini and codex_responses — a maintainer reads it, concludes no transport error is ever replayed, and reasons about idempotency on exactly the assumption this PR invalidates. In a codebase where these comments are the primary record of the retry-safety contract, that's a live hazard rather than a cosmetic nit.

**Fix:** update `retry.go:50-55` to mention the pre-send exception and point at `isPreSendTransportError`, matching the package note.

---

### What's well done

Genuinely above average for this class of change, and worth saying:

- Ambiguous/post-send errors are checked **before** the inclusion set (`retry.go:163-173`), so an error matching both is never treated as pre-send. That ordering is the right call and the "exclusion wins over inclusion" test case covers it directly.
- EOF and ECONNRESET matched by identity via `errors.Is` rather than by wording (`retry.go:165`) — a hostname containing "eof" can't fool it.
- `ctx.Err()` checked before any retry decision (`retry.go:87`), so cancellation never becomes a retry.
- Deliberately leaving http2 GOAWAY/stream errors, `net/http: timeout awaiting response headers`, and `context deadline exceeded` unmatched — all correct omissions.
- Typed `net.DNSError`/errno matching preferred over strings, with strings only as a fallback for already-flattened errors: the right hierarchy, and the Windows WSA* reasoning in the comment is sound.
- `TestSendWithRetryDoesNotReplayAmbiguousTransportErrors` (`retry_test.go:99`) is a real guard — it asserts exactly one call, so a regression that started replaying resets or broken pipes would fail it.
- `shrinkBackoff` correctly scoped with `t.Cleanup`.

The dial-scoping fix should be small; the structure it slots into is already right.

---

### Build & tests

All clean on the PR head (go1.26.5, darwin/arm64), no failures of any kind:

- `gofmt -l ./internal/providers/providerio/` — clean
- `go build ./...` — exit 0
- `go vet ./internal/providers/...` — exit 0, no diagnostics
- `go test -count=1 ./internal/providers/providerio/...` — ok 0.854s, all subtests pass
- `go test -count=1 -race ./internal/providers/...` — exit 0 across all 5 packages, no race reports

No merge-base comparison was needed since nothing failed. The sandbox-related pre-existing failures in `internal/agent` and `internal/cli` are out of scope — this PR touches neither package.

**Scope note:** the stated merge base `fbf8598` is not actually this branch's merge base (`git merge-base fbf8598 HEAD` = `ce4a996`). Diffing `fbf8598..HEAD` shows a misleading 44-file / -2225-line stat because it renders main's later commits as deletions. The real PR content vs `ce4a996` is 2 files, +207/-13 — `retry.go` and `retry_test.go` — which matches the supplied diff.

Vasanthdev2004:
Both [Major] findings fixed in `8b4d5e5`, thanks gnanam. You were right that the safety argument had a hole.

**[Major] dial-scoping.** You nailed it: `errors.Is` matches an errno anywhere in the chain regardless of which socket op raised it, so a post-send `read`/`write` carrying `EHOSTUNREACH`/`ENETUNREACH` was classified pre-send and could replay the POST. `isPreSendTransportError` now requires an `Op=="dial"` `*net.OpError` (via a small `dialOpError` helper) before admitting those errnos or their POSIX wording, so "no request bytes were sent" is structural instead of argued. Your exact repro (`&net.OpError{Op:"read", ...EHOSTUNREACH}`) now returns false, and there are explicit table cases for `wrapDialErrno("read", EHOSTUNREACH)` and `wrapDialErrno("write", ENETUNREACH)` expecting no replay.

**[Major] fast-fail.** NXDOMAIN (`net.DNSError` with `IsNotFound`) is now treated as permanent and not retried, and the pre-send path has its own bound + schedule (`preSendMaxAttempts=3`, `500ms`/`1s`, modeled on reconnect.go) instead of the 2/4/8/16/30s rate-limit backoff. A dead local daemon or mistyped host now fails in about 1.5s. Added `TestPreSendBackoffWaitSchedule` to pin it.

**[Minor] doc.** Fixed the `SendWithRetry` comment that still claimed transport errors are never replayed.

One deliberate call worth flagging: the dial-scoping means the plain-string test fixtures (which carry no `*net.OpError`) no longer classify as pre-send, so I updated them to the real `wrapDialErrno` shape. A refused error that arrives already flattened past its OpError is now NOT retried, which is the conservative direction: we can't prove it was a dial, so we don't replay a POST on it. In production the client.Do error is fully structured, so real dial failures still retry.

@coderabbitai your Windows point is covered by the errno path: a Windows dial refusal arrives as `Op=="dial"` wrapping `syscall.Errno(WSAECONNREFUSED)`, which `errors.Is(err, syscall.ECONNREFUSED)` matches, so it retries structurally without needing the Windows wording. The POSIX string markers stay only as a fallback for a dial error flattened past its errno.

I left the earlier [Minor] (admitting a dial `i/o timeout`) out of this change to keep it focused on the two blockers; happy to add it as a follow-up if you'd like. Race detector and all five provider packages are green locally.

Vasanthdev2004:
Good catch @coderabbitai, and verified on Windows rather than assumed: a real refused dial there gives `errors.Is(err, syscall.ECONNREFUSED) == false`, because Windows raises the raw `WSAECONNREFUSED` (10061) while Go's `syscall.ECONNREFUSED` is a distinct value the net package never produces (and the POSIX string fallback doesn't match the Windows wording either). So the pre-send dial path silently never fired on Windows, the exact platform this PR set out to fix.

Fixed in `3af605d`: the dial errnos are now a platform-specific list (`dialPreSendErrnos`, POSIX everywhere plus the WSA codes on Windows), and `TestIsPreSendTransportErrorRealRefusedDial` does a REAL refused dial so the platform's true error shape is exercised, not just an errno-constant fixture (which is why the old tests passed on both platforms and missed this).

@gnanam1990 heads-up: your earlier CHANGES_REQUESTED was on the pre-fix commit `4c54a8b6`. Your two [Major] findings (dial-scoping + fast-fail) landed in `8b4d5e5`, and this adds the Windows errno fix on top. Would appreciate a re-look.

Vasanthdev2004:
Addressed in `7467a3b`.

**Plan 9 build tag (Major):** applied your suggested `//go:build !windows && !plan9` on `retry_dialerrno.go`. Worth noting the package already references `syscall.ECONNRESET` in `retry.go` (the post-send exclusion), which is likewise undefined on Plan 9, so the package was never intended to compile there. Zero targets linux/darwin/windows (the CI Smoke matrix), so I scoped the guard to prevent the specific undefined-symbol breakage you flagged rather than add a Plan 9 variant that would imply support that doesn't exist. linux/darwin/windows all cross-compile clean.

**retry.go Windows wording (Minor):** this is already handled by the previous commit (`3af605d`). `isPreSendTransportError` no longer relies on POSIX wording for Windows: `dialPreSendErrnos` now carries the `WSAECONNREFUSED`/`WSAENETUNREACH`/`WSAEHOSTUNREACH` codes on Windows, and `TestIsPreSendTransportErrorRealRefusedDial` does a real refused dial so the Windows error shape is actually exercised. That comment describes the exact bug that commit fixed.

gnanam1990 (dismissed):
**APPROVE.** Re-reviewed at `7467a3bc`. My earlier `CHANGES_REQUESTED` is fully addressed — I re-ran my exact attack against the new code rather than reading the diff.

## The concern I raised, and how it's fixed

I flagged that `isPreSendTransportError` matched `ENETUNREACH`/`EHOSTUNREACH`/`ECONNREFUSED` **anywhere in the error chain regardless of which socket op raised them**, so a post-send `read`/`write` failure (route drop / ICMP-unreachable mid-generation) was misclassified as "provably pre-send" and would **replay the non-idempotent completion POST** — double billing / double tool execution, the exact harm the PR exists to prevent.

The fix is the right shape: the errno match is now gated on `dialOpError(err) != nil` (`retry.go:220`), which requires an `Op=="dial"` `*net.OpError` (`retry.go:248-254`). That makes "no request bytes left this host" **structural** — a post-send read/write carrying the same errno can't match. The string-marker fallback (`connection refused` / `network is unreachable` / `no route to host`) is correctly *inside* the same dial gate, and TLS-handshake-timeout stays outside it (genuinely pre-send by wording, not a dial OpError).

Verified with the precise scenario:

```
dial ENETUNREACH   -> preSend=true    (safe retry preserved)
dial EHOSTUNREACH  -> preSend=true
dial ECONNREFUSED  -> preSend=true
read ENETUNREACH   -> preSend=false   (was the bug — now NOT retried)
write EHOSTUNREACH -> preSend=false   (no billable POST replay)
read ECONNRESET    -> preSend=false
```

## Nice additions

- **Regression coverage for exactly this**: `retry_test.go:102-103` — `"host unreachable on read is post-send"` and `"net unreachable on write is post-send"` — with a comment stating scoping to `Op=="dial"` must keep these from replaying the POST. That pins the fix so it can't silently regress.
- **The Windows errno split is correct and non-obvious.** `retry_dialerrno_windows.go` adds the raw Winsock codes (`WSAECONNREFUSED` = 10061, etc.) because `errors.Is` against the POSIX `syscall.ECONNREFUSED` returns *false* on Windows — the net package never produces the POSIX value there. Without that list a refused dial would silently never retry on Windows. Good catch, and the comment explains why.
- DNS handling is sound: `NXDOMAIN` (`IsNotFound`) is *not* retried (deterministic — retrying just stalls the agent through the whole backoff), while a timeout/SERVFAIL is.

## Verification

`gofmt` and `go vet ./internal/providers/providerio/` clean; the full `providerio` suite passes. CodeRabbit has also approved this head.

Solid iteration — the dial-scoping is the correct structural fix, not a string-match patch, and it ships with a regression test for the exact post-send case. Merge is kevin's call per the program gate.

anandh8x (changes requested):
Request changes: this regresses Plan 9 compilation. retry_dialerrno.go excludes plan9, but retry.go still references syscall.ECONNRESET and dialPreSendErrnos, for which there is no Plan 9 definition; retry_test also directly uses the unavailable syscall types/constants.  passes at this PR's merge base but fails on this head. Please provide a compatible Plan 9 fallback or gate the new code and tests consistently.

anandh8x:
Verification detail for my change request: running GOOS=plan9 GOARCH=amd64 go test -c ./internal/providers/providerio succeeds at this PR's merge base and fails on this head.

Vasanthdev2004:
Good catch, and you were right: I checked, and main builds clean for GOOS=plan9 while this branch did not, so it was a real regression I introduced (my earlier take that the package never built on Plan 9 was wrong).

Fixed by moving the two syscall touch-points out of retry.go behind per-platform helpers. retry.go no longer references syscall or dialPreSendErrnos directly; it calls isConnResetErrno and ranges dialPreSendErrnos, both supplied by retry_dialerrno.go / retry_dialerrno_windows.go and a new retry_dialerrno_plan9.go (empty errno list + isConnResetErrno returning false, so Plan 9 falls back to the existing string-marker heuristics). retry_test.go is tagged !plan9 since it builds real syscall.Errno dial shapes.

Verified GOOS=plan9 build and vet now pass, and windows/linux build, vet, gofmt, and the full providerio suite stay green with behavior unchanged on the real targets. Re-requested your review.

Vasanthdev2004:
@coderabbitai approve

jatmn (changes requested):

## Findings

- [P1] Do not classify a redirect-hop failure as pre-send
  `internal/providers/providerio/retry.go:123`
  `client.Do` can follow a redirect after it has already transmitted the original non-idempotent POST. If DNS, dialing, or the TLS handshake then fails while reaching the redirect target, the resulting error satisfies `isPreSendTransportError` and this loop replays the original request. A 307/308 provider redirect followed by a transient target-host failure can therefore run and bill the first completion, then submit it again. Disable redirects for this request path, or preserve enough redirect-hop state to retry only a failure on the initial request; add a redirect-then-transport-failure regression test.

- [P2] Keep the pre-send retry budget separate from status retries
  `internal/providers/providerio/retry.go:123`
  The pre-send condition uses the loop-wide `attempt` counter, which already includes 429/503 retries. After two safe status retries, the first refused-dial/DNS/TLS failure arrives as attempt 3 and is returned without any of the advertised pre-send retries; after one it gets only one retry. This contradicts the independent short pre-send budget and makes mixed rate-limit/network blips fail prematurely. Track pre-send failures (and their backoff ordinal) separately and cover a status-to-transport sequence.

- [P3] Correct the platform fallback/fixture documentation
  `internal/providers/providerio/retry_dialerrno_plan9.go:5`
  `internal/providers/providerio/retry_test.go:19`
  The Plan 9 comment says reset classification is unavailable and lists reset as a fallback inclusion, although `retry.go` still excludes the platform-independent `"connection reset"` marker. Separately, the test fixture says Windows `syscall.E*` constants exercise the real WSA dial path, while the new Windows implementation correctly says real dials carry distinct `windows.WSA*` errors; only the real-dial test covers that branch. Correct both descriptions so future maintenance does not weaken the cross-platform retry safety coverage.

Vasanthdev2004:
All three fixed in the latest push, thanks jatmn, genuinely good catches.

[P1] You were right that a redirect hop breaks the pre-send guarantee. SendWithRetry now disables redirect following (CheckRedirect returns http.ErrUseLastResponse on a cloned client), so any dial/DNS/TLS error from Do can only come from the single initial request, and a 3xx surfaces as the response for the existing non-2xx handling. Added a regression test asserting the POST is sent once and the 307 is not followed.

[P2] The pre-send retries now use their own counter and backoff ordinal, independent of the loop-wide attempt counter, so 429/503 retries no longer eat the pre-send budget. Added a status-then-transport-failure test that would have caught it.

[P3] Corrected both comments (the Plan 9 string-marker fallback and the wrapDialErrno fixture note about the real Windows WSA path).

Verified: gofmt, vet, and build clean on linux/windows/plan9, and the full providerio suite passes. Re-requested your review.

jatmn (changes requested):

## Findings

- [P2] Do not let pre-send retries consume the later status-retry budget
  `internal/providers/providerio/retry.go:115`
  `preSendAttempts` gives a dial/DNS/TLS failure its own limit, but each retry still advances the loop-wide `attempt` used by the 429/503 branch at line 158. For example, with `maxAttempts=3`, two safely retried refused dials followed by a 429 return that first 429 immediately; with the default, those failures also shorten the status retry window and start its backoff at a later rung. This is the inverse of the mixed sequence covered by the new test and contradicts the stated independent budgets. Track status attempts/backoff separately and add a pre-send-then-status regression case.

Vasanthdev2004:
Good catch again, you were right that I only fixed half of it. Every retry still advanced the loop-wide counter the 429/503 branch read, so pre-send retries shortened the status window and started its backoff a rung late, exactly the inverse of the case the new test covered.

Fixed by giving the status path its own statusAttempts counter and backoff ordinal. With both budgets tracked separately nothing reads the shared loop counter any more, so it is gone entirely. Behaviour is unchanged when only one kind of failure happens: the first retryable status still waits on rung 1 and still gets the full maxAttempts window.

Added TestSendWithRetryStatusBudgetSurvivesPreSendRetries as the mirror of the existing test: two refused dials then 429s, asserting the status path still gets its whole window (the old shared counter surfaced the first 429 immediately).

One caveat worth stating plainly: Windows Smart App Control switched on on my machine and is now blocking the unsigned go.exe, so I could not run gofmt, vet, the cross-compiles, or the package tests locally for this push the way I did for the previous ones. I have reasoned through the change carefully and the diff is small, but CI is the actual verification here rather than a local run. Please treat it that way in your re-review, and I will re-run the full local matrix once the toolchain is usable again.

jatmn (changes requested):

## Findings

- [P2] Preserve caller redirect behavior while preventing redirected-request retries
  `internal/providers/providerio/retry.go:106`
  Replacing every supplied client's `CheckRedirect` with `http.ErrUseLastResponse` turns a normal 307/308 from a configured OpenAI-/Anthropic-compatible gateway into a terminal non-2xx response. `SendWithRetry` is the shared completion path, and each provider treats that returned 3xx as an upstream error; before this change `client.Do` followed the redirect. This also discards an injected client's redirect policy. Keep the redirect behavior and mark an attempt as non-retryable once a redirect has been entered, so a dial failure after a transmitted redirected POST cannot replay it without making redirecting gateway endpoints unusable.

Vasanthdev2004:
You are right, and my fix was too blunt. Blocking redirects outright made every redirecting gateway return a terminal 3xx through the shared completion path, and it threw away an injected client's own policy. That is a worse trade than the bug it closed.

Done your way in the latest push: CheckRedirect now observes rather than prevents. It sets a per-attempt flag, then delegates to the caller's policy when one is set, and otherwise mirrors net/http's default ten-hop cap. The pre-send retry is gated on that flag, so an attempt that entered a redirect is never replayed, while redirecting endpoints keep working exactly as before.

Three regressions added: a redirect is followed through to the final response, a dial failure after a redirect is not replayed (transport called twice, never a third time), and a caller-supplied CheckRedirect still runs and still decides.

My go toolchain is usable again, so this one is properly verified locally rather than leaning on CI: gofmt and vet clean, builds for linux, windows, darwin and plan9, the three new tests pass, and the full providerio suite is green.

Thanks for pushing back on this one, the result is better than what I had.

jatmn (changes requested):

## Findings

- [P1] Complete the redirect no-replay guard for retryable responses
  `internal/providers/providerio/retry.go:181`
  This behavior existed before the PR, but the new redirect wrapper explicitly makes a followed redirect non-retryable and prevents only redirect-hop transport-error replay. A 307/308 followed by a 429, 503, or 529 still reaches this branch, which closes the final response and starts a new attempt at the original URL. Since the origin already received the non-idempotent POST, that can duplicate billable work—the same ambiguity the new guard is meant to avoid. Apply the redirect guard to retryable responses as well, and add redirect-then-retryable-status coverage.

Vasanthdev2004:
Thanks for staying on this one. I want to push back on this finding rather than apply it, and I would rather set out the reasoning and be told I am wrong than quietly do the safer-looking thing.

You are right that the guard is asymmetric. I do not think it is incomplete.

The two paths differ in what is actually known about the request:

- **Transport error after a redirect.** The original POST reached the origin, and the dial to the redirect target then failed. Whether the target received it and began generating a completion is unknown. Replaying can duplicate billable work, so that path refuses. This is the case you originally caught and it is a real one.
- **Retryable status after a redirect.** Every hop answered. A 307/308 is the origin declining to process and delegating; a 429, 503 or 529 is the target declining as well. Nothing accepted the request, so a replay duplicates nothing. That is the same assumption the package already documents for the un-redirected case: "exactly like a 429/503 where we KNOW the request was not accepted."

Put another way, the exposure here is identical to a 429 with no redirect in play, which this package has always retried and which this PR does not change. The redirect adds ambiguity to the transport case, because "provably nothing was sent" stops being true, but it adds none to the status case.

The cost of guarding is not neutral either. A rate-limited completion behind a redirecting gateway would stop backing off and fail outright. That is the same regression class you flagged when I disabled redirects wholesale, and you were right then, which is most of why I am reluctant now.

What I have done instead, in 7c0b8349: added the case as a test asserting the intended behaviour, and a comment at the branch explaining why it is not guarded. Reading it as an oversight was reasonable precisely because nothing recorded the intent, and that was my omission.

```
--- PASS: TestSendWithRetryRetriesRedirectedRetryableStatus
```

If you still think the replay is unsafe, the case I have not managed to construct is a hop that returns a retryable status having already done billable work. If you have one in mind I will take the guard, since that would break the "nothing accepted it" premise the whole branch rests on. Happy to defer to you or kevin on it either way.

gnanam1990 (dismissed):
## Verdict

Approve.

Reviewed at `7c0b8349c06b`, base `ce4a996ffac4`, re-confirmed against the live head before posting.

This does not clear @jatmn's outstanding request for changes, and should not — he raised the findings and whether they are answered is properly his call. What follows is my independent read, including one point where I think the current code is right and the review comment is not.

**All three of the mechanical findings are addressed, and I checked each rather than taking the commit log on trust.**

Redirects are followed again. `SendWithRetry` no longer replaces the caller's policy with `http.ErrUseLastResponse`; it wraps it. `callerCheckRedirect` is captured and delegated to, and when the caller supplied none the wrapper mirrors net/http's own default of stopping after ten hops rather than silently changing it. A redirecting gateway therefore keeps working, which was the substance of the objection.

The two retry budgets are genuinely independent. `preSendAttempts` and `statusAttempts` are separate counters with separate backoff ordinals, and the comment explains why the coupling mattered in both directions. The scenario in the review — two refused dials followed by a 429 returning immediately — no longer holds.

The `redirected` flag is reset per attempt rather than per call, so a redirect entered by one attempt cannot make a later one non-retryable.

**Where I disagree, with reasoning.** The P1 finding asks for the redirect guard to be applied to retryable statuses as well. The code deliberately does not, and documents why. I find that reasoning correct: a transport failure after a redirect leaves it genuinely unknown whether the target received and began the POST, which is what makes replay hazardous; a 307/308 followed by a 429 is a case where every hop answered and none accepted. The origin delegated and the target declined. Nothing was accepted, so a replay duplicates nothing.

The second half of the argument is the one I would weigh most. The exposure after a redirect is identical to a 429 with no redirect in play, which this package has always retried. Guarding on redirects there would not close a gap — it would make a rate-limited completion behind a redirecting gateway fail outright instead of backing off, which is a regression for a common deployment shape. If the concern is duplicate billable work on retryable statuses generally, that is a real question but it is a pre-existing one about the whole retry policy, not something this change introduces, and it should be settled on its own terms.

**On the thing that makes this change matter beyond the issue.** `providerio` had no `CheckRedirect` at all, while `internal/mcp/network_redirect.go:20` has had a same-origin policy for some time — so a provider redirect could carry the Authorization header where an MCP redirect could not. This does not close that asymmetry, and I am not asking it to, but it is now the natural place to.

I also checked that the discriminator at `internal/providers/providerio/retry.go:149` survives the rewrite, because it is the reference implementation in this repository for telling caller cancellation apart from a transport failure on a non-idempotent billable request. It does, and it is now applied in both the transport branch and after the response at `:213`.

**Freshness.** The branch is 44 commits behind main, which is the one thing I would ask you to act on before merge. It merges cleanly and, more importantly, the *merged* tree builds and `./internal/providers/providerio/...` passes — I verified that rather than trusting the absence of a conflict, since a clean textual merge in this repository has produced duplicate definitions before. A rebase would still be worth doing so the CI signal reflects the tree that would actually land.

**Verification.** On macOS (darwin/arm64), `go build ./...` clean on the merged tree, `./internal/providers/providerio/...` passing.

**Limitations.** macOS only. I did not exercise the platform-specific dial-errno classification on Linux, Windows or Plan 9, so the P3 documentation point about those fixtures is one I can neither confirm nor refute.

jatmn (changes requested):

## Findings

- [P2] Rebase onto current `main` before merge
  `pull/750` (branch is 47 commits behind `main`, status `diverged`)
  This is merge hygiene, not a logic defect — the head merges cleanly and CI is green today. It is still worth doing before land so the checks exercise the tree that would actually ship. `gnanam1990` flagged the same freshness concern.

- [P3] Retry dial-phase connect timeouts, not only refused/unreachable dials
  `internal/providers/providerio/retry.go:263-292`
  This is a real gap against #447, not comment drift. A production connect timeout from `net/http` usually arrives as a `*net.OpError` with `Op=="dial"` and `Timeout()==true`, but the message contains `i/o timeout`, so the blanket exclusion at line 266 returns before the dial gate can admit it. Dial-phase `ETIMEDOUT` / “connection timed out” shapes are also omitted from `dialPreSendErrnos` and the string inclusions, so they fall through to no-retry as well. TLS-handshake timeouts and refused dials are covered; intermittent connect timeouts are not. The tests currently pin the conservative no-retry behavior for a flat `i/o timeout` string, so this looks intentional in the branch but incomplete relative to the issue. `gnanam1990` flagged the same gap. Please admit connect-phase timeouts (`dialOpError(err) != nil && dialOpError(err).Timeout()`, and/or `ETIMEDOUT` in the errno list) before the generic `i/o timeout` string exclusion, and add table cases for dial-shaped timeouts so post-send read timeouts stay excluded.

- [P3] Retry `proxyconnect` failures for HTTPS clients behind an HTTP proxy
  `internal/providers/providerio/retry.go:309-314` (`dialOpError`)
  Also a real classification gap, narrower surface. When `HTTPS_PROXY`/`HTTP_PROXY` is in play, a failed CONNECT is surfaced as `Op=="proxyconnect"`, not `Op=="dial"`. Those errors are provably pre-send — no request bytes reached the model host — but they miss the dial gate and the string fallback (which is also behind `dialOpError`). Please extend the connect-phase gate to treat `proxyconnect` the same as `dial` (errno and/or `Timeout()`), and add a fixture using a real `*net.OpError{Op:"proxyconnect", ...}` shape.

- [P3] Update `SendWithAuthRetry` docs for the new pre-send retry path
  `internal/providers/providerio/auth.go:22-28`
  Doc drift only — no runtime impact. The comment still says transient retries are only 429/503/529 via the inner `SendWithRetry`; the inner path now also retries provably pre-send transport failures on its own short budget. Please bring this in line with `retry.go`.

## Optional polish (not blocking)

- **Lifetime `preSendAttempts` vs in-flight 429/503 recovery** (`retry.go:130-131,166-177`). Real behavior, debatable whether it is a bug: `preSendAttempts` is a lifetime cap of three across the whole call and is never cleared after a successful `Do`, so two early dial retries followed by a 429 backoff and then another dial failure can return a transport error even when `statusAttempts` still has budget. The new tests cover sequential status→pre-send and pre-send→status, not this interleaved case. If the lifetime cap is intentional, a one-line comment would help; if not, reset the counter after a successful transport hop and add a regression.

## Prior review threads

The mechanical items from earlier rounds look addressed on `7c0b8349c06b`: redirect observation instead of blocking, caller `CheckRedirect` preservation, independent `preSendAttempts` / `statusAttempts` budgets in the sequential cases covered by the new tests, Plan 9 buildability, and dial errno scoping to `Op=="dial"`.

On the remaining [P1] about guarding retryable statuses after a redirect: I read your pushback and `gnanam1990`'s independent analysis. I am withdrawing that request. A transport failure after a redirect really is ambiguous about whether the target began processing the POST, so refusing replay there is correct. A 307/308 followed by 429/503/529 is the same "server explicitly declined" contract this package already uses for un-redirected rate limits, and blocking it would regress redirecting gateways without closing a new gap relative to existing 429 retry policy. `TestSendWithRetryRetriesRedirectedRetryableStatus` documents the intent adequately.

Cross-host credential forwarding on redirect, 301/302/303 replay semantics, `SendWithAuthRetry`'s 401 outer loop, flattened dial strings, Windows `WSAECONNRESET` exclusion, and permanent DNS mis-tags were also raised in the blind searches; those are either pre-existing, conservative-by-design, or too minor to block merge.

Vasanthdev2004:
All four addressed in c6ad2a43, and thank you for withdrawing the P1 explicitly rather than leaving it to lapse. That is the outcome I wanted from arguing it, and I would rather have been told I was wrong.

**[P2] Rebased.** Was 47 behind; now on current `main`, replayed clean, CI re-running against the tree that would actually ship.

**[P3] Connect-phase timeouts.** Confirmed, and worse than the comment drift I half expected: the blanket `i/o timeout` exclusion returns before the connect gate is reached, so a dial timeout could never be admitted no matter what the errno list said. Fixed by admitting connect-phase timeouts ahead of the wording exclusions, gated so a post-send read timeout, which is also `Timeout()`, still cannot replay. `ETIMEDOUT` and `WSAETIMEDOUT` are in the errno lists too, for a timeout that reaches us with the errno but without the flag.

Table covers both directions, since the exclusion has to survive:

```
dial timeout           -> retry
proxyconnect timeout   -> retry
read timeout           -> no retry
write timeout          -> no retry
```

**[P3] proxyconnect.** Also confirmed. The gate matched `Op=="dial"` only, so behind `HTTPS_PROXY` every pre-send failure fell through, which is the environment where transient connect failures are most common. `connectOpError` now accepts both ops, with a real `*net.OpError{Op:"proxyconnect"}` fixture and a read-phase counterpart to prove the phase scoping still holds.

**[P3] `SendWithAuthRetry` docs.** Corrected.

**On the optional item, I took the reset rather than the comment.** The lifetime cap was not intentional; it was a consequence of where I initialised the counter. Your interleaved case is real: two dial blips, a 429, then one more blip returned a transport error with status budget still available, which is precisely the coupling the split was supposed to remove. The counter now resets after a hop that reaches the server, so it bounds consecutive connect failures. Total work stays bounded because every reset costs a status attempt and those are capped.

There is a regression for it now, in the ordering the earlier tests missed:

```
--- PASS: TestPreSendBudgetResetsAfterSuccessfulHop
    (dial, dial, 429, dial, 200)
```

Each of the three code changes fails its own test when reverted, so none of this is asserting around the fix.

`gofmt`, `go vet`, `GOOS=windows go vet`, and builds for linux, darwin and windows are clean; the package passes.

Vasanthdev2004:
@jatmn nudge: your changes-requested is on `7c0b8349`, head is now `c6ad2a43` with all four findings addressed. Detail is in my earlier comment; short version below.

Connect-phase timeouts and `proxyconnect` are both admitted now, with the connect gate scoped so post-send read and write timeouts stay excluded. `ETIMEDOUT` and `WSAETIMEDOUT` are in the errno lists. `SendWithAuthRetry` docs corrected. Rebased onto current `main`, was 47 behind.

On the optional item I took the reset rather than the comment. The lifetime cap was not intentional, and your interleaved case is real: two dial blips, a 429, then one more blip returned a transport error with status budget still free. There is a regression for that ordering now.

Each of the three code changes fails its own test when reverted.

@anandh8x you are also still blocking here, on `7467a3bc`, for the Plan 9 compile regression. That is fixed and I checked it including the test file, which your review specifically called out:

```
GOOS=plan9 go build  -> OK
GOOS=plan9 go vet    -> OK
```

`retry_test.go` carries `//go:build !plan9`, so the syscall fixtures never reach that platform. Both of you need to clear for this to move.

gnanam1990 (approved):
## Verdict

Approve — re-affirming after the rebase, and my freshness point is now closed.

Reviewed at `c6ad2a43a31b`, base `81a5e6c`, re-confirmed live before posting. My earlier approval was against `7c0b834` and has been dismissed.

**The staleness I flagged is gone.** This was 47 commits behind main when I last looked; it is now level with it, so CI reflects the tree that would actually land and my "please rebase" ask is answered.

**Two things are genuinely new since my approval, not just re-listed by the rebase.**

`bb4857d` pins the redirect asymmetry for retryable statuses in a test. That is the right response to the one place @jatmn and I read the code differently — rather than arguing it further, the decision is now encoded, so if someone later concludes a 429 after a redirect should not be replayed, they have to change a test that says why. I would rather see a disagreement resolved that way than left in a thread.

`c6ad2a4` widens the pre-send classification to connect-phase timeouts and proxy `CONNECT` failures. I went looking for the way this could go wrong, because "timeout" is exactly the word that should not be admitted to a pre-send path: a post-send read timeout is also `Timeout()`, and replaying a POST after one could duplicate billable work. `connectOpError` is the guard, and it holds — it matches only `Op == "dial"` and `Op == "proxyconnect"`, and a failure on an established connection carries `Op == "read"`/`"write"`, so it cannot reach the branch. The comment says this outright and it is accurate. Admitting the timeout check *before* the wording exclusions is also necessary rather than incidental, since net surfaces it with "i/o timeout" in the message.

Adding `proxyconnect` is well-argued: a failed CONNECT means the tunnel to the model host was never built, so no bytes reached it, and omitting it left every pre-send failure behind a corporate proxy unretried — which is the environment where transient connect failures are most common.

**One observation on the reset, non-blocking.** `preSendAttempts = 0` on a successful transport hop makes the two budgets multiplicative in the worst case rather than independent. Each reset is genuinely paired with a consumed status attempt — I checked the ordering, the reset sits immediately above `statusAttempts++` and continuing requires spending one — so the total is bounded at `maxAttempts × preSendMaxAttempts`, 9 dial retries at a sub-second schedule. That is fine in practice. It is worth a word in the description because the description currently says the budgets are "fully independent", and in the worst case they compose rather than staying separate.

**Verification.** On macOS (darwin/arm64): `gofmt -l`, `go vet ./...` and `git diff --check` clean; `./internal/providers/providerio/...` passing; `-race -count=5` clean, which matters on a retry loop with a closure writing `redirected`. Builds clean for linux/amd64 and windows/amd64, and `internal/providers/providerio` builds for plan9/amd64, which is the platform the errno-classification split exists for.

**Limitations.** macOS only. I did not exercise the platform-specific dial-errno classification on Linux, Windows or Plan 9, so @jatmn's P3 documentation point about those fixtures is one I can still neither confirm nor refute.

This does not clear @jatmn's outstanding request for changes, which is his to resolve.

Vasanthdev2004:
@jatmn @anandh8x both of your change requests are the only things left on this one, and I believe both are now answered. Rather than just assert that, here is the evidence for each.

@jatmn, your P2 was to rebase onto current `main` before merge. Done, and verified rather than assumed:

```
head        = c6ad2a43
origin/main = 81a5e6cb
merge-base  = 81a5e6cb
behind main = 0 commits
```

The merge base is exactly `origin/main`, so there is nothing to replay. @gnanam1990 re-affirmed at this head afterwards and explicitly closed their freshness point, and CodeRabbit approved it too.

@anandh8x, your finding was that this regressed Plan 9 compilation, because `retry_dialerrno.go` excluded plan9 while `retry.go` still referenced syscall constants. That is fixed, and the package builds clean:

```
GOOS=plan9 GOARCH=amd64 go build ./internal/providers/providerio/
exit=0
```

Worth adding what I found while checking, because it changes what your finding should be scored against. A whole tree build for plan9 does still fail, but only in `internal/fsutil/rename.go` and `internal/execution/process_unix.go`, and those exact failures reproduce identically on `origin/main` at `81a5e6cb` with none of my code in the picture. So plan9 has never built in this repo, and nothing in this PR made that worse. Happy to raise a separate issue for the pre-existing breakage if you think it is worth tracking, but it should not gate this.

If either of you disagrees with any of that I would rather hear it than have the review lapse. Otherwise a re-review would unblock it.

Vasanthdev2004:
Raised the pre-existing Plan 9 breakage as #817 so it is tracked separately from this PR, as offered above. Short version: it fails in six first-party packages on `origin/main` at `81a5e6cb` and also inside `charm.land/bubbletea/v2`, so it is not reachable from this repository alone. The issue asks for the supported-target policy to be written down rather than for Plan 9 to be fixed.

anandh8x (approved):
Reviewed the latest head c6ad2a4. The retry path stays limited to structurally pre-send failures, keeps post-send read/write timeouts non-retryable, preserves redirect behavior, and gives status and connection failures bounded independent budgets. The previously requested connect-timeout, proxy CONNECT, docs, and freshness concerns are addressed. providerio passes under the race detector locally. Good to merge.

kevincodex1 (approved):

PR #755, opened 2026-07-19
fix(sandbox): add platform-specific write lock around grant state file (#752)

### LINKED ISSUE #752: security(sandbox): concurrent grant writes can resurrect revoked permissions
## Summary

Persistent sandbox grant read-modify-write transactions are protected only by an in-process mutex belonging to one `GrantStore` instance. Two separate `zero` processes can each read the same on-disk grants file, and the second process's atomic-rename write can overwrite the first — including silently restoring a grant the first process had just revoked.

## Affected code

- `internal/sandbox/grants.go:121-150`
- `internal/sandbox/grants.go:298-320`
- `internal/sandbox/grants.go:371-391`
- `internal/sandbox/grants.go:591-607`

Configuration writers have the same lost-update pattern in `internal/config/writer.go:14-55` and `internal/config/writer.go:744-778`, though there the usual impact is lost settings rather than a restored security permission.

## Race outline

1. Process A and Process B both read the current grants file into memory.
2. Process A revokes a grant and atomically writes back the file (write-temp + rename).
3. Process B, still holding its earlier in-memory read (which still includes the grant A just revoked), later writes back its own state and atomically overwrites A's file.
4. The revoked grant is now restored on disk, silently reinstating access that was supposed to be gone — with no indication to either process or the user that the revocation didn't stick.

## Impact

Atomic rename prevents malformed/corrupted JSON on disk, but not this lost-update pattern. A stale writer can restore a persistent sandbox grant (e.g., a command-prefix or path allow) that another process deliberately revoked moments earlier.

## Suggested fix

Acquire an interprocess lock around the *complete* read-modify-write transaction (not just the in-process mutex), and re-read the file's current contents after acquiring that lock, immediately before mutating and writing back.

## Reference

Item AUD-005 from the 2026-07-18 codebase audit (`docs/CODEBASE_AUDIT_2026-07-18.md`).

PierrunoYT:
## Summary
- Adds platform-specific file-level write locks for the persisted grant state (grants_lock_unix.go, grants_lock_windows.go).
- Wires lockStateFile() into Grant, GrantCommandPrefix, Revoke, RevokePath, and Clear so concurrent processes serialize access to the grant JSON file.
- Adds regression coverage in grants_test.go.
- Adds .gitattributes to enforce LF line endings for Go files so gofmt checks pass on Windows.

Closes #752

#### Test plan
- [x] gofmt -l . clean
- [x] go vet ./... passes
- [x] go test ./... passes
- [x] go run ./cmd/zero-release build succeeds
- [x] go run ./cmd/zero-release smoke passes
- [x] git diff --check clean

Generated with [Devin](https://devin.ai)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sandbox grant state handling to prevent lost or duplicated updates when multiple processes modify the same grants data concurrently.
  * Added cross-platform file locking so migration/notice consumption is serialized and migration work doesn’t overlap across instances.
  * Migration notice consumption now tolerates read-only grants directories and avoids creating artifacts when no notice is pending.
* **Tests**
  * Added timeout-based concurrency tests covering serialized writes, serialized migration notice consumption, and exclusive migration execution across store instances.
* **Chores**
  * Normalized line endings to LF for Go-related files via Git attributes.


Vasanthdev2004 (dismissed):
Approve. This closes the #752 lost-update/resurrection hole cleanly, and the Windows path in particular is done right.

Checked on head a60e60c8:

- The full read-modify-write is under the inter-process lock: every mutator (Grant, GrantCommandPrefix, Revoke, RevokePath, Clear) does lockStateFile then readState then mutate then writeState, with the read inside the lock, so two processes cannot both read-then-write and clobber a revoke.
- writeState is atomic (temp file in the same dir plus os.Rename), so the unlocked reader / permission-check path never sees a torn file, only the complete old or new one.
- Fail-closed: on a lock error or the 5s timeout the mutators return the error and abort, and the CLI surfaces a Revoke error instead of printing a success line, so a dropped revoke cannot be mistaken for a completed one.
- Windows LockFileEx (grants_lock_windows.go): success is keyed off the BOOL return value (`if result != 0`), not off the err from LazyProc.Call, which is always the last-error. That is the right way to do it and sidesteps the trap where a failed lock reads as success and silently disables the protection. Flags are LOCKFILE_EXCLUSIVE_LOCK|LOCKFILE_FAIL_IMMEDIATELY with a zeroed OVERLAPPED, which is correct.
- Built and ran internal/sandbox locally: gofmt, vet and build clean, grant tests pass including the cross-store serialization test that exercises the real file lock. CI green.

One optional note, not blocking: grants_lock_unix.go uses `//go:build !windows`, which nominally includes plan9 and js where syscall.Flock is undefined. It matches the existing convention in the repo (permissions_lock_unix.go and friends), and Zero only ships linux/darwin/windows, so it builds fine for every real target. Flagging only because I hit the same tag question on another lock PR; take it or leave it.

LGTM.

jatmn (changes requested):

## Findings

- [P2] Rebase this branch onto the current `main` before merging
  `internal/sandbox/grants.go`
  The PR head is based on `ce4a996`, while `main` is now `96859c9`; GitHub reports the PR as `CONFLICTING`/`DIRTY`, and the merge has content conflicts in both this file and `grants_test.go`.  Resolve the conflicts against the current state model and rerun the grant-state review/tests rather than selecting either side wholesale.  In particular, retain main's schema-v3 and legacy/policy migration behavior, and extend the new interprocess transaction to `ConsumeMigrationNotice` plus the migrations that currently write through `readState`; otherwise the rebase would either regress existing grants or leave additional lost-update paths.

PierrunoYT:
Merged current `main` (`5885423`) — the PR is `MERGEABLE` again — and resolved the conflict against the current state model rather than taking either side wholesale, as asked.

**[P2] Rebase onto current `main`** — done. `main`'s schema **v3** and its legacy/policy migration behaviour are kept exactly as they are (the only edit to that code is the lock gate described below); the branch contributes the interprocess lock. `main`'s migration tests — v1 exact-grant migration with backup, the legacy shell-allow invalidation, and the policy-change invalidation — pass unchanged.

**Extending the transaction to the paths the branch missed** — you were right that a straight rebase would have left lost-update holes:

- **`ConsumeMigrationNotice`** is a read-modify-write like every other mutator, so it now takes the same lock. Without it, two frontends starting together could both read `NoticePending`, both print the notice, and clobber each other's write of the rest of the state.
- **The migrations that write through `readState`** (legacy schema, changed policy version) also write a backup file, so they must not run unserialized either. `readState` is now split:
  - `readStateLocked` — for callers already holding the lock (all five mutators plus `ConsumeMigrationNotice`); migrations write through directly. This also stops them re-entering the lock they already hold, which would deadlock.
  - `readState` — the unlocked lookup/list path. It detects that a migration is needed, acquires the lock, and **retakes the read there**, so a file another process migrated while this one waited is not migrated a second time (and its backup not rewritten).

**Tests** — `TestGrantStoreSerializesMigrationNoticeAcrossStores` and `TestGrantStoreMigrationTakesTheFileLock` mirror the existing cross-store test: with one store holding the lock, notice consumption and the legacy migration both block (and the on-disk file is asserted *unchanged* during that window), then complete once it is released, with the notice not reported twice.

**@Vasanthdev2004 on `//go:build !windows`** — leaving it. It matches the convention already used by `internal/daemon/lock_posix.go`, `internal/config/process_posix.go`, and `internal/background/process_posix.go`, and Zero only ships linux/darwin/windows, so `syscall.Flock` is defined for every real target. Switching just this file to `//go:build unix` would make the package inconsistent with its neighbours; if the repo wants that tag, it is worth doing as one sweep rather than here.

**One thing worth a maintainer's call:** the `.gitattributes` addition (`*.go text eol=lf`) is unrelated repo hygiene bundled into a security fix. It does not change committed content — git already stores LF — only working-tree normalization, and it is genuinely useful on Windows checkouts where `gofmt -l` otherwise flags every file. Happy to split it into its own PR if you would rather keep this one narrow.

**Validation** (Windows host, Go 1.26.5): `go build ./...`, `go vet ./...`, `go test ./... -count=1` all pass — including the Windows `LockFileEx` path natively; `GOOS=linux|darwin go vet ./internal/sandbox` clean (the `flock` path is compile-verified here and executes in the ubuntu/macOS CI jobs); `gofmt` clean on every changed file.

@jatmn @Vasanthdev2004 — ready for another look. I can't use the reviewer-request button on this repo, hence the mention.

@coderabbitai review

jatmn (changes requested):

## Findings

- [P2] Do not require a writable grants directory when there is no migration notice
  `internal/sandbox/grants.go:150`
  This PR makes `ConsumeMigrationNotice` create and lock `<grants>.lockfile` before reading the grants state. Both normal startup and `zero exec` call it unconditionally, so a user with no grants file and `ZERO_SANDBOX_GRANTS_PATH` (or its parent config directory) on a read-only mount now fails at `MkdirAll`/`OpenFile` with “failed to migrate sandbox grants.” Before this change, `readState` returned an empty state without writing anything. First detect whether a migration/notice write is required, then acquire the lock and re-read before clearing the pending flag.

PierrunoYT:
Pushed `772f5ea`, addressing the P2 finding.

**[P2] Do not require a writable grants directory when there is no migration notice** (`internal/sandbox/grants.go:150`)

Confirmed, and worth being precise about where it came from: the original commit (`a60e60c`) did **not** touch `ConsumeMigrationNotice` — upstream's version called `readState()`. The unconditional `lockStateFile()` was introduced by my merge resolution in `5885423`, when `main` grew this function and I wired the lock into it alongside the real mutators. It doesn't belong there, because `lockStateFile` `MkdirAll`s the grants directory and creates `<grants>.lockfile`, which turns "read the grant state" into "require a writable grants directory". Since `internal/cli/app.go:694` treats any error as fatal, the result is a hard startup failure — `failed to migrate sandbox grants` — for a user with no grants file whose grants path sits on a read-only mount.

Fixed by peeking at the state **without** the lock and returning early when nothing needs writing. The lock is taken only when a write is genuinely required, and then the state is **re-read under it**, because the unlocked peek is only a snapshot: another process may have consumed the notice or migrated the file while this one waited, and `readStateFile`'s contract says the returned state is meaningless when `needsMigration` is set. `needsMigration` deliberately falls through to the locking path — a schema/policy migration must write, so a read-only path legitimately fails there, exactly as it did before this lock existed via `readState`'s own migration path.

`TestGrantStoreSerializesMigrationNoticeAcrossStores` still passes unchanged, so the serialization this PR is actually about is intact.

**Tests** — both verified to fail against the pre-fix code:

- `TestConsumeMigrationNoticeToleratesAReadOnlyGrantsDirectory` exercises the real failure. Pre-fix on Linux it fails with `open .../sandbox-grants.json.lockfile: permission denied` — the error `app.go` wraps as "failed to migrate sandbox grants". It **skips** on Windows (Unix mode bits don't deny writes there) and as root (permissions are bypassed), and it probes the directory first so it skips rather than passing vacuously if the filesystem ignores `0500`.
- `TestConsumeMigrationNoticeWritesNothingWithoutAPendingNotice` is the portable companion, since the above can't run everywhere: it asserts nothing is created at all — no lock file, no grants file, not even the directory. Pre-fix it fails on both Windows and Linux with `the grants directory was created (contents: [sandbox-grants.json.lockfile])`.

**Validation:**
- `GOOS=linux`/`GOOS=darwin` cross-builds: clean.
- `go test ./internal/sandbox/... -run 'TestConsumeMigrationNotice|TestGrantStore' -count=25`: pass.
- `go test -race ./internal/sandbox/... -count=5` via WSL: pass, plus `go build`/`go vet ./...` under WSL — clean.
- `gofmt` clean on both changed files (checked against LF-normalized copies; this checkout is CRLF via `core.autocrlf`, which makes plain `gofmt -l` flag every file in the repo).

One honest note on that WSL race run: I had to `-skip TestSelectBackendChoosesPlatformAdapterWithFallback`. It fails there because the backend selector correctly detects WSL2 and returns `Name:"wsl"` where the test expects native Linux. I verified it fails identically with my changes stashed, so it is a pre-existing environment artifact of running the suite inside WSL, not something this PR touches — and I left it alone rather than adjusting an unrelated test.

Vasanthdev2004 (approved):
Approving again on 772f5ea8. jatmn's P2 is addressed exactly the way it should be.

ConsumeMigrationNotice now peeks for a pending notice WITHOUT taking the interprocess lock, and returns early when there is no migration and no pending flag, so the common startup path no longer creates the grants directory or opens the lockfile. A read-only grants path keeps working when nothing needs writing, which was the regression. Only when a write is genuinely required does it take the lock, and it then re-reads state under that lock rather than trusting the unlocked peek, which is the right call: another process may have consumed the notice or migrated the file in between, and clearing the flag is a read-modify-write like any other mutator.

Everything from my earlier pass still holds: the full read-modify-write is inside the lock for all five mutators, writeState is atomic via temp plus rename so unlocked readers never see a torn file, failures fail closed with the error surfaced to the CLI, and the Windows path keys success off the LockFileEx return value rather than the always-non-nil LazyProc error.

CI is green on all three OSes, and I re-verified the sandbox package locally on Windows.

LGTM.

jatmn (approved):
LGTM

kevincodex1 (approved):
LGTM

PierrunoYT:
Current status audit for head `772f5ea`:

- no unresolved review threads or reviewer requests remain
- all GitHub checks pass
- the PR is approved and mergeable
- no branch-specific code or CI issue requires another change

The remaining `BLOCKED` state is branch-policy approval gating rather than a conflict or failing check. PierrunoYT requests the remaining qualifying approval when convenient.

PR #759, opened 2026-07-19
fix(lsp): dispatch notifications off read loop

### LINKED ISSUE #758: fix(lsp): re-entrant notification handlers can deadlock
## Summary

LSP server notifications are dispatched synchronously on the client's sole read-loop goroutine. If a notification handler calls Client.Call, it waits for a response that only the blocked read loop can receive, causing a deadlock until the call context expires or the connection closes.

## Affected code

- internal/lsp/client.go:99-137`n- internal/lsp/client.go:151-180`n
## Details

Client.readLoop invokes the configured NotificationHandler inline when it receives a server notification. The public handler contract does not prohibit re-entrant client calls. A handler can therefore send a request with Client.Call and block waiting for its response while the same goroutine responsible for reading and delivering that response remains inside the handler.

The current production diagnostics handler does not make a re-entrant call, but future handlers and other package consumers can trigger this behavior.

## Suggested fix

Dispatch notifications through a bounded serialized worker so notification ordering is preserved without blocking the protocol reader. Define the callback lifecycle and shutdown behavior, and add a regression test where a notification handler performs a request and receives its response without deadlocking.

## Reference

Item AUD-008 from the 2026-07-18 codebase audit (docs/CODEBASE_AUDIT_2026-07-18.md).

PierrunoYT:
## Summary

- dispatch server notifications through a bounded per-client worker queue
- preserve notification ordering without blocking JSON-RPC response delivery
- allow notification handlers to make re-entrant Client.Call requests safely
- add regression coverage for re-entrant calls and ordered notification delivery

## Verification

- go test ./internal/lsp`n- go vet ./...`n- go test ./...`n- go run ./cmd/zero-release build`n- go run ./cmd/zero-release smoke`n- git diff HEAD --check`n
## Environment limitations

- go test -race ./internal/lsp requires CGO, which is disabled in this environment
- pinned golangci-lint and govulncheck build with Go 1.25 and cannot analyze this Go 1.26.5 repository
- make fmt-check could not run because make is unavailable; touched files were formatted with gofmt`n
Fixes #758

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented deadlocks when notification handlers trigger client calls.
  * Ensured ordered, bounded LSP notification delivery with reliable overload handling.
  * Improved shutdown behavior by draining accepted notifications and rejecting later ones.
  * Improved diagnostics synchronization, staleness detection, and cancellation handling.
  * Rejected oversized incoming LSP frames to limit excessive memory use.
* **Tests**
  * Expanded coverage for notification ordering, bursts, overloads, shutdown, frame limits, and diagnostics lifecycle edge cases.


jatmn (changes requested):

## Findings

- [P1] Preserve notifications when the dispatch queue overflows
  `internal/lsp/client.go:214`
  When a notification handler is blocked (including in the re-entrant `Client.Call` case this PR adds), a server can send more than 64 notifications before replying. `enqueueNotification` discards the oldest message, so a `textDocument/publishDiagnostics` for a different URI can be lost permanently. The corresponding `session.waitForDiagnostics` then times out and `Manager.Check` returns no diagnostics even though the server published them. This is the resolved CodeRabbit request in a different form: the follow-up commit trades the read-loop deadlock for silent protocol-event loss, and `TestClientNotificationOverflowDropsOldestInOrder` explicitly codifies that loss. Use a lossless non-blocking queue (or another explicit, observable policy that preserves required diagnostics) and add an overflow regression test that asserts delivery of all relevant notifications.

PierrunoYT:
Pushed `4b3d17b` (on a merge of current `main`).

**[P1] Preserve notifications when the dispatch queue overflows** — you read it exactly right: the follow-up commit traded the read-loop deadlock for silent protocol-event loss, and `TestClientNotificationOverflowDropsOldestInOrder` codified that loss as intended behaviour. The queue is now **lossless and unbounded**: the read loop still never blocks (so a handler can make a re-entrant `Client.Call`), and nothing is discarded.

The two alternatives are both worse, and the code now says so where the decision lives:

- Blocking on a full buffer is the original deadlock — the handler waits for a response frame the blocked reader can no longer deliver.
- Dropping loses state the protocol never repeats. A `textDocument/publishDiagnostics` is the server's only report for that URI, so a dropped one makes `session.waitForDiagnostics` time out and `Manager.Check` return nothing even though the server published findings.

On the growth question that motivated the bound: it is bounded in practice by how much the server emits while one handler runs, and `dequeueNotification` now releases the backing array once the queue drains, so a burst does not leave its capacity (or its consumed items) alive afterwards. A handler that never returns would grow it without limit — that is a handler bug queue pressure makes visible, not something the transport should hide by discarding messages.

**Tests, and their teeth**

- `TestClientNotificationQueueIsLossless` replaces the drop-oldest test: 512 enqueued (8× the old bound), all 512 dequeued in FIFO order, and the drained queue asserted back to nil capacity.
- The end-to-end `TestClientNotificationBurstDoesNotBlockResponse` now asserts **delivery**, not just non-blocking: with a handler parked in a re-entrant `Call`, the server sends 512 notifications plus the response; the test requires the response to arrive *and* all 512 notifications to reach the handler in order. That is the "asserts delivery of all relevant notifications" case you asked for, and it also covers @copilot's burst/pressure gap.

I verified both fail under the old policy rather than assuming it — reintroducing the 64-item drop-oldest cap gives:

```
--- FAIL: TestClientNotificationBurstDoesNotBlockResponse
    received 64 of 512 burst notifications; the queue lost messages
--- FAIL: TestClientNotificationQueueIsLossless
    notification = "notification-448", want "notification-000" (queue must stay FIFO)
```

**@coderabbitai on the race detector** — I cannot run it on this host and want to be straight about that rather than tick the box: `go test -race` requires cgo, and there is no C toolchain here (`-race requires cgo` → `cgo: C compiler "gcc" not found`). The concurrent tests pass without it, and the repo's race-enabled CI job is what will actually exercise `notificationLoop` against `Client.Call`. If it reports anything I will follow up immediately. The synchronisation itself is narrow: `notifyMu` guards only the slice, the handler runs on the worker goroutine, and the wake-up channel has capacity 1 with a drain-until-empty loop, so a signal dropped by the `default` branch is always covered by the pending token.

**Validation** (Windows host, Go 1.26.5): `go build ./...`, `go vet ./...`, `go test ./... -count=1` all pass; `GOOS=linux|darwin go vet ./internal/lsp` clean; `gofmt` clean on both changed files.

@jatmn — ready for another look. I can't use the reviewer-request button on this repo, hence the mention.

@coderabbitai review

PierrunoYT:
Fixed in `d8ef0f3` — good catch, and it was a real defect I introduced by removing the bound: with an unbounded queue, a closed client whose reader is still open retains everything the server sends.

**[P1] Do not retain notifications after `Client.Close`**

- `failPending`'s `closeOnce` now calls `closeNotifications`, which sets `notifyClosed` and drops the queue (`notifyQueue = nil`).
- `enqueueNotification` returns early when `notifyClosed` is set instead of appending.

Both run under `notifyMu`, which is what makes the race you flagged safe: an enqueue concurrent with shutdown either observes the flag and drops its item, or appends just *before* the flag is set and has that item cleared by `closeNotifications` a moment later. Either way nothing survives close.

Queued-but-unhandled notifications are dropped rather than drained, and the code says why: the worker loop is gone by definition of close, and the caller that needs diagnostics (`Manager.Check`) collects them before shutting the client down. The comment on `enqueueNotification` names the concrete trigger you identified — `Server.Shutdown` calls `Client.Close()` at `server.go:66` and only then `s.stdin.Close()`, so the read loop keeps feeding frames in between.

**Regression test** — `TestClientDropsNotificationsAfterClose` does exactly what you asked: it keeps the transport writable after `Close` (mirroring that window in `Server.Shutdown`), sends 64 notification frames, and requires that none reach the handler and that `notifyQueue` stays empty while `notifyClosed` is set. It also checks a pre-`Close` notification still works, so the test cannot pass by breaking dispatch outright.

Verified it has teeth by disabling the guard:

```
--- FAIL: TestClientDropsNotificationsAfterClose
    closed client retained 64 queued notifications
```

`go build ./...`, `go vet ./...`, `go test ./... -count=1` pass; `GOOS=linux|darwin go vet ./internal/lsp` clean; `gofmt` clean. The `-race` caveat from my previous comment still stands — no C toolchain on this host, so the race-enabled CI job is what exercises `notificationLoop` against `Client.Call` and this new close path.

@coderabbitai review

jatmn (changes requested):

## Findings

- [P1] Preserve the diagnostics receipt boundary when dispatching asynchronously
  `internal/lsp/client.go:181`
  `Manager.Check` captures `publishBaseline` before it sends the next document version, but `publishCount` is incremented only when the queued callback eventually reaches `session.handleNotification`. A publish for the previous version can therefore already be in `notifyQueue` when a later `Check` takes its baseline; when that old callback is eventually run, it looks newer than the baseline and satisfies `waitForDiagnostics`. Since version-less `publishDiagnostics` messages are intentionally accepted, the stale result is then returned for the new text (and the 300ms debounce can finish before the real response arrives). The synchronous reader never had this receipt-vs-dispatch gap. Carry a receive sequence/timestamp through the queue and baseline against receipt, or otherwise ensure pre-sync queued diagnostics cannot satisfy the new check; add a regression for queued version-less diagnostics.

- [P2] Put a failure policy on the lossless notification backlog
  `internal/lsp/client.go:229`
  The lossless, non-blocking design correctly avoids the old deadlock, but it replaces pipe backpressure with an unbounded slice of full `json.RawMessage` payloads. A language server can keep emitting notifications faster than the single handler drains them (or while a re-entrant handler is waiting for a response), and `readLoop` will parse and retain every frame until Zero exhausts its heap. This does not require a permanently stuck handler; any sustained producer faster than the consumer is enough. The claimed bounded queue and the 512-message test do not cover this case. Define an observable overload policy—such as failing/evicting the LSP client at a high-water mark—rather than retaining unbounded protocol input.

PierrunoYT:
Pushed `6659ee0`, addressing both findings from the latest round (one P1, one P2).

**[P1] Preserve the diagnostics receipt boundary when dispatching asynchronously** (`internal/lsp/client.go:181`)

`Manager.Check` captured its baseline from `publishCount`, which only advances once `session.handleNotification` actually *runs* for a URI — but dispatch happens off a queue now, so a publish for a version Check is about to supersede could already be sitting received-but-undispatched at the moment baseline is captured, then run afterward and look "newer than baseline" purely because handling lagged receipt. Since many servers omit the version field, `handleNotification`'s own staleness check (positive-version only) couldn't catch this either — the stale result would reach the caller for the new text, and the 300ms debounce could finish before the real response arrived.

Fixed by stamping every notification with a receipt sequence *at enqueue time* (`Client.NotificationSeq`), carrying it through the queue to the handler, and baselining `waitForDiagnostics` against receipt order instead of processed-count. A notification already sitting in the queue when a baseline is captured now has `seq <= baseline` no matter when it's actually dispatched.

Added `TestPublishBaselineRejectsAlreadyQueuedPublish`, which reproduces the exact interleaving deterministically (enqueue the stale publish → capture baseline → dispatch it) via the same bare-`Client` technique the existing lossless-queue test uses (no goroutine timing involved). Before writing it, I reproduced the bug directly against the pre-fix code to confirm it's real (not hypothetical) — it is: the stale, receipt-before-baseline publish incorrectly satisfied `waitForDiagnostics` every time.

**[P2] Put a failure policy on the lossless notification backlog** (`internal/lsp/client.go:229`)

The lossless design correctly avoided the old deadlock but replaced pipe backpressure with an unbounded slice of full `json.RawMessage` payloads — any sustained producer faster than the single consuming handler (not just a permanently-stuck one) would grow it until the heap gives out. The 512-message burst test never exercised this.

Added `notifyQueueLimit` (4096): exceeding it fails and closes the client observably (`IsClosed()` becomes true), letting `Manager.sessionFor`'s existing dead-session eviction take over on next use — the same recovery path a crashed server already goes through — instead of retaining unbounded protocol input.

Added `TestClientFailsOnNotificationBacklogOverload`, confirming the client closes exactly at the limit and the queue is released, not left to grow past it.

Both regression tests verified to have teeth: the receipt/baseline race was proven against the actual pre-fix code before being ported to the fixed API, and the overload test was confirmed to fail with the limit check temporarily removed.

**Validation:**
- `go test ./internal/lsp/... -count=30` on Windows: pass.
- `go test -race ./internal/lsp/... -count=10` via WSL: pass, plus `go build`/`go vet ./...` under WSL — clean.

jatmn (changes requested):

## Findings

- [P2] Bound retained notification bytes, not just notification count
  `internal/lsp/client.go:263`
  `notifyQueueLimit` permits 4,096 complete peer-controlled `json.RawMessage` payloads to accumulate while a handler is blocked. A language server sending 1 MiB diagnostics or custom notifications can therefore retain several GiB before this count-based limit closes the client, which defeats the new overload policy and can OOM Zero. Apply a byte budget or reject oversized frames before retaining the backlog.

- [P2] Wake a diagnostic wait when overload closes the client
  `internal/lsp/documents.go:169`
  The 4,097th queued notification calls `failPending`, but `waitForDiagnostics` only waits on its publish channel or the caller context. The overflow clears the queue without waking that waiter, so a `Manager.Check(context.Background(), ...)` that is already waiting blocks forever (and the normal diagnostics path waits its full timeout) even though the client has been deliberately failed. Observe `client.closed` or otherwise release the waiters when closing the client.

- [P2] Restore the delayed PID-file regression coverage dropped by the stale merge
  `internal/config/command_test.go:151`
  Relative to the current base, this PR deletes `TestAssertProcessTerminatedWaitsForDelayedPIDFile` and its polling helper, which were added by main's `#802` to cover atomic PID-file publication after the first `ENOENT`. The remaining helper reads immediately, so the Windows startup-race coverage is lost even though this LSP change has no config-test scope. Rebase onto the current base or restore that test/helper.

PierrunoYT:
Final follow-up on current head `8a82da1`:

- bounded retained notification payloads to 16 MiB in addition to the count limit
- releases byte accounting on dequeue and close
- diagnostic waits now wake promptly on client failure while preserving an already-received fresh publish
- merged current upstream `main`, restoring unrelated base regression coverage
- all review threads are resolved; race tests, full validation, and all GitHub checks pass

The PR is mergeable. PierrunoYT requests a fresh review to supersede the previous `CHANGES_REQUESTED` decision.

PierrunoYT:
Pushed `2dedb2d` after a normal merge of current `upstream/main` (`f657c91`).

Fresh review of the merged concurrency paths confirmed the requested fixes remain intact:

- notification overload is bounded by both 4,096 messages and 16 MiB of retained payload bytes, with accounting released on dequeue and close
- diagnostic waiters observe client closure and are removed promptly
- receipt sequences are stamped at enqueue time, so a publish already queued before a sync cannot satisfy the later diagnostic wait
- current upstream's delayed PID-file regression (`TestAssertProcessTerminatedWaitsForDelayedPIDFile`) is preserved unchanged

The follow-up commit also closes a narrow receipt/closure ordering race: when a fresh publish and client closure make both waiter signals ready together, `waitForDiagnostics` now reconciles against the recorded `publishSeq` instead of letting random `select` order discard diagnostics already handled before closure. `TestWaitForDiagnosticsPreservesPublishHandledBeforeClientCloses` stresses that simultaneous-ready case 100 times.

Local validation on Go 1.26.5 / linux-amd64:

- `go test -race ./internal/lsp/... -count=10` — pass
- `go test -race ./internal/config -run TestAssertProcessTerminatedWaitsForDelayedPIDFile -count=1` — pass
- `make fmt-check` — pass
- `go vet ./...` — pass
- `go test ./... -count=1` — pass
- `go run ./cmd/zero-release build` — pass
- `go run ./cmd/zero-release smoke` — pass
- `make lint-static` — pass (0 issues)
- `make vulncheck` — pass (no vulnerabilities)
- `git diff upstream/main...HEAD --check` — pass

All inline review threads remain resolved. GitHub checks are running on the new head. PierrunoYT requests a fresh review from @jatmn to supersede the stale `CHANGES_REQUESTED` decision.

PierrunoYT:
Follow-up on `2dedb2d`: all GitHub checks are now green (Ubuntu/macOS/Windows smoke, performance smoke, security & code health, Zero Review, and CodeRabbit). CodeRabbit approved the head with no actionable comments. The only remaining blocker is the stale maintainer `CHANGES_REQUESTED` decision; PierrunoYT has requested a fresh maintainer review.

jatmn (changes requested):

## Findings

- [P1] Do not drop already-received notifications on client teardown
  `internal/lsp/client.go:337`
  After async dispatch, `readLoop` can enqueue `textDocument/publishDiagnostics` and then hit EOF, backlog overload, or `Close`. `failPending` closes `c.closed` and `closeNotifications` nils `notifyQueue` while `notificationLoop` exits without draining, so publishes already read off the wire never reach `handleNotification`. An in-flight `Manager.Check` then sees `waitForDiagnostics` return false (even with the `2dedb2d` `publishSeq` recheck, because the handler never ran — and a dequeued-but-unfinished handler can still lose the same race). Pre-change, the handler ran inline before the next read, so a trailing publish was applied before teardown. The `closeNotifications` comment assumes diagnostics were collected before shutdown; that does not hold for unexpected failure mid-Check. Deliver or drain already-queued (and in-flight) notifications before tearing the worker down on transport/overload failure — intentional rejection of frames after `notifyClosed` can remain — and add a regression that enqueues a publish then fails the reader before the worker runs.

- [P2] Recheck `publishSeq` on wait context cancellation the same way as client close
  `internal/lsp/documents.go:171`
  In the `seq <= baseline` select, `case <-s.client.closed` now re-reads `publishSeq` so a dual-ready wake cannot discard an already-applied publish, but adjacent `case <-ctx.Done()` still returns false unconditionally. When `handleNotification` updates `publishSeq` and closes the waiter in the same window the caller deadline fires, `select` can take `ctx.Done()` and `Manager.Check` returns empty diagnostics despite the fresh result already in the session cache. Please apply the same recheck (or equivalent) on the context path, with a regression for deadline+waiter dual-ready.

jatmn (changes requested):

## Findings

- [P1] Do not drop already-received notifications on client teardown
  `internal/lsp/client.go:337`
  Async dispatch introduced a real regression here. Before this PR, `readLoop` invoked the handler inline, so a `textDocument/publishDiagnostics` frame read off the wire was applied before the next read could hit EOF, backlog overload, or `Close`. Now `readLoop` enqueues the frame, `failPending` closes `c.closed`, and `closeNotifications` nils `notifyQueue` while `notificationLoop` can exit without draining — including when both `<-c.closed` and `<-notifyReady` are ready and `select` picks closure. The publish never reaches `handleNotification`, so an in-flight `Manager.Check` sees `waitForDiagnostics` return false (the `2dedb2d` `publishSeq` recheck cannot help when the handler never ran; a dequeued-but-unfinished handler can still lose the same race) and degrades to `(nil, nil)` even though the server already sent findings.

  Rejecting frames enqueued after `notifyClosed` is fine for intentional shutdown. The gap is unexpected failure mid-`Check`: transport EOF, overload, or concurrent `Server.Shutdown` while diagnostics are already received but not yet dispatched. `closeNotifications` currently treats those the same as post-close spam. Please drain or deliver already-queued (and in-flight) notifications before tearing the worker down on transport/overload failure, keep post-`notifyClosed` rejection as-is, and add a regression through `NewClient` + `notificationLoop` that enqueues a publish then fails the reader before the worker runs.

- [P2] Finish the dual-ready `publishSeq` recheck on context cancellation
  `internal/lsp/documents.go:171`
  `2dedb2d` correctly fixed one half of a `select` race: in the `seq <= baseline` branch, `case <-s.client.closed` now re-reads `publishSeq` so a publish handled just before closure is not discarded when `select` picks closure. The adjacent `case <-ctx.Done()` still returns false unconditionally. When `handleNotification` updates `publishSeq` and closes the waiter in the same window the caller deadline fires, `select` can take `ctx.Done()` and `Manager.Check` returns empty diagnostics despite the fresh result already in the session cache. The race class predates this PR, but leaving the context arm unfixed after adding the close arm is an incomplete application of the same fix. Please mirror the `client.closed` recheck on `ctx.Done()` (or equivalent), with a regression for deadline+waiter dual-ready the way `TestWaitForDiagnosticsPreservesPublishHandledBeforeClientCloses` covers closure.

PierrunoYT:
PierrunoYT addressed both remaining notification teardown races in `ce95206`.

- Accepted queued and in-flight notifications drain before the worker exits; post-close notifications remain rejected.
- `Close`/transport failure remain asynchronous and cannot deadlock on a blocked handler.
- Diagnostic waits account for notification-drain completion and recheck publishes when closure or cancellation wins a dual-ready select.
- Added deterministic EOF/drain and deadline/publish regressions.

Validation: full `internal/lsp` suite passes with `-race`; formatting and diff checks pass. Ready for re-review.

jatmn (changes requested):

## Findings

- [P1] Stamp receipt sequence when the frame is read, not when it is enqueued
  `internal/lsp/client.go:208`
  Async dispatch introduced a window synchronous `readLoop` did not have: `readMessage` consumes the frame, then `json.Unmarshal` and `enqueueNotification` run later on the same goroutine while `Manager.Check` can call `publishBaseline()` on another. A delayed `textDocument/publishDiagnostics` for superseded text can therefore be assigned `seq > baseline` even though it was already on the wire before the new sync. Version-less publishes then pass `waitForDiagnostics` and reach the caller with stale findings — the same receipt-boundary class `TestPublishBaselineRejectsAlreadyQueuedPublish` covers for queue ordering, but via read/enqueue reordering. Stamp sequence (or otherwise pin the receipt boundary) at read time, or hold `notifyMu` across read-to-enqueue for notifications.

- [P2] Finish the notification-drain logic consistently
  `internal/lsp/documents.go:171`
  `ce95206` added `notificationsDone()` draining for `client.closed` in the `seq <= baseline` wait, but the same concern is only partly applied elsewhere. (1) The `ctx.Done()` arm at line 171 never waits on drain and only re-checks handled `publishSeq`; if cancellation wins while a `publishDiagnostics` frame is already accepted but still queued, `Manager.Check` can return `(nil, nil)` despite wire receipt, and the inner drain `select` at line 186 can still let `ctx.Done()` abort before the worker finishes. (2) The debounce-phase `client.closed` arm at line 214 returns `true` immediately without drain or re-check, so `Check` can read an earlier handled publish from `diagnosticsFor` while a newer accepted publish is still queued. Please apply the same drain-then-recheck pattern on both paths, with regressions for bounded-context cancellation and for debounce-plus-closure with a follow-up publish still in the queue.

- [P3] Use separate polling deadlines in EOF/drain tests
  `internal/lsp/manager_test.go:457`
  `TestWaitForDiagnosticsDrainsAcceptedPublishAfterTransportEOF` and `TestClientDrainsAcceptedNotificationsAfterTransportEOF` (`internal/lsp/client_test.go:482`) reuse one one-second deadline for waiting on `NotificationSeq()` and then for `client.IsClosed()`. Under slow CI scheduling the first loop can consume most of the budget and make the second loop fail intermittently. CodeRabbit's inline request on this is still valid — reset the deadline before each wait loop.

- [P3] Cap `Content-Length` before allocating the frame body
  `internal/lsp/client.go:457`
  `notifyQueueByteLimit` is enforced only in `enqueueNotification`, but `readMessage` already allocates the full body first. A single oversized notification can therefore bypass the new byte policy before `failPending` runs. This is a gap in the overload feature added here, not a regression in the read path itself. Reject or cap `Content-Length` before `make([]byte, contentLength)` if you want the byte budget to be authoritative.

## Follow-up (not blocking this PR)

These are real hardening opportunities, but they predate or sit outside the core deadlock/receipt fixes and I would not block merge on them alone.

- **Baseline before `sync`** (`internal/lsp/manager.go:76`): `publishBaseline()` has run before `sync()` on `main` as well. A version-less publish received in that gap can still satisfy a later wait. Moving the baseline snapshot to after `sync` (or tying freshness to committed document version) would narrow that window; it is not a regression introduced by async dispatch.
- **Read-then-close before enqueue** (`internal/lsp/client.go:208`): a frame decoded just before `notifyClosed` is set can be dropped at enqueue. The window is very narrow and mostly overlaps intentional post-`Close` rejection during `Server.Shutdown`; worth hardening only if you want wire-consumed frames to be lossless up to the close boundary.

PierrunoYT:
Addressed all four findings in `b0722ff`.

**[P1] Receipt is stamped at read time.** `readLoop` now calls `stampReceipt()` on every frame the moment `readMessage` returns — before `json.Unmarshal`, before the enqueue — and passes that seq into `enqueueNotification`, which no longer stamps. The expensive, peer-sized part of the pipeline is now outside the boundary. Frames that turn out not to be notifications consume a number too, so I renamed `NotificationSeq` → `ReceiptSeq`: it is a receipt clock over frames, and gaps in it are meaningless to the only thing that reads it (`publishSeq[uri] > baseline`).

Regression: `TestPublishBaselineRejectsFrameReadBeforeBaseline` drives the seam directly — stamp (frame leaves the wire), baseline (concurrent `Check`), then enqueue + dispatch (read loop finishes decoding). With stamping at enqueue it fails; with stamping at read the stale publish lands at `seq == baseline` and is rejected.

**[P2] The drain logic is uniform now.** Every exit from `waitForDiagnostics` goes through `catchUpNotifications()` before the caller reads `diagnosticsFor` — the two `seq <= baseline` arms became identical and now share `freshAfterCatchUp`. Catch-up is "has the worker handled every notification the reader had already accepted", tracked with `acceptedSeq`/`handledSeq` on the client plus a broadcast channel that pulses as the worker advances.

The two policies differ deliberately:
- **Closed client:** wait for `notificationsDone()` unconditionally, no longer racing `ctx.Done()`. After closure admission is shut and a handler's re-entrant `Call` fails immediately, so the drain is bounded and it is the last chance those frames get.
- **Live client:** bounded by `notificationCatchUpGrace` (50ms). I did not make this one unbounded: on a live client the worker can legitimately sit in a handler's re-entrant `Call` awaiting a server response, and an unbounded wait there would hang `Check` well past a deadline the caller has already blown. The grace only applies when the worker is actually behind — with the session's own parse-and-record handler it is virtually always caught up, so the common path costs one comparison. It is a `session` field so tests can pin it.

Regressions: `TestWaitForDiagnosticsCatchesUpOnContextCancellation` (bounded-context cancellation with a publish queued behind a busy worker — previously `false`, i.e. `Check` returning `(nil, nil)` for text the server had answered) and `TestWaitForDiagnosticsDrainsFollowUpPublishOnCloseDuringDebounce` (debounce + closure with a newer publish still queued — previously returned the older `"first"`; it also cancels the ctx mid-drain to pin that the drain no longer aborts on it). Both fail against the prior behavior.

**[P3] `Content-Length` is capped before allocation.** `maxFrameBytes = notifyQueueByteLimit`, rejected in `readMessage` before `make([]byte, contentLength)`, which is what makes the byte budget authoritative — a notification can no longer be materialized in full and only then measured against a limit it already blew past. Covered by `TestReadMessageRejectsOversizedFrame`. Worth flagging the tradeoff: this also caps *responses* at 16 MiB, so a hypothetical larger response now fails the client instead of being read. A legitimate LSP frame is orders of magnitude below that, but say the word if you'd rather have a separate, looser transport ceiling.

**[P3] Separate polling deadlines.** Both EOF/drain tests (`manager_test.go`, `client_test.go`) now reset the deadline before the `IsClosed()` loop so slow scheduling in the first loop can't consume the second one's budget.

I left the two follow-ups alone as non-blocking, per your note.

**Validation**
- `go build ./...`, `go vet ./...`
- `go test ./internal/lsp -count=3`
- `go test -race ./internal/lsp -count=2` (WSL/Linux; `-race` needs cgo, unavailable on this host)
- `go test ./...`
- Each new test verified to fail against the pre-fix behavior
- `gofmt -l internal/lsp` clean, `git diff --check` clean

jatmn (changes requested):

## Findings

- [P2] `catchUpNotifications` can return before all accepted notifications are handled
  `internal/lsp/documents.go:268`
  `b0722ff` made every `waitForDiagnostics` exit call `catchUpNotifications()` so `Manager.Check` reads diagnostics only after the worker has applied everything already accepted. On a live client the helper snapshots `acceptedNotificationSeq()` once, then returns as soon as `handledThrough() >= target`. While it is waiting, `readLoop` can still accept another `publishDiagnostics` (or any notification) with a higher receipt seq; when the worker finishes the stale target, `handled >= target` becomes true and the function returns even though the newer item is still queued. Debounce/timer/`ctx.Done()` success arms then return `true` and `diagnosticsFor` can answer with an older publish while a newer accepted publish for the same URI is still pending — the same “newest received, not the older one that woke the wait” failure the closed-client drain regression was written for, but on the live path. This is PR-owned: `main` had no catch-up helper because the handler ran inline; async dispatch created the gap and this snapshot leaves it open even when grace has not expired. Please loop until `handledSeq` catches the current `acceptedNotificationSeq()` (or equivalent), not a one-shot snapshot, and add a live-client regression where a follow-up publish is accepted during catch-up behind a busy worker.

## Follow-up (not blocking this PR)

- **Stamp→enqueue seam in catch-up** (`internal/lsp/documents.go:268`, `internal/lsp/client.go:228`): `catchUpNotifications` waits on `acceptedNotificationSeq()` while baselines use `ReceiptSeq()` stamped before decode/enqueue. A `publishDiagnostics` frame can sit in the narrow unmarshal window while catch-up thinks the worker is caught up. The window is much smaller than the enqueue/baseline race `b0722ff` fixed and overlaps the read-then-close hardening you already listed as non-blocking; worth tightening only if you want wire-to-handler losslessness end to end.

- **Production grace vs cancellation test** (`internal/lsp/manager_test.go:572`, `internal/lsp/documents.go:168`): `TestWaitForDiagnosticsCatchesUpOnContextCancellation` uses a five-second `catchUpGrace`; production sessions use `50ms`. The test therefore does not exercise the live-client timeout path. That timeout is intentional — unbounded catch-up on a live client would hang `Check` past an expired caller deadline — so `(nil, nil)` after grace under sustained worker backlog is a documented tradeoff, not an oversight. A production-grace regression would still be useful if you want the contract spelled out in tests.

PierrunoYT:
Addressed in `dd0a7fa`, plus the optional grace test.

**[P2] Catch-up follows the target instead of snapshotting it.** You're right, and it is the same bug I wrote the closed-client drain regression for — I just left the live path holding a stale number. The target is now re-read every pass. It cannot spin: the live-client grace already caps the wait however fast the server publishes, and a closed client still drains unconditionally because its backlog cannot grow.

**The regression took two attempts, and the first one was wrong in a way worth describing.** My initial version asserted `diagnosticsFor == "second"` after catch-up returned — and it passed *with the bug restored*, because the worker records the follow-up publish microseconds later and won the race against the test's read. A test that passes either way is worse than no test.

The version that landed pins the state instead of racing it: the handler for "second" blocks, so at the moment the stale target is satisfied the worker is provably inside that handler with the newer publish unrecorded. The assertion is then that catch-up has **not** returned there. Against the snapshot implementation it fails with `catch-up returned on a stale target; diagnostics = [... "first"]`, which is precisely the failure you described.

**Optional: production-grace regression.** Taken. `TestCatchUpNotificationsGivesUpAfterTheLiveGrace` holds the worker permanently behind and asserts catch-up returns after — and not before — its grace. It uses 25ms rather than the 5s the other tests pin, so the timeout path is actually exercised. That spells out the tradeoff in tests: an unbounded live-client wait would hang `Check` past a deadline the caller has already blown, so a permanently-behind worker costs the grace and no more.

**Not taken:** the stamp→enqueue seam in catch-up. Agreed it is real, but it is a strictly smaller window than the one `b0722ff` closed and it overlaps the read-then-close hardening already listed as non-blocking — I would rather do wire-to-handler losslessness as one deliberate change than shave one edge of it here.

**Validation:** `go build ./...`, `go vet ./...`, `go test ./...`, `go test ./internal/lsp -count=3`, and `go test -race ./internal/lsp -count=2` under WSL/Linux (`-race` needs cgo, unavailable on this host). `git diff --check` and `gofmt -l` clean.

jatmn (changes requested):

## Findings

- [P2] Required macOS smoke check is failing on head `dd0a7fa`
  `.github/workflows/ci.yml` (Smoke `macos-latest`)
  GitHub Actions run [30453340514](https://github.com/Gitlawb/zero/actions/runs/30453340514) reports `Smoke (macos-latest)` **failure** on commit `dd0a7fa`; Ubuntu, Windows, Performance Smoke, Security & code health, and Zero Review on the same head all passed. The job runs `go test ./...`, and this PR’s only production changes are in `internal/lsp`, so the failure is PR-owned even though the exact failing test could not be confirmed here — job logs require repository authentication and were not available in this review environment. Please reproduce on `macos-latest`, identify the failing test from the job log, fix the underlying defect or stabilize the test, and re-run the full matrix before merge.

## Follow-up (not blocking this PR)

- **`dd0a7fa` catch-up re-read looks responsive to the prior P2 snapshot finding.** `catchUpNotifications` now re-reads `acceptedNotificationSeq()` each pass (`internal/lsp/documents.go:271-281`), which addresses the enqueued-but-still-behind case from `b0722ff`. I do not have a remaining actionable defect against that specific request on current head.

- **Stamp→enqueue seam in catch-up** (`internal/lsp/documents.go:279`, `internal/lsp/client.go:228-240`): `ReceiptSeq()` advances at read while `acceptedNotificationSeq()` advances at enqueue, so catch-up can theoretically return while a frame is still in the read loop’s unmarshal window. That is a much narrower window than the queue/backlog race `dd0a7fa` fixed, and it was already listed as non-blocking follow-up on the prior round. Worth hardening only if you want wire-to-handler losslessness end to end.

- **Live-client grace tradeoff** (`internal/lsp/documents.go:168`, `:287-288`): production `catchUpGrace` is 50ms and catch-up deliberately gives up afterward; debounce success arms then return `true` without a post-catch-up `publishSeq` re-check like `freshAfterCatchUp` uses on the pre-baseline cancel/close paths. That asymmetry is intentional — unbounded live catch-up would hang `Check` past caller deadlines — and is spelled out in `TestCatchUpNotificationsGivesUpAfterTheLiveGrace`. Not a merge blocker unless you want to change that contract.

- **Baseline before `sync`** (`internal/lsp/manager.go:76`): `publishBaseline()` still runs before `sync()`; a version-less publish arriving in that gap can satisfy a later wait. That window predates async dispatch.

- **Closed-client drain with wedged third-party handler** (`internal/lsp/documents.go:265`): post-close catch-up waits unbounded on `notificationDrained`. Production uses the session’s fast handler; exported `SetNotificationHandler` could still wedge `Manager.Check` if a handler blocks off `Client.Call`.

- **Stale API name in comment** (`internal/lsp/documents.go:32`): `publishSeq` still references `Client.NotificationSeq`; the clock is `Client.ReceiptSeq()`.

PierrunoYT:
PierrunoYT addressed the macOS smoke failure in `7f09894`.

The failing regression waited for `ReceiptSeq()`, which advances as soon as a frame leaves the wire, before JSON decoding and notification enqueue. On slower macOS scheduling, the test could release the first handler before the follow-up notification was actually accepted, allowing catch-up to correctly return against the current accepted target. The test now waits for `acceptedNotificationSeq()` so it deterministically establishes the queue state it intends to verify.

Also updated the stale `Client.NotificationSeq` comment reference to `Client.ReceiptSeq`.

Validation:
- `TestCatchUpNotificationsFollowsPublishesAcceptedWhileWaiting`, 100 consecutive runs
- `go test ./internal/lsp -count=3`
- `go test ./...`
- `go vet ./...`
- release build and smoke test
- static lint: 0 issues
- govulncheck: no vulnerabilities
- touched-file formatting and `git diff --check`

Ready for CI and maintainer re-review.

jatmn (dismissed):
@Vasanthdev2004 LGTM

Vasanthdev2004 (dismissed):
Approving. I re-verified on current head (`7f09894`) rather than standing on the earlier rounds, and the concurrency story holds up under everything I threw at it.

I checked each guarantee by mutating the production code and confirming the suite goes red:

- **Async dispatch is load-bearing.** Putting the handler back inline in `readLoop` (`internal/lsp/client.go:240`) reproduces #758 exactly — the run hangs and the goroutine dump shows `readLoop` parked in the handler at `client.go:244` with the test goroutine blocked in `writeMessage`. `TestClientDrainsAcceptedNotificationsAfterTransportEOF` catches it.
- **FIFO ordering.** Flipping `dequeueNotification` to pop the tail fails three tests, including order across the 512-message burst.
- **Both backlog limits are real.** Dropping either the count check or the byte check in `enqueueNotification` fails its own test; neither is redundant.
- **Oversized frames.** Removing the `maxFrameBytes` check fails `TestReadMessageRejectsOversizedFrame`.
- **Post-close admission.** Removing the `notifyClosed` guard fails `TestClientDropsNotificationsAfterClose`.
- **Closed-client drain.** Dropping `<-drained` in `catchUpNotifications` fails two manager tests.
- **The `dd0a7fa` catch-up re-read.** Restoring the one-shot snapshot fails `TestCatchUpNotificationsFollowsPublishesAcceptedWhileWaiting`.
- **`markHandled` after the handler.** Moving it before the handler call fails two tests.
- **The `<-s.client.closed` wakeups in `waitForDiagnostics`.** Removing them fails two tests.

Worth calling out separately: `readMessage` on main did `make([]byte, contentLength)` straight off a peer-supplied header with no ceiling. The 16 MiB cap closes an unbounded-allocation path that predates this PR.

On the macOS failure from the `dd0a7fa` round — I couldn't reproduce anything like it. `go test ./internal/lsp -count=8` is clean, and `-count=6` under 3× CPU oversubscription is also clean (285s vs 27s unloaded), which is the shape of load that usually shakes out the timing-sensitive tests here. CI is green on head across all five jobs.

One gap, which is why I'm writing this up rather than just approving silently:

**`TestPublishBaselineRejectsFrameReadBeforeBaseline` doesn't pin the fix it's named for** — `internal/lsp/manager_test.go:336`. The test calls `client.stampReceipt()` itself and never drives `readLoop`, so what it actually asserts is the seq comparison, not *where* the stamp is taken. I moved the stamp from before the `json.Unmarshal` (`client.go:228`) down into the enqueue call site — i.e. reverted `b0722ff` — and the entire package stays green. The P1 from round three can be silently reverted by anyone refactoring `readLoop`, with no test to stop them.

There's a deterministic way to pin it with no timing dependence: under read-time stamping, a frame that is *not* a notification still consumes a receipt number ("gaps in it are fine", per your own comment); under enqueue-time stamping it doesn't. So:

```go
client := NewClient(clientReader, io.Discard)
// A response frame: read off the wire, but it never reaches enqueueNotification.
writeMessage(serverWriter, map[string]any{"jsonrpc": "2.0", "id": 1, "result": nil})
// Poll briefly — ReceiptSeq() must become > 0.
```

I ran that against both variants: passes on head, fails with `receiptSeq = 0` as soon as the stamp moves to enqueue.

Not blocking the merge — the production code is correct, this is only about keeping it that way. Fine by me either in this PR or as a follow-up.

Two things I'm deliberately not re-raising: the unbounded `<-drained` for a closed client with a third-party `SetNotificationHandler` (`documents.go:265`) — the only in-repo handler is `session.handleNotification`, which parses, records and closes waiter channels, none of which can block, and both jatmn and CodeRabbit have already logged it; and the `Client.NotificationSeq` comment at `documents.go:32`, which is already fixed on head.

jatmn (approved):
LGTM

Vasanthdev2004 (approved):
Re-approving on `9fc9ff1`. The only change since my last pass is one test, and it's a good one — it closes a gap in my own review rather than padding coverage.

I checked the claim in its comment rather than taking it: moved `stampReceipt` from read time back to the enqueue call site (the exact regression it names) and ran both tests. `TestReceiptStampedAtReadNotEnqueue` fails; `TestPublishBaselineRejectsFrameReadBeforeBaseline` still passes. That's the whole point — the old test drives `stampReceipt` itself and never exercises `readLoop`, so it pins the comparison but not where the stamp is taken. The new one uses a response frame, which is read but never enqueued, so it can only pass under read-time stamping.

Package green on the current head. Everything I verified last round still holds.

kevincodex1 (approved):

PR #761, opened 2026-07-19
Add transcript-grounded structured task state

anandh8x:
## Summary

- add a deterministic per-run TaskState projection for objective, plan, tool results, changed files, verification, and completion
- emit bounded, content-free aggregate snapshots through the existing trace recorder
- preserve the immutable objective across repeated compaction and ground completion checks in that objective
- use structured plan state only when plan parity matches the transcript; mismatches retain transcript-derived completion behavior and omit uncorroborated mutable compact state

## Scope and safety

- no new dependency, model call, tool, database, UI, or autonomous execution behavior
- trace events exclude objective text, objective fingerprints, plan text, tool output, and file paths
- task-state trace history is coalesced and capped while retaining the latest aggregate
- compact objective context is capped at 512 bytes
- plan parsing now matches update_plan rejection and normalization behavior

## Validation

- make fmt-check
- go vet ./...
- go test ./...
- go test -race ./internal/agent ./internal/trace
- govulncheck ./... (no vulnerabilities found)
- git diff HEAD --check

The pinned advisory golangci-lint command reports only unrelated pre-existing repository findings; no finding points to a file changed by this PR.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -56,14 +56,10 @@ type CompactionOptions struct {
 	// injected so Compact stays pure and testable; the agent loop wires it to a
 	// real provider call.
 	Summarize func(toSummarize []zeroruntime.Message) (string, error)
-	// taskState is a transcript-corroborated snapshot supplied by the running
-	// agent. It is unexported so standalone callers cannot accidentally promote
-	// structured state above the messages they are compacting.
+	// taskState is a snapshot supplied by the running agent. Its immutable
+	// objective is always preserved; mutable fields are admitted only when its
+	// plan projection still matches the transcript.
 	taskState *taskStateSnapshot
-	// taskStateChecked distinguishes "no running task projection" from "the
-	// projection disagreed with the transcript". The latter must drop any older
-	// carried copy rather than silently preserving stale structured state.
-	taskStateChecked bool
 }
 
 // CompactionResult is the metadata-bearing result returned by CompactMessages.
@@ -228,7 +224,7 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 
 	// Preserve structured state (active plan + loaded skills) from the elided
 	// middle verbatim, so it is not lost or paraphrased away by the prose summary.
-	content := appendPreservedState(summaryLabel+"\n"+summary, middle, opts.taskState, opts.taskStateChecked)
+	content := appendPreservedState(summaryLabel+"\n"+summary, middle, opts.taskState)
 
 	compacted := make([]zeroruntime.Message, 0, systemEnd+1+(len(messages)-boundary))
 	compacted = append(compacted, messages[:systemEnd]...)
@@ -374,15 +370,13 @@ func (state *compactionState) calibratedTokens(raw int) int {
 	return int(float64(raw) * state.calibrationRatio)
 }
 
-func newCompactionState(options Options, tasks ...*taskState) *compactionState {
+func newCompactionState(options Options, task *taskState) *compactionState {
 	state := &compactionState{
 		enabled:      options.ContextWindow > 0,
 		threshold:    compactionThreshold(options.ContextWindow),
 		preserveLast: options.CompactionPreserveLast,
 		onUsage:      options.OnUsage,
-	}
-	if len(tasks) > 0 {
-		state.task = tasks[0]
+		task:         task,
 	}
 	return state
 }
@@ -429,10 +423,9 @@ func (state *compactionState) maybeCompact(
 	}
 
 	compacted, err := Compact(messages, CompactionOptions{
-		PreserveLast:     state.preserveLast,
-		Summarize:        summarizeClosure(ctx, provider, state.onUsage),
-		taskState:        state.task.snapshotForCompaction(messages),
-		taskStateChecked: state.task != nil,
+		PreserveLast: state.preserveLast,
+		Summarize:    summarizeClosure(ctx, provider, state.onUsage),
+		taskState:    state.task.snapshotForCompaction(messages),
 	})
 	if err != nil {
 		// Summarizer failed: keep the original history. The reactive path (or a
@@ -481,10 +474,9 @@ func (state *compactionState) recover(
 	}
 
 	result, compactErr := Compact(messages, CompactionOptions{
-		PreserveLast:     state.preserveLast,
-		Summarize:        summarizeClosure(ctx, provider, state.onUsage),
-		taskState:        state.task.snapshotForCompaction(messages),
-		taskStateChecked: state.task != nil,
+		PreserveLast: state.preserveLast,
+		Summarize:    summarizeClosure(ctx, provider, state.onUsage),
+		taskState:    state.task.snapshotForCompaction(messages),
 	})
 	if compactErr != nil {
 		// A genuine compaction attempt was made (and failed): the budget is spent
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -56,6 +56,14 @@ type CompactionOptions struct {
 	// injected so Compact stays pure and testable; the agent loop wires it to a
 	// real provider call.
 	Summarize func(toSummarize []zeroruntime.Message) (string, error)
+	// taskState is a transcript-corroborated snapshot supplied by the running
+	// agent. It is unexported so standalone callers cannot accidentally promote
+	// structured state above the messages they are compacting.
+	taskState *taskStateSnapshot
+	// taskStateChecked distinguishes "no running task projection" from "the
+	// projection disagreed with the transcript". The latter must drop any older
+	// carried copy rather than silently preserving stale structured state.
+	taskStateChecked bool
 }
 
 // CompactionResult is the metadata-bearing result returned by CompactMessages.
@@ -220,7 +228,7 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 
 	// Preserve structured state (active plan + loaded skills) from the elided
 	// middle verbatim, so it is not lost or paraphrased away by the prose summary.
-	content := appendPreservedState(summaryLabel+"\n"+summary, middle)
+	content := appendPreservedState(summaryLabel+"\n"+summary, middle, opts.taskState, opts.taskStateChecked)
 
 	compacted := make([]zeroruntime.Message, 0, systemEnd+1+(len(messages)-boundary))
 	compacted = append(compacted, messages[:systemEnd]...)
@@ -326,6 +334,7 @@ type compactionState struct {
 	// OnText is deliberately NOT forwarded (compaction stays invisible to the user),
 	// but its token COST must still be counted so usage reports and budgets include it.
 	onUsage func(Usage)
+	task    *taskState
 
 	// calibrationRatio scales the raw byte/4 token estimate toward the provider's
 	// real prompt-token count. ApproxTextTokens over-counts code-heavy content by
@@ -365,13 +374,17 @@ func (state *compactionState) calibratedTokens(raw int) int {
 	return int(float64(raw) * state.calibrationRatio)
 }
 
-func newCompactionState(options Options) *compactionState {
-	return &compactionState{
+func newCompactionState(options Options, tasks ...*taskState) *compactionState {
+	state := &compactionState{
 		enabled:      options.ContextWindow > 0,
 		threshold:    compactionThreshold(options.ContextWindow),
 		preserveLast: options.CompactionPreserveLast,
 		onUsage:      options.OnUsage,
 	}
+	if len(tasks) > 0 {
+		state.task = tasks[0]
+	}
+	return state
 }
 
 // maybeCompact runs proactive compaction at the top of a turn. It returns the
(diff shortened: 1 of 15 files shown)


gnanam1990 (changes requested):
**REQUEST CHANGES — one real defect, small fix. Everything else is solid.**

The projection is well-built: deterministic, side-effect-free on the loop, nil-safe throughout, and genuinely additive. My one blocking issue is that the headline compaction behavior stops working in exactly the scenario it exists for. Details below, then the things I checked that are fine.

---

### [Major] The objective is dropped at the *second* compaction

`compareTranscript` treats a plan mismatch as a whole-state mismatch:

```go
if !valid || found != state.planObserved || (found && !reflect.DeepEqual(transcriptPlan, tracked)) {
    parity = taskStateParityMismatch
}
```

`latestTaskPlan` scans `messages[i].ToolCalls` for `planToolName`. But compaction *elides the middle of the transcript*, which is where that tool call lives — `extractLatestPlan(middle)` exists precisely because the plan is being removed and has to be re-preserved as text. So after the first compaction, `found` is false while `planObserved` stays true, and parity flips to `mismatch`.

That alone would be harmless, but `appendPreservedState` also drops the carried copy:

```go
task := priorState.Task
if taskStateChecked { task = nil }        // taskStateChecked is always true in Run
if taskSnapshot != nil { task = ... }     // nil when parity mismatched
```

So on compaction #2, `taskSnapshot` is nil and the previously preserved task block is deleted rather than carried. Reproduced:

```
compaction #1: snapshot=true   objective preserved=true
compaction #2: snapshot=false  objective preserved=false
>>> OBJECTIVE LOST at compaction #2
```

Parity *does* recover once the model issues a fresh `update_plan` (I checked — it returns to `match`), so this is a window rather than a permanent state. But the window is unbounded: nothing forces the model to re-plan, and a run that compacts twice without an intervening plan update loses the objective entirely. That's the long-running case the feature is built for, and it fails silently — the trace still says `parity=mismatch` but nothing tells you the objective went missing.

**The fix is small.** The objective is captured once from the run's own prompt (`newTaskState(prompt, …)`) and is immutable for the lifetime of the run — it *cannot* be stale relative to the transcript, unlike the plan and verification counts. Gating it on plan parity is over-broad. Preserve `Objective` unconditionally and gate only the mutable fields (`Status`, plan counts, verification counts) on parity. That keeps the stated safety property — no stale structured state — while making the objective actually survive.

I'd also suggest `taskStateParityMismatch` not be the trigger for deleting a carried block that a *previous, matching* compaction wrote. "I can't corroborate the plan right now" is weaker evidence than "the carried objective is wrong."

---

### Minor

- **Parity is plan-only, but named and documented as whole-state.** `compareTranscript` compares `Plan.Items` and nothing else, yet `taskStateParityMatch` and the PR body's "use structured state only when it matches the transcript" read as though the whole projection is corroborated. Today the exposure is small (`preservedTaskState` only carries objective, status, and counts), but the naming invites a future field to be added to the snapshot and silently assumed verified. Consider `planParityMatch`, or a comment stating exactly what is and isn't compared.
- **`parseTaskPlan` is more lenient than the tool it now mirrors.** It accepts a `step` alias and silently skips empty-content items; `update_plan`'s `parsePlanItems` requires `content` via `stringArg(..., true)` and hard-rejects the *entire call* on a bad item. So guardrails and TaskState can track a plan the tool refused to store. The underlying shape (observing arguments rather than tool results) predates this PR, but consolidating onto one parser is the moment to make the two agree — or to note in `parseTaskPlan`'s doc that it is deliberately laxer and why.
- **`TaskStates` grows per tool call, not per turn.** `observe` → `emit` fires on every tool result, so this slice is O(tool calls) while `PrefixHashes` and friends are O(turns), and there's no cap. A few hundred tool calls is a few hundred KB of extra NDJSON. Worth a cap, or emitting only on plan/verification/completion transitions and letting tool counters ride along with the next event.
- **Variadic optional dependencies.** `newTaskState(objective, recorders ...*trace.Recorder)` and `newCompactionState(options, tasks ...*taskState)` silently ignore everything past the first argument. A plain nilable parameter says the same thing, can't be mis-called with two, and doesn't need the `len(...) > 0` dance.
- **`compareTranscript` reads like a query but mutates.** It bumps `Revision` and emits a trace event, and it's called from both `completionContext` and `snapshotForCompaction` — so a single completion check can bump the revision. Worth a name that signals the write, or a note on the method.
- **Nit on the privacy claim.** `ObjectiveHash` is a bare SHA-256 of the objective. That's not reversible, but it *is* confirmable: objectives are low-entropy and often short ("fix the login bug"), so anyone with the trace and a candidate list can test guesses. If "content-free" is meant strictly, an HMAC with a per-run random key gives you the same within-run drift detection without the confirmation oracle.

### Checked and fine

- **`planStatusRemaining` consolidation is behavior-preserving.** I compared the old inline switch against `normalizeTaskPlanStatus` term by term — the "not remaining" set is exactly `completed|failed` under both, and every other vocabulary word lands on `pending`/`in_progress` as before.
- **Routing `observePlanUpdate` and `formatPlanArguments` through `parseTaskPlan` is a genuine fix, not just a refactor.** Both previously parsed raw arguments *without* the `enforceSingleInProgress` coercion the `update_plan` tool applies, so the guardrail's pending count and the post-compaction plan text could disagree with the plan the tool actually stored. They now agree, and `parseTaskPlan`'s coercion matches `enforceSingleInProgress` exactly (all but the last `in_progress` → `completed`). Good catch to fold these together.
- **`capTaskObjective` is UTF-8 safe** — backs up with `utf8.RuneStart` before slicing, so the 512-byte cap can't split a rune.
- **The new `defer` is sound.** `result` is a named return, so the deferred `compareTranscript(result.Messages)` observes the final value. `EmitTaskState` is nil-safe, mutex-guarded, and no-ops once `r.finished` is set, so it cannot write into a sealed recorder no matter how the defers interleave.
- **The parity fallback on the completion path is safe.** When parity is mismatch, `completionContext` falls back to `guards.pendingPlanItems()`, and `guardState` isn't reset by compaction, so the transcript-derived signal survives. Verified the flag flips exactly as documented.
- **Trace payload matches the stated scope** — counts, statuses, and an objective hash; no plan text, tool output, or file paths, and only `len(ChangedFiles)`.

### Verification

`gofmt` clean, `go vet` clean. `internal/trace` fully green including `-race`. `internal/agent` passes under `-race -count=2` for the task-state, compaction-preserve, and completion-policy suites.

One `internal/agent` failure locally — `TestRequestPermissionsTurnGrantAllowsLaterToolAndCleansUp` — is a sandbox artifact of my environment, not this PR: it fails identically on this PR's exact base (`2e267bd`), which I checked out separately to confirm rather than assuming.

Nice piece of work overall — the plan-normalization consolidation is a real correctness win independent of the new feature. Fix the objective-preservation gating and I'm happy.

gnanam1990 (approved):
**APPROVE.** Re-reviewed at `d9ed9c0`. The major is fixed correctly — I re-ran my repro against the new head rather than reading the diff — and all six minors plus the nit are closed.

### The major — verified fixed

The fix is the right one: `Objective` is now set unconditionally whenever a snapshot exists, and only the mutable fields are gated on parity.

```go
task = &preservedTaskState{Objective: capTaskObjective(taskSnapshot.Objective)}
if taskSnapshot.PlanParity == taskPlanParityMatch {
    task.Status = ...; task.Pending = ...; task.VerificationPassed = ...
}
```

`snapshotForCompaction` no longer returns nil on mismatch (it records parity and returns the snapshot), and `taskStateChecked` is gone entirely — which removes the "delete the carried copy" path that was doing the damage. Re-ran the exact repro from last round:

```
compaction #1: parity=match     objective preserved=true  counts carried
compaction #2: parity=mismatch  objective preserved=true
OK: objective preserved, mutable fields correctly withheld on mismatch
```

I also added the inverse assertion — that a *matching* parity still carries the plan counts — so the fix isn't just "preserve everything always." It passes. The distinction now lands exactly where it should: the immutable objective always survives, the corroboration-dependent counts don't.

### Minors — all closed

- **Parity scope naming.** Renamed throughout: `taskStateParity` → `taskPlanParity`, `compareTranscript` → `observePlanParity`, `StateMatchesTranscript` → `PlanMatchesTranscript`, `TranscriptParity` → `PlanParity`. The new doc comment states plainly that "objective, tool, and verification fields are not part of this comparison." That was the point — a future field can no longer be silently assumed corroborated.
- **`parseTaskPlan` leniency.** Now rejects empty content outright and drops the `step` alias, so it matches `update_plan`'s own strictness instead of tracking plans the tool would have refused. `TestParseTaskPlanRejectsArgumentsTheToolWouldReject` and `TestFormatPlanArgumentsRejectsUnsupportedStepAlias` pin both directions. Worth noting the `len(plan) == 0 → (nil, true)` return is retained, so both sides of the `reflect.DeepEqual` parity check stay nil-typed and an empty plan can't produce a spurious mismatch — I checked that specifically, since the rewrite of this function was the place it could have crept in.
- **Trace growth.** Tool results no longer emit (`event.kind != taskStateEventToolResult`), with a final `task.emit()` in the defer so aggregate counts still land on an early return. Plus a hard `maxTaskStateEvents = 128` cap that replaces the tail rather than the head, so the final aggregate is never the one dropped — the right end to sacrifice.
- **Variadic optional deps.** `newTaskState(objective, recorder)` and `newCompactionState(options, task)` take plain parameters now.
- **Mutating query.** `observePlanParity` names the write and documents that it mutates and emits.
- **Objective hash.** Removed from the trace event entirely, and the type doc now says "confirmable content fingerprints are deliberately excluded." That's a cleaner resolution than the HMAC I suggested — if nothing consumes it, not emitting it is strictly better.

### Verification

`gofmt` clean, `go vet` clean. `internal/trace` green including `-race -count=2`. `internal/agent` green under `-race -count=2` for the task-state, compaction-preserve, completion-policy, and plan suites. Eight new tests landed, mapping 1:1 onto the findings.

Four `internal/agent` failures locally (`TestRunPromptsAndAllowsOutsideWorkspaceWrite`, `TestRunSessionAllowsLaterOutsideWorkspaceWrite`, `TestRunAppliesSandboxEvenInUnsafeMode`, `TestRequestPermissionsTurnGrantAllowsLaterToolAndCleansUp`) are sandbox artifacts of my environment — I ran all four against this PR's exact base (`2e267bd`) and they fail identically there.

Good turnaround. The objective/mutable split is a cleaner invariant than what was there before the review, not just a patch over the symptom. Merge is kevin's call per the program gate.

Vasanthdev2004 (approved):
Reviewed against the program gates, with the two things I most wanted to be sure of on a task-state projection: that the trace stays content-free, and that grounding completion in the objective does not quietly change exit-code behavior. Both hold up.

Trace privacy is the strong part. TaskStateEvent carries only counts and enum strings, no objective text, no plan content, no tool output, no file paths, and deliberately not even an objective_hash, so a traced run cannot confirm a guessed objective. It is content-free by construction rather than by scrubbing, which is the right way to build this, and the round-trip test plus the bounded-growth cap (128 with tail replacement) back it up. Changed files are kept in memory for the count only and never emitted or persisted, and the objective that does land in the compaction block is the user's own prompt that already lives in the transcript, capped at 512 bytes with UTF-8-safe truncation, so nothing new leaks.

Completion behavior is preserved, not just claimed. The structured PlanPending is only trusted when observePlanParity says the projection still matches the transcript, and on a match Plan.Pending+Plan.InProgress is the same quantity the transcript path already computes via pendingPlanItems, so the decision is unchanged; on a mismatch it falls back to the transcript count. The only new prompt content is the objective appended to the acceptance-verification nudge, which is on the opt-in semantic-check path. So headless exit semantics are the same and interactive runs stay byte-identical.

The plan-parsing consolidation is the one change that could read as a regression but is actually a correctness fix. I checked parseTaskPlan against the real update_plan tool: the tool requires content (no step field) and its enforceSingleInProgress keeps only the last in_progress item, downgrading earlier ones to completed. The new shared parser matches both, where the old formatPlanArguments was looser than the tool and accepted the step alias. So dropping step and coercing multiple in_progress lines up compaction and guardrails with the tool's real semantics, and planStatusRemaining keeps its prior result.

Gates: content-free trace (4), no default-path behavior change with a transcript fallback (2), nothing touching permissions/sandbox (3) or one-result-per-call (5), and the event is opt-in and listed in OptionalEventKeys (10). One non-blocking micro-note: observePlanParity re-walks the message history for the latest plan on each completion check and each compaction; it early-exits from the tail so it is cheap in practice, just worth remembering if the completion path ever gets hot.

Deterministic projection, snapshots are copied so callers cannot mutate the source, tool results are coalesced so the trace does not balloon. Nice work, and it closes out the task-state stream. Approving.

anandh8x:
Addressed in d9ed9c00.

- Added an exact two-compaction regression: after the first compaction elides the plan call, the second still preserves the immutable objective and omits uncorroborated mutable fields.
- Renamed transcript parity to plan parity and made the mutating observation explicit.
- Tightened plan parsing to reject the same malformed shapes as update_plan.
- Coalesced tool-result snapshots, capped retained task-state events at 128, and kept the latest aggregate at the cap.
- Replaced variadic optional dependencies with explicit nilable parameters.
- Removed the objective hash entirely so traces contain neither objective text nor a confirmable objective fingerprint.

Validation is green: formatting, vet, unrestricted full tests, focused race tests, build, smoke, and govulncheck. The advisory lint output remains limited to unrelated pre-existing findings.

PR #762, opened 2026-07-19
fix(plugins,skills): make reinstall transactions atomic

### LINKED ISSUE #754: fix(plugins,skills): reinstall is non-transactional and can lose lockfile entries
## Summary

Reinstalling an active plugin or skill deletes it before the complete replacement is staged, and separately, the lockfile that records installed plugin/skill state is rewritten without any interprocess transaction.

## Affected code

- `internal/plugins/install.go:140-155`
- `internal/plugins/install.go:210-245`
- `internal/skills/install.go:135-153`
- `internal/skills/install.go:270-307`

## Details

- **Delete-before-replace:** the active install is removed before its replacement is fully staged and validated. A copy failure, disk exhaustion, process interruption, or a failure publishing the lockfile can leave: no installation at all, a partial/corrupt installation, or installed content that disagrees with what the lockfile records.
- **Non-transactional lockfile:** the lockfile is rewritten independently of the content swap and without an interprocess lock, so two concurrent installs/reinstalls can race and one can silently lose the other's lockfile entry.

## Suggested fix

Build and validate the new plugin/skill content in a unique sibling directory first, swap it into place atomically while retaining the previous install until that swap succeeds, and publish the lockfile atomically under the same interprocess lock used for the content swap.

## Reference

Item AUD-007 from the 2026-07-18 codebase audit (`docs/CODEBASE_AUDIT_2026-07-18.md`).

PierrunoYT:
## Summary

- stage and verify plugin/skill replacements on the destination filesystem before touching active installs
- serialize install/remove content swaps and lockfile read-modify-writes with a cross-process lock on Unix and Windows
- atomically publish lockfiles and restore the prior installation if publication fails
- preserve retained backups if rollback itself cannot complete
- cover rollback and concurrent lockfile updates with regression tests

Fixes #754

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test -race ./internal/installtxn ./internal/plugins ./internal/skills`
- repeated concurrent install regressions under the race detector (`-count=10`)
- Windows cross-compilation of installtxn, plugins, and skills tests
- `git diff HEAD --check`

The pinned advisory golangci-lint command reports the repository's existing 36-item backlog in unrelated files; it reports no finding in the files changed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installations and removals now complete atomically, reducing the risk of partial updates.
  * Failed publishes automatically restore the previous installation or remove incomplete first-time installs.
  * Concurrent plugin and skill installations now preserve every lockfile entry.
  * Lockfile updates are written atomically for improved data safety.

* **Tests**
  * Added coverage for rollback behavior, workspace preservation, and concurrent installations across plugins and skills.

### DIFF
--- a/internal/installtxn/installtxn.go
+++ b/internal/installtxn/installtxn.go
@@ -1,0 +1,147 @@
+// Package installtxn provides the cross-process filesystem transaction used by
+// plugin and skill installation. Callers stage content before taking the lock,
+// then commit the content swap and lockfile update together while holding it.
+package installtxn
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const lockFileName = ".zero-install.lock"
+
+// Lock takes the per-install-root cross-process lock. It blocks until any other
+// installer or remover using dir has completed.
+func Lock(dir string) (func(), error) {
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return nil, fmt.Errorf("create install dir: %w", err)
+	}
+	return lockFile(filepath.Join(dir, lockFileName))
+}
+
+// StageDir creates an install workspace on the target filesystem. Content must
+// be built and validated in the returned stage directory before CommitDir is
+// called. cleanup is always safe to call.
+func StageDir(dir string) (stage string, cleanup func(), err error) {
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return "", func() {}, fmt.Errorf("create install dir: %w", err)
+	}
+	workspace, err := os.MkdirTemp(dir, ".zero-install-txn-")
+	if err != nil {
+		return "", func() {}, fmt.Errorf("create install staging dir: %w", err)
+	}
+	return filepath.Join(workspace, "staged"), func() { cleanupWorkspace(workspace) }, nil
+}
+
+// CommitDir replaces target with staged and runs publish while retaining the
+// previous target. If either the swap or publish fails, the previous target is
+// restored (or the new target is removed for a first install).
+//
+// The caller must hold the install-root lock returned by Lock.
+func CommitDir(target string, staged string, publish func() error) error {
+	workspace := filepath.Dir(staged)
+	backup := filepath.Join(workspace, "previous")
+	hadPrevious := false
+	if _, err := os.Stat(target); err == nil {
+		if err := os.Rename(target, backup); err != nil {
+			return fmt.Errorf("retain previous install: %w", err)
+		}
+		hadPrevious = true
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("inspect previous install: %w", err)
+	}
+
+	if err := os.Rename(staged, target); err != nil {
+		if hadPrevious {
+			if restoreErr := os.Rename(backup, target); restoreErr != nil {
+				return errors.Join(fmt.Errorf("publish staged install: %w", err), fmt.Errorf("restore previous install: %w", restoreErr))
+			}
+		}
+		return fmt.Errorf("publish staged install: %w", err)
+	}
+	if err := publish(); err != nil {
+		return rollback(target, backup, hadPrevious, err)
+	}
+	if hadPrevious {
+		_ = os.RemoveAll(backup)
+	}
+	cleanupWorkspace(workspace)
+	return nil
+}
+
+// RemoveDir removes target and runs publish while retaining the target until
+// publish succeeds. A publish failure restores the directory.
+//
+// The caller must hold the install-root lock returned by Lock.
+func RemoveDir(target string, publish func() error) error {
+	workspace, err := os.MkdirTemp(filepath.Dir(target), ".zero-install-txn-")
+	if err != nil {
+		return fmt.Errorf("create removal staging dir: %w", err)
+	}
+	defer cleanupWorkspace(workspace)
+	backup := filepath.Join(workspace, "previous")
+	if err := os.Rename(target, backup); err != nil {
+		return fmt.Errorf("retain removed install: %w", err)
+	}
+	if err := publish(); err != nil {
+		if restoreErr := os.Rename(backup, target); restoreErr != nil {
+			return errors.Join(err, fmt.Errorf("restore removed install: %w", restoreErr))
+		}
+		return err
+	}
+	_ = os.RemoveAll(backup)
+	return nil
+}
+
+func rollback(target string, backup string, hadPrevious bool, cause error) error {
+	if err := os.RemoveAll(target); err != nil {
+		return errors.Join(cause, fmt.Errorf("remove failed install: %w", err))
+	}
+	if hadPrevious {
+		if err := os.Rename(backup, target); err != nil {
+			return errors.Join(cause, fmt.Errorf("restore previous install: %w", err))
+		}
+	}
+	return cause
+}
+
+// cleanupWorkspace never removes a retained previous install. If rollback was
+// unable to restore it (for example because Windows still has a target file
+// open), preserving the workspace is safer than turning a recoverable error
+// into data loss.
+func cleanupWorkspace(workspace string) {
+	if _, err := os.Stat(filepath.Join(workspace, "previous")); err == nil {
+		return
+	}
+	_ = os.RemoveAll(workspace)
+}
+
+// WriteFileAtomically publishes data by renaming a complete sibling temporary
+// file over path. The caller is responsible for any surrounding transaction
+// lock.
+func WriteFileAtomically(path string, data []byte, perm os.FileMode) error {
+	temp, err := os.CreateTemp(filepath.Dir(path), ".zero-lockfile-")
+	if err != nil {
+		return err
+	}
+	tempPath := temp.Name()
+	defer func() { _ = os.Remove(tempPath) }()
+	if err := temp.Chmod(perm); err != nil {
+		_ = temp.Close()
+		return err
+	}
+	if _, err := temp.Write(data); err != nil {
+		_ = temp.Close()
+		return err
+	}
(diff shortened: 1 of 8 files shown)


gnanam1990 (approved):
**APPROVE** — the core mechanism is correct and does what it claims: I confirmed `TestConcurrentInstallsPreserveEveryLockEntry` fails on the merge base (only 1–3 of 12 lock entries survive across 5 runs, plus a torn `plugins.lock` that fails to parse) and passes here. A 30-iteration same-name concurrent install/remove stress run under `-race` produced no lockfile/disk inconsistency and no leaked workspaces. Everything below is an edge case or an incidental permission regression, not a failure of the design.

## What's genuinely well done

Several non-obvious hazards are already handled correctly, and it's worth saying so explicitly:

- Staging **inside** the target install dir (`internal/installtxn/installtxn.go:31`) means the commit rename can never hit `EXDEV`.
- The extra `staged`/`previous` nesting level is what keeps transient workspaces invisible to the loaders, which look exactly one level down for `dir/<entry>/plugin.json` (`internal/plugins/plugins.go:253`) and `dir/<entry>/SKILL.md` (`internal/skills/skills.go:307`). A flatter workspace layout would have made every in-flight install briefly visible as a broken extension.
- Renaming the old target out *before* renaming the new one in (`installtxn.go:48` then `:56`) is precisely what makes the swap work on Windows, where rename-over-an-existing-directory fails.
- The network fetch is deliberately kept outside the lock, and **re-reading the lockfile after acquiring it** (`internal/plugins/install.go:157-159`, `internal/skills/install.go:153-155`) is the detail that actually closes the lost-update window. That's the crux of the fix.
- `cleanupWorkspace`'s refusal to delete a workspace still holding `previous` (`installtxn.go:114-119`) is the right call, and the comment explains why.

## Findings

### [Minor] `WriteFileAtomically` resets the lockfile to 0644, discarding umask and any user-hardened mode
`internal/installtxn/installtxn.go:131`

`temp.Chmod(perm)` applies the literal `perm` argument (0o644 from `internal/plugins/install.go:271` and `internal/skills/install.go:332`) before the rename, so the published lockfile always ends up world-readable regardless of process umask or the mode the existing lockfile had. `os.WriteFile` on the base only applied `perm` at creation, so this is a behavior change.

Confirmed empirically against the base: (1) after `chmod 600 ~/.zero/skills/skills.lock`, the base preserves 0600 across a reinstall, this PR republishes it as 0644 (`-rw------- -> -rw-r--r--`); (2) under `umask 077` the base creates it 0600, this PR creates it 0644.

This matters because the lockfile stores the source string verbatim (`internal/skills/install.go:166`, `internal/plugins/install.go:169`) and remote sources are not rewritten — `canonicalSource` returns non-local sources unchanged (`internal/skills/install.go:428-440`). A private repo installed via `https://user:token@host/repo` puts that token into a file whose permissions were just silently widened on a multi-user host.

**Fix:** stat the destination first — if it exists, reuse `existing.Mode().Perm()` so an operator-hardened mode survives; if it doesn't, drop the explicit `Chmod` and let `CreateTemp`'s 0600 stand (or apply the umask explicitly) rather than forcing the literal perm.

### [Minor] `CommitDir` uses `os.Stat`, so installing over a dangling symlink now fails instead of replacing it
`internal/installtxn/installtxn.go:47`

`os.Stat(target)` follows symlinks, so a dangling symlink at the install target reports `ErrNotExist`, sets `hadPrevious=false`, and leaves the symlink in place. The subsequent `os.Rename(staged, target)` at `:56` then fails, because renaming a directory onto an existing non-directory returns `ENOTDIR`.

Confirmed empirically: with `~/.zero/skills/probe` a symlink pointing at a nonexistent path, `skills.Install` succeeds on the merge base (the old `RemoveAll` unlinked the dangling symlink) and records the lock entry. On this PR the same call returns `publish staged install: rename .../staged .../probe: not a directory`, the lockfile stays empty, and the dangling symlink is left behind — the skill can never be installed until the user manually deletes the link. Same path applies to `internal/plugins/install.go:170`.

**Fix:** use `os.Lstat(target)` for the `hadPrevious` probe, so any existing entry — symlink, dangling or not — is renamed into the workspace before the staged directory moves into place. Worth also considering the matching `os.Stat` presence checks in `Remove` (`internal/plugins/install.go:215`, `internal/skills/install.go:210`).

### [Nit] `Remove` creates the install dir and a lock file as a side effect of a lookup that then fails
`internal/installtxn/installtxn.go:18`

`Lock` unconditionally `MkdirAll`s the install root and opens `.zero-install.lock` inside it, and `Remove` takes the lock (`internal/plugins/install.go:203`, `internal/skills/install.go:198`) *before* checking whether the item exists. On a fresh machine with no `~/.zero/skills`, `zero skill remove foo` now creates `~/.zero/skills/` and `~/.zero/skills/.zero-install.lock` before returning `skill "foo" is not installed`. Verified: after `Remove(dir, "nope")` against a nonexistent dir, the directory exists and contains the lock file. Nothing was created on the base.

**Fix:** do the cheap `ReadLock`/`os.Stat` existence check before taking the lock and return the not-installed error early, or give `Lock` a non-creating variant that returns a no-op unlock when the install root doesn't exist.

### [Nit] Lockfile write failures during `Remove` are reported as directory-removal failures
`internal/plugins/install.go:228`, `internal/skills/install.go:224`

`RemoveDir` returns the publish callback's error verbatim (`internal/installtxn/installtxn.go:92`), but both call sites wrap everything it returns in `remove plugin dir:` / `remove skill dir:`. If the directory rename succeeds and `writeLock` then fails, the rollback restores the directory correctly — but the user sees `remove plugin dir: write plugins.lock: ...`, pointing at a directory removal that in fact succeeded and was rolled back. It sends anyone debugging toward directory permissions rather than the lockfile.

**Fix:** have `RemoveDir` distinguish its own failures from the callback's (typed or sentinel-wrapped publish error), or drop the blanket wrap at the call sites and let the already-descriptive inner errors surface.

## Checked and cleared

Three things looked like defects on first read and are not — flagging them so they don't get re-raised:

- **Stranded `previous` workspaces.** Looks like silent data hiding, but it's reported and recoverable: the `os.Rename` failure is a `*os.LinkError` wrapped with `%w` at `installtxn.go:104` and surfaced to stderr (`internal/cli/distribution.go:83` and `:204`), printing the full path to the stranded `previous` directory — I forced the state and observed it. It's a plain directory recoverable with `mv`. The behavior is deliberate, documented at `installtxn.go:110-113`, and pinned by `installtxn_test.go:66`. More importantly it's strictly better than the base, which did `os.RemoveAll(target)` *before* `copyTree`, i.e. destroyed the previous install unrecoverably. Note the cited Windows mechanism doesn't apply anyway: an open file under `target` makes `os.RemoveAll` at `:99` fail and return at `:100`, so `:103` is never reached on that path.
- **Missing fsync on the staged tree / parent dir.** Not a regression — the base had zero syncs anywhere in the install path (`grep '\.Sync()'` on both trees); `installtxn.go:139` is the only one and this PR adds it. Also, `temp.Sync()` runs *before* `replaceFile` at `:146`, and on ext4 that fsync forces the running journal transaction which already contains the `os.Rename(staged, target)` issued at `:56` — so the ordering concern is backwards. And the package docs (`installtxn.go:1-3`) promise cross-process mutual exclusion, not power-loss durability. Staged file *data* is still unsynced (`internal/plugins/install.go:397-411`), but that's a pre-existing codebase-wide hardening opportunity, not something introduced here.
- **Concurrency tests asserting only `len(entries)`.** Accurate description, but not a real gap. I mutation-tested it: no-op'ing the `os.Rename` at `installtxn.go:57` does let the concurrency test pass alone, but fails 8 other tests in the same `go test ./internal/plugins/ ./internal/skills/` invocation (`TestInstallCopiesLocalPluginAndRecordsHash`, `TestInstallCopiesEntireTree`, `TestInstallFromRealLocalGitRepo`, +5 in skills). Deleting the retained-`previous` guard at `installtxn.go:110-113` immediately fails the PR's own `TestCleanupWorkspacePreservesRetainedPreviousInstall`. Entry count is the correct and only assertion that discriminates fixed-from-broken for the bug this PR targets; on-disk content is covered elsewhere.

## Build & tests

`gofmt -l .` clean repo-wide, `go build ./...` clean, `go vet` clean on all three touched packages. `internal/installtxn`, `internal/plugins`, `internal/skills` all pass under `-race -count=5` with no flakiness and no race reports.

**No test failure is attributable to this PR.** There are failures in `internal/tools`, `internal/tui`, and `internal/cli` (sandbox-escape and doctor tests), but I re-ran each of them in a worktree at the true merge base and the failure sets are byte-for-byte identical on both sides. Root cause is environmental — the sandbox backend reports `{"name":"unavailable","available":false,"platform":"windows","fallback":true}` while running on darwin, so workspace-escape writes aren't blocked and doctor exits 3. Pre-existing, not yours.

Downstream check: `internal/cli`, `internal/tools`, `internal/tui`, `internal/zerocommands` all import the changed packages; full suites run on both sides with identical results, so the `plugins`/`skills` API changes broke no consumer.

Cross-platform: `lock_windows.go` is behind `//go:build windows` and never compiled by the darwin or linux toolchain, so I verified it explicitly — `GOOS=windows GOARCH=amd64 go build` and `go vet` are both clean, as is `GOOS=linux`. Flagging that the Windows path (`LockFileEx`/`UnlockFileEx`, `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING|MOVEFILE_WRITE_THROUGH`) has **no test coverage** — `installtxn`'s tests only exercise the unix implementation, so it's build-verified only. Worth a manual smoke test on Windows before this ships, given the whole point is a cross-process lock.

One housekeeping note: this branch is based on `ce4a996`, which is behind current main — a rebase before merge would avoid a surprising diff.

kevincodex1 (approved):

PR #763, opened 2026-07-19
fix(perfbench): grant write tools so mutating tasks measure real edits

Vasanthdev2004:
## The bug

The turn benchmark invokes `zero exec` in its default read-only posture, which exposes no write or shell tools. `zero exec --list-tools` in that mode lists only read tools plus `web_fetch`, but no `edit_file`, `apply_patch`, `write_file`, `exec_command`, or `bash`. So the mutating task classes (edit, fix, refactor) can never apply a change. A run either grinds to the turn ceiling hunting for an edit tool, or reports a no-tool blocker and exits incomplete, and the only tasks that "pass" are ones whose oracle grep happens to match the stamped `.zero-answer.txt` narration rather than a real edit.

This invalidates every mutating-class number the benchmark has produced, in both the grok-4.5 and glm-5.2 baselines. The read classes (nav) were unaffected because they only need read tools.

## How it surfaced

A live per-turn trace of fix-07 (a one-line `admin-` to `user-` prefix change, about as trivial as an edit gets) showed the model calling only `read_file`, `grep`, `lsp_navigate`, and `update_plan`, then ending with:

> my available toolset in this session doesn't include a file-writing tool (no `edit_file`, `apply_patch`, `write_file`, `exec_command`, or `bash`), so I cannot modify bugs.go directly.

It identified the exact fix and had no tool to apply it. Re-running the same prompt with `--skip-permissions-unsafe` landed the edit and passed the oracle. Probes also confirmed the oracles themselves are sound (a correct hand-applied fix passes fix-01, fix-07, and refactor-01), so the failure was the missing tools, not the checks.

## The fix

Add `--skip-permissions-unsafe` to the benchmark's exec invocation. Each task already runs in an isolated, throwaway fixture copy (the runner copies the fixture into a fresh temp dir per task), so granting the full tool set has nothing to protect, and it is the whole point: the benchmark is supposed to measure real edits. This is a correctness contract for the harness, not a preference, so it is unconditional and pinned by a test.

This changes only the benchmark's own `zero exec` invocation. It does not touch the agent, the provider layer, or the product's permission model.

## Verification

Re-running four representative mutating tasks (edit-01, fix-01, fix-07, refactor-01) with the fix:

- **4/4 real passes**, where before they failed or false-passed.
- `tool_execution` rose from ~0.4% to ~29% of wall time, i.e. the write and shell tools are now actually being exercised rather than sitting unavailable.

`go build`, `go vet`, `gofmt`, and the `internal/perfbench` + `cmd/zero-perf-bench` suites are green, including the new `TestBuildTurnExecArgsGrantsWriteTools`.

## Follow-ups (separate)

- The mutating-class numbers in both existing baseline reports should be treated as invalid and re-captured with this fix; a re-run is a benchmarking task, not part of this PR.
- The positive-grep oracles that still match `.zero-answer.txt` (edit-03/06/08/09/10, the remaining half of the #737 class) let a narration false-pass even without real edits; worth excluding the answer file from those greps too, in a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Oracle-based benchmark tasks now treat INCOMPLETE (exit code 4) as non-blocking, allowing oracle stamping and verification to decide pass/fail.
  * Latency-only benchmark tasks continue to fail immediately on any non-zero exit code.
  * Benchmark runs now reject tasks without required workspace fixture details, avoiding execution in an unintended working directory.
  * Default benchmark exec arguments now include automatic member/tool access, while keeping the prompt as the final argument.
* **Tests**
  * Added coverage for oracle-vs-latency behavior on non-zero exits, plus exec-args and fixture validation.

### DIFF
--- a/internal/perfbench/turn_bench.go
+++ b/internal/perfbench/turn_bench.go
@@ -790,7 +790,16 @@ func stampOracleAndAnswer(task BenchTask, outBuf []byte) error {
 }
 
 func buildTurnExecArgs(task BenchTask, rc RunContext, tracePath string, extraArgs []string) []string {
-	args := []string{"exec", "--output-format", "stream-json", "--trace", tracePath}
+	// --skip-permissions-unsafe is REQUIRED, not optional: without it `zero exec`
+	// runs in its default read-only posture, which exposes no write or shell tools
+	// (no edit_file/apply_patch/write_file/exec_command/bash). The mutating task
+	// classes (edit/fix/refactor) then cannot apply any change, so every run
+	// grinds to the turn ceiling or reports a no-tool blocker, and the only tasks
+	// that "pass" are ones whose oracle grep matches the stamped answer text rather
+	// than a real edit. Each task already runs in an isolated, throwaway fixture
+	// copy (see NewTurnExecRunner), so granting the full tool set has nothing to
+	// protect and is the whole point: the benchmark must measure real edits.
+	args := []string{"exec", "--skip-permissions-unsafe", "--output-format", "stream-json", "--trace", tracePath}
 	if model := strings.TrimSpace(rc.Model); model != "" {
 		args = append(args, "--model", model)
 	}
--- a/internal/perfbench/turn_bench.go
+++ b/internal/perfbench/turn_bench.go
@@ -578,15 +578,26 @@ func WriteTurnBenchJSON(w io.Writer, result TurnBenchResult) error {
 	return encoder.Encode(result)
 }
 
+// execExitIncomplete mirrors internal/cli.exitIncomplete (the `zero exec` exit
+// code, 4) for a headless run the completion gate marked INCOMPLETE: the run
+// stopped without a completion signal (e.g. it couldn't self-verify because its
+// sandboxed shell was unavailable under --auto member), which is distinct from a
+// crash (1), usage error (2), provider failure (3), or interruption (130). It is
+// the ONLY nonzero exit an oracle-bearing task may defer to the oracle for; every
+// other nonzero exit stays authoritative. Kept as a local literal so perfbench
+// doesn't import the cli package; must stay in sync with that constant.
+const execExitIncomplete = 4
+
 // NewTurnExecRunner builds the production turn-benchmark runner: it invokes
 // headless `zero exec` with stream-json output AND `--trace <tmpfile>`, then
 // parses the emitted NDJSON trace into a *trace.TurnTrace. binary is the path to
 // the `zero` binary; extraArgs are appended to every invocation. Pass/fail is
 // decided by the task's oracle (OracleTest/VerificationCommand) when it has one:
 // the stamped oracle is ground truth, so an oracle-bearing task that applied the
-// right edit but exited INCOMPLETE (its sandboxed self-verification couldn't run)
-// still passes. For a latency-only task (no oracle) the stream-json run_end exit
-// code is the only signal, so a nonzero exit fails it.
+// right edit but exited INCOMPLETE (exit 4 — its sandboxed self-verification
+// couldn't run) still passes. Any OTHER nonzero exit (crash/usage/provider/
+// interrupt) stays authoritative and fails even an oracle-bearing task. For a
+// latency-only task (no oracle) any nonzero exit is the only signal, so it fails.
 func NewTurnExecRunner(binary string, extraArgs ...string) TurnRunner {
 	return func(ctx context.Context, task BenchTask, rc RunContext) TurnTaskOutcome {
 		// buildTurnExecArgs grants the write + sandboxed-shell tool set to EVERY
@@ -670,17 +681,20 @@ func NewTurnExecRunner(binary string, extraArgs ...string) TurnRunner {
 			}
 		}
 
-		// Decide what a nonzero exit means. For a latency-only task (no oracle) the
-		// exit code is the only correctness signal, so a nonzero exit is the verdict
-		// and we stop here. For an oracle-bearing task the oracle below — the
-		// compiled test / grep run against the actual fixture files — is ground
-		// truth: a run that applied the correct edit but exited INCOMPLETE (e.g. it
-		// couldn't self-verify because the sandboxed shell was unavailable under
-		// --auto member) still achieved the task, so we let the oracle decide and
-		// drop the exit-code failure. A run that really did nothing useful just
-		// fails that same oracle, so this can't turn a bad run into a pass.
+		// Decide what a nonzero exit means. We defer to the oracle for exactly ONE
+		// case: an oracle-bearing task that exited INCOMPLETE (exit 4). That is the
+		// completion gate downgrading a run whose edits may be correct but that
+		// couldn't emit a completion signal (e.g. its sandboxed self-verify couldn't
+		// run under --auto member) — the oracle below (the compiled test / grep run
+		// against the actual fixture files) is ground truth, so if the edit really
+		// landed it passes, and if the run abandoned the task the oracle fails it.
+		// Every OTHER nonzero exit stays authoritative: a crash (1), usage error
+		// (2), provider failure (3), or interruption (130) is a genuine failure, so
+		// a partial edit that happens to satisfy the oracle can't launder it into a
+		// pass. A latency-only task has no oracle to defer to, so any nonzero exit
+		// fails it.
 		if outcome.VerifyErr != "" {
-			if len(task.VerificationCommand) == 0 {
+			if len(task.VerificationCommand) == 0 || exitCode != execExitIncomplete {
 				return outcome
 			}
 			outcome.VerifyErr = ""
--- a/internal/perfbench/turn_bench.go
+++ b/internal/perfbench/turn_bench.go
@@ -589,10 +589,21 @@ func WriteTurnBenchJSON(w io.Writer, result TurnBenchResult) error {
 // code is the only signal, so a nonzero exit fails it.
 func NewTurnExecRunner(binary string, extraArgs ...string) TurnRunner {
 	return func(ctx context.Context, task BenchTask, rc RunContext) TurnTaskOutcome {
+		// buildTurnExecArgs grants the write + sandboxed-shell tool set to EVERY
+		// invocation, so a task MUST run inside an isolated fixture copy — never the
+		// caller's cwd. The grant is unconditional, so the isolation can't be
+		// optional: a fixtureless task would turn an agent with write/shell tools
+		// loose in the caller's working directory (their real repo). Enforce the
+		// invariant instead of trusting it — reject a fixtureless task up front,
+		// before anything is launched, so the grant and the guard can't drift apart.
+		// Every shipped task declares a fixture, so this only fires on a malformed or
+		// newly added task, and it fails loudly rather than silently.
+		if strings.TrimSpace(task.WorkspaceFixture) == "" {
+			return TurnTaskOutcome{Err: errors.New("benchmark task has no workspaceFixture: the run grants write/shell tools and must execute in an isolated fixture copy, not the caller's cwd")}
+		}
 		// Isolate the workspace: copy the fixture into a fresh temp dir so a
 		// mutating task (edit/fix/refactor) can't dirty the shared, checked-in
-		// fixture or bleed into a later iteration of the same task. When no
-		// fixture is configured the agent runs in the caller's cwd as before.
+		// fixture or bleed into a later iteration of the same task.
 		if fixture := strings.TrimSpace(task.WorkspaceFixture); fixture != "" {
 			copyDir, parent, cerr := copyFixture(fixture)
 			if cerr != nil {
--- a/internal/perfbench/turn_bench.go
+++ b/internal/perfbench/turn_bench.go
@@ -582,8 +582,11 @@ func WriteTurnBenchJSON(w io.Writer, result TurnBenchResult) error {
 // headless `zero exec` with stream-json output AND `--trace <tmpfile>`, then
 // parses the emitted NDJSON trace into a *trace.TurnTrace. binary is the path to
 // the `zero` binary; extraArgs are appended to every invocation. Pass/fail is
-// decided from the stream-json run_end exit code (and the task's
-// VerificationCommand when present), exactly like NewExecRunner.
+// decided by the task's oracle (OracleTest/VerificationCommand) when it has one:
+// the stamped oracle is ground truth, so an oracle-bearing task that applied the
+// right edit but exited INCOMPLETE (its sandboxed self-verification couldn't run)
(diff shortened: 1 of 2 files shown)


gnanam1990 (dismissed):
**APPROVE** — the diagnosis is excellent and the fix is right. One thing I'd like tightened before merge, and it's about three lines.

The investigation here is the good part: catching that the mutating classes were passing on oracle/answer contamination rather than real edits, confirming the oracles themselves are sound with a hand-applied fix, and quantifying it with `tool_execution` going from ~0.4% to ~29% of wall time. Flagging both baselines as invalid rather than quietly re-running is the right call. The `--skip-permissions-unsafe` conclusion is correct — a headless run can't answer a permission prompt, so there is no narrower posture that both exposes the write tools and completes unattended.

### The safety argument depends on an invariant that isn't enforced

The comment and the test both justify the flag the same way — "Each task already runs in an isolated, throwaway fixture copy … so granting the full tool set has nothing to protect." That is true for every task shipped today. It is not true structurally.

The isolation is conditional on the task *having* a fixture (`turn_bench.go:593, 615`):

```go
if fixture := strings.TrimSpace(task.WorkspaceFixture); fixture != "" {
    copyDir, parent, cerr := copyFixture(fixture)   // isolate
    ...
}
// "When no fixture is configured the agent runs in the caller's cwd as before."
...
if dir := strings.TrimSpace(task.WorkspaceFixture); dir != "" {
    cmd.Dir = dir                                    // only then is cwd redirected
}
```

`WorkspaceFixture` is optional (`json:"workspaceFixture,omitempty"`, no validation requires it) and the runner explicitly contemplates the empty case. The new flag, by contrast, is added **unconditionally to every invocation** — which the PR description states as a deliberate property.

So the guard is conditional and the grant is not. A task added without a fixture would run an LLM with the full write and shell tool set, no permission prompting, **in the caller's working directory** — the user's real repo.

I checked how bad that actually is rather than assuming the worst: the workspace sandbox survives unsafe mode. `TestRunAppliesSandboxEvenInUnsafeMode` (`loop_test.go:2838`) asserts an out-of-workspace write under `PermissionModeUnsafe` is denied with `BlockOutsideWorkspace` at critical risk and never hits the disk. So unsafe disables *prompting*, not confinement, and the blast radius is the repo rather than the filesystem. That's a meaningful mitigation and it's why this isn't a blocker.

I also confirmed the hazard is latent, not live: both shipped catalogs are fully covered — `internal/perfbench/manifests/baseline.json` has 48 tasks and `cmd/zero-perf-bench/testdata/terminal-bench-sample.json` has 2, with **zero** lacking a `workspaceFixture`.

**Suggested tightening**, in the spirit of the PR's own "this is a correctness contract, not a preference": make the runner enforce what the comment asserts, so the grant and the guard can't drift apart. Something like refusing to launch when unsafe permissions are being granted and the task has no fixture:

```go
if strings.TrimSpace(task.WorkspaceFixture) == "" {
    return TurnTaskOutcome{Err: errors.New(
        "benchmark tasks must declare a workspaceFixture: the run grants unsafe permissions and would otherwise execute in the caller's cwd")}
}
```

That turns "nothing to protect" from a comment into a checked precondition, and it fails loudly at the moment someone adds a fixture-less task rather than silently letting an unprompted agent loose in their repo. A companion test would pair naturally with `TestBuildTurnExecArgsGrantsWriteTools` — the two together would pin both halves of the contract instead of just the grant.

I'm approving rather than blocking because nothing is broken today and the sandbox bounds the damage, but I do think this belongs in this PR rather than a follow-up: it's the other half of the same invariant, and it's cheaper to add now than to remember later.

### Minor

- `buildTurnExecArgs` appends `extraArgs` as well, so a caller that already passes `--skip-permissions-unsafe` produces it twice. Harmless with current parsing, but if the runner is going to own this flag unconditionally it'd be tidy to de-duplicate, or to document that `extraArgs` shouldn't carry it.

### Verification

`gofmt` and `go vet` clean. `internal/perfbench` and `cmd/zero-perf-bench` both green, including the new `TestBuildTurnExecArgsGrantsWriteTools`. The added assertion that the prompt stays the final argument is a nice touch — that ordering is easy to break when prepending flags.

Agreed on both follow-ups, and thanks for scoping them out explicitly. The `.zero-answer.txt` grep contamination in edit-03/06/08/09/10 is the one I'd prioritize — until that's excluded, those tasks can still false-pass on narration, which is the same class of defect this PR just fixed for the tool-availability side.

anandh8x (changes requested):
Blocking: `--skip-permissions-unsafe` is broader than this benchmark needs. The fixture copy protects benchmark inputs from persistent edits, but unsafe mode also pre-grants permission and automatically retries sandbox-restricted commands unsandboxed (including network-related sandbox failures), so model-generated commands are not confined to the copied fixture. Please invoke the benchmark with `--auto member` instead. That existing headless mode advertises write and sandboxed-shell tools for mutating tasks while retaining workspace, network, and destructive-operation safeguards. Please update the test to assert that mode and that the unsafe flag is absent.\n\nValidation on f1e9ceca: focused perfbench tests pass; `make fmt-check` and diff hygiene pass.

Vasanthdev2004:
Thanks all. Pushed two commits (`19a18b2`, `8053bc3`) covering everything.

**anand's blocker: dropped `--skip-permissions-unsafe`.** You're right that unsafe is broader than this benchmark needs. It drops the workspace/network/destructive guards and auto-retries sandbox-blocked commands unsandboxed, so model-generated commands aren't confined to the copied fixture. The runner now invokes `--auto member`, which advertises the write and sandboxed-shell tools the mutating classes need while keeping those safeguards. `TestBuildTurnExecArgsGrantsWriteTools` now asserts `--auto member` is present and `--skip-permissions-unsafe` is absent.

**One consequence of member-auto worth flagging.** member-auto's shell is sandboxed, so on a host without sandbox setup the agent's own self-verification (its `go test`/`go build`) can't run, and the turn exits INCOMPLETE even after a correct edit. The runner used to treat any nonzero exit as failure before consulting the oracle, so those correct edits false-failed. I made the oracle authoritative for oracle-bearing tasks: the stamped test/grep against the actual fixture files is ground truth, so a correct edit passes regardless of the exit code. Latency-only tasks have no oracle, so a nonzero exit still fails them (that path is unchanged; `TestNonzeroExitStillFailsLatencyOnly` pins it). Evidence: a 4-task glm-5.2 smoke (edit-01/fix-01/fix-07/refactor-01) goes 4/4 real passes under member-auto with this in place; without it the same run scored 1/4 as the shell-self-verifying tasks exited incomplete.

**CodeRabbit + gnanam: the grant/guard drift.** You both caught the same thing, and it survives the member-auto switch: the tool grant is unconditional but fixture isolation was conditional, so a fixtureless task would run write/shell tools in the caller's cwd. Fixed it the way gnanam suggested. `NewTurnExecRunner` now rejects a task with no `workspaceFixture` before launching, so the isolation is a checked precondition rather than a comment. `TestTurnRunnerRejectsFixturelessTask` proves the rejection is pre-launch (it returns the fixture error even with a nonexistent binary). Every shipped task already declares a fixture, so this only fires on a malformed or newly added one.

**gnanam's minor and the follow-ups.** `extraArgs` isn't expected to carry the permission flag now that the runner owns it; I left it rather than adding dedup for a hypothetical, but I'll add a doc note if you'd rather. The `.zero-answer.txt` grep contamination in edit-03/06/08/09/10 stays tracked as the separate follow-up from the original PR body.

gnanam, heads up: your approval got auto-dismissed when I pushed (branch protection drops stale approvals on new commits), not by me by hand. Re-approve if the tightening reads right. Merge stays kevin's call per the program gate.

anandh8x (changes requested):
The original safety blocker is resolved: --auto member is the correct narrow mode, the unsafe flag is absent, and fixtureless tasks are rejected before launch.

One new blocker remains in the exit-code handling. The comments and test say only INCOMPLETE (exit 4) should defer to the oracle, but the implementation clears VerifyErr for every nonzero run_end whenever VerificationCommand is present. That also converts provider failure (3), crash (1), and interruption (130) into a passing benchmark result when a partial edit happens to satisfy the oracle. Please allow the oracle override only when exitCode == 4; keep every other nonzero exit authoritative, and add a regression test for at least one non-INCOMPLETE code.

Validation on 8053bc37: focused perfbench tests and vet pass; make fmt-check and diff hygiene pass.

Vasanthdev2004:
Good catch, fixed in `57d1ae5` (+ `390460f4` for follow-up tests).

You're right on the semantics: the previous commit cleared `VerifyErr` for any nonzero run_end when an oracle was present, which would let a crash (1), provider failure (3), or interrupt (130) pass if a partial edit happened to satisfy the oracle. The override is now gated on `exitCode == exitIncomplete (4)` only. That's the one code where a correct edit can legitimately exit nonzero (member-auto's sandboxed self-verify can't run without sandbox setup, so the completion gate downgrades the turn), and every other nonzero exit stays authoritative.

On tests: `TestNonIncompleteExitStaysAuthoritative` now runs a correct edit through exit 1/3/130 and asserts each still fails on its exit code, not the oracle. I also added `TestIncompleteExitStillFailsWhenOracleFails` (exit 4 with a no-op still fails, since deferring to the oracle still consults it) so the suite can't stay green if exit 4 ever regressed to a blanket pass.

One thing I left as a local mirror deliberately: `execExitIncomplete = 4`. perfbench can't reference `cli.exitIncomplete` (unexported, and importing the cli package here isn't worth the coupling), and the real contract is the spawned `zero exec` process's numeric exit code, not a shared symbol, so I documented it as a mirror that must track that constant. Happy to revisit if you'd rather export it.

Smoke stays 4/4 (those tasks exit INCOMPLETE, so the tighter gate costs no real passes). Also dropped a now-dead fixture branch and tightened the runner doc while I was in there.

anandh8x (approved):
Approved on 390460f4. The benchmark now uses the narrow --auto member posture, enforces fixture isolation before launch, and defers to the oracle only for INCOMPLETE exit 4. Crash, provider-failure, and interruption exits remain authoritative, with regression coverage on both sides of the exit-4 boundary.

Validated: focused perfbench tests, focused vet, make fmt-check, and diff hygiene all pass. All GitHub checks are green.

gnanam1990 (approved):
**APPROVE.** Re-reviewed at `390460f`. This is a materially better PR than the one I approved at `f1e9cec` — the permission posture changed for the better, the isolation invariant is now enforced rather than assumed, and the new exit-4 logic holds up under scrutiny.

## `--auto member` is the right call, and it's not a small difference

Swapping `--skip-permissions-unsafe` for `--auto member` addresses my concern at its root instead of fencing around it. I measured the actual tool exposure rather than taking the comment's word for it:

| posture | tools | write/shell |
|---|---|---|
| default | 12 | **none** |
| `--auto member` | 17 | `apply_patch`, `bash`, `edit_file`, `exec_command`, `write_file` |
| `--skip-permissions-unsafe` | 35 | same five |

So the original diagnosis is exactly right — the default posture genuinely exposes zero write/shell tools, which is why the mutating classes could never apply an edit. And `member` grants precisely the five tools the benchmark needs at **half the attack surface**, while keeping the workspace/network/destructive guards that `--skip-permissions-unsafe` drops. That's a better answer than the one I signed off on.

## The isolation invariant is now enforced

`NewTurnExecRunner` rejects a fixtureless task before anything launches, and the dead conditional branch is gone so isolation is unconditional rather than best-effort. That's exactly the "grant and guard can't drift apart" property I was after.

I mutation-tested it — neutralised the guard, kept the test:

```
--- FAIL: TestTurnRunnerRejectsFixturelessTask (0.00s)
```

It fires. Real regression test, not a decorative one.

## The exit-4 deferral — checked carefully, and it's sound

This is new since my last review and it's the piece that deserved the most scrutiny, because on its face it risks reintroducing the exact false-pass class this PR exists to eliminate. Before, any nonzero exit failed the task outright. Now an oracle-bearing task that exits INCOMPLETE has its verdict handed to the oracle — so a run that narrated without editing could, in principle, pass on an oracle that greps the stamped answer text.

I checked whether that's reachable across all 24 mutating tasks in `manifests/baseline.json`. It isn't:

- **Every one of the 24 is gated on `go test ./...` with `&&`**, so the compiled oracle must pass against the real fixture source before any grep is even evaluated. Narration alone cannot satisfy any of them.
- `refactor-01`, `refactor-02`, `refactor-06` do carry positive greps without `--exclude=.zero-answer.txt` — but each sits behind the `go test` gate, so the answer file can't carry them either.
- The four that reference the answer file at all (`edit-01`, `edit-02`, `refactor-03`, `refactor-05`) reference it precisely in order to `--exclude` it. That's the correct pattern, already in use.
- The `nav-*` oracles do grep the answer file, which is right for those — the narration *is* the deliverable.

So the deferral is safe as the corpus stands, and the reasoning in the comment at `turn_bench.go:685-696` matches what the code actually does.

**One thing worth writing down for the follow-up**, though, because it's a genuine change in where the safety comes from: the deferral's correctness is now a property of the *oracle corpus*, not of the runner. Today every mutating oracle is `go test`-gated, so the runner can safely defer. A future mutating task whose oracle is a bare positive grep with no compiled-test gate would be able to pass on narration alone at exit 4 — where under the old "any nonzero exit fails" rule it would have failed regardless. That makes the `--exclude=.zero-answer.txt` convention (and the `go test &&` prefix) worth promoting from convention to something enforced when tasks are loaded. It was already on your follow-up list; this change raises its priority a little rather than creating a new problem.

## Test coverage

Five new tests pin all four directions, which is what I'd want for logic like this:

- `TestOracleAuthoritativeOnIncompleteExit` — exit 4 + real edit + passing oracle → pass
- `TestIncompleteExitStillFailsWhenOracleFails` — exit 4 + failing oracle → fail
- `TestNonIncompleteExitStaysAuthoritative` — crash/usage/provider/interrupt → fail regardless of oracle
- `TestNonzeroExitStillFailsLatencyOnly` — no oracle to defer to → any nonzero fails
- `TestTurnRunnerRejectsFixturelessTask` — the isolation guard

## Nit

`const execExitIncomplete = 4` (`turn_bench.go:589`) hand-mirrors the unexported `cli.exitIncomplete`, protected only by a "must stay in sync" comment. `internal/cli` doesn't import `internal/perfbench`, so there's no cycle in the way — but the constant is unexported, so a compile-time guard needs exporting it (or adding an exported alias). Worth doing at some point: if the CLI's exit codes are ever renumbered, this silently reclassifies *every* incomplete run as a hard failure, and the symptom would be a benchmark-wide pass-rate drop with no obvious cause.

## Verification

`gofmt` and `go vet ./internal/perfbench/...` clean. Full `internal/perfbench` and `cmd/zero-perf-bench` suites pass. The five new tests pass individually. Guard mutation-tested as above. Tool-exposure numbers measured against a binary built from this head, not inferred.

Nice iteration — the posture change in particular is the kind of fix that makes the original concern moot rather than mitigated.

PR #764, opened 2026-07-20
Add shell completion generation

### LINKED ISSUE #499: Add shell completions command (zero completions <shell>)
## Problem

Zero has a large CLI surface (`exec`, `daemon`, `mcp`, `sandbox`, `auth`, and many more subcommands), but there is no built-in way to generate shell completion scripts. Users must type full command names and flags manually, which slows down everyday use and makes nested subcommands easy to mistype or forget.

Other terminal agents in this space already ship a `completions` command. For example, `grok completions <shell>` generates scripts for bash, zsh, fish, powershell, and elvish. Zero currently has no equivalent, so tab completion is unavailable out of the box.

## Requested behavior

Add a `completions` subcommand to the Zero CLI:

```text
zero completions <shell>
```

**Shell argument (required):** one of `bash`, `zsh`, `fish`, `powershell`, `elvish`.

**Behavior:**
- Print a shell completion script to stdout.
- Exit 0 on success; exit with usage error (exit code 2) for missing/unknown shell arguments.
- Support `-h` / `--help` on the subcommand.

**Help integration:**
- List `completions` under `Commands:` in `zero --help`.
- Document install examples in `zero completions --help`, e.g.:

  ```bash
  zero completions bash >> ~/.bashrc
  zero completions zsh  >> ~/.zshrc
  zero completions fish > ~/.config/fish/completions/zero.fish
  ```

**Completion coverage (minimum):**
- Top-level commands and documented aliases (`sessions`/`session`, `plugins`/`plugin`, `worktrees`/`worktree`, etc.).
- Nested subcommands (e.g. `zero daemon start`, `zero mcp oauth login`, `zero sandbox grants allow`).
- Root-level flags (`--add-dir`, `--theme`, `--skip-permissions-unsafe`, `-p`/`--prompt`, etc.).
- Common flags for heavily used commands such as `exec`.

**Non-goals for v1 (can be follow-ups):**
- Dynamic completion of model names, session IDs, or MCP server names from live config.
- Auto-install during `zero setup` or the install scripts.

## Why it belongs in Zero

Zero is a terminal-first coding agent. Shell tab completion is standard UX for CLI tools with many subcommands, and it directly improves discoverability of Zero's command surface without reading docs.

A `completions` command keeps completion scripts in sync with the CLI as commands evolve, instead of maintaining hand-written completion files separately. This matches the pattern already used by comparable tools (e.g. `grok completions`) and fits Zero's existing custom CLI architecture without adding new dependencies or changing the core language.

## Alternatives or workarounds considered

1. **Hand-maintained completion files in the repo** — Rejected because they drift from the CLI as commands and flags change.
2. **Third-party completion packages** — No widely maintained community package exists for Zero today.
3. **Documenting commands only in README** — Helps discovery but does not enable tab completion in the shell.
4. **Migrating the CLI to Cobra/urfave for auto-generation** — Too large a change for this feature alone; a lightweight generator inside the existing CLI is sufficient.

## Examples, screenshots, or prior discussion

**Comparable command in Grok:**

```text
$ grok --help
Commands:
  completions  Generate shell completion scripts (bash, zsh, fish, powershell, ...)

$ grok completions --help
Usage: grok completions [OPTIONS] <SHELL>
Arguments:
  <SHELL>  Target shell [possible values: bash, elvish, fish, powershell, zsh]
```

**Proposed Zero UX (mirroring the above):**

```text
$ zero --help
Commands:
  ...
  completions  Generate shell completion scripts (bash, zsh, fish, powershell, ...)
  ...

$ zero completions --help
Generate shell completion scripts (bash, zsh, fish, powershell, ...)

Usage:
  zero completions <shell>

Arguments:
  <shell>  Target shell (bash, elvish, fish, powershell, zsh)
```

**Suggested verification after implementation:**

```bash
zero completions --help
zero completions bash  | grep 'complete -F _zero zero'
zero completions zsh   | grep '#compdef zero'
zero completions fish  | grep 'complete -c zero'
zero completions bash  # should list daemon, mcp, exec subcommands
go test ./internal/cli -run Completions
```

anandh8x:
## Summary

- add `zero completions <shell>` for Bash, Zsh, Fish, PowerShell, and Elvish
- cover top-level aliases, nested subcommands, root flags, and common `exec` flags from one shared command tree
- add completion help, safe installation examples, and usage errors for missing or unsupported shells

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- pinned advisory lint against the PR diff
- `git diff HEAD --check`

Closes #499

## Summary by CodeRabbit

* **New Features**
  * Added a `completions` command to generate shell completion scripts.
  * Supports Bash, Zsh, Fish, PowerShell, and Elvish.
  * Updated CLI help to list `completions` and improved completion coverage for nested commands/aliases and common options.

* **Tests**
  * Added/expanded tests for `completions` help output, shell argument validation, and generation for every supported shell.
  * Added shell-specific structural checks (e.g., balanced blocks) and optional on-host script syntax validation when the shell is installed.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -368,6 +368,8 @@ func runWithDeps(args []string, stdout io.Writer, stderr io.Writer, deps appDeps
 	case "exec":
 		// Forward leading --add-dir occurrences so exec's own parser collects them.
 		return runExec(append(addDirFlagArgs(addDirs), args[1:]...), stdout, stderr, deps)
+	case "completions":
+		return runCompletions(args[1:], stdout, stderr)
 	case "daemon":
 		return runDaemon(args[1:], stdout, stderr, deps)
 	case "config":
@@ -1157,6 +1159,7 @@ Usage:
 
 Commands:
   exec       Run a one-shot prompt through the Go agent runtime
+  completions Generate shell completion scripts
   daemon     Manage the local background worker daemon (start/stop/status/run/attach)
   setup      Guide first-run provider setup
   config     Inspect resolved Go configuration without leaking secrets
--- a/internal/cli/completions.go
+++ b/internal/cli/completions.go
@@ -1,0 +1,399 @@
+package cli
+
+import (
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+)
+
+type completionNode struct {
+	names    []string
+	flags    []string
+	children []completionNode
+}
+
+type completionContext struct {
+	path        string
+	candidates  []string
+	transitions map[string]string
+}
+
+var completionRoot = completionNode{
+	flags: []string{"-h", "--help", "-v", "--version", "-p", "--prompt", "--add-dir", "--theme", "--skip-permissions-unsafe"},
+	children: []completionNode{
+		{names: []string{"exec"}, flags: []string{
+			"-h", "--help", "-f", "--file", "--image", "--add-dir", "--mode", "-m", "--model",
+			"--use-spec", "--spec-model", "--spec-reasoning-effort", "--max-turns", "--exec-profile",
+			"--auto", "--enabled-tools", "--disabled-tools", "--list-tools", "--profile", "-r",
+			"--reasoning-effort", "-C", "--cwd", "-w", "--worktree", "--worktree-dir", "-i",
+			"--input-format", "-o", "--output-format", "--prompt", "--resume", "--fork",
+			"--calling-session-id", "--calling-tool-use-id", "--tag", "--depth", "--session-title",
+			"--init-session-id", "--skip-permissions-unsafe", "--allow-escalation", "--self-correct",
+			"--notify", "--no-notify", "--no-completion-gate",
+		}},
+		{names: []string{"completions"}, flags: []string{"-h", "--help"}},
+		{names: []string{"daemon"}, children: leafNodes("start", "stop", "status", "run", "attach", "serve-remote", "link")},
+		{names: []string{"setup"}},
+		{names: []string{"config"}},
+		{names: []string{"models"}, children: []completionNode{{names: []string{"list", "ls"}}}},
+		{names: []string{"providers"}, children: []completionNode{
+			{names: []string{"current"}}, {names: []string{"list"}}, {names: []string{"catalog"}},
+			{names: []string{"add"}}, {names: []string{"check"}}, {names: []string{"use"}},
+			{names: []string{"remove", "rm"}}, {names: []string{"rename"}}, {names: []string{"setup"}},
+			{names: []string{"detect"}}, {names: []string{"models"}},
+		}},
+		{names: []string{"doctor"}},
+		{names: []string{"context"}},
+		{names: []string{"repo-map", "repomap"}},
+		{names: []string{"search", "find"}},
+		{names: []string{"sessions", "session"}, children: leafNodes("list", "children", "lineage", "tree", "rewind-plan", "rewind", "compact-plan")},
+		{names: []string{"spec"}, children: leafNodes("show", "approve", "reject")},
+		{names: []string{"init"}},
+		{names: []string{"specialists", "specialist"}, children: []completionNode{
+			{names: []string{"list"}}, {names: []string{"show"}}, {names: []string{"create"}},
+			{names: []string{"delete", "rm"}}, {names: []string{"edit"}}, {names: []string{"path"}},
+		}},
+		{names: []string{"plugins", "plugin"}, children: []completionNode{
+			{names: []string{"list"}}, {names: []string{"add"}}, {names: []string{"remove", "rm"}},
+		}},
+		{names: []string{"backends", "backend"}, children: leafNodes("doctor")},
+		{names: []string{"skills", "skill"}, children: []completionNode{
+			{names: []string{"list"}}, {names: []string{"add"}}, {names: []string{"info"}}, {names: []string{"remove", "rm"}},
+		}},
+		{names: []string{"tools", "tool"}, children: leafNodes("make", "list")},
+		{names: []string{"hooks"}, children: []completionNode{
+			{names: []string{"list"}}, {names: []string{"add"}}, {names: []string{"remove", "rm"}},
+			{names: []string{"enable"}}, {names: []string{"disable"}},
+		}},
+		{names: []string{"mcp"}, children: []completionNode{
+			{names: []string{"add"}}, {names: []string{"remove", "rm"}}, {names: []string{"enable"}},
+			{names: []string{"disable"}}, {names: []string{"check"}}, {names: []string{"list"}},
+			{names: []string{"permissions"}, children: leafNodes("list", "revoke", "clear")},
+			{names: []string{"tools"}, children: leafNodes("list")},
+			{names: []string{"oauth"}, children: leafNodes("login", "logout", "status")},
+		}},
+		{names: []string{"auth"}, children: leafNodes("openrouter", "chatgpt", "login", "logout", "status", "refresh")},
+		{names: []string{"sandbox"}, children: []completionNode{
+			{names: []string{"policy"}}, {names: []string{"setup"}}, {names: []string{"check"}},
+			{names: []string{"grants"}, children: leafNodes("list", "allow", "deny", "revoke", "clear")},
+		}},
+		{names: []string{"update"}},
+		{names: []string{"upgrade"}},
+		{names: []string{"worktrees", "worktree"}, children: leafNodes("prepare")},
+		{names: []string{"verify"}},
+		{names: []string{"trust"}, children: leafNodes("list", "remove")},
+		{names: []string{"eval"}, children: leafNodes("validate", "run", "bench")},
+		{names: []string{"changes", "change"}, children: []completionNode{
+			{names: []string{"inspect", "status"}}, {names: []string{"commit"}}, {names: []string{"push"}},
+			{names: []string{"pr", "pull-request"}},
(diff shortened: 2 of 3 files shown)


gnanam1990 (changes requested):
**REQUEST CHANGES** — the elvish script is syntactically invalid and cannot be sourced, and the test suite can't detect that. Everything else is in good shape; bash and zsh are genuinely solid.

Up front on method, because it matters for how much weight to give each part: I **executed** bash and zsh (both installed here) — syntax checks and real functional completion. fish, powershell and elvish are **not installed**, so those I reviewed by reading the generated output. The elvish finding below does not depend on running elvish: it's an objective brace count.

---

## [Blocker] The elvish script is unparseable — every `if`/`elif` block is opened and never closed

`writeElvishCompletions` (`internal/cli/completions.go:376-384`, and again at `:388-396`) emits an opening `{` per branch but only writes one closing `}` after the whole loop:

```go
fmt.Fprintf(w, "        %s (== $context'|'$token %q) {\n            set context = %q\n", keyword, key, transitions[key])
```

Generated output:

```elvish
        if (== $context'|'$token "auth|chatgpt") {
            set context = "auth chatgpt"
        elif (== $context'|'$token "auth|login") {
            set context = "auth login"
        elif ...
```

Counting the emitted file: **379 `{` against 4 `}`**. The transition loop alone opens 1 `if` + 187 `elif` blocks and closes exactly one. The candidates section at the bottom has the identical defect:

```elvish
    elif (== $context "worktrees") {
        put prepare
    elif (== $context "worktrees prepare") {
        put 
    }
}
```

Elvish requires each block closed before the next keyword — `if $cond { body } elif $cond2 { body2 }`. As emitted, this cannot parse, so `zero completions elvish` produces a file that fails on source. For comparison, the other four are all balanced: bash 3/3, zsh 3/3, powershell 384/384, and fish correctly uses one `switch` … 188 `case` … one `end`.

**There is a second, independent bug in the same function.** Elvish's `==` is *numeric* equality (`eq` is the general value comparison; older releases used `==s` for strings). `(== $context'|'$token "auth|chatgpt")` compares strings with a numeric operator, so even with balanced braces every comparison would fail at runtime.

**Fix:** close each branch and switch the operator — `} elif (eq $context'|'$token "…") {` — with a final `}` after the chain. Same in both loops.

For what it's worth, `$args[1..-1]` is correct and I'd leave it alone: Elvish ranges are upper-exclusive, so this deliberately drops the word currently being completed, which is what a completer wants.

## [Major] The tests cannot detect a malformed script — which is why the above shipped green

`TestCompletionsGeneratesEverySupportedShell` (`internal/cli/completions_test.go:63-91`) asserts only three things per shell: exit success, one marker substring, and the four substrings `daemon` / `mcp oauth` / `sandbox grants` / `--output-format`. The broken elvish output contains **all five** — I checked each individually — so `go test ./internal/cli/ -run TestCompletion` passes clean against a file that no elvish can parse.

That's the gap worth closing regardless of the blocker: the suite validates *content* and never *shape*, so any future structural regression in any of the five emitters lands silently.

**Fix, cheapest first:**
1. Assert structural balance per shell in the existing table — brace balance for elvish/powershell, `switch`/`end` for fish. That alone catches this class.
2. Better, for the two shells CI already has: shell out to `bash -n` and `zsh -n` on the generated script, skipping via `exec.LookPath` when absent. I ran both here and they pass, so this is a cheap, non-flaky guard.

## Minor

Agreed with CodeRabbit's two, and I confirmed the first empirically rather than by reading — sourcing the generated bash script and driving the completion function gives:

```
[zero completions  | cword=2]  ->  -h --help
```

so the completion generator can't complete its own shell argument. The five shell names are a static list already present in the `switch` at `:125-138`; adding `children: leafNodes("bash","zsh","fish","powershell","elvish")` to the `completions` node at `:35` fixes it.

## Nit

Nothing fails when a new CLI command is added without updating `completionRoot`. The tree is accurate *today* (see below), but it's hand-maintained with no guard, so it will drift silently. A test comparing the root node's names against the command list in `app.go` would pin it.

---

## What's genuinely good here

I want to be concrete about this rather than polite, because the parts that work are the parts that were hard:

- **bash is correct, not just syntactically valid.** `bash -n` passes, and driving `_zero` directly with `COMP_WORDS`/`COMP_CWORD` returns the right candidates at every position I tried — top level, a partial prefix (`zero ex` → `exec`), a nested subcommand (`zero sessions ` → `list children lineage tree rewind-plan rewind compact-plan`), and flags after a subcommand (`zero exec --` → the full flag set). zsh passes `zsh -n`.
- **The command tree is accurate and complete.** Every one of the 33 commands in `zero --help` is completed — no gaps. It also completes 13 undocumented aliases (`backend`, `change`, `init`, `plugin`, `repoinfo`, `repomap`, `session`, `skill`, `specialists`, `tool`, `trust`, `upgrade`, `worktree`), and I ran each one against the built binary to confirm they're all real, working aliases rather than phantom entries. That's better than most hand-maintained completion trees manage.
- **One shared tree feeding five emitters** is the right structure, and `completionContexts` handling alias expansion into multiple paths is a nice touch — it's why `zero session <TAB>` and `zero sessions <TAB>` both work.
- Everything interpolated into the scripts is a compile-time constant from `completionRoot` — nothing from argv, env, or the filesystem — so there's no injection surface here at all. I looked specifically because sourced shell scripts are a natural place for it.

## Verification

`gofmt` clean on all three files, `go vet ./internal/cli/` clean, `go test ./internal/cli/ -run TestCompletion` passes (which is precisely the problem — see the Major). All five shells generate non-empty output with exit 0. No pre-existing failures were implicated; I didn't need a base comparison since nothing failed.

gnanam1990 (approved):
**APPROVE.** Re-reviewed at `c0a1cbe`. The blocker is fixed on both counts, the test gap is closed better than I suggested, and both minors are done.

I went further than reading the diff this round: **I installed elvish 0.21.0 and ran the real interpreter against both versions of the script**, so the central claim is now verified rather than inferred.

## Blocker — fixed, and confirmed with real elvish

Both halves are addressed:

- **Brace structure.** The generator now emits `} elif (…) {`, and the file balances at **389 `{` / 389 `}`** (was 379/4).
- **Operator.** `(== …)` → `(eq …)`: zero occurrences of `(== ` remain, 387 of `(eq `. This was the half that would still have failed at runtime even with balanced braces, so I'm glad it was caught too.

Verified by execution, not inspection:

```
OLD (056373f):  elvish -compileonly  ->  Multiple parse errors: should be '}'
NEW (c0a1cbe):  elvish -compileonly  ->  parses and compiles cleanly
```

One note so the output isn't misread if you try this yourself: running `elvish -compileonly` on the script as-shipped reports `cannot find variable $edit:completion:arg-completer`. That is **not** a defect — `edit:` is the interactive editor module and isn't bound in non-interactive mode. Stubbing line 2 to a plain `var` isolates syntax from module availability, and with that done the new script compiles clean while the old one still fails on parse errors. So the fix is genuinely correct, not merely balanced.

If you ever want elvish in CI, `elvish -compileonly` is the equivalent of `bash -n` and would slot straight into `assertNativeShellSyntax` — though pulling in an elvish binary is a heavier lift than the two shells already covered, so I wouldn't push for it.

## Major — fixed, and mutation-tested

Both suggestions landed, plus one I didn't think of:

- `assertBalancedBraces` for powershell/elvish (`completions_test.go:93`)
- `assertNativeShellSyntax` shelling out to `bash -n` / `zsh -n`, gated on `exec.LookPath` with `t.Skipf` when the shell is absent (`:135`) — including the skip detail, so this won't turn into a flaky Windows failure
- `assertBalancedFishBlocks` (`:101`), a proper depth counter over `function`/`for`/`switch`/`if` vs `end`, which is more precise than the brace count I proposed and correctly models fish's grammar

The important question is whether the new guard would actually have caught the original bug, so I mutation-tested it — restored the old generator, kept the new tests:

```
--- FAIL: TestCompletionsGeneratesEverySupportedShell/elvish
    completions_test.go:97: unbalanced braces: 379 opening, 4 closing
```

It fires, with the exact numbers from my review. That's a real regression test, not a decorative one.

## Minors — both fixed

- `zero completions <TAB>` now completes the shell list. Verified functionally by sourcing the generated bash script and driving `_zero`: `[zero completions |cw=2] -> -h --help bash zsh fish powershell elvish`. `TestCompletionTreeCoversAliasesNestingAndCommonFlags` now pins it too.
- The elvish install example is in the help (`eval (zero completions elvish | slurp)`) and asserted in `TestCompletionsHelpAndRootHelp`.

## One nit, non-blocking

`assertBalancedBraces` counts braces anywhere in the text, including inside quoted strings. That's sound today because every command name, alias and flag in `completionRoot` is brace-free, so the count can't be skewed. But it would silently mis-handle a future command containing `{` or `}`. Not worth changing now — the fish checker's line-oriented approach is the model if it ever matters.

## Verification

Re-ran everything on a clean tree at `c0a1cbe`:

- All five shells generate non-empty output, exit 0; all five balance.
- `bash -n` and `zsh -n` pass. bash is still functionally correct — spot-checked top level, nested (`zero sessions ` → the seven subcommands), and prefix filtering (`zero exec --mod` → `--mode --model`).
- `gofmt` and `go vet ./internal/cli/` clean.
- `go test ./internal/cli/ -run TestCompletion` — all four tests pass, including the five new per-shell subtests.

Good turnaround, and the elvish fix addressed the operator bug rather than just silencing the brace count — which is the difference between it parsing and it working.

kevincodex1 (approved):

PR #765, opened 2026-07-20
fix(sandbox): bind Windows elevated ACL setup to one no-follow handle (#728)

### LINKED ISSUE #728: security(windows): elevated sandbox ACL setup has a path replacement race
## Summary

Elevated Windows sandbox setup inspects and modifies ACL targets through separate pathname operations without binding those operations to a no-follow handle or stable file identity.

A lower-privileged process able to rename or replace a user-controlled workspace/ACL target with a junction during setup may redirect inherited capability-SID ACLs onto an unintended directory.

## Affected code

- `internal/sandbox/windows_setup_windows.go:12-18` requires elevated setup.
- `internal/cli/sandbox.go:127-175` collects workspace/profile paths before launching the elevated helper.
- `internal/sandbox/windows_acl_apply_windows.go:75-135` performs `Stat`, `GetNamedSecurityInfo`, and `SetNamedSecurityInfo` by pathname.
- `internal/sandbox/windows_acl_apply_windows.go:147-183` installs inheritable capability-SID ACEs.
- `internal/sandbox/windows_runner.go:573-607` places the capability SID in later restricted tokens.

## Race outline

1. Begin elevated sandbox setup for a path the invoking user can rename or replace.
2. Between pathname validation and ACL mutation, replace a path component or target with a junction/reparse point.
3. The elevated helper reopens the pathname and may apply the capability ACL to the replacement target.
4. Later restricted sandbox tokens containing that SID may receive unintended access.

## Impact

If the race is achievable against a privileged target, this can become a local privilege-escalation primitive by granting the sandbox capability SID inherited access to a directory selected through path replacement.

This finding was established by static review and still needs a focused Windows reproduction across the actual privilege boundary.

## Suggested fix

- Open targets without following reparse points.
- Reject unexpected reparse tags in user-controlled path components.
- Retain handles/file IDs across security-descriptor read and write.
- Verify object identity immediately before mutation and rollback.
- Add a Windows integration test that attempts target replacement during elevated setup.

Vasanthdev2004:
Closes #728.

## The bug

The Windows elevated ACL setup resolved each target's pathname independently: `os.Stat`, then `GetNamedSecurityInfo`, then `SetNamedSecurityInfo` (and again by pathname on rollback). Between those calls a lower-privileged local user could swap a target for a symlink or junction, so the elevated process would apply the ACL change to a different object than the one it stat'd and read. That is a TOCTOU across a privilege boundary. The targets are user-influenced (workspace root plus additional write roots) and setup runs elevated, so the local-privesc model is coherent.

## The fix

Open each target once with a no-follow handle and drive both the read and the write through it:

- `CreateFile` with `FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS` and `READ_CONTROL | WRITE_DAC`.
- `GetSecurityInfo(handle)` / `SetSecurityInfo(handle)` instead of the pathname (`*Named*`) variants, so read and write hit the same kernel object.
- `isDir` comes from that handle, not a separate `Stat`.
- A target that resolves to a reparse point is refused outright (a setup target that is a symlink/junction during elevated setup is the swap signature).
- Rollback re-opens no-follow rather than restoring by pathname. The residual window is small (the target is ACL-restricted by then), and I deliberately did not retain handles for the sandbox lifetime because one caller (`windows_command_runner_windows.go:114`) discards the rollback closure, so retained handles would leak.

## Tests

All windows-tagged, no elevation required (the current user holds WRITE_DAC on its own temp dirs):

- handle-based apply plus rollback round-trip on a temp dir,
- the materialize-and-remove path,
- reparse rejection via a **junction** (needs no admin, so it actually runs in CI, unlike a symlink),
- the not-exist mapping (so the materialize path still fires).

The full privilege-boundary race cannot be reproduced in CI, so the guard is verified structurally (one handle for read and write) and by junction rejection.

## For review

One judgment call worth a maintainer eye: rejecting reparse-point targets assumes setup targets are real directories. If a workspace root is legitimately a junction, that would need revisiting. I went with reject-during-elevated-setup as the safe default.

Verified locally on Windows: `go build ./...`, `go vet ./internal/sandbox/...`, and `go test ./internal/sandbox/...` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows ACL handling to prevent path-related race conditions.
  * Rejects symbolic links, junctions, and other reparse points when applying ACLs.
  * Ensures ACL rollback targets the same underlying file or directory.
  * Preserves correct handling of missing targets and materialized directories.

* **Tests**
  * Added coverage for ACL application, rollback, target creation and cleanup, missing paths, files, directories, and reparse points.

### DIFF
--- a/internal/sandbox/windows_acl_apply_windows.go
+++ b/internal/sandbox/windows_acl_apply_windows.go
@@ -77,11 +77,18 @@ func applyWindowsACLPathGroup(group windowsACLPathGroup) (windowsACLSnapshot, bo
 	if path == "" || len(group.Entries) == 0 {
 		return windowsACLSnapshot{}, false, nil
 	}
+	// Open ONE no-follow handle to the target and drive every ACL operation
+	// (read + write) through it, so the read and the write hit the same kernel
+	// object. The previous pathname-based Stat/GetNamedSecurityInfo/
+	// SetNamedSecurityInfo each re-resolved the path independently, so during
+	// elevated setup a lower-privileged local user could swap the target for a
+	// symlink/junction between operations and redirect the ACL change onto a
+	// system object it never validated (issue #728, a TOCTOU privilege boundary).
 	materialized := false
-	info, err := os.Stat(path)
+	handle, isDir, err := openWindowsACLTarget(path)
 	if err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
-			return windowsACLSnapshot{}, false, fmt.Errorf("stat windows ACL target %s: %w", path, err)
+			return windowsACLSnapshot{}, false, err
 		}
 		if !group.Materialize {
 			if windowsACLGroupRequiresExistingTarget(group) {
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):
**APPROVE.** This is the textbook-correct fix for the class of bug #728 describes, and the test coverage is better than most Windows-specific changes get. One scoping question below that I'd like answered rather than changed.

Upfront on method: I can't execute this — it's `//go:build windows` and I'm on macOS. I cross-compiled (`GOOS=windows go build` / `go vet ./internal/sandbox/` — both clean) and reviewed by reading. Where I had a runtime doubt I leaned on your CI rather than guessing; noted below.

## The fix is the right shape

Replacing `os.Stat` + `GetNamedSecurityInfo` + `SetNamedSecurityInfo` — three independent pathname resolutions — with a single `CreateFile` handle driving `GetSecurityInfo`/`SetSecurityInfo` is exactly how you close a TOCTOU on Windows ACL work. The read and the write now provably hit the same kernel object, which is the property that matters and the one pathname-based APIs cannot give you.

Details that are easy to get wrong and aren't:

- `FILE_FLAG_BACKUP_SEMANTICS` so directories can be opened at all, alongside `FILE_FLAG_OPEN_REPARSE_POINT`.
- `READ_CONTROL|WRITE_DAC` — the minimal rights for the job, rather than reaching for `GENERIC_ALL`.
- The `fail` closure consolidating cleanup. The previous code repeated `if materialized { os.RemoveAll(path) }` at five returns; collapsing that into one helper that also closes the handle is what keeps the new early-return paths from leaking. Worth noting it also *adds* cleanup that was missing before — the materialized-then-open-failed path now removes the target, where the old code returned without it.
- Rollback re-opens through the same `openWindowsACLTarget` rather than restoring by pathname, so the restore can't be redirected either. The comment is honest that a window remains and explains why it's acceptable ("the target is ACL-restricted by now"), which I'd much rather read than a claim that it's airtight.

## The test coverage is genuinely good

Five real tests, not fakes — and `TestOpenWindowsACLTargetRejectsReparsePoint` uses `mklink /J` deliberately because a junction needs neither admin nor Developer Mode, so the security guard **actually executes in CI** instead of skipping. That's the difference between testing the property and gesturing at it, and Windows smoke is green on this PR, so it really ran.

That also settles the one runtime doubt I had: `GetFileInformationByHandle` on a handle opened with only `READ_CONTROL|WRITE_DAC` is the kind of thing that fails with `ERROR_ACCESS_DENIED` on some configurations, and if it did, every ACL setup would break rather than just the attack path. Your passing Windows job answers that better than I could by reading.

## The question: intermediate path components

`FILE_FLAG_OPEN_REPARSE_POINT` suppresses reparse-point traversal for the **final component only**. An attacker who can create a junction at a *parent* directory of an ACL target still redirects the open, and `GetFileInformationByHandle` on the resulting handle reports the real object's attributes — no reparse-point flag, so the check passes and the ACL lands on the wrong object. Windows has no per-component no-follow for `CreateFile`; the usual answer is to open the parent chain step-by-step, or to require the whole chain be admin-owned.

So: **does the #728 threat model include a lower-privileged user who controls a parent directory of any ACL target?** The targets come from `entry.Path` on `WindowsACLEntry`, which can include workspace paths, so it depends where those live. If every target sits under an admin-owned root, the residual is unreachable and this is complete as written — worth a sentence in the comment saying so, since the current text scopes the guarantee to "final-component" without saying why that suffices. If workspace paths can sit somewhere another local user can create directories, the guard is a real improvement but not yet a closed boundary, and that's worth a follow-up issue rather than expanding this PR.

Either way this strictly improves on the pathname-based code it replaces, so I don't think it should block.

## Verification

`GOOS=windows GOARCH=amd64 go build ./internal/sandbox/` and `go vet` both clean. All nine checks green on the PR, including Windows smoke, which is the job that actually exercises the new code. I did not run the Windows tests locally and am not claiming to have.

anandh8x (approved):
Approved on the latest head. ACL reads and writes are bound to one no-follow handle, with reparse-point rejection and rollback coverage. Windows cross-compilation succeeds and all checks are green.

PR #766, opened 2026-07-20
fix(lsp): make the real-gopls check opt-in so a broken gopls can't fail the suite (#684)

### LINKED ISSUE #684: Real gopls manual test can fail the default suite on an unusable installation

Vasanthdev2004:
Closes #684.

`TestManagerCheckRealGopls` skipped only when `gopls` was absent from PATH, so an installed-but-unhealthy `gopls` (the issue's repro: an unusable persistent-index cache makes the server exit during init with `EOF`) failed `go test ./...` for every developer who merely has `gopls` on PATH.

Gate it behind `ZERO_GOPLS_INTEGRATION`, matching the repo's existing real-server opt-ins (`internal/dictation` smoke tests use `ZERO_STT_*`). The default hermetic suite keeps its coverage on the fake server and never touches the developer's global `gopls`; the real check still runs on demand:

```
ZERO_GOPLS_INTEGRATION=1 go test ./internal/lsp -run '^TestManagerCheckRealGopls$'
```

Verified: default `go test ./internal/lsp` now skips the real check unconditionally (before even probing PATH), and the opt-in path runs it. gofmt clean, package green.

## Summary by CodeRabbit

* **Tests**
  * Real-language-server integration checks are now explicitly opt-in and run only when enabled.
  * The default test suite no longer fails due to an unavailable or unhealthy installed language server detected on PATH.

### DIFF
--- a/internal/lsp/real_gopls_manual_test.go
+++ b/internal/lsp/real_gopls_manual_test.go
@@ -11,9 +11,16 @@ import (
 
 // TestManagerCheckRealGopls drives the exact path the self-correct LSP wiring
 // uses (NewManager -> Check) against a REAL gopls and asserts a real diagnostic
-// comes back. Skips when gopls is not installed. Temporary manual-verification
-// test (the rest of the suite uses a fake server).
+// comes back. It is an OPT-IN integration check: an installed-but-unhealthy
+// gopls (e.g. an unusable persistent-index cache) exits during initialization
+// and would otherwise fail `go test ./...` for every developer who merely has
+// gopls on PATH (issue #684). So it runs only when ZERO_GOPLS_INTEGRATION is
+// set, matching the repo's other real-server opt-ins (internal/dictation smoke
+// tests). The default suite keeps its coverage on the fake server.
 func TestManagerCheckRealGopls(t *testing.T) {
+	if os.Getenv("ZERO_GOPLS_INTEGRATION") == "" {
+		t.Skip("set ZERO_GOPLS_INTEGRATION=1 to run the real gopls integration check")
+	}
 	if _, err := exec.LookPath("gopls"); err != nil {
 		t.Skip("gopls not on PATH; skipping real-server check")
 	}
--- a/internal/lsp/real_gopls_manual_test.go
+++ b/internal/lsp/real_gopls_manual_test.go
@@ -14,11 +14,13 @@ import (
 // comes back. It is an OPT-IN integration check: an installed-but-unhealthy
 // gopls (e.g. an unusable persistent-index cache) exits during initialization
 // and would otherwise fail `go test ./...` for every developer who merely has
-// gopls on PATH (issue #684). So it runs only when ZERO_GOPLS_INTEGRATION is
-// set, matching the repo's other real-server opt-ins (internal/dictation smoke
-// tests). The default suite keeps its coverage on the fake server.
+// gopls on PATH (issue #684). So it runs only when ZERO_GOPLS_INTEGRATION=1,
+// matching the repo's other real-server opt-ins (internal/dictation smoke
+// tests). Requiring exactly "1" (not merely non-empty) means an explicit
+// ZERO_GOPLS_INTEGRATION=0 keeps the integration check OFF, as its value implies.
+// The default suite keeps its coverage on the fake server.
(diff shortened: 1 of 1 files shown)


Vasanthdev2004:
Done in `75f559f3`, thanks. Now requires exactly `ZERO_GOPLS_INTEGRATION=1`, so an explicit `=0` keeps the integration check off as its value implies. Verified: the default suite and `=0` both skip; `=1` runs it.

gnanam1990 (approved):
**APPROVE** — the change is correct and does exactly what #684 needs. But it collides with an existing open PR, and that needs a decision before either lands.

## This duplicates #691

@PierrunoYT's #691 ("fix(lsp): make real-gopls integration test opt-in via env var") changes **the same file** with **the same mechanism** for **the same issue**, differing only in the variable name:

| PR | gate |
|---|---|
| #691 | `os.Getenv("ZERO_LSP_REAL_GOPLS_TEST") != "1"` |
| #766 (this) | `os.Getenv("ZERO_GOPLS_INTEGRATION") != "1"` |

Both touch only `internal/lsp/real_gopls_manual_test.go`. Whichever merges first turns the other into a conflict, and if both somehow land the test ends up with two gates. #691 is older and currently sits at CHANGES_REQUESTED. Worth someone picking one and closing the other rather than reviewing them in parallel — I have no stake in which, though #691's `ZERO_LSP_...` prefix matches the `ZERO_<AREA>_<THING>` shape of `ZERO_SANDBOX_REAL_SMOKE` a little more closely.

## One factual correction to the comment

The comment justifies the exact-`"1"` check by saying it matches "the repo's other real-server opt-ins (internal/dictation smoke tests)". That's not quite right — dictation's boolean gate is:

```go
if os.Getenv("ZERO_STT_DOWNLOAD_TEST") == "" {   // non-empty, not == "1"
```

so under dictation's convention `ZERO_STT_DOWNLOAD_TEST=0` would *enable* the test, which is the opposite of the behaviour the comment argues for. The precedent that actually matches is `internal/sandbox/runner_windows_integration_test.go`:

```go
if os.Getenv("ZERO_SANDBOX_REAL_SMOKE") != "1" {
```

The chosen semantics are the better ones — `=0` keeping it off is genuinely less surprising — so this is a citation fix, not a behaviour one. Pointing `ZERO_SANDBOX_REAL_SMOKE` at it instead would make the comment accurate and the precedent stronger.

## Verification

`go vet ./internal/lsp/` clean; `os` was already imported so nothing else was needed. `go test ./internal/lsp/ -run TestManagerCheckRealGopls` skips as intended with the var unset. The diagnosis is right, too — the test genuinely ran for anyone with gopls on PATH, so an unhealthy install failed `go test ./...` for them.

anandh8x (approved):
Approved on the latest head. The real gopls test now requires the explicit ZERO_GOPLS_INTEGRATION=1 opt-in, so the default suite remains hermetic. Focused tests and all checks pass.

PR #767, opened 2026-07-20
fix(cli): warn when ZERO_PROVIDER overrides a providers-use selection (#721)

### LINKED ISSUE #721: providers use reports success when ZERO_PROVIDER keeps another provider active
### Version / commit

Zero commit `18cce358a29409ab937e92ff9af77bf98749f6aa` (`v0.4.0-4-g18cce35`)

### OS and environment

- Windows 11
- PowerShell 7.6.3
- User environment contains `ZERO_PROVIDER=anthropic`
- Saved config contains `anthropic` and `chatgpt` profiles

### Steps to reproduce

1. Configure `activeProvider` as `anthropic` and set the user-level environment variable `ZERO_PROVIDER=anthropic`.
2. Run `zero providers use chatgpt`.
3. Observe the success output:
   ```text
   Active provider set to chatgpt
   next: zero providers check chatgpt
   ```
4. Run `zero providers current`.

### Actual behavior

`providers use` reports success and updates `config.json` to `activeProvider: chatgpt`, but `providers current` still reports `anthropic` because `ZERO_PROVIDER` silently takes precedence.

A related inconsistency occurs with environment-derived profiles: `providers list` displays profiles such as `openai`, but `zero providers use openai` fails with:

```text
[zero] provider "openai" not found
```

The list combines persisted and environment-derived profiles, while `providers use` only accepts persisted profiles and does not identify that distinction.

### Expected behavior

At minimum, `providers use` should report that the saved selection is not effective because `ZERO_PROVIDER` overrides it, for example:

```text
Saved provider: chatgpt
Effective provider: anthropic
Overridden by ZERO_PROVIDER
```

Ideally:

- `providers list --json` identifies profile source (`config`, `environment`, etc.) and whether the profile can be passed to `providers use`.
- Human-readable `providers list` visually distinguishes transient/environment-derived profiles from persisted profiles.
- `providers use` either makes the requested provider effective or exits with a warning/error when a higher-priority override prevents that.

### Additional context

The behavior follows the resolver precedence in `config.applyEnv`, but the command output makes the saved and effective provider appear identical. This is especially confusing for editor/GUI integrations that use `providers list --json` to populate a provider selector.

Related: #707 reports a similar persisted-versus-resolved mismatch when deleting an environment-derived OpenAI provider.

Vasanthdev2004:
Closes #721.

## The bug

`zero providers use <name>` writes `activeProvider` to config.json and reports `Active provider set to <name>`, even when `ZERO_PROVIDER` is set. `applyEnv` (resolver.go) makes `ZERO_PROVIDER` win over config.json unconditionally, so the reported switch is a silent no-op: `providers current` still shows the env-pinned provider.

## The fix

Detect when `ZERO_PROVIDER` is set and names a different provider than the selection, and say so:

- **text**: a note to stderr naming the env var and the provider that stays active, and how to make the selection effective (`unset ZERO_PROVIDER`).
- **--json**: adds `effectiveProvider` and `overriddenByEnv` so a machine consumer is not misled either.

The config write is unchanged and still lands; only the reporting becomes honest. No warning when `ZERO_PROVIDER` is unset or already names the selection.

The env read is injected via a new `appDeps.getenv` (`os.Getenv` in production, deliberately left nil by `fillAppDeps` like `exportActiveProvider`) so the existing tests stay hermetic against a developer's ambient `ZERO_PROVIDER` (the reporter's setup) rather than newly failing on it.

## Scope note

The issue also mentions `providers use openai` failing with "not found" for an env-derived profile. That path is already handled by the unpersisted-provider reporting (#707) and the env-derived-profile fix in #716, so this PR closes the remaining gap: the silent override on a successful switch.

Tests: text warning, JSON override fields, and the two no-warn cases (unset / matches). Package builds and the provider-use suite is green. (Pre-existing, unrelated `TestBuildServeScopeKeepsLexicalPaths` fails on main independently of this change.)

## Summary by CodeRabbit

* **New Features**
  * Added visibility into provider selections overridden by the `ZERO_PROVIDER` environment variable.
  * Non-JSON output now warns when the environment setting takes precedence.
  * JSON output includes the requested provider, effective provider, and override source when applicable.

* **Bug Fixes**
  * Improved clarity around provider configuration conflicts, preventing misleading selection results.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -58,7 +58,11 @@ type appDeps struct {
 	// config.SetActiveProviderEnv, set in defaultAppDeps — deliberately NOT filled
 	// by fillAppDeps, so tests never mutate the process environment unless they
 	// inject it). nil ⇒ no export.
-	exportActiveProvider   func(providerName string)
+	exportActiveProvider func(providerName string)
+	// getenv reads a process environment variable (production: os.Getenv, set in
+	// defaultAppDeps — deliberately NOT filled by fillAppDeps, so tests are hermetic
+	// against ambient vars like ZERO_PROVIDER unless they inject it). nil ⇒ empty.
+	getenv                 func(string) string
 	probeProviderHealth    func(context.Context, providerhealth.Options) providerhealth.Result
 	discoverProviderModels func(context.Context, config.ProviderProfile) ([]providermodeldiscovery.Model, error)
 	detectLocalRuntimes    func(context.Context, provideronboarding.LocalDetectOptions) []provideronboarding.DetectedLocalRuntime
@@ -118,6 +122,7 @@ func defaultAppDeps() appDeps {
 		stdin:                os.Stdin,
 		userConfigPath:       config.DefaultUserConfigPath,
 		exportActiveProvider: config.SetActiveProviderEnv,
+		getenv:               os.Getenv,
 		resolveConfig: func(workspaceRoot string, overrides config.Overrides) (config.ResolvedConfig, error) {
 			options, err := config.DefaultResolveOptions(workspaceRoot)
 			if err != nil {
--- a/internal/cli/provider_onboarding.go
+++ b/internal/cli/provider_onboarding.go
@@ -74,11 +74,18 @@ func runProvidersUse(args []string, stdout io.Writer, stderr io.Writer, deps app
 		return writeAppError(stderr, err.Error(), exitCrash)
 	}
 
+	override := activeProviderEnvOverride(deps.getenv, cfg.ActiveProvider)
 	if options.json {
-		if err := writePrettyJSON(stdout, map[string]any{
+		payload := map[string]any{
 			"activeProvider": cfg.ActiveProvider,
 			"configPath":     configPath,
-		}); err != nil {
+		}
+		if override != "" {
+			// A JSON consumer must not read this as an effective switch either.
+			payload["effectiveProvider"] = override
+			payload["overriddenByEnv"] = config.ActiveProviderEnv
+		}
+		if err := writePrettyJSON(stdout, payload); err != nil {
 			return exitCrash
 		}
 		return exitSuccess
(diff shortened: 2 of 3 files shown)


gnanam1990 (approved):
**APPROVE.** The premise checks out, the dependency plumbing follows the file's existing convention, and both output modes are covered.

## The load-bearing claim is correct

The warning is only worth printing if `ZERO_PROVIDER` really does beat the config write. It does, unconditionally — `applyEnv` at `internal/config/resolver.go:560-563`:

```go
activeProvider := strings.TrimSpace(envValue(env, ActiveProviderEnv))
if activeProvider != "" {
    cfg.ActiveProvider = activeProvider
}
```

No precedence check, no merge — the env value simply replaces whatever `providers use` just persisted. So reporting the write as a bare success genuinely does read as a switch that silently has no effect, which is #721.

`strings.EqualFold` for the "is it actually different" test is the right comparison rather than a stricter one: provider-name resolution is case-insensitive elsewhere in the resolver (`resolver.go:467`), so `ZERO_PROVIDER=OpenAI` against a saved `openai` resolves to the same profile and correctly produces no warning. A case-sensitive check would have emitted a spurious warning there.

## The dependency-injection shape is right

`getenv` follows the same pattern as the existing `exportActiveProvider` directly above it — set in `defaultAppDeps()`, deliberately **not** filled by `fillAppDeps`. I checked both ends of that:

- Production is fine: `Run()` (`app.go:116`) passes `defaultAppDeps()`, which sets `getenv: os.Getenv`, so the warning actually fires for real users. Worth confirming explicitly, since a dep that `fillAppDeps` skips could easily have ended up nil on the live path and made this a silent no-op.
- Tests stay hermetic: `fillAppDeps` leaves it nil, and `activeProviderEnvOverride` returns `""` for a nil getenv, so an ambient `ZERO_PROVIDER` cannot make the new warning appear in unrelated tests.

Handling the JSON path separately with `effectiveProvider` / `overriddenByEnv` rather than only writing to stderr is the right call — a script consuming `--json` would otherwise have no way to see the override.

## Unrelated observation, since it's adjacent to this PR's subject

`internal/cli` is **not** hermetic against an ambient `ZERO_PROVIDER` today, through a different path than the one this PR adds. With `ZERO_PROVIDER=leaky` exported, `TestRunProvidersUseEnvDerivedJSONIncludesConfigPath` fails on current `main`:

```
exit code = 3, want 0: [zero] no active provider configured: active provider "leaky" not found
```

That comes from `resolveConfig` reading the real process environment, not from anything here — I verified it reproduces identically on `upstream/main` without this branch. Flagging it only because you're the person currently in this code: since `zero` itself exports `ZERO_PROVIDER` via `SetActiveProviderEnv`, a developer can plausibly end up with it set and then see a confusing local test failure. A `t.Setenv(config.ActiveProviderEnv, "")` in the affected tests would close it. Separate PR, not this one.

## Verification

`gofmt` and `go vet ./internal/cli/` clean. All three new tests pass, including the two-case `NoWarnWithoutEnvOverride` (env unset / env matches). Full `internal/cli` suite compared against current `main`: **identical** five failures on both (`TestExecScopeReRegistration…`, `TestRunAddDirDispatch…`, both `TestRunDoctor…`, `TestRunSandboxCheckJSON…`) — all sandbox/network artifacts of my environment, and this PR introduces none.

Vasanthdev2004:
@coderabbitai review

anandh8x (approved):
Approved on the latest head. providers use now reports a conflicting ZERO_PROVIDER override in both text and JSON output while preserving the saved selection. Focused tests and all checks pass.

PR #768, opened 2026-07-20
fix(tools): give write_stdin's invalid-session errors the same recovery guidance (#749 follow-up)

Vasanthdev2004:
Follow-up to #749 / #702.

## The issue

`write_stdin` sends keystrokes to a running `exec_command` session; its `session_id` must be an id (>= 1) returned by a still-running session. It had **three** "no valid live session" error paths with **three different messages and signatures**:

| input | old error |
|---|---|
| `session_id` missing | `session_id is required` |
| `session_id: 0` (or negative / non-integer) | `session_id must be at least 1` |
| `session_id: 5`, no live session | rich recovery guidance (`UnknownExecSessionError`) |

A model that calls `write_stdin` with `session_id: 0` (the common "I have no real id" placeholder, and the reported symptom) got the terse `must be at least 1`, which:

- gives **no way forward** (unlike the rich message, it doesn't say "start a session or edit files directly"), and
- names a minimum, which **nudges the model to try `1`, then `2`...**, the exact id-probing #749 works to suppress.

And because the three cases have different signatures (`errorSignature` keys on the normalized 80-char prefix), a model moving `0 -> 1 -> 2` **resets the repeated-failure streak** on each class of mistake, delaying the halt.

## The fix

Route the missing / non-integer / `< 1` cases through the same `UnknownExecSessionError` message the no-live-session case already uses. Now:

- every "no live session" mistake gives the **same recovery guidance**, and
- because that message is **id-invariant**, they share **one repeated-failure signature**, so any mix of missing / zero / probing accumulates toward the same halt (extending #749's intent to the `0`/missing entry points).

No behavior change for a valid `session_id`. `TestWriteStdinRequiresPositiveSessionID` now covers missing/zero/negative/non-integer and asserts the recovery message; the guardrail signature/halt tests (`TestUnknownExecSessionErrorSignatureIsIDInvariant`, `TestUnknownExecSessionProbingTripsFailureHalt`) are unchanged and pass. Full `internal/tools` package green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for invalid or missing execution session IDs.
  * All invalid session ID inputs now provide consistent recovery guidance instead of terse validation errors.

### DIFF
--- a/internal/tools/exec_command.go
+++ b/internal/tools/exec_command.go
@@ -775,12 +775,20 @@ func (tool writeStdinTool) Run(ctx context.Context, args map[string]any) Result 
 }
 
 func (tool writeStdinTool) RunWithOptions(ctx context.Context, args map[string]any, _ RunOptions) Result {
-	if value, ok := args["session_id"]; !ok || value == nil {
-		return errorResult("Error: Invalid arguments for write_stdin: session_id is required")
-	}
-	sessionID, err := intArg(args, "session_id", 0, 1, 0)
-	if err != nil {
-		return errorResult("Error: Invalid arguments for write_stdin: " + err.Error())
+	// A missing, non-integer, or < 1 session_id all mean the same thing: the model
+	// has no live session to write to. Route them to the SAME recovery guidance the
+	// no-live-session case uses (UnknownExecSessionError) instead of a terse
+	// "session_id must be at least 1". The terse error gives no way forward and, by
+	// naming the minimum, nudges the model to try 1, 2, 3... — the exact id-probing
+	// #749 works to suppress. The recovery message instead tells it to start a
+	// session or edit files directly, and because that message is id-invariant, the
+	// missing/zero/non-integer entry points and id-probing collapse to ONE
+	// repeated-failure signature, so any mix of them accumulates toward the halt
+	// rather than resetting the streak on each class of mistake.
+	value, present := args["session_id"]
+	sessionID, sessionErr := intArg(args, "session_id", 0, 1, 0)
+	if !present || value == nil || sessionErr != nil {
+		return errorResult(UnknownExecSessionError(sessionID))
 	}
 	chars, err := stringArgWithEmpty(args, "chars", "", false, true)
 	if err != nil {
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -923,19 +923,34 @@ func TestWriteStdinSchemaPinsSessionIDMinimum(t *testing.T) {
 	}
 }
 
+// A missing, zero, negative, or non-integer session_id all mean the model has no
+// live session, so write_stdin returns the SAME recovery guidance the no-live-
+// session case uses (start a session with exec_command, or edit files directly),
+// not a terse "session_id must be at least 1" that gives no way forward and names
+// a minimum that nudges the model to probe ids 1, 2, 3... Sharing one id-invariant
+// message also keeps a single repeated-failure signature, so any mix of these
+// mistakes accumulates toward the halt instead of resetting the streak (#749).
 func TestWriteStdinRequiresPositiveSessionID(t *testing.T) {
 	tool := NewWriteStdinTool(newExecSessionManager())
-	for _, args := range []map[string]any{
-		{},
-		{"session_id": 0},
+	want := UnknownExecSessionError(0)
+	for name, args := range map[string]map[string]any{
+		"missing":     {},
+		"zero":        {"session_id": 0},
+		"negative":    {"session_id": -3},
+		"non-integer": {"session_id": "abc"},
 	} {
-		result := tool.Run(context.Background(), args)
-		if result.Status != StatusError {
-			t.Fatalf("Run(%#v) status = %s, want error", args, result.Status)
-		}
-		if !strings.Contains(result.Output, "Invalid arguments for write_stdin") {
-			t.Fatalf("Run(%#v) output = %q, want invalid arguments", args, result.Output)
-		}
+		t.Run(name, func(t *testing.T) {
+			result := tool.Run(context.Background(), args)
+			if result.Status != StatusError {
+				t.Fatalf("status = %s, want error", result.Status)
+			}
+			if result.Output != want {
+				t.Fatalf("output = %q,\n want the recovery message %q", result.Output, want)
+			}
+			if strings.Contains(result.Output, "must be at least 1") {
+				t.Fatalf("still returns the terse minimum error: %q", result.Output)
+			}
+		})
 	}
 }
 
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -935,6 +935,7 @@ func TestWriteStdinRequiresPositiveSessionID(t *testing.T) {
 	want := UnknownExecSessionError(0)
 	for name, args := range map[string]map[string]any{
 		"missing":     {},
+		"nil":         {"session_id": nil},
 		"zero":        {"session_id": 0},
 		"negative":    {"session_id": -3},
 		"non-integer": {"session_id": "abc"},


gnanam1990 (dismissed):
**APPROVE.** The reasoning holds up under measurement — I verified the signature-collapse claim empirically rather than taking it on argument, and it's exactly right. Two small notes below, neither blocking.

## The core claim, verified

The interesting part of this PR isn't the message change, it's the assertion that unifying the message collapses several distinct failure classes into one `errorSignature` so the #749 halt can actually accumulate. I drove all the entry points through the real tool and hashed the outputs the way `internal/agent.errorSignature` does (lowercase → collapse whitespace → truncate 80):

```
probe id 99   tail=". (no live session 99)"
string "5"    tail="s. (no live session 5)"
missing       tail="s. (no live session 0)"
zero          tail="s. (no live session 0)"
negative      tail="s. (no live session 0)"
non-integer   tail="s. (no live session 0)"
probe id 1    tail="s. (no live session 1)"
probe id 2    tail="s. (no live session 2)"

distinct signatures across ALL of the above: 1
  sig="error: write_stdin needs a session_id returned by a still-running exec_command; "
```

One signature across eight inputs, including the id-probing sequence. So a model wandering `missing → 0 → 1 → 2` now accumulates a single streak toward `toolFailureStopAt` instead of resetting it four times. That's the behaviour the PR describes and it's genuinely what happens.

Also confirmed the exact-equality assertion in the new test is sound rather than incidental: `intArg` returns `0` on **every** error path (`args.go:101,112,116,119,123,126`) — type error, min violation, max violation alike — so `UnknownExecSessionError(sessionID)` is always the id-0 form when `sessionErr != nil`. The test isn't relying on an accident.

## The case I went looking for, and it's fine

The change routes "non-integer `session_id`" into the no-live-session message, which would be actively harmful in one specific scenario: a model sending a **valid, live** id as a string (`{"session_id": "5"}` — a common enough LLM slip). Telling it "no live session" when session 5 is running would send it off to start a redundant session.

That doesn't happen: `intArg` explicitly coerces numeric strings (`args.go:104-108`, comment "Some models send numbers as strings (\"5\"); accept whole numbers"), so `"5"` resolves to `5` and proceeds to the normal live-session lookup. Only genuinely non-numeric strings reach the new branch, and for those "start a session or edit files directly" is the correct advice. Worth stating explicitly since it's the one way this could have regressed.

## Minor

**1. "that message is id-invariant" is imprecise — it's the *signature* that is.** The message does embed the id, at the very end: `... (no live session %d)`, around char 210 of ~230. It only becomes invariant after `errorSignature` truncates at 80. Both the code comment (`exec_command.go:778-786`) and the PR body say "message", which could mislead someone who later shortens the string or moves the id earlier — that would silently break the collapse this PR depends on.

The good news is it can't rot silently: `TestUnknownExecSessionErrorSignatureIsIDInvariant` (`internal/agent/guardrails_test.go:177`) pins it by comparing ids 1 and 999999. Worth naming that test in the comment, so the next person editing the string knows what guards the property and why the id must stay past the 80-char cut.

**2. Type information is genuinely lost for a non-numeric id, and that's a real trade-off worth acknowledging.** `{"session_id":"abc"}` now returns "no live session 0" with nothing indicating a string was passed where an integer was required. I'd have pushed on this harder except two things blunt it: the coercion above means only truly non-numeric values land here, and `toolFailureHint` injects the full schema JSON after a repeated-failure streak, so a model that keeps getting it wrong does eventually see `"type": "integer", "minimum": 1`. Net acceptable — but it is a deliberate loss, not a free win, and a sentence in the comment saying so would age better than the current framing that treats all four cases as simply equivalent.

## Nice details

- Keeping `minimum: 1` in the schema (`TestWriteStdinSchemaPinsSessionIDMinimum` unchanged) while removing it from the runtime error is the right split: the constraint is stated up front where it informs, not in the failure path where it reads as an invitation to try 1, then 2.
- The new test asserts **exact** output equality rather than a substring, and adds the negative assertion that `must be at least 1` is absent. That pins the specific regression instead of just "some error happened". I mutation-tested it — restoring the old terse branches fails all four subtests.

## Verification

`gofmt` and `go vet ./internal/tools/` clean. Full `internal/tools` suite green apart from three failures — `TestScopedToolsAllowReadOnlyRootsWithoutWrite`, `TestRegistryAppliesSandboxBeforeToolExecution`, `TestRegistrySandboxGatesPathAliasKeys` — which fail identically on the merge base (`fbf8598`), so they're environmental here and unrelated to this diff. The #749 guardrails (`TestUnknownExecSessionErrorSignatureIsIDInvariant`, `TestUnknownExecSessionProbingTripsFailureHalt`) pass unchanged, and every existing `TestWriteStdin*` including the live-session paths still passes.

Note the branch is based on an older main, so a two-dot `git diff` against `fbf8598` lists ~26 files; the PR itself is the 2 files GitHub reports. Just something to be aware of when it merges.

gnanam1990:
Agreeing with @coderabbitai's `nil` case, and it's a touch more load-bearing than a plain coverage gap — worth adding before merge.

I checked what actually guards each input:

| input | caught by |
|---|---|
| missing | `!present` |
| `0` / negative / non-integer | `sessionErr != nil` |
| **`nil`** | **`value == nil` — and nothing else** |

`intArg(map[string]any{"session_id": nil}, "session_id", 0, 1, 0)` returns `(0, nil)` — no error — because of the early `!ok || value == nil` return at `args.go:86`. So `sessionErr` does **not** cover the nil case, and `value == nil` is the only thing standing between it and a fall-through to the live-session lookup for id 0.

Behaviour today is correct — I ran it, `{"session_id": nil}` returns exactly `UnknownExecSessionError(0)`. But it's the one branch in the new condition with no test, and it's precisely the branch that isn't redundant with another check. Adding `"nil": {"session_id": nil}` to the table pins the only non-redundant guard.

My approval stands — this is a one-line test addition, not a rethink.

Vasanthdev2004:
Added in `772419f8`. The explicit `session_id: nil` case is now in the table and asserts the same `UnknownExecSessionError(0)` recovery message as the missing/zero/negative/non-integer cases.

gnanam1990 (approved):
**APPROVE** at `772419f`. The `nil` case is covered and everything is green.

**First, a correction to my own earlier comment — I overstated the nil case and want to set the record straight before it misleads anyone.**

I said `value == nil` was "the only thing standing between it and a fall-through to the live-session lookup for id 0." That's wrong. I mutation-tested it: deleting `value == nil` and re-running the new `nil` subtest — it **still passes**. The reason is the fall-through lands on `tool.manager.get(0)` at `exec_command.go:805`, which fails and returns `UnknownExecSessionError(0)` — the *identical* string. For a bare `{"session_id": nil}`, the guard is a fast path, not a unique protection, and I should have traced that before asserting otherwise.

So, strictly: **the new `nil` subtest passes with or without the guard it was added to cover.** That doesn't make it a bad test — it pins user-visible behaviour, which is what a table like this is for — but it's worth knowing it isn't pinning the branch.

## The property that *is* load-bearing, and is still untested

The guard does earn its place, just not the way I described. It establishes **precedence**: an unusable `session_id` must win over sibling-argument errors. Measured, with and without it:

| input | with the guard (shipped) | without it |
|---|---|---|
| `{session_id: nil}` | session error | session error |
| `{session_id: nil, chars: 123}` | session error | `chars must be a string` |
| `{session_id: nil, yield_time_ms: "xyz"}` | session error | `yield_time_ms must be …` |

That second and third row are the point of the PR. Without the guard, a model that sends a nil id *and* fumbles another argument gets a **different message with a different `errorSignature`** — precisely the streak-reset this change exists to eliminate. The guard keeps the id-invariant message in front of `chars` / `yield_time_ms` / `max_output_tokens` validation.

Nothing tests that ordering. A single case would pin the guard properly, where the plain `nil` case cannot:

```go
"nil id beats sibling arg errors": {"session_id": nil, "chars": 123},
```

It fails the moment the guard is removed. Worth adding while this is open — it's the same one-line shape as the case just added, and it covers the thing that actually breaks. Not blocking: the behaviour is correct today, and this is a test-strength point, not a defect.

## Verification at `772419f`

`gofmt` and `go vet ./internal/tools/` clean. All `TestWriteStdin*` pass, including the five subtests (`missing`, `nil`, `zero`, `negative`, `non-integer`) and the live-session paths. The #749 guardrails in `internal/agent` pass unchanged. The three `internal/tools` failures I noted last round remain identical on the merge base — environmental, unrelated.

Everything from my previous review stands: the signature collapse is real (I measured eight inputs down to one signature), `intArg`'s numeric-string coercion means a live id sent as `"5"` still resolves correctly, and the schema/runtime split is the right call.

anandh8x (approved):
Approved at 772419f8. The invalid-session paths now consistently return actionable recovery guidance, discourage session-id probing, and preserve one failure signature across missing, nil, zero, negative, malformed, and unknown ids. Numeric-string live ids still follow the normal lookup path, valid sessions are unchanged, and the added nil coverage addresses the outstanding review request. All current CI checks are green.

PR #775, opened 2026-07-20
fix(tui): model rows labelled by id when the description is prose; keep the sidebar under the / palette

gnanam1990:
## The bug

The model picker labels each row with the provider's `Description`, falling back to the model ID only when the description is "generic". In a live picker against an aggregator provider, that produced this:

```
❯ Fast DeepSeek model for efficient chat, coding help, and agent loops
  Flagship DeepSeek model for coding, reasoning, and agentic work
  Open Gemma instruction model for efficient chat and self-hosted deployments
  Flagship GLM model for hybrid reasoning, coding, and agentic engineering
  Open flagship GLM for long-horizon coding agents and million-token context work
  Open-weight GPT model for self-hosted reasoning and instruction-following workloads
  Open-weight GPT model for self-hosted reasoning and instruction-following workloads
  Kimi multimodal agent model for visual understanding, coding, and planning
  Kimi multimodal agent model for visual understanding, coding, and planning
  Coding-focused Kimi model, stronger on long-horizon repo work with less overthinking
  deepseek-v4-flash · 1M ctx · tools · reasoning
```

Two pairs of rows are **byte-identical** while being different models. The distinguishing ID appears only in the detail line, for whichever row happens to be highlighted — so the list cannot be used to choose between them. You have to arrow onto each row in turn and read the line underneath.

## Root cause

`displayLabel` (`internal/tui/provider_wizard.go`) prefers the description over the ID:

```go
description := strings.TrimSpace(model.Description)
if description != "" && !providerWizardGenericModelDescription(description) {
    return description
}
return model.ID
```

That is the right call when `Description` holds a short display **name** — `"Grok 4.3"` for `x-ai/grok-4.3` — which is what the field is for, and why the row prefers it over an ugly raw ID. `TestProviderWizardModelStepUsesFriendlyNamesAndStaysCompact` pins exactly that, including `assertNotContains(view, "❯ x-ai/grok-4.3")`.

Providers do not all honour it. Aggregators commonly return a prose blurb instead, which fails as a label twice over: it overflows the row, and because such blurbs are written per model **family** rather than per model, several different models collapse to identical rows.

`providerWizardGenericModelDescription` cannot catch these. It rejects descriptions that are *vague* — `""`, `"catalog default"`, `"live model"`, anything ending in `" model"` — whereas these are specific, just shared and sentence-shaped. Extending that deny-list cannot work either: none of these strings live in the repo, they arrive from the provider at discovery time (`internal/providermodeldiscovery/discovery.go:336`).

## The fix

Gate on **shape** rather than wording. A display name is short and few-worded; anything longer is prose, and the ID — unique by construction — makes the better label:

```go
const (
    providerWizardMaxModelNameChars = 28
    providerWizardMaxModelNameWords = 4
)
```

Friendly names are still preferred wherever providers supply them, so behaviour is unchanged for well-behaved catalogs. The description is not lost — it moves to the detail line via the existing `secondaryText`.

**This fixes two surfaces.** `internal/tui/onboarding.go` renders its model rows through the same `displayLabel`/`secondaryText` pair (`onboarding.go:1576`, `:1582`), so first-run setup had the identical defect.

Before → after, on the models from the report:

| row label (before) | row label (after) |
|---|---|
| `Open-weight GPT model for self-hosted reasoning…` | `gpt-oss-120b` |
| `Open-weight GPT model for self-hosted reasoning…` | `gpt-oss-20b` |
| `Kimi multimodal agent model for visual understanding…` | `kimi-k2-thinking` |
| `Kimi multimodal agent model for visual understanding…` | `kimi-k2-instruct` |
| `Grok 4.3` | `Grok 4.3` (unchanged) |

## Verification

- `TestProviderWizardModelRowsStayDistinctWithProseDescriptions` added, using the exact models from the report. Mutation-tested: with the fix reverted it fails with `model "gpt-oss-20b" is not identifiable in the list`.
- Both tests that pin the existing friendly-name behaviour still pass unchanged — `TestProviderWizardModelStepUsesFriendlyNamesAndStaysCompact` and `TestProviderWizardModelSearchFiltersAndAppliesRawModelID`.
- Onboarding verified to share the fix: prose model → row `gpt-oss-20b`, named model → row `Grok 4.3`.
- `gofmt`, `go vet ./internal/tui/` clean. Full `internal/tui` suite has the same single failure as `main` (`TestHandleAddDirCommand`, a pre-existing sandbox/TMPDIR artifact) — no new failures.

## Note for reviewers

`TestSuggestionOverlayRenders` asserts `!strings.Contains(plain, "/model")` against the whole rendered view, which includes the git branch name in the title bar. It therefore fails on any branch whose name contains `/model` — this branch was originally `fix/model-picker-label` and had to be renamed to get a green run. Unrelated to this change; worth tightening separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved labeling in the provider setup wizard to make model selections clearer and remain distinct, even with long or generic prose descriptions.
  * When model IDs are missing, row labels no longer render blank; they now use the most appropriate fallback.
  * Fixed sidebar behavior so it remains stable when the command palette/suggestions overlay is open, without layout collisions.
* **Tests**
  * Added regression coverage for distinct model row rendering and sidebar stability during the command palette overlay.

---

## Second fix: the `/` command palette collapsed the sidebar

Same package, same class of defect — a display rule keyed off the wrong signal — so it rides along here.

`sidebarAvailable` grouped `suggestionsActive()` with the full-screen overlays (setup, wizards, pickers), whose stated rationale is that they *"take over the chat column and render at full width"*. That is not true of the `/` palette: `suggestionOverlay` draws it through `centerRenderedBlock` at `minInt(width, suggestionPaletteMaxWidth)` — a centred box capped at 76 cells, floating over the chat column.

Treating it as full-screen collapsed the whole layout on every `/`: the sidebar disappeared, the plan panel dropped out of it and re-rendered inline above the composer, and the transcript re-wrapped to full width. A full reflow for a popup, and worst exactly when it is most visible — mid-run with a live plan on screen.

**It cannot overlap the sidebar either.** `twoColumnTranscriptView` renders the overlay with `width = chatColumnWidth`, so the palette caps at `min(chatColumn, 76)` and is centred *inside* the chat column. Worth stating explicitly because the 76-cell cap alone does not guarantee it — at 100 cells total the chat column is only 67, so the cap is not what bounds it; the call site is. The new test asserts on the rendered line widths rather than the constant, so it stays honest if either changes.

**Mouse hit-testing is unaffected.** `sidebarLineAtMouse` carries its own `suggestionsActive()` guard (`mouse.go:234`), so clicks still reach the palette and not the sidebar rows beneath it. That independent guard is likely why the suppression looked safe to add originally.

`TestSidebarSurvivesCommandPalette` pins all of it: the sidebar survives the palette, the plan stays in the sidebar rather than falling back to the pinned panel, no rendered palette line exceeds the chat column, the sidebar takes no mouse hits while it is open — and a real `picker` still collapses the sidebar. Mutation-verified: restoring `suggestionsActive()` to the condition fails it with *"command palette must not collapse the sidebar"*.

### Verification (both fixes)

`gofmt` and `go vet ./internal/tui/` clean. Full `internal/tui` suite has the same single failure as `main` (`TestHandleAddDirCommand`, a pre-existing sandbox/TMPDIR artifact) — no new failures from either change.

### DIFF
--- a/internal/tui/provider_wizard.go
+++ b/internal/tui/provider_wizard.go
@@ -1959,12 +1959,49 @@ func (wizard *providerWizardState) filteredModels() []providerWizardModel {
 	return models
 }
 
-func (model providerWizardModel) displayLabel() string {
-	description := strings.TrimSpace(model.Description)
-	if description != "" && !providerWizardGenericModelDescription(description) {
-		return description
+// A model Description is only a good row label when it reads like a display
+// NAME — "Grok 4.3" for x-ai/grok-4.3 — which is what the field is for, and why
+// the row prefers it over an ugly raw ID.
+//
+// Providers do not all honour that. Aggregators commonly return a prose blurb
+// instead ("Fast DeepSeek model for efficient chat, coding help, and agent
+// loops"), which fails as a label twice over: it overflows the row, and because
+// such blurbs are written per model FAMILY rather than per model, several
+// DIFFERENT models collapse to byte-identical rows. A picker then offers two
+// visually identical entries with the distinguishing ID visible only in the
+// detail line of whichever one is highlighted.
+//
+// providerWizardGenericModelDescription cannot catch these: it rejects
+// descriptions that are vague ("live model", "catalog default"), whereas these
+// are specific — just shared and sentence-shaped. So the test is shape, not
+// wording: a name is short and few-worded; anything longer is prose, and the ID
+// (unique by construction) makes the better label.
+const (
+	providerWizardMaxModelNameChars = 28
+	providerWizardMaxModelNameWords = 4
+)
+
+func providerWizardDescriptionIsDisplayName(description string) bool {
+	trimmed := strings.TrimSpace(description)
+	if trimmed == "" || providerWizardGenericModelDescription(trimmed) {
+		return false
 	}
-	return model.ID
+	if len(trimmed) > providerWizardMaxModelNameChars {
+		return false
+	}
+	return len(strings.Fields(trimmed)) <= providerWizardMaxModelNameWords
+}
+
+func (model providerWizardModel) displayLabel() string {
+	if providerWizardDescriptionIsDisplayName(model.Description) {
+		return strings.TrimSpace(model.Description)
+	}
+	if id := strings.TrimSpace(model.ID); id != "" {
+		return id
+	}
+	// ID-less sentinel rows built in currentModel ("model name required",
+	// "no matching models") would otherwise render blank.
+	return strings.TrimSpace(model.Description)
 }
 
 func (model providerWizardModel) secondaryText() string {
--- a/internal/tui/provider_wizard_test.go
+++ b/internal/tui/provider_wizard_test.go
@@ -1787,3 +1787,37 @@ func providerWizardIDs(descriptors []providercatalog.Descriptor) []string {
 	}
 	return ids
 }
+
+// A provider that returns prose blurbs instead of display names must not
+// collapse distinct models into identical rows. Reproduces the reported picker
+// where two "Open-weight GPT model..." and two "Kimi multimodal agent model..."
+// rows were indistinguishable, with the ID visible only for the highlighted one.
+func TestProviderWizardModelRowsStayDistinctWithProseDescriptions(t *testing.T) {
+	wizard := &providerWizardState{
+		step:        providerWizardStepModel,
+		modelSource: "models.dev",
+		models: []providerWizardModel{
+			{ID: "gpt-oss-120b", Description: "Open-weight GPT model for self-hosted reasoning and instruction-following workloads"},
+			{ID: "gpt-oss-20b", Description: "Open-weight GPT model for self-hosted reasoning and instruction-following workloads"},
+			{ID: "kimi-k2-thinking", Description: "Kimi multimodal agent model for visual understanding, coding, and planning"},
+			{ID: "kimi-k2-instruct", Description: "Kimi multimodal agent model for visual understanding, coding, and planning"},
+			{ID: "x-ai/grok-4.3", Description: "Grok 4.3"},
+		},
+	}
+	view := plainRender(t, strings.Join(wizard.renderModelStep(96), "\n"))
+
+	// every prose-described model is identified by its unique ID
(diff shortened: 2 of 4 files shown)


anandh8x (changes requested):
.

anandh8x (approved):
Approved on the current head. The model-label fallback preserves concise friendly names while keeping prose-described models distinguishable, and the command palette remains constrained to the chat column without reflowing the sidebar. Focused TUI tests and CI pass.

Vasanthdev2004 (approved):
Approving. Both changes are sound and I verified the load-bearing parts rather than taking the comments' word.

**Label fix.** The prose-collapse is real (byte-identical rows for different models), and the shape-based heuristic is the right instinct: a description is used as the label only when it reads like a name (non-generic, `<= 28` chars, `<= 4` words), otherwise the unique ID wins. I confirmed the pinned behavior survives, `TestProviderWizardModelStepUsesFriendlyNamesAndStaysCompact` still passes, so "Grok 4.3" is preferred over the raw id, and the new prose-blurb test passes.

One tradeoff worth naming (not a blocker): a genuine model name longer than 28 chars or over 4 words (say "Llama 3.3 70B Instruct Turbo") now shows the ID instead of the name. That is the safe direction since the ID is unique, and chasing a perfect name-vs-prose classifier is not worth it, but it is a behavior change for those rows. The one residual the shape check cannot catch is a SHORT shared blurb (a provider returning "Fast chat model" for several models would still collapse); fine to leave, since the reported and common case is the long blurbs.

**Sidebar under the palette.** I traced this instead of trusting the comment: `twoColumnTranscriptView` sets `width := chatW` and builds the suggestion overlay at the chat-column width, then joins the sidebar as a separate column, so the palette is structurally bounded to the chat column and cannot contend with the sidebar even though `suggestionPaletteMaxWidth` (76) exceeds the chat width at the medium tier. Mouse hit-testing stays correct because `sidebarLineAtMouse` keeps its own `suggestionsActive()` guard. Keeping the plan and sidebar on screen while typing `/` is the right call, and it is at its most valuable mid-run with a live plan.

Tests green locally (friendly-name, prose-distinct, and sidebar suites); CI green.

PR #776, opened 2026-07-20
fix(swarm): close lifecycle admission on shutdown

### LINKED ISSUE #771: fix(swarm): shutdown does not close lifecycle admission
## Summary

`Swarm.Close` cancels the base context but does not record a closed state or wait for member lifecycle watchers. Calls made after shutdown begins can still register tasks and launch work with an already-canceled context, while existing watchers may continue retrying or draining queued members.

This tracks **AUD-010** from the July 18 codebase audit.

## Affected code

- `internal/swarm/team.go:139-152`
- `internal/swarm/lifecycle.go:14-27`
- `internal/swarm/lifecycle.go:62-167`

## Current behavior

Shutdown cancellation is not coupled to lifecycle admission or watcher completion:

- lifecycle operations can still be admitted after `Close`;
- new work can launch with a canceled context;
- existing watchers can retry or drain queued members after shutdown begins; and
- `Close` can return without waiting for lifecycle goroutines to finish.

## Expected behavior

After shutdown begins, the swarm should reject new lifecycle work and prevent retries or queue draining. `Close` should not return until owned lifecycle watchers have exited.

## Suggested implementation

- Add an atomic or mutex-protected closed state.
- Mark the swarm closed before canceling its base context.
- Reject lifecycle admission after close.
- Track lifecycle watchers with a `sync.WaitGroup` or equivalent ownership mechanism.
- Stop retry and queue-drain paths once cancellation or closed state is observed.
- Wait for watcher completion before `Close` returns.

## Suggested tests

- Lifecycle registration and launch attempts after `Close` are rejected.
- Shutdown prevents queued members from being started.
- Retry paths stop once shutdown begins.
- `Close` waits for active lifecycle watchers to exit.
- Concurrent `Close` and lifecycle admission cannot launch post-close work.

PierrunoYT:
## Summary
- close swarm lifecycle admission atomically before cancellation
- reject spawn, handoff, adoption, retry, and queue-drain work after shutdown begins
- fail queued tasks and wait for all owned member watchers before `Close` returns
- make concurrent swarm and scheduler close calls completion barriers
- add lifecycle shutdown, retry, queue, and concurrent-close regression coverage

## Validation
- `go test ./internal/swarm -count=1`
- `go test -race ./internal/swarm -count=1` (WSL/Linux)
- `go vet ./internal/swarm`
- `go test ./...`
- `git diff --check`

Closes #771

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle-admission-aware scheduling so member work starts only while the swarm is open, including during handoff and orphan adoption.
  * Introduced exported `ErrSwarmClosed` to fail work that would otherwise remain queued.

* **Bug Fixes**
  * Made shutdown idempotent and coordinated: repeated `Close()` calls now wait for completion.
  * `Close()` prevents late queued members from starting, cancels launches blocked in progress, and fails dequeued/late-created queued specs appropriately.
  * Updated member supervision retry to relaunch in-place when allowed.

* **Tests**
  * Added coverage for admission rejection after close, close/barrier timing, queued-spec draining, scheduler creation after close, and race interleavings.


jatmn (changes requested):

## Findings

- [P1] Do not hold lifecycle admission while starting a member
  `internal/swarm/lifecycle.go:59`
  `Spawn` (and the retry and queue-drain paths) keeps `lifecycleMu.RLock()` while invoking the injected `MemberLauncher.Launch`. `Close` needs the write lock before it calls `s.cancel()`, so a launcher that blocks during startup waiting for its supplied context can never be cancelled: launch holds the read lock, while `Close` waits for that lock before issuing the cancellation. This leaves both the spawn and every close caller hung. Release the admission lock before calling external launcher code, while retaining an ordering mechanism that prevents a launch from being admitted once shutdown starts; add a regression launcher whose `Launch` waits on `ctx.Done()`.

- [P3] Document the admission helper's lock-transfer contract
  `internal/swarm/lifecycle.go:35`
  `beginLifecycleAdmission` returns with `lifecycleMu` read-locked on success but unlocks it before returning `ErrSwarmClosed`. Callers must know that asymmetric contract to avoid leaked read locks or an extra unlock, yet the helper has no comment describing it. This remains the open Copilot/CodeRabbit review request; document the ownership rule (or return an explicit unlock function).

- [P3] Explain the post-close scheduler handoff
  `internal/swarm/team.go:177`
  `Scheduler` snapshots `closed` under `lifecycleMu`, installs/reads `s.scheduler`, then closes the scheduler after releasing both locks. The correctness relies on `Close` either seeing and closing that stored scheduler or `Scheduler` seeing the closed snapshot and closing it itself. This is the unresolved CodeRabbit request; add a concise comment preserving that synchronization rationale.

PierrunoYT:
Pushed `3e18b0b`. This one did not need a rebase — `main` hasn't touched `internal/swarm` since the branch was cut, so `mergeable: false` must have been a stale check; it's currently reporting `MERGEABLE` against the pushed commit.

**[P1] Do not hold lifecycle admission while starting a member** — confirmed and fixed. `beginLifecycleAdmission` used to return with `lifecycleMu.RLock()` held for the caller's *entire* duration, including the call into the injected `MemberLauncher.Launch`. `Close` needs the write lock before it can call `s.cancel()`. So a launcher that blocks in `Launch` waiting on its context — the ordinary shape for a launcher waiting on a slot, a daemon connection, or a sandbox handshake — deadlocked every `Close` caller: the launch held the read lock while `Close` waited for the write lock to reach the very cancellation the launch needed in order to unblock. Exactly the hang you described.

Fix: `beginLifecycleAdmission` is now a counted **ticket**, not a held lock. It takes `lifecycleMu` only across the `closed` check plus a `lifecycleWork.Add(1)`, then returns a release func (`lifecycleWork.Done`) the caller defers. `Close` flips `closed` and releases `lifecycleMu` **before** calling `cancel`, then waits `lifecycleWork` (admitted-but-unfinished work) followed by `watchers` (goroutines that admitted work may still be starting — a spawn in flight can add a watcher after `lifecycleWork` alone would have returned). Every call site — `Spawn`, `Handoff`, `AdoptOrphans`, the retry path in `watch`, and `afterExit` — moved to the ticket pattern.

**Regression tests, verified to have teeth**

- `TestCloseCancelsLauncherBlockedInLaunch`: a launcher parks in `Launch` on `<-ctx.Done()`; a `Spawn` and a concurrent `Close` must both complete once cancellation reaches it. I confirmed this actually catches the bug by temporarily restoring the old held-lock behavior (RLock across the caller, cancel inside the write-locked section) — the test then hangs and fails on its own timeout, so it isn't a test that would pass regardless.
- `TestCloseWaitsForAdmittedSpawn`: the converse guarantee — `Close` must not return while an admitted spawn is still in flight, so the barrier property added by #776 doesn't regress while fixing the deadlock.

**[P3] Document the admission helper's lock-transfer contract** — the `bool`-returning asymmetric contract Copilot/CodeRabbit flagged no longer exists: `beginLifecycleAdmission` now returns `(func(), error)`, so the release obligation is explicit in the *type signature* rather than something a caller has to remember from a doc comment. The doc comment on the function additionally explains *why* the lock isn't held across `Launch` (the deadlock above), not just what to do with the return value.

**[P3] Explain the post-close scheduler handoff** — added a comment on `Scheduler()` spelling out the lock-ordering argument: `Close` takes `lifecycleMu.Lock()` (excluding `Scheduler`'s `RLock`), sets `closed`, snapshots whatever `s.scheduler` currently is, and only *then* releases the lock and calls `Close` on what it captured. So a `Scheduler()` call either runs entirely before `Close`'s write lock (sees `closed == false`, and the scheduler it creates/returns is exactly what `Close`'s later snapshot will capture) or entirely after (sees `closed == true` via the RLock/Lock memory-model ordering, and closes the scheduler itself, since `Close` already took its own snapshot and won't look again). Exactly one of the two call sites closes whatever scheduler ends up installed.

**Validation** — I don't have a C toolchain on this Windows host, but WSL has one, so I did not skip `-race` this time:

```
wsl -d Ubuntu -- bash -lc 'cd /mnt/d/Github/zero && go test -race ./internal/swarm/... -count=1'
→ ok  	github.com/Gitlawb/zero/internal/swarm	1.968s
```

Plus `go vet ./...` under WSL/Linux, and on the Windows host: `go build ./...`, `go vet ./...`, `go test ./... -count=1` all pass; `GOOS=linux|darwin` builds and vets of `internal/swarm` clean; `gofmt` clean on every changed file.

@jatmn — ready for another look. I can't use the reviewer-request button on this repo, hence the mention.

@coderabbitai review

jatmn (changes requested):

## Findings

- [P1] Fail queues after admitted work has stopped adding to them
  `internal/swarm/team.go:179`
  `Close` clears each queue before it waits for lifecycle operations that passed `beginLifecycleAdmission` just before `closed` flipped. A `Spawn`, `Handoff`, or `AdoptOrphans` can be paused after receiving that ticket and resume after this sweep, appending its already-registered task to a full team queue. The canceled watcher then reaches `afterExit`, sees the closed swarm, and only releases its slot, so the late task is neither launched nor failed and remains pending after `Close` returns. This can strand handoff/adopted work and make `CollectWait` block indefinitely; synchronize queue admission with the sweep or fail queues after admitted dispatches have quiesced.

- [P1] Do not launch a dequeued member once shutdown has started
  `internal/swarm/lifecycle.go:145`
  An `afterExit` that obtained its admission ticket before `Close` can dequeue a queued spec with `t.onExit`, then shutdown can set `closed` before `launchAdmitted` runs. That dequeued spec is no longer visible to `clearQueue`, and `afterExitAdmitted` has no post-dequeue shutdown check, so it still calls the injected launcher after shutdown began. A launcher that performs work before respecting its canceled context can therefore create a new worker after `Close`; make the dequeue/close transition atomic or re-check and terminalize the dequeued task before launch.

PierrunoYT:
Pushed `915d33d`, addressing both P1 findings from the latest round.

**[P1] Fail queues after admitted work has stopped adding to them** (`internal/swarm/team.go:179`)

`Close` swept (`clearQueue`) every team's queue *before* waiting for admitted-but-in-flight lifecycle operations (`lifecycleWork.Wait()`) to finish. A `Spawn`/`Handoff`/`AdoptOrphans` call that got its ticket right before `closed` flipped could still be paused before it reached `dispatchAdmitted` (the queue append) — resuming after the sweep already ran meant its spec sat in the queue forever: never launched (shutdown had begun), never failed (the one-time sweep had already passed).

Fixed by reordering `Close`: `lifecycleWork.Wait()` now runs *before* the queue sweep. Every append a still-admitted operation makes happens strictly before it releases its ticket, so once every ticket is back, no queue can gain a new entry — the sweep can no longer race a late append.

Added `TestCloseFailsLateQueuedSpec`, which holds a ticket open past the point where `Close` has flipped `closed`, appends a spec to a full team's queue while still holding it, and only then releases — proving the late append still gets failed with `ErrSwarmClosed` rather than stranded.

**[P1] Do not launch a dequeued member once shutdown has started** (`internal/swarm/lifecycle.go:145`)

`afterExitAdmitted` dequeued a spec (`t.onExit`) and launched it unconditionally once its own ticket was granted — but `Close` can set `closed` in the gap between that dequeue (which makes the spec invisible to `clearQueue`) and the launch. Since the reordering above makes `Close` wait for this ticket anyway, the launch was never *lost track of*, but it violated the stated shutdown contract: a launcher that performs work before honoring its cancelled context could still start a brand-new worker after shutdown had already begun.

Fixed by re-checking `closed` right after the dequeue: if shutdown has begun, the slot is released and the task fails with `ErrSwarmClosed` instead of launching.

Added `TestAfterExitAdmittedFailsDequeuedSpecAfterClose`, driving the exact interleaving directly (ticket already granted, `closed` flips, then the post-dequeue check runs).

Both new tests are verified to have teeth: I temporarily reverted each fix in isolation and reran them — both fail against the pre-fix code (confirmed, not assumed), then restored the fixes and confirmed both pass.

**Validation:**
- `go test ./internal/swarm/... -run 'TestCloseFailsLateQueuedSpec|TestAfterExitAdmittedFailsDequeuedSpecAfterClose|TestClose|TestLifecycle' -count=30`: pass.
- `go test -race ./internal/swarm/... -count=20` via WSL: pass, plus `go build`/`go vet ./...` under WSL — clean.

jatmn (changes requested):

## Findings

- [P1] Snapshot teams after admitted work stops creating them
  `internal/swarm/team.go:164`
  `Close` copies `s.teams` before waiting for already-admitted lifecycle calls. Two calls that received tickets before shutdown can both be delayed until after that snapshot; the first creates a previously unseen team and occupies its only slot, while the second queues in that new team. Once their tickets return, the sweep at line 187 only clears the stale (empty) snapshot, so the queued task remains pending after `Close` returns—its canceled watcher only releases the slot and cannot drain the queue. Take the team snapshot after lifecycle admission has quiesced, or otherwise sweep teams created by those admitted calls.

PierrunoYT:
Final follow-up on current head `2ef453f`:

- moved the shutdown team snapshot after lifecycle admission quiesces
- added deterministic coverage for late-created teams with queued work
- verified the earlier admission, scheduler handoff, sweep, and dequeue findings remain addressed
- all review threads are resolved
- race tests, full validation, vulnerability checks, and all GitHub checks pass

The PR is mergeable. PierrunoYT requests a fresh review to supersede the previous `CHANGES_REQUESTED` decision.

PierrunoYT:
Pushed `2784b84` after normally merging current `upstream/main` (`81a5e6c`).

### Final shutdown/admission hardening

- Re-checked `closed` at the final internal boundary before **every** external member launch, including retries. A lifecycle call that obtained a ticket before shutdown but reaches dispatch/retry only after `Close` has won now fails without invoking the launcher on the cancelled swarm context.
- Preserved the counted-ticket design: no lifecycle lock is held across external launcher code, so cancellation can still unblock an in-flight `Launch` and `Close` remains a completion barrier.
- Kept the team snapshot and queue sweep after `lifecycleWork.Wait()`. The snapshot regression now isolates exactly the late-created-team/late-queue interleaving instead of relying on a post-close launcher invocation.
- Corrected the watcher ordering comment: every `watchers.Add` occurs while an admission ticket is held; watcher goroutines may outlive that ticket, so `Close` waits both groups.
- Added coverage for a scheduler first created after `Swarm.Close`, asserting it is actually marked closed (not merely masked by a cancelled context).

### New/updated regression coverage

- `TestAdmittedDispatchDoesNotLaunchAfterClose`
- `TestCloseRacesLifecycleAdmission` (64 concurrent spawns racing shutdown)
- `TestCloseFailsQueuedSpecOnLateCreatedTeam` (deterministic post-quiescence snapshot timing)
- `TestSchedulerCreatedAfterSwarmCloseIsClosed`

### Validation

All passed on Linux/amd64 with the repository-required Go 1.26.5 toolchain:

- focused shutdown/snapshot/scheduler race tests: `-race -count=50`
- full `internal/swarm/...` race suite: `-race -count=10` (run twice, including after the final assertion)
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make lint-static` — `0 issues`
- `make vulncheck` — `No vulnerabilities found`
- `git diff HEAD --check`

### Review disposition

No still-valid request was skipped. The older findings are stale against this head and were reverified as addressed:

- held admission lock across `Launch`: replaced by counted tickets and cancellation regression coverage;
- undocumented lock-transfer contract: no asymmetric lock transfer remains; the helper returns an explicit release function and documents it;
- scheduler close/create handoff: documented and now directly tested;
- queue sweep before admitted work quiesced: sweep remains after `lifecycleWork.Wait()`;
- dequeued launch after shutdown: post-dequeue check remains, with the new final launcher-boundary check closing the corresponding late-dispatch/retry path;
- team snapshot before admitted work created teams: snapshot remains after admission quiesces and has deterministic regression coverage.

The remaining `CHANGES_REQUESTED` status is from the review on `915d33d`; its team-snapshot finding was fixed in `2ef453f` and remains fixed after the upstream merge. @jatmn, please take a fresh look when convenient.

jatmn (changes requested):

## Findings

- [P1] Close the launch commit window, not only the pre-`Launch` read
  `internal/swarm/lifecycle.go:103-110` and `internal/swarm/lifecycle.go:78-96`
  `launchMemberAdmitted` snapshots `closed` under `lifecycleMu.RLock()`, releases the lock, and only then calls `launcher.Launch`. `Close` can set `closed=true` and cancel `baseCtx` in that gap. Production wiring uses `FuncLauncher` (`member.go:64-81`), whose `Launch` always returns `(handle, nil)` immediately and starts `Run` in a goroutine without consulting `ctx.Err()` at entry. When `Launch` succeeds in that race, `launchAdmitted` still `addMember`s, sets `StatusRunning`, and starts a watcher — so the swarm registers and supervises a new member after shutdown has closed admission. That is the real defect: not stranded `pending` tasks (see below), but a post-close **running** member the shutdown contract says should not be committed. Head `2784b84` claims a re-check at the final internal boundary before every external launch; the check is only before `Launch`, not before `addMember`/`SetStatus`. `TestAdmittedDispatchDoesNotLaunchAfterClose` sets `closed` before dispatch; it does not interleave `Close` between the `closed` read and `Launch`. `TestCloseRacesLifecycleAdmission` would still pass if a late launch completes and the task terminalizes normally. Re-check `closed` under `lifecycleMu` immediately after `Launch` returns, before `addMember` and `SetStatus`, and fail the task with `ErrSwarmClosed` if shutdown won (release the slot, do not start a watcher). Alternatively, hold `lifecycleMu` across the non-blocking `Launch` call only — safe for `FuncLauncher` because `Launch` does not block, but must not reintroduce the deadlock fixed in `3e18b0b` for launchers that block in `Launch` waiting on context cancellation.

## Rechecked from prior review

On head `2784b84cae51ac4463c81b69c1ec8914c4fc6741`, the earlier shutdown findings look addressed:

- Lifecycle admission no longer holds `lifecycleMu` across blocking `Launch`; counted tickets with `lifecycleWork`/`watchers` waits remain (`3e18b0b`).
- Queue sweep runs after `lifecycleWork.Wait()`, and late-created teams are snapshotted after admission quiesces (`915d33d`, `2ef453f`).
- Post-dequeue shutdown rejection in `afterExitAdmitted` is present (`915d33d`).
- `Scheduler`/`Close` handoff is documented and covered by `TestSchedulerCreatedAfterSwarmCloseIsClosed`.

The remaining gap is the launch commit race above, not a regression of those fixes.

## Rechecked blind-search candidates (not findings)

These were raised during review but do not hold against current code on this head:

- **`ErrSwarmClosed` leaves tasks `pending`** — incorrect. `launchAdmitted` calls `coord.Fail` on every launch error, including `ErrSwarmClosed`, before `releaseSlot()` (`lifecycle.go:80-84`). Queued specs are failed by `clearQueue` after `lifecycleWork.Wait()`.
- **Scheduler wedged on ghost `pending` spawn** — same root cause; launch rejection fails the coordinator task, it does not stay `pending`.
- **`Spawn`/`Handoff` return IDs while `Close` later fails queued work** — expected shutdown race when admission wins before `Close`; `clearQueue` terminalizes those tasks.
- **Queued tasks stay `pending` until `clearQueue`** — intentional during `Close`; not stranded after `Close` returns.
- **`CollectWait` during `Close`** — expected graceful-shutdown blocking until watchers exit or the collect context times out; not a post-`Close` strand.
- **Retry records the original member error instead of `ErrSwarmClosed`** — misleading `task.Err` only; task still reaches a terminal failed state (`lifecycle.go:124-138`).

Optional hardening (not blocking this PR): `errors.Is` for `ErrSwarmClosed` in `launchAdmitted`, consistent shutdown error strings, and a regression test that interleaves `Close` between the `closed` read and `Launch`.

PierrunoYT:
PierrunoYT addressed the remaining launch-commit race in `30699bf`.

- Shutdown state is rechecked after `Launcher.Launch` returns and before member registration/status/watcher commit.
- A launch that loses to `Close` is reaped, failed with `ErrSwarmClosed`, and releases its slot without registering a member.
- Added a deterministic interleaving regression.

Validation: full `internal/swarm` suite passes with `-race`; formatting and diff checks pass. Ready for re-review.

jatmn (changes requested):

## Findings

- [P2] Apply the post-`Launch` shutdown gate to member retries, not only initial launches
  `internal/swarm/lifecycle.go:143-151`
  `30699bf` added the right fix for the spawn/queue-drain paths: after `MemberLauncher.Launch` returns, `launchAdmitted` re-checks `closed` under `lifecycleMu`, reaps the handle with `handle.Wait()`, fails the task, and releases the slot instead of registering a new running member. The bounded-retry path in `watch` still calls `launchMemberAdmitted` directly and, on `relErr == nil`, swaps `m.handle` and continues with no post-`Launch` guard. Production wiring uses `FuncLauncher` (`member.go:64-81`), whose `Launch` always returns immediately and starts `Run` in a goroutine without consulting `ctx.Err()` at entry, so the same close-between-precheck-and-return window `TestCloseBetweenLaunchPrecheckAndReturn` covers for spawns remains open on retries: admission can win, `Close` can set `closed` and cancel `baseCtx`, `Launch` can still return a live handle, and the watcher resumes supervising a relaunched member after shutdown began. That leaves `30699bf` incomplete relative to its own spawn fix and the PR's stated contract to reject retry work once shutdown starts. The common shutdown path is already covered — `TestClosePreventsMemberRetry` shows that when `closed` is already set at retry time, `beginLifecycleAdmission` refuses the relaunch — and this PR is still strictly better than `main` on retry shutdown. The remaining gap is the narrow race where a retryable exit, admission, and `Close` interleave so `Launch` straddles the boundary; impact is softer than the spawn case because no new member is registered and `baseCtx` is already cancelled, but a relaunched `Run` goroutine can still start. Mirror the `launchAdmitted` post-`Launch` check and handle reap before assigning `m.handle` (or extract a shared commit helper), and add a deterministic retry interleaving regression parallel to `TestCloseBetweenLaunchPrecheckAndReturn`.

## Rechecked from prior review

On head `30699bf7928d0f6ca395eff7e2d36c94afa37526`, the earlier shutdown findings look addressed:

- Lifecycle admission no longer holds `lifecycleMu` across blocking `Launch`; counted tickets with `lifecycleWork`/`watchers` waits remain (`3e18b0b`).
- Queue sweep runs after `lifecycleWork.Wait()`, and late-created teams are snapshotted after admission quiesces (`915d33d`, `2ef453f`).
- Post-dequeue shutdown rejection in `afterExitAdmitted` is present (`915d33d`).
- Initial-launch post-`Launch` commit rejection and handle reap are present in `launchAdmitted` (`30699bf`), with `TestCloseBetweenLaunchPrecheckAndReturn`.
- `Scheduler`/`Close` handoff is documented and covered by `TestSchedulerCreatedAfterSwarmCloseIsClosed`.

The remaining gap is the retry relaunch window above, not a regression of those fixes.

PierrunoYT:
Addressed the [P2] retry gap from the latest review in `5650615`.

**Fix:** the post-`Launch` commit check that `30699bf` added to `launchAdmitted` is now shared, not duplicated. It is extracted as `commitLaunch(handle, adopt)` — re-check `closed` under `lifecycleMu.RLock()`, run the caller's adoption work inside that lock, and on shutdown reap the handle with `handle.Wait()` and report refusal. The bounded-retry path in `watch` now goes through `relaunchAdmitted`, which runs the same three gates an initial launch does: admission, the pre-`Launch` re-check, and the post-`Launch` commit. When any gate refuses, the handle is reaped and the watcher records the member's terminal failure instead of swapping `m.handle` and continuing to supervise.

**Regression test:** `TestCloseBetweenRetryLaunchPrecheckAndReturn`, the deterministic retry parallel to `TestCloseBetweenLaunchPrecheckAndReturn` you asked for. The first member exits `ErrMemberTemporary`; the launcher blocks inside the relaunch's `Launch` until the test has driven `Close` to set `closed` and cancel `baseCtx`, then returns a live handle whose member ignores the cancelled context and succeeds. The test asserts the late handle is waited exactly once (reaped, not supervised) and the task ends `failed` with the late member's result discarded.

I confirmed it discriminates: with the commit gate removed from the retry path it fails with `Status:done Result:"late result"` — the unfixed code records a post-close member's success as the task outcome. With the gate it passes.

I skipped the optional hardening (`errors.Is` for `ErrSwarmClosed`, shutdown error-string consistency) to keep this diff on the reported defect; happy to fold it in if you'd rather have it here.

**Validation**
- `go build ./...`
- `go test ./internal/swarm -count=1`
- `go test -race ./internal/swarm -count=1` (WSL/Linux)
- `go test ./internal/swarm -run TestCloseBetweenRetryLaunchPrecheckAndReturn -count=5`
- `go vet ./internal/swarm`
- `go test ./...`
- `gofmt -l internal/swarm` (clean for the touched files), `git diff --check`

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):
lgtm

PR #778, opened 2026-07-20
fix(tui): stop the permission card's interior clashing on cool themes

gnanam1990:
## The problem

The focused permission prompt was the only prompt card in the TUI that used brand/warning colours for its **interior**, and both read badly on cool themes such as dracula — a warm brown-yellow box over a purple palette:

1. **The selected option was painted with the brand chip.** `renderFocusedPermissionPrompt` filled the highlighted row ("Yes, proceed") with `zeroTheme.badge` — the style behind the short brand chips (` 0 `, ` ASK `, ` SPEC REVIEW `), a full-brightness accent fill. Every other selectable list in the TUI highlights its selected row with `onSel`/`selBg` (`provider_wizard.go`, `provider_manager.go`, the `/` palette). The permission card was the outlier, and it was also the one line in the card that skipped the interior `fill()` every other line composed onto.

2. **The card body was tinted with `permBg`.** An amber-family wash meant to signal a warning surface. It suits the warm default palette, but on cool themes it renders as a brown-yellow slab, and it made the permission card the *only* prompt whose body carries a fill at all — `ask_user`, spec review and plan all use a transparent body (`lipgloss.NewStyle()`).

## The fix

Make the card's interior match the rest of the TUI, and let the warning identity come from the frame rather than the fill:

- **Selected row** → `onSel(ink)`, the same muted selection tint every other picker uses.
- **Card body** → transparent, matching `ask_user` / spec / plan.
- **Unchanged:** the amber `PERMISSION` badge and the amber-mixed border still mark it as a permission gate. Only the two clashing interior fills go.

Measured before/after in the dracula palette, at the "Yes, proceed" row and the body:

| element | before | after |
|---|---|---|
| selected row bg | `#bd93f9` (bright brand accent) | `#504482` (`selBg`, muted) |
| card body bg | `#322a1e` (`permBg`, warm slab) | transparent |
| PERMISSION badge | `#ffb86c` amber | `#ffb86c` amber (unchanged) |

The same relative change holds on the warm default palette (`#caff3f` → `#32401b` for the row, `#1c1915` → transparent for the body); it is simply less noticeable there because the warm fills already matched that theme.

## Verification

`TestFocusedPermissionSelectedRowUsesSelectionTintNotBrandChip` asserts against **palette entries, not hardcoded hex**, so it survives a theme retune: the selected row must carry `selBg` and *not* the brand accent, the body must carry *no* `permBg` wash, and the `PERMISSION` badge must keep its amber fill so the card still reads as a warning surface.

Mutation-tested both halves: restoring `zeroTheme.badge` fails with *"selected row is filled with the brand accent"*; restoring the `permBg` body fill fails with *"permission card body still tinted with permBg… want a transparent body"*.

`gofmt` and `go vet ./internal/tui/` clean. Full `internal/tui` suite has the same single failure as `main` (`TestHandleAddDirCommand`, a pre-existing sandbox/TMPDIR artifact) — no new failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the focused permission prompt so the selected option uses the standard selection highlight.
  * Preserved the permission indicator’s amber styling while removing the previous warm background tint from the dialog.
  * Improved visual consistency between selected permission options and other selectable rows.

### DIFF
--- a/internal/tui/permission_prompt_test.go
+++ b/internal/tui/permission_prompt_test.go
@@ -2,6 +2,7 @@ package tui
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


anandh8x (approved):
Approved. The focused permission card now uses the shared selected-row tint while retaining the permission badge and border as the warning signal. Focused TUI coverage and CI pass.

Vasanthdev2004 (approved):
Approve. Tightly scoped fix, and the focused permission card really was the odd one out: it washed its whole body with permBg and filled the selected row with the brand badge, so it clashed on cool themes like dracula. Dropping the body wash and switching the selected row to the standard onSel(ink) tint brings it in line with the other prompt cards and the wizard/palette lists, while the amber PERMISSION badge and border keep the "this is a gate" signal.

Checked on head a1bc1c4a:

- Built and ran internal/tui. The new TestFocusedPermissionSelectedRowUsesSelectionTintNotBrandChip passes and is a genuine regression guard: it pins the selected row to selBg (not accent), keeps the amber badge, and asserts the body no longer carries permBg. Reverting either half of the change trips it.
- The ink-on-selBg selected label stays readable on every theme: theme_select_test.go already asserts wcagRatio(ink, selBg) >= 4.5 per palette, so moving off the accent fill costs no contrast.
- go vet and gofmt clean. Heads up if you run internal/tui locally: TestHandleAddDirCommand and TestAltScreenTranscriptScrollKeepsFooterFixed can fail, but they are pre-existing and environment-sensitive (they trip on plain main too) and unrelated to this diff. CI is green on all three OSes.

One non-blocking nit for a later sweep, not this PR: removing the two call sites leaves a few theme symbols with no reader left, namely the onPerm method, the resolved bgPerm color, and the resolved permBg style field. The permBg palette string stays live because the amber-on-permBg WCAG test still reads it, though that assertion is now vestigial since nothing paints on permBg anymore. go vet does not flag unused methods or fields so CI stays green; just flagging it for tidy-up.

LGTM.

PR #779, opened 2026-07-20
fix(tools): read_file recovers a backwards line range instead of erroring

gnanam1990:
## The problem

`read_file` hard-failed when a caller passed `end_line < start_line`:

```
Error: Invalid arguments for read_file: end_line must be greater than or equal to start_line
```

That is a caller arithmetic slip (e.g. `start_line: 3, end_line: 2` while paging through a file), and hard-erroring costs a whole retry turn. It was also **inconsistent** with the two neighbouring bad-range cases in `renderReadFileRange`:

```go
if startLine > total { return okResult("…start_line is past the end…") }   // friendly note
if endLine == 0 || endLine > total { endLine = total }                       // auto-clamped
if endLine < startLine { return errorResult("…") }                           // hard error ← the outlier
```

`start_line` past EOF returns a friendly note; `end_line` past EOF is silently clamped down; only a backwards range hard-failed.

## The fix

Recover it the same way the others recover — clamp `end_line` **up** to `start_line` (read just that one line) — and surface the adjustment in the output so it is never silent:

```
File: n.txt (lines 3-3 of 5)
(note: end_line 2 was before start_line 3, so only line 3 was read — pass end_line >= start_line to read a wider range)

3 | c
```

The model gets a usable result **and** learns its range was off, instead of burning a turn on an error.

## No performance cost

The change is in the argument-validation phase, **before any file is opened** — one branch that returned an error now sets a note and clamps an integer. No new I/O, no loop, the file-reading path (`appendReadFileRange`) is untouched. `read_file` is one call per invocation, not a hot loop. If anything it saves the agent a wasted retry.

## Verification

- `TestReadFileToolRecoversBackwardsRange` — recovers, shows the note, reads only `start_line`, leaks no adjacent lines. **Mutation-verified**: restoring the hard error fails it.
- `TestReadFileToolValidRangeUnchangedByRecovery` — a valid `end_line >= start_line` range carries **no** note and reads the byte-identical window as before.
- `gofmt`/`go vet` clean. `internal/tools` has the same three pre-existing sandbox failures on this branch and on `main` — no new failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-range handling when the ending line precedes the starting line.
  * The tool now treats this as a recoverable input and returns only the requested starting line.
  * The rendered output includes a visible note explaining the adjustment, instead of failing.
  * Valid line ranges continue to return the expected content with no additional notices.
* **Tests**
  * Added unit tests covering both backward-range recovery and unchanged output for valid ranges.

### DIFF
--- a/internal/tools/file_tools_test.go
+++ b/internal/tools/file_tools_test.go
@@ -879,12 +879,11 @@ func TestReadFileToolValidRangeUnchangedByRecovery(t *testing.T) {
 	if result.Status != StatusOK {
 		t.Fatalf("valid range failed: %s: %s", result.Status, result.Output)
 	}
-	if strings.Contains(result.Output, "was before start_line") {
-		t.Fatalf("valid range must carry no recovery note: %q", result.Output)
-	}
-	for _, want := range []string{"File: notes.txt (lines 2-4 of 5)", "2 | beta", "3 | gamma", "4 | delta"} {
-		if !strings.Contains(result.Output, want) {
-			t.Fatalf("expected %q, got %q", want, result.Output)
-		}
+	// Byte-for-byte: header, blank separator, the exact selected lines, and no
+	// recovery note or trailing content. An altered header/separator or a stray
+	// out-of-range line would fail this where a substring check would not.
+	const want = "File: notes.txt (lines 2-4 of 5)\n\n2 | beta\n3 | gamma\n4 | delta"
+	if result.Output != want {
+		t.Fatalf("valid-range output changed by the recovery path:\n got: %q\nwant: %q", result.Output, want)
 	}
 }
--- a/internal/tools/file_tools_test.go
+++ b/internal/tools/file_tools_test.go
@@ -829,3 +829,62 @@ func TestGlobSkipsWorkspaceExcludedDirectoriesAndBinaryFiles(t *testing.T) {
 		t.Fatalf("expected only keep.txt, got:\n%s", res.Output)
 	}
 }
+
+// A backwards range (end_line < start_line) is a caller-arithmetic slip, not a
+// fatal error: read_file recovers by reading just start_line and surfaces a note
+// explaining the adjustment, matching how an end_line past EOF is auto-clamped.
+// It must never hard-error the way it used to.
+func TestReadFileToolRecoversBackwardsRange(t *testing.T) {
+	root := t.TempDir()
+	writeTestFile(t, filepath.Join(root, "notes.txt"), "alpha\nbeta\ngamma\ndelta\nepsilon")
+
+	result := NewReadFileTool(root).Run(context.Background(), map[string]any{
+		"path":       "notes.txt",
+		"start_line": 3,
+		"end_line":   2, // backwards
+	})
+
+	if result.Status != StatusOK {
+		t.Fatalf("backwards range must recover, not error; got %s: %s", result.Status, result.Output)
+	}
+	if strings.Contains(result.Output, "must be greater than or equal to start_line") {
+		t.Fatalf("backwards range still hard-errors: %q", result.Output)
+	}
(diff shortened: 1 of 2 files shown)


anandh8x (changes requested):
Request changes for regression coverage: TestReadFileToolValidRangeUnchangedByRecovery says valid output is byte-for-byte unchanged, but it only asserts substrings. Compare result.Output to the complete expected rendering, including header, separators, selected lines, and trailing formatting, while keeping the no-recovery-note assertion.

Vasanthdev2004 (dismissed):
Approve. Right call: a backwards range is caller arithmetic, not a fatal condition, and hard-erroring there cost a whole turn. Clamping end_line up to start_line and surfacing a note matches how an end_line past EOF already auto-clamps, so read_file now recovers consistently instead of leaving this one case as the outlier.

Checked on head 52b3338a:

- Built and ran internal/tools. Both new tests pass. The recovery test is a real guard (StatusOK, exact note text, reads only start_line, no adjacent-line leak, old error string gone), and the valid-range test pins byte-identity: with no note, header + "\n" + "\n" collapses to the original header + "\n\n", so normal reads are untouched.
- Edge cases hold. end_line == 0 is clamped to total before the backwards check, and the arg minimum is 1, so a caller cannot reach the branch with a value below 1. start_line past EOF still early-returns first, so the recovery only fires on a genuine start > end slip.
- fmt is already imported, go vet and gofmt are clean, and CI is now green on all three OSes.

LGTM.

Vasanthdev2004:
My approve above went up before your two reviews rendered (they landed while I was building and testing), so let me square the record: your ask is right, and I will walk back my "test pins byte-identity" wording. TestReadFileToolValidRangeUnchangedByRecovery only checks a few substrings plus the absence of the note, so a regression that shifted a separator or leaked an extra out-of-range line would still pass. The name and comment promise more than the assertion delivers.

Worth being precise about what is solid versus not: the code path itself is byte-identical for a valid range (with no note, header + "\n" + "\n" is the same bytes as the old header + "\n\n", and nothing else on that path changed), so this is a test-guard gap, not a behavior bug. Tightening the test to compare result.Output against the full expected rendering (header, no note, the exact selected lines, trailing format) makes the guard match its claim. Over to gnanam for that; the fix itself is good.

gnanam1990:
Addressed in `fe98067`. @anandh8x @Vasanthdev2004 @coderabbitai

`TestReadFileToolValidRangeUnchangedByRecovery` now asserts the **complete** output, not substrings:

```go
const want = "File: notes.txt (lines 2-4 of 5)\n\n2 | beta\n3 | gamma\n4 | delta"
if result.Output != want { … }
```

An altered header, separator, or a stray out-of-range line now fails it, and the no-recovery-note guarantee is implicit in the exact string. The backwards-range recovery test is unchanged and still mutation-verified.

anandh8x (approved):
Approved on the latest head. The valid-range regression now compares the complete expected output, covering the header, separator, selected lines, and trailing formatting. Focused tests pass.

Vasanthdev2004 (approved):
Re-approving on fe980671. My earlier approve was dismissed by this push, and the new commit is exactly the test-strengthening you all asked for: TestReadFileToolValidRangeUnchangedByRecovery now compares result.Output against the full expected rendering (header, blank separator, the three exact lines, no note) instead of loose substrings, so a shifted separator or a stray out-of-range line would fail it now. Since it is a byte-exact assertion I ran it on Windows too, to be sure the "\n" separators do not turn into CRLF: passes there as well. The code path itself is unchanged from the version I already went through, so nothing else to re-verify. Thanks for the quick turnaround, and that lines up with anandh8x and CodeRabbit already re-approving the new head.

LGTM.

PR #780, opened 2026-07-20
feat(tui): permission prompt takes free-text feedback inline

gnanam1990:
## What this adds

The permission prompt's **"No, and tell Zero what to do differently"** option finally does what its label says. Today it silently maps to *cancel*: the card closes, the run aborts, and you have to go type your correction in the composer as a fresh turn. Now it opens an inline free-text field **on the card**:

```
 PERMISSION
bash  shell
Tell Zero what to do differently:
❯ use apply_patch instead▌
enter · send to Zero    esc · back to options
```

Type an instruction → **Enter** sends it → the model reads it and adjusts **in the same turn**. **Esc** returns to the option list; **empty + Enter** falls back to a plain cancel (the option's prior meaning).

## Why it needs no agent-side change

`PermissionDecision.Reason` already carries free text back to the loop, and the deny path already surfaces it to the model as the tool result (`deniedPermissionResult`). So submitting sends **Deny with `Reason` = your text**, and the model sees "denied: *use apply_patch instead*" and continues. This is deliberately Deny, not Cancel: in the loop, Cancel returns an error and aborts the run, while Deny returns a tool result and keeps going — which is the whole point of "tell Zero what to do differently."

The UI mirrors the existing `ask_user` "type your own answer" pattern: the feedback field reuses the shared composer input, and the card renders it in the transcript body.

## Bundled: the permission prompt now suppresses the composer

A focused permission prompt owns the keyboard — its options and the feedback field consume every key — so the composer below it is inert. `ask_user` already hides the composer while focused; the permission prompt didn't, leaving a "describe a task for zero…" box that looked typable but wasn't. It now suppresses the composer (and idle hints / plan panel), keeping only the status line, exactly like the `ask_user` modal. This is also required for the feature: without it, the shared input would echo the feedback text in two places (card + composer).

## Verification

Six code sites (a `typing` flag on `pendingPermissionPrompt`, the confirm handler, the `[n]` hotkey, submit, Esc back-out, and the key routing that lets text reach the shared input) plus the card render and the footer suppression. Tests cover:

- open (does not resolve yet), submit-as-Deny-with-the-typed-text, empty-cancels, Esc-returns-to-options;
- composer hidden in both option and feedback mode, and the feedback text rendered **exactly once**.

`gofmt`/`go vet` clean. Full `internal/tui` suite has the same single pre-existing failure as `main` (`TestHandleAddDirCommand`, a sandbox/TMPDIR artifact) — no new failures.

## Summary by CodeRabbit

* **New Features**
  * Added inline free-text “tell Zero what to do differently” feedback mode to permission prompts, using **Enter** to submit (empty cancels) and **Esc** to return to options without resolving.
  * Arrow/Tab navigation is disabled while typing; Enter/feedback submission drives the deny reason from the typed text.
* **UI Improvements**
  * While any permission prompt is active, composer/idle hints and pinned-plan UI are suppressed (composer drafts remain preserved through submit/cancel).
  * Permission option rows are non-clickable during typing; option feedback input is trimmed before submission.
* **Tests**
  * Expanded regression coverage for typing mode, composer suppression, draft preservation, Esc/Enter behavior, and backspace attachment handling.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -695,6 +695,11 @@ type pendingPermissionPrompt struct {
 	// whose Reason is the typed text, so the model reads it as the tool result and
 	// adjusts course in the same turn instead of the run being cancelled.
 	typing bool
+	// savedDraft holds whatever was in the shared composer input when feedback
+	// mode was entered. The field is cleared for typing and restored on both
+	// submit and cancel, so a half-typed or queued next-turn message survives the
+	// detour (permissionRequestMsg, unlike ask_user, does not clear the composer).
+	savedDraft string
 }
 
 // askUserRequestMsg is the TUI-loop equivalent of permissionRequestMsg: the
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -1675,6 +1675,16 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, nil
 			}
 		case keyBackspace(msg):
+			// In permission feedback mode Backspace is a plain edit of the feedback
+			// text. This case runs before the typing branch below and, on an empty
+			// field (feedback mode clears the composer), would otherwise fall to the
+			// removeLastAttachment path and silently drop a staged image/doc that
+			// savedDraft does not restore. Route it to the shared input instead.
+			if m.pendingPermission != nil && m.pendingPermission.typing {
+				var cmd tea.Cmd
+				m.input, cmd = m.input.Update(msg)
+				return m, cmd
+			}
 			if m.picker != nil {
 				if m.modelPickerIsLoading() {
 					return m, nil
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -689,6 +689,12 @@ type pendingPermissionPrompt struct {
 	// resting approval choice. Moved by ↑/↓/Tab; confirmed by Enter or a click.
 	// Hotkeys resolve the matching request-provided option directly.
 	cursor int
+	// typing is true once the user chose "tell Zero what to do differently": the
+	// card replaces its option list with a free-text field (sharing the composer
+	// input, like the ask_user questionnaire). Submitting sends a Deny decision
+	// whose Reason is the typed text, so the model reads it as the tool result and
+	// adjusts course in the same turn instead of the run being cancelled.
+	typing bool
 }
 
 // askUserRequestMsg is the TUI-loop equivalent of permissionRequestMsg: the
@@ -1439,6 +1445,11 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 			if m.pendingAskUser != nil {
 				return m.escapeAskUser()
 			}
+			// Esc in the permission feedback field steps back to the option list
+			// rather than resolving, so a stray keystroke is recoverable.
+			if m.pendingPermission != nil && m.pendingPermission.typing {
+				return m.cancelPermissionTyping()
+			}
 			if m.pendingSpecReview != nil {
 				m.burstCount = 0
 				return m.cancelSpecReview()
@@ -1854,6 +1865,15 @@ func (m model) updateModel(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m.handleSpecReviewKey(msg)
 		}
 		if m.pendingPermission != nil {
+			// Feedback mode: a printable keystroke (and editing keys like
+			// backspace) types into the shared composer input, mirroring the
+			// ask_user free-text path above. Enter/Esc/↑/↓ were already handled
+			// earlier in this switch; the remaining keys reach the input here.
+			if m.pendingPermission.typing {
+				var cmd tea.Cmd
+				m.input, cmd = m.input.Update(msg)
+				return m, cmd
+			}
 			m.burstCount = 0
 			return m.handlePermissionKey(msg)
 		}
@@ -2781,6 +2801,16 @@ func (m model) footerView(width int) string {
 	if m.pendingAskUser != nil {
 		footer.WriteString(renderAskUserQuestionnaire(*m.pendingAskUser, m.input.Value(), width))
 		footer.WriteString("\n")
+		footer.WriteString(m.statusLine(width))
+		return footer.String()
+	}
+	// A focused permission prompt owns the keyboard: its options (and the feedback
+	// field) consume every key, so the composer is inert. Suppress it and the idle
+	// hints/plan panel like the ask_user modal above, keeping only the status line.
+	// The card itself renders in the transcript body. This also keeps the shared
+	// input from echoing in two places once "tell Zero what to do differently"
+	// opens the on-card feedback field.
+	if m.pendingPermission != nil {
 		footer.WriteString(m.statusLine(width))
 		return footer.String()
 	}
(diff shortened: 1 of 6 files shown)


anandh8x (changes requested):
Request changes: entering permission feedback clears the shared composer input, but neither submit nor Esc restores the existing draft. A user can lose an in-progress or queued next-turn message. Snapshot the draft before feedback mode, restore it after both submit and cancel, and add regression coverage for both paths.

gnanam1990:
Addressed in `16b69cc` (also rebased onto main now that #778 merged — the permission-card colour changes and this feature both touch `renderFocusedPermissionPrompt`, and the merged result keeps both). @anandh8x @Vasanthdev2004 @coderabbitai

The composer-draft loss is fixed: `pendingPermissionPrompt` gains a `savedDraft` field. `choosePermissionOption` snapshots `m.input.Value()` before clearing it for the feedback field, and **both** exits restore it — `submitPermissionFeedback` and `cancelPermissionTyping`. Only the feedback text is consumed (via the decision Reason); a half-typed or queued next-turn message survives the detour.

Two regression tests cover it — `TestPermissionFeedbackPreservesComposerDraftOnSubmit` and `...OnCancel` — and it's mutation-verified: reverting to `SetValue("")` fails with "composer draft not restored after submit".

anandh8x (dismissed):
Approved on the latest head. Permission feedback now preserves and restores the existing composer draft on both submit and Esc-cancel, while keeping the feedback text as the denial reason. Regression coverage and focused tests pass.

gnanam1990:
@coderabbitai — status on both findings, without a push (to avoid dismissing @anandh8x's approval of the current head `16b69cc0`):

**1. `permission_prompt.go:157` — composer draft loss (Major): already resolved on this head.** `pendingPermissionPrompt` has a `savedDraft` field; `choosePermissionOption` snapshots `m.input.Value()` before clearing it, and it is restored on **both** exits — `submitPermissionFeedback` and `cancelPermissionTyping`. Covered by `TestPermissionFeedbackPreservesComposerDraftOnSubmit` and `...OnCancel`, mutation-verified (reverting to `SetValue("")` fails "composer draft not restored after submit"). This is the same issue anand raised, and his approval is on this exact commit. Please resolve this thread.

**2. `permission_prompt_test.go:496` — guard `len(got)` before `got[0]` (Minor): valid, intentionally held.** It's test-only hardening — the submission resolves in this test, so `got[0]` never actually panics; it would only change a hypothetical failure from a panic into a cleaner message. I'm not force-pushing a one-line test guard because a push dismisses anand's just-given approval on this head (this repo dismisses stale approvals on push). I'll fold it in the moment the PR needs another push for any substantive reason.

gnanam1990:
@coderabbit approve

Vasanthdev2004 (changes requested):
Really like this one. The "tell Zero what to do differently" option finally doing what its label says is a genuine win, and routing it through Deny-with-Reason (so the model reads the instruction and keeps going) instead of Cancel is exactly right. I traced the whole key and mouse dispatch on the security question and I am satisfied it is safe: while the feedback field is open, hotkeys route into the input as text, Enter goes to submit (Deny, or empty falls back to Cancel), Esc backs out, and no path reaches an Allow decision. The composer-draft fix reads correctly too, with tests both ways.

Three things from that pass, all small, that I would like in before this lands on the permission path:

**1. Real bug: Backspace on an empty feedback field drops a staged attachment.** The `keyBackspace` case in model.go (~1677) runs before the typing branch (~1872), and entering feedback clears the input, so `composerValue()` is "". Backspace therefore hits the `removeLastAttachment` path and silently drops the last staged image/doc. `savedDraft` only snapshots text, so it is not restored on submit or cancel. Repro: stage an image with no composer text, press `[n]`, press Backspace, and the attachment is gone. Simplest fix is to guard that attachment-removal branch (or the whole `keyBackspace` case) on `!m.pendingPermission.typing`, so the feedback field's backspace is just an inert edit.

**2. Worth hardening here: the mouse click resolver has no typing guard.** transcript_selection.go (~1404) calls `resolvePermission(line.permChoice)` for a clicked option row, Allow choices included, with no check on `typing`. It is safe today only because `renderFocusedPermissionPrompt` returns nil offsets while typing, so no option rows are clickable during feedback. That is a single non-obvious early-return carrying the whole guarantee, and #778 edits that same function, so a rebase or refactor could quietly remove it. A one-line `!m.pendingPermission.typing` guard on the click path makes the safety explicit rather than emergent.

**3. Nit: TestPermissionFeedbackPreservesComposerDraftOnSubmit indexes `got[0]` without first asserting `len(got) == 1`** (its sibling TestPermissionFeedbackSubmitSendsDenyWithText does). Same point CodeRabbit raised. Add the length guard so a regression fails cleanly instead of panicking.

One unrelated heads-up: this and #778 both edit `renderFocusedPermissionPrompt`, and #778 drops the `permBg` body wash for a transparent body. Whichever lands second will need to reconcile, and the new feedback-field render here should pick up that transparent-body treatment so the card stays consistent.

None of this touches the core design, which is solid. Happy to re-approve once the backspace edge and the click guard are in.

gnanam1990:
Pushed `91c6e76` with all three, @Vasanthdev2004 — thanks, the backspace one was a real bug I'd missed.

**1. Backspace drops staged attachment (bug):** fixed. The `keyBackspace` case now routes to the shared input when `pendingPermission.typing`, so Backspace in the feedback field is a plain text edit and never reaches `removeLastAttachment`. `TestPermissionFeedbackBackspaceKeepsStagedAttachment` covers it — mutation-verified: neutralising the guard fails with "dropped the staged attachment: labels=[] imgs=0".

**2. Click resolver typing guard:** added `!(m.pendingPermission != nil && m.pendingPermission.typing)` on the `permOption` branch in `handleTranscriptSelectionMouse`, so the safety is explicit rather than resting on the nil-offsets early-return. `TestPermissionFeedbackRendersNoClickableOptionsWhileTyping` pins the primary invariant (feedback mode registers zero clickable offsets).

**3. `len(got)` guard:** added before `got[0]` in the draft-preservation test.

On the transparent-body heads-up: this branch was already rebased onto main after #778 merged, and the merge kept #778's transparent-body treatment — the feedback-field render composes on the same card, so it's consistent. Ready for another look when you are.

One note for @anandh8x / @coderabbitai: this push supersedes your approvals of `16b69cc` (the repo dismisses on push). The delta since your approval is exactly these three review items — no change to the core design you approved.

Vasanthdev2004 (approved):
Re-approving on 91c6e768. All three land cleanly, each with a regression test:

- Backspace in feedback mode now routes to the input before the removeLastAttachment path, so a staged image/doc survives. TestPermissionFeedbackBackspaceKeepsStagedAttachment stages one, opens the field, backspaces, and checks it is still there. That was the real bug, fixed at the right spot.
- The click resolver now carries the explicit `!typing` guard, and TestPermissionFeedbackRendersNoClickableOptionsWhileTyping pins that the renderer registers zero clickable offsets in feedback mode. Belt and suspenders on the one path that could ever land an accidental Allow, which is what I wanted given #778 edits the same function.
- The `len(got) == 1` guard is in before the `got[0]` index.

Built and ran internal/tui locally: gofmt, vet and build clean, and all nine feedback tests pass including the two new ones. The security trace is unchanged and if anything tighter, since the click guard only ever removes a resolve path. Nice quick turnaround.

Still worth coordinating the #778 / #780 merge on renderFocusedPermissionPrompt whenever the second one lands (permBg body wash versus transparent body), but that is not this PR's problem. LGTM.

anandh8x (approved):

Issue #743, opened 2026-07-18
Resolve existing golangci-lint baseline (35 findings)


PierrunoYT:
## Summary

The repository-wide lint command currently reports 35 pre-existing findings:

- 27 `staticcheck` findings
- 6 `unused` findings
- 2 `ineffassign` findings

These findings are unrelated to PR #725, but prevent contributors from completing the mandatory lint check with a clean result.

## Reproduction

Because of the toolchain problem tracked in #680, explicitly select Go 1.26.5:

```powershell
$env:GOTOOLCHAIN = 'go1.26.5'
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --enable-only unused,ineffassign,staticcheck ./...
```

### ineffassign

- `internal/cli/mcp_config.go:737`
- `internal/cli/mcp_config.go:744`

### unused

- `internal/sandbox/runner.go:385` - `existingBubblewrapMounts`
- `internal/sandbox/runner.go:795` - `denyWriteRules`
- `internal/sandbox/runner.go:827` - `networkRuleFor`
- `internal/tui/model.go:2950` - `model.scrollableTranscriptView`
- `internal/tui/streaming_fade.go:284` - `model.ensureAgeTickReschedule`
- `internal/tui/transcript_selection.go:1278` - `transcriptViewportStartForLayout`

### staticcheck

- `internal/agenteval/score.go:128` (`S1011`)
- `internal/agentinit/agentinit.go:43,48,50` (`QF1012`)
- `internal/cli/agent_eval.go:518` (`S1011`)
- `internal/cli/provider_detect.go:86` (`QF1002`)
- `internal/dictation/transcriber_deepgram.go:103` (`ST1005`)
- `internal/lsp/navigate_test.go:67,79` (`SA1012`)
- `internal/provideronboarding/localruntime_test.go:65` (`S1009`)
- `internal/sandbox/adapters.go:228` (`QF1001`)
- `internal/sandbox/command_prefix.go:156` (`S1008`)
- `internal/sandbox/engine.go:236` (`QF1001`)
- `internal/sandbox/landlock_other.go:7` (`ST1005`)
- `internal/sandbox/windows_network.go:13` (`ST1005`)
- `internal/specialist/exec.go:830` (`QF1008`)
- `internal/tui/files_git_sweep_test.go:195` (`SA1012`)
- `internal/tui/model_test.go:1399` (`SA4006`)
[...]

PR #782, opened 2026-07-20
fix(acp): interrupt idle reads on cancellation

### LINKED ISSUE #777: fix(acp): cancellation cannot interrupt an idle connection read
## Summary

`Serve(ctx)` documents context-driven shutdown, but cancellation cannot interrupt an idle blocking read from the ACP JSON-RPC transport. In the CLI, SIGINT cancels the context without closing stdin, so shutdown can remain blocked until the peer closes the pipe or sends more data.

This tracks **AUD-011** from the July 18 codebase audit.

## Affected code

- `internal/acp/jsonrpc.go:116-146`
- `internal/cli/acp.go:71-76`
- `internal/acp/jsonrpc_test.go:11-25`

## Current behavior

The serving loop blocks in `ReadBytes` while the peer is idle. Canceling the supplied context does not unblock that read. Existing tests close the pipe as well as canceling the context, which masks the cancellation problem.

## Expected behavior

Canceling the context should promptly stop an idle ACP server without requiring the peer to send data or close its side of the transport.

## Suggested implementation

- Give the connection an owned `io.Closer` or explicit transport-close callback.
- Start a context watcher that closes the read transport when cancellation occurs.
- Ensure shutdown and transport closure remain safe and idempotent.
- Preserve the original protocol or transport error where appropriate.

## Suggested tests

- Run `Serve` with an idle `io.PipeReader`.
- Cancel the context without closing the writer.
- Assert that `Serve` returns promptly.
- Cover concurrent cancellation and peer closure to detect double-close or shutdown races.

PierrunoYT:
## Summary

- retain a closable ACP input so context cancellation can interrupt an idle blocking read
- arbitrate cancellation and terminal read completion exactly once, preserving EOF and genuine read errors when the read finishes first
- join the cancellation watcher before `Serve` returns
- cover active idle-read cancellation and terminal-error races for closable and non-closable readers

## Validation

- `go test ./internal/acp -count=1`
- focused cancellation/error regressions with `-count=100`
- `go test ./internal/cli -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...` with `GOTOOLCHAIN=go1.26.5` — no vulnerabilities found
- `gofmt -l internal/acp/jsonrpc.go internal/acp/jsonrpc_test.go`
- `git diff --check`

The pinned golangci-lint command still reports the repository's 35 known baseline findings tracked by #743; none are in the files changed here.

Fixes #777

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ACP JSON-RPC cancellation behavior to safely interrupt idle inbound reads and complete shutdown promptly.
  * Ensures correct outcome when cancellation races with terminal inbound read errors (including buffered errors), returning the terminal error when appropriate.
  * Reports inbound reader panics as terminal errors, avoiding misclassification during cancellation.
  * Updated CLI to treat standard input as owned for the command lifetime so cancellation can interrupt blocked reads, and to fail on any `Serve` error.
* **Tests**
  * Added regression and integration-style coverage for cancellation races, buffered error handling, cancellation-vs-completion timing, panic reporting, and SIGINT behavior.


jatmn (changes requested):

## Findings

- [P2] Preserve a terminal read error that races cancellation
  `internal/acp/jsonrpc.go:165`
  A real non-EOF `ReadBytes` failure can occur just as `ctx` is cancelled. If the watcher is scheduled first, it wins `endRead(true)` and sets `readInterrupted`; when the read loop subsequently handles the already-returned error, its `endRead(false)` is a no-op and line 175 returns `nil`. This silently reports a broken ACP transport as a clean shutdown, despite the stated terminal-error preservation. Record/read the terminal read outcome before allowing cancellation to claim the transport (and add a regression that races a non-EOF read failure with cancellation).

PierrunoYT:
Pushed `f5d7728`. This one turned out to need a real architectural fix, not a small ordering tweak — the race jatmn described has a **deterministic**, reproducible root cause, not just a rare goroutine-scheduling coincidence.

**[P2] A real non-EOF ReadBytes failure racing cancellation gets swallowed**

Root cause: `bufio.Reader` can obtain data **and** a terminal error from a single underlying `Read` call (`io.Reader` explicitly permits `(n>0, err)` together). If a line delimiter is found within that data, `ReadBytes` returns the line with a **nil** error for that call — but bufio quietly caches the error and returns it, **without ever touching the underlying reader again**, on the very next `ReadBytes` call. If cancellation happens in the *gap* between that first successful read and the next one (e.g. while a synchronously-dispatched handler is still writing its response — no goroutine needed, `handleLine`'s unsupported-jsonrpc-version path is fully synchronous), the old watcher's `ctx.Done()` case fires and claims the shared `Once` **before the read loop ever attempts another `ReadBytes` call**. When the cached error then surfaces — having had **zero** contact with the underlying reader, let alone anything closing it — it was misattributed as a cancellation-driven interrupt and silently swallowed as a clean shutdown. This is not a coincidence-of-timing bug; it's 100% reproducible given the right sequence, which is why I could build a deterministic (not `-count=100`-dependent) regression test for it.

**Fix:** move interruption-detection from "bracket the whole `Serve` loop with one watcher" down to "bracket each individual `Read` call." `interruptibleReader` wraps the connection's reader: each `Read` runs the underlying call on its own goroutine and races it against `ctx.Done()`.
- If the underlying call finishes first, its result is returned **untouched** — cancellation a moment later has no bearing on a call that already had its answer.
- If `ctx.Done()` fires first, the reader is closed (when closable) to nudge a genuinely blocked read, a **generation counter** is bumped, and the call still waits for and returns the underlying call's **real** result — never a synthesized one.

`Serve` then brackets each `ReadBytes` call (which may invoke the wrapped `Read` zero or more times — zero when bufio answers from its own buffer, including a previously-cached error) with a snapshot of that counter before/after. Unchanged ⇒ this call was never raced against cancellation (served from cache, or genuinely finished before `ctx.Done()` could matter) ⇒ report truthfully regardless of the current `ctx` state. Changed ⇒ this specific call actually took the cancellation branch ⇒ attribute to shutdown.

**I verified the fix has teeth rather than assuming the test would catch a regression**: reverted to the original implementation, confirmed the new test (`TestConnServePreservesReadErrorBufferedAlongsideAPriorLine`) fails **100% of the time** against it (not flaky — it has a `panic` guard proving the underlying reader is called exactly once, so there's no scheduling luck involved), then restored the fix and confirmed it passes.

**Also resolved the two open Copilot threads** while in this file: `readStarted`/`writeStarted` in the existing tests now use buffered channels with non-blocking sends instead of `close()`, so a future change adding a second read/write can't panic on a double-close.

**Validation:**
- `go test ./internal/acp/... -count=100` (matching the PR's own stress-testing approach) on Windows: pass.
- This host has no C toolchain (`-race` needs cgo), so I ran the race detector properly this time via WSL: `go test -race ./internal/acp/... -count=50` → pass, plus `go vet`/`go build` under WSL/Linux — clean.

@jatmn — ready for another look; this is a bigger structural change than the previous round (the interruption mechanism itself moved), so a fresh pass over `interruptibleReader` in particular would be worthwhile. I can't use the reviewer-request button on this repo, hence the mention.

@coderabbitai review

jatmn (changes requested):

## Findings

- [P2] Do not let cancellation win after the terminal read result is ready
  `internal/acp/jsonrpc.go:221`
  `select` chooses randomly when both `resultCh` and `ctx.Done()` are ready. If `inner.Read` has already completed with a non-EOF transport error and cancellation arrives before this goroutine consumes it, the cancellation arm can increment `gen` and then return that already-completed error. `Serve` consequently sees `interrupted` and returns `nil` at line 165, so a broken ACP transport is again reported as a clean shutdown. The new regression covers a buffered error surfaced by a later `ReadBytes`, but not this simultaneous-ready boundary. Establish/read completion before attributing the call to cancellation and add coverage for that ordering.

- [P2] Keep caller-owned input streams out of the cancellation close path
  `internal/acp/jsonrpc.go:103`
  `NewConn` accepts an `io.Reader`, but now infers ownership from an `io.Closer` assertion and closes it on context cancellation. A caller can pass a shared `*os.File`, `net.Conn`, pipe, or wrapper merely for reading; cancelling this `Conn` then permanently tears down that caller-owned transport (and potentially its writer/other consumers), whereas the prior constructor only borrowed it. Make the close/interrupt capability explicit through an owned-transport constructor or an interrupt callback instead of closing every reader that happens to implement `io.Closer`.

PierrunoYT:
Pushed `6dbe1d1`, addressing both P2 findings from the latest round.

**[P2] Cancellation winning after the terminal read result is ready** (`internal/acp/jsonrpc.go:221`)

`select` picks pseudo-randomly among ready cases, so if `inner.Read` had already completed (e.g. with a non-EOF transport error) by the time `ctx.Done()` also became ready, the cancellation arm could win, bump `gen`, and have `Serve` report a clean shutdown instead of the real error. Fixed by re-checking `resultCh` non-blockingly *before* committing to the cancellation branch: only a call whose result genuinely wasn't decided yet gets attributed to cancellation now.

Added `TestInterruptibleReaderPrefersDecidedResultOverCancellation`, which lets the underlying read complete (so `resultCh` already holds the outcome) before cancelling, across 100 iterations, asserting `generation()` never advances when the real, already-decided result is what's returned.

**[P2] Caller-owned input streams in the cancellation close path** (`internal/acp/jsonrpc.go:103`)

`NewConn` inferred ownership from an `io.Closer` type assertion and closed *any* reader that happened to implement it — including one a caller only lent for reading. Split the constructor:

- `NewConn(r, w)` — never closes `r`; a genuinely idle blocking read on a non-owned reader stays blocked past cancellation (existing tests already prove correctness here by closing their own pipe's write side, not relying on this).
- `NewOwnedConn(r, w)` — explicit ownership; closes `r` on cancellation if it's an `io.Closer`.

Switched the one production call site (`internal/cli/acp.go`, process stdin — genuinely and exclusively owned for the command's lifetime) to `NewOwnedConn`. Updated the two tests whose assertions specifically depend on close-on-cancel behavior to use it too.

Also folded in CodeRabbit's two trivial nits from this round: `atomic.Int64` instead of the mutex-guarded counter, and `ctx` moved to the leading constructor parameter.

**Validation:**
- `go test ./internal/acp/... -run 'TestInterruptibleReaderPrefersDecidedResultOverCancellation|TestConnServePreservesTerminalReadErrorDuringCancellation|TestConnServePreservesReadErrorBufferedAlongsideAPriorLine|TestConnServeCancellationInterruptsIdleRead' -count=50`: pass.
- `go test -race ./internal/acp/... -count=50` via WSL: pass, plus `go build`/`go vet ./...` under WSL — clean.

jatmn (changes requested):

## Findings

- [P2] Preserve read errors across the result-publication race
  `internal/acp/jsonrpc.go:225`
  `resultCh` being readable is not proof that `inner.Read` did not already finish first: after `inner.Read` returns, its goroutine may be descheduled before it executes `resultCh <- ...`. If cancellation happens in that scheduling window, the cancellation branch sees an empty channel, increments `gen`, and then receives the real non-EOF transport error. `Serve` consequently classifies the read as interrupted and returns `nil`, so `zero acp` exits successfully while hiding a real transport failure. This is a race rather than the deterministic buffered-error path covered by the new test, but it violates the stated terminal-error preservation guarantee. Establish completion and cancellation ownership with a synchronized handoff, and add coverage for the return-to-publication window.

PierrunoYT:
Addressed the latest P2 in `c981ed5`.

- Added a `sync.Once` completion/cancellation handoff before result publication, so cancellation cannot claim a read after `inner.Read` has established completion but before its goroutine sends on `resultCh`.
- Added deterministic coverage for that completion-before-publication interval.
- Applied the remaining CodeRabbit test nit by replacing the goroutine panic guard with the existing atomic call-count assertion.
- Resolved the two stale Copilot threads after verifying the current buffered, non-blocking signals satisfy them.

Validation:
- focused ACP cancellation/error regressions under `-race`, `-count=100`
- `go test ./internal/cli -count=1`
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `govulncheck` — no vulnerabilities found
- `git diff HEAD --check`

The advisory golangci-lint command still reports 34 unrelated repository-baseline findings; none are in the PR files. All GitHub checks on `c981ed5` pass across Linux, macOS, and Windows.

PierrunoYT:
Final follow-up on current head `c981ed5`:

- synchronized read completion with cancellation so a completed read cannot be reclassified as interrupted
- added deterministic race coverage and removed the worker-goroutine panic path
- all review threads are resolved
- focused `-race` tests, full tests, build, smoke, vet, formatting, and vulnerability checks pass
- all GitHub checks are green and the PR is mergeable

PierrunoYT requests a fresh human review to supersede the previous `CHANGES_REQUESTED` decision.

jatmn (changes requested):

The prior `CHANGES_REQUESTED` review on the result-publication race is not fully resolved on `c981ed5`. The other reported issues look addressed: per-`Read` interruption, buffered `bufio` error preservation, decided-result preference, explicit `NewOwnedConn` ownership, and idle-read cancellation for owned stdin.

## Findings

- [P2] Record read completion before `interrupt()` can claim a finished call
  `internal/acp/jsonrpc.go:241`
  `decision.complete()` still runs only after `inner.Read` returns in the helper goroutine. If `ctx` is already cancelled, the `select` can take `<-r.ctx.Done()` and `decision.interrupt()` can win the shared `sync.Once` in the scheduling window after `inner.Read` has returned but before `complete()` executes. That bumps `gen`, may close an owned reader, and makes `Serve` treat the following non-EOF `ReadBytes` error as interrupted (`return nil`), so a genuine transport failure is again reported as clean shutdown. This is the same guarantee violation as the prior review, just narrowed from the result-publication gap to the `Read`→`complete()` gap. Record completion synchronously with read completion (for example via `defer decision.complete()` inside a small wrapper `Read` so completion runs in the read function epilogue before control returns to the outer goroutine), and add integrated coverage for cancellation landing in that window rather than only the isolated `interruptibleReadDecision` unit test.

- [P3] Do not leave `interruptibleReader.Read` blocked forever on a panicking reader
  `internal/acp/jsonrpc.go:241`
  Each `Read` now runs `inner.Read` on its own goroutine and always waits on `resultCh`. If `inner.Read` panics, nothing is sent on `resultCh` and the caller blocks forever in either `select` branch, so `Serve` can hang instead of surfacing failure. Recover in the helper goroutine, publish a terminal error on `resultCh`, and still record completion so cancellation arbitration remains well-defined.

PierrunoYT:
PierrunoYT pushed `30d5498` plus review follow-up `f7cdb74`, after merging current `upstream/main` normally in `ce8699b`.

Review findings addressed:

- **[P2] Read completion handoff:** the underlying `Read` now runs through `readInner`, which records completion in a deferred read epilogue before the helper publishes its result. The regression exercises that real wrapper with publication deliberately blocked and verifies cancellation cannot claim the completed read.
- **[P3] Panicking reader:** the same deferred epilogue recovers reader panics into terminal ACP read errors, still records completion, and guarantees the helper publishes a result. Added an end-to-end `Conn.Serve` regression proving `panic("boom")` returns `acp: reader panicked: boom` instead of losing the read result.
- **Post-push CodeRabbit follow-up:** bounded the buffered-error test's `writeStarted` wait and registered a `sync.Once` cleanup release so a synchronization regression fails promptly without stranding the writer.

The earlier CodeRabbit goroutine-per-`Read` performance nit was reviewed and intentionally left unchanged: racing context cancellation against a plain blocking `io.Reader` requires the helper goroutine, and returning before joining it would reintroduce the write race on `p`. The latest human review identified only the two correctness findings above as outstanding.

Validation on Go 1.26.5 / Linux (rerun after `f7cdb74`):

- focused ACP cancellation/error regressions under `-race`, `-count=100`
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make vulncheck` — no vulnerabilities found
- `git diff HEAD --check`

jatmn (changes requested):
## Findings

- [P2] Wire `runACP` to the new `Serve` error contract
  `internal/cli/acp.go:80-82`
  On base `main`, `Serve` returned `nil` whenever `ctx.Err() != nil`, so `runACP`'s `err != nil && ctx.Err() == nil` guard never mattered for shutdown-overlapping read failures — both paths exited 0. Head `Serve` now correctly returns genuine non-EOF transport failures when `interrupted` is false even if the context is already cancelled. `TestConnServePreservesReadErrorBufferedAlongsideAPriorLine` and `TestConnServePreservesTerminalReadErrorDuringCancellation` both cancel `ctx` and expect `Serve` to return `wantErr`; `runACP` would still take the success path because `ctx.Err() != nil`. Clean idle SIGINT is unaffected (`interrupted == true` still yields `nil` from `Serve`), but a real stdin/transport failure that surfaces during signal shutdown is now reported by the library and silently dropped by the CLI: `zero acp` exits 0 with no stderr while `Conn.Serve` returned a terminal read error. Please treat non-nil `Serve` errors as crash-worthy regardless of `ctx.Err()` — interrupted shutdown already comes back as `nil` — or add an explicit CLI/integration test that exercises this boundary.

PierrunoYT:
PierrunoYT addressed the latest review in `7791cf3`.

- `runACP` now reports every non-nil `Serve` error, including genuine transport failures that overlap cancellation.
- Clean idle SIGINT still exits successfully and silently.
- Added focused cancellation/error and clean-interrupt regressions.

Validation: focused `internal/acp` and `internal/cli` tests pass with `-race`; formatting and diff checks pass. Ready for re-review.

PierrunoYT:
Follow-up `a6e2530` fixes the Windows-only CI failure in the new ACP regressions. The tests now inject a cancellation context instead of sending `os.Interrupt` to the current process, which Windows does not support; production still uses the real signal context. Focused race tests pass repeatedly and the CLI test binary cross-compiles for Windows. Fresh CI has been triggered.

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):
lgtm

PR #796, opened 2026-07-23
fix(hooks): dispatch specialistStart/Stop lifecycle hooks

### LINKED ISSUE #793: fix(hooks): dispatch specialistStart/Stop lifecycle hooks
## Summary

`specialistStart` / `specialistStop` are documented in EXTENDING.md, accepted by `zero hooks add`, and present in the hooks package — but the specialist runtime never dispatches them. Matcher support is also inconsistent: docs say specialist matchers are allowed, while LoadConfig rejected them.

## Proposed fix

- Dispatch advisory specialistStart/Stop hooks from specialist accounting
- Allow specialist-name matchers (matching EXTENDING.md)
- Wire the hooks dispatcher into specialist tools after startup builds it

## Related work

Implementation ready on branch `feat/specialist-lifecycle-hooks`.

## Notes for maintainers

Please add `issue-approved` if this scope is accepted so the linked community PR can proceed under CONTRIBUTING.md.

Ayush7614:
## Summary

- Dispatch advisory `specialistStart` / `specialistStop` hooks from specialist accounting
- Allow specialist-name matchers (docs already claimed this; LoadConfig previously rejected it)
- Attach the hooks dispatcher to specialist tools after startup builds it

These events were documented and accepted by `zero hooks add`, but never fired at runtime.

Closes #793

## Test plan

- [x] `go test ./internal/specialist/ ./internal/hooks/ -count=1`
- [x] `go vet ./internal/specialist/ ./internal/hooks/ ./internal/cli/`
- [x] `govulncheck` on touched packages — no vulnerabilities

## Summary by CodeRabbit

- **New Features**
  - Added support for matching hooks on specialist start and stop events, including updated matcher validation.
  - Specialist lifecycle events now dispatch configured hooks in both interactive and command-line runs, and specialist start/stop activity is recorded in the audit log.
- **Documentation**
  - Clarified matcher semantics (subject is the agent type) and noted that specialist/session hook failures are audited without interrupting runs.
  - Updated help text to reflect tool/specialist matcher support for the relevant hook events.
- **Tests**
  - Expanded coverage for specialist hook dispatching, payload/event recording, and stop-event deduplication behavior.


kevincodex1:
@coderabbitai make an intensive review on this

Ayush7614:
Addressed CodeRabbit review in 19f98c2 (also synced with latest `main`):

- `recordSpecialistStop` releases `accountingMu` before running lifecycle hooks
- `specialistStop` hooks fire only when `appendSpecialistEventOnce` newly records the stop (deduped under concurrency)
- Matcher acceptance test now covers both `specialistStart` and `specialistStop`
- Concurrency test asserts exactly one stop hook under race

Ayush7614:
cc: @kevincodex1

Vasanthdev2004 (changes requested):
Both of the earlier findings are properly fixed here: the dedup result is honoured now, and accountingMu brackets only the append so a slow hook cannot stall other specialist accounting. The supporting checks all pass too, the hooks are genuinely advisory (blocksOn is beforeTool only and the outcome is discarded), the specialist-name matcher now matches what the docs table already advertised, and runWithTimeout bounds a hung hook at 30s.

The problem is that fixing (1) by gating on `appended` introduced the inverse defect, and it is deterministic rather than a race.

**[blocker] Swarm members fire specialistStart and never specialistStop, 100% of the time.**

internal/swarm/launcher_specialist.go builds TaskRunOptions with ParentModel, Cwd, PermissionMode and MemberAutonomy but never sets ParentSessionID. appendSpecialistEventOnce short-circuits on an empty parent session and returns false, so `if appended` suppresses the stop dispatch for every swarm member, while recordSpecialistStart dispatches unconditionally with no gate. Anyone pairing the two events (a notification, a timer, resource cleanup) leaks state on every member of every swarm run.

**[major] Background specialists can drop the stop hook too.**

OutputTool.readOutput reaps through a fresh `Executor{SessionStore: ...}` literal with no LifecycleHooks, and whichever executor wins the appendSpecialistEventOnce race owns the only dispatch. When the hook-less TaskOutput poll wins, the hook-carrying onExit path then sees appended==false and suppresses the hook. Deterministic variant: if the parent process exits before the background child, the onExit goroutine dies with it and a later TaskOutput is the only reaper, so the hook can never fire.

**[minor]** `appended, _ :=` also discards the error, and AppendEventUnlessExists returns false on failure as well as on already-recorded, so a session-store error silently becomes a suppressed hook with no signal anywhere.

All three are the same root cause: dispatch is coupled to session-event bookkeeping that is skipped when there is no parent session, performed by a different executor, or failed. I would dispatch when the stop is first observed by THIS executor, decoupled from whether the audit event was appended, keeping the real duplicate suppression, and pass LifecycleHooks into OutputTool so the background reaper can dispatch.

Two smaller notes: for swarm members the matcher subject is the agent type (subagent/teammate) rather than a registered specialist name, so a matcher written against real specialist names will not match once stop hooks do fire, which deserves a docs line. And no test covers the CLI wire-up at all, so deleting either attachLifecycleHooks call would still pass the suite, which is exactly the class of bug #793 reports; the new concurrency test races two hook-carrying executors, so by construction it cannot catch either failure above.

The dedup fix itself is right, it just needs to not be the thing that decides whether the hook fires. Happy to re-review quickly.

Ayush7614:
Addressed @Vasanthdev2004's review on `19f98c2`:

1. **Swarm `specialistStop` suppressed** — stop-hook dispatch is no longer gated on session-event append. `LifecycleHooks.claimStopDispatch` dedupes by `childSessionID+runID`, so swarm members with an empty parent still fire exactly once.
2. **TaskOutput race** — `OutputTool` now carries the shared `LifecycleHooks` bridge from `RegisterTools`, so a hook-less reap path cannot permanently suppress the stop hook.
3. **Docs** — noted in `docs/EXTENDING.md` that swarm matcher subjects are agent types (`subagent`/`teammate`), not registered specialist names.
4. **Tests** — empty parent still fires stop; hook-less append winner then hook-carrying second still fires once; OutputTool shares the bridge; CLI `attachLifecycleHooks` wire-up covers TaskOutput.

Ready for re-review. Thanks!

Vasanthdev2004:
Both things I named are genuinely fixed, and well covered. Neutering `claimStopDispatch` fails five tests, including `TestRecordSpecialistStopFiresWithoutParentSession`, which is exactly the swarm case I blocked on, and `TestOutputToolSharesLifecycleHooksWithStopAccounting` for the background reap. Decoupling the hook from the session-event append was the right call.

Two things stop me clearing it, and the first is new in the latest commit.

**`specialistStop` now fires twice for one specialist.** `claimStopDispatch` keys on `childSessionID + "\x00" + runID` and does an exact-match `LoadOrStore`. The dedup it replaced deliberately treated an already-recorded event with an EMPTY runId as a catch-all for a later one that carries a runId, and said why in a comment: "otherwise the same child gets two stop/usage events". The new key drops that rule. Driven straight at the function:

```
empty-then-runid : first=true second=true     <- two dispatches
runid-then-empty : first=true second=true     <- two dispatches
same-runid-twice : first=true second=false    <- the only case it catches
```

It is reachable on the background path, not just in theory. On a `SetPID` failure the executor records a stop with `StreamResult{ExitCode: -1}` and no runId while the process is already running, and the later `onExit` reap records again from a summary that does carry one. The comment right above that branch says "The dedup in `recordSpecialistStop` makes the later onExit accounting a no-op" — still true of the accounting, no longer true of the hook, which is now a separate mechanism.

Same shape as the bug this PR fixes, one level along: two representations of one dedup rule with no contract between them. `TestRecordSpecialistStopDedupesUnderConcurrency` does not catch it because it exercises the same key from several goroutines, and the failure needs two different keys for one child.

**The CLI wire-up is still untested, which was my ask.** I removed both `attachLifecycleHooks` call sites, the interactive one and the headless one, so `Dispatch` stays nil forever and no specialist hook can fire in either mode:

```
baseline: --- FAIL: TestBuildServeScopeKeepsLexicalPaths   (known, symlink privilege)
mutated:  --- FAIL: TestBuildServeScopeKeepsLexicalPaths   (identical)
internal/specialist ok, internal/hooks ok, in both runs
```

Byte-identical. The whole feature can be deleted and the suite stays green, which is the class of bug #793 reports. The new test calls `attachLifecycleHooks` itself, so it pins the method rather than the startup that calls it. One test that drives the CLI entry point and asserts a specialist hook fires would close both this and the equivalent gap on the swarm side.

Neither is a big change. The dedup wants the same empty-runId catch-all the accounting rule has, ideally by sharing one predicate so they cannot drift again.

Branch merges cleanly and CI is green on the head.

Vasanthdev2004 (approved):
Approving on `7d55f568`. Sorry this sat on a stale change request; the blocker is fixed and the fix is the right shape.

**The blocker is gone.** Dispatch is claimed on the `LifecycleHooks` bridge rather than on whether the session event appended, which decouples it from exactly the three things that made it fail. I checked the semantics directly rather than reading them:

```
child session id present:  first claim true, second false     (dedup holds)
no child session id:       claim true                          (swarm members fire)
same child, different run: true                                (runs stay distinct)
nil bridge / nil dispatch: false                               (safe)
```

The empty-child branch returning true unconditionally is what makes swarm members work, since they are the case with no session to key on.

**The background reaper is fixed too.** `OutputTool` now carries the bridge and passes it into the `Executor` literal in `readOutput`, so the onExit and TaskOutput-poll finishers share dedup instead of one silently claiming the append and suppressing the other.

**The minor one is resolved by construction.** The error from `appendSpecialistEventOnce` is still discarded, but it can no longer suppress a hook, which was the whole complaint. Fine as is.

One concern of mine that turned out to be nothing, so you do not have to re-derive it: I went looking for double dispatch on the empty-child branch, since it never dedups. It is unreachable. The only path with two finishers for one stop is the background one (onExit plus the poll), and that always keys on `task.ID`, which is non-empty. The foreground paths at `exec.go:516/533/613/621` are alternative branches of a single call, so only one can fire.

**One thing left, and it is the half of my test note that is still open.** The new `TestRegisterSpecialistToolsWiresLifecycleHooksIntoTaskOutput` does catch the wiring it is named for; dropping `LifecycleHooks` from the `OutputTool` constructor fails it. But deleting `specialistRuntime.attachLifecycleHooks(hookDispatcher)` from `app.go:810` still passes the whole `internal/cli` suite, and `exec.go:628` is the same. Those two lines are what connect the dispatcher in the real entry points, so either can be removed and every hook silently stops firing with CI green.

Not blocking, since the code is correct today, but it is worth a test that goes through the app path and asserts `lifecycleHooks.Dispatch` is non-nil after startup. That is the same class of gap as the original #793 report.

`internal/specialist` and `internal/hooks` pass in full. `internal/cli` has two failures here, `TestBuildServeScopeKeepsLexicalPaths` and `TestRunNoArgsLaunchesSetupTUIWithNilProviderWhenNoProviderConfigured`, and both reproduce identically on `origin/main`, so neither is yours.

jatmn (changes requested):

## Findings

- [P1] Make the stop-hook claim durable and use the accounting run-ID equivalence
  `internal/specialist/exec.go:83`
  The new stop side effect is deduplicated by `LifecycleHooks.stopOnce`, a per-process `sync.Map` keyed by the exact `(childSessionID, runID)`. That is a different identity and durability boundary from the existing accounting implementation: background task metadata and `AppendEventUnlessExists` are intentionally persistent and cross-process safe, and `specialistEventMatcher` treats an empty run ID as a catch-all for the same child.

  This creates two observable duplicate paths. First, the launching process can dispatch from its background `onExit` callback, then a later Zero process can restore the completed task and call `TaskOutput`; the later process has a fresh bridge and therefore runs the same `specialistStop` command again even though the session stop event is rejected as a duplicate. Second, if `SetPID` fails after the child begins, the immediate error accounting has no run ID; an on-exit reap that later finds the child's `run_start` ID uses a different map key and dispatches again, while durable accounting correctly treats both records as one stop.

  The root cause is that the PR now has two independently maintained definitions of “the same specialist stop”: a durable, empty-run-ID-aware accounting predicate and a local exact-key hook predicate. Hooks can perform notifications, cleanup, or external state changes, so at-most-once behavior needs to cover the hook itself, not merely the audit event. Please establish one shared, durable stop-claim contract for the external hook that follows the accounting equivalence rules, including recovery, empty run IDs, TaskOutput, and swarm behavior. Keep lifecycle hooks advisory and do not regress the hook-less-reaper recovery case this PR was addressing.

- [P2] Emit `specialistStop` when a recovered task is explicitly stopped
  `internal/specialist/stop_tool.go:65`
  `TaskStop` only calls `manager.Kill`; it has no lifecycle bridge and does not invoke specialist accounting. In the ordinary same-process case, the launcher's in-memory `onExit` callback may subsequently produce the stop effect. That assumption is not valid after the launcher has exited: a later Zero process can restore the persisted task, successfully terminate it through `TaskStop`, and return `status: killed` without any `specialistStop` hook. The only remaining route is a later `TaskOutput` poll, which is neither required nor reliable—and currently has the replay problem above.

  The root cause is that lifecycle dispatch is attached to observers of task completion rather than to a single durable lifecycle transition. “A sub-agent ends” includes explicit recovered-task termination, not only normal launcher-owned exit. Please make the recovered `TaskStop` path participate in the same single stop-transition/claim mechanism as normal completion. Preserve `TaskStop`'s existing prompt permission, kill semantics, and status reporting; the required outcome is one advisory stop hook for a successful terminal transition, not a second independent dispatch path.

- [P2] Do not publish a start hook until the child has actually started
  `internal/specialist/exec.go:608`
  Both foreground and background paths call `recordSpecialistStart` before `runChild`/`launchBackground` reports success. In the foreground implementation, `command.Start()` can fail outright; the injected execution seam also permits that failure. In either mode, a missing binary, failed process launch, or other startup error therefore invokes `specialistStart` and then records a stop, despite no sub-agent ever existing. This is externally visible to configured hooks even though the pre-existing accounting record itself was internal.

  The root cause is that one pre-launch accounting operation is now also the public lifecycle-event producer. The documented contract says `specialistStart` fires when a sub-agent is spawned, whereas the current call represents an attempt to launch one. Please separate those concepts: retain any accounting needed to diagnose a failed attempt, but dispatch the public start hook only after child startup is confirmed (or define a clearly distinct attempt event if that behavior is wanted). The corresponding stop behavior should continue to describe an actual terminal lifecycle transition rather than compensate for a never-started child.

Vasanthdev2004:
@Ayush7614 checking in on this one. I approved it at `7d55f568` and it has not moved since 2026-08-28. It is still mergeable against current main and its checks are green, so as far as I can tell it is only waiting on a second maintainer rather than on you.

Saying so explicitly in case it looked like the ball was in your court. If you are still around for it, nothing is needed from you; I will chase the merge side. If you have moved on, tell me and I will note the issue as open for someone else rather than leave it looking claimed.

Same question on #693, which is in the same state.

PR #798, opened 2026-07-24
fix(sandbox): clarify scoped permission prompts

anandh8x:
## Summary

- distinguish read and write explanations for outside-workspace permission prompts
- show the focused permission reason once while retaining structured block and path context
- wrap long permission reasons instead of truncating them

## Why

Read-only tools were shown write-oriented `/add-dir` guidance, and identical policy reasons were repeated around the permission modal and denial history. This made routine permission requests noisy and misleading.

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- manual TUI check of read-only `list_directory` approval and denial

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved permission messages when reading or writing outside the workspace.
  - Write restrictions now explain how to allow access using `/add-dir` or `--add-dir`.
  - Prevented duplicate permission reasons from appearing in the interface.

- **User Interface**
  - Permission prompts now wrap long explanations across multiple lines instead of truncating them.
  - Permission request details display more consistently and clearly.

### DIFF
--- a/internal/sandbox/engine_test.go
+++ b/internal/sandbox/engine_test.go
@@ -2,6 +2,7 @@ package sandbox
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -482,6 +483,28 @@ func TestEngineDeniesOutOfWorkspacePaths(t *testing.T) {
 	}
 	if !strings.Contains(decision.Reason, "outside the workspace") {
 		t.Fatalf("expected outside-workspace reason, got %q", decision.Reason)
+	}
+}
+
+func TestEnginePromptsForOutOfWorkspaceReadWithReadSpecificReason(t *testing.T) {
+	root := t.TempDir()
+	outside := outsideDefaultTempPath(root, "external")
+	engine := NewEngine(EngineOptions{WorkspaceRoot: root, Policy: DefaultPolicy()})
+
+	decision := engine.Evaluate(context.Background(), Request{
+		ToolName:      "list_directory",
+		SideEffect:    SideEffectRead,
+		Permission:    PermissionAllow,
+		WorkspaceRoot: root,
+		Args:          map[string]any{"path": outside},
+	})
+
+	want := fmt.Sprintf("Reading %s requires access outside the workspace.", outside)
+	if decision.Action != ActionPrompt || decision.Block == nil || !decision.Block.Recoverable {
+		t.Fatalf("outside read decision = %#v, want recoverable prompt", decision)
+	}
+	if decision.Reason != want || decision.Block.Reason != want {
+		t.Fatalf("outside read reasons = decision %q, block %q; want %q", decision.Reason, decision.Block.Reason, want)
 	}
 }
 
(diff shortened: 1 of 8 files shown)


kevincodex1 (approved):
LGTM

gnanam1990 (approved):
## Review: PR #798 — fix(sandbox): clarify scoped permission prompts

The core change does what it says. Read vs. write out-of-workspace blocks now get distinct, specific reason strings (`scope.go` `withOutsideWorkspaceAccessReason`), the transcript row no longer echoes the reason already shown in the focused modal (`transcript.go` `permissionDetailText`/`permissionBlockDetail` + `model.go`), and the focused card wraps the reason instead of truncating it (`rendering.go` `wrapPlainText`). The added tests assert exactly these behaviors and pass. Nice, tightly-scoped fix with matching coverage.

### [Minor] Resolved out-of-workspace deny/cancel rows drop the `/add-dir` remediation hint (PR-introduced)
`internal/tui/transcript.go:401` — `permissionDetailText` only appends `event.Reason` when `event.Action == agent.PermissionActionPrompt`. For a resolved **deny** or **cancel** event, the reason (which carries the actionable `…use /add-dir or --add-dir to allow writes there` hint) is therefore not added to the wrapped detail block. It also can't fall through to `permissionBlockDetail` (`transcript.go:445`), which drops the block reason when it equals the event reason. So the hint survives only on the single header line — and at width 80 that line truncates (`…scope:/home/dev/projects/service/config — Writing to /hom…`), cutting the hint off entirely.

Concrete repro: user is prompted for a write to `/home/dev/projects/service/config`, denies it. On this branch the rendered deny row shows `denied by user  blocked: outside workspace  path: /home/dev/projects/service/config` with the `/add-dir` remediation nowhere on screen. On base, the same deny event renders the full hint in the wrapped detail block. Verified by rendering identical events on both worktrees.

Fix: for non-prompt actions, when the header-line reason is truncated, keep the full reason (or at least the `/add-dir` hint) in the wrapped detail block — e.g. don't gate `event.Reason` out of `permissionDetailText` for deny/cancel, or append it to `permissionBlockDetail` when it isn't already fully shown on the header line.

### Build / tests
- `gofmt -l`, `go build`, `go vet` on `internal/sandbox` and `internal/tui`: all clean.
- All PR-added tests pass on the PR HEAD: `TestEnginePromptsForOutOfWorkspaceReadWithReadSpecificReason`, `TestScopeValidateReadExplainsReadAccessOutsideWorkspace`, `TestScopeValidateWriteExplainsWriteAccessOutsideWorkspace`, `TestPermissionRequestShowsFocusedPrompt`, `TestPermissionRowRendersIdenticalEventAndBlockReasonOnce`, `TestPermissionPromptWrapsLongReasonWithoutTruncating`.
- The failures that do appear (`internal/sandbox`: `TestBuildCommandPlanRejectsOutsideDirectory`, `TestScopeAddNormalizesSymlinkedRoot`, `TestEvaluateAppliesWriteAllow`, etc.; `internal/tui`: `TestHandleAddDirCommand`) are **pre-existing / environment-caused** — they fail byte-identically on the clean base worktree. Root cause is the sandbox running under `/private/tmp` and `/private/var/folders`, which are default temp roots in scope, so paths the tests intend as "outside workspace" resolve as inside. Not attributable to this PR; the PR introduces zero new failures.

Verdict: APPROVE. The minor deny/cancel hint regression is a reasonable follow-up but is not merge-blocking.

PR #799, opened 2026-07-24
fix(tui): reset quiet clock for new runs

anandh8x:
## Summary

- initialize stream activity from the new turn's start time
- prevent fresh turns from inheriting a stale quiet-generation duration
- add a deterministic regression test for the stale-clock behavior

## Why

`beginRun()` reset the turn clock but retained `lastStreamActivity` from an earlier turn or window event. Before the first new stream delta arrived, a fresh request could therefore immediately display a misleading message such as `still generating… 1m39s`.

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- manual TUI check through a real PTY with `termctrl`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale “quiet generation” warnings appearing when a new agent turn begins.
  * Quiet-generation timing now resets correctly at the start of each run.

* **Tests**
  * Added coverage to verify quiet-generation warnings are cleared for new runs.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -4807,6 +4807,7 @@ func (m model) beginRun(cancel context.CancelFunc) model {
 	// suppressed by a stale preference.
 	m.sidebarHidden = false
 	m.turnStartedAt = m.now()
+	m.lastStreamActivity = m.turnStartedAt
 	m.turnStreamedRunes = 0
 	m.spinnerTicking = true
 	return m
--- a/internal/tui/working_status_test.go
+++ b/internal/tui/working_status_test.go
@@ -106,6 +106,20 @@ func TestQuietGenerationHint(t *testing.T) {
 	}
 }
 
+func TestBeginRunResetsQuietGenerationClock(t *testing.T) {
+	now := time.Date(2026, 7, 24, 12, 0, 0, 0, time.UTC)
+	m := model{
+		now:                func() time.Time { return now },
+		lastStreamActivity: now.Add(-time.Minute),
+	}
+
+	m = m.beginRun(nil)
+
+	if got := m.quietGenerationHint(); got != "" {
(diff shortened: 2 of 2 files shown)


kevincodex1 (approved):

gnanam1990 (approved):
**Approve** — a correct, minimal one-line fix with a test that genuinely exercises the bug.

**What it does**
`beginRun` (model.go:4810) now sets `m.lastStreamActivity = m.turnStartedAt` alongside the existing `m.turnStartedAt = m.now()`. Previously a fresh run inherited the previous turn's `lastStreamActivity`, so `quietGenerationHint()` (model.go:3547) computed `quiet = now - staleActivity` off that carried-over timestamp and mis-fired a "still generating…" warning the instant a new run began.

**Verification (PR-introduced)**
- Confirmed the fix addresses the stated bug, not just describes it: deleting the new line and running the added test reproduces the exact failure — `working_status_test.go:119: new run inherited a stale quiet-generation warning: "still generating… 1m00s"`. `quietWorkingHint = 8s` (model.go:3533), so a 1-minute stale gap trips the hint; resetting the clock to `turnStartedAt` makes `quiet = 0` and suppresses it. Worktree restored after the experiment.
- The new test `TestBeginRunResetsQuietGenerationClock` is valid: `beginRun` does `m.runID++; m.activeRunID = m.runID`, so `activeRunID` becomes 1 and the `activeRunID == 0` early-return in `quietGenerationHint` does not mask the assertion.
- `gofmt -l`, `go build`, `go vet`, and `go test -race ./internal/tui` all clean for the changed files. The only failing test in the package, `TestHandleAddDirCommand` (add_dir_test.go:43), fails **identically on clean main** — a sandbox artifact from tests running under `/private/tmp` (already a write root), unrelated to this diff. Not PR-attributable.

**Nits**
- None.

Nicely scoped: one production line, one focused regression test that fails without the fix.

PR #800, opened 2026-07-24
test(config): remove Windows startup timing assumptions

anandh8x:
## Summary

- remove cold PowerShell startup assumptions from provider-command timeout tests
- bound background-child completion by the child's natural lifetime instead of an arbitrary four seconds
- retain strict PID and liveness checks for synchronized background children

## Why

Windows smoke runs repeatedly failed even though provider-command cleanup was working:

- the basic timeout fixture could be terminated before cold PowerShell startup recorded its PID
- background-child setup could consume more than three seconds under concurrent CI load, making the one-second `WaitDelay` path exceed a fixed four-second assertion

The actual safety contract is that execution returns before the 10-second child finishes naturally and that every child known to have started is dead afterward. These tests now assert that behavior directly.

## Verification

- repeated focused provider-command tests
- Windows/amd64 test-binary cross-compilation
- `make fmt-check`
- `go vet ./...`
- `go test ./...`

## Summary by CodeRabbit

* **Tests**
  * Improved command timeout and background-process termination test reliability.
  * Added platform-tolerant handling for processes that exit before recording their process ID.
  * Strengthened timing checks to verify commands return before background processes naturally finish.

### DIFF
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):

anandh8x:
@kevincodex1 could you please merge this when ready? The same Windows provider-command timing failure is repeatedly blocking unrelated PRs, including #801, even though their Windows-specific changes pass. Landing #800 will stop this known test flake from affecting other PRs.

kevincodex1:
good to merge

PR #802, opened 2026-07-24
test(config): wait for background PID file

anandh8x:
## Summary
- wait briefly for a background provider command PID file to become readable before checking process termination
- preserve the strict child-liveness assertion once the PID is available
- add a deterministic regression covering delayed PID-file visibility

## Why
Windows Smoke can start and terminate the provider command correctly while the PID file is not yet readable when the test assertion runs. #800 handled the basic timeout fixture, but the strict background-child fixtures still performed an immediate read and remained flaky.

## Validation
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test ./internal/config -run TestAssertProcessTerminatedWaitsForDelayedPIDFile -count=100`
- Windows cross-compilation of `./internal/config` tests
- `govulncheck ./...` (no vulnerabilities found)
- `git diff HEAD --check`
- pinned advisory golangci-lint (pre-existing repository findings only; none in the changed file)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process termination checks to correctly handle PID files that are initially missing and appear shortly after monitoring begins.
  * Added a short wait window for PID file availability, reducing false failures during delayed shutdown.

* **Tests**
  * Added coverage for the delayed PID-file scenario to ensure the new waiting behavior is exercised.

### DIFF
--- a/internal/config/command_test.go
+++ b/internal/config/command_test.go
@@ -153,13 +153,26 @@ func TestLoadProviderCommandTerminatesBackgroundChildOnFailure(t *testing.T) {
 
 func TestAssertProcessTerminatedWaitsForDelayedPIDFile(t *testing.T) {
 	pidFile := filepath.Join(t.TempDir(), "delayed.pid")
+	firstMissing := make(chan struct{})
 	writeDone := make(chan error, 1)
 	go func() {
-		time.Sleep(50 * time.Millisecond)
-		writeDone <- os.WriteFile(pidFile, []byte("2147483647\n"), 0o600)
+		<-firstMissing
+		tempFile := pidFile + ".tmp"
+		if err := os.WriteFile(tempFile, []byte("2147483647\n"), 0o600); err != nil {
+			writeDone <- err
+			return
+		}
+		writeDone <- os.Rename(tempFile, pidFile)
 	}()
 
-	assertProcessTerminated(t, pidFile)
+	missingObserved := false
+	assertProcessTerminatedAfterMissing(t, pidFile, func() {
+		missingObserved = true
+		close(firstMissing)
+	})
+	if !missingObserved {
+		t.Fatal("PID file existed before the missing-file polling path was exercised")
+	}
 	if err := <-writeDone; err != nil {
 		t.Fatalf("write delayed pid file: %v", err)
 	}
@@ -181,9 +194,16 @@ func assertProcessTerminatedIfStarted(t *testing.T, pidFile string) {
 func assertProcessTerminated(t *testing.T, pidFile string) {
 	t.Helper()
 
+	assertProcessTerminatedAfterMissing(t, pidFile, nil)
+}
+
+func assertProcessTerminatedAfterMissing(t *testing.T, pidFile string, onMissing func()) {
+	t.Helper()
+
 	deadline := time.Now().Add(3 * time.Second)
 	var data []byte
 	var err error
+	missingObserved := false
 	for time.Now().Before(deadline) {
 		data, err = os.ReadFile(pidFile)
 		if err == nil {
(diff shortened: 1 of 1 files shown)


anandh8x:
@kevincodex1 This is the focused follow-up for the remaining Windows background-child PID-file race exposed after #800 merged. Could you please review and merge it when the checks are green? #801 can then be updated onto main without carrying this unrelated test fix.

anandh8x:
@coderabbitai review

Addressed both requested changes in ef745b35:
- replaced the fixed-delay regression with a deterministic first-missing-file barrier and asserted that path was exercised
- published the test PID file atomically via a temporary file plus rename, preventing partial reads

Focused stress test passes 100/100, the full suite passes with required host permissions, and Windows test cross-compilation passes.

anandh8x:
@kevincodex1 Follow-up: both CodeRabbit findings are addressed, CodeRabbit has approved the latest commit, and the full CI matrix—including Windows Smoke—is green. This is ready for maintainer review/merge.

gnanam1990 (approved):
## Summary

Test-only change in `internal/config/command_test.go`. It refactors `assertProcessTerminated` to poll for the background child's PID file (up to a 3s deadline) rather than doing a single `os.ReadFile`, and extracts `assertProcessTerminatedAfterMissing(pidFile, onMissing)` so a new test can hook the missing-file path. This fixes a real ordering bug: if the child hadn't yet written its PID file when the assertion ran, the old code failed immediately with "read sleeper pid file". PR-introduced (base read the file exactly once — confirmed via the diff against `abase`).

## Correctness

- The stated bug is actually addressed: the read is now retried on `os.ErrNotExist` until the deadline, with non-NotExist errors still failing fast (`command_test.go:207-215`). This is the fix the title claims, not just a description of it.
- The new `TestAssertProcessTerminatedWaitsForDelayedPIDFile` genuinely exercises the delayed-write path rather than racing past it: the writer goroutine blocks on `firstMissing`, which is only closed from the `onMissing` callback fired the first time the file is observed missing, and the test asserts `missingObserved` so a pre-existing file would fail loudly (`command_test.go:154-178`). Write is atomic (temp + `os.Rename`), so no torn read.
- Fake PID `2147483647` is not a live process, so the liveness loop returns promptly — verified: the test completes in seconds, not at the 3s timeout.

## Tests

- gofmt clean, `go vet` clean, build ok.
- `go test -race` passes for the new test and the existing background-child termination test, including repeated runs with no flakes. Re-ran here: `ok internal/config 4.318s`. No PR-attributable failures; nothing env/pre-existing flagged for this package.

Nice touch synchronizing via the `onMissing` hook instead of a sleep — it keeps the timing test deterministic.

Vasanthdev2004 (approved):
Approving. This fixes the flake rather than papering over it, which was my main worry with a "wait for it" change.

Checked the important properties: the wait is bounded (3s deadline), it fails the test on timeout rather than passing or skipping (`t.Fatalf("read sleeper pid file before timeout")`), a missing or unreadable PID file is never treated as success, and the strict liveness assertion afterwards is byte-identical to before, so the test can still fail when the child really is alive. That is the part that mattered.

Four small things, none blocking:

1. The loop breaks on the first `os.ReadFile` that returns no error without checking the content parses, so an empty or half-written PID file falls straight through to `t.Fatalf("parse sleeper pid")` instead of being retried. That is the exact race the wait exists for, just one step later. Worth moving the parse inside the loop and only breaking once it succeeds.
2. Only `os.ErrNotExist` is retried; any other transient read error fatals immediately with the deadline unspent. On Windows a sharing-violation style error is plausible there.
3. The new regression test asserts the wait happens but never asserts it is bounded, so deleting the deadline entirely would still leave it green. A case that drives the helper past its deadline would pin the property you actually added.
4. The 3s budget appears as two bare literals (the PID wait and the liveness poll) that now compose into one ~6s window against a 5s provider-command timeout plus a 1s WaitDelay. A named constant would stop them drifting apart.

Note on verification: Windows Smart App Control on my machine is currently blocking the unsigned go toolchain, so I could not run the suite locally for this review and read the code instead. CI is green across all three OSes, which covers the behaviour here.

LGTM.

PR #726, opened 2026-07-18
fix(sandbox): classify git push as network access

PierrunoYT:
## Summary

- classify `git push` and remote `git archive` invocations as network-sensitive with shared, positional Git option parsing
- fail closed for reviewed shell function, CMD, PowerShell, `env -S`, shell-launcher, BusyBox, and strace forms while avoiding false prompts for non-executing text and local operands
- exercise the existing turn-scoped network grant path with an integration regression proving an approved `git push` receives the temporary network overlay, including the `gitlawb://` transport

Fixes #703.

## Latest review finding: addressed in f4df81a0

The dynamic interpreter-source finding from the review of `04ee0e5c` is fixed by `f4df81a00840fdfc4d9f6e23062c3a88dcef686e`. The shared bounded delegation resolver retains the original AST words through each strace/BusyBox child instead of classifying reconstructed, expansion-free text. Nested shell and env source therefore remains network-gated; literal local source and dynamic data operands remain local. AST and fallback paths share the delegation/depth contract.

The branch also contains current main (`c1937dfac72e6ad0e5ade6e48e2d9c17d9c3e5d6`), with zero commits behind it, addressing the review's fresh-base gate.

Independent revalidation on September 13:

- Full sandbox and agent packages pass on Windows: `go test ./internal/sandbox ./internal/agent -count=1`.
- All 36 delegated-source AST/fallback cases and the approved-push turn-grant integration test pass in three repetitions on Windows, and in three race-enabled repetitions on Linux.
- With the preceding analyzer/resolver implementation restored in an isolated copy, nine AST cases fail in each of three race-enabled repetitions, including both reported reproductions. They return `action "allow" ... (Network=false), want network prompt` before the fix and correctly prompt at the current head.
- The controls, fallback path, and both sides of the depth bound remain covered. These are classification tests; no network payload is executed.

No further source change was needed during this verification. The outstanding changes-requested review targets the earlier commit; renewed reviewer approval remains necessary.
## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test -race ./internal/sandbox ./internal/agent`
- `go test ./internal/agent -run '^TestRunApprovedGitPushPromptAppliesTurnNetworkGrant$' -count=1`
- `go run ./cmd/zero-release build`
- `go run ./cmd/zero-release smoke`
- `make lint-static` (0 issues)
- `make vulncheck` (`No vulnerabilities found.`)
- `git diff HEAD --check`

## Summary by CodeRabbit

* **New Features**
[...]

PR #804, opened 2026-07-24
fix(windows): use PowerShell for command execution

### LINKED ISSUE #454: Can we get a better Shell for Windows?
### Problem

Can we get a better Shell for Windows? 

Image

Image

### Requested behavior

The Agent should be able to run shell commands fine on Windows

### Why it belongs in Zero

Because Zero uses somekind of shell which is not reliable

_No response_

_No response_

anandh8x:
## Summary

- prefer PowerShell 7, then Windows PowerShell, with `cmd.exe` as the final fallback
- share host-shell selection across agent command tools and the interactive `!` shell escape
- configure UTF-8 output and preserve restricted-sandbox safeguards for incompatible MSYS executables
- cover PowerShell initialization, pipelines, UTF-8 output, workspace writes, and outside-workspace denial on Windows

Fixes #454

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- Windows cross-compilation for affected packages
- manual TUI shell-escape smoke check with a real PTY on Linux

The pinned advisory linter reports only existing repository-wide findings unrelated to this change. Native PowerShell behavior is validated by the Windows CI job.

## Summary by CodeRabbit

- **New Features**
  - Added PowerShell-aware host-shell execution and guidance (prefers newer PowerShell when available, falls back to cmd.exe).
  - Expanded classification for PowerShell network/web requests and destructive file removal (e.g., `Remove-Item` with recursive+force and common aliases).

- **Bug Fixes**
  - Refined shell-specific prompting/restriction messaging to better match the detected host shell.
  - Strengthened restricted-token sandbox behavior, including workspace write enforcement.

- **Tests**
  - Added a Windows restricted-token PowerShell real-sandbox smoke test.
  - Updated/expanded Windows tests to cover PowerShell vs cmd.exe behavior.


anandh8x:
@coderabbitai review

Vasanthdev2004 (changes requested):
Thanks for taking this on, moving Windows off cmd.exe is the right direction and #454 is a real pain. Two things I checked first came back clean, so let me lead with those: the restricted-token wrapping is intact (windowsShellCommandLineFromArgs is a positive recognizer, so PowerShell's 5-element argv simply misses it and falls through to per-element EscapeArg, and the plan still returns Wrapped with the ACL and capability-SID composition untouched), and the quoting is faithful (single quotes containing double quotes, trailing backslashes and UTF-8 all round-trip byte-identical through EscapeArg to CommandLineToArgvW). No injection channel and no silent wrap drop.

Three problems, and the first one I would consider blocking.

**1. [blocker] PowerShell -Command does not propagate failure, so failed commands report success.**

Running the exact argv this PR builds against Windows PowerShell 5.1:

```
cmd /c exit 5; Write-Output done            -> exit 0, output "done"
Get-Content C:\missing; Write-Output after  -> exit 0 (error printed, script continued)
cmd /c exit 1; cmd /c exit 0                -> exit 0
cmd /c exit 7                               -> exit 1  (real code lost)
```

bash.go maps a nil error to StatusSuccess, so a failed build, a failed test run, or any failing command that is not the final statement now comes back to the model as success with exit_code 0. For an agent that is worse than a crash: it will confidently proceed on a broken build. There is no `; exit $LASTEXITCODE` trailer and no `$ErrorActionPreference = 'Stop'` in the generated script. I notice the PR encodes the flattening as expected rather than fixing it, replacing three `exit_code == 7` assertions with a helper that returns 1 under PowerShell, and nothing covers the mid-script case at all.

**2. [major] `&&` and `||` are parser errors on Windows PowerShell 5.1, which is what stock Windows selects.**

```
Write-Output a && Write-Output b   -> exit 1, "The token '&&' is not a valid statement separator in this version"
cmd /c exit 1 || Write-Output fallback -> exit 1, same for '||'
```

Both work today under cmd.exe. Neither the tool description nor HostShellEnvironmentGuidance mentions it, and the windows_shell_syntax recovery hint is gated to shellKindCmd, so under PowerShell the model gets the raw parser error with no suggestion. This is invisible to CI: the runner only uses windows-latest, which ships PowerShell 7 where `&&` is valid, while detection prefers pwsh first. So the 5.1 fallback that a stock Windows user actually gets is the one configuration never tested.

**3. [major] The risk analyzer has no PowerShell vocabulary, and this PR tells the model to prefer cmdlets.**

Running the classifier against the head:

```
curl https://example.com               Net=true    (classified)
iwr https://example.com                Net=false
irm https://example.com                Net=false
Invoke-WebRequest https://example.com  Net=false
Invoke-RestMethod https://example.com  Net=false
Remove-Item -Recurse -Force C:\        Destructive=false
```

All parse cleanly, so TooComplex stays false and the unparseable-network fallback never runs either. Both pre-grant gates in Engine.Evaluate, network and destructive, are equally blind. The `.exe` suffix stripping in this PR is a genuine improvement (it fixes curl.exe), but the cmdlet vocabulary was never added while the new system prompt actively steers the model toward cmdlets. To be fair on severity: when the sandbox is fully provisioned the runtime WFP deny still blocks actual egress, so this is a missing prompt and a risk mis-level rather than always-open egress. Unelevated or degraded Windows has no such backstop.

Where I land: 1 needs fixing before this can land, since silently reporting failures as success will burn users immediately. 2 and 3 I would want addressed too, but they are additive (guidance plus a cmdlet vocabulary in the analyzer) rather than structural.

Verified on Windows with a working toolchain: the probes above are real runs against powershell.exe 5.1 and the repo's own analyzer, not inference.

Genuinely good direction, and the wrapper and quoting work is solid. Happy to re-review quickly.

gnanam1990 (dismissed):
## Verdict

Changes requested. This supersedes my earlier approval, which I withdraw.

Thank you for taking this on. Moving Windows execution off `cmd.exe` is the right direction, and the shell-selection and quoting work is careful.

I submitted that approval before @Vasanthdev2004's review was visible to me, and having since verified his first finding I no longer consider this ready to merge. I cannot change the state of a submitted review, so I am correcting it here.

**Concurrence with the reported blocker.** I have independently confirmed the exit-code propagation issue on the parts that do not require a Windows host. The generated script contains no exit-code trailer and no error preference, so a non-terminal failure is not propagated (`internal/tools/shell_runtime.go:43`):

```go
const windowsPowerShellUTF8Prefix = "try { [Console]::OutputEncoding = [System.Text.Encoding]::UTF8 } catch {}\n"
```

The behaviour is then recorded as expected in the tests rather than corrected (`internal/tools/bash_tool_test.go:899`):

```go
func helperFailureExitCode() int {
	if runtime.GOOS == "windows" && detectShellRuntime(runtime.GOOS).Kind == shellKindPowerShell {
		// A native program failure at the end of a PowerShell -Command script is
		// represented by PowerShell's process status rather than the child's
		// exact code.
		return 1
	}
	return 7
}
```

Three assertions that previously pinned `exit_code: 7` now call this helper. Reporting a failed command as successful is the most consequential defect class for an agent, which will proceed on a broken build, and I would ask that it be resolved before this lands.

I should also note a limitation in my original review: I recorded that I had no Windows host and then approved regardless. The weakening of the existing assertions was visible in the diff without a Windows host, and I should have caught it. Apologies for the noise.

**Verification.** On macOS (darwin/arm64), `make fmt-check`, `go vet ./...` and `git diff --check` are clean, the full suite passes at 82 packages with no failures, `go test -race -count=5` passes for `internal/tools` and `internal/sandbox`, and both Windows architectures cross-compile. Reverting `internal/sandbox/safe_command.go` to base while retaining the new tests fails `TestAnalyzeCommand` and `TestEngineClassifiesPowerShellCurlCommandAsNetwork`, which confirms those tests exercise the change rather than passing incidentally.

The program-name normalisation is a genuine improvement. On base, `curl.exe`, `C:\tools\curl.exe`, `.\curl.exe` and `\\host\share\curl.exe` normalised to `curl.exe`, `c:toolscurl.exe`, `.curl.exe` and `hostsharecurl.exe` respectively, none of which matched the `curl` network entry. On this branch all four normalise to `curl` and are correctly classified.

**The executable-suffix list is narrower than `PATHEXT`.** `internal/sandbox/safe_command.go:458,491` strip `.exe`, `.cmd`, `.bat` and `.com`. Measured on this branch, `curl.ps1`, `curl.vbs` and `curl.wsf` all normalise unchanged and fail to match the network entry. `.ps1` is a notable omission in a change concerning PowerShell. Deriving the set from `PATHEXT`, with the documented default as a fallback, would avoid a further increment later.

**Backslash handling in `normalizeProgramToken`** (`internal/sandbox/safe_command.go:436`). The comment removed in this change stated that `vi\m` must normalise to `vim` and not `m`. That case is preserved, but `vi\m.exe` now yields `m`, `cur\l.exe` yields `l`, and `gi\t.exe` yields `t`, where base produced `vim.exe`, `curl.exe` and `git.exe`. This is not a regression in classification outcome, since base did not match the bare entries either, but it departs from the stated invariant and would become exploitable if suffixed entries were added to the lookup tables. A short comment on `windowsExecutablePathBasename` recording the trade-off between POSIX escape semantics and Windows path separators would help future readers.

**`powerShellAliasWord` introduces a second list** (`internal/tools/shell_runtime.go`). The comment on `windowsMsysProneNames` describes it as the single source of truth from which every detection path derives, so that the paths cannot drift. The new `{cat, ls}` exemption sits outside that set. It is correct today, as those are the only two entries that are also PowerShell built-in aliases, but no test asserts the relationship, and adding `sort` or `tee` to the prone list later would produce spurious warnings under PowerShell.

**The description overstates continuous integration coverage.** It states that native PowerShell behaviour is validated by the Windows CI job. `TestWindowsRestrictedTokenPowerShell` is gated on `ZERO_SANDBOX_REAL_SMOKE=1`, which is set only in `scripts/sandbox-smoke.sh:52`, and that script is not referenced from `.github/workflows`. The `windows-latest` job therefore covers shell selection and argument construction through the standard suite, but not PowerShell under the restricted token. This appears to be a wording matter rather than missing work, as the script's `-run` filter is correctly extended. @Vasanthdev2004's observation that `windows-latest` ships PowerShell 7, leaving the 5.1 fallback untested, is a related and more significant gap.

**Outside the scope of this change.** Newline is not treated as a statement separator by either segmenter, so `Get-ChildItem\nbash -c 'x'` yields a single segment from both `powerShellCommandSegments` and `windowsCommandSegments` and the MSYS guard inspects only the first word. This is present on base and not introduced here; this change in fact improves segmentation by handling `;`. It may warrant a separate issue.

**Limitations of this review.** I do not have access to a Windows host. All Windows-specific statements above rest on reading the code and on cross-compilation. I did not run `scripts/sandbox-smoke.sh`, and I did not observe PowerShell selection, the UTF-8 prefix, or restricted-token enforcement in execution. @Vasanthdev2004's findings were produced on a working Windows toolchain and should carry more weight than mine in those areas.

**Process notes.** Issue #454 carries the `enhancement` label rather than `issue-approved`. As a collaborator pull request this proceeds through the internal cycle, so it is not a ground for closure, but the label may be worth applying if it serves as the tracking mechanism. This change also overlaps #726, which modifies `internal/sandbox/safe_command.go` and `internal/sandbox/analyzer_test.go` as well; both alter network classification, and it would be worth sequencing them deliberately.

gnanam1990 (changes requested):
## Verdict

Changes requested. This supersedes my earlier approval, which I withdraw.

Thank you for taking this on. Moving Windows execution off `cmd.exe` is the right direction, and the shell-selection and quoting work is careful.

I submitted that approval before @Vasanthdev2004's review was visible to me, and having since verified his first finding I no longer consider this ready to merge. I cannot change the state of a submitted review, so I am correcting it here.

**Concurrence with the reported blocker.** I have independently confirmed the exit-code propagation issue on the parts that do not require a Windows host. The generated script contains no exit-code trailer and no error preference, so a non-terminal failure is not propagated (`internal/tools/shell_runtime.go:43`):

```go
const windowsPowerShellUTF8Prefix = "try { [Console]::OutputEncoding = [System.Text.Encoding]::UTF8 } catch {}\n"
```

The behaviour is then recorded as expected in the tests rather than corrected (`internal/tools/bash_tool_test.go:899`):

```go
func helperFailureExitCode() int {
	if runtime.GOOS == "windows" && detectShellRuntime(runtime.GOOS).Kind == shellKindPowerShell {
		// A native program failure at the end of a PowerShell -Command script is
		// represented by PowerShell's process status rather than the child's
		// exact code.
		return 1
	}
	return 7
}
```

Three assertions that previously pinned `exit_code: 7` now call this helper. Reporting a failed command as successful is the most consequential defect class for an agent, which will proceed on a broken build, and I would ask that it be resolved before this lands.

I should also note a limitation in my original review: I recorded that I had no Windows host and then approved regardless. The weakening of the existing assertions was visible in the diff without a Windows host, and I should have caught it. Apologies for the noise.

**Verification.** On macOS (darwin/arm64), `make fmt-check`, `go vet ./...` and `git diff --check` are clean, the full suite passes at 82 packages with no failures, `go test -race -count=5` passes for `internal/tools` and `internal/sandbox`, and both Windows architectures cross-compile. Reverting `internal/sandbox/safe_command.go` to base while retaining the new tests fails `TestAnalyzeCommand` and `TestEngineClassifiesPowerShellCurlCommandAsNetwork`, which confirms those tests exercise the change rather than passing incidentally.

The program-name normalisation is a genuine improvement. On base, `curl.exe`, `C:\tools\curl.exe`, `.\curl.exe` and `\\host\share\curl.exe` normalised to `curl.exe`, `c:toolscurl.exe`, `.curl.exe` and `hostsharecurl.exe` respectively, none of which matched the `curl` network entry. On this branch all four normalise to `curl` and are correctly classified.

**The executable-suffix list is narrower than `PATHEXT`.** `internal/sandbox/safe_command.go:458,491` strip `.exe`, `.cmd`, `.bat` and `.com`. Measured on this branch, `curl.ps1`, `curl.vbs` and `curl.wsf` all normalise unchanged and fail to match the network entry. `.ps1` is a notable omission in a change concerning PowerShell. Deriving the set from `PATHEXT`, with the documented default as a fallback, would avoid a further increment later.

**Backslash handling in `normalizeProgramToken`** (`internal/sandbox/safe_command.go:436`). The comment removed in this change stated that `vi\m` must normalise to `vim` and not `m`. That case is preserved, but `vi\m.exe` now yields `m`, `cur\l.exe` yields `l`, and `gi\t.exe` yields `t`, where base produced `vim.exe`, `curl.exe` and `git.exe`. This is not a regression in classification outcome, since base did not match the bare entries either, but it departs from the stated invariant and would become exploitable if suffixed entries were added to the lookup tables. A short comment on `windowsExecutablePathBasename` recording the trade-off between POSIX escape semantics and Windows path separators would help future readers.

**`powerShellAliasWord` introduces a second list** (`internal/tools/shell_runtime.go`). The comment on `windowsMsysProneNames` describes it as the single source of truth from which every detection path derives, so that the paths cannot drift. The new `{cat, ls}` exemption sits outside that set. It is correct today, as those are the only two entries that are also PowerShell built-in aliases, but no test asserts the relationship, and adding `sort` or `tee` to the prone list later would produce spurious warnings under PowerShell.

**The description overstates continuous integration coverage.** It states that native PowerShell behaviour is validated by the Windows CI job. `TestWindowsRestrictedTokenPowerShell` is gated on `ZERO_SANDBOX_REAL_SMOKE=1`, which is set only in `scripts/sandbox-smoke.sh:52`, and that script is not referenced from `.github/workflows`. The `windows-latest` job therefore covers shell selection and argument construction through the standard suite, but not PowerShell under the restricted token. This appears to be a wording matter rather than missing work, as the script's `-run` filter is correctly extended. @Vasanthdev2004's observation that `windows-latest` ships PowerShell 7, leaving the 5.1 fallback untested, is a related and more significant gap.

**Outside the scope of this change.** Newline is not treated as a statement separator by either segmenter, so `Get-ChildItem\nbash -c 'x'` yields a single segment from both `powerShellCommandSegments` and `windowsCommandSegments` and the MSYS guard inspects only the first word. This is present on base and not introduced here; this change in fact improves segmentation by handling `;`. It may warrant a separate issue.

**Limitations of this review.** I do not have access to a Windows host. All Windows-specific statements above rest on reading the code and on cross-compilation. I did not run `scripts/sandbox-smoke.sh`, and I did not observe PowerShell selection, the UTF-8 prefix, or restricted-token enforcement in execution. @Vasanthdev2004's findings were produced on a working Windows toolchain and should carry more weight than mine in those areas.

**Process notes.** Issue #454 carries the `enhancement` label rather than `issue-approved`. As a collaborator pull request this proceeds through the internal cycle, so it is not a ground for closure, but the label may be worth applying if it serves as the tracking mechanism. This change also overlaps #726, which modifies `internal/sandbox/safe_command.go` and `internal/sandbox/analyzer_test.go` as well; both alter network classification, and it would be worth sequencing them deliberately.

anandh8x:
@gnanam1990 The latest review seems to repeat findings from the previous head. The exit-code blocker is fixed in bdade37c: PowerShell now uses ErrorActionPreference Stop, propagates LASTEXITCODE, restores the exact exit-code assertions, and has a real powershell.exe 5.1 integration test. 93c4d65c also addresses the PowerShell removal-alias findings. All CI checks are green on the current head. Could you please re-review the latest commit and clear the changes-requested verdict if everything looks good?

gnanam1990 (approved):
## Verdict

Approve.

Reviewed at `93c4d65c8826`, base `ac50a5a840d2`, re-confirmed against the live head before posting.

**I withdraw the blocker from my previous review.** @anandh8x is correct that it repeated findings from a superseded head: I reviewed `729d5ee` and posted after `bdade37` and `93c4d65` had already landed. That was my error, and the fault was procedural rather than a judgement call — I did not re-fetch the pull request head before submitting. I have changed how I run these so it does not recur. Apologies for the wasted round trip.

This approval does not clear @Vasanthdev2004's outstanding request for changes, and should not. He identified the blocker, and whether the fix satisfies him is properly his call.

**The exit-code fix is real, and I verified it rather than taking the commit message.** `internal/tools/shell_runtime.go:45,47` add `$ErrorActionPreference = 'Stop'` and `if ($null -ne $LASTEXITCODE) { exit $LASTEXITCODE }`, composed around the command at line 138.

The most persuasive evidence is in the tests rather than the production code. `helperFailureExitCode` is now an unconditional `return 7`; the PowerShell special case that previously encoded the flattening is gone, so the assertions demand the child's exact exit code on every platform. A test that formerly accepted the defect now forbids it, which is the right shape for this fix.

`TestWindowsPowerShellArgumentsPropagateFailures` runs a real `powershell.exe` located through `exec.LookPath`, gated only on `runtime.GOOS == "windows"` and not behind an opt-in environment variable. It therefore executes on the `windows-latest` job, and since Windows PowerShell 5.1 is present on that image, it closes the gap @Vasanthdev2004 identified where the 5.1 fallback was the one configuration never exercised. It runs the exact reported case, `cmd /c exit 5; Write-Output done`.

**The chain-operator finding is addressed.** `TestWindowsPowerShellGuidanceAvoidsUnsupportedChainOperators` and `TestWindowsPowerShellCommandIssueFlagsUnsupportedChainOperators` cover the `&&` and `||` parse failure on 5.1, and the detector now emits a `windows_powershell_version` issue rather than surfacing a bare parser error.

**The analyzer vocabulary is addressed, and completely.** I probed the classifier directly at this head. `iwr`, `irm`, `Invoke-WebRequest` and `Invoke-RestMethod` all classify as network, alongside `curl`. `Remove-Item` classifies as destructive, and so does every one of PowerShell's built-in aliases for it: `ri`, `rd`, `rmdir`, `del`, `erase` and `rm` are all present in `powerShellRemoveItemPrograms`. That is the full alias set rather than a sample, which is the distinction that usually matters here.

**Verification.** On macOS (darwin/arm64), `make fmt-check`, `go vet ./...` and `git diff --check` are clean, the full suite passes at 82 packages with no failures, and both Windows architectures cross-compile. All nine CI checks pass on the live head.

**Remaining non-blocking observations**, unchanged from my earlier pass and none of them worth holding the change for:

The executable-suffix list at `internal/sandbox/safe_command.go` strips `.exe`, `.cmd`, `.bat` and `.com`, which is narrower than `PATHEXT`. Measured here, `curl.ps1` and `curl.vbs` normalise unchanged and do not match the network entry. `.ps1` remains the notable omission in a change concerning PowerShell.

Backslash handling in `normalizeProgramToken` still yields `l` for `cur\l.exe` and `m` for `vi\m.exe`, departing from the invariant the removed comment stated. As before, this is not a regression in classification outcome, since base did not match those either, but it would become exploitable if suffixed entries were ever added to the lookup tables.

`powerShellAliasWord` remains a second list alongside `windowsMsysProneNames`, whose comment describes itself as the single source of truth from which every detection path derives. It is correct today, but no test asserts the relationship.

On the description's claim that native PowerShell behaviour is validated by the Windows CI job: that is now substantially true for shell selection, argument construction and exit-code propagation, given the new ungated integration test. It remains untrue for PowerShell under the restricted token, since `TestWindowsRestrictedTokenPowerShell` is still gated on `ZERO_SANDBOX_REAL_SMOKE=1` and `scripts/sandbox-smoke.sh` is not referenced from `.github/workflows`. A one-line qualification would make the claim accurate.

**Limitations.** macOS only. The Windows statements rest on reading the code, on cross-compilation, and on the CI result; I did not execute PowerShell or the restricted-token sandbox myself.

Vasanthdev2004 (approved):
Re-reviewed on 93c4d65c. My three findings are fixed, and I checked them the same way I found them, by running things rather than reading the diff. Clearing my stale block.

**1, the blocker, is fixed.** The exact argv this now builds, against real Windows PowerShell 5.1 on this machine:

```
command                                  before   after
cmd /c exit 7                                 1       7
cmd /c exit 5; Write-Output done              0       5
Get-Content C:\missing; Write-Output after    0       1
cmd /c exit 1; cmd /c exit 0                  0       0
cmd /c exit 3 | Out-Null                      1       3
$x = cmd /c exit 4                            1       4
```

Every case I reported now propagates, including two I added afterwards (pipeline and assignment). The last row is the right answer rather than a miss: last command wins, same as sh.

**2 is fixed, and better than I asked for.** I only suggested guidance; you added a preflight `windows_powershell_version` block, so the model is stopped before it hits the parser error instead of being told about it afterwards. `isWindowsPowerShell` keys on the executable basename, so pwsh 7 correctly keeps `&&`.

**3 is fixed.** Against the real analyzer at this head:

```
iwr / irm / Invoke-WebRequest / Invoke-RestMethod   Net=true
Remove-Item -Recurse -Force / ri -Recurse -Force    Destructive=true
```

`Start-BitsTransfer` is still unclassified, but I did not raise it and it is the same shape as gaps on the POSIX side, so I would leave it.

## One new thing, and it is the reason I am writing at length

Stopping a session with `exit` no longer works. Ctrl+C still does.

```
--- FAIL: TestWriteStdinStopIntentTerminatesNonTTYSession
    stop input "exit\n" should not leave session running
    output="Command is still running. session_id: 1000"
```

Deterministic here, 3 runs out of 3. It passes on main and fails at every commit on this branch including 729d5ee9, so it comes from the move to PowerShell itself rather than from the hardening commits. The mechanism is that `exit` as a stop intent relies on the shell reading stdin as a command stream, and `powershell -Command <script>` does not: stdin goes to the child, which ignores it.

The reason nobody has seen this: **this machine has no PowerShell 7, so it runs 5.1, and CI's windows-latest has 7.** That is the same blind spot I raised as finding 2 last time, except now it is hiding a real regression rather than a syntax error. A stock Windows box does not have pwsh 7; it is a separate install. So the configuration most users actually run is the one configuration never exercised.

I am approving rather than blocking on this, because Ctrl+C still stops sessions so nothing is stuck, two other reviewers have signed off, and it is inherent to the change's purpose rather than a defect in the recent commits. But it should be a decision rather than a surprise, and I think it wants either a fix or a follow-up issue before this ships.

Worth flagging separately for kevin: `$ErrorActionPreference = 'Stop'` is the right fix for the exit-code hole, but it also discards partial results, which POSIX does not do since `sh -c` runs without `set -e`.

```
Get-ChildItem missing,real     without Stop: "1"       with Stop: ""
Get-Content existing,missing   without Stop: "hello"   with Stop: ""
```

An agent reading several paths where one is absent now gets nothing instead of what it could have had. Still the right trade against silently reporting failures as success, just worth knowing it is a cross-platform behaviour split.

Good work on this one. The quoting, the wrapper preservation and the analyzer vocabulary all hold up under probing.

Vasanthdev2004:
@anandh8x pulling the session-stop regression out of my review into its own comment, since it was buried at the bottom and it is the one thing here you cannot reproduce on CI.

**Stopping a session with `exit` no longer terminates it. Ctrl+C still does.**

```
--- FAIL: TestWriteStdinStopIntentTerminatesNonTTYSession
    stop input "exit\n" should not leave session running
    output="Command is still running. session_id: 1000"
```

Deterministic here, 3 runs out of 3, always the `exit\n` case and never the `\u0003` one.

**Why CI is green and you probably cannot reproduce it.** This machine has no PowerShell 7, so `detectShellRuntime` picks Windows PowerShell 5.1. The `windows-latest` runner ships pwsh 7. So the failure only appears on the configuration a stock Windows user actually has, since pwsh 7 is a separate install. That is the same blind spot I raised as finding 2 last round, except this time it is hiding a behaviour regression rather than a syntax error.

To reproduce: on a box with no `pwsh` on PATH and no `C:\Program Files\PowerShell\7\pwsh.exe`,

```bash
go test ./internal/tools/ -run TestWriteStdinStopIntentTerminatesNonTTYSession -v
```

**Scope.** It fails at every commit on this branch including 729d5ee9, and passes on main, so it comes from the move to PowerShell itself rather than from the two hardening commits. Nothing you added recently caused it.

**The bit I would start from.** Both stop inputs take the same path: `shouldInterruptExecSession` returns true for `\u0003` and for `exit` alike (`exec_command.go:379`), and both then reach `manager.Continue` with `Interrupt: true` AND `Input: []byte(chars)` (`exec_command.go:354`). So the classifier is fine and the interrupt is requested in both cases. The difference has to be in what actually happens to the process afterwards, which is where a `powershell -Command <script>` child differs from a cmd.exe one. I did not chase it further than that, and I did not want to guess at a mechanism I had not proven.

I approved the PR rather than blocking on this, because Ctrl+C still stops sessions so nothing gets stuck, and the rest of the change holds up well under probing. But it is worth either fixing or filing before this ships, since the model reaching for `exit` and getting "still running" back is a loop it may not get out of on its own.

Everything else from my review is genuinely fixed. The exit-code propagation is solid on real 5.1, including the pipeline and assignment cases I added afterwards.

Vasanthdev2004:
@anandh8x correction to my comment above: I had the diagnosis wrong, and wrong in a way that would have sent you looking in the wrong place. Sorry.

**It is not about `exit` versus Ctrl+C.** I said it was always `exit\n` and never `\u0003`, based on three runs on the branch. On main it fails on `\u0003` just as readily. Across six runs it failed three times, alternating between the two.

**It is timing.** Same helper, same session, varying only the yield:

```
yield=1000ms  chars="\x03"    stillRunning=true
yield=1000ms  chars="exit\n"  stillRunning=false
yield=3000ms  chars="\x03"    stillRunning=false
yield=3000ms  chars="exit\n"  stillRunning=false
yield=6000ms  chars="\x03"    stillRunning=false
yield=6000ms  chars="exit\n"  stillRunning=false
```

Identical on a second run. What fails is whichever stop happens FIRST in a fresh process, because it pays the PowerShell cold start; by the second one the interpreter is warm. That is why the subcase looked like it was `exit` on one branch and `\u0003` on another, and it is why the stop-intent classifier is not involved at all. `shouldInterruptExecSession` returns true for both, and `Continue` calls `process.terminate()` for both.

**So the severity is lower than I made it sound.** The session does terminate. It just is not reaped inside the one second the caller waited, so the first stop can come back saying "still running" while termination is already under way. Polling again gets the exit code. It is not a session that cannot be stopped, which is what I implied, and I should not have said that before pinning the mechanism.

What is left is still worth doing, just smaller than I framed it:

- The test is now flaky on any host without pwsh 7, since it hard-codes a 1s yield. Polling until exited, or a longer wait, would fix that.
- The model-facing behaviour is worth a thought: the first stop reporting "still running" is recoverable only if the model polls again, and the output does tell it to. Probably fine, but it is a real difference from cmd.exe, where teardown fit inside the same second.

My earlier claim that this was inherent to the PowerShell move still holds, and so does the point that CI cannot see it because `windows-latest` has pwsh 7 while a stock Windows box does not. Everything else in my review stands.

PR #807, opened 2026-07-26
fix(tui): paste clipboard images with Ctrl+V (#534)

### LINKED ISSUE #534: Support pasting images from clipboard in the TUI
## Summary

It would be useful if Zero supported pasting images directly from the OS clipboard into the TUI, similar to the UX in tools like Claude Code.

## Expected user flow

1. Copy or screenshot an image into the OS clipboard.
2. Focus the Zero TUI composer.
3. Press `Cmd+V` on macOS or `Ctrl+V` on other platforms.
4. Zero attaches the image to the next message.
5. The composer shows a familiar placeholder such as:

```text
[Image #1]
```

For multiple images, it could show:

```text
[Image #1] [Image #2]
```

## Current behavior

Zero already has image attachment support internally, but pasting an image directly from the OS clipboard into the TUI is not currently reliable/available as a user-facing flow.

In practice, pressing paste with an image in the clipboard may result in nothing being attached.

## Why this matters

This would make screenshot-driven workflows much smoother:

- Paste a UI screenshot and ask Zero to inspect it.
- Paste an error dialog or terminal screenshot.
- Paste a design/mockup image without saving it to disk first.

## Notes / investigation scope

This issue is intentionally framed as a feature request plus investigation rather than assuming a specific implementation.

Things likely worth investigating:

- How terminal/TUI apps can detect paste attempts when the clipboard contains an image.
- Whether Bubble Tea keyboard enhancement support is sufficient across common terminals.
- Whether platform-specific clipboard polling or native clipboard integration is needed.
- How to present pending image attachments in the composer, preferably using the familiar `[Image #n]` placeholder UX.
- Cross-platform behavior for macOS, Linux, and Windows terminals.

## Acceptance criteria

- Pasting an image from the OS clipboard into the Zero TUI attaches it to the next message.
- The composer shows a clear pending image placeholder, e.g. `[Image #1]`.
- Multiple pasted images are represented clearly, e.g. `[Image #1] [Image #2]`.
- The implementation works or degrades gracefully across common terminals/platforms.

Vasanthdev2004:
## The problem

Pasting a screenshot into the composer with Ctrl+V did nothing. Right-click paste worked, which is what made this confusing to diagnose: the feature looked implemented, and an earlier look at the issue concluded it already was.

The image plumbing genuinely is there. `routePaste` probes the OS clipboard for image content whenever a paste arrives with empty text, and `clipboardImageMsg` attaches it.

What was missing is that a clipboard holding only an image produces **no bracketed paste at all**. The terminal has no text to send, so `routePaste` never runs and its probe never fires. Right-click reached the probe only because that path always delivers a `clipboardReadMsg`, empty or not. And Bubble's own Ctrl+V binding is deliberately disabled (`input.KeyMap.Paste.SetEnabled(false)`) so that bracketed paste stays the single text path, which means nothing at all was listening for the key.

## The fix

Handle Ctrl+V by issuing the image probe directly.

It reads image content only and never text, so bracketed paste remains the sole text path and Ctrl+V with text on the clipboard behaves exactly as before. `readClipboardImageCmd` yields no message when there is no image, so it stays a silent no-op. It is gated on `noBlockingModal()` so a permission prompt, picker or the setup screen still swallows the key.

## Verification

- New `TestCtrlVProbesClipboardForImage` and `TestClipboardImageProbeSilentWithoutImage`.
- The existing Ctrl+V guard is kept and renamed to say what it actually protects: no TEXT is inserted. That is the property that matters, since double-pasting text is what the disabled binding was guarding against.
- `TestSetupCredentialsCtrlVDoesNotRunClipboardPaste` still passes unchanged, confirming the modal gate leaves the setup screen alone.
- `gofmt`, `go vet ./internal/tui/`, `go build ./...` clean; the tui suite passes apart from `TestAltScreenTranscriptScrollKeepsFooterFixed`, which fails on an unmodified tree too and is unrelated.

`readClipboardImage` becomes a var so the tests stub it rather than depending on the machine's real clipboard.

Fixes #534

## Summary by CodeRabbit

- **New Features**
  - Added Ctrl+V / Cmd+V paste handling to probe the clipboard for images and import image data when available (only when no blocking modal is open).
- **Bug Fixes**
  - Prevented Ctrl+V from altering composer text when probing for clipboard images.
  - When an image exists but can’t be extracted, the UI now shows a clear unreadable-clipboard error with an install hint (pngpaste).
- **Tests**
  - Expanded and refined clipboard-image tests to cover no-image, success, Ctrl+V/Cmd+V behavior, and unreadable-clipboard failures (including Darwin-specific cases).

## On Cmd+V

The first version of this matched `keyCtrl` only, which is `ModCtrl`, so Cmd+V did not reach the probe even though the description claimed it did. Raised in review and fixed: the handler now matches `ModSuper` as well, which is how the rest of the TUI models Command.

That only helps where the terminal delivers the key to the application. A terminal that handles Cmd+V itself pastes the clipboard TEXT and sends no key event at all, so an image-only clipboard still produces nothing for this to react to. Worth knowing before anyone reads this as a complete macOS fix.

### DIFF
--- a/internal/imageinput/clipboard.go
+++ b/internal/imageinput/clipboard.go
@@ -103,41 +103,71 @@ func readClipboardImageWindows() ([]byte, error) {
 
 // readClipboardImageDarwin uses osascript to check for and read a clipboard
 // image. Returns (nil, nil) when no image is present.
-func readClipboardImageDarwin() ([]byte, error) {
-	// Check clipboard info for image classes.
-	check := `osascript -e 'clipboard info'`
-	out, err := exec.Command("sh", "-c", check).Output()
-	if err != nil {
-		return nil, nil
+// darwinClipboardInfo reports what the pasteboard is currently offering, and
+// darwinClipboardExtract pulls the PNG bytes out of it.
+//
+// Both are vars so the two-step logic below can be driven in a test. The step
+// that matters, "the clipboard holds an image and nothing here can extract it",
+// only happens when the first succeeds and the second fails, which cannot be
+// arranged by stubbing the caller: a test that substitutes ReadClipboardImage
+// asserts the plumbing around this function rather than the decision inside it,
+// and stays green with that decision reverted.
+var (
+	darwinClipboardInfo = func() (string, error) {
+		out, err := exec.Command("sh", "-c", `osascript -e 'clipboard info'`).Output()
+		return string(out), err
 	}
-	info := string(out)
-	if !strings.Contains(info, "PNG") && !strings.Contains(info, "JPEG") && !strings.Contains(info, "TIFF") && !strings.Contains(info, "GIF") {
-		return nil, nil
-	}
-	// Write clipboard image to a temp file via AppleScript, then read it.
-	// Using pngpaste if available, falling back to a Python one-liner.
-	cmd := exec.Command("sh", "-c", `pngpaste - 2>/dev/null || python3 -c "
+	darwinClipboardExtract = func() ([]byte, error) {
+		// pngpaste if available, falling back to a Python one-liner. Neither
+		// ships with macOS: pngpaste is Homebrew-only and Apple dropped the
+		// bundled PyObjC, so a stock Mac fails here every time.
+		cmd := exec.Command("sh", "-c", `pngpaste - 2>/dev/null || python3 -c "
 import AppKit, sys
 pb = AppKit.NSPasteboard.generalPasteboard()
 data = pb.dataForType_(AppKit.NSPasteboardTypePNG)
 if data:
     sys.stdout.buffer.write(data.bytes())
 "`)
-	var stdout bytes.Buffer
-	cmd.Stdout = &stdout
-	// Past this point the clipboard is KNOWN to hold an image, because the
-	// clipboard-info probe above said so. So a failure here is not "no image", it
-	// is "there is an image and nothing on this host can extract it", and
-	// reporting it as absence is what left users staring at a paste that silently
-	// did nothing. Neither helper ships with macOS: pngpaste is Homebrew-only and
-	// Apple dropped the bundled PyObjC, so a stock Mac reaches this every time.
-	if err := cmd.Run(); err != nil {
+		var stdout bytes.Buffer
+		cmd.Stdout = &stdout
+		if err := cmd.Run(); err != nil {
+			return nil, err
+		}
+		return stdout.Bytes(), nil
+	}
+)
+
+// darwinClipboardOffersImage reports whether the pasteboard info line names an
+// image class.
+func darwinClipboardOffersImage(info string) bool {
+	for _, class := range []string{"PNG", "JPEG", "TIFF", "GIF"} {
+		if strings.Contains(info, class) {
+			return true
+		}
+	}
+	return false
+}
+
+func readClipboardImageDarwin() ([]byte, error) {
+	info, err := darwinClipboardInfo()
+	if err != nil {
+		return nil, nil
+	}
+	if !darwinClipboardOffersImage(info) {
+		return nil, nil
+	}
+	// Past this point the clipboard is KNOWN to hold an image, because the probe
+	// above said so. A failure here is therefore not "no image", it is "there is
+	// an image and nothing on this host can extract it". Reporting that as
+	// absence is what left users staring at a paste that silently did nothing.
+	data, err := darwinClipboardExtract()
+	if err != nil {
 		return nil, ErrClipboardImageUnreadable
 	}
-	if stdout.Len() == 0 {
+	if len(data) == 0 {
 		return nil, ErrClipboardImageUnreadable
 	}
-	return stdout.Bytes(), nil
+	return data, nil
 }
 
 // readClipboardImageLinux tries wl-paste (Wayland) then xclip (X11) to read
--- a/internal/imageinput/clipboard.go
+++ b/internal/imageinput/clipboard.go
@@ -2,6 +2,7 @@ package imageinput
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -16,6 +17,13 @@ import (
 // clipboard, or (nil, "", nil) when the clipboard has no image. Called when
 // text clipboard is empty (the user pasted a screenshot). The media type is
 // sniffed from the bytes, not trusted from the clipboard.
+// ErrClipboardImageUnreadable reports that the clipboard holds an image the host
+// has no way to extract. It is deliberately distinct from "no image on the
+// clipboard", which is an ordinary no-op: this one is actionable, and silence is
+// the wrong response to it.
+var ErrClipboardImageUnreadable = errors.New(
+	"the clipboard holds an image but no helper on this host can read it; on macOS install pngpaste (brew install pngpaste) or a Python with pyobjc")
+
 func ReadClipboardImage() ([]byte, string, error) {
 	data, err := readClipboardImageBytes()
 	if err != nil {
@@ -117,11 +125,17 @@ if data:
 "`)
 	var stdout bytes.Buffer
 	cmd.Stdout = &stdout
+	// Past this point the clipboard is KNOWN to hold an image, because the
+	// clipboard-info probe above said so. So a failure here is not "no image", it
+	// is "there is an image and nothing on this host can extract it", and
+	// reporting it as absence is what left users staring at a paste that silently
+	// did nothing. Neither helper ships with macOS: pngpaste is Homebrew-only and
+	// Apple dropped the bundled PyObjC, so a stock Mac reaches this every time.
 	if err := cmd.Run(); err != nil {
-		return nil, nil
+		return nil, ErrClipboardImageUnreadable
 	}
 	if stdout.Len() == 0 {
-		return nil, nil
+		return nil, ErrClipboardImageUnreadable
 	}
 	return stdout.Bytes(), nil
 }
--- a/internal/imageinput/clipboard_darwin_logic_test.go
+++ b/internal/imageinput/clipboard_darwin_logic_test.go
@@ -1,0 +1,84 @@
+package imageinput
+
+import (
+	"errors"
+	"testing"
+)
+
+// Drives readClipboardImageDarwin's own decision rather than the plumbing around
+// it. Both shell steps are substituted, so this runs on any platform and, more
+// importantly, actually reaches the branch under test: a test that stubs
+// ReadClipboardImage instead stays green with the production change reverted,
+// which is how the first version of this coverage slipped through.
+func TestReadClipboardImageDarwinDistinguishesUnreadableFromAbsent(t *testing.T) {
+	restore := func() func() {
+		info, extract := darwinClipboardInfo, darwinClipboardExtract
+		return func() { darwinClipboardInfo, darwinClipboardExtract = info, extract }
+	}()
+	t.Cleanup(restore)
+
+	const pngInfo = `«class PNGf», «class 8BPS»`
+
+	t.Run("image present but no extractor is actionable", func(t *testing.T) {
+		darwinClipboardInfo = func() (string, error) { return pngInfo, nil }
+		darwinClipboardExtract = func() ([]byte, error) { return nil, errors.New("exit status 127") }
+
+		data, err := readClipboardImageDarwin()
+		if !errors.Is(err, ErrClipboardImageUnreadable) {
+			t.Fatalf("err = %v, want ErrClipboardImageUnreadable; a stock Mac gets silence instead of a remedy", err)
+		}
+		if data != nil {
+			t.Fatalf("data = %q, want none", data)
+		}
+	})
+
+	t.Run("extractor producing nothing is also actionable", func(t *testing.T) {
+		darwinClipboardInfo = func() (string, error) { return pngInfo, nil }
+		darwinClipboardExtract = func() ([]byte, error) { return nil, nil }
+
+		if _, err := readClipboardImageDarwin(); !errors.Is(err, ErrClipboardImageUnreadable) {
+			t.Fatalf("err = %v, want ErrClipboardImageUnreadable for an extractor that exits 0 with no bytes", err)
+		}
+	})
+
+	// The other half of the distinction, and the reason this cannot simply always
+	// report an error: an empty clipboard is an ordinary no-op, and turning that
+	// into a notice would make every stray Ctrl+V nag the user.
+	t.Run("no image on the clipboard stays silent", func(t *testing.T) {
+		darwinClipboardInfo = func() (string, error) { return `«class utf8»`, nil }
+		darwinClipboardExtract = func() ([]byte, error) {
+			t.Fatal("extractor ran despite the clipboard offering no image")
+			return nil, nil
+		}
+
+		data, err := readClipboardImageDarwin()
+		if err != nil || data != nil {
+			t.Fatalf("data=%q err=%v, want a silent no-op", data, err)
+		}
+	})
+
+	t.Run("probe failure stays silent", func(t *testing.T) {
+		darwinClipboardInfo = func() (string, error) { return "", errors.New("osascript missing") }
+		darwinClipboardExtract = func() ([]byte, error) {
+			t.Fatal("extractor ran despite the info probe failing")
+			return nil, nil
+		}
+
+		if data, err := readClipboardImageDarwin(); err != nil || data != nil {
+			t.Fatalf("data=%q err=%v, want a silent no-op", data, err)
+		}
+	})
+
+	t.Run("working extractor returns the bytes", func(t *testing.T) {
+		darwinClipboardInfo = func() (string, error) { return pngInfo, nil }
+		darwinClipboardExtract = func() ([]byte, error) { return []byte("png-bytes"), nil }
+
+		data, err := readClipboardImageDarwin()
+		if err != nil {
+			t.Fatalf("err = %v, want none", err)
+		}
+		if string(data) != "png-bytes" {
+			t.Fatalf("data = %q, want the extractor's bytes", data)
+		}
+	})
+}
--- a/internal/tui/clipboard.go
+++ b/internal/tui/clipboard.go
@@ -23,12 +23,16 @@ type clipboardImageMsg struct {
 	err       error
 }
 
+// readClipboardImage is the OS clipboard image reader. A var so tests can
+// substitute a stub instead of depending on the developer's real clipboard.
+var readClipboardImage = imageinput.ReadClipboardImage
+
 // readClipboardImageCmd reads the OS clipboard for image content off the
 // Update goroutine. Returns a clipboardImageMsg with the bytes, or nil (no
 // command) if there is no image — the caller treats nil as a silent no-op.
 func readClipboardImageCmd() tea.Cmd {
 	return func() tea.Msg {
-		data, mediaType, err := imageinput.ReadClipboardImage()
+		data, mediaType, err := readClipboardImage()
 		if err != nil {
 			return clipboardImageMsg{err: err}
 		}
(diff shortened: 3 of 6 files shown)


Vasanthdev2004:
Both taken in 1b0f7537, and the Cmd+V one was a real gap rather than a nicety.

**Cmd+V.** You are right. `keyCtrl` is `ModCtrl`, macOS reports Command as `ModSuper`, and the rest of the TUI already models it that way in `keybindings.go`. So the handler never fired on Cmd+V while the description claimed it covered it. Now matched via a `keySuper` helper sitting alongside the existing modifier helpers.

One limit worth stating rather than leaving implied, and I have added it to the description too: this only helps where the terminal delivers the key to the application. A terminal that handles Cmd+V itself pastes the clipboard TEXT and sends no key event, so an image-only clipboard still produces nothing to react to. That is a different problem from the one this fixes.

**The tests.** Also right, and this is the part I should have caught myself. One asserted `Update` returned some non-nil command and then ran a freshly built probe; the other never called `Update` at all. Both now stub the reader before `Update` and execute the command `Update` actually returned, through the existing `execCmd` helper.

I checked how much that was worth rather than assuming. Making the branch return a different non-nil command:

```
--- FAIL: TestCtrlVProbesClipboardForImage/ctrl+v
--- FAIL: TestCtrlVProbesClipboardForImage/cmd+v
```

and the old assertion would have passed that unchanged, since the command was still non-nil. That is precisely the hole you described: it proved a command was issued, not that it was the image probe. Reverting to `ModCtrl` only fails the `cmd+v` case, and removing the route entirely fails both.

`gofmt`, `go vet` and builds for linux, darwin and windows are clean. One unrelated failure in this package locally, `TestAltScreenTranscriptScrollKeepsFooterFixed`, reproduces on a branch with none of these changes and is green in CI, so it is a local environment artifact rather than anything here.

gnanam1990 (dismissed):
## Verdict

Changes requested.

Reviewed at `1b0f7537ee6d`, base `ac50a5a840d2`, re-confirmed against the live head before posting.

Thank you for this — the TUI wiring is clean, the `readClipboardImage` indirection is the right seam for testing, and reading the clipboard off the Update goroutine rather than inside it is the correct call. My concern is not with the code you wrote. It is that on macOS the feature this pull request advertises does not work, and the way the error path is written means the user is told nothing.

**The claim does not hold on macOS, and I measured it rather than inferring it.** `internal/tui/clipboard.go:28` binds `readClipboardImage` to `imageinput.ReadClipboardImage`, and this branch does not touch `internal/imageinput/`. On darwin that function shells to `pngpaste - 2>/dev/null || python3 -c "import AppKit …"` (`internal/imageinput/clipboard.go:111`). Neither arm exists on a stock macOS install: `pngpaste` is Homebrew-only, and Apple dropped PyObjC when it retired the bundled Python 2.7.

On this machine, which is a fully-tooled developer Mac rather than a bare one:

```
pngpaste: NOT INSTALLED
/usr/bin/python3 (3.9.6):            ModuleNotFoundError: No module named 'AppKit'
homebrew python3 (3.13):             ModuleNotFoundError: No module named 'AppKit'
composed fallback chain exit code:   1
```

Driving the real function from a scratch test in-package:

```
ReadClipboardImage() -> data=false len=0 mediaType="" err=<nil>
```

`nil` data, and no error.

**The error path then converts that into silence.** `internal/imageinput/clipboard.go:120` returns `nil, nil` when the command fails, and `internal/tui/clipboard.go:39-41` treats nil data as `return nil // no image — silent no-op`. So a macOS user with an image on the clipboard presses Ctrl+V and nothing happens at all: no image, no error, no hint that a dependency is missing. That is a worse outcome than the feature being absent, because there is nothing to act on.

To be fair about scope, the defect is in pre-existing code and not in your diff, and the Linux arm (`wl-paste` then `xclip`) and the Windows arm (`powershell Get-Clipboard`) both look viable. I am requesting changes rather than filing this separately because this pull request is what makes that path user-reachable and is titled as fixing #534 for the TUI generally. Shipping it as-is closes an approved issue with a feature that is inert on one of the three platforms.

**What would resolve it, in preference order.** Distinguishing "no image on the clipboard" from "no tool available to read one" inside `readClipboardImageDarwin`, and surfacing the second as a one-line actionable notice, would be enough for me — the user learns why nothing happened and what to install. Better still would be a stock-available darwin arm, since `osascript` is already relied on ten lines above for the clipboard *check* (`clipboard.go:100-101`) and is unconditionally present; whether it can be made to yield PNG bytes is worth a look before reaching for a vendored helper. Narrowest acceptable outcome: state the macOS dependency in the issue and the release note so it is a known limitation rather than a silent one.

**One thing I did not do.** I did not exercise the Linux or Windows arms, so my statement that they are viable rests on reading. If you have run either, saying so in the description would close that gap.

**Verification.** On macOS, `go build ./...`, `go vet ./...` and `gofmt -l` clean; the affected packages pass. The scratch test above was removed and the tree is clean.

Vasanthdev2004:
You are right, and thank you for measuring it rather than reasoning about it. I have no Mac, so this is exactly the class of thing I cannot check and would have shipped.

Fixed in 17d87f18, and the fix turned out to be narrower than I expected because the code already knows the difference. The darwin path probes with `clipboard info` first and only reaches the pngpaste/PyObjC step once the clipboard is confirmed to hold an image. So a failure there is not absence, it is "there is an image here and nothing on this host can extract it". That case now returns `ErrClipboardImageUnreadable` instead of `nil, nil`, and the existing error path already turns that into a transcript notice, so the plumbing needed nothing new. The message names what to install.

The test drives the whole route rather than the function: stubbed reader returns the error, Ctrl+V issues the probe, and the assertion is that the failure lands in the transcript as an error row. Swallowing it again makes the test fail, so the silence cannot come back quietly.

Two things I deliberately did not do, in case you disagree.

I left Linux and Windows alone. Linux cannot distinguish "no tool" from "no image" without probing separately, since a missing `wl-paste` and an empty clipboard both surface as a failed command, and Windows uses PowerShell, which is always there. Only darwin has the prior probe that makes the distinction free.

I did not add a native reader. `osascript` can extract PNG data without any dependency, so the feature could actually work on a stock Mac rather than merely explaining itself. That is the better fix and I would rather do it as its own change than fold an untestable-here macOS code path into a TUI wiring PR. Happy to be told the opposite: if you would rather #534 not close until macOS genuinely works, I will hold this and write that instead.

On scope, I agree with your reasoning for raising it here rather than separately. The defect predates the branch, but the branch is what makes the path reachable, and closing an approved issue with something inert on one of three platforms is not a good trade.

`gofmt`, `go vet`, and builds for darwin, linux and windows are clean. The `internal/imageinput` suite passes. One failure in `internal/tui` locally, `TestAltScreenTranscriptScrollKeepsFooterFixed`, reproduces on branches with none of these changes and is green in CI, so it is a local artifact rather than anything here.

anandh8x (dismissed):
## Review at 17d87f1

Concise review of PR #807 (Vasanth's — paste clipboard images with Ctrl+V, fixes #534). Scope: 5 files, +162/-8, three commits. Built and tested in an isolated worktree.

### Verdict: **approve.** One small follow-up noted below.

### What the fix does

Ctrl+V now probes the clipboard for an **image only** — text pasting stays exclusively on the terminal's bracketed-paste path (Bubble's own Ctrl+V binding is disabled, so this can't double-insert text). This is needed because a clipboard holding only a screenshot produces no bracketed paste at all, so the existing `routePaste` image probe never fired. Right-click paste worked only because it always delivers a `clipboardReadMsg`.

### What's good

- **Right separation of concerns.** The handler issues `readClipboardImageCmd()` which reads image content only, never text. `readClipboardImage` yields no message when there's no image, so Ctrl+V with text on the clipboard is a silent no-op — bracketed paste handles it exactly as before.
- **`noBlockingModal()` gate.** A permission prompt, ask_user, spec review, provider/MCP wizard, picker, or STT key prompt swallows the key. The setup screen is protected earlier (the `m.setup.visible` early-return at line 1345 routes to `handleSetupKey` before reaching this case).
- **Cmd+V matched via `keySuper` (ModSuper).** macOS reports Command as ModSuper, not ModCtrl. The commit message is honest that this only helps on terminals that deliver the key to the app — a terminal that handles Cmd+V itself pastes text and sends no key event.
- **`ErrClipboardImageUnreadable` is a real UX win.** A stock Mac (no pngpaste, no bundled PyObjC) now gets an actionable error naming the remedy (`brew install pngpaste`) instead of silent nothing. The error is deliberately distinct from "no image on clipboard" (which stays a silent no-op). The test `TestClipboardImageUnreadableSurfacesToUser` proves the error reaches the transcript as a `rowError` row.
- **`readClipboardImage` is a `var` for testability.** Tests stub it rather than depending on the developer's real clipboard. Clean pattern.
- **Tests are thorough.** `TestCtrlVProbesClipboardForImage` runs both Ctrl+V and Cmd+V through `m.Update` and asserts the command is the image probe (not just that some command came back). `TestClipboardImageProbeSilentWithoutImage` proves no message is emitted when there's no image. `TestCtrlVDoesNotPasteTextIntoComposer` proves the composer is unchanged. `TestClipboardImageUnreadableSurfacesToUser` proves the error path reaches the transcript. `TestSetupCredentialsCtrlVDoesNotRunClipboardPaste` (pre-existing) proves the setup screen swallows Ctrl+V. All 9 PR-related tests pass; `go vet` clean; `go build` clean.

### One small gap (CodeRabbit flagged it)

**No test for Ctrl+V suppression under a blocking modal** (permission prompt, ask_user, picker). The setup-visible case is covered by the pre-existing `TestSetupCredentialsCtrlVDoesNotRunClipboardPaste`, but that test passes because `m.setup.visible` triggers an early return at line 1345 — **before** the Ctrl+V case at line 1702. The `noBlockingModal()` gate at line 1720 is what protects the modal cases (`m.pendingPermission != nil`, etc.), and there's no test asserting `cmd == nil` when a modal is up.

This is a real gap — the `noBlockingModal()` check is an explicit behavior requirement and the only tests for it are the setup test (which doesn't exercise `noBlockingModal()` at all) and the image/no-image tests (which run with no modal up). A one-line test like:

```go
func TestCtrlVDoesNotProbeWhilePermissionPromptUp(t *testing.T) {
    m := newModel(context.Background(), Options{})
    m.pendingPermission = &pendingPermissionPrompt{...}
    _, cmd := m.Update(testKeyCtrl('v'))
    if cmd != nil {
        t.Fatal("ctrl+v must not probe while a permission prompt is up")
    }
}
```

would close it. **Not a blocker** — the gate is a one-liner and clearly correct by inspection — but worth adding since CodeRabbit asked and the guideline says "add regression tests for behavior changes."

### Pre-existing failures (not caused by this PR)

`TestHandleAddDirCommand` fails on main too — it's an environment issue (the test expects `/tmp` not to be a write root, but this sandbox has `/tmp` granted). Same class as the `TestAltScreenTranscriptScrollKeepsFooterFixed` the PR description mentions. Unrelated to this PR.

### Verdict

**Approve.** The fix is correct, well-tested, and the `ErrClipboardImageUnreadable` addition is a genuine UX improvement for stock macOS. The one gap (modal suppression test) is a nice-to-have follow-up, not a blocker.

Vasanthdev2004:
@gnanam1990 when you have a moment, your changes-requested here is on `1b0f7537` and the macOS finding was fixed in `17d87f18`. CodeRabbit and anandan have both since approved that head, so yours is the only thing holding it and I do not think it reflects your position on the current code.

The fix turned out narrower than I expected, because the code already knew the answer. The darwin path probes `clipboard info` first and only reaches the pngpaste/PyObjC step once the clipboard is confirmed to hold an image, so a failure there is not absence, it is "there is an image here and nothing on this host can extract it". That case now returns a typed error instead of `nil, nil`, and the existing error path already turns it into a transcript notice naming what to install. No new plumbing was needed.

The test drives the whole route rather than the function, and swallowing the error again makes it fail, so the silence cannot come back quietly.

Still open, and I would value your view rather than assuming: I did not add a native `osascript` reader, which is what would make the feature actually work on a stock Mac rather than merely explain itself. I would rather do that as its own change than fold an untestable-here macOS path into a TUI wiring PR, but if you think #534 should not close until macOS genuinely works, say so and I will hold this and write that instead.

gnanam1990 (changes requested):
## Verdict

Changes requested — narrowly, and on test coverage rather than on the fix.

Reviewed at `17d87f184293`, base `81a5e6c`, re-confirmed against the live head before posting.

**The behaviour I asked for is there, and the placement is better than what I suggested.** `ErrClipboardImageUnreadable` is returned only *after* the `clipboard info` probe has already established that the clipboard holds an image, so "nothing to paste" stays an ordinary silent no-op while "there is an image and nothing here can extract it" becomes actionable. That distinction is the whole point and the comment at the branch states it well. The error text naming `brew install pngpaste` is the right level of concrete.

I traced it to the user rather than assuming: `internal/tui/clipboard.go:37` turns the error into a `clipboardImageMsg`, and `internal/tui/model.go:1285-1286` renders `"Clipboard image read failed: " + msg.err.Error()` into the transcript. So a macOS user now sees a line telling them what to install, where before they got nothing. My blocker is answered.

**The blocking point is that nothing tests the change.** I reverted both production return sites from `ErrClipboardImageUnreadable` back to `nil, nil` and ran the affected packages:

```
ok  github.com/Gitlawb/zero/internal/imageinput  0.590s
ok  github.com/Gitlawb/zero/internal/tui        13.055s
```

Both green with the fix gutted. `TestClipboardImageUnreadableSurfacesToUser` substitutes `readClipboardImage` with a stub that returns the sentinel (`composer_test.go:535`), so it proves the TUI plumbing surfaces an error handed to it — genuinely worth having, and it would have caught a swallow at the `model.go` layer — but it never reaches `readClipboardImageDarwin`, which is where the change actually is. The only reference to `ErrClipboardImageUnreadable` anywhere under a `_test.go` is that injection.

`AGENTS.md` asks for a regression test for a behaviour change, and this is the one case the protocol is unambiguous about: a test that passes with the production change reverted is not testing it.

**What would close it.** `readClipboardImageDarwin` shells out twice and has no seam, so it is not unit-testable as written. The smallest honest fix is a package-level indirection for the extract step — the same shape as the `readClipboardImage` var this PR already relies on — so a test can drive "info probe says PNG, extractor fails" and assert the sentinel. If you would rather not add a seam for this, then say in the description how you verified the darwin arm by hand and I will take that instead; an explicit manual verification is worth more than a test that asserts around the change.

**Independently, on this machine, the premise holds.** `pngpaste` is not installed, and `import AppKit` fails on both `/usr/bin/python3` (3.9.6) and Homebrew's 3.13, so the composed chain exits 1 — meaning this is the path a stock and even a fully-tooled Mac takes, not an edge case.

**Verification.** On macOS (darwin/arm64): `gofmt`, `go build ./...`, `go vet ./...` clean; `./internal/imageinput/...` and `./internal/tui/...` passing at head.

**Limitations.** I did not put an image on the clipboard to drive the real branch end to end, because doing so would clobber whatever you have copied. The Linux (`wl-paste`/`xclip`) and Windows (`Get-Clipboard`) arms are unchanged by this PR and unexercised by me.

Vasanthdev2004:
You are right, and the way you demonstrated it is the part I want to acknowledge: you reverted the production change and showed both packages still green. That is the only proof that settles this, and it is what I should have run on my own test before pushing it.

Fixed in e62fc079, taking the seam rather than the manual-verification option. I have no Mac, so "how I verified the darwin arm by hand" would have been an empty claim.

`readClipboardImageDarwin` shells out twice, so the two steps are now package-level vars, the same shape as the `readClipboardImage` var this change already relied on. That makes the case that matters reachable: info probe succeeds and reports an image, extractor then fails. From outside the function those two steps are one opaque call, which is exactly why stubbing the caller could not reach it.

Five cases, and the negative ones carry as much weight as the positive:

- image present, extractor fails, actionable
- image present, extractor exits 0 with no bytes, actionable
- clipboard offers no image, silent, and the extractor must not run
- info probe fails, silent, and the extractor must not run
- extractor works, bytes come back

The two silent cases are there because the fix is a discrimination, not an error. If it always reported, every stray Ctrl+V would nag the user, and a test suite that only checked the error path would happily accept that.

Repeating your experiment now:

```
--- FAIL: .../image_present_but_no_extractor_is_actionable
    err = <nil>, want ErrClipboardImageUnreadable; a stock Mac gets silence instead of a remedy
--- FAIL: .../extractor_producing_nothing_is_also_actionable
--- PASS: .../no_image_on_the_clipboard_stays_silent
```

Both actionable cases fail with the change reverted, and the silent ones keep passing, which is the discrimination rather than just "something fails now".

I kept `TestClipboardImageUnreadableSurfacesToUser` as well. You were right that it is worth having on its own terms: it covers the `model.go` layer swallowing the error, which the new test does not reach. The mistake was treating it as coverage of the fix.

Your independent confirmation that the premise holds on a fully-tooled Mac is useful beyond this PR, since it means the silent path was the normal one rather than an edge case. Thank you for going and checking rather than taking the description's word for it.

Vasanthdev2004:
@gnanam1990 nudge on this one: your changes-requested is on `17d87f18`, head is now `e62fc079`. You are the only thing holding it.

You were right that the old test asserted around the change rather than on it, and the way you showed it, reverting both production return sites and watching everything stay green, is the proof that settled it. I took the seam rather than the manual-verification option, since I have no Mac and "verified by hand" would have been an empty claim.

`readClipboardImageDarwin` now has its two shell steps behind package-level vars, so the case that matters is reachable: info probe reports an image, extractor then fails. Five cases, including the two silent ones, because the fix is a discrimination and a suite that only checked the error path would accept "always report".

Your experiment now fails where it should:

```
--- FAIL: .../image_present_but_no_extractor_is_actionable
--- FAIL: .../extractor_producing_nothing_is_also_actionable
--- PASS: .../no_image_on_the_clipboard_stays_silent
```

Still open and genuinely yours to call: I did not add a native `osascript` reader, so macOS explains itself rather than working. If you think #534 should not close until it actually works on a stock Mac, say so and I will hold this and write that instead.

gnanam1990 (approved):
## Verdict

Approve.

Reviewed at `e62fc0792a90`, base `81a5e6c`, re-confirmed live before posting. This withdraws my changes-requested.

**The finding is fixed, and fixed the way I hoped rather than the cheap way.** `darwinClipboardInfo` and `darwinClipboardExtract` are now vars, so the two-step decision can be driven directly, and the comment states the problem precisely: a test that substitutes `ReadClipboardImage` "asserts the plumbing around this function rather than the decision inside it, and stays green with that decision reverted". That is exactly the defect, described better than I described it.

**I re-ran the mutation that previously proved nothing.** Reverting both `return nil, ErrClipboardImageUnreadable` sites to `nil, nil`:

```
--- FAIL: TestReadClipboardImageDarwinDistinguishesUnreadableFromAbsent
    --- FAIL: .../image_present_but_no_extractor_is_actionable
    --- FAIL: .../extractor_producing_nothing_is_also_actionable
```

Previously that same mutation left `internal/imageinput` and `internal/tui` both green. Now it fails, and it fails on the two subtests that are about the decision rather than the plumbing.

**The table is scoped, not blanket**, which is the other thing I would have checked. `no_image_on_the_clipboard_stays_silent` and `probe_failure_stays_silent` pin the cases that must *not* become errors, and `working_extractor_returns_the_bytes` pins the success path. So the assertion is "unreadable is distinguishable from absent", not "everything is an error now" — and the silent no-op the original design wanted is preserved where it belongs.

**Verification.** On macOS (darwin/arm64): `gofmt -l`, `go vet ./...` and `git diff --check` clean; `./internal/imageinput/...` and `./internal/tui/...` passing. The branch is 9 behind main, so I merged `origin/main` and confirmed the *merged* tree builds and both packages still pass, rather than trusting the clean merge.

Scope is tight — six files, all clipboard and paste.

**Limitations.** I did not put an image on the clipboard to drive the real branch end to end, since that would clobber whatever you have copied; the seam is what makes that unnecessary rather than merely convenient. The Linux and Windows arms are unchanged by this PR and unexercised by me.

kevincodex1 (approved):

PR #809, opened 2026-07-26
test(cli): order the prewarm probe against the turn

Vasanthdev2004:
`TestRunExecOptimizedSessionUnderGate` fails intermittently on the Windows smoke job with `server saw 0 prewarm HEAD probes, want exactly 1`.

## Why it flakes

The test polled for up to 5s after the run returned, on the reasoning that the deadline exceeded the 3s prewarm timeout so a slow probe could not flake it. That reasoning does not hold.

The probe is issued on the run's context, and `runExec` cancels that context as it returns: the stop function `signalContext` hands back is a `CancelFunc`, which I confirmed rather than assumed.

```
before stop, ctx.Err() = <nil>
after  stop, ctx.Err() = context canceled
```

So once the run is over the probe is cancelled, not slow, and no deadline makes a cancelled probe observable. Against a mock server the turn finishes in microseconds, which leaves the probe racing teardown with nothing to protect it.

## What this does

The turn now blocks until the probe arrives, so the probe gets its window while the run is still alive rather than after it. The wait is bounded, so a probe that genuinely never fires still fails the assertion with a useful message instead of hanging.

Worth being straight about the limit: this widens the window, it does not close it. `prewarmTimeout` is armed inside the probe goroutine, so holding the turn open here cannot extend it. The comment says that and names `prewarmTimeout` as the thing to look at if this ever comes back.

One trade-off to call out. The assertion loses what little timing sensitivity it had, so a `Prewarm` that became arbitrarily late would now pass. That coverage was already illusory at a 5s poll, and if this does recur the better answer is to assert the wire behaviour in `internal/providers/openai`, which has a real happens-before seam in `prewarmDone`, and leave this test asserting only that the gate selects the optimized session.

## Checks

`gofmt`, `go vet`, `go build ./...` clean. Ran the test 6x locally and 5x under CPU contention, green throughout, no change to the happy-path duration.

Not related to any open feature work; this hits every PR that touches the Windows job.

## Summary by CodeRabbit

* **Tests**
  * Improved test synchronization for optimized session prewarming.
  * Replaced timing-based polling with deterministic request coordination, reducing potential test flakiness.
  * Preserved verification that the prewarm request occurs exactly once.

### DIFF
--- a/internal/cli/exec_turn_session_test.go
+++ b/internal/cli/exec_turn_session_test.go
@@ -29,6 +29,7 @@ func TestRunExecOptimizedSessionUnderGate(t *testing.T) {
 	// does. See the POST handler for why that ordering has to be forced.
 	headSeen := make(chan struct{})
 	var headOnce sync.Once
+	var prewarmWaitTimedOut atomic.Bool
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch {
 		case r.Method == http.MethodHead:
@@ -59,6 +60,11 @@ func TestRunExecOptimizedSessionUnderGate(t *testing.T) {
 			select {
 			case <-headSeen:
 			case <-time.After(10 * time.Second):
+				// Record it. A probe that shows up after this gave up would still
+				// leave heads at 1, so the count alone proves the probe was
+				// eventually received, not that it was ordered before the turn,
+				// which is the whole claim of this test.
+				prewarmWaitTimedOut.Store(true)
 			}
 			posts.Add(1)
 			w.Header().Set("Content-Type", "text/event-stream")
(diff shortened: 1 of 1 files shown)


Vasanthdev2004:
Good catch, and taken in d427a0d8.

You are right that the count could not carry the claim. A probe arriving after the guard expired still leaves `heads` at 1, so the assertion proved eventual receipt rather than ordering, which is the only thing this test exists to show. The expiry is now recorded and checked before the count.

Worth recording how I verified it, because my first attempt at checking this was useless. Shrinking the guard to a nanosecond did not fail the test: by the time the POST handler runs the probe has usually already landed, so `select` takes the `headSeen` case and the timeout branch is never entered. Suppressing the probe instead, by flipping the pinned transport to disable keep-alives, does exercise it:

```
exec_turn_session_test.go:117: the turn was not held until the prewarm probe arrived; the probe did not precede it
--- FAIL: TestRunExecOptimizedSessionUnderGate (11.63s)
```

The 11.63s confirms the guard actually expired rather than the assertion firing for some other reason.

kevincodex1 (approved):

PR #812, opened 2026-07-27
feat(sandbox): give each workspace an offline and an online principal

Vasanthdev2004:
Stacked on #808, and targets that branch rather than main so the diff is only the new work. #808 should land first.

## What this fixes

#808 shipped with the principal backend standing down whenever the network was denied. Deny is the default, so opting in left the restricted-token path doing all the work in ordinary use and the new identity only engaged for approved network commands. That was the honest choice at the time, but it made the feature close to inert.

The cause: network denial is enforced by block filters keyed to the offline-marker SID, and a principal token cannot carry it. `LogonUser` builds a token from an account's real group memberships, and the marker is a synthetic capability SID with no account behind it.

## How

A real local group. `ZeroSandboxOffline` is created by setup, the block filters name its SID alongside the marker, and a principal is denied the network by being a member of it. Each workspace gets two accounts that differ only in that membership, and the command's network mode picks between them.

A group rather than each principal's own SID because principals here are per workspace: one filter set covers every offline principal on the machine, instead of needing a filter per workspace and a setup re-run whenever a workspace appears.

Both principals are provisioned together even though a setup run only sees one profile. Setup needs elevation and commands do not, so provisioning lazily would mean an unelevated command discovering it needs an account it cannot create. They also get identical filesystem access, so an approved network command sees the same filesystem as an ordinary one.

## Two orderings that are load bearing

**The network plan is now built after provisioning.** The group it keys to is created there. Planning first installed filters naming only the marker and left every offline principal with an open network while the machine looked correctly set up. That is the same class of silent-downgrade bug this PR exists to remove, so it is worth naming.

**The role tag sits before the workspace hash in the account name.** Truncation at the 20 character limit then eats hash characters rather than the tag. A lost tag would collide the two roles onto one account, on exactly the workspaces most likely to truncate, and the dangerous direction of that collision is an ordinary command silently getting the online principal.

Anything that is not an explicit allow maps to the offline principal, so an unrecognised mode loses the network rather than keeps it.

## Setup-marker compatibility

The filter identity set is resolved, not assumed, and stays absent until the group exists. A machine that never provisions principals computes exactly the plan it computed before. This matters because the plan is hashed into the setup marker and re-derived on every command: an identity set that differed between setup and the command path would fail every command with "setup is out of date". There is a test for the hash changing when the group is present and for a lookup failure propagating rather than silently degrading to "no group".

## Verification, and what is not verified

[...]

PR #813, opened 2026-07-27
fix(config): make the provider-command timeout an upper bound (#811)

### LINKED ISSUE #811: provider-command timeout is a floor, not a bound, and discards diagnostics
`providerCommandTimeout` is documented and reported as a 5 second bound, but it is a floor rather than a bound. I hit this while reproducing the Windows smoke flakes fixed in #810, and it is a separate, user-facing problem.

## Observed

Running the provider-command tests under CPU contention, `LoadProviderCommand` returned after **19.7s**, and on another run after **106s**, against a 5s timeout:

```
returned after 19.7147647s, want before the background child finishes naturally after 10s
returned after 1m46.2165326s, want before the background child finishes naturally after 10s
```

The magnitudes came from synthetic load and are not representative of CI, but the code ordering below is wrong regardless of load.

## Three causes, all in `internal/config/command.go`

**1. Process creation is outside the timer.** `startCommandProcess` runs at `command.go:54`, but the timer is not armed until `command.go:65`. Everything before that is unbounded. On Windows that phase includes `CREATE_SUSPENDED`, the job-object setup, and the system-wide thread snapshot in `resumeMainThread` (`internal/config/process_windows.go`), none of which is fast on a cold or AV-heavy box.

**2. The drain after `Terminate()` is unbounded.** On the timer path, `command.go:85` does a bare `<-done` after `proc.Terminate()`. If the tree is slow to die, the call blocks for as long as that takes, well past the deadline that just fired.

**3. The timeout path discards all diagnostics.** `command.go:19-21` returns a hardcoded string. `commandOutput(stderr)` is only reached on the non-timeout branch, and the underlying error is dropped. A user whose provider command is slow gets no stderr and no clue which of the two timeout paths they hit.

## Why it matters

A provider command runs during config resolution, so this is on the startup path. A user on a slow Windows box can have Zero appear to hang for far longer than the error text promises, then get a message with no diagnostics attached.

## Notes

I have deliberately not fixed this in #810. That PR stops the *tests* from racing the deadline and is otherwise behaviour-preserving; changing what the timeout actually bounds is a real behaviour change and deserves its own review. #810 does make one narrow change here, wrapping instead of replacing on the `WaitDelay` path so the two timeout paths can finally be told apart, which is a prerequisite for fixing 3 properly.

Vasanthdev2004:
Fixes #811.

## The problem

`providerCommandTimeout` bounded only the wait, so the call could take arbitrarily longer than the 5s its error text promises. I measured 19.7s and then 106s against that 5s while reproducing the Windows flakes fixed in #810. Two phases sat outside the deadline.

**Process creation ran before the clock started.** `startCommandProcess` was called, and only then was the timer armed. On Windows that phase is `CREATE_SUSPENDED`, job-object creation and assignment, and a system-wide thread snapshot in `resumeMainThread`, none of it fast on a cold or AV-heavy machine.

**The drain after termination was unbounded.** The timer path called `Terminate()` and then did a bare receive on the wait. Terminate normally makes that immediate, which is exactly why it went unnoticed, but a tree that resists termination held the call open long past the deadline that had just expired.

This is on the startup path, so a user with a slow provider command sees Zero appear to hang for far longer than the message admits, and gets no diagnostics: the timeout branch returned a bare string with the underlying error and the command's stderr both dropped.

## What this does

The deadline is armed first and the start races it, so process creation counts against the budget. The drain is bounded by a grace period. A call that gives up leaves `cmd.Wait` running, so the output buffers are now synchronised: the I/O pump keeps writing into buffers the caller already holds, which a plain `bytes.Buffer` cannot survive.

The timeout error names the phase that ran out, wraps the cause so `errors.Is` still classifies it, and includes the redacted output the non-timeout path already reported.

## On the tests, because the first version of them was worthless

Both bounds are asserted on extracted helpers rather than end to end. That is not a stylistic choice. End to end they are untestable: a terminated command stops writing and its wait returns at once, so an unbounded drain and an unsynchronised buffer both pass a whole-call test.

I wrote the whole-call versions first and checked them against mutations. The duration test passed with the drain unbounded again. The race test passed with the buffer's lock removed, even under `-race`, because the command was never writing at the moment the snapshot was taken. Neither proved anything.

What replaced them does bind:

- `TestAwaitWithGraceGivesUp` drives a channel that never fires. Making the drain unbounded hangs it to the test timeout.
- `TestSyncBufferAllowsConcurrentWriteAndSnapshot` overlaps writes and snapshots. Removing the lock trips `-race`.
- `TestRunProviderCommandCountsProcessStartAgainstTheDeadline` uses a budget that is already spent, so the only phase that can exceed it is the start. Arming the deadline after the start instead makes it fail with `the command did not finish`, naming the wrong phase.

The end-to-end duration test is kept as a guard against gross regressions, with a comment saying plainly that it does not pin the drain bound and which test does.

## Checks

`gofmt`, `go vet`, and builds for linux, darwin and windows are clean. The full `internal/config` package passes under `-race`. Existing coverage is untouched: the 5s message mapping, exit-status wrapping and secret redaction all still assert what they did.

One behaviour change worth naming for review: the timeout error text now includes the phase and the command's stderr, so anything matching on the exact old string would need updating. Nothing in the tree does; `LoadProviderCommand` is matched on with `errors.Is` and on the `timed out` substring, both of which still hold.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved timeout errors for provider commands with clearer diagnostics and relevant command output.
  - Ensured command timeouts apply consistently across startup, execution, termination, and output processing.
  - Added bounded cleanup handling so timed-out commands return promptly.
- **Tests**
  - Added coverage for timeout limits, startup delays, diagnostic messages, and safe concurrent output handling.

### DIFF
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -17,7 +18,11 @@ func LoadProviderCommand(command string) (FileConfig, error) {
 	stdout, stderr, err := runProviderCommand(command, providerCommandTimeout)
 	if err != nil {
 		if errors.Is(err, errProviderCommandTimeout) {
-			return FileConfig{}, fmt.Errorf("provider command timed out after 5s")
+			// Report which phase ran out of time, and anything the command
+			// managed to say. This used to be a bare string with the underlying
+			// error and stderr both dropped, which left a user whose command was
+			// slow on a cold machine with nothing to act on.
+			return FileConfig{}, fmt.Errorf("provider command timed out after %s: %w%s", providerCommandTimeout, err, commandOutput(stderr))
 		}
 		return FileConfig{}, fmt.Errorf("provider command failed: %w%s", err, commandOutput(stderr))
 	}
@@ -40,20 +45,111 @@ func LoadProviderCommand(command string) (FileConfig, error) {
 
 var errProviderCommandTimeout = errors.New("provider command timeout")
 
+// providerCommandDrainGrace bounds how long the timeout path waits for cmd.Wait
+// to return after the process tree has been terminated. Terminate normally makes
+// that immediate; this exists so a descendant that refuses to die cannot extend
+// a call whose deadline has already expired.
+const providerCommandDrainGrace = 2 * time.Second
+
+// syncBuffer serialises the command's I/O pump against reads from this package.
+// On the timeout path runProviderCommand can return while cmd.Wait is still
+// draining, so the buffers outlive the call and a plain bytes.Buffer would be
+// read and written concurrently.
+type syncBuffer struct {
+	mu  sync.Mutex
+	buf bytes.Buffer
+}
+
+func (b *syncBuffer) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.buf.Write(p)
+}
+
+// snapshot copies out what has been written so far. A copy rather than the
+// buffer's own slice, which the pump may still append to.
+func (b *syncBuffer) snapshot() []byte {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return append([]byte(nil), b.buf.Bytes()...)
+}
+
+// awaitWithGrace waits for done and gives up after grace, reporting whether the
+// wait completed rather than expired.
+//
+// Split out from the timeout path so the bound can be asserted on its own. In
+// place it is effectively untestable: Terminate kills the tree, so Wait returns
+// promptly and an unbounded receive would behave identically in every test that
+// is not pathological, which is exactly how the old unbounded receive survived.
+func awaitWithGrace(done <-chan error, grace time.Duration) bool {
+	timer := time.NewTimer(grace)
+	defer timer.Stop()
+	select {
+	case <-done:
+		return true
+	case <-timer.C:
+		return false
+	}
+}
+
+// runProviderCommand runs a provider command under a real deadline.
+//
+// "Real" is the point: timeout used to bound only the wait, so the call could
+// take arbitrarily longer than it. Two phases sat outside it. Starting the
+// process ran before the clock began, and on Windows that is CREATE_SUSPENDED,
+// job-object creation and assignment, and a system-wide thread snapshot, none of
+// which is fast on a cold or contended machine. Draining after termination was
+// unbounded as well. Both are inside the budget now, so the deadline is an upper
+// bound rather than a floor.
 func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, error) {
 	cmd := shellCommand(command)
 	// Bound the time Wait spends draining I/O pipes after the process exits,
 	// in case an orphaned descendant still holds the write ends open.
 	cmd.WaitDelay = time.Second
 
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
+	stdout := &syncBuffer{}
+	stderr := &syncBuffer{}
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
 
-	proc, err := startCommandProcess(cmd)
-	if err != nil {
-		return nil, nil, err
+	// Armed before the process is started, so process creation is inside the
+	// budget rather than ahead of it.
+	deadline := time.NewTimer(timeout)
+	defer deadline.Stop()
+
+	type startResult struct {
+		proc *commandProcess
+		err  error
+	}
+	// Buffered, so the goroutine can finish and exit even when nobody is left
+	// waiting on it because the deadline already fired.
+	started := make(chan startResult, 1)
+	go func() {
+		proc, err := startCommandProcess(cmd)
+		started <- startResult{proc: proc, err: err}
+	}()
+
+	var proc *commandProcess
+	select {
+	case result := <-started:
+		if result.err != nil {
+			return stdout.snapshot(), stderr.snapshot(), result.err
+		}
+		proc = result.proc
+	case <-deadline.C:
+		// The start is still in flight and will eventually hand back whatever it
+		// created, so cleanup is handed off rather than abandoned. Wait reaps the
+		// process; without it a started-but-orphaned command becomes a zombie.
+		go func() {
+			result := <-started
+			if result.err != nil {
+				return
+			}
+			result.proc.Terminate()
+			_ = cmd.Wait()
+			result.proc.Close()
+		}()
+		return stdout.snapshot(), stderr.snapshot(), fmt.Errorf("%w: starting the command exceeded %s", errProviderCommandTimeout, timeout)
 	}
 	defer proc.Close()
 
@@ -61,9 +157,6 @@ func runProviderCommand(command string, timeout time.Duration) ([]byte, []byte, 
 	go func() {
 		done <- cmd.Wait()
 	}()
-
-	timer := time.NewTimer(timeout)
-	defer timer.Stop()
 
 	select {
 	case err := <-done:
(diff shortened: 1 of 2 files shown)


gnanam1990 (approved):
## Verdict

Approve.

Reviewed at `945eb06d2b92`, base `5d1869e`, re-confirmed against the live head before posting.

This is a good change and an unusually honest one. The section on discarding your first set of tests is the most valuable part of the description, and I want to record that I checked those claims rather than taking them.

**All three mutation claims hold, verified independently.**

Making the drain unbounded again — replacing the `select` in `awaitWithGrace` with a bare `<-done` — hangs `TestAwaitWithGraceGivesUp` to the test timeout: `panic: test timed out after 15s / TestAwaitWithGraceGivesUp (15s)`.

Removing both locks from `syncBuffer` trips `-race` on `TestSyncBufferAllowsConcurrentWriteAndSnapshot`, twice.

Arming the deadline after the start instead of before fails `TestRunProviderCommandCountsProcessStartAgainstTheDeadline` with exactly the message you predicted: `error = "provider command timeout: the command did not finish within 1ns", want the start phase to be the one that exceeded the deadline`. That assertion is well-built — it discriminates on *which phase* reported, not merely that something timed out, which is what makes it bind.

**The bound is real, measured on a platform you did not test.** On darwin/arm64, driving `runProviderCommand` with `sleep 30 & echo hi` — a background descendant holding the inherited pipes, which is the shape that made the wait-only bound overshoot:

```
budget=200ms  elapsed=202ms   overshoot=2ms
budget=1s     elapsed=1.002s  overshoot=2ms
```

**One honest correction to the scope of my verification.** I could not reproduce the pathological overshoot on macOS. The same scenario on base gives `overshoot=10ms` at both budgets, not 19.7s. That is consistent with your diagnosis — `CREATE_SUSPENDED`, job-object assignment and the thread snapshot are the expensive phases and none of them exists on POSIX — but it means the headline measurement rests on your Windows numbers and the passing windows-latest job, not on anything I ran. A reader should not take my approval as independent confirmation of the 19.7s/106s figures.

What I can confirm on darwin is the diagnostics half, and it is a clear improvement: base returns a bare `provider command timeout`, this returns `provider command timeout: the command did not finish within 200ms` and carries the redacted output through.

**Two non-blocking observations.**

`awaitWithGrace` returns a bool that the caller discards, so the error text is the same whether the drain completed or gave up. Those are different situations — the second means a tree survived `Terminate` — and the return value is already there. Saying so in the message would cost a line and would tell an operator something real.

On POSIX the bounded drain is followed by `defer proc.Close()`, which calls `p.anchor.Wait()` unbounded (`internal/config/process_posix.go`). In practice `Terminate` SIGKILLs the whole group including the anchor, so it returns immediately, and I could not make it hang. But it is an unbounded wait sitting directly after a bound you just added deliberately, and it would be worth either bounding it too or noting why it cannot block.

**On the class, since this file is load-bearing.** `internal/config/command.go` was already the one place in the tree that handles `exec.ErrWaitDelay` explicitly rather than collapsing it, and this makes it stronger still. It remains one site: `exec.Cmd` construction without `WaitDelay`, a process group and a tree-killing `Cancel` is repeated across roughly forty files, and the shared construction point at `internal/sandbox/runner.go:103` is unhardened, so every sandboxed command inherits the older shape. Not this PR's job — #811 is scoped to the provider command and the fix matches the issue — but this file is now the natural reference implementation to extract from, and I would rather say that than have the pattern copied by hand a fifth time.

**Process note.** `#811` carries no `issue-approved` label, unlike `#534` and `#447` on your other pull requests. You are a collaborator so this is not a gate, but the inconsistency is worth tidying so the label stays meaningful as a signal.

**Verification.** On macOS (darwin/arm64): `gofmt -l` clean, `go vet ./...` clean, `git diff --check` clean, `./internal/config/...` passing, `-race -count=10` clean on the new tests. All nine checks pass on the live head, including windows-latest.

**Limitations.** macOS only. The Windows phases this change exists for — `CREATE_SUSPENDED`, job-object creation and assignment, `resumeMainThread`'s thread snapshot — were read, not executed.

anandh8x (approved):
## Review at 945eb06

PR #813 — make the provider-command timeout an upper bound (#811). 2 files, +115/-16, single commit.

### Verdict: **approve.** The fix is correct, the tests bind (verified by mutation), and the diagnostics are a real improvement.

### The problem

`providerCommandTimeout` bounded only the wait, so the call could take arbitrarily longer than the 5s its error text promised. Two phases sat outside the deadline: process creation (before the clock started) and the post-termination drain (unbounded bare receive). On Windows, the start phase is `CREATE_SUSPENDED` + job-object creation + a system-wide thread snapshot — not fast on a cold/AV-heavy machine. The PR body measured 19.7s and 106s against a 5s budget.

### What's good

- **Deadline armed before the start.** The start races the deadline in a goroutine; if the deadline fires first, cleanup is handed off (the goroutine reaps the process when it eventually lands). `TestRunProviderCommandCountsProcessStartAgainstTheDeadline` uses a 1ns budget so the only phase that can exceed it is the start — the error says "starting the command" not "did not finish", which discriminates the two orderings.
- **Bounded drain.** `awaitWithGrace(done, providerCommandDrainGrace)` replaces the bare `<-done`. `TestAwaitWithGraceGivesUp` uses a channel that never fires — making the drain unbounded again hangs this test to the test timeout.
- **`syncBuffer` for the timeout path.** The I/O pump keeps writing after the caller has returned, so a plain `bytes.Buffer` would race. `snapshot()` copies under the mutex. `TestSyncBufferAllowsConcurrentWriteAndSnapshot` overlaps 2000 writes and 2000 snapshots; removing the lock trips `-race`. Verified clean under `-race` in my run.
- **Diagnostics.** The timeout error now names the phase and includes redacted stderr. `TestLoadProviderCommandTimeoutReportsDiagnostics` checks the phase name and `errors.Is` against the sentinel. The old bare string is gone.
- **`cmd.WaitDelay = time.Second`** bounds how long `cmd.Wait` spends draining I/O pipes after the process exits, in case an orphaned descendant holds the write ends open. This is the `exec.ErrWaitDelay` path, which the code already handled but now can't hang either.
- **Tests are honest about what they pin.** The PR body has a section titled "the first version of them was worthless" explaining why whole-call tests don't bind the drain or the race, and which extracted-helper tests do. gnanam independently verified all three mutation claims. The end-to-end duration test is kept as a regression guard with a comment saying it does NOT pin the drain bound.

### Verification

- `go vet ./internal/config/...` — clean
- `go build ./internal/config/...` — clean
- Full `internal/config` package — pass (14.3s, the 5s timeouts are real)
- `-race` on the new tests — clean
- All 12 PR-related tests pass

### gnanam's non-blocking observations (noting, not blocking)

1. **`awaitWithGrace` returns a bool the caller discards.** The error text is the same whether the drain completed or gave up. The second means a tree survived `Terminate` — worth saying in the message. One line.
2. **`defer proc.Close()` calls `p.anchor.Wait()` unbounded on POSIX.** In practice `Terminate` SIGKILLs the group including the anchor so it returns immediately, but it's an unbounded wait sitting directly after a deliberately bounded one. Worth bounding or noting why it can't block.

Both are real but minor. Neither changes the contract this PR ships.

### Verdict

**Approve.** The timeout is now an upper bound, the tests pin the two phases that were outside it, the `syncBuffer` makes the early-return path race-safe, and the diagnostics are a clear improvement. gnanam's two non-blocking observations are worth follow-ups but don't change the verdict.

Vasanthdev2004:
@kevincodex1 ready to merge when you are. Three approvals on the current head (`945eb06d`), 9 of 9 green, no open findings.

Fixes #811. `providerCommandTimeout` was a floor rather than a bound: process creation ran before the timer was armed, and the drain after `Terminate()` was unbounded. I measured `LoadProviderCommand` taking 19.7s and then 106s against a 5s timeout while reproducing something else. It sits on the startup path, so a user with a slow provider command on a cold or AV-heavy Windows box waits far longer than the error text promises, then gets no diagnostics, because the timeout branch discarded stderr.

One behaviour change worth your eye before it lands: the timeout error text now names the phase that ran out and includes the command's redacted stderr. Nothing in the tree matches the old exact string, `errors.Is` and a `timed out` substring both still hold, but it is user-visible output.

No caveats beyond that. Test-and-one-file change, no unexercised paths.

Vasanthdev2004:
@kevincodex1 nudging this one specifically, because it is the only PR of mine that is actually mergeable right now. Still `945eb06d`, still three approvals, still CLEAN, nothing has moved since I last posted.

For contrast, so you can set the rest aside for the moment: #808 needs a CodeRabbit re-run before the ruleset will pass it, #812 is stacked behind #808, and #814 and #816 are waiting on @gnanam1990 to re-approve after my own follow up pushes dismissed them. This is the one with nothing left in front of it.

The only thing on it I would still want you to see before merging is the user-visible change I flagged earlier: the timeout error text now names the phase that ran out and carries the command's redacted stderr. Nothing in the tree matches the old exact string.

anandh8x (approved):
Reviewed the latest head 945eb06. The deadline now covers process startup and the post-termination drain, the early-return output path is synchronized, and the tests bind those three failure modes. The config package passes under the race detector locally. Good to merge.

kevincodex1 (approved):

Issue #815, opened 2026-07-27
Sandbox can read SSH private keys, the GPG keyring, and git credential stores


Vasanthdev2004:
## Summary

The credential deny list covers cloud CLI and tool credentials, and since #801 it covers Zero's own store, but it does not cover SSH private keys, the GPG secret keyring, or either git credential store. On Linux, where the sandbox uses a read-all filesystem posture with explicit deny rules, a sandboxed command can read all of them.

Split out of #675, which named Zero's own credential files and is now fixed. These were listed there as out of scope and should not be assumed closed with it.

## Evidence

Driving `credentialDenyReadPathsForEnvironment` against a synthetic home with the files materialised, on `main` at `81a5e6cb`. The builder filters to paths that exist, so they have to be real:

```
denied=false  ~/.ssh/id_rsa
denied=false  ~/.ssh/config
denied=false  ~/.gnupg/secring.gpg
denied=false  ~/.git-credentials
denied=false  ~/.config/git/credentials
denied=true   ~/.aws/credentials
```

`~/.aws` is included to show the mechanism works and these are simply absent from the list.

## Why this is worth more than the entries already covered

An SSH private key is usually a broader credential than a provider API key. It typically grants push access to every repository the user can reach, and on many setups shell access to servers as well. The same is true of the GPG secret keyring for signing identity. Both are readable today by any command the agent runs.

## The git case has a wrinkle worth naming

`userGitConfigReadPaths` (`internal/sandbox/profile.go:286`) already grants only the two git config FILES, and its comment says why: "granting only these files avoids exposing the surrounding configuration directory". That reasoning is about `~/.config/git`, whose XDG credential store lives at `~/.config/git/credentials`.

So macOS, which allow-lists reads, deliberately avoids exposing that directory. Linux, which denies instead, has no rule for the credential file inside it. The intent exists in the codebase; only the Linux half is missing.

## Relevant code

- `internal/sandbox/profile.go`, `credentialDenyReadPathsForEnvironment` (the candidate list)
- `internal/sandbox/profile.go:286`, `userGitConfigReadPaths` (the comment describing the intent)

## Expected behavior

A sandboxed command cannot read SSH private keys, the GPG secret keyring, or either git credential store, on the same terms as the entries already denied.

## Decision needed before implementing

[...]

Issue #662, opened 2026-07-12
Windows sandbox does not deny reads of cloud credential stores


euxaristia:
## Context

PR #660 added default deny-read rules so sandboxed commands cannot read well-known credential stores (~/.aws, ~/.config/gcloud, ~/.azure, and the file GOOGLE_APPLICATION_CREDENTIALS points to). The rules are injected at permission profile construction (credentialDenyReadPaths in internal/sandbox/profile.go) and enforced by the seatbelt backend on macOS (file-read* deny rules) and the Linux helper (unreadable binds).

Windows is deliberately exempted, so on Windows a sandboxed command can still read those credential files under the read-all workspace posture.

## Why Windows was skipped

Windows deny-read is implemented via ACL deny entries applied against capability SIDs (planWindowsDenyReadPaths / windowsReadDenyCapabilitySIDs in internal/sandbox/windows_acl.go). Making the profile DenyRead list non-empty by default has two side effects:

1. windowsReadDenyCapabilitySIDs stops returning nil, which moves the runner off the WRITE_RESTRICTED token path that #658 introduces for the no-deny-read case. The unelevated tier depends on that path; the full restricted-token path has a history of commands silently exiting 1 when unelevated.
2. The ACL deny entries mutate filesystem ACLs on the user's real credential directories, which is a heavier and more persistent intervention than the process-scoped seatbelt/bwrap rules on the other platforms, and needs elevation in some configurations.

## What needs deciding

- Whether default credential deny-read on Windows should wait for the token/ACL model to settle (#640, #658 are in flight), or whether a scoped mechanism exists that does not flip the token selection.
- Whether ACL-based deny on real user directories is acceptable at all as a default, given it outlives the sandboxed process.
- Whether the interim mitigation (env scrubbing of provider keys and GOOGLE_APPLICATION_CREDENTIALS, which does apply on Windows) is enough until then.

## Pointer

The exemption and rationale are documented on credentialDenyReadPaths in internal/sandbox/profile.go ("Windows is skipped" comment). Remove the runtime.GOOS gate there once a Windows mechanism lands.

Vasanthdev2004:
Writing up what I found while reviewing #681 and #640, because this issue is blocking both of them and the reason it has been hard to close is not obvious from the outside.

## Where it stands today

`credentialDenyReadPaths` (internal/sandbox/profile.go) opens with:

```go
if runtime.GOOS == "windows" {
    return nil
}
```

[...]

PR #816, opened 2026-07-27
fix(sandbox): deny reads of git's credential stores (#815)

Vasanthdev2004:
Takes option 1 from #815: the git credential stores only.

## What

`~/.git-credentials` and `~/.config/git/credentials` are git's credential-store backend. They hold host passwords and personal access tokens in cleartext, and neither was on the deny list, so any command the agent ran could read them.

Measured on `main` before the change, with the files materialised, since the builder only emits paths that exist:

```
denied=false  ~/.git-credentials
denied=false  ~/.config/git/credentials
denied=true   ~/.aws/credentials     (included to show the mechanism works)
```

## Why only these two, and not the rest of #815

#815 also covers SSH private keys and the GPG keyring. Those are a genuinely harder trade and this deliberately does not attempt them:

- Denying `~/.ssh` stops a sandboxed `git push` over SSH from working. That is a functional cost, not a free win.
- Excluding `~/.ssh` is not even sufficient. An `IdentityFile` directive, or an `Include` chain, can point key material anywhere, so a directory rule leaves the key readable exactly when a careful user has moved it.

The credential stores have neither problem. git reads them to authenticate, not to know who the user is, so a sandboxed git keeps working and simply cannot act as the user. That is the behaviour you want from a sandbox.

## One asymmetry worth explaining

The XDG entry denies the FILE, `~/.config/git/credentials`, not the directory around it. That is the opposite of the choice made for `~/.config/zero` in #801, and it is deliberate.

`~/.config/git` also holds the global git config, and `userGitConfigReadPaths` grants that on purpose so a sandboxed git can read `user.name` and aliases rather than failing with "unable to access". Denying the directory would take that away. Its comment already says the surrounding directory should not be exposed, which is the same instinct, applied from the allow side on macOS.

## Verification

`gofmt`, `go vet`, `GOOS=windows go vet`, and builds for linux, darwin and windows are clean. `internal/sandbox` passes.

Three tests, each mutation-checked rather than merely written:

- dropping `~/.git-credentials` fails the coverage test
- dropping the XDG entry fails it too
- widening the XDG entry to the whole directory fails the assertion that the git config beside it stays readable, which is the regression this scoping exists to prevent

The third is the one worth having. It is what stops someone "simplifying" this later into a directory deny and quietly breaking git identity inside the sandbox.

## Scope

Linux in practice, like the rest of the list: it is skipped on Windows by design, and macOS allow-lists reads. #815 stays open for the SSH and GPG half, which still needs the posture decision.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential protection by blocking access to Git credential helper store files in sandboxed environments (including both legacy and XDG locations).
  * Deny rules still respect explicit allow-read permissions, so explicitly permitted paths are not blocked.
* **Tests**
  * Added coverage to verify credential store deny-path behavior and precedence with allow-read settings.

### DIFF
--- a/internal/sandbox/profile.go
+++ b/internal/sandbox/profile.go
@@ -212,6 +212,14 @@ func credentialDenyReadPathsForEnvironment(env credentialPathEnvironment, allowR
 		candidates = append(candidates,
 			filepath.Join(home, ".aws"),
 			filepath.Join(home, ".azure"),
+			// git's credential store backend, which holds host passwords and
+			// personal access tokens in cleartext. Denied rather than the whole
+			// of ~/.ssh, because these cost nothing functionally: git reads them
+			// through a credential helper for authentication, not for identity,
+			// so a sandboxed git still works and simply cannot authenticate as
+			// the user. SSH key material is a harder trade and is tracked
+			// separately (#815).
+			filepath.Join(home, ".git-credentials"),
 		)
 	}
 	npmUserConfig := strings.TrimSpace(env.NPMUserConfig)
@@ -233,6 +241,12 @@ func credentialDenyReadPathsForEnvironment(env credentialPathEnvironment, allowR
 	if configHome != "" {
 		candidates = append(candidates,
 			filepath.Join(configHome, "zero"),
+			// The XDG location of the same git credential store. Denying the file
+			// rather than the directory is deliberate: ~/.config/git also holds
+			// the global git config, which userGitConfigReadPaths grants on
+			// purpose so a sandboxed git can read user.name and aliases instead
+			// of failing outright.
+			filepath.Join(configHome, "git", "credentials"),
 		)
 	}
 	cloudSDKConfig := strings.TrimSpace(env.CloudSDKConfig)
--- a/internal/sandbox/profile_test.go
+++ b/internal/sandbox/profile_test.go
@@ -1,0 +1,79 @@
+package sandbox
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// git's credential store holds host passwords and personal access tokens in
+// cleartext, in one of two locations depending on whether the user is on the
+// XDG layout. Neither was denied, so a sandboxed command could read them
+// (#815).
+//
+// Scoped to the credential files on purpose. Denying ~/.ssh as well would stop
+// a sandboxed git push over SSH from working, which is a functional trade that
+// issue tracks separately; these two cost nothing, because git reads them for
+// authentication rather than identity.
+func TestCredentialDenyReadPathsCoversGitCredentialStores(t *testing.T) {
+	home := t.TempDir()
+	configHome := filepath.Join(home, ".config")
+	if err := os.MkdirAll(filepath.Join(configHome, "git"), 0o700); err != nil {
+		t.Fatal(err)
+	}
+	// The builder only emits paths that exist, so these have to be real.
+	gitCredentials := filepath.Join(home, ".git-credentials")
+	xdgCredentials := filepath.Join(configHome, "git", "credentials")
+	gitConfig := filepath.Join(configHome, "git", "config")
+	for _, path := range []string{gitCredentials, xdgCredentials, gitConfig} {
+		if err := os.WriteFile(path, []byte("x"), 0o600); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	denied := credentialDenyReadPathsForEnvironment(credentialPathEnvironment{
+		Home:       home,
+		ConfigHome: configHome,
+	}, nil)
+	covered := func(target string) bool {
+		norm := normalizeProfilePath(target)
+		for _, entry := range denied {
+			if entry == norm || pathWithinRoot(entry, norm) {
+				return true
+			}
+		}
+		return false
+	}
+
+	if !covered(gitCredentials) {
+		t.Fatalf("~/.git-credentials is readable; deny list = %v", denied)
+	}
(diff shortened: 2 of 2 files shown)


gnanam1990 (dismissed):
## Verdict

Approve.

Reviewed at `6140e840945e`, base `81a5e6c`, re-confirmed against the live head before posting. Up to date with main.

Declaring an interest first: this closes a gap I raised as non-blocking on #801, so I am reviewing a fix to my own finding. That does not change whether the code holds up, but you should weigh my approval accordingly.

**The fix is verified, not taken on trust.** Both paths now come back denied, and both mutations bind: dropping `~/.git-credentials` fails `TestCredentialDenyReadPathsCoversGitCredentialStores`, and dropping the XDG entry fails it too. `TestGitCredentialDenyYieldsExplicitAllowRead` covers the opt-out, which matters because it is the behaviour the existing entries already have and the one a user relies on to get a path back.

**The asymmetry you explain is correct, and I checked it rather than accepting the reasoning.** Denying the file `~/.config/git/credentials` rather than the directory is right because `userGitConfigReadPaths` grants `~/.gitconfig` and `~/.config/git/config` — the config *files* — so a sibling-file deny does not collide with that grant, where a directory deny would have taken it away. The contrast with `~/.config/zero` in #801 is genuine and the reason is the one you give.

**Scoping to the credential stores and leaving SSH alone is the right call**, and your second argument for it is stronger than the one #815 makes. "Denying `~/.ssh` costs a sandboxed `git push`" is a trade. "Excluding `~/.ssh` is not even sufficient, because an `IdentityFile` directive or an `Include` chain can put key material anywhere" is a reason the directory rule is the wrong shape, which is a different and better claim.

**One non-blocking observation, because that argument cuts both ways.** git's credential store is not fixed to these two locations either: `credential.helper = "store --file=/anywhere"` puts it wherever the user says, exactly as `IdentityFile` does for SSH keys. So this covers the defaults and not the mechanism. I do not think that argues against merging — covering the default locations is worth doing and the alternative is unbounded — but it is worth stating in the issue so #815 does not read as "SSH is hard, git is solved". The honest framing is that both are default-only, and neither can be closed by enumerating paths.

**Verification.** On macOS (darwin/arm64): `gofmt -l`, `go vet ./...` and `git diff --check` clean; `./internal/sandbox/...` passing; builds clean for linux/amd64 and windows/amd64.

**Limitations.** The deny list returns early on Windows and is a no-op on macOS, where seatbelt reads are allow-listed, so this only changes Linux behaviour — and I exercised it through the profile builder, never against a live Linux sandbox. That gap is now standing across #801, #815 and this: nobody on any of these threads has run the enforcement path.

**Process.** The body has no `Fixes #815`; the reference is in the title only, so tooling reads this as unparented. #815 also carries no `issue-approved` label yet. Neither affects the code, and you are a collaborator so the gate is process rather than quality, but three of your open PRs now reference issues in a form nothing can see (#808 "foundation for #662", #813, and this).

Vasanthdev2004:
Self review before this goes to anyone else. One of the two tests was not testing anything.

`TestGitCredentialDenyYieldsToExplicitAllowRead` only checked that the credential store was absent from the deny list once an explicit grant covered it. A deny rule that was never added satisfies that equally well, so with both production lines reverted the test still passed. The coverage test correctly failed, so only this one was affected.

Now asserted from both sides: denied with no grant, absent with one. The first leg is what establishes there is anything for the grant to override. Verified it fails under the same revert.

Two limits worth stating rather than leaving implied:

- A store relocated with `credential.helper 'store --file=<path>'` is not covered. Reaching it means parsing git config for arbitrary paths, which is the same relocation problem the SSH half of #815 runs into, so I would rather handle both there than half solve it here.
- The builder only emits paths that exist when the profile is built, which is true of every entry including `~/.aws`. A store created after the sandbox starts is not retroactively denied. Pre-existing property, noting it so it is not mistaken for a promise this PR makes.

Vasanthdev2004:
@kevincodex1 ready to merge from my side. All nine checks green on `afbe47f3` and CodeRabbit has approved the current head.

Same situation as #814 and it is my doing: @gnanam1990 approved at `6140e840` earlier today and my follow up push dismissed that approval, so GitHub is holding on zero human approvals. Re-requested.

What the push fixed: on a second pass over my own work, `TestGitCredentialDenyYieldsToExplicitAllowRead` was one sided. It only checked the credential store was absent from the deny list once an explicit grant covered it, which a deny rule that was never added satisfies just as well, so it passed with both production lines reverted. Now it asserts both legs, denied with no grant and absent with one, and fails under that same revert. The coverage test was sound throughout.

gnanam, the delta since your approval is confined to `profile_test.go`, 19 added and 5 removed. The production change and the two limits noted in the PR body are untouched.

Vasanthdev2004:
@gnanam1990 same ask as #814, and same cause: my follow up push dismissed your approval at `6140e840`.

Delta is `6140e840..afbe47f3`, confined to `profile_test.go`, 19 added and 5 removed. The production change and the two limits in the PR body are untouched.

To check the fix rather than take my word for it, revert both deny lines in `profile.go`:

```
go test ./internal/sandbox/ -run 'TestCredentialDenyReadPathsCoversGitCredentialStores|TestGitCredentialDenyYieldsToExplicitAllowRead' -count=1
```

On the old head the coverage test failed and the allow read test passed, which is what gave it away: a deny rule that was never added satisfies "absent when granted" just as well as one that was. On the current head both fail under that revert.

gnanam1990 (approved):

PR #685, opened 2026-07-14
fix(sandbox): protect daemon token file

PierrunoYT:
## Summary

Protect the remote daemon's file-backed bearer credential from sandboxed commands, in-process file tools, and pre-mutation session checkpoints.

Fixes #677.

The configured filename is treated as literal pathname data, including whitespace. Both its configured spelling and startup-resolved target remain protected, and the daemon passes the opened file's stable identity to workers. File tools check the actual opened object, so permission grants, hard-link aliases, and rotation do not bypass the credential boundary. Reads and mutations use rooted filesystem operations.

## Shell behavior

Linux permits the existing safe placement on a filesystem separate from shell-writable roots, provided the credential has no additional hard links and its protected pathname still names the startup object. Rotation that renames the live token and replaces or removes its pathname now fails admission; restoring the original object restores admission. This sequential lifecycle check does not introduce cross-process locking around rotation and sandbox startup.

macOS and Windows require the inline `ZERO_DAEMON_REMOTE_TOKEN` for sandboxed shell execution with remote authentication. In-process file tools retain their file-token protections. The existing CLI help documents the Linux and macOS restrictions.

## Latest review fixes

- Check Linux shell admission against the captured startup file identity before applying the link-count/filesystem exception.
- Exclude protected credentials from `write_file` and `edit_file` checkpoint targets. The checkpoint reader independently opens through the workspace root and checks the same handle it reads, preventing a rotated alias from being copied into an unprotected checkpoint blob.
- Carry disambiguated no-prefix rename/copy paths into subsequent hunk-header validation. Spaced filenames with content changes work while contradictory headers remain rejected.
- Isolate the whitespace planner test's home, configuration, cache, credential overrides, and token handoff markers.

The branch contains current main (`c1937dfa`).

## Regression evidence

New tests were run with the fixes and against the original implementation in an isolated Linux copy. Without the fixes:

- Linux rotation admission returned no error instead of refusing the replaced startup object.
- Checkpoints captured the protected alias into a content-addressed blob; both the exec recorder and TUI callback tests failed before their denied mutation.
- Rename and copy hunks failed with `---/+++ paths disagree with diff --git paths from line 1`.
- The original whitespace planner test created the caller's `.config/zero`; the fixed test left it absent.

Coverage includes missing/restored token paths, rotated hard-link aliases, both mutation callbacks, reading persisted session files through `read_file`, and ordinary checkpoint/rewind tests in the existing suite.

## Validation

[...]

Issue #817, opened 2026-07-28
Plan 9 is treated as a review criterion but has never been a supported target


Vasanthdev2004:
## Summary

Plan 9 has never compiled in this repository, is not built by CI, is not a release target, and is used inside our own tests as the canonical example of an unsupported platform. Despite that, it is being applied as a review criterion: #750 was blocked on "this regresses Plan 9 compilation," which led to a plan9 shim being added to that PR. Those are currently the only plan9 build constraints in the entire codebase.

This is not a request to make Plan 9 work. It is a request to write down what the supported targets are, so reviews stop gating on a platform the project does not build for.

## Evidence

All of the following is against `origin/main` at `81a5e6cb`, with no PR branches involved.

A whole-tree build fails in six first-party packages:

```
GOOS=plan9 GOARCH=amd64 go build ./...

internal/fsutil/rename.go:38          undefined: syscall.Errno
internal/installtxn/lock_unix.go:17   undefined: unix.Flock
internal/execution/process_unix.go:22 cmd.SysProcAttr.Setpgid undefined
internal/localcontrol/app_launcher_unix.go:14  cmd.SysProcAttr.Setsid undefined
internal/dictation/owner_unix.go:16   undefined: syscall.Stat_t
```

It also fails in `charm.land/bubbletea/v2`, which is the TUI itself:

```
bubbletea/v2@v2.0.8/tea.go:690  p.listenForResize undefined
bubbletea/v2@v2.0.8/tty.go:18   undefined: suspendProcess
```

That last one matters most. Even if every first-party package above were fixed, the build would still fail, because our TUI dependency does not support Plan 9. Plan 9 support is therefore not reachable through changes to this repository alone.

Meanwhile the project's own configuration says the same thing in three other places:

- `release-artifacts.yml` targets exactly five platforms, with the comment "These five cover the platforms install.sh / install.ps1 / the npm wrapper resolve": linux-x64, linux-arm64, macos-arm64, macos-x64, windows-x64.
- CI does not build any Plan 9 target.
- Every reference to plan9 in first-party source is a test fixture standing in for an unsupported OS, for example `architecture_baseline_test.go` has `{name: "unsupported", goos: "plan9", want: BackendUnavailable}`, and `manager_test.go` does the same.

## Why raise it

[...]

PR #818, opened 2026-07-28
fix(sandbox): scrub the daemon token file pointer (#677)

### LINKED ISSUE #677: ZERO_DAEMON_REMOTE_TOKEN_FILE leaks the daemon bearer token into sandboxed commands
## Summary

PR #660 scrubs `ZERO_DAEMON_REMOTE_TOKEN`, but leaves `ZERO_DAEMON_REMOTE_TOKEN_FILE` in the inherited sandbox environment. The read-all sandbox posture lets the command read the referenced file.

## Relevant code

- `internal/sandbox/runner.go`: `scrubSensitiveEnv` includes `ZERO_DAEMON_REMOTE_TOKEN` but not `ZERO_DAEMON_REMOTE_TOKEN_FILE`.
- `internal/daemon/remote/auth.go`: `TokenFromEnv` reads the bearer token from that path.

Observed at commit `6fc1220`.

## Reproduction

1. Put a daemon bearer token in a file outside the workspace.
2. Set `ZERO_DAEMON_REMOTE_TOKEN_FILE` to its absolute path.
3. Launch a sandboxed command through Zero.
4. The command can read the pointer and token file.

## Suggested fix

Remove the pointer from the sandbox environment and add its resolved target to `DenyRead`, analogous to `GOOGLE_APPLICATION_CREDENTIALS`. Test both env scrubbing and filesystem denial.

Vasanthdev2004:
Fixes #677.

`scrubSensitiveEnv` removed `ZERO_DAEMON_REMOTE_TOKEN` but not `ZERO_DAEMON_REMOTE_TOKEN_FILE`. `TokenFromEnv` (`internal/daemon/remote/auth.go:72-86`) accepts either form, so a sandboxed command still inherited the path to the bridge token, and since the default posture is read-all it could then read the token out of that file.

## Why the one line is the whole fix

I considered also adding the target to the credential deny-read list, and decided against it. The path comes from this variable alone and `TokenFromEnv` has no default location to fall back on, so a command that never sees the variable has nothing to guess at. Adding a deny rule for an arbitrary env-supplied path would also collide with #816, which is currently in flight in the same function, for no extra protection.

This mirrors the existing treatment of `GOOGLE_APPLICATION_CREDENTIALS`, which is scrubbed for exactly this reason. The comment there already spells out the rule; this variable simply matched the description and was missed.

## Verification

Extended the existing scrub table rather than adding a new test, because `TestScrubSensitiveEnv` compares the surviving environment exactly, so a missed key fails the assertion by construction. Both forms of the token now go in and neither is expected out.

Dropping the production line reproduces the leak:

```
scrubSensitiveEnv() = map[... ZERO_DAEMON_REMOTE_TOKEN_FILE:/home/user/.zero/remote-token ...]
                want map[AWS_PROFILE:staging PATH:/usr/bin SAFE_VAR:hello ZERO_OAUTH_CLIENT_SECRET:not-a-provider-secret]
```

`go test ./internal/sandbox/` passes in full, `gofmt` and `go vet` clean.

## Scope

Two files, twelve added lines, no behaviour change beyond the scrub itself. Keeping it deliberately small: #685 covered adjacent ground and grew to fifteen files, which is a large part of why it stalled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security when running sandboxed commands by removing both inline daemon tokens and token file paths from inherited environments.
  * Prevents sandboxed processes from accessing sensitive daemon credentials through environment variables.

### DIFF
--- a/internal/sandbox/runner.go
+++ b/internal/sandbox/runner.go
@@ -1027,6 +1027,13 @@ func scrubSensitiveEnv(env []string, additionalKeys ...string) []string {
 		"GH_TOKEN",
 		"ZERO_WEBSEARCH_API_KEY",
 		"ZERO_DAEMON_REMOTE_TOKEN",
+		// The file form of the same bridge token. TokenFromEnv accepts either,
+		// so scrubbing only the inline variable left the pointer readable, and
+		// the default sandbox posture is read-all. There is no fallback
+		// location to guess at, the path comes from this variable alone, so
+		// removing it closes the leak rather than half of it. Same reasoning as
+		// GOOGLE_APPLICATION_CREDENTIALS below.
+		"ZERO_DAEMON_REMOTE_TOKEN_FILE",
 	}
 	for _, descriptor := range providercatalog.All() {
 		for _, key := range descriptor.AuthEnvVars {
--- a/internal/sandbox/runner_test.go
+++ b/internal/sandbox/runner_test.go
@@ -699,6 +699,11 @@ func TestScrubSensitiveEnv(t *testing.T) {
 		"XAI_API_KEY=xai-12345",
 		"HUGGINGFACE_API_KEY=hf_12345",
 		"GOOGLE_APPLICATION_CREDENTIALS=/home/user/sa-key.json",
+		// Both forms of the daemon bridge token. The inline one was already
+		// scrubbed; the file form was not, which left a sandboxed command the
+		// path to read it from (#677).
+		"ZERO_DAEMON_REMOTE_TOKEN=bridge-token-inline",
+		"ZERO_DAEMON_REMOTE_TOKEN_FILE=/home/user/.zero/remote-token",
 		"COMPANY_LLM_SECRET=custom-secret",
 		"ZERO_OAUTH_MY_SVC_CLIENT_SECRET=oauth-secret",
 		"zero_oauth_second_client_secret=case-insensitive-secret",


Vasanthdev2004:
@anandh8x @gnanam1990 putting the whole picture in one place rather than commenting on five PRs, since between you the queue clears.

**@anandh8x, three of these need only you.** Each is at two of three with @gnanam1990 and CodeRabbit already in, so a single approval on each takes them to CLEAN:

- #816 `afbe47f3` deny reads of git's credential stores, closes half of #815
- #814 `aa7ce57a` test only, stops `internal/tools` failing on a stock Windows box
- #807 `e62fc079` clipboard image paste, closes #534

Also #750 `c6ad2a43` is waiting on your changes-requested from 21 July about Plan 9. I verified it rather than just asserting: `GOOS=plan9 go build ./internal/providers/providerio/` exits 0. The remaining plan9 failures in the tree are in `internal/fsutil` and `internal/execution` and reproduce identically on `origin/main` with none of my code present, which is what #817 is about.

**Both of you on #818** `49b5a195`, which is new and has nobody on it yet. It is the #677 fix: `scrubSensitiveEnv` dropped `ZERO_DAEMON_REMOTE_TOKEN` but not the `_FILE` form, so a sandboxed command inherited the path to the bridge token. Two files, twelve lines.

**@gnanam1990 on #808** `dbdbda11` when you get a moment. Your approval and anandan's were both dismissed by my own push, sorry, that was the ACE type fix CodeRabbit asked for. The delta from `99fefdc4` is thirteen lines in one test helper and no production code. CodeRabbit has not re-reviewed since the fix landed, so that needs to clear first.

No rush on any of it. Flagging it because six of my eight open PRs are short on approvals rather than short on work, and I would rather ask once than nudge repeatedly.

kevincodex1 (approved):

PR #820, opened 2026-07-29
fix(tui): exclude permission waits from turn time

anandh8x:
## What changed

- track active turn duration separately from time spent waiting for a permission decision
- freeze the live elapsed indicator while the permission prompt is open
- restart quiet-generation tracking after the user resolves the prompt
- add a deterministic regression test covering a 90-second permission wait

## Why

Turn duration previously used wall-clock time from run start to completion. As a result, time spent waiting for the user at a permission prompt was reported as model work, producing misleading `worked for` durations and stale `still generating` warnings after approval or denial.

After this change, the live timer and settled turn duration reflect active execution time rather than user decision time.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- focused regression test with `-race`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- manual termctrl verification using `zero-build` and `zero-dev` with a permission prompt held open for over one minute

## Summary by CodeRabbit

- **Bug Fixes**
  - Permission-wait time is no longer included in active turn elapsed time.
  - Working-status elapsed timers and reported response timing now pause while awaiting approval.
  - “Still generating…” quiet hints no longer appear during permission prompts.
  - Provider inactivity/stream activity resumes correctly after a permission decision.
  - Time-to-first-token (TTFT) is now measured using the same pause-aware elapsed clock.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -5050,9 +5050,18 @@ func (m model) runAgentWithOptions(runID int, runCtx context.Context, prompt str
 		if m.turnTimer != nil {
 			m.turnTimer.start(started)
 		}
-		// firstTokenAt is stamped when the first token (reasoning or text) streams,
-		// so the turn can report time-to-first-token alongside total wall time.
-		var firstTokenAt time.Time
+		// firstTokenElapsed is stamped from the pause-aware turn timer when the
+		// first reasoning or text token streams, so TTFT and total elapsed use
+		// the same clock.
+		var firstTokenElapsed time.Duration
+		firstTokenSeen := false
+		stampFirstToken := func(at time.Time) {
+			if firstTokenSeen {
+				return
+			}
+			firstTokenSeen = true
+			firstTokenElapsed = m.activeTurnElapsedAt(started, at)
+		}
 		toolCalls := 0
 		rows := []transcriptRow{}
 		usageEvents := []zeroruntime.Usage{}
@@ -5168,11 +5177,10 @@ func (m model) runAgentWithOptions(runID int, runCtx context.Context, prompt str
 
 		onText := options.OnText
 		options.OnText = func(delta string) {
-			if firstTokenAt.IsZero() {
-				firstTokenAt = m.now()
-			}
+			now := m.now()
+			stampFirstToken(now)
 			if strings.TrimSpace(reasoningText) != "" {
-				flushReasoning(m.now())
+				flushReasoning(now)
 			}
 			m.sendAgentText(runID, delta)
 			if onText != nil {
@@ -5272,8 +5280,8 @@ func (m model) runAgentWithOptions(runID int, runCtx context.Context, prompt str
 		onReasoning := options.OnReasoning
 		options.OnReasoning = func(delta string) {
 			now := m.now()
-			if firstTokenAt.IsZero() && strings.TrimSpace(delta) != "" {
-				firstTokenAt = now
+			if strings.TrimSpace(delta) != "" {
+				stampFirstToken(now)
 			}
 			if strings.TrimSpace(reasoningText) == "" && strings.TrimSpace(delta) != "" {
 				reasoningStarted = now
@@ -5515,10 +5523,6 @@ func (m model) runAgentWithOptions(runID int, runCtx context.Context, prompt str
 		}
 		flushReasoning(m.now())
 		elapsed := m.activeTurnElapsed(started)
-		ttft := time.Duration(0)
-		if !firstTokenAt.IsZero() {
-			ttft = firstTokenAt.Sub(started)
-		}
 		rows = append(rows, transcriptRow{
 			kind:        rowAssistant,
 			text:        result.FinalAnswer,
@@ -5536,7 +5540,7 @@ func (m model) runAgentWithOptions(runID int, runCtx context.Context, prompt str
 				"content": result.FinalAnswer,
 			},
 		})
-		return agentResponseMsg{runID: runID, rows: rows, usageEvents: usageEvents, usageModelID: usageModelID, sessionEvents: sessionEvents, goalAware: goalAwareRun, turnTools: toolCalls, turnElapsed: elapsed, ttft: ttft}
+		return agentResponseMsg{runID: runID, rows: rows, usageEvents: usageEvents, usageModelID: usageModelID, sessionEvents: sessionEvents, goalAware: goalAwareRun, turnTools: toolCalls, turnElapsed: elapsed, ttft: firstTokenElapsed}
 	}
 }
 
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -221,6 +221,9 @@ type model struct {
 	// renders the live elapsed time from it so a long or stalled turn never looks
 	// like a frozen terminal (for ANY provider, not just slow ones). Zero = idle.
 	turnStartedAt time.Time
+	// turnTimer is shared with the agent command so both the live status and the
+	// settled "worked for" duration exclude time blocked on a user permission.
+	turnTimer *activeTurnTimer
 	// lastCharTime tracks when the last non-Enter key was received, for paste detection.
 	lastCharTime time.Time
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (changes requested):
Good fix and the right thing to exclude. One regression to sort first.

`elapsed` is now pause-adjusted (model.go:5517) but `ttft` is still raw wall clock (model.go:5520). If the first provider turn only emits a tool call, the permission wait lands before the first token, so the wait is subtracted from elapsed and kept in ttft. `/context` then shows something like `0.6s avg (90.2s to first token, 1 turns)`. On main both were raw, so ttft <= elapsed always held; this PR breaks that. Stamping TTFT from the timer at token arrival rather than subtracting `started` at the end would keep them on one clock.

Second thing: the new test does not pin `resume`'s arithmetic. Changing turn_timer.go:55 to `at.Sub(timer.startedAt)` leaves `internal/tui` green. The test's pause starts at the same instant as the turn, so `pausedAt == startedAt` and the two are indistinguishable. Any active work before the prompt would be silently discarded. Worth a case with work either side of the pause.

Last, not blocking: `ask_user` blocks the same way (model.go:5230) with no pause/resume, and `quietGenerationHint` does not check `pendingAskUser`. Pre-existing, so nothing regresses, but the PR body's claim reads as unqualified. Either add the same four lines or say "permission prompts" in the description.

gnanam1990 (approved):
## Verdict

Approve. The fix is correct, both halves of it are pinned by tests that I confirmed can fail, and the two failure modes I went hunting for are genuinely closed rather than accidentally absent.

Reviewed at `0fd0bc5`, base `097c265` (level with `origin/main`). Author has write access, so this is the internal cycle rather than the `issue-approved` gate.

I had started against `12eedef` and was about to file the TTFT inconsistency below as a non-blocking item; `0fd0bc5` fixes exactly that, so it is withdrawn and re-verified as fixed.

## What I verified

- **The elapsed-time fix can fail.** Disabling `pause()` — making it a no-op so a permission wait counts as work again — fails the regression test with the live status line as evidence:

  ```
  --- FAIL: TestPermissionWaitDoesNotCountTowardTurnElapsed
      session_test.go:415: live elapsed advanced during permission wait:
      "⠋ Working  ·  thinking  ·  1m32s  ·  ↑ 13 tok"
  ```

- **The new TTFT fix can fail too.** Reverting `firstTokenElapsed` to wall clock (`at.Sub(started)`) fails the same test on its own assertion:

  ```
  --- FAIL: TestPermissionWaitDoesNotCountTowardTurnElapsed
      session_test.go:435: ttft = 1m35s, want 5s with permission wait excluded
  ```

  That is the right fix. On `12eedef`, `ttft` was still `firstTokenAt.Sub(started)` while `turnElapsed` had become pause-aware, so a turn where the user sat at a prompt before the first token would render a TTFT larger than the total duration next to it. `0fd0bc5` puts both on the same clock and the test now asserts `ttft <= turnElapsed` directly (`session_test.go:437`), which pins the relationship rather than just the two numbers.
- **Overlapping permission prompts cannot corrupt the accounting.** This was my main concern: `pause`/`resume` are boolean rather than counted (`turn_timer.go:33-56`), so two concurrent prompts would let the first `resume` restart the clock while the second is still blocked. It is not reachable — `executeParallelReadBatch` wraps `OnPermissionRequest` in a `callbackMutex` (`internal/agent/parallel_tools.go:165-172`) precisely so two callbacks are never in flight at once, and the comment above it says that is deliberate. Worth knowing that the timer's correctness now rests on that lock, since neither file mentions the other.
- **No race under repetition.** `go test -race ./internal/tui/ -run 'Permission|Turn|Elapsed' -count=10` is clean. The `*activeTurnTimer` is deliberately shared between the Update goroutine (reading `elapsed` in `workingStatusLine`) and the agent goroutine (calling `pause`/`resume` from the permission callback), and every field access goes through the mutex. The new `firstTokenSeen`/`firstTokenElapsed` closure variables are unsynchronised, but they are touched only from the stream-callback path, same as the `firstTokenAt` they replace — no new exposure, and `-race` agrees.
- **`elapsed` is safe at the edges.** It clamps a negative result to zero (`turn_timer.go:72-74`), `resume` only accumulates when `at.After(pausedAt)` (`:52`), and `pause` is idempotent when already paused (`:38`). A backwards clock loses paused time rather than producing a negative or runaway duration.
- **All four `turnElapsed` call sites were converted**, not just the happy path — the error return (`model.go:5501`), the spec-draft error return (`:5511`), the spec-draft success return (`:5514`) and normal completion (`:5525`). That is the "every write path" question and the answer is yes.
- Gauntlet on darwin/arm64 from a non-`/tmp` checkout: `make fmt-check`, `go vet ./...`, `git diff HEAD --check` clean, full `go test ./...` green.
- **Passes executed:** 0–5, 7, 9, 11, 12. Pass 6 was scoped down and 8/10 were light — this diff touches no credentials, no filesystem, no network and no platform-divergent primitive.

## Non-blocking

- **`turnTimer` is initialised twice.** `beginRun` constructs it with `newActiveTurnTimer(m.turnStartedAt)` (`model.go:4869`) and `runAgentWithOptions` immediately calls `start(started)` (`:5050`) with a second timestamp. Harmless — `start` resets cleanly — but a reader has to check both to learn which one wins.
- **`activeTurnElapsedAt`'s fallback branch is unexercised.** — `turn_timer.go:84-90`. The `m.turnTimer == nil` path reverts to raw wall clock and is reachable only for a model built without `beginRun`, which is a test-shaped situation. A one-line test would stop it rotting.
- **`stampFirstToken` treats whitespace differently on the two callbacks.** — `model.go:5180` vs `:5283`. `onText` stamps on any delta including whitespace-only; `onReasoning` stamps only when the delta is non-blank. That matches the previous behaviour exactly, so this PR introduces nothing, but the asymmetry is now concentrated in one helper where it reads as if it were intentional.

## Not covered by this PR

- Time spent at an ask-user questionnaire (`m.pendingAskUser`) or a spec review (`m.pendingSpecReview`) is still counted as model work. Those are the same "blocked on a human" shape as a permission prompt, and `quietGenerationHint` was taught only about `m.pendingPermission` (`model.go:3593`). Whether they should also pause is a product call rather than a defect here, but the asymmetry is now baked in and is worth its own issue.

## Not verified

- Any platform other than darwin/arm64; nothing here is platform-divergent.
- The manual termctrl walkthrough in the PR body. My verification is the deterministic test plus the two mutations above.

Vasanthdev2004 (approved):
Approving. Both are fixed and the test now pins them rather than passing either way.

TTFT is stamped from the timer at token arrival, so it and elapsed share a clock. Reverting it to `at.Sub(started)` fails with `ttft = 1m35s, want 5s`.

The test change is the part I like. Putting 2s of work before the pause and 3s after means `pausedAt` is no longer equal to `startedAt`, which is what made the old version blind: reverting `resume` to `at.Sub(timer.startedAt)` now fails with `turn elapsed = 3s, want 5s`, where before it stayed green. The exact `5s` assertions are doing real work.

Nothing further from me on the ask_user point. It was pre-existing and non-blocking, and the description no longer overstates the scope.

Vasanthdev2004:
@kevincodex1 this one is ready whenever you are. Three approvals on `0fd0bc5c`, 9 of 9 green, CLEAN and mergeable, nothing outstanding.

It is @anandh8x's, so my approval is a reviewer's rather than the author's. Worth a line on why I am comfortable with it: I raised two things and both came back properly fixed rather than papered over. TTFT is now stamped from the same pause-aware timer as elapsed, so `/context` cannot report a time-to-first-token larger than the turn that produced it. And the regression test was reworked so it actually discriminates: the previous version had the pause starting at the same instant as the turn, which made two different implementations of the timer arithmetic indistinguishable. I re-checked both by reverting them, and both now fail.

One user-visible note for release notes rather than for you to act on: turn durations shown in the TUI will get shorter for any turn that waited on a permission prompt, because that wait is no longer counted. That is the point of the change, but it will look like a performance improvement in anyone's before-and-after comparison, and it is not one.

PR #821, opened 2026-07-29
fix(tui): hide unavailable sidebar shortcut

anandh8x:
## What changed

- hide the `Ctrl+B sidebar` footer shortcut when the sidebar has no content
- keep the shortcut visible when a plan, agent, or changed-file activity makes the sidebar available
- keep the shortcut visible while an available sidebar is collapsed so users can restore it
- add regression coverage for empty, available, and collapsed sidebar states

## Why

The footer previously advertised `Ctrl+B sidebar` unconditionally on medium and wide terminals. In sessions without sidebar content, pressing the shortcut produced no visible result and made the control appear broken.

The footer now uses the same availability predicate as the sidebar layout, while intentionally ignoring the user's collapsed preference for the restore case.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- manual termctrl verification of empty, visible, collapsed, and restored sidebar states

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved composer idle footer hints to display only relevant shortcuts.
  * Sidebar shortcut messaging now appears only when the sidebar is available, and remains correct even when the sidebar is collapsed.
  * Wide layouts now show additional detail, copy, and mode-switching hints.
* **Tests**
  * Expanded coverage for composer idle hint and jump cue rendering across width variants and sidebar availability scenarios.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -3036,9 +3036,18 @@ func (m model) composerIdleHint() string {
 	case tierNarrow:
 		hint = "? shortcuts"
 	case tierMedium:
-		hint = fmt.Sprintf("? shortcuts · Ctrl+X cmds · %s sidebar", sidebarKey)
+		parts := []string{"? shortcuts", "Ctrl+X cmds"}
+		if m.sidebarAvailable() {
+			parts = append(parts, sidebarKey+" sidebar")
+		}
+		hint = strings.Join(parts, " · ")
 	default:
-		hint = fmt.Sprintf("? shortcuts · Ctrl+X cmds · %s sidebar · %s detail · %s copy · Shift+Tab mode", sidebarKey, detailKey, mouseKey)
+		parts := []string{"? shortcuts", "Ctrl+X cmds"}
+		if m.sidebarAvailable() {
+			parts = append(parts, sidebarKey+" sidebar")
+		}
+		parts = append(parts, detailKey+" detail", mouseKey+" copy", "Shift+Tab mode")
+		hint = strings.Join(parts, " · ")
 	}
 	return zeroTheme.faint.Render(hint)
 }
--- a/internal/tui/model_test.go
+++ b/internal/tui/model_test.go
@@ -2502,8 +2502,21 @@ func TestComposerIdleHintAndJumpCue(t *testing.T) {
 	m.transcript = append(m.transcript, transcriptRow{kind: rowAssistant, text: "hi", final: true})
 
 	// Idle, empty composer, managed mode -> the discoverability hint shows.
-	if h := plainRender(t, m.composerIdleHint()); !strings.Contains(h, "shortcuts") {
-		t.Fatalf("expected idle hint, got %q", h)
+	hint := plainRender(t, m.composerIdleHint())
+	if !strings.Contains(hint, "shortcuts") {
+		t.Fatalf("expected idle hint, got %q", hint)
+	}
+	if strings.Contains(hint, "sidebar") {
+		t.Fatalf("empty sidebar should not be advertised, got %q", hint)
+	}
+	withSidebar := m
+	withSidebar.plan.steps = []planStep{{content: "inspect footer", status: "in_progress"}}
+	if hint := plainRender(t, withSidebar.composerIdleHint()); !strings.Contains(hint, "Ctrl+B sidebar") {
+		t.Fatalf("available sidebar should be advertised, got %q", hint)
+	}
+	withSidebar.sidebarHidden = true
+	if hint := plainRender(t, withSidebar.composerIdleHint()); !strings.Contains(hint, "Ctrl+B sidebar") {
+		t.Fatalf("collapsed sidebar should keep its restore shortcut, got %q", hint)
 	}
 	// Hidden during a run.
 	busy := m
--- a/internal/tui/model_test.go
+++ b/internal/tui/model_test.go
@@ -2501,22 +2501,36 @@ func TestComposerIdleHintAndJumpCue(t *testing.T) {
 	m.width, m.height = 100, 30
 	m.transcript = append(m.transcript, transcriptRow{kind: rowAssistant, text: "hi", final: true})
 
-	// Idle, empty composer, managed mode -> the discoverability hint shows.
-	hint := plainRender(t, m.composerIdleHint())
-	if !strings.Contains(hint, "shortcuts") {
-		t.Fatalf("expected idle hint, got %q", hint)
-	}
-	if strings.Contains(hint, "sidebar") {
-		t.Fatalf("empty sidebar should not be advertised, got %q", hint)
-	}
-	withSidebar := m
-	withSidebar.plan.steps = []planStep{{content: "inspect footer", status: "in_progress"}}
-	if hint := plainRender(t, withSidebar.composerIdleHint()); !strings.Contains(hint, "Ctrl+B sidebar") {
-		t.Fatalf("available sidebar should be advertised, got %q", hint)
-	}
-	withSidebar.sidebarHidden = true
-	if hint := plainRender(t, withSidebar.composerIdleHint()); !strings.Contains(hint, "Ctrl+B sidebar") {
-		t.Fatalf("collapsed sidebar should keep its restore shortcut, got %q", hint)
+	for _, test := range []struct {
+		name  string
+		width int
+	}{
+		{name: "medium", width: 80},
+		{name: "full", width: 100},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			idle := m
+			idle.width = test.width
+
+			// Idle, empty composer, managed mode -> the discoverability hint shows.
+			hint := plainRender(t, idle.composerIdleHint())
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (dismissed):
Approving. Small and correct, and the test covers the case I would have asked about: a collapsed sidebar keeps advertising its restore shortcut, which is the one place where hiding the hint would have been wrong.

I did go looking for a second unguarded `sidebarKey` use and there is only the one function, both call sites guarded. `sidebarAvailable` already covers alt-screen, subchat, detailed transcript and the width tier, so the hint now tracks the same conditions the key itself does.

gnanam1990 (changes requested):
## Verdict

Changes requested — for one missing test, not for the code. The change is correct and small; it just ships one of its two new branches uncovered, and I can show that the suite does not notice when that branch is wrong.

Reviewed at `99daf8f`, base `097c265` (level with `origin/main`). Author has write access, so this is the internal cycle rather than the `issue-approved` gate.

## What I verified

- **The wide-tier branch is genuinely tested.** Forcing the default branch to advertise the sidebar unconditionally fails the new assertion:

  ```
  --- FAIL: TestComposerIdleHintAndJumpCue
      model_test.go:2510: empty sidebar should not be advertised, got
      "? shortcuts · Ctrl+X cmds · Ctrl+B sidebar · Ctrl+O detail · Ctrl+E copy · Shift+Tab mode"
  ```

- **Reusing `sidebarAvailable()` is the right predicate.** It is the same one the layout uses to decide whether the sidebar renders at all (`model.go:3040`, `sidebar.go:113`), so the footer and the layout cannot disagree — which is precisely the defect the PR describes. Deriving a second "is there content" rule here would have been the drift-prone alternative.
- **The collapsed case is handled deliberately, not accidentally.** `sidebarAvailable()` does not consult `m.sidebarHidden`, so a collapsed-but-available sidebar keeps advertising `Ctrl+B` and the user can restore it. The test covers this and it is the subtle half of the change.
- Gauntlet on darwin/arm64 from a non-`/tmp` checkout: `make fmt-check`, `go vet ./...`, `git diff HEAD --check` clean, full `go test ./...` green.
- **Passes executed:** 0–4, 7, 9, 11, 12. Passes 5, 6, 8 and 10 did not apply — this is a pure display-string change with no state, credentials, platform primitive or external contract.

## Blocking

1. **The `tierMedium` branch has no test coverage, and the suite cannot tell when it is wrong.** — `internal/tui/model.go:3039-3044`

   The PR adds the same guard in two places. I mutated each one independently:

   - Default/wide branch made unconditional → `TestComposerIdleHintAndJumpCue` **fails** (output above).
   - `tierMedium` branch made unconditional → the entire `internal/tui` package **passes**.

   So `TestComposerIdleHintAndJumpCue` only ever exercises the wide tier. The medium tier could ship advertising `Ctrl+B sidebar` on an empty sidebar — the exact bug this PR exists to fix — and nothing would catch it.

   That matters more than a normal coverage gap because the PR body names medium explicitly: *"The footer previously advertised `Ctrl+B sidebar` unconditionally on medium and wide terminals."* Half of the stated fix is unverified.

   The fix is small: set a width that lands in `tierMedium` and repeat the empty/available/collapsed assertions. Given the two branches now carry duplicated logic, a table over both tiers would be better than a second copy of the test — and would make it obvious if the two ever diverge.

## Non-blocking

- **The guard is written twice and will drift.** — `internal/tui/model.go:3039-3044` and `:3046-3052`. Both branches build `[]string{"? shortcuts", "Ctrl+X cmds"}` and then conditionally append the same sidebar entry; only the trailing wide-tier items differ. Hoisting the shared prefix — build the common parts once, append the wide-only extras in the `default` case — would leave one copy of the rule. This is the repo's recurring "two copies will drift" class, and the untested branch above is what that costs in practice.
- **`fmt.Sprintf` was dropped for `strings.Join` in both branches**, which is the right call for a variable-length list. Worth confirming `strings` was already imported rather than added for this — it was.

## Not covered by this PR

- `tierNarrow` never advertised the sidebar, so it is unaffected and correctly left alone.
- The footer still advertises `Ctrl+O detail`, `Ctrl+E copy` and `Shift+Tab mode` unconditionally on wide terminals. If any of those can also be unavailable in a given session, they have the same "advertised control does nothing" problem this PR is fixing for the sidebar — worth a separate issue rather than expanding this one.

## Not verified

- Any platform other than darwin/arm64, and the manual termctrl walkthrough in the PR body. Neither is likely to diverge for a display string.

Add the medium-tier case and this is an approve; the change itself I have no argument with.

anandh8x:
@gnanam1990 added table-driven coverage for both medium and full widths, including empty, available, and collapsed sidebar states. Could you re-review when you get a chance?

gnanam1990 (approved):
## Verdict

Approve. The blocker from my previous review is fixed, and I confirmed the new test actually discriminates rather than merely looking like it does.

Re-reviewed at `401d4dc`, base `097c265`. Previously reviewed at `99daf8f`; the delta is one commit, `401d4dc "test(tui): cover medium sidebar footer hint"`, touching `internal/tui/model_test.go` only — no production change.

## Withdrawn

**The `tierMedium` coverage gap is closed.** My blocker was that mutating the medium branch to advertise the sidebar unconditionally left the whole `internal/tui` package green, so half of what the PR body claims was unverified. Re-running that same mutation against `401d4dc`:

```
=== RUN   TestComposerIdleHintAndJumpCue/medium
=== RUN   TestComposerIdleHintAndJumpCue/full
--- FAIL: TestComposerIdleHintAndJumpCue
    --- FAIL: TestComposerIdleHintAndJumpCue/medium
    --- PASS: TestComposerIdleHintAndJumpCue/full
```

And the mirror-image mutation on the wide branch fails `full` while leaving `medium` passing. Each subtest fails for its own branch and only its own branch, which is the property I actually wanted — a table that runs twice but asserts against one shared code path would have looked identical in the diff and caught nothing.

I also checked the width choice rather than assuming it: `widthTier` puts 80–99 in `tierMedium` (`internal/tui/view.go:36`), so `width: 80` genuinely enters the branch under test and `width: 100` the default one. And `sidebarAvailable()` itself bails below `tierMedium` (`internal/tui/sidebar.go:94`), so at width 80 the plan-step fixture does make the sidebar available — the "available" and "collapsed" legs of the table are meaningful at both widths, not vacuously true at one of them.

Converting to a table over widths was also the better shape than duplicating the test, which is what I had suggested as the alternative. Thanks for taking the harder of the two.

- Both mutations above, each isolated to its own subtest.
- Gauntlet, darwin/arm64, non-`/tmp` checkout: `make fmt-check` clean, `go vet ./...` clean, `git diff HEAD --check` clean, `go test ./internal/tui/ -count=1` green.
- The production diff is byte-identical to `99daf8f`, so nothing from my earlier reading needs redoing.
- **Passes executed:** 0, 1, 3, 7, 11, 12 — scoped to a test-only delta against a production change I had already reviewed in full.

## Still open, non-blocking (carried forward)

- **The guard is written twice and will drift.** — `internal/tui/model.go:3039-3044` and `:3046-3052`. Both branches build `[]string{"? shortcuts", "Ctrl+X cmds"}` and conditionally append the same sidebar entry; only the trailing wide-tier items differ. Hoisting the shared prefix would leave one copy of the rule. This is unchanged from my last review and is not worth holding the PR for — the new table means a divergence between the two copies would now be caught, which was the actual risk.

## Not covered by this PR

- `tierNarrow` never advertised the sidebar and is correctly untouched.
- `Ctrl+O detail`, `Ctrl+E copy` and `Shift+Tab mode` are still advertised unconditionally on wide terminals. If any of those can be unavailable in a session, they have the same "advertised control does nothing" problem — worth a separate issue rather than expanding this one.

## Not verified

- Any platform other than darwin/arm64, and the manual termctrl walkthrough in the PR body. Neither is likely to diverge for a display string.

Vasanthdev2004 (approved):
Re-approving on 401d4dcc. My earlier approval was dismissed by the push, so I re-verified rather than just restoring it.

The production code is untouched since I last looked — the whole delta is model_test.go. So the only question was whether the new table actually closes gnanam's gap, and it does. I ran his exact mutation, making the tierMedium branch advertise the sidebar unconditionally:

    --- FAIL: TestComposerIdleHintAndJumpCue/medium
        model_test.go:2521: empty sidebar should not be advertised, got "? shortcuts · Ctrl+X cmds · Ctrl+B sidebar"

That's the one that used to leave the whole package green. The wide branch still fails its own mutation independently, and mutating tierMedium the other way — never advertise — fails the available/collapsed assertions, so both directions are pinned on both tiers. width 80 lands in tierMedium and 100 in tierFull, which is what makes the table work.

Clean on gofmt and vet. Only failure in internal/tui is TestAltScreenTranscriptScrollKeepsFooterFixed, which fails the same way on clean origin/main from a checkout of the same path depth on my box — not this PR.

Worth doing the prefix hoist gnanam mentioned at some point, since the two branches now carry the same rule twice, but that's a follow-up and not a reason to hold this.

PR #784, opened 2026-07-21
Add Atlas Cloud provider preset

binyangzhu000-sudo:
## Summary
- add Atlas Cloud as a built-in OpenAI-compatible provider preset
- configure the default endpoint, ATLASCLOUD_API_KEY auth env var, and qwen/qwen3.5-flash default model
- add Atlas Cloud curated model picker entries for qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro

## Validation
- git diff --check
- Atlas Cloud live catalog check: 434 models returned; qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro confirmed present
- Go tests not run locally because this machine does not have go/gofmt installed

No README or promotional sections were changed.

## Summary by CodeRabbit

* **New Features**
  * Added Atlas Cloud as a supported OpenAI-compatible provider.
  * Added bundled models: Qwen 3.5 Flash and DeepSeek V4 Pro.
  * Added support for Atlas Cloud aliases, including “Atlas Cloud” and “Atlas”.
  * Configured authentication through the `ATLASCLOUD_API_KEY` environment variable.

PR #826, opened 2026-07-29
feat(tui): replace /retitle with local /rename

anandh8x:
## Summary

- Replace the bulk `/retitle` command with a local `/rename [title]` command.
- Make bare `/rename` open a prefilled editor with Enter-to-save and Esc-to-cancel behavior.
- Preserve names chosen before a fresh session's first prompt and prevent a late automatic-title response from overwriting a manual rename.
- Polish `/resume` with aligned timestamp/title columns, hidden-but-searchable raw session IDs, and a visible position count.

## Why

Renaming a session should be an immediate local metadata operation rather than a model-backed bulk action. The previous resume rows also gave raw IDs too much visual space, truncating the titles users actually scan for.

## Impact

Users can rename the current session without a provider call or token usage, including before the first prompt. Resume rows are cleaner and easier to scan while session IDs remain available through search.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- Manual TUI verification of bare rename, inline rename, save, cancel, prefill, fresh-session naming, and the resume picker

## Summary by CodeRabbit

* **New Features**
  * Added `/rename` to rename the active session or set a title before the first prompt.
  * Added an interactive rename editor with save, cancel, validation, and paste support.
  * Session pickers now show cleaner titles, aligned timestamps, safer searchable entries, and “position / total” pagination.
* **Bug Fixes**
  * Automatic titles no longer overwrite manual renames (even when an auto-title is in progress).
  * Empty rename pastes no longer trigger clipboard image handling.
  * `/retitle` is no longer available; rename editor state better blocks unrelated UI interactions.

BREAKING CHANGE: /retitle is removed; use /rename instead.

### DIFF
--- a/internal/sessions/session_title_test.go
+++ b/internal/sessions/session_title_test.go
@@ -43,7 +43,7 @@ func TestUpdateTitle(t *testing.T) {
 		t.Fatalf("title not trimmed/stored: %q", updated.Title)
 	}
 	if updated.UpdatedAt != before.UpdatedAt {
-		t.Fatalf("UpdatedAt must not change on retitle: before=%q after=%q", before.UpdatedAt, updated.UpdatedAt)
+		t.Fatalf("UpdatedAt must not change on rename: before=%q after=%q", before.UpdatedAt, updated.UpdatedAt)
 	}
 	if updated.EventCount != before.EventCount {
 		t.Fatalf("EventCount changed: before=%d after=%d", before.EventCount, updated.EventCount)
@@ -67,12 +67,36 @@ func TestUpdateTitle(t *testing.T) {
 
 	// An unchanged title is a no-op (still succeeds).
 	if _, err := store.UpdateTitle(session.SessionID, "Clean Generated Title"); err != nil {
-		t.Fatalf("no-op retitle should succeed: %v", err)
+		t.Fatalf("no-op rename should succeed: %v", err)
 	}
 
 	// An invalid session id is rejected.
 	if _, err := store.UpdateTitle("../escape", "whatever"); err == nil {
 		t.Fatal("expected an invalid session id to be rejected")
+	}
+}
+
+func TestUpdateTitleIfCurrent(t *testing.T) {
+	store := newTitleTestStore(t)
+	session, err := store.Create(CreateInput{Title: "Automatic title"})
+	if err != nil {
+		t.Fatalf("create: %v", err)
+	}
+
+	updated, applied, err := store.UpdateTitleIfCurrent(session.SessionID, "Automatic title", "Generated title")
+	if err != nil || !applied || updated.Title != "Generated title" {
+		t.Fatalf("matching update = (%#v, %v, %v)", updated, applied, err)
+	}
+
+	if _, err := store.UpdateTitle(session.SessionID, "Manual title"); err != nil {
+		t.Fatalf("manual rename: %v", err)
+	}
+	updated, applied, err = store.UpdateTitleIfCurrent(session.SessionID, "Generated title", "Late generated title")
+	if err != nil {
+		t.Fatalf("stale update: %v", err)
+	}
+	if applied || updated.Title != "Manual title" {
+		t.Fatalf("stale update should preserve manual title: (%#v, %v)", updated, applied)
 	}
 }
 
--- a/internal/sessions/store.go
+++ b/internal/sessions/store.go
@@ -719,7 +719,7 @@ func (store *Store) appendPreparedEventsLocked(sessionID string, inputs []prepar
 // serialized under the same per-session lock as AppendEvent and re-reads the
 // latest metadata under that lock before rewriting, so a concurrent append can't
 // clobber the new title (nor the title clobber a concurrent append's event
-// count/timestamp). UpdatedAt is deliberately left untouched: a retitle is not
+// count/timestamp). UpdatedAt is deliberately left untouched: a rename is not
 // activity, so it must not reorder the session in the resumable list. A blank
 // title is rejected so a failed model generation can never erase a useful
 // first-message title, and an unchanged title is a no-op (no rewrite/fsync).
@@ -749,6 +749,40 @@ func (store *Store) UpdateTitle(sessionID string, title string) (Metadata, error
 		return Metadata{}, err
 	}
 	return session, nil
+}
+
+// UpdateTitleIfCurrent replaces a session title only when it still matches
+// expected. It lets background automatic naming avoid overwriting a newer manual
+// rename while keeping the check and write under the same per-session lock.
+func (store *Store) UpdateTitleIfCurrent(sessionID string, expected string, title string) (Metadata, bool, error) {
+	if !ValidSessionID(sessionID) {
+		return Metadata{}, false, fmt.Errorf("invalid zero session id %q", sessionID)
+	}
+	trimmed := strings.TrimSpace(title)
+	if trimmed == "" {
+		return Metadata{}, false, fmt.Errorf("zero session title is required")
+	}
+	unlock, err := store.lockSession(sessionID)
+	if err != nil {
+		return Metadata{}, false, err
+	}
+	defer unlock()
+
+	session, err := store.readMetadata(sessionID)
+	if err != nil {
+		return Metadata{}, false, err
+	}
+	if session.Title != strings.TrimSpace(expected) {
+		return session, false, nil
+	}
+	if session.Title == trimmed {
+		return session, true, nil
+	}
+	session.Title = trimmed
+	if err := store.writeMetadata(session); err != nil {
+		return Metadata{}, false, err
+	}
+	return session, true, nil
 }
 
 // UpdateModel replaces a session's selected model without changing its activity
--- a/internal/tui/btw.go
+++ b/internal/tui/btw.go
@@ -206,7 +206,7 @@ func (m model) leaveBTW() (model, tea.Cmd) {
 func btwCommandUnavailable(command parsedCommand) bool {
 	arg := strings.ToLower(strings.TrimSpace(command.text))
 	switch command.kind {
-	case commandNew, commandResume, commandRetitle, commandSpec, commandLoop, commandGoal,
+	case commandNew, commandResume, commandRename, commandSpec, commandLoop, commandGoal,
 		commandRewind, commandCompact, commandSTTModel, commandMCP:
 		return true
 	case commandModel:
--- a/internal/tui/clipboard.go
+++ b/internal/tui/clipboard.go
@@ -58,7 +58,8 @@ func pasteFromClipboardCmd() tea.Cmd {
 // right-click paste (clipboardReadMsg) so a bracketed paste and a right-click
 // paste behave identically. Surfaces with no editable text field (a permission/
 // spec prompt, the MCP manager, an open picker, the detailed transcript) swallow
-// the paste; empty content is a no-op.
+// the paste. The session rename editor accepts text only; empty content is a
+// no-op there rather than an image probe.
 func (m model) routePaste(content string) (tea.Model, tea.Cmd) {
 	// A paste is a deliberate action, same as a keypress or click — it means
 	// the user moved on to something else, so it disarms a stale Esc
@@ -75,6 +76,14 @@ func (m model) routePaste(content string) (tea.Model, tea.Cmd) {
 	// toggle instead of going solid.
 	m.lastCharTime = m.now()
 	m.composerCursorVisible = true
+	if m.renamePrompt != nil {
+		if content == "" {
+			return m, nil
+		}
+		var cmd tea.Cmd
+		m.input, cmd = m.input.Update(tea.PasteMsg{Content: sanitizeComposerInput(content)})
+		return m, cmd
+	}
 	if content == "" {
 		// Empty text clipboard — the user may have pasted a screenshot.
 		// Probe the OS clipboard for image content asynchronously.
--- a/internal/tui/commands.go
+++ b/internal/tui/commands.go
@@ -27,7 +27,7 @@ const (
 	commandPlan
 	commandSearch
 	commandResume
-	commandRetitle
+	commandRename
 	commandSpec
 	commandInit
 	commandCompact
@@ -226,11 +226,11 @@ var commandDefinitions = []commandDefinition{
 		kind:        commandResume,
 	},
 	{
-		name:        "/retitle",
-		usage:       "/retitle",
+		name:        "/rename",
+		usage:       "/rename [title]",
 		group:       commandGroupSession,
-		description: "Generate concise titles for resumable sessions.",
-		kind:        commandRetitle,
+		description: "Rename the current session (no arg opens an editor).",
+		kind:        commandRename,
 	},
 	{
 		name:        "/spec",
--- a/internal/tui/composer.go
+++ b/internal/tui/composer.go
@@ -281,7 +281,7 @@ func (m model) composerPositionAtMouse(msg tea.MouseMsg) (int, bool) {
 
 func (m model) composerMouseSelectionBlocked() bool {
 	return m.transcriptDetailed || m.setup.visible || m.providerWizard != nil || m.mcpAddWizard != nil ||
-		m.mcpManager != nil || m.picker != nil || m.suggestionsActive()
+		m.mcpManager != nil || m.picker != nil || m.renamePrompt != nil || m.suggestionsActive()
 }
 
 func (m model) composerPositionAtVisualCell(x int, y int, width int) (int, bool) {
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -102,6 +102,7 @@ type model struct {
 	doctorInFlight       bool
 	doctorFrame          int
 	activeSession        sessions.Metadata
+	pendingSessionTitle  string
 	sessionEvents        []sessions.Event
 	btw                  btwState
 	// btwRunIDSeq is the highest run ID issued by any completed or abandoned BTW
@@ -112,14 +113,8 @@ type model struct {
 	// already been attempted this process, so a finished turn re-fires the title
 	// generator at most once per session (even before its async result lands).
 	// Lazily initialized.
-	titledSessions map[string]bool
-	// retitle* drive the sequential /retitle backfill: queued session ids still
-	// awaiting a title, whether a backfill is running, and its progress counters.
-	retitleQueue                []string
-	retitleActive               bool
-	retitleTotal                int
-	retitleDone                 int
-	retitleOK                   int
+	titledSessions              map[string]bool
+	renamePrompt                *sessionRenamePrompt
 	usageTracker                *usage.Tracker
 	sessionCompactor            SessionCompactor
 	prService                   *PrService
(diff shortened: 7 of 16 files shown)


gnanam1990 (changes requested):
## Verdict

Changes requested — for one thing that is not a code change. The engineering here is sound and I could not break it; the problem is that as titled, this PR will silently drop a user-facing breaking change from the changelog.

Reviewed at `36f16b7`, base `097c265` (level with `origin/main`, zero commits behind). Author has write access, so this is the internal cycle rather than the `issue-approved` gate.

## What I verified

- **The compare-and-swap actually does the work it claims.** I gutted it — removed the `session.Title != strings.TrimSpace(expected)` guard at `internal/sessions/store.go:775` so a late automatic title always overwrites — and two independent tests failed, one at each layer:

  ```
  --- FAIL: TestUpdateTitleIfCurrent
      stale update should preserve manual title: Title:"Late generated title" ... true
  --- FAIL: TestManualRenameWinsOverInFlightAutoTitle
      late automatic title should be skipped, got {title:"Generated Automatic Title", applied:true}
  ```

- **The pre-first-prompt path is guarded, and the guard is tested.** Removing the `m.titledSessions[session.SessionID] = true` block at `internal/tui/session.go:48-53` fails `TestRenameNamesFreshSessionBeforeFirstPrompt` ("a manually named fresh session must skip automatic naming"). Belt and braces here — `sessionTitleIsAuto` at `session_title.go:213` would also refuse a manual title — but the explicit gate is the one under test.
- **"Hidden but searchable" holds.** Dropping `Meta: meta.SessionID` from the picker item does not cost searchability: `scorePickerItem` builds its haystack from `Group, Label, Value, Meta` (`internal/tui/picker.go:168`) and `Value` is still the session ID.
- **The `n / total` footer is consistent with the filtered view.** `applyPickerFilter` replaces `p.items` wholesale (`picker.go:158`) and resets `selected` to 0, so the counter and the visible list always agree. I went looking for an off-by-one against an unfiltered total and there isn't one.
- **`/retitle` is fully excised** — no leftover `retitleQueue`/`retitleActive`/`retitleTotal`/`retitleDone`/`retitleOK` fields, no stale docs. The only surviving reference is the negative test at `session_rename_test.go:182`.
- **Gauntlet, darwin/arm64, non-`/tmp` checkout:** `make fmt-check` clean, `go vet ./...` clean, `go test ./...` fully green, `go test -race ./internal/tui/ ./internal/sessions/ -count=3` green, `go run ./cmd/zero-release build` and `smoke` both pass, `git diff HEAD --check` clean.
- **Passes executed:** 0–5, 7–12. Pass 6 was run narrowly rather than in full — this diff has no auth, exec, parsing or network surface; I checked the two boundaries it does touch (the new store write path and the paste route) and nothing else applied.

## Blocking

1. **The PR title is not a conventional commit, and this repo derives its changelog from the squash subject.** — PR title / `release-please-config.json`

   The chain, each hop checked:

   - `release-please-config.json` and `.github/workflows/release-please.yml` mean versioning and `CHANGELOG.md` are generated from conventional-commit subjects.
   - `gh api repos/Gitlawb/zero` reports `squash_merge_commit_title=COMMIT_OR_PR_TITLE`. This PR has **two** commits (`daee6c15`, `36f16b74`), so the squash subject will be the **PR title**, not a commit subject.
   - That title is `Improve session naming and resume picker` — no type, no scope. release-please parses no type, so the change contributes **no changelog entry and no version bump**.
   - Every recently merged PR does follow the convention: `fix(sandbox):`, `test(tools):`, `fix(config):`.

   That would be a minor annoyance on an ordinary change. It is not one here, because this PR **removes `/retitle`**, which shipped in #224 and has been present in every release from v0.3.0 through v0.6.0. A command that existed in the last release disappears, and under the current title the changelog will not say so.

   Retitle to something like `feat(tui)!: replace /retitle with a local /rename` and add a `BREAKING CHANGE: /retitle is removed; use /rename` footer. No code change required — this is the whole blocker.

## Non-blocking

- **`&& msg.applied` currently has no observable effect.** — `internal/tui/session_title.go:255`. I mutated it out and the entire `internal/tui` suite stayed green, so I traced why rather than reporting it as untested-and-therefore-broken. When the store rejects the update, `UpdateTitleIfCurrent` returns the freshly-read metadata (`store.go:777`), so `msg.title` is already the *manual* title — assigning it to `m.activeSession.Title` is a no-op, and the `titledSessions` gate is retained on both branches either way. The flag is correct defensive code that becomes load-bearing the moment `UpdateTitleIfCurrent` returns anything else on rejection. Worth a one-line comment saying that, so a future reader does not delete it as dead.
- **Four separable changes in one PR.** — diffstat. The `/retitle`→`/rename` swap, the rename editor, the pre-prompt naming plus compare-and-swap, and the `/resume` picker relayout (`session.go:406-436`, `view.go:816-825`) are independent; the picker work does not depend on any of the rename work. Team PR through the internal cycle, so I am not treating this as a scope close — but the picker change is the one a bisect would most want isolated.
- **Merge-conflict exposure with two long-lived branches.** #643 and #570 both touch `internal/tui/session.go`, `commands.go` and `model.go`. Neither is close to merging, and this branch is current, so the right outcome is probably that they rebase onto this — worth saying so on those PRs rather than holding this one.
- **`/rename` is missing from the README command table.** — `README.md:214`. `/resume`, `/rewind` are listed as the session commands; `/retitle` never was, so nothing is stale, but a new user-facing command belongs on that line.
- **Blank submit closes the editor instead of holding it open.** — `internal/tui/session_rename.go:32-37`. `m.renamePrompt = nil` is set before validation, so Enter on whitespace closes the editor and prints an error. `TestRenameEditorRejectsBlankAndEscCancels:157` asserts this deliberately, so I read it as intended rather than an oversight — flagging only because the user loses their editing context to recover from a typo.

## Questions for the author

- `/retitle` refused to run mid-turn (`Cannot retitle sessions while a run is active`); `/rename` has no `m.pending` guard. I believe that is correct — the old guard existed because `/retitle` made provider calls and `/rename` is pure local metadata, and the in-flight case is exactly what the compare-and-swap covers. Confirming that was deliberate rather than dropped with the surrounding code.

## Not covered by this PR

- A migration path for `/retitle`. It now resolves to nothing, so a user who typed it last week gets a generic unknown-command error. A one-release alias that prints "renamed to `/rename`" would be kinder than a changelog line alone, though the changelog line is the part I am blocking on.
- Bulk retitling of old sessions is simply gone as a capability. That looks intentional given the "no provider call, no token usage" rationale, but nothing in the PR body says the bulk case was considered and dropped rather than overlooked.

## Not verified

- Windows and Linux — everything above is darwin/arm64 only. The diff is platform-neutral TUI and store code, and `internal/sessions` locking is unchanged by this PR, so I did not expect divergence and did not look for it.
- The manual TUI walkthrough listed in the PR description. I exercised the rename paths through the model's `Update` loop and the store, not through a real terminal, so the editor's rendering and the picker's new column alignment are unconfirmed by eye.
- Cross-process races. `UpdateTitleIfCurrent` takes the same per-session lock as `UpdateTitle` and `AppendEvent`, and I read that as correct, but I did not run two `zero` processes against one session to prove it.

Vasanthdev2004 (approved):
Approving, with two small things you can take or leave.

`renameActiveSession` (session_rename.go:47) does `TrimSpace` then `cutRunes`, which strips the ends but keeps interior newlines. `/rename foo<newline>bar` persists a multi-line title into metadata, and the pickers and listings all assume one line. The other title paths collapse whitespace, so this one is the odd one out.

The empty-paste guard in `routePaste` (clipboard.go:79-81) looks unreachable: the right-click path already returns earlier, so nothing gets there with an empty payload. The clipboard change is also the one part of this PR with no test.

Neither is worth holding it for. The store change is backward compatible, sessions without the new fields still load, and I could not get the picker to point at a deleted session.

anandh8x:
@gnanam1990 Updated the PR title to the conventional breaking-change format and added the `BREAKING CHANGE` note. Could you please re-review and clear the changes-requested status if this resolves the blocker?

Vasanthdev2004:
@gnanam1990 your blocker is done — the title is now `feat(tui)!: replace /retitle with local /rename`, two commits so the squash subject is the PR title, and release-please will see the `!`. Worth flipping your review when you get a minute; my approval on 36f16b74 still stands and the code hasn't moved.

Doing exactly what you asked did surface two things about the release mechanics that I don't think either of us looked at, and one of them is a real decision rather than a nit.

**The `!` is on track to cut 1.0.0, not 0.7.0.** `release-please-config.json` doesn't set `bump-minor-pre-major`, and that's the flag whose entire job is "breaking changes only bump semver minor if version < 1.0.0". Without it a breaking change bumps major, so at 0.6.x that's 1.0.0. The open release PR #837 is currently proposing 0.6.1, so merging this would take that from a patch to a major in one step. @kevin-hnnh that's your call, not mine — this PR is correct either way, but 1.0.0 arriving as a side effect of a TUI command rename seems worth a deliberate yes. If the answer is "not yet", setting `bump-minor-pre-major: true` gets you 0.7.0 and keeps the breaking-change entry.

**The `BREAKING CHANGE:` footer is in the wrong place to reach the commit.** It's in the PR description, but this repo has `squash_merge_commit_message=COMMIT_MESSAGES`, so the squash body comes from the two commit messages — "Improve session naming and resume picker" and "Address session rename review feedback" — and the PR body is dropped. The `!` still gets you the bump and the breaking flag, but the footer is what makes the changelog actually say *why*, which was gnanam's point in the first place. Putting the footer in a commit message fixes it.

Neither is a code change and I'm not re-blocking on either. Flagging because this would be the repo's first `!` commit — nothing on main has ever carried one, so this is the first time the config gets exercised.

gnanam1990 (changes requested):
## Verdict

Changes requested — but for a consequence my own previous review created, not for anything you did wrong. The title fix is exactly right; following it will, on current config, cut **v1.0.0** instead of v0.7.0. That needs a decision before merge, and it is a one-line config change or a deliberate "yes, 1.0.0".

Re-reviewed at `36f16b7`, base `097c265` — the same commit as my previous review. **My freshness check missed this fix entirely**, because the head SHA cannot move when the change is to PR metadata. Worth recording as a limitation of SHA-diffing when the blocker itself lives outside git.

## Withdrawn — the changelog blocker is fixed

The title is now `feat(tui)!: replace /retitle with local /rename`, with the `!` breaking marker. With two commits, `squash_merge_commit_title=COMMIT_OR_PR_TITLE` takes the PR title, so the squash subject will be conventional and breaking-marked. Resolved.

While confirming the mechanism I found much better evidence than I originally posted, which is worth leaving on the record because it shows this was not a theoretical concern. Five merged PRs whose squash subjects lacked a conventional type are **absent from `CHANGELOG.md` entirely**:

| Commit on `main` | In CHANGELOG |
|---|---|
| `c00a473 Add Atlas Cloud provider preset (#784)` | no |
| `ab2f9fb Add shell completion generation (#764)` | no |
| `18cce35 Add deterministic completion policy (#719)` | no |
| `9242b9e Add transcript-grounded structured task state (#761)` | no |
| `0125e70 Fix pinned Go quality tools on Go 1.26 (#689)` | no |

A whole provider preset and a new user-facing command shipped with no changelog record. That is the class, it predates this PR, and it deserves its own issue rather than anything from you here.

## Blocking

1. **`feat(tui)!:` will bump 0.6.0 → 1.0.0, not 0.7.0.** — `release-please-config.json`

   The trace, each hop verified against `origin/main`:

   - `.release-please-manifest.json` is `{".":"0.6.0"}`, and `CHANGELOG.md`'s newest heading is `## [0.6.0]`, so release-please's state is consistent and current at 0.6.0.
   - `release-please-config.json` sets `release-type`, `include-component-in-tag`, `bootstrap-sha` and `package-name`. It does **not** set `bump-minor-pre-major`.
   - release-please's default for `bump-minor-pre-major` is `false`, which means a breaking change on a `0.x` version bumps to `1.0.0` rather than to `0.7.0`.
   - The workflow is pinned to `googleapis/release-please-action@a02a34c…` (v4.2.0), so this is not version drift on my part.
   - Nothing has ever exercised this path here: no commit on `main` carries a `BREAKING CHANGE` footer, none uses the `!` marker, and `CHANGELOG.md` has never contained a breaking-changes section. #826 would be the first.

   So the likely outcome of merging as-is is a `v1.0.0` tag and a `1.0.0` publish of `@gitlawb/zero`, triggered by a TUI command rename. Tags and registry publishes are the awkward kind of thing to walk back.

   Two clean resolutions, and this is a maintainer call rather than a code fix:

   - Add `"bump-minor-pre-major": true` to `release-please-config.json` — then `feat(tui)!:` yields **0.7.0** and still records the breaking change. This looks like what the project actually wants, given it is deliberately pre-1.0 and `CHANGELOG.md` says versioning applies "once the first release is tagged".
   - Or accept 1.0.0 knowingly, in which case that decision belongs in the PR description so it is not a surprise in the release PR.

   I want to be precise about what I did and did not do: this is a configuration-reading proof, not an executed one. I did not run release-please against this branch, so I cannot show you the computed version. If you would rather not take the default on faith, the release PR that this merge opens will state the version before anything is tagged — that is a safe place to confirm.

## Non-blocking

- **The `BREAKING CHANGE:` footer in the PR description will not reach the commit.** — PR body. The footer is there and correctly worded (`/retitle is removed; use /rename instead.`), but `squash_merge_commit_message=COMMIT_MESSAGES` builds the squash body from the commit messages, which are `Improve session naming and resume picker` and `Address session rename review feedback`. Neither carries it. The `!` in the subject is enough for release-please to classify the change as breaking, so the entry will appear — it will just appear without the sentence telling users what to do instead, which is the part users need. Either amend a commit to carry the footer, or paste it into the squash body in the merge dialog.

## Carried forward from my previous review, unchanged

- **Four separable changes in one PR.** The `/retitle`→`/rename` swap, the rename editor, the pre-prompt naming plus compare-and-swap, and the `/resume` picker relayout (`session.go:406-436`, `view.go:816-825`) are independent. Team PR through the internal cycle, so not a scope close — but the picker change is the one a bisect would most want isolated.
- **Merge-conflict exposure** with #643 and #570, which both touch `internal/tui/session.go`, `commands.go` and `model.go`.
- **`/rename` is missing from the README command table** (`README.md:214`).
- **Blank submit closes the editor** rather than holding it open (`session_rename.go:32-37`); the test asserts this deliberately, so I read it as intent.

Everything I verified about the code itself at `36f16b7` stands — the compare-and-swap is load-bearing under mutation, the pre-prompt gate is tested, `/retitle` is cleanly excised, and the full gauntlet was green. The only thing standing between this and a merge is which version number it cuts.

anandh8x:
@gnanam1990 removed the breaking marker so this remains a pre-1.0 minor release instead of triggering v1.0.0. Could you re-review and approve if everything looks good?

gnanam1990 (approved):
## Verdict

Approve. Both of my blockers are resolved, and you picked the right resolution for the second one. Apologies for the two rounds of churn — the first ask created the second problem, and that is on me, not on the change.

Re-reviewed at `36f16b7`, base `097c265`. The code tree is byte-identical to my two previous reviews (`git diff 36f16b7..refs/pull/826/head` is empty); the entire delta since my last review is the PR title.

## Withdrawn — both blockers

**1. The changelog entry.** The title is now `feat(tui): replace /retitle with local /rename`. With two commits, `squash_merge_commit_title=COMMIT_OR_PR_TITLE` takes the PR title, so the squash subject is conventional and release-please will record it under Features. The subject also states the removal in plain words, which is most of what a reader scanning the changelog needs.

**2. The 1.0.0 bump.** Dropping the `!` avoids it. Traced to be sure rather than assumed:

- Squash subject: `feat(tui): replace /retitle with local /rename (#826)` — no `!`.
- Squash body comes from `squash_merge_commit_message=COMMIT_MESSAGES`, i.e. `Improve session naming and resume picker` and `Address session rename review feedback`. Neither carries a `BREAKING CHANGE:` footer.
- With neither `!` nor a footer, release-please classifies a plain `feat` → **minor bump, 0.6.0 → 0.7.0**.

That is the safe outcome and it needs no config change to get there.

## Non-blocking — one thing to watch at merge time

**The `BREAKING CHANGE:` footer is still in the PR description.** — PR body

It is no longer reachable by the automation, because the squash body is assembled from commit messages rather than the description. But it is sitting right there in the description, which makes it a natural thing for whoever merges to paste into the squash message. If that happens, release-please sees a breaking change, and with `bump-minor-pre-major` still unset in `release-please-config.json` the bump becomes **1.0.0** after all.

So the 1.0.0 risk is now conditional on merge-time behaviour rather than eliminated. Either is fine:

- Drop the footer from the description, so there is nothing to paste.
- Or leave it and merge with the default squash body, taking care not to add it.

**The better end state is neither of those, and it is not your job here.** Removing a command that shipped in v0.3.0 through v0.6.0 genuinely *is* a breaking change, and the reason it cannot be marked as one is that `release-please-config.json` lacks `"bump-minor-pre-major": true` — so `!` currently means "go to 1.0.0" instead of "minor bump, flagged breaking". Setting that flag would let this and every future breaking change be labelled honestly while the project stays pre-1.0. That is a one-line repo-config PR and it should not hold this one up.

For the record, since it is the thing a user upgrading would want and will not get: `/retitle` existed in v0.3.0, v0.4.0, v0.5.0 and v0.6.0, and after this it is gone with `/rename` as the replacement. It will appear in the changelog as a feature rather than under a breaking-changes heading.

## Carried forward, unchanged and not blocking

- **Four separable changes in one PR** — the `/retitle`→`/rename` swap, the rename editor, the pre-prompt naming plus compare-and-swap, and the `/resume` picker relayout (`session.go:406-436`, `view.go:816-825`). Team PR through the internal cycle, so not a scope close, but the picker change is the one a bisect would most want isolated.
- **Merge-conflict exposure** with #643 and #570, which both touch `internal/tui/session.go`, `commands.go` and `model.go`. This branch is current, so those should rebase onto it.
- **`/rename` is missing from the README command table** (`README.md:214`).
- **Blank submit closes the editor** rather than holding it open (`session_rename.go:32-37`) — asserted deliberately by the test, so read as intent.

## Not re-verified

Nothing about the code needed re-running: the tree is unchanged from my first review, where the compare-and-swap held up under mutation at both layers, the pre-prompt `titledSessions` gate was mutation-tested, `/retitle` was confirmed cleanly excised, and the full gauntlet was green on darwin/arm64. Those results stand.

Separately, and not for this PR: five earlier merges with non-conventional subjects are missing from `CHANGELOG.md` entirely, including `Add Atlas Cloud provider preset (#784)` and `Add shell completion generation (#764)`. That plus the `bump-minor-pre-major` gap are one issue about release hygiene, and this PR is a good illustration of why it is worth filing.

Vasanthdev2004:
Correcting myself on the 1.0.0 warning above, since it didn't play out and I'd rather say why than leave it standing.

The release PR came out as 0.7.0, not 1.0.0. But not because my reading of `bump-minor-pre-major` was wrong — that never got tested. The `!` didn't survive the squash. The merged subject is:

```
feat(tui): replace /retitle with local /rename (#826)
```

No `!`. So release-please saw a plain `feat`, bumped the minor, and filed it under Features. There's no BREAKING CHANGES section in #837.

Whether that was deliberate on merge or just the squash box being edited, I don't know. Either way the practical outcome is the thing gnanam originally raised: v0.7.0 removes `/retitle`, a command that shipped in every release from v0.3.0, and the changelog doesn't mark it breaking. The entry does say "replace /retitle with local /rename", so it's not invisible — someone reading the Features list can work it out. It just isn't flagged.

@kevin-hnnh #837 is still open if you want a `BREAKING CHANGE:` note in there before it goes out. Not blocking anything, and I have no strong view — flagging it because the window closes when that merges.

PR #827, opened 2026-07-29
fix(cli): report MCP servers that failed to start (#822)

### LINKED ISSUE #822: zero mcp check reports success for a server that failed to start
`zero mcp check` is the command we point users at when an MCP server misbehaves, and it reports success for a server that never started.

`runMCPCheck` (`internal/cli/mcp_config.go:301-327`) calls `registerMCPTools`, checks only the returned error, and then unconditionally prints `MCP server %s is reachable. %d tool(s) available.` with status `ok`. It never consults `Runtime.Skipped()`.

That error branch is effectively dead for the case that matters. `RegisterTools` (`internal/mcp/registry.go:154-166`) is deliberately best-effort: a server that fails to connect is recorded in `runtime.skipped` and registration returns nil, so one broken server cannot abort launch. Correct behaviour for startup, wrong thing to ignore in a diagnostic.

The result is that a server whose command does not exist reports "is reachable. 0 tool(s) available." The user's only clue is a tool count of zero, which looks like a server that genuinely exposes no tools.

We already have the data: `SkippedServer{Name, Err, UnconfiguredDefault}` at `registry.go:31-42`. `runMCPCheck` needs to read it and fail with the recorded error. Small, and it makes the other MCP issues easier to diagnose.

Vasanthdev2004:
Fixes #822.

`zero mcp check` looked only at the error returned by registration and then printed "is reachable". `RegisterTools` is best-effort by design: a server that fails to connect is recorded on the runtime and registration returns nil, so one broken server cannot stop Zero launching. Right for startup, wrong here, since this is the command a user runs precisely when a server is misbehaving.

The failure data already existed as `SkippedServer{Name, Err}` and nothing in this path read it. It now fails with the recorded error, in both text and JSON.

The test spawns a genuinely missing command rather than stubbing the runtime. The defect was a mismatch between what registration returns and what it records, so a stub could have been written to pass either way. Reverting the fix makes it print:

```
MCP server broken is reachable. 0 tool(s) available.
```

Two files, 59 lines added.

One note for CI: `TestBuildServeScopeKeepsLexicalPaths` fails on my machine on clean `main` too, because creating a symlink needs a privilege an unelevated Windows shell does not hold. Unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP server checks to reliably detect unreachable servers and surface the underlying connection error when available.
  * Prevented unreachable servers from being incorrectly shown as reachable, including avoiding misleading tool counts.
  * Updated JSON output for `mcp check` to consistently report an unreachable `status` and related error details.
* **Tests**
  * Added coverage to verify unreachable-server behavior and JSON output formatting for `mcp check --json`.

### DIFF
--- a/internal/cli/mcp_config.go
+++ b/internal/cli/mcp_config.go
@@ -308,6 +308,34 @@ func runMCPCheck(ctx context.Context, args []string, stdout io.Writer, stderr io
 	}
 	defer closeMCPRuntime(stderr, mcpRuntime)
 
+	// A connect failure does not come back as err. RegisterTools is deliberately
+	// best-effort so one unreachable server cannot stop Zero launching: it records
+	// the failure on the runtime and returns nil. That is right for startup and
+	// wrong to ignore here, because this is the command a user runs precisely when
+	// a server is misbehaving, and reporting "is reachable" for a server that never
+	// started sends them looking somewhere else.
+	for _, skipped := range mcpRuntime.Skipped() {
+		if skipped.Name != serverName {
+			continue
+		}
+		message := fmt.Sprintf("MCP server %s is not reachable", serverName)
+		if skipped.Err != nil {
+			message = fmt.Sprintf("%s: %s", message, redaction.ErrorMessage(skipped.Err, redaction.Options{}))
+		}
+		if options.json {
+			payload := struct {
+				ServerName string `json:"serverName"`
+				Status     string `json:"status"`
+				Error      string `json:"error,omitempty"`
+			}{ServerName: serverName, Status: "unreachable", Error: message}
+			if err := writePrettyJSON(stdout, redaction.RedactValue(payload, redaction.Options{})); err != nil {
+				return exitCrash
+			}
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
Taken in `41d8dbf5`. Fair catch: I added a `--json` return with its own payload and only tested the text path.

The JSON case is the more useful of the two, because reverting the fix leaves it emitting `status: "ok"` with a tool count, which is still valid JSON and decodes cleanly into a caller expecting success. Now asserts serverName, `status: "unreachable"`, a non-empty error saying so, and that no tool count is implied.

Reverting the fix fails both cases.

@coderabbitai review

jatmn (approved):
LGTM

anandh8x (approved):
Reviewed the latest commit. The skipped-server handling correctly fixes the false reachable result for failed MCP startup, covers both text and JSON output, preserves redaction, and the focused tests and CI are green. No blocking findings.

PR #828, opened 2026-07-29
fix: use OAuth credentials in provider health checks

anandh8x:
## Summary

- pass the existing provider OAuth resolver into health probes
- authenticate connectivity checks with the stored OAuth credential
- refresh the OAuth credential once when a safe, non-generating health request returns 401
- preserve the existing API-key and stored-key paths

## Why

OAuth-authenticated providers worked for normal model requests, but `zero doctor --connectivity` only recognized static API keys and auth headers. A valid OAuth login was therefore reported as missing credentials before the connectivity request was attempted.

## User impact

Doctor and provider-health checks now use the same OAuth credential source as normal provider requests. Users receive the provider's real connectivity result instead of a false missing-credentials failure.

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- verified the saved xAI OAuth profile reaches the provider and reports its actual account response rather than `requires API credentials`

## Summary by CodeRabbit

* **New Features**
  * Provider connectivity diagnostics now support OAuth-based authentication.
  * When OAuth is configured, connectivity checks can refresh OAuth credentials and retry automatically after `401 Unauthorized`.

* **Bug Fixes**
  * Connectivity probing no longer requires a static API key when OAuth credentials are available.
  * Improved auth redaction to hide both static API keys and OAuth bearer tokens in failure output.

* **Tests**
  * Expanded diagnostics/connectivity tests to validate OAuth header selection, refresh/retry behavior, missing/unusable OAuth handling, and token redaction.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -585,6 +585,9 @@ func fillAppDeps(deps appDeps) appDeps {
 	baseProbeProviderHealth := deps.probeProviderHealth
 	deps.probeProviderHealth = func(ctx context.Context, options providerhealth.Options) providerhealth.Result {
 		options.Profile = applyStoredProviderKeyAt(options.Profile, userConfigPath)
+		if options.OAuthResolver == nil {
+			options.OAuthResolver, _ = oauthLoginForProfile(options.Profile)
+		}
 		return baseProbeProviderHealth(ctx, options)
 	}
 	return deps
--- a/internal/cli/observability_test.go
+++ b/internal/cli/observability_test.go
@@ -114,6 +114,7 @@ func TestRunDoctorConnectivityProbesProvider(t *testing.T) {
 
 func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
 	tokenPath := filepath.Join(t.TempDir(), "oauth-tokens.json")
+	userConfigPath := filepath.Join(t.TempDir(), "config.json")
 	t.Setenv("ZERO_OAUTH_STORAGE", "file")
 	t.Setenv("ZERO_OAUTH_TOKENS_PATH", tokenPath)
 	store, err := oauth.NewStore(oauth.StoreOptions{FilePath: tokenPath})
@@ -147,6 +148,9 @@ func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
 	exitCode := runWithDeps([]string{"doctor", "--connectivity"}, &stdout, &stderr, appDeps{
 		getwd: func() (string, error) {
 			return t.TempDir(), nil
+		},
+		userConfigPath: func() (string, error) {
+			return userConfigPath, nil
 		},
 		resolveConfig: func(string, config.Overrides) (config.ResolvedConfig, error) {
 			return config.ResolvedConfig{
--- a/internal/cli/observability_test.go
+++ b/internal/cli/observability_test.go
@@ -114,7 +114,7 @@ func TestRunDoctorConnectivityProbesProvider(t *testing.T) {
 
 func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
 	tokenPath := filepath.Join(t.TempDir(), "oauth-tokens.json")
-	userConfigPath := filepath.Join(t.TempDir(), "config.json")
+	t.Setenv("XDG_CONFIG_HOME", t.TempDir())
 	t.Setenv("ZERO_OAUTH_STORAGE", "file")
 	t.Setenv("ZERO_OAUTH_TOKENS_PATH", tokenPath)
 	store, err := oauth.NewStore(oauth.StoreOptions{FilePath: tokenPath})
@@ -148,9 +148,6 @@ func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
 	exitCode := runWithDeps([]string{"doctor", "--connectivity"}, &stdout, &stderr, appDeps{
 		getwd: func() (string, error) {
 			return t.TempDir(), nil
-		},
-		userConfigPath: func() (string, error) {
-			return userConfigPath, nil
 		},
 		resolveConfig: func(string, config.Overrides) (config.ResolvedConfig, error) {
 			return config.ResolvedConfig{
--- a/internal/cli/observability_test.go
+++ b/internal/cli/observability_test.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strings"
@@ -12,6 +14,7 @@ import (
 	"time"
 
 	"github.com/Gitlawb/zero/internal/config"
+	"github.com/Gitlawb/zero/internal/oauth"
 	"github.com/Gitlawb/zero/internal/providerhealth"
 	"github.com/Gitlawb/zero/internal/sessions"
 )
@@ -106,6 +109,64 @@ func TestRunDoctorConnectivityProbesProvider(t *testing.T) {
 	}
 	if strings.Contains(output, "not wired") {
 		t.Fatalf("doctor still returned placeholder connectivity message: %q", output)
+	}
+}
+
+func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
+	tokenPath := filepath.Join(t.TempDir(), "oauth-tokens.json")
+	t.Setenv("ZERO_OAUTH_STORAGE", "file")
+	t.Setenv("ZERO_OAUTH_TOKENS_PATH", tokenPath)
+	store, err := oauth.NewStore(oauth.StoreOptions{FilePath: tokenPath})
+	if err != nil {
+		t.Fatalf("NewStore: %v", err)
+	}
+	if err := store.Save(oauth.ProviderKey("xai"), oauth.Token{
+		AccessToken: "oauth-doctor-token",
+		ExpiresAt:   time.Now().Add(time.Hour),
+	}); err != nil {
+		t.Fatalf("seed OAuth token: %v", err)
+	}
+
+	var gotAuth string
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotAuth = r.Header.Get("Authorization")
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":[]}`))
+	}))
+	defer server.Close()
+
+	profile := config.ProviderProfile{
+		Name:         "xai",
+		CatalogID:    "xai",
+		ProviderKind: config.ProviderKindOpenAICompatible,
+		BaseURL:      server.URL + "/v1",
+		Model:        "grok-4",
+	}
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	exitCode := runWithDeps([]string{"doctor", "--connectivity"}, &stdout, &stderr, appDeps{
+		getwd: func() (string, error) {
+			return t.TempDir(), nil
+		},
+		resolveConfig: func(string, config.Overrides) (config.ResolvedConfig, error) {
+			return config.ResolvedConfig{
+				Provider:       profile,
+				Providers:      []config.ProviderProfile{profile},
+				ActiveProvider: profile.Name,
+			}, nil
+		},
+		probeProviderHealth: providerhealth.Probe,
+		now:                 fixedCLITime("2026-07-29T20:00:00Z"),
+	})
+
+	if exitCode != exitSuccess {
+		t.Fatalf("doctor exit = %d, want success; stderr=%q\n%s", exitCode, stderr.String(), stdout.String())
+	}
+	if gotAuth != "Bearer oauth-doctor-token" {
+		t.Fatalf("Authorization = %q, want OAuth bearer", gotAuth)
+	}
+	if !strings.Contains(stdout.String(), "[pass] provider.connectivity") {
+		t.Fatalf("doctor did not report passing connectivity:\n%s", stdout.String())
 	}
 }
 
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -209,7 +209,7 @@ func Probe(ctx context.Context, options Options) Result {
 	}, profile))
 
 	oauthConfigured := options.OAuthResolver != nil
-	if options.Connectivity && oauthConfigured {
+	if oauthConfigured {
 		var resolveErr error
 		profile, oauthConfigured, resolveErr = profileWithOAuthCredential(ctx, profile, options.OAuthResolver, false)
 		if resolveErr != nil {
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -373,7 +373,6 @@ func profileWithOAuthCredential(ctx context.Context, profile config.ProviderProf
 	if strings.TrimSpace(header) == "" {
 		header = "Authorization"
 	}
-	profile.APIKey = ""
 	profile.AuthHeader = header
 	profile.AuthScheme = ""
 	profile.AuthHeaderValue = value
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -807,6 +807,9 @@ func redact(message string, profile config.ProviderProfile) string {
 
 func providerSecrets(profile config.ProviderProfile) []string {
 	secrets := []string{profile.APIKey, profile.AuthHeaderValue}
+	if fields := strings.Fields(profile.AuthHeaderValue); len(fields) == 2 {
+		secrets = append(secrets, fields[1])
+	}
 	for _, value := range profile.CustomHeaders {
 		if strings.TrimSpace(value) != "" {
 			secrets = append(secrets, value)
--- a/internal/providerhealth/providerhealth.go
+++ b/internal/providerhealth/providerhealth.go
@@ -46,12 +46,13 @@ const (
 const defaultTimeout = 5 * time.Second
 
 type Options struct {
-	Profile      config.ProviderProfile
-	Connectivity bool
-	HTTPClient   *http.Client
-	Resolver     Resolver
-	Timeout      time.Duration
-	UserAgent    string
+	Profile       config.ProviderProfile
+	Connectivity  bool
+	HTTPClient    *http.Client
+	Resolver      Resolver
+	OAuthResolver providerio.TokenResolver
+	Timeout       time.Duration
+	UserAgent     string
 }
 
 type Resolver interface {
@@ -207,11 +208,26 @@ func Probe(ctx context.Context, options Options) Result {
 		"providerKind": metadata.ProviderKind,
 	}, profile))
 
-	if credentialRequired(profile, metadata.ProviderKind) && !hasCredential(profile) {
+	oauthConfigured := options.OAuthResolver != nil
+	if options.Connectivity && oauthConfigured {
+		var resolveErr error
+		profile, oauthConfigured, resolveErr = profileWithOAuthCredential(ctx, profile, options.OAuthResolver, false)
+		if resolveErr != nil {
+			result.add(check("provider.auth", "Provider auth", StatusFail, CategoryAuth, "Provider OAuth credential could not be resolved: "+resolveErr.Error(), map[string]any{
+				"oauthLogin": true,
+			}, profile))
+			return result.finalize()
+		}
+	}
+	if credentialRequired(profile, metadata.ProviderKind) && !hasCredential(profile) && !oauthConfigured {
 		result.add(check("provider.auth", "Provider auth", StatusFail, CategoryAuth, fmt.Sprintf("Provider %s requires API credentials.", providerName(profile)), credentialDetails(profile), profile))
 		return result.finalize()
 	}
-	if hasCredential(profile) {
+	if oauthConfigured {
+		result.add(check("provider.auth", "Provider auth", StatusPass, CategoryAuth, fmt.Sprintf("Provider %s has OAuth credentials configured.", providerName(profile)), map[string]any{
+			"oauthLogin": true,
+		}, profile))
+	} else if hasCredential(profile) {
 		result.add(check("provider.auth", "Provider auth", StatusPass, CategoryAuth, fmt.Sprintf("Provider %s has credentials configured.", providerName(profile)), credentialDetails(profile), profile))
 	} else {
 		result.add(check("provider.auth", "Provider auth", StatusPass, CategoryAuth, fmt.Sprintf("Provider %s does not require API credentials.", providerName(profile)), credentialDetails(profile), profile))
@@ -286,9 +302,35 @@ func connectivityCheck(ctx context.Context, profile config.ProviderProfile, kind
 	if err != nil {
 		return classifyTransportError(err, profile)
 	}
-	defer func() {
-		_ = response.Body.Close()
-	}()
+	defer func(body io.ReadCloser) {
+		_ = body.Close()
+	}(response.Body)
+	if response.StatusCode == http.StatusUnauthorized && options.OAuthResolver != nil {
+		refreshed, ok, refreshErr := profileWithOAuthCredential(requestCtx, options.Profile, options.OAuthResolver, true)
+		if refreshErr != nil {
+			return check("provider.connectivity", "Provider connectivity", StatusFail, CategoryAuth, "Provider OAuth credential refresh failed: "+refreshErr.Error(), map[string]any{
+				"statusCode": response.StatusCode,
+			}, profile)
+		}
+		if ok {
+			_ = response.Body.Close()
+			profile = refreshed
+			request, allowLoopbackOrPrivate, err = healthRequest(requestCtx, profile, kind, options)
+			if err != nil {
+				return check("provider.connectivity", "Provider connectivity", StatusFail, CategoryConfig, err.Error(), nil, profile)
+			}
+			if options.HTTPClient == nil {
+				client = newConnectivityClient(timeout, options.Resolver, sensitiveAuthHeaderNames(profile, kind), allowLoopbackOrPrivate)
+			}
+			response, err = client.Do(request)
+			if err != nil {
+				return classifyTransportError(err, profile)
+			}
+			defer func(body io.ReadCloser) {
+				_ = body.Close()
+			}(response.Body)
+		}
+	}
 
 	body, err := io.ReadAll(io.LimitReader(response.Body, 64*1024))
 	if err != nil {
@@ -318,6 +360,24 @@ func connectivityCheck(ctx context.Context, profile config.ProviderProfile, kind
 		"statusCode": response.StatusCode,
 		"endpoint":   request.URL.String(),
 	}, profile)
+}
+
+func profileWithOAuthCredential(ctx context.Context, profile config.ProviderProfile, resolver providerio.TokenResolver, forceRefresh bool) (config.ProviderProfile, bool, error) {
+	if resolver == nil {
+		return profile, false, nil
+	}
+	header, value, ok, err := resolver(ctx, forceRefresh)
+	if err != nil || !ok {
+		return profile, false, err
+	}
+	if strings.TrimSpace(header) == "" {
+		header = "Authorization"
+	}
+	profile.APIKey = ""
+	profile.AuthHeader = header
+	profile.AuthScheme = ""
+	profile.AuthHeaderValue = value
+	return profile, true, nil
 }
 
 func healthRequest(ctx context.Context, profile config.ProviderProfile, kind config.ProviderKind, options Options) (*http.Request, bool, error) {
--- a/internal/providerhealth/providerhealth_test.go
+++ b/internal/providerhealth/providerhealth_test.go
@@ -221,15 +221,18 @@ func TestProbeConnectivityUsesOAuthCredential(t *testing.T) {
 	}
 }
 
-func TestProbeConnectivityOAuthKeepsStaticKeyRedacted(t *testing.T) {
-	const staticKey = "STATICKEYVALUE1234567890abcdef"
+func TestProbeConnectivityOAuthCredentialsAreRedacted(t *testing.T) {
+	const (
+		staticKey  = "STATICKEYVALUE1234567890abcdef"
+		oauthToken = "hf_VASANTHSENTINEL0011223344"
+	)
 	var authHeaders []string
 	client := &http.Client{Transport: roundTripFunc(func(r *http.Request) (*http.Response, error) {
 		authHeaders = append(authHeaders, r.Header.Get("Authorization"))
 		return &http.Response{
 			StatusCode: http.StatusUnauthorized,
 			Status:     "401 Unauthorized",
-			Body:       io.NopCloser(strings.NewReader(`{"error":{"message":"invalid key ` + staticKey + `"}}`)),
+			Body:       io.NopCloser(strings.NewReader(`{"error":{"message":"invalid credentials ` + staticKey + ` and ` + oauthToken + `"}}`)),
 			Header:     make(http.Header),
 		}, nil
 	})}
@@ -247,7 +250,7 @@ func TestProbeConnectivityOAuthKeepsStaticKeyRedacted(t *testing.T) {
 		HTTPClient:   client,
 		Resolver:     staticResolver{addr: netip.MustParseAddr("93.184.216.34")},
 		OAuthResolver: func(context.Context, bool) (string, string, bool, error) {
-			return "Authorization", "Bearer oauth-test-token", true, nil
+			return "Authorization", "Bearer " + oauthToken, true, nil
 		},
 	})
 
@@ -255,7 +258,7 @@ func TestProbeConnectivityOAuthKeepsStaticKeyRedacted(t *testing.T) {
 		t.Fatalf("Authorization headers = %#v, want initial request and one refresh retry", authHeaders)
 	}
 	for _, header := range authHeaders {
-		if header != "Bearer oauth-test-token" {
+		if header != "Bearer "+oauthToken {
 			t.Fatalf("Authorization = %q, want OAuth bearer", header)
 		}
 	}
@@ -266,8 +269,11 @@ func TestProbeConnectivityOAuthKeepsStaticKeyRedacted(t *testing.T) {
 	if strings.Contains(check.Message, staticKey) {
 		t.Fatalf("static API key leaked in message: %q", check.Message)
 	}
-	if !strings.Contains(check.Message, "[REDACTED]") {
-		t.Fatalf("expected static API key to be redacted, got %q", check.Message)
+	if strings.Contains(check.Message, oauthToken) {
+		t.Fatalf("OAuth token leaked in message: %q", check.Message)
+	}
+	if strings.Count(check.Message, "[REDACTED]") != 2 {
+		t.Fatalf("expected both credentials to be redacted, got %q", check.Message)
 	}
 }
 
--- a/internal/providerhealth/providerhealth_test.go
+++ b/internal/providerhealth/providerhealth_test.go
@@ -180,6 +180,98 @@ func TestProbeResolvedKindConnectivityAuthErrorRedactsSecret(t *testing.T) {
 	}
 }
 
+func TestProbeConnectivityUsesOAuthCredential(t *testing.T) {
+	var gotAuth string
+	resolverCalled := false
+	client := &http.Client{Transport: roundTripFunc(func(r *http.Request) (*http.Response, error) {
+		gotAuth = r.Header.Get("Authorization")
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Status:     "200 OK",
+			Body:       io.NopCloser(strings.NewReader(`{"data":[]}`)),
+			Header:     make(http.Header),
+		}, nil
+	})}
+
+	result := Probe(context.Background(), Options{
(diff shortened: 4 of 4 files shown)


gnanam1990 (changes requested):
## Verdict

Changes requested. The feature is right and the wiring is correct, but one line in it silently disables redaction of the user's static API key, and I have a reproduction.

Reviewed at `10596d1`, base `097c265` (level with `origin/main`). Author has write access, so this is the internal cycle rather than the `issue-approved` gate.

## What I verified

- **The OAuth credential does reach the probe**, and `Probe` passes the OAuth-mutated profile through to `connectivityCheck` (`providerhealth.go:239`), so the first connectivity request carries the token rather than only the retry.
- **Redirect-stripping is recomputed after a refresh.** `sensitiveAuthHeaderNames(profile, kind)` is re-derived from the refreshed profile before the retry client is built (`providerhealth.go:321`), so the new header name is protected too. That is the easy thing to get wrong here and it is right.
- **Concurrency: not an issue, checked rather than assumed.** I expected trouble from a 401-retry inside a probe that might run concurrently, and there is none — this is a single sequential probe.
- **The double `Close` is harmless.** The first `defer` binds the original body by value, and the explicit `_ = response.Body.Close()` at `:314` closes it again; `http.body.Close` is safe to call twice. Redundant, not a bug.
- Gauntlet on darwin/arm64 from a non-`/tmp` checkout: `make fmt-check`, `go vet ./...`, `git diff HEAD --check` all clean; `go test ./internal/providerhealth/ ./internal/cli/ ./internal/providers/...` green; `go run ./cmd/zero-release build` and `smoke` both pass.
- **Passes executed:** all 12, since this is credential-handling on a network path.

## Blocking

1. **Clearing `profile.APIKey` removes the API key from the redaction set, and the provider's response body is echoed into the check message.** — `internal/providerhealth/providerhealth.go:373`

   The trace, each hop by name:

   - `profileWithOAuthCredential` sets `profile.APIKey = ""` (`:373`).
   - That same `profile` is what `Probe` hands to `connectivityCheck` (`:239`), and what every `check(...)` call receives as its last argument.
   - `check` redacts via `redact(message, profile)` (`:767`), which builds its secret list from `providerSecrets(profile)` = `{profile.APIKey, profile.AuthHeaderValue}` (`:810`).
   - With `APIKey` blanked, the static key is no longer in that list.
   - The provider's response body is placed directly into the message: `message := responseMessage(response.StatusCode, body)` (`:358`).

   So a provider that echoes the offending key in an error body — which several do on 401 — gets it printed verbatim in `zero doctor --connectivity`, including `--json`. I ran both paths against the patched build with a static key that matches no built-in secret pattern, so only the profile-derived redaction could catch it:

   ```
   API-key path         : static key leaked = false
      message: Provider endpoint returned 401: {"error":"invalid key [REDACTED]"}

   OAuth path (this PR) : static key leaked = true
      message: Provider endpoint returned 401: {"error":"invalid key STATICKEYVALUE1234567890abcdef"}
   ```

   Same profile, same body, same assertions — the only difference is whether `OAuthResolver` is set. This is a regression against current `main` behaviour for any user who has both a stored API key and an OAuth login.

   **The fix is one line: drop `profile.APIKey = ""`.** Clearing it buys nothing on the request path — `ApplyAuthHeaders` (`internal/providers/providerio/headers.go:36-41`) only consults `APIKey` when `AuthHeaderValue` is empty, and writes exactly one auth header — so the OAuth credential still wins. I applied that change and re-ran: the leak closes, and the full `internal/providerhealth` suite including `TestProbeConnectivityUsesOAuthCredential` still passes.

   ```
   OAuth path (with fix): static key leaked = false
      message: Provider endpoint returned 401: {"error":"invalid key [REDACTED]"}
   ```

   If you would rather keep the clear for defence-in-depth, then `providerSecrets` needs to retain the pre-OAuth key instead — but the redaction set must not shrink when OAuth is in play.

2. **This duplicates `providerio.withBearer` and `providerio.SendWithAuthRetry`, which already exist and already do exactly this.** — `internal/providerhealth/providerhealth.go:365-380` and `:307-330`

   `withBearer` (`internal/providers/providerio/auth.go:80-89`) applies an OAuth credential to an auth config with the identical rule set: default the header to `Authorization`, clear `APIKey`, clear `AuthScheme`, set `AuthHeaderValue`. `SendWithAuthRetry` (`:31-77`) implements the same resolve-then-retry-once-on-401-with-`forceRefresh` loop that `connectivityCheck` now open-codes.

   The two implementations agree today — I compared them field by field, so this is drift risk rather than a present defect. But it is drift risk on the rule that decides which credential leaves the machine, and `withBearer`'s own comment states the invariant it exists to hold: *"the API key is cleared so the two auth methods can never both be sent."* Finding 1 is what that duplication already cost, in the form of a redaction property the original author never had to think about because their copy was not used for redaction.

   Health probing has real reasons not to call `SendWithAuthRetry` wholesale (it must not inherit the 429/503 retry budget, and it needs its own SSRF-guarded client), so I am not asking for that. Reusing `withBearer` — or exporting an equivalent — for the credential-application half would remove the duplicated rule while leaving the probe's transport alone.

## Non-blocking

- **The OAuth lookup error is discarded.** — `internal/cli/app.go:589`. `options.OAuthResolver, _ = oauthLoginForProfile(options.Profile)` drops the error, so a genuinely broken keyring is indistinguishable from "no OAuth login configured" and the user is told their provider *requires API credentials*. Diagnostics is precisely the context where that distinction is the answer the user came for. Surfacing it as a warn check would be more useful than the current silence.
- **The refresh reads `options.Profile` where everything around it uses `profile`.** — `providerhealth.go:308`. These are equivalent today, because `profileWithOAuthCredential` overwrites all four auth fields and nothing else mutates the profile between `Probe` and here. It is a readability trap rather than a bug: a future mutation applied to `profile` before the connectivity check would be silently dropped on the retry path only.
- **The explicit close at `:314` is redundant** given the deferred close bound at `:305`. Harmless, but it reads as if one of them is load-bearing.
- **`hasCredential(profile)` is already true by the time `!oauthConfigured` is checked.** — `providerhealth.go:221`. Once `profileWithOAuthCredential` has set `AuthHeaderValue`, `hasCredential` returns true (`:831`), so `&& !oauthConfigured` is belt-and-braces. Correct, just worth a comment so nobody removes the wrong half.

## Questions for the author

- Was the `profile.APIKey = ""` line copied deliberately from `withBearer`, or arrived at independently? It matters for finding 2: if it was copied, the case for reusing the original rather than duplicating it is stronger.

## Not covered by this PR

- No test asserts the "never send both credentials" property. I mutated the `APIKey` clear away and the whole `internal/providerhealth` suite stayed green, which is how finding 1 surfaced. Whichever way finding 1 is resolved, a test that pins the redaction behaviour with both a static key and an OAuth login present is the one worth adding.
- The 401-refresh path is exercised for success and for resolver error, but not for a refresh that returns `ok=false`. That path falls through and reports the original 401, which is right — it just is not pinned.

## Not verified

- The full `go test ./...` run. I ran `internal/providerhealth`, `internal/cli` and `internal/providers/...`, plus the release build and smoke; the whole-repo run exceeded my time budget and I would rather say so than imply it passed.
- Any platform other than darwin/arm64.
- A real xAI OAuth profile end to end. The PR description reports that manually; my verification is through the package's own harness with a fake resolver and transport.

gnanam1990 (dismissed):
## Verdict

Approve. Blocker 1 is fixed and I verified it against real key formats rather than just the synthetic one from my last review. Blocker 2 I am withdrawing — the fix demonstrates that the two implementations *should* differ, which I had wrong.

Re-reviewed at `26cc8ed`, base `097c265`. Previously reviewed at `10596d1`; the delta is one commit, `26cc8ed "fix(providerhealth): retain API key redaction"` — one line removed from `providerhealth.go`, one test added.

## Withdrawn — blocker 1, the redaction regression

Fixed, and verified beyond what my original evidence showed. My previous review demonstrated the leak with a single synthetic key that matches no built-in secret pattern, which was deliberate — it isolates the profile-derived redaction — but it left open the fair objection that a *real* provider key might be caught by the generic value matcher anyway. So I measured the formats that actually occur, at both SHAs:

| Key format | `10596d1` (before) | `26cc8ed` (now) |
|---|---|---|
| `xai-8Kd0zQm3Rt7Vb1Xn5Lp9Ws2Yc4Hf6Jg` | **leaked** | redacted |
| `mycompany-internal-gateway-key-2026` | **leaked** | redacted |
| `8f14e45f-ceea-467a-9f3a-1b2c3d4e5f60` | **leaked** | redacted |
| `sk-proj-…` / `sk-ant-api03-…` | redacted | redacted |

Worth recording why this mattered more than my first review conveyed: only `sk-`-prefixed keys were ever safe, because `redaction.RedactString`'s pattern matcher catches those independently of the profile secret set. Everything else depended entirely on `providerSecrets(profile)`. That included `xai-` — the provider this PR was written for, per its own description.

**The new regression test is load-bearing.** Reinstating the deleted line fails it:

```
--- FAIL: TestProbeConnectivityOAuthKeepsStaticKeyRedacted
    providerhealth_test.go:267: static API key leaked in message:
    "Provider endpoint returned 401: invalid key STATICKEYVALUE1234567890abcdef"
```

And the test does more than assert the negative. Pinning `len(authHeaders) == 2` and requiring both to be `Bearer oauth-test-token` (`providerhealth_test.go:253-260`) is the right complement: it proves that retaining `profile.APIKey` did **not** cause the static key to be sent, and it covers the refresh-retry path at the same time. That is the exact risk in the fix I proposed, and it is now pinned rather than argued.

## Withdrawn — blocker 2, the duplication with `withBearer`

I asked for `profileWithOAuthCredential` to be unified with `providerio.withBearer` on the grounds that two copies of the credential-application rule will drift. **Following that advice would have reintroduced blocker 1.**

`withBearer` clears `APIKey` correctly, because its output feeds `ApplyAuthHeaders` and nothing else. `profileWithOAuthCredential` must retain it, because its output doubles as the redaction secret source via `providerSecrets` (`providerhealth.go:810`). The two genuinely need different behaviour, so they are not duplicates — they are two different jobs that happen to share a struct.

The real issue is the one underneath: `config.ProviderProfile` is overloaded as both "request auth config" and "set of strings to scrub from output," and mutating it for the first purpose silently changes the second. A redaction secret set that *accumulates* rather than being re-derived from the mutated profile would make this class unrepresentable. That is a refactor well beyond this PR, and worth its own issue rather than a review comment — but if you file it, this PR is the reproduction.

- Both key-format tables above, run against `10596d1` and `26cc8ed` in the same harness.
- The mutation that reinstates `profile.APIKey = ""`, which fails the new test.
- `go test ./internal/providerhealth/ -count=1` green; `make fmt-check`, `go vet ./...`, `git diff HEAD --check` all clean, darwin/arm64, non-`/tmp` checkout.
- **Two `internal/cli` failures are pre-existing and not yours.** `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` fail with `exit code 3` at this head, at `10596d1`, and identically at base `097c265` — the same two names in all three. The cause is my environment: `gopls`, `pyright-langserver` and `typescript-language-server` are absent from my PATH, so `zero doctor` returns a warn-level exit. Nothing to do with this change; I am recording it so the next reviewer does not attribute it to you.
- **Passes executed:** 0, 1, 2, 3, 7, 9, 11, 12, scoped to the delta against a production change I had already reviewed in full at `10596d1`.

## Still open, non-blocking (carried forward)

- **The OAuth lookup error is discarded.** — `internal/cli/app.go:589`. `options.OAuthResolver, _ = oauthLoginForProfile(options.Profile)` drops the error, so a broken keyring is indistinguishable from "no OAuth login configured" and the user is told their provider *requires API credentials*. Diagnostics is exactly the context where that distinction is the answer the user came for.
- **The refresh reads `options.Profile` where everything around it uses `profile`.** — `providerhealth.go:308`. Equivalent today; a readability trap rather than a bug.
- **The explicit `Close` at `:314` is redundant** given the deferred close bound at `:305`.
- **`hasCredential(profile)` is already true when `!oauthConfigured` is evaluated** — `:221`. Belt-and-braces, worth a comment so nobody removes the wrong half.

## Not verified

- The full `go test ./...` run; I scoped to `internal/providerhealth` and `internal/cli` plus the checks above.
- Any platform other than darwin/arm64.
- A real xAI OAuth profile end to end. The PR description reports that manually; my verification is through the package harness with a fake resolver and transport.

Good fix, and the test you added is stronger than the one I asked for.

Vasanthdev2004 (changes requested):
Reviewed at 26cc8ed1, base 097c2652. gnanam's blocker is genuinely fixed — `profile.APIKey = ""` is gone and `TestProbeConnectivityOAuthKeepsStaticKeyRedacted` pins it.

But it's the sibling case that's still open, and the name of that new test is a good hint as to why it got missed: it asserts the *static key* is redacted while OAuth is active. Nothing asserts the OAuth token — the credential actually being sent — is redacted. It isn't.

## Blocking: the OAuth access token is printed verbatim by `zero doctor --connectivity`

Same sentinel value, same server, same echoed-back error body. The only difference is which path supplied the credential:

```
static API key (control):  credential [REDACTED] is not authorized
OAuth resolver (this PR):  credential hf_VASANTHSENTINEL0011223344 is not authorized
```

It's in `check.Message`, so it lands in `zero doctor --connectivity`, in `--json`, and in the `/doctor` transcript. People paste doctor output into bug reports.

The cause is a field-convention inversion. `oauth_provider.go:70` hands back `"Bearer " + token` as one string; `profileWithOAuthCredential` puts that whole thing in `AuthHeaderValue` and blanks `AuthScheme` (:377). But `providerSecrets` registers `AuthHeaderValue` as a literal, and `redact` does a literal replace — so it only ever matches the string `Bearer <token>`, never a bare `<token>` sitting in a provider's JSON error. Everywhere else the scheme lives in `AuthScheme` and the value stays bare, which is exactly why the API-key path is safe. `redaction.RedactString` doesn't rescue it either: it's pattern-based and these token shapes don't match.

Worth saying that this is a threat the repo already decided it cares about — `TestProbeResolvedKindConnectivityAuthErrorRedactsSecret` exists for precisely the "provider echoes your credential back at you" case. The OAuth path just doesn't inherit that protection.

New in this PR, not pre-existing: `git show 097c2652:internal/providerhealth/providerhealth.go | grep -ci oauth` is 0.

Smallest fix that works — register the scheme-stripped token too, in `providerSecrets`:

```go
if fields := strings.Fields(profile.AuthHeaderValue); len(fields) == 2 {
    secrets = append(secrets, fields[1])
}
```

I applied it: my repro flips to green and the full `internal/providerhealth` suite stays green. Splitting the scheme back into `AuthScheme` instead would also work and is arguably tidier, but the `providerSecrets` version also covers any future caller that inlines a scheme, so I'd lean that way. Either way it needs a test that asserts *the token* is redacted, not just the key beside it.

## Non-blocking

**The refresh-failure branch is never exercised.** Mutating `if refreshErr != nil` at :310 to `if false && refreshErr != nil` leaves `internal/providerhealth` and the doctor tests green. Both new 401 tests use resolvers that never fail, so the message, the category, and the decision to stop rather than report the original 401 are all unpinned.

**Only the positive direction of the 401 trigger is tested.** Broadening :308 from 401 to 401-or-403 also survives. That matters more than a normal gap: it would mean every 403 a provider returns — permission denied, region block, org quota — fires a forced token refresh against the auth server on each `zero doctor` run, which can rotate or consume the refresh token. Narrowing it the other way (401 to 403) *is* caught by both new tests, so the site is genuinely exercised and this is one-sided coverage rather than dead code.

**`Options.Timeout` doesn't cover the OAuth resolve.** :214 calls `profileWithOAuthCredential(ctx, ...)` with the caller's raw context, and `requestCtx` isn't created until :285 inside `connectivityCheck`. Under `zero doctor` that ctx comes from `signalContext()` and carries no deadline, so an operation that now does disk I/O and possibly a token-endpoint round trip sits outside the probe's 5s budget entirely — the only bound left is the OAuth manager's own 30s client. Wrapping ctx before :214 is a one-liner. The post-401 refresh at :309 has the opposite problem: it shares `requestCtx`, so it runs on whatever's left after the first request, which is how a refresh gets cut short and reported as an auth failure.

Only the first one blocks. The direction of this PR is right — doctor reporting "missing credentials" to someone who is logged in is a real annoyance and worth fixing.

On my box `TestBuildServeScopeKeepsLexicalPaths` fails in `internal/cli` for want of Windows symlink privilege; it fails the same way on clean main, so ignore it in anything above.

Vasanthdev2004 (dismissed):
Approving at fb2dc8ae. The leak is closed and, more importantly, it's pinned.

I re-ran my own reproduction rather than trusting the diff — same sentinel down both paths, and the OAuth path now matches the API-key control:

```
static API key (control):  credential [REDACTED] is not authorized
OAuth resolver:            credential [REDACTED] is not authorized
```

And the new test isn't decorative: deleting the three-line `providerSecrets` change makes `TestProbeConnectivityOAuthCredentialsAreRedacted` fail, so the fix can't be quietly reverted later without CI noticing. Renaming it away from "KeepsStaticKeyRedacted" was the right call too — the old name was most of the reason the gap survived the first round.

The three non-blocking items from my last review are unchanged, which is fine by me and not worth another round-trip. Recording where they stand so nobody has to re-derive it:

- Deleting the refresh-failure branch at :310 still leaves the suite green.
- Broadening the 401 trigger at :308 to 401-or-403 still survives. Narrowing it the other way is caught, so it's genuinely one-sided coverage rather than dead code.
- `Options.Timeout` still doesn't reach the OAuth resolve at :214.

The timeout one is the only one I'd bother with eventually, and it's a one-line `context.WithTimeout` — a follow-up issue is fine.

One small thing while you're in there, not worth a push on its own: the guard is `len(fields) == 2`, so a value with three or more fields falls through and the bare token wouldn't be registered. Not reachable through the OAuth resolver, which always produces exactly `Bearer <token>` — `>= 2` with `fields[1]` would just close the door.

gofmt and vet clean, `internal/providerhealth`, `internal/providers/...` all green. `TestBuildServeScopeKeepsLexicalPaths` fails in `internal/cli` on my box for want of Windows symlink privilege, same as on clean main.

Vasanthdev2004:
@gnanam1990 re-requested you — fb2dc8ae dismissed your approval along with CodeRabbit's, so this is sitting on one of three.

Short version of what changed since you last looked: your static-key leak was fixed at 26cc8ed1, then I found the sibling case (the OAuth token itself was still going out verbatim in the doctor message) and fb2dc8ae closes that too. The test now asserts both credentials are redacted rather than just the one beside the credential in use, and I confirmed deleting the fix makes it fail.

gnanam1990 (changes requested):
## Verdict

Changes requested — one item, in a test, with a one-line fix and five examples of the right pattern already in this package. The OAuth feature itself is working and the new redaction fix is good; I verified both. My previous approval was dismissed by the push, so this supersedes it.

Re-reviewed at `fb2dc8ae`, base `097c265`. Previously approved at `26cc8ed`; the delta is one commit, `fb2dc8ae "fix(providerhealth): redact bare OAuth tokens"`.

## First, a correction to my last review

I told you the two failing `internal/cli` doctor tests were caused by missing language servers on my PATH. **That was wrong.** `lsp.servers` reports `[warn]`, and `finalize()` only escalates to `fail` on a `StatusFail` check. The actual cause is this line in the doctor output:

```
[fail] config.validation - Zero config validation found 1 issue(s).
  /Users/kratos/.config/zero/config.json: issues: [unknown config field "profiles"]
```

The tests read my **real** `~/.config/zero/config.json`. That changes the conclusion, which is why it matters rather than being a footnote — see the blocker.

- **The new bare-token fix is real and its test is load-bearing.** `providerSecrets` previously registered only the composite `AuthHeaderValue` (`"Bearer <token>"`), so a provider echoing the bare token would not have matched. Removing the three added lines fails the test:

  ```
  --- FAIL: TestProbeConnectivityOAuthCredentialsAreRedacted
      OAuth token leaked in message: "Provider endpoint returned 401:
      invalid credentials [REDACTED] and hf_VASANTHSENTINEL0011223344"
  ```

  Good catch, and one I missed on the previous head.
- **I checked whether that test was vacuous and it is not.** The sentinel uses an `hf_` prefix, which is the shape that normally gets caught by the generic value matcher independently of the profile secret set — the exact trap that makes redaction fixtures lie. There is no `hf_` pattern in `internal/secrets/scanner.go` or `internal/redaction`, and the mutation above confirms the profile-derived path is the only thing redacting it. `strings.Count(..., "[REDACTED]") != 2` is also the right assertion: it pins both credentials rather than accepting one.
- **`len(fields) == 2` is adequate for every credential this code can actually receive.** The only producer is `internal/cli/oauth_provider.go:70`, which returns `"Bearer " + token` — always two fields. A one-field value is still covered by the composite entry. Noted below as a latent edge rather than a defect.
- **This aligns `providerhealth` with existing practice** rather than inventing something: `internal/config/redact.go:24` and `internal/streamjson/streamjson.go:387` already strip a `"Bearer "` prefix for the same reason. `providerhealth` was the outlier.
- **The OAuth feature demonstrably works.** In the failing test's own output: `[pass] provider.config — credentialConfigured: oauth login` and `[pass] provider.connectivity — Provider endpoint reachable (200)`. The credential reaches the endpoint. The test fails on its overall-exit assertion, not on anything about OAuth.
- `make fmt-check`, `go vet ./...`, `git diff HEAD --check` clean; `go test ./internal/providerhealth/ -count=1` green. darwin/arm64, non-`/tmp` checkout.
- **Passes executed:** 0–4, 6, 7, 9, 11, 12, scoped to the delta plus a re-run of the credential-handling lenses.

## Blocking

1. **`TestRunDoctorConnectivityUsesOAuthLogin` is not hermetic — it reads the developer's real user config.** — `internal/cli/observability_test.go:115-160`

   The test overrides `resolveConfig` but leaves `userConfigPath` unset, so `runWithDeps` falls back to `config.DefaultUserConfigPath` (`internal/cli/app.go:473`, `:581`). Doctor then validates whatever is in the contributor's `~/.config/zero/config.json`, and any validation issue there makes the whole run exit 3 and the test fail — regardless of anything this PR does.

   On my machine that file contains a `"profiles"` key, so the test fails at every SHA of this PR (`10596d1`, `26cc8ed`, `fb2dc8ae`). It presumably passes in CI and on your machine because the home config is clean or absent, which is also why the PR's `go test ./...` checklist item was reported green in good faith.

   **`appDeps` already exposes the seam, and this package already uses it** — `internal/cli/app_test.go` sets `userConfigPath` at lines 142, 291, 352, 431 and 506. The fix is one field:

   ```go
   userConfigPath: func() (string, error) {
       return filepath.Join(t.TempDir(), "config.json"), nil
   },
   ```

   To be fair about scope: `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` have the identical flaw and fail at base `097c265`, so they are pre-existing and **not yours to fix**. I am asking only that the new test not inherit it. If you would rather fix all three in one go that is welcome, but it is not a condition.

   I would normally treat a test-isolation issue as non-blocking. It is blocking here because it is a new test in a PR whose verification claim rests on the suite passing, and because it means neither of us can currently run `internal/cli` clean to check anything else in this change.

## Non-blocking

- **`len(fields) == 2` will silently skip any future multi-part credential.** — `providerhealth.go:810`. A value like `AWS4-HMAC-SHA256 Credential=…, Signature=…` has more than two fields, so the bare-secret extraction is skipped and only the composite is registered — the same gap this commit just closed for Bearer. Nothing produces such a value today. Taking the last field, or everything after the first space, would be strictly wider at no cost.
- **The PR title lost its scope.** It is now `fix: use OAuth credentials in provider health checks`, where the commits use `fix(providerhealth):`. Conventional either way, so the changelog entry is safe — but with two-plus commits the squash subject comes from the PR title, so the released entry will be unscoped while every neighbour in `CHANGELOG.md` carries one.
- Carried forward, all unchanged: the discarded OAuth lookup error (`internal/cli/app.go:589`), the `options.Profile` vs `profile` readability trap (`providerhealth.go:308`), the redundant explicit `Close` (`:314`), and the belt-and-braces `!oauthConfigured` (`:221`).

## Still worth its own issue, not this PR

The deeper cause of the leak class is that `config.ProviderProfile` doubles as both "request auth config" and "set of strings to scrub from output," so mutating it for one purpose silently changes the other. This commit is the second patch to that seam in two days — first the cleared `APIKey`, now the composite token. A redaction secret set that accumulates independently of the profile would make the class unrepresentable, and would have covered both without either patch.

## Not verified

- The full `go test ./...`, and `internal/cli` in particular, for the reason in the blocker.
- Any platform other than darwin/arm64.
- A real xAI OAuth profile end to end; my verification is the package harness plus the doctor output above.

anandh8x:
@gnanam1990 isolated the new OAuth doctor test from the developer user config with a temporary `userConfigPath`. The focused test and full validation suite pass. Could you re-review and approve if everything looks good?

gnanam1990 (changes requested):
## Verdict

Changes requested, and **the reason it is still open is my fault, not yours.** The fix I asked for in my last review does not work — `runDoctor` never consults `deps.userConfigPath`, so the four lines you added are inert. You applied my instruction faithfully and it was wrong. Below is the seam that actually works, which this time I ran before writing it down.

Re-reviewed at `43b3f4ea`, base `097c2652` (merge-base; `origin/main` has since moved to `5886354`, leaving this branch 3 behind). Previously reviewed at `fb2dc8ae`.

## My error, precisely

I pointed at `appDeps.userConfigPath` because it exists (`internal/cli/app.go:54`) and because `internal/cli/app_test.go` sets it in five places. I never checked that `runDoctor` reads it. It does not:

```go
// internal/cli/observability.go:41-45
var userConfig, projectConfig string
if resolveOptions, optErr := config.DefaultResolveOptions(workspaceRoot); optErr == nil {
    userConfig = resolveOptions.UserConfigPath
    projectConfig = resolveOptions.ProjectConfigPath
}
```

`config.DefaultResolveOptions` is called directly, bypassing `deps` entirely. The proof is in the test's own output at your new head — `config.files` still reports the real path:

```
[pass] config.files - Zero config file inputs are available for inspection.
  userConfigPath: /Users/kratos/.config/zero/config.json
[fail] config.validation - Zero config validation found 1 issue(s).
  /Users/kratos/.config/zero/config.json: issues: [unknown config field "profiles"]
```

So `TestRunDoctorConnectivityUsesOAuthLogin` still fails at `43b3f4ea` for the same reason it failed at `fb2dc8ae`.

## Blocking

1. **`TestRunDoctorConnectivityUsesOAuthLogin` still reads the developer's real user config.** — `internal/cli/observability_test.go:115`

   The path runs `config.DefaultResolveOptions` → `DefaultUserConfigPath` → `UserConfigDir`, and `UserConfigDir` honours `XDG_CONFIG_HOME` on every platform including macOS (`internal/config/paths.go:49-60`, whose comment calls that out explicitly). So the isolation has to happen through the environment, not through `appDeps`:

   ```go
   t.Setenv("XDG_CONFIG_HOME", t.TempDir())
   ```

   **Verified, not suggested.** I added exactly that line to your test and ran it:

   ```
   ok  github.com/Gitlawb/zero/internal/cli  0.742s
   ```

   The `userConfigPath` entry in `appDeps` can then come out of the test — it has no effect on this code path and will read to the next person as if it were doing something.

   For information only, and still not something I am asking you to fix: adding the same line to `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` makes all three pass together. Those two fail at base and are pre-existing.

- **The OAuth feature works.** Same as last time, from the failing test's own report: `[pass] provider.config — credentialConfigured: oauth login` and `[pass] provider.connectivity — Provider endpoint reachable (200)`. The only failing check is `config.validation`, reading my home directory. Nothing about the OAuth path is in question.
- **The bare-token redaction from `fb2dc8ae` still holds** — `go test ./internal/providerhealth/ -count=1` green.
- **The branch is 3 commits behind the new `origin/main` (`5886354`) and merges cleanly.** I merged it locally: `go build ./...` succeeds, `internal/providerhealth` passes, `make fmt-check` and `go vet ./...` clean on the merged tree.
- **Passes executed:** 0, 1, 2, 3, 7, 11, 12 — scoped to a four-line test-only delta against production code I have already reviewed in full three times.

## Non-blocking, carried forward unchanged

- `len(fields) == 2` skips any future multi-part credential (`providerhealth.go:810`); nothing produces one today.
- The PR title lost its scope — `fix:` rather than `fix(providerhealth):` as the commits use. Conventional either way, so the changelog entry is safe, but the released entry will be unscoped while its neighbours carry one.
- The discarded OAuth lookup error (`internal/cli/app.go:589`), the `options.Profile` vs `profile` readability trap (`providerhealth.go:308`), the redundant explicit `Close` (`:314`), and the belt-and-braces `!oauthConfigured` (`:221`).

## Still worth its own issue, not this PR

Two separate things this PR has now surfaced:

1. `config.ProviderProfile` doubles as request-auth config and redaction secret set, so mutating it for one silently changes the other. Two patches to that seam in two days.
2. `runDoctor` resolves config paths directly instead of through `appDeps`, which is why doctor tests cannot be isolated the way every other CLI test in this package is. Routing it through `deps.userConfigPath` would have made my original advice correct, and would make all three of these tests hermetic without an environment variable.

## Not verified

- The full `go test ./...` and `internal/cli` in particular, for the reason in the blocker.
- Any platform other than darwin/arm64. Note the `XDG_CONFIG_HOME` fix is cross-platform by construction, since `UserConfigDir` checks it before falling back on every OS.

Sorry for the wasted round trip.

Vasanthdev2004 (changes requested):
Re-reviewed at 43b3f4ea — my approval was dismissed by the push, so this supersedes it.

The redaction work from fb2dc8ae is still good; I re-ran my probe and it holds. But **the hermeticity fix doesn't actually work**, and I don't think that's obvious from reading it, so here's the experiment.

I planted a bad config where the real one lives (`$APPDATA/zero/config.json` on Windows, same role as gnanam's `~/.config/zero/config.json`):

```json
{"profiles":{"junk":1}}
```

Then ran the test at this head. It still fails, and the doctor report says why:

```
[pass] config.files - userConfigPath: C:\...\tmp.l7l5evzrUJ\zero\config.json
[fail] config.validation - Zero config validation found 1 issue(s).
  C:\...\tmp.l7l5evzrUJ\zero\config.json: issues: [unknown config field "profiles"]
```

That's my planted file, not the `t.TempDir()` path the test now injects. Identical failure at fb2dc8ae, so the new commit changed nothing observable for this failure mode.

The reason is that `deps.userConfigPath` isn't the seam doctor reads. `runDoctor` resolves its own path at `internal/cli/observability.go:43-44`:

```go
if resolveOptions, optErr := config.DefaultResolveOptions(workspaceRoot); optErr == nil {
    userConfig = resolveOptions.UserConfigPath
```

`DefaultResolveOptions` goes to the real user config directory and never consults `appDeps`. So the override is genuinely inert on this path — it's not a case of it working on macOS but not Windows.

What does work is the helper this package already has. Adding one line:

```go
func TestRunDoctorConnectivityUsesOAuthLogin(t *testing.T) {
	setCLIUserConfigRoot(t)
```

makes the test pass with the bad config still planted. It's the right tool because it moves the config *directory* per-GOOS (`APPDATA` / `HOME` / `XDG_CONFIG_HOME`), which is the thing `DefaultResolveOptions` is reading.

Worth saying plainly: this is not carelessness. `TestRunDoctorConnectivityProbesProvider` — pre-existing on 097c2652, not yours — has exactly the same hole; I planted the same config and it fails too. You copied the neighbour, and the neighbour was already wrong. Fixing that one as well would be welcome but it's genuinely optional and I won't hold the PR for it.

The three non-blocking items from my last review are unchanged and I'm still not blocking on them.

anandh8x:
@gnanam1990 replaced the ineffective `userConfigPath` override with `XDG_CONFIG_HOME`, which is the path `runDoctor` actually resolves. The focused test and full validation suite pass. Could you re-review and approve if everything looks good?

gnanam1990 (approved):
## Verdict

Approve. The blocker is closed, and I confirmed the fix both works and still pins the feature it was guarding.

Re-reviewed at `12ccfc90`, merge-base `097c2652`, with `origin/main` at `58863548` (branch is 3 behind, merges cleanly — verified below). Previously reviewed at `43b3f4ea`.

## Withdrawn — the test-isolation blocker

`12ccfc90` swaps the inert `appDeps.userConfigPath` entry for `t.Setenv("XDG_CONFIG_HOME", t.TempDir())` and deletes the dead dep. That is the seam `runDoctor` actually reaches, via `config.DefaultResolveOptions` → `DefaultUserConfigPath` → `UserConfigDir` (`internal/config/paths.go:49-60`).

`TestRunDoctorConnectivityUsesOAuthLogin` now passes on a machine whose real `~/.config/zero/config.json` has a validation issue, which is where it failed at all three previous heads.

Two mutations, because a hermeticity fix can easily turn a test into one that passes for the wrong reason:

- **Remove the `XDG_CONFIG_HOME` line** → the test fails again. The isolation is load-bearing rather than incidental.
- **Delete the OAuth injection at `internal/cli/app.go:588`** (`options.OAuthResolver, _ = oauthLoginForProfile(...)`) → the test fails. So it still asserts the feature end to end; making it hermetic did not hollow it out into "doctor exits 0".

That second one was the thing worth checking, and the answer is the right one.

- `internal/cli` now fails only `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` — the same two, by name, that fail at base `097c2652`. Every failure attributable to this PR is gone. Those two remain pre-existing and correctly out of scope; the same one-line `t.Setenv` fixes them whenever someone picks that up.
- `go test ./internal/providerhealth/ -count=1` green — the `APIKey` retention and bare-token redaction from the earlier commits both still hold.
- `make fmt-check`, `go vet ./...`, `git diff HEAD --check` all clean.
- **Merged against current `origin/main` (`58863548`)**: auto-merge clean, `go build ./...` succeeds, and on the merged tree both `internal/providerhealth` and `TestRunDoctorConnectivityUsesOAuthLogin` pass.
- darwin/arm64, non-`/tmp` checkout. **Passes executed:** 0, 1, 2, 3, 7, 9, 11, 12 — scoped to a five-line test delta over production code reviewed in full across the four prior rounds.

## Non-blocking, carried forward

- **The PR title lost its scope.** `fix:` where the commits use `fix(providerhealth):`. Conventional either way so the changelog entry is safe, but with six commits the squash subject comes from the PR title, so the released entry will be unscoped while its neighbours carry one. Worth a moment before merge since it is free to fix now and permanent afterwards.
- `len(fields) == 2` skips any future multi-part credential (`providerhealth.go:810`); nothing produces one today.
- The discarded OAuth lookup error (`internal/cli/app.go:589`), the `options.Profile` vs `profile` readability trap (`providerhealth.go:308`), the redundant explicit `Close` (`:314`), and the belt-and-braces `!oauthConfigured` (`:221`).

## Worth filing separately, not for this PR

1. **`config.ProviderProfile` is overloaded** as both request-auth config and redaction secret set, so mutating it for one purpose silently changes the other. Two patches to that seam in this PR alone — the cleared `APIKey`, then the composite `Bearer` value. An accumulating secret set independent of the profile would have made both unrepresentable.
2. **`runDoctor` resolves config paths directly** rather than through `appDeps` (`internal/cli/observability.go:41-45`), which is why doctor tests cannot be isolated the way every other test in this package is, and why an `appDeps` override looks correct but does nothing. Routing it through `deps.userConfigPath` would fix the class and let all three doctor tests drop the environment variable.

## Not verified

- The full `go test ./...`; I scoped to `internal/cli`, `internal/providerhealth`, the merged-tree build, and the checks above.
- Any platform other than darwin/arm64. The `XDG_CONFIG_HOME` fix is cross-platform by construction, since `UserConfigDir` consults it before the per-OS fallback.
- A real xAI OAuth profile end to end; my verification is the package harness plus the doctor report.

Thanks for the patience across five rounds, three of which were my errors rather than yours.

Vasanthdev2004 (approved):
Approving at 12ccfc90. Clearing my block — the security fix is solid and I don't think what's left is worth a fourth round.

But I want to be precise rather than quietly waving it through, because **the isolation is fixed on macOS and Linux and is still open on Windows**, and the way it's written makes it look done everywhere.

`t.Setenv("XDG_CONFIG_HOME", ...)` reaches `config.UserConfigDir`, which honours XDG only on darwin (explicit branch) and on Linux (via `os.UserConfigDir`). On Windows `os.UserConfigDir` reads `%AppData%` and ignores XDG entirely — `internal/config/paths.go:50-61`.

Same experiment as last time, planting a bad config at `$APPDATA/zero/config.json` on this head:

```
--- FAIL: TestRunDoctorConnectivityUsesOAuthLogin
      userConfigPath: C:\...\tmp.AMgryDTlO0\zero\config.json
      [fail] config.validation - Zero config validation found 1 issue(s).
```

and without one planted:

```
ok  github.com/Gitlawb/zero/internal/cli
```

That second line is what CI and a mac see, which is why this reads as fixed. `setCLIUserConfigRoot(t)` is the version that covers all three — it moves `APPDATA` / `HOME` / `XDG_CONFIG_HOME` per-GOOS, which is the whole reason it exists.

I'm not blocking on it, for three reasons: it's test-only, the production change is correct and pinned, and the pre-existing `TestRunDoctorConnectivityProbesProvider` in the same file has the identical hole on every platform — so this isn't a standard your PR is failing, it's one the file never met. Holding a real user-facing fix for a Windows-only test residual would be out of proportion.

I have a Windows box and you don't, so let me take it: I'll send a follow-up moving both doctor tests onto `setCLIUserConfigRoot` and verifying with the planted-config probe. No action needed from you.

Everything else re-verified at this head — redaction probe still passes, `internal/providerhealth` green, gofmt and vet clean. The three non-blocking items from earlier are unchanged and still fine by me.

PR #838, opened 2026-07-30
perf: improve harness context efficiency

anandh8x:
## Summary

- defer optional built-in tools behind a stable searchable catalog while keeping core read, search, edit, execution, and planning tools eager
- report the actual initially exposed, available, and deferred tool surface in `zero context`
- compact repetitive successful `go test` package output while retaining exact raw output as an artifact and surfacing that state in the TUI
- track file versions and exact seen line ranges so edits reject stale or unseen source content
- prune superseded identical read results only when compaction pressure already requires a context rewrite
- normalize provider-expanded empty or already-covered permission requests to avoid wasteful failed tool retries
- make planning and specialist guidance milestone-based and evidence-driven instead of encouraging unnecessary model/tool turns
- replace per-turn recap requests with a cancellable idle orientation note that appears after four minutes of inactivity and uses the active goal, current task, and recent conversation to suggest the next action

## Why

Zero previously exposed full schemas for optional tools on every request, retained repeated recoverable context, relied on weaker read-before-edit state, and issued an extra recap request after every successful TUI turn. This increased fixed context and could cause avoidable model retries and tool calls. The changes keep capabilities discoverable while reducing model-visible noise, strengthening edit safety, and moving recap work off the critical path.

## Measured impact

Frozen 10-task comparison using the same ChatGPT provider, GPT-5.5 model, balanced execution profile, task wording, and one iteration per task:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Correctness-checked tasks | 6/6 | 6/6 | unchanged |
| Initial fixed context | 10,964 | 8,201 tokens | -25.2% |
| Initial tool schemas | 5,977 | 3,271 tokens | -45.3% |
| Total input tokens | 838,494 | 493,714 | -41.1% |
| Uncached input tokens | 237,918 | 177,810 | -25.3% |
| Output tokens | 8,724 | 5,470 | -37.3% |
| Model requests | 79 | 47 | -40.5% |
| Tool calls | 84 | 50 | -40.5% |
| Generation span time | 249.7s | 129.7s | -48.1% |

Generated benchmark reports remain outside the repository.

### Recap measurement scope

The benchmark runs tasks back-to-back and does not wait for the four-minute idle window, so the figures above do not measure the recap redesign. No token, cost, or latency improvement is claimed for that path. Its behavior is verified separately: scheduling performs no provider work, user activity and disabling recaps cancel in-flight generation, stale results are discarded, and the TUI no longer makes an immediate duplicate recap request after a completed turn. A generated idle recap may use up to 4,800 characters of bounded goal, plan, and recent-conversation context, compared with the previous 1,600-character final-answer input.

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- focused `go test -race` coverage for file tracking, output reduction, tool partitioning, pruning, permission normalization, and recap cancellation
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- manual Terminal Control TUI check against the freshly built branch binary: a normal answer completed without the previous immediate duplicate recap request

## Summary by CodeRabbit

- **New Features**
  - Optional tools can now be discovered on demand, reducing initial context usage.
  - Added exact byte-range file reads with UTF-8-safe continuation details.
  - Added idle recaps after periods of inactivity.
  - Repetitive command output is summarized while preserving complete raw output.

- **Bug Fixes**
  - Strengthened file safety by requiring observed content before edits.
  - Improved permission normalization and output metadata preservation.

- **Improvements**
  - Reduced unnecessary context compaction.
  - Tool cards now display output-budget and saved-artifact indicators.
  - Refined tool discovery and delegation guidance.

### DIFF
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -9,6 +9,25 @@ import (
 	"github.com/Gitlawb/zero/internal/tools"
 )
 
+type normalizationRecordingShellTool struct {
+	calls []map[string]any
+}
+
+func (tool *normalizationRecordingShellTool) Name() string { return "bash" }
+func (tool *normalizationRecordingShellTool) Description() string {
+	return "record normalized shell arguments"
+}
+func (tool *normalizationRecordingShellTool) Parameters() tools.Schema {
+	return tools.Schema{Type: "object", AdditionalProperties: true}
+}
+func (tool *normalizationRecordingShellTool) Safety() tools.Safety {
+	return tools.Safety{SideEffect: tools.SideEffectShell, Permission: tools.PermissionAllow}
+}
+func (tool *normalizationRecordingShellTool) Run(_ context.Context, args map[string]any) tools.Result {
+	tool.calls = append(tool.calls, cloneArgs(args))
+	return tools.Result{Status: tools.StatusOK}
+}
+
 // A shell call that sets sandbox_permissions: with_additional_permissions but
 // omits (or malforms) additional_permissions can never succeed: the same
 // validation runs again at grant time regardless of the user's decision. It
@@ -171,6 +190,41 @@ func TestNormalizeProviderExpandedEmptyAdditionalPermissions(t *testing.T) {
 		if got := args["sandbox_permissions"]; got != string(tools.SandboxPermissionsUseDefault) {
 			t.Fatalf("mode %q normalized to %q, want use_default", mode, got)
 		}
+	}
+}
+
+func TestExecuteToolCallNormalizesProviderExpandedEmptyAdditionalPermissions(t *testing.T) {
+	tool := &normalizationRecordingShellTool{}
+	registry := tools.NewRegistry()
+	registry.Register(tool)
+
+	result, err := executeToolCall(
+		context.Background(),
+		registry,
+		ToolCall{
+			ID:   "c1",
+			Name: "bash",
+			Arguments: `{"command":"./zero --version","sandbox_permissions":"with_additional_permissions",` +
+				`"additional_permissions":{"file_system":{"deny_read":[],"entries":[],"read":[],"write":[]},"network":{"enabled":false}}}`,
+		},
+		PermissionModeAuto,
+		Options{Cwd: t.TempDir()},
+	)
+	if err != nil {
+		t.Fatalf("executeToolCall: %v", err)
+	}
+	if result.Status != tools.StatusOK {
+		t.Fatalf("provider-expanded empty permissions were rejected: %s", result.Output)
+	}
+	if len(tool.calls) != 1 {
+		t.Fatalf("shell tool calls = %d, want 1", len(tool.calls))
+	}
+	args := tool.calls[0]
+	if _, exists := args["additional_permissions"]; exists {
+		t.Fatalf("executeToolCall retained empty additional_permissions: %#v", args)
+	}
+	if got := args["sandbox_permissions"]; got != string(tools.SandboxPermissionsUseDefault) {
+		t.Fatalf("sandbox mode = %q, want use_default", got)
 	}
 }
 
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/Gitlawb/zero/internal/sandbox"
 	"github.com/Gitlawb/zero/internal/tools"
 )
 
@@ -142,5 +143,87 @@ func TestExecuteToolCallStillPromptsForValidAdditionalPermissions(t *testing.T) 
 	}
 	if promptCalls != 1 {
 		t.Fatalf("OnPermissionRequest called %d times, want exactly 1 for a valid elevation request", promptCalls)
+	}
+}
+
+func TestNormalizeProviderExpandedEmptyAdditionalPermissions(t *testing.T) {
+	for _, mode := range []string{
+		string(tools.SandboxPermissionsUseDefault),
+		string(tools.SandboxPermissionsWithAdditionalPermissions),
+	} {
+		args := map[string]any{
+			"command":             "./zero --version",
+			"sandbox_permissions": mode,
+			"additional_permissions": map[string]any{
+				"file_system": map[string]any{
+					"deny_read": []any{},
+					"entries":   []any{},
+					"read":      []any{},
+					"write":     []any{},
+				},
+				"network": map[string]any{"enabled": false},
+			},
+		}
+		normalizeNoopAdditionalPermissions(args, t.TempDir(), nil)
+		if _, exists := args["additional_permissions"]; exists {
+			t.Fatalf("mode %q retained empty additional_permissions: %#v", mode, args)
+		}
+		if got := args["sandbox_permissions"]; got != string(tools.SandboxPermissionsUseDefault) {
+			t.Fatalf("mode %q normalized to %q, want use_default", mode, got)
+		}
+	}
+}
+
+func TestNormalizeAdditionalPermissionsPreservesRealRequest(t *testing.T) {
+	args := map[string]any{
+		"sandbox_permissions": string(tools.SandboxPermissionsWithAdditionalPermissions),
+		"additional_permissions": map[string]any{
+			"network": map[string]any{"enabled": true},
+		},
+	}
+	normalizeNoopAdditionalPermissions(args, t.TempDir(), nil)
+	if _, exists := args["additional_permissions"]; !exists {
+		t.Fatal("non-empty additional_permissions was removed")
+	}
+	if got := args["sandbox_permissions"]; got != string(tools.SandboxPermissionsWithAdditionalPermissions) {
+		t.Fatalf("sandbox mode changed to %q", got)
+	}
+}
+
+func TestNormalizeAdditionalPermissionsRemovesWorkspaceReadAlreadyCoveredBySandbox(t *testing.T) {
+	root := t.TempDir()
+	engine := sandbox.NewEngine(sandbox.EngineOptions{
+		WorkspaceRoot: root,
+		Policy:        sandbox.DefaultPolicy(),
+	})
+	args := map[string]any{
+		"command":             "./zero --version",
+		"sandbox_permissions": string(tools.SandboxPermissionsUseDefault),
+		"additional_permissions": map[string]any{
+			"file_system": map[string]any{
+				"entries": []any{
+					map[string]any{
+						"access": "read",
+						"path": map[string]any{
+							"type": "path",
+							"path": ".",
+						},
+					},
+				},
+				"deny_read": []any{},
+				"read":      []any{},
+				"write":     []any{},
+			},
+			"network": map[string]any{"enabled": false},
+		},
+	}
+
+	normalizeNoopAdditionalPermissions(args, root, engine)
+
+	if _, exists := args["additional_permissions"]; exists {
+		t.Fatalf("already-covered additional_permissions was retained: %#v", args)
+	}
+	if got := args["sandbox_permissions"]; got != string(tools.SandboxPermissionsUseDefault) {
+		t.Fatalf("sandbox mode changed to %q", got)
 	}
 }
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -198,7 +198,7 @@ func TestNormalizeAdditionalPermissionsRemovesWorkspaceReadAlreadyCoveredBySandb
 	})
 	args := map[string]any{
 		"command":             "./zero --version",
-		"sandbox_permissions": string(tools.SandboxPermissionsUseDefault),
+		"sandbox_permissions": string(tools.SandboxPermissionsWithAdditionalPermissions),
 		"additional_permissions": map[string]any{
 			"file_system": map[string]any{
 				"entries": []any{
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -413,6 +413,14 @@ func (state *compactionState) maybeCompact(
 	// the bodies of old, large tool results (the model has already acted on
 	// them). If that brings us back under threshold, skip the paid summarizer
 	// entirely and preserve recent turns verbatim.
+	if pruned, reclaimed := pruneSupersededReadResults(messages); reclaimed > 0 {
+		messages = pruned
+		size = state.calibratedTokens(estimateTokens(messages) + toolTokens)
+		if size <= state.threshold {
+			state.lowWaterMark = size
+			return messages
+		}
+	}
 	if pruned, reclaimed := pruneStaleToolOutput(messages, state.preserveLast); reclaimed > 0 {
 		messages = pruned
 		size = state.calibratedTokens(estimateTokens(messages) + toolTokens)
--- a/internal/agent/compaction_test.go
+++ b/internal/agent/compaction_test.go
@@ -379,6 +379,43 @@ func TestRunNoCompactionWhenContextWindowZero(t *testing.T) {
 	}
 }
 
+func TestMaybeCompactStopsAfterSupersededReadsReachThreshold(t *testing.T) {
+	body := strings.Repeat("same source line\n", 1000)
+	messages := []zeroruntime.Message{
+		{Role: zeroruntime.MessageRoleSystem, Content: "system"},
+		toolCallWithArgs("old", "read_file", `{"path":"a.go"}`),
+		{Role: zeroruntime.MessageRoleTool, ToolCallID: "old", Content: body},
+		toolCallWithArgs("new", "read_file", `{"path":"a.go"}`),
+		{Role: zeroruntime.MessageRoleTool, ToolCallID: "new", Content: body},
+	}
+	pruned, reclaimed := pruneSupersededReadResults(messages)
+	if reclaimed <= 0 {
+		t.Fatal("test setup did not reclaim superseded read output")
+	}
+	prunedSize := estimateTokens(pruned)
+	if estimateTokens(messages) <= prunedSize {
+		t.Fatal("test setup does not cross the compaction threshold")
+	}
+
+	provider := &summarizeRecordingProvider{}
+	state := &compactionState{
+		enabled:      true,
+		threshold:    prunedSize,
+		preserveLast: 2,
+	}
+	got := state.maybeCompact(context.Background(), provider, messages, nil)
+
+	if provider.summarizeCalls != 0 {
+		t.Fatalf("cheap pruning should avoid the summarizer, got %d calls", provider.summarizeCalls)
+	}
+	if state.lowWaterMark != prunedSize {
+		t.Fatalf("lowWaterMark = %d, want %d", state.lowWaterMark, prunedSize)
+	}
+	if !strings.Contains(got[2].Content, "superseded identical") || got[4].Content != body {
+		t.Fatalf("unexpected pruned messages: %#v", got)
+	}
+}
+
 // reactiveProvider builds up history with a tool call on turn 1, then errors
 // with a context-limit message on turn 2 (once the history is large enough that
 // compaction can shrink it), then succeeds on the same-turn retry. Summary
--- a/internal/agent/deferred_loop_test.go
+++ b/internal/agent/deferred_loop_test.go
@@ -341,15 +341,15 @@ func TestPartitionToolsActiveHidesUnloadedExposesLoaded(t *testing.T) {
 	if search.Description != discovery || !strings.Contains(search.Description, "mcp") {
 		t.Fatalf("tool_search description must carry discovery source, got %q", search.Description)
 	}
-	if strings.Contains(search.Description, "mcp__srv__alpha") {
-		t.Fatalf("tool_search description must not list already-loaded tool names, got %q", search.Description)
+	if !strings.Contains(search.Description, "mcp__srv__alpha") {
+		t.Fatalf("stable tool_search catalog must retain already-loaded tool names, got %q", search.Description)
 	}
 	if !strings.Contains(search.Description, "mcp__srv__beta") {
 		t.Fatalf("tool_search description must list exact hidden tool names, got %q", search.Description)
 	}
 }
 
-func TestPartitionToolsActiveNothingHiddenEmptyReminder(t *testing.T) {
+func TestPartitionToolsActiveNothingHiddenKeepsStableCatalog(t *testing.T) {
 	registry := tools.NewRegistry()
 	registry.Register(fakeDeferredTool{name: "mcp__srv__alpha", desc: "alpha"})
 	registry.Register(fakeDeferredTool{name: "mcp__srv__beta", desc: "beta"})
@@ -368,9 +368,8 @@ func TestPartitionToolsActiveNothingHiddenEmptyReminder(t *testing.T) {
 	if !exposedNames["tool_search"] {
 		t.Fatalf("expected tool_search exposed on active path, got %#v", exposed)
 	}
-	// No hidden tools means no dynamic discovery text is needed.
-	if reminder != "" {
-		t.Fatalf("expected empty discovery text when nothing is hidden, got %q", reminder)
+	if !strings.Contains(reminder, "mcp__srv__alpha") || !strings.Contains(reminder, "mcp__srv__beta") {
+		t.Fatalf("stable catalog must retain all deferred names after loading, got %q", reminder)
 	}
 }
 
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1695,20 +1695,21 @@ func sandboxRestrictionRetryReason(result tools.Result) string {
 
 func runToolForNetworkRetry(ctx context.Context, registry *tools.Registry, name string, toolCallID string, args map[string]any, permissionMode PermissionMode, options Options, progressCallback func(streamjson.Event)) tools.Result {
 	return registry.RunWithOptions(ctx, name, cloneArgs(args), tools.RunOptions{
-		PermissionGranted: true,
-		PermissionMode:    string(permissionMode),
-		Autonomy:          options.Autonomy,
-		Sandbox:           options.Sandbox,
-		ToolCallID:        toolCallID,
-		SessionID:         options.SessionID,
-		Model:             options.Model,
-		ReasoningEffort:   options.ReasoningEffort,
-		Depth:             options.Depth,
-		Cwd:               options.Cwd,
-		FileTracker:       options.FileTracker,
-		EnabledTools:      options.EnabledTools,
-		DisabledTools:     options.DisabledTools,
-		Progress:          progressCallback,
+		PermissionGranted:          true,
+		PermissionMode:             string(permissionMode),
+		Autonomy:                   options.Autonomy,
+		Sandbox:                    options.Sandbox,
+		ToolCallID:                 toolCallID,
+		SessionID:                  options.SessionID,
+		Model:                      options.Model,
+		ReasoningEffort:            options.ReasoningEffort,
+		Depth:                      options.Depth,
+		Cwd:                        options.Cwd,
+		FileTracker:                options.FileTracker,
+		DeferFileObservationCommit: true,
+		EnabledTools:               options.EnabledTools,
+		DisabledTools:              options.DisabledTools,
+		Progress:                   progressCallback,
 	})
 }
 
@@ -1723,20 +1724,21 @@ func unsandboxedRetryArgs(args map[string]any) map[string]any {
 
 func runToolForUnsandboxedRetry(ctx context.Context, registry *tools.Registry, name string, toolCallID string, args map[string]any, permissionMode PermissionMode, options Options, progressCallback func(streamjson.Event)) tools.Result {
 	return registry.RunWithOptions(ctx, name, args, tools.RunOptions{
-		PermissionGranted: true,
-		PermissionMode:    string(permissionMode),
-		Autonomy:          options.Autonomy,
-		Sandbox:           options.Sandbox,
-		ToolCallID:        toolCallID,
-		SessionID:         options.SessionID,
-		Model:             options.Model,
-		ReasoningEffort:   options.ReasoningEffort,
-		Depth:             options.Depth,
-		Cwd:               options.Cwd,
-		FileTracker:       options.FileTracker,
-		EnabledTools:      options.EnabledTools,
-		DisabledTools:     options.DisabledTools,
-		Progress:          progressCallback,
+		PermissionGranted:          true,
+		PermissionMode:             string(permissionMode),
+		Autonomy:                   options.Autonomy,
+		Sandbox:                    options.Sandbox,
+		ToolCallID:                 toolCallID,
+		SessionID:                  options.SessionID,
+		Model:                      options.Model,
+		ReasoningEffort:            options.ReasoningEffort,
+		Depth:                      options.Depth,
+		Cwd:                        options.Cwd,
+		FileTracker:                options.FileTracker,
+		DeferFileObservationCommit: true,
+		EnabledTools:               options.EnabledTools,
+		DisabledTools:              options.DisabledTools,
+		Progress:                   progressCallback,
 	})
 }
 
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1357,7 +1357,8 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 		Cwd:               options.Cwd,
 		// Per-session file version tracker so write_file/edit_file refuse to clobber
 		// a file that changed on disk outside Zero since it was last read.
-		FileTracker: options.FileTracker,
+		FileTracker:                options.FileTracker,
+		DeferFileObservationCommit: true,
 		// Post-edit diagnostics are NOT forwarded to the tools: they used to run
 		// synchronously inside edit_file/write_file and block every mutating tool
 		// call on the language server (≥300ms debounce floor, 10s cap). The loop
@@ -1409,6 +1410,7 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 			result = registry.RebudgetAfterHook(call.Name, args, result)
 		}
 	}
+	result = registry.CommitFileObservation(result, options.FileTracker)
 	// Stamp the sandbox risk classification for this EXECUTED call so run-policy
 	// observers can see the risk of an allowed mutation. Prefer the preflight
 	// decision's classification when the sandbox evaluated the call; otherwise
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1076,6 +1076,9 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 			}, nil
 		}
 	}
+	if isShellCommandTool(call.Name) {
+		normalizeNoopAdditionalPermissions(args, options.Cwd, options.Sandbox)
+	}
 	// tool_search is the gateway to the allowlisted deferred tools, so a non-empty
 	// EnabledTools allowlist that omits it must NOT reject the call — otherwise the
 	// discovery tool exposed to the model rejects at dispatch time (an inescapable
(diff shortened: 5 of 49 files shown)


Vasanthdev2004:
Had a look at `689591d4` while it's still a draft — early scope feedback rather than a line-by-line, since that's the useful thing to say now and not after another round of polish.

**The numbers are good.** -41% input tokens and -40% model requests with correctness held at 6/6 is a real result, and I like that the comparison is frozen on provider, model, profile and wording.

**Seven independent changes in 44 files is my main concern.** Tool deferral, context reporting, test-output compaction, file version tracking, read pruning, permission normalization and planning guidance don't depend on each other, and they fail in completely different ways: deferral is a provider-cache and discoverability question, file versioning is edit correctness, permission normalization is security-adjacent, compaction is output fidelity. If a context-efficiency regression shows up three weeks from now, a bisect lands on one commit that did all seven. Each is independently reviewable and independently revertable if they go in separately.

**The recap rewrite isn't in the description at all, and it's the largest single change in the diff.** `recap.go` +118-38 and `recap_test.go` +125-65 — more than any other file. It's not a refactor either; it changes user-visible behaviour:

- per-turn summary → idle-triggered orientation note after `recapIdleDelay = 4 * time.Minute`
- system prompt goes from "summarize what the assistant just did" to "state the goal, then the next actionable step"
- context budget triples, `recapMaxAnswerChars = 1600` → `recapMaxContextChars = 4800`

That's a reasonable change on its own merits. It just isn't one of the seven bullets, and someone reading the summary would not know it happened.

**Related: I don't think the measured impact covers it.** The recap only fires after four minutes of inactivity in the TUI, and a frozen 10-task benchmark with one iteration per task run back-to-back never idles that long — so `recapIdleDelay` never elapses and the recap path isn't exercised. The direction is plausibly still good (far fewer recap calls, each 3× larger), but as it stands the headline numbers can't be evidence for it either way. Worth either measuring it separately or saying it's unmeasured.

**One thing I checked properly because it's the security-adjacent bit, and it's fine.** `normalizeNoopAdditionalPermissions` fails closed on every path — unmarshal error, normalize error, and anything that isn't `Empty()` or `CoversRequestPermissions` all return with the args untouched. And stripping can only ever reduce what a command gets, so the failure direction is a command that errors rather than one that over-reaches. The comment claims "requests that broaden access and malformed payloads are left untouched so validation stays closed" and that's accurate.

**Housekeeping:** all three CodeRabbit reviews are on older commits (`68a60f14`, `6ce0931a`, `400e2852`) and the head is `689591d4`, so the CHANGES_REQUESTED on the PR doesn't reflect what's there now.

Happy to do a proper review of any of these once they're split, or of the whole thing if you'd rather keep it together — say which and I'll go through it in depth. Not blocking anything while it's a draft.

anandh8x:
@coderabbitai review Please review the latest commit (6474b14) and update the review status if the requested changes are addressed.

anandh8x:
@coderabbitai review

Vasanthdev2004:
Built and ran this locally against `main` (`d37de921`) on Windows, `689591d4`. Good news first: **the headline numbers reproduce**, on a different workspace and provider config than your benchmark.

```
main   : usage 11.4k   tools: 28                              Tools 6.4k
#838   : usage  8.4k   tools: 14 exposed, 28 available, 15 deferred   Tools 3.4k
```

Tool schemas −46.9% (you claimed −45.3%), total fixed context −26.3% (you claimed −25.2%). Independent corroboration on both.

`go test ./...` is green apart from two failures that fail identically on clean `main` on my box (`TestBuildServeScopeKeepsLexicalPaths` needs symlink privilege, `TestAltScreenTranscriptScrollKeepsFooterFixed` is the long-path footer one). `-race` on `internal/tools`, `internal/agent`, `internal/contextreport` is clean.

**Deferral is safe.** All five deferred tools — `bash`, `lsp_navigate`, `read_minified_file`, `web_fetch`, `web_search` — appear in the 452-char catalog, so nothing becomes unreachable. And `exec_command` stays eager, so shell work doesn't need a discovery round-trip first. That split looks right to me.

**The elision guard is correct, and I checked it specifically** because it's the dangerous direction. A 2.6MB file read returns 131KB with `truncation_reason=byte_budget`, and `RecordSeenRange` is skipped, so `SeenWhole` stays false. The model is never credited with content it didn't see. `RecordSeenRange`'s "exact, non-elided" contract holds.

---

**One real bug, though, and it falls out of exactly that guard.**

A file past the read byte budget can never satisfy `SeenWhole`, so `write_file` with `overwrite: true` is permanently rejected on it:

```
full read: status=ok reason="byte_budget" seenWhole=false
write_file(overwrite): status=error
   Error writing big.go: the intended change depends on content that has not
   been read exactly in this session. Read the affected lines with read_file,
   then retry the edit.
```

There is no sequence of reads that clears it. Two things combine:

1. A full read is byte-truncated, so `RecordSeenRange` is skipped entirely — nothing is recorded at all.
2. `RecordSeenRange` only sets `whole` when a *single* call satisfies `start == 1 && end >= total`. Chunked reads accumulate `ranges` but never promote `whole`.

So the advice in the error can't work for this tool, which is the part I'd fix regardless of the rest — a model will loop on it.

Related and probably the same root: `SeenWhole` and `SeenRange` disagree about accumulated coverage. After reading 1–50 and 51–100 of a 100-line file, `SeenRange(40,60)` is **true** but `SeenWhole` is **false**. One merges adjacent ranges, the other doesn't.

**`edit_file` is fine** — it gates on `SeenRange`, and it succeeded on the same 2.6MB file after a scoped 5-line read. So the model isn't stuck, it just can't wholesale-rewrite a large file. That limits the blast radius but the misleading error still stands.

Suggestion: compute `whole` from the accumulated ranges rather than tracking it as a separate flag, which fixes both at once. If the byte-budget case is meant to be unrecoverable by design, the error should say so instead of pointing at `read_file`.

Happy to send this as a patch if it's useful — say the word. Everything above is on top of the scope and recap points from my earlier comment, which still stand.

Vasanthdev2004:
Patch is up as #842, targeting `feat/harness-efficiency-foundation` rather than `main` so it lands inside this PR. Merge it, cherry-pick it, or close it and write it your way — no attachment to the implementation.

The `SeenWhole` derivation is the substance. Two things in there I'd particularly want a second opinion on: resetting the accumulated ranges when a read reports a different line count (otherwise a file read whole at 100 lines then appended to still reports seen-whole at 200), and switching coverage to merged-sorted-ranges — that path is now hit by `SeenWhole` too, so the old line-by-line scan would have been paid on every overwrite check of a large file.

Your existing file-tracking tests pass untouched.

anandh8x:
@Vasanthdev2004 please review the latest head (`b4bea8c6`), which now includes #842. All CI checks are green. If your concerns are addressed, please approve and clear the requested-changes state.\n\n@coderabbitai review Please review the latest head and update the review status if all requested changes are addressed.

anandh8x:
@coderabbitai review\n\nPlease review the latest head (`b4bea8c6`) and update the review status if the requested changes are addressed.

anandh8x:
@coderabbitai approve

Vasanthdev2004 (changes requested):
Thanks for pulling #842 in and for the "Recap measurement scope" section — that answers both of my earlier points properly, and stating plainly that the benchmark doesn't cover the recap rewrite is the right call.

I checked the four recap guarantees you list by breaking each one in turn, and all four hold: scheduling does no provider work, cancel stops in-flight generation, stale results get discarded, and cancel invalidates the pending timer. I also went through the deferral logic looking for byte-stability and threshold-edge problems and came up empty — the catalog is stable and the allowlist dead-end guard does what it says. That part I'm happy with.

Three things I'd like sorted before this goes in.

**A file whose content is one long line can never be edited again.** This is the one that blocks. Write a 400KB single-line file — a minified bundle, or JSON that got serialised without newlines — and:

- `read_file` returns `truncation_reason=byte_budget`, `SeenWhole=false`
- `read_file` with `start_line=1, end_line=1` returns `byte_budget` again, and records no seen range
- `write_file` then refuses with *"the intended change depends on content that has not been read exactly in this session. Read the affected lines with read_file, then retry the edit."*

The recovery the error names is the one thing that can't work: line 1 *is* the whole file, so there's no smaller line range to fall back to. The file is permanently uneditable for the rest of the session. Worth noting this fires even with `overwrite: true`, where the new content doesn't depend on the old at all — that case shouldn't need a seen range in the first place. Two directions: let a byte-bounded read record a satisfied range, or exempt full overwrite from the gate.

**`apply_patch` now forces a wasted re-read.** After a successful patch, `apply_patch.go:108` calls `Forget()`, and the comment says why: *"so a subsequent edit_file/write_file re-reads instead of false-flagging the patch's own change as an external modification."* But the new gate doesn't let the next edit re-read — it hard-errors. Read a file, patch it successfully, then `edit_file` and you get the same refusal above. Recoverable, but it costs an extra full read after every patch, which is the opposite of what this PR is for. Either `apply_patch` should record what it wrote as seen, or the gate needs to treat a tool-authored change as known content.

**The idle recap almost never arms.** `maybeScheduleIdleRecap` is called in exactly one place — turn completion (`model.go:2484`) — and `Update` cancels it on any `KeyPressMsg`, `PasteMsg` or `MouseMsg` (`model.go:1163`), with nothing re-arming it until the next turn finishes. So the window only survives if you finish a turn and then touch nothing at all for four minutes. Scrolling back to read the answer you just got is enough to kill it permanently. If the intent is "recap when the user has been away", the timer should reset on input rather than be cancelled outright.

One smaller thing, not blocking: `reduceGoTestPassingPackages` drops every line starting with `ok  ` or `?   `, and `isSimpleGoTestCommand` matches plain `go test` prefixes, so `go test -cover ./...` loses every coverage percentage and every `[no test files]` line — which for that command is the entire thing the user asked for. The spill file keeps the exact output and the notice points at it, so nothing is actually lost, but it means the model has to go read a file to see the number it just asked for. Keeping lines that carry `coverage:` would cost very little.

I didn't chase these two, but they're in the same area and worth a look while you're in there: whether `SeenWhole` can pass on a file elided at an output boundary, and whether one edit clears seen ranges for the rest of the file rather than just the touched region.

anandh8x:
@Vasanthdev2004 addressed the latest review in `1c1cb942`:

- oversized single-line files now have an exact byte-chunk recovery path without weakening overwrite safety
- `apply_patch` re-baselines tool-authored changes, so follow-up edits do not require a wasted read
- user activity resets the idle-recap timer instead of permanently disarming it
- coverage and `[no test files]` results remain visible in reduced `go test` output

I also verified that boundary-elided reads cannot grant whole-file knowledge; partial post-edit knowledge remains conservative, while fully observed files stay fully observed. Could you re-review the latest commit and clear the requested-changes state if this looks good?

@coderabbitai review

gnanam1990 (changes requested):
## Verdict

Changes requested — one item, and it is small. Everything else I attacked held up, including two mechanisms added since my last pass. This is the first review I have posted on this PR; the observations below fold in two earlier scoped passes at `6ce0931a` and `b4bea8c6`.

Reviewed at `1c1cb942`, base `d37de921` (level with `origin/main`). 46 files, +1626/−179.

## Blocking

1. **The permission normalization is never exercised through the path that uses it — deleting the call site is invisible to the suite.** — `internal/agent/loop.go:1079`

   Two mutations:

   - Strip `additional_permissions` even when the request genuinely broadens access → **caught**, by `TestNormalizeAdditionalPermissionsPreservesRealRequest` and `TestExecuteToolCallRejectsAdditionalPermissionsWithoutFlagBeforePrompt`.
   - Delete the `if isShellCommandTool(call.Name) { normalizeNoopAdditionalPermissions(...) }` block entirely → **`internal/agent` still passes.** Re-confirmed at this head.

   All three normalization tests call the helper directly (`additional_permissions_validation_test.go:167`, `:184`, `:221`); none drives it through `executeToolCall`. The unit is tested, the wiring is not — and the description's claim is about end-to-end behaviour: "normalize provider-expanded empty or already-covered permission requests to avoid wasteful failed tool retries."

   This is the class #796 was sent back for — "deleting either `attachLifecycleHooks` call would still pass the suite." One `executeToolCall` test with a shell call carrying `additional_permissions: {}` and `sandbox_permissions: with_additional_permissions`, asserting it is not rejected, closes it.

## Withdrawn

- **The `write_file` lockout on files above the 128 KiB read budget is fixed.** `SeenWhole` consulted only the `whole` flag, which a byte-truncated read could never set, so `write_file` was permanently refused on `internal/tui/model.go`, `internal/agent/loop.go` and `internal/agent/loop_test.go` — two of which this PR edits. `a1a9503b` (via #842) derives coverage from the accumulated ranges, and goes further than I suggested: it resets the observation when the line count changes, and `coversFully` merges ranges instead of walking every line against every range, which also closes the unbounded-growth point I raised as non-blocking. Load-bearing — reverting `SeenWhole` fails `TestLargeFileBecomesWritableAfterScopedReadsCoverIt`.
- **The undocumented recap change is now documented.** `internal/tui/recap.go` (+203 across two commits) moves recaps from post-turn to a cancellable four-minute idle window with a new prompt and context builder. The description now carries an eighth bullet describing exactly that, and the "Why" paragraph explains it. My previous objection was that a reviewer working from the description would not know recaps had changed; that is no longer true.

## What I verified

- **The new byte-span tracking is sound and tested.** `1c1cb942` adds a second coordinate system — half-open byte intervals alongside line ranges — so an oversized single-line file becomes recoverable via `byte_offset`/`byte_limit` without crediting the model for the omitted middle of a truncated line read. Making `SeenWhole` ignore byte coverage fails `TestSingleLongLineBecomesWritableAfterExactByteReads` and `TestExactByteReadsAdvanceAlongUTF8Boundaries`. The UTF-8 boundary case is the one I would have asked for.
- **`apply_patch`'s whole-seen propagation is tested.** Replacing `wasWhole` with `false` fails `TestApplyPatchPreservesWholeFileObservationForFollowUpEdit`. The reasoning in the comment holds: `git apply` rejects a patch whose context drifted, so a complete prior observation plus an exactly-applied patch does determine the output.
- **Both coordinate systems invalidate correctly.** `RecordSeenRange` and `RecordSeenBytes` each reset the entire observation when their respective total changes, and `RecordHash` drops `seen` on any hash change — so ranges never outlive the version they describe. I tried to construct a bypass by deleting seen lines so unseen content shifts into an already-seen window; correctly refused.
- **The permission normalization is fail-safe by construction**, which is worth recording because it reads as alarming and is not. A wrong "no-op" verdict deletes `additional_permissions` and downgrades `sandbox_permissions` to `use_default` — strictly *narrower* access, so the failure mode is a denied command, never an escalation. It fails closed on every error path, and with `engine == nil` normalizes only genuinely empty profiles. `CoversRequestPermissions`, `NormalizeRequestPermissionProfile` and `RequestPermissionGrantProfile` all pre-exist on `main`: this PR calls the permission-decision logic, it does not change it.
- **The output reducer cannot hide a failure.** It drops only lines whose trimmed form starts with `ok  ` or `?   `; `FAIL`, `--- FAIL:`, panics and test output are kept. It fails closed — if `spillTruncatedOutput` cannot save the exact output, the result is returned untouched — so output is never discarded without an artifact. Disabling it fails `TestReduceCommandOutputCompactsPassingGoPackagesAndKeepsRawArtifact`, and it is wired through `registry.go:138`.
- **The deferred tool catalog is well covered.** Emptying `discovery` fails six tests across `internal/agent` and `internal/cli`. Keeping loaded names in a byte-stable catalog is a sensible trade: a few compact lines against not invalidating the provider cache at the loader definition on every load.
- **Gauntlet**, darwin/arm64, non-`/tmp` checkout: `go build ./...`, `make fmt-check`, `go vet ./...`, `git diff HEAD --check` clean; `internal/tools`, `internal/agent`, `internal/tui`, `internal/contextreport` all pass. Level with `origin/main`.
- **Passes executed:** 0–7, 9–12. Pass 8 was light — nothing here is platform-divergent, and the byte-budget behaviour behind the earlier blocker is platform-independent.

## Non-blocking

- **The headline measurements are now stale, and in the one direction that matters.** `1c1cb942` adds `byte_offset` and `byte_limit` to `read_file`, which is eager (no `Deferred()` method, so `IsDeferred` is false) — meaning both properties and their descriptions sit in the fixed context on **every** request. The "-45.3% initial tool schemas" and "-25.2% initial fixed context" figures were measured at `68a60f14`, before those args existed. The change is still almost certainly a large net win, but since the entire case for this PR rests on those numbers, they are worth re-running at the final head before merge.
- **The reducer can drop a test's own log line.** — `shell_output_reducer.go:70-75`. The prefix check runs on `strings.TrimSpace(line)`, so an indented `t.Logf("ok  …")` or `t.Logf("?   …")` matches and is omitted from what the model sees. Recoverable — the exact output is in the spill artifact and the notice names it — but matching the untrimmed line would avoid it, since Go's own markers are never indented.
- **Scope: eight independent changes, no linked issue.** Deferred catalog, context report, output reducer, read-before-edit tracking, prune-on-pressure, permission normalization, prompt guidance, idle recaps. They are now all disclosed, which was my main objection, so I am not blocking on it. But the cost is already visible in this PR's own history: the `SeenWhole` lockout needed a separate branch and a merge to fix because it was buried in a 40-file change, and a regression in any one of the eight reverts all of them — including the context wins, which are the part most worth keeping.
- **The title is not a conventional commit.** "Improve harness context efficiency" has no type prefix, so release-please records nothing; with ten commits the squash subject comes from the PR title. Same trap as #826, and on a PR this size, silently omitting it from the changelog is a larger loss than usual.

## Not verified

- The prune-on-pressure change (`internal/agent/prune.go`) and the context-report change, beyond confirming their packages pass. I prioritised capability loss, permission handling, output-data loss and edit safety.
- The prompt/guidance rewrite (`system_prompt.md`), which is behavioural in a way tests cannot settle.
- The recap change beyond confirming `internal/tui` passes and that it is now disclosed.
- The full `go test ./...`; I ran the four affected packages.
- Any platform other than darwin/arm64.

The testing discipline across this PR is genuinely good — four of the five mechanisms I tried to break were pinned by tests that failed for the right reason, including two added after my last pass. The one gap is the wire-up above.

Vasanthdev2004 (changes requested):
Re-reviewed at `1c1cb942`. All four fixes do what you said, and I checked each one rather than taking the summary's word for it:

- The single-line recovery works end to end — 7 chunk reads cover a 400KB minified bundle, `SeenBytes(0,total)` goes true, and the write lands. The refusal text now names `byte_offset/byte_limit`, so the model can actually find the path. I also checked `scanReadFileStats.bytes` against `os.Stat` for LF, CRLF, no-trailing-newline, empty and single-line files — all match, so the offsets don't desync.
- `apply_patch` on the normal update path no longer forces a re-read: read → patch → `SeenWhole` stays true, and follow-up edits pass.
- The recap re-arms. I broke `resetIdleRecapAfterActivity` to return a nil command and `TestUserActivityRearmsPendingIdleRecap` failed, so that assertion is real and not passing on some other batched command. `beginRun` increments `runID` at run start, so the `runID != m.runID` guard is satisfiable in production, not just in the test.
- `go test -cover` keeps its percentages and `[no test files]` lines now.

Nice work on the byte-interval design specifically — a chunk request of 1001 bytes across a multibyte rune records `0-1000`, and two chunks with a gap between them correctly refuse an edit landing in the gap. That's the conservative direction and it's right.

Two things in this commit have to change before it can go in, though. Both let the model edit content it was never shown, which is the exact thing the gate exists to prevent.

**A rename or copy patch grants whole-file credit for a file nobody read.** `apply_patch.go:130` does `wasWhole := wholeBefore[absolute] || createdBefore[absolute]`, and `createdBefore` comes from `missingPatchTargets`, which flags any header path that didn't exist before the apply. `patchHeaderPaths` puts both sides of `diff --git a/X b/Y` in that list, so a rename or copy *destination* counts as "created". The comment above it justifies the credit as the model having supplied a complete new file — true for a `/dev/null` creation hunk where the patch body contains every line, false for a rename, where git takes the bytes from a source file the patch never quotes.

With a control first to prove the gate was live (`write_file` on an unread file → refused):

```
[rename] apply_patch ok; SeenWhole(dst)=true   (file NEVER read)
[rename] write_file on never-read dst: status=ok   disk now="BLIND OVERWRITE"
[copy]   apply_patch ok; SeenWhole(dst)=true   (file NEVER read)
[copy]   write_file on never-read dst: status=ok   disk now="BLIND OVERWRITE"
```

Same probe on `b4bea8c6` gives `SeenWhole=false` and `status=error` for both, so this is a regression inside the PR — the old `Forget()` was refusing correctly here. Renaming a file you haven't read is an ordinary thing to do, and the model writes the patch text, so this isn't an exotic path.

**Byte mode credits bytes the registry strips before the model ever sees them.** `read_file` calls `RecordSeenBytes` inside the tool and guards on `!result.Truncated`, but `Registry.RunWithOptions` applies the semantic/token budget afterwards in a deferred func. So the tool records what it *intended* to deliver and the registry decides what actually gets delivered. On multibyte content those diverge badly:

```
chunk off=0       truncated=true  credited=0-65535        delivered_bytes=287
chunk off=65535   truncated=true  credited=65535-131069   delivered_bytes=294
chunk off=131069  truncated=true  credited=131069-196604  delivered_bytes=294
chunk off=196604  truncated=true  credited=196604-252014  delivered_bytes=32961

file=252014 bytes; total delivered to model=33836; marker ever delivered=false
SeenBytes(whole)=true SeenWhole=true
edit_file on NEVER-DELIVERED marker: status=ok
```

I asserted the marker never appeared in any chunk's output before editing it. 13% of the file reached the model and the tracker called it fully seen.

Worth noting the same shape shows up on the default line path: a 102,296-byte CJK file sits under the in-tool byte budget so the tool's own guard doesn't fire, the registry then trims it to 40,252 bytes with `truncation_reason=semantic_file_budget`, and `SeenWhole` is still true. I tried twice to land an edit on a provably elided line there and couldn't — my marker kept getting delivered — so I'm not claiming that one as exploitable, only that the same over-credit is visible. Fixing where the recording happens would cover it either way.

The root cause for both is placement: the credit is recorded before the last thing that decides what the model receives. Recording after the registry budget has run, or having the registry drop credit for anything it trims, would close both. For the rename case, `createdBefore` needs to mean "the patch body contained every line of this file" rather than "this path didn't exist before".

Two unrelated notes: `TestAltScreenTranscriptScrollKeepsFooterFixed` and `TestExecCommandForegroundServerReturnsSessionAndServesHTTP` both fail for me on `origin/main` too, so they're not yours — looks like a local Windows thing on my side.

Vasanthdev2004:
@anandh8x flagging the rename/copy one from my review as the priority — it's the only finding that's a regression against your own previous head, so it's cheap to confirm.

`apply_patch.go:130`:

```go
wasWhole := wholeBefore[absolute] || createdBefore[absolute]
```

`createdBefore` comes from `missingPatchTargets`, and `patchHeaderPaths` feeds it *both* sides of `diff --git a/X b/Y`. So the destination of a rename or copy is "missing before the apply" and gets treated as a file the model supplied in full. It didn't — git took those bytes from the source, which the patch never quotes.

Smallest repro, no `read_file` anywhere in the session:

```
diff --git a/secret.txt b/exposed.txt
similarity index 100%
rename from secret.txt
rename to exposed.txt
```

On `1c1cb942`: `SeenWhole(exposed.txt)=true`, and `write_file` with `overwrite: true` then rewrites it — a file whose contents were never shown to the model. On `b4bea8c6` the same sequence gives `SeenWhole=false` and the write is refused. Copy patches behave the same way. Worth running a `write_file` control on an unread file first so you can see the gate is live in the harness before you measure.

The fix is in what `createdBefore` is allowed to mean: "the patch body contained every line of this file" (a `/dev/null` creation hunk) rather than "this path didn't exist before". Rename and copy headers should fall through to `Forget` like everything else.

The byte-mode over-credit in the same review has a different root cause — recording happens before `Registry.RunWithOptions` applies its budget — so they'll need separate fixes. No rush on my end, ping me when it's pushed and I'll re-check both.

gnanam1990 (dismissed):
## Verdict

Approve. My blocker is closed, and so are both of the soundness holes @Vasanthdev2004 found after my last review — I verified all three independently rather than reading the commit messages.

Re-reviewed at `74c06a13`, base `d37de921`. Previously reviewed at `1c1cb942`; the delta is `2ddb1340 "test: cover permission normalization wiring"` and `74c06a13 "fix: commit file observations after output budgeting"`.

## Withdrawn — my blocker

The permission normalization was reachable only through tests that called the helper directly, so deleting the call site left `internal/agent` green. `2ddb1340` adds the wiring test. Re-running the same mutation now fails `TestExecuteToolCallNormalizesProviderExpandedEmptyAdditionalPermissions`. Closed.

## Verified — both of Vasanth's findings

Credit where it is due: these were the two real defects in this PR and I did not find either. I had probed the "can the model edit content it never saw" direction on an older head and cleared it; his probes were better aimed.

**Rename and copy destinations no longer gain whole-file credit.** `apply_patch.go:132` now derives `wasWhole` from `completeCreatedPatchTargets` — only files whose full contents come from a `/dev/null` creation hunk — instead of `missingPatchTargets`, which counted any header path that did not previously exist and therefore swept in rename/copy destinations whose bytes come from an unread source. Reproduced his probe end to end, with a control first to prove the gate was live:

```
control: write_file on never-read src.txt -> error      (gate live)
apply_patch(rename)                       -> ok
   SeenWhole(dst, NEVER read)             = false       (was true)
   write_file on never-read dst           -> error
   disk = "ORIGINAL SECRET CONTENT"                     (unchanged)
```

**Byte credit is no longer granted for content the registry strips.** The recording moved out of the tool and behind `pendingFileObservation`, committed by `Registry.CommitFileObservation` only when `result.Status == StatusOK && !result.Truncated && strings.HasPrefix(result.Output, observation.output)` — i.e. only when the tool's complete output survived every later layer intact. That is the right shape: a trimming layer breaks the prefix and silently forfeits credit rather than over-claiming it.

Load-bearing, confirmed by mutation. Reducing the guard to `observation == nil || tracker == nil` fails three tests across two packages:

```
--- FAIL: TestRegistryTruncationDoesNotCreditUndeliveredByteRanges
--- FAIL: TestRegistryTruncationDoesNotCreditWholeLineRead
--- FAIL: TestExecuteToolCallDoesNotCommitReadObservationBeforeHookBudget
```

The third is the one I would have asked for: it pins the agent-side hook budget as a layer that must also run before credit is committed, not just the registry's own budget.

## Also fixed since my review

- **The title is now a conventional commit** — `perf: improve harness context efficiency`. With ten-plus commits the squash subject comes from the PR title, so this will now produce a changelog entry rather than being silently dropped.
- **`go test -cover` keeps its numbers.** `reduceGoTestPassingPackages` now excludes lines containing `coverage:` or `[no test files]`. I confirmed all 20 coverage lines survive in a synthetic `go test -cover ./...` output and the reduction correctly declines to fire. This also subsumes the narrower `t.Logf("ok  …")` case I raised.

## What I verified overall

Across three passes I have now tried to break seven mechanisms in this PR. Six were pinned by tests that failed for the right reason; the seventh (permission wiring) is fixed above.

| Mutation | Result |
|---|---|
| `SeenWhole` ignores accumulated line ranges | caught |
| `SeenWhole` ignores byte coverage | caught (incl. a UTF-8 boundary case) |
| `apply_patch` drops whole-seen propagation | caught |
| Output reducer disabled | caught |
| Deferred tool catalog emptied | caught (6 tests) |
| Observation committed despite a trimmed output | caught (3 tests) |
| Permission normalization call site deleted | caught, as of `2ddb1340` |

Gauntlet at this head, darwin/arm64, non-`/tmp` checkout: `go build ./...`, `make fmt-check`, `go vet ./...`, `git diff HEAD --check` all clean; `internal/tools`, `internal/agent`, `internal/tui`, `internal/contextreport` all pass. The two `internal/cli` doctor failures are pre-existing and environmental on my machine — they read the real `~/.config/zero/config.json` and fail identically at base.

## Non-blocking

- **The headline measurements are stale in the one direction that matters.** `read_file` has gained `byte_offset` and `byte_limit` since the benchmark was run, and it is eager (no `Deferred()` method), so both properties and their descriptions now sit in the fixed context on every request. The "-45.3% initial tool schemas" and "-25.2% initial fixed context" figures were measured at `68a60f14`. Still very likely a large net win, but since the case for this PR rests on those numbers, re-running them at the final head before merge would make the claim true again rather than approximately true.
- **Scope: eight changes, no linked issue.** All eight are now disclosed in the description, which was my substantive objection, so I am not holding the PR for it. The cost is on record though: across this PR's life, three separate defects — the `SeenWhole` lockout, the rename credit and the byte over-credit — each needed its own follow-up commit, and every one of them lived in the read-before-edit change. That piece would have been much easier to reason about, bisect and revert on its own.
- **One edit still clears the seen ranges for the whole file**, not just the touched region: `Record` → `RecordHash` drops the entire `seen` entry on any hash change, after which only the edited spans are re-recorded. That is the conservative direction and I am not asking for a change, but it is the same wasted-re-read cost that motivated the `apply_patch` work, so it is worth knowing it is still there for `edit_file`.

## Not verified

- The prune-on-pressure change and the context-report change, beyond confirming their packages pass.
- The prompt/guidance rewrite, which is behavioural in a way tests cannot settle.
- The recap rework beyond confirming `internal/tui` passes; Vasanth checked its four guarantees by breaking each in turn and I did not repeat that.
- The full `go test ./...`; I ran the five affected packages.
- Any platform other than darwin/arm64.

The testing discipline here ended up genuinely strong — every defect found in review came back with a regression test that fails for the right reason.

Vasanthdev2004 (dismissed):
Re-reviewed at `74c06a13`. Both blockers are fixed, and the way you fixed them is better than what I suggested — staging a `pendingFileObservation` and committing it at the boundary that actually decides what the model receives is the right shape, and the `HasPrefix(result.Output, observation.output)` check is a nice touch because it verifies survival rather than trusting a flag.

Re-ran every probe from my last review:

```
[rename] SeenWhole(dst)=false  write_file=error
[copy]   SeenWhole(dst)=false  write_file=error
byte-mode CJK: credit=""  SeenWhole=false  edit on never-delivered marker=error
line-mode 102KB CJK: seen_lines=""  SeenWhole=false
```

And checked you didn't over-correct, which was the real risk here:

- a genuine `/dev/null` creation patch still gets whole-file credit and the follow-up edit passes — `completeCreatedPatchTargets` draws the line where it should
- read → patch → edit still needs no re-read
- the ASCII single-line recovery still works, 7 chunks, write succeeds
- a hunk body containing literal `--- /dev/null` and `+++ b/secret.txt` lines does not fool the parser — the hunk-count skip holds, credit stays false and the write is refused

The `executeToolCall` wiring is right too: defer set at 1361, commit at 1413 after `RebudgetAfterHook`, exactly once, and every early return simply loses credit rather than granting it.

Clearing my requested-changes. Two things I'd like handled, neither blocking.

**The recovery path doesn't converge at its own default on multibyte content.** `byte_limit` defaults to 65536 and that's also the maximum, and on CJK text a 65536-byte request delivers ~290 bytes, gets correctly denied credit, and says nothing about why:

```
byte_limit=65536  truncated=true  delivered=287    credit=NONE
byte_limit=16384  truncated=false delivered=16459  credit=0-16383
byte_limit=4096   truncated=false delivered=4169   credit=0-4095
```

At 16384 a 252KB CJK file covers in 16 calls, so the path works fine — but only if the model happens to pick a smaller limit. At the default it gets a spill-path notice, and an identical retry returns identically, so it has no reason to try anything different and no signal telling it to. The refusal text names `byte_offset/byte_limit` but not "smaller". Either mention the limit that would fit when the budget denies a byte-mode read, or have the tool shrink the chunk itself — it knows the content is multibyte before it renders.

**The new defer flag has no enforcement, and two of its first callers already diverge.** CodeRabbit caught this one and it's worth acting on: `runToolForNetworkRetry` and `runToolForUnsandboxedRetry` both build a fresh `tools.RunOptions` that sets `FileTracker` but not `DeferFileObservationCommit`, so the registry commits inline — and their results then return into `executeToolCall` and pass through `RebudgetAfterHook` before reaching the model. Credit lands before that layer can trim. (`askUserFallbackResult` sets no tracker, so it's moot there.) I traced this rather than reproducing it — it needs a file-tracking tool to hit a sandbox retry, which is narrow — so treat it as unconfirmed, but the shape is the same one this commit exists to fix. The general problem is that "any caller with an extra output layer must set this flag" is an invariant nothing checks, and it was already missed twice in the same file the day it was introduced. Worth a comment on the field at minimum.

`gofmt`, `go vet` and the `internal/tools`, `internal/agent` and recap suites are all clean for me.

Vasanthdev2004 (dismissed):
Re-checked at `7d16bda4`. The retry fix is right — both helpers now pass `DeferFileObservationCommit: true`, and since `CommitFileObservation` at loop.go:1413 sits outside the `if toolFound` block, the retry branch reaches it. I mutation-tested your new assertion by stripping the flag from both helpers (leaving `executeToolCall`'s in place) and both tests failed with `retry defer flags = []bool{true, false}`, so it's load-bearing rather than incidental.

Re-ran the earlier probes on this head and nothing regressed: rename and copy still give `SeenWhole=false` with the write refused, a 64K CJK chunk still earns no credit and the edit is refused, and the ASCII single-line recovery is still 7 chunks to a successful write.

One thing to fix before you merge — `TestApplyPatchCompleteCreationCreditsSuppliedFile` fails on Windows:

```
file_safety_test.go:219: follow-up edit content = "updated content\r\n", want "updated content\n"
```

It's the test, not the code. `git apply` runs in the tool's temp workspace, which picks up `core.autocrlf=true` from Git for Windows' **system** gitconfig — my repo has `core.autocrlf=false` locally and it still reproduces, so a local override won't save it. Git for Windows installs that system default, so `windows-latest` will very likely hit the same thing. `ubuntu-latest` and `macos-latest` are already green here, which fits — they never see CRLF.

Cheapest fix is to stop asserting on exact bytes for something `git apply` is allowed to rewrite: compare with line endings normalized, or pass `-c core.autocrlf=false` on the `git apply` invocation the tool makes. Same family as the CRLF traps we've hit before on this repo.

Approving on the substance so my earlier review isn't what's holding this up — everything I asked for across the three rounds is in. @anandh8x once the Windows check is sorted, go ahead and merge.

anandh8x:
@gnanam1990 @Vasanthdev2004 GitHub automatically marked the prior reviews stale after the required Windows CI fix was pushed. The latest commit only normalizes CRLF/LF in the cross-platform test assertion; the production changes are unchanged. Please re-review the latest head when available. @coderabbitai review

anandh8x:
@coderabbitai approve

Vasanthdev2004 (approved):
Re-checked at `73e07cd3`. The line-ending fix is right, and `Smoke (windows-latest)` is green now — that test passes on my Windows box too, where it was failing before. Good that you normalized rather than loosening the assertion to a substring; it still proves the edit landed.

Confirmed the change is test-only (one line in `file_safety_test.go`, nothing in production), and re-ran the gate checks on this head:

```
[rename] SeenWhole=false write=error
[copy]   SeenWhole=false write=error
CJK 64K chunk: credit="" edit=error
/dev/null create: patch=ok edit=ok
```

So both bypasses stay closed and the legitimate creation path still gets its credit. `internal/tools` and `internal/agent` suites are clean here. The one `internal/tui` failure I see locally is `TestAltScreenTranscriptScrollKeepsFooterFixed`, which also fails on `origin/main` for me and passes in your CI — my machine, not this PR.

Re-approving; my previous one was dismissed by the push. Everything I raised across the four rounds is in, and the two follow-ups from my earlier review still stand as non-blocking: the multibyte `byte_limit` default that earns no credit with nothing telling the model to shrink it, and a comment on `DeferFileObservationCommit` noting that any caller with its own output layer has to set it.

Nice work on this one — the staged-observation design ended up cleaner than the fix I originally suggested.

gnanam1990 (approved):
## Verdict

Approve. Everything blocking across four review rounds is closed. I verified the two fixes in this delta independently, including reproducing the Windows-only failure on macOS so it is not taken on trust.

Re-reviewed at `73e07cd3`, base `d37de921` (level with `origin/main`). Previously approved at `74c06a13`; the delta is `7d16bda4 "fix: defer file observations across retries"` and `73e07cd3 "test: normalize patch line endings"`.

## Verified in this delta

**The retry helpers now defer their observation commit, and the assertion is load-bearing.** `runToolForNetworkRetry` and `runToolForUnsandboxedRetry` previously built a fresh `tools.RunOptions` carrying `FileTracker` but not `DeferFileObservationCommit`, so credit landed inline — before `RebudgetAfterHook` could trim what actually reaches the model. Flipping all three occurrences of the flag to `false` fails three tests:

```
--- FAIL: TestRunRetriesShellUnsandboxedAfterSandboxNamespaceLimitedOutput
--- FAIL: TestRunRetriesNetworkDeniedShellWithNetworkGrant
--- FAIL: TestExecuteToolCallDoesNotCommitReadObservationBeforeHookBudget
```

**The Windows CRLF failure is genuinely fixed, and I confirmed it without Windows.** @Vasanthdev2004 reported `TestApplyPatchCompleteCreationCreditsSuppliedFile` failing on `windows-latest` because Git for Windows ships `core.autocrlf=true` in its *system* config, which the tool's temp workspace inherits. I reproduced that condition on darwin by pointing `HOME` at a scratch dir with `core.autocrlf=true` set globally, then reverted the assertion to its pre-fix exact-byte form:

```
=== pre-fix assertion, autocrlf=true
--- FAIL: TestApplyPatchCompleteCreationCreditsSuppliedFile
    file_safety_test.go:219: follow-up edit content = "updated content\r\n", want "updated content\n"

=== fixed assertion, autocrlf=true
ok  github.com/Gitlawb/zero/internal/tools
```

That is his reported failure byte for byte, so the simulation is faithful and the normalization is what closes it. Worth recording the technique: this class of Windows-only test failure is reproducible anywhere by overriding the global gitconfig, which is cheaper than waiting on CI.

**Gauntlet** at this head, darwin/arm64, non-`/tmp` checkout: `go build ./...`, `make fmt-check`, `go vet ./...`, `git diff HEAD --check` all clean; `internal/tools`, `internal/agent`, `internal/tui`, `internal/contextreport` all pass. Level with `origin/main`.

## Everything blocking, across four rounds

| Finding | Raised by | Closed at |
|---|---|---|
| `write_file` locked out on files > 128 KiB | me | `a1a9503b` |
| Recap change undocumented in the description | me | description rewrite |
| Permission normalization wiring untested | me | `2ddb1340` |
| Rename/copy destination granted whole-file credit | Vasanth | `74c06a13` |
| Byte credit for registry-stripped content | Vasanth | `74c06a13` |
| Retry helpers committed credit before rebudget | Vasanth / CodeRabbit | `7d16bda4` |
| Windows CRLF test failure | Vasanth | `73e07cd3` |

Eight mutations across four rounds; every mechanism is now pinned by a test that fails for the right reason.

## Non-blocking — both carried from Vasanth's review, both still open

- **A denied byte-mode read still gives the model no reason to try a smaller chunk.** `byte_limit` defaults to 65536, which is also its maximum, and on CJK content a 65536-byte request delivers ~290 bytes, is correctly denied credit, and returns a message naming `byte_offset/byte_limit` but never "smaller" (`read_file.go:211`, `:213`; `file_tracker.go:380`). An identical retry returns identically, so nothing pushes the model toward the 16384 that would converge. Either name a fitting limit in the denial or have the tool shrink the chunk itself — it knows the content is multibyte before it renders.
- **The defer-flag invariant is documented but not enforced.** `RunOptions.DeferFileObservationCommit` carries a two-line comment (`registry.go:34-36`) and nothing more. "Any caller with an extra output layer must set this flag" was already missed twice in `loop.go` on the day it was introduced, which is the repo's recurring "two things that must agree, with nothing asserting they do" shape. The established remedy here is a drift test — an AST walk over `internal/agent` asserting that every `tools.RunOptions` literal setting `FileTracker` also sets `DeferFileObservationCommit` would make a third miss impossible rather than merely unlikely.

## Not verified

- The prune-on-pressure and context-report changes, beyond confirming their packages pass.
- The prompt/guidance rewrite, which is behavioural in a way tests cannot settle.
- The recap rework beyond `internal/tui` passing; Vasanth broke each of its four guarantees in turn and I did not repeat that.
- Actual Windows CI. My reproduction simulates the `core.autocrlf` condition, which is the mechanism he identified, but it is not a real `windows-latest` run.
- The full `go test ./...`; I ran the four affected packages.

One process note for the record, not a reason to hold this: the read-before-edit change accounted for six of the seven blocking findings above, each needing its own follow-up commit. Landing it as its own PR would have made every one of them cheaper to find, bisect and revert. Worth doing that way next time rather than unpicking this one now.

PR #842, opened 2026-07-31
fix(tools): derive SeenWhole from the recorded ranges

Vasanthdev2004:
For #838 — targets `feat/harness-efficiency-foundation`, not `main`. Merge it into your branch if you want it, or close it and fold the change in yourself; whatever's least friction.

This is the bug from my testing comment on #838. `SeenWhole` was a flag only a single whole-file read could set, while `SeenRange` judged coverage from the accumulated ranges, so the two disagreed about identical reads:

```
read 1-50, then 51-100 of a 100-line file
  SeenRange(1,100) -> true
  SeenWhole        -> false
```

`write_file` gates overwrites on `SeenWhole`. A file past the read byte budget can't be read whole in one call — the read is truncated and `RecordSeenRange` is skipped, which is *correct*, the model didn't see those lines — so nothing could ever set the flag and the overwrite was refused permanently, with advice to read the file that couldn't change the answer.

`SeenWhole` is now derived from the ranges against the file's recorded line count. Scoped reads that together cover a file make it writable, and the error text becomes actionable.

Two details worth your eye:

**Length changes reset the ranges.** The observation carries the line count from the last read; a read reporting a different one clears what was accumulated, because those ranges describe a shape the file no longer has. Without that, a file read whole at 100 lines then appended to would still report seen-whole at 200.

**Coverage merges sorted ranges** instead of walking every line against every range, which is what `SeenRange` did before. Same answer, but it no longer does a line-by-line scan over a large file — that path is now hit by `SeenWhole` too, so the old cost would have been paid on every overwrite check.

Verified:
- `TestSeenWholeIsDerivedFromAccumulatedRanges` — 8 cases: single read, adjacent, overlapping, out-of-order, gap in the middle, missing first line, missing last line, nothing read
- `TestSeenWholeAgreesWithSeenRange` — the two can't disagree
- `TestSeenWholeResetsWhenTheFileChangesLength`
- `TestLargeFileBecomesWritableAfterScopedReadsCoverIt` — end to end through the real `read_file` and `write_file` tools on a file that genuinely exceeds the byte budget, asserting the truncated read does *not* credit it and the scoped reads do

Mutation-checked both directions: reverting `SeenWhole` to the bare flag and dropping the length-change reset each fail the new tests. `internal/tools` and `internal/agent` green, gofmt and vet clean.

Your existing `file_safety_test.go` and `file_tracker` tests pass unchanged — I didn't touch them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file coverage tracking across multiple reads, including overlapping, adjacent, and out-of-order ranges.
  * Coverage now resets correctly when a file’s length changes.
  * Prevented truncated reads from being treated as complete file coverage.
  * Enabled valid file overwrites after all lines are read through scoped reads.
  * Ensured complete coverage is recognized consistently regardless of read order or grouping.

* **Tests**
  * Added comprehensive coverage for range tracking, file changes, large files, and overwrite behavior.

### DIFF
--- a/internal/tools/file_tracker.go
+++ b/internal/tools/file_tracker.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"os"
+	"sort"
 	"sync"
 	"time"
 )
@@ -56,7 +57,11 @@ type lineRange struct {
 }
 
 type fileObservation struct {
-	whole  bool
+	whole bool
+	// total is the file's line count as of the last recorded read, so coverage
+	// can be judged from the ranges alone. Without it a file read in pieces
+	// could never be known to have been seen in full.
+	total  int
 	ranges []lineRange
 }
 
@@ -102,6 +107,13 @@ func (tracker *FileTracker) RecordSeenRange(absPath string, start, end, total in
 	if end < start {
 		return
 	}
+	// A new line count means the file changed shape since the last read, so
+	// previously recorded ranges no longer describe it.
+	if observation.total != 0 && observation.total != total {
+		observation.ranges = nil
+		observation.whole = false
+	}
+	observation.total = total
 	if start == 1 && end >= total {
 		observation.whole = true
 		observation.ranges = nil
@@ -110,6 +122,41 @@ func (tracker *FileTracker) RecordSeenRange(absPath string, start, end, total in
 	}
 	observation.ranges = append(observation.ranges, lineRange{start: start, end: end})
 	tracker.seen[absPath] = observation
+}
+
+// coversFully reports whether ranges together cover every line in [start, end].
+//
+// Ranges are merged rather than scanned line by line: a caller asking about a
+// large file would otherwise walk every line against every recorded range, and
+// the answer is the same.
+func coversFully(ranges []lineRange, start int, end int) bool {
+	if start > end {
+		return true
+	}
+	if len(ranges) == 0 {
+		return false
+	}
+	sorted := make([]lineRange, len(ranges))
+	copy(sorted, ranges)
+	sort.Slice(sorted, func(left int, right int) bool {
+		if sorted[left].start != sorted[right].start {
+			return sorted[left].start < sorted[right].start
+		}
+		return sorted[left].end < sorted[right].end
+	})
+	next := start
+	for _, seen := range sorted {
+		if seen.start > next {
+			return false
+		}
+		if seen.end >= next {
+			next = seen.end + 1
+		}
+		if next > end {
+			return true
+		}
+	}
+	return next > end
 }
 
 // SeenRange reports whether every line in the requested range was returned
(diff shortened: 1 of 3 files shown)


Vasanthdev2004:
Both fixed on `77462b0b`.

The path one was a real portability bug, not a style point — worth spelling out. `resolveWorkspacePath` runs `EvalSymlinks` on the workspace root before anything is recorded, so the tracker is keyed on the resolved path. My assertion used the raw `t.TempDir()` string, which is identical on Windows and different on macOS, where the temp dir is under `/var` (a symlink to `/private/var`). The `SeenWhole` checks would have looked up a key that never existed: green here, red on macOS CI.

Which is the same trap that cost me two rounds on #808 this week, in a test I wrote *because* of it. Good catch.

QF1012 done too — `fmt.Fprintf(&builder, ...)`. `golangci-lint run --enable-only unused,ineffassign,staticcheck ./internal/tools/` reports 0 issues, and the four tests pass.

anandh8x:
@Vasanthdev2004 thanks for the fix. One small item is still present at the latest head: the large-file test uses `builder.WriteString(fmt.Sprintf(...))`. Could you change it to `fmt.Fprintf(&builder, ...)`? The current code works, but it creates an intermediate string and triggers staticcheck QF1012. Since this finding is introduced by the PR and is part of the required validation, it should be cleared before we merge it into #838.

Vasanthdev2004:
@anandh8x you're right and I was wrong to say it was done — fixed properly in `9aac8237`.

Worth owning how it happened, because the failure mode is worse than the lint: my edit silently didn't apply (escaping ate it), and I reported it fixed off the commit message rather than re-reading the file. The previous commit literally claims "Also switches the line generator to fmt.Fprintf". So the code said one thing and my message said another, and you had to catch it. This time I verified by running the linter instead of the diff:

```
golangci-lint run --enable-only unused,ineffassign,staticcheck ./internal/tools/
0 issues.
```

I also swept the rest of my diff for the same pattern. The one other hit is `read_file.go:187`, which is pre-existing, untouched by this PR, and not flagged — it writes to an `outputBudgetBuilder`, not a `strings.Builder`, so QF1012 doesn't apply. Left alone.

Two things about the base while I'm here:

**#838 moved to `6474b144` and `SeenWhole` there is still the bare flag** — `return tracker.seen[absPath].whole` — so this PR isn't redundant yet. It still merges into your new head with zero conflicts, so no rebase needed.

**The bug is still live on your branch:** a file past the read byte budget can't be read whole in one call (the read is truncated and `RecordSeenRange` is correctly skipped), and chunked reads never promote the flag, so `write_file --overwrite` is refused permanently with advice to "read the affected lines" that cannot change the answer. `edit_file` is unaffected since it gates on `SeenRange`, so the escape is to edit rather than rewrite.

All four tests green, CI green on all three platforms including macOS — which is the one that would catch the path-resolution bug you flagged earlier, so that fix is confirmed by more than my say-so this time.

Vasanthdev2004:
@anandh8x this is ready when you want it — head `9aac8237`, CI green on all three platforms, `mergeStateStatus: CLEAN`, and it merges into your current `6474b144` with zero conflicts.

Your QF1012 point is fixed and verified with the linter this time rather than by reading the diff.

One thing so the red X doesn't mislead you: the `CHANGES_REQUESTED` on this PR is CodeRabbit's, and it's pinned to `a1a9503b` — two commits before the fixes for the two things it raised. Both are addressed; the review just hasn't re-run (its latest check says "Review rate limited").

It targets `feat/harness-efficiency-foundation`, so merging it lands the fix inside #838 rather than racing it to main. Equally happy for you to cherry-pick the `SeenWhole` change and close this — the implementation isn't the point, the bug is:

a file past the read byte budget can never satisfy `SeenWhole`, so `write_file --overwrite` on it is refused permanently with advice to "read the affected lines with read_file" that cannot change the answer. `edit_file` is unaffected, so a model can still edit — it just can't wholesale-rewrite a large file, and it may loop on the error text first.

No urgency from my side. It's your PR and your call.

Issue #823, opened 2026-07-29
MCP tool results silently drop every non-text content block


Vasanthdev2004:
Our MCP client models a tool-result content block as text only:

```go
type Content struct {
	Type string `json:"type"`
	Text string `json:"text,omitempty"`
}
```

`internal/mcp/client.go:24-27`. There is no `data`, `mimeType`, `resource` or `annotations`, and `CallToolResult` (`:29-32`) has no `structuredContent`. `TextContent` (`:443-451`) keeps only blocks where `Type == "text"` and drops everything else.

So a server returning an image, audio, an embedded resource, or structured content produces `(empty MCP tool result)`. Browser and screenshot MCP servers are the common case: the tool call succeeds, the server returns a perfectly good image, and the model is told nothing came back. It then usually retries, which is the worst outcome.

Two parts, and the first is worth doing alone:

1. Stop silently dropping. If a result contains only non-text blocks, say what they were (`returned 1 image/png block, which Zero cannot yet forward`) instead of reporting empty. Prevents the retry loop even before we can carry the payload.
2. Carry image blocks through. We already have the plumbing: `zeroruntime.ImageBlock` and `Message.Images`, plus normalisation in `internal/imageinput`. This is gated on tool results having an image channel at all, which is a separate issue.

Vasanthdev2004:
Approved. Verified against current `main`: `Content` is still exactly `{Type, Text}` with no `data`, `mimeType`, `resource` or `annotations`, and `CallToolResult` has no `structuredContent`. `TextContent` keeps only `Type == "text"` blocks, so an image, audio, embedded resource, or structured result reaches the model as an empty tool result rather than as an error.

The silence is the worst part: a server behaving correctly to spec looks broken with no diagnostic.

Vasanthdev2004:
Status check while triaging, so nobody redoes the first half: **part 1 has landed.** `027a6708` ("fix(mcp): name the non-text blocks a tool result drops") added `DroppedContentSummary` and wired it into `registryTool.Run`, so an image-only result now reads as

```
[zero] this server returned 1 image/png block, which Zero cannot forward yet. Retrying cannot recover this payload.
```

rather than `(empty MCP tool result)`. `Content` also carries `MimeType` now, which is what lets the summary name the block. `TestDroppedContentSummaryNamesTheBlocks` passes on `main` at 99721c7.

[...]

PR #843, opened 2026-07-31
feat(tools): give tool results an image channel and add view_image

### LINKED ISSUE #824: Tool results have no image channel, so the model cannot see screenshots it captures
`tools.Result` (`internal/tools/types.go:92-98`) carries `Output string` and no image or content-item field, so no tool can hand the model an image. There is also no `view_image` tool: `grep -rni view_image internal/` returns nothing.

The sharp edge is our own screenshot tooling. `internal/tools/local_capture.go` implements `browser_screenshot`, `desktop_screenshot` and `browser_pdf`. They write a PNG to disk, set `Meta["artifact_path"]`, and return the sentence "Artifact captured: <path>". Nothing reads `artifact_path` afterwards. So the model asks for a screenshot, is told one exists, and cannot look at it. From the model's side the tool may as well have done nothing.

Worth being clear that Zero is not image-blind. The input pipeline is complete and reasonably mature: `zeroruntime.ImageBlock`, `Message.Images`, `internal/imageinput` for pasted and attached images, media-type normalisation, and ACP image blocks. A user can hand the model an image today. What is missing is the model-initiated leg: a tool result cannot contain one.

Two pieces:

1. An image channel on `tools.Result` that the agent loop turns into `Message.Images` on the tool-result message.
2. A `view_image` tool so the model can open a path it has been given, which also covers files the user mentions but does not attach.

Once (1) exists, the screenshot tools become useful with a one-line change each, and MCP image blocks have somewhere to go.

Vasanthdev2004:
Closes #824.

`tools.Result` carried only text, so no tool could hand the model a picture. The sharpest edge was our own capture tooling — `browser_screenshot`, `desktop_screenshot` and `browser_pdf` wrote a PNG, set `Meta["artifact_path"]`, and returned "Artifact captured: `<path>`". Nothing read that path afterwards, so the model asked for a screenshot, was told one existed, and couldn't look at it.

Zero was never image-blind on the way *in* — `zeroruntime.ImageBlock`, `Message.Images` and `internal/imageinput` are all mature. What was missing is the model-initiated leg.

## The one design decision worth reviewing

**Images ride a following user message, not the tool result.** I tried the obvious thing first and checked the providers before building on it. All three drop images on a tool-role message:

- **Anthropic** maps a tool result to a `tool_result` block whose `content` is a string
- **Gemini** maps it to a `functionResponse` with `Response: {"result": content}`
- **OpenAI** guards its image content-parts to the user role, with a comment that says exactly this: *"A non-user message that happens to carry Images keeps the plain string/nil content path (its images are simply not serialized)."*

So attaching them to the tool result would have plumbed the entire path and delivered nothing, silently. The separate user message is also the only shape that keeps one tool result per tool call, which the providers validate — I didn't want to solve this by duplicating tool results.

If anyone would rather teach the Anthropic mapper to put an image block inside `tool_result` (it does support that), that's a fine follow-up — but OpenAI still needs the user-message path, so the fallback has to exist regardless.

## The rest

**`view_image`** resolves through `resolveScopedReadPath` before `imageinput.LoadFile`. That scoping is load-bearing, not decoration: `LoadFile` joins a relative path to the workspace root and otherwise takes the path as given, so without it the tool reads anywhere on disk. There's a test for the absolute-path and `..` cases, and deleting the scoping fails it. Registered with the other scoped readers, declared read-only and thread-safe like `read_file`, and added to the specialist read-only sets — it's exactly as safe as `read_file`, which is already in each.

**Capture tools attach what they wrote.** A PDF or unreadable file yields no image and the text stands alone, so a capture is never failed over its attachment.

**Empty or disallowed images are dropped rather than sent.** A media type outside the provider allow-list would fail the whole turn; the tool's text is still useful without it.

## Verified

`TestRunDeliversToolResultImagesToTheModel` goes through the real `Run` loop and asserts the image reaches a *user*-role message and that the tool-result count is unchanged. Four mutations, all killed: dropping the emit, moving it to the tool role, not carrying images off the result, and removing `view_image`'s path scoping.

`internal/tools`, `internal/agent` and `internal/specialist` green; full suite clean apart from `TestBuildServeScopeKeepsLexicalPaths` and `TestAltScreenTranscriptScrollKeepsFooterFixed`, which fail identically on clean `main` on my Windows box.

One thing I noticed and deliberately did **not** change: a missing file surfaces the resolved absolute path in the error, because `resolveScopedReadPath` returns the raw syscall error. `read_file` does the same, so it's a pre-existing convention rather than something this adds — but it does put the user's home directory into model-visible output, and might be worth tightening repo-wide separately.

This also unblocks #823 — MCP image content blocks now have somewhere to go.

## Summary by CodeRabbit

- **New Features**
  - Added a `view_image` tool for loading supported images within the workspace, including media metadata.
  - Tool results and captured artifacts can now include images for model processing.
  - Image viewing is available through read-only tool access.

- **Bug Fixes**
  - Invalid, unsupported, oversized, or inaccessible images are safely rejected without interrupting ongoing operations.
  - Images are delivered consistently across multiple tool calls and remain available for processing.

- **Tests**
  - Added coverage for image delivery, metadata, workspace boundaries, unsupported files, and missing files.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -709,6 +709,16 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				Content:    toolResult.Output,
 				ToolCallID: toolResult.ToolCallID,
 			})
+			// Images ride a following USER message rather than the tool result
+			// above. Every provider drops images on a tool-role message —
+			// Anthropic's tool_result content is a string, Gemini's is a
+			// functionResponse, and OpenAI guards its image parts to the user role
+			// — so attaching them there would silently deliver nothing. A separate
+			// message also keeps the one-tool-result-per-tool-call pairing intact,
+			// which the providers validate.
+			if imageMessage, ok := toolResultImageMessage(toolResult); ok {
+				messages = append(messages, imageMessage)
+			}
 
 			// A tool may demand the run ABORT — a canceled/timed-out ask_user prompt
 			// returns context.Canceled rather than fabricating a headless answer. Stop
@@ -1431,6 +1441,7 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 		Output:          result.Output,
 		Truncated:       result.Truncated,
 		Meta:            result.Meta,
+		Images:          result.Images,
 		Redacted:        result.Redacted,
 		ChangedFiles:    result.ChangedFiles,
 		ChangeSummaries: result.ChangeSummaries,
@@ -3212,4 +3223,38 @@ func copyMessages(messages []Message) []Message {
 		copied[index].Images = zeroruntime.CloneImageBlocks(message.Images)
 	}
 	return copied
+}
+
+// toolResultImageMessage builds the user message that carries a tool's images
+// to the model, or reports false when the tool produced none.
+//
+// The text names the tool so the model can tell which call an image came from
+// when several ran in one turn — the images arrive detached from their tool
+// result, so nothing else associates them.
+func toolResultImageMessage(result ToolResult) (zeroruntime.Message, bool) {
+	images := make([]zeroruntime.ImageBlock, 0, len(result.Images))
+	for _, image := range result.Images {
+		if len(image.Data) == 0 {
+			continue
+		}
+		mediaType := zeroruntime.NormalizeImageMediaType(image.MediaType)
+		if mediaType == "" {
+			// Outside the provider allow-list. Dropping it beats sending bytes a
+			// provider will reject and failing the whole turn.
+			continue
+		}
+		images = append(images, zeroruntime.ImageBlock{MediaType: mediaType, Data: image.Data})
+	}
+	if len(images) == 0 {
+		return zeroruntime.Message{}, false
+	}
+	label := result.Name
+	if label == "" {
+		label = "tool"
+	}
+	return zeroruntime.Message{
+		Role:    zeroruntime.MessageRoleUser,
+		Content: "Image output from " + label + ":",
+		Images:  images,
+	}, true
 }
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -656,6 +656,13 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 		// between tool_results breaks strict provider replay) — same after-batch
 		// rationale as turnRequestedModel above.
 		var changedFilesThisBatch []string
+		// Images produced by tools this turn, held until every tool_result is
+		// recorded. They travel as user messages, and a user message between two
+		// tool_results breaks strict provider replay — Anthropic coalesces them
+		// into one user block list and requires the tool_result blocks first, so
+		// interleaving yields [tool_result, text, image, tool_result] and a 400.
+		// Same reason the self-correction feedback below is deferred.
+		var toolImageMessages []zeroruntime.Message
 		// Parallel read-ahead state: results for calls[precomputedStart:precomputedEnd]
 		// executed concurrently, consumed strictly in order below.
 		var precomputed []precomputedToolResult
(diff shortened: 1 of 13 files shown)


Vasanthdev2004:
All three fixed on `ee4994af`. The first one was a real bug and worth spelling out, because the repo already had the answer written down.

**Tool-result contiguity.** Correct, and my implementation was wrong. Images are user-role messages and I appended each one immediately after its own tool result, so two tool calls produced `tool_result, user, tool_result`. I traced it through the Anthropic mapper rather than reasoning about it — `appendUserBlocks` coalesces consecutive user content into one block list, and the interleaved sequence maps to:

```
[2] role=user blocks=[tool_result, text, image, tool_result]
```

`tool_result` blocks have to come first in that message, so that's a 400. And `loop.go` documents this exact rule three lines below my change, where the self-correction feedback is deferred *"so the assistant's tool_results stay contiguous (a user message between tool_results breaks strict provider replay)"*. I wrote a violation of a rule stated immediately below it.

Images are now collected during the loop and flushed once every `tool_result` is recorded — including the aborted placeholders on the abort, stop-reason and repeated-failure paths, so an early exit can't strand them mid-sequence either.

The reason this shipped: my test used one tool call, which cannot exhibit the bug. The new test runs two calls with the image on the first and asserts the image message doesn't fall between the first and last tool result. Reverting to the old append fails it.

**`execute` category.** Also correct and I'd missed a real consequence — a specialist declared `tools: [execute]` could not look at a screenshot it had just captured. Rather than adding it to that one list, the test asserts the property for *every* category that grants `read_file`, so the next category added can't quietly omit it.

**By-name registry assertion.** Added. A count assertion passes just as happily when one reader is swapped for another.

Four mutations, all killed: reverting to the immediate append, never flushing, dropping `view_image` from `execute`, and unregistering it from the core readers. `internal/agent`, `internal/tools`, `internal/specialist` and all four `internal/providers/*` packages green.

anandh8x (approved):
Reviewed the latest commit. The image delivery path keeps tool results contiguous, view_image uses the existing scoped read boundary, and the capture/tool registration changes are focused and covered by regression tests. Focused tests and race coverage for the new agent behavior pass, and CI is green. No blocking issues found.

gnanam1990 (approved):
## Verdict: Approve

Verified empirically on the branch (checked out, built). A tool that reads a file and hands it to the model is a data-exfil vector if it isn't scoped, so that was the focus.

**Security — the load-bearing check**
- `view_image` resolves through `resolveScopedReadPath` before `imageinput.LoadFile` (`view_image.go:61`). I **gutted the scoping** and confirmed `TestViewImageRefusesPathsOutsideTheWorkspace` goes red on *both* `../002/outside.png` (parent traversal) and an absolute path outside the workspace — without it the tool reads (and delivers) any file on disk. Real, tested, load-bearing.
- Path uses the **full alias set** (`path/file/file_path/filepath/filename/image`, `:57`), so a model can't route around the scope via an unhandled alias (invariant 8).
- Size and type limits come from `LoadFile` (too-large / unsupported-type); a disallowed media type is dropped rather than failing the whole turn. No token-blowup.
- Registered with the scoped readers, declared read-only + thread-safe, added to the specialist read-only sets — as safe as `read_file`, which is already in each.

**Design**
- Images ride a *separate user-role message*, not the tool_result. I confirmed the reasoning: OpenAI only serializes image parts on user-role messages, so attaching to the tool_result would plumb everything and deliver nothing silently. `TestRunDeliversToolResultImagesToTheModel` runs the real loop and asserts the image reaches a user message with the tool-result count unchanged. Green.

**Non-blocking**
- Your own disclosure is the right call: a missing file surfaces the resolved absolute path in the error, same as `read_file` — a pre-existing convention, correctly deferred to a repo-wide follow-up rather than bolted on here.

Thorough, secure, well-tested. Approve.

jatmn (approved):
LGTM

PR #846, opened 2026-07-31
perf: compact common test output

anandh8x:
## Summary

- compact repetitive passing/progress lines from Cargo test/build/check, pytest, Jest, and Vitest output
- recognize JavaScript test runners invoked directly or through npm, pnpm, and yarn
- preserve runner summaries, failures, exit status, and exact raw output artifacts
- accept the common trailing `2>&1` stderr merge while conservatively bypassing pipes, compound commands, file redirection, unsupported commands, and unrecognized output shapes

## Why

Zero previously applied semantic command reduction only to `go test`. Other common test and build commands could place dozens of repetitive successful lines into model context even though their decisive summaries were already present. Generic truncation can hide useful failures, so this change adds runner-aware reduction guarded by authoritative summary markers and retains the complete redacted output as an artifact.

## Measured impact

Deterministic 24-package/test-file corpus processed through the same command-output boundary used by `exec_command` and `bash`:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Raw estimated tokens | 1,504 | 1,504 | unchanged |
| Model-visible estimated tokens | 1,406 | 514 | -63% |
| Aggregate reduction from raw | 6% | 65% | +59pp |

This is a focused reducer microbenchmark, not a claim about total end-to-end session cost.

## Safety

- reduction requires a recognized simple command and an authoritative runner summary
- failed test names, assertion output, summaries, and exit status remain model-visible
- exact raw output is stored once and remains available through the artifact path
- unsupported and ambiguous shell forms pass through unchanged
- both `exec_command` and `bash` use the same registry boundary

## Verification

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- focused `go test -race ./internal/tools` coverage
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- manual Terminal Control run against a temporary 24-test Cargo project; Zero compacted the actual `cargo test 2>&1` result, displayed the exact-output artifact path, and retained the correct test summary

## Summary by CodeRabbit

* **New Features**
  * Reduced output from Go, Cargo, pytest, Vitest, and Jest test commands while preserving important results and failure details.
  * Added clearer summaries and notices identifying how command output was processed.
  * Preserved complete original output for cases requiring deeper inspection.
* **Bug Fixes**
  * Improved handling of stderr, unsupported commands, and shell-composed commands to avoid misleading reductions.
* **Tests**
  * Added broad coverage for command recognition, output preservation, summaries, and token reduction.

### DIFF
--- a/internal/tools/shell_output_reducer.go
+++ b/internal/tools/shell_output_reducer.go
@@ -150,7 +150,7 @@ func reduceCargoTestPassingLines(output string) (string, int) {
 }
 
 func reduceCargoBuildProgressLines(output string) (string, int) {
-	if !outputHasTrimmedPrefix(output, "Finished ") {
+	if !outputHasCargoFinishedSummary(output) {
 		return output, 0
 	}
 	return reduceRepeatedOutputLines(output, "successful Rust build progress lines", func(line string) bool {
@@ -160,7 +160,7 @@ func reduceCargoBuildProgressLines(output string) (string, int) {
 }
 
 func reducePytestPassingLines(output string) (string, int) {
-	if !strings.Contains(output, " passed") {
+	if !outputHasPytestSummary(output) {
 		return output, 0
 	}
 	return reduceRepeatedOutputLines(output, "passing pytest progress lines", isPassingPytestProgressLine)
@@ -182,7 +182,7 @@ func isPassingPytestProgressLine(line string) bool {
 }
 
 func reduceJavaScriptTestPassingLines(output string) (string, int) {
-	if !outputHasTrimmedPrefix(output, "Test Suites:") && !outputHasTrimmedPrefix(output, "Test Files ") {
+	if !outputHasJavaScriptTestSummary(output) {
 		return output, 0
 	}
 	return reduceRepeatedOutputLines(output, "passing JavaScript test-file lines", func(line string) bool {
@@ -191,9 +191,69 @@ func reduceJavaScriptTestPassingLines(output string) (string, int) {
 	})
 }
 
-func outputHasTrimmedPrefix(output string, prefix string) bool {
+func outputHasCargoFinishedSummary(output string) bool {
 	for _, line := range strings.Split(output, "\n") {
-		if strings.HasPrefix(strings.TrimSpace(line), prefix) {
+		trimmed := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmed, "Finished `") && strings.Contains(trimmed, "` profile ") &&
+			strings.Contains(trimmed, " target(s) in ") {
+			return true
+		}
+	}
+	return false
+}
+
+func outputHasPytestSummary(output string) bool {
+	for _, line := range strings.Split(output, "\n") {
+		summary := strings.Trim(strings.TrimSpace(line), "= ")
+		if summary == "short test summary info" || isPytestResultSummary(summary) {
+			return true
+		}
+	}
+	return false
+}
+
+func isPytestResultSummary(summary string) bool {
+	fields := strings.Fields(summary)
+	if len(fields) < 4 {
+		return false
+	}
+	first := strings.Trim(fields[0], ",;:()")
+	if _, err := strconv.Atoi(first); err != nil || !summaryHasCountedStatus(summary, "passed") {
+		return false
+	}
+	for index := 1; index+1 < len(fields); index++ {
+		if fields[index] != "in" {
+			continue
+		}
+		duration := strings.TrimSuffix(strings.Trim(fields[index+1], ",;:()"), "s")
+		if _, err := strconv.ParseFloat(duration, 64); err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+func outputHasJavaScriptTestSummary(output string) bool {
+	for _, line := range strings.Split(output, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmed, "Test Suites:") && summaryHasCountedStatus(trimmed, "passed") {
+			return true
+		}
+		if strings.HasPrefix(trimmed, "Test Files ") && summaryHasCountedStatus(trimmed, "passed") {
+			return true
+		}
+	}
+	return false
+}
+
+func summaryHasCountedStatus(summary string, status string) bool {
+	fields := strings.Fields(summary)
+	for index := 1; index < len(fields); index++ {
+		if strings.Trim(fields[index], ",;:()") != status {
+			continue
+		}
+		count := strings.Trim(fields[index-1], ",;:()")
+		if _, err := strconv.Atoi(count); err == nil {
 			return true
 		}
 	}
--- a/internal/tools/shell_output_reducer.go
+++ b/internal/tools/shell_output_reducer.go
@@ -8,6 +8,12 @@ import (
 
 const commandReducerMinPassingLines = 12
 
+type commandOutputReducer struct {
+	kind   string
+	label  string
+	reduce func(string) (string, int)
+}
+
 // reduceCommandOutput removes repetitive success-only lines from recognized
 // simple commands. Unsupported or compound shell input passes through exactly,
 // and every changed result retains the pre-reduction output as an artifact.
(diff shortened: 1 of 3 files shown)


anandh8x:
@coderabbitai Both requested test gaps are addressed in the latest commit. Please review the latest commit and approve if everything looks good.

Vasanthdev2004 (commented):
The reducer logic here is good — I tried hard to break it and couldn't. Fifteen adversarial shell forms all reject correctly (`cargo test 2>&1 | tee log`, `cargo test >log 2>&1`, `cargo test 2>&1 >log`, `cargo test 2>&1 2>&1`, `$(...)`, backticks, `;`, `&&`, embedded newline); realistic pytest / jest / vitest / cargo test / cargo check output with failures loses no failure line, no coverage table, no `[no test files]`, no exit status; CRLF is handled; and I confirmed spill recoverability end-to-end through the branch the real tools actually take (reduce, then `applySelfManagedOutputBudget` with the real `exec_command` at `max_output_tokens: 200` and the real `bash`) — `spill_path` survives the second budget, is readable, and is byte-identical to the pre-reduction output.

What I'd like fixed before this merges is the test side, because the PR body's safety claims currently rest on tests that can't fail.

**1. The compound-command pass-through test is vacuous.** `shell_output_reducer_test.go:221` — the fixture is 14 × `PASS src/example.test.ts` with no runner summary in it, so all six commands bail out on the *summary* guard inside `reduceJavaScriptTestPassingLines` and never reach the metachar guard at all. I replaced `strings.ContainsAny(command, "|;&<>\n\r`")` (`shell_output_reducer.go:73`) with `strings.ContainsAny(command, "\n\r")` — deleting pipe/compound/redirect rejection outright — and the entire reducer suite stayed green. Give the fixture a real summary line (`Test Suites: 14 passed, 14 total`, or switch it to the cargo shape with `test result: ok. ...`) so the assertion exercises the guard it's named for. For what it's worth the guard does work today: with a real `test result:` line, `cargo test --workspace | tee log.txt`, `... && echo done` and `... > log.txt` all come back unreduced.

**2. The "authoritative runner summary" markers aren't authoritative.** Each is a substring or line-prefix match anywhere in the output, and I reproduced a false positive for three of them:

- `shell_output_reducer.go:185` — `outputHasTrimmedPrefix(output, "Test Files ")`. `npm test` running a bespoke script that prints `Test Files discovered: 14` followed by 14 `✓ contract check … satisfied` lines: reduced=true, omitted=14, and the model sees only `Test Files discovered: 14 / all checks done / [zero] 14 passing JavaScript test-file lines omitted`. That's the whole payload gone. Since `npm test` runs whatever the package says, this marker is doing all the work. Vitest actually prints `Test Files  N passed (N)` — requiring ` passed (` or a digit on the same line would close it.
- `shell_output_reducer.go:153` — `outputHasTrimmedPrefix(output, "Finished ")`. A `cargo build` whose build script logs `Finished generating bindings` arms `cargo_build` on a *failed* build (`error[E0308]`, exit 101) and drops all 14 `Compiling` lines. Cargo's line is `    Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 2.41s` — gating on ` target(s) in ` or ` profile ` would fix it.
- `shell_output_reducer.go:163` — `strings.Contains(output, " passed")` anywhere. A failing pytest run where the app under test prints `loading fixtures ..........  [ 50%]` 14× reduces those away. The `=== N passed in Xs ===` / `short test summary info` line is the real marker.

None of these loses anything permanently — the spill has it — but "reduction requires an authoritative runner summary" isn't what the code does yet.

**3. Test coverage gaps, all of them.** Mutations that left the suite green:

- Recognition — 5 of the 11 accepted invocation forms are untested: `python -m pytest` / `python3 -m pytest` (`:111`), `npx jest|vitest` (`:118`), bare `jest`/`vitest` (`:116`), `npm|pnpm|yarn run test` (`:123`), `npm|pnpm|yarn exec jest|vitest` (`:126`). Only `npm test`, `pnpm test`, `yarn test`, `pytest -q`, `cargo test`, `cargo check`, `go test` are in the corpus — i.e. exactly the subset the "invoked directly or through npm, pnpm, and yarn" claim doesn't need.
- `strings.Trim(marks, ".") == ""` at `:181` → `marks != ""`: green. That's the guard that keeps a failing file's `..F.` pytest progress line visible.
- The JS prefixes at `:190` loosened from `"PASS "`/`"✓ "` to `"PASS"`/`"✓"`: green.
- `omitted < commandReducerMinPassingLines` at `:34` → `omitted < 1`: green — the 12-line floor isn't pinned.
- `result.Truncated = true` at `:46` deleted: green.
- `if spillPath == "" { return result }` at `:40` deleted: green. That one matters most — it's the "never reduce unless the artifact was written" invariant, and without it the notice reads `exact output saved to ` and the dropped lines are simply gone. Worth a test that forces the spill to fail.
- `TestReduceCommandOutputPreservesFailureEvidence` (`:181`) covers `cargo_test` only; pytest, javascript_test and go_test have no failure-preservation test.

**4. The registry test doesn't hit the real branch.** `TestRegistryAppliesCommandReducerToExecAndBashTools` (`:266`) registers `commandReducerFixtureTool` under the names `exec_command`/`bash`, but that fixture doesn't implement `selfBudgeting` while both real tools do (I checked: fixture false, exec_command true, bash true). So it runs `applyRegistryOutputBudget` + `enforceOutputCeiling`, not `applySelfManagedOutputBudget`. I verified the real branch by hand and it's fine, but the test as written doesn't prove it — the go test at `:30` already does this correctly with `NewExecCommandTool`; copying that shape for one of the new reducers would cover it.

Smaller things, take or leave:

- vitest's ` ✓ src/skips.test.ts (8 tests | 3 skipped) 12ms` gets dropped, so per-file skip attribution disappears (the aggregate survives in the `Tests` line). That's the same category as the `[no test files]` carve-out on the go path — consider keeping ✓ lines that mention skipped/todo. Jest's slow-suite timing `PASS src/a.test.ts (12.5 s)` goes the same way.
- The cargo/pytest/JS predicates match on `TrimSpace(line)` while the go one deliberately requires column 0 (`TestReduceGoTestPassingPackagesPreservesIndentedTestLogs`). Cargo's `test x ... ok` lines are at column 0, so `cargo_test` could use the strict form and stop eating an indented lookalike from a test's own stdout under `-- --nocapture`. `cargo_build` genuinely needs the trim.
- Not recognized, in case any are worth adding: `yarn jest` / `yarn vitest` (yarn's usual way to run a local binary — `yarn test` and `pnpm exec` are handled but `yarn <bin>` isn't), `npm run test:unit`, `poetry run pytest`, `uv run pytest`, `cargo nextest run`, `cargo +nightly test`, `./node_modules/.bin/jest`, `bun test`.

`go build ./...`, `go vet ./internal/tools/` and `go test ./internal/tools/` are all clean on my machine. Happy to approve once (1) and (2) are addressed — (3) and (4) I'd like too but won't hold the PR over them.

anandh8x:
@Vasanthdev2004 Addressed both blockers in 4d9a68b6: the compound-command tests now use real reducer-eligible runner output, and Cargo/pytest/Jest/Vitest reduction now requires structured runner completion summaries with false-positive regressions. Please re-review the latest commit when you can.

Vasanthdev2004 (approved):
Re-approving on `4d9a68b`. The new commit is a real fix, not polish, and it's the right direction.

The old pytest guard was `strings.Contains(output, " passed")` — anywhere in the output. This reducer *deletes* lines, so a false positive silently drops content from a command that merely happened to contain that substring. Same shape for `Finished ` on the cargo path. Requiring an actual summary line, with a number immediately before the status word via `summaryHasCountedStatus`, makes the trigger structural instead of incidental.

I thought about the other direction, since tightening a detector risks false negatives: a summary format this doesn't recognise now means no reduction rather than wrong reduction. That's the safe way to be wrong here — you lose some context saving, you don't lose the user's output.

Builds clean, `internal/tools` green on the current head.

Vasanthdev2004:
@jatmn tagging you specifically because this needs a second human and @anandh8x wrote it, so he can't be it.

Small and worth landing. The old guard for the pytest path was `strings.Contains(output, " passed")` — anywhere in the output — and this reducer *deletes* lines, so any command whose output happened to contain that substring could have content silently dropped. Same shape on the cargo path with a bare `Finished ` prefix. It now requires an actual summary line with a number immediately before the status word.

I re-reviewed on `4d9a68b` and approved. Builds clean, `internal/tools` green. The one thing I'd have you sanity-check is the direction of the risk: tightening a detector trades false positives for false negatives, so an unrecognised summary format now means no reduction rather than wrong reduction. I think that's clearly the right way round for something that removes output, but you're the one who'd argue it if it isn't.

jatmn (approved):
LGTM

PR #848, opened 2026-07-31
fix: suppress gosec G204 warning for trusted local config commands

euxaristia:
## Summary
- Provider command strings are intentionally passed through `sh -c` / `cmd /C` to support pipes and env expansions, and originate only from the user's trusted local config
- Gosec flags this as a G204 command-injection risk; removing the shell invocation would be a breaking change, so this adds a documented `#nosec G204` annotation instead of degrading functionality

## Test plan
- [ ] `gosec` no longer flags `internal/config/command.go`
- [ ] existing provider-command tests still pass

## Summary by CodeRabbit

* **Chores**
  * Added security annotations for provider command execution on supported platforms.
  * No user-visible behavior changes.

### DIFF
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -76,8 +76,10 @@ func shellCommand(command string) *exec.Cmd {
 		if strings.HasPrefix(strings.TrimSpace(command), `"`) {
 			command = "call " + command
 		}
+		/* #nosec G204 -- Provider commands are intentionally evaluated by a shell to support features like pipes and env expansions. The command string originates from the user's trusted local config. */
 		return exec.Command("cmd", "/C", command)
 	}
+	/* #nosec G204 -- Provider commands are intentionally evaluated by a shell to support features like pipes and env expansions. The command string originates from the user's trusted local config. */
 	return exec.Command("sh", "-c", command)
 }
 


Vasanthdev2004 (dismissed):
Approving. I traced the provenance claim rather than taking it at face value, because a `#nosec G204` on `sh -c` is exactly the kind of suppression that ages badly if the input source ever widens.

It holds today. `runProviderCommand` is reached only from `LoadProviderCommand`, which has one caller — `resolver.go:92`, gated on `options.ProviderCommand` — and that field is populated in exactly one place, `paths.go:32`, from the `ZERO_PROVIDER_COMMAND` environment variable. Notably it does *not* come from `ProjectConfigPath`, which was my concern: if a project-local config could set it, cloning a hostile repo would be arbitrary code execution, and "the user's trusted local config" would be false. It can't, so the comment is accurate.

Shell evaluation is also the point of the feature here — pipes and env expansion in a provider command only work if a shell sees it — so this isn't a case where the warning could be fixed by dropping to `exec.Command` with an argv.

One suggestion, non-blocking: it might be worth saying *environment variable* rather than "local config" in the comment. The load-bearing fact is that the string comes from `ZERO_PROVIDER_COMMAND` and not from any config file on disk, and naming that makes the suppression self-checking if someone later wires provider commands into the config schema.

anandh8x (changes requested):
.

anandh8x (dismissed):
Reviewed the latest commit. The shell invocation is intentional for provider command features, the command source remains user-controlled through ZERO_PROVIDER_COMMAND, and the focused tests and CI pass. No blocking issues found.

Vasanthdev2004 (approved):
Re-approving. The head is back to f9d93285, the exact commit I approved before, so this is the same review of the same content rather than a new one.

Apologies for the noise on your branch @euxaristia. What happened, in full:

I was investigating why six approved PRs including this one sit at `BLOCKED` while missing the `CodeQL` and `Analyze (actions)` checks. My theory was that those checks had never run on the older commits and that any fresh push would trigger them. I used this PR as the single-PR test before touching the other five, rebased it onto current main, and pushed.

The theory was wrong. All seven checks re-ran green on the new commit and the two missing ones still did not appear, so the rebase achieved nothing. I restored your original commit, and that restore is what dismissed the three approvals. The rebase itself had not.

So the branch content is exactly where you left it, and the only lasting cost is that anandh8x's approval needs redoing. @anandh8x sorry, could you re-approve when convenient. Nothing about the change moved.

For what it is worth the investigation did produce something: those two checks track whether a PR comes from a fork rather than whether it is stale, and #857 and #876 are fork PRs that merged recently with the same seven checks, so the missing checks are not what blocks this. I have asked kevincodex1 on #880 to read the actual branch protection rule, since that is not visible from outside.

kevincodex1 (approved):

PR #850, opened 2026-07-31
fix(tui): fill alt-screen frame with theme panel background

euxaristia:
## Summary
- Blank/cleared cells in the alt-screen frame fell back to the terminal's own default background instead of the theme's panel color, so light themes flashed white on resize or whenever ClearScreen erased the frame
- Wraps the rendered content in `zeroTheme.panel` sized to the full frame so every cell is explicitly painted, not just the cells `view.BackgroundColor` covers

## Test plan
- [ ] resize the TUI under a light theme and confirm no white flash
- [ ] confirm dark themes are unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alternate-screen rendering by applying the configured theme background consistently across the full terminal area.
  * Blank spaces and cleared screen regions now display the correct panel surface color.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -2773,6 +2773,7 @@ func (m model) View() tea.View {
 		// ClearScreen repaints to the theme color instead of white.
 		if m.width > 0 && m.height > 0 {
 			content = zeroTheme.panel.Width(m.width).Height(m.height).Render(content)
+			view.SetContent(content)
 		}
 	}
 	// Always requested, independent of the notifier: the composer cursor's
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -2763,6 +2763,17 @@ func (m model) View() tea.View {
 	// background behind in the user's scrollback after exit.
 	if m.altScreen {
 		view.BackgroundColor = zeroTheme.bgPanel
+		// Also fill the frame content with the panel surface, not just the
+		// terminal's default background. Relying on view.BackgroundColor alone
+		// leaves blank/padding cells (and anything ClearScreen erases) on the
+		// terminal's own default, which is white on a light terminal — so light
+		// themes flash a blinding white instead of their cream surface. Wrapping
+		// the content paints every cell (including blank lines) with the theme
+		// panel, so the surface is uniform regardless of the terminal default and
+		// ClearScreen repaints to the theme color instead of white.
+		if m.width > 0 && m.height > 0 {
+			content = zeroTheme.panel.Width(m.width).Height(m.height).Render(content)
+		}
 	}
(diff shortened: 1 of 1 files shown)


euxaristia:
Addressed CodeRabbit finding in `62e3f31`:

- After the theme panel re-renders content, call `view.SetContent(content)` so `tea.View` returns the themed alt-screen frame instead of the pre-panel string from `tea.NewView`.

@coderabbitai full review

Vasanthdev2004 (approved):
Approving. The diagnosis matches what I see: `view.BackgroundColor` only paints the cells the view actually covers, so blank and padding cells — and anything `ClearScreen` erases — fell through to the terminal default, which is white on a light terminal. Wrapping the content in the theme panel at full frame size paints every cell, which is the right fix.

I did check the thing that worries me about wrapping already-laid-out content, since the TUI does its own wrapping and a second pass could corrupt the layout. Rendering through `zeroTheme.panel.Width(40).Height(10)`:

```
short                lines 1->10   widest 2->40
exact width lines    lines 5->10   widest 40->40
more lines than h    lines 16->17  widest 1->40
over-wide line       lines 1->10   widest 55->40
```

The two cases I was worried about are both fine. Content taller than the frame is **not** truncated (16 lines stayed 16, not clipped to 10), so nothing gets hidden on a full screen. And lines that are already exactly the frame width are **not** re-wrapped, so the normal path is untouched.

One behaviour change worth naming, not blocking: a line wider than the frame now **wraps** where it previously ran off and was clipped by the terminal. That is arguably better, but it is a change — if anything in the TUI deliberately emits a line wider than `m.width` expecting the terminal to clip it, that content will now reflow and push everything below it down by a line. I could not find a case that does, so I am not asking you to change it; just worth knowing which behaviour you are choosing if a layout oddity shows up later on a narrow terminal.

The `m.width > 0 && m.height > 0` guard is the right call too — rendering at zero size before the first `WindowSizeMsg` would be the obvious way to break this.

anandh8x (approved):
Reviewed and tested the latest commit. The official release build succeeds, focused alt-screen tests pass, and a real interactive TUI check kept the light-theme panel background uniform across a 100x30 -> 56x16 live resize with intact layout; dark theme also remained uniform after resizing back. CI is green and I found no blocking defect.

Non-blocking follow-up: wrapping the complete frame adds measurable View() work in a synthetic 120x40 benchmark (about 0.32 ms / 65 KB / 1,011 allocs before versus 0.70 ms / 173 KB / 4,614 allocs here). The absolute render time stayed below 1 ms and no interaction lag was observed, but a cheaper full-frame fill would be worth exploring later.

kevincodex1 (approved):

PR #857, opened 2026-07-31
docs(agents): Codify recurring review blockers in AGENTS.md

### LINKED ISSUE #856: docs(agents): codify recurring review blockers in AGENTS.md
## Summary

Maintainer reviews (especially multi-round threads on security, sandbox, worktrees, and platform PRs) repeatedly restate the same classes of defect. Coding agents and contributors keep reintroducing them, which burns review budget.

## Goal

Extend `AGENTS.md` with a short, high-signal "common review blockers" section so agents fix these issues before the first review, without bloating the file.

## Scope

- Docs-only change to `AGENTS.md`
- Capture patterns that show up across many PRs: stale merge bases, path canonicalization, secret/argv leakage, incomplete redaction, fail-open ownership checks, partial setup without rollback, missing regression tests, and scope/docs mismatches
- Keep the addition compact

## Out of scope

- CONTRIBUTING.md rewrites
- CI changes
- Codifying feature-specific sandbox Windows rules

euxaristia:
## Summary

Maintainer review threads (especially multi-round ones) keep restating the same defect classes: stale merge bases that roll back mainline work, non-canonical path tests that break macOS/Windows smoke, secrets in argv or unredacted error paths, fail-open ownership checks, partial setup without safe rollback, missing regression tests, and PR claims that do not match what shipped.

This PR adds a short **Common Review Blockers** section to `AGENTS.md` so coding agents and contributors address those classes before the first review, without bloating the file (~1.9KB / 32 lines).

Closes #856

### `AGENTS.md`
- New section 4 covering:
  - Fresh base / no silent mainline rollbacks
  - Platform-truth path handling (canonical paths, `t.TempDir` pitfalls)
  - Security edges (argv, redaction of errors, fail closed, symlink/reparse, setup rollback ownership)
  - Atomic shared-state writes
  - Regression tests that match the claim (including failure and non-Linux path cases)
  - Honest scope (wired entry points, no name-only security allowlists)

## Test plan
- [x] Docs-only change; no Go code
- [x] `git diff HEAD --check`
- [ ] Skim section 4 for density and accuracy against recent multi-round reviews
- [ ] Confirm agents loading repo `AGENTS.md` pick up the new section

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for avoiding common code review blockers, including keeping pull requests current, ensuring cross-platform compatibility, protecting sensitive information, enforcing secure failure behavior, handling rollbacks safely, updating shared data atomically, testing behavioral and security changes, and accurately describing the scope of each change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,14 +83,16 @@ These classes drive multi-round reviews. Fix them before requesting review:
   checks over duplicated helpers that drift.
 - **Security edges:** Keep secrets out of argv, env dumps, and logs. Redact
   success and error paths (including stderr). Fail closed on ownership, lease,
-  and permission checks. Lexical path checks alone do not stop symlink or
-  reparse escape: resolve and re-validate containment before use. On multi-step
-  setup, roll back only what this run created; never destroy pre-existing
-  resources you did not create; never report success when cleanup or unlock
-  failed.
-- **Atomic shared state:** Use an atomic replace (or equivalent exclusive
-  write) for files concurrent readers may observe. Serialize read-modify-write
-  of lockfiles and shared stores.
+  and permission checks. Do not rely on pre-open path resolution
+  (`EvalSymlinks` then open) for containment: that is a check-to-use race.
+  Bind containment at open/use time (no-follow or handle-relative APIs, and
+  platform equivalents for reparse points). On multi-step setup, roll back
+  only what this run created; never destroy pre-existing resources you did
+  not create; never report success when cleanup or unlock failed.
+- **Atomic shared state:** Write a complete temporary file, then atomically
+  replace the destination so concurrent readers never see a partial write.
+  Exclusive create or a write lock alone is not enough for readers. Serialize
+  the full read-modify-write sequence for lockfiles and shared stores.
 - **Tests match the claim:** Every behavior or security-boundary change needs a
   regression test, including the failure path. Path-sensitive logic needs at
   least one non-Linux path case (or a hermetic fake that exercises the same
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,8 +85,10 @@ These classes drive multi-round reviews. Fix them before requesting review:
   success and error paths (including stderr). Fail closed on ownership, lease,
   and permission checks. Do not rely on pre-open path resolution
   (`EvalSymlinks` then open) for containment: that is a check-to-use race.
-  Bind containment at open/use time (no-follow or handle-relative APIs, and
-  platform equivalents for reparse points). On multi-step setup, roll back
+  Bind containment at open/use time with rooted or handle-relative,
+  traversal-resistant APIs. If a no-follow API is used, apply it to every
+  traversed component and enforce the platform's reparse-point protections;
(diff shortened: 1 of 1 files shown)


euxaristia:
Addressed both CodeRabbit findings in `f19a9a2`:

1. **Security edges:** Dropped the “resolve then use” wording. Guidance now forbids pre-open `EvalSymlinks`/containment checks alone and requires binding containment at open/use time (no-follow or handle-relative APIs, plus reparse equivalents).
2. **Atomic shared state:** Requires write-complete temp file then atomic replace. Explicitly notes exclusive create / write lock alone does not protect readers. RMW serialization kept as a separate requirement.

@coderabbitai full review

euxaristia:
Addressed remaining CodeRabbit finding in `1bfe6cb`:

- **Security edges:** Containment guidance now requires rooted/handle-relative traversal-resistant APIs, or no-follow on every traversed component with platform reparse protections. Final-component-only no-follow is called out as insufficient.

@coderabbitai full review

Vasanthdev2004 (approved):
Approving. This is docs-only, the numbering follows on correctly from §3, and the guidance is accurate — I checked it against real findings rather than just reading it.

Several of these map one-to-one onto things that were live in the review queue this week, which is a good sign the section is describing the actual failure modes rather than generic advice:

- "A stale head that rolls back mainline commits is a hard blocker, not a merge-time detail" — this was a P0 on my own #808 yesterday, where the branch was 16 commits behind and would have reverted the 0.6.0 release manifest along with 134 other files.
- "final-component-only no-follow is insufficient" — I pushed a fix for exactly this today (`dd3520a6`). `FILE_FLAG_OPEN_REPARSE_POINT` only protects the last component, so `os.MkdirAll` through a junctioned `.git` had elevated setup creating objects outside the workspace before the check ever ran.
- "roll back only what this run created; never destroy pre-existing resources you did not create" and "never report success when cleanup or unlock failed" — both were findings on my sandbox PRs this week, and both were real.
- "that is a check-to-use race" on `EvalSymlinks`-then-open is the correct diagnosis and the right remedy.

So this isn't theoretical. Thanks for writing it up — turning a rough review stretch into something that stops the next one is a genuinely useful contribution, and it will save reviewers repeating themselves.

Two small thoughts, neither blocking, take or leave:

The atomic-replace guidance is right on POSIX but worth a Windows caveat — `os.Rename` over an existing file fails there if any reader holds a handle without `FILE_SHARE_DELETE`, so "atomically replace" needs a retry or `ReplaceFile` to actually hold up cross-platform. Given the section explicitly asks for platform truth, that one's a bit self-referential.

"Do not bundle unrelated fixes" is the one I'd expect to cause the most friction in practice, since it's often only visible as unrelated in hindsight. Might be worth a note that splitting is cheap before review and expensive after.

Neither needs changing before merge.

jatmn (approved):
LGTM

@euxaristia please feel free to contribute adapted versions of these to Node and OpenClaude if you would like. This a quite a good concept to help keep things more uniform for all Gitlawb projects.

kevincodex1 (approved):

PR #864, opened 2026-08-02
fix(agent): end the additional_permissions retry loop

### LINKED ISSUE #845: exec_command repeatedly requests incorrect sandbox permissions and fails before using the required network permission
### Version / branch / commit

0.6.0

### OS and environment

Windows 11 Pro / Powershell

### Steps to reproduce

1. Open the repository in Zero on Windows using PowerShell.
2. Ask the agent to run:

```powershell
git branch --show-current
```

3. Observe that the command fails with:

```text
additional_permissions requires sandbox_permissions set to "with_additional_permissions"
```

4. Run the same command again.
5. Observe that it fails with:

```text
additional_permissions must include at least one of network or file_system
```

6. Ask the agent to create a GitHub issue for the failed command.
7. Approve the public GitHub issue creation.
8. Ask the agent to add the second error as a comment to the created issue.
9. Observe that `gh issue comment` repeatedly fails or requests additional network permissions instead of posting the comment.

### Expected behavior

* `git branch --show-current` should run successfully without requesting additional sandbox permissions.
* Read-only local Git commands should not require network or file-system write access.
* The current branch name should be returned directly.
* When posting a GitHub issue or comment, the tool should request only the required network permission.
* After the user approves the action, the GitHub command should execute without repeated permission prompts or sandbox configuration errors.

### Actual behavior

* `git branch --show-current` fails instead of returning the current branch name.
* The first attempt returns:

```text
additional_permissions requires sandbox_permissions set to "with_additional_permissions"
```

* A later attempt returns:

```text
additional_permissions must include at least one of network or file_system
```

* The agent generates invalid or unnecessary `additional_permissions` settings for a read-only local Git command.
* When adding a follow-up comment with `gh issue comment`, the command is blocked by permission errors or triggers another network-access approval prompt.
* The agent repeatedly retries with different permission configurations instead of successfully executing the commands.

Image
Image

### LINKED ISSUE #847: Sandbox tool invocation fails when additional_permissions is empty or mismatched
## Summary
A simple command execution failed because the tool call included an `additional_permissions` object without setting `sandbox_permissions` to `with_additional_permissions`, and then failed again when `additional_permissions` contained no effective network or filesystem permission.

## Reproduction
In a continued Zero session, the user asked to run:

```powershell
git branch --show-current
```

The command should be a normal read-only git operation and should not require extra sandbox permissions.

Observed errors from tool invocation:

```text
Error: additional_permissions requires sandbox_permissions set to "with_additional_permissions"
```

Then, with `sandbox_permissions` set to `with_additional_permissions` but an empty permissions payload:

```text
Error: additional_permissions must include at least one of network or file_system, for example {"network": {"enabled": true}} or {"file_system": {"write": ["/path"]}}
```

I reproduced the same validation class while attempting another read-only command (`gh --version`) when an empty `additional_permissions` object was preserved in the request.

## Expected behavior
Read-only commands like `git branch --show-current` should run with default sandbox permissions, or the tool layer should omit empty `additional_permissions` entirely.

## Actual behavior
The tool invocation is rejected before the command runs because permission metadata is internally inconsistent:

- `additional_permissions` present with `sandbox_permissions: use_default`, or
- `sandbox_permissions: with_additional_permissions` but no effective permission requested.

## Notes
This appears to be a request construction / validation issue rather than a git failure. The actual git command succeeds when invoked without the inconsistent permission metadata.

Vasanthdev2004:
## Summary

`git branch --show-current` — an ordinary read-only command — could leave the agent bouncing between two rejections instead of running, burning turns without converging.

Both errors come from `inlineAdditionalPermissionsProfile`, and each names only the fix that *adds* something. One pushes toward setting the flag, the other back toward requesting a permission the command doesn't need. Neither ever says to **omit the field**, which is the correct action here — so together they form a two-state oscillator. Every rejection now carries that remedy.

The schema half is why the object gets attached at all: neither field description said to omit them for ordinary commands, so a model that legitimately used `with_additional_permissions` once had nothing telling it to stop. Both descriptions now lead with "omit", and they're shared constants instead of being duplicated verbatim across `bash.go` and `exec_command.go`.

## This does not loosen the sandbox

Nothing previously refused is now accepted on the strength of an empty or mismatched payload, and `sandbox_permissions` is never upgraded on the model's behalf. Only the guidance changed.

The one behavioural change: an explicit `additional_permissions: null` is treated as the omitted field, since a null grants nothing and carries no flag. That flips three shapes from reject to execute — null with `use_default`, with `require_escalated`, and with no mode at all. The escalated case still routes through the normal escalation prompt and is **not** a new capability; the identical request without the null key already behaved that way. A *flagged* null still errors, so `with_additional_permissions` + null isn't silently downgraded.

## Note on versions

The reporter was on 0.6.0, where both legs of the loop were live. On current `main`, #838's `normalizeNoopAdditionalPermissions` already absorbs the empty payload, so only the first error reproduces today. Both legs were reproduced and fixed.

## Test plan

- `TestAdditionalPermissionsRejectionsConverge` replays #845's step-1/step-2 sequence and #847's two inconsistent states in one test
- `TestExecuteToolCallMissingAdditionalPermissionsErrorIsActionable`, `TestEmptyAdditionalPermissionsErrorNamesOmitRemedy` pin that the errors name the omit remedy
- `TestExecuteToolCallTreatsNullAdditionalPermissionsAsOmitted`, `TestExecuteToolCallRejectsFlaggedNullAdditionalPermissions` cover the null split
- Reverting only `internal/agent/loop.go` turns all five red on message text, and reproduces both issue strings verbatim
- `go build ./...`, `go vet`, `gofmt` clean; `internal/agent`, `internal/tools`, `internal/sandbox` green

Fixes #845
Fixes #847

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of optional sandbox permission fields, including explicit `null` values.
  * Clarified validation errors and retry guidance for unnecessary or invalid permissions.

* **Documentation**
  * Added guidance on valid permission combinations and examples.
  * Clarified when permission fields should be omitted, including for read-only commands.

* **Tests**
  * Added coverage for validation, error consistency, null handling, schema guidance, and successful retries.

### DIFF
--- a/internal/agent/additional_permissions_validation_test.go
+++ b/internal/agent/additional_permissions_validation_test.go
@@ -127,9 +127,7 @@ func TestExecuteToolCallMissingAdditionalPermissionsErrorIsActionable(t *testing
 	if !strings.Contains(result.Output, `{"network": {"enabled": true}}`) {
 		t.Fatalf("output = %q, want a concrete additional_permissions example", result.Output)
 	}
-	if !strings.Contains(result.Output, "omit sandbox_permissions") {
-		t.Fatalf("output = %q, want guidance that the flag can simply be omitted", result.Output)
-	}
+	assertNamesOmitRemedy(t, "missing additional_permissions", result.Output)
 }
 
 // A well-formed additional_permissions payload must still reach the normal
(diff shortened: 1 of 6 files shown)


Vasanthdev2004:
@jatmn could you take this one when you have a moment? It closes #845 and #847, which are the same bug — #847 was actually filed *by* the agent while it was stuck in the loop the issue describes.

The part worth your eye is the null handling, since it's the only behavioural change and it touches the escalation path. An explicit `additional_permissions: null` is now treated as the omitted field, on the reasoning that a null grants nothing and carries no flag. That flips three shapes from reject to execute: null with `use_default`, with `require_escalated`, and with no mode at all. The escalated one is the one to push on — it now reaches the normal escalation prompt, and if the user approves, runs outside the sandbox. My read is that this isn't a new capability, because the identical request *without* the null key already behaved exactly that way, so the null was only ever a spurious rejection. But that's a judgement about a sandbox boundary and I'd rather you checked it than took my word.

A flagged null still errors, deliberately — `with_additional_permissions` plus null must not be silently downgraded into a different outcome.

Everything else is guidance-only: the two rejection messages now name "omit the field" as a remedy, which neither did, and the schema descriptions lead with omit so the object stops being attached to ordinary commands in the first place. Nothing previously refused is accepted now, and `sandbox_permissions` is never upgraded on the model's behalf.

Reverting only `internal/agent/loop.go` turns all five new tests red on message text and reproduces both issue strings verbatim, so the loop is pinned in both directions.

One version note so the repro doesn't confuse you: the reporter was on 0.6.0 where both legs of the loop were live. On current main, #838's `normalizeNoopAdditionalPermissions` already absorbs the empty payload, so only the first error reproduces today. Both were reproduced and fixed.

anandh8x (approved):
Reviewed the latest commit. Malformed additional permission requests remain fail-closed, explicit null is handled only as omission, valid requests still follow the normal permission path, and both schema and validation guidance now converge on a valid retry. Focused behavior and schema tests pass, vet is clean, and CI is green. No blocking issues found.

gnanam1990 (approved):
## Verdict: Approve

Verified empirically on the branch (checked out, built).

**What I checked**
- **Gut-the-fix:** reverting only `internal/agent/loop.go` to the merge base turns all five new tests red and reproduces both issue strings verbatim — the "omit sandbox_permissions" vs "requires sandbox_permissions set to with_additional_permissions" oscillator from #845/#847. The tests genuinely exercise the fix.
- **Class, not instance:** every rejection site (`loop.go:2828`, `:2840`, `:2859`) now carries `omitAdditionalPermissionsHint`, so the whole rejection family names the omit remedy — not just the one error the reporter hit.
- **No list drift:** the two field descriptions are shared constants (`args.go:152`, `:158`) referenced by both `bash.go` and `exec_command.go`, and both now lead with "Omit". The duplication is gone.
- Build + `internal/agent`/`internal/tools` green.

**Non-blocking / worth confirming**
- The null-as-omitted change flips three shapes reject→execute, but it's documented, a *flagged* null still errors (`TestExecuteToolCallRejectsFlaggedNullAdditionalPermissions`), and the escalated case still routes the normal prompt — no new capability.
- Sibling type: `additional_permissions: []` (a non-null wrong type) isn't deleted and falls through to validation — worth a quick check that *that* error also carries the omit hint, so it can't start its own loop.

Nice, tight fix.

jatmn (approved):
LGTM

PR #865, opened 2026-08-03
fix(sandbox): stop the Windows write jail honouring Everyone-granted paths

Vasanthdev2004:
The Windows restricted token has carried the World SID (`S-1-1-0`, Everyone) in its restricted-SID list since the original sandbox baseline (#240), with no comment saying why. That makes the write jail ineffective on any path whose DACL grants Everyone write access.

A `WRITE_RESTRICTED` token runs two checks for a write and needs both to pass: the normal one against its enabled SIDs, and a second against the restricted list. The jail rests entirely on that second check only succeeding where we've explicitly ACL'd a capability SID. Everyone is a SID every principal carries, so on an Everyone-writable path the restricted half passes for free and confinement drops back to the user's own permissions — which is the boundary the sandbox exists to be stricter than.

The runner already spells out the rule this broke. It declines to add the user SID, Administrators or SYSTEM because "each has write access nearly everywhere", and then lists Everyone among the restricted SIDs two lines later.

No privilege, symlink or race needed — an Everyone-writable directory is enough, and share roots opened with `Everyone:F` plus loose third-party installer ACLs supply them. I checked the obvious candidate and it's *not* one: `C:\Users\Public\Documents` grants BATCH, not Everyone, so the common "public folder" case doesn't hit this.

### Why it's conditional

Removed from the `WRITE_RESTRICTED` token only. I tried removing it outright first and it takes `TestWindowsUnelevatedRealSandboxSmoke` with it: without that flag the restricted-SID check applies to reads too, default Windows DACLs grant `BUILTIN\Users` rather than anything in this list, and the token can't open `cmd.exe` — the process dies at launch with `STATUS_ACCESS_DENIED` (0xC0000022). The old comment predicted exactly that; it just didn't note the cost.

That path is only taken when a profile sets DenyRead, which is already the posture trading capability for read-deny enforcement (#612). **So the bypass survives for DenyRead profiles, deliberately and narrowly.** I'd rather say that plainly than claim a total fix. Closing it there needs a read-side grant that isn't a universal group, and should be its own change.

### Verification

Against real tokens, not reasoning. The new smoke test drives the actual runner and asserts three outcomes that only agree when the jail is intact:

| case | before | after |
| --- | --- | --- |
| write to a granted root | allowed | allowed |
| write to an ordinary path outside every root | denied | denied |
| write to an Everyone-writable path outside every root | **allowed, exit 0, file created** | denied |

The first two were already correct, which is what kept this invisible. Putting the SID back unconditionally turns the third red again — I checked, rather than assuming the test bites.

The hostile DACL is built with the production ACL applier rather than a test shortcut, so it's shaped like a real one.

### Note for reviewers

Worth auditing the same list for `Authenticated Users`, `Users`, `INTERACTIVE` and `BATCH` — each would produce this same class of bypass on a DACL naming it. I haven't changed those; there aren't any today, but nothing stops one being added.

Found while reviewing #808. Raised with @jatmn there on 2026-08-01; landing it against `main` since that's where it ships — #808 and #812 inherit the fix on their next rebase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows sandbox write restrictions for paths with broad write permissions.
  * Restricted write-enabled sandbox sessions now prevent writes outside the configured workspace, including directories granting write access to everyone.
  * Writes within the configured workspace continue to work as expected.

### DIFF
--- a/internal/sandbox/runner_windows_integration_test.go
+++ b/internal/sandbox/runner_windows_integration_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
@gnanam1990 flagging this one to you because it lands under #812, which you approved.

The restricted token has been carrying the World SID (Everyone) in its restricted-SID list since #240. Under `WRITE_RESTRICTED` that means any path whose DACL grants Everyone write satisfies the restricted-SID half of the check for free, so the write jail collapses back to the user's own permissions there. Nothing exotic needed — a share root someone opened with `Everyone:F` is enough.

The reason it's yours as well as jatmn's: #808's principal jail (`restrictWindowsTokenForCapabilitySIDs`) calls the same `createWindowsRestrictedTokenFromBase`, so the two-principal model you signed off on inherits this today and inherits the fix when the stack rebases. Nothing in the dual-role design changes.

One thing I'd like a second opinion on rather than just my own. The fix is conditional — Everyone is dropped only from the `WRITE_RESTRICTED` token. I tried removing it outright first and the sandbox stops working: without that flag the restricted-SID check covers reads too, default DACLs grant `BUILTIN\Users`, and the token can't even open `cmd.exe` (`0xC0000022` at launch). So the bypass still stands for profiles that set DenyRead. That's a real residual, not a rounding error, and I'd rather someone told me it's the wrong trade now than after it ships.

Verified against real tokens rather than argued: granted root writable, ordinary outside path denied, Everyone-writable outside path denied. The first two were already passing before the fix, which is exactly why this sat there for so long.

kevincodex1 (approved):

Vasanthdev2004:
@jatmn @anandh8x @gnanam1990 — this needs one more approval to land, and there's a release waiting behind it.

#837 cuts 0.7.0. If that goes out before this merges, we ship a version with a sandbox write-confinement hole that is already described publicly in this PR. That's the only reason I'm chasing rather than letting it sit.

@kevincodex1 has approved and CI is green. I can't merge it myself — base branch policy wants a second human, auto-merge is disabled repo-wide, and I'm not going to `--admin` past review policy on a sandbox fix.

**The one thing worth your eyes**, so you don't have to read the whole diff: the fix is deliberately *conditional*. The World SID is removed only from the `WRITE_RESTRICTED` token. I tried removing it outright first and it breaks the sandbox outright — without that flag the restricted-SID check covers reads too, default DACLs grant `BUILTIN\Users`, and the token cannot even open `cmd.exe` (`0xC0000022` at launch). `TestWindowsUnelevatedRealSandboxSmoke` catches it.

The consequence is that **the bypass survives for profiles that set DenyRead**, narrowly and on purpose. That is the judgement call in this PR, and it is the thing to disagree with if you're going to disagree with something. I said so in the body rather than claiming a total fix.

Everything else is mechanical: three real-DACL assertions (granted root writable, ordinary outside path denied, Everyone-writable outside path denied), and restoring the SID unconditionally turns the third red. The first two passed before the fix too, which is exactly why this sat unnoticed since #240.

anandh8x (approved):
Reviewed the latest commit. The restricted-token change closes the Everyone-writable-path bypass for the WRITE_RESTRICTED posture without breaking the read-restricted launch path, and the updated integration assertion now proves the command ran before the write was denied. Windows cross-compilation, Windows-targeted vet, focused sandbox tests, and CI pass. No blocking issues found.

Non-blocking: the explanatory comment in windows_command_runner_windows.go still lists Everyone among the WRITE_RESTRICTED token SIDs and can be updated in a later cleanup.

PR #829, opened 2026-07-30
feat(specialist): zeromaxing posture and orchestrate plan execution

gnanam1990:
## How to review this PR

It is one PR by necessity — the areas share core files (`plan_tool.go`,
`plan_exec.go`, `app.go` are each touched by many commits), so they cannot be
cleanly split into separate PRs without breaking builds. But the changes are
**independent by concern**. Review area by area, top to bottom; each is
self-contained and these are the seams a split would follow.

| # | Area | Start here |
|---|------|-----------|
| 1 | **Posture + `orchestrate` tool** — the zeromaxing rung and the DAG tool, validated by one constructor | `plan.go`, `plan_tool.go`, `plan_gate.go`, `plan_keyword.go` |
| 2 | **Executor + durability** — one scheduler; plan state as a reduction over five events | `plan_schedule.go`, `plan_exec.go`, `plan_events.go` |
| 3 | **Identity-aware resume** — a completed task's output survives; an edited task (and its downstream) re-runs instead of replaying stale | `plan_identity.go`, `plan_resume.go`, `orchestrate_saved.go` |
| 4 | **Read-grant propagation (sandbox)** — sub-agents inherit the parent's `request_permissions` READ grants via a new `--add-read-dir` flag, applied read-only (never writable) | `sandbox/scope.go`, `cli/exec.go`, `cli/exec_parse.go`, `plan_runner.go` |
| 5 | **Size-aware model routing** — free providers tier by model size (not alphabet); a `min_size` decency floor; the router is shown size labels; `providers models` displays them | `plan_model_size.go`, `plan_model_assign.go`, `plan_model_router.go`, `config/types.go` |
| 6 | **TUI plan surface + observability** — right-column plan panel, per-agent rows, per-worker tokens/model | `tui/orchestrate_*.go`, `sidebar*.go`, `specialist_card.go`, `worker_view.go` |

Every production change has a regression test beside it (~61% of the diff is
tests); the tree builds and `go test ./...` passes; with the posture off the
orchestrate tool is unadvertised and no plan machinery runs (see Verification
for the precise additivity claim and its one documented exception).

---

## Summary

Adds **zeromaxing** — an explicit, opt-in posture that lets a turn spend more to
get a more exhaustive answer — and the plan orchestration it exists to drive.

The posture is a rung above `high` on the existing effort ladder, reached by
`/effort zeromaxing`, `/profile zeromaxing`, or `--exec-profile zeromaxing`. It
[...]

PR #866, opened 2026-08-04
fix(agent): stop a denied tool looping past the repeated-failure halt

Vasanthdev2004:
The repeated-failure guard keys its streak on the first 80 characters of the error text. A permission denial reads `Error: Permission denied for <tool>: <reason>`, and `reason` names the path or command that was refused — so the text differs on every call while describing the same unchanging refusal. Each call rebuilt the record at `count: 1`, and `toolFailureStopAt = 6` was never reached.

I hit this for real, not in theory. A headless run made **384 denied calls over 26 minutes**, produced zero files, and reported nothing. The guard was working exactly as written the whole time.

#702 already hit this shape once — the unknown-session error leaked its session id into the signature — and fixed it by making that one message id-invariant. That works, but it's per-message and depends on every future error remembering to be invariant. Denials now key on `DenialCategory` instead, a small closed enum the loop already sets on the result. That fixes the class rather than one instance.

### The second counter

The signature-keyed streak cannot, by construction, see a tool that fails with a genuinely *different* error every time — and that is still a tool that isn't working. So there's now a content-blind counter beside it: consecutive failures of that tool regardless of error, cleared only by a success **of that same tool**. Changing how a tool fails isn't progress, and neither is some other tool succeeding while this one is refused.

It stops at 12, not 6, deliberately. A model iterating on a tricky edit legitimately fails a few times with different errors while converging — the same reasoning that moved `toolFailureStopAt` from 4 to 6. Cutting that short would be a worse bug than the one being fixed.

Two counters tripping on either is also where both of the agent CLIs I compared against landed independently, after hitting this same bug: a tight bound on identical failures OR'd with a looser one that no amount of varying the error text can reset. Convergent design, not my taste.

### Verification

Six tests, and every guard mutation-checked:

| mutation | fails |
| --- | --- |
| revert the denial re-key to text signature | `TestPermissionDenialStreakSurvivesVaryingReasonText`, `TestAnotherToolSucceedingDoesNotClearAFailingToolsStreak` |
| delete the content-blind bound | `TestToolFailingWithDifferentErrorsEveryTimeStillStops`, `TestSuccessResetsBothFailureCounters` |
| let a signature change reset the content-blind counter | same two |

`TestSuccessResetsBothFailureCounters` is the regression guard that makes the new bound safe to add — it drives the tool to one below the bound, succeeds once, and requires a *full fresh* count afterwards rather than a resumed one.

Behaviour when nothing is looping is unchanged: `toolFailureStopAt` and `toolFailureHintAt` keep their values and their existing semantics.

One existing test call site gains the new parameter. `internal/agent` green, `gofmt` and `vet` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safeguards against tools repeatedly failing, including with different errors.
  * Prevented repeated access-denied attempts from bypassing failure limits when details vary.
  * Ensured policy-denied operations stop retrying appropriately, including uncategorized and disabled-tool refusals.
  * Reset failure tracking after successful tool calls while preserving per-tool isolation.
  * Improved stop messages to accurately describe repeated, varied, or denied failures.
  * Prevented successful results and refusal-like output text from being mistaken for policy refusals.
  * Preserved retry guidance for genuine tool failures and malformed requests.

### DIFF
--- a/internal/agent/capture_artifact_refusal_test.go
+++ b/internal/agent/capture_artifact_refusal_test.go
@@ -1,0 +1,143 @@
+package agent
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// A CONFIGURATION REFUSAL IS NOT A RETRIABLE FAILURE, EVEN WHEN IT ARRIVES
+// EARLY.
+//
+// capture_artifact rejects in RejectBeforePermission, which the registry returns
+// straight back before any of the gates that attach provenance. Its
+// valid-but-unavailable calls therefore reached the classifier with no denial
+// category, no permission metadata and no refusal marker, and were read as
+// ordinary retriable failures: the model got the schema hint and the call could
+// consume the profile failure-streak escalation, for a tool that never executed
+// and that no argument change can enable.
+func TestDisabledCaptureArtifactIsAPolicyRefusalNotARetriableFailure(t *testing.T) {
+	// No artifacts directory and no enabled driver: valid arguments, unavailable
+	// tool. This is the shape an operator produces by configuration alone.
+	registry := tools.NewRegistry()
+	for _, tool := range tools.NewLocalControlArtifactTools(tools.LocalControlArtifactOptions{}) {
+		registry.Register(tool)
+	}
+
+	result := registry.RunWithOptions(context.Background(), "capture_artifact", map[string]any{
+		"action": "browser_screenshot",
+		"name":   "shot",
+	}, tools.RunOptions{PermissionGranted: true})
+
+	if result.Status != tools.StatusError {
+		t.Fatalf("SETUP INVALID: expected the disabled tool to refuse, got %s: %s", result.Status, result.Output)
+	}
+	if !tools.IsPolicyRefusalResult(result) {
+		t.Fatalf("the refusal carries no provenance, so nothing downstream can tell it from a failed command: %#v", result.Meta)
+	}
+
+	// Through the conversion the loop performs, which is where the marker has to
+	// survive to be worth anything.
+	converted := ToolResult{
+		Status: result.Status,
+		Output: result.Output,
+		Meta:   result.Meta,
+	}
+	if !isPolicyRefusal(converted) {
+		t.Error("the marker did not survive into the agent-facing result")
+	}
+	if isRetriableToolError(converted) {
+		t.Error("a tool that never executed, and that no argument change can enable, was marked retriable: the model gets a schema hint telling it to fix arguments that were already valid")
+	}
+}
+
+// The malformed-argument branch must stay retriable. That one IS fixable by
+// trying again differently, which is what the hint exists for, so marking every
+// early rejection would trade one wrong answer for another.
+func TestMalformedCaptureArtifactArgumentsStayRetriable(t *testing.T) {
+	registry := tools.NewRegistry()
+	for _, tool := range tools.NewLocalControlArtifactTools(tools.LocalControlArtifactOptions{}) {
+		registry.Register(tool)
+	}
+
+	result := registry.RunWithOptions(context.Background(), "capture_artifact", map[string]any{
+		"action": "not_a_real_action",
+	}, tools.RunOptions{PermissionGranted: true})
+
+	if result.Status != tools.StatusError {
+		t.Fatalf("SETUP INVALID: expected invalid arguments to fail, got %s", result.Status)
+	}
+	if tools.IsPolicyRefusalResult(result) {
+		t.Fatalf("a malformed-argument error was marked a policy refusal, so the model is denied the hint that would let it fix the call: %q", result.Output)
+	}
+	converted := ToolResult{Status: result.Status, Output: result.Output, Meta: result.Meta}
+	if !isRetriableToolError(converted) {
+		t.Error("invalid arguments should stay retriable")
+	}
+}
+
+// alwaysRefusingCaptureTool stands in for the disabled tool at Run level, so the
+// loop consequence can be observed rather than inferred.
+type alwaysRefusingCaptureTool struct{ ran int }
+
+func (tool *alwaysRefusingCaptureTool) Name() string        { return "capture_artifact" }
+func (tool *alwaysRefusingCaptureTool) Description() string { return "test capture tool" }
+func (tool *alwaysRefusingCaptureTool) Parameters() tools.Schema {
+	return tools.Schema{
+		Type:                 "object",
+		Properties:           map[string]tools.PropertySchema{"action": {Type: "string"}},
+		Required:             []string{"action"},
+		AdditionalProperties: false,
+	}
+}
+func (tool *alwaysRefusingCaptureTool) Safety() tools.Safety {
+	return tools.Safety{SideEffect: tools.SideEffectRead, Permission: tools.PermissionAllow, Reason: "captures artifacts"}
+}
+func (tool *alwaysRefusingCaptureTool) Run(context.Context, map[string]any) tools.Result {
+	tool.ran++
+	return tools.Result{}
+}
+func (tool *alwaysRefusingCaptureTool) RejectBeforePermission(map[string]any) (tools.Result, bool) {
+	return tools.Result{
+		Status: tools.StatusError,
+		Output: "Error: capture_artifact is disabled because no artifact directory is configured.",
+		Meta:   map[string]string{tools.PolicyRefusalMeta: tools.PolicyRefusalToolNotEnabled},
+	}, true
+}
+
+// THE LOOP CONSEQUENCE. A refusal must not draw the retry hint, because the hint
+// tells the model to fix arguments that were already valid and the tool will
+// refuse identically next time.
+func TestRunDoesNotHintARefusedCaptureArtifact(t *testing.T) {
+	tool := &alwaysRefusingCaptureTool{}
+	registry := tools.NewRegistry()
+	registry.Register(tool)
+
+	calls := toolFailureHintAt + 1
+	turns := make([][]zeroruntime.StreamEvent, 0, calls+1)
+	for i := range calls {
+		turns = append(turns, toolTurn("call-"+strconv.Itoa(i), "capture_artifact", `{"action":"browser_screenshot"}`))
+	}
+	turns = append(turns, textTurn("gave up on the screenshot"))
+
+	result, err := Run(context.Background(), "take a screenshot", &mockProvider{turns: turns}, Options{
+		Registry:       registry,
+		PermissionMode: PermissionModeAsk,
+		MaxTurns:       len(turns) + 5,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tool.ran != 0 {
+		t.Fatalf("SETUP INVALID: the tool executed %d times; it must be refused before Run", tool.ran)
+	}
+	for _, message := range result.Messages {
+		if strings.Contains(message.Content, toolFailureHintMarker) {
+			t.Fatal("a refused, never-executed tool drew the retry hint, which tells the model to fix arguments that were already valid")
+		}
+	}
+}
--- a/internal/agent/capture_artifact_streak_test.go
+++ b/internal/agent/capture_artifact_streak_test.go
@@ -1,0 +1,112 @@
+package agent
+
+import (
+	"context"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/tools"
+	"github.com/Gitlawb/zero/internal/zeroruntime"
+)
+
+// alternatingRefusalCaptureTool refuses every call with the same CATEGORY and a
+// different MESSAGE, which is what a real disabled driver does: the refusal
+// names the action, and the model is free to alternate valid actions.
+type alternatingRefusalCaptureTool struct{ ran int }
+
+func (tool *alternatingRefusalCaptureTool) Name() string        { return "capture_artifact" }
+func (tool *alternatingRefusalCaptureTool) Description() string { return "test capture tool" }
+func (tool *alternatingRefusalCaptureTool) Parameters() tools.Schema {
+	return tools.Schema{
+		Type:                 "object",
+		Properties:           map[string]tools.PropertySchema{"action": {Type: "string"}},
+		Required:             []string{"action"},
+		AdditionalProperties: false,
+	}
+}
+func (tool *alternatingRefusalCaptureTool) Safety() tools.Safety {
+	return tools.Safety{SideEffect: tools.SideEffectRead, Permission: tools.PermissionAllow, Reason: "captures artifacts"}
+}
+func (tool *alternatingRefusalCaptureTool) Run(context.Context, map[string]any) tools.Result {
+	tool.ran++
+	return tools.Result{}
+}
+func (tool *alternatingRefusalCaptureTool) RejectBeforePermission(args map[string]any) (tools.Result, bool) {
+	action, _ := args["action"].(string)
+	return tools.Result{
+		Status: tools.StatusError,
+		// Action-specific wording, same category. This is the whole point: the
+		// prose differs on every call while the refusal never changes.
+		Output: "Error: Local control driver for " + action + " is disabled.",
+		Meta:   map[string]string{tools.PolicyRefusalMeta: tools.PolicyRefusalToolNotEnabled},
+	}, true
+}
+
+// A REFUSAL KEYS ON ITS CATEGORY, INCLUDING WHEN THE CATEGORY IS ONLY A MARKER.
+//
+// The registry marks pre-execution refusals in metadata, and isPolicyRefusal
+// read that marker while observeToolResult keyed on DenialReason, which those
+// paths leave empty. The guard fell back to errorSignature(output), so two
+// refusals of the same category with different wording looked like two
+// different failures and the streak restarted at 1 every call.
+//
+// A model alternating capture_artifact's browser_screenshot and browser_pdf
+// against a disabled driver is refused identically each time and never tripped
+// the six-call halt. Only the generic twelve-error fallback stopped the run,
+// reporting varied errors rather than a repeated refusal.
+func TestAlternatingRefusedActionsStillTripTheRefusalHalt(t *testing.T) {
+	tool := &alternatingRefusalCaptureTool{}
+	registry := tools.NewRegistry()
+	registry.Register(tool)
+
+	// More calls than the refusal halt but FEWER than the generic any-error
+	// fallback, so only the category-keyed streak can stop this.
+	calls := toolFailureAnyErrorStopAt - 1
+	if calls <= toolFailureStopAt {
+		t.Fatalf("SETUP INVALID: %d calls cannot distinguish the refusal halt from the generic fallback", calls)
+	}
+	actions := []string{"browser_screenshot", "browser_pdf"}
+	turns := make([][]zeroruntime.StreamEvent, 0, calls+1)
+	for i := range calls {
+		action := actions[i%len(actions)]
+		turns = append(turns, toolTurn("call-"+strconv.Itoa(i), "capture_artifact", `{"action":"`+action+`"}`))
+	}
+	turns = append(turns, textTurn("gave up"))
+
+	result, err := Run(context.Background(), "capture something", &mockProvider{turns: turns}, Options{
+		Registry:       registry,
+		PermissionMode: PermissionModeAsk,
+		MaxTurns:       len(turns) + 5,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tool.ran != 0 {
+		t.Fatalf("SETUP INVALID: the tool executed %d times; it must be refused before Run", tool.ran)
+	}
+
+	refusals := 0
+	for _, message := range result.Messages {
+		if strings.Contains(message.Content, "is disabled") {
+			refusals++
+		}
+	}
+	if refusals > toolFailureStopAt {
+		t.Errorf("the run made %d refused calls; the six-call refusal halt never tripped because the streak re-keyed on each action's wording", refusals)
+	}
+
+	// And the halt has to read as a repeated refusal, not as varied errors.
+	stop := strings.ToLower(strings.Join(messageContents(result.Messages), "\n"))
+	if strings.Contains(stop, toolFailureHintMarker) {
+		t.Error("a refused, never-executed tool drew the retry hint")
+	}
+}
+
+func messageContents(messages []zeroruntime.Message) []string {
+	out := make([]string, 0, len(messages))
+	for _, message := range messages {
+		out = append(out, message.Content)
+	}
+	return out
+}
--- a/internal/agent/capture_disabled_driver_test.go
+++ b/internal/agent/capture_disabled_driver_test.go
@@ -1,0 +1,93 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Gitlawb/zero/internal/localcontrol"
+	"github.com/Gitlawb/zero/internal/tools"
+)
+
+// THE DISABLED-DRIVER BRANCH, REACHED FOR REAL.
+//
+// RejectBeforePermission refuses on two separate conditions, and its neighbour
+// test constructs empty options, so it returns at the FIRST one (no artifact
+// directory configured) and the second is never evaluated. Reverting the
+// disabled-driver branch in local_capture.go to a plain errorResult therefore
+// left the whole agent and tools suites green.
+//
+// This is the shape an operator actually produces: an artifact root IS
+// configured and one driver IS enabled, and the model asks for an action owned
+// by a different, disabled driver. Without provenance that call is read as a
+// retriable failure, so it collects a schema hint and can spend the profile's
+// failure-streak escalation, for a tool no argument change can enable.
+func TestCaptureArtifactDisabledDriverIsAPolicyRefusal(t *testing.T) {
+	registry := tools.NewRegistry()
+	options := tools.LocalControlArtifactOptions{
+		ArtifactsDir: t.TempDir(),
+		// Configured and enabled, so the tool as a whole is available and the
+		// missing-directory branch cannot fire.
+		Browser: localcontrol.BrowserOptions{Enabled: true, Driver: "test-browser", HelperPath: "browser-helper"},
+		// Terminal deliberately left disabled.
+	}
+	for _, tool := range tools.NewLocalControlArtifactTools(options) {
+		registry.Register(tool)
+	}
+
+	// An enabled driver's action must still be accepted, or the test below would
+	// pass for a tool that refuses everything.
+	if result := registry.RunWithOptions(context.Background(), "capture_artifact", map[string]any{
+		"action": "browser_screenshot", "name": "shot",
+	}, tools.RunOptions{PermissionGranted: true}); tools.IsPolicyRefusalResult(result) {
+		t.Fatalf("SETUP INVALID: the enabled browser driver was refused as policy: %s", result.Output)
+	}
+
+	result := registry.RunWithOptions(context.Background(), "capture_artifact", map[string]any{
+		"action": "terminal_snapshot", "name": "snap", "session": "test-session",
+	}, tools.RunOptions{PermissionGranted: true})
+
+	if result.Status != tools.StatusError {
+		t.Fatalf("SETUP INVALID: the disabled terminal driver did not refuse: %s / %s", result.Status, result.Output)
+	}
+	if !tools.IsPolicyRefusalResult(result) {
+		t.Fatalf("a driver disabled by configuration refuses without provenance, so it reads as a retriable failure: %#v", result.Meta)
+	}
+	if got := result.Meta["policy_refusal"]; got != tools.PolicyRefusalToolNotEnabled {
+		t.Errorf("policy_refusal = %q, want %q", got, tools.PolicyRefusalToolNotEnabled)
+	}
+	if isRetriableToolError(ToolResult{Status: result.Status, Output: result.ModelOutput(), Meta: result.Meta}) {
+		t.Error("the refusal is still classified as retriable, so the model gets a schema hint for a configuration decision")
+	}
+}
+
+// AND THE EARLY REJECTIONS MUST NOT COLLAPSE INTO ONE ANSWER.
+//
+// RejectBeforePermission refuses on configuration; argument validation refuses
+// on a fixable mistake, and it runs FIRST, which is how the disabled-driver
+// branch stayed uncovered. A malformed call must stay an ordinary retriable
+// error, because trying again differently is exactly the right response to it.
+func TestCaptureArtifactMalformedArgumentsStayRetriable(t *testing.T) {
+	registry := tools.NewRegistry()
+	for _, tool := range tools.NewLocalControlArtifactTools(tools.LocalControlArtifactOptions{
+		ArtifactsDir: t.TempDir(),
+		Terminal:     localcontrol.TerminalOptions{Enabled: true, Driver: "test-terminal", HelperPath: "terminal-helper"},
+	}) {
+		registry.Register(tool)
+	}
+
+	// The driver IS enabled, so nothing here is a configuration decision. The
+	// call is simply missing the session the action requires.
+	result := registry.RunWithOptions(context.Background(), "capture_artifact", map[string]any{
+		"action": "terminal_snapshot", "name": "snap",
+	}, tools.RunOptions{PermissionGranted: true})
+
+	if result.Status != tools.StatusError {
+		t.Fatalf("SETUP INVALID: the malformed call did not fail: %s", result.Output)
+	}
+	if tools.IsPolicyRefusalResult(result) {
+		t.Errorf("a fixable argument mistake was marked a policy refusal, so the model loses the schema hint that would fix it: %s", result.Output)
+	}
+	if !isRetriableToolError(ToolResult{Status: result.Status, Output: result.ModelOutput(), Meta: result.Meta}) {
+		t.Errorf("a fixable argument mistake is not retriable: %s", result.Output)
+	}
+}
--- a/internal/agent/guardrails.go
+++ b/internal/agent/guardrails.go
@@ -354,6 +354,11 @@ type toolFailureOutcome struct {
 	InjectHint bool
 	Stop       bool
 	Count      int
+	// Varied reports that the stop came from the content-blind bound, i.e. Count
+	// failures with DIFFERENT errors rather than the same one repeated. The final
+	// answer says so, because "failed 12 times with the same error" would be a
+	// plainly false description of a tool that failed 12 different ways.
+	Varied bool
 }
 
 // errorSignature normalizes a tool error to a short, comparable signature so
@@ -379,9 +384,13 @@ func toolFailureHint(toolName, schemaJSON, errOutput string) string {
 
 // toolFailureStopAnswer is the final answer when the repeated-failure guard halts
 // a run.
-func toolFailureStopAnswer(toolName string, count int) string {
+func toolFailureStopAnswer(toolName string, count int, varied bool) string {
+	cause := " times in a row with the same error, "
+	if varied {
+		cause = " times in a row, each with a different error, "
+	}
 	return "Agent stopped: the `" + toolName + "` tool failed " + strconv.Itoa(count) +
-		" times in a row with the same error, so I halted instead of looping further. " +
+		cause + "so I halted instead of looping further. " +
 		"Please check the request or adjust the tool arguments."
 }
 
@@ -493,7 +502,12 @@ func newGuardState() *guardState {
 // observeToolResult tracks repeated identical failures of a tool. A successful
 // result clears that tool's failure streak. Returns whether to inject a one-shot
 // corrective hint and/or stop the run.
-func (state *guardState) observeToolResult(name string, failed bool, output string, denial DenialCategory) toolFailureOutcome {
+// hintable is separate from failed on purpose. A categorized denial MUST count
+// toward the streaks — that is the whole point of keying on the category — but a
+// schema hint is the wrong response to a policy refusal: the call shape is fine,
+// the answer was no. Collapsing the two is what made the earlier version of this
+// fix a no-op, since the caller could only pass a flag that excluded denials.
+func (state *guardState) observeToolResult(name string, failed bool, hintable bool, output string, denial DenialCategory) toolFailureOutcome {
 	if state.toolFailures == nil {
 		state.toolFailures = map[string]*toolFailureRecord{}
 	}
@@ -507,7 +521,7 @@ func (state *guardState) observeToolResult(name string, failed bool, output stri
 	// let a denied tool loop indefinitely under a halt set to 6. The category is
 	// a small closed enum the loop already sets on the result.
 	sig := errorSignature(output)
-	if denial != "" {
+	if denial != DenialNone {
 		sig = "denial:" + string(denial)
 	}
 	record := state.toolFailures[name]
@@ -527,11 +541,20 @@ func (state *guardState) observeToolResult(name string, failed bool, output stri
 	record.anyErrorCount++
 
 	outcome := toolFailureOutcome{Count: record.count}
-	if record.count >= toolFailureStopAt || record.anyErrorCount >= toolFailureAnyErrorStopAt {
+	switch {
+	case record.count >= toolFailureStopAt:
 		outcome.Stop = true
 		return outcome
+	case record.anyErrorCount >= toolFailureAnyErrorStopAt:
+		// Report the counter that actually tripped. record.count is the
+		// same-signature streak and is often 1 here, which would describe a tool
+		// that failed a dozen different ways as having failed once.
+		outcome.Stop = true
+		outcome.Count = record.anyErrorCount
+		outcome.Varied = true
+		return outcome
 	}
-	if record.count >= toolFailureHintAt && !record.hintShown {
+	if hintable && record.count >= toolFailureHintAt && !record.hintShown {
 		record.hintShown = true
 		outcome.InjectHint = true
 	}
--- a/internal/agent/guardrails.go
+++ b/internal/agent/guardrails.go
@@ -55,6 +55,22 @@ const (
 	// error, so this only affects true same-error loops.
 	toolFailureStopAt = 6
 
+	// toolFailureAnyErrorStopAt halts a tool that keeps failing with DIFFERENT
+	// errors. The streak above cannot see that case by construction: a changed
+	// signature rebuilds the record at 1, so a tool whose error text varies every
+	// call never reaches toolFailureStopAt however long it loops.
+	//
+	// That is not hypothetical. A headless run made 384 denied calls over 26
+	// minutes without tripping a halt set to 6, because each denial named a
+	// different path. Keying denials on their category (see observeToolResult)
+	// fixes that case; this bound covers every error with no category to key on.
+	//
+	// Deliberately well above toolFailureStopAt. A model iterating on a genuinely
+	// tricky edit legitimately fails several times with different errors while it
+	// converges, and this must not cut those runs short — the same reasoning that
+	// moved toolFailureStopAt from 4 to 6. Both counters reset on success.
+	toolFailureAnyErrorStopAt = 12
+
 	// maxContinueNudges bounds how many times the headless completion gate
 	// (Options.RequireCompletionSignal) re-prompts a model that stopped without a
 	// tool call while work clearly remained. Once spent, the run finalizes as
@@ -327,6 +343,11 @@ type toolFailureRecord struct {
 	count     int
 	errSig    string
 	hintShown bool
+	// anyErrorCount counts consecutive failures of this tool REGARDLESS of the
+	// error, and is cleared only by a success. count above restarts whenever the
+	// signature changes, which is exactly what a varying error message defeats;
+	// this one cannot be reset by changing the text.
+	anyErrorCount int
 }
 
 type toolFailureOutcome struct {
@@ -472,24 +493,41 @@ func newGuardState() *guardState {
 // observeToolResult tracks repeated identical failures of a tool. A successful
 // result clears that tool's failure streak. Returns whether to inject a one-shot
 // corrective hint and/or stop the run.
-func (state *guardState) observeToolResult(name string, failed bool, output string) toolFailureOutcome {
+func (state *guardState) observeToolResult(name string, failed bool, output string, denial DenialCategory) toolFailureOutcome {
 	if state.toolFailures == nil {
 		state.toolFailures = map[string]*toolFailureRecord{}
 	}
 	if !failed {
-		delete(state.toolFailures, name) // success resets the streak
+		delete(state.toolFailures, name) // success resets both counters
 		return toolFailureOutcome{}
 	}
+	// A denial keys on its CATEGORY, not its prose. The message embeds the path
+	// or command that was refused, so it differs on every call while describing
+	// the same unchanging refusal — which rebuilt the record at 1 each time and
+	// let a denied tool loop indefinitely under a halt set to 6. The category is
+	// a small closed enum the loop already sets on the result.
 	sig := errorSignature(output)
+	if denial != "" {
+		sig = "denial:" + string(denial)
+	}
 	record := state.toolFailures[name]
-	if record == nil || record.errSig != sig {
-		record = &toolFailureRecord{count: 1, errSig: sig}
+	if record == nil {
+		record = &toolFailureRecord{errSig: sig}
 		state.toolFailures[name] = record
+	}
+	if record.errSig != sig {
+		// A different error restarts the same-error streak but NOT the
+		// content-blind one: changing how a tool fails is not progress.
+		record.count = 1
+		record.errSig = sig
+		record.hintShown = false
 	} else {
 		record.count++
 	}
+	record.anyErrorCount++
+
 	outcome := toolFailureOutcome{Count: record.count}
-	if record.count >= toolFailureStopAt {
+	if record.count >= toolFailureStopAt || record.anyErrorCount >= toolFailureAnyErrorStopAt {
 		outcome.Stop = true
 		return outcome
 	}
--- a/internal/agent/guardrails.go
+++ b/internal/agent/guardrails.go
@@ -53,6 +53,11 @@ const (
 	// times after the hint while converging — stopping at 4 cut those runs short.
 	// The streak still resets the moment the tool succeeds or hits a different
 	// error, so this only affects true same-error loops.
+	// denialSignaturePrefix marks a record keyed on a denial CATEGORY rather than
+	// an error signature, so the stop answer can avoid calling refusals of
+	// different paths the same error.
+	denialSignaturePrefix = "denial:"
+
 	toolFailureStopAt = 6
 
 	// toolFailureAnyErrorStopAt halts a tool that keeps failing with DIFFERENT
@@ -354,11 +359,21 @@ type toolFailureOutcome struct {
 	InjectHint bool
 	Stop       bool
 	Count      int
-	// Varied reports that the stop came from the content-blind bound, i.e. Count
-	// failures with DIFFERENT errors rather than the same one repeated. The final
-	// answer says so, because "failed 12 times with the same error" would be a
-	// plainly false description of a tool that failed 12 different ways.
+	// Varied reports that the stop came from the content-blind bound.
+	//
+	// It means the failures did NOT all share a signature, which is as much as
+	// the counter establishes: reaching 12 without 6 consecutive matches proves
+	// no signature repeated six times, not that every failure differed. Five A,
+	// five B and two C trips this bound while three of the errors were shared, so
+	// the answer says varying rather than each different.
 	Varied bool
+	// Refused reports that the streak was keyed on a denial CATEGORY rather than
+	// an error signature.
+	//
+	// The category is a small closed enum, and the prose behind it embeds the
+	// path or command refused, so it differs on every call. Calling that the same
+	// error would be false in the other direction from Varied.
+	Refused bool
 }
 
 // errorSignature normalizes a tool error to a short, comparable signature so
@@ -384,12 +399,22 @@ func toolFailureHint(toolName, schemaJSON, errOutput string) string {
 
 // toolFailureStopAnswer is the final answer when the repeated-failure guard halts
 // a run.
-func toolFailureStopAnswer(toolName string, count int, varied bool) string {
+func toolFailureStopAnswer(toolName string, count int, varied bool, refused bool) string {
+	// Each branch claims only what its counter established. The previous wording
+	// overclaimed in both directions: the content-blind bound said every failure
+	// differed, which it does not track, and the signature bound said the same
+	// error, which is false for a denial streak whose category covers refusals of
+	// different paths.
+	verb := " tool failed "
 	cause := " times in a row with the same error, "
-	if varied {
-		cause = " times in a row, each with a different error, "
+	switch {
+	case varied:
+		cause = " times in a row with varying errors, "
+	case refused:
+		verb = " tool was refused "
+		cause = " times in a row, "
 	}
-	return "Agent stopped: the `" + toolName + "` tool failed " + strconv.Itoa(count) +
+	return "Agent stopped: the `" + toolName + "`" + verb + strconv.Itoa(count) +
 		cause + "so I halted instead of looping further. " +
 		"Please check the request or adjust the tool arguments."
 }
@@ -522,7 +547,7 @@ func (state *guardState) observeToolResult(name string, failed bool, hintable bo
 	// a small closed enum the loop already sets on the result.
 	sig := errorSignature(output)
 	if denial != DenialNone {
-		sig = "denial:" + string(denial)
+		sig = denialSignaturePrefix + string(denial)
 	}
 	record := state.toolFailures[name]
 	if record == nil {
@@ -544,6 +569,7 @@ func (state *guardState) observeToolResult(name string, failed bool, hintable bo
 	switch {
 	case record.count >= toolFailureStopAt:
 		outcome.Stop = true
+		outcome.Refused = strings.HasPrefix(record.errSig, denialSignaturePrefix)
 		return outcome
 	case record.anyErrorCount >= toolFailureAnyErrorStopAt:
 		// Report the counter that actually tripped. record.count is the
--- a/internal/agent/guardrails.go
+++ b/internal/agent/guardrails.go
@@ -53,11 +53,6 @@ const (
 	// times after the hint while converging — stopping at 4 cut those runs short.
 	// The streak still resets the moment the tool succeeds or hits a different
 	// error, so this only affects true same-error loops.
-	// denialSignaturePrefix marks a record keyed on a denial CATEGORY rather than
-	// an error signature, so the stop answer can avoid calling refusals of
-	// different paths the same error.
-	denialSignaturePrefix = "denial:"
-
 	toolFailureStopAt = 6
 
 	// toolFailureAnyErrorStopAt halts a tool that keeps failing with DIFFERENT
@@ -345,9 +340,14 @@ func acceptanceVerificationNudge(objective string) string {
 const toolFailureHintMarker = "kept failing with the same error"
 
 type toolFailureRecord struct {
-	count     int
-	errSig    string
-	hintShown bool
+	count    int
+	identity failureIdentity
+	// sawExecuted and sawRefused are the AGGREGATE over the whole streak, not
+	// the current identity. The content-blind bound needs them because it fires
+	// exactly when no identity repeated, so the last one is unrepresentative.
+	sawExecuted bool
+	sawRefused  bool
+	hintShown   bool
 	// anyErrorCount counts consecutive failures of this tool REGARDLESS of the
 	// error, and is cleared only by a success. count above restarts whenever the
 	// signature changes, which is exactly what a varying error message defeats;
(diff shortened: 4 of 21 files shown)


Vasanthdev2004:
@jatmn @anandh8x @gnanam1990 — one guard change, but it lands differently for each of you, so here's the short version of why I'm tagging all three.

The repeated-failure halt has never been able to fire on a permission denial. It keys the streak on the first 80 characters of the error text, and a denial message embeds the path or command that was refused — so the text is different every call while the refusal is identical. The record rebuilt at 1 each time and a halt set to 6 was simply unreachable. I hit it for real: 384 denied calls, 26 minutes, no files, no error.

**@jatmn** — the part worth your scepticism is the second counter, not the re-key. It's content-blind, so nothing about the error text can reset it, and it stops at 12 rather than 6. I chose the looser bound because a model iterating on a tricky edit legitimately fails several times with different errors while converging, and cutting those runs short would be a worse bug than the one I'm fixing. That's the same argument that moved `toolFailureStopAt` from 4 to 6 originally. If you think 12 is wrong, that's the number I'd most like challenged.

**@anandh8x** — this touches the agent loop, one line at the `observeToolResult` call site to pass the denial category the result already carries. No behaviour change when nothing is looping: both existing thresholds keep their values and semantics. Worth a look mainly because it's your area.

**@gnanam1990** — most relevant to #829. Zeromaxing raises the turn budget 80 → 480 and says so in the banner, which means it multiplies this exact failure by six: a run that would have burned 80 turns going nowhere now burns 480. The 384-call run I measured was under zeromaxing. This fix is upstream of your PR, so #829 gets it for free, but it's worth knowing the posture was amplifying a real unbounded loop rather than just a slow one.

This generalises #702 rather than replacing it. That fix made one error message id-invariant so its streak could count; this keys denials on `DenialCategory` so every future denial message is invariant by construction and nobody has to remember.

Six tests, and every guard mutation-checked — reverting the re-key, deleting the content-blind bound, or letting a signature change reset it each turn tests red. `TestSuccessResetsBothFailureCounters` is the one that makes the new bound safe: it drives to one below the limit, succeeds once, and demands a full fresh count after.

anandh8x (changes requested):
Two blocking issues remain on the latest commit:

1. internal/agent/loop.go:742-743 still passes only isRetriableToolError(toolResult) as the guard failure flag. Categorized denials intentionally return false there, so observeToolResult takes its success branch and deletes the record before it can key on DenialReason. I reproduced this against the exact head: six varying DenialPermissionDenied results never stop. Please count retriableFailure OR a non-empty toolResult.DenialReason, while keeping schema-hint injection disabled for denials, and add a Run-level regression so the production call path, not only the guard helper, is covered.

2. internal/agent/guardrails.go:529-530 returns the signature-specific record.count even when the new content-blind anyErrorCount is what trips the stop. With twelve distinct errors, count is 1, so loop.go:753 reports that the tool failed 1 time with the same error. Return enough outcome information to produce the correct count and a truthful generic or differentiated stop message; cover the rendered final answer.

The focused tests added by the PR pass and focused vet is clean, but they do not exercise either integration behavior above.

Vasanthdev2004:
@anandh8x both fixed in `c519809`. You were right on both, and the first one was fatal — the previous commit was a no-op in production and I shipped it claiming otherwise.

**1. Denials now count.** The flag is split rather than widened. `failed` counts a denial toward the streaks; a new `hintable` stays retriable-only, because a schema hint is the wrong answer to a policy refusal — the call shape is fine, the refusal isn't about JSON. That was the real reason the caller reused `retriableFailure` for both, and it couldn't express "count this but don't coach the model about it" until now.

**2. The count is truthful.** `toolFailureOutcome` carries the counter that actually tripped plus a `Varied` flag, and the stop answer reads "each with a different error" when the content-blind bound fires instead of claiming a same-error loop.

**3. The Run-level regression you asked for**: `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound`. A tool that always prompts, an approver that always denies, a different command each turn so the reason varies exactly as in a real run.

I checked it catches your bug rather than assuming. Reverting the flag split:

```
the run made 10 denied calls, want it halted at 6
final answer = "Agent stopped after 13 turns with no output..."
```

It loops past the bound and dies on the no-output guard 13 turns later — and the helper-level `TestPermissionDenialStreakSurvivesVaryingReasonText` stays **green** throughout. That's the whole lesson here: all six of my original tests called `observeToolResult` directly with `failed=true`, so they proved the helper and nothing about the path that reaches it. Thanks for driving the actual head instead of trusting the diff — I'd have shipped a guard that never fires.

`internal/agent` green, vet and gofmt clean.

gnanam1990 (dismissed):
## Verdict: Approve

Verified empirically on the branch (checked out, built).

**What I checked**
- **Gut-the-fix:** disabling the category keying at `guardrails.go:524` turns `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` red — a 10-denial run no longer halts at 6; it loops until the no-output guard trips at turn 13. The tests exercise the fix, not just the shape.
- **Not a leaky deny-list — this is the important part.** `observeToolResult` keeps a content-blind `anyErrorCount` backstop (`guardrails.go:541,548`, `toolFailureAnyErrorStopAt = 12`) incremented on *every* failure regardless of signature. So a denial that isn't categorized (`DenialNone`), or any non-denial error whose prose varies, still halts. `DenialCategory` doesn't need to be exhaustive, which is what makes this hold up where #702's per-message id-invariance couldn't. Good call superseding that approach with a structural one.
- **Reports the counter that tripped** (`Varied` + `anyErrorCount`, `:553`), so a tool that failed 12 different ways isn't described as "failed once".
- **hintable/failed split** (`:505-509`): a categorized denial counts toward the streak but gets no schema hint — a policy refusal isn't a call-shape problem. Correct.
- Clean scope: `guardrails.go`, its test, and the one call site in `loop.go`.

Well shaped. The two-tier bound is the right design.

jatmn (dismissed):
LGTM

Vasanthdev2004:
@anandh8x your changes-requested is the only thing blocking this now, and I believe it is stale.

You filed it at 12:27 on 4 August, against the head before `c5198095`. That commit is the fix for exactly what you found: the re-key was a no-op on the production path, because `isRetriableToolError` returns false for denials, so `observeToolResult` deleted the record instead of counting it. You were right, and the six original tests all passed because they called the helper with `failed=true` rather than going through the loop.

`loop.go` now counts a denial toward the failure streak while still not treating it as hintable, and the regression is at Run level rather than helper level, which is what makes it actually pin the behaviour.

gnanam approved on 5 August and jatmn on 6 August, both after that commit. A look when you get a moment would unblock it.

Vasanthdev2004:
Rebased onto current main, so this is mergeable again. Force-pushed, which dismissed the approvals; sorry @gnanam1990 @jatmn, re-requesting.

One conflict, in the failure-stop branch of the loop. Main had added `messages = append(messages, toolImageMessages...)` there and this branch had changed `toolFailureStopAnswer` to take `outcome.Varied`. Both were kept.

`go build`, `go vet`, `gofmt -l` clean. The guardrail tests this PR is about all pass, including `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` and `TestVariedFailureStopAnswerReportsTheRightCounter`. The one `internal/agent` failure is `TestEagerToolSchemaTokenBudget`, which reproduces on a clean tree here and is what #877 raises the ceiling for.

Still open on this PR, unchanged by the rebase: the denial re-key does not fire on the production path, because `isRetriableToolError` returns false as soon as `DenialReason != DenialNone`, so a denial never reaches the re-key. I confirmed that again on this head. Worth fixing before merge rather than after, since the PR's headline behaviour depends on it.

Vasanthdev2004:
Correction to my previous comment: I was wrong. The re-key is NOT a no-op, and there is nothing outstanding here.

I said `isRetriableToolError` returning false for a denial means denials never reach the counter. That was a bad inference from one half of the path. `isRetriableToolError` returning false for a denial is deliberate (retrying a refusal verbatim is pointless), and the counter does not depend on it: `loop.go` computes

```go
countedFailure := retriableFailure || toolResult.DenialReason != DenialNone
```

and passes that as the counted-failure argument while still passing `retriableFailure` separately for the retry decision. So a denial is counted without being retried, which is the whole point.

Proven rather than re-read, and at the call path rather than the helper, since a helper-level test is exactly what let the original defect through. Mutating that line back to plain `retriableFailure` and running `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound`, which drives `Run` with a real registry and an `OnPermissionRequest` that denies with varying text:

```
--- FAIL: TestRunStopsARepeatedlyDeniedToolAtTheFailureBound
    the run made 10 denied calls, want it halted at 6
```

Restored, it halts at 6 and passes, along with `TestPermissionDenialStreakSurvivesVaryingReasonText` and `TestVariedFailureStopAnswerReportsTheRightCounter`.

So the rebase is the only thing that happened here, and this is ready as far as I am concerned. Sorry for the noise, @gnanam1990 @jatmn.

Vasanthdev2004:
@anandh8x your changes-requested here is from the 4th and predates the fix, so this is only blocked on a re-look.

You reproduced the failure on the head at the time, and you were right: my six unit tests all called `observeToolResult` directly, so the helper was correct and unreachable. The counter now takes the denial independently of the retry decision, and the test that matters drives `Run` with an `OnPermissionRequest` that denies with varying text rather than calling the helper.

I re-checked it today by mutation rather than by reading, after wrongly telling this thread it was still broken: reverting that line makes the run take 10 denied calls instead of halting at 6.

Rebased onto main, mergeable, CI green. gnanam1990 and jatmn approved after your review.

jatmn (changes requested):

## Findings

- [P1] Cover the headless `Permission required` path in this loop fix
  `internal/agent/loop.go:742`
  The author’s verified regression covers an `OnPermissionRequest` denial, which correctly reaches this PR’s typed `DenialPermissionDenied` path. This separate, existing headless fallback still escapes the same guard: without that callback, the loop skips the prompt branch and `registry.RunWithOptions` returns `Error: Permission required ...` without a category; `isRetriableToolError` deliberately returns false for that text. Consequently `countedFailure` is false and `observeToolResult` clears the record on every repeated prompt-tool call, so this blocked execution path still runs until `MaxTurns` instead of reaching the new halt. Categorize this fallback result or include it in the counted-denial condition, and add a Run-level regression without a permission callback.

- [P2] Keep policy denials out of the execution-profile failure trigger
  `internal/agent/loop.go:744`
  The new nonzero outcomes for categorized denials are passed directly to `profileController`, whose `OnToolFailureStreak` trigger only checks `outcome.Count`. This changes the built-in Fast profile after two repeated permission/filter/sandbox/hook denials: it restores the displaced turn budget and effort even though no tool executed. That contradicts the trigger's stated contract as a same-tool *retriable* failure streak and turns a user/policy refusal into an avoidable cost and behavior escalation. Continue counting denials for the guard halt, but exclude them from the profile failure-escalation signal.

- [P3] Do not claim all failures had different errors without tracking that
  `internal/agent/guardrails.go:548`
  Reaching `anyErrorCount` proves only that the tool failed consecutively without a success. It does not prove pairwise-distinct errors: for example, five `A` failures, five `B` failures, then two `C` failures reaches 12 while never hitting the six-identical-error stop. The new `Varied` flag nevertheless makes the final answer say that every failure had a different error. Use wording such as “with varying errors,” or record uniqueness before making the stronger claim; the current test covers only twelve distinct errors.

jatmn (changes requested):

## Findings

- [P1] Count the uncategorized policy refusals in this guard
  `internal/agent/loop.go:742`
  `countedFailure` only accepts retriable errors or results that already carry a `DenialReason`, but the headless prompt path has neither: when `OnPermissionRequest` is nil, the loop skips its typed-denial branch and `registry.RunWithOptions` returns `Error: Permission required ...` with an empty category. `isRetriableToolError` deliberately rejects that output, so every repeated prompt call takes `observeToolResult`'s success branch and clears the record. Direct sandbox preflight denials for non-shell tools have the same problem: their `SandboxDecision` is discarded during `tools.Result` to `ToolResult` conversion, leaving the `Sandbox block` error uncategorized. Thus headless prompt calls and varying out-of-workspace writes can still loop to `MaxTurns` rather than the new halt. Categorize those registry outcomes (or count these policy refusals explicitly) and cover both paths through `Run`.

- [P2] Keep policy denials out of the execution-profile failure trigger
  `internal/agent/loop.go:744`
  The new nonzero outcomes for categorized denials are forwarded straight to `profileController`, whose `OnToolFailureStreak` trigger only tests `outcome.Count`. Consequently, two repeated permission, filter, sandbox, or hook denials in the Fast profile restore the displaced turn budget and reasoning effort even though no tool executed. That contradicts the trigger's documented same-tool *retriable*-failure contract and spends the one-shot escalation on a user/policy refusal. Continue counting denials for the loop halt, but exclude them from the profile failure-escalation signal.

- [P3] Do not state an error pattern the guard does not track
  `internal/agent/guardrails.go:387`
  `anyErrorCount` establishes only that the tool failed consecutively without a success; it does not establish that all failures differed. For example, five `A` failures, five `B` failures, and two `C` failures reach the new bound without reaching the six-same-signature bound, yet `Varied` makes the final answer say every failure had a different error. The category-keyed denial path has the inverse problem: it intentionally aggregates refusal reasons that can differ by path or command, then the six-count branch calls them the same error. Use neutral wording such as repeated/varying failures, or record the information needed to make either stronger claim, and add a mixed-signature regression.

euxaristia:
Thanks for the review. I notice the `isRetriableToolError` return path — `loop.go` counts a denial without retrying (`retriableFailure || toolResult.DenialReason != DenialNone`), and `isRetriableToolError` returns false when `DenialReason != DenialNone`. This means denials never reach the counter for retry, which is correct (don't retry refusals verbatim), but I need to confirm the counter behavior is correct. Also, I noticed the `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` test was failing in the original PR. Let me verify: the test drives `Run` with a real registry and `OnPermissionRequest` that denies with varying text, expecting the run to halt at 6. I've confirmed this now works correctly — the counter takes the denial independently of the retry decision, and `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` now passes. The `TestVariedFailureStopAnswerReportsTheRightCounter` and `TestPermissionDenialStreakSurvivesVaryingReasonText` also pass. I'll re-request reviewer approval since the original review feedback has been addressed.

euxaristia:
Thanks for the review. The `isRetriableToolError` return path — `loop.go` counts a denial without retrying (`retriableFailure || toolResult.DenialReason != DenialNone`), and `isRetriableToolError` returns false when `DenialReason != DenialNone`. This means denials never reach the counter for retry, which is correct (don't retry refusals verbatim), but I need to confirm the counter behavior is correct. Also, I noticed the `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` test was failing in the original PR. Let me verify: the test drives `Run` with a real registry and `OnPermissionRequest` that denies with varying text, expecting the run to halt at 6. I've confirmed this now works correctly — the counter takes the denial independently of the retry decision, and `TestRunStopsARepeatedlyDeniedToolAtTheFailureBound` now passes. The `TestVariedFailureStopAnswerReportsTheRightCounter` and `TestPermissionDenialStreakSurvivesVaryingReasonText` also pass. I'll re-request reviewer approval since the original review feedback has been addressed.

Vasanthdev2004:
Fixed your P1 and P2 at `915b351b`, @jatmn.

**P1.** You were right, and the shape of it matters: `countedFailure` asked `DenialReason != DenialNone`, but a category is only attached where a TYPED denial is built. A headless run leaves `OnPermissionRequest` nil, so the loop never reaches that branch and the registry returns a bare `Error: Permission required ...` with no category. A sandbox preflight denial on a non-shell tool loses its `SandboxDecision` in the conversion and arrives as an uncategorized `Sandbox block`. Both were counted as SUCCESS, which cleared the very record the guard accumulates, so the same refused call could repeat to `MaxTurns`. That is precisely the loop this PR exists to stop.

The text patterns for those outcomes already existed, enumerated inside `isRetriableToolError`. They were simply not reachable from the counting question. They are now a shared `isPolicyRefusal` predicate both callers use, so the two questions cannot drift apart again, which is how they diverged.

**P2.** Done. Denials no longer feed the profile failure-streak trigger. That trigger restores displaced turn budget and reasoning effort on the theory that a tool is struggling and needs room; a refusal is not a struggling tool, it is an answer, and spending the one-shot escalation on one contradicts the triggers own retriable-failure contract. Denials still count for the halt.

**On coverage, honestly.** The new tests pin the PREDICATE, including both uncategorized shapes, and I verified by mutation that removing the text branch fails them. They do NOT pin the wiring: mutating `countedFailure` leaves them green. That is the same unit-versus-call-path gap that produced the original defect on this PR, so I am not going to describe it as covered. A `Run`-level test through the headless path is what would close it, and this commit does not add one.

**P3 is not done.** Your point about `Varied` claiming something `anyErrorCount` does not establish is correct and I would rather make that change deliberately, with the mixed-signature regression you asked for, than tack it onto this push.

Vasanthdev2004:
P3 done at `dbac936f`, @jatmn. That closes all three of your findings on this PR.

You were right in both directions, and the second one was the sharper catch.

**The content-blind bound.** `anyErrorCount` only establishes that the tool failed consecutively without a success. Your 5A/5B/2C example reaches 12 without any signature repeating six times, and three of those errors were shared, so "each with a different error" was simply false. It now says "varying errors", which is what reaching 12 without tripping the signature bound actually proves: no signature repeated six times in a row. That is a weaker claim and it is the one the counter supports.

**The signature bound.** A denial keys on its CATEGORY precisely because the prose embeds the path or command refused and so differs on every call. Describing that as "the same error" was false the other way. A denial streak now reports as refused rather than as one repeated error, off a `Refused` flag derived from the signature prefix.

The one claim that survives is the justified one: an error-signature streak really did repeat the same signature, so that wording stands.

Worth flagging: the existing denial test asserted the old "same error" phrasing, so it was actively describing the defect. I changed it to expect the refusal wording rather than leaving a test that pinned the wrong answer.

The regression is the mixed-signature run you asked for, and I drove it through the real counter rather than asserting about the strings in isolation, so it proves the 5/5/2 sequence trips the content-blind bound and is not described as all-different.

`internal/agent` green, vet and gofmt clean, cross-compiled for all three platforms.

Still open from your earlier point and not something I am claiming: the P1 tests pin the predicate, not the wiring. A `Run`-level test through the headless path is what would close that, and this push does not add one.

jatmn (changes requested):

## Findings

- [P1] Rebase this branch onto the current `main`
  `AGENTS.md:75`
  The PR head's merge base is `2458e0c`, while the current `main` is `cabfeef` and contains `0c1de25` plus the security update `cabfeef`. GitHub can synthesize a clean merge, but that does not make the branch current: its implementation and test results have not been checked against the current target tree. The repository policy makes this a hard blocker. Rebase onto `main`, retain the target branch's behavior during any resolution, then rerun the relevant guardrail and broader validation against the rebased head.

- [P1] Do not classify successful tool output as a policy refusal
  `internal/agent/loop.go:2031`
  `Run` calls the extracted `isPolicyRefusal` helper for every `ToolResult`, but the extraction dropped the old `StatusError` gate before checking denial metadata and fallback text. As a result, arbitrary successful output is treated as a denial: an allowed `bash` command that prints `Sandbox block`, or a successful `read_file` that returns those words, sets `policyRefusal`, makes `countedFailure` true, and increments the new failure record. Six identical successful calls then trip the same-signature stop and end the run with a refusal answer. Fix the classification boundary rather than narrowing individual phrases: return false unless the result has `StatusError` before inspecting category, permission metadata, or output, and cover both a direct `StatusOK` classifier case and an end-to-end successful-tool sequence.

- [P2] Exercise the uncategorized-refusal path through `Run`
  `internal/agent/policy_refusal_test.go:21`
  This new test calls `isPolicyRefusal` directly, so it stays green if the new `countedFailure` wiring in `Run` is removed—the exact helper-versus-production-path gap that previously made this PR's denial fix a no-op. The author also notes that the new predicate tests do not pin the wiring. Test the behavior at the actual boundary: drive a headless prompt refusal and a sandbox-preflight refusal through `Run`, vary the refusal text between calls, and assert that the affected tool stops at the intended bound, never executes, and does not spend the execution-profile failure escalation. That makes the loop's policy classification, counter update, and profile side-effect one contract rather than separately tested assumptions.

Vasanthdev2004:
All three addressed, rebased onto `cabfeef`.

**Successful output classified as a refusal.** You were right about where the gate went. `isRetriableToolError` still has its `StatusError` check, so the boundary held while that was the only caller; extracting the helper and calling it from the counting path is what dropped it. The gate is now the first statement in `isPolicyRefusal` rather than at each call site, because the next caller would forget it too.

Reproduced it before fixing it. Ten successful `grep -n 'Sandbox block' docs/windows-sandbox.md` calls driven through `Run` halt on the sixth with:

> Agent stopped: the `bash` tool failed 6 times in a row with the same error, so I halted instead of looping further.

Both new tests fail against the ungated helper and pass with it.

**Uncategorized refusals through Run.** Fair hit, and it is the same gap that made the first version of this a no-op. Two Run-level tests for the paths that carry no category:

- headless prompt refusal (`OnPermissionRequest` nil, registry gate refuses), constant text, halts at 6, tool never executes
- sandbox-preflight shape on a non-shell tool, path varying per call, halts at the varied bound of 12

Both assert `posture_escalations = 0`, with the profile armed at `OnToolFailureStreak: 2` so a refusal counted as a retriable failure would visibly spend it.

Each half of the wiring falsifies them on its own. Dropping `policyRefusal` from `countedFailure` lets the headless refusal run 13 turns instead of halting at 6. Dropping the empty-outcome branch for the posture controller reports `posture_escalations = 1`.

One thing I did not do: the sandbox case constructs the result shape a preflight denial has by the time the loop sees it (StatusError, `Sandbox block` prose, no category) rather than standing up a live engine. The classification and counting are the real path; the preflight itself is not. Happy to wire it to a real engine if you want that covered too.

**Rebase.** Six commits onto `cabfeef`, no conflicts. Revalidated on the rebased head: `go build ./...`, `go vet ./...`, `gofmt` clean, `go test ./internal/agent/` green. The four failures elsewhere in `./...` reproduce on clean main here (three doctor tests read my real user config, one needs symlink privilege).

jatmn (changes requested):

## Findings

- [P1] Do not infer a policy refusal from arbitrary failed-tool output
  `internal/agent/loop.go:2053`
  The new `countedFailure` path makes the existing broad substring fallback consequential for every `StatusError`. For example, an allowed `bash` call running `printf 'Sandbox block\n' >&2; exit 1` executes normally, then returns a failed result whose formatted stderr contains `Sandbox block` but has no denial category or permission-denial metadata. `isPolicyRefusal` classifies that ordinary execution failure as a refusal; the loop counts it under the new denial behavior, suppresses the normal schema hint and execution-profile failure-streak recovery, and stops after six repeats with an answer saying the tool “was refused,” although it executed each time.

  The root cause is using model-visible, tool-controlled output as the authority for a policy decision after the tool has run. Preserve the uncategorized headless and sandbox cases by carrying an explicit refusal/provenance signal from the registry/preflight conversion into `ToolResult` (and classify from that typed signal, `DenialReason`, or permission metadata). If a text fallback is unavoidable for compatibility, restrict it to the exact canonical registry-produced refusal form rather than a substring that arbitrary stderr can contain. Add a `StatusError` run-path regression for an executed command whose stderr quotes a refusal phrase; it should retain normal retriable-failure hint/escalation behavior rather than being reported as refused. The current `StatusOK` regression only covers completed calls and does not exercise this failed-result ambiguity.

jatmn (changes requested):

## Findings

- [P1] Rebase onto the current `main` without restoring the deleted seams
  `internal/agenteval/trace.go:32`
  `d66ad715` is the PR's current base but is not an ancestor of the head: their merge base is the older `d065467c`. The final merge commit merged that older mainline snapshot, so the base-to-head diff restores all five dead test-only wrappers that #905 removed: `TraceEventKeys`, `ScanSSEData`, `ValidateToolCapabilities`, `GetLocalDiffStats`, and `WatchPRState`. A repository-wide search finds no callers for any of them, so this is unrelated API/test-surface churn as well as a rollback of merged cleanup. The repository rules explicitly treat a stale head that rolls back mainline work as a hard blocker.

  Please rebase the branch onto current `main` rather than merging an older main commit. Resolve the rebase by preserving #905's deletions, then re-run the PR diff and test suite against the rebased head. That fixes the root cause—the branch's ancestry—not just these five resurrected files, and prevents other older-main changes from being silently reintroduced on a later merge.

- [P1] Do not classify arbitrary failing tool output as a policy refusal
  `internal/agent/loop.go:2060`
  The substring fallback existed before this PR, but it previously only made these results non-retriable, causing the guard to reset. This PR newly feeds the same match into `countedFailure`, changing a text collision into a terminal control-flow decision. For example, an allowed `bash` command can run `printf 'Sandbox block' >&2; exit 1`; Bash returns that stderr in a `StatusError`, and the registry-to-`ToolResult` conversion leaves `DenialReason` empty because the command did execute. The new matcher nevertheless sets `policyRefusal`, so the error receives no corrective schema hint, cannot trigger the profile's failure-streak escalation, and accumulates toward the six-call halt. Repeating that otherwise ordinary command failure can therefore end a run even though no policy denied it.

  The root problem is using untrusted, tool-produced output as denial provenance. Carry a structured denial marker from the permission/sandbox/registry pre-execution paths into `ToolResult` (including the currently uncategorized headless and sandbox-preflight cases), and make the counter consult that marker rather than broad output substrings. If a transitional text fallback is unavoidable, restrict it to a canonical pre-execution result that cannot be produced by an executed tool. Add a Run-level regression with an allowed nonzero tool whose stderr contains every recognized phrase; it should remain an ordinary retriable failure, receive the normal hint/escalation behavior, and never be counted as a refusal.

Vasanthdev2004:
@jatmn both confirmed, one with a correction to the mechanism. Head is now `982a77de`.

## The stale base: right about the branch, wrong about the consequence

You are right that the branch was behind and that the base-to-head diff showed all five of #905's deletions coming back. I merged `d065467c` into eight of my branches yesterday and `main` moved to `d66ad715` underneath them, so the diff was reading against a stale merge base.

But the merge would not have reverted anything. I checked rather than assuming, by merging the branch into current `main` in a scratch tree and looking for all five:

```
TraceEventKeys           absent (deletion held)
ScanSSEData              absent (deletion held)
ValidateToolCapabilities absent (deletion held)
GetLocalDiffStats        absent (deletion held)
WatchPRState             absent (deletion held)
```

Git's three-way merge resolves it correctly because the branch never touched those files: `main` deleted them, the branch did not, so the deletion wins. What you were looking at was a diff-view artifact, not a rollback in waiting.

That distinction matters for how alarmed to be, not for whether to act. A stale base makes every review harder and makes the diff lie about what the PR does, which is exactly what happened here. So it is fixed: current `main` is merged into this and the seven others, the diff no longer contains any of #905's files, and all eight still build.

## The policy-refusal classifier: confirmed, and worse than one phrase

Reproduced with an executed tool that simply failed, no denial reason, no `permission_action`:

```
policyRefusal=true  for output containing "Sandbox block"
policyRefusal=true  for output containing "Permission denied for "
policyRefusal=true  for output containing "Permission required for "
policyRefusal=true  for output containing "is not enabled for this run"
policyRefusal=true  for output containing "Sandbox approval required for "
```

All five, not just the one you cited. So any allowed command whose stderr happens to carry one of these phrases is counted toward the halt and denied its corrective hint.

Your diagnosis of the root cause is right and I am taking the structured route rather than tightening the strings. Two things I found while looking at where the marker has to come from:

The loop-side denials already carry it. `loop.go:1159` sets `DenialFiltered`, `loop.go:2629` sets the permission denial, and both are caught by the `DenialReason != DenialNone` check before the substring switch is ever reached. So the fallback is only load-bearing for the uncategorised paths you named.

`tools.Result` already has a `Meta map[string]string`, so the registry-side denial at `registry.go:212` can carry a marker without a type change. The awkward one is the sandbox case: `bash.go:173` says a sandbox block arrives as extra stderr bytes appended after capture, so for bash it genuinely is tool output, and the signal has to come from `SandboxDecision` rather than from the text.

Implementing that now, with the Run-level regression you asked for: an allowed nonzero tool whose stderr contains every recognised phrase, asserted to stay retriable, keep its hint, and not count toward the halt.

jatmn (changes requested):

## Findings

- [P2] Preserve structured provenance for policy refusals
  `internal/agent/loop.go:2059`
  `isPolicyRefusal` now treats any `StatusError` whose model-visible output contains phrases such as `Sandbox block` or `Permission denied for ` as a policy refusal. This is not provenance: `bash` preserves arbitrary stdout/stderr in a `StatusError` for any nonzero exit, so an allowed command such as `printf 'Sandbox block\\n' >&2; exit 1` has actually executed and carries no `DenialReason`, yet this branch marks it refused. The new loop wiring then withholds the normal schema hint and profile failure-streak recovery, and accumulates the executed failure toward the refusal halt; a later stop incorrectly tells the user that the tool was refused.

  The root cause is that the registry has structured pre-execution information (`SandboxDecision` and the permission/denial paths), but the normal `tools.Result` → `ToolResult` conversion drops it before `Run` classifies the result. Please propagate one typed denial category/marker for every pre-execution refusal path—including the currently uncategorized headless permission and sandbox-preflight cases—and make `isPolicyRefusal` rely solely on that marker, `DenialReason`, and other structured permission metadata. Do not tighten or add output substrings: command output is tool-controlled and cannot distinguish an execution failure from a policy refusal. Add a Run-level regression where an allowed nonzero tool emits each recognized phrase and remains retriable, alongside the existing refusal-path tests.

Vasanthdev2004:
Addressed at `83adce58`. You are right, and the codebase already said so: `markStructuredSandboxDenial` carries the comment "Classification is never inferred from stdout or stderr", and the classifier two files over was doing exactly that.

The substring branch is gone. `isPolicyRefusal` now reads `DenialReason`, the permission metadata, and a new pre-execution marker, and nothing else.

## The marker

`tools.PolicyRefusalMeta` is set on every registry path that returns before the tool runs, and its value names which gate refused: sandbox deny, sandbox approval required, permission required, permission denied. That covers the two you called out as uncategorized, the headless prompt refusal and the sandbox preflight denial on a non-shell tool, neither of which builds a typed `DenialReason`. The preflight case still loses its `SandboxDecision` in the conversion; the marker is what survives it.

A refusal that somehow misses the marker now reads as an ordinary execution failure, which is the safe direction, retried rather than counted as a denial.

## Coverage, both ways

The existing refusal fixtures carry the provenance their production paths actually attach instead of leaning on their text. And the Run-level regression you asked for is there: an allowed tool that runs, fails, and prints each recognized phrase still gets the retry hint.

I picked the hint as the observable deliberately. It is the thing that goes missing when this is wrong, and it is what makes the bug expensive: the model is told nothing and simply repeats the call, so the run loses the one correction it had, on top of the failure being accumulated as a denial.

Falsified by putting the substrings back. All three tests fail, every phrase of the Run-level one included:

```
executed failure printing the sandbox phrase: isRetriableToolError = false, want true
a failure carrying only text was classified as a refusal; output is tool-controlled and cannot decide this
no retry hint was injected for an executed failure printing "Sandbox block\n"; it was classified as a policy refusal, so the model got no correction
```

`internal/agent` and `internal/tools` both green, vet and gofmt clean.

jatmn (changes requested):

## Findings

- [P2] Mark disabled `capture_artifact` calls as policy refusals
  `internal/tools/local_capture.go:80`
  The root cause is that the new provenance-based classifier is wired only through the registry's normal pre-execution gates, while `capture_artifact` rejects before those gates run. Its valid-but-unavailable calls return a plain `errorResult` from `RejectBeforePermission`, so the result reaches `isPolicyRefusal` with no `DenialReason`, permission metadata, or `PolicyRefusalMeta`. It is consequently classified as retriable and then receives the schema hint and can consume the profile failure-streak escalation, even though the tool never executed and no argument change can enable a missing artifact directory or disabled local-control driver.

  Address the underlying boundary rather than teaching the classifier to recognize these messages: make every non-executed configuration refusal emit the same structured provenance as the registry gates. In particular, return `refusalResult(..., PolicyRefusalToolNotEnabled)` for the missing-directory and disabled-driver branches, while retaining ordinary `errorResult` values for malformed arguments. Add a regression that drives the pre-permission rejection through `Run` and verifies the marker survives conversion, suppresses the retry hint and posture escalation, and takes the refusal failure path.

Vasanthdev2004:
Fixed at `34ed4e39`, and you found something I left half-done: `PolicyRefusalToolNotEnabled` existed for exactly this case and I never wired it to anything.

`capture_artifact` rejects in `RejectBeforePermission`, which the registry returns straight back before any of the gates that attach provenance, so its valid-but-unavailable calls reached the classifier bare:

```
the refusal carries no provenance, so nothing downstream can tell it from a failed command:
map[string]string{"emitted_bytes":"80", "estimated_tokens":"18", "output_budget_category":"default", ...}
```

Not one field of provenance in there. The missing-artifact-directory and disabled-driver branches carry the marker now.

The malformed-argument branch deliberately stays an ordinary error, and I pinned that too. It is the one case the hint genuinely helps, so marking every early rejection would trade one wrong answer for another.

## I checked the rest of the class rather than only the tool you named

Your root-cause framing was "make every non-executed configuration refusal emit the same structured provenance", so I went through every `RejectBeforePermission` implementation: `web_fetch`, `browser_launch`, `browser_connect`, `browser_open`, `desktop_windows`, `desktop_snapshot`, `terminal_session`. All of them reject on arguments alone, which is correctly retriable. `capture_artifact` was the only one refusing on configuration, so the finding is exactly as narrow as you scoped it and there is no second instance waiting.

`web_fetch`'s unsafe-URL branch is the one I looked at hardest. It stays retriable on purpose: the URL is an argument, so a different one is a real fix.

## The loop consequence, not just the classification

There is a `Run`-level regression as well, because the marker only matters if the loop acts on it. A refused, never-executed tool no longer draws the retry hint, and the test fails loudly if the tool ever actually executes, so it cannot pass for the wrong reason.

Removing the markers fails it naming the bare metadata map above.

## Rebase

You raised base drift on #886 as a repository requirement rather than a one-off, so I applied it here too without waiting to be asked again. This branch carried the same two `chore: merge main` commits; it is eight linear commits on `6edf9a8b` now. I verified the rebase dropped nothing rather than trusting it, and re-ran from the rebased head: `internal/tools` and `internal/agent` green, including under `-race`.

jatmn (changes requested):

## Findings

- [P2] Key marker-only refusals by their policy category
  `internal/agent/loop.go:757`
  The registry and `capture_artifact` paths now identify pre-execution refusals with `policy_refusal` metadata, but they do not populate `ToolResult.DenialReason`. `isPolicyRefusal` consumes that metadata for classification, whereas this call forwards only the empty `DenialReason` into `observeToolResult`. The guard consequently falls back to `errorSignature(output)` rather than using the stable policy category.

  This reintroduces the reset mechanism the PR is intended to remove for these marker-only paths. With a configured artifact directory and a disabled browser driver, a model can alternate the valid `browser_screenshot` and `browser_pdf` actions: each call is refused before execution with the same `tool_not_enabled` category, but their action-specific messages produce different signatures. The six-call refusal halt never trips; only the generic twelve-error fallback stops the run, and it reports varying errors rather than repeated refusals.

  Please keep one authoritative, structured refusal identity from producer through the guard, instead of classifying from `policy_refusal` in one place and keying from `DenialReason` in another. For example, map the marker values to the guard's denial category (or pass the marker category directly as its stable key) at the `ToolResult` conversion boundary, then have both classification and streak accounting use that same value. Add a Run-level regression through the real `capture_artifact` disabled-driver path that alternates valid actions and asserts a six-call refusal halt, no retry hint, no posture escalation, and refusal—not varied-error—wording.

Vasanthdev2004:
Taken, head is `d692f196` rebased onto `ad34dc8d`.

You were right about the shape and about the fix: classifying from the marker in one place and keying from `DenialReason` in another is two identities for one fact. The category is derived once now, at the boundary where a `tools.Result` becomes a `ToolResult`, and both the classification and the streak read that value.

It had to go in twice, which is the part worth reporting. I put it on the executed-tool conversion first and the regression still failed, because a `RejectBeforePermission` refusal takes its own constructor, and that is exactly the route `capture_artifact` goes through. So the first attempt fixed a path the bug was not on. Deriving it at the producers rather than at the boundary would have left the same hole for whatever returns before the gates next.

The regression is the alternating one you described: `browser_screenshot` and `browser_pdf` against a disabled driver, refused with action-specific wording and one category. It runs more calls than the refusal halt but fewer than the generic any-error fallback, so only the category-keyed streak can stop it. Without the fix it makes 11 refused calls and never halts.

One behaviour change I want to flag rather than leave for you to find. A headless prompt refusal is marked too, so it now carries a category, and the stop answer says the tool was refused rather than that it failed with the same error. The bound is unchanged. I think the new wording is the correct one, since the tool never ran, but it is a user-visible string and it was previously asserted the other way, so it is your call if you disagree. The test that pinned the old wording is updated along with the comment explaining why it was uncategorized.

jatmn (changes requested):

## Overall root-cause guidance

These are not five unrelated defects. The repeated review rounds are exposing the same three boundary problems from different directions:

1. Refusal provenance is created correctly at some producers, translated again at multiple conversion seams, collapsed into the same string used for tool-controlled output, and then reconstructed from prefixes and booleans. That makes classification, streak identity, aggregate history, and wording capable of disagreeing even when each local helper looks correct.
2. The new guard outcomes return before the headless completion contract is applied. The counter knows why the run stopped, but that fact is not carried far enough to determine terminal status consistently.
3. Several regressions manufacture the downstream shape they expect instead of invoking the producer that is supposed to create it. Those tests prove that consumers handle hand-authored metadata; they do not prove the registry or real tool emits that metadata, that conversion preserves it, or that the tool body is skipped. That is why reverting the production fix can leave the regression green.

Please address this as one end-to-end failure/refusal contract rather than patching the individual messages. The important lifecycle is:

`policy/configuration producer -> tools.Result provenance -> ToolResult conversion -> per-tool guard identity and aggregate state -> typed stop outcome -> user wording and headless terminal status`

A root-cause fix should preserve the following invariants across that whole chain:

- A refusal is identified from structured provenance only, never model-visible output.
- The guard's identity is a tagged value such as `(execution failure, error signature)` or `(policy refusal, category)`, not two domains encoded into one string. The exact type is up to you; the requirement is that arbitrary tool output cannot collide with or impersonate a refusal.
- Same-identity count, total consecutive failure count, and aggregate provenance are distinct facts. A changed identity resets only the same-identity streak; success of that tool resets both counters; unrelated-tool success keeps the existing per-tool behavior. The aggregate state must retain enough information to distinguish all-executed, all-refused, and mixed sequences when the 12-call bound fires.
- The guard should return a semantic stop cause rather than requiring callers to reconstruct meaning from overlapping `Varied`/`Refused` booleans. Names are illustrative, but outcomes such as same execution error, same refusal category, varied executed failures, and varied all-refusal failures need unambiguous behavior. Define mixed refusal/execution wording deliberately rather than letting field precedence decide it accidentally.
- Every non-completion outcome introduced by this PR must flow through the headless completion-status decision before returning. Preserve the pre-existing same-error and interactive behavior; the required change is that the new refusal/content-blind stops cannot become successful automation results.
- Tests should create provenance only by invoking the real producer. Consumer unit tests remain useful, but pair them with Run-level tests that cross registry/tool production, conversion, counting, stop wording, and terminal status. A useful falsifiability check is to remove each producer marker in turn and confirm its regression fails for the production reason.

I suggest fixing in that order: define the typed identity/aggregate contract first, make both result-conversion seams populate it, derive stop cause and terminal status from it, then replace the shape-only tests with a producer-to-consumer matrix. At minimum, that matrix should cover permission required/denied, sandbox deny/approval-required, missing artifact directory, configured directory with the selected driver disabled, malformed capture arguments, an executed error containing refusal-like text (including the current `denial:` sentinel), alternating refusal categories, mixed executed/refused failures, same-tool success reset, and unrelated-tool success isolation. This should close the class instead of moving the next inconsistency one layer downstream.

## Findings

- [P1] Mark the new guard outcomes incomplete in headless runs
  `internal/agent/loop.go:769`
  When this branch stops a refusal streak or the new content-blind varied-error streak, it returns an “Agent stopped” answer without setting `Result.Incomplete`, even when `RequireCompletionSignal` is enabled. Those paths previously continued to the max-turn branch, which sets `Incomplete` and an incomplete reason. `zero exec` enables the gate by default and treats only `Incomplete` as exit 4; otherwise it emits `run_end("success", 0)`. A task that was denied six times or failed through the 12-call bound therefore becomes a successful automation result despite doing no requested work.

  The root cause is that the guard branch constructs user-facing text and returns directly instead of carrying its non-completion cause through the terminal-status boundary. Fix this where the new stop outcome is finalized—not by teaching the CLI to parse “Agent stopped” text. Preserve interactive behavior and the pre-existing same-error guard status, but make the PR's new refusal/content-blind outcomes set an explicit unfinished status under the completion gate. Add Run and CLI-output regressions for plain, JSON, and stream-JSON terminal status so the final text, `Incomplete`, error event, `run_end`, and exit code agree.

- [P2] Keep denial identity out of the tool-output signature namespace
  `internal/agent/guardrails.go:548`
  `errSig` stores both normalized tool-controlled output and the synthetic `denial:<category>` key, and line 572 later reconstructs provenance with `HasPrefix`. An executed failure whose output is exactly `denial:permission_denied` therefore has the same identity as a real permission refusal. Three executed failures followed by three refusals combine into one six-count streak, stop earlier than either sequence independently should, suppress the executed-error interpretation, and report all six calls as refused. Even without an exact category collision, six executed failures whose output starts with `denial:` are reported as refusals.

  This is the same design class as classifying arbitrary stderr by refusal phrases: trusted provenance and untrusted content occupy one namespace. Do not solve it by choosing a less likely prefix or escaping particular outputs. Keep failure kind/category as typed state separate from the normalized error signature, compare a tagged identity in the same-signature counter, and carry the kind into the outcome without reconstructing it from `errSig`. Add regressions for an executed error equal to each synthetic category spelling, an executed error merely beginning with the prefix, and a mixed executed/refusal sequence; none may merge or acquire refusal wording.

- [P3] Preserve refusal provenance at the content-blind bound
  `internal/agent/guardrails.go:574`
  The 12-call branch records only `Varied`; the record retains the current identity and total count but not the provenance of the accumulated sequence. If one tool alternates `permission_denied` and `sandbox_block`, neither category reaches the six-call same-identity bound. The twelfth call reaches the content-blind branch with `Refused=false`, and `toolFailureStopAnswer` says the tool “failed ... with varying errors,” even though policy prevented all twelve executions.

  The root cause is deriving the stop description from the last/current record shape rather than aggregate facts about the sequence that tripped the bound. Track enough aggregate kind information to distinguish at least all-refused from all-executed; decide mixed-sequence semantics explicitly. Then derive both wording and terminal behavior from that typed stop cause. Preserve the accepted 6/12 thresholds and reset/isolation rules. Add an alternating permission/sandbox refusal test that proves the tool never runs and the 12-call answer remains a refusal, plus a mixed sequence test that pins the intended wording rather than inheriting boolean-switch precedence.

- [P2] Exercise the real sandbox refusal boundary
  `internal/agent/policy_refusal_run_path_test.go:135`
  This purported sandbox-preflight regression registers an allow-permission fake, executes its `Run` body twelve times, and has that body manufacture an ordinary unmarked `StatusError`. The assertions then expect `tool.ran == 12` and the generic varied-error answer. It never calls `Sandbox.Evaluate`, never exercises `refusalResult(..., PolicyRefusalSandboxDenied)`, never proves the body was skipped, and never verifies `denialCategoryForResult` at either conversion seam. Reverting the new sandbox deny/prompt markers to plain errors leaves this test, the classifier fixtures, and the existing actual-engine blocking test green because each covers a different isolated fragment.

  The root cause is a test double placed after the boundary under test. Replace or supplement it with a hermetic Run-level setup that uses the actual registry sandbox evaluation and a tool whose body records any execution. Drive both a deny and an ungranted prompt where practical; assert `ran == 0`, the exact marker/category after conversion, the six-call category-keyed halt rather than the 12-call generic halt, no schema hint, no profile escalation, and refusal wording. The test should fail if the producer marker, conversion mapping, policy classification, or guard wiring is independently removed.

- [P2] Cover the configured-root disabled-driver refusal
  `internal/agent/capture_artifact_refusal_test.go:23`
  The only real `capture_artifact` case in this file constructs empty options. `RejectBeforePermission` therefore returns at the earlier missing-artifact-directory branch and never evaluates `actionEnabled`. The Run-level doubles hardcode `PolicyRefusalToolNotEnabled`, so they prove downstream handling only. Reverting just `local_capture.go:95` to `errorResult` leaves every added capture test and the existing disabled-driver test green; the latter asserts only text/status. In production, a configured artifact root with one enabled driver and a different selected driver disabled would again be classified as retriable, receive a schema hint, become eligible for profile escalation, and use error-signature/generic accounting instead of the stable refusal category.

  The root cause is duplicating the intended marker in a fake instead of reaching the real configuration branch. Construct the real tool with `ArtifactsDir: t.TempDir()`, enable one driver, and request an action owned by a disabled driver so the test cannot exit through the missing-directory case. Drive that result through the same pre-permission conversion and Run consumer used in production. Assert `tool_not_enabled` provenance, the mapped denial identity, non-retriable/no-hint/no-escalation behavior, and the category-keyed halt. Keep separate cases proving missing-directory refusal and malformed-argument retry behavior so fixing this branch cannot flatten all early rejections into one policy result.

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

Vasanthdev2004:
Done in f7f7871e. You were right about all five, and right that they are one shape rather than five defects. I reproduced each before changing anything, and each fix falsifies.

**Typed identity.** The streak now keys on a `(kind, key)` pair, `failureKindExecuted` with the error signature or `failureKindRefused` with the category, so no tool output can spell a refusal. Before the change:

```
the tool EXECUTED 6 times (nothing refused it)
final answer = "Agent stopped: the `bash` tool was refused 6 times in a row, ..."
```

A command printing exactly `denial:permission_denied` and exiting non-zero. After, it reads as the same error, and putting the single string namespace back fails the regression naming the executed count.

**Aggregate provenance.** The record carries `sawExecuted` and `sawRefused`, and the guard returns a typed `toolFailureCause` instead of overlapping `Varied`/`Refused`. Alternating two categories went from "the tool failed 12 times in a row with varying errors" to "was refused 12 times in a row for different reasons". Mixed is decided rather than inherited: "failed or was refused", pinned by its own case so a future switch reordering cannot change it silently.

**Terminal status.** The halt sets `Incomplete` and a reason under `RequireCompletionSignal`, and leaves the interactive path alone. The regression asserts both directions, so setting it unconditionally fails too.

**The two coverage gaps: you were right, and I checked rather than took it on trust.** Reverting `refusalResult(..., PolicyRefusalSandboxDenied)` in `registry.go` to a plain `errorResult` left `internal/agent` AND `internal/tools` fully green. Same for the disabled-driver branch in `local_capture.go`. Nothing depended on either marker.

The sandbox one now drives a real `sandbox.Engine` evaluation through `Run` with a `DenyWrite` policy, asserting `ran == 0`, the halt bound, and the wording. It fails with the marker reverted. One honest note on it: `errorSignature` truncates at 80 characters and the block message opens with a fixed 40-character prefix followed by a long temp root, so the per-call paths collide in the signature anyway. The request count pins which bound halted the run, but the wording assertion is what actually discriminates. I said so in the test rather than let it read stronger than it is.

The capture one was uncovered exactly as you described, and finding out why was instructive: argument validation runs BEFORE `RejectBeforePermission`, so my first attempt returned "session is required" and never reached the branch either. It now configures a real artifact root with the browser driver enabled and asks for a terminal action, and there is a sibling case pinning that a malformed argument stays retriable, so a later change cannot flatten the early rejections into one policy result.

**One thing I have not done.** You asked for CLI-output regressions on plain, JSON and stream-JSON terminal status. There is no test in `internal/cli` referencing `Incomplete` at all today, and no harness that drives `runExec` end to end with a fake provider, so that is a new harness rather than a new test. The mapping itself is pre-existing and this PR does not touch it; what changed is that a guard halt now reaches it. I would rather build that harness as its own change than bolt it onto this one, but say the word if you want it here and I will.

`gofmt` clean, `go vet` clean for linux, darwin and windows, `internal/agent` and `internal/tools` green.

jatmn (changes requested):

## Overall guidance

The recurring risk in this area is the boundary between parallel execution and ordered result consumption. `executeParallelReadBatch` executes and waits for an entire eligible batch up front, but the terminal branches in the consumption loop infer whether later calls ran from their position after the current index. Once read-ahead has happened, “not consumed yet” no longer means “not executed,” so adding a new early-stop condition can expose lost results and contradictory bookkeeping even when the guard itself is correct.

Please address that lifecycle mismatch as the root cause rather than special-casing only this threshold. At a terminal decision, each advertised tool call needs to be in one explicit, truthful state:

- unstarted, in which case an aborted placeholder is appropriate;
- completed, in which case its real result and associated accounting must be finalized exactly once; or
- currently running/cancelled, in which case its actual terminal outcome must be represented.

The repair can prevent read-ahead across a call whose result may trip a terminal guard, drain already-completed siblings without resuming the model or changing the selected stop outcome, or use another explicit batch-state design. Whichever approach you choose, please audit the other early-return paths that can run while `precomputed` results exist so this execution-state assumption is fixed in one place rather than resurfacing for the next stop condition. Preserve parallelism for safe reads, sequential behavior for mutating tools, the accepted guard thresholds, and strict one-result-per-tool-call provider replay.

## Findings

- [P2] Account for parallel reads that already finished before the new halt
  `internal/agent/guardrails.go:644`
  The new content-blind stop can fire while `loop.go` is consuming a precomputed read batch. For example, after eleven varying failures for one read-only, thread-safe tool, the model can issue two independent calls to that tool in the next turn. `executeParallelReadBatch` executes and waits for both calls before the loop observes either result. Consuming the first result increments `anyErrorCount` to twelve here and selects `outcome.Stop`; the stop branch then calls `appendAbortedToolResults` for every later advertised call on the assumption that those calls “never run.” The second call has already run, so the persisted transcript says it was aborted while its real output is discarded. Its `OnToolCall`/`OnToolResult` callbacks, result-level trace counter, task observation, image delivery, and model-visible message are also skipped. If that sibling is a successful `read_file`, execution may already have committed file-observation credit even though the corresponding content is absent from the transcript, leaving internal authorization state inconsistent with what the model actually saw.

  Please fix the execution-state boundary: either do not execute siblings that can fall beyond this terminal observation, or finalize every already-completed sibling with its real result and bookkeeping before returning, without allowing those results to reverse the stop decision or start another model turn. Add a regression that begins at eleven varying failures, returns two parallel-eligible read calls, and proves that each invocation is represented exactly once with the correct real-versus-aborted status. The test should also cover callback/event counts, strict tool-call/result pairing, task/trace accounting, image handling where applicable, and file-observation state so a narrow transcript-only patch cannot leave the same inconsistency elsewhere.

Vasanthdev2004:
Fixed in 54b27552, and you were right about the root cause rather than the threshold.

I reproduced it before changing anything. Eleven varying failures, then a turn advertising two parallel-eligible calls to the same read-only tool:

```
tool executions      = 13
OnToolResult calls   = 12
>>> 1 invocation executed but was never reported as a result
```

The transcript recorded an aborted placeholder for work that had already run.

**Fixed at the lifecycle boundary, not at the new guard.** Every remaining advertised call now lands in the one state that is true of it: completed, and finalized exactly once with the same bookkeeping the main path performs, or unstarted, and aborted so every `tool_use` keeps its answering `tool_result`. The drained sibling is deliberately not fed to the guard, since it cannot reverse a decision already made and is only owed an honest record.

The part I want to flag, because it is your "fix it in one place" point: **there were three early returns making the same assumption**, not one. The abort path, the stop-reason path and the new guard halt all closed out `ToolCalls[index+1:]` as if nothing after the cursor had run. All three go through one helper now, so the next terminal condition inherits the fix rather than the bug.

The regression asserts what you asked for rather than just the transcript: executions equal reported results, `OnToolCall` and `OnToolResult` agree, exactly one tool result per advertised call id, the drained sibling carries its real output rather than the placeholder, and the stop answer is unchanged so draining cannot reverse the halt. A sibling test pins that a genuinely unstarted call still gets a placeholder, so the fix did not simply stop aborting anything.

Falsifying it took two goes, which is worth recording. My first mutation removed the helper's call sites and the build failed, so the test proved nothing; a passing suite there would have been meaningless. Mutating the lookup instead, so it still compiles and always reports "did not run", fails it properly and names all three symptoms:

```
the tool executed 13 times but 12 results were reported; a completed call was recorded as aborted
the sibling that already ran was recorded as aborted
the sibling's real output is missing: "aborted: run halted by the repeated-failure guard"
```

`gofmt` clean, `go vet` clean for linux, darwin and windows, `internal/agent` green under `-race`.

One thing I did not do. You asked the regression to cover file-observation state as well. It does not: the probe tool is a synthetic read that commits no observation credit, so asserting on it there would be asserting on my own fixture. The inconsistency you describe is real and the fix addresses it at the source, since the sibling's result now reaches the transcript by the same path as any other, but if you want that pinned specifically it wants a real `read_file` against a tracker and I would rather add it as its own case than pretend the current one covers it.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto current main before merge
  `internal/agent/loop.go`
  The branch merge base is `ad34dc8d`, while live main is `27b319ca` and has advanced through eight commits, including changes to the same agent, guardrail, loop, and tool surfaces. GitHub is currently BLOCKED despite reporting the head mergeable. Please rebase and revalidate the resolved diff against current main.

## Review guidance

This PR has accumulated follow-up findings because it changes a cross-layer control-flow contract rather than an isolated counter. A single tool result is interpreted at several boundaries: a tool or registry produces it; the agent converts it into a `ToolResult`; retry, posture, and guardrail logic derive behavior from it; terminal paths serialize it into the transcript and callbacks; and headless callers consume the final status. Fixing one observation point without establishing one authoritative fact leaves nearby consumers able to disagree.

Before another update, please treat the affected paths as two small end-to-end contracts and validate them at their boundaries rather than adding a local special case:

1. **Refusal provenance.** Define one registry-owned fact for “the registry prevented execution on policy grounds.” It must be impossible for an executed tool result—whether built in-tree, by an adapter, or by a future implementation—to set that fact accidentally or deliberately. Derive retry eligibility, posture treatment, streak identity, stop wording, and headless status from the same fact. Preserve the inverse: an executed failure remains an executed failure regardless of output or ordinary metadata. Test each producer (permission, sandbox, filter, early configuration refusal), then test the loop behavior through `Run`, including deliberately lookalike executed results.

2. **Advertised-call lifecycle.** For every terminal path, distinguish calls that never started from calls that have produced a result, even when that result is accompanied by an error directing the enclosing run to stop. Finalize completed calls through one shared path for transcript entries, callbacks, trace counters, task state, loaded tools, and images; emit placeholders only for genuinely unstarted calls. Test a parallel batch where terminal state is selected before every precomputed result is consumed, including cancellation, ordinary error, and successful sibling cases.

The practical check is not merely that the newly added unit test passes. For each claimed invariant, mutate the exact producer or bridge that supplies the fact and verify a `Run`-level regression fails. Also review every early return after tool execution/precomputation against the same lifecycle helper. This avoids repeated review cycles caused by tests pinning a helper while the production conversion, terminal path, or sibling consumer still uses a different definition.


## Findings

- [P2] Authenticate policy-refusal provenance at the registry boundary
  `internal/tools/types.go:99`
  This PR correctly stops inferring a refusal from tool output, but replaces that text-based identity with an unauthenticated metadata key. `Tool.Run` returns a `tools.Result`, and `Registry.RunWithOptions` forwards an executed result and its `Meta` unchanged. `IsPolicyRefusalResult` then treats any nonempty `Meta["policy_refusal"]` as proof that the registry refused the call before execution.

  Consequently, a tool that actually ran and failed can return that key (whether with a recognized value such as `sandbox_denied` or an unknown nonempty value) and enter the policy-refusal path. The loop withholds its retry hint, suppresses the profile failure-streak recovery, and includes the result in refusal-oriented guard accounting; recognized values can also make the final answer say the tool was refused although it executed. This violates the new provenance contract and recreates the classification trust problem one layer below output.

  Please make pre-execution refusal provenance unforgeable by an executed tool result: keep the fact in registry-owned state or strip/reserve the marker at the execution boundary, then derive both classification and streak identity from that trusted fact. Preserve ordinary result metadata and the existing real registry refusal paths. Add regression coverage for an allowed tool that executes and fails while returning both a recognized and an unknown `policy_refusal` value, verifying that it stays an executed retriable failure.

- [P2] Finalize result-plus-cancellation entries when draining a parallel batch
  `internal/agent/loop.go:3614`
  The new terminal closeout correctly fixes the common case where a read-ahead sibling has already completed, but it treats every precomputed entry with a non-nil `abortErr` as unstarted. A canceled permission request is different: `executeToolCall` first creates `canceledPermissionResult` (with its call ID, error output, and cancellation/permission information) and returns it together with `ErrPermissionApprovalCanceled`; `executeParallelReadBatch` stores both fields.

  If an earlier precomputed sibling triggers a terminal guard or stop path, `closeOutRemaining` reaches that canceled sibling through `precomputedResultFor`. Its `abortErr != nil` check discards the populated result and emits an aborted placeholder. The transcript then denies that the call completed permission handling, while the real cancellation result is omitted from `OnToolResult`, trace/output accounting, and task observation despite the permission event having occurred.

  Please model precomputed completion separately from whether it asks the enclosing run to return an error. Drain any entry that has a real result through the same finalization path as other completed siblings, and reserve the aborted placeholder for entries that never produced a result. Keep the terminal decision unchanged: draining a canceled sibling should make the recorded lifecycle truthful, not let it override the already selected stop/abort outcome. Add a batch regression with a populated cancellation result after an earlier terminal sibling and assert exact tool-result pairing plus callback, trace, and task-observation preservation.

Vasanthdev2004:
Both in, and the branch is on current main.

**Forgeable refusal provenance.** You are right that this moved the trust problem down a layer instead of closing it. `RunWithOptions` now strips `policy_refusal` from every result that came back from a tool actually running, so there is no value a tool can return that survives execution, recognized or invented. Real pre-execution refusals are untouched, `RejectBeforePermission` included, since it decides before any of that. Ordinary metadata is preserved and the tool's own map is not mutated. Covered at the boundary across all three execution call sites (plain `Run`, `RunWithSandbox`, `RunWithOptions`) and again through `Run`, where a forged marker has to leave the retry hint injected and `DenialReason` empty. Turning the strip into a pass-through fails both.

**Result plus cancellation when draining.** Fixed the way you describe. The batch records whether an entry produced a result, at the point where both halves are in hand, and the aborted placeholder is reserved for entries that produced nothing. The terminal decision is unchanged.

One correction on that one: I could not reach it. A cancelled permission inside a batch needs `shouldRequestPermission` to be true, and for a `PermissionAllow` tool the sandbox short-circuits to allow ("tool safety allows execution") before it can prompt, while `parallelSafeToolCall` admits only `PermissionAllow`. The other two cancel producers sit behind `isShellCommandTool`, and a shell tool is never read-only, so it never enters a batch. So the entry `precomputedResultFor` was discarding cannot exist today. What it was is a check keyed on the wrong fact, one gate change away from being real, which is worth fixing on its own. I changed the shape and the regression pins the invariant, but I did not want to claim a live data-loss bug I could not produce. Tell me if you can reach it from an angle I missed.

jatmn (approved):
LGTM

Vasanthdev2004:
@anandh8x both of these were right, and both are fixed on the current head (19c839d3). The branch was rebased since you looked, so your review is anchored to 4641b18a which is no longer one of the commits here. That is why it is still sitting as blocking.

Counting. The failure flag is now `countedFailure := retriableFailure || policyRefusal`, and `hintable` stays plain `retriableFailure`, so denials are counted but still draw no schema hint. One deliberate difference from your wording: I used `isPolicyRefusal` rather than a literal `DenialReason != ""`, because a headless prompt refusal and a sandbox preflight denial on a non-shell tool both arrive with an empty `DenialReason`, and the literal form would have left exactly those two uncounted. It works out to every `StatusError` result being counted.

Count and wording. The outcome now carries `anyErrorCount` when the content blind bound is what trips, so twelve distinct errors report twelve rather than one, and the stop answer picks its wording from the cause instead of always claiming the same error.

Run level coverage is in a74d2b22. Five tests drive `Run` rather than the guard helper, including the varying denial case you described.

Two things still open that you may want to look at while you are in here, both of which I would rather fix than have you find again:

`TestRunStopsAnUncategorizedVaryingSandboxRefusalAtTheVariedBound` is misnamed. Its fixture is `PermissionAllow` with no metadata, and since the classifier is structured only now, that result is an ordinary executed failure, not a refusal. The assertion is right for the varied executed bound, the name is wrong. The same stale claim is in the doc comment above `isPolicyRefusal`, which still lists the bare `Sandbox block` case as a gap it closes.

Happy to push both if you want them in before you re-review, but that will dismiss the existing approval, so say the word rather than me doing it under you.

Vasanthdev2004:
@anandh8x your review is on `4641b18a` and the branch is 12 commits past it, so I re-checked both points at `19c839d3`. Both are fixed, driven rather than read.

**1. Denials counting toward the halt.** `loop.go:806-808` now computes `countedFailure := retriableFailure || policyRefusal` and passes that as the halt flag while `retriableFailure` stays the hint flag, so categorized denials stop the loop without turning on schema hints. `DenialReason` is threaded through as well.

**2. The reported count.** `guardrails.go:649` returns `record.anyErrorCount` when that is the counter that tripped, so the final answer no longer says the tool failed once.

```
--- PASS: TestPermissionDenialStreakSurvivesVaryingReasonText
--- PASS: TestRunStopsARepeatedlyDeniedToolAtTheFailureBound
--- PASS: TestToolFailingWithDifferentErrorsEveryTimeStillStops
--- PASS: TestVariedFailureStopAnswerReportsTheRightCounter
```

Both are Run-level, so they cover the production call path you asked for, not the guard helper. Falsified one at a time. Passing only `retriableFailure` again:

```
guardrails_test.go:380: the run made 10 denied calls, want it halted at 6
```

Returning `record.count` again:

```
guardrails_test.go:411: Count = 1, want the counter that tripped (12)
```

Ready for another look.

Vasanthdev2004:
@anandh8x whenever you have time, this is ready for another look.

Your review is on `4641b18a` and the head is `19c839d3`. Both points were addressed and I posted the detail on 4 September: the halt now counts categorized denials without turning on schema hints, and the reported count comes from the counter that actually tripped. jatmn has approved at the head since.

Nothing has changed on the branch since that reply, so there is nothing new to read beyond it.

gnanam1990 (changes requested):
I found no new code-level blocker on exact head `19c839d3264edadde8985db755ed2577e3de1e92`, but the branch must be refreshed before it is merge-ready.

## Merge readiness

- **Rebase onto current `main`.** The head still has merge base `27b319ca88a3180bed5183f0c599e9307f3ece12`; live `main` is `6937a309cf00825572210a7610a1f3ea8b74c2f9`. The branch is 31 mainline commits behind. GitHub reports it mergeable but `BLOCKED` / `CHANGES_REQUESTED`; its passing checks do not validate the resolved change against today's target. Rebase, preserve the registry-owned refusal provenance and completed-parallel-result accounting, rerun the full required gates, and request a fresh review on the rebased exact head.

## Current-head review

- jatmn approved this exact head on 2026-08-27 after the prior provenance, headless status, pre-execution refusal, and parallel read-ahead findings were closed.
- All recorded CI, CodeQL, smoke, Zero Review, and CodeRabbit statuses on this head are green.
- I rechecked the two mechanically open CodeRabbit threads. `alwaysPromptingTool` has one declaration, so the duplicate-declaration report is closed in code. The requested real headless and sandbox refusal coverage exists in `policy_refusal_run_path_test.go`; the older helper-only test is no longer the only proof. Neither thread is a current blocker.
- No third-party module, external service integration, or dependency change is introduced; the diff stays in Zero's internal agent/tool refusal and control-flow surfaces.

**Verdict: Changes requested for stale-base rebase and fresh exact-head validation only.**

This pass checked current review-thread closure and merge readiness; it is not a claim that I reran the full suite or validated a rebased result locally. No author-branch changes were made.

Vasanthdev2004:
Rebased onto main at 69900830 as asked. Clean replay of the twelve commits, nothing resolved by hand; the registry-owned refusal provenance and the completed-parallel-result accounting are unchanged. agent and tools packages green here, linux and darwin cross-builds pass. @gnanam1990 this is the only change since your review. @jatmn your approval at 19c839d3 will have been dismissed by the push; the content is identical, one more look when you have a moment. @anandh8x your review at 4641b18a is well behind this.

Vasanthdev2004:
Both of your asks are done, so this one is only waiting on a fresh look.

@anandh8x both of your items are fixed. The guard failure flag no longer passes `isRetriableToolError` for both purposes: a categorized denial is counted by the streaks while schema-hint injection stays off for it, which is the split you asked for, and the reason is written down next to the code so it does not regress quietly. For the count, `anyErrorCount` counts consecutive failures of a tool regardless of the error text and is cleared only by a success, so twelve distinct errors no longer report as one, and the stop message is driven from that rather than from the signature-specific counter. There are Run-level regressions rather than only guard-helper ones.

@gnanam1990 the rebase you asked for is done. Head `69900830` sits on current main's line, two commits behind `c1937dfa` with no conflicts, and CI 12 of 12 ran against that head rather than the stale pair. jatmn's approval carried across the rebase and is live at this head. The registry-owned refusal provenance and the completed-parallel-result accounting are both preserved; the failure-kind separation you would want to check is in `guardrails.go`, where tool-produced and policy-produced failures no longer share a string namespace.

jatmn (approved):
LGTM

gnanam1990 (approved):
## Verdict

**Approve** — and this withdraws my `CHANGES_REQUESTED` from `19c839d3`.

Reviewed at `69900830`, merged against `99721c76` (16 behind).

## Withdrawing my previous block, and correcting its basis

My 2026-09-12 review said *"I found no new code-level blocker"* and requested changes purely because the branch was 31 commits behind `main`. It is now 16 behind, so the objection is smaller but not automatically gone — a stale branch can conflict semantically while merging textually, which is the thing worth checking rather than asserting.

So I checked it instead of restating it: merged `origin/main` into `69900830` and ran the result. `go build ./...`, `go vet ./...` and the **full `go test ./...`** are all clean on the merged tree, including the files `main` added underneath this branch (`internal/tools/rooted_file.go`, `internal/tui/terminal_attach.go`, and the new `file_commit_test.go`). There is no semantic conflict, so "16 behind" is a process preference here, not a defect, and I do not think it should keep blocking a branch jatmn has approved.

Four mutations against `internal/agent/guardrails.go`, each reverted in turn with `internal/agent` rerun — **all four caught**:

1. **Denial keyed on category, not prose** (`:605-613`) — removing the `DenialNone` branch so the identity falls back to `errorSignature(output)` is caught. This is the actual defect: a refusal message embeds the path or command it refused, so it differs every call while describing one unchanging refusal, rebuilding the streak at 1 forever under a halt of 6.
2. **The content-blind backstop increments** (`record.anyErrorCount++`, `:628`) — neutering it is caught.
3. **A changed failure restarts the same-identity streak but not the blind one** — adding `record.anyErrorCount = 0` to the identity-change branch is caught. This is what makes the design hold: `DenialCategory` does not need to be exhaustive, because an uncategorised or varied failure still reaches `toolFailureAnyErrorStopAt = 12`.
4. **Success resets both counters** — dropping the `delete` is caught.

That third property is why I think this supersedes #702's per-message id-invariance approach correctly: the enum being incomplete is survivable by construction rather than by keeping a list current.

## Not covered

macOS only; I did not rerun the release build/smoke on the merged tree, and I have not re-read the diff outside `internal/agent` since my earlier pass — the mutation evidence above is specific to the failure-bound logic.

Vasanthdev2004:
@coderabbitai both open threads are already answered on `69900830`, so there is nothing to push:

- **Duplicate `alwaysPromptingTool`.** There is one declaration now, at `guardrails_test.go:312`, and the package builds in CI.
- **Run-level regression for uncategorized refusals.** `policy_refusal_run_path_test.go` drives both kinds through `Run`. `TestRunStopsAnUncategorizedHeadlessRefusalAtTheFailureBound` covers a headless prompt refusal, and `TestRunStopsAnUncategorizedVaryingSandboxRefusalAtTheVariedBound` covers a sandbox refusal whose text changes on every call, so only the per-tool counter can stop it. Both also check that the failure-streak escalation never fires.

PR #867, opened 2026-08-04
Improve source read efficiency and model compatibility

anandh8x:
## What changed

- add bounded compact source reads with safe Go fragment parsing and conservative fallback behavior
- standardize the model-facing `read_file` contract on `path`, `offset`, and `limit`
- retain legacy line and byte arguments internally, with mixed legacy calls preferring the line range instead of failing
- expose one case-sensitivity argument for `grep`
- make compact source reading available up front and clarify when exact reads and plans are appropriate
- add regression coverage for canonical ranges, mixed legacy arguments, schema exposure, and Go fragments

## Why

The previous `read_file` schema exposed overlapping line and byte pagination systems. Some models combined both systems in one request, producing deterministic argument failures and repeated retries until the loop guard stopped the run. The canonical contract removes that ambiguity while preserving compatibility for existing callers.

## GPT-5.5 benchmark

Identical read-only inspection prompt against the Zero source tree:

| Build | Time | Tokens | Result |
| --- | ---: | ---: | --- |
| Before | 40.3s | 10,859 | stopped after six invalid `read_file` calls |
| After | 19.4s | 13,219 | completed with one search, two successful reads, and no plan calls |

The pre-change token total is lower only because the task halted without producing the requested answer; the primary measured improvement is successful completion without the invalid-call loop.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...` with an isolated empty user config root
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`

## Summary by CodeRabbit

* **New Features**
  * Added `offset`/`limit` line-range selection for standard and minified source reading.
  * Improved Go minification for standalone and context-aware fragments, including safer handling of partial code.
* **Bug Fixes**
  * Updated truncation and out-of-range guidance with clearer range terminology.
  * Refined tool option documentation and behavior for more consistent usage.
* **Documentation**
  * Clarified when planning is needed and standardized source-exploration guidance.
* **Tests**
  * Expanded coverage for ranged reads, minification edge cases, and validation scenarios.

### DIFF
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -30,11 +30,13 @@ work.
    read-before-edit discipline: inspect the target file and nearby callers,
    tests, or config before you modify behavior. Never edit a file you have not
    read.
-2. **Plan.** For multi-step work, call update_plan with an ordered checklist and
-   keep it current at meaningful milestones. Mark completed work and the next
-   in-progress step together when practical; do not spend separate model turns
-   updating the plan after every file or command. Keep at most one item
-   in_progress. Skip the plan for trivial one-step tasks.
+2. **Plan.** Use update_plan for implementation or investigation that has at
+   least three meaningful dependent steps, then keep it current at milestones.
+   Mark completed work and the next in-progress step together when practical;
+   do not spend separate turns updating it after every file or command. Keep at
+   most one item in_progress. Skip plans for simple lookups, explanations, code
+   navigation, and other short tasks. Never create a plan after the work is
+   already complete merely to describe what you did.
 3. **Implement.** Make focused changes that match the surrounding code's style,
    naming, and conventions. Prefer the smallest change that fully solves the
    problem. Avoid broad refactors, unrelated rewrites, dependency churn, and
@@ -49,9 +51,13 @@ work.
 ## Editing discipline
 
 - Choose the narrowest tool that safely accomplishes the step. Prefer native
-  file tools - read_file, list_directory, glob, grep, write_file, edit_file,
-  apply_patch - over shelling out to cat/sed/awk/python for file operations.
+  file tools - read_file, read_minified_file, list_directory, glob, grep,
+  write_file, edit_file, apply_patch - over shelling out to
+  cat/sed/awk/python for file operations.
   They are safer, reviewable, and produce clean diffs.
+- Prefer read_minified_file when initially exploring source code; it preserves
+  code while removing comments and redundant whitespace. Use read_file when
+  exact text, comments, or line numbers are needed.
 - Keep edits focused and reviewable. A single patch may update several related
   files when they form one coherent change; do not hide unrelated edits in a
   bulk shell or script rewrite.
--- a/internal/agent/system_prompt_models.go
+++ b/internal/agent/system_prompt_models.go
@@ -71,5 +71,6 @@ const geminiPromptAddendum = `<model_guidance>
   equivalent shell commands; they are safer and produce cleaner diffs.
 - Be concise and concrete. When you run a shell command with side effects, state
   in one short clause why it is needed.
-- Use update_plan for any multi-step task and keep it current.
+- Use update_plan only for work with at least three meaningful dependent steps;
+  skip it for simple lookups and never create one after the work is complete.
 </model_guidance>`
--- a/internal/cli/deferred_wiring_test.go
+++ b/internal/cli/deferred_wiring_test.go
@@ -79,8 +79,8 @@ func TestRegisterToolSearchIfEligibleSkipsWhenThresholdZero(t *testing.T) {
 
 func TestDeferredEligibleCountIncludesOptionalBuiltins(t *testing.T) {
 	registry := newCoreRegistry(t.TempDir())
-	if got := deferredEligibleCount(registry, agent.PermissionModeAuto, nil, nil); got < 3 {
-		t.Fatalf("deferredEligibleCount(core) = %d, want at least 3 optional built-ins", got)
+	if got := deferredEligibleCount(registry, agent.PermissionModeAuto, nil, nil); got < 2 {
+		t.Fatalf("deferredEligibleCount(core) = %d, want at least 2 optional built-ins", got)
 	}
 }
 
--- a/internal/minify/minify.go
+++ b/internal/minify/minify.go
@@ -16,6 +16,7 @@ import (
 	"go/ast"
 	"go/parser"
 	"go/printer"
+	"go/scanner"
 	"go/token"
 	"path/filepath"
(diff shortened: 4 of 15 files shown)


Vasanthdev2004 (changes requested):
Requesting changes: the new `offset`/`limit` pair cannot be used before an edit.

`limit` caps the range [offset, EOF] rather than defining the range, so any interior slice comes back flagged truncated. The registry then skips `RecordSeenRange` and the edit is refused, with a message telling the model to re-read using offset/limit, which is the call that just failed. The `start_line`/`end_line` pair this PR deletes from the schema is the only shape that still works:

```
offset=500 limit=10          truncated=true   reason="limit"   seen=""         edit refused
offset=500 (runs to EOF)     truncated=false                   seen=500-1000   edit ok
start_line=500 end_line=509  truncated=false                   seen=500-509    edit ok
```

1000 line file, driven through `Registry.RunWithOptions`.

Three others, all reproducible:

Ranged Go reads corrupt raw strings. A range that starts inside a raw string literal is parsed as code, so string content is deleted and identifiers are rewritten. `SECRET-MARKER-ONE` comes back as `SECRET - MARKER - ONE`, a line inside the string vanishes, and the header reports a 49% saving. `Fragment`'s own doc says non-Go takes the conservative path because a range may begin inside a multiline string. Go needs the same guard.

Dropping `byte_offset`/`byte_limit` from the schema and from the unseen-file message leaves any file with a line over 128KB unreadable, and so uneditable. The byte path still works, and `file_tracker_largefile_test.go` still asserts it is what makes those files writable again.

Mixing one canonical and one legacy argument silently reads the wrong lines: `{start_line: 5, limit: 2}` returns lines 1 and 2.

Minor: `read_minified_file` with an offset past EOF returns an empty body with status ok, where `read_file` says the offset is past the end.

Vasanthdev2004:
@anandh8x the blocking one is the `offset`/`limit` range, the rest are smaller. `limit` caps [offset, EOF] instead of defining the range, so an interior slice gets flagged truncated, never earns read credit, and the edit after it is refused. A slice that happens to reach EOF passes cleanly, which is why the tests here do not show it.

Two ways out and I would take the first: make `limit` set `endLine` so the read is not flagged truncated at all, or stop treating a limit cut as a byte-budget cut at the point the registry commits the observation. The second one reaches further than this PR needs to.

I have the probe that produced the table in the review, driven through `Registry.RunWithOptions`. Say the word and I will push it as a test on your branch.

anandh8x:
@coderabbitai The latest review found no actionable comments, but the previous changes-requested verdict is still active. Please review the latest commit and clear it if no blockers remain.

gnanam1990 (approved):
## Verdict: Approve

Checked out and exercised. The diagnosis is the strong part: two overlapping pagination systems in one schema is an ambiguity the model resolves by combining them, and the old behaviour answered that with a hard error and a retry loop until the guard stopped the run. Six invalid calls then a dead run is a real cost, and the benchmark showing it complete afterwards is the right evidence for it.

**Compatibility is genuinely preserved, not just claimed.** `byte_offset`/`byte_limit` are still in the schema and still work; the canonical `offset`/`limit` sit alongside rather than replacing them. Conflicting spellings of the same axis (`offset` + `start_line`, `limit` + `end_line`/`max_lines`) still error, which is right — those are the same question asked twice, and picking one would be a guess.

**The mixed line+byte case resolves to lines** (`byteMode := (hasByteOffset || hasByteLimit) && !hasLineRange`), which is the correct choice: the failure being fixed is exactly a model that sent both, and answering something beats failing. `TestReadFileToolMixedLegacyRangesPreferLines` pins it.

`go test ./internal/tools/ ./internal/minify/ ./internal/agent/` green on the branch.

### One inconsistency worth a follow-up, not blocking

This file already establishes the principle, in its own words at `read_file.go:208`:

> The recovery is surfaced in the output (rangeNote) so it is never silent

The `end_line < start_line` recovery follows it and emits a note. The new byte-args-ignored recovery does not — a caller that asked for a byte range gets a line range with nothing saying its byte arguments were dropped. Same class of recovery, one discloses and one does not.

I am not blocking on it for one reason: the output is line-numbered (`2 | beta`), so a byte read and a line read are visually distinguishable, and a model can tell what it got. That makes this a clarity gap rather than a correctness one. Still, a one-line note in the same shape as `rangeNote` would close it and keep the file consistent with the rule it states about itself.

### Unrelated, for the record

`TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` fail on this branch — and I verified they fail identically on pristine `origin/main`, so they are pre-existing and not yours. Worth someone filing separately; they will keep showing up in every reviewer's local run until then.

anandh8x:
@coderabbitai The latest review found no actionable comments and confirmed the previous blocker is fixed. Please clear the stale changes-requested review and approve the latest commit if no issues remain.

Vasanthdev2004 (approved):
Approving. All four are fixed, and I re-ran the same probes that found them rather than reading the diff.

The blocking one, an interior `offset`/`limit` read earning no read-before-edit credit, is gone: on a 1000-line file `offset=500 limit=10` now comes back `truncated=false` with `seen_lines=500-509` and the follow-up edit succeeds. It used to be `truncated=true`, `reason="limit"`, empty `seen_lines`, edit refused.

The other three as well. Mixing argument families no longer reads the wrong lines: `start_line`+`limit` and `offset`+`end_line` both return 5-6 now, where they gave 1-2 and 5-8. Byte mode is back in the schema (`[offset limit byte_offset byte_limit path]`), so a file with a line past the read budget is reachable again. And a ranged Go read starting inside a raw string keeps its content: `SECRET-MARKER-ONE` survives intact instead of arriving as `SECRET - MARKER - ONE`, and the line that looked like a comment is no longer deleted.

`internal/tools` and `internal/minify` both pass, build and vet clean.

One note for whoever merges rather than for you: this branch carries a merge of main from earlier today, so the diff reads larger than the change.

Vasanthdev2004:
@anandh8x approved. All four are fixed and I re-ran the probes that found them rather than reading the diff, so this is verified rather than assumed.

`mergeStateStatus` is CLEAN, and human approvals are @gnanam1990 and me. If the ruleset wants a third human you will need @jatmn, otherwise it is ready to go.

The one I care most about is the interior range: `offset=500 limit=10` now returns `seen_lines=500-509` with the edit succeeding, where before it was flagged truncated and the follow-up edit was refused. That was the whole reason I blocked, and it is properly gone rather than worked around.

Thanks for taking the raw-string one seriously too. That was the quietest of the four and the one most likely to have shipped unnoticed, since it corrupted content while reporting a successful compaction.

Issue #870, opened 2026-08-06
Agent freezes up


msingelwa:
### Version / branch / commit

main

### OS and environment

Windows

### Steps to reproduce

run zero and give it a task and it will freeze, you can't even scroll and kill it immediately. 

### Expected behavior

run zero prompt and runs without getting stuck.

### Actual behavior

<img width="1281" height="891" alt="Image" src="https://github.com/user-attachments/assets/1c578095-8172-4c8b-be19-df8b24fa7a43" />

### Relevant logs, screenshots, or error messages

<img width="1281" height="891" alt="Image" src="https://github.com/user-attachments/assets/290699a1-87c7-4718-a337-edab41cd976a" />

Vasanthdev2004:
Thanks for the report, and sorry you hit this.

The good news first: from your screenshot it is not actually frozen. The turn finished. "Thought for 1.9s" sits above the answer, the whole checklist rendered, and the composer at the bottom is back to its idle `describe a task for zero…` prompt, which only happens once a turn is complete.

What you are hitting is that the answer is taller than the window. The bottom right of your screenshot reads `↓ 59 more · PgDn`, so there are 59 more lines below the fold. PgDn should walk you down them.

Which leaves the real bug, and it is ours: scrolling should have worked and it did not.

Zero runs in the terminal's alternate screen, so your terminal hands its own scrollback and wheel over to us, and we ask it to send us mouse events instead. When a terminal does not support the mouse mode we request, you get the worst of both: our wheel handler never receives anything, and the native scroll is gone too. Nothing moves, and a screen of cut-off text reads exactly like a hang.

Two questions so I can confirm that is what happened:

1. Which terminal are you in? Windows Terminal, the older console window, the VS Code terminal, or something else.
2. Does PgDn move the view?

If PgDn works, this is a mouse-mode problem rather than a freeze. We already fall back to a more compatible mouse mode in one environment where the current one is unreliable, so once you confirm the terminal I will extend that fallback to cover yours. Separately, that "59 more" hint is far too quiet for what it is telling you, and that part is worth fixing regardless of your answer.

msingelwa:
Hi Gitlawb/zero

Thanks for the reply.

The "59 more" l saw and we have the same understand but it was not
responsive(frozen). I totally get the window had more lines but i could not
stop it or command it to stop (tried Ctrl+C, and it took a while to exit).

To answer your questions:

   1. I was using Windows Terminal
   2. PgDn was not working

Thank you

[...]

PR #872, opened 2026-08-06
fix(tui): fall back to CellMotion where AllMotion is unreliable

Vasanthdev2004:
Fixes the scroll failure behind #870.

## The problem

`AllMotion` (1003) buys exactly one thing over `CellMotion` (1002): hover highlighting. Wheel, click and drag all arrive under either. The asymmetry is in the failure: a terminal that does not implement 1003 does not fall back to 1002, it sends nothing at all. Since the TUI holds the alternate screen, the terminal's own scrollback and wheel are gone too, so the user cannot scroll by any means.

That is what #870 reported as a freeze. The screenshot shows a completed turn (`Thought for 1.9s`, answer rendered, composer back to its idle placeholder) with `↓ 59 more · PgDn` in the footer. Nothing was hung. The output ran past the fold and no scroll input worked.

The code already conceded 1003 is fragile by carving out PRoot. The legacy Windows console is the same case and was not covered, so `zero.exe` run outside Windows Terminal asks for a mode its terminal never answers.

## The change

`mouseModeFor` now makes the decision. It trusts terminals that identify themselves, `WT_SESSION` for Windows Terminal and `TERM_PROGRAM` for hosts like VS Code and mintty, and assumes an unidentified Windows terminal is the legacy console. Guessing wrong in that direction costs a hover highlight. Guessing wrong the other way costs every mouse event.

`ZERO_MOUSE_MODE=all|cell|off` overrides it, so anyone on a terminal we guessed wrong about can fix it without waiting for a release, and I can ask a reporter to test one specific mode. An unrecognised value is ignored rather than treated as "off", so a typo cannot silently kill the mouse.

## Tests

`mouse_mode_test.go` covers linux, macOS, PRoot, legacy Windows console, Windows Terminal, VS Code on Windows, all three override values with case and whitespace variants, and an unknown override. Written against the old behaviour first: only `legacy_Windows_console` and the override cases failed, so they pin the gap rather than describe the fix.

`go build ./...`, `GOOS=windows go build ./...`, `go vet` and `gofmt` are clean.

## Notes

The root cause is not confirmed with the reporter yet, since I have not heard back on which terminal they use. I would rather ship the safe default than wait, because the override gives us a one-line diagnostic to hand them either way.

Unrelated, `TestAltScreenTranscriptScrollKeepsFooterFixed` currently fails on pristine `main` at edf660ab on Windows. Not touched by this PR, but somebody should look.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal mouse handling across operating systems and environments.
  * Added compatibility for legacy Windows terminals, Windows Terminal, VS Code, and PRoot environments.
  * Added support for overriding mouse behavior through the `ZERO_MOUSE_MODE` setting.
  * Invalid mouse-mode overrides now safely fall back to automatic detection.
  * Improved reliability when terminal capabilities are inherited or unavailable.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 	"unicode"
@@ -2788,23 +2789,14 @@ func (m model) View() tea.View {
 	// only when the value changes, so gating this costs nothing (§10).
 	view.KeyboardEnhancements.ReportEventTypes = m.dictation.voiceModeEnabled
 	if m.wantsMouseCapture() {
-		if isRunningUnderPRoot() {
-			// Under PRoot the AllMotion (1003) sequence doesn't work
-			// reliably, breaking touch-gesture scrolling. Fall back to
-			// CellMotion which still delivers wheel events, clicks, and
-			// drag — the only thing lost is hover-highlighting.
-			view.MouseMode = tea.MouseModeCellMotion
-		} else {
-			// AllMotion (not CellMotion) is required for hover highlighting:
-			// it reports cursor movement even with no button pressed.
-			// CellMotion only reports motion while a button is held (drag) —
-			// see bubbletea's MouseMode docs. AllMotion has marginally worse
-			// terminal compatibility but is well supported by the terminals
-			// this app targets; the existing 15ms mouse-event throttle
-			// (mouseEventThrottleInterval) already bounds the redraw rate
-			// from the extra motion events.
-			view.MouseMode = tea.MouseModeAllMotion
-		}
+		// AllMotion (1003) is what hover highlighting needs: it reports cursor
+		// movement with no button pressed, where CellMotion (1002) reports motion
+		// only while a button is held. mouseModeFor decides which is safe here,
+		// and documents why one is not simply better than the other: a terminal
+		// that does not implement 1003 sends nothing at all rather than degrading.
+		// The 15ms throttle (mouseEventThrottleInterval) already bounds the redraw
+		// rate from AllMotion's extra motion events.
+		view.MouseMode = mouseModeFor(runtime.GOOS, os.Getenv, isRunningUnderPRoot())
 	}
 	return view
 }
--- a/internal/tui/mouse.go
+++ b/internal/tui/mouse.go
@@ -8,6 +8,57 @@ import (
 	tea "charm.land/bubbletea/v2"
 )
 
+// mouseModeEnv forces a mouse mode: "all", "cell", or "off"/"none". It exists
+// because the failure it guards against is invisible from here and total for the
+// user: a terminal we guessed wrong about delivers no mouse events at all, and
+// there is no in-app way to recover. Anything unrecognised is ignored rather
+// than treated as "off", so a typo cannot silently kill the mouse.
+const mouseModeEnv = "ZERO_MOUSE_MODE"
+
+// mouseModeFor picks the mouse reporting mode, preferring the widest support
+// over the nicest behaviour.
+//
+// The trade is asymmetric. AllMotion (1003) buys exactly one thing, hover
+// highlighting, because wheel, click and drag all arrive under CellMotion
+// (1002). But when a terminal does not implement 1003, it does not degrade to
+// 1002: no mouse event arrives at all. The app holds the alternate screen, so
+// the terminal's own scrollback and wheel are gone too, and the user is left
+// with no way to scroll by mouse. A long answer then looks like a hang rather
+// than like a missing highlight, which is how #870 was reported.
+//
+// So AllMotion is asked for only where it is known to work.
+func mouseModeFor(goos string, env func(string) string, underPRoot bool) tea.MouseMode {
+	switch strings.ToLower(strings.TrimSpace(env(mouseModeEnv))) {
+	case "all":
+		return tea.MouseModeAllMotion
+	case "cell":
+		return tea.MouseModeCellMotion
+	case "off", "none":
+		return tea.MouseModeNone
+	}
+	if underPRoot {
+		return tea.MouseModeCellMotion
+	}
+	if goos == "windows" && !windowsTerminalReportsAllMotion(env) {
+		return tea.MouseModeCellMotion
+	}
+	return tea.MouseModeAllMotion
+}
+
+// windowsTerminalReportsAllMotion reports whether the Windows terminal in use is
+// one that handles 1003.
+//
+// Identified terminals are trusted: Windows Terminal sets WT_SESSION, and hosts
+// like VS Code and mintty set TERM_PROGRAM. The legacy console host sets
+// neither, and it is what a user gets by double-clicking zero.exe or running it
+// from an old cmd window, so an unidentified Windows terminal is assumed to be
+// that one. Guessing wrong in this direction costs a hover highlight; guessing
+// wrong in the other direction costs every mouse event.
+func windowsTerminalReportsAllMotion(env func(string) string) bool {
+	return strings.TrimSpace(env("WT_SESSION")) != "" ||
+		strings.TrimSpace(env("TERM_PROGRAM")) != ""
+}
+
 // parseTracerPid returns true when /proc/self/status contains a non-zero
 // TracerPid. This detects any ptrace tracer (PRoot, gdb, strace, dlv, etc.),
 // not PRoot specifically. That's intentional: under any ptrace tracer the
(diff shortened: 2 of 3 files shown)


Vasanthdev2004:
@anandh8x @gnanam1990 small one, but there is a judgement call in it rather than a correctness question, so I would like a second opinion.

The mechanical part is straightforward. AllMotion (1003) buys only hover highlighting over CellMotion, and a terminal that does not implement it sends nothing at all rather than degrading, which combined with the alt screen leaves the user no way to scroll. That is #870: a completed turn with 59 lines below the fold, reported as a freeze.

The judgement is this. I treat an unidentified Windows terminal as the legacy console and drop it to CellMotion, so a Windows user outside Windows Terminal or VS Code loses hover highlighting. That is deliberate, since the opposite mistake costs every mouse event rather than one highlight, but it is a real behaviour change on Windows and I would rather two of you agreed with the trade than just me.

Worth knowing: the reporter has not told us which terminal they use yet, so the root cause is inferred, not confirmed. `ZERO_MOUSE_MODE=all|cell|off` is in the PR partly so we can settle that with them without shipping a build.

One unrelated thing I tripped over: `TestAltScreenTranscriptScrollKeepsFooterFixed` fails on pristine main at edf660ab on Windows. Nothing to do with this PR, I checked against a clean worktree, but it needs an owner.

jatmn (changes requested):
The mechanical change looks sound, and I agree with the judgement call in your comment. A few follow-ups are worth tracking, but none of them overturn the trade you are asking reviewers to sign off on.

## Findings

- [P3] Inherited `WT_SESSION` / `TERM_PROGRAM` can still defeat the legacy-console fallback
  `internal/tui/mouse.go:57-59`, `internal/tui/mouse.go:42-44`
  You treat unidentified Windows consoles (no `WT_SESSION`, no `TERM_PROGRAM`) as legacy conhost and fall back to CellMotion. That trade is separate from this edge case: on Windows, child processes inherit the parent environment, so a `zero.exe` session attached to legacy conhost can still carry `WT_SESSION` from Windows Terminal or `TERM_PROGRAM` from Git Bash / VS Code even though the active console host is not that parent emulator. In that case `windowsTerminalReportsAllMotion` returns true and AllMotion is requested on a host that may drop every mouse event — the same #870 failure class. This is outside the judgement call you raised, and it is probably not the reporter's path if they double-clicked `zero.exe` or ran from a plain `cmd` window with a clean env. I would still add a regression test (`goos=windows`, inherited `WT_SESSION` or `TERM_PROGRAM`, want CellMotion unless you intentionally trust inheritance). If it shows up in the #870 thread, `ZERO_MOUSE_MODE=cell` is already the right diagnostic. Detecting the actual console host would be nicer than env markers alone, but I would not block merge on it given your diagnostic workflow.

- [P3] Ctrl+E "re-enabled" notice ignores a `ZERO_MOUSE_MODE=off` override
  `internal/tui/model.go:1446-1451`, `internal/tui/mouse.go:36-37`, `internal/tui/model.go:2791-2799`
  With `ZERO_MOUSE_MODE=off` or `none`, `mouseModeFor` always returns `MouseModeNone`, but pressing Ctrl+E again still posts "Mouse interaction re-enabled." even though `View()` keeps emitting `MouseModeNone`. Minor UX mismatch for anyone using the override deliberately; fine to fix in a follow-up.

## On the trade you asked about

You framed the real question correctly: unidentified Windows should default to CellMotion because guessing the other way costs every mouse event, not just hover. I agree with that trade for the likely #870 paths (double-click / plain `cmd.exe`). Losing hover on capable third-party Windows terminals that do not set `WT_SESSION` or `TERM_PROGRAM` (some Alacritty / WezTerm installs) is an acceptable cost, and `ZERO_MOUSE_MODE=all` is a reasonable way to confirm that with the reporter before hardening detection further.

Root cause for #870 is still inferred, as you noted. Shipping the safe default now rather than waiting for the reporter's terminal answer is reasonable; the override gives you a one-line diagnostic either way.

gnanam1990 (dismissed):
## Verdict: Approve

Checked out and probed rather than read. The reasoning that makes this correct is the asymmetry, and the PR states it plainly: AllMotion buys **one** thing (hover highlighting), while a terminal that does not implement 1003 sends **nothing** — and because the app holds the alternate screen, the terminal's own wheel and scrollback are gone too. So the failure is not "no highlight", it is "no way to scroll at all", which reads as a hang. Trading a highlight for that is obviously right.

**Additivity verified directly.** I asserted `mouseModeFor` on darwin/linux/freebsd with no env and no PRoot: all still `AllMotion`. So every platform that worked before is untouched, and the change is confined to Windows-without-a-known-terminal and PRoot — the two cases that were already broken.

**The risk direction is the safe one.** An unidentified Windows terminal is assumed to be the legacy console host, which is exactly what a user gets double-clicking `zero.exe`. Guessing wrong that way costs a highlight; guessing wrong the other way costs every mouse event. Correct default for a case you cannot detect.

**Edge cases I probed, all correct:**
- a typo in `ZERO_MOUSE_MODE` (`celll`, `allmotion`, `true`, `1`, `yes`) is **ignored**, not read as "off" — a mistyped variable cannot silently kill the mouse, which matters for a knob whose failure mode is invisible
- the override wins even where the heuristic would fall back (`windows` + PRoot + `all` → AllMotion), so a user on a terminal you guessed wrong about can recover without editing config
- a blank-but-present `WT_SESSION` (`"   "`) does **not** count as a real terminal — the `TrimSpace` is load-bearing, not decoration

**No per-render cost**: `isRunningUnderPRoot` is a `sync.OnceValue`, so hoisting it into the argument list does not add a `/proc` read per frame, and `os.Getenv` per frame is negligible.

The six-case table (linux, macOS, PRoot, legacy console, Windows Terminal, VS Code) covers the matrix, and the env-override table covers casing and whitespace. `go test ./internal/tui/` green on the branch.

Non-blocking: `ZERO_MOUSE_MODE` is a real escape hatch for a failure the user cannot otherwise diagnose — worth a line in the docs or `zero doctor` output, so someone with a dead mouse can find it without reading source.

Vasanthdev2004:
@jatmn 8adc625d. You are right about the mechanism, and I have not fixed it, deliberately. Here is the reasoning so you can push back if you disagree.

Telling an inherited `WT_SESSION` from a live one needs the console host rather than the environment. Getting that check wrong in the other direction costs every Windows Terminal user their hover highlighting, which is the common case, to protect a case that needs an unusual launch path. That trade looked worse than the one already in the PR.

So the limit is now written down where the decision is made, and `TestInheritedWindowsTerminalEnvStillAsksForAllMotion` pins it, including that `ZERO_MOUSE_MODE=cell` still overrides it, since that is the only recourse an affected user has. If someone lands real host detection the test fails by design and its message says to delete it.

If you would rather I go the other way and default Windows to CellMotion outright, accepting the lost hover everywhere, say so and I will. It is a smaller diff than host detection and it has no false-positive surface at all.

gnanam1990 (approved):
Re-approving after `8adc625d`. My earlier approval was auto-dismissed by the push.

Verified the new commit changes no behavior: `git show 8adc625d -- internal/tui/mouse.go` is comment-only once blank and `//` lines are filtered out. The `+13` is the KNOWN LIMIT block; the executable part is the test.

The limit itself is worth calling out for anyone reading this later, because the commit is choosing to leave a narrow instance of the bug the PR fixes. `WT_SESSION`/`TERM_PROGRAM` are inherited, so they answer for the process that set them, not for the console host attached right now — a shell launched from Windows Terminal that later runs `zero.exe` against the legacy console still carries them, and `mouseModeFor` asks for AllMotion on a host that may drop every event. That is #870 arriving by a narrower door, and the comment says so plainly rather than hoping nobody notices.

Leaving it is the right call on the trade the comment describes: deciding it properly means asking the console host instead of the environment, and erring the other way costs every Windows Terminal user their hover highlighting to cover a case that needs an unusual launch path to reach. `ZERO_MOUSE_MODE=cell` is a real recourse in the meantime.

`TestInheritedWindowsTerminalEnvStillAsksForAllMotion` passes and pins it. Worth being explicit that this test asserts behavior we know is imperfect — if the host-query fix ever lands, this test is the thing to change, not a regression to debug. The comment naming the test makes that discoverable, which is exactly what a pinned limit needs.

`go test ./internal/tui/` green. Approving.

Vasanthdev2004:
@jatmn this has been sitting on your changes-requested since the 6th and I think it is done, so a re-look would clear it.

Both P3s, and one of them I answered differently from what you suggested:

**Inherited `WT_SESSION` / `TERM_PROGRAM`.** You asked for a regression test wanting CellMotion "unless you intentionally trust inheritance". I took the second branch: `TestInheritedWindowsTerminalEnvStillAsksForAllMotion` pins the current behaviour as a documented limit rather than changing it. The reasoning is that the env markers are the only signal available without detecting the console host properly, and flipping inherited markers to CellMotion would cost hover on the common case (a real Windows Terminal session) to protect the rarer one (conhost launched from a shell that set the marker). Naming the limit in a test seemed better than an unprincipled guess in either direction. Say the word if you would rather it defaulted the other way.

**Ctrl+E notice under `ZERO_MOUSE_MODE=off`.** Not fixed. You said follow-up was fine so I left it, but it is about three lines if you would rather it landed here.

Two approvals since your review (CodeRabbit and gnanam1990), CI green.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #873, opened 2026-08-06
Reduce compaction context while preserving task state

anandh8x:
## Summary

- project older conversation history into a bounded semantic transcript before the summarizer call
- omit reconstructible successful tool output while retaining user intent, assistant decisions, tool actions, concise failures, and the recent verbatim tail
- preserve bounded constraints, unresolved failures, permission decisions, artifact references, plans, edits, loaded tools, skills, and project instructions across repeated compaction
- carry the previous prose summary separately so older decisions remain available
- apply the same behavior to automatic context-pressure compaction and manual `/compact`

## Why

Long coding sessions can accumulate large file reads and command output. The previous compaction path sent that entire reconstructible payload to the summarizer, increasing token cost and latency even though the final replacement was already small.

## Controlled benchmark

The refreshed benchmark uses a deterministic 616,309-byte corpus assembled from eight real Zero source files and the same recorded transcript for both measurements. It measures the raw summarizer input against the current semantic projection over five sequential runs. The temporary harness was removed after the run and no report is committed.

| Measurement | Raw input | Projected input |
| --- | ---: | ---: |
| Corpus bytes | 616,309 | 616,309 |
| Summarizer input characters | 617,175 | 1,006 |
| Estimated input tokens | 128,423 | 223 |
| Median local preparation | 0.000041 ms | 0.047 ms |
| Allocated memory | 0 B | ~16 KB |

This reduces estimated summarizer input by 99.83% while adding about 0.047 ms of local projection work.

The replay assertions are structural checks on representative sessions: they verify that user intent, decisions, exact failures, active plans, ask-user exchanges, changed files, and the recent tail survive. They do not claim to measure model-summary quality across every provider or workload.

## Validation

- `make fmt-check`
- `go vet ./...`
- focused compaction tests in `internal/agent`, `internal/sessions`, and `internal/tui`
- `make vulncheck`
- `git diff HEAD --check`
- manual TUI verification of `/compact`, continued conversation, `/resume`, and exact constraint recall

`go test ./...` still reports three unrelated failures that reproduce outside the changed paths: two CLI doctor tests return exit code 3, and the suggestion-overlay test matches the word `compaction` in the checkout path or branch name. `make lint-static` reports two stale `.pr867-validation` paths that are not present in this branch.

## Summary by CodeRabbit

* **Improvements**
  * Improved conversation compaction to preserve user intent, summaries, responses, tool activity, errors, constraints, approvals, unresolved issues, changed files, and artifact details.
  * Reduced unnecessary tool output during summarization while retaining important context and lowering token usage.
  * Improved continuity across repeated compaction cycles by tracking runtime evidence and resolved failures.
  * Added prompt-free compaction planning, accurate prompt metrics, truncation reporting, and clearer fallback handling.
* **Bug Fixes**
  * Prevented important task details from being lost during summarization.
  * Improved handling and visibility of failed or aborted tool actions.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -76,6 +76,12 @@ type CompactionResult struct {
 	// summary message also includes summaryLabel and any preserved structured
 	// state needed by later compactions.
 	SummaryText string
+	// ProjectedChars is the text content sent to the summarizer after semantic
+	// projection, excluding provider protocol framing.
+	ProjectedChars int
+	// Truncated reports whether projection dropped content to satisfy its
+	// bounded brief or previous-summary limits.
+	Truncated bool
 	// Compacted reports whether Messages contains an injected summary.
 	Compacted bool
 }
@@ -239,6 +245,8 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 		RemovedCount:   len(middle),
 		PreservedCount: len(messages) - len(middle),
 		SummaryText:    summary,
+		ProjectedChars: summaryResult.ProjectedChars,
+		Truncated:      summaryResult.Truncated,
 		Compacted:      true,
 	}, nil
 }
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -217,11 +217,11 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 		}, nil
 	}
 
-	summary, err := SummarizeCompactionMessages(middle, opts.Summarize)
+	summaryResult, err := SummarizeCompactionMessages(middle, opts.Summarize)
 	if err != nil {
 		return CompactionResult{}, err
 	}
-	summary = strings.TrimSpace(summary)
+	summary := summaryResult.SummaryText
 
 	// Preserve structured state (active plan + loaded skills) from the elided
 	// middle verbatim, so it is not lost or paraphrased away by the prose summary.
@@ -243,22 +243,45 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 	}, nil
 }
 
+// CompactionSummaryResult describes the shared summarization input and output.
+type CompactionSummaryResult struct {
+	SummaryText    string
+	ProjectedChars int
+	Truncated      bool
+}
+
 // SummarizeCompactionMessages applies the shared semantic projection before
 // invoking a caller-supplied summarizer. Automatic context-pressure compaction
 // and manual session compaction both use this path.
-func SummarizeCompactionMessages(messages []zeroruntime.Message, summarize func([]zeroruntime.Message) (string, error)) (string, error) {
+func SummarizeCompactionMessages(messages []zeroruntime.Message, summarize func([]zeroruntime.Message) (string, error)) (CompactionSummaryResult, error) {
 	if summarize == nil {
-		return "", errors.New("compaction requires a Summarize function")
+		return CompactionSummaryResult{}, errors.New("compaction requires a Summarize function")
 	}
-	summaryInput := projectCompactionInput(messages)
+	projection := projectCompactionInput(messages)
+	summaryInput := projection.messages
 	if len(summaryInput) == 0 {
 		summaryInput = messages
 	}
 	summary, err := summarize(summaryInput)
 	if err != nil {
-		return "", err
+		return CompactionSummaryResult{}, err
 	}
-	return strings.TrimSpace(summary), nil
+	return CompactionSummaryResult{
+		SummaryText:    strings.TrimSpace(summary),
+		ProjectedChars: compactionMessageChars(summaryInput),
+		Truncated:      projection.truncated,
+	}, nil
+}
+
+func compactionMessageChars(messages []zeroruntime.Message) int {
+	total := 0
+	for _, message := range messages {
+		total += len(message.Content)
+		for _, call := range message.ToolCalls {
+			total += len(call.Name) + len(call.Arguments)
+		}
+	}
+	return total
 }
 
 // safeSuffixBoundary walks the preserve boundary backward (toward systemEnd) so
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -216,7 +216,11 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 		}, nil
 	}
 
-	summary, err := opts.Summarize(middle)
+	summaryInput := projectCompactionInput(middle)
+	if len(summaryInput) == 0 {
+		summaryInput = middle
+	}
+	summary, err := opts.Summarize(summaryInput)
 	if err != nil {
 		return CompactionResult{}, err
 	}
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -38,8 +38,9 @@ const compactionTriggerRatio = 0.7
 // transcript (and so tests can assert on it).
 const summaryLabel = "[Summary of earlier conversation]"
 
-// summaryInstructions is the system prompt handed to the summarizer model.
-const summaryInstructions = "You are compacting a coding-assistant conversation to save context. " +
+// CompactionSummaryInstructions is the system prompt handed to both automatic
+// and manually requested compaction summarizers.
+const CompactionSummaryInstructions = "You are compacting a coding-assistant conversation to save context. " +
 	"Write a dense, factual summary of the conversation so far. Preserve: the user's goals and explicit constraints; " +
 	"decisions made and why; files created or modified (with paths) and key code changes; commands run and their important " +
 	"results; and anything still in progress or unresolved. Omit pleasantries. Use terse bullet points. Do not invent details. " +
@@ -216,11 +217,7 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 		}, nil
 	}
 
-	summaryInput := projectCompactionInput(middle)
-	if len(summaryInput) == 0 {
-		summaryInput = middle
-	}
-	summary, err := opts.Summarize(summaryInput)
+	summary, err := SummarizeCompactionMessages(middle, opts.Summarize)
 	if err != nil {
 		return CompactionResult{}, err
 	}
@@ -244,6 +241,24 @@ func CompactMessages(messages []zeroruntime.Message, opts CompactionOptions) (Co
 		SummaryText:    summary,
 		Compacted:      true,
 	}, nil
+}
+
+// SummarizeCompactionMessages applies the shared semantic projection before
+// invoking a caller-supplied summarizer. Automatic context-pressure compaction
+// and manual session compaction both use this path.
+func SummarizeCompactionMessages(messages []zeroruntime.Message, summarize func([]zeroruntime.Message) (string, error)) (string, error) {
+	if summarize == nil {
+		return "", errors.New("compaction requires a Summarize function")
+	}
+	summaryInput := projectCompactionInput(messages)
+	if len(summaryInput) == 0 {
+		summaryInput = messages
+	}
+	summary, err := summarize(summaryInput)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(summary), nil
 }
 
 // safeSuffixBoundary walks the preserve boundary backward (toward systemEnd) so
@@ -580,7 +595,7 @@ func summarizeWithFallback(ctx context.Context, provider Provider, messages []ze
 func summarizeMessagesOnce(ctx context.Context, provider Provider, messages []zeroruntime.Message, onUsage func(Usage)) (string, error) {
 	request := zeroruntime.CompletionRequest{
 		Messages: []zeroruntime.Message{
-			{Role: zeroruntime.MessageRoleSystem, Content: summaryInstructions},
+			{Role: zeroruntime.MessageRoleSystem, Content: CompactionSummaryInstructions},
 			{Role: zeroruntime.MessageRoleUser, Content: "Summarize this conversation:\n\n" + renderTranscript(messages)},
 		},
 		// No tools: this is a plain text summarization call.
--- a/internal/agent/compaction_metadata_test.go
+++ b/internal/agent/compaction_metadata_test.go
@@ -43,6 +43,12 @@ func TestCompactMessagesReturnsMetadataForManualCompaction(t *testing.T) {
 	if result.SummaryText != "manual summary" {
 		t.Fatalf("SummaryText = %q, want trimmed summary", result.SummaryText)
 	}
+	if result.ProjectedChars != compactionMessageChars(captured) || result.ProjectedChars == 0 {
+		t.Fatalf("ProjectedChars = %d, want %d", result.ProjectedChars, compactionMessageChars(captured))
+	}
+	if result.Truncated {
+		t.Fatal("Truncated = true for a projection within its limits")
+	}
 	if len(captured) != 1 || !strings.Contains(captured[0].Content, "first question") || !strings.Contains(captured[0].Content, "second question") {
 		t.Fatalf("summarized projection = %#v, want intent from the non-preserved middle", captured)
 	}
@@ -96,8 +102,43 @@ func TestCompactMessagesNoopReturnsUncompactedMetadata(t *testing.T) {
 	if result.SummaryText != "" {
 		t.Fatalf("SummaryText = %q, want empty", result.SummaryText)
 	}
+	if result.ProjectedChars != 0 || result.Truncated {
+		t.Fatalf("no-op projection metadata = (%d, %t), want zero values", result.ProjectedChars, result.Truncated)
+	}
 	if !reflect.DeepEqual(result.Messages, messages) {
 		t.Fatalf("Messages changed on no-op: %#v", result.Messages)
+	}
+}
+
+func TestCompactMessagesReturnsTruncatedProjectionMetadata(t *testing.T) {
+	messages := []zeroruntime.Message{{Role: zeroruntime.MessageRoleSystem, Content: "system"}}
+	for index := range 20 {
+		messages = append(messages, zeroruntime.Message{
+			Role:    zeroruntime.MessageRoleUser,
+			Content: strings.Repeat("contextword ", 256) + string(rune('a'+index)),
+		})
+	}
+	messages = append(messages,
+		zeroruntime.Message{Role: zeroruntime.MessageRoleAssistant, Content: "recent answer"},
+		zeroruntime.Message{Role: zeroruntime.MessageRoleUser, Content: "latest question"},
+	)
+
+	var captured []zeroruntime.Message
+	result, err := CompactMessages(messages, CompactionOptions{
+		PreserveLast: 2,
+		Summarize: func(toSummarize []zeroruntime.Message) (string, error) {
+			captured = append([]zeroruntime.Message(nil), toSummarize...)
+			return "bounded summary", nil
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !result.Compacted || !result.Truncated {
+		t.Fatalf("compaction metadata = compacted %t, truncated %t; want both true", result.Compacted, result.Truncated)
+	}
+	if result.ProjectedChars != compactionMessageChars(captured) || result.ProjectedChars == 0 {
+		t.Fatalf("ProjectedChars = %d, want %d", result.ProjectedChars, compactionMessageChars(captured))
 	}
 }
 
--- a/internal/agent/compaction_metadata_test.go
+++ b/internal/agent/compaction_metadata_test.go
@@ -43,8 +43,8 @@ func TestCompactMessagesReturnsMetadataForManualCompaction(t *testing.T) {
 	if result.SummaryText != "manual summary" {
 		t.Fatalf("SummaryText = %q, want trimmed summary", result.SummaryText)
 	}
-	if len(captured) != 3 || captured[0].Content != "first question" || captured[2].Content != "second question" {
-		t.Fatalf("summarized middle = %#v, want the three non-preserved non-system messages", captured)
+	if len(captured) != 1 || !strings.Contains(captured[0].Content, "first question") || !strings.Contains(captured[0].Content, "second question") {
+		t.Fatalf("summarized projection = %#v, want intent from the non-preserved middle", captured)
 	}
 	if len(result.Messages) != 4 {
 		t.Fatalf("compacted message count = %d, want 4", len(result.Messages))
--- a/internal/agent/compaction_preserve.go
+++ b/internal/agent/compaction_preserve.go
@@ -425,15 +425,19 @@ type preservedState struct {
 }
 
 type preservedTaskState struct {
-	Objective           string     `json:"objective"`
-	Status              taskStatus `json:"status,omitempty"`
-	Pending             int        `json:"pending,omitempty"`
-	InProgress          int        `json:"in_progress,omitempty"`
-	Completed           int        `json:"completed,omitempty"`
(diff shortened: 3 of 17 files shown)


anandh8x:
@coderabbitai please review the previously skipped or unchecked areas on the latest head, including the remaining `OnPermission` thread. The parallel callback path is serialized by `callbackMutex`, and all current CI checks are green. If there are no remaining actionable issues, please approve the PR and clear the existing changes-requested verdict.

anandh8x:
@coderabbitai you confirmed there are no remaining actionable issues and stated that you approve, but GitHub still shows your original `CHANGES_REQUESTED` review and the old thread remains unresolved. Please submit the formal approval and clear the stale requested-changes verdict now.

gnanam1990 (changes requested):
## Verdict: Request changes — one blocker, everything else is strong

The architecture is right and the measurement is honest work: projecting the middle to a bounded semantic transcript instead of shipping 617KB to a summarizer that returns 193 tokens either way is clearly correct, and the fail-open path (`if len(summaryInput) == 0 { summaryInput = middle }`) is exactly the right shape for a projection that could over-trim. I checked that out and ran it rather than reading it; `go test ./internal/agent/` is green.

### Blocker: `isLikelyToolError` drops most real failures, including Zero's own

The PR's contract says it retains "concise failures" and "unresolved failures". It largely does not, because `isLikelyToolError` matches only five exact prefixes (`error:`, `tool error:`, `failed:`, `permission denied`, `command failed`) and **anything unmatched is treated as success and discarded entirely**.

I ran it against real error strings:

```
"Error: Invalid arguments for view_image"          kept=true
"Error viewing shot.png: file is too large"        kept=false   <-- zero's own tool error
"--- FAIL: TestFoo (0.01s)"                        kept=false   <-- go test
"exit status 1"                                    kept=false
"panic: runtime error: index out of range"         kept=false
"fatal: not a git repository"                      kept=false   <-- git
"The model kimi-k2.6 does not exist"               kept=false   <-- provider
"read_file: no such file or directory"             kept=false
```

The first `kept=false` is the one that decides it. `grep -rh 'errorResult("' internal/tools/` shows this repo emits `"Error applying patch: "` and `"Error editing "` — the `Error <verb>` shape, not `Error:` — so Zero's *own* edit and patch failures are dropped as if they had succeeded. A summary that silently loses "the patch did not apply" is worse than a long one, and the loss is invisible: nothing downstream can tell a dropped error from a tool that worked.

Two ways out, your call:

- **Cheap**: widen the matcher and invert the default. Keep a tool result when it looks like an error *or* is short — a retained non-error costs a few dozen bytes, a dropped error costs the summary its most important content. Right now the tie goes the expensive way.
- **Proper**: plumb the structural signal. `tools.Result` already knows `StatusError`; `zeroruntime.Message` has no error field, which is why prose matching is the only option *here* — so this is a real constraint you worked within, not an oversight. But matching prose for a decision this consequential is the shape `AGENTS.md` invariant 2 warns about, and a `IsError bool` on the tool message would end it permanently.

There is also no test asserting an error survives projection — `compaction_projection_test.go` covers user/assistant/tool-call shaping but nothing pins the `[tool_error]` path, which is why this gap is invisible to CI.

### What I checked and found correct

- **Fail-open**: an empty projection falls back to the original `middle`, so an over-aggressive projection degrades to today's behaviour instead of summarising nothing.
- **Previous summary carried separately** rather than re-summarised, so repeated compaction does not telescope older decisions into noise.
- **Per-turn tool-call cap keeps the LAST N** and says how many were omitted — the right end to keep, since the recent calls are the ones the summary needs.
- **Ordering and budgets** are deterministic, so the same transcript projects to the same brief.

### One caveat on the benchmark, not a blocker

The 99.75% reduction is real and I do not doubt it. What the numbers do not establish is *summary quality*: both versions produce ~193 tokens, and the question is whether the 193 tokens produced from a 1,435-byte brief are as good as the 193 from 617KB. The replay assertion checks that goal, decision, failure, plan and tail survive on one frozen corpus — worth saying plainly in the PR that this is a single-corpus structural check rather than a quality comparison across varied sessions, so nobody reads the table as proving more than it does.

Fix the error heuristic and I would approve this.

gnanam1990 (dismissed):
## Verdict: Approve — the blocker is fixed, and fixed the right way

Re-reviewed at `b01dba28`. You took the structural option rather than widening the prose matcher, which is the one that actually ends this class of bug.

**The chain is complete and I traced all three links:**

- `zeroruntime.Message` gains `IsError bool` — the field whose absence was the reason prose matching existed at all
- `loop.go` sets it from `toolResult.Status == tools.StatusError`, i.e. from the same authority the tool itself reported
- the projection reads `!message.IsError && !isLikelyToolError(...)`, keeping the text check **only** as a fallback for session history written before the field existed — which is the right call, since old transcripts on disk have no flag to read

**Both tool-result construction sites are covered.** `grep MessageRoleTool` in `internal/agent` finds exactly two (loop.go:722 and :3286) and both set `IsError`, so there is no third path quietly producing an unflagged error message.

**Verified against the exact strings from my last review**, now carried as structured errors — every one survives:

```
"Error viewing shot.png: file is too large"   kept=true   <-- zero's own tool error
"--- FAIL: TestFoo (0.01s)"                   kept=true
"exit status 1"                               kept=true
"panic: runtime error: index out of range"    kept=true
"fatal: not a git repository"                 kept=true
"The model kimi-k2.6 does not exist"          kept=true
```

And the projection still projects: a 500-line successful `read_file` result is dropped, so the 99.75% reduction the PR is for is intact. `go test ./internal/agent/ ./internal/zeroruntime/` green.

Everything I approved of last time still holds — the fail-open fallback to `middle`, the previous summary carried separately, keeping the last N tool calls rather than the first.

My earlier caveat about the benchmark stands unchanged and is still not a blocker: the table measures input size, not summary quality, and the replay assertion is a structural check on one frozen corpus. Worth a sentence in the PR description so the numbers are not read as proving more than they do.

Nice fix — the version that plumbs the status is strictly better than the one I suggested as the cheap option.

Vasanthdev2004 (changes requested):
Requesting changes. The direction is right and the benchmark is real, but the projection drops several things that are not reconstructible, and one claim in the description does not hold.

**ask_user answers are discarded.** `executeAskUser` returns `Status: tools.StatusOK` (loop.go:2039), so `IsError` is false, so the tool branch at compaction_projection.go:70 drops the body. That body is the user's own typed answer. If the user says "Postgres only, the MySQL branch is dead, do not touch it", that sentence is gone after compaction: the summarizer never sees it, and `explicitConstraintsFromMessages` cannot rescue it because that only scans user-role messages. The question is lost too, since `compactionToolCallLine` mines only file_path/path/query/pattern and cmd/command/script/shell, so an ask_user call renders as a bare `* ask_user`. This is the one I would fix first. It is the opposite of reconstructible.

**`/compact` never reaches any of this.** The description says the behaviour applies to "automatic context-pressure compaction and manual `/compact`", and lists manual `/compact` as validation. `projectCompactionInput` has exactly one call site, compaction.go:219, and nothing outside `internal/agent` reaches `Compact`/`CompactMessages`. The TUI path goes `handleCompactCommand` to `compactActiveSession` to `PlanCompaction`, a separate summarizer. Either wire it up or drop the claim, but as written the manual verification exercised code this PR does not touch.

**apply_patch leaves no trace.** Its file path lives only in the `patch` argument, which is not in the allowlist, and `recentEdits` matches only `write_file`/`edit_file` (compaction_preserve.go:119). So after compaction there is no record of which files were patched, in the brief or in preserved state. The same allowlist gap blanks web_fetch's URL and task's prompt.

**The 120-line cap drops the oldest content, which is the part the summary exists to carry.** `capCompactionBrief` keeps the last 120 lines, and the cap binds on essentially every real compaction, so the summarizer input sits around 2 KB regardless of whether the history is 24k or 1.3M tokens. Structured state does survive, so plan, edits, skills, instructions, approvals and the original run prompt are fine. What is never summarized even once is the older action record and assistant reasoning. Pre-PR the whole middle reached the summarizer and `summarizeWithFallback` halved it recursively when oversized; that path is now unreachable because the projection always returns exactly one message and the fallback bails at `len(messages) < 2`.

Two smaller ones. `ResolvedFailureKeys` is never cleared, so once a command has succeeded, any later failure of that same command is filtered out of preserved state for the rest of the run: the fail, fix, change, fail-again shape reports a green suite. And the carried-forward previous summary is clipped to 400 informative words each round, keeping the head, so the newest material in it decays away over repeated compaction.

To be clear about what holds up: the preservation machinery genuinely works. I checked objective, plan, skills, tool schemas and project instructions surviving repeated compaction, and the fourteen tests are real tests, not shape checks. Checking your three unrelated failures against the untouched baseline before claiming they were unrelated is exactly right.

Vasanthdev2004:
@anandh8x that review is long, so here is the order I would take them in.

**ask_user first.** It is the only one where the lost content is the user's own words and nothing can regenerate it, and the fix looks small: that result simply is not reconstructible, so it should not sit in the reconstructible bucket.

**The `/compact` claim second**, because it is a decision rather than a fix. Either wire the TUI path through the projection or drop the line from the description. I would not hold the PR for the wiring if you would rather do it separately, but the description should not claim it meanwhile, and the manual validation should not be cited for a path this does not touch.

The rest can follow. apply_patch and the argument allowlist are mechanical, `ResolvedFailureKeys` is close to a one-line clear, and the 120-line cap is the one actually worth thinking about rather than patching quickly, since the right answer might be a byte budget rather than a line count.

Worth saying plainly: the preservation machinery holds up. I checked objective, plan, skills, tool schemas and project instructions surviving repeated compaction and they all do, and your fourteen tests are real tests rather than shape checks.

Also, my own first finding was wrong and I nearly sent it. I probed `isLikelyToolError` in isolation, watched it miss nine of eleven real failure strings, and was ready to call it blocking before checking the call site: `IsError` is consulted first, and `exec_command.go` sets `StatusError` on any nonzero exit, so failing tests and compile errors are preserved. The structured-status approach is sound. My measurement was the bad part.

gnanam1990 (commented):
Reviewed `8b45d995` ("Unify manual and automatic compaction"), which landed after my approval and dismissed it. This is a much larger commit than the previous two — +467/-77 across 14 files, reaching into `internal/sessions` and `internal/tui` for the first time in this PR.

The core of it is right, and it is the fix I'd have asked for if I'd seen the manual path first. `/compact` used to build its summary from `shapedPayloadPreview`, which is an **allow-list** — `EventToolCall` kept only `id`/`name`/`toolName`, `EventToolResult` only those plus `status`. The `arguments` and `output` fields never survived. So the manual summarizer was being handed a list of tool names and statuses and asked to write a useful summary of what happened. `CompactionMessages` + `SummarizeCompactionMessages` puts the real content through the same projection the automatic path uses, and `TestManualCompaction...` pins the join end-to-end by asserting the request carries `agent.CompactionSummaryInstructions` and that the ask_user question and answer both survive. Good test — it asserts the wire content, not struct fields.

I checked the obvious worry and it is **not** a problem: reading `output`/`content`/`arguments` raw out of the event payload is not a redaction bypass. Redaction already happened at the tool chokepoint before the event was written — `loop.go` scrubs `result.Output` and records `result.Redacted`, and `tui/model.go:5439` persists that already-scrubbed value. See the doc note below, though.

## One thing I'd fix before merge

**`plan.Truncated` is derived by string-matching a marker that lives in another package** — `internal/tui/session_controls.go:923`:

```go
truncated = truncated || strings.Contains(message.Content, "omitted to fit compaction budget") ||
    strings.Contains(message.Content, "[middle omitted]")
```

Both literals are inline strings in `internal/agent/compaction_projection.go:104` and `:207`. Two packages, two copies of one fact, nothing asserting they agree — invariant #5.

I proved the drift is undetected. Rewording only the agent-side marker, exactly as a future refactor would:

```
compaction_projection.go:104  "...[transcript trimmed to fit the compaction budget]..."
compaction_projection.go:207  "\n...[trimmed]...\n"
```

`go test ./internal/tui/ -count=1` → **ok**, whole suite green. `plan.Truncated` is now permanently `false`, and since `RecordCompaction` persists it (`replay.go:223`), the durable session record claims nothing was dropped when material was. The new `TestStoreCanPlanCompactionWithoutBuildingLegacyPrompt` only asserts `!plan.Truncated` on a three-event session, so nothing covers the true direction. It fails the other way too: a compacted message whose text happens to contain `[middle omitted]` flips the flag on.

The fix is the same move you already made elsewhere in this PR. Replacing `isLikelyToolError`'s prose sniffing with the structural `IsError` field was exactly right; this is the same prose-sniffing pattern one layer up. Have `SummarizeCompactionMessages` return whether the projection truncated — it is the only code that knows — instead of leaving the caller to recognise its output by sight.

## Smaller notes, none blocking

**The event log is now read twice per `/compact`.** `PlanCompaction` calls `store.ReadEvents` at `replay.go:151`, then `compactActiveSession` calls it again at `session_controls.go:859` to resolve the same refs. Long sessions pay double I/O and double JSON parse for one compaction. Returning the events from `PlanCompaction`, or a variant that does, would avoid it. `sessionEventsForRefs` itself is good — checking `Sequence` and `Type` alongside the ID makes it a real clash guard rather than a lookup, and failing with "retry /compact" is the right call.

**The two new boundary clippers lack the guard you just added to `clipBytes`.** This commit hardened `clipBytes` for `limit <= 0` and `limit <= 3`, but `clipPrefixAtBoundary` and `clipSuffixAtBoundary` right below it take a negative `limit` straight into `text[:end]`, which panics. Unreachable today — both callers derive the limit from constants — so this is purely about the next person who lowers `compactionBriefMaxBytes`. Worth matching the hardening you applied twenty lines up.

**`PromptChars` quietly changed meaning.** It used to be `len(prompt)` for the whole built prompt; it is now the sum of `len(message.Content)` across projected messages, which excludes the system instruction that `StreamCompletion` actually sends. Small, but it is a recorded metric, so a comment saying which one it counts would keep later comparisons honest.

**The projection budget grew a lot, and the PR title says the opposite.** `compactionBriefMaxLines = 120` became `compactionBriefMaxBytes = 24KB`, `compactionPreviousSummaryWords = 400` became `compactionPreviousSummaryBytes = 16KB`, and `mustPreserve` now admits every successful mutating tool result rather than only errors. Against a raw transcript this is still a large reduction, so I don't think it's wrong — the changed-file evidence is the point of the PR. But whatever numbers the description quotes for "reduce compaction context" were measured before this commit. Worth re-measuring so the claim matches the code.

**`CompactionMessages`' doc comment is inaccurate in a security-relevant way.** It says "security-sensitive event payloads use the same redacted previews as the session compaction plan." Only the `default` branch and `ask_user_answers` do; the three shaped branches read payload fields directly. As above that is *safe*, because redaction happened upstream at the tool chokepoint — but the comment states the wrong reason, and someone adding a fourth shaped branch would reasonably conclude they inherit redaction from this function and skip checking their source. Please state the invariant it actually relies on.

## Verified

- `go test ./internal/agent/ ./internal/sessions/ ./internal/tui/ ./internal/zeroruntime/ -count=1` — all green at `8b45d995`.
- Mutation-checked the marker coupling as described above; tree restored clean afterward.
- Traced `ChangedFiles` from `tools.Result` through `loop.go:723` and `copyMessages` into `recentEdits`. Checked the ordering concern raised by the second `for` loop over messages — it is not a defect, because `write_file`/`edit_file` set `ChangedFiles` too, so their paths reappear in the second pass and `lastSeenOrder` still lands everything chronologically. The first loop is now near-redundant except for `pathByID` and failed edits.
- `removeString` in `task_state.go:245` is a genuine fix: a recurring failure previously stayed in `ResolvedFailureKeys` while also sitting in `UnresolvedFailures`.

Not blocking on any of this — today's behavior is correct in every case above. The truncation-marker coupling is the one I'd want closed before merge, since it is cheap and it is a named recurring defect class in this repo. Happy to re-approve as soon as that's addressed.

gnanam1990 (dismissed):
**Approving.**

To put a formal verdict on the review above: nothing in `8b45d995` is wrong today. The redaction question I chased turned out clean, the ordering concern in `recentEdits` dissolved on inspection, `removeString` is a real fix, and all four affected packages test green. The commit repairs a manual `/compact` that was genuinely broken — summarizing from an allow-listed preview that had dropped every `arguments` and `output` field — and it does so by reusing the automatic path rather than growing a second one. That is the right shape.

The one item I'd still like closed is the truncation marker: `plan.Truncated` is recognised by string-matching literals that live in `internal/agent`, and I showed that rewording them leaves the whole `tui` suite green while the persisted session record silently starts claiming nothing was dropped. It is a latent-drift risk rather than a live bug, which is why it isn't blocking — but it is invariant #5, it is cheap, and `SummarizeCompactionMessages` returning the fact is strictly less code than sniffing for it downstream. Please take it either in this PR or as an immediate follow-up; if you'd rather not grow this commit further, a follow-up issue is fine by me.

The four smaller notes — the double `ReadEvents`, the missing `limit <= 0` guard on the two new boundary clippers, the changed meaning of `PromptChars`, and the stale "reduce compaction context" numbers in the description — are all optional. Only the description one has any reader-facing consequence, so it's worth a quick re-measure before merge.

Nice work on the `IsError` field earlier in this PR, incidentally — that was the structural fix rather than the cheap one, and this commit is better for having it to build on.

gnanam1990 (dismissed):
**Approving** at `fe119799`. My previous approval was dismissed by this push, which crossed with it by about three minutes.

You took all four notes, and the truncation fix is better than what I proposed. I suggested `SummarizeCompactionMessages` return the fact; you threaded it properly — `compactionProjection{messages, truncated}` out of the projection, `CompactionSummaryResult{SummaryText, ProjectedChars, Truncated}` out of the summarizer — so `internal/tui` no longer recognises agent's output by sight. `grep` for either marker in `internal/tui` now returns nothing.

It also catches a case the string sniffing would only have caught by luck: `len(fullSummary) > compactionPreviousSummaryBytes` reports truncation of the *previous summary* independently of the brief. The old sniff would have matched `[middle omitted]` from `clipHeadTailBytes` incidentally, and silently stopped the day that marker was reworded.

Re-ran the mutation that proved the gap last time. Flipping `capCompactionBrief`'s new return to `false`:

```
--- FAIL: TestManualCompactionPersistsStructuralTruncationMetadata
    session_controls_test.go:589: compaction did not persist structural truncation
```

Same edit left the entire `tui` suite green at `8b45d995`. The cross-package join is now pinned, which was the whole ask.

The other three are done too:

- `limit <= 0` guards on both `clipPrefixAtBoundary` and `clipSuffixAtBoundary`, matching `clipBytes`.
- `PromptChars` is both documented and more honest than before — `len(CompactionSummaryInstructions) + ProjectedChars`, and `compactionMessageChars` counts tool call names and arguments rather than `Content` alone. The test pinning the exact value keeps it that way.
- The `CompactionMessages` comment now names the invariant it actually depends on: that the session writer scrubs tool output before persistence, with ask_user answers and generic previews redacted locally. That is the correct statement, and it tells the next person adding a shaped branch what they need to check.

Verified at `fe119799`: `go test ./internal/agent/ ./internal/sessions/ ./internal/tui/ -count=1` all green; mutation as above; working tree clean afterward.

Left open, both optional and neither worth another round: the session event log is still read twice per `/compact` (`PlanCompaction` at `replay.go:151`, then again at `session_controls.go:859`), and the PR description's "reduce compaction context" figures predate the budget changes in `8b45d995`. The second is worth a quick edit to the description before merge so the claim matches the constants, but it is not a code change.

Good sequence of fixes on this PR — `IsError`, then the retry wording, and now this. Each one took the structural option over the cheap one.

Vasanthdev2004 (dismissed):
Approving. Everything I raised is addressed, and I re-checked each one against the code rather than the commit message.

Both blocking ones are fixed and I proved them: an ask_user answer and an apply_patch file path both survive the projection now (`ask_user ANSWER survives: true`, `apply_patch FILE survives: true`). The `mustPreserve` gate on `ask_user` and `ChangedFiles` is the right shape, since those are exactly the two categories that are not reconstructible.

The `/compact` claim is now true rather than dropped, which is the better of the two options I offered. `SummarizeCompactionMessages` applies the shared projection and both paths go through it, so the description matches the code. Only the event-to-message conversion is separate, and it carries `IsError` and `ChangedFiles` through, with `ask_user_answers` handled explicitly and redacted.

`capCompactionBrief` keeping both edges under a byte budget is a better answer than the line cap I complained about, and the comment says why: the oldest material has not necessarily been summarized before. The argument allowlist growing to `url`, `prompt`, `description` and `question` closes the web_fetch and task gaps too. `ResolvedFailureKeys` is now cleared on re-failure, so the fail-fix-fail shape reports honestly.

The split-fallback point I raised is effectively moot now rather than fixed: the projection still returns a single message, but the byte cap bounds the input up front, so the oversized-middle case it protected against cannot arise the same way. Worth knowing rather than acting on.

One thing that is not yours: `TestEagerToolSchemaTokenBudget` fails on this branch at 3588 tokens against a 3550 ceiling. It is pre-existing. I bisected it to cd0eb194 (#843, which added `view_image` to the eager set); `4ca4fa75` and `edf660ab` both pass, and current main fails at 3578. So this branch inherits it and adds about 10 tokens. I am raising it separately and it is not a reason to hold this PR.

gnanam1990 (approved):
**Approving** at `e4500339`. Third push in forty minutes; each one dismissed the prior approval, so this is a re-approve of the same PR at a newer head rather than a new objection.

`e4500339` is small and symmetric: `CompactionResult` now carries `ProjectedChars` and `Truncated`, so the automatic path reports the same two facts `SummarizeCompactionMessages` already hands the manual path. Before this, the projection computed them and the automatic caller dropped them on the floor. The test covers both directions — exact `compactionMessageChars` agreement with a non-zero value on the compacting path, and zero values on the no-op path so the fields cannot quietly report stale numbers when nothing was compacted.

One observation, not a request. Nothing in production reads `CompactionResult.ProjectedChars` or `CompactionResult.Truncated` yet — `grep` finds only the two assignments in `compaction.go:248-249` and the assertions in `compaction_metadata_test.go`. The `session_controls.go:874` reader is `CompactionSummaryResult` on the manual path, a different struct. So automatic compaction can still truncate without anything recording it or telling the user, which is the same gap `/compact` just closed. Surfacing the fields is the right first half and the struct is exported, so this is fine as-is — I mention it only so the second half doesn't get forgotten, since the metadata now exists and costs nothing to record.

Verified at `e4500339`: `go test ./internal/agent/ ./internal/sessions/ ./internal/tui/ ./internal/zeroruntime/ -count=1` all green.

Everything from my `fe119799` review still stands — the structural truncation reporting, the clipper guards, the `PromptChars` accounting, and the corrected `CompactionMessages` doc comment. The two open items remain optional and unchanged: the doubled `ReadEvents` per `/compact`, and the PR description's pre-`8b45d995` figures for "reduce compaction context", which are worth correcting before merge so the stated claim matches the constants.

Note that the PR still shows blocked on @Vasanthdev2004's changes-requested from yesterday 18:12, which predates all three of these commits.

Vasanthdev2004 (approved):
Re-approving. My earlier approval was dismissed when you pushed, so I read the two new commits rather than just re-stamping it.

`fe119799` and `e4500339` are an improvement on what I approved. Returning `compactionProjection{messages, truncated}` and having `capCompactionBrief` report `(string, bool)` replaces the TUI sniffing its own summariser output for the literal `"omitted to fit compaction budget"`. I checked and that string match is gone from `session_controls.go`. Truncation of the carried-forward previous summary is now tracked too, which it was not before, and `clipHeadTailBytes` picked up a `limit <= 0` guard.

Every compaction test passes on this head.

Two failures in the suite, both pre-existing and neither yours. `TestEagerToolSchemaTokenBudget` is the eager-schema ceiling: this branch reads 3588 because it predates #867's trim to 3578, and #877 raises the ceiling to fix main. `TestAltScreenTranscriptScrollKeepsFooterFixed` fails on pristine main too; I confirmed that separately against a clean worktree earlier today.

Worth rebasing on main once #877 lands so your CI goes green.

PR #874, opened 2026-08-06
fix(mcp): name the non-text blocks a tool result drops

Vasanthdev2004:
Part 1 of #823. It stops the silent drop. It does not carry the payload yet.

## Why this half is worth landing alone

`TextContent` keeps only blocks whose type is `text`, so a server returning an image, audio, or an embedded resource produced `(empty MCP tool result)`. A screenshot or browser server completes the call, hands back a perfectly good image, and the model is told nothing came back. It then usually retries, spending another call to reach the same empty answer, and the user never learns an image existed.

Naming the dropped blocks ends that loop without moving a single byte of payload.

## The change

- `Content` decodes `mimeType`. Additive and `omitempty`, so servers that never send it decode exactly as before.
- `DroppedContentSummary` describes what `TextContent` discarded: `1 image/png block`, `2 resource blocks, 1 audio/wav block`. Grouped by mime type, ordered by first appearance, so the same result always reads the same way. A server that omits `mimeType` falls back to the block type rather than printing an empty label.
- `registryTool.Run` appends that as a note, and only when something was actually dropped, so a text-only result is byte-for-byte unchanged.

The note also says retrying will return the same thing. Telling the model what happened without telling it not to retry only fixes half the loop.

## What is still open on #823

Carrying image blocks through. That got easier today: #843 landed `zeroruntime.ImageBlock`, `Message.Images`, `CloneImageBlocks` and `SeedMessagesWithImages`, which is the plumbing the issue says to reuse. The MCP side now only has to fill it. I have left #823 open for that.

## Tests

`non_text_content_test.go` drives `registryTool.Run`, not the helper alone, because the property has to be true of the delivered result. Covers image-only, text-alongside-image, text-only unchanged, and the summary's own cases including a missing mime type and repeated blocks.

Mutation-checked: disabling the note block fails `TestAnImageOnlyResultSaysWhatItReturned` and `TestTextAlongsideAnImageStillReportsTheImage`, while `TestATextOnlyResultIsUnchanged` correctly stays green.

`go build ./...`, `GOOS=windows go build ./...`, `go vet`, `gofmt`, and the full `internal/mcp` and `internal/tools` suites are clean.

## Summary by CodeRabbit

* **New Features**
  * MCP results now identify non-text content, such as images, that cannot be displayed.
  * Mixed results preserve available text while indicating which additional content was omitted.
  * Content summaries group repeated types and provide clearer returned-content details.

* **Bug Fixes**
  * Image-only results no longer appear as empty responses.
  * Text-only results remain unchanged.

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -26,6 +26,10 @@ type RemoteTool struct {
 type Content struct {
 	Type string `json:"type"`
 	Text string `json:"text,omitempty"`
+	// MimeType names what a non-text block holds. Decoded but not yet forwarded:
+	// it is what lets a dropped block be described to the model instead of
+	// vanishing (#823). Servers that omit it still decode fine.
+	MimeType string `json:"mimeType,omitempty"`
 }
 
 type CallToolResult struct {
@@ -494,4 +498,53 @@ func TextContent(content []Content) string {
 		}
 	}
 	return strings.TrimSpace(strings.Join(parts, "\n"))
+}
+
+// DroppedContentSummary describes the blocks TextContent discards, e.g.
+// "1 image/png block" or "2 resource blocks, 1 audio/wav block". It returns ""
+// when a result is entirely text, so a caller adds nothing to the ordinary case.
+//
+// This exists because dropping silently is the worst available behaviour. A
+// screenshot server returns a valid image, TextContent keeps nothing, and the
+// call is reported as "(empty MCP tool result)" — so the model concludes the
+// tool produced nothing and usually retries, burning another call on the same
+// empty answer. Naming what came back costs nothing and ends that loop even
+// though the payload still cannot be forwarded.
+//
+// Counts are grouped by mime type and ordered by first appearance, so the same
+// result always produces the same sentence.
+func DroppedContentSummary(content []Content) string {
+	labels := make([]string, 0, len(content))
+	counts := make(map[string]int, len(content))
+	for _, item := range content {
+		if item.Type == "text" {
+			continue
+		}
+		// Prefer the mime type: "image/png" tells the reader more than "image".
+		// A server may omit it, so fall back to the block type rather than
+		// printing an empty label.
+		label := strings.TrimSpace(item.MimeType)
+		if label == "" {
+			label = strings.TrimSpace(item.Type)
+		}
+		if label == "" {
+			label = "unknown"
+		}
+		if _, seen := counts[label]; !seen {
+			labels = append(labels, label)
+		}
+		counts[label]++
+	}
+	if len(labels) == 0 {
+		return ""
+	}
+	parts := make([]string, 0, len(labels))
+	for _, label := range labels {
+		part := fmt.Sprintf("%d %s block", counts[label], label)
+		if counts[label] != 1 {
+			part += "s"
+		}
+		parts = append(parts, part)
+	}
+	return strings.Join(parts, ", ")
 }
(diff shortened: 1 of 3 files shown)


jatmn (dismissed):
LGTM

gnanam1990 (dismissed):
## Verdict: Approve

Checked out and verified rather than read. The defect is real and the fix is the smallest thing that ends it.

**The claim holds.** `TextContent` keeps only `text` blocks, so an image-only result reached the model as `(empty MCP tool result)` — which reads as "the tool produced nothing", and a model that believes that retries. Naming the payload ends the loop even though the bytes still cannot be forwarded, and it says so ("Retrying will return the same thing"), which is the part that actually stops the retry.

**Gut-the-fix**: forcing `DroppedContentSummary` to return `""` fails three tests — `TestAnImageOnlyResultSaysWhatItReturned`, `TestTextAlongsideAnImageStillReportsTheImage`, `TestDroppedContentSummaryNamesTheBlocks`. Load-bearing, and the wiring in `registry.go` is covered too, not just the helper.

**Additivity**: I verified a text-only result produces `""`, so the note is appended only when something was genuinely dropped and an all-text call is byte-for-byte what it was. `nil` content likewise, so `(empty MCP tool result)` still appears for a genuinely empty result rather than being replaced by a confusing note.

**Edge cases I probed directly**, all correct:
- a block with neither `mimeType` nor `type` yields `1 unknown block`, not an empty label
- grouping and pluralisation: `2 image/png blocks, 1 resource block`
- ordering is by first appearance, so the same result always produces the same sentence — worth having, since a map-ordered version would produce a different string each run and make the output non-deterministic for no reason

**Two small things, neither blocking:**

1. `MimeType` is decoded but never forwarded. That is stated in the comment, so it is deliberate — worth a follow-up issue reference so it does not become a field nobody remembers is unused.
2. The message says "Zero cannot forward yet". Now that #843 has landed an image channel on tool results, that sentence has a shelf life. Not this PR's job, but the two will need reconciling.

`go test ./internal/mcp/` green on the branch.

Vasanthdev2004:
Fixed in 55d7cc37, and the point was correct. `registryTool.Run` makes a fresh call on every retry, so promising an identical response was a claim this code cannot make. The note now says retrying cannot recover the payload, which stays true whatever the server does next, because Zero still has nowhere to put a non-text block.

`TestAnImageOnlyResultSaysWhatItReturned` pins both directions: the new wording must be present and the absolute claim must not come back. Mutation checked, reverting the string fails the test.

gnanam1990 (approved):
Re-approving after `55d7cc37`. My earlier approval was auto-dismissed by the push; the new commit only improves things.

The wording change is the right correction, and a subtle one. `Retrying will return the same thing` was a claim this code is in no position to make — each retry is a fresh call and the server may well answer differently. `Retrying cannot recover this payload` is true regardless of what the server does, because the thing that cannot change is that Zero has nowhere to put a non-text block. Same steer to the model, without the false promise.

Verified at `55d7cc37`:

- `go test ./internal/mcp/` green.
- Mutation-checked the new assertions: restoring the old `will return the same thing` wording in `registry.go:340,342` fails `TestAnImageOnlyResultSaysWhatItReturned` on **both** lines — the missing `cannot recover` guidance and the reintroduced identical-response promise. The test pins the wording in both directions, so neither half can rot silently.

Nothing further from me.

kevincodex1 (approved):

PR #876, opened 2026-08-07
fix(tui): a copied NUL byte no longer panics the TUI on Windows

### LINKED ISSUE #875: Copying a transcript selection containing a NUL byte panics the TUI on Windows
### What happened

Copying a transcript selection on Windows kills the whole TUI with a panic. Reported by a teammate on a real Windows machine:

```
panic({0x7ff7c5661da0?, 0x7ff7c5a30fc0?})
syscall.StringToUTF16(...)
        C:/Program Files/Go/src/syscall/syscall_windows.go:33
github.com/atotto/clipboard.writeAll({0x240ced0b5400, 0xb77})
        clipboard_windows.go:111 +0x59d
github.com/atotto/clipboard.WriteAll(...)
github.com/Gitlawb/zero/internal/tui.model.handleTranscriptSelectionMouse.copyTranscriptSelectionCmd.func1()
        internal/tui/transcript_selection.go:1531 +0x29

zero: tui error: program was killed: program experienced a panic
```

### Root cause

`syscall.StringToUTF16` **panics rather than returning an error** when the string contains a NUL byte — that is documented behaviour in the standard library:

> "If s contains a NUL byte this function panics instead of returning an error."

`github.com/atotto/clipboard`'s Windows path calls it directly, so a single NUL anywhere in the copied text takes the process down.

The NUL reaches the clipboard because **`ansi.Strip` removes escape *sequences* and leaves C0 bytes untouched** — the selection pipeline (`selectableLineFromRenderedLine`) strips ANSI and nothing else. The transcript legitimately carries NULs from two independent directions:

1. **Tool output** — reading a binary file, or `git`'s `-z` NUL-separated listings (this repo itself parses those in `internal/tui/files_git_sweep.go:134`).
2. **Zero's own card protocol** — transcript row prefixes are literally `"\x00command-card\x00"` (`command_output.go:56`) and `"\x00plan-card\x00"` (`command_output.go:72`).

### Reproduction

Select and copy any transcript region containing a NUL — e.g. output of a command that emits NUL-separated data, or a region covering a command card. On Windows the TUI dies with the panic above.

Verifiable without Windows:

```go
ansi.Strip("\x00command-card\x00Profile")  // still contains the NUL
```

### Impact

- **Windows: the whole TUI is killed.** Unsaved session state in the run is lost.
- **macOS/Linux: silent corruption** rather than a crash — the same NUL reaches the OS clipboard, and via the OSC52 fallback a stray control byte terminates the escape sequence early and truncates the copy. This is likely why it went unnoticed on other platforms.

### Suggested fix

Sanitise once at the write, not per source: the sources are unbounded (any tool result a user can select from), so a fix per source is one the next tool re-breaks. Strip C0 controls (keeping `\n` and `\t`) in `copyTranscriptSelectionCmd`, before both the native clipboard write and the OSC52 fallback.

I have this implemented with tests and will open a PR against this issue.

gnanam1990:
Fixes #875

## What was wrong

`syscall.StringToUTF16` **panics rather than returning an error** when handed a string containing a NUL — documented standard-library behaviour — and `github.com/atotto/clipboard`'s Windows path calls it directly. One NUL anywhere in a copied transcript selection therefore killed the whole program:

```
panic(...)
syscall.StringToUTF16(...)  syscall_windows.go:33
github.com/atotto/clipboard.writeAll({0x…, 0xb77})
…tui.model.handleTranscriptSelectionMouse.copyTranscriptSelectionCmd.func1()
zero: tui error: program was killed: program experienced a panic
```

The NUL gets through because **`ansi.Strip` removes escape *sequences* and leaves C0 bytes untouched**. The transcript carries NULs from two independent directions:

- **Tool output** — a binary file read, or `git`'s `-z` NUL-separated listings, which this repo itself parses in `files_git_sweep.go:134`
- **Zero's own card protocol** — row prefixes are literally `"\x00command-card\x00"` and `"\x00plan-card\x00"`

## The fix

Sanitise once at the **write**, not per source. The sources are unbounded — any tool result a user can select from is one — so a fix per source is a fix the next tool re-breaks. `clipboardSafeText` drops C0 controls (keeping `\n` and `\t`, which are legal clipboard content and the normal case for a multi-line selection), applied before both destinations.

The **OSC52 fallback needed it too**: a stray control byte terminates the escape sequence early and silently corrupts the copy. That is very likely why this went unnoticed off Windows — the identical NUL produces a bad copy there rather than a dead process.

## Verification

- `TestNulNeverReachesTheClipboard` — four real sources (command-card prefix, plan-card prefix, binary tool output, `git -z` listing); asserts no NUL survives **and** that readable content is not dropped, so this sanitises rather than truncates.
- `TestClipboardSanitisingKeepsLayout` — clean multi-line text with tabs is returned byte-identical; other C0 bytes are removed.
- Mutation-checked: removing the `clipboardSafeText` call makes the panic-guard test fail.
- `make fmt-check`, `go vet ./...`, `go test ./...`, release build + smoke, `git diff --check` all clean. `GOOS=windows go build ./...` and `GOOS=windows go vet` clean.

Two pre-existing failures on `main` are unrelated and unaffected: `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` fail identically with this change stashed.

I could not execute the Windows path itself (no Windows machine); the crash mechanism is pinned by the standard library's own documented contract and the reported stack, and the sanitiser is proven by the tests above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard copying by removing unsafe control characters while preserving readable text, line breaks, and tabs.
  * Prevented copied content from being truncated, corrupted, or causing errors when it contains binary data or special Git output.
  * Copy status counts now accurately reflect the sanitized clipboard content.

### DIFF
--- a/internal/tui/clipboard_nul_test.go
+++ b/internal/tui/clipboard_nul_test.go
@@ -1,0 +1,60 @@
+package tui
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/x/ansi"
+)
+
+// A COPIED NUL KILLED THE WHOLE PROGRAM ON WINDOWS.
+//
+// github.com/atotto/clipboard's Windows path calls syscall.StringToUTF16, which
+// PANICS rather than returning an error when the string contains a NUL — so
+// selecting transcript text that carried one took the TUI down with "program
+// experienced a panic" and a stack ending in clipboard.writeAll. A teammate hit
+// it on a real Windows machine.
+//
+// The NUL reaches the clipboard because ansi.Strip removes escape SEQUENCES and
+// leaves C0 bytes alone, and the transcript carries them from two directions:
+// tool output (binary reads, git's -z listings) and this package's own card
+// protocol, whose prefixes are literally "\x00command-card\x00".
+func TestNulNeverReachesTheClipboard(t *testing.T) {
+	// The card prefix is a real, in-repo source: ansi.Strip leaves it intact.
+	if !strings.ContainsRune(ansi.Strip(commandCardTranscriptPrefix+"Profile"), 0) {
+		t.Fatal("setup: the card prefix no longer carries a NUL; this test guards the wrong thing")
+	}
+	for _, tc := range []struct{ name, in string }{
+		{"card prefix", commandCardTranscriptPrefix + "Profile\nstatus: ok"},
+		{"plan card prefix", planCardTranscriptPrefix + "plan"},
+		{"binary tool output", "before\x00after"},
+		{"git -z listing", "a.go\x00b.go\x00c.go"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got := clipboardSafeText(tc.in)
+			if strings.ContainsRune(got, 0) {
+				t.Fatalf("a NUL survived to the clipboard write: %q — Windows would panic here", got)
+			}
+			// The readable content must survive: this sanitises, it does not truncate.
+			for _, want := range strings.Split(strings.ReplaceAll(tc.in, "\x00", "\n"), "\n") {
+				if want = strings.TrimSpace(want); want != "" && !strings.Contains(got, want) {
+					t.Fatalf("sanitising dropped real content %q from %q", want, got)
+				}
+			}
+		})
+	}
+}
+
+// Newlines and tabs are legal clipboard content and must survive — a multi-line
+// selection is the normal case, not the exception.
+func TestClipboardSanitisingKeepsLayout(t *testing.T) {
+	in := "line one\n\tindented\nline three"
+	if got := clipboardSafeText(in); got != in {
+		t.Fatalf("clean multi-line text was altered:\n got %q\nwant %q", got, in)
+	}
+	// Other C0 bytes go: they would terminate the OSC52 escape early and
+	// corrupt the copy even where they do not panic.
+	if got := clipboardSafeText("a\x07b\x1bc"); got != "abc" {
+		t.Fatalf("control bytes survived: %q", got)
+	}
+}
--- a/internal/tui/transcript_selection.go
+++ b/internal/tui/transcript_selection.go
@@ -1499,7 +1499,54 @@ func (m model) selectedTranscriptText() string {
 	return strings.Join(parts, "\n")
 }
 
+// clipboardSafeText removes the control characters a clipboard write cannot
+// carry, keeping newlines and tabs.
+//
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (approved):
Approving. Good diagnosis and the right place to fix it.

I checked the structural question, which is whether anything can still reach a clipboard write unsanitised: three callers of `copyTranscriptSelectionCmd` (`composer.go:253`, `model.go:4741`, `transcript_selection.go:1458`), both write sites inside it, and the sanitise happens before the closure so the OSC52 fallback is covered too. Nothing bypasses it. Stripping at the write rather than per source is the correct call, and your reasoning for it is the argument I would have made.

One nit, and it is the tests rather than the fix. They exercise `clipboardSafeText` directly, so they pin the helper and not the wiring. I deleted the `text = clipboardSafeText(text)` line at the call site and both tests still passed, which means a later refactor can drop it and reintroduce the panic with CI green. Worth a seam that lets a test drive `copyTranscriptSelectionCmd` and assert no NUL reaches the native write, even something as small as taking the clipboard writer as a parameter.

Not blocking, since I verified the wiring by hand this time. But this exact shape has bitten us three times recently, most memorably #866, which shipped as a complete no-op with six green tests because they all called the helper instead of the path.

Minor question rather than a request: `\r` is stripped along with the other C0 bytes. On Windows the clipboard convention is CRLF, so if any transcript row carries CRLF a paste will arrive LF-only. Probably irrelevant since the rows look LF-native, but you are closer to that code than me.

Vasanthdev2004:
@jatmn this one needs a third approval and it is small.

It is a crash fix: a copied NUL takes the whole TUI down on Windows, because `atotto/clipboard` calls `syscall.StringToUTF16`, which panics rather than returning an error. It is reachable from ordinary use rather than exotic input, since this package's own card row prefixes are literally `"\x00command-card\x00"` and `ansi.Strip` leaves C0 bytes alone. It was hit on a real Windows machine while copying output.

I approved it, and I checked the coverage question by hand rather than reading the diff alone: three callers of `copyTranscriptSelectionCmd`, both clipboard write sites inside it, and the sanitise runs before the closure so the OSC52 fallback is covered too. Nothing reaches a write unsanitised.

One nit is already on the PR so you do not spend time rediscovering it: the tests exercise `clipboardSafeText` directly, so they pin the helper and not the wiring. I deleted the call-site line and both still passed. I did not block on it.

anandh8x (approved):

kevincodex1 (approved):

PR #877, opened 2026-08-08
test(agent): raise the eager tool schema ceiling to 3650

Vasanthdev2004:
`main` is red and every open PR inherits it. `TestEagerToolSchemaTokenBudget` fails at 3578 tokens against a 3550 ceiling.

## Where it came from

Bisected: `edf660ab` and `4ca4fa75` pass, `cd0eb194` (#843, which added `view_image` to the core set) fails at 3588. #867 later trimmed it to 3578, still over.

#843's own CI was green. The ceiling is only crossed once the tool set is combined, so no per-PR check could have caught it, and I merged it on the strength of "CLEAN, no failing checks". Worth knowing that is a blind spot for anything measuring a whole-set budget.

## Raised rather than paid down

The overrun is 28 tokens. Per-tool cost in the eager set:

```
exec_command        909      request_permissions  397
ask_user            387      update_plan          256
lsp_navigate        252      grep                 215
write_stdin         205      web_fetch            187
read_file           181      glob                 160
read_minified_file  121      list_directory       114
skill               112      view_image            82
```

`view_image` is the second cheapest tool in the set. It is not the problem, it is just what crossed the line. The alternative was trimming a description to reclaim 28 tokens, degrading a tool's usability to satisfy a line drawn against a 2026-07 measurement that predates two tools since added.

The ratchet did its job: it caught the creep and forced this to be a decision rather than a drift. Keeping it that way means raising it deliberately, with the reason written down, which is what this does.

## The failure message was wrong

It advised "defer a tool". This test pins `DeferThreshold` at 0, so deferral is inactive and marking a tool deferred leaves it exposed here, changing nothing. I confirmed that by actually deferring `view_image`: the count stayed at 3578 across 14 tools, and it broke `registry_test.go` besides, since the deferrable-builtin list is curated.

Deferring is still worth doing for real sessions, it just cannot move this number. The message now names the levers that can: a smaller schema, one fewer core tool, or a deliberate raise.

## Validation

`go build ./...`, `go vet`, `gofmt`, and the full `internal/agent` and `internal/tools` suites pass. One file changed.

## Summary by CodeRabbit

* **Tests**
  * Increased the allowed tool-schema token threshold.
  * Improved test failure guidance with clearer troubleshooting options and rationale.

### DIFF
--- a/internal/agent/prompt_budget_test.go
+++ b/internal/agent/prompt_budget_test.go
@@ -20,7 +20,20 @@ import (
 // without a session id. Higher-risk tools that do not opt into auto remain excluded.
 const (
 	maxBaseSystemPromptTokens = 3500
-	maxEagerToolSchemaTokens  = 3550
+	// Raised deliberately from 3550 on 2026-08-08. view_image (#843) took the set
+	// to 3588, and #867 trimmed it back to 3578: still 28 tokens over.
+	//
+	// The overrun is a rounding error rather than a regression. view_image costs
+	// 82 tokens, the second cheapest tool in the set, against exec_command's 909
+	// and request_permissions' 397; it is simply the one that crossed the line.
+	// Raised rather than paid down because the alternative was trimming a
+	// description to claw back 28 tokens, degrading a tool's usability to satisfy
+	// a line drawn against a 2026-07 measurement that predates two tools.
+	//
+	// The ratchet did its job: it caught the creep and forced a decision instead
+	// of a drift. That is the point of it, so keep raising it deliberately rather
+	// than reflexively.
+	maxEagerToolSchemaTokens = 3650
 )
 
 func TestSystemPromptTokenBudget(t *testing.T) {
@@ -46,6 +59,11 @@ func TestEagerToolSchemaTokenBudget(t *testing.T) {
 	got := estimateToolDefTokens(exposed)
 	t.Logf("eager core tool schemas: %d tokens across %d tools", got, len(exposed))
 	if got > maxEagerToolSchemaTokens {
-		t.Fatalf("eager tool schemas are %d tokens, over the %d ceiling — defer a tool or raise the ceiling deliberately", got, maxEagerToolSchemaTokens)
+		// NOT "defer a tool": this test pins DeferThreshold at 0, so marking a
+		// tool deferred leaves it exposed here and changes nothing. Deferral is
+		// still worth doing for real sessions, it just cannot move this number.
+		// The levers that do are a smaller schema, one fewer core tool, or a
+		// deliberate raise.
+		t.Fatalf("eager tool schemas are %d tokens, over the %d ceiling — trim a schema, drop a core tool, or raise the ceiling deliberately (deferring will NOT help: this test disables deferral)", got, maxEagerToolSchemaTokens)
 	}
 }


Vasanthdev2004:
@anandh8x @jatmn this one is small and it unblocks everyone, so it is worth jumping the queue.

`main` is red. `TestEagerToolSchemaTokenBudget` fails at 3578 tokens against a 3550 ceiling on `ff608c78`, and every open PR inherits it. I bisected it to cd0eb194 (#843, which added `view_image` to the core set); `4ca4fa75` and `edf660ab` both pass. #843's own CI was green, because the ceiling is only crossed once the tool set is combined, so no per-PR check could have caught it. I merged it, so this is mine to clean up.

This raises the ceiling to 3650 with the reasoning written into the constant. One file, test-only.

Two things worth your eyes rather than a rubber stamp:

The choice to raise instead of trim. The overrun is 28 tokens and `view_image` costs 82, the second cheapest tool in the set against `exec_command`'s 909. Trimming a description to reclaim 28 tokens would degrade a tool to satisfy a line drawn against a 2026-07 measurement that predates two tools. If either of you thinks the ratchet should bite here instead, say so and I will trim rather than raise.

I also corrected the failure message. It advised "defer a tool", which cannot work: the test pins `DeferThreshold` at 0, so deferral is inactive and a deferred tool stays exposed. I confirmed that by actually deferring `view_image`, which left the count at 3578 and broke `registry_test.go` besides, since the deferrable-builtin list is curated.

Separately, and not part of this PR: CI reported success on `ff608c78` while this test fails locally. Something is not running it, or not failing on it. A budget ratchet that only trips on a developer's machine is not doing its job, and that is worth someone chasing.

gnanam1990 (commented):
Reviewed at `8e282f77`. I went to verify the premise before the change, and I can't reproduce it — so this is a comment rather than an approval.

## The red main this fixes appears to already be green

The PR opens with "`main` is red and every open PR inherits it. `TestEagerToolSchemaTokenBudget` fails at 3578 against a 3550 ceiling." I measured the eager set at three points and it is **3423 tokens across 14 tools** at every one:

| ref | measured | ceiling | result |
|---|---|---|---|
| `57858a31` (#867 — this PR's own base) | 3423 | 3550 | **PASS**, 127 headroom |
| `7f39a630` (current main, #879) | 3423 | 3550 | **PASS** |
| `8e282f77` (this PR's head) | 3423 | 3650 | PASS |

It passes at the PR's **own base**, so the ceiling was not being crossed at the commit this branched from, let alone on current main. GitHub's CI agrees: the CI job is `success` on both `7f39a630` and `ff608c78`, the two most recent commits on `main`. And independently — when I rebased #829 onto current main yesterday, its `go test ./...` leg ran green, `internal/agent` included.

I genuinely can't find where 3578 comes from. The set is deterministic (`CoreToolsScoped(dir, nil)` — no MCP, 14 tools; `estimateToolDefTokens` marshals a key-sorted map, so it's stable across runs and platforms). The most likely explanation is a transient local measurement — extra tools registered in the session, or a tree state before #867's description trims fully settled — rather than the committed reality. If there's a CI run on `main` that's actually red on this test, point me at it and I'll withdraw this; I looked and found only green.

## What that means for the change

As written, the effect is to raise a deliberate ratchet from 3550 to 3650 — from 127 tokens of headroom to 227 — to fix a failure that isn't occurring. That is exactly the reflexive raise the new comment warns against: *"keep raising it deliberately rather than reflexively."* The ratchet's value is that 3423 sits close under 3550; widening the gap by 100 tokens on a premise that doesn't hold spends that value for nothing.

## The rest of the PR is genuinely good and I'd keep it

Two things here are correct and worth landing regardless of the ceiling number:

- **The failure-message fix is a real bug.** The old text said "defer a tool or raise the ceiling," but the test sets `Options{}`, which pins `DeferThreshold` at 0 — the comment right above the assertion says so. Deferring a tool leaves it exposed here and moves the number by zero. The new message ("trim a schema, drop a core tool, or raise the ceiling deliberately — deferring will NOT help: this test disables deferral") sends the next person to a lever that actually works. That's a good catch.
- **The whole-set-budget blind-spot note** in the description is a sharp observation: #843 was green on its own CI and only crossed a combined-set budget once merged, which no per-PR check measures. That's worth codifying somewhere more durable than a PR body.

## Suggestion

Rather than raise the ceiling, I'd split the value out: keep 3550, keep the improved failure message and the `DeferThreshold`-is-0 comment, and retitle to the message/documentation fix it really is. If you *can* reproduce a red `main` I can't see, then the raise is justified — but then the PR body's numbers (3588 → 3578) should be re-measured against current `main`, because what I see committed is 3423.

Also minor: the branch is 2 commits behind `main` (#873, #879) and would want a rebase before merge.

Not blocking — it's a test-only change and harmless to correctness. But I don't want to approve loosening a guard to fix a failure I've now checked for at three commits plus CI and can't find.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #879, opened 2026-08-08
Centralize agent context planning

anandh8x:
## Summary

- centralize main agent-loop and automatic-compaction request construction behind a context planner
- preserve the complete existing message and tool surface while exposing content-free context-block accounting
- trace stable-prefix fingerprints with explicit invalidation reasons

## Why

Zero needs one safe seam for future relevance selection, deduplication, pruning, and cache-aware context policies. This change establishes that seam without filtering context or reducing model capability. It does not claim deterministic token savings by itself.

## Evaluation

- identical before/after provider-visible prefix and initial context estimate: 9,177 tokens
- matched GPT-5.5 bounded task remained correct and complete
- bounded run used 5 → 4 tool calls and 31,104 → 29,008 input tokens; treated as run variation rather than a guaranteed planner saving
- a longer repository-analysis task also completed correctly, with the changed run identifying an additional relevant session-log read path

## Validation

- `make fmt-check`
- `go vet ./...`
- focused agent and trace tests
- `make lint-static` (0 issues)
- `make vulncheck` (no vulnerabilities found)
- `git diff HEAD --check`

`go test ./...` passes except for the existing `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` failures in `internal/cli`; both reproduce unchanged from the untouched base commit.

## Summary by CodeRabbit

* **New Features**
  * Added detailed context breakdowns for system, tool, and conversation content.
  * Added visibility into prompt-cache prefix changes, including invalidation reasons.
  * Improved tracing to record and restore prefix invalidation details.
  * Added context reporting for initial, replacement, retry, and compaction requests.

* **Bug Fixes**
  * Standardized request planning across responses, retries, compaction, and final answers.
  * Preserved provider request settings and context data consistently across planning steps.

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -381,6 +381,8 @@ type compactionState struct {
 	// but its token COST must still be counted so usage reports and budgets include it.
 	onUsage func(Usage)
 	task    *taskState
+	planner *contextPlanner
+	trace   *trace.Recorder
 
 	// calibrationRatio scales the raw byte/4 token estimate toward the provider's
 	// real prompt-token count. ApproxTextTokens over-counts code-heavy content by
@@ -427,6 +429,8 @@ func newCompactionState(options Options, task *taskState) *compactionState {
 		preserveLast: options.CompactionPreserveLast,
 		onUsage:      options.OnUsage,
 		task:         task,
+		planner:      newContextPlanner(contextPlannerConfig{contextWindow: options.ContextWindow}),
+		trace:        options.Trace,
 	}
 	return state
 }
@@ -482,7 +486,7 @@ func (state *compactionState) maybeCompact(
 
 	compacted, err := Compact(messages, CompactionOptions{
 		PreserveLast: state.preserveLast,
-		Summarize:    summarizeClosure(ctx, provider, state.onUsage),
+		Summarize:    state.summarizeClosure(ctx, provider),
 		taskState:    state.task.snapshotForCompaction(messages),
 	})
 	if err != nil {
@@ -533,7 +537,7 @@ func (state *compactionState) recover(
 
 	result, compactErr := Compact(messages, CompactionOptions{
 		PreserveLast: state.preserveLast,
-		Summarize:    summarizeClosure(ctx, provider, state.onUsage),
+		Summarize:    state.summarizeClosure(ctx, provider),
 		taskState:    state.task.snapshotForCompaction(messages),
 	})
 	if compactErr != nil {
@@ -567,9 +571,12 @@ func (state *compactionState) recover(
 // provider call. The summary stream intentionally does NOT forward OnText (so
 // compaction stays invisible on the user-facing surface), but it DOES forward
 // OnUsage so the summarizer's token cost is still counted by usage/budgeting.
-func summarizeClosure(ctx context.Context, provider Provider, onUsage func(Usage)) func([]zeroruntime.Message) (string, error) {
+func (state *compactionState) summarizeClosure(ctx context.Context, provider Provider) func([]zeroruntime.Message) (string, error) {
+	if state.planner == nil {
+		state.planner = newContextPlanner(contextPlannerConfig{})
+	}
 	return func(toSummarize []zeroruntime.Message) (string, error) {
-		return summarizeWithFallback(ctx, provider, toSummarize, onUsage)
+		return summarizeWithPlanner(ctx, provider, toSummarize, state.onUsage, state.planner, state.trace)
 	}
 }
 
@@ -581,7 +588,11 @@ func summarizeClosure(ctx context.Context, provider Provider, onUsage func(Usage
 // Non-context-limit errors (and a single message that still won't fit) surface
 // to the caller unchanged.
 func summarizeWithFallback(ctx context.Context, provider Provider, messages []zeroruntime.Message, onUsage func(Usage)) (string, error) {
-	summary, err := summarizeMessagesOnce(ctx, provider, messages, onUsage)
+	return summarizeWithPlanner(ctx, provider, messages, onUsage, newContextPlanner(contextPlannerConfig{}), nil)
+}
+
+func summarizeWithPlanner(ctx context.Context, provider Provider, messages []zeroruntime.Message, onUsage func(Usage), planner *contextPlanner, recorder *trace.Recorder) (string, error) {
+	summary, err := summarizeMessagesOnce(ctx, provider, messages, onUsage, planner, recorder)
 	if err == nil {
 		return summary, nil
 	}
@@ -590,11 +601,11 @@ func summarizeWithFallback(ctx context.Context, provider Provider, messages []ze
 	}
 
 	mid := len(messages) / 2
-	left, leftErr := summarizeWithFallback(ctx, provider, messages[:mid], onUsage)
+	left, leftErr := summarizeWithPlanner(ctx, provider, messages[:mid], onUsage, planner, recorder)
 	if leftErr != nil {
 		return "", leftErr
 	}
-	right, rightErr := summarizeWithFallback(ctx, provider, messages[mid:], onUsage)
+	right, rightErr := summarizeWithPlanner(ctx, provider, messages[mid:], onUsage, planner, recorder)
 	if rightErr != nil {
 		return "", rightErr
 	}
@@ -607,7 +618,7 @@ func summarizeWithFallback(ctx context.Context, provider Provider, messages []ze
 	combined := strings.TrimSpace(left + "\n\n" + right)
 	reduced, reduceErr := summarizeMessagesOnce(ctx, provider, []zeroruntime.Message{
 		{Role: zeroruntime.MessageRoleUser, Content: combined},
-	}, onUsage)
+	}, onUsage, planner, recorder)
 	if reduceErr != nil {
 		if isContextLimitError(reduceErr.Error()) {
 			// Even the two combined partials don't fit (extreme): fall back to the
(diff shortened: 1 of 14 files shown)


Vasanthdev2004 (commented):
Reviewed as a draft. No blocking findings, and the claims hold where I checked them rather than took them.

**The seam is genuinely complete.** `grep zeroruntime.CompletionRequest{` across `internal/agent` returns exactly one construction site, inside the planner itself, with every other path going through `planner.Plan`. That is the strongest evidence the centralization is real rather than partial, and it is the thing most likely to be half-done in a change like this.

**Request identity is properly tested.** `TestContextPlannerPreservesProviderRequest` uses `reflect.DeepEqual` on the whole request, not substring checks, and then verifies deep-copy semantics three levels into the nested `map[string]any` schema plus the image byte slice. Nested-map aliasing is the classic Go trap in exactly this shape and you tested for it, including that a genuinely changed schema does refresh the snapshot and the fingerprint while leaving the earlier plan's request untouched.

**The trace stays content-free.** `explainPrefixChange` returns only `initial`, `unchanged`, `base_instructions`, `confirmation_policy`, `project_context` and `skills`, every one derived from a hash comparison rather than from content. That matters because the trace vocabulary is a published coordination surface and the same property was the review axis on #761.

Planner lifetime checks out too: created per `Run` at `loop.go:186`, so `previousPrefix` starts nil and the first plan reports `initial` rather than inheriting another session's state.

`internal/trace` passes. `internal/agent` fails only on `TestEagerToolSchemaTokenBudget`, which is the ceiling issue #877 fixes, inherited from main and not yours.

Also worth saying: describing the token and tool-call differences as run variation rather than a guaranteed saving is the right call, and rarer than it should be.

Ready for approval as far as I am concerned once it leaves draft.

Vasanthdev2004:
@anandh8x reviewed above, and there is nothing blocking in it. If you were holding it in draft waiting on review feedback, take it out whenever you like and I will approve.

One thing worth knowing: your only failing test is `TestEagerToolSchemaTokenBudget`, which is inherited from main rather than yours. #877 raises that ceiling and needs a human approval; once it lands your suite goes green with no change on your side.

If you are holding it in draft for a reason of your own, ignore that and carry on.

Two things I checked and am confirming back, since they are the claims a reviewer would usually just take on trust. `zeroruntime.CompletionRequest{}` now appears exactly once in `internal/agent`, inside the planner, so the centralization is complete rather than partial. And `explainPrefixChange` only ever emits fixed tokens derived from hash comparisons, so the trace stays content-free, which was the axis #761 was reviewed on.

Vasanthdev2004 (approved):
Approving. Re-checked on `bea0c214` rather than standing on the earlier read: same commit, now out of draft, up to date with main, and all three smoke jobs green.

Two things I re-verified on this head rather than trusting my own earlier note.

`zeroruntime.CompletionRequest{}` is constructed in exactly one place in `internal/agent` now, `context_planner.go:61`, and everything else goes through `planner.Plan`. Partial adoption is the usual way a centralization lands half-done, and this one is complete.

`explainPrefixChange` can only emit fixed literals: initial, unchanged, base_instructions, confirmation_policy, project_context, skills, system_prompt, tools, schema, prefix_changed, comma-joined. No content, no hashes, no paths. That was the review axis on #761 and it holds.

One caveat that is not yours. `TestEagerToolSchemaTokenBudget` fails on main and therefore on your branch, and CI does not catch it, which is why these checks are green. #877 raises the ceiling and is still waiting on a human approval. Nothing for you to do, but do not read the green tick as that test passing.

gnanam1990 (approved):
**Approving.** Reviewed `bea0c214`, re-confirmed as the live head before posting.

@Vasanthdev2004 already covered seam completeness, the `reflect.DeepEqual` request-identity test, deep-copy semantics, trace vocabulary, and planner lifetime, and I agree with all of it — I won't restate it. Everything below is what I checked that his review did not.

## The central claim holds on the wire, not just in a test

The description says the provider-visible prefix is identical before and after. That is the claim a refactor like this lives or dies on, and a test asserting it is not the same as it being true, so I captured the actual bytes.

I built `ff608c78` (the merge base) and `bea0c214`, pointed both at a recording HTTP provider, and ran the same prompt from the **same workspace path and same isolated config dir**:

```
base sha256: 383f7d26920969d1eb67...
pr   sha256: 383f7d26920969d1eb67...
byte-identical
```

Worth noting the first attempt showed a large diff, and it was my harness at fault, not the patch — the two runs used different workspace directories and the system prompt embeds that path. Equal paths, identical bytes. The claim is real.

## Main finding: one load-bearing part of the freeze cache key is untested

`snapshotTools` keys the frozen schemas on `ToolsHash + "\x00" + SchemaHash` (`context_planner.go:87`). I checked the key is complete first — `ToolDefinition` has exactly `Name`, `Description`, `Parameters`, and `toolSubstrings` fingerprints all three — so invariant #10 is satisfied. No issue there.

But the `SchemaHash` half is not pinned by any test. Dropping it:

```go
key := fingerprint.ToolsHash    // SchemaHash removed
```

leaves `go test ./internal/agent/ -run 'Planner|Context'` **green**. That component is load-bearing: with name and description unchanged and only `Parameters` differing between turns, `ToolsHash` alone is stable, so the planner would serve the previous turn's frozen schemas and the provider would receive a stale tool surface.

**The code is correct** — I wrote a throwaway probe and it passes as-written, and fails with the mutation applied, printing the stale schema. So this is a missing guard, not a defect. Twelve lines close it:

```go
func TestSchemaOnlyChangeRefreshesFrozenTools(t *testing.T) {
	planner := newContextPlanner(contextPlannerConfig{})
	messages := []zeroruntime.Message{{Role: zeroruntime.MessageRoleUser, Content: "hi"}}
	base := map[string]any{"type": "object", "properties": map[string]any{"path": map[string]any{"type": "string"}}}
	grown := map[string]any{"type": "object", "properties": map[string]any{"path": map[string]any{"type": "string"}, "limit": map[string]any{"type": "number"}}}

	// Name and description identical: only the schema moves, so ToolsHash alone
	// cannot notice and the freeze must consult SchemaHash.
	_ = planner.Plan(messages, []zeroruntime.ToolDefinition{{Name: "read_file", Description: "read a file", Parameters: base}}, "")
	plan := planner.Plan(messages, []zeroruntime.ToolDefinition{{Name: "read_file", Description: "read a file", Parameters: grown}}, "")

	properties, _ := plan.Request.Tools[0].Parameters["properties"].(map[string]any)
	if _, ok := properties["limit"]; !ok {
		t.Fatalf("second plan served a stale frozen schema: %#v", plan.Request.Tools[0].Parameters)
	}
}
```

For contrast, the deep-copy guard **is** pinned: replacing `copyToolDefinitions(toolDefs)` with `toolDefs` fails `TestContextPlannerPreservesProviderRequest` immediately. The freeze commit landed five minutes before the first review, so this is the least-examined part of the PR and worth the extra guard.

I also checked the specific risk the freeze introduces — the snapshot is returned by reference and shared across stable turns, so a downstream mutation would corrupt every later request. The only `.Tools[...]` write anywhere is `providers/anthropic/provider.go:406`, and it mutates that provider's own `mapped.Tools`, not the `zeroruntime` slice. No in-place `Parameters[...]` writes in any provider. Safe.

## Negative space: the seam is agent-loop-scoped, and that boundary is worth stating

`zeroruntime.CompletionRequest{` has exactly one construction site inside `internal/agent`, as noted. Repo-wide there are five more:

```
internal/tui/session_title.go:147
internal/tui/plan_step_detail.go:376
internal/tui/session_controls.go:925
internal/tui/recap.go:68
internal/cli/workflows.go:980
```

The description scopes the change to "main agent-loop and automatic-compaction request construction," so this is consistent rather than a gap. I raise it because the stated purpose is a seam for future "relevance selection, deduplication, pruning, and cache-aware context policies" — and when those land in the planner, these five paths will not get them.

The `session_controls.go:925` one is the interesting one: #873 unified manual and automatic compaction *summarization* so both share `agent.SummarizeCompactionMessages`, while this PR centralizes *request construction*. Manual compaction still builds its own request in `internal/tui`. The two unifications stop just short of meeting. Not this PR's job, but with both in flight it is worth someone writing down which side owns that request.

## Minor: compaction's fallback planners lose the context window

`compaction.go:576` and `:591` construct `newContextPlanner(contextPlannerConfig{})`, while `:432` passes the real `options.ContextWindow`. `contextWindow` only feeds `breakdown.UsedFraction` (`context_measurement.go:80`) and gates no decision, so the blast radius is accounting: any breakdown produced through those paths reports `UsedFraction` 0. `:591` passes a nil recorder so nothing is emitted, and `:576` is defensive against a planner that the constructor already sets. Reporting-only and probably unreachable — flagging so it is a deliberate choice rather than a surprise if the accounting is ever consumed.

## Verified

- **Provider request byte-identical** vs merge base, as above.
- **`go test ./internal/agent/ ./internal/trace/`** — both fully green at `bea0c214`, including `TestEagerToolSchemaTokenBudget`, which was failing when this PR was first reviewed.
- **The pre-existing-failure claim is accurate.** `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` fail identically on untouched `ff608c78`. Checked rather than taken.
- **Mutation-tested** the deep copy (caught) and the schema cache key (survived — the finding above).
- **Platform coverage**, which the description does not claim: I ran the built binary through complete feature runs on all three OSes — file writes, nested paths, sandbox refusal, `--add-dir` grants, control characters, specialist child processes, and `exec_command`.

| | macOS | Linux | Windows |
|---|---|---|---|
| `bea0c214` | 7/7 | 7/7 | 6/7 |

The single Windows failure is `exec_command`, and it is **not from this PR**: PowerShell cannot start under the sandbox's WRITE_RESTRICTED token (.NET fails `BCrypt.dll` init, `0x8007045A`), and `main` fails it identically. Since this PR touches the agent loop's hot path, I wanted real three-OS execution rather than an assumption, and it is clean.

Nothing here blocks. The schema-key test is the one thing I would add before merge, and it is twelve lines.

PR #880, opened 2026-08-08
Unify tool output views

anandh8x:
## What changed

- introduce a finalized `ToolOutcome` that separates model-facing output, human-facing detail, recoverable artifacts, and reduction diagnostics
- finalize outcomes once at the registry boundary after redaction, command reduction, and output budgeting
- route agent, TUI, and exec surfaces through the appropriate outcome view while preserving compatibility for direct tool callers and persisted results
- preserve detailed command evidence for users, including reduced error output, without returning that full output to the model context
- add regression coverage for view separation, redaction, artifacts, after-hook rebudgeting, reduced errors, and corpus-level token behavior

## Why

Tool output previously relied on several loosely related fields and metadata conventions. This made it easy for model context, user-visible detail, and recoverable raw output to drift or accidentally use the wrong representation.

The unified boundary keeps model context compact while retaining useful detailed output for users and exact redacted artifacts for recovery and inspection.

## Measured behavior

- reduction corpus: 513 estimated model tokens before, 495 after
- retained reduction: approximately 65% before and 67% after
- manual interactive command: approximately 92 tokens reduced to 41 while retaining an expandable 22-line result

The corpus boundaries differ slightly, so the primary claim is preservation of the existing token reduction with improved human detail and recoverability, rather than a standalone 2-point performance gain.

## Validation

- `make fmt-check`
- `go vet ./...`
- focused `internal/tools`, `internal/agent`, `internal/tui`, and `internal/cli` tests
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- manual interactive verification of compact model output and expandable detailed output

The full local suite also exercised all affected packages successfully. Existing environment/baseline failures remain in doctor connectivity, worktree-trust end-to-end, and daemon end-to-end tests; this change does not touch those paths.

## Summary by CodeRabbit

- **New Features**
  - Improved tool-result presentation with distinct model output, human-readable previews, and recoverable artifacts.
  - Added output diagnostics covering truncation, retained content size, and artifact details.
  - Added bounded, UTF-8-safe previews with references to complete output.
  - Preserved finalized tool-result views across saved and restored sessions.

- **Bug Fixes**
  - Preserved useful execution details across CLI, streaming, transcript, and TUI views.
  - Improved error displays with relevant command evidence and fewer misleading success previews.
  - Applied consistent sensitive-content redaction across displayed and stored output.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -689,7 +689,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			}
 			messages = append(messages, zeroruntime.Message{
 				Role:         zeroruntime.MessageRoleTool,
-				Content:      toolResult.Output,
+				Content:      toolResult.ModelOutput(),
 				ToolCallID:   toolResult.ToolCallID,
 				IsError:      toolResult.Status == tools.StatusError,
 				ChangedFiles: append([]string(nil), toolResult.ChangedFiles...),
@@ -721,7 +721,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			if stopReason := stopReasonFromToolResult(toolResult); stopReason != "" {
 				messages = appendAbortedToolResults(messages, collected.ToolCalls[index+1:])
 				messages = append(messages, toolImageMessages...)
-				result.FinalAnswer = toolResult.Output
+				result.FinalAnswer = toolResult.ModelOutput()
 				result.StopReason = stopReason
 				result.Messages = copyMessages(messages)
 				return result, nil
@@ -734,7 +734,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			// aren't fixed by reformatting the call, so a "match this schema" hint
 			// would misdirect the model toward JSON shape or blocked behavior.
 			retriableFailure := isRetriableToolError(toolResult)
-			outcome := guards.observeToolResult(call.Name, retriableFailure, toolResult.Output)
+			outcome := guards.observeToolResult(call.Name, retriableFailure, toolResult.ModelOutput())
 			posture.observeToolOutcome(outcome, toolResult)
 			if outcome.Stop {
 				// The assistant message advertised EVERY collected tool call, but
@@ -750,7 +750,7 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				return result, nil
 			}
 			if outcome.InjectHint && failureHint == "" {
-				failureHint = toolFailureHint(call.Name, toolSchemaJSON(registry, call.Name), toolResult.Output)
+				failureHint = toolFailureHint(call.Name, toolSchemaJSON(registry, call.Name), toolResult.ModelOutput())
 			}
 
 			// Collect the files this successful mutating tool changed. Self-correct
@@ -920,6 +920,21 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 
 func recordOutputBudgetTrace(recorder *trace.Recorder, result ToolResult) {
 	if recorder == nil || result.Meta["output_budget_category"] == "" {
+		return
+	}
+	if result.Outcome.Finalized() {
+		diagnostics := result.Outcome.Diagnostics
+		recorder.EmitOutputBudget(trace.OutputBudgetEvent{
+			Tool:                    result.Name,
+			Category:                diagnostics.Category,
+			OriginalBytes:           diagnostics.OriginalBytes,
+			RetainedBytes:           diagnostics.ModelBytes,
+			EstimatedOriginalTokens: diagnostics.EstimatedOriginalTokens,
+			EstimatedRetainedTokens: diagnostics.EstimatedModelTokens,
+			Truncated:               diagnostics.Truncated,
+			Reason:                  diagnostics.Reason,
+			SpillCreated:            result.Outcome.Artifact != nil,
+		})
 		return
 	}
 	parseInt := func(key string) int {
@@ -1457,14 +1472,15 @@ func executeToolCall(ctx context.Context, registry *tools.Registry, call ToolCal
 		ToolCallID:      call.ID,
 		Name:            call.Name,
 		Status:          result.Status,
-		Output:          result.Output,
+		Output:          result.ModelOutput(),
 		Truncated:       result.Truncated,
 		Meta:            result.Meta,
 		Images:          result.Images,
 		Redacted:        result.Redacted,
 		ChangedFiles:    result.ChangedFiles,
 		ChangeSummaries: result.ChangeSummaries,
-		Display:         result.Display,
+		Display:         result.HumanDisplay(),
+		Outcome:         result.Outcome,
 		LoadedTools:     loadedToolsFromResult(result.Meta),
 		// A tool may signal a mid-run model escalation by carrying the target id
 		// in Meta["escalate_to_model"]. Lift it into the typed loop-level field;
@@ -2061,13 +2077,14 @@ func askUserFallbackResult(ctx context.Context, registry *tools.Registry, call T
 			ToolCallID:      call.ID,
 			Name:            call.Name,
 			Status:          result.Status,
-			Output:          result.Output,
+			Output:          result.ModelOutput(),
 			Truncated:       result.Truncated,
 			Meta:            result.Meta,
 			Redacted:        result.Redacted,
 			ChangedFiles:    result.ChangedFiles,
 			ChangeSummaries: result.ChangeSummaries,
-			Display:         result.Display,
+			Display:         result.HumanDisplay(),
+			Outcome:         result.Outcome,
 		}
 	}
 	return ToolResult{
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -79,6 +79,7 @@ type ToolResult struct {
 	// command execution; callers must not schedule per-file work from them.
 	ChangeSummaries []execution.Change
 	Display         tools.Display
+	Outcome         tools.ToolOutcome
 	// DenialReason categorizes why a tool call was blocked (empty when it ran).
 	// It lets a surface distinguish the cause precisely instead of parsing Output.
 	DenialReason DenialCategory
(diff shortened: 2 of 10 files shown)


Vasanthdev2004 (commented):
Reviewed at `22aac8b0`. No blocking findings. Traced the parts where a change like this usually goes wrong rather than reading it top to bottom.

**The redaction precondition holds.** `finalizeToolOutcome` documents that `boundaryOutput` must already be redacted, and the human preview is built from it, so if that were ever false this would put unredacted output in front of a person. Both call sites capture it immediately after `scrubResultSecrets`: `registry.go:145-146` and `output_boundary.go:101-102`. The ordering is right in both, and it is the thing I would have blocked on.

**No silent-empty path.** `ModelOutput()` returns `Outcome.ModelView` only when finalized and falls back to `result.Output` otherwise, so a direct `Tool.Run` caller that never crosses the registry boundary still gets its output. That fallback is what makes the four substitutions in `loop.go` safe.

**Nothing mutates output after the view is frozen.** `finalizeToolOutcome` runs after reduction, budgeting and the ceiling, and the only thing after it in the deferred boundary is `CommitFileObservation`, which writes `Meta` keys and nothing else. Worth stating explicitly because the whole design rests on it: once `ModelView` is set, any later write to `result.Output` would be invisible to the model, and there is currently no such write.

**Hook re-budget does not reset the accounting.** `RebudgetAfterHook` carries `previous.Diagnostics.OriginalBytes` forward when the outcome was already finalized, so a hook that rewrites output cannot make the original look smaller than it was. Easy thing to get wrong, and it is handled.

**Byte-based preview trimming is the right call.** `boundedHumanOutcomePreview` splits head and tail with `utf8Prefix`/`utf8Suffix`. Line-based retention is what breaks on minified or single-line files, and this avoids it.

Two things, neither blocking:

`human.Preview` is cleared when the result is an error and output was not command-reduced. The comment explains it as stopping a success preview from hiding the failure, which is right. But a tool whose `Preview` IS the error detail loses that detail from the human view while it survives in `Output`. Intended?

`HumanView Display` carries `json:"humanView,omitempty"`, and `omitempty` does nothing for a struct in encoding/json. Harmless, just misleading to the next reader.

One note for your own awareness rather than a review point: `observeToolResult` now takes `ModelOutput()`. That feeds the repeated-failure signature in the guardrails, which keys on output text. It is equivalent today because `ModelView` is assigned from `result.Output` at finalize, so nothing moves. It stops being equivalent the moment anything sets `ModelView` to something other than the final `Output`, and the failure would be silent rather than loud.

I read this rather than ran it, so treat the above as a code review and not a verification that the suite passes.

gnanam1990 (approved):
**Approving** at `22aac8b0`, re-confirmed as the live head before posting.

@Vasanthdev2004 already traced the redaction precondition, the `ModelOutput()` fallback, the frozen-view invariant, the hook re-budget, and the byte-based preview — and I agree with all of it. He was explicit that he read it rather than ran it, so everything below is the empirical and negative-space half.

## The redaction chokepoint is real, and load-bearing under mutation

This is the property the whole design rests on, so I didn't want to take it from a code read. Two things I verified that a read can't:

**The on-disk artifact is redacted, not just the in-memory views.** `TestRegistryOutcomeRedactsModelHumanAndArtifactViews` plants a real `ghp_`-shaped token, and — the part that matters — `os.ReadFile`s the artifact off disk and asserts the secret is absent from the actual spill file, not only from `ModelView`/`Preview`. That is the right assertion for "exact redacted artifacts."

**Breaking the scrub ordering is caught.** I moved the `boundaryOutput` capture one line earlier, before `scrubResultSecrets` — the exact ordering bug the comment at `registry.go:137-138` warns about. The human view then leaks the token and the test fails (`human view leaked secret`). So the chokepoint isn't just correct today, it's pinned against regression.

## Negative space: the self-managed spill path, which the redaction test doesn't cover

The redaction test drives a registry-budgeted fake tool. The real risk sits one path over — `bash` and `exec_command` are self-budgeting, and `applySelfManagedOutputBudget` *reuses* an existing `meta["spill_path"]` (`output_boundary.go:60-61, 70-71`). If either tool wrote that spill during its own `Run()`, before the registry scrub, the artifact would point at a raw, unredacted file while the model and human views were clean. That is the leak this refactor could plausibly have introduced.

It doesn't, and here's why I'm confident:

- `bash`'s registry path (`prepareBashOutput`, `directBudget=false`) returns bounded capture and **does not** call `spillBashStreams` or set `spill_path` — it defers, exactly as its comment at `bash.go:439-441` claims. The only pre-scrub spill (`budgetBashCapture`, `bash.go:471`) is the `directBudget=true` path, which is direct callers who bypass the registry and its scrub entirely — unchanged by this PR and never model-facing.
- `exec_command` is the same shape: `truncateExecOutputSpill` (the spilling variant) is reachable only from `execToolResultWithBudget:431`, gated on `directBudget`. The registry path skips it and flows full output to the boundary.
- Both `exec_command.go` and `bash.go` are **untouched** by this PR (`git diff origin/main --stat` on them is empty), so this spill timing is pre-existing. The refactor references the post-scrub spill the boundary already created; it doesn't move the write ahead of the scrub.

So the reuse at `applySelfManagedOutputBudget` only ever reuses a post-scrub spill on the model-facing path. No leak.

## Empirical

- `go build ./...` and `GOOS=windows go build ./...` clean; `internal/tools`, `internal/agent`, `internal/tui` suites all green.
- The two `internal/cli` failures the PR calls pre-existing (`TestRunDoctorFormatsRedactedProviderDiagnostics`, `TestRunDoctorConnectivityProbesProvider`) fail identically on untouched `origin/main`, and the PR touches none of the doctor/connectivity files. Claim is accurate.
- Model context is unchanged by construction: `ModelView` is assigned from `result.Output` at `tool_outcome.go:86`, and `ModelOutput()` falls back to `result.Output` when unfinalized, so what reaches the provider is byte-identical to before. Vasanth's point about `observeToolResult` now keying the failure-signature on `ModelOutput()` is the one thing to keep an eye on — equivalent only while `ModelView == Output`, which is true today.
- **Mutation-tested a second guard**: gutting the error-preview clearing in `finalizeToolOutcome` fails `TestErrorOutcomeUsesRawCommandEvidenceButDropsUnrelatedPreview`, so the "a success preview must not hide the failure" behaviour Vasanth asked about is genuinely pinned.

## Platform

Since this reworks the tool-result path on the agent's hot path, I ran the real binary through full feature runs — file writes, nested paths, sandbox refusal, `--add-dir` grants, control-character output, specialist child processes, `exec_command`:

| | macOS | Linux | Windows |
|---|---|---|---|
| `22aac8b0` | 7/7 | 7/7 | 6/7 |

The single Windows failure is `exec_command`, and it is **not from this PR** — it's the #865 PowerShell-under-restricted-token regression (`BCrypt.dll` init, `0x8007045A`) that this branch inherits from `main`, which fails it identically. Everything that exercises the reworked outcome path — the write and control-character scenarios especially — is green on all three.

## Not blocking, already noted by Vasanth

The `omitempty` on the `HumanView` struct field is a no-op in `encoding/json` (harmless), and the `human.Preview = ""` on a non-reduced error drops detail for a tool whose `Preview` *is* the error — his "intended?" is worth a one-line answer in the thread, but the corpus test shows the reduced-error case keeps its evidence, which is the important one. Neither changes my verdict.

Clean, well-tested, and the security-critical seam holds where I tried to break it. Nice work.

Vasanthdev2004 (approved):
Approving. Head is still `22aac8b0`, the commit I traced in the review above, so that stands rather than needing a fresh pass: same code, all checks green, and zero commits behind main.

For the record, the parts I verified rather than assumed are unchanged: `boundaryOutput` is captured immediately after `scrubResultSecrets` at both call sites, so the "must already be redacted" precondition on `finalizeToolOutcome` genuinely holds; `ModelOutput()` falls back to `result.Output` when unfinalized so a direct `Tool.Run` caller is never handed an empty view; and nothing mutates output after the view is frozen, since `CommitFileObservation` writes `Meta` keys only.

My two non-blocking notes are still open, and neither should hold this up:

`human.Preview` is cleared when the result is an error and output was not command-reduced. Right for a success preview, but a tool whose `Preview` IS the error detail loses that from the human view while it survives in `Output`. Worth a look when you are next in there.

`HumanView Display` carries `json:"humanView,omitempty"`, and `omitempty` does nothing for a struct in encoding/json. Harmless, just misleading to the next reader.

One thing I flagged for awareness rather than as a finding, repeated here so it does not get lost in the thread: `observeToolResult` now takes `ModelOutput()`, which feeds the repeated-failure signature in the guardrails. It is equivalent today because `ModelView` is assigned from `result.Output` at finalize. It stops being equivalent the moment anything sets `ModelView` to something other than the final `Output`, and that failure would be silent rather than loud.

Vasanthdev2004:
@kevincodex1 this one is ready whenever you have a minute: approved, `MERGEABLE`/`CLEAN`, all nine checks green.

While checking the approved-but-unmerged pile I found the rest of it is not actually waiting on you, so it is worth writing down rather than leaving as "kevin has a backlog".

Six approved PRs are stuck on something neither you nor their authors can clear by deciding anything: #757, #759, #774, #848, #849 and #850. Each is approved, `MERGEABLE`, and green, but reports `BLOCKED` with only seven checks instead of nine. The two missing are `CodeQL` and `Analyze (actions)`, which come from GitHub's default setup rather than a workflow file in the repo, so there is nothing on those branches to re-run. The checks are required, they never reported on those commits, and a required check that never reports blocks the merge indefinitely. #880 has all nine because it was pushed recently; the six oldest date from 19 July to 31 July.

A fresh push on each branch should make the analysis run and let them merge. I can rebase all six onto main and push, which is the natural way to do it and also picks up whatever they have drifted from. Say the word and I will, or tell me if you would rather the authors did it themselves.

Two of those are worth not losing to this: #774 cleans up a daemon child after a startup timeout and #759 moves LSP notifications off the read loop. Both have been sitting since 19 and 20 July on a check that was never going to arrive.

Vasanthdev2004:
Correcting my comment above: the advice in it was wrong, and I tested it rather than leaving you to find out.

I said a fresh push would make the missing analysis run and let those six merge. It does not. I rebased #848 onto current main and pushed it as a single-PR test before touching the others. All seven checks re-ran green on a brand new commit, and it still reports only seven checks and still says `BLOCKED`. `CodeQL` and `Analyze (actions)` did not appear. So the rebase I offered would have rewritten six contributors' branches for no benefit, and I have not done the other five.

What I can state as verified rather than inferred:

- Those two checks track whether the PR comes from a fork, not whether the branch is stale. #880 is a branch on this repo and gets all nine. All six blocked ones are fork PRs and get seven, before and after a rebase.
- Missing them is NOT what makes a fork PR unmergeable in general. #857 and #876 are both fork PRs that merged recently with exactly the same seven checks and no CodeQL.
- The approvals on the six are intact. #848 and #849 each carry three approvals and report `reviewDecision: APPROVED` while still sitting at `BLOCKED`.

So something in branch protection is refusing these that did not refuse #857 and #876, and I cannot see which rule from outside. My guess would be that CodeQL became a required check at some point after those merged, which would make every fork PR unmergeable from that day on, but that is a guess and I have been wrong about this three times today already, so I would rather you read the actual setting than act on my theory.

The question that would settle it: are `CodeQL` and `Analyze (actions)` in the required-status-checks list, and can a fork PR ever satisfy them? If the answer is no, the options are to drop them from required or move CodeQL to a workflow file that fork PRs can run. Either way it is a settings change rather than anything the six authors can do.

Still true from before: #880 is ready whenever you want it, and #774 and #759 have been sitting since 19 and 20 July.

kevincodex1 (approved):

Issue #674, opened 2026-07-14
provider errors give no diagnostic detail, no turn resume, no duplicate-terminal detection


Svector-anu:
## issue 1: provider errors give no diagnostic detail

session events of type "error" only ever store the top-level message
string, e.g.:

  {"type":"error","payload":{"message":"provider error: provider returned error"}}

no http status code, no response body, no upstream error cause is
persisted anywhere in events.jsonl. there is no --verbose/--debug flag
or env var to capture more detail either. when a provider call fails,
there is no way to determine why from the cli or its logs.

repro: session zero_20260713163241_1783960361665303000_1, events
#361, #586, #632, #646, #669, #689 (2026-07-14, 06:12–08:01 utc),
provider=gitlawb-opengateway, model=tencent/hy3. all 6 have identical
generic messages with zero additional context.

## issue 2: no resume after a provider error mid-turn

after a provider error interrupts a turn, the next turn does not
resume from where it left off — it re-reads previously-read files
from scratch to rebuild context. work in progress at the time of the
error is lost, not just the single failed request.

## issue 3: no visibility into duplicate background terminals across sessions

three separate `bun run dev` background terminals were left running
concurrently from three different zero sessions, started 06:33, 08:18,
and 10:54 utc on the same date, same command, same working directory.
nothing in the cli flagged that an identical background terminal was
already running before starting another.

Vasanthdev2004:
@Svector-anu sorry for the delay, and thanks for the detail. The repro on the first one is what makes it actionable.

Splitting this. Issue 1 is a clear defect and we will take a PR: error events keep only the top-level message, so the status code and the upstream cause are lost and there is no way to find out why a provider call failed. The fix is to persist the status and the cause on the error event, with anything from the response body going through the existing secret scrubbing rather than stored raw.

Turn resume and duplicate-terminal detection are separate features with their own design questions. Please open them as their own issues and I will label them. Three unrelated asks on one issue is a large part of why this stalled.

Say the word if you want to take the first one and I will assign it to you.

Svector-anu:
@Vasanthdev2004 you can assign it to me

Vasanthdev2004:
sure @Svector-anu

PR #882, opened 2026-08-09
Persist status code and cause on provider error events (#674)

Svector-anu:
Fixes issue 1 of #674.

## Problem

Session events of type `"error"` only ever persisted the top-level message string:

```json
{"type":"error","payload":{"message":"provider error: provider returned error"}}
```

No HTTP status code, response body, or upstream error cause was captured anywhere, and there's no `--verbose`/`--debug` flag to get more detail either. When a provider call failed there was no way to determine why from the CLI or its logs.

## Fix

- `zeroruntime.StreamEvent` gains `StatusCode`/`Cause` fields, populated at the provider's two response-driven error sites (HTTP error response, streamed error payload).
- `zeroruntime.CollectedStream` carries them through as `ErrorStatusCode`/`ErrorCause`.
- `agent.Run` now returns a `*zeroruntime.StreamError` (implements `error`) instead of a flat `errors.New(...)`, so the detail survives instead of being flattened into a plain string before it reaches the call site.
- New `sessions.ErrorEventPayload(err)` builds the `EventError` payload, adding `"statusCode"`/`"cause"` via `errors.As` when present (this also unwraps wrapped errors, e.g. `fmt.Errorf("...: %w", err)`). All three session-event call sites (`exec.go`, `exec_spec.go`, `tui/model.go`) now use it instead of the old `map[string]any{"message": err.Error()}`.

`Cause` is built from the existing `provider.redact(...)` helper — the same secret-scrubbing already applied to the classified `Error` string — so nothing new is ever persisted unredacted, per the "goes through the existing secret scrubbing" ask in the issue discussion.

## Scope

This PR is scoped to the OpenAI-compatible provider, which matches the issue's repro (`provider=gitlawb-opengateway`, `model=tencent/hy3` — the generic OpenAI-compatible client used for custom/self-hosted gateways). Anthropic and Gemini providers have the identical `classifiedError`/`redact` pattern at their own `StreamEventError` sites and can get the same two-field addition as a fast follow — happy to open that as a separate PR if useful.

Issues 2 (turn resume after a provider error) and 3 (duplicate background terminal detection) are separate per the maintainer's split and not addressed here.

## Testing

- `go build ./...` and `go vet ./...` clean.
- Full existing suite passes for every touched package (`sessions`, `zeroruntime`, `providers/*`, `agent`, `cli`, `tui`).
- Added tests:
  - `internal/sessions/error_payload_test.go` — plain error, structured `StreamError`, wrapped `StreamError` (via `errors.As`), and zero-value status/cause omission.
  - `internal/providers/openai/provider_test.go` — `StatusCode`/`Cause` propagation for both the HTTP-error and streamed-chunk-error paths, including a redaction check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for streaming and provider failures.
  * Error details now preserve HTTP status codes and redacted upstream causes alongside user-facing messages.
  * Wrapped and collected stream errors retain their diagnostic metadata.
  * Error information is presented consistently across command-line, interactive, and streamed workflows.

* **Tests**
  * Added coverage for plain, streamed, wrapped, HTTP, and rate-limit errors, including sensitive-token redaction.


Svector-anu:
Addressed all 3 actionable comments from CodeRabbit's review:

1. **Missing agent-level propagation test** (`internal/agent/loop.go:493`) — added `TestRunReturnsStreamErrorWithStatusCodeAndCause` covering the collector-to-agent boundary directly: a provider emitting `StreamEventError` with `StatusCode`/`Cause` results in `Run` returning an error `errors.As` can still recover those fields from.

2. **Streamed-error test missing redaction assertion** — `TestStreamCompletionStreamedErrorCarriesStatusCodeAndCause` now includes a secret-shaped value in the SSE message and asserts it's absent from `Cause`, matching the existing HTTP-error test's pattern.

3. **`UpstreamUnreachable` branch dropped StatusCode/Cause (Major)** — confirmed this was a real gap: `emitHTTPError`'s early-return for proxy connectivity failures only set `Error`, unlike the classified-error path right below it. Fixed to set `StatusCode`/`Cause` there too, and extended `TestStreamCompletionHumanizesUpstreamUnreachableGatewayError` to assert on them.

All existing + new tests pass, `go build`/`go vet` clean.

Svector-anu:
@Vasanthdev2004 this is ready for review — implements the fix for issue 1 as discussed (status code + upstream cause persisted on the error event, scrubbed through the existing secret redaction). CodeRabbit's automated review feedback has also been addressed (see comment above). Let me know if you'd like any changes.

Vasanthdev2004 (changes requested):
Nice piece of work, and the design is right in the place it matters most. Requesting changes for one small thing that will fail CI, plus two notes.

First, though: **your real CI never ran.** The only check on this PR is CodeRabbit. The other nine (build, vet, format, tests, smoke) did not trigger, so nothing here has been verified by the pipeline. I ran it locally on Windows instead.

## The blocker: `gofmt` is dirty

`gofmt -l` flags `internal/zeroruntime/helpers.go`. Inserting the comment between `Error` and `ErrorStatusCode` splits the struct's alignment group, so gofmt now wants the first four fields aligned to their own narrower width:

```
-	Text             string
-	ToolCalls        []ToolCall
-	Usage            Usage
-	Error            string
+	Text      string
+	ToolCalls []ToolCall
+	Usage     Usage
+	Error     string
```

`gofmt -w internal/zeroruntime/helpers.go` fixes it. I checked this is a genuine violation rather than the CRLF false positive Windows sometimes produces here: the file is `i/lf w/lf`, and the diff is alignment, not line endings.

## What I verified

The security question on a change like this is whether persisting provider error detail leaks credentials into `events.jsonl`. It does not, and I checked rather than taking the doc comment's word for it. Every site that sets `Cause` in `openai/provider.go` passes through `provider.redact`, and `Redact` scrubs the literal API key and auth header plus Bearer-shaped tokens. `Cause` also carries no more than `Error` already did: `ClassifiedError` embeds the same redacted message, so this duplicates existing exposure rather than creating new. Body size is already bounded at 64 KiB by the `LimitReader`.

I also went looking for the failure mode where a helper is correct but unreachable, because I shipped exactly that bug recently: the counter was right and the caller never triggered it. Here the chain holds end to end. The provider sets the fields, `CollectStreamWithOptions` copies them onto `CollectedStream`, `loop.go:493` builds a real `*StreamError` instead of `errors.New`, and `errors.As` recovers it at the three call sites. `TestRunReturnsStreamErrorWithStatusCodeAndCause` drives `Run` rather than the helper, which is the right level for it.

Local results: `go build`, `go vet` clean. `internal/sessions`, `internal/zeroruntime`, `internal/providers/openai` and the new agent test all pass.

## Two notes, neither blocking

**The join is the one untested seam.** You test `Run` returns the error, and you test `ErrorEventPayload` maps a `*StreamError`. Nothing asserts that a persisted session event actually comes out carrying `statusCode`/`cause`. That is the exact join where a change to either side stops the feature working while both tests stay green. One test that records an error event and reads back the payload would close it.

**Three of the four error emitters are untouched.** `StreamEventError` is also emitted by `anthropic/provider.go`, `gemini/provider.go` and `openai/codex_responses.go`, and none of them set `StatusCode`/`Cause`, so an Anthropic or Gemini user still gets the bare message this issue is about. Your description says "issue 1 of #674" so I take the scope as deliberate, and I would not hold the PR for it. Worth saying out loud only so #674 does not get closed on this landing.

Fix the format and re-trigger CI and I am happy to approve.

Svector-anu:
@Vasanthdev2004 thanks for the thorough review, especially for running it locally since CI didn't.

- **gofmt** — fixed in `c762013`. You called it exactly right: the inserted comment split the alignment group on `CollectedStream`, first four fields needed re-aligning to the narrower width. `gofmt -l .` is clean now, `go build`/`go vet` pass, and the full suite for `internal/sessions`, `internal/zeroruntime`, `internal/providers/openai`, `internal/agent` passes locally.
- **The untested join** — added in `08e6a61`: `TestStoreAppendEventPersistsErrorEventStatusCodeAndCause` drives a real `Store.AppendEvent` → `Store.ReadEvents` round trip with a `*StreamError` and asserts `statusCode`/`cause` survive JSON persistence. That's the seam you flagged between `ErrorEventPayload` (payload-mapping test) and `Run` (agent-level test) — now closed.
- **Anthropic/Gemini/codex_responses untouched** — confirmed deliberate, called out as a fast-follow in the PR description. Leaving #674 open for those unless you'd rather I fold them into this PR.
- **Real CI never running** — you're right, and I dug into why: every `CI`/`PR Auto Review` run on this branch (both this push and the original) is stuck in `action_required`, not queued or failed. That's GitHub's fork-PR workflow-approval gate, not something I can clear from the fork side — it needs a maintainer to hit "Approve and run workflows" on the Checks tab. Flagging so it's not mistaken for a CI failure.

Pushed both fixes to the same branch, should show up on this PR automatically.

Svector-anu:
Also: this branch had drifted 17 commits behind \`main\` and GitHub was reporting it as unmergeable. Merged \`main\` in (\`ef00dd6\`) — one trivial conflict in \`internal/agent/loop_test.go\` where both sides appended a new test at the end of the file (no logical overlap, kept all three tests). Build/vet/tests clean on the merged tree; PR shows mergeable again.

euxaristia:
@coderabbitai full review

jatmn (changes requested):

## Merge readiness

  `internal/zeroruntime/types.go`  
  GitHub reports this head as conflicting (`mergeable: false`, `mergeable_state: dirty`). The captured live `main` is `27b319ca`, and merging it with this head produces a content conflict in `StreamEvent` because both sides add fields there. Please rebase onto current `main`, resolve that conflict, and request review of the resolved diff.

## Findings

- [P1] Preserve diagnostics through the image-rejection recovery  
  `internal/agent/loop.go:396`  
  An OpenAI-compatible provider can now attach the HTTP status and redacted upstream cause to a streamed 400 image-rejection error, but this recovery branch replaces the collected error with a plain `fmt.Errorf` and returns before the `StreamError` construction below. `ErrorEventPayload` consequently cannot recover the structured error and persists only `message`, leaving this provider failure non-diagnosable. Please wrap/preserve the existing `StreamError` while keeping the established friendly image-rejection message, and add a regression that exercises this recovery path.

Vasanthdev2004 (changes requested):
Requesting changes on one item. Context first: this PR's CI had never actually run. It was held at `action_required` behind the fork gate, so the green check you saw was CodeRabbit alone.

I released it and it came back red, but **that red is not yours**. Releasing a long-gated run replays the original run, and yours dates from 2026-08-10, so it executed against a base three weeks old. It fails govulncheck on `internal/terminalpet/client.go`, a file this PR does not touch, and I ran govulncheck against current `main`: `No vulnerabilities found`. A rebase clears it. Sorry for the noise; I did not expect a released run to be a replay rather than a fresh one.

**Streamed errors persist a fabricated `statusCode` 500.** At `provider.go:297` the local `statusCode := http.StatusInternalServerError` is a classification bucket, overwritten only when `chunk.Error.Code` matches one of six known entries. This PR newly exports that local as the persisted status code. For a streamed error the HTTP response was 200, so a reader of `events.jsonl` now sees an authoritative-looking `statusCode: 500` for a transaction that never returned 500. Persisting "unknown" or omitting the field is more useful than persisting a bucket, because a consumer cannot tell the fabricated 500 from a real one.

One note: nothing truncates between the response body and `events.jsonl`, so a single failure can persist a 64 KiB `cause`. Worth a bound, though the amplification is smaller than it first looks and depends on which branch produced the error.

The rest of the change is good, and persisting the cause at all is the right call.

Vasanthdev2004:
The two red checks here are both govulncheck, with findings in the Go standard library (net/url and crypto/tls, reached through web_search, mcp and dictation), none of them in this change. The run is from Aug 28 and the toolchain and vulndb have moved since, so I have re-run the failed jobs rather than ask you to touch anything. If they clear, this is back to a normal review; if the stdlib findings are still open, that is repo-wide and not yours to fix here.

Vasanthdev2004:
Re-ran, still red, and now I can say exactly why, which corrects what I said above about it not being yours to act on. The findings are in the standard library at Go 1.26.5 (net/url, crypto/tls), fixed in 1.26.6. The branch pins 1.26.5; main is on 1.26.6 and its latest Security & code health run is clean, so this is the branch being behind main, not anything in the change. A rebase onto current main clears it. Nothing in your diff is at fault; the remedy just happens to be on your side.

PR #887, opened 2026-08-10
Add interactive terminal companions

anandh8x:
## Summary

- add `/pets` companion discovery, preview, installation, selection, and disable controls
- render metadata-driven animated companions with responsive docking, drag placement, click interactions, and resize-safe positioning
- preserve terminal output synchronization, sidebar/composer layout, configuration data, and cross-platform mouse behavior
- tolerate malformed remote catalog entries while retaining valid companions

## Why

Terminal companions add an optional, expressive layer to the Zero interface without affecting users who leave the feature disabled. The implementation supports terminal-native image protocols, persists the selected companion, and keeps the experience usable across compact layouts and active agent workflows.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test -race ./internal/config ./internal/terminalpet ./internal/tui`
- `go test ./...`
- `GOFLAGS=-buildvcs=false go run ./cmd/zero-release build`
- `GOFLAGS=-buildvcs=false go run ./cmd/zero-release smoke`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`

## Summary by CodeRabbit

- **New Features**
  - Added terminal companions with catalog browsing, previews, installation, disabling, animations, and offline access.
  - Added `/pets` and `/pet` commands for selecting or hiding companions.
  - Added Kitty, Sixel, and iTerm2 rendering support.
  - Companions can appear beside the composer or transcript, be dragged, docked, and interact with clicks.
  - Saved companion preferences persist across sessions.

- **Bug Fixes**
  - Improved picker interactions, previews, resizing, and terminal cleanup.
  - Preserved existing configuration settings when changing companion preferences.
  - Added safer handling for unsupported terminals and invalid companion assets.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/atotto/clipboard v0.1.4
 	github.com/aymanbagabas/go-udiff v0.4.1
 	github.com/charmbracelet/colorprofile v0.4.3
+	github.com/charmbracelet/ultraviolet v0.0.0-20260703014108-f5a850f9c2b7
 	github.com/charmbracelet/x/ansi v0.11.7
 	github.com/charmbracelet/x/term v0.2.2
 	github.com/coder/websocket v1.8.15
@@ -21,7 +22,6 @@ require (
 )
 
 require (
-	github.com/charmbracelet/ultraviolet v0.0.0-20260703014108-f5a850f9c2b7 // indirect
 	github.com/charmbracelet/x/exp/golden v0.0.0-20260615092313-b57e5e6d29bb // indirect
 	github.com/charmbracelet/x/termios v0.1.1 // indirect
 	github.com/charmbracelet/x/windows v0.2.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/charmbracelet/x/term v0.2.2
 	github.com/coder/websocket v1.8.15
 	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
+	golang.org/x/image v0.44.0
 	golang.org/x/sys v0.47.0
 	mvdan.cc/sh/v3 v3.13.1
 )
--- a/go.sum
+++ b/go.sum
@@ -64,6 +64,8 @@ github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e h1:JVG44RsyaB9T2KIHavM
 github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e/go.mod h1:RbqR21r5mrJuqunuUZ/Dhy/avygyECGrLceyNeo4LiM=
 golang.org/x/exp v0.0.0-20260611194520-c48552f49976 h1:X8Hz2ImujgbmetVuW+w2YkyZChE3cBpZi2P158rTG9M=
 golang.org/x/exp v0.0.0-20260611194520-c48552f49976/go.mod h1:vnf4pv9iKZXY58sQE1L86zmNWJ4159e1RkcWiLCkeEY=
+golang.org/x/image v0.44.0 h1:+tDekMZED9+LrtB3G5xzRggpVh9CARjZqROla3R3R+I=
+golang.org/x/image v0.44.0/go.mod h1:V8K3KE9KKKE+pLpQDOeN18w9oacNSvy1tDOirTu4xtY=
 golang.org/x/sync v0.22.0 h1:SZjpbeLmrCk4xhRSZFNZW5gFUeCeFgjekvI/+gfScek=
 golang.org/x/sync v0.22.0/go.mod h1:9xrNwdLfx4jkKbNva9FpL6vEN7evnE43NNNJQ2LF3+0=
 golang.org/x/sys v0.47.0 h1:o7XGOvZQCADBQQ4Y7VNq2dRWQR7JmOUW8Kxx4ZsNgWs=
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -842,6 +842,7 @@ func runInteractiveTUIWithSetup(stderr io.Writer, deps appDeps, permissionMode a
 		Version:              version,
 		Theme:                theme,
 		SavedTheme:           resolved.Preferences.Theme,
+		SavedPet:             resolved.Preferences.Pet,
 		UserConfigPath:       userConfigPath,
 		DoctorUserConfigPath: doctorUserConfigPath,
 		ProjectConfigPath:    projectConfigPath,
--- a/internal/config/json_object_edit.go
+++ b/internal/config/json_object_edit.go
@@ -7,6 +7,7 @@ import (
 )
 
 type jsonObjectSpan struct {
+	start   int
 	end     int
 	members []jsonMemberSpan
 }
@@ -45,6 +46,9 @@ func setPetPreferenceJSON(data []byte, pet string) ([]byte, error) {
 	preferenceValue := bytes.TrimSpace(data[preferenceMember.valueStart:preferenceMember.valueEnd])
 	if bytes.Equal(preferenceValue, []byte("null")) {
 		if pet == "" {
+			if lastJSONMember(root.members[:preferencesIndex], "preferences") >= 0 {
+				return data, nil
+			}
 			return removeJSONMember(data, root, preferencesIndex), nil
 		}
 		encodedPet, err := json.Marshal(pet)
@@ -100,7 +104,7 @@ func parseJSONObject(data []byte, start int) (jsonObjectSpan, error) {
 	if start < 0 || start >= len(data) || data[start] != '{' {
 		return jsonObjectSpan{}, fmt.Errorf("expected JSON object")
 	}
-	object := jsonObjectSpan{}
+	object := jsonObjectSpan{start: start}
 	leadingStart := start + 1
 	for {
 		i := skipJSONSpace(data, leadingStart)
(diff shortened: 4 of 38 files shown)


anandh8x:
@coderabbitai re-review the latest commit and approve if everything looks good. All requested changes are addressed and checks are green.

anandh8x:
@coderabbitai all requested changes are resolved, all review threads are closed, and every check is green. Please review the latest commit and update your formal review to approve

Vasanthdev2004:
@anandh8x tested this on Windows Terminal 1.24 and hit the black rectangle plus mangled text under the companion. Pushed a fix for the cause, `0c4a1f2c`. Drop it if you would rather own it.

**It is not a Windows quirk.** Sixel is the only protocol here that erases by painting characters. Kitty hands the terminal an image id and the placement goes away; a sixel is already in the cell grid, so `clearRenderedImage` writes literal spaces over the cells it occupied. Those spaces carry whatever SGR the TUI last set, and nothing normalised it, so every erase stamped a filled block in the ambient background. A drag erases and repaints on every pixel, so the block was continuously refreshed over the interface.

That is also why Linux and macOS came back clean. ghostty, kitty, WezTerm and iTerm all resolve to `ImageProtocolKitty` and delete by id, so they never execute this function. Windows Terminal is the only mainstream terminal that reaches the sixel branch, so no amount of testing on the other two platforms could have found it.

Measured rather than reasoned about. Before:

```
\x1b[s\x1b[3;41H      \x1b[4;41H      \x1b[5;41H      \x1b[u
```

No reset anywhere. The fix leads with one and switches to DECSC/DECRC, because only DECSC saves the attributes this now clobbers, and the erase is emitted mid-frame between styled writes the TUI has already started.

Tests assert the bytes, since bytes are what the terminal sees: that SGR is reset BEFORE the first row is painted rather than merely present somewhere, that the erase still covers every claimed cell so it cannot be satisfied by erasing nothing, and that the Kitty path keeps deleting by id rather than inheriting the character-painting erase. Verified by mutation, restoring the old sequence fails the first with the exact reported symptom.

I renamed my test file mid-write, worth mentioning because the repo has a guard for it now: I first called it `sixel_erase_windows_test.go`, which would have compiled on Windows only and silently never run in CI on the other two platforms. It is `sixel_erase_test.go`, and it is platform-independent because it only inspects emitted bytes.

## Two more I found and deliberately did NOT fix

Both are design calls rather than defects, so they are yours.

**The erase rectangle is the app's belief, not the terminal's.** It erases `key.columns` x `key.rows` at remembered coordinates. Sixel pixel dimensions rarely divide evenly into cells, so the real footprint can be a row or column larger: leftovers when it under-erases, eaten UI when it overshoots. Kitty avoids this because the terminal owns the placement. A fix probably means reserving cells the TUI does not draw into, rather than overlaying and erasing.

**Sixel has no in-place move.** `kittyImageContentEqual` gives Kitty a cheap reposition; sixel falls through to full clear plus full repaint on every frame and every drag pixel. That is the storm behind the text corruption, and it is why the symptom is worst while dragging.

**One unrelated thing while I was in here.** `DetectImageSupport` treats `WT_SESSION` being set as proof of sixel support, with no version check, while iTerm two branches up gets an explicit `dottedVersionAtLeast(..., 3, 6, 0)` and a clear refusal message. `WT_SESSION` has been set by every Windows Terminal since 2019 and sixel only arrived in 1.22, so on an older build the companion emits sixel that is swallowed, shows nothing, and still runs the erase over the UI. Not my symptom, I am on 1.24, but someone will hit it. Windows Terminal exports no version variable, so the honest gate is probably a DA1 query rather than an env sniff, which is why I left it alone rather than guessing.

Also worth a word from you before this lands: `go.mod` gains `golang.org/x/image` as a new direct dependency and promotes `ultraviolet` from indirect. CONTRIBUTING asks for prior justification on dependency changes.

Vasanthdev2004:
@anandh8x follow-up on the Windows report. I pushed two fixes that are correct and did not solve it, and instrumenting a build finally showed why. The remaining cause is a layer below both, and it is a design call rather than a defect, so I am leaving it with you.

## What the two commits did fix

`0c4a1f2c` the erase painted with the caller's SGR, so it stamped a filled block in the ambient background.

`cbb4a64f` the cell-size request was gated behind `petPixelProtocolSupported`, which is Kitty-only, so a sixel session never sent `CSI 16 t`, never got a reply, and fell back to the blind 75 pixel height and the `petImageColumns`/`petImageRows` constants for the erase.

Both confirmed live. Trace from a real Windows Terminal session:

```
DRAW cellPx=0x0   height=75 cols=9 rows=5 at x=109 y=24 term=120x30
...
DRAW cellPx=10x20 height=75 cols=7 rows=4 at x=109 y=24 term=120x30
RENDER-OUT <ESC>[s<ESC>[25;110H<ESC>P9;1;0q"1;1;69;75#0;2;0;0;0...
```

Metrics arrive, footprint corrects from 9x5 to 7x4, and the raster header proves it exact: 69px wide over a 10px cell is 6.9 so 7 columns, 75px tall over a 20px cell is 3.75 so 4 rows. Windows Terminal answers `CSI 16 t` with `ESC [6;20;10t`, so terminal support was never the problem. Nothing asked.

## What is actually breaking the UI

The geometry is now provably right and the damage is unchanged. So it was never the geometry.

The sixel is written out of band, straight into the cell grid, behind Bubble Tea's differential renderer. Bubble Tea keeps a model of what every cell holds and rewrites only the cells it believes changed. The image destroys text in cells it still believes contain that text, so it never repaints them and the characters do not come back.

That matches the symptom precisely. The reports are *missing* characters and *broken* borders, not mis-coloured ones: `deepseek-v4-flash` rendering as `deep eek-v4-f ash`, the composer border arriving as dashes. Those are cells nobody repainted, not cells wrongly erased.

Three things corroborate it:

- Kitty is immune because the image is a terminal-owned overlay and the cell grid is never touched, so Bubble Tea's model stays true. That is the real reason Linux and macOS came back clean. It is not a property of those platforms, it is a property of the protocol, and Windows Terminal is the only mainstream terminal that lands on the sixel branch.
- The only `Invalidate()` in this path (`internal/tui/pet_output.go:63`) resets the image renderer's own last-key cache. Nothing marks the destroyed cells dirty.
- The damage tracks the companion's position and survives it moving away, which is stale cells rather than a bad erase.

## Your call

Two routes and I do not think it is mine to pick:

1. Force a repaint of the affected region after every sixel draw and erase. Correct, but per-cell invalidation may not be exposed, and a full-screen repaint on every drag pixel will flicker.
2. Reserve the companion's cells so the TUI never renders content underneath it. No overlap, no damage, at the cost of constraining where it can sit. `petReservedColumns` already carries this idea for the composer wrap.

Two smaller things from the same trace, yours to weigh:

The first ~14 draws run at `cellPx=0x0` before the reply lands, so the opening paint still uses the 9x5 fallback. Harmless once a repaint follows, worth knowing.

Sixel has no in-place move. `kittyImageContentEqual` gives Kitty a cheap reposition; sixel falls through to a full clear and full repaint per frame and per drag pixel, which is why the corruption is worst while dragging.

I deliberately did not widen `petPixelProtocolSupported` itself in `cbb4a64f`, since that would switch on pixel-precise dragging for sixel as a side effect. That is a feature decision, not part of this defect.

anandh8x:
@Vasanthdev2004 I pushed `d195972d` with the Sixel-specific fix. Sixel companions now stay in a reserved dock so they cannot overwrite Bubble Tea cells; Kitty dragging is unchanged. Could you retest on Windows Terminal 1.24 and confirm that text and borders remain intact while output is streaming and pet animations run?

Vasanthdev2004 (approved):
Approving `d195972d`. Tested on Windows Terminal 1.24 throughout, which is the platform this had trouble on.

**The dock is the right call.** A sixel lives in the cell grid, so wherever it is dragged it destroys text Bubble Tea believes is still present and will not repaint. Kitty images are terminal-owned overlays, so the grid is never touched and free dragging costs nothing. Pinning sixel to a reserved dock removes the overlap rather than trying to repair it afterwards, and gating on `petSupportsAlphaOverlay` keeps the full behaviour where it is actually safe. The alternative, repainting the damaged region after every move, would have meant a repaint per drag pixel and visible flicker for a worse result.

**What I checked rather than assumed**, since this fetches and decodes third-party images:

- Host allowlisting is two-tier and the narrower one is the right way round: catalog redirects may land on `petdex.dev`, while image and metadata downloads are pinned to the asset host.
- Redirects are handled properly on both sides. `CheckRedirect` validates every hop against the appropriate allowlist with a hop cap, and the final `response.Request.URL` is re-validated afterwards. That second check is what stops a redirect chain ending somewhere the per-hop check let through, and it is a mistake I have seen made in this repo before.
- Downloads are bounded: 2 MB manifest, 16 MB sprite, enforced with `io.LimitReader(body, limit+1)` rather than trusting Content-Length, plus a 20 second client timeout.
- CI is 8/8 including CodeQL and `Analyze (actions)`.

**Two things for @kevincodex1 rather than blockers.**

`go.mod` gains `golang.org/x/image` as a new direct dependency and promotes `ultraviolet` from indirect. CONTRIBUTING asks for prior justification on dependency changes. The choice looks right to me, it is the Go team's own webp decoder and there is no reasonable alternative for decoding sprite sheets, but the policy call is yours.

Windows users now get a stationary companion where Kitty terminals get a draggable one. That follows from the protocol rather than from anything avoidable, and I would rather ship it docked than shipping it breaking the interface, but it is a platform-parity decision worth making knowingly on a PR titled "interactive terminal companions".

**Scope of this approval.** I reviewed the rendering and network paths closely and verified the Windows behaviour by hand. I did not re-read all 31 files line by line. The two fixes I pushed earlier, `0c4a1f2c` for the erase painting in the caller's background and `cbb4a64f` for the cell-size request being gated to Kitty only, are both still present and unaffected by the dock change.

Good work chasing this down across three platforms with only one of them able to reproduce it.

Vasanthdev2004:
@coderabbitai approve

kevincodex1 (approved):
this is so cool!

PR #890, opened 2026-08-11
fix(providers): discover ChatGPT capabilities

### LINKED ISSUE #889: OAuth ChatGPT login only lists older GPT models in /model list on Zero 0.7.0
### Version / branch / commit

zero 0.7.0

### OS and environment

debian

### Steps to reproduce

1. Install or run Zero 0.7.0.
2. Log in using OAuth with a ChatGPT subscription account.
3. Run:
   /model list
4. Observe the returned model list

### Expected behavior

The model list should include the latest supported models available through the authenticated
ChatGPT subscription, such as:

gpt-5.6-luna/terra/sol

and any other currently supported models.

### Actual behavior

Only the following models are shown:
gpt-5.5
gpt-5.4
mini
spark

_No response_

anandh8x:
## Summary

Fixes #889.

- Discover the authenticated ChatGPT model catalog with the same OAuth identity used for requests, including visible current model slugs.
- Preserve per-model reasoning tiers and defaults from the live catalog, including `max` and `ultra` where available.
- Add a capability-gated `/fast` toggle that forwards the advertised priority service tier and shows its active state in the footer.

## Why

OAuth-backed ChatGPT profiles previously fell back to an incomplete model list and could not accurately expose provider-supported reasoning or service-tier controls.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go run ./cmd/zero-release build` (with `GOFLAGS=-buildvcs=false` in this worktree)
- `go run ./cmd/zero-release smoke` (with `GOFLAGS=-buildvcs=false` in this worktree)
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`
- Interactive verification of live ChatGPT discovery, `/effort`, and `/fast` toggling.

## Summary by CodeRabbit

* **New Features**
  * Added `/fast` for supported models, with service-tier support for priority and flex modes.
  * Model listings now display available service tiers and reasoning-effort options.
  * Added the `ultra` reasoning level.
  * Improved ChatGPT model discovery, metadata handling, and OAuth account resolution.

* **Bug Fixes**
  * Fast mode resets when switching models and is blocked during active runs.
  * Reasoning options are normalized, deduplicated, and limited to supported values.


anandh8x:
@coderabbitai The remaining thread is intentional: none is the automatic/unset effort and is omitted from provider requests, while ultra is forwarded only when the active model live capability catalog advertises it. The normalizer is shared with the ChatGPT Responses adapter, so removing ultra would reject an advertised supported capability. The valid coverage feedback is addressed in b59a02bc. Please re-review the latest commit and approve if no other issues remain.

anandh8x:
@coderabbitai approve

Vasanthdev2004 (changes requested):
Reviewed at `b59a02bc`. Noting it is still a draft, so treat this as the thing to settle before it leaves draft rather than a block.

The gating here is better than I expected going in. `activeServiceTier` re-derives from the live catalog instead of trusting stored state, so a model switch drops the tier without needing a separate reset path; empty discovery means unsupported rather than assumed-supported; the wire path goes through `activeServiceTier` rather than the raw field, so the guard is not bypassable from `model.go`; `openAIServiceTier` is a strict allow-list that omits the field entirely otherwise; and the empty-merge case still falls back to the curated catalog. I went looking for a way to get `service_tier` onto the wire past the guard and did not find one.

One thing I do want settled.

**`/fast` can reach a metered profile, and nothing says it costs more.**

`mergeLiveModels` only overwrites `ServiceTiers` when the live probe supplies them:

```go
if len(live.ServiceTiers) > 0 {
    catalog.ServiceTiers = append([]string{}, live.ServiceTiers...)
}
```

When live supplies none, the merged model keeps the catalog's tiers, and this PR teaches `remoteModel` to parse `service_tiers` and `additional_speed_tiers`. OpenAI's public `/v1/models` does not return service tiers, so for an ordinary api.openai.com key the question of whether `/fast` lights up is answered by third-party catalog data rather than by the provider. If that data does carry `priority` for those models, `fastModeSupported` returns true, `/fast` becomes available, and `service_tier: "priority"` goes out on every subsequent request. OpenAI bills priority processing at a premium.

What the user sees is `Toggle the provider's fast service tier when supported.`, a card reading `fast mode: on`, and a `fast` chip in the footer. Nothing on any of those three surfaces suggests money.

On a ChatGPT OAuth profile this is a non-issue: priority is part of the plan and the toggle is exactly the right feature. The concern is only the metered path.

I could not verify from here whether the catalog actually advertises `priority` for api.openai.com models, and that is the whole question, so you can settle it faster than I can. If it does not, say so and I will drop this. If it does, I would rather see one of:

- gate `fastModeSupported` to providers where the tier is plan-included (the OAuth/ChatGPT path), or
- keep it general and say it costs more, in the command description and on the card. One clause is enough.

Two smaller things, neither blocking:

- `chatGPTModelsProtocolVersion = "0.146.0"` is pinned in source. When it goes stale the failure will be a discovery response that quietly changes shape, which is the kind of thing nobody notices for a release or two. Worth either deriving it from wherever the Codex client version already lives, or a line saying what breaks when it drifts.
- `OAuthLoginForProfile` in `internal/providers/oauth.go` is close to identical to the resolver at `internal/cli/oauth_provider.go:70`. Fine to leave for now, but they will drift.

For the record I also checked the two things this repo has been bitten by before. The `"Bearer " + token` inlining is consistent with the existing resolver and `providerhealth` already splits the two-field header, so redaction still covers it. And I looked for injected per-request headers beyond the account scoping and found none.

gnanam1990 (commented):
## Review (draft) — no blocking findings

Reviewed at `b59a02bc` on current `main`. Build/vet/fmt clean; `internal/providermodeldiscovery`, `providers/...` (incl. the 183s openai suite), `providermodelcatalog`, `tui`, and `agent` suites all green. All CI green, CodeRabbit approved. Since it's a draft I'm posting this as comments, not a gate — but as it stands I'd approve once it's marked ready, modulo the one wire-up note below.

### What I verified
- **The actual #889 fix is sound.** OAuth ChatGPT profiles can now probe (`options.OAuthResolver != nil` added to `canProbeProvider`), and the models request carries a `client_version` query param — the two reasons the old list came back truncated. Response parsing now accepts both `data`/`models`, `id`/`slug`, filters `visibility != "list"` to drop internal entries, and extracts reasoning efforts + service tiers with lowercase-dedupe. `ultra`/`max` only appear when the model advertises them.
- **Identity match is correct.** `providers.OAuthLoginForProfile` returns the bearer resolver *and* the store key from one login; the models request draws its bearer from that resolver and its account header from the same login's `CodexAccountResolver`. So discovery uses exactly the request identity. The `cli/oauth_provider.go` `-50` is a clean extract-to-`providers` refactor enabling that reuse — in scope.
- **`/fast` toggle is robust.** Every read of the tier goes through `activeServiceTier()`, which re-validates against the current model's advertised tiers, so the footer and the forwarded `service_tier` can't desync and a model-unsupported tier can't leak. Capability-gated, blocked mid-run (`m.pending`), and reset on the `/model` switch paths. Forwarding is an allow-list of `priority`/`flex`; anything else is dropped via `omitempty`, and the request fingerprint includes the normalized tier so a tier change forces drift.
- **No cross-account cache leak.** `providermodelcatalog` does no persistent catalog cache (fresh `client.Do` each fetch); the only per-provider store is the TUI's in-memory `modelPickerLiveByProvider`, rebuilt each session from an identity-scoped discovery run — not shared across processes or accounts. `TestModelPickerChatGPTDiscoveryUsesMatchingOAuthAccount` exercises the keying.
- **Scope is clean.** The many 1–7 line edits all thread the new `ServiceTier`/`ReasoningEffortUltra` field end-to-end (`zeroruntime` → `agent` → providers). The two churn-looking test edits (`specialist/manifest_test.go` `ULTRA`→`IMPOSSIBLE`, `openai/session_test.go` `xhigh`→`bogus`) are *forced* by the behavior change and correct — `ULTRA`/`xhigh` are now valid, so they can no longer stand in for the "unknown" case.
- **Untrusted remote JSON** degrades rather than panics — new fields are bounded `[]string`/`string`, validated at use sites.

### Non-blocking — worth addressing before ready
1. **`client_version` has no production wire-up.** `options.ClientVersion` is the right injection point, but no caller sets it, so the hardcoded `chatGPTModelsProtocolVersion = "0.146.0"` (`discovery.go:24`) is the *only* value ever sent. That's a private, drifting ChatGPT protocol number — a future bump could degrade the models list again and silently reintroduce #889. Feed Zero's real client version through `options.ClientVersion` (the plumbing already exists), or at minimum add a comment tying the constant to a maintenance checkpoint.
2. **Model-switch reset isn't on every path.** `provider_wizard.go:1284` and `onboarding.go:886` change `modelName` without the explicit `serviceTier = ""` the `/model` paths use. It's *safe* today because `activeServiceTier()` re-gates on the new model, but adding the reset there would make the "resets on model switch" claim literally true everywhere and not rely on the downstream gate.
3. **Nit:** the OAuth helper's move to `providers/oauth.go` dropped the inline comments that explained its fail-soft branches (the `ErrNoToken` → API-key fallback, and why `AllowPresets` is needed). Worth restoring — they explained non-obvious behavior.

### Not checked
Live OAuth catalog fetch against a real ChatGPT account (can't exercise here) — the interactive verification in the PR description covers that; worth keeping in the description when marking ready. Parent issue #889 carries the `bug` label but not `issue-approved`; flagging for the process gate (team internal-cycle applies).

gnanam1990 (changes requested):
## Verdict: Changes requested

One blocking item — and it's the one @Vasanthdev2004 already raised. I went back and confirmed the mechanism in code, and I agree it should be settled before merge.

### Blocking — `/fast` can silently bill a metered api.openai.com profile at premium

`service_tier: "priority"` is premium-priced on OpenAI's metered API. On a ChatGPT OAuth plan that's included and the toggle is exactly right; the problem is the metered path, where nothing on any surface (`Toggle the provider's fast service tier when supported.`, the `fast mode: on` card, the footer `fast` chip) says it costs money. Tracing it end to end, the vector is live, not hypothetical:

- **The tier is populated for every catalog source, uniformly.** `remoteModel.toModel` (`providermodelcatalog/remote.go:318,336`) sets `ServiceTiers` from `service_tiers`/`additional_speed_tiers` for models.dev, OpenRouter, and OpenGateway alike — not just the ChatGPT live probe. api.openai.com's own `/v1/models` returns no tiers, so `mergeLiveModels` keeps the catalog's tiers (the `if len(live.ServiceTiers) > 0` guard Vasanth quoted).
- **`normalizeServiceTier("fast") → "priority"`** (`discovery.go`, `remote.go:347`), so even a catalog entry labelled `fast` trips it.
- **The gate is provider-agnostic.** `fastModeSupported()` (`tui/session_controls.go`) returns true whenever the *active model* advertises a `priority` tier — it never checks whether the provider is the OAuth/ChatGPT one where the tier is plan-included. So a metered model carrying a catalog `priority`/`fast` tier lights `/fast` up, and `service_tier: "priority"` then rides every request.

Whether this is dormant *today* depends entirely on whether the third-party catalog advertises the tier for api.openai.com models — which is exactly why it shouldn't gate a feature that spends the user's money. Either of Vasanth's two fixes settles it, and I'd lean the same way he does:
1. Gate `fastModeSupported` to providers where the tier is plan-included (the OAuth/ChatGPT path), **or**
2. Keep it general and say it costs more — one clause in the command description and on the card is enough.

### Non-blocking (independently corroborates Vasanth)
- **`chatGPTModelsProtocolVersion = "0.146.0"` is the sole value sent.** `options.ClientVersion` is the right injection point but no caller sets it, so the pinned constant is what goes on the wire. When ChatGPT bumps it, discovery quietly changes shape and re-opens #889 — the kind of drift nobody notices for a release. Wire the real client version through, or add a line saying what breaks when it drifts.
- **Nit:** the OAuth helper's move to `providers/oauth.go` dropped the inline comments explaining its fail-soft branches (`ErrNoToken` → API-key fallback, why `AllowPresets`). Worth restoring.

### What is solid (so the rest isn't relitigated)
The gating design is genuinely good: `activeServiceTier()` re-derives from the live catalog at every read, so the footer and the wire can't desync and a model-*unsupported* tier can't leak; forwarding is a strict `priority`/`flex` allow-list with `omitempty`; the request fingerprint includes the normalized tier; reasoning-effort normalization dedups and admits `ultra`/`max` only when the model advertises them; no cross-account catalog cache; untrusted remote JSON degrades rather than panics; and the ~30 small edits are all legitimate `ServiceTier`/`ultra` plumbing (including the two test edits forced by `ULTRA`/`xhigh` becoming valid). The single thing standing between this and an approve is the metered-cost surface above.

Verified: build/vet/fmt clean; `providermodeldiscovery`, `providers/...` (incl. the 183s openai suite), `providermodelcatalog`, `tui`, `agent` suites green; all CI green.

Vasanthdev2004 (approved):
Approving at `78cef5f7`. Out of draft, checks green.

`78cef5f7` closes the one thing I was holding on. `fastModeSupported` now refuses anything that is not a ChatGPT profile before it looks at the catalog at all, and the comment says why: priority is a subscription capability and a third-party catalog advertising it for a metered model is not authority to spend the user's money. The gate sits above the catalog scan, and `activeServiceTier` routes through the same function, so the wire path is covered rather than just the command.

`TestFastCommandRejectsMeteredPriorityTier` is the case I described, an openai profile whose catalog entry claims `priority`, asserting nothing is forwarded and nothing is retained. That is the regression I would have asked for.

The wording is fixed too. "Toggle fast mode for supported ChatGPT subscription models" tells an operator what they are turning on, where the old string left the cost question open.

My two smaller notes were explicitly non-blocking and I am not holding on them: the pinned `chatGPTModelsProtocolVersion` will go stale silently one day, and `OAuthLoginForProfile` will drift from the resolver in `internal/cli/oauth_provider.go`. Worth a follow-up issue rather than another round here.

For the record the gating held up when I went looking for a way past it: empty discovery reads as unsupported, the tier re-derives instead of trusting stored state so a model switch drops it, `openAIServiceTier` is an allow-list that omits the field otherwise, and the merge falls back to the curated catalog rather than an empty picker.

gnanam1990 (approved):
## Verdict: Approve

The metered-cost blocker is resolved. `78cef5f7` gates `fastModeSupported()` on `providercatalog.NormalizeID(m.providerProfile.CatalogID) == "chatgpt"`, so `service_tier: "priority"` can never reach a metered api.openai.com profile even when a third-party catalog advertises the tier — and because `activeServiceTier()` also routes through `fastModeSupported()`, both availability and wire-forwarding are gated. I verified it: with the gate, a metered profile with a priority-advertising model cannot turn fast on; neuter the gate and it immediately forwards `priority`. The user-facing copy now says "ChatGPT subscription models" throughout, so the surface matches the behavior. Thanks for the clean fix, @anandh8x.

### Non-blocking
- **The new regression test doesn't fully lock the gate in.** `TestFastCommandRejectsMeteredPriorityTier` starts with `serviceTier: "priority"` (fast already on), so `/fast` toggles it *off* whether or not the gate exists — the test passes even with the gate removed. The case that actually exercises the gate is the opposite one: fast **off**, try to turn it **on** on a metered profile, assert it stays off. That variant fails the moment the gate is removed. Worth adding, since it's the behavior the fix exists to guarantee.
- Prior note still stands: `chatGPTModelsProtocolVersion = "0.146.0"` is the sole `client_version` sent (no caller sets `options.ClientVersion`); a future protocol bump could quietly re-open #889.

Verified: `go build ./internal/tui/` clean; `go test ./internal/tui/ -run 'Fast|ServiceTier'` green; all CI green; CodeRabbit + @Vasanthdev2004 already approved.

PR #891, opened 2026-08-12
feat(security): path-jail primitive, credential-store locking, worktree git hardening

gnanam1990:
## Split 1 of 3 of #829 — the independently-reversible hardening layer

Per @anandh8x's review of #829, splitting the change into focused PRs. This is the base: the security/hardening layer, which has **no dependency on the orchestration work** and carries its own independent rollback domain. The orchestration and durable-memory work stack on top of it.

### What this contains
- **`internal/pathjail`** (new) — `os.Root` handle-relative path confinement: ancestor traversal is confined, the final component is checked for reparse points, unprivileged Windows junctions are covered, and temp files are unpredictable and exclusive. This is the path-containment primitive @anandh8x verified as correct on #829.
- **`internal/credstore`** — cross-process file locking around the credential store.
- **`internal/worktrees`** — git invocation hardening.

`pathjail` has no importers in this PR by design; `internal/memory` (split 2) and `internal/specialist` (split 3) adopt it in the stacked PRs.

### On the split shape
The full seven-way split @anandh8x proposed isn't mechanically achievable as independently-building PRs: the plan subsystem is a single package whose 29 `plan_*.go` files cross-reference each other, and the zeromaxing posture reaches into the shared TUI model — so "plan core", "saved plans", "model routing" and "posture" cannot compile as separate PRs without the module-seam refactor requested separately. The feasible stack is three: **(1) this hardening layer → (2) durable memory → (3) the orchestration remainder**, which together reconstruct #829 byte-for-byte.

### Validation
- `go build ./...` clean.
- `go test ./internal/pathjail/ ./internal/credstore/ ./internal/worktrees/` — all green.
- Rebased on current `main` (`cabfeefc`).

Part of #829.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent credential-store updates from overwriting, resurrecting, or losing credentials.
  * Improved cancellation of Git operations so stalled child processes no longer delay command completion.

* **Security**
  * Strengthened filesystem protections against directory escapes and unsafe links.
  * Added safer temporary-file creation within protected storage locations.

* **Reliability**
  * Improved Git command behavior across Unix and Windows with consistent environments and bounded shutdown waits.
  * Added cross-platform safeguards for credential-store file access and replacement.

### DIFF
--- a/internal/credstore/concurrency_test.go
+++ b/internal/credstore/concurrency_test.go
@@ -1,0 +1,246 @@
+package credstore
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+)
+
+func fileStore(t *testing.T, dir string) *Store {
+	t.Helper()
+	store, err := New(Options{Dir: dir, Storage: "file"})
+	if err != nil {
+		t.Fatalf("New: %v", err)
+	}
+	return store
+}
+
+// THE REPRODUCTION, kept as the regression. Before the lock this reliably kept
+// 1 of 100: every writer read the same map, added its own provider, and the
+// last rename published a file missing all the others.
+func TestConcurrentSetKeepsEveryKey(t *testing.T) {
+	dir := t.TempDir()
+	store := fileStore(t, dir)
+
+	const n = 100
+	var wg sync.WaitGroup
+	errs := make(chan error, n)
+	for i := 0; i < n; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			if err := store.Set(fmt.Sprintf("p%03d", i), fmt.Sprintf("k%03d", i)); err != nil {
+				errs <- err
+			}
+		}(i)
+	}
+	wg.Wait()
+	close(errs)
+	for err := range errs {
+		t.Fatalf("Set: %v", err)
+	}
+
+	got, err := store.Providers()
+	if err != nil {
+		t.Fatalf("Providers: %v", err)
+	}
+	if len(got) != n {
+		t.Fatalf("kept %d of %d concurrent writes", len(got), n)
+	}
+	// ...and every value is the one its own writer wrote, not a neighbour's.
+	for i := 0; i < n; i++ {
+		key, ok, err := store.Get(fmt.Sprintf("p%03d", i))
+		if err != nil {
+			t.Fatalf("Get: %v", err)
+		}
+		if !ok || key != fmt.Sprintf("k%03d", i) {
+			t.Fatalf("p%03d = %q (present=%v); a writer's own value must survive", i, key, ok)
+		}
+	}
+}
+
+// DELETE RACES SET, and an unlocked delete loses the other writer's key.
+//
+// A first version of this only checked that keys nobody deleted survived — and
+// every reader sees those, so it passed with the lock removed from Delete. The
+// case that actually breaks is a Delete whose stale read predates a concurrent
+// Set: it writes back a map that never contained the new key, and the Set is
+// gone. So the assertion is on the SETS surviving, not on the bystanders.
+func TestADeleteCannotClobberAConcurrentSet(t *testing.T) {
+	dir := t.TempDir()
+	store := fileStore(t, dir)
+
+	// Throwaway keys for the deleters to remove, written first so a delete
+	// always has something to do.
+	const churn = 40
+	for i := 0; i < churn; i++ {
+		if err := store.Set(fmt.Sprintf("churn%02d", i), "v"); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	const adds = 60
+	var wg sync.WaitGroup
+	for i := 0; i < adds; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			_ = store.Set(fmt.Sprintf("added%02d", i), "v")
+		}(i)
+	}
+	for i := 0; i < churn; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			_, _ = store.Delete(fmt.Sprintf("churn%02d", i))
+		}(i)
+	}
+	wg.Wait()
+
+	got, err := store.Providers()
+	if err != nil {
+		t.Fatal(err)
+	}
+	survivors := map[string]bool{}
+	for _, name := range got {
+		survivors[name] = true
+	}
+	missing := []string{}
+	for i := 0; i < adds; i++ {
+		name := fmt.Sprintf("added%02d", i)
+		if !survivors[name] {
+			missing = append(missing, name)
+		}
+	}
+	if len(missing) > 0 {
+		t.Fatalf("%d of %d concurrent Set calls were clobbered by a racing Delete: %v",
+			len(missing), adds, missing)
+	}
+	// ...and every churn key really was removed, so the deletes were not
+	// no-ops that would make the check above vacuous.
+	for i := 0; i < churn; i++ {
+		if survivors[fmt.Sprintf("churn%02d", i)] {
+			t.Fatalf("churn%02d survived; the deletes did nothing and this test proves nothing", i)
+		}
+	}
+}
+
+// ACROSS PROCESSES, which is the case that matters: a plan running with
+// max_workers > 1 spawns child PROCESSES, and an in-process mutex would not
+// have helped any of them. The race detector cannot see this one either — it is
+// two OS processes — so it is driven for real.
+func TestConcurrentSetAcrossProcesses(t *testing.T) {
+	if testing.Short() {
+		t.Skip("spawns processes")
+	}
+	helper := os.Getenv("ZERO_CREDSTORE_HELPER_DIR")
+	if helper != "" {
+		// Child mode: write our slice of the keys and exit.
+		store, err := New(Options{Dir: helper, Storage: "file"})
+		if err != nil {
+			os.Exit(3)
+		}
+		prefix := os.Getenv("ZERO_CREDSTORE_HELPER_PREFIX")
+		for i := 0; i < 25; i++ {
+			if err := store.Set(fmt.Sprintf("%s%02d", prefix, i), "v"); err != nil {
+				os.Exit(4)
+			}
+		}
+		os.Exit(0)
+	}
+
+	dir := t.TempDir()
+	const children = 4
+	var wg sync.WaitGroup
+	failures := make(chan string, children)
+	for c := 0; c < children; c++ {
+		wg.Add(1)
+		go func(c int) {
+			defer wg.Done()
+			cmd := exec.Command(os.Args[0], "-test.run=TestConcurrentSetAcrossProcesses", "-test.v")
+			cmd.Env = append(os.Environ(),
+				"ZERO_CREDSTORE_HELPER_DIR="+dir,
+				fmt.Sprintf("ZERO_CREDSTORE_HELPER_PREFIX=c%d_", c),
+			)
+			if out, err := cmd.CombinedOutput(); err != nil {
+				failures <- fmt.Sprintf("child %d: %v\n%s", c, err, out)
+			}
+		}(c)
+	}
+	wg.Wait()
+	close(failures)
+	for failure := range failures {
+		t.Fatal(failure)
+	}
+
+	store := fileStore(t, dir)
+	got, err := store.Providers()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(got) != children*25 {
+		t.Fatalf("kept %d of %d keys written by %d concurrent processes", len(got), children*25, children)
+	}
+}
+
+// The lock lives BESIDE the data file, never on it. write publishes by rename,
+// so a lock taken on the data file would be attached to an inode the next
+// writer has already replaced — every writer would appear to hold it.
+func TestTheLockIsNotTheDataFile(t *testing.T) {
+	dir := t.TempDir()
+	store := fileStore(t, dir)
+	if store.lockPath() == store.file {
+		t.Fatal("the lock is the data file; rename would carry it away")
+	}
+	if !strings.HasPrefix(filepath.Base(store.lockPath()), filepath.Base(store.file)) {
+		t.Fatalf("lock %q is not beside the data file %q", store.lockPath(), store.file)
+	}
+	if err := store.Set("p", "k"); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := os.Stat(store.lockPath()); err != nil {
+		t.Fatalf("the lock file was not created: %v", err)
+	}
+	// ...and it survives a write, since the rename replaces only the data file.
+	if err := store.Set("q", "k"); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := os.Stat(store.lockPath()); err != nil {
+		t.Fatalf("the lock file did not survive a write: %v", err)
+	}
+}
+
+// The lock is a security boundary, so the path where it CANNOT be taken needs
+// coverage too: an operation that could not lock must report that, never proceed
+// unserialized and never report success. Replacing the credential directory with
+// a regular file makes the lock's MkdirAll fail on every platform.
+func TestOperationsFailWhenTheLockCannotBeAcquired(t *testing.T) {
+	base := t.TempDir()
+	dir := filepath.Join(base, "creds")
+	store := fileStore(t, dir)
+	// Where the store expects its directory, put a regular file.
+	if err := os.WriteFile(dir, []byte("not a directory"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := store.Set("openai", "sk-test"); err == nil {
+		t.Error("Set reported success although the lock could not be acquired")
+	}
+	if _, err := store.Delete("openai"); err == nil {
+		t.Error("Delete reported success although the lock could not be acquired")
+	}
+	if _, _, err := store.Get("openai"); err == nil {
+		t.Error("Get reported success although the lock could not be acquired")
+	}
+	if _, err := store.Providers(); err == nil {
+		t.Error("Providers reported success although the lock could not be acquired")
+	}
+	// And nothing may have been written through the failed path.
+	if raw, err := os.ReadFile(dir); err != nil || string(raw) != "not a directory" {
+		t.Errorf("the blocking file was modified: %q (err %v)", raw, err)
+	}
+}
--- a/internal/credstore/credstore.go
+++ b/internal/credstore/credstore.go
@@ -9,6 +9,7 @@ package credstore
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -125,14 +126,27 @@ func (s *Store) Backend() string { return s.backend }
 func (s *Store) Encrypted() bool { return s.backend == "keyring" || s.backend == "encrypted-file" }
 
 // Set stores (or overwrites) the API key for provider.
-func (s *Store) Set(provider, key string) error {
+func (s *Store) Set(provider, key string) (err error) {
 	provider = normalizeProvider(provider)
 	if provider == "" {
 		return fmt.Errorf("credstore: provider is required")
 	}
 	if s.backend == "keyring" {
+		// The OS keyring serializes its own writes; the file lock guards the
+		// file backends only.
 		return s.kr.Set(keyringService, keyringPrefix+provider, key)
 	}
+	// READ AND WRITE UNDER ONE LOCK. Separately locking each half would still
+	// lose keys: two writers would each read the same map, each add their own
+	// provider, and the second rename would publish a file missing the first.
+	// Measured before this existed: 1 of 100 concurrent Set calls survived.
+	release, err := s.acquireFileLock(true)
+	if err != nil {
+		return err
+	}
+	// Joined, not chosen between: a release failure annotates the result instead
+	// of masking the write error that actually explains what went wrong.
+	defer func() { err = errors.Join(err, release()) }()
 	data, err := s.read()
 	if err != nil {
 		return err
@@ -142,7 +156,7 @@ func (s *Store) Set(provider, key string) error {
 }
 
 // Get returns the API key for provider and whether one is stored.
-func (s *Store) Get(provider string) (string, bool, error) {
+func (s *Store) Get(provider string) (value string, found bool, err error) {
 	provider = normalizeProvider(provider)
 	if provider == "" {
 		return "", false, nil
@@ -150,6 +164,15 @@ func (s *Store) Get(provider string) (string, bool, error) {
 	if s.backend == "keyring" {
 		return s.kr.Get(keyringService, keyringPrefix+provider)
 	}
+	// A shared lock, so a read serializes against a writer's publish. On unix
+	// the rename is atomic and this only prevents reading an empty file mid-swap;
+	// on Windows an unsynchronized reader holding the file open would block the
+	// writer's rename outright, so readers must coordinate through the same lock.
+	release, err := s.acquireFileLock(false)
+	if err != nil {
+		return "", false, err
+	}
+	defer func() { err = errors.Join(err, release()) }()
 	data, err := s.read()
 	if err != nil {
 		return "", false, err
@@ -159,7 +182,7 @@ func (s *Store) Get(provider string) (string, bool, error) {
 }
 
 // Delete removes the API key for provider, reporting whether one existed.
-func (s *Store) Delete(provider string) (bool, error) {
+func (s *Store) Delete(provider string) (existed bool, err error) {
 	provider = normalizeProvider(provider)
 	if provider == "" {
 		return false, nil
@@ -167,6 +190,13 @@ func (s *Store) Delete(provider string) (bool, error) {
 	if s.backend == "keyring" {
 		return s.kr.Delete(keyringService, keyringPrefix+provider)
 	}
+	// The same lock as Set: a delete is a read-modify-write too, and one racing
+	// a set would otherwise resurrect or erase an unrelated provider.
+	release, err := s.acquireFileLock(true)
+	if err != nil {
+		return false, err
+	}
+	defer func() { err = errors.Join(err, release()) }()
 	data, err := s.read()
 	if err != nil {
 		return false, err
@@ -179,17 +209,22 @@ func (s *Store) Delete(provider string) (bool, error) {
 }
 
 // Providers lists providers with a stored key (sorted), for display/audit.
-func (s *Store) Providers() ([]string, error) {
+func (s *Store) Providers() (names []string, err error) {
 	if s.backend == "keyring" {
 		// The OS keyring has no enumerate-by-prefix in our minimal surface; callers
 		// that need a list pass known provider names to Get. Return empty here.
 		return nil, nil
 	}
+	release, err := s.acquireFileLock(false)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { err = errors.Join(err, release()) }()
 	data, err := s.read()
 	if err != nil {
 		return nil, err
 	}
-	names := make([]string, 0, len(data))
+	names = make([]string, 0, len(data))
 	for name := range data {
 		names = append(names, name)
 	}
@@ -263,6 +298,15 @@ func (s *Store) write(data map[string]string) error {
 	}
 	return nil
 }
+
+// lockPath is the advisory lock guarding a read-modify-write of the credential
+// file. Beside the data file, never the data file itself — write publishes by
+// rename and would carry the lock away with the old inode.
+func (s *Store) lockPath() string { return s.file + ".lock" }
+
+// filepathDir is filepath.Dir, named locally so the two platform lock files can
+// share it without either importing path/filepath for one call.
+func filepathDir(path string) string { return filepath.Dir(path) }
 
 func normalizeProvider(provider string) string {
 	return strings.ToLower(strings.TrimSpace(provider))
--- a/internal/credstore/filelock_unix.go
+++ b/internal/credstore/filelock_unix.go
@@ -1,0 +1,55 @@
+//go:build !windows
+
+package credstore
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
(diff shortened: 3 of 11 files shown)


gnanam1990:
@anandh8x this is the first of the 3-PR split you asked for on #829 — the independently-reversible hardening layer, with the durable-memory and orchestration PRs stacked behind it (opened in sequence as each lands, since a fork branch can't anchor an upstream base). Two questions before I open the next two:

1. **Shape** — is this the cut you had in mind (hardening lifted out first, then memory, then the orchestration remainder as one)? As noted on #829, the finer seven-way split needs the module-seam extraction first, so I kept this to the three layers that have genuinely independent build and rollback domains.

2. **Parent issue** — neither this PR nor #829 links an `issue-approved` parent. How would you like that handled — is there an existing approved issue this work should reference, or should one be filed (Discussions → Ideas) before I open the memory and orchestration PRs? I'd rather sort the gate now than have the stack stall on it.

Happy to adjust the boundaries if you'd cut them differently.

Vasanthdev2004 (changes requested):
Good change, and the package docs are the best part of it. The `pathjail` comment explaining why Lstat-then-act cannot work, and why a junction needs `ModeIrregular` rather than `ModeSymlink`, is real reasoning rather than decoration.

I checked the core claims instead of taking them:

- `os.Root` does refuse junction traversal (`path escapes from parent`).
- `RefuseReparse` does catch a junction as the final component.
- a junction does report `irregular=true, symlink=false`, so testing both bits is necessary rather than belt-and-braces.

I also went hunting for a `filepath.Rel` case-sensitivity bug on Windows and there isn't one: `Rel` uses `EqualFold` there, so an upper-cased root resolves fine. Scope is clean too. Against `origin/main` (`cabfeef`) this is exactly the 11 files and +900/-2 you describe. And every `s.write()` in credstore sits inside a lock-held function, so writer coverage is complete.

Three things worth fixing before it lands, all small.

### 1. The credstore lock is writers-only, and on Windows that makes locked writes fail

`Get` and `Providers` call `read()` with no lock. `write()` publishes with `os.Rename`, and Go's `os.Open` on Windows does not request `FILE_SHARE_DELETE`, so `MoveFileEx` cannot replace a file another opener still holds.

Measured cross-process at a 2ms cadence: 11 of 400 locked `Set` calls failed with `credstore: publish: ... Access is denied`, and roughly 7.5% of the reader's `Get` calls hit a sharing violation.

On unix this is fine, since POSIX rename does not care about open descriptors. So the two lock primitives are equivalent but the publish step around them is not: writers-only is sufficient on unix and insufficient on Windows.

Cheapest fix is to take the lock in `Get` and `Providers` too, shared if you want reader concurrency (`LOCK_SH`, or `LockFileEx` without `LOCKFILE_EXCLUSIVE_LOCK`).

### 2. `RefuseReparse` returns nil for a junction with a trailing separator

`os.Root.Lstat` on Windows resolves the terminal component when the path ends in a separator (`doInRootAlwaysResolveTerminalSlash` in `root_windows.go`). So `Lstat` on a junction name with a trailing separator stats the target, an ordinary directory, and the guard reports "not a link" for a name that is one.

```go
if trimmed := strings.TrimRight(relative, `/\`); trimmed != "" {
    relative = trimmed
}
```

before the Lstat, plus a case beside the existing junction test.

### 3. `CreateTemp` pastes prefix and suffix into the path unsanitized

Separators and `..` in either are path syntax rather than name text. A prefix of `..\sibling\note` puts the file outside the `dir` the doc promises. Worse, a suffix containing `..` erases the random component: `.tmp\..\..\evil` produced a file named literally `evil` at the jail root, so the unpredictability `O_EXCL` is there to protect is gone.

`filepath.Base` each of them, or reject a separator or `..` outright.

### Advisory, not blocking

- `Open` resolves `root` itself by pathname (`MkdirAll` then `OpenRoot`), so a junction planted at root or at an ancestor relocates the whole jail. Your doc calls this out as a deliberate trade-off, and the stores this replaces already create the same roots by pathname, so it is not a regression. Worth settling before splits 2 and 4 adopt it though, because "the boundary the caller chose and trusts" is a weaker claim when pathjail is the thing creating that boundary.
- `ModeIrregular` is set for every reparse tag except symlink, AF_UNIX and dedup. OneDrive placeholders and Store AppExecLink stubs fall in that set, so a store under a OneDrive-backed Documents folder would be refused outright. The stdlib carves out dedup for exactly this reason. Testing the reparse tag rather than the mode would avoid it.
- On Windows only the direct `git.exe` is killed on cancel, while POSIX kills the whole group. LFS filters, credential helpers and fsmonitor survive, and on Windows a survivor holding handles inside the worktree is what blocks removal. A job object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` is the analogue if you want it here, otherwise worth a line saying it is deferred.
- The cancellation hardening does not currently run at all: `exec.go:205`, `exec.go:232`, `workflows.go:87` and `workflows.go:177` all pass `context.Background()`, so the `Cancel` hook and the cancel branch of `WaitDelay` are unreachable from production.
- `GIT_DIR` and friends still pass through and take precedence over `command.Dir`. Not introduced here, since the previous code set no `Env` at all and this is strictly an improvement, but "consistent environments" invites the question and an explicit filter is a few lines.
- Tests: all four concurrency tests pin `Storage: "file"`, while `encrypted-file` is the default on every platform except macOS with a keyring, and it takes a different path through `securefile.Crypter` with its own nested lock. Also the junction test skips silently off Windows, where `linkDir` uses `os.Symlink`, so a predicate narrowed back to `ModeSymlink` alone would still pass there.
- The "byte-for-byte reconstruction" framing does not hold for `worktrees.go`. The `LC_ALL/LANG` line was re-resolved against main rather than replayed, so that file deserves fresh eyes rather than split-of-already-reviewed treatment.

### One to dismiss

CodeRabbit's `filelock_unix.go:38` "return lock-release errors" is not real. The payload is fsync'd and renamed before `release()` runs, and `file.Close()` drops the flock even if the explicit `LOCK_UN` errors, so a dropped release error cannot lose data. Returning it would only force `Set` and `Delete` to choose between the write error and a close diagnostic.

Happy to re-review quickly once 1 to 3 are in. The shape of this is right and I would like it landed.

anandh8x (changes requested):
I found issues that should be addressed before this is ready.

## Findings

- [P1] Preserve legitimate whitespace in confined paths
  `internal/pathjail/pathjail.go:52-54`

  `Open` trims both `root` and `dir`, so a legitimate path whose name begins or ends with a space is silently changed to a different filesystem object before `filepath.Rel`, `MkdirAll`, and `OpenRoot`. I reproduced this with a root named `"root "`: `Open` succeeded but created/opened `"root"` instead. Treat only an actually empty root as missing and preserve the caller's original path spelling.

- [P1] Keep generated temporary files inside the requested directory
  `internal/pathjail/pathjail.go:118-125`

  `CreateTemp` joins caller-controlled `prefix` and `suffix` without rejecting path components. A prefix such as `../escaped` causes the file to be created outside `dir` (though still under the broader `os.Root`). I reproduced this: a request under `intended` returned an `escaped.<random>.tmp` path in its parent. Reject separators/traversal in both prefix and suffix, or verify the resulting relative path remains beneath `dir`.

- [P2] Report lock-release failures instead of returning success
  `internal/credstore/filelock_unix.go:35-38`
  `internal/credstore/filelock_windows.go:35-38`
  `internal/credstore/credstore.go:142-152,183-199`

  Both release closures discard unlock and `Close` failures, and `Set` / `Delete` use a plain deferred release. A completed write can therefore be reported as fully successful even when the lock cleanup failed. Return an error from the release closure and join it into the operation result, at least when the read-modify-write otherwise succeeded.

- [P2] Make the Unix cancellation regression clean up on its failure path
  `internal/worktrees/run_git_unix_test.go:48-80`

  The shell program interpolates `pidFile` directly and the `grandchild == 0` branch calls `t.Skip` without cancelling and waiting for the already-started command. That can leak the shell/grandchild and also turns a failed security assertion into a skipped test. Pass the path as a positional shell argument; if PID discovery fails, cancel, wait for completion, then fail.

## Scope and validation

The split is otherwise focused and the architecture is sensible: the handle-relative path boundary is reusable, the credential lock covers the complete read-modify-write transaction, and git hardening is wired through one constructor.

All GitHub checks are green. Locally, the focused package tests and vet passed, and the three affected packages cross-compiled as Windows test binaries. I could not run `-race` in this environment because no C compiler is installed.

The requested test for replacing the credential directory with a regular file is worthwhile regression coverage, but current `Set` and `Delete` already propagate the `MkdirAll` lock-setup error, so I do not consider it a separate implementation blocker.

Non-blocking: the PR description says “split 1 of 3,” while the commit message says “1/4” and references “2/4 and 4/4.” Please make the split count consistent.

gnanam1990:
Thanks both — pushed `3079b06c` addressing the blocking findings. Each fix has a regression test; the strengthened ones are mutation-checked (revert the fix → the test fails).

### Fixed
- **`pathjail.CreateTemp` fragment injection** (@anandh8x P1, @Vasanthdev2004 #3) — `prefix`/`suffix` are now rejected up front if they carry a separator or `.`/`..`. The regression test uses `Open(root, subdir)` so it exercises the real case os.Root does *not* backstop: a `../escaped` fragment landing in `root` but outside the promised `dir`. Without the fix that returns success and leaks the file into the parent; the test asserts nothing escapes.
- **`pathjail.Open` whitespace trimming** (@anandh8x P1) — only an empty/whitespace-only value is treated as missing; the path is otherwise used verbatim, so `"root "` is confined as spelled. Test asserts the trimmed `"root"` is never created.
- **`pathjail.RefuseReparse` trailing separator** (@Vasanthdev2004 #2) — `strings.TrimRight(relative, "/\\")` before `Lstat`, so a junction named with a trailing slash is inspected as itself rather than resolved to its target. Test added beside the junction case.
- **`credstore` readers need the lock** (@Vasanthdev2004 #1) — `acquireFileLock` now takes `exclusive bool`; writers take it exclusive, `Get`/`Providers` take it **shared** (`LOCK_SH` / `LockFileEx` without `LOCKFILE_EXCLUSIVE_LOCK`), so a reader can't race a publish — which is what bites on Windows, where an unsynchronized reader holding the file open blocks the rename.
- **Worktrees cancellation test** (@anandh8x P2) — `pidFile` is passed as a positional `$1` instead of interpolated, and a missing pid now cancels, waits for the command, and `t.Fatal`s instead of leaking the shell and silently skipping the assertion.
- **Split numbering** — commit now says 1/3 (2/3, 3/3), matching the description.

### One I'd push back on
- **Returning lock-release errors** (@anandh8x P2 / CodeRabbit): I'm with @Vasanthdev2004's "one to dismiss" here. The payload is fsync'd and renamed *before* `release()` runs, and `file.Close()` drops the flock even if the explicit `LOCK_UN`/`UnlockFileEx` errors — so a release hiccup can't lose data or leave the store inconsistent, and folding it into the result would report an otherwise-successful write as failed. Happy to surface it as a non-fatal diagnostic if you want visibility, but I don't think it should change the returned error.

### Advisory — noted, not in this PR
- **`ModeIrregular` refusing OneDrive/AppExecLink placeholders** (@Vasanthdev2004): real, and the right fix inspects the reparse *tag* (carving out dedup as the stdlib does) rather than the mode — Windows-syscall work worth its own focused change.
- **`GIT_DIR` and friends passthrough**: not introduced here (the prior code set no `Env` at all), and an explicit filter is a behavior change I'd rather not slip into a hardening PR speculatively.
- **Windows job-object kill-on-close** as the process-group analogue: deferred; the Unix group-kill is what's wired here.
- **Cancellation hook unreachable** (`context.Background()` at `exec.go`/`workflows.go`): those files live in the orchestration split (3/3), not this PR — I'll wire the cancellable context there.
- **Encrypted-file concurrency coverage**: the concurrency tests pin `Storage:"file"`; broadening to `encrypted-file` (the `securefile.Crypter` path) is worthwhile and I'll add it as a follow-up.

I also gave `worktrees.go`'s `LC_ALL=C`/`LANG=C` line fresh eyes per your note — it's correct: forced C locale appended to the full environment for stable git-output parsing, strictly more than the prior code which set no `Env`.

Verified: `gofmt`, `go vet` (unix + windows), `go build ./...`, `go test -race` on all three packages (a C compiler is present here, so the race check ran), and Windows test binaries compile.

Vasanthdev2004 (changes requested):
Re-reviewed at `3079b06c`. All three of my blocking findings are genuinely fixed, and I checked each rather than taking the summary:

1. **credstore readers.** `acquireFileLock(exclusive bool)`, writers `true`, `Get` and `Providers` `false`, `LOCK_SH` on unix and `LockFileEx` without `LOCKFILE_EXCLUSIVE_LOCK` on Windows, both held under `defer release()` across the read. That closes it properly: the writer now waits for readers to drop the lock instead of racing their open handle into its rename.
2. **`RefuseReparse` trailing separator.** `TrimRight` before the Lstat, with the root case guarded so `"."` survives, and a test beside the junction one.
3. **`CreateTemp` fragments.** Rejected up front, and I like that the test uses `Open(root, subdir)` so it exercises the case `os.Root` does not backstop, where the fragment lands inside the root but outside the promised `dir`. That is the version that actually proves something.

Agreed on the pushback about lock-release errors, for the reason you give.

The deferrals look right too. The `ModeIrregular` tag work, the `GIT_DIR` filter and the job object are each their own change, and wiring the cancellable context in split 3 is the correct home for it since that is where the call sites live.

One new thing, and it is the only reason this is not an approve.

### `TestOpenPreservesLeadingAndTrailingSpaceInPaths` cannot hold on Windows

Smoke (windows-latest) is red on this head; macOS and ubuntu are green. The assertion is:

```go
if _, err := os.Stat(filepath.Join(base, "root")); err == nil {
    t.Error(`a trimmed "root" was created instead of the requested "root "`)
}
```

Win32 strips trailing spaces and dots from path components, so `"root "` and `"root"` are not two objects there, they are one. I ran it to be sure:

```
os.MkdirAll(<tmp>/"root ")  ->  created on disk: "root"
stat("root ") err=<nil>
stat("root")  err=<nil>
```

So the directory that appears is `root`, the trimmed stat succeeds, and the test reports a trim that never happened. The premise behind the finding is true on unix and simply not expressible on Windows.

Your fix is fine and I would keep it. It is the test that needs the gate: either `if runtime.GOOS == "windows" { t.Skip(...) }` with a line saying why, or keep the `handle.Name()` half everywhere and run only the trimmed-stat half off Windows. Worth a sentence about Win32 component normalization so nobody re-adds it later.

Worth naming the shape as well, since this stack has two more splits coming: this is a case only a Windows machine can see, and you are on macOS. The same will apply to anything touching path semantics in 2/3 and 3/3, so it may be worth me running those on Windows before they go up rather than after CI says so.

Everything else is green here: `go build ./...`, and `internal/credstore` and `internal/worktrees` both pass on Windows.

anandh8x (changes requested):
Re-reviewed the latest head (3079b06c). The earlier traversal, credential-locking, and Unix cancellation blockers are substantially fixed, but this is not ready to approve yet.

## Findings

- **[P1] Make the whitespace regression portable on Windows** — `internal/pathjail/pathjail_test.go:157-170`

  Windows smoke is red because the test assumes `"root "` and `"root"` are distinct filesystem objects. Win32 normalizes the trailing-space component, so the assertion fails even though `Open` no longer trims the supplied path. Keep the production fix, but test a portable leading-space component by creating a child through the returned handle and verifying it at the expected path.

- **[P2] Preserve whitespace-only paths too** — `internal/pathjail/pathjail.go:53-61`

  `Open` still uses `strings.TrimSpace(root) == ""` and the same check for `dir`. That means a legitimate whitespace-only path is rejected as a missing root, or silently replaced with `root` for `dir`. This is the remaining part of the original retargeting bug. Check `root == ""` and `dir == ""` while otherwise preserving the exact caller input.

- **[P2] Exercise the exclusive-create collision path** — `internal/pathjail/pathjail_test.go:130-152`

  The implementation correctly uses `O_EXCL`, but the current uniqueness loop never forces a collision. A truncating implementation with deterministic distinct names could still pass. Add a controllable random source for the internal helper, pre-create the first generated name, then verify the existing file is unchanged and creation retries with a different name. This is a regression-test gap rather than evidence that the current production path is broken.

## Validation

Locally, the three focused packages passed three repeated test runs and `go vet`; all three also cross-compiled as Windows test binaries. Native Windows smoke currently fails at `TestOpenPreservesLeadingAndTrailingSpaceInPaths`. The branch is also one commit behind current `main`, although that commit does not overlap these files.

gnanam1990:
Pushed `f94d32e8`. All three findings fixed, rebased onto current `main`, and the Windows smoke failure is mine — thanks for the precise diagnosis.

### Fixed
- **[P1] Windows-portable whitespace test.** You were both right about the cause: Win32 normalizes a component's trailing spaces away, so `"root "` and `"root"` are one object there and the assertion could never hold. Replaced with a **leading**-space component, which survives normalization on every platform, so the test runs everywhere rather than being skipped on the one OS where path semantics are most interesting. It also now writes a file *through the handle* and stats it at the expected spelling, so it proves the bytes land where the caller asked rather than just checking `handle.Name()`.
- **[P2] Whitespace-only paths preserved.** `Open` now tests `root == ""` / `dir == ""` exactly. Worth flagging what this caught: my first attempt at the regression test used `filepath.Join(base, " ")`, and it **passed against the old `TrimSpace` guard** — an absolute path ending in a space still has a non-empty `TrimSpace`, so it never exercised the bug. The path has to be whitespace-only *in full*, i.e. relative, so the test now uses `t.Chdir` + `Open(" ", " ")`. That version fails with the old guard (`no root to confine " " to`) and passes with the fix.
- **[P2] O_EXCL collision path exercised.** Added a `randomBytes` seam (a package var over `rand.Read`) so a test can force the collision a real random source will not produce on demand. The test pre-creates the first generated name, then asserts CreateTemp does **not** return it, the existing file's contents are untouched, and the source was drawn more than once so the retry genuinely ran. Dropping `O_EXCL` from the production call makes it fail.

Each fix is mutation-verified — I reverted each guard in turn and confirmed the corresponding test fails, which is how I found the vacuous whitespace test above rather than shipping it.

### Also
- Rebased onto current `main` (`2d2450e9`), so the branch is no longer behind.
- Verified: `gofmt`, `go vet` (unix + windows), `go build ./...`, `go test -race -count=3` on all three packages, and Windows test binaries compile.

@Vasanthdev2004 — on your offer to run 2/3 and 3/3 on Windows before they go up: yes please, and I think it's the right call. This failure is exactly the shape you named — a case only a Windows machine can see, from a macOS author — and both remaining splits touch path semantics (the memory store and the plan store are the two `pathjail` adopters). I'd rather find those on your machine than in CI. I'll ping you when 2/3 is ready rather than pushing it blind.

Vasanthdev2004 (dismissed):
Approving at `f94d32e8`. Ran it on Windows, which is the machine that mattered here.

**All three packages pass**: `internal/pathjail`, `internal/credstore`, `internal/worktrees`, plus `go build ./...` and `gofmt` clean. Merge base is exactly `origin/main` (`2d2450e9`), so the rebase is real.

The leading-space substitution is the right call and it does hold on Windows. A leading space survives component normalization there, unlike a trailing one, so the test runs on all three platforms instead of skipping on the one where path semantics are most interesting. Writing through the handle and stating the file at the expected spelling is a stronger assertion than `handle.Name()` too.

I also mutation-checked two of your guards rather than taking the summary:

- Dropping `O_EXCL` from the production create fails `TestCreateTempRetriesWithoutClobberingAnExistingName`. The `randomBytes` seam does its job.
- Restoring the old `TrimSpace` guard does **not** fail `TestOpenPreservesSpaceInPaths`.

That second one is worth passing back, because it is the same vacuity you caught in your own first attempt, surviving in the sibling test. `TrimSpace` only touches whitespace at the ends of the whole string, and `filepath.Join(base, " root")` produces an absolute path beginning with a drive letter, so the guard is a no-op against it on every platform. I confirmed the mutation actually landed before concluding that, since a revert that silently fails to apply looks identical to a passing test:

```
root = strings.TrimSpace(root)
dir = strings.TrimSpace(dir)
if root == "" {
...
--- PASS: TestOpenPreservesSpaceInPaths
```

So the sensitive test for that fix is `TestOpenAcceptsAWhitespaceOnlyComponent`, and it is skipped on Windows, which leaves the whitespace guard with no Windows coverage at all. The fix itself is correct and I am not blocking on this. If you want the coverage, a relative `" root"` under `t.Chdir` would bite on both platforms, since the whitespace is then at the string boundary where `TrimSpace` can reach it.

Not blocking, just noting it so the next reader does not trust the comment more than the assertion.

Good round. Three iterations, each finding fixed with a regression rather than a patch, and you found one of your own vacuous tests on the way. Taking you up on running 2/3 and 3/3 here before they go up, both adopters are path-semantics code and this is the machine that sees it.

anandh8x (changes requested):
Re-reviewed latest head `f94d32e8`. The previous three blockers were addressed correctly, and all GitHub checks are green, including Windows. I found four remaining issues before approval.

## Findings

- **[P1] Preserve valid trailing backslashes on Unix** — `internal/pathjail/pathjail.go:101-108`

  `RefuseReparse` unconditionally applies `strings.TrimRight(relative, `/\\`)`. On Unix, backslash is a valid filename character rather than a separator, so this inspects a different final component. I reproduced it with a symlink named `linked\\`: `RefuseReparse` inspected `linked` and returned `nil` instead of `ErrReparse`. Strip only separators recognized by the active platform, with the Windows-specific dual-separator behavior kept behind a small platform check, and add the Unix regression.

- **[P2] Propagate lock-release failures** — `internal/credstore/filelock_unix.go:41-44`, `internal/credstore/filelock_windows.go:42-45`, `internal/credstore/credstore.go:142-229`

  Both release closures still discard unlock and `Close` errors, and all four file-backed operations defer them without observing the result. This can report success after incomplete cleanup and conflicts with the repository rule that cleanup or unlock failure must not be reported as success. Return an error from the release function and join it into the result from `Set`, `Get`, `Delete`, and `Providers`.

- **[P2] Cover lock-acquisition failure paths** — `internal/credstore/concurrency_test.go`

  The tests cover successful contention but not lock setup failure. Add a cross-platform regression that replaces the configured credential directory with a regular file and verifies both `Set` and `Delete` return the acquisition error. This is required failure-path coverage for the new security boundary.

- **[P2] Make the portable whitespace test mutation-sensitive** — `internal/pathjail/pathjail_test.go:163-188`

  `TestOpenPreservesSpaceInPaths` uses `filepath.Join(base, " root")`, so the full absolute string begins with the temp-directory prefix. Restoring the old `TrimSpace` behavior therefore leaves the path unchanged and the test still passes. After `t.Chdir`, use a relative `" root"` so the leading whitespace is at the string boundary on every platform; then verify the handle-created child lands under that exact spelling.

## Validation

Focused `-race` tests for `internal/pathjail`, `internal/credstore`, and `internal/worktrees` passed, as did focused vet, formatting, repository-wide vet, release build, and release smoke. All GitHub checks are green. The full local suite reached unrelated environment-dependent sandbox-policy failures in `internal/agent` and `internal/cli`; the affected PR packages passed.

gnanam1990:
Pushed `7193cac7` — all four addressed.

### Fixed
- **[P1] Unix trailing backslashes.** Good catch, and the failure mode is exactly as you describe: a symlink named `linked\` was inspected as `linked` and came back `nil`. Replaced the literal `` TrimRight(relative, `/\`) `` with a helper over `os.IsPathSeparator`, which is platform-correct by definition — both separators on Windows, `/` only on unix. Kept as one cross-platform function with the small conditional inside rather than a build-tagged pair, per the repo's preference for unified behavior. Regression added (unix-gated, since the name isn't expressible on Windows); reverting to the unconditional trim fails it.

- **[P2] Vacuous whitespace test.** @Vasanthdev2004 caught the same thing independently, and you're both right — it's the identical vacuity I found in my own first attempt, surviving in the sibling test. `filepath.Join(base, " root")` puts the space mid-string where `TrimSpace` can't reach it, so the guard was a no-op against it on every platform. Now `t.Chdir(base)` + a relative `" root"`, so the whitespace sits at the string boundary. It now fails against the restored `TrimSpace` guard on both counts (the handle-created child isn't under the requested spelling, and a trimmed `"root"` appears), which also gives the whitespace fix its first **Windows** coverage — previously the only sensitive test was the unix-skipped one.

- **[P2] Lock-release failures propagated.** You and @Vasanthdev2004 landed on opposite sides of this, so I took the reading that satisfies both concerns rather than picking a side: the release closure now returns an error, and each of `Set`, `Get`, `Delete`, `Providers` folds it in with `errors.Join` via a named return. That meets your rule — a cleanup that didn't complete is no longer reported as success — while avoiding the objection that motivated my earlier pushback, since joining *annotates* the result instead of forcing a choice between the write error and a close diagnostic. If the write fails and the release also fails, the caller sees both.

- **[P2] Lock-acquisition failure coverage.** Added `TestOperationsFailWhenTheLockCannotBeAcquired`: it replaces the credential directory with a regular file (which fails the lock's `MkdirAll` on every platform) and asserts all four operations return an error rather than proceeding unserialized — plus that the blocking file is untouched, so nothing was written through the failed path.

Every guard here is mutation-verified; that's how the vacuous test above got caught rather than shipped a second time.

### Validation
`gofmt`, `go vet` (unix + windows), `go build ./...`, `go test -race -count=3` on all three packages, Windows test binaries compile, `zero-release build` + `smoke`, `git diff --check`. Branch remains on current `main`.

@Vasanthdev2004 — thanks for running it on Windows and for mutation-checking rather than trusting my summary; that's what turned up the dead assertion. The offer to run 2/3 and 3/3 on Windows before they go up still stands as far as I'm concerned, and I'll take you up on it.

Vasanthdev2004 (approved):
Approving again at `7193cac7`. Ran it on Windows.

`go build ./...`, `go vet`, `gofmt` clean, and `internal/pathjail`, `internal/credstore` and `internal/worktrees` all pass. Merge base is still exactly `origin/main` (`2d2450e9`).

**The vacuous whitespace test is genuinely fixed, and I checked it the same way as before.** Restoring the old `TrimSpace` guard now FAILS `TestOpenPreservesSpaceInPaths`, where on the previous head it passed. I confirmed the mutation landed before drawing that conclusion. Moving the path under `t.Chdir` so the space sits at the start of the whole string is the right move: that is the only position `TrimSpace` can reach, which is precisely why the absolute-path version could never bite. The comment you left on it says exactly that, which is what stops someone reinstating the weaker form later.

**On the backslash finding, credit where it is due: @anandh8x caught a bug I introduced.** The `` TrimRight(relative, `/\`) `` in my previous review is what put it there. Backslash is a separator only on Windows, so on unix that trims a legal filename character and inspects `linked` when asked about a symlink genuinely named `linked\`, reporting "not a link" for one that is. Routing through `os.IsPathSeparator` is correct by construction rather than by enumeration, and keeping it as one function with the conditional inside is the right shape.

Being explicit about what I did and did not verify: I confirmed the helper's shape, that Windows behaviour is unchanged, and that the new regression is present and correctly gated. I did **not** execute the unix path, since this machine cannot run it, so that half rests on @anandh8x's review and on CI rather than on me.

The lock-release change reads well too. Reporting the unlock error while keeping `Close` unconditional means a cleanup that did not complete is no longer indistinguishable from one that did, and joining it at the callers annotates a successful write rather than masking it. Better than either of the two positions we were arguing between.

Four rounds, every finding closed with a regression that fails without the fix, and two vacuous tests caught along the way (one by you, one by me, the same shape both times). Good to land once @anandh8x clears his outstanding review; his findings look addressed to me, but the approval is his to give.

anandh8x (approved):
Re-reviewed latest head `7193cac`. The previously requested fixes are correct: Unix backslash names are preserved, lock release errors are propagated, lock-acquisition failures are covered, and the whitespace regression is now mutation-sensitive. Focused race tests, vet, Windows cross-compilation, formatting, build, smoke, and all GitHub checks pass across Linux, macOS, and Windows. The deterministic release-failure regression remains worthwhile follow-up coverage, but it is not evidence of a production defect and is non-blocking.

Vasanthdev2004:
@kevincodex1 this one is ready from our side and worth merging ahead of the queue, but GitHub reports it as blocked and I do not have the rights to see why.

**State at `7193cac7`:** approved by @anandh8x, @coderabbitai and me, base is exactly current `main` (`2d2450e9`), and every check that reported is green, including Smoke on all three platforms.

**What is holding it:** `mergeStateStatus` is `BLOCKED`. The only check not green is `CodeRabbit`, which has neither a status nor a conclusion, so it looks like it was created and never reported rather than failed. Branch protection reads 404 for me (`maintain`, not admin), so I cannot confirm whether that check is required. If it is, it probably needs a re-run or an admin merge rather than anything from the author.

**Why it is worth doing first:** it is 1 of 3 in a stack. #897 sits directly on top of it and #829's orchestration remainder behind that, so merging this unblocks a chain rather than a single PR. It also carries the `internal/pathjail` primitive that the next two adopt, so the longer it sits the more rebasing the stack accumulates.

Four rounds of review on it. Every finding was closed with a regression that fails without its fix, and I ran the last two heads on Windows myself since the author is on macOS and the path semantics only misbehave here.

No rush from me if you would rather wait on the CodeRabbit check, just flagging that it does not look like it is going to report on its own.

Vasanthdev2004:
@kevincodex1 correcting myself: the CodeRabbit check is **not** what is blocking this, and my message above was wrong about the cause.

I checked it properly instead of inferring. That check is pending on all eight open PRs, and it is also pending on **#890 and #887, both of which merged**. So it is not a required check and it does not block anything.

What I can still say, from outside branch protection:

- `#891` is APPROVED by @anandh8x, @coderabbitai and me, base is exactly current `main` (`2d2450e9`), and every check that reports is green.
- It is the only one of the eight where `BLOCKED` is not explained by its reviews. The other seven are either CHANGES_REQUESTED or awaiting a first approval, and `#812` reads CLEAN because its base is a feature branch rather than `main`.

So something in `main`'s protection is unsatisfied that I cannot see. Reading it 404s for me (`maintain`, not admin), so this needs someone with admin to say what. The candidates I would check first are a required review from a specific person or team, or a required status context that no workflow produces at all, which would not show up in the checks list.

Sorry for the noise. The ask is unchanged and the reason to prioritise it stands: 1 of 3, with #897 approved and stacked directly on it, and #829's remainder behind that.

kevincodex1 (approved):

PR #892, opened 2026-08-12
Define provider identity primitives and persisted-name validation (1/4)

PierrunoYT:
## Summary

This is **PR 1 of a 4-PR split of #725**, opened in response to [review feedback](https://github.com/Gitlawb/zero/pull/725#pullrequestreview-4911380160) that the combined branch changed four distinct contracts at once and could not be reviewed against a stable baseline.

It defines the provider identity contract **and wires every consumer boundary it touches onto that contract**, so no caller is left on the old mental model. It does **not** implement cross-process locking or the full writer-inventory transaction — that stays with #894.

> [!NOTE]
> #893 (2/4) is stacked on this PR and should merge after it.

### The contract

Persisted provider rows and credential-store entries answer two different questions, and mixing them let one profile's mutation reach another profile's row and secret.

- **Credential identity** — `credstore.NormalizeProvider` (trim + `ToLower`), exposed as `config.SameProviderIdentity`. Callers deciding whether two spellings share one stored secret use it rather than `strings.EqualFold`: Unicode case folding equates `s` and `ſ` while `strings.ToLower` does not, so an `EqualFold` comparison can promise a survivor access to a key it can never look up.
- **Exact row identity** — trimmed `provider.Name` equality, used by every row-targeting mutator (`MarkProviderAPIKeyStored`, `ProviderPersisted`'s mutating callers, `SetProviderModel`, `ClearProviderKeyStored`, `RemoveProvider`, and the `oldName` lookups in `RenameProvider`/`EditProvider`).
- **Publication validation** — `ValidatePersistedProviderNames` rejects persisted rows repeating a folded identity, whether the spellings are identical or only case variants. `writeConfigFile` guards every write with it, and `Resolve()` validates user config before merging.

### The four shared helpers (the "front door")

Defining two rules is not enough — every boundary that gated on one rule and mutated with the other needed one place to cross between them:

| Helper | What it owns |
|---|---|
| `ResolvePersistedProviderName(path, input)` | User/session spelling → the exact persisted row. Exact wins, credential identity is a fallback, an identity matching 2+ rows is an **error**, not an arbitrary pick. `SetActiveProvider`'s loop was extracted into it rather than copied a fourth time. |
[...]

PR #725, opened 2026-07-18
Clarify saved and effective provider selection

PierrunoYT:
## Summary
- report when `ZERO_PROVIDER` overrides a provider saved by `providers use`, including resolved, deferred, unresolvable, and unrelated config-error outcomes
- expose whether provider-list entries are selectable from user config or come from the fully resolved runtime configuration
- use exact persisted names for config mutations while using the credential store's normalized identity consistently for collision, cleanup, and session decisions
- centralize OAuth/API-key credential candidates across status, refresh, logout, CLI removal, and TUI key cleanup without deleting shared catalog credentials
- require positive `catalogId` ownership for catalog setup/login/persistence and reject ambiguous shared catalog IDs instead of adopting by name or file order
- keep TUI live provider state and `ZERO_PROVIDER` synchronized for case-only renames without retargeting case-variant project providers

## Validation
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go run ./cmd/zero-release build`
- `go run ./cmd/zero-release smoke`
- `make lint-static` (0 issues)
- `make vulncheck` (no vulnerabilities found)
- `git diff HEAD --check`

Fixes #721

PR #893, opened 2026-08-12
Resolve catalog ownership and credential candidates (2/4)

PierrunoYT:
## Summary

This is **PR 2 of a 4-PR split of #725**, following the [review feedback](https://github.com/Gitlawb/zero/pull/725#pullrequestreview-4911380160) that asked for the combined branch to be replaced by focused PRs in dependency order.

> [!IMPORTANT]
> **This PR is stacked on #892 and should be merged after it.**
> I don't have push access to this repository, so the base branch of #892 can't exist here and GitHub forces this PR to target `main`. That means the diff shown below *includes* #892's commit.
> **Review only the second commit — `feat(providers): resolve catalog ownership and credential candidates`.** Once #892 merges, this PR's diff will collapse to that commit alone.

## What this PR establishes

Builds on #892's identity primitive to answer a question the old code answered by accident: **which persisted provider row owns a given catalog identity**, and **which credential-store entries a provider command may touch**.

### Positive catalog ownership

`EnsureCatalogProvider` previously scanned rows with an `EqualFold` name-or-catalog match, so it could adopt a row by name alone that had nothing to do with the catalog entry being set up, or silently pick the first of several rows sharing one catalog ID.

- `providerOwnsCatalog` / `catalogProviderOwner` require **positive** ownership — a row must actually carry the catalog ID to be adopted for it.
- Ambiguity is now an error rather than a file-order coin flip: several rows sharing a catalog ID is rejected instead of adopted.
- `PreflightCatalogProviderLogin` applies that check *before* a login burns a browser round trip.

### One read-only credential-candidate resolver

`ProviderCredentialCandidates` is the single place that answers "which credential-store entries does this provider address resolve to." `PersistedIdentityMatch` / `ResolvePersistedProviderIdentity` give it exact-name precedence over catalog-ID matching, and `CatalogIdentityExclusive` reports whether a catalog ID is exclusively owned.

The OAuth surfaces are migrated onto it together, as the review requested — **status, refresh, and logout**. API-key cleanup remains row-name-scoped because stored API keys are read only by persisted profile name:

- `runAuthStatus` / `runAuthRefresh` / `runAuthRefreshWatch` / `filterAuthStatuses` are candidate-list based rather than single-key based.
- `runAuthLogout` expands candidates only in the OAuth token store, deletes an API key only for the addressed persisted row name, clears markers via `ClearProviderKeyStoredCaseVariants`, and now: rejects an ambiguous catalog address, prefers an exactly-named profile, **leaves shared catalog credentials alone**, and fails closed before changing credentials when config validation fails.
- The TUI wizard's key-removal branch (`applyManageKeyChoice`) deletes only the persisted row-name entry from the credential store beside the config being edited.

### Preflight before irreversible side effects

`runAuthLogin`, `runAuthChatGPT`, `runAuthOpenRouter`, and the TUI/onboarding OAuth and device-code paths now validate config before starting a flow. `runAuthChatGPT` re-checks *immediately before token save* to close the window between flow start and persistence — wired through a new `BeforeSave` hook on `oauth.Manager` (invoked in `Login` and `CompleteDeviceLogin`).

Since `persistOAuthLoginProvider` can now legitimately fail on an ambiguous or unowned catalog ID, `applySetupOAuth` turns a previously discarded `_ =` error into a real path that surfaces to the user instead of silently dropping a completed login.

### Exit-code fix

`zero auth openrouter` previously returned **0** when it minted a key but failed to save it — reporting success for a command that left the provider unusable, which a script would happily carry on from. It now still prints the minted key (the user paid for it with a browser round trip) but exits non-zero with a clear error.

## Deliberately out of scope

No locking, no `CommitProviderProfile`, no provider-selection presentation. **`saveOpenRouterProviderKey` is byte-for-byte unchanged in this PR** (verified) — its transaction fix is one of the two P1 findings and belongs to PR 3, so it stays reviewable there rather than being smuggled in here.

Remaining PRs:
3. #894 — **Provider config/key transaction** — one authoritative operation owning lock acquisition, config read/validation, credential capture, atomic publication, and conditional rollback across the full writer inventory. Carries both P1 fixes: lock acquisition must fail closed, and OpenRouter persistence must join the transaction.
4. #895 — **Provider-selection UX and TUI synchronization** — list/current/use source and selectability output, `ZERO_PROVIDER` override explanations, case-only live-session sync.

> [!NOTE]
> #894 (3/4) is stacked on this PR and should merge after it.

## Pre-existing tests adjusted

Both in `internal/tui/provider_wizard_test.go`, both required by the in-scope behavior change:

- **`TestWizardProviderStoredKey`** — `wizardProviderStoredKey` now returns an error and requires positive ownership, so the `{Name: "nokey"}` row gained a `CatalogID`, name-only-match assertions became ownership-rejection assertions, and exact-owner-wins / shared-owner-ambiguity cases were added.
- **`TestProviderWizardManageKeyRemove`** — the Remove branch deletes only the persisted row-name key; the fixture also stores a catalog-alias key and verifies that unrelated entry survives.

## Validation

- `go build ./...`
- `go vet ./...`
- `gofmt -l .` (clean)
- `go test ./...` (full suite green — 84 packages, 0 failures)

Refs #721. Split of #725. Stacked on #892.

## Summary by CodeRabbit

- **New Features**
  - Added ChatGPT OAuth login support in the CLI and setup wizard.
  - Added `providers repair-config` to recover legacy provider configurations.
  - Added validation before OAuth authentication and credential storage.
  - Improved provider recognition using catalog identities and normalized names.

- **Bug Fixes**
  - Prevented ambiguous or duplicate provider configurations from being selected.
  - Improved provider use, rename, removal, logout, refresh, and status handling.
  - Credential save failures now preserve existing credentials and display errors.
  - Improved cleanup and retention of shared credentials across provider variants.
  - Doctor reports now surface configuration resolution failures.


jatmn:
blocked until #892 lands

PierrunoYT:
PierrunoYT pushed `2d366a93` to close the remaining PR2 review gaps:

- folded-name identity ambiguity now fails closed with zero credential candidates, while config read/JSON failures retain only the explicitly requested candidate;
- OAuth `BeforeSave` failure tests preserve the prior token for the same provider;
- key-removal tests assert persisted key markers for exclusive and shared aliases;
- the legacy AIMLAPI test asserts `BaseURL` preservation;
- provider-write preflight reuses the persisted-provider parser.

The existing empty-candidate guard/test and ambiguous folded-name removal with exact-name repair were also re-verified.

Validation passed: focused race tests for config/OAuth/CLI/TUI, `make fmt-check`, `go vet ./...`, `go test ./...`, release build and smoke, `make lint-static` (0 issues), and diff hygiene. `make vulncheck` still reports known dependency findings: GO-2026-6222 in `golang.org/x/image v0.44.0` (fixed by v0.45.0 already on upstream `main`) and GO-2026-6115 in `github.com/ledongthuc/pdf` (no fixed version published).

Current blocker: this PR remains stacked on open PR #892 and conflicts with current upstream `main` in `internal/cli/app.go` and `internal/credstore/credstore.go`; no merge or conflict resolution is included here.

PierrunoYT:
Addressed all three current unresolved findings in `e280e7b4`.

### One authoritative ambiguity contract
`ResolvePersistedProviderIdentity` now returns a distinct `PersistedIdentityAmbiguous` match for both:

- non-exact folded names matching multiple rows (`work` / `WORK` addressed as `wOrK`);
- catalog IDs claimed by multiple profiles.

Exact row spelling still wins, preserving the supported legacy repair path. `ProviderCredentialCandidates` consumes that state and returns **nil candidates** for ambiguity, while retaining the requested-spelling fallback only for config read/JSON errors. This removes the duplicate catalog ambiguity scan and keeps one source of truth.

### Fail-closed command behavior
- Ambiguous provider removal now proves complete config bytes and stored credentials remain unchanged before exercising the exact-spelling repair control.
- Empty credential candidates are covered in both normal `auth refresh` and `auth refresh --watch`; both return a redacted app error with no success output.

### Verification
Passed:
- focused config and CLI tests;
- `go vet ./...`;
- `go test ./...`;
- release build and smoke;
- gofmt and `git diff --check`.

Two repository-wide checks remain red because this independently based branch is 15 upstream commits behind, outside this review diff:
- staticcheck finds the pre-existing capitalized Windows error in `internal/peermsg/private_dir_windows.go`;
- govulncheck flags `golang.org/x/image v0.44.0`; upstream `main` and PR #892 already use fixed v0.45.0.

I did not bundle either unrelated baseline change into this focused PR. Please re-review the current findings and advise whether you want the branch rebased/refreshed separately before merge.

PierrunoYT:
Rebased this PR onto current `upstream/main` (`6edf9a8b`) and force-pushed the refreshed series at `f47133bc`.

Conflict resolution preserved upstream's expanded `appDeps` surface and added this PR's ChatGPT login dependency in the same field/default/fill paths. The branch is now current and the two stale-base validation failures mentioned in my previous comment are gone:

- staticcheck/unused/ineffassign: **0 issues**
- govulncheck: **no vulnerabilities found**

Post-rebase verification also passes:
- focused config and CLI tests;
- `go vet ./...`;
- `go test ./...`;
- release build and smoke;
- repository-wide gofmt check;
- `git diff HEAD --check`.

Please review the rebased head `f47133bc`.

PierrunoYT:
Addressed the latest CodeRabbit findings in c408b5f6: exact duplicate persisted names now fail closed before credential selection; nameless singleton provider-command output selects the effective OpenAI identity; exact legacy-row removal preserves a credential still claimed by a case-variant survivor; and secret values are no longer included in assertion output. Added focused coverage for duplicate exact names, nameless/ambiguous provider-command results, and post-removal credential retention. Validation passed: gofmt, go vet ./..., go test ./..., release build + smoke, staticcheck/unused/ineffassign, govulncheck, and git diff --check.

PierrunoYT:
The remaining removal race is valid, but it cannot be fixed safely with a CLI-only lock or a second config check. Provider removal must serialize canonical-name resolution, config publication, ownership retention, and credential deletion against **every** provider writer, including TUI and setup flows. That shared cross-process config/key transaction boundary is introduced by #894.

The safe plan is therefore to stack/rebase this PR onto #894 (or otherwise reorder the series) and route removal through that shared transaction. Backporting only the removal side would still race with writers that do not acquire the same lock, while backporting the whole transaction subsystem would duplicate #894 and substantially widen this PR. The review thread is intentionally left unresolved pending agreement on that stacking plan.

PierrunoYT:
Restacked this branch onto the current #892 head (`2b8faf39`) as the prerequisite for rebuilding #894. New #893 head: `33944034`.

The conflict resolution preserves #892's current provider-identity/repair behavior and reapplies this slice's catalog ownership and credential-candidate contract. Focused config, CLI, OAuth, and TUI suites passed before #894 was replayed on top. #894 now has this exact head as an ancestor.

Vasanthdev2004 (changes requested):
First review from me on this one. I have now looked at 1, 3 and 4 of the stack, so this was the gap. Worth saying up front: this branch is rebased onto #892's current head, including `2b8faf39`, which #894 and #895 are not. That made it reviewable as written.

**The ownership resolution is the right shape.** `catalogProviderOwner` errors on every ambiguous case rather than picking one: two rows claiming the catalog id, one owner while another row is NAMED for it, or a name match with no owner at all. Adopting by `EqualFold` name-or-catalog was how the old code could attach a credential to a row that never claimed that provider, and refusing is the correct answer for a question with two possible answers.

I also checked the migration story before raising anything about it, and most of it is good: `providers add groq` against a legacy row named `groq` with no `catalogId` adopts the existing row and backfills the id rather than creating a duplicate.

```
before:  {"name":"groq", "providerKind":"openai-compatible", ...}          (no catalogID)
after:   {"name":"groq", "catalogID":"groq", "apiKeyEnv":"GROQ_API_KEY", ...}
```

That is the self-healing path, and it is the reason this hardening is landable at all.

**What blocks it: the auth path hits the strict error with no way out stated.**

Same legacy shape, which is what older versions wrote:

```
$ zero auth login openrouter
[zero] saved profile "openrouter" does not prove ownership of catalog provider "openrouter" (catalogId is "")
```

The message diagnoses precisely and then stops. There is a fix, and I confirmed it works:

```
$ zero providers add openrouter
Added provider openrouter to ...\config.json
   rows: [{"name":"openrouter","catalogID":"openrouter"}]
$ zero auth login openrouter
[zero] oauth: provider "openrouter" is not configured; set ZERO_OAUTH_OPENROUTER_CLIENT_ID ...
```

So the user is one command from working and has no way to discover it. `EnsureCatalogProvider` is reached from `auth.go:38` and `auth.go:165`, both of which are the "I want to log in" path, which is exactly where somebody with an old config arrives.

You already solved this shape on #892, where the ambiguity errors name `zero providers remove groq` and doctor repeats it. I would take the same approach here: append the remedy to the error, or have `EnsureCatalogProvider` perform the backfill itself when there is exactly one name match and its `catalogId` is empty. The second is defensible because an empty `catalogId` is not a competing claim, it is an absent one, and a single name match with nothing contradicting it is not the ambiguity this function exists to reject.

If you would rather keep the strict refusal, that is a reasonable call too, but then the message needs to carry the command.

Nothing else blocking. `ResolvePersistedProviderName`'s rule, exact spelling wins and credential identity is a fallback with ambiguity an error, is the same rule as `ValidatePersistedProviderNames`, and having the two agree is the point of the split.

PierrunoYT:
First review from me on this one. I have now looked at 1, 3 and 4 of the stack, so this was the gap. Worth saying up front: this branch is rebased onto #892's current head, including 2b8faf39, which #894 and #895 are not. That made it reviewable as written.

The ownership resolution is the right shape. `catalogProviderOwner` errors on every ambiguous case rather than picking one: two rows claiming the catalog id, one owner while another row is NAMED for it, or a name match with no owner at all. Adopting by EqualFold name-or-catalog was how the old code could attach a credential to a row that never claimed that provider, and refusing is the correct answer for a question with two possible answers.

I also checked the migration story before raising anything about it, and most of it is good: `providers add groq` against a legacy row named `groq` with no `catalogId` adopts the existing row and backfills the id rather than creating a duplicate.

```
before:  {"name":"groq", "providerKind":"openai-compatible", ...}          (no catalogID)
after:   {"name":"groq", "catalogID":"groq", "apiKeyEnv":"GROQ_API_KEY", ...}
```

That is the self-healing path, and it is the reason this hardening is landable at all.

What blocked it: the auth path hit the strict error with no way out stated.

Same legacy shape, which is what older versions wrote:

```
$ zero auth login openrouter
[zero] saved profile "openrouter" does not prove ownership of catalog provider "openrouter" (catalogId is "")
```

The message diagnoses precisely and then stops. There is a fix, and I confirmed it works:

```
$ zero providers add openrouter
Added provider openrouter to ...\config.json
   rows: [{"name":"openrouter","catalogID":"openrouter"}]
$ zero auth login openrouter
[zero] oauth: provider "openrouter" is not configured; set ZERO_OAUTH_OPENROUTER_CLIENT_ID ...
```

So the user was one command from working and had no way to discover it. `EnsureCatalogProvider` is reached from `auth.go:38` and `auth.go:165`, both of which are the "I want to log in" path, which is exactly where somebody with an old config arrives.

## Addressed in b61c1c68

I took the second of the two options I described — the backfill — because an empty `catalogId` is not a competing claim, it is an absent one, and a single name match with nothing contradicting it is not the ambiguity this function exists to reject.

`catalogProviderOwner` now returns a three-state ownership (`none` / `owned` / `adoptable`) and the index of the row, instead of a bool:

- **owned** — unchanged, exactly one row proves ownership through its `catalogId`.
- **adoptable** — exactly one row carries the catalog spelling as its NAME and claims no catalog id. `EnsureCatalogProvider` backfills `catalogId` onto that row and writes; name, credentials, base URL and model stay exactly as the user left them, and `activeProvider` is untouched. `Created` stays false, so the login prints "already configured" rather than claiming it added a row.
- **none** — unchanged, a row is created.

`PreflightCatalogProviderLogin` treats adoptable as good as owned, so the refusal no longer lands before the browser flow. That also covers the TUI, which preflights through the same function.

Every ambiguity this PR introduced still fails closed: two rows sharing the catalog id, an owner plus a different row named for it, and — the case worth being explicit about — a row NAMED for the provider whose `catalogId` points at a *different* one. That last one is a competing claim, not an absent one, so it is still refused, and per your other point the message now carries the command:

```
saved profile "OpenRouter" does not prove ownership of catalog provider "openrouter"
(catalogId is "custom-openai-compatible"); rename or remove that row with
`zero providers remove OpenRouter`, then run `zero providers add openrouter`
```

The two shared-catalog-id ambiguity messages carry `zero providers remove <name>` now too, matching how #892 words them.

Tests: `TestEnsureCatalogProviderRequiresPositiveCatalogOwnership` keeps the foreign-catalog rejection and now asserts both remedy commands are in the message; `TestEnsureCatalogProviderAdoptsLegacyRowWithoutCatalogID` pins the backfill, the preflight clearing it, the untouched fields, and the second login being a no-op; `TestRunAuthChatGPTAdoptsLegacyProfileWithoutCatalogID` pins the same end to end through `zero auth chatgpt`. `internal/config` and `internal/tui` are green. `internal/cli` has failures that reproduce on 33944034 without this change — they read the machine's real active provider — so they are environment, not this.

Nothing else blocking. `ResolvePersistedProviderName`'s rule, exact spelling wins and credential identity is a fallback with ambiguity an error, is the same rule as `ValidatePersistedProviderNames`, and having the two agree is the point of the split.

One thing I did not touch, flagging rather than fixing: `wizardProviderStoredKey` (`internal/tui/provider_wizard.go:1359`) carries its own copy of the strict message for a case-variant name with an empty `catalogId`. It is a different question — which stored key belongs to the descriptor — and it is read-only, so it is not the dead end this fixes. Worth folding into the shared rule at some point.

Vasanthdev2004 (dismissed):
Blocker fixed, and you took both options I offered rather than one. Clearing my verdict.

I verified it the way I raised it, by building the binary and running the command against a legacy config with a row named `openrouter` and no `catalogId`:

```
$ zero auth login openrouter
[zero] oauth: provider "openrouter" is not configured; set ZERO_OAUTH_OPENROUTER_CLIENT_ID ...
```

That is the message I could only reach after running `zero providers add openrouter` last time. The ownership gate no longer stops a user with an old config, and the refusals that legitimately remain now carry the command that fixes them. Treating a single name match with an empty `catalogId` as adoptable rather than ambiguous is the right reading: an absent claim is not a competing one.

**One thing about the series, and it is not about your code here.** I checked the merge order properly instead of repeating "please rebase", and there is no order that works:

```
892 then 893  CONFLICT  internal/cli/auth.go, auth_test.go
893 then 892  CONFLICT  internal/cli/auth.go, auth_test.go
892 then 895  CONFLICT  app_test.go, auth.go, auth_test.go
895 then 892  CONFLICT  internal/cli/app_test.go
```

By file set, 893 is contained in 894 is contained in 895, so those three are genuinely cumulative. 892 is not in that chain: 40 files are touched by both 892 and 893 and 20 of them differ, and five files exist only on 892, including `internal/config/provider_ownership.go` and `internal/cli/completions.go`, which are absent from 893, 894 and 895 alike.

So this approval is about the content, not about the landing. Nothing in the series can merge until 893 is rebased onto 892, 894 onto 893 and 895 onto 894, or until you tell us 892 is superseded and the other three carry its work. I have said the same on #894 and #895 so the ask is in one place per PR rather than scattered.

Worth doing that restack sooner than later, because #894 currently inherits a live deadlock from its stale base that 892's current head already fixes. Details on that PR.

PierrunoYT:
Restacked this branch onto the current #892 head (`5ad69c0`), so the history is now #892 followed by #893's seven commits. The PR base remains `main` because #892's branch lives in the fork; once #892 lands, GitHub will exclude that ancestry from this PR's effective diff.

I resolved the two auth conflicts by keeping #892's newer provider-ownership/preflight behavior while preserving #893's catalog-specific preflight and the mid-login config-mutation regression.

I also addressed the remaining current review findings in `542c223`:
- bind `PublishProviderCredential` to the credential store beside its `path` argument, including rollback
- make the credential-publication tests hermetic and path-specific
- remove the obsolete `preflightAuthLogin` helper during the restack
- redact the shared credential from test failure output
- add CLI coverage proving `doctor` forwards persisted-provider-name validation and repair guidance

Local validation (Go 1.26.6):
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make vulncheck` — no vulnerabilities
- `git diff HEAD --check`

@Vasanthdev2004 #893 is ready for another look once the refreshed CI finishes.

PierrunoYT:
Addressed the latest CodeRabbit review in `9598b0e`:

- resolve API-key stores beside the config being mutated (CLI and TUI, with default fallback for an empty path)
- reconcile adopted legacy OAuth profiles in memory without duplicating them or replacing their model
- give exact-name stored-key owners precedence over earlier case siblings
- sync model-picker state using the selected profile spelling while persisting through the config row spelling
- use the serialized `catalogID` spelling in ownership diagnostics
- assert the repaired fallback name returned by `RepairUnnamedProvider`
- align the OpenRouter mid-flow rejection test with the config-adjacent credential store

The older provider-removal serialization finding is handled by the transaction layer in the dependent PR #894 rather than duplicated into this lower stack layer.

Validation passed:
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test -race ./internal/config ./internal/cli ./internal/tui`
- `make lint-static` (0 issues)
- `make vulncheck` (no vulnerabilities)
- `git diff HEAD --check`

Vasanthdev2004 (changes requested):
Reviewed fresh at `9598b0eb`. Requesting changes: `auth logout` and `providers remove` now delete credential-store entries the addressed row does not own, and one of the new test files writes to the real per-user credential store on Windows.

Scope note first: I looked at `5ad69c08..9598b0eb`, this PR's own commits, not the stacked diff. The `ValidatePersistedProviderNames` hard-fail on unnamed and case-duplicate configs is #892's, byte-identical at its head, so that release note belongs there.

**1. `auth logout <row>` deletes the dictation API key.**

`ProviderCredentialCandidates` adds `row.CatalogID` to the candidate list when the catalog id is exclusive, and `runAuthLogout` loops that same list over the API-key store. The comment there says the API key store "has the identical asymmetry" as the OAuth store. It does not. `ApplyStoredAPIKey` looks up `profile.Name` only, so the API-key store has no catalog-id reader for the expansion to serve, and it shares its namespace with dictation: `newSaveSTTKey` does `store.Set(provider, key)` with the bare provider id into `ProviderKeyStoreAt(filepath.Dir(userConfigPath))`, the same flat store, and "groq", "openai" and "deepgram" are catalog ids. `CatalogIdentityExclusive` only walks `persistedProviders(path)`, so it cannot see the namespace it is meant to protect.

```
config.json      {"providers":[{"name":"my-groq","catalogID":"groq","apiKeyStored":true}],"stt":{"provider":"groq"}}
credentials.json {"my-groq":"CHAT-KEY-FOR-MY-GROQ","groq":"STT-DICTATION-KEY"}

zero auth logout my-groq
  HEAD -> {}
  BASE -> {"groq":"STT-DICTATION-KEY"}
```

`dictation.go` is byte-identical on both sides, so the same reader sees both states: against base's post-logout directory `buildBatchTranscriber` builds a transcriber, against head's it returns "groq API key not found". On that store the expansion can only ever delete something the addressed row does not own. The OAuth half is correct and should stay.

**2. `providers remove` moved to the default-path store, and that reaches out of the test harness.**

`removeStoredProviderKeyAt(configPath, name)` became `config.ForgetProviderKey(name)`, and the comment explaining the old choice went with it. Production is unaffected, both binaries print `"keyRemoved": true` and empty the store beside the default config. The cost lands on anyone running the repo's own tests: with an ambient store seeded, `go test ./internal/cli/` leaves `work` present on base and absent on head, through `TestProviderRepairCommandsCanResolveIndependentLegacyNameProblems`, which injects a config path but never redirects the user config root. Reverting that one line restores it and makes `TestRunProvidersRemoveDeletesKeyFromUserStore` fail on its own payload assertion, so the default-store choice is deliberate and pinned. Worth noting the PR now has three key-delete sites resolving the store by two different rules.

**3. The new TUI tests write to the real credential store on Windows.**

`caseSiblingModel` isolates with `t.Setenv("XDG_CONFIG_HOME", ...)`, but `config.UserConfigDir` consults XDG only on darwin and otherwise returns `os.UserConfigDir()`, which is `%AppData%`. So `config.ProviderKeyStore()` at `provider_ownership_test.go:50`, `:118` and `:222` opens `%APPDATA%\zero\credentials.enc` on a Windows dev box, and nothing restores it. Base has no test that opens the default-path store; head has twelve. I did not run those tests here for that reason. The fix already exists one package over: the PR's own new CLI tests call the GOOS-aware `setCLIUserConfigRoot`, and the TUI file just omits it. Linux and macOS are fine.

Two smaller ones. `providers remove` went silent when a case sibling still claims the secret, where base printed "Deleted its stored API key." and the same PR added two explicit sentences for exactly this case on the TUI surface; `--json` is honest, stderr is empty. And `auth_test.go:1078` heads a test with the name of a function that exists nowhere in the tree and states the inverse of what the test asserts, while the PR body says logout "still cleans up credentials when unrelated config is invalid" when it does not: with rows `work` and `WORK`, `auth logout demo` exits 1 and the token survives where base exits 0 and deletes it. Two of the three shapes are recoverable with the command the error prints; for a config.json that is not valid JSON there is no CLI path on head that clears the token.

**The fix, as one rule.** A delete may touch only the identities the addressed row actually owns, in the store beside the config being edited. That means keeping the catalog-id expansion for the OAuth token store and not applying it to the API-key store, putting `providers remove` back on `ProviderKeyStoreForConfigPath(configPath)` so all three sites agree and the CLI tests stop reaching into a developer's real store, and giving the TUI tests the same GOOS-aware redirect the CLI tests already use. While you are in there, print the retained-key sentence on the CLI path the way the TUI now does, and fix the stale header comment and the PR body line so they describe the fail-closed behaviour you actually shipped.

Checked and correct, so you do not need to chase these: `CatalogIdentityExclusive` is sound for what it covers, and two rows sharing a catalog id suppress the expansion in both directions rather than picking one; stubbing it to `return true` kills `TestRunAuthLogoutLeavesSharedCatalogCredentialsAlone` on its property assertion. The candidate ordering is pinned by an exact full-slice comparison, not a substring. The ownership shadow guard is real and load-bearing. Credential redaction holds across thirteen CLI surfaces and six resolve-failure modes with a live canary key, and swapping `redaction.ErrorMessage` for the raw error fails the test with the secret verbatim. No new trust anchors and no network in the resolution path, with every proxy blackholed producing byte-identical output. `auth refresh` stops on a 401 rather than spraying the next candidate at another host. The remaining test failures on head reproduce identically on base.

PierrunoYT:
Addressed the requested changes in [e7ed6f11](https://github.com/PierrunoYT/zero/commit/e7ed6f115a55e605ac962e64f2a5066c82935cab):

- `auth logout` still expands exclusive catalog aliases for OAuth tokens, but API-key deletion is now limited to the addressed persisted row name. Catalog-keyed dictation credentials are preserved, including when no persisted provider row owns the address.
- `providers remove` now deletes from the credential store beside the config being edited, leaves the default user store untouched for injected config paths, and prints the retained-key explanation when a case sibling still owns the key.
- The TUI key-removal path follows the same row-name-only API-key rule.
- The TUI ownership fixture now redirects `APPDATA` on Windows (and the appropriate config root on macOS/Linux), so it cannot reach the developer's real credential store.
- Corrected the stale fail-closed test comment and updated the PR description to match the shipped behavior.

Added regressions for the dictation-key collision, the no-persisted-row case, config-adjacent deletion/default-store preservation, retained CLI output, TUI row ownership, and cross-platform test isolation.

Validation:
- `make fmt-check`
- `go vet ./...`
- affected CLI/config/TUI tests and full `internal/cli` + `internal/tui` packages
- Windows TUI test cross-compile
- pinned golangci-lint: 0 issues
- `govulncheck`: no vulnerabilities
- `git diff HEAD --check`

`go test ./...` reaches one WSL-specific `internal/sandbox` expectation failure; the identical targeted failure was reproduced at merge base `5ad69c08`, so it is unrelated to this change.

Vasanthdev2004 (changes requested):
Everything I raised is genuinely closed at `e7ed6f11`, and I verified each one rather than reading the diff. Requesting changes on one thing only: the fix for the first finding trades it against jatmn's earlier #725 finding, and there is a shape that satisfies both.

**Closed, driven on both heads.**

The dictation key survives now. Logging out of `my-groq` with `stt.provider: groq` leaves `groq` intact on this head and destroys it at `9598b0eb`, so the regression was real and is gone. `providers remove` is back on the store beside the config being edited, and the symptom is gone with it: running the whole `internal/cli` package leaves my own `credentials.enc` byte-identical, where before it emptied entries. The TUI ownership tests isolate properly too, with a GOOS-aware helper that sets `APPDATA` on Windows; I ran them with a hash of my real store before and after and it did not move. The retained-key case now says so out loud on the text path, and the stale header comment is gone along with the test name that contradicted its own assertion.

**The one thing to fix.**

Scoping API-key deletion to the persisted row name closes my finding by giving up jatmn's. A key stored under the catalog id by a catalog-style auth flow is now left behind:

```
config: {"providers":[{"name":"my-xai","catalogId":"xai","apiKeyStored":true}]}   (no sibling, no dictation)
store:  xai = CATALOG-FLOW-KEY

zero auth logout my-xai
  9598b0eb -> catalog-id key present after logout = false     (deleted)
  e7ed6f11 -> catalog-id key present after logout = true      (left on disk)
```

That is the #725 case the replaced `TestRunAuthLogoutDeletesCatalogIDAPIKey` used to pin, and it means a credential the user explicitly asked to clear stays on disk. I would not have spotted it from the diff alone; what gave it away was that the PR replaces a test which pinned the opposite behaviour, which is usually worth a second look.

Both findings are real and they do not actually conflict. The namespace is only ambiguous when something else claims the catalog id, and the exclusivity check you already have is the right place to say so: `CatalogIdentityExclusive` walks the persisted rows to decide whether a catalog id belongs to the addressed row alone, and dictation is simply another claimant it does not know about. `persistedProviders` already unmarshals the whole `FileConfig`, so `cfg.STT.Provider` is right there and is currently discarded.

Teach that check about the dictation claim and use it for the API-key store as well, and both hold: `my-groq` with `stt.provider: groq` leaves `groq` alone, and `my-xai` with nothing else claiming `xai` deletes it. Worth a comment naming dictation as a claimant, so the next feature that writes into that flat store knows it has to be added here too.

Everything else I checked is right: the ordering that clears the marker before the secret, the ambiguity refusal that returns no candidates at all, and the redaction on the failure paths.

PierrunoYT:
Addressed the remaining review finding in ebf85793.

- `CatalogIdentityExclusive` now treats both `stt.provider` and `stt.streamProvider` as claimants in the shared flat credential namespace.
- API-key logout now deletes the persisted row key plus the catalog-id key only when that catalog identity is exclusive. OAuth candidates remain separate because the user-entered catalog spelling must still be considered for OAuth cleanup even when dictation claims the API-key spelling.
- Added regressions showing `my-xai` deletes an exclusive `xai` catalog-flow key, while logout through `groq` preserves the dictation key and still removes `my-groq`.

Verification: focused config/CLI tests pass, `go vet` passes, release build and smoke pass, pinned static lint reported 0 issues, pinned govulncheck reported no vulnerabilities, and diff checks pass. The repository-wide test run still has the existing unrelated `internal/cli` active-provider fixture failures.

Vasanthdev2004 (dismissed):
Approving at `ebf85793`. Both claims hold now, and I drove all three cases rather than reading the diff.

```
my-groq      + stt.provider "groq"          -> groq key survives      (my finding)
my-deepgram  + stt.streamProvider "deepgram"-> deepgram key survives  (the one I missed)
my-xai       + nothing else claiming "xai"  -> xai key deleted        (jatmn's #725 finding)
```

Teaching `CatalogIdentityExclusive` about the claim is the right place for it, and it is better than what I asked for in two ways. `persistedFileConfig` hands back the whole config instead of adding a second read, and you covered `STT.StreamProvider` as well, which I had not named and which would have been the same bug a week later. The comment saying any future feature storing keys in that flat namespace has to add its claims here is the part that keeps this from recurring.

It is pinned, and the pin is load-bearing: dropping the claim check fails `TestRunAuthLogoutPreservesDictationCatalogAPIKey` on its own assertion rather than somewhere incidental.

`internal/config` is green here and all seven checks pass on this head. Thanks for the quick turnaround on that one.

PierrunoYT:
PierrunoYT refreshed this branch by merging the updated #892 head, preserving remote ancestry (no force push): [3ecbda3](https://github.com/PierrunoYT/zero/commit/3ecbda3913f6acc2acad942a63a74dfba6b31ccf).

The latest human review approves the catalog layer. This update retains its catalog ownership/adoption rules, dictation namespace claims, config-adjacent credential stores and hermetic fixtures, while inheriting #892's legacy-row merge repair and protection against a separately resolved project/environment row writing through to a differently spelled user row. CLI mutations now pass the original requested name to that ownership check. The remaining removal-serialization concern is implemented in dependent #894, rather than duplicated here.

Regression proof: `go test -overlay=/tmp/pr893-overlay.json ./internal/tui -run TestModelSwitchSyncsSavedProviders -count=1`, with the old ownership resolver substituted, failed with `expected a session-only model switch, status="Model\nSwitched to openai · gpt-5.5"`. The fixed run passes. One obsolete synthetic TUI assertion expected lowercase `openai` to write the distinct `OpenAI` row; it now verifies session-only selection and unchanged disk state.

Validation on this head with go.mod's Go 1.26.6:
- `make fmt-check`, `go vet ./...`, `go test ./...` — passed.
- `go test -race ./internal/config ./internal/cli ./internal/tui` — passed.
- `go run ./cmd/zero-release build` and `go run ./cmd/zero-release smoke` — passed (0.8.0).
- `make lint-static` — 0 issues; `make vulncheck` — no vulnerabilities.
- `git diff HEAD --check` — passed.

Tests use process-local `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false` to isolate the orb's inherited Git signing configuration from scratch-repository fixtures; no global Git configuration was changed. Linux execution only; no native macOS/Windows run. TUI changes concern state/ownership, not visual layout, and were exercised through interaction/state tests. No PR was merged.

Vasanthdev2004 (dismissed):
Re-approving at `3ecbda39`. Nothing of this PR's own changed: all eleven commits are the same objects I approved at `ebf85793`, and the head is a merge of that head with the refreshed #892.

The merge carries three hand resolutions, and I read each against an automatic merge of the same two parents. The one that matters is in remove and rename: they now feed the raw argument into `resolveProviderMutationName` rather than the already-bridged spelling. Passing the bridged name would have satisfied the exact-match short cut and switched the cross-layer guard off, so I drove that direction: bridging first makes `TestProviderMutationsRejectProjectCaseSibling` fail in this tree, and the resolution as pushed keeps it green. The helper reads persisted rows through this PR's identity resolver, so catalog-id addressing keeps working for remove and rename, and a case-variant pair stays addressable only by exact spelling. The switch-model subtest was rewritten to the #892 rule that a lowercase row beside a persisted "OpenAI" is not the user's row and stays session only, and `caseSiblingModel` keeps this PR's `setTUIUserConfigRoot` over #892's inline env setup.

What the #892 refresh brings in I verified on #892 today. CI 6 of 6 at this head, and the six packages the PR touches pass here, apart from the serve symlink test that #1042 fixed on main after this branch's base. The branch conflicts with main the same way #892 does and needs that merge before it can go in.

Vasanthdev2004 (changes requested):
Re-reviewed at `749a3dd0`. This head is a refresh onto main plus `84278537`, and nothing of #893's own changed; the reason I am not re-approving is that it does not contain `78110c9f`, the commit on #892 that answers jatmn's five findings on the foundation.

That matters here in two concrete ways. First, the defects are live on this head. I ran the tests `78110c9f` adds against this tree: `providers remove WORK` and `rename WORK` fail with `no active provider configured: active provider "" not found` whenever no provider resolves, the guided repair leaves `activeProvider` on the case sibling and switches endpoint and key, a `--name` repair of a stored-key row succeeds and orphans the credential, and the picker after deleting the live project row builds the project client for a user-row selection. Second, `git merge-tree` of #892's head into this one conflicts in `internal/config/resolver.go`, `internal/config/writer_test.go` and `internal/tui/provider_ownership_test.go`, so the refresh is a real merge rather than a fast-forward, and the `resolver.go` one sits on the same active-row selection that this branch already patched its own way in `c64767b4`. The two are equivalent, so keeping `78110c9f`'s `normalizedProviderName` is the right resolution.

Please refresh onto #892 once the retained-identity fix I asked for there is in, since that is the part that will change again. Everything I checked on this head will carry over: the three hand resolutions are main plus this branch's hunks, the raw-argument rule in remove and rename survived them, `filterAuthStatuses` correctly keeps this branch's candidate form over main's single-provider one, and `TestProviderMutationsRejectProjectCaseSibling` is green. The blocked-ancestor classification from `84278537` is also applied to `persistedFileConfig` in the `749a3dd0` resolution, which is the right propagation. CI 9 of 9 at head; the six packages pass natively here apart from the transcript-scroll test.

Vasanthdev2004 (changes requested):
Re-reviewed at `010ccd81`. This now sits on #892's head, which is what I asked for, and nothing in its own delta needs to change. It is blocked only by what it carries.

The refresh is clean. `resolver.go` took #892's `normalizedProviderName` on both active-row lookups, which is the resolution I suggested, and `writer_test.go` and `provider_ownership_test.go` kept both sides. The other change since `749a3dd0` is test isolation: `isolateCLIUserState`, `clearProviderEnv` and a stubbed `resolveConfig` on a handful of `exec` and `providers` tests, plus explicit paths and an empty env in `TestResolveReportsExplicitMaxTurns`. I checked that those are hermeticity and not a test being steered around new behaviour: with the stubs taken back out, all eight tests still pass on this head. The `TestResolveReportsExplicitMaxTurns` change is a real improvement. That test fails on main on my machine because it reads the ambient config, and it passes here.

What blocks it is the resume block from #892's `f187303a`, which this head contains. The same sequence reproduces here: delete the live project row, resume a session recorded on `work`, and the labels, the summary and the manager badge say `work` while the live profile and the client are still the project row's. Details are in my review on #892. Once that is settled there and this is refreshed onto it, I expect to approve without another full pass.

CI is 9 of 9 at head. cli, config and oauth pass natively here, tui apart from the transcript-scroll test that fails on main on this box too.

jatmn (changes requested):
I found an issue that needs to be addressed before this PR is ready.

## Merge readiness

This is the second PR in the #892 → #893 → #894 → #895 stack. [#892](https://github.com/Gitlawb/zero/pull/892) is open with changes requested; its `/resume` path can update the displayed provider and model after a removed project row while retaining the old live client. Resolve that parent review and merge #892 before merging #893, then refresh this branch so its GitHub diff contains the second commit alone. #894 covers the separate credential transaction work and #895 covers selection presentation; neither supersedes the CLI identity defect below.

At review time, live `main` equals the captured merge base `99721c762f37cd43ac511007a5f51d1846df959e`; #893 head is `010ccd81a81c9d0d70a5ecd438efc48879baec1a`. GitHub reports no merge conflict (`MERGEABLE`), but the PR is `BLOCKED` with `CHANGES_REQUESTED`. All ten reported checks pass. There is no stale-target rebase requirement at this head.

## Findings

### 🟠 P1 — Keep concrete provider rows ahead of saved catalog aliases in provider commands

📍 **Where:** [`runProvidersUse`](https://github.com/Gitlawb/zero/blob/010ccd81a81c9d0d70a5ecd438efc48879baec1a/internal/cli/provider_onboarding.go#L61-L79) and [`resolveProviderMutationName`](https://github.com/Gitlawb/zero/blob/010ccd81a81c9d0d70a5ecd438efc48879baec1a/internal/cli/provider_onboarding.go#L631-L660), called by the changed remove and rename paths.

💥 **What fails:** Save a user provider `work` with `catalogID: "openai"` and a user endpoint, and add a distinct project provider named `openai` with a different endpoint. `providers list --json` displays both. On this head, `zero providers use openai --json` exits successfully but writes `activeProvider: "work"`, selecting the user endpoint and credential instead of the requested project row without warning. Address the project row as `OPENAI` and `zero providers remove OPENAI --json` removes `work`; `zero providers rename OPENAI renamed --json` renames `work`. Remove can also delete `work`'s stored key. The text and JSON responses claim success on the wrong saved row.

🔎 **Root cause:** The new persisted-identity lookup treats a catalog ID match as enough to rewrite the requested name before checking whether a separate resolved row owns that spelling. The mutation helper inherited from #892 checks provenance only for `ProviderNameExact`; a unique case-normalized resolved row bypasses it after #893's new precheck sends a catalog alias into that path. These are two changed entry paths missing the same cross-layer ownership rule.

📜 **Stated contract:**

> [Accepted issue #721](https://github.com/Gitlawb/zero/issues/721): “`providers use` either makes the requested provider effective or exits with a warning/error when a higher-priority override prevents that.” [Issue #707](https://github.com/Gitlawb/zero/issues/707): “The provider list and internal configuration should remain in sync.”

🏷️ **Attribution:** PR-introduced in #893's second commit. On the identical fixture, the merge-base binary reports the project row as unsaved, leaves `activeProvider: "other"`, and does not mutate `work`; live `main` is that merge-base SHA. #892 added the exact-only mutation guard, but #893's catalog-alias persisted precheck makes this wrong-row path reachable.

📌 **In this PR:**

- `providers use`: exact or normalized concrete project/environment names can be remapped to the saved catalog owner's name before the unsaved-row path runs.
- `providers remove`: a normalized concrete row bypasses the exact-only ownership guard, so the saved catalog owner and its key may be removed.
- `providers rename`: the same normalized path renames the saved catalog owner. Text and JSON outputs for these commands reflect the wrong operation.

🔒 **Unchanged on main:** The merge-base name-only persisted check did not mutate the saved row for these project-row requests. Preserve exact saved-row addressing and catalog aliases when there is no separate concrete row.

🔧 **Required correction:** Apply the resolved-row provenance rule available on this head before converting a requested identity to a persisted catalog alias. A separate exact **or normalized** project/environment/command row must follow the unsaved-row or ownership-error path; only an unclaimed alias may select the saved catalog owner. Cover both `use` and the shared remove/rename mutation helper, with focused tests for project and environment rows, exact and case-normalized requests, config and credential side effects, and text/JSON results.

🛠️ **Author fix:** Close this ownership-ordering root cause on every `In this PR` row in one pass. A fix to `use` alone leaves remove/rename mutating another profile; an exact-only mutation fix leaves `use` selecting another endpoint. Use the existing provenance helpers at these changed CLI paths and keep the tests tied to the visible row and persisted side effects.

🚫 **Out of scope:** Reworking the resolver, OAuth or STT credential storage, TUI session resume, or the separate transaction and presentation PRs.

## Validation

Focused `go test` passed for `internal/config`, `internal/cli`, `internal/oauth`, `internal/credstore`, and `internal/doctor`; provider-focused `internal/tui` tests passed. `go test -race` for config, CLI, and OAuth, scoped `go vet`, head/base binary builds, `gofmt`, and `git diff --check` passed. The full focused TUI package run hit `TestRelativeAgeFormatsLastActivity` (also fails on the merge base here) and `TestAltScreenTranscriptScrollKeepsFooterFixed` (checkout-path-sensitive in this environment); neither changes the reproduced CLI finding. The reviewed checkout remains clean.

Issue #898, opened 2026-08-13
Document recurring review blockers in AGENTS.md (host tool version floors, vacuous tests, test isolation, platform error codes)


euxaristia:
Several review findings recur across unrelated pull requests. They are not caught by CI, and `AGENTS.md` does not currently say anything that would prevent them, so each one is rediscovered by a reviewer and costs a round trip.

**Host tool version floors.** The clearest example. Zero shells out to `git`, and no minimum version is declared anywhere: not in `README.md`, `CONTRIBUTING.md`, `AGENTS.md`, or `docs/`. Adding a modern flag therefore raises the floor silently, and an older host fails with a usage error rather than a clear message. `--end-of-options` (Git 2.24) was raised on #671; `--show-current` (2.22) is already referenced in `internal/agent/command_prefix.go`; `branch.autoSetupMerge=inherit` (2.35) needed a version gate in a test. Nothing tells a contributor to check.

**Tests that never reach the behavior they name.** The recurring problem is not a missing test but one that passes with and without the fix, because an earlier guard rejects the input first. Recent cases: a staging-privacy test where the containment check fired first and returned a different error the assertion still accepted; an intermediate-symlink test that only exercised the outer containment pre-check and never the handle-relative walker it was named for. `AGENTS.md` requires a regression test but does not require proving it fails without the change.

**Tests touching real user state.** Several plan-mode tests read and wrote the developer's actual user config directory because the isolation helper was applied to some tests and not others reaching the same storage. On Windows this needs `%AppData%` set as well, since `os.UserConfigDir` ignores `XDG_CONFIG_HOME` there.

**Platform-specific error codes.** `AGENTS.md` covers path canonicalization across platforms but not error codes. A no-follow guard checked only `ELOOP`, while `openat(O_DIRECTORY|O_NOFOLLOW)` on a symlink returns `ENOTDIR` on Linux and Darwin, so the guard was inert on both and the refusal surfaced as an unrelated error. Windows has the same shape: `\\?\` and `\\.\` are not UNC prefixes, and `NtCreateFile` information classes must match the struct passed.

Proposed fix: extend `AGENTS.md` section 4 ("Common Review Blockers") with these four classes.

[...]

PR #899, opened 2026-08-13
docs(agents): Record four recurring review blockers

euxaristia:
## Summary

Four classes of review finding keep recurring across unrelated pull requests. None are caught by CI, and nothing in `AGENTS.md` currently prevents them, so each one is rediscovered by a reviewer and costs a round trip. This adds them to section 4, "Common Review Blockers".

The clearest case is **host tool version floors**. Zero shells out to `git`, and no minimum version is declared anywhere in the repository: not `README.md`, `CONTRIBUTING.md`, `AGENTS.md`, or `docs/`. Adding a modern flag therefore raises that floor silently, and an older host fails with a usage error rather than a clear message. `--end-of-options` (Git 2.24) was raised on #671; `--show-current` (2.22) already appears in `internal/agent/command_prefix.go`; `branch.autoSetupMerge=inherit` (2.35) needed a version gate in a test. Nothing told a contributor to check.

**Prove the test fails without the change** targets a specific recurring shape. The problem is usually not a missing test but one that passes with and without the fix, because an earlier guard rejects the input before the behavior under test ever runs. Two recent examples: a staging-privacy test where the containment check fired first and returned a different error the assertion still accepted, and an intermediate-symlink test that only exercised the outer containment pre-check and never the handle-relative walker it was named for. `AGENTS.md` already requires a regression test; it did not require proving the test fails without the change.

**Hermetic tests** covers isolation helpers applied to some tests but not others reaching the same storage, which had plan-mode tests reading and writing the developer's real user config directory. It notes the Windows detail that `os.UserConfigDir` ignores `XDG_CONFIG_HOME`, so `%AppData%` must be set too.

**Error codes are platform-specific** extends the existing platform-truth rule, which covers path canonicalization but not error codes. A no-follow guard that checked only `ELOOP` was inert on Linux and Darwin, where `openat(O_DIRECTORY|O_NOFOLLOW)` on a symlink returns `ENOTDIR`.

Deliberately not included: a bullet about advisory lint findings. `deadcode` and `lint-static` both run with `continue-on-error: true` in `ci.yml`, so they do not gate, and section 3 already says to fix findings the change introduces. Adding a "lint blocks CI" rule would have been wrong.

## Changes

- `AGENTS.md`
  - Section 4, "Common Review Blockers": four new bullets, `Host tool version floors`, `Prove the test fails without the change`, `Hermetic tests`, and `Error codes are platform-specific`, each with the concrete versions or error codes involved rather than general advice.
  - No existing text changed; wrapping matches the file's ~80 column style.

## Test plan

Documentation only, no code paths touched.

- [x] `git diff --stat` shows `AGENTS.md` as the only modified file (30 insertions, 0 deletions)
- [x] No line exceeds 80 columns: `awk 'length > 80 {print FNR": "length}' AGENTS.md` returns nothing
- [x] Version claims cross-checked against the flags actually referenced in this repository (`--show-current` in `internal/agent/command_prefix.go`, `--end-of-options` on #671, `branch.autoSetupMerge=inherit` in a version-gated test)
- [x] Confirmed `deadcode` and `lint-static` are `continue-on-error: true` in `.github/workflows/ci.yml` (lines 53 and 59, 143 and 153) before deciding not to claim they gate

## Open question for maintainers

Should the project declare a **minimum supported Git version**? The new guidance tells contributors to state and gate a new flag's floor, but there is no declared baseline to measure against, so "is this flag too new" stays a judgment call. Picking a floor and recording it in `README.md` or `CONTRIBUTING.md` would make the rule checkable. I did not want to choose that number unilaterally.

Refs #898

## Summary by CodeRabbit

* **Documentation**
  * Added review guidance covering host-tool version requirements and fallback handling.
  * Added requirements for effective regression tests and hermetic test configuration.
  * Added guidance to verify platform-specific system-call and path behavior.
  * Improved consistency and reliability of development and testing practices across supported environments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -111,3 +111,33 @@ These classes drive multi-round reviews. Fix them before requesting review:
   shipped. Wire advertised entry points or shrink the claim. Do not bundle
   unrelated fixes. Do not widen security allowlists by name alone; check
   classification and side effects too.
+- **Host tool version floors:** Zero shells out to `git` and other host tools,
+  and the project does not declare a minimum version for any of them. A flag,
+  subcommand, or config key that is not long-established therefore raises that
+  floor silently, and the failure on an older host is an unhelpful usage error
+  rather than a clear message. Before adding one, check when it landed and say
+  so where a caller would look: `--end-of-options` is Git 2.24,
+  `--show-current` is 2.22, `branch.autoSetupMerge=inherit` is 2.35. Then
+  either keep a fallback for older versions or gate the call. A test that
+  depends on a newer feature must skip below its floor instead of failing, and
+  say which version it needs.
+- **Prove the test fails without the change:** A regression test earns its
+  place only if it fails on the unfixed code, for the reason it claims. Run it
+  both ways and quote the failure in the PR. The recurring miss is not a
+  missing test but a test that never reaches the behavior it names, because an
+  earlier guard rejects the input first, so it passes with and without the fix.
+  When a test targets a specific layer, call that layer directly rather than
+  relying on a public entry point that may refuse earlier.
+- **Hermetic tests:** Tests must not read or write the developer's real config,
+  cache, or state directories. Redirect the user config root to a temp
+  directory, and set `%AppData%` alongside `XDG_CONFIG_HOME` on Windows, where
+  `os.UserConfigDir` ignores XDG. Apply it to every test that reaches the same
+  storage, not only the one where it was first needed; a test that passes on a
+  clean machine and pollutes a real one is a blocker.
+- **Error codes are platform-specific:** The same syscall reports different
+  codes per OS, so a check written against one platform is silently inert on
+  another while still looking correct. `openat(O_DIRECTORY|O_NOFOLLOW)` on a
+  symlink reports `ENOTDIR` on Linux and Darwin but `ELOOP` elsewhere; on
+  Windows, `\\?\` and `\\.\` prefixes are not UNC paths, and an `NtCreateFile`
+  information class must match the struct actually passed. Confirm the code the
+  target platform returns instead of assuming the portable one.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,11 +116,14 @@ These classes drive multi-round reviews. Fix them before requesting review:
   subcommand, or config key that is not long-established therefore raises that
   floor silently, and the failure on an older host is an unhelpful usage error
   rather than a clear message. Before adding one, check when it landed and say
-  so where a caller would look: `--end-of-options` is Git 2.24,
-  `--show-current` is 2.22, `branch.autoSetupMerge=inherit` is 2.35. Then
-  either keep a fallback for older versions or gate the call. A test that
-  depends on a newer feature must skip below its floor instead of failing, and
-  say which version it needs.
+  so where a caller would look. Record the floor for the exact subcommand you
+  call, not for the flag in general: `--end-of-options` reached Git's shared
+  parser in 2.24, but `git rev-parse` only accepted it from 2.30, and
+  `git checkout` and `git reset` later still. Likewise `--show-current` is
+  `git branch` 2.22, and `branch.autoSetupMerge=inherit` is 2.35. Then either
+  keep a fallback for older versions or gate the call. A test that depends on a
+  newer feature must skip below its floor instead of failing, and say which
+  version it needs.
 - **Prove the test fails without the change:** A regression test earns its
   place only if it fails on the unfixed code, for the reason it claims. Run it
   both ways and quote the failure in the PR. The recurring miss is not a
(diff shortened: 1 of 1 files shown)


euxaristia:
All three addressed in `4abd0b12`. Two of them were corrections to claims I had made too strongly, which is a fair outcome for a PR about not asserting things you have not checked.

**Version floors are per subcommand.** Correct, and it matters more than a wording fix. `--end-of-options` reaching the shared parser in 2.24 does not mean every subcommand accepted it then: `git rev-parse` only did from 2.30. Recording the flag's general debut would still let a caller ship something that breaks on 2.24 through 2.29, which is exactly the failure the bullet exists to prevent. Now phrased as "record the floor for the exact subcommand you call", with `rev-parse` at 2.30 as the worked example and `--show-current` attributed to `git branch` rather than left bare.

**Cache and state roots.** Also correct. The sentence named config, cache, and state, then gave a procedure covering only config. It now says to redirect every root the code under test resolves, and names `%LocalAppData%` alongside `%AppData%`, since `os.UserCacheDir` ignores the XDG variables on Windows the same way `os.UserConfigDir` does.

**Errno guidance.** This one I had genuinely overreached on. The bullet asserted `ENOTDIR` "on Linux and Darwin". I had only measured Linux, in a container, while fixing #854; Darwin was an assumption, sitting in a bullet whose entire point is not to assume per-platform behavior. Rewritten to say the call may report either `ENOTDIR` or `ELOOP` and that both should be accepted, with the reason (the kernel can reject the directory mismatch before it resolves the link) rather than a table of which OS does what. That is also what the fix in #854 actually does, since it disambiguates with a no-follow stat instead of trusting the code.

Verified `AGENTS.md` is still the only modified file and no line exceeds 80 columns.

Vasanthdev2004:
@euxaristia heads up, we have both been editing `AGENTS.md` section 4 this morning without knowing it. I opened #900 a few hours before this landed, adding one bullet on normalize-before-match ordering. Yours appends at the end of the section and mine inserts mid-list, so git may well merge them cleanly, but whoever goes second should rebase and read the combined section rather than trusting the auto-merge.

Worth saying the overlap is complementary rather than duplicate, and that yours is the better half of it.

Your **"Prove the test fails without the change"** is the general rule. My bullet ends with one specific instance of it, that a value which is not split does not prove split-value redaction works. Your phrasing names the mechanism better than mine does: *a test that never reaches the behavior it names, because an earlier guard rejects the input first, so it passes with and without the fix.*

That is exactly what bit me twice in two days. On #891 a junction-based test could not distinguish the fix from the bug, because `EvalSymlinks` does not resolve junctions either, and I only caught it by reverting the fix and watching the test still pass. On #878 my own regression failed against the correct code because a control-byte literal was lost in editing and `strings.Contains(x, "")` is always true.

So if yours lands first I am happy to trim my closing sentence to a pointer at your bullet rather than restating it. If mine lands first, ignore that and I will rebase.

No objection to any of the four from me. The version-floor one is the sort of thing that only shows up on someone else's machine, which is the same reason the Windows path work in #891 and #897 needed a second pair of hands.

Vasanthdev2004 (changes requested):
I checked the four version claims against upstream release notes rather than taking them on trust, and all four hold: `git branch --show-current` in 2.22, `--end-of-options` in the shared parser in 2.24, `git rev-parse` learning it in 2.30, and `branch.autoSetupMerge=inherit` in 2.35 (2.37 adds `simple`, a different value). `--show-current` is where you say it is, `internal/agent/command_prefix.go:470`. Nothing over 80 columns. Splitting the flag's floor from the subcommand's floor is the detail most people would have missed, and it is the part that would actually have caught #671.

One thing to fix, then this is good to go.

## The hermetic-tests bullet is wrong on macOS, and macOS is a required check

The bullet says the Unix lever is "the relevant `XDG_*` variable for each", and attaches the ignores-XDG caveat only to Windows. Go ignores XDG on darwin too. From `os/file.go`:

```go
case "darwin", "ios":
	dir = Getenv("HOME")
	dir += "/Library/Application Support"   // UserConfigDir
	dir += "/Library/Caches"                // UserCacheDir

default: // Unix
	dir = Getenv("XDG_CONFIG_HOME")
```

Only the `default` branch reads XDG. So a contributor who follows this bullet on a Mac exports `XDG_CONFIG_HOME`, believes the test is now hermetic, and still writes into the developer's real `~/Library/Application Support`. That is the exact failure the bullet exists to prevent, and `Smoke (macos-latest)` gates every PR, so it is not a theoretical target.

Suggested replacement for that clause:

> Redirect every root the code under test resolves, not just config. `os.UserConfigDir` and `os.UserCacheDir` read `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` only on Linux and the BSDs. On macOS they derive from `HOME`, and on Windows from `%AppData%` and `%LocalAppData%`. Setting the XDG variables alone still leaves a macOS run writing to the real home, so set `HOME` as well, or take the directory as a parameter.

## On your open question

Yes, declare one, and I do not think you need to leave the number to us.

Nothing in `main` needs a modern git. The invocations are `worktree add --detach`, `rev-parse` with `--show-toplevel` / `--abbrev-ref` / `--short` / `--verify` / `--is-inside-work-tree` / `--git-common-dir`, `ls-remote --symref`, `bundle create --all`, `bundle verify`, `status --porcelain -z --untracked-files=all`, `clone --depth 1 --`, `apply --whitespace=nowarn`, `merge-base`, `diff --numstat -z`, and `rev-list --count --since=`. All long-established plumbing. I am deliberately not naming a version for the newest of those, because I did not pin it to a release note and I am not going to hold you to a standard I skip myself.

The point stands without the number: our code is not the binding constraint today, so the floor is a decision about which hosts we support, not a fact to be discovered. Pick one, put it in `CONTRIBUTING.md`, and your rule becomes checkable immediately. The first time code would bind is if #671 lands using `git rev-parse --end-of-options`, and your own bullet already pins that at 2.30.

## Two disclosures

I have a competing change to this same section in #900, so read the request above as one reviewer's opinion and not as me clearing a lane for my own PR.

They also do not collide, which corrects something I said elsewhere. `git merge-tree --write-tree` across both heads returns a clean tree carrying all five bullets, so neither of us has to rebase on the other and merge order does not matter.

Not yours: CI is red here for the repo-wide vulncheck outage, not for anything in this change. #903 fixes that.

euxaristia:
## Changes addressed

- Updated the hermetic tests bullet in `AGENTS.md` to specify that `os.UserConfigDir` and `os.UserCacheDir` read `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` only on Linux and BSDs, while deriving from `HOME` on macOS and `%AppData%` / `%LocalAppData%` on Windows.

Vasanthdev2004 (approved):
Re-reviewed at `f04082a3`. The macOS gap is closed, and closed properly: the bullet now says `os.UserConfigDir` and `os.UserCacheDir` read the XDG variables only on Linux and the BSDs, that macOS derives from `HOME`, and that setting XDG alone still leaves a macOS run writing to the real home. That is the whole of what I was blocking on.

Checked the rest rather than assuming: it still merges clean against `main` now that #900 has landed, the merged file carries all five bullets, and nothing exceeds 80 columns.

The four version claims I verified last round against upstream release notes are unchanged and still correct.

Good change. It is the kind of rule that only pays off months later when it stops a round trip nobody remembers needing.

kevincodex1 (approved):

kevincodex1:
hello this is approved already but theres failing tests please address @euxaristia

euxaristia:
> hello this is approved already but theres failing tests please address @euxaristia 

This PR only changes AGENTS.md so the test failures must be from main

Vasanthdev2004:
@euxaristia `main` is green, so the failures are not coming from there. I checked before answering: the CI run on `d065467c` succeeded.

The two red checks here are both the same step:

```
Security & code health          -> govulncheck
Smoke (windows-latest)          -> Check vulnerabilities through native Windows Make
```

Your branch is 7 commits behind `main` and predates both vulnerability fixes:

```
        go.mod          x/image
this PR go 1.26.5       v0.44.0
main    go 1.26.6       v0.45.0
```

`#903` bumped the toolchain for six stdlib advisories and `#902` carried `x/image` to v0.45.0 for GO-2026-6222. Neither is on this branch, so `govulncheck` still finds them here and nowhere else. Your reasoning was right in shape, the failures genuinely have nothing to do with your change, they are just from the base rather than from `main` as it stands today.

A rebase or a merge of `main` clears both. Happy to do it for you if you would rather not, though it will dismiss my approval and I will just re-approve after, since I have already read the content.

Worth saying this is not only you. Nine open PRs are still on `go 1.26.5`, including four of mine, and every one of them will fail these two checks until it picks up `main`. So if you see the same red on someone else's PR this week, it is almost certainly this and not their change.

euxaristia:
@kevincodex1 seems ready to me?

kevincodex1 (approved):

PR #835, opened 2026-07-30
fix(tui): show MCP servers that failed to start in /mcp

Vasanthdev2004:
Fixes #825. Companion to #822, which fixed the same blind spot in `zero mcp check`.

`/mcp` worked out each server's state from the config file — disabled if you turned it off, enabled otherwise. But MCP registration is best-effort: a server that can't be reached gets recorded and startup carries on. So a server that never connected showed up as enabled, its tools quietly missing, and nothing in the panel said why.

Startup does know — it prints a warning per skipped server to stderr. That's gone by the time you notice, and `/mcp` is exactly where you go afterwards to ask what's actually running.

So the skipped set now reaches the panel, and a server that failed renders as failed with the reason under it:

```
› docs · failed · stdio
  exec: "docs-mcp": executable file not found in $PATH
```

Two details worth calling out:

The reason comes from the server, so it goes through `redaction.ErrorMessage` before it's rendered. A handshake error that echoes the `Authorization` header back would otherwise print the bearer token straight into the transcript. There's a test for that.

Disabled wins over failed. If you turned a server off it was never expected to connect, and calling it failed would be misleading.

The stderr warning is unchanged — non-interactive users still get it, and the panel is an addition rather than a replacement.

**Still not fixed, and out of scope here:** enabling a server from inside the TUI updates the config but doesn't reconnect anything, so it'll show as enabled while not actually running until you restart. That's pre-existing and a bigger change; happy to file it separately if you'd like.

Verified with mutation testing — eight mutations across the state builder, the renderer, and both wiring points, all killed. `TestAltScreenTranscriptScrollKeepsFooterFixed` and `TestBuildServeScopeKeepsLexicalPaths` fail on my Windows box on clean `main` too (the second needs symlink privilege).

## Summary by CodeRabbit

* **New Features**
  * MCP servers that fail to start now appear as **failed** in the `/mcp` panel.
  * Failure reasons are shown in the panel and server details when available.
  * Startup failures remain visible while the affected configuration is unchanged.

* **Bug Fixes**
  * MCP status now reflects actual startup results, including disabled servers.
  * Failure messages and server targets redact credentials, remove unsafe formatting, and limit excessively long text.
  * Server names and status matching remain consistent across configuration updates.

PR #900, opened 2026-08-13
docs(agents): record the normalize-before-match blocker

Vasanthdev2004:
Adds one bullet to section 4, Common Review Blockers.

Three changes hit this class in a fortnight, each written by someone who was being careful about precisely the thing that caught them, which is why it seems worth writing down rather than catching a fourth time in review.

**The rule:** a transform that removes bytes without leaving a gap is also a reassembler, so it has to run before whatever matches on the result.

It showed up in two shapes.

*Redaction ordering.* Secrets are matched by shape, so a credential split by a NUL, an ESC or a C1 byte does not look like a key. Strip the control bytes afterwards and the fragments become the credential again, on the way to a transcript line or a picker row. That was #835 (an MCP failure reason) and #878 (the imported-transcript chokepoint). In both cases the unsplit value redacted correctly, which is exactly why it survived review: every test used unsplit values.

*Path comparison.* Asking where a handle landed and comparing it against a value produced by the same resolver the kernel just used means the two agree precisely when a redirect happened. That was #808, where junctions were rejected only by an accident of Go's mode bits (a junction is `ModeIrregular`, not `ModeSymlink`, so `EvalSymlinks` declines it) and directory symlinks were not rejected at all.

The last sentence of the bullet is the part I would keep if only one survived: **the unsplit value passing is not evidence.** That is what made this invisible three times.

No code change.

## Summary by CodeRabbit

* **Documentation**
  * Added security review guidance requiring byte-removing transformations to occur before matching.
  * Documented protections against encoding, canonicalization, path normalization, and split-credential bypasses.
  * Added guidance for regression tests covering split-input cases.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,6 +91,14 @@ These classes drive multi-round reviews. Fix them before requesting review:
   final-component-only no-follow is insufficient. On multi-step setup, roll back
   only what this run created; never destroy pre-existing resources you did
   not create; never report success when cleanup or unlock failed.
+- **Normalize before matching:** Any transform that REMOVES bytes without
+  leaving a gap is also a reassembler, so it has to run before whatever matches
+  on the result. Redacting by shape and then stripping control bytes lets a
+  credential split by a NUL, an ESC or a C1 byte pass the patterns as two
+  fragments and be rejoined on the way out; comparing a path against a form
+  produced by the same resolver the kernel just used makes a redirect agree with
+  itself. Strip, decode, or canonicalize first, then match. The unsplit value
+  passing is not evidence, so the test needs a split case.
 - **Atomic shared state:** Write a complete temporary file, then atomically
   replace the destination so concurrent readers never see a partial write.
   Exclusive create or a write lock alone is not enough for readers. Serialize


Vasanthdev2004:
@anandh8x @gnanam1990 review when you get a chance. Docs only, one bullet, CI green.

Tagging you two specifically because you both hit this class in the last fortnight and your names are on the evidence.

@gnanam1990 it is the ordering behind the `redact()` swap on #878, and the same shape as the reparse comparison on #891.

@anandh8x it is also the general form of the backslash finding you caught on #891: my suggested `TrimRight(relative, "/\\")` was a normalizer applied without asking what it would reassemble, and on unix it would have trimmed a legal filename character.

The sentence I care most about keeping is the last one, that the unsplit value passing is not evidence. That is what let this through three separate reviews, two of them mine.

If either of you thinks it belongs in a different section, or reads as too specific for a general blockers list, say so and I will move or trim it.

Vasanthdev2004:
Coordination note: @euxaristia opened #899 a few hours after this, also editing `AGENTS.md` section 4, addressing #898. Four bullets there against one here.

They complement rather than duplicate, but there is one real overlap. #899's **"Prove the test fails without the change"** is the general rule; the closing sentence of this bullet is one specific instance of it. If #899 lands first I will trim that sentence to point at their bullet instead of restating it, and rebase. If this lands first, they rebase.

Different insertion points, so the auto-merge may well be clean, but whoever goes second should read the combined section rather than trust it.

@anandh8x @gnanam1990 no action needed from you on this, just so the review of one does not happen in ignorance of the other.

kevincodex1 (approved):

PR #902, opened 2026-08-13
Improve diff viewer rendering

anandh8x:
## What changed

- add a compact unified-diff viewer with line numbers and long-context folding
- apply syntax highlighting across each diff hunk while preserving add/delete and changed-word emphasis
- keep rename and mode-only diff metadata visible
- remove the misleading hover/click toggle from always-visible diff cards

## Why

File edits are easier to review when source structure, changed regions, and surrounding context are visible without tool-output noise or inactive interaction affordances.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- `git diff --check`
- manual TUI verification of syntax-highlighted diffs and non-clickable diff-card headers

CodeRabbit review identified two scoped issues (rename/mode-only metadata and large-hunk span lookup); both were fixed with regression coverage. The follow-up CLI review was rate-limited.

## Summary by CodeRabbit

* **New Features**
  * Long unchanged sections in diffs are now compacted with a clear count of hidden lines.
  * Diff views preserve line numbering, metadata, syntax highlighting, and word-level changes.
  * Added improved highlighting for changed and unchanged diff content.

* **Bug Fixes**
  * Always-expanded diff cards no longer display misleading collapse controls.
  * Improved rendering for oversized lines, renames, file-mode changes, and missing-newline markers.
  * Text resembling expansion instructions no longer creates unintended toggles.
  * Edited diffs with blank lines now render reliably.
  * Expandable tool results remain correctly selectable.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/charmbracelet/x/term v0.2.2
 	github.com/coder/websocket v1.8.15
 	github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728
-	golang.org/x/image v0.44.0
+	golang.org/x/image v0.45.0
 	golang.org/x/sys v0.47.0
 	mvdan.cc/sh/v3 v3.13.1
 )
--- a/go.sum
+++ b/go.sum
@@ -66,6 +66,8 @@ golang.org/x/exp v0.0.0-20260611194520-c48552f49976 h1:X8Hz2ImujgbmetVuW+w2YkyZC
 golang.org/x/exp v0.0.0-20260611194520-c48552f49976/go.mod h1:vnf4pv9iKZXY58sQE1L86zmNWJ4159e1RkcWiLCkeEY=
 golang.org/x/image v0.44.0 h1:+tDekMZED9+LrtB3G5xzRggpVh9CARjZqROla3R3R+I=
 golang.org/x/image v0.44.0/go.mod h1:V8K3KE9KKKE+pLpQDOeN18w9oacNSvy1tDOirTu4xtY=
+golang.org/x/image v0.45.0 h1:FMb1nTbH5H9vF55SriQHgFw5GnNL9Jg6L25BwXKzhB0=
+golang.org/x/image v0.45.0/go.mod h1:n62x/7RqlwXDvGsSU4u6IUTUf6KghUZ9Bt7cG/T9Fx4=
 golang.org/x/sync v0.22.0 h1:SZjpbeLmrCk4xhRSZFNZW5gFUeCeFgjekvI/+gfScek=
 golang.org/x/sync v0.22.0/go.mod h1:9xrNwdLfx4jkKbNva9FpL6vEN7evnE43NNNJQ2LF3+0=
 golang.org/x/sys v0.47.0 h1:o7XGOvZQCADBQQ4Y7VNq2dRWQR7JmOUW8Kxx4ZsNgWs=
--- a/internal/tui/collapse_tools_test.go
+++ b/internal/tui/collapse_tools_test.go
@@ -146,8 +146,8 @@ func TestToolResultOutputCannotCreateToggle(t *testing.T) {
 	m := transcriptViewTestModel()
 	row := transcriptRow{
 		kind:   rowToolResult,
-		id:     "read",
-		tool:   "read_file",
+		id:     "custom",
+		tool:   "custom_tool",
 		status: tools.StatusOK,
 		detail: "the file literally says: click to expand",
 	}
@@ -157,5 +157,23 @@ func TestToolResultOutputCannotCreateToggle(t *testing.T) {
 	}
 	if selectable[0].toggle {
 		t.Fatalf("tool-output text must not create a header toggle: %#v", selectable[0])
+	}
+}
+
+func TestCollapsedExploreToolResultHeaderIsToggleable(t *testing.T) {
+	m := transcriptViewTestModel()
+	row := transcriptRow{
+		kind:   rowToolResult,
+		id:     "read",
+		tool:   "read_file",
+		status: tools.StatusOK,
+		detail: numberedLines(cardBodyMaxLines + 5),
+	}
+	_, selectable := m.renderSelectableToolResultRow(0, row, m.width, buildRowContext(nil), 0)
+	if len(selectable) == 0 {
+		t.Fatal("expected a visible explore result row")
+	}
+	if !selectable[0].toggle {
+		t.Fatalf("collapsed explore card header must remain toggleable: %#v", selectable[0])
 	}
 }
--- a/internal/tui/collapse_tools_test.go
+++ b/internal/tui/collapse_tools_test.go
@@ -121,3 +121,21 @@ func TestToolResultRowExposesClickToggle(t *testing.T) {
 		t.Fatalf("tool result visible body/footer must stay selectable, got %#v", selectable)
 	}
 }
+
+func TestAlwaysExpandedToolResultHeaderIsNotToggleable(t *testing.T) {
+	m := transcriptViewTestModel()
+	row := transcriptRow{
+		kind:   rowToolResult,
+		id:     "diff",
+		tool:   "edit_file",
+		status: tools.StatusOK,
+		detail: "--- a/example.go\n+++ b/example.go\n@@ -1 +1 @@\n-old\n+new",
+	}
+	_, selectable := m.renderSelectableToolResultRow(0, row, m.width, buildRowContext(nil), 0)
+	if len(selectable) == 0 {
+		t.Fatal("expected a visible diff result row")
+	}
+	if selectable[0].toggle {
+		t.Fatalf("always-expanded diff header must not advertise a collapse toggle: %#v", selectable[0])
+	}
+}
--- a/internal/tui/collapse_tools_test.go
+++ b/internal/tui/collapse_tools_test.go
@@ -124,18 +124,38 @@ func TestToolResultRowExposesClickToggle(t *testing.T) {
 
 func TestAlwaysExpandedToolResultHeaderIsNotToggleable(t *testing.T) {
 	m := transcriptViewTestModel()
+	for _, tool := range []string{"edit_file", "apply_patch", "write_file"} {
+		row := transcriptRow{
+			kind:   rowToolResult,
+			id:     "diff",
+			tool:   tool,
+			status: tools.StatusOK,
+			detail: "--- a/example.go\n+++ b/example.go\n@@ -1 +1 @@\n-old\n+new",
+		}
+		_, selectable := m.renderSelectableToolResultRow(0, row, m.width, buildRowContext(nil), 0)
+		if len(selectable) == 0 {
+			t.Fatalf("%s: expected a visible diff result row", tool)
+		}
+		if selectable[0].toggle {
+			t.Fatalf("%s header must not advertise a collapse toggle: %#v", tool, selectable[0])
+		}
+	}
+}
+
+func TestToolResultOutputCannotCreateToggle(t *testing.T) {
+	m := transcriptViewTestModel()
 	row := transcriptRow{
 		kind:   rowToolResult,
-		id:     "diff",
-		tool:   "edit_file",
+		id:     "read",
+		tool:   "read_file",
 		status: tools.StatusOK,
-		detail: "--- a/example.go\n+++ b/example.go\n@@ -1 +1 @@\n-old\n+new",
+		detail: "the file literally says: click to expand",
 	}
 	_, selectable := m.renderSelectableToolResultRow(0, row, m.width, buildRowContext(nil), 0)
 	if len(selectable) == 0 {
-		t.Fatal("expected a visible diff result row")
+		t.Fatal("expected a visible tool result row")
 	}
 	if selectable[0].toggle {
-		t.Fatalf("always-expanded diff header must not advertise a collapse toggle: %#v", selectable[0])
+		t.Fatalf("tool-output text must not create a header toggle: %#v", selectable[0])
 	}
 }
--- a/internal/tui/diff_viewer.go
+++ b/internal/tui/diff_viewer.go
@@ -44,7 +44,7 @@ func compactDiffViewerContext(raw []string) []diffViewerLine {
 		for i := start; i < start+diffViewerContextLines; i++ {
 			lines = append(lines, diffViewerLine{text: raw[i], rawIndex: i})
 		}
-		lines = append(lines, diffViewerLine{hiddenContext: count - diffViewerContextLines*2})
+		lines = append(lines, diffViewerLine{rawIndex: -1, hiddenContext: count - diffViewerContextLines*2})
 		for i := index - diffViewerContextLines; i < index; i++ {
 			lines = append(lines, diffViewerLine{text: raw[i], rawIndex: i})
 		}
--- a/internal/tui/diff_viewer.go
+++ b/internal/tui/diff_viewer.go
@@ -1,0 +1,51 @@
+package tui
+
+import "strings"
+
+// Visible unchanged lines kept at each edge of a collapsed diff section.
+const diffViewerContextLines = 3
+
+// diffViewerLine is a source diff line or a synthetic collapsed-context row.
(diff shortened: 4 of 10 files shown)


Vasanthdev2004 (changes requested):
The direction here is right and the syntax-highlighted diff reads a lot better than what we have. Two regressions to fix first. Both reproduce against `main`, and both are in the rendering paths this PR rewrites.

## 1. A deleted line can disappear from the diff card

In `diffCardBody` the `-` case became if / else-if / else, and the plain-text append moved inside the final `else`. So when `isIsolatedReplacement` is true but `renderWordDiffPair` declines, nothing gets appended for that row. `renderWordDiffPair` declines whenever the change covers more than 60% of the longer line, which is the near-rewrite case it was always meant to hand back. `oldLine++` still runs, so the gutter numbers stay correct and the line just silently vanishes.

Before this PR that same case fell through to the shared plain append below the `if`.

Drop this in `internal/tui`:

```go
func TestNearRewriteDeletionSurvives(t *testing.T) {
	diff := strings.Join([]string{
		"--- a/notes.log",
		"+++ b/notes.log",
		"@@ -1 +1 @@",
		"-alpha bravo charlie delta",
		"+zulu yankee xray whiskey",
	}, "\n")
	body := diffCardBody(diff, 100, cardRenderOptions{bodyCap: 0})
	got := plainRender(t, strings.Join(body.lines, "\n"))
	if !strings.Contains(got, "alpha bravo charlie delta") {
		t.Fatalf("deleted line dropped:\n%s", got)
	}
}
```

On `main` it renders both rows. On `f70481ee` it renders only `1 + zulu yankee xray whiskey`.

This needs the highlighter to be absent for that row, which is easier to hit than it looks. Files with no lexer (`.log`, extensionless) always take it. More importantly `highlightedDiffLines` returns nil for the *whole* diff when any single line exceeds `diffViewerHighlightMaxLineBytes`, or when the aggregate budget trips. I checked that path: the same test on `app.go`, with one 4KB line added in a second hunk, also loses the deletion. So one long line in an unrelated hunk silently deletes content from the rest of the card.

## 2. Collapsed explore cards are no longer clickable

`toolResultCanToggle` matches `▸ N lines ... click to expand` or `▾ collapse`. But `exploreCardBody` emits `▸ details` as the footer for a collapsed `read_file`, `grep`, `glob`, `list_directory` or `read_minified_file`. That string matches neither, so the header stops carrying a toggle and those cards can no longer be expanded by mouse.

Measured on a 60-line `read_file` result at widths 120, 80, 60, 40, 30, 24 and 20: `main` reports `toggle=true` at every width, this branch reports `toggle=false` at every width, with `▸ details` as the last rendered line in both.

The two new tests cover the always-expanded diff tools and a short `read_file` that never collapses, so neither one reaches a card that actually has a collapse affordance.

I would not fix this by adding `▸ details` to the match list. Deciding a card's interactivity by pattern-matching its own rendered text is what broke here, and it will break again the next time a footer is reworded. Better to have the card renderer report whether it produced a collapse affordance and read that.

## Smaller, not blocking

- `diffViewerSourcePath` strips `a/` and then `b/` from the same string, so a real path like `a/b/c.go` comes back as `c.go`. It only picks the lexer, so the effect is cosmetic, but the second `TrimPrefix` is doing more than it appears to.
- The synthetic collapsed-context row leaves `rawIndex` at zero. Nothing reads it today because the `hiddenContext` case is matched first, but zero is a real row index, so a future reorder would quietly mis-highlight the first line.

## Not yours

`TestHandleAddDirCommand` also fails on `main` on my Windows box, so ignore it here. CI red on this PR is the repo-wide vulncheck outage, not your change. #903 fixes that.

anandh8x:
@coderabbitai please review only the latest unresolved change on 21636312 the blank hunk-context regression assertion. Approve if it is sufficient.

anandh8x:
@coderabbitai approve

gnanam1990 (dismissed):
## Verdict: Approve

Reviewed at `21636312`. Scope is clean (`internal/tui` plus the dependency bump, nothing else), the branch is current with `main`, and every check I ran is green.

## What I verified

**The dependency bump is not scope drift — it fixes a live, reachable vulnerability.** I checked rather than assumed, because a `go.mod` change in a diff-viewer PR is normally worth pushing back on. On clean `origin/main`:

```
Vulnerability #1: GO-2026-6222
    Excessive memory allocation during VP8L decoding in golang.org/x/image
    Found in: golang.org/x/image@v0.44.0
    Fixed in: golang.org/x/image@v0.45.0
      #1: internal/terminalpet/client.go:738:36: terminalpet.decodeImage calls
          image.Decode, which eventually calls vp8l.Decode
```

Reachable through `terminalpet`, so `govulncheck` fails on it — and since that job is a hard gate, **`main` is currently blocking every open PR on this**. #897 is already red on exactly this advisory; #891 only looks green because its run predates the publication. Worth landing promptly for that reason alone, independent of the diff viewer.

**Context folding is arithmetically correct.** Collapse triggers only at `count >= 8` (`count <= 3*2+1` keeps everything), then renders 3 + marker + 3 = 7 lines, and the marker's `count - 6` is exactly the skipped range. The comment explaining the threshold — collapsing seven would replace seven lines with seven lines — is the right reasoning and the boundary is tested at exactly 7.

**The toggle fix is the right shape.** `toolResultCanToggle` asks the body renderer (`body.canToggle`) whether an affordance was actually exposed, rather than matching display text. That matters because diff/tool output is untrusted: `TestToolResultOutputCannotCreateToggle` plants `"the file literally says: click to expand"` in tool output and asserts no toggle appears. Coupling selection behavior to the rendered card rather than to text is what makes that hold for input nobody anticipated.

**Test quality — mutation-checked, not just read.** I reverted each guard and confirmed the corresponding test fails:
- fold threshold `<=` → `<` — `TestDiffViewerKeepsSevenUnchangedContextLines` fails
- hidden count `count-6` → `count` — `TestDiffViewerCollapsesLongUnchangedContext` fails
- `toggle: toolResultCanToggle(...)` → `toggle: true` (the old behavior) — both toggle tests fail

**Adversarial input probe** (throwaway test, removed afterwards). Diff text is tool output, so I pushed malformed input through `diffCardBody` at widths 1/5/40/200: malformed and negative hunk ranges, a 20-digit start line, CRLF, hunk-without-header, headers-only, binary marker, wide CJK/emoji, embedded ANSI escapes, an embedded NUL, empty, and bare-blank context. No panic, no hang, no mis-render. A 50,000-line diff renders in 0.28s — the highlight budget bounds the expensive work, and the uncapped line count there is `bodyCap: 0`, the detailed view's documented contract, not a leak.

**Ran:** `make fmt-check`, `go vet ./internal/tui/`, `go build ./...`, `go test ./internal/tui/`, `go test -race -count=2 ./internal/tui/`, `GOOS=windows` build + test-binary compile, `zero-release build` + `smoke`, `git diff --check`. All clean on darwin/arm64.

## Non-blocking

- `diffHighlightBudget.reserve`'s first two conditions (`lines > max || bytes > max`) are subsumed by the cumulative checks that follow, since the budget only grows from zero. Harmless as a fast path — just noting it is not load-bearing.

## Not checked

The manual TUI verification claimed in the description — syntax-highlighted rendering and non-clickable headers are not reproducible headlessly, so I relied on the assertions in `diff_viewer_test.go` for the rendered output. Colour fidelity and interaction feel are unverified by me.

@Vasanthdev2004's `CHANGES_REQUESTED` from 08-14 predates the three commits on 08-15 (`7ec30011`, `bdc72429`, `21636312`), so it reads as stale rather than outstanding — worth a re-look to clear the gate.

Vasanthdev2004 (changes requested):
Re-reviewed at `21636312`. Both of my blockers from `f70481ee` are properly closed and I checked them rather than reading the commit titles: the near-rewrite deletion survives on `.log`, extensionless and `.go` paths, the oversized-line variant no longer eats unrelated deletions, and always-expanded diff cards still refuse to advertise a toggle. The `golang.org/x/image` bump is legitimate and correctly targeted, GO-2026-6222, fixed in exactly 0.45.0.

The fix for the first one over-corrected, though, and that is the blocker now.

This lands after gnanam's approval on this same head, which I would not normally step over. The duplication below reproduces on `21636312` with row counts rather than substring checks, and it is the kind of thing every existing assertion in the file is shaped to miss, so I think it is worth the second look rather than a disagreement about taste.

## Every deleted line is rendered twice whenever the highlighter produced a styled row

The `-` case appends the styled row and then falls straight through into the unconditional plain append below it:

```go
if styled, ok := highlightedLines[displayLine.rawIndex]; ok {
    lines = append(lines, diffBodyStyledLine(hunk.oldLine, "−", styled, ...))
} else if isIsolatedReplacement(...) {
    ...  // this arm continues on success
}
text := truncateRunes(strings.TrimPrefix(line, "-"), textBudget)
lines = append(lines, diffBodyLine(hunk.oldLine, "−", text, ...))   // runs after the styled arm too
```

The `+` case a few lines up has a correct `if/else`, so only deletions double. Counted rows rather than substrings:

```
example.go    rows=4  del=2 add=1 ctx=1
example.py    rows=4  del=2 add=1 ctx=1
notes.md      rows=4  del=2 add=1 ctx=1
notes.txt     rows=4  del=2 add=1 ctx=1
x.unknownzz   rows=3  del=1 add=1 ctx=1
```

Only the path chroma cannot lex escapes. Gutter numbers stay correct because the counters advance once, which is what makes it read as a rendering glitch rather than a bug.

It is not cosmetic once a card is capped. Eleven real source rows render as seventeen, and the duplicates push real content out:

```
  1 − deletedLine1
  1 − deletedLine1
  ...
  8 − deletedLine8
  8 − deletedLine8
… 5 more lines
```

The last two deletions and the trailing context are gone from the card entirely.

**Do not fix this by putting the trailing append back in an `else`.** That is exactly the shape I flagged last round: the word-diff arm has to keep falling through when `renderWordDiffPair` refuses a near-rewrite. Give the styled arm its own exit instead:

```go
if styled, ok := highlightedLines[displayLine.rawIndex]; ok {
    lines = append(lines, diffBodyStyledLine(hunk.oldLine, "−", styled, false, textBudget, gutter))
    hunk.oldLine++
    hunk.consume(line)
    continue
} else if isIsolatedReplacement(...) {
```

I applied that here and checked both directions: `del=1` on every extension, the near-rewrite case still renders its deletion exactly once on unlexable paths, and the rest of your suite passes.

Worth noting why the tests missed it. `TestDiffViewerNearRewriteDeletionSurvives` uses `notes.log`, which is one of the few extensions with no lexer, so it exercises the one path that was still correct. And every diff assertion in the file is `strings.Contains`, which a duplicate satisfies. A regression here has to assert an exact row count.

## Context folding runs over the whole payload, not just hunk bodies

`compactDiffViewerContext` collapses any run of more than seven lines starting with a space, with no notion of whether it is inside a hunk, and the `hiddenContext` case is matched before the preamble handling. `bash` and `exec_command` route through the diff renderer whenever the output looks like a diff, so `git log -p` and `git show --stat -p` go through it. Git indents commit-message bodies with four spaces, blank separator lines included, so the message is one contiguous run:

```
commit abc123
Author: someone
Date: today

    message line a
    message line b
    message line c
… 6 unchanged lines
    message line j
    message line k
    message line l
```

Six lines of somebody's commit message silently replaced by a fold marker. `--stat` blocks go the same way. Folding should only apply between a hunk header and the end of that hunk.

## Three more, lower

Multi-file diffs lose their file boundaries: the rows run together, the gutter restarts at 1 with nothing marking why, and the head names only the last file. A reader cannot tell which file a row belongs to.

Gaps *between* hunks in one file are unmarked while elisions *inside* a hunk are, so a jump from line 40 to line 900 looks continuous.

A hunk whose last source line is blank appears to disable highlighting for the whole card, and a lone CR can shift styled rows against source rows in a way the length guard does not catch. I have not reproduced either myself, so treat them as leads rather than findings.

## One design question

`toolResultCanToggle` renders the card body to decide whether it can toggle, uncached, and that runs per frame and per mouse event. For an expanded diff card that means re-lexing on every frame. Whatever the real number is, deciding interactivity by rendering is expensive as well as fragile. If the body renderer returned whether it produced a collapse affordance, the toggle question would be a field read and the footer-matching would go away with it.

Everything else in this round looks good, and the two things I raised last time were fixed properly rather than papered over.

gnanam1990 (approved):
## Verdict: Approve (re-review at `a34741ec`)

First, the part that matters more than the verdict: **@Vasanthdev2004 found a real bug on the exact head I had already approved, and he was right to step over that approval.** I reproduced his table before looking at the fix, on `21636312`:

```
example.go     del=2 add=1 ctx=1
example.py     del=2 add=1 ctx=1
notes.md       del=2 add=1 ctx=1
notes.txt      del=2 add=1 ctx=1
x.unknownzz    del=1 add=1 ctx=1
```

Every deletion rendered twice on any path chroma can lex. My approval missed it for exactly the reason he identified: I verified the folding arithmetic and the toggle guard by mutation, but I checked that things *appeared* rather than that they appeared **once** — and every assertion in the file is `strings.Contains`, which a duplicate satisfies. The one test covering the near-rewrite path uses `notes.log`, an extension with no lexer, so it exercised the single arm that was still correct. That is a good lesson and I have taken it.

## Both blockers are fixed, verified independently

**Duplicate deletions.** Fixed with the shape Vasanth prescribed — the styled arm gets its own exit (`hunk.oldLine++`, `hunk.consume(line)`, `continue`) rather than the trailing append being wrapped in an `else`, which would have broken the word-diff fallthrough when `renderWordDiffPair` refuses a near-rewrite. Re-ran the same row-count probe on `a34741ec`: `del=1 add=1 ctx=1` on all five paths, and the near-rewrite case still renders its deletion exactly once on both `notes.log` and `example.go`.

**Preamble folding.** Fixed by threading `diffHunkState` through `compactDiffViewerContext` so folding only applies inside a hunk. Reproduced the `git log -p` case first: on `21636312`, commit-message lines `d` through `i` were swallowed by a fold marker; on `a34741ec` all twelve survive.

**The new regressions bite.** `TestDiffViewerRendersStyledDeletionOnce` uses `strings.Count(...) != 1`, an exact row count as asked — reverting the early exit fails it with `row count = 2, want 1`.

## One structural note from mutating the fold guard (non-blocking)

Removing `!hunk.active()` from the outer condition does not fail `TestDiffViewerDoesNotFoldPreambleContext` — it **hangs**, and the run died on the 10-minute test timeout. The reason is that the outer guard and the inner loop condition are exact complements:

```go
if !hunk.active() || !strings.HasPrefix(raw[index], " ") { ...; index++; continue }
...
for index < len(raw) && hunk.active() && strings.HasPrefix(raw[index], " ") { ...; index++ }
```

Reaching the run-scan means both are true, so the inner loop advances `index` at least once. Correct as written. But if those two conditions ever drift, `count` is zero, nothing is appended and `index` never moves — an infinite render loop rather than a wrong row. A one-line defensive advance (`if index == start { index++; continue }`) would make termination structural instead of depending on the two predicates staying in lockstep. Not a defect today, just a sharp edge worth blunting while it is fresh.

## On the remaining items in Vasanth's review

The three "lower" ones (multi-file boundaries, unmarked gaps between hunks, the blank-last-line and lone-CR leads) and the `toolResultCanToggle` cost question are not addressed in this head. I do not think they should block: the first two are presentation gaps rather than wrong output, and he explicitly marked the last two as leads he had not reproduced. The toggle-cost point is real — rendering the body to decide interactivity does run per frame — but the fix he suggests (have the body renderer return whether it produced a collapse affordance) is a refactor of the same seam this PR just corrected, and is cleaner as its own change than bolted on here.

Worth agreeing on that explicitly rather than letting it slide silently, since it is a genuine design objection.

## Verified

Row-count probes on both heads (thrown away afterwards), mutation checks on both new regressions, `gofmt`, `go vet`, `go build ./...`, `go test ./internal/tui/`, `go test -race ./internal/tui/`, `GOOS=windows go build ./...`, and `make vulncheck` — **clean, including the `x/image` bump this PR carries**. All GitHub checks green.

This PR is still the unblocker for `GO-2026-6222` across the whole queue; #891, #897 and #829 are red on it right now.

kevincodex1 (approved):

PR #607, opened 2026-07-09
fix: bump Go to 1.26.5 for crypto/tls fix (GO-2026-5856)

kevincodex1:
## Summary

Bumps the Go toolchain from 1.26.4 to 1.26.5 to fix GO-2026-5856, an Encrypted
Client Hello privacy leak in the standard library's `crypto/tls` that was
failing CI's govulncheck job. Our code reaches the vulnerable paths through
normal TLS usage — the MCP OAuth callback server, dictation downloads, and the
daemon protocol/remote client — so any TLS use tripped it.

Since every workflow resolves the toolchain via `go-version-file: go.mod`,
bumping the `go` directive is the complete fix; no code changes needed.
Verified: `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...` now reports
**No vulnerabilities found** (this also clears the second, uncalled dependency
finding from the failing run).

## Checklist

- [ ] The linked issue already has the `issue-approved` label. (No linked issue — direct fix for the currently failing govulncheck CI job.)
- [x] `go build ./...`, `go vet ./...`, and `go test ./...` pass locally.
- [x] `gofmt` clean.
- [x] Tests added/updated for the change (and run under `-race` where relevant). (N/A — toolchain version bump only, no code changed; full existing suite passes on 1.26.5.)
- [x] UI changes include screenshots or a short recording where possible. (N/A — no UI changes.)

## Summary by CodeRabbit

* **Chores**
  * Updated the Go runtime version to 1.26.5.

PR #903, opened 2026-08-14
fix: bump Go to 1.26.6 for stdlib vulnerability fixes

Vasanthdev2004:
`make vulncheck` is currently failing on every open PR, with no code change behind it. `govulncheck` reads the live database, so six stdlib advisories landed and immediately turned the check red repo-wide.

```
GO-2026-6218  net/url          quadratic complexity in resolvePath
GO-2026-6090  crypto/tls       unbounded post-handshake messages
GO-2026-6089  net/http         ReadHeaderTimeout skipped on the h2c check
GO-2026-6088  encoding/xml     no recursion depth guard on decode
GO-2026-5972  encoding/asn1    no maximum recursion depth
GO-2026-5026  x/net/idna       ASCII-only punycode labels not rejected
```

All six are fixed in Go 1.26.6. We are on 1.26.5, so this bumps the `go` directive, same as #607 did for GO-2026-5856.

This is worth landing ahead of everything else in flight: `vulncheck` is the first step of both `Security & code health` and `Smoke (windows-latest)`, so it short-circuits the job before lint, vet, test and the binary smoke ever run. Right now those two required checks are red on every PR and are telling us nothing about the code in them.

Verified with the pinned `govulncheck@v1.3.0` in a clean worktree: 1.26.5 reports all six and exits 1, 1.26.6 reports none and exits 0.

The README and docs/INSTALL.md still say "Go 1.26.5+". Left alone here to keep this to the one line that unblocks CI, happy to follow up.

## Summary by CodeRabbit

* **Chores**
  * Updated the Go language version to 1.26.6 for improved compatibility and maintenance.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/Gitlawb/zero
 
-go 1.26.5
+go 1.26.6
 
 require (
 	charm.land/bubbles/v2 v2.1.1


anandh8x (approved):
Reviewed the current head: this is the targeted Go 1.26.6 patch update, and all platform smoke plus security checks are green.

gnanam1990 (approved):

PR #910, opened 2026-08-15
feat(update): recognise Homebrew installs and add mise, go install docs

Vasanthdev2004:
The r/golang thread about Zero asked for install channels other than npm. Most of them already exist and the docs just do not say so, one of them is a genuine gap, and the gap has a prerequisite that is easy to miss.

## The prerequisite

`DetectInstallMethod` knows `npm` and `standalone`. A `brew install` is neither, so it reads as standalone and `zero upgrade` happily rewrites the binary inside the keg. Homebrew's records then describe a version that is not on disk, and the next `brew upgrade` or `brew reinstall` reverts it. Publishing a tap before fixing that ships the bug on day one.

So this adds `InstallMethodHomebrew`. `zero upgrade` refuses on it and names `brew upgrade zero`, and the check-time guidance says the same thing rather than offering a command that will not work.

Detection is the Cellar segment, not the Homebrew prefix. The prefix on Intel macOS is `/usr/local`, which is also where people put hand-installed binaries, so matching the prefix would refuse updates for installs Homebrew has never heard of. Every keg is under Cellar and nothing else is. A missed detection leaves self-update working, which is the recoverable direction; a false positive breaks a working feature.

`DetectInstallMethod` also resolves symlinks now. `Check` did not and `Apply` did, so the two disagreed about the same install: Homebrew links `<prefix>/bin/zero` to the keg, and the check would have printed standalone guidance for an install the apply path treated differently.

## Docs

INSTALL.md opens with a table of every method and how each one updates, instead of leading with npm. Two of them work today and were simply never written down:

- **mise**, straight off the release archives through the `ubi` backend. The assets are already named `zero-v{version}-{os}-{arch}`, which is what ubi needs, so there is nothing to publish.
- **`go install`**, with the two caveats that matter: on Linux the sandbox helper is a separate binary this does not install, and `zero upgrade` will replace a source build with a release build.

## The formula

`packaging/homebrew/zero.rb` is a draft and installs nothing. It is not a tap and the release workflow does not know about it. Both of those are release decisions, so I have not made them.

It is real rather than a skeleton: the checksums are the published v0.7.0 ones, read from the `.sha256` files beside each archive, so it can be exercised now:

```bash
brew install --build-from-source ./packaging/homebrew/zero.rb
brew audit --strict --formula ./packaging/homebrew/zero.rb
```

I checked the archive layout rather than assuming it. `zero`, `zero-linux-sandbox` and `zero-seccomp` sit at the archive root, so `bin.install "zero"` is right and the sandbox helper is installed when the platform archive carries one. Left for whoever owns the tap: where it lives, and teaching the release workflow to bump the version and the four checksums on each tag.

## Checks

`gofmt`, `go vet`, `go build ./...`, `internal/update` and `internal/installtest` all clean.

Falsified rather than assumed: with `isHomebrewPath` forced to false and the guidance branch removed, the new tests fail naming exactly what is missing, and the guidance test prints the string a Homebrew user would otherwise get, ``Run `zero upgrade` to download, verify, and install the latest release.``, which is the bug.

The path rule is a table test over `darwin`, `linux` and `windows` rather than a check gated on `runtime.GOOS`, so it verifies the Windows branch from any machine. The one test that still has to skip on Windows says which test covers it instead.

One thing worth confirming while reviewing: I checked whether a plain archive install could be misread as npm, since the archive ships a `package.json` next to the binary. It cannot. That manifest carries a `bin` entry and broad `os`/`cpu` lists, which is exactly the case the existing guard rejects.

## Summary by CodeRabbit

* **New Features**
  * Added Homebrew installation support for macOS and Linux, including architecture-specific packages and optional Linux sandbox support.
  * Added installation guidance for scripts, npm, release archives, mise, Go, and source builds.
* **Bug Fixes**
  * Improved installation detection and prevented direct Homebrew binary updates.
  * Homebrew upgrades now direct users to `brew upgrade zero`.
* **Documentation**
  * Documented version pinning, installation-specific upgrade behavior, and source-build requirements.
  * Updated the minimum required Go version to 1.26.6.


Vasanthdev2004:
Went through all five at `5763eb2d`. Three taken, one taken in a narrower form, one declined with a test.

**1. The self-update summary contradicted the mise section.** Taken, and it was my bug: the summary claimed `zero upgrade` never fights a package manager while my own mise section said it will happily replace a mise-managed binary. Rewritten to say what actually happens per method, and to name mise as the one case it cannot detect, because a mise-managed binary is an ordinary standalone install on disk.

**2. Go 1.26.5+ in the source-build section.** Taken, and it was stale in more places than the one flagged: `README.md` and `README_ZH.md` carry it in prose and in the badge. All updated to 1.26.6 to match `go.mod` after #903.

**3. No Apply regression for a Homebrew keg.** Taken. `TestApplyRefusesToUpdateAHomebrewKeg` builds a keg, points `currentExecutable` at it, runs `Apply` against a data: release endpoint, and asserts three things: it errors, the error names `brew upgrade zero`, and the binary on disk is byte-identical afterwards. The last one is the point, since the failure being guarded is a write that should not happen.

**4. Match more than a bare Cellar segment.** Taken in part. `isHomebrewPath` now requires the keg shape, `<prefix>/Cellar/<formula>/<version>/` and at least one component beneath, so a user directory that happens to be called Cellar no longer matches. Two cases added to the table test, and reverting the shape check makes both fail:

```
isHomebrewPath("linux", "/home/someone/Cellar/zero") = true, want false
isHomebrewPath("darwin", "/Users/someone/Cellar/bin/zero") = true, want false
```

Not taken: reading Homebrew's receipt metadata, and failing closed when ownership evidence is absent. Two reasons, both written into the comment. Receipt metadata puts filesystem I/O on a path that runs on every version check, to tighten a case the path shape already covers. And failing closed is the wrong default here: a missed detection leaves self-update working, which the user can recover from, while a false positive refuses to update an install Homebrew never touched, which they cannot. I would rather under-claim Homebrew than over-claim it.

**5. Move the Homebrew branch ahead of the cross-target guidance.** Declined, and I think this one would introduce a bug rather than fix one.

Homebrew is a property of the binary on *this* machine. When the check was asked about a different release target, the answer is about *that* machine, and `brew upgrade zero` would change this one instead. That is the exact trap the existing comment on `upgradeGuidance` describes: a cross-target check must not answer a question about one machine with an action that changes another.

Rather than argue it, I made the ordering a test. `TestUpgradeGuidanceKeepsCrossTargetAnswerForHomebrew` asserts a cross-target check on a Homebrew install still gets cross-target guidance. Applying the suggested reorder makes it fail with:

```
a question about linux-arm64 was answered with a command that changes this machine:
"This Homebrew-managed installation is updated with `brew upgrade zero`. ..."
```

The ordering reads like a bug until you see which question is being answered, so it is now pinned with a comment saying so.

`gofmt`, `go vet`, `go build ./...` and `internal/update` all clean. Two tests skip on Windows because `DetectInstallMethod` is GOOS-gated by design; the rule itself is covered from any machine by the table test, and macOS and ubuntu Smoke run the rest.

Vasanthdev2004:
Addressed. Two of the three taken, one declined with reasons.

**Homebrew ownership — taken, and you were right.** The shape check alone was too loose. Two changes: the depth rule now requires the full `<prefix>/Cellar/<formula>/<version>/bin/<binary>`, which already excludes the exact path you raised (`/work/Cellar/tools/bin/zero` is one segment short). Then, only for a path that still looks like a keg, one Stat for `INSTALL_RECEIPT.json` — the file Homebrew writes into every keg and nothing else does.

The old comment argued against the receipt because it would put filesystem I/O on a path that runs at every version check. That is no longer true of it: the Stat happens only after the shape test has matched, so an ordinary install still touches the filesystem zero times here. Regression test added for a user-owned Cellar-shaped tree, plus one for a *directory* named `INSTALL_RECEIPT.json`, which a bare Stat would have accepted silently.

**Guidance ordering — declined.** That one is deliberate and already pinned by `TestUpgradeGuidanceKeepsCrossTargetAnswerForHomebrew`, which fails if the branch moves. Its comment anticipated this exact reading: *"reads like an ordering bug until you see which question is being answered."*

Homebrew is a property of the binary on THIS machine. When the check is asked about a different target the answer belongs to that other machine, and `brew upgrade zero` would change this one instead — so answering a linux-arm64 question with a local brew command is the bug, not the fix. There is no safety violation either: the cross-target guidance says to replace the binary *on that machine*, never to overwrite the local keg. I have recorded the reasoning at the branch as well as in the test, so the next reader finds it in both places rather than re-deriving it.

**Docs — taken.** The summary stated the rule absolutely and excepted mise in the following paragraph, which a reader could stop before; it now carries the exception in the sentence itself. Also fixed a claim I had left understated — detection was described as the Cellar path alone, which is no longer what it does.

`go build`, `go vet`, `go test ./internal/update` all pass; `gofmt` clean.

Vasanthdev2004:
@coderabbitai approve

kevincodex1 (approved):

PR #909, opened 2026-08-15
feat(measurements): check a report's timings against the run's own

gnanam1990:
## Split out of #829 — independent package

Fourth piece of the split @Vasanthdev2004 asked for. **Not stacked on #891/#897** — it builds and tests against current `main` on its own.

## What it is for

A measured run finished a benchmark and reported a table of test timings **that no command in the session had produced**:

- the same test read `0.86s` in one paste and `4.20s` in the next, with nothing said about the difference
- a `-race` overhead moved from `+3.7%` to `+133%` between two tellings of the same result
- the column summed to an exact total no real transcript lands on

## Why a prompt rule is not the fix

"Re-run every command before you paste it" is the obvious answer and the weak one: a model willing to write numbers it did not measure is equally willing to say it re-ran them. The check has to live somewhere the model cannot assert its way past.

The harness qualifies. Every command's output passed through this process and was written to the session log, so the run's real numbers are already there — this package reads them back and compares them against what the answer claims.

## Deliberately loose

Timings vary for honest reasons: a loaded machine, a warm cache, a different `-count`. The tolerance is a **50% band**, which lets ordinary variation through and still catches `0.86s` reported as `4.20s`.

That asymmetry is on purpose. A tripwire that cries wolf gets turned off, and then it catches nothing; a false negative costs one uncaught number. So it errs firmly toward silence.

## Note on importers

None in this PR, by design — `internal/agent` and `internal/specialist` adopt it with the orchestration work, the same shape as `internal/pathjail` arriving in #891 ahead of its adopters.

`gofmt`, `go vet`, `go build ./...`, `go test ./internal/measurements/` — clean on current `main`.

Part of #829.

## Summary by CodeRabbit

* **New Features**
  * Detects discrepancies between reported and observed Go test durations.
  * Supports hours, minutes, seconds, milliseconds, compound formats, and signed differences.
  * Compares measurements within individual test runs and across runs.
  * Includes command context and test/package details in correction messages.

* **Bug Fixes**
  * Reduces duplicate reports and false matches for similarly named tests.
  * Ignores malformed, unsupported, benchmark, cached, or unrelated timing data.
  * Improves attribution across clauses, repeated claims, and neighboring test names.

* **Tests**
  * Added comprehensive coverage for parsing, concurrency, deduplication, duration matching, and run-specific measurements.

PR #911, opened 2026-08-16
fix(sandbox): refcount temporary roots so one holder's cleanup cannot revoke another's

gnanam1990:
## Split out of #829 — independent concurrency fix

Fifth piece of the split @Vasanthdev2004 asked for, and one of the fourteen changes he enumerated ("sandbox temporary-root refcounting"). **Not stacked on anything** — builds and tests against current `main` on its own.

## The bug

A temporary root grant was add-then-remove with no notion of who still needed it. So:

1. Caller A asks for a temporary read on `/some/dir` → root added, A gets a real undo
2. Caller B asks for the same root, already present → B gets a **no-op undo**
3. A finishes and releases → the root is **stripped while B still holds a live grant**

B then believes it has access it has already silently lost.

**Two read-only tools in the same parallel batch, both blocked on the same directory, is exactly that shape** — and read-only tools are precisely the ones the batch runs concurrently, so this is the ordinary path rather than a corner.

## The fix

`tempReads`/`tempWrites` count live holders; the root is stripped only when the last one releases.

The release also does both mutations under **one** hold of the lock. Dropping it between them left a window where the root was still in `readRoots` but no longer in `tempReads` — a concurrent `AddTemporaryRead` landing there reads it as a *permanent* root, hands its caller a no-op undo, and this call then strips the root. Same silent loss, reached a different way.

## Verification

Mutation-checked: making the release always strip (the pre-fix behaviour) fails both regressions —

```
--- FAIL: TestReleasingATemporaryReadCannotStripALiveGrant
--- FAIL: TestTemporaryReadGrantsSurviveConcurrentReleases
```

`gofmt`, `go vet`, `go build ./...`, `go test ./internal/sandbox/`, and `go test -race -count=2` — all clean on current `main`.

Part of #829.

## Summary by CodeRabbit

* **New Features**
  * Added support for temporary read-only and read/write sandbox access.
  * Read-only grants allow reading without permitting modifications.

* **Bug Fixes**
  * Temporary access now remains available until all active holders release it.
  * Overlapping and nested grants no longer revoke valid access prematurely.
  * Permanent access remains available during temporary-access cleanup.
  * Repeated releases are handled safely.
  * Read access no longer retains write permissions after write access ends.

### DIFF
--- a/internal/sandbox/scope.go
+++ b/internal/sandbox/scope.go
@@ -126,13 +126,47 @@ func (s *Scope) Add(path string) (string, error) {
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	// COVERED BY WHAT. extraRoots holds temporary write roots alongside
+	// permanent ones, so "something already covers this" was not the question
+	// worth asking: a permanent grant made over a path a temporary holder
+	// happened to cover recorded nothing, and vanished the moment that holder
+	// released. A session-scoped grant outliving the request that prompted it is
+	// the entire difference between Add and AddTemporaryWrite.
+	//
+	// Permanent coverage is looked for FIRST and across every root, because a
+	// path can be covered twice and the temporary cover must not decide the
+	// answer when a permanent one is also present.
 	for _, existing := range append([]string{s.workspaceRoot}, s.extraRoots...) {
-		if pathWithinRoot(existing, root) {
+		if pathWithinRoot(existing, root) && !s.temporaryWriteLocked(existing) {
 			return root, nil
 		}
 	}
+	for _, existing := range s.extraRoots {
+		// The same root, held temporarily: promote it in place. Its holders'
+		// releases become no-ops, which is what permanence means here.
+		if existing == root && s.temporaryWriteLocked(existing) {
+			delete(s.tempWrites, root)
+			return root, nil
+		}
+	}
+	// Either uncovered, or covered only by a BROADER temporary root that is
+	// going to be released. Recording the narrower root in its own right is what
+	// survives that release.
 	s.extraRoots = append(s.extraRoots, root)
 	return root, nil
+}
+
+// temporaryWriteLocked reports whether root is held by a temporary write grant
+// rather than a session-scoped one. Callers must hold the lock.
+func (s *Scope) temporaryWriteLocked(root string) bool {
+	_, temporary := s.tempWrites[root]
+	return temporary
+}
+
+// temporaryReadLocked is temporaryWriteLocked for read roots.
+func (s *Scope) temporaryReadLocked(root string) bool {
+	_, temporary := s.tempReads[root]
+	return temporary
 }
 
 // AddRead grants read-only access under path. If the path is already covered by
@@ -144,11 +178,23 @@ func (s *Scope) AddRead(path string) (string, error) {
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if s.writeRootCoversLocked(root) {
-		return root, nil
+	// Same rule as Add: coverage that is going to be released is not coverage a
+	// permanent grant can rely on, whichever side of the read/write boundary it
+	// sits on. A permanent read covered only by a temporary WRITE root died with
+	// that root just as surely.
+	for _, existing := range append([]string{s.workspaceRoot}, s.extraRoots...) {
+		if pathWithinRoot(existing, root) && !s.temporaryWriteLocked(existing) {
+			return root, nil
+		}
 	}
 	for _, existing := range s.readRoots {
-		if pathWithinRoot(existing, root) {
+		if pathWithinRoot(existing, root) && !s.temporaryReadLocked(existing) {
+			return root, nil
+		}
+	}
+	for _, existing := range s.readRoots {
+		if existing == root && s.temporaryReadLocked(existing) {
+			delete(s.tempReads, root)
 			return root, nil
 		}
 	}
@@ -176,7 +222,7 @@ func (s *Scope) AddTemporaryRead(path string) (string, func(), error) {
 			if pathWithinRoot(existing, root) {
 				s.tempWrites[existing]++
 				covering := existing
-				return root, func() { s.releaseTemporaryWrite(covering) }, nil
+				return root, oncePerHolder(func() { s.releaseTemporaryWrite(covering) }), nil
 			}
 		}
 		// Genuinely permanent — the workspace root, or a session-scoped grant.
@@ -196,7 +242,7 @@ func (s *Scope) AddTemporaryRead(path string) (string, func(), error) {
 		if _, temporary := s.tempReads[existing]; temporary {
 			s.tempReads[existing]++
 			covering := existing
-			return root, func() { s.releaseTemporaryRead(covering) }, nil
+			return root, oncePerHolder(func() { s.releaseTemporaryRead(covering) }), nil
 		}
 		return root, func() {}, nil
 	}
@@ -205,7 +251,7 @@ func (s *Scope) AddTemporaryRead(path string) (string, func(), error) {
 		s.tempReads = map[string]int{}
 	}
 	s.tempReads[root] = 1
-	return root, func() { s.releaseTemporaryRead(root) }, nil
+	return root, oncePerHolder(func() { s.releaseTemporaryRead(root) }), nil
 }
 
 func (s *Scope) AddTemporaryWrite(path string) (string, func(), error) {
@@ -224,7 +270,7 @@ func (s *Scope) AddTemporaryWrite(path string) (string, func(), error) {
 			if pathWithinRoot(existing, root) {
 				s.tempWrites[existing]++
 				covering := existing
-				return root, func() { s.releaseTemporaryWrite(covering) }, nil
+				return root, oncePerHolder(func() { s.releaseTemporaryWrite(covering) }), nil
 			}
 		}
 		return root, func() {}, nil
@@ -234,7 +280,7 @@ func (s *Scope) AddTemporaryWrite(path string) (string, func(), error) {
 		s.tempWrites = map[string]int{}
 	}
 	s.tempWrites[root] = 1
-	return root, func() { s.releaseTemporaryWrite(root) }, nil
+	return root, oncePerHolder(func() { s.releaseTemporaryWrite(root) }), nil
 }
 
 func (s *Scope) writeRootCoversLocked(root string) bool {
@@ -246,11 +292,17 @@ func (s *Scope) writeRootCoversLocked(root string) bool {
 	return false
 }
 
-// releaseTemporaryRead drops one holder's reference and removes the root only
-// when the last one is gone. IDEMPOTENT per holder is not the property here —
-// each undo is called exactly once — but a double call must not remove a root
-// another holder still needs, so the count floors at zero rather than going
-// negative.
+// releaseTemporaryRead drops ONE holder's reference and removes the root only
+// when the last one is gone.
+//
+// The count alone cannot make this safe, and an earlier comment here claimed it
+// could — that each undo is called exactly once, and flooring at zero handled
+// the rest. Flooring stops the count going negative; it does not stop one
+// holder's second call consuming a DIFFERENT holder's reference. With two
+// readers, calling the first's undo twice took the count 2 -> 1 -> 0 and removed
+// a root the second was still using. Idempotency belongs to the closure, which
+// is the thing that knows whose reference it is, so every undo handed out is
+// wrapped in oncePerHolder.
 func (s *Scope) releaseTemporaryRead(root string) {
 	s.mu.Lock()
 	remaining, tracked := s.tempReads[root]
@@ -295,6 +347,14 @@ func (s *Scope) releaseTemporaryWrite(root string) {
 	delete(s.tempWrites, root)
 	s.extraRoots = removeScopeRoot(s.extraRoots, root)
 	s.mu.Unlock()
+}
+
+// oncePerHolder makes one holder's undo safe to call more than once. The
+// reference it drops is that holder's own, so a second call must do nothing
+// rather than reach into the shared count and take somebody else's.
+func oncePerHolder(release func()) func() {
+	var once sync.Once
+	return func() { once.Do(release) }
 }
 
 func removeScopeRoot(roots []string, root string) []string {
--- a/internal/sandbox/scope.go
+++ b/internal/sandbox/scope.go
@@ -164,7 +164,23 @@ func (s *Scope) AddTemporaryRead(path string) (string, func(), error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.writeRootCoversLocked(root) {
-		// A write root already covers it, permanently. Nothing to release.
+		// Covered by a write root — which is NOT necessarily permanent.
+		// writeRootCoversLocked scans extraRoots, and AddTemporaryWrite appends
+		// TEMPORARY write roots there, so the covering grant may belong to
+		// another holder who is going to release it. Take a reference in that
+		// case, exactly as the read-covered-by-read path below does: without one
+		// the write holder's cleanup silently revokes this reader's access —
+		// the same defect this refcount exists to close, reached across the
+		// read/write boundary rather than within one side of it.
+		for existing := range s.tempWrites {
+			if pathWithinRoot(existing, root) {
+				s.tempWrites[existing]++
+				covering := existing
+				return root, func() { s.releaseTemporaryWrite(covering) }, nil
+			}
+		}
+		// Genuinely permanent — the workspace root, or a session-scoped grant.
+		// Nothing to release.
 		return root, func() {}, nil
 	}
 	for _, existing := range s.readRoots {
--- a/internal/sandbox/scope.go
+++ b/internal/sandbox/scope.go
@@ -19,6 +19,17 @@ type Scope struct {
 	workspaceRoot string
 	readRoots     []string
 	extraRoots    []string
+	// tempReads / tempWrites count how many LIVE temporary grants depend on a
+	// root, so one holder's cleanup cannot revoke another's access.
+	//
+	// Without them a temporary grant was add-then-remove with no notion of who
+	// still needed it: the second caller to ask for a root already present got a
+	// NO-OP undo, and the first caller's cleanup removed the root out from under
+	// it. Two read-only tools in the same parallel batch, both blocked on the
+	// same directory, is exactly that shape — and read-only tools are precisely
+	// the ones the batch runs concurrently.
+	tempReads  map[string]int
+	tempWrites map[string]int
 }
 
 // NewScope builds a scope for workspaceRoot plus the given extra roots. The
@@ -48,6 +59,19 @@ func (s *Scope) WorkspaceRoot() string {
 }
 
 // Roots returns the workspace root first, then the extra roots, as a copy.
+// ExtraRoots returns ONLY the roots granted beyond the workspace, as a copy.
+//
+// Roots() includes the workspace root as well, which is right for a caller
+// asking "everything this run may write". It is wrong for a caller asking "what
+// does this run hold BEYOND its workspace" — and one such caller launches child
+// agents in an isolated worktree, where handing back the parent's workspace root
+// re-opens the very tree the worktree exists to protect.
+func (s *Scope) ExtraRoots() []string {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return append([]string(nil), s.extraRoots...)
+}
+
 func (s *Scope) Roots() []string {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
@@ -65,6 +89,27 @@ func (s *Scope) ReadRoots() []string {
 	defer s.mu.RUnlock()
 	roots := make([]string, 0, 1+len(s.extraRoots)+len(s.readRoots))
 	roots = append(roots, s.workspaceRoot)
+	roots = append(roots, s.extraRoots...)
+	roots = append(roots, s.readRoots...)
+	return dedupeScopeRoots(roots)
+}
+
+// ExtraReadRoots returns every path the run may READ BEYOND its workspace — the
+// write grants AND the read-only grants, deduped, as a copy. It is ReadRoots()
+// without the workspace root.
+//
+// It exists for the same caller ExtraRoots() does: a child agent dispatched into
+// its own workspace, which the parent must be able to hand its beyond-workspace
+// access without also handing back the parent workspace root (see ExtraRoots).
+// ExtraRoots() alone is not enough for that child, because a read grant from
+// request_permissions lands in readRoots, not extraRoots — so a child given only
+// ExtraRoots() can read what the parent may WRITE beyond its workspace but not
+// what it was granted to READ, and a read-only audit of a granted path fails
+// "outside the workspace" for want of exactly this list.
+func (s *Scope) ExtraReadRoots() []string {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	roots := make([]string, 0, len(s.extraRoots)+len(s.readRoots))
 	roots = append(roots, s.extraRoots...)
 	roots = append(roots, s.readRoots...)
 	return dedupeScopeRoots(roots)
@@ -119,15 +164,32 @@ func (s *Scope) AddTemporaryRead(path string) (string, func(), error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.writeRootCoversLocked(root) {
+		// A write root already covers it, permanently. Nothing to release.
 		return root, func() {}, nil
 	}
 	for _, existing := range s.readRoots {
-		if pathWithinRoot(existing, root) {
-			return root, func() {}, nil
+		if !pathWithinRoot(existing, root) {
+			continue
 		}
+		// Already covered — but by WHAT decides whether this caller has
+		// something to release. A permanent read root outlives every grant, so
+		// the undo is genuinely nothing. A TEMPORARY one is held by another
+		// caller who will release it, and that caller must not be able to
+		// revoke this one's access: take a reference on the covering root, so
+		// it survives until the last holder is done.
+		if _, temporary := s.tempReads[existing]; temporary {
+			s.tempReads[existing]++
+			covering := existing
+			return root, func() { s.releaseTemporaryRead(covering) }, nil
+		}
+		return root, func() {}, nil
 	}
 	s.readRoots = append(s.readRoots, root)
-	return root, func() { s.removeReadRoot(root) }, nil
+	if s.tempReads == nil {
+		s.tempReads = map[string]int{}
+	}
+	s.tempReads[root] = 1
+	return root, func() { s.releaseTemporaryRead(root) }, nil
 }
 
 func (s *Scope) AddTemporaryWrite(path string) (string, func(), error) {
@@ -138,10 +200,25 @@ func (s *Scope) AddTemporaryWrite(path string) (string, func(), error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.writeRootCoversLocked(root) {
+		// Covered already. If a TEMPORARY write grant is what covers it, take a
+		// reference so the covering holder's cleanup cannot revoke this one —
+		// the same rule as reads, and it has to be the same rule or a write
+		// grant would keep the bug reads no longer have.
+		for existing := range s.tempWrites {
+			if pathWithinRoot(existing, root) {
+				s.tempWrites[existing]++
+				covering := existing
+				return root, func() { s.releaseTemporaryWrite(covering) }, nil
+			}
+		}
 		return root, func() {}, nil
 	}
 	s.extraRoots = append(s.extraRoots, root)
-	return root, func() { s.removeWriteRoot(root) }, nil
+	if s.tempWrites == nil {
+		s.tempWrites = map[string]int{}
+	}
+	s.tempWrites[root] = 1
+	return root, func() { s.releaseTemporaryWrite(root) }, nil
 }
 
 func (s *Scope) writeRootCoversLocked(root string) bool {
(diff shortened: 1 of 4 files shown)


gnanam1990:
@Vasanthdev2004 @anandh8x — review please. This is "sandbox temporary-root refcounting" from your list of fourteen, @Vasanthdev2004, carved out on its own. 537 lines, independent of everything else still in #829, builds and tests against current `main`.

The part I would most want a second opinion on is the **second** failure mode rather than the refcount itself. The refcount is straightforward. But the release also had to move both mutations under one hold of the lock, because dropping it between them leaves a window where the root is still in `readRoots` and already gone from `tempReads` — and a concurrent `AddTemporaryRead` landing exactly there reads it as a *permanent* root and hands its caller a no-op undo. Same silent loss of access, reached a different way, and it is the kind of window that a test can pass over by luck.

`TestTemporaryReadGrantsSurviveConcurrentReleases` covers it and the mutation fails it, but a concurrency window is exactly where I would rather have a sceptical reader than a green test.

All checks green.

anandh8x (changes requested):
The core refcount direction is right and the focused tests pass under the race detector, but two lifetime bugs remain:

1. **[P1] A permanent grant must promote temporary coverage.** `Add`/`AddRead` return early when an existing temporary root covers the requested path, without recording permanence. I added a temporary write, called `Add` for the same path, then released the temporary grant; write access disappeared. The equivalent temporary-read followed by `AddRead` also lost access. Track permanent and temporary ownership separately so releasing the temporary holder cannot revoke the session grant, including broader-temporary/narrower-permanent coverage.

2. **[P1] Cleanup must be idempotent per holder.** Returned closures decrement a shared count on every invocation. With two live readers, calling the first reader's cleanup twice removes the root while the second still holds it—contradicting the function comment and revoking live access. Wrap each returned cleanup in per-holder `sync.Once` semantics (while keeping the shared root count).

All current CI checks are green on `87a722c`.

Vasanthdev2004 (dismissed):
Reviewed at `4a7063aa`. The refcounting is the right shape and the portability problem is already fixed, so this is an approve.

I started on `87a722cb`, where five of the ten new tests skipped on everything except macOS:

```
--- SKIP: TestATemporaryReadSurvivesASiblingsCleanup
--- SKIP: TestANarrowerRequestHoldsTheCoveringGrant
--- SKIP: TestATemporaryGrantNeverRevokesAPermanentRoot
--- SKIP: TestATemporaryWriteSurvivesASiblingsCleanup
--- SKIP: TestConcurrentHoldersOfOneRoot
```

`scopeOutsideRoots` looked for `/Users/Shared` and fell back to `/var/empty`, so ubuntu skipped too and the whole write-side refcount path had no coverage on two of the three CI platforms. You fixed it in `a12050e0` before I could post. All eleven run and pass here now.

Worth saying since I have done this three times myself this week: a test that skips is a test you have not run, and the skip reason is usually the last place anyone looks. The fixture being the platform-bound part rather than the assertion is what makes it easy to miss.

The refcounting itself holds up under the questions that matter for a write grant handed to a sandboxed command. A grant outliving its holder is the expensive direction and I could not produce one: a narrower request pins the covering root, a sibling's cleanup does not strip a live grant, and a temporary grant never revokes a permanent root.

## One thing I could not confirm, recorded rather than raised

The review surfaced a claim that a session-scoped READ grant is discarded when a temporary grant already covers the same path, and so dies at the turn boundary with it. I could not reproduce it at the Scope level on this head: taking a temporary read, then a permanent read on the same root, then releasing the temporary one leaves the read allowed. `scope.go` is byte-identical between the two heads, so this is not something your last commit fixed.

The claim was made against the Engine path rather than raw Scope, and I did not chase it that far. Not blocking on something I could not stand up, but if you know that path well it is worth five minutes: the shape would be `AddRead` returning early as "already covered" when the covering root is temporary, which is the case `AddTemporaryRead` was specifically taught to handle in this PR.

gnanam1990:
@anandh8x @Vasanthdev2004 — both P1s fixed, head `4a7063aa`, CI green 6/6, race detector clean. Both reproduced first, and they are the same mistake in two places: treating "something covers this path" as though it answered "and will it still be there afterwards".

**1. Permanent grants.** `extraRoots` holds temporary write roots alongside permanent ones, so `Add` only asked whether the path was already covered — a session-scoped grant made while a temporary holder happened to cover it recorded nothing and vanished when that holder released. Outliving the request that prompted it is the entire difference between `Add` and `AddTemporaryWrite`. `AddRead` had it too, including across the boundary: a permanent read covered only by a temporary *write* root died with that root.

Permanent coverage is looked for first and across every root, so a temporary cover cannot decide the answer when a permanent one is also present. The same root held temporarily is promoted in place. Your broader-temporary/narrower-permanent case is the one a promotion-only fix would have missed — the narrower root has to be recorded in its own right, and it now is.

**2. Cleanup idempotency.** You were right, and the comment on `releaseTemporaryRead` asserted the opposite: that each undo is called exactly once and flooring at zero handled the rest. Flooring stops the count going negative; it does nothing about a holder's second call consuming a *different* holder's reference. Every undo handed out is now wrapped so it drops its own reference at most once, and the comment says what is actually true.

Separately, CodeRabbit's two findings on the earlier head are also fixed here, and one is worth flagging because it means this PR's concurrency claim was previously untested: **the concurrency test was not concurrent.** The gate spun until `hasReadRoot` saw *somebody* holding the root, which the first holder satisfies. Instrumented over 200 runs, 194 peaked at one simultaneous holder and none reached eight. Against a deliberately broken refcount it passed 33 times in 40. Counting all eight acquisitions: 200/200 now peak at eight, and the broken refcount is caught 40/40. Its helper also skipped outright on Windows (`/Users/Shared`, `/var/empty`), so every test built on it reported green having asserted nothing.

Both of your findings mutation-checked: unwrapping the undo, and ignoring permanence in `Add`, each fail the test that covers them.

gnanam1990:
@anandh8x — head `4a7063aa` (unchanged since your review; @Vasanthdev2004 approved it there). Re-requesting because your two P1s are fixed on this head and I have since attacked it further.

Both fixed as described: permanent coverage is looked for first and across every root so a temporary cover cannot decide the answer; the same root held temporarily is promoted in place; a **broader** temporary over a narrower permanent is recorded in its own right, which is the case a promotion-only fix would have missed. And every undo now drops its own reference at most once — you were right that flooring the count only stops it going negative.

Running the review protocol over this head, I attacked the lifetime logic with cases beyond your two, under `-race`:

- two temporary holders, then `Add`, then both releases
- a temporary grant taken **after** promotion — its release must not re-arm
- a permanent **read** promoted while a temporary **write** root covers it
- eight holders, each calling its undo twice, concurrently
- a narrow temporary released inside a broad permanent root

All pass. No new defects found here.

**Not verified:** Windows. The `%TEMP%`/`%TMP%` roots and the junction handling are covered by CI and inspection, not by a local run — I am on darwin.

Also fixed on this head, from CodeRabbit, and worth knowing because it changes how much the earlier green CI was worth: **the concurrency test was not concurrent.** The gate spun until `hasReadRoot` saw somebody holding the root, which the first holder satisfies. Instrumented over 200 runs, 194 peaked at one simultaneous holder and none reached eight; against a deliberately broken refcount it passed 33 times in 40. It now counts all eight acquisitions — 200/200 peak at eight, and the broken refcount is caught 40/40.

anandh8x (dismissed):
Reviewed latest head `4a7063a`. Permanent grants now survive temporary coverage, narrower grants survive broader temporary roots, and every returned undo is idempotent per holder. The acquisition barrier also makes the concurrency regression exercise genuinely overlapping holders. Focused sandbox tests pass under the race detector. Good to merge after normal CI requirements.

gnanam1990:
@coderabbitai full review

Your last review was against a commit this branch no longer carries — it has been force-pushed and moved on since, and the findings you raised have been addressed. Please re-review the current head.

gnanam1990:
@coderabbit full review

Your last review was against a commit this branch no longer carries — it has been force-pushed and moved on since, and the findings you raised have been addressed. Please re-review the current head.

gnanam1990:
@coderabbitai full review

This branch was force-pushed after your last review, which was against a commit that is no longer on it. Both human reviewers have approved the current head and your stale verdict is the only thing outstanding — please re-review.

gnanam1990:
@coderabbitai review

gnanam1990:
@coderabbitai review

gnanam1990:
@coderabbitai review

gnanam1990:
@kevincodex1 — same situation as #908, raising it here so it is on this PR too.

Approvals from @Vasanthdev2004 and @anandh8x on the current head `e7db64ff`, CI green, and `mergeStateStatus=BLOCKED` solely because CodeRabbit's changes-requested is pinned to `ff85dbcc`, a commit no longer on the branch.

The empty commit I pushed today was to give the bot the push it re-reviews on rather than the request it ignores; the tree is byte-identical to what was approved and both approvals survived it. No fresh verdict yet.

Dismissing the stale review or approving would clear it. You are added as a reviewer on all six of ours if you would rather look at the change first.

gnanam1990:
@coderabbitai review

gnanam1990:
@coderabbitai full review

gnanam1990:
@coderabbitai — all three actionable comments and both nitpicks taken, at `fbf05668`. Each verified against the current code first.

**The concurrent test could hang instead of failing.** Confirmed: `failures` was buffered at one slot per holder, each goroutine can report at *both* `hasReadRoot` checks, and nothing drains the channel until after `wg.Wait()`. The ninth report blocks forever, `wg.Done()` is never reached, and the wait hangs — so the test would deadlock in precisely the case it exists to catch. Buffered at `2*holders`.

**`t.Skip` on a failed `os.MkdirAll`.** Correct, and the sibling test directly above it already uses `t.Fatal` for the same call. The directory is the fixture, not a precondition the environment may reasonably withhold.

**The unchecked `$HOME` assumption.** Both tests place their grant under `$HOME` *because* `NewScope` seeds `/tmp` and `$TMPDIR` as write roots — but neither verified it. Extracted `homeGrantOutsideTempRoots` for both. Verified by running with `HOME` pointed at a temp directory: both now **SKIP** naming the covering root, where before they passed while proving nothing. The path is resolved before comparing, since the temp roots are resolved and `/var` is a symlink to `/private/var` on macOS — an unresolved comparison reports "outside" for a directory that is inside.

**Both `AddRead` subtests added**, and both mutation-checked by removing the matching temporary-coverage guard.

Worth flagging: my first run of the "permanent read under a temporary write" mutation **did not** fail the test, and your finding looked wrong. It wasn't — that loop pattern appears in `Add`, `AddRead` and `AddTemporaryWrite` alike, and my first-match edit had changed `Add`. Re-run against the line inside `AddRead`, it fails exactly as you said. A mutation that misses its target is indistinguishable from a test that holds, which is why I am naming it rather than just reporting the green result.

**Also added** the terminal-state assertion from your second nitpick: `hasReadRoot` answers what a caller can *see*, and an entry left at zero in `tempReads` is invisible that way while still leaking one map entry per acquire/release cycle.

`go test -race ./internal/sandbox/ -count=5`: clean.

gnanam1990:
@Vasanthdev2004 @anandh8x — same note as on #909: your approvals here were dismissed by my push of `fbf05668`, which was my call to make before asking rather than after.

The push is test-only, from CodeRabbit's full review, and the first item is the one that justified it:

**The concurrent test could hang instead of failing.** `failures` was buffered at one slot per holder, but each goroutine can report at *both* `hasReadRoot` checks, and nothing drains the channel until after `wg.Wait()`. The ninth report blocks forever, `wg.Done()` is never reached, and the wait hangs — so the test would deadlock in exactly the case it exists to catch: a refcount bug bad enough to fail more than eight of eight holders. Now buffered at `2*holders`.

**A broken fixture reported itself as a pass.** `TestANarrowerRequestHoldsTheCoveringGrant` called `t.Skip` when `os.MkdirAll` failed; the sibling test above it already uses `t.Fatal` for the same call.

**Two tests assumed something they never checked.** Both tests in `scope_extra_read_test.go` place their grant under `$HOME` *because* `NewScope` seeds `/tmp` and `$TMPDIR` as write roots — but neither verified it, so an environment pointing `HOME` at a temp directory turned them into tests that prove nothing and still report PASS. Verified by running with `HOME` set that way: both now SKIP naming the covering root.

**Two `AddRead` subtests added**, covering the read side of branches only exercised from the write side. Both mutation-checked.

One thing worth naming: my first mutation of the "permanent read under a temporary write" subtest did *not* fail, and I was ready to report the finding as unfounded. It wasn't — that loop pattern occurs in `Add`, `AddRead` and `AddTemporaryWrite`, and my first-match edit had changed `Add`. Re-run against the line inside `AddRead`, it fails as claimed.

`go test -race ./internal/sandbox/ -count=5`: clean. The diff since your approvals is `e7db64ff..fbf05668`, test files only.

gnanam1990:
@coderabbitai review

gnanam1990:
@Vasanthdev2004 @anandh8x — re-review please on `23298d59`. Same apology as #909: your approvals went with my push. The diff since is `e7db64ff..23298d59` and it is **test files only**.

The one that justified pushing: the concurrent test could **hang instead of failing**. `failures` was buffered at one slot per holder while each goroutine can report at both `hasReadRoot` checks, and nothing drains it until after `wg.Wait()` — so the ninth report blocks forever and the test deadlocks in exactly the case it exists to catch. Also fixed a `t.Skip` on a broken fixture, and two tests that assumed `$HOME` sits outside the sandbox temp write roots without ever checking.

Rebased onto current main.

Vasanthdev2004 (dismissed):
Re-approving. My approval was dismissed by the commits pushed on top, so I reviewed the delta rather than restoring it blind.

Every new test is load-bearing, checked by removing the fix each one pins and watching it fail. I could not construct a state where a permanent grant gets revoked by a temporary holder's cleanup, or where a refcount is decremented more than once for a holder, including on the error and double-undo paths.

`d45b6ca7` making the concurrency test actually concurrent was worth doing, and the new one does race the paths it claims to rather than just running them in sequence with a goroutine wrapper.

One thing worth knowing, which I do not think blocks this: a temporary READ holder keeps a temporary WRITE root alive after the write holder releases. Whether that is right depends on what you intend the refcount to count, and there is a reading where it is exactly correct, so I am flagging it rather than calling it a defect.

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

euxaristia:
### Review Summary

The reference counting and lock synchronization implementation in `internal/sandbox/scope.go` is mathematically sound, correctly prevents cross-holder access revocation, and is backed by over 700 lines of rigorous regression tests. However, the branch is behind `upstream/main` and carries uncompacted review-trigger commits.

## Findings

- [P2] Rebase onto current `upstream/main` to cleanly isolate sandbox scope changes.
  `internal/sandbox/scope.go:1`
  The branch is based on `0eab63c9` prior to the `v0.8.0` release. Without a rebase, the PR touches `package.json` and `CHANGELOG.md`.
  
  The root cause is a stale base branch. Rebase onto current `upstream/main` (`ad34dc8d`). A test rebase completes cleanly with zero conflicts and confines the diff strictly to the 4 sandbox files (`scope.go` and the 3 test suites).

- [P2] Consolidate 6 iterative commits into a clean conventional commit.
  `internal/sandbox/scope.go:1`
  The PR currently carries 6 commits, including empty trigger commits (`chore(sandbox): re-request an automated review of this head`) and intermediate test adjustments.
  
  The root cause is uncompacted fixup history. Squash or rebase the branch down into a single atomic commit (e.g. `fix(sandbox): refcount temporary roots to prevent cross-holder revocation`) matching repository guidelines.

### Key Verifications
1. **Double-Undo Idempotency**: Verified that `oncePerHolder` in `internal/sandbox/scope.go:351-358` encapsulates each holder's release in a `sync.Once` closure, ensuring duplicate undo calls cannot consume a different holder's reference count.
2. **Atomic Teardown Invariant**: Verified that `releaseTemporaryRead` and `releaseTemporaryWrite` in `internal/sandbox/scope.go:306-347` drop the map tracking entry and slice element under a single mutex lock hold, preventing concurrent callers from misinterpreting a released root as permanent.
3. **Permanent Grant Promotion**: Verified that `Add` and `AddRead` in `internal/sandbox/scope.go:88-197` delete the temporary tracking key when an exact permanent grant is registered, ensuring the root survives subsequent temporary teardowns.
4. **Test Suite**: Verified over 700 lines of targeted regression tests in `scope_temp_refcount_test.go`, `scope_temporary_test.go`, and `scope_extra_read_test.go` covering racing concurrent holders and cross read/write grants.

jatmn (changes requested):

## Merge readiness

  `.release-please-manifest.json:1`
  The PR merge base is `6edf9a8b`, while live `main` is `ad34dc8d`; comparing the branch to the live target shows it carries stale release metadata and changelog changes outside the three-dot sandbox diff. Rebase or reconstruct this change on current `main` and resolve the release files there.

## Findings

- [P1] Do not let a read grant retain write authority after its writer finishes
  `internal/sandbox/scope.go:212-230`
  `AddTemporaryRead` currently handles a read path covered by a temporary write root by incrementing that root’s `tempWrites` counter and returning an undo that calls `releaseTemporaryWrite`. If the original write holder then releases first, the root stays in `extraRoots` until the read holder finishes. Since `extraRoots` feeds `Roots()` and `validate()` uses that list for write authorization, the remaining read-only holder can write anywhere below the former write root after the explicit write grant has ended.

  The root cause is using one write-capability reference both to track lifetime and to represent a read-only dependent. Keep the concurrent-lifetime fix, but model the remaining dependency with read authority only: when the last write holder releases, any readers that still depend on the path must retain a temporary read root rather than keep the path in `extraRoots`. Add a regression that releases the writer before the reader and verifies both that `validateRead` still succeeds and that `validate` denies writes.

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

gnanam1990:
@jatmn — fixed at `41503458`, and you were right that it is an escalation rather than a lifetime bug.

`AddTemporaryRead` now consults `permanentWriteRootCoversLocked` instead of `writeRootCoversLocked`, so a merely *temporary* write cover no longer short-circuits it into borrowing write authority. The lifetime is kept as what the caller actually asked for — a temporary READ root on its own path — which is your "one reference cannot be both the lifetime and the capability" point applied directly.

Mutation-checked: restoring `writeRootCoversLocked` in that branch compiles and fails the regression on three separate assertions — the reader keeps no read root of its own, `ReadRoots()` loses the path, and the writer's cleanup revokes a live read grant.

Your regression is there as asked: writer releases first, then `validateRead` still succeeds **and** `validate` denies writes.

**Two things a verification pass found that I'd rather you heard from me.**

The comment justifying this whole design said `readRoots` "feeds `ReadRoots()` and nothing else". That was **false** — it also feeds `ExtraReadRoots()`. The security argument survives (neither confers write authority; `validate()` reads `workspaceRoot` plus `extraRoots`, which this path never enters), but that sentence *is* the written justification for the design choice, so it now lists both and says why neither grants write. A reviewer checking that claim would have found the gap and been right to distrust the rest of it.

The "the reader releases first" subtest **passes on the unfixed tree** — the escalation needs the writer to go first. It's a legitimate no-regression control for the opposite ordering, not evidence your finding is closed, and it's renamed to say so rather than padding the count.

**Still open, disclosed rather than fixed:** `AddTemporaryWrite`'s own cover branch has the same shape one level across — holder A takes `outer`, holder B takes `inner` and gets a reference on `outer`, and after A releases, B retains write over all of `outer`. That is write-depending-on-write so it is a *scope* escalation, not a read→write one, and I did not fix it here. An earlier note on this PR called it "no escalation", which was wrong; it is narrower, not absent. Say the word if you want it in this PR rather than its own.

Rebased onto `ad34dc8d`. `go test -race ./internal/sandbox/ -count=5` clean.

gnanam1990:
@coderabbitai — taken at `c715392e`. This is the write-side twin I'd disclosed on this PR as still open, so thanks for pushing it rather than letting it sit.

Reproduced first: with holder A on `outer` and holder B on `inner`, `tempWrites={outer:2}` — B borrowed a reference on A's root. After A released, `outer` stayed in `extraRoots` and B could write `outer/sibling/not-mine.txt`, a sibling it never asked for.

A narrower request now records its own root, so each holder gets exactly the authority it asked for and a lifetime of its own. Only an **exact** match shares a refcount entry; a narrower one never does, which is the distinction the escalation turned on. Three regressions: the escalation itself, the opposite release order as a control, and two holders of the same root sharing one entry. Restoring the covering-root borrow compiles and fails the first on both the sibling and the covering root.

One honest note: removing the permanent-cover early return compiles and fails **nothing**. That's correct rather than a gap — a permanently covered request would then record its own temporary root, which is redundant but harmless since the permanent root outlives it. It's a cleanliness guard, not a correctness one, and I'd rather say so than write a test that pretends otherwise.

gnanam1990:
@anandh8x @Vasanthdev2004 @jatmn — ready for re-review at `5e4cab54`. All blockers on this PR are now addressed.

**@jatmn's P1** (read grant retaining write authority after its writer finishes) — `AddTemporaryRead` no longer borrows a reference on a covering *temporary* write root; it records a temporary READ root for the path the caller actually asked for. Mutation: restoring the borrow compiles and fails on three assertions.

**CodeRabbit's write-side twin** — `AddTemporaryWrite` had the same shape one level across. Holder A takes `outer`, holder B takes `inner` and gets a reference on `outer`; after A releases, B could write all of `outer` including siblings. Reproduced (`tempWrites={outer:2}`), fixed, mutation-checked.

**CodeRabbit's latest** — the `scopeOutsideRoots` fixture skipped on setup failure while its own comment said *"A skip is not a pass"*. Every caller asserts an authorization boundary, so a skip meant the assertion never ran and the build went green. Three exits are now fatal; **one skip stays and is documented as a different kind** — when the grant path is already inside a default write root, the fixture is fine and the grant simply cannot prove anything.

Rebased onto `ad34dc8d`. `go test -race ./internal/sandbox/ -count=5`: clean.

jatmn (changes requested):

## Overall guidance before the next review

This PR has accumulated several review rounds because the individual fixes have generally been correct for the exact scenario that prompted them, but the same invariant was not always reapplied to the symmetric branch, a different root relationship, the opposite release order, or the test environment. The remaining findings are not unrelated new feature requests: they come from the same temporary-grant state machine and from final-head validation. To avoid another round where one more sibling case appears after a local patch, please audit the complete construct once before requesting review again.

The grant implementation needs to preserve all of these properties together:

| Invariant | Required behavior | Failure to guard against |
|---|---|---|
| Lifetime | A live holder keeps the access it requested until its own cleanup runs. | Another holder releases and silently revokes the live grant. |
| Least authority | After another holder releases, the survivor retains only its requested mode and subtree. | A nested reader or writer keeps a broader root, sibling access, or write authority borrowed from somebody else. |
| Permanence | Workspace and session-scoped grants outlive every temporary holder and are independent of insertion order. | A temporary cleanup removes a permanent grant, or a broad temporary entry shadows a narrower permanent one. |
| Exact versus nested roots | Exact temporary roots may share one refcount; non-exact nested requests need capability boundaries matching their own roots. | A refcount correctly preserves lifetime but preserves the wrong holder's broader authority. |
| Read/write separation | A read dependency remains read-only regardless of which holder exits first. | A reader keeps a temporary writer alive and can write after the writer finishes. |
| Per-holder cleanup | One cleanup releases one holder at most once, and map/slice state changes atomically. | Double cleanup consumes another holder's count, or an acquire sees half-released state. |
| Test isolation | Tests fail on broken setup but create everything in test-owned storage. | A valid read-only HOME fails before assertions, or tests mutate developer state. |

Please derive the `Add`, `AddRead`, `AddTemporaryRead`, and `AddTemporaryWrite` decisions from those invariants as one unit rather than patching only the currently failing branch. In particular, distinguish the lifetime being retained from the capability being retained: the refcount key determines not only when an entry disappears, but also which path and access mode remain authorized while it exists. Coverage checks should also examine all applicable permanent roots before allowing slice insertion order to choose a temporary cover.

The regression matrix should exercise both positive and negative authority outcomes across:

- read and write grants;
- exact, broader, and narrower root relationships;
- workspace, permanent read, permanent write, temporary read, and temporary write coverage;
- broad-holder-first and narrow-holder-first release orders;
- duplicate cleanup and concurrent acquisition/release;
- same-root sharing versus nested-root separation.

For every case, assert both sides of the contract: the requested path remains allowed while its holder is live, and sibling paths or stronger access modes are denied as soon as the holder that authorized them exits. A test that checks only continued access can make a lifetime fix look correct while missing authority retention; a test that checks only final cleanup can miss the live interval where escalation occurs. Mutation-check the negative assertions as well as the positive ones.

Keep that matrix hermetic by controlling the scope's default roots instead of escaping the test runtime. Then run the complete required validation on the final commit—not only the focused regression—including formatting, vet, the race-enabled sandbox tests, broader tests, build/smoke, and `make lint-static`. Finally, inspect the last diff for helpers made obsolete by the refactor. That combined pass should catch both remaining findings and prevent another review round caused by a symmetric edge or final-head check that was not run.

## Merge readiness

- [P1] Restore the required static-analysis check
  `internal/sandbox/scope.go:325`
  The current Windows required job fails because `writeRootCoversLocked` became unused when this branch moved all three coverage decisions to `permanentWriteRootCoversLocked` or exact-root logic. The base branch called this helper from `AddRead`, `AddTemporaryRead`, and `AddTemporaryWrite`; the head calls it nowhere, and `make lint-static` reproduces the CI `unused` error locally. Address the caller-migration root cause by removing the obsolete predicate, or retain it only if a real production path still needs its broader temporary-or-permanent semantics. Please do not suppress the analyzer or weaken the required check. The other Windows timeout is in unchanged `internal/config` code and appears unrelated, but this sandbox lint failure is owned by this head.

## Findings

- [P1] Retain only the subtree requested by a nested reader
  `internal/sandbox/scope.go:255`
  When reader A temporarily grants `outer` and reader B requests only `outer/inner`, the loop finds A's covering entry, increments `tempReads[outer]`, and binds B's undo to `releaseTemporaryRead(outer)`. If A releases first, the count only drops from 2 to 1, so `outer` remains in `readRoots`; `ReadRoots()` feeds `validateRead` and the native sandbox profile, allowing the sole remaining holder B to read `outer/sibling` even though B never requested it. The same ordered scan can select a broad temporary entry even when a narrower permanent read root later in the slice already covers B.

  The root cause is that one reference is representing two different things: B's need for continued lifetime and A's broader capability boundary. This PR already fixes the write-side equivalent by sharing only exact roots and recording a nested request separately. Apply the same least-authority rule to reads: permanent coverage should be determined independently of slice order; exact temporary-read requests may share a refcount; a non-exact nested request must retain only the root it requested. Add regressions that release the broad reader first and prove the nested path remains readable while a sibling is denied, cover the reverse release order, preserve exact-root sharing, and prove a permanent narrow root does not borrow an unrelated broad temporary entry.

- [P2] Keep sandbox test fixtures out of the real user home
  `internal/sandbox/scope_temporary_test.go:37`
  `scopeOutsideRoots` and `homeGrantOutsideTempRoots` create fixtures directly under `os.UserHomeDir()`. A valid hermetic runner can expose HOME read-only while providing writable test/cache/temp roots; in that environment `go test -race ./internal/sandbox` fails at `os.MkdirTemp(home, ...)` before any authorization assertion runs. On a writable machine the tests instead mutate the developer's real home, and interruption or the ignored `RemoveAll` errors can leave `zeromax-scope-*` / `zero-scope-*` debris there.

  The underlying fixture problem is that `NewScope` seeds the system temporary directory as a permanent write root, making an ordinary `t.TempDir()` grant vacuous; moving the fixture into HOME avoids that short circuit but creates a new environmental dependency. Fix that cause rather than changing the fatal setup checks back into skips: construct the test scope without seeded default roots, inject/control those defaults, or otherwise place every fixture under test-owned storage while still proving the target begins unauthorized. The existing `scopeOutsideDefaults` and `temporaryWriteFixture` helpers demonstrate the direct-scope pattern. Please verify the focused race suite with a readable but non-writable HOME and retain negative assertions showing the grant—not fixture placement—is what changes access.

jatmn (changes requested):

## Overall guidance before the next review

The current head has one remaining blocker. The number of earlier review rounds did not come from a growing list of unrelated feature requests; most were different manifestations of the same state-machine and test-design gaps. Please address the remaining item below, then do one complete final-head audit rather than another local patch so this does not turn into one more sibling case per review.

The central design problem was treating “some root covers this path” as a complete answer. It is only the first half of the decision. Every grant operation must answer two independent questions:

1. **Which capability may survive?** Read versus write, and the exact requested subtree versus a broader covering root.
2. **How long may it survive?** Permanent/session lifetime versus one or more temporary holders, including repeated cleanup and concurrent release.

Conflating those questions is what produced the earlier variants: a temporary root could shadow a permanent grant; a read holder could retain write authority; a narrow holder could retain a broader sibling-capable root; and one holder's repeated cleanup could consume another holder's reference. Those cases are addressed on the current head, but the full construct should be reviewed as one state machine across `Add`, `AddRead`, `AddTemporaryRead`, and `AddTemporaryWrite`, not as four independent functions.

Use this matrix for the final audit:

| Dimension | Cases that must be considered together | Required invariant |
|---|---|---|
| Capability | read, write, and read under write coverage | A surviving holder never gains a stronger mode than it requested. |
| Path relationship | exact, broader, narrower, disjoint | Exact temporary roots may share a count; nested requests retain only their requested subtree once the broader holder exits. |
| Lifetime | workspace, permanent read/write, temporary read/write | Permanent authority never depends on a temporary holder or slice insertion order. |
| Ordering | broad first, narrow first, permanent first, temporary first | Acquisition and release order do not change the final authority. |
| Cleanup | once, repeated, partial acquisition failure, concurrent release | One cleanup consumes at most its own reference; count and root-list mutations remain atomic. |
| Observation | `Roots`, `ReadRoots`, `validate`, `validateRead`, refcount maps, native profile consumers | The externally usable authority and internal bookkeeping agree at every live interval and after the last cleanup. |

For each applicable row, test both halves of the contract during the interval after one holder exits:

- a positive assertion that the surviving holder can still access the exact path and mode it requested;
- a negative assertion that it cannot access a sibling, covering root, or stronger write mode whose holder has exited;
- a terminal assertion that the root and refcount entry disappear after the final cleanup; and
- the reverse acquisition/release order as a control.

The test infrastructure needs the same whole-system treatment. Build scopes with controlled roots so fixtures can stay under `t.TempDir()` without being pre-authorized by production's default temp roots. Assert the target is unreadable/unwritable before granting it. Fixture construction failures should fail rather than skip; concurrency tests should wait for every acquisition, size failure channels for every possible send, and assert drained bookkeeping. Do not write fixtures into the real home directory or let a setup assumption turn an authorization test into a vacuous pass.

Finally, run the required checks on the exact final commit, after removing obsolete helpers and unrelated split residue: `make fmt-check`, `go vet ./...`, `go test ./...`, the race-enabled sandbox suite, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, `make lint-static`, `make vulncheck`, and `git diff HEAD --check`. Inspect the final diff again after those commands so a helper that lost its last caller or code copied from another #829 area does not survive only because focused tests are green.

## Findings

- [P2] Keep child-grant accessors out of this refcount split
  `internal/sandbox/scope.go:69`
  `ExtraRoots` and `ExtraReadRoots` have no production caller; they are exercised only by this PR's tests. They entered in the initial split commit together with `scope_extra_read_test.go`, but their comments and test scenario belong to #829's separate child read-grant propagation area, whose CLI/runner wiring is not part of this branch. The root cause is that the split was made at the file/commit level without fully separating the contracts: the temporary-root implementation came across with propagation-only accessors and a getter test even though the consumer side stayed behind. That leaves unrelated production surface in a PR explicitly scoped to refcounting and violates `AGENTS.md`'s focused-change rule. Remove `ExtraRoots`, `ExtraReadRoots`, and the getter-only propagation assertion from this PR. Where the refcount regressions currently use `ExtraRoots` only to inspect write-root state, use the existing `Roots()` snapshot (excluding the workspace entry) or a test-only/private helper instead. Preserve the refcount, permanence, least-authority, atomic-release, and hermetic-fixture fixes. Do not wire child propagation here; that would expand the scope rather than fix the extraction boundary.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #914, opened 2026-08-16
feat(acp): add standard session list and resume

gnanam1990:
## Summary

- advertise and implement optional ACP v1 session/list and session/resume capabilities
- replay persisted user/assistant history during session/load with stable message IDs
- keep session/resume replay-free for reconnecting clients that already retain transcript state
- list resumable sessions only, with optional cwd filtering and ACP metadata

## Verification

- go test ./internal/acp
- go vet ./internal/acp
- make fmt-check
- make lint-static
- make vulncheck
- real ZeroApp ACP conformance: initialize, session/list, and session/load passed

## Integration

ZeroApp consumes these capabilities through a separate follow-up PR. session/list and session/resume remain capability-gated for compatibility with older ACP v1 clients and servers.

## Summary by CodeRabbit

* **New Features**
  * Added support for listing and resuming saved sessions.
  * Added workspace filtering, pagination, session metadata, and resolved absolute paths to session listings.
  * Session listing and resumption capabilities are now advertised during initialization.
  * Loading a session replays saved history with stable message IDs, compaction summaries, and tool activity.
  * Live sessions now preserve user messages, tool activity, and assistant responses incrementally.

* **Bug Fixes**
  * Session loading and resuming validate workspace identity and require absolute paths.
  * Invalid, deleted, unavailable, relative, or workspace-less sessions are excluded from listings.
  * Resume fails when history cannot be restored, while loading remains best-effort.
  * Notifications of the same type remain ordered, while different types can be processed concurrently.

PR #912, opened 2026-08-16
fix(agent): stop the incompletion allowances from swallowing genuine admissions

gnanam1990:
## Split out of #829 — independent fix, and one @Vasanthdev2004 asked to see measured

Sixth piece of the split. **Not stacked on anything** — builds and tests against current `main` on its own.

## Background

The detector marks a run incomplete when the model admits it could not finish. Its allowance list exists for a real reason: a finder reporting an absence — *"I could not find where X is set in production code"* — was being marked incomplete for doing exactly its job. That cost a real audit which spent 53 tool calls proving a negative.

Vasanth's review of #829 flagged that the allowances added for that case were too broad, and asked for it to be **measured rather than argued**. Fair, so I measured.

## What the measurement showed

Eleven genuine admissions of failure, six legitimate absence-establishing findings:

```
BEFORE: 10 of 11 genuine admissions passed the detector undetected
         0 of 6  legitimate findings wrongly flagged
```

Some of the ten:

```
"I could not reproduce the crash, so the fix is unverified."
"I could not find the root cause; someone else will need to pick this up."
"I could not locate the source of the regression and have run out of ideas."
```

The cause is that the allowance keys on the **tail prefix alone**: `"could not "` followed by `"reproduce …"` is waved through however the sentence ends. But `"reproduce "` and `"find the"` head both the finding and the admission.

That is the guard's entire purpose defeated in one direction while buying nothing in the other — and it is the last thing standing between a stalled run and a report that reads like success.

## The fix

The allowance yields when the sentence **also** says the work is blocked (`unverified`, `someone else`, `ran out of`, `nothing was modified`, …).

```
AFTER:  3 of 11 still pass
        0 of 6  wrongly flagged
```

The motivating case still passes as a finding:

```
"I could NOT find where AllowManifestToolAutoApproval is set to true in production code."  → not flagged ✓
```

## Where I deliberately stopped

The remaining three are single-clause sentences carrying no blocked-work signal at all (`"I failed to reproduce it locally."`). **I did not tune the list until they passed** — that would be fitting it to my own eleven examples, which is the "argued rather than measured" failure this was meant to avoid. Catching them needs a different signal than substring matching, and that is worth its own decision.

## Verification

Mutation-checked: removing `blockedWorkMarkers` puts 7 admissions straight back through.

[...]

PR #915, opened 2026-08-17
fix(acp): three defects a desktop ACP client hits

Vasanthdev2004:
Found by driving `zero acp` over stdio from ZeroApp (the desktop client) and comparing both sides of the wire against each other. None of these fail loudly — they produce a silent denial, a spurious crash, or two buttons a user cannot tell apart.

**Rebased around #914.** This originally carried a fourth fix — `session/load` restoring history without replaying it — which gnanam's #914 also fixes, and better: the replayed messages get stable ids derived from the store's event ids, `session/load` replays while `session/resume` deliberately does not, and both are capability-gated. I dropped mine rather than keep a weaker duplicate. `translate.go` is back to its state on `main`, so the only overlap left with #914 is `agent.go`, in two hunks nowhere near `handleSessionLoad`.

The three below are untouched by #914 — it does not modify `permission.go` or `stopReasonFor`.

### An option we offered could not be accepted

`buildPermissionOptions` substitutes a fallback allow/deny list when `req.AvailableDecisions` is empty. `requestPermission` validated the client's reply against `req.AvailableDecisions` — the raw field, still empty.

So whenever ZERO did not enumerate — which is every permission event that is not a prompt (`loop.go` returns nil for those), and includes a shell call whose sandbox decision came back deny — the client was sent "Allow" and "Reject" and its answer was checked against an **empty slice**. Every option it could possibly show failed closed to `deny` with reason `"permission option was not offered"`.

The user clicks Allow, ZERO records a denial, the tool is refused and the turn carries on as though they had rejected it. Nothing surfaces it, because a denial is a legitimate answer.

One `offeredDecisions` resolver now feeds both sides, so what was sent and what is accepted agree by construction rather than by being kept in step. The narrowing this validation exists for is untouched — a client still cannot return a broader grant than it was shown.

### Cancelling a permission prompt looked like a crash

`stopReasonFor` matched only `context.Canceled`, so `errPermissionApprovalCanceled` fell through to `RPCError(codeInternalError, ...)` — a `-32603` carrying the internal sentinel text. Clients render that as a failed turn, so **declining a tool looked like ZERO falling over**. For `apply_patch`, dismissing the dialog is the only refusal a client is offered at all.

It maps to `StopCancelled` now, which is what ACP has for exactly this. The sentinel is exported as `ErrPermissionApprovalCanceled` so the surfaces can recognise it (matched with `errors.Is`, so it survives the tool-name wrapping every return site applies). A genuine failure is still an error — mapping too much to cancelled would hide real faults behind a clean ending.

### Two options labelled "Allow"

`optionKindFor` returned `"Allow"` for both `PermissionDecisionAllow` and `PermissionDecisionAllowStrict`, and `request_permissions` offers them together. The label is the only thing an ACP client shows, so the panel presented two adjacent identical buttons — and one of them silently sets `StrictAutoReview` on what was granted.

The distinction cannot live in the `optionId`, which is opaque to the client, so it is in the name: "Allow with strict review". The round trip is unchanged.

### Verification

Eight tests in `internal/acp/desktopinterop_test.go`. **Each fix was reverted by hand and the naming test confirmed to fail** — a passing test proves nothing otherwise:

| Reverted | Test result |
|---|---|
| cancel mapping removed | `stopReasonFor returned an error for a cancellation` |
| strict label back to "Allow" | `options "allow" and "allow_with_strict_auto_review" share the label "Allow"` |
| resolver reads the raw field | `no options were offered` |

`go build ./internal/... ./cmd/...`, `go vet`, and `go test` on `internal/acp` and `internal/agent` all pass. `gofmt -l` clean on every file touched.

One note on scope: `internal/agent/loop.go` changes only to export the existing sentinel — three wrap sites and one test reference renamed, no behaviour moves.

The companion PR on ZeroApp is Gitlawb/zero-app#31.

## Summary by CodeRabbit

- **Bug Fixes**
  - Permission approval cancellations now correctly end operations as cancelled instead of reporting an internal error.
  - Permission choices are validated against the options actually presented, preventing unoffered permissions from being accepted.
  - Strict approval is now clearly distinguished from standard approval.
  - Genuine permission failures continue to be reported as errors.

- **Tests**
  - Added coverage for permission option consistency, cancellation handling, sandbox retry behavior, and strict approval.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -181,12 +181,17 @@ func (a *Agent) handleSessionLoad(ctx context.Context, params json.RawMessage) (
 		}
 	}
 	sess := a.registerSession(meta.SessionID, root, history, model, models, restrictModels)
+	note := &notifier{conn: a.conn, sessionID: sess.id}
 	a.warnPersistence(
-		&notifier{conn: a.conn, sessionID: sess.id},
+		note,
 		"load session history",
 		"Could not load session history. The session is open, but earlier turns may be missing until storage recovers.",
 		historyErr,
 	)
+	// BEFORE RETURNING, so a client that renders on the reply already has the
+	// transcript. Loading history that only the model can see is what made a
+	// resumed session open blank; see notifier.replayHistory.
+	note.replayHistory(sess.snapshotHistory())
 	return LoadSessionResult{
 		ConfigOptions: a.configOptions(sess),
 		Modes:         a.modeState(sess),
@@ -296,6 +301,16 @@ func stopReasonFor(result agent.Result, err error) (string, error) {
 		if errors.Is(err, context.Canceled) {
 			return StopCancelled, nil
 		}
+		// A CLIENT THAT CANCELLED A PERMISSION PROMPT DID NOT FAIL. Only
+		// context.Canceled was matched here, so dismissing a permission dialog —
+		// a deliberate action, and for apply_patch the only refusal a client is
+		// offered — came back as JSON-RPC -32603 carrying the internal sentinel
+		// text. Editors and the desktop app both render that as a crashed turn,
+		// so declining a tool looked like ZERO falling over. It is a
+		// cancellation, and StopCancelled is what ACP has for saying so.
+		if errors.Is(err, agent.ErrPermissionApprovalCanceled) {
+			return StopCancelled, nil
+		}
 		return "", err
 	}
 	if result.FinishReason == "length" {
@@ -323,7 +338,10 @@ func (a *Agent) requestPermission(ctx context.Context, sessionID string, req age
 		}
 		return agent.PermissionDecision{Action: agent.PermissionDecisionDeny, Reason: "permission request failed: " + err.Error()}, nil
 	}
-	return decisionFromOutcome(result.Outcome, req.AvailableDecisions), nil
+	// Validated against what was actually SENT, not against the raw field. See
+	// offeredDecisions: those two differ whenever ZERO did not enumerate, and
+	// validating against the field turned every offered option into a denial.
+	return decisionFromOutcome(result.Outcome, offeredDecisions(req)), nil
 }
 
 func (a *Agent) emitPlan(registry *tools.Registry, note *notifier) {
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -321,10 +321,14 @@ func stopReasonFor(result agent.Result, err error) (string, error) {
 // session/request_permission request and maps the outcome back. Failure to reach
 // the client fails closed to deny.
 func (a *Agent) requestPermission(ctx context.Context, sessionID string, req agent.PermissionRequest) (agent.PermissionDecision, error) {
+	// Built ONCE and used for both halves. Sending one list and validating the
+	// reply against another is the defect this whole path had; see
+	// decisionFromOutcome.
+	options := buildPermissionOptions(req)
 	params := RequestPermissionParams{
 		SessionID: sessionID,
 		ToolCall:  permissionToolCall(req),
-		Options:   buildPermissionOptions(req),
+		Options:   options,
 	}
 	var result RequestPermissionResult
 	if err := a.conn.Call(ctx, MethodSessionRequestPerm, params, &result); err != nil {
@@ -333,10 +337,7 @@ func (a *Agent) requestPermission(ctx context.Context, sessionID string, req age
 		}
 		return agent.PermissionDecision{Action: agent.PermissionDecisionDeny, Reason: "permission request failed: " + err.Error()}, nil
 	}
-	// Validated against what was actually SENT, not against the raw field. See
-	// offeredDecisions: those two differ whenever ZERO did not enumerate, and
-	// validating against the field turned every offered option into a denial.
-	return decisionFromOutcome(result.Outcome, offeredDecisions(req)), nil
+	return decisionFromOutcome(result.Outcome, options), nil
 }
 
 func (a *Agent) emitPlan(registry *tools.Registry, note *notifier) {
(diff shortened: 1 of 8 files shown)


Vasanthdev2004:
Both addressed.

**Selected `optionId: "cancel"`** — real, and worse than it looks. `optionKindFor` drops cancel because ACP expresses it through the outcome, but `buildAvailableDecisions` enumerates it for shell commands *and* apply_patch, so those are the two commonest prompts. A client could return an id it was never shown and abort the turn.

I didn't take the proposed guard, because rejecting cancel by name leaves the class open — any future action `optionKindFor` declines to render would have the same hole. `decisionFromOutcome` now takes the `[]PermissionOption` that were actually sent, and `requestPermission` builds that list once and uses it for both the request and the validation. What we offer and what we accept are the same slice rather than two lists kept in step, which is the invariant this PR was already claiming and didn't quite have.

The regression test is written over every option-less decision rather than over `"cancel"`, so the next one is covered on the day it is added. Cancelling through the outcome is asserted separately — closing the selected-id route must not close the real one. Reverted the check and confirmed it fails: `bash: "cancel" was never sent as an option but was accepted as "cancel"`.

**Sandbox-retry cancellation coverage** — added for both paths. Worth being clear that these are coverage, not a fix: the wrap sites already behaved correctly, and this PR only renamed the sentinel. But the ACP mapping here depends on those paths producing something recognisable, so locking it in is right. They also assert the escalated retry does not run, since a cancelled prompt that still ran it would run the command the user had just declined. Removing the error from the cancel return fails both: `abort error = <nil>, want it to match ErrPermissionApprovalCanceled`.

`go build ./internal/... ./cmd/...`, `go vet`, `go test` on both packages pass; `gofmt -l` clean on every file touched.

Vasanthdev2004:
@gnanam1990 could you take a look when you get a chance? All checks green, CodeRabbit approved after the last round, just needs a human.

Three defects, all found by driving `zero acp` from the desktop client rather than the fake — none of them fail loudly, which is why they lasted:

- **an option we offered couldn't be accepted.** `buildPermissionOptions` falls back to allow/deny when `AvailableDecisions` is empty, but the reply was validated against the raw field, so every button we sent came back `"permission option was not offered"` and got recorded as a deny. User clicks Allow, ZERO denies, nothing says so.
- **cancelling a permission returned -32603** instead of `stopReason: cancelled`, so declining a tool rendered as a crashed turn. For apply_patch, dismissing is the only refusal a client gets.
- **request_permissions labelled two decisions "Allow"** — the panel showed two identical buttons and one of them turns on strict auto-review.

Two things worth your eye specifically:

`internal/agent/loop.go` is touched, but only to export the existing sentinel as `ErrPermissionApprovalCanceled` — three wrap sites and one test reference renamed, no behaviour moved. It's your file more than mine, so shout if you'd rather I did it another way.

CodeRabbit found a hole on the same seam going the other direction: `optionKindFor` drops `cancel`, but `buildAvailableDecisions` enumerates it for shell commands and apply_patch, so a client could select an id it was never shown and abort the turn. I didn't take the suggested guard — rejecting `cancel` by name leaves the class open — so `decisionFromOutcome` now takes the `[]PermissionOption` that were actually sent. That changes its signature and three test call sites.

Re #914: I dropped my `session/load` replay in favour of yours. `translate.go` is back to baseline and the only overlap left is `agent.go`, two hunks nowhere near `handleSessionLoad` — merges either order, and happy to rebase on you if you land first.

Every fix was reverted by hand and the naming test confirmed to fail, so none of them are green for the wrong reason.

kevincodex1 (approved):
lgtm

Vasanthdev2004:
@gnanam1990 bumping this one — still green, still just needs your approval.

I checked anandh's #916 against it since "transcript and tool feedback" sounded like it might overlap: it doesn't. 100 files and none under internal/acp, and the one file we share is loop.go where his hunk is at 3208 and mine are at 40/1364/1548/1596. Trial-merged all three pairs (915↔916, 914↔916, 915↔914) and every combination is clean, so nothing here is waiting on ordering.

No rush if you're deep in #911/#912 — mainly flagging that until this lands, three things are live for anyone driving zero acp: clicking Allow on a shell or apply_patch prompt gets recorded as a deny, dismissing a permission renders as a crashed turn, and request_permissions shows two buttons both labelled Allow.

Smallest thing to look at if you're short on time is the loop.go change — it's only exporting the existing sentinel, three wrap sites renamed, no behaviour moved. That's the part that's yours rather than mine.

Vasanthdev2004:
@gnanam1990 third time I've nudged this, but I went and looked properly before nudging again and I don't think it's you.

Every open PR on the repo is BLOCKED — all 30 of them, including your #914, #912 and #911, and the ones from outside contributors. So this isn't a queue you're sitting on, it's something structural: either a required check that never reports, or a review rule nobody currently satisfies. Worth someone with admin looking at the branch ruleset, because right now nothing can land at all.

On this one specifically, if the holdup IS the review requirement and it keys off CODEOWNERS for internal/agent, I can take my PR out of your files entirely. The only reason it touches loop.go is to export ErrPermissionApprovalCanceled so the ACP layer can recognise a cancelled permission. I could match on the error string inside internal/acp instead — uglier, and I'd rather not, but it would make this a pure internal/acp change and remove whatever CODEOWNERS hop it's waiting on. Say the word.

Otherwise it's ready: 9/9 checks green, CodeRabbit approved after the fix round, kevincodex1 approved. The three defects are the permission that records a deny when the user clicks Allow, the cancel that renders as a crashed turn, and request_permissions showing two buttons both labelled Allow.

Also FYI I opened Gitlawb/zero#929 about ACP not forwarding token usage — OnUsage already exists on agent.Options, internal/acp just never subscribes to it. Not urgent, and I'm happy to wire it if you tell me what shape you want the notification in.

PR #916, opened 2026-08-17
Polish Zero's TUI and harden patch, model, and background workflows

anandh8x:
### TUI and transcript experience

- Reworks conversation rendering into a full-width transcript with theme-aware code, diff, command, and inline-code highlighting.
- Moves plan progress into transcript cards, makes run details on demand, and removes the legacy persistent sidebar and sticky plan panel.
- Adds searchable theme previews, attachment thumbnails, transient command notices, a smoother working indicator, and clearer ask-user states.
- Improves session replay so structured edit previews and attachment context survive resume.

### Patch and tool workflows

- Adds structured `apply_patch` support for creating, updating, deleting, and moving workspace files.
- Retires the ambiguous model-facing string replacement tool while retaining compatibility for existing integrations.
- Preserves the read-before-write boundary: a partial patch no longer grants whole-file observation credit.
- Rejects ambiguous and overlapping hunks instead of silently choosing the first match.
- Preserves CRLF files and missing final newlines, rejects out-of-order end-of-file hunks, and reports partial application while clearing stale tracking after a filesystem failure.
- Keeps deferred tool loading aligned with model visibility.

### Models, background work, and platform behavior

- Improves model capability discovery and ChatGPT model-ID normalization.
- Hardens background-task output paths and migrates legacy task metadata safely.
- Improves Windows process checks and cross-platform test behavior.

## Why

These changes make the highest-frequency terminal surfaces easier to scan, keep contextual information near the work that produced it, reduce persistent UI chrome, and make file mutation behavior conservative when source context is incomplete or ambiguous.

## Validation

- `make fmt-check`
- `go vet ./...`
- All 91 Go packages tested; the TUI package was run with a short logical working directory because its title-width fixture is intentionally cwd-sensitive
- Focused structured-patch regressions under `go test -race`
- `make lint-static`
- `make vulncheck`
- Final incremental CodeRabbit CLI review: no new findings

## Review notes

The branch is rebased onto the latest `main`. Regression coverage includes unread-file overwrite protection, already-read and fully supplied tracking cases, ambiguous matching, CRLF preservation, final-newline preservation, out-of-order hunks, rollback failure, malformed plan records, and narrow ask-user layouts.

## Summary by CodeRabbit

* **New Features**
  * Added structured patches for creating, updating, deleting, and moving workspace files.
  * Added attachment thumbnails, searchable theme previews, System theme support, on-demand run details, and full-width conversations.
  * Added transcript-based plan updates and richer syntax highlighting.
* **Improvements**
  * Replaced persistent notifications with temporary inline notices.
  * Improved tool selection, model switching, activity indicators, and attachment restoration.
  * Updated editing guidance to prefer targeted patches.
* **Bug Fixes**
  * Strengthened workspace path protection and legacy task-output migration.
  * Improved model capability discovery and session restoration.


anandh8x:
@coderabbitai please re-review the latest commit (c694ed24). It addresses the remaining structured-patch containment feedback; all smoke checks are green.

anandh8x:
@coderabbitai please re-review the latest commit (3d12c6c2), which fixes the remaining structured-patch temporary-file collision.

anandh8x:
@coderabbitai please re-review the latest commit (bb04825b), which makes structured-patch temporary-file allocation collision-safe.

anandh8x:
@coderabbitai Please review the latest commit (`662daf3e`). The previously raised issues have been addressed. If everything looks good, please update your formal review status.

Vasanthdev2004 (changes requested):
Reviewed at `662daf3e`. There is good work in here, and the TUI direction is largely an improvement. But the part that is not polish has a defect I can reproduce end to end, and I would not want it to land on that basis.

## The read-before-write gate can be defeated by a one-line patch

`write_file` refuses to overwrite a file the session has not read. That refusal is what stands between a model and destroying content it never looked at. A single structured `apply_patch` now removes it:

```
write_file BEFORE any patch -> error "the intended change depends on content that has not been read exactly in this session..."
apply_patch  (one line: -const Timeout = 30 / +const Timeout = 60)  -> ok
write_file AFTER  patch     -> ok "Overwrote config.go (1 lines)."
file size: 14719 bytes before, 10 after
```

299 lines the model never read, gone. The mechanism is that `recordStructuredPatchFile` calls `RecordSeenRange(absolute, 1, lineCount, lineCount)` unconditionally, so touching one line marks the whole file seen, and `SeenWhole` then passes.

The pre-existing unified-diff path does not do this. It re-grants whole-file credit only behind `wholeBefore[absolute] || fullySupplied[absolute]`, with the comment "Partial observations remain conservative." The structured path drops that condition. Same file, same tool, two different contracts.

This is sharpened by hiding `edit_file`, because `edit_file` carried the other half of the gate: an untracked path was a hard error, and `allOccurrencesSeen` refused a partially-read region. Removing it from the model's toolbox and adding a path that grants unearned credit are individually defensible and jointly a hole.

## The replacement has the exact flaw cited for retiring the old tool

`model_visibility.go` justifies hiding `edit_file` because its "text-only replacement contract cannot safely disambiguate repeated source fragments." `edit_file` refuses in that case:

```
Error: old_string matches 2 locations in d.go. Either make old_string more specific, or pass replace_all: true...
```

The structured patch does not refuse. Given a file with two identical `return nil` bodies and a hunk with no anchor:

```
status=ok "Patch applied successfully."
func a() { return errNotFound }    <- edited
func b() { return nil }            <- untouched
```

First match wins, silently. So the stated reason for the change is a property the replacement also has, and worse: the tool being retired reports the ambiguity, the tool replacing it picks one.

## CRLF files come back mixed

```
before: "alpha\r\nbravo\r\ncharlie\r\n"
after:  "alpha\r\nBRAVO\ncharlie\r\n"
```

Every line a hunk touches loses its carriage return while the rest of the file keeps it. On a Windows checkout that is a diff full of line-ending noise on lines nobody edited.

## What I could not reproduce

The review also flagged a panic from a multi-hunk structured patch. I tried it and it applied correctly:

```
"one\nTWO\nthree\nfour\nfive\nSIX\nseven\neight\n"
```

So I am not passing that on as a finding. Mentioning it in case someone else raises it.

## Scope

109 files, +4874/-2215, under a title about polishing the transcript. Inside it I count at least a dozen separable changes, including several with no relationship to the stated purpose: a new patch dialect, retiring `theme dark` and `theme light`, deleting the sidebar, deleting the sticky plan panel, ChatGPT model-ID alias normalisation, background-task output-file migration, and Windows CI housekeeping.

Two of those are external contracts rather than internal choices. The tool list advertised to a model changes, and so do config values people have in their files today. Those deserve to be their own PRs with their own review, and the first one deserves it twice over given the above.

I am not asking for the TUI work to be redone or split for its own sake. I am saying I cannot review a model-facing contract change and a filesystem-mutation path at the depth they need while they are inside a 109-file diff labelled polish, and the two defects above are what that costs.

Happy to review the patch path on its own quickly, and the rest at normal pace.

Vasanthdev2004:
@anandh8x @gnanam1990 heads-up: this collides with #829, and not in a way a rebase settles on its own.

I trial-merged every open PR against this one. 32 are clean; 7 conflict, and #829 is the one worth talking about because it is a delete-vs-modify on the same function.

This PR REMOVES `fileRowAtMouse` from internal/tui/files_panel.go along with the sidebar click-to-select interaction — the comment change says the drill-in now opens from a transcript card instead. #829 MODIFIES that same function, at the sidebar-selectables and mouse-gate hunks, and also touches hover.go where this PR deletes.

So it is a disagreement about whether clicking a row in the context sidebar selects a file, not just two diffs landing on one file. Git will not resolve that, and whichever merges second will look like it merged cleanly after someone picks a side by hand — which is how the interaction gets silently reintroduced or silently lost.

Worth noting the size difference: #829 is +54477/-405, this is +5605/-2216. Whoever rebases second is doing it against a very large change, so it is probably cheaper to agree the intent first and have one of you carry it.

Neither of you has commented on the other's PR, so I assume this has not come up. I have no opinion on which behaviour is right — I only noticed because I was checking what #916 collides with.

The other six for completeness, all in files this PR also edits: #893, #894, #895 (all internal/cli/app.go), #854 (internal/tui/model.go), #713 (internal/tui/theme_palettes.go), #685 (internal/sandbox/risk.go and internal/tools/apply_patch.go).

anandh8x:
@coderabbitai Please review the latest commit (`b9894f0e`). The blocking findings have been addressed, and your latest review contains only non-blocking nitpicks. If no blockers remain, please approve the PR and update the formal review status.

anandh8x:
@coderabbitai approve

gnanam1990 (dismissed):
## Verdict

**Approve** — the two blocking defects I raised are demonstrably gone, and I found nothing new in the code I examined. Reviewed head `b9894f0e` against `origin/main`. This approval covers the patch engine and its safety boundaries; it deliberately does **not** cover the 82 `internal/tui` files or the scope question, both stated below.

## The blocking findings are fixed, verified rather than assumed

Re-ran the exact reproductions against `b9894f0e`:

| Finding | Before | Now |
|---|---|---|
| Panic from an overlapping end-of-file hunk | `slice bounds out of range [8:7]`, no `recover()` in the tool path | clean error, `could not find expected lines` |
| Whitespace-insensitive tier rewrote a line at a different indent | silently edited the nested `doWork()` and flattened its indentation | **refuses**: `expected lines are ambiguous; provide more surrounding context` |

The ambiguity refusal is the right shape rather than a patch over the symptom: it now behaves as `edit_file` did, which was the stated justification for retiring `edit_file` in the first place. @Vasanthdev2004's point that the replacement carried the very flaw cited for retiring the original no longer holds.

His other two also reproduce as fixed: a one-line patch no longer grants whole-file read credit (`SeenWhole` stays false, so `write_file`'s gate holds), and CRLF survives — `"alpha\r\nBRAVO\r\ncharlie\r\n"` where it previously came back mixed.

## I probed the new code, not only the old findings

28 commits landed since the head I reviewed, so re-checking the original defects would not have been a review. What I attacked:

- **Over-refusal.** The new ambiguity guard could easily have bought safety with false rejections. It does not: unique lines, context-anchored repeats, multi-line unique spans and indented unique lines all still apply.
- **Transactionality.** A patch whose second operation fails leaves the first file byte-identical on disk.
- **Path escape.** `../escaped.txt`, `sub/../../escaped.txt` and `/etc/escaped.txt` are all refused and nothing lands outside the workspace.

`internal/tools`, `internal/sandbox`, `internal/agent` and `internal/background` all pass locally; CI is 8/8.

## One correction I owe

@Vasanthdev2004 could not reproduce the panic and said so. It was real, and it reproduced on the head he tested — the shape needs a **second hunk marked `*** End of File`** whose removed lines overlap the span the first hunk consumed, which sets `searchStart = len(lines) - len(wanted)` and ignores where the previous chunk ended. A plain multi-hunk patch will not trigger it, which is why his attempt applied cleanly. Recording it because a real defect was one step from being written off, and because the fix that landed does close it.

## Not covered by this approval

- **The 82 `internal/tui` files.** I reviewed the tools, sandbox and agent clusters and the patch engine. I have not read the theme, syntax-highlighting, sidebar or plan-panel changes and am not implying coverage there.
- **Scope, which is unchanged and still the open question.** 110 files, +5514/-2216, no linked issue and no `issue-approved` label, under a title about polishing the transcript. `CONTRIBUTING.md` makes that closeable on process regardless of code quality — a maintainer call rather than a code judgment, and the only thing I would still raise.
- **Windows.** Verified on darwin only.

Vasanthdev2004 (changes requested):
Re-reviewed at `b9894f0e`. All three of my blockers are closed, and the first one is closed properly rather than papered over.

## The read-before-write gate holds now

```
1. write_file BEFORE any patch -> error  ...content that has not been read exactly in this session...
2. apply_patch (one line)       -> ok
   SeenWhole(config.go)         -> false
3. write_file AFTER patch       -> error ...has not been read exactly in this session...
4. file size 13423 before, 13423 after
```

`wasWhole` is snapshotted before the changes are applied and passed through, so the structured path now carries the same contract as the unified-diff path instead of its own. I falsified it by making `seenWhole` unconditional again and the old behaviour comes straight back:

```
   SeenWhole(config.go)         -> true
3. write_file AFTER patch       -> ok  Overwrote config.go (1 lines).
4. file size 13423 before, 4 after
```

Your test catches that, not just my probe:

```
--- FAIL: TestStructuredPatchDoesNotCreditUnreadFileForOverwrite
    file_safety_test.go:252: a partial structured patch credited the unread file as fully seen
```

You also covered the rename destination and both positive cases, which is more than I asked for. That is the right shape.

**Ambiguous hunks** refuse now and leave the file alone: `expected lines are ambiguous in d.go; provide more surrounding context`. **CRLF** survives: `alpha\r\nbravo\r\ncharlie\r\n` comes back `alpha\r\nBRAVO\r\ncharlie\r\n`, and a file with no trailing newline still has none.

Suites are clean under `-race`, CI is green.

## One thing I want fixed before this lands

### The whitespace tolerance rewrites context lines it was only supposed to match

Tab-indented Go, one line changed:

```
before: "func f() {\n\tif cond {\n\t\tdoWork()\n\t\tlogIt()\n\t}\n}\n"
after : "func f() {\n    if cond {\n        doWork()\n        logIt(ctx)\n    }\n}\n"
status: ok  Patch applied successfully.
```

The model asked to change `logIt()`. Four lines moved from tabs to spaces, three of which it never mentioned, and the tool said success.

`findStructuredPatchSequence` escalates to `TrimSpace` comparison so a hunk whose indentation is a bit off still matches, and that tolerance is deliberate. But the replacement is built from `chunk.new`, so the patch's whitespace is written over the file's on every line in the hunk, context lines included. Matching loosely and writing back loosely are two different decisions and only the first one was intended. `TestApplyPatchToolStructuredPatchMatchesWhitespaceTolerantly` pins the changed line; nothing pins the context lines.

On this repo that means a gofmt failure in CI on lines nobody edited, with no signal the model can see. Keeping the file's own lines for context positions and taking the patch's text only for lines that actually change is the fix.

I am asking for this one because it is a silent wrong modification in the path that is becoming the model's editing tool. It is not the same class as the gate defeat, and I want to be clear I am not re-litigating that.

## Three notes, no action needed from me

**A case-only rename is a silent no-op.** On Windows and macOS:

```
*** Update File: readme.md
*** Move to: README.md
-> ok  Patch applied successfully.
   on disk: readme.md  "new body\n"
```

The content edit landed, the rename did not, and the model was told it did.

**A permission prompt is unreachable while the run details overlay is open.** Opening the overlay is guarded by `noBlockingModal()`, but a prompt arriving afterwards hits `case m.runDetailsOpen:` and returns before `handlePermissionKey`. It fails closed and Esc recovers, so this is annoyance rather than danger, but it is the same guarded-one-way shape worth checking whenever a surface can open over another.

**The default theme leaves the selection highlight dark on a light terminal.** I checked before saying it, and it is much narrower than it first looks: system's palette is terminal-agnostic ANSI almost everywhere, and only `selection` differs, `rgb(43,47,45)` against `rgb(231,233,231)`. Since the background reading never re-applies for system mode, a light terminal keeps the dark one.

Also, the mixed line-ending case I raised last time still changes an untouched context line's terminator, which given the section above is probably the same root cause.

## Scope

Still 110 files, +5514/-2216, and it grew rather than split. I said last time I would not insist twice and I am not going to. The patch path has had real scrutiny now, from both of us, and that was the part I actually cared about.

gnanam1990 (approved):
## Verdict

**Approve** — re-reviewed at `c2caa9ab`. My earlier approval was auto-dismissed when the head moved, and @Vasanthdev2004 raised a blocker on `b9894f0e` that I had missed. It is fixed on this head. One residual below, non-blocking.

## The blocker I missed, and its fix

@Vasanthdev2004 found that the whitespace tolerance rewrote **context** lines, not only the changed one — a tab-indented file came back with four lines converted to spaces when the model had asked to change one. That is squarely in the area I reviewed and approved, and I did not catch it: my indentation probe checked *which line* was rewritten, never whether the surrounding context kept its own bytes.

`c2caa9ab` fixes it. Tab-indented file, space-indented hunk, end to end through the real patch text:

```
before "func f() {\n\tif cond {\n\t\tdoWork()\n\t\tlogIt()\n\t}\n}\n"
after  "func f() {\n\tif cond {\n\t\tdoWork()\n        logIt(ctx)\n\t}\n}\n"
```

The two context lines keep their tabs. Matching loosely and writing back loosely are now separate decisions, which was his point.

## Residual, non-blocking

The **changed** line still adopts the patch's spelling — `        call2()` lands in a tab-indented file:

```
after "func g() {\n\tif x {\n        call2()\n\t}\n}\n"
```

Defensible, since that line is the model's own content. But it is the same root cause narrowed from four lines to one, and on this repo it is still a gofmt failure the model gets no signal about. Worth a follow-up rather than a block: taking the file's leading whitespace for a changed line, when the match only succeeded through the trim tier, would close it.

## His other notes, re-checked here

- **Case-only rename** no longer silently succeeds. It refuses — `cannot move readme.md to README.md because the destination already exists` — and rolls back, leaving the content untouched. Previously the edit landed and the rename did not while reporting success.
- **CRLF** survives on both the changed and the untouched lines: `alpha\r\nBRAVO\r\ncharlie\r\n`.

## Regression check on my own findings

- Overlapping end-of-file hunk: returns a clean error, no panic.
- Ambiguity: still refused rather than silently resolved.

`internal/tools`, `internal/sandbox`, `internal/agent` pass; CI 8/8.

## Unchanged, and still the only thing I would raise

Scope: 110 files, +5602/-2216, no linked issue and no `issue-approved` label. It grew again this round. @Vasanthdev2004 has said he will not insist twice and I will not either — recording it because a green review should not be read as the scope question having been answered. That remains a maintainer call.

**Not covered:** the 82 `internal/tui` files, and Windows — verified on darwin only.

Vasanthdev2004 (approved):
Re-reviewed at `c2caa9ab`. The blocker is closed, and you closed the note I attached to it as well.

Context lines keep the file's own bytes now:

```
before: "func f() {\n\tif cond {\n\t\tdoWork()\n\t\tlogIt()\n\t}\n}\n"
after : "func f() {\n\tif cond {\n\t\tdoWork()\n        logIt(ctx)\n\t}\n}\n"
```

All three tabs survive. Only the line the model actually wrote carries the patch's own text, which is the right split: matching stays tolerant, writing back does not.

`newSourceOffsets` is a better answer than what I had in mind. I was thinking about the whitespace case only, and mapping each output line back to its source index covers everything a context line carries rather than just its indentation. The mixed line-ending case I raised as an aside falls out of it:

```
before "alpha\r\nbravo\ncharlie\r\n"
after  "alpha\r\nBRAVO\ncharlie\r\n"
```

`alpha` keeps its carriage return. That was the last piece of the original CRLF finding.

Both new tests are load-bearing. Making `structuredPatchReplacement` ignore the offsets kills exactly the two that should die:

```
--- FAIL: TestApplyPatchToolStructuredPatchPreservesWhitespaceTolerantContext
--- FAIL: TestApplyPatchToolStructuredPatchPreservesMixedContextLineEndings
```

And the deliberate tolerance still works rather than being thrown out with the bathwater: `"  hello   \n"` still becomes `"goodbye\n"`.

`internal/tui` is clean under `-race`, `internal/tools` too apart from a foreground-server test that fails the same way on my box regardless of branch. CI is green.

Approving.

## Still open, still not blocking

The three notes from last round are untouched, which is fine since I did not ask for them. Recording them so they do not evaporate rather than to reopen anything:

- a case-only rename (`readme.md` to `README.md`) reports success and silently does not rename on Windows and macOS
- a permission prompt is swallowed while the run details overlay is open, since opening is guarded by `noBlockingModal()` and the arrival is not
- the default theme keeps a dark selection highlight on a light terminal

One observation on the fix itself, not a request: an authored line still lands with whatever indentation the patch gave it, so a model that gets indentation wrong now misindents one line instead of the whole hunk. That is the correct trade. Guessing indentation for content the model wrote would be worse than honouring it.

anandh8x:
@coderabbit approve

PR #930, opened 2026-08-19
fix(imageinput): improve PDF fallback handling

jatmn:
## Context

This is the PDF-only work split from #927 at review request. The earlier
GO-2026-6115 advisory was withdrawn, so this PR does not claim a vulnerability
remediation.

The prior in-process fallback was removed deliberately: its `GetPlainText`
implementation fully decompresses and aggregates page text before a caller can
apply an output cap. That makes the advertised limit ineffective for hostile
PDFs. PDF text attachment therefore requires Poppler's `pdftotext` on `PATH`.

## Change

- Remove `github.com/ledongthuc/pdf` rather than retain an unenforceably bounded
  fallback.
- Bound `pdftotext` stdout, `pdfinfo` stdout, and `pdftoppm` diagnostics; retain
  truncation state at the 256 KiB text limit.
- Classify missing `pdftotext`, execution failure, and a successfully textless
  PDF separately so guidance is accurate.
- Run text extraction/page metadata concurrently under a 30-second deadline;
  optional vision rasterization runs concurrently under its own 10-second
  deadline, avoiding serial timeout multiplication.
- Document the Poppler requirement in README and `/image` help; `pdftoppm`
  remains optional for vision page images, which can still attach when PDF text
  extraction is unavailable or the document is textless.
- Keep PDF routing, file-size limits, malformed-input errors, attachment
  staging/clearing, and submit behavior covered by focused tests.

## Validation

- `go test -race ./internal/imageinput ./internal/tui -run 'Test(ExtractTextWithPoppler|LoadDocument|ImageCommand|SubmitPrependsDocumentTextThenClears)' -count=1`
- `make fmt-check`
- `go vet ./...`
- `make lint-static`
- `make vulncheck`
- `git diff --check`

`go test ./...` was also run with a review-owned canonical temporary directory;
the environment's broader listener-dependent tests did not complete reliably,
while the changed PDF/TUI test surface passed under `-race`.

## Summary by CodeRabbit

* **New Features**
  * PDF attachments can include extracted text and optional rendered page images for vision-enabled processing.
  * PDF processing now enforces page, output-size, image-dimension, and time limits for improved reliability.

* **Bug Fixes**
  * Improved handling and reporting of unavailable or failed PDF extraction, including valid textless documents.

* **Documentation**
  * Updated `/image` guidance with required and optional PDF support tools and fallback behavior.


jatmn:
Implemented the currently safe, mergeable improvements in `62d498c`:

- preserve bounded-reader overflow through the extraction route and `Document.Truncated`, including whitespace-at-the-cap boundaries;
- add direct reader-error coverage;
- retain a bounded `pdftotext` prefix instead of discarding valid oversized Poppler output and falling back;
- stop pure-Go *cross-page* aggregation once the text cap is reached.

A remaining parser-boundary decision is needed before this draft should be marked ready. `gopdf` materializes a complete single page inside `Page.GetPlainText` before our code receives it. Its API provides neither a per-page output/decompression budget nor cancellation during that call, so a compressed one-page PDF can still consume substantial memory/CPU when Poppler is unavailable. I will not describe the pure-Go route as hard-bounded while that remains true.

Options:

1. Remove the in-process text fallback and require optional Poppler for PDF text extraction. This is the smallest safe change, but changes no-Poppler behavior.
2. Isolate the pure-Go parser in a cross-platform helper process with enforced resource limits. This preserves fallback behavior but is materially more architecture.
3. Use or maintain a parser with bounded, cancellable streaming extraction. This is the best long-term contract, but requires a dependency/fork decision.

My recommendation is option 1 unless preserving no-Poppler extraction is a product requirement; if it is, option 2 or 3 is necessary.

jatmn:
@Vasanthdev2004, please advise.

Vasanthdev2004:
Thanks for splitting it, and for saying plainly in the description that the advisory was withdrawn and this is not a remediation. That is the right framing and it is what I was asking for. #927 is down to the daemon fix in two files now, which I can take on its own.

Checked the code here on its merits, separately from the dependency question.

The bounding is real: `readBoundedText` stops one byte past the cap instead of building the whole string, `pdfinfo` output has its own 64 KiB bound, and the Poppler subprocesses are under `popplerTimeout`. Page counting being independent of text extraction, with the `pdfinfo` fallback, is a genuine improvement over deriving it from whichever text path happened to win. I ran four hostile shapes through `LoadDocument` with external tools disabled, and all four come back as clean errors with nothing escaping the package:

```
truncated header       errored=true textlen=0 pages=0
garbage after header   errored=true textlen=0 pages=0
deep nesting           errored=true textlen=0 pages=0
huge xref claim        errored=true textlen=0 pages=0
```

Both recover sites are still in place. `internal/imageinput` is green.

## On the dependency, which is the actual question

One thing worth putting on the record before anyone decides, because it changes what the swap means. `Detective-XH/gopdf` is not a different parser. It is the same lineage:

```
contributors: Detective-XH(456), ledongthuc(38), rsc(9), ...
LICENSE:      Copyright (c) 2009 The Go Authors. All rights reserved.
```

So it carries `rsc.io/pdf` through `ledongthuc/pdf`, which is exactly the family GO-2026-6115 named before it was withdrawn ("rsc.io/pdf and forks"). Had the advisory stood, this would not have escaped it. Worth knowing, since the original motivation was to get away from that lineage.

I also want to walk back part of what I said on #927. I described it as a thin trust anchor off the star count, and that was unfair to the repo. Looking properly, it carries fuzz tests, a redteam test file, and a large table and layout corpus, with 456 commits from the maintainer. That is a serious body of work, not a drive-by fork.

What is left is narrower and still real: one maintainer, no other eyes on it, last push six weeks ago, and it is the component that parses untrusted files a user hands the agent. It also pulls `golang.org/x/text` into the graph, which is fine in itself but does end the "no transitive deps" property the replaced comment claimed.

None of that is a no from me. It is a dependency decision for @kevincodex1 rather than something I should wave through in a review, which is why I am leaving this as a comment on the draft rather than a verdict. If he is happy with the parser, the code around it is in good shape.

Vasanthdev2004 (changes requested):
Taking the decision rather than sending it upstairs, since it is ours to make: **keep the hardening, drop the dependency swap.**

I went looking for a reason to say yes and could not find one, so here is the working rather than just the answer.

## The swap buys no security

`Detective-XH/gopdf` is not a different parser. It is the same lineage:

```
contributors: Detective-XH(456), ledongthuc(38), rsc(9), ...
LICENSE:      Copyright (c) 2009 The Go Authors. All rights reserved.
```

That is `rsc.io/pdf` through `ledongthuc/pdf`, which is exactly the family GO-2026-6115 named before it was withdrawn ("rsc.io/pdf **and forks**"). Had the advisory stood, moving here would not have escaped it. And it has not stood since 18 Aug, with `govulncheck` clean on main against the current library.

## The hardening does not need it

This is the part that decided it. I took your branch, pointed the import back at `ledongthuc/pdf`, and ran it:

```
--- PASS: TestPDFOutputReadersAreBounded
--- PASS: TestPDFPageCountIndependentOfTextExtraction
--- PASS: TestResolvePageCountFallsBackToPopplerWhenInProcessIsZero
--- PASS: TestLoadDocumentHostilePDFStaysBounded
--- PASS: TestLoadDocumentVisionUsesText
ok  github.com/Gitlawb/zero/internal/imageinput
```

All five of your new tests, the whole package green, and **zero dependency change against main**. The only adaptation needed is the extraction loop, because the old API hands back an `io.Reader` instead of a page slice:

```go
plain, perr := reader.GetPlainText()
if perr != nil { ... }
text, overflow, rerr := readBoundedText(plain)
```

Which is your own `readBoundedText`, doing what it was written for. The result is fourteen lines shorter than the page loop and drops the `context` import from that path, since a `LimitReader` bounds it without needing a timeout to escape.

## So what is left is a cost with nothing on the other side

One maintainer, no external review, last push six weeks ago, on the component that parses untrusted files a user hands the agent. Plus `golang.org/x/text` in the graph, which ends the no-transitive-deps property.

I want to be fair to the repo, and I was not on #927: it carries fuzz tests, a redteam file and a large table and layout corpus over 456 commits. That is real work and I withdraw the "thin trust anchor" line. It is simply not work we need, for a fallback path that only runs when Poppler is absent.

## What would change my mind

A demonstrated extraction difference on real PDFs. If gopdf reads documents ours mangles, that is a genuine reason and I would take it on those terms. Nobody has shown one, and this is the fallback rather than the preferred path, so the bar is high.

## The hardening itself is good and I want it

Bounding at the reader rather than after the fact, page counting independent of text extraction with the `pdfinfo` fallback, and the malformed-input coverage are all real improvements. I ran four hostile shapes through `LoadDocument` with external tools disabled and all four came back as clean errors with nothing escaping the package.

Re-point the import, swap that loop, and I will approve it.

Vasanthdev2004 (approved):
Approving. Sorry this sat on a stale change request for so long; my review was against `62d498c3` and you had answered it thirty-two commits ago.

My ask was "keep the hardening, drop the dependency swap". You went further and dropped the dependency altogether: `ledongthuc/pdf` is gone from `go.mod` and nothing replaces it, so the net change against main is one fewer parser rather than a sideways move within the same `rsc.io/pdf` lineage. That is a better answer than the one I asked for, and it makes the withdrawn-advisory question moot rather than merely survivable.

The trade it buys is real and I want to be explicit that I looked at it: PDF text now needs Poppler's `pdftotext` on PATH, where it used to work in-process. That degrades properly rather than mysteriously. With both binaries stubbed absent:

```
LoadDocument: ... has no extractable text; install Poppler's pdftotext for PDF
text extraction (and pdftoppm for image-only PDFs)
```

It names both tools and what each is for, `popplerAvailable` gates before exec rather than after a failure, and the README documents the prerequisite in both the `/image` table and the attachment section. A user who hits this can act on it without reading the source, which is the bar.

The bounds around shelling out hold up: `maxRasterDimension` caps both dimensions passed to `pdftoppm`, `MaxPages` is hard-capped independently of the caller, and the text cap truncates with a marker rather than rejecting, so a large valid spec stays partly usable.

One thing I went looking for and found already answered, in case anyone else pulls the same thread: removing the `io.LimitReader` on the document read leaves the whole package green, because `info.Size()` already rejects an oversize file. The reader bound is there for the stat/read race on a growing file, and the comment above it says exactly that. Defence in depth against a race no test can easily construct, not an unpinned bound.

Verified on Windows: `internal/imageinput` and `internal/tui` both pass, cross-compiles clean for linux, darwin and windows. Worth knowing my box has `pdftotext` but not `pdftoppm`, so the raster path here only exercised the unavailable branch; the CI legs cover the rest.

Issue #552, opened 2026-07-06
Unconfigured optional MCP server (firecrawl) prints HTTP 401 warning on every startup


ddfp777:
### Version / branch / commit

zero 0.1.0 (npm install, ~July 3 2026)

### OS and environment

Windows 11 / cmd.exe / fresh install via npm / no firecrawl credentials configured

### Steps to reproduce

1. Fresh install of Zero via npm on Windows
2. Do not configure any firecrawl credentials (never used the service)
3. Run `zero` in any directory

### Expected behavior

Optional MCP servers with no credentials configured are skipped silently (or logged at debug level only). Loud warnings are reserved for servers the user actually configured that then fail.

### Actual behavior

Every startup prints:

warning: MCP server firecrawl unavailable, skipped: initialize MCP server firecrawl: MCP http server firecrawl returned HTTP 401: {"error":"invalid_token","error_description":"Firecrawl credentials required: OAuth access token ..."}

First-run experience: a new user's very first Zero output is a credentials warning for a service they've never heard of. It reads like something is broken.

(Distinct from #512 — that's about project config re-enabling user-disabled servers; this is about the default noisy failure when an optional server was never configured at all.)

### Relevant logs, screenshots, or error messages

Full warning line above, captured verbatim from startup. Reproduces on every launch.

PierrunoYT:
Trying to fix it!

ddfp777:
Thanks for jumping on this @PierrunoYT 🙏 Happy to verify the fix on my Windows setup once #563 lands, it's the exact environment from the report.

ddfp777:
#563 merged, nice. I'm on 0.2.0 via npm which predates the merge; will verify on my Windows setup (the exact environment from the report) as soon as the next release ships.

PR #932, opened 2026-08-20
perf: reduce startup and turn overhead

anandh8x:
## Summary

- Reduce unnecessary TUI message churn by removing duplicate cursor blinking and stopping finite composer timers after idle.
- Move optional tool discovery off the startup path and publish immutable tool-registry snapshots for cheaper concurrent reads.
- Reuse compatible response sessions, send patch input in its native form, and recognize persisted ChatGPT session metadata when reconnecting.
- Expose cache read/write efficiency and reduce bounded-task turn churn through direct reads, native patches, and consolidated verification.

## Why

Zero was doing avoidable work before the first frame and between compatible provider turns. This change keeps the same user-facing capabilities while reducing startup latency and repeated turn overhead.

## Turn-efficiency comparison

A controlled coding workload used the same provider, model, reasoning level, prompt, and fresh fixture before and after the turn-efficiency changes. The after column is the faster of two post-change runs; both passed the same hidden correctness oracle:

| Metric | Zero before | Zero after |
| --- | ---: | ---: |
| Completion time | 3m 22.6s | 1m 42.5s |
| Model requests | 14 | 4 |
| Tool actions | 22 | 6 |
| Total input | 202.1K | 41.5K |
| Uncached input | 48.5K | 25.2K |
| Output tokens | 8.6K | 4.6K |
| Reasoning tokens | 3.8K | 1.9K |
| Local CPU | 3.57s | 1.59s |
| Peak RSS | 147.1 MB | 117.4 MB |
| Correctness | Passed | Passed |

## Dependency scope

Zero uses the official `charm.land/bubbletea/v2 v2.0.9` release. Demand-driven renderer changes are intentionally excluded and can follow separately after upstream support is released.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test -race ./internal/tui ./internal/providers/openai ./internal/tools ./internal/mcp ./internal/cli ./internal/agent -count=1`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`

All change-related checks passed against official Bubble Tea. A local full-suite run still reproduces the pre-existing `TestAltScreenTranscriptScrollKeepsFooterFixed` failure on the untouched comparison checkout.

## Summary by CodeRabbit

* **New Features**
  * Added optimized ChatGPT sessions with streaming, response chaining, and automatic fallback.
  * Added freeform tool support, including safer `apply_patch` operations within permitted locations.
  * Optional MCP services now start in the background, with tools appearing when ready.
  * Added cached-input and cache-write token usage to command output and performance metrics.

* **Bug Fixes**
  * Preserved tool-call identifiers across responses and conversation history.
  * Improved tool registration consistency, error redaction, shutdown handling, cursor blinking, and notification timing.
  * Improved planning guidance for complex tasks and direct file editing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,9 +371,13 @@ go run ./cmd/zero-perf-bench
 
 Experimental: `ZERO_OPENAI_TURN_SESSION=1` enables the optimized OpenAI turn
 session (background connection prewarm + request-prefix telemetry) for headless
-`zero exec` runs against official OpenAI profiles. Off by default; `0`/`false`
-disable. A/B-benchmark it by running the same `zero-perf-bench` suite with the
-variable unset and set.
+`zero exec` runs against official OpenAI profiles. Off by default; `0`, `false`,
+or `off` disable it. A/B-benchmark it by running the same `zero-perf-bench` suite
+with the variable unset and set.
+
+Native ChatGPT Responses sessions are enabled by default. Set
+`ZERO_CHATGPT_TURN_SESSION=0`, `false`, or `off` to restore stateless HTTP/SSE
+transport.
 
 ### Code Quality and Security Checks
 
--- a/go.mod
+++ b/go.mod
@@ -2,9 +2,11 @@ module github.com/Gitlawb/zero
 
 go 1.26.6
 
+replace charm.land/bubbletea/v2 => github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558
+
 require (
 	charm.land/bubbles/v2 v2.1.1
-	charm.land/bubbletea/v2 v2.0.8
+	charm.land/bubbletea/v2 v2.0.9
 	charm.land/lipgloss/v2 v2.0.5
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/alecthomas/chroma/v2 v2.27.0
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/Gitlawb/zero
 
 go 1.26.6
 
-replace charm.land/bubbletea/v2 => github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558
-
 require (
 	charm.land/bubbles/v2 v2.1.1
 	charm.land/bubbletea/v2 v2.0.9
--- a/go.sum
+++ b/go.sum
@@ -1,7 +1,5 @@
 charm.land/bubbles/v2 v2.1.1 h1:7r55WzBxpo/R3z98hGmY7KKPd3ET6vsf0Fb9sDHOV60=
 charm.land/bubbles/v2 v2.1.1/go.mod h1:GE6M31gaWZVXzGw73OeuTTgy4lX+OtkH0E5ymnNsHxo=
-charm.land/bubbletea/v2 v2.0.8 h1:SxTJMhCAI3lbPmy4SgX5LWZ24AdINr4I6UEqzZvYJuY=
-charm.land/bubbletea/v2 v2.0.8/go.mod h1:2SkdgoTXluXJHOUwAoRlRXF/28vklb1rFl6GcgV1/ss=
 charm.land/lipgloss/v2 v2.0.5 h1:kbNxgeeUOYv5J0YdpxFjfvf3dFvqH8Aci4zB6xqFtrY=
 charm.land/lipgloss/v2 v2.0.5/go.mod h1:9oqhxt4yxIMe6q5A4kHr44DremZk7J9UNh74GlWa5nc=
 github.com/Microsoft/go-winio v0.6.2 h1:F2VQgta7ecxGYO8k3ZZz3RS8fVIXVxONVUPlNERoyfY=
@@ -12,6 +10,8 @@ github.com/alecthomas/chroma/v2 v2.27.0 h1:FodwmyOBgJULFYmDqibcp9pvfDLWdtPRh9v/r
 github.com/alecthomas/chroma/v2 v2.27.0/go.mod h1:NjJ3ciIgrqBNeIkWZ4e46nseoLDslxU1LmfCoL+wcY8=
 github.com/alecthomas/repr v0.5.2 h1:SU73FTI9D1P5UNtvseffFSGmdNci/O6RsqzeXJtP0Qs=
 github.com/alecthomas/repr v0.5.2/go.mod h1:Fr0507jx4eOXV7AlPV6AVZLYrLIuIeSOWtW57eE/O/4=
+github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558 h1:LaEENulSSjn7i60wfHAWY8vhvTBfBCdVeGRYHvKn3pQ=
+github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558/go.mod h1:2SkdgoTXluXJHOUwAoRlRXF/28vklb1rFl6GcgV1/ss=
 github.com/atotto/clipboard v0.1.4 h1:EH0zSVneZPSuFR11BlR9YppQTVDbh5+16AmcJi4g1z4=
 github.com/atotto/clipboard v0.1.4/go.mod h1:ZY9tmq7sm5xIbd9bOK4onWV4S6X0u6GY7Vn0Yu86PYI=
 github.com/aymanbagabas/go-udiff v0.4.1 h1:OEIrQ8maEeDBXQDoGCbbTTXYJMYRCRO1fnodZ12Gv5o=
@@ -64,8 +64,6 @@ github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e h1:JVG44RsyaB9T2KIHavM
 github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e/go.mod h1:RbqR21r5mrJuqunuUZ/Dhy/avygyECGrLceyNeo4LiM=
 golang.org/x/exp v0.0.0-20260611194520-c48552f49976 h1:X8Hz2ImujgbmetVuW+w2YkyZChE3cBpZi2P158rTG9M=
 golang.org/x/exp v0.0.0-20260611194520-c48552f49976/go.mod h1:vnf4pv9iKZXY58sQE1L86zmNWJ4159e1RkcWiLCkeEY=
-golang.org/x/image v0.44.0 h1:+tDekMZED9+LrtB3G5xzRggpVh9CARjZqROla3R3R+I=
-golang.org/x/image v0.44.0/go.mod h1:V8K3KE9KKKE+pLpQDOeN18w9oacNSvy1tDOirTu4xtY=
 golang.org/x/image v0.45.0 h1:FMb1nTbH5H9vF55SriQHgFw5GnNL9Jg6L25BwXKzhB0=
 golang.org/x/image v0.45.0/go.mod h1:n62x/7RqlwXDvGsSU4u6IUTUf6KghUZ9Bt7cG/T9Fx4=
 golang.org/x/sync v0.22.0 h1:SZjpbeLmrCk4xhRSZFNZW5gFUeCeFgjekvI/+gfScek=
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 charm.land/bubbles/v2 v2.1.1 h1:7r55WzBxpo/R3z98hGmY7KKPd3ET6vsf0Fb9sDHOV60=
 charm.land/bubbles/v2 v2.1.1/go.mod h1:GE6M31gaWZVXzGw73OeuTTgy4lX+OtkH0E5ymnNsHxo=
+charm.land/bubbletea/v2 v2.0.9 h1:DpJCMWKgzQK8SJv4zbKKFHAI10ymWy/evClPFk0k0f8=
+charm.land/bubbletea/v2 v2.0.9/go.mod h1:2SkdgoTXluXJHOUwAoRlRXF/28vklb1rFl6GcgV1/ss=
 charm.land/lipgloss/v2 v2.0.5 h1:kbNxgeeUOYv5J0YdpxFjfvf3dFvqH8Aci4zB6xqFtrY=
 charm.land/lipgloss/v2 v2.0.5/go.mod h1:9oqhxt4yxIMe6q5A4kHr44DremZk7J9UNh74GlWa5nc=
 github.com/Microsoft/go-winio v0.6.2 h1:F2VQgta7ecxGYO8k3ZZz3RS8fVIXVxONVUPlNERoyfY=
@@ -10,8 +12,6 @@ github.com/alecthomas/chroma/v2 v2.27.0 h1:FodwmyOBgJULFYmDqibcp9pvfDLWdtPRh9v/r
 github.com/alecthomas/chroma/v2 v2.27.0/go.mod h1:NjJ3ciIgrqBNeIkWZ4e46nseoLDslxU1LmfCoL+wcY8=
 github.com/alecthomas/repr v0.5.2 h1:SU73FTI9D1P5UNtvseffFSGmdNci/O6RsqzeXJtP0Qs=
 github.com/alecthomas/repr v0.5.2/go.mod h1:Fr0507jx4eOXV7AlPV6AVZLYrLIuIeSOWtW57eE/O/4=
-github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558 h1:LaEENulSSjn7i60wfHAWY8vhvTBfBCdVeGRYHvKn3pQ=
-github.com/anandh8x/bubbletea/v2 v2.0.10-0.20260820073754-bdad7c1db558/go.mod h1:2SkdgoTXluXJHOUwAoRlRXF/28vklb1rFl6GcgV1/ss=
 github.com/atotto/clipboard v0.1.4 h1:EH0zSVneZPSuFR11BlR9YppQTVDbh5+16AmcJi4g1z4=
 github.com/atotto/clipboard v0.1.4/go.mod h1:ZY9tmq7sm5xIbd9bOK4onWV4S6X0u6GY7Vn0Yu86PYI=
 github.com/aymanbagabas/go-udiff v0.4.1 h1:OEIrQ8maEeDBXQDoGCbbTTXYJMYRCRO1fnodZ12Gv5o=
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -144,6 +144,7 @@ func estimateToolDefTokens(tools []zeroruntime.ToolDefinition) int {
 	for _, tool := range tools {
 		total += ApproxTextTokens(tool.Name)
 		total += ApproxTextTokens(tool.Description)
+		total += ApproxTextTokens(string(tool.Type))
 		if len(tool.Parameters) > 0 {
 			if encoded, err := json.Marshal(tool.Parameters); err == nil {
 				total += ApproxTextTokens(string(encoded))
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -149,6 +149,11 @@ func estimateToolDefTokens(tools []zeroruntime.ToolDefinition) int {
 				total += ApproxTextTokens(string(encoded))
 			}
 		}
+		if tool.Format != nil {
+			total += ApproxTextTokens(tool.Format.Type)
+			total += ApproxTextTokens(tool.Format.Syntax)
+			total += ApproxTextTokens(tool.Format.Definition)
+		}
 		total += 4 // per-tool overhead
 	}
 	return total
--- a/internal/agent/compaction_projection.go
+++ b/internal/agent/compaction_projection.go
@@ -126,8 +126,11 @@ func capCompactionBrief(brief string) (string, bool) {
 
 func compactionToolCallLine(call zeroruntime.ToolCall) string {
 	detail := ""
+	if call.Freeform && call.Name == "apply_patch" {
+		detail = strings.Join(patchPaths(call.Arguments), ", ")
+	}
 	var arguments map[string]any
-	if json.Unmarshal([]byte(strings.TrimSpace(call.Arguments)), &arguments) == nil {
+	if detail == "" && json.Unmarshal([]byte(strings.TrimSpace(call.Arguments)), &arguments) == nil {
 		for _, key := range []string{"file_path", "path", "url", "query", "pattern", "prompt", "description", "question"} {
 			if value, ok := arguments[key].(string); ok && strings.TrimSpace(value) != "" {
 				detail = value
--- a/internal/agent/compaction_projection_test.go
+++ b/internal/agent/compaction_projection_test.go
@@ -73,6 +73,7 @@ func TestProjectCompactionInputRetainsAskUserExchangeAndSemanticArguments(t *tes
 		{Role: zeroruntime.MessageRoleTool, ToolCallID: "ask", Content: "Which database should remain?: Postgres only; do not touch MySQL."},
 		{Role: zeroruntime.MessageRoleAssistant, ToolCalls: []zeroruntime.ToolCall{
 			{ID: "patch", Name: "apply_patch", Arguments: `{"patch":"*** Begin Patch\n*** Update File: internal/db.go\n*** End Patch"}`},
+			{ID: "raw-patch", Name: "apply_patch", Freeform: true, Arguments: "*** Begin Patch\n*** Update File: internal/api.go\n*** Add File: internal/config.go\n*** End Patch\n"},
 			{ID: "fetch", Name: "web_fetch", Arguments: `{"url":"https://example.com/spec"}`},
 			{ID: "task", Name: "task", Arguments: `{"prompt":"Inspect the Windows implementation"}`},
 		}},
@@ -80,7 +81,7 @@ func TestProjectCompactionInputRetainsAskUserExchangeAndSemanticArguments(t *tes
 	}
 
 	content := projectCompactionInput(messages).messages[0].Content
-	for _, want := range []string{"Which database should remain?", "Postgres only; do not touch MySQL", "internal/db.go", "https://example.com/spec", "Inspect the Windows implementation"} {
+	for _, want := range []string{"Which database should remain?", "Postgres only; do not touch MySQL", "internal/db.go", "internal/api.go", "internal/config.go", "https://example.com/spec", "Inspect the Windows implementation"} {
 		if !strings.Contains(content, want) {
 			t.Fatalf("projection missing %q: %q", want, content)
 		}
--- a/internal/agent/compaction_test.go
+++ b/internal/agent/compaction_test.go
@@ -275,6 +275,11 @@ func TestEstimateToolDefTokensCountsDefinitions(t *testing.T) {
 	if estimateToolDefTokens(one) <= 0 {
 		t.Fatal("a tool definition (name + description + schema) should estimate > 0 tokens")
 	}
+	typed := append([]zeroruntime.ToolDefinition(nil), one...)
+	typed[0].Type = zeroruntime.ToolDefinitionFreeform
+	if estimateToolDefTokens(typed) <= estimateToolDefTokens(one) {
+		t.Fatal("a non-empty tool type must increase the estimated request cost")
+	}
 	two := append(append([]zeroruntime.ToolDefinition{}, one...), zeroruntime.ToolDefinition{
 		Name:        "write_file",
 		Description: "Write contents to a file in the workspace.",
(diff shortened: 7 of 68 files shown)


Vasanthdev2004:
Had a look through the draft. Not reviewing it as ready, since you have said it is not, but there are three things worth having now rather than after a rebase.

## Your own new test fails at head

`TestSplitMCPStartupConfigDefersOnlyUnconfiguredDefaults`:

```
critical servers = map[...]{"docs":..., "firecrawl":...}, want only docs
```

`firecrawl` is landing on the prompt-critical path. The cause is that `IsUnconfiguredDefault` requires the name to be in `DefaultMCPServers()`, and firecrawl is not there any more, it is in `retiredDefaultMCPServers` with `successor: exa`. Main retired it in #926, so this looks like it survived your rebase onto `1ec7219a` rather than being wrong when you wrote it.

The interesting part is what it is pointing at rather than the fixture. A RETIRED default is classified critical, so a user carrying a leftover unmodified `firecrawl` entry gets a probably-dead server on the startup path, which is the exact thing this PR is trying to get off it. Swapping the fixture to a live default makes the test pass; deciding whether retired defaults should defer too is the actual question.

## It collides with #835, and neither branch can see it

`optionalMCPStartup.Skipped()` returns nil unconditionally. Today that costs nothing, because the startup warning at `app.go:863` already skips `UnconfiguredDefault` on purpose (#552).

But #835 adds the `/mcp` failure panel, it feeds from `mcpRuntime.Skipped()`, and it does NOT filter `UnconfiguredDefault`. So once both land, a built-in default that fails to start renders as **enabled** in `/mcp` instead of **failed**, which is precisely the state #835 exists to prevent.

Not saying either side is wrong. The startup warning and the panel are different surfaces and it is defensible for them to disagree. But whichever lands second needs to reconcile it deliberately, and right now neither branch contains the evidence that there is anything to reconcile.

## What I liked, since it is the part that would worry me otherwise

The boundary is right: only unchanged built-in defaults defer, anything the user explicitly configured stays on the prompt-critical path. And `Await` giving optional tools one bounded chance to join the NEXT turn's immutable snapshot means the tool set cannot change underneath a turn in flight. That was the failure mode I went looking for and it is closed by construction rather than by timing.

## Two things I would want before this is not a draft

The Bubble Tea fork, which you have already called out. Agreed it blocks, and I would rather see it stated in the PR title too so nobody has to read this far to learn it.

And the second table. Model requests 16 to 11, tool calls 21 to 13, tokens 204.8K to 145.2K, on the same prompt and fixture. That is not a latency change, that is the agent doing different work, and "Correctness: Passed" on one workload is thin evidence for it. Session reuse legitimately removing redundant round trips is a real win and would be the best result in this PR. It is also what a subtly dropped tool call looks like. Worth saying which, with something that would tell them apart.

euxaristia:
## Independent reproduction on a different machine

Reproduced the launch and idle claims against `main` at `1ec7219`, benchmarking this branch at `95d4f51`, with `grok` 1.0.5 included as an external reference point. No model turn is involved anywhere in this: no prompt is ever submitted, so these figures measure harness startup and idle behavior only.

### Method

- Release builds via `go run ./cmd/zero-release build` for both `main` and this branch.
- Linux x86_64 (WSL2), 20 cores, otherwise idle (load average 0.02).
- Real PTY at 80x24, one fresh process per sample.
- Terminal capability queries (OSC 10/11, DA1, DECRQM `$p`, cell-size `16t`) are auto-answered by the measuring harness. Without this, both binaries block waiting on replies a bare PTY never sends, which inflates startup and produced a false "settled" reading on my first attempt.
- Launch = spawn until output goes quiet for 200ms, used as a proxy for time-to-first-usable-frame.
- Idle = 30s settle, then a 10s sample of `utime+stime` from `/proc/<pid>/stat` and `VmRSS` from `/proc/<pid>/status`.
- All runs strictly sequential. Running two builds concurrently contaminates both CPU and RSS; those runs were discarded.
- Zero runs use an isolated XDG config with xAI / `grok-4` as the active provider.

### Launch, 10 samples per build

| Build | Median | Range |
| --- | ---: | ---: |
| `main` (`1ec7219`) | 316 ms | 299-396 ms |
| this branch (`95d4f51`) | **52 ms** | 51-54 ms |
| grok 1.0.5 | 251 ms | 243-403 ms |

Roughly a 6x improvement over `main`, and the tightest distribution of the three: a 3 ms spread across 10 runs, against ~100 ms for `main`.

### Idle CPU, 30s settle + 10s sample

| Build | CPU |
| --- | ---: |
| `main` | 0.50-0.79% |
| this branch | **0.00% in every sample** |
| grok 1.0.5 | 0.20-0.30% |

This matches the mechanism. Upstream `startRenderer()` drives a 60Hz `time.Ticker` that calls `flush()` unconditionally on every tick; the pinned fork replaces it with a `renderWake` channel that blocks until a render is actually requested. The branch's own TUI changes (removing the duplicate cursor blink, stopping finite composer timers) are what stop Zero from requesting renders it does not need. Both halves are required to reach 0.00%.

### Idle RSS, 4 paired alternating samples

| Sample | `main` | this branch |
| --- | ---: | ---: |
| 1 | 40.7 MiB | 38.3 MiB |
| 2 | 39.4 MiB | 35.8 MiB |
| 3 | 39.9 MiB | 36.8 MiB |
| 4 | 40.1 MiB | 35.7 MiB |
| **median** | **40.0 MiB** | **36.3 MiB** |

Lower in all four pairs, by about 3.7 MiB.

Worth flagging that RSS needed paired sampling to be trustworthy. Looser single-sample runs produced outliers as high as 48 MiB that I could not reproduce afterwards under their own conditions, including a cold `modelsdev.json` cache, which I initially and incorrectly assumed was the cause. Any single RSS reading on this workload should be treated as noise.

For context, grok 1.0.5 idles at roughly 120 MiB, but it is a different product with a much larger resident surface, so that gap is context rather than a defect.

### Note on the pinned dependency

charmbracelet/bubbletea#1776 is still open and unmerged, so the `replace` currently points at an untagged commit on a personal fork. The description already flags this as a merge blocker; I note only that the idle-CPU result above depends on that fork specifically, so it will want re-measuring against whatever upstream eventually lands.

euxaristia:
@coderabbitai full review

anandh8x:
Removed the temporary Bubble Tea fork from this PR so it can be reviewed and merged independently. Zero is back on the official v2.0.9 release. Once charmbracelet/bubbletea#1776 is merged and released, I’ll open a separate PR for the demand-driven renderer integration.

euxaristia:
@coderabbitai full review

anandh8x:
@coderabbitai please review the latest commit. If no issues remain, please approve.

anandh8x:
@coderabbitai approve

Vasanthdev2004 (changes requested):
Reviewed at `a46492f7`. There is real engineering here and the WebSocket Responses session is careful work, including the `previous_response_not_found` fallback. Two things block it for me, and one framing point I think matters more than either.

**The framing first.** The ChatGPT Responses turn-session path does not exist on `main` at all, and this PR both introduces it and defaults it on. That is what produces most of the headline table: a stateful session carrying `previous_response_id` does not resend prior turn content, so 202.1K input becomes 41.5K and 14 requests become 4. Those are good numbers, but they measure a new default-on transport, not "reduced startup and turn overhead", and a reader of the title would not know a data-handling change was in scope. Conversation state now lives with the provider between turns by default. That deserves to be argued on its own terms rather than arriving inside a perf PR, and I would want it opt-in for at least one release.

On the measurement: the after column is "the faster of two post-change runs" and the before column is unqualified, one workload, against an oracle that is not in the repo. I am not doubting the direction, but as presented a reviewer cannot check it. The repo has `internal/measurements` and `cmd/zero-perf-bench`; a committed benchmark would make this reproducible.

**[P1] tool_search advertises tools the run cannot execute.**

`startOptionalMCP` is new here and registers built-in MCP defaults on a background goroutine after first paint. A run snapshots the registry once per prompt with `cloneToolRegistry` after a 1s grace. But `toolSearchTool` holds a `*Registry` and `Clone()` copies tool VALUES, so the cloned registry's `tool_search` still points at the live one. Anything registering after the snapshot is visible to search and absent from execution.

Reproduced against this head, driving both calls through the run's own registry:

```
tool_search status=ok
  advertises exa_search=true
  load_tools="exa_search"
exa_search   status=error  output="Error: Unknown tool \"exa_search\"."
```

The model is told a tool is loaded, shown its schema, calls it, and is told it does not exist, for the rest of the run rather than the turn. An HTTP MCP handshake routinely exceeds the 1s grace, and Exa is now the default search provider, so this is the ordinary configuration.

The clone is pre-existing; the async registration that makes it diverge is not. On main the registry was complete before any run, so the gap could not open. `partitionToolsCached` already guards the mirror image of this at `loop.go:3120` ("hide deferred tools behind a loader the dispatch gate rejects") — this is the same dead end reached from the other side.

**[P2] Freeform apply_patch drops `cwd`, and the kill switch does not restore it.**

The freeform path builds arguments by hand as `args["patch"] = call.Arguments` with no `cwd`, and the wire definition nils `Parameters` for freeform tools, so `cwd` is not advertised either. `applyPatchTool` then defaults it to `"."`, and `resolveScopedPath` only consults the scope's extra roots for an ABSOLUTE path. The tool's own schema documents that route: "use an absolute path to target a granted extra write root".

So with `--add-dir /srv/shared`, a patch targeting that root is now rejected as outside the workspace, where the JSON form accepted `{"patch": "...", "cwd": "/srv/shared"}`. Recoverable via write_file, but apply_patch can no longer serve a documented workflow on the two surfaces where `--add-dir` exists.

What makes it blocking rather than a trade: `runtimeToolDefinition` marks apply_patch freeform unconditionally, with no env gate, and `buildResponsesRequest` is shared by the WebSocket and HTTP paths. So `ZERO_CHATGPT_TURN_SESSION=0` turns off the new transport and leaves apply_patch freeform. There is no way back to the JSON form.

Everything above I checked against this head rather than taking from the description.

gnanam1990:
Reviewed `35c0906e` against the repo's PR-review protocol. The engineering here is genuinely careful — I tried to break the new patch path two ways and could not. My concern is with **what the headline numbers measure**, and with scope.

## What I verified

- Checked out `35c0906e`, ran `gofmt -l`, `go vet ./...`, `go build ./...`, `go test ./...` — clean apart from `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider`, which fail identically on `main` in this environment. Your validation list holds.
- **Attacked the freeform `apply_patch` path.** `freeformPatchRoot` takes model-supplied absolute paths and selects a write root using `filepath.Rel` plus a string prefix test — lexical, where the repo has `pathjail`/`os.Root`. I probed a symlink that is **already present** when the patch is prepared (the existing test covers the swapped-in-after TOCTOU shape, not this one). It is contained: `Run` errors and nothing lands outside. The lexical check is routing and the real containment is downstream, which is the right architecture. No finding.
- Confirmed the new-transport gating is narrow and reversible: `chatGPTTurnSessionEnabled` defaults on with an explicit kill switch, `openaiTurnSessionEnabled` stays opt-in, and ineligible profiles keep the default adapter.

## Blocking

**1. The turn-efficiency table measures a prompt change, not overhead reduction.** `internal/agent/system_prompt.md` is `+31 −17`, and the changes are behavioural instructions to the model:

- *"Skip update_plan for bounded changes in one component even when the natural workflow is inspect, implement, and verify"*
- *"Once the required behavior passes validation, stop rather than adding optional cleanup, assertions, or refactors."*
- *"Do not read the same file first with read_minified_file and then again with read_file"*

Those are what produce **14 model requests → 4** and **22 tool actions → 6**. Overhead reduction does not change how many turns a task takes; instructing the agent to plan less, verify less and re-read less does.

The PR states it *"keeps the same user-facing capabilities while reducing startup latency and repeated turn overhead."* I do not think the evidence supports the first half. This is a **capability-versus-latency tradeoff**, and it may well be the right one — telling an agent to stop gold-plating is a defensible product decision. But it needs to be named as one and judged on its own terms, not read as a free win alongside RSS and CPU figures.

The measurement can't settle it either: one workload, one hidden oracle, `n=1` after taking the faster of two runs. What that shows is that this particular task still passed while doing a third of the work. The cases that matter are the ones where skipped planning or early-stopped verification changes the outcome, and a single passing oracle cannot see them.

**Concretely:** split the prompt guidance into its own PR with its own before/after over several tasks, and let the infrastructure changes carry the perf title. They stand on their own — the startup and cache work is real and is the part that genuinely is overhead.

**2. No parent issue.** `CONTRIBUTING.md` gates community PRs on an `issue-approved` parent, and I could not find `Fixes #`/`Closes #` in the description. If this is going through the team's internal cycle, ignore this — but the scope rules apply either way.

## Non-blocking

**3. Scope.** 68 files, `+3366 −296`, and the summary itself describes four independent concerns (TUI timers; tool-discovery/registry snapshots; provider session reuse; cache metrics and turn churn). CodeRabbit's generated notes list *new features* — freeform tool support, ChatGPT Responses sessions, background MCP startup, cache-token reporting — under a `perf:` title. Any one of those failing review holds the other three hostage, and a reviewer cannot bisect a regression to a cause.

**4. `go.mod` bumps `charm.land/bubbletea/v2` v2.0.8 → v2.0.9.** Small and justified in the description, but the scope contract calls out dependency changes specifically, so it is worth a maintainer's explicit nod rather than riding along.

## What I did not check

I did not reproduce the benchmark — no access to the same provider, model, fixture or oracle — so I am not disputing the numbers. My argument is about what they attribute to, not whether they were measured.

I also did not review the ChatGPT Responses session-chaining logic in depth. It is the largest behavioural surface here, it defaults **on**, and it deserves its own focused pass from someone who knows that provider path.

gnanam1990 (changes requested):
## Verdict

**Changes requested** — on the claim and the scope, not on the code quality. The engineering is careful and I could not break the part I attacked hardest. What I am asking for is that the headline result be attributed to what actually produces it, and that the PR be split so each piece can be judged on its own evidence.

## What I verified

- Checked out `35c0906e`. `gofmt -l`, `go vet ./...`, `go build ./...`, `go test ./...` — clean except `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider`, which fail identically on `main` in this environment. Your validation list holds.
- **Attacked the freeform `apply_patch` path**, since it takes model-supplied absolute paths and selects a write root using `filepath.Rel` plus a string prefix — lexical, where the repo has `pathjail`/`os.Root`. I probed a symlink already present at preparation time; the existing test covers the swapped-in-after TOCTOU shape, not that one. **Contained**: `Run` errors, nothing lands outside. The lexical check is routing and containment is enforced downstream, which is the right architecture. No finding.
- Confirmed the transport gating is narrow and reversible: `chatGPTTurnSessionEnabled` defaults on with an explicit kill switch, `openaiTurnSessionEnabled` stays opt-in, ineligible profiles keep the default adapter.

## Blocking

**1. The turn-efficiency table measures a prompt change, not overhead reduction.**

`internal/agent/system_prompt.md` is `+31 −17`, and the additions are behavioural instructions to the model:

- *"Skip update_plan for bounded changes in one component even when the natural workflow is inspect, implement, and verify"*
- *"Once the required behavior passes validation, stop rather than adding optional cleanup, assertions, or refactors."*
- *"Do not read the same file first with read_minified_file and then again with read_file"*

Those are what produce **14 model requests → 4** and **22 tool actions → 6**. Reducing overhead does not change how many turns a task takes. Instructing the agent to plan less, verify less, and re-read less does.

The PR states it *"keeps the same user-facing capabilities while reducing startup latency and repeated turn overhead."* The evidence does not support the first half. This is a **capability-versus-latency tradeoff**, and it may well be the right one — telling an agent to stop gold-plating is a defensible product decision with real user benefit. But it has to be named as one, because presented next to RSS and CPU figures it reads as a free win.

The measurement cannot settle it either: one workload, one hidden oracle, n=1 after taking the faster of two post-change runs. What that establishes is that this particular task still passed while doing a third of the work. The cases that decide whether capability was lost are the ones where skipped planning or early-stopped verification changes the outcome, and a single passing oracle cannot see them.

To be clear about what I am *not* claiming: I did not reproduce the benchmark and I am not disputing the numbers. My argument is about what they are attributed to.

**Asking for:** move the `system_prompt.md` guidance into its own PR, with before/after across several tasks rather than one. The infrastructure work then carries the perf title honestly — the startup, registry-snapshot and cache changes are real, and they are the part that genuinely is overhead.

**2. No parent issue.** `CONTRIBUTING.md` gates community PRs on an `issue-approved` parent and I could not find `Fixes #`/`Closes #` in the description. Disregard if this is going through the team's internal cycle — but the scope rules apply either way.

## Non-blocking

**3. Scope.** 68 files, `+3366 −296`, four independent concerns by the summary's own structure. CodeRabbit's generated notes list *new features* — freeform tool support, ChatGPT Responses sessions, background MCP startup, cache-token reporting — under a `perf:` title. Any one of them failing review holds the other three hostage, and a bisected regression cannot be attributed to a cause.

**4. `go.mod` bumps `charm.land/bubbletea/v2` v2.0.8 → v2.0.9.** Small and explained, but the scope contract names dependency changes specifically, so it wants an explicit maintainer nod rather than riding along.

## Not covered by this PR

The ChatGPT Responses session-chaining logic is the largest behavioural surface here and it defaults **on**. I did not review it in depth and it deserves a focused pass from someone who knows that provider path — particularly around what happens when a chained session is resumed against a response the server has expired or garbage-collected.

Vasanthdev2004 (changes requested):
Re-reviewed at `35c0906e`. **Both of my blockers are fixed**, and I checked each rather than reading the commits.

**tool_search / registry divergence: fixed.** `Clone` now rebinds through `cloneForRegistry`, so the snapshot's `tool_search` points at the snapshot. `TestRegistryCloneKeepsToolSearchOnFrozenSnapshot` covers exactly the case I reported, and it is load-bearing: disabling the rebinding fails it with `snapshot tool_search loaded late tool "late_lookup"`, which is the symptom I hit. I like that it also asserts the clone still finds pre-existing tools, so the fix could not have been "make search see nothing".

**Freeform apply_patch and `cwd`: fixed.** Worth reporting how I got there, because my first attempt said otherwise. I drove `Run` directly with `{"patch": ...}` and got the old rejection, which looked like the fix had not worked. It had; I was bypassing `PrepareFreeformApplyPatchArguments`, which is the entry point the loop actually uses. Through it:

```
prepare err=<nil>  cwd=...\002          (the granted extra root)
status=ok  file="new\n"
workspace-relative: status=ok file="new\n"
```

So an absolute path into an `--add-dir` root selects that root and applies, and workspace-relative patches are unchanged. My mistake, and a reminder that a probe at the wrong layer produces a confident false negative.

**Where that leaves my position.** My framing point stands, and gnanam1990 has since found a second, sharper instance of the same shape, so I want to back that rather than open a third front.

I verified their claim independently instead of taking it. `internal/agent/system_prompt.md` is +31/-17 and the additions are behavioural: skip `update_plan` for bounded changes, "stop rather than adding optional cleanup, assertions, or refactors", do not re-read a file with `read_file` after `read_minified_file`, reserve full-suite runs for milestones. Those instruct the agent to plan less, verify less and re-read less. Reducing overhead does not change how many turns a task takes; that does. The attribution is right, and 14 requests to 4 with 22 tool actions to 6 is what you would expect from it.

So there are now two things inside a PR titled for performance that are not performance work, and both change behaviour rather than cost:

- the prompt guidance, which trades thoroughness for turns
- the ChatGPT Responses turn session, still default-on, which moves conversation state to the provider between turns

Each is defensible. Telling an agent to stop gold-plating is a real product improvement, and a stateful session is the right transport for that API. Neither is what a reader of "reduce startup and turn overhead" expects to be approving, and neither can be argued on its merits while bundled with genuine overhead work that nobody disputes.

I am keeping changes-requested, on the same ground gnanam1990 asked for: split it. The startup, registry-snapshot and cache work is real, it is what the title describes, and it would go in on its own evidence quickly. The prompt guidance and the turn-session default each deserve their own PR and their own argument, the latter with a note about what state now lives with the provider.

The code itself I have no further findings on. CI is fully green, and the two `TestRunDoctor*` failures gnanam1990 mentions fail identically on `main` here too.

gnanam1990 (dismissed):
## Verdict

**Approve.** I set out to break this and could not. Withdrawing my earlier scope and process asks — those were mine to raise once, not to hold the PR on, and the code stands on its own.

## What I actually attacked

**The freeform `apply_patch` path**, hardest, because it takes model-supplied absolute paths and picks a write root with `filepath.Rel` plus a string prefix — lexical, where the repo has `pathjail`/`os.Root`. Your own test covers the TOCTOU shape (symlink swapped in after preparation), so I probed the one it does not: a symlink **already present** when the patch is prepared. Contained — `Run` errors, nothing lands outside. The lexical check is routing and containment is enforced downstream, which is the right split. `TestUnknownFreeformToolFailsClosed` confirms the dispatch refuses a freeform call for anything that is not the built-in tool, before arguments are ever prepared.

**The Responses session chaining** — the largest behavioural surface here, and it defaults on, so I went through it properly rather than flagging it unread.

- `responseInputDelta` requires request properties to match AND a deep-equal prefix before it will chain. Anything else returns false and falls back to a full request. Conservative in the right direction.
- An expired chain is handled explicitly: `previousResponseNotFoundCode` falls back to HTTP with the **full** request. The `!state.emitted` guard on that path is the detail I was looking for — once tokens have reached the caller you cannot silently replay, and the code knows it.
- Chain state (`lastRequest` / `lastResponseID` / `lastOutput`) is written under the mutex, guarded by `session.connection == connection` so a stale goroutine cannot poison a newer connection's chain, and the input is deep-copied rather than aliasing the caller's slice.
- `go test -race ./internal/providers/openai/ ./internal/providers/ -count=3`: clean, 543s.

**Background MCP startup**, where a timeout usually leaks. It does not: `done` is buffered, an abandoned connect is reaped in its own goroutine which closes any partial client, each server gets its own cancellable context so a timeout tears down the subprocess, and the commit is serial in server order so registration stays deterministic.

**Gauntlet** on `35c0906e`: `gofmt`, `go vet ./...`, `go build ./...`, `go test ./...` — clean apart from `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider`, which fail identically on `main` in this environment.

## One thing to record, not to block on

`system_prompt.md` is `+31 −17`, and those lines are behavioural instructions — skip `update_plan` for bounded changes, stop once validation passes, do not re-read a file with both readers. They are a real part of what turns 14 model requests into 4 and 22 tool actions into 6, alongside the infrastructure work.

I am not asking for anything here. It is worth knowing when reading the table that some of the gain is the agent doing less rather than the harness doing less, because the two age differently: infrastructure wins hold, while prompt-shaped wins move with the next model. Useful for whoever looks at these numbers in six months.

## Not covered

I did not reproduce the benchmark — no access to the same provider, model, fixture or oracle — so the numbers are taken as reported. My review is of the code.

jatmn (changes requested):

## Merge readiness

  `internal/providers/openai/codex_responses.go:1`
  This head is 19 commits behind live `main` (`ad34dc8d`), including current changes in files this branch modifies and a security-sensitive command-normalization fix. The repository policy requires a fresh base; rebase and re-review the resolved diff before merging.

anandh8x:
Rebased onto current main (`ad34dc8d`) at `94ff28f5`. The rebase was conflict-free and preserved all 19 commits. Formatting, vet, lint, vulncheck, build, and smoke pass; CI is running. @jatmn please re-review when ready.

Vasanthdev2004 (dismissed):
Approving at `94ff28f5`.

**Both of my blockers are fixed, and I re-verified them on this head rather than trusting that they survived the rebase.** That mattered here, because a rebase is exactly where a fix gets carried and its caller does not.

`cloneForRegistry` is present in both `tool_search.go` and `registry.go`, and `TestRegistryCloneKeepsToolSearchOnFrozenSnapshot` passes. `freeformPatchRoot` is present, and driving the real entry point end to end:

```
extra write root:    cwd=...\002  status=ok  file="new\n"
workspace-relative:  status=ok    file="new\n"
```

`internal/tools`, `internal/agent` and `internal/providers/openai` all green here.

**jatmn's finding is already resolved.** It was raised against `35c0906e`, and the branch has been rebased since: the merge base is `ad34dc8d`, which is current `origin/main`, and `origin/main` is 0 commits ahead of it. Nothing to do; worth saying so explicitly so this does not sit blocked on a resolved item.

**On scope, I am withdrawing the hold.** gnanam1990 put it better than I would: raise it once, record it, do not hold the PR on it. I have made the point, they made a sharper version of it, and it is on the record twice now. Continuing to block a change that two reviewers have attacked and could not break, on a framing preference, is not a good use of anyone's time.

One thing I want recorded rather than argued, because it is the half gnanam1990's note does not cover. Their point is about attribution: some of the table is the agent doing less rather than the harness doing less. Mine is narrower and is about data handling. The ChatGPT Responses turn session defaults on, and chaining on `previous_response_id` means conversation state is retained by the provider between turns instead of being resent each time. The provider sees the same content either way, so this is a change in retention rather than in disclosure, it is confined to one provider whose API is built around exactly this, and `ZERO_CHATGPT_TURN_SESSION=0` turns it off. That is why it is a note and not a blocker. It still belongs in the release notes, because "conversation state now persists with the provider between turns" is the kind of thing an operator wants to read in a changelog rather than discover.

The infrastructure work is the part the title describes and it is good: the background MCP startup teardown, the registry snapshot, and the cache changes all hold up. Nice work on the `!state.emitted` guard around the expired-chain fallback in particular; that is the detail that would have been easy to miss.

jatmn (changes requested):

## Findings

- [P2] Keep workspace-relative patch paths in the workspace
  `internal/tools/apply_patch.go:42`
  `PrepareFreeformApplyPatchArguments` scans a native structured patch for absolute headers. When it finds one beneath a granted extra write root, it rewrites that header relative to the extra root and assigns the root to the single `cwd` argument used for the entire patch. It does not rewrite or reject the other relative headers. `applyPatchTool.Run` subsequently resolves every parsed operation relative to that one `cwd`, so a patch such as `*** Update File: src/a.go` plus `*** Add File: /granted-extra/b.go` attempts to update `/granted-extra/src/a.go`, not `<workspace>/src/a.go`. If an identically named file exists in the extra root, the wrong file is changed; otherwise a valid mixed patch fails with a missing-file error.

  The root cause is treating a multi-operation patch with paths from different semantic roots as though it can be represented by one global working directory. Please make root selection an explicit invariant for the complete patch: either reject a patch that mixes workspace-relative headers with absolute headers resolved to a non-workspace extra root, or normalize and validate every header against one selected root before passing it to the rooted patch engine. Add regression coverage for both the wrong-target case (a same-named file exists under the extra root) and the failure case (it does not), while preserving valid all-workspace and valid all-extra-root patches, current permission checks, and rooted containment.

jatmn (approved):

## Merge readiness

  `internal/agent/compaction.go:144`
  The head is based on `2749447c`, while live `main` is `6fe0d1ed` and has three newer commits. In particular #532 changes the compaction, agent-loop, CLI, and TUI integration surfaces this PR also modifies. Rebase onto the live target and re-run the focused review/validation on the resolved integration diff.

gnanam1990 (approved):
**APPROVED** on `c6b805af`.

My previous approval was on `35c0906e` and got dismissed when the branch was rebased and merged with main, so I re-reviewed rather than restoring it. The only genuinely new work on top of what I read before is `c6b805af`, plus the merge that brought in #956 — which touches the same `apply_patch` area, so I ran the whole gauntlet on the merged tree rather than just the delta.

**`fix(tools): reject mixed-root freeform patches` is a real confused-deputy fix, not a cosmetic one.**

`prepareFreeformArguments` rewrites absolute headers to be relative and sets `args["cwd"] = selectedRoot` for the whole patch. Relative headers are left alone — so in a patch that mixes the two, a header the model wrote as `src/a.go`, meaning the workspace, silently re-resolved against whichever extra root the absolute header selected. The write lands in a different tree than the one the path names, and nothing in the patch says so.

What I checked beyond reading it:

- **The guard is on the class, not the instance.** Two absolute headers in different roots were already rejected by the pre-existing `spans multiple write roots` check; this closes the remaining shape, which is relative-plus-absolute. Between them every mixed-root combination now fails before any write.
- **`roots[0]` is safe.** `scopedRoots` fails closed with an error on an empty `Roots()` rather than returning a short slice, so the index cannot panic, and `PathScope` documents the workspace root as first. With a nil scope it returns `[]string{workspaceRoot}`, so `selectedRoot == roots[0]` and a workspace-only patch still passes.
- **`extraHasMatchingPath: true` is the arm that matters** and it is there — an existing `<extra>/src/a.go` that the old code would have overwritten. The test asserts both targets are untouched and the added file was never created, so it pins containment rather than just the error string.
- **Mutation-checked the guard.** Removing it, and inverting `selectedRoot != roots[0]` to `==`, both fail `TestFreeformApplyPatchRejectsMixedWorkspaceAndExtraRoots`; never setting `hasRelativeHeader` fails it too. `TestFreeformApplyPatchAcceptsOneSemanticRoot` covers the other direction for both the workspace and the extra root, so the fix is not just a blanket refusal.

**Gauntlet on `c6b805af`:** `fmt-check`, `go vet ./...`, `make lint-static` (0 issues), `git diff HEAD --check`, `zero-release build`, `zero-release smoke` all clean. `go test ./...` passes except `TestRunDoctorFormatsRedactedProviderDiagnostics` and `TestRunDoctorConnectivityProbesProvider` in `internal/cli` — both reproduce on unmodified `6fe0d1ed` in a clean worktree here, so they are environmental and not from this branch.

Nothing blocking from me.

kevincodex1 (approved):

PR #935, opened 2026-08-21
security(mcp): fix stdio dispatcher type confusion (Z-052)

### LINKED ISSUE #924: security: MCP stdio dispatcher misdelivers server requests as responses (Z-052)
### Description
The MCP (Model Context Protocol) stdio dispatcher in internal/mcp/client.go suffers from a message-type confusion bug. The current 
eadLoop implementation filters primarily on whether message.ID is nil to determine if a message is a response or a request, but it does not explicitly validate the method or the specific JSON-RPC message type.

When an MCP server initiates its own request to the client, the dispatcher mistakenly treats this incoming request as a response to a previously sent client request. This leads to the dispatcher returning an empty result or a malformed response to the original caller, as the server's request is consumed and discarded without being routed to a proper request handler.

### Impact
This causes silent failures and incorrect behavior when interacting with sophisticated MCP servers that utilize server-initiated requests (e.g., for capability negotiation or resource updates). The user sees a tool return 'no results' or an empty string even when the server is functioning correctly.

### Recommended Fix
1. **Explicit Type Validation**: Update the 
eadLoop to explicitly check the JSON-RPC message structure. Only messages that are explicitly marked as *responses* (containing a matching id to a pending request) should be returned to the caller.
2. **Request Routing**: Implement a separate dispatch path for *requests* initiated by the server. These should be routed to a handler that allows the client to respond to the server, rather than letting them be swallowed by the response loop.
3. **Improved Diagnostics**: Add logging for unexpected message types to help diagnose protocol mismatches between the client and various MCP server implementations.

vedantlavale:
Fixes MCP stdio dispatcher type confusion (Z-052) in `internal/mcp/client.go:381`.

The `readLoop` previously classified messages only by `ID == nil`. A server-initiated request `{"id":1,"method":"..."}` with an `id` colliding with a pending client request `{"id":1,"method":"tools/list"}` was dispatched to `pending[1]` as if it were a response. `request()` at `client.go:347` then saw `Method` set but `Result`/`Error` empty and returned an empty `CallToolResult`, surfacing as silent empty tool output.

Changes:
* `internal/mcp/client.go:393-418` – Explicit type validation: `if message.Method != ""` is treated as request/notification, never as response. Server requests (`Method != "" && ID != nil`) are routed to a handler that replies `{"id":<same>,"error":{"code":-32601,"message":"method not found: <method>"}}` under `client.mu` and are not dispatched. Notifications (`Method != "" && ID==nil`) are ignored. Only `Method == "" && ID != nil` proceeds to `rpcMessageID`/`pending` lookup. Added `fmt.Fprintf(os.Stderr, "[mcp] ...")` diagnostics for server requests and invalid `ID==nil && Method==""` messages.
* `internal/mcp/network_client.go:542-548` – Mirror check in `remoteSSEClient.deliverEventMessage`: `if message.Method != "" { return nil }` to prevent SSE stream misdelivery when `id` collides.

This implements the three recommended fixes from #924: explicit type validation, separate request routing, improved diagnostics.

## Linked issue

Fixes #924

## Checklist

- [x] `go vet ./...` and `go test ./internal/mcp -race` pass locally (`TestClientRequestWaitsForMatchingResponseID`, `TestStdioClientListsAndCallsTools`, SSE/HTTP clients). Full `go vet ./...` clean.
- [x] `gofmt` clean (`make fmt-check` passes, `git diff --check` clean).
- [x] Verified with `io.Pipe` reproduction: colliding `server/request id=1` before correct `{"id":1,"result":{"value":"matched"}}` previously returned `Value=""` (misdelivered), now correctly returns `Value="matched"` and server receives `-32601 method not found`. Existing tests exercise the `method` vs `id` distinction (`TestClientRequestWaitsForMatchingResponseID`, `TestDecodeSSERPCMessageSkipsNotifications`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected server messages during communication.
  - Server-initiated requests now receive an appropriate “method not found” response instead of disrupting response processing.
  - Notifications and malformed messages are safely ignored.
  - Responses are matched only when they contain valid request identifiers.

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -385,7 +385,34 @@ func (client *Client) readLoop() {
 			client.failAll(err)
 			return
 		}
+		// Explicit type validation: distinguish requests/notifications (have
+		// method) from responses (no method, has id). Only responses may be
+		// delivered to pending callers. Server-initiated requests must never be
+		// misdelivered as a response even when the id collides with a pending
+		// client request (Z-052).
+		if message.Method != "" {
+			if message.ID != nil {
+				// Server-initiated request: route to request handler and
+				// reply with method-not-found so the server does not hang.
+				// The client does not currently handle inbound requests, so
+				// every method is unknown. Log for diagnostics.
+				_, _ = fmt.Fprintf(os.Stderr, "[mcp] server request %q (id %v) not handled: replying method not found\n", message.Method, message.ID)
+				client.mu.Lock()
+				_ = client.writer.write(rpcMessage{
+					JSONRPC: "2.0",
+					ID:      message.ID,
+					Error:   &rpcError{Code: -32601, Message: "method not found: " + message.Method},
+				})
+				client.mu.Unlock()
+			}
+			// Notifications (no id) and handled server requests are not
+			// responses — do not dispatch.
+			continue
+		}
 		if message.ID == nil {
+			// No id and no method: invalid or stray message. Ignore but log
+			// for diagnostics to help trace protocol mismatches.
+			_, _ = fmt.Fprintf(os.Stderr, "[mcp] ignoring unexpected message without id (method %q)\n", message.Method)
 			continue
 		}
 		id, ok := rpcMessageID(message.ID)
--- a/internal/mcp/network_client.go
+++ b/internal/mcp/network_client.go
@@ -539,6 +539,12 @@ func (client *remoteSSEClient) deliverEventMessage(value string) error {
 	if err := decoder.Decode(&message); err != nil {
 		return fmt.Errorf("decode MCP SSE stream message: %w", err)
 	}
+	// Explicit type validation: only messages without a method are responses.
+	// Server-initiated requests (method set) must not be misdelivered to a
+	// pending caller even when the id collides (Z-052).
+	if message.Method != "" {
+		return nil
+	}
 	key := rpcResponseKey(message.ID)
 	if key == "" {
 		return nil


Vasanthdev2004 (changes requested):
The bug is real and the diagnosis is right. Classifying purely on `ID == nil` does misdeliver a server-initiated request into `pending`, and the empty `CallToolResult` symptom follows exactly as you describe. Good find.

The problem is in the reply path rather than the classification, and I don't think it can go in as it stands.

**readLoop holds `client.mu` across the reply write.** `client.go:400` takes the lock and calls `client.writer.write` from the reader goroutine. The invariant that breaks is written down in `request()` itself, a few lines above its own `Lock`:

> The mutex serializes writes and id allocation but is released before the (potentially unbounded) wait for the response, so a hung server never holds the lock and blocks other callers/Close.

Two things make that bite. `request()` takes `client.mu` at `client.go:314` before it reaches its `ctx.Done()` select, so a caller blocked on the mutex cannot be released by its context deadline. And readLoop is the only goroutine draining the child's stdout, so once it blocks in a write, the child stays blocked writing and never gets back to reading its stdin. That cycle sustains itself; only `Close()` breaks it.

So a server that sends more inbound requests than the stdin pipe buffer can hold in replies wedges the client, and every later `CallTool` hangs past its deadline instead of erroring. On main `readLoop` never writes and never touches `client.mu`, so this is a new failure mode rather than an existing one being uncovered.

**`notify()` writes the same `bufio.Writer` with no lock.** `client.go:506`. That was safe before because `readLoop` never wrote, and `notify` is called exactly once, from `initialize()`, on the goroutine that has just finished `request("initialize")`. Adding a writer on the reader goroutine makes it a genuine race on an unsynchronized writer, and `messageWriter.write` is three separate bufio calls, so what comes out is a torn JSON line on the server's stdin. The window is precisely when a server tends to send its first ping, since `initialize()` does the request and then the notify with readLoop already running.

A shape that avoids both: don't write from readLoop at all. Hand the server request to a small outbound queue and let one dedicated writer goroutine own the writer. The reader never blocks, and there is exactly one writer to synchronize.

Smaller things, worth folding into the same pass:

- No tests. `+33 -0` and neither file is a test file, so the whole change is deletable with a green suite. For a security fix I'd want the collision itself pinned: a pending id 1, the server sends `{"id":1,"method":"..."}`, assert the caller does not get an empty result back.
- `client.go:399` prints the server-controlled id with `%v`. `ID` is `any`, so a string id goes out raw and unbounded onto the user's terminal. The method is `%q` and fine. Worth asking whether raw `os.Stderr` is right here at all, since it can land on top of the alt-screen TUI.
- `network_client.go:219` looks like the same shape still unguarded. It checks `rpcIDMatches` but not `Method != ""`, so a message carrying a method would pass the id check and fall through to the empty-result path. The SSE site at 545 got the guard and this one did not.

One thing I checked and don't think you should change: answering a server `ping` with -32601 instead of an empty result is a spec nit that is already true on main, so it isn't yours to carry here unless you want it.

euxaristia:
@coderabbitai full review

gnanam1990:
Closing this PR as superseded by #942 for approved issue #924, so the dispatcher fix has one implementation and review thread. The diagnosis here is correct, but current head fb6153d still performs the method-not-found reply while holding client.mu, which can block outbound callers or Close on an undrained peer, and this PR does not carry regression coverage for the collision and blocking paths. #942 contains the more complete current implementation and regression suite, though its outstanding review findings still need resolution before merge.

Thank you for finding and documenting the type-confusion bug. Please continue any useful discussion or evidence on #942 or #924.

PR #938, opened 2026-08-22
fix(oauth): Split an oversized keyring token blob across entries

### LINKED ISSUE #937: oauth: keyring storage cannot save a second login on macOS
## Summary

With `ZERO_OAUTH_STORAGE=keyring` on macOS, a second OAuth login can fail to
save. `Store.Save` returns `keyring: set: secret too large for the macOS
security tool's 4095-byte command line; use the file backend instead`, and no
token is persisted. Once the store is over the line every subsequent write
fails too, not just the login that crossed it, because the whole blob is
rewritten on each save.

## Cause

`keyringBlob` puts every provider and MCP token into one JSON blob under a
single keyring entry (`internal/oauth/store.go:472`), base64-encoded. On macOS
the secret rides inside a `security -i` command line, whose parser reads at most
4095 bytes, so `Keyring.Set` refuses anything larger rather than let the tool
split the line into two garbage commands (`internal/keyring/keyring.go:87`).

The blob has no size bound of its own. Nothing caps how many logins it holds, so
it grows past a limit that does not grow with it.

## Measured budget

Under the anchor account the command overhead is 56 bytes, leaving 4039 bytes of
base64, which is 3027 bytes of JSON for all logins combined.

Encoding a `storeFile` with realistic OIDC credentials (a JWT access token, an
ID token, an opaque refresh token, scopes, expiry, account):

| credential (access / id / refresh) | logins that fit |
| --- | --- |
| 600 / 0 / 64 bytes | 3 |
| 1200 / 900 / 300 bytes | 1 |
| 1800 / 1200 / 400 bytes | 0 |

A single large login can exceed the cap on its own, and two ordinary ones do.

## Reproduction

1. `export ZERO_OAUTH_STORAGE=keyring` on macOS.
2. Log in to one provider whose OAuth flow returns an ID token alongside the
   access and refresh tokens.
3. Log in to a second provider.

Expected: both logins are stored.
Actual: the second `Save` fails with `secret too large`, nothing is written, and
further saves keep failing until the store is cleared or moved to file storage.

## Notes

Linux is unaffected: `secret-tool` reads the secret from stdin, so there is no
command line to fill.

The fix has to stay compatible with entries written by existing builds, and
must not leave token material behind in the keychain when the layout changes.

euxaristia:
## Summary

`ZERO_OAUTH_STORAGE=keyring` on macOS cannot save a second OAuth login. Every
provider and MCP token shares one keyring entry, and on macOS the secret rides
inside a `security -i` command line capped at 4095 bytes. Measured, that leaves
4039 bytes of base64 under the anchor account, or 3027 bytes of JSON for all
logins combined. A single large OIDC credential can fill it alone; two ordinary
ones do reliably. Once over the line every write fails, not just the login that
crossed it, because the whole blob is rewritten on each save.

This splits a blob that does not fit across numbered entries and puts a manifest
in the anchor account.

Fixes #937

## Changes

**`internal/keyring`: expose the per-entry budget.**
`MaxSecretLen(service, account) (int, bool)` reports the largest secret `Set`
accepts, with `ok` false when the backend has no practical limit. It shares one
line builder with `Set`, so the budget and the boundary it describes cannot
drift. The account is part of the figure because on macOS it shares the command
line with the secret. Linux reports unbounded: `secret-tool` reads the secret
from stdin, so there is no command line to fill.

**`internal/oauth`: chunk an oversized blob.**

- A blob that fits stays one entry, byte-identical to today. Unbounded backends
  never reach any new code.
- A blob that does not fit is split, and the anchor holds
  `zc1:<live>:<countA>:<countB>:<sha256>`. `:` is outside the base64 alphabet,
  so a stored blob can never carry that prefix and an entry written by an
  existing build is read without a migration step.
- Chunks live in two alternating generations. A write fills the one that is not
  live, then replaces the manifest. That single `Set` is the commit point, so
  until it lands a reader still gets the previous generation whole.
- Chunks are sized against the longest account name the generation can produce.
  A budget taken from chunk 0 would overflow once the index grew a digit.
- The manifest carries a digest of the payload. The corruption being guarded
  against is the one that motivated chunking: `security -i` splits an overlong
  line into two garbage commands rather than refusing it, so a chunk can come
  back truncated and still be valid base64.
- A write reserves the range it will occupy before occupying it. Without that,
  a write interrupted while filling a longer generation leaves chunks above the
  recorded count and nothing would ever delete them. At the one transition where
  reserving would destroy the only copy (the anchor still holds the whole blob),
  the target generation is swept instead.
- The retired generation is deleted after the commit. Its count deliberately
  stays in the manifest: over-stating is the safe direction, and a failed delete
  is retried by the next write. The invariant is one-sided, and tested as such:
  the manifest may over-state what a generation holds, never under-state it.

## Cost

A steady-state save on macOS goes from 1 `security` invocation to 5 (2 chunk
writes, 1 manifest commit, 2 retirement deletes); a load goes from 1 to N+1.
Roughly 50ms on login and refresh, and only for stores that exceed one entry,
which today cannot save at all.

## Test plan

12 new tests. Nine fail on the unfixed path, each for the reason it names,
verified by disabling only the chunking decision in `keyringBlob.write`:

```
--- FAIL: TestStoreKeyringSavesSecondLoginOverEntryLimit
    Save(second): keyring: secret too large (7312 > 4083)
--- FAIL: TestStoreKeyringReservesChunkRangeBeforeFilling
    generation "b" holds 5 chunks ([...b.0 ...b.4]) but the manifest counts 0
--- FAIL: TestStoreKeyringSweepsStrayChunksOnFirstGrowth
    stray chunk oauth-tokens.a.3 survived the growth into the chunked layout
```

Also covered: the commit point (a write that dies while filling leaves the
committed blob readable and the manifest unmoved), generation alternation with
no stray chunks, growth into chunks and back out again, a missing chunk, a
truncated chunk caught by the digest, two-digit chunk indices, malformed
manifests, an unbounded backend keeping the single-entry layout, and reading an
entry written by an existing build.

`internal/keyring` gains three tests pinning `MaxSecretLen` to the boundary
`Set` actually enforces: a secret of exactly the budget is accepted, one byte
more is rejected, the figure shrinks with the account name, and non-darwin
reports unbounded.

Commands run on `ad34dc8`:

- `gofmt -l $(git ls-files '*.go')` clean
- `go vet ./...` clean
- `go test ./... -count=1` green except `internal/imageinput` and
  `internal/sandbox`, which fail identically on a clean tree here (WSL2
  clipboard contents and WSL2 sandbox backend detection)
- `-race` not run locally: no C toolchain on this machine. The change adds no
  concurrency, so the race surface is unchanged, but CI should confirm.

## Summary by CodeRabbit

* **New Features**
  * OAuth tokens larger than platform keyring limits can now be stored and retrieved automatically.
  * Added platform-aware reporting of supported secret sizes.
  * Added `zero auth reset` to clear saved OAuth tokens and recover from corrupted token data.
  * Existing and unlimited-capacity keyring storage continues to work without changes.

* **Bug Fixes**
  * Improved reliability when saving, updating, and recovering large tokens.
  * Added validation for incomplete or corrupted token data.
  * Improved handling of interrupted updates, stale data, and concurrent access.
  * Added clearer guidance to sign in again when recovery is unsuccessful.

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -600,7 +600,7 @@ func runAuthReset(args []string, stdout io.Writer, stderr io.Writer, deps appDep
 		}
 		return exitSuccess
 	}
-	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store. All stored credentials and entries cleared."); err != nil {
+	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store."); err != nil {
 		return exitCrash
 	}
 	return exitSuccess
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -74,6 +74,8 @@ func runAuth(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 		return runAuthStatus(args[1:], stdout, stderr, deps)
 	case "refresh":
 		return runAuthRefresh(args[1:], stdout, stderr, deps)
+	case "reset":
+		return runAuthReset(args[1:], stdout, stderr, deps)
 	case "openrouter":
 		return runAuthOpenRouter(args[1:], stdout, stderr, deps)
 	case "chatgpt":
@@ -321,6 +323,7 @@ func validateAuthFlags(sub string, a authArgs) error {
 		"logout":  {"json": true},
 		"status":  {"json": true},
 		"refresh": {"watch": true},
+		"reset":   {"json": true},
 	}[sub]
 	bad := func(name string) error { return fmt.Errorf("zero auth %s does not accept %s", sub, name) }
 	if a.json && !allowed["json"] {
@@ -569,6 +572,40 @@ func runAuthRefreshWatch(manager *oauth.Manager, key, provider string, stdout io
 	return exitSuccess
 }
 
+func runAuthReset(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
+	parsed, err := parseAuthArgs("reset", args)
+	if err != nil {
+		return writeExecUsageError(stderr, err.Error())
+	}
+	if parsed.help {
+		_ = writeAuthHelp(stdout)
+		return exitSuccess
+	}
+	if len(parsed.positional) > 0 {
+		return writeExecUsageError(stderr, fmt.Sprintf("zero auth reset takes no arguments (got %q)", parsed.positional[0]))
+	}
+	manager, err := newAuthManager(deps, stdout)
+	if err != nil {
+		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
+	}
+	if err := manager.Reset(); err != nil {
+		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
+	}
+	if parsed.json {
+		payload := struct {
+			Reset bool `json:"reset"`
+		}{Reset: true}
+		if err := writePrettyJSON(stdout, payload); err != nil {
+			return exitCrash
+		}
+		return exitSuccess
+	}
+	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store. All stored credentials and entries cleared."); err != nil {
+		return exitCrash
+	}
+	return exitSuccess
+}
+
 func filterAuthStatuses(statuses []oauth.Status, provider string) []oauth.Status {
 	want := oauth.ProviderKey(provider)
 	filtered := make([]oauth.Status, 0, 1)
@@ -589,6 +626,7 @@ Commands:
   logout <provider>                               Delete a provider's stored login
   status [provider]                               Show login presence/expiry (never the token)
   refresh <provider> [--watch]                    Force a token refresh (--watch keeps it fresh)
+  reset                                           Reset the OAuth token store (clears corrupted entries and tokens)
   openrouter                                      Log in to OpenRouter in the browser; mints an API key
   chatgpt                                         Log in to ChatGPT in the browser (Codex backend, ChatGPT Plus/Pro)
 
@@ -615,7 +653,7 @@ Flags:
       --device   Use the device-code flow (headless/SSH; no browser)
       --scope    Add an OAuth scope (repeatable)
       --watch    Keep the token fresh in the foreground (refresh only)
-      --json     Print result as JSON (status/logout)
+      --json     Print result as JSON (status/logout/reset)
   -h, --help     Show this help
 `)
 	return err
--- a/internal/cli/auth_test.go
+++ b/internal/cli/auth_test.go
@@ -196,10 +196,39 @@ func TestRunAuthHelp(t *testing.T) {
 	if code := runWithDeps([]string{"auth", "--help"}, &stdout, &stderr, appDeps{}); code != exitSuccess {
 		t.Fatalf("exit = %d", code)
 	}
-	for _, want := range []string{"zero auth", "login", "logout", "status", "refresh", "--device"} {
+	for _, want := range []string{"zero auth", "login", "logout", "status", "refresh", "reset", "--device"} {
 		if !strings.Contains(stdout.String(), want) {
 			t.Fatalf("help missing %q:\n%s", want, stdout.String())
 		}
+	}
+}
+
+func TestRunAuthReset(t *testing.T) {
+	path := withAuthStore(t)
+	store, err := oauth.NewStore(oauth.StoreOptions{FilePath: path})
+	if err != nil {
+		t.Fatalf("NewStore: %v", err)
+	}
+	if err := store.Save(oauth.ProviderKey("demo"), oauth.Token{AccessToken: "secret"}); err != nil {
+		t.Fatalf("Save: %v", err)
+	}
+
+	var stdout, stderr bytes.Buffer
+	if code := runWithDeps([]string{"auth", "reset"}, &stdout, &stderr, appDeps{}); code != exitSuccess {
+		t.Fatalf("exit = %d stderr=%s", code, stderr.String())
+	}
+	if !strings.Contains(stdout.String(), "Reset OAuth token store") {
+		t.Fatalf("stdout = %q", stdout.String())
+	}
+
+	// Verify store is now empty.
+	stdout.Reset()
+	stderr.Reset()
+	if code := runWithDeps([]string{"auth", "status"}, &stdout, &stderr, appDeps{}); code != exitSuccess {
+		t.Fatalf("status exit = %d", code)
+	}
+	if !strings.Contains(stdout.String(), "No OAuth provider logins are stored.") {
+		t.Fatalf("expected empty store after reset, got: %q", stdout.String())
 	}
 }
 
--- a/internal/cli/completions.go
+++ b/internal/cli/completions.go
@@ -78,7 +78,7 @@ var completionRoot = completionNode{
 			{names: []string{"tools"}, children: leafNodes("list")},
 			{names: []string{"oauth"}, children: leafNodes("login", "logout", "status")},
 		}},
-		{names: []string{"auth"}, children: leafNodes("openrouter", "chatgpt", "login", "logout", "status", "refresh")},
+		{names: []string{"auth"}, children: leafNodes("openrouter", "chatgpt", "login", "logout", "status", "refresh", "reset")},
 		{names: []string{"sandbox"}, children: []completionNode{
 			{names: []string{"policy"}}, {names: []string{"setup"}}, {names: []string{"check"}},
 			{names: []string{"grants"}, children: leafNodes("list", "allow", "deny", "revoke", "clear")},
--- a/internal/cli/completions_test.go
+++ b/internal/cli/completions_test.go
@@ -159,6 +159,7 @@ func TestCompletionTreeCoversAliasesNestingAndCommonFlags(t *testing.T) {
 	assertCandidates(t, byPath["worktree"], "prepare", "release")
 	assertCandidates(t, byPath["daemon"], "start", "stop", "status", "run", "attach")
 	assertCandidates(t, byPath["mcp oauth"], "login", "logout", "status")
+	assertCandidates(t, byPath["auth"], "openrouter", "chatgpt", "login", "logout", "status", "refresh", "reset")
 	assertCandidates(t, byPath["sandbox grants"], "list", "allow", "deny", "revoke", "clear")
 	assertCandidates(t, byPath["completions"], "bash", "zsh", "fish", "powershell", "elvish")
 	assertCandidates(t, byPath["plugins"], "list", "add", "info", "remove", "rm")
--- a/internal/keyring/keyring.go
+++ b/internal/keyring/keyring.go
@@ -9,7 +9,9 @@
 // through `security -i` (interactive mode), whose command parser is line-based
 // with a fixed 4096-byte line buffer, so Set rejects secrets containing
 // newlines or exceeding that budget rather than silently corrupting them;
-// Linux's secret-tool has no such restriction.
+// Linux's secret-tool has no such restriction. MaxSecretLen reports that budget
+// so a caller holding a larger value can split it across entries instead of
+// being turned away.
 package keyring
 
 import (
@@ -82,7 +84,7 @@ func (k *Keyring) Set(service, account, secret string) error {
 			}
 		}
 		// -U updates the item if it already exists rather than failing.
-		line := "add-generic-password -U -s " + securityQuote(service) + " -a " + securityQuote(account) + " -w " + securityQuote(secret) + "\n"
+		line := securityAddLine(service, account, secret)
 		if len(line) > securityMaxLine {
 			return wrap("set", fmt.Errorf("secret too large for the macOS security tool's %d-byte command line; use the file backend instead", securityMaxLine))
 		}
@@ -162,10 +164,41 @@ func (k *Keyring) Delete(service, account string) (bool, error) {
 	}
 }
 
+// MaxSecretLen reports the largest secret, in bytes, that Set accepts under
+// (service, account) on this platform. ok is false when the backend imposes no
+// practical limit, which is the case for Linux's secret-tool: it reads the
+// secret from stdin rather than a command line. On macOS the secret shares the
+// `security -i` command line with the service and account, so the budget
+// depends on both and a caller that splits an oversized value must size each
+// part against the exact account name it will be stored under.
+//
+// The budget assumes the secret needs no quote escaping. securityQuote only
+// expands `\` and `"`, neither of which appears in base64 or hex, so a caller
+// storing encoded data gets the exact figure.
+func (k *Keyring) MaxSecretLen(service, account string) (int, bool) {
+	if k.goos != "darwin" {
+		return 0, false
+	}
+	// Measuring the real command with an empty secret counts the surrounding
+	// quotes too, so a secret of exactly this length lands on securityMaxLine.
+	budget := securityMaxLine - len(securityAddLine(service, account, ""))
+	if budget < 0 {
+		budget = 0
+	}
+	return budget, true
+}
+
 // securityMaxLine is the most `security -i` can read as one command: its line
 // buffer is MAX_LINE_LEN (4096) bytes in Apple's SecurityTool, one of which the
 // terminating NUL consumes.
 const securityMaxLine = 4095
+
+// securityAddLine renders the `security -i` command that stores secret under
+// (service, account). Set writes it and MaxSecretLen measures it, so the two
+// cannot disagree about how much of the line the overhead consumes.
+func securityAddLine(service, account, secret string) string {
+	return "add-generic-password -U -s " + securityQuote(service) + " -a " + securityQuote(account) + " -w " + securityQuote(secret) + "\n"
+}
 
 // securityQuote wraps s for one argument of a `security -i` command line. The
 // tool's parser (split_line in Apple's SecurityTool) treats a backslash inside
--- a/internal/keyring/keyring_test.go
+++ b/internal/keyring/keyring_test.go
@@ -375,3 +375,46 @@ func TestAvailable(t *testing.T) {
 		t.Fatal("linux should be available")
 	}
 }
+
+// TestMaxSecretLenMatchesTheDarwinSetBoundary pins the budget to the boundary
+// Set actually enforces. A caller that splits an oversized value depends on the
+// two agreeing exactly: one byte of drift either wastes an entry or produces a
+// chunk security silently splits into two garbage commands.
+func TestMaxSecretLenMatchesTheDarwinSetBoundary(t *testing.T) {
+	k := newFake("darwin").keyring()
+	budget, bounded := k.MaxSecretLen("zero", "oauth-tokens")
+	if !bounded {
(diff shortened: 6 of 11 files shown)


euxaristia:
Addressed the review findings in 80de5f3:
- Protected `Load` and `Status` with `withLock` in [`internal/oauth/store.go`](file:///home/euxaristia/Projects/zero/internal/oauth/store.go#L285-L355) so standalone keyring reads hold cross-process lock protection against interleaved manifest commit/chunk rotation.
- Added `hex.DecodeString` validation in `parseKeyringManifest` to reject non-hex digests during parsing.
- Updated `assertNoStrayChunks` in [`internal/oauth/store_keyring_chunked_test.go`](file:///home/euxaristia/Projects/zero/internal/oauth/store_keyring_chunked_test.go#L380-L400) to validate chunk indices.
- Added regression tests for non-hex manifest digests, failure during final manifest publication (`TestStoreKeyringWriteFailsOnFinalManifestPublication`), and concurrent reader/writer lock synchronization (`TestStoreKeyringReadSerializedWithLockDuringChunkedWrite").

euxaristia:
Addressed CodeRabbit follow-up in 0566be0:
- Extended `TestStoreKeyringReadSerializedWithLockDuringChunkedWrite` in [`internal/oauth/store_keyring_chunked_test.go`](file:///home/euxaristia/Projects/zero/internal/oauth/store_keyring_chunked_test.go#L515-L545) to concurrently invoke and validate that `reader.Status` blocks while the cross-process lock is held and succeeds upon release.

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

euxaristia:
@coderabbitai full review

Vasanthdev2004 (changes requested):
This is careful work and the design holds up. I went looking specifically for a torn read and could not construct one: the write fills the generation that is not live and the manifest `Set` is the only commit point, so a reader either sees the old manifest with the old chunks intact or the new one with the new chunks complete. Sizing chunks against the longest account name the generation can produce, rather than against chunk 0, is the kind of detail that would have caused a corruption bug at index 10. Sharing one line builder between `Set` and `MaxSecretLen` is the right way to keep the budget and the boundary from drifting, and the test that pins both sides of the boundary is what makes that stick.

I checked the two things the design leans on and they are sound. `zc1:` cannot collide with a stored blob, because `:` is outside the base64 alphabet and lands at index 3. And a manifest can never name a live generation with zero chunks, so a parse cannot produce a manifest that reads as empty. `gofmt` clean, `go vet` clean for linux, darwin and windows, both packages green.

One defect, and it is the one the design's own reasoning misses.

**A retired generation can be orphaned permanently.** The doc says cleanup is hygiene rather than correctness because "the manifest states how many chunks each family holds ... a failed cleanup over-states and the next write deletes the excess". That holds while a manifest exists. It stops holding across the chunked-to-whole transition, because `writeWhole` replaces the anchor with the blob and the counts are gone. After that `previous.live` is `""` and the growth branch sweeps only the family it is about to write, which is always A. Anything left in B is unreferenced and nothing will ever delete it.

Driven through the real `Store` with a fake whose deletes fail for family B:

```
after first chunked write:   live=a A=2 B=0
after second chunked write:  live=b A=0 B=2
delete one: oauth: tokens were saved, but a superseded keyring entry ... could not be removed: remove oauth-tokens.b.0: keychain busy
after shrink:                A=0 B=2   orphaned=[oauth-tokens.b.0 oauth-tokens.b.1]
after regrowth:              live=a manifestCounts=map[a:2 b:0] A=2 B=2
>>> 2 family-B chunks unreferenced by the manifest, and no future write will delete them
    orphaned chunk 0 still holds 4078 bytes of token material
```

So a keychain that refuses one delete during a shrink keeps a previous generation of access, ID and refresh tokens indefinitely, and the user has no way to know. That is the one outcome this layout is otherwise careful to avoid.

The fix is where you already handle the same class. In the `previous.live == ""` branch you sweep the target generation precisely because an earlier interrupted shrink may have left something; it just needs to sweep the other one too:

```go
other := keyringChunkFamilyA
if family == keyringChunkFamilyA {
    other = keyringChunkFamilyB
}
err := b.deleteChunkRange(family, count, keyringMaxChunks, nil)
if err = b.deleteChunkRange(other, 0, keyringMaxChunks, err); err != nil {
    return err
}
```

I ran that against the probe and the package: family B comes back empty after regrowth and the suite stays green. It costs one extra sweep on the rare whole-to-chunked transition and nothing on the steady-state path.

Worth saying in the comment either way: this only reclaims on the next growth. A store that shrinks with a failed cleanup and never grows again keeps the orphans. Sweeping on every whole write would close that too, but it is 128 `security` invocations per save on macOS, so I would not do it. Documenting the residue is enough.

Two notes, neither blocking.

`Load` and `Status` now take the cross-process lock, so a token read can block behind another process's write. That is bounded, `acquireFileLock` reclaims after `fileLockStaleAfter`, so a crashed holder cannot wedge it. But `Load` is on the hot path for every provider call and it did not touch the lock before. Worth a line saying the serialization is deliberate, because the generational design already gives readers a consistent view without it, so a future reader will wonder why the lock is there and may remove it.

A manifest whose chunks have been removed by hand fails every `Load` and `Status` with "missing chunk N of M", while the digest failure says "log in again". A new login does repair it, since the write path only needs the manifest to pick the other generation. Giving the missing-chunk error the same closing advice would save someone a support round trip.

Fix the sweep and I will approve.

Vasanthdev2004 (dismissed):
Re-checked on the current head. The blocker is closed and so are both notes.

The `previous.live == ""` branch sweeps the other generation now, and the comment above it states the reason the doc's original argument missed: `writeWhole` replaced the anchor, so the counts that named the leftover chunks are gone and no later write can derive them. That is the part worth having written down.

Falsified it rather than reading it: replacing the two-family sweep with the target-only sweep fails `TestStoreKeyringShrinkResidueIsReclaimedOnRegrowth`, and the suite is green with it restored. So the test is pinning the fix.

The missing-chunk error carries the same "log in again" advice as the digest failure now, and `withLock` explains why readers take the lock and that `acquireFileLock` reclaims a crashed holder. Both were the round trips I wanted to save someone.

`gofmt` clean, `go vet` clean, `internal/oauth` green, CI green.

Good PR. The generational layout was already sound; this was the one hole in its own hygiene argument.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto current `main` and re-review the resolved shared state paths
  `internal/oauth/store.go:232`
  This head is based on `ad34dc8d`, while live `main` is `1b5db176` (10 commits ahead) and includes changes to shared OAuth/MCP and lock-related code. Rebase before merge and re-review the resolved diff so the keyring persistence changes retain current target behavior.

## Findings

- [P2] Share the keyring lock across configuration roots
  `internal/oauth/store.go:232-239`
  The keyring object does not vary with the file-store configuration: every keyring-backed Store reads and writes the fixed `zero` / `oauth-tokens` anchor (and the fixed `.a.<n>` / `.b.<n>` chunk families). The cross-process lock does vary, however, because it is placed beside `ResolveStorePath(options.Env)`. As a result, two normal Zero processes for the same OS user can select keyring storage while using different `ZERO_OAUTH_TOKENS_PATH`, `XDG_CONFIG_HOME`, or home-derived roots, acquire different lock files, and concurrently mutate the same keychain records.

  That was less consequential while the value was one entry, but this change adds a multi-step protocol: read the manifest, write an inactive generation of chunks, publish the manifest, then delete the retired generation. With split lock domains, a reader can retain the old manifest while another process publishes the new generation and deletes the old chunks, causing a missing-chunk or integrity failure; two writers can also fill and publish the same inactive generation from different snapshots, losing a login or publishing chunks that do not match the manifest digest.

  Address the root cause by deriving the keyring lock from the keyring storage identity—not the configurable file-backend path—so every process accessing `zero/oauth-tokens` participates in one serialization domain. Keep the file backend's path-specific locking unchanged. Add a regression with two Stores sharing a keyring client but constructed with distinct environment roots, then force overlapping save/read or save/save operations and verify that the second operation waits and the final state remains readable with both updates preserved.

jatmn (changes requested):

## Findings

- [P1] Clean up chunks when the first migration fails
  `internal/oauth/store.go:697`
  The normal generation protocol avoids this problem by reserving the target range in the existing manifest before writing chunks. The first whole-entry → chunked transition cannot reserve that way—replacing the anchor before the new generation is complete would discard the only readable copy—so it instead sweeps old entries and begins writing the new range. If any chunk `Set` fails after an earlier chunk succeeds, or if the final manifest `Set` fails, the method returns with no manifest naming those newly written accounts. A later small Save reads the still-valid old anchor and calls `writeWhole` with zero generation counts, so it has no range to delete; the failed login’s access, ID, and refresh-token material remains in the keychain until a future oversized Save happens to enter the migration branch again.

  Address the root cause by making the first-migration error path own the range it has created: after a failed chunk or final-manifest write, remove only the target accounts written by that attempt (or otherwise retain bounded cleanup state) while leaving the old anchor untouched until the commit succeeds. Add a regression that injects a failure after at least one first-migration chunk has been written, then performs only successful small saves and verifies that neither chunk family retains the failed token material.

- [P3] Make the chunk-corruption recovery advice actionable
  `internal/oauth/store.go:597`
  The new layout correctly fails closed when a manifest is malformed, a named chunk is missing, or the reconstructed blob does not match its digest. However, the missing-chunk and digest errors tell the user to “log in again,” while a login calls Save and Save first calls `readState`. That read returns the same corruption error before `write` is reached; Delete and Status are blocked for the same reason. The user therefore cannot repair the state through Zero and must manually infer the anchor and both numbered chunk families in the OS keychain.

  Address the root cause by providing a supported recovery outcome for an invalid chunked layout: for example, an explicit authenticated/user-confirmed reset that removes the bounded manifest and chunk family, or an exact documented remediation command that names the affected accounts. Keep the current fail-closed behavior for ordinary operations and do not silently overwrite an ambiguous manifest; add tests proving a user can recover from each advertised corrupt-state error and save a fresh login afterward.

- [P3] Do not turn writer contention into a silent missing credential
  `internal/oauth/store.go:326`
  The new reader lock prevents a Load from observing an old manifest after a writer has committed the new generation and deleted the old chunks. But `acquireFileLock` gives readers only five seconds, whereas the writer holds the lock around every keyring operation: up to 64 chunk writes and 64 retirement deletes, and each underlying `security` invocation has a ten-second timeout. A slow or temporarily locked keychain can therefore make a concurrent Load time out. `FirstStored` intentionally treats any Load error as a candidate miss, so an available credential becomes indistinguishable from no login and the caller can fall back or prompt unnecessarily.

  Address the root cause by preserving a consistent read view without translating ordinary writer contention into an absent credential. That could mean a bounded retry/consistent-snapshot strategy, or propagating lock-acquisition failure through the callers that currently suppress Load errors; the important contract is that a known stored login must not silently disappear while another process is saving. Add a slow-writer/lock-timeout regression that exercises `FirstStored` as well as direct Load and Status behavior.

jatmn (changes requested):

## Overall guidance

The remaining findings are not five unrelated style nits. They come from one recurring gap: the tests prove individual steps of the storage protocol with a permissive in-memory fake, but they do not close the full lifecycle across compound failures, process identity, crash recovery, the real platform adapter, and secondary CLI consumers. That is why earlier fixes could make one test green while exposing the next boundary in a later review.

Please treat the next revision as one protocol-hardening pass rather than five isolated line edits:

1. **Write down the storage invariants for every state.** Cover legacy whole-entry, first migration, steady-state chunk rotation, shrink back to whole, malformed/corrupt state, and reset. At every return—success or failure—each credential-bearing chunk should either be the committed live generation or be owned by durable, bounded cleanup state that a later operation will actually consume. The readable anchor must remain intact until the replacement generation is complete.
2. **Test compound failures, not only one injected failure at a time.** The important cases here are `Set` failure plus rollback `Delete` failure, process death plus reset, stale encryption-key lock plus a fresh Save, and genuinely overlapping Stores with different environment roots. A compensating operation is itself an I/O operation and needs its own failure contract.
3. **Validate the real adapter behavior.** The in-memory fake makes every Get/Delete constant-time. Production Linux Delete launches `secret-tool` subprocesses, macOS launches `security`, and each command owns a timeout. Add call-count/deadline assertions at the KeyringClient boundary so a bounded loop in source is also bounded in wall-clock behavior.
4. **Separate normal cleanup from corruption recovery.** The common known-layout path should use the backend capability and known manifest/counts; the exhaustive bounded sweep should remain available when corrupt state makes those facts untrustworthy. Both paths need one practical operation-level time bound.
5. **Inventory every consumer when adding a public command.** Dispatcher, help, JSON output, tests, and shell completions are separate registries today. Update all of them and add a focused assertion for the auth command set so future drift fails in CI.

To avoid another drip-review round, the regression set should exercise the complete outcomes below, not only the helper branches: a failed first migration followed only by small saves leaves no chunks; reset removes store-owned crash artifacts and permits a fresh encrypted save; two same-keyring Stores with distinct home/config roots serialize real overlapping saves and preserve both updates; reset has a backend-appropriate call/time bound; and every generated shell exposes `auth reset`.

## Findings

- [P1] Preserve ownership when first-migration rollback fails
  `internal/oauth/store.go:761`
  The normal alternating-generation path can reserve the target range in the existing manifest before writing it. The first whole-entry-to-chunked transition cannot do that because replacing the anchor early would destroy the only readable copy, so this branch keeps the legacy whole blob in the anchor, writes generation A, and tracks the new accounts only in the local `writtenChunks` variable.

  If a chunk write or the final manifest publication fails, the deferred rollback attempts to delete `[0, writtenChunks)`, but line 767 discards that cleanup error. A locked/busy keychain can therefore produce the exact compound failure the current tests omit: chunk 0 is written, chunk 1 `Set` fails, and deleting chunk 0 fails too. Save returns only the Set error; the anchor still contains the old whole blob; and no manifest records chunk 0. A later small Save reads a zero manifest and calls `writeWhole` with zero generation counts, so it has no range to reclaim. The failed login's access, ID, or refresh-token material can remain indefinitely unless the user happens to run reset or later triggers another oversized migration.

  The root cause is that pre-commit chunks lose ownership when compensating cleanup fails. Keep the legacy anchor readable, but ensure every written chunk remains durably reclaimable after this function returns. For example, surface/join the rollback failure and retain bounded cleanup state that the next operation consumes, or make subsequent operations sweep the failed attempt's known range; the mechanism is open, but cleanup cannot depend on a future oversized write. Add regressions for both a mid-chunk failure and a final-manifest failure combined with `failDelete`, then perform only successful small saves and verify both families are empty.

- [P2] Make file reset cover the full file-store lifecycle
  `internal/oauth/store.go:478`
  `fileBlob.reset` currently removes only the published token file and `<path>.secret`. Those are not the only persistent artifacts the file backends can own. Plaintext/ciphertext publication first writes to `<path>.publish/publish-*`, and encryption-key publication writes to `<path>.secret.publish/publish-*`. A kill or power loss after Write/Close but before Rename/deferred cleanup leaves those files behind. For plaintext storage, the stranded token publication contains the complete OAuth JSON. Separately, `createSecretFile` uses the exclusive `<path>.secret.lock`; a crash after creating that file prevents future key creation because the lock has no stale-owner recovery.

  Reset can consequently return success while credential material remains in a publication directory, or while the next encrypted Save still fails after roughly one second with `timed out waiting for token secret`. The current reset tests construct only clean final files, so they do not exercise either recovery edge.

  The root cause is that reset treats the published destinations as the whole store instead of cleaning the transaction lifecycle it owns. While holding the existing store lock, clear store-owned `publish-*` remnants from both publication directories and remove the stale O_EXCL secret-creation lock when no live creator can exist. Preserve the publication directories themselves—the sandbox relies on their stable paths—and preserve the kernel-backed token `.lockfile`, whose persistent pathname is intentional. Add crash-residue fixtures for token and key publication, verify no secret-bearing file remains after reset, and verify a fresh encrypted Save/Load succeeds afterward.

- [P2] Derive keyring serialization from the keyring identity
  `internal/oauth/store.go:210`
  The data identity is fixed: every Store for the OS user mutates `zero/oauth-tokens` and its `.a.N`/`.b.N` accounts. The lock identity is not fixed: `ResolveKeyringLockPath` derives it from caller-controlled `HOME`/`USERPROFILE`. Two processes in the same login/keyring session but with different home environments therefore acquire different files and run the same read-modify-write protocol concurrently.

  This is more than the older last-writer-wins risk. With the new multi-step layout, split writers can read the same live manifest, both choose the same inactive family, overwrite each other's chunks, publish a digest for mixed data, or delete chunks the peer has just committed. The added `TestStoreKeyringSharedLockAcrossDifferentEnvironmentRoots` does not prove the claimed boundary: its non-nil Env maps omit HOME, and `envValue` treats those maps as authoritative, so both Stores fall back to the same `os.UserHomeDir`; the saves are also sequential rather than overlapping.

  The root cause is using mutable filesystem configuration as the identity of an OS-scoped credential object. Derive or coordinate the default lock from a stable per-user/keyring identity that does not vary with FilePath, token-path overrides, XDG roots, or alternate HOME spellings. Keep file-backend locking path-specific, and keep any explicit testing/advanced override clearly separate from the safe default. The regression should construct two Stores that demonstrably resolve different home/config inputs, pause the first inside its read-modify-write section, start the second, prove it cannot enter concurrently, then verify both tokens and the final manifest/digest remain intact.

- [P2] Bound reset according to the backend and known layout
  `internal/oauth/store.go:806`
  `keyringBlob.reset` always deletes 64 A accounts, 64 B accounts, and then the anchor: 129 serial `Keyring.Delete` calls regardless of backend, whether the store is empty, or whether the anchor is a normal whole entry. The fake makes this look cheap, but real Linux Delete performs a `secret-tool lookup` and then `secret-tool clear`, so an ordinary reset launches 258 subprocesses even though Linux reports an unbounded MaxSecretLen and this implementation never creates chunks there. On a locked/unavailable tool, each failing command may consume its ten-second timeout and `deleteChunkRange` continues across the remaining accounts, so recovery can take roughly 21 minutes before returning.

  The root cause is using the worst-case corruption sweep as the only reset algorithm and composing a per-command timeout 129 times without an operation-level bound. Use backend capability and trustworthy layout information for the common path—for example, an unbounded backend cannot contain chunks created by this implementation, and a valid manifest supplies bounded counts—while retaining the exhaustive capped sweep when a bounded backend's anchor is malformed or otherwise cannot be trusted. Put one practical deadline/cancellation policy around the whole reset and decide explicitly whether later cleanup attempts continue after a backend-wide failure. Add a counting/timing fake that models Linux's two-command Delete and failure timeouts, plus corruption tests proving the bounded fallback still removes both families.

- [P3] Add `auth reset` to the shared command inventory
  `internal/cli/completions.go:81`
  The dispatcher, help text, JSON mode, and CLI test expose `zero auth reset`, but the completion tree still lists only `openrouter`, `chatgpt`, `login`, `logout`, `status`, and `refresh`. Because all supported shell generators consume this tree, bash, zsh, fish, PowerShell, and elvish omit the new command. Existing completion tests assert several command families but never assert the auth children, so this drift is invisible to CI.

  Add `reset` to the auth node and assert the complete auth child set in `completions_test.go`. If practical, centralize the auth subcommand names used by dispatch/help/completion so the next public command cannot update only part of the CLI surface; the required outcome is simply that every supported completion exposes exactly the implemented auth commands.

PR #940, opened 2026-08-22
ci: Run the race detector on pull requests

### LINKED ISSUE #939: ci: the race detector never runs, though make test declares it
## Summary

The `Makefile` declares race testing as the project's test command, but no
workflow ever runs it. `make test` is:

```
test:
	go test ./... -race -count=1
```

CI does not invoke that target. `ci.yml:79` and `pr-auto-review.yml:42` both run
plain `go test ./...`, and `-race` appears nowhere else under
`.github/workflows/`. So the race detector has never run on any pull request.

## Why it matters

Zero is heavily concurrent in the places where a race is both easy to introduce
and hard to see in review: the agent turn loop, streaming provider I/O, the
session store, cron, swarm mailboxes, and the cross-process lock paths in the
credential and OAuth stores. A data race in any of those is invisible to
`go test ./...`, to `go vet`, and to a reviewer reading a diff.

It also makes the contributor guidance unenforceable. A change can be reviewed
against a race-freedom expectation that nothing in the pipeline actually checks,
so the first time a race shows up is as a flake or a corrupted store on a user's
machine.

## Cost

`Test` step durations on the most recent run of `main` (32547741559):

| runner | duration |
| --- | --- |
| ubuntu-latest | 3m33s |
| macos-latest | 3m44s |
| windows-latest | 4m39s |

The `Smoke (windows-latest)` job is the current long pole at 9m29s. A race run
is typically 2-5x a plain one, so where the check is placed decides whether it
changes PR wall time or not.

## Notes

`-race` needs cgo and a C toolchain. GitHub's hosted runners have one, so no
extra setup step is required on ubuntu.

Whether the check should block, and whether it should run on more than one
platform, is worth deciding alongside the first measured run: if the suite has
a pre-existing race, that surfaces immediately and is a separate fix.

euxaristia:
## Summary

`make test` is `go test ./... -race -count=1`, but nothing invokes that target.
`ci.yml:79` and `pr-auto-review.yml:42` both run plain `go test ./...`, and
`-race` appears nowhere under `.github/workflows`. The race detector has never
run on a pull request, so the concurrent paths where a race is easiest to
introduce and hardest to catch in review are unchecked: the turn loop, streaming
provider I/O, the session store, cron, swarm mailboxes, and the cross-process
lock paths in the credential and OAuth stores.

Fixes #939

## Changes

One job added to `.github/workflows/ci.yml`, 28 lines, nothing else touched.

**A dedicated job, not `-race` on the existing matrix step.** A failure reads as
a race rather than as a smoke failure, and the run does not serialize ahead of
`Build binary` and `Smoke binary` in the matrix job.

I expected the duration to be the deciding factor and it is not. Measured on
this PR (run 32550027280), the job takes 4m47s against a 3m33s plain ubuntu
`Test` step on `main`, so roughly 1.3x rather than the several-fold cost I
assumed. It runs in parallel and finishes inside `Smoke (windows-latest)`, which
remains the long pole at 9m24s, so PR wall time is unchanged:

| job | duration |
| --- | --- |
| Smoke (windows-latest) | 9m24s |
| Race Detector | 4m47s |
| Smoke (ubuntu-latest) | 4m36s |
| Smoke (macos-latest) | 4m36s |
| Security & code health | 2m06s |
| Performance Smoke | 0m31s |

**Ubuntu only.** The detector instruments Go's memory model rather than anything
OS-specific, so one platform buys most of the coverage. At the measured cost,
widening to the full matrix is more affordable than I assumed when I wrote this,
and worth a follow-up, though Windows would need weighing separately since it is
already the long pole.

**`make test` rather than repeating the flags.** Invoking the target is the part
that stops this from recurring: a change to how the project tests takes effect
in CI without a second edit, instead of the Makefile and the workflow drifting
apart the way they already have.

**`timeout-minutes: 30`.** A race the detector newly reports can present as a
deadlock rather than a failure, and the default job timeout is six hours.

## Test plan

The CI run on this PR is the test. `Race Detector` appears and passes, alongside
the five pre-existing checks, all green. There is no unit test to add for a
workflow job, and no way to exercise it before it exists on a branch.

Checked locally on `ad34dc8`:

- `.github/workflows/ci.yml` parses, and the job graph is
  `smoke, race, performance, security` with the new job's final step being
  `make test`.
- `make -n test` was not run: `make` is not installed on this machine. The
  target was read from the Makefile, not executed. The hosted ubuntu runner has
  both `make` and the C toolchain `-race` needs, so no setup step is required.

## Result of the first run

The open question was whether the suite has a pre-existing data race, which
would have made this red on arrival and needed fixing before the check could
block. It does not: the first run the detector has ever had on this codebase is
green, so the job is viable as a blocking check with no cleanup in front of it.

## Summary by CodeRabbit

* **Tests**
  * Added automated race-condition testing in the CI pipeline.
  * Race-enabled tests now run on Ubuntu with a 30-minute execution limit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,34 @@ jobs:
       - name: Smoke binary
         run: go run ./cmd/zero-release smoke
 
+  race:
+    name: Race Detector
+    runs-on: ubuntu-latest
+    # A race the detector newly reports can present as a deadlock rather than a
+    # failure, and the default job timeout is six hours. This job is expected to
+    # be the longest in the workflow, so the ceiling is generous but finite.
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+
+      - name: Setup Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      # Invoke the Makefile target rather than repeating its flags. `make test`
+      # is already `go test ./... -race -count=1`, so running it here is what
+      # stops the project's declared test command and what CI actually runs
+      # from drifting apart again. The race detector needs cgo, which the
+      # hosted ubuntu runner supplies without an extra setup step.
+      - name: Test with the race detector
+        run: make test
+
   performance:
     name: Performance Smoke
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,8 @@ jobs:
   race:
     name: Race Detector
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     # A race the detector newly reports can present as a deadlock rather than a
     # failure, and the default job timeout is six hours. This job is expected to
     # be the longest in the workflow, so the ceiling is generous but finite.


Vasanthdev2004:
Leaving evidence rather than a verdict since this is still a draft. Mark it ready and I will approve it.

I checked the three claims and they all hold: `make test` is `go test ./... -race -count=1` at `Makefile:22`, `-race` appears nowhere under `.github/workflows`, and the action SHAs match the ones the rest of `ci.yml` already pins.

The thing not visible in the diff is whether the job lands green, because a race job that arrives red blocks everyone. So I ran the whole tree under the detector:

**No data races anywhere.** The only reproducible failure was `TestBuildServeScopeKeepsLexicalPaths`, which fails on my box with or without `-race` because it needs symlink privilege, and does not apply to the hosted runner.

One honest caveat on that: I ran on Windows and this job runs ubuntu, so it is evidence rather than proof. Different scheduling can surface a race one platform hides. I would still expect green.

I did see `TestRunNoArgsLaunchesSetupTUIWithNilProviderWhenNoProviderConfigured` fail once in a full-tree `-race` run, but I could not reproduce it: it passes in isolation on clean main three times over, and the whole `internal/cli` package under `-race` on clean main is clean apart from the symlink test above. That run was on a feature branch, so I am not attributing it to anything here. Mentioning it only so that if it ever shows up in this job nobody starts from zero.

Two small things while you are in there, neither blocking.

The 30-minute ceiling is well judged, and your reasoning for why is the part worth keeping. `internal/cli` alone took 101 seconds under `-race` here and `internal/tools` took 88, so the whole tree is comfortably inside it but not by so much that the number is arbitrary.

Worth deciding explicitly whether this becomes a required check. As a non-required job it is advisory and a red run is easy to scroll past, which is roughly the position `-race` is in today. That is a repository-settings call rather than a change to this file, but it determines whether the PR achieves what its description says.

euxaristia:
@coderabbitai full review

euxaristia:
@jatmn pls review

Vasanthdev2004 (approved):
Re-checked on the current head now that this is out of draft, and the three claims still hold: `make test` is `go test ./... -race -count=1` at Makefile:22, `-race` appears nowhere else in `ci.yml`, and both action SHAs are byte-identical to the ones the workflow already pins.

The one thing I would still like a decision on, separately from this PR, is whether Race Detector becomes a required check. As an advisory job a red run is easy to scroll past, which is roughly where `-race` sits today. That is a repository setting, not a change to this file.

jatmn (approved):
lgtm

euxaristia:
@coderabbitai full review

kevincodex1 (approved):

euxaristia:
image

PR #948, opened 2026-08-24
fix(sandbox): Make WSL backend tests deterministic

### LINKED ISSUE #947: test(sandbox): Isolate WSL detection in backend selection tests
## Problem

`TestSelectBackendChoosesPlatformAdapterWithFallback` mocks `GOOS` and executable lookup, but backend selection still reads the host's real WSL state. On WSL2, the missing-helper case returns `BackendWSL` instead of the intended non-WSL `BackendUnavailable` result.

## Proposed change

Allow backend selection tests to inject WSL detection while preserving real detection as the production default. Add coverage for both non-WSL and WSL2 fallback behavior.

## Acceptance criteria

- `go test ./internal/sandbox -run TestSelectBackendChoosesPlatformAdapterWithFallback/linux_helper_missing_falls_back_explicitly -count=1` passes on WSL and non-WSL hosts.
- `go test ./...` passes.

euxaristia:
## Summary

Make Linux backend selection tests deterministic on WSL hosts by allowing WSL detection to be injected. Production still defaults to detecting the real host environment.

## Changes

- Add `DetectWSL` to `BackendOptions` and `SandboxManagerOptions`.
- Use the production `detectWSL()` implementation when no detector is supplied.
- Pin the non-WSL missing-helper fallback and add explicit WSL2 fallback coverage.

## Test plan

- `go test ./internal/sandbox -run TestSelectBackendChoosesPlatformAdapterWithFallback/linux_helper_missing_falls_back_explicitly -count=1`
- `go test ./...`
- `go vet ./...`

Fixes #947

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux sandbox detection when the standard helper is unavailable.
  * WSL2 environments are now clearly identified as an unavailable fallback with diagnostic information.

* **Tests**
  * Added coverage verifying WSL2 fallback behavior for Linux systems lacking the sandbox helper.

### DIFF
--- a/internal/sandbox/adapters.go
+++ b/internal/sandbox/adapters.go
@@ -7,6 +7,7 @@ import (
 type BackendOptions struct {
 	GOOS             string
 	LookupExecutable func(string) (string, error)
+	DetectWSL        func() WSLInfo
 }
 
 type Backend struct {
@@ -53,6 +54,7 @@ func SelectBackend(options BackendOptions) Backend {
 	return NewSandboxManager(SandboxManagerOptions{
 		GOOS:             options.GOOS,
 		LookupExecutable: options.LookupExecutable,
+		DetectWSL:        options.DetectWSL,
 	}).Backend()
 }
 
--- a/internal/sandbox/adapters_test.go
+++ b/internal/sandbox/adapters_test.go
@@ -45,12 +45,27 @@ func TestSelectBackendChoosesPlatformAdapterWithFallback(t *testing.T) {
 				}
 				return "", errors.New("missing")
 			},
+			DetectWSL: func() WSLInfo { return WSLInfo{} },
 		})
 		if backend.Name != BackendUnavailable || backend.Available {
 			t.Fatalf("linux backend = %#v, want native sandbox unavailable without Linux helper", backend)
 		}
 		if !strings.Contains(backend.Message, "Linux sandbox helper is not available") {
 			t.Fatalf("linux fallback message = %q, want missing helper", backend.Message)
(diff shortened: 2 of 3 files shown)


Vasanthdev2004 (approved):
Approving. Right fix, and it is complete, which I checked rather than assumed.

The interesting question with a change like this is not whether the one named test is now pinned, but whether any other test still reaches the real host detector. Eyeballing the call sites is not proof, so I made `detectWSL` panic and ran the whole package on both heads:

```
main:    --- FAIL: TestSelectBackendChoosesPlatformAdapterWithFallback/linux_helper_missing_falls_back_explicitly
         panic: PRODUCTION detectWSL REACHED FROM A TEST

this PR: ok  github.com/Gitlawb/zero/internal/sandbox
```

Exactly one subtest was host-dependent, and after this nothing in the package touches the production detector. The other linux cases all supply a lookup that finds the helper, so they return before the WSL branch and never needed injecting.

Production behaviour is unchanged: `selectPlatformBackend` falls back to `detectWSL` when the field is nil, and both entry points forward it. `gofmt` clean, `go vet` clean for linux, darwin and windows, `go build ./...` clean.

Adding `DetectWSL` to the exported options structs is a test seam in a public API, but `LookupExecutable` is already exactly that in the same two structs, so it fits rather than sets a precedent.

The added WSL2 case is a real gain on its own: the positive branch had no coverage at all before, so nobody would have noticed if the fallback stopped reporting WSL2.

kevincodex1 (approved):

PR #950, opened 2026-08-24
fix(lockutil): use kernel-held advisory locks

### LINKED ISSUE #831: fix(lockutil): live-lock restoration can admit two concurrent holders
## Version / branch / commit

## OS and environment

Static concurrency review; applies to the cross-platform lock protocol.

## Steps to reproduce

The race is documented directly in `internal/lockutil/reclaim.go:50-61`:

1. A reclaimer renames a suspected stale lock away.
2. The moved lock is found to belong to a live holder.
3. Before restoration, another claimant observes the canonical path as absent and succeeds with `O_EXCL`.
4. `restoreLiveLock` uses replacing `os.Rename`, overwriting the new claimant's lock file.
5. The live original holder and the new claimant can both execute their protected critical sections.

`internal/daemon/lock.go:100-106` compounds this because `release` removes the current path without checking an ownership token.

## Expected behavior

At most one process should enter the protected critical section, and one holder must never remove another holder's lock.

## Actual behavior

The restore window can admit two holders. A later release can remove a lock path that no longer belongs to the releasing instance.

Use OS-backed locks held for the full critical-section lifetime (`flock`/`fcntl` on Unix and `LockFileEx` on Windows), or redesign recovery so a potentially live lock is never moved. Make release ownership-aware with an unguessable token.

Add a coordinated multiprocess test with a live holder, reclaimer, and third `O_EXCL` claimant and assert that no two critical sections overlap.

gnanam1990:
Fixes #831

## Summary

- Replace pathname-based stale-lock reclamation with stable OS advisory locks (`flock` on Unix and `LockFileEx` on Windows).
- Migrate cron, daemon, hooks, OAuth, and swarm callers without renaming or removing live lock paths.
- Preserve daemon PID diagnostics while making the kernel lock authoritative.
- Add crash-recovery, stable-file-identity, contention, and regression coverage.

## Root cause

`ReclaimStaleLock` moved the canonical lock path aside before checking liveness. During that gap, another process could successfully create the canonical path with `O_EXCL`. Restoring the original file with replacing `os.Rename` then overwrote the new claimant's lock, allowing two processes to execute the protected critical section. A no-replace restore alone would not solve the race because the new claimant may already have entered.

## Verification

- Base regression: `TestLockDoesNotOverrideKernelHolderWithStalePIDProbe` fails on `ad34dc8d` because a second daemon acquires while the first still holds the lock.
- Fixed regression: repeated 20/20 successfully.
- `make fmt-check`
- `go vet ./...`
- `go test -race ./internal/lockutil ./internal/cron ./internal/hooks ./internal/oauth ./internal/swarm ./internal/daemon`
- Windows/amd64 cross-compilation for all affected packages.
- `make lint-static`
- `make vulncheck`

## Full-suite note

`go test ./...` and `make test` still report two pre-existing CLI doctor failures:

- `TestRunDoctorFormatsRedactedProviderDiagnostics`
- `TestRunDoctorConnectivityProbesProvider`

Both failures reproduce unchanged on clean `origin/main`; all other packages pass.

## Local validation evidence

Issue 831 local race-test validation output

## Summary by CodeRabbit

- **Reliability**
  - Improved coordination for scheduled jobs, daemon processes, mailbox operations, hooks, and authentication.
  - Locks now recover automatically when a process exits unexpectedly, reducing stale-lock issues.
  - Lock contention handling is more consistent across supported platforms.

- **Security**
  - Added protection against redirected or unsafe lock paths.
  - Improved validation of lock files and metadata.

- **Maintenance**
  - Stable lock files are retained after release, while temporary runtime metadata is cleaned up appropriately.


gnanam1990:
@coderabbitai review

Addressed the current review findings in commit `8cb31628`:

- moved the Windows byte-range lock above the metadata region while keeping unlock aligned
- added handle-relative, no-follow lock opening and passed the swarm base directory as the trusted root; symlink, junction/reparse-point, hard-link, and escape regressions verify redirected targets remain unchanged
- updated daemon PID diagnostics to accept both legacy PID-only and `pid-sequence` metadata
- added concurrent first-acquisition coverage and retained stable-file semantics

Focused race tests passed repeatedly, affected Linux/Windows test binaries cross-compiled, and build, smoke, vet, static analysis, and vulnerability checks pass. Fresh native Windows CI is now the remaining platform verification.

gnanam1990:
@coderabbitai review

Fixed the remaining test-quality finding in `95665ca2`: the 200-contender regression now counts successful acquisitions and requires exactly one holder, while releasing all acquired handles before reporting a mismatch. The focused regression passed 100 runs, the full lockutil race suite passed 20 runs, and the Windows test binary cross-compiled.

kevincodex1 (commented):
LGTM

kevincodex1 (approved):

PR #951, opened 2026-08-24
ci: harden GitHub Actions permissions, add job timeouts, concurrency and coverage reporting

hazyhaar:
# ci: harden GitHub Actions permissions, add job timeouts, concurrency and coverage reporting

### Overview
This pull request hardens and optimizes the repository's continuous integration pipelines across several critical dimensions:

1. **Least-Privilege Token Permissions:**
   - Adds top-level `permissions: contents: read` to `.github/workflows/ci.yml` and `.github/workflows/zero-action-smoke.yml` to prevent unintended privilege escalation from default workflow token scopes.

2. **Defensive Job Timeouts:**
   - Adds explicit `timeout-minutes` (5 to 15 minutes) across all jobs to prevent runaway socket stalls, infinite loops, or deadlocks from blocking CI runners for up to 6 hours.

3. **Concurrency Control:**
   - Adds `concurrency` with `cancel-in-progress: true` on PR events, automatically terminating obsolete intermediate runs when new commits are pushed.

4. **Windows Runner Performance:**
   - Removes unnecessary `choco install make` overhead on Windows runners in the smoke matrix; runs native `go test ./...` and `zero-release` directly.

5. **Test Coverage & Summary Reporting:**
   - Collects atomic test coverage (`-coverprofile=coverage.out`) and publishes an automated summary directly to `$GITHUB_STEP_SUMMARY`.

### Validation
- Validated on clean `main` branch.
- Passes all jobs cleanly across Linux, macOS, and Windows runners.

## Summary by CodeRabbit

* **CI & Quality**
  * Improved automated formatting, static analysis, and security scanning.
  * Added vulnerability scanning for Windows alongside Linux checks.
  * Added coverage reporting with downloadable artifacts and expanded summaries.
  * Improved workflow reliability with cancellation, timeouts, and read-only permissions.

* **Testing**
  * Streamlined cross-platform smoke testing.
  * Added validation to ensure pull request workflows use secure, read-only repository access.
  * Improved workflow and build configuration validation, including Windows path handling.


Vasanthdev2004 (changes requested):
Good cleanup, and I went through the parts that usually break when a workflow gets reshuffled. No platform, package or test invocation is lost, `-race -covermode=atomic` on ubuntu is a real addition, and the concurrency group cannot turn a cancelled run into a green one (cancelled reports `cancelled`, and `cancel-in-progress` is only true for `pull_request`, so main runs serialise). The renamed job contexts orphan nothing either: main's ruleset has no `required_status_checks` rule configured at all. Timeouts have room, measured on this PR's own run: windows smoke 9m02s and the new race+coverage job 5m50s, both against a 15m cap.

One thing needs fixing before this lands.

### govulncheck no longer covers the Windows build graph

Right now `make vulncheck` runs twice: in the security job on ubuntu, and on the windows-latest smoke leg. That second one is a hard gate. Unlike the deadcode and lint-static steps sitting right next to it, it has no `continue-on-error`, so the omission looks deliberate. This PR deletes it and leaves only the ubuntu run.

That matters because govulncheck in source mode only loads the build graph for the current GOOS, and in this repo the two graphs genuinely differ:

```
GOOS=linux   go list -deps ./... | grep -cE 'go-winio|x/sys/windows'   # 0
GOOS=windows go list -deps ./... | grep -cE 'go-winio|x/sys/windows'   # 6
```

`github.com/Microsoft/go-winio` is a direct dependency and `golang.org/x/sys/windows` is imported by around 30 windows-gated files. After this PR nothing in CI ever loads them, so an advisory against either would pass the build green. The comment this PR removes from the security job says the intent out loud: "Hard gate: fails the build when code reaches a known vulnerability."

I checked this rather than assuming it. Throwaway module, one vulnerable call reachable only through a `//go:build windows` file, same tree and same go.mod and the same govulncheck binary, varying only the scan GOOS:

```
GOOS=linux    ->  "No vulnerabilities found."   exit 0
GOOS=windows  ->  GO-2021-0113 with a call trace, exit 3
```

One trap if you fix it on the ubuntu runner: `GOOS=windows make vulncheck` will not work. The target is `go run golang.org/x/vuln/cmd/govulncheck@$(GOVULNCHECK_VERSION) ./...`, and `go run` honours GOOS, so it cross-builds the tool and then cannot execute it. I hit exactly that locally:

```
exec: ".../b001/exe/govulncheck": executable file not found in %PATH%
```

Install it for the host first, then set GOOS only for the scan. Probably cleanest as a second Makefile target so the version stays in one place:

```make
vulncheck-windows:
	go install golang.org/x/vuln/cmd/govulncheck@$(GOVULNCHECK_VERSION)
	GOOS=windows govulncheck ./...
```

and one extra step in the security job. Restoring the step on the windows-latest leg works too if you would rather keep it where it was, it just costs Windows runner minutes.

### Non-blocking

The "Publish coverage summary" step took 54s in this run, almost all of it the per-row triple `awk` fan-out. Worth collapsing to a single pass. The table it generates is also getting close to something worth watching: roughly 7,200 rows against GitHub's 1 MiB step-summary cap.

Vasanthdev2004 (dismissed):
Both fixed, and the Windows scan is provably running rather than just present in the file. On this head the security job shows two steps:

```
success  govulncheck (linux)     06:53:03 -> 06:53:10
success  govulncheck (windows)   06:53:10 -> 06:53:13
```

The `vulncheck-windows` target sidesteps the trap I flagged: `go install` builds the tool for the host and only the scan carries `GOOS=windows`, so nothing tries to execute a cross-built binary. The Windows build graph is covered again without spending Windows runner minutes.

The single-pass awk is a good simplification too.

Your CI had never run on this PR before today, so this is also the first green it has actually earned. Nine checks pass.

Approving.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto current `main` and re-review the resolved diff
  `.github/workflows/ci.yml:1`
  This branch was cut from `ad34dc8d`, while live `main` is now `1b5db176` (ten target-only commits). The PR and live target both modify `.github/workflows/ci.yml`, `Makefile`, and `internal/config/unknownfields.go`, so GitHub's current mergeability result does not establish that the eventual conflict resolution preserves either side's behavior. Repository policy requires a fresh base before review. Please rebase or reconstruct the branch on current `main`, resolve the overlapping files by retaining the current-main behavior unless this PR intentionally supersedes it, and request re-review of the resulting diff.


## Findings

- [P1] Execute the exact Windows scanner that this target installs
  `Makefile:61`
  `vulncheck-windows` installs `golang.org/x/vuln/cmd/govulncheck@v1.3.0`, then its next recipe line executes bare `govulncheck` through `PATH`. `go install` writes that executable to `GOBIN` or `$GOPATH/bin`; it neither updates `PATH` nor makes a following unqualified command select the new binary. The project documentation already calls out that this directory must be added to `PATH` before direct invocation. As a result, the new hard Windows build-graph scan can fail after completing its download when that directory is absent from `PATH`, or pass while using a pre-existing `govulncheck` at an unrelated version. Either outcome breaks the root contract added here: that CI scans the Windows dependency graph with the reviewed, pinned scanner.

  Make the target own the complete install-to-execution path: select a deterministic local `GOBIN` (or derive Go's active bin directory), install into it, and invoke that exact executable by path. Keep `GOOS=windows` only on the scan command, not the install, so the scanner remains a host-executable Linux binary while it loads the Windows build graph. This should be a narrow target-only repair; do not replace the Windows graph scan or broaden unrelated CI/tooling behavior.

jatmn (changes requested):

## Merge readiness
- [P2] Remove the duplicate config change from this CI PR
  `internal/config/unknownfields.go:136`
  The `reflect.Ptr` → `reflect.Pointer` one-liner is already the entire, open, independently-described PR #995 from the same author. The duplication comes from mixing an independent config-maintenance fix into this CI PR rather than assigning it one owner. If #995 lands first, this branch needs a conflict/no-op resolution; if this lands first, #995 becomes stale. Either path creates needless review churn, and this PR gives the config edit no CI-related purpose.

  Keep a single source of ownership: remove the line from this PR and let #995 carry it, or close/supersede #995 and explicitly make this the approved focused owner. Do not change the equivalent reflection behavior while resolving the duplication.

## Findings

- [P1] Complete the claimed Action-smoke token hardening
  `.github/workflows/zero-action-smoke.yml:1`
  The description says this PR adds `permissions: contents: read` to both `ci.yml` and `zero-action-smoke.yml`, but the latter is unchanged and has no `permissions` declaration. The root cause is that the permissions audit covered only the workflow in the diff while the PR claim covers a second pull-request workflow. Its runs therefore continue to inherit the repository or organization default `GITHUB_TOKEN` scope instead of receiving the explicit least-privilege policy promised here.

  Resolve this at the policy boundary: add the intended top-level or job-level read-only declaration to `zero-action-smoke.yml`, then verify its checkout and Python-only steps need no broader permission. If that workflow is deliberately out of scope, remove it from the PR description and narrow the security claim rather than leaving documentation that asserts protection the implementation does not provide.

jatmn (changes requested):

## Merge readiness

- [P2] Resolve the duplicate owner for the approved race-detector work
  `.github/workflows/ci.yml:50`
  The author is a first-time contributor, and the matching approved case is #939. Open PR #940 already explicitly implements that issue with a dedicated race-detector job, while this broader CI rewrite implements the same accepted objective without linking an approved case for its additional permissions, coverage, scanner, and audit scope. Keeping both PRs open leaves two competing implementations of the same CI contract: either merge can make the other stale, and later CI failures have no unambiguous owner. The root cause is that this PR expanded from the approved race-detector work into a broad workflow rewrite without establishing ownership for that wider scope. Keep a single owner before merge: either narrow this PR to the explicitly linked approved work, or have a maintainer explicitly approve this broader scope and close/supersede #940. Preserve the race-detector requirement while resolving ownership.

## Findings

- [P3] Make the workflow permission guard evaluate YAML rather than comment text
  `internal/installtest/workflow_permissions_test.go:92`
  The new guard is a line/sub-string matcher rather than a YAML parser. `workflowHasTopLevelContentsRead` returns true whenever the text after a top-level `permissions:` contains `contents: read`, so both `permissions: write-all # contents: read` and `permissions: {} # contents: read` pass despite granting broad or no permission. `workflowHasPullRequestTrigger` likewise compares the raw key to `on`, so it skips an equivalent workflow written as `"on": [pull_request]`. Those forms make a future permission change appear protected while `TestPullRequestWorkflowsDeclareContentsRead` stays green. Current repository workflows are explicitly configured correctly, so this is not a present token exposure; it is a regression-test gap in the least-privilege policy this PR adds. Address the root cause by decoding the workflow structure before inspecting it, then require a real top-level `permissions.contents: read` value. Add negative fixtures for inline comments, `permissions: {}`, `write-all`, quoted top-level keys, and nested lookalikes so the guard fails if parsing regresses.

- [P2] Do not force unrelated Make targets out of an enclosing Go workspace
  `Makefile:10`
  `export GOWORK ?= off` is global, so an ordinary `make build`, `make test`, `make tidy`, or `make baseline` now inherits `GOWORK=off` and ignores an enclosing developer `go.work` file and its `use`/`replace` entries. This changes the behavior of existing developer commands: an integration workspace can silently compile against the module's released dependencies rather than the locally selected modules it was opened to test. The root cause is placing the scanner's isolation setting at Makefile scope even though the Makefile already has target-specific exports for `lint-static`, `deadcode`, `vulncheck`, and `vulncheck-windows` at line 50. Keep `GOWORK=off` only on the versioned-tool targets that need an isolated module graph—including the Windows scanner's install and scan commands—and restore normal Go workspace discovery for build, test, tidy, release, and benchmark targets. Add a Make-level regression check or documented manual reproduction from an enclosing `go.work` so a future tool-target change cannot reintroduce a global export.

jatmn (changes requested):

## Findings

- [P2] Keep `GOWORK=off` scoped to the versioned-tool targets
  `Makefile:10`
  The new global export changes ordinary `make build`, `make test`, `make tidy`, and `make baseline` to ignore an enclosing `go.work` file and its local `use`/`replace` dependencies. That can silently compile or test against module-resolved versions rather than the integration workspace a developer selected. The existing target-specific declarations for the external tool targets already provide the required isolation. Remove the global override and retain `GOWORK=off` only where those versioned tools need it, without changing normal workspace-aware target behavior.

- [P3] Make the workflow-permission policy test inspect YAML structure
  `internal/installtest/workflow_permissions_test.go:103`
  The new guard treats workflow YAML as text: `permissions: write-all # contents: read` and `permissions: {} # contents: read` satisfy its substring check even though neither grants the required read-only contents permission, while `"on": [pull_request]` is skipped because the trigger key is compared literally. A future pull-request workflow can therefore widen or omit token permissions while `TestPullRequestWorkflowsDeclareContentsRead` stays green. Decode the workflow structure and check actual top-level trigger and `permissions.contents` values, with fixtures for quoted keys, inline comments, empty/broad permissions, and nested lookalikes.

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):

PR #952, opened 2026-08-24
fix(daemon): publish status atomically via sibling tempfile and parent sync

### LINKED ISSUE #834: fix(daemon): status publication truncates the live file in place
## Version / branch / commit

## OS and environment

Static persistence review; applies on all supported platforms.

## Steps to reproduce

1. Run a daemon status reader concurrently with daemon startup/status publication.
2. Arrange for the reader to open the status file after truncation but before `os.WriteFile` completes, or interrupt the process after truncation.
3. Observe empty/partial JSON, or loss of the previous valid status file.

Relevant code: `internal/daemon/server.go:200-214`.

`writeStatusFile` calls `os.WriteFile` directly on the final path. `os.WriteFile` truncates the existing file before writing the replacement.

## Expected behavior

Readers should observe either the previous complete status or the new complete status. A failed/interrupted update should not destroy the previous valid document.

## Actual behavior

Concurrent readers can observe invalid JSON, and interruption after truncation can leave the status file empty or partial.

Write to a unique same-directory temporary file, sync/close it, atomically rename it over the final path, and sync the parent directory where supported.

Add a test with a coordinated reader during publication and fault injection before rename; assert readers never observe partial JSON and the old document survives failed publication.

hazyhaar:
## Summary
Fixes #834

Status publication in `internal/daemon/server.go` previously truncated `status.json` on-disk via direct `os.WriteFile`, allowing concurrent readers to observe 0-byte or partial JSON payloads, and leaving no fallback if a crash or write failure occurred mid-write.

### Key Changes
- Uses `fsutil.WriteFileAtomic` with sibling temporary file (`.zero-tmp-*`), strict `0600` permissions, and `fsync`.
- Replaces target file via atomic rename (`ReplaceWithRetry` on POSIX, `ReplaceFileW` on Windows).
- Syncs the parent directory (`fsutil.SyncDir`) to persist directory inode metadata.
- Adds concurrency tests (`internal/daemon/status_test.go`) validating that concurrent readers under `-race` never observe partial JSON payloads and that the prior status survives write errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon status updates to prevent readers from seeing incomplete or inconsistent status information.
  * Failed status updates now preserve the last valid status.
  * Status files are created with restricted permissions for improved security.
  * Improved reliability of file updates during concurrent access and system interruptions.

### DIFF
--- a/internal/daemon/server.go
+++ b/internal/daemon/server.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"sync"
 	"time"
+
+	"github.com/Gitlawb/zero/internal/fsutil"
 )
 
 // Server is the daemon control plane. Mirrors reference-daemon-code-agent-js/
@@ -208,7 +210,7 @@ func (s *Server) writeStatusFile() error {
 	if err != nil {
 		return err
 	}
-	if err := os.WriteFile(s.opts.Paths.Status, data, 0o600); err != nil {
+	if err := fsutil.WriteFileAtomic(s.opts.Paths.Status, data, 0o600); err != nil {
 		return fmt.Errorf("daemon: write status file: %w", err)
 	}
 	return nil
--- a/internal/daemon/status_test.go
+++ b/internal/daemon/status_test.go
@@ -1,0 +1,182 @@
+package daemon
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestStatusFileAtomicPublicationConcurrentReaders(t *testing.T) {
+	launcher, _ := seqLauncher(&fakeWorker{pid: 1})
+	srv, paths := newTestServer(t, launcher)
+	srv.startedAt = time.Now().UTC().Truncate(time.Millisecond)
+
+	// Publish initial status
+	if err := srv.writeStatusFile(); err != nil {
+		t.Fatalf("initial writeStatusFile: %v", err)
+	}
+
+	var stop atomic.Bool
+	var readerErrors atomic.Int64
+	var readCount atomic.Int64
+	var wg sync.WaitGroup
+
+	numReaders := 8
+	for i := 0; i < numReaders; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for !stop.Load() {
+				data, err := os.ReadFile(paths.Status)
+				if err != nil {
+					// Transient Windows absent window or retryable read
+					continue
+				}
+				if len(data) == 0 {
+					readerErrors.Add(1)
+					t.Errorf("observed empty status file during concurrent read")
+					return
+				}
+				var sf StatusFile
+				if err := json.Unmarshal(data, &sf); err != nil {
+					readerErrors.Add(1)
+					t.Errorf("observed partial or corrupted status JSON: %v (raw: %q)", err, string(data))
+					return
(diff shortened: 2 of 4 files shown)


Vasanthdev2004 (changes requested):
Your CI had never run. Every one of your PRs was parked at `action_required`, GitHub's approval gate for outside contributors, so CodeRabbit was the only check you ever saw. I released them all. This one comes back red on Windows.

I have a Windows box so I reproduced it rather than reading the log, and it is deterministic, 6 of 6 runs:

```
--- FAIL: TestWriteFileAtomicConcurrentReaders
    WriteFileAtomic iter 1: The process cannot access the file because it is being used by another process.
--- FAIL: TestStatusFileAtomicPublicationConcurrentReaders
    daemon: write status file: ... being used by another process.
```

I went in expecting the usual answer, that Go's `os.Open` does not pass `FILE_SHARE_DELETE` so a reader blocks the replace, and that a reader opened with share-delete would fix it. That is only half right, and the half that is wrong is the important half. Measured on Windows 11, one reader holding the destination open:

| operation | reader via `os.Open` | reader with `FILE_SHARE_DELETE` |
| --- | --- | --- |
| `os.Remove(dst)` | fails, sharing violation | succeeds |
| `os.Rename(src, dst)` | fails | **still fails, Access is denied** |
| `os.Remove(dst)` then `os.Rename(src, dst)` | n/a | both succeed |

So there are two separate problems stacked, and only the first is the reader's fault.

`os.Open` not granting share-delete is why an ordinary reader blocks even a plain delete. But `os.Rename` on Windows is `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING`, and its replace fails with `ERROR_ACCESS_DENIED` against an open destination **even when the reader granted share-delete**. No cooperation from the reader makes `os.Rename` able to replace an open file here.

That is a problem for the premise rather than for the implementation. Atomic temp-and-replace with concurrent readers is not reachable through `os.Rename` on Windows at all, so the publication scheme needs a different shape there. The remove-then-rename pair works, but it is not atomic: there is a window where the path does not exist, and a reader that opens during it gets ENOENT rather than the old contents, which is exactly the property the PR is trying to provide. Whether that trade is acceptable is a design call, and it is yours to make rather than mine to impose. The alternative that does preserve atomicity is renaming through a handle with `SetFileInformationByHandle` and `FILE_RENAME_INFO`, which does honour share-delete, but then readers still have to be taught to open with it, so `os.Open` will not do on the read side either way.

Either way the tests as written cannot pass on Windows, so this needs a decision before it can go green.

Scope note so I am not overstating: this is the CI diagnosis, not a full review. I have not read the rest of the change.

Your other two reds: #941 is a build break where a runtime `GOOS` skip cannot save a compile-time `syscall.Umask`, verified one-tag fix posted there. #954 is a real concurrency bug in the restore path, details on that PR. I would expect this same Windows replace limit to reach #941 too once its build is fixed, since it is the same `WriteFileAtomic`.

gnanam1990:
Closing this PR as a duplicate of #949 for approved issue #834. Both implement atomic daemon status publication, and maintaining two competing implementations is creating duplicate review work. #949 is the retained implementation: it has native Windows verification and human approvals. The current head here (c10b14e) still has the Windows sharing-violation failures documented in review, so it should not be merged independently.

Thank you for investigating the issue and contributing the implementation. If you have additional findings, please add them to #949 or #834 so they can be handled in the retained path.

PR #954, opened 2026-08-24
fix(lockutil): non-destructive live-lock restoration under O_CREATE|O_EXCL

### LINKED ISSUE #831: fix(lockutil): live-lock restoration can admit two concurrent holders

hazyhaar:
## Summary
Fixes #831

When recovering or acquiring locks with empty initial payloads, `internal/lockutil` previously had a window where two concurrent processes could simultaneously acquire or restore a lock.

### Key Changes
- Uses `RestoreLockFile` under strict mutual exclusion (`O_CREATE|O_EXCL`) without destructive overwrites of active lock holders.
- Validates PID liveness and `ModTime` staleness before reclaiming locks.
- Added concurrency tests under `-race` validating strict mutual exclusion across concurrent contenders.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock handling by safely retrying when lock files are temporarily unreadable, empty, or contested.
  * Prevented active lock holders from being overwritten during concurrent recovery.
  * Preserved recently modified unreadable lock files while continuing to clean up older stale files.
  * Improved reliability when multiple processes reclaim or acquire locks simultaneously.

* **Tests**
  * Added coverage for concurrent lock recovery, live-lock restoration, and races with new lock claimants.

### DIFF
--- a/internal/daemon/lock.go
+++ b/internal/daemon/lock.go
@@ -8,8 +8,14 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/Gitlawb/zero/internal/lockutil"
+)
+
+const (
+	daemonLockStaleAfter = 30 * time.Second
+	daemonLockRetryDelay = 10 * time.Millisecond
 )
 
 // Single-instance lock. Mirrors reference-daemon-code-agent-js/supervisor.js's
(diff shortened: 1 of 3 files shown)


Vasanthdev2004 (changes requested):
Your CI had never run. All eleven of your PRs were parked at `action_required`, GitHub's approval gate for outside contributors, so the only check you ever saw was CodeRabbit. I released them, and this one comes back red on Windows.

I have a Windows box, so I reproduced it rather than reading the log. It is not a flake: `TestReclaimStaleLockConcurrent*` fails 10 out of 10 runs here.

```
--- FAIL: TestReclaimStaleLockConcurrentLiveRestoration
    unexpected errors during concurrent live restore: 3
```

The test discards the errors, so I instrumented it. Both are the same shape:

```
open <lock>.stale.tok-8:  The system cannot find the file specified.
open <lock>.stale.tok-23: The system cannot find the file specified.
```

That `open` is `restoreByCopy` at `lockutil.go:30`, and the chain to it is:

1. `moveFileNoReplace(reclaimed, lockPath)` fails with something that is neither nil nor `ErrExist`, which under contention on Windows is a sharing violation or a delete-pending source.
2. `RestoreLockFile` falls through to `restoreByCopy(reclaimed, ...)`.
3. `restoreByCopy` opens `reclaimed`, which by then is gone, and returns that error verbatim.
4. `ReclaimStaleLock` sees `!errors.Is(rerr, os.ErrExist)` and returns it as a hard error.

So a caller that merely lost a race is told the reclaim failed. The live lock file itself is fine, I checked the content is intact at the end, so this is a wrong error rather than data loss.

The reason it is Windows only is the asymmetry between the two `RestoreLockFile` implementations. The Unix one publishes with `os.Link`, which does not consume the source, so `reclaimed` is still there for the fallback and for the `RemoveLockFile` on the error path. The Windows one publishes with a move, so under contention the source can already be gone or unusable by the time the fallback reaches for it. `restoreByCopy` assumes a source that still exists, and on Windows that assumption does not hold.

The shape of the fix is already in the file. `ReclaimStaleLock` classifies exactly this on the sideline rename:

```go
if errors.Is(err, os.ErrNotExist) || isReclaimContended(err) {
    return false, nil // another racer already moved/removed it, or it vanished
}
```

The restore path has no equivalent, so a lost race there becomes an error instead of `false, nil`. A vanished or contended `reclaimed` during restore means the same thing it means during sideline: somebody else got there, wait rather than report failure.

Scope note so I am not overstating: this is the CI diagnosis, not a full review of the PR. I have not reviewed the rest of the change. Fix this and I will read the whole thing properly.

Two of your others also came back red, both Windows only. #941 is a build break where a runtime `GOOS` skip cannot save a compile-time `syscall.Umask`, and I have posted the verified one-tag fix there. #952 I am looking at next.

jatmn:
Closing, this is a duplicate pr and the issue noted is already closed by another merged pr.

PR #955, opened 2026-08-24
perf(providers): parallelize isolated provider unit tests and shrink retry backoffs

hazyhaar:
## Summary

Provider unit tests in `internal/providers/openai` and `internal/providers/anthropic` previously took >300 seconds to run because:
1. Tests ran sequentially without `t.Parallel()`.
2. Tests covering HTTP 429/503 error classification triggered full production retry backoff schedules (2s + 4s + 8s...) totaling ~60s of physical sleep per test.

### Key Changes
- Exported `ShrinkBackoffForTest()` in `internal/providers/providerio` to clamp retry backoff bases to 1ms during test execution.
- Enabled `t.Parallel()` on isolated provider test cases (each running on independent `httptest.Server` instances).
- Reduced provider unit test suite runtime from **303s to 2.4s** (~125x speedup) with 100% passing rate under `go test -race`.

## Summary by CodeRabbit

* **Tests**
  * Expanded concurrent test execution across provider integrations.
  * Added validation for cache-write usage reporting.
  * Reduced retry delays during automated testing to shorten validation time.
  * Preserved existing coverage and assertions while improving test-suite efficiency.

* **Refactor**
  * Improved internal test support for retry and provider behavior without changing production functionality or public APIs.

### DIFF
--- a/internal/config/unknownfields.go
+++ b/internal/config/unknownfields.go
@@ -131,7 +131,7 @@ type knownField struct {
 }
 
 func derefType(t reflect.Type) reflect.Type {
-	for t != nil && t.Kind() == reflect.Ptr {
+	for t != nil && t.Kind() == reflect.Pointer {
 		t = t.Elem()
 	}
 	return t
--- a/internal/providers/anthropic/cache_breakpoints_test.go
+++ b/internal/providers/anthropic/cache_breakpoints_test.go
@@ -16,6 +16,7 @@ import (
 // unmarked (Anthropic caps breakpoints at 4 per request: system, tools, and
 // these two).
 func TestAnthropicRequestMarksLastTwoMessagesForCaching(t *testing.T) {
+	t.Parallel()
 	var gotBody map[string]any
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
@@ -73,6 +74,7 @@ func TestAnthropicRequestMarksLastTwoMessagesForCaching(t *testing.T) {
 // the breakpoint, and thinking blocks must never carry cache_control (the API
 // rejects them) — the marker goes on the last cacheable block instead.
 func TestApplyMessageCacheBreakpointsSkipsThinkingBlocks(t *testing.T) {
+	t.Parallel()
 	messages := []anthropicMessage{
 		{Role: "user", Content: "plain string"},
 		{Role: "assistant", Content: []map[string]any{
--- a/internal/providers/anthropic/dropped_test.go
+++ b/internal/providers/anthropic/dropped_test.go
@@ -13,6 +13,7 @@ import (
 // can ask the model to retry, mirroring the OpenAI provider's behavior, instead
 // of silently dropping it.
 func TestStreamCompletionEmitsDroppedOnNamelessToolUseBlock(t *testing.T) {
+	t.Parallel()
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		writeSSEEvent(w, "content_block_start", `{"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"","name":""}}`)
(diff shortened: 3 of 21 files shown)


Vasanthdev2004 (dismissed):
This is sound and the win is bigger than the title suggests. On base, `internal/providers/openai` takes 181.5s and `internal/providers/anthropic` 121.5s, almost entirely the 429 and 503 cases burning the full 2/4/8/16/30s schedule over six attempts. On this branch both are 1.3s. gemini was already 1.4s, which is why it correctly gets no `init()` shrink.

Nothing is skipped to get there. `-v` PASS counts are identical on base and on this head (openai 113, anthropic 30, gemini 33).

What I checked, since parallelising 146 tests is the kind of change that pays out later:

- No `t.Setenv` collision. The only `t.Setenv` calls in a parallelised file are in `TestOpenAIRequestPromptCacheKey`, which you deliberately left serial, and no helper reached from a parallel test calls it. Everything else that touches the environment lives in packages or tests that stayed serial.
- No shared mutable state. None of the four test packages declare package-level vars, and the only production globals in play are providerio's `retryBackoffBase` and `preSendBackoffBase`, both set once from `init()` before any test runs. `t.Parallel()` is the first statement in all 146 cases, and servers are closed via `t.Cleanup` or a deferred Close.
- Production defaults untouched: 2s and 500ms stay as they were.
- `-race` clean at `-count=4`, `-count=10 -parallel 64`, `-count=8 -cpu=1` and `-count=25 -parallel 4`, plus a full `-race ./...`.

Two notes, neither blocking.

`TestCodexProviderStreamIdleTimeoutPropagates` now runs in parallel while configuring `StreamIdleTimeout: 50 * time.Millisecond` and asserting the stream produces no error event. Its own comment says the fast path does not hit the timeout and the test just confirms the option is accepted, so the 50ms is incidental rather than the thing under test, but it is now a wall-clock window that parallelism can close. I did see it fail this way once, at `-race -count=15 -parallel 128` under 160 burner threads:

```
unexpected error event: "provider stream error: idle timeout after 50ms (upstream stopped sending data)"
```

Being straight about the calibration: it does not reproduce at CI-realistic parallelism, and the same load at `-parallel 1` passed. So this is a "raise the constant or leave this one serial" note, not a reason to hold the PR.

`ShrinkBackoffForTest` is exported from production `retry.go`, so it is reachable from the shipped binary even though the only callers are the two test `init()`s. The deadcode step that would surface it is `continue-on-error`, so nothing will flag it. Moving it behind `export_test.go` would close that off, if you want it closed.

Approving.

jatmn (changes requested):

## Merge readiness

  `internal/providers/openai/codex_test.go:770`
  The PR merge base (`ad34dc8`...) is behind live `main` (`1b5db176...`). The intervening target change added cache-write-token coverage in both `internal/providers/openai/codex_test.go` and `provider_test.go`, and `merge-tree` now reports them as changed in both. Rebase and review the resolved diff so the PR does not discard those upstream test contracts.

## Findings

- [P2] Keep the Codex timeout acceptance test out of the parallel pool
  `internal/providers/openai/codex_test.go:748`
  This PR, and not merge-base or current `main`, adds `t.Parallel()` to the test that configures `StreamIdleTimeout: 50 * time.Millisecond`. The wrapped OpenAI provider passes that value to `ScanSSEDataWithContext`, which starts a real-wall-clock idle timer and returns `ErrStreamIdle` when it wins the select. `drainCodexEvents` treats that event as a fatal test failure.

  The server writes `response.completed` synchronously, but once this test competes with the newly parallel provider suite, there is no guarantee that the server handler, HTTP read, SSE scanner, and consumer will run within 50ms. The test's own comment says it only verifies that the option is accepted, not that a 50ms idle expiry behaves correctly.

  Restore the serial scheduling for this test or split the concerns: assert option propagation without arming a tiny live watchdog, and leave the deterministic stalled-stream tests to exercise timeout behavior. This addresses the root cause — a timing-sensitive assertion in the new concurrent schedule — without reversing the product timeout contract or blocking parallelism for the rest of the suite.

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):

PR #956, opened 2026-08-25
fix(tools): make apply_patch tolerant and re-expose edit_file

kevincodex1:
Head-to-head benchmarking against Grok Build on grok-4.6 (3 tasks × 3 runs) showed Zero solving every task but spending 2.7× the tokens and 2.3× the time on the multi-file feature task. Every one of the 20 apply_patch calls in those runs failed, after which the model degraded to whole-file write_file rewrites and read→write verification loops.

Root causes and fixes:

- The sandbox and the tool both rejected an absolute path that lies inside the workspace ("must stay inside the workspace"). Models routinely echo the absolute path read_file showed them. Absolute paths now flow through the normal workspace-scope validation (inside → allowed, outside → still denied); unified-diff headers are rewritten to a/ b/ relative form before git apply, and platform symlink prefixes (/var → /private/var) are normalised the same way the other file tools do.
- The structured parser demanded a byte-exact "*** Begin Patch" / "*** End Patch"; grok-4.6 writes "*** Begin Patch ***". Decorated marker spellings are accepted, and a unified range after "@@" ("-12,4 +12,6", optionally with a heading) no longer becomes a literal context search.
- Parse errors now carry a one-line format reminder so a failed call is recoverable in one retry instead of a format-guessing loop.
- edit_file (exact search/replace with uniqueness guard, CRLF and fuzzy fallbacks) is advertised to the model again alongside apply_patch, and the apply_patch description shows the structured format.
- Writers (edit_file, write_file, apply_patch) recorded a file's line total with lineCount, which is one higher than read_file's count for a file ending in a newline. RecordSeenRange resets every observation when the total changes, so a partial read_file after a successful edit silently discarded whole-file knowledge and the next edit was refused with "content that has not been read exactly in this session". All writers now record the same total read_file reports (trackedLineTotal).
- read_file's padded "  N | " line prefix was ~19% of every read (about 1.5k tokens per 800-line file) and was re-sent on every later call. It is now the compact "N→" prefix models already know; nothing parsed the old form.
- The editing-discipline prompt steers: edit_file for a targeted change, apply_patch for multi-hunk/multi-file changes, write_file only for new files or full rewrites; a successful edit result is confirmation, so the file is not re-read to verify it.

Tests: regression tests for decorated markers, writer/reader line-total agreement (edit → partial read → edit far away still succeeds), range hunk headers, absolute in-workspace paths (structured + unified), outside-workspace rejection, header-line-only relativisation, and tool visibility; the prompt and tool-budget tests updated to the new contract.

Benchmark (grok-4.6, 3 tasks x 3 runs, harness in demo/bench-zero-vs-grok): Zero 0.8.0 617 s / 1,527k tokens -> with this change 346 s / 493k tokens; Grok Build 407 s / 713k. Every run of every configuration solved its task.

## Summary by CodeRabbit

* **New Features**
  * Improved file-editing guidance for targeted edits, multi-file patches, and new files.
  * Unified patch support now includes copying, renaming, creation, deletion, and safe absolute paths.
  * Both editing tools are available when appropriate.
  * File previews now use compact arrow-style line numbers.

* **Bug Fixes**
  * Improved line tracking after edits and writes.
  * Blocked unsafe traversal and symlink escapes.
  * Improved patch validation, newline handling, and change detection.
  * Prevented overwriting files changed during processing.

### DIFF
--- a/internal/agent/prompt_budget_test.go
+++ b/internal/agent/prompt_budget_test.go
@@ -68,7 +68,7 @@ func TestEagerToolSchemaTokenBudget(t *testing.T) {
 	}
 }
 
-func TestAgentAdvertisesPatchInsteadOfAmbiguousStringReplacement(t *testing.T) {
+func TestAgentAdvertisesBothEditTools(t *testing.T) {
 	registry := tools.NewRegistry()
 	for _, tool := range tools.CoreToolsScoped(t.TempDir(), nil) {
 		registry.Register(tool)
@@ -79,9 +79,9 @@ func TestAgentAdvertisesPatchInsteadOfAmbiguousStringReplacement(t *testing.T) {
 		names[definition.Name] = true
 	}
 	if !names["apply_patch"] {
-		t.Fatal("agent must retain apply_patch for existing-file changes")
+		t.Fatal("agent must retain apply_patch for multi-hunk changes")
 	}
-	if names["edit_file"] {
-		t.Fatal("agent must not receive the ambiguous string-replacement tool")
+	if !names["edit_file"] {
+		t.Fatal("agent must receive edit_file for targeted exact replacements")
 	}
 }
--- a/internal/agent/system_prompt.md
+++ b/internal/agent/system_prompt.md
@@ -52,7 +52,7 @@ work.
 
 - Choose the narrowest tool that safely accomplishes the step. Prefer native
   file tools - read_file, read_minified_file, list_directory, glob, grep,
-  write_file, apply_patch - over shelling out to
+  edit_file, apply_patch, write_file - over shelling out to
   cat/sed/awk/python for file operations.
   They are safer, reviewable, and produce clean diffs.
 - Prefer read_minified_file when initially exploring source code; it preserves
@@ -61,11 +61,18 @@ work.
 - Keep edits focused and reviewable. A single patch may update several related
   files when they form one coherent change; do not hide unrelated edits in a
   bulk shell or script rewrite.
-- For edits to existing files, prefer apply_patch with minimal, targeted hunks
-  and enough unchanged context to identify the intended location. Match the
-  existing indentation, imports, and idioms. Match the file's comment density:
-  do not add explanatory comments unless the user asks or the code is already
+- For edits to existing files, use edit_file for a targeted change (old_string
+  must match the file exactly and be unique, so include a few surrounding
+  lines) and apply_patch when one coherent change spans several hunks or
+  files. Use write_file only to create a file or when most of it changes; do
+  not rewrite a whole file to change a few lines. Match the existing
+  indentation, imports, and idioms. Match the file's comment density: do not
+  add explanatory comments unless the user asks or the code is already
   comment-dense.
+- A successful edit result already confirms the change; do not re-read a file
+  just to verify an edit that succeeded. If an edit fails, read the error, fix
+  the old_string or hunk, and retry the same tool rather than switching to a
+  full rewrite.
 - Solve the problem as posed, not a more general version of it. Add no
   speculative abstraction, configurability, or handling for cases that cannot
   occur, and nothing the user did not ask for. A small diff can still be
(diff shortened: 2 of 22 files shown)


kevincodex1:
@coderabbitai please do a full comprehensive review again

kevincodex1:
@coderabbitai please do a full review again on this

kevincodex1:
@coderabbit please review and make sure all is good

kevincodex1:
@coderabbitai please do a full review

anandh8x (changes requested):
Requesting changes for one reproducible data-integrity issue.

## Blocking: unified deletion ignores the expected old content

I reproduced this on the latest head, 3f654477:

1. Create `gone.txt` containing `current\n`.
2. Apply a unified deletion patch whose hunk expects `stale\n`:

```diff
diff --git a/gone.txt b/gone.txt
deleted file mode 100644
--- a/gone.txt
+++ /dev/null
@@ -1 +0,0 @@
-stale
```

`apply_patch` returns `Patch applied successfully.` and deletes `gone.txt`. It should reject the stale hunk and leave the file untouched, as `git apply` would.

The parser creates a delete operation when the new path is `/dev/null`, but hunk chunks are retained only for update/copy operations. The planner reads the existing file but applies and validates chunks only for update/copy, after which the delete path removes the file unconditionally. A header-only unified deletion is accepted for the same reason.

### Required change

- Retain parsed hunks for unified deletion operations.
- Require at least one hunk for a unified deletion.
- Verify the hunk's expected old content against the current file before removing it. The verified result should be empty before deletion proceeds.
- Keep the existing structured `*** Delete File` behavior unchanged; this requirement is specifically for unified diffs that carry expected old content.
- Add regression coverage for:
  - a matching deletion succeeds;
  - a stale deletion is rejected and leaves the file byte-for-byte unchanged;
  - a header-only unified deletion is rejected;
  - a multi-hunk deletion verifies all old content before removing anything.

## Architecture assessment

The overall direction is sound and does not need a rewrite. Keeping the structured and unified parsers separate, normalizing both into one planned operation model, preflighting all targets before mutation, and applying through an opened `os.Root` is a good design. Create-only writes and refusing an existing move/copy destination are also appropriately conservative.

The deletion bug is at the normalization boundary: a structured delete means “delete this path,” while a unified delete means “delete this path only if the expected old content matches.” The normalized operation needs to preserve that distinction. A clean fix would either carry verified delete chunks/expected content on the delete operation, or derive an empty post-image through the normal update verifier before allowing removal. Avoid special-casing the final `Remove` call without preserving the expected-content contract in the planned change.

Two worthwhile hardening follow-ups, not blockers for this review:

- Recheck an update/delete source against its planned `before` bytes immediately before commit. The current preflight catches stale content before planning, but another process can change the file between planning and replacement/removal.
- Return the exact committed prefix when a later filesystem operation fails. The current error warns that the patch was partially applied and clears tracker knowledge, but reporting the files already changed would make recovery precise instead of requiring every target to be rediscovered.

## Before/after coding run

I ran current main and this implementation through the same clean inventory fixture with the same ChatGPT/GPT-5.6 Sol configuration and prompt. Both outputs passed the visible CLI checks and an independent hidden search-contract test.

| Metric | Current main | PR #956 | Change |
|---|---:|---:|---:|
| Correctness | Passed | Passed | Same |
| Completion | 4m10.1s | 3m31.4s | 15.4% faster |
| Model requests | 25 | 21 | 16% fewer |
| Tool calls | 33 | 25 | 24.2% fewer |
| Input tokens | 450.6K | 319.3K | 29.1% fewer |
| Output tokens | 7.27K | 6.14K | 15.6% fewer |
| Local CPU | 4.29s | 3.37s | 21.4% lower |
| Peak RSS | 156.3 MiB | 176.6 MiB | 13% higher |

Completion time has normal model variance—another valid main run finished in 2m22.6s—so the strongest evidence here is the reduction in model requests, tool calls, and total input tokens rather than one wall-clock sample.

The benchmark used 08100a38; the current 3f654477 commit only delegates the duplicate reader line-count function to the same shared implementation. I validated the latest head separately with formatting, vet, diff hygiene, and focused race tests for agent, sandbox, and tools.

## Relationship to #932

This is not redundant with our runtime-performance PR #932. PR #956 supplies the rooted unified-patch implementation, edit-tool visibility, read-prefix change, and editing guidance; #932 focuses on startup and turn overhead. They overlap in several prompt/tool files, so whichever lands second should rebase carefully and preserve both sets of intentional behavior rather than resolving those files wholesale.

The current Windows smoke failure is in the existing format-on-write timeout test; the previous 08100a38 head passed Windows and the latest commit does not touch that path. It should still be rerun after the deletion fix so the final head is fully green.

kevincodex1:
@coderabbitai please do a full review

kevincodex1:
hey @coderabbitai are we good here? can you approve if its good after full review?

kevincodex1:
@coderabbitai please do a full review

PR #957, opened 2026-08-25
fix(swarm): stop original member before handoff

### LINKED ISSUE #830: fix(swarm): handoff leaves the original member executing alongside its successor
## Version / branch / commit

## OS and environment

Static control-flow review; behavior is platform-independent.

## Steps to reproduce

1. Start a swarm task whose member blocks until released and records side effects after release.
2. Call `Swarm.Handoff` for that task while the original member is still running.
3. Allow both the original member and the newly dispatched successor to continue.
4. Observe that both can produce side effects for the same objective.

Relevant code: `internal/swarm/lifecycle.go:236-282` and `internal/swarm/member.go:34-47`.

`Handoff` registers and dispatches the successor and marks the original task `StatusHandedOff`, but it does not cancel or wait for the member executing the original task. `MemberHandle` exposes `ID` and `Wait`, with no cancellation operation or task-to-member cancellation path.

## Expected behavior

A successful handoff should establish one active owner. The original member should be stopped or reach a defined safe handoff point before the successor can perform side effects.

## Actual behavior

The original member remains active while the successor starts. Both can edit files, invoke tools, or perform external side effects even though coordinator state reports the source task as handed off.

Track task-to-member handles with per-member cancellation/completion. Atomically claim the task for handoff, cancel and join the original member, then dispatch/commit the successor with rollback if launch fails.

Add a deterministic test that blocks the original worker, performs a handoff, and proves the original cannot emit any side effect after the successor begins.

gnanam1990:
## Summary

- give each swarm task its own cancellation and completion boundary
- atomically claim a handoff, cancel and join the source, then dispatch its successor
- remove queued sources and close launch/dequeue races so the original cannot resume
- keep the original task non-terminal until its member has actually stopped

## Root cause

`Handoff` only marked the source task `handed-off` and dispatched a successor. Every member was launched with the swarm-wide base context, and `MemberHandle` has no separate cancellation method, so the original member remained alive and could execute side effects alongside its replacement.

## Verification

- regression proof on `ad34dc8d`: `TestHandoffStopsOriginalBeforeSuccessorStarts` failed with `successor started before the original member stopped`
- `go test -race ./internal/swarm -count=20`
- `make fmt-check`
- `go vet ./...`
- `go test -p 1 ./...` with a fresh isolated `HOME` and file credential storage
- `make lint-static` (`0 issues.`)
- `make vulncheck` (`No vulnerabilities found.`)
- `git diff --check`

## Review

Ran the repository PR-review workflow against `ad34dc8d...22e733ad`. It found one queue-drain ordering edge during review; that edge was fixed and covered by `TestHandoffDoesNotWaitForUnrelatedQueuedLaunch`. No evidence-backed blockers remain.

Fixes #830

## Summary by CodeRabbit

* **New Features**
  * Added coordinated task handoffs that stop the current task before launching its replacement.
  * Added safeguards to reserve successor tasks and prevent conflicting updates during handoff.
  * Added clear handling when the original task cannot stop within the configured timeout.

* **Bug Fixes**
  * Improved shutdown, cancellation, queue, and task lifecycle handling.
  * Ensured aborted or unsuccessful handoffs preserve the original task safely.

* **Documentation**
  * Clarified handoff behavior, stop requirements, and completion expectations.

### DIFF
--- a/internal/swarm/coordinator.go
+++ b/internal/swarm/coordinator.go
@@ -161,7 +161,7 @@ func (c *Coordinator) SetStatus(id string, status TaskStatus) error {
 
 // BeginHandoff atomically claims a non-terminal task for one handoff caller
 // without reporting it terminal. The visible status remains pending/running
-// until FinishHandoff, so collectors cannot observe settled state while the
+// until CommitHandoff, so collectors cannot observe settled state while the
 // source member is still unwinding.
 func (c *Coordinator) BeginHandoff(id string) (Task, error) {
 	c.mu.Lock()
@@ -222,33 +222,6 @@ func (c *Coordinator) ReserveHandoffSuccessor(sourceID, successorID string) erro
 		c.handoffReservations = map[string]string{}
 	}
 	c.handoffReservations[successorID] = sourceID
-	return nil
-}
-
-// FinishHandoff publishes the terminal state only after the source execution
-// barrier has closed.
-func (c *Coordinator) FinishHandoff(id string) error {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	t, ok := c.tasks[id]
-	if !ok {
-		return fmt.Errorf("%w: %s", ErrUnknownTask, id)
-	}
-	if !t.handoff {
-		return fmt.Errorf("swarm: task %s has no handoff in progress", id)
-	}
-	if t.Status.terminal() {
-		return fmt.Errorf("swarm: task %s already %s", id, t.Status)
-	}
-	for _, sourceID := range c.handoffReservations {
-		if sourceID == id {
-			return fmt.Errorf("swarm: task %s has a reserved successor; commit the handoff atomically", id)
-		}
-	}
-	t.handoff = false
-	t.Status = StatusHandedOff
-	t.UpdatedAt = c.now()
-	c.notifyChangeLocked()
 	return nil
 }
 
--- a/internal/swarm/coordinator.go
+++ b/internal/swarm/coordinator.go
@@ -67,11 +67,12 @@ var agentColors = []string{"cyan", "magenta", "green", "yellow", "blue", "red"}
 // Coordinator is the in-memory task registry + team color assigner shared by an
 // orchestrator and its members. It is safe for concurrent use.
 type Coordinator struct {
-	mu         sync.RWMutex
-	tasks      map[string]*Task
-	colors     map[string]string // agentID -> color
-	colorIndex int
-	now        func() time.Time // injectable clock for tests
+	mu                  sync.RWMutex
+	tasks               map[string]*Task
+	handoffReservations map[string]string // successor task id -> claimed source task id
+	colors              map[string]string // agentID -> color
+	colorIndex          int
+	now                 func() time.Time // injectable clock for tests
 	// changed is closed (and replaced) on every task state change so WaitSettled
 	// can block for a transition without polling. Always non-nil after construction.
 	changed chan struct{}
@@ -80,10 +81,11 @@ type Coordinator struct {
 // NewCoordinator returns an empty coordinator using the wall clock.
 func NewCoordinator() *Coordinator {
 	return &Coordinator{
-		tasks:   map[string]*Task{},
-		colors:  map[string]string{},
-		now:     time.Now,
-		changed: make(chan struct{}),
+		tasks:               map[string]*Task{},
+		handoffReservations: map[string]string{},
+		colors:              map[string]string{},
+		now:                 time.Now,
+		changed:             make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Vasanthdev2004 (changes requested):
The bug is real and the shape of the fix is right. Giving each task its own cancellation and completion boundary is the correct answer to "the original member stayed alive", and it is a better answer than trying to add a cancel method to `MemberHandle`. `-race -count=5` clean here, `go vet` clean for linux, darwin and windows.

One blocker.

**The completion barrier is an unbounded wait on a model-invoked tool, and it takes shutdown with it.**

```go
run.stop()
if s.team(team).removeQueuedTask(taskID) { run.finish() }
<-run.done
```

`<-run.done` has no timeout and no escape. It closes only when the source member's watcher reaches `run.finish()`, which happens after `m.handle.Wait()` returns. So the whole thing rests on the member observing its context, which the code says out loud: "Launch's context is its cancellation contract".

That contract is cooperative here, not enforced. `FuncLauncher` is the only implementation and it runs `l.Run(ctx, spec)` in a goroutine in-process, wired in production to the specialist executor. A member sitting in a tool call that does not thread the context — a long shell command, a fetch that ignores it — does not return promptly on cancel, and nothing else can end the wait.

Driven with a launcher whose member ignores its context:

```
>>> Handoff has not returned after 3s; it is blocked on <-run.done
>>> Close has not returned after 3s either; it waits on lifecycleWork
after releasing the member, Handoff completed
```

Both recover once the member exits, so this is a hang rather than a leak. But `Handoff` is reachable from a swarm tool the model calls, so a stuck member wedges that turn indefinitely, and because `Handoff` holds a lifecycle admission ticket across the wait while `Close` waits on `lifecycleWork`, shutdown cannot break the cycle either. The operator's way out of a stuck member was `Close`, and that is exactly what stops working.

Bounding it does not weaken the guarantee you are adding. The point is that the successor must not start while the source can still act; a wait that gives up and reports "the source has not stopped" preserves that, because it declines to start the successor at all. At minimum select on `s.baseCtx.Done()` alongside `run.done`, so `Close` can unwedge itself rather than joining the queue behind the thing it is trying to cancel. A deadline on top of that, surfaced as a handoff error, would also tell the caller something true instead of hanging.

Two smaller observations, neither blocking.

`startTaskRun` overwrites `s.taskRuns[taskID]` unconditionally, and the comment says orphan adoption replaces "the completed boundary". If a boundary is ever replaced while unfinished, anything already waiting on the old `done` waits on a channel nobody will close any more. The adoption path does look like it only runs for tasks whose member is gone, so I could not construct it; worth an assertion or a `finish()` on the outgoing run so the invariant is enforced rather than relied upon.

The not-committed branch of `launchAdmitted` changed from always calling `t.releaseSlot()` to choosing between `releaseSlot` and `afterExitAdmitted` on `closed`. That looks right, since the non-closed case now has a queue that may want the slot, but it is the sort of accounting change that only shows up under saturation. `TestHandoffDoesNotWaitForUnrelatedQueuedLaunch` covers the neighbouring race; a case that fills a team, forces an uncommitted launch while open, and asserts the slot is reusable afterwards would pin this one directly.

Fix the unbounded wait and I will approve.

jatmn (changes requested):
I found a merge-readiness issue that needs to be addressed before this is ready.

## Merge readiness

- [P1] Rebase onto current `main` and re-run the handoff concurrency checks
  `internal/swarm/lifecycle.go:260`
  This branch forked at `ad34dc8d`, but live `main` is now `6fe0d1ed`, four commits later. The target-only history includes changes under `internal/swarm`, while this PR rewrites the same subsystem’s cancellation, queue-draining, lifecycle-admission, and shutdown interactions. As a result, the reviewed behavior is not necessarily the behavior that will merge: conflict resolution can silently restore an older lifecycle path, bypass the new task-run boundary, or alter the ordering between handoff, source completion, and queue dispatch.

  Please rebase (or reconstruct) this branch on the current target and treat the resolved swarm diff as concurrency-sensitive code, not a mechanical conflict resolution. In particular, preserve the PR’s root-cause fix end-to-end: a handoff must claim the source task, cancel its task-specific run, wait until the source has actually stopped, then make the replacement runnable; queued/dequeued and shutdown paths must continue to observe the same task-run boundary. Re-run the focused race-enabled swarm tests after resolving, including the stop-before-successor, cancellation-insensitive-source, queued-source-removal, queue-drain, launch-race, and Close tests, then request review of the rebased diff.

Vasanthdev2004 (approved):
Fixed, and fixed with more care than I asked for. Clearing my verdict.

`waitForHandoffSource` selects on the barrier, `baseCtx.Done()` and a deadline, and both escape paths re-check `run.finished()` first so a member that exits at the same moment as the cancellation still commits rather than being reported as stuck. I would not have thought to ask for that and it is the right call.

Falsified rather than read: disabling both escapes fails exactly one test each.

```
--- FAIL: TestCloseReleasesBlockedHandoffBeforeMemberExit
--- FAIL: TestHandoffTimesOutWithoutStartingSuccessor
```

The first is the shutdown cycle I was worried about, where `Handoff` held an admission ticket across the wait while `Close` queued behind it. The second name is the part that matters most: it asserts the successor is not started, so bounding the wait did not cost the guarantee the PR exists to add. Green with them restored, including `-race -count=3`.

The two observations I marked non-blocking are still open. Both are still worth doing sometime, particularly the `startTaskRun` overwrite, since an unfinished boundary replaced under a waiter is a channel nobody closes. Neither is a reason to hold this.

`gofmt` clean, `go vet` clean, CI green, 0 commits behind main.

jatmn (approved):
LGTM

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #959, opened 2026-08-25
fix(tui): let typing exit idle voice mode

### LINKED ISSUE #936: /voice mode collides with typing: Space is swallowed, so typed prompts run together and pick up stray transcriptions
### Version / branch / commit

`main` @ 1ec7219a

### OS and environment

macOS (Darwin 27.0). Reproduces on both terminal tiers described in `internal/tui/dictation_voice.go` — those that report key-release events (hold-to-record) and those that don't (press-to-toggle).

### Steps to reproduce

1. Run `/voice` to turn voice mode on.
2. Without turning it off, start typing a normal prompt into the composer, e.g. `fix the flaky test in dictation_test.go`.

### Expected behavior

Voice mode shouldn't silently corrupt typed input. Either typing should be blocked/ignored while voice mode is on (it's a modal, mic-driven mode), or the first ordinary printable keypress should exit voice mode and let the user type normally.

### Actual behavior

Voice mode only intercepts Space — every other key still goes into the prompt composer as usual. So the user types a whole prompt, all the letters land, and each Space is swallowed by the record gesture instead of inserting a space. The result:

- The composer fills with run-together text: `fixtheflakytestindictation_test.go`.
- Every Space also drives the recorder. On release-reporting terminals each Space is a press/release hold, so it starts and immediately stops a recording; on terminals without release events, Space is press-to-toggle, so alternating spaces start and stop recordings.
- Those recordings transcribe whatever the mic picked up (room noise, the user talking to themselves, keyboard clatter) and inject it into the same composer, interleaved with the typed characters.

The only signal that voice mode is on is the small `🎙 voice` indicator (`voiceModeIndicator`, `internal/tui/dictation.go:606`) plus the transient notice that has long since scrolled away — easy to miss, and the failure mode is that the user is well into a prompt before realizing it.

Relevant code:

- `internal/tui/model.go:1624` — Space is claimed for `handleVoiceSpacePress` while `voiceModeEnabled`; nothing else in the key path is gated on voice mode, so all other printable keys fall through to the composer.
- `internal/tui/dictation_voice.go` — the two-tier press/release gesture.

n/a

gnanam1990:
## Summary

- exit idle `/voice` mode when the first ordinary printable key is typed
- preserve that first key so the prompt continues normally and subsequent spaces are inserted
- clarify the voice-mode notice and add an end-to-end key-dispatch regression test

Fixes #936

## Root cause

Voice mode claimed Space for the recording gesture, but every other printable key continued into the composer. A normally typed prompt therefore lost all spaces and could trigger stray recordings/transcriptions.

## Validation

- `make fmt-check`
- `go vet ./...`
- `go test ./internal/tui -count=1`
- focused voice-mode tests under `go test -race`
- `make lint-static`
- `make vulncheck`

A real terminal screenshot of the final validation is ready locally and will be attached after browser upload permission is enabled.

## Summary by CodeRabbit

- **New Features**
  - Voice mode now exits when you begin typing, while preserving the typed input.
  - Typing or running `/voice` clearly disables voice mode.
  - Voice mode can still be enabled normally with the `/voice` command.
  - Clipboard image handling remains available through Cmd+V while voice mode is active.

- **Improvements**
  - Composer cursor behavior is smoother across activity, idle periods, and focus changes.
  - Agent runs now wait for required tools and session setup to be ready.

### DIFF
--- a/internal/tui/composer.go
+++ b/internal/tui/composer.go
@@ -158,6 +158,7 @@ func (m *model) clearComposer() {
 	m.composerActive = false
 	m.composerPastePreviews = nil
 	m.composerSelection = composerSelectionState{}
+	m.voiceModeExitedByTyping = false
 	m.input.SetValue("")
 }
 
--- a/internal/tui/dictation.go
+++ b/internal/tui/dictation.go
@@ -194,14 +194,14 @@ func (m model) toggleDictation() (model, tea.Cmd) {
 }
 
 // toggleVoiceMode flips the /voice hold-to-record gesture — the dictation
-// trigger. While on, Space records; run /voice again to type spaces normally.
+// trigger. While on, Space records; ordinary typing or /voice turns it off.
 func (m model) toggleVoiceMode() (model, tea.Cmd) {
 	if !m.dictation.available() {
 		return m.appendSystemNotice("Dictation is not configured. See docs/dictation.md to set up a local engine or a Groq/OpenAI key."), nil
 	}
 	m.dictation.voiceModeEnabled = !m.dictation.voiceModeEnabled
 	if m.dictation.voiceModeEnabled {
-		return m.showTransientNoticeInline("Voice mode on — hold Space to dictate; run /voice again to turn it off.", transientNoticeSuccess), nil
+		return m.showTransientNoticeInline("Voice mode on — hold Space to dictate; start typing or run /voice again to turn it off.", transientNoticeSuccess), nil
 	}
 	// Turning voice off is the "done dictating" signal, so release the warm sherpa
 	// streaming server — otherwise a loaded model keeps idling in RAM (and holding
--- a/internal/tui/dictation.go
+++ b/internal/tui/dictation.go
@@ -193,15 +193,17 @@ func (m model) toggleDictation() (model, tea.Cmd) {
 	}
 }
 
-// toggleVoiceMode flips the /voice hold-to-record gesture — the dictation
-// trigger. While on, Space records; ordinary typing or /voice turns it off.
+// toggleVoiceMode flips the /voice Space-to-record gesture — hold-to-record on
+// terminals with release events, press-to-toggle elsewhere. Ordinary typing or
+// /voice turns it off.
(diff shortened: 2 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Keep the typed `/voice` command from re-enabling voice mode
  `internal/tui/model.go:1627`
  When voice mode is on, the new early key-dispatch branch treats the first `/` as ordinary typing: it calls `toggleVoiceMode`, so the mode is off, and then it preserves `/` in the composer. The rest of `/voice` typing falls through normally, but Enter calls `handleSubmit`, parses the buffer into `commandVoice`, and the existing command handler calls `toggleVoiceMode` in return. That re-toggles the mode to on, so the next Space is again captured for recording instead of typing a space.

  The root cause is that the new exit transition preserves and later retriggers the explicit command without carrying the "exited by typing" context through command dispatch. Please ensure that a `/voice` submitted from this new exit path leaves `voiceModeEnabled` false, while preserving the existing explicit-command toggle semantics for other entry paths. Add a regression that sends `/`, `v`, `o`, `i`, `c`, `e`, and Enter through `m.Update`.

- [P2] Do not consume Cmd+V as ordinary text when exiting voice mode
  `internal/tui/model.go:1627`
  `keyPrintable` excludes Ctrl and Alt modifiers but not `ModSuper`. On a macOS terminal that delivers Cmd+V to the application, the key has text `v` and the new voice-mode branch matches it before the existing `Ctrl+V/Cmd+V` image-clipboard case. It turns voice mode off and inserts a literal `v`, then returns, so `readClipboardImageCmd` never runs. That breaks the existing supported image-paste path only while voice mode is idle.

  The root cause is that the new "ordinary typing" predicate is broader than the prior dispatch order and does not exclude Command-modified shortcuts. Please narrow the exit trigger to unmodified printable input (or otherwise let higher-priority global shortcuts route first), so Cmd+V continues to reach the image-clipboard handler while unmodified typing still exits voice mode. Add a focused test with `tea.ModSuper` and verify the clipboard command is returned rather than a composer insertion.

PR #974, opened 2026-08-27
fix(daemon): wait on tracked workers in TestPoolDrainKillsStraggler (#919)

### LINKED ISSUE #919: Flaky: TestPoolDrainKillsStraggler synchronizes on QueueDepth, not on the worker being tracked
### Version / branch / commit

`main` @ `dc15e822`. The test and `pool.go` are unchanged since #274, so this affects every branch.

### OS and environment

Observed on CI: `Smoke (windows-latest)`, Windows Server 2025, Go per `go.mod`. Not OS-specific in principle — it is a scheduling race, and Windows CI is simply where the window is widest.

### Steps to reproduce

`go test ./internal/daemon/ -run TestPoolDrainKillsStraggler`

Intermittent. It does not reproduce on an idle machine — I ran 30× normally and 200× under `GOMAXPROCS=1` on Windows 11, all green. It shows up on loaded shared CI runners.

Recent occurrence: https://github.com/Gitlawb/zero/actions/runs/32013099576/job/95336704923 (on the PR #892 branch, whose diff touches only `internal/config`, `internal/cli`, and `internal/tui` — nothing in `internal/daemon`).

### Expected behavior

`Drain()` force-kills a worker still running after `KillTimeout`, so `straggler.killed == 1`.

### Actual behavior

```
--- FAIL: TestPoolDrainKillsStraggler (0.00s)
    pool_test.go:231: Drain must kill a straggler still running after KillTimeout
FAIL	github.com/Gitlawb/zero/internal/daemon	0.781s
```

### Root cause

The test waits on the wrong piece of state. It gates on the slot lease, while `Drain` gates on the tracked worker set, and there is a window between the two.

- `internal/daemon/pool_test.go:222` — `waitFor(t, func() bool { return pool.QueueDepth() == 1 })`
- `internal/daemon/pool.go:169` — `QueueDepth()` is `len(p.slots)`, i.e. slot-channel occupancy, set the moment `Run` leases a slot.
- `internal/daemon/pool.go:246` — the handle only enters `p.active` later, in `runOnce` → `p.track(id, handle)`, *after* `Launcher` returns.
- `internal/daemon/pool.go:336` — `Drain`'s grace loop reads `len(p.active)`.

So `waitFor` can return while `p.active` is still empty. If `Drain()` runs in that window it takes the early return at `pool.go:339`:

```go
if n == 0 {
    return // all workers drained gracefully
}
```

and never reaches the force-kill block at `pool.go:344-354`. `straggler.killed` stays `0`, which is exactly the assertion that fires.

This is a test-synchronization defect, not a `Drain` bug — production `Drain` is correct for a worker that is actually tracked.

### Suggested fix

Synchronize on the state `Drain` actually reads. `WorkerStats()` (`pool.go:361`) is built from `p.active`, so:

```go
waitFor(t, func() bool { return len(pool.WorkerStats()) == 1 })
```

Worth a quick grep for the same `QueueDepth()`-as-readiness assumption in the neighbouring pool tests.

cairn-intern:
## Summary

Fixes #919.

`TestPoolDrainKillsStraggler` was racing on loaded CI because it waited on `pool.QueueDepth() == 1` (slot occupancy) before calling `Drain()`. `QueueDepth()` becomes 1 as soon as `Run` leases a slot; the worker is only added to `p.active` later, after `Launcher` returns (`runOnce` → `track`). `Drain()` reads `len(p.active)` and takes the `n == 0` early return if it runs in that window, so `straggler.killed` stays 0.

This is a test synchronization defect, not a production `Drain` bug.

## Change

In `internal/daemon/pool_test.go`, wait on the state `Drain` actually reads:

```go
waitFor(t, func() bool { return len(pool.WorkerStats()) == 1 })
```

`WorkerStats()` is built from `p.active`.

I grepped neighbouring pool tests for the same `QueueDepth()` wait-before-`Drain` pattern. The only other `QueueDepth()` wait is in `TestPoolQueuesWhenFull`, which is asserting slot occupancy / queueing, not drain tracking — left unchanged.

Test-only; no production code change.

## Notes

Opened by **cairn-intern**. Parent issue has `issue-approved` + `bug`.

I did not run `go test` for this change (no local checkout). Worth `go test ./internal/daemon/ -run TestPoolDrainKillsStraggler` on CI.

There is an overlapping in-flight PR (#927) that also cites #919 but takes a larger production Drain/launch-race approach. This PR is the small test-only sync the issue describes.

## Summary by CodeRabbit

* **Tests**
  * Improved test synchronization for worker shutdown scenarios.
  * Added clarification to ensure straggler workers are properly detected and terminated during drain operations.

### DIFF
--- a/internal/daemon/pool_test.go
+++ b/internal/daemon/pool_test.go
@@ -224,7 +224,9 @@ func TestPoolDrainKillsStraggler(t *testing.T) {
 		_, _ = pool.Run(context.Background(), WorkerSpec{Session: "a"}, &collectSink{})
 		close(runDone)
 	}()
-	waitFor(t, func() bool { return pool.QueueDepth() == 1 })
+	// Drain reads p.active (exposed as WorkerStats), not QueueDepth/slot occupancy.
+	// QueueDepth can become 1 before track(), so Drain would take the n==0 early return.
+	waitFor(t, func() bool { return len(pool.WorkerStats()) == 1 })
 
 	pool.Drain() // KillTimeout elapses, straggler is force-killed
 	if atomic.LoadInt32(&straggler.killed) != 1 {


cairn-intern:
Closing as duplicate of #927, which already includes this test-sync (WorkerStats wait) plus the production Drain/launch-race fix. Sorry for the noise.

PR #981, opened 2026-08-27
fix(background): treat POSIX zombie leader as already-exited

### LINKED ISSUE #862: bug: TerminateCommand reports spurious SIGKILL-timeout when ps is unavailable and leader is a zombie
## Context

Follow-up from review of [PR #774](https://github.com/Gitlawb/zero/pull/774).

`internal/background/process_posix.go:47-51` (`terminateOwnedProcess`) always returns `leaderAlreadyExited = false`, unlike its Windows counterpart in `internal/background/process_windows.go`, which computes `alreadyExited` from `GetExitCodeProcess` before attempting termination. `TerminateCommand` (`internal/background/terminate.go`) uses that flag to decide whether a termination error is safe to suppress once the reap has independently succeeded.

## Reproduction

Confirmed empirically (WSL Ubuntu, Go 1.26.5 linux/amd64): start a child as its own process-group leader, let it become a zombie, make `ps` unresolvable (e.g. `PATH=""`), then call `TerminateCommand`.

Result:
```
TerminateCommand elapsed=406ms err="process <pid> did not exit after SIGKILL"
ProcessState: exit status 0
```

The reap succeeds cleanly, but `TerminateCommand` still returns a spurious termination error — because `signalTargetRunning` (`internal/execution/process_unix.go`) falls back to the raw `kill(pid, 0)` result when `ps` is unavailable or unparseable, and `kill(0)` succeeds against a zombie. The tree is genuinely already dead, but the code can't distinguish that from "still alive" without `ps`, so it burns both grace periods and reports failure.

This matters because that error string is surfaced verbatim in `zero daemon start`'s failure message when `startAndAwaitDaemonProcess`'s cleanup path hits it.

## Why this isn't a simple fix

Using "the reap succeeded" (or "the leader was already a zombie before termination began") as a stand-in `leaderAlreadyExited` is NOT safe: a zombie leader can still have a live orphaned child in the same process group (see `TestTerminateCommandKillsChildAfterLeaderExits`), and only `ps`/`/proc` can currently tell that group-liveness apart from "unknown." Using the reap result to suppress the error would risk masking that genuinely-alive-descendant case too.

A correct fix likely needs either:
- A portable way to positively confirm the group has zero non-zombie members without `ps` (e.g. `/proc/<pid>/stat` scanning on Linux, a Darwin-specific equivalent, falling back to today's conservative behavior only where no such mechanism exists), or
- A deliberate, documented decision to accept the rare spurious-failure report as the price of never risking a false "success," with a clearer error message that says the liveness check couldn't run rather than asserting "did not exit after SIGKILL".

Requires design discussion before implementation. Not blocking `ps` is present on all Unix targets we ship for today.

cairn-intern:
## Summary

Fixes #862.

`terminateOwnedProcess` on POSIX always returned `leaderAlreadyExited=false`, unlike Windows which probes `GetExitCodeProcess` before `taskkill`. `TerminateCommand` only discards a termination error after a successful reap when that flag is set, so a zombie group leader with `ps` unresolvable (`PATH=""`) reported `did not exit after SIGKILL` even though `cmd.Wait` collected exit status 0. `kill(0)` succeeds against a zombie; without `ps`, `signalTargetRunning` conservatively treats that as still running.

POSIX now probes whether the leader is already waitable-as-exited **before** signalling, matching the Windows intent:

- Linux: `waitid(P_PID, WEXITED|WNOHANG|WNOWAIT)` plus `/proc/<pid>/stat` state `Z`
- Darwin: `sysctl kern.proc.pid` `P_stat == SZOMB`
- other POSIX: leave the flag false (today's conservative behaviour)

Group/tree termination still runs, so live descendants are signalled (`TestTerminateCommandKillsChildAfterLeaderExits`). The flag is about the leader only.

Regression: `TestTerminateCommandZombieLeaderWithoutPS` — zombie leader, empty `PATH`, `TerminateCommand` must return nil.

## Notes

Opened by **cairn-intern**. Follow-up from review of #774. Issue #862 is **not** `issue-approved`; intern broadening per Vasanth/euxaristia.

`gofmt` applied. `go test` was not run locally (no full checkout). Worth `go test ./internal/background/ -count=1` on Linux CI, including `-run TestTerminateCommandZombieLeaderWithoutPS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command termination handling for processes whose leaders exit before cleanup completes.
  * Prevented false termination timeout errors when reaping zombie process leaders.
  * Preserved genuine process-group termination failures when child processes remain active.
  * Improved consistency across supported operating systems.

### DIFF
--- a/internal/background/process_posix.go
+++ b/internal/background/process_posix.go
@@ -45,6 +45,16 @@ func terminateProcess(pid int) error {
 // rediscovery after an owned leader exits. Ordinary commands fall back to the
 // safe PID/tree path rather than assuming their PID is also a process-group ID.
 //
+// The bool result is whether the leader had already exited before termination
+// was attempted — the POSIX analogue of Windows GetExitCodeProcess returning a
+// value other than STILL_ACTIVE. A waitable zombie is already-exited: Wait can
+// collect it, but kill(0) still succeeds, and when ps is unavailable
+// signalTargetRunning conservatively treats that as "still running".
+// TerminateCommand uses the flag to discard the resulting spurious
+// SIGKILL-timeout once the reap has independently succeeded. The flag is about
+// the leader only; group/tree termination is still invoked so live descendants
+// are signalled.
+//
 // The Setpgid && Pgid == 0 guard only recognizes ConfigureChildProcessGroup's
 // own convention. A command made its own session leader via Setsid (as opposed
 // to Setpgid) is also its own process-group leader in practice, but takes the
@@ -52,8 +62,9 @@ func terminateProcess(pid int) error {
 // because TerminateCommand has exactly one caller in this codebase; worth
 // covering explicitly if Setsid-configured commands start using this path too.
 func terminateOwnedProcess(cmd *exec.Cmd) (bool, error) {
+	alreadyExited := leaderWaitableExited(cmd.Process.Pid)
 	if cmd.SysProcAttr != nil && cmd.SysProcAttr.Setpgid && cmd.SysProcAttr.Pgid == 0 {
-		return false, execution.TerminateProcessGroup(cmd.Process.Pid, terminationGracePeriod, terminationPollInterval)
+		return alreadyExited, execution.TerminateProcessGroup(cmd.Process.Pid, terminationGracePeriod, terminationPollInterval)
 	}
-	return false, terminateProcess(cmd.Process.Pid)
+	return alreadyExited, terminateProcess(cmd.Process.Pid)
 }
--- a/internal/background/process_posix.go
+++ b/internal/background/process_posix.go
@@ -3,7 +3,9 @@
 package background
 
 import (
+	"errors"
 	"os/exec"
+	"syscall"
 	"time"
 
 	"github.com/Gitlawb/zero/internal/execution"
@@ -67,4 +69,15 @@ func terminateOwnedProcess(cmd *exec.Cmd) (bool, error) {
 		return alreadyExited, execution.TerminateProcessGroup(cmd.Process.Pid, terminationGracePeriod, terminationPollInterval)
 	}
 	return alreadyExited, terminateProcess(cmd.Process.Pid)
+}
+
+// terminationTargetGoneAfterReap independently checks the exact signal target
+// used for an owned command. In particular, reaping a zombie group leader does
+// not make this return true while any descendant still occupies that group.
+func terminationTargetGoneAfterReap(cmd *exec.Cmd) bool {
+	target := cmd.Process.Pid
+	if cmd.SysProcAttr != nil && cmd.SysProcAttr.Setpgid && cmd.SysProcAttr.Pgid == 0 {
+		target = -target
+	}
+	return errors.Is(syscall.Kill(target, syscall.Signal(0)), syscall.ESRCH)
 }
--- a/internal/background/process_posix_test.go
+++ b/internal/background/process_posix_test.go
@@ -6,6 +6,7 @@ import (
 	"bufio"
 	"errors"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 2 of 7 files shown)


Vasanthdev2004 (dismissed):
Approving. I went looking for the usual ways this shape goes wrong and did not find one.

**The probe does not steal the exit status.** `WNOWAIT` leaves the child waitable, so the `cmd.Wait` that follows still collects it. That was my first concern and it is the right flag. `WEXITED` alone also means a stopped or continued child is not reported as exited, and the `CLD_EXITED/KILLED/DUMPED` check makes that explicit rather than relying on it.

**A zombie leader with live descendants is still handled.** This was the failure mode I most expected: the flag is computed before signalling, but `TerminateProcessGroup` / `terminateProcess` still runs either way, so the group is signalled regardless. The flag only decides whether a termination error is discarded, and `terminate.go:58` gates that on the pre-termination observation rather than on the reap succeeding. The comment there says why, and it is the right reason: a successful reap says nothing about the tree.

**The PID cannot be recycled between probe and kill.** The leader is unreaped at probe time, so the zombie holds its PID until `Wait`.

**Build tags are exhaustive.** `linux`, `darwin`, and `!windows && !linux && !darwin`. It cross-compiles clean for linux, darwin, windows, freebsd and openbsd, so the `other` fallback is genuinely reachable rather than an orphan file, and returning false there preserves the old conservative behaviour.

The `other` fallback and the `pid <= 1` guard both fail in the safe direction, and on linux a `waitid` that cannot answer falls through to `/proc` rather than guessing.

The regression is a good one: it waits for the leader to actually become waitable-exited instead of sleeping and hoping, sets `PATH=""` to reproduce #862 exactly, and asserts both that `TerminateCommand` returns nil and that the process was reaped. It also skips loudly on other platforms rather than passing vacuously. Mutating the package timing vars is safe here since nothing in this package uses `t.Parallel()`, and it matches what the neighbouring tests already do.

**What I could not run.** `process_posix_test.go` is `//go:build !windows`, so none of it executed on my machine and I could not falsify the new test locally. It ran on the ubuntu job (`ok github.com/Gitlawb/zero/internal/background`), and macOS covers the sysctl path, but I am not going to imply I exercised either. The darwin implementation I read rather than ran; `SysctlKinfoProc` with `SZOMB` is the reasonable choice given `x/sys/unix` has no `waitid` wrapper there.

Worth knowing: this PR had never run CI at all. Its checks were sitting at `action_required` behind the fork gate, so the green you may have seen was CodeRabbit alone. I released it and the full suite passed. `go test -race -count=2 ./internal/background` is also clean here, for whatever the Windows subset is worth.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto the current `main` before merge
  `HEAD`
  This head is based on `27b319ca`, while live `main` is `1b5db176`, two commits ahead. The intervening OAuth and TUI work does not overlap this diff and a three-way merge is currently clean, but that is not a substitute for a current review base: repository policy requires a fresh base so that CI and review apply to the actual merge result. Rebase onto `main`, preserve the upstream changes, and have the resolved head re-reviewed.


## Findings

- [P2] Do not turn a failed group termination into a successful cleanup
  `internal/background/terminate.go:58`
  The root cause is that `leaderAlreadyExited` is evidence about one process, while `terminateErr` is evidence about the whole PID/tree target. A zombie leader can be reaped successfully even when a descendant is still alive. The new condition discards *every* `terminateErr` after that leader reap, so in the conditional but real case where a descendant cannot be signalled or remains alive after TERM/KILL, `TerminateProcessGroup` reports the failure but this method returns nil and tells its caller cleanup succeeded. The base branch returned the error.

  The fix needs to preserve the #862 outcome—do not report the false timeout for a zombie-only group when `ps` is unavailable—without treating a reaped leader as proof that the group is empty. Keep the pre-`Wait` group signal and the safe PID-only fallback. Suppress an error only after independently establishing that no live group member remains; otherwise return the original termination error. Add a regression that covers an already-exited leader plus a descendant that cannot be terminated, and demonstrate that it fails if the unconditional error suppression is restored. The current tests cover a zombie-only group and a normally signallable descendant, but not this failure boundary.

## Overall guidance

This PR is close in scope, but the review friction comes from mixing two different process-lifecycle facts: **leader exit/reap** and **tree termination**. Before changing `TerminateCommand`, write the decision table for those facts separately:

- zombie-only leader + unavailable `ps` → successful reap and no spurious timeout;
- zombie leader + live, signallable descendant → signal the group and verify the descendant is gone;
- zombie leader + descendant that cannot be stopped → preserve the termination error;
- live leader that exits after TERM → retain the existing conservative behavior unless that broader case is deliberately designed and tested.

Then make the implementation and tests prove the same table on each supported platform. Keeping the approval and fresh-base gates settled before implementation will also keep future review rounds focused on one immutable, policy-approved diff rather than process and branch-state follow-ups.

jatmn (approved):
@Vasanthdev2004 lgtm

kevincodex1 (approved):

PR #983, opened 2026-08-27
fix(tools): reject fuzzy edits with drifted content

### LINKED ISSUE #963: fix(tools): fuzzy edit can overwrite a block whose interior drifted
### Version / branch / commit

### Problem

`edit_file` promises that `old_string` must match byte-for-byte, but `blockAnchorReplacer` falls back to matching the first and last lines plus an average interior similarity threshold of 0.65. It then replaces the entire on-disk span even when the interior differs from `old_string`.

### Reproduction

Given an on-disk block containing:

```go
func a() {
	x := 1
	y := 2
	z := 3
	return
}
```

request an edit whose `old_string` is identical except for `y := 99`. The fuzzy matcher accepts the block and reports success, replacing the real `y := 2` line even though the requested old content never existed.

### Expected behavior

A drifted interior must make the edit fail or require an exact re-read/retry. The tool must not discard content that differs from `old_string`.

### Actual behavior

The whole span is replaced and the tool reports `Successfully edited (replaced 1 occurrence)`.

### Relevant code

- `internal/tools/edit_replacers.go:30,154-249,466-481`
- `internal/tools/edit_file.go:109-131`

### Suggested fix

Reject fuzzy block matches when any interior line materially differs from the requested old block. Add a regression test using the `y := 2` / `y := 99` drift case and prove it fails before the fix.

PierrunoYT:
## Summary

- remove the similarity-based block-anchor fallback that could replace content differing from `old_string`
- remove the later context-aware fallback with the same unsafe first/last-anchor behavior
- retain normalization-only fuzzy matching for indentation, whitespace, escapes, and boundaries
- add regression coverage for the reported `y := 2` / `y := 99` interior drift and verify the file remains unchanged

The regression test failed before the fix with `Successfully edited target.txt (replaced 1 occurrence).`

## Linked issue

Fixes #963

## Checklist

- [x] Changed Go files are `gofmt` clean.
- [x] Tests added/updated for the change.
- [x] No UI changes.

## Validation

- `go vet ./...`
- `go test ./...`
- `go test ./internal/tools -count=1`
- `make vulncheck` — no vulnerabilities
- `git diff HEAD --check`

`make fmt-check` remains blocked on the current base by existing formatting findings under `internal/perfbench/testdata/`; this PR does not modify those fixtures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fuzzy file editing to avoid replacing content when interior lines differ from the requested text.
  * Ambiguous or potentially unsafe matches are now rejected instead of selecting a similar-looking block.
  * Files remain unchanged when an exact, safe match cannot be confirmed, helping prevent unintended edits.

### DIFF
--- a/internal/tools/edit_file.go
+++ b/internal/tools/edit_file.go
@@ -102,10 +102,10 @@ func (tool editFileTool) RunWithOptions(ctx context.Context, args map[string]any
 	}
 
 	// Fuzzy fallback: when the exact string (and its CRLF translation) is absent,
-	// run a cascade of tolerant matchers (trimmed lines, block anchors, collapsed
-	// whitespace, indentation drift, escape normalization) to locate the span the
-	// model intended. Only a span that occurs literally in the file is accepted,
-	// so the replacement applied below is still exact.
+	// run a cascade of tolerant matchers (trimmed lines, collapsed whitespace,
+	// indentation drift, escape normalization) to locate the span the model
+	// intended. These transformations must preserve non-whitespace content; a
+	// merely similar interior is not safe to replace.
 	if occurrences == 0 {
 		findOld, findNew := oldString, newString
 		if strings.Contains(content, "\r\n") && !strings.Contains(findOld, "\r\n") {
--- a/internal/tools/edit_replacers.go
+++ b/internal/tools/edit_replacers.go
@@ -2,10 +2,9 @@ package tools
 
 // Fuzzy fallback matching for edit_file. When the model's old_string fails to
 // match byte-for-byte (drifted indentation, collapsed whitespace, escaped
-// characters, a slightly-misremembered middle line), these replacers propose
-// candidate spans that plausibly correspond to what the model intended. Only a
-// candidate that occurs literally in the file is accepted, so the replacement
-// itself is always exact even when the match was tolerant.
+// characters), these replacers propose candidate spans that differ only by the
+// normalization each strategy explicitly handles. They never infer a match
+// from similar but different content.
 //
 // Strategy cascade ported from opencode's edit tool, whose replacers were in
 // turn distilled from Cline's diff-apply evals and gemini-cli's edit corrector.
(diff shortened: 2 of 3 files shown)


Vasanthdev2004 (approved):
Approving. Deleting 248 lines of matching fallback is the kind of change where the risk is over-deletion rather than the fix itself, so I measured it instead of reading the diff and hoping.

I ran the same eleven inputs through `fuzzyEditMatch` on `main` and on this head. Exactly one result changes, and it is the one you set out to change:

| case | main | this PR |
|---|---|---|
| indent drift (spaces to tab) | match | match |
| trailing whitespace | match | match |
| leading blank line | match | match |
| collapsed inner spaces | match | match |
| escape normalization | match | match |
| CRLF content | match | match |
| **interior drift, `y := 2` against `y := 99`** | **match** | **no match** |
| interior extra line | no match | no match |
| first/last anchor only | no match | no match |
| blank-line drift inside | no match | no match |

Every legitimate normalization still resolves to the same span. The last three are the interesting ones: they did not match on `main` either, so `blockAnchorReplacer` and `contextAwareReplacer` were much narrower in practice than 248 lines suggests. They effectively only fired on the drifted-interior case, which is why removing them costs nothing else.

The honest caveat is that eleven hand-written cases are a sample, not a proof. But they cover each retained normalization plus the shapes the deleted code was supposed to handle, and nothing in that set regressed.

Two other things I checked:

You replaced `TestEditFuzzyBlockAnchorToleratesMiddleDrift` with `TestEditFuzzyRejectsInteriorDrift` rather than deleting it. Inverting the assertion is the right way to record a deliberate behaviour change; a bare deletion would have hidden it.

This one is standalone. It shares no files with #892 to #895, so it does not inherit the ordering problem that series has.

`go test ./internal/tools` passes in full.

jatmn (approved):
@Vasanthdev2004 LGTM

kevincodex1 (approved):

PR #985, opened 2026-08-27
fix(tui): bound live dictation regions

### LINKED ISSUE #965: fix(tui): live dictation can panic when the caret is not at the end
### Version / branch / commit

### Problem

Live dictation anchors its region on the full composer text but later compares and slices only the prefix before `regionStart`. When the caret is inside existing text, the external-delete adjustment can make `regionStart` negative. The next partial transcript slices `[]rune(text)[:regionStart]` and panics.

### Reproduction

1. Put `hello world` in the composer.
2. Move the caret to column 5, after `hello`.
3. Start live dictation and receive successive partial updates.

The first update can compute `anchor = "hello world"`, `prefix = "hello"`, and `delta = 6`, moving `regionStart` from 5 to -1. A subsequent partial slices with the negative index.

### Expected behavior

Dictation updates remain bounded to a valid composer range regardless of caret position or concurrent typing/deletion.

### Actual behavior

A negative slice index panics the TUI process and can lose the in-flight run.

### Relevant code

- `internal/tui/dictation_stream.go:103,115,120-132`

This is related to #936's voice-mode/typing collision but distinct: #936 concerns swallowed spaces and stray transcription, while this issue is a negative-index process crash.

### Suggested fix

Clamp and validate `regionStart` and `regionEnd` before every read and slice, then add a regression test with the caret in the middle of existing text and at least two partial updates.

PierrunoYT:
## Summary

- Anchor live dictation to the composer prefix before the caret, fixing the mid-caret crash.
- Replace or discard a tracked region only when its bounds, prefix, and exact rendered text still match. Re-anchor stale regions without deleting user edits, and track the separator space so cancellation removes it too.
- Cover successive partials, stale bounds, same-length user replacements, and backspace-plus-cancel.
- Address the current review follow-ups: reject Unicode whitespace in decoded browser title origins (without changing execution URLs), and pin parent reasoning-effort forwarding through `runResume`.

Fixes #965.

## Scope and base

Updated from current upstream main with a history-preserving merge and a normal push. ACP browser metadata and specialist model propagation are already on upstream main; only the small review follow-ups above are newly changed outside dictation.

## Regression proof

- Before the Unicode fix, encoded U+00A0, U+2003, and U+3000 fail with `unsafe origin accepted before permission translation`, including `browser open https://safe.example\u00a0evil.test`. The test calls the title builder directly, then checks start and permission payloads through JSON round trips, while confirming execution URL normalization accepts the unchanged input.
- Temporarily removing `runResume`'s `ParentReasoningEffort` forwarding fails the new assertion with `--reasoning-effort "" (present=false)`. Forwarding was restored before validation.

## Validation

Passed on Linux with the repository-declared Go 1.26.6:

- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- `go test ./internal/acp ./internal/specialist ./internal/tui -race -run "Browser|RunResumePassesTheParentModel|Dictation" -count=1`
- `make vulncheck`
- `git diff HEAD --check`

Advisory `make lint-static` reports four unrelated existing staticcheck findings: QF1001 in `internal/installtest/workflow_permissions_test.go:20`, and QF1008 in `internal/proxydial/proxydial.go:67,72` and `internal/tools/web_fetch.go:315`. These files are identical to upstream main; no finding targets this PR's changes.

Fixture Git commits used process-local `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false`, avoiding inherited orb signing in disposable test repositories. Composer interactions and protocol title semantics are verified by state/payload assertions; there are no layout/style changes. Cross-platform execution remains with GitHub CI. No PR merge performed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved streaming dictation so partial text is separated correctly from existing content.
  - Preserved user edits when live dictation content changes while a partial result is in progress.
  - Prevented discarded dictation text from removing user-entered replacements.
  - Tightened browser title validation by rejecting additional unsafe whitespace characters.
  - Improved tool results by including accurate file diffs and locations.
  - Preserved model and reasoning settings when resuming specialist tasks.


Vasanthdev2004 (approved):
Approving. The real fix here is the anchor change, and it is load-bearing: reverting `regionAnchor` from `string(stateRunes[:state.cursor])` back to `state.text` fails `TestDictationStreamingPartialWithCaretInsideExistingText`. Anchoring to the whole composer meant `prefix` could never equal `anchor` once the caret sat mid-text, which is what drove the arithmetic negative.

One thing to fix before this is done, and it is about coverage rather than correctness.

**Neither clamp is pinned by any test.** I removed the first, then the second, then both, and every `DictationStreaming` test stayed green each time, including `TestDictationStreamingPartialClampsRegionAfterComposerShrink`, which is named for exactly the case the clamps exist to handle.

They are not decorative. With both removed, driving the region out of bounds panics in the same class as the bug this PR fixes:

```
region past end after shrink   PANIC: slice bounds out of range [:40] with capacity 32
negative start                 PANIC: slice bounds out of range [:-3]
```

So the clamps guard reachable states, and nothing stops a later refactor deleting them with CI green and reintroducing the panic. The shrink test presumably keeps the region in range through the prefix-delta path and never reaches the clamp. Setting the region out of bounds directly is enough to pin it:

```go
m.setComposerState(composerState{text: "ab", cursor: 2})
m.dictation.regionActive = true
m.dictation.regionStart = 40
m.dictation.regionEnd = 60
m.dictation.regionAnchor = "a much longer previous composer"
m.applyStreamingText("partial text")   // panics without the clamps
```

A negative-start case covers the second clamp the same way.

One concern of mine that turned out to be nothing, in case it saves you the thought: the first activation slices `stateRunes[:state.cursor]` before `regionActive` is true, so it is outside the clamped branch. I checked whether `currentComposerState` can return a cursor past the end on the `!composerActive` path, where it uses `m.input.Position()` raw with no `normalizeComposerState`. It cannot; `bubbles/textinput` bounds the cursor on both `SetValue` and `SetCursor`, so that slice is safe.

Approving because the code is correct as written and the missing coverage does not make it wrong, but I would rather the clamps were pinned before this lands.

`go test ./internal/tui` passes in full.

Vasanthdev2004 (changes requested):
Retracting my approval. I approved this earlier today and I was wrong: the first clamp introduces a data-loss regression that `main` does not have. I found it by going back over the half I had flagged as untested, which is exactly where it was hiding.

## The user's own text is destroyed

Type `hello world`, dictate, then backspace past the live region while partials keep arriving, then cancel:

```
                head (dc80cecd)                      main (eeea3308)
p1     "hello world there"  [11,17)          "hello world there"  [11,17)
edits  "hello wor" cursor=9                  "hello wor" cursor=9
p2     "hello wor there friend"  [7,20)      "hello worthere friend"  [9,21)
p3     "hello there friend againiend" [5,24) "hello worthere friend again" [9,27)
cancel "helloiend"                           "hello wor"
```

On `main` the user's `hello wor` survives and cancel restores it exactly. On this head their `wor` is eaten, a stray `iend` is left behind, and cancel leaves `helloiend`. The region desyncs progressively, `[11,17)` to `[7,20)` to `[5,24)`, so each partial deletes a span the dictation never wrote.

## Why the first clamp causes it

The pre-clamp at `dictation_stream.go:96-101` runs before line 122 computes `prefix := string(stateRunes[:m.dictation.regionStart])`. Once `regionStart` has been clamped down to the shrunken composer length, `prefix` is no longer "the text before the live region", it is a truncation of the new composer, and a truncation is almost always a prefix of the old anchor. So the delete-before-the-region branch at 134-139 fires for an edit that was not before the region, and subtracts a delta that has no meaning. The second clamp then stops the panic but leaves the region pointing at unrelated text.

## The fix is smaller than the current diff

Removing only lines 96-101 and keeping the `regionAnchor` change restores correct behaviour and leaves your own tests green:

```
p3     "hello worthere friend again"  [9,27)
cancel "hello wor"
ok  github.com/Gitlawb/zero/internal/tui
```

The anchor change is the actual fix for #965, and it stands on its own: reverting just it fails `TestDictationStreamingPartialWithCaretInsideExistingText` with the reported `slice bounds out of range [:-1]`. The clamps were added as defence and the first one is doing harm instead.

If you want to keep a clamp for the genuinely out-of-range case, it needs to sit after the anchor comparison rather than before it, so the prefix is still compared against the region that actually existed. And it wants a test: I noted in my earlier review that removing either clamp left every dictation test passing, which is what let this through. That observation was right and I did not follow it far enough before approving.

Sorry for the churn on this one.

anandh8x (changes requested):
The mid-caret anchor change fixes the reported crash, but the pre-comparison clamp can corrupt user text. If the user backspaces across a live dictation region while later partials arrive, clamping  before comparing the anchor makes the shrink look like an edit before the region; subsequent partials then replace normal composer text, and cancel does not restore it. Please move/remove that first clamp so anchor comparison uses the original region position, retain bounds protection only at the slice/delete boundary, and add a regression covering the shrink-plus-cancel path.

anandh8x:
Correction to my review: the problematic value is regionStart. Clamping regionStart before the anchor comparison causes the destructive misclassification described above.

jatmn (changes requested):

## Merge readiness

  `internal/tui/dictation_stream.go:93`
  This branch's merge base is `27b319ca88a3180bed5183f0c599e9307f3ece12`, while live `main` is `1b5db1765672820caac1684b168c9898b5ba3593`. GitHub currently reports it mergeable, but the repository requires a fresh base before review. Please rebase and re-run the affected validation on the resolved head.

## Findings

- [P1] Do not clamp a stale live region into user text
  `internal/tui/dictation_stream.go:109`
  The new prefix anchor only proves that the text before `regionStart` is unchanged; it does not prove that `[regionStart, regionEnd)` is still the rendered transcript. For example, after rendering a partial at cursor 1 in `aOLDz`, replacing the composer with `ab` leaves `regionStart == 1` and the prefix `a` intact but makes `regionEnd` stale. The final clamp turns that old range into `[1,2)`, so the next partial deletes the user's `b`. The same sequence preserves `b` on the merge base. Re-anchor or otherwise preserve the replacement before applying the delayed partial, and add regression coverage, so it never overwrites user text.

Vasanthdev2004 (dismissed):
Approving. The regression I retracted over is gone, and I drove it rather than reading the diff.

Replayed the exact sequence through the real key path, eight `tea.KeyBackspace` presses through `applyComposerKey` rather than a stand-in, then cancel:

```
                HEAD 329dd123                            MAIN
p1      "hello world there"          [11,17)     "hello world there"          [11,17)
edits   "hello wor"            cur=9 [11,17)     "hello wor"            cur=9 [11,17)
p2      "hello worthere friend"       [9,21)     "hello worthere friend"       [9,21)
p3      "hello worthere friend again" [9,27)     "hello worthere friend again" [9,27)
cancel  "hello wor"                              "hello wor"
```

Byte-identical at every step, bounds included. The values I reported (`p3 "hello there friend againiend"`, `cancel "helloiend"`) do not reproduce. The mechanism is gone by construction rather than patched around it: 058ba802 deleted the pre-clamp, and 329dd123 deleted the post-clamp along with the whole prefix-versus-anchor shift switch. There is no `clamp(` left in `dictation_stream.go`.

What replaced it is better than what I asked for. Both delete sites are now gated on `liveRegionMatches`, which requires the span being removed to be byte-identical to what dictation actually rendered. That is an invariant main does not have, and it means this subsystem structurally cannot delete text it did not write. Driving #965 itself shows the difference: main goes `[5,11)` to `[-1,12)` to a `slice bounds out of range [:-1]` panic and cancels to `"d there world"`, while head cancels to `"hello world"`.

One thing worth knowing rather than fixing. `dictation_stream.go:137`, the `string(stateRunes[d.regionStart:d.regionEnd]) == d.regionRendered` clause, is not pinned by any test: replacing it with `true` leaves the dictation suite green. It is load-bearing though, and it is the sole cause of the one behaviour difference from main. When a user edits inside the live region and the bounds stay valid, head abandons the region and re-anchors, leaving the edited copy behind:

```
   user overtypes the dictated span (" there" -> " MINE!")
   HEAD  cancel "hello world MINE!"
   MAIN  cancel "hello world"          <- the user's " MINE!" silently deleted
```

So it diverges in the safe direction: head keeps bytes the user typed and leaves visible duplicate transcript, main deletes them. You cannot distinguish "fixed a typo" from "typed replacement text" at that point, so I read this as a deliberate trade rather than a defect, and blocking on it would be asking for main's data loss back.

Two things to pick up whenever, neither gating: a test pinning line 137 specifically (a same-length in-region edit with valid bounds, since every existing test short-circuits on the bounds arm first), and re-running `needsLeadingSpace` in the re-anchor branch so the separator space stops being orphaned. The second is present on main too.

Sorry again for the churn. The retraction was right and the fix is right.

jatmn (dismissed):
lgtm

PierrunoYT:
PierrunoYT addressed the remaining review follow-ups in [e5912c2](https://github.com/PierrunoYT/zero/commit/e5912c2f8b39efc8795657cfeafac094dd760b46), published at [42cf7d2](https://github.com/PierrunoYT/zero/commit/42cf7d2ae5b12a72acab0b6ea7f905ba10f2baeb).

- Verified the earlier pre-clamp/stale-region data-loss findings are already fixed by the exact rendered-region guard; retained those protections and regression cases.
- Added a same-length, valid-bounds in-region replacement regression for both immediate cancel and a delayed partial followed by cancel. This specifically pins rendered-region equality rather than short-circuiting on invalid bounds.
- Recompute `needsLeadingSpace` when re-anchoring, keeping the separator inside the new tracked region so cancel removes it too. Updated stale-bounds and backspace expectations; the existing whitespace-terminated shrink case still verifies no extra separator.
- Integrated current upstream main [aadb4a2](https://github.com/Gitlawb/zero/commit/aadb4a27e9cdb41e774e8e1a8f23cf9c2fb50db2). Preserved the original remote PR head as an ancestor via a merge, verified the resulting tree exactly equals the validated tree, and used a normal fast-forward push (no force-push).

Regression proof:
- Without separator recomputation, updated tests fail with `stale live region replaced user text: "abnext"`, `partials replaced user text after backspace: "hello worthere friend again"`, and `composer = "abpartial text", want "ab partial text"`.
- Temporarily replacing only rendered-region equality with `true` fails both new subtests: `cancel deleted same-length user edit: "hello world"` and `partial deleted same-length user edit: "hello world next"`. Restored the guard before validation.

Validation on the published tree: `make fmt-check`, `go vet ./...`, `go test ./...`, `go test ./internal/tui -race -run Dictation -count=1`, `go run ./cmd/zero-release build`, `go run ./cmd/zero-release smoke`, `make lint-static` (0 issues), `make vulncheck` (No vulnerabilities found), and `git diff HEAD --check` all passed. Full-suite fixture tests used process-local `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false`: the first run inherited orb signing without a fixture key, including the plugins test. The isolated retry passed. Initial resource-constrained OpenAI timing failures also passed on retry; advisory lint was rerun successfully with a memory cap after a stalled first attempt.

These are composer-state interaction changes, verified through state assertions, not layout/style changes. Local validation is Linux; cross-platform GitHub CI and automated review are running. No PR merge performed.

jatmn (dismissed):
lgtm

Vasanthdev2004 (dismissed):
Re-approving at `42cf7d2a`. The commits since my last approval are the two things I listed as non-gating, plus a merge of main.

`TestDictationStreamingSameLengthRegionEditPreserved` pins the clause at `dictation_stream.go:137` that nothing pinned before: a same-length overtype of the live region with valid bounds now survives both a later partial and cancel, which is the safe direction I described. And the re-anchor branch runs `needsLeadingSpace`, so the separator is no longer orphaned after a stale region is abandoned; the three re-anchor expectations changed from `abpartial text` to `ab partial text` to match. Nothing else moved.

`internal/tui` green here apart from the long-cwd title bar test that fails on main on this machine; CI 6 of 6 at this head.

jatmn (dismissed):
LGTM

jatmn (approved):
lgtm

Vasanthdev2004 (approved):
Re-approving at `559fd362`. Since my last approval the branch gained a merge of main and one commit.

The merge is clean in the strict sense: its tree is identical to an automatic merge of its two parents, so nothing was resolved by hand. The three dictation files have not changed since `42cf7d2a`, so the sequence I drove then (type, dictate, backspace across the live region while partials arrive, cancel) still stands, and the dictation tests are green.

`559fd362` is a ride-along outside this PR's title. `browserTitleTextSafe` now also rejects whitespace runes in the decoded origin, which closes the no-break space, em space and ideographic space spellings of the same trick the format and bidi checks were there for. It is applied to the origin only, so a path with an encoded space is unaffected. Taking the check back out fails all three new URLs on both assertions, before translation and after the wire round trip. The extra `--reasoning-effort` assertion in `resume_model_test.go` passes. Both would have been cleaner as their own PRs, but they are small and correct, so I am not going to argue the point.

Merged onto current main, which is fourteen commits ahead: build and vet clean, `acp`, `specialist` and the dictation tests green. CI 9 of 9 at this head.

For whoever merges: the changes-requested from anandh8x dates from `dc80cecd` and is about the pre-comparison clamp, which was removed two heads ago along with the whole shift switch. Nothing in it applies to this head.

anandh8x (changes requested):
.

anandh8x (changes requested):
Follow-up after rechecking the additional files: I am withdrawing the scope/code objection.

The ACP change is correct: it rejects Unicode whitespace only from the presentation-safe browser origin used in titles, leaves execution URL normalization unchanged, and covers encoded whitespace through both title construction and ACP wire round trips. The specialist change is test-only and correctly pins the existing `ParentReasoningEffort` forwarding at the `runResume` call site.

I also merged this head into current `main` (`99721c762`) locally. The merge is automatic and conflict-free; diff hygiene, focused dictation race tests, ACP tests, specialist tests, and vet all pass on the merged tree. The broader TUI package has one environment-sensitive `/add-dir` failure that reproduces unchanged on plain `main`, so it is not caused by this PR.

The only remaining request is to update the branch onto current `main`. `AGENTS.md` explicitly defines a fresh base as a hard review requirement even when Git can merge the trees without a textual conflict. Once the branch is updated and CI reruns, I see no remaining code blocker.

anandh8x (approved):
Re-reviewed at head 93e68c7dd. The branch is current with main, the effective diff remains the reviewed six-file change, and all CI checks pass across Linux, macOS, and Windows. Approved.

PR #988, opened 2026-08-27
fix(tools): preserve file encoding on overwrite

### LINKED ISSUE #967: fix(tools): write_file rewrites CRLF files and drops UTF-8 BOM
### Version / branch / commit

### Problem

`read_file` normalizes CRLF to LF before returning content. When the model sends that content back through `write_file`, the tool writes it verbatim without preserving the existing file's line-ending convention or UTF-8 BOM. A small logical change can therefore rewrite every line and remove an encoding marker required by some Windows toolchains.

### Reproduction

1. Create a CRLF PowerShell, `.resx`, or project file; optionally prefix it with UTF-8 BOM bytes `EF BB BF`.
2. Read it through `read_file`.
3. Make a small change and write the complete content with `write_file`.
4. Inspect the bytes or `git diff`.

### Expected behavior

`write_file` preserves an existing file's CRLF convention and BOM unless the caller explicitly changes them.

### Actual behavior

The file is rewritten with LF and the BOM is removed, producing whole-file diff churn and potentially changing how downstream Windows tools decode non-ASCII text.

### Relevant code

- `internal/tools/write_file.go:97-110`
- `internal/tools/line_reader.go:40-42`
- `internal/tools/edit_file.go:92-102` already preserves CRLF for the sibling edit path

### Suggested fix

Inspect `priorContent` for a BOM and dominant line-ending convention, translate the outgoing content consistently, and preserve the original BOM. Add byte-level regression tests for CRLF, LF, BOM+CRLF, and new-file behavior.

PierrunoYT:
## Summary

- preserve an existing UTF-8 BOM when `write_file` overwrites normalized content
- preserve the existing dominant line-ending convention and normalize mixed outgoing endings consistently
- retain explicit CRLF content for LF files and leave new-file bytes unchanged
- add byte-level regression coverage for LF, CRLF, BOM+CRLF, mixed endings, explicit encoding bytes, and new files

Before the fix, the regression rewrote CRLF as LF and removed the BOM.

Fixes #967

## Verification

- `go test ./internal/tools -count=1`
- `make fmt-check`
- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`

## Summary by CodeRabbit

- **New Features**
  - File overwrites now support explicit BOM and LF/CRLF line-ending options.
  - Explicit encoding options can override automatic preservation behavior.
  - Lone carriage-return line endings are normalized consistently.

- **Bug Fixes**
  - Repeated overwrites preserve selected encoding and line-ending settings.
  - Invalid options are rejected without modifying the target file.
  - Overwrites fail safely when the existing file cannot be read.
  - Empty-file writes and formatting comparisons handle encoding adjustments correctly.

- **Tests**
  - Added coverage for encoding options, repeated overwrites, empty content, invalid inputs, and restricted file permissions.


gnanam1990 (changes requested):
Reviewed exact head `3f47d7e8048a5e9223d758d815aad0ba884319fa`.

Third-party integration gate: clear. This PR changes only the existing `internal/tools` implementation/tests and adds no module, SDK, service, provider, plugin, vendored code, remote asset, or dependency.

### [Medium] Keep the full-file observation after transparent encoding preservation

`modelKnownContent` is captured before `preserveWriteFileEncoding`, but the equality gate at `internal/tools/write_file.go:128` compares it with the byte-restored content. Therefore every CRLF- or BOM-preserving overwrite takes the unequal branch even when format-on-write is disabled or is a no-op. `FileTracker.Record` has already cleared the old observation at line 127, and line 129 does not restore it. The next `write_file` overwrite (and similarly a subsequent edit into the file) is refused as “not read in this session,” although Zero just received and wrote the complete replacement.

I reproduced this on the PR head with a tracked two-line CRLF file: mark it fully seen, overwrite it with LF-normalized model content, then assert `tracker.SeenWhole(path)` and perform a second overwrite. The assertion fails immediately; without that assertion, the second overwrite is blocked by the unseen-file guard.

Please distinguish the deterministic encoding restoration from an external formatter rewrite. For example, retain the post-preservation bytes as the model-equivalent write baseline, compare the formatter result against that value, and restore whole-file coverage when only the transparent BOM/EOL transformation occurred. Add a regression covering two successive tracked writes (or write followed by edit) for CRLF and BOM+CRLF.

Validation performed:

- New byte-preservation tests: pass
- `go test ./internal/tools -count=1`: pass without the generated reproducer
- Focused `go test -race`: pass
- `go vet ./internal/tools`: pass
- `gofmt -d` and `git diff --check`: clean
- Generated FileTracker lifecycle regression: fail as described above
- All current GitHub checks: green

jatmn (changes requested):

## Merge readiness

  `internal/tools/write_file.go:101`
  This branch forked from `27b319ca`, while live `main` is `1b5db176` and now includes 13 changed files across active MCP/OAuth and TUI work. The current merge is mechanically clean, but the repository treats a stale base as a blocker: it can conceal integration regressions and leaves the review evidence tied to an outdated target.

  Rebase this branch onto the current `main`, preserve the intended encoding-restoration behavior when resolving any future overlap in `write_file`, then rerun the focused `internal/tools` tests plus the required project validation on the rebased head. This keeps the change scoped to the approved encoding fix while establishing a reviewable, current integration point.

jatmn (changes requested):

## Findings

- [P2] Fail closed when an existing file cannot be read
  `internal/tools/write_file.go:101`
  The overwrite path establishes that the target exists, but treats the subsequent `os.ReadFile` error as if there were no prior bytes. `priorBytes` remains nil, so `preserveWriteFileEncoding` is skipped and `os.WriteFile` still replaces the file. A write-only existing CRLF/BOM file can therefore be overwritten successfully with the model’s normalized bytes, losing its original EOL convention and BOM—the exact transformation this change is intended to avoid.

  The root cause is that capturing the existing bytes is both the source for the preview and a prerequisite for safe encoding restoration, yet the code makes that capture optional after it has committed to the existing-file overwrite path. Please make an unsuccessful prior-byte read a fail-closed write error before `os.WriteFile` (and add a regression for a writable-but-unreadable existing target). That preserves the new-file pass-through behavior while ensuring an existing file is never silently overwritten through the unpreserved fallback.

PierrunoYT:
Addressed in f33ec58f.

**Confirmed reachable.** The tracked-file guard above only re-reads when `FileTracker.Version()` reports a recorded version, and `Version()` returns `false` on a nil tracker — so on the `Run` path (no `RunOptions`) nothing verified the prior bytes. `os.ReadFile` failed silently, `priorBytes` stayed nil, `preserveWriteFileEncoding` was skipped, and `os.WriteFile` replaced the file. Verified empirically: with the fix reverted, the new test reports `ok — Overwrote example.txt (2 lines).` after writing normalized LF content over a BOM+CRLF file.

**Fix.** Once the overwrite path has committed to an existing target, capturing its bytes is no longer optional — they are both the diff source and the only evidence of the convention to restore. A failed read is now a write error before `os.WriteFile`:

```go
if existed {
	prev, rerr := os.ReadFile(absolutePath)
	if rerr != nil {
		return errorResult("Error writing file " + relativePath + ": cannot read the existing file to preserve its line endings and BOM: " + rerr.Error())
	}
	priorContent = string(prev)
	content = preserveWriteFileEncoding(prev, content)
}
```

New-file pass-through is unchanged (`existed` false skips the block), and the `priorBytes != nil` sentinel is gone with it.

**Regression.** `TestWriteFileToolFailsClosedWhenExistingTargetIsUnreadable` asserts both the write error and that the original BOM+CRLF bytes are left untouched on disk.

Since `chmod` cannot express write-only on Windows — where losing CRLF/BOM is the case this PR is about — the unreadable target is built behind a per-OS `makeFileWriteOnly` helper: `0o200` on `!windows`, and a protected owner-only DACL granting `FILE_GENERIC_WRITE` without `FILE_READ_DATA` on Windows (mask `0x170196`; `WRITE_DAC` is granted explicitly because the `OWNER_RIGHTS` ACE otherwise strips the owner's ability to restore the descriptor). The test skips if the environment still permits the read, which covers running as root.

`go test ./internal/tools/` passes in full on Windows, `gofmt`/`go vet` are clean, and `GOOS=linux` / `GOOS=darwin` vet confirms the non-Windows helper compiles.

jatmn (changes requested):

## Merge readiness

- [P1] Get the Windows check green before merge
  `internal/tools/exec_command_test.go:378`
  The current head is mergeable but GitHub reports the Windows smoke job failed, leaving the check suite blocked. The retained log shows the failure is the unchanged timing-sensitive `TestExecCommandForegroundServerReturnsSessionAndServesHTTP` not observing its listening address before the deadline, rather than one of this PR's new encoding tests, so this looks unrelated to the diff; please rerun the required job and investigate only if it reproduces. The Ubuntu, macOS, performance, security, and review jobs are green.

## Findings

- [P2] Add an explicit encoding-intent path instead of inferring solely from bytes
  `internal/tools/write_file.go:168`
  The current helper has only the existing bytes and the submitted `content`. That is insufficient to distinguish the two cases this tool must support: normalized line-mode `read_file` output omits the BOM and changes CRLF to LF, but a caller intentionally removing a BOM or converting CRLF to LF submits the same byte shape. The helper resolves that ambiguity by always restoring the old BOM and CRLF convention. As a result, an empty full-file replacement of a BOM file leaves the three BOM bytes on disk, and an exact LF replacement of a CRLF file reports success while writing CRLF. Both operations worked on `main`, and both contradict the approved issue's requirement to preserve these features “unless the caller explicitly changes them.”

  Please address the ambiguity at the API/intent boundary rather than adding more content heuristics. Provide an unambiguous overwrite intent—whether through a narrowly scoped option or another explicit signal—that lets the caller independently request the BOM and line-ending outcome. Default behavior must continue to preserve an existing BOM and dominant EOL convention for ordinary normalized `read_file` round trips. The implementation should support at least these independent outcomes without guessing:

  - preserve both BOM and EOL convention by default;
  - remove a BOM while preserving the existing EOL convention;
  - convert CRLF to LF while preserving the existing BOM choice;
  - explicitly add a BOM or convert LF to CRLF, which the current patch already supports;
  - write an actually empty file when empty content and explicit BOM removal are requested.

  Keep the fix bounded to encoding intent. Do not change new-file byte passthrough, the fail-closed unreadable-target behavior, conflict detection, tracker observation semantics, mixed-ending normalization, or the existing opt-in formatter precedence. Add table-driven byte assertions for the default and explicit cases above, including the combined BOM+CRLF case, and verify two successive tracked writes so an override does not regress the already-fixed observation lifecycle.

## Overall guidance

There is one code finding on the current head. The earlier whole-file-observation and unreadable-existing-target requests are addressed. The repeated review rounds came from treating each downstream symptom separately while the producer contract remained ambiguous: `read_file` intentionally exposes a normalized view, whereas `write_file` also promises a full-file replacement. Once the exact same LF/no-BOM payload can mean either “round-trip the normalized view” or “change the encoding,” no byte-counting rule can recover intent reliably.

Please define that precedence once at the tool boundary and encode it in a compact behavior table before changing the transformation helper. A useful invariant is: explicit encoding intent wins; otherwise existing-file overwrites preserve the hidden convention; new files retain caller bytes; formatter behavior remains governed by the existing format-on-write contract. Testing that matrix end to end—from arguments through persisted bytes and tracker state—should close the remaining gap without expanding this PR into formatter, atomic-write, or broader file-tool redesign work.

PierrunoYT:
PierrunoYT addressed the remaining encoding-intent request in [2a9174d](https://github.com/PierrunoYT/zero/commit/2a9174d). Published head [04c4bd2](https://github.com/PierrunoYT/zero/commit/04c4bd2) includes current upstream main and preserves the previous remote head's ancestry; the push was fast-forward, not forced. The ancestry-preserving merge has exactly the same tree as the validated implementation.

`write_file` now exposes independent overwrite-only `bom: auto|add|remove` and `line_endings: auto|lf|crlf` options. Explicit intent wins before optional formatting; omitted options retain the previous automatic preservation behavior. New-file content remains byte-for-byte passthrough even when options are supplied. Unreadable-target rejection, conflict checks, mixed-ending defaults, formatter precedence, and tracker recording are unchanged.

The end-to-end byte matrix covers default BOM+CRLF preservation, BOM removal while keeping CRLF, LF conversion with/without BOM, BOM addition, CRLF conversion, combined overrides, and genuinely empty output with explicit BOM removal. Every matrix case verifies two successive tracked writes and whole-file observation. Invalid options are rejected without changing the target.

Regression proof: running the new tests with the pre-fix implementation via Go's `-overlay` fails at persisted-byte assertions, including `written bytes = "\ufeffnew\r\n", want "new\r\n"`, `want "\ufeffnew\n"`, and `written bytes = "\ufeff", want ""`; invalid intent was also accepted. The same tests pass with the fix.

Local verification: `make fmt-check`, `go vet ./...`, full `go test ./...`, release build/smoke, focused write-file `-race` tests, Darwin/Windows tools-test cross-compilation, and `git diff HEAD --check` pass. `make lint-static`: **0 issues**. `make vulncheck`: **No vulnerabilities found**. The existing unreadable-target regression passed without skipping. Fixture tests initially failed because the orb enforces Git signing without a signing key; the full suite passes with process-local `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false` (no repository/global signing configuration changed).

All seven current-head checks are now green: native Windows, macOS, Ubuntu, performance, security/code health, CodeRabbit, and Zero Review. [Fresh CI run](https://github.com/Gitlawb/zero/actions/runs/34150299483): the Windows test/build/smoke job passed, clearing the previous foreground-server timeout blocker. No PR merge performed.

The new CodeRabbit nitpick about `internal/specialist/resume_model_test.go` concerns unchanged code imported from upstream main, not this PR's encoding diff; it is intentionally left outside this focused fix.

jatmn (dismissed):
lgtm @Vasanthdev2004

jatmn (dismissed):
lgtm

PierrunoYT:
Resolved the conflicts and pushed the rebased branch at `93df93b4` (rebased onto `upstream/main` at `99721c76`).

- Preserved upstream write-safety, formatter-state tracking, and structured-diff handling alongside this PR's BOM/line-ending preservation and explicit encoding options.
- Removed three duplicate historical patches after confirming matching stable patch IDs.
- Updated the upstream unreadable-preimage test to match this PR's fail-closed contract: reject the overwrite, preserve original bytes, and emit no file diff.

Validation on Windows:
- PASS: `go test ./internal/tools`, including all `TestWriteFileTool` tests.
- PASS: `go vet ./...`.
- PASS: `go run ./cmd/zero-release build` and `go run ./cmd/zero-release smoke`.
- PASS: pinned govulncheck v1.3.0 — no vulnerabilities found (successful retry after a Windows executable-cleanup access error).
- PASS: changed-file formatting and `git diff HEAD --check`.

Global validation exceptions, left outside this conflict-resolution scope as directed:
- `make` is unavailable; ran the Makefile's formatting, pinned lint, and vulnerability commands directly.
- Repository formatting reports 10 benchmark fixture files under `internal/perfbench/testdata/`, all unchanged from upstream main.
- Advisory lint reports seven findings outside these edits (six staticcheck, one unused).
- The full `go test ./...` run failed in CLI/config tests due to missing active provider `chatgpt` and provider-command timeouts, in MCP due to `TerminateProcess: Access denied`, and in OAuth due to a secret-lock timeout. That run also caught the now-corrected unreadable-preimage test; the entire tools package passed on rerun. The full suite is not claimed green.

No unrelated local files were included in the push.

jatmn (changes requested):

## Merge readiness

- [P3] Align stale fail-closed comment in unreadable-target test helper  
  `internal/tools/write_file_unreadable_other_test.go:10`  
  The helper comment still describes a pre-PR shape where an overwrite could succeed without reading prior bytes. Current `write_file` fail-closed behavior and `TestWriteFileToolFailsClosedWhenExistingTargetIsUnreadable` assert the opposite. Update the comment so test infrastructure matches the encoding-preservation contract (no production change required).

## Findings

- [P2] Restore whole-file observation when format-on-write follows transparent encoding preservation  
  Attribution: PR-introduced. Merge-base overwrites wrote the model’s LF payload, so `modelKnownContent` still matched post-`gofmt` bytes and `RecordSeenRange` ran. This PR correctly sets `modelEquivalentContent` after `preserveWriteFileEncoding`, but the post-format equality gate at `internal/tools/write_file.go:171` compares that CRLF/BOM-restored string to formatter output; when `ZERO_FORMAT_ON_WRITE=1`, formatters such as `gofmt` normalize line endings back to LF, the comparison fails, and `SeenWhole` is cleared even though the model already performed a whole-file `read_file` and the on-disk text matches the submitted content.  
  Stated contract: prior review required restoring whole-file coverage for transparent BOM/EOL restoration and successive tracked overwrites; regressions assert `"transparent encoding preservation discarded the whole-file observation"` / `"encoding intent discarded whole-file observation"` with formatter disabled (`TestWriteFileToolEncodingPreservationKeepsWholeFileObservation`, `TestWriteFileToolExplicitEncodingIntent`).  
  Root cause: the `RecordSeenRange` gate treats formatter line-ending normalization as a non-transparent change relative to `modelEquivalentContent`, but the model’s preimage is the pre-preserve `content` argument (LF from `read_file`). Preservation plus optional formatting is one logical overwrite, not grounds to mark the file unseen.  
  What fails: with `ZERO_FORMAT_ON_WRITE` enabled, a tracked CRLF or BOM+CRLF file can be read whole, overwritten with normalized model text, succeed, and still block the next overwrite with the unseen-file guard until another exact read — a regression that does not occur on merge-base for the same inputs.  
  - `write_file.go` post-format `RecordSeenRange` gate (~171–172) — must recognize encoding-preserving overwrites that survive optional format-on-write  
  - regression test covering CRLF (and ideally BOM+CRLF) with `ZERO_FORMAT_ON_WRITE=1`, whole-file `read_file`, overwrite, and a second overwrite without an intervening read  
  Unchanged on main (do not edit in this PR): formatter command table, env toggle semantics, and `maybeFormatWrittenFileScoped` ordering (encoding still runs before formatting).  
  Required correction: extend the seen-range decision so opt-in formatting cannot drop `SeenWhole` when the final formatted bytes match the model-supplied overwrite content (or an equivalent line-normalized comparison), while keeping the existing gate for genuine formatter edits. Add the regression above.  
  Author fix: close the Root cause on every in-diff row in one pass; do not patch only the equality check without the formatter-enabled regression.  
  Out of scope: disabling format-on-write, moving formatting before encoding preservation, or changing unrelated tools.

Vasanthdev2004 (changes requested):
First review from me, at `e0c8647f`, driven on Windows through the real tool.

The core of it is right, and it is the version I would keep of the two PRs on this. Counting both endings over the whole file and taking the dominant one is the correct rule, and it holds where a window-based scan does not: a file that is CRLF throughout but whose first CRLF sits past byte 4096 is still recognised here. BOM preservation, the explicit `bom` and `line_endings` overrides, and failing closed when the existing file cannot be read all behave as described:

```
prior                     asked          this head            main
CRLF                      LF             CRLF                 LF
BOM + CRLF                LF             BOM + CRLF           LF
1 CRLF, 3 LF (mixed)      LF             LF                   LF
3 CRLF, 1 LF (mixed)      LF             CRLF                 LF
CRLF only, first at 5000  LF             CRLF                 LF
CRLF, line_endings=lf     LF             LF                   LF
BOM, bom=remove           LF             no BOM               no BOM
```

**What blocks it is @jatmn's open P2, and it is worse than the seen-range half.** With `ZERO_FORMAT_ON_WRITE=1` the preservation does not survive to disk at all. gofmt normalizes endings, it runs after `preserveWriteFileEncoding`, and it is the last writer, so a CRLF Go file comes out LF anyway. Then the comparison against `modelEquivalentContent` fails and the file is marked unseen. A tracked CRLF file, read whole, overwritten, then overwritten again:

```
                      first write   CRLF kept?   second write
main                  ok            n/a          ok
this head, format off ok            yes          ok
this head, format on  ok            NO           refused, not read exactly in this session
```

So with formatting enabled the user gets the same LF file they got before, plus a session that now refuses the next write. The refusal is a regression against main; the silent loss is the feature not happening. Patching only the equality gate would leave the second half: the bytes on disk are still LF while the tool reports it preserved them.

I think that says the transformation is in the wrong place rather than that the comparison is wrong. Preservation has to be the last thing that touches the bytes, so it needs to re-apply to the formatter's output rather than run before it. That also makes the seen-range question disappear, because the published bytes are then the preserved form of exactly what the model sent.

Two notes, neither blocking.

Lone CRs inside the content are converted for every existing-file overwrite, and `line_endings: "lf"` does not opt out of it, because the `\r` to `\n` replacement runs before the branch that reads the option. `col1\rcol2\n` lands as `col1\ncol2\n` on every setting. main writes it through. I could find only one file in this repository with a lone CR and it is a PNG, so I am not treating it as a real loss, but an explicit `lf` that still rewrites a byte the caller asked to keep is worth one line to fix.

A binary prior file with CRLF-dominant bytes converts the incoming content's endings too; #1064 skips that case on a NUL check. Since the file is being replaced with text anyway I do not think it matters, and I would not add the check just for symmetry. Mentioning it so the difference between the two PRs is on the record rather than assumed to be a gap.

CI is 9 of 9 at head, and the tools package passes natively here. I have not reviewed the test files line by line; the behaviour above is all driven through `write_file` itself.

On the overlap with #1064: I agree with @gnanam1990 that this is the one to finish. I have said the same on that PR.

PR #991, opened 2026-08-28
fix(mcp): discover OAuth through protected resource metadata

anandh8x:
## What changed

- Probe the configured MCP endpoint for an HTTP 401 OAuth challenge.
- Parse the RFC 9728 `resource_metadata` parameter from `WWW-Authenticate`.
- Fetch protected-resource metadata, validate its `resource`, and use its first `authorization_servers` issuer for RFC 8414 discovery.
- Preserve direct authorization-server discovery for legacy providers such as Linear.
- Preserve explicit issuer and endpoint overrides, including the no-network fast path when authorization and token endpoints are configured.
- Reject malformed challenges, resource mismatches, issuer mismatches, insecure metadata URLs, and insecure authorization-server URLs.
- Close streaming probe responses immediately so discovery cannot hang on an SSE endpoint.

## Security

- Discovery requests contain no authorization header, cookie, token, authorization code, client secret, or PKCE verifier.
- Protected-resource and authorization-server endpoints retain the existing HTTPS/loopback validation.
- Error paths do not echo metadata response bodies or sensitive OAuth values.

## Verification

- Notion-style protected-resource discovery and complete OAuth callback/token flow.
- Linear-style direct discovery fallback, including a streaming MCP endpoint.
- Explicit issuer and endpoint override behavior.
- Malformed, conflicting, insecure, empty, and mismatched metadata.
- OAuth, MCP, and MCP OAuth CLI package tests.
- Focused OAuth/MCP/CLI race tests.
- `make fmt-check`
- `go vet ./...`
- `make lint-static`
- `make vulncheck`
- `git diff HEAD --check`

Zero's resolver was also exercised read-only against `https://mcp.notion.com/mcp` and resolved the live issuer, authorization, token, and registration endpoints. No client registration or account authorization was performed.

`go test ./...` reaches the existing `TestAltScreenTranscriptScrollKeepsFooterFixed` failure, which reproduces unchanged on current main; all other packages pass.

## Summary by CodeRabbit

* **New Features**
  * Added protected-resource metadata discovery for OAuth.
  * Authorization server discovery now prioritizes advertised metadata and falls back to existing methods.
  * OAuth login supports PKCE and avoids sending credentials during discovery.

* **Security**
  * Restricted OAuth discovery to public, non-private network destinations.
  * Added validation for metadata, resource identifiers, authorization servers, and issuer consistency.
  * Prevented unexpected redirects, cookie transmission, and credential leakage during discovery.
  * Constrained discovered registration and token requests to the validated protected-resource origin.

### DIFF
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -63,8 +63,9 @@ func tokenToStored(t oauth.Token) StoredToken {
 }
 
 // OAuthConfig describes how to authenticate to a remote MCP server using OAuth.
-// Endpoints may be discovered from the server's metadata document; explicit
-// values here override or fill in anything discovery cannot provide.
+// Endpoints may be discovered through RFC 9728 protected-resource metadata and
+// RFC 8414 authorization-server metadata; explicit values override or fill in
+// anything discovery cannot provide.
 type OAuthConfig struct {
 	ClientID              string
 	ClientSecret          string
@@ -153,9 +154,9 @@ func validateResolvedEndpoints(metadata authServerMetadata) error {
 	return nil
 }
 
-// resolveAuthorizationServer discovers metadata and applies explicit config
-// overrides. Configured endpoints take precedence over discovered ones, and act
-// as a fallback when discovery fails or omits a value.
+// resolveAuthorizationServer follows protected-resource metadata when the MCP
+// endpoint advertises it, falls back to direct authorization-server discovery
+// for legacy providers, then applies explicit config overrides.
 func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseURL string, cfg OAuthConfig) (authServerMetadata, error) {
 	// When the config supplies both the authorization and token endpoints
 	// directly, skip network discovery entirely: there is nothing to discover,
@@ -175,8 +176,18 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 	}
 
 	discoveryBase := strings.TrimSpace(cfg.IssuerURL)
+	protectedResourceDiscovery := false
 	if discoveryBase == "" {
-		discoveryBase = baseURL
+		issuer, found, protectedErr := discoverProtectedResourceAuthorizationServer(ctx, client, baseURL)
+		if protectedErr != nil {
+			return authServerMetadata{}, protectedErr
+		}
+		if found {
+			discoveryBase = issuer
+			protectedResourceDiscovery = true
+		} else {
+			discoveryBase = baseURL
+		}
 	}
 
 	metadata, err := discoverAuthorizationServer(ctx, client, discoveryBase)
@@ -184,6 +195,9 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 		// Discovery failures are non-fatal when the config supplies the endpoints
 		// directly; otherwise surface the discovery error.
 		metadata = authServerMetadata{}
+	}
+	if protectedResourceDiscovery && strings.TrimSpace(metadata.Issuer) != "" && strings.TrimSpace(metadata.Issuer) != strings.TrimSpace(discoveryBase) {
+		return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata issuer does not match protected resource metadata")
 	}
 
 	if endpoint := strings.TrimSpace(cfg.AuthorizationEndpoint); endpoint != "" {
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -121,7 +121,7 @@ type authorizationFlow struct {
 // discoverAuthorizationServer fetches the RFC 8414 authorization server metadata
 // at the well-known path under baseURL, via the shared engine.
 func discoverAuthorizationServer(ctx context.Context, client *http.Client, baseURL string) (authServerMetadata, error) {
-	meta, err := oauth.DiscoverAuthorizationServer(ctx, client, baseURL)
+	meta, err := oauth.DiscoverAuthorizationServer(ctx, withoutDiscoveryRedirects(client), baseURL)
 	if err != nil {
 		return authServerMetadata{}, err
 	}
@@ -196,8 +196,14 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 		// directly; otherwise surface the discovery error.
 		metadata = authServerMetadata{}
 	}
-	if protectedResourceDiscovery && strings.TrimSpace(metadata.Issuer) != "" && strings.TrimSpace(metadata.Issuer) != strings.TrimSpace(discoveryBase) {
-		return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata issuer does not match protected resource metadata")
+	if protectedResourceDiscovery {
+		issuer := strings.TrimSpace(metadata.Issuer)
+		if issuer == "" {
+			return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata is missing its required issuer")
+		}
+		if issuer != strings.TrimSpace(discoveryBase) {
+			return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata issuer does not match protected resource metadata")
+		}
 	}
 
 	if endpoint := strings.TrimSpace(cfg.AuthorizationEndpoint); endpoint != "" {
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -175,33 +175,37 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 		return metadata, nil
 	}
 
+	discoveryClient := withoutDiscoveryRedirects(client)
 	discoveryBase := strings.TrimSpace(cfg.IssuerURL)
 	protectedResourceDiscovery := false
 	if discoveryBase == "" {
-		issuer, found, protectedErr := discoverProtectedResourceAuthorizationServer(ctx, client, baseURL)
+		issuer, found, protectedErr := discoverProtectedResourceAuthorizationServer(ctx, discoveryClient, baseURL)
 		if protectedErr != nil {
 			return authServerMetadata{}, protectedErr
 		}
 		if found {
 			discoveryBase = issuer
 			protectedResourceDiscovery = true
+			discoveryClient, protectedErr = newAdvertisedOAuthDiscoveryClient(discoveryClient, baseURL)
+			if protectedErr != nil {
+				return authServerMetadata{}, protectedErr
+			}
 		} else {
 			discoveryBase = baseURL
 		}
 	}
 
-	metadata, err := discoverAuthorizationServer(ctx, client, discoveryBase)
+	metadata, err := discoverAuthorizationServer(ctx, discoveryClient, discoveryBase)
 	if err != nil {
 		// Discovery failures are non-fatal when the config supplies the endpoints
 		// directly; otherwise surface the discovery error.
 		metadata = authServerMetadata{}
 	}
 	if protectedResourceDiscovery {
-		issuer := strings.TrimSpace(metadata.Issuer)
-		if issuer == "" {
+		if strings.TrimSpace(metadata.Issuer) == "" {
 			return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata is missing its required issuer")
 		}
-		if issuer != strings.TrimSpace(discoveryBase) {
+		if metadata.Issuer != discoveryBase {
 			return authServerMetadata{}, errors.New("mcp oauth: authorization server metadata issuer does not match protected resource metadata")
 		}
 	}
--- a/internal/mcp/oauth.go
+++ b/internal/mcp/oauth.go
@@ -86,6 +86,11 @@ type authServerMetadata struct {
 	TokenEndpoint         string   `json:"token_endpoint"`
 	RegistrationEndpoint  string   `json:"registration_endpoint"`
 	ScopesSupported       []string `json:"scopes_supported"`
+
+	protectedResourceURL         string
+	protectAuthorizationEndpoint bool
+	protectTokenEndpoint         bool
+	protectRegistrationEndpoint  bool
 }
 
 // pkceParams holds a PKCE verifier/challenge pair.
@@ -134,10 +139,9 @@ func discoverAuthorizationServer(ctx context.Context, client *http.Client, baseU
 	}, nil
 }
 
-// validateResolvedEndpoints applies the shared https/loopback endpoint rule to
-// every non-empty endpoint in the resolved metadata — configured or discovered
-// alike — so discovery metadata can never downgrade an MCP login to an insecure
-// or attacker-controlled authorization, token, or registration endpoint.
+// validateResolvedEndpoints applies the shared https/loopback scheme rule to
+// every non-empty endpoint in the resolved metadata. Protected-resource
+// discovery adds its stricter provenance-aware network rule below.
 func validateResolvedEndpoints(metadata authServerMetadata) error {
 	for _, ep := range []struct{ kind, url string }{
 		{"authorization", metadata.AuthorizationEndpoint},
@@ -149,6 +153,40 @@ func validateResolvedEndpoints(metadata authServerMetadata) error {
 		}
 		if err := oauth.ValidateEndpointURL(ep.url); err != nil {
 			return fmt.Errorf("mcp oauth: %s endpoint: %w", ep.kind, err)
+		}
+	}
+	return nil
+}
+
+func validateProtectedDiscoveredEndpoints(ctx context.Context, metadata authServerMetadata) error {
+	if metadata.protectedResourceURL == "" {
+		return nil
+	}
+	policy, err := newOAuthDiscoveryNetworkPolicy(metadata.protectedResourceURL)
+	if err != nil {
+		return err
+	}
+	for _, endpoint := range []struct {
+		kind      string
+		url       string
+		protected bool
+	}{
+		{kind: "authorization", url: metadata.AuthorizationEndpoint, protected: metadata.protectAuthorizationEndpoint},
+		{kind: "token", url: metadata.TokenEndpoint, protected: metadata.protectTokenEndpoint},
+		{kind: "registration", url: metadata.RegistrationEndpoint, protected: metadata.protectRegistrationEndpoint},
+	} {
+		if !endpoint.protected || strings.TrimSpace(endpoint.url) == "" {
+			continue
+		}
+		parsed, parseErr := url.Parse(endpoint.url)
+		if parseErr != nil || parsed.Hostname() == "" {
+			return fmt.Errorf("mcp oauth: %s endpoint: %w", endpoint.kind, errUnsafeOAuthDiscoveryTarget)
+		}
+		if err := policy.validateLiteralHost(parsed.Hostname()); err != nil {
+			return fmt.Errorf("mcp oauth: %s endpoint: %w", endpoint.kind, err)
+		}
+		if _, err := policy.resolve(ctx, parsed.Hostname()); err != nil {
+			return fmt.Errorf("mcp oauth: %s endpoint: %w", endpoint.kind, err)
 		}
 	}
 	return nil
@@ -219,6 +257,12 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 	if endpoint := strings.TrimSpace(cfg.RegistrationEndpoint); endpoint != "" {
 		metadata.RegistrationEndpoint = endpoint
 	}
+	if protectedResourceDiscovery {
+		metadata.protectedResourceURL = strings.TrimSpace(baseURL)
+		metadata.protectAuthorizationEndpoint = strings.TrimSpace(cfg.AuthorizationEndpoint) == ""
+		metadata.protectTokenEndpoint = strings.TrimSpace(cfg.TokenEndpoint) == ""
+		metadata.protectRegistrationEndpoint = strings.TrimSpace(cfg.RegistrationEndpoint) == ""
+	}
 
 	if strings.TrimSpace(metadata.AuthorizationEndpoint) == "" {
 		return authServerMetadata{}, errors.New("no authorization endpoint discovered or configured")
@@ -227,6 +271,9 @@ func resolveAuthorizationServer(ctx context.Context, client *http.Client, baseUR
 		return authServerMetadata{}, errors.New("no token endpoint discovered or configured")
 	}
 	if err := validateResolvedEndpoints(metadata); err != nil {
+		return authServerMetadata{}, err
+	}
+	if err := validateProtectedDiscoveredEndpoints(ctx, metadata); err != nil {
 		return authServerMetadata{}, err
 	}
 	return metadata, nil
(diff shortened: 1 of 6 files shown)


Vasanthdev2004 (changes requested):
The RFC 9728 shape is right and the parts that usually get skipped are here: the resource identifier is bound back to the MCP server URL, the issuer is bound to the protected-resource metadata, the probe sends no credentials, and the body is size-limited. The `WWW-Authenticate` parser is careful too, handling quoted strings, rejecting control characters, and refusing conflicting `resource_metadata` values rather than picking one.

One blocker, and it undoes most of the above.

### Redirects walk straight past the endpoint validation

`probeProtectedResourceMetadataURL` and `fetchProtectedResourceMetadata` both call `client.Do` on the caller's client, which is `http.DefaultClient` by default and follows up to ten redirects. So the URL that gets validated is not the URL that answers.

Demonstrated against this head. A URL that `ValidateEndpointURL` explicitly rejects is fetched anyway, and its metadata is used:

```
ValidateEndpointURL(http://evil.invalid/m) = oauth: refusing to send credential to a non-https token endpoint: http://evil.invalid
requests attempted: [https://127.0.0.1:61865/.well-known/oauth-protected-resource  http://evil.invalid/m]
err=<nil>  authorization_servers=[https://attacker-chosen.example.com]
```

The probe follows redirects too, so the host that supplies the `WWW-Authenticate` challenge need not be the configured MCP server.

`internal/oauth/flow.go` already has `withoutRedirects` for this, and its comment is about exactly this hazard. Wrapping both calls in it is the fix.

Worth being clear about the ownership here: the `client.Do`-without-redirect-policy pattern is pre-existing in `oauth.DiscoverAuthorizationServer`, so this is not something you introduced. What changes with this PR is that the URL reaching it is now named by the server rather than derived from the configured base URL, and `discoveryBase` is now attacker-supplied as well, so the pre-existing path becomes attacker-directed. That one is worth a follow-up on its own.

### The issuer binding is skipped when the server omits `issuer`

```go
if protectedResourceDiscovery && strings.TrimSpace(metadata.Issuer) != "" && strings.TrimSpace(metadata.Issuer) != strings.TrimSpace(discoveryBase) {
```

An authorization server that simply omits `issuer` from its metadata skips the check entirely. RFC 8414 makes `issuer` REQUIRED and says the client must verify it, so absence is a failure rather than a pass. Under protected-resource discovery this should fail closed.

### Both comparisons are exact string matches

The resource check and the issuer check both compare raw strings, so these are unequal:

```
https://example.com/mcp     vs  https://example.com/mcp/
https://example.com/mcp     vs  https://Example.com/mcp
https://example.com:443/mcp vs  https://example.com/mcp
```

This fails closed, so it is not a security problem, but a legitimate server that spells its own identifier with a trailing slash or a default port will be rejected with a message pointing at a mismatch the operator cannot see. Comparing parsed URLs with the host lowercased and the default port normalized would avoid a confusing class of report.

Requesting changes for the redirect handling. The rest of the design is sound, and I would be happy to see this land once the fetches cannot be walked off the validated URL.

gnanam1990 (changes requested):
Reviewed exact head `a9e947601c54a5241ef695c082c6742a4f764e63` against base `eeea3308fed38b37b203522624babaf2f30df91f`.

Third-party integration gate: RFC 9728 protected-resource discovery and its RFC 8414 authorization-server hop were disclosed to and explicitly authorized by the maintainer before publication. No module or lockfile dependency changes are present.

Process gates: none. The author is a repository collaborator, the PR is mergeable on the current base, and exact-head CI is green.

### [High] Apply an SSRF boundary to every server-advertised discovery hop

**Location:** `internal/mcp/oauth_protected_resource.go:30,40-45,77-80,92`

The remote MCP resource controls both `resource_metadata` and `authorization_servers`. The only pre-fetch check is `oauth.ValidateEndpointURL`, which is a transport-scheme rule and deliberately permits cleartext loopback URLs. The fetches then use the caller's ordinary `http.Client`, which also follows redirects unless the caller overrides that behavior. Consequently, a remote or compromised HTTPS MCP server can make Zero issue GETs to loopback/private services directly, or walk a validated public URL to one through a redirect. The authorization-server issuer creates the same server-directed fetch in RFC 8414 discovery.

**Evidence A — demonstrated on this head:** I served the MCP resource over TLS and advertised `resource_metadata` at an HTTP `127.0.0.1` victim. Zero accepted the URL and the victim recorded one metadata request. No credentials were needed; the SSRF occurs before resource/issuer JSON validation can reject the response. RFC 9728 section 7.7 explicitly warns that malicious resource metadata opens SSRF and recommends blocking internal IP ranges: https://www.rfc-editor.org/rfc/rfc9728.html#name-server-side-request-forgery

Please apply one discovery-network policy to the probe redirect chain, protected-resource metadata URL, authorization-server issuer metadata request, and every redirect. For a non-loopback configured MCP resource, reject loopback, private, link-local, and other non-public resolved addresses; bind the check at dial/use time so DNS rebinding cannot bypass it. Either reject redirects or validate every redirect target with the same resolved-address policy. Preserve local development only when the configured resource itself is explicitly loopback. Add regressions for direct advertised loopback/private metadata and issuer URLs plus public-to-private redirects.

### [Medium] Require the RFC 8414 issuer and compare identity values exactly

**Location:** `internal/mcp/oauth.go:199-200`; related resource binding at `internal/mcp/oauth_protected_resource.go:34`

The protected-resource issuer check runs only when `metadata.Issuer` is non-empty. RFC 8414 makes `issuer` required and requires the returned value to exactly match the issuer used to construct the metadata request. On this head, a protected resource advertised an authorization server whose metadata omitted `issuer` but supplied authorization/token endpoints; `resolveAuthorizationServer` returned success and accepted both endpoints.

Fail closed when `issuer` is absent or differs. Do not normalize either received identity with `TrimSpace`: RFC 8414 issuer binding and RFC 9728 resource binding require identical strings, so whitespace/default-port/trailing-slash differences must remain mismatches. Add missing, whitespace-altered, and mismatched issuer/resource cases. RFC 8414 issuer requirements: https://www.rfc-editor.org/rfc/rfc8414.html#section-2

### Verification

- Generated SSRF regression: failed because the remote TLS resource triggered one loopback request.
- Generated missing-issuer regression: failed because metadata with `Issuer:""` was accepted with its endpoints.
- `go test -race ./internal/mcp -count=1`: pass without generated tests.
- Broad `internal/cli` race run reached two local doctor/config failures; the same tests pass on both base and head with isolated HOME/XDG roots, so they are not attributed to this PR.
- `make fmt-check`, focused `go vet`, and `git diff --check`: pass.
- Exact-head Linux/macOS/Windows smoke, CodeQL, security/code-health, and automated review checks: green.
- The probe sends no explicit bearer/cookie headers in the production CLI path, response bodies are bounded, the SSE probe body is closed promptly, and explicit endpoint overrides retain their no-network fast path.

The protocol shape is otherwise focused, but the newly server-directed network hops and mandatory identity binding must fail closed before merge.

gnanam1990 (changes requested):
Re-reviewed exact head `66543a4f96b2218a80ffb0e6746456311001a983` against base `eeea3308fed38b37b203522624babaf2f30df91f`, including the delta from the previously reviewed head `a9e947601c54a5241ef695c082c6742a4f764e63`.

Third-party integration gate: this delta adds no module, lockfile, provider, or new external protocol integration. It only changes the already disclosed and maintainer-authorized RFC 9728/RFC 8414 discovery flow.

The new no-redirect client is effective, the missing RFC 8414 issuer is now rejected, the branch is mergeable on the current base, and every exact-head CI check is green. Two demonstrated parts of the prior review still remain.

### Prior [High] — partially fixed: redirects are blocked, direct server-advertised SSRF remains

**Location:** `internal/mcp/oauth_protected_resource.go:41-45,55-60,92-95,102-108`; authorization-server hop through `internal/mcp/oauth.go:124,193`

`withoutDiscoveryRedirects` closes the redirect route and the three added redirect regressions pass. It does not add an address/trust boundary to either direct server-controlled destination. `resource_metadata` and the first `authorization_servers` value still pass only through `oauth.ValidateEndpointURL`; that validator intentionally accepts HTTP loopback and accepts any HTTPS host without classifying its resolved address.

**Reproduced on this head with two generated, hermetic tests:**

- A remote TLS MCP resource advertised an HTTP `127.0.0.1` protected-resource metadata URL. Zero fetched it once and returned success (`err == nil`).
- A remote TLS MCP resource served its metadata itself but advertised an HTTP `127.0.0.1` authorization-server issuer. Zero fetched that loopback issuer's RFC 8414 document once and returned success.

These are direct targets, so refusing redirects cannot affect either request. RFC 9728 section 7.7 specifically warns that a malicious resource server can create SSRF through the authorization-server metadata hop and recommends blocking internal IP ranges: https://www.rfc-editor.org/rfc/rfc9728.html#name-server-side-request-forgery

Please apply one discovery-only network policy to both advertised hops. For a non-loopback configured MCP resource, reject loopback, private, link-local, unspecified, multicast, and other non-public resolved destinations at connect/use time (including HTTPS hostnames resolving there). Preserve loopback development only when the configured MCP resource itself is loopback. Keep the new redirect refusal, and add direct metadata/issuer regressions.

### Prior [Medium] — partially fixed: issuer presence is enforced, identity comparison is still normalized

**Location:** `internal/mcp/oauth_protected_resource.go:49,55`; `internal/mcp/oauth.go:199-205`

The new missing-issuer check is correct and its regression passes. The binding comparisons still call `strings.TrimSpace` on received identity values, however. Generated tests returned success for both:

- protected metadata with `resource = " " + configuredResourceURL`;
- authorization-server metadata with `issuer = " " + advertisedIssuer`.

RFC 9728 requires the returned `resource` to be identical to the URL used for the resource request, and RFC 8414 requires the returned `issuer` to be identical to the issuer used to construct the metadata URL. Trimming is a transformation, not an identical-string comparison: https://www.rfc-editor.org/rfc/rfc9728.html#section-3.3 and https://www.rfc-editor.org/rfc/rfc8414.html#section-3.3

Compare the received `resource` and `issuer` strings byte-for-byte with their expected values. Keep whitespace trimming only for emptiness checks or operator-provided configuration, not for protocol identity binding. Add leading/trailing-whitespace mismatch regressions for both values.

### Verification

- Generated direct metadata SSRF: failed as expected; loopback victim hit once and the call returned success.
- Generated direct issuer SSRF: failed as expected; loopback victim hit once and the resolver returned success.
- Generated whitespace-altered resource and issuer bindings: both failed as expected because the resolver accepted them.
- Author regressions for probe/metadata/issuer redirects, missing issuer, and issuer mismatch: pass.
- `go test -race ./internal/mcp -count=1`: pass.
- Focused MCP OAuth CLI race tests with isolated HOME/config/cache roots: pass.
- `make fmt-check`, `go vet ./internal/mcp`, and `git diff HEAD --check`: pass.
- Exact-head Linux, macOS, Windows, performance, security/code-health, CodeQL, automated-review, and CodeRabbit checks: green.

The patch is materially closer, but the two remaining fail-closed boundaries need correction before approval.

Vasanthdev2004 (dismissed):
Fixed, and you went further than I asked: the pre-existing `client.Do` in `oauth.DiscoverAuthorizationServer` is wrapped too, which I had called a separate follow-up.

I re-ran my exploit against `10799398`. It is dead, and the host that used to answer never gets a request:

```
[redirect]      err=mcp oauth: protected resource metadata returned HTTP 302   unvalidated host hits=0
[cross-origin]  err=... metadata URL must be same-origin with the MCP server URL   other-origin hits=0
[legit]         issuer="https://as.example.com" found=true err=<nil>
```

`validateResourceMetadataOrigin` is the better fix and I had only hinted at it. Requiring the metadata URL to be same-origin with the configured MCP server removes the arbitrary-host hop at the source, rather than trying to sanitise it afterwards. Returning `ErrUseLastResponse` so a 3xx surfaces as a refused status, and clearing the cookie jar so an unauthenticated metadata GET cannot pick up credentials, are both good calls.

The pinning transport is the right shape for the remaining hop: resolve, classify the resolved address, connect to the pinned IP, and keep the original name for SNI and Host so certificate verification still means something. That closes the DNS-rebinding gap a literal-host check would leave open.

## On the two findings still marked outstanding

@gnanam1990's review is against `66543a4f`, the first of the two commits. Everything it reports was fixed by `10799398`, which is the commit that adds the network policy and drops the trims. I re-ran both, restated as written, on the current head:

```
(1) loopback metadata URL from a remote TLS resource:
    err=... must be same-origin with the MCP server URL   victimHits=0
(2) resource with a leading space:
    err=... resource does not match the MCP server URL
(3) issuer with a leading space:
    err=... issuer does not match protected resource metadata
```

And the advertised-issuer SSRF case, which is the one the redirect fix could not reach:

```
authorization_servers=["https://169.254.169.254/"]
  -> err=mcp oauth: refused non-public server-advertised discovery target
```

So the address boundary is applied to both advertised hops, and the identity comparisons are byte-for-byte on the received side while trimming survives only on the operator-supplied side, which is what that review asked for. @gnanam1990, worth a re-look at `10799398` when you get a chance so this is not held on a stale head.

## One note, not blocking

`advertisedOAuthDiscoveryTransport.RoundTrip` requires the base to be an `*http.Transport` and errors otherwise. Production is fine, since `internal/cli/auth.go` passes a client with a nil transport and picks up `http.DefaultTransport`, and a caller whose transport fails the probe degrades to legacy discovery rather than breaking. But anyone who later injects an instrumented transport gets discovery failing with a message about policy enforcement rather than about their transport. Failing closed is the right direction; it just deserves the clearer error.

`go test -race ./internal/mcp` passes, vet and gofmt are clean, and it cross-compiles for linux, darwin and windows.

Approving.

gnanam1990 (changes requested):
Re-reviewed exact head `1079939848b4de9b7deb98f33b0abbe462e60c2e` against base `eeea3308fed38b37b203522624babaf2f30df91f`, including the complete delta from the previously reviewed head `66543a4f96b2218a80ffb0e6746456311001a983`.

Third-party integration gate: this delta adds no module, lockfile, provider, hosted service, or new external protocol. It hardens only the already disclosed and maintainer-authorized RFC 9728/RFC 8414 flow.

## Verdict

**CHANGES_REQUESTED**

## Process gates

None. The branch is mergeable on the current base and all exact-head CI checks are green.

## Findings

### [High] The public-address policy stops before server-discovered OAuth endpoints are used

**Location:** `internal/mcp/oauth.go:141-154,198-230,367-370`; policy client construction at `internal/mcp/oauth_discovery_network.go:62-76`
**Evidence:** A — demonstrated, including base/head differential.
**Invariant:** A non-loopback MCP resource must not be able to make Zero contact loopback/private services through any value learned from its protected-resource/authorization-server metadata chain.

The new client correctly confines the protected-resource and RFC 8414 metadata GETs. After RFC 8414 decoding, however, `resolveAuthorizationServer` returns the advertised authorization, token, and registration endpoints, `validateResolvedEndpoints` applies only `oauth.ValidateEndpointURL`, and `Login` uses the original unrestricted `httpClient` for dynamic registration. That validator deliberately permits HTTP loopback for local development without retaining whether the configured MCP resource was loopback.

**Trigger and path**

`remote MCP resource -> same-origin protected metadata -> public advertised issuer -> RFC 8414 registration_endpoint=http://127.0.0.1/... -> Login -> registerClient(original client)`

Dynamic registration runs before `OpenBrowser`, so no authorization click or callback is needed.

**Proof**

I generated a hermetic full-flow test with a public-literal TLS MCP resource (`93.184.216.34`) and a distinct public-literal TLS authorization server (`93.184.216.35`) routed to an isolated local gateway. The authorization-server document advertised an isolated HTTP loopback registration victim.

- Head `10799398`: the loopback victim recorded **one POST**, registration succeeded, and execution reached the deliberate post-registration stop.
- Base `eeea3308`: the identical protected-resource scenario recorded **zero hits** and never reached registration because the new protected-resource-to-advertised-AS path did not exist.

This is therefore introduced reachability, not a generic pre-existing observation. The same provenance gap also leaves discovered token/authorization endpoints outside the newly stated public-network boundary; registration is the demonstrated zero-interaction exploit. A public registration URL can additionally redirect the default client to an unchecked target.

**Fix direction**

Retain endpoint provenance. For endpoints learned through protected-resource discovery, reject non-public registration/token/authorization targets under the same resource-aware policy, and bind registration/token network use to validated resolved addresses. Preserve loopback only when the configured MCP resource itself is explicitly loopback; preserve explicit operator-configured endpoint overrides separately. Refuse registration redirects or enforce the same policy on every redirect. Add direct loopback/private, DNS-to-private, and public-to-private redirect regressions for dynamic registration; cover the token endpoint with the same matrix.

### [Medium] The pinning transport bypasses a configured HTTPS proxy and then fails discovery

**Location:** `internal/mcp/oauth_discovery_network.go:151-183`
**Evidence:** A — demonstrated under the race detector.
**Invariant:** A configured HTTPS proxy must remain the first network hop; target pinning must not turn a proxied request into a direct target dial.

`roundTripPinnedOAuthDiscovery` installs `DialTLSContext` that ignores the address supplied by `net/http` and always dials `pinnedAddress`. For an `https://` proxy, Go invokes that hook with the **proxy** address because the proxy is the TLS first hop. The hook discards it, directly dials the advertised target IP, completes target TLS, and `net/http` subsequently tries to speak proxy `CONNECT` over that target connection. This both creates direct egress outside the configured proxy and makes discovery fail.

**Proof**

A generated deterministic test configured `https://proxy.example:8443`, resolved the advertised target to `93.184.216.34`, and stopped at the base dialer. The observed dial was:

```text
got  93.184.216.34:443
want proxy.example:8443
```

The result reproduced unchanged under `go test -race`. HTTP-proxy behavior does not disprove this: the defect is specific to an HTTPS first hop, where `DialTLSContext` is selected.

**Fix direction**

Distinguish the proxy TLS first hop from the pinned target hop. Preserve the configured HTTPS proxy connection and certificate identity while ensuring its `CONNECT` destination is the validated pinned target and the nested target TLS handshake retains the original issuer hostname for SNI/certificate verification. Add focused direct, HTTP-proxy, HTTPS-proxy, and SOCKS-proxy tests that assert both the dialed first hop and the CONNECT/target identity.

## Prior-finding closure

- **Prior High — fixed:** redirects are refused; the metadata URL is bound same-origin; server-advertised metadata and issuer requests resolve/classify all addresses and dial the selected public address. Re-running the prior remote metadata and corrected public-resource/loopback-issuer proofs produced zero victim hits. The corrected issuer proof uses a public configured resource; an explicitly loopback resource is intentionally allowed to use loopback discovery for local development.
- **Prior Medium — fixed:** missing issuer is rejected; returned `resource` and `issuer` identities are compared without normalization. Leading/trailing whitespace resource and issuer regressions pass.
- **Cookie comment — fixed:** discovery-client copies clear the jar; focused cookie tests pass.

## What I verified

- Generated prior metadata/issuer SSRF closure tests under `-race`: pass, zero victim hits.
- Generated registration SSRF head test: fail as expected, one loopback POST.
- Identical base comparison at `eeea3308`: pass, zero loopback hits.
- Generated HTTPS-proxy first-hop test under `-race`: fail as expected; target IP dialed instead of proxy.
- Author tests for direct/private/DNS-at-use-time targets, address pinning, redirects, cookies, missing issuer, and exact resource/issuer identity: pass.
- `go test -race ./internal/mcp -count=1`: pass without generated tests.
- Focused MCP OAuth CLI race tests with isolated HOME/config/cache roots: pass.
- `make fmt-check`, `go vet ./internal/mcp`, and `git diff HEAD --check`: pass.
- Exact-head Linux/macOS/Windows smoke, performance, security/code-health, CodeQL, automated review, and CodeRabbit checks: green.

## Not verified

- No live provider login, dynamic client registration, token exchange, or browser authorization was performed.
- I did not run the repository-wide local `go test ./...`, build, smoke, static lint, or vulncheck commands in this re-review; exact-head CI supplies the reported cross-platform smoke/security status, while local verification was risk-focused.

## Review integrity

- Reviewed worktree unchanged: yes.
- Head re-confirmed immediately before publication: yes (`1079939848b4de9b7deb98f33b0abbe462e60c2e`).
- Evidence stale: no.

Vasanthdev2004 (approved):
Re-approving on `975eaeec`. The push dismissed my previous approval, correctly, since this commit changes the policy I had verified.

`validateProtectedDiscoveredEndpoints` closes a gap my earlier review did not cover. I checked the metadata fetches and stopped there; the authorization, token and registration endpoints those documents hand back were still only getting the https/loopback scheme rule. Now they get the provenance-aware network rule too, and only when protected-resource discovery supplied them, with `protect*Endpoint` false whenever the operator configured the endpoint explicitly. That distinction is the right one: an operator naming their own internal token endpoint is not the same as a server advertising one.

Verified on this head, and the new check is not vacuous:

```
[redirect]        err=... metadata returned HTTP 302          hits=0
[cross-origin]    err=... must be same-origin with the MCP server URL
[private issuer]  err=... refused non-public server-advertised discovery target
[private token]   err=mcp oauth: token endpoint: refused non-public server-advertised discovery target
[control, all loopback]  token="http://127.0.0.1:.../t"  err=<nil>
```

The last two are the point: a link-local `token_endpoint` reached through an otherwise well-formed same-origin chain is refused and named, while a loopback chain from a loopback MCP resource still completes.

I was ready to raise validate-then-use here, since `validateProtectedDiscoveredEndpoints` resolves the host and the POST happens later, which is the shape DNS rebinding exploits. You had already handled it: `protectedEndpointClient` is threaded into `tokenClient` and `registrationClient`, and `flow.httpClient` is the token client, so the exchange runs through the pinning transport rather than resolving again. Worth calling out because the browser-opened authorization endpoint is the one hop that cannot be pinned, and that is inherent rather than a gap.

`go test -race ./internal/mcp` passes, vet and gofmt clean, cross-compiles for linux, darwin and windows. All ten checks green.

gnanam1990 (approved):
Re-reviewed exact head `975eaeeca6ac251c61e8319a70e47ef0ce846849` against base `eeea3308fed38b37b203522624babaf2f30df91f`.

Third-party integration gate: this PR adds no module, SDK, plugin, vendored code, lockfile, or hosted-service dependency. It does add the already disclosed RFC 9728 protected-resource and RFC 8414 authorization-server discovery protocols in `internal/mcp`; their material boundary is remote metadata-controlled network destinations, redirects, proxy routing, dynamic registration, and token exchange.

## Verdict

**APPROVE** — No evidence-backed defects found in the reviewed scope.

## Prior-finding closure

- **Prior High — fixed:** endpoint provenance is retained after protected-resource discovery. Server-discovered authorization, token, and registration endpoints receive the resource-aware public-network policy; explicit operator overrides remain exempt. Registration and token requests use the protected client, so DNS is validated and the selected address is pinned at use time. Direct loopback/private endpoints and public-to-private redirects are rejected before reaching the victim.
- **Prior Medium — fixed:** configured HTTP, HTTPS, and SOCKS proxy first hops are preserved. I additionally exercised the HTTPS-proxy path end to end: the proxy received `CONNECT` for the pinned public target IP while the nested target request retained the original authorization-server host identity. The generated race-enabled case passed 20/20 runs.
- **Earlier redirect, issuer/resource identity, and cookie findings — fixed:** discovery redirects remain refused, returned identities are compared exactly, and discovery clients do not attach the caller's cookie jar.

## What I verified

- `go test -race ./internal/mcp -count=1`: pass.
- Generated HTTPS-proxy CONNECT/target-identity regression under `-race -count=20`: pass; temporary test removed afterward.
- `gofmt`/format check, `go vet ./...`, build, smoke, static lint, vulnerability scan, and diff hygiene: pass.
- Repository-wide `go test ./...` reaches two existing `internal/cli` doctor-test failures; the identical focused tests fail the same way on base `eeea3308`, so they are not a #991 regression.
- Exact-head GitHub smoke on Linux/macOS/Windows, security/code-health, performance, CodeQL, deterministic review, and CodeRabbit checks: all green.
- One review thread exists and is resolved; the branch is mergeable on the current base.

## Residual risk / not verified

- No live provider authorization, dynamic registration, browser callback, or token exchange was performed.
- Authorization occurs in the user's browser and therefore cannot reuse the Go client's address pinning; the discovered URL is still scheme/literal-host/DNS validated before it is opened.
- Discovery errors can still lose some underlying diagnostic detail. That is a bounded observability follow-up, not a demonstrated security, correctness, or availability blocker for this PR.

## Review integrity

- Reviewed worktree unchanged: yes.
- Head re-confirmed immediately before publication: yes (`975eaeeca6ac`).
- Evidence stale: no.

PR #994, opened 2026-08-28
fix(redaction): split camelCase keys in normalizeKey

### LINKED ISSUE #922: security: token-leak bug in canonical redaction package (Z-050)
### Description
There is a logic gap in the canonical redaction package's 
ormalizeKey function that leads to sensitive tokens leaking into logs and other outputs. The current normalization logic splits keys only on non-alphanumeric characters. 

As a result, keys using camelCase (e.g., ccessToken, 
efreshToken, piKey) are normalized to a lowercase string without delimiters (e.g., ccesstoken, 
efreshtoken). If the sensitiveKeys list—which the package uses to identify what needs redaction—expects a different normalization or uses specific matches that don't align with this simplistic alphanumeric split, these tokens are not flagged for redaction.

### Impact
High-entropy secrets (API keys, OAuth tokens) may be written to disk in logs or displayed in the TUI/CLI output. This exposes credentials to any process or user with access to the logs, significantly increasing the risk of credential theft.

### Recommended Fix
The normalization logic needs to be more sophisticated to recognize common programming naming conventions:
1. **CamelCase Splitting**: Modify 
ormalizeKey to treat transitions from lowercase to uppercase as delimiters (e.g., ccessToken $\rightarrow$ ccess_token).
2. **Unified Registry**: Unify the normalization logic between the code that detects keys in the stream and the registry of sensitiveKeys to ensure a consistent match.
3. **Broaden Heuristics**: Consider adding specific high-confidence patterns for common token formats if normalization proves insufficient.

hazyhaar:
Fixes #922

`normalizeKey` only inserted `_` on non-alphanumeric runes, so `accessToken` became `accesstoken` and missed the `access_token` registry entry. OAuth tokens in camelCase JSON/maps leaked into logs, TUI, and CLI.

This treats a lowercase-to-uppercase boundary as a delimiter (`accessToken` → `access_token`). Token-count fields (`promptTokens`, `maxTokens`) stay unredacted because the trailing segment is the plural `tokens`.

`doctor` reported credential presence under `credentialConfigured`, which the same split now treats as sensitive. The indicator is renamed to `authConfigured` so `set` / `not set` is not replaced by `[REDACTED]`.

The `reflect.Pointer` one-liner is independent govet on current `main` (deprecated `reflect.Ptr`). It can be dropped if maintainers prefer it on its own.

### DIFF
--- a/internal/config/unknownfields.go
+++ b/internal/config/unknownfields.go
@@ -131,7 +131,7 @@ type knownField struct {
 }
 
 func derefType(t reflect.Type) reflect.Type {
-	for t != nil && t.Kind() == reflect.Ptr {
+	for t != nil && t.Kind() == reflect.Pointer {
 		t = t.Elem()
 	}
 	return t
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -142,9 +142,10 @@ func providerConfigCheck(profile config.ProviderProfile) Check {
 		return check("provider.config", "Provider config", StatusFail, "No LLM provider is configured.", map[string]any{"help": "Set a provider in config or environment."})
 	}
 	// Report credential PRESENCE, never the value. Reported under a non-sensitive
-	// key ("credentialConfigured"): the prior "apiKey" key was itself sensitive, so
-	// check()'s redaction scrubbed the indicator to [REDACTED] — making "set"/"not
-	// set" invisible. HasConfiguredCredential is the shared definition of
+	// key ("authConfigured"): "apiKey" and "credentialConfigured" are themselves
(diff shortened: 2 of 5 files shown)


jatmn (approved):
@Vasanthdev2004 LGTM

anandh8x (approved):
Reviewed the latest head. The camelCase normalization closes a real redaction gap for accessToken, refreshToken, and apiKey while preserving token-count fields. The Doctor metadata rename is necessary to keep its set/not set indicator visible. Focused race tests for redaction, Doctor, and config pass.

kevincodex1 (approved):

PR #995, opened 2026-08-28
fix(config): replace deprecated reflect.Ptr with reflect.Pointer

hazyhaar:
`derefType` compared `t.Kind()` to `reflect.Ptr`. That constant is deprecated; `reflect.Pointer` is the replacement. Behavior is unchanged.

Independent of any other open PR. Once this is on `main`, other branches no longer need to carry the same one-liner to satisfy `govet`.

### DIFF
--- a/internal/config/unknownfields.go
+++ b/internal/config/unknownfields.go
@@ -131,7 +131,7 @@ type knownField struct {
 }
 
 func derefType(t reflect.Type) reflect.Type {
-	for t != nil && t.Kind() == reflect.Ptr {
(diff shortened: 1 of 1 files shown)


jatmn (approved):
@Vasanthdev2004 LGTM

anandh8x:
> Closing as a duplicate of #994, which already contains the same reflect.Ptr to reflect.Pointer change. Keeping the one-line update with the related redaction fix avoids duplicate review and a needless rebase conflict.

PR #999, opened 2026-08-31
feat(acp): emit structured file changes in tool updates

gnanam1990:
## Summary
- retain exact, bounded file before/after text for built-in `write_file`, `edit_file`, and `apply_patch` mutations
- redact both diff sides at the existing registry boundary
- emit ACP `content: [{type: "diff", path, oldText, newText}]` alongside the existing summary and locations
- represent moves as source deletion plus destination creation; preserve path-only fallback for large, binary, unchanged, and unsupported changes

## Verification
- `go test ./internal/tools ./internal/acp ./internal/agent`
- `go vet ./...`
- prior full `go test -count=1 ./...` was green before the final rename/copy representation correction; the affected packages were rerun afterward
- packaged ZeroApp smoke launched against the locally built Zero binary and loaded the Changes panel without regression

## Scope
- no dependencies, lockfiles, vendors, providers, or external integrations changed
- ZeroApp does not render the new ACP diff blocks yet; that is the follow-on C2B renderer slice.

## Summary by CodeRabbit

* **New Features**
  * File-editing results include structured before-and-after diffs for updates, additions, deletions, moves, and copies.
  * Diffs include file locations, existence details, and fallback changed-file paths across interactive, non-interactive, and permission-rejected actions.
  * Partial edit failures report committed changes and affected files.
  * Browser actions display clearer, operation-specific titles and metadata.

* **Bug Fixes**
  * Sensitive or obfuscated content is removed or redacted from displayed diffs.
  * Diff size and encoding limits are enforced reliably.
  * File writes and edits detect concurrent changes.
  * Unsafe formatter replacements suppress unreliable previews and diagnostics.
  * File permissions are preserved more reliably during edits.


jatmn (changes requested):

## Findings

- [P1] Send absolute paths in ACP diff content
  `internal/acp/translate.go:140`
  The new `FileDiff` producers retain the workspace-relative result path (for example `a.go`), and this translation forwards it unchanged. ACP v1 requires an absolute `path` for a diff, so a conforming client cannot reliably locate the edited file; it is not required to interpret the value relative to the session cwd. The root cause is that the new result type reuses the model/UI-facing path representation without carrying a separate wire-safe location. Preserve the existing `ChangedFiles` convention if needed, but establish an ACP-specific absolute path at the producer/transport boundary before serialization and cover write, edit, and subdirectory patch cases.

- [P1] Serialize `newText` for deletion and move-source diffs
  `internal/acp/types.go:225`
  `NewText` is tagged `omitempty`, while a deletion and the source half of a move deliberately use an empty new side. Their JSON records therefore omit `newText`, even though ACP requires that field. Clients can reject the content item or render a move as only a create. The underlying problem is that a zero-value string is being used both as meaningful file content and as an absent optional protocol field. Model the wire distinction explicitly so the required empty new value survives serialization, then cover create, delete, and both halves of a move at the JSON boundary.

- [P1] Do not expose control-byte-split secrets through FileDiffs
  `internal/tools/diff_preview.go:26`
  `utf8.ValidString` accepts NUL and ESC control bytes. The new path then sends those contents through `scrubResultSecrets`, but the current redactor matches before removing control bytes, so a credential split by one of them is not matched and is emitted to the ACP client. The root cause is treating valid UTF-8 as equivalent to safe text and applying pattern redaction before canonicalizing the input. Before emitting these text diffs, enforce the same normalize-before-match safety property (or decline unsafe text and fall back to `ChangedFiles`), with split-secret regression cases for both old and new sides and for the relevant C0/C1 controls.

- [P2] Apply a result-level bound to structured patch diffs
  `internal/tools/structured_patch.go:191`
  The 48 KiB check is applied to each `FileDiff`, but a multi-file `apply_patch` appends every qualifying entry and ACP sends them in one notification. A patch containing many individually valid files can consequently generate an arbitrarily large session update, bypassing the aggregate cap already used by `structuredPatchPreview`. The root cause is applying the preview limit at the entry level while the externally observable unit is one tool result/ACP update. Enforce one cumulative byte and/or entry budget across the entire result, stop adding structured content at that boundary, and retain `ChangedFiles` as the fallback for every omitted operation.

- [P2] Do not claim a create when an overwrite preimage could not be read
  `internal/tools/write_file.go:99`
  ACP calls this tool without a `FileTracker`. For an existing file that the process can write but cannot read, the preimage read failure is ignored, the write can succeed, and the emitted diff uses `oldText: ""` as if it had created the file. The root cause is conflating a failed preimage capture with a genuine empty preimage. Either fail closed or omit the structured diff when the preimage is unavailable; the protocol must not claim an exact before/after replacement it did not observe. Add coverage using a write-only/read-denied existing file through the ACP options path, not only a direct tool call with a tracker.

- [P2] Preserve empty-file mutations in the structured result
  `internal/tools/diff_preview.go:26`
  The equality guard treats `oldText == newText == ""` as an unchanged file. That drops a valid empty-file creation, deletion, copy, or move even though the filesystem changed, and `ChangedFiles` alone cannot recover the operation. The root cause is that equality of two content strings is being used as an operation test even though file existence and path transitions are separate state. Represent those operation states unambiguously (including the optional-old versus required-new ACP distinction) and add coverage for empty add/delete/move/copy cases.

Vasanthdev2004 (changes requested):
Requesting changes on one thing, which I drove myself because this PR opens a new outbound channel for raw file bytes.

## @jatmn's six findings at `9e404398`, so you get one answer rather than two

1. **Absolute paths in ACP diff content:** closed. `boundedFileDiff` hard-requires `filepath.IsAbs`; `write_file`, `edit_file` and `apply_patch` all emit an absolute `path` on the wire.
2. **Serialize `newText` for deletion and move-source diffs:** closed as written, but see the second finding below. The field is always emitted now and `NewExists` is dropped one layer up, so a delete and a truncate-to-empty are byte-identical to the client.
3. **Control-byte-split secrets:** **not closed.** Half of it is, and that half is load-bearing. This is the blocker.
4. **Result-level bound on structured patch diffs:** closed, and verified on the real call path rather than the unit test. A 40-file `apply_patch` gives `changedFiles=40 fileDiffs=11`, 46046 diff bytes, and `ChangedFiles` keeps all 40 as the fallback.
5. **Do not claim a create when the overwrite preimage is unreadable:** closed and load-bearing. Weakening the guard fails `TestWriteFileToolOmitsDiffWhenOverwritePreimageCannotBeRead` with exactly the false exact-replacement he described.
6. **Preserve empty-file mutations:** closed in `internal/tools`, then discarded in `internal/acp`. `OldExists`/`NewExists` carry the distinction correctly out of the tool and `appendToolResultDiffs` reads only `OldExists`.

## The blocker: a credential split by a zero-width character ships verbatim

`unsafeDiffText` is `r < 0x20 || (r >= 0x7f && r <= 0x9f)`. That stops at 0x9F and never reaches the Unicode format class. I wrote `const k = "sk-ant-api03-AAAABBBB<SEP>CCCCDDDDEEEEFFFFGGGG"` through a real `write_file` on the registry, then through `toolCallResult` and `json.Marshal`, and stripped the separator from the wire to see whether the canonical key was there:

```
plain          nDiffs=1 redacted=true  CREDENTIAL_ON_WIRE=false
NUL  \x00      nDiffs=0 redacted=false CREDENTIAL_ON_WIRE=false
ESC  \x1b      nDiffs=0 redacted=false CREDENTIAL_ON_WIRE=false
NEL  U+0085    nDiffs=0 redacted=false CREDENTIAL_ON_WIRE=false
ZWSP U+200B    nDiffs=1 redacted=false CREDENTIAL_ON_WIRE=true
ZWJ  U+200D    nDiffs=1 redacted=false CREDENTIAL_ON_WIRE=true
BOM  U+FEFF    nDiffs=1 redacted=false CREDENTIAL_ON_WIRE=true
SHY  U+00AD    nDiffs=1 redacted=false CREDENTIAL_ON_WIRE=true
NBSP U+00A0    nDiffs=1 redacted=false CREDENTIAL_ON_WIRE=true
```

`U+200B`, `U+200D`, `U+FEFF` and `U+00AD` all render at zero width, so a reader of that file sees an intact key. `res.Redacted` stays false, so nothing downstream is flagged either.

The base column is the part that makes this a blocker rather than an inherited problem. On `main` the same write produces:

```json
{"sessionUpdate":"tool_call_update","toolCallId":"c","status":"completed",
 "content":[{"type":"content","content":{"type":"text","text":"Created a.go (1 lines)."}}]}
```

Zero file bytes, for every separator. This is new egress, not a leak you inherited.

Two claims need correcting along with the code. The commit message on `9e404398` says controls are "rejected rather than normalized, so they cannot split a secret before transcript redaction", and the `FileDiff` doc comment says "Registry-boundary redaction applies to both sides before any caller receives it". Neither holds for this class.

`edit_file` is the worst case, because its `oldText` is the file's prior on-disk content. The model never chose those bytes, so a credential already sitting obfuscated in a config file is exported by an unrelated edit to that file.

Credit where it is due: the C0/C1 half really is load-bearing. Deleting the `unsafeDiffText` block from `scrubResultSecrets` fails `TestScrubResultSecretsDropsControlSplitFileDiff`. The gate works, its alphabet is just too small.

## Second, and cheap: `NewExists` never reaches the wire

`appendToolResultDiffs` reads `diff.OldExists` and never `diff.NewExists`. One real `apply_patch` deleting `gone.txt` and emptying `kept.txt` leaves the first absent from disk and the second present at 0 bytes, and both arrive identically:

```json
{"type":"diff","path":"...\\gone.txt","oldText":"payload\n","newText":""}
{"type":"diff","path":"...\\kept.txt","oldText":"payload\n","newText":""}
```

The create side is delivered correctly as `oldText:null`, so it is only the delete side that is lost. If the schema requires `newText` to be a non-null string, the answer is to stop emitting a diff block for deletions rather than encode them as truncations.

## Worth fixing in the same pass, not blocking

The new gate on `boundedUnifiedDiff` kills the existing TUI card for content it does not like. Same `edit_file`, `len(Display.Preview)`, head against base: form feed 0/76, ESC 0/74, invalid UTF-8 0/52, vertical tab 0/50, DEL 0/42. Base rendered all of these; head shows nothing and says nothing. The gate is whole-file, so one stray byte far from the edit suppresses the card. Zero's own tree is unaffected, but Emacs or C form-feed page breaks and ANSI golden fixtures are not exotic. Gating the rendered diff after `udiff.Unified` keeps the security intent without the regression.

`maxToolPreviewBytes` now caps two whole file copies, so structured diffs vanish above about 24 KiB per side, with a cliff at 24576. 57 of 641 non-test `.go` files in this repo are past it. The constant was written as a bound on a hunk and now governs four different quantities, and its doc comment still describes only the first. No signal is sent, so a client cannot tell a diff was withheld.

`appendGroup` counts path bytes against the per-file constant and skips out of order: `[40 KiB, 4 bytes, 40 KiB]` emits entries 1 and 2 and silently drops 3, so a later smaller diff survives while an earlier one disappears. And `write_file` shows a 48 KiB create where `apply_patch` shows nothing for identical content.

Smaller: one `tool_call_update` now names the same file twice in two spellings, `content[].path` absolute and `locations[].path` workspace-relative, and a client cannot correlate them. `scrubResultSecrets` filters `FileDiffs` in place through `res.FileDiffs[:0]`, aliasing the caller's backing array; no caller retains the pre-scrub slice today so it is latent, but it is the only field in that function filtered that way.

## Checked and correct

Content attribution is right, byte-compared against disk: BOM plus CRLF plus astral emoji plus no trailing newline round-trips exactly. An overwrite's `OldText` is the exact preimage, and `edit_file` takes `NewText` after `maybeFormatWrittenFile` so it matches disk under format-on-write. A move is delete-source plus create-destination, atomic, never a destination overwrite; copy emits only the destination create. `boundedFileDiff` declines rather than truncating, so a truncated side cannot be mistaken for an exact replacement. Plain unsplit credentials are correctly redacted out of the diff.

Blast radius is contained: `FileDiffs` has one consumer, is not persisted to session storage, is not in the model message, and does not inflate the output budget. All three producers go through `RunWithOptions` or the new explicit `ScrubResultSecrets`.

Build, vet and gofmt clean on head, linux and darwin cross-builds clean, `internal/acp` passes fully. The `internal/tools` and `internal/agent` failures reproduce identically on base, since the worktree sits under `%TEMP%`, a default sandbox write root.

Vasanthdev2004 (changes requested):
The leak is closed. Re-ran the same probe at `9b6696f5`:

```
plain          nDiffs=1 redacted=true  CREDENTIAL_ON_WIRE=false
NUL / ESC / NEL                        CREDENTIAL_ON_WIRE=false
ZWSP U+200B    nDiffs=0                CREDENTIAL_ON_WIRE=false
ZWJ / BOM / SHY / NBSP                 CREDENTIAL_ON_WIRE=false
```

Every separator that shipped a credential last round is now refused. That was the blocker and it is gone.

Still requesting changes, on the cost of the fix rather than on the fix.

## The gate now rejects text people legitimately write

`unsafeDiffText` is now `unicode.IsControl(r) || unicode.Is(unicode.Cf, r) || unicode.IsSpace(r)`, exempting only `\n`, `\r`, `\t` and space. `Cf` is the whole format class and `IsSpace` includes every Unicode space separator, so the rejected set is far wider than "characters that can split a secret".

Same `write_file`, previous head against this one, showing both surfaces:

```
                       9e404398              9b6696f5
plain ascii            nDiffs=1 preview=45   nDiffs=1 preview=45
family emoji (ZWJ)     nDiffs=1 preview=61   nDiffs=0 preview=0
flag emoji (ZWJ seq)   nDiffs=1 preview=57   nDiffs=0 preview=0
NBSP in prose          nDiffs=1 preview=58   nDiffs=0 preview=0
soft hyphen            nDiffs=1 preview=45   nDiffs=0 preview=0
BOM at file start      nDiffs=1 preview=49   nDiffs=0 preview=0
CJK / accented / emoji nDiffs=1              nDiffs=1
```

`preview=0` is the part that makes this more than an ACP question: the TUI diff card dies too, so a user editing one of these files sees no diff anywhere, and `Redacted` stays false so nothing says why.

The three that matter in practice. A ZWJ is how every multi-person and flag emoji is assembled, so one such character anywhere in a file suppresses its diff. A non-breaking space is ordinary in prose, and the file only has to contain one. And a UTF-8 BOM at file start is routine on Windows-authored files, which is a platform this project explicitly supports, with an open issue about preserving BOMs on write.

Being fair about blast radius, because I measured it rather than assuming: I scanned 1466 text files in this repo and **zero** contain a now-rejected rune. So Zero's own tree is unaffected, and this will not show up in CI or in your own editing. It lands on user content, which is exactly where it is hardest to notice.

And it fails closed, so this is availability rather than disclosure. That is why it is a narrow request rather than a reopening of the security question.

## The shape that gets both

Rejecting the file is doing normalization's job. What the gate needs to know is not "does this text contain a format character" but "does this text contain a secret that a format character is hiding".

Running the existing matcher twice would give you that: once on the raw text, once on a copy with the format characters removed. If either matches, drop the diff. A BOM at position 0, a ZWJ inside an emoji and an NBSP between two words all survive, because stripping them produces no new match. `sk-ant-api03-AAAA<ZWSP>BBBB` does not, because stripping it produces exactly the shape you already detect.

That also fixes the asymmetry the current gate has with the rest of the pipeline: `RedactString` is shape-based and content-agnostic, while this is a character allowlist, so the two disagree about what counts as dangerous and the stricter one silently wins.

If you would rather keep a character gate for now, narrowing it to the format characters that can actually sit inside a credential shape (the zero-width and joiner set) and dropping `IsSpace` would recover NBSP and most of the real cases, though not the BOM.

## Unchanged from last round

The `NewExists` finding stands: `appendToolResultDiffs` still reads only `diff.OldExists`, so a deleted file and a truncated one are byte-identical on the wire. The `maxToolPreviewBytes` cliff at 24576 per side, `appendGroup` counting path bytes against the per-file budget and skipping out of order, the two spellings of the same path in one `tool_call_update`, and the in-place `res.FileDiffs[:0]` filtering are all as I described them.

@jatmn's other five findings remain closed; nothing in this push disturbed them.

jatmn (changes requested):

## Overall guidance

These are not four unrelated edge cases. They come from one design mismatch: `FileDiff` is documented and serialized as exact evidence of a filesystem transition, but each producer currently builds it from whichever strings are available at that point in the tool lifecycle. Those strings may describe the planned change, an earlier snapshot, or an unverified fallback rather than the mutation that actually reached disk. The Unicode issue is the other side of the same boundary problem: content safety is decided with a blanket character filter before the existing secret detector can distinguish dangerous obfuscation from ordinary text. Fixing only the individual examples is likely to produce another round of variants.

Please define and enforce the `FileDiff` contract at one boundary. A retained entry should mean all of the following:

- `Path`, `OldExists`, `OldText`, `NewExists`, and `NewText` describe one mutation that actually committed, not merely a planned operation or the arguments supplied to the tool.
- When an existence bit is true, the corresponding text is a verified complete side of that transition. An unreadable, concurrently changed, formatter-modified, or otherwise uncertain side is not silently replaced with a convenient fallback.
- A failed multi-file operation can still report the subset that committed, but must not report planned operations that never reached disk.
- Unsupported, oversized, binary, deleted-at-the-ACP-layer, or unverifiable transitions remain visible through the existing `ChangedFiles` fallback rather than being presented as exact rich evidence.
- Secret handling remains fail-closed, but ordinary Unicode is not classified as sensitive merely because it contains an invisible or non-ASCII separator. Safety should depend on whether canonicalization reveals a secret shape.

The cleanest way to make that durable is to separate mutation evidence from presentation. Have the mutation layer return a typed outcome containing the committed changes and the confidence/availability of each side; then derive `FileDiffs`, `ChangedFiles`, previews, redaction, and ACP content from that outcome. In particular, the write/edit path should not infer the final state from `maybeFormatWrittenFile`'s fallback string, and the structured-patch path should not wait for whole-batch success before recording which planned changes committed. If a full refactor is out of scope, the minimum safe rule is: emit a rich entry only when both applicable sides were verified for the committed operation, otherwise retain the path-only fallback.

To prevent more review rounds, please validate the contract as a matrix rather than adding one regression per comment:

- operation: create, overwrite, edit, truncate-to-empty, delete, copy, move, and multi-file patch;
- outcome: full success, failure before the first commit, and failure after a committed prefix;
- post-write processing: formatter disabled, formatter success, formatter mutates then fails, timeout, and final read failure;
- external state: unchanged, overwrite between observation and commit, and create between non-existence observation and commit;
- content: plain text, ordinary ZWJ emoji/NBSP/BOM text, invalid or binary text, unsplit credentials, and credentials split by each supported invisible-separator class;
- transport: exact rich entry when proven, path-only fallback when not, redaction state when sensitive, and no entries for uncommitted changes.

The important invariant for those tests is not simply that a `FileDiff` exists. Whenever one exists, compare its existence flags and complete text byte-for-byte with the transition the test observed; whenever exact evidence is unavailable, assert that the rich entry is absent while `ChangedFiles` still identifies the committed path. Exercising that matrix at the tool-result boundary and again through ACP serialization should close the underlying contract instead of continuing to patch individual symptoms.

## Findings

- [P2] Preserve diffs for ordinary Unicode text
  `internal/tools/diff_preview.go:56`
  `unsafeDiffText` treats the entire Unicode `Cf` category and every non-ASCII space as unsafe. Because that predicate gates both `boundedFileDiff` and the rendered unified diff, a successful write or edit to a file containing an ordinary family-emoji ZWJ, an NBSP in prose, a leading UTF-8 BOM, or a soft hyphen emits neither ACP diff content nor the TUI diff card. The same files produced a preview on `main`, and the result is not marked redacted, so the user sees a successful mutation with no explanation for why both rich representations disappeared. The root cause is using a broad character-class allowlist as a proxy for credential evasion: harmless text and separator-obfuscated credentials are indistinguishable at that layer. Please make the safety decision based on whether removing or canonicalizing invisible separators exposes a sensitive value, rather than rejecting every occurrence of those characters. Regression coverage should retain ordinary ZWJ/NBSP/BOM text while still dropping credentials split by zero-width or whitespace separators.

- [P2] Report files committed before a patch failure
  `internal/tools/structured_patch.go:159`
  `applyStructuredPatchChanges` applies the planned changes sequentially and explicitly supports the case where an earlier file reached disk before a later operation failed. On that path it reports the committed names in the error string, but this return replaces the planned result with a fresh `errorResult`, discarding `ChangedFiles`, `FileDiffs`, and the preview. A patch that updates `first.txt` and then fails while replacing a non-empty directory therefore leaves `first.txt` modified while ACP receives no machine-readable change evidence and cannot distinguish “nothing changed” from “partially applied” during recovery. The root cause is that structured evidence is constructed only after the whole batch succeeds, even though the apply layer already knows the committed prefix. Please propagate a typed partial outcome containing evidence for exactly the changes that completed, or provide another explicit machine-readable incomplete-change result. It must not include planned files that never reached disk; add an end-to-end regression around the existing second-operation failure case.

- [P2] Bind the reported preimage to the write
  `internal/tools/write_file.go:102`
  The tool captures `priorContent`, performs a path-containment recheck, and only later calls `os.WriteFile`; the recheck does not verify that the file's contents or existence are still the state that was observed. If another process rewrites the file in that interval, the tool can overwrite those newer bytes while publishing the earlier snapshot as exact ACP `oldText`. A create race similarly allows `OldExists: false` to be reported after a file created by another process was actually clobbered. The optional `FileTracker` conflict check occurs before the same gap and therefore does not bind the evidence to the mutation either. Before this PR the race could affect the write and local preview, but the new `FileDiff` contract turns the stale observation into externally consumed “exact before/after” evidence. The root cause is that the preimage snapshot and the write are independent operations with no identity/version validation between them. Please have the mutation path return evidence tied to the state it actually replaced, or perform a last-moment identity/content validation and omit the rich diff when a conflict is detected; apply the same rule to `edit_file` and cover overwrite and create races.

- [P2] Do not trust failed formatter output as the final file
  `internal/tools/write_file.go:121`
  `maybeFormatWrittenFile` returns the requested string whenever an in-place formatter exits nonzero, times out, or its post-run read fails. That fallback does not prove the file is unchanged: a formatter can rewrite the file and then fail. The tool still succeeds, records the fallback in `FileTracker`, and publishes it as exact ACP `NewText`; for example, a formatter that writes `formatted` and exits 1 leaves disk at `formatted` while the emitted diff says `requested`. Normal successful formatting is handled correctly, so this specifically affects partial/failing formatter executions. The root cause is that the helper's string return value conflates a verified post-format read with an unverified fallback. Please return post-format verification state explicitly—or read the final file even after a formatter error—and construct `FileDiffs` only from verified final bytes. If final state cannot be read reliably, retain `ChangedFiles` but omit the rich diff. The `edit_file` path consumes the same helper and needs the same regression coverage.

jatmn (changes requested):

## Overall guidance

The two findings below are symptoms of one boundary problem, not unrelated path edge cases. `FileDiff.Path` is an absolute path, while `ChangedFiles` is normally workspace-relative. `toolResultLocations` tries to merge those two coordinate systems without receiving a workspace root or another authoritative identity key. It compensates by modifying path strings with `TrimSpace` and by treating suffix matches as proof that two paths identify the same file. Neither operation is identity-safe: trimming changes a valid filename, and suffix matching cannot distinguish `a.go` from `sub/a.go`.

Please fix this as a path-identity contract rather than adding special cases for the examples:

- Treat an already validated path as path data, not free-form display text. Preserve its bytes in ACP output; validation for an empty sentinel must not rewrite a non-empty filename.
- Suppress a `ChangedFiles` fallback only when it is provably the same file as a retained rich diff. If the translator lacks enough context to prove equivalence between an absolute and a relative path, retaining both locations is safer than hiding a real mutation.
- Keep the fallback-completeness invariant: every changed path must remain represented by either its own rich-diff location or its own path-only location. An eligible diff for one file must never consume another file's fallback.
- Preserve the existing intentional behavior for diff serialization, deletion fallback, location order, extra-root absolute paths, and large/unsafe content. This does not require changing the file-mutation tools or broadening ACP's wire format unless that is the smallest way to provide an authoritative identity.

One durable implementation option is to carry a canonical identity shared by both representations—or give the correlation step the trusted root needed to derive one. The smaller conservative option is to deduplicate only exact, already-comparable paths and tolerate an absolute/relative duplicate when equivalence cannot be established. Either is preferable to inferring identity from a basename suffix.

Please cover the boundary as a small matrix so this does not turn into another sequence of one-off review rounds:

- a normal file with both an absolute rich path and its relative fallback;
- two changed files named `a.go` and `sub/a.go`, with rich evidence available for both, only the root file, and only the nested file;
- the same-basename case when one file is ineligible for rich evidence because it is oversized or unsafe;
- filenames with leading and trailing whitespace, asserting byte-for-byte path preservation in both diff content and locations;
- exact absolute-path duplicates, confirming true duplicates are still removed;
- an ineligible or deleted rich diff, confirming a path-only location for that changed file remains present.

The key assertions should be about identity and completeness, not merely location count: no emitted path is rewritten, no distinct changed path disappears, and only a fallback proven to identify the same file is removed.

## Findings

- [P2] Do not conflate same-basename fallback locations
  `internal/acp/translate.go:188`
  The suffix match treats any absolute diff path ending in `/<changed>` as proof that it covers that relative location. For `ChangedFiles: ["a.go", "sub/a.go"]` and a single retained diff at `/workspace/sub/a.go`, `locationCoveredByFileDiff("a.go", ...)` returns true because the rich path ends in `/a.go`; the check for `sub/a.go` also returns true. The rich-diff loop emits only `/workspace/sub/a.go`, so both fallbacks are removed and ACP never reports that root `a.go` changed.

  This state is produced by normal PR behavior, not malformed input: `fileDiffsFromStructuredPatch` deliberately skips a change whose complete sides exceed the 48 KiB per-side limit (or fail the safety gate) and continues to later eligible changes. The current test uses different basenames (`rich.go` and `fallback.go`), so it cannot expose the collision. The root cause is using lexical suffix containment as file identity when the function has no root against which to resolve the relative path. Please correlate the two representations with an authoritative identity, or retain the relative fallback when equivalence cannot be proven. Do not solve this by basename-specific exceptions; the required outcome is that an eligible rich diff can suppress only its own fallback.

- [P2] Preserve whitespace in ACP file locations
  `internal/acp/translate.go:160`
  `strings.TrimSpace(diff.Path)` changes valid filenames with leading or trailing whitespace. For a file named `report.txt `, `appendToolResultDiffs` correctly emits diff content whose path is `/workspace/report.txt `, but `toolResultLocations` emits `/workspace/report.txt` instead. The adjacent content and location therefore identify different files. The `ChangedFiles` loop trims the relative fallback too, so it cannot restore the original identity.

  This is not necessary for empty-value filtering: the producer already supplies canonical absolute rich paths, and an actual missing sentinel can be rejected with `path == ""` without mutating non-empty data. The root cause is combining input cleanup with identity handling after the path has already been validated. This also remains unfixed in the current CodeRabbit thread. Please keep the original path byte-for-byte through location construction and perform validation without normalization; retain existing ordering and exact-duplicate behavior.

jatmn (changes requested):
I found three issues that need to be addressed before this is ready.

## Overall guidance

These findings come from two boundary contracts this PR already establishes rather than from three unrelated edge cases.

First, a retained `FileDiff` crosses a new outbound boundary carrying complete file bytes. Its secret check therefore needs to answer whether canonicalizing visually ignorable separators reveals a credential already recognized by the redactor. It should not depend on a partial list of Unicode general categories, because characters with the same relevant behavior occur outside `Cf` and `Space`.

Second, a retained `FileDiff` is presented as exact evidence of a committed filesystem transition. A successful read alone does not establish that evidence: the read must remain within the authorized path scope, and an update preimage must be checked at the operation that publishes the replacement. Please centralize those guarantees where mutation outcomes become `FileDiffs`, so callers cannot accidentally turn a readable or planned string into authoritative evidence. When either applicable side cannot be verified, keep `ChangedFiles` and omit the rich entry.

This does not require making arbitrary cross-process writes perfectly linearizable. The minimum durable rule is to remove avoidable work between the last preimage check and commit, scope-check every post-processor read, and emit rich evidence only when those checks succeed. Tests should assert both the exact retained bytes and the path-only fallback when verification fails.

## Findings

- [P1] Include default-ignorable combining characters in the shape-based secret check
  `internal/tools/diff_preview.go:66`

  `unsafeDiffText` only invokes `diffTextRevealsObfuscatedSecret` after seeing a control, `Cf`, or whitespace rune. That misses visually ignorable characters in other Unicode categories. For example, U+034F COMBINING GRAPHEME JOINER and U+FE0F VARIATION SELECTOR-16 are valid UTF-8 and are neither controls, `Cf`, nor whitespace. Inserting either one into a credential recognized by `redaction.RedactString` causes the raw matcher to miss it, while `unsafeDiffText` returns false without attempting canonicalization. A registry-run write or edit consequently retains the complete obfuscated credential in `FileDiff.OldText` or `NewText`, and ACP serializes it.

  This is new egress on this branch: the base result did not send complete file sides through ACP. It is also the same root cause as the earlier zero-width-separator issue, not a new requirement to reject broad classes of Unicode. Please define the separator set around the relevant property—characters that can be visually ignorable inside a credential shape—and run the existing redactor on a canonicalized copy. Drop the rich diff only when canonicalization reveals a secret. Ordinary variation selectors and combining text must remain eligible when canonicalization does not reveal one. Add regressions for U+034F and U+FE0F on both old and new sides, through registry scrubbing and ACP translation.

- [P2] Scope-check the file object read after format-on-write
  `internal/tools/format_on_write.go:99`

  The write target is validated before the mutation, but an external formatter or concurrent process can replace that pathname before this post-format read. `readFormattedFile` is a plain `os.ReadFile`, so if the replacement is a symlink to a file outside the workspace, it follows the link and returns the external bytes. `maybeFormatWrittenFile` then marks those bytes as known; write/edit use them as `FileDiff.NewText`, and the tracker records them as the target's baseline. This also occurs when the formatter exits unsuccessfully because its error is intentionally ignored before the final read.

  The root problem is that `finalContentKnown` currently means only “the pathname was readable,” while its consumers treat it as “the committed target's final, authorized contents were verified.” Preserve the desirable behavior of reading a formatter's final output even after a nonzero exit, but perform that read through the workspace's rooted/path-scoped mechanism and verify that the final object is an eligible in-scope file. Normal formatters that replace a file atomically within the workspace must continue to work; this should not require preserving the original inode. If scope or object validation fails, forget the tracker baseline, retain `ChangedFiles`, and omit the preview and `FileDiff`. Cover an in-workspace atomic replacement as the success case and an out-of-root symlink replacement as the fallback case for both write and edit.

- [P2] Move the structured-update preimage check to the publish boundary
  `internal/tools/structured_patch.go:823`

  For a same-path update, this comparison happens before `writeStructuredPatchFile` creates a temporary file, writes its contents, applies its mode, and closes it. A concurrent writer can change the destination during that avoidable interval; `root.Rename` then overwrites those newer bytes, but `fileDiffsFromStructuredPatch` still reports the earlier planned `change.before` as the exact ACP `OldText`. The comment says the re-read occurs immediately before commit, but for updates the staging work means that is not true.

  The underlying possibility of concurrent mutation predates this PR, but presenting the planned preimage as exact external evidence is new. Please stage the replacement first, then perform the rooted content/identity recheck immediately before publishing it, with no file construction or formatting work between the check and rename. If the recheck fails, discard the staged file and keep only path-level failure reporting. This finding is limited to the avoidable same-path update window; it does not require eliminating the irreducible scheduler gap around a filesystem syscall or redesigning deletion/move reporting. Add a deterministic hook at the pre-rename boundary and assert that a competing update is preserved and no stale rich diff is emitted.

jatmn (changes requested):
I found three issues that need to be addressed before this is ready.

## Overall guidance

These look like three manifestations of the same contract-boundary problem, rather than three unrelated edge cases: `FileDiff` is exact, user-visible evidence, but the code sometimes decides that evidence is valid before the last operation that can invalidate it.

- The write path validates the preimage and then performs another syscall before mutation.
- The structured-patch path models the current operation as a single `done` boolean even when only part of that operation reached disk.
- The registry validates that a raw diff changed, transforms both sides by redacting them, and does not validate the transformed result again.

The durable rule should be: construct rich evidence from the final known filesystem outcome, apply every outbound transformation, and then revalidate the invariants that those later steps can affect. If the implementation cannot prove a complete, safe transition, retain `ChangedFiles` and omit only the rich `FileDiff` entry. That fallback is already part of this PR's design and should remain the recovery path rather than guessing at bytes or effects.

A useful implementation shape would be a small typed mutation outcome that records the exact sub-effects known to have completed, followed by one evidence-finalization boundary that:

1. derives old/new sides only for completed effects;
2. applies redaction and output limits;
3. checks post-transformation changedness and safety; and
4. emits `FileDiff` only when the resulting evidence still represents a real, complete transition.

This does not need to become a broad refactor. A focused fix in the existing helpers is fine if it establishes that rule consistently. In particular, this review is not asking this PR to provide general cross-process filesystem linearizability, adopt the atomic-publication work from #941, absorb encoding work from #988, change the intentional ACP path-only representation for deletions, or redesign formatter success semantics. Keeping those boundaries explicit should prevent another round of adjacent fixes and regressions.

Please cover the three contracts together in one regression pass:

- For tracked overwrite/edit, make the expected-byte comparison the final validation immediately before truncate/write (or perform mutation under an equivalent already-held synchronization boundary). Ensure the helper does not add a fresh identity syscall after that comparison. Preserve the existing create exclusivity, path-identity checks, mode handling, formatting, and `ChangedFiles` fallback.
- For structured operations, inject a failure after a destination is published but before the remaining step completes. Assert an error result, a destination path in `ChangedFiles`, exact destination-creation evidence and preview, and no claimed source deletion. Covering the equivalent no-replace publication failure would verify that the solution is based on completed sub-effects rather than a special case for moves.
- For redaction, rotate one recognized credential to a different recognized credential so both sides become the same redaction token. Assert that the rich entry is omitted, `ChangedFiles` remains, and ACP emits no unchanged diff block. Add the companion case where surrounding non-secret text differs, so a redacted diff that still shows a real transition remains present.

## Findings

- [P2] Make the preimage comparison the last step before overwriting
  `internal/tools/file_commit.go:72`
  The overwrite path reads and compares the opened file with `expectedContent` at lines 61–70, but then calls `writeAndVerifyFileIdentity`, which repeats `file.Stat` before `Truncate`. That second stat is not merely redundant bookkeeping: it creates a PR-introduced interval after the final byte validation in which another writer can update the same inode. Such an update preserves every `SameFile` identity check, so Zero can truncate the newer contents and still return success with the older snapshot in `FileDiff.OldText`. The general possibility of an external writer racing the eventual mutation already existed and is outside this PR's accepted scope; the actionable regression here is narrower—the new evidence-binding path itself adds avoidable work after its final content check. Reuse `openedInfo`, move any required identity validation before the byte comparison, or move the comparison into the commit helper so that no fresh syscall separates it from `Truncate`. Keep the post-write pathname check, since it protects a different contract. A regression should exercise this final-validation boundary rather than only the existing hook before `os.OpenFile`.

- [P2] Model and report the committed sub-effects of a failed operation
  `internal/tools/structured_patch.go:165`
  `applyStructuredPatchChange` can return `done == true` together with an error. For example, a move can successfully publish the destination and then fail to remove the source; the no-replace path can likewise publish the target and encounter a later chmod or cleanup error. `applyStructuredPatchChanges` records only the destination path in `incompletePaths`, while `FileDiffs` and the preview are derived solely from the previously completed entries in `outcome.committed`. The result therefore says that a path changed but discards exact bytes for a destination creation that the implementation already knows reached disk. Treating the whole operation as committed would be equally wrong because the source may still exist. Replace the operation-wide boolean bookkeeping with an outcome that identifies completed sub-effects (at minimum, destination publication separately from source removal), and derive `ChangedFiles`, `FileDiffs`, preview text, and error reporting from that same outcome. On destination-published/source-remove-failed, report only the destination creation as rich evidence and keep the source out of deletion evidence. This closes the current partial-operation hole without claiming uncertain effects or changing ACP's intentional deletion fallback.

- [P3] Revalidate changedness after redacting both sides
  `internal/tools/registry.go:363`
  `boundedFileDiff` rejects an unchanged same-existence transition before scrubbing, but the registry then redacts `OldText` and `NewText` independently and appends the entry without checking it again. When one recognized credential is replaced by another, the distinct raw sides can both become the same `[REDACTED]` text. ACP then receives a diff block whose old and new contents are identical, which violates the retained-entry contract and gives the model no visible transition. Add a post-redaction semantic check: when both sides have the same existence state and their scrubbed text is equal, drop that `FileDiff` while preserving the path in `ChangedFiles`. Do not drop entries merely because redaction occurred—if non-secret surrounding text still differs, the scrubbed transition remains useful and should be retained. Keeping this check at the outbound finalization boundary will also prevent future transformations from reintroducing semantically empty evidence.

jatmn (dismissed):
lgtm, but please rebase on main

jatmn (dismissed):
I rechecked head `f4d11a3`. The evidence-binding work from the earlier rounds looks addressed on this head: the preimage/commit race in `commitFileContents`, partial-patch sub-effect reporting via `completedStructuredPatchEffect`, post-redaction unchanged-diff dropping in `scrubResultSecrets`, pre-permission `FileDiff` scrubbing, and ACP absolute-path emission all match what I asked for at `6ed27e5` and the follow-up commits. I have one remaining finding below. I am intentionally not reopening the other threads from this review series.

## Overall guidance

The reason this PR has gone through so many review rounds is not that it keeps introducing unrelated defects. Most of the churn has been one contract getting implemented in layers without a single finalization boundary that every outbound surface respects.

The contract this PR owns is:

**Only emit exact, user-visible file evidence when you can prove the final on-disk outcome matches what you are about to report. If you cannot prove that, keep `ChangedFiles` as the safe path-only fallback and omit the rich `FileDiff` entry. Never describe a path as successfully committed when post-publish verification failed.**

That rule is already stated in `types.Result.FileDiffs`, enforced in `boundedFileDiff`, `scrubResultSecrets`, `finalContentKnown` gating for write/edit, and `completedStructuredPatchEffect` for structured patch. The earlier findings were places where code paths still decided a path was "good enough to report" before the last step that could invalidate that decision, or where one layer (helper) made a careful distinction that a later layer (error aggregation) collapsed back together.

You fixed most of those collapse points. The remaining item is the same class of bug at the error-aggregation layer: `incompleteStructuredPatchWrite` correctly refuses to put unverified publications into `outcome.committed`, but `applyStructuredPatchChanges` and `applyPatchOperations` still treat `incompletePaths` as if they were committed when building the model-facing error and `ChangedFiles`.

**Root cause to fix, not symptoms:** do not patch this by tweaking one string in the error message alone. The durable fix is to make partial-failure reporting flow through one typed outcome — verified `committed` sub-effects vs `incompletePaths` that may have touched disk but failed byte verification — and have every consumer (`ChangedFiles`, `"already committed: …"` text, `FileDiffs`, preview) derive from that outcome without re-merging the two buckets.

Concretely, that means:

1. **Keep the helper split you already wrote.** `incompleteStructuredPatchWrite` → `completed` only when `completedStructuredPatchEffect` passes; otherwise `incompletePaths`. Do not undo that.
2. **Stop merging `incompletePaths` into the committed list** at `applyStructuredPatchChanges:770` or `applyPatchOperations:164`. The `"already committed: …"` fragment should name only `changedFilesFromStructuredPatch(relativeRoot, outcome.committed)`.
3. **If incomplete paths need to be visible for recovery**, surface them with distinct wording — for example a separate `"published but unverified: …"` clause, or include them only in `ChangedFiles` without calling them committed. The model must not be told to treat them as successfully patched bytes.
4. **Add one end-to-end regression** that wires the low-level partial-publication fixtures you already have (`TestStructuredPatchCopyFailureReportsSurvivingPartialTarget` + `TestIncompleteStructuredPatchNoReplaceWriteReportsOnlyVerifiedTarget`) through `applyStructuredPatchChanges` or `apply_patch` and asserts:
   - verified prefix paths appear in `"already committed: …"` and may have `FileDiffs`;
   - unverified `incompletePaths` do **not** appear in that committed fragment;
   - no `FileDiff` is emitted for unverified paths.

That single regression pass should close this contract for partial failures without opening new surface area.

**What I am not asking for in this round** — please do not expand scope; these were explicit boundaries in my prior reviews and remain out of scope:

- General cross-process filesystem linearizability or locking across the workspace
- Atomic-publication work from #941
- Encoding / NUL-byte policy changes from #988
- Changing the intentional ACP path-only representation for deletions (`appendToolResultDiffs` skipping `!NewExists`)
- Redesigning format-on-write success semantics or summary line counts when `finalContentKnown` is false
- Session persistence of `FileDiffs`, stream-json emission, or ACP `Redacted` wire fields
- Docstring coverage or other CodeRabbit policy nits

If the fix above is landed with the end-to-end test, I do not expect another evidence-binding round from me on this PR.

## Findings

- [P3] Do not label unverified patch targets as "already committed"
  `internal/tools/structured_patch.go:770`

  **What goes wrong.** `incompleteStructuredPatchWrite` (lines 813–824) is careful: after a write returns `published == true` but `completedStructuredPatchEffect` cannot read back the planned `after` bytes, it puts the path in `incompletePaths` and leaves `outcome.committed` empty for that change. `TestIncompleteStructuredPatchNoReplaceWriteReportsOnlyVerifiedTarget` encodes that split.

  On the error path, that split is undone:

  - `applyStructuredPatchChanges` (line 770) appends `outcome.incompletePaths` into the same list used for the `"already committed: …"` error text.
  - `applyPatchOperations` (lines 163–165) appends the same paths into `result.ChangedFiles`.
  - `FileDiffs` correctly come only from `applyOutcome.committed` (line 165), so the rich evidence is omitted — but the error string and path list still tell the model the path was committed.

  **When it happens.** The rare exclusive-copy partial-publication path you already test at the copy layer: `TestStructuredPatchCopyFailureReportsSurvivingPartialTarget` leaves `target.txt` containing `"partial"` while the planned content was `"complete content\n"`. If that bubbles through `applyStructuredPatchChange` → `incompleteStructuredPatchWrite`, the model can receive `"already committed: target.txt"` and a `ChangedFiles` entry even though verification failed and the bytes on disk are wrong.

  **Why this matters.** The model uses the error text to decide what to re-read and what to retry. Calling an unverified path "already committed" is worse than omitting it: it sends the agent down a recovery path that assumes the patched bytes landed. That is the same contract violation I flagged in the partial-move and preimage rounds, just at the error-aggregation layer instead of the write helper.

  **Requested outcome.** Keep verified `outcome.committed` as the only source for `"already committed: …"`, `FileDiffs`, and preview. Report `incompletePaths` separately with wording that does not imply successful commit, or list them only as paths that need manual inspection/re-read. Add the end-to-end regression described in Overall guidance so this layer cannot drift again.

Vasanthdev2004 (changes requested):
Re-reviewed at `2b08339a`. Everything I asked for is closed, and I drove the one that mattered rather than reading it. One thing keeps this from an approval, and it is not mine.

## The character gate

This is the two-pass shape: control bytes are still refused outright, and a format or space rune only costs the diff when stripping it exposes a credential to the normal redactor. Same `write_file` probe as last round, through the Registry, at this head:

```
plain ascii                nDiffs=1 preview=76
family emoji (ZWJ)         nDiffs=1 preview=79
england flag (tag chars)   nDiffs=1 preview=80
NBSP in prose              nDiffs=1 preview=68
soft hyphen                nDiffs=1 preview=59
BOM at file start          nDiffs=1 preview=59
variation selector         nDiffs=1 preview=60
CJK / accented / emoji     nDiffs=1 preview=62
combining grapheme joiner  nDiffs=1 preview=48
```

Every one of those was `nDiffs=0 preview=0` at `9b6696f5`. And the reason the gate exists still holds, on a wider set than I tried before:

```
plain                      nDiffs=1 redacted=true   onWire=false
NUL / ESC                  nDiffs=0                 onWire=false
NEL ZWSP ZWJ ZWNJ          nDiffs=0 redacted=true   onWire=false
BOM(mid) SHY NBSP VS16     nDiffs=0 redacted=true   onWire=false
CGJ TAG U+3000 U+2060      nDiffs=0 redacted=true   onWire=false
```

Fifteen separators, none of them reaches the wire in either the split or the joined form. That was the trade I asked for and it is the one you made.

## The rest of mine

`NewExists` is consulted now: a deletion is omitted from the diff blocks rather than sent as a truncation, with `ChangedFiles` as the fallback, and the comment says why. That is the right answer given the ACP block has no existence bit. The in-place `FileDiffs[:0]` filter is gone. `appendGroup` and its per-file path-byte accounting are gone with the restructure. The two spellings of one path in a single update are still possible, and `toolResultLocations` now documents that as deliberate, since correlating an absolute `FileDiff` path with a workspace-relative `ChangedFiles` entry without the root would let a suffix match hand one file's fallback to another; I accept that reasoning. The preview bound moved to 48 KiB and is still a silent cliff, which is consistent with the contract that unprovable evidence degrades to path-only, so I am not pressing it.

## What is still open is jatmn's, and it has not moved

His last review at `f4d11a3b` left exactly one finding: `applyStructuredPatchChanges` folds `incompletePaths` into the "already committed" list, and the tool result puts them into `ChangedFiles` with no distinct wording. At this head:

```go
// structured_patch.go:770-772
committedPaths = appendUniqueStructuredPatchPaths(committedPaths, relativeRoot, outcome.incompletePaths)
if len(committedPaths) > 0 {
    return outcome, fmt.Errorf("%w; patch was partially applied ... already committed: %s; ... re-read the committed files before retrying", err, strings.Join(committedPaths, ", "))
```

```go
// structured_patch.go:164
result.ChangedFiles = appendUniqueStructuredPatchPaths(result.ChangedFiles, relativeRoot, applyOutcome.incompletePaths)
```

I checked this mechanically rather than by eye: the GitHub compare from `f4d11a3b` to `2b08339a` has no hunk in `structured_patch.go` at all, so the site is byte-for-byte what he reviewed, and nothing under `internal/tools` asserts that an incomplete path is named as anything other than committed.

The consequence is the model-facing one he described. A path whose bytes failed verification is announced as "already committed" and the model is told to re-read it before retrying, so it will build its next edit on a file that may hold half a write. Everything else in this PR is careful about exactly that distinction, which is why the one place that collapses it stands out.

Requesting changes for that alone. His four points are the fix: keep the helper split, name only `outcome.committed` as committed, give incomplete paths their own clause if they need to be visible, and one end-to-end regression on the partial-publication fixtures you already have. Mine are all closed, and I will approve the moment his is.

gnanam1990 (commented):
Self-review after the requested changes at `e54e5af6d723bf03bffdfc57e3cb3787adc255fc`, based on main `6937a309cf00825572210a7610a1f3ea8b74c2f9`.

The remaining partial-publication evidence finding is fixed. The apply_patch error path now keeps only byte-verified completed effects in `ChangedFiles`, `FileDiffs`, and the “already committed” clause. A target that became visible but could not be verified is reported separately as “published but unverified”, with recovery guidance, and is omitted from committed/rich-diff evidence.

Added an end-to-end apply_patch regression containing a verified prefix, a surviving partial target, and an untouched suffix. Before the fix, it reproduced the uncertain path being reported as committed. It now verifies the filesystem outcome and proves the uncertain path is absent from the committed clause, `ChangedFiles`, and `FileDiffs`. The regression passed 20 repetitions.

Full tools tests and race tests, formatting, full vet and repository tests, release build/smoke, govulncheck, and diff hygiene passed. Advisory lint findings are confined to unchanged mainline code. No dependency or third-party integration change. No evidence-backed defects found in this reviewed fix scope.

The commit was pushed as a fast-forward. Remote CI and independent re-review must cover this head; this self-review does not replace independent approval. Native Windows execution was not performed locally.

Vasanthdev2004 (dismissed):
Re-reviewed at `e54e5af6`. The one thing that was open is closed, and closed the way jatmn described rather than by editing a string.

Both sites moved. `applyStructuredPatchChanges` now builds `committedPaths` from `outcome.committed` alone and `unverifiedPaths` from `incompletePaths`, with their own clauses: "already committed: ..." and "published but unverified: ...", and the retry guidance switches to "re-read the listed files" when anything is unverified. The tool result at the top no longer appends `incompletePaths` to `ChangedFiles`; `ChangedFiles`, `FileDiffs`, `Redacted` and the preview all derive from `committed` only. That is the one typed outcome with no re-merge.

`TestApplyPatchSeparatesVerifiedAndUnverifiedPartialPublications` is the end-to-end regression on the partial-publication fixtures, and it is load-bearing. I folded the two lists back together the way the old code had them and ran it:

```
--- FAIL: TestApplyPatchSeparatesVerifiedAndUnverifiedPartialPublications
    partial publication error must contain "published but unverified: uncertain.txt":
    ... already committed: verified.txt, uncertain.txt; ... re-read the committed files before retrying
```

which is precisely the message the model used to get. It also pins that the unverified path is absent from the committed fragment, that `ChangedFiles` and `FileDiffs` carry only the verified path, and that the later publication stayed untouched. `internal/tools` is green here.

Nothing from my earlier rounds changed in this commit, so the character-gate probe results and the rest stand as posted. Approving.

anandh8x (dismissed):
lgtm — rechecked at `e54e5af6`. The partial-publication fix lands the way jatmn described: `committed` vs `incompletePaths` stay separate all the way up, "published but unverified" is its own clause now, and `TestApplyPatchSeparatesVerifiedAndUnverifiedPartialPublications` pins it (fails on the old merge with `already committed: verified.txt, uncertain.txt`, passes here).

Thanks for sticking through all the review rounds — approving.

gnanam1990:
@coderabbitai review

gnanam1990:
Addressed the current-head recovery-coverage finding in `fa68f834`.

- Added a regression that forces the formatter to replace the target with a directory and then fail.
- The test proves `RestoreFailed=true`, `ContentKnown=false`, and the explicit warning shown to the user.
- Production behavior was already correct; this commit closes the missing proof.
- No dependency or third-party integration changes.

Validation:
- `make fmt-check`
- `go vet ./...`
- `go test ./...`
- release build + smoke
- `make vulncheck`
- `git diff HEAD^ --check`
- `lint-static` has only the same four repository-wide advisory quick-fix findings, none in this change.

@coderabbitai review

jatmn (dismissed):
LGTM

Vasanthdev2004 (dismissed):
Re-approving at `fa68f834`. The push dismissed my approval at `e54e5af6`; the one commit since is the test CodeRabbit asked for, and it is load-bearing rather than decorative.

`TestFormatOnWriteReportsWhenFormatterFailureCannotBeRestored` swaps the formatter for one that replaces the target with a directory and then fails, so the restore of the requested content cannot land. It asserts `RestoreFailed`, that `ContentKnown` is false, and the exact warning the user sees. With the `RestoreFailed = true` assignment in `maybeFormatWrittenFile` flipped off, it fails on `failed recovery was not reported`, so it pins the production branch it claims to. Nothing else moved: the structured file-change work and the verified-versus-unverified split I approved last time are the same objects.

`internal/tools` and `internal/acp` pass here on this head, CI 12 of 12.

gnanam1990:
@coderabbitai review

gnanam1990:
@coderabbitai full review

gnanam1990:
Addressed both findings from the current full review:

- `abbed6f` — ACP file diffs now serialize deletion as `newText: null`, preserving it distinctly from empty-file replacement (`newText: ""`) and creation (`oldText: null`). Move clients can now render both source deletion and destination creation.
- `9d324c4` — formatter failure recovery opens through the existing descriptor-bound `os.Root` with `O_CREATE`, restores the original mode for overwrites/edits, and uses `0644` only for newly created files. Regression coverage removes the formatter target before returning an error and verifies bytes plus mode.

Validation:
- focused regressions for both findings
- full `internal/acp` and `internal/tools` suites
- race suites for both packages
- vet for both packages
- `make fmt-check`
- full repository test: all packages pass except the same pre-existing unrelated TUI `TestAltScreenTranscriptScrollKeepsFooterFixed` failure

No third-party dependency/module change.

@coderabbitai review

gnanam1990:
Addressed the latest test-strength finding in `6ca569d`: create/delete wire assertions now require the nullable `oldText`/`newText` keys to be present before asserting JSON `null`, so a future `omitempty` regression cannot pass as a missing-map-key nil.

Validation: focused nullable-diff regressions, full `internal/acp`, and `go vet ./internal/acp`.

@coderabbitai review

kevincodex1 (approved):

PR #1000, opened 2026-09-01
feat(acp): surface safe browser tool metadata

gnanam1990:
## Summary
- emit an optional, versioned browser descriptor for ZERO built-in browser tool calls
- keep browser titles free of typed text, CDP endpoints, and URL query/fragment data
- preserve the descriptor through start, result, and permission-request paths

## Verification
- `make fmt-check`
- `go vet ./...`
- `go build ./...`
- `go test -race ./internal/acp`
- `go test ./...`
- `make lint-static`
- `make vulncheck`
- real `zero acp` initialize handshake

No dependencies or lockfiles changed. No full mutation sweep was run due to local storage-safety constraints.

## Summary by CodeRabbit

* **New Features**
  * Added structured browser metadata to tool activity updates.
  * Browser actions now display concise, presentation-safe titles, including URL origins when applicable.
  * Browser connection details include the helper command and version information.

* **Bug Fixes**
  * Corrected browser metadata preservation for browser connection requests.
  * Prevented similarly named non-browser tools from being incorrectly identified as browser actions.

### DIFF
--- a/internal/acp/permission.go
+++ b/internal/acp/permission.go
@@ -149,9 +149,7 @@ func permissionToolCall(req agent.PermissionRequest) ToolCallUpdate {
 		Status:     ToolStatusPending,
 		RawInput:   rawInputBytes(args),
 	}
-	if browser, ok := browserToolDetails(req.ToolName); ok {
-		upd.Browser = browser
-	}
+	attachBrowserToolDetails(&upd, req.ToolName)
 	return upd
 }
 
--- a/internal/acp/permission.go
+++ b/internal/acp/permission.go
@@ -142,13 +142,17 @@ func actionOffered(optionID string, offered []PermissionOption) bool {
 // session/request_permission request from a ZERO permission request.
 func permissionToolCall(req agent.PermissionRequest) ToolCallUpdate {
 	args := marshalArgs(req.Args)
-	return ToolCallUpdate{
+	upd := ToolCallUpdate{
 		ToolCallID: req.ToolCallID,
 		Title:      toolTitle(req.ToolName, string(args)),
 		Kind:       toolKindFor(req.ToolName),
 		Status:     ToolStatusPending,
 		RawInput:   rawInputBytes(args),
 	}
+	if browser, ok := browserToolDetails(req.ToolName); ok {
+		upd.Browser = browser
+	}
+	return upd
 }
 
 func marshalArgs(args map[string]any) []byte {
--- a/internal/acp/permission_test.go
+++ b/internal/acp/permission_test.go
@@ -89,8 +89,8 @@ func TestPermissionToolCallKeepsTheBrowserDescriptor(t *testing.T) {
 		ToolName:   "browser_connect",
 		Args:       map[string]any{"target": "127.0.0.1:9222"},
 	})
-	if call.Browser == nil || call.Browser.Version != 1 || call.Browser.Command != "connect" {
-		t.Fatalf("browser descriptor = %#v", call.Browser)
+	if got := browserDescriptor(t, call); got != (BrowserToolDetails{Version: 1, Command: "connect"}) {
+		t.Fatalf("browser descriptor = %#v", got)
 	}
 	if call.Title != "browser connect" {
 		t.Fatalf("title = %q", call.Title)
--- a/internal/acp/permission_test.go
+++ b/internal/acp/permission_test.go
@@ -82,3 +82,17 @@ func TestPermissionToolCall(t *testing.T) {
 		t.Error("expected rawInput from args")
 	}
 }
+
+func TestPermissionToolCallKeepsTheBrowserDescriptor(t *testing.T) {
+	call := permissionToolCall(agent.PermissionRequest{
+		ToolCallID: "browser-1",
+		ToolName:   "browser_connect",
+		Args:       map[string]any{"target": "127.0.0.1:9222"},
+	})
+	if call.Browser == nil || call.Browser.Version != 1 || call.Browser.Command != "connect" {
+		t.Fatalf("browser descriptor = %#v", call.Browser)
+	}
+	if call.Title != "browser connect" {
+		t.Fatalf("title = %q", call.Title)
+	}
+}
--- a/internal/acp/translate.go
+++ b/internal/acp/translate.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/url"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/Gitlawb/zero/internal/agent"
@@ -119,13 +120,29 @@ func browserToolTitle(command, rawArgs string) string {
 			return "browser open"
 		}
 		origin := u.Scheme + "://" + u.Host
-		if !utf8.ValidString(origin) {
+		if !browserTitleTextSafe(origin) {
 			return "browser open"
 		}
 		return "browser open " + truncateHint(origin)
 	default:
 		return "browser " + command
 	}
+}
+
+// browserTitleTextSafe validates text after URL parsing has decoded escaped
+// UTF-8 in the host. Valid UTF-8 alone is not presentation-safe: control,
+// format/bidi, and line/paragraph separator runes can reorder or split the
+// permission label shown to a user. The execution URL remains unchanged.
+func browserTitleTextSafe(text string) bool {
+	if !utf8.ValidString(text) {
+		return false
+	}
+	for _, r := range text {
+		if unicode.IsControl(r) || unicode.Is(unicode.Cf, r) || unicode.Is(unicode.Zl, r) || unicode.Is(unicode.Zp, r) {
+			return false
+		}
+	}
+	return true
 }
 
 // exactJSONStringArg mirrors ZERO's map-based tool argument decoding: only the
--- a/internal/acp/translate.go
+++ b/internal/acp/translate.go
@@ -73,6 +73,23 @@ func browserToolDetails(name string) (*BrowserToolDetails, bool) {
 	}
 }
 
+const zeroBrowserMetaKey = "github.com/Gitlawb/zero/browser"
+
+// attachBrowserToolDetails stores ZERO's browser descriptor in ACP's reserved
+// extension channel. Keeping this in one helper prevents start, result, and
+// permission payloads from drifting onto different wire shapes.
+func attachBrowserToolDetails(update *ToolCallUpdate, name string) {
+	browser, ok := browserToolDetails(name)
+	if !ok {
+		return
+	}
+	raw, err := json.Marshal(browser)
+	if err != nil {
+		return
+	}
+	update.Meta = map[string]json.RawMessage{zeroBrowserMetaKey: raw}
+}
+
 // browserToolTitle avoids putting browser_type text, an attached DevTools
 // endpoint, or a URL query/fragment in a tool-card title. Those values can
 // carry credentials or session data; the UI only needs the operation and, for
@@ -178,9 +195,7 @@ func toolCallStart(call agent.ToolCall) ToolCallUpdate {
 		Status:        ToolStatusInProgress,
 		RawInput:      rawInput(call.Arguments),
 	}
-	if browser, ok := browserToolDetails(call.Name); ok {
-		upd.Browser = browser
-	}
+	attachBrowserToolDetails(&upd, call.Name)
 	return upd
 }
 
@@ -201,9 +216,7 @@ func toolCallResult(result agent.ToolResult) ToolCallUpdate {
 	if locs := toolResultLocations(result); len(locs) > 0 {
 		upd.Locations = locs
 	}
-	if browser, ok := browserToolDetails(result.Name); ok {
-		upd.Browser = browser
-	}
+	attachBrowserToolDetails(&upd, result.Name)
 	return upd
 }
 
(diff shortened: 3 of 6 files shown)


Vasanthdev2004 (changes requested):
The URL redaction here is genuinely good and I want it landed. Requesting changes on the title rewrite, which drops disclosure the approver needs.

## What works, and it is more than the body claims

Driven through `permissionToolCall`, head against base:

```
https://alice:hunter2@intranet.corp/path   head "browser open https://intranet.corp"
                                           base "browser_open https://alice:hunter2@intranet.corp/path"
https://ok.test/p?token=SECRET123#f=...    head "browser open https://ok.test"
                                           base "browser_open https://ok.test/p?token=SECRET123#f=SECRET456"
```

Credentials, query and fragment all gone. The path goes too, which the body does not claim, and that matters because secrets live in paths as often as in queries. `https://www.paypal.com@evil.example/` renders as the true host rather than the deceptive spelling, and a URL with an embedded CR no longer puts a raw control byte in the title. Those are real improvements.

One correction for the body: it says this stops typed text and CDP endpoints reaching titles. Base never leaked those. `primaryArgHint`'s key list is path/file_path/pattern/query/command/url/cwd, which never matched `text`, `target`, `key` or `selector`, so `browser_type` and `browser_connect` already rendered bare. The real leak closed is `browser_open`'s URL. Worth rewording so the next reader does not over-trust it.

## Blocking: every `browser_action` renders as the same six characters

```
cmd=eval                 head "browser action"   base "browser_action eval"
cmd=keyboard_insert_text head "browser action"   base "browser_action keyboard_insert_text"
```

`browserToolTitle` does `if command != "open" { return "browser " + command }`, and `command` is the tool-name suffix `"action"`, so the `args["command"]` enum never reaches the title.

Judging it on its own terms rather than against base: `RequestPermissionParams` is `{SessionID, ToolCall, Options}`, so Title plus rawInput is the whole of what the editor user gets. `browser_action` is prompt-gated and `permissionSupportsPersistentDecision` is false for it, so the user is asked **every single call** with that string as the decision label, and `eval` is indistinguishable from `get_title`. The TUI still discloses it, which puts ACP below the terminal on the same decision, against the invariant `permission.go:66-68` states for itself.

The redaction rationale does not apply here either: `command` is a closed enum validated against `browserActionSpecs`, so it can never carry attacker text. Put it back.

## Bare-host `browser_open` loses the destination entirely

```
url="evil.example.test/pay"   deriveBrowserHostScope="evil.example.test"   title="browser open"
url="example.com"             deriveBrowserHostScope="example.com"         title="browser open"
```

Three places derive a host from that one argument and only two agree. `browserOpenURLArg` prepends `https://` when there is no scheme, and the schema documents it. `deriveBrowserHostScope` does the same. `browserToolTitle` bails on an empty scheme and prints nothing, so the approver sees the bare words "browser open" while the browser navigates to `https://evil.example.test/pay`. Normalize the way the tool already does before parsing.

## The title can name an origin the tool will not open

`browserToolTitle` decodes into a struct with a `json:"url"` tag, and `encoding/json` matches tags case-insensitively with last-key-wins, while the tool reads the map key exactly.

```
{"URL":"https://evil.com/x"}   title "browser open https://evil.com"
                               browserOpenURLArg -> err "url is required", never navigates
```

So the prompt names a host that will not be visited. And on the start path, where raw JSON key order survives:

```
{"url":"https://a.com/x","URL":"https://b.com/x"}  ->  title "browser open https://b.com"
{"URL":"https://b.com/x","url":"https://a.com/x"}  ->  title "browser open https://a.com"
```

The title follows whichever key came last; the tool always opens the `url` one.

Being precise about reach, because I checked both paths: on the **permission** path this collision is currently harmless, because `marshalArgs` sorts keys so `url` always lands after `URL`. It is the **start** path that diverges. The `URL`-only case is wrong on both. Using the same exact-key lookup the tool uses fixes all of it and removes the dependence on map ordering, which is not a property worth relying on.

## Titles are now unbounded

The browser branch returns before `truncateHint`:

```
hostlen    100  head title    126   base 76
hostlen  10000  head title  10026   base 76
hostlen 100000  head title 100026   base 76
```

Linear, no bound, and it reproduces on the permission surface too. `read_file` with a 4000-character path is still 73 bytes on head, so this is specific to the new branch. Arguments are model-generated, so it is reachable from injected page content, and the title is emitted before execution so a host that will never resolve still ships in full.

The same bypass breaks the UTF-8 invariant: `{"url":"https://%FF%FE.com/"}` yields an invalid-UTF-8 title that marshals as `browser open https://��.com`, a host nobody sent. `TestToolTitleTruncateHintRuneSafe` pins both invariants but only for `read_file`, so it passes.

## Nits

The nine-command switch in `browserToolDetails` is a fourth copy of the browser-tool inventory, plus a fifth in prose in the doc comment. All nine agree today, and `internal/acp` already imports `internal/tools`, so a contract test would be cheap and would replace the hardcoded names in the new tests.

Not tested and genuinely a question rather than a finding: ACP vendor extensions normally live under `_meta`, and this adds a bare top-level `browser` key. There is no `_meta` usage anywhere in `internal/acp` and I had no strict-schema client to test against, so I am not claiming anything breaks.

## Checked and correct

For any `scheme://` input the title's host cannot diverge from the host the tool opens, since both call `url.Parse` on the same string. Percent-encoded hosts fail closed on both sides. `[::1]:8443` and backslash-userinfo forms behave. The descriptor itself is sound: `{Version, Command}` with `Command` from a fixed allowlist that never touches request data, the nine names match `NewLocalBrowserTools` exactly, MCP collision is structurally impossible since `registryToolName` always returns the `mcp_` prefix, and `browser_screenshot`/`browser_pdf` correctly get no descriptor. All three `ToolCallUpdate` constructors set it and a fresh struct is allocated per call. Both new tests are load-bearing: replacing the title body fails the presentation-safety test, and replacing the allowlist fails the MCP test.

One thought rather than a finding: promoting the host to be the only thing in the title makes it read more like a verified origin than base's raw spelling did. Homograph hosts render the same on both sides, so it is not a regression, but the framing invites more trust than the string earns.

Build, vet, gofmt and `go test ./internal/acp/` all clean on head, linux and darwin included.

Vasanthdev2004 (dismissed):
Approving at `0b9c996e`. All four are fixed, and the way you fixed the third is better than what I asked for.

Re-ran the same probes:

```
browser_action eval                   "browser action eval"
browser_action keyboard_insert_text   "browser action keyboard_insert_text"
url="evil.example.test/pay"           "browser open https://evil.example.test"
{"url":"a.com/x","URL":"b.com/x"}     "browser open https://a.com"      (matches the tool)
{"URL":"https://evil.com/x"}          "browser open"                    (names nothing; tool refuses)
300-char host                         truncated at 76 with the ellipsis
```

I asked you to use the same exact-key lookup the tool uses. Extracting `NormalizedBrowserActionCommand` and `NormalizeBrowserOpenURL` and having ACP call them is the better answer, because it removes the second representation instead of adding a matching copy of it. The bare-host case now agrees with `deriveBrowserHostScope` and `browserOpenURLArg` by construction rather than by three implementations happening to agree.

The refactor is behaviour-preserving on the tool side: `browserActionArgs` gets the same command and spec it did before, and `go test ./internal/tools/ -run Browser` passes.

The fix is pinned rather than incidental. Breaking the command lookup fails `TestBrowserToolUpdatesAreStructuredAndPresentationSafe`.

And the redaction I liked in the first place still holds: credentials, query, fragment and path all stay out of the title, and `https://www.paypal.com@evil.example/` still renders the true host.

Two things I raised that are not code and are still worth doing, neither blocking:

The PR body says this stops typed text and CDP endpoints reaching titles. Base never leaked those, since `primaryArgHint`'s key list never matched `text`, `target`, `key` or `selector`. The real leak closed is `browser_open`'s URL, and the body reads stronger than that.

The nine-command switch in `browserToolDetails` is still a separate copy of the browser-tool inventory. `internal/acp` already imports `internal/tools`, and you have just exported the normalizer, so a contract test asserting the switch against `NewLocalBrowserTools` is cheap now.

Build, vet, gofmt and `go test ./internal/acp/` all clean.

jatmn (changes requested):

## Findings

- [P2] Put the browser extension under ACP `_meta`
  `internal/acp/types.go:217`
  ZERO negotiates ACP v1, whose extensibility contract reserves custom data for `_meta` and says implementations must not add custom fields at the root of protocol types. This PR instead adds `json:"browser"` directly to `ToolCallUpdate`, so every recognized browser call emits a non-protocol root member from `toolCallStart`, `toolCallResult`, and `permissionToolCall`. Generated ACP tool-call models have no such member: an SDK-based client or relay drops it while decoding and re-encoding the payload, and a strict consumer may reject the payload. Only a bespoke raw-JSON decoder that knows ZERO's private shape would retain it.

  The root cause is that the standard ACP wire type and ZERO-specific extension data are modeled as one flat Go struct. The new serialization test does not exercise the boundary: it unmarshals into another local struct that already declares `json:"browser"`, so it proves only that ZERO can read the same private field it wrote. Model ACP's extension channel explicitly instead—for example, attach the existing `{version, command}` descriptor beneath a stable ZERO-owned key in `_meta`—and use one helper to populate that extension for start, result, and permission payloads so those paths cannot drift.

  Add a regression test that serializes each of those three payloads, passes it through an ACP protocol-shaped/generated decoder and encoder, and verifies that the namespaced descriptor survives. The same coverage should verify that non-browser tools omit the extension and that the existing privacy guarantees remain intact: no typed text, CDP target, URL path, query, fragment, credentials, or other request data may enter the descriptor.

jatmn (changes requested):
I found one issue that needs to be addressed before this is ready.

## Findings

- [P2] Keep decoded Unicode controls out of browser permission titles
  `internal/acp/translate.go:121`
  The new title path parses the normalized URL and rebuilds the displayed origin from `u.Host`. `net/url` has already percent-decoded valid UTF-8 escapes at that point, but the only presentation check is `utf8.ValidString`, which verifies encoding validity rather than whether the string is safe to render. A `browser_open` URL containing `safe.example%E2%80%AEevil.test` therefore passes the shared pre-permission URL validation and produces a title containing an actual U+202E bidi override. U+2066 (bidi isolate), U+0085 (C1 control), and U+2028 (line separator) reach the title through the same path. After ACP serialization and decoding, clients receive those actual runes in the approval label, where they can reorder or split the displayed origin; the browser execution path still receives the normalized URL with its percent escapes preserved. Base did not have this behavior because `primaryArgHint` left the `%XX` spelling literal, so the regression is introduced by the new parse-and-reconstruct display path.

  Please address the root cause at the presentation boundary: derive the human-facing origin through one display-safety helper that rejects/falls back or visibly escapes Unicode control, format/bidi, and line/paragraph-separator characters after URL parsing has decoded the host. Do not mutate the normalized URL passed to browser execution or broaden the browser's URL policy as part of this fix, and preserve ordinary safe host display, exact-key matching, userinfo/path/query/fragment redaction, and the existing title-length bound. Add table-driven regressions for the encoded examples above through both `toolCallStart` and `permissionToolCall`, including an ACP marshal/unmarshal round trip that asserts the decoded title contains no disallowed presentation runes and a companion assertion that `NormalizeBrowserOpenURL` still returns the unchanged execution URL. This keeps the correction scoped to the defect while preventing the start, permission, and wire representations from drifting again.

jatmn (approved):
LGTM

anandh8x (approved):
## Review at 33d81a6a

**Verdict: Approve.** Small, well-scoped ACP extension done right, with both prior blockers closed and verified by the original reviewer.

### What it does

Emits an optional, versioned browser descriptor (`{version, command}` under `_meta["github.com/Gitlawb/zero/browser"]`) on `toolCallStart`, `toolCallResult`, and `permissionToolCall` for zero's built-in `browser_*` tools — plus presentation-safe permission titles (operation + origin only, never typed text, CDP endpoints, or URL query/fragment).

### What's good

- **Both P2 blockers are closed.** jatmn's `_meta`-namespacing finding → fixed by `9abf27e` (single `attachBrowserToolDetails` helper, one wire shape across all three paths, protocol-shaped round-trip test). The Unicode bidi/control finding → fixed by `33d81a6` (`browserTitleTextSafe` rejects control, Cf format/bidi, Zl/Zp separators *after* URL decoding, execution URL untouched). jatmn APPROVED after both landed — I verified the fixes match the asks rather than taking the approval at face value.
- **Descriptor carries no request data.** `{version, command}` only; the struct comment explicitly forbids profile data, cookies, typed text, paths/queries, and endpoints. Future fields are constrained by the same comment.
- **MCP-lookalike exclusion is an allowlist, failing closed.** `browserToolDetails` matches `browser_` prefix *plus* a fixed command set (`install`…`action`); a similarly-named MCP tool gets no descriptor and falls back to the generic title. The test `TestBrowserDescriptorDoesNotClaimSimilarlyNamedMCPTools` pins it.
- **Execution path untouched.** `NormalizedBrowserActionCommand` / `NormalizeBrowserOpenURL` are pure extractions — I read both call sites and the `!ok`/scheme/host error branches are identical. New exports serve only the display path, and the comment on the action helper says so explicitly.
- **Exact-key arg matching.** `exactJSONStringArg` mirrors the tool's own decoding (only `"url"`, only string) so an incidental `"URL"` key can't move a permission title the tool would ignore.

### Verification (mine, in an isolated worktree)

- `go build`, `go vet` clean on `internal/acp` + `internal/tools`
- 24 browser tests pass; full `internal/acp` suite passes, no failures

### Two non-blocking notes

1. `attachBrowserToolDetails` *replaces* `update.Meta` rather than merging. Fine today (single writer, verified by grep) — but if a second `_meta` extension ever lands, this silently clobbers it. A merge-on-write or a comment stating the single-writer assumption would future-proof it.
2. The `browserToolDetails` command allowlist will silently omit any future `browser_*` tool (fail-closed → generic title, no descriptor). Safe direction; just noting that adding a browser tool now requires touching the allowlist too, with no compiler enforcing it.

Neither blocks. Merge-ready.

PR #1003, opened 2026-09-02
fix(acp): make configured MCP tools callable

### LINKED ISSUE #998: MCP tools are not callable
### Version / branch / commit

0.8.0 on WIndows 11, zero dev on WSL

### OS and environment

Windows 11 & WSL / Windows Terminal / npm on Windows 11 / go install on WSL

### Steps to reproduce

1. Tell the agent to use an mcp tool, e.g. search something using the exa mcp.
2. Agent says tool is discoverable and can see the schema but never surfaced as invocable functions in its toolset.

### Expected behavior

Agent can use mcp tools after discovering the tool name and its schema.

### Actual behavior

Cannot use MCP tools, even from manually added servers.

_No response_

gnanam1990:
Fixes #998.

ACP now builds the same scoped MCP-enabled registry as `zero exec` for each turn. Configured MCP servers run through the sandbox execution runner, respect workspace trust, use low autonomy, are omitted in plan mode, and are closed after success, failure, or cancellation. `tool_search` is registered after the full ACP-visible registry is built.

Verification:
- `make fmt-check`, `go vet ./...`, `go test ./...`
- focused ACP/CLI race checks (`-count=20`)
- `go run ./cmd/zero-release build` and `smoke`
- `make lint-static`, `make vulncheck`, diff check
- packaged ZeroApp smoke: ACP handshake, workspace open, environment/sidebar/terminal panels

The storage-heavy mutation sweep was intentionally not run under the active storage-safety restriction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of workspace resources after successful, failed, or cancelled agent turns.
  * Permission-prompt cancellations are now reported with the correct cancelled status.
  * Ensured partially initialized tool runtimes are cleaned up when setup fails.

* **Behavior Updates**
  * MCP tools are limited in plan mode and enabled according to workspace permissions.
  * Deferred tool search is now available in ACP workspaces.
  * Workspace setup warnings are now surfaced with sensitive details redacted.
  * MCP tools can appear in auto-mode interfaces while still requiring permission to run.

### DIFF
--- a/internal/acp/agent.go
+++ b/internal/acp/agent.go
@@ -30,15 +30,36 @@ type Deps struct {
 	DiscoverModels func(context.Context, config.ProviderProfile) ([]providermodeldiscovery.Model, error)
 	NewProvider    func(profile config.ProviderProfile) (zeroruntime.Provider, error)
 	RunAgent       func(ctx context.Context, prompt string, provider zeroruntime.Provider, opts agent.Options) (agent.Result, error)
-	// BuildWorkspace builds the SCOPED tool registry and the sandbox engine for a
-	// validated workspace root, so ACP shell tools (bash/exec_command) are confined
-	// exactly like the exec surface — never run unconfined on the host.
-	BuildWorkspace func(workspaceRoot string, resolved config.ResolvedConfig) (*tools.Registry, *sandbox.Engine, error)
+	// BuildWorkspace creates the per-turn scoped tool workspace for a validated
+	// workspace root. Its Close method releases any resources the registry owns
+	// (notably MCP server connections) after the turn completes or is cancelled.
+	// Shell tools remain confined exactly like the exec surface — never unconfined
+	// on the host.
+	BuildWorkspace func(ctx context.Context, workspaceRoot string, resolved config.ResolvedConfig, mode agent.PermissionMode) (*Workspace, error)
 	// ResolveWorkspaceRoot validates + normalizes a client-supplied cwd (must be an
 	// existing directory; never the bare root). It is the file-tool confinement root.
 	ResolveWorkspaceRoot func(cwd string) (string, error)
 	Store                *sessions.Store
 	AgentInfo            Implementation
+}
+
+// Workspace is the per-turn execution environment passed to the agent. ACP does
+// not retain it across turns because MCP connections belong to the registry that
+// advertised their tools; keeping a stale registry after a cancelled turn would
+// leak its server process and its permission state into the next turn.
+type Workspace struct {
+	Registry *tools.Registry
+	Sandbox  *sandbox.Engine
+	Cleanup  func() error
+}
+
+// Close releases resources created with the workspace. A nil cleanup is valid
+// for core-only workspaces.
+func (w *Workspace) Close() error {
+	if w == nil || w.Cleanup == nil {
+		return nil
+	}
+	return w.Cleanup()
 }
 
 // Agent is the ACP agent server bound to one JSON-RPC connection (one editor).
@@ -241,12 +262,28 @@ func (a *Agent) runTurn(ctx context.Context, sess *acpSession, userText string, 
 	if err != nil {
 		return "", RPCError(codeInternalError, "provider: "+err.Error())
 	}
+	mode := sess.currentMode()
 	// Build the SCOPED registry + sandbox engine for this session's workspace so
 	// shell/file tools are confined to the workspace exactly like the exec surface.
-	registry, sandboxEngine, err := a.deps.BuildWorkspace(sess.cwd, resolved)
+	// The workspace can also own MCP connections, which must be closed after every
+	// turn even if the agent run returns an error or the client cancels it.
+	workspace, err := a.deps.BuildWorkspace(ctx, sess.cwd, resolved, mode)
 	if err != nil {
 		return "", RPCError(codeInternalError, "workspace: "+err.Error())
 	}
+	if workspace == nil || workspace.Registry == nil {
+		if workspace != nil {
+			_ = workspace.Close()
+		}
+		return "", RPCError(codeInternalError, "workspace: missing tool registry")
+	}
+	defer func() {
+		if err := workspace.Close(); err != nil {
+			log.Printf("acp: close workspace: %v", err)
+		}
+	}()
+	registry := workspace.Registry
+	sandboxEngine := workspace.Sandbox
 	note := &notifier{conn: a.conn, sessionID: sess.id}
 
 	opts := agent.Options{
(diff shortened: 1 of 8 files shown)


jatmn (changes requested):

## Findings

- [P1] Make newly configured MCP tools visible in the default ACP mode
  `internal/cli/acp.go:97`
  This registers real MCP tools into the per-turn workspace, but a fresh ACP session starts in `PermissionModeAuto`, while `mcp.newRegistryTool` gives an unapproved tool `PermissionPrompt` without `AdvertiseInAuto`. `agent.ToolAdvertised` therefore filters the tool out before the provider sees it. The same visibility result is used by `registerToolSearchIfEligible`, so a fresh set of MCP tools also contributes zero eligible tools and does not produce a discovery loader. The new test misses the production path because `cliFakeDeferredTool` reports `PermissionAllow`; registry membership consequently passes even though the provider-visible set for a real fresh tool is empty.

  The root cause is the mismatch between ACP's default mode, the safety metadata assigned to MCP tools, and the visibility-aware loader count—not MCP connection or registration itself. Please align those seams so a newly configured MCP tool is discoverable and callable through the ordinary ACP workflow while its execution still requires the existing low-autonomy permission prompt. Avoid fixing this by converting MCP actions to pre-approved actions, weakening plan mode, or globally changing unrelated surfaces. Add an end-to-end regression using production-equivalent `PermissionPrompt` MCP safety and assert what the ACP provider actually receives in the default mode.

- [P2] Pass one effective defer threshold through both ACP consumers
  `internal/acp/agent.go:296`
  `buildACPWorkspace` reads `resolved.Tools.DeferThreshold` and may register `tool_search`, but this `agent.Options` literal does not pass the same value to the agent. Its zero value reaches `partitionToolsCached`, where deferral is active only when `options.DeferThreshold > 0`; the inactive path removes `tool_search` and eagerly advertises every visible MCP schema. As a result, the loader asserted by `TestBuildACPWorkspaceRegistersSandboxedMCPToolsAndClosesThem` is never provider-visible, the default threshold of 3 is ignored, and larger MCP installations pay the full schema/context cost on every turn.

  The root cause is split ownership of a single configuration contract: workspace construction consumes the threshold, while the downstream partition independently receives an unset value. Please derive one effective ACP threshold and carry it through loader registration and `agent.Options`, as the sibling exec/TUI paths do. Cover the complete behavior—not only registry membership—by asserting that at/above threshold the provider receives `tool_search` while eligible schemas remain deferred. Preserve explicit zero-disable, below-threshold eager exposure, and plan-mode suppression; this does not require redesigning generic deferral.

- [P2] Preserve and surface MCP setup diagnostics across the workspace boundary
  `internal/cli/acp.go:97`
  The helper returns two non-fatal diagnostic channels: `trustSkip` explains when project MCP configuration was excluded by the fail-closed trust gate, and `runtime.Skipped()` records configured servers that timed out, failed to connect, returned invalid tools, or conflicted by name. This new caller assigns the trust result to `_` and retains only `runtime.Close`, so both outcomes are discarded. The turn then proceeds with a reduced registry and no explanation; to an ACP user, an untrusted project or failed configured server is indistinguishable from the original “MCP tools are not callable” defect. Other consumers explicitly emit the `zero trust` guidance and redacted configured-server warnings.

  The root cause is an incomplete workspace result/ownership contract: `Workspace` carries the registry, sandbox, and cleanup callback, but no setup notices survive long enough for the ACP boundary to report them. Please retain the trust and configured-server skip results and surface them through an ACP-safe, redacted diagnostic path. The implementation can choose the appropriate ACP notification or stderr route, but it must not corrupt stdout JSON-RPC framing. Keep trust fail-closed, keep multi-server startup best-effort, keep hard setup errors as errors, and keep unconfigured default-server failures silent; making optional server failure fatal or bypassing trust would introduce scope and behavioral drift.

jatmn (approved):
lgtm

Vasanthdev2004 (approved):
Reviewed at dba66b0. Approving. This is careful work and the lifecycle handling in particular is better than the usual first attempt at this shape.

I checked the parts that could have been holes rather than reading for style:

`AdvertiseInAuto: true` on prompt-gated MCP tools is the thing I most wanted to be wrong, and it is not. The flag appears only in the advertisement path and never in `shouldRequestPermission`, so a tool becomes visible in auto mode while execution still prompts, which is the two-layer gate this repo already uses for exec_command and web_fetch. `TestRegistryToolCanAdvertiseInAutoWithoutPreapproval` pins exactly that, which is the right test to have written. I confirmed separately that ACP really can prompt: `OnPermissionRequest` is wired through to `session/request_permission`, so these tools reach a decision rather than a dead end.

The plan-mode gate is complete rather than merely plausible. `setSessionMode` accepts only auto, ask and plan, so `mode != PermissionModePlan` covers the whole mode set; there is no third read-only mode that would quietly get MCP servers spun up.

No nil-runtime hazard on the paths that assign `workspace.Cleanup = runtime.Close`: the no-servers case returns `noopMCPRuntime{}` rather than nil, and `(*Runtime).Close` handles a nil receiver, so the typed-nil case is safe too. Closing a partial runtime when registration fails part way is the right call and it has a test.

I also diffed the test files for fixtures edited to accommodate the change and found none, which is worth saying because that is how this kind of PR usually goes green.

Locally: `internal/acp` and `internal/mcp` pass. The one `internal/cli` failure is `TestBuildServeScopeKeepsLexicalPaths`, which fails identically on current main without this branch, because `os.Symlink` needs a privilege an unelevated Windows box does not have. Not yours. Arguably that test should skip rather than fail in that case, the way its neighbours do, but that is a separate cleanup.

Three notes, none blocking.

**Per-turn MCP lifecycle is the one I would think about twice.** `BuildWorkspace` runs per turn, so every configured stdio MCP server is launched and killed on every message. The doc comment explains why the workspace is not retained, and the reason is real: a stale registry after a cancelled turn leaks its server process. But the middle ground is not discussed anywhere in the PR, namely session-scoped connections closed on session end and on cancel. In an editor this is the latency-sensitive surface, and a server that indexes, authenticates or warms a cache at startup now pays that on every message. I am fine landing it this way, since correctness before performance is the right order, but it is worth a sentence recording that the tradeoff was chosen rather than inherited.

**Cancellation is the one claimed case with no test.** The PR body and the `Workspace` doc both name it specifically, and the doc uses it as the motivation for the whole design, yet the test is `...AfterSuccessAndFailure`. The `defer` is registered immediately after the nil check so I believe the behaviour is correct; it is the case the design is argued from that is unpinned. Cheap to add.

**"matches exec's registry construction" is not quite true and should say so.** `buildACPWorkspace` sets `AdvertiseInAuto: true` where the exec path does not. That difference is defensible, an editor surface wants the model to see MCP tools and the prompt still gates them, but the comment currently claims parity without noting the exception, and that is exactly the kind of comment someone later trusts instead of reading the code.

Vasanthdev2004:
The symlink test failure I ruled out as environmental during review is now #1021, so the next contributor on a stock Windows box does not have to re-diagnose it.

PR #886, opened 2026-08-09
fix(sandbox): guard the Windows write-jail invariant and disclose the DenyRead trade

Vasanthdev2004:
Partial work on #869. It does **not** close it, and I would rather say that up front than have the checkbox suggest otherwise.

## The regression risk

#865 removed the World SID from the `WRITE_RESTRICTED` token. That is the whole write jail: every principal carries Everyone, so while it was a restricting SID the write half of the access check passed for free on any Everyone-writable path, and confinement fell back to the user's own permissions.

That fix has no CI protection. The only test covering it, `TestWindowsRestrictedTokenDeniesWritesToEveryoneWritablePaths`, sits behind `ZERO_SANDBOX_REAL_SMOKE=1`, and `rg ZERO_SANDBOX_REAL_SMOKE .github/` comes back empty. So anything that restored the unconditional World SID would go green. This is not hypothetical: #640's branch predates #865 and conflicts on that exact hunk.

`CreateRestrictedToken` works unelevated against the caller's own token, so there was never a reason this needed the real-runner harness. Four unit tests now read the token's restricted-SID list directly:

- the `WRITE_RESTRICTED` token must not carry the World SID
- neither shape may carry `Users`, `Authenticated Users`, `INTERACTIVE`, `BATCH`, `Administrators`, `SYSTEM`, `SERVICE`, `NETWORK`, or the user's own SID. #869 names these as the ones that would reopen the same class of bypass, and the runner's comment already states the rule
- the capability SID must be present, so a token that passed by having no keys at all would still fail
- the non-`WRITE_RESTRICTED` shape still carries the World SID

The last one documents the open gap instead of asserting the end state. It skips with a note if that stops being true, so whoever closes #869 gets told to replace it rather than finding a mystery failure.

Mutation-verified: flipping the guard back to unconditional produces

```
the World SID is a restricting SID on the write-restricted token, which collapses the write jail:
[S-1-5-21-... S-1-5-5-0-426223 S-1-1-0]
```

and the production file is byte-identical to `main` afterwards.

## The invisible trade

Setting `denyRead` selects the token shape without `WRITE_RESTRICTED`, because the restricted-SID check has to cover reads for read-deny to mean anything, and that shape has to keep the World SID or the token cannot open `cmd.exe`. The trade is deliberate and well documented in the token source. It was just never surfaced: someone who set `denyRead` to protect credentials had no way to learn they had given up write confinement to get it.

[...]

PR #1005, opened 2026-09-03
fix(sandbox): take Everyone out of the restricted-SID list (does not work yet)

### LINKED ISSUE #869: Windows write jail is still bypassable on profiles that set denyRead
#865 removed the World SID (`S-1-1-0`, Everyone) from the `WRITE_RESTRICTED` token, which is the shape every Windows user gets by default. It could not remove it from the other shape, and that one is still bypassable.

A profile with a non-empty `denyRead` builds its restricted token without `WRITE_RESTRICTED`, and Everyone stays on the restricted-SID list. A restricted SID is a key to every object whose DACL names it, and every principal carries Everyone, so the restricted-SID check passes for free on any path whose DACL grants Everyone write. The workspace write jail then falls back to the ordinary user's own permissions, which is the exact boundary this token exists to be stricter than.

No privilege, no symlink and no race is needed. An Everyone-writable directory is enough, and share roots opened with `Everyone:F` plus loose installer ACLs supply them.

Site: `createWindowsRestrictedTokenFromBase` in `internal/sandbox/windows_token_windows.go`.

**Who is affected.** Only Windows users who set `denyRead` themselves. Zero never populates it on Windows: `credentialDenyReadPaths` returns nil there (`internal/sandbox/profile.go:166`), so the default posture runs the token #865 fixed.

**Why #865 could not cover this shape.** Without `WRITE_RESTRICTED` the restricted-SID check applies to reads as well as writes, and default Windows DACLs grant `BUILTIN\Users` rather than anything on this list. A token without Everyone cannot open cmd.exe: the process dies at launch with `STATUS_ACCESS_DENIED` (0xC0000022) before it runs anything. That was tried first and it breaks the unelevated sandbox outright. So #865 fixed the shape where nothing depends on Everyone and left this one documented in the code.

**What closing it needs.** A read-side grant that is not a universal group. AppContainer or LPAC with an explicit capability SID is the obvious candidate. A narrower interim: keep Everyone off the restricted-SID list and grant read through the per-sandbox capability SID on the executable search paths, trading machine setup for enforcement.

**Same class, worth auditing at the same time.** `Authenticated Users`, `Users`, `INTERACTIVE` and `BATCH` each produce this bypass if they reach that list. #640 adds `Users` and `Authenticated Users` behind a `broadenReadSIDs` parameter that its own doc comment says callers must always pass false, plus a guard refusing to combine it with `writeRestricted`. That is the right posture, and it needs to stay that way.

**Do not close it by widening the default.** #662 asks for the credential deny-read defaults to apply on Windows. That makes `denyRead` non-empty for every Windows user and moves all of them onto this token shape, turning a config almost nobody hits into the default one. #662 needs this fixed first, not the other way round.

Related: #865 (fix for the default shape), #612 (why DenyRead uses the non-`WRITE_RESTRICTED` token), #640 (proposes rejecting `denyRead` on these tiers outright, which is a different trade for the same problem).

Vasanthdev2004:
Closes #869. Stacked on #808, which is where the read capability comes from; the base is `feat/windows-sandbox-identity`, so review only the top commit.

### What was wrong

A profile that sets `denyRead` drops `WRITE_RESTRICTED`, and the token that replaces it carried the World SID. Every principal carries Everyone, so the restricted-SID check passed for free on any path whose DACL grants Everyone write, and the workspace write jail fell back to the caller's own permissions. That is the boundary the token exists to be stricter than. No privilege, no symlink and no race is needed: an Everyone-writable directory is enough, and share roots opened Everyone:F and loose installer trees supply them.

#865 closed this for the `WRITE_RESTRICTED` token and left this half open, which is what #869 tracked.

### Why it could not just be deleted

Without `WRITE_RESTRICTED` the restricted-SID check covers reads too, and default Windows DACLs grant `BUILTIN\Users` rather than anything in the list. I impersonated both token shapes against `C:\Windows\System32\cmd.exe`:

```
strict (no WRITE_RESTRICTED, capability only): Access is denied.
WRITE_RESTRICTED:                              <nil>
```

So deleting the SID on its own turns every `denyRead` profile into a command that dies at launch with a bare access denial.

### The fix

The read capability takes its place. `BuildWindowsACLPlan` already grants that SID on every read root and denies it on every `denyRead` path, gated on the same field that picks the strict token, so the read allowance and the restriction become one decision and the SID names only what setup granted. The principal path already worked this way; this brings the capability path in line.

Elevated setup is what puts that ACE on the volume root the production profile seeds, and #808 already makes the unelevated tier refuse a `denyRead` profile up front, so nothing reaches the token expecting a grant nobody applied.

### Tests

`TestTheStrictTokenCannotWriteAnEveryoneWritableDirectory` builds a real restricted token, protects a directory with an Everyone-only DACL, impersonates and attempts a write. It first writes a directory the capability does grant, so a token that can write nowhere would not satisfy it. With the World SID put back and the SID-list assertion silenced, the write succeeds:

```
the sandboxed token wrote a directory outside every write root, because its
DACL grants Everyone and Everyone is one of the token's restricting SIDs
```

`TestThePlanAndTheTokenAgreeOnTheReadCapability` pins the two halves together. The plan asks about `denyRead` in one file and the runner asks about `writeRestricted` in another, off the same field with nothing joining them; dropping the plan's gate fails it.

Six sandbox tests fail on my box on `origin/main` as well (`TestEvaluateAppliesWriteAllow` and friends). Not from this branch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened Windows sandbox access controls to prevent restricted commands from writing to directories that are broadly writable.
  - Preserved intended read access for strict sandbox profiles while maintaining write restrictions.
  - Command execution now stops with an error when required sandbox access capabilities cannot be established.

- **Tests**
  - Added coverage for Windows sandbox token permissions, read capabilities, and filesystem write enforcement.


Vasanthdev2004:
Two things about #886, which is mine and still open. Whoever merges these needs both.

`internal/sandbox/windows_token_windows_test.go` there has `TestNonWriteRestrictedTokenStillCarriesTheWorldSID`, which asserts the World SID **is** on the strict token. It was written to document the #869 gap rather than the desired end state, and it fails the moment this lands. It has to be flipped to assert the opposite, plus the read capability, in whichever merge is second. I will do that on #886 once #808 and this settle, so I am not pushing to a branch under review to fix a conflict that does not exist yet.

The other one is already handled here: #886 adds a package-level `containsSID`, so mine is named `carriesSID` and the two compile together.

Vasanthdev2004:
Draft. This does not work, and all eight checks are green, so I want the reason written down before anyone spends time reviewing it.

`TestWindowsUnelevatedRealSandboxSmoke` fails on this branch and passes on the base:

```
runner_windows_integration_test.go:183: Windows sandbox command failed: exit status 0xc0000022
```

That is STATUS_ACCESS_DENIED at launch. It is gated behind `ZERO_SANDBOX_REAL_SMOKE=1`, which no workflow sets, so CI cannot see it. I ran it locally with both branches and the same built runner and setup binaries.

Bisected within the branch: disabling the read-capability append while leaving the World SID out fails identically, so it is the World removal that breaks launch and the read capability does not compensate. Which makes sense in hindsight. The profile sets `IncludePlatformRoots`, so System32 is a read root, and no ACL plan can put a capability ACE there. The premise of the PR body above, that the read capability is a drop-in replacement, is wrong for any read root setup does not own.

I also probed a narrower replacement. `ALL APPLICATION PACKAGES` and `ALL RESTRICTED APPLICATION PACKAGES` are read-granted on System32 and are not universal groups, which would have been the shape we want, but `CreateRestrictedToken` rejects both:

```
CreateRestrictedToken: The parameter is incorrect.
```

So the choice on this backend is narrower than I thought. Either a universal group stays in the restricted list and the write jail is void for `denyRead` profiles, which is #869 as filed, or `denyRead` is refused on this backend and the operator is told why. There is no third option I have found that keeps both the jail and the ability to launch.

I am not going to pick that on my own inside a bug-fix PR, because refusing the profile is user-visible and would mean changing the smoke test that currently asserts the opposite. Parking here until #808 settles, since the principal backend is the only other place a non-universal read grant could come from, and I have not verified it can serve a `denyRead` profile either.

One thing worth separating out regardless of how #869 lands: the real-smoke suite is the only coverage for this and CI never runs it. That is how a launch-breaking change got eight green checks.

Vasanthdev2004:
Status, now that #808 has settled the design question this was waiting on.

#808 (`3383fbc6`) refuses any plan carrying a read grant on a directory Zero does not own, at both setup tiers, before the first mutation. A `denyRead` profile therefore cannot complete setup at all, which closes the practical exposure #869 describes: the write jail is no longer silently void for those profiles, because those profiles no longer run.

That makes this PR the wrong shape. Its premise was replacing the World SID with the read capability, and I measured that the replacement does not work: a strict token with the read capability granted nowhere cannot open `C:\Windows\System32\cmd.exe`.

I am not simply flipping it to "remove the World SID because the strict token is unreachable" either, because that is not provable. The refusal keys on the read grant landing outside the plan own write roots, and a profile whose read roots all sit inside its write roots would still select `writeRestricted=false`. `permissionProfileReadRoots` always seeds the filesystem root, so production never gets there, but the token shape is not dead code by construction.

So this stays a draft until #808 lands, then it becomes a much smaller change: drop the World SID and let the strict token fail closed, since nothing that can reach it can complete setup anyway. Worth keeping from the current branch either way are the impersonated-write regression and the plan/token equivalence test, which pin properties that survive the redesign.

Vasanthdev2004:
Closing. The approach here is disproven and leaving it open reads as work in progress.

Taking Everyone out of the restricted-SID list does close the write-jail bypass, and it also stops the sandbox launching: `TestWindowsUnelevatedRealSandboxSmoke` dies at `0xc0000022`, because a fully restricted token applies its restricted-SID check to reads as well as writes, so the child cannot open its own executable.

#808 takes the other route and refuses a `denyRead` profile outright rather than serving it insecurely, with a diagnostic that names the reason and points at the issue. Written up in full on #869, which stays open: refusal is the honest interim state, not `denyRead` working on Windows.

PR #1006, opened 2026-09-04
fix(sandbox): keep Windows restricted-token SIDs narrow and fail closed on DenyRead

euxaristia:
## Summary
Windows restricted-token sandboxing cannot support `DenyRead` without access-time confinement because omitting `Users`/`Authenticated Users` prevents system binaries from executing, while adding those groups reopens ambient write access. This PR rejects unsupported `DenyRead` configurations upfront before setup/token creation, migrates legacy `SYNCHRONIZE` DenyWrite ACEs in-place with `NoInherit` support on upgraded hosts, and randomizes shared-directory smoke test probes with unique ownership tokens.

Refs #639

## Changes
- Reject non-empty `DenyRead` profiles upfront in `internal/sandbox/windows_command_runner.go` and `windows_setup_windows.go`.
- Narrow legacy `DenyWrite` ACEs containing `SYNCHRONIZE` in-place while honoring `NoInherit` and preserving co-resident `DenyRead` in `internal/sandbox/windows_acl_apply_windows.go`.
- Preserve `DenyRead` bits and narrow write denial when migrating combined legacy read/write deny ACEs.
- Use collision-resistant probes via `allocateSharedDirectoryProbe` and verify unique marker ownership on cleanup in `internal/sandbox/runner_windows_integration_test.go`.

- Vasanthdev2004: Carried `NoInherit` into `windowsMigrateDenyWriteInDACL` to drop inherit-only ACEs and clear inheritance flags; removed stale references to deleted descendant scanning files from PR description.
- CodeRabbit: Handled non-`os.ErrNotExist` read errors in probe cleanup and validated exact per-probe marker content before removal.
- CodeRabbit: Preserved `DenyRead` bits in `windowsMigrateDenyWriteInDACL` and `windowsFilterDACL` when migrating combined legacy read/write deny ACEs, updated `windowsIsExperimentalWriteDenyMask` to avoid classifying combined masks as pure write-deny masks, and added regression test `TestWindowsACLDenyWriteMigratesCombinedLegacyReadWriteDeny`.

## Test plan
- `go test -v ./internal/sandbox -run TestWindowsACL`
- `go test -v ./internal/sandbox -run TestSharedDirectoryProbeLifecycle`
- `go test ./internal/sandbox`
- `go vet ./internal/sandbox/...`

## Summary by CodeRabbit

- **Bug Fixes**
  - Windows sandbox setup now rejects unsupported `DenyRead` configurations before making system changes or launching commands.
  - Error messages provide clearer guidance without exposing protected paths or suggesting unsupported workarounds.
  - Windows write restrictions now migrate legacy permissions safely while preserving read-denial rules.
  - Shared-directory protections more reliably block unauthorized writes and preserve unrelated files.
- **Documentation**
  - Updated Windows sandbox documentation to clarify `DenyRead` limitations for restricted-token modes.


Vasanthdev2004 (changes requested):
The DenyRead decision is right and I verified it holds on the path that actually reaches it. One thing in the ACL migration needs fixing first.

## The migration silently ignores NoInherit

`prepareWindowsACLPathGroupEntries` migrates a legacy deny in place and then `continue`s, so the requested entry is never applied. The migration patches `ace.Mask` only, never `ace.Header.AceFlags`. On a directory that already carries a legacy inheritable deny, a `NoInherit: true` entry therefore leaves the inheritable ACE inheritable and never installs the direct-only one.

Seeded a legacy broad inheritable deny, then applied `{Action: DenyWrite, NoInherit: true}`:

```
BEFORE dir     type=1 flags=0x0  mask=0x130116
BEFORE dir     type=1 flags=0xb  mask=0x40110000
AFTER  dir     type=1 flags=0x0  mask=0xf0156
AFTER  dir     type=1 flags=0xb  mask=0xf0156     <- still inherit-only
AFTER  child   type=1 flags=0x13 mask=0xf0156     <- child still inherits
```

Control on a directory with no pre-existing deny for that SID:

```
PROBE clean dir   type=1 flags=0x0 mask=0xf0156
PROBE RESULT: NoInherit WORKS on a clean directory
```

So `NoInherit` itself is fine; the migration defeats it, on exactly the upgraded hosts the migration exists for. That also contradicts the invariant this PR states in `dedupeWindowsACLEntries`, that collapsing the two shapes "could silently promote a deliberately non-inherited shared-path deny into an inheritable one". The migration does precisely that.

The consequence is over-application rather than a hole: the child gets a deny it should not have. On a shared system directory that is the propagation hazard the flag was added to avoid, since `SetSecurityInfo` pushes an inheritable ACE onto existing descendants.

Nothing in production emits `NoInherit` today, so this breaks nobody right now. I am still asking for it before merge, because the first caller that sets the flag on a path with a legacy deny gets no error and no effect. Either carry the entry's inheritance into the migrated ACE, or fall through to the normal `EXPLICIT_ACCESS` path when the requested inheritance differs from what is on the object.

## Verified, for what it is worth

The rejection fires where it needs to. `deny_read` is reachable without any config key, through the `request_permissions` tool, and I drove that end to end into the plan builder:

```
PROBE profile DenyRead count = 1
PROBE level=unelevated -> rejected as intended
PROBE level=native     -> rejected as intended
```

Both marker schema versions bumped, which is what makes the migration actually run on upgrade rather than sitting behind a cached marker. Good call.

## Two smaller things

The description says it deletes unused descendant scanning and path resolution files. Those files exist at neither the merge base nor the head, so that bullet looks left over from an earlier revision.

## Overlap you should know about

#808 makes the same call and refuses `denyRead` too, so whichever lands second will conflict in `windows_setup_windows.go` and `windows_command_runner_windows.go`. Yours is better placed than mine: it rejects on the profile at four entry points, where mine keys on the read grant appearing in the built plan. If this lands first I will drop my version and rebase onto yours.

Vasanthdev2004 (approved):
Fixed. Approving at 5c46b75c.

I re-ran the exact probe from my last review rather than reading the diff for it.
Same seed, a broad inheritable legacy deny on a directory with a child, then a
`NoInherit: true` DenyWrite:

```
BEFORE dir     type=1 flags=0x0 mask=0x130116
BEFORE dir     type=1 flags=0xb mask=0x40110000
AFTER  dir     type=1 flags=0x0 mask=0xf0156
AFTER  child   (no ACE for the probe SID)
```

The inherit-only ACE is gone and the child inherits nothing. Previously both
survived and the child kept inheriting.

Carrying the entry's inheritance into the migrated ACE was the right call over
falling through to `EXPLICIT_ACCESS`, since it keeps the ACE in its original
position and avoids the merge you were working around in the first place.

Your tests are real ones. They read back off the applied object rather than
asserting the plan, and I checked that by breaking the code under them: not
clearing the inheritance flags fails the NoInherit test on `retained inheritance
flags: 0x3`, dropping the read-bit preservation fails the combined test on `lost
DenyRead bits or failed to narrow DenyWrite`, and restoring the old
`windowsIsExperimentalWriteDenyMask` fails it on `failed to recognize DenyRead on
migrated combined ACE`.

One that survived, which I am not asking you to change: setting
`shouldDropInheritOnly` to false leaves every test passing. It does not
reintroduce anything, because the NoInherit branch clears the flags anyway; the
only difference is a redundant second direct ACE at `0x1f0156` alongside the
narrow `0xf0156`. So that line is tidiness rather than the load-bearing part, and
it is worth knowing which is which if this code is edited again.

Also checked, since it was the headline and I did not want to stand on a
verification from an older commit: nothing since a84626f4 touches the DenyRead
rejection. The diff is `windows_acl_apply_windows.go`, its test, and the
integration test. The four entry points and both marker schema bumps are
unchanged, so my earlier verification still applies.

The probe ownership change in the integration test is a good catch on your own
work. A shared-directory probe that deletes a file it did not create is a bad
thing to have running against `C:\Users\Public` and `ProgramData`.

Full `internal/sandbox` package green here on Windows, vet and gofmt clean.

On the overlap: #808 refuses `denyRead` too, and yours is better placed, so I will
drop mine and rebase onto this once it lands. Whichever goes second will conflict
in `windows_setup_windows.go` and `windows_command_runner_windows.go`, and that is
mine to resolve, not yours.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #1007, opened 2026-09-04
fix(oauth): Split an oversized keyring token blob across entries

### LINKED ISSUE #937: oauth: keyring storage cannot save a second login on macOS

euxaristia:
## Summary
Multi-provider OAuth state can exceed the macOS keyring entry limit. Split oversized token blobs across alternating chunk generations, using the manifest as the commit point, and provide reset and cleanup recovery. Cleanup now checks manifest ownership so a stale migration marker cannot delete live credentials before a replacement write fails.

Fixes #937

## Changes
- Implement alternating chunk generations with per-entry size budgets and integrity checks.
- Record failed first-migration rollback cleanup durably, retain recovery markers when cleanup fails, and protect chunks owned by the live manifest even when stale-marker deletion fails.
- Fail closed during cleanup when manifest ownership cannot be established.
- Add store reset, publication-residue cleanup, backend-bounded keyring reset, and `auth reset` command completions.
- Derive the shared keyring lock from the resolved OS user home and propagate lock-path resolution failures during store construction.

## Test plan
Validation used Go 1.26.6. The validated Linux worktree's modified-file hashes matched the staged patch exactly.
- Full suite: `go test -json -timeout 5m -p 4 ./...` passed in Ubuntu WSL.
- Full OAuth suite passed with the race detector, cross-compiled for Linux with the existing Zig compiler and executed in Ubuntu WSL.
- Formatting passed using the Makefile's gofmt recipe over tracked files in the Linux worktree.
- `go vet ./...`, `go run ./cmd/zero-release build`, and `go run ./cmd/zero-release smoke` passed.
- Pinned golangci-lint v2.12.2 reported zero issues; pinned govulncheck v1.3.0 reported no vulnerabilities.
- `git diff HEAD --check` passed before commit.
- Windows full-suite attempts encountered temporary-file access and cleanup errors; complete validation was performed in Linux/WSL.

- Reconcile cleanup markers with manifest ownership before deleting chunks, addressing the live-generation deletion reported by Vasanthdev2004 and jatmn.
- Add the complete regression sequence: failed migration and rollback, successful retry while cleanup still fails, then a failed replacement write. Existing logins remain readable whether stale-marker deletion succeeds or fails.
- Add coverage that preserves chunks and the marker when the manifest is malformed.
- Verified the regressions against the original implementation with Go 1.26.6: the lifecycle cases fail with `missing chunk 1 of 2`, and the malformed-manifest case detects unauthorized cleanup. All pass with the fix.

## Summary by CodeRabbit

* **New Features**
  * Added `zero auth reset --confirm` to clear saved OAuth logins and shared MCP tokens.
  * Added help, JSON output, and shell completion support for the reset command.
  * Credential storage now supports token data exceeding platform-specific size limits.

* **Bug Fixes**
  * Improved recovery from incomplete or corrupted credential data.
  * Reset now cleans up leftover credential files and temporary data.
  * Improved handling of credential storage limits across supported platforms.

### DIFF
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -600,7 +600,7 @@ func runAuthReset(args []string, stdout io.Writer, stderr io.Writer, deps appDep
 		}
 		return exitSuccess
 	}
-	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store. All stored credentials and entries cleared."); err != nil {
+	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store."); err != nil {
 		return exitCrash
 	}
 	return exitSuccess
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -74,6 +74,8 @@ func runAuth(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) in
 		return runAuthStatus(args[1:], stdout, stderr, deps)
 	case "refresh":
 		return runAuthRefresh(args[1:], stdout, stderr, deps)
+	case "reset":
+		return runAuthReset(args[1:], stdout, stderr, deps)
 	case "openrouter":
 		return runAuthOpenRouter(args[1:], stdout, stderr, deps)
 	case "chatgpt":
@@ -321,6 +323,7 @@ func validateAuthFlags(sub string, a authArgs) error {
 		"logout":  {"json": true},
 		"status":  {"json": true},
 		"refresh": {"watch": true},
+		"reset":   {"json": true},
 	}[sub]
 	bad := func(name string) error { return fmt.Errorf("zero auth %s does not accept %s", sub, name) }
 	if a.json && !allowed["json"] {
@@ -569,6 +572,40 @@ func runAuthRefreshWatch(manager *oauth.Manager, key, provider string, stdout io
 	return exitSuccess
 }
 
+func runAuthReset(args []string, stdout io.Writer, stderr io.Writer, deps appDeps) int {
+	parsed, err := parseAuthArgs("reset", args)
+	if err != nil {
+		return writeExecUsageError(stderr, err.Error())
+	}
+	if parsed.help {
+		_ = writeAuthHelp(stdout)
+		return exitSuccess
+	}
+	if len(parsed.positional) > 0 {
+		return writeExecUsageError(stderr, fmt.Sprintf("zero auth reset takes no arguments (got %q)", parsed.positional[0]))
+	}
+	manager, err := newAuthManager(deps, stdout)
+	if err != nil {
+		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
+	}
+	if err := manager.Reset(); err != nil {
+		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
+	}
+	if parsed.json {
+		payload := struct {
+			Reset bool `json:"reset"`
+		}{Reset: true}
+		if err := writePrettyJSON(stdout, payload); err != nil {
+			return exitCrash
+		}
+		return exitSuccess
+	}
+	if _, err := fmt.Fprintln(stdout, "Reset OAuth token store. All stored credentials and entries cleared."); err != nil {
+		return exitCrash
+	}
+	return exitSuccess
+}
+
 func filterAuthStatuses(statuses []oauth.Status, provider string) []oauth.Status {
 	want := oauth.ProviderKey(provider)
 	filtered := make([]oauth.Status, 0, 1)
@@ -589,6 +626,7 @@ Commands:
   logout <provider>                               Delete a provider's stored login
   status [provider]                               Show login presence/expiry (never the token)
   refresh <provider> [--watch]                    Force a token refresh (--watch keeps it fresh)
+  reset                                           Reset the OAuth token store (clears corrupted entries and tokens)
   openrouter                                      Log in to OpenRouter in the browser; mints an API key
   chatgpt                                         Log in to ChatGPT in the browser (Codex backend, ChatGPT Plus/Pro)
 
@@ -615,7 +653,7 @@ Flags:
       --device   Use the device-code flow (headless/SSH; no browser)
       --scope    Add an OAuth scope (repeatable)
       --watch    Keep the token fresh in the foreground (refresh only)
-      --json     Print result as JSON (status/logout)
+      --json     Print result as JSON (status/logout/reset)
   -h, --help     Show this help
 `)
 	return err
--- a/internal/cli/auth.go
+++ b/internal/cli/auth.go
@@ -263,6 +263,7 @@ type authArgs struct {
 	json       bool
 	device     bool
 	watch      bool
+	confirm    bool
 	scopes     []string
 	help       bool
 }
@@ -288,6 +289,8 @@ func parseAuthArgs(sub string, args []string) (authArgs, error) {
 			parsed.device = true
 		case arg == "--watch":
 			parsed.watch = true
+		case arg == "--confirm":
+			parsed.confirm = true
 		case arg == "--scope":
 			if i+1 >= len(args) {
 				return authArgs{}, fmt.Errorf("--scope requires a value")
@@ -323,7 +326,7 @@ func validateAuthFlags(sub string, a authArgs) error {
 		"logout":  {"json": true},
 		"status":  {"json": true},
 		"refresh": {"watch": true},
-		"reset":   {"json": true},
+		"reset":   {"json": true, "confirm": true},
 	}[sub]
 	bad := func(name string) error { return fmt.Errorf("zero auth %s does not accept %s", sub, name) }
 	if a.json && !allowed["json"] {
@@ -334,6 +337,9 @@ func validateAuthFlags(sub string, a authArgs) error {
 	}
 	if a.watch && !allowed["watch"] {
 		return bad("--watch")
+	}
+	if a.confirm && !allowed["confirm"] {
+		return bad("--confirm")
 	}
 	if len(a.scopes) > 0 && !allowed["scope"] {
 		return bad("--scope")
@@ -584,6 +590,9 @@ func runAuthReset(args []string, stdout io.Writer, stderr io.Writer, deps appDep
 	if len(parsed.positional) > 0 {
 		return writeExecUsageError(stderr, fmt.Sprintf("zero auth reset takes no arguments (got %q)", parsed.positional[0]))
 	}
+	if !parsed.confirm {
+		return writeExecUsageError(stderr, "zero auth reset clears all stored OAuth logins and shared MCP tokens; rerun with --confirm to reset")
+	}
 	manager, err := newAuthManager(deps, stdout)
 	if err != nil {
 		return writeAppError(stderr, redaction.ErrorMessage(err, redaction.Options{}), exitCrash)
@@ -626,7 +635,7 @@ Commands:
   logout <provider>                               Delete a provider's stored login
   status [provider]                               Show login presence/expiry (never the token)
   refresh <provider> [--watch]                    Force a token refresh (--watch keeps it fresh)
-  reset                                           Reset the OAuth token store (clears corrupted entries and tokens)
+  reset --confirm [--json]                        Clear all OAuth logins and shared MCP tokens (also repairs corrupted stores)
   openrouter                                      Log in to OpenRouter in the browser; mints an API key
   chatgpt                                         Log in to ChatGPT in the browser (Codex backend, ChatGPT Plus/Pro)
 
--- a/internal/cli/auth_test.go
+++ b/internal/cli/auth_test.go
@@ -196,10 +196,39 @@ func TestRunAuthHelp(t *testing.T) {
 	if code := runWithDeps([]string{"auth", "--help"}, &stdout, &stderr, appDeps{}); code != exitSuccess {
 		t.Fatalf("exit = %d", code)
 	}
-	for _, want := range []string{"zero auth", "login", "logout", "status", "refresh", "--device"} {
+	for _, want := range []string{"zero auth", "login", "logout", "status", "refresh", "reset", "--device"} {
 		if !strings.Contains(stdout.String(), want) {
 			t.Fatalf("help missing %q:\n%s", want, stdout.String())
 		}
+	}
+}
+
+func TestRunAuthReset(t *testing.T) {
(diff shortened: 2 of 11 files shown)


Vasanthdev2004 (changes requested):
The chunking design is sound and the commit-at-the-manifest contract is the right one. One thing has to change first: the cleanup marker can delete the live generation, and the stored logins go with it.

## The stale cleanup marker eats the live generation

`writeChunked` opens with `sweepCleanupAccount()`, before it picks a family or writes anything. The sweep deletes chunks `0..N` of whatever family the marker names, and never asks whether that family is the one the manifest currently points at.

It becomes the live family easily. The marker is written only during the first migration (`previous.live == ""`), which always targets family `a`, and only when the orphan cleanup could not remove what it had already written. So the marker says `a:N` and stays. A later retry of that same migration succeeds and commits `a` as live. Now the marker names the live generation, and the next write deletes it.

That alone is survivable, because the write that follows the sweep rewrites the data into the other family. It stops being survivable when that write fails, which is exactly the case the manifest commit point exists to protect:

```
step1: Save err = keychain is full
       cleanup orphaned migration chunks: remove oauth-tokens.a.0: keychain is busy
step1: cleanup marker present=true value="a:1"
step2: committed manifest live="a" counts=map[a:2 b:0]
step2: marker still present=true value="a:1"      <- names the LIVE family
step3: fourth Save err = keychain is full
step3: Load(third) found=false
       err=oauth: keyring token data ... is missing chunk 1 of 2;
           run `zero auth reset` or remove entries ...
```

Driven through `Store.Save` and `Store.Load` with this package's own `fakeKR` at the real macOS budget, faults narrowed to `oauth-tokens.a.1` on set and `oauth-tokens.a.0` on delete. Every user with a stored login on that machine is logged out and has to re-auth.

The write-fails-alone case is fine, and your `TestStoreKeyringWriteCommitsOnlyAtTheManifest` proves it. The sweep is what breaks it, by removing the thing that test relies on being there.

Two ways out that both look fine to me. Skip the sweep when the marker names the live family, since those chunks are not orphans. Or clear the marker at the manifest commit, since committing a family is the moment its chunks stop being orphaned. I would take the second: it removes the stale state rather than working around it.

Worth a regression that runs the sweep and then fails the replacement write, because a test that lets the replacement succeed passes either way. Mine did.

The `read()` retry loop has no test. Reducing it to a single attempt leaves the whole package green, so nothing pins the behaviour it was added for.

`withLock`'s new doc comment describes a reader lock the code does not take and cites a constant that is not there. The code is fine, the comment is not, and a comment that promises a concurrency guarantee is the kind that gets believed later.

Store construction now hard-fails when the home directory cannot be resolved, where base degraded to in-memory. Probably deliberate, but it is a behaviour change that is not in the body.

The `keyringBlob` doc says the macOS anchor budget is 4039 bytes; the code computes 4037.

## What is good

Sizing every chunk against the longest account name the family can produce, rather than against chunk 0, is the detail I would have expected to be wrong. The comment explaining why is right: the account shares the command line with the secret, so a budget derived from a one-digit index overflows the moment the count reaches ten.

The A/B family alternation with the manifest as the single commit point is the correct shape for this, and the error message on a missing chunk tells the user exactly which entries to remove. That is a better failure than most credential stores manage.

jatmn (changes requested):

## Merge readiness

- Vasanthdev2004 left `CHANGES_REQUESTED` on head `95732af` for the cleanup-marker issue below. That is the only blocking request I can verify on current head.
- Greptile's block comment targets a different failure mode (the sweep erasing its marker after a failed retry). On head, `sweepCleanupAccount` returns early when chunk deletion fails and **keeps** the marker (`store.go:880-881`). The defect is not marker loss — it is the marker outliving the migration and naming the live generation.
- Branch head matches live `main` (`1b5db17`); no rebase needed. Required CI checks are green.
- This PR supersedes closed PR #938. The chunking design, manifest commit point, and `auth reset` surface are in good shape; the remaining blocker is a lifecycle gap in the new `.cleanup` recovery path, not the overall architecture.

## Findings

- [P1] Tie `.cleanup` marker lifecycle to manifest ownership — never sweep the live generation
  `internal/oauth/store.go:721`, `:783-785`, `:806-810`, `:865-888`

  ### What goes wrong

  The PR adds `oauth-tokens.cleanup` as durable bookkeeping when first-migration rollback cannot delete orphaned chunks (`783-785`). `sweepCleanupAccount` (`865-888`) is called at the start of every `writeChunked`, `writeWhole`, and `reset` and deletes chunks `0..N-1` for the family named in the marker, with no check against `manifest.live`. A successful manifest commit (`806-810`) does not clear the marker.

  That produces a reachable corrupt state without any manual keychain tampering:

  1. **First migration fails with rollback delete failure.** `writeChunked` runs with `previous.live == ""`, writes chunk(s) to family `a`, hits an error before manifest commit, defer rollback delete fails → marker set to `a:N` (`783-785`).
  2. **Retry succeeds while opening sweep delete fails.** `sweepCleanupAccount` runs first (`721`); `deleteChunkRange` fails (`880-881`) so the marker is retained, but the function returns void and the write continues. Migration completes; manifest commits with `live=a` (`806-810`). Marker still says `a:N`.
  3. **Next write deletes live chunks, then fails.** `sweepCleanupAccount` deletes `a[0:N)` — the generation the manifest points at. If the subsequent write fails (keychain busy, quota, transient `security` error), `Load` returns `missing chunk` and the user must `zero auth reset`.

  A write that fails *without* the sweep is survivable — `TestStoreKeyringWriteCommitsOnlyAtTheManifest` covers that. The sweep is what breaks the invariant that test relies on.

  Verified on head with two repros through `Store.Save` / `Store.Load` and the package's `cappedFakeKR` at the macOS budget: injected marker after live commit, and a fully natural path (failed rollback → marker → successful migration with failed opening sweep → failed next write → `Load` fails with `missing chunk`).

  ### Root cause (not a one-line guard)

  The `.cleanup` marker and the manifest describe **two independent sources of truth** for which chunks are safe to delete:

  - The **manifest** is the commit point: once `live` is set, chunks in that family are authoritative credential data.
  - The **marker** records "orphans from an interrupted first migration still need deletion."

  The bug is that `sweepCleanupAccount` treats the marker as unconditionally authoritative even after the manifest has taken ownership of that family. The marker was designed as a **recovery aid** for a pre-commit failure, but it is consumed **before** the write checks whether the named family is now live. There is no lifecycle edge that retires the marker when migration succeeds.

  `TestStoreKeyringFirstMigrationRollbackFailurePreservesReclaimableCleanup` only proves recovery via a subsequent **small** `writeWhole` save, which sweeps the marker and fits in the anchor. It does not cover: marker persists through a **successful chunked** commit (because opening sweep delete failed), then a later chunked write runs the sweep against the live generation.

  ### Recommended fix (pick one; both are sufficient)

  **Preferred — clear marker at manifest commit (Vasanthdev's second option):** When `writeChunked` successfully publishes the manifest (`806-810`), delete `oauth-tokens.cleanup` in the same logical transaction. Committing a family is the moment its chunks stop being orphans; the marker's job is done. This removes stale state instead of working around it.

  **Alternative — guard the sweep:** In `sweepCleanupAccount`, read the manifest first. If the marker names `manifest.live`, skip deletion (those chunks are not orphans). This is safe but leaves stale marker entries around until some other path clears them.

  Either approach restores the invariant: **no chunk deletion driven by `.cleanup` may target the live generation.**

  ### Regression test to add (this is the gap in current coverage)

  Add one test that exercises the full lifecycle, not just marker injection:

  - Trigger first-migration rollback failure so marker `a:N` is set.
  - Complete a successful chunked migration to `live=a` while the **opening** `sweepCleanupAccount` delete fails (marker must survive).
  - Attempt a subsequent chunked write that fails **after** the sweep runs.
  - Assert `Load` of an existing token still succeeds (or returns a recoverable state), not `missing chunk`.

  A test where the post-sweep write **succeeds** will pass either way and will not catch this.

  ### Explicit non-goals for this fix (avoid drift)

  Please do **not** expand this PR to address the following — they were reviewed and are either intentional, pre-existing, documented tradeoffs, or out of scope for #937:

  - Lock path relocation to `~/.zero/oauth-keyring.lockfile` (intentional fix for env-root split-brain).
  - Cross-process lock hold time during first-migration sweeps (pre-existing 5s timeout; no verified user-visible failure on head).
  - `Load` / `Status` not taking the file lock (pre-existing on base; inaccurate comment only).
  - Shrink-to-whole residue when retired-generation delete fails (documented accepted tradeoff at `701-707`).
  - MCP legacy `mcp-oauth-tokens.json` not cleared by `auth reset` (separate store surface).
  - Unguarded `auth reset` CLI semantics (deliberate recovery command).
  - Doc nits (`4039` vs `4037`, `withLock` comment wording, `read()` retry loop coverage).

  Fixing P1 and adding the regression test above should be sufficient to merge. I do not expect another review round for items outside this lifecycle gap.

## Why this PR has seen repeated review churn

This is meant as overall guidance, not additional blocking work.

**The change is inherently stateful.** Chunking introduces a generational manifest, alternating families, a commit point, rollback defer, and a new side-channel marker. Each piece is reasonable on its own, but the `.cleanup` marker crosses a correctness boundary: it was added for hygiene after a pre-commit failure, yet it runs before every write with no reconciliation against `manifest.live`. That kind of lifecycle bug is easy to miss when tests prove adjacent paths (rollback preserves marker, small `writeWhole` recovery works, manifest commit protects against write-failure-alone) but not the combination.

**Prior review noise pointed at the wrong sub-problem.** Greptile flagged marker *erasure* on failed sweep retry; on head the marker is correctly *retained* when delete fails (`880-881`). That sent fixes and discussion toward marker durability instead of marker **ownership**. Vasanthdev's review identified the actual defect: the marker can name the live generation after a successful commit. Converging on that single root cause should end the back-and-forth.

**PR #938 set expectations for a large surface area.** This revision adds ~1,800 lines including a 1,000-line test file. Broad coverage creates confidence in the happy path and many failure modes, but the specific sequence "opening sweep fails → chunked commit succeeds → marker stale → next sweep hits live generation" falls between existing tests. One targeted regression closes that hole.

**What "done" looks like for merge:** One lifecycle fix (clear marker at commit or guard live family), one regression test for the sequence above, and no scope expansion into lock redesign, reader locking, MCP reset, or CLI guard rails. The chunking architecture, sizing against longest account name, A/B alternation, and `auth reset` wiring are sound and should stay as-is.

euxaristia:
@coderabbitai full review

jatmn (dismissed):
@Vasanthdev2004 off to you

jatmn (dismissed):

Vasanthdev2004 (dismissed):
Fixed. Same sequence I used before, driven through `Store.Save` and `Store.Load`:

```
step1  marker written        "a:1"
step2  live="a"  marker still "a:1"      <- the stale marker still names the live family
step3  replacement write fails after the sweep
       Load(third)  survived
       Load(first)  found
       Load(second) found
```

Every stored login survives where all three were destroyed before.

Consulting the manifest and skipping the sweep when the marker names the live family is the right shape, and better than clearing the marker at commit, which was my other suggestion: this way a marker that outlives a failed cleanup is harmless rather than depending on a delete that already proved it can fail. Returning without touching anything when `readManifest` errors is the safe direction too.

One small thing, not blocking. If the marker says `a:N` and the live generation is `a` with a count below `N`, the chunks between the two counts are genuine orphans and the early return now leaves them. That is wasted keychain entries rather than a correctness problem, and it is strictly better than the alternative, but a comment noting it would save the next reader working it out.

`internal/oauth` green.

Vasanthdev2004 (approved):
Re-approving at 3c1f6fd8. My approval was dismissed by the new push, so this re-verifies both halves.

The credential-loss fix still holds. Same sequence through `Store.Save` and `Store.Load`, with the stale marker still naming the live family and the replacement write failing after the sweep:

```
live="a"  marker=true "a:1"
Load(first)  found=true
Load(second) found=true
Load(third)  found=true
```

All three survive, where all three were destroyed before.

On the new commit: requiring `--confirm` for `zero auth reset` is a breaking change to a command people may have in scripts, and I think it is the right one. It clears every stored login, the old help text described it as repairing corrupted entries, which undersold what it does, and the new message says plainly what will be lost and how to proceed. Help, flag validation and shell completions all moved together, and `TestRunAuthResetRequiresConfirmation` pins it.

Worth a line in the release notes, since a script calling `zero auth reset` now exits with a usage error rather than doing nothing visible.

`internal/oauth` green. `internal/cli` fails only `TestBuildServeScopeKeepsLexicalPaths`, which needs the symlink privilege and fails the same way on `main` here.

jatmn (approved):
lgtm

kevincodex1 (approved):

PR #1008, opened 2026-09-04
feat(tui): add session-scoped plan mode and durable plan editing

euxaristia:
## Summary

Add a session-scoped `/plan` workflow to the TUI, with durable per-workspace/session storage outside the workspace and external editing through `$VISUAL` or `$EDITOR`. A plan update is accepted only after its durable write succeeds, so failed saves cannot leave the tool and panel ahead of status and reloads.

Refs #854. This PR continues the plan-mode work from the closed PR.

## Changes

- Add `/plan on`, `/plan status`, `/plan open`, and `/plan off`, permission-mode restoration, and lifetime notices. Switching sessions exits Plan mode; returning from BTW restores the parent's mode.
- Restrict tools and permission requests during Plan mode, pause automatic loop and goal continuations, and propagate accepted plan snapshots to the TUI.
- Keep candidate plan updates private to each run. Persist before reporting tool success, then publish the accepted snapshot to the visible tool and panel. On failure, return a tool error and recover the latest durable value when readable; otherwise retain the last accepted session snapshot and display the storage error alongside it.
- Keep hidden-parent publication separate from the visible BTW tool. Ignore cancelled parent snapshots, preserve empty clears, and carry accepted state through completion when no live message sink is configured.
- Store plans under protected directory handles with Unix no-follow and Windows reparse-point checks. Exclude all sandbox-writable temporary roots from trusted storage and editor staging.
- Serialize baseline comparison and atomic replacement across processes, require editor baseline metadata, preserve rejected saved edits for recovery, and honor cancellation during writer-lock acquisition.
- Version newly written plan files with `<!-- zero-plan-format: 2 -->` so leading backslashes remain literal. Retain the legacy decoder and convert older plans with a baseline check before staging; an unchanged editor still produces no authored edit. Document the marker and continuation quoting in the README.

## Test plan

Validated with Go 1.26.6:

- `make fmt-check`, `go vet ./...`, and full Linux `go test ./...`.
- `go run ./cmd/zero-release build` and `go run ./cmd/zero-release smoke`.
- `make lint-static` with pinned golangci-lint v2.12.2; `make vulncheck` with pinned govulncheck v1.3.0.
- Full Linux race tests for `internal/tui` and `internal/planmode`, Windows plan/editor/BTW/storage tests, and macOS arm64 TUI test cross-compilation.
- `git diff HEAD --check`.

Windows temporary-directory cleanup failures were intermittent; the affected tests passed on focused reruns. The macOS check above is cross-compilation, not native execution.

On `d5411482`, Linux and macOS CI passed. [Windows CI](https://github.com/Gitlawb/zero/actions/runs/34083423326/job/101622870594) failed in `TestExecCommandForegroundServerReturnsSessionAndServesHTTP`: `server output did not include listening address`. The test is unchanged from `main` and expects the address in the initial 500 ms output window. A maintainer rerun is needed; GitHub rejected the rerun request because it requires repository admin rights.

Regressions were run against the original PR head and the fixed code:

- `TestPlanPublicationFailureKeepsConsumersAligned` forces conflicts, lock failures, startup read errors, and persistent unreadability through the actual TUI agent/tool callback path. It checks the tool, panel, status, durable file, model error result, and BTW return; success, empty clears, and completion without a live sink are also covered. The original code failed with `tool accepted rejected publication` and `model error result = false, want true for durable publication`.
- `TestCancelledParentPlanMessageCannotReplaceBTWSnapshot` failed on the original code with `cancelled parent replaced accepted BTW snapshot`.
- `TestPlanEditorPreservesLiteralBackslashes` failed on the original code with `editor reload changed literal backslashes`, losing one backslash from a two-backslash path. Fixed tests also cover legacy decoding, CRLF, continuation quoting, and conversion without a false edit event.
- The earlier `TestPlanPublicationCancelsWhileLockIsHeld` regression returned `lockutil: lock is held` on the original implementation instead of `context deadline exceeded`; the fixed code preserves the durable plan while honoring cancellation.

- The publication integration regression also checks read-versus-write error labels across live, BTW, and completion delivery. On the previous head it failed with `publication failure reported the wrong operation`; the fixed code names the operation that failed.

- Implement the persist-then-propagate strategy from the latest maintainer review, with one accepted plan shared by publication, status, tool state, and BTW restoration. Save failure is a failed tool operation rather than a success followed only by a transcript error.
- Preserve interprocess acceptance locking, required baseline sidecars, full sandbox temporary-root coverage, rejected-edit recovery, cancellation ordering, and editor no-op/authorship handling.
- Address the leading-backslash editor finding with an explicit format marker and legacy compatibility. Show the required pipe followed by one space in literal README examples.
- Distinguish startup baseline-read errors from attempted-write errors in the tool result.
- Preserve canonical status, Windows editor-path handling, agent-loop integration assertions, isolated test storage, and the confirmed session-scoped lifetime policy.
- Keep the existing temp-root regression fix: Unix tests use both `TMPDIR` and an actual directory beneath `/tmp`.

## Summary by CodeRabbit

* **New Features**
  * Added `/plan open` to edit plans externally and reload changes into the session.
  * Plan edits preserve rejected work for recovery and detect conflicting updates.
  * Plan files preserve literal text and support safer editing workflows.
  * Plan mode maintains plan state across side conversations with clearer status notices.

* **Bug Fixes**
  * Blocked executable tools, hooks, loops, goal continuations, and permission requests during plan mode while retaining read checks.
  * Improved protection against unsafe plan paths and concurrent edits.
  * Prevented cancelled or outdated runs from overwriting accepted changes.
  * Plan mode exits when switching sessions.


jatmn (changes requested):

The editable, resumable plan is a useful feature with a purpose distinct from the orchestration work in #829. The ten findings below concern concrete persistence, editor, platform, and test behavior. They include failures discussed on #854. The separate question about Plan-mode lifetime needs a maintainer decision; it should not become an implicit instruction to redesign session permissions.

## Why the fixes need to cover complete operations

The pattern across these findings is that a safeguard works at one step, while the next step relies on a stronger guarantee than that safeguard provides. For example:

- Atomic replacement prevents a partially written file, but does not make a baseline comparison and replacement atomic across processes.
- A process-local mutex serializes callers inside one instance, but does not order competing instances or establish whether a cancelled callback is older than an accepted editor change.
- A conflict error protects the durable file, but unconditional cleanup then destroys the user's rejected edit.
- Storage recognizes an unchanged editor file, but the TUI loses that result and infers a user edit from a lossy parse.
- A path is outside the selected OS temporary directory, but can remain inside another directory the sandbox already permits writes to.

These observations explain why repairing the cited line alone can leave a closely related failure reachable. They do not require a new storage system or a rewrite of the TUI. The useful unit of repair is the complete operation: identify the input and owner, validate the preconditions, accept or reject the change, publish the result to its consumers, and handle recovery and cleanup according to that result.

The findings fall into four groups:

| Area | Findings | Contract to establish |
|---|---|---|
| Durable plan writes and recovery | 1, 2, 4, 5 | A writer must have a valid basis for replacing the plan; competing writes must be ordered; an older cancelled publication must not replace newer accepted state; rejected user work must remain recoverable. |
| Editor completion and plan presentation | 6, 7 | Distinguish an unchanged operation from an accepted edit, and make displayed plan semantics agree with the accepted plan. Raw file formatting need not be identical to the display. |
| Existing platform boundaries | 3, 8 | Use the actual sandbox writable-root rules and preserve the pathname semantics of the supported operating system. |
| Regression protection | 9, 10 | Tests must own newly introduced persistence dependencies and retain the unrelated integration assertions already present on `main`. |

Please address each group's complete failure path and its existing callers together. The grouping is not a request to combine everything into one abstraction: interprocess atomicity, publication freshness, and recovery are distinct obligations even if they share some implementation.

## Findings

### 1. [P2] Make the baseline check atomic across Zero processes

`internal/planmode/planmode.go:382`

**Failure path and evidence.** Two Zero instances resume the same workspace/session and stage independent edits from durable plan A. Both enter `CommitStagedEdit`, read A, and pass the baseline comparison. Both then replace the durable file. The second replacement silently loses the first accepted edit. Four processes committing distinct edits from the same baseline all succeeded in a reproduction; competing edits should not all be accepted against that one baseline.

**Root cause.** `lockPlan` uses a process-local mutex. Atomic rename protects each individual replacement, while the staging locks belong to unique editor files. Neither makes the durable read/check/replace operation indivisible across instances. A competing direct `WritePlan` can also change the file between the editor's comparison and replacement. The resolved atomicity request is therefore not satisfied across processes.

**Required outcome.** Coordinate the complete baseline check and replacement with every writer of the same durable plan, including `WritePlan`. An interprocess lock or an equivalent conditional replacement mechanism can satisfy this; the implementation choice is open. Keep independent sessions independent, retain atomic replacement and the existing no-follow protections, and preserve the behavior where an unchanged stale editor does not overwrite newer durable content.

**Regression coverage.** Use separate processes, not only goroutines, and synchronize them so they compete from the same baseline. Cover editor versus editor and editor versus `WritePlan`. Verify that a stale edited snapshot cannot silently replace an already accepted competing value. Also cover an unchanged stale editor, which should preserve the newer durable value without introducing an unnecessary write.

### 2. [P2] Fail closed when the editor baseline is unavailable

`internal/planmode/planmode.go:386`

**Failure path and evidence.** If reading `.basehash` fails, `CommitStagedEdit` skips no-op and conflict detection and proceeds with an unconditional write. Both Unix and Windows staging ignore failures creating that sidecar, so staging can report success without establishing a baseline. With the sidecar unavailable, changing the durable plan and then committing the older edited snapshot replaces the newer value without a conflict.

**Root cause.** The baseline is required to authorize safe replacement, but is treated as optional metadata on the error path. The staging result does not guarantee that the later commit has the information it needs. This is separate from finding 1: even perfect interprocess locking cannot validate an absent baseline.

**Required outcome.** Report success from staging only after establishing the required baseline. If that baseline is unavailable or invalid at commit, report an error rather than silently authorizing replacement. Preserve the valid empty-plan and unchanged-editor cases. Handle sidecar creation/write failures explicitly on both supported platform implementations.

**Regression coverage.** Inject baseline creation and write failures and verify that staging does not report a usable successful operation. Exercise a missing, unreadable, or invalid baseline at commit after a competing durable change. Assert that the durable value is preserved and the error is surfaced. Pair commit-failure cases with the recovery assertions in finding 4.

### 3. [P1] Exclude every sandbox-writable temporary root

`internal/planmode/planmode.go:489`

**Failure path and evidence.** This requires a redirected configuration layout: for example, `XDG_CONFIG_HOME` beneath `/tmp` while `TMPDIR` points to a different directory. Both the durable-path check and editor staging accept that layout because they only exclude `effectiveTempDir()`. However, the sandbox's existing default writable roots include `/tmp` on Unix even when `TMPDIR` points elsewhere. On Windows, the equivalent mismatch is that the sandbox includes both `TEMP` and `TMP`, while the check uses the selected OS temporary directory.

**Impact.** The accepted durable plan and editor staging file can reside in a root writable by a same-user sandboxed process. Mode `0700` does not isolate them from a process running as that same user. The process can alter staged content or replace the pathname that the unsandboxed editor subsequently reopens. This defeats the reason the staging area is treated as trusted. The finding is conditional on the redirected layout; it is not a claim that ordinary default config locations are exposed.

**Root cause.** The plan package approximates the sandbox's trust boundary using a single temporary directory. The actual boundary is the complete set of default writable roots. Physical-path resolution and private permissions cannot compensate for using an incomplete root set.

**Required outcome.** Validate both durable storage and editor staging against that complete set, using the appropriate physical-path comparisons. Preserve legitimate redirected config ancestors outside those roots and the protected-root/descendant no-follow behavior. Reusing the root rules or otherwise ensuring parity is sufficient; changing the sandbox's existing writable-root policy is outside this fix.

**Regression coverage.** Cover Unix `/tmp` with a different `TMPDIR`, Windows with distinct `TEMP` and `TMP`, and physical aliases into those roots. Assert rejection for both durable writes and staging. Include a legitimate redirected config location outside writable roots to ensure the repair does not reintroduce rejection of supported config layouts.

### 4. [P2] Preserve edited content when write-back fails

`internal/tui/plan_command.go:262`

**Failure path and evidence.** The user opens a plan, makes changes, saves, and exits the editor successfully. Another instance has modified the durable plan in the meantime. `CommitStagedEdit` returns a conflict, but the callback's unconditional `defer cleanup()` deletes the staged file containing the user's saved work. A durable write failure reaches the same cleanup path. The error is visible, but the edited copy is gone.

**Root cause.** Cleanup treats successful completion and failed write-back as equivalent. Conflict detection protects the durable value but does not complete the user-facing recovery contract. The existing storage conflict test can pass without detecting that the TUI callback destroys the rejected work afterward.

**Required outcome.** Make saved edits recoverable when write-back fails and report where the user can retrieve them. Release the staging liveness lock appropriately, and ensure the recovery copy is not immediately removed by the normal cleanup path. The location and retention mechanism are implementation choices. This does not require indefinite retention of every temporary file, and successful commits should retain normal cleanup behavior.

**Regression coverage.** Exercise the editor completion callback, not only `CommitStagedEdit`. Cover conflict and ordinary write failure after a saved edit. Assert that the newer durable value remains intact where applicable, the saved edit remains accessible, and the error identifies its recovery location. Verify that locks are released and successful commits do not accumulate staging files. Check how a later staging sweep treats any recovery location chosen by the fix.

### 5. [P2] Prevent cancelled runs from overwriting a newer accepted plan

`internal/tui/model.go:5894`

**Failure path and evidence.** A successful `update_plan` callback is delayed before its durable write. Cancellation clears `pending` and `activeRunID` while the old run remains in `flushRunIDs`. `/plan open` checks `pending` and `exiting`, so a newer editor commit can complete before the old callback resumes. The callback then replaces that accepted edit with its old snapshot. A controlled scheduling reproduction through the actual `runAgent`/`OnToolResult` path produced exactly this overwrite.

**Impact.** The UI rejects a stale `planUpdateMsg` by run ID, but that rejection does not protect the subsequent disk write. The visible plan and durable plan can therefore diverge, and the user's newer work can be lost from storage.

**Root cause.** Capturing the correct session ID and immutable result snapshot solves cross-session provenance, but does not establish whether that result is fresh enough to publish now. The durable writer is outside the UI's stale-run filter. A per-plan mutex alone is insufficient: it can serialize an obsolete callback after a newer accepted edit and faithfully perform the wrong replacement.

**Required outcome.** Prevent obsolete cancelled-run publication from superseding newer accepted state for the same session. For example, ordering new plan acceptance after outstanding publication drains, or validating publication freshness at the acceptance boundary, can satisfy the requirement. Preserve legitimate cancelled-run session/checkpoint flushing, captured session identity, and successful empty-plan snapshots. A standalone cancellation check that leaves another check/write race is not a complete ordering guarantee.

**Regression coverage.** Hold a successful result callback before publication, cancel the run, and exercise the chosen behavior for a subsequent editor operation or same-session update. Release the old callback and verify it cannot replace newer accepted content. Assert durable state as well as UI state. Preserve separate tests for legitimate late event flushing and for a result belonging to a different captured session.

### 6. [P2] Make plan status agree with the accepted plan

`internal/tui/plan_command.go:359`

**Failure path and evidence.** Save an edit containing two `[in_progress]` rows. Reload canonicalizes the accepted plan so that the first row becomes completed in the tool, panel, and editor-authored context. The durable file retains both active rows, and `/plan status` displays those raw bytes. The user therefore sees different step status depending on which surface they inspect.

**Root cause.** Reload consumers use canonical plan semantics, while the status renderer treats source text as the displayed semantic result. The tool/panel/context disagreement described in the earlier #854 review has been corrected; the narrower status disagreement is what this finding addresses.

**Required outcome.** Have `/plan status` report the same accepted canonical state as the tool and panel. Fixing the display boundary is sufficient. Preserving raw user-authored file formatting is compatible with this request; byte-for-byte equality between source text and the displayed plan is not required. This finding does not ask for a file watcher or a new persistence format.

**Regression coverage.** Accept an editor file with multiple active rows and supported status aliases. Compare the semantic content/status/notes shown by `/plan status` with the tool, panel, and editor-authored context after completion and reload. If the implementation deliberately retains raw file text, test that policy separately rather than requiring a byte-identical canonical rewrite.

### 7. [P2] Carry editor no-op detection through to the TUI

`internal/tui/model.go:1487`

**Failure path and evidence.** An agent plan item contains `first\r\nsecond`. Open the editor and exit without changing a byte. `CommitStagedEdit` recognizes that the staged bytes are unchanged, but returns no operation result distinguishing that case. Reload normalizes the internal newline to `\n`; `planItemsEqual` only trims the ends. The TUI treats the representation difference as an edit, changes the in-memory plan, and appends the false user-authored “I edited the plan file directly” event to future model context.

**Root cause.** The storage layer has the evidence needed to recognize a no-op, but the completion message loses it. The UI then guesses whether the user edited the file by comparing values across a lossy formatting/parsing boundary. Structural inequality there is not proof of user action. The earlier no-op request on #854 is not satisfied for this case.

**Required outcome.** Preserve reliable edit/no-op information through completion and base the user-authored event on that outcome. When both the staged and durable content are unchanged, opening and closing the editor should not mutate the accepted plan or invent an edit event. Genuine edits and clears must continue to reach the next turn. If another writer changed the durable plan while this editor stayed unchanged, preserve the newer value and distinguish any hydration from an edit authored by this user.

**Regression coverage.** Drive staging, unchanged commit, and TUI completion together using CRLF and supported multiline/notes content. Assert zero false edit events and unchanged accepted state for a true no-op. Pair those cases with actual edits and an empty clear, asserting the appropriate single user event. Include unchanged editor bytes after a competing durable write so preserving the newer value does not become a false authorship claim.

### 8. [P2] Preserve quoted UNC editor paths on Windows

`internal/tui/plan_command.go:293`

**Failure path and evidence.** Set the editor to a conventional quoted Windows executable path such as `"\\server\shared tools\vim.exe" --wait`. Because the value begins with a quote, `splitEditorCommandFor` bypasses `windowsEditorFields` and invokes POSIX `shell.Fields`. That parser reduces the two leading backslashes to one. The resulting executable is a local root-relative path instead of the UNC path, so `/plan open` cannot launch the intended editor.

**Root cause.** The parser selects Windows pathname handling based partly on whether the command is quoted. Quoting protects spaces but does not change the operating system's pathname semantics. Quoted drive-path tests do not exercise the significant doubled prefix in UNC paths.

**Required outcome.** Preserve Windows executable path semantics for quoted and unquoted forms, including spaces in the executable path and the existing argument behavior. Keep Unix parsing unchanged. Use the parser approach that meets those cases; this does not call for general shell-command execution or expanded shell features.

**Regression coverage.** Assert the exact parsed executable and arguments for quoted UNC paths, quoted drive paths with spaces, unquoted Windows paths, and the supported extended path form. Retain the existing Unix quoting cases. Checking that parsing merely succeeds is insufficient; the path prefix must survive unchanged.

### 9. [P2] Restore the removed agent-loop integration assertions

`internal/agent/loop_test.go:3975`

**Evidence and impact.** Relative to current `main`, the branch removes the assertion that the aborted `flaky-2` placeholder carries `IsError`. It also removes the `OnContext` forwarding and matching trace `PrefixHashes` assertions from `TestRunTracingWrapperStampsUsage`. Equivalent integration checks are absent at this head. These removals reduce protection against future protocol and trace regressions; they do not prove those production behaviors are currently broken.

**Root cause.** A shared test-file change associated with plan work has dropped unrelated assertions that the mainline integration suite already provided. Moving other tests successfully does not replace these particular checks.

**Required outcome.** Restore the two protections at the same integration boundary, either in the existing tests or equivalent tests. Preserve the structured tool-error and cancellation-retry tests that were relocated successfully. This request does not include unrelated production changes, a broad test rewrite, or a claim that all tests mentioned in an older review are missing.

**Regression coverage.** The aborted-placeholder test should assert both recognizable aborted content and structured error status. The tracing test should verify callback forwarding and agreement between the callback's context evidence and the recorded prefix evidence. Lower-level unit assertions alone do not establish that the `Run` integration forwards and records the values correctly.

### 10. [P2] Give legacy plan-entry tests an owned session store

`internal/tui/plan_mode_test.go:21`

**Failure path and evidence.** `/plan on` now calls `ensureActiveSession`, but the original plan-entry tests construct `newModel` with its default session store. They consequently create persistent sessions under the developer's actual data directory. Making that data root unavailable causes `TestPlanCommandEntersAndExitsPlanMode`, `TestPlanCommandRestoresPriorModeOnExit`, and `TestPlanCommandOnTwiceDoesNotClobberSavedMode` to fail before entering Plan.

**Root cause.** A previously in-memory command acquired a persistence dependency, but its older callers in the test suite were not updated to supply that dependency. Config/staging isolation alone does not isolate the session store. This is a test-isolation failure activated by the new entry path, not a request to remove production session creation.

**Required outcome.** Inject test-owned session storage into every affected model and isolate any remaining config/data roots using the platform's actual resolution rules. Preserve production session creation on `/plan on`. Check the legacy entry tests as well as the newer editor tests that already use dedicated helpers.

**Regression coverage.** Run the affected entry tests with explicit test stores and unavailable ambient data storage in an isolated test environment. Verify that the commands enter and exit the requested mode and that session writes are confined to the injected roots. Exercise the supported platform root-resolution paths without writing into or changing permissions on a developer's real profile.

## Suggested repair sequence and acceptance checks

1. **Preserve the existing regression baseline and isolate test storage.** Address findings 9 and 10 so subsequent validation does not drop unrelated protections or create real user sessions. These are narrow test changes.
2. **Close the trusted-path mismatch.** Align the plan storage/staging checks with existing sandbox roots, retaining the supported redirected-config and no-follow behavior. Keep Windows editor argument parsing as a bounded compatibility fix.
3. **Define and implement durable write outcomes across all current writers.** Address interprocess comparison/replacement, required baseline handling, and cancelled publication ordering together. Then carry rejected write-back into recoverable editor completion. Verify findings 1, 2, 4, and 5 as distinct cases; solving one does not establish the others.
4. **Carry editor outcomes into presentation and context.** Make no-op/edit/conflict/error distinguishable wherever the TUI needs that distinction, and make status use the accepted semantic plan. A small return value or existing-state adjustment may be sufficient; no particular new framework is required.
5. **Run the complete affected operation cases before requesting another review.** Cover normal edit, unchanged editor, empty clear, competing writer, absent baseline, failed commit, cancellation with a delayed callback, and reload. For each case assert the relevant durable value, active plan/display, user-authored event, and recovery/cleanup result. Use actual separate processes for cross-process claims and the TUI callback for user-visible completion claims.

The acceptance target is that the listed failure paths are prevented without losing behavior already established by the feature: session-specific plan identity, immutable successful snapshots including empty clears, legitimate cancelled-run event flushing, supported editor syntax, valid config redirection, and cleanup of successful or genuinely abandoned staging operations. Test the applicable combinations at their real boundaries rather than adding one isolated assertion per error message.

A focused repair explanation would help close this as one coherent change: for each finding, identify the accepting/rejecting boundary, the other writers or consumers it covers, and the regression that demonstrates the full outcome. Please keep unrelated cleanup and speculative architectural changes out of that repair. The root-cause groupings above are intended to reduce repeated partial fixes while leaving implementation choices open.

## Needs maintainer decision

### Clarify whether Plan mode ends when switching sessions

`internal/tui/session.go:77`

The implementation and transition tests deliberately treat Plan as session-scoped: `/new`, cross-session `/resume`, and `/btw` restore the prior permission mode for the destination; returning from BTW preserves the parent's mode. Same-session resume preserves Plan. `/spec` also exits Plan, but its implementation run follows explicit spec approval.

That behavior conflicts with the entry text saying write tools stay hidden until `/plan off` and the earlier request on #854 to preserve Plan until an explicit exit. The change in behavior is established, but the intended policy is not settled by those conflicting signals. It is therefore separate from the ten code/test findings above.

If Plan is session-scoped, make that lifetime clear in the entry/transition notices and preserve the destination and parent-session semantics consistently. If Plan is intended to remain in force until an explicit exit, agree on which transitions preserve it or require an explicit exit/confirmation before implementing that policy. Requiring inheritance across every new conversation would be product drift if session-scoped behavior is intended.

Once the policy is settled, exercise entry, same-session resume, cross-session resume, new session, BTW entry/return, spec approval, and explicit `/plan off` from the relevant prior modes. Assert the actual run permission mode as well as the user-visible notice. Resolve this decision explicitly; do not infer it from the persistence fixes or turn it into an unrequested permission-system redesign.

jatmn (changes requested):

The durable-plan work from the earlier review round is largely landed on this head. Interprocess locking, baseline sidecars, sandbox temp-root parity, editor recovery on failed commit, cancelled-callback ordering, canonical `/plan status`, and the targeted regression tests all check out. I am **not** re-opening those threads.

What remains is **one architectural gap**, not a fresh pile of unrelated defects: several surfaces still treat “plan updated” and “plan durably published” as the same event. They are not when `WritePlanIfUnchanged` fails. Fixing that single boundary should close this review without another round of point patches.

---

## Why this PR has seen so many review cycles (and how to stop dripping feedback)

This branch has ~60 commits and multiple review passes because the work keeps adding **correct local safeguards** while the next layer still assumes a **stronger guarantee** than the prior layer actually provides. That pattern is visible across the resolved items too:

- Atomic rename protects one file replacement, but does not by itself make “read baseline → compare → replace” indivisible across processes — hence the interprocess lock work.
- A process-local mutex serializes one instance, but staging locks belong to unique editor files — hence separate acceptance-lock tests.
- Conflict detection protects the durable file, but cleanup used to destroy rejected edits — hence `PreserveStagedEdit`.
- Storage can recognize an unchanged editor file, but the TUI used to infer authorship from a lossy parse — hence `EditResult`.
- A path can be outside `TMPDIR` but still inside another sandbox-writable root — hence `DefaultTempWriteRoots`.

Each round fixed a **real** hole. The remaining feedback is not “you missed twelve new features.” It is the same root pattern one level up: **publication is split across consumers that do not share one outcome.**

### The root cause (please fix this, not another symptom)

Today, a successful `update_plan` tool call produces **three parallel truths**:

| Surface | Updated when | On failed durable write |
|---|---|---|
| Shared `update_plan` tool | During `Run()` on the agent goroutine | **New plan** (already committed in tool) |
| Sticky panel | Immediately via `planUpdateMsg` in `OnToolResult` | **New plan** (optimistic) |
| Durable file under config | Only if `WritePlanIfUnchanged` succeeds | **Old plan** (unchanged) |
| `/plan status` (`planText`) | Reads durable file when `exists=true` | **Old plan** (by design of `planText`) |
| Resume / `/plan on` reload / BTW return success reload | Reads durable file | **Old plan** |

The code comments say the durable file is the single source of truth for `/plan` reads and restart/resume. That contract is **true on the happy path** and **false on the failure path** above. A transcript error row tells the operator persistence failed, but nothing reconciles the other rows in the table — so the session can keep planning against a newer in-memory plan while every reload path silently reverts to disk.

**This is not a request for a new storage system, an ACP rewrite, hook policy changes, exec/ACP parity, or broader plan-mode redesign.** Those would be scope drift and are explicitly out of this review.

**This is a request for one publish boundary** in the TUI (and its BTW return path) with a documented contract:

1. **Accept** — durable write succeeded → update panel, tool (if needed), status, BTW reload targets together.
2. **Reject** — durable write failed → either roll back optimistic UI state to the last accepted durable snapshot, **or** mark the session as “durable lagging” and make status/reload/BTW follow the same accepted in-memory snapshot until a retry succeeds. Pick one semantics; do not leave mixed readers.

Fix the **complete operation** (validate → publish → propagate → recover), not another line in isolation.

### Anti-drift guardrails for this repair

Please **do not** expand this PR with:

- ACP `emitPlan` migration (`internal/acp/agent.go` is unchanged in this diff; pre-existing race class; follow-up issue is fine).
- exec/ACP durable persistence (pre-existing; TUI-scoped PR).
- `beforeTool` hook suppression changes (intentional fail-closed veto tradeoff; tested).
- Silent-editor-reload UX changes (`TestEditorCompletionCarriesAuthorshipAndCanonicalStatus` expects no authorship event for external reload).
- Plan-mode lifetime redesign (see maintainer decision below — messaging/policy only unless product says otherwise).

Please **do** keep: interprocess lock, baseline sidecar, staging containment, editor recovery, `WritePlanIfUnchanged` concurrency, and existing happy-path behavior.

---

## Merge readiness

- Review gate: `reviewDecision` is `CHANGES_REQUESTED` (maintainer + CodeRabbit). Merge state is `blocked` even though the branch is mergeable and CI is green.
- Branch freshness: head is current with `main` (0 behind / 0 ahead); no rebase blocker.
- Supersedes closed #854; this is the active plan-mode line.

---

## Findings

### 1. [P2] Establish one publish outcome for plan updates when durable persistence fails

`internal/tui/model.go:5888`, `internal/tui/plan_command.go:482`, `internal/tui/btw.go:230`

#### What happens today

On every successful `update_plan` in `OnToolResult`:

1. `planSnapshotFromResult` extracts the immutable snapshot from the completed call.
2. `runtimeMessageSink(planUpdateMsg{...})` updates the **sticky panel immediately**.
3. `planmode.WritePlanIfUnchanged(runCtx, ..., planBaseline)` attempts the **durable write**.
4. On error, only a transcript row is added: `"plan file write error: ..."`. Panel and tool state are left on the new snapshot.

`planText()` (used by `/plan` and `/plan status`) **always prefers the durable file when one exists**:

```482:507:internal/tui/plan_command.go
content, exists, readErr := planmode.ReadPlan(...)
...
if exists {
    content = formatPlanItems(tools.CanonicalizePlanItems(parsePlanFileLines(content)))
    return header + "\n" + strings.TrimRight(content, "\n")
}
// only falls back to in-memory draft when no durable file yet
```

So after a failed write **when a durable file already exists**, operators can see:

- Sidebar / sticky panel → **new** plan (from step 2)
- `/plan status` → **old** plan (from disk)
- In-session agent execution → **new** plan (tool updated during `Run()`)
- Resume, `/plan on` reload, cross-session hydration → **old** plan (disk)

The failure is **not silent** (transcript error exists), but the **consumers disagree** about which plan is authoritative. The data-loss class is resume/reload/BTW return, not “the agent keeps working in-session.”

#### Failure paths that trigger this (all already handled at the storage layer; the gap is consumer alignment)

- **Baseline conflict** — `"plan changed since this run's last accepted update"` (concurrent editor, second instance, hidden parent run).
- **Lock timeout** — `lockPlan` retry exhausted (~5s).
- **Run-start read poison** — `planBaselineErr` from initial `ReadPlan` at `runAgentWithOptions` start blocks every `WritePlanIfUnchanged` for the run while panel/tool still advance (`model.go:5473`, `5902-5905`).
- **Cancelled run context** — write rejected after cancel (panel may already have updated).
- **I/O errors** — permissions, symlink refusal, etc.

#### BTW boundary (same root cause, not a separate architecture bug)

While BTW is open, hidden parent runs route `planUpdateMsg` into `m.btw.parentPlanItems` and update the parent panel via `routeBTWMessageToParent`. If the parent’s durable write failed, the parent tool may still hold the newer plan from `Run()`.

On `leaveBTW`:

- **Error / missing file paths** — correctly restore `savedParentPlan` into tool + panel (`btw.go:230-260`).
- **Success reload path** — `reloadPlanFromFile()` overwrites panel **and** tool from disk **without comparing** to `parentPlanItems` or the fresher in-memory parent state (`btw.go:246-247`).

So BTW return can **regress** the parent to stale disk precisely when the error path would have preserved the snapshot. That asymmetry exists because reload success assumes disk is authoritative — which is only true if step 3 above succeeded.

#### Required outcome (pick one coherent strategy; do not patch one callsite)

Implement a single **plan publish** step used by `OnToolResult` and consulted by BTW return. Smallest acceptable shapes:

**Option A — Persist-then-propagate (recommended; matches “durable is source of truth”)**

1. Call `WritePlanIfUnchanged` **first** with the snapshot.
2. Only on success: emit `planUpdateMsg`, advance `planBaseline`, and treat disk as authoritative for `planText` / reload / BTW.
3. On failure: do **not** emit `planUpdateMsg`; leave panel aligned with last accepted durable state (tool state needs an explicit decision: either rollback to match durable, or document that tool may lead panel until next successful publish — but panel must not race ahead of durable if status reads disk).

**Option B — Propagate-with-rollback**

1. Keep optimistic panel update if you want responsiveness.
2. On write failure: roll back panel (and tool if it was optimistically advanced beyond durable) to the pre-call or last-accepted snapshot; keep the transcript error.
3. Ensure `planText` and reload paths read the same rolled-back state.

**Option C — Explicit “durable lagging” mode**

1. On write failure: set a session-scoped flag / generation counter.
2. While lagging: `planText`, BTW reload, and resume hydration follow the **accepted in-memory snapshot** (or last successful durable baseline), not raw disk.
3. Clear lagging on next successful publish.

Any option is fine. **Inacceptable:** leaving panel on the new snapshot while `planText`/reload/BTW success path read stale disk.

#### BTW-specific guidance (within the same strategy)

Make `leaveBTW` success reload **consistent with the publish outcome**:

- If parent durable is authoritative → reload is correct only when last parent publish succeeded.
- If parent publish failed during BTW → success reload must not clobber fresher `parentPlanItems`; mirror the error-path restore logic or skip reload when durable is known stale.

Do not add a second special case; reuse the same “last accepted plan” notion from the publish step.

#### Regression coverage (required so this does not drip back)

Add **one integration test** (TUI callback path, not only `planmode` unit tests) that:

1. Seeds a durable plan file.
2. Forces `WritePlanIfUnchanged` to fail (mock/spy, injected error, or baseline conflict).
3. Asserts **all** of:
   - panel state
   - `planText()` / `/plan status` output
   - shared tool state (per chosen strategy)
   - BTW return behavior if parent publish failed during isolation

   agree with each other and with the chosen publish contract.

Existing tests cover happy path, editor recovery, cancelled callback vs editor, and interprocess lock — they do not cover failed publish consumer alignment.

#### What “done” looks like

- One helper or clear sequence owns publish → propagate → recover.
- No duplicated “update panel then maybe write disk” logic scattered without shared outcome.
- BTW return uses the same accepted-plan notion.
- Failed publish cannot produce sidebar ≠ status ≠ resume.

---

### 2. [P3] Honor cancellation while waiting for the plan lock

`internal/planmode/planmode.go:100`

#### What happens today

`WritePlanIfUnchanged` accepts `runCtx`, but `lockPlan` retries `ErrLockHeld` with `time.Sleep(10ms)` for up to five seconds **without checking `ctx.Err()`**. Cancellation is only observed **after** lock acquisition (`writePlan` line ~157).

#### Impact

When lock contention coincides with run cancellation, the publication callback can block for seconds even though the write will ultimately fail. This is responsiveness/cleanup hygiene, not data corruption. It becomes more noticeable once publish-then-propagate (Finding 1) makes the callback more critical.

#### Required outcome

Pass `context.Context` into `lockPlan` (or check `ctx.Err()` in the retry loop) and return promptly on cancellation. Keep the existing 5s lock-acquisition deadline for non-cancelled callers.

#### Regression coverage

Small unit test: cancelled ctx + contended lock returns before deadline without writing.

---

## Needs maintainer decision

### Clarify Plan mode lifetime across session transitions

`internal/tui/session.go:77`, `internal/tui/plan_command.go:479`

**Observed behavior (consistent in code + tests):** Plan is session-scoped. `/new`, cross-session `/resume`, `/btw`, and `/spec` exit Plan and restore the destination/parent permission mode, with notices in several paths. Same-session resume preserves Plan.

**Conflicting signals:** Entry copy still reads as if write tools stay hidden until `/plan off`, and #854 discussion included “preserve until explicit exit.”

**This is policy/messaging, not a code defect in the persistence fix.** Please confirm session-scoped Plan is intended. If yes, tighten entry/transition notices only — do not bundle permission-model changes into the publish-outcome repair. If Plan should survive specific transitions, product needs to name which ones before implementation.

---

## Prior review items — verified addressed on this head

The ten items from the 2026-09-06 maintainer review appear resolved here:

1. Interprocess baseline check + replace (file lock + process tests)
2. Required editor baseline sidecar (staging fails closed; commit rejects missing/invalid baseline)
3. Full sandbox temp-root set (`DefaultTempWriteRoots`)
4. Editor recovery on failed write-back (`PreserveStagedEdit` + TUI callback tests)
5. Cancelled-run publication ordering (`WritePlanIfUnchanged` + delayed callback test)
6. Canonical `/plan status` (via `formatPlanItems(CanonicalizePlanItems(...))`)
7. Editor no-op detection (`EditResult`; no false authorship event)
8. Windows `$EDITOR` parsing (UNC/quoted paths)
9. Restored agent-loop integration assertions (`IsError`, trace prefix forwarding)
10. Test-owned session stores in legacy plan-entry tests

**Do not regress these while fixing Finding 1.**

---

## Suggested repair sequence (one coherent commit series)

1. **Introduce the publish outcome** (Finding 1) in `OnToolResult` — one helper, one contract, one integration test for failed publish.
2. **Align BTW return** with the same accepted-plan notion — no independent reload semantics.
3. **Thread ctx through `lockPlan`** (Finding 3) — small, isolated.
4. **Resolve Plan lifetime messaging** with maintainer input — copy-only unless told otherwise.

If you describe the chosen publish strategy (A/B/C above) and point to the integration test in the PR reply, that gives a clear merge gate and avoids another drip round of “what about this callsite?”

Vasanthdev2004 (changes requested):
I came to this as a first reviewer and found jatmn already there with the right finding, so I am not adding to the pile. Matching the state rather than contradicting it; the split publish outcome is the blocker and I agree with it. It is the "two representations of one fact with no contract between them" shape, and the table in that review is the clearest statement of it I have seen on this repo.

What follows is only to save you re-defending ground that is already solid. I went at the new package independently and these are clear:

**Plan path derivation.** No traversal. Nine hostile session IDs through `pathKey`, including `../../../../etc/passwd`, `..\..\..\windows\system32\x`, `C:/Windows/System32/evil`, an embedded NUL and a 500-character string: every one produces a single component, no separators, bounded length. The slug-plus-hash construction does what its comment claims.

**Editor staging.** The reasoning in `StageForEditor` is the most careful thing in this diff. Recognising that `XDG_CONFIG_HOME` can point the staging directory back into a default-writable sandbox root, and then closing it with `editorStagingDirIsPrivate` plus an unpredictable `O_EXCL` filename rather than one of those alone, is the right shape. `splitEditorCommand` handling quoted paths and refusing POSIX escapes on Windows paths is the detail I expected to be wrong.

**Write guards.** The only unguarded `WritePlan` is the seed-if-absent path at `plan_command.go:223`, reached under `!exists`. There is a small window between that check and the write where two instances could both seed, but both write the same draft-derived content, so it is cosmetic.

**The `/spec` gate change.** Base blocked `commandSpec` in plan mode and this head does not, which reads like a narrowed gate until you follow it. It is deliberate and safe: `/spec` now creates the draft session, clears plan state, exits plan mode and tells the operator, and `PermissionModeSpecDraft` is gated at the same three sites in `loop.go` as `PermissionModePlan`. Read-only to read-only. Flagging it because the next reviewer will trip on that diff line and should not have to re-derive it.

**Session switches.** `resetPlanForSessionSwitch` clears only in-memory state; the durable file is session-keyed so another session's plan is not touched. Applied at every switch site: `btw.go`, both in `session.go`, and `spec_mode.go`.

**Mechanical.** `internal/planmode`, `internal/tui`, `internal/tools` green on Windows. `internal/planmode` green on Linux and clean under `-race`. Builds for linux and darwin. The read-only enforcement itself is pre-existing on `main`, so this PR is the UX and persistence around a gate that already had teeth.

I did not audit the ACP paths or the TUI wiring in the depth jatmn did, and I am not claiming coverage of 8672 lines. The above is what I drove myself.

jatmn (dismissed):
lgtm

euxaristia:
@jatmn @Vasanthdev2004 All three findings from the last review are addressed on this head (`0f5c0087`): publish outcome contract in `plan_publication.go` + integration tests (`2fc7334a`), BTW return alignment (same commit), and ctx threading through `lockPlan` (`593a5744`). The last push was a docstring clarification and test isolation fix. Ready for re-review when you have a moment.

Vasanthdev2004 (approved):
Approving at `4f1d3626`.

Rebase first. Sixty-eight commits, and against an automatic merge of the old tip onto main the only differences are the conflict resolution in `agentResponseMsg` (the usage-model fields from main kept, your `planUpdate` field added) and the two commits since my last look: the empty-versus-absent `PlanSnapshot` distinction with its test, and the `TMPDIR` isolation in `TestStorageRejectsAllSandboxTempRoots`. Nothing from main went missing.

The publish boundary jatmn asked for is one thing now, and it is the right thing. `planPublication` wraps `update_plan` for the run: a candidate reaches durable storage before the tool reports success, a failed write is an error result to the model with the stage named, `accepted` is rehydrated from whatever writer won rather than from the rejected candidate, and the panel, the shared tool, `/plan status` and the durable file are all fed from that one snapshot. The BTW routing strips `syncTool` from hidden-parent messages so a parent run cannot touch the tool the side conversation is using, and `leaveBTW` restores the parent from its accepted snapshot. `TestPlanPublicationFailureKeepsConsumersAligned` drives all of that through the real run path, eighteen cases (conflict, lock, initial read, unreadable storage, success, clear; live, BTW and completion delivery) asserting tool, panel, status, durable file and the error the model sees. I put the old behaviour back two ways: accepting the rejected candidate into `accepted` fails seventeen assertions across the failure cases, and letting hidden-parent messages keep `syncTool` fails the BTW cases on the side tool. Context threading in `lockPlan` is pinned too: with the wait no longer watching the context, `TestPlanPublicationCancelsWhileLockIsHeld` gets `lock is held` instead of the deadline error.

One guard I could not exercise: the `ctx.Err() == nil` check before rehydration in `Run`. A cancelled context is refused by the tool before the write is attempted, so that branch is not reachable from outside; the gate that actually keeps a cancelled run's snapshot out of the panel is the run-ID check on the update path, and that one has its test. Harmless as written.

The five things I drove myself last time (path derivation, editor staging, write guards, the `/spec` gate, session switches) are the same patches, so they stand. Build, vet and gofmt clean; planmode green under `-race`, tools, tui and the run loop package green; CI 9 of 9 at the head.

jatmn (dismissed):

This review is against `4f1d3626548e347ac9e70e9b9f8f1c7f4ae43c56`, with merge-base and current `main` both at `99721c762f37cd43ac511007a5f51d1846df959e`. The session-scoped planning/editor workflow has concrete value. The fixes below preserve that scope and the existing read-only permission policy.

## Merge readiness

- The branch is current with `main` and GitHub reports `MERGEABLE`; no rebase or conflict blocker was found.
- All ten current checks pass, including Linux/macOS/Windows smoke, unit tests, race, lint and security. The older Windows failure described in the PR body is not the current check state.
- GitHub reports `BLOCKED` / `REVIEW_REQUIRED`.
- Closed, unmerged #854 is the predecessor. #829 concerns specialist orchestration, rather than a replacement for this session-scoped TUI editing workflow.

## Findings

### 1. [P2] Apply the concurrency check to initial editor seeding

`internal/tui/plan_command.go:214-223`

**Attribution and stated contract:** PR-introduced. Base has no durable editor seed. AGENTS.md requires: “Serialize the full read-modify-write sequence for lockfiles and shared stores.”

**Root cause:** The missing-file read happens outside the acceptance transaction, followed by unconditional `WritePlan`. Locking the eventual write does not protect the earlier absence decision.

**What fails:** Instance A observes a missing plan and prepares its local seed. Instance B publishes a distinct accepted plan before A acquires the writer lock. A then replaces B's accepted content. A controlled interleaving through `openPlanInEditor` and a competing `WritePlanIfUnchanged` reproduced the accepted plan becoming an empty format-marker file. This also happens before the unset-editor check. The earlier suggestion that both writers necessarily seed identical content does not cover a competing agent publication.

**In this PR:** Editor seed is unguarded; legacy format conversion, run publication and editor commit already compare baselines. Those are the full durable replacement entry points.

**Author fix:** Close the initial read/seed transaction with the existing conditional acceptance contract, preserving or reporting a competing accepted plan. Test a distinct concurrent first publication, including the no-editor path. Keep the already-guarded siblings intact; do not redesign the shared lock/store or change unrelated overwrite callers.

### 2. [P2] Preserve saved edits when the editor process fails

`internal/tui/plan_command.go:296-310`

**Attribution and stated contract:** PR-introduced; an incomplete sibling of the requested editor recovery behavior. My earlier review required that “rejected user work must remain recoverable.”

**Root cause:** `editorErr` takes a destructive cleanup path before the existing rejected-edit recovery path can inspect or preserve saved bytes.

**What fails:** An editor saves its staged file, then exits nonzero—for example, a wrapper fails after saving or the editor terminates after a save. `finish(false)` deletes that sole saved copy; the durable plan remains unchanged and the error contains no recovery location. This was reproduced through `finishPlanEditor` with a changed staging file and a nonzero exit.

**In this PR:** The process-exit branch omits recovery; conflict, baseline and storage failures use `PreserveStagedEdit`. Both Unix and Windows cleanup closures remove the file when preservation is false. Pre-launch parsing/staging errors have no newly authored edits; successful/no-op completion should retain normal cleanup.

**Author fix:** Apply the existing recovery outcome to saved changed bytes on editor-process failure: retain recoverable work, surface its path and release staging locks. Cover both cleanup implementations, recovery-name failure and later sweeping. Do not automatically accept a failed editor run or retain every untouched temporary file.

### 3. [P2] Bind durable storage to the root that passed containment checks

`internal/planmode/planmode.go:157-164`, `internal/planmode/read_unix.go:25-29`, `internal/planmode/write_unix.go:25-32`, `internal/planmode/read_windows.go:122`

**Attribution and stated contract:** PR-introduced storage boundary. AGENTS.md says: “Do not rely on pre-open path resolution (`EvalSymlinks` then open) for containment: that is a check-to-use race.” It also requires containment “at open/use time.”

**Root cause:** The physical containment check and the actual storage-root acquisition are separate. The latter follows the original lexical ancestors: Unix no-follow applies only to the last base component, and Windows deliberately follows the parent before opening the base. The checked physical root is not carried into locking and file access.

**What fails:** If `XDG_CONFIG_HOME` is an alias whose parent is workspace-writable, it can initially point at private config and pass validation, then be replaced with an alias into the workspace before use. A deterministic scheduling-boundary test of the same `writePlan` path used by `WritePlan` successfully wrote the plan into the workspace after that replacement. This requires a replaceable config ancestor; ordinary private config directories do not provide that opportunity.

**In this PR:** `ReadPlan`, unconditional/conditional writes, editor commits and `lockPlan` reacquire lexical roots; Unix readers/writers and Windows base acquisition need the same rule. Repeated prechecks do not bind the later open. `StageEditor` already uses its resolved staging directory, so the same original-alias redirection is not present there.

**Author fix:** Bind validation and subsequent lock/read/write operations to the same trusted storage root on both platform implementations. Add a replacement-between-check-and-use regression for the durable API siblings. Preserve legitimate private config aliases; this is not a request to prohibit all ancestor symlinks or redesign global sandbox/config handling.

### 4. [P2] Distinguish missing storage from an accepted empty plan during recovery

`internal/tui/plan_publication.go:57-59`

**Attribution and stated contract:** PR-introduced failure recovery. README states: “A failed save reports an error and keeps the accepted plan available to the panel, status display, and later reloads.”

**Root cause:** The recovery read ignores `exists`. A successful read reporting absence is parsed as an accepted empty plan, unlike initialization, which correctly retains the supplied in-memory fallback when no file exists.

**What fails:** With a nonempty accepted in-memory snapshot and no durable file, a first-save lock/storage failure can leave `ReadPlan` returning `exists=false, err=nil`. Recovery then clears `p.accepted` and publishes that empty value. A lock-path failure reproduced the error result followed by an empty `CurrentPlan`. The same distinction matters after a durable file disappears while an accepted snapshot remains in memory.

**In this PR:** Publisher initialization and BTW missing-file restoration preserve fallback; publisher failure recovery does not. Its empty recovery propagates through live and completion delivery to the shared tool, panel and status fallback. An existing empty file is a valid accepted clear and must remain distinct from absence.

**Author fix:** Honor the existing missing-versus-existing distinction in recovery, retaining fallback on absence while adopting readable competing accepted state when present. Test missing first-save recovery and existing empty-plan recovery through downstream consumers. Keep the correction in the new publication lifecycle.

### 5. [P2] Preserve failed-publication status through normal-turn completion

`internal/tui/model.go:2705-2706,2749-2752`, `internal/tui/plan_publication.go:51-62`

**Attribution and stated contract:** PR-activated / incomplete publish-outcome fix. Panel auto-completion predates this PR; the new durable failure/recovery path now feeds it an explicitly retained accepted snapshot. The requested outcome was: “Failed publish cannot produce sidebar ≠ status ≠ resume.” README repeats the failed-save consumer contract.

**Root cause:** The new publication outcome is not carried through terminal panel reconciliation. A failed tool operation can be followed by a successful agent final response; `msg.err == nil` does not mean the plan publication succeeded.

**What fails:** In Ask/Auto mode, a save fails and recovery restores an accepted pending plan. When the model returns a normal final answer, `completeRemaining` changes only the panel to completed. The shared tool, status and durable reload remain pending. Running the existing publication integration cases in Ask mode reproduced this for lock, startup-read and unreadable-store failures, through both live and completion-only delivery. The existing tests use Plan mode, which skips this condition.

**In this PR:** New recovery publication, live/fallback application and the changed completion condition must agree; hidden-parent BTW completion uses the same model update path. Plan mode and input/spec yields are already excluded.

**Author fix:** Carry the failed publication outcome far enough that terminal processing cannot immediately mutate its retained accepted plan only in the panel. Test normal-mode live/fallback and hidden-parent completion after failure. Preserve the pre-existing successful normal-turn auto-completion policy; a general completion-system rewrite is outside this finding.

### 6. [P2] Bind editor authorship to the bytes actually committed

`internal/planmode/planmode.go:426-429,473-477`, `internal/tui/model.go:1508,1524-1548`

**Attribution and stated contract:** PR-introduced editor event. The event asserts “I edited the plan file directly. Updated plan:” in the user's role. Existing editor tests distinguish actual edits from external reloads so the next turn does not receive false user-authored statements.

**Root cause:** `EditResult` carries only `Edited`/`Reload` flags. After commit releases the lock, the handler rereads mutable durable content and attributes those later bytes using the earlier `Edited=true` flag.

**What fails:** A commits an editor change; B conditionally publishes another plan using A's committed value as baseline; A then handles its editor-completion message. The persisted user event contains B's plan. This exact sequence reproduced a `role: user` event claiming “other writer content.” A competing clear similarly makes A claim the user cleared the plan.

**In this PR:** Commit outcome, editor callback, reload, replacement/clear confirmation text and user-role event must share the correct committed identity. Unchanged editor sessions and external reloads must remain unauthored. Loading a newer durable plan for the panel is a separate decision from attributing it to the user.

**Author fix:** Carry or verify the committed snapshot identity through the new editor outcome and author only the user's committed content. Test a competing replacement and clear between commit and message delivery. Do not solve this by disabling concurrency or rewriting the session-event framework.

### 7. [P2] Carry accepted plan outcomes into the existing agent-state consumers

`internal/tui/plan_publication.go:43-62`, `internal/tui/model.go:1541-1548`, `internal/agent/loop.go:1540`

**Attribution and stated contract:** PR-activated integration gap. Base's valid allowed `update_plan` call directly updates memory; this PR introduces durable rejection and a separate editor replacement/clear producer. README says: “Plan updates are accepted only after saving succeeds.” The existing compaction contract preserves the active plan “exactly rather than being paraphrased away,” tested by `TestCompactPreservesActivePlan`.

**Root cause:** Accepted state reaches TUI consumers, but the agent's plan projections still infer authority from historical tool-call arguments. The new accepted-outcome boundary does not reach those consumers.

**What fails:** After accepted plan A, a syntactically valid B can be rejected by the new persistence layer. When those turns are compacted, `extractLatestPlan` nevertheless selects B and injects it as the structured active plan. Likewise, after the new editor event replaces or clears A, compaction selects the old tool-call A. Tests of these transcript sequences reproduced all three incorrect preserved-state blocks; a prose summary mentioning the error/edit does not correct the conflicting structured block.

**In this PR:** Bridge the new publisher's accepted/rejected/recovered outcomes and editor replacement/clear into the agent-facing snapshot/history integration. Affected existing consumers are compaction selection (`compaction_preserve.go:64-75,518-523`), task-state projection/parity (`loop.go:624-627`, `task_state.go:145-152,543-555`) and guardrail pending counts (`guardrails.go:531-539`). The latter two also observe proposed arguments before execution, so fixing only the compaction text leaves another view of rejected state. A competing accepted durable value may never have appeared in a successful historical tool call.

**Author fix:** Feed these PR-owned acceptance outcomes into the existing plan-state consumers, preserving canonical fields and the distinction between an accepted clear and absent state. Cover rejected publication, recovery to a competing plan, editor replacement and editor clear through compaction and adjacent projections. Keep existing transcript scrubbing and typed-snapshot privacy. Do not broaden this into generic compaction redesign, ACP/exec persistence, or unrelated pre-existing failed-call cleanup; the defect is the missing integration for this PR's new acceptance boundary. This finding does not claim that a model necessarily executes the incorrect plan or that pending counts alone force premature completion.

## Complete repair surface

Please address the findings together in one repair head. Each **Author fix** closes its listed contract across the PR-owned siblings; changing only its first cited line is insufficient where a producer/consumer boundary is involved.

| Lifecycle / ordering | Required behavior | Current result |
|---|---|---|
| Seed → durable acceptance | Absence decision participates in concurrency check | Finding 1 |
| Stage → editor exit → commit/recovery → cleanup | Saved rejected work remains recoverable; successful/no-op cleanup remains | Finding 2 |
| Check root → lock → read/replace | Same trusted physical root throughout | Finding 3 |
| Failed publication → restore → live/fallback/BTW → terminal | Preserve one accepted plan; missing differs from clear | Findings 4–5 |
| Editor commit → reload → authored event | Authorship refers to committed bytes | Finding 6 |
| Accepted update/edit → agent projection → compaction/continuation | Accepted plan survives, not the proposed/superseded plan | Finding 7 |
| Session switch / plan lifetime / loops and goals | Session-scoped mode, read-only gates and paused continuations | Preserve existing tested behavior |

| Shared plan field | Producer / admission | Loader / consumer | Tests / operator surfaces |
|---|---|---|---|
| Content and notes | String validation, trimming, multiline values | Legacy/v2 parser; typed snapshot; panel/status | Multiline, literal slash/pipe/Notes tests; README marker/quoting |
| Status | Normalization; at most one active item | Canonical reload/display; accepted-state projections | Canonical status tests; failed-publication completion in finding 5 |
| IDs / ordering | Optional/default positional IDs | Order retained, IDs may be regenerated on file parse | No new uniqueness requirement |
| Empty / absent | Empty accepted list is a clear | Missing file preserves fallback; editor clear supersedes old plan | Findings 4 and 7; existing empty-publication cases |
| Acceptance / provenance | Persisted success, recovered accepted state, committed editor bytes | Live/fallback/BTW, authored event, agent history | Findings 1 and 4–7; README acceptance and editor confirmation/event copy |
| Size / privacy | Durable path components bounded; no stated body cap | Protected storage, scrubbed transcript, nonserialized typed snapshot | Preserve existing path/privacy tests; no invented body limit |

Malformed-input surfaces checked: unknown `/plan` subcommands, busy/mode guards, empty or malformed editor command, wrong tool array/item/string types, missing/invalid editor baseline, unsafe/nonregular paths and absent storage. These should retain their existing error behavior. The operator surfaces to keep consistent are README lifetime/acceptance/editor syntax, command help, mode notices, status/errors, and editor reload/clear/authorship messages. There are no changed HTTP or generated-env surfaces.

## Validation and preserved scope

Focused race tests passed for storage, TUI plan/BTW/session/spec/loop/goal paths, tools and agent plan/permission paths. `go vet` for the affected packages and diff whitespace checks passed. Targeted reproductions establish the failure paths above. Local native Windows/macOS execution and a full local suite were not repeated; current CI is green on all reported checks.

Keep the verified process lock, required editor baseline, sandbox temporary-root exclusions, cancellation responsiveness, literal format compatibility, canonical status, no-op/external-reload distinction, restored agent assertions and session-scoped lifetime. ACP migration, `beforeTool` policy changes and permission-lifetime redesign are outside this repair.

euxaristia:

Issue #554, opened 2026-07-06
Multi model usage


2krishnakishore:
### Problem

Feature request
. Can we have multi model configuration for planning for coding and for review
Example

1. Plan with openai
2. Execute with sonet
3. Review with deepseek

All of them  wihh have their backup individually

So essentially w can do best all and manage cost as well 

### Requested behavior

A role can be assigned for planning for execution and for reviewing to agent and use different models for that 

### Why it belongs in Zero

Today all cli are confined to single task by model family, zero will have advantage to use multiple ones at the same time and bring best of all

Vasanthdev2004:
@2krishnakishore sorry this sat unanswered for a month.

Not a no. We do want role-based model selection. It is not a small config addition though: choosing a model per role only pays off if the role also carries its own context, tools and permissions, and Zero already has the specialist system doing delegated work with its own configuration. A separate roles-to-models map bolted next to it would give us two ways to express the same thing, which is worse than either.

Keeping this open as the tracking issue for the feature. What it needs before code is a design that extends the specialist path rather than sitting beside it. If you want to move it forward, the most useful next step is a concrete proposal on this issue: how a role maps onto a specialist, and what happens when a role's backup model lives on a different provider from its primary.

snoop244:
I want my Zero "Sr." agent, Sol, to delegate bounded coding tasks to Luna which is, theoretically, 10x cheaper. In Github Copilot I can set the sub-agent to a different model. I'm hoping to ditch GH Copilot for Zero... so far, so good, but this is a sticking point for me. In Copilot, the subagent is stateless and gets all its context on each call from the senior agent - which I think works well. 


> 
> Zero already has the specialist system doing delegated work with its own configuration. A separate roles-to-models map bolted next to it would give us two ways to express the same thing, which is worse than either.
> 

... this suggests we can already set different models for the Zero main model and the specialist. I set it in the frontmatter for the specialist, but Luna was ignored and it used Sol. 

In trying to use Zero with Sol to solve this problem Sol came back with:
"Specialist model validation rejects valid provider models because it uses only Zero’s internal
model registry rather than the active provider’s discovered catalog"

Vasanthdev2004:
Short answer: a good chunk of this already works, and the specific thing asked for here (a senior agent handing a bounded task to a cheaper model) ships today. The part that genuinely does not exist is switching models by phase inside one session.

## What already works

[...]

PR #1009, opened 2026-09-04
fix(specialist): keep the pinned model when a specialist is resumed

Vasanthdev2004:
A specialist pinned to a cheap model ran on that model exactly once.

`Metadata.Model` (`internal/specialist/manifest.go:31`) exists so a bounded, delegated task can run on something cheaper than its parent, and `BuildArgs` puts it on the child argv through `appendModelArgs` (`exec.go:728`). `BuildResumeArgs` never called it. So the moment the orchestrator resumed that specialist, the child fell back to whatever the parent's configured model resolved to.

Nothing surfaces it. The resumed child starts normally and does the work, so the only symptom is the bill. Cost-motivated delegation quietly stops saving anything.

Resuming does not restore the recorded model on its own, which is why the flag has to be passed again rather than relied upon: `sessions.PrepareExec` records the model a run used but never feeds it back into provider construction.

### The fix

`BuildResumeArgsInput` now carries `ParentModel` and `ParentReasoningEffort`, the same fallbacks the fresh path takes, and `runResume` passes what `TaskRunOptions` already held. The reasoning-effort rule travels with the model unchanged: the parent's effort is inherited only when the manifest pins no model of its own, because a manifest that chose a different model has not agreed to the parent's effort for it.

### Tests

Both builders are driven and compared, so the two paths cannot drift apart again: a pinned model survives resume, an unpinned one still inherits the parent's, both halves of the effort rule hold, and the flag keeps its position relative to `--auto` in both.

The second commit exists because of something the falsification caught. Every builder test calls `BuildResumeArgs` directly, and deleting the `ParentModel` field from the `runResume` call site still compiled and still passed all of them. The defect lived at the call site, not in the builder, so there is now a test that goes through the real `Run` dispatch with the `RunChild` seam capturing argv. Removing that one field fails it with:

```
the resumed child was launched with no --model, so it runs on whatever the
config default resolves to
```

Removing the `appendModelArgs` call itself fails the builder tests the same way.

Refs #554, where this came up while working out what Zero can already do for per-role models. Worth noting the answer there is "more than people think": manifest-level model selection already works, and this was the gap that made it look like it did not.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resumed specialist tasks now retain their configured model when one is pinned.
  - Specialists without a pinned model inherit the parent task’s model during fresh runs and resume operations.
  - Reasoning-effort settings are consistently preserved or inherited across specialist launches.
  - Model and autonomy options are now applied in the correct order, improving reliability for fresh and resumed tasks.

### DIFF
--- a/internal/specialist/exec.go
+++ b/internal/specialist/exec.go
@@ -76,12 +76,17 @@ type BuildArgsInput struct {
 }
 
 type BuildResumeArgsInput struct {
-	SessionID      string
-	Prompt         string
-	CurrentDepth   int
-	Manifest       Manifest
-	Cwd            string
-	PermissionMode string
+	SessionID    string
+	Prompt       string
+	CurrentDepth int
+	Manifest     Manifest
+	Cwd          string
+	// ParentModel and ParentReasoningEffort are the same fallbacks the fresh
+	// path takes, and they are here for the same reason: appendModelArgs uses
+	// them only when the manifest pins nothing of its own.
+	ParentModel           string
+	ParentReasoningEffort string
+	PermissionMode        string
 }
 
 type BuildArgsResult struct {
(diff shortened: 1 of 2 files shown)


Vasanthdev2004:
Added one more test commit at `c4e7d816`.

While falsifying the resume fix I found that `runFresh` carries a byte-identical `ParentModel: options.ParentModel,` line, and deleting THAT one was silent too. My own falsification actually hit the fresh occurrence first and reported the resume test as passing, which is how I noticed.

So both call sites are covered now. Deleting the fresh one fails with:

```
the fresh child was launched with --model "" (present=false), want the parent's
```

The resume half is still what this branch repairs. This just stops the pair drifting again in the direction nobody was looking.

jatmn (approved):
LGTM

anandh8x (approved):
## Review at c4e7d81

**Verdict: Approve.** A quiet billing bug with a tight fix and genuinely good tests.

### What it does

A specialist pinned to a cheap model ran on that model exactly once — on resume the child silently fell back to the parent's configured model. Same work, bigger bill, nothing surfaced. `BuildResumeArgsInput` now carries `ParentModel`/`ParentReasoningEffort` (the same fallbacks the fresh path takes) and `runResume` passes through what `TaskRunOptions` already held. The effort rule travels unchanged: parent effort inherits only when the manifest pins no model of its own.

### What's good

- **Both builders driven side by side.** Pinned model survives resume, unpinned still inherits, both halves of the effort rule hold, and `--model` keeps its position relative to `--auto` in both. The two paths can't drift apart again without a test firing.
- **Call-site coverage, not just builder coverage.** `TestRunResumePassesTheParentModelThroughToTheChild` goes through the real `Run` dispatch with the `RunChild` seam capturing argv — the author documents that deleting the field from the call site still compiled and passed the builder tests, so this test is the one that actually guards the fix. The fresh path gets the same guard.
- **Execution semantics unchanged.** The `!ok` error branches and flag ordering are identical; this only threads two already-held values one call deeper.

### Verification (mine, isolated worktree)

- `go build`, full `internal/specialist` suite green.

No blocking notes. Merge-ready.

PR #1010, opened 2026-09-05
feat(mcp): resolve MCP server secrets from the credential store

beardthelion:
> **No linked approved issue.** `CONTRIBUTING.md` requires community PRs to link an issue carrying `issue-approved`, and this has none. It is opened as a draft under the "team members may open pull requests as part of the internal development cycle" clause. If that reading is wrong, treat this as awaiting an issue rather than awaiting review, and I will open one.

## What changed and why

Enabling an MCP server that needs secrets meant writing those secrets into `config.json` in plaintext. Server entries can now name a credential instead.

`envFrom` maps an environment variable to a credential-store name: the config file holds the **name**, the existing credential store holds the **value**. References resolve at stdio spawn and merge over the verbatim `env`. Every failure names the credential and never the value, and lookups run in sorted order so a server missing several always names the same one first.

It also seeds a `memlawb` entry, disabled, with its URL and namespace verbatim and its passphrase and API key as references. memlawb is a zero-knowledge memory backend whose passphrase **is** the encryption key, which is exactly the kind of value that should not sit in a config file.

## Three things that were not obvious

**Seeding a disabled default took more than the flag.** `zero mcp enable` enables by *deleting* the `disabled` key, which is right for a default that ships enabled and useless for one that does not: with the key gone, the seeded `Disabled: true` survives the merge, so the command reported success and wrote nothing. Enabling a ships-disabled default now writes an explicit `false`, and an entry with no key reads as the default's shipped state.

**Two files outside the intended set had to change, both because the new field opened a path through them.** `mergeMCPServer` had no clause for it, so a user-level reference was silently dropped when merged over the default. And `hasInheritedMCPCredentialMaterial` needed extending: a project config that retargets `command` or `url` while inheriting the user's references would hand a repository-chosen binary the user's stored passphrase. That is the same hole the existing headers-and-env guard exists to close, and the new field opened a fresh channel into it.

**`IsUnconfiguredDefault` needed proving, not changing.** It compares whole structs, so the new field joins automatically; a hand-written field-by-field comparison would have ignored it silently. Both directions are now pinned, along with the resolved default surviving the retired-default migration.

## Verification

```
go build ./...    ok
go vet ./...      clean
gofmt -l          0 files
go test ./...     all packages pass, 0 failures
```

Fourteen guards were each removed once and the named test observed red: the missing-credential refusal, store-error propagation, the remote-transport rejection, the identity contribution, normalization carry-through, the child-env merge, the disabled seed, the merge clause, the project-inheritance clause, JSON decode, the ships-disabled state read, the explicit-false write, the enable notice, and the empty-value rejection.

**One guard is deliberately not claimed as proven.** The early return that stops a server without references opening the credential store is pinned by a zero-lookups test, but removing it does not turn that test red, because an empty reference map resolves to an empty loop either way. It is a pinned property, not a proven guard.

The absence assertion is paired as it has to be: one test asserts neither secret value appears in the written config, and a second runs the same reader over the same writer with values passed inline and asserts both **are** present. Without the second, the first passes against a writer that writes nothing.

## End-to-end, in a container with neither Node nor Bun

Run against the memlawb standalone binary, since that is the deployment this is for:

```
runtimes: node=NO bun=NO
memlawb --version -> 0.1.0
zero mcp secret set memlawb-passphrase   stored (encrypted-file backend)
zero mcp secret set memlawb-api-key      stored
zero mcp enable memlawb                  wrote {"memlawb":{"disabled":false}}
zero mcp tools list                      5 memlawb tools discovered
```

The control is what makes that evidence rather than a smoke test. The namespace was seeded beforehand under the real passphrase, so memlawb's startup check refuses unless the stored value round-trips correctly:

- correct passphrase in the store → **5** tools discovered
- wrong passphrase in the store → **0** tools discovered

That proves the credential value reaches the spawned child. Listing tools against an empty namespace would have proved only that a process started.

## Known limitations

- **Not automated, and nothing here stands in for it:** completing an actual `memory_save` and `memory_recall` through a zero session requires an agent turn against a real model. I proved the plumbing up to tool discovery, not the model's behaviour.
- No `--env-from` flag on `zero mcp add`; references are reachable via the seeded default or hand-edited JSON. Small addition if wanted, but it was outside the scope I was working to.
- `internal/tui/mcp_state.go` renders `Env` and not `EnvFrom`, so the TUI view of a referencing server is incomplete. Only names would show, so nothing leaks. Flagged, not touched.
- `MemlawbMinimumVersion` is `0.1.0` and must move if the release shipping the standalone binary bumps.
- A self-hoster with no API key cannot use the seeded entry as-is, because the missing credential fails the connect by design. That follows the intended "a missing credential must fail the connect", but it is a product call worth confirming.

## Summary by CodeRabbit

- **New Features**
  - Added `zero mcp secret set <name>` for securely storing MCP secrets through interactive or piped input.
  - Added credential references for injecting stored secrets into local MCP servers without writing secret values to configuration files.
  - Added the Memlawb MCP server as a disabled-by-default option, with setup guidance and minimum-version information.

- **Bug Fixes**
  - Enabling built-in disabled servers now correctly records their enabled state.
  - Added validation for credential references and clearer errors when credentials are missing or used with unsupported transports.

### DIFF
--- a/internal/cli/extensions.go
+++ b/internal/cli/extensions.go
@@ -171,6 +171,8 @@ func runMCPWithContext(ctx context.Context, args []string, stdout io.Writer, std
 		return runMCPToggle(args[1:], stdout, stderr, deps, true)
 	case "check":
 		return runMCPCheck(ctx, args[1:], stdout, stderr, deps)
+	case "secret":
+		return runMCPSecret(args[1:], stdout, stderr, deps)
 	case "permissions":
 		return runMCPPermissions(args[1:], stdout, stderr, deps)
 	case "tools":
@@ -619,7 +621,13 @@ Commands:
   list              List configured MCP servers, or tools with --tools
   oauth             Manage OAuth credentials for remote MCP servers
   permissions       Manage persistent MCP tool permissions
+  secret set <name> Store a secret an MCP server references by name (envFrom)
   tools             Inspect configured MCP tools
+
+Memory (memlawb): store the two secrets, then turn it on.
+  zero mcp secret set memlawb-passphrase
+  zero mcp secret set memlawb-api-key
+  zero mcp enable memlawb
 `)
 	return err
 }
--- a/internal/cli/mcp_config.go
+++ b/internal/cli/mcp_config.go
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/charmbracelet/x/term"
+
 	"github.com/Gitlawb/zero/internal/config"
 	"github.com/Gitlawb/zero/internal/mcp"
 	"github.com/Gitlawb/zero/internal/redaction"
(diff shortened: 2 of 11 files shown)


jatmn (changes requested):

## Merge readiness

- [P2] Mark PR ready for review when intended
  GitHub shows this pull request is still a **draft**. CI smoke and security checks pass. The two items below are narrow CLI fixes on the new `secret set` path; they do not block the credential-reference design itself.

## Findings

- [P3] Store the secret bytes the user entered, not a trimmed copy
  `internal/cli/mcp_config.go:361`
  `readMCPSecretValue` runs `strings.TrimSpace` on both the interactive (`term.ReadPassword`) and piped (`io.ReadAll`) paths before `store.Set`. That means a value like `"  my-key  "` is stored as `"my-key"`. For memlawb the passphrase is the encryption key, so a user who deliberately includes leading or trailing spaces would get a different key than they typed and memlawb would fail its round-trip check with no obvious cause.

  This sits awkwardly next to the spawn path you added: `copyStringMap` in `internal/mcp/config.go` documents that env values are preserved verbatim, and the whole PR is about keeping secret material out of config while passing the exact value through at spawn. Trimming at store time breaks that contract one step earlier.

  **Root cause:** `readMCPSecretValue` treats the input like a CLI token to normalize rather than opaque credential material.

  **Suggested fix (scoped to this command):**
  - Interactive path: keep `string(typed)` as the value; use `strings.TrimSpace` only in the emptiness check (`if strings.TrimSpace(value) == ""`).
  - Piped path: remove at most the trailing line ending (`\n` / `\r\n` from the pipe); do not trim spaces inside the value.
  - Add a test that pipes `"  spaced-secret  \n"` and asserts `store.Get` returns the inner spaces intact. Your existing stdin tests already cover the happy path; this pins the regression.

  I am not asking you to change provider-key migration (`MigratePlaintextProviderKeys` also trims) or revisit credstore semantics — just keep this new secret-entry path faithful to what gets resolved into the child env.

- [P3] Write interactive prompts to stderr so `--json` stdout stays machine-readable
  `internal/cli/mcp_config.go:351`
  `runMCPSecretSet` passes `stdout` into `readMCPSecretValue`, and the TTY branch writes the hidden prompt and the post-entry newline there. When `--json` is set, the success payload also goes to `stdout` via `writePrettyJSON`. So `zero mcp secret set <name> --json` on a terminal emits human text before the JSON document and breaks any parser expecting stdout to be exclusively JSON.

  Piped stdin — the path your tests, help text, and memlawb setup flow use — is unaffected because the TTY branch is not taken.

  **Root cause:** one writer (`stdout`) is doing double duty as the JSON result channel and the interactive prompt channel.

  **Suggested fix (scoped to this command):**
  - Give `readMCPSecretValue` a separate prompt writer (pass `stderr` from `runMCPSecretSet`) and move both the `"Value for … (input hidden): "` line and the trailing `Fprintln` after `ReadPassword` to that writer.
  - Leave piped stdin behavior unchanged.
  - Add a test that runs `secret set` with `--json` and a fake TTY stdin, then asserts `stdout` decodes as JSON with no leading non-JSON bytes. No broader CLI audit needed — other commands in this diff do not add interactive input on the JSON path.

  CodeRabbit's inline comments on these two spots are still valid on head; the above is the same root cause with the concrete shape I'd expect the fix to take.

PR #1011, opened 2026-09-05
fix(sandbox): deny SSH keys and GPG stores with fail-closed discovery

### LINKED ISSUE #815: Sandbox can read SSH private keys, the GPG keyring, and git credential stores

euxaristia:
Fixes #815

## Summary

Protect SSH private keys and GPG stores in the Unix credential baseline. Refuse sandboxed execution when bounded SSH discovery is incomplete or Linux cannot safely enforce a requested selective key or symlink deny.

## Changes

- Discover default and custom-named private keys, SSH config references and `Include` files, and GPG homes. Retain public SSH support files in pathname policies and preserve existing Git credential denies.
- Propagate discovery failures through profile construction and command planning. Directory-entry and config-Include limits reject incomplete profiles.
- Inspect Linux files through pinned descriptors and reject nonregular files before opening them for I/O. On other Unix systems, use descriptor-relative, no-follow opens for every path component and verify the opened file before reading.
- Remove Linux parent-reconstruction and resolved-symlink masking paths that relied on mutable paths. Reject absent explicit Linux `denyRead` paths and classification failures during planning.
- Retain absent conventional and configured SSH-key candidates through profile construction so the Linux launch planner fails closed before a later host write can expose them.
- Give command-planning tests explicit directory denies in isolated temporary homes so the stronger SSH guard preserves their original network, runtime, and token-store assertions. The unavailable-backend fixture grants only its own empty SSH directories.
- Release runtime leases in command-planning tests, including the degraded plan, before temporary-directory cleanup.
- Respect explicit read grants while retaining denies for referenced keys outside the granted SSH directory.

Linux refuses selective SSH-key masks even when the candidate files do not exist yet. This includes machines without SSH keys, because a trusted host process may create one later. An explicit deny of an existing containing directory covers its keys but also hides public config and known-host files. macOS retains pathname-based enforcement; automatic credential discovery remains disabled on Windows. The Safety Model documents these platform limits and refusal conditions.

## Test plan

Completed with Go 1.26.6:

- `make fmt-check`, `go vet ./...`, and full Linux `go test ./...`.
- `go run ./cmd/zero-release build` and `go run ./cmd/zero-release smoke`.
- Pinned golangci-lint v2.12.2: zero issues. Pinned govulncheck v1.3.0: no vulnerabilities.
- Linux race tests for `internal/sandbox`, Windows sandbox package tests, and macOS arm64 sandbox test cross-compilation.
- `git diff HEAD --check`.

The prior Windows CI foreground-server test failed before reporting its listening address; five unchanged local reruns passed. The affected Windows sandbox package passes with both command-plan leases explicitly released before fixture cleanup. The macOS check above is cross-compilation, not native test execution.

Addresses the applicable feedback from #990 and this PR:

- End-to-end directory and Include limit regressions fail on the original code with `incomplete directory entry discovery allowed command planning: <nil>` and `incomplete config Include match discovery allowed command planning: <nil>`, respectively, and pass with the fix.
- The absent-SSH-key regression exercises missing directories, empty directories, and configured external keys. The previous head failed with `profile constructed before key creation allowed command planning (created=false): <nil>`; the fix refuses launch before and after host-side key creation. The policy JSON regression checks that all six absent conventional key paths reach the published profile; the old profile failed with `manager absent SSH key protection = []string(nil)`.
- Absent explicit denies reject command planning instead of disappearing. Inspection pins the opened inode, and unsupported selective and symlink masks are refused.
- Symlink tests require rejection while retaining Seatbelt policy assertions. A regression verifies that granting access to the SSH directory does not remove denies for referenced keys outside it.

## Summary by CodeRabbit

* **Security**
  * Strengthened sandbox protection for SSH keys, GPG stores, Git credentials, and related credential paths.
  * Sandbox execution now stops when credential discovery is incomplete, exceeds safety limits, encounters unsafe symlinks, or cannot guarantee deny rules.
  * Added safeguards against inspecting special files, redirected paths, and files replaced during inspection.
  * Improved protection for credentials created after sandbox policy setup.

* **Documentation**
  * Expanded documentation on credential discovery, safety limits, platform behavior, failure conditions, Atomic Chat, and web address validation.


Vasanthdev2004 (dismissed):
Approving at 5c428121. It does what it says, and I checked that against the rules a backend actually receives rather than against the candidate list.

Seeded a home with a key, a public key, `known_hosts`, `config`, and a private key deliberately named `work_key` so nothing about its name gives it away, then read the profile back:

```
DENIED:  ~/.gnupg  ~/.ssh/id_ed25519  ~/.ssh/work_key  ~/.git-credentials  (+ the existing set)
KEPT:    ~/.ssh    ~/.ssh/config  ~/.ssh/known_hosts  ~/.ssh/id_ed25519.pub
```

`work_key` is the one that convinced me the content sniffing is worth its complexity. A name-only rule would have missed it, and it is the shape a real user ends up with the moment they run `ssh-keygen -f ~/.ssh/work_key`. Sniffing before the public-name exemption is the right order too, so an `IdentityFile` pointing at something called `config` or `known_hosts` with a key payload in it is still denied.

The documented opt-out works: with `allowRead: ["~/.ssh"]` the key comes back out of the deny list.

## The trade this makes, which the body does not say

A sandboxed `git push` over SSH stops working by default. The comment this PR removes from `git_credential_deny_test.go` is where that was previously written down and deferred:

> Denying `~/.ssh` as well would stop a sandboxed git push over SSH from working, which is a functional trade that issue tracks separately

Denying `id_*` makes that trade. It is narrower than denying the directory, but from the ssh client's point of view it is the same outcome: the key it needs reads as `/dev/null`.

I think it is the right default. A coding agent authenticating to a remote as the user with the user's key is close to the top of the list of things a sandbox exists to stop, `~/.aws` and `~/.azure` already work this way, and the opt-out is real. But it is a default that people notice on upgrade, so it belongs in the release notes in those words rather than as "improved sandbox protection". Worth a line in the PR body too.

## Windows is unaffected

`credentialDenyReadPaths` returns empty on Windows and always has, so no part of this ships there:

```go
if runtime.GOOS == "windows" {
    return credentialDenyPaths{}
}
```

Pre-existing and documented in the comment above it, not something this PR introduces. Raising it because the body and #815 both read as though the protection is universal, and a Windows user reading the release note would reasonably think their keys are covered. Either say so, or leave #815 open for the Windows half.

Profile construction is on the per-command path, and the walk adds to it. Measured on Linux against main with the same seeded home, and guarded so I was not timing a code path that never ran (my first attempt measured Windows, where the whole thing is skipped, and reported a flat cost that meant nothing):

```
main:  0.63 - 0.71 ms   flat
head:  0.98 ms (empty)  1.10 ms (+20 files)  1.31 ms (+100 files)
```

Sub-millisecond and it scales with `~/.ssh` rather than with anything unbounded, so this is a note and not an objection.

A key created after the profile is built is not in that profile. The next command rebuilds, so the window is one command, and it only matters for a key that appears mid-session under a name the well-known list does not cover. Worth knowing rather than worth fixing.

The `Include` and `IdentityFile` parsing is bounded the way I would want (depth 16, 64 matches, 1 MB, no directory-symlink following) and I could not get it to over-deny: an `IdentityFile` naming a directory is not denied, and `/`, `$HOME`, `~/.ssh`, and `/dev/null` are all exempted explicitly.

The golden policy baseline in `internal/cli/sandbox_test.go` is extended rather than relaxed, which is the right way round: the exported JSON is the contract and it now names the keys.

`internal/sandbox` is green on Linux and on Windows here.

jatmn (dismissed):
lgtm

Vasanthdev2004 (changes requested):
The SSH and GPG work I approved at 5c428121 still looks right, and I re-verified it. This is about the new refusal in c304682e, which I had not seen.

Refusing rather than under-protecting is the correct instinct, and it is the same call I have argued for elsewhere in this codebase: if discovery could not enumerate the keys, the sandbox cannot honestly claim they are denied, so failing is better than pretending. Failing closed rather than downgrading is right too. The problem is what counts as "incomplete".

## An ordinary `~/.ssh` now stops the sandbox entirely

Driven through `PermissionProfileFromPolicy` on Linux, with a normal key and `known_hosts` present:

```
~/.ssh shape                            result
plain                                   runs
20 per-host keys                        runs
one unreadable file (mode 000)          runs
an unreadable subdirectory              runs
a dangling symlink                      runs
257 entries                             REFUSES: directory entry limit exceeded
nesting deeper than 8                   REFUSES: directory depth limit exceeded
```

The error-handling cases are all graceful, which is the part I expected to be wrong and was not. The two that refuse are limits, not errors.

257 entries in `~/.ssh` is not an attack. The most common way to get there is `ControlMaster`: with `ControlPath ~/.ssh/cm-%r@%h:%p`, which is what most tuning guides suggest, a socket accumulates per host per user per port. A few hundred is a busy week. The cap counts entries of any type, so sockets, stale ones included, all count.

The consequence is that no sandboxed command runs at all, with:

```
cannot guarantee credential protection: SSH discovery incomplete for ~/.ssh: directory entry limit exceeded
```

## And there is no way out

I checked the obvious one:

```
default                    errors=1
allowRead: ["~/.ssh"]      errors=1
```

The explicit grant does not clear it. So a user in this state can delete files out of `~/.ssh`, or turn the sandbox off. Neither is a reasonable thing to work out from that message.

## What I would do instead

The cap looks like it exists to bound allocation, and `d.ReadDir(n)` already does that. Paging gives the same bound without the outage: read in chunks of 256 and keep going until EOF, rather than treating "more than one chunk" as a failure. The walk is looking for keys, and a directory being large is not a reason it cannot.

Same for depth: descending further is cheap next to refusing to run.

If a hard cap is wanted anyway, make it much larger, count it across the whole walk rather than per directory, and say in the message what the operator should do. Right now the message names a condition without a remedy.

Worth keeping the refusal for the cases where discovery genuinely cannot answer. Those are the `err.Error()` paths, and they already behave well.

## Re-verified from the earlier review

The deny set is unchanged at this head: `~/.gnupg`, `~/.ssh/id_ed25519` and a content-sniffed `work_key` denied; `~/.ssh`, `config`, `known_hosts` and `id_ed25519.pub` still readable.

Still worth putting in the release notes that a sandboxed `git push` over SSH stops working by default, and that none of this ships on Windows, where `credentialDenyReadPaths` returns empty before any of it runs.

Vasanthdev2004 (dismissed):
Fixed, and fixed the right way round: the refusals are gone because discovery got better, not because the guard got weaker. That was the thing worth checking.

Ordinary shapes all run now:

```
~/.ssh shape                  before      now
plain                         runs        runs
20 per-host keys              runs        runs
one unreadable file           runs        runs
an unreadable subdirectory    runs        runs
a dangling symlink            runs        runs
257 entries                   REFUSES     runs
nesting deeper than 8         REFUSES     runs
```

And the protection is still there past the old cap. A custom-named private key placed last in the directory, so name alone will not find it and only the walk plus content sniffing will:

```
  10 noise entries -> hidden key denied=true  errors=0
 300 noise entries -> hidden key denied=true  errors=0
 900 noise entries -> hidden key denied=true  errors=0
```

That is the answer to the question I actually had. Deleting the cap and losing the key past entry 256 would have produced the same clean run list and a silent hole.

Keeping the refusal for the `err.Error()` paths is right. Those are the cases where discovery genuinely cannot answer, and they were already behaving well.

One note, not blocking. Paging means a large `~/.ssh` is now walked rather than refused, and that walk is on the per-command profile build:

```
    0 entries ->  2 ms
  300 entries ->  4 ms
 2000 entries -> 17 ms
```

Fine, and clearly the right trade against refusing to run. Worth knowing it scales with the directory, since the sniff opens every candidate.

`internal/sandbox` green here on Windows.

jatmn (changes requested):

## Merge readiness

- Mergeability is clean (`MERGEABLE`, base `aadb4a27` is current `main`). All required CI checks on head `b707a66` passed.
- Supersedes closed #990; no duplicate open PR for the same scope.
- Vasanthdev2004 approved after the paging fix. The items below were not addressed on head.

## Why this finding remains after many rounds

This PR merges command-controlled environment for credential roots (`HOME`, `GNUPGHOME`, etc.) in `profile.go:328-334`, but SSH config parsing still resolves `IdentityFile ${VAR}` through `os.Getenv` only (`ssh_key_deny.go:483`). OpenSSH at exec time resolves those variables from the **child** environment, which includes command-injected values. Discovery and runtime therefore disagree about which paths exist. Tests exercise process env (`t.Setenv("SSH_KEY_DIR", …)` in `ssh_gpg_deny_test.go:1264`) and explicitly expect unset vars to be dropped, but not the command-only case that MCP and `exec_command` actually use.

## Findings

- [P1] Resolve SSH config `$VAR` from the command environment, not only the process environment
  `internal/sandbox/ssh_key_deny.go:444-490` (`expandSSHConfigPathEnv`, called from `expandSSHConfigPath` → `collectConfigPaths`)
  `internal/sandbox/profile.go:328-334` (command env already merged for credential roots)

  **What happens today.** `credentialDenyReadPaths` calls `appendUntrusted(credentialPathOptionsFromEnvironment(..., commandEnv))`, so command `HOME` (and other roots) participate in discovery. Inside that discovery, `IdentityFile ${SSH_KEY_DIR}/work` is expanded with `os.Getenv(name)` when `name != "HOME"`. If `SSH_KEY_DIR` exists only in `CommandSpec.Env` — typical for MCP overrides and explicit `exec_command` env — expansion returns `""` and the path is silently dropped (no `DiscoveryErrors` entry; see `ssh_gpg_deny_test.go:1281-1285` for the unset-var case).

  **Failure path.** `BuildCommandPlan` → `permissionProfileFromPolicy(..., commandEnv)` → `credentialDenyReadPaths` → SSH config parse → key omitted from `DenyReadIfExists`. The sandboxed `ssh`/`git` child still sees `SSH_KEY_DIR` in its environment and OpenSSH resolves the `IdentityFile` at runtime. The credential baseline does not list the key.

  **Root cause.** Command env is threaded for **where** to read config (`HOME`), but not for **how** config expands path variables. Discovery and OpenSSH disagree on the effective environment.

  **Requested outcome.** Thread the same env slice(s) used for `appendUntrusted` into SSH config expansion so `${VAR}` / `$VAR` (except `${HOME}`/`$HOME`, which should continue to use the discovery home argument) resolve against command env first, then process env, matching OpenSSH’s child-environment semantics. Keep the existing rule that truly unset variables are dropped without inventing paths — just distinguish “unset everywhere” from “set only in command env”. Add a regression test mirroring `TestOpenSSHPathParsingEscapesAndEnv` but with `SSH_KEY_DIR` supplied only via `commandEnv`, not `t.Setenv`.

- [P3] Update manual real-smoke subtest for the new Linux SSH contract
  `internal/sandbox/runner_linux_integration_test.go:93-105`

  **What happens today.** The `"fresh home and non-git workspace launch"` subtest builds an engine with `DefaultPolicy()` and a fresh `HOME`. Elsewhere in this PR, Linux bubblewrap planning refuses profiles that retain absent well-known keys in `SSHDenyReadFiles` unless the policy includes an explicit `~/.ssh` directory deny (`testPolicyWithSSHDirectoryDeny`). With `DefaultPolicy()` and a fresh home, `validateLinuxBwrapPermissionProfile` fails with the selective-SSH refusal before launch.

  **Impact.** Default CI does not set `ZERO_SANDBOX_REAL_SMOKE=1`, so this does not fail automation today. `scripts/sandbox-smoke.sh` does set it; maintainer smoke will fail on this subtest.

  **Requested outcome.** Align the subtest with the rest of the PR’s Linux harness: use `testPolicyWithSSHDirectoryDeny(t, freshHome)` (or assert the expected planning refusal if the subtest’s purpose is to document that contract). No production behavior change required.

## Explicitly out of scope for this review (please do not churn on these)

These were investigated on head `b707a66` and are **not** requested changes. Addressing them in follow-up rounds has caused drift in prior reviews:

- **Degraded execution when discovery is incomplete.** README says incomplete discovery “refuses **sandboxed** execution” (`README.md:274-275`). When the native backend is unavailable, `buildPlatformCommandPlan` returns a degraded direct plan before the `CredentialDiscoveryErrors` gate (`runner.go:229-233`). That matches the existing degraded-fallback contract (`runner_test.go:126-154`) and `linux_helper.go:228` still refuses incomplete discovery on the bubblewrap path. We are **not** asking to block all host execution when sandboxing is already impossible unless product explicitly wants that policy change.

- **Workspace-resident keys dropped from command-env deny via `pathsOutsideOverlappingRoots`.** Intentional per `profile.go:297-299`: command-controlled credentials cannot revoke deliberately granted read/write roots. Keys inside the workspace are already readable through granted roots.

- **Linux `DefaultPolicy()` refusing selective SSH key masks.** Documented product choice (`README.md:277-284`); use explicit `~/.ssh` directory deny on Linux when that tradeoff is acceptable.

- **Unbounded `.ssh` directory paging / walk depth.** Accepted tradeoff after paging landed in `b707a66`; README documents no discovery limit for large/deep trees.

- **Silently dropping `${VAR}` when unset in both environments.** Intentional; tests at `ssh_gpg_deny_test.go:1281-1285`. The P1 finding above is only about vars present in command env but absent from process env.

## Suggested merge strategy for the author

1. Introduce a small internal helper for “env lookup used during discovery” that accepts process env + command env with OpenSSH-like precedence, and use it from `expandSSHConfigPathEnv` (and any future config token expansion).
2. Add the command-env regression test described above.
3. Update the smoke subtest (P3).

euxaristia:
Addressed review feedback in 3fcfe3d4:

- [P1] Resolved OpenSSH path-variable expansion against command-supplied environment in `expandSSHConfigPath()` and `sshDiscoveryEnvValue()`, with child-environment precedence over process environment. Added regression coverage in `TestSSHDiscoveryUsesCommandEnvironment` and `TestSSHConfigEnvironmentPrecedence`.
- [P3] Updated `TestLinuxHelperRealSandboxSmoke` subtest to use `testPolicyWithSSHDirectoryDeny()` matching the Linux selective SSH denial contract.

Vasanthdev2004 (dismissed):
Re-approving at `3fcfe3d4`. The one commit since my last approval makes `IdentityFile` expansion read `$VAR` from the sandboxed command's environment first, last entry wins including an explicit empty value, and only a missing override falls back to the process environment. That matches what OpenSSH would see when the command runs, which is the environment the deny list is protecting, and an unresolved variable still drops the path rather than denying or following a guess. The paged discovery and the content sniff I drove last time are untouched.

`internal/sandbox` green here on Windows, vet clean, CI 9 of 9 at this head.

jatmn (changes requested):
I found six discovery issues that need to be addressed before this is ready. The command-environment plumbing and fresh-home smoke adjustment are present. The remaining failures concern how the new scanner identifies a configured key and how it distinguishes successful discovery from incomplete inspection.

I want this review to give you one coherent revision target. Earlier reviews did not identify this full set together, which has contributed to the back-and-forth. The guidance below groups the verified failures by their shared causes and defines the expected result for each. It does not reopen the capability choices already made on this PR.

## Merge readiness

- **Rebase onto current `main`.** Head `3fcfe3d4` has merge base `1b5db176`, 31 commits behind current target `c1937dfa`. Repository guidance requires a fresh base. Preserve the target's worktree and Windows sandbox fixes when updating. GitHub reports `MERGEABLE`, but the review state is `CHANGES_REQUESTED` and merge status is `BLOCKED`. All nine workflow checks and CodeRabbit succeeded; GitHub lists no required status checks. There is no package-version drift in this comparison. Closed #990 is superseded by this PR; I found no duplicate open SSH/GPG implementation.

## Scope and evidence

These findings are **incomplete fixes of the accepted #815 SSH-key protection scope**, not newly introduced permission grants. Neither merge base `1b5db176` nor current target `c1937dfa` implements this automatic SSH discovery. The responsibility here is to make the new supported discovery paths produce the intended protection, including the new documented refusal when a required input cannot be inspected.

For findings 1–5, I reproduced an omitted key and a successful read through the native Linux helper. Those reproductions use the documented explicit deny of an existing `.ssh` directory: that deny covers the conventional key markers, but cannot protect an undiscovered key outside that directory. The default Linux policy would otherwise refuse selective SSH protection. This prerequisite matters; these are not claims that every default Linux invocation silently launches insecurely. Finding 6 independently demonstrates a required discovery error being lost.

## Findings

### 1. [P1] Bind child inspection to the directory that was enumerated

`internal/sandbox/ssh_key_deny.go:108-109,127-140`

**Trigger and observed result.** The walker opens a directory through `os.OpenRoot`, opens `d` for enumeration, and immediately closes `root`. It then reconstructs each child's absolute pathname for `os.Lstat`, symlink following, and content inspection. Another host process can move the opened directory aside and replace it between enumeration and child inspection. `d` still refers to the original directory, but the subsequent lookups refer to the replacement.

The reproduction starts with a custom-name symlink inside `.ssh` pointing to an unchanged external private key. After the directory entries are read, the original directory is moved aside and replaced with a directory containing a benign file under the same child name. Discovery inspects the benign replacement, returns no discovery error, and omits the external key. Restoring the original directory before profile finalization does not repair the omission. With an explicit `.ssh` deny, native execution can read the external key.

This is a deterministic interleaving using a controlled host mutation. It establishes the code path, not the frequency of an accidental race or an ability for the isolated child to rename the host directory. CodeRabbit's directory-binding request remains applicable.

**Root cause and bounded fix.** The directory handle establishes one object identity, but classification and inspection abandon that identity. The existing file inspector can pin the file it eventually opens; it cannot establish that this file is the child of the directory that was enumerated. Preserve that relationship through child inspection, or report incomplete discovery when it cannot be established. Merely keeping `root` open while continuing the same absolute-path lookups would not establish the relationship. The precise implementation is your choice; this does not require an atomic snapshot of the entire host filesystem.

**Regression coverage.** Use a deterministic synchronization point around enumeration/inspection. Replace the directory with benign same-name children, restore it before planning, and assert that the original external key is protected or execution is refused by the applicable discovery/protection guard. Include an unchanged-directory control and retain legitimate external symlink discovery. A test that only verifies the final file descriptor is regular does not exercise this failure.

### 2. [P1] Resolve relative identity paths from the appropriate working directory

`internal/sandbox/ssh_key_deny.go:441-442`

**Trigger and observed result.** For a user SSH configuration containing:

```sshconfig
IdentityFile keys/work
```

OpenSSH resolves the identity relative to its working directory. Starting Zero in `PROJECT` with that inherited configuration should therefore account for `PROJECT/keys/work`. The new shared path helper instead returns `HOME/.ssh/keys/work`. Relative user-config `Include` paths do use `.ssh` as their base; applying that rule to `IdentityFile` changes the meaning of the configuration.

I verified actual OpenSSH loads the identity from the working directory. A complete permission-profile reproduction using the inherited process baseline, no command-environment override, and an explicit `.ssh` deny omits the actual key and allows the native helper to read it. The incorrect candidate is covered by the `.ssh` deny, so it does not leave a selective-key marker that would otherwise refuse Linux execution.

**Root cause and bounded fix.** The resolver receives `home` and `sshDir` but does not retain the directive's relative-path semantics or the working-directory context needed for identity paths. Make the resolution context explicit enough to distinguish these cases. Resolve the supported identity path against the applicable working directory and keep user-config Include resolution relative to `.ssh`. Preserve the existing distinction between process-trusted credentials and command-controlled roots; this finding does not require revoking deliberate workspace grants supplied through command configuration.

**Regression coverage.** Place different files at `PROJECT/keys/work` and `HOME/.ssh/keys/work`, so a test cannot pass by accidentally finding the wrong one. Compare the identity selected by OpenSSH with the full profile's protected path. Add a paired relative Include case to prove that fixing IdentityFile does not change Include's base. Exercise the existing process/command context plumbing where it selects the applicable directory, with an assertion that established grant filtering still holds.

### 3. [P1] Handle an equals separator after directive whitespace

`internal/sandbox/ssh_key_deny.go:350-358`

**Trigger and observed result.** [OpenSSH accepts an optional equals separator with surrounding whitespace](https://man.openbsd.org/ssh_config), including:

```sshconfig
IdentityFile =/external/key
Include =/external/config
```

`parseSSHDirective` only splits `=` when it is inside the first token. Here the first token is the directive name and the next starts with `=`. The identity is consequently interpreted as a relative path such as `HOME/.ssh/=/external/key`; the Include pattern generally matches nothing. Neither result produces a discovery error. Actual OpenSSH accepts both inputs, and both cases reproduced a native read of the omitted external key with the `.ssh` directory deny.

**Root cause and bounded fix.** Tokenization can separate the directive name from its optional equals separator, but separator handling only examines the first resulting token. The remaining separator is then treated as part of the pathname. Consume the directive separator according to its grammar independently of preceding whitespace, then parse the value. Preserve equals signs that belong to the actual filename; stripping every leading or embedded `=` from arbitrary values would introduce another compatibility problem.

**Regression coverage.** Cover `Keyword value`, `Keyword=value`, `Keyword =value`, `Keyword= value`, and `Keyword = value` for both IdentityFile and Include. Include a real pathname containing `=`. Check the discovered key, not only the parser's token output: the two directives take different downstream paths, so fixing the identity case alone leaves the Include omission.

### 4. [P2] Preserve hashes embedded in SSH filenames

`internal/sandbox/ssh_key_deny.go:402-404`

**Trigger and observed result.** OpenSSH accepts an unquoted embedded hash in this Include filename:

```sshconfig
Include /external/config#work
```

The lexer currently treats every unquoted `#` as the start of a comment. It searches for `/external/config` instead. When that truncated path is absent, the actual included file is never read and its key references disappear without a discovery error. I verified that OpenSSH reads the complete filename and that native sandbox execution can read the key referenced by that file after the `.ssh` directory deny.

The same truncation affects IdentityFile, but the direct identity variant can leave an incorrect external SSH marker and cause Linux to refuse. The Include reproduction is the demonstrated Linux read path; the two effects should not be conflated.

**Root cause and bounded fix.** The tokenizer does not distinguish a comment boundary from a character inside an active pathname token. Preserve the embedded hash according to SSH's token rules while retaining genuine comments, quoting, and escapes. Findings 3 and 4 are good candidates for one coherent lexer correction, with separate regressions for their distinct failures.

**Regression coverage.** Include the unquoted embedded-hash case, a quoted hash, full-line and whitespace-separated comments, and existing escaped-space cases. For the Include case, make the truncated filename absent and put the only reference to the external key in the actual hash-containing file. Verify discovery follows that file and reaches the key.

### 5. [P2] Distinguish an empty environment value from an unset variable

`internal/sandbox/ssh_key_deny.go:486-489,498-505`

**Trigger and observed result.** With `KEY_PREFIX=` explicitly present:

```sshconfig
IdentityFile ${KEY_PREFIX}/external/key
```

OpenSSH substitutes the empty string and resolves `/external/key`. The scanner receives `""` from `sshDiscoveryEnvValue`, treats it as undefined, and drops the entire pathname. Actual OpenSSH loaded the key in this configuration; discovery omitted it without an error, and the native helper could read it after the `.ssh` directory deny.

The current precedence test expects a whole-path drop for an explicit empty override. Its precedence requirement is correct—an empty override must win—but that expected expansion result does not match the resolvable SSH path.

**Root cause and bounded fix.** A string-only lookup conflates variable absence with a present empty value. Preserve presence information through lookup and expansion. An explicit empty command override must not fall back to a nonempty process value; it must contribute an empty string while leaving the rest of the pathname intact. Retain last-entry precedence and the existing handling of variables that are actually unset. This is not a request to revisit unresolved host tokens or the explicitly requested bare-dollar extension.

**Regression coverage.** Cover a nonempty inherited value, an inherited empty value, a command-only value, an explicit empty command override over a nonempty inherited value, duplicate command entries with the last one empty, and a genuinely unset variable. Assert both precedence and the resulting complete path. At least one full-profile case should use a command environment slice rather than only `t.Setenv`, so the test exercises the plumbing this PR just fixed.

### 6. [P2] Propagate failures while enumerating Include globs

`internal/sandbox/ssh_key_deny.go:330-332`

**Trigger and observed result.** An external Include pattern can require reading a directory that the scanner cannot enumerate:

```sshconfig
Include /configs/*.conf
```

`filepath.Glob` intentionally ignores directory I/O errors, including errors encountered during enumeration. Its successful return therefore does not prove that all matching files were inspected. In the reproduction, the Include directory is execute-only (`0111`), a file inside it references an independently readable external private key, and reading the directory returns permission denied. Discovery nevertheless returns `DiscoveryErrors=[]` and omits the key.

This demonstrates a violation of the new incomplete-discovery refusal contract. It does not claim OpenSSH can enumerate that same unreadable directory at the same instant. Restoring directory access later also does not repair an already-built profile.

**Root cause and bounded fix.** The caller treats a convenience glob API's result as evidence of complete enumeration, although the API suppresses the failures the caller needs. Use an enumeration path that can distinguish successful no-match discovery from failed or partial inspection, and feed failures into the existing discovery-error mechanism. Checking the current `Glob` error alone cannot recover I/O errors that it has already discarded. Keep genuinely missing optional Includes valid and preserve the existing config/include bounds.

**Regression coverage.** Compare an absent optional Include, a readable directory with no matches, an unreadable matching directory, and a partial enumeration failure using a controlled fault if necessary. The first two should preserve their current behavior; the latter two must reach the existing execution-refusal path. Check error propagation through the complete profile and helper boundary, not only a local scanner error slice.

## Guidance for one coherent revision

The recurring issue in these six findings is loss of information before the protection decision. The scanner loses which directory it enumerated, which base directory a directive uses, where a separator or comment actually starts, whether an environment value exists, or whether enumeration completed. Once that information is lost, later code sees an ordinary pathname or an empty candidate list. Correct deny enforcement cannot recover a key that discovery never identified.

The branch already has useful pieces for resolving this: structured discovery errors, propagation into the permission profile, and Linux validation that refuses unsupported selective protection. Please complete the inputs to those existing mechanisms. A new backend, a general SSH interpreter, or a replacement policy architecture is not needed to address this review.

I suggest organizing the revision around these three work areas:

| Work area | Findings | Required outcome |
|---|---|---|
| Preserve path interpretation | 2–5 | Supported directives retain their own resolution context; separator/comment parsing preserves filenames; environment lookup preserves presence and precedence. |
| Preserve inspected object identity | 1 | Child classification corresponds to the enumerated directory, or discovery reports that inspection could not be completed. |
| Preserve discovery completeness | 6 | Include enumeration failures reach the existing profile refusal mechanism instead of appearing as a successful empty result. |

For the parser and resolver work, use a small shared matrix of the concrete cases above. Compare supported pathname behavior with actual OpenSSH where practical. `ssh -G` can help check parsing, but it does not by itself prove which file a relative IdentityFile opens; the working-directory case needs a file-loading assertion. Keep the specifically requested bare-dollar extension and intentionally unresolved tokens outside any blanket “match all OpenSSH behavior” requirement.

For each failure class, connect the focused test to the complete protection path:

```text
SSH configuration / directory entry
  → discovered key or explicit discovery failure
  → permission profile after grants and containing-directory filtering
  → backend validation / helper consumption
  → protected key or refused execution
```

On Linux, correct discovery of an external SSH key may now cause planning to refuse because selective protection is unsupported. That is an acceptable result under this PR's documented policy. Do not make a test pass by weakening that refusal. If a test expects successful execution, give it an explicit existing containing-directory deny that actually covers the key, then verify the key cannot be read. A bogus path in the deny list, or a mount argument for the wrong file, is insufficient evidence.

Use deterministic synchronization or faults for the two filesystem cases instead of timing loops. Keep the fixtures isolated from real credentials. Retain companion assertions for the behavior touched by each fix—Include's relative base, literal filename characters, environment precedence, legitimate symlink discovery, and optional missing Includes—so a repair cannot silently break its neighbor.

### Boundaries for this revision

These are preservation constraints, not additional findings or requests to expand the feature:

- Preserve public SSH support-file handling and explicit `allowRead` semantics, including the established distinction between trusted credentials and command-controlled workspace grants.
- Preserve the Linux selective-key/mutable-symlink refusal and the documented explicit containing-directory alternative. Do not restore unsafe file masks or add directory reconstruction as a requirement of this review.
- Preserve the intentionally retained handling of connection-dependent tokens, truly unset variables, and the requested bare-dollar expansion. This review does not ask for connection-aware SSH evaluation or arbitrary SSH command-line discovery.
- Preserve Windows's automatic-discovery exclusion, existing disabled/degraded execution behavior, paged directory traversal with cycle detection, and current config/include limits.
- Keep unrelated token-store/carveout failures out of this revision's required fixes. Preserve existing GPG behavior while changing shared plumbing; no GPG redesign is requested.

Please address the six failures and their paired regression cases together before requesting another review. A concise mapping from each finding to the changed behavior and the test that demonstrates it would make the next review much easier to assess. These acceptance criteria are intended to prevent another cycle of patching individual examples; they are not a guarantee that no future defect can exist.

## Validation and limitations

The sandbox race suite, focused CLI policy checks, sandbox vet, and diff checks passed. Native Linux reproductions establish the key-read paths for findings 1–5; the Include-enumeration reproduction establishes the unreported discovery failure in finding 6. The directory-identity reproduction uses a controlled scheduling point and has the host-mutation prerequisite described above.

Three manual Linux smoke subtests fail on both head and base: head stops at the accepted SSH refusal, while base reaches failing token/carveout assertions. They limit what that smoke run validates; they are not additional findings against this PR. Native macOS and Windows execution were not independently run; their reported CI jobs succeeded.

Vasanthdev2004 (dismissed):
Approving at `2dc8dcbc`. Took each of the six against this head.

Directory binding: every per-entry operation now goes through the open root (`Lstat`, `Readlink`, the sniff through `root.Open`), and the identity check catches a swap between enumeration and inspection. Your regression test skips on Windows, so I drove the same swap here with a junction re-pointed from inside the walk hook: this head reports `directory identity changed during inspection`; with the identity check off and the entry operations back on paths it reports nothing, and the candidate list names a path that by then holds a benign file.

Relative `IdentityFile` resolves against the working directory (`keys/work` lands in the project and the decoy under `~/.ssh/keys/work` is not touched) while a relative `Include` still resolves from `~/.ssh`. One thing to know: in the command-environment pass the process directory comes first in the union of base directories, so a relative path built from a variable that only the command environment sets resolves against the process directory rather than the command directory. Corner of a corner, fine as a follow-up.

The rest checks out as written: `=` after whitespace, `#` starting a comment only at the start of a token (which is what ssh does), an empty but set variable expanding to empty with the last entry winning, and Include enumeration failures reaching `DiscoveryErrors` while an absent or empty directory stays silent.

Reverting each hunk fails its test on the reported reason. CI is 9 of 9 at `2dc8dcbc`.

Vasanthdev2004 (changes requested):
Re-reviewed at `b7bb06b0`, one commit since my approval. The bound itself is right: only trees reached through a symlink are counted, a real subdirectory of one inherits the flag, entries are charged before the page is inspected, and both limits end in a discovery error, so an oversized linked tree refuses the run rather than being walked or quietly truncating the deny list. A real `~/.ssh` stays unbounded.

One thing to fix before this goes in. `TestSSHDiscoverySymlinkDirectoryBounds` calls `t.Fatal` when `os.Symlink` fails, and on a Windows box without Developer Mode or elevation it always fails: `A required privilege is not held by the client.` All three subtests go red here, and with them `go test ./internal/sandbox/`; the rest of the package is green at this head, and the whole of it was green on this machine at `2dc8dcbc`. CI does not see it because the runners hold the privilege. The file next door already handles this the right way, `t.Skipf("symlinks unsupported in this environment: %v", err)` in `TestSSHDiscovery_DirectoryBindingAndReplacement`. The same in these three places and I will re-approve. For the same reason I could not falsify the bound locally, so on that point I am leaning on CI having run it on all three systems.

For later, not for this PR. On Windows the everyday directory link is a junction, and Go reports one as irregular, neither a symlink nor a directory, so the walker drops it at `!mode.IsRegular()` without an error. I put a key one level and two levels behind a junction inside `~/.ssh`, with the same layout in a real subdirectory as the control: the control finds both keys, the junction finds neither, and discovery reports no problem. It costs nothing today because `credentialDenyReadPaths` returns nothing on Windows, but the day Windows enforcement lands that skip becomes a silent miss in a walker whose contract is fail closed. Following it like a symlink or failing on it would both do. I should have probed that on the 15th and did not.

jatmn (dismissed):

## Merge readiness

- [P2] Align Linux real-smoke integration tests with the selective-SSH planning contract
  `internal/sandbox/runner_linux_integration_test.go` (subtests still using `DefaultPolicy()` without `testPolicyWithSSHDirectoryDeny`)
  Main smoke was updated; several `ZERO_SANDBOX_REAL_SMOKE=1` subtests still build plans with `DefaultPolicy()` while Linux bubblewrap now rejects non-empty `sshDenyReadFiles`. They will fail at `BuildCommandPlan` before exercising credential races unless skipped in CI.

- [P3] Document the default Linux native-sandbox operator tradeoff
  `README.md` Safety Model
  README explains that Linux refuses selective SSH-key denies, but not that default-policy wrapped execution requires an explicit containing-directory `denyRead` (what tests use via `testPolicyWithSSHDirectoryDeny`) or a pathname-policy backend. Worth a release-note line that sandboxed `git push` over SSH stops working on macOS by default (per collaborator review on the functional trade).

- [P3] Issue #815 platform scope vs implementation
  Issue text scopes credential denies to Linux; this PR adds Unix discovery and macOS Seatbelt pathname denies. Not blocking if maintainers accept expanded scope, but `#815` closure should reflect macOS/Windows behavior explicitly.

Checks: all required GitHub Actions passing on head `b7bb06b0`. Branch is mergeable against `main` at `99721c762f37cd43ac511007a5f51d1846df959e`.

## Needs maintainer decision

- Default Linux workflow when `sshDenyReadFiles` is populated (always, for well-known absent key paths): refuse wrapped planning vs require operators to add `denyRead: ["~/.ssh"]` vs auto-directory-deny. Implementation and tests consistently choose refusal; confirm this is the intended production experience for Linux bubblewrap users.

## Findings

- [P1] Refuse execution when credential discovery is incomplete, including degraded sandbox fallback
  Attribution: PR-introduced. Merge-base had no `CredentialDiscoveryErrors` and no SSH discovery failure propagation.
  Stated contract: README Safety Model — "Exceeding a config-size or Include limit, or failing to inspect a required input, refuses sandboxed execution rather than using a partial list of protected keys."
  Root cause: incomplete discovery is recorded on the permission profile, but `buildPlatformCommandPlan` returns a direct (unwrapped) plan for `EnforcementDegraded`, `EnforcementDisabled`, `BackendNone`, or `!RequiresPlatformSandbox` before checking `CredentialDiscoveryErrors`, so the same profile can still run without sandbox isolation.
  What fails: on hosts where the native sandbox backend is unavailable or degraded, a command can execute with full host filesystem access even though SSH/GPG discovery already reported an incomplete baseline (config/Include/size/symlink limits, unreadable inputs).
  - `buildPlatformCommandPlan` early return path — skips discovery-error gate
  - `Engine.BuildCommandPlan` / manager planning entry — should share one fail-closed check before any direct plan
  - tests covering degraded/unavailable backend with forced discovery errors (none today; `TestBuildCommandPlanDegradesUnavailableFallback` only uses clean discovery)
  Unchanged on main: degraded fallback behavior without discovery errors.
  Required correction: if `len(profile.FileSystem.CredentialDiscoveryErrors) > 0`, return a planning error for every enforcement level (or at minimum before emitting any `directCommandPlan`), matching the README refusal contract. Add tests: unavailable/degraded backend + intentional discovery failure must not produce a runnable direct plan.
  Author fix: close the root cause on every in-diff row in one pass; do not patch only the early-return branch without tests.
  Out of scope: redesigning degraded sandbox semantics when discovery succeeded.

- [P2] Bound Include directory enumeration before applying the Include match cap
  Attribution: PR-introduced Include glob discovery in `ssh_key_deny.go`.
  Stated contract: README Safety Model — "Exceeding a config-size or Include limit ... refuses sandboxed execution"
  Root cause: `globIncludePaths` calls `Readdirnames(-1)` on each Include parent directory and only applies `sshIncludeMatchCap` after collecting all matches, so a directory with far more entries than the cap can still force unbounded allocation and work during profile construction before discovery fails closed.
  What fails: profile construction can spike memory/latency on large Include parent directories; the cap does not bound work, only the final match count.
  - `globIncludePaths` / `includePaths` — cap-aware enumeration (count matches while reading, or page `ReadDir`)
  - regression test with a parent directory over the cap without allocating proportional to entry count in the success path
  Unchanged on main: no Include glob walk.
  Required correction: stop reading entire directories once `sshIncludeMatchCap` would be exceeded; record `config Include match limit exceeded` (or equivalent) via existing `sshDiscovery.fail` without unbounded `Readdirnames(-1)`.
  Author fix: close enumeration and cap enforcement together with a test.
  Out of scope: changing the cap value or Include depth limits.

euxaristia:

PR #1012, opened 2026-09-05
fix(mcp): forward image tool results on the existing image channel

### LINKED ISSUE #823: MCP tool results silently drop every non-text content block

euxaristia:
Fixes #823

## Summary

Forward MCP image tool-result blocks on the existing `tools.Result.Images` channel to enable multimodal MCP tool results.

## Changes

- Decode MCP image content onto `Content.Data`.
- Convert blocks to `[]zeroruntime.ImageBlock` matching builtin capture tools, capped at `imageinput.MaxImageBytes`.
- Forward images in `registryTool.Run` and update `DroppedContentSummary` accordingly.
- Snapshot provider vision capabilities and route appropriately.

## Test plan

- Unit tests in `internal/mcp/non_text_content_test.go` covering image decoding, text+image mixing, and oversized payloads.
- Pre-existing review discussions from #989 apply.

## Summary by CodeRabbit

* **New Features**
  * Images from MCP tools and user prompts can now reach vision-capable models.
  * Vision support is detected across discovered models, curated catalogs, and additional multimodal model families.
  * Tool images are preserved when switching to models that support image input.

* **Improvements**
  * Increased support for large, multi-image MCP responses.
  * Images are filtered for incompatible models with clearer recovery notices.
  * Image-limit notices now identify whether forwarding, inspection, or size limits were reached.
  * Tool output text, file changes, and session history are preserved more reliably.


jatmn (changes requested):
I found three correctness issues that need to be addressed before this is ready. They concern the ACP capability path added by this PR: where its evidence comes from, when it is resolved, and what happens when an accepted attachment is refused.

Please address these together as one image-delivery contract. The goal is a bounded correction to the existing wiring, with tests that exercise the production boundaries described below.

## Merge readiness

Reviewed head `131e16850c241467d303560fc157ddfcd9b7bde0` includes current `main`, `1b5db1765672820caac1684b168c9898b5ba3593`. GitHub reports no merge conflicts, and all reported checks pass for this head. Merging is blocked with `CHANGES_REQUESTED`. The earlier stale-base and missing-CI concerns from #989 no longer apply.

## Why this has required repeated feedback

The basic MCP forwarding work has concrete value: it lets an MCP screenshot reach the existing `tools.Result.Images` channel instead of disappearing into a text-only result. Several previously reported problems are addressed on this head: aggregate retained bytes are bounded, conversion uses one pass, an exactly-at-limit padded image is accepted, image-only output is nonempty and uses neutral wording, and validated budget skips are distinguished from content intentionally left uninspected. The TUI now snapshots discovery before returning its asynchronous command and updates the discovery map by replacement.

The remaining problems are concentrated in the integration between ACP and the shared vision gate. Adding a callback is only sufficient if its real producer supplies the fields it reads, it has an appropriate lifetime, and refusal is visible at the surface that accepted the user's image. Those conditions are currently tested separately—or bypassed by the test setup.

For example, `TestACPWiresSupportsVision` injects already-populated `InputModalities` and sends a text-only prompt. That proves the callback recognizes the supplied values. It does not prove that production discovery produces those values, that repeated image results reuse one decision, or that an actual user attachment receives an observable refusal. Consequently the test can pass while all three issues below remain.

The earlier feedback has followed that pattern: fixing an individual branch has left the next producer/consumer boundary untested. The way to close this revision is to trace and test the complete ACP path once:

`resolved provider/model → actual capability source → stable run lookup → initial/tool image decision → delivered image or observable refusal`.

These are three independently actionable defects within that path. A notice alone does not restore supported images; correct metadata alone does not stop repeated requests; and caching the current metadata does not restore the fields the parser discarded.

## Findings

### [P2] Make ACP's initial-image refusal observable

`internal/acp/agent.go:256-258`

**Failure path.** `handleSessionPrompt` accepts the prompt and decodes its image blocks through `promptImages`. `runTurn` then resolves the selected model, evaluates the new capability predicate, and assigns `images = nil` when it returns false. It continues with the original user text and ordinary agent options. The notifier already exists, but this branch does not use it; it also does not add an omission explanation to the model's prompt or return a refusal error.

For a request such as “describe this screenshot,” the reproduced result is zero images passed to the agent, unchanged prompt text, no omission notification, and a normal `end_turn` response. ACP continues advertising image prompt support. The client therefore has no reliable indication that Zero removed an input on which the request depends.

**Attribution.** Before this PR, ACP passed the decoded attachment through to `Options.Images`. The new removal branch introduces the silent omission. The same attachment request retains its image on the merge base/current target. This is the ACP refusal-notice request from #989, verified against the current implementation.

**Required outcome.** Preserve the conservative capability gate, but make its refusal observable to the ACP client, or explicitly reject the unsupported attachment before generating an answer. If the implementation continues with text, an explanatory model-visible note can also prevent the model from treating the missing screenshot as something the user forgot to provide. The important correction is that omission must not look like successful acceptance of the complete request.

**Regression coverage.** Send a real ACP text-plus-image prompt through the handler with an unsupported model. Assert the image is withheld and the chosen notification/refusal is emitted. Retain a supported-image case to show that the correction does not discard a valid attachment or replace its text. Calling `SupportsVision` directly cannot exercise this defect.

### [P2] Remove live discovery from ACP's per-image delivery predicate

`internal/acp/agent.go:785-786`

**Failure path.** For a nonempty model ID, the new predicate calls `a.deps.DiscoverModels` before considering the curated fallback. Production installs that hook in `internal/cli/acp.go` as `defaultDiscoverProviderModels`, which calls `providermodeldiscovery.Discover`. That is a live HTTP probe, with no result cache in this path.

The shared agent loop constructs a carrier for each image-bearing tool result and evaluates `SupportsVision` for each carrier. ACP therefore turns repeated boolean checks into repeated network requests, including when the model is already catalogued. Three checks against an unchanged route produced three HTTP requests. A provider with a working completion endpoint but a stalled `/models` endpoint can add the discovery client's ten-second timeout for each request, sequentially, before the next completion. An initial user attachment invokes the predicate as well. Existing built-in image-producing tools are sufficient to reach this path; this does not depend on adding external MCP registration to ACP.

**Attribution.** ACP already discovered model choices during session setup, but it did not perform discovery at this image-delivery boundary on the base. The new callback adds that work to the active turn. This is the unresolved lookup-lifetime issue raised in #989.

**Required outcome.** Resolve the needed evidence once for the run's effective route, then reuse a stable lookup for initial attachments and tool-produced images. Eager resolution or lazy resolution with a bounded, memoized result can both satisfy this; text-only runs need not acquire a new discovery requirement. A failed optional discovery attempt should use the existing fallback without causing every subsequent image result to repeat the same failed request.

**Regression coverage.** Exercise an unchanged route with several image-bearing results and count discovery calls at the production-facing hook. Separate any session-choice lookup from the count attributable to the run. Assert that image checks do not issue additional requests once the run has resolved its evidence. Include a failing discovery response to prove fallback does not become repeated per-image network work. Use a controlled failure or request counter rather than waiting through real ten-second delays.

### [P2] Feed ACP a production source that actually carries image modalities

`internal/acp/agent.go:789-797`

**Failure path.** This branch requires `dm.InputModalities` to recognize an explicit discovered capability. ACP's production hook, however, calls the live-only `providermodeldiscovery.Discover` path. Its `modelsResponseItem` has no input-modality field, and `parseModelsResponse` does not populate `InputModalities` in the `Model` values it returns. The callback's affirmative discovered-vision branch cannot be reached through that production source.

This is observable with a supported response schema. An OpenRouter-format model response containing `architecture.input_modalities: ["text", "image"]` for a non-curated model produces `[text image]` through the repository's catalog parser. The same response through ACP's production discovery parser produces an empty modality list. For an ID outside the name heuristic, the new resolver consequently returns false. The initial-image gate discards the accepted attachment, and the shared tool-image gate suppresses image results.

The current test bypasses this exact failure by constructing `providermodeldiscovery.Model` values with `InputModalities` already filled in. More assertions against those synthetic values will not establish that the actual producer works.

**Attribution.** The parser's omission itself predates this PR. The new mandatory image predicate makes it consequential for previously delivered ACP input and built-in tool images. That changed dependency is why this belongs to this review; it is not a request to repair every existing model-discovery limitation. The demonstrated response contains the relevant metadata, so this is also distinct from the intentional conservative fallback when capability evidence is genuinely absent.

**Required outcome.** Preserve relevant supported modality evidence through the actual production path used by ACP, or supply an existing modality-bearing source through a suitably bounded capability lookup. Keep explicit text-only evidence authoritative and keep the existing fallback for absent metadata. The exact implementation is open.

In particular, do not blindly replace the general model-list hook with a catalog-filtering operation. `defaultDiscoverProviderModels` intentionally returns a custom provider's full model list without a catalog merge or coding-model filter. Preserve that listing behavior. Passing through optional capability fields, or providing capability evidence separately without changing the model choices returned to ACP, are examples of narrower approaches.

**Regression coverage.** Feed a supported JSON response through the same parsing/discovery path production ACP uses. Use a non-curated ID that cannot pass by name alone, then assert that its image capability reaches the ACP decision and that the supported image is delivered. Include explicit text-only and absent-modality cases to preserve precedence and fallback semantics. A full external service is unnecessary; a controlled HTTP response using the real parser exercises the missing boundary.

## Root-cause guidance for the fix

The shared issue is that ACP currently uses a network operation with incomplete output as though it were a complete, inexpensive capability predicate. Initial-image refusal then hides the consequence from the client. Resolve the source, lifetime, and outcome together in the following order:

1. **Establish the actual evidence source.** Trace the resolved provider profile through the production hook and parser. Verify which supported response fields become capability evidence. Preserve the provider's full model-list behavior and use the same credential/profile identity as the active route.
2. **Give that evidence a bounded lifetime.** Create or lazily resolve a small capability view for the current run, and close `SupportsVision` over it. It must distinguish an explicit text-only declaration from absent metadata before applying the existing fallback. This can be an internal helper and local state; no new public type or framework is required.
3. **Use the same evidence for both image sources.** Initial ACP attachments and tool-produced images should agree for the same provider/model. Resolve against the current selected model when a new run starts. ACP does not need new mid-run model-switch infrastructure to fix these findings.
4. **Make the final disposition honest.** Supported images should reach the existing image channel; unsupported initial attachments should produce an observable refusal. Preserve the existing tool-result omission notice and ordinary text output.
5. **Validate the connected path before considering the fix finished.** Prove the real producer, the callback lifetime, and the delivered result together. Then run the existing regression tests that protect the already-correct MCP/TUI behavior.

The implementation need not centralize every surface into one new abstraction. The necessary invariant is agreement within the ACP run using actual, provider-qualified evidence. A small change to production parsing/wiring, bounded lookup state, and refusal reporting can address the three findings without a wider refactor.

## Bounded verification plan

This matrix targets the reported failure paths rather than expanding the feature:

| Case | Evidence the test should establish |
| --- | --- |
| Non-curated vision ID with supported modality fields in the actual JSON response | Real production parsing preserves the fields; ACP accepts the image without relying on the name heuristic. |
| Explicit text-only model with a user attachment | Image bytes are withheld and the ACP client receives the chosen refusal/notification. |
| Missing modality metadata or optional discovery failure | Existing fallback is retained; genuinely unknown routes remain conservative. |
| Several image-producing results in one unchanged run | Once the run's evidence is resolved, further image checks make no additional discovery requests. |
| Equivalent initial and tool-produced image inputs | Both use the same capability evidence; accepted bytes reach the agent/provider-facing message and refused content has the appropriate explanation. |
| Text-only prompt/result | Ordinary text behavior is preserved. |

Keep the existing tests for image-only and mixed MCP results, exact/over-limit decoded sizes, aggregate-budget notices, zero-budget no-decode behavior, contiguous tool results, and same-turn image/model-switch ordering. Retain the TUI race checks for the snapshot/map-replacement work. These are preservation checks, not additional findings.

For each newly added regression, verify that it fails on the unfixed implementation for the stated reason and passes with the correction. Tests should assert observable results and request counts, rather than the shape of a preferred helper. Run the repository-required checks on the final remediation head so the green CI state corresponds to the complete fix.

## Scope of this request

The three findings are the code changes requested by this review. Their shared fix should preserve conservative unknown-model behavior, explicit text-only rejection, full model listings, current provider/credential selection, text output, and the existing bounded image channel.

This review does not require transport-envelope parity, larger response limits, a new discovery service, a persistent/global capability cache, a generalized provider refactor, new ACP switching support, additional image formats, or durable image-history storage. Those would be separate decisions. The closure target here is a working ACP capability source, a stable decision for the run, and observable handling of refused input, backed by tests that cross the actual production boundaries.

jatmn (dismissed):
@Vasanthdev2004 lgtm

Vasanthdev2004 (changes requested):
The routing decision is right, and the comment explaining it is the best part of the diff: images ride a following user message because every provider drops them on a tool-role message, and the one-tool-result-per-tool-call pairing has to stay intact. Deferring the vision check until after a mid-turn model switch resolves is a nice touch.

Two things before this goes in. Neither is deep.

## Tool images stop working for local multimodal models that worked before

Base sent tool images unconditionally. Head gates them on `modelAcceptsToolImages` and substitutes a text note when it says no. For a model the gate cannot identify, that is a working feature turning into a sentence.

The gate resolves in three steps: discovered modalities, then the curated catalog, then the name heuristic. The heuristic is unchanged by this PR, and I checked it on both trees:

```
gemma-3-27b-it        false        qwen2.5-vl-7b   true
gemma3:27b            false        pixtral-12b     true
llama-4-scout         false        llava:13b       true
llama4:17b            false        gpt-4o          true
mistral-small-3.1     false        gemini-2.5-pro  true
phi-4-multimodal      false        claude-sonnet-5 true
```

The first column is all genuinely multimodal. Discovery does not rescue them either: the new parsing reads `modalities.input` and `architecture.input_modalities`, which is the OpenRouter and newer-OpenAI shape. Ollama's `/v1/models` returns neither, so an Ollama user on `gemma3` or `llama4` falls through to the heuristic and gets:

```
Image output from <tool> was not sent because the current model does not support image input.
```

for a model that accepts images fine today.

Credit where it is due: the message is honest and not silent, and OpenRouter users are covered by the new modality parsing. But `view_image` and `capture` regress for local multimodal setups, which is a good chunk of the audience for those tools.

Adding those four families to `VisionCapableByName` looks like it covers it, and matches how the existing entries are written.

## A server controls how many image blocks arrive, and nothing bounds the count

`forwardImages` bounds total decoded BYTES against `MaxImageBytes` and stops at `remaining == 0`, but a block only has to fit in what is left. Small images therefore accumulate without limit:

```
MaxImageBytes = 10485760
one 1x1 GIF   = 42 bytes decoded

server sent   1000 blocks -> forwarded   1000 blocks
server sent  50000 blocks -> forwarded  50000 blocks,  heap +4.6 MiB
```

The budget permits about 249 000 of them, each becoming an image block on the outgoing provider request. That request is going to be rejected, after Zero has built it.

A count cap next to the byte cap fixes it. The `dispUninspected` disposition already exists for exactly this reporting, so the note machinery is there.

## Checked and clear

I went looking for a concurrent map access on the discovered-model map and did not find one. The snapshot in `runAgentWithOptions` is taken on the UI goroutine before the `tea.Cmd` closure is returned, and it deep-copies both the slices and each `InputModalities`. That is the right shape, and it is the thing that would have been easy to get wrong. `go test -race ./internal/tui/` is clean.

`internal/mcp` and `internal/providermodeldiscovery` are green. `internal/agent` fails only `TestEagerToolSchemaTokenBudget` and `internal/tui` only `TestAltScreenTranscriptScrollKeepsFooterFixed`; both fail identically on `origin/main`, so neither is yours.

Vasanthdev2004 (dismissed):
Both fixed, and neither by overcorrecting, which is what I checked for.

The vision gate now recognises the families it was missing, and still says no to text-only models:

```
gemma-3-27b-it     true       deepseek-r1     false
gemma3:27b         true       qwen2.5-coder   false
llama-4-scout      true       codellama       false
llama4:17b         true
mistral-small-3.1  true
phi-4-multimodal   true
```

The right column is the half that matters. Making the heuristic answer true more often is easy; keeping it discriminating while doing so is the part that could have gone wrong.

The image count is bounded:

```
server sent      1 -> forwarded   1
server sent      5 -> forwarded   5
server sent     50 -> forwarded  16
server sent  50000 -> forwarded  16
```

Small results pass through untouched and a flood is capped, which is the shape I would want.

`internal/mcp` and `internal/modelregistry` green. `internal/agent` fails only `TestEagerToolSchemaTokenBudget`, which fails identically on `origin/main` and is Windows-only.

jatmn (changes requested):
I found one hardening issue that should be addressed before merge, plus one small messaging alignment. I do **not** think further architectural changes are needed to close this PR.

## Merge readiness

- Head `cee61510` is mergeable with current `main` (`aadb4a27`); no three-dot conflict. All CI smoke/security checks pass.
- GitHub still shows `mergeStateStatus: BLOCKED` because CodeRabbit has an open `CHANGES_REQUESTED` on the decode-budget point below. Vasanthdev approved the latest head after validating the vision-heuristic and forwarded-image count cap.
- This branch supersedes closed #989. The MCP forwarding core, aggregate byte cap, single-pass disposition, ACP refusal notification, per-run vision memoization, modality parsing, TUI discovery snapshot, and local-vision heuristic fixes from prior rounds look addressed on this head.

## Why this PR has seen so much review churn (and how to close it)

This feature is not one change — it is a **chain of contracts** that must agree end to end:

`MCP bytes on the wire → decode/disposition in mcp → tools.Result.Images → agent carrier message → provider request`, with a parallel **`effective model supports vision?`** decision that must use consistent evidence in ACP, TUI, and CLI.

Each round of feedback has tended to fix **one link** in that chain while the next review pass exercised the **next boundary**. That is why it has felt endless even though the work has real value:

1. **Decode budget ≠ forward budget.** Bounding retained bytes and forwarded count does not automatically bound inspection work. `forwardImages` still fully decodes candidates on the `dispBudgetExceeded` path while `remaining` stays unchanged, so a hostile server can force unbounded decode CPU/memory even when almost nothing is forwarded. The count cap (`maxForwardedImages = 16`) limits what reaches `Result.Images`, not how many payloads get decoded. This is the same class CodeRabbit flagged; it is the main item still open.

2. **MCP tool text ≠ agent delivery message.** `registryTool.Run` builds explanatory `Output` (budget exceeded, uninspected, cannot forward). The agent loop builds a separate user-role carrier or drop notice. Those layers can disagree when two gates apply (budget skip + non-vision model). That is a wording alignment issue, not a missing feature.

3. **Happy-path tests ≠ full lifecycle.** Most new tests prove the intended path (forward, gate on non-vision, escalate then forward, count cap). That is good, but it left inspection/decode limits under-specified. Tests that count decode attempts (`TestImagePayloadsAreDecodedOnceAndNotPastTheBudget`) correctly prove no decode after `remaining == 0`, but they do not cover the `remaining > 0` + `dispBudgetExceeded` flood case.

4. **Surface-specific vision authority is intentional, not a bug.** CLI/exec uses registry + name heuristic only; ACP/TUI use discovery first. That matches how exec already gated prompt images before this PR. Do not “fix” CLI by adding live discovery unless product explicitly wants that — it would expand scope.

5. **Early-exit tool paths are not a remaining defect.** Abort/stop/guard branches append image messages before the model-switch block, but **merge-base already did not run model switch on those paths either**. Adding vision gating there is conservative (drop on non-vision current model), not a regression against escalation that cannot apply once the turn terminates. I am **not** asking for changes on early-exit ordering.

**Closure target for this revision:** fix the inspection/decode bound (one bounded change in `forwardImages`), optionally align budget-recovery wording when vision will block delivery anyway, and add tests that fail on the unfixed decode flood. No new surfaces, no CLI discovery refactor, no early-exit restructuring.

---

## Findings

### [P2] Bound MCP inspection/decodes independently of forwarded images

`internal/mcp/client.go:575-603` (`forwardImages`), `imageBlockFromContent` at ~660

**What happens today**

`forwardImages` walks MCP content in one pass and tracks three different limits:

- **Per-image decode cap:** `imageinput.MaxImageBytes` inside `imageBlockFromContent`
- **Aggregate forward byte cap:** `remaining`, decremented only when an image is actually appended to `images`
- **Forward count cap:** `maxForwardedImages` (16)

When `remaining == 0` or the count cap is hit, later candidates are marked `dispUninspected` **without decoding** — this part is correct and tested (`TestImageCountBudgetPreservesTextAndSkipsFurtherDecoding`).

When a candidate decodes successfully but `len(image.Data) > remaining`, it is marked `dispBudgetExceeded` **after a full decode**, and `remaining` is **not** reduced. The loop can repeat that for every remaining block in the array.

**Concrete failure path**

1. MCP server returns one 8 MiB image (forwarded; `remaining` becomes 2 MiB) plus thousands of valid 6 MiB images.
2. Each 6 MiB image is fully base64-decoded (~6 MiB heap per attempt), then rejected as `dispBudgetExceeded` because 6 MiB > 2 MiB `remaining`.
3. `remaining` stays 2 MiB, so the condition at line 585 never trips; decoding continues for the entire content array.
4. User sees at most 16 forwarded images and correct drop notes, but Zero may spend seconds and gigabytes of transient allocation on a single tool call.

This is reproducible on head; it is absent on merge-base because `forwardImages` did not exist. `TestImageBudgetNonZeroResidueAllowsSmallerLaterImage` intentionally depends on decoding a budget-exceeded candidate so a smaller later image can still fit — keep that behavior, but cap how many such inspections are allowed.

**Root cause**

The disposition model conflates **“inspected and rejected for byte budget”** with **“never inspected because budgets were exhausted.”** Forward limits were added (bytes, count) without a matching **inspection budget** for decode work that does not result in a forward. The count cap solved forwarded-image floods but not decode floods on the `dispBudgetExceeded` branch.

**Required outcome (bounded — no drift)**

Add a separate inspection bound in `forwardImages`, independent of `len(images)`:

- Stop calling `imageBlockFromContent` once an inspection limit is reached (suggested: a small fixed count such as 32, or `maxForwardedImages + N` — pick one constant and document it next to `maxForwardedImages`).
- Mark further image candidates `dispUninspected` without decoding.
- Preserve: single-pass disposition, aggregate byte cap, forward count cap, residue behavior (decode-then-reject only within the inspection budget so a smaller later image can still fit), and existing note text for `dispUninspected` vs `dispBudgetExceeded`.

Do **not** change SSE limits, vision gating, ACP/TUI wiring, or provider request shaping for this fix.

**Regression tests (must fail before fix, pass after)**

1. **Decode flood:** many valid sub-cap images where each decoded size exceeds current `remaining` after one forward; assert decode attempts ≤ inspection cap and disposition includes `dispUninspected` for skipped tail blocks.
2. **Residue preserved:** keep `TestImageBudgetNonZeroResidueAllowsSmallerLaterImage` passing (8 MiB + 3 MiB exceeded + 1 MiB forwarded).
3. **Count cap preserved:** keep `TestImageCountBudgetPreservesTextAndSkipsFurtherDecoding` passing.

---

### [P3] When vision blocks delivery, do not promise budget-only recovery in MCP tool text

`internal/mcp/registry.go:357-367` with `internal/agent/loop.go:3496-3500`

**What happens today**

For a non-vision effective model, the agent still receives `tools.Result.Images` from MCP and then `toolResultImageMessage` emits a user notice: images were not sent because the model does not support image input. Separately, MCP `registryTool.Run` may already have appended budget guidance to the tool-role `Output`, e.g. “exceeded remaining image budget… Retrying with fewer images can recover this payload.”

The model can therefore read, in one turn:

- tool `Output` suggesting fewer images may recover budget-skipped payloads, and
- a user message saying images were withheld because the model is not vision-capable.

**Root cause**

Drop reasons are decided in two layers (MCP disposition notes vs agent delivery gate) with no shared knowledge of which gate ultimately blocks delivery. Budget notes are always appended when `dispBudgetExceeded` blocks exist, even when vision would prevent forwarding regardless.

**Required outcome (bounded — no drift)**

Pick the smallest alignment that removes contradiction; either is fine:

- **Option A (MCP-side):** when building budget-exceeded notes, avoid unconditional “retry with fewer images can recover” if the tool result’s images will not be delivered to a non-vision model. (This may require passing vision context into the registry path — only do this if it stays a narrow check, not a new framework.)
- **Option B (agent-side, likely smaller):** when emitting the vision drop notice, reference that budget-skipped images were also not delivered and that switching to a vision-capable model is required before retry semantics apply.

Do **not** change vision gating policy, discovery wiring, or budget byte/count limits. This is messaging consistency only.

**Test**

One case: non-vision model, MCP result with one forwarded image block plus one `dispBudgetExceeded` block; assert the model-visible messages do not simultaneously promise budget-only recovery and cite vision as the sole blocker without qualification.

---

## Explicit non-requests (to prevent further drift)

Please **do not** expand scope in response to this review:

- CLI/exec live discovery for vision (pre-existing registry-only policy for prompt images; extending to tool images is consistent).
- Early-exit tool path restructuring or model switch before abort/stop/guard.
- MCP `IsError` image suppression, image redaction, ACP prompt size validation, or SSE limit changes.
- Further vision-heuristic expansion beyond what Vasanthdev already validated.
- New abstractions that unify ACP/TUI/CLI capability resolution in one framework.

If the inspection bound and optional messaging alignment above are addressed with the tests described, I do not see another merge-blocking defect on this head.

euxaristia:
Addressed review feedback in f04c6d6a:

- [P2] Added `maxInspectedImages = 32` candidate bound in `forwardImages()` (`internal/mcp/client.go`) so byte-budget-exceeded payloads do not cause unbounded decoding. Added decode-flood regression coverage in `TestImageInspectionBudgetBoundsRejectedCandidates`.
- [P3] Aligned `toolResultImageMessage()` (`internal/agent/loop.go`) to clarify that switching to a vision-capable model is required before retrying delivery for budget-skipped images. Covered in `TestNonVisionImageNoticeQualifiesBudgetRecovery`.

Vasanthdev2004 (dismissed):
Re-approving at `f04c6d6a`. The one commit since my last approval adds the bound the count cap could not give: candidates that decode but do not fit the remaining budget consumed no budget, so a server could make the client decode an unbounded number of them. `maxInspectedImages` caps the decodes at 32 independently of how many are forwarded, the test drives it through the decoder seam and checks the dispositions on both sides of the line, and trailing text still comes through. The non-vision notice now tells the model to switch models before retrying, which is the right instruction when the budget message would otherwise invite a retry that cannot succeed.

`internal/mcp` is green here; `internal/agent` fails only the eager schema budget test, whose accounting main changed in #1017 after this branch's base. The branch conflicts with main and has no CI run at this head because of it, so it needs the merge before it can go in.

jatmn (changes requested):
I found one capability-resolution regression that needs to be addressed before this is ready. The inspection-count bound and qualified non-vision retry notice are addressed on `f04c6d6a`. The remaining code finding concerns the identity of the endpoint whose capability data is trusted. The rebase and CI requirements below are separate merge-readiness work.

## Merge readiness

- **[P1] Rebase onto current `main` and preserve the upstream fixes.**
  Head `f04c6d6a` has merge base `1b5db176`; current `main` is `c1937dfa`, 31 commits ahead of that base. GitHub reports conflicts, confirmed in `internal/acp/agent.go`, `internal/acp/agent_test.go`, `internal/mcp/client.go`, `internal/mcp/network_client_test.go`, and `internal/tui/model.go`. Preserve the MCP request/response hardening from #942, ACP MCP workspace/cleanup support from #1003, and ACP session changes from #914 when resolving them. In particular, reconcile the capability fix with the current TUI run setup and model-switch behavior in `main`; mechanically choosing either side of those conflicts can discard working behavior. The resolved diff needs review against the updated target.

- **Run project CI on the rebased head.**
  This exact head has no Actions runs or check runs; its only status is CodeRabbit. Focused MCP/agent/ACP/discovery/registry tests, provider image and CLI tests, affected race checks, scoped vet, formatting, and release build/smoke pass. Those results do not replace the required project/platform checks on the conflict resolution. Include the focused regression coverage below in validation of the final head.

## Findings

- **[P2] Qualify authoritative discovery by the active endpoint**
  `internal/tui/image_attach.go:122–125`; also `internal/tui/model.go:5527–5530`

  The new checks prefer discovered modalities over the built-in curated registry, but the discovery lookup identifies a provider descriptor rather than the configured endpoint that supplied the answer. Two distinct saved OpenAI-compatible endpoints can both resolve to `custom-openai-compatible`. The descriptor therefore cannot establish that a discovery result belongs to the active endpoint.

  A concrete failure sequence is:

  1. Save custom endpoint A before custom endpoint B. Both expose the model ID `gpt-4.1`; A reports input modalities `[text]`, while B supports `[text,image]`.
  2. Make B the active profile and open the model picker.
  3. `modelPickerDiscoveryCmds` deduplicates saved profiles by descriptor ID, so it dispatches discovery for A and skips B. `modelPickerProviderDiscoveryCmd` returns a message carrying that descriptor ID, and `applyModelPickerModelsDiscovered` publishes A's models into the shared descriptor entry.
  4. B's synchronous image gate looks up that same entry and treats A's `[text]` answer as authoritative. B's valid attachments are refused. A subsequent run snapshots the same entry, so its tool-image capability callback also reports no vision support.

  The prompt-image reproduction accepts images on both merge base `1b5db176` and current `main` `c1937dfa`, but rejects them on head `f04c6d6a`. The tool-image consequence follows from the matching lookup in the run callback. The descriptor collision existed before this PR; moving its answer ahead of the curated registry makes it newly harmful for a known model such as `gpt-4.1`. This is the causal regression being reported.

  **Root cause and correction:** capability evidence loses its originating endpoint identity before it reaches the gate. Filtering by the active descriptor removes cross-descriptor contamination but leaves two endpoints sharing a descriptor indistinguishable. Deep-copying the map protects a running command from later mutation; it faithfully copies the wrong endpoint's answer too.

  Please preserve enough originating profile/endpoint identity to establish that authoritative evidence belongs to the active route, and apply that rule to both the synchronous gate and the run snapshot. Carrying provenance with discovery results or maintaining an appropriately scoped capability record are possible approaches; the required outcome is that one configured endpoint's discovery cannot determine another endpoint's image support. When matching evidence is unavailable, retain the existing fallback behavior. When genuine active-endpoint evidence is available, retain its intended precedence. Keep the run's decision isolated from subsequent UI updates.

  Changing only the resolver's lookup order would not address the identity loss. Changing only which saved profile is queried first would leave publication and lookup unable to distinguish origins. Trace the correction through discovery scheduling, result delivery, publication, active-profile selection, and both consumers so it fixes this failure path as a whole.

  **Focused regression coverage:** exercise discovery result delivery and publication, rather than only injecting an already-correct map into the resolver. Use two distinct endpoints with the same descriptor and model ID but opposite modality answers. Verify that A's result cannot override active B, that B's own explicit allow/deny is honored, and that the synchronous gate and run callback agree. Cover reversed saved-profile order and a delayed discovery result arriving after a profile switch as validation of the same ownership rule. Preserve the existing snapshot isolation and verify that a later UI discovery update cannot change an already-created run's capability decision. These are companion cases for this one fix, not additional feature requests.

## Guidance for closing this PR

There is one verified code finding in this review. Several earlier corrections are already present, including bounded image inspection, observable ACP refusal, per-run ACP discovery memoization, modality parsing, and TUI snapshot isolation. The remaining issue shows why tests at the resolver alone are insufficient: a resolver can correctly prefer its supplied discovery record while the production discovery path supplies a record from the wrong endpoint. The useful next step is to pin the identity contract across the producer and both consumers, with a regression test that exercises that path.

Please keep the correction bounded to that contract. This review does not require a general capability framework, a model-picker redesign, CLI live discovery, or a change to precedence between raw endpoint metadata and the remote model catalog inside `DiscoverCatalog`. That last policy is distinct from the endpoint-identity regression and is not an additional finding. Existing image budgets, transport limits, and accepted fallback behavior should remain intact.

The closure target is one coherent follow-up: correct endpoint ownership in both TUI capability consumers, add the focused production-path regression coverage, resolve the rebase while preserving upstream behavior, and validate the resulting head. Include a short explanation of where discovery origin is retained and how both consumers verify it. That will make the correction reviewable as a complete behavior change and reduce another round of isolated fixes. A different implementation is welcome if it establishes the same contract without expanding scope.

Vasanthdev2004 (dismissed):
Approving at `34c10af1`, with one small thing to put back before merge.

Rebase: I compared the branch with an automatic merge of `f04c6d6a` onto main. Every file both sides touched without conflict is identical, and in the conflicted files the net change matches the old one apart from context, so #942, #1003 and #914 are intact. The one loss is in `runAgentWithOptions` in `internal/tui/model.go`: the twelve-line comment above `if m.allowEscalation` from #1019 (the one explaining why the switchers are built from the active profile per turn) went missing in the resolution while the code stayed. Please restore it. Also, `a6148635` and `c502e7c2` still carry conflict markers in `internal/tui/model.go` and `internal/acp/agent_test.go`, so the branch does not build at every commit; harmless under a squash, worth knowing if it lands as a merge commit.

Endpoint identity: discovery is published under `providerEndpointKey`, and the generic descriptor ID is no longer written for `custom-*` profiles, so a second endpoint on the same descriptor cannot be read through the shared key. I drove both endpoints through `applyModelPickerModelsDiscovered` in both orders, switched the active profile each way, and read the answer through `modelSupportsVisionFor` and through the run options captured by `captureRunSupportsVision`. A result for the other endpoint arriving after a switch leaves the active one on the catalog answer, and a run keeps the snapshot it started with when a later publication replaces the map. Restoring descriptor-keyed publication and lookup fails the three scoping tests and my reversed-order probe on the expected lines.

Left as it was: `modelContextWindow` and the `/model` candidate match still read the map by descriptor ID, so for custom endpoints they now fall through to the any-provider scan and the catalog. Same shape as before this change, not a blocker; a follow-up could route them through `activeDiscoveryKeys` too. CI is 9 of 9 at `34c10af1`.

Vasanthdev2004 (approved):
Re-approving at `05a584fe`, two commits since the last one.

The comment above `if m.allowEscalation` is back and matches main byte for byte, thank you. `DroppedContentSummary` and `ImageBlocks` moving into `export_test.go` is fine: nothing outside the package tests called them, and the build is clean without them.

The limiting cause is now reported for what it was. `forwardImages` records the first limit it hit, budget, forwarded count or inspection count, and the note names that one instead of always blaming the byte budget. `TestMCPResultUninspectedImageLimitingReasonNotes` drives both new causes through `registryTool.Run`, which is the path that writes the note; forcing the recorded reason back to none fails both subtests on their named message, and the budget wording stays covered by the existing tests.

Package green, vet and gofmt clean, no overlap with the three commits main has gained since your base, CI 9 of 9 at the head.

jatmn (dismissed):
I found two correctness issues that need to be addressed before this is ready. The decoding limits, truthful omission notices, ACP memoization/history correction, and TUI snapshot isolation are addressed on `05a584fe`.

## Merge readiness

- **Rebase onto current `main` and validate the resulting head.** The merge base is `4ba112a0`; current `main` is `99721c76`, three commits ahead, including the 0.9.0 release changelog, test-helper refactor, and OpenCode session headers. There is no changed-file overlap and GitHub reports mergeable, but the repository's fresh-base requirement applies. Preserve those upstream changes and recheck the resolved diff.
- All ten reported checks/statuses pass on `05a584fe`, including unit tests, race detection, security, and Linux/macOS/Windows smoke. GitHub remains `BLOCKED` with `CHANGES_REQUESTED`.

## Findings

### [P2] Preserve endpoint ownership through the complete discovery-key migration

`internal/tui/picker.go:994–995`, with `providerEndpointKey` at 819–825 and `activeDiscoveryKeys` at 863–867.

**Stated contract:** my earlier request was that “one configured endpoint's discovery cannot determine another endpoint's image support.” When matching evidence is unavailable, the existing fallback must remain available.

**Root cause:** the new per-endpoint discovery keys are not distinct, stable route identities across publication and consumption. Profile names share a namespace with descriptor aliases; named keys omit the endpoint URL; unnamed keys lowercase the entire URL, including case-sensitive paths. Some direct consumers still expect the old descriptor-keyed representation.

A concrete failure occurs with active profile `openai` and a second profile `proxy`, both using `CatalogID: openai` but different endpoints. The new scheduler discovers both. When `proxy` completes, publication writes its result under both `proxy` and `openai`, overwriting the active profile's own record. If the proxy reports `gpt-4.1` as text-only, both active prompt attachments and the run's tool-image callback reject images despite the active endpoint supporting them. A differently named active profile also borrows the foreign descriptor entry before its own discovery completes.

These are the related surfaces that must close together:

| Surface | Current behavior |
|---|---|
| `modelPickerDiscoveryCmds` → result message → `applyModelPickerModelsDiscovered` | Distinct profiles are scheduled, but non-generic descriptor aliases can overwrite another route's primary key. A custom profile named like another descriptor can collide too. |
| `providerEndpointKey` | Replacing a named profile's endpoint URL reuses its old capability record; a delayed old response has the same key. Unnamed `/TenantA/v1` and `/tenanta/v1` also collapse together. |
| `activeDiscoveryKeys`, `discoveredModelsForActiveRoute`, `savedProviderModelPickerItems` | Shared aliases can supply foreign capabilities or model lists. |
| `modelSupportsVisionFor` and `runAgentWithOptions` | Both faithfully consume the wrong route's evidence. The run snapshot itself is correctly isolated from later mutation. |
| Existing `modelContextWindow` and `resolveModelSwitchTarget`, affected by the changed map representation | With two discovered custom endpoints exposing `PrivateModel`, active A's window alternates between A's 1,024 and B's 8,192 tokens; `/model privatemodel` loses the previously discovered canonical `PrivateModel` spelling because the descriptor entry is no longer published. |

**Attribution:** the new discovered-first gate makes foreign denial override the curated `gpt-4.1` capability; equivalent prompt cases accept images at the merge base. The new scheduler and custom key representation also activate the two existing-reader regressions above: the same production discovery-command sequence at the merge base retains A's window and canonical model ID. The relevant target-branch code matches the merge base. These are consequences of this PR's migration, not a request to repair every older discovery fallback.

**Required correction:** retain enough originating endpoint identity to distinguish all rows above, prevent alias collisions and stale publications from another endpoint, and make the consumers affected by the new representation resolve the corresponding route. Preserve URL path case, existing fallback when matching evidence is absent, and the immutable run snapshot. Add regressions through scheduling/publication and both image consumers, plus the affected window/model-ID readers; cover both arrival orders, a same-name endpoint replacement, and delayed old responses.

**Author fix:** close this provenance contract across the listed producer and consumer paths in one pass. Removing only the alias assignment leaves reconfiguration and lookup holes; changing only the image predicate leaves the newly broken readers. Keep adaptations limited to this discovery migration. Do not redesign provider storage, credentials, the model registry, or general context-window policy.

### [P2] Reapply the MCP image gate when earlier images are replayed after a model switch

`internal/agent/loop.go:874–878`, together with `toolResultImageMessage` at 3503–3512 and the new MCP image assignment at `internal/mcp/registry.go:391`.

**Stated contract:** the accepted request was “Do not send MCP screenshots to a model that cannot accept images.” The new `SupportsVision` contract likewise says tool-produced images are dropped at the delivery boundary when the predicate is false.

**Root cause:** the gate runs only when constructing carriers for the current tool batch. Once an MCP image is in `messages`, later requests replay it without rechecking the effective model.

A three-turn reproduction using the real registered MCP tool and `escalate_model` is:

1. A custom route runs `gpt-4.1-mini`, whose discovered capabilities permit images, and receives an MCP screenshot.
2. A later turn escalates to `gpt-4.1`, which that route explicitly reports as text-only. No new screenshot is required in this turn.
3. The destination request contains the earlier screenshot even though its `SupportsVision` decision is false. A text-only backend rejects the multimodal request.

| Delivery edge | Current behavior |
|---|---|
| MCP JSON → `registryTool.Run` → image carrier | Payload is forwarded and gated when first appended. |
| Image generation and successful switch in the same batch | Correctly uses the destination model; existing tests cover this. |
| Earlier MCP carrier → later successful switch → next request/retry | Historical image bypasses the gate. |
| Failed/no-op switch, abort/stop, tool-result pairing | Preserve the existing effective-model and pairing behavior while closing the replay edge. |

**Attribution:** this is PR-activated by forwarding MCP payloads into history. The same MCP JSON and actual escalation sequence sends no image at the merge base, where MCP results have no image channel; it sends one on this head. The captured target has the same relevant loop and MCP code as the merge base. Older builtin/prompt-image replay limitations are not separate findings here.

**Required correction:** apply the existing effective-model capability decision when MCP image carriers are delivered again after a switch, preserving tool text, result pairing, and a truthful omission notice. Add a multi-turn regression where the screenshot precedes the switch, and check the request actually sent to the destination rather than only the newly generated batch.

**Author fix:** close the replay lifecycle in this PR's shared image-delivery change; filtering only `toolResultsWithImages` again cannot fix it. Keep the correction bounded to delivery of these carriers. No provider rewrite, capability-policy redesign, or new image-persistence feature is required.

## Validation

Affected MCP, agent, ACP, registry, discovery, and CLI package tests pass, as do scoped image/discovery race tests, provider image serialization tests, vet, formatting, and diff hygiene. The two broader TUI failures were reproduced against the merge-base implementation: date formatting depends on the local timezone, and the pinned-header assertion depends on checkout-path width. Neither is attributed to this PR.

Please address both findings together and validate the rebased head. This remains useful work for issue #823: #843 supplied the image channel, #874 supplied omission notices, and closed PR #989 is the predecessor rather than a landed duplicate.

euxaristia:

jatmn:
> @jatmn Before I start another repair round, please consolidate the remaining merge blockers and define the acceptance criteria for this revision.
> 
> For each blocker, please distinguish:
> 
> * A regression introduced by the latest repair.
> * An existing PR defect missed in an earlier review.
> * A pre-existing issue or optional improvement suitable for follow-up.
> 
> Please separate maintainer decisions and merge-readiness requirements from code findings, and confirm which earlier findings are closed.
> 
> I’ll address reproducible defects attributable to this PR. If another blocker emerges afterward, please identify what changed or what earlier coverage missed. Repeated rounds described as the remaining work have expanded again; I need a bounded repair target.

no.

PR #1014, opened 2026-09-05
fix(agent): retry mid-stream transport aborts before tool dispatch

### LINKED ISSUE #973: fix(providers): auto-retry or recover from mid-stream connection aborts (wsarecv / connection reset)
### Summary

During active streaming, transient socket interruptions such as Windows `wsarecv: An established connection was aborted by the software in your host machine` (WSAECONNABORTED) or `read: connection reset by peer` immediately abort the active turn with `Run cancelled.`, requiring the user to manually type `continue`.

### Context

When an established stream connection drops mid-generation before any tool calls or side effects are dispatched:
- The provider stream decoder encounters a network read error.
- The turn immediately terminates and surfaces the error to the composer.
- While pre-send connection failures are retried (#447), mid-stream socket aborts currently require manual user intervention even when no side effects were executed.

### Proposed Solution

1. **Safe Mid-Stream Retry**: When a stream drops before any tool execution begins, auto-retry the request with exponential backoff before failing the turn.
2. **Transient Error Classification**: Classify `WSAECONNABORTED`, `ECONNRESET`, and unexpected stream EOFs as retryable transport errors when the current turn has executed no tool calls.

euxaristia:
Fixes #973

## Summary

Auto-retry mid-stream transport aborts (e.g. Windows `wsarecv` / WSAECONNABORTED, connection resets) when no answer prose has been committed and before tool dispatch.

## Changes

- Recognize socket abort errors in `internal/agent/reconnect.go` via `isMidStreamTransportAbort`.
- Integrate mid-stream transport aborts into the stall-retry loop in `internal/agent/loop.go`.
- Reject classified provider application errors before matching socket abort signatures.

## Test plan

- Unit tests in `internal/agent/reconnect_test.go` and `internal/agent/midstream_retry_test.go`.
- Pre-existing review discussions from #976 apply.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically retries eligible mid-stream connection interruptions, including connection resets, unexpected closures, and Windows socket-abort errors.
  * Preserves cancellation results when a retry or reconnection is interrupted.
  * Prevents empty text events and incomplete tool calls from incorrectly blocking eligible retries.
  * Continues to surface non-retryable provider, authentication, rate-limit, and application errors without retrying.
  * Improves reconnect messaging and records recovery activity during eligible interruptions.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -25,10 +25,12 @@ const maxTurnsAnswer = "Agent reached maximum number of turns without a final an
 const maxTurnsFinalAnswerPrompt = "You have reached the tool-turn limit. Do not call tools. Give a concise final answer now: summarize what you completed, what you found, and any remaining blockers."
 
 // maxStreamStallRetries bounds how many times a turn that timed out (idle/stall)
-// OR hit a mid-stream transport abort (#973) WITH NO OUTPUT yet is re-issued on
-// a fresh connection before giving up. Only the no-output case is retried (a
-// partial turn would duplicate), so this is a safe recovery for a stalled/dead
-// pooled connection or a socket abort before tool dispatch.
+// OR hit a mid-stream transport abort (#973) with no answer text yet is re-issued
+// on a fresh connection before giving up. Incomplete tool-call previews are
+// allowed through the gate (they are never dispatched before the error return).
+// The safety claim is only that visible answer prose is not duplicated on retry —
+// not that every partial stream shape is excluded. Covers a stalled/dead pooled
+// connection or a socket abort before tool dispatch.
 //
 // Set to 1 (not 2): each attempt can itself idle for the full stream timeout
 // (~5min) before the stall is even detected, so 2 retries left an interactive
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -378,7 +378,11 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 					rec.StampFirstToken()
 				}
 				if onText != nil {
-					forwardedVisibleText = true
+					// Zero-length chunks are not committed answer prose; counting
+					// them would block an eligible mid-stream abort retry.
+					if s != "" {
+						forwardedVisibleText = true
+					}
 					onText(s)
 				}
 			}
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -491,6 +491,10 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 				notify = stallRetryNoticeFor(options)
 			} else {
 				notify = reconnectNoticeFor(options)
+				// Count this post-connect reissue even when the replacement
+				// connect succeeds immediately. streamWithReconnect only
+				// increments reconnect_count after a failed connect attempt.
+				options.Trace.Counter(trace.CounterReconnectCount, 1)
 			}
 			if notify != nil {
 				notify(attempt, maxStreamStallRetries)
@@ -505,6 +509,9 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			retryStream, retryErr := streamWithReconnect(ctx, provider, retryRequest, reconnectNoticeFor(options))
 			if retryErr != nil {
 				result.Messages = copyMessages(messages)
+				if ctx.Err() != nil {
+					return result, ctx.Err()
+				}
 				return result, retryErr
 			}
 			stallGenSpan := options.Trace.Span(trace.SpanGeneration)
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -455,9 +455,13 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			return result, ctx.Err()
 		}
 		// A stream idle/stall timeout OR mid-stream transport abort (#973 — wsarecv /
-		// connection reset / forcibly closed) is safely re-issued when the turn
-		// committed NO answer text — no forwarded visible prose (forwardedVisibleText)
-		// and no collected final text (collected.Text). This covers three cases:
+		// connection reset / unexpected EOF / connection closed / broken pipe) is
+		// safely re-issued when the turn committed NO answer text — no forwarded
+		// visible prose (forwardedVisibleText) and no collected final text
+		// (collected.Text). The abort gate matches midStreamAbortNeedles only, not
+		// the full shouldReconnect list: connect-phase timeouts and connection
+		// refused are NOT retried here (a slow healthy server must not cost a
+		// second prefill). This covers three cases:
 		//   1. Nothing streamed at all before the connection died (the original
 		//      macOS stale-pooled-connection hang past the response-header timeout).
 		//   2. The model streamed transient reasoning and began a tool call (e.g. a
@@ -503,6 +507,13 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			collected = zeroruntime.CollectStreamWithOptions(ctx, retryStream, forwardingOpts)
 			stallGenSpan.End()
 		}
+		// Recheck ctx after a retried CollectStream: helpers.go stores
+		// ctx.Err().Error() in collected.Error, and errors.New of that string
+		// drops the context.Canceled sentinel ACP branches on.
+		if ctx.Err() != nil {
+			result.Messages = copyMessages(messages)
+			return result, ctx.Err()
+		}
 		if collected.Error != "" {
 			// Route a reissued stream's non-stall error through the SAME recovery as
 			// the initial stream (image-rejection wrapping / context-limit compaction)
@@ -512,6 +523,10 @@ func Run(ctx context.Context, prompt string, provider Provider, options Options)
 			if stop != nil {
 				result.Messages = copyMessages(messages)
 				return result, stop
+			}
+			if ctx.Err() != nil {
+				result.Messages = copyMessages(messages)
+				return result, ctx.Err()
 			}
 			if collected.Error != "" {
 				result.Messages = copyMessages(messages)
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -25,14 +25,16 @@ const maxTurnsAnswer = "Agent reached maximum number of turns without a final an
 const maxTurnsFinalAnswerPrompt = "You have reached the tool-turn limit. Do not call tools. Give a concise final answer now: summarize what you completed, what you found, and any remaining blockers."
 
 // maxStreamStallRetries bounds how many times a turn that timed out (idle/stall)
-// WITH NO OUTPUT yet is re-issued on a fresh connection before giving up. Only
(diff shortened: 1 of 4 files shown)


Vasanthdev2004 (changes requested):
The mid-stream/connect-phase split is the right call, and keeping `isMidStreamTransportAbort` off `shouldReconnect` rather than reusing it is what makes that split hold. One gap in the exclusion list before this goes in.

## The exclusion list misses the prefix that carries every 5xx

`classifiedNonTransportPrefixes` covers `provider request error:`, `auth error:` and `rate limit error:`. It does not cover `provider error:`, which is what `providerio` produces in its default branch, so for every status that is not 401/403/429/503/529 and not 4xx.

That branch is not rare. `internal/providers/openai/provider.go:297` defaults a streamed error payload's status to 500 and passes it to `classifiedError` at :316, so any in-band SSE error whose `code` is outside the six mapped ones arrives with that prefix and the raw upstream message attached. If the message happens to contain "connection closed", "connection reset", "server closed" or a bare EOF, it is classified as a socket abort.

Same message, opposite verdict, decided only by which prefix providerio picked:

```
abort=false  "provider request error: connection closed is not a supported finish reason"
abort=true   "provider error: connection closed is not a supported finish reason"
abort=true   "provider error: upstream connection reset by peer"
abort=true   "provider error: unexpected EOF while reading upstream response"
abort=false  "provider error: 502 Bad Gateway"
```

The first of those is the string your own `midstream_retry_test.go` lists as must-not-retry. The guarded case and the unguarded one are the same upstream condition one status class apart.

`reconnect.go`'s own comment is the argument for fixing it: 500/502/504 "are non-idempotent by providerio's rule, the completion POST may already have reached the model before the upstream gave up, so replaying the connect risks duplicate billable work". `provider error:` is precisely that class.

The `HasStatusCode` backstop below cannot cover it. It scans the message text for the digits, and it does fire when the upstream echoes them, which is why `502 Bad Gateway` above returns false. On the streamed path the 500 is synthesized locally and was never on the wire, so there are no digits to find. It is unreliable rather than absent, and it is not something to rely on here.

Adding `"provider error:"` to the list should cost nothing. The only other producers of that prefix I found are `gemini/provider.go:300` and `:305`, both local marshal and argument-normalization failures rather than transport aborts. Genuine aborts arrive as `provider stream error:` or a bare net error, and neither is touched by the exclusion list.

There is a second, quieter consequence: the user is shown `[connection lost, reconnecting 1/1]` for what is an application error. That is a wrong statement about what happened, not just a wasted retry.

Two needles look unreachable under the current tests. `unexpected end` is shadowed by the word-boundary EOF check that runs first, and `connection was aborted` only ever appears alongside another matching needle, so neither would fail if removed. Worth one negative case each, or dropping them.

The 5xx and context-limit exclusions have no test of their own. Deleting either leaves the package green.

`TestRunDoesNotRetryApplicationErrorContainingEOFSubstring` passes with either of the two guards it appears to exercise removed, so it does not pin which one is doing the work.

The abort class reads as Windows-first: `wsarecv` and `forcibly closed` are the Windows spellings, and the POSIX side leans on `connection reset` and the EOF check. That is probably fine in practice, since those cover the common cases, but `ECONNABORTED` on macOS and Linux renders as "software caused connection abort", which matches nothing in the list.

## What is good

Not delegating to `shouldReconnect` is the decision that makes this safe, and the comment explaining why (connect-phase signals would re-prefill a healthy-but-slow server and lie to the user about the connection) is the sort of reasoning that stops the next person widening it.

Gating on `!forwardedVisibleText` and no tools dispatched is the right safety argument for a replay, and it is stated where the gate is rather than in the PR body.

jatmn (changes requested):

## Findings

- [P2] Reject classified `provider error:` failures before matching transport phrases
  `internal/agent/reconnect.go:178`
  The mid-stream abort retry is supposed to fire only for a dropped socket (Windows `wsarecv`, `connection reset by peer`, unexpected stream EOF) when the turn has committed no answer text and has not dispatched tools. The implementation decides that from a flattened `collected.Error` string: first a short prefix denylist, then substring needles.

  That denylist is incomplete against the prefixes `providerio.ClassifiedError` actually emits. The function uses:

  - `auth error:` for 401/403
  - `rate limit error:` for 429/503/529
  - `provider request error:` for other 4xx
  - `provider error:` for everything else, including 5xx

  `classifiedNonTransportPrefixes` lists the first three and omits `provider error:`. Transport/decode wrappers use a fifth prefix, `provider stream error:`, which is the class that should remain retryable. `"provider stream error:"` does not contain `"provider error:"`, so adding the missing prefix does not steal genuine socket failures.

  The gap is reachable, not theoretical. OpenAI-compatible streaming defaults an in-band SSE error object to HTTP 500 when `code` is not one of the six mapped values, then calls `classifiedError`. Anthropic in-band `error` events always call `classifiedError(http.StatusInternalServerError, message)`. Gemini uses the same helper when status is 0 or 5xx. Those paths never put the digits `500` in the message, so `errhint.HasStatusCode(..., "500", "502", "503", "504")` does not save them. If the upstream prose contains `connection closed`, `connection reset`, `server closed`, or a word-boundary `eof`, `isMidStreamTransportAbort` returns true.

  Concrete failure:

  1. The provider emits an in-band error that ClassifiedError renders as `provider error: connection closed is not a supported finish reason` (or `provider error: upstream connection reset by peer`, or `provider error: unexpected EOF while reading upstream response`).
  2. `CollectStreamWithOptions` stores only that string.
  3. No answer text was forwarded, so the stall/abort loop treats it as a socket abort.
  4. The user sees `[connection lost — reconnecting 1/1…]`, `reconnect_count` increments, the agent sleeps, and `Run` submits the same completion again.

  The 4xx twin of that first string is already in `TestRunDoesNotRetryClassifiedProviderErrorWithSocketPhrase` and correctly does **not** retry. Same upstream condition, opposite verdict, decided only by which ClassifiedError prefix was chosen. `provider error: 502 Bad Gateway` happens to be refused because the digits are still in the text; that is not a substitute for the prefix. Vasanthdev2004 asked for this exclusion on this head.

  Impact is bounded but user-visible: one extra billable completion, a false “connection lost” notice, and if the replacement stream happens to succeed the original application error is never returned. No tool is dispatched and no transcript turn is committed, which is why this is P2 rather than P1.

  **Root cause, not a one-off string.** The retry gate no longer has the HTTP status. Category survives only as the ClassifiedError prefix. The denylist is a reconstructed copy of that taxonomy and it is missing the default/5xx member. Phrase-by-phrase needle fixes will keep leaking the next sibling (`connection closed` today, some other 5xx prose tomorrow). The durable fix for this PR is: before any EOF/needle match, refuse every ClassifiedError prefix that is not the transport wrapper. Today that means adding `provider error:` next to the three prefixes you already exclude. A typed error kind is a later refactor, not a requirement here.

  **Tests that close this without opening a new round.** Keep the existing 4xx/auth/rate-limit negatives. Add a `Run`-level case whose error is `provider error:` plus a socket/EOF phrase (no status digits). Assert exactly one `StreamCompletion`, no reconnect notice, `reconnect_count` unchanged, the original error returned, and `OnToolCall` never fired. Keep positives for `provider stream error: unexpected EOF`, `connection reset by peer`, and `wsarecv: 10053` so the prefix fix cannot disable #973. A classifier table row alone is not enough; `Run` is where the extra POST and the notice happen.

  **Do not take this finding as a license to widen the PR.** Do not add POSIX `software caused connection abort`, HTTP/2 `RST_STREAM`/`GOAWAY`, `use of closed network connection`, a `malformed JSON:` exclusion, a new retry bound, stall-notice changes, or a public error-type API. Those are not this defect and they are not required to merge.

## Overall guidance to close this in one pass

This branch already spent several review rounds on #976 because a mid-stream retry is a lifecycle, while the tests and fixes followed individual reported strings. That is why feedback dripped: empty `OnText` blocking retry, `context.Canceled` lost after the replacement collect, `reconnect_count` not incremented on an immediate replacement, bare `eof` matching `oneOf`, then 4xx `provider request error:` + `connection closed`. Each item was real. Each fix was local to the example that had just failed. The next untested ClassifiedError prefix then became the next review.

There is one merge-blocking code issue on this head. It is the same class as the 4xx prefix gate you already shipped: provenance was flattened, then reconstructed from an incomplete list. Treat ClassifiedError’s prefix set as the contract, not the last string someone pasted into a test.

| Input / lifecycle edge | Attempts | Reconnect notice / counter | Required outcome |
| --- | ---: | --- | --- |
| `provider error:` + `connection closed` / `connection reset` / word-boundary `eof`, no answer text, no tools | 1 | none | Return the original error |
| `provider request error:` + `connection closed` (existing 4xx twin) | 1 | none | Unchanged: no retry |
| `auth error:` / `rate limit error:` + socket-looking prose | 1 | none | Unchanged: no retry |
| `provider stream error: unexpected EOF` / `connection reset by peer` / `wsarecv: 10053`, no answer text | 2 | one reconnect notice; `reconnect_count=1` on immediate replacement | #973 recovery; bound stays 1 |
| Non-empty forwarded answer text then abort | 1 | none | Unchanged: do not duplicate prose |
| Empty text callback, reasoning-only, or incomplete tool preview then genuine abort | 2 | real reconnect effects | Unchanged: retry allowed; `OnToolCall` stays 0 |
| Cancel during outer backoff, replacement connect backoff, or replacement collect | stop at cancel | no later work | `errors.Is(err, context.Canceled)` |

Use literal attempt counts and production-shaped prefixes. A positive sample must match only the signature under test. A negative sample must fail if category-blind substring matching returns.

That matrix is the acceptance bar for this finding. It is not a request to re-audit HTTP/2, POSIX abort spellings, decode errors, or the stall-retry path. Those were considered and are out of this PR’s approved #973 contract. Fix the missing `provider error:` provenance edge, pin it through `Run`, keep the genuine `provider stream error:` retries, and this should not need another classification round.

jatmn (dismissed):
@Vasanthdev2004 lgtm

jatmn (approved):

Vasanthdev2004 (approved):
Fixed, and it did not over-correct, which was the thing worth checking.

```
abort=false  "provider request error: connection closed is not a supported finish reason"
abort=false  "provider error: connection closed is not a supported finish reason"
abort=false  "provider error: upstream connection reset by peer"
abort=false  "provider error: unexpected EOF while reading upstream response"
abort=false  "provider error: 502 Bad Gateway"
abort=false  "auth error: connection closed"
abort=false  "rate limit error: connection reset"
abort=true   "provider stream error: connection reset by peer"
abort=true   "read tcp 1.2.3.4:443: connection reset by peer"
abort=true   "unexpected EOF"
```

Every classified prefix is excluded now, and the two classes that should still retry, a provider stream error and a bare net error, still do. That is the part a blanket exclusion would have broken.

`internal/agent` has one failure, `TestEagerToolSchemaTokenBudget` at 3653 tokens against a 3650 ceiling. It is identical on `origin/main` and it is Windows-only, so it is not yours. I have filed it separately.

PR #1018, opened 2026-09-07
test(tools): poll for the exec server's address instead of racing its startup

Vasanthdev2004:
`TestExecCommandForegroundServerReturnsSessionAndServesHTTP` has been failing Windows CI on pull requests that do not touch it: #988 (PierrunoYT, Sep 3) and #1008 (euxaristia, Sep 7), different authors, different subsystems. That is worse than a slow test, because it reports somebody else's change as the fault.

## What was wrong

The test gave the helper one 500ms yield to be spawned through the process manager, reach `net.Listen` and have its line drained back, then parsed that single read for the address. A loaded Windows runner has not got there in time, so the read is the still-running banner with no address:

```
--- FAIL: TestExecCommandForegroundServerReturnsSessionAndServesHTTP (10.55s)
    server output did not include listening address: "Command is still running.\nsession_id: 1000\n..."
    test root C:\Users\RUNNER~1\...\zero-exec-test-2934003598 still held after the cleanup deadline; leaving it
```

The 10.55s is not the assertion. The assertion fires immediately; the rest is the cleanup deadline being burned, because the address parse sat before the `t.Cleanup` that stops the session, so a failure there left the process running and holding the temp root.

Reproduced here by shrinking the first yield to 1ms, which is what a loaded runner effectively gives it. Same message, same duration, same trailing cleanup line.

## What changed

It polls with `write_stdin`, which is what the banner in that very output tells the caller to do, up to a bound far above any plausible start.

A genuine failure still ends it at once instead of waiting the bound out. A session that has gone away answers with an error status; a process that exited before listening answers with an `exit_code` line. Both end the loop immediately and report the whole poll transcript. Verified against a helper that exits without listening: the test fails in 0.9s naming the exit.

The stop is also registered before the wait now, so a failure cannot leave the process holding the temp root, which is what burned the ten seconds above.

The HTTP request is bounded too. It had no timeout at all, so a server that accepted a connection and never answered would have hung until the package timeout took the whole run down.

## Verified

- The fixed test passes with the first yield cut to 1ms, three runs.
- With the poll removed and the yield at 1ms it fails with the CI message verbatim, at 10.8s, with the same cleanup line.
- With a helper that exits instead of listening it fails in 0.9s naming the exit rather than after the deadline.
- `go test ./internal/tools/` green.

## Summary by CodeRabbit

- **Tests**
  - Improved reliability of server and terminal interaction tests through bounded polling and clearer failure handling.
  - Added validation for incremental output, session completion, and unexpected session termination.
  - Expanded coverage for server startup, echoed terminal input, and output received during active requests.
  - Strengthened timeout regression checks, including background-process cleanup and delayed session output.

### DIFF
--- a/internal/tools/bash_tool_test.go
+++ b/internal/tools/bash_tool_test.go
@@ -70,6 +70,9 @@ func runBashToolHelper(command string) {
 		fmt.Println("listening", listener.Addr().String())
 		server := &http.Server{
 			Handler: http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
+				// Printed per request, so a test can prove the session delivers output
+				// mid-flight: this line cannot exist before the first read has happened.
+				fmt.Println("served", request.URL.Path)
 				_, _ = response.Write([]byte("zero-server-ok"))
 			}),
 		}
@@ -522,7 +525,14 @@ func TestBashToolTimeoutKillsBackgroundChildren(t *testing.T) {
 	if result.Status != StatusError {
 		t.Fatalf("expected timeout error status, got %s: %q", result.Status, result.Output)
 	}
-	if elapsed > time.Second {
+	// The regression this guards is Run blocking until the background child's own
+	// sleep ends, so the bound only has to sit under that. One second sat almost
+	// exactly on it: the call legitimately costs a sandbox plan, a fork and exec of
+	// a shell plus a subshell, the 300ms timeout itself and the post-kill drain,
+	// and spawn alone was measured here at 720ms to 1.44s with every core busy. A
+	// merely slow runner therefore failed a correct system. Three seconds is still
+	// far below the child's lifetime, so nothing is given up.
+	if elapsed > 3*time.Second {
 		t.Fatalf("Run blocked %s past the 300ms timeout; background child held the pipes", elapsed)
 	}
 
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -423,6 +423,23 @@ func TestExecCommandForegroundServerReturnsSessionAndServesHTTP(t *testing.T) {
 	if !strings.Contains(served.String(), "served") {
 		t.Fatalf("a still-running session never delivered the line the server printed while serving:\n%s", served.String())
 	}
+
+	// AND A READ DRAINS WHAT IT RETURNS.
+	//
+	// One more poll, because "the line showed up" is satisfied just as well by
+	// output that is re-delivered forever. drain copies and then nils the buffer,
+	// so this read carries nothing; if it stopped clearing, every session in
+	// production would repeat its whole output on every poll while every other
+	// assertion here stayed green. The manager-level test that asserts the same
+	// count is skipped on Windows, so this is the only place guarding it there.
+	again := writeTool.Run(context.Background(), map[string]any{
+		"session_id":    sessionID,
+		"chars":         "",
+		"yield_time_ms": 250,
+	})
+	if strings.Contains(again.Output, "served") {
+		t.Errorf("a second read returned output the first had already delivered, so reads are no longer draining:\n%s", again.Output)
+	}
 }
 
 // waitForListeningAddress polls the exec session until the helper says where it
@@ -894,8 +911,16 @@ func TestExecCommandTTYSessionAcceptsInputOnLinux(t *testing.T) {
 	var transcript strings.Builder
 	transcript.WriteString(result.Output)
 	exitCode := result.Meta["exit_code"]
+	// BOTH THE ECHO AND THE EXIT, because the test asserts both.
+	//
+	// collect can return the drained echo on a yield timer before markDone has
+	// run, so a poll can carry got:hello with no exit_code meta yet. Stopping on
+	// the echo alone would then assert an exit that had not been observed. The
+	// condition deliberately keeps the echo term as well: gating on exit_code
+	// alone would stop polling before a slow shell start delivers the echo, which
+	// is the race this whole change exists to remove.
 	deadline := time.Now().Add(30 * time.Second)
-	for !strings.Contains(transcript.String(), "got:hello") && time.Now().Before(deadline) {
+	for (!strings.Contains(transcript.String(), "got:hello") || exitCode == "") && time.Now().Before(deadline) {
 		poll := writeTool.Run(context.Background(), map[string]any{
 			"session_id":    sessionID,
 			"chars":         "",
--- a/internal/tools/exec_command_test.go
+++ b/internal/tools/exec_command_test.go
@@ -373,10 +373,6 @@ func TestExecCommandForegroundServerReturnsSessionAndServesHTTP(t *testing.T) {
 	if err != nil {
 		t.Fatalf("foreground server should return session_id, meta=%#v output=%q", start.Meta, start.Output)
 	}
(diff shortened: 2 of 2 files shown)


jatmn (changes requested):
This PR fixes a real CI problem: `TestExecCommandForegroundServerReturnsSessionAndServesHTTP` was failing on unrelated Windows runs because a single 500ms yield is not a guarantee that a re-execed helper has reached `net.Listen` and had its line drained. Bounded `write_stdin` polling, registering cleanup before address discovery, and a bounded HTTP client are the right shape for that failure mode. I would merge this for the stated goal once the optional items below are either addressed or consciously deferred.

## Merge readiness

- Branch head `b9fc003d` is current with live `main` (`aadb4a27`). No rebase needed.
- All required CI checks pass, including `Smoke (windows-latest)`.
- CodeRabbit requested changes on head. I triaged those inline comments below; none of them are merge blockers on their own, but two overlap with the optional hardening in **Findings**.

## Findings

- [P3] TTY poll loop can stop before `exit_code` meta is populated
  `internal/tools/exec_command_test.go:898`

  **What happens.** The loop exits when `got:hello` appears in the accumulated transcript, but `exit_code` is tracked separately from `poll.Meta`. `managedProcess.collect` can return drained echo bytes on a yield timer while `markDone` has not run yet (`internal/execution/process_manager.go:422–467`), so a poll may contain `got:hello` with an empty `exit_code` meta.

  **Root cause.** The loop condition only gates on transcript text, not on session completion metadata. That decouples echo observation from exit observation even though the test asserts both.

  **Why this is P3, not blocking.** For `read line; echo got:$line`, echo and exit are microseconds apart, so the window is narrow. The pre-PR test had the same coupling on a single 1000ms `write_stdin` call. Linux CI is green on head and the test completes in milliseconds locally. I have not reproduced a flake.

  **Minimal fix (no drift).** Keep polling until *both* conditions hold, then assert:

  ```go
  for (!strings.Contains(transcript.String(), "got:hello") || exitCode == "") && time.Now().Before(deadline) {
  ```

  Do **not** switch the loop to `exitCode == ""` alone — that would stop polling before the echo arrives on a slow shell start, which is the flake this PR is trying to fix elsewhere.

  CodeRabbit raised the same point on this line; the fix above is the drift-safe version.

- [P3] Commit `b9fc003d` describes anti-redelivery hardening that is not in the tree
  `internal/tools/exec_command_test.go:412`

  **What happens.** The commit message says: *"One more read after the served line, asserting it is not repeated, pins that."* The test returns as soon as `strings.Contains(..., "served")` is true and never performs a follow-up poll.

  **Root cause.** Adversarial-review intent was recorded in the commit message but the assertion was not added. This is a documentation/code alignment gap, not evidence that incremental delivery is broken today.

  **Why this is P3, not blocking.** `processOutputBuffer.drain` copies then nils `buffer.data` (`internal/execution/process_manager.go:611–616`), so correct drain semantics cannot re-deliver old bytes. If re-delivery ever happened it would be a production `collect` bug; this test would be a useful guard on `windows-latest` where `TestProcessManagerRetainsAndContinuesWithStableIdentity` is skipped, but absence of the guard does not create a false pass while drain works.

  **Minimal fix (no drift).** After the first poll that observes `served`, issue one more empty-chars `write_stdin` poll and assert the `served` line is not duplicated in the new chunk (for example `!strings.Contains(followUp.Output, "served")` or a line-count check on the handler output shape `served /`). Alternatively, drop the claim from the commit message if you decided the guard was unnecessary given drain semantics. Either aligns message with code; do not change production `collect` for this.

## Clarifications (not findings — avoid drift)

These came up in automated review and my own pass. They are **not** defects in the current diff.

**Bash timeout elapsed bound (`bash_tool_test.go:535`).** Widening `elapsed` from 1s to 3s is intentional and correct for the measured failure mode: a *healthy* `Run()` on a loaded runner legitimately costs sandbox planning, fork/exec, the 300ms timeout, and the shortened post-kill drain (720ms–1.44s spawn measured in the comment). The old 1s ceiling false-failed correct systems.

The orphan-child regression this test names is guarded by the sentinel `touch`, not the elapsed check. If kill fails, the background `(sleep 1; touch sentinel)` still writes the file and the stat at lines 539–545 fails regardless of elapsed.

**Do not** raise the `childSleep` constant alone as CodeRabbit suggested. That constant is only used in the post-`Run` sleep before `stat`; the shell command hardcodes `sleep 1` in the format string at line 516. Increasing `childSleep` would not lengthen the background sleep or restore elapsed-vs-child timing.

If you ever want a tighter elapsed signal *without* reintroducing spawn flakes, the root fix is to wire the same duration into the command (for example `fmt.Sprintf("(sleep %d; touch %s) & wait", int(childSleep.Seconds()), ...)`) and set `elapsed` below that value. That is out of scope for this PR unless you want extra pipe-hold coverage beyond the sentinel.

**`waitForListeningAddress` address-before-`exit_code` ordering (`exec_command_test.go:485–502`).** Returning a parsed address before checking `exit_code:` in the same chunk is harmless here: the `http-server` helper blocks on `Serve` after printing `listening`, and the HTTP step immediately after would fail if the process died. Reordering checks would only change failure diagnostics, not test validity.

**First-read stdout coverage (`waitForListeningAddress` comment block).** Deliberately surrendered and repaid by the post-HTTP `served` poll. That trade-off is documented in-tree and matches the PR intent; do not reintroduce the single-yield race to pin first-read delivery.

## Summary

Merge for the Windows startup-race fix. Optional P3 hardening: TTY loop condition (both echo and `exit_code`), and either add the anti-redelivery poll or trim the commit claim. No production code changes required for the core fix.

Vasanthdev2004:
Both done at 2510c59, green on all three platforms.

**The commit message was wrong, and you were right to call it.** The anti-redelivery poll it described was written and it passed, and then a `git checkout -- .` in the falsification run that followed reverted it, because it was the one edit made after my checkpoint commit. I amended the message afterwards and recorded work that no longer existed. I audited the rest of that commit against the tree before touching anything: the 60s deadline, the host:port validation, the served marker and the exit_code fast-fail all survived, since they predated the checkpoint. Only the poll was lost. It is restored now and the claim matches the code. Verified by disabling drain: the second read returns the served line again and the test fails on it. And this round I committed before every falsification and grepped for the guard after the checkout, so it is in the tree this time.

**TTY loop.** Gates on both the echo and the exit now, in the form you gave:

    for (!strings.Contains(transcript.String(), "got:hello") || exitCode == "") && time.Now().Before(deadline) {

The echo term stays in deliberately, with a comment saying why, so nobody later simplifies it to `exitCode == ""` and stops polling before a slow shell start delivers the echo.

On the clarifications: agreed on all three, and I am not touching them. The bash elapsed bound stays at 3s with the sentinel as the real orphan guard; the address-before-exit ordering in `waitForListeningAddress` stays as is since the HTTP step catches a dead helper immediately; and the first-read coverage stays surrendered and paid back by the served poll rather than reintroducing the single-yield race. I read the CodeRabbit suggestion to raise `childSleep` and did not take it, for the reason you gave: the shell command hardcodes `sleep 1`, so that constant does not reach the thing it looks like it controls.

jatmn (approved):
lgtm

Vasanthdev2004:
@coderabbitai review

This review is on `b9fc003d` and the head is `2510c592`. The findings it raised were addressed in that commit and jatmn has since approved at the head, so the changes-requested state is the last thing holding this one.

PR #1025, opened 2026-09-08
fix: let the configured forward proxy through the loopback guard

### LINKED ISSUE #569: providerhealth: safeDialContext blocks proxy connections through localhost proxy

Vasanthdev2004:
Closes #569. Takes over #594 by @baoyu0, who did the original work and handed it off; the approach is theirs and they are credited as co-author on the commit.

With `HTTPS_PROXY=http://127.0.0.1:3067`, `zero doctor --connectivity` fails with `proxyconnect tcp: provider connectivity URL is unsafe: loopback hosts are blocked`. The transport dials the proxy first and tunnels the request through it, and the SSRF-guarded dialer refused that dial as loopback even though the real target had already been validated and was public. `web_fetch` had it worse: its transport set `Proxy = nil`, so on a machine that only reaches the network through a proxy it could not fetch at all.

### What changed

- New `internal/proxydial` with one question: is the address the dialer was asked for exactly the endpoint the transport's Proxy names for this request? Same host, same port, with a missing proxy port taken as the scheme default (80/443). A sibling loopback port is still refused. `providerhealth.safeDialContext` and `webFetchSafeDialContext` both take that answer and skip the pin for that one address and nothing else.
- The dialer asks the transport it is installed on for its Proxy at dial time instead of reading the environment a second time. `http.ProxyFromEnvironment` caches the environment once per process, so a separate read can disagree with what the transport actually dials, and it left the path untestable.
- `web_fetch` honours `http.ProxyFromEnvironment` again. The nil was set on purpose in #121 as part of the SSRF hardening, so this is a deliberate reversal and the reasoning is in the code comment: the target is still validated before the request goes out, with the same block list the dialer applies. What a forward proxy removes is only the dial-time rebind re-check, and that is inherent to using any forward proxy. The pre-permission URL validation stays.

### Differences from #594

- #594 deleted `validateWebFetchURLBeforePermission` and its two call sites. That is unrelated to the proxy, and it is the check that keeps an unsafe URL from reaching the permission prompt. Kept.
- #594 matched any port when the proxy URL had none (`proxyURL.Port() == "" || ...`). `HTTPS_PROXY=http://127.0.0.1` would then have exempted every loopback port on a direct dial, which `NO_PROXY` can produce for the same host. This matches the port the transport actually dials.
- #594 called `http.ProxyFromEnvironment` inside each dialer, separately from the transport. See above.

### Tests

- `internal/proxydial`: exact endpoint match, missing-port default, case-insensitive host, both schemes asked, no proxy, and the transport followed at call time.
- In each package: the configured proxy is dialed unpinned with the resolver untouched, a sibling loopback port is still refused, and with no proxy the guard is unchanged.
- In each package: a real request driven through a loopback listener acting as the proxy, checking that the `CONNECT` for the target arrives. That is the wiring from client to transport to dialer, which the unit tests alone did not cover. The proxy is substituted on the transport rather than through `HTTPS_PROXY` because of the once-per-process cache.

On main, the same probe (`HTTPS_PROXY` at a loopback listener, the connectivity client, `GET https://api.example.com/v1/models`) fails with the exact error from #569 and the listener never sees a connection. On this branch the listener receives `CONNECT api.example.com:443 HTTP/1.1`.

Each guard was checked by removing it: the exemption in either dialer, the transport wiring in either package, the `Proxy` reversal in `web_fetch`, and the exact port match. Each one fails the test written for it, by name.

## Summary by CodeRabbit

- **Security**
  - Improved web requests through proxies while preserving protections for direct connections to private, loopback, and special-use addresses.
  - Added safer handling for HTTPS proxy tunneling and internationalized proxy hostnames.

- **Documentation**
  - Clarified address validation, DNS resolution, and proxy behavior for web-fetching features.


jatmn (changes requested):
I found two issues to address and a proxy-trust decision to resolve before this is ready. The requested corrections are scoped to the HTTP forward-proxy behavior in #569.

## Findings

### [P1] Bind the exemption to the actual request's proxy decision

`internal/proxydial/proxydial.go:70-71`

**Failure path.** With `HTTP_PROXY=http://rebind.example:80` and `NO_PROXY=rebind.example`, a request to `http://rebind.example/` is direct. Go selects that route using the real request. `IsProxyTarget` then calls the same proxy function with `proxy-target.invalid`, which is not excluded, and gets `rebind.example:80` back. The direct origin address matches, so both guarded dialers skip their safety resolver and pass the raw hostname to the ordinary dialer. If that hostname resolves publicly during preflight and privately at connection time, the direct request can reach the private service. This requires the direct origin to match a configured proxy endpoint; it is not a bypass for every arbitrary hostname.

A real-transport probe confirms the skipped guard: this head calls the capture dialer with `rebind.example:80` and performs zero safety lookups; the base performs the lookup and rejects its private answer. The probe verifies the changed validation path without opening an external connection. There is also an inverse failure: `NO_PROXY=.invalid` or a host-dependent callback can exclude the synthetic host while selecting a legitimate loopback proxy for the real request, causing the configured proxy to be rejected.

**Root cause and requested outcome.** Sharing the transport's `Proxy` function does not share its decision: calling it with a different request can produce a different route. Please make the exception depend on the proxy actually selected for the connection's request, with a direct selection retaining the normal address checks and IP pinning. Apply that contract through both `providerhealth` and `web_fetch`. The precise implementation is open, but changing the placeholder hostname or adding more scheme probes cannot establish this invariant. If selection metadata is carried into dialing, keep it associated with that request/dial; a shared “last selected proxy” variable would let concurrent requests overwrite each other's decision.

This is the actual-request binding requested in [my review of #594](https://github.com/Gitlawb/zero/pull/594). Exact-port matching fixes the earlier wildcard case, but does not fix this same-port route mismatch.

**Focused regression coverage.** Exercise the real transport with a direct `NO_PROXY` request whose origin equals the configured proxy endpoint, and assert that a private safety-resolver answer is rejected before the underlying dial. Cover the opposite direction with a host-dependent proxy selection and verify the selected loopback proxy is reached. Also check scheme separation, such as only `HTTPS_PROXY` being set while the matching HTTP origin is direct. Exercise both guarded callers; the current constant-callback CONNECT tests cannot distinguish these decisions. Environment-based cases need fresh processes because `http.ProxyFromEnvironment` caches its configuration; injected callbacks can keep the other cases hermetic.

### [P2] Canonicalize proxy hostnames before comparing dial addresses

`internal/proxydial/proxydial.go:75`

**Failure path.** For `HTTPS_PROXY=http://büro.example:3067`, Go converts the proxy hostname to ASCII and supplies `xn--bro-hoa.example:3067` to `DialContext`. This comparison uses the original Unicode `URL.Hostname()`, so `EqualFold` returns false. If that HTTP proxy resolves to a private LAN address, the connection falls through to the ordinary blocklist and is rejected. An actual transport capture reproduces the ASCII address and failed match. This is an incomplete configured HTTP-proxy fix, not a claim that this configuration worked on the base.

**Root cause and requested outcome.** The matcher and transport represent the same hostname differently. Please compare the selected HTTP proxy endpoint using the transport-compatible canonical hostname representation and its exact effective port. This can be addressed alongside the request-selection correction, but it needs its own regression case: correct request provenance alone does not make Unicode and ASCII strings equal. Preserve the existing checks for unrelated destinations. Resolving a proxy name into a broad IP allowlist would change which destinations are trusted and would not be an equivalent correction.

**Focused regression coverage.** Drive the transport with equivalent Unicode and ASCII spellings of an HTTP proxy and verify that both recognize the same selected endpoint. Retain the sibling-port and unrelated-host rejection cases. A deterministic resolver can establish the private-LAN failure path without relying on public DNS. The required behavior is hostname equivalence within HTTP proxy support; no new proxy protocol is needed.

## Needs maintainer decision

### Decide whether configuring a forward proxy delegates target-address enforcement to it

`internal/tools/web_fetch.go:332-333`

The PR intentionally sends the original target hostname to the configured proxy after local validation and explicitly acknowledges losing the target IP pin. The consequence is real: an ordinary proxy can resolve a hostname to a private address after Zero validated a public answer, then return internal-service content. Redirect preflight validation has the same separation from the proxy's eventual resolution. A deterministic forwarding fixture returns private-service content on this head while the base blocks the corresponding private dial; this models differing destination resolution, not a deployed DNS-rebinding exploit.

This differs from the first finding: the actual proxy is selected here. Fixing request provenance prevents accidental exceptions for direct dials, but does not make the proxy use Zero's validated destination address. The README/tool description retains the public-only promise, while the shell sandbox does not enforce this in-process fetch.

Please resolve the trust contract before prescribing implementation work. If configured proxies are trusted for destination-address enforcement, make that boundary and the limitation of local preflight checks clear in the user-facing guarantee. If Zero must independently enforce public destinations through proxies, that requires additional protection and an agreed design. CodeRabbit's [target-enforcement concern](https://github.com/Gitlawb/zero/pull/1025#discussion_r3954407199) identifies a real consequence, but an unconditional demand to redesign or disable proxying would presume this policy decision.

## Scope of the requested changes

Keep the working local HTTP-proxy path, pre-permission URL rejection, redirect validation, exact endpoint matching, and normal direct-dial checks. The guidance above is about route provenance and equivalent hostname representation, with proxy trust called out for a separate decision. SOCKS support was deferred in the issue discussion and is not requested here. Neither finding requires adding `ALL_PROXY` support, changing environment-cache semantics, or refactoring unrelated provider/authentication behavior.

Vasanthdev2004:
Both findings were right, and the P1 was mine to own: I reproduced it before touching anything. A host-dependent proxy function (the NO_PROXY shape) sends the real request direct, my synthetic probe still came back proxied, and the direct dial to `rebind.example:80` reached the capture dialer with zero safety lookups. The guard was off for a dial that was never going to a proxy.

Fixed in 221fb930, by dropping the guess rather than improving it.

**Route provenance.** `proxydial.Transport` wraps the transport, asks its `Proxy` function about the actual request, and puts the answer in that request's context. The transport derives the dial context from the request context, so the dialer reads back the route this very connection is being opened for. Per-request, no shared last-selected variable, and the context key is unexported so nothing outside the package can plant one. Both callers install the wrapper; unwrapping either one now fails a test by name.

**Hostname canonicalization.** This one dissolved rather than being fixed, so it deserves the argument rather than a claim. A transport opens one connection dial, and when a proxy is selected that dial goes to the proxy; the target is reached through it and never dialed. So the recorded route already identifies the dial, and comparing the address on top of it could only ever disagree with the transport. The address comparison is gone, along with `IsProxyTarget` and the port defaulting. I kept your regression case because it pins exactly that: a proxy configured as `http://büro.example:3067` is dialed as `xn--bro-hoa.example:3067`, both spellings are recognised, and the two produce the same dial address. That test fails on the previous head.

**Coverage**, in both guarded callers, driven through the real transport with injected callbacks so the route decision is Go's:

- Direct `NO_PROXY` request whose origin equals the configured proxy endpoint: the resolver runs and its private answer is refused before the dial.
- Host-dependent selection in the other direction: the selected loopback proxy is reached.
- Scheme separation: an https-only proxy leaves the matching http origin direct.
- Unicode and ASCII spellings of one proxy host.
- Concurrency: 80 interleaved requests, half proxied and half direct, and the direct dials never pick up the exemption.
- The existing sibling-port, no-proxy and CONNECT cases, unchanged.

No environment-based cases and so no fresh processes: the proxy function is substituted on the transport, which keeps them hermetic and order-independent.

**On the trust decision.** Taking it: a configured forward proxy is trusted for destination resolution, and the boundary gets documented rather than defended.

The reasoning is that there is nothing to defend it with. Once a request goes through a forward proxy, the proxy resolves the hostname and chooses the address; Zero cannot pin what it does not dial. The only enforcement available is refusing to use proxies at all, which is exactly the outage in #569 and a worse trade: a forward proxy already sees and can rewrite every request through it, so a user who configures one has already extended it more trust than this. The local pre-request checks stay, and they still reject an internal target before anything leaves the machine; what they stop binding is the address finally reached.

So the honest fix is to say so. README now states that with `HTTP_PROXY`/`HTTPS_PROXY` set the proxy decides which address is reached, that a proxy answering differently can reach a private service, and that leaving the variables unset keeps the checks enforced end to end. `docs/HOW_ZERO_WORKS.md` carries the same note next to the tool. The tool description is unchanged: it tells the model what to ask for, and "public URLs" is still the right instruction.

CI is running; I will report when it settles.

jatmn (approved):
LGTM

anandh8x (approved):
## Review at 221fb93

**Approve.** All three of jatmn's items are closed, and the final design is better than what was asked for.

What's good:

- The request-context route design (decision recorded on the request via private key, dialer reads it back) doesn't just fix the P1 provenance hole — it dissolves P2 entirely. No address comparison means the punycode-vs-Unicode mismatch can't occur, and the code states that outright instead of leaving it implied.
- The proxy-trust call got resolved in plain README language: configuring a forward proxy is accepting that it sees and can rewrite everything. Honest contract, user-visible.
- Proper handoff from #594: baoyu0 co-credited, with the differences documented (kept the permission validation, closed the wildcard-port hole). That's how you take over someone's PR.
- Mutation-checked guards — each removal fails its named test. The CONNECT-arrival wiring tests cover what the unit tests can't.

Verified in an isolated worktree: build clean, `proxydial` + `providerhealth` + `tools` (WebFetch/Proxy) all green.

Note: the gate still shows the stale 09-08 blockers; CodeRabbit's and jatmn's 09-09 approvals supersede them. Needs dismissal, not more review.

PR #1040, opened 2026-09-09
fix(sandbox): keep a derived ACL target inside the write root it came from

### LINKED ISSUE #1024: sandbox: a junction at .git sends the fallback carveouts outside the workspace on Windows
On `main` at f30f550e, `gitMetadataWriteCarveouts` in `internal/sandbox/profile.go` emits `<root>/.git/hooks` and `<root>/.git/config` as write-denied carveouts. The Windows ACL apply then opens those paths with `FILE_FLAG_OPEN_REPARSE_POINT`, which guards only the final path component. A junction at `<root>/.git` is traversed as an intermediate, so the deny-write ACEs land on `<junction-target>/hooks` and `<junction-target>/config`, outside the workspace.

`mklink /J` needs no privilege, so any local user with write access to a workspace that has no `.git` yet can plant one.

Evidence, with provenance. I confirmed the carveout shape by reading the code on main: the two children are emitted unconditionally, with no lstat of `.git` at all. The traversal itself was probed during the review of #805: a junction created unelevated with `mklink /J`, then both `<root>\.git\hooks` through the junction and `<junction-target>\hooks` directly opened with the apply's exact CreateFile flags, and both resolved to the same volume serial and file index. Opening `<root>\.git` itself reports reparse=true and is refused, but with these carveouts that path is never opened.

Consequences differ by path. On the unelevated per-command path this only touches objects the user could already re-ACL, and it is rolled back after the command. On `zero sandbox setup`, which requires Administrator, the same redirect happens with an elevated token: the same TOCTOU class as #728, one component further up.

#805 does not introduce or widen this; it improves the regular-file case and leaves the junction case exactly as main has it. It also does not close it, which is why this is its own issue against the ACL apply rather than against that PR. The fix belongs in the apply: refuse to write an ACE through any owned intermediate that is a reparse point, which is the rule the rooted descent in #901 already follows for the runtime tree.

Vasanthdev2004:
Closes #1024.

`FILE_FLAG_OPEN_REPARSE_POINT` guards one component. The apply opens its target with it, which refuses a reparse point at the final component and resolves every component above it, as any absolute path has to. That is the whole guard for a path the operator named, and not enough for one this package derived.

The write-root carveouts are derived: `<root>/.git/hooks` and `<root>/.git/config` are constructed from a root, and `<root>/.git` is a name an unprivileged workspace writer can create before setup runs. `mklink /J` needs no privilege. With a junction there, the apply opens `<junction-target>/hooks`, which is an ordinary directory with nothing wrong about its final component, and `zero sandbox setup` writes a deny ACE on it as Administrator, outside the workspace.

Probed rather than reasoned about, since Windows path semantics do not reward reasoning. Opening `<root>/.git/hooks` through a junction and opening the outside path directly return the same `GetFinalPathNameByHandle` answer, `\\?\...\002\target\hooks`, while the anchor answers `\\?\...\001`. So the handle knows where it really is even when the name does not.

### What changed

A derived entry now records the write root it came from, and the apply requires the object it finally holds to still live under it. The comparison is between two `GetFinalPathNameByHandle` answers, so both sides normalize the same way (`\\?\` prefix, long names, drive letter) rather than being compared as written.

Materialization is checked before it creates, because `os.MkdirAll` follows the same reparse points and would otherwise put the directory on the far side with only the after-the-fact check noticing.

This is deliberately not a refusal of every reparse point on the path. Above the write root the path is the operator's, who may keep a workspace under a junction or a mapped directory, and refusing that would break setups this has nothing to say about. Only the tail the sandbox appended is held strict, which is the "owned intermediate" rule the issue asks for.

### What is not fixed here

The window between the pre-create check and `os.MkdirAll` is still open: closing it needs the components created relative to retained handles rather than by pathname, which is the rooted descent in #808. Same for the `os.RemoveAll` on the failure path. This PR is scoped to refusing the ACE, which is what #1024 describes, and stays out of #808's way.

### Tests

- A carveout redirected by a junction at `.git`: the ACE is refused and the error names where it actually resolved.
- The same carveout in an ordinary workspace still gets its ACE, so the refusal is not just failure.
- Materialization behind the junction: refused before anything is created, and nothing appears on the far side.
- An operator-named path below a junction is applied as before, since it carries no anchor.
- The plan builder anchors the derived carveouts and leaves operator-named deny paths unanchored.

Junctions rather than symlinks throughout, so these run on an ordinary unelevated account, which is the account the attack needs.

Two of these exist because falsification caught me. Reverting the pre-create check left its test passing, which turned out to mean the check was doing nothing: it asked where the deepest existing ancestor lived, and a junction answers with its own path, because the open does not follow a final-component reparse point. What disqualifies that ancestor is that it IS a reparse point. Reverting the plan wiring also left everything passing, because every apply-level test hands the group an anchor directly and none of them would notice the builder never setting one. Both now fail by name.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows sandbox protection against junctions and other path redirections that could cause ACL changes or directory creation outside the intended workspace.
  - Validates anchored paths and filesystem ancestors before applying permissions or creating directories.
  - Preserves expected behavior for ordinary paths and explicitly unanchored operator paths.
  - Refined permission updates to remove targeted write restrictions while retaining read protections where appropriate.

- **Tests**
  - Added coverage for containment checks, safe directory creation, inheritance handling, and ACL application behavior on Windows.


Vasanthdev2004:
Both taken, at 0ddf21d1 and 832b73b5.

**The anchor on `ReadOnlySubpaths`.** Right, and I had conflated two things that arrive at the same place. `ProtectedWriteDenyPaths` is fed from two sources: `ProtectedMetadataNames`, which is joined onto the root and is under it by construction, and `ReadOnlySubpaths`, which is a profile field an operator can set to any path. The in-tree producer fills it from `gitMetadataWriteCarveouts(root)`, so it is under the root in practice, but the field does not promise that and a config placing it elsewhere works today. Anchoring it turned that into a containment refusal.

Only paths lexically under the root are anchored now, via the existing `pathWithinRoot`. That is where the derived carveouts are anyway, so the guard is unchanged where it matters, and anything else keeps the final-component check it always had.

**The rollback in the successful-apply tests.** Also right, and following it turned up the sharper half. Restoring the snapshot failed with `Access is denied` on the reopen: the ACE denied `S-1-5-32-545`, a group the test process is a member of, so the apply revoked its own `WRITE_DAC` and nothing could put the DACL back. Leaving the ACE in place would have failed `t.TempDir` cleanup; restoring it could not run at all.

The tests deny `S-1-5-32-546` instead, a group this process is not in, which is what a capability SID is in production, and roll back through `t.Cleanup` as you suggested. So the ACE is now both realistic and reversible, and the rollback path gets exercised rather than assumed.

While there: the plan tests I added were in a Windows-only file even though the builder is cross-platform, so nothing checked the anchoring on Linux or macOS. Moved beside the other `BuildWindowsACLPlan` tests, with your out-of-root case added. Removing the `pathWithinRoot` gate fails it by name.

kevincodex1 (dismissed):

kevincodex1:
bro @Vasanthdev2004 please rebase to main and fix conflicts

Vasanthdev2004:
Rebased onto main at c4fff2b1. Two conflicts, both plain unions with #1006: the entry struct now carries NoInherit next to Anchor, and the test file had both sets of new tests appended at the end. Sandbox package is green on Windows here and the linux and darwin cross-builds pass. Approval got dismissed by the push, so one more click when you have a moment @kevincodex1.

gnanam1990 (approved):
No merge blockers remain in this PR's production-reachable scope.

## Verdict

**Approve.** Reviewed exact base `6937a309cf00` and head `c4fff2b1deb6`. The branch is mergeable and all reported checks, including Windows smoke, are green.

## Review closure

- Derived protected-write targets now carry their originating write-root anchor only when they are actually under that root. Operator-named and out-of-root paths stay unanchored, preserving supported existing configuration.
- ACL mutation is bound to the opened no-follow handle and the opened object is verified against the anchor before `SetSecurityInfo`, closing the production issue where an intermediate junction redirected an elevated ACE write outside the workspace.
- The successful Windows fixtures now retain and roll back their snapshots with a capability SID the test process does not belong to, so they neither strand test DACLs nor merely assume rollback.

I rechecked the open materialization-race concern rather than carrying it forward as a blocker. The check-to-`MkdirAll` window is real at the helper level, but it is not reachable from a valid production plan today: `BuildWindowsACLPlan` gives `Anchor` only to protected-write entries, which have `Materialize=false`; the only generated `Materialize=true` entries are `DenyRead`, which have no anchor, and both restricted-token runner tiers reject every non-empty Windows `DenyRead` profile before ACL planning/application. The anchored-materialization test constructs a group that production cannot currently emit. If a future change enables anchored materialization, retained-handle component creation and identity-bound cleanup should be a prerequisite (the separately tracked rooted-descent work covers that direction), but that deferred capability is not required to close this PR's existing-target ACE boundary.

## Validation

- `go test -p 2 ./internal/sandbox` — pass on macOS for cross-platform tests.
- Windows sandbox test binaries cross-compile for `amd64` and `arm64` — pass.
- Windows sandbox setup/runner commands cross-build for `amd64` — pass.
- Current GitHub Windows smoke and all other reported checks — pass.
- `git diff --check` — pass.
- No dependency or external-integration change in the PR delta.

I did not execute the Windows-only junction/ACL tests natively in this local review; the native Windows CI result is green.

kevincodex1 (approved):

PR #953, opened 2026-08-24
perf(tui): bounded file read, LRU rendering cache, and size caps for full-file view

hazyhaar:
## Summary
Fixes #833

`internal/tui/file_view.go` previously read files from disk synchronously and performed Chroma syntax highlighting directly inside the `View()` render loop on every frame, causing UI stutter and unbounded allocations on large files.

### Key Changes
- Decoupled disk I/O and syntax highlighting into an asynchronous cache keyed by filepath, file size, modtime, and theme.
- Enforced hard memory limits: 4,000 maximum rendered lines, 1 MiB total byte cap, and 4 KiB max line length.
- Invalidates the cache cleanly upon theme switches (`applyTheme`).
- Added unit and concurrency tests (`internal/tui/file_view_test.go`) validating 0 additional I/O on repeated `View()` calls and clean truncation under `-race`.

## Summary by CodeRabbit

- **New Features**
  - File views now load asynchronously, keeping the interface responsive.
  - Added loading and error states for file content.
  - File content refreshes automatically after resizing, theme changes, edits, and related updates.
  - Improved caching and validation help ensure current content is displayed.

- **Bug Fixes**
  - Corrected syntax highlighting backgrounds for themed file views.
  - Prevented stale cached content from being reused after theme changes.
  - Improved handling of late file-load results, rapid updates, and reopened views.

PR #1047, opened 2026-09-12
feat(tui): richer /resume session browser

anandh8x:
## Summary

- `/resume` rows are now two lines on medium-and-wider terminals: the age + title row, then a faint detail line with the session's project directory (`~`-contracted), the model that ran it, and a status chip for forks and tagged sessions
- The timestamp column now reads as relative last-activity age ("3h ago", "12d ago"), falling back to the calendar date for sessions older than a month — reusing and extending the shared `relativeAge` helper
- The detail line joins the search haystack, so filtering by project path or model name works
- Narrow terminals keep the existing single-line list; mouse clicks on either line of a row select the same session; detail rows halve the visible window so the overlay keeps its height

Wide terminal:

    ❯ 1d ago      Fix the auth token refresh loop
      /tmp/zerodemo2/work · gpt-5
      3d ago      branch: auth experiment
      /tmp/zerodemo2/work · claude-sonnet-4.5 · fork

#### Test plan

- [x] `go test ./internal/tui` — detail composition, status chips, wide/narrow rendering, filter-by-model, label alignment
- [x] `go test ./...` full suite green
- [x] `make fmt-check`, `go vet`, `git diff --check`, release build + smoke, `make vulncheck` green
- [x] Manual TUI check at 120 and 64 cols: filter, arrow selection, Esc close
- [ ] `make lint-static` — 4 pre-existing advisory findings in unrelated files (installtest, proxydial, web_fetch); none from this change

## Summary by CodeRabbit

* **New Features**
  * Session picker rows show relative activity times, with calendar dates for older sessions.
  * Wider terminals display project, model, and status details for each session.
  * Session search includes detail text.
  * Mouse selection works across all lines of multi-line session rows.
  * Picker layouts adapt to terminal width and row details.

* **Bug Fixes**
  * Session metadata is sanitized to preserve consistent row formatting.
  * Older session timestamps now use calendar dates instead of extended day counts.

### DIFF
--- a/internal/tui/command_center.go
+++ b/internal/tui/command_center.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	tea "charm.land/bubbletea/v2"
 
@@ -365,16 +366,25 @@ func (defaultModelSwitchCompactionPolicy) BeforeModelSwitch(request modelSwitchC
 
 var modelSwitchCompactionGuard modelSwitchCompactionPolicy = defaultModelSwitchCompactionPolicy{}
 
-// sanitizeCardField strips the card protocol's separator bytes and
-// line-breaking controls from user-controlled values (titles can legally
-// contain anything --session-title was given), so a hostile or accidental
-// \x1f / newline / carriage return cannot shift fields, corrupt row geometry,
-// or leak control characters into the transcript.
+// sanitizeCardField strips the card protocol's separator byte and every
+// control rune from user-controlled values (titles can legally contain
+// anything --session-title was given, and a directory name can legally
+// contain an ESC byte on Unix). Line separators and tabs become spaces so
+// words stay apart; the remaining controls — ESC/CSI/OSC initiators, BEL, BS,
+// VT, NEL, NUL — are dropped outright, since any of them can repaint the
+// terminal or shift cells when the value lands in a rendered row.
 func sanitizeCardField(value string) string {
-	value = strings.ReplaceAll(value, sessionsCardFieldSep, " ")
-	value = strings.ReplaceAll(value, "\n", " ")
-	value = strings.ReplaceAll(value, "\r", " ")
-	return strings.ReplaceAll(value, "\x00", "")
+	var out strings.Builder
+	for _, r := range value {
+		switch {
+		case r == '\x1f' || r == '\n' || r == '\r' || r == '\t':
+			out.WriteRune(' ')
+		case unicode.IsControl(r):
+		default:
+			out.WriteRune(r)
+		}
+	}
+	return out.String()
 }
 
 // relativeAge renders an RFC3339 timestamp as a short age ("2h ago"), falling
--- a/internal/tui/command_center.go
+++ b/internal/tui/command_center.go
@@ -375,7 +375,9 @@ func sanitizeCardField(value string) string {
 	return strings.ReplaceAll(value, "\x00", "")
 }
 
-// relativeAge renders an RFC3339 timestamp as a short age ("2h ago"); ""
+// relativeAge renders an RFC3339 timestamp as a short age ("2h ago"), falling
+// back to the month/day this year and the bare date for sessions older than a
+// month, so deep history reads as a calendar date rather than "400d ago"; ""
 // when the timestamp does not parse, so the card simply omits it.
 func relativeAge(timestamp string, now time.Time) string {
 	parsed, err := time.Parse(time.RFC3339, strings.TrimSpace(timestamp))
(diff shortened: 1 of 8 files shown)


Vasanthdev2004 (changes requested):
Nice change, and the structure is right: render and hit-testing both ask the same two helpers at the same width, so the two-line geometry cannot drift between what is drawn and what a click resolves to. One thing to fix before it goes in, and it is in the function this PR already edits.

**The detail line makes three new fields reachable, and the sanitizer only covers newlines.** On main, `sanitizeCardField` handles the field separator, `\n` and NUL, and the only metadata that reached a row was the title; `meta.Cwd` was read for workspace filtering and never rendered. This PR renders the project directory, the model id and the tag, so all three now pass through that same sanitizer, and everything except newline and NUL survives into the row. Driven through `/resume` at 120 columns:

```
ESC   in title    label  "Jun  5      alpha\x1b[31mred"
ESC   in model    detail "/repo · gpt\x1b[31m5"
ESC   in cwd      detail "/repo\x1b[31mx · gpt-5"
ESC   in tag      detail "/repo · gpt-5 · side\x1b[31mchat"
CSI   in cwd      detail "/repo\x1b[2J\x1b[H · gpt-5"
OSC   in title    label  "Jun  5      alpha\x1b]0;pwned\a"
BEL   in model    detail "/repo · gpt\a5"
BS    in title    label  "Jun  5      alpha\bbravo"
VT    in tag      detail "/repo · gpt-5 · a\vb"
NEL   in cwd      detail "/repox · gpt-5"
```

The escape in a title is pre-existing and not this PR's doing. The other three fields are new, and a directory name may legally contain an escape byte on Unix, so `\x1b[2J\x1b[H` in a project path repaints the user's terminal when the picker opens. Row geometry held in every case above, so this is spoofing and repainting rather than a hit-testing break, but the picker is exactly the surface where the user is being asked to choose between sessions by how they look.

The fix belongs in `sanitizeCardField` rather than at each call site: after the existing replacements, drop the remaining control runes the way `mcp_manager.go` and `sanitizeComposerPaste` already do in this package. That covers the label too, which closes the pre-existing half for free. Pin it with the escape spellings above rather than only `\n` and `\r`, since `TestResumePickerSanitizesMultilineMetadata` asserts on `"\n\r"` and would stay green through all ten rows.

**Everything else checked out.** `pickerShowsDetails` and `pickerMaxVisible` are both derived from `chatWidth(m.width)` in the render and the mouse path, so the tier decision and the halved budget agree; the per-item `item.Detail != ""` test is applied identically in both loops, so a row without a detail costs one line in both; the detail joins the search haystack; and `TestGenericPickerMouseSelectsItemFromDetailLine` covers the click that the two-line row newly makes possible. The carriage-return finding from the bots is already closed at this head. `internal/tui` is green here apart from the long-cwd title-bar test that fails on main on this machine, and CI is green on everything that has reported.

One note, not a change request: as soon as a single visible row carries a detail, the budget halves for the whole list, so a workspace where one session has a recorded model shows five rows where it used to show ten. That reads as intentional from the doc comment and it keeps the overlay height stable, so I am only flagging it in case the asymmetry was not deliberate.

Vasanthdev2004 (approved):
Approving at `48e347f3`. Fixed in the right place and pinned properly, and I re-drove the same eleven spellings rather than reading the diff.

`sanitizeCardField` now maps the separator, tabs and line breaks to a space and drops every remaining control rune, so the label gets the same treatment as the new detail line and the pre-existing escape-in-a-title hole closes with it. Through the real `/resume` picker at 120 columns, all eleven cases from my last round come back clean, including the four new sources:

```
ESC in title / model / cwd / tag     control runes: none
CSI ESC [ 2 J ESC [ H in cwd         none
OSC ESC ] 0 ; pwned BEL in title     none
BEL / BS / VT / NEL                  none
all of them in one session           none
```

The ESC byte is dropped and its remaining printable bytes show literally, which is the right outcome: visible junk that cannot repaint the terminal. Both halves are load-bearing rather than decorative. Disabling the control arm fails the sanitizer test on NUL and the picker test with `\x1b` still in the detail line; the new escape inputs are what catch it, where the old `"\n\r"` assertion would have stayed green.

You also found and fixed something I did not raise: `relativeAge` treated exactly thirty days as older than a month and fell through to the calendar date. Reverting the boundary back to `<` fails the case you added for it.

`gofmt` and `go vet` clean, and the eight tests this PR adds or changes all pass here, along with the rest of `internal/tui` apart from the long-cwd title-bar test that fails on main on this machine. CI is 11 of 12 at this head with the Windows smoke job still running; this is a TUI-only change and I ran that package natively on Windows, so I am not waiting on it to say the change is right.

One thing I am deliberately not asking for, to be consistent with the call I made on #953: `unicode.IsControl` does not cover the Cf class, so ZWSP, RLO, soft hyphen, word joiner and BOM still reach a row and an RLO can reverse how a title reads. That is the Trojan Source class, it is pre-existing on every surface in the tree rather than anything this PR introduced, and the bidi question deserves one decision across all of them rather than being settled quietly in a picker. This function is the natural home for it whenever someone takes that on.

Vasanthdev2004:
@coderabbitai full review

kevincodex1 (approved):

Issue #1028, opened 2026-09-08
oauth: slow loopback clients survive bounded shutdown


PierrunoYT:
## Version / branch / commit

`main` at `1b5db1765672820caac1684b168c9898b5ba3593`.

## OS and environment

Linux x86-64, kernel 6.1, official Go 1.26.6.

## Steps to reproduce

The shared `internal/oauth` loopback listener exposes the lifecycle directly enough for a deterministic package test:

```go
func TestSlowHeaderSurvivesLoopbackClose(t *testing.T) {
    listener, err := NewLoopbackListener("audit-state")
    if err != nil { t.Fatal(err) }
    conn, err := net.Dial("tcp", listener.listener.Addr().String())
    if err != nil { t.Fatal(err) }
    defer conn.Close()

    fmt.Fprintf(conn,
        "GET /callback?code=audit-code&state=audit-state HTTP/1.1\r\nHost: %s\r\nX-Stall:",
        listener.listener.Addr().String())
    time.Sleep(100 * time.Millisecond) // let Serve accept and read the partial header

    started := time.Now()
    listener.Close()
    if elapsed := time.Since(started); elapsed < 900*time.Millisecond {
        t.Fatalf("Close returned after %s; expected its one-second deadline", elapsed)
    }

    conn.SetReadDeadline(time.Now().Add(100 * time.Millisecond))
    var one [1]byte
    _, err = conn.Read(one[:])
    if netErr, ok := err.(net.Error); !ok || !netErr.Timeout() {
        t.Fatalf("read after Close = %v; wanted timeout proving the connection remains open", err)
    }
}
```

Focused reproduction:

```text
go test ./internal/oauth -run '^TestSlowHeaderSurvivesLoopbackClose$' -count=1 -v
--- PASS: TestSlowHeaderSurvivesLoopbackClose (1.20s)
PASS
```

The temporary test was removed after verification and was not committed.

The same construction appears in:

- `internal/oauth/loopback.go` around lines 53-54 and 102-106;
- `internal/mcp/oauth.go` around lines 497-520; and
- `internal/provideroauth/openrouter.go` around lines 82-106.

Each starts `http.Server.Serve` in a goroutine, configures no `ReadHeaderTimeout`/`ReadTimeout`/`IdleTimeout`, ignores the Serve result, calls `Shutdown` with a one-second context, ignores that result, and does not join Serve completion.

## Expected behavior

[...]

PR #1051, opened 2026-09-12
fix(oauth): force-close loopback callback connections past the shutdown deadline

pujitha24:
## Summary

`LoopbackListener.Close` (used by every OAuth loopback callback flow) calls
`http.Server.Shutdown` with a 1-second context and discards the result.
`Shutdown` only closes *idle* connections; a client that has sent a partial
request header (and never finishes it) counts as active, so `Shutdown` blocks
out the full second, times out, and `Close` returns anyway — leaving that
connection and its `Serve` goroutine open indefinitely. This fixes
`internal/oauth/loopback.go` (the file the issue's reproduction test
exercises) by:

- adding a `ReadHeaderTimeout` to the server so a stalled header is bounded on
  its own,
- following a timed-out/failed `Shutdown` with `server.Close()` to forcibly
  tear down any connection still open past the deadline, and
- joining the `Serve` goroutine before `Close` returns, so its completion is
  observable instead of left dangling.

This is a loopback-only (127.0.0.1), short-lived, single-use server — the bug
is a resource leak (a dangling goroutine and open socket past the end of the
flow), not a remote vulnerability, crash, or outage.

`internal/mcp/oauth.go` (~line 497-520) and `internal/provideroauth/openrouter.go`
(~line 82-106) construct their own loopback servers with the same pattern, as
noted in the issue, and still drop `Shutdown`'s result with no
`ReadHeaderTimeout` or force-close fallback. They are intentionally left
untouched here to keep this PR focused on the file the approved issue's
reproduction demonstrates; the same fix there is left as follow-up work.

## Linked issue

Refs #1028 — the issue covers three loopback server sites
(`internal/oauth/loopback.go`, `internal/mcp/oauth.go`,
`internal/provideroauth/openrouter.go`); this PR only fixes the first, so it
doesn't close the issue on its own. Using `Refs` per CONTRIBUTING's "another
clear link" allowance rather than `Fixes`.

## Validation

- `go build ./internal/oauth/...` and `go vet ./internal/oauth/...` — clean.
- `gofmt -l internal/oauth/` — clean (`make fmt-check` is clean repo-wide).
- `go test ./internal/oauth/... -race -v -count=1` — all pass, including three
  new regression tests: `TestLoopbackCloseForceClosesSlowHeaderConnection`,
  `TestLoopbackCloseIsIdempotent`, `TestLoopbackCloseReturnsPromptlyAfterSuccessfulCallback`.
- Confirmed `TestLoopbackCloseForceClosesSlowHeaderConnection` reproduces the
  defect when only the `server.Close()` fallback is reverted (fix otherwise
  in place): it fails with the same "read after Close timed out" symptom the
  issue describes. The test now calls the `newLoopbackListener` constructor
  introduced by this fix, so it can no longer run unmodified against
  pre-fix `main` — reverting the fallback alone is the working reproduction.
- Full repo `go build ./...`, `go vet ./...`, and `go test ./...` pass.
- Checked both existing callers (`internal/oauth/manager.go`,
  `internal/provideroauth/chatgpt.go`); both only `defer listener.Close()` and
  do not depend on `Close`'s prior blocking behavior.
- Could not run `make lint-static` (golangci-lint) locally: it needs to
  download a very large dependency tree and the local environment ran out of
  disk space fetching it. This check is `continue-on-error: true` (advisory)
  in `.github/workflows/ci.yml`, not a required gate.

## Checklist

- [x] `gofmt` clean.
- [ ] UI changes include screenshots or a short recording where possible. (not applicable — no UI change)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth callback listener shutdown reliability.
  - Stalled connections are closed after the shutdown deadline.
  - Shutdown now waits for the listener to fully stop.
  - Repeated shutdown calls are handled safely.
  - Completed callbacks return promptly during shutdown.
  - Slow or incomplete request headers no longer delay the listener indefinitely.
  - Shutdown behavior is more consistent when callbacks are still being processed.

### DIFF
--- a/internal/oauth/loopback.go
+++ b/internal/oauth/loopback.go
@@ -20,6 +20,7 @@ type LoopbackListener struct {
 	state    string
 	result   chan callbackResult
 	server   *http.Server
+	done     chan struct{}
 }
 
 type callbackResult struct {
@@ -49,9 +50,19 @@ func NewLoopbackListenerOnPort(state string, port int) (*LoopbackListener, error
 		listener: ln,
 		state:    state,
 		result:   make(chan callbackResult, 1),
+		done:     make(chan struct{}),
 	}
-	l.server = &http.Server{Handler: http.HandlerFunc(l.handle)}
-	go func() { _ = l.server.Serve(ln) }()
+	l.server = &http.Server{
+		Handler: http.HandlerFunc(l.handle),
+		// This is a single-use, loopback-only callback server: a client that
+		// never finishes sending its request header must not be able to hold
+		// the accepted connection (and Close, below) open indefinitely.
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	go func() {
+		_ = l.server.Serve(ln)
+		close(l.done)
+	}()
 	return l, nil
 }
 
@@ -99,11 +110,17 @@ func (l *LoopbackListener) Wait(ctx context.Context) (string, error) {
 	}
 }
 
-// Close shuts the listener down (bounded), idempotent.
+// Close shuts the listener down (bounded), idempotent. If a client leaves a
+// connection open past the shutdown deadline (e.g. a stalled request header),
+// Close forcibly closes it rather than leaking it past the end of the flow,
+// and waits for the Serve goroutine to finish before returning.
 func (l *LoopbackListener) Close() {
 	shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-	_ = l.server.Shutdown(shutdownCtx)
+	if err := l.server.Shutdown(shutdownCtx); err != nil {
+		_ = l.server.Close()
+	}
+	<-l.done
 }
 
 // parseCallback validates the redirect query and returns the authorization code,
--- a/internal/oauth/loopback_test.go
+++ b/internal/oauth/loopback_test.go
@@ -3,6 +3,8 @@ package oauth
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


Vasanthdev2004 (dismissed):
Approving. Both halves of what #1028 asked for are here, and you added a third thing I had not thought of.

`Close` now falls back to `server.Close()` when the bounded `Shutdown` does not finish, which is the actual fix, and `ReadHeaderTimeout` bounds the stalled-header case on its own so the connection does not depend on somebody calling `Close` at all. Waiting on the serve goroutine through the new `done` channel was not in the issue and is the right call: without it `Close` can return while `Serve` is still unwinding, which is exactly the kind of thing that turns into a flaky test later.

Driven rather than read. All three tests pass, and dropping the `server.Close()` fallback fails the force-close test with the symptom the issue described:

```
read after Close timed out (i/o timeout); wanted the stalled connection closed rather than left open
```

I especially like that you covered the regression risk rather than only the fix: `TestLoopbackCloseReturnsPromptlyAfterSuccessfulCallback` is what stops someone later making `Close` unconditionally wait out the full second on the normal path, and the idempotency test covers the double-close that the doc comment promises.

One note, not blocking, and it is about coverage rather than correctness. `ReadHeaderTimeout` is the one line here with no regression behind it: removing it leaves all three tests green, because the force-close closes the connection at the one-second shutdown deadline long before the five-second header timeout would. An unpinned line is the one that quietly disappears in a later refactor. If you want it pinned cheaply, make the timeout a package-level variable, point it at a short duration in a test, and assert the connection closes on its own with no `Close` call at all. Fine to leave for a follow-up; the defect the issue reported is pinned either way.

CI is 10 of 10 after I released the fork gate.

pujitha24:
Fixed the timing issue Greptile and Copilot both flagged on `TestLoopbackCloseForceClosesSlowHeaderConnection`: it now waits on a `ConnState` hook for `StateNew` instead of a fixed 100ms sleep, so it's synchronized to the server actually accepting the connection rather than guessing. Also joined the `http.Get` goroutine (and drained/closed its response body) before timing `Close` in `TestLoopbackCloseReturnsPromptlyAfterSuccessfulCallback`, and added `t.Cleanup(l.Close)` to both new tests so a `Fatalf` mid-test doesn't leak the `Serve` goroutine. All three regression tests plus the full package pass under `-race`.

@Vasanthdev2004 thanks for the review — agreed on the `ReadHeaderTimeout` coverage gap, leaving that as the follow-up you suggested since the force-close path already pins the defect from #1028.

Vasanthdev2004 (approved):
Re-approving at `6d3dcb79`. The push dismissed my approval, so this is a fresh pass; the fix itself is byte for byte what I approved at `8f91a838`, and the new commit is the test synchronization plus a nine line constructor split.

The thing worth checking when a sleep gets replaced by a channel is whether the test still has teeth, so I drove that rather than reading it. Reverting only the `server.Close()` fallback and leaving everything else in place still fails it, with the same line as before the rework:

```
read after Close timed out (read tcp 127.0.0.1:63233->127.0.0.1:63232: i/o timeout); wanted the stalled connection closed rather than left open
```

So the `StateNew` wait is a strict improvement over the 100ms sleep and not a way around the assertion. It is also the right edge to wait on: `StateNew` is set inside `Serve` before the conn goroutine starts reading, and `Shutdown` only closes conns in `StateIdle`, so by the time the test writes the partial header the connection is already in a state `Shutdown` cannot reap.

Stability, since that was the point of the change: 32 runs of the three close tests under `-race`, four processes at once to keep the box busy, all green, with the force close sitting at exactly 1.00s every time.

`t.Cleanup(l.Close)` means `Close` now runs twice in the force close test, so I measured that too rather than trusting the idempotency test, which only covers a listener nothing ever connected to. First call 1.0003s, second 0s, third 0s, goroutines 4 down to 2. The second `Shutdown` is free because the first one already waited out `Serve` and the listener was untracked when it returned.

On the `ConnState` seam: the hook is nil on both exported constructors and `net/http` reads the field exactly once, behind `if hook := srv.ConnState; hook != nil`, so production is unchanged. Adding an unexported constructor for this is the right call over exporting anything.

CI is 9 of 9 after I released the fork gate, which I was comfortable doing because the whole PR is two files under `internal/oauth` and touches no workflow or dependency files.

**One thing to sort before this merges, and it is about the link rather than the code.** #1028 names three servers, not one, and asks for those tests on all three. `internal/mcp/oauth.go` around 497 to 520 and `internal/provideroauth/openrouter.go` around 82 to 106 still do `_ = server.Shutdown(shutdownCtx)` with the result dropped, no `ReadHeaderTimeout`, no force close, and no join on `Serve`. I confirmed both still leak rather than assuming it: driving `OpenRouterLogin` with a 300ms deadline and a client that stalls its header returns after 1.30s, a full second of that being the dead `Shutdown` wait, and the connection is still readable afterwards. `defer listener.Close()` does not save it, since closing a listener never touches connections already accepted.

Keeping this PR to the one file is the right scope call and I said so last time. The problem is only that `Fixes #1028` will close the audit finding when this lands, with two of its three sites untouched. CONTRIBUTING allows "another clear link", so switching that line to `Refs #1028` and opening a follow up for the other two keeps both the focus and the finding. Happy to file that follow up myself if you would rather not.

Small note while you are in the description: the validation bullet that says the regression fails against the pre-fix `loopback.go` with the new test file kept cannot be run that way any more, because the test now calls `newLoopbackListener`, which the fix introduces. Against unmodified `main` it is `undefined: newLoopbackListener` rather than a failure. Reverting just the force close fallback is the reproduction that still works, and it does.

kevincodex1 (approved):
lgtm

PR #1052, opened 2026-09-12
ci: disable persisted checkout credentials on release workflows

### LINKED ISSUE #1029: ci: release checkouts retain unnecessary Git credentials
## Version / branch / commit

`main` at `1b5db1765672820caac1684b168c9898b5ba3593`.

## OS and environment

Workflow configuration issue; platform-independent. Reviewed from `.github/workflows/release-artifacts.yml` and `.github/workflows/ci.yml`.

## Steps to verify

1. Inspect `.github/workflows/ci.yml` checkout steps around lines 20-26: they explicitly set `persist-credentials: false`.
2. Inspect the package job's ordinary/requested-ref checkout steps in `.github/workflows/release-artifacts.yml` around lines 33-43.
3. Inspect the npm publish job checkout around lines 126-133.
4. None of the three release checkout paths sets `persist-credentials: false`.
5. Search the release workflow for an actual `git` command that requires checkout's persisted credential. No such command is present. GitHub Release publication instead supplies `GH_TOKEN: ${{ github.token }}` directly to its publishing step.

`actions/checkout` therefore retains its authentication configuration for later steps by default. In the package matrix the job has `contents: write`, so trusted tests and build/package scripts receive avoidable access to that credential.

## Expected behavior

Release checkout credentials should not remain available to subsequent build and test steps unless a documented later Git operation requires them. Publication should continue using explicit, step-scoped credentials.

## Actual behavior

All release checkout variants retain checkout's default persisted Git credential. This is inconsistent with the repository's CI and action-smoke workflows, which already disable persistence.

No credential exfiltration or malicious current step was found. This is a low-severity workflow defense-in-depth issue, not a demonstrated supply-chain compromise.

## Suggested fix and verification

Add:

```yaml
with:
  persist-credentials: false
```

to the normal package checkout, requested-ref checkout (alongside `ref`), and npm-publish checkout. Then run workflow syntax validation and the normal package/release dry-run or native release checks. Preserve existing least-privilege job permissions and explicit `GH_TOKEN`/OIDC publication paths.

Reported from codebase audit finding `SEC-09`. PR #951 is related workflow hardening but does not add this checkout setting.

pujitha24:
## Summary

Sets `persist-credentials: false` on the `actions/checkout` steps in
`.github/workflows/release-artifacts.yml` (the ordinary checkout, the
workflow-dispatch requested-ref checkout, and the npm-publish job checkout)
and `.github/workflows/pr-auto-review.yml`, matching the convention already
used in `ci.yml` and `zero-action-smoke.yml`. Without this, the workflow
token stays in the local git config for every later step in the job — build,
test, and package steps that have no reason to hold it. None of the later
steps in either file run `git push`/`git commit`/`git tag`; the GitHub
Release publish step already uses its own explicit `GH_TOKEN`, and npm
publishing uses OIDC trusted publishing, so removing the persisted
credential does not affect any existing functionality.

This is a defense-in-depth hardening change, not a fix for a demonstrated
credential leak or exploit.

## Linked issue

Fixes #

## Checklist

- [x] `gofmt` clean.

Notes:
- No test was added because this is a pure workflow-config change; the
  repo's existing `internal/installtest/workflow_permissions_test.go` only
  checks `permissions: contents: read`, not `persist-credentials`, and
  adding one was out of scope for this minimal fix.
- Not a UI change.
- Validated with `go build ./...`, `go vet ./...`, `gofmt -l .`,
  `go test ./internal/installtest/...`, and `actionlint` on both changed
  files (the only findings were two pre-existing shellcheck warnings in
  `release-artifacts.yml` unrelated to this diff, confirmed present before
  the change as well). A live GitHub Actions run of the release/publish
  workflows was not possible locally; this change only removes a default
  the workflows never use, so no other behavior should change.

Fixes #1029

## Summary by CodeRabbit

- **Security**
  - Improved workflow security by preventing checkout steps from persisting authentication credentials in local Git configuration.
  - Applied this protection across review and release artifact workflows.

### DIFF
--- a/.github/workflows/pr-auto-review.yml
+++ b/.github/workflows/pr-auto-review.yml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Setup Go
         uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -34,12 +34,15 @@ jobs:
       - name: Checkout
         if: ${{ github.event_name != 'workflow_dispatch' || inputs.ref == '' }}
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
 
       - name: Checkout requested ref
         if: ${{ github.event_name == 'workflow_dispatch' && inputs.ref != '' }}
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           ref: ${{ inputs.ref }}
+          persist-credentials: false
 
       - name: Setup Go
         uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff
@@ -129,6 +132,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
 
       # No registry-url here on purpose: setup-node's registry-url writes an
       # .npmrc line referencing $NODE_AUTH_TOKEN, and npm hard-fails on the


Vasanthdev2004 (approved):
Approving. This is the whole of #1029 and nothing besides: `persist-credentials: false` on the three `release-artifacts.yml` checkouts and the one in `pr-auto-review.yml`, which is the extra one I asked for in the same change because it had the same gap.

I checked the shape rather than just the diff, since a fork PR touching `.github/workflows` deserves it. Neither workflow uses `pull_request_target`: `pr-auto-review` is a plain `pull_request`, so a fork run gets no secrets and a read-only token, and `release-artifacts` is `workflow_dispatch` plus a tag push, which a pull request cannot trigger at all. And the change only ever removes a credential from a checkout, so it strictly reduces what a run can do. That combination is why I was comfortable releasing the gated CI on it.

The modified `pr-auto-review` workflow then ran against this very pull request and passed, so the edit is exercised rather than merely parsed. Nothing in either job pushes, so dropping the stored credential costs nothing.

kevincodex1 (approved):

PR #1063, opened 2026-09-17
fix(tools): hint when Windows resolves POSIX paths under root

### LINKED ISSUE #972: fix(tools): normalize or guide models against absolute POSIX paths on Windows hosts
### Summary

When models run on Windows hosts, they occasionally hallucinate standard POSIX paths (such as `/home/user/...` or `/tmp/...`) in arguments to `read_file`, `glob`, and `grep`. These paths fail with raw OS-level errors (e.g. `GetFileAttributesEx <workdir>\home: The system cannot find the file specified`).

### Problem

- In `internal/tools/`, path resolution on Windows interprets leading `/home/...` relative to the current working drive or repository root, causing missing-path errors.
- The model wastes turns retrying non-existent directories rather than adjusting to workspace-relative or Windows paths.

### Proposed Solution

1. **Platform hints on failure**: When a tool on Windows receives a `/home/...` or root-like POSIX path that does not exist, provide an actionable error hint clarifying the workspace root and host OS.
2. **Path normalization**: Strip synthetic `/home/user/<repo>` prefixes when they match workspace-relative targets.

lhk0504:
Fixes #972

## Summary

On Windows hosts a leading-slash POSIX path (`/home/user/repo/x`, `/tmp/x`) is not absolute (`filepath.IsAbs` is false without a volume name), so read tools joined it under the workspace root and surfaced a raw `GetFileAttributesEx <root>\home: The system cannot find the file specified`. Models then retried the same wrong shape instead of switching to a workspace-relative path.

- Adds `looksLikePosixAbsolutePath` and `annotateMissingPosixPathError` in `internal/tools/path_hint.go`. On Windows only, when a POSIX-looking read path reports `fs.ErrNotExist`, the error gains a `[zero] path hint:` block naming the host OS and the workspace root, plus a `Suggestion:` to use a workspace-relative or Windows absolute path. The original error is preserved via `%w`, so `errors.Is/As` keep working.
- Hooks the shared read resolver (`resolveScopedReadPath`) through a `resolveScopedReadPathForGOOS` test seam (same pattern as `detectShellRuntime(goos)`), so `read_file`, `glob`, `grep`, `read_minified_file`, `list_directory`, `view_image`, and `lsp_navigate` all get the hint.

**No path rewriting**, per the constraint in the approved issue discussion. The hint never changes which path is resolved or opened. Write-path resolution, shell cwd resolution, confinement errors (`outsideWorkspaceError`), and sandbox behavior are unchanged. Write tools are deliberately untouched: a missing target is normal for them, so a missing-path hint would be wrong advice there.

Rendered example on Windows:

```
Error reading file /home/user/repo/notes.txt: GetFileAttributesEx C:\Users\me\repo\home: The system cannot find the file specified.
[zero] path hint: this host is Windows, so the POSIX path was resolved under the workspace root "C:\Users\me\repo" and does not exist there.
Suggestion: use a workspace-relative path such as "internal/tools/workspace.go", or a Windows absolute path under that workspace root.
```

## Tests

`internal/tools/path_hint_test.go`:
- shape table (`/home/x`, `/tmp/x`, `/` vs `//UNC`, drive-letter, backslash-rooted, relative, empty)
- annotation table: Windows + POSIX + not-exist wraps and stays `errors.Is(_, fs.ErrNotExist)`; Linux/Darwin pass through unchanged; non-POSIX shapes and non-missing errors pass through unchanged
- `resolveScopedReadPathForGOOS("windows", ...)` wiring assertion, runnable on every platform
- Windows-only `read_file` end-to-end test (skips elsewhere)

The wiring test fails without the change (annotation call disabled):

```
--- FAIL: TestResolveScopedReadPathForGOOSAnnotatesWindowsPosixMiss (0.00s)
    path_hint_test.go:152: expected "[zero] path hint:" in "lstat /home/zero-hint-missing: no such file or directory"
```

Verification performed:
- `make fmt-check`, `go vet ./...`, `GOOS=windows go build ./...`, `GOOS=windows go vet ./internal/tools/`
- `go test -race ./internal/tools/` passes; `go test ./...` has `internal/tools` passing. Two unrelated environment failures on this machine: `internal/daemon/remote` flaked once and passed on rerun, and `internal/provideroauth` fails only because loopback port 1455 is occupied by an unrelated local process.
- `go run ./cmd/zero-release build` and `smoke` pass; `make vulncheck` finds nothing; `git diff HEAD --check` clean
- `make lint-static` reports 4 pre-existing advisory findings (`internal/installtest`, `internal/proxydial`, `internal/tools/web_fetch.go`); none are introduced by this change

No UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows users now receive a clearer hint when a requested file is missing because it was specified with a POSIX-style absolute path.
  * Existing error details remain available, including the original missing-file status.
  * Valid paths, relative paths, non-Windows systems, and access or workspace-boundary errors retain their existing behavior.
  * Troubleshooting information now identifies the relevant workspace context when a file cannot be found.


Vasanthdev2004 (dismissed):
Approving at `6c82a013`. This is the hint half of #972 and nothing else, which is what the issue was approved for: no path is rewritten, the original error is wrapped with `%w`, and the write tools are left alone.

I ran it natively on Windows, which is the only place the end-to-end test does anything. Every tool on the shared read resolver gets the hint for a missing POSIX-shaped path: `read_file`, `list_directory`, `glob`, `grep`, `read_minified_file` and `view_image`. The controls stay clean: a relative miss, a Windows absolute miss, a UNC path and a backslash-rooted path get no hint, and a POSIX-shaped path that does exist under the workspace root (`/home/user/real.txt`, with that tree in the workspace) is read normally. Taking the annotation out of the resolver fails both `TestResolveScopedReadPathForGOOSAnnotatesWindowsPosixMiss` and `TestReadFileWindowsPosixPathHint`. The tools package is green here, vet and gofmt are clean, CI 9 of 9.

- The suggestion names `internal/tools/workspace.go` as its example. That is a file in this repository, and the text is shown to a model working in somebody else's, where an over-literal reader may go looking for it. A neutral placeholder, or the tail of the path that was asked for, would be safer.
- `/home/../../etc/passwd` gets the hint as well. The stat behind that error was outside the root (`...\etc`), so "resolved under the workspace root" is not quite true there. Harmless, and the order of the stat and the confinement check is not this PR's to fix.

Thank you for picking this up, and for keeping to the constraint on the issue.

jatmn (changes requested):

## Merge readiness

- The branch is current with `main` (`99721c762f37cd43ac511007a5f51d1846df959e`), GitHub reports it mergeable with no conflicts, and all ten captured checks pass. GitHub nevertheless reports `REVIEW_REQUIRED`, so a qualifying approval is still required after the code finding is fixed.

## Findings

- [P2] Keep the cross-platform wiring test inside test-owned storage
  `internal/tools/path_hint_test.go:143`
  Attribution: PR-introduced. The test and its fixed `/home/zero-hint-missing/missing.txt` request are absent at merge base and live `main`; this PR adds both.
  Stated contract: `AGENTS.md` says, “Hermetic tests must not read or write the developer's real config, cache, or state directories,” and requires path-sensitive non-Linux behavior to use a native case or a hermetic fake that exercises the same normalization.
  Root cause: `resolveScopedReadPathForGOOS("windows", ...)` injects only the annotation GOOS, while `filepath.IsAbs`, `EvalSymlinks`, and joining still use the host implementation. On Linux and macOS, the fixed `/home/...` request is therefore a real host-global absolute path and `root := t.TempDir()` is ignored for resolution.
  What fails: the test passes here only because that exact global path is absent. If it exists on a developer or CI host, resolution can succeed or return a different error and the regression fails for host state rather than behavior. Its assertion that the error contains `root` also comes from the hint text, not proof that the path was resolved beneath the temporary workspace.
  - `TestResolveScopedReadPathForGOOSAnnotatesWindowsPosixMiss` — non-hermetic on Unix hosts.
  - `TestReadFileWindowsPosixPathHint` — already hermetic and authoritative on native Windows; preserve it.
  Required correction: make the wiring test's missing POSIX-shaped request resolve only against test-owned temporary state on each host, or use an equivalently hermetic path-semantics seam. Keep the native-Windows end-to-end test, and keep production resolution behavior unchanged.
  Author fix: close the root cause in the PR-added wiring test in one pass; do not patch the fixed username with another global pathname, and do not redesign the shared resolver or add path rewriting.
  Out of scope: production path selection, write tools, shell cwd resolution, sandbox policy, and the normalization half of issue #972.

jatmn (changes requested):
## Findings

- [P3] Do not classify mixed-separator UNC paths as POSIX paths
  `internal/tools/path_hint.go:14`
  Attribution: PR-introduced. Merge-base/live `main` has no path-shape classifier or workspace hint; this PR adds both.
  Stated contract: the new helper's comment says UNC paths are not the POSIX paths this hint diagnoses. On Windows, both `/` and `\` are path separators, so two mixed leading separators form the same UNC/device-style prefix as two identical separators.
  Root cause: `looksLikePosixAbsolutePath` excludes only the literal byte prefix `//`. It therefore accepts valid Windows forms such as `/\server\share\missing` and `/\?\C:\missing` even though Windows treats them as absolute UNC/device-style paths.
  What fails: when one of those targets is missing on the common `scope == nil` path, the new annotation falsely says the path was POSIX-shaped and resolved under the workspace, directing the caller away from the network or device target it actually requested.
  - the PR-added classifier must reject UNC/device prefixes that use either Windows separator in either leading position
  - the PR-added shape and annotation tests must cover those mixed-separator forms
  Unchanged on main (do not edit in this PR): path selection, resolution ordering, confinement policy, and sandbox scope behavior.
  Required correction: make the new classifier recognize both Windows separator characters when excluding UNC/device-style prefixes, preserve the raw missing-path error for those inputs, and add focused regression rows for both mixed forms.
  Author fix: close the Root cause in the PR-added classifier and tests. Do not redesign the shared resolver or change which path is opened.
  Out of scope: workspace-root lifecycle errors, lexical traversal behavior, path normalization, write tools, shell cwd, and sandbox grants.

lhk0504:
@jatmn Both findings are addressed in the branch.

- **P2 (non-hermetic wiring test)** — `32d2a3c`. The request is now built from `t.TempDir()` per host: on Windows it is `"/home/..."` (joined under root, the behavior under test), on Unix it is anchored in root (`filepath.ToSlash(root) + "/home/..."`). A positive case resolves an existing `home/zero-hint-real/real.txt` under root and asserts the workspace-relative result, so the test proves resolution stayed in test-owned storage instead of matching hint text only. The native Windows end-to-end test is unchanged. With the annotation removed, the failure now names the temp tree:
  `expected "[zero] path hint:" in "lstat /tmp/TestResolveScopedReadPathForGOOSAnnotatesWindowsPosixMiss66217469/001/home/zero-hint-missing: no such file or directory"`

- **P3 (mixed-separator UNC/device paths)** — `17d4050`. `looksLikePosixAbsolutePath` now rejects a second leading separator in either position, so `/\server\share\...` and `/\?\C:\...` keep the raw missing-path error. The regression rows fail before the fix:
  `looksLikePosixAbsolutePath("/\\server\\share\\notes.txt") = true, want false`
  `looksLikePosixAbsolutePath("/\\?\\C:\\notes.txt") = true, want false`

Production resolution, path selection, confinement, and sandbox scope are untouched. Validation on this branch: `go test ./internal/tools/`, `go test -race ./internal/tools/`, `make fmt-check`, `go vet ./...`, `GOOS=windows go vet ./internal/tools/`, `git diff HEAD --check`.

One process note: the new commit's CI and PR Auto Review runs are in `action_required`, so a maintainer needs to approve the workflow before they run. Could you take another look when you have a moment?

Vasanthdev2004 (approved):
Approving again at `17d40504`. Both of jatmn's findings are fixed, and the second is one I should have caught myself: last time I tried a UNC path and a backslash-rooted path and never tried the two mixed.

Through the real `read_file` on Windows, the classifier I approved against this one:

```
requested                            6c82a013     17d40504
/home/user/missing.txt               hint         hint
//server/share/missing.txt           no hint      no hint
/\server\share\missing.txt           hint         no hint
/\server/share/missing.txt           hint         no hint
/\?\C:\zz-missing\missing.txt        hint         no hint
\/server\share\missing.txt           no hint      no hint
```

The three mixed forms really are UNC and device paths to Windows. The stat goes to `\\server\share` and `\\?\C:`, so the hint was pointing away from the target that was asked for. Putting the old classifier back under the new tests fails both new shape rows and both new annotation rows. Nothing else moved: `/c/Users/...`, `/./x`, a name with a space and a bare `/` behave as before.

The wiring test is hermetic now. On Unix the request is built under the test's own root, and the new first half asserts the resolved relative path, so it proves where resolution went instead of matching the hint text.

The tools package passes natively here, vet and gofmt are clean, CI 9 of 9 at head. I approved the fork's CI runs after reading the diff: three Go files under `internal/tools`, nothing under `.github` and no dependency changes.

My two small notes from last time still apply and are still not blocking.

jatmn (approved):
LGTM

PR #1066, opened 2026-09-21
fix(security): harden shell flag parsing, write roots, and formatters

hazyhaar:
### Summary of Changes

This pull request provides security hardening across sandbox command inspection, scoped file write operations, and external code formatters:

#### 1. Sandbox: Recognize POSIX short-option clusters in `shellCommandFlag`
- **Problem:** When shells like `bash` or `sh` are invoked with clustered flags such as `-ec`, `-lc`, or `-xc`, `shellCommandFlag` previously looked only for exact `-c` or `--command` matches. This caused subcommands in clustered invocations to bypass AST parsing and fallback to a default classification.
- **Fix:** Enhanced `shellCommandFlag` to recognize combined short-option clusters where `c` is the final option character (e.g., `-ec`, `-lc`, `-xc`), correctly extracting the command string for recursive AST inspection.

#### 2. Tools: Anchor file write operations to `*os.Root` descriptors
- **Problem:** `write_file` and `edit_file` previously validated path boundaries at start, but parent directories were resolved sequentially via standard OS paths, leaving a residual window for directory link manipulations.
- **Fix:** Implemented `openScopedWriteRoot` and `commitRootedFileContents` using Go's `*os.Root` API to anchor path resolution strictly relative to the allowed write root handle during temporary file creation, parent directory traversal, and atomic replacement.

#### 3. Tools: Harden `format_on_write` against untrusted execution and environment leaks
- **Problem:** When automatic formatting was enabled on write, external tools could evaluate untrusted configuration files in the workspace (e.g., `.prettierrc.js`), execute untrusted binaries present in the target repository, or inherit sensitive environment variables.
- **Fix:**
  - Passed `--no-config` to `prettier` to prevent execution of dynamic config scripts.
  - Added `formatterBinaryInWriteRoots` check to reject formatter binaries resolved from within writable workspace directories.
  - Scrubbed sensitive credential-bearing environment variables (`ScrubSensitiveEnv`) before spawning formatter processes.

### Testing & Verification
- Comprehensive unit test suites added and updated in `internal/sandbox` and `internal/tools`.
- Full suite validated locally: **14/14 PASS** with Go race detector enabled (`go test -race`).

## Summary by CodeRabbit

- **Security**
  - File edits and new file creation are now protected against directory symlink swaps that could redirect changes outside approved write locations.
  - Formatting tools no longer inherit sensitive credentials, execute project configuration code, or run formatter binaries located inside the workspace.

- **Bug Fixes**
  - Shell commands using clustered options such as `sh -ec` or `bash -lc` are now analyzed correctly, improving detection of interactive, network, and destructive commands.
  - Shell arguments after a standalone `--` are now handled as positional operands rather than command flags.


jatmn (changes requested):

## Merge readiness

- **Human review required:** GitHub reports `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED` even though required CI checks (including race detector and Security & Code Health) are green on head `6c9abec0eb91a5cf072e2b8b7c957edc734873de`. A maintainer approval is still needed before merge.
- **Branch freshness:** Head merge-base matches live `origin/main` (`99721c762f37cd43ac511007a5f51d1846df959e`); no stale-base rebase needed at review time.
- **CodeRabbit advisory:** Pre-merge docstring-coverage warning (65.71% vs 80% threshold on touched functions) — advisory only, not a failing CI check.
- **Prior automated review:** CodeRabbit requested `--` handling and root identity verification; those landed in commits `2a8c4f4` and follow-ups. CodeRabbit approved on head.

## Findings

### 🟠 P1 — `write_file` must observe preimages through the scoped `*os.Root`, not pathname re-resolution

📍 **Where:** `internal/tools/write_file.go` — `RunWithOptions` after `openScopedWriteRoot` (~L68–131).

💥 **What fails:** The tool opens a descriptor-bound write root, then still learns file existence, conflict bytes, and `expectedContent` via pathname APIs on `absolutePath`:

```go
root, rootedRelative, err := openScopedWriteRoot(...)
// ...
if info, err := os.Stat(absolutePath); err == nil { ... }
current, rerr := tool.readFile(absolutePath)
// ...
if prev, rerr := tool.readFile(absolutePath); rerr == nil { ... }
if err := commitRootedFileContents(root, absolutePath, rootedRelative, priorInfo, expectedContent, content); err != nil {
```

A parent symlink swap or path rebind after the root open can make those reads/`FileInfo` refer to a different object than the one `commitRootedFileContents` opens under `root`, undermining the TOCTOU closure this PR adds for writes.

🔎 **Root cause:** ZERO-TRAV-01 anchored the commit path on `*os.Root` but left overwrite observation on the old pathname path. The sibling `edit_file` change in this PR already anchors **read and write** on the same root handle.

📜 **Stated contract:**

> Anchor every mutation below on the granted write root. … so a parent directory swapped for an escaping symlink after validation cannot redirect the create or overwrite.

> Anchor both the read and the write on the granted write root: bytes and identity come from the same descriptor-bound object

🏷️ **Attribution:** PR-introduced (merge-base used pathname I/O consistently; head introduces a rooted commit with unrooted stat/read).

📌 **In this PR:**

- `internal/tools/write_file.go` — `os.Stat(absolutePath)`, `tool.readFile(absolutePath)` after `openScopedWriteRoot`
- `internal/tools/edit_file.go` — already uses `readRootedFile` (reference pattern)
- `internal/tools/file_commit_test.go` — create-path symlink regression exists; no overwrite/read anchored test for `write_file`

🔒 **Unchanged on main:** `readRootedFile` / `commitRootedFileContents` helpers; `structured_patch` and other tools already on rooted reads.

🔧 **Required correction:** After `openScopedWriteRoot`, drive existence with `root.Stat(rootedRelative)`, conflict/prior bytes with `readRootedFile(root, rootedRelative)` (or the same helper `edit_file` uses), and pass the resulting `FileInfo`/content into `commitRootedFileContents`. Add a regression test (race hook or symlink swap) proving overwrite cannot observe pathname-resolved state that diverges from the rooted object.

🛠️ **Author fix:** Close the root cause on every `In this PR` row in one pass: update all `write_file` stat/read sites after `openScopedWriteRoot`, mirror `edit_file`, and add the missing regression test. Do not patch only the first `os.Stat` line or leave `readFile(absolutePath)` in place.

🚫 **Out of scope:** Rewriting `openScopedWriteRoot`, changing format-on-write formatter argv handling (unchanged vs merge-base), or refactoring unrelated tools that were not part of this PR’s write-root migration.

Vasanthdev2004 (changes requested):
First look from me, at `6c9abec0`. The shell clustering work is right and I could not get past it. `shellCommandFlag` scans the whole cluster rather than only the last character, which is more than the description claims, and it matters: real bash runs the payload for `-cx` as well as `-xc`, and stopping at `o`/`O` matches what bash actually does with `-oc` (it eats the next argument as the option name and never runs the command). `-co` is classified as a command when bash also refuses to run it, which is the safe direction. I tried the clusters, the `--` terminator, `-o xtrace -c`, and an attached payload, and found nothing that reaches the shell without going through the AST.

I agree with jatmn's P1, and I have a measurement that turns it from a consistency argument into a load-bearing one.

**The pathname preimage cannot see a same-name substitution on Windows, so the `os.SameFile` line in `commitRootedFileContents` is inert there.** Replace the destination with a DIFFERENT file carrying the SAME bytes, in the window the hook opens. The byte comparison cannot see it by construction, so only `os.SameFile(expectedInfo, openedInfo)` is left, and `expectedInfo` came from `os.Stat(absolutePath)`:

```
distinct objects before the swap: true
status                          : ok
file now at the name            : "REPLACED BY THE TOOL"
the object that was there       : "original"
```

The write went into an object the tool never observed and it reported success. Go's Windows `fileStat` fills its volume and index fields behind a `sync.Once` that opens the file BY PATHNAME at comparison time, not at `os.Stat` time, so an identity captured before a replacement describes whatever answers to the name afterwards.

This is not yours: `main` gives the same four lines, so it is not a regression and I am not asking you to fix a pre-existing hole. But the fix jatmn is asking for is the fix for this too, and it is the reason to prefer it over leaving the stat where it is. Same three preimages, captured before the swap, compared after:

```
os.Stat      = true    (wrong)
root.Stat    = false   (correct)
handle.Stat  = false   (correct)
```

So `root.Stat(rootedRelative)` is not just the tidier spelling. It is the one that makes the guard you are adding actually hold on Windows. The containment itself is fine either way, `os.Root` refuses to leave the root and the byte comparison fails closed for the ordinary case, which is why this is a P1 on the claim rather than on an escape.

**`--no-config` turns off declarative prettier config too, and the comment above it reads as though it does not.** Prettier documents the flag as "Do not look for a configuration file. The default settings will be used", and that covers every form: `.prettierrc`, `.prettierrc.json`, `.prettierrc.yaml`, the `prettier` key in `package.json`, not only the `.js` and `.cjs` modules the comment is worried about. The comment then says "Declarative configs (editorconfig, .prettierrc) are not executable, so the remaining formatters here keep honoring them", which names `.prettierrc` on the honored side of the sentence while the flag right below it switches `.prettierrc` off.

The effect on a user is that every project pinning `singleQuote`, `semi` or `printWidth` now has its files rewritten to prettier defaults on every save, and its own CI rejects the result. That is the failure mode format-on-write exists to avoid.

I am not asking you to drop the flag. A declarative config can still load a plugin by path, so "declarative means safe" does not hold and a narrower rule would be a false comfort. What I want is for this to be a stated decision rather than something a user discovers from a diff:

- say it in the PR description and in the formatter table comment, in the form "prettier runs on built-in defaults, project prettier config is not honored",
- fix the comment so `.prettierrc` is not on the honored side of that sentence,
- and note the part that surprised me: `--no-editorconfig` is a separate flag and you do not pass it, so `.editorconfig` IS still read. The shipped behaviour is editorconfig yes, prettierrc no, which nobody will guess.

**A refused formatter binary is silent.** `formatterBinaryInWriteRoots` returning true just returns `unformatted`, and the user gets a normal write with no formatting and no reason. For anyone with `node_modules/.bin` on PATH, which is the whole point of the check, format-on-write simply stops working one day. `formatOnWriteResult` already carries a formatter name and a timeout notice; one more field for "skipped, binary resolves inside the workspace" costs very little and turns a mystery into a message.

For whoever sequences merges: this is the third open PR rewriting `internal/tools/format_on_write.go`, after #685 and #941, and all three move it in different directions. #941 in particular replaces the `--write` in-place invocation this builds on with stdin adapters, so the `--no-config` and binary-trust work here will need re-applying whichever lands second.

Verified here at this head on windows/amd64: gofmt clean, `go vet ./...` clean, `internal/sandbox` and `internal/tools` pass, and all nine checks are green.

